PR #4145, opened 2020-08-25
allow random.choice to accept ndarray input

mattjj:
This is a follow-up to #4137. After that change (but before this one) passing a numpy ndarray as the second argument to jax.random.choice would raise an indexing warning/error.

### DIFF
--- a/jax/random.py
+++ b/jax/random.py
@@ -98,6 +98,14 @@ def _is_prng_key(key: jnp.ndarray) -> bool:
 ### utilities
 
 
+# TODO(mattjj,jakevdp): add more info to error message, use this utility more
+def _asarray(x):
+  """A more restrictive jnp.asarray, only accepts JAX arrays and np.ndarrays."""
+  if not isinstance(x, (np.ndarray, jnp.ndarray)):
+    raise TypeError(f"Function requires array input, got {x} of type {type(x)}.")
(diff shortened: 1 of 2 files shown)


jax/random.py
```diff
@@ -98,6 +98,14 @@ def _is_prng_key(key: jnp.ndarray) -> bool:
 ### utilities
 
 
+# TODO(mattjj,jakevdp): add more info to error message, use this utility more
+def _asarray(x):
+  """A more restrictive jnp.asarray, only accepts JAX arrays and np.ndarrays."""
+  if not isinstance(x, (np.ndarray, jnp.ndarray)):
+    raise TypeError(f"Function requires array input, got {x} of type {type(x)}.")
```
jakevdp, line 105: Should we accept python & numpy scalars here as well?

mattjj: I'm not sure; it might depend on how this ends up getting used. I mainly wanted to fix this particular bug (jax.random.choice not accepting ndarrays) for now, so that we could iterate on this `_asarray` method (and maybe something like it in many more places) later.

WDYT?

jakevdp (approved):

jax/random.py
jakevdp: Sounds good.

PR #4152, opened 2020-08-26
Use omnistaging env var even when not using absl flags for config.

tomhennigan:


mattjj (approved):
Great idea! This bit me a while ago.

tomhennigan:
I think this is not quite enough, looks like we need to also call the enablers and that should be done once everything has finished initialising... Any suggestions on where to put that?

mattjj:
Ah, I gave you a brain-dead review on this because I was in a meeting. I messed up!

This attribute `omnistaging_enabled` is meant to indicate that omnistaging is already enabled (and therefore we shouldn't run any omnistaging enablers), rather than that we should enable omnistaging. The latter is just determined by the `jax_omnistaging` config option.

So this PR is currently messed up, as you've noticed...

mattjj:
Ah, I gave you a brain-dead review on this because I was in a meeting. I messed up!

This attribute `omnistaging_enabled` is meant to indicate that omnistaging is already enabled (and therefore we shouldn't run any omnistaging enablers), rather than that we should enable omnistaging. The latter is just determined by the `jax_omnistaging` config option.

So this PR is currently messed up, as you've noticed...

mattjj:
I think we want to make a version of this change lower down, on the line around 117 that says

```python
if FLAGS.jax_omnistaging:
```

we probably want to change that to

```python
if FLAGS.jax_omnistaging or bool_env('JAX_OMNISTAGING', False):
```

That means, in ~English~ American, "let's run the omnistaging enablers if either the config option is set (configured via absl) or if we separately notice the env var is set".

I think...

tomhennigan:
Thanks for the suggestion, have changed the CL so the default value for omnistaging enabled is set based on the env var. The enablers now run immediately if we register them when omnistaging is enabled. This required moving some of them to the end of their files (so that symbols were defined in order that they could be deleted).

mattjj:
Awesome, thanks for sorting that out.

PR #4155, opened 2020-08-26
jax.numpy: change `asarray` to `_strict_asarray`

jakevdp:
This makes more explicit our policy of not silently pushing lists and tuples to device within `jax.numpy` code. The current API is inconsistent on this front.

This extends the initial work in #4145 and uses the utility in more places.

Note that this *may* cause breakages downstream, because lists are no longer accepted in places where they once were.

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -799,7 +799,7 @@ def ldexp(x1, x2):
 @_wraps(np.frexp)
 @jit
 def frexp(x):
-  x = asarray(x)
+  x = _strict_asarray(x)
   if issubdtype(x.dtype, complexfloating):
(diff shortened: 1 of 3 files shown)


froystig (approved):
Nice!

jakevdp:
This breaks ~a few dozen dependent test targets... seems like a lot of people are relying on implicit list to array conversion. Do you think this deserves a deprecation cycle, or should we move fast and break things?

mattjj:
Hrm, tough call. Let's discuss in jax-core chat, or even at our next team meeting?

jakevdp:
I think this is a bit too big to do at once. I'm going to do a more piecemeal approach, and use existing validators (e.g. `_check_arraylike()`) for better error messages.

First change is in #4195

PR #4159, opened 2020-08-27
Fix bug in omnistaging_enabler

gnecula:
This code was failing with "KeyError: psum" for the tests
"//third_party/py/flax/...". I suspect that the error is due to the
ordering of the omnistaging enablers, changed in #4152.

I am not sure of this fix, but this seemed to be enough for all the
presubmit tests to pass and allow the copybara import.


gnecula:
I am merging to enable copybara import, but we should revisit if this is not the right fix.

PR #3903, opened 2020-07-29
WIP: exploring weakly-dtyped device arrays

jakevdp:
This exploration is motivated by a breakage of jit invariance in cases where the working dtype does not match the default dtype. For example:
```python
import jax.numpy as jnp
from jax import jit

x = jnp.arange(4, dtype='float16')

def add(x, y):
  return x + abs(y)

print(add(x, 1.0).dtype)
# float16

print(jit(add)(x, 1.0).dtype)
# float32
```
This is somewhat related to the issue raised in #3840. It is a corner case that shows up whenever the default dtype differs from the working dtype, which comes up much more frequently in the potentialy re-designed x64 flag behavior (in progress at #3773)

The above issue can be boiled down to a simpler demonstration:
```python
print(jnp.dtype(x + abs(1.0)))
# float16

print(jnp.dtype(x + lax.abs(1.0)))
# float32
```
JAX type promotion knows to treat python scalars as less important than JAX arrays, so despite the fact that a python float like `1.0` is 64-bit, the result of adding it to a `float16` is still a `float16`.

On the other hand, the result of `lax.abs(1.0)` is a float32 or float64, depending on the state of the X64 flag, and this type **does** affect the type promotion of the output.

The proposed solution is to allow device arrays to act more like the former case by extending the `weak_dtype` concept that is currently used in abstract arrays, which will allow us to maintain jit invariance even in the presence of lax operations that don't have definitive type information.

PR #4161, opened 2020-08-27
raise_to_shaped: preserve weak_type by default

jakevdp:
As we look at deprecating the X64 flag, the notion of weakly-typed abstract values becomes more important (see #3903). This will require `weak_type` to be preserved within `aval` transformations wherever possible; this PR makes that change for the `raise_to_shaped` utility.

### DIFF
--- a/jax/lax/lax_control_flow.py
+++ b/jax/lax/lax_control_flow.py
@@ -264,16 +264,29 @@ def while_loop(cond_fun: Callable[[T], bool],
       # transformation on it), so we fall back to the primitive version.
       pass
 
-  init_vals, in_tree = tree_flatten((init_val,))
-  init_avals = tuple(_map(_abstractify, init_vals))
-  cond_jaxpr, cond_consts, cond_tree = _initial_style_jaxpr(cond_fun, in_tree, init_avals)
-  body_jaxpr, body_consts, body_tree = _initial_style_jaxpr(body_fun, in_tree, init_avals)
-  if not treedef_is_leaf(cond_tree) or len(cond_jaxpr.out_avals) != 1:
-    msg = "cond_fun must return a boolean scalar, but got pytree {}."
-    raise TypeError(msg.format(cond_tree))
-  if cond_jaxpr.out_avals[0].strip_weak_type() != ShapedArray((), np.bool_):
-    msg = "cond_fun must return a boolean scalar, but got output type(s) {}."
-    raise TypeError(msg.format(cond_jaxpr.out_avals))
+  # The body input and output avals must match exactly. However, we want to account for
+  # the case when init contains weakly-typed values (e.g. Python scalars), with avals that
+  # may not match the output despite being compatible by virtue of their weak type.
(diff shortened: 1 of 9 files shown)


jax/lax/lax_control_flow.py
```diff
@@ -1031,8 +1031,8 @@ def _cond_transpose(cts, *args, branches, linear):
 
   branches_trans = tuple(
       _transpose_cond_jaxpr(jaxpr, num_res) for jaxpr in branches)
-  lin_in_avals = _map(
-      raise_to_shaped, [a for a, l in zip(in_avals, linear) if l])
+  lin_in_avals = [a.strip_weak_type() for a in _map(
```
mattjj, line 1034:
```suggestion
  lin_in_avals = [raise_to_shaped(a).strip_weak_type() for a, l in zip(in_avals, linear) if l]
```

mattjj (approved):
LGTM

jakevdp:
Marking as *pull ready* so I can test internally... this swaps-out the foundation of a lot of operations, and I still have to kick the tires on it a bit.

jax/lax/lax_control_flow.py
```diff
@@ -1213,21 +1211,46 @@ def scan(f, init, xs, length=None):
...
+    jaxpr, consts, out_tree = _initial_style_jaxpr(f, in_tree, carry_avals + x_avals)
+    out_tree_children = out_tree.children()
+    if len(out_tree_children) != 2:
+      msg = "scan body output must be a pair, got {} in {} pass."
+      raise TypeError(msg.format(tree_unflatten(out_tree, jaxpr.out_avals), pass_))
+    carry_avals_out = jaxpr.out_avals[:out_tree_children[0].num_leaves]
+
+    # weak type logic run only on the first pass.
+    if pass_ != 'first':
+      break
```
mattjj, line 1237: Another way to structure this code, rather than the loop over `['first', 'second']`, might be to factor the jaxpr formation part into a function, then just call it once, check for weak type mismatches, and call it again if needed. I think that might be cleaner than the loop with breaks. We can try that out later if you'd rather land this immediately, though.

mattjj (approved):
New changes LGTM! One suggestion on code organization but it can wait if you want.

jax/lax/lax_control_flow.py
jakevdp: I initially did that, but the subsequent code uses 7 or 8 of the intermediate values computed in the function. Returning all of those seemed error-prone. I thought about defining a custom namedtuple to organize all of those, but that seems overly-engineered. I also thought about using `nonlocal` to let all those values leak passively into the parent scope, but that struck me as confusing. I finally landed on the loop approach as the least-worst option.

mattjj: I think a function returning a tuple is probably the best approach. I don't think it's error-prone because it's a function with effectively a single call site.

jakevdp: OK, I'll make an update.

jakevdp: #4512

PR #4166, opened 2020-08-28
Implement a proper shape checking rule for gather.

bchetioui:
The implementation is based on the corresponding shape inference
code in `tensorflow/compiler/xla/service/shape_inference.cc`. The
tests added in `tests/lax_test.py` are similarly mirroring the
corresponding tests in tensorflow, with slight adaptations for the
particular setting of JAX. Fixes google/jax#2826, and in principle
fixes google/jax#4154, fixes google/jax#3905, fixes google/jax#4003.

PR #4167, opened 2020-08-28
Addition of one more conclusive polynomial comparison case.

bchetioui:
In the case when the difference between two polynomials is a
constant, it is possible to conclusively compare them. This commit
adds such a case to `masking.Poly.__ge__`.

I ran into such a case in the tests of #4166 ([link](https://github.com/google/jax/pull/4166/checks?check_run_id=1040679020)). This would fix the issue.


gnecula (approved):
Looks good, thanks!

PR #4168, opened 2020-08-28
Add support for axis_index inside vmap

apaszke:
Also, reorganize the code to put all `axis_index` related functions in
`lax_parallel.py`, next to all other parallel collectives.

### DIFF
--- a/jax/lax/lax_parallel.py
+++ b/jax/lax/lax_parallel.py
@@ -628,8 +628,6 @@ def _axis_index_translation_rule(c, *, axis_name, axis_env, platform):
 def _axis_index_bind(*, axis_name):
   dynamic_axis_env = pxla._thread_local_state.dynamic_axis_env
   frame = dynamic_axis_env[axis_name]
-  sizes = dynamic_axis_env.sizes[:dynamic_axis_env.index(frame)+1]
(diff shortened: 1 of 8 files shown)


jax/lax/lax_parallel.py
```diff
@@ -579,6 +618,41 @@ def all_gather(x, axis_name):
...
 
+def _axis_index_translation_rule(c, *, axis_name, axis_env, platform):
+  div = xb.constant(c, np.array(axis_env.nreps // prod(axis_env.sizes),
+                                dtype=np.uint32))
+  mod = xb.constant(c, np.array(axis_env.sizes[-1], dtype=np.uint32))
+  unsigned_index = xops.Rem(xops.Div(xops.ReplicaId(c), div), mod)
+  return xops.ConvertElementType(unsigned_index, xb.dtype_to_etype(np.int32))
+
+def _axis_index_bind(*, axis_name):
+  dynamic_axis_env = pxla._thread_local_state.dynamic_axis_env
```
mattjj, line 629: Could add a comment for our future selves explaining that axis_index has a custom bind rule to make it work with vmap.

mattjj (approved):
Fantastic! Thanks for doing both omni and non-omni versions :)

PR #4089, opened 2020-08-18
Add a fast C++ jit codepath.

jblespiau:
This starts a C++ jit codepath to speed up dispatch time.
It will be enabled only after we support all features and extensive
testing within Google.

Supported features:
- scalar, numpy array and DeviceArray argument support:
  - integer, floats, boolean, and complex scalars arguments are supported.
  - The jax_enable_x64 flag will be used at object-creation type to cast scalars and numpy arrays.
  - The Jax `weak_type` attribute for arguments is supported (DeviceArray and scalars).
- The donate_argnums argument.
- Use an XLA tuple for more than 100 arguments

Unsupported features:
- jax._cpp_jit on methods e.g
    @functools.partial(jax.jit, static_argnums=0)
    def _compute_log_data(self, ...)
      ...
  This is currently not supported by the C++ codepath, because "self" won't be automatically added.
- disable_jit.

This is currently only for the XLA execution path, not the tracing path.

PR #4169, opened 2020-08-28
[Ready to be submitted] Add more features to the C++ jax.jit.

jblespiau:
This mainly follows https://github.com/google/jax/pull/4089 by adding:

- support for disable_jit from C++
- support for jax._cpp_jit on methods.
- supporting applying @jax.jit on top-level functions, by delaying the retrieval of the device and backend.
- concurrency support.

Now remains the unknowns bugs and missing features. I will be hunting them down until all tests pass. I think some differences are due to the differences between xla_computation and _xla_callable.)

benchmark

See:

- https://i.ibb.co/ZMvZ4nK/benchmark.png for the benchmarking comparison (see
 cr/328899906 + benchmarks for how numbers were generated)
- The results of the Jax tests when enabling this:
http://sponge2/4a67d132-209f-45c5-ab7b-83716d329ec2 (110 fails, 92 passes, but many common cause of failure).

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -35,6 +35,7 @@ import numpy as np
 from contextlib import contextmanager, ExitStack
 
 from . import core
+from . import lib
 from . import linear_util as lu
 from . import ad_util
 from . import dtypes
@@ -230,10 +231,10 @@ def _cache_for_cpp_jit(call):
   fun_caches = weakref.WeakKeyDictionary()
 
   def memoized_fun(fun: Callable,
-                   static_argnums=(),
+                   static_argnums: Union[int, Iterable[int]] = (),
                    device=None,
                    backend=None,
-                   donate_argnums=()):
+                   donate_argnums: Union[int, Iterable[int]] = ()):
     _check_callable(fun)
     # We need tuples if we want them to be hashable.
     static_argnums = _ensure_tuple(static_argnums)
@@ -253,11 +254,10 @@ def _cache_for_cpp_jit(call):
 @_cache_for_cpp_jit
 def _cpp_jit(
     fun: Callable,
-    static_argnums: Union[int, Iterable[int]] = (),
+    static_argnums: Iterable[int] = (),
     device=None,
     backend: Optional[str] = None,
-    donate_argnums: Union[int, Iterable[int]] = ()
-) -> Callable:
+    donate_argnums: Iterable[int] = ()) -> Callable:
   """An implementation of `jit` that tries to do as much as possible in C++.
 
   The goal of this function is to speed up the time it takes to process the
@@ -267,8 +267,6 @@ def _cpp_jit(
   the C++ code will fallback to `_python_jit` when it faces some unsupported
   feature.
   """
-  # TODO(jblespiau): Add support for `disable_jit.`
-
   # `_check_callable(fun)` and the normalization of the argnums are in
   # `_cache_for_cpp_jit`.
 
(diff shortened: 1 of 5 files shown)


jblespiau:
TODO: Update to use a tensorflow version after the fix in cr/328926660

jax/api.py
```diff
@@ -297,22 +296,77 @@ def cache_miss(*args, **kw):
...
+
+class _IsMethodDecorator:
+  """Detects whether `fun` is a method and resolves the instance.
+
+  A function is decorated *before* it is bound, so we cannot directly
+  determine whether it is a 'method' or 'function'. We can rely on the
+  Descriptor protocol (see below) to detect, at *call time* whether it was a
+  method and retrieve the associated object.
+
+  We prefer to do this only when we think it' needed as the goal is to remove
```
gnecula, line 349: Typo: it' -> it's

jax/api.py
```diff
@@ -297,22 +296,77 @@ def cache_miss(*args, **kw):
...
+
+  https://docs.python.org/3/howto/descriptor.html#functions-and-methods
+  """
+
+  def __init__(self, func):
+    self.func = func
+    self.is_method = False
+
+  def __get__(self, instance, owner):
+    if not self.is_method:
```
gnecula, line 360: Is it important to have this conditional?

jax/api.py
```diff
@@ -297,22 +296,77 @@ def cache_miss(*args, **kw):
...
+  # hashable, so maybe such usage will be forbidden in the future.
+  # Prefer for example an attribute or property access:
+  # class A:
+  #   def __init__(self):
+  #     self.jitted_f = jax.jit(self.f)
+
+  # We try to do this more expensive path only when it's needed and assume
+  # users respect the `self` convention.
+  # See `_IsMethodDecorator`
+  # Note that @classmetod and @staticmethod are *not* supported.
```
gnecula, line 334: I wonder what happens if the method is a @classmethod or @staticmethod. Would there be a reasonable error message?

jax/api.py
```diff
@@ -297,22 +296,77 @@ def cache_miss(*args, **kw):
...
     return (
         # We call xla.backend_compile to ensure the XLA computation appears
         # separately in Python profiling results.
-        xla.backend_compile(backend, xla_result.xla_computation, options),
+        xla.backend_compile(backend_, xla_result.xla_computation, options),
         xla_result.out_pytree_def,
         xla_result.shaped_arrays,
         xla_result.lazy_expressions)
 
   # Delay the import, because it requires a new version of jaxlib.
```
gnecula, line 311: nit: I think that this deserves a 
```
# TODO(jblespiau): Remove when C++ jit has landed (jaxlib.version >= 0.1.54)
```

gnecula (approved):

gnecula:
Very nice, thanks! Is this ready to merge?

jax/api.py
jblespiau: Done.

jax/api.py
jblespiau: I guess not, removed.

jax/api.py
jblespiau: Added a test.

Both classmethod and staticmethod will raise an exception (also in the current version), because they won't be callable objects.

jax/api.py
jblespiau: Done.

jblespiau:
> Very nice, thanks! Is this ready to merge?

Still running tests internally, I think I have one failure. I will confirm when it works.

jblespiau:
> > Very nice, thanks! Is this ready to merge?
> 
> Still running tests internally, I think I have one failure. I will confirm when it works.

Updated tensorflow with the fix. It works internally: http://sponge2/8c06a190-71a8-4efd-9e15-a75817fcca24

PR #4170, opened 2020-08-28
Allow vmapping all_to_all and implement a (slow) CPU and GPU translation

### LINKED ISSUE #4141: No batching rule for `all_to_all`
```py
import jax
from jax import lax, vmap, pmap
import jax.numpy as jnp

def f(x):
  tiled_x = jax.tree_map(lambda x: jnp.tile(x[None], [jax.device_count()] + [1] * len(x.shape)), x)
  all_x = lax.pswapaxes(tiled_x, 'batch', 0)
  return all_x
pmap(vmap(f), axis_name='batch')(jnp.arange(jax.device_count() * 5 * 2).reshape((jax.device_count(), 5, 2)))
```

Currently this throws an assertion error elsewhere but with #4140, it will correctly throw `NotImplementedError: Batching rule for 'all_to_all' not implemented`.

apaszke:
This allows pmapping vmapped computations that use `all_to_all` or
`pswapaxes` inside. It also includes a very slow CPU and GPU translation
rule that might be useful for debugging programs locally, since XLA only
implements the `AllToAll` collective on TPUs.

Fixes #4141.

### DIFF
--- a/jax/lax/lax_parallel.py
+++ b/jax/lax/lax_parallel.py
@@ -16,6 +16,7 @@ Parallelization primitives.
 """
 
 import collections
+import warnings
 
 import numpy as np
 
@@ -243,9 +244,7 @@ def pswapaxes(x, axis_name, axis):
       ``axis_name``.
 
   Returns:
-    Array(s) with shape ``np.insert(np.delete(x.shape, axis), axis, axis_size)``
-    where ``axis_size`` is the size of the mapped axis named ``axis_name`` in
-    the input ``x``.
+    Array(s) with the same shape as ``x``.
   """
   return all_to_all(x, axis_name, axis, axis)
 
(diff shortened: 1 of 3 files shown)


jax/lax/lax_parallel.py
```diff
@@ -531,11 +546,23 @@ def _moveaxis(src, dst, x):
...
 
+def _all_to_all_batcher(vals_in, dims_in, *, axis_name, split_axis, concat_axis):
+  x, = vals_in
+  d, = dims_in
+  if split_axis <= d:
+    split_axis += 1
+  if concat_axis <= d:
+    concat_axis += 1
+  result = all_to_all_p.bind(x, axis_name=axis_name, split_axis=split_axis, concat_axis=concat_axis)
+  return result, d
```
apaszke, line 557: Hm I just realized that I might need to adjust `d` here...

jax/lax/lax_parallel.py
```diff
@@ -497,12 +496,28 @@ def _ppermute_batcher(vals_in, dims_in, axis_size, axis_name, perm):
 batching.collective_rules[ppermute_p] = _ppermute_batcher
 
 
+def _all_to_all_via_psum(x, *, axis_name, split_axis, concat_axis):
+  zero_full = lax.full((x.shape[split_axis],) + x.shape, 0, dtype=x.dtype)
+  idx = axis_index(axis_name)
+  local_full = lax.dynamic_update_slice_in_dim(zero_full, x[None], idx, 0)
+  global_full = psum(local_full, axis_name=axis_name)
```
mattjj, line 503: I'm not sure, but can you call `_allgather` here? It seems similar.

jax/lax/lax_parallel.py
mattjj: Maybe try to add a test that fails?

mattjj (approved):
LGTM, modulo the batching rule you're looking at.

jax/lax/lax_parallel.py
Thenerdstation: I think it should be 

```python
# Vmapped axis got shifted down one.
if split_axis < concat_axis: 
  if split_axis <= d and concat_axis >= d:
    d -= 1
# Vmapped axis got shifted up one. 
if split_axis > concat_axis:
  if split_axis >= d and concat_axis <= d:
    d += 1
# In the case where the are equal, leave d alone.
```

Thenerdstation:
Bump on this.

jax/lax/lax_parallel.py
apaszke: Nice catch, I used `all_gather` in the latest implementation

PR #4171, opened 2020-08-28
Use xla.abstractify in xla_computation.

jblespiau:
The previous version was dropping weak-types for scalars.


jblespiau:
This one is broken.

PR #3583, opened 2020-06-27
add lax.associative_scan to docs

mattjj:
Also tweaked the docstring to say "arrays" instead of "Tensors".

PR #4172, opened 2020-08-28
lax.associative_scan: fix docstring examples

### LINKED ISSUE #4165: Documentation issue - associative_scan
https://github.com/google/jax/blob/e95d5701e33bda30299a8c87aa43b61a686c1b01/jax/lax/lax_control_flow.py#L2352

Hi,

It should clearly be used from lax, and not numpy + it seems that the Example format does not follow the docstring style, so it doesn't show properly in the readthedocs page.

Adrien

jakevdp:
Fixes #4165


jakevdp:
Oh, looks like there's already some work on this: #3583

mattjj:
How about we kill #3583 in favor of this one? You could steal any useful bits from it.

PR #4174, opened 2020-08-28
applied simple find+sed for 'master' -> 'main'

mattjj:
I just ran

```
find . -name '*.py' -exec sed -i 's/master_/main_/g' '{}' \;
find . -name '*.py' -exec sed -i 's/Master/Main/g' '{}' \;
```


jax/core.py
```diff
@@ -687,9 +687,9 @@ def cur_sublevel() -> Sublevel:
   return thread_local_state.trace_state.substack[-1]
 
 @contextmanager
-def new_master(trace_type: Type[Trace], bottom=False) -> Generator[MasterTrace, None, None]:
+def new_master(trace_type: Type[Trace], bottom=False) -> Generator[MainTrace, None, None]:
```
jakevdp, line 690: Maybe also do `find . -name '*.py' -exec sed -i 's/_master/_main/g' '{}' \;`

mattjj: Yeah, good call. Roy pointed out some others too:

```
jax/core.py:    self.master = master
jax/core.py:    self.level = master.level
jax/core.py:    if val._trace.master is self.master:
```

I've got to set this aside for a while but if anyone else gets a chance, feel free to push commits onto this branch!

gnecula:
I had trouble pushing to this branch, so I created #4178

PR #4025, opened 2020-08-11
Add GMRES to `jax.scipy.sparse.linalg`

gehring:
Implementation of GMRES based off of @shoyer's [prototype](https://gist.github.com/shoyer/dc33a5850337b6a87d48ed97b4727d29) augmented with:
- support for pytrees
- looped with termination condition
- scipy-like `restart` argument (at least my best interpretation of it. I had a hard time parsing scipy's source code for GMRES)

To do:
- [ ] Improve numerical stability. I can't currently get it to reliably pass tests when not running in Float64 mode, i.e., `jax_enable_x64 == True`.
- [ ] Add documentation
- [ ] Add tests
- [ ] Make sure scipy's implementation behaves similarly

Relevant issues:
- #1531

Issue #1531, opened 2019-10-20
Differentiable matrix-free linear algebra, optimization and equation solving


shoyer:
This is a meta-issue for keeping track of progress on implementing differentiable higher-order functions from SciPy in JAX, e.g.,

- [ ] `scipy.sparse.linalg.gmres` and `cg`: matrix-free linear solves
- [ ] `scipy.sparse.linalg.eigs` and `eigsh`: matrix-free eigenvalue problems
- [ ] `scipy.optimize.root`: nonlinear equation solving
- [ ] `scipy.optimize.fixed_point`: solving for fixed points
- [ ] `scipy.integrate.odeint`: solving ordinary differential equations
- [ ] `scipy.optimize.minimize`: nonlinear minimization

These higher-order functions are important for implementing sophisticated differentiable programs, both for scientific applications and for machine learning.

Implementations should leverage and build upon JAX's custom transformation capabilities. For example, `scipy.optimize.root` should leverage autodiff for calculating the Jacobians or Jacobian-vector products needed for Newton's method.

In most cases, I think the right way to do this involves two separate steps, which could happens in parallel:
1. Higher order primitives for defining automatic differentiation rules, but not specialized to any particular algorithm, e.g., `lax.custom_linear_solve` from https://github.com/google/jax/pull/1402.
2. Implementations of particular algorithms for the forward problems, e.g., a conjugate gradient method for linear solves. These could either be implemented from scratch using JAX's functional control flow (e.g., `while_loop`) or could leverage existing external implementations on particular backends. Either way they will almost certainly need custom derivative rules, rather than differentiation through the forward algorithm.

There's lots of work to be done here, so please comment if you're interested in using or implementing any of these.

shoyer:
I'll start things off.

I've written initial versions of higher order primitives for non-linear root finding (see discussion in https://github.com/google/jax/issues/1448) and linear equation solving (https://github.com/google/jax/pull/1402).

I've also written experimental implementations in JAX for `gmres` and a Thick-Restart Lanczos method (for `eigsh`). These are not shared publicly yet, but I could drop them into `jax.experimental`.

gehring:
@shoyer With this addition and, in general, implicit diff related features, are there any plans for a mechanism to extract intermediate/aux values when differentiating to allow us to log things like number of iterations for tangent solve, residual, etc.? (related: #1197 #844)

shoyer:
I'm glad to hear @jekbradbury is thinking about, because I have not! I agree it's important.

[...]

PR #4175, opened 2020-08-28
Gmres qr

alewis:
A synthesis of @gehring's and @shoyer's GMRES implementation #4025 with the one in TensorNetwork. 

Per those authors, this implementation supports both preconditioning and pytrees, but I've modified it to also use a more efficient QR implementation of the main Arnoldi loop that builds the QR decomposition of the Krylov matrix instead of the matrix itself. This greatly reduces the overhead of the final linear solve, since the relevant system is now triangular. It also allows the iteration to terminate early if convergence is reached midway through the Arnoldi iterations.

The behaviour of the maxiter and restart arguments has also been modified (I think) to mirror that in SciPy. Finally, a few bugs have been fixed such that the implementation passes some simple tests, but I haven't tested very extensively. 

@gehring @shoyer, it would be great if you had a look!

Relevant issues: #1531


jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,254 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  givens_factors = givens_rotation(R_row[k], R_row[k + 1])
+  givens = givens.at[k, :].set(givens_factors)
+  cs_k, sn_k = givens_factors
+
+  R_row = R_row.at[k].set(cs_k * R_row[k] - sn_k * R_row[k + 1])
+  R_row = R_row.at[k + 1].set(0.)
+  return R_row, givens
+
+
+def _gmres(A, b, x0, tol, n_kry, M):
```
shoyer, line 273: I don't love the `n_kry` name, which took me a while to figure out. Could you stick with the name `maxiter`, or maybe something more verbose like `n_krylov` or `krylov_dim`?

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,254 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  """
+
+  v = tree_map(lambda x: x[..., k], V)  # Gets V[:, k]
+  v = A(M(v))
+
+  def gram_schmidt_step(r, v_i):
+    h_i = _vdot_tree(r, v_i)
+    r_i = _sub(r, _mul(h_i, v_i))
+    return r_i, h_i
+  v, h = lax.scan(gram_schmidt_step, v, xs=V.T)
```
shoyer, line 232: It looks like this scan is scanning over the columns of `V`?

One key implementation detail from https://github.com/google/jax/pull/4025 / https://github.com/google/jax/issues/1531#issuecomment-657164860 is the use of a matrix/vector product rather than a loop. As I noted in that comment, "This is perhaps 10-20% faster on CPU and ~10x faster on GPUs". So I suspect this is missing that aspect.

My version also used two iterations of classical gram Schmidt, which apparently can be important for ill-conditioned matrices per https://slepc.upv.es/documentation/reports/str1.pdf

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,254 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    err = jnp.abs(sn) / b_norm
+    return k + 1, err, V, R, beta_vec, givens
+
+  carry = (0, beta, V, R, beta_vec, givens)
+  carry = lax.while_loop(loop_cond, arnoldi_qr_step, carry)
+  k, err, V, R, beta_vec, _ = carry
+  converged = k < n_kry - 1
+
+  y = jsp.linalg.solve_triangular(R[:, :-1].T, beta_vec[:-1])
+  dx = M(tree_map(lambda X: jnp.dot(X[..., :-1], y), V))
```
shoyer, line 320: This is a bit of a "gotcha" currently but if you're doing linear algebra on TPUs it's important to set `precision=lax.Precision.HIGHEST` whenever doing dot-product. Otherwise it defaults to bfloat16, which is usually not enough.

jax/scipy/sparse/test_gmres.py
```diff
@@ -0,0 +1,152 @@
+import numpy as np
```
shoyer, line 1: This test file should go in `tests/`, for consistency with JAX's other tests.

jax/scipy/sparse/test_gmres.py
```diff
@@ -0,0 +1,152 @@
...
+    return A @ x
+  tol = A.size*jax.numpy.finfo(dtype).eps
+  x, _ = linalg.gmres(A_mv, b, x0=x0, tol=tol, atol=tol, restart=n_kry,
+                      maxiter=maxiter)
+  solution = jax.numpy.array([2., 1.], dtype=dtype)
+  try:
+    np.testing.assert_allclose(x, solution, atol=tol)
+  except AssertionError as e:
+    print("Failed at dtype ", dtype)
+    print(e)
```
shoyer, line 34: Generally you should not be adding these assertion wrappers :)

Instead, use parameterized tests (see the existing test suite for examples).

jax/scipy/sparse/test_gmres.py
```diff
@@ -0,0 +1,152 @@
...
+#    cs, sn = gmres.givens_rotation(*v)
+#    rot = np.zeros((2, 2), dtype=dtype)
+#    rot[0, 0] = cs
+#    rot[1, 1] = cs
+#    rot[0, 1] = -sn
+#    rot[1, 0] = sn
+#    rot = jax.numpy.array(rot)
+#    result = rot @ v
+#    tol = 4*jax.numpy.finfo(dtype).eps
+#    np.testing.assert_allclose(result[-1], 0., atol=tol)
```
shoyer, line 152: One other unit test that would be nice to have is a consistency check with SciPy's GMRES after a small fixed number of iterations (before convergence). I have a version of that in my gist: https://gist.github.com/shoyer/dc33a5850337b6a87d48ed97b4727d29

shoyer:
> I've modified it to also use a more efficient QR implementation of the main Arnoldi loop that builds the QR decomposition of the Krylov matrix instead of the matrix itself. This greatly reduces the overhead of the final linear solve, since the relevant system is now triangular. It also allows the iteration to terminate early if convergence is reached midway through the Arnoldi iterations.

This makes complete sense! I think it may also improve numerical stability.

On the other hand, `apply_givens_rotation` is another un-vectorized loop, so this _might_ significantly slow things down on accelerators. If that is the case, then we might want to have two implementations: the Given rotation version allowing for early termination, and the version that runs for some fixed number of iterations.

To inform this decision, it would be helpful to do have microbenchmarks results, similar to those I had in https://gist.github.com/shoyer/dc33a5850337b6a87d48ed97b4727d29

gehring:
@googlebot I consent.

jax/scipy/sparse/linalg.py
alewis: I think the relevant parameter is actually very confusingly named `restart`, which is why I favour mentioning the Krylov dimension explicitly. However, I do support either of your two options - perhaps `krylov_dim` slightly more.

jax/scipy/sparse/linalg.py
alewis: I had not realized this was so carefully considered, and will restore the old code.

jax/scipy/sparse/linalg.py
alewis: Sounds good, thank you!

jax/scipy/sparse/test_gmres.py
alewis: Ok, I'll write parameterized tests there.

jax/scipy/sparse/test_gmres.py
alewis: Looks great

alewis:
You are right the extra loop could conceivably lead to a large speedup on accelerators in certain use cases - those in which the matvec operation does not entirely dominate the expense. However, a GPU user may also encounter a case where this operation does dominate, but the ability to terminate the Arnoldi iterations early is important (for example, because GMRES is to be run on many different problems in sequence over the course of an iteration, and the appropriate incoming Krylov dimension is therefore not known. I would therefore, in the case that the original implementation were found to be sometimes favourable, recommend that the user be provided with both options when switching to the GPU.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,254 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+def _safe_normalize(x, return_norm=False, thresh=None):
+  norm = jnp.sqrt(_vdot_tree(x, x, assume_real=False))
+  if thresh is None:
+    thresh = jnp.finfo(x.dtype).eps
+  thresh = thresh.astype(norm.dtype)
+
+  normalized_x, norm = lax.cond(
+    norm > thresh,
+    lambda y: (_div(y, norm), norm),
+    lambda y: (_mul(0.*y, thresh), thresh),  # To get the dtype right
```
gehring, line 206: `0.*y` will fail if `y` is a pytree.
```suggestion
    lambda y: tree_map(jnp.zeros_like, y),  # To get the dtype right
```

alewis:
Just letting you all know I'm back from vacation (which is where I've been) and am working on this again. Updates to follow once I've made them.

alewis:
@mganahl Your comments don't seem to have attached themselves to the relevant lines of code, so I'm not sure exactly what you're referring to, but in sequence:

- All print commands will indeed likely be removed;
- This implementation of GMRES allows b and x0 to be a data structure called a pytree that I honestly don't understand very well (it might be something to look into for symmetric tensors however). Along with a few other quirks, the tree_map statements are there to support this structure.
- Early stopping of Arnoldi should I think be redundant with early stopping of GMRES. But we do have a tolerance threshold (you're right to notice it currently does nothing) that stops the normalizations from blowing up just in case.

alewis:
I've added better tests and done a bit of fiddling. The present implementation appears to give comparable (actually slightly better) error as SciPy when it works, but is also much more unstable than SciPy, failing completely in many cases that SciPy handles easily (for example, in most cases when maxiter > 1, and even for maxiter == 1 in some cases). The reason for this is not yet clear.

shoyer:
It might be make sense to look over SciPy's implementation to see if it handles any edge cases that we don't handle here. The source code for SciPy's iterative solvers can be found here:
https://github.com/scipy/scipy/tree/master/scipy/sparse/linalg/isolve

The GMRES implementation uses a mix of Python and Fortran:
https://github.com/scipy/scipy/blob/914523af3bc03fe7bf61f621363fca27e97ca1d6/scipy/sparse/linalg/isolve/iterative.py#L432
https://github.com/scipy/scipy/blob/914523af3bc03fe7bf61f621363fca27e97ca1d6/scipy/sparse/linalg/isolve/iterative/GMRESREVCOM.f.src

(You might also or alternatively consider implementing L-GMRES, which I've found to be generally a superior alternative to GMRES with restarts.)

alewis:
The SciPy code isn't the easiest on Earth to read, is it? Still, I'm hopeful because the TensorNetwork version does not suffer these instabilities. Lemme poke a bit more.

gehring:
Current, I've identified 2 failure modes in the previous. The 1st was the normalization causing nans and is already handled, but the second I only recently identified and is caused by trying to solve a zero/singular matrix once the basis was constructed (a problem that is made more common when thresholding the normalization). It might be worth checking if that is what is happening here. The fix should be straightforward if that is the case.

alewis:
Excellent, this is very likely the problem. This may have something to do
with the input matrix having large condition number. That, however, is not
the full picture because I find the algorithm consistently breaks down when
A is the identity matrix, which is obviously invertible.

gehring:
I've also encountered it when the residual is very small right from the start, e.g. the b vector is close to zero or the initial guess is very good. This was obviously an issue for the batched version but maybe this also affects this implementation if it always runs at least one iteration.

alewis:
I think I know the problem: in the TensorNetwork version, the final solve
is sliced to include only the submatrix formed during the Arnoldi process.
In this code this is not (easily) possible since everything is traced, so
the final solve involves the full Arnoldi matrix, which is mostly zeroes in
the case of early termination. The resulting problem can thus very easily
be singular.

To solve it we need to find a way to involve only the relevant submatrix in
the solve, a tricky proposition given that the number of iterations is a
traced value.

gehring:
Ah, that makes sense! Would initializing the triangular matrix with the identity instead of zeros then simply throwing/masking out the dummy variables after the fact be a reasonable workaround?

alewis:
Yes, that might well work. I will try when I get a second.

alewis:
It works! You're a genius.

gehring:
Awesome!

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,306 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  norm = jnp.sqrt(jnp.abs(_vdot_tree(x, x, assume_real=False)))
+  if thresh is None:
+    thresh = jnp.finfo(x.dtype).eps
+  thresh = thresh.astype(norm.dtype)
+
+  normalized_x, norm = lax.cond(
+    norm > thresh,
+    lambda y: (_div(y, norm), norm),
+    lambda y: (tree_map(jnp.zeros_like, y),
+               (thresh*0.).astype(norm.dtype)),  # To get the dtype right
```
gehring: This should work and is a bit more direct
```suggestion
               jnp.zeros((), dtype=norm.dtype)),  # To get the dtype right
```

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,306 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+
+
+def _safe_normalize(x, return_norm=False, thresh=None):
+  """
+  Returns the L2-normalized vector (which can be a pytree) x, and optionally
+  the computed norm. If the computed norm is less than the threshold `thresh`,
+  which by default is the machine precision of x's dtype, it will be
+  taken to be 0, and the normalized x to be the zero vector.
+  """
+  norm = jnp.sqrt(jnp.abs(_vdot_tree(x, x, assume_real=False)))
```
gehring, line 206: I would `_vdot_tree` to always return a non-negative number. Is this not the case?

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,306 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  residual = _sub(b, A(x0))
+  unit_residual, beta = _safe_normalize(residual, return_norm=True)
+
+  V = tree_map(
+    lambda x: jnp.pad(x[..., None], ((0, 0),) * x.ndim + ((0, restart),)),
+    unit_residual,
+  )
+  #R = jnp.zeros((restart, restart + 1), jnp.result_type(*tree_leaves(b)))
+  R = jnp.eye(restart, restart + 1, dtype=jnp.result_type(*tree_leaves(b)))
+  b_norm = jnp.sqrt(jnp.abs(_vdot_tree(b, b, assume_real=False)))
```
gehring, line 326: See other comment about `_vdot_tree`

jax/scipy/sparse/linalg.py
alewis: It's to cast to real

shoyer: I think this should use `assume_real=True` then? `assume_real` appears to refer to the output rather than the input variables.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +183,307 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    beta_vec = beta_vec.at[k + 1].set(sn)
+    err = jnp.abs(sn) / b_norm
+    return k + 1, err, V, R, beta_vec, givens
+
+  carry = (0, beta, V, R, beta_vec, givens)
+  carry = lax.while_loop(loop_cond, arnoldi_qr_step, carry)
+  k, err, V, R, beta_vec, _ = carry
+
+  y = jsp.linalg.solve_triangular(R[:, :-1].T, beta_vec[:-1])
+  Vy = tree_map(lambda X: _dot(X[..., :-1], y), V)
```
gehring, line 356: This likely will never have any noticeable impact but I'll mention it just for the sake of keeping things as simple as possible. Using `_dot` here might be unnecessary. Is there a reason why it was changed from `jnp.dot`, e.g., `tree_map(lambda X: jnp.dot(X[..., :-1], y), V)`?

jax/scipy/sparse/linalg.py
gehring: I've never used complex numbers in practice so I'm far from familiar with the conventions surrounding how they are handled numerically. I'm the one who added the `assume_real` argument here since I wasn't quite sure what the impact of the assumptions of the original implementation were. I opted to play it safe and check with @shoyer later but I never got that far with that first PR.

All that to say that it's very possible that we could just revert `_vdot_tree` to the original one without the `assume_real` argument. Someone would just have to validate that it makes sense.

alewis: Getting rid of assume_real makes sense to me, yeah (in normal array world, I would usually write x = np.abs(np.vdot(a, b)) if I wanted a real number).

jax/scipy/sparse/linalg.py
alewis: _dot has the precision explicitly specified, and I didn't want to keep typing it out

alewis:
What's the certified way to do `assert a <= b` in a test? Right now I've had to resort to `assert` and exception handling.

jax/scipy/sparse/linalg.py
gehring: Ah, right, I forgot about that. Good catch!

jax/scipy/sparse/linalg.py
```diff
@@ -28,25 +28,9 @@
...
-  # https://en.wikipedia.org/wiki/Definiteness_of_a_matrix#Definitions_for_complex_matrices
-  result = _vdot(x.real, y.real)
-  if jnp.iscomplexobj(x) or jnp.iscomplexobj(y):
-    result += _vdot(x.imag, y.imag)
-  return result
-
-
 # aliases for working with pytrees
-
-def _vdot_tree(x, y, assume_real=True):
```
gehring: Does this change affect the CG implementation? I believe it used the `assume_real=True` branch. Would the CG calls to `_vdot_tree` need to be wrapped with `jnp.abs` as well?

alewis: Well, the tests seem to still pass, but I'll take a look

shoyer:
> What's the certified way to do `assert a <= b` in a test? Right now I've had to resort to `assert` and exception handling.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLess

jax/scipy/sparse/linalg.py
shoyer: Yes, we definitely want to still use `_vdot_real_part` for CG. Otherwise we are doing a bunch of redundant computation for the imaginary part of the solution, even though we know a priori that it is zero.

alewis: OK, fair enough. I think the assume_real is a little confusing though, so I made this two different functions.

alewis:
I did some benchmarking. CPU performance seems to be comparable to SciPy, but the mystery Colab GPU performs much (~ 10X) more slowly. The GPU runtimes appear to be almost independent of the matrix size, which strongly suggests something about the GMRES overhead (rather than the matvec itself) is the culprit. I haven't checked the QR version against the fixed-iteration one yet.

https://colab.research.google.com/drive/1DZBFNQeXiDuiyPxXRCJFnuSknaUzB1HU?usp=sharing

alewis:
The difference between the fixed and QR implementations seems to be minor; in addition CPU performance is actually a bit slower than SciPy. Perhaps there's some sort of vectorization trickery I'm missing?

shoyer:
> The difference between the fixed and QR implementations seems to be minor; in addition CPU performance is actually a bit slower than SciPy. Perhaps there's some sort of vectorization trickery I'm missing?

You want to be careful to always `jit` the outer-most function when you are benchmarking in JAX, and to use `block_until_ready()`. I did a few tweaks to make that work in your notebook:
https://gist.github.com/shoyer/2f86513877fcbf044cea964060dd8de3

Otherwise you are mostly measuring JIT compilation time, not computation time.

alewis:
Alright, well I don't quite understand why I'm unable to reproduce your profiling here, but based on your notebook we seem to have the following results at the given parameters:

- On the CPU, the QR implementation is significantly (~30%) faster than the fixed one, even when we don't terminate early.
- On the GPU, the QR implementation is about 10x slower than the fixed one.
- The final linear solve consumes about 40% of the overall runtime.

The latter result would seem to indicate that SciPy's ability to dynamically resize the Arnoldi matrix potentially confers a significant advantage, and thus our workaround is perhaps not ideal (although it may be unavoidable). In addition, it seems that maintaining both implementations is desirable.

It is worth pointing out that the case where the matvec operator multiplies by a known dense matrix is kind of a weird edge case - unless the matrix were known to be well approximated by a low-dimensional Krylov space, a dense solver would probably be preferred. So the large impact of the GMRES overhead compared to the matvec itself may be misleading. Certainly this is the case in TensorNetwork applications.

alewis:
The "fixed" implementation could be made to end the Arnoldi process once an invariant subspace was found, without necessarily involving the QR loop. This would probably have similar properties re: early termination as the QR one without breaking the GPU so badly. It would still be worth keeping both I think, since the QR has other advantages.

alewis:
I've noticed that both CG and gmres currently store the Arnoldi vectors as the columns of V rather than the rows. Since by default we presumably use row-major linear ordering this seems potentially problematic, especially on the GPU. Would it be worth transposing the program?

alewis:
I found a bug I need to fix before this is ready; but does anyone know why the Read the Docs build is failing?

alewis:
Ok, this one seems to work. However I'm still not sure what's going on with the documentation.

jax/scipy/sparse/linalg.py
```diff
@@ -17,68 +17,89 @@
...
+
 def _vdot_real_part(x, y):
-  """Vector dot-product guaranteed to have a real valued result."""
+  """Vector dot-product guaranteed to have a real valued result despite
+     possibly complex input. Thus neglects the real-imaginary cross-terms.
+     The result is a real float.
+  """
   # all our uses of vdot() in CG are for computing an operator of the form
-  # `z^T M z` where `M` is positive definite and Hermitian, so the result is
+  #  z^T M z
```
rmlarsen, line 38: This is the same as z^H M z = conj(z)^H M z, which is the natural inner product for complex-valued vector spaces.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,377 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  """Orthogonalize x against the columns of Q."""
+  # "twice is enough"
+  # http://slepc.upv.es/documentation/reports/str1.pdf
+
+  # This assumes that Q's leaves all have the same dimension in the last
+  # axis.
+  r = jnp.zeros((tree_leaves(Q)[0].shape[-1]))
+  q = x
+
+  for _ in range(iterations):
```
rmlarsen, line 249: You only really need the second iteration if the vectors are nearly linearly dependent. A robust test is to only do the second iteration of ||r||  < 1/sqrt(2) ||x||.

tests/lax_scipy_sparse_test.py
```diff
@@ -60,7 +61,34 @@ def rand_sym_pos_def(rng, shape, dtype):
...
+      M = np.linalg.inv(rand_sym_pos_def(rng, A.shape, A.dtype))
+    elif preconditioner == 'exact':
+      M = np.linalg.inv(A)
+    else:
+      M = None
+
+    if M is None or not return_function:
+      return M
+    else:
+      return lambda x: jnp.dot(M, x, precision=lax.Precision.HIGHEST)
```
rmlarsen, line 90: Why is the highest precision needed here? The preconditioner is often an approximation of the inverse of A anyway.

rmlarsen: Ah, I see this is just for testing. Fair enough.

mattjj:
Achievement unlocked: @rmlarsen lending his expertise to JAX linear algebra 🥇 Woohoo!

jax/scipy/sparse/linalg.py
alewis: I think the T here was meant to denote Hermitian conjugation, but I'll change the docstring

tests/lax_scipy_sparse_test.py
alewis: We do use the highest precision throughout though, so it's a fair point

alewis:
The main thing holding me back here is that I don't understand why Read the Docs is failing - I've never used it before and the error messages in this case are not very helpful, at least to me. Does anyone have an idea?

jax/scipy/sparse/linalg.py
alewis: Addressed in the latest commit

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,392 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  """
+  v_proj = tree_multimap(
+    lambda X, y: jnp.einsum("...n,...->n", X.conj(), y),
+    A,
+    v,
+  )
+  return tree_reduce(operator.add, v_proj)
+
+
+def _iterative_classical_gram_schmidt(Q, x, iterations=2):
```
rmlarsen, line 239: rename `iterations` to `max_iterations`

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,392 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    q = _sub(q, Qh)
+    r = _add(r, h)
+    return (k + 1, q, r)
+
+  def cond_function(carry):
+    k, _, r = carry
+    _, rnorm = _safe_normalize(r, return_norm=True)
+    return lax.cond(
+        k < (iterations - 1),
+        lambda x: rnorm < xnorm_scaled,
```
rmlarsen, line 264: For `max_iterations > 2` it is not sufficient to check against the original norm scaled. You need to check whether ||r|| < 1/sqrt(2) ||x|| in every iteration. It is true that "twice is enough" (almost always), but you might as well implement the more general condition, since (if you skip it for k==max_iterations) it will not add extra cost.

jax/scipy/sparse/linalg.py
alewis: Done

jax/scipy/sparse/linalg.py
alewis: I had to complicate the function a lot to realize this while staying functional (since `cond` evaluates both branches), but done

alewis:
This might be ready to merge!

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,434 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    cs, sn = givens[k, :] * beta_vec[k]
+    beta_vec = beta_vec.at[k].set(cs)
+    beta_vec = beta_vec.at[k + 1].set(sn)
+    err = jnp.abs(sn) / b_norm
+    return k + 1, err, V, R, beta_vec, givens
+
+  carry = (0, residual_norm, V, R, beta_vec, givens)
+  carry = lax.while_loop(loop_cond, arnoldi_qr_step, carry)
+  k, residual_norm, V, R, beta_vec, _ = carry
+  _ = k # Until we figure out how to pass this to the user.
```
shoyer, line 414: Maybe use `del` here for unused variables? That enforces that you can't use `k`.
```suggestion
  del k  # Until we figure out how to pass this to the user.
```

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,434 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+
+  if x0 is None:
+    x0 = tree_map(jnp.zeros_like, b)
+  if M is None:
+    M = _identity
+
+  try:
+    size = sum(bi.size for bi in tree_leaves(b))
+  except AttributeError:
+    size = len(tree_leaves(b))
```
shoyer, line 587: If you use `  b, x0 = device_put((b, x0))` first (like in `cg`), then you won't need this fall-back condition.

shoyer (approved):
This looks great!

My main suggestion is to try to revert some of the pervasive use of `lax.cond` with `jnp.where` instead. This should result in much improved performance on accelerators, especially GPUs.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,434 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  else:
+    return normalized_x
+
+
+def _project_on_columns(A, v):
+  """
+  Returns A.T.conj() @ v.
+  """
+  v_proj = tree_multimap(
+      lambda X, y: jnp.einsum("...n,...->n", X.conj(), y),
```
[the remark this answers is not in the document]
gehring: I original wrote it this way to handle a `b` vector that is a pytree of arrays that aren't 1D. I added this to allow "inverting" some hvp related to a NN where the `b` vector is a pytree of the gradients.

I'd have to look at it more carefully to remind myself why I thought it was necessary but I believe we'd have to change how the basis vectors are stored if we wanted this to be 2D. If I recall correctly I was trying to avoid having to ravel/unravel linear systems produced by vjps/jvps.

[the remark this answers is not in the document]
alewis: On my end I don't really understand how pytrees work and in similar places was trying to basically copy @gehring's idioms as best as possible

jax/scipy/sparse/linalg.py
```diff
@@ -17,68 +17,89 @@
...
+  #  z^H M z
+  #  where M is positive definite and Hermitian, so the result is
   # real valued:
   # https://en.wikipedia.org/wiki/Definiteness_of_a_matrix#Definitions_for_complex_matrices
   vdot = partial(jnp.vdot, precision=lax.Precision.HIGHEST)
   result = vdot(x.real, y.real)
   if jnp.iscomplexobj(x) or jnp.iscomplexobj(y):
     result += vdot(x.imag, y.imag)
-  return result
+  return result.real
```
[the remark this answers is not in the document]
alewis: Without the .real, I believe the result will be a complex float whose imaginary part is 0j. The .real casts it to a plain old float, thus halving the memory footprint.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,434 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    thresh = jnp.finfo(norm.dtype).eps
+  thresh = thresh.astype(dtype).real
+
+  normalized_x, norm = lax.cond(
+      norm > thresh,
+      lambda y: (_div(y, norm), norm),
+      lambda y: (tree_map(jnp.zeros_like, y), jnp.zeros((),
+                                                        dtype=thresh.dtype)),
+      x,
+  )
```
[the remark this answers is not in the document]
alewis: I just hadn't realized `where` was preferred. I'll look it over and have a think about it.

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +188,434 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+    Qh = tree_map(lambda X: _dot_tree(X, h), Q)
+    q = _sub(q, Qh)
+    r = _add(r, h)
+
+    def qnorm_cond(carry):
+      k, not_done, _, _ = carry
+      return lax.cond(not_done,
+                      lambda x: k < (max_iterations - 1),
+                      lambda x: False,
+                      0)
```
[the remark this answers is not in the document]
alewis: I think this might need to be wrapped in a function (not necessarily cond), but I'll try it

jax/scipy/sparse/linalg.py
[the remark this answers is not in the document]
shoyer: `result` is either `vdot(x.real, y.real)` or `vdot(x.real, y.real) + vdot(x.imag, y.imag)` here. In either case, `vdot()` is applied to purely real (float) arrays, so I believe the result should always be a 0d float array.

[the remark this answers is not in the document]
alewis: Yes, fair enough, I think you're right.

jax/scipy/sparse/linalg.py
alewis: Fixed

alewis:
Ok, this one should have all the cond statements removed, mostly in favour of jnp.logical_and (just `and` doesn't seem to work).

alewis:
How's this looking?

jax/scipy/sparse/linalg.py
```diff
@@ -167,10 +191,419 @@ def cg(A, b, x0=None, *, tol=1e-5, atol=0.0, maxiter=None, M=None):
...
+  dtype = jnp.result_type(*tree_leaves(b))
+  H = jnp.eye(restart, restart + 1, dtype=dtype)
+
+  def loop_cond(carry):
+    _, _, breakdown, k = carry
+    return jnp.logical_and(k < restart, jnp.logical_not(breakdown))
+    #  return lax.cond(k < restart,
+    #                  lambda x: ~x,
+    #                  lambda x: False,
+    #                  breakdown)
```
shoyer, line 443: delete?

jax/scipy/sparse/linalg.py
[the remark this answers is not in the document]
shoyer: Any luck trying `where` here?

(We can also do this in a follow up -- this shouldn't block submitting this PR)

shoyer (approved):
Sorry for the delayed review! I think we should go ahead and merge this. It's well tested and looks quite useful. There is likely room for future improvements, but that will be easier to do in follow-up PRs :).

shoyer:
could you kindly approve the changes in https://github.com/google/jax/pull/4832?

I just squashed a bunch of commits and fixed up / simplified `_safe_normalize` a little bit

shoyer:
This was merged in #4832

rmlarsen:
Woohoo!

shoyer:
See https://github.com/google/jax/pull/5034 for a few follow-on fixes / refinements to the interface.

PR #4180, opened 2020-08-31
[jax2tf] Turn on with_gradient by default

gnecula:
As I was writing the demo I realized that it makes more sense for
with_gradient to be set to True by default.

I have also fixed a bug with tie_in in omnistaging.


bchetioui (approved):
LGTM, thanks!

PR #4184, opened 2020-08-31
Bug fix in axis checking in batching.py

hamzamerzic:
A series of PRs renaming the frame entries have been submitted, one of them introducing a bug when using omnistaging. This PR fixes that and removes a print comment (assuming added for debugging purposes).


gnecula:
@apaszke I am going to merge this right now, because it is failing many tests, but please check that this is the right fix.

@hamzamerzic Thank you for the fix!

apaszke (commented):
Yes, this looks good! Thanks!

PR #4192, opened 2020-09-02
[jax2tf] Implementation of random_gamma

gnecula:
The simplest implementation is by converting the JAX own impl_rule,
which rewrites gamma into other JAX primitives.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -134,7 +134,7 @@ def _tfval_add_unit(vals: Sequence[TfValOrUnit],
 tf_impl: Dict[core.Primitive,
               Callable[..., Any]] = {}
 
-def convert(fun, with_gradient=True):
+def convert(fun, with_gradient=False):
   """Transforms `fun` to be executed by TensorFlow.
 
   Args:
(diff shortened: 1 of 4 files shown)


jax/random.py
```diff
@@ -988,18 +988,19 @@ def _gamma_grad(sample, a):
...
-def _gamma_impl(key, a):
+def _gamma_impl(key, a, use_vmap=False):
   a_shape = jnp.shape(a)
   # split key to match the shape of a
   key_ndim = jnp.ndim(key) - 1
   key = jnp.reshape(key, (-1, 2))
   key = vmap(split, in_axes=(0, None))(key, prod(a_shape[key_ndim:]))
   keys = jnp.reshape(key, (-1, 2))
   alphas = jnp.reshape(a, -1)
-  if xla_bridge.get_backend().platform == 'cpu':
```
gnecula, line 999: I do not plan to merge this PR until I understand better the reason why we don't use vmap on CPU. Presumably it does not help to use vmap over map, but does it hurt?

gnecula: @hawkinsp I would appreciate your insight in this.

shoyer: My guess is that on CPU the loop has good performance (and saves a bit of memory) but it's much slower on GPU.

This might be a nice usecase for a "gmap"

gnecula: @shoyer That must be it: the memory saving.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -218,6 +218,30 @@ def _interpret_fun(fun: lu.WrappedFun,
...
+  Args:
+    jax_impl: typically the impl-rule for a primitive, with signature
+      `(*args: JaxVal, **kwargs) -> Sequence[JaxVal]`. This function implements
+      a primitive in terms of other primitives.
+    multiple_results: whether `jax_impl` returns a sequence of results.
+
+  Returns:
+     a function with signature `(*args: TfValOrUnit, **kwargs) -> Sequence[TfValOrUnit]`.
+  """
+  def wrapped(*tf_args: TfValOrUnit, **kwargs) -> Sequence[TfValOrUnit]:
```
SIben, line 233: Is it OK to make this return `Union[TfValOrUnit, Sequence[TfValOrUnit]]` to satisfy mypy, as when `multiple_results == False` then the return type is not a Sequence.

bchetioui (approved):
LGTM

jax/experimental/jax2tf/jax2tf.py
gnecula: I'll try that, but in the past we got strange failures from pytype when we started to use more `Union`.

PR #4193, opened 2020-09-02
[jax2tf] Draft of a generator for the documentation of operations with limited support (WIP)

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +127,119 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+  def __call__(self, *args, **kwargs):
+    if not self.instance:
+      self.instance = self.cls(*args, **kwargs)
+    return self.instance
+
+PrimitiveName = str
+ErrorType = str
+ErrorString = str
+Devices = Tuple[str, ...]
+Limitation = Tuple[PrimitiveName, ErrorType, ErrorString, Devices]
```
gnecula, line 145: Maybe it is better to use a collections.namedtuple here?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +127,119 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+  all_devices = ["CPU", "GPU", "TPU"]
+
+  def _report_failure(error_type: str, msg: str,
+                      devs: Sequence[str] = all_devices) -> None:
+    limitations.append((prim.name, error_type, msg, tuple(devs)))
+
+  tf_unimpl = functools.partial(_report_failure, "Missing TF support")
+
+  if prim in [lax.min_p, lax.max_p]:
+    tf_dtype = args[0].dtype
```
gnecula, line 172: This suggests that args are `TfVal`. This is fine for use in tests, but not for use in classifying usage. In that case the arguments are JAX values. That is why I had suggested that perhaps `np.shape` and `np.dtype` works for all kinds of values that we care about.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +127,119 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+    output_file: the file in which to write the summary of the encountered
+      limitations. If None, the summary is written to stdout. (default: None).
+  """
+  def __init__(self, inner_categorizer: Callable = categorize,
+               inner_prettifier: Callable = prettify,
+               output_file: Optional[str] = None):
+    self.all_limitations: Sequence[Limitation] = []
+    self.inner_categorizer = inner_categorizer
+    self.inner_prettifier = inner_prettifier
+    atexit.register(self.pprint_limitations, output_file)
```
gnecula, line 219: I would not put something with such global power (`atexit`) into the constructor of the class. I think that this belongs in the test harness somewhere.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +127,119 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+
+  Args:
+    prim: the primitive to call.
+    args: the arguments to pass to prim.
+    kwargs: the keyword arguments to pass to prim.
+
+  Returns:
+    A list of limitations
+  """
+  limitations: List[Limitation] = []
```
gnecula, line 162: Shouldn't this list be at a more global scope?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -356,7 +470,7 @@ def post_process_map(self, map_primitive, out_tracers, params):
 
   def get_primitive_impl(self, p):
     try:
-      return tf_impl[p]
+      return Categorizer().wrap(p, tf_impl[p])
```
gnecula, line 473: I am not yet sold on the need for the singleton. 
I don't quite see how the categorizer is invoked. Wouldn't it be enough for the test harness to overwrite get_primitive_impl with a categorizer? Then people who read the code don't have to wonder what the categorizer is doing (since it does not matter outside of tests).

jax/experimental/jax2tf/jax2tf.py
SIben: I agree 100%. I actually just put this here so it'd be somehow self-contained and easy to understand the draft. I'll make that change.

jax/experimental/jax2tf/jax2tf.py
SIben: The way this is executed, the args are indeed `TfVal` for now; maybe I should make it explicit in the type signature?

jax/experimental/jax2tf/jax2tf.py
SIben: My intention was to keep `categorize` self-contained; the list that is at a more global scope right now is `Categorizer().all_limitations`. If we decide that we want to simply populate a global list instead, then we can do that, but I was not really sold on this idea.

jax/experimental/jax2tf/jax2tf.py
gnecula: But then I cannot use this function to count usage in JAX. If I use it there, the args are JAX tracers. That is Ok, as long as we have a way to get the shape and dtype that works for both.

jax/experimental/jax2tf/jax2tf.py
SIben: Addressed!

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -29,6 +30,19 @@ class JaxToTfTestCase(jtu.JaxTestCase):
 
   def setUp(self):
     super().setUp()
+
+    # Initialize categorizer
+    categorizer = jax2tf.jax2tf.Categorizer()
+    # Making sure the exit handler is not registered several times
+    atexit.unregister(categorizer.pprint_limitations)
+    # Register limitation summary atexit printing handler
+    atexit.register(categorizer.pprint_limitations, None)
```
gnecula, line 39: I suspect that this will register multiple printers, one for each test.

gnecula: I see that you are unregistering it first. Why not do everything at global level somehow: create the categorizer, register the printer? Maybe use this: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

jax/experimental/jax2tf/jax2tf.py
gnecula: I did not see where still is added to `all_limitations`

jax/experimental/jax2tf/jax2tf.py
SIben: FYI, `np.dtype(dtype)` fails if `dtype` is a TensorFlow dtype (one needs to get `dtype.as_numpy_dtype` first, or better, `to_jax_dtype(dtype)`).
I think I addressed your comment in my latest commit (?).

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: Thanks, I was actually looking for something like that – but missed it

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +126,122 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+    self.inner_prettifier = inner_prettifier
+
+  def wrap(self, prim: core.Primitive, func: Callable) -> Callable:
+    """
+    Wraps a primitive and its corresponding TF implementation with the inner
+    categorizer.
+    """
+    def wrapper(*args, **kwargs):
+      limitations = self.inner_categorizer(prim, *args, **kwargs)
+      self.all_limitations += limitations
```
SIben, line 230: @gnecula here is where the limitations are added to the state within the singleton

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -26,6 +27,17 @@
...
 class JaxToTfTestCase(jtu.JaxTestCase):
+  @classmethod
+  def setUpClass(cls):
+    # Initialize categorizer
+    categorizer = jax2tf.jax2tf.Categorizer()
+    # Register limitation summary atexit printing handler
+    atexit.register(categorizer.pprint_limitations, None)
+    # Monkey-patch jax2tf.TensorFlowTrace.get_primitive_impl to wrap the
+    # resulting primitive in the categorizer.
+    original_impl = jax2tf.jax2tf.TensorFlowTrace.get_primitive_impl
```
gnecula, line 38: I think it would be better to restore the `original_impl` on exit from this class.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -26,6 +27,17 @@
 from jax import numpy as jnp
 
 class JaxToTfTestCase(jtu.JaxTestCase):
+  @classmethod
+  def setUpClass(cls):
+    # Initialize categorizer
+    categorizer = jax2tf.jax2tf.Categorizer()
+    # Register limitation summary atexit printing handler
+    atexit.register(categorizer.pprint_limitations, None)
```
gnecula, line 35: We must have a flag or environment variable to guard the `atexit.register`. I generally prefer an environment variable for something that needs to be checked only once. There are going to be executions when we do not want the atexit handler to run, e.g., inside automated tests in Google (any kind of file writing will fail unless it is directed to some special directories), or when we run selected tests manually for debugging.

SIben: Yes, this was always the final intent; I'll add it now.

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: Right, good catch!

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -26,6 +27,17 @@
 from jax import numpy as jnp
 
 class JaxToTfTestCase(jtu.JaxTestCase):
+  @classmethod
+  def setUpClass(cls):
+    # Initialize categorizer
+    categorizer = jax2tf.jax2tf.Categorizer()
```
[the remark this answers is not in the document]
SIben: Surely we can do that, and I think it makes sense when it comes to "just" documentation generation. The goal was to have more flexibility (as discussed in the design doc) to work with other categorizers, stat generators, or even test generators; in that case, it either becomes inconvenient to define functions working with one single global state (as it needs to be accessible in all the files where the state is defined), or we end up having several loci aggregating different types of data.
That was my rationale, but I see now that the plan was not to make such considerations right now, so I'll make a change in the direction you suggest.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -25,7 +26,28 @@
 from jax import test_util as jtu
 from jax import numpy as jnp
 
+import os
+
 class JaxToTfTestCase(jtu.JaxTestCase):
+  @classmethod
+  def setUpClass(cls):
```
gnecula, line 33: I would move this whole thing to global scope. As it is now, every test case will setup its own collection and would overwrite the same file. If this is in the global scope, there is no need to restore the hooks.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -126,6 +126,95 @@ def add_unit(v: TfValOrUnit, aval: core.AbstractValue):
...
+    args: the arguments to pass to prim.
+    kwargs: the keyword arguments to pass to prim.
+
+  Returns:
+    A list of limitations
+  """
+  limitations: List[Limitation] = []
+  all_devices = ["CPU", "GPU", "TPU"]
+
+  def _report_failure(error_type: str, msg: str,
```
gnecula, line 154: I think it may make sense to actually run this as it is intended to generate the report, and to include the report in the PR.

I would also change one of the tests with `min` and `max` to see what would it take for this solution to be used to control the test failure. Ideally, we would see that we can remove this conditional from the test.

SIben: Addressed both of these in the next commits.
I also added a flag for test generation without file generation, as I assume the tests may fail in Google otherwise.

jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -0,0 +1,14 @@
+# Primitives with limited support
+
+| Affected primitive | Type of limitation | Description | Devices affected |
```
gnecula, line 3: Looks good

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -25,8 +26,22 @@
 from jax import test_util as jtu
 from jax import numpy as jnp
 
-class JaxToTfTestCase(jtu.JaxTestCase):
+import os
+
+# JAX2TF_CATEGORIZE_OUT is either not set, or contains the name of the
+# target file for the output.
+output_categories = os.getenv('JAX2TF_CATEGORIZE_OUT')
```
gnecula, line 33: Wouldn't it be easier to use if the file name is hard-coded and the path is computed relative to the current file?

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -402,17 +402,7 @@ def test_add_mul(self, harness: primitive_harness.Harness):
 
   @primitive_harness.parameterized(primitive_harness.lax_min_max)
   def test_min_max(self, harness: primitive_harness.Harness):
-    expect_tf_exceptions = False
-    dtype = harness.params["dtype"]
-
-    if dtype in [np.bool_, np.int8, np.uint16, np.uint32, np.uint64,
-                 np.complex64, np.complex128]:
-      # TODO(bchetioui): tf.math.maximum and tf.math.minimum are not defined for
```
gnecula, line 410: Nice cleanup! Some of these TODO's deserve to be moved to the categorizer, but probably not this one.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -111,12 +128,23 @@ def run_tf(mode):
...
+
+        if generate_tests_from_categorize:
+          new_limitations = (
+            jax2tf.jax2tf.all_limitations[len(current_limitations):])
+          detected_tf_exception = any(map(is_tf_exception, new_limitations))
+
+        if not (expect_tf_exceptions or detected_tf_exception):
           raise e
         else:
           print(f"Encountered expected exception for mode={mode}: {e}")
```
gnecula, line 150: Can we print here the messages from the relevant limitations?

SIben: Done in `4cb23e6`

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -402,38 +402,18 @@ def test_add_mul(self, harness: primitive_harness.Harness):
 
   @primitive_harness.parameterized(primitive_harness.lax_min_max)
   def test_min_max(self, harness: primitive_harness.Harness):
-    expect_tf_exceptions = False
-    dtype = harness.params["dtype"]
```
gnecula, line 406: In the design doc you made the strong point that this code occurs 3 times. I am curious if the technique works to remove all those uses? I suspect that it will work in some sense, but the error message is not going to be about `reduce_window` but about `max`.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -111,14 +125,27 @@ def run_tf(mode):
       else:
         assert False
 
+    def is_tf_exception(lim: jax2tf.jax2tf.Limitation):
+      return (lim.ErrorType == 'Missing TF support' and
+              self.tf_default_device.device_type in lim.Devices)
+
     result_tf = None
     for mode in ("eager", "graph", "compiled"):
+      current_limitations = jax2tf.jax2tf.all_limitations[:]
```
gnecula, line 134: I am wondering if it makes sense to move `current_limitations` outside the loop. What if one of these executions has cached the jax2tf.convert? Then we do not see any limitations. 

At the same time, this gives me the idea that it may be helpful to document of the limitations still occur with `exeprimental_compile=True`, I suspect that many do not. Maybe that is for the next PR.

jax/experimental/jax2tf/tests/primitives_test.py
SIben: Correct, this is my interpretation as well. Which is fine, since the limitation is really coming from `max`.

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: This seems a bit tricky. If we are 100% sure that they should always have the same limitations, this'd make sense to move out `current_limitations` outside. Otherwise, as you point out, it may not be the case, and then I think this deserves some thinking.

PR #4194, opened 2020-09-02
Add axis_index_groups support to all_gather.

inoryy:
Ensures `all_gather` works correctly while utilizing replica groups. I've changed `_expand` API a bit to make the conditional index clamping easier but afaik it's only used in `_all_gather` anyway.

In a test environment with 8 devices with the typical `[[0, 1, 2, 3], ..., [28, 29, 30, 31]]` input the expected output would be of shape `(8, 4, 4)` with first half containing four copies of `[[0, 1, 2, 3], ..., [12, 13, 14, 15]]` followed by four copies of `[[16, 17, 18, 19], ..., [28, 29, 30, 31]]` for the second half.

### DIFF
--- a/jax/lax/lax_parallel.py
+++ b/jax/lax/lax_parallel.py
@@ -579,17 +579,17 @@ ad.deflinear(all_to_all_p, _all_to_all_transpose_rule)
 pxla.multi_host_supported_collectives.add(all_to_all_p)
 
 
-def _expand(dim, size, axis_name, x):
+def _expand(dim, size, index, x):
   shape = list(x.shape)
   shape.insert(dim, size)
   out = lax.full(shape, lax._const(x, 0))
-  return lax.dynamic_update_index_in_dim(out, x, axis_index(axis_name), dim)
+  return lax.dynamic_update_index_in_dim(out, x, index, dim)
 
-def _allgather(x, dim, size, axis_name):
-  outs = tree_util.tree_map(partial(_expand, dim, size, axis_name), x)
-  return psum(outs, axis_name)
+def _allgather(x, dim, size, index, axis_name, axis_index_groups=None):
+  outs = tree_util.tree_map(partial(_expand, dim, size, index), x)
+  return psum(outs, axis_name, axis_index_groups=axis_index_groups)
 
-def all_gather(x, axis_name):
+def all_gather(x, axis_name, *, axis_index_groups=None):
   """Gather values of x across all replicas.
 
   If ``x`` is a pytree then the result is equivalent to mapping this function to
@@ -601,13 +601,17 @@ def all_gather(x, axis_name):
     x: array(s) with a mapped axis named ``axis_name``.
     axis_name: hashable Python object used to name a pmapped axis (see the
       :func:`jax.pmap` documentation for more details).
+    axis_index_groups: optional list of lists containing axis indices (e.g. for
+      an axis of size 4, [[0, 1], [2, 3]] would run all gather over the first
+      two and last two replicas). Groups must cover all axis indices exactly
+      once, and all groups must be the same size.
 
   Returns:
     Array(s) representing the result of an all-gather along the axis
     ``axis_name``. Shapes are the same as ``x.shape``, but with a leading
     dimension of the axis_size.
 
-  For example, with 2 XLA devices available:
+  For example, with 4 XLA devices available:
 
   >>> x = np.arange(4)
   >>> y = jax.pmap(lambda x: jax.lax.all_gather(x, 'i'), axis_name='i')(x)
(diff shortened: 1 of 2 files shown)


tests/pmap_test.py
```diff
@@ -494,6 +494,29 @@ def sum_helper(a):
     ans = f(x)
     self.assertAllClose(ans, expected, check_dtypes=False)
 
+  def testGatherReplicaGroups(self):
+    replicas = xla_bridge.device_count()
+    if replicas % 2 != 0:
+      raise SkipTest
```
gnecula, line 500: It would be nice to add a message for `SkipTest`. I see that some tests in this file don't do that, but it is still a good thing to do.

jax/lax/lax_parallel.py
```diff
@@ -601,6 +604,10 @@ def all_gather(x, axis_name):
     x: array(s) with a mapped axis named ``axis_name``.
     axis_name: hashable Python object used to name a pmapped axis (see the
       :func:`jax.pmap` documentation for more details).
+    axis_index_groups: optional list of lists containing axis indices (e.g. for
+      an axis of size 4, [[0, 1], [2, 3]] would run all gather over the first
+      two and last two replicas). Groups must cover all axis indices exactly
+      once, and all groups must be the same size.
```
gnecula, line 610: While you are changing this, I think we should change below "with 2 XLA devices available" to "with 4 XLA devices available".

jax/lax/lax_parallel.py
```diff
@@ -579,17 +579,20 @@ def _moveaxis(src, dst, x):
...
-  return lax.dynamic_update_index_in_dim(out, x, axis_index(axis_name), dim)
+  return lax.dynamic_update_index_in_dim(out, x, index, dim)
 
-def _allgather(x, dim, size, axis_name):
-  outs = tree_util.tree_map(partial(_expand, dim, size, axis_name), x)
-  return psum(outs, axis_name)
+def _allgather(x, dim, size, axis_name, axis_index_groups=None):
+  index = axis_index(axis_name)
+  if axis_index_groups is not None:
+    index = index % size
```
gnecula, line 591: This area is new to me, but I am wondering if this clamping computation is correct. I assume that this is needed because `axis_index` is agnostic of the `axis_index_groups`. Then I suspect that the clamping computation is not correct. E.g., if `axis_index_groups=[[0, 2], [1, 3]`. Am I misunderstanding something?

jax/lax/lax_parallel.py
```diff
@@ -617,7 +624,8 @@ def all_gather(x, axis_name):
    [0 1 2 3]
    [0 1 2 3]]
```
gnecula, line 625: Maybe add an example with `axis_index_groups`?

jax/lax/lax_parallel.py
inoryy: The reason is indeed because `axis_index` doesn't account for `axis_index_groups`.
As a workaround I've added a separate `axis_index_to_group_index` kwarg which provides the necessary mapping from global to group index. I've also added a test case that covers your example.

jax/lax/lax_parallel.py
inoryy: Done.

tests/pmap_test.py
inoryy: Done.

jax/lax/lax_parallel.py
inoryy: Done.

jax/lax/lax_parallel.py
```diff
@@ -601,13 +607,19 @@ def all_gather(x, axis_name):
     x: array(s) with a mapped axis named ``axis_name``.
     axis_name: hashable Python object used to name a pmapped axis (see the
       :func:`jax.pmap` documentation for more details).
+    axis_index_groups: optional list of lists containing axis indices (e.g. for
+      an axis of size 4, [[0, 1], [2, 3]] would run all gather over the first
+      two and last two replicas). Groups must cover all axis indices exactly
+      once, and all groups must be the same size.
+    axis_index_to_group_index: list of indices mapping from global axis index
```
gnecula, line 614: Since`axis_index_to_group_index` is uniquely determined by `axis_index_groups`, I feel that it is redundant to ask the user to provide the argument. I think that a better approach is to add code to compute it. I think it makes sense to have this code run at tracing time (use `np` instead of `jnp` for the calculations).

tests/pmap_test.py
```diff
@@ -494,6 +494,66 @@ def sum_helper(a):
     ans = f(x)
     self.assertAllClose(ans, expected, check_dtypes=False)
 
+  def testGatherReplicaGroups(self):
```
gnecula, line 497: I would parameterize this test to run with two groups, to ensure that the case we fixed is working.

jax/lax/lax_parallel.py
inoryy: Done.

tests/pmap_test.py
inoryy: I've added the case we discussed in the `testGatherReplicaGroupsInterleaved`. 
It's a bit tricky to generalize it into parameterized test case since the expected outputs need to be prepared differently.

jax/lax/lax_parallel.py
```diff
@@ -616,8 +620,37 @@ def all_gather(x, axis_name):
...
+  ... x, 'i', axis_index_groups=[[0, 2], [1, 3]]))(x)
+  >>> print(y)
+  [[[ 0.  1.  2.  3.]
+    [ 8.  9. 10. 11.]]
+   [[ 4.  5.  6.  7.]
+    [12. 13. 14. 15.]]
+   [[ 0.  1.  2.  3.]
+    [ 8.  9. 10. 11.]]
+   [[ 4.  5.  6.  7.]
+    [12. 13. 14. 15.]]]
```
gnecula, line 642: Nice!

jax/lax/lax_parallel.py
```diff
@@ -616,8 +620,37 @@ def all_gather(x, axis_name):
...
   """
-  return _allgather(x, 0, psum(1, axis_name), axis_name)
+
+  index = axis_index(axis_name)
+  if axis_index_groups is not None:
+    indices = np.array(axis_index_groups).flatten()
+    axis_index_to_group_index = indices.argsort() % len(axis_index_groups[0])
```
gnecula, line 648: This is clever, perhaps too clever! It is not obvious that it works. What is the intended result for all_gather with groups? If `index_groups=[[2, 1], [3, 0]]` then should the result be exactly the same as when the index groups are sorted? (`index_groups=[[1, 2], [0, 3]]`)? Essentially in both cases we should have `to_group_index=[0, 0, 1, 1]`? 

how would you argue that this code is correct?

inoryy: But the intra-group order matters when gathering, no? As an end-user I'd definitely expect the result for the two cases to be different which is only possible if `to_group_index` preserves the ordering of `axis_index_groups`.

gnecula: The semantics is up to us. It is Ok to gather respecting the intra-group order, as you propose. In that case, I recommend changing slightly the example in the documentation to have the index groups not sorted intra-group.

inoryy: Makes sense, updated the example to illustrate the importance of intra-group ordering.

PR #4201, opened 2020-09-03
[jax2tf] Cleanup the correctness stats layout.

bchetioui:


jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -0,0 +1,115 @@
+import functools
```
gnecula, line 1: This file needs the License notice at the top. See the other files

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -0,0 +1,115 @@
...
+import numpy as np
+import tensorflow as tf # type: ignore[import]
```
gnecula, line 3: Would it be possible to remove the dependency on tf for this module? It is not important, but I feel that we are adding such a big dependency for so little benefit.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -23,22 +23,16 @@
 from jax.config import config
 from jax import dtypes
 from jax.experimental import jax2tf
+from jax.experimental.jax2tf.tests import correctness_stats as jcs
```
gnecula, line 26: I think that the Google style rule would prefer:
```
from jax.experimental.jax2tf.tests import correctness_stats
```

We do not strictly use the Google style rules in this project, but I think in this case there is little benefit from the more cryptic `jcs` to be worth it (there are few uses).

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -100,6 +94,16 @@ def ConvertAndCompare(self, func_jax: Callable, *args,
...
         as in JAX.
     """
+    original_impl = jax2tf.jax2tf.TensorFlowTrace.get_primitive_impl
+
+    def patch_get_primitive_impl():
+      wrapper = jcs.collect_limitations
+      jax2tf.jax2tf.TensorFlowTrace.get_primitive_impl = ( # type: ignore
+        lambda s, p: wrapper(p, original_impl(s, p)))
+
+    def restore_get_primitive_impl():
```
gnecula, line 104: There is a useful pattern that you can use here. Check out https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -125,18 +129,24 @@ def run_tf(mode):
...
 
-    def is_tf_exception(lim: jax2tf.jax2tf.Limitation):
+    def is_tf_exception(lim: jcs.Limitation):
       return (lim.ErrorType == 'Missing TF support' and
               self.tf_default_device.device_type in lim.Devices)
 
     result_tf = None
     for mode in ("eager", "graph", "compiled"):
-      current_limitations = jax2tf.jax2tf.all_limitations[:]
+      current_limitations = jcs.all_limitations[:]
```
gnecula, line 138: nit: perhaps it is worth saving just the `len(current_limitations)`. This makes it a little clearer to the user that the old limitations are not really needed below.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -125,18 +129,24 @@ def run_tf(mode):
...
+      # Monkey-patch jax2tf.TensorFlowTrace.get_primitive_impl to wrap the
+      # resulting primitive in a categorizer. We do that at every loop
+      # iteration because we want to restore the original implementation
+      # before returning, and the code inside the loop may throw exceptions
+      # and thus return early.
+      patch_get_primitive_impl()
```
gnecula, line 144: nit: I would just inline here the function here. I am not really sure though, for the cleanup you will need to define a function.

SIben: Addressed in `6fc28be`. I actually put everything out of the loop now, since we can use `addCleanup`.

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: Addressed in `d069c18`

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: Addressed in `697c06d`

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: Done in `7da907f`, but the alternative looks a little bit hacky. OK with it if you are and the cost of importing TF justifies it (I envision that this function is probably the only reason we'd need TF there)

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -0,0 +1,136 @@
...
+
+from jax import core
+from jax import dtypes
+from jax import lax
+from jax import numpy as jnp
+
+def to_jax_dtype(tf_dtype):
+  np_dtype = tf_dtype.as_numpy_dtype
+  # All the TF numpy dtypes come from numpy except for bool, bfloat16.
+  if np_dtype.__module__ == np.__name__:
```
gnecula, line 27: Wouldn't it be clearer to look up the name of the type:
```
if tf_dtype.name == "bfloat16":
   return dtypes.bfloat16
elif ...
```

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -0,0 +1,136 @@
...
+    return np_dtype
+  # Only bool has less than 5 elements; in this case, we can still return
+  # np_dtype.
+  elif np_dtype(4) != 4:
```
gnecula, line 31: I think this is a very hacky, that is why I thought us just look at the name of the type.

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -0,0 +1,136 @@
...
+    return np_dtype
+  return jnp.bfloat16
+
+Limitation = NamedTuple("Limitation", [ ("PrimitiveName", str)
```
gnecula, line 35: I did not notice this before, but attribute names normally should be lowercase, only type names are capitalized.

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: You're… 100% right, I don't know why I didn't do that in the first place. Thank you, fixed in `806f909`.

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: You're… 100% right, I don't know why I didn't do that in the first place. Thank you, fixed in `806f909`.

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: Thanks, addressed in `09df6e9`.

PR #4204, opened 2020-09-04
[jax2tf] Use the JAX impl rule for threefry instead of writing our own

gnecula:
The previous implementation copied the threefry implementation; instead we just interpret and convert on the fly
the existing JAX impl rule. However, we need to cast uint32 to int32 because tf.add is not implemented for uint32.

### DIFF
--- a/jax/experimental/jax2tf/tests/primitives_test.py
+++ b/jax/experimental/jax2tf/tests/primitives_test.py
@@ -612,23 +612,14 @@ class JaxPrimitiveTest(tf_test_util.JaxToTfTestCase):
     values = np.array([True, False, True], dtype=np.bool_)
     self.ConvertAndCompare(f_jax, values)
 
-  def test_random_gamma(self):
-    f_jax = jax.jit(jax.random.gamma)
-    for alpha in [1.0,
-                  np.array([1.0, 0.2, 1.2], np.float32),
-                  np.array([1.0, 0.2, 1.2], np.float64)]:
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -612,23 +612,14 @@ def test_reduce_ops_with_boolean_input(self, f_jax):
...
-                    np.array([0, 0xFFFFFFFF], dtype=np.uint32),
-                    np.array([0xFFFFFFFF, 0xFFFFFFFF], dtype=np.uint32)
-                    ]:
-      self.ConvertAndCompare(f_jax, rng_key)
+  @primitive_harness.parameterized(primitive_harness.random_gamma)
+  def test_random_gamma(self, harness: primitive_harness.Harness):
+    self.ConvertAndCompare(harness.dyn_fun, *harness.dyn_args_maker(self.rng()),
+                           rtol=1e-5)
+
+  @primitive_harness.parameterized(primitive_harness.random_gamma)
```
SIben, line 620: Shouldn't this be `@primitive_harness.parameterized(primitive_harness.random_split)`?

bchetioui (changes requested):
LGTM apart from the wrongly named test case.

jax/experimental/jax2tf/tests/primitives_test.py
gnecula: Indeed, good catch!

tomhennigan:
Neat!

PR #4205, opened 2020-09-04
fix docs build by adding requirement

### LINKED ISSUE #4202: Broken docs check
The `docs/readthedocs.org:jax` check fails across PRs lately, i. e. [here](https://readthedocs.org/projects/jax/builds/11796506/). The only error in the [log](https://readthedocs.org/api/v2/build/11796506.txt) is 

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

myst-parser 0.12.8 requires docutils>=0.15, but you'll have docutils 0.14 which is incompatible.
```

froystig:
This is an attempt to resolve #4202

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -9,6 +9,7 @@ sphinx-autodoc-typehints <1.11
 jaxlib
 ipykernel
 nbsphinx
+docutils >=0.15
 myst-parser[sphinx]
 # The next packages are for notebooks
 matplotlib


gnecula:
Should this PR be merged?

froystig:
It is no longer needed!

PR #4209, opened 2020-09-07
[jax2tf] Expand coverage of primitives by categorize.

bchetioui:
This commit adds handling logic for the limitations of:
- qr
- svd
- select_and_gather_add
- reduce_window/reduce_window_{min,max,sum}
- add
- mul
- scatter/scatter_{min,max,mul,add}

Also fixes a bug in a call to _infer_shape_jax, which wasn't
compatible with boolean operands and went undetected due to the
high-level handling of TF exceptions in higher-order primitives.

`expect_tf_exception` is now only explicitly set in tests where the issue
is numerical differences, and for `population_count` where it requires
implementing tests for `bitcast`, which will be done in a separate PR.

### DIFF
--- a/jax/experimental/jax2tf/tests/correctness_stats.py
+++ b/jax/experimental/jax2tf/tests/correctness_stats.py
@@ -73,7 +73,7 @@ def categorize(prim: core.Primitive, *args, **kwargs) \
     np_dtype = _to_np_dtype(args[0].dtype)
     if np_dtype in [np.float16, dtypes.bfloat16]:
       # b/158006398: TF kernels are missing for 'rem' and 'atan2'
-      tf_unimpl(f"Missing TF kernels for {prim.name} with dtype {np_dtype}")
+      tf_unimpl(f"{prim.name} is unimplemented for dtype {np_dtype}")
 
   if prim is lax.nextafter_p:
     np_dtype = _to_np_dtype(args[0].dtype)
--- a/jax/experimental/jax2tf/tests/correctness_stats.py
+++ b/jax/experimental/jax2tf/tests/correctness_stats.py
@@ -14,7 +14,7 @@
 
 import functools
 import numpy as np
-from typing import Callable, List, NamedTuple, Optional, Tuple, Sequence
+from typing import Any, Callable, List, NamedTuple, Optional, Tuple, Sequence
 
 from jax import core
 from jax import dtypes
@@ -31,6 +31,8 @@ Limitation = NamedTuple("Limitation", [ ("primitive_name", str)
                                       , ("error_string", str)
                                       , ("devices", Tuple[str,...])
                                       ])
+
+NpDType = Any
 
 def categorize(prim: core.Primitive, *args, **kwargs) \
     -> List[Limitation]:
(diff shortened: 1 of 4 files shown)


jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -69,16 +70,59 @@ def _to_np_dtype(dtype):
       tf_unimpl(f"{prim.name} is unimplemented for dtype {np_dtype}")
 
   if prim in [lax.rem_p, lax.atan2_p]:
-    # b/158006398: TF kernels are missing for 'rem' and 'atan2'
     np_dtype = _to_np_dtype(args[0].dtype)
     if np_dtype in [np.float16, dtypes.bfloat16]:
+      # b/158006398: TF kernels are missing for 'rem' and 'atan2'
       tf_unimpl(f"Missing TF kernels for {prim.name} with dtype {np_dtype}")
```
gnecula, line 76: I think we should change the text to "{prim.name} is unimplemented ...", like the others

SIben: Done in `99a745e`

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -69,16 +70,59 @@ def _to_np_dtype(dtype):
       tf_unimpl(f"{prim.name} is unimplemented for dtype {np_dtype}")
```
gnecula, line 70: It seems that it may be useful to change `tf_unimpl` to take just the `np_dtype` as an argument and keep the rest constant.

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -69,16 +70,59 @@ def _to_np_dtype(dtype):
...
+      tf_unimpl(f"{prim.name} is unimplemented for dtype {np_dtype}")
+
+  if prim is lax_linalg.svd_p:
+    np_dtype = _to_np_dtype(args[0].dtype)
+    if np_dtype in [np.float16, dtypes.bfloat16]:
+      # TODO: SVD on TPU for bfloat16 seems to work for JAX but fails for TF
+      tf_unimpl(f"{prim.name} is unimplemented for dtype {np_dtype}",
+                devs=["TPU"])
+    elif np_dtype in [np.complex64, np.complex128]:
+      # TODO: on CPU and GPU "No registered 'Svd' OpKernel for XLA_CPU_JIT
```
gnecula, line 97: I am wondering if the `Works on JAX because JAX uses a custom implementation` deserves to be in the generated documentation?

gnecula (commented):
This looks like a nice improvement

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: Added it in `c289a06`

jax/experimental/jax2tf/tests/correctness_stats.py
SIben: Addressed in `c289a06`

PR #4210, opened 2020-09-07
Add more context to the axis_frame error message.

apaszke:
Some of the vmap and gmap collective tests have been failing on master
and I can't seem to be able to reproduce them locally. Hopefully, if
this happens again, this extra bit of information will be useful in
debugging the problem.


gnecula:
This seems also a useful improvement to the error message.

gnecula (approved):

PR #4211, opened 2020-09-07
[jax2tf] Do not use jax.random.PRNGKey before in primitive harness

gnecula:
We cannot execute JAX functions before the program is initialized


bchetioui (approved):
LGTM

PR #4219, opened 2020-09-08
[jax2tf] Add a template file for documentation generation.

bchetioui:
The documentation now gives instructions about how to
regenerate it, as well as when it was last generated.


jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -29,10 +29,13 @@
 
 import os
 
-if os.getenv('JAX2TF_CATEGORIZE_OUT'):
+if not os.getenv('JAX2TF_CATEGORIZE_OUT') is None:
```
gnecula, line 32: I normally see this written as "x is not None"; feels more readable, wdyt?

jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -1,5 +1,15 @@
...
+*Last generated on (YYYY-MM-DD): 2020-09-08*
+
+## Updating the documentation
+
+This file is automatically generated by running the jax2tf tests in
+`jax/experimental/jax2tf/tests/primitives_test.py` with the
+`JAX2TF_CATEGORIZE_OUT` environment variable set.
+
+## Generated summary of primitives with limited support
+
```
gnecula, line 12: I suggest adding some text here: 

The following JAX primitives are converted to Tensorflow but the result of the conversion may trigger runtime errors when run on certain devices and with certain data types.

jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -29,3 +39,29 @@
 | scatter-mul | Missing TF support | scatter-mul is unimplemented for dtype complex64 | TPU |
 | select_and_gather_add | Missing TF support | select_and_gather_add is unimplemented for dtype float32 | TPU |
 | svd | Missing TF support | svd is unimplemented for dtype complex64; this works on JAX because JAX uses a custom implementation | CPU, GPU |
+
+## Not yet implemented primitive conversions
+
```
gnecula, line 44: I would add some text here as well:

The conversion of the following JAX primitives is not yet implemented. 

Nit: I would either put them all in one line, or put them in a table with similar format as the primitives above.

jax/experimental/jax2tf/tests/tf_test_util.py
SIben: Changed this in `bd4528a2`!

jax/experimental/jax2tf/primitives_with_limited_support.md
SIben: Added & put all the primitives on one line.

PR #4222, opened 2020-09-08
[jax2tf] Add tests for the conversion of conv_general_dilated

bchetioui:
This also adds the precision argument to the tfxla call which
was previously ignored.

### DIFF
--- a/jax/experimental/jax2tf/tests/primitive_harness.py
+++ b/jax/experimental/jax2tf/tests/primitive_harness.py
@@ -876,98 +876,76 @@ random_split = tuple(
                                np.array([0xFFFFFFFF, 0xFFFFFFFF], dtype=np.uint32)])
 )
 
-lax_conv_general_dilated_general = tuple(
-  Harness(f"_general_lhs={jtu.format_shape_dtype_string(lhs_shape, dtype)}_rhs={jtu.format_shape_dtype_string(rhs_shape, dtype)}_windowstrides={window_strides}_padding={padding}_lhsdilation={lhs_dilation}_rhsdilation={rhs_dilation}_dimensionnumbers={dimension_numbers}_featuregroupcount={feature_group_count}_batchgroupcount={batch_group_count}_precision={precision}".replace(' ', ''),
-    lax.conv_general_dilated,
-    [RandArg(lhs_shape, dtype), RandArg(rhs_shape, dtype),
-     StaticArg(window_strides), StaticArg(padding), StaticArg(lhs_dilation),
-     StaticArg(rhs_dilation), StaticArg(dimension_numbers),
-     StaticArg(feature_group_count), StaticArg(batch_group_count),
-     StaticArg(precision)],
-    lhs_shape=lhs_shape,
-    rhs_shape=rhs_shape,
-    dtype=dtype,
-    window_strides=window_strides,
-    padding=padding,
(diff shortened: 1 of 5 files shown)


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -875,3 +875,99 @@ def _fft_rng_factory(dtype):
                                np.array([0, 0xFFFFFFFF], dtype=np.uint32),
                                np.array([0xFFFFFFFF, 0xFFFFFFFF], dtype=np.uint32)])
 )
+
+lax_conv_general_dilated_general = tuple(
```
gnecula, line 879: How long do these tests take to run? Is there a way to reduce the number of tests by thinking which of the features are orthogonal?

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -875,3 +875,99 @@ def _fft_rng_factory(dtype):
...
+      ((1, 1), (2, 3)), # dilation only on RHS (atrous)
+      ((2, 3), (3, 2))  # dilation on both LHS and RHS (transposed & atrous)
+  ]
+  # Dimension numbers and corresponding permutation
+  for dimension_numbers, perms in [
+      (("NCHW", "OIHW", "NCHW"), ((0, 1, 2, 3), (0, 1, 2, 3))), # default
+      (("NHWC", "HWIO", "NHWC"), ((0, 2, 3, 1), (2, 3, 1, 0))), # TF default
+      (("NCHW", "HWIO", "NHWC"), ((0, 1, 2, 3), (2, 3, 1, 0))), # custom
+  ]
+  for lhs_shape, rhs_shape in [[
```
gnecula, line 931: I am wondering if this would be easier to write if it is merged with the `for` above.

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -455,6 +455,20 @@ def test_dynamic_update_slice(self, harness):
   def test_squeeze(self, harness: primitive_harness.Harness):
     self.ConvertAndCompare(harness.dyn_fun, *harness.dyn_args_maker(self.rng()))
 
+  @primitive_harness.parameterized(jtu.cases_from_list(
+      primitive_harness.lax_conv_general_dilated_general
+    ) + list(primitive_harness.lax_conv_general_dilated_known_limitations))
```
SIben, line 460: @gnecula this is my idea for reducing the length of the tests; we only take a subset of this general tests, and make sure to run the cases with known limitations every time.

gnecula: I am thinking that we can avoid an exponential number of cases if we think of some features as orthogonal. For example, let's say that the precision feature is orthogonal to the others. Then instead of generating all the tests for all precisions, we generate a set of base tests and then every 3rd one get the HIGHEST precision. This reduces the tests by a factor of 3. Perhaps we can do that for a few more features. 

If you think in terms of code coverage, we replace path coverage for a sequence of conditionals with branch coverage: each branch is covered but not necessarily all combinations are covered.

SIben: Oh I see! That may be indeed possible – and more desirable than taking a random subset of tests everytime, I think. I'll try to break it down like that.

Issue #3271, opened 2020-06-01
Leaking trace stack global state in presence of exceptions


gnecula:
I ran into a situation when the core.trace_state was left in a bad state after a failing test. Here is a simple repro:

```
  def test_leaking_trace_stack(self):
    assert not core.trace_state.trace_stack.upward   # Assert clean trace_state
    def func(x):
      raise ValueError(x)
    try:
      jax.jvp(func, (0.,), (0.,))
    except:
      print(core.trace_state.trace_stack.upward)  # We should have a clean trace_state
```

The above test will print a non-empty trace_state (with a JVPTrace). 

I think that this comes from the following use of generators within context managers. 

```
@lu.transformation
def jvpfun(instantiate, primals, tangents):
  with new_master(JVPTrace) as master:
    out_primals, out_tangents = yield (master, primals, tangents), {}
    del master
  ...
```

Since we yield from the body context manager, if an exception is raised the context manager will not be on the stack and it won't be cleaned up by the context manager. Instead, the cleanup will run later when the context manager is garbage collected. By then, the next test may have started already and will use the wrong trace_state.

Perhaps we need to review the linear_util to catch exceptions and pass them to the generators?

mattjj:
I think `new_master` is implemented in terms of `try/finally` to avoid the issue you described. I'll take a look at that test case and see what's up...

mattjj:
I think it's getting cleaned up, just in a confusing order. If we put a print statement like `print(trace_state.trace_tsack.upward)` after [this line](https://github.com/google/jax/blob/dd81a8dded3191110a2db884581c1d4dacd364fb/jax/core.py#L647) then we get this printed:

```
[MasterTrace(0,JVPTrace)]
[]
```

In other words, this version of the test in the OP passes:

```python
import jax
from jax import core

assert not core.trace_state.trace_stack.upward   # Assert clean trace_state
def func(x):
  raise ValueError
try:
  jax.jvp(func, (0.,), (0.,))
except:
  pass
print(core.trace_state.trace_stack.upward)  # We should have a clean trace_state
```

```
[]
```

I'm still puzzling over the logic over the order of evaluation, but since it's getting cleaned up eventually, do we care about the order?

mattjj:
I put the assignment back to @gnecula because AIUI he's working on another repro.

gnecula:
Fixed by #4226

PR #4226, opened 2020-09-08
Interrupt lu transformation generators whenever an exception occurs

apaszke:
This fixes some errors that have been appearing in our CI from time to
time. All transformations are implemented as generators, but they
haven't been explicitly aborted when an exception has been raised.
Instead, they only got closed when they got garbage collected, which
could happen at an unspecified later time, potentially leading to a
corruption of global state, which could have been modified after the
exception was handled.

Note that this implementation doesn't propagate the original exception
into the argument transformations, and doesn't allow them to handle the
error either. Such an extension would be possible, but throwing an
exception into a generator mutates the exception object, clobbering
the nice traceback that we would usually carry. One can work around
those issues, but it feels really hacky and we don't need it right now
anyway, so I figured we'll be better off with the simple thing for the
time being.

### DIFF
--- a/jax/linear_util.py
+++ b/jax/linear_util.py
@@ -147,7 +147,17 @@ class WrappedFun(object):
       stack.append((gen, out_store))
     gen = None
 
-    ans = self.f(*args, **dict(self.params, **kwargs))
+    try:
+      ans = self.f(*args, **dict(self.params, **kwargs))
+    except:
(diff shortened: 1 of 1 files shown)


apaszke:
Fixes #3271 (I think?)

froystig (approved):

PR #4227, opened 2020-09-08
Extend axis env while translating the pmapped jaxpr to XLA

apaszke:
This is normally unnecessary, because the XLA translation usually
doesn't bind any of the primitives in the jaxpr, but this is not true in
case of scan! Its translation rule reevaluates the jaxpr as a function,
and if it contains collectives such as `axis_index` it can fail due to
axis being missing.


mattjj (approved):
Nice!

PR #4229, opened 2020-09-08
Expose jax.image.scale_and_translate

johnpjf:
Expose scale_and_translate as a public function, fix a bug in implementation when translation is not 0.

### DIFF
--- a/docs/jax.image.rst
+++ b/docs/jax.image.rst
@@ -13,4 +13,5 @@ Image manipulation functions
   :toctree: _autosummary
 
     resize
+    scale_and_translate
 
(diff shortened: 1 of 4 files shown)


hawkinsp (changes requested):
You should add the new function to the documentation and verifies that it renders correctly (either build the docs yourself, or look at the readthedocs preview you get as a presubmit). To add the new function to the docs, you add it to `jax.image.rst`.

johnpjf:
Ok, added to docs/jax.image.rst.

hawkinsp (approved):

PR #4230, opened 2020-09-08
Add a test that fails due to unsupported strides.

zhangqiaorjc:
- test fails during dlpack tensor to buffer conversion
- XLA layout does not support arbitrary dlpack strides
- users should explicit materialze such tensors by making a copy

### DIFF
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -5,3 +5,4 @@ pillow
 pytest-benchmark
 pytest-xdist
 wheel
+torch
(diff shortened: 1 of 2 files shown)


zhangqiaorjc:
I'm going to do a binary release for jaxlib which has the actual fix.

mattjj (approved):
LGTM!

hawkinsp:
Can we merge this PR?

zhangqiaorjc:
I totally lost track of this PR. Should we still merge it?

hawkinsp:
I think so. The `jax.lib.version` guard is now stale, though.

zhangqiaorjc:
Isn't the version guard still correct? or you meant we can safely remove it now?

hawkinsp:
Isn't the minimum version now higher than the guard?

zhangqiaorjc:
this PR is superceded by #6013

PR #4232, opened 2020-09-09
Fix Dockerfile wheel installation issues. #4066

zhangqiaorjc:
Needs to add "packaging" after

https://github.com/google/jax/pull/4066


mattjj (approved):
LGTM

PR #4234, opened 2020-09-09
improve lax.pad shape rule

mattjj:
It's now:
  * better tested
  * better at catching errors
  * faster
  * easier to read


mattjj:
Internal tests pass!

PR #4236, opened 2020-09-09
[jax2tf] Clean up test_dynamic_slice.

bchetioui:
With the XLA nested compilation bug fixed, this should now work
fine.


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1096,10 +1096,8 @@ def _dynamic_slice(operand, *start_indices, slice_sizes):
   # tf.gather. But those have different semantics for index-out-of-bounds than
   # JAX (and XLA). We have tried to force compilation, by wrapping into
   # tf.xla.experimental.compile, or tf.function(experimental_compile=True), but
-  # those solutions are brittle because they do not work when nested into an
```
gnecula, line 1099: Neither of those bugs are fixed. So, this text should stay as it was.

SIben: Oops – I thought these were the ones concerning the nested compilation. Caveat of not having access to b/ links :)
Fixed!

gnecula (approved):

PR #4237, opened 2020-09-09
[jax2tf] Moved the limitations for XlaSort to correctness_stats

gnecula:


jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -29,7 +29,7 @@
 
 import os
 
-if os.getenv('JAX2TF_CATEGORIZE_OUT'):
+if os.getenv('JAX2TF_OUTPUT_LIMITATIONS'):
```
SIben, line 32: If we're settling on that variable name, maybe it is better to leave it out of this PR and make the change in #4219 so that it remains consistent across code/template file/doc?

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -56,9 +56,12 @@ def _report_failure(error_type: str, msg: str,
...
+  def tf_unimpl(np_dtype: Optional[NpDType] = None,
+                additional_msg: Optional[str] = None,
                 devs: Sequence[str] = all_devices) -> None:
-    msg = f"{prim.name} is unimplemented for dtype {np_dtype}"
+    msg = f"{prim.name} is unimplemented"
+    if np_dtype is not None:
+      msg += f" for dtype {np_dtype}"
     if additional_msg:
       msg += '; ' + additional_msg
     _report_failure("Missing TF support", msg, devs=devs)
```
SIben, line 67: This change is in concurrence with a change I'm proposing in #4222. Which one do you think we should go for? A mix of the 2 is also possible :)

gnecula: If that PR is ready, we can merge it and I adapt. Otherwise, we merge this and you can adapt. I think that we will iterate on this quite a bit as we expand the categorization. Most likely there will be several helper functions, as in your PR.

SIben: It is not ready. Let's merge this one, and I will adapt.

jax/experimental/jax2tf/tests/tf_test_util.py
gnecula: I have merged #4219. I will update the template here, if you are fine with the new flag.

SIben: Since #4219 was just merged, please update template & doc as well here :)

SIben: Yes, that is fine by me!

PR #3983, opened 2020-08-06
Allow JAX objects to be represented by multiple buffers

jakevdp:
Why? This will allow us to support more flexible extensions of JAX types, such as sparse matrices (which are single objects wrapping multiple buffers).

SInce this change touches large parts of the package, I plan to approach it incrementally, changing a bit at a time and ensuring tests pass:

- [x] make `xla.device_put()` return a length-1 tuple of buffers
- [x] make `xla.aval_to_xla_shape` return length-1 tuple of xla_shapes
- [x] partition flattened buffers properly for xla result handlers
- [x] generalize above to multiple buffers & add a test of a basic jax object backed by two buffers
- [ ] ~modify `unit` to return an empty tuple instead of a place-holder~ ...may be possible, but is not trivial. For example, there are multiple flavors of object that use `AbstractUnit` as their `aval`, including `Unit`, `UnitVar`, and `DropVar`. The latter in particular cannot be lowered to zero buffers, becuase it is used to ingest (and then ignore) a buffer produced by a primitive that returns multiple buffers, when that buffer happens to be unused in the jaxpr. I tried working around this by creating a custom abstract value for `DropVar`, but there remained some failures related to `UnitVar` for which it was difficult to find the root cause.
- [ ] ~possible to remove `multiple_results` from primitives?~ ...probably not; we still want some primitives to return values and some to return tuples, though we need to be careful with flattening when there are multiple results that have multiple buffers (covered by lax loop test case)

PR #4238, opened 2020-09-09
Add new public API: pxla.device_put()

jakevdp:
Why? We are changing the behavior of `device_put()` in #3983; this causes issues for downstream libraries that use internal APIs. An example is the trax project [here](https://github.com/google/trax/blob/67a4edf83fb77cd1a9df872005130cd2c04d1854/trax/layers/acceleration.py#L260-L264).

This PR introduces a new API to cover this, `pxla.device_put()`, so that we can change internal APIs without breaking downstream users.

Once this change is in, trax can use this new API and not be broken by #3983.

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -237,7 +237,7 @@ def shard_args(devices: Sequence[xb.xla_client.Device],
 
 shard_arg_handlers: Dict[Any, Callable[[Any, Any, Any], Sequence[Any]]] = {}
 shard_arg_handlers[core.Unit] = \
-    lambda x, devices, _: [xla.device_put(core.unit, d) for d in devices]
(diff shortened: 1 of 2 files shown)


mattjj:
Hmm, nothing in pxla.py (or xla.py) should be public; we'd need to define a public function in api.py.

Moreover, xla.device_put (and now pxla.device_put) returns a Buffer (or a list of Buffers), and we don't want to return those to users. So long as trax needs to work with Buffers directly, it's using internal APIs. That is, I suspect we need to make different revisions to trax so that it's no longer using internal APIs.

jakevdp:
Fair enough. Perhaps we should provide `jax.api.multi_device_put` that more or less does what the trax function does?

jakevdp:
Looks like this will be accomplished by @skye's work on a `replicate` function.

PR #4240, opened 2020-09-09
Update install_cuda script to specify cublas.

zhangqiaorjc:
CUBLAS packaging changed in CUDA 10.1 according to

https://forums.developer.nvidia.com/t/cublas-for-10-1-is-missing/71015

so the dependency has to be specified explicitly. I followed examples from TF Dockerfile

https://github.com/tensorflow/tensorflow/blob/3182f482fd7cc19fe5f75bf502f148c8583e3c18/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu.Dockerfile#L45

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu.Dockerfile

### DIFF
--- a/build/install_cuda.sh
+++ b/build/install_cuda.sh
@@ -4,22 +4,39 @@ set -xe
 CUDA_VERSION=$1
 
 LIBCUDNN=libcudnn7
-if [ $CUDA_VERSION = "10.2" ]; then
+if [ $CUDA_VERSION = "10.0" ]; then
+  CUBLAS=libcublas10
+  CUBLAS_DEV=libcublas-dev
+elif [ $CUDA_VERSION = "10.1" ]; then
+  # Have to pin to libcublas10=10.2.1.243-1 due to bug in TF, see
+  # https://github.com/tensorflow/tensorflow/issues/9489#issuecomment-562394257
+  CUBLAS=libcublas10=10.2.1.243-1
+  CUBLAS_DEV=libcublas-dev=10.2.1.243-1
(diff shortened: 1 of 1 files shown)


build/install_cuda.sh
```diff
@@ -4,22 +4,39 @@ set -xe
 CUDA_VERSION=$1
 
 LIBCUDNN=libcudnn7
-if [ $CUDA_VERSION = "10.2" ]; then
+if [ $CUDA_VERSION = "10.0" ]; then
+  CUBLAS=libcublas10
+  CUBLAS_DEV=libcublas-dev
+elif [ $CUDA_VERSION = "10.1" ]; then
+  # Have to pin to libcublas10=10.2.1.243-1 due to bug in TF, see
```
hawkinsp, line 11: I suspect this isn't true: we're mostly using the cublas installation in the docker container just for its headers, not for its library. We will dlopen() the library at run time I think.

That said, a pin does no harm.

hawkinsp (approved):

build/install_cuda.sh
zhangqiaorjc: SG, let's keep it in sync with TF Dockerfile for now, which pins to this version

https://github.com/tensorflow/tensorflow/blob/3182f482fd7cc19fe5f75bf502f148c8583e3c18/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu.Dockerfile#L44

PR #4244, opened 2020-09-10
Add comments for residuals from f_bwd.

zhangqiaorjc:
In my first reading of this notebook, i found it hard to understand what the 2nd return value of f_fwd is, and had to dig through API docstring for f_fwd. It might be easier to use comments to point out that this residual is immediately used by f_bwd.


gnecula (approved):
Thanks for contributing improvements to parts that are unclear for a JAX beginner.

PR #4246, opened 2020-09-10
[jax2tf] Cleanup test_unary_elementwise.

bchetioui:


jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -16,10 +16,25 @@ certain data types.
 
 | Affected primitive | Type of limitation | Description | Devices affected |
 | --- | --- | --- | --- |
+| acosh | Missing TF support | acosh is unimplemented for dtype bfloat16 | CPU, GPU |
```
gnecula, line 19: Maybe to make a change in a future PR: the text should not repeat the name of the primitive, I think. Perhaps something like "Primitive is unimplemented in TF for: bfloat16, int32, ...".

gnecula (approved):

jax/experimental/jax2tf/primitives_with_limited_support.md
SIben: Addressed directly when rebasing, since I had to regenerated the file to address the conflict 👍

PR #4248, opened 2020-09-10
[jax2tf] Add converted primitives without tests to the generated doc.

bchetioui:


jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -75,3 +75,10 @@ conversion to Tensorflow.
 The conversion of the following JAX primitives is not yet implemented:
 
 `after_all`, `all_to_all`, `axis_index`, `cholesky`, `create_token`, `cummax`, `cummin`, `custom_linear_solve`, `eig`, `eigh`, `igamma_grad_a`, `infeed`, `lu`, `outfeed`, `pmax`, `pmin`, `ppermute`, `psum`, `random_gamma_grad`, `reduce`, `rng_uniform`, `triangular_solve`, `xla_pmap`
+
+## Primitive conversions with missing tests
+
+The following JAX primitives have a defined conversion but are known to be
+missing tests:
+
+`add_any`, `argmax`, `argmin`, `broadcast`, `call`, `clamp`, `complex`, `cond`, `conj`, `conv_general_dilated`, `custom_jvp_call_jaxpr`, `custom_lin`, `custom_vjp_call_jaxpr`, `dot_general`, `imag`, `integer_pow`, `real`, `remat_call`, `rev`, `scan`, `select_and_scatter`, `select_and_scatter_add`, `stop_gradient`, `transpose`, `xla_call`
```
gnecula, line 84: xla_call, remat_call, call: these are not really seen by the converter, they are inlined. You should exclude them. Technically xla_pmap_p also. Search for xla_call_p in jax2tf.

scan, cond, while (why is it not on the list?): these are tested in control_flow_ops_test. 

I am surprised about select_and_scatter_add.

SIben: `while` is called within the `random_gamma` conversion it seems.

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -200,19 +205,35 @@ def _pipewrap(columns):
...
-                       template_file: str) -> None:
+prettify_not_yet_implemented = lambda: prettify_as_ordered_list(tf_not_yet_impl)
+
+def prettify_not_yet_covered(covered_set: Set[core.Primitive]) -> str:
+  """
+  Builds an ordered summary markdown list of all the primitives that are
+  implemented but not in the set passed as an argument.
+  """
+  ignore = set([xla.xla_call_p, pxla.xla_pmap_p, pe.remat_call_p, core.call_p,
+                lax.cond_p, lax.scan_p, lax.while_p])
```
gnecula, line 223: The "custom_xxx_call_jaxpr" are also well-tested, in the control-flow-ops-test. Perhaps remove them from here. Alternatively (may be even better) run both primitives_tests and control-flows tests when generating the documentation?

SIben: Addressed in `7725d8db`; we now also run tests in `control_flow_ops_test` and this instruction is added to the template.

PR #4250, opened 2020-09-10
Clarify when jaxlib version should be bumped.

skye:


skye:
@jblespiau @hawkinsp please chime in if you don't think this is the right policy

jblespiau:
The versioning is a side effect of building the C++ and Python code in 2 different packages. At a given point in time, the Python jax is expected to work with the latest released jaxlib and that's is: there is no stronger guarantee (e.g. no table of supported versions).
Thus, I think that many policies will work, and this one is a simple one, so LGTM.

skye:
Just a minor clarification: there's also the [minimum required jaxlib version](https://github.com/google/jax/blob/master/jax/lib/__init__.py#L25), so we do guarantee more than just the latest jaxlib working. But doesn't change your point that other policies would be possible.

PR #4262, opened 2020-09-11
[jax2tf] Add input and output names to the TensorFlow graph generated by jax2tf.

petebu:


gnecula:
Can you please give some background for this change? Is this to help with profiling or for some other graph inspection? 

Is there a standard practice for doing something like this in TF? I am asking because depending on how this is used, it has the potential to become part of the requirements for jax2tf.convert, and I'd like to understand those requirements.

tomhennigan:
Peter is OOO this week but I can provide some more context.

The goal is to allow graph inspection and manipulation (e.g. prune the resulting TF graph between input and output tensors).

Currently we are using `tf.function` to extract a graph, and this provides names for input tensors (in the form `arg_{idx}_{n}` iirc) but not outputs (although I suspect if we look hard enough we can find output tensor names).

Peter and I wanted to ensure that the name matching was indeed a supported part of the public API that we were using and it was not clear if this was a supported part of `tf.function`s contract (although I suspect it is). It seemed "cleaner" to do this in `jax2tf` regardless.

We did discuss whether we could simply do this "around" jax2tf.convert (e.g. in another library). This would be possible, but is slightly complicated by the custom gradient registered by jax2tf (which we cannot decorate/wrap). We currently do not need the gradient for our use case so this is not a blocker, but we felt that on balance it was a useful feature for jax2tf users so wanted to upstream.

No worries if this is not a part of the public API you want to commit to, we can use a similar approach in a wrapper function to name function inputs/outputs and then solve "the gradient problem" in the future if we need to.

gnecula:
Thanks for the explanation. 

Since the use case seems to be something that would apply to a tf.function even if not coming from jax2tf, I feel that it would be best if we can use existing TF mechanisms for this, with the vision that jax2tf produces something that behaves as if it was written with TF, and can be processed with existing TF mechanisms. 

If it is more expedient to make this change, let's do that now, but keep this behavior undocumented, hoping that we can remove it, or replace it with an expanded naming metadata mechanism (which we are thinking of doing anyway, for visibility in xprof and such). 

I will wait for your answer, and then I am ready to merge this change.

gnecula:
@petebu I think that now is a good time to merge this. Servomatic will also use this mechanism to do some static analysis for the jax2tf models.

PR #4263, opened 2020-09-11
[jax2tf] Fix lax.div_p

petebu:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -501,7 +501,7 @@ def _population_count(x):
 
 def _div(lhs, rhs):
   if lhs.dtype.is_integer:
-    quotient = tf.math.floor_divide(lhs, rhs)
+    quotient = tf.math.floordiv(lhs, rhs)
```
gnecula, line 504: This suggests that we have no tests for _div for integer types?

@SIben was looking at expanding test coverage. FYI.

gnecula (approved):

jax/experimental/jax2tf/jax2tf.py
SIben: Yes, it is somehow the same problem that we had for `add` and `mul` – it is only included in `test_binary_elementwise` afaict, which only considers floating point types. Perhaps it'd be good to gather the untested paths by dtype to get a better overview.

gnecula:
But aren't these tested in `test_type_promotion` for integer types? I suspect that the problem there is that the right combination that would trigger the error is not picked up.

SIben:
In `test_type_promotion` we go through `jnp.divide` which ends up promoting the args to inexact types before they encounter the primitive.

PR #4264, opened 2020-09-11
Support non-fragmenting mask of reshape

juliuskunze:
Supports `mask(reshape)` for all non-fragmenting cases. 

Restrictions:
- Polymorphic (non-constant) dimensions have to stay in the same order (no swapping/splitting/merging allowed) and
- the product of each polymorphic size and all constant sizes before the next polymorphic dimension has to be unchanged.

For example, reshapes between the following shapes are allowed:
- `(n, 2)` <=> `2n`
- `(n, 2, 4, m)` <=> `(4n, 2, m)`

Examples of reshapes not allowed:
- `(2, n)` <≠> `2n`
- `(a, b)` <≠> `ab`
- `(a, b)` <≠> `(b, a)`

These would require `gather`/`scatter` operations. As we discussed in Jamie's talk, it is to be seen whether the resulting performance would be acceptable. This PR only implements parts that should be supported in any case.

### DIFF
--- a/jax/interpreters/masking.py
+++ b/jax/interpreters/masking.py
@@ -403,6 +403,8 @@ class MaskTrace(Trace):
     out_aval = primitive.abstract_eval(*(t.aval for t in tracers), **params)
     vals, polymorphic_shapes = unzip2((t.val, t.polymorphic_shape) for t in tracers)
     logical_shapes = map(shape_as_value, polymorphic_shapes)
+    # TODO(mattjj): generalize mask rule signature
+    if primitive.name == 'reshape': params['polymorphic_shapes'] = polymorphic_shapes
     out = masking_rule(vals, logical_shapes, **params)
     if primitive.multiple_results:
       return map(partial(MaskTracer, self), out, (o.shape for o in out_aval))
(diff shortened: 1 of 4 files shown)


jax/interpreters/masking.py
```diff
@@ -403,6 +403,7 @@ def process_primitive(self, primitive, tracers, params):
     out_aval = primitive.abstract_eval(*(t.aval for t in tracers), **params)
     vals, polymorphic_shapes = unzip2((t.val, t.polymorphic_shape) for t in tracers)
     logical_shapes = map(shape_as_value, polymorphic_shapes)
+    if primitive.name == 'reshape': params['polymorphic_shapes'] = polymorphic_shapes
```
mattjj, line 406: We should probably avoid this kind of pattern. I think the general form of a masking rule should be to pass in polymorphic shapes, and if the common case doesn't need it we can just have convenience wrappers to handle those.

I can make that refactor, but can you leave a "TODO(mattjj): generalize mask rule signature" above this line?

mattjj (approved):
Thanks for doing this!

I know I have a backlog of PRs from you, @JuliusKunze, (sorry!) but I figured at least I can offer prompt reviews for new small PRs like this one.

One suggestion for adding a TODO. Actually I played with this refactor last weekend already, so it's good to see another reason to pass in polymorphic_shapes in general.

mattjj:
The docs build failure is unrelated, and a pre-existing condition. We can ignore it.

JuliusKunze:
@mattjj Can we merge? Got another PR in the pipeline.

mattjj:
Unfortunately there are now conflicts... does it look easy for you to resolve them?

JuliusKunze:
@mattjj Sure, resolved!

JuliusKunze:
@mattjj merge?

mattjj:
Merged!

PR #4265, opened 2020-09-11
Use a TypeVar for vmap/pmap/jit function types, not just their return values.

aslanides:
This has the benefit of preserving both call and return signatures.

Note that apparently at least two types must be passed as constraints to TypeVar, so we have the slightly odd and redundant Callable, Callable[..., Any].

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -75,16 +75,14 @@ traceback_util.register_exclusion(__file__)
 
 AxisName = Any
 
-# This TypeVar is used below to express the fact that function call signatures
-# are invariant under the jit, vmap, and pmap transformations.
-# Specifically, we statically assert that the return type is invariant.
-# Until PEP-612 is implemented, we cannot express the same invariance for
-# function arguments.
-# Note that the return type annotations will generally not strictly hold
+# These TypeVars are used below to express the fact that function types
+# (i.e. call signatures) are invariant under the jit, vmap, and pmap
+# transformations.
+# Note that the function type annotations will generally not strictly hold
 # in JIT internals, as Tracer values are passed through the function.
 # Should this raise any type errors for the tracing code in future, we can disable
 # type checking in parts of the tracing code, or remove these annotations.
-T = TypeVar("T")
+F = TypeVar("F", bound=Callable)
 
 map = safe_map
 zip = safe_zip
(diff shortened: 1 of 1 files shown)


aslanides:
Hmm, seems like `mypy` has a problem with this:


```
$ mypy jax
> jax/image/scale.py:222: error: "object" not callable
> jax/random.py:433: error: "object" not callable
```

Both of these errors seem to originate at applications of jax.jit via functools.partial ... although they are far from the only instances in the codebase and I can't seem to see anything distinctive about them.

jax/api.py
```diff
@@ -68,16 +68,14 @@
...
-# Note that the return type annotations will generally not strictly hold
+# This TypeVar is used below to express the fact that function types
+# (i.e. call signatures) are invariant under the jit, vmap, and pmap
+# transformations.
+# Note that the function type annotations will generally not strictly hold
 # in JIT internals, as Tracer values are passed through the function.
 # Should this raise any type errors for the tracing code in future, we can disable
 # type checking in parts of the tracing code, or remove these annotations.
-T = TypeVar("T")
+F = TypeVar("F", Callable, Callable[..., Any])
```
shoyer, line 78: It looks like the [recommendation from mypy](https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators) is to use `bound`, e.g.,:
```suggestion
F = TypeVar("F", bound=Callable[..., Any])
```

I'm not entirely sure why that would be any different, but it might be worth a try...

aslanides:
Presubmit failure seems like a flaky doc test.

shoyer (approved):

aslanides:
Failing test(s) seem to be due to unrelated flakes -- I think we're safe to merge.

jekbradbury:
This change causes pytype to fail on certain downstream code patterns. For example, [this code](https://github.com/deepmind/dm-haiku/blob/master/examples/impala/learner.py#L40):
```python
mean_per_batch = jax.vmap(rlax.policy_gradient_loss, in_axes=1)(logits, *args)
```
fails to typecheck, with
```
File "[...]/haiku/examples/impala/learner.py", line 40, in policy_gradient_loss: Missing parameter 'adv_t' in call to function rlax.policy_gradient_loss [missing-parameter]
         Expected: (logits_t, a_t, adv_t, w_t)
  Actually passed: (logits_t, a_t)
```

This case might be a pytype bug, since `policy_gradient_loss` is in fact called with the four arguments it expects.

trevorcai:
Happy to fix Haiku in the above scenario (as it's a pretty straightforward improvement over my lazy code), but do want to register that `*args` is useful in some cases and it'd be a pity to lose this.

jekbradbury:
There are enough distinct failures in downstream code resulting from `*args` that I don't think we should merge this unless/until we can fix that (either in this PR, or in pytype).

aslanides:
Thanks -- I've raised a bug with the PyType team: b/174530851.

aslanides:
Update on this: the pytype team have now resolved the issue with unpacking *args, and I have fixed other breakages in downstream internal code. PTAL @jekbradbury :)

shoyer (approved):

PR #3882, opened 2020-07-28
Parameterize geev/eig to allow for not computing the left/right eigenvectors

bchetioui:
In the current JAX, calls to `eig`/`eigvals`/`geev` systematically compute both the left and right eigenvectors, whether or not they are needed. This PR adds the `jobvl` and `jobvr` parameters to allow for only computing the eigenvalues or only the desired eigenvectors, in both `jaxlib.lapack.?geev` and `jax.lax_linag.eig`.
`jobvl` and `jobvr` are by default set to True in `jaxlib.lapack.geev`, for backwards compatibility with JAX; they are however by default set to False in `jax.lax_linalg.eig`, as it seems more natural to explicitly request additional computations if they are needed.

PR #4266, opened 2020-09-11
Modifies eig_p and related operations to take advantage of the new jaxlib geev API

bchetioui:
This follows from #3882. @hawkinsp's comments on that PR have been addressed here.

### DIFF
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -1,5 +1,5 @@
 flake8
-jaxlib==0.1.55
+jaxlib==0.1.55 # must be kept in sync with the version in jax/lib/__init__.py
 mypy==0.770
 pillow
 pytest-benchmark
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -1,5 +1,6 @@
 flake8
-jaxlib==0.1.55 # must be kept in sync with the version in jax/lib/__init__.py
+ # Must be kept in sync with the minimum jaxlib version in jax/lib/__init__.py
+jaxlib==0.1.55
 mypy==0.770
 pillow
 pytest-benchmark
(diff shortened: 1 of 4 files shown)


build/test-requirements.txt
```diff
@@ -1,5 +1,5 @@
 flake8
-jaxlib==0.1.51
+jaxlib==0.1.55
```
skye, line 2: You should also bump the minimum jaxlib here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L25
(In fact, can you also add a comment to the line here saying to always keep these versions in sync?)

skye (changes requested):

build/test-requirements.txt
SIben: Done, thank you!

build/test-requirements.txt
```diff
@@ -1,5 +1,5 @@
 flake8
-jaxlib==0.1.51
+jaxlib==0.1.55 # must be kept in sync with the version in jax/lib/__init__.py
```
skye, line 2:
```suggestion
 # Must be kept in sync with the minimum jaxlib version in jax/lib/__init__.py
jaxlib==0.1.55
```

skye (approved):
One tiny nit, otherwise LGTM!

build/test-requirements.txt
```diff
@@ -1,5 +1,6 @@
 flake8
-jaxlib==0.1.51
+ # Must be kept in sync with the minimum jaxlib version in jax/lib/__init__.py
```
gnecula, line 2: Isn't the minimum version 1.51 still? What I would do is to bump the minimum version (and the version here) in a few days, unless we discover a problem with 1.55

gnecula: I see you've done it below. Ok.

PR #4038, opened 2020-08-12
omnistaging on by default

mattjj:
fixes #4114, fixes #3397, fixes #3108,

PR #3370, opened 2020-06-08
omnistaging

mattjj:
### Motivation

JAX transformations like `jit` and `pmap` stage out computations to XLA. That is, we apply them to functions comprising multiple `jax.numpy` operations so that rather being executed one at a time from Python the operations are all part of one end-to-end optimized XLA computation.

But exactly which operations get staged out? Here’s an example:

```python
import jax.numpy as jnp
from jax import lax

@jit
def select_tril(x):
  mask = jnp.arange(x.shape[0])[:, None] > jnp.arange(x.shape[1])
  return lax.select(mask, x, jnp.zeros_like(x))  # lax.select is like jnp.where

x = np.arange(12).reshape((3, 4))
select_tril(x)
```

```
ENTRY jit_select_tril.8 {
  constant.3 = pred[] constant(false)
  constant.1 = pred[3,4]{1,0} constant({...})
  parameter.2 = s32[3,4]{1,0} parameter(0)
  constant.4 = s32[] constant(0)
  broadcast.5 = s32[3,4]{1,0} broadcast(constant.4), dimensions={}
  select.6 = s32[3,4]{1,0} select(constant.1, parameter.2, broadcast.5)
  ROOT tuple.7 = (s32[3,4]{1,0}) tuple(select.6)
}
```

The `select` operation is staged out, but the operations for constructing the constant `mask` are not. Rather than being staged out, the operations that construct `mask` are executed op-by-op at Python tracing time, and XLA only sees a compile time constant `constant.1` representing the value of `mask`. That’s unfortunate, because if we had staged out the operations for constructing `mask`, XLA could have fused them into the `select` and avoided materializing the result at all. **As a result we end up wasting memory with a large constant, wasting time dispatching multiple un-fused op-by-op XLA computations, and potentially even fragmenting memory.**

(The `broadcast` that corresponds to the construction of the zeros array for `jnp.zeros_like(x)` is staged out because JAX is lazy about very simple expressions from #1668.)

The reason the creation of `mask` is not staged out is that `jit` operates based on data dependence. That is, `jit` stages out only those operations in a function that have a data dependence on an augment. Control flow primitives and `pmap` behave similarly. In the case of `select_tril`, the operations to construct the constant `mask` do not have a data dependence on the argument `x`, so they are not staged out; only the `lax.select` call has a data dependence.

[...]

PR #4271, opened 2020-09-12
omnistaging experiment: print partial eval stats

mattjj:
The idea here is that we can add a useful debugging tool for omnistaging (#3370, #4038) by printing out expressions that would have been constant-folded with pre-omnistaging behavior.

```python
import jax.numpy as jnp
from jax.lax import scan

from jax.config import config
config.enable_omnistaging()

def cumsum(xs):
  _, ys = scan(lambda c, x: (c + x, c + x + jnp.add(1, 1)), 0., xs)
  return ys

cumsum(jnp.arange(10))
```

```
from tracing <lambda> at example.py:8, this was not constant folded:

{ lambda  ; _ _.
  let c = add 1 1
      d = convert_element_type[ new_dtype=float32
                                old_dtype=int32 ] c
  in (*, *, d) }
```


hawkinsp:
Closing, because I doubt many folks need this particular debugging facility any more...

PR #4272, opened 2020-09-12
thread backend in pxla.replicate

### LINKED ISSUE #4223: Nested pmap issue on GPU
Running the code with nested pmap/vmap in the reproducer below works with CPU/TPU but fails on GPU. The explicitly specified `cpu` backend when running `pmap` seems to be ignored, as there are clearly enough devices available.

Changing 

`updates_at_idxs = updates[..., None] * rlax.one_hot(indices, num_classes)` 

to 

`updates_at_idxs = updates[..., None]` 

makes things work on GPU as well. Not entirely sure how this is related to the mapped axes though.

- [Colab reproducer](https://colab.research.google.com/drive/1C8i4xUbu8wDWJKy0EvGhu2Vn7ptAky6W?usp=sharing)

mattjj:
fixes #4223

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -925,7 +925,7 @@ def replicate(val, axis_size, nrep, devices=None, backend=None):
     A ShardedDeviceArray of length `axis_size` where each shard is equal to
     ``val``.
   """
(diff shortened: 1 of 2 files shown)


mattjj:
Internal tests pass!

PR #4273, opened 2020-09-12
fixing typo

alexminnaar:
I assume "...one of more type parameters..." was intended to read "...one or more type parameters..."


shoyer:
thanks!

PR #4276, opened 2020-09-14
Avoid rank promotion in np.outer

romanngg:


mattjj:
Is this meant to make it more consistent with NumPy behavior? If so, can you add a test?

mattjj:
Oh, never mind, I get it: you want to avoid using automatic rank promotion in our own implementations!

mattjj:
Thanks!

PR #4279, opened 2020-09-14
[jax2tf] Build a primitive harness for test_type_promotion.

bchetioui:
We were previously generating the cases using `jtu.cases_from_list`,
which by default dropped 2 test cases (JAX_NUM_GENERATED_CASES=10,
number of generated cases = 12).

### DIFF
--- a/jax/experimental/jax2tf/tests/primitive_harness.py
+++ b/jax/experimental/jax2tf/tests/primitive_harness.py
@@ -875,3 +875,11 @@ random_split = tuple(
                                np.array([0, 0xFFFFFFFF], dtype=np.uint32),
                                np.array([0xFFFFFFFF, 0xFFFFFFFF], dtype=np.uint32)])
 )
+
+type_promotion = tuple(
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -875,3 +875,11 @@ def _fft_rng_factory(dtype):
                                np.array([0, 0xFFFFFFFF], dtype=np.uint32),
                                np.array([0xFFFFFFFF, 0xFFFFFFFF], dtype=np.uint32)])
 )
+
+type_promotion = tuple(
```
gnecula, line 879: This is not really a primitive harness. I am thinking that perhaps all that is needed is to replace the `jtu.cases_from_list` with `tuple`?

SIben: That's fair; I wasn't 100% sure about making it a "primitive harness" since it is really testing something at a higher level. I made the relevant change in `c104889`. Please use that second commit message when merging, then :)

PR #4281, opened 2020-09-14
revert #4277 (google failure)

mattjj:
Some downstream user is relying on the rank of stax's biases being 1.

### DIFF
--- a/jax/experimental/stax.py
+++ b/jax/experimental/stax.py
@@ -19,7 +19,6 @@ For an example of its use, see examples/resnet50.py.
 
 
 import functools
-import itertools
 import operator as op
 
 from jax import lax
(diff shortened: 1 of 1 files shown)


jax/experimental/stax.py
```diff
@@ -49,8 +50,7 @@ def Dense(out_dim, W_init=glorot_normal(), b_init=normal()):
   def init_fun(rng, input_shape):
     output_shape = input_shape[:-1] + (out_dim,)
     k1, k2 = random.split(rng)
-    b_shape = (1,) * (len(output_shape) - 1) + (out_dim,)
-    W, b = W_init(k1, (input_shape[-1], out_dim)), b_init(k2, b_shape)
+    W, b = W_init(k1, (input_shape[-1], out_dim)), b_init(k2, (out_dim,))
```
romanngg, line 53: Could you please only revert this change, but keep the rest? It appears that this is the only part that breaks them

romanngg (approved):

jax/experimental/stax.py
mattjj: Done.

jekbradbury (approved):

PR #4282, opened 2020-09-14
[jax2tf] Added support for x64 for the remaining test files

bchetioui:
This includes:
- control_flow_ops_test.py
- jax2tf_test.py
- saved_model_test.py
- stax_test

### DIFF
--- a/jax/experimental/jax2tf/tests/savedmodel_test.py
+++ b/jax/experimental/jax2tf/tests/savedmodel_test.py
@@ -17,7 +17,6 @@ import os
 from absl.testing import absltest
 
 import jax
-from jax import dtypes
 from jax import lax
 import jax.numpy as jnp
 import numpy as np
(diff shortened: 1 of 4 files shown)


jax/experimental/jax2tf/tests/savedmodel_test.py
```diff
@@ -41,25 +42,28 @@ def save_and_load_model(self, model: tf.Module) -> tf.Module:
 
   def test_eval(self):
     f_jax = jax.jit(lambda x: jnp.sin(jnp.cos(x)))
+    signature_type = tf.float64 if config.FLAGS.jax_enable_x64 else tf.float32
```
gnecula, line 45: I am sure that this works, but is it needed? It should be possible to save a model that takes f32 arguments in x64 mode.

SIben: It is possible to save the model. However, unless we explicitly cast `x` to `(j)np.float32`, `jax` will treat it as 64-bit float by default in x64 mode, while the saved model will still load it as a 32-bit float, and `assertAllClose` will fail because of that. We could instead do `x = np.array(0.7, dtype=jnp.float32)` and then have `tf.float32` as our signature type all the time.

gnecula: I think it is better to cast the input.

PR #4283, opened 2020-09-14
Allow device_get to pass Python scalars through unchanged

trevorcai:


jax/api.py
```diff
@@ -1970,7 +1970,10 @@ def device_put(x, device: Optional[xc.Device] = None):
 def _device_get(x):
   if isinstance(x, core.Tracer):
     return x
-  return x.copy()
+  try:
+    return x.copy()
+  except AttributeError:
+    return x
```
shoyer, line 1976: A more robust version of this would be:
```suggestion
  try:
    copy = x.copy
  except AttributeError:
    return x
  else:
    return copy()
```

trevorcai: ...because it only activates when copy is not an attribute of x, rather than if copy() throws an AttributeError within it?
done.

mattjj (approved):
LGTM

jax/api.py
shoyer: Yes exactly. Otherwise this might hide an error inside the `copy()` implementation.

PR #4284, opened 2020-09-14
Partially revert #4277

jekbradbury:
@romanngg says this is the only part of the change that broke internal code.


jekbradbury:
Oops, raced with @mattjj. Closing.

PR #4287, opened 2020-09-15
Add api.device_put_sharded()

jakevdp:
There are a number of downstream libraries that roll these functions by reaching into internal APIs. Examples are flax [here](https://github.com/google/flax/blob/14883b01c9f75889577e4f6cd54b1ffff0925452/flax/jax_utils.py#L48-L61) and [here](https://github.com/google/flax/blob/14883b01c9f75889577e4f6cd54b1ffff0925452/flax/jax_utils.py#L154-L161), and trax [here](https://github.com/google/trax/blob/67a4edf83fb77cd1a9df872005130cd2c04d1854/trax/layers/acceleration.py#L227-L264), as well as a number of DM projects.

This is a first pass at providing the types of functionality that these downstream libraries need, so that we can decouple them from private APIs.

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -1967,12 +1967,11 @@ def device_put(x, device: Optional[xc.Device] = None):
   return tree_map(lambda y: xla.device_put_p.bind(y, device=device), x)
 
 
-def device_put_sharded(x, devices: Sequence[xc.Device]):
+def device_put_sharded(x: Sequence[Any], devices: Sequence[xc.Device]) -> pxla.ShardedDeviceArray:
   """Shards the input to specified devices, returning ShardedDeviceArrays.
 
   Args:
-    x: An array of leading dimension n_devices, or a sequence of length n_devices
-      containing arrays, scalars, or (nested) standard Python containers thereof.
+    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof.
     devices: A sequence of devices()
 
   Returns:
@@ -1980,14 +1979,14 @@ def device_put_sharded(x, devices: Sequence[xc.Device]):
     of x sharded across the specified devices.
 
   Examples:
-    Sharding a single array results in a sharded copy of that array. Note that the
-    array's leading dimension must equal the number of devices:
+    Passing a list of arrays results in a sharded array containing a stacked version
+    of the inputs. Note that the array's leading dimension must equal the number of devices:
 
     >>> from jax import api, numpy as jnp
     >>> devices = api.local_devices()
-    >>> x = jnp.ones((len(devices), 5))
+    >>> x = [jnp.ones(5) for device in devices]
     >>> y = api.device_put_sharded(x, devices)
-    >>> np.allclose(x, y)
+    >>> np.allclose(y, jnp.stack(x))
     True
 
     Sharding a list of nested objects is equivalent to sharding the list
@@ -2009,6 +2008,8 @@ def device_put_sharded(x, devices: Sequence[xc.Device]):
   - device_put
   - device_put_replicated
   """
+  if not isinstance(x, Sequence):
+    raise ValueError(f"x must be a sequence; got {type(x)}")
   # TODO(jakevdp): provide a default for devices that considers both local devices and pods
   assert len(x) == len(devices), f"len(x) = {len(x)} must equal len(devices) = {len(devices)}."
   def _device_put_sharded(*xs) -> pxla.ShardedDeviceArray:
@@ -2021,7 +2022,7 @@ def device_put_sharded(x, devices: Sequence[xc.Device]):
   return tree_multimap(_device_put_sharded, *x)
 
 
-def device_put_replicated(x, devices: Sequence[xc.Device]):
+def device_put_replicated(x: Any, devices: Sequence[xc.Device]) -> pxla.ShardedDeviceArray:
   """Replicates the input across specified devices, returning ShardedDeviceArrays.
 
   Args:
--- a/jax/api.py
+++ b/jax/api.py
@@ -1971,17 +1971,50 @@ def device_put_sharded(x, devices: Sequence[xc.Device]):
   """Shards the input to specified devices, returning ShardedDeviceArrays.
 
   Args:
-    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof
+    x: An array of leading dimension n_devices, or a sequence of length n_devices
+      containing arrays, scalars, or (nested) standard Python containers thereof.
     devices: A sequence of devices()
 
   Returns:
     A ShardedDeviceArray or (nested) Python container thereof containing a copy
     of x sharded across the specified devices.
+
+  Examples:
+    Sharding a single array results in a sharded copy of that array. Note that the
+    array's leading dimension must equal the number of devices:
+
+    >>> from jax import api, numpy as jnp
+    >>> devices = api.local_devices()
+    >>> x = jnp.ones((len(devices), 5))
+    >>> y = api.device_put_sharded(x, devices)
+    >>> np.allclose(x, y)
+    True
+
+    Sharding a list of equivalent nested objects is equivalent to sharding the list
+    of each entry and repackaging the results to mach
+
+    >>> x = [(i, jnp.arange(i, i + 4)) for i in range(len(devices))]
+    >>> y = api.device_put_sharded(x, devices)
+    >>> type(y)
+    tuple
+    >>> y0 = api.device_put_sharded([a for a, b in x], devices)
+    >>> y2 = api.device_put_sharded([b fr a, b in x], devices)
+    >>> np.allclose(y[0], y0)
+    True
+    >>> np.allclose(y[1], y1)
+    True
+
+  See also:
+  - device_put
+  - device_put_replicated
   """
   # TODO(jakevdp): provide a default for devices that considers both local devices and pods
   assert len(x) == len(devices), f"len(x) = {len(x)} must equal len(devices) = {len(devices)}."
   def _device_put_sharded(*xs) -> pxla.ShardedDeviceArray:
-    aval = ShapedArray((len(xs),) + np.shape(xs[0]), dtypes.result_type(xs[0]))
+    avals = [core.get_aval(x) for x in xs]
+    assert all(aval == avals[0] for aval in avals), f"abstract values must all match; got {avals}"
+    x_aval = core.raise_to_shaped(avals[0])
+    aval = ShapedArray((len(devices),) + x_aval.shape, x_aval.dtype)
     buffers = [xla.device_put(x, d) for x, d in zip(xs, devices)]
     return pxla.ShardedDeviceArray(aval, buffers)
   return tree_multimap(_device_put_sharded, *x)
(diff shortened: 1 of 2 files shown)


jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+  Args:
+    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof
+    devices: A sequence of devices()
+
+  Returns:
+    A ShardedDeviceArray or (nested) Python container thereof containing a copy
+    of x sharded across the specified devices.
+  """
+  # TODO(jakevdp): provide a default for devices that considers both local devices and pods
+  assert len(x) == len(devices), f"len(x) = {len(x)} must equal len(devices) = {len(devices)}."
```
skye, line 1981: I think this should go inside `_device_put_sharded`, `x` is the pytree right?

jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+  def _device_put_sharded(*xs) -> pxla.ShardedDeviceArray:
+    aval = ShapedArray((len(xs),) + np.shape(xs[0]), dtypes.result_type(xs[0]))
```
skye, line 1983: assert that all xs have the same shape and dtype

jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+
+  Args:
+    x: An array, scalar, or (nested) standard Python container thereof
+    devices: A sequence of devices()
+
+  Returns:
+    A ShardedDeviceArray or (nested) Python container thereof containing a copy
+    of x replicated across the specified devices.
+  """
+  return device_put_sharded(len(devices) * [x], devices)
```
skye, line 2000: I think you have to `tree_multimap` over `x` again

tests/api_test.py
```diff
@@ -625,6 +625,24 @@ def test_device_put_across_platforms(self):
...
+
+  def test_device_put_replicated(self):
+    devices = api.local_devices()
+    n_devices = len(devices)
+    x = np.arange(4)
+    y = api.device_put_replicated(x, devices)
+    self.assertEqual(len(y.device_buffers), len(devices))
+    self.assertTrue(all(b.device() == d for b, d in zip(y.device_buffers, devices)))
+    self.assertAllClose(y, jnp.stack(n_devices * [x]))
+
```
skye, line 645: Can you also add a pytree test for put_sharded and put_replicate? I think testing a single pytree for each should be sufficient.

jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+  """Shards the input to specified devices, returning ShardedDeviceArrays.
+
+  Args:
+    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof
+    devices: A sequence of devices()
+
+  Returns:
+    A ShardedDeviceArray or (nested) Python container thereof containing a copy
+    of x sharded across the specified devices.
+  """
```
skye, line 1979: The flax replicate function has `x = jax.numpy.asarray(x)`. The other versions don't have this, so I'm not sure it's necessary, but you might wanna ask about it or include it to be safe (maybe it's to handle replicating scalars? Does device_put not handle that?)

skye (changes requested):

jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
   return tree_map(lambda y: xla.device_put_p.bind(y, device=device), x)
 
 
+def device_put_sharded(x, devices: Sequence[xc.Device]):
+  """Shards the input to specified devices, returning ShardedDeviceArrays.
+
+  Args:
+    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof
```
jekbradbury, line 1973: Not a sequence, right?

jax/api.py
```diff
@@ -1966,6 +1966,40 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+  assert len(x) == len(devices), f"len(x) = {len(x)} must equal len(devices) = {len(devices)}."
+  def _device_put_sharded(*xs) -> pxla.ShardedDeviceArray:
+    aval = ShapedArray((len(xs),) + np.shape(xs[0]), dtypes.result_type(xs[0]))
+    buffers = [xla.device_put(x, d) for x, d in zip(xs, devices)]
+    return pxla.ShardedDeviceArray(aval, buffers)
+  return tree_multimap(_device_put_sharded, *x)
+
+
+def device_put_replicated(x, devices: Sequence[xc.Device]):
+  """Replicates the input across specified devices, returning ShardedDeviceArrays.
```
jekbradbury, line 1990: I would add "broadcasted over an extra logical axis" or something to that effect (and/or add an API to create SDAs that are replicated _without_ that extra axis)

jax/api.py
jakevdp: I called this out as a sequence, because this function requires that the length of `x` be equal to the length of `devices`, to distinguish this from `device_put_replicated` in which there is no such constraint.

jax/api.py
jakevdp: That's right - I think to exactly replicate the various forms of this that are used by downstream libraries will require some extra wrappers along these lines.

Other implementations pass nested containers to `x`, so calling `jnp.asarray(x)` would not work in all cases.

jax/api.py
jakevdp: I think either would work, this catches issues earlier.

jax/api.py
jakevdp: If I understand your comment correctly, I think the version that replicates without the extra axis is `device_put_sharded`.

jax/api.py
jakevdp: I ended up reimplimenting this in terms of `tree_map` for efficiency & clarity.

tests/api_test.py
jakevdp: Done.

jax/api.py
jakevdp: Added the additional language as well.

jax/api.py
jakevdp: Done.

jax/api.py
jakevdp: Updated the description to be more clear.

jekbradbury: Ahh, ok, then I'm not sure that "Shards the input" is the right description, because it doesn't cut any input into pieces. Maybe "Creates a sharded device array representing the input stacked over devices", although I don't really like that language either?

jax/api.py
jekbradbury: Hmm. By "replicates without the extra axis", I mean "creates an SDA that uses `replication_factors` to represent an array replicated over devices" (the aval of the SDA result is the same as the aval of the DA input).

jakevdp: Maybe the issue is the terminoligy I'm using. I want this function to be something that takes an input and returns a sharded device array containing the input stacked over the number of devices, because empirically this is a common use-case in downstream libraries. Is there a better name for this than "replicate"?

jax/api.py
jakevdp: It does cut the input into pieces if the input is an array. If the input is a list, it's essentially treated as a list of pre-determined shards.

jax/api.py
jekbradbury: There are two separate questions: what is the function doing logically, and what is it doing in terms of devices/buffers:

- Logically, you can imagine functions that concatenate or stack a list of inputs, and functions that represent an identity or a broadcast of a single input.
- Device-wise, we want all of these to have the resulting array reside across multiple devices, but there are a few ways to do this: either an identical copy of the whole logical array on every device, or a separate piece of the logical array on each device.

jax/api.py
jekbradbury: Thanks, that clears things up for me. Allowing either an array or a list seems like a good approach here and one that's consistent with NumPy expectations, but it that should maybe be documented more explicitly in the JAX context since we don't do it as much.

jax/api.py
jekbradbury: In MLPerf code, we had a `replicate` function with a `broadcast: bool` kwarg. When `broadcast=False`, it represented an identity function that results in an output SDA with the same aval as the input, but contains copies of the whole array on every device and uses `replication_factors` to encode this. When `broadcast=True`, it represented a broadcast function that results in an output SDA with an extra logical axis, containing copies of the whole _original_ array on every device (but now they're slices of the new logical array), and for which `replication_factors` is empty.

jakevdp: OK, maybe more directly to the point: I want the functions to do what I wrote in the code here, because this is the functionality that downstream libraries need. What should the functions be called?

jax/api.py
jakevdp: I'm happy to not allow an array here, because the functionality that downstream libraries need to turn a list of arrays into a stacked sharded array.

jax/api.py
jekbradbury: Sorry for being a little obnoxious: it's the functionality that downstream libraries currently have, but not necessarily what they need/should have in the long run. A "replicate" function that's logically a broadcast is only needed while `pmap` doesn't fully support `(in/out)_axes=None`, and becomes tech debt as soon as that support lands (since the broadcast doesn't do anything for users, and e.g. makes this API useless with `soft_pmap`).

So I'd suggest using a name (or kwarg) with "broadcast" in it for this function. Maybe `api.device_put_broadcasted`? Or `api.device_put_replicated(broadcast=True)`? For the other one, `api.device_put_sharded` sounds fine.

jakevdp: OK, I see. Maybe better then would be to remove the `device_put_replicated(x, devices)` function altogether, since it's functionally equivalent to `device_put_sharded([x] * len(devices), devices)` (though there's a bit more validation overhead this way).

jax/api.py
jakevdp: Done

jax/api.py
jakevdp: Alright, all that's left is the `device_put_sharded` function, which has clearer semantics.

jax/api.py
```diff
@@ -1967,6 +1967,61 @@ def device_put(x, device: Optional[xc.Device] = None):
   return tree_map(lambda y: xla.device_put_p.bind(y, device=device), x)
 
 
+def device_put_sharded(x: Sequence[Any], devices: Sequence[xc.Device]) -> pxla.ShardedDeviceArray:
+  """Shards the input to specified devices, returning ShardedDeviceArrays.
```
skye, line 1971:
```suggestion
  """Transfers pre-sharded input to the specified devices, returning ShardedDeviceArrays.
```

This comment threw me off initially. It's not actually sharding the input, it's assembling ShardedDeviceArrays from already-sharded data.

I would also consider renaming the function, maybe `multi_device_put`? I'm ok with the current name, but it still makes me think it's gonna do the sharding itself (that might be my prior bias though).

jax/api.py
```diff
@@ -1967,6 +1967,61 @@ def device_put(x, device: Optional[xc.Device] = None):
...
+
+  Args:
+    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof.
+    devices: A sequence of devices()
+
+  Returns:
+    A ShardedDeviceArray or (nested) Python container thereof containing a copy
+    of x sharded across the specified devices.
```
skye, line 1979:
```suggestion
    A ShardedDeviceArray or (nested) Python container thereof containing a stacked
    version of x sharded across the specified devices.
```

I'm less attached to this suggestion, but same idea as above.

skye (approved):
I was confused before, your expanded comment was very helpful!

jekbradbury (approved):

jakevdp:
Thanks!

PR #4301, opened 2020-09-16
Support polynomial division for mask

juliuskunze:
Add support for multivariate polynomial division on polymorphic sizes without remainder, allowing `mask` of

- `jnp.reshape` with -1 size and
- `lax.slice` for polymorphic stride for sizes  `poly * stride`, i. e. `(n^2+2n)//n = n+2`

Also clean up `Poly` class, improve error messages.

### DIFF
--- a/jax/interpreters/masking.py
+++ b/jax/interpreters/masking.py
@@ -18,7 +18,7 @@ from functools import partial, reduce
 from itertools import chain, product
 import operator as op
 import string
-from typing import Callable, Dict, Sequence, Union
+from typing import Callable, Dict, Sequence, Union, Tuple
 
 import numpy as np
 
@@ -107,80 +107,91 @@ def eval_poly_shape(shape, values_dict):
 def eval_poly(poly, values_dict):
   return poly.evaluate(values_dict) if type(poly) is Poly else poly
 
-def _ensure_poly(p):
-  if type(p) is Poly:
-    return p
+def _ensure_poly(p: 'Size') -> 'Poly':
+  if isinstance(p, Poly): return p
   return Poly({Mon(): p})
 
 def _polys_to_ints(shape):
   return tuple(int(d) if type(d) is Poly and d.is_constant else d
                for d in shape)
 
-def is_polymorphic(shape: Sequence[Union[int, 'Poly']]):
+def is_polymorphic(shape: Sequence['Size']):
   return any(map(lambda d: type(d) is Poly, shape))
 
 class Poly(dict):
-  """Polynomial with nonnegative integer coefficients for polymorphic shapes."""
+  """Polynomial with integer coefficients for polymorphic shapes."""
 
-  def __init__(self, coeffs):
+  def __init__(self, coeffs: Dict['Mon', int]):
     # Makes sure Polynomials are always in canonical form
     coeffs = {mon: op.index(coeff)
               for mon, coeff in coeffs.items() if coeff != 0}
     coeffs = coeffs or {Mon(): 0}
     super().__init__(coeffs)
 
-  def __add__(self, other):
+  def __add__(self, other: 'Size') -> 'Poly':
     coeffs = self.copy()
     for mon, coeff in _ensure_poly(other).items():
       coeffs[mon] = coeffs.get(mon, 0) + coeff
     return Poly(coeffs)
 
-  def __sub__(self, other):
+  def __sub__(self, other: 'Size') -> 'Poly':
     return self + -other
 
-  def __neg__(self):
+  def __neg__(self) -> 'Poly':
     return Poly({mon: -coeff for mon, coeff in self.items()})
 
-  def __mul__(self, other):
+  def __mul__(self, other: 'Size') -> 'Poly':
     other = _ensure_poly(other)
-    coeffs = {}
+    coeffs: Dict[Mon, int] = {}
     for (mon1, coeff1), (mon2, coeff2) in product(self.items(), other.items()):
       mon = mon1 * mon2
       coeffs[mon] = coeffs.get(mon, 0) + coeff1 * coeff2
     return Poly(coeffs)
 
-  def __rmul__(self, other):
+  def __rmul__(self, other: 'Size') -> 'Poly':
     return self * other  # multiplication commutes
 
-  def __radd__(self, other):
+  def __radd__(self, other: 'Size') -> 'Poly':
     return self + other  # addition commutes
 
-  def __rsub__(self, other):
+  def __rsub__(self, other: 'Size') -> 'Poly':
     return _ensure_poly(other) - self
 
-  def __floordiv__(self, divisor):
-    q, _ = divmod(self, divisor)  # pytype: disable=wrong-arg-types
+  def __floordiv__(self, divisor: 'Size') -> 'Poly':
+    q, _ = divmod(self, divisor)
     return q
 
-  def __mod__(self, divisor):
-    _, r = divmod(self, divisor)  # pytype: disable=wrong-arg-types
+  def __mod__(self, divisor: 'Size') -> int:
+    _, r = divmod(self, divisor)
     return r
 
-  def __divmod__(self, divisor):
-    if self.is_constant:
-      return divmod(int(self), divisor)
-    else:
-      def divided(count):
-        q, r = divmod(count, divisor)
-        if r != 0:
-          raise ValueError('shapecheck  and masking currently only support '
-                           'strides that exactly divide the strided axis '
-                           'length.')
-        return q
+  def __divmod__(self, divisor: 'Size') -> Tuple['Poly', int]:
+    """
+    Floor division with remainder (divmod) generalized to polynomials. To allow
+    ensuring '0 <= remainder < divisor' for consistency with integer divmod, the
+    divisor must divide the dividend (up to a constant for constant divisors).
+    :return: Quotient resulting from polynomial division and integer remainder.
+    """
+    divisor = _ensure_poly(divisor)
+    dmon, dcount = divisor._leading_term
+    dividend, quotient, remainder = self, _ensure_poly(0), _ensure_poly(0)
+    while dividend != 0:  # invariant: dividend == divisor*quotient + remainder
+      mon, count = dividend._leading_term
+      qcount, rcount = divmod(count, dcount)
+      try:
+        qmon = mon // dmon
+      except ValueError:
+        raise ValueError(f"Stride {divisor} must divide size {self} "
+                         f"(up to a constant for constant divisors).")
+      r = Poly({mon: rcount})
+      q = Poly({qmon: qcount})
+      quotient += q
+      remainder += r
+      dividend -= q * divisor + r
+    return quotient, int(remainder)
 
-      return Poly(
-        {k: coeff // divisor if k.degree == 0 else divided(coeff)
-         for k, coeff in self.items()}), self.get(Mon(), 0) % divisor
+  def __rdivmod__(self, dividend: 'Size') -> Tuple['Poly', int]:
+    return divmod(_ensure_poly(dividend), self)
 
   def __hash__(self):
     return hash(tuple(sorted(self.items())))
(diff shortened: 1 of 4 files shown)


JuliusKunze:
@mattjj This small and new PR is now ready to review. (;

jax/interpreters/masking.py
```diff
@@ -239,6 +245,12 @@ def evaluate(self, env):
   def is_constant(self):
     return len(self) == 1 and next(iter(self)).degree == 0
 
+  @property
+  def _leading_term(self) -> Tuple['Mon', int]:
```
gnecula, line 249: What is the significance of the leading_term? Is it the constant if one is present?

jax/numpy/lax_numpy.py
```diff
@@ -1135,7 +1135,7 @@ def check(size):
       int, size, "The error arose in jax.numpy.reshape.")
   newshape = [check(size) for size in newshape] if iterable else check(newshape)
   newsize = _prod((newshape,) if type(newshape) is Poly else newshape)
-  if newsize < 0:
+  if np.any(np.equal(newshape, -1)):
```
gnecula, line 1138: I think this is a better check than before. I am wondering though if other negative dimensions will be caught reliably somewhere else.

jax/interpreters/masking.py
```diff
@@ -107,77 +107,93 @@ def eval_poly_shape(shape, values_dict):
...
-def _ensure_poly(p):
-  if type(p) is Poly:
-    return p
-
+def _ensure_poly(p: 'Size') -> 'Poly':
+  if isinstance(p, Poly): return p
   return Poly({Mon(): p})
 
-def is_polymorphic(shape: Sequence[Union[int, 'Poly']]):
+def is_polymorphic(shape: Sequence['Size']):
```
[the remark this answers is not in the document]
gnecula: I saw the definition later. If you move it up, you won't need to put in in quotes

jax/lax/lax.py
```diff
@@ -3599,9 +3599,9 @@ def _slice_shape_rule(operand, *, start_indices, limit_indices, strides):
       msg = "slice strides must be positive, got {}"
       raise TypeError(msg.format(strides))
 
-  result_shape = np.floor_divide(
-      np.add(np.subtract(limit_indices, start_indices), strides) - 1, strides)
-  return tuple(result_shape)
+  diff = np.subtract(limit_indices, start_indices)
+  # Not np.divmod since Poly.__rdivmod__ is ignored by NumPy, breaks poly stride
+  return tuple(q + np.where(r, 1, 0) for q, r in map(divmod, diff, strides))
```
gnecula, line 3604: I was initially confused by the use of `np.where`, thinking that it is going to break in presence of polynomials. But `r` is always a scalar, right? Then perhaps `(r > 0)` is cleaner than `np.where(r, 1, 0)`?

jax/interpreters/masking.py
```diff
@@ -253,33 +265,43 @@ def mul(coeff, mon):
...
 
-  def __mul__(self, other):
+  def __mul__(self, other: 'Mon') -> 'Mon':
     return Mon(Counter(self) + Counter(other))
 
   @property
   def degree(self):
     return sum(self.values())
 
+  def __floordiv__(self, divisor: 'Mon') -> Optional['Mon']:
```
gnecula, line 294: Since you clearly figured out the representation of these monomials, do you mind adding a few comments to this class at the top?

jax/interpreters/masking.py
```diff
@@ -107,77 +107,93 @@ def eval_poly_shape(shape, values_dict):
...
+    return quotient, int(remainder)
+
+  def __rdivmod__(self, dividend: 'Size') -> Tuple['Poly', int]:
+    return divmod(_ensure_poly(dividend), self)
+
+  def _polynomial_divmod(self, divisor: 'Poly') -> Tuple['Poly', 'Poly']:
+    """Multivariate polynomial division, returning quotient and remainder."""
+    dmon, dcount = divisor._leading_term
+    dividend, quotient, remainder = self, _ensure_poly(0), _ensure_poly(0)
+    while dividend != 0:
```
gnecula, line 187: I think it would help if you wrote down the loop invariant that dividend = divisor * quotient + remainder. 

I am trying to convince myself that this is "complete". E.g., if the leading term of the dividend and the divisor are not the same, then the monomial division will return None, and the entire leading term of the dividend is added to the remainder. I guess I am not sure what means the leading term. If it were the one with the highest degree, I could see how this terminates. But otherwise, why does it terminate?

gnecula (commented):
This is a welcome change, thank you! I have a few comments, perhaps I did not fully understand the details.

mattjj:
Multivariate polynomial division in general requires a doubly exponential time algorithm; this is what [Grobner bases](https://en.wikipedia.org/wiki/Gr%C3%B6bner_basis) are for. I don't think that's a direction we want to go.

mattjj:
Thanks for continuing to try new ideas @JuliusKunze ! To say a bit more, it's intentional that we don't support `lax.slice` with polymorphic stride: in general `a // b` is not a polynomial for general polynomials, and a key desideratum is that the language of polymorphic shapes be closed and tractable to decide equality (polynomial equality can be checked in BPP).

Our current plan with strides is to support only division of the form `(c * poly + k) // c == poly + k // c` for integers `c` and `k`. If we can enlarge that to a bigger family of tractable polynomial divisions then we should! But the key question is what that family is, as we know general multivariate polynomial division is intractable.

gnecula:
I have made some comments in a related design doc that is not yet external, so I repeat them here. 

I think it is important to have a clear mental model of what is supported and what is not in terms of polynomial division, but we should think how to enlarge the class of supported division operation. It would be unsatisfying to see an error that '6x' cannot be divided by 'x'. 

I propose at least to add to the supported class the dual of the one @mattjj pointed out: `(c * poly) // poly ==  c`. 

I am also wondering how often people will only need univariate polynomials. In a use case I am thinking about, we use one shape variable to be standing for the batch dimension. Perhaps it makes sense to also support division of univariate polynomials. 

Yet another case that I think may occur frequently is when the polynomials are just monomials (no `+`). I am not sure, and it would be best to get some experience using this before we decide what further exceptions to add.

jax/interpreters/masking.py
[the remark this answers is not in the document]
JuliusKunze: Had the same thought, but one can't move it up, since it requires Poly to be defined already.

jax/interpreters/masking.py
```diff
@@ -107,77 +107,93 @@ def eval_poly_shape(shape, values_dict):
...
-def is_polymorphic(shape: Sequence[Union[int, 'Poly']]):
+def is_polymorphic(shape: Sequence['Size']):
   return any(map(lambda d: type(d) is Poly, shape))
 
 class Poly(dict):
-  """Polynomial with nonnegative integer coefficients for polymorphic shapes."""
+  """Polynomial with integer coefficients for polymorphic shapes."""
 
-  def __init__(self, coeffs):
+  def __init__(self, coeffs: Dict['Mon', int]):
```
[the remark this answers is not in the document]
JuliusKunze: I did intentionally not do that, for now, to minimize the PR diff.

jax/interpreters/masking.py
JuliusKunze: It's needed by polynomial division to determine the next term to divide.

jax/numpy/lax_numpy.py
JuliusKunze: Comment a few lines above said: "# other errors, like having more than one -1, are caught downstream".
In any case, the change does not worsen error handling.

jax/lax/lax.py
JuliusKunze: Good point, updated to `r > 0`!

jax/interpreters/masking.py
```diff
@@ -107,77 +107,93 @@ def eval_poly_shape(shape, values_dict):
...
-        return q
-
-      return Poly(
-        {k: coeff // divisor if k.degree == 0 else divided(coeff)
-         for k, coeff in self.items()}), self.get(Mon(), 0) % divisor
+  def __divmod__(self, divisor: 'Size') -> Tuple['Poly', int]:
+    """
+    Floor division with remainder (divmod) generalized to polynomials.
+    To allow ensuring consistency with integer divmod,
+    the polynomial division remainder divided by the divisor must be constant.
```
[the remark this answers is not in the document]
JuliusKunze: Clarified the description!

jax/interpreters/masking.py
JuliusKunze: Ok, done!

jax/interpreters/masking.py
JuliusKunze: @gnecula I described the invariant in the code. 

The leading terms is the highest degree term that comes first lexicographically. (The monomials are ordered in [graded reverse lexicographical order](https://en.wikipedia.org/wiki/Monomial_order#Graded_reverse_lexicographic_order): First by total degree and then lexicographically to break ties.)

JuliusKunze:
@mattjj perfect division (`(c * poly) // poly == c` as described by @gnecula) is exactly the class this PR adds. +1 for adding it to the spec since it's useful for the listed cases and doesn't require Gröbner bases.

@gnecula The cases of univariate polynomials and polynomials with only one term are already covered by that extension, or did you have another use case in mind?

gnecula:
This LGTM but I know that @mattjj has been working on updating this part of the code. He should be the one to give this PR the final verdict.

jax/interpreters/masking.py
```diff
@@ -253,33 +262,53 @@ def mul(coeff, mon):
...
   else:
-    return  0 if coeff == 0 else mon if coeff == 1 else coeff * mon
+    return 0 if coeff == 0 else mon if coeff == 1 else coeff * mon
 
 
 abstract_arrays._DIMENSION_TYPES.add(Poly)
 
-
 class Mon(dict):
+  """Represents a multivariante monomial, such as n^3 * m."""
```
gnecula, line 271: Typo: multivariate

jax/interpreters/masking.py
```diff
@@ -253,33 +262,53 @@ def mul(coeff, mon):
...
 
   def __str__(self):
-    return ' '.join('{}**{}'.format(k, v) if v != 1 else str(k)
-                    for k, v in sorted(self.items()))
+    return ' '.join(f'{key}^{exponent}' if exponent != 1 else str(key)
+                    for key, exponent in sorted(self.items()))
 
-  def __lt__(self, other):
-    # sort by total degree, then lexicographically on indets
+  def __lt__(self, other: 'Mon'):
```
gnecula, line 279: I find this strange, and hard to understand. IIUC, this is not the numeric comparison, just a sorting order for monomials based on their representation. 

The "-self.degree" along with ">" used in an implementation of "__lt__" is confusing.  

Is this needed only for _leading term? If so, wouldn't it be clearer if in that function we define this as sorting key without overloading __lt__?

gnecula:
@mattjj I think that this PR is worth merging. It does enable a common idiomatic use of `reshape(-1)`.

jax/interpreters/masking.py
```diff
@@ -107,77 +107,90 @@ def eval_poly_shape(shape, values_dict):
...
+    divisor = _ensure_poly(divisor)
+    dmon, dcount = divisor._leading_term
+    dividend, quotient, remainder = self, _ensure_poly(0), _ensure_poly(0)
+    # loop invariant: dividend = divisor * quotient + remainder
+    while dividend != 0:
+      mon, count = dividend._leading_term
+      qmon = mon // dmon
+      qcount, rcount = divmod(count, dcount)
+      q = Poly({qmon: qcount}) if qmon else _ensure_poly(0)
+      r = Poly({mon: rcount if qmon else count})
```
gnecula, line 180: I still find this code very hard to understand. 

Does it even make sense to continue if qmon is None or if rcount is not 0? The remainder will be non-constant. An exception has to be made for the case when the divisor is a constant. If we abort early, the code is easier to understand: we only consider the case when the leading term of the divisor divides the leading term of the dividend.

jax/interpreters/masking.py
JuliusKunze: @gnecula I think what you suggest is not equivalent, see https://en.wikipedia.org/wiki/Monomial_order#Graded_reverse_lexicographic_order

I replaced the direction here since `x^2<x` (how it was before) is unintuitive.
Semantically, various other orders could be used without breaking poly division. I chose this order to allow creating string representations following `PolyTest.test_parse_spec` just by sorting.

jax/interpreters/masking.py
JuliusKunze: @gnecula Good point, simplified it.

gnecula:
@JuliusKunze I am sorry this took so long, I will merge it. 

Unfortunately, there are a few pytype failures:
```
File "third_party/py/jax/interpreters/masking.py", line 161, in __floordiv__: Built-in function divmod was called with the wrong arguments [wrong-arg-types]
         Expected: (x: int, ...)
  Actually passed: (x: Poly, ...)
File "third_party/py/jax/interpreters/masking.py", line 165, in __mod__: Built-in function divmod was called with the wrong arguments [wrong-arg-types]
         Expected: (x: int, ...)
  Actually passed: (x: Poly, ...)
File "third_party/py/jax/interpreters/masking.py", line 194, in __rdivmod__: Built-in function divmod was called with the wrong arguments [wrong-arg-types]
         Expected: (x: int, ...)
  Actually passed: (x: Poly, ...)
```

These come from pytype, which we use inside Google.

gnecula:
I will take care of fixing these errors on import.

mattjj:
LGTM. Thanks for following up on this, George and @JuliusKunze.

I agree about supporting special cases of division that are easy. Thanks for clarifying the scope of these changes!

mattjj (approved):
LGTM. I started (but have yet to finish) a refactor that will likely affect some of the implementation details here, but not the functionality, which sounds good!

PR #4312, opened 2020-09-16
improve an escaped tracer error message

mattjj:
Before this commit, encountering an escaped tracer in a specific way would lead to a bad internal error. This change
1. raises an UnexpectedTracerError instead, and
2. includes in the error message the user source line which created the
tracer.

Here's a repro for the original issue, which models an error several users started encountering with Flax and Haiku after we switched omnistaging on by default (#4038):

```python
from jax import jit
from jax import random

key = random.PRNGKey(0)

def init():
  global key
  return random.normal(subkey, ())

print(init())  # -1.2515389
print(init())  # -0.58665067

init = jit(init)
print(init())  # 0.48648298
print(init())  # 0.48648298  !!

# That last call has repeated randomness but no error, because we aren't
# re-executing the Python function and thus aren't actually touching the global
# Python key. But if we look at the key, we see an escaped tracer when
# omnistaginig is on (and no tracer when omnisitaging is off):

print(key)

# If we touch that key again, we'll get an error.

# If we touch it this way:
#
#   random.normal(key, ())
#
# We'll get this error:
#
#   jax.core.UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global      state from a previously traced function.
#   The functions being transformed should not save traced values to global state.
#   Details: Can't lift sublevels 1 to 0.


# But if we touch it in just the right way, we can get a crappy internal error:

from jax import lax

def f(x, c):
  random.normal(key, ())
  return None, None

@jit
def g():
  lax.scan(f, None, None, length=2)

g()

#    File "/usr/local/google/home/mattjj/packages/jax/jax/interpreters/partial_eval.py", line 909, in new_const
#      self.frame.tracers.append(tracer)
#    File "/usr/local/google/home/mattjj/packages/jax/jax/interpreters/partial_eval.py", line 899, in frame
#      def frame(self): return self.main.jaxpr_stack[-1]  # pytype: disable=attribute-error
#  IndexError: tuple index out of range
```

That's a bad internal error!

Here's the error after this change:

```
jax.core.UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this
tracer escaped through global state from a previously traced function.
Detail: tracer created on line example.py:8 (init).
```

To make for a better error message, we
1. add an optional mechanism for `Tracer`s to implement for checking whether its corresponding main trace is still alive;
2. implement that mechanism for `DynamicJaxprTracer` (used in `jit`, `pmap`, and control flow), since those main traces already have a flag for whether it's still alive (i.e. whether the `jaxpr_stack` is nonempty);
3. update `find_top_trace` to check for liveness using that mechanism (it's the latest we can check for liveness while still having the offending escaped tracer in scope); and
4. add source info to each `DynamicJaxprTracer` instance for it to use in its `UnexpectedTracerError` message (we were already computing the source info for each jaxpr eqn created, and we just had to plumb it into the tracers).

cc @LenaMartens

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -464,9 +464,8 @@ class Tracer:
   def aval(self):
     raise NotImplementedError("must override")
 
-  @property
-  def _live(self):
-    return True  # Override for liveness checking
+  def _assert_live(self) -> None:
+    pass  # Override for liveness checking
 
   # Python looks up special methods only on classes, not instances. This means
   # these methods needs to be defined explicitly rather than relying on
@@ -743,7 +742,7 @@ def full_lower(val):
     return val
 
 def find_top_trace(xs) -> Trace:
-  traces = [x._live and x._trace.main for x in xs if isinstance(x, Tracer)]
+  traces = [x._assert_live() or x._trace.main for x in xs if isinstance(x, Tracer)]  # type: ignore
   top_main = max(traces, default=None, key=attrgetter('level'))
   dynamic = thread_local_state.trace_state.trace_stack.dynamic
   top_main = (dynamic if top_main is None or dynamic.level > top_main.level
--- a/jax/core.py
+++ b/jax/core.py
@@ -422,12 +422,14 @@ class Trace:
     del primitive, fwd, bwd, out_trees  # Unused.
     return fun.call_wrapped(*tracers)
 
-def escaped_tracer_error(detail):
+def escaped_tracer_error(detail=None):
   msg = ("Encountered an unexpected tracer. Perhaps this tracer escaped "
          "through global state from a previously traced function.\n"
          "The functions being transformed should not save traced values to "
-         "global state.\nDetails: {}.")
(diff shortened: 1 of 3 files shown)


jax/core.py
```diff
@@ -462,6 +464,10 @@ def __len__(self):
   def aval(self):
     raise NotImplementedError("must override")
 
+  @property
+  def _live(self):
```
jekbradbury, line 468: Do you expect other kinds of tracers to want to implement this as returning `True` or `False`? Otherwise, the implementation you're using here (True or error) is maybe a better fit for an `_assert_live(self)` method that errors or returns `None`.

jekbradbury (approved):

jax/core.py
mattjj: Good point! I think True or error is better. I'll try to tweak it.

PR #4313, opened 2020-09-16
Implement jnp.ravel_multi_index()

jakevdp:


jakevdp:
Internal tests pass.

froystig (approved):
Nice. Do we want to check upfront, and err with a more specific message, if this is called in `'raise'` mode with non-static first argument?

jakevdp:
> Nice. Do we want to check upfront, and err with a more specific message, if this is called in 'raise' mode with non-static first argument?

Good idea. I added an explicit `ValueError` when the `multi_index` value is traced. Is that what you had in mind, or is there a standard utility for this that I'm not aware of?

froystig:
I had `core.concrete_or_error` in mind, which is used elsewhere throughout `lax_numpy`. It allows tracers as long as they're concrete.

jakevdp:
OK, Done.

PR #4314, opened 2020-09-17
Make expm transposable and remove custom_jvp rule.

### LINKED ISSUE #3447: Add VJP expm
The pull request #2062 implements the jvp of the matrix exponential by using the Fréchet derivative. I think the vjp can be easily added by transposing the input matrix in `expm_frechet`:
```python
defvjp(expm, lambda g, ans, matrix: expm_frechet(_T(matrix), g, compute_expm=False))
```
I can test and implement this solution if #2062 is merged.

zhangqiaorjc:


jax/scipy/linalg.py
```diff
@@ -238,18 +237,28 @@ def triu(m, k=0):
...
 
-@partial(api.custom_jvp, nondiff_argnums=(1,))
-def _expm(A, upper_triangular):
+def _expm(A, upper_triangular, max_n_squarings):
   P, Q, n_squarings = _calc_P_Q(A)
-  R = _solve_P_Q(P, Q, upper_triangular)
-  R = _squaring(R, n_squarings)
+
+  def _nan(_):
+    return jnp.zeros(A.shape, dtype=A.dtype) * jnp.nan
```
shoyer, line 254: Can you use `jnp.full_like` here?
```suggestion
    return jnp.full_like(A, jnp.nan)
```

jax/scipy/linalg.py
```diff
@@ -300,11 +309,19 @@ def _precise_dot(A, B):
...
+  def _identity(x):
+    return x
+
+  def scan_loop(n_squaring, R):
+    def scan_f(c, i):
+      return lax.cond(i < n_squaring, _squaring_precise, _identity, c), None
+    res, _ = lax.scan(scan_f, R, jnp.arange(16))
+    return res
+
+  return scan_loop(n_squarings, R)
```
shoyer, line 324: nit: you could remove one layer of indirection here:
```suggestion
  def scan_f(c, i):
    return lax.cond(i < n_squaring, _squaring_precise, _identity, c), None
  res, _ = lax.scan(scan_f, R, jnp.arange(16))
  return res
```

jax/scipy/linalg.py
```diff
@@ -238,18 +237,28 @@ def triu(m, k=0):
...
+
+  def _nan(_):
+    return jnp.zeros(A.shape, dtype=A.dtype) * jnp.nan
+
+  def _compute(_):
+    R = _solve_P_Q(P, Q, upper_triangular)
+    R = _squaring(R, n_squarings)
+    return R
+
+  R = lax.cond(n_squarings > max_n_squarings, _nan, _compute, None)
```
shoyer, line 261: I think it's important (at least, pre-omnistaging) to include the function arguments as the `operand`, otherwise the computation doesn't have the right data dependency. But omnistaging is the future, so maybe this OK :)

jax/scipy/linalg.py
```diff
@@ -238,18 +237,28 @@ def triu(m, k=0):
...
 
 @_wraps(scipy.linalg.expm, lax_description=_expm_description)
-def expm(A, *, upper_triangular=False):
-  return _expm(A, upper_triangular)
+def expm(A, *, upper_triangular=False, max_n_squarings=16):
+  return _expm(A, upper_triangular, max_n_squarings)
 
-@partial(api.custom_jvp, nondiff_argnums=(1,))
-def _expm(A, upper_triangular):
+def _expm(A, upper_triangular, max_n_squarings):
```
shoyer, line 250: Can you put this whole helper function under a `jit`? The last two arguments should be static.

jax/scipy/linalg.py
```diff
@@ -238,18 +237,32 @@ def triu(m, k=0):
...
   P, Q, n_squarings = _calc_P_Q(A)
-  R = _solve_P_Q(P, Q, upper_triangular)
-  R = _squaring(R, n_squarings)
+
+  def _nan(args):
+    A, *_ = args
+    return jnp.zeros(A.shape, dtype=A.dtype) * jnp.nan
+
+  def _compute(args):
+    A, P, Q, upper_triangular, n_squarings = args
```
shoyer, line 259: Can you remove `upper_triangular` and  `n_squarings` from `args`? That way they will remain static variables (instead of dynamic variables that are explicitly included in the arguments to `lax.cond`)
```suggestion
    A, P, Q = args
```

jax/scipy/linalg.py
zhangqiaorjc: Done

jax/scipy/linalg.py
zhangqiaorjc: Done

jax/scipy/linalg.py
zhangqiaorjc: Good idea! Added explicit arguments in case people turn off omnistaging.

jax/scipy/linalg.py
zhangqiaorjc: Done.

jax/scipy/linalg.py
zhangqiaorjc: Fixed.

zhangqiaorjc:
Thanks Stephan! All comments resolved.

tests/linalg_test.py
```diff
@@ -1318,7 +1318,9 @@ def testExpmGrad(self, n, dtype, rng_factory):
      rng = rng_factory(self.rng())
      jtu.skip_if_unsupported_type(dtype)
      a = rng((n, n), dtype)
-     jtu.check_grads(jsp.linalg.expm, (a,), modes=["fwd"], order=1)
+     def expm(x):
+       return jsp.linalg.expm(x, upper_triangular=False, max_n_squarings=16)
+     jtu.check_grads(expm, (a,), modes=["fwd", "rev"], order=1)
```
shoyer, line 1323: could we test `order=2`, also? I think the gradient should also be differentiable.

zhangqiaorjc: Done.

jax/scipy/linalg.py
```diff
@@ -238,18 +237,31 @@ def triu(m, k=0):
 _expm_description = textwrap.dedent("""
 In addition to the original NumPy argument(s) listed below,
 also supports the optional boolean argument ``upper_triangular``
-to specify whether the ``A`` matrix is upper triangular.
+to specify whether the ``A`` matrix is upper triangular, and the optional
+argument ``max_n_squarings`` to specify the max number of squarings allowed
+in the scaling-and-squaring approximation method. Return nan if the actual
+number of squarings required is more than ``max_n_squarings``.
```
shoyer, line 243: It would be nice to add a reference on appropriate values of `max_n_squarings`, or at least some guidance on why you would want a bigger number (e.g., for larger matrix norm). You could even note the formula, e.g., `n_squarings = max(0, floor(log2(sum(abs(A))) - c))` with `c=2.43` for float64 and `c=1.97` for float32.

jax/scipy/linalg.py
```diff
@@ -238,18 +237,31 @@ def triu(m, k=0):
...
 @_wraps(scipy.linalg.expm, lax_description=_expm_description)
-def expm(A, *, upper_triangular=False):
-  return _expm(A, upper_triangular)
+def expm(A, *, upper_triangular=False, max_n_squarings=16):
+  return _expm(A, upper_triangular, max_n_squarings)
 
-@partial(api.custom_jvp, nondiff_argnums=(1,))
-def _expm(A, upper_triangular):
+@partial(jit, static_argnums=(1, 2))
+def _expm(A, upper_triangular, max_n_squarings):
```
shoyer, line 251: Would `max_squarings` be a better variable name? `n` seems a bit redundant.

zhangqiaorjc: SG. Done.

tests/linalg_test.py
```diff
@@ -1318,7 +1318,10 @@ def testExpmGrad(self, n, dtype, rng_factory):
      rng = rng_factory(self.rng())
      jtu.skip_if_unsupported_type(dtype)
      a = rng((n, n), dtype)
-     jtu.check_grads(jsp.linalg.expm, (a,), modes=["fwd"], order=1)
+     def expm(x):
+       return jsp.linalg.expm(x, upper_triangular=False, max_squarings=16)
+     jtu.check_grads(expm, (a,), modes=["fwd", "rev"], order=1)
```
shoyer, line 1323: You can remove this line. `order=2` means "all orders up to 2", so this is covered in the line below.

shoyer (approved):

tests/linalg_test.py
zhangqiaorjc: Done.

PR #4319, opened 2020-09-17
[jax2tf] Disable complex convolution test on GPU: crash in TF

gnecula:


bchetioui (approved):

bchetioui (commented):
Is it possible that this would also happen on TPU?

gnecula:
> Is it possible that this would also happen on TPU?

For some reason I do not see this failure on TPU.

Issue #2347, opened 2020-03-04
jnp.array(x, dtype=object) Causes Error(Shape Checking Issue)


Yeachan-Heo:
x = List[Tuple[jnp.array, float, float, jnp.array, bool]]
jnp.array(x, dtype=object) does not work. when np.array(x, dtype=object) works. 
i expect result as jnp.ndarray[Tuple[jnp.array, float, float, jnp.array, bool]]
Following Error is:
RuntimeError: Invalid argument: Convert does not allow non-arrays, so cannot convert from s32[4] to TOKEN.: 
This is a bug in JAX's shape-checking rules; please report it!

mattjj:
It's intentional that JAX doesn't support dtype=object arrays, but we could improve this error message and the documentation about it.

Yeachan-Heo:
> It's intentional that JAX doesn't support dtype=object arrays, but we could improve this error message and the documentation about it.

Looking Forward to it! Thanks XD

tudorcebere:
Hello! Is this still open?

jakevdp:
Hi @tudorcebere, thanks for your interest! I took a look and it seems like the error is better now:
```python
>>> import jax.numpy as jnp
>>> jnp.array([(1, 2), 2, 3], dtype=object)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-7af97dc5d062> in <module>()
      1 import jax.numpy as jnp
----> 2 jnp.array([(1, 2), 2, 3], dtype=object)

/usr/local/lib/python3.6/dist-packages/jax/numpy/lax_numpy.py in array(object, dtype, copy, order, ndmin)
   2234   if order is not None and order != "K":
   2235     raise NotImplementedError("Only implemented for order='K'")
-> 2236   lax._check_user_dtype_supported(dtype, "array")
   2237   dtype = dtype and dtypes.canonicalize_dtype(dtype)
   2238 

/usr/local/lib/python3.6/dist-packages/jax/lax/lax.py in _check_user_dtype_supported(dtype, fun_name)
   5894   if np_dtype.kind not in "biufc" and np_dtype.type != dtypes.bfloat16:
   5895     msg = f"JAX only supports number and bool dtypes, got dtype {dtype}"
-> 5896     raise TypeError(msg)
   5897   if dtype is not None and np_dtype != dtypes.canonicalize_dtype(dtype):
   5898     msg = ("Explicitly requested dtype {} {} is not available, "

[...]

PR #4323, opened 2020-09-17
Solving #2347

tudorcebere:
Covering an execution path on array creation: when an array is created with an unspecified type (None), there was no type checking that the inferred type is compatible with JAX. The issue has been documented and a test with the use case added.

This issue solves #2347

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2733,6 +2733,8 @@ def atleast_3d(*arys):
 def array(object, dtype=None, copy=True, order="K", ndmin=0):
   if order is not None and order != "K":
     raise NotImplementedError("Only implemented for order='K'")
+
+  # check if the given dtype is compatible with JAX
   lax._check_user_dtype_supported(dtype, "array")
 
   weak_type = dtype is None and dtypes.is_weakly_typed(object)
@@ -2740,9 +2742,12 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
 
   if _can_call_numpy_array(object):
     object = _np_array(object, dtype=dtype, ndmin=ndmin, copy=False)
+
   assert type(object) not in dtypes.python_scalar_dtypes
 
   if type(object) is np.ndarray:
+    _inferred_dtype = object.dtype and dtypes.canonicalize_dtype(object.dtype)
+    lax._check_user_dtype_supported(_inferred_dtype, "array")
     out = _device_put_raw(object, weak_type=weak_type)
     if dtype: assert _dtype(out) == dtype
   elif isinstance(object, (DeviceArray, core.Tracer)):
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -4664,6 +4664,13 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     with self.assertRaisesRegex(jax.core.ConcretizationTypeError, msg('stop')):
       jax.jit(lambda stop: jnp.arange(0, stop))(3)
 
+  def testIssue2347(self):
+    # https://github.com/google/jax/issues/2347
+    object_list = List[Tuple[jnp.array, float, float, jnp.array, bool]]
+    self.assertRaises(TypeError, jnp.array, object_list)
+
+    np_object_list = np.array(object_list)
+    self.assertRaises(TypeError, jnp.array, np_object_list)
 
 # Most grad tests are at the lax level (see lax_test.py), but we add some here
 # as needed for e.g. particular compound ops of interest.


tudorcebere:
@googlebot  I signed it!

jax/numpy/lax_numpy.py
```diff
@@ -2277,11 +2277,16 @@ def atleast_3d(*arys):
...
+  # check if the given dtype is compatible with JAX
   lax._check_user_dtype_supported(dtype, "array")
   dtype = dtype and dtypes.canonicalize_dtype(dtype)
 
   if _can_call_numpy_array(object):
     object = _np_array(object, dtype=dtype, ndmin=ndmin)
+    # check if the inferred type is compatible with JAX
+    _inferred_dtype = object.dtype and dtypes.canonicalize_dtype(object.dtype)
+    lax._check_user_dtype_supported(_inferred_dtype, "array")
+
```
jakevdp, line 2289: I would suggest putting this check under `if type(object) is np.ndarray`, as that is the place where the error can crop up.

I'm fairly certain that the choice to put it here is fine, because the only way you can end up in that lower block is if you are first in this block, but it would require less reasoning about the control flow if the check is put below.

Does that sound right to you?

jakevdp (commented):
Thanks for addressing this! One comment below:

jax/numpy/lax_numpy.py
tudorcebere: Sounds good! Made the requested change and rebased.

jax/numpy/lax_numpy.py
```diff
@@ -2277,11 +2277,18 @@ def atleast_3d(*arys):
...
   dtype = dtype and dtypes.canonicalize_dtype(dtype)
 
   if _can_call_numpy_array(object):
     object = _np_array(object, dtype=dtype, ndmin=ndmin)
+
+  if type(object) is np.ndarray:
+    # check if the inferred np type is compatible with JAX
+    _inferred_dtype = object.dtype and dtypes.canonicalize_dtype(object.dtype)
+    lax._check_user_dtype_supported(_inferred_dtype, "array")
+
```
jakevdp, line 2291: I think  the check should be put in the existing `if type(object) is np.ndarray` block just below this.

tudorcebere: Actually, the problem appears when a dtype is inferred by numpy at line 2285 (on the new object). I believe that the check should be after that point (that's the path in the current codebase that is not covered - a numpy array with the dtype=object).

jakevdp: Understood - but another path to this is if the input is a numpy array with dtype `object`, and it was not clear to me without looking at the source of `_can_call_numpy_array` whether that block would be executed in that case. It turns out it is.  But due to that obfuscation I thought it would be clearer to put the check in the place where it's obvious at first glance that it will be called for *every* numpy array input.

What do you think?

tudorcebere: I understand your point, nice catch, thanks. The check flow now should be: check current object dtype -> check specified dtype -> check inferred dtype. I don't think we can move it to the top, because in the case when the dtype is `None` and we can't infer the type from the object (like in the test case), we gen the dtype after the numpy array is being created and we check that. You are right that a check should be at the top as well in the case when a numpy array is being passed with a bad dtype to not propagate the error until the constructor of numpy array is called.

What do you think?

jakevdp: The new version seems redundant: if the input is a numpy array, the typecheck now happens twice. What I had in mind is this:
```python
def array(object, dtype=None, copy=True, order="K", ndmin=0):
  if order is not None and order != "K":
    raise NotImplementedError("Only implemented for order='K'")

  # check if the given dtype is compatible with JAX
  lax._check_user_dtype_supported(dtype, "array")
  dtype = dtype and dtypes.canonicalize_dtype(dtype)

  if _can_call_numpy_array(object):
    object = _np_array(object, dtype=dtype, ndmin=ndmin)

  assert type(object) not in dtypes.python_scalar_dtypes

  if type(object) is np.ndarray:
    _inferred_dtype = object.dtype and dtypes.canonicalize_dtype(object.dtype)
    lax._check_user_dtype_supported(_inferred_dtype, "array")
    out = _device_put_raw(object)
    if dtype: assert _dtype(out) == dtype
  elif isinstance(object, (DeviceArray, core.Tracer)):
    #...
```
a single check that takes care of every case where `object` is an array.

tudorcebere: Sorry for being painfully stupid, I did not observe that the next block checked exactly for that, now it totally makes sense.

jakevdp: No worries!

jakevdp (approved):
Looks great! Once tests are green, I'll merge. Thanks so much for the contribution!

jakevdp:
This somehow fell off my radar, and now it needs a rebase to be merged - sorry about that! Are you able to rebase this on current master?

tudorcebere:
@jakevdp rebased, sorry for the delay.

jakevdp:
Sorry, you caught me during my week off and now there are conflicts. Can you rebase again and resolve the conflicts? Thanks!

PR #4328, opened 2020-09-17
rename and simplify TypedJaxpr -> ClosedJaxpr

mattjj:
rename and simplify TypedJaxpr -> ClosedJaxpr

This change:
* simplifies code that constructs TypedJaxprs/ClosedJaxprs (because in_avals/out_avals no longer need to be constructed), making them easier to work with;
* correspondingly rules out a class of errors (mismatches between invars/outvars and in_avals/out_avals);
* provides a more descriptive class name (ClosedJaxprs are like jaxprs but they're closed in that they are packaged with their constant values).

This is part 1 of an attempt to remove TypedJaxprs completely, or at least significantly reduce our use of them. However, I'm not getting rid of them entirely in this first step because it'd require bigger changes (basically allowing all constants to be represented as literals, rather than only scalars) that would not only touch a lot more code (jaxpr formation, jaxpr-to-jaxpr transformations, control flow, XLA lowering) but also might affect XLA lowering right before a  conference deadline (ICLR). Plus I'm trying to make big changes in smaller steps :)

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -67,15 +67,14 @@ class Jaxpr:
   def __init__(self, constvars: Sequence['Var'], invars: Sequence['Var'],
                outvars: Sequence['Atom'], eqns: Sequence['JaxprEqn']):
     """
-    Params:
-      constvars: list of variables introduced for constants (either literals
-        in the Python program, or the result of constant folding during the
-        generation of the Jaxpr). Array constants are replaced with such variables
-        while scalar constants are kept inline.
+    Args:
+      constvars: list of variables introduced for constants. Array constants are
+        replaced with such variables while scalar constants are kept inline.
       invars: list of input variables. Together, `constvars` and `invars` are
         the inputs to the Jaxpr.
       outvars: list of output variables.
-      eqns: list of equations."""
+      eqns: list of equations.
+    """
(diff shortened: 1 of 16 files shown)


mattjj:
@gnecula @froystig No need to review every line. Maybe just take a look at the docs, and think about the idea behind the change.

jax/core.py
```diff
@@ -92,54 +91,46 @@ def jaxprs_in_params(params) -> Iterator[Jaxpr]:
...
-               in_avals: Sequence['AbstractValue'],
-               out_avals: Sequence['AbstractValue']):
-    assert len(literals) == len(jaxpr.constvars)
-    assert len(in_avals) == len(jaxpr.invars)
+  def __init__(self, jaxpr: Jaxpr, consts: Sequence):
+    assert len(consts) == len(jaxpr.constvars)
+    self.jaxpr = jaxpr
+    self.consts = list(consts)
 
-    if not skip_checks:
```
gnecula, line 119: Being able to remove this invariant check is a powerful reason to make the change.

gnecula (approved):
Nice!

mattjj:
Global internal tests pass!

PR #4330, opened 2020-09-18
insert a stop_gradient in lax.create_token

froystig:
see #4292 

The data dependency forced on the operand is false. There's no need to follow it for derivatives, and doing so can lead to unexpected errors.


mattjj:
LGTM but perhaps we should just make this omnistaging-only and remove the dummy arg entirely. That can be done in follow-up stuff.

mattjj (approved):
LGTM, see comment in thread about omnistaging.

PR #4331, opened 2020-09-18
Use jvp(expm) to compute expm_frechet.

zhangqiaorjc:


shoyer (approved):
It would be great to see some tests with larger numbers (rather than rand_small), just to be sure we hit all the code paths for expm

zhangqiaorjc:
Added test matrices with a range of target l1 norm values to trigger different code paths. I had to bump tol for some dtypes to pass 2nd order grad. Left a TODO to improve tol for those ExpmGrad tests.

shoyer (approved):
Looks great to me! Thanks for adding these careful test cases.

PR #4334, opened 2020-09-18
[jax2tf] Add support for Add for uint32

gnecula:


gnecula:
I will have to try #4278 again, I had to revert due to bugs in TF.

PR #4338, opened 2020-09-18
[jax2tf] Implementation of the conversion of eig_p.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1332,6 +1332,28 @@ def _svd(operand, full_matrices, compute_uv):
 
 tf_impl[lax_linalg.svd_p] = _svd
 
+def _eig(operand: TfVal, compute_left_eigenvectors: bool,
+         compute_right_eigenvectors: bool):
+  if compute_left_eigenvectors and compute_right_eigenvectors:
+    # TODO(bchetioui): didn't find a 100% reliable, easy and satisfying way to
```
gnecula, line 1338: It is not clear to me what sorting is relevant here. Presumably it has something to do with what tf.linalg.eigvals and tf.linalg.eig do?

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -109,6 +109,14 @@ def _to_np_dtype(dtype) -> NpDType:
       # experimental_compile=True breaks for complex types.
       tf_unimpl(np_dtype)
 
+  if prim is lax_linalg.eig_p:
+    tf_unimpl(additional_msg=("this is a problem only in compiled mode "
```
gnecula, line 113: Does this mean that in compiled mode `eig` will not work? Does JAX have custom calls for CPU and GPU?

SIben: Yes, the implementations are provided in `jaxlib.lapack.pyx` (`*geev` functions).

jax/experimental/jax2tf/jax2tf.py
SIben: Given the triplet `(w, vl, vr)` of eigenvalues, left eigenvectors, and right eigenvectors, we expect that `vl[i]` is a left eigenvector with eigenvalue `w[i]`, and the same on the right for `vr[i]`, so the order needs to be consistent across these three.

gnecula: I understand, I suggest adding the clarification for the sorting in the comment itself.

jax/experimental/jax2tf/tests/correctness_stats.py
gnecula: I suggested adding a comment, or perhaps even text in the message saying that XLA does not support this case.

SIben: I should have said implementation* (this is actually only on CPU). I will add it.

PR #4346, opened 2020-09-18
Disable CPP Jax jit tests.

jblespiau:
it's because I will do a breaking change in jax_jit. As I need 2 Cls (one on google3 for TF, one on github for JAX), and that they are incompatible, I must disable this (otherwise google3/ tests will fail).

We do need to move jax_jit and co to JAX.

The follow-up will be cr/332509389.


tests/api_test.py
```diff
@@ -63,7 +63,7 @@ def jit(self):
     # TODO(jblespiau,phawkins): Remove this when jaxlib has been released.
     # This is in the future, because we are making a breaking change to
     # Tensorflow.
-    if version < (0, 1, 54):
```
hawkinsp, line 66: Nit: shouldn't we just skip the test? Won't `PythonJitTest` also run the Python tests?

jblespiau: it's not just the test we need to skip, it's all the tests of this class.

It's a base class used by PythonJitTest, so if disable tests, it will also be disabled there.

By doing this, it will run CPPJitTest, but as the PythonTests

hawkinsp: Doesn't PythonJitTest override this method?

hawkinsp (approved):

jekbradbury:
This went in as https://github.com/google/jax/commit/9f53d2a8d883bc6eefc1f701cdc6639bdc774f84; seems to have been a fluke in the new copybara flow.

PR #4349, opened 2020-09-18
Fixed ppermute batching rule

### LINKED ISSUE #4348: Ppermute batching errors
This code.

```python
import jax
import jax.numpy as jnp
from functools import partial

@partial(jax.pmap, axis_name="i")
@jax.vmap
def f(a):
  jax.lax.ppermute(a, "i", [0, 1, 2, 3, 4, 5, 6, 7, 8])

# Run it.
f(jnp.arange(8 * 8 * 8).reshape((8, 8, 8)))
```

Dies with this
```
TypeError: _allreduce_translation_rule() got an unexpected keyword argument 'perm'
```

I think this is because of this line
https://github.com/google/jax/blob/40e20242db0ed1d8cceb1d93b78d07c248e693a6/jax/lax/lax_parallel.py#L550

For some reason, the batching primative was set as `pmin`. 

I'll go ahead and make a PR to fix this.

chaserileyroberts:
Removed incorrect binding to pmin

Fixes #4348


mattjj:
Thanks for looking at this! (This is a batching rule, not a translation rule as in the issue title.)

PR #4350, opened 2020-09-18
Add jax.numpy.array_equiv

ajcr:
Adds jax implementation of `np.array_equiv`.


ajcr:
Ah - I see this was done 20 minutes ago! https://github.com/google/jax/pull/4347

PR #4351, opened 2020-09-18
Return a function in _cpp_jit as jax.jit must return a function.

jblespiau:


jax/traceback_util.py
```diff
@@ -19,6 +19,7 @@
 
 from .api_util import wraps
 
+
 _jax_path = os.path.dirname(__file__)
```
jblespiau, line 23: I did not modify this in my CitC, so it must be a difference between google3/ HEAD and github.

I would prefer not to have to change this, it would involve syncing again, hoping this to be solved, exporting again etc.
I do not think a new line is worth 8 minutes of my time (for a google3/ CL, it would be easy, but because of github first, I have less flexibility (I use a script exporting from critique to here).

Issue #70, opened 2018-12-11
Unimplemented NumPy core functions


alexbw:
Remaining functions to be implemented:

- [x] np.argpartition (tricky because XLA does not provide any partial sort functionality) 
- [x] np.partition (tricky because XLA does not provide any partial sort functionality)

The list above was made by inspecting `jnp._NOT_IMPLEMENTED` and excluding deprecated functions (such as `np.alen`, `np.ipmt`, etc.), functions not relevant to JAX (such as `np.setbufsize`, `np.ascontiguousarray`, etc), and functions that modify buffers in-place (`np.put`, `np.place`, etc.):

Bugs for high-level categories:
 - [x] np.linalg: https://github.com/google/jax/issues/1999
 - [x] np.fft: https://github.com/google/jax/issues/1877
 - [x] searching: https://github.com/google/jax/issues/2080
 - [x] sorting: https://github.com/google/jax/issues/2079
 - [x] sets: https://github.com/google/jax/issues/2078
 - [x] NaN functions: https://github.com/google/jax/issues/2077
 - [x] Index tricks: https://github.com/google/jax/issues/5850

mattjj:
The list needs a bit of refinement, since some things were already implemented and other things are out of scope, like to do with loading/saving/string manipulation. I deleted a few of the out-of-scope ones, and marked as done some things that were already included and some things that are coming in #96.

rainwoodman:
Is there a guide on how to write customized primitive functions (or is that even possible?). I am looking for the equivalent to autograd's `defvjp`.

A quick glance into the code suggests everything is delegated to XLA -- if an operator is nontrivial then it has to be first implemented in XLA?

kovasb:
Looking at contributing a function or two from this list. 

What is the approach to functions with nonstatically-sized return values? For example, setxor1d. 

My understanding is XLA needs to compile for each specific shape. Is there a pattern for avoiding worse-possible-case behavior, or should functions like setxor1d be removed from this list?

souravsingh:
@mattjj I am looking to take a stab at the issue. How do I start?

hawkinsp:
@souravsingh Awesome! In general, lots of these are fairly easy to do. Pick one you like off the list, and take a look at previous PRs implementing numpy ops. For example, I just sent out https://github.com/google/jax/pull/298 adding np.cumprod/cumsum support. Usually, it's just a question of implementing a numpy op in terms of the primitives in the `lax.py` library.

[...]

PR #4354, opened 2020-09-19
ravel_multi_index implementation

alexminnaar:
Hello, I wanted to implement numpy's ravel_multi_index however I just noticed [here](https://github.com/google/jax/issues/70) that it was marked as implemented but I don't actually see it anywhere.  If it actually has been implemented and I just missed it I can go ahead and close this.


alexminnaar:
Looks like it got implemented 2 days ago

PR #4357, opened 2020-09-20
Added all to all batching support

### LINKED ISSUE #4344: NotImplementedError: Batching rule for 'all_to_all' not implemented
I would like this to be supported.

chaserileyroberts:
Just opening as a draft as I'm not confident I did this correctly.

Fixes #4344

I realized that #4170 is also open that does the same thing.


hawkinsp:
It looks like a subsequent PR did this!

PR #4359, opened 2020-09-21
Add back the batching rule for ppermute

apaszke:
Just make sure it's correct this time and add a test.


mattjj (approved):
LGTM!

It might be good (maybe in a separate follow-up) to have a test for `_collective_batcher` that tries other in_axes/out_axes (though that test may already exist for some other collective, and also I know it's super simple and seems impossible to break!).

apaszke:
Included the test in another commit!

PR #4360, opened 2020-09-21
[jax2tf] Added support for shape polymorphism conversion.

gnecula:
The intended usage, and some design challenges, for this feature
are described in README.md.

In order to enable this change, I had to first change
TensorFlowTracer to carry explicitly the JAX abstract value
for the undelying TF value (previously, we would just compute
the abstract value from the TF value. This is not possible
when the TF value has partially known shape. In that case we
want the JAX abstract value, using masking.ShapeSpec.

As a beneficial side-effect of adding explicit abstract values
to TensorFlowTracer we can clean up all the hacky handling of
core.unit (we would store core.unit as a TF value, hence the
need for TfValOrUnit, and we would swap it with tf.nan
when going to TF). Now we can just store tf.name as the TF value
and core.abstract_unit for the abstract value.

The key smarts in this change are just reused from jax.interpreters.masking.
All we really added is carefully carrying that information through.

A tricky part was carrying the abstract shapes for tf.custom_gradient.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -13,13 +13,14 @@
 # limitations under the License.
 """Experimental module transforms JAX functions to be executed by TensorFlow."""
 import functools
+import itertools
 import string
-from typing import Any, Callable, Dict, Iterable, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Optional, Sequence, Tuple, Union
 
 import jax
-from jax import abstract_arrays
 from jax import ad_util
 from jax import api
+from jax import api_util
 from jax import config
 from jax import core
 from jax import custom_derivatives
@@ -33,11 +34,13 @@ from jax import tree_util
 from jax import util
 from jax.api_util import flatten_fun
 from jax.interpreters import ad
+from jax.interpreters import masking
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import pxla
 from jax.interpreters import xla
 from jax.lax import lax_control_flow
 from jax.lax import lax_fft
+
 import numpy as np
 import tensorflow as tf  # type: ignore[import]
 
@@ -52,18 +55,20 @@ from jaxlib import xla_client
 # A value suitable in a TF tracing context: tf.Tensor, tf.Variable,
 # or Python scalar or numpy.ndarray. (A tf.EagerTensor is a tf.Tensor.)
 TfVal = Any
+DType = Any
 def _is_tfval(v: TfVal) -> bool:
   if isinstance(v, (tf.Tensor, tf.Variable)):
     return True
   try:
     # Note: this conversion is overkill and just intended as a type check; this code
     # is in principle only run if core.skip_checks is False.
+    # TODO: it is not true that this code is run only without skip_checks
     _safe_convert_to_tensor(v)
     return True
   except ValueError:
     return False
 
-def _safe_convert_to_tensor(val, dtype=None):
+def _safe_convert_to_tensor(val, dtype=None) -> TfVal:
   """Converts val to a Tensor.
 
   This method wraps TensorFlow's `convert_to_tensor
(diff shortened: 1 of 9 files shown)


jax/interpreters/masking.py
```diff
@@ -326,7 +330,7 @@ def _parse_dim(spec):
     return prod(map(_parse_dim, spec.split('*')))
   elif spec.isdigit() or spec.startswith('-') and spec[1:].isdigit():
     return _parse_lit(spec)
-  elif spec in _identifiers:
+  elif spec[0] in _identifiers:
```
gnecula, line 333: I am not sure I understand the original intent behind this code. It seems to work for single-letter identifiers. I thought that it makes sense to allow multi-letter identifiers that start with lowercase.

jax/interpreters/masking.py
```diff
@@ -311,13 +311,17 @@ def finalize_spec(polymorphic_shape, padded_shape):
...
 
 def parse_spec(spec=''):
-  if not spec:
-    return ShapeSpec(())
-  if spec[0] == '(':
-    if spec[-1] != ')': raise ShapeSyntaxError(spec)
-    spec = spec[1:-1]
-  dims = map(_parse_dim, spec.replace(' ', '').strip(',').split(','))
-  return ShapeSpec(dims)
+  try:
```
gnecula, line 314: There are two changes here, the string "()" was not processed properly, and I wanted an error message that includes the original in_shapes not just the part that failed. 

I guess it makes sense to add some masking_test.py, but I don't know what is the status of that file.

gnecula: I have reverted this change

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -410,6 +415,235 @@ def outer(y):
...
+    self.assertAllClose(res_jax, res_tf)
+    self.assertAllClose(res_jax_grad, res_tf_grad)
+
+    # Now use TF tracing for the gradient
+    tf_grad = tf.function(tf_value_and_grad, autograph=False).get_concrete_function(*self.MakeInputSignature("(batch1, batch2, 8, 9)"))
+    # The shape of the value
+    self.assertEqual((None, None, 8, 8), tuple(tf_grad.output_shapes[0]))
+    # The shape of the gradient should match the input
+    # TODO: there seems to be a bug here, the output should be (None, None, 8, 9)
+    self.assertEqual((None, None, 8, None), tuple(tf_grad.output_shapes[1]))
```
bchetioui, line 563: Should we really keep this assertEqual if we believe this is a bug?

bchetioui (approved):
A few minor issues, otherwise LGTM :)

jax/experimental/jax2tf/tests/jax2tf_test.py
gnecula: Good point, I removed.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -58,12 +59,13 @@ def _is_tfval(v: TfVal) -> bool:
   try:
     # Note: this conversion is overkill and just intended as a type check; this code
     # is in principle only run if core.skip_checks is False.
+    # TODO: it is not true that this code is run only without skip_checks
```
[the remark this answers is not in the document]
gnecula: Yes, I worry that the performance penalty may be significant

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -410,6 +415,238 @@ def outer(y):
...
+    """Shape polymorphism with gradients and pytrees for inputs and outputs."""
+    def f(x):
+      # x: dict(x=[b, 3, 4])
+      # res: dict(res=[b, 3, 4])
+      return dict(res=x["x"] * 2.)
+
+    f_tf = self.CheckShapePolymorphism(f, [dict(x=("b, 3, 4"))],
+                                       expected_output_shape=dict(res="(b, 3, 4)"))
+    x = dict(x=np.ones((2, 3, 4), dtype=np.float32))
+    xv = tf.Variable(x["x"], dtype=dtypes.canonicalize_dtype(jnp.float_))
```
bchetioui, line 579: The dtype here should instead be `jnp.float32`. As it is, the test is not compatible with the x64 flag
```suggestion
    xv = tf.Variable(x["x"], dtype=jnp.float32)
```

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -410,6 +415,238 @@ def outer(y):
...
+    f.defvjp(f_fwd, f_bwd)
+    x = np.ones((2, 3, 4, 5), dtype=np.float32)
+    res_jax = f(x)
+    res_jax_grad = jax.grad(lambda x: jnp.sum(f(x)))(x)
+
+    f_tf = self.CheckShapePolymorphism(f, ["(batch1, batch2, d1, d2)"],
+                                       expected_output_shape="(batch1, batch2, d1, d1)")
+    self.assertAllClose(res_jax, f_tf(x))
+
+    xv = tf.Variable(x, dtype=dtypes.canonicalize_dtype(jnp.float_))
```
bchetioui, line 545: As it is, it will fail in x64 mode (same as below). The line should be:
```suggestion
    xv = tf.Variable(x, dtype=jnp.float32)
```

bchetioui (changes requested):
Ran the tests in x64 mode and noticed two more issues I overlooked before :)

jax/experimental/jax2tf/tests/jax2tf_test.py
gnecula: Fixed

jax/experimental/jax2tf/tests/jax2tf_test.py
gnecula: Fixed

bchetioui (approved):

PR #4361, opened 2020-09-21
[jax2tf] Implementation of the conversion of eigh_p.

bchetioui:
The main issue is a huge loss of precision in compiled mode.

I believe (but couldn't test) that `always_custom_assert` can be set to False on TPU, since the implementation for that platform does not seem to use custom calls.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -420,7 +420,6 @@ tf_not_yet_impl = [
   lax.reduce_p, lax.rng_uniform_p,
 
   lax.linear_solve_p,
-  lax_linalg.eigh_p,
   lax_linalg.lu_p,
   lax_linalg.triangular_solve_p,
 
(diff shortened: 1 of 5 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1350,6 +1349,20 @@ def _eig(operand: TfVal, compute_left_eigenvectors: bool,
...
 tf_impl[lax_linalg.eig_p] = _eig
 
+def _eigh(operand: TfVal, lower: bool):
+  if operand.shape[-1] == 0:
+    return tuple([operand, tf.reshape(operand, operand.shape[:-1])])
+  if not lower:
+    operand = tf.linalg.adjoint(operand)
+  w, v = tf.linalg.eigh(operand)
+  dtype = operand.dtype
+  if dtype in [tf.complex64, tf.complex128]:
```
gnecula, line 1359: Nit: this could be written as:
```
cast_type = { tf.complex64: tf.float32, tf.complex128: tf.float64 }.get(dtype)
if cast_type is not None:
  w = tf.cast(w, cast_type)
```

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -337,6 +337,68 @@ def check_eigenvalue_is_in_array(eigenvalue, eigenvalues_array):
...
+  def test_eigh(self, harness: primitive_harness.Harness):
+    operand = harness.dyn_args_maker(self.rng())[0]
+    lower = harness.params["lower"]
+    # Make operand self-adjoint
+    operand = (operand + np.conj(np.swapaxes(operand, -1, -2))) / 2
+    # Make operand lower/upper triangular
+    triangular_operand = np.tril(operand) if lower else np.triu(operand)
+    dtype = harness.params["dtype"]
+
+    if dtype in [np.float16]:
```
gnecula, line 350: I think that we should start to put these in the primitive_harnesses, e.g., filter these cases out.

gnecula (approved):

PR #4363, opened 2020-09-21
Add a prototype implementation of recursive checkpointing

apaszke:
Example:
```py
import jax                                               
import jax.numpy as jnp                                  
from jax.experimental.checkpoint import checkpoint_recursive
                                                         
@checkpoint_recursive                                     
def f(x):                                                
  for i in range(8):                                     
    x = jnp.exp(x)                                       
  return x                                               
                                                         
print(jax.make_jaxpr(f)(jnp.ones((4,))))                 
```

produces the following jaxpr:

<details>

```
{ lambda  ; a.
  let b = remat_call[ call_jaxpr={ lambda  ; a.
                                   let b = remat_call[ call_jaxpr={ lambda  ; a.
                                                                    let b = remat_call[ call_jaxpr={ lambda  ; a.
                                                                                                     let b = exp a
                                                                                                     in (b,) }
                                                                                        concrete=False
                                                                                        name=<unnamed wrapped function> ] a
                                                                        c = exp b
                                                                    in (c,) }
                                                       concrete=False
                                                       name=<lambda> ] a
                                       c = remat_call[ call_jaxpr={ lambda  ; a.
                                                                    let b = exp a
                                                                    in (b,) }
                                                       concrete=False
                                                       name=<unnamed wrapped function> ] b
                                       d = exp c
                                   in (d,) }
                      concrete=False
                      name=<lambda> ] a
      c = remat_call[ call_jaxpr={ lambda  ; a.
                                   let b = remat_call[ call_jaxpr={ lambda  ; a.
                                                                    let b = exp a
                                                                    in (b,) }
                                                       concrete=False
                                                       name=<unnamed wrapped function> ] a
                                       c = exp b
                                   in (c,) }
                      concrete=False
                      name=<lambda> ] b
      d = remat_call[ call_jaxpr={ lambda  ; a.
                                   let b = exp a
                                   in (b,) }
                      concrete=False
                      name=<unnamed wrapped function> ] c
      e = exp d
  in (e,) }

```

</details>

**TODO:**
* [ ] Tests
* [ ] Better cost estimation (loops, calls, conditionals all can have an arbitrary number of primitives inside)
* [ ] Loop splitting, when the half-point of operation count lies inside a loop

### DIFF
--- a/jax/experimental/checkpoint.py
+++ b/jax/experimental/checkpoint.py
@@ -1,0 +1,50 @@
+from typing import Callable
+from functools import partial
+import itertools as it
+
+from .. import checkpoint
+from .. import core
+from .. import linear_util as lu
+from ..interpreters import partial_eval as pe
+from ..api_util import wraps, flatten_fun
+from ..tree_util import tree_flatten, tree_unflatten
+from ..traceback_util import api_boundary
+
+# TODO: Support concrete?
+def checkpoint_recursive(fun: Callable, threshold=1) -> Callable:
+  if threshold < 1:
+    raise ValueError(f"threshold has to be at least 1, got {threshold}")
+  @wraps(fun)
+  @api_boundary
+  def fun_recursive(*args, **kwargs):
+    args_flat, in_tree = tree_flatten((args, kwargs))
+    flat_fun, out_tree = flatten_fun(lu.wrap_init(fun), in_tree)
+    avals_flat = [core.raise_to_shaped(core.get_aval(arg)) for arg in args_flat]
+    jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(flat_fun, avals_flat)
+    jaxpr_noconst = pe.convert_constvars_jaxpr(jaxpr)
+    checkp_fun = _recursive_checkpoint_jaxpr(jaxpr_noconst, threshold=threshold)
+    out_flat = checkp_fun(*consts, *args_flat)
+    return tree_unflatten(out_tree(), out_flat)
+  return fun_recursive
+
+def _recursive_checkpoint_jaxpr(jaxpr, threshold):
+  assert not jaxpr.constvars
+  eqns = jaxpr.eqns
+  if len(jaxpr.eqns) <= threshold:
+    return partial(core.eval_jaxpr, jaxpr, ())
+  split_point = (len(eqns) + 1) // 2
+  eqns1, eqns2 = eqns[:split_point], eqns[split_point:]
+  residuals = list(_jaxpr_free_vars(core.Jaxpr([], [], jaxpr.outvars, eqns2)))
+  jaxpr1 = core.Jaxpr([], jaxpr.invars, residuals, eqns1)
+  jaxpr2 = core.Jaxpr([], residuals, jaxpr.outvars, eqns2)
+  f1 = checkpoint(_recursive_checkpoint_jaxpr(jaxpr1, threshold))
+  f2 = _recursive_checkpoint_jaxpr(jaxpr2, threshold)
+  return lambda *args: f2(*f1(*args))
+
+def _jaxpr_free_vars(jaxpr):
+  free = set(jaxpr.outvars)
+  for eqn in jaxpr.eqns[::-1]:
+    free -= set(eqn.outvars)
+    free |= set(eqn.invars)
+  free -= set(it.chain(jaxpr.invars, jaxpr.constvars))
+  return free
--- a/tests/checkpoint_test.py
+++ b/tests/checkpoint_test.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import numpy as np
 from absl.testing import absltest
+from unittest import skipIf
 
 import jax
 from jax import core
@@ -26,12 +28,13 @@ config.parse_flags_with_absl()
 FLAGS = flags.FLAGS
 
 class CheckpointTest(jtu.JaxTestCase):
+  @skipIf(not jax.config.omnistaging_enabled, "need omnistaging")
   def testSimpleChain(self):
     def f(x):
       for i in range(8):
         x = jnp.exp(x)
       return x
-    x = jnp.ones((4,))
+    x = np.ones((4,))
     jaxpr = jax.make_jaxpr(C.checkpoint_recursive(f))(x).jaxpr
     def verify(jaxpr, lvl):
       if lvl < 0: return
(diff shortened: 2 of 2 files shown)


shoyer:
Nit: do you plan to actually implement [binomial checkpointing](https://opus4.kobv.de/opus4-matheon/files/76/86.pdf), or just a form of recursive "multilevel checkpointing"? Binomial checkpointing is definitely more efficient given a fixed memory budget, but may not be worth the trouble...

apaszke:
Yeah @mattjj has pointed out yesterday that this is not really binomial checkpointing. I'll change the name today. I would be interested in implementing the binomial scheme as well, but honestly I can't find a single paper that has a reasonable description of how this is done. Everything I find either uses the `revolve` procedure as a black box, or only discusses the lower bounds... Do you have any good references?

sriharikrishna:
@apaszke Would you benefit from having a python implementation of Revolve? @jhueckelheim and others have spent time on this before and would be interested in contributing.  There exists an implementation by Julien Hermann associated with [https://arxiv.org/pdf/1911.13214v1.pdf](https://arxiv.org/pdf/1911.13214v1.pdf) that might be straightforward to adapt. Unfortunately, the software artifact does not seem available.

shoyer:
> There exists an implementation by Julien Hermann associated with https://arxiv.org/pdf/1911.13214v1.pdf that might be straightforward to adapt

This is definitely a nice description of the binomial checkpointing algorithm, at the very least! Thanks for the pointer.

From an implementation perspective, it seems like the first step would be to write the general code that implements an arbitrary "checkpointing strategy" in JAX's autodiff. Then we could explore swapping in different strategies, e.g., binomial vs binary checkpointing.

Gaupy:
> @apaszke Would you benefit from having a python implementation of Revolve? @jhueckelheim and others have spent time on this before and would be interested in contributing. There exists an implementation by Julien Hermann associated with https://arxiv.org/pdf/1911.13214v1.pdf that might be straightforward to adapt. Unfortunately, the software artifact does not seem available.

I believe the implementation that @sriharikrishna was refering to is that one: 
https://gitlab.inria.fr/adjoint-computation/disk-revolve-public

The details can be found in the attached paper: https://dl.acm.org/doi/10.1145/3378672

apaszke:
Great reference @sriharikrishna, thanks for posting that! That's a very understandable formulation and I think we could implement something like that relatively easily. I think that memory persistent schedules are also the one we could desugar into some nesting of remats, so that's a good property to have.

Agreed with @shoyer that it would be great to write a toolbox for that and I'm happy to drive that. The code that's in this PR is a starting point which allows you to suspend the program after a fixed set of iterations, following https://openreview.net/pdf?id=BkYYXJ9i-, and we already have the single-level checkpointing implemented as `jax.checkpoint`. What else do we need?

I guess one thing that would be nice to have is an interpreter for the imperative code generated by most checkpointing tools. But I wonder if it wouldn't be better to just reimplement them in less stateful way...

@Gaupy thanks for the links. Unfortunately we won't be able to integrate that directly into JAX, because it's GPL licensed 😕 Still, the interpreter I mentioned above might help with an integration outside of the core if you'd be interested in that.

Gaupy:
Ok I understand. What kind of licence would you need? Changing the licence is certainly not a problem (we did not give too much thoughts about the licence and I don't expect Julien to care either about  moving to a LGPL type of licence).

apaszke:
I think that Apache 2.0 or BSD are generally acceptable, although I'd have to check. @shoyer might know more.

Still, it looks like those programs are generally designed to output a sequence of steps and I think that for JAX purposes it would be better to get some kind of a tree representation that shows how the individual checkpoints are performed, so we'll probably need to rewrite it quite a bit anyway.

shoyer:
> Ok I understand. What kind of licence would you need? Changing the licence is certainly not a problem (we did not give too much thoughts about the licence and I don't expect Julien to care either about moving to a LGPL type of licence).

We could distribute Apache, BSD or MIT licensed code with JAX.

We could depend on LGPL licensed code. I'm not entirely sure about GPL, but we definitely can't use AGPL (the current license) in any form: https://opensource.google/docs/using/agpl-policy/

All that said, I guess the most valuable component of this code/paper for us would be just having a reference for understanding exactly how the "revolve" algorithm works. Eventually we probably want a version of that in JAX.

Gaupy:
Not sure this is the best location to describe the algorithm, but to understand the code there are two ways:
 - the first (simplest way but more computational intensive) is to compute the full schedule ahead of time via a dynamic program. The main dynamic program is the following: 
`T(l,c) = min_j j  + T(l-j,c-1) + T(j-1,c)`
Essentially it means that to reverse a graph with l time steps with c memory slots, you go forward j time steps, write a checkpoint, then reverse the end of the graph (l-j time steps but only c-1 checkpoints remaining). One you've reversed the end of the graph, you reverse the beginning of the graph (j-1 initial steps) [I'm not 100% sure about the indexes]. Then you take the j such that this time is minimized. This is what our library H-Revolve does and why it provides the full schedule, it also works with a storage hierarchy.
I made some slides to describe it for multiple levels of storage, but it's similar, see here: http://people.bordeaux.inria.fr/gaupy/ressources/pub/slides/adjoint_comp.pdf (slide 15).


 - the second (more complicated, but can be computed online) is the same, but you can actually guess what the "j" that minimizes the progdyn should be (derived by Griewank and Walther in their revolve paper):
it says that given l and c, any j in the domain `[max(β(c, t − 1), l − β(c − 1, t + 1)+1),min(β(c, t), l − β(c − 1, t)+1)]` is optimal, where where t is the unique integer satisfying β(c, t) ≤ l < β(c, t + 1), and where `β(c, t)= (c+t) !/ (c! t!)` (c chooses c+t).

shoyer:
Thanks @Gaupy! I think the "full schedule" version is exactly what we want for JAX.

Gaupy:
Ok. In the mean time I think we'll change the licence of our code to MIT if you want to try using it while you reimplement it. 

I don't think we chose AGPL on purpose: I was pretty sure we licenced it GPL... I'm definitely interested in any feedback on usability. Note that if you use the "Revolve" function, that gives you the optimal binomial strategy assuming write and reads are negligible in front of the forward step cost. 

If it is not the case you can use Hrevolve that can take as input both these times (more info on the TOMS paper linked below).

shoyer:
Awesome, thanks so much @Gaupy

joglekara:
Are there any traps we should be aware of in using this implementation of recursive checkpointing? Looks really handy but I realize that there could be a good reason why it hasn't been merged in yet.

And as always, thanks so much for your efforts.

PR #4364, opened 2020-09-21
Adding an option to return the output tree in make_jaxpr

botev:


mattjj:
Thanks!

Could we make this match `xla_computation`'s `return_shape` option? (Both in parameter name and in what gets returned.)

botev:
I think this should now be in that format - e.g. returning the out_avals in the correct tree format.

mattjj:
Awesome, thanks! I should've also suggested adding a simple test. I pushed a commit that did that, and also returned ShapeDtypeStruct instances rather than ShapedArrays (to match `xla_computation` better).

mattjj (approved):
LGTM!

botev:
Oh, sorry I did not noticed that the leaves are `ShapeDtypeStruct`, but thanks for making this ready to merge :)

PR #4375, opened 2020-09-22
[jax2tf] Implement proper tracking of which primitives have dedicated tests (WIP)

bchetioui:
We want to know exactly which primitive conversions are missing dedicated tests.
The method we used before was just showing that some primitives were never called throughout the tests, which didn't give much insight into whether they were actually properly tested or not.


gnecula:
I am not yet convinced that we want this code change. I do think that we will need a way to organize the harnesses, but I feel that we have not yet thought through the use cases. E.g., an alternative way would be to add to the harness a tuple of primitives that are covered. We would also need a way to list all the harnesses.

hawkinsp:
Closing as stale.

PR #4377, opened 2020-09-22
Delete dead axis_index code

apaszke:
The primitive was moved to `lax_parallel.py` some time ago, so the one
in `core` should no longer be used. This is probably a result of a
botched rebase.


mattjj (approved):
Thanks! Indeed looks like a botched merged/rebase.

PR #4378, opened 2020-09-22
Fix axis_index inside nested pmaps

apaszke:
The previous translation rule has assumed that `axis_index` is always
taken over the outermost axis in the `axis_env`, and was always producing
the same output, no matter which axis has been specified. This fixes the
translation rule to start taking the `axis_name` into account.

Additionally, this adds support for querying the index along multiple
axes, which will be useful for `gmap`.

### DIFF
--- a/jax/lax/lax_parallel.py
+++ b/jax/lax/lax_parallel.py
@@ -667,9 +667,11 @@ def all_gather(x, axis_name, *, axis_index_groups=None):
 
 
 def _axis_index_translation_rule(c, *, axis_name, axis_env, platform):
-  div = xb.constant(c, np.array(axis_env.nreps // prod(axis_env.sizes),
+  axis_pos = list(axis_env.names).index(axis_name)
+  nreplicas = axis_env.nreps // prod(axis_env.sizes)
+  div = xb.constant(c, np.array(nreplicas * prod(axis_env.sizes[axis_pos+1:]),
                                 dtype=np.uint32))
-  mod = xb.constant(c, np.array(axis_env.sizes[-1], dtype=np.uint32))
+  mod = xb.constant(c, np.array(axis_env.sizes[axis_pos], dtype=np.uint32))
   unsigned_index = xops.Rem(xops.Div(xops.ReplicaId(c), div), mod)
   return xops.ConvertElementType(unsigned_index, xb.dtype_to_etype(np.int32))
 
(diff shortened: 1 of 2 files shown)


apaszke:
@mattjj I'm not sure what does `nreps` represent in the axis env, so I tried my best to retain the semantics of the translation rule while fixing it. Please make sure to read that part carefully. The part that's confusing for me is why do we divide the replica ID by `nreps / prod(axis_env.sizes)`. Is it ever the case that we replicate the computation without extending the `axis_env`?

jax/interpreters/pxla.py
```diff
@@ -955,9 +955,10 @@ def _pmap_translation_rule(c, axis_env,
     _xla_shard(c, aval, new_env, in_node) if in_node_mapped else in_node
     for aval, in_node, in_node_mapped in zip(in_avals, in_nodes, mapped_invars))
 
-  sharded_outs = xla.jaxpr_subcomp(
-      c, call_jaxpr, backend, new_env, (),
-      extend_name_stack(name_stack, wrap_name(name, 'pmap')), *in_nodes_sharded)
+  with core.extend_axis_env(axis_name, global_axis_size, None):
+    sharded_outs = xla.jaxpr_subcomp(
+        c, call_jaxpr, backend, new_env, (),
+        extend_name_stack(name_stack, wrap_name(name, 'pmap')), *in_nodes_sharded)
```
mattjj, line 961: I think we don't have to make this change; we're meant to be using the static axis env `new_env` for the axis environment, not the trace-time one.

jax/lax/lax_parallel.py
```diff
@@ -690,11 +692,20 @@ def _axis_index_soft_pmap_rule(vals, mapped, chunk_size, *, axis_name):
...
+  for name in reversed(axis_name):
+    frame = core.axis_frame(name)
+    if frame.main_trace is not None:
+      trace = frame.main_trace.trace_type(frame.main_trace, core.cur_sublevel())
+      name_idx = trace.process_axis_index(frame)
+    else:
+      name_idx = core.Primitive.bind(axis_index_p, axis_name=name)
+    index += name_idx * inner_size
+    inner_size *= psum(1, name)
+  return index
```
mattjj, line 708: I think we can instead leave the bind rule as it was, and just update the axis_index translation rule to actually use axis_name (and look it up in its axis_env argument).

apaszke: This change is not a fix for the bug, but adds the "Additionally, this adds support for querying the index along multiple
axes, which will be useful for gmap." part.

mattjj:
Looks like I broke this in #3370, and we didn't have test coverage for it! (Before #3370 [the `axis_index` bind rule correctly used `axis_name`](https://github.com/google/jax/pull/3370/files#diff-6609803133099eb146acd9d3128dccc3L463), but #3370 deleted that and I didn't fix the translation rule.)

jax/lax/lax_parallel.py
mattjj: Oh, I see. Thanks for explaining. I think that makes sense.

jax/interpreters/pxla.py
apaszke: You're right. Reverted!

jax/lax/lax_parallel.py
mattjj: So IIUC the idea here is to handle multiple axes in the bind rule, rather than the alternative of handling it in the translation rule and in `process_axis_index`. Is that right? That seems reasonable, though the alternative seems reasonable too. I think the `psum` primitive and its rules work more like the latter (i.e. allowing axis_name to be a list/tuple, ultimately calling into [this helper](https://github.com/google/jax/blob/997e56d0b2cd45fede4cdec117bb6520b8d0cec0/jax/interpreters/xla.py#L462) to handle it), but we don't need to be consistent with that.

mattjj (approved):
LGTM! Thanks for finding and fixing this; I'm embarrassed that I added such a bad bug in #3370 (we think), and moreover that we didn't have test coverage for nested axis index (even though it appears in at least one docstring, and sometimes in our pmap demos!).

PR #4388, opened 2020-09-23
Fix the abstract eval and translation rule for all_to_all

### LINKED ISSUE #1332: pmap all_to_all shape error
On an 8-device machine, if I run a trivial pmap + all_to_all with concat_axis=1:
```
x = np.ones((8, 8, 64))
y = jax.pmap(lambda x: lax.all_to_all(x, axis_name='i', split_axis=0, concat_axis=1), axis_name='i')(x)
print(y.shape, onp.array(y).shape)  # --> returns (8, 8, 64) (8, 1, 512) !!!
```
whereas
```
x = np.ones((8, 8, 64))
y = jax.pmap(lambda x: lax.all_to_all(x, axis_name='i', split_axis=0, concat_axis=0), axis_name='i')(x)
print(y.shape, onp.array(y).shape)  # --> returns (8, 8, 64) (8, 8, 64)  no shape disagreement
```
even just trying to print the repr for the former case `y` yields the error `<repr(<jax.interpreters.pxla.ShardedDeviceArray at 0x7fe547474d08>) failed: RuntimeError: Invalid argument: Argument does not match host shape or layout of computation parameter 0: want f32[8,8,64]{2,1,0}, got f32[8,1,512]{2,0,1}>`

apaszke:
The previous rules assumed that `split_axis == concat_axis` (i.e. that
the used collective is equivalent to `pswapaxes`). Since we expose this
as part of our API, we should probably make sure that we handle other
cases too.

Fixes #1332.

### DIFF
--- a/jax/lax/lax_parallel.py
+++ b/jax/lax/lax_parallel.py
@@ -557,13 +557,25 @@ def _all_to_all_translation_rule(c, x, *, split_axis, concat_axis, axis_name,
   replica_groups = _replica_groups(axis_env, axis_name, None)
   if len(replica_groups[0]) == 1:
     return x
-  else:
+  elif platform == 'tpu':
     split_count = len(replica_groups[0])
     if not all(split_count == len(g) for g in replica_groups):
       raise ValueError('Replica groups must be equally sized')
     replica_groups_protos = xc.make_replica_groups(replica_groups)
-    return xops.AllToAll(x, split_axis, concat_axis, split_count,
-                         replica_groups_protos)
+    if concat_axis == split_axis:
+      return xops.AllToAll(x, split_axis, concat_axis, split_count,
+                           replica_groups_protos)
+    else:
+      if concat_axis < split_axis:
+        split_axis += 1
+      elif split_axis < concat_axis:
+        concat_axis += 1
+      x = xla.lower_fun(partial(lax.expand_dims, dimensions=(concat_axis,)), multiple_results=False)(c, x)
+      x = xops.AllToAll(x, split_axis, concat_axis, split_count, replica_groups_protos)
+      x = xla.lower_fun(partial(lax.squeeze, dimensions=(split_axis,)), multiple_results=False)(c, x)
+      return x
+  else:
+    raise NotImplementedError("all_to_all and pswapaxes only supported on TPU")
 
 def _all_to_all_split_axis_rule(vals, which_mapped, split_axis, concat_axis,
                                 axis_name):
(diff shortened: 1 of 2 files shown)


jax/lax/lax_parallel.py
```diff
@@ -557,13 +557,25 @@ def _all_to_all_translation_rule(c, x, *, split_axis, concat_axis, axis_name,
...
+      return xops.AllToAll(x, split_axis, concat_axis, split_count,
+                           replica_groups_protos)
+    else:
+      if concat_axis < split_axis:
+        split_axis += 1
+      elif split_axis < concat_axis:
+        concat_axis += 1
+      x = xla.lower_fun(partial(lax.expand_dims, dimensions=(concat_axis,)), multiple_results=False)(c, x)
+      x = xops.AllToAll(x, split_axis, concat_axis, split_count, replica_groups_protos)
+      x = xla.lower_fun(partial(lax.squeeze, dimensions=(split_axis,)), multiple_results=False)(c, x)
```
apaszke, line 575: The semantics of XLA's `AllToAll` differ from the JAX's interpretation in that in XLA the collective doesn't insert or remove any dimensions, but only moves chunks of data from one existing dimension to the other existing dimension. Here, I simulate our behavior by inserting the concat axis, running all_to_all, and then squeezing out the now singleton split axis.

jax/lax/lax_parallel.py
```diff
@@ -557,13 +557,25 @@ def _all_to_all_translation_rule(c, x, *, split_axis, concat_axis, axis_name,
...
+      return x
+  else:
+    raise NotImplementedError("all_to_all and pswapaxes only supported on TPU")
```
apaszke, line 578: Interestingly enough, when I tried running `testAllToAll` on CPU then I was actually getting incorrect results, instead of loud "not supported" crashes, so I figured out it might be a good idea to gate that within JAX. My other patch fixes this limitation by adding the fallback through `all_gather`, but I need to land this one first or else the TPU tests start failing.

mattjj: Should we leave a TODO here, and maybe a corresponding bug report / feature request for the XLA:CPU/GPU team? It may be quick for them to fix the issue, so we may not need the all_gather fallback at the JAX level for long.

jax/lax/lax_parallel.py
mattjj: I admire your laziness in using `xla.lower_fun` here rather than writing the reshapes in terms of the HLO builder :P

mattjj (approved):
LGTM, thanks for fixing this longstanding silently buggy stuff.

jax/lax/lax_parallel.py
apaszke: 🙃

apaszke: I'm still trying to put off the day in which I'll have to learn more XLA operators.

jax/lax/lax_parallel.py
apaszke: Sure, I'll create one. I wanted to submit the CPU bug anyway

BohdanK-1QBit: apologies for restarting an old thread (should I rather open a new one) but is there any update on the cpu support for all_to_all? (it seems that since this discussion the GPU support has been added)

PR #4391, opened 2020-09-23
Consider lists as groups of axis names too

apaszke:


apaszke:
I forgot to include that after our discussion yesterday @mattjj

PR #4397, opened 2020-09-24
avals with names

jekbradbury:
List of changes in this PR:

- add named_shape to ShapedArray avals (an unordered set of AxisName: size pairs) modeling “all SPMD axes along which this value may vary”
- revise calls to the ShapedArray constructor to set/propagate named_shape (possible follow-up: dataclass-like aval.update() since most constructor calls ask for “this aval but with these changes”)
- revise mapped_aval and unmapped_aval to use avals with a named axis (so they now implement the bidirectional mapping between a named and positional axis)
- add maybe_unmapped_aval to implement the mapping from a named axis that might or might not be present to a leading positional axis that might or might not be present
- use these new avals for the operands of functions wrapped in map primitives (with the named axis present for mapped operands)
- revise the typechecker’s check_map to enforce this usage 
- revise the checks for call/map primitives in new_eqn_recipe to use the typechecker’s check_call/check_map
- use these new avals in make_jaxpr and xla_computation of “open SPMD” functions (with the named axis present for all operands, for now)
- add a pbroadcast primitive that adds a named axis; have axis_size as a parameter so that jaxprs never require information from the axis environment to be typed
- add axis_size as a parameter to axis_index for the same reason
- add extend_axis_env or the similar replace_axis_env any time implementation code is tracing a function that might call the pbroadcast or axis_index traceables (and therefore read the axis environment)
- require primitive arguments to have matching named_shape, with non-collective primitives implying a pure map over the named axes in their operands
- add named_shape promotion (a kind of implicit broadcasting that’s significantly safer than the ordered-axis case) to multi-operand traceables, using pbroadcast
- pbroadcast the initial loop-carried values for while_loop and scan along all active named axes, in order to avoid named_shape mismatches with the updated version of the loop carry (possible follow-ups: relax this to a fixed-point iteration or remove it entirely and require users to pbroadcast)
- require arguments to collective primitives to have the relevant named axis
- pbroadcast arguments to collective traceables (other than psum) without the relevant named axis
- revise the special case that implements psum(constant) -> axis_size * constant to instead be psum(value without axis X, X) -> axis_size(X) * value; move it from a custom bind to the traceable
- revise all-reduce collectives to remove the named axis they operate over when axis_index_groups is None (but it can be put back by a pbroadcast later)
- revise the transpose of this non-broadcasting psum to be pbroadcast, and set the transpose of pbroadcast to be a non-broadcasting psum
- check the per-shard buffer shape in the ShardedDeviceArray constructor
- fix the jaxpr metadata name stack usage in soft_pmap
- fix ShardingSpec usage in soft_pmap
- revise map primitives to treat return values of wrapped functions that lack the named axis as unmapped outputs (which don’t get an extra positional axis outside the map)
- revise the translation rule for xla_pmap to take into account these unmapped outputs
- use this ability to have an unmapped output in partial_eval (so unmapped constants can become unmapped outputs) and transpose (so unmapped outputs can become unmapped inputs, and vice versa)
- delete the logic that performs fan-in-sum in transpose-of-map when unmapped invars are forced to become mapped outvars
- revise map frontend APIs to add pbroadcasts to the end of their wrapped functions, so all outputs will be mapped as a temporary implementation of the out_axes=0 status quo (follow-up: implement out_axes kwarg that allows both 0 and None)
- add all named axes in the dynamic environment to the scalar 1.0 used as the vjp seed in grad, in order to match the status quo behavior where grad acts “vectorized” or “elementwise” over all SPMD axes (follow-up: implement axis_name kwarg that allows specifying axes that grad should treat akin to bulk axes instead)
- add all named axes in the dynamic environment to the arguments of functions passed to vjp, in order to match the status quo behavior of acting “vectorized” over all SPMD axes even if the function includes pbroadcasts (follow-up: remove this and require all autodiff’ed functions to be pure maps in any named axes that aren’t explicitly listed in grad/vjp/jvp/jacobian)

TODO before merge:

- figure out where helper functions that use pbroadcast should live
- figure out what to do in the non-omnistaging case
- figure out any new tests we need?
- rebase on named vmap; use avals-with-names for batching
- remove extra “mapped” bit from batching and soft_pmap rules/tracers (it’s now redundant with the aval)
- run internal tests

TODO later:

- figure out how this interacts with mask
- (maybe) figure out whether/how to use this in papply
- figure out whether to match the named shape of the argument in things like full_like
- figure out the named shape semantics of the rng_uniform (the XLA RNG)
- (maybe) use named axes in the scan body function (but enforce no collectives)
- support batching control flow through named axes in the condition (in particular, map that to vectorized control flow in batching rules)
- allow calling out named axes in autodiff functions (“treat these named axes like you would positional axes, as opposed to acting elementwise over them”, aka “allow the Jacobian to be non-diagonal along these axes”)
- enforce requirement that functions passed to autodiff actually do elementwise over named axes that aren’t called out in that way

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -61,6 +61,7 @@ _reduce = functools.reduce
 T = TypeVar('T')
 Array = Any
 
+
 @cache()
 def _initial_style_open_jaxpr(fun: Callable, in_tree, in_avals):
   wrapped_fun, out_tree = flatten_fun_nokwargs(lu.wrap_init(fun), in_tree)
@@ -81,6 +82,9 @@ def _initial_style_jaxprs_with_common_consts(funs: Sequence[Callable],
   # their (input) signatures to match. This function "joins" the staged jaxprs:
   # for each one, it makes another that accepts *all* constants, but only uses
   # those that it needs (dropping the rest).
+  # Additionally, this function joins the outputs of the jaxprs in terms of
+  # named-axis avals, by adding `pbroadcast` calls to branches that return
+  # unmapped values where other branches return mapped ones.
 
   jaxprs, all_out_avals, all_consts, all_out_trees = unzip4(
       _initial_style_open_jaxpr(fun, in_tree, in_avals) for fun in funs)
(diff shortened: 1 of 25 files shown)


jekbradbury:
(This code is currently too old for CI to be able to run on it, but it passes tests locally with 8 host devices and omnistaging enabled.)

jakevdp:
Closing as obsolete.

PR #4402, opened 2020-09-24
[host_callback] Fixes and documentation changes for omnistaging

gnecula:
As a nice benefit from omnistaging, the id_print are executed in program order
even if the result is not ued. We keep the functions returning values for
backwards compatibility, but also so the can be used inside expressions and
also so that they can tap values from the backward computation.

Omnistaging introduces jaxpr with no invars, which broke host_callback.

### DIFF
--- a/tests/host_callback_test.py
+++ b/tests/host_callback_test.py
@@ -294,6 +294,8 @@ class HostCallbackTest(jtu.JaxTestCase):
 
   def test_jit_result_unused(self):
     """We can id_print even if we don't use the result."""
+    if not config.omnistaging_enabled:
+      raise SkipTest("Test requires omnistaging")
     def func(x):
       hcb.id_print(x, where="1", output_stream=testing_stream)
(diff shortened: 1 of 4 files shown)


tests/host_callback_test.py
```diff
@@ -712,24 +742,31 @@ def test_jvp(self):
...
-    raise SkipTest("broken by omnistaging")  # TODO(mattjj,gnecula): update
-
     # The output of id_print is not needed for backwards pass
     def func(x):
       return 2. * hcb.id_print(x * 3., what="x * 3",
                                output_stream=testing_stream)
 
     grad_func = api.grad(func)
     jaxpr = str(api.make_jaxpr(grad_func)(5.))
-    # Just making the Jaxpr invokes the id_print once
```
gnecula, line 724: @mattjj Another way in which omnistaging makes the behavior more predictable.

mattjj (commented):
Woohoo, awesome!

PR #4404, opened 2020-09-24
Add support for `all_to_all` in vmap

apaszke:
This allows us to run `all_to_all` over vmapped axes, and even over axis groups mixing vmapped and pmapped axes.

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -964,9 +964,10 @@ def _pmap_translation_rule(c, axis_env,
     _xla_shard(c, aval, new_env, in_node) if in_node_mapped else in_node
     for aval, in_node, in_node_mapped in zip(in_avals, in_nodes, mapped_invars))
 
-  sharded_outs = xla.jaxpr_subcomp(
-      c, call_jaxpr, backend, new_env, (),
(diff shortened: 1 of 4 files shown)


mattjj (approved):
Woohoo! LGTM!

jax/interpreters/pxla.py
```diff
@@ -955,9 +955,10 @@ def _pmap_translation_rule(c, axis_env,
     _xla_shard(c, aval, new_env, in_node) if in_node_mapped else in_node
     for aval, in_node, in_node_mapped in zip(in_avals, in_nodes, mapped_invars))
 
-  sharded_outs = xla.jaxpr_subcomp(
-      c, call_jaxpr, backend, new_env, (),
-      extend_name_stack(name_stack, wrap_name(name, 'pmap')), *in_nodes_sharded)
+  with maybe_extend_axis_env(axis_name, global_axis_size, None):  # type: ignore
```
apaszke, line 958: @mattjj I know you didn't like this line at first, so please re-review. I had to add it to get the tests to pass. The problem is caused by the fact that even at translation time we sometimes evaluate jaxprs via `xla.lower_fun`. Those might contain calls to `axis_index` which will fail if we don't extend the env here.

PR #4407, opened 2020-09-25
[jax2tf] Added Getting Started notebook

gnecula:
This was a Google internal notebook, which we want to open up.

### DIFF
--- a/jax/experimental/jax2tf/JAX2TF_getting_started.ipynb
+++ b/jax/experimental/jax2tf/JAX2TF_getting_started.ipynb
@@ -1,0 +1,1836 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "mB5eSZXZIO9W"
+      },
+      "source": [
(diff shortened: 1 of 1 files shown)


bchetioui (changes requested):
On the public colab, we require the following changes for now:
```python
# line 'hello_tf = jax2tf.convert(hello_jax)' becomes
hello_tf = jax2tf.convert(hello_jax, with_gradient=True)
```

Add the following line:
```
!pip install flax==0.2.1
```

Add lines to enable omnistaging manually before the Flax example:
```python
from jax.config import config
config.enable_omnistaging()
```

Change all occurrences of `"params"` to `"param"` in the Flax example

Tolerance needs to be bumped up on CPU:

```python
elif tf_accelerator.device_type == "CPU":
  tolerances = dict(atol=1e-5, rtol=1e-4) # instead of 1e-6 and 1e-6
```

gnecula:
Thank you @SIben for the feedback. I am tempted to ask JAX core to release a new colab version and only then submit this.

gnecula:
This is superseded by #4460.

PR #4412, opened 2020-09-25
Use core.concrete_or_error() to improve errors in reductions

jakevdp:
Old error:
```python
>>> import jax.numpy as jnp
>>> from jax import jit                                                                             
>>> jit(jnp.sum)(jnp.arange(4), 0) 
[...]
TypeError: Unexpected type of axis argument: <class 'jax.interpreters.partial_eval.DynamicJaxprTracer'>
```
New error:
```python                                                               
>>> jit(jnp.sum)(jnp.arange(4), 0) 
[...]
ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected.

axis argument to jnp.sum().

The error occured while tracing the function sum at /Users/vanderplas/github/google/jax/jax/numpy/lax_numpy.py:1737.

You can use transformation parameters such as `static_argnums` for `jit` to avoid tracing particular arguments of transformed functions.

See https://jax.readthedocs.io/en/latest/faq.html#abstract-tracer-value-encountered-where-concrete-value-is-expected-error for more information.

Encountered tracer value: Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
```

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -867,6 +867,8 @@ def concretization_function_error(fun, suggest_astype=False):
 
 def concrete_or_error(force: Any, val: Any, context=""):
   """Like force(val), but gives the context in the error message."""
+  if force is None:
+    force = lambda x: x
   if isinstance(val, Tracer):
     if isinstance(val.aval, ConcreteArray):
       return force(val.aval.val)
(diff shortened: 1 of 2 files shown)


froystig (approved):

mattjj:
What's the error if you just use `int`?

mattjj:
Sorry, that was cryptic. Let me explain...

mattjj:
Okay, don't mind me; I got confused because I thought I had already merged #4342, and I was confused why it wasn't triggering here. But, well, I haven't merged it!

After #4342 it should show us the additional information about _which_ positional argument to a jitted function was problematic.

Carry on!

PR #4413, opened 2020-09-26
numpy histogramdd implementation

alexminnaar:
Hello, first real contribution here.  I wanted to implement numpy's [histogramdd](https://numpy.org/doc/stable/reference/generated/numpy.histogramdd.html#numpy.histogramdd).  One possible issue is that I wasn't able to use `self._CompileAndCheck` within my test because histogramdd uses multi_ravel_index with mode = 'raise' which causes issues with `self._CompileAndCheck` as shown [here](https://github.com/google/jax/blob/5b3cbc5e18bd4750fb27a0021f2d7012c97013f1/tests/lax_numpy_test.py#L2818).

### DIFF
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -36,7 +36,7 @@ from .lax_numpy import (
     fabs, finfo, fix, flatnonzero, flexible, flip, fliplr, flipud, float16, float32,
     float64, float_, float_power, floating, floor, floor_divide, fmax, fmin,
     fmod, frexp, full, full_like, function, gcd, geomspace, gradient, greater,
-    greater_equal, hamming, hanning, heaviside, histogram, histogram_bin_edges,
+    greater_equal, hamming, hanning, heaviside, histogram, histogram_bin_edges, histogramdd,
     hsplit, hstack, hypot, i0, identity, iinfo, imag,
     indices, inexact, in1d, inf, inner, int16, int32, int64, int8, int_, integer, 
     interp, intersect1d, invert,
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -897,6 +897,44 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
     counts = counts / bin_widths / counts.sum()
   return counts, bin_edges
 
+@_wraps(np.histogramdd)
+def histogramdd(sample, bins=10, range=None, weights=None, density=None):
+  _check_arraylike("histogramdd", sample)
+  N, D = shape(sample)
+
+  if weights is not None and weights.shape != (N,):
+    raise ValueError("should have one weight for each sample.")
+
+  bin_idx_by_dim = D*[None]
+  nbins = np.empty(D, int)
+  bin_edges_by_dim = D*[None]
+  dedges = D*[None]
+
+  for i in builtins.range(D):
+    bin_edges = histogram_bin_edges(sample[:, i], bins[i], range, weights)
+    bin_idx = searchsorted(bin_edges, sample[:, i], side='right')
+    bin_idx = where(sample[:, i] == bin_edges[-1], bin_idx - 1, bin_idx)
+    bin_idx_by_dim[i] = bin_idx
+    nbins[i] = len(bin_edges) + 1
+    bin_edges_by_dim[i] = bin_edges
+    dedges[i] = diff(bin_edges_by_dim[i])
+
+  xy = ravel_multi_index(bin_idx_by_dim, nbins, mode='clip')
+  hist = bincount(xy, weights, length=nbins.prod())
+  hist = reshape(hist, nbins)
+  core = D*(slice(1, -1),)
+  hist = hist[core]
+
+  if density:
+    s = sum(hist)
+    for i in builtins.range(D):
+      _shape = np.ones(D, int)
+      _shape[i] = nbins[i] - 2
+      hist = hist / reshape(dedges[i], _shape)
+
+    hist /= s
+
+  return hist, bin_edges_by_dim
 
 @_wraps(np.heaviside)
 def heaviside(x1, x2):
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -2277,7 +2277,7 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     }
     for shape in [(5,), (5, 5)]
     for dtype in default_dtypes
-    # We only test explicit integer-valued bin edges beause in other cases
+    # We only test explicit integer-valued bin edges because in other cases
     # rounding errors lead to flaky tests.
     for bins in [np.arange(-5, 6), [-5, 0, 3]]
     for density in [True, False]
@@ -2296,6 +2296,37 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     # "TypeError: invalid type promotion with custom data type"
     if dtype != jnp.bfloat16:
       self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False,
+                              tol=tol)
+    self._CompileAndCheck(jnp_fun, args_maker)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+    {"testcase_name": "_{}_bins={}_weights={}_density={}".format(
+      jtu.format_shape_dtype_string(shape, dtype), bins, weights, density),
+      "shape": shape,
+      "dtype": dtype,
+      "bins": bins,
+      "weights": weights,
+      "density": density
+    }
+    for shape in [(5, 3), (10, 3)]
+    for dtype in int_dtypes
+    # We only test explicit integer-valued bin edges because in other cases
+    # rounding errors lead to flaky tests.
+    for bins in [(2, 2, 2), [[-5, 0, 4], [-4, -1, 2], [-6, -1, 4]]]
+    for weights in [False, True]
+    for density in [False, True]
+  ))
+  def testHistogramdd(self, shape, dtype, bins, weights, density):
+    rng = jtu.rand_default(self.rng())
+    _weights = lambda w: abs(w) if weights else None
+    np_fun = lambda a, w: np.histogramdd(a, bins=bins, weights=_weights(w), density=density)
+    jnp_fun = lambda a, w: jnp.histogramdd(a, bins=bins, weights=_weights(w), density=density)
+    args_maker = lambda: [rng(shape, dtype), rng((shape[0],), dtype)]
+    tol = {jnp.bfloat16: 2E-2, np.float16: 1E-1}
+    # np.searchsorted errors on bfloat16 with
+    # "TypeError: invalid type promotion with custom data type"
+    if dtype != jnp.bfloat16:
+        self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False,
                               tol=tol)
     self._CompileAndCheck(jnp_fun, args_maker)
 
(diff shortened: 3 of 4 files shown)


jakevdp:
Thanks for the pull request! I'll plan to do a close review tomorrow morning.

One note: before we merge, we will ask you to squash all your commits into one. If you're not familiar with how to do that, I'll be happy to help you out! Enjoy your weekend :grin:

alexminnaar:
Thanks @jakevdp .  As you could see I was having some trouble with the tests (it looked like some builds were passing and some were not).  I suspect what might be happening (assuming my implementation is correct) is that the jax/numpy histogram differences accumulate as you go to higher dimensions such that it was going beyond the error tolerance.  I tried to relax the tolerances passes in to `_CheckAgainstNumpy` but it appears that `1E-6` was being used no matter what value I passed in.  In the end I just restricted the test to using `int_dtypes` to get the tests to pass but I'm not sure if that's the best way to handle it.

jakevdp:
The way we addressed this in the histogram tests was to only use integer bin edges, where floating point errors will not affect binning: https://github.com/google/jax/blob/2cb795e0d9dd513d613d059fbf5d6b4a7a9b9eaf/tests/lax_numpy_test.py#L2306-L2307

jax/numpy/__init__.py
```diff
@@ -36,7 +36,7 @@
     fabs, finfo, fix, flatnonzero, flexible, flip, fliplr, flipud, float16, float32,
     float64, float_, float_power, floating, floor, floor_divide, fmax, fmin,
     fmod, frexp, full, full_like, function, gcd, geomspace, gradient, greater,
-    greater_equal, hamming, hanning, heaviside, histogram, histogram_bin_edges,
+    greater_equal, hamming, hanning, heaviside, histogramdd, histogram, histogram_bin_edges,
```
jakevdp, line 39: minor: I think `histogramdd` should be after `histogram_bin_edges` to maintain sorted order.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,46 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
     counts = counts / bin_widths / counts.sum()
   return counts, bin_edges
 
+_range = range
```
jakevdp, line 900: You can remove `_range`, and use `builtins.range` below.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,46 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+@_wraps(np.histogramdd)
+def histogramdd(sample, bins=10, range=None, weights=None, density=None):
+  sample = asarray(sample)
```
jakevdp, line 903: Before calling `asarray(sample)`, call `_check_arraylike("histogramdd", sample)`. The reason for this is that we try to avoid silent conversion of lists to arrays, because it can be a silent source of poor performance.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,46 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+
+  if weights is not None and weights.shape != (N,):
+    raise ValueError("should have one weight for each sample.")
+
+  bin_idx_by_dim = D*[None]
+  nbins = D*[None]
+  bin_edges_by_dim = D*[None]
+  dedges = D*[None]
+
+  for i in _range(D):
```
jakevdp, line 914: Would it be possible to call `histogram()` on each dimension, and then combine the results? It seems like this duplicates much of the logic spelled-out there.

tests/lax_numpy_test.py
```diff
@@ -2299,6 +2299,35 @@ def testHistogram(self, shape, dtype, bins, density, weights):
...
+  def testHistogramdd(self, shape, dtype, bins, weights, density):
+    rng = jtu.rand_default(self.rng())
+    _weights = lambda w: abs(w) if weights else None
+    np_fun = lambda a, w: np.histogramdd(a, bins=bins, weights=_weights(w), density=density)
+    jnp_fun = lambda a, w: jnp.histogramdd(a, bins=bins, weights=_weights(w), density=density)
+    args_maker = lambda: [rng(shape, dtype), rng((shape[0],), dtype)]
+    tol = {jnp.bfloat16: 2E-2, np.float16: 1E-1}
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False,
+                              tol=tol)
+    # histogramdd uses ravel_multi_index with mode = 'raise' so cannot perform this check
```
jakevdp, line 2328: It might be worth using a different mode to allow compilation. Would `mode='clip'` be appropriate? Do we ever expect out-of-bounds indices within the `histogramdd` computation?

jakevdp (changes requested):
Looks great, thanks for working on this! A few questions & suggestions below:

jax/numpy/__init__.py
alexminnaar: will do

jax/numpy/lax_numpy.py
alexminnaar: will do

jax/numpy/lax_numpy.py
alexminnaar: will do

jax/numpy/lax_numpy.py
alexminnaar: That's what I was hoping to do initially.  The problem is that we combine across dimensions based on the bin index for each example - unfortunately `histogram()` only returns the counts so the bin indexes would get lost.  [Numpy's implementation ](https://github.com/numpy/numpy/blob/92ebe1e9a6aeb47a881a1226b08218175776f9ea/numpy/lib/histograms.py#L942-L1128) doesn't reuse code from `histogram()` either it looks like.

tests/lax_numpy_test.py
alexminnaar: I'll give it a try

alexminnaar: I tried making the change   
`xy = ravel_multi_index(bin_idx_by_dim, nbins, mode='clip')`
and uncommenting the `_CompileAndCheck` test but end up with that JIT error 
```

jax.traceback_util.FilteredStackTrace: jax.core.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected.

```

jakevdp: It would be worth figuring out what's causing this, and trying to fix it. `jnp.histogram` is compatible with JIT, so I don't think there should be any reason that `jnp.histogramdd` could not be.

jax/numpy/lax_numpy.py
jakevdp: Makes sense, thanks.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,45 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+  hist = reshape(hist, nbins)
+  hist = lax.convert_element_type(hist, float32)
+  core = D*(slice(1, -1),)
+  hist = hist[core]
+
+  if density:
+    s = sum(hist)
+    for i in builtins.range(D):
+      _shape = ones(D, int32)
+      _shape = ops.index_update(_shape, i, nbins[i] - 2)
```
jakevdp, line 933: This might be the problematic piece for compilation. try using numpy operations for the shape to avoid pushing this computation to the device.

alexminnaar: I did some debugging and it looks like the line where the jit error starts happening is 927 i.e. `hist = hist[core]`.  However the stack trace mentions mentions line 923 i.e. `hist = bincount(xy, weights, minlength=prod(asarray(nbins)))`
```
  File "tests/lax_numpy_test.py", line 2323, in <lambda>
    jnp_fun = lambda a, w: jnp.histogramdd(a, bins=bins, weights=_weights(w), density=density)
  File "/home/alexminnaar/jax/jax/numpy/lax_numpy.py", line 923, in histogramdd
    hist = bincount(xy, weights, minlength=prod(asarray(nbins)))
  File "/home/alexminnaar/jax/jax/numpy/lax_numpy.py", line 1520, in bincount
    length = _max(length, minlength)
jax.traceback_util.FilteredStackTrace: jax.core.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected.

```
Any idea what might be the culprit?

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,45 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+    bin_edges = histogram_bin_edges(sample[:, i], bins[i], range)
+    bin_idx = searchsorted(bin_edges, sample[:, i], side='right')
+    bin_idx = where(sample[:, i] == bin_edges[-1], bin_idx - 1, bin_idx)
+    bin_idx_by_dim[i] = bin_idx
+    nbins[i] = len(bin_edges) + 1
+    bin_edges_by_dim[i] = bin_edges
+    dedges[i] = diff(bin_edges_by_dim[i])
+
+  xy = ravel_multi_index(bin_idx_by_dim, nbins)
+  hist = bincount(xy, weights, minlength=prod(asarray(nbins)))
```
jakevdp, line 923: If you're using `jnp.bincount`, you need to pass the `length` parameter in order for it to be compatible with `jit`. See the note here: https://github.com/google/jax/blob/aaa5724eabe8c88036ac3ee808a4bc0714c578ea/jax/numpy/lax_numpy.py#L1496-L1503

jax/numpy/lax_numpy.py
jakevdp: I think the issue is with how you're calling `bincount`. See the comment below

jax/numpy/lax_numpy.py
alexminnaar: Thanks yep it looks like that was the problem.  There were also compilation issues associated with the `prod()` call so I changed that numpy.  I also fixed the compilation issues related to your previous comment.  It looks like the tests are passing now.  Thanks again.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,44 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+  if weights is not None and weights.shape != (N,):
+    raise ValueError("should have one weight for each sample.")
+
+  bin_idx_by_dim = D*[None]
+  nbins = np.empty(D, int)
+  bin_edges_by_dim = D*[None]
+  dedges = D*[None]
+
+  for i in builtins.range(D):
+    bin_edges = histogram_bin_edges(sample[:, i], bins[i], range)
```
jakevdp, line 914: Please pass weights to `histogram_bin_edges`: it does nothing currently, but will future-proof us in case that behavior is ever implemented.

jax/numpy/lax_numpy.py
```diff
@@ -897,6 +897,44 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
...
+    bin_idx = searchsorted(bin_edges, sample[:, i], side='right')
+    bin_idx = where(sample[:, i] == bin_edges[-1], bin_idx - 1, bin_idx)
```
jakevdp, line 916: I think this step may not be required when using `mode='clip'` below.

jakevdp (commented):
Looks great! A couple more minor comments below. In addition, please add histogramdd to the list in https://github.com/google/jax/blob/master/docs/jax.numpy.rst to ensure it appears in the documentation.

Once that is all addressed, please squash everything into a single commit if you're comfortable doing that. Thanks!

jax/numpy/lax_numpy.py
alexminnaar: Are you referring to line 916? When I remove that some of the tests don't pass.

jakevdp: OK, thanks for checking!

jakevdp (approved):
Awesome job, thanks for working on this!

I'm going to trigger our worker to run tests on GPUs and TPUs, and assuming that passes we can get this merged!

PR #4421, opened 2020-09-29
Allow jax objects to be represented by multiple buffers

jakevdp:
Re-landing #3983, which was reverted in #4417.

This version removes the two `isinstance` checks that were problematic for TPU device buffers.


mattjj (approved):
LGTM

PR #4425, opened 2020-09-30
Use `raise .. from ..` to chain exceptions

akihironitta:
I'd like to suggest an enhancement in the way that Python 3's exception chaining is used.

## Motivation
Almost all places where `raise`ing exceptions in `except` use `raise .. from ..` syntax to explicitly chain exceptions ([PEP 3134](https://www.python.org/dev/peps/pep-3134/)). However, at some places, `raise .. from ..` syntax is not used, which can lead users to confusion (because it outputs `The above exception was the direct cause of the following exception` between tracebacks instead of `During handling of the above exception, another exception occurred`).

## Description of the changes
I used either:
- `raise .. from e` to chain exceptions where the exception `e` has valuable information and seems useful for users.
- `raise .. from None` to unchain exceptions where the previous exception has no valuable information and seems noisy and redundant.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -981,8 +981,8 @@ class ShapedArray(UnshapedArray):
   def __len__(self):
     try:
       return self.shape[0]
-    except IndexError:
-      raise TypeError("len() of unsized object")  # same as numpy error
+    except IndexError as err:
+      raise TypeError("len() of unsized object") from err  # same as numpy error
 
   def _len(self, ignored_tracer):
     return len(self)
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -936,9 +936,9 @@ def _initialize_outfeed_receiver(
   """
   try:
     outfeed_receiver_module = xla_extension.outfeed_receiver
-  except AttributeError:
+  except AttributeError as err:
     raise NotImplementedError(
-        "id_tap works only with jaxlib version 0.1.51 and higher")
+        "id_tap works only with jaxlib version 0.1.51 and higher") from err
 
   with _outfeed_receiver.lock:
     if _outfeed_receiver.receiver is not None:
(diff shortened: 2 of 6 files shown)


jakevdp (approved):
Looks good! To satisfy my curiosity, I tried the before/after for one of these using the following code block:
```python
from jax import numpy as jnp
x = jnp.arange(9).reshape(3, 3)
jnp.linalg.matrix_power(x, 2.5)
```
Old traceback looked like this:
```pytb
~/github/google/jax/jax/numpy/linalg.py in matrix_power(a, n)
     73   try:
---> 74     n = operator.index(n)
     75   except TypeError:

TypeError: 'float' object cannot be interpreted as an integer

<ipython-input-1-1ee64e9a3763> in <module>
      1 from jax import numpy as jnp
      2 x = jnp.arange(9).reshape(3, 3)
----> 3 jnp.linalg.matrix_power(x, 2.5)

~/github/google/jax/jax/numpy/linalg.py in matrix_power(a, n)
     74     n = operator.index(n)
     75   except TypeError:
---> 76     raise TypeError("exponent must be an integer, got {}".format(n))
     77 
     78   if n == 0:

TypeError: exponent must be an integer, got 2.5
```
New traceback looks like this:
```pytb
~/github/google/jax/jax/numpy/linalg.py in matrix_power(a, n)
     73   try:
---> 74     n = operator.index(n)
     75   except TypeError as err:

TypeError: 'float' object cannot be interpreted as an integer

<ipython-input-1-1ee64e9a3763> in <module>
      1 from jax import numpy as jnp
      2 x = jnp.arange(9).reshape(3, 3)
----> 3 jnp.linalg.matrix_power(x, 2.5)

~/github/google/jax/jax/numpy/linalg.py in matrix_power(a, n)
     74     n = operator.index(n)
     75   except TypeError as err:
---> 76     raise TypeError("exponent must be an integer, got {}".format(n)) from err
     77 
     78   if n == 0:

TypeError: exponent must be an integer, got 2.5
```

akihironitta:
@jakevdp Thank you for reviewing so quickly and providing the code block that will make the change very easy for others to understand!

PR #4431, opened 2020-10-01
Remove scalar float0 XLA representation.

tomhennigan:
Some places in the code (e.g. lower_fun in xla.py) we need to go from an XLA shape to a ShapedArray and cannot do this if we drop shape information earlier.

BUG=169849944


mattjj (approved):
LGTM thanks for catching this and for identifying the workaround. I think @jakevdp is working on a better fix to `lower_fun` (for other reasons, and the timing is coincidental!).

PR #4437, opened 2020-10-02
Add support for creating float0 arrays.

### LINKED ISSUE #4433: jnp.zeros does not support float0 as a dtype
After #4039, it's no longer possible to pass an integer zero tangent to `jax.jvp`, it instead needs to be a `float0` tangent. However, `jnp.zeros` doesn't actually support creating arrays of dtype `float0`, giving:

```
TypeError: JAX only supports number and bool dtypes, got dtype [('float0', 'V')]
```
Using `np.zeros` works fine, but it seems like `jnp.zeros` should work here too, especially since float0 is a jax-specific thing.

(For context, I have some code that looks like
```
def some_function_custom_jvp_rule(primals, tangents):
    x, y = primals
    dx, _ = tangents
    _, f_jvp_x = jax.jvp(lambda x: foo(x, y), (x,), (dx,))
    # ... do something with f_jvp_x
```
and sometimes `x` is a pytree containing integers. Before, this just worked; now, I have to go in and change the dtypes, because `custom_jvp` gives me integer tangents but `jax.jvp` expects `float0` tangents. Perhaps this is a bug too? I'm not sure.)

tomhennigan:
Fixes #4433.

### DIFF
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -67,7 +67,8 @@ complex_ = np.complex128
 # complex_ = np.complex64
 
 # Trivial vectorspace datatype needed for tangent values of int/bool primals
-float0 = np.dtype([('float0', np.void, 0)])
+float0 = np.dtype([('float0', np.bool_, 0)])
+float0 = np.array((0,), float0).dtype
 
 _dtype_to_32bit_dtype = {
     np.dtype('int64'): np.dtype('int32'),
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -2565,6 +2565,10 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
                                 "JAX only supports number and bool dtypes.*"):
       jnp.array(3, [('a','<i4'),('b','<i4')])
 
+  def testArraySupportsFloat0(self):
+    jnp.array((0,), jax.float0)
+    jnp.zeros([], jax.float0)
+
   def testIssue121(self):
     assert not np.isscalar(jnp.array(3))
 
--- a/jax/lax/lax.py
+++ b/jax/lax/lax.py
@@ -6188,7 +6188,9 @@ def _abstractify(x):
 
 def _check_user_dtype_supported(dtype, fun_name=None):
   np_dtype = np.dtype(dtype)
-  if np_dtype.kind not in "biufc" and np_dtype.type != dtypes.bfloat16:
+  is_bfloat16 = np_dtype.type is dtypes.bfloat16
+  is_float0 = np_dtype == dtypes.float0
+  if np_dtype.kind not in "biufc" and not (is_bfloat16 or is_float0):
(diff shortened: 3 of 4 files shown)


tests/lax_numpy_test.py
```diff
@@ -2565,6 +2565,12 @@ def testArrayUnsupportedDtypeError(self):
                                 "JAX only supports number and bool dtypes.*"):
       jnp.array(3, [('a','<i4'),('b','<i4')])
 
+  def testArraySupportsFloat0(self):
+    with self.assertRaisesRegex(TypeError,
+                                "JAX only supports number and bool dtypes.*"):
+      jnp.array(3, jax.float0)
+      jnp.zeros([], jax.float0)
```
LenaMartens, line 2572: Should you be testing if `jnp.zeros([], jax.float0)` is supported instead of throwing an error? Do we want to support `jnp.array(3, jax.float0)`?

tomhennigan: You are totally right, have fixed.

jax/dtypes.py
```diff
@@ -67,7 +67,8 @@ class _bfloat16_finfo(object):
 # complex_ = np.complex64
 
 # Trivial vectorspace datatype needed for tangent values of int/bool primals
-float0 = np.dtype([('float0', np.void, 0)])
+float0 = np.dtype([('float0', np.bool_, 0)])
+float0 = np.array((0,), float0).dtype
```
tomhennigan, line 71: I don't understand why we need to do this dance.. Anyone know?

mattjj: @shoyer @jakevdp we pray to you for wisdom...

shoyer: This second line creates a differently structured dtype:
```
In [22]: import numpy as np

In [23]: float0 = np.dtype([('float0', np.bool_, 0)])

In [24]: float0_v2 = np.array((0,), float0).dtype

In [25]: float0
Out[25]: dtype([('float0', '?', (0,))])

In [26]: float0_v2
Out[26]: dtype({'names':['float0'], 'formats':[('?', (0,))], 'offsets':[0], 'itemsize':1})
```

I don't know exactly _why_ `dtype` is not idempotent here (it is for the old version of `float0`). My guess is it's a bug in NumPy. As you can imagine, size 0 dtypes are not well tested or used often.

Regardless, I would hesitate to recommend using this version of `float0`, because it does actually allocate 1 bytes of memory per array element:
```
In [3]: np.zeros((10, 10), jax.float0).nbytes
Out[3]: 100
```

jakevdp: Maybe it's a bug that's been fixed in recent numpy version?
```python
import numpy as np
print(np.__version__)
float0 = np.dtype([('float0', np.bool_, 0)])
float0_v2 = np.array((0,), float0).dtype
print(float0)
print(float0_v2)
```
```
1.18.5
[('float0', '?', (0,))]
[('float0', '?', (0,))]
```

PR #4439, opened 2020-10-02
Change select_and_scatter_add's batching rule to avoid unrolling.

hawkinsp:
Add a test for select_and_scatter_add batching.


tests/lax_vmap_test.py
```diff
@@ -594,6 +594,41 @@ def fun(operand, tangents):
...
+  def testSelectAndScatterAdd(self, dtype, padding, rng_factory):
+    rng = rng_factory(self.rng())
+    all_configs = itertools.chain(
+        itertools.product(
+            [(4, 6)],
+            [(2, 1), (1, 2)],
+            [(1, 1), (2, 1), (1, 2)]),
+        itertools.product(
+            [(3, 2, 4, 6)], [(1, 1, 2, 1), (2, 1, 2, 1)],
+            [(1, 2, 2, 1), (1, 1, 1, 1)]))
```
mattjj, line 613: Would it make sense to move this into the parameters of the test case?

mattjj (approved):
LGTM!

tests/lax_vmap_test.py
hawkinsp: Done.

PR #4440, opened 2020-10-02
wrap long line

froystig:
minor change, in order to test source sync.


gnecula:
It seems that the CL for this has been garbage collected. I will try to re-label it "pull ready"

PR #4441, opened 2020-10-02
Implement jnp.choose

jakevdp:


jax/numpy/lax_numpy.py
```diff
@@ -2345,6 +2345,33 @@ def column_stack(tup):
...
+  elif mode == 'wrap':
+    a = a % N
+  elif mode == 'clip':
+    a = clip(a, 0, N - 1)
+  else:
+    raise ValueError(f"mode={mode!r} not understood. Must be 'raise', 'wrap', or 'clip'")
+
+  a, *choices = broadcast_arrays(a, *choices)
+  choices = array(choices).reshape(N, -1)
+  return choices[a.ravel(), arange(a.size)].reshape(a.shape)
```
hawkinsp, line 2372: Is the `.ravel()` needed here? Wouldn't advanced indexing do what you need?

Reshapes can be expensive on TPU, so it's better to avoid them if you can.

hawkinsp (approved):

jax/numpy/lax_numpy.py
jakevdp: Thanks, and Done - it turns out this could be expressed more efficiently using the `indices()` function.

PR #4446, opened 2020-10-03
Fix differentially_private_sgd.py Example

n2cholas:
[This](https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py) example ran with the following two errors.

First, [this](https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py#L141) line caused the following error:
```
TypeError: max requires ndarray or scalar arguments, got <class 'tuple'> at position 0.
```

[This](https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py#L170) line caused the following error:
```
TypeError: can only concatenate list (not "range") to list.
```

This PR fixes both of those so the script runs with the instructions given.

### DIFF
--- a/examples/differentially_private_sgd.py
+++ b/examples/differentially_private_sgd.py
@@ -138,7 +138,7 @@ def clipped_grad(params, l2_norm_clip, single_example_batch):
   nonempty_grads, tree_def = tree_flatten(grads)
   total_grad_norm = jnp.linalg.norm(
       [jnp.linalg.norm(neg.ravel()) for neg in nonempty_grads])
-  divisor = jnp.max((total_grad_norm / l2_norm_clip, 1.))
+  divisor = jnp.max(jnp.array((total_grad_norm / l2_norm_clip, 1.)))
   normalized_nonempty_grads = [g / divisor for g in nonempty_grads]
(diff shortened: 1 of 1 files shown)


examples/differentially_private_sgd.py
```diff
@@ -138,7 +138,7 @@ def clipped_grad(params, l2_norm_clip, single_example_batch):
   nonempty_grads, tree_def = tree_flatten(grads)
   total_grad_norm = jnp.linalg.norm(
       [jnp.linalg.norm(neg.ravel()) for neg in nonempty_grads])
-  divisor = jnp.max((total_grad_norm / l2_norm_clip, 1.))
+  divisor = jnp.max(jnp.array((total_grad_norm / l2_norm_clip, 1.)))
```
mattjj, line 141: How about this instead?

```suggestion
  divisor = jnp.maximum(total_grad_norm / l2_norm_clip, 1.)
```

mattjj (approved):
Thanks! One suggestion.

mattjj:
Thanks for finding and fixing these issues! Now we've just got to solve the compilation problem you identified in the other thread.

PR #4451, opened 2020-10-04
Histogram2d implementation

alexminnaar:
Implementing numpy's [histogram2d](https://numpy.org/doc/stable/reference/generated/numpy.histogram2d.html).  It should be a straight-forward application of histogramdd for the 2d case.

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -1459,6 +1459,8 @@ def _intersect1d_sorted_mask(ar1, ar2, return_indices=False):
 
 @_wraps(np.intersect1d)
 def intersect1d(ar1, ar2, assume_unique=False, return_indices=False):
+  ar1 = core.concrete_or_error(asarray, ar1, "The error arose in intersect1d()")
+  ar2 = core.concrete_or_error(asarray, ar2, "The error arose in intersect1d()")
 
   if not assume_unique:
     if return_indices:
@@ -1558,6 +1560,9 @@ def bincount(x, weights=None, minlength=0, *, length=None):
     msg = f"x argument to bincount must have an integer type; got {x.dtype}"
     raise TypeError(msg)
   if length is None:
+    x = core.concrete_or_error(array, x,
+      "The error occured because of argument 'x' of jnp.bincount. "
+      "To avoid this error, pass a static `length` argument.")
     length = max(x) + 1
   length = _max(length, minlength)
   if ndim(x) != 1:
@@ -1997,7 +2002,7 @@ because its output shape is data-dependent.
 def nonzero(a):
   # Note: this function cannot be jitted because its output has a dynamic
   # shape.
-  a = atleast_1d(a)
+  a = core.concrete_or_error(atleast_1d, a, "The error arose in jnp.nonzero")
   dims = shape(a)
   ndims = len(dims)
   ds = [lax.broadcasted_iota(int_, dims + (1,), i) for i in range(ndims)]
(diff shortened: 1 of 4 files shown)


jax/numpy/lax_numpy.py
```diff
@@ -921,6 +921,19 @@ def histogram(a, bins=10, range=None, weights=None, density=None):
     counts = counts / bin_widths / counts.sum()
   return counts, bin_edges
 
+@_wraps(np.histogram2d)
+def histogram2d(x, y, bins=10, range=None, weights=None, density=None):
+
+  N = len(bins)
```
jakevdp, line 927: It looks like this would result in a `TypeError` for the default value.

tests/lax_numpy_test.py
```diff
@@ -2325,6 +2325,38 @@ def testHistogram(self, shape, dtype, bins, density, weights):
...
+    for bins in [[2, 2], [[0, 1, 3, 5], [0, 2, 3, 4, 6]]]
+    for weights in [False, True]
+    for density in [False, True]
+  ))
+  def testHistogram2d(self, shape, dtype, bins, weights, density):
+    rng = jtu.rand_default(self.rng())
+    _weights = lambda w: abs(w) if weights else None
+    np_fun = lambda a, b, w: np.histogram2d(a, b, bins=bins, weights=_weights(w), density=density)
+    jnp_fun = lambda a, b, w: jnp.histogram2d(a, b, bins=bins, weights=_weights(w), density=density)
+    args_maker = lambda: [rng(shape, dtype), rng(shape, dtype), rng((shape[0],), dtype)]
```
jakevdp, line 2350: In this context, I believe all three results should be `rng(shape, dtype)`

tests/lax_numpy_test.py
```diff
@@ -2325,6 +2325,38 @@ def testHistogram(self, shape, dtype, bins, density, weights):
...
+      "shape": shape,
+      "dtype": dtype,
+      "bins": bins,
+      "weights": weights,
+      "density": density
+    }
+    for shape in [(5,), (12,)]
+    for dtype in int_dtypes
+    # We only test explicit integer-valued bin edges because in other cases
+    # rounding errors lead to flaky tests.
```
jakevdp, line 2340: This comment doesn't match the code below (`bins=[2, 2]` is specifying a number of bins for each axis, rather than explicit integer-valued bin edges)

jakevdp (changes requested):
Looks great! Just a couple minor comments.

jax/numpy/lax_numpy.py
alexminnaar: Yes you're right

tests/lax_numpy_test.py
alexminnaar: yep

tests/lax_numpy_test.py
alexminnaar: Ah that's right.  I think I also made that mistake [here](https://github.com/google/jax/blob/cf1fefcdc65872786e7ed8aafdade35dd4fe8e6c/tests/lax_numpy_test.py#L2339), should I add that change to this PR?

jakevdp: Sure, that sounds good.

tests/lax_numpy_test.py
```diff
@@ -2325,6 +2325,36 @@ def testHistogram(self, shape, dtype, bins, density, weights):
...
+      jtu.format_shape_dtype_string(shape, dtype), bins, weights, density),
+      "shape": shape,
+      "dtype": dtype,
+      "bins": bins,
+      "weights": weights,
+      "density": density
+    }
+    for shape in [(5,), (12,)]
+    for dtype in int_dtypes
+    for bins in [[2, 2], [[0, 1, 3, 5], [0, 2, 3, 4, 6]]]
```
jakevdp, line 2339: I'd also test `bins=2` to cover the case of passing an integer.

jakevdp, line 2339: I'd also test `bins=2` to cover the integer case.

jakevdp:
Looks great, thanks! I'm going to trigger some GPU & TPU tests for this code, and if everything passes we'll get it merged. Thanks for all the work on this, especially through the multiple rounds of review!

tests/lax_numpy_test.py
alexminnaar: Good catch, histogrammdd wasn't handling this case correctly.  The way it should work is if bins is an integer then `bins=D*[bins]`.

jakevdp (approved):

jakevdp:
One thing: can you rebase on master, squash all the commits into one and force-push to this branch? Thanks.

Let me know if you need help with that.

alexminnaar:
I'm not sure what that failure is referring to
` jax/experimental/jax2tf/jax2tf.py:221: error: Cannot call function of unknown type`
Is that due to my changes?

jakevdp:
The error looks unrelated to this change.

Try rebasing on master, and then force-push.

PR #4459, opened 2020-10-06
Add `lax.conv_general_dilated_patches`

romanngg:
Conv-based approach, inspired originally by @jbcdnr (also thanks @blakehechtman for pointing out where to look!)

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -551,7 +551,6 @@ def conv_general_dilated(
   If `dimension_numbers` is `None`, the default is `('NCHW', 'OIHW', 'NCHW')`
   (for a 2D convolution).
   """
-  dnums: ConvDimensionNumbers
   dnums = conv_dimension_numbers(lhs.shape, rhs.shape, dimension_numbers)
   if lhs_dilation is None:
     lhs_dilation = (1,) * (lhs.ndim - 2)
@@ -5934,7 +5933,8 @@ def _canonicalize_precision(precision):
                      f"or a tuple of two lax.Precision values; got {precision}")
 
 
-def conv_dimension_numbers(lhs_shape, rhs_shape, dimension_numbers):
+def conv_dimension_numbers(lhs_shape, rhs_shape, dimension_numbers
+                           ) -> ConvDimensionNumbers:
   """Converts convolution `dimension_numbers` to a `ConvDimensionNumbers`.
 
   Args:
--- a/jax/lax/__init__.py
+++ b/jax/lax/__init__.py
@@ -338,6 +338,9 @@ from jax._src.lax.parallel import (
   psum_p,
   pswapaxes,
 )
+from jax._src.lax.other import (
+  conv_general_dilated_patches
+)
 
 # TODO(phawkins): fix callers and remove these imports.
 from . import lax
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -493,6 +493,190 @@ class LaxTest(jtu.JaxTestCase):
 
     self._CompileAndCheck(fun, args_maker)
 
+  def testConvGeneralDilatedPatchesOverlapping1D(self):
+    lhs = np.array([[1]], np.float32).reshape((1, 1))
+    patches = lax.conv_general_dilated_patches(
+      lhs=lhs,
+      filter_shape=(),
+      window_strides=(),
+      padding='SAME'
+    )
+    self.assertAllClose(lhs, patches)
+
+    dn = ('NHC', 'OIH', 'NHC')
+    lhs = np.array([1, 2, 3, 4, 5], np.float32).reshape((1, -1, 1))
+
+    patches = lax.conv_general_dilated_patches(
+        lhs=lhs,
+        filter_shape=(2,),
+        window_strides=(2,),
+        padding='VALID',
+        dimension_numbers=dn
+    )
+    self.assertAllClose(
+        np.array([[1, 2],
+                  [3, 4]], np.float32).reshape((1, 2, 2)), patches)
+
+    patches = lax.conv_general_dilated_patches(
+        lhs=lhs,
+        filter_shape=(3,),
+        window_strides=(1,),
+        padding='SAME',
+        dimension_numbers=dn
+    )
+    self.assertAllClose(
+        np.array([[0, 1, 2],
+                  [1, 2, 3],
+                  [2, 3, 4],
+                  [3, 4, 5],
+                  [4, 5, 0]], np.float32).reshape((1, 5, 3)), patches)
+
+    patches = lax.conv_general_dilated_patches(
+        lhs=lhs,
+        filter_shape=(3,),
+        window_strides=(1,),
+        padding='SAME',
+        rhs_dilation=(2,),
+        dimension_numbers=dn
+    )
+    self.assertAllClose(
+        np.array([[0, 1, 3],
+                  [0, 2, 4],
+                  [1, 3, 5],
+                  [2, 4, 0],
+                  [3, 5, 0]], np.float32).reshape((1, 5, 3)), patches)
+
+  def testConvGeneralDilatedPatchesOverlapping2D(self):
+    lhs = np.array([[1, 2, 3],
+                    [4, 5, 6]], np.float32).reshape((1, 2, 3, 1))
+    patches = lax.conv_general_dilated_patches(
+        lhs=lhs,
+        filter_shape=(2, 2),
+        window_strides=(1, 1),
+        padding='SAME',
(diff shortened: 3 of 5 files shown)


jax/image/patches.py
```diff
@@ -0,0 +1,118 @@
...
+      after each spatial dimension.
+    lhs_dilation: `None`, or a sequence of `n` integers, giving the
+      dilation factor to apply in each spatial dimension of `lhs`. LHS dilation
+      is also known as transposed convolution.
+    rhs_dilation: `None`, or a sequence of `n` integers, giving the
+      dilation factor to apply in each spatial dimension of `rhs`. RHS dilation
+      is also known as atrous convolution.
+    dimension_numbers: either `None`, or a 3-tuple
+      `(lhs_spec, rhs_spec, out_spec)`, where each element is a string
+      of length `n+2`. `None` defaults to `("NCHWD..., OIHWD..., NCHWD...")`.
```
shoyer, line 68: I find all these references to "LHS" and "RHS" a little confusing. What do they mean in the context of extracting image patches from a single image?

shoyer:
For API inspiration, NumPy is currently considering adding a "sliding window" generation function: https://github.com/numpy/numpy/pull/17394/files

It uses strides instead of convolutions, so it probably isn't directly suitable for JAX, but I do think the API is more intuitive (referring only to axes in the original image). It's also nice how it returns the patches in the indicated shape, rather than as 1D vectors. Is it possible to revise this API along these lines, while still retaining the flexibility/power?

romanngg:
Thanks for the suggestion Stephan, 

I tried to make the API as close to `lax.conv_general_dilated` as possible, since I think (at least this was so in my case, when implementing locally-connected networks) there are cases when users would want to answer "what input patches are subject to the receptive field of a given CNN layer" (same as in https://www.tensorflow.org/api_docs/python/tf/image/extract_patches). In this case  going through a different API could be a bit tedious, and potentially wasteful since internally it would reduce back to a CNN call.

What if perhaps I rather move this function to something like `lax.conv_general_dilated_patches`, and then `np.sliding_window_view` view could be built on top of it later?

jax/image/patches.py
romanngg: These mean the batch of inputs and the (imaginary) convolutional kernel - please see my other comment on trying to follow the `lax.conv_general_dilated` API.

jax/image/patches.py
```diff
@@ -0,0 +1,118 @@
...
+      window_strides=window_strides,
+      padding=padding,
+      lhs_dilation=lhs_dilation,
+      rhs_dilation=rhs_dilation,
+      dimension_numbers=dimension_numbers,
+      precision=precision,
+      feature_group_count=n_channels
+  )
+
+  c = out_spec.index('C')
```
romanngg, line 108: @shoyer I am returning non-flat patches here, but after thinking about it more, I think this may actually be premature (and afaik potentially costly e.g on TPUs), as for example in my usecase (implementing a locally-connected layer) I don't need unflattened patches. Apparently TF doesn't unflatten them either https://www.tensorflow.org/api_docs/python/tf/image/extract_patches, so I think this could be removed too and left to the user / downstream `np.sliding_window_view` function to do the unflattening - lmk what you think!

shoyer:
> What if perhaps I rather move this function to something like `lax.conv_general_dilated_patches`, and then `np.sliding_window_view` view could be built on top of it later?

Sounds good to me! I really like this idea of separate low-level and high-level interfaces. It's a great fit for `jax.lax` vs `jax.numpy` (or `jax.image`).

jax/image/patches.py
shoyer: Agreed!

romanngg:
Thanks, moved to lax!

jax/lax/lax.py
```diff
@@ -573,6 +572,94 @@ def conv_general_dilated(
...
+
+  spatial_size = np.prod(filter_shape)
+  n_channels = lhs.shape[lhs_spec[1]]
+
+  # Move separate `lhs` spatial locations into separate `rhs` channels.
+  rhs = _eye(lhs.dtype, (spatial_size,) * 2, 0).reshape(filter_shape * 2)
+
+  rhs = rhs.reshape((spatial_size, 1) + filter_shape)
+  rhs = np.tile(rhs, (n_channels,) + (1,) * (rhs.ndim - 1))
+  rhs = np.moveaxis(rhs, (0, 1), (rhs_spec[0], rhs_spec[1]))
```
shoyer, line 648: In the post-omnistaging future, it seems like it would probably be wiser to use `jnp` functions for generating these arrays rather than using `np`?

shoyer (approved):

jax/lax/lax.py
romanngg: Good point, I didn't notice I was using `onp`! Given that I can't import `jnp` here since it itself depends on `lax` and this would lead to a circular dependency, where would you advice me to put this function? I could put it in a separate file `lax/lax_other.py` and import in `lax/__init__.py` from there. Lmk what you think, thanks!

shoyer: I would guess we could also get away with a local import inside the function, but `lax/lax_other.py` is probably a better idea!

jax/lax/lax.py
```diff
@@ -573,6 +572,94 @@ def conv_general_dilated(
...
+
+  out = conv_general_dilated(
+      lhs=lhs,
+      rhs=rhs,
+      window_strides=window_strides,
+      padding=padding,
+      lhs_dilation=lhs_dilation,
+      rhs_dilation=rhs_dilation,
+      dimension_numbers=dimension_numbers,
+      precision=precision,
```
shoyer, line 658: This would be a nice place to use the newly support option to use tuple for precision:
https://github.com/google/jax/pull/4586

The LHS precision can always be `Precision.DEFAULT`, because 0 and 1 can be represented exactly in bfloat16. This should make `conv_general_dilated_patches` twice as fast on TPU when using full float32 precision.

romanngg: Done! (set `Precision.DEFAULT` for `rhs`, not `lhs`, right?)

jax/lax/lax.py
romanngg: Moved to lax_other!

tests/lax_test.py
```diff
@@ -493,6 +493,113 @@ def fun(lhs, rhs):
...
+      if c == 'N':
+        self.assertEqual(out_c, patch_c)
+      elif c == 'C':
+        self.assertEqual(out_c * np.prod(filter_shape), patch_c)
+      else:
+        self.assertEqual(out_c, patch_c * filter_shape[filter_spec.index(c)])
+
+        source += [patches_spec.index(c), patches_spec.index(c.lower())]
+
+    # Test that stacking patches together gives the source image, padded.
```
shoyer, line 586: I am correct in understanding that this property only holds for carefully chosen pairs of `filter_shape` and `strides`?

It would be nice to have at least one test cases with explicitly constructed inputs where this doesn't hold.

e.g., patches of size 3 in 1D with strides=1:
[1, 2, 3, 4, 5]
-> [[0, 0, 1], [0, 1, 2], [1, 2, 3], [2, 3, 4], [3, 4, 5], [4, 5, 0], [5, 0, 0]]

romanngg: Great suggestion, done!

jax/lax/__init__.py
```diff
@@ -336,3 +336,6 @@
   psum_p,
   pswapaxes,
 )
+from .lax_other import (
+  conv_general_dilated_patches
```
shoyer, line 340: To get this documented, also add this function to the list in `jax.lax.rst`.

shoyer (approved):

shoyer:
Someone is going to have to remind me how our internal pulldowns from GitHub work (which is how we finalize merging PRs these days), but this looks good to me!

jax/lax/__init__.py
romanngg: Thanks, added!

hawkinsp:
You'll need to rebase this PR on top of head: we moved some files around. Thanks!

romanngg:
rebased

jax/_src/lax/lax.py
```diff
@@ -551,7 +551,6 @@ def conv_general_dilated(
   If `dimension_numbers` is `None`, the default is `('NCHW', 'OIHW', 'NCHW')`
   (for a 2D convolution).
   """
-  dnums: ConvDimensionNumbers
```
shoyer, line 554: did you mean to delete this?

romanngg: Yes, AFAIK this should not be needed since I have annotated the return type of `conv_dimension_numbers` function

PR #4466, opened 2020-10-07
Matmul Transpose Functionality

antoniojkim:
Added transpose features to `jax.numpy.matmul`.

Behaviour should be similar to how transpose is handled with in [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/linalg/matmul). User can optionally choose to pass in arguments `transpose_a` and `transpose_b` to change the contraction dimensions that simulate a similar behaviour.

### Example

```python
a = np.random.random((3, 2))
b = np.random.random((3, 2))
c = jnp.matmul(a, b, transpose_b=True)
```
The above should be functionally and mathematically equivalent to

```python
c = jnp.matmul(a, jnp.transpose(b))
```
However, the first code uses less instructions to accomplish the same task.
```python
>>> jax.make_jaxpr(lambda a, b: jnp.matmul(a, jnp.transpose(b)))(a, b)
{ lambda  ; a b.
  let c = transpose[ permutation=(1, 0) ] b
      d = dot_general[ dimension_numbers=(((1,), (0,)), ((), ()))
                       precision=None ] a c
  in (d,) }
>>> jax.make_jaxpr(lambda a, b: jnp.matmul(a, b, transpose_b=True))(a, b)
{ lambda  ; a b.
  let c = dot_general[ dimension_numbers=(((1,), (1,)), ((), ()))
                       precision=None ] a b
  in (c,) }
```
I have found that the removal of this unnecessary, explicit transpose helped improve performance quite a bit for very large matrices.

## Testing

Created additional matmul test to verify that it produces expected behaviour. Ran all `pytest -n auto tests/lax_numpy_test.py` to verify that all previous behaviour is consistent.

## Help

Not sure if my testing is sufficient. If anybody could help check them and if they are not sufficient, maybe guide me on how to properly add tests to this repo, that would be great.

Also, not sure how to update the `jax.numpy.matmul` documentation. If someone could please help with that, that would be awesome.

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -3154,7 +3154,7 @@ def dot(a, b, *, precision=None):  # pylint: disable=missing-docstring
 
 
 @_wraps(np.matmul, lax_description=_PRECISION_DOC)
-def matmul(a, b, *, precision=None):  # pylint: disable=missing-docstring
+def matmul(a, b, *, precision=None, transpose_a=False, transpose_b=False):  # pylint: disable=missing-docstring
   _check_arraylike("matmul", a, b)
   for i, x in enumerate((a, b)):
     if ndim(x) < 1:
@@ -3209,8 +3209,12 @@ def matmul(a, b, *, precision=None):  # pylint: disable=missing-docstring
 
   a = lax.squeeze(a, tuple(a_squeeze))
   b = lax.squeeze(b, tuple(b_squeeze))
+
+  a_contract_dims = ndim(a) - 1 - (a_is_mat and transpose_a)
+  b_contract_dims = ndim(b) - 1 - (b_is_mat and not transpose_b)
+
   out = lax.dot_general(
-    a, b, (((ndim(a) - 1,), (ndim(b) - 1 - b_is_mat,)), (a_batch, b_batch)),
+    a, b, (((a_contract_dims,), (b_contract_dims,)), (a_batch, b_batch)),
     precision=precision)
   return lax.transpose(out, perm)
 
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1016,6 +1016,27 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
                             tol=tol)
     self._CompileAndCheck(jnp_fun, args_maker)
 
+
+  def testMatmulTransposed(self):
+    a = np.random.random((3, 2))
+    b = np.random.random((3, 2))
+
+    ab1 = jnp.matmul(jnp.transpose(a), b)
+    ab2 = jnp.matmul(a, b, transpose_a=True)
+    jtu._assert_numpy_allclose(ab1, ab2)
+
+    ab1 = jnp.matmul(a, jnp.transpose(b))
+    ab2 = jnp.matmul(a, b, transpose_b=True)
+    jtu._assert_numpy_allclose(ab1, ab2)
+
+    a = np.random.random((3, 2))
+    b = np.random.random((2, 3))
+
+    ab1 = jnp.matmul(jnp.transpose(a), jnp.transpose(b))
+    ab2 = jnp.matmul(a, b, transpose_a=True, transpose_b=True)
+    jtu._assert_numpy_allclose(ab1, ab2)
+
+
   def testTensordotErrors(self):
     a = np.random.random((3, 2, 2))
     b = np.random.random((2,))
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1018,6 +1018,7 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
 
 
   def testMatmulTransposed(self):
+    # Test Matrix Matrix Behaviour
     a = np.random.random((3, 2))
     b = np.random.random((3, 2))
 
@@ -1034,6 +1035,18 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
 
     ab1 = jnp.matmul(jnp.transpose(a), jnp.transpose(b))
     ab2 = jnp.matmul(a, b, transpose_a=True, transpose_b=True)
+    jtu._assert_numpy_allclose(ab1, ab2)
+
+    # Test Tensor Tensor Behaviour
+    a = np.random.random((3, 4, 2))
+    b = np.random.random((3, 4, 2))
+
+    ab1 = jnp.matmul(jnp.transpose(a, [0, 2, 1]), b)
+    ab2 = jnp.matmul(a, b, transpose_a=True)
+    jtu._assert_numpy_allclose(ab1, ab2)
+
+    ab1 = jnp.matmul(a, jnp.transpose(b, [0, 2, 1]))
+    ab2 = jnp.matmul(a, b, transpose_b=True)
     jtu._assert_numpy_allclose(ab1, ab2)
 
 


shoyer:
We _could_ do this, but I would lean towards preserving NumPy's interface for `matmul` and leaving advanced manipulations like this to `einsum` (which lowers to a mix of `dot_general` and `transpose`).

Unfortunately `einsum` doesn't do the desired thing here -- instead it adds another transpose afterwards:
```
>>> jax.make_jaxpr(lambda a, b: jnp.einsum('ij,kj->ik', a, b))(a, b)
{ lambda  ; a b.
  let c = xla_call[ backend=None
                    call_jaxpr={ lambda  ; a b.
                                 let c = dot_general[ dimension_numbers=(((1,), (1,)), ((), ()))
                                                      precision=None ] b a
                                     d = transpose[ permutation=(1, 0) ] c
                                 in (d,) }
                    device=None
                    donated_invars=(False, False)
                    name=_einsum ] a b
  in (c,) }
```

This could possibly be improved with some digging, however.

antoniojkim:
@shoyer Adding the optional transpose functionality does to a certain degree preserve the NumPy interface as all new arguments have default values and were positionally added to the end. No previous usages of the `matmul` function would have to change. Only going forward, the user would have slightly more control over the contraction dimensions than before. The motivation for doing this obviously would be that TensorFlow exposes this functionality in it's [`matmul`](https://www.tensorflow.org/api_docs/python/tf/linalg/matmul) function.

I still think it's a good idea to do this. What are your thoughts?

hawkinsp:
Another reason that it doesn't necessarily make sense to add functionality like `transpose` to `matmul` is that in JAX you have the freedom to form these kinds of composite ops using `jit`.

e.g., you can easily write things like

```
my_matmul = jax.jit(lambda x, y: x @ y.T)
```

I don't think "fewer instructions" is necessarily a good metric for better or worse code; under a `jit`, XLA will perform whole-program optimizations, and it may end up fusing the transpose with the dot anyway.

antoniojkim:
@hawkinsp its not about fewer instructions per-say. It's about performance. I submitted the PR because I found the change to have quite a bit of performance impact with large matrices. Sometimes the `jax.jit` operator does not optimize these out enough. e.g.

```python
>>> a = np.random.random((7, 2, 3))
>>> b = np.random.random((7, 2, 3))
>>> my_matmul = jax.jit(lambda x, y: jnp.matmul(x, jnp.transpose(y, [0, 2, 1])))
>>> jax.make_jaxpr(my_matmul)(a, b)
{ lambda  ; a b.
  let c = xla_call[ backend=None
                    call_jaxpr={ lambda  ; a b.
                                 let c = transpose[ permutation=(0, 2, 1) ] b
                                     d = dot_general[ dimension_numbers=(((2,), (1,)), ((0,), (0,)))
                                                      precision=None ] a c
                                 in (d,) }
                    device=None
                    donated_invars=(False, False)
                    name=<lambda> ] a b
  in (c,) }
```

Clearly, the jit did not optimize out the transpose in favour of changing the `dot_general` contraction dimensions. The additional explicit transpose can have performance implications, especially if the transposed matrix multiplication is used frequently and on large matrices.

> XLA will perform whole-program optimizations, and it may end up fusing the transpose with the dot anyway.

In my experience, while this _can_ happen, it does not always reliably do so. Having the option to explicitly change the contraction dimensions just makes it a lot easier and more reliable.

hawkinsp:
`jax.make_jaxpr` shows JAX IR before optimization, so I wouldn't expect it to have any transpose/dot folding.

If you want to view the optimized XLA HLO (note: internal API use):

```
>>> print(jax.lib.xla_bridge.get_backend().compile(jax.xla_computation(my_matmul)(a, b)).hlo_modules()[0].to_string())
HloModule xla_computation__lambda___6.14

ENTRY %xla_computation__lambda___6.14 (parameter.1: f32[7,2,3], parameter.2: f32[7,2,3]) -> (f32[7,2,2]) {
  %parameter.1 = f32[7,2,3]{2,1,0} parameter(0)
  %parameter.2 = f32[7,2,3]{2,1,0} parameter(1)
  %transpose = f32[7,3,2]{2,1,0} transpose(f32[7,2,3]{2,1,0} %parameter.2), dimensions={0,2,1}, metadata={op_type="transpose" op_name="xla_computation(<lambda>)/jit(<lambda>)/transpose[ permutation=(0, 2, 1) ]" source_file="<ipython-input-44-e6e04a279036>" source_line=3}
  %dot = f32[7,2,2]{2,1,0} dot(f32[7,2,3]{2,1,0} %parameter.1, f32[7,3,2]{2,1,0} %transpose), lhs_batch_dims={0}, lhs_contracting_dims={2}, rhs_batch_dims={0}, rhs_contracting_dims={1}, metadata={op_type="dot_general" op_name="xla_computation(<lambda>)/jit(<lambda>)/dot_general[ dimension_numbers=(((2,), (1,)), ((0,), (0,)))\n                                                     precision=None ]" source_file="<ipython-input-44-e6e04a279036>" source_line=3}
  ROOT %tuple.13 = (f32[7,2,2]{2,1,0}) tuple(f32[7,2,2]{2,1,0} %dot)
}
```

In this case you are right: XLA's CPU backend has not chosen to fuse the transpose with the `dot`. I filed a bug with the XLA folks, given that the underlying kernel does support fused transposes.

(By contrast, on TPU, XLA will fuse these operations, particularly if the matrices are larger.)

I'd also agree with @shoyer that we want to be cautious about extending NumPy's API. It's possible we could change `einsum` to try even harder to avoid adding a transpose to the output.

antoniojkim:
@hawkinsp That is fine. If the API change is a big deal, I'll just change the contraction dimensions explicitly in my code. I look forward to this being resolved internally in the future so that I don't have to do it manually. Thanks for your timely responses!

hawkinsp:
The XLA folks noted that this optimization appears not to be implemented for batched matrix multiplications at the moment. But it will happen for unbatched ones. e.g., if I remove the batch dim from your example, I get:
```
HloModule xla_computation__lambda___2.14

ENTRY %xla_computation__lambda___2.14 (parameter.1: f32[2,3], parameter.2: f32[2,3]) -> (f32[2,2]) {
  %parameter.1 = f32[2,3]{1,0} parameter(0)
  %parameter.2 = f32[2,3]{1,0} parameter(1)
  %dot.1 = f32[2,2]{1,0} dot(f32[2,3]{1,0} %parameter.1, f32[2,3]{1,0} %parameter.2), lhs_contracting_dims={1}, rhs_contracting_dims={1}, metadata={op_type="dot_general" op_name="xla_computation(<lambda>)/jit(<lambda>)/dot_general[ dimension_numbers=(((1,), (0,)), ((), ()))\n                                                     precision=None ]" source_file="<ipython-input-8-81b4a98a625c>" source_line=3}
  ROOT %tuple.13 = (f32[2,2]{1,0}) tuple(f32[2,2]{1,0} %dot.1)
}
```

antoniojkim:
@hawkinsp I see. Is there a ticket for getting it implemented for batched matrix multiplications? Especially when it comes to high-performance ML, having fast batched operations is ideal. While it is possible to implement this functionality manually by explicitly calling `lax.dot_general` and specifying the contraction dimensions, it would be nice if it was conveniently made available within the package.

hawkinsp:
I've reported a bug and hopefully they will fix it. The relevant XLA compilation pass is here:

https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/service/transpose_folding.h?q=transpose_folding.h&ss=tensorflow%2Ftensorflow

Note we weren't ruling out exposing transposed matmuls, just expressing caution about extending NumPy's API and expressing a desire to use composition where possible.

hawkinsp:
https://github.com/tensorflow/tensorflow/commit/19db1548e8b6b614377dbc314207eab92652ae94 fixed the XLA transpose folding and #4507 incorporates it into jaxlib. (We'll need to make a jaxlib release or you'll need to build from source to see it, though.)

antoniojkim:
@hawkinsp Awesome! I'll be sure to try it out. Do you know when the next jaxlib release is going to be?

hawkinsp:
We're going to try to make a jaxlib release in the next few days.

PR #4470, opened 2020-10-07
[jax2tf] Port jax2tf to use omnistaging

gnecula:
The main change is that we use `dynamic=True` to use an
omnistaging-based tracer. This has the benefit that we can
convert to TF even functions with no arguments (previously
they would be constant-folded by JAX prior to the conversion).

We also add an explicit error if the jax2tf.convert transformation
is nested under other JAX transformations.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -564,6 +564,7 @@ aval_method = namedtuple("aval_method", ["fun"])
 
 
 class EvalTrace(Trace):
+  # See comments in https://github.com/google/jax/pull/3370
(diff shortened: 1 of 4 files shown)


gnecula:
I am aware of a test failure, where omnistaging tracer breaks a custom_vjp tests (grad of vmap). Still debugging that.

gnecula:
After discussing with @mattjj I changed the code to use `core.new_base_main` instead of `core.new_main(dynamic=True)`. Now all the tests pass.

PR #4475, opened 2020-10-07
[jax2tf] Fix min_max and add_mul harnesses.

bchetioui:
Some of the array parameters were wrapped in StaticArg(), which
resulted in the function not being actually converted through
jax2tf.

One oddity is that the test of min_max now requires "always_custom_assert" to
be set to True to run properly, implying that even in compiled
mode, the results produced by jax and jax2tf are inconsistent.

I checked the HLO and did not find an obvious reason why this is
the case, so I left it as a TODO until I get the chance to look at it
more closely.

### DIFF
--- a/jax/experimental/jax2tf/tests/primitives_test.py
+++ b/jax/experimental/jax2tf/tests/primitives_test.py
@@ -471,7 +471,16 @@ class JaxPrimitiveTest(tf_test_util.JaxToTfTestCase):
 
   @primitive_harness.parameterized(primitive_harness.lax_min_max)
   def test_min_max(self, harness: primitive_harness.Harness):
-    self.ConvertAndCompare(harness.dyn_fun, *harness.dyn_args_maker(self.rng()))
+    # TODO(bchetioui): discrepancies between TF & JAX when comparing with NaN;
+    # JAX always returns NaN, while TF returns the value NaN is compared with.
+    def custom_assert(result_jax, result_tf):
+      mask = np.isnan(result_jax)
+      self.assertAllClose(result_jax[~ mask], result_tf[~ mask])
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -471,7 +471,16 @@ def test_add_mul(self, harness: primitive_harness.Harness):
 
   @primitive_harness.parameterized(primitive_harness.lax_min_max)
   def test_min_max(self, harness: primitive_harness.Harness):
-    self.ConvertAndCompare(harness.dyn_fun, *harness.dyn_args_maker(self.rng()))
+    # TODO(bchetioui): discrepancies between TF & JAX when comparing with NaN;
```
gnecula, line 474: I think it makes sense to follow-through with this TODO in a subsequent PR. I think we can add to the classifier a class about "tf discrepancies". 

I am not sure how to tie this to the tests. It would be nice if it were true that we use a custom_assert iff there is a known discrepancy. This is probably "almost true".

gnecula (approved):
LGTM

jax/experimental/jax2tf/tests/primitives_test.py
bchetioui: This is indeed "almost true". There are also cases where there exists a set of equally valid results – such as the left and right eigenvectors for `eig` from which JAX and TF may pick different elements. At the same time, there are not that many primitives that implement custom asserts; although we want to avoid duplication, maybe it is OK to manually add a case in the classifier for each primitive that has such discrepancies. Then, we could still detect if this documentation is "outdated" by checking that custom_assert is indeed set in ConvertAndCompare in the case that such a limitation was output by the classifier.

PR #4483, opened 2020-10-07
fix a 'store occupied' error in jax_debug_nans

mattjj:
This code snippet could cause a 'store occupied' error:

```python
@jit
def f(x):
  return x + np.nan

FLAGS.jax_debug_nans = True
f(1)
```

The reason is that in `xla._xla_call_impl` we would run a `linear_util.WrappedFun` twice, first via `xla._xla_callable` and then again directly (i.e. in op-by-op) if we got a nan on the output. Things would work fine if the second execution also raised a nan error, since then the `WrappedFun` wouldn't complete execution, but if the second execution does not raise an error (as in the above case, because `1 + np.nan` doesn't involve any jax primitive executions) then we'd end up with a StoreOccupied error from running the `WrappedFun` twice.

This StoreOccupied error could crop up any time the de-optimized version didn't raise a nan issue, e.g. because of `jit` changing numerical behavior.

The fix is just to intentionally allow re-running the `WrappedFun`, since the whole point of jax_debug_nans is to re-run functions that in normal circumstances we would only want to execute exactly once.

Thanks @jblespiau for finding this!

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -558,8 +558,18 @@ def _xla_call_impl(fun: lu.WrappedFun, *args, device, backend, name, donated_inv
   try:
     return compiled_fun(*args)
   except FloatingPointError:
+    assert FLAGS.jax_debug_nans  # compiled_fun can only raise in this case
     print("Invalid value encountered in the output of a jit function. "
           "Calling the de-optimized version.")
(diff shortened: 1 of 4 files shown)


jblespiau:
Thanks, it makes sense.

PR #4496, opened 2020-10-08
[jax2tf] Ensure that in tests TF does not constant-fold in eager mode…

gnecula:
… before compiling

I had noticed that `tf.function(tf.math.maximum, experimental_compile=True)(1., 2.)` would actually just constant fold the operation in eager mode and compile a constant. This can be prevented by using `input_signature` (which this PR does). Alternatively it seems that it would be prevented by ensuring that the arguments are arrays.

### DIFF
--- a/jax/experimental/jax2tf/tests/tf_test_util.py
+++ b/jax/experimental/jax2tf/tests/tf_test_util.py
@@ -16,7 +16,7 @@ import atexit
 import contextlib
 import logging
 import numpy as np
-from typing import Any, Callable, Optional, Tuple
+from typing import Any, Callable, List, Optional, Tuple
 import tensorflow as tf  # type: ignore[import]
 
 import jax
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -128,14 +128,26 @@ def convert_if_bfloat16(v):
 
     tf_args = tuple(map(convert_if_bfloat16, args))
 
+    def make_input_signature(*tf_args) -> List[tf.TensorSpec]:
+      # tf_args can be PyTrees
+      def make_one_arg_signature(tf_arg):
+        return tf.TensorSpec(np.shape(tf_arg), tf_arg.dtype)
+      return tf.nest.map_structure(make_one_arg_signature, list(tf_args))
```
bchetioui, line 135: Is `tf.nest.map_structure` equivalent to using `jax.tree_util.tree_map` in this case? Also, does that mean we should also replace the simple `map` at line 129 with a similar operation?

gnecula: Indeed, `tf.nest.map_structure` is like `tree_map` (except the latter has some fancy custom nodes, which are not relevant for our tests). Maybe we could use the latter, but I felt that since we are in TF world here, perhaps it is best to stick to TF mechanisms. 

You are right that the map in line 129 should also use `tf.nest.map_structure`. Looking at `convert_if_bfloat16`, should it not use `tf.convert_to_tensor` always?

bchetioui: The case when `tf.convert_to_tensor` is not called is typically gonna occur when the arg in question is a scalar; not sure if it is necessary to explicitly convert there?

gnecula: I have changed the PR to apply tf.nest.map_structure for convert_if_bfloat16

gnecula:
Closing in favor of #4501.

PR #4498, opened 2020-10-08
Add a flag to control the C++ jax.jit behavior.

jblespiau:


jax/api.py
```diff
@@ -86,6 +86,10 @@
 FLAGS = flags.FLAGS
 flags.DEFINE_bool("jax_disable_jit", bool_env("JAX_DISABLE_JIT", False),
                   "Disable JIT compilation and just call original Python.")
+flags.DEFINE_bool("experimental_cpp_jit", False,
+                  "A temporary flag to opt-out of the C++ jax.jit fast path."
```
tomhennigan, line 90: Nit: the flag actually enables the C++ fast path so I'd suggest rewording to "Experimental: Enables C++ fast path for jitted functions. Will be removed in a future release of JAX."

tomhennigan: Also, can we have an environment variable to set the default value like `JAX_DISABLE_JIT`? This is useful for cases where folks are not using `jax.config.config_with_absl()`

mattjj (approved):
LGTM!

PR #4500, opened 2020-10-08
Recast int/bool tangents to float0 in custom_jvp/vjps 

LenaMartens:
This should fix the second concern in #4433: `because custom_jvp gives me integer tangents but jax.jvp expects float0 tangents. Perhaps this is a bug too?`

### DIFF
--- a/jax/interpreters/ad.py
+++ b/jax/interpreters/ad.py
@@ -146,6 +146,17 @@ def unpair_pval(pval):
     aval_1, aval_2 = aval
     return (aval_1, const_1), (aval_2, const_2)
 
+def replace_float0s(primals, tangents):
+  return [core.zeros_like_float0(tangent, dtype(primal))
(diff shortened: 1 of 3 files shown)


jax/interpreters/ad.py
```diff
@@ -146,6 +146,17 @@ def unpair_pval(pval):
     aval_1, aval_2 = aval
     return (aval_1, const_1), (aval_2, const_2)
 
+def remove_float0s(primals, tangents):
```
mattjj, line 149:
```suggestion
def replace_float0s(primals, tangents):
```

mattjj (approved):
Awesome! Great job figuring this out and fixing it in a super clear way. I made one naming sugggestion, "remove" -> "replace", but overall LGTM.

jax/interpreters/ad.py
LenaMartens: Renamed!

PR #4512, opened 2020-10-08
lax_control_flow: retry via function rather than via loop

jakevdp:
Following-up on post-merge review: https://github.com/google/jax/pull/4161#discussion_r501995045

### DIFF
--- a/jax/lax/lax_control_flow.py
+++ b/jax/lax/lax_control_flow.py
@@ -264,29 +264,30 @@ def while_loop(cond_fun: Callable[[T], bool],
       # transformation on it), so we fall back to the primitive version.
       pass
 
-  # The body input and output avals must match exactly. However, we want to account for
-  # the case when init contains weakly-typed values (e.g. Python scalars), with avals that
-  # may not match the output despite being compatible by virtue of their weak type.
-  # To do this, we compute the jaxpr in two passes: first with the raw inputs, and if
-  # necessary, a second time with modified init values.
-  for pass_ in ["first", "second"]:
+  def _create_jaxpr(init_val):
     init_vals, in_tree = tree_flatten((init_val,))
     init_avals = tuple(_map(_abstractify, init_vals))
     cond_jaxpr, cond_consts, cond_tree = _initial_style_jaxpr(cond_fun, in_tree, init_avals)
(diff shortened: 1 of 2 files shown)


mattjj:
I like this version better, but wdyt? Do you like the other way? I think I just prefer functions to loops with breaks, but maybe it just comes down to personal preference...

mattjj (approved):
LGTM if you like it, but if you don't you should follow your heart.

jax/lax/lax_control_flow.py
```diff
@@ -264,29 +264,29 @@ def while_loop(cond_fun, body_fun, init_val):
...
-    new_init_vals, changed = _promote_weak_typed_inputs(init_vals, init_avals, body_jaxpr.out_avals)
-    if not changed:
-      break
-    init_val, = tree_unflatten(in_tree, new_init_vals)
+  # The body input and output avals must match exactly. However, we want to account for
+  # the case when init contains weakly-typed values (e.g. Python scalars), with avals that
+  # may not match the output despite being compatible by virtue of their weak type.
+  # To do this, we compute the jaxpr in two passes: first with the raw inputs, and if
+  # necessary, a second time with modified init values.
+  init_vals, in_tree, init_avals, cond_jaxpr, cond_consts, body_jaxpr, body_consts, body_tree = _create_jaxpr(init_val)
```
mattjj, line 285: To save a little horizontal space we could re-arrange the returned tuple and write this as
```python
init_vals, init_avals, body_jaxpr, *rest = _create_jaxpr(init_val)
new_init_vals, changed = _promote_weak_typed_inputs(init_vals, init_avals, body_jaxpr.out_avals)
if changed:
  new_init_val, = tree_unflatten(in_tree, new_init_vals)
  init_vals, init_avals, body_jaxpr, *rest = _create_jaxpr(new_init_val)
in_tree, cond_jaxpr, cond_consts, body_consts, body_tree = rest
```

We could also consider recomputing stuff if it's too annoying to plumb out.

Anyway we always end up revising code over time, and this code'll be no different, so I suggest just finishing off a version you're happy with and moving on :)

jakevdp: I like that idea. Done.

PR #4513, opened 2020-10-08
pretty print improvements

mattjj:
For primitives that take jaxprs as parameters, print params on a newline.

Also improve print_shapes=True.


gnecula:
It would be useful to see an example of what this prints.

PR #4517, opened 2020-10-09
optimize scan partial_eval to fix #4510

### LINKED ISSUE #4510: VJP of `scan` is unnecessarily copying primals that are also inputs to the scan
Consider the following computation:
```
import jax
import jax.lax as lax
import jax.numpy as jnp
import numpy as np

def cumprod(x):
  s = jnp.ones((64, 1024), jnp.float32)
  return lax.scan(lambda s, x: (x*s, s), s, x)

def forward_and_backward(x, ct, ct_acc):
  primals, pullback = jax.vjp(cumprod, x)
  return pullback((ct_acc, ct))

np.random.seed(1234)
x = jnp.asarray(np.random.randn(1024, 64, 1024))
ct = jnp.asarray(np.random.randn(1024, 64, 1024))
ct_acc = jnp.asarray(np.random.randn(64, 1024))

print(jax.make_jaxpr(forward_and_backward)(x, ct, ct_acc))
```

From the jaxpr:
```
{ lambda  ; a b c.
  let d = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(64, 1024) ] 1.0
      _ _ _ _ e f =
        scan[ jaxpr={ lambda  ; a b c d.
                      let e = mul c a
                      in (e, *, a, *, a, c) }
              length=1024
              linear=(False, True, False, True)
              num_carry=2
              num_consts=0
              reverse=False
              unroll=1 ] d * a *
      _ _ _ g =
        scan[ jaxpr={ lambda  ; a b c d e f.
                      let g = mul b f
                          h = add_any d g
                          i = mul b e
                      in (*, h, *, i) }
              length=1024
              linear=(True, True, True, True, False, False)
              num_carry=2
              num_consts=0
              reverse=True
              unroll=1 ] * c * b e f
  in (g,) }
```
we are apparently choosing to save a copy of `x` inside the forward pass for use in the backward pass. But this is pointless: `x` was an input to the forward pass anyway, so we could have saved ourselves a somewhat expensive copy inside a loop and just forwarded the original input to the backward pass.

mattjj:
Fixes #4510. See that thread for a description of the issue.

In short, in some cases the extensive residuals generated by the vjp-of-scan forward pass would be identical to extensive inputs. ("Extensive" means "scanned-over".) Yet in the forward pass those residuals were being produced by in effect scanning an identity function over the inputs, basically implementing a copy in terms of a loop of dynamic-update-slices. It's more efficient to elide that copy and just forward the inputs to the outputs.

This PR adds that optimization: in scan's partial eval rule, we identify extensive outputs (both residuals and non-residuals, because why not) that are just forwarding inputs (with a copy) and we replace them with the inputs (without a copy).

Before this change, the jaxpr in #4510 looked like

```
{ lambda  ; a b c.
  let d = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(64, 1024) ] 1.0
      _ _ _ _ e f =
        scan[ jaxpr={ lambda  ; a b c d.
                      let e = mul c a
                      in (e, *, a, *, a, c) }
              length=1024
              linear=(False, True, False, True)
              num_carry=2
              num_consts=0
              reverse=False
              unroll=1 ] d * a *
      _ _ _ g =
        scan[ jaxpr={ lambda  ; a b c d e f.
                      let g = mul b f
                          h = add_any d g
                          i = mul b e
                      in (*, h, *, i) }
              length=1024
              linear=(True, True, True, True, False, False)
              num_carry=2
              num_consts=0
              reverse=True
              unroll=1 ] * c * b e f
  in (g,) }
```

After the change, it looks like

```
{ lambda  ; a b c.
  let d = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(64, 1024) ] 1.0
      _ _ _ _ e =
        scan[ jaxpr={ lambda  ; a b c d.
                      let e = mul c a
                      in (e, *, a, *, a) }
              length=1024
              linear=(False, True, False, True)
              num_carry=2
              num_consts=0
              reverse=False
              unroll=1 ] d * a *
      _ _ _ f =
        scan[ jaxpr={ lambda  ; a b c d e f.
                      let g = mul b f
                          h = add_any d g
                          i = mul b e
                      in (*, h, *, i) }
              length=1024
              linear=(True, True, True, True, False, False)
              num_carry=2
              num_consts=0
              reverse=True
              unroll=1 ] * c * b e a
  in (f,) }
```

Notice that before the change the first scan operation (the forward pass) produced two live outputs, while after the change it produces only one. The second scan has the same number of inputs before and after the change, but after the change one of those inputs is the top-level jaxpr's input argument `a` rather than being an output from the first scan.

Since scan is so heavily used, I'm going to run a global presubmit check on this before merging.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -232,9 +232,6 @@ class Literal:
   def __hash__(self):
     assert False
 
-  def __eq__(self, other):
-    assert False
-
   def __repr__(self):
     if hasattr(self, 'hash'):
       return '{}'.format(self.val)
--- a/jax/lax/lax_control_flow.py
+++ b/jax/lax/lax_control_flow.py
@@ -1601,8 +1601,8 @@ def _scan_partial_eval(trace, *tracers, reverse, length, num_consts, num_carry,
 
   # Propagate the forwarded extensive outputs using fwd_extensive.
   out_carry, out_extensive = split_list(out_flat, [num_carry])
-  out_extensive = iter(out_extensive)
-  out_extensive = [next(out_extensive) if i is None else
+  out_extensive_iter = iter(out_extensive)
+  out_extensive = [next(out_extensive_iter) if i is None else
                    tracers[i].pval[1] if tracers[i].is_known() else tracers[i]
                    for i in fwd_extensive]
   out_flat = out_carry + out_extensive
--- a/jax/lax/lax_control_flow.py
+++ b/jax/lax/lax_control_flow.py
@@ -1566,7 +1566,6 @@ def _scan_partial_eval(trace, *tracers, reverse, length, num_consts, num_carry,
   jaxpr_1_opt, out_pvals_1, consts_1 = pe.trace_to_jaxpr(
       lu.wrap_init(core.jaxpr_as_fun(jaxpr_1)), in_pvals_1,
       instantiate=[True] * (num_carry + num_ys) + [False] * num_res)
-
   jaxpr_1_opt = pe.ClosedJaxpr(pe.convert_constvars_jaxpr(jaxpr_1_opt), ())
   num_consts_1 = num_consts + len(consts_1)
   # any now-known residuals are intensive, so we want to revise jaxpr_2 to take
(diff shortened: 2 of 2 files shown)


jekbradbury (approved):

mattjj:
In the "after" jaxpr shown above, we're still forwarding an _intensive_ input (the carry) to an extensive output:

```
      _ _ _ _ e =
        scan[ jaxpr={ lambda  ; a b c d.  # this a is intensive
                      let e = mul c a
                      in (e, *, a, *, a) }  # forwarding to an extensive output broadcasts it
              length=1024
              linear=(False, True, False, True)
              num_carry=2
              num_consts=0
              reverse=False
              unroll=1 ] d * a *
```

froystig (approved):

PR #4518, opened 2020-10-09
[jax2tf] Fix the behavior for shift when the shift amount is negative…

gnecula:
… or large.

It turns out that TF shift behavior is "implementation defined" if the shift
amount is negative or larger or equal to the bitsize of the operand. This is
different than XLA.

We add conditionals to check and fix the corner cases, we expand the tests.

A better solution would be to use an XlaShift operation.


gnecula:
@bchetioui It was quite painful to find this bug, because it manifested itself by a test failure in a Flax model (the model was producing different results with jax2tf, but in some cases the differences looked like numerical errors).

There were TODOs in the test file for negative inputs that I think were caused by the same problem. Also, the shift harnesses did not have enough coverage for large shift amounts. 

I am also wondering how to debug these problems next time they occur in a large example.

bchetioui:
For this example in particular, it seems that there was a known discrepancy, and a simple way to detect it; maybe it would make sense to handle such cases by throwing a `NotImplementedError` from the conversion directly?

If the silent error is not known, maybe we could have some kind of "debug mode" where ops are simultaneously executed in JAX and converted using JAX2TF, and the result is compared at every step? Just a thought.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -151,36 +151,60 @@ def run_tf(mode):
       else:
         assert False
 
-    def is_tf_exception(lim: correctness_stats.Limitation):
-      return (lim.error_type == 'Missing TF support' and
+    def expected_missing_tf_support(lim: correctness_stats.Limitation):
+      return (lim.error_type == correctness_stats.CATEGORY_MISSING_TF_SUPPORT and
+              self.tf_default_device.device_type in lim.devices)
+    def expected_possible_incorrect(lim: correctness_stats.Limitation):
```
bchetioui, line 157: nit: maybe we can be more consistent between the name of the function here and the name of the category?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -856,6 +895,7 @@ def pack(a, b):
 
     # Unpacks the first element of a tuple.
     def fst(t):
+      assert t.dtype == double_word_dtype
```
bchetioui, line 898: nit: was this assert here for debugging purpose or do we actually want to keep it here?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -874,7 +914,8 @@ def snd(t):
 
   def reducer(x, y):
     which = tf_impl[select_prim]
-    return tf_impl[lax.select_p](which(fst(x), fst(y)), x=x, y=y)
+    res = tf_impl[lax.select_p](which(fst(x), fst(y)), x=x, y=y)
+    return res
```
bchetioui, line 918: nit: this could be a one-liner

bchetioui (approved):
LGTM!

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -71,12 +74,15 @@ def _report_failure(error_type: str, msg: str,
...
+    msg = "Primitive is unimplemented in TF"
     if additional_msg:
       msg += '; ' + additional_msg
-    _report_failure(missing_tf_support, msg, np_dtype, devs=devs)
+    _report_failure(CATEGORY_MISSING_TF_SUPPORT, msg, np_dtype, devs=devs)
+
+  def tf_maybe_incorrect(np_dtype: Optional[NpDType] = None,
+                         msg: Optional[str] = None,
+                         devs: Sequence[str] = all_devices) -> None:
+    _report_failure(CATEGORY_INCORRECT_RESULTS, msg, np_dtype, devs=devs)
```
bchetioui, line 85: There is no default message in `tf_maybe_incorrect`, so `msg` should have type `str`!

jax/experimental/jax2tf/jax2tf.py
gnecula: It acts as a documentation for the type of `t`, which matters because of the size of the shift.

PR #4522, opened 2020-10-09
[jax2tf] Add paths that do not use XLA in conv_general_dilated.

bchetioui:
This adds some amount of support for people who want to run
convolutions without having XLA linked in. These paths can
seemingly be converted for TFJS as well. I might have missed
some other "obvious" cases, but this already includes calls like
```python
flax.linen.Conv(features=16, kernel_size=(3, 3), padding='SAME')(x)
```

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -675,6 +675,128 @@ def _conv_general_precision_config_proto(precision):
   proto.operand_precision.append(int(precision))
   return proto
 
+# _try_tf_conv returns a Tensor when it succeeds, or a string describing why
+# it did not succeed otherwise.
+def _try_tf_conv(lhs, rhs, window_strides, padding, lhs_dilation, rhs_dilation,
+                 dimension_numbers, feature_group_count, batch_group_count,
+                 out_shape) -> Union[str, TfVal]:
+  # TODO(bchetioui): this function is not exhaustive wrt which convolution cases
+  # can be translated into TF primitives. Further investigation is needed to
+  # fully flesh it out.
+  if not lhs.dtype in [tf.float16, tf.float32, tf.float64]:
+    return f"tf.nn.convolution is not supported for dtype {lhs.dtype}"
+  if feature_group_count != 1:
+    return "tf.nn.convolution does not support grouped convolutions"
+  # TODO(bchetioui): is there something to do with batch_group_count?
+  if batch_group_count != 1:
+    return "Unimplemented support for batch_group_count != 1"
+  nb_spatial_dimensions = len(lhs.shape) - 2
+  # TF can only deal with 1D, 2D and 3D convolution
+  if nb_spatial_dimensions < 1 or nb_spatial_dimensions > 3:
+    return ("TensorFlow can only handle convolutions with 1, 2, or 3 "
+            "spatial dimensions")
+  # TODO(bchetioui): handle different stride cases
+  if list(window_strides) != [1] * nb_spatial_dimensions:
+    return ("Unimplemented support for window_strides != "
+            f"{tuple([1] * nb_spatial_dimensions)}")
+
+  success = lambda res: (res, None)
+  failure = lambda msg: (None, msg)
+
+  def convert_padding():
+    # TODO(bchetioui): in this instance, we can not use padtype_to_pads as
+    # string padding is not implemented for transposed convolution.
+    if list(lhs_dilation) != [1] * nb_spatial_dimensions:
+      return failure("Padding conversion is not supported for transposed "
+                     "convolution.")
+    lhs_perm, rhs_perm, _ = dimension_numbers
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -683,6 +784,21 @@ def _conv_general_dilated(lhs, rhs, window_strides, padding, lhs_dilation,
...
       dimension_numbers, feature_group_count, batch_group_count, lhs_shape,
       rhs_shape, precision)
+
+  tf_fun, tf_padding, tf_data_format, tf_dilation = _can_use_tf_conv(
+      lhs, rhs, window_strides, padding, lhs_dilation, rhs_dilation,
+      dimension_numbers, feature_group_count, batch_group_count, out_shape
+  )
+  if tf_fun is not None:
+    result = None
+    if tf_fun is tf.nn.convolution:
```
gnecula, line 794: This feel brittle. Wouldn't it be better to just have _can_use_tf_conv return the result, and then you don't have to split the logic between that function and this one?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -675,6 +675,107 @@ def _conv_general_precision_config_proto(precision):
...
+  if list(window_strides) != [1] * nb_spatial_dimensions:
+    return failure_result
+
+  def convert_padding():
+    # TODO(bchetioui): in this instance, we can not use _infer_shape_jax as
+    # string padding is not implemented for transposed convolution.
+    if list(lhs_dilation) != [1] * nb_spatial_dimensions:
+      return None
+    # TF only allows 'VALID' and 'SAME' padding
+    out_valid, = _infer_shape_jax(
```
gnecula, line 703: The `_infer_shape_jax` function is going away (see PR 4360). I am also wondering why is this the right way to pick the padding. I assume from this code that it is not enough to check `padding`, but then why is the fact that the shape is the same as with `VALID` enough reason to pick that padding?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -675,6 +675,107 @@ def _conv_general_precision_config_proto(precision):
   proto.operand_precision.append(int(precision))
   return proto
 
+def _can_use_tf_conv(lhs, rhs, window_strides, padding,
```
gnecula, line 678: Thinking ahead towards the use case classifier, I think it would be useful for this function to return also a string with the reason why one cannot use tf conv. The classifier can use that string to classify what are the common obstacles.

gnecula (commented):
Convolution is really complicated. Thanks for tackling it.

jax/experimental/jax2tf/jax2tf.py
bchetioui: You are right, that assumption is erroneous.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -675,6 +675,98 @@ def _conv_general_precision_config_proto(precision):
...
+  nb_spatial_dimensions = len(lhs.shape) - 2
+  # TF can only deal with 1D, 2D and 3D convolution
+  if nb_spatial_dimensions < 1 or nb_spatial_dimensions > 3:
+    return failure_result
+  # TODO(bchetioui): handle different stride cases
+  if list(window_strides) != [1] * nb_spatial_dimensions:
+    return failure_result
+
+  def convert_padding():
+    # TODO(bchetioui): in this instance, we can not use _infer_shape_jax as
```
gnecula, line 698: Nit: I would reword this comment because soon there won't be any _infer_shape_jax in the code.

gnecula (approved):
LGTM

jax/experimental/jax2tf/jax2tf.py
bchetioui: Very good idea, and done in the latest version!

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -675,6 +675,116 @@ def _conv_general_precision_config_proto(precision):
...
+  if tf_padding is None:
+    return error
+  tf_dim_nums, error = convert_dim_nums()
+  if tf_dim_nums is None:
+    return error
+  return convert_dilation_and_compute_result(tf_padding, tf_dim_nums)
+
+# TODO(bchetioui): enabling this flag permits using a conversion path purely
+# based on TF (and not XLA) for _conv_general_dilated in cases when it is
+# possible. It is disabled by default due to a so far unknown bug.
```
gnecula, line 785: I think it is worth explaining more here, even if we do not understand what is going on. Otherwise, we ourselves in a few months won't know what we meant. 

One of the tests (actually, what is special about this test?), fails with an assertion if tf.nn.convolution is used. 
The result of tf.nn.convolution has a different shape than what JAX expects, but only when experimental_compile=True. We cannot reproduce it as a pure TF test.

PR #4524, opened 2020-10-09
Fix extraneous dtype warning in jnp.mean

### LINKED ISSUE #4490: np.mean raises warning when input is float64 in jax 0.2.1
With the new update, the following script
```python
import numpy as np
import jax.numpy as jnp
jnp.mean(np.ones(30))  # or jnp.mean(1)
```
raises the warning
```
UserWarning: Explicitly requested dtype float64 requested in sum is not available,
and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64
configuration option or the JAX_ENABLE_X64 shell environment variable. See
https://github.com/google/jax#current-gotchas for more.
```
I think this new behavior is unexpected so I would like to open this issue.

jakevdp:
Fixes #4490

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -1865,6 +1865,7 @@ def mean(a, axis=None, dtype=None, out=None, keepdims=False):
       dtype = float_
     else:
       dtype = _dtype(a)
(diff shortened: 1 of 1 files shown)


froystig (approved):

fehiepsi:
Thanks, @jakevdp !

PR #1668, opened 2019-11-12
Lazy sublanguage

mattjj:
**TLDR**

Before this commit, this computation would avoid materializing the iota (`np.arange`) array at trace time:

```python
@jit
def f(x):
  m, n = x.shape
  return x + np.arange(n)
```

But this one would materialize the iota array at trace time and stage it into the computation as a potentially large array constant:

```python
@jit
def f(x):
  m, n = x.shape
  return x + np.arange(m)[:, None]
```
The difference is that previously operations like broadcasts, transposes, and singleton-dimension-adding reshapes (as above) would force otherwise lazy values to be materialized, while after this commit broadcasts, transposes, and those reshapes are all lazy operations that only update metadata and reuse the same device buffer as their input rather than compiling and executing XLA computations and producing new buffers.

Also, `np.eye` and `np.tri` become lazy (in addition to `np.zeros`, `np.ones`, `np.full`).

Finally, this PR replaces the ad-hoc "lazy device constant" system.

In an earlier version of this PR, I had also included the feature that lazy expressions would be fused into eager mode op-by-op compiled computations. After thinking it through with @hawkinsp we decided to split that functionality out into a follow-up PR, if we decide to include it at all.

**TODO**
- [x] add a way to make identities and triangular matrices
- [x] systematic tests
- [x] make `jit` force its arguments (see [comment below](https://github.com/google/jax/pull/1668#issuecomment-561699616))
- [x] microbenchmarks
- [x] write implementation notes below

**Follow-up work** for subsequent PRs
- [x] check that this resolves the issues underlying #1640 (or tweak it to resolve them) [EDIT: actually it's already fixed by #1848]
- [ ] fuse lazy expressions into eager op-by-op computations
- [ ] consider adding dtype conversion to the lazy language
- [ ] remove use of `lax.tie_in` in as many places as possible
- [ ] leverage laziness in `pmap` dispatch
- [ ] also handle reshapes that remove singleton dimensions

## Design idea

The basic idea is to introduce a kind of lazy sublanguage. Each DeviceArray carries an expression in this lazy sublanguage, and when we apply certain operations we'll produce a result that refers to the same underlying buffer as the input but has an updated expression (rather than compiling and executing an XLA computation to produce a new result buffer). Only when we apply functions to the array that don't exist in the lazy sublanguage will we compile and execute an XLA computation (taking into account the any lazy expressions on the inputs).

[...]

PR #4536, opened 2020-10-11
remove constants from lazy sublanguage

mattjj:
Follow-up to #4535 (it's treated as a diff base).

This is **removing**  the device constant part of #1668. We can do this because after #3370 and #4038 omnistaging removes the need for lazy device constants in a jitted context. (They could still in principle be useful in an op-by-op context, but the power:weight isn't worthwhile anymore.)

After this change, the only parts of the lazy sublanguage that remain are those to do with broadcasts and transposes. We may or may not kill those in a follow-up (it hinges on whether any benefit to op-by-op execution is worth the extra complexity).

This change regresses non-omnistaging users. As one particular example, `test_eval_shape_big_random_array` no longer passes with omnistaging disabled.

fyi @jblespiau


mattjj:
Note to self: it makes sense to block this on the removal of `disable_omnistaging()`, since landing this PR beforehand could cause performance regressions.

PR #4539, opened 2020-10-11
Add jax.numpy.polyint

schmrlng:
Though not mentioned in the numpy [documentation](https://numpy.org/doc/stable/reference/generated/numpy.polyint.html), this implementation matches `np.polyint` in the cases `m > 1 and len(k) == 1` as well as `len(k) > m`.

I've also added some `np.` (i.e., base `numpy`) qualifiers in the `polyder` implementation as my impression is that this improves constant-folding for `jax.make_jaxpr` and `jax.jit`.

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -3055,13 +3055,37 @@ def polyadd(a1, a2):
 def polyder(p, m=1):
   p = asarray(p)
   if m < 0:
-    raise ValueError("Order of derivative must be positive")
+    raise ValueError("Order of derivative must be nonnegative")
   if m == 0:
     return p
   if m % 1:
     raise ValueError("m must be an integer")
-  coeff = (arange(len(p), m, -1) - 1 - arange(m)[:, newaxis]).prod(0)
+  coeff = (np.arange(len(p), m, -1) - 1 - np.arange(m)[:, newaxis]).prod(0)
   return p[:-m] * coeff
+
+
+@_wraps(np.polyint)
+def polyint(p, m=1, k=None):
+  p = asarray(p)
+  if m < 0:
+    raise ValueError("Order of antiderivative must be nonnegative")
+  if m == 0:
+    return p
+  if m % 1:
+    raise ValueError("m must be an integer")
+  if k is None:
+    k = zeros(m)
+  else:
+    k = atleast_1d(k)
+    if len(k) == 1 and m > 1:
+      k = k * ones(m)
+    if ndim(k) > 1 or len(k) < m:
+      raise ValueError("k must be a scalar or a rank-1 array of length 1 or >=m.")
+  coeff = 1 / np.prod([np.pad(np.arange(len(p) + i, 0, -1),
+                              (0, m - i), "constant", constant_values=1)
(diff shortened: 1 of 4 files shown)


jax/numpy/lax_numpy.py
```diff
@@ -3055,14 +3055,38 @@ def polyadd(a1, a2):
 def polyder(p, m=1):
```
jakevdp, line 3055: If we're requiring `m` to be concrete (as implied by the change from `arange` to `np.arange` below), then we should call
```python
m = core.concrete_or_error(int, m, "The error arose in jnp.polyder argument `m`")
```
Same in `polyint` below.

jax/numpy/lax_numpy.py
```diff
@@ -3055,14 +3055,38 @@ def polyadd(a1, a2):
...
   if m % 1:
     raise ValueError("m must be an integer")
-  coeff = (arange(len(p), m, -1) - 1 - arange(m)[:, newaxis]).prod(0)
+  coeff = (np.arange(len(p), m, -1) - 1 - np.arange(m)[:, newaxis]).prod(0)
   return p[:-m] * coeff
 
+
+@_wraps(np.polyint)
+def polyint(p, m=1, k=None):
+  p = asarray(p)
```
jakevdp, line 3069: We're moving toward requiring inputs to be arrays to start with (though proceeding carefully for existing functions for backward compatibility). For a new function like this, we should call `_check_arraylike('polyint', p)` before `p = asarray(p)`.

Similar for `k` when it is not `None`.

tests/lax_numpy_test.py
```diff
@@ -1348,6 +1348,25 @@ def testPolyDer(self, a_shape, order, dtype):
...
+       "dtype": dtype, "a_shape": a_shape, "order" : order, "k_shape": k_shape}
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes
+      for order in range(5)
+      for k_shape in set([None, (), (1,), (order,), (order + 1,)])))
+  def testPolyInt(self, a_shape, order, k_shape, dtype):
+    rng = jtu.rand_default(self.rng())
+    np_fun = lambda arg1, arg2: np.polyint(arg1, m=order, k=arg2)
+    jnp_fun = lambda arg1, arg2: jnp.polyint(arg1, m=order, k=arg2)
+    args_maker = lambda: [rng(a_shape, dtype), None if k_shape is None else rng(k_shape, dtype)]
```
jakevdp, line 1365: I can never remember the precedence order of commas and ternary operators. For readability, could you add some parentheses to make this clearer? e.g.
```
lambda: [rng(a_shape, dtype), (None if k_shape is None else rng(k_shape, dtype))]
```

jakevdp (changes requested):
Looks great! Thanks for the contribution. A few minor comments, and then we can get this in.

Also, if you're able to squash the additional changes so that this PR is a single commit, that would be much appreciated.

hawkinsp:
Just as an FYI: we moved some of these files into `jax/_src/numpy`, so you'll need to rebase your PR on top of head.

jakevdp:
`polyint` was added in #5731

PR #4541, opened 2020-10-11
Export expit and logsumexp in jax.nn.functions

alexalemi:
As a convenience.

### DIFF
--- a/jax/nn/functions.py
+++ b/jax/nn/functions.py
@@ -21,7 +21,8 @@ from jax import custom_jvp
 from jax import dtypes
 from jax import lax
 from jax import core
-from jax.scipy.special import expit, logsumexp
+from jax.scipy.special import expit
+from jax.scipy.special import logsumexp as _logsumexp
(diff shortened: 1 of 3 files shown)


jax/nn/functions.py
```diff
@@ -22,6 +22,7 @@
 from jax import lax
 from jax import core
 from jax.scipy.special import expit
+from jax.scipy.special import logsumexp as _logsumexp
```
hawkinsp, line 25: Seems like we could just:
```
from jax.scipy.special import logsumexp
```
and save the extra alias below?

alexalemi:
I had started with that but the linter got upset, complaining `logsumexp`
was imported but not used.

jax/nn/functions.py
alexalemi: I had started with that but the linter then complained that `logsumexp` was imported but not used.

hawkinsp (approved):

hawkinsp (approved):

Issue #4540, opened 2020-10-11
Unexpected error from jax.numpy.asarray when providing explicit dtype


phinate:
This works with regular numpy, but not with jax (==0.2.1):

```
jax.numpy.asarray([2,3], dtype='int')
```
leads to
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-9-515fb738910a> in <module>
      1 import numpy as np
----> 2 jnp.asarray([2,3], dtype='int')

~/melody/.env/lib/python3.8/site-packages/jax/numpy/lax_numpy.py in asarray(a, dtype, order)
   2501 def asarray(a, dtype=None, order=None):
   2502   lax._check_user_dtype_supported(dtype, "asarray")
-> 2503   return array(a, dtype=dtype, copy=False, order=order)
   2504 
   2505 

~/melody/.env/lib/python3.8/site-packages/jax/numpy/lax_numpy.py in array(object, dtype, copy, order, ndmin)
   2463   if type(object) is np.ndarray:
   2464     out = _device_put_raw(object)
-> 2465     if dtype: assert _dtype(out) == dtype
   2466   elif isinstance(object, (DeviceArray, core.Tracer)):
   2467     if isinstance(object, DeviceArray) and copy:

AssertionError: 
```

mattjj:
Hrm, this works for me on github master:

```python
In [1]: import jax

In [2]: jax.numpy.asarray([2,3], dtype='int')
/usr/local/google/home/mattjj/packages/jax/jax/lax/lax.py:6190: UserWarning: Explicitly requested dtype int requested in asarray is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  warnings.warn(msg.format(dtype, fun_name , truncated_dtype))
/usr/local/google/home/mattjj/packages/jax/jax/lax/lax.py:6190: UserWarning: Explicitly requested dtype int requested in array is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  warnings.warn(msg.format(dtype, fun_name , truncated_dtype))
[...]

PR #4543, opened 2020-10-11
remove a double warning message with asarray

mattjj:
Observed in #4540:

```
In [1]: import jax

In [2]: jax.numpy.asarray([2,3], dtype='int')
/usr/local/google/home/mattjj/packages/jax/jax/lax/lax.py:6190: UserWarning: Explicitly requested dtype int requested in asarray is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
/usr/local/google/home/mattjj/packages/jax/jax/lax/lax.py:6190: UserWarning: Explicitly requested dtype int requested in array is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
/usr/local/google/home/mattjj/packages/jax/jax/lib/xla_bridge.py:130: UserWarning: No GPU/TPU found, falling back to CPU.
  warnings.warn('No GPU/TPU found, falling back to CPU.')
Out[2]: DeviceArray([2, 3], dtype=int32)
```

The double warning arises because `jax.numpy.asarray` checks if the user-requested dtype is supported, then calls into `jax.numpy.array` which does the same check. The check on `jax.numpy.asarray` is redundant.

EDIT: Well, I guess the messages aren't completely redundant: one mentions `asarray` while the other mentions `array`.

@jakevdp what do you think? Better to have double warnings? Or maybe we should figure out a better warning mechanism?

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -2500,6 +2500,7 @@ def _can_call_numpy_array(x):
 @_wraps(np.asarray)
 def asarray(a, dtype=None, order=None):
   lax._check_user_dtype_supported(dtype, "asarray")
+  dtype = dtypes.canonicalize_dtype(dtype) if dtype is not None else dtype
   return array(a, dtype=dtype, copy=False, order=order)
 
 


jakevdp:
I would lean slightly toward keeping the warning, and calling `canonicalize_dtype` before passing the dtype to `array` in order to prevent the double warning.

mattjj:
That sounds like a better plan! I'll do that.

jakevdp (approved):

PR #4547, opened 2020-10-12
Enrich the error messages with the bound names that are available.

jblespiau:
The user often do not know whether it's not the correct name, or whether it was not defined, etc. It's easier to get this information when debugging.


mattjj (approved):
LGTM other than a perhaps-extraneous `print`. Take a look at the `print` and then we can merge this!

jblespiau:
Thanks for your sharp eyes

PR #4548, opened 2020-10-12
Ensure values returned by jax.random.truncated_normal() are in range.

hawkinsp:
A user observed -inf values being returned by truncated_normal(), which occur if the uniform random value passed to erfinv() is out of range, e.g., due to rounding. Do more of the computation using jax.random.uniform(), which promises correct behavior in the face of rounding.

As an added security measure, also clamp the outputs of the function to the open interval.

### DIFF
--- a/jax/random.py
+++ b/jax/random.py
@@ -700,10 +700,10 @@ def _multivariate_normal(key, mean, cov, shape, dtype) -> jnp.ndarray:
 
 
 def truncated_normal(key: jnp.ndarray,
-                    lower: Union[float, jnp.ndarray],
-                    upper: Union[float, jnp.ndarray],
-                    shape: Optional[Sequence[int]] = None,
-                    dtype: np.dtype = dtypes.float_) -> jnp.ndarray:
+                     lower: Union[float, jnp.ndarray],
+                     upper: Union[float, jnp.ndarray],
+                     shape: Optional[Sequence[int]] = None,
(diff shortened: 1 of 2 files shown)


jax/random.py
```diff
@@ -739,12 +740,19 @@ def _truncated_normal(key, lower, upper, shape, dtype) -> jnp.ndarray:
...
+  lower = lax.convert_element_type(lower, dtype)
+  upper = lax.convert_element_type(upper, dtype)
+  a = lax.erf(lower / sqrt2)
+  b = lax.erf(upper / sqrt2)
   if not jnp.issubdtype(dtype, np.floating):
     raise TypeError("truncated_normal only accepts floating point dtypes.")
-  u = uniform(key, shape, dtype, minval=jnp.finfo(dtype).tiny)
-  return sqrt2 * lax.erf_inv(a + u * (b - a))
+  u = uniform(key, shape, dtype,
+              minval=lax.nextafter(a, np.array(np.inf, dtype=dtype)), maxval=b)
```
jakevdp, line 750: Why is this `nextafter` required if we are doing the clamping below?

hawkinsp: I guess it isn't needed. Done.

jakevdp (approved):

jakevdp:
Test failures appear to be associated with https://github.com/google/jax/pull/4546, and should disappear if this is synced to master.

PR #4556, opened 2020-10-12
BUG: fix indexing error

### LINKED ISSUE #4551: Jit functions can't necessarily specialize on argument shapes
I'll disclaim that I'm not sure if this is a bug that should be fixed in Jax, or if the Jax "Sharp Bits" documentation should discuss this.

I originally stumbled upon an issue where my jit function was behaving strangely, and behaving differently than the non-jit version. Upon going over the Sharp Bits, I think this pertains to the "functions with argument-**value dependent shapes**" section. At first I thought it was simply my mistake, but then it dawned on me that the function doesn't in fact have value dependent shapes.

Here's a reproducible example:

```python
import jax
import jax.numpy as jnp

def fn(a, qs, idx):
  return a[idx, :, qs].mean(axis=-1).squeeze()

jit_fn = jax.jit(fn)

a = jnp.ones((24, 8, 11))
qs = jnp.arange(11)
fn(a, qs, 4).shape # outputs (8,) as expected
jit_fn(a, qs, 4).shape # outputs (11,)
``` 

Note that the shape of `fn` does not actually depend on the values of `qs`. In fact, none of the "intermediate steps" of the function produce any array whose shape depends on the values of `qs`. I fixed the problem by rewriting the functions as follows:

```python
def better_fn(a, idx):
  return a[idx, :].mean(axis=-1).squeeze()

jit_better_fn = jax.jit(better_fn)

b = a[:, :, qs]
better_fn(b, idx).shape # outputs (8,)
jit_better_fn(b, idx).shape # outputs (8,)
```

This is more evidence to me that the Sharp Bits section of the documentation applies to this case, however I believe that the statement that "specializing on argument shapes is ok" from the documentation is ambiguous.

jakevdp:
Fixes #4551 

Confirmed that the new tests fail without the `lax_numpy.py` change.

### DIFF
--- a/jax/numpy/lax_numpy.py
+++ b/jax/numpy/lax_numpy.py
@@ -3993,7 +3993,7 @@ def _index_to_gather(x_shape, idx):
     idx_no_nones = [(i, d) for i, d in enumerate(idx) if d is not None]
     advanced_pairs = (
       (asarray(e), i, j) for j, (i, e) in enumerate(idx_no_nones)
-      if isinstance(e, (Sequence, ndarray)))
+      if isscalar(e) or isinstance(e, (Sequence, ndarray)))
     advanced_pairs = ((_normalize_index(e, x_shape[j]), i, j)
                       for e, i, j in advanced_pairs)
     advanced_indexes, idx_advanced_axes, x_advanced_axes = zip(*advanced_pairs)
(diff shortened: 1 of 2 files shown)


mattjj (approved):
LGTM, thanks for fixing and testing!

PR #4048, opened 2020-08-13
README.md correct typo

chrisfcarroll:

PR #4560, opened 2020-10-13
Update README.md

hieuza:
Fix a typo in example code.


jakevdp:
Hi - thanks for looking into this! This is actually correct as currently written: the idea is that the output vector from one iteration becomes the input vector for the next. I agree it's a bit confusing... We discussed a bit how to improve this in #4048, but didn't land on any good options.

PR #4563, opened 2020-10-13
Improve pmap axis error in the presence of pytrees

### LINKED ISSUE #4552: pmap error message for mismatched rank with PyTree args
Small usability request :)

Currently, if you pass in an argument with an incorrect rank to a pmap you get the error: `ValueError: pmap got arg {arg_num} of rank 0 but axis to be mapped 0`

The `arg_num` index will correctly point to the position of the faulty argument, but only if your arguments do not contain PyTrees. If your arguments contain PyTrees you could for example get a `ValueError: pmap got arg 35 of rank 0 but axis to be mapped 0` on a function with 5 arguments.

If you pass in an incorrectly sized argument, there is a separate error message for PyTree args (https://github.com/google/jax/blob/ee9ca569890fae79c7e82fe5bad26e5ce9e72227/jax/api.py#L1275).   Maybe something similar can be done here for an incorrect rank?

jakevdp:
Fixes #4552.

Example:
```python
from jax import pmap
import jax.numpy as jnp

@pmap
def f(x):
  return x

z = [jnp.ones((1, n)) for n in range(6)]
x = [z, z, z + [0], z]
f(x)
```
Old error:
```pytb
ValueError: pmap got arg 18 of rank 0 but axis to be mapped 0
``` 
New error:
```pytb
ValueError: pmap got arg 18 of rank 0 but axis to be mapped 0. The tree of shapes is:
(([[(1, 0), (1, 1), (1, 2), (1, 3), (1, 4), (1, 5)], [(1, 0), (1, 1), (1, 2), (1, 3), (1, 4), (1, 5)], [(1, 0), (1, 1), (1, 2), (1, 3), (1, 4), (1, 5), ()], [(1, 0), (1, 1), (1, 2), (1, 3), (1, 4), (1, 5)]],), {})
```

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -1234,14 +1234,15 @@ def vmap(fun: Callable[..., T], in_axes=0, out_axes=0, axis_name=None) -> Callab
 
   return batched_fun
 
-def _get_axis_size(name: str, i:int, shape: Tuple[int, ...], axis: int):
-  try:
-    return shape[axis]
-  except (IndexError, TypeError) as e:
-    raise ValueError(f"{name} got arg {i} of rank {len(shape)} "
-                     f"but axis to be mapped {axis}") from e
-
 def _mapped_axis_size(tree, vals, dims, name):
+  def _get_axis_size(name: str, i:int, shape: Tuple[int, ...], axis: int):
+    try:
+      return shape[axis]
+    except (IndexError, TypeError) as e:
+      ranks = tree_unflatten(tree, [np.ndim(x) for x, d in zip(vals, dims)])
+      raise ValueError(f"{name} got arg {i} of rank {len(shape)} but axis to be mapped {axis}. "
+                       f"The tree of ranks is:\n{ranks}") from e
+
   mapped_axis_sizes = {_get_axis_size(name, i, np.shape(x), d)
                        for i, (x, d) in enumerate(zip(vals, dims))
                        if d is not None}


LenaMartens (approved):
LGTM (and thanks for the quick PR)! 
One comment, does it make more sense to print out the rank of the arguments instead of the shape, or would this be more confusing? I care about getting my ranks right but I don't know if printing `2` instead of `(x, y)` is more readable.

jakevdp:
> does it make more sense to print out the rank of the arguments instead of the shape

I'll defer to your judgment on that: what would you have found most useful when you encountered this?

LenaMartens:
Asked the user about this and we agree showing ranks would make more sense, because it allows you to match the "got rank of x" to the ranks in the printout. 

Rank error would look like:
```
ValueError: pmap got arg 18 of rank 0 but axis to be mapped 0. The tree of ranks is:
(([[2, 2, 2, 2, 2, 2], [2, 2, 2, 2, 2, 2], [2, 2, 2, 2, 2, 0], [2, 2, 2, 2, 2, 2]],), {})
```

jakevdp:
OK, pushed the change to show ranks instead. The error from the above code is now
```python
ValueError: pmap got arg 18 of rank 0 but axis to be mapped 0. The tree of ranks is:
(([[2, 2, 2, 2, 2, 2], [2, 2, 2, 2, 2, 2], [2, 2, 2, 2, 2, 2, 0], [2, 2, 2, 2, 2, 2]],), {})
```

Issue #4569, opened 2020-10-13
No GPU/TPU found, falling back to CPU.


AashinShazar:
Hey guys, I'm trying to run some code on an AWS server with JAX but I seem to be running into the following error:

```
/home/ubuntu/anaconda3/envs/python3/lib/python3.6/site-packages/jax/lib/xla_bridge.py:130: UserWarning: No GPU/TPU found, falling back to CPU.
  warnings.warn('No GPU/TPU found, falling back to CPU.')
```

I've followed the instructions as outlined in the installation guide and installed the appropriate JAX version with GPU support which was CUDA 10.0 in my case. I've tried manually setting the symlink and XLA_FLAGS environment variable but with no luck. 

Has anyone else faced this issue and resolved it? 

I'm using a deep learning image from AWS with ubuntu 16.04 and CUDA 10.0.

skye:
Can you set `export TF_CPP_MIN_LOG_LEVEL=0`, run your code again, and post the full output? It should be enough to just run `import jax; jax.device_count()` in a python shell.

AashinShazar:
```
>>> import jax
2020-10-14 00:38:37.574012: I external/org_tensorflow/tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.0
2020-10-14 00:38:38.159461: I external/org_tensorflow/tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.0
>>> jax.device_count()
2020-10-14 00:38:46.578107: I external/org_tensorflow/tensorflow/compiler/xla/service/service.cc:168] XLA service 0x5650e540a080 initialized for platform Interpreter (this does not guarantee that XLA will be used). Devices:
2020-10-14 00:38:46.578263: I external/org_tensorflow/tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Interpreter, <undefined>
2020-10-14 00:38:46.618461: I external/org_tensorflow/tensorflow/core/platform/profile_utils/cpu_utils.cc:108] CPU Frequency: 2600035000 Hz
2020-10-14 00:38:46.618861: I external/org_tensorflow/tensorflow/compiler/xla/service/service.cc:168] XLA service 0x5650e5399890 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
[...]

PR #4571, opened 2020-10-14
Mention how to turn on logging in "No GPU/TPU found" warning.

skye:
This is useful when debugging when a GPU/TPU is expected to be found,
e.g. #4569, and hopefully isn't too noisy when expecting to run on
CPU.

Also changes the warning to use logging.warn instead of warnings.warn,
and suppresses an inactionable warning from the absl logging module
itself.

Old message:
```
>>> import jax; jax.devices()
/usr/local/google/home/skyewm/jax/jax/lib/xla_bridge.py:130: UserWarning: No GPU/TPU found, falling back to CPU.
  warnings.warn('No GPU/TPU found, falling back to CPU.')
[CpuDevice(id=0)]
```

New message looks like:
```
>>> import jax; jax.devices()
W1013 18:33:31.047047 139848542200192 xla_bridge.py:132] No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
[CpuDevice(id=0)]
```

### DIFF
--- a/jax/lib/xla_bridge.py
+++ b/jax/lib/xla_bridge.py
@@ -23,9 +23,10 @@ XLA. There are also a handful of related casting utilities.
 from functools import partial
 import os
 from typing import Callable, Dict, Tuple, Union
-import warnings
 
 from absl import logging
+# Disable "WARNING: Logging before flag parsing goes to stderr." message
+logging._warn_preinit_stderr = 0
 
 from ..config import flags
 from .. import util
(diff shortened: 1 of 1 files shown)


mattjj (approved):
Just curious, what makes logging.warn better?

skye:
It doesn't print the message twice! Plus it's consistent with how we log elsewhere, so it would show up wherever your logs usually show up (if anyone can figure out how to configure absl...). I'll update the description with the old message for comparison.

PR #4575, opened 2020-10-14
Check that the init_value in reduce_window is indeed a scalar.

### LINKED ISSUE #4574: Missing check in shape rule of reduce_window
`reduce_window` expects to build a reduction function of type `T -> T -> T`, where `T` is the type of the operand and both arguments are scalar. If `init_value` is not a scalar, the reduction function that is constructed is invalid, and we get a `RuntimeError`.

This behavior can be reproduced using:
```python
from jax import lax
from jax import numpy as jnp

params = { "operand": jnp.ones((4, 4), dtype=jnp.int32)
         , "init_value": jnp.zeros((1,), dtype=jnp.int32) 
         , "computation": lax.max
         , "window_dimensions": (2, 2)
         , "window_strides": (2, 2)
         , "padding": "VALID"
         , "base_dilation": (1, 1)
         , "window_dilation": (1, 1)
         }
lax.reduce_window(**params)
```
which results in
```bash
RuntimeError: Invalid argument: Reduction function must return a scalar or tuple of scalars but returns shape: s32[1]: 
This is a bug in JAX's shape-checking rules; please report it!
https://github.com/google/jax/issues
```

bchetioui:
This fixes google/jax#4574.

### DIFF
--- a/jax/lax/lax.py
+++ b/jax/lax/lax.py
@@ -4882,6 +4882,10 @@ def _reduce_window_shape_rule(operand, init_value, *, jaxpr, consts,
     msg = ("reduce_window got inconsistent dtypes for operand and init_value: "
            " got operand dtype {} and init_value dtype {}.")
     raise TypeError(msg.format(operand.dtype, init_value.dtype))
+  if init_value.shape != ():
+    msg = ("reduce_window expected init_value to be a scalar but init_value "
+           "has shape {}.")
+    raise TypeError(msg.format(init_value.shape))
   return _common_reduce_window_shape_rule(
     operand, window_dimensions, window_strides, padding, base_dilation,
     window_dilation)
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2031,6 +2031,22 @@ class LaxTest(jtu.JaxTestCase):
     with self.assertRaisesRegex(ValueError, msg):
       lax.conv_general_dilated(lhs, rhs, **kwargs)
 
+  def test_reduce_window_scalar_init_value_shape_rule(self):
+    # https://github.com/google/jax/issues/4574
+    args = { "operand": np.ones((4, 4), dtype=np.int32)
+           , "init_value": np.zeros((1,), dtype=np.int32)
+           , "computation": lax.max
+           , "window_dimensions": (2, 2)
+           , "window_strides": (2, 2)
+           , "padding": "VALID"
+           , "base_dilation": (1, 1)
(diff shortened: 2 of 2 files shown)


tests/lax_test.py
```diff
@@ -2031,6 +2031,21 @@ def test_conv_general_dilated_different_input_ranks_error(self):
...
+           , "init_value": np.zeros((1,), dtype=np.int32)
+           , "computation": lax.max
+           , "window_dimensions": (2, 2)
+           , "window_strides": (2, 2)
+           , "padding": "VALID"
+           , "base_dilation": (1, 1)
+           , "window_dilation": (1, 1)
+           }
+
+    msg = ("reduce_window expected init_value to have shape")
```
skye, line 2046: Can you put the full error message here? I realize we don't always do it, but I think it's a good practice for verifying the error message is actionable and well-formatted.

jax/lax/lax.py
```diff
@@ -4882,6 +4882,10 @@ def _reduce_window_shape_rule(operand, init_value, *, jaxpr, consts,
     msg = ("reduce_window got inconsistent dtypes for operand and init_value: "
            " got operand dtype {} and init_value dtype {}.")
     raise TypeError(msg.format(operand.dtype, init_value.dtype))
+  if init_value.shape != ():
+    msg = ("reduce_window expected init_value to have shape () but init_value "
```
skye, line 4886: nit: maybe say "reduce_window expected init_value to be a scalar"

skye (changes requested):

jax/lax/lax.py
bchetioui: I was a bit hesitant about that, because depending on context we might expect shape `(1,)` to correspond to a scalar (I think?), but in this case, only `()` is acceptable. WDYT?

skye: I don't think `(1,)` is ever considered a scalar, but I see how that could be confusing. I'm fine with either then. The reason I suggested saying "a scalar" is because I think it reads more clearly and makes it more clear it _always_ needs to be a scalar and doesn't depend on other circumstances (I feel like `()` is used more when you're printing a shape that may be something else too).

bchetioui: After chatting offline, decided to make this change!

tests/lax_test.py
bchetioui: Done! Thanks!

skye (approved):

gnecula (approved):

PR #4578, opened 2020-10-14
Change scatter JVP rule to avoid using floating-point representation …

hawkinsp:
…of ID values.

We can avoid floating point IDs by using the same masking approach we use for tangent values for the primals.


mattjj (approved):
Aw man, I drafted these comments a few hours ago then forgot to hit 'approve'!

PR #4579, opened 2020-10-14
multi-buf: fix donated_invars in _xla_callable

jakevdp:
This fixes the `custom_object_test.py` failures on GPU and TPU.


mattjj (approved):
LGTM!

PR #4589, opened 2020-10-15
[jax2tf] Added a non-XLA conversion path for reduce_window_max.

bchetioui:
The new path uses tf.nn.max_pool. The primitive harness for
reduce_window and related primitives has been rewritten completely
to produce better coverage in less tests.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1266,9 +1266,12 @@ def _select_and_gather_add(tangents: TfVal,
   init_identity = lambda x: pack(const(dtype, init), const(dtype, 0))
 
   out = _specialized_reduce_window(reducer, init_identity,
-                                   pack(operand, tangents), window_dimensions,
-                                   window_strides, padding, base_dilation,
-                                   window_dilation, _in_avals, _out_aval)
+                                   pack(operand, tangents),
+                                   window_dimensions=window_dimensions,
+                                   window_strides=window_strides,
+                                   padding=padding, base_dilation=base_dilation,
+                                   window_dilation=window_dilation,
+                                   _in_avals=_in_avals, _out_aval=_out_aval)
 
   return snd(out)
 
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1330,31 +1330,93 @@ def reducer(arg1: TfVal, arg2: TfVal) -> TfVal:
...
+                          tf_padding, tf_data_format)
+  if not has_batch_dim:
+    result = tf.squeeze(result, 0)
+  if not has_channel_dim:
+    result = tf.squeeze(result, -1)
+  return result
+
+def _specialized_reduce_window(reducer, identity, operand, window_dimensions,
+                               window_strides, padding, base_dilation,
+                               window_dilation, _in_avals, _out_aval, *,
```
gnecula, line 1387: I am very surprised by the position of the '*'. I expected that it should be right after `operand`, because all the others are parameters to the primitive.

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -814,88 +814,77 @@ def _fft_rng_factory(dtype):
...
-          init_value=init_value,
-          computation=computation,
-          window_dimensions=window_dimensions,
-          window_strides=window_strides,
-          padding=padding,
-          base_dilation=base_dilation,
-          window_dilation=window_dilation)
-  for computation in [lax.add, lax.max, lax.min, lax.mul]
-  for dtype in { lax.add: filter(lambda t: t != np.bool_, jtu.dtypes.all)
-               , lax.mul: filter(lambda t: t != np.bool_, jtu.dtypes.all)
```
gnecula, line 835: I meant to point out that I have never see this formatting style, with the lines starting with commas. It is a minor matter, but it does seem annoying. I think that there is no "information content" in the comma and seeing it first in the line seems distracting.

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -814,88 +814,77 @@ def _fft_rng_factory(dtype):
...
-  for base_dilation in [(2, 3)]
+def _make_reduce_window_harness(name, *, shape=(4, 6), base_dilation=(1, 1),
+                                computation=lax.add, window_dimensions=(2, 2),
+                                window_dilation=(1, 1), init_value=0,
+                                window_strides=(1, 1), dtype=np.float32,
+                                padding=((0, 0), (0, 0))):
+  return Harness(f"{name}_shape={jtu.format_shape_dtype_string(shape, dtype)}_initvalue={init_value}_computation={computation.__name__}_windowdimensions={window_dimensions}_windowstrides={window_strides}_padding={padding}_basedilation={base_dilation}_windowdilation={window_dilation}".replace(' ', ''),
+                 lax.reduce_window,
+                 [RandArg(shape, dtype),
+                  StaticArg(np.array(init_value, dtype=dtype)),
```
gnecula, line 825: Is it important for the init_value to be a compile-time constant? If not, I think it is better to have it a dynamic argument (just drop `StaticArg`).

gnecula (approved):
Very nice!

jax/experimental/jax2tf/tests/primitive_harness.py
bchetioui: It's quite common in Haskell code, but it's true I haven't really seen it elsewhere! I find it personally more readable, and it makes it easier to add/remove lines – also the diff contains less lines when doing so, since we don't need to add/remove a comma on the previous line when doing that. Nevertheless, noted :)

jax/experimental/jax2tf/jax2tf.py
bchetioui: They are, and it would probably make more sense to do it like that, you're right. I'll make that change shortly.

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -96,7 +96,8 @@ def _to_np_dtype(dtype) -> NpDType:
   else:
     np_dtype = None
 
-  if prim in [lax.min_p, lax.max_p]:
+  if prim in [lax.min_p, lax.max_p, lax.reduce_window_min_p,
```
gnecula, line 99: Wouldn't this lead to changes in the unimplemented primitives documentation? Maybe you need to re-generate that documentation?

Issue #2491, opened 2020-03-23
Jax scans are slower than expected


dionhaefner:
I am implementing the [tridiagonal matrix algorithm](https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm) (TDMA) to solve many tridiagonal systems of the same shape in two sweeps (one forward and one backward pass).

The shape of each diagonal is something like `(100_000, 100)`, and I vectorize over the leading axis, so this should be reasonably efficient.

In pure NumPy, I would do it like this:

```python
def tdma_naive(a, b, c, d):
    """
    Solves many tridiagonal matrix systems with diagonals a, b, c and RHS vectors d.
    """
    assert a.shape == b.shape and a.shape == c.shape and a.shape == d.shape

    n = a.shape[-1]

    for i in range(1, n):
        w = a[..., i] / b[..., i - 1]
        b[..., i] += -w * c[..., i - 1]
        d[..., i] += -w * d[..., i - 1]

    out = np.empty_like(a)
    out[..., -1] = d[..., -1] / b[..., -1]

    for i in range(n - 2, -1, -1):
        out[..., i] = (d[..., i] - c[..., i] * out[..., i + 1]) / b[..., i]

    return out
```

The JAX implementation looks like this:

```python
def tdma_jax_kernel(a, b, c, d):
    def compute_primes(last_primes, x):
        last_cp, last_dp = last_primes
        a, b, c, d = x

        denom = 1. / (b - a * last_cp)
        cp = c * denom
        dp = (d - a * last_dp) * denom

        new_primes = (cp, dp)
        return new_primes, new_primes

    diags = (a.T, b.T, c.T, d.T)
    init = jnp.zeros((a.shape[1], a.shape[0]))
    _, (cp, dp) = jax.lax.scan(compute_primes, (init, init), diags)

    def backsubstitution(last_x, x):
        cp, dp = x
        new_x = dp - cp * last_x
        return new_x, new_x

    _, sol = jax.lax.scan(backsubstitution, init, (cp[::-1], dp[::-1]))

    return sol[::-1].T
```

I implemented the algorithm in a handful of backends (including a sloppily written CUDA kernel). You can see the results in this Gist:

https://gist.github.com/dionhaefner/a97ef80b77e02b36e4b248bb97541161

The executive summary is that Jax is 2.5x slower than Numba on CPU, and 3x slower than my amateurish CUDA kernel on GPU (but is on par with Numba here).

[...]

PR #4591, opened 2020-10-15
Add axis parameter to lax.scan

### LINKED ISSUE #2509: Expand scan to take in_axes and out_axes parameters
This would allow scanning over an arbitrary axis, not just the leading one, and perhaps even different axes for the different arguments. 

This was mentioned in #2491.

alonfnt:
I have tried to add a bit more convenience on the scan function by specifying the axis to iterate over. It should basically get rid of the need of transposing the input before doing the scan without a performance loss. 

Hoping it can partially fix #2509. I am still not getting the expected performance on #2491 but it think should be faster than working with transposes.  (Not really clear why it isn't in my tests)

The basic idea behind it is to get the correct `index_array`/`dynamic_index_array` of the appropiate axis instead of axis=0.

The current implementation is:
```
In [1]: import jax

In [2]: xs = jax.numpy.full((100, 100, 100), 42)

In [3]: f = lambda carry, x: (carry + x.max(), x.max())

In [4]: %timeit jax.lax.scan(f, 0.0, xs)
86.3 ms ± 1.15 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %timeit jax.lax.scan(f, 0.0, xs, axis=1)
86.2 ms ± 388 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit jax.lax.scan(f, 0.0, xs, axis=2)
88.3 ms ± 753 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
So it does not seem to be any overhead on using axis > 0, and of course the results I have checked are correct. Although I guess there should be some unit tests to cover it. 

```
In [14]: xs2 = jax.numpy.ones((5,10))

In [15]: jax.lax.scan(f, 0.0, xs2, axis=0)
Out[15]:
(DeviceArray(5., dtype=float32),
 DeviceArray([1., 1., 1., 1., 1.], dtype=float32))

In [16]: jax.lax.scan(f, 0.0, xs2, axis=1)
Out[16]:
(DeviceArray(10., dtype=float32),
 DeviceArray([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], dtype=float32))
```
Please, let me know if this is something you are interested in merging and if you have any ideas to improve it.

### DIFF
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -1,7 +1,7 @@
 cloudpickle
 colorama>=0.4.4
 matplotlib
-pillow>=8.3.1
+pillow>=9.1.0
 pytest-benchmark
 pytest-xdist
 wheel
--- a/tests/image_test.py
+++ b/tests/image_test.py
@@ -42,16 +42,8 @@ except ImportError:
 
 try:
   from PIL import Image as PIL_Image
-  # TODO(jakevdp): remove this try/except when pillow requirement is updated.
-  try:
-    # pillow >=9.1.0
-    PIL_Resampling = PIL_Image.Resampling
-  except AttributeError:
-    # pillow <9.1.0
-    PIL_Resampling = PIL_Image
 except ImportError:
   PIL_Image = None
-  PIL_Resampling = None
 
 config.parse_flags_with_absl()
 
@@ -114,10 +106,10 @@ class ImageTest(jtu.JaxTestCase):
     args_maker = lambda: (rng(image_shape, dtype),)
     def pil_fn(x):
       pil_methods = {
-        "nearest": PIL_Resampling.NEAREST,
-        "bilinear": PIL_Resampling.BILINEAR,
-        "bicubic": PIL_Resampling.BICUBIC,
-        "lanczos3": PIL_Resampling.LANCZOS,
+        "nearest": PIL_Image.Resampling.NEAREST,
+        "bilinear": PIL_Image.Resampling.BILINEAR,
+        "bicubic": PIL_Image.Resampling.BICUBIC,
+        "lanczos3": PIL_Image.Resampling.LANCZOS,
       }
       img = PIL_Image.fromarray(x.astype(np.float32))
       out = np.asarray(img.resize(target_shape[::-1], pil_methods[method]),


hawkinsp:
This look reasonable to me, but it needs some tests (probably in `lax_control_flow_test.py`).

mattjj:
Thanks for exploring this!

I think thorough tests are the top priority here. There are some things, like the scan batching rule, that assume the scanned-over axis is always the leading one, and so e.g. it's [safe to place the batch axis as the second axis](https://github.com/google/jax/blob/81b7212cf9f67f26cac68e2465ab995ef4bd8d92/jax/lax/lax_control_flow.py#L1763). I think some code with those assumptions will need to be revised. The only way to ferret out all such assumptions is to have really thorough tests.

Luckily I don't think there are any fundamental blockers to this generalization!

AlbertAlonso:
Hi @mattjj, could you show me some short snippets of user code that **should** work. That way I can extend them into unit tests and then check that everything is working, since Jax is a large project and I am having a hard time thinking of tests.

I am eager to get this properly tested and merged now that I can focus on it!

hamzamerzic:
Just checking in what the latest on this PR is @mattjj. I have an open TODO to include this in Haiku [link](https://github.com/deepmind/dm-haiku/blob/30b2cb5ce29ea13fc75087146658ec158c007c0a/haiku/_src/recurrent.py#L177). Thanks!

hamzamerzic:
Is there anything blocking here? Would be great to get this one in. Happy to take it on myself. Cheers.

mattjj:
Thanks for the pings. Sorry for the 6+ month delay time!

The blocker is adding thorough tests. I think the ideal tests would cover cases like scanning over different axis indices (maybe making the current test a parametrized test?), autodiff, and vmapping. I'm especially concerned about vmapping, because I suspect the `_scan_batching_rule` function may still assume that the scanned-over axis is the leading axis (though that's just a suspicion!). The file lax_vmap_test.py has some examples for how to write thorough vmap tests, which cover lots of combinations of axes for different arguments.

Does that make sense?

hamzamerzic:
Oh, I just realized I completely missed your message. That makes sense! I'll see if I can take a look at this next week. Thanks!

alonfnt:
@hamzamerzic As it has been a while since I opened this PR, I cleaned a bit the commits and parametrized the current tests as Mathew suggested. Everything is currently working for the normal implementation, which is what I needed, but when I tried to parametrize the tests that deal with batching such as `testScanLinearize` and `testScanGrad`, I did get failing results which none of my attempts were able to fix. It seems that the reason is, as Mathews correctly guessed, the `_scan_batching_rule` function and its assumption of leading axis. 

You are more than welcome to continue this PR or throw it all and start from scratch, whichever suits you better :smile:

PR #4593, opened 2020-10-15
Fix init with constant Poly shapes

jheek:
I'm getting issues with initializer functions in combination with automasking.
Typically when automasking the batch dim of a NN there will be constant `Poly` dims like `Poly(num_features)` passed to the initializers.

As a workaround we can cast the shape to ints we evaluate the Poly's such that the downstream arithmetic will work out correctly.

In the future it would probably be better if the init functions also support non-constant Poly dims but that needs quite a bit more work.

### DIFF
--- a/jax/interpreters/masking.py
+++ b/jax/interpreters/masking.py
@@ -110,8 +110,11 @@ def eval_poly(poly, values_dict):
 def _ensure_poly(p):
   if type(p) is Poly:
     return p
-
   return Poly({Mon(): p})
+
+def _polys_to_ints(shape):
+  return tuple(int(d) if type(d) is Poly and d.is_constant else d
+               for d in shape)
 
 def is_polymorphic(shape: Sequence[Union[int, 'Poly']]):
   return any(map(lambda d: type(d) is Poly, shape))
@@ -408,9 +411,10 @@ class MaskTrace(Trace):
     if primitive.name == 'reshape': params['polymorphic_shapes'] = polymorphic_shapes
     out = masking_rule(vals, logical_shapes, **params)
     if primitive.multiple_results:
-      return map(partial(MaskTracer, self), out, (o.shape for o in out_aval))
+      out_shapes = map(_polys_to_ints, [o.shape for o in out_aval])
+      return map(partial(MaskTracer, self), out, out_shapes)
     else:
-      return MaskTracer(self, out, out_aval.shape)
+      return MaskTracer(self, out, _polys_to_ints(out_aval.shape))
 
   def process_call(self, call_primitive, f, tracers, params):
     assert call_primitive.multiple_results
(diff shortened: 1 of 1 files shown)


jekbradbury:
I'm not super up to date on the implementation of masking, but don't think we should have to add a bunch of casts to downstream code to support monomorphic shapes in a polymorphic context. (As in, I kind of think it's a bug if you give me a value that could have been a Python `int` but is instead a `Poly`.)

gnecula:
What is happening here is that the shapes are leaking into array computation, e.g.:
```
  receptive_field_size = prod(shape) / shape[in_axis] / shape[out_axis]
  fan_in = shape[in_axis] * receptive_field_size
  fan_out = shape[out_axis] * receptive_field_size
```

I think that this is a part of the masking transformation that is not quite done yes. I think that this will require some programmer intervention. 

This workaround works only when the polynomial is actually a constant, and will actually break down the line when you really want the batch dimension to be variable. A somewhat better workaround might be for the polynomial rules to produce actual ints for constant polynomials. 

This is on top of @mattjj's mind as well

mattjj:
I think if we're careful we can make sure that any time a constant Poly is produced it's turned into an int. That way we can never leak constant Polys. I even think I made that change on a branch somewhere...

jheek:
Yes another solution would be to avoid construction of constant Poly instances. @mattjj is there a reason for not just leaving constant dimensions to be ints? I guess it would require some explicit downcasting in the `Poly` logic like `Poly(a * x) / Poly(x) => int(a)`.

Also is there already an api to evaluate a shape? Because that should solve the problem here more generally (for non-constant) Poly shapes.

mattjj:
Yeah I think that's roughly what I have in mind too: don't promote into Polys if we don't have to, and lower out of them into ints whenever we can. I don't have an example but I suspect the latter may be necessary in some cases; maybe it's not!

There are APIs to evaluate shapes (see `masking.shape_as_value` and `masking.shape_as_padded_value`) but these things are very much in flux.

jheek:
I changed to  PR to avoid literal promotion instead. It does not demote Poly division that results in a constant

mattjj:
I pushed a commit that does the demotion back to ints when each shape rule is applied, so that whenever we produce a new MaskTracer in process_primitive we make shapes as concrete as possible.

We should really add a test, but I consider this a quick fix/workaround pending more revisions to `mask` anyway. @jheek if you have a test available please add it, but otherwise I think we should just merge when CI is green.

jekbradbury (approved):

jheek:
Thanks, as for testing I still feel like this is a bit of a workaround. In the end you want this to work even when initializing from actual dynamic shapes. Some flax tests will depend on the int behavior soon though.

PR #4596, opened 2020-10-15
[jax2tf] Implemented conversion for triangular_solve.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1718,6 +1718,30 @@ def _eigh(operand: TfVal, lower: bool):
 
 tf_impl[lax_linalg.eigh_p] = _eigh
 
+def _triangular_solve(a: TfVal, b: TfVal, *, left_side: bool, lower: bool,
+                      transpose_a: bool, conjugate_a: bool,
+                      unit_diagonal: bool):
+  if unit_diagonal:
+    a = (a - tf.linalg.diag(tf.linalg.diag_part(a)) +
```
hawkinsp, line 1725: This would not be correct if there are NaNs on the diagonal.

hawkinsp (commented):
One small nit.

jax/experimental/jax2tf/jax2tf.py
bchetioui: Thanks Peter! `d840c34` should fix this!

hawkinsp (approved):

cghawthorne:
Do you need to remove `lax_linalg.triangular_solve_p` from `tf_not_yet_impl`?

gnecula:
> Do you need to remove `lax_linalg.triangular_solve_p` from `tf_not_yet_impl`?

Good catch. There is a presubmit test that will fail because of this. We just don't run that test in GitHub

bchetioui:
Removed it. Thanks @cghawthorne :)

Issue #4135, opened 2020-08-24
Cumsum slower than NumPy on CPU


dionhaefner:
I noticed that `cumsum` is several times slower than NumPy on 1 CPU, even on large arrays.

`jax.lax.scan` is faster, but only works along the leading axis (#2509).

(this might be connected to #4134)

#### Example script

```python
import jax
jax.config.update('jax_enable_x64', True)

import numpy as np
import jax.numpy as jnp


@jax.jit
def bench_cumsum(arr):
    return jnp.cumsum(arr, axis=0)


@jax.jit
def bench_scan(arr):
    def accumulate(carry, y):
        new_y = y + carry
        return new_y, new_y

    init = jnp.zeros(arr.shape[1:], dtype=arr.dtype)
    return jax.lax.scan(accumulate, init, arr)[1]


def bench_np(arr):
    return np.cumsum(arr, axis=0)


if __name__ == '__main__':
    import time
    shape = (360, 160, 120, 3)

    key = jax.random.PRNGKey(17)
    arr = jax.random.uniform(key, shape=shape, dtype='float64')

    res_cumsum = bench_cumsum(arr).block_until_ready()
    start = time.perf_counter()
    for _ in range(10):
        bench_cumsum(arr).block_until_ready()
    end = time.perf_counter()
    print(f'bench_cumsum: {end-start:.3f}s')

    res_scan = bench_scan(arr).block_until_ready()
    start = time.perf_counter()
    for _ in range(10):
        bench_scan(arr).block_until_ready()
    end = time.perf_counter()
    print(f'bench_scan: {end-start:.3f}s')
    np.testing.assert_allclose(res_cumsum, res_scan)

    arr_np = np.array(arr)
    start = time.perf_counter()
    for _ in range(10):
        bench_np(arr_np)
    end = time.perf_counter()
    print(f'bench_np: {end-start:.3f}s')

```

#### Output

```
$ taskset -c 0 python bench_cumsum.py
bench_cumsum: 6.635s
bench_scan: 0.885s
bench_np: 2.230s
```

mattjj:
Thanks for the excellent benchmarking script! This certainly sounds like a performance bug.

It's true without 64bit mode too; I got these timings with 32bit (commenting out the 64bit-enable liine):

```
bench_cumsum: 3.366s
bench_scan: 0.183s
bench_np: 0.749s
```

My timings with 64bit were similar to yours (running the original script):

```
bench_cumsum: 7.550s
bench_scan: 0.382s
[...]

PR #4603, opened 2020-10-16
Optimize lax.associative_scan, reimplement cumsum, etc. on top of associative_scan.

hawkinsp:
Add support for an axis= parameter to associative_scan.

We previously had two associative scan implementations, namely lax.associative_scan, and the implementations of cumsum, cumprod, etc.

lax.associative_scan was more efficient in some ways because unlike the cumsum implementation it did not pad the input array to the nearest power of two size. This appears to have been the cause of a significant part of #4135.

The cumsum/cummax implementation used slightly more efficient code to slice and
interleave arrays, which this change adds to associative_scan as well. Since we
are now using lax primitives that make it easy to select an axis, add support
for user-chosen scan axes as well.

We can also simplify the implementation of associative_scan: one of the
recursive base cases seems unnecessary, and we can simplify the code by removing
it.

Benchmarks from #4135 on my workstation:
Before:
bench_cumsum: 0.900s
bench_associative_scan: 0.597s
bench_scan: 0.359s
bench_np: 1.619s

After:
bench_cumsum: 0.435s
bench_associative_scan: 0.435s
bench_scan: 0.362s
bench_np: 1.669s

Before, with taskset -c 0:
bench_cumsum: 1.989s
bench_associative_scan: 1.556s
bench_scan: 0.428s
bench_np: 1.670s

After, with taskset -c 0:
bench_cumsum: 1.271s
bench_associative_scan: 1.275s
bench_scan: 0.438s
bench_np: 1.673s


mattjj (approved):
LGTM, woohoo awesome!

PR #4606, opened 2020-10-16
[jax2tf(js)] Convertible QuickDraw model using Linen

bchetioui:
This builds on #4522, #4562 and #4589 to allow converting a Flax model into TF.js for inference.
Just opening as draft for now, let's see what exactly we do with this.


jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+
+from jax.config import config # type: ignore
+config.config_with_absl()
+
+from jax.experimental import jax2tf # type: ignore
+jax2tf.jax2tf.ENABLE_TF_CONVOLUTION = True
+
+TfVariable = Any
+
+class ExportWrapper(tf.train.Checkpoint):
```
gnecula, line 38: I would try to not make changes to the ExportWrapper because we expect to make that a library function that just implements: https://www.tensorflow.org/hub/reusable_saved_models

gnecula: Also, I would like to know what is different here.

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+      fn: a function taking two arguments, the parameters and the batch of
+        images.
+      params: the parameters, as a list/tuple/dictionary of np.ndarray, to be
+        used as first argument for `fn`.
+     input_shape: the shape of the second argument of `fn` (except the batch size)
+     batch_sizes: a sequence of batch sizes for which to save the function.
+    """
+    super().__init__()
+    # Convert fn from JAX to TF.
+    self._fn = jax2tf.convert(fn, with_gradient=True)
```
gnecula, line 54: Was it important to add `with_gradient`?

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+    # Create tf.Variables for the parameters.
+    self._params = tf.nest.map_structure(
+        lambda param: tf.Variable(param, trainable=False), params)
+    self.saved_signatures = {}
```
gnecula, line 58: Afaik, `saved_signatures` is not part of the API from `reusable_saved_models`. Did you add it just as a place to store some data for your need, or it actually matters to be passed to `tf.saved_model.save`?

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+    # TODO: hackish? We need a serving_default signature for the conversion
+    # to work properly, but we might be able to change that through some
+    # parameter of the converter.
+    self.saved_signatures["serving_default"] = (
+      self.__call__.get_concrete_function(
```
gnecula, line 63: Isn't this achieved also by the code in line 73?

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+def save_model(fn: Callable, params,
+               model_dir: str, *, input_shape: Sequence[int],
+               batch_sizes: Sequence[int]):
+  """Saves the SavedModel for a function"""
+  print(jax.make_jaxpr(fn)(params, np.ones([1] + list(input_shape))))
+  wrapper = ExportWrapper(
+    fn, params, input_shape,
+    batch_sizes=batch_sizes)
+  print(f"Saving the model to {model_dir}")
+  tf.saved_model.save(wrapper, model_dir, signatures=wrapper.saved_signatures)
```
gnecula, line 90: If this is the only place where we use `saved_signatures` then I think it is better to not have them in the ExportWrapper.

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+      train_acc = FlaxQuickDraw.accuracy(
+          FlaxQuickDraw.predict, optimizer.target, train_ds)
+      test_acc = FlaxQuickDraw.accuracy(
+          FlaxQuickDraw.predict, optimizer.target, test_ds)
+      print("Epoch {} in {:0.2f} sec".format(epoch, epoch_time))
+      print("Training set accuracy {}".format(train_acc))
+      print("Test set accuracy {}".format(test_acc))
+
+    return (FlaxQuickDraw.predict, optimizer.target,
+            FlaxQuickDraw.feature_extractor,
```
gnecula, line 200: We don't need the feature extractor and its params, let's inline the feature extractor in the model.

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+  wrapper = ExportWrapper(
+    fn, params, input_shape,
+    batch_sizes=batch_sizes)
+  print(f"Saving the model to {model_dir}")
+  tf.saved_model.save(wrapper, model_dir, signatures=wrapper.saved_signatures)
+
+# The code below is an adaptation for Flax from the work published here:
+# https://blog.tensorflow.org/2018/07/train-model-in-tfkeras-with-colab-and-run-in-browser-tensorflowjs.html
+
+class FlaxQuickDraw:
```
gnecula, line 95: This class existed in the other notebook because I had two experiments that I wanted to keep separate. I think that here we can just move everything to the top-level

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+        {'params': extractor_params}, inputs)
+
+  class Module(nn.Module):
+    """The whole model."""
+    def setup(self):
+      self.feature_extractor = FlaxQuickDraw.FeatureExtractorModule()
+
+    @nn.compact
+    def __call__(self, x):
+      x = self.feature_extractor(x)
```
gnecula, line 132: Just inline here the feature extractor module

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+      start_time = time.time()
+      for inputs, labels in train_ds:
+        optimizer = jax.jit(FlaxQuickDraw.update)(optimizer, inputs, labels)
+      epoch_time = time.time() - start_time
+      train_acc = FlaxQuickDraw.accuracy(
+          FlaxQuickDraw.predict, optimizer.target, train_ds)
+      test_acc = FlaxQuickDraw.accuracy(
+          FlaxQuickDraw.predict, optimizer.target, test_ds)
+      print("Epoch {} in {:0.2f} sec".format(epoch, epoch_time))
+      print("Training set accuracy {}".format(train_acc))
```
gnecula, line 196: To what accuracy were you able to train it? and how long it took?

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
```diff
@@ -0,0 +1,278 @@
...
+      print("Test set accuracy {}".format(test_acc))
+
+    return (FlaxQuickDraw.predict, optimizer.target,
+            FlaxQuickDraw.feature_extractor,
+            optimizer.target["feature_extractor"])
+
+def download_dataset(dir_path=os.path.join(os.path.dirname(__file__), "data")):
```
gnecula, line 203: This has to be called before? How long does the download take and how big is the data set?

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: It is solely to pass them to `tf.saved_model.save`, but they can be passed manually as you saw below. So it is perfectly fine to store them outside of the `ExportWrapper`.

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: It is! Only that you then need to explicitly pass a `signature_def` argument to `convert_tf_saved_model` to tell which signature to use, as it looks for one called `serving_default` otherwise

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: It depends on how many classes we want to use. The example had 100, with each data file roughly ~100 to 200 Mb (so 10 to 20 Gb total if we stick to this number of classes)

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: Yes, as `PreventGradient` is not a valid op for the conversion!

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: I undid the differences I made. The only difference that is important is that `with_gradient` must be set to True, so as to avoid having `PreventGradient` ops in the graph (as they are not supported by Tensorflow JS)

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: After getting the whole dataset, I can say: the download is not *that* long (of course, it depends on your connection speed), and the considered subset of 100 classes is roughly 11 Gb. How many classes we want to work with (and thus how many files we want to download) can be adjusted using the `NB_CLASSES` global variable in `sketcher.py`.

jax/experimental/jax2tf/tfjs_demo/Sketcher/utils.py
```diff
@@ -0,0 +1,152 @@
...
+    self._fn = jax2tf.convert(fn, with_gradient=with_gradient)
+    # Create tf.Variables for the parameters.
+    self._params = tf.nest.map_structure(
+        # If with_gradient=False, we mark the variables behind as non-trainable,
+        # or else the Keras model below fails for trying to access them
+        # (even with hub.KerasLayer(..., trainable=False), surprisingly).
+        lambda param: tf.Variable(param, trainable=with_gradient),
+        params)
+    self._signatures: Dict[str, Callable] = {}
+
```
bchetioui, line 52: @gnecula this was part of the `ExportWrapper`, but I don't see it mentioned in the spec. Is it needed?

gnecula: I do not think so.

bchetioui: Thank you, I removed it.

jax/experimental/jax2tf/tfjs_demo/Sketcher/sketcher.py
bchetioui: Took roughly 27 min to train it with 77% accuracy (where we consider a prediction to be accurate if the expected label is the most likely prediction, which is different from the original TF example which considers a prediction accurate if the expected label is one of the 5 most likely).

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,147 @@
...
+    print("Test set accuracy {}".format(test_acc))
+
+  return optimizer.target
+
+def main(*args):
+  dir_path = os.path.join(os.path.dirname(__file__), "data")
+  classes = download_dataset(dir_path, NB_CLASSES)
+  assert len(classes) == NB_CLASSES, classes
+  print(f"Classes are: {classes}")
+  flax_params = train(dir_path, classes)
```
gnecula, line 133: nit: What do you think about loading the classes here, and calling `train` with the data sets? Then all of the interaction with `dir_path` is in the `main` function and we just pass data around in the other functions.

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,147 @@
...
+
+  model_dir = os.path.join(os.path.dirname(__file__), "saved_models")
```
gnecula, line 135: I think that it would be better to use a temporary directory for saving the models. Saving next to the sources will never work when run inside Google. I suggest either hard-coding /tmp/jax2tf/tf_js_quickdraw or `tempfile`.

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
```diff
@@ -0,0 +1,8 @@
+# JAX2TF.js Quickdraw example
+
+This example contains the code necessary to train a CNN on the Quickdraw
+dataset using Flax, and subsequently convert it using JAX2TF into a model that
+can be loaded in TF.js.
+
```
gnecula, line 6: Can you please add some text about what the user is supposed to run, and what to look for? Also perhaps what to do with the tarball and how it fits with the rest of the code here?

jax/experimental/jax2tf/examples/tf_js/quickdraw/utils.py
```diff
@@ -0,0 +1,75 @@
...
+from jax import numpy as jnp # type: ignore
+
+import numpy as np # type: ignore
+import os
+import requests # type: ignore
+
+def download_dataset(dir_path, nb_classes):
+
+  if not os.path.exists(dir_path):
+    os.mkdir(dir_path)
```
gnecula, line 23: I think that `os.makedirs` will create the directory and its parents, if needed.

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,147 @@
...
+from flax import linen as nn
+from flax.training import common_utils # type: ignore
+
+import jax # type: ignore
+from jax import lax
+from jax import numpy as jnp
+from jax.experimental import jax2tf # type: ignore
+jax2tf.jax2tf.ENABLE_TF_CONVOLUTION = True
+
+from jax.experimental.jax2tf.examples.saved_model_lib import save_model # type: ignore
```
gnecula, line 29: I suggest importing modules, not functions:
```
from jax.experimental.jax2tf.examples impot saved_model_lib
```

Then when we see `saved_model_lib.save_model` we know right away what is being called.

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,149 @@
...
+    train_acc = accuracy(predict, optimizer.target, train_ds)
+    test_acc = accuracy(predict, optimizer.target, test_ds)
+    print("Epoch {} in {:0.2f} sec".format(epoch, epoch_time))
+    print("Training set accuracy {}".format(train_acc))
+    print("Test set accuracy {}".format(test_acc))
+
+  return optimizer.target
+
+def main(*args):
+  dir_path = os.path.join(os.path.dirname(__file__), "data")
```
gnecula, line 128: Please change this one to base_model_path/data

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,149 @@
...
+  print(f"Classes are: {classes}")
+  train_ds, test_ds = utils.load_classes(dir_path, classes)
+  flax_params = train(train_ds, test_ds, classes)
+
+  base_model_path = "/tmp/jax2tf/tf_js_quickdraw"
+  model_dir = os.path.join(base_model_path, "saved_models")
+  # the model must be converted with with_gradient set to True to be able to
+  # convert the saved model to TF.js, as "PreventGradient" is not supported
+  input_signatures = [
+    tf.TensorSpec([batch_size, 28, 28, 1]) for batch_size in [1, 256]
```
gnecula, line 140: Do we even need to save for batch_size 256? Do you know how convert_tf_saved_model handles the batch size? It converts only the first input_signature? Converts all (I'd be surprised)?

bchetioui: No, it is indeed useless. It converts the one corresponding to the `serving_default` signature, which I believe would now be the first signature with the new Exporter code you pushed this morning

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
bchetioui: PTAL at the latest README and let me know what you think!

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
```diff
@@ -0,0 +1,38 @@
...
+The training loop runs for 5 epochs, and the model as well as its equivalent
+TF.js-loadable model are subsequently saved into `/tmp/jax2tf/tf_js_quickdraw`.
+
+## Interacting with the model
+
+Here is a (third party) [demo](https://zaidalyafeai.github.io/sketcher/) of how
+interacting with the trained model looks like.
+
+If you want to try it yourself, the trained model along with third party front
```
gnecula, line 30: You said that there is a readme in the tarball. Perhaps it is worth saying here also that the tarball has the same exact files as the above linked website, with one line change to ..., along with the trained weights. Do you happen to have training accuracy numbers?

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,146 @@
...
+
+import flax # type: ignore
+from flax import linen as nn
+from flax.training import common_utils # type: ignore
+
+import jax # type: ignore
+from jax import lax
+from jax import numpy as jnp
+from jax.experimental import jax2tf # type: ignore
+jax2tf.jax2tf.ENABLE_TF_CONVOLUTION = True
```
gnecula, line 27: Let's wait until you land the other mechanism.

jax/experimental/jax2tf/examples/tf_js/quickdraw/quickdraw.py
```diff
@@ -0,0 +1,146 @@
...
+  classes = utils.download_dataset(dataset_path, NB_CLASSES)
+  assert len(classes) == NB_CLASSES, classes
+  print(f"Classes are: {classes}")
+  train_ds, test_ds = utils.load_classes(dataset_path, classes)
+  flax_params = train(train_ds, test_ds, classes)
+
+  model_dir = os.path.join(base_model_path, "saved_models")
+  # the model must be converted with with_gradient set to True to be able to
+  # convert the saved model to TF.js, as "PreventGradient" is not supported
+  saved_model_lib.save_model(predict, flax_params, model_dir,
```
gnecula, line 139: Have you tried this with the now default "convert_model=True"?

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
bchetioui: I will upgrade. Yes, I have training accuracy numbers:
```
Training set accuracy 0.7825493812561035
Test set accuracy 0.764135718345642
```
where "accurate" means "the prediction with the highest likelihood is the expected class".

gnecula: Please update the accuracy numbers in the README

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
```diff
@@ -0,0 +1,38 @@
+# Quickdraw
+
+This directory contains the code necessary to train a CNN on the Quickdraw
+dataset using Flax, and subsequently convert it using jax2tf into a model that
+can be loaded in TF.js.
+
+## Training the model
+
+To train and export the model yourself, simply run
```
gnecula, line 9: Is this step going to take a while? It would be good to tell the user, something like "You can train and export the model yourself, which will take 30 minutes, or you can skip to [next section](#interacting-with-the-model)"

jax/experimental/jax2tf/examples/README.md
```diff
@@ -13,6 +13,10 @@ At the moment, the open-source TensorFlow model server is missing XLA support,
 but the Google version can be used, as shown in the
 [serving examples](https://github.com/google/jax/blob/master/jax/experimental/jax2tf/examples/serving/README.md).
 
+A jax2tf-generated SavedModel can also be converted to a format usable with
+TensorFlow.js in some cases where the conversion does not require XLA support,
+as show in the [Quickdraw TensorFlow.js example](https://github.com/google/jax/blob/master/jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md).
```
gnecula, line 18: as shown

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Training the model
+
+You can train and export the model yourself by running
+```bash
+$ python3 quickdraw.py
+```
+
+The training itself lasts roughly 30 minutes, assuming the dataset has already
+been downloaded. You can also skip to the [next section](https://github.com/google/jax/blob/master/jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md#interacting-with-the-model)
```
gnecula, line 15: nit: In this particular case you can just write [#interacting-with-the-model] because that works both in github and in Google.

PR #4607, opened 2020-10-16
[jax2tf] Remove the argument type promotion in jax2tf.

gnecula:
This was added as a safety mechanism in the past when jax2tf tracking
of types was imprecise. Now it is not needed anymore.

Also removed the mention that shift operations on TPU may produce
incorrect results.

### DIFF
--- a/jax/experimental/jax2tf/primitives_with_limited_support.md
+++ b/jax/experimental/jax2tf/primitives_with_limited_support.md
@@ -1,6 +1,6 @@
 # Primitives with limited support
 
-*Last generated on (YYYY-MM-DD): 2020-10-14*
+*Last generated on (YYYY-MM-DD): 2020-10-16*
 
 ## Updating the documentation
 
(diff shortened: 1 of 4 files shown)


jax/experimental/jax2tf/primitives_with_limited_support.md
```diff
@@ -58,7 +58,6 @@ conversion to Tensorflow.
 | nextafter | Missing TF support | Primitive is unimplemented in TF | bfloat16, float16 | CPU, GPU, TPU |
 | population_count | Missing TF support | Primitive is unimplemented in TF | uint32, uint64 | CPU, GPU, TPU |
 | qr | Missing TF support | Primitive is unimplemented in TF; this is a problem only in compiled mode (experimental_compile=True)) | complex128, complex64 | CPU, GPU, TPU |
-| reduce_window_sum | Missing TF support | Primitive is unimplemented in TF | uint16, uint32, uint64 | CPU, GPU, TPU |
```
gnecula, line 61: @bchetioui Do you know why this was removed? It seems to still be declared in correctness_stats.

gnecula: I figured it out. I told you for #4589 to make the initial value dynamic in the tests for reduce_window. It turns out that lax.reduce_window will only pick special reducers if the initial  value is a constant. So, it has to be StaticArg. I fixed that and new stuff appeared in the documentation.

bchetioui, line 61: It seems that it was a mistake to remove `StaticArg` around the `init_value` in the `reduce_window` harness maker, as the `reduce_window_{min,max,sum}` specialized paths are not tested anymore, and are thus dropped out of this list.

bchetioui: Oh, see my response below, we were looking into the same thing :)

bchetioui (approved):
LGTM! This `reduce_window` testing issue makes me think that it'd be nice to extract the general list of limitations from the classifier, and check how many times a limitation is hit. We could detect which limitations are documented but not properly tested (like happened here), but also possibly detect cases that are tested several times & thus potential opportunities to remove some of our tests.

Just a thought, it may not be super useful yet, but if we end up having too many tests (as you said might become the case), it may be worth doing something like this.

Issue #4615, opened 2020-10-16
Print-debugging jax?


johnsalmon:
I've got some jax code that relies on grad, jvp, etc.  I'd like to do some simple print-style debugging.  E.g., some lines like:

    print("JUtotal: ", ret)
    return ret

In a function whose gradient is required.  I'd really like to see something like:

`JUtotal: -0.924010929514282`

but instead, I see:

```
JUtotal:  Traced<ConcreteArray(-0.9240109029514282)>with<JVPTrace(level=6/0)>
  with primal = Traced<ConcreteArray(-0.9240109029514282)>with<JVPTrace(level=4/0)>
                  with primal = DeviceArray(-0.9240109, dtype=float64)
                       tangent = Traced<ShapedArray(float64[])>with<BatchTrace(level=3/0)>
                                   with val = DeviceArray([ 4.31834536e-04, -5.96999899e-06,  1.25216808e-01,
                                                           -1.07182239e-01, -7.46824003e-07, -8.62923763e-07],            dtype=float64)
                                        batch_dim = 0
       tangent = Traced<ShapedArray(float64[]):JaxprTrace(level=5/0)>
```

I understand why the latter is sometimes important (e.g., when debugging the autograd machinery itself).  But for now, I just want to know the numerical value of 'ret'.  Is there a way to obtain just the numerical value of an object in a traced function with all the extra "hair" removed?

shoyer:
I recommend using `jax.experimental.host_callback` for this -- see `id_print()` in particular: https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html

mattjj:
It fills my heart with joy to see you're still playing with JAX, @johnsalmon !

`id_print` is one option, though its primary purpose is being able to perform prints from code that was staged out to XLA via `jit` (or `pmap`, or control flow primitives). To reach that goal it might not be as ergonomic as Python printing, like you might want in interactive pdb sessions.

The text you see right now is from [the `Tracer.__repr__` method defined in core.py](https://github.com/google/jax/blob/7e801cae3a35d9fb76705f0c23aa715dc6ab9b77/jax/core.py#L538-L544), which is meant to recursively dissect a Tracer and print everything it finds. But as you say, maybe that can be too much information for some common use cases.

[...]

PR #4618, opened 2020-10-16
add a Tracer.__str__ for simpler printing

mattjj:
```
In [1]: def f(x):
   ...:     print(x)
   ...:     print(repr(x))
   ...:     return x
   ...:

In [2]: from jax import grad
gra
In [3]: grad(f)(3.14)
Traced<3.14>
Traced<ConcreteArray(3.14, weak_type=True)>with<JVPTrace(level=2/0)>
  with primal = Traced<ConcreteArray(3.14, weak_type=True):JaxprTrace(level=1/0)>
       tangent = Traced<ShapedArray(float32[], weak_type=True):JaxprTrace(level=1/0)>
Out[3]: array(1.)

In [4]: from jax import jit

In [5]: jit(f)(3.14)
Traced<float32[]>
Traced<ShapedArray(float32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
Out[5]: DeviceArray(3.14, dtype=float32)

In [6]: grad(grad(f))(3.14)
Traced<3.14>
Traced<ConcreteArray(3.14, weak_type=True)>with<JVPTrace(level=4/0)>
  with primal = Traced<ConcreteArray(3.14, weak_type=True):JaxprTrace(level=3/0)>
       tangent = Traced<ShapedArray(float32[], weak_type=True):JaxprTrace(level=3/0)>
Out[6]: array(0., dtype=float32)
```

Experiments for #4615

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -543,6 +543,14 @@ class Tracer:
                                     for name, pp_payload in contents)
     return str(base)
 
+  def __str__(self):
+    if isinstance(self.aval, ConcreteArray):
+      return f'Traced<{self.aval.val}>'
+    elif isinstance(self.aval, (ShapedArray, AbstractUnit)):
+      return f'Traced<{self.aval.str_short()}>'
+    else:
+      return repr(self)
+
   def _contents(self):
     try:
       return [(name, pp(repr(getattr(self, name)))) for name in self.__slots__]


mattjj:
I think the upside here is not worth the potential surprise of having repr differ from str. Maybe we can address the original issue in some other way, like by providing some print utility functions.

PR #4008, opened 2020-08-10
fix custom_jvp/vjp closure issues, and nondiff_argnums too!

mattjj:
fixes #4566, fixes #4521, fixes #2912, fixes #3822, fixes #4173, fixes #2520, fixes #3808 

If you survey those issues, you'll come to the conclusion that `jax.custom_jvp` and `jax.custom_vjp` don't work with closing over JAX Tracers, and moreover that `nondiff_argnums` can be just as problematic. Indeed, `nondiff_argnums` was effectively implemented in terms of lexical closure. The original plan was not to make `custom_jvp`/`custom_vjp` work with lexical closure, but I never documented that, and the error messages were terrible!

This PR makes `custom_jvp` and `custom_vjp` work with closed-over Tracers. Woo! That is, now `custom_jvp` and `custom_vjp` functions and rules can close over Tracers to our hearts' content. For all non-autodiff transformations, things will Just Work. For autodiff transformations, we'll get a clear error message about why we can't differentiate with respect to values over which a custom_jvp or custom_vjp closes:
> Detected differentiation of a custom_jvp function with respect to a closed-over value. That isn't supported because the custom JVP rule only specifies how to differentiate the custom_jvp function with respect to explicit input parameters. Try passing the closed-over value into the custom_jvp function as an argument, and adapting the custom_jvp rule.

This PR accomplishes that goal by following through on the original _ansatz_ of `custom_jvp` and `custom_vjp`, namely to make them work like `core.call`. That is, we do all the `core.process_env_traces` stuff properly.

TODO:
- [x] make pre-omnistaging version work again
- [x] fix jax2tf sublevels interaction
- [x] fix all google internal users
- [x] update custom derivatives notebook with new custom_vjp nondiff_argnums advice
- [x] write a pr message

PR #4620, opened 2020-10-16
Add default implementation of process_custom_jvp_call process_custom_vjp_call to `jax.experimental.callback`

sharadmv:
As of #4008, there are no more default implementations of `process_custom_jvp_call` and `process_custom_vjp_call`. This results in an error whenever a function that has a custom jvp/vjp appears in a `CallbackTrace`.

This PR adds the default implementation, though it should probably preserve the custom info, since `CallbackTrace` keeps most primitive the same. We can iterate on the initial implementation below, if that is the best way forward.


mattjj (approved):
LGTM!

PR #4625, opened 2020-10-17
Make pad more robust to non-hashable values.

jblespiau:
The `pad_width` can be an iterable of iterbles, thus, we need to cast both the inner and outer list to tuples.
Also, the values should be intergers, but some use-cases (mostly Tensorflow) pass e.g. DeviceArray(1) as values in there.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -748,13 +748,13 @@ def correlate(a, v, mode='valid', *, precision=None):
 
 
 def _normalize_float(x):
-    info = finfo(_dtype(x))
-    cond = lax.abs(x) < info.tiny
-    x1 = where(cond, x * (1 << info.nmant), x)
-    x2 = where(cond,
-               full_like(x, -info.nmant, dtype=np.int32),
-               zeros_like(x, dtype=np.int32))
-    return lax.convert_element_type(x1, _dtype(x)), x2
+  info = finfo(_dtype(x))
+  cond = lax.abs(x) < info.tiny
+  x1 = where(cond, x * (1 << info.nmant), x)
+  x2 = where(cond, full_like(x, -info.nmant, dtype=np.int32),
+             zeros_like(x, dtype=np.int32))
+  return lax.convert_element_type(x1, _dtype(x)), x2
+
 
 _INT_DTYPES = {
   16: np.int16,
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2324,10 +2325,12 @@ def _pad(array, pad_width, mode, constant_values):
     msg = "Unimplemented padding mode '{}' for np.pad."
     raise NotImplementedError(msg.format(mode))
 
+
 @_wraps(np.pad)
-def pad(array, pad_width, mode='constant', constant_values=0):
-  if isinstance(pad_width, list):
-    pad_width = tuple(pad_width)
+def pad(array, pad_width, mode="constant", constant_values=0):
```
jblespiau, line 2330: This is the change, the rest is only identication fixes.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2324,10 +2325,12 @@ def _pad(array, pad_width, mode, constant_values):
...
+  pad_width = tuple((tuple(int(i)
```
hawkinsp, line 2331: Formatting nit: I'd break the line before `if` not before `for`.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2324,10 +2325,12 @@ def _pad(array, pad_width, mode, constant_values):
...
+                           for i in x) if not isinstance(x, int) else x
```
hawkinsp, line 2332: Wouldn't it be simpler just to write:
`tuple(tuple(int(i) for i in x) for x in pad_width)`

Why not unconditionally convert to int?

hawkinsp: Oh, I see, numpy.pad accepts ints and tuples. This is fine.

hawkinsp:
You have some test failures as well.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2324,10 +2325,13 @@ def _pad(array, pad_width, mode, constant_values):
...
     raise NotImplementedError(msg.format(mode))
 
+
 @_wraps(np.pad)
-def pad(array, pad_width, mode='constant', constant_values=0):
-  if isinstance(pad_width, list):
-    pad_width = tuple(pad_width)
+def pad(array, pad_width, mode="constant", constant_values=0):
+  if isinstance(pad_width, Sequence):
+    pad_width = tuple(  # new line
```
hawkinsp, line 2332: Sometimes I've just done this via a trailing `#` with no text. (Note we don't use pyformat, so you can also just *not* use it here...)

hawkinsp (approved):

PR #4639, opened 2020-10-19
Add fast path for jnp.repeat when repeats is a scalar constant.

petebu:
The default code-path generates a complicated soup of ops for this very simple case what can be implemented using reshapes and broadcasts only.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2959,6 +2959,19 @@ def repeat(a, repeats, axis=None, *, total_repeat_length=None):
       "When jit-compiling jnp.repeat, the total number of repeats must be static. "
       "To fix this, either specify a static value for `repeats`, or pass a static "
       "value to `total_repeat_length`.")
+
+    # Fast path for when repeats is a scalar.
+    if np.ndim(repeats) == 0 and ndim(a) != 0:
+      input_shape = a.shape
+      aux_axis = axis if axis < 0 else axis + 1
+      a = expand_dims(a, aux_axis)
+      reps = [1] * len(a.shape)
+      reps[aux_axis] = repeats
+      a = tile(a, reps)
+      result_shape = list(input_shape)
+      result_shape[axis] *= repeats
+      return reshape(a, result_shape)
+
     repeats = np.ravel(repeats)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2883,6 +2883,16 @@ def repeat(a, repeats, axis=None, *, total_repeat_length=None):
       "To fix this, either specify a static value for `repeats`, or pass a static "
       "value to `total_repeat_length`.")
     repeats = np.ravel(repeats)
+
+    # Special case when repeats is a scalar.
+    if ndim(repeats) == 0:
+      input_shape = a.shape
+      a = reshape(a, np.insert(input_shape, axis, 1))
```
hawkinsp, line 2890: `lax.expand_dims`

?

hawkinsp (changes requested):

jax/_src/numpy/lax_numpy.py
mattjj: +1

mattjj (approved):
Thanks for this improvement!

+1 to hawkinsp@'s comment.

Consider this optional, but since all untested code is (eventually) broken, I wonder how we might add a test for this. Maybe we could just call `make_jaxpr` and check the number of eqns is only ~3?

jax/_src/numpy/lax_numpy.py
petebu: That's better. Thanks!

petebu:
> Thanks for this improvement!
> 
> +1 to hawkinsp@'s comment.
> 
> Consider this optional, but since all untested code is (eventually) broken, I wonder how we might add a test for this. Maybe we could just call `make_jaxpr` and check the number of eqns is only ~3?

Good idea. In fact, the original code was wrong but wasn't being called so the tests passed anyway. I've just committed a fix and added a test for the number of eqns.

petebu:
Before this PR, `jnp.repeat(jnp.array([1,2,3,4]), repeats=2)` generated 21 Jaxpr eqns. Now it's down to 6.

hawkinsp (approved):

mattjj (approved):
Extra LGTM for an extra good PR.

PR #4641, opened 2020-10-19
Add deprecation warning for indexing with non-tuple sequences

### LINKED ISSUE #4564: Should JAX deprecate indexing with lists?
Since numpy 1.16, indexing with a list in place of a tuple has led to a `FutureWarning` (See https://github.com/numpy/numpy/pull/9686 for a discussion of the rationale for this):
```python
>>> import numpy as np
>>> x = np.arange(6).reshape(2, 3)
>>> idx = [[0], [1]]
>>> x[idx]
FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
array([1])
```
As mentioned in the warning, the current behavior treats the indices as identical to a tuple:
```python
>>> x[tuple(idx)]
array([1])
```
while in the future, the indices will be treated as an array:
```python
>>> x[np.array(idx)]
array([[[0, 1, 2]],
       [[3, 4, 5]]])
```

JAX currently implements the old, deprecated behavior, without any warning:
```python
>>> import jax.numpy as jnp
>>> jnp.array(x)[idx]
DeviceArray([1], dtype=int32)
```

This is setting us up for a future where numpy and JAX have different indexing semantics for lists of indices. I would propose that we follow numpy and start warning about this behavior now, so that when a numpy release finally does deprecate this indexing behavior, jax will be ready to immediately follow suit.

Thoughts?

jakevdp:
Fixes #4564 

The goal here is to prevent JAX indexing from giving different results than numpy indexing in future numpy versions.

Note that with this change, JAX will raise a warning in some cases that numpy does not; e.g.
```python
x = jnp.arange(6).reshape(3, 2)
idx = [0, 1]
x[idx]
```
This PR is setting up JAX to be **more** strict than numpy, in that it will always error on non-tuple sequences of indices. This is consistent with JAX's behavior in other settings, where we avoid silent promotion of lists to arrays within `jax.numpy` functions.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4232,10 +4232,28 @@ def _eliminate_deprecated_list_indexing(idx):
   # deprecated by NumPy and exists for backward compatibility.
   if not isinstance(idx, tuple):
     if isinstance(idx, Sequence) and not isinstance(idx, ndarray):
+      # As of numpy 1.16, some non-tuple sequences of indices result in a warning, while
+      # others are converted to arrays, based on a set of somewhat convoluted heuristics
+      # (See https://github.com/numpy/numpy/blob/v1.19.2/numpy/core/src/multiarray/mapping.c#L179-L343)
+      # In JAX, we raise a warning for *all* non-tuple sequences, and in the future will
+      # *always* raise a TypeError here, rather than silently converting to an array or tuple
+      # depending on the contents of the list as numpy will. "Explicit is better than implicit".
+      # TODO(jakevdp): raise a TypeError here.
       if _any(_should_unpack_list_index(i) for i in idx):
+        msg = ("Using a non-tuple sequence for multidimensional indexing is deprecated; "
+               "use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will "
+               "result in a TypeError. See https://github.com/google/jax/issues/4564 "
+               "for discussion of why this type of indexing is being deprecated.")
         idx = tuple(idx)
       else:
+        msg = ("Using a non-tuple sequence for multidimensional indexing is deprecated; "
+               "use `arr[array(seq)]` instead of `arr[seq]`. In the future this will "
+               "result in a TypeError. See https://github.com/google/jax/issues/4564 "
+               "for discussion of why this type of indexing is being deprecated.")
         idx = (idx,)
+      # TODO(jakevdp): this stacklevel is appropriate for x[idx]; for ops.index_update
+      # we should use stacklevel=5; for x.at[idx].set() we should use stacklevel=6.
+      warnings.warn(msg, FutureWarning, stacklevel=4)
     else:
       idx = (idx,)
   return idx
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -3696,7 +3696,7 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
   def testIssue776(self):
     """Tests that the scatter-add transpose rule instantiates symbolic zeros."""
     def f(u):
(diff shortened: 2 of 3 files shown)


tests/lax_numpy_test.py
```diff
@@ -3696,7 +3696,7 @@ def testIssue764(self):
   def testIssue776(self):
     """Tests that the scatter-add transpose rule instantiates symbolic zeros."""
     def f(u):
-      y = jax.ops.index_add(np.ones(10,), [2, 4, 5], u)
+      y = jax.ops.index_add(np.ones(10,), np.array([2, 4, 5]), u)
```
hawkinsp, line 3699: I'd actually suggest this should be `jax.ops.index[2, 4, 5]`.

It's never been intentional that you can pass an array here...

jax/_src/numpy/lax_numpy.py
```diff
@@ -4232,10 +4232,26 @@ def _eliminate_deprecated_list_indexing(idx):
...
     if isinstance(idx, Sequence) and not isinstance(idx, ndarray):
+      # As of numpy 1.16, some non-tuple sequences of indices result in a warning, while
+      # others are converted to arrays, based on a set of somewhat convoluted heuristics
+      # (See https://github.com/numpy/numpy/blob/v1.19.2/numpy/core/src/multiarray/mapping.c#L179-L343)
+      # In JAX, we raise a warning for *all* non-tuple sequences, and in the future will
+      # *always* raise a TypeError here, rather than silently converting to an array or tuple
+      # depending on the contents of the list as numpy will. "Explicit is better than implicit".
+      # TODO(jakevdp): raise a TypeError here.
       if _any(_should_unpack_list_index(i) for i in idx):
+        msg = ("Using a non-tuple sequence for multidimensional indexing is deprecated; "
```
hawkinsp, line 4243: Would it make sense to link to any NumPy deprecation plans?

hawkinsp (approved):

tests/lax_numpy_test.py
jakevdp: I think that's something different, right? In numpy it would be akin to
```python
x[2, 4, 5] += u
```
which results in an index error for 1D arrays.

The intent here is broadcasting indexing along one dimension; something like:
```python
idx = np.array([2, 4, 5])
x[idx] += u
```

jax/_src/numpy/lax_numpy.py
jakevdp: Initially I'd thought that, but given JAX's relative flexibility compared to numpy, I think it would be beneficial to deprecate list-based indices before numpy does. Then there will never be a time when the latest JAX and the latest numpy return different results for the same indexing operation.

tests/lax_numpy_test.py
hawkinsp: Then it probably should be `jax.ops.index[np.array([2, 4, 5])]`.

Although perhaps
`x.at[np.array([2, 4, 5])].set(y)` is even better?

(We should probably prefer that syntax everywhere.)

jax/_src/numpy/lax_numpy.py
hawkinsp: I mean not so much the concrete timeline, but more something that lays out the rationale for why it is to be deprecated (either our doc or theirs.)

tests/lax_numpy_test.py
jakevdp: Done.

jax/_src/numpy/lax_numpy.py
jakevdp: Done.

hawkinsp (approved):

PR #4642, opened 2020-10-19
A couple of typo/gap fixes in PRNG design notes

lmmx:
I hope ‘tidying’ PRs are welcome. While reading the design notes (in a terminal as I find it easier to read there than in a browser) I noticed a few mistakes in the doc on PRNGs, and fixed them:

1. [invalid Python syntax] `global` is a statement separate from the assignment of the variable it references
2. empty line in code section (not displayed but fixed anyway while I’m at it)
3. gap in the sentence “We use the key together with the hash function to implement ___: that is, splitting is a way to generate two new keys from an existing one.”
    - I wasn’t completely sure what this gap should be, so I went with “splittable PRNGs”, but please change this if it should be something else! In the absence of any text here, the link was invisible in the docs page
4. [invalid Python syntax] you cannot put a comment in the body of a definition, you get an IndentationError. You can however use the [Ellipsis literal](https://docs.python.org/3/library/constants.html#Ellipsis) `...` (treated like a `pass` statement but indicates a TODO)

First contribution so CLA submitted as per contributing guidelines.

### DIFF
--- a/design_notes/prng.md
+++ b/design_notes/prng.md
@@ -29,7 +29,8 @@ def foo(): return bar() + baz()
 def bar(): return rand(RNG, (3, 4))
 def baz(): return rand(RNG, (3, 4))
 def main():
-  global RNG = RandomState(0)
+  global RNG
+  RNG = RandomState(0)
   return foo()
 ```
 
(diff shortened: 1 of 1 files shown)


jakevdp (approved):
Looks good, thanks!

PR #4644, opened 2020-10-19
Also convert numpy arrays to hashable values.

jblespiau:
This was working for a combination of lists and tuples, but the user can also pass a numpy array. Iterable is the correct type here, not Sequence.


mattjj (approved):
LGTM!

mattjj:
Can you remedy that test failure though?

jblespiau:
Done.

Issue #4558, opened 2020-10-13
factor named_call primitive into jax core


froystig:
Both [Flax](https://github.com/google/flax/blob/717d50489d46efb92fa846878d51c4c06936b7cd/flax/core/named_call.py) and [Haiku](https://github.com/deepmind/dm-haiku/blob/1909345c550f56825e8c8aa42621c8a307741ac8/haiku/_src/named_call.py) implement a `named_call` JAX primitive. How about we move this into JAX core?

This is the same as JAX's call primitive (see `jax.core.call_p`), except in how it compiles to XLA, where the name that it carries is included in the HLO computation name. This makes it useful for annotating profiles.

cc @LenaMartens @tomhennigan @trevorcai @levskaya @jheek @avital

levskaya:
I'd be a big fan of doing this!  I tried suggesting it before, but I think people were too busy to think about it at the time.  It would be much better having such a simple, basic primitive live in a single place in JAX.

shoyer:
Yes, please!

froystig:
As I understand it, there might be differences between what we'd upstream to JAX and what's currently in Haiku, namely due to Haiku's state threading (?). The Flax version is roughly what I had in mind originally.

qiuminxu:
The stateful_named_call in haiku looks like a user of the named_call_p that doesn't need to be upstreamed. I attempted a draft in PR #4733 .

froystig:
closed by #4733

PR #4649, opened 2020-10-20
Preserve name scope of named_call primitive in jax2tf

qiuminxu:
named_call primitive has been used in Flax to add higher level model information.  The namescope information has not been preserved in jax2tf converter. 

This PR looks for the named_call primitive and uses tf.name_scope to pass the name scope information to the converted tensorflow ops. The PR also sanitizes the names to valid tensorflow name scopes.

screenshot before:
Namescope not perserved after jax2tf conversion.
namescope_before

screenshot after:
Profiler traces shows namescopes from named_call primitive.
namescope_after

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -753,8 +753,7 @@ class TensorFlowTrace(core.Trace):
         vals_out: Sequence[Tuple[TfVal,
                                  core.AbstractValue]] = f.call_wrapped(*vals)
     else:
-      vals_out: Sequence[Tuple[TfVal,
-                               core.AbstractValue]] = f.call_wrapped(*vals)
+      vals_out = f.call_wrapped(*vals)
     return [TensorFlowTracer(self, v, a) for v, a in vals_out]
 
   def post_process_call(self, call_primitive: core.Primitive,
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Experimental module transforms JAX functions to be executed by TensorFlow."""
 import functools
+import re
 import string
 from typing import Any, Callable, Dict, Iterable, Optional, Sequence, Tuple, Union
 
@@ -44,6 +45,18 @@ from tensorflow.compiler.xla import xla_data_pb2  # type: ignore[import]
 
 from jaxlib import xla_client
 
+
+# The scope name need to be a valid TensorFlow name. See
+# https://github.com/tensorflow/tensorflow/blob/r2.3/tensorflow/core/framework/node_def_util.cc#L731
+_VALID_SCOPE_REGEX = re.compile("^[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*$")
+_INVALID_SCOPE_CHAR = re.compile("[^A-Za-z0-9_.\\/>-]")
+
+
+def _sanitize_scope_name(name):
+  scope_name = _INVALID_SCOPE_CHAR.sub("_", name)
+  if not _VALID_SCOPE_REGEX.match(scope_name):
+    scope_name = ".{}".format(scope_name)
+  return scope_name
 
 # A value suitable in a TF tracing context: tf.Tensor, tf.Variable,
 # or Python scalar or numpy.ndarray. (A tf.EagerTensor is a tf.Tensor.)
@@ -735,7 +748,13 @@ class TensorFlowTrace(core.Trace):
     assert call_primitive.multiple_results
     vals: Sequence[TfVal] = [t.val for t in tracers]
     f = _interpret_subtrace(f, self.main, tuple(t.aval for t in tracers))
-    vals_out: Sequence[Tuple[TfVal, core.AbstractValue]] = f.call_wrapped(*vals)
+    if call_primitive == core.named_call_p:
+      with tf.name_scope(_sanitize_scope_name(params["name"])):
+        vals_out: Sequence[Tuple[TfVal,
+                                 core.AbstractValue]] = f.call_wrapped(*vals)
+    else:
+      vals_out: Sequence[Tuple[TfVal,
+                               core.AbstractValue]] = f.call_wrapped(*vals)
     return [TensorFlowTracer(self, v, a) for v, a in vals_out]
 
   def post_process_call(self, call_primitive: core.Primitive,
--- a/jax/experimental/jax2tf/tests/jax2tf_test.py
+++ b/jax/experimental/jax2tf/tests/jax2tf_test.py
@@ -410,6 +410,18 @@ class Jax2TfTest(tf_test_util.JaxToTfTestCase):
         ValueError, "convert must be used outside all JAX transformations"):
       self.TransformConvertAndCompare(outer, np.ones((4,)), transform)
 
+  def test_name_scope(self):
+    log = []
+
+    @jax.named_call
+    def my_test_function(x):
+      y = tf.Variable(1., name="foo")
+      log.append(y.name)
+      return x * x
+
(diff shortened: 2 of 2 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -539,7 +551,12 @@ def process_call(self, call_primitive: core.Primitive, f: lu.WrappedFun,
     assert call_primitive.multiple_results
     vals: Sequence[TfVal] = [t.val for t in tracers]
     f = _interpret_subtrace(f, self.main, tuple(t.aval for t in tracers))
-    vals_out: Sequence[Tuple[TfVal, core.AbstractValue]] = f.call_wrapped(*vals)
+    if call_primitive.name == "named_call":
```
tomhennigan, line 554: It feels bad to anchor on the string name of a primitive. For named_call in particular I think there are plans to upstream the implementation to JAX core (there are equivalent implementations in Haiku and Flax). I suggest we do that first and then your test can be `if call_primitive == lax.named_call_p`.

cc @LenaMartens

qiuminxu: I'm happy to wait for the upstream the implementation to JAX core. Is there already a PR or I'm also happy to help with the upstreaming. @levskaya @gnecula

froystig: Upstreaming is tracked as issue #4558. I had assigned it to myself, but at the moment anyone is welcome to take it if they'd like.

qiuminxu: PR #4733 has merged, changed to use if call_primitive == lax.named_call_p.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -46,6 +47,17 @@
 from jaxlib import xla_client
 
 
+# TF scope validation, refer to tensorflow/python/framework/ops.py
```
tomhennigan, line 50: At a guess this xref could get out of date quickly.. Is there a more canonical place (e.g. documentation for what a valid name scope is?). The `ops.py` implementation also seems to be branched from `node_def_util.cc:737`. Should we link to that?

Also I've spotted some similar code in other projects, is this (sanitizing a name scope) something we should ask TF team to support (e.g. `tf.name_scope(..)` would do this for you?).

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -734,7 +746,12 @@ def process_call(self, call_primitive: core.Primitive, f: lu.WrappedFun,
     assert call_primitive.multiple_results
     vals: Sequence[TfVal] = [t.val for t in tracers]
     f = _interpret_subtrace(f, self.main, tuple(t.aval for t in tracers))
-    vals_out: Sequence[Tuple[TfVal, core.AbstractValue]] = f.call_wrapped(*vals)
+    if call_primitive == lax.named_call_p:
+      with tf.name_scope(_sanitize_scope_name(params["name"])):
+        vals_out: Sequence[Tuple[TfVal,
+                                 core.AbstractValue]] = f.call_wrapped(*vals)
+    else:
+      vals_out = f.call_wrapped(*vals)
```
tomhennigan, line 754: Preserve the type info here too?

tomhennigan (changes requested):
Looks good! I think this code would benefit from tests to avoid regressions. I think writing these could be a bit of a pain, so here are two suggestions for how you might approach it:

1. Take advantage of the fact jax2tf is an interpreter and test the current scope name during conversion.

```python
def test_name_scope(self):
  log = []

  @named_call
  def my_test_function(x):
    y = tf.Variable(1., name='foo')
    log.append(y.name)
    return x ** 2

  jax2tf.convert(my_test_function)(2)
  name, = log
  self.assertContains(name, 'my_test_function/foo')
```

2. Stage out the computation as a graph `tf.function(jax2tf.convert(..)).get_concrete_function(..).graph` and then walk the graph asserting at least one node contained an expected name scope.

jax/experimental/jax2tf/jax2tf.py
qiuminxu: Done.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -46,6 +47,17 @@
...
 
+# TF scope validation, refer to tensorflow/python/framework/ops.py
+_VALID_SCOPE_REGEX = re.compile("^[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*$")
+_INVALID_SCOPE_CHAR = re.compile("[^A-Za-z0-9_.\\/>-]")
+
+
+def _sanitize_scope_name(name):
+  scope_name = _INVALID_SCOPE_CHAR.sub("_", name)
+  if not _VALID_SCOPE_REGEX.match(scope_name):
+    scope_name = "F_{}".format(scope_name)
```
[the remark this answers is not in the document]
qiuminxu: Done.

jax/experimental/jax2tf/jax2tf.py
qiuminxu: Replaced with https://github.com/tensorflow/tensorflow/blob/r2.3/tensorflow/core/framework/node_def_util.cc#L731 which should be stable.

I consulted with TF team member who touched the code over a year ago, the name scope code in TF is quite old and it would be hard to introduce new behaviors.  The new hidden behavior might cause a bit of confusion, and the backward compatibility could also be an issue. It doesn't look too promising for TF team to support this.

qiuminxu:
Thank you @tomhennigan for the review and suggestion for the test

> Looks good! I think this code would benefit from tests to avoid regressions. I think writing these could be a bit of a pain, so here are two suggestions for how you might approach it:
> 
> 1. Take advantage of the fact jax2tf is an interpreter and test the current scope name during conversion.
> 
> ```python
> def test_name_scope(self):
>   log = []
> 
>   @named_call
>   def my_test_function(x):
>     y = tf.Variable(1., name='foo')
>     log.append(y.name)
>     return x ** 2
> 
>   jax2tf.convert(my_test_function)(2)
>   name, = log
>   self.assertContains(name, 'my_test_function/foo')
> ```
> 
> 1. Stage out the computation as a graph `tf.function(jax2tf.convert(..)).get_concrete_function(..).graph` and then walk the graph asserting at least one node contained an expected name scope.

The first suggestion works great! Thank you!

qiuminxu:
The jax2tf_test needs to wait for https://github.com/google/jax/pull/4796 to merge to pass.

tomhennigan (approved):

gnecula (approved):
LGTM! Thanks!

Please fix the lint errors and squash as appropriate, and we are ready to merge.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -539,7 +551,12 @@ def process_call(self, call_primitive: core.Primitive, f: lu.WrappedFun,
     assert call_primitive.multiple_results
     vals: Sequence[TfVal] = [t.val for t in tracers]
     f = _interpret_subtrace(f, self.main, tuple(t.aval for t in tracers))
-    vals_out: Sequence[Tuple[TfVal, core.AbstractValue]] = f.call_wrapped(*vals)
+    if call_primitive.name == "named_call":
+      with tf.name_scope(_sanitize_scope_name(params["name"])):
+        vals_out: Sequence[Tuple[TfVal,
```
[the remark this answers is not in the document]
qiuminxu: Yes, I did a bisect and find after https://github.com/google/jax/pull/4470/, the name_scope of forward pass (which was missing) show up in xprof. Before that, the jax2tf process_call function doesn't capture the named_call of the forward pass, after the PR, those start to appear. 

The named_call function does have an argument, I'm not sure why that will be constant folded by jax prior to the conversion. (Maybe because that doesn't have an effect on the output?)

PR #4650, opened 2020-10-20
Ensure that check_jaxpr is done with abstract values

gnecula:
Prior to this it was possible, e.g., for code that contains a Literal,
such as  `broadcast_to_dim[shape=(1000)] 0.0` to result in FLOPS during checking.

The assertion is broken by many tests unless we raise_to_shape for Literals.

I have timed the checks on my laptop and I do not see a reduction in the
total test time.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -1344,7 +1344,7 @@ def _check_jaxpr(jaxpr: Jaxpr, in_avals: Sequence[AbstractValue]):
     try:
       in_avals = map(read, eqn.invars)
       typecheck_assert(all(not isinstance(ina, ConcreteArray) for ina in in_avals),
-                       f"Equation given ConcreteArray type inputs")
+                       "Equation given ConcreteArray type inputs")
       if prim in custom_typechecks:
         custom_typechecks[prim](*in_avals, **eqn.params)
       if prim.call_primitive:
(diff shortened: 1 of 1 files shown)


mattjj (approved):
LGTM!

jax/core.py
```diff
@@ -1343,6 +1343,8 @@ def write(v: Var, a: AbstractValue) -> None:
     prim = eqn.primitive
     try:
       in_avals = map(read, eqn.invars)
+      typecheck_assert(all(not isinstance(ina, ConcreteArray) for ina in in_avals),
+                       f"Primitive {prim} takes ConcreteArray abstract inputs")
```
froystig, line 1347:
```suggestion
                       f"Equation given ConcreteArray type inputs")
```

Minor: the primitive name and equation are already included as context whenever `typecheck_assert` raises in this block (see a few lines down in `except ...`)

froystig (approved):

jax/core.py
```diff
@@ -1343,6 +1343,8 @@ def write(v: Var, a: AbstractValue) -> None:
     prim = eqn.primitive
     try:
       in_avals = map(read, eqn.invars)
+      typecheck_assert(all(not isinstance(ina, ConcreteArray) for ina in in_avals),
+                       f"Equation given ConcreteArray type inputs")
```
froystig, line 1347:
```suggestion
                       "Equation given ConcreteArray type inputs")
```

PR #4658, opened 2020-10-20
[multi-buf] simplify custom object test avals

jakevdp:


mattjj (approved):
LGTM.

This suggestion isn't related to this particular change, and isn't urgent, but I suggest using the term "data type" rather than "object", so e.g. we rename the file to custom_data_type_test.py, and in the comment near the top of the file we say "jaxpr value" or "member of a jaxpr data type" rather than "jaxpr object". The reason is that "object" is a Python concept rather than a jaxpr one.

This file is a great illustration of how to add a new data type and make it work with compilation! Orthogonally, we could make these data types work with autodiff too (by registering them with the ad_util.py functions, and defining `at_least_vspace` on the AbstractValue subclass to associate a tangent data type).

jakevdp:
Thanks! Renaming test files requires some corresponding internal changes, so I'll leave that for now.

lukepfister:
This PR has been incredibly helpful for my own work. Thank you!

I have a question (and I apologize if this isn't a good place to ask).  Is it possible to make your Sparse data type work with some of the `jnp` "ufuncs", eg `jnp.abs`, `jnp.sin`, etc?  Is it just a matter of registering the type with the appropriate handlers?  

Any pointers would be appreciated.

PR #4661, opened 2020-10-20
Emit a warning on non-hashable static arguments in jax.jit.

jblespiau:
The message looks like, e.g.:

Static argument (index 1) of type <class 'numpy.ndarray'> for function f is non-hashable. As this can lead to unexpected cache-misses, it will raise an error in a near future.

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -41,8 +41,9 @@ from . import ad_util
 from . import dtypes
 from .core import eval_jaxpr
 from .api_util import (wraps, flatten_fun, apply_flat_fun, flatten_fun_nokwargs,
-                       flatten_fun_nokwargs2, argnums_partial, flatten_axes,
-                       donation_vector, rebase_donate_argnums)
+                       flatten_fun_nokwargs2, argnums_partial,
+                       argnums_partial_except, flatten_axes, donation_vector,
+                       rebase_donate_argnums)
 from .traceback_util import api_boundary
 from .tree_util import (tree_map, tree_flatten, tree_unflatten, tree_structure,
                         tree_transpose, tree_leaves, tree_multimap,
@@ -130,11 +131,12 @@ def jit(fun: Callable[..., T],
       arguments to treat as static (compile-time constant). Operations that only
       depend on static arguments will be constant-folded in Python (during
       tracing), and so the corresponding argument values can be any Python
-      object. Calling the jitted function with different values for these
-      constants will trigger recompilation. If the jitted function is called
-      with fewer positional arguments than indicated by ``static_argnums`` then
-      an error is raised. Arguments that are not arrays or containers thereof
-      must be marked as static. Defaults to ().
+      object. Static arguments should be hashable, meaning both ``__hash__``
+      and ``__eq__` are implemented. Calling the jitted function with different
+      values for these constants will trigger recompilation. If the jitted
+      function is called with fewer positional arguments than indicated by
+      ``static_argnums`` then an error is raised. Arguments that are not arrays
+      or containers thereof must be marked as static. Defaults to ().
     device: This is an experimental feature and the API is likely to change.
       Optional, the Device the jitted function will run on. (Available devices
       can be retrieved via :py:func:`jax.devices`.) The default is inherited from
@@ -198,8 +200,7 @@ def _python_jit(
       raise ValueError(msg.format(static_argnums, donate_argnums, len(args)))
     f = lu.wrap_init(fun)
     if static_argnums:
-      dyn_argnums = [i for i in range(len(args)) if i not in static_argnums]
-      f, dyn_args = argnums_partial(f, dyn_argnums, args)
+      f, dyn_args = argnums_partial_except(f, static_argnums, args)
     else:
       dyn_args = args
     args_flat, in_tree = tree_flatten((dyn_args, kwargs))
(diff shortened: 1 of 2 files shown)


jblespiau:
Fixed mypy (by ignoring it).

jax/api.py
```diff
@@ -130,11 +131,12 @@ def jit(fun: Callable[..., T],
       arguments to treat as static (compile-time constant). Operations that only
       depend on static arguments will be constant-folded in Python (during
       tracing), and so the corresponding argument values can be any Python
-      object. Calling the jitted function with different values for these
-      constants will trigger recompilation. If the jitted function is called
-      with fewer positional arguments than indicated by ``static_argnums`` then
-      an error is raised. Arguments that are not arrays or containers thereof
-      must be marked as static. Defaults to ().
+      object. Static arguments should be hashable and comparable. Calling the
```
mattjj, line 134: I think ["comparable" entails more than just equality comparison](https://docs.python.org/2.0/ref/comparisons.html). Instead, [just saying hashable entails both `__hash__` and `__eq__`](https://docs.python.org/3.1/glossary.html#term-hashable). So I suggest we either just say "Static arguments should be hashable." or else just inline the standard Python definition for explicitness:

```suggestion
      object. Static arguments should be hashable, meaning both ``__hash__`` and ``__eq__` are implemented. Calling the
```

jax/api_util.py
```diff
@@ -74,16 +75,78 @@ def flatten_fun_nokwargs2(in_tree, *args_flat):
...
-  fixed_args = tuple([unit if i in dyn_argnums else wrap_hashably(arg)
-                      for i, arg in enumerate(args)])
+  fixed_args = tuple(
+      unit if i in dyn_argnums else wrap_hashably(arg)
+      for i, arg in enumerate(args))
   dyn_args = tuple(args[i] for i in dyn_argnums)
   return _argnums_partial(f, dyn_argnums, fixed_args), dyn_args
 
+
+def argnums_partial_except(f,
```
mattjj, line 105:
```suggestion
def argnums_partial_except(f: lu.WrappedFun,
```

jax/api_util.py
```diff
@@ -74,16 +75,78 @@ def flatten_fun_nokwargs2(in_tree, *args_flat):
...
+                           static_argnums: Tuple[int, ...],
+                           args,
```
mattjj, line 107:
```suggestion
                           args: Tuple[Any],
```

jax/api_util.py
```diff
@@ -74,16 +75,78 @@ def flatten_fun_nokwargs2(in_tree, *args_flat):
...
+                           raise_error=False):
```
mattjj, line 108:
```suggestion
                           raise_error: bool = False) -> Tuple[lu.WrappedFun, Tuple[Any]]:
```

jax/api_util.py
```diff
@@ -74,16 +75,78 @@ def flatten_fun_nokwargs2(in_tree, *args_flat):
...
+
+  Args:
+    f: The function being wrapped.
+    static_argnums: The indices of the static arguments.
+    args: The positional arguments.
+    raise_error: If `True`, an error will be raised instead of a warning.
+
+  Returns:
+    A tuple of:
+      The function wrapped into a partial with static-arguments wrapped.
```
mattjj, line 119: It's not accurate to say the function is wrapped into a partial, because what's actually being returned is a lu.WrappedFun instance.

I think this docstring isn't saying much more than the type annotations above. We could either
1. adapt the language so that it's precise (like the type annotations)
2. delete most of the docstring.
I suggest the latter.

How about a one-line docstring that just says

```python
"""Version of ``argnums_partial`` that checks hashability of static_argnums."""
```

jax/api_util.py
```diff
@@ -74,16 +75,78 @@ def flatten_fun_nokwargs2(in_tree, *args_flat):
...
+
+  Args:
+    f: The function being wrapped.
+    dync_arguments: The indices of the dynamic arguments (i.e. non-static
+      positial arguments).
+    args: The positional arguments.
+
+  Returns:
+    A tuple of:
+      The function wrapped into a partial with the non dynamic arguments
```
mattjj, line 90: This docstring isn't quite accurate (first element of the return value isn't a partial, it's an lu.WrappedFun), and moreover I don't think this docstring conveys more than type annotations.

Can you revert the changes to `argnums_partial`, both in the code and in the docstring? IIUC it doesn't need to change in this PR.

jax/api_util.py
mattjj, line 108: Can we remove the `raise_error` argument? I don't think it's used by any callers, and having options on a function is kind of an anti-pattern. Instead, we can just have the version of the function body that prints a warning for now, and in the future edit it to raise an error. (If we want to make it switchable later, we could have a flag rather than an argument to an internal-only function.)

mattjj (changes requested):
Thanks for this! I have some changes to suggest.

jax/api.py
jblespiau: Done (manually, copying the suggested edits

jax/api_util.py
jblespiau: Done.

jax/api_util.py
jblespiau: Done.

jax/api_util.py
jblespiau: Done.

jblespiau: Done.

jax/api_util.py
jblespiau: Done.

jblespiau:
mypy was unhappy with the returned value annotation, so i removed them:

: error: Incompatible return value type (got "Tuple[Any, Tuple[Any, ...]]", expected "Tuple[WrappedFun, Tuple[Any]]")

mattjj (approved):
LGTM!

mattjj:
Thanks for your patience, and your hard work!

tomweingarten:
A number of the examples pass dicts as static_args to create_model. Should the examples be updated so they don't throw these errors?

For example:
https://github.com/google/flax/blob/master/examples/wmt/train.py#L257

jekbradbury:
Yes, they should! Can you open an issue on the Flax repo for that? (Unless there are also instances of this in the JAX repo.)

PR #4671, opened 2020-10-21
Add `lax.conv_general_dilated_local`

romanngg:
Also known as locally connected layer, convolution without weight sharing.


romanngg:
(closed accidentally, reopening)

Sohl-Dickstein:
A ping on this. It would be great to have this functionality in JAX! I have a new project now that requires this functionality, and will have to patch it in since it's not in JAX yet.

mattjj (approved):

mattjj:
Sorry for somehow not noticing this for so long!

jax/_src/lax/other.py
```diff
@@ -114,3 +114,115 @@ def conv_general_dilated_patches(
...
+  Args:
+    lhs: a rank `n+2` dimensional input array.
+    rhs: a rank `n+2` dimensional array of kernel weights. Unlike in regular
+      CNNs, its spatial coordinates (`H`, `W`, ...) correspond to output spatial
+      locations, while input spatial locations are fused with the input channel
+      locations in the single `I` dimension, in the order of
+      `"C" + ''.join(c for c in rhs_spec if c not in 'OI')`, where
+      `rhs_spec = dimension_numbers[1]`. For example, if `rhs_spec == "WHIO",
+      the unfolded kernel shape is
+      `"[output W][output H]{I[receptive window W][receptive window H]}O"`.
```
shoyer, line 148: Why not make the argument `2*n+2` dimensional, e.g., (w, h, x, y, i, o) ?

romanngg: I agree the shape is not intuitive, but the reasoning here was to avoid redundant reshapes. Namely, `conv_general_dilated_patches` returns flattened patches (due to using a conv-based implementation - https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.conv_general_dilated_patches.html), hence I wanted the local convolution to have weights that can be multiplied with patches without reshapes.

If we make `rhs` higher-dimensional, then each forward/backward pass would have a reshape of `rhs` before applying. One alternative is to have `conv_general_dilated_patches` itself return unflattened patches, but then the reshapes will be happening inside `conv_general_dilated_patches`.

I know that some reshapes are easily optimized away, but some aren't and can introduce a performance hit (see e.g. b/168747617), so I thought it would be safer to keep it without reshapes. Lmk if this makes sense, or do you think it won't impact performance here?

shoyer: Personally I think both functions would be clearer if they took/returned higher dimensional arrays, but I see your point.

PR #4336, opened 2020-09-18
Implicitly restarted arnoldi

mganahl:
Hi! 

This is a draft of a JAX implementation of an implicitly restarted arnoldi method for non-symmetric problems. This PR is part of an ongoing effort to upstream some of the sparse linalg methods of `TensorNetwork` to JAX.  I will add tests for the method soon (there are some tests in `TensorNetwork` that I'll reuse).

For the moment I just wanted to ask if the JAX team was  interested in supprting this (and a symmetric version of it). 
If so, I'd be happy to extend this further. It would be great if you could have a quick look at it and let me know how this looks from your point of view (e.g. are there any JAX specific things/gotchas I should be aware of or could improve upon in the code).

thanks!

PR #3114, opened 2020-05-15
Experimental thick-restart Lancozs method for sparse eigenvalue problems

shoyer:
I'm not sure if actually makes sense to merge this code. It probably needs some additional work/testing to make it more robust, but maybe `jax.experimental` would be a suitable incubator? Either way, hopefully it's a useful point of reference.

xref #3112

PR #4675, opened 2020-10-22
implicitly restarted lanczos method

mganahl:
Hi!
This is an implementation of an implicitly restarted Lanczos method for JAX.  The implementation is similar to #4336, but makes explicit use of the symmetry/hermiticity of the linear operator. This is very related to #3114, which uses thick restarts instead of implicit ones.

As one non-trivial test case I added computation of the lowest eigenvalue-eigenvector pairs of a free-fermion lattice Hamiltonian for up to N=18 lattice sites, for which exact results are available.

Comments highly appreciated!


mganahl:
It seems that one of the python 3.6 builds is failing, but I'm having trouble figuring out from the logs what's the exact reason.

mganahl:
commented out the scipy consistency test for now because it breaks one of the builds for unknown reasons,

mganahl:
Hi @shoyer, just kindly asking if you could have a look at this. Thanks!

PR #4681, opened 2020-10-22
Fix a bug in `make_jaxpr`, as static-arguments should be passed as is

jblespiau:


mattjj (approved):
LGTM thanks for catching this.

PR #4682, opened 2020-10-22
[x64] Add initial weak_type logic to abstract eval rules

jakevdp:
This is a small part of the exporation from #3903. The main effect of this change is to make unary lax operations propagate the `weak_type` from input to output. This solves some of the jit invariance issues outlined in #3903, while punting on the more difficult questions of how to handle the full dtype promotion table in the presence of weak types.

Example repro:
```python
import jax.numpy as jnp
from jax import jit

x = jnp.arange(4, dtype='float16')

def add(x, y):
  return x + abs(y)

print(add(x, 1.0).dtype)
print(jit(add)(x, 1.0).dtype)
```
Output before this change:
```
float16
float32
```
Output after this change
```
float16
float16
```

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -402,7 +402,7 @@ def convert_element_type(operand: Array, new_dtype: DType) -> Array:
   if type(operand) in dtypes.python_scalar_dtypes:
     operand = np.asarray(operand, new_dtype)
   old_dtype = dtypes.canonicalize_dtype(_dtype(operand))
-  if old_dtype == new_dtype:
+  if old_dtype == new_dtype and not dtypes.is_weakly_typed(operand):
     return operand
   if (dtypes.issubdtype(old_dtype, np.complexfloating) and
       not dtypes.issubdtype(new_dtype, np.complexfloating)):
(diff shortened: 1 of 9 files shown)


jax/_src/lax/lax.py
```diff
@@ -402,8 +402,6 @@ def convert_element_type(operand: Array, new_dtype: DType) -> Array:
   if type(operand) in dtypes.python_scalar_dtypes:
     operand = np.asarray(operand, new_dtype)
   old_dtype = dtypes.canonicalize_dtype(_dtype(operand))
-  if old_dtype == new_dtype:
-    return operand
```
jakevdp, line 406: Note: I removed this because `convert_element_type` also now removes the weak type from the input, and it is cleaner to let this happen via the common abstract eval than to create a new primitive to strip the weak type from operand.

tests/dtypes_test.py
```diff
@@ -138,6 +143,23 @@ def testPromoteDtypes(self):
...
+    for xtype, ytype in itertools.product(
+      [int, float, jnp.int16, jnp.int32, jnp.float16, jnp.float32], repeat=2)
+    for xfun, yfun in itertools.product(
+      [identity, abs, jnp.array], repeat=2)
+    )
+  def testBinaryPromotionJitInvariance(self, xtype, ytype, xfun, yfun):
+    """Test jit invariance of simple binary promotion rules with and without weak types."""
+    f = lambda x, y: xfun(x) + yfun(y)
+    args_maker = lambda: [xtype(1), ytype(1)]
+    self._CompileAndCheck(f, args_maker, check_dtypes=True)
```
jakevdp, line 161: This test is not strictly necessary for this PR, but it is a good TDD case for the next step in implementing weak types, so I thought it would be useful to include.

jakevdp:
Replaced by #4850

PR #4684, opened 2020-10-22
Cleanup: pass function name rather than function object

jakevdp:
I quickly searched for other instances of this in `_promote_args` and related functions, but this seems to be the only one.


mattjj (approved):
Good catch!

PR #4691, opened 2020-10-23
[jax2tf] Fix implementation and tests of select_and_scatter_add conversion.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1470,18 +1470,20 @@ def _select_and_scatter(
...
-  # TODO(phawkins): handle the select and scatter jaxprs correctly.
-  a = tf.constant(0, operand.dtype)
-  select_fn = _ge_fn.get_concrete_function(a, a)
-  scatter_fn = _add_fn.get_concrete_function(a, a)
-  return tfxla.select_and_scatter(operand, window_dimensions, window_strides,
-                                  padding, source, init_value, select_fn,
-                                  scatter_fn)
-tf_impl[lax.select_and_scatter_add_p] = _select_and_scatter_add
+@functools.partial(bool_to_int8, argnums=(0, 1))
+def _select_and_scatter_add(source, operand, *, select_prim, window_dimensions,
```
gnecula, line 1474: Was this primitive not tested at all? How come the previous binding had so different kwargs?

gnecula (approved):
LGTM

jax/experimental/jax2tf/jax2tf.py
bchetioui: No, it was in the list of primitives that were completely untested. I wouldn't be able to tell regarding the bindings, it seems that the initial import of jax2tf (`8e0a012666932c1929f08af889c8f6aa247d47c4`) had it already. I stumbled upon this by chance when working on benchmarking the Quickdraw model.

gnecula: Good catch!

PR #4693, opened 2020-10-23
Fixes #4692

### LINKED ISSUE #4692: jax.scipy.minimize optional function args not being passed properly
In `jax.scipy.optimize._minimize`,
```python
fun_with_args = partial(fun, *args)
```
incorrectly puts the `args` before the parameter when `args` are provided.
E.g. 

```python
from functools import partial
func = lambda a,b: print(a,b)

args = ('b',)
partial(func, *args)('a')
# b a instead of a b as intended
```

Joshuaalbert:
Just a minor correction.

### DIFF
--- a/jax/_src/scipy/optimize/line_search.py
+++ b/jax/_src/scipy/optimize/line_search.py
@@ -112,7 +112,7 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
 
     # This will cause the line search to stop, and since the Wolfe conditions
     # are not satisfied the minimization should stop too.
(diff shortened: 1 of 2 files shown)


Joshuaalbert:
Fixes #4692

@googlebot I fixed it. (There is nothing to fix that I can see. I'm using the same github account I've always used, and the contact details are correct)

jakevdp:
Regarding the CLA issues: you should check your local git configuration. The commit email is listed as `45720Physics`, which  I suspect does not match the email associated with your github ID. Once you fix that, you can undo the previous commits and re-commit with the correct info.

Issue #4594, opened 2020-10-15
Jax-based BFGS minimize fails to converge, head-to-head with SciPy


gully:
Hello,
Thank you for making Jax!
I recently compared the jax-based BFGS minimize with the scipy-based BFGS minimize on an astronomical scene modeling demo of 3 parameters.  I find that the Jax version failed with error code 3 "zoom failed", while the SciPy version succeeds:

https://gist.github.com/gully/577c6d8eb96d26379f8e6ce89b9d2a2e 

This might be a corner case, since the inverse-Hessian appears to be non-Positive-semidefinite, perhaps pointing towards some other problem.  At first I thought it was a `float32` issue, but I'm pretty sure all my arrays are `float64` (I change the config at the top).  I'm using jax version 2.3, and jaxlib 0.1.55 with CUDA 10.2.  

Any guidance is appreciated!  Thanks again.

Joshuaalbert:
Hi @gully glad to have you trying it out. It looks like the jacobian is very large, `[    8428.3213677 , -2903921.86042314,  6528682.31072694]` in magnitude at the initial point. The output says number of iterations is 1. Meaning that it failed to do a line search on the first iteration because the line search didn't let small enough values be tried. I'll look into adjusting this behaviour to favour very steep scapes better. But for now just normalise your loss function by some suitably large number like `jnp.linalg.norm(jnp.array([    8428.3213677 , -2903921.86042314,  6528682.31072694]))`. This will let smaller step sizes work at the start.

gully:
@Joshuaalbert Thank you so much for a careful response and looking through the notebook.  I was following your PR to get this functionality merged.  I will try as you describe.

Joshuaalbert:
Glad this solved it. To address the underlying issue of line search in regions of large gradients I'll need to adjust the linesearch. In principle once the Hessian inverse starts to approximate the actual Hessian inverse (typically several iterations in) the step size tends to unity. But, at the start the optimal step size can be very large or very small which can be a problem. Maybe simply letting a large range of step sizes during zoom part of line search would solve this.

Joshuaalbert:
Can be closed

PR #4695, opened 2020-10-23
Fixes #4692 and #4594

### LINKED ISSUE #4692: jax.scipy.minimize optional function args not being passed properly

Joshuaalbert:
Minor code changes to fix #4692 and #4594

### DIFF
--- a/tests/scipy_optimize_test.py
+++ b/tests/scipy_optimize_test.py
@@ -86,6 +86,14 @@ class TestBFGS(jtu.JaxTestCase):
     scipy_res = scipy.optimize.minimize(func(np), x0, method='BFGS').x
     self.assertAllClose(scipy_res, jax_res, atol=2e-5, check_dtypes=False)
 
+  def test_fixes4594(self):
+    n = 2
+    A = jnp.eye(n) * 1e4
+    def f(x):
+      return jnp.mean((A @ x) ** 2)
+    results = jax.scipy.optimize.minimize(f, jnp.ones(n), method='BFGS')
+    self.assertAllClose(results.x, jnp.zeros(n), atol=1e-6, rtol=1e-6)
+
 
 if __name__ == "__main__":
   absltest.main()
--- a/jax/_src/scipy/optimize/line_search.py
+++ b/jax/_src/scipy/optimize/line_search.py
@@ -112,7 +112,7 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
 
     # This will cause the line search to stop, and since the Wolfe conditions
     # are not satisfied the minimization should stop too.
-    state = state._replace(failed=state.failed | (dalpha <= 1e-5))
+    state = state._replace(failed=state.failed | (dalpha <= 1e-10))
 
     # Cubmin is sometimes nan, though in this case the bounds check will fail.
     a_j_cubic = _cubicmin(state.a_lo, state.phi_lo, state.dphi_lo, state.a_hi,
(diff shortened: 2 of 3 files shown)


jakevdp:
Hi - thanks for the contribution! I think the fix for #4692 is fine, but I'm not sure about the change related to #4594. How did you come up with `1E-5` here, as opposed to, say, `1E-6` or `1E-4`? Is there a test we could add that would catch regressions along the lines of the issue reported in #4594?

All told, I think it might be beneficial to contribute these changes separately, since they require distinct discussions.

Joshuaalbert:
Hi @jakevdp, when I wrote the initial line search method I used a threshold for zooming such that if too-small step sizes are tried then it stops. This typically means you're at a saddle point. I used the threshold of 1e-5 because that worked well in the tests I used, and indeed all the comparitive tests with scipy in the unit tests pass with 1e-5. However 1e-5 is not small enough when the jacobian large near the initial variable of a minimisation. That was the issue that arose in #4594. I've now chosen a value of 1e-10 to allow less well behaved objectives near the initial point. Note, that after a few iterations of BFGS or any other approximate variable metric method the step size will settle around unity so you don't need to worry about small lengthscales, except when you hit a saddle point. The unit tests are still passing because this change doesn't impact cases with well-behaved jacobians near the initial point. I'm happy to revert 1e-10 to 1e-5.

jakevdp:
Can you add a test case for the issue in #4594; i.e. one that passes with `1E-10` but fails with `1E-5`?

jakevdp:
Looks great, thanks!

One last thing: could you please squash these changes into a single commit? In the meantime, I'll trigger GPU and TPU versions of the new test.

Joshuaalbert:
squashed

jakevdp (approved):

tests/scipy_optimize_test.py
```diff
@@ -86,6 +86,14 @@ def min_op(x0):
     scipy_res = scipy.optimize.minimize(func(np), x0, method='BFGS').x
     self.assertAllClose(scipy_res, jax_res, atol=2e-5, check_dtypes=False)
 
+  def test_fixes4594(self):
+    n = 2
+    A = jnp.eye(n) * 1e4
+    def f(x):
+      return jnp.mean((A @ x) ** 2)
+    results = jax.scipy.optimize.minimize(f, jnp.ones(n), method='BFGS')
+    assert jnp.isclose(results.x, jnp.zeros(n)).all()
```
jakevdp, line 95: This test fails on TPU; we'll need to change the precision

Also, since this is in the test framework, we should be using `self.assertAllClose`. If it had been using that, I'd have been able to tell you what precision we need 😀

Joshuaalbert: Oops, I changed it now to try with 1e-8.

tests/scipy_optimize_test.py
```diff
@@ -86,6 +86,14 @@ def min_op(x0):
     scipy_res = scipy.optimize.minimize(func(np), x0, method='BFGS').x
     self.assertAllClose(scipy_res, jax_res, atol=2e-5, check_dtypes=False)
 
+  def test_fixes4594(self):
+    n = 2
+    A = jnp.eye(n) * 1e4
+    def f(x):
+      return jnp.mean((A @ x) ** 2)
+    results = jax.scipy.optimize.minimize(f, jnp.ones(n), method='BFGS')
+    self.assertAllClose(results.x, jnp.zeros(n), atol=1e-8)
```
jakevdp, line 95: Still fails on TPU. Let's switch this to `atol=1e-6, rtol=1e-6`, then squash into a single commit, and I think we'll be good.

Joshuaalbert: done

jakevdp (approved):

Issue #2952, opened 2020-05-04
FFT precision/performance


zymone:
I find the noticeably difference between outputs of `numpy.fft.fft` and `jax.numpy.fft.fft`.
The difference also changes with different device. For `cpu` device error is bigger than for `gpu` device. On the other hand mean absolute error for `gpu` implementation of `fft` from e.g. PyTorch is around 1e-8 which seems reasonably. I guess that might be some minor bug.

Second issue is the performance of the `jax.numpy.fft.fft` on `cpu` device. I am aware that `jax` is intended for GPU/TPU, but the overhead of jax for `fft` using `cpu` seems weirdly big.

Below is the simple code for reproduction. 

```Python
%env JAX_ENABLE_X64=1
%env JAX_PLATFORM_NAME=cpu

import time
import numpy as np

import jax
from jax import numpy as jnp

np.random.seed(0)

signal = np.random.randn(2**20)
signal_jax = jnp.array(signal)

jfft = jax.jit(jnp.fft.fft)

X_np = np.fft.fft(signal)
X_jax = jfft(signal_jax)

print(np.mean(np.abs(X_np)))
print('max:\t', jnp.max(jnp.abs(X_np - X_jax)))
print('mean:\t', jnp.mean(jnp.abs(X_np - X_jax)))
print('min:\t', jnp.min(jnp.abs(X_np - X_jax)))

### CPU
# 907.3490574884647
# max:	 2.8773885332210747
# mean:	 0.3903197564919141
# min:	 2.4697454729898156e-05

### GPU
# 907.3490574884647
# max:	 0.001166179716824765
# mean:	 0.00020841654559267488
# min:	 2.741492442122853e-07

R = 100
ts = time.time()
for i in range(R):
    _ = np.fft.fft(signal)
print('numpy fft execution time [ms]:\t', (time.time()-ts)/R * 1000)

# Compile
_ = jfft(signal_jax).block_until_ready()

ts = time.time()
for i in range(R):
    _ = jfft(signal_jax).block_until_ready()
print('jax fft execution time [ms]:\t', (time.time()-ts)/R * 1000)

### CPU
# numpy fft execution time [ms]:	 36.75990343093872
# jax fft execution time [ms]:	         219.37960147857666

### GPU
# numpy fft execution time [ms]:	 38.53107929229736
# jax fft execution time [ms]:	         0.38921356201171875
```

jakevdp:
Please see the related discussion in #2874.

mattjj:
[...]

PR #4697, opened 2020-10-23
[JAX] Use PocketFFT for FFTs on CPU instead of Eigen.

hawkinsp:
PocketFFT is the same FFT library used by NumPy (although we are using the C++ variant rather than the C variant.)

For the benchmark in https://github.com/google/jax/issues/2952 on my workstation:

Before:
```
907.3490574884647
max:     4.362646594533903e-08
mean:    6.237288307614869e-09
min:     0.0
numpy fft execution time [ms]:   37.088446617126465
jax fft execution time [ms]:     74.93342399597168
```

After:
```
907.3490574884647
max:     1.9057386696477137e-12
mean:    3.9326737908882566e-13
min:     0.0
numpy fft execution time [ms]:   37.756404876708984
jax fft execution time [ms]:     28.128278255462646
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,6 +50,9 @@ tf_workspace(
 
 tf_bind()
 
+load("//third_party/pocketfft:workspace.bzl", pocketfft = "repo")
+pocketfft()
+
 # Required for TensorFlow dependency on @com_github_grpc_grpc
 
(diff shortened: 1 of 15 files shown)


hawkinsp:
Submitted #4699 which is a clone of this PR (for reasons to do with version control.)

PR #4698, opened 2020-10-23
add jax.custom_gradient wrapper for jax.custom_vjp

mattjj:
There was a deprecated version of this wrapper implemented in terms of jax.custom_transforms (which itself is deprecated, and hopefully soon to be removed), but this commit adds an implementation in terms of jax.custom_vjp. One drawback it has relative to jax.custom_vjp is that it doesn't support Python control flow in the backward-pass function (e.g. for inserting a debugger into the backward pass).

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -65,7 +65,7 @@ from .interpreters import batching
 from .interpreters import masking
 from .interpreters import invertible_ad as iad
 from .interpreters.invertible_ad import custom_ivjp
-from .custom_derivatives import custom_jvp, custom_vjp
+from .custom_derivatives import custom_jvp, custom_vjp, custom_gradient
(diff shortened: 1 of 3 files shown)


froystig (approved):

sharadmv:
Thanks for the addition!

mattjj:
Unfortunately I can't merge this immediately because it breaks some internal tests where folks are relying on (1) disabling omnistaging and (2) using custom_gradient. (Sounds like old code!) So I either need to switch between the new and old versions based on the value of config.omnistaging_enabled, or I need to remove those users of custom_gradient, or I need to have another idea. I think the first approach might work.

Internal test failure link for my future self: test/OCL:338776806:BASE:338776875:1603500800896:426c4a7a

PR #4702, opened 2020-10-24
FIX: example usage of optimizers API in module docstring

fabianp:
The example usage of the optimizers API in this docstring did not correspond with the current API where optimizers return a triplet (opt_init, opt_update, get_params) and not an object with .init, .update and .get_params fields. Using the previous code would result in the error "AttributeError: 'Optimizer' object has no attribute 'init'"

### DIFF
--- a/jax/experimental/optimizers.py
+++ b/jax/experimental/optimizers.py
@@ -69,12 +69,12 @@ Example Usage:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


jakevdp (approved):
Looks good! I think we don't run doctests on experimental code, so that's why this wasn't caught automatically.

fabianp:
thanks for the review @jakevdp !

PR #4707, opened 2020-10-26
register filtered stack trace path exclusions from outside of `traceback_util`

froystig:
Previously, in `traceback_util.py`, we maintained a list of include/exclude paths under which stack frames were filtered. This PR introduces a module-level function for registering a path exclusion, and uses it throughout the codebase.

Hopefully this is easier maintain consistently. Our recent factor/move of several files under `_src` inadvertently affected what frames were filtered. This PR fixes that in particular.

### DIFF
--- a/jax/abstract_arrays.py
+++ b/jax/abstract_arrays.py
@@ -20,6 +20,9 @@ from . import ad_util
 from . import core
 from . import dtypes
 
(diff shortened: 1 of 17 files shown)


mattjj (approved):
LGTM, really nice improvement.

PR #4711, opened 2020-10-26
Add typing annotations and expose py.typed

NeilGirdhar:


jax/_src/lax/control_flow.py
```diff
@@ -1116,7 +1116,16 @@ def cond_bind(*args, branches, linear):
...
 ### scan
 
-def scan(f, init, xs, length=None, reverse=False, unroll=1):
+Carry = TypeVar('Carry')
+X = TypeVar('X')
+Y = TypeVar('Y')
+
+def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
+         init: Carry,
+         xs: Optional[Union[X, Sequence[X]]],
```
hawkinsp, line 1125: Unfortunately this type is too restrictive. `scan` accepts arbitrary pytrees, not just values and sequences.

We also don't know how to describe pytrees to the Python type checkers, so I think the best we can do here is something like `Any` or something that is sugar for `Any`.

NeilGirdhar: Thanks for looking at this PR!  Although, I don't understand how `X` is too restrictive.  `X` is an unbound `TypeVar`, which means it is bound to `Any`.  Perhaps you could illustrate a valid call to scan that doesn't satisfy the annotation?

If anything, I wish `scan` were more restrictive accepting only:
```python
def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
         init: Carry,
         xs: Optional[X],
         length: Optional[int] = None,
         reverse: bool = False,
         unroll: int = 1) -> Tuple[Carry, Y]:
```
And for the cases where `xs` used to be a sequence, users could pass in `tree_multimap(jnp.stack, xs)` in that place.

jax/api.py
```diff
@@ -1743,6 +1744,16 @@ def _vjp_pullback_wrapper(cotangent_dtypes, io_tree, fun, py_args):
   return tree_unflatten(out_tree, ans)
 
 
+@overload
```
NeilGirdhar, line 1747: I could add a TODO here about using PEP 586's `Literal` on `has_aux` to make this more ergonomic when JAX drops Ptyhon 3.7 one day.

NeilGirdhar:
@hawkinsp Any chance you could take a look at this again?  I have to merge this PR every time JAX releases a new version in order to keep the type annotations working.

jax/_src/lax/control_flow.py
hawkinsp: I'm think was confused by the `Sequence`. `X` can be pretty much anything (e.g., a dict of list of arrays is fine). But why isn't the type annotation just:

```
Carry = TypeVar('Carry')
X = TypeVar('X')
Y = TypeVar('Y')

def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
         init: Carry,
         xs: X,
         length: Optional[int] = None,
         reverse: bool = False,
         unroll: int = 1) -> Tuple[Carry, Y]:
```

You don't, as far as I can tell, need to special case either `Optional` or `Sequence`. `xs` can already be `None` or a sequence by virtue of being a free type variable.

jax/_src/nn/functions.py
```diff
@@ -25,10 +26,12 @@
 from jax.scipy.special import logsumexp as _logsumexp
 import jax.numpy as jnp
 
+Array = Any
+
 # activations
 
 @custom_jvp
-def relu(x):
+def relu(x: Array) -> Array:
```
hawkinsp, line 34: No action required, but I suspect that ultimately we will need to accept an `ArrayLike` type (that includes a number of non-arrays, e.g., scalars,) but return an `Array`. Obviously this doesn't matter yet given `Array = Any`.

jax/tree_util.py
```diff
@@ -152,7 +159,7 @@ def tree_unflatten(cls, aux_data, children):
   register_pytree_node(cls, op.methodcaller('tree_flatten'), cls.tree_unflatten)
   return cls
 
-def tree_map(f, tree):
+def tree_map(f: Callable[[Any], Any], tree: T) -> T:
```
hawkinsp, line 162: I don't think it's correct to conclude that the input and output are the same type.
Counterexample:
```
jax.tree_util.tree_map(lambda x: "hi", 3)
```

Same several other places in this file.

hawkinsp (changes requested):

jax/_src/lax/control_flow.py
NeilGirdhar: The reason that you need to special case `Optional` and `Sequence` is because the `TypeVar` `X` ensures that `xs` matches the argument to `f`.   With your definition, you won't be able to accept:
```python
scan(f, some_carry, None, 10)  # error: Cannot infer type argument 1 of "scan"
```
I'm perfectly happy removing the `Sequence` annotation if you prefer that.

jax/_src/nn/functions.py
NeilGirdhar: Yeah, I'm looking to see what the JAX team comes up with.  NumPy 1.20 already has type annotations for `ndarray`, and it looks like they're making it generic on shape and dtype soon.

jax/tree_util.py
NeilGirdhar: Oh, great point!  I'll fix this.

NeilGirdhar: With the new overloads, we get
```python
reveal_type(tree_map(lambda x: "hi", 3))  # Revealed type is 'builtins.int'
reveal_type(tree_map(lambda x: 2 * x, [3.0, 4.0]))  # Revealed type is 'builtins.list[builtins.float]'
reveal_type(tree_multimap(f, [4.5, 3.2], [3.0, 4.0]))  # Revealed type is 'Union[builtins.list[builtins.float], builtins.float*]'
reveal_type(tree_multimap(g, 4.5, 2))  # Revealed type is 'Union[builtins.float, builtins.str*]'
```

jax/_src/lax/control_flow.py
hawkinsp: The sequence example you give first is not type correct, so there's nothing wrong with rejecting it. Arrays and lists are not interchangeable from the perspective of scan (or indeed any pytree-accepting function).

e.g.
```
import jax
from jax import lax

from typing import Tuple

def f(carry: float, x: int) -> Tuple[float, float]:
    print(carry)
    print(x)
    return carry, x + 0.5

lax.scan(f, 0.7, [1, 2, 3, 4])
```

```
AttributeError                            Traceback (most recent call last)
~/p/jax/jax/_src/lax/control_flow.py in scan(f, init, xs, length, reverse, unroll)
   1196   try:
-> 1197     lengths = [x.shape[0] for x in xs_flat]
   1198   except AttributeError as err:

~/p/jax/jax/_src/lax/control_flow.py in <listcomp>(.0)
   1196   try:
-> 1197     lengths = [x.shape[0] for x in xs_flat]
   1198   except AttributeError as err:

AttributeError: 'int' object has no attribute 'shape'

<ipython-input-3-0b5bbb8fa958> in <module>
      9     return carry, x + 0.5
     10
---> 11 lax.scan(f, 0.7, [1, 2, 3, 4])

~/p/jax/jax/_src/lax/control_flow.py in scan(f, init, xs, length, reverse, unroll)
   1200     raise ValueError(
   1201       msg.format(', '.join(str(x) for x in xs_flat
-> 1202                            if not hasattr(x, 'shape')))) from err
   1203
   1204   if length is not None:

ValueError: scan got value with no leading axis to scan over: 1, 2, 3, 4.
```

I'm also quite hesitant about the need to special case `Optional`. If anything I think that says "it might be better to just use `Any` instead of `X` and `Y`, because the type checker has no good way to understand pytrees".

jax/tree_util.py
```diff
@@ -152,7 +160,15 @@ def tree_unflatten(cls, aux_data, children):
   register_pytree_node(cls, op.methodcaller('tree_flatten'), cls.tree_unflatten)
   return cls
 
-def tree_map(f, tree):
+@overload
+def tree_map(f: Callable[[U], U], tree: T) -> T:
+    ...
+
+@overload
+def tree_map(f: Callable[[T], U], tree: T) -> U:
```
hawkinsp, line 168: I still don't think this is a correct annotation. Counterexample:
```
def f(x: int) -> float: return x + 2.5
jax.tree_util.tree_map(f, {"foo": 3})
```
This cannot match the first overload because the input and output of `f` do not have the same type, and it cannot match the second overload because the input and output of `f` do not match the input and output of the `tree_map`.

I'm not convinced there is any meaningful type you can give this function that improves on the `Any` type you have below.

Similarly below.

jax/_src/lax/control_flow.py
NeilGirdhar: Oh, that's good to know.  I assumed that sequences would be handled by the flatten that happens inside `scan`.  I always use dataclasses containing arrays, so I never thought to try it.  Thanks for setting me straight!  I'll remove that now.

Regarding making `xs` optional, it seems logical to me.  Could you give me an example that should be accepted by the type variable approach, but isn't?  I think that would make it clear.

jax/tree_util.py
NeilGirdhar: Yeah, great point.   It would be really nice to be able to keep the overloads since the first one covers the really common case where `f` maps array to array, transforming the components of a pytree.  If I add a third overload using `Any` as you suggest, then we get:

```python
def f(x: int) -> str:
    return "hi"

def g(x: int) -> int:
    return x

reveal_type(tree_map(f, 3))  # Revealed type is 'builtins.str*'
reveal_type(tree_map(f, {'foo': 3}))  # Revealed type is 'Any'
reveal_type(tree_map(g, [3, 3, 3]))  # Revealed type is 'builtins.list[builtins.int]'
```

Does that miss any important cases?

jax/_src/lax/control_flow.py
NeilGirdhar: Okay, you're right!  I misunderstood the way the flattening inside scan works.  Thanks for setting me straight.  I'll fix the annotation now.

jax/tree_util.py
NeilGirdhar: @hawkinsp Does it look okay to you now?

jax/tree_util.py
NeilGirdhar: Just pinging on this since I think this is all that's left to get this change pushed in.

hawkinsp (changes requested):
Would you mind squashing your commits as well?

jax/_src/nn/functions.py
```diff
@@ -212,7 +215,7 @@ def glu(x, axis=-1):
 logsumexp = _logsumexp
 
 
-def log_softmax(x, axis=-1):
+def log_softmax(x: Array, axis: int = -1) -> Array:
```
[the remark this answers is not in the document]
NeilGirdhar: Done.

NeilGirdhar:
Squashed and requested changes applied.

jax/api.py
```diff
@@ -26,10 +26,11 @@
 import functools
 import inspect
 import itertools as it
-import operator
+import sys
 import threading
 import weakref
```
NeilGirdhar, line 31: By the way, this is unused.  Should I remove it?

hawkinsp: I wouldn't do it in this PR.

hawkinsp (approved):

NeilGirdhar:
@hawkinsp Is there anything else I need to do?  I see a merging blocked, and I see something about pytype, but when I run pytype on my machine, I don't get any errors relevant to this change.  Please let me know if you need me to make any more changes.

hawkinsp:
I'm still working on getting this merged. Python type changes are risky because they can cause code that you would not think to be affected to suddenly stop type checking.

NeilGirdhar:
@hawkinsp Oh, no problem!  I just didn't understand if the merging blocked error was something I needed to address.  Thanks again for your fast reviews!

jax/tree_util.py
```diff
@@ -247,7 +274,21 @@ def _replace_nones(sentinel, tree):
       return tree
 
 no_initializer = object()
-def tree_reduce(function, tree, initializer=no_initializer):
+
+@overload
```
hawkinsp, line 278: Oddly enough the annotations on `tree_reduce` are blocking me from submitting this PR. They reveal a type error in some code that uses `tree_reduce`. I'm corresponding with the author if it's a real bug or a type checker problem. I think however I could submit this right now if we temporarily removed the two overloads on `tree_reduce`.

jax/_src/nn/functions.py
```diff
@@ -212,7 +215,7 @@ def glu(x, axis=-1):
 logsumexp = _logsumexp
 
 
-def log_softmax(x, axis=-1):
+def log_softmax(x: Array, axis: Union[int, Tuple[int, ...]] = -1) -> Array:
```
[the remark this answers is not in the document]
NeilGirdhar: Done, and I did the same for [softmax](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.softmax.html), which has a similar interface.

jax/tree_util.py
NeilGirdhar: Okay, as you wish!

jax/_src/nn/functions.py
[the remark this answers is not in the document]
NeilGirdhar: I also broadened `normalize` since it relies only on `mean` and mean accepts `Optional[Union[int, Tuple[int, ...]]]`.

jax/tree_util.py
```diff
@@ -247,7 +274,23 @@ def _replace_nones(sentinel, tree):
...
+#                 tree: Any) -> T:
+#     ...
+#
+# @overload
+# def tree_reduce(function: Callable[[T, Any], T],
+#                 tree: Any,
+#                 initializer: T) -> T:
+#     ...
+
+def tree_reduce(function: Callable[[T, Any], T],
```
hawkinsp, line 291: You'll actually have to remove the type annotation completely for me to submit this. (It's the output type causing the problem, so you could also just change the output type to `Any` for now.)

That said, we think this annotation caught a real bug in a RL agent, so it's also fine to reinstate the overloads if you don't mind waiting another day to get this in... Up to you.

NeilGirdhar: Oh, great, happy to hear the type annotations caught something! :smile: 

Let's wait then, so that we can do a more perfect job.

hawkinsp: I think I might be able to get this in now. Please push a version with the overloads added back!

NeilGirdhar: Done!

jax/_src/nn/functions.py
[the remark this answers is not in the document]
NeilGirdhar: (Wrapped some lines for the 80 character limit.)

PR #4713, opened 2020-10-26
simplify select jvp

mattjj:
also remove some coverage of broadcast_p, which jax never generates now


mattjj:
@jekbradbury I had to revert the bit that removed broadcasts because it caused some annoying test failures.

jekbradbury (approved):

PR #4117, opened 2020-08-21
Support overriding implementations of JAX functions within a scope

shoyer:
The idea here is to support overriding the implementation of higher order JAX
functions within a limited scope, for libraries such as Haiku and Flax that
implement their own versions of this functions that support mutation.

Usage example:

    with jax.override_context({'lax.scan': my_scan}):
      # all calls to lax.scan() are replaced by my_scan()
      ...

Ultimately, it would be great to replace this with a unified interface for
mutable state in JAX, but this could be a convenient temporary measure.

PR #4076, opened 2020-08-15
Add experimental __array_module__ method

shoyer:
xref https://github.com/google/jax/issues/1565

`__array_module__` (see [NEP 37](https://numpy.org/neps/nep-0037-array-module.html)) is an experimental alternative to `__array_function__` and `__array_ufunc__` for "duck array" compatibility with NumPy that promises to be much less invasive.

Example usage, for writing a generic version of `np.stack()` on top of `concatenate`, `asarray`, `shape` attributes and array indexing:

```python
import numpy_dispatch

def duckarray_stack(arrays):
    """This "stack" function should work with any array library, including JAX."""
    npx = numpy_dispatch.get_array_module(*arrays)  # returns jax.numpy for JAX arrays
    arrays = [npx.asarray(arr) for arr in arrays]
    shapes = {arr.shape for arr in arrays}
    if len(shapes) != 1:
        raise ValueError('all input arrays must have the same shape')
    expanded_arrays = [arr[npx.newaxis, ...] for arr in arrays]
    return npx.concatenate(expanded_arrays, axis=0)
```

For now, you need to use `numpy_dispatch.get_array_module()` to try this feature (with https://github.com/seberg/numpy-dispatch) but in the long term the hope is to merge this into NumPy proper as `np.get_array_module()`.

My reasoning for merging it into JAX (on an experimental basis with no guarantees, of course) is that:

1. It's not invasive -- the implementation is small and self-contained.
2. No backwards compatibility issues. Unlike `__array_function__` and `__array_ufunc__`, `__array_module__` will always require an explicit opt-in by libraries that use it by calling `get_array_module()`.
2. Other NumPy developers [want evidence](https://github.com/numpy/numpy/pull/16935#issuecomment-673951287) that this is actually feasible.
3. Scikit-Learn developers like @thomasjpfan are interested in exploring supporting scikit-learn on top of NumPy-like libraries like JAX, and experimental support for this protocol will make that easier.

Note: this PR does add `numpy-dispatch` as a optional testing requirement in order to verify that this works. If desired, we could remove this from CI, but installing numpy-dispatch appears to only add about 7 extra seconds of build time.

PR #4725, opened 2020-10-28
an experiment in handling instances with __jax_array__

mattjj:
Before raising an error on an unrecognized type, first check if the object defines a `__jax_array__` method. If it does, call it!

This provides a way for custom types to be auto-converted to JAX-compatible types.

We didn't add a check for `__array__` because that might entail a significant change in behavior. For example, we'd be auto-converting `tf.Tensor` values. Maybe it's better to remain loud in those cases.

Implementing this method is *not* sufficient for a type to be duck-typed enough for use with `jax.numpy`. But it may be necessary. That is, someone trying to add a duck-typed array to be used with JAX identified a need for `__jax_array__` or similar.

This feature is experimental, so it may disappear, change arbitrarily, or never be documented.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -300,9 +300,11 @@ def _result_dtype(op, *args):
   return _dtype(op(*args))
 
 
-def _arraylike(x): return isinstance(x, ndarray) or isscalar(x)
+def _arraylike(x):
+  return isinstance(x, ndarray) or isscalar(x) or hasattr(x, '__jax_array__')
+
 def _check_arraylike(fun_name, *args):
-  """Check if all args fit JAX's definition of arraylike (ndarray or scalar)."""
+  """Check if all args fit JAX's definition of arraylike."""
   assert isinstance(fun_name, str), f"fun_name must be a string. Got {fun_name}"
   if _any(not _arraylike(arg) for arg in args):
     pos, arg = next((i, arg) for i, arg in enumerate(args)
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -122,10 +122,13 @@ xla_result_handlers: Dict[Type[core.AbstractValue], Callable[..., Callable]] = {
 }
 
 def device_put(x, device: Optional[Device] = None) -> Tuple[Any]:
-  x = canonicalize_dtype(x)
-  try:
-    return device_put_handlers[type(x)](x, device)
-  except KeyError as err:
+  handler = device_put_handlers.get(type(x))
+  if handler:
+    x = canonicalize_dtype(x)
+    return handler(x, device)
+  elif hasattr(x, '__jax_array__'):
+    return device_put(x.__jax_array__(), device)
+  else:
     raise TypeError(f"No device_put handler for type: {type(x)}") from err
 
 def _device_put_array(x, device: Optional[Device]):
@@ -151,6 +154,8 @@ def canonicalize_dtype(x):
   for typ in typ.mro():
     handler = canonicalize_dtype_handlers.get(typ)
     if handler: return handler(x)
+  if hasattr(x, '__jax_array__'):
+    return canonicalize_dtype(x.__jax_array__())
   raise TypeError(f"No canonicalize_dtype handler for type: {type(x)}")
 
 def _canonicalize_ndarray_dtype(x):
@@ -173,6 +178,8 @@ def abstractify(x) -> core.AbstractValue:
   for typ in typ.mro():
     aval_fn = pytype_aval_mappings.get(typ)
     if aval_fn: return aval_fn(x)
+  if hasattr(x, '__jax_array__'):
+    return abstractify(x.__jax_array__())
   raise TypeError(f"Argument '{x}' of type '{type(x)}' is not a valid JAX type")
 
 def _make_abstract_python_scalar(typ, _):
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -129,7 +129,7 @@ def device_put(x, device: Optional[Device] = None) -> Tuple[Any]:
   elif hasattr(x, '__jax_array__'):
     return device_put(x.__jax_array__(), device)
   else:
-    raise TypeError(f"No device_put handler for type: {type(x)}") from err
+    raise TypeError(f"No device_put handler for type: {type(x)}")
 
 def _device_put_array(x, device: Optional[Device]):
   backend = xb.get_device_backend(device)
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2068,6 +2068,27 @@ class APITest(jtu.JaxTestCase):
     with self.assertRaisesRegex(ValueError, "tangent values inconsistent"):
       f_jvp(np.ones(2, np.int32))
 
+  def test_dunder_jax_array(self):
+    # https://github.com/google/jax/pull/4725
+
+    class AlexArray:
+      def __init__(self, jax_val):
+        self.jax_val = jax_val
+      def __jax_array__(self):
+        return self.jax_val
+      dtype = property(lambda self: self.jax_val.dtype)
+      shape = property(lambda self: self.jax_val.shape)
+
+    x = AlexArray(jnp.array([1., 2., 3.]))
+    y = jnp.sin(x)
+    self.assertAllClose(y, jnp.sin(jnp.array([1., 2., 3.])))
+    y = api.grad(api.jit(lambda x: jnp.sin(x).sum()))(x)
+    self.assertAllClose(y, jnp.cos(jnp.array([1., 2., 3.])))
+
+    x = AlexArray(jnp.array([[1., 2., 3.]]))
+    y = api.pmap(jnp.sin)(x)
+    self.assertAllClose(y, jnp.sin(jnp.array([[1., 2., 3.]])))
+
 
 class RematTest(jtu.JaxTestCase):
 
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2089,6 +2089,11 @@ class APITest(jtu.JaxTestCase):
     y = api.pmap(jnp.sin)(x)
     self.assertAllClose(y, jnp.sin(jnp.array([[1., 2., 3.]])))
 
+    x = jnp.array(1)
+    a = AlexArray(x)
+    for f in [jnp.isscalar, jnp.size, jnp.shape, jnp.dtype]:
+      self.assertEqual(f(x), f(a))
+
 
 class RematTest(jtu.JaxTestCase):
 
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -437,6 +437,9 @@ def convert_element_type(operand: Array, new_dtype: DType = None,
     msg = "Casting complex values to real discards the imaginary part"
     warnings.warn(msg, np.ComplexWarning, stacklevel=2)
 
+  if hasattr(operand, '__jax_array__'):
+    operand = operand.__jax_array__()
+
   if not isinstance(operand, (core.Tracer, xla.DeviceArray)):
     return _device_put_raw(np.asarray(operand, dtype=new_dtype),
                            weak_type=new_weak_type)
--- a/jax/core.py
+++ b/jax/core.py
@@ -859,6 +859,8 @@ def concrete_aval(x):
   for typ in type(x).mro():
     handler = pytype_aval_mappings.get(typ)
     if handler: return handler(x)
+  if hasattr(x, '__jax_array__'):
+    return concrete_aval(x.__jax_array__())
   raise TypeError(f"{type(x)} is not a valid JAX type")
 
 
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -280,7 +280,10 @@ def is_python_scalar(x):
   try:
     return x.aval.weak_type and np.ndim(x) == 0
   except AttributeError:
-    return type(x) in python_scalar_dtypes
+    if hasattr(x, '__jax_array__'):
+      return is_python_scalar(x.__jax_array__())
+    else:
+      return type(x) in python_scalar_dtypes
 
 def dtype(x):
   if type(x) in python_scalar_dtypes:


lukepfister:
This would be very helpful for some work I'm doing.

However, I'd expect `jnp.sin(a)` to also return an `AlexArray`.  Thoughts?

mattjj:
@lukepfister hmm, I think that would be both very hard to implement and also not consistent with NumPy's handling of `__array__`.

lukepfister:
@mattj Ah, fair enough.  I was thinking of the [__array_ufunc__](https://numpy.org/doc/stable/user/basics.subclassing.html#array-ufunc-for-ufuncs) and `__array_wrap__` functionality.   Then the numpy equivalent would indeed return another `AlexArray`.

Would love to know if something similar is possible with jax.

mattjj:
@shoyer can probably help educate me about those angles! (He's probably explained them to me already and I forgot...)

The aim in this particular PR is just specifying the one-way translation from some instance of a class defining `__array__` to a JAX value, though it may be a first step in thinking along the lines of more general overloading.

shoyer:
> @MattJ Ah, fair enough. I was thinking of the [**array_ufunc**](https://numpy.org/doc/stable/user/basics.subclassing.html#array-ufunc-for-ufuncs) and `__array_wrap__` functionality. Then the numpy equivalent would indeed return another `AlexArray`.
> 
> Would love to know if something similar is possible with jax.

We certainly could implement something like this. Both PyTorch and TensorFlow has simplified versions of NumPy's protocols for overriding functions.

That said, there are costs to flexibility:

- checking for protocols like this adds a (small) amount of overhead to every function call (this mattered for NumPy, but the incremental cost is probably negligible compared to JAX's existing overhead from `jit`)
- it adds a layer of indirection into the JAX source code
- it adds a layer of indirection into user code, which now may or may not be using JAX's original implementations

On the whole, I think it might be easier to try using something like NumPy's `__array_module__` protocol for this use-case, which is already implemented on JAX arrays: https://github.com/google/jax/pull/4076. `AlexArray` could implement _NumPy's_ protocol to define a new implementation of the `numpy` namespace on `AlexArray` objects.

In theory, we could do make something like `__jax_module__` in the same style as `__array_module__` to enable generic versions of non-NumPy JAX functions, but I'm not sure it's really worth the trouble (see also https://github.com/google/jax/pull/4117).

jax/_src/numpy/lax_numpy.py
```diff
@@ -300,9 +300,11 @@ def _result_dtype(op, *args):
   return _dtype(op(*args))
 
 
-def _arraylike(x): return isinstance(x, ndarray) or isscalar(x)
+def _arraylike(x):
+  return isinstance(x, ndarray) or isscalar(x) or hasattr(x, '__jax_array__')
```
jakevdp, line 304: Would it be worth checking that `__jax_array__` is callable?

jax/interpreters/xla.py
```diff
@@ -122,11 +122,14 @@ def array_result_handler(device: Optional[Device], aval: core.ShapedArray):
...
-  try:
-    return device_put_handlers[type(x)](x, device)
-  except KeyError as err:
-    raise TypeError(f"No device_put handler for type: {type(x)}") from err
+  handler = device_put_handlers.get(type(x))
+  if handler:
+    x = canonicalize_dtype(x)
+    return handler(x, device)
+  elif hasattr(x, '__jax_array__'):
+    return device_put(x.__jax_array__(), device)
```
jakevdp, line 130: Maybe overkill, but we might add a guardrail here against infinite recursion here and below.

tests/api_test.py
```diff
@@ -2068,6 +2068,27 @@ def test_linearize_aval_error(self):
...
+    class AlexArray:
+      def __init__(self, jax_val):
+        self.jax_val = jax_val
+      def __jax_array__(self):
+        return self.jax_val
+      dtype = property(lambda self: self.jax_val.dtype)
+      shape = property(lambda self: self.jax_val.shape)
+
+    x = AlexArray(jnp.array([1., 2., 3.]))
+    y = jnp.sin(x)
```
jakevdp, line 2083: It's a bit strange that this returns a jax array rather than wrapping the output in an AlexArray. I wonder if it wouldn't be better to provide something analogous to [NEP 18](https://numpy.org/neps/nep-0018-array-function-protocol.html) or some of the followups to that?

jakevdp (commented):
This looks pretty good, but it's not clear to me whether this is the right approach in the long run. Should we instead follow numpy and dispatch these kinds of things via `__array_function__`?

mattjj:
I'm not familiar with `__array_function__`, but after reading [NEP 18](https://numpy.org/neps/nep-0018-array-function-protocol.html) briefly my thinking is that this change is about something pretty different: this `__jax_array__` change is about allowing user data types to be compatible with jax.numpy and jax.lax functions, whereas NEP 18 is about making `numpy` functions overloadable. Am I misunderstanding?

jax/interpreters/xla.py
mattjj: Hm, what kind of guard do you have in mind?

That sounds like it'd just be a bug in the `__jax_array__` implementation, and trying to guard against a potential infinite recursion just to raise a different error doesn't seem clearly better than just letting people get the infinite recursion error.

shoyer:
I think `__array_function__`/`__array_module__` are for a complementary but distinct purpose: making it possible to overload the `numpy` (or perhaps `jax.numpy`) namespaces.

Potentially conflicting behavior would arise only if a custom array object adds `__jax_array__` now and later adds `__jax_array_function__`, which would then change the behavior of JAX functions on that object from coercing to switching to a separate implementation.

jakevdp:
Ah, you're right - I guess I was actually thinking of the `__array_ufunc__` mechanism. The way the test case is structured made me think that this would be the more natural approach, though it would require a much deeper change to JAX.

If this is primarily about making it possible for `jnp.array(x)` to work for user-defined `x`, then this looks reasonable.

If this is primarily about making it so e.g. `jnp.sin(x)` works for user-defined `x`, then I think piggy-backing on the `__array_ufunc__` mechanism might make more sense.

jax/_src/numpy/lax_numpy.py
mattjj: Hmm, interesting, but that seems a bit over-defensive; if it's not callable, then we'll almost immediately after this get a callability error from trying to call it.

AlexeyKurakin:
@jakevdp 
I was original requestor of this feature and initial motivation was to allow `jnp.sin(x)` to allow with `x` which is instance of user-defined class. In other words, to allow user-defined classes to be "automatically casted" to JAX types.

mattjj:
> If this is primarily about making it so e.g. jnp.sin(x) works for user-defined x, then I think piggy-backing on the __array_ufunc__ mechanism might make more sense.

I think `__array_ufunc__` might also be for something slightly different: it'd let the user data type decide how to dispatch a `numpy` (not `jax.numpy`) ufunc. But we're not trying to add that capability; we just want user data types to be able to present that they know how to turn themselves into jax arrays.

AlexeyKurakin:
I'm also not familiar with `__array_ufunc__`, but I think `__jax_array__` is a reasonable way to achieve what I described.

mattjj:
It's really helpful to check through all this, since these subtle differences are pretty confusing! I think I puzzled through this 1.5 months ago when we first started this PR, but didn't write it down for posterity. So it's good to double-check!

shoyer:
Conceptually, I would put `__array_ufunc__` in the same category as `__array_function__` for overriding the behavior of functions in a namespace.

I'm not sure we want that behavior, but certainly making it possible to explicit _convert_ arbitrary objects into JAX arrays seems useful.

jakevdp:
Sounds good, I guess I was mainly just confused by the test case calling jnp.sin rather than jnp.array - it made me think the motivation for this change was different than what it is.

mattjj:
I wrote the test that way because we don't want to require people to call `jnp.array` to do the conversion. If we required that, then folks might as well just call their own conversion function. This is just a convenience layer so that e.g. Objax user code can avoid explicit conversion to jax arrays.

jakevdp:
Got it – in that case, we'll have to do a pass on `lax_numpy.py` to make certain we're calling `array` or `asarray` on relevant inputs: I don't think we currently do that in all cases.

mattjj:
No, actually this PR covers what we need. We don't need to call `array` or `asarray` because we check for the `__jax_array__` method at a lower level,  namely on our dispatch path (that's the change in xla.py) and in `core.get_aval` (that's the core.py change). That's how `jnp.sin` works now!

jakevdp:
Agreed for anything that calls directly into `lax`. But there are routines in `jax.numpy` that don't. For example:
```python

**Comment by mattjj** [comment]
You're right that there might still be some unexpected gaps, though I expect we can just wait for @AlexeyKurakin to tell us about any.

For these particular examples, it's not completely clear to me that we want `jnp.isscalar` (or `jnp.dtype`, `jnp.shape`, etc) to change their behavior based on the existence of `__jax_array__`. Maybe those just show that the author of `AlexArray` must do more work to sufficiently duck-type JAX arrays, i.e. more than just implementing `__jax_array__`? Perhaps some of those start working once there's a `shape` attribute available?

In other words, it does seem like a nice invariant if `jnp.shape(x) == jnp.shape(jnp.array(x))`, but I wonder if that's up to the person doing the duck-typing rather than up to JAX.

(By the way, this is tangential, but I found it surprising based on the function name alone that `dtypes.is_python_scalar(jnp.array(1))` is True! I think I get the reasoning, but it is surprising. From #4737 AIUI.)

**Comment by jakevdp** [comment]
Agreed that those functions are a bit of a gray area in themselves, but throughout lax_nunpy.py we trigger different code-paths based on their outputs, so I anticipate anyone using this feature for anything beyond basic array conversion will be in for some surprises unless we put a concerted effort into handling and texting this case throughout.

**Comment by mattjj** [comment]
Well, again, it's not clear to me whether we need to do anything there, or if that's more up to the user writing the custom data type to do better duck typing.

It seems that `jnp.isscalar` is the odd duck:

```python
import jax.numpy as jnp
from jax import jit, grad

class AlexArray:
  def __init__(self, jax_val):
    self.jax_val = jax_val
  def __jax_array__(self):
    return self.jax_val
  dtype = property(lambda self: self.jax_val.dtype)
  shape = property(lambda self: self.jax_val.shape)

x = jnp.array(1)
a = AlexArray(x)

for f in [jnp.dtype, jnp.shape, jnp.size]:
  print(f(x))
  print(f(a))
  print()
```

```
int32
int32

()
()

1
1
```

I wonder if the implementation of `dtypes.is_python_scalar` should just be tweaked. I'll add a change to this PR.

In any case, let's leave the gap analysis to @AlexeyKurakin to share with us as he experiments with this feature, through issues and PRs. This experimental feature is not meant to entail sweeping changes up-front; if it did, we'd just drop it entirely! My guess is Alex would rather we land and experiment.

mattjj:
Pushed 5eb3685 to change `dtypes.is_python_scalar` and to test the specific issues @jakevdp brought up so far. WDYT about merging?

jakevdp:
One note: `size()` above only works because the underlying array is size 1. If it were anything else, the size would not match and a fair number of `jax.numpy` routines would work incorrectly (anything relying on `ravel()`, for instance, which depends on calling `size()`)

I'm fine merging this for now, because I don't think we'll have many users depending on it. But if we ever do, I think it will require a fair bit more effort to make it usable for anything beyond simple array conversions.

mattjj:
Ah, but `size` can be handled just by exposing a `size` attribute, as in

```python
class AlexArray:
  def __init__(self, jax_val):
    self.jax_val = jax_val
  def __jax_array__(self):
    return self.jax_val
  dtype = property(lambda self: self.jax_val.dtype)
  shape = property(lambda self: self.jax_val.shape)
  size = property(lambda self: self.jax_val.size)
```

So that's up to the user duck-typing, not up to JAX, IMO. We get to make up the rules of our own duck-typing game, after all :)

> But if we ever do, I think it will require a fair bit more effort to make it something we can recommend for anything beyond simple array conversions.

This might be true, but it might not be; it's not clear to me yet, and the examples so far have all been pretty easy.

jakevdp:
Sure, but what else needs to be handled by the user in order to make JAX work properly with objects of this type? I can't answer that right now.

jakevdp:
As a user, I would find it surprising that I need to implement `shape`, `dtype`, `size`, ??, ??, and ?? along with `__jax_array__` in order to make things work correctly. Why not just `__jax_array__` alone?

mattjj:
Well, since `jnp.shape`, `jnp.size`, and `jnp.dtype` are just aliases to `np.shape`, `np.size`, and `np.dtype`, to make the `jnp` versions work one has to do exactly what is needed to make the `np` versions work. We're not changing anything here.

The new method `__jax_array__` is for only the narrow purpose of converting to a jax array when a jax array value is needed, and nothing more. It has no implications for how `jnp.shape`, `jnp.size`, `jnp.dtype` etc work.

I wonder if perhaps you're reasoning by an analogy between how the NumPy `__array__` method works and how `__jax_array__` _could_ work. But that analogy doesn't apply; we're choosing a narrower definition of what `__jax_array__` does. In other words, when you ask "Why not just __jax_array__ alone?", that's just not the spec that we're going for here. And the reasons you bring up are indeed reasons why we _shouldn't_ pursue that spec, namely that it'd be a lot more work.

jakevdp:
The other reason I'm hesitant here is because, on balance, the added capability does not seem commensurate with the support burden. A user could just as easily create a `to_jax()` method and call it when they want to use a jax method. But instead of that we're creating a new jax entrypoint with an intended use that currently has no test coverage, and no estimate of the work it will take to support it properly.

On the balance, ISTM it would be better to recommend users call `to_jax()`. Alternatively, if we want to support `__jax_array__()`, we should do a good accounting of what it would actually take to support it fully before introducing an incomplete implementation.

mattjj:
> the added capability does not seem commensurate with the support burden

> no estimate of the work it will take to support it properly

What's the support burden?

To be clear, my estimate is that this won't take any additional work to support properly, as this PR is by definition all the proper support needed (modulo bugs). That is, there are no additional features in scope, i.e. no changes to how `jnp.shape`, `jnp.size`, `jnp.dtype` etc work.

> A user could just as easily create a to_jax() method and call it when they want to use a jax function.

I believe this was discussed in some detail in the original chat thread, with me as the original proponent! But I came to believe I was wrong, and the goal now is to figure out whether we can easily _avoid_ requiring all Objax user code have this kind of explicit conversion.

If this change proves to have a support burden, we'll remove it. The whole point here is to experiment, as per the PR title. That's also why there's no documentation. If you have examples of a support burden, please provide them, but otherwise it's just speculation, and speculation on which we disagree.

jakevdp:
OK, sounds good. My understanding had been that the goal here was that a custom object could be used in-place of a JAX array in all `jnp` and `lax` functions, and the results would be the same as if you used the underlying array.

It sounds like you're saying that's not the goal or the promise of this feature, in which case this is fine.

jakevdp:
I wanted to get a more quantitative idea of the aspects of this that I was concerned about, so I replaced the `_CheckAgainstNumpy` test utility with this:
```python
  def _CheckAgainstNumpy(self, numpy_reference_op, lax_op, args_maker,
                         check_dtypes=True, tol=None,
                         canonicalize_dtypes=True):
    args = args_maker()
    
    import jax.numpy as jnp
    class AlexArray:
      def __init__(self, jax_val):
        self.jax_val = jax_val
      def __jax_array__(self):
        return self.jax_val
      dtype = property(lambda self: self.jax_val.dtype)
      shape = property(lambda self: self.jax_val.shape)
      size = property(lambda self: self.jax_val.size)
      ndim = property(lambda self: self.jax_val.ndim)

    lax_args = [AlexArray(jnp.array(arg)) if isinstance(arg, jnp.ndarray) else arg for arg in args]
    lax_ans = lax_op(*lax_args)
    numpy_ans = numpy_reference_op(*args)
    self.assertAllClose(numpy_ans, lax_ans, check_dtypes=check_dtypes,
                        atol=tol, rtol=tol,
```
When I ran the test suite with
```
$ pytest -n auto test/lax_numpy_test.py
```
The result is:
```
801 failed, 2797 passed, 419 skipped
```
From a spot-check of the eight hundred test failures, these appear to be real failures: i.e. places where `__jax_array__` is not properly handled by the `jax.numpy` API.

mattjj:
Wow, thanks for doing that! That could be very helpful as a test pattern for someone wanting to duck-type ndarrays for use with `jax.numpy`.

But is it just testing the extent to which that the `AlexArray` implementation is incomplete? That is, those test failures might not have anything to do with `__jax_array__`, and instead just show one needs to do more work to duck-type arrays. Indeed, we'd expect it to need things like `__add__`!

The goal here is not to implement another duck-typed ndarray, or claim that just implementing `__jax_array__` is in any way sufficient for duck typing. As you've shown, it certainly is not!

> My understanding had been that the goal here was that a custom object could be used in-place of a JAX array in all jnp and lax functions

Ah, perhaps this is the crux of the issue! Actually, that's not the goal with this PR. That's Alex's ultimate goal, but in this PR we're just trying to add one missing piece needed so that he can finish his own duck type class. There's still plenty more work needed to implement a complete duck, but that's on his end.

> We should just make sure to note that clearly when documenting it.

Well, I have no plans to document this. Right now it's just an experiment to see if it helps Objax. Think of it like something brand new in `jax.experimental`, except it can't literally live in `jax.experimental`.

I'll revise the OP to clarify that this is an experiment.

(It's also already been rolled back due to test failures. I've got to look into those...)

AlexeyKurakin:
I opened an issue to keep track of this change: https://github.com/google/jax/issues/5356

jax/interpreters/xla.py
```diff
@@ -122,11 +122,14 @@ def array_result_handler(device: Optional[Device], aval: core.ShapedArray):
...
 
 def device_put(x, device: Optional[Device] = None) -> Tuple[Any]:
-  x = canonicalize_dtype(x)
-  try:
-    return device_put_handlers[type(x)](x, device)
-  except KeyError as err:
-    raise TypeError(f"No device_put handler for type: {type(x)}") from err
+  handler = device_put_handlers.get(type(x))
+  if handler:
+    x = canonicalize_dtype(x)
```
shoyer, line 127: It looks like this PR broke internal tests due to this line. In particular, some type like `IntEnum` no longer are accepted by `device_put`.

mattjj: Oh man, I think you nailed it! Is it just because I dropped the `x = canonicalize_dtype(x)` beforehand?

shoyer: That was the only thing I saw that could have broken the existing code path.

PR #4731, opened 2020-10-28
Add explicit tests of weak/strong promotion semantics

jakevdp:
This will help us see more explicitly via code diffs what has changed as we evolve weak type semantics.


froystig (approved):
Nice!

PR #4733, opened 2020-10-29
Upstream implementation of named_call_p primitive to JAX

qiuminxu:
Upstream the implementation of named_call to JAX (there are equivalent implementations in Haiku and Flax).

https://github.com/google/jax/issues/4558

Reference:
Flax implementation:
https://github.com/google/flax/blob/master/flax/core/named_call.py

Haiku implementation:
https://github.com/deepmind/dm-haiku/blob/master/haiku/_src/named_call.py

### DIFF
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2246,7 +2246,7 @@ class LaxTest(jtu.JaxTestCase):
     self.assertLen(jaxpr.jaxpr.eqns, 2)
 
   def test_named_call(self):
-
+    # TODO(qiuminxu) Make named_call a public function in lax. 
     def named_call(f, name):
       def named_f(*args):
         f_ = jax.linear_util.wrap_init(lambda: (f(*args),))
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2245,6 +2245,25 @@ class LaxTest(jtu.JaxTestCase):
                                              (x,), (1.,)))(1.)
     self.assertLen(jaxpr.jaxpr.eqns, 2)
 
+  def test_named_call(self):
+
+    def named_call(f, name):
+      def named_f(*args):
+        f_ = jax.linear_util.wrap_init(lambda: (f(*args),))
+        out, = lax.named_call_p.bind(f_, name=name)
+        return out
+      return named_f
+
+    def square(x):
+      return x ** 2
+
+    @jax.jit
+    def f(x):
+      return named_call(square, 'name_for_testing')(x)
+
+    c = jax.xla_computation(f)(2)
+    self.assertIn('name_for_testing', c.as_hlo_text())
+
 
 class LazyConstantTest(jtu.JaxTestCase):
   def _Check(self, make_const, expected):
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2246,7 +2246,7 @@ class LaxTest(jtu.JaxTestCase):
     self.assertLen(jaxpr.jaxpr.eqns, 2)
 
   def test_named_call(self):
-    # TODO(qiuminxu) Make named_call a public function in lax. 
+    # TODO(qiuminxu) Make named_call a public function in lax
     def named_call(f, name):
       def named_f(*args):
         f_ = jax.linear_util.wrap_init(lambda: (f(*args),))
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2449,6 +2449,24 @@ ad.defjvp2(min_p,
            lambda g, ans, x, y: mul(_brcast(g, y), _balanced_eq(x, ans, y)),
            lambda g, ans, x, y: mul(_brcast(g, x), _balanced_eq(y, ans, x)))
 
+named_call_p = core.CallPrimitive('named_call')
+named_call_p.def_impl(core.call_impl)
+ad.primitive_transposes[named_call_p] = functools.partial(ad.call_transpose,
+                                                         named_call_p)
+
+
+def _named_call_translation_rule(c, axis_env, in_nodes, name_stack,
+                                 *, name='core_call', backend, call_jaxpr):
+  subc = xla.xb.make_computation_builder(name)
+  args = [xla.xb.parameter(subc, i, c.GetShape(n))
+          for i, n in enumerate(in_nodes)]
+  out_nodes = xla.jaxpr_subcomp(subc, call_jaxpr, backend, axis_env, (),
+                                jax.util.extend_name_stack(name_stack, name),
+                                *args)
+  subc = subc.Build(xla.xops.Tuple(subc, out_nodes))
+  return xla.xops.Call(c, subc, list(in_nodes))
+
+xla.call_translations[named_call_p] = _named_call_translation_rule
 
 shift_left_p = standard_naryop([_int, _int], 'shift_left')
 ad.defjvp_zero(shift_left_p)
--- a/jax/lax/__init__.py
+++ b/jax/lax/__init__.py
@@ -168,6 +168,7 @@ from jax._src.lax.lax import (
   min_p,
   mul,
   mul_p,
+  named_call_p,
   naryop,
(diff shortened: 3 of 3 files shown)


qiuminxu:
cc @froystig, @LenaMartens @tomhennigan @trevorcai @levskaya @jheek @avital

levskaya:
@qiuminxu - this looks fine to me on the flax side, thank you!

tests/lax_test.py
```diff
@@ -2245,6 +2245,25 @@ def test_select_jvp_complexity(self):
                                              (x,), (1.,)))(1.)
     self.assertLen(jaxpr.jaxpr.eqns, 2)
 
+  def test_named_call(self):
+
+    def named_call(f, name):
```
tomhennigan, line 2250: I think this `named_call` should be a public function in `lax`. We do not want users having to bind the primitive themselves. I'm happy for that to be done in a follow up PR though since having a single shared primitive is already a strict improvement.

LenaMartens:
Thanks for starting this! 

The named_call_p looks good to me, but +1 on having a public function which wraps the the call_primitive. I think ideally we would want to replace [named_fun](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call.py#L109) in Haiku with an upstreamed version because it still depends on internal-ish APIs. 

One interesting edge case we found for this `named_fun`, is that using the named_call primitive raises the abstraction level of the arguments (test for that edge case [here](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call_test.py#L51)). This caused issues when applying the call_primitive to Haiku modules that use non JaxType arguments(eg. a string). These would be correctly static_argnumed by the user, but the call_primitive would abstract them anyway. We fixed (?) this by stashing away the arguments we don't want abstracted. Not sure if this should be included in an upstreamed version.

jekbradbury:
I think it's possible to avoid raising the abstraction level of the arguments, but I'm not sure. Is that specific to the Haiku `stateful_named_call`, or is it caused by the handling of closed-over `*args` when tracing `lambda: f(*args)`?

jekbradbury (approved):

tests/lax_test.py
qiuminxu: Thanks for the review! A named_call public function is great. Let me think about how to implement it since it is where the stateful_named_call in haiku and named_call in flax diverges. Let me start a draft for feedback first.

qiuminxu:
Thanks @LenaMartens for pointing out the interesting point on raising the abstraction level of the arguments.

Flax also implements a higher level implementation of [named_call](https://github.com/google/flax/blob/84674691f91c210dcdbe0c144f7b37c3800fd22a/flax/core/lift.py#L658).

I wonder how to merge the two implementation. Let's find out the best way to do that.

Would it make sense to have the primitive in first and start iterations on shared named_call on a separate PR?

qiuminxu:
> Thanks for starting this!
> 
> The named_call_p looks good to me, but +1 on having a public function which wraps the the call_primitive. I think ideally we would want to replace [named_fun](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call.py#L109) in Haiku with an upstreamed version because it still depends on internal-ish APIs.
> 
> One interesting edge case we found for this `named_fun`, is that using the named_call primitive raises the abstraction level of the arguments (test for that edge case [here](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call_test.py#L51)). This caused issues when applying the call_primitive to Haiku modules that use non JaxType arguments(eg. a string). These would be correctly static_argnumed by the user, but the call_primitive would abstract them anyway. We fixed (?) this by stashing away the arguments we don't want abstracted. Not sure if this should be included in an upstreamed version.

Thanks, is there a test to repro this issue?

qiuminxu:
Thanks everyone for the insights and review!

LenaMartens:
To clarify: there's two utility functions in Haiku, one wrapping the named_call primitive (`_named_call`) and one handling Haiku state (`stateful_named_call`). We wouldn't need to upstream `stateful_named_call` or the Haiku state handling, but it would be nice to be able to replace `_named_call`.

I think the Haiku implementation of `_named_call` is very close to Flax's implementation of `_named_call`. Both functions do a linear_util.wrap_init and flattening/unflattening of the arguments, and in the case of Haiku we stash away the arguments to prevent raising the abstraction level:
```python
# Avoid abstracting inputs by calling as a thunk
f_thunk = lu.wrap_init(lambda: flat_f.call_wrapped(*flat_args),)
out_flat = named_call_p.bind(f_thunk, name=name) 
```

> Thanks, is there a test to repro this issue?

I think actually after omnistaging the edge case has become larger, and all three of `test_static_argnums`, `test_partial_eval` and `test_non_jaxtype` test the abstraction raising [here](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call_test.py#L37).

qiuminxu:
Thanks @LenaMartens. I don't fully understand how the abstracting inputs works, can you explain more?
When I replace the Haiku [named_fun](https://github.com/deepmind/dm-haiku/blob/24d17d498029bc388dda629aa6c46c79c8f82da6/haiku/_src/named_call.py#L109) with the following flax implementation, the named_call test still passes. Are the two implementation equivalent? Does jax.api_util.flatten_fun_nokwargs solves the abstraction raising in another way? @tomhennigan @levskaya @jekbradbury 
 
```
  def named_call(f, name):
    _, in_tree = jax.tree_flatten(())
    def named_f(*args, **kwargs):
      lu_f = jax.linear_util.wrap_init(lambda: f(*args, **kwargs))
      flat_f, out_tree = jax.api_util.flatten_fun_nokwargs(lu_f, in_tree)
      out_flat = lax.named_call_p.bind(flat_f, name=name)
      return jax.tree_unflatten(out_tree(), out_flat)
    return named_f
```

tomhennigan:
Hi @qiuminxu, we do have tests covering this (e.g. `NamedCallTest.test_non_jaxtype_arg`) so I its a good sign that the Flax implementation passes.

Lets not block this PR on the wrapper function, I'd suggest sequencing this in a few changes across JAX/Haiku/Flax as follows:

1. Submit this PR with the primitive.
2. Update Haiku and Flax to use the JAX primitive instead of their own.
3. Align Haiku and Flax implementations of (non stateful) wrapper function (copy from one codebase to the other).
4. Wait for a day or so incase there needs to be a rollback
5. Upstream that unified implementation to JAX.
6. Drop wrapper implementation in Haiku + Flax.

tests/lax_test.py
tomhennigan: Please feel free to just make this a TODO for now.

tomhennigan (approved):

tests/lax_test.py
qiuminxu: Done.

qiuminxu:
> Hi @qiuminxu, we do have tests covering this (e.g. `NamedCallTest.test_non_jaxtype_arg`) so I its a good sign that the Flax implementation passes.
> 
> Lets not block this PR on the wrapper function, I'd suggest sequencing this in a few changes across JAX/Haiku/Flax as follows:
> 
> 1. Submit this PR with the primitive.
> 2. Update Haiku and Flax to use the JAX primitive instead of their own.
> 3. Align Haiku and Flax implementations of (non stateful) wrapper function (copy from one codebase to the other).
> 4. Wait for a day or so incase there needs to be a rollback
> 5. Upstream that unified implementation to JAX.
> 6. Drop wrapper implementation in Haiku + Flax.

Like this plan. Let me follow this and submit the changes in sequential.

froystig (approved):

PR #4737, opened 2020-10-29
Switch to declarative weak dtype promotion rules.

jakevdp:
This changes our handling of weak types in type promotion from a procedural approach to a declarative approach.

I constructed the type promotion table so as to have no change to the existing promotion behavior, as confirmed by the test added in #4731

We plan to make some changes to weak dtype promotion behavior going forward, and this will make it easier to specify the precise behavior we want JAX to have.

### DIFF
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -166,6 +166,19 @@ can_cast = np.can_cast
 issubsctype = np.issubsctype
 
 
+class _WeakType:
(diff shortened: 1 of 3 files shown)


jakevdp:
I'm going to wait on #4738 so we have more explicit test coverage for this in x32 mode.

hawkinsp (approved):

PR #4740, opened 2020-10-30
[jax2tf] Reduce number of tests for select_and_gather_add.

bchetioui:
Reduces the number of tests for the conversion of `select_and_gather_add_p` from 216 to 11.


gnecula (approved):
Nice savings!

PR #4744, opened 2020-10-30
Update weak dtype promotion rules

jakevdp:
This PR updates JAX's type promotion table in preparation for a more systematic handling of weak dtypes. The updated promotion table is generated via least upper bounds of the following type promotion lattice, where `*` indicates weakly-typed values:
lattice

Introducing this lattice leads to two types of user-visible changes: first is the promotion of strongly-typed floats with weakly-typed complex values in x64 mode:

*Before:*
```python
>>> jnp.float32(1) + 1j
DeviceArray(1.+1.j, dtype=complex128)
```
*After:*
```python
>>> jnp.float32(1) + 1j
DeviceArray(1.+1.j, dtype=complex64)
```
The precision of the strongly-typed float now dictates the precision of the output.

Secondly is the promotion of `uint64` with signed integers, which currently results in a strongly-typed float64, will now result in a weakly-typed float. Because the weak float is backed by a float64, this doesn't actually change the resulting promotion behavior for two values (i.e. `result_type(uint64, int16)` is still `float64` as before), but it does ensure associativity of promotion in the presence of more than two values; for example:

*Before:*
```python
>>> a, b, c = jnp.uint64(1), jnp.int16(1), jnp.float16(1)
>>> jnp.result_type(a, b, c)
dtype('float64')
>>> jnp.result_type(c, a, b)
dtype('float16')
```

*After:*
```python
>>> a, b, c = jnp.uint64(1), jnp.int16(1), jnp.float16(1)
>>> jnp.result_type(a, b, c)
dtype('float16')
>>> jnp.result_type(c, a, b)
dtype('float16')
```

Note that weak types still do not propagate through lax operations (there is an outstanding TODO for that, and it will take care to implement well), so the promotion test tables still show strongly-typed equivalents for all expected outputs.

### DIFF
--- a/tests/dtypes_test.py
+++ b/tests/dtypes_test.py
@@ -182,11 +182,11 @@ class DtypesTest(jtu.JaxTestCase):
 
 class TestPromotionTables(jtu.JaxTestCase):
 
-  def testWeakDtypesPromotionTable(self):
+  def testObservedPromotionTable(self):
     """Test that the weak & strong dtype promotion table does not change over time."""
     # Note: * here refers to weakly-typed values
     typecodes = \
-      ['b1','u1','u2','u4','u8','i1','i2','i4','i8','bf','f2','f4','f8','c4','c8','i*','f*','c*']
+        ['b1','u1','u2','u4','u8','i1','i2','i4','i8','bf','f2','f4','f8','c4','c8','i*','f*','c*']
     if FLAGS.jax_enable_x64:
       expected = [
         ['b1','u1','u2','u4','u8','i1','i2','i4','i8','bf','f2','f4','f8','c4','c8','i8','f8','c8'],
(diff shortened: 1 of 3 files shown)


jakevdp:
I'm going to pull-in and run internal tests before review - I suspect the uint + python int promotion rule changes might cause a few ripples in dependent code.

jakevdp:
Internal tests are passing for the updated PR.

jakevdp:
This includes some changes that I spun out to be reviewed separately in #4748. I'll rebase on that once it's merged.

jakevdp:
Rebased after merging #4748; this is ready for review!

gnecula:
I like this very much. I wonder if it is possible to put the lattice in the documentation, in lieu of a table. The table is easy to use if you want to answer the type promotion for one pair of types, the lattice is better if you want to understand the "rule behind" the table.

gnecula:
Please also update the CHANGELOG

hawkinsp:
I'm still quite uncomfortable with the u8/s8 -> f* promotion.

jakevdp:
If we're going to make `u8/i8` promotion an error, I'd prefer to do that separately from this change since it's entirely orthogonal.

hawkinsp:
> If we're going to make `u8/i8` promotion an error, I'd prefer to do that separately from this change since it's mostly orthogonal and likely to be a breaking change for some users.

I don't see it as orthogonal: the lattice shown above isn't correct. It should not have any strong -> weak edges. But I'm happy if we make that change first.

tests/dtypes_test.py
```diff
@@ -182,11 +182,11 @@ def testScalarCastInsideJitWorks(self):
 
 class TestPromotionTables(jtu.JaxTestCase):
 
-  def testWeakDtypesPromotionTable(self):
+  def testObservedPromotionTable(self):
```
hawkinsp, line 185: Delete this test? It's now redundant with the lattice test.

(Alternatively, change the code in dtypes.py to generate its table with the lattice, but don't repeat the table twice if it's a generated artifact.)

hawkinsp (approved):

tests/dtypes_test.py
jakevdp: This is a bit different: it's an integration test to confirm that the actual observed behavior of type promotion implemented in binary operators (typified by `jnp.add`) does not change.

The lattice test is an internal unit test that confirms the internal promotion table can be generated from a lattice.

For performance reasons, I suspect it's not a good idea to have `dtypes.py` generate the promotion table from the lattice, as that generation would happen at import time (although I've not benchmarked it).

hawkinsp: If the goal is to verify `jnp.add` has the behavior in the table, I suggest importing the type table from dtypes.py.

I'm mostly unhappy about the redundancy.

jakevdp: JAX is already a pretty costly import all told, I'd rather avoid making it worse:
```python
%time import jax
# CPU times: user 395 ms, sys: 41.8 ms, total: 437 ms
# Wall time: 439 ms

%time import numpy
# CPU times: user 21 µs, sys: 0 ns, total: 21 µs
# Wall time: 24.6 µs
```

jakevdp: The lattice->promotion table code currently takes ~1.5ms... so it's not much compared to JAX's full startup cost, but does contribute.

What do you think?

hawkinsp: We don't need a second "does not change" test, and for an end-to-end test, we don't need to repeat the table because we are testing the external behavior matches the table.

jakevdp: The other option would be to use the lattice directly within `promote_dtypes`, possibly along with `lru_cache` to compute promotion rules on demand.

jakevdp: > We don't need a second "does not change" test,

I'm not sure what the second test is that you're referring to.

> for an end-to-end test, we don't need to repeat the table because we are testing the external behavior matches the table.

What external behavior are you referring to?

jakevdp:
I just pushed an update that adds a changelog describing the two user-visible changes caused by this PR.

tests/dtypes_test.py
hawkinsp: The lattice test also has the effect of a "does not change" test. You do not need two such tests.

For "external behavior" I mean what you refer to as "it's an integration test to confirm that the actual observed behavior of type promotion implemented in binary operators". You do not need to repeat the type promotion table to write such a test. There should only be one copy of the promotion table anywhere, and I'm happy for it to be the one in dtypes.py and for the test to merely check the actual behavior (as produced by operators) matches the expected behavior (as described by the table).

jakevdp: > The lattice test also has the effect of a "does not change" test. You do not need two such tests.

I don't think that's quite true: the lattice test generates the internal table. The outputs of a binary operation are the internal table as interpreted by `result_type`, which had nontrivial logic prior to this PR. A big reason I created that "does not change" test was to ensure that the change of logic in `result_type` within this PR would lead to consistent results. For that reason, I think it's better not to remove it in the same commit that it's designed to validate; would you be open to me removing it in a followup?

> For "external behavior" I mean...

Ah, got it. There would need to be some non-trivial transformation logic, though, becuase the internal table includes weak types and the results of binary operations currently do not return weak types; also the behavior outside x64 mode differs. 

Because of the first point, I really think it would be better to keep this test for now, even if it's removed in a followup PR. What do you think?

hawkinsp (approved):

PR #4746, opened 2020-10-31
Add multi-host support to sharded_jit.

skye:
This is done similarly to how multi-host pmap works: for each input, each host contributes one "local" piece of a conceptual "global" array, which is further sharded across that host's local devices. Each host similarly gets a local piece of each global output array.

To support this, three new arguments are added to sharded_jit: local_in_parts, local_out_parts, and local_num_partitions. These are similar to the existing in_parts, out_parts, and num_partitions arguments (which are now implicitly assumed to be global), except they specify the partitioning local to each host. These new arguments are only needed when the local and global partitionings are different, i.e. when partitioning across multiple hosts.

Note this doesn't support multi-host partitioning when using pmap(sharded_jit). I'll add that in a subsequent change.

### DIFF
--- a/jax/interpreters/sharded_jit.py
+++ b/jax/interpreters/sharded_jit.py
@@ -15,6 +15,7 @@
 from functools import partial
 from typing import Callable, Optional, Tuple
 
+from absl import logging
 import numpy as np
 
 from .. import core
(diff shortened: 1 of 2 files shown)


skye:
FYI @apaszke in case you wanna take a look

jax/interpreters/sharded_jit.py
```diff
@@ -73,13 +74,28 @@ def _sharded_callable(
     fun: lu.WrappedFun, num_partitions: Optional[int],
     in_parts: Tuple[pxla.PartitionsOrReplicated, ...],
     out_parts_thunk: Callable[[], Tuple[pxla.PartitionsOrReplicated, ...]],
-    name: str, *abstract_args):
+    local_in_parts: Optional[Tuple[pxla.PartitionsOrReplicated, ...]],
+    local_out_parts_thunk: Callable[[], Optional[Tuple[pxla.PartitionsOrReplicated, ...]]],
+    local_num_parts: Optional[int], name: str, *abstract_args):
```
jekbradbury, line 79: It doesn't really matter, but I was tripped up once or twice by the fact that these parameters are called "num_parts" in some places and "num_partitions" in others.

jax/interpreters/sharded_jit.py
```diff
@@ -96,24 +112,59 @@ def _sharded_callable(
...
   assert num_partitions is not None
-  if num_partitions > xb.local_device_count():
+  if num_partitions > xb.device_count():
     raise ValueError(
         f"sharded_jit computation requires {num_partitions} devices, "
-        f"but only {xb.local_device_count()} devices are available.")
+        f"but only {xb.device_count()} devices are available.")
+  if xb.local_device_count() < num_partitions < xb.device_count():
+    raise NotImplementedError(
+        f"sharded_jit across multiple hosts must use all available devices. "
```
jekbradbury, line 121: The one place in P5X where sharded_jit not inside a pmap might be useful on its own (the init function) would require allowing pod sharded_jit on a subset of global devices (since we don't have partial replication support). But in practice it's even better to just use pmap(sharded_jit) for that too, so that we don't need a separate step that copies the single initialized replica (or single initialized replica per replica set) onto the devices in the other replicas. So I think this restriction is fine.

jax/interpreters/sharded_jit.py
```diff
@@ -258,15 +311,30 @@ def sharded_jit(fun: Callable, in_parts, out_parts, num_partitions: int = None):
...
+    local_in_parts: Optional. This should be set when partitioning across
+      multiple processes, and says how each process's worth of data should be
+      partitioned (vs. in_parts which is the "global" partitioning across all
+      processes). This API is likely to change in the future.
+    local_out_parts: Optional. This should be set when partitioning across
+      multiple processes, and says how each process's worth of data should be
+      partitioned (vs. out_parts which is the "global" partitioning across all
+      processes). This API is likely to change in the future.
+    local_num_partitions: Optional. Explicitly specifies the numbers of local
+      devices to partitions across in a multi-process setting.
```
jekbradbury, line 323:
```suggestion
      devices to partition across in a multi-process setting. This API is likely
      to change in the future.
```

jekbradbury (approved):
Looks good! I especially appreciate the `get_local_aval`/`get_global_aval` factoring.

jax/interpreters/sharded_jit.py
skye: We can also lift this without too much trouble if/when we need it, we just don't have something analogous to `devices` for sharded_jit yet.

jax/interpreters/sharded_jit.py
skye: Done.

jax/interpreters/sharded_jit.py
skye: You're right, this is super inconsistent. I adopted the pxla convention and changed everything to `nparts` and `local_nparts` (for brevity!), except the argument names to `sharded_jit`, which I kept as [local_]num_partitions.

PR #4748, opened 2020-10-31
Simplify weak types in promotion table

jakevdp:
I realized that the custom class is not necessary.

A side-effect of this change is that `jnp.promote_types` will now treat `int`, `float`, and `complex` as weakly-typed, which is consistent with how they are handled internally; e.g.
```python
>>> jnp.promote_types(jnp.float16, float)                                                                                                                                
dtype('float16')
```
This is a breaking change, in that this previously returned `dtype('float64')`, but I think it's the right API to have here.

### DIFF
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -166,21 +166,10 @@ can_cast = np.can_cast
 issubsctype = np.issubsctype
 
 
-class _WeakType:
-  def __init__(self, type):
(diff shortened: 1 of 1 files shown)


mattjj (approved):
LGTM, if this is a breaking change, should we do a global check for internal failures before merging?

jakevdp:
I made a minor tweak that causes `promote_types` to be backward compatible; internal global check passes.

PR #4755, opened 2020-11-02
Add a prototype of the new xmap API

apaszke:
At the moment all resources are emulated through vectorization, but it
should be good enough to let us play with some toy examples.

The current implementation is very much a prototype and might end up
being very user unfriendly, because it never checks its assumptions.
Violating them will likely result in a pretty weird error.


apaszke:
Oh and I guess another important thing to note is that collectives are not working just yet.

PR #4759, opened 2020-11-02
add weak_type framework to abstract eval

jakevdp:
This has no user-visible effect as written, but puts in place the framework whereby standard abstract evaluation rules can specify how weak types propagate


jakevdp:
Replaced by #4850

PR #4760, opened 2020-11-03
[jax2tf] Expanded the SavedModel library to allow the compilation of …

gnecula:
…models.

This is important for use with TensorFlow serving.
Also removed the servo_main.py (only applies to OSS TF Serving, which
does not yet support XLA)


jax/experimental/jax2tf/examples/README.md
```diff
@@ -36,19 +41,24 @@ into functions that behave as if they had been written with TensorFlow.
 Therefore, if you are familiar with how to generate SavedModel, you can most
 likely just use your own code for this.
 
-The file `saved_model_main.py` is an executable that shows who to perform the following
-sequence of steps:
+The file `saved_model_main.py` is an executable that shows who to perform the
+following sequence of steps:
```
bchetioui, line 45:
```suggestion
The file `saved_model_main.py` is an executable that shows how to perform the
following sequence of steps:
```

jax/experimental/jax2tf/examples/saved_model_lib.py
```diff
@@ -77,49 +100,46 @@ def save_model(jax_fn: Callable,
       with_gradient=with_gradient,
       in_shapes=[None, shape_polymorphic_input_spec])
 
-  wrapper = _ExportWrapper(tf_fn, params, params_trainable=with_gradient)
+  # Create tf.Variables for the parameters.
+  param_vars = tf.nest.map_structure(
+      # If with_gradient=False, we mark the variables behind as non-trainable,
+      # to ensure that users of the SavedModel will not try to fine tune them.
+      lambda param: tf.Variable(param, trainable=with_gradient),
+      params)
```
bchetioui, line 108: nit: the docstring specifies that this should be some kind of `tuple/list/dictionary`, but then do we need to use `tf.nest.map_structure`? Or maybe we want to abstract in the docstring to say that the params and inputs must be some kind of traversable?

bchetioui (approved):
Small change, otherwise LGTM!

jax/experimental/jax2tf/examples/saved_model_lib.py
gnecula: I do not think that I understand the comment. I think that the docstring is correct, and we do need to use `tf.nest.map_structure` to produce the `param_vars` with the same structure.

Issue #4734, opened 2020-10-29
SIGFPE + RuntimeError caused by missing shape rule for FFT


bchetioui:
The following code

```python
from jax import lax 
from jaxlib import xla_client
import numpy as np

x = np.ones((14, 15))
kwargs = { "fft_type": xla_client.FftType.FFT
         , "fft_lengths": [0]
         }
         
lax.fft(x, **kwargs)
```

triggers a floating point exception and thus crashes.

bchetioui:
It seems that in general, JAX is missing a shape rule for FFT, leading to `RuntimeError`s for other examples such as

```python
from jax import lax 
from jaxlib import xla_client
import numpy as np

x = np.ones((14, 15))
kwargs = { "fft_type": xla_client.FftType.IRFFT
         , "fft_lengths": [1]
         }
         
lax.fft(x, **kwargs)
```
that results in a

```
RuntimeError: Invalid argument: IRFFT requires innermost dimension matches fft_length/2+1, but dimension 1 is 15 and should be 1.: 
This is a bug in JAX's shape-checking rules; please report it!
https://github.com/google/jax/issues
```

PR #4765, opened 2020-11-03
[jax2tf] Fix and trim FFT conversion test.

bchetioui:
It turns out that most of the tests in the `"dtypes"` harnesses were useless, as [lax.fft promotes the dtypes as needed to pass them to lax.fft_p](https://github.com/google/jax/blob/f58f1ee4563852de2853115d885c2eec7a69a77b/jax/_src/lax/fft.py#L55:L66). Passing arrays with other dtypes to `lax.fft_p` results in `RuntimeError`s, similarly to the cases outlined in #4734.

### DIFF
--- a/jax/experimental/jax2tf/tests/primitives_test.py
+++ b/jax/experimental/jax2tf/tests/primitives_test.py
@@ -170,8 +170,10 @@ class JaxPrimitiveTest(tf_test_util.JaxToTfTestCase):
                                   "only 1D FFT is currently supported."):
         harness.dyn_fun(*harness.dyn_args_maker(self.rng()))
     else:
+      tol = None if jtu.device_under_test() == "tpu" else 1e-3
       self.ConvertAndCompare(harness.dyn_fun,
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -170,15 +170,8 @@ def test_fft(self, harness: primitive_harness.Harness):
                                   "only 1D FFT is currently supported."):
         harness.dyn_fun(*harness.dyn_args_maker(self.rng()))
     else:
-      tol = None
-      if jtu.device_under_test() in ("cpu", "gpu"):
```
gnecula, line 174: It is suspicious that these tolerances are not necessary anymore. Were they only needed for some of the dtypes that are now disabled?

bchetioui: No, they were generally needed. I didn't encounter any problematic instance anymore, but I guess the tests might indeed become flaky if we remove it. Adding it back.

PR #4766, opened 2020-11-03
Cleanup outdated jaxlib TODOs in jax/_src/lax/fft.py

bchetioui:
I believe we can remove this since `_minimum_jaxlib_version = (0, 1, 55)`!


hawkinsp (approved):

hawkinsp:
Thanks! I've been meaning to grep through and remove a few of these...

PR #4767, opened 2020-11-03
[jax2tf] Add enable_xla parameter to convert

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -107,9 +107,20 @@ def _safe_convert_to_tensor(val, dtype=None) -> TfVal:
 tf_impl_with_avals: Dict[core.Primitive,
                          Callable[..., Any]] = {}
 
+# XLA is not linked in all environments; when converting a primitive, if the
+# conversion path involves a TFXLA operation and this variable is disabled, an
+# exception is thrown instead.
+ENABLE_XLA = True
```
gnecula, line 113: I would suggest to name this "_enable_xla", with "_" to mark it private, and with lower case because it is a variable.

gnecula (commented):
I would also add at least a test, to check the mechanism. Perhaps you can use this to add a test for convolution, which has both XLA and non-XLA paths?

gnecula:
I think that this deserves some changes to the README.md. Or, perhaps we can wait until we see if this flag is useful at all (it would not be useful if most models fail to convert)

bchetioui:
I am partial to checking which models we can already convert without XLA/which ones we can manage by adding TF conversion paths first, but I don't feel that strongly about it!

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1106,7 +1107,11 @@ def _make_conv_harness(name, *, lhs_shape=(2, 3, 9, 10), rhs_shape=(3, 3, 4, 5),
                        padding=((0, 0), (0, 0)), lhs_dilation=(1, 1),
                        rhs_dilation=(1, 1), feature_group_count=1,
                        dimension_numbers=("NCHW", "OIHW", "NCHW"),
-                       batch_group_count=1):
+                       batch_group_count=1, enable_xla=True,
+                       expect_xla_disabled_exception=False):
+  expect_xla_disabled_exception = expect_xla_disabled_exception or (
+      not (enable_xla or jax2tf.jax2tf.ENABLE_TF_CONVOLUTION))
```
gnecula, line 1113: I was assuming that if we disable XLA we should enable `jax2tf.jax2tf.ENABLE_TF_CONVOLUTION`. I think that there were some cases when it was failing though

bchetioui: Yes, we should, but if you remember there was some weird transposition happening that we unsuccessfully tried to debug for a while. I can enable it and you can test internally!

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -107,9 +107,20 @@ def _safe_convert_to_tensor(val, dtype=None) -> TfVal:
 tf_impl_with_avals: Dict[core.Primitive,
                          Callable[..., Any]] = {}
 
+# XLA is not linked in all environments; when converting a primitive, if the
```
gnecula, line 110: I would extend this comment to say that if XLA is disabled then we try harder to use only standard TF ops, if they are applicable to the concrete use case.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -158,10 +169,15 @@ def convert(fun: Callable, *,
       an error will be raised if a gradient computation is attempted.
       This is due to a current bug in TensorFlow.
 
+    enable_xla: if unset, the converter will raise an error if the function
```
gnecula, line 172: Please expand this comment as above (we try harder to use TF)

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1229,15 +1237,17 @@ def _make_conv_harness(name, *, lhs_shape=(2, 3, 9, 10), rhs_shape=(3, 3, 4, 5),
...
     # tf.nn.convolution only supports a subset of the possible dtypes for JAX
     # convolutions (float16, float32, float64). With the below tests, we ensure
     # that we avoid this branch in cases when it would not succeed.
     _make_conv_harness("tf_conversion_path_dtype", lhs_shape=lhs_shape,
                      padding='VALID', rhs_shape=rhs_shape, dtype=dtype,
                      dimension_numbers=dimension_numbers, lhs_dilation=(1, 1),
-                     rhs_dilation=(1, 1))
-  for dtype in jtu.dtypes.all_inexact
+                     rhs_dilation=(1, 1), enable_xla=False,
+                     expect_xla_disabled_exception=True)
```
gnecula, line 1249: Why is there an exception here?

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1248,5 +1258,15 @@ def _make_conv_harness(name, *, lhs_shape=(2, 3, 9, 10), rhs_shape=(3, 3, 4, 5),
                       lhs_shape=(1, 28, 28, 32), rhs_shape=(3, 3, 16, 8),
                       padding='VALID', batch_group_count=1, lhs_dilation=(1, 1),
                       rhs_dilation=(1, 1), feature_group_count=2,
-                      dimension_numbers=('NHWC', 'HWIO', 'NHWC'))]
+                      dimension_numbers=('NHWC', 'HWIO', 'NHWC'),
+                      enable_xla=False, expect_xla_disabled_exception=True)]
```
gnecula, line 1262: Why is there an exception here?

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -742,8 +742,40 @@ def test_conv_general_dilated(self, harness: primitive_harness.Harness):
...
+    if harness.name == "_tf_conversion_path_3d_lhs=float32[1,4,28,28,1]_rhs=float32[2,3,3,1,16]_windowstrides=(1,1,1)_padding=VALID_lhsdilation=(1,1,1)_rhsdilation=(1,1,2)_dimensionnumbers=('NDHWC','DHWIO','NDHWC')_featuregroupcount=1_batchgroupcount=1_precision=None_enablexla=False":
+      raise unittest.SkipTest("TODO: known but unidentified bug in compiled "
+                              "mode")
+
+    convert_and_compare = partial(
+        self.ConvertAndCompare, harness.dyn_fun,
+        *harness.dyn_args_maker(self.rng()), atol=tol, rtol=tol,
+        enable_xla=harness.params["enable_xla"])
+
+    if harness.params["expect_xla_disabled_exception"]:
```
gnecula, line 771: If it weren't for the above cases when `expect_xla_disabled_exception=True`, wouldn't it be better to just 
set up a special harness that is only testing this exception and have a separate test for it? Then we don't need to pass around the `expect_xla_disabled_exception`?

jax/experimental/jax2tf/tests/primitive_harness.py
bchetioui: These cases where we expect an exception were intended as checks that we properly fell back to XLA when the TF conversion could not be achieved. Since there is no scenario when we fallback anymore, it is not that useful. I will remove them.

jax/experimental/jax2tf/tests/primitive_harness.py
bchetioui: See above.

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -441,6 +441,20 @@ def _make_scatter_harness(name, *, shape=(5,), f_lax=lax.scatter_min,
   for unique_indices in [False]
 )
 
+disable_xla = tuple(
```
gnecula, line 444: I like this!

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1191,11 +1208,13 @@ def _make_conv_harness(name, *, lhs_shape=(2, 3, 9, 10), rhs_shape=(3, 3, 4, 5),
     # TODO(bchetioui): the NCW data format is not supported on CPU for TF
     # for now. That path is thus disabled to allow the code to use XLA instead.
   ]
+  for enable_xla in [False, True]
```
gnecula, line 1211: I like this as well: all tests either use the default enable_xla=True or try both cases.

PR #4782, opened 2020-11-04
[jax2tf] Add support for cummin/cummax.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1662,6 +1661,13 @@ def _get_min_identity(tf_dtype):
 tf_impl_with_avals[lax.reduce_window_p] = _reduce_window
 # pylint: enable=protected-access
 
+tf_impl_with_avals[lax_control_flow.cummin_p] = _convert_jax_impl(
+    functools.partial(lax_control_flow._cumred_tpu_translation_rule,
```
gnecula, line 1665: I am wondering why in JAX we use this for TPU only, and use associative scan for CPU and GPU. I don't see any exceptions here, so it works on all platforms? Perhaps it is faster on CPU/GPU to use the associative scan?

bchetioui: Yes, this is not efficient on all backends! There is the following comment in `_cumred_tpu_translation_rule`:
```
# On TPU, an implementation using reduce_window is handled specially by the
# compiler and is efficient. On other backends, it is O(n^2)
```

PR #4788, opened 2020-11-04
Promotion via lattice

jakevdp:
Now that JAX type promotion is derived from a lattice rather than specified in a table, we could use the lattice directly and avoid pre-computation of the table as suggested by @hawkinsp in review of #4748. This PR explores that possibility.

The main potential downside is greater overhead in determining output types, but it turns out that with careful cacheing, the lattice approach has slightly faster amortized performance.

TLDR; micro-benchmarks show:

- import time precomputation goes from 350 µs to 160 µs
- first-time type promotion is a factor of 2 slower than current table lookup
- subsequent type promotions with `lru_cache` are 10% faster than current table lookup

Timing on the master branch for retrieving the promotion table:
```python
In [1]: import numpy as np 
   ...: from jax import dtypes, config 
   ...: config.update('jax_enable_x64', True) 
   ...: %timeit [dtypes.result_type(a, b) for a in dtypes._jax_types for b in dtypes._jax_types]                                                                           
1.59 ms ± 22.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Amortized time on this branch for computing the promotion table:
```python
In [1]: import numpy as np 
   ...: from jax import dtypes, config 
   ...: config.update('jax_enable_x64', True) 
   ...: %timeit [dtypes.result_type(a, b) for a in dtypes._jax_types for b in dtypes._jax_types]                                                                           
1.43 ms ± 14.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
In this PR, the table of lattice upper bounds is computed on import. This is the cost of this operation:
```python  
In [1]: from jax import dtypes                                                                                                                                             
In [2]: %timeit dtypes._make_lattice_upper_bounds()                                                                                                                        
161 µs ± 4.06 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
This is about 2x as fast as the current method that constructs the full promotion table on import:
```python
In [1]: from jax import dtypes                                                                                                                                             

In [2]: %timeit dtypes._make_type_promotion_table()                                                                                                                        
349 µs ± 5.04 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Removing `functools.lru_cache` from `_least_upper_bound` shows that the average first-time type promotion cost is only about 40% greater than the cost of a cached type promotion:
```python
In [1]: import numpy as np 
   ...: from jax import dtypes, config 
   ...: config.update('jax_enable_x64', True) 
   ...: %timeit [dtypes.result_type(a, b) for a in dtypes._jax_types for b in dtypes._jax_types]                                                                           
1.86 ms ± 24.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
All told, I think this lattice-based approach is a win in terms of simplicity of implementation, and avoids the excessive overhead that I feared would make it a non-starter.

### DIFF
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -166,10 +166,8 @@ can_cast = np.can_cast
 issubsctype = np.issubsctype
 
 
-# List of all valid JAX dtypes, in the order they appear in the type promotion
-# table.
+# Enumeration of all valid JAX types in order.
 _weak_types = [int, float, complex]
-
 _jax_types = [
   np.dtype('bool'),
   np.dtype('uint8'),
@@ -190,7 +188,7 @@ _jax_types = [
 
 def _jax_type(value):
   """Return the jax type for a value or type."""
-  # Note: `value in _weak_types` can return false positives due to dtype comparator overloading.
+  # Note: `x in _weak_types` can return false positives due to dtype comparator overloading.
   if any(value is typ for typ in _weak_types):
     return value
   dtype_ = dtype(value)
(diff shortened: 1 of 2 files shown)


hawkinsp:
Note my preference was to leave the table as a literal in `dtypes.py` and calculate it in the test (only).

jakevdp:
OK - I got a different impression from https://github.com/google/jax/pull/4744#discussion_r516973503

hawkinsp:
Well, the thing I would still like is not to have two explicitly written out copies of the type promotion table. This was one of the two possible ways to achieve that.

jakevdp:
There are not two copies of the promotion table. There is one copy of the internal promotion table in `dtypes.py`, and there are two different tables in `dtypes_test.py` reflecting the output of `result_type`, both of which differ in important ways from the internal table in `dtypes.py`.

jakevdp:
I changed to a different cacheing approach which makes the micro benchmarks much more competetive with the status quo. Please take a look at the updated description!

jax/dtypes.py
```diff
@@ -200,37 +198,39 @@ def _jax_type(value):
...
+      if not new_upper_bounds.difference(upper_bounds[n]):
+        break
+      upper_bounds[n] |= new_upper_bounds
+  return upper_bounds
+_lattice_upper_bounds = _make_lattice_upper_bounds()
+
+@functools.lru_cache(512)
+def _least_upper_bound(*nodes):
+  common_upper_bounds = set.intersection(*(_lattice_upper_bounds[n] for n in nodes))
+  least_upper_bounds = common_upper_bounds - set.union(*(_lattice_upper_bounds[n] - {n}
```
gnecula, line 228: It is not obvious to me when reading this that it computes LUB. If CUB(a, b) are the common upper bounds of a and b, then LUB(a, b) is defined as {c in CUB(a, b) | forall d in CUB(a, b). c <= d}. By definition this set is either empty or  a singleton (if <= is an order). 

It would be more satisfying if the code were more obviously correct. BTW, if we pre-compute the LUB before we start then we can afford to have a less efficient but more obviously correct implementation. 

Perhaps this is an academic question, since this code will not be used to compute LUB for arbitrary lattices, and I am sure it works for our lattice.

jax/dtypes.py
```diff
@@ -200,37 +198,39 @@ def _jax_type(value):
...
   return dtype_
 
-# Mapping from types to their type numbers.
-_jax_type_nums = {t: i for i, t in enumerate(_jax_types)}
-
-
-def _make_type_promotion_table():
-  # Note: this promotion table is generated via the least upper bounds over a type
-  # promotion lattice. See testPromotionTableLattice for details of this.
+def _type_promotion_lattice():
```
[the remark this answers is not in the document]
jakevdp: I have a type promotion documentation overhaul separately in the works.

jax/dtypes.py
```diff
@@ -200,37 +198,39 @@ def _jax_type(value):
...
+      new_upper_bounds = set().union(*(lattice[b] for b in upper_bounds[n]))
+      if n in new_upper_bounds:
+        raise ValueError(f"cycle detected in type promotion lattice for node {n}")
+      if not new_upper_bounds.difference(upper_bounds[n]):
+        break
+      upper_bounds[n] |= new_upper_bounds
+  return upper_bounds
+_lattice_upper_bounds = _make_lattice_upper_bounds()
+
+@functools.lru_cache(512)
```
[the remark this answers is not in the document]
jakevdp: That's a possibility - it would add an additional ~2.5ms to JAX's import overhead (by comparison, numpy's total import time is measured in microseconds) and would not cover `result_type` with 3 or more inputs, so I judged that lazy cacheing would be better.

jax/dtypes.py
```diff
@@ -200,37 +198,39 @@ def _jax_type(value):
...
+def _least_upper_bound(*nodes):
```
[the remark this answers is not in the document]
jakevdp: I think it's better to use a single computation. Here are representative timings without cacheing, indicating the cost of the first computation:
```python
%timeit dtypes.result_type(np.uint8, np.int32)                                                                                  
9.94 µs ± 427 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit dtypes.result_type(np.uint8, np.int32, np.float16) 
10.9 µs ± 130 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit dtypes.result_type(dtypes.result_type(np.uint8, np.int32), np.float16)    
17.8 µs ± 86.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
And here are the timings with cacheing, incidating the cost of subsequent computations over the same values:
```python
%timeit dtypes.result_type(np.uint8, np.int32)                                                                                
5.32 µs ± 97.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit dtypes.result_type(np.uint8, np.int32, np.float16) 
7.62 µs ± 53.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit dtypes.result_type(dtypes.result_type(np.uint8, np.int32), np.float16)      
10.7 µs ± 99.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
Even the first time a trinomial `result_type` is computed, the cost is comparable to that of two cache lookups when computing it pairwise. Subsequent computations for the same inputs are then 20% faster.

jax/dtypes.py
```diff
@@ -200,37 +198,39 @@ def _jax_type(value):
...
-_type_promotion_table = _make_type_promotion_table()
+  return {
+    b1: [i_],
+    u1: [i2, u2], u2: [i4, u4], u4: [i8, u8], u8: [f_],
+    i_: [u1, i1], i1: [i2], i2: [i4], i4: [i8], i8: [f_],
+    f_: [bf, f2, c_], bf: [f4], f2: [f4], f4: [f8, c4], f8: [c8],
+    c_: [c4], c4: [c8], c8: [],
+  }
+
+def _make_lattice_upper_bounds():
```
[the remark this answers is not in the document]
jakevdp: Yes, I believe that's correct.

jax/dtypes.py
jakevdp: I believe this code does exactly that. This is the reasoning:
- `c < d` if and only if `d ∈ UB(c) - {c}`
- Therefore, any node `d ∈ UB(c) - {c}` given `c ∈ CUB(a,b)` is greater than some other node in `CUB(a,b)`, and thus cannot be the least upper bound.
- It follows that removing all such nodes `d` from `CUB(a, b)` leaves only the least upper bound, so long as the input graph is actually a lattice.

If there is more than one remaining node, then the two are unorderable, and the input graph must not be a lattice.

jakevdp: I added a comment describing this reasoning... let me know if it's clear.

jax/dtypes.py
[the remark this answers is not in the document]
jakevdp: See #4806

jax/dtypes.py
gnecula: BTW, I think we can merge this. The discussion at this point is more fun than useful. 

But the proof is still tricky. LUB is by definition the unique element in CUB that is <= all CUB. In bullet 2 above you subtract from CUB nodes `d` that cannot be LUB. But I do not think that it follows that if you are left with a single node it must be LUB (i.e., bullet 3 is not obviously true). 

You would need to remove from CUB all nodes `d` such that there exists `c` in CUB with `not (d <= c)`. It is true that if `c < d` then `not (d <= c)` but not the other way around (unless you have a total order). 

The last statement "If there is more than one remaining node, then the two are unorderable, and the input graph must not be a lattice" also I do not think that is strictly true.

jakevdp: Hmmm.... I'm out of practice in my proof formalisms :grin:. I'll think about this some more, but I think we can somehow show the third point so long as `<=` is transitive across partially ordered sets.

jakevdp: OK, how about this:
- We want `LUB(a, b) = {c ∈ CUB(a, b) | forall d in CUB(a, b). c <= d}`
- `c <= d` implies `d ∈ UB(c)` (I think this is true, otherwise `c` and `d` are unorderable?)
- So we want `LUB(a, b) = {c ∈ CUB(a, b) | forall d in CUB(a, b). d ∈ UB(c)}`
- In other words `LUB(a, b) = {c ∈ CUB(a, b) | cardinality[CUB(a, b)] = cardinality[UB(c)]}`
So, all we need is to compute the cardinality of each node `c` in `CUB(a, b)` and choose the one whose upper bounds have the same cardinality as `CUB(a, b)`.

jakevdp: Alright, I think I landed on an approach that's more obviously correct, and is about 20% faster to boot. See the explanation in the code.

gnecula: Your second bullet is true by definition of UB. 

Very nice, I believe this proof.

jax/dtypes.py
```diff
@@ -200,37 +198,61 @@ def _jax_type(value):
...
+  #   LUB(N) ≡ {c ∈ CUB(N) | ∀ d ∈ CUB(N), c ≤ d}
+  # The definition of an upper bound implies that c ≤ d if and only if d ∈ UB(c),
+  # so the LUB can be expressed:
+  #   LUB(N) = {c ∈ CUB(N) | ∀ d ∈ CUB(N): d ∈ UB(c)}
+  # or, equivalently:
+  #   LUB(N) = {c ∈ CUB(N) | CUB(N) ⊆ UB(c)}
+  # By definition, LUB(N) has a cardinality of 1 for a partially ordered set.
+  # Note a potential algorithmic shortcut: from the definition of CUB(N), we have
+  #   ∀ c ∈ N: CUB(N) ⊆ UB(c)
+  # So if N ∪ CUB(N) is nonempty, if follows that LUB(N) = N ∪ CUB(N).
```
gnecula, line 247: I think that you want intersection here, not union.

jakevdp: Yep, thanks. Fortunately the code is correct 😁

gnecula: Brilliant!

jakevdp: Thanks! Comment is updated.

PR #4796, opened 2020-11-05
Add a named_call public function in lax

qiuminxu:
Add a named_call public function in lax.

The implementation copies from the flax [_named_call](https://github.com/google/flax/blob/bb3170165010d508e7950d2c66eb9846d3661471/flax/core/lift.py#L674)

[Tests](https://github.com/deepmind/dm-haiku/blob/3fedc4725da8f58b4dae80bb21b195b0e3d9b670/haiku/_src/named_call_test.py#L25) passed after replacing the haiku [named_fun](https://github.com/deepmind/dm-haiku/blob/3fedc4725da8f58b4dae80bb21b195b0e3d9b670/haiku/_src/named_call.py#L109) with this implementation.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2449,25 +2449,6 @@ ad.defjvp2(min_p,
            lambda g, ans, x, y: mul(_brcast(g, y), _balanced_eq(x, ans, y)),
            lambda g, ans, x, y: mul(_brcast(g, x), _balanced_eq(y, ans, x)))
 
-named_call_p = core.CallPrimitive('named_call')
-named_call_p.def_impl(core.call_impl)
-ad.primitive_transposes[named_call_p] = functools.partial(ad.call_transpose,
-                                                         named_call_p)
-
-
-def _named_call_translation_rule(c, axis_env, in_nodes, name_stack,
-                                 *, name='core_call', backend, call_jaxpr):
-  subc = xla.xb.make_computation_builder(name)
-  args = [xla.xb.parameter(subc, i, c.GetShape(n))
-          for i, n in enumerate(in_nodes)]
-  out_nodes = xla.jaxpr_subcomp(subc, call_jaxpr, backend, axis_env, (),
-                                jax.util.extend_name_stack(name_stack, name),
-                                *args)
-  subc = subc.Build(xla.xops.Tuple(subc, out_nodes))
-  return xla.xops.Call(c, subc, list(in_nodes))
-
-xla.call_translations[named_call_p] = _named_call_translation_rule
-
 shift_left_p = standard_naryop([_int, _int], 'shift_left')
 ad.defjvp_zero(shift_left_p)
 
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -32,6 +32,7 @@ Just-in-time compilation (:code:`jit`)
     make_jaxpr
     eval_shape
     device_put
+    named_call
 
 Automatic differentiation
 -------------------------
--- a/jax/__init__.py
+++ b/jax/__init__.py
@@ -58,6 +58,7 @@ from .api import (
   linear_transpose,
   make_jaxpr,
   mask,
+  named_call,
   partial,  # TODO(phawkins): update callers to use functools.partial.
   pmap,
   pxla,  # TODO(phawkins): update users to avoid this.
--- a/jax/api.py
+++ b/jax/api.py
@@ -2276,6 +2276,48 @@ def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
 remat = checkpoint
 
 
+def named_call(
+    fun: Callable[..., Any],
+    *,
+    name: Optional[str] = None,
+) -> Callable[..., Any]:
+  """Adds a user specified name to a function when staging out JAX computations.
+
+  When staging out computations for just-in-time compilation to XLA (or other
+  backends such as TensorFlow) JAX runs your Python program but by default does
+  not preserve any of the function names or other metadata associated with it.
+  This can make debugging the staged out (and/or compiled) representation of
+  your program complicated because there is limited context information for each
+  operation being executed.
+
+  `named_call` tells JAX to stage the given function out as a subcomputation
+  with a specific name. When the staged out program is compiled with XLA these
+  named subcomputations are preserved and show up in debugging utilities like
+  the TensorFlow Profiler in TensorBoard. Names are also preserved when staging
+  out JAX programs to TensorFlow using :func:`experimental.jax2tf.convert`.
+
+  Args:
+    fun: Function to be wrapped. This can be any Callable.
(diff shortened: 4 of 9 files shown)


qiuminxu:
cc @froystig, @LenaMartens @tomhennigan @trevorcai @levskaya @jheek @avital

Thanks @LenaMartens for verifying this implementation also has raised the abstraction level of the arguments. Looks like the two implementation are equivalent. Are we good to upstream this implementation or would you prefer to copy the implementation to double check and then upstream?

tomhennigan:
If @LenaMartens is happy I'm happy 😄

jax/_src/lax/lax.py
```diff
@@ -2468,6 +2468,16 @@ def _named_call_translation_rule(c, axis_env, in_nodes, name_stack,
 
 xla.call_translations[named_call_p] = _named_call_translation_rule
 
+
+def named_call(f, name):
```
tomhennigan, line 2472: Doc string? This is a new public symbol after all 😄 

We should also be adding this to https://github.com/google/jax/blob/master/docs/jax.lax.rst.

tomhennigan (changes requested):
Looks good to me, but documentation would be very welcome.

jax/_src/lax/lax.py
qiuminxu: Done.

docs/jax.lax.rst
```diff
@@ -162,6 +162,14 @@ Custom gradient operators
     custom_linear_solve
     custom_root
 
+Function operators
```
tomhennigan, line 165: Is `named_call` an operator (I understand that the symbol is a function that operates on functions, but since it is a higher order function I find the word "operator" a bit confusing, might just be me..)?

If we keep this in `jax.lax` would put this at the bottom of the file under a heading `Debugging` or `Compilation`. Alternatively we could move this to `jax.named_call` and then document it under `Just-in-time-compilation (jit)`. I think that might actually make the most sense since `named_call` is only useful once JAX programs are staged out.

@gnecula wdyt?

jax/_src/lax/lax.py
```diff
@@ -2468,6 +2468,33 @@ def _named_call_translation_rule(c, axis_env, in_nodes, name_stack,
 
 xla.call_translations[named_call_p] = _named_call_translation_rule
 
+
+def named_call(
+    fun: Callable[..., Any],
+    *,
+    name: Optional[str] = None,
+) -> Callable[..., Any]:
+  """Wraps a function in a name_scope with the provided name.
```
tomhennigan, line 2477: JAX does not have the concept of a "name scope" and we should not assume users are familiar with TF before using JAX. I think the important thing to call out is that named call is used to add names to staged jax programs.

Here's a concrete suggestion of a perhaps more useful doc string:

```
Adds a user specified name to a function when staging out JAX computations.

When staging out computations for just-in-time compilation to XLA (or other
backends such as TensorFlow) JAX runs your Python program but by default does
not preserve any of the function names or other metadata associated with it.
This can make debugging the staged out (and/or compiled) representation of your
program complicated because there is limited context information for each
operation being executed.

`named_call` tells JAX to stage the given function out as a subcomputation with
a specific name. When the staged out program is compiled with XLA these named
subcomputations are preserved and show up in debugging utilities like the
TensorFlow Profiler in TensorBoard. Names are also preserved when staging out
JAX programs to TensorFlow using :func:`experimental.jax2tf.convert`.
```

jax/_src/lax/lax.py
```diff
@@ -2468,6 +2468,33 @@ def _named_call_translation_rule(c, axis_env, in_nodes, name_stack,
...
+    within the name scope. Use the fun.__name__ if not specified.
+
+  Returns:
+    A version of `fun` that is wrapped in a name_scope.
+  """
+  if name is None:
+    name = fun.__name__
+
+  _, in_tree = jax.tree_flatten(())
+  def named_f(*args, **kwargs):
```
tomhennigan, line 2491: nit: Should we `@functools.wraps(fun)` to preserve docstring et al?

tomhennigan (changes requested):
Looks good, final comments around documentation and location.

docs/jax.lax.rst
gnecula: I like the idea of putting this in `jax` instead of `lax`.

jax/_src/lax/lax.py
gnecula: I like this. Ideally we should have a documentation of how to interpret the name_stack that shows up in xprof.

jax/_src/lax/lax.py
qiuminxu: Done.

jax/_src/lax/lax.py
qiuminxu: Thanks for the doc string, it looks very nice! 

> I like this. Ideally we should have a documentation of how to interpret the name_stack that shows up in xprof.

Should we add a screenshot and some explanations in this page? https://jax.readthedocs.io/en/latest/profiling.html?

docs/jax.lax.rst
qiuminxu: Done. I created a `named_call.py` under `jax/_src` which has the implementation, and another `jax/named_call.py` that only contains only imports. Then I exposed both symbols directly under jax (`jax.named_call` and `jax.named_call_p` )

gnecula:
I am not sure it is a good idea to have named_call in its own file. It is a rather small feature to deserve a file on its own. Perhaps it should go into `api`, even though it is kind of large? @hawkinsp has reorganized the files recently, at least partially. I am wondering what he thinks is the right place for this going forward. 

Since you added `named_call_test.py` this PR will need some changes to the tests/BUILD file. If, however, we merge named_call somewhere else, then it makes sense to merge the tests also.

froystig:
Let's add the implementation (definitions of primitive and rules) to `core.py`, right below the corresponding implementation of `call_p`. We can then import `named_call` in `api.py` so that it is available externally as `jax.named_call`.

qiuminxu:
> I am not sure it is a good idea to have named_call in its own file. It is a rather small feature to deserve a file on its own. Perhaps it should go into `api`, even though it is kind of large? @hawkinsp has reorganized the files recently, at least partially. I am wondering what he thinks is the right place for this going forward.
> 
> Since you added `named_call_test.py` this PR will need some changes to the tests/BUILD file. If, however, we merge named_call somewhere else, then it makes sense to merge the tests also.

Done. I moved named_call_p close to call_p. Also added named_call public function in api.py. Tests are also added to api_test.py.

qiuminxu:
@tomhennigan @gnecula @froystig Does the latest version look good?

tomhennigan (approved):

froystig (approved):

gnecula (approved):
Thanks!

PR #4806, opened 2020-11-05
DOC: add lattice to type promotion doc

jakevdp:
Preview of the new docs here: https://jax--4806.org.readthedocs.build/en/4806/type_promotion.html

### DIFF
--- a/docs/type_promotion.rst
+++ b/docs/type_promotion.rst
@@ -4,15 +4,45 @@ Type promotion semantics
 ========================
 
 JAX's type promotion rules (i.e., the result of
-:func:`jax.numpy.promote_types` for each pair of types) are given by the
-following table, where, for example
+:func:`jax.numpy.promote_types` for each pair of types) is determined via
+the following type promotion lattice:
 
-* "b1" means :code:`np.bool_`,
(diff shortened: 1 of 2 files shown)


docs/type_promotion.rst
```diff
@@ -5,14 +5,42 @@ Type promotion semantics
...
 * "b1" means :code:`np.bool_`,
-* "s2" means :code:`np.int16`,
+* "i2" means :code:`np.int16`,
 * "u4" means :code:`np.uint32`,
 * "bf" means :code:`np.bfloat16`,
-* "f2" means :code:`np.float16`, and
-* "c8" means :code:`np.complex128`.
+* "f2" means :code:`np.float16`,
+* "c8" means :code:`np.complex128`,
+* "i*" means Python :code:`int`,
```
gnecula, line 16: Perhaps we can use i_ like in the code?

gnecula (approved):

docs/type_promotion.rst
jakevdp: I prefer `i*` because using the standard wildcard character communicates that the precision is variable.

PR #4812, opened 2020-11-06
[DOCS] Improve docs for custom_jvp and custom_vjp

j-towns:
Correct the `custom_jvp` docstring to include the `defjvps` instance method. Add the `defjvp`/`defvjp` instance methods to the `jax.rst` sphinx doc.

### DIFF
--- a/jax/custom_derivatives.py
+++ b/jax/custom_derivatives.py
@@ -79,10 +79,18 @@ class custom_jvp:
   This class is meant to be used as a function decorator. Instances are
   callables that behave similarly to the underlying function to which the
   decorator was applied, except when a differentiation transformation (like
-  :py:func:`jax.jvp` or :py:func:`jax.grad`) is applied, in which case a custom user-supplied
-  JVP rule function is used instead of tracing into and performing automatic
-  differentiation of the underlying function's implementation. There is a single
-  instance method, ``defjvp``, which defines the custom JVP rule.
+  :py:func:`jax.jvp` or :py:func:`jax.grad`) is applied, in which case a custom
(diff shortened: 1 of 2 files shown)


jax/custom_derivatives.py
```diff
@@ -79,10 +79,18 @@ class custom_jvp:
...
-  JVP rule function is used instead of tracing into and performing automatic
-  differentiation of the underlying function's implementation. There is a single
-  instance method, ``defjvp``, which defines the custom JVP rule.
+  :py:func:`jax.jvp` or :py:func:`jax.grad`) is applied, in which case a custom
+  user-supplied JVP rule function is used instead of tracing into and
+  performing automatic differentiation of the underlying function's
+  implementation.
+
+  There are two instance methods available for defining the custom JVP rule:
+  :py:func:`~jax.custom_jvp.defjvp` for defining a *single* custom JVP rule for
```
gnecula, line 88: I am curious what is the role of "~" in the function reference?

gnecula (approved):

jax/custom_derivatives.py
j-towns: It’s documented [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-syntax):

> If you prefix the content with ~, the link text will only be the last component of the target. For example, ``:py:meth:`~Queue.Queue.get` `` will refer to `Queue.Queue.get` but only display `get` as the link text.

I did that because it’s obvious that the links are to the class’s instance methods, so no need for the full name.

PR #4815, opened 2020-11-06
[DOCS] [cosmetic] Add JAX favicon to sphinx docs

j-towns:
Add this 64px x 64px favicon to the docs: 
favicon


shoyer:
You mean like this?

image

I guess this still might be a little hard to read at favicon size, though.

j-towns:
Thanks! I don’t think it needs to be readable, just recognisable. I’ll try out the version you just sent and see if it looks better.

j-towns:
That looks a tad better in my opinion.

shoyer (approved):

PR #4816, opened 2020-11-06
[jax2tf] Add the convert_and_save_model to the jax2tf API.

gnecula:


gnecula:
@tomhennigan I have tried to write down a recipe for how one would save a model from Haiku. PTAL.

jax/experimental/jax2tf/README.md
```diff
@@ -76,12 +81,62 @@ my_model.f(tf.ones([16, 28, 28]))  # a batch size of 16
 tf.saved_model.save(my_model, '/some/directory')
 ```
 
-More involved examples of using SavedModel, including how to prepare
-Flax models for conversion and saving, are described in the
+### Usage: saved models from Flax
```
tomhennigan, line 84: Shouldn't this section be in the Haiku or Flax documentation instead of in JAX?

gnecula: Perhaps it should. Maybe it will migrate there. For now, I feel that it makes sense for it to be hosted here, because from a Flax/Haiku perspective jax2tf is such a niche, and so fluid, and more intertwined with the way the `convert_and_save_model` works than with the libraries themselves.

gnecula: Is what I wrote about how to split a Haiku model into a predict function and a pytree of parameters correct?

jax/experimental/jax2tf/README.md
```diff
@@ -76,12 +81,62 @@ my_model.f(tf.ones([16, 28, 28]))  # a batch size of 16
...
+pass it to `model.apply` as multiple inputs.)
+
+If you are using Haiku, then the recipe is along these lines:
+
+  ```
+  model_fn = ...define your Haiku model...
+  net = hk.transform(model_fn)  # get a pair of init and apply functions
+  params = ... train your model starting from net.init() ...
+  net_no_rng = hk.without_apply_rng(net)
+  predict_fn = net_no_rng.apply
```
tomhennigan, line 135: nit: you can make this a one liner:

```
predict_fn = hk.without_apply_rng(net).apply
```

jax/experimental/jax2tf/saved_model_lib.py
```diff
@@ -101,17 +87,17 @@ def jax_fn_for_save_model(params, inputs):
...
+    in_shapes=[None, shape_polymorphic_input_spec],
+    enable_xla=enable_xla)
 
   # Create tf.Variables for the parameters.
   param_vars = tf.nest.map_structure(
-      # If with_gradient=False, we mark the variables behind as non-trainable,
-      # to ensure that users of the SavedModel will not try to fine tune them.
-      lambda param: tf.Variable(param, trainable=with_gradient),
-      params)
+    # If with_gradient=False, we mark the variables behind as non-trainable,
```
tomhennigan, line 97: This seems odd to me, if the model is not trainable wouldn't it be better to keep the params as tensors baked into the graph (e.g. `if not with_gradient: param_vars = params`)? The typical use for non-trainable variables is for things like moving averages (not for model parameters) so I am a bit surprised to see them used in this way.

Using non-trainable tensors here may also impact how models that use non-trainable variables (e.g. batch norm) need to be handled. Right now this API only supports closing over batch norm state (which I believe will be typically fine for inference).

jax/experimental/jax2tf/saved_model_lib.py
```diff
@@ -21,47 +21,32 @@
...
-               params,
-               model_dir: str,
-               *,
-               input_signatures: Sequence[tf.TensorSpec],
-               shape_polymorphic_input_spec: Optional[str] = None,
-               with_gradient: bool = False,
-               enable_xla: bool = True,
-               compile_model: bool = True):
-  """Saves the SavedModel for a function.
+def convert_and_save_model(jax_fn: Callable,
```
tomhennigan, line 28: We can make the definition here more concrete, for example we require a callable taking two args `Callable[[Any, Any], Any]`.

jax/experimental/jax2tf/saved_model_lib.py
```diff
@@ -21,47 +21,32 @@
...
+                           params,
+                           model_dir: str,
+                           *,
+                           input_signatures: Sequence[tf.TensorSpec],
```
tomhennigan, line 32: Bit confused by this, each signature should be a Pair and the elements of the pair should be optionally nested tf.TensorSpec no?

    Params = Inputs = Nest[tf.TensorSpec]
    PredictSignature = Tuple[Params, Inputs]
    PredictSignatures = Sequence[PredictSignature]

jax/experimental/jax2tf/saved_model_lib.py
```diff
@@ -21,47 +21,32 @@
...
+                           shape_polymorphic_input_spec: Optional[str] = None,
+                           with_gradient: bool = False,
+                           enable_xla: bool = True,
+                           compile_model: bool = True,
+                           save_model_options=None):
```
tomhennigan, line 37: We can hint this: `: tf.saved_model.SaveOptions`

jax/experimental/jax2tf/saved_model_lib.py
tomhennigan: This also makes me wonder, do we need to support models that want randomness as part of their inference (e,g, `predict_fn(params, rng, x)`)? Are there other signatures we might need to support in the future? It feels to me like the scope of this function might grow..

jax/experimental/jax2tf/saved_model_lib.py
```diff
@@ -101,17 +87,17 @@ def jax_fn_for_save_model(params, inputs):
...
-      with_gradient=with_gradient,
-      in_shapes=[None, shape_polymorphic_input_spec],
-      enable_xla=enable_xla)
+    jax_fn,
+    with_gradient=with_gradient,
+    in_shapes=[None, shape_polymorphic_input_spec],
+    enable_xla=enable_xla)
 
   # Create tf.Variables for the parameters.
   param_vars = tf.nest.map_structure(
```
tomhennigan, line 96: Since we're using nest, I'd suggest using `map_structure_with_path` (I'm not sure if this is exposed in TF2 so you might need to use `import tensorflow.compat.v1 as tf1` or `dm-tree`) here and including a name for the variables based on their path. Could be a follow up, but otherwise the default names of `Variable_{1..N}` will not be super useful.

tomhennigan (approved):
Don't block on my comments if you disagree!

jax/experimental/jax2tf/saved_model_lib.py
gnecula: Done

jax/experimental/jax2tf/README.md
gnecula: Done

jax/experimental/jax2tf/saved_model_lib.py
gnecula: The input signatures are just for the input, not including the parameters. 

I wish I can use `Nest[tf.TensorSpec]` notation, but I don't know of an easy way to define the `Nest` generic hint, so I just use the type hint as if we have a leaf nest.

jax/experimental/jax2tf/README.md
gnecula: This touches on some organization and API aspects that I am still trying to optimize. On one hand, I think it would be strategically important for jax2tf to position itself as completely independent of SavedModel: you get a TF function and you package it as a SavedModel entirely as you wish. This is important because it addresses a lot of the worries of the form "will JAX SavedModel support X kind of assets or input_signatures or metadata"? Hence, my initial thinking was to NOT have an API that creates SavedModel. 

However, when constructing examples to work with Keras, TF Hub, TF.js, and Servo, in combination with Flax, it took quite a bit of time to figure out how to put everything together. It was in this process that I wrote the saved_model_lib.py, to be used in the examples. Last week I had to send twice over email the information for how to package Flax models for use with SavedModel. I think that I owe to our users to write down how this can be done. And I do think that *at the moment* the `convert_and_saved_model` documentation and how it is used with Flax and Haiku is too temporary to deserve being in the Flax or Haiku documentation. 

This through this comment makes me think to not include saved_model_lib.py as part of jax2tf library, but put it in jax2tf/examples. The documentation of how to use it with Flax and Haiku also goes in the examples for now. Eventually, we can migrate the documentation to Flax and Haiku. I would also prefer to not even export the library in google3, and tell people to just copy the code.

jax/experimental/jax2tf/saved_model_lib.py
gnecula: I fixed the type hint for `jax_fn`. 

You are right that the scope of this function will grow. Or, we can decide that this function is just a convenience for the simple cases. For the more complicated cases, you should roll your own SavedModel.

jax/experimental/jax2tf/saved_model_lib.py
gnecula: I thought that even for non-trainable variables it is better to keep them separate from the graph as a matter of SavedModel organization. Perhaps it would only make a difference when the size of the variables are larger than the proto size. 

If the user wants to keep the parameters baked into the graph, they can do it before calling this function: pass [] for params. 

BTW, I have not looking into how this would be used with BatchNorm.

tomhennigan: That makes sense to me, I would still suggest `params` are represented as `trainable` (even if we disallow gradients) just to allow us to represent non-trainable mutable values in the future.

The `trainable` bit in TensorFlow is a bit of a weird beast, for example in Keras if you mark a `Layer` as non-trainable, it's parameters still have `trainable=True` (since this bit is immutable and set at construction time). Personally I think of this bit as `parameter=True`.

```python
import tensorflow as tf
d = tf.keras.layers.Dense(10)
d(tf.ones([1, 1]))

assert len(d.trainable_variables) == 2
assert len(d.variables) == 2

d.trainable = False  # Layer.trainable != Variable.trainable
assert len(d.trainable_variables) == 0
assert len(d.variables) == 2
assert all(v.trainable for v in d.variables)
```

jax/experimental/jax2tf/README.md
tomhennigan: It makes sense while you are iterating on this feature and it is alpha for the documentation to be in one place, I think once it is stable that we should move the documentation out of JAX and into the respective libraries (we can always x-link from JAX if that is useful).

Perhaps longer term the JAX+TF utility layer (e.g. **not** jax2tf, but things built on top of that) could be a separate library that can be evolved using the public API.

jax/experimental/jax2tf/saved_model_lib.py
gnecula: `map_structure_with_path` is not exported as part of `tf.nest` in TF2. I just added a comment that one can use `dm-tree`

jax/experimental/jax2tf/saved_model_lib.py
gnecula: It turns out that Keras insists on wanting to take gradients through these variables even if the layer is marked as non-trainable.

PR #4822, opened 2020-11-07
jax.scipy.signal functions for n-dim arrays

### LINKED ISSUE #4775: jax.scipy.signal.convolve should support n-dimensional inputs
`scipy.signal.convolve()` supports N-dimensional inputs; the JAX equivalent should as well.

There is already a `_convolve_nd` utility function in the module; this would involve calling that, and writing a test case to ensure the outputs match scipy's.

ashridh:
Fixes: #4775 .

Changes include:

1. `jax.scipy.signal.correlate` was giving incorrect results for tensors with ndim >= 2. Fixed this.
2. Added `_convolve_nd` in definition of `convolve()` and `correlate()`
3. Added tests for running these functions with 2d arrays.
4. Fix a small error where `jnp.scipy.signal.correlate2d` was wrapping `scipy.signal.correlate`.

### DIFF
--- a/tests/scipy_signal_test.py
+++ b/tests/scipy_signal_test.py
@@ -29,6 +29,7 @@ config.parse_flags_with_absl()
 
 onedim_shapes = [(1,), (2,), (5,), (10,)]
 twodim_shapes = [(1, 1), (2, 2), (2, 3), (3, 4), (4, 4)]
+threedim_shapes = [(2, 2, 2), (3, 3, 2), (4, 4, 2), (5, 5, 2)]
 
 
 default_dtypes = jtu.dtypes.floating + jtu.dtypes.integer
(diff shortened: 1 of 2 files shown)


tests/scipy_signal_test.py
```diff
@@ -49,10 +49,12 @@ class LaxBackedScipySignalTests(jtu.JaxTestCase):
       for mode in ['full', 'same', 'valid']
       for op in ['convolve', 'correlate']
       for dtype in default_dtypes
-      for xshape in onedim_shapes
-      for yshape in onedim_shapes))
+      for xshape in onedim_shapes + twodim_shapes
+      for yshape in onedim_shapes + twodim_shapes))
```
jakevdp, line 53: It would be worth also testing some three-dimensional shapes as well.

jakevdp (commented):
Looks great, thanks for taking this on! Just a few more tests and I think this will be good to go.

tests/scipy_signal_test.py
```diff
@@ -49,10 +50,12 @@ class LaxBackedScipySignalTests(jtu.JaxTestCase):
...
       for op in ['convolve', 'correlate']
       for dtype in default_dtypes
-      for xshape in onedim_shapes
-      for yshape in onedim_shapes))
+      for xshape in onedim_shapes + twodim_shapes + threedim_shapes
+      for yshape in onedim_shapes + twodim_shapes + threedim_shapes))
   def testConvolutions(self, xshape, yshape, dtype, mode, jsp_op, osp_op):
     rng = jtu.rand_default(self.rng())
+    if xshape != yshape:
+      return
```
jakevdp, line 58: This seems strange to only test cases where the two arrays have equal shape. Maybe instead in the parametrization we can do something like:
```python
for shapeset in [onedim_shapes, twodim_shapes, threedim_shapes]
for xshape in shapeset
for yshape in shapeset
```
Then tests will cover mixed pairs of shapes with the same dimension.

ayush-1506: Makes sense, will do!

jakevdp (approved):
Looks great, thanks!

PR #4824, opened 2020-11-07
add gpu determinism note

mattjj:


gnecula:
Thanks!

docs/gpu_determinism.rst
```diff
@@ -0,0 +1,15 @@
+Determinism on GPUs
+======================
+
+It can be tricky to make floating point computations on GPUs fully
+deterministic, in the sense of computing precisely the same values
+across separate program executions on the same hardware, and in general doing
+so may require sacrificing some performance. Some sources of nondeterminism in
+XLA:GPU are:
```
hawkinsp, line 8: I'm unsure whether this is actually true of the readthedocs rst renderer, but github's syntax highlighting is parsing `:GPU are:` as a rst directive. Maybe just write `XLA/GPU`?

docs/gpu_determinism.rst
```diff
@@ -0,0 +1,15 @@
...
+
+1. `XLA:GPU autotuning <https://docs.nvidia.com/deeplearning/frameworks/tensorflow-user-guide/index.html#xla-autotune>`_ automatically selects which cuBLAS and cuDNN kernels to call for best performance on your device, but exactly which gets picked for any given program execution may depend on exogenous factors;
+2. `GPU atomics can cause reductions to be nondeterministic <https://github.com/tensorflow/tensorflow/commit/e31955d9fb34ae7273354dc2347ba99eea8c5280>`_.
+
```
hawkinsp, line 12: The third source  that we know of is `scatter` operations. I would mention it here.

docs/gpu_determinism.rst
```diff
@@ -0,0 +1,15 @@
...
+One can remove some sources of nondeterminism by setting the environment
```
hawkinsp, line 13: I think we should (not in this PR) add a `jax.config.deterministic()` mode that sets these two options and addition warns on non-deterministic scatters.

hawkinsp (approved):

docs/gpu_determinism.rst
agarwl: +1 to providing a `jax.config.deterministic()` mode with the caveat for scatter updates.

mattjj:
It sounds like the `--xla_gpu_deterministic_reductions` flag is now gone (or it will be when we push an updated jaxlib) because it's now effectively always on by default. So hopefully this will get less surprising...

jakevdp:
Closing as obsolete.

PR #4826, opened 2020-11-08
Pass around the flattened args and outputs in invertible_ad

apaszke:
The previous version worked fine when the argument list was already
flat, but would fail when it was a more complicated pytree. This should
fix it.


mattjj (approved):
LGTM, but WDYT about adding a test that would've failed before and passes now?

apaszke:
Yeah, good call. Added the test.

PR #4832, opened 2020-11-08
Gmres QR (rebased)

shoyer:
from https://github.com/google/jax/pull/4175

### DIFF
--- a/tests/lax_scipy_sparse_test.py
+++ b/tests/lax_scipy_sparse_test.py
@@ -28,6 +28,7 @@ import jax.scipy.sparse.linalg
 
 from jax.config import config
 config.parse_flags_with_absl()
+config.update('jax_enable_x64', True)
 
 float_types = [np.float32, np.float64]
 complex_types = [np.complex64, np.complex128]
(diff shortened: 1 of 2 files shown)


shoyer:
@alewis @gehring could you kindly consent to this PR again?

(I wanted to consolidate commits a little bit to de-clutter Git history)

gehring:
@googlebot I consent.

tests/lax_scipy_sparse_test.py
```diff
@@ -178,6 +209,201 @@ def tree_unflatten(cls, aux_data, children):
...
+            preconditioner,
+            qr_mode),
+       "shape": shape, "dtype": dtype, "preconditioner": preconditioner,
+       "qr_mode": qr_mode}
+      for shape in [(3, 3)]
+      # TODO(shoyer): get working for np.complex128 and qr_mode=True
+      for dtype in [np.float64]
+      for preconditioner in [None, 'identity', 'exact']
+      for qr_mode in [False]))
+  def test_gmres_against_scipy(self, shape, dtype, preconditioner, qr_mode):
```
shoyer, line 251: This new test (verifying consistency with scipy) currently only passes for float64 (not complex128) and only for qr_mode=False. I am concerned that there is bug in the GMRES implementation for these other cases.

We should still merge this PR to facilitate iteration, but my suggestion is that we keep `gmres` out of the main `jax.scipy.sparse` namespace until we fix this.

alewis:
@googlebot I consent

tests/lax_scipy_sparse_test.py
alewis: It's not clear to me that the code necessarily should agree with SciPy for the unconverged solution. Why is this?

shoyer: Assuming exact precision arithmetic, shouldn't GMRES be deterministic algorithm on non-pathological inputs?

These tests are running on small matrices with double precision, so loss of numerical precision cannot explain the difference.

alewis: I'm honestly not totally sure. Different choices of orthogonalization procedure for the Arnoldi process could perhaps influence the basis found. But perhaps you are right.

I suppose an implication of your results is that the qr_mode choices don't agree with one another. So that's something to understand as well.

shoyer: GMRES could still be useful even if restricted to `qr_mode=False` and `dtype=float64`. Perhaps that's a better intermediate state of affairs, but I'm going to try to merge this now so it's easier to iterate.

I suspect there may be two lingering bugs:

- Handling of complex numbers. Perhaps we're missing a complex conjugate somewhere, e.g., in the orthogonalization step?
- Something in QR mode.

shoyer: Indeed, there were a couple of lingering bugs! See https://github.com/google/jax/pull/5034 for the fixes.

PR #4835, opened 2020-11-09
simplify the implementation of vmap collectives

mattjj:
These changes came from experimenting with a `pdot` primitive.

The main intended changes here are:
1. remove the need for split axis rules in batching.py, and instead rely on just the collective rules (namely to handle vectorizing their operation over a single named axis even if the collective is applied over multiple named axes)
2. simplify `BatchTrace.process_primitive` so that we don't pass tracers into rules and rely on a somewhat subtle recursion

Another minor change is we factor some tests to be more granular, so as to make it easier to select individual cases from the command line.

---

Before this PR, we had two separate tables in batching.py which held rules for handling collectives:
1. `collective_rules` which only needed to handle the case of operating over single named axes (rather than tuples of named axes), e.g. `psum(x, 'i')` but not `psum(x, ('i', 'j')`, and
2. `split_axis_rules` which would split a collective operating over more than one axis into a composition of primitives operating on single axes, e.g. `psum(x, ('i', 'j'))` would be split into `psum(psum(x, 'i'), 'j')`.

These two rules were used together to support collectives with `vmap`, where the latter would decompose things so that they could be handled by the former.

However, there was a complexity cost: the logic in `BatchTrace.process_primitive` involved a subtle recursion, and we had two tables of rules instead of one.

After this PR, we just have a table of `collective_rules`, where these rules do need to handle the case of collectives that operate over multiple axis names. This rule setup is more general, in that one can always implement this kind of a rule by composing the functions that constituted the previous set of rules, but we aren't forced to factor the code that way. Moreover, we avoid the subtle recursion in `BatchTrace.process_primitive`.

---

Follow-up work needed:
* This change breaks multi-axis all_to_all. We can either implement the previous logic in terms of the new rule organization (not included in this PR just because it's very tricky logic and supporting it is not a blocker), or we can avoid the need for multi-axis all_to_all by adjusting how gmap/xmap lowering works (so we don't need to do any axis splitting for all_to_all, thus avoiding the trickiness).
* Make `pmin_p` and `pmax_p` be multi-input/multi-output to bring them in line with `psum_p` so that we can share more helper code among these three monoid reducer collectives (#5011).
* Rename the `axis_name` parameter used by collectives to be `axis_names`, and correspondingly ensure at the "traceable" level (e.g. in the `psum` function that calls `psum_p.bind`) that it's always a tuple, rather than a tuple-or-singleton as it is now. This way we handle the polymorphism as early as possible, so the growing amount of downstream code that processes collectives doesn't need to handle both cases.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -380,6 +380,9 @@ def _batched_reduction_collective2(
     prim, if_mapped, if_unmapped, frame, vals_in, dims_in, axis_name,
     axis_index_groups):
   assert not prim.multiple_results  # cf. _batched_reduction_collective
+  if axis_index_groups is not None:
+    raise NotImplementedError("axis_index_groups not supported in vmap collectives. "
+                              "Please open a feature request!")
   (v,), (d,) = vals_in, dims_in
   val_out = (if_mapped(v, d) if d is not batching.not_mapped
              else if_unmapped(v, frame.size))
@@ -538,6 +541,7 @@ def _ppermute_transpose_rule(t, perm, axis_name):
 
 def _ppermute_batcher(frame, vals_in, dims_in, axis_name, perm):
   assert len(perm) == frame.size, "Permutation doesn't match the axis size!"
+  assert axis_name == frame.name, "ppermute batcher called with wrong axis name"
   (v,), (d,) = vals_in, dims_in
   assert d is not batching.not_mapped
   perm_indices = [None] * frame.size
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -74,6 +74,8 @@ def psum(x, axis_name, *, axis_index_groups=None):
   >>> print(y)
   [ 0.          0.16666667  0.33333334  0.5       ]
   """
+  if not isinstance(axis_name, (tuple, list)):
+    axis_name = (axis_name,)
   _validate_axis_index_groups(axis_index_groups)
   leaves, treedef = tree_util.tree_flatten(x)
   leaves = [lax.convert_element_type(l, np.int32)
@@ -134,6 +136,8 @@ def pmax(x, axis_name, *, axis_index_groups=None):
     Array(s) with the same shape as ``x`` representing the result of an
     all-reduce max along the axis ``axis_name``.
   """
+  if not isinstance(axis_name, (tuple, list)):
+    axis_name = (axis_name,)
   _validate_axis_index_groups(axis_index_groups)
   return tree_util.tree_map(partial(
       pmax_p.bind, axis_name=axis_name, axis_index_groups=axis_index_groups), x)
@@ -157,6 +161,8 @@ def pmin(x, axis_name, *, axis_index_groups=None):
     Array(s) with the same shape as ``x`` representing the result of an
     all-reduce min along the axis ``axis_name``.
   """
+  if not isinstance(axis_name, (tuple, list)):
+    axis_name = (axis_name,)
   _validate_axis_index_groups(axis_index_groups)
   return tree_util.tree_map(partial(
       pmin_p.bind, axis_name=axis_name, axis_index_groups=axis_index_groups), x)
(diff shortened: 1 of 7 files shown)


jax/_src/lax/parallel.py
```diff
@@ -157,6 +161,8 @@ def pmin(x, axis_name, *, axis_index_groups=None):
     Array(s) with the same shape as ``x`` representing the result of an
     all-reduce min along the axis ``axis_name``.
   """
+  if not isinstance(axis_name, (tuple, list)):
+    axis_name = (axis_name,)
```
apaszke, line 165: BTW one interesting issue with treating all tuples as collections of axes is that I once tried using a `taggedtuple` as a wrapper that would wrap an axis name and make it distinct form other axis names, but it later got unpacked in this code, because it satisfied `isinstance(x, tuple)`. Not necessary a comment to this PR, but a slightly weird behavior that might be surprising at times.

jax/_src/lax/parallel.py
```diff
@@ -347,42 +353,41 @@ def _allreduce_translation_rule(prim, c, val, *, axis_name, axis_index_groups,
...
+  if len(axis_name) > 1:
+    remaining_axis_names = tuple(n for n in axis_name if n != frame.name)
+    vals_out = prim.bind(*vals_out, axis_name=remaining_axis_names,
+                         axis_index_groups=None)
+  return vals_out, [batching.not_mapped] * len(vals_out)
+
+# TODO(mattjj): update pmin/pmax to have multiple_results like psum, delete this
+def _batched_reduction_collective2(
+    prim, if_mapped, if_unmapped, frame, vals_in, dims_in, axis_name,
+    axis_index_groups):
```
apaszke, line 381: This is missing a check that `axis_index_groups` are `None`!

apaszke: Also, couldn't we implement this in terms of a call to some list wrapping + `_batched_reduction_collective` + list unwrapping? This would make sure that it's easier to keep the impls in sync and would avoid errors like this one.

jax/_src/lax/parallel.py
```diff
@@ -534,14 +536,14 @@ def _ppermute_transpose_rule(t, perm, axis_name):
...
 
-def _ppermute_batcher(vals_in, dims_in, axis_size, axis_name, perm):
-  assert len(perm) == axis_size, "Permutation doesn't match the axis size!"
-  perm_indices = np.full((axis_size,), -1, dtype=np.int32)
-  for s, d in perm:
-    perm_indices[s] = d
-  vals_out = [lax_numpy.take(v, perm_indices, d) if d is not batching.not_mapped else v
-              for v, d in zip(vals_in, dims_in)]
-  return vals_out, dims_in
+def _ppermute_batcher(frame, vals_in, dims_in, axis_name, perm):
```
apaszke, line 539: This should check that `axis_name == frame.name` or otherwise it will silently compute something incorrect.

apaszke (approved):
LGTM once the few missing checks are added!

jax/_src/lax/parallel.py
mattjj: Good points! Though #5011 will remove this function anyway so I think we don't need to refactor much on this PR. I can add the `None` check though.

jax/_src/lax/parallel.py
mattjj: Hrm interesting. Indeed having tuples represent collections of axis names is inconsistent with the description that anything hashable can be an axis name.

We could just make lists represent collections of axis names. But I think using lists and tuples interchangeably is so common in Python APIs that having them behave differently here would be confusing. So it seems okay to have tuples (and all subclasses) be an exception so that they act like collections rather than singletons.

Are there other options for how we could set things up?

jax/_src/lax/parallel.py
mattjj: I think this should only be called when `axis_name == frame.name`, but I'll add an assertion. LMK if that sounds incorrect!

mattjj: I guess it's analogous to how `_batched_reduction_collective` includes `assert frame.name in axis_name`.

mattjj:
Thanks for the eagle-eyed review!

PR #4848, opened 2020-11-09
Add support for pmap in_axes other than 0 and None 

apaszke:
... and in map primitives in general (which is why the patch touches
most traces).

This also fixes a bug in the transpose rule for map primitives, which
would fail to adjust the aval associated with zeros returned from the
map body.


mattjj (approved):
LGTM, nice! Consider running a global presubmit check before merging (to reduce the chance we have to roll back in case some internal test covers a weird case our tests don't).

PR #4850, opened 2020-11-10
Propagate weak types through unary and binary ops

jakevdp:
This PR takes the first step toward propagating weak types through JAX operations, adding machinery to abstract evaluation rules to control how they handle weak types, as well as adding a `weak_type` argument to `convert_element_type` that defaults to `False`, which is the current behavior.

Two new tests ensure that weak types are correctly propagated through basic unary and binary operations.

### DIFF
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2312,5 +2312,29 @@ class LazyConstantTest(jtu.JaxTestCase):
                                 "index_dtype must be an integer type"):
       jax_fn(np.ones((2, 2)), axis=0, index_dtype=index_dtype)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+        {"testcase_name": "_{}".format(rec.op),
+         "op_name": rec.op, "rec_dtypes": rec.dtypes}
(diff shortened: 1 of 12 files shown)


tests/lax_test.py
```diff
@@ -2312,5 +2312,34 @@ def testArgMinMaxIndexDtypeError(self, jax_fn, index_dtype):
                                 "index_dtype must be an integer type"):
       jax_fn(np.ones((2, 2)), axis=0, index_dtype=index_dtype)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+        {"testcase_name": "_{}".format(rec.op),
+         "op_name": rec.op, "rec_dtypes": rec.dtypes}
+      for rec in LAX_OPS if rec.nargs == 1))
+  def testUnaryWeakTypes(self, op_name, rec_dtypes):
+    """Test that all lax unary ops propagate weak_type information appropriately."""
+    if op_name in ['acos', 'asin', 'atan', 'tan']:
```
jakevdp, line 2321: I believe these exceptions can be removed once #5130 is merged.

mattjj (approved):

PR #4854, opened 2020-11-10
Rm redundant pre-omnistaging attribute annotations from core.TraceStack

j-towns:


mattjj:
Thanks! I suspect this was a result of a bad rebase.

I think we want to add back correct annotations though, which should probably be

```python
class TraceStack:
  stack: List[MainTrace]
  dynamic: MainTrace
```

I'll see if I can push those onto this PR.

mattjj (approved):
LGTM!

PR #4855, opened 2020-11-10
Fix batch polymorphism for take and gather

gnecula:


jax/_src/lax/lax.py
```diff
@@ -1413,8 +1413,8 @@ def iota(dtype: DType, size: int) -> Array:
   """
   if config.omnistaging_enabled:
     dtype = dtypes.canonicalize_dtype(dtype)
-    size = core.concrete_or_error(int, size, "size argument of lax.iota")
```
gnecula, line 1416: I am thinking that here we can use the same code to handle input shapes for primitives. Incidentally, the previous code fails for Poly.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3780,9 +3780,6 @@ def take(a, indices, axis=None, out=None, mode=None):
 
 def _normalize_index(index, axis_size):
   """Normalizes an index value in the range [-N, N) to the range [0, N)."""
-  if type(axis_size) is Poly:
```
gnecula, line 3783: This code does not work if `index` is a tracer, and it seems like a partial attempt to make this work for polymorphism. Nothing breaks if I remove this. I am curious if something breaks in Flax.

jax/_src/numpy/lax_numpy.py
```diff
@@ -4102,7 +4099,8 @@ def _index_to_gather(x_shape, idx):
 
       # Broadcast gather_indices from [..., k] to [..., 1, 1, ..., 1, k].
       gather_indices = lax.broadcast_in_dim(
-        gather_indices, np.insert(gather_indices.shape, -1, shape),
```
gnecula, line 4105: `np.insert` does not work if `shape` has Poly.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2669,9 +2669,9 @@ def arange(start, stop=None, step=None, dtype=None):
   require = partial(core.concrete_or_error, _np_asarray)
   msg = "It arose in jax.numpy.arange argument `{}`.".format
   if stop is None and step is None:
-    start = require(start, msg("stop"))
+    start = require(start, msg("start"))
     dtype = dtype or _dtype(start)
-    return lax.iota(dtype, np.ceil(start)) # avoids materializing
+    return lax.iota(dtype, int(np.ceil(start))) # avoids materializing
```
bchetioui, line 2674: The error string change seems wrong to me. The only mandatory argument should – while being called "start" – correspond to `stop` here; maybe we can do:
```python
stop = require(start, msg("stop"))
dtype = dtype or _dtype(stop)
return lax.iota(dtype, int(np.ceil(stop)))
```
to clarify instead? The docstring inherited from np.arange does say: `arange([start,] stop[, step,], dtype=None)`, so this seems inconsistent.

If you maintain this change, it seems you also need to change `tests/testArangeConcretizationError` in `lax_numpy_test.py`

gnecula: These are good points. I have changed back the error string, and renamed `start` to `stop` as you suggested.

PR #4868, opened 2020-11-11
Add support for XLA variadic reduce

sharadmv:
Enable Pytree support for `lax.reduce`(i.e. expose XLA's variadic reduce)

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -28,9 +28,11 @@ import jax
 from jax import core
 from jax import ad_util
 from jax import api
+from jax import api_util
 from jax import linear_util as lu
 from jax import dtypes
 from jax import lazy
+from jax import tree_util
 from jax.config import flags, config
 from jax.core import Primitive, _canonicalize_dimension
 from jax.abstract_arrays import (UnshapedArray, ShapedArray, ConcreteArray, array_types,
@@ -42,7 +44,8 @@ from jax.interpreters import ad
 from jax.interpreters import invertible_ad as iad
 from jax.interpreters import batching
 from jax.interpreters import masking
-from jax.util import cache, safe_zip, partial, prod, safe_map, canonicalize_axis
+from jax.util import (cache, safe_zip, partial, prod, safe_map, canonicalize_axis,
+                      split_list)
 from jax.tree_util import tree_map
 from jax.lib import pytree
 from jax.lib import xla_bridge
@@ -1088,19 +1091,30 @@ def argmax(operand: Array, axis: int,
   return argmax_p.bind(operand, axes=(axis,),
                        index_dtype=dtypes.canonicalize_dtype(index_dtype))
 
-def reduce(operand: Array, init_value: Array, computation: Callable,
+def reduce(operands: Array, init_values: Array, computation: Callable,
            dimensions: Sequence[int]) -> Array:
   """Wraps XLA's `Reduce
   <https://www.tensorflow.org/xla/operation_semantics#reduce>`_
   operator.
   """
-  monoid_reducer = _get_monoid_reducer(computation, init_value)
+  flat_operands, operand_tree = tree_util.tree_flatten(operands)
+  flat_init_values, init_value_tree = tree_util.tree_flatten(init_values)
+  if operand_tree != init_value_tree:
+    raise ValueError('Operands must have the same tree structure as init_values:'
+                     f' {operand_tree} vs. {init_value_tree}')
+  if len(flat_operands) != len(flat_init_values):
+    raise ValueError('Must have same total number of operands as init_values: '
+                     f' {len(flat_operands)} vs. {len(flat_init_values)}')
+  monoid_reducer = _get_monoid_reducer(computation, flat_init_values)
   if monoid_reducer:
-    return monoid_reducer(operand, dimensions)
+    return monoid_reducer(*flat_operands, dimensions)
   else:
-    jaxpr, consts = _reduction_jaxpr(computation, _abstractify(init_value))
-    return reduce_p.bind(operand, init_value, computation=computation,
+    flat_init_avals = safe_map(_abstractify, flat_init_values)
+    jaxpr, consts, out_tree = _variadic_reduction_jaxpr(
+        computation, tuple(flat_init_avals), init_value_tree)
+    out = reduce_p.bind(*(flat_operands + flat_init_values), computation=computation,
                          jaxpr=jaxpr, consts=consts, dimensions=tuple(dimensions))
+    return tree_util.tree_unflatten(out_tree, out)
 
 @cache()
 def _reduction_jaxpr(computation, aval):
@@ -1109,12 +1123,27 @@ def _reduction_jaxpr(computation, aval):
   jaxpr, _, consts = pe.trace_to_jaxpr(comp, (pval, pval), instantiate=False)
   return jaxpr, consts
 
-def _get_monoid_reducer(monoid_op: Callable, x: Array) -> Optional[Callable]:
+@cache()
+def _variadic_reduction_jaxpr(computation, flat_avals, aval_tree):
+  avals = tree_util.tree_unflatten(aval_tree, flat_avals)
+  flat_in_avals, in_tree = tree_util.tree_flatten((avals, avals))
+  pvals = safe_map(pe.PartialVal.unknown, flat_in_avals)
+  comp = lu.wrap_init(computation)
+  flat_comp, out_tree = api_util.flatten_fun_nokwargs(comp, in_tree)
+  jaxpr, _, consts = pe.trace_to_jaxpr(flat_comp, tuple(pvals),
+                                       instantiate=False)
+  return jaxpr, consts, out_tree()
+
+def _get_monoid_reducer(monoid_op: Callable, xs: Array) -> Optional[Callable]:
+  if len(xs) != 1:
+    return None
+  x, = xs
   aval = core.get_aval(x)
   dtype = _dtype(x)
   if (type(aval) is ConcreteArray) and aval.shape == ():
     if monoid_op is add:
-      return np.equal(aval.val, 0) and _reduce_sum
+      return np.equal(aval.val, 0) and partial(
+          _reduce_sum)
     if monoid_op is mul:
       return np.equal(aval.val, 1) and _reduce_prod
     elif monoid_op is bitwise_or and dtype == np.bool_:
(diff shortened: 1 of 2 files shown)


jakevdp:
Have you considered instead generalizing the existing `reduce_p` primitive to accept a sequence of operands? Since XLA does variadic & non-variadic reduce via a single operation, it might simplify things.

sharadmv:
It's an option, but it requires changing `reduce_p` to have `multiple_results=True`. I wasn't sure whether that's an option or not.

sharadmv:
I've updated the PR with two features.

1. Made `reduce` handle Pytrees
2. Add a `use_kahan_summation` kwarg to `jnp.sum` that's threaded into `lax._reduce_sum` that modifies the translation rule.

jax/_src/lax/lax.py
```diff
@@ -1088,19 +1091,29 @@ def argmax(operand: Array, axis: int,
...
+def reduce(operands: Array, init_values: Array, computation: Callable,
+           dimensions: Sequence[int], **reduce_kwargs: Any) -> Array:
   """Wraps XLA's `Reduce
   <https://www.tensorflow.org/xla/operation_semantics#reduce>`_
   operator.
   """
-  monoid_reducer = _get_monoid_reducer(computation, init_value)
+  flat_operands, operand_tree = tree_util.tree_flatten(operands)
+  flat_init_values, init_value_tree = tree_util.tree_flatten(init_values)
+  assert operand_tree == init_value_tree, (
```
hawkinsp, line 1102: I'd probably raise a TypeError or ValueError explicitly rather than using assert. I'd also include details of the structure to help the user. Same everywhere.

jax/_src/lax/lax.py
```diff
@@ -1088,19 +1091,29 @@ def argmax(operand: Array, axis: int,
...
+  assert len(flat_operands) == len(flat_init_values), (
+      'Must have same number of init values as operands')
+  monoid_reducer = _get_monoid_reducer(computation, flat_init_values, **reduce_kwargs)
   if monoid_reducer:
-    return monoid_reducer(operand, dimensions)
+    return monoid_reducer(*flat_operands, dimensions)
   else:
-    jaxpr, consts = _reduction_jaxpr(computation, _abstractify(init_value))
-    return reduce_p.bind(operand, init_value, computation=computation,
+    init_avals = tree_util.tree_map(_abstractify, init_values)
```
hawkinsp, line 1110: Nit: I might be tempted to flatten and then abstractify.

jax/_src/lax/lax.py
```diff
@@ -1109,12 +1122,27 @@ def _reduction_jaxpr(computation, aval):
...
+  flat_in_avals, in_tree = tree_util.tree_flatten((avals, avals))
+  pvals = safe_map(pe.PartialVal.unknown, flat_in_avals)
+  comp = lu.wrap_init(computation)
+  flat_comp, out_tree = api_util.flatten_fun_nokwargs(comp, in_tree)
+  jaxpr, _, consts = pe.trace_to_jaxpr(flat_comp, tuple(pvals),
+                                       instantiate=False)
+  return jaxpr, consts, out_tree()
+
+def _get_monoid_reducer(monoid_op: Callable, xs: Array, **reduce_kwargs) -> Optional[Callable]:
+  if len(xs) > 1:
```
hawkinsp, line 1137: Presumably you should test for exact equality. What happens if I attempt a nullary reduction?

(That may be illegal at the XLA level (we should check), but there's clearly nothing stopping me from getting this far with one.)

jax/_src/lax/lax.py
```diff
@@ -4681,53 +4711,92 @@ def _scatter_jvp(primals, tangents, *, update_jaxpr, update_consts,
   partial(_scatter_batching_rule, scatter))
 
 
-def _reduce_shape_rule(operand, init_value, *, computation, jaxpr, consts,
-                       dimensions):
-  return tuple(np.delete(operand.shape, dimensions))
+def _reduce_abstract_eval_rule(*avals, **params):
```
hawkinsp, line 4714: Do we have to use `**params` here or can we revert to using keyword arguments?

hawkinsp, line 4714: This rule should probably verify that the `init_values` are scalars.

jax/_src/lax/lax.py
```diff
@@ -4681,53 +4711,92 @@ def _scatter_jvp(primals, tangents, *, update_jaxpr, update_consts,
...
+  assert all(isinstance(aval, UnshapedArray) for aval in avals), avals
+  avals = tuple(raise_to_shaped(aval) for aval in avals)
+  operand_avals, init_value_avals = split_list(avals, [len(avals) // 2])
+  least_specialized = _max(
```
hawkinsp, line 4718: It feels a bit excessive to need to write out this logic. Is there a reason we couldn't keep using `standard_abstract_eval` here?

jax/_src/lax/lax.py
```diff
@@ -4736,17 +4805,43 @@ def _reduce_number_dtype_rule(name, operand, *args, **kw):
...
-  return xops.Reduce(c, [operand], [xb.constant(c, np.array(0, dtype))],
-                     xla.primitive_subcomputation(add_p, scalar, scalar),
-                     axes)
+  zero = xb.constant(c, np.array(0, dtype))
+
+  if not use_kahan_summation:
+    return xops.Reduce(c, [operand], [xb.constant(c, np.array(0, dtype))],
+                       xla.primitive_subcomputation(add_p, scalar, scalar),
+                       axes)
+
```
hawkinsp, line 4821: Another trick you may prefer, but it's up to you (this is fine also):

Use `xla.lower_fun` and call your new `lax.reduce` inside the translation rule.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1865,8 +1866,12 @@ def _reduction_init_val(a, init_val):
 
 _cast_to_bool = partial(lax.convert_element_type, new_dtype=bool_)
 
-sum = _make_reduction("sum", np.sum, lax.add, 0, upcast_f16_for_computation=True,
-                      bool_op=lax.bitwise_or)
+def sum(a, axis=None, dtype=None, out=None, keepdims=False, use_kahan_summation=False):
```
hawkinsp, line 1869: I might err on the side of *not* adding this feature to `jax.numpy.sum` for now. We want to be somewhat cautious about extensions to the numpy api.

sharadmv: I removed it from numpy, but exposed a `lax.reduce_sum` with the `use_kahan_summation` kwarg, if that's a better alternative.

jax/_src/lax/lax.py
sharadmv: I'm okay with the implementation as an XLA subcomputation; I think I'll leave as-is.

jax/_src/lax/lax.py
sharadmv: `standard_abstract_eval` currently only handles primitives that don't have `multiple_results=True` (it only returns a single aval).

jax/_src/lax/lax.py
sharadmv: Done.

sharadmv: Done.

jax/_src/lax/lax.py
sharadmv: Done.

jax/_src/lax/lax.py
sharadmv: Done.

jax/_src/lax/lax.py
sharadmv: Done.

jax/_src/lax/lax.py
hawkinsp: Perhaps it's worth doing that generalization and making `standard_abstract_eval` support such primitives? I think it's simpler to have separate shape and dtype rules where possible.

jax/_src/lax/lax.py
```diff
@@ -4681,53 +4710,97 @@ def _scatter_jvp(primals, tangents, *, update_jaxpr, update_consts,
...
+  xla_computation = _reduction_computation(c, jaxpr, consts, init_values, singleton=False)
+  return xops.Reduce(c, operands, init_values, xla_computation, dimensions)
+
+def _reduce_batch_rule(batched_args, batch_dims, *, computation, jaxpr,
+                       consts, dimensions):
+  num_operands = len(batched_args) // 2
+  operands, init_values = split_list(batched_args, [num_operands])
+  operand_bdims, init_value_bdims = split_list(batch_dims, [num_operands])
+  if all(init_value_bdim is None for init_value_bdim in init_value_bdims):
+    assert all(operand_bdim is not None for operand_bdim in operand_bdims)
```
hawkinsp, line 4753: It's not clear to me that these two assertions always hold. I think you actually need to handle all of these cases.

sharadmv: I added a TODO.

hawkinsp (changes requested):
The change looks good. Could you please squash the commits?

sharadmv:
Done!

hawkinsp (approved):

jax/_src/lax/lax.py
```diff
@@ -1952,13 +1983,25 @@ def standard_abstract_eval(prim, shape_rule, dtype_rule, *args, **kwargs):
...
+    shapes, dtypes = shape_rule(*args, **kwargs), dtype_rule(*args, **kwargs)
+    if not prim.multiple_results:
+      shapes, dtypes = [shapes], [dtypes]
+    out_avals = safe_map(ShapedArray, shapes, dtypes)
   elif least_specialized is UnshapedArray:
-    return UnshapedArray(dtype_rule(*args, **kwargs))
+    dtypes = dtype_rule(*args, **kwargs)
+    if not prim.multiple_results:
+      dtypes = [dtypes]
+      out_avals = safe_map(UnshapedArray, dtypes)
```
gnecula, line 1999: I think that the indentation is off here.

tests/lax_test.py
```diff
@@ -2238,6 +2238,30 @@ def test_reduce_window_scalar_init_value_shape_rule(self):
     with self.assertRaisesRegex(TypeError, msg):
       lax.reduce_window(**args)
 
+  def test_reduce_correctly_works_with_pytrees(self):
```
gnecula, line 2241: Shouldn't there be some test with multiple operands?

gnecula:
@bchetioui can you please see what is needed to handle this in jax2tf?

jax/_src/lax/lax.py
sharadmv: Thanks for the catch!

tests/lax_test.py
sharadmv: Done.

sharadmv: Test has been updated to use multiple operands.

PR #4871, opened 2020-11-11
[jax2tf] Added testing and fixed conversion of conj.

bchetioui:


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -822,6 +822,25 @@ def _make_harness(name, *, shape=(4, 4), dtype=np.float32, symmetric=False,
...
+                 **kwargs)
+
+lax_conj = tuple( # Validate dtypes
+  _make_conj_harness("dtypes", dtype=dtype)
+  for dtype in jtu.dtypes.floating + jtu.dtypes.complex
+) + tuple( # Validate kwargs
+  _make_conj_harness("kwargs", **kwargs)
+  for kwargs in [
+    { "_input_dtype": np.float32 },             # expected kwarg for ad
+    { "random_kwarg1": 1, "random_kwarg2": 2 }, # irrelevant kwargs
```
gnecula, line 840: Is this important to test? I saw your conversation with Peter about the `**kwargs` in the implementation of `conj_p`, but this does not mean that the primitive would arise with random kwargs. Have you thought of instead replacing `**kwargs` with  an optional `input_dtype`?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -876,7 +876,18 @@ def _population_count(x):
 tf_impl[lax.bessel_i1e_p] = tf.math.bessel_i1e
 
 tf_impl[lax.complex_p] = tf.complex
-tf_impl[lax.conj_p] = tf.math.conj
+
+def _conj(x, **kwargs):
+  # The only dtypes that are allowed are: float32, float64, complex64, and
+  # complex128.
+  if x.dtype == tf.float32:
```
gnecula, line 883: was the problem here that `tf.math.conj` returns a float32 if the input is a float32, while HLO returns a complex64?

jax/experimental/jax2tf/tests/primitive_harness.py
bchetioui: Maybe it is not very important. I only made it this way for the sake of consistency with how the primitive is implemented, i.e. the implementation in lax & this one will behave the same regardless of kwargs. I thought about modifying both this impl and the one in lax, but was reluctant to change the prototype only in jax2tf. WDYT?

jax/experimental/jax2tf/jax2tf.py
bchetioui: Yes!

jax/experimental/jax2tf/tests/primitive_harness.py
gnecula: I think that we should remove this harness. This would be the only harness where we test passing random kwargs to primitives.

PR #4872, opened 2020-11-11
[jax2tf] Replaced cumsum/cumprod implementation and added testing.

bchetioui:
By adding testing, I noticed that `cumsum` and `cumprod` do not support many dtypes when converted using pure TF ops (`int`s, `uint`s, `complex`s, though this is a problem only in `compiled` mode). With this implementation, `reduce_window` is used instead, and all dtypes supported by add/mul are supported. The drawback is that these conversions now require XLA, but they can be specialized later on to use TF for the compatible dtypes if need be.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1400,9 +1400,6 @@ tf_impl[lax.argmax_p] = functools.partial(_argminmax, tf.math.argmax)
 _add_fn = tf.function(tf.math.add, autograph=False)
 _ge_fn = tf.function(tf.math.greater_equal, autograph=False)
 
-tf_impl[lax_control_flow.cumsum_p] = tf.math.cumsum
-tf_impl[lax_control_flow.cumprod_p] = tf.math.cumprod
-
 def _select_and_gather_add(tangents: TfVal,
                            operand: TfVal,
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1657,16 +1654,24 @@ def _get_min_identity(tf_dtype):
...
+# cummin, cumsum and cumprod. This is efficient on TPU, but the complexity is
+# O(n^2) on other backends. This may be implemented using associative_scan
+# instead to favor different backends.
 tf_impl_with_avals[lax_control_flow.cummin_p] = _convert_jax_impl(
     functools.partial(lax_control_flow._cumred_tpu_translation_rule,
                       lax._reduce_window_min), multiple_results=False)
 tf_impl_with_avals[lax_control_flow.cummax_p] = _convert_jax_impl(
     functools.partial(lax_control_flow._cumred_tpu_translation_rule,
                       lax._reduce_window_max), multiple_results=False)
+# TODO(bchetioui): cumsum and cumprod can be converted using pure TF ops for
```
gnecula, line 1667: Since this is fresh in your mind now, can you add a bit more detail here, e.g., which dtypes can be implemented?

gnecula (approved):

PR #4874, opened 2020-11-11
[jax2tf] Add a harness for broadcast_p and fix conversion.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1316,7 +1316,8 @@ def _dot_general(lhs, rhs, dimension_numbers, precision):
 
 
 def _broadcast(operand, *, sizes):
-  return tf.broadcast_to(operand, sizes + tf.shape(operand))
+  result_shape = tf.TensorShape(sizes).concatenate(operand.shape)
```
gnecula, line 1319: I am pretty sure we will need to `eval_shape(sizes)` here to create the TF shape, for shape polymrphism.

gnecula: I am curious, what was the error here?

bchetioui: There are actually several errors here:
(1) `+` is not concatenation, but pointwise addition
(2) in non-eager mode, `tf.shape(operand)` and `operand.shape` behave differently (see https://www.tensorflow.org/api_docs/python/tf/shape)

It was actually a bit tricky!

gnecula: Good catch

PR #4875, opened 2020-11-11
[jax2tf] Fix implementation and add tests for broadcast_in_dim.

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1321,8 +1321,9 @@ def _broadcast(operand, *, sizes):
 
 
 def _broadcast_in_dim(operand, *, shape, broadcast_dimensions):
-  inshape = tuple(1 if i not in broadcast_dimensions else d
-                  for i, d in enumerate(shape))
+  inshape = [1] * len(shape)
```
gnecula, line 1324: Good catch!

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1321,8 +1321,9 @@ def _broadcast(operand, *, sizes):
...
+  for orig_shape_i, broadcast_dim_i in zip(operand.shape, broadcast_dimensions):
```
gnecula, line 1325: I am having difficulty understanding why is this correct. The conditional looks suspicious, since there is nothing conditional in the documented semantics: "the i'th dimension of the operand is mapped to the broadcast_dimension[i]'th dimension of the output shape"

It is not obvious to see from your code that `prod(inshape) == prod(operand.shape)`, since only then the `tf.reshape` is correct. 

Wouldn't the following code be more obviously following the semantics, and more obviously have the above reshape property?

```
inshape = [1] * len(shape)
for orig_shape_i, broadcast_dim_i in zip(operand.shape, broadcast_dimensions):
   inshape[broadcast_dim_i] = orig_shape_i
```

I think that your code is likely to be correct because of the side conditions on how the dimensions must relate: either `orig_shape_i == 1` or `orig_shape_i == shape[broadcast_dim_i]`. But I find the alternative I am proposing easier to compare to the semantics.

bchetioui: I agree with you. I think my code is correct because the shape rule would anyway throw out any code where `orig_shape_i != 1` or `orig_shape_i != shape[broadcast_dim_i]`, but this is more obvious.
I pushed a fix, thank you!

bchetioui: FTR, this change makes it difficult to work with shape polymorphism (because it requires carrying avals, and it is not straightforward to add them in `_while` that makes use of `broadcast_in_dim`). Since `shape` always correspond to `_in_avals[0]` it is simpler to keep the code as it was originally.

PR #4878, opened 2020-11-11
[masking] Ensure that Poly operations are sound

gnecula:
We want to ensure that an operation on polynomials throws
an explicit error when the result of the operation cannot
be soundly represented as a polynomial.

This change implements polynomials ranging over non-negative
shape variables. (It may be a good idea to restrict the range
to positive values only.)

Formally, an operation `op` on polynomial `p` either
throws an error `UndefinedPoly` or produces a polynomial
`res` such that `op(Val(p)) = Val(res)`, for any `Val`, a
value with non-negative values for shape variables.

Prior to this there were several sources of unsoundness:

 * The comparison `p1 = p2` would return True when the
   polynomials are identical (sound), and False when they
   are not identical (not sound)
 * The comparison `p >= 0` would always return True. This
   is unsound if `p` has monomial with negative coefficients.


tests/masking_test.py
```diff
@@ -83,8 +83,12 @@ def test_Poly_equal(self):
...
     assert constant_poly(3) != constant_poly(4)
     assert Poly({Mon(): 3, Mon({'n': 1}): 4}) == Poly({Mon({'n': 1}): 4, Mon(): 3})
-    assert Poly({Mon(): 3, Mon({'n': 1}): 4}) != Poly({Mon(): 3, Mon({'n': 2}): 4})
-    assert Poly({Mon(): 3, Mon({'m': 1}): 4}) != Poly({Mon(): 3, Mon({'n': 1}): 4})
+    assert Poly({Mon(): 3, Mon({'n': 1}): 4}) != Poly({Mon(): 4, Mon({'n': 1}): 4})
+    assert Poly({Mon(): 3, Mon({'n': 1}): 4}) != Poly({Mon(): 2})
+    with self.assertRaisesRegex(UndefinedPoly, "inconclusive"):
+      Poly({Mon(): 3, Mon({'n': 1}): 4}) != Poly({Mon(): 3, Mon({'n': 2}): 4})
+    with self.assertRaisesRegex(UndefinedPoly, "inconclusive"):
+      Poly({Mon(): 3, Mon({'m': 1}): 4}) != Poly({Mon(): 3, Mon({'n': 1}): 4})
```
bchetioui, line 91: I'm curious why this is inconclusive, and why in general `__eq__` is inconclusive in so many cases. My interpretation is that all the polynomials should be defined as parameterized by all the possible input variables, i.e. comparing these two polynomials `p(m) = 4*m + 3` and `q(n) = 4*n + 3` in this case is really a comparison of `p(m, n) = 4*m + 3` and `q(m, n) = 4*m + 3`, with two independent variables. In that case, it seems to me that we can easily conclude that these polynomials are different.

The other interpretation (that would make this more ambiguous) is if we considered `p` and `q` as polynomials with one input variable, such that they'd be equivalent up to renaming of the input variable; I don't think this is the setting we are in though?

gnecula: You make a good point about how to interpret the soundness requirement. I do think that it makes sense to treat each shape variable as distinct. The main goal of soundness is to ensure that JAX tracing with polynomial shapedarray is a conservative approximation of all the possible behaviors of tracing under constant shapearray. In any trace the shape variables are all introduced at the start of the trace, so they are global to the trace. Consider the following code:

```
def f(x):
  return 1 if x.shape[0] == x.shape[1] else return 1
``` 

If you trace this with `in_shapes=['n', 'm']`, you want to fail, because there is no single trace that abstract all the possible traces for 2D arrays. You may think that such user code is rare, but conditionals on shape values abound inside the JAX implementation of jax.numpy and the shape checking rules.

bchetioui: I think I understand: I feel that my confusion stems from my interpretation of how `__eq__` should behave, which is a purely boolean comparison of whether the polynomials are formally the same or not. But what we are actually interested in is (as exhibited in the code) whether, for two polynomials `p` and `q`
- `p == q` in all points OR
- `p != q` in all points (i.e. `p - q` has no real roots)
as these two conditions ensure we can always pick a single trace.

Thanks for the clarification!

mattjj (approved):
This is a great idea, @gnecula! Thanks for explaining it. This is a huge step forward in the design (though technically a step backward in the coverage, it's a worthwhile one)!

PR #4883, opened 2020-11-12
[jax2tf] Add testing for argmax and argmin.

bchetioui:


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -226,6 +226,47 @@ def _get_min_identity(dtype):
...
+)
+# TODO(bchetioui): the below documents a limitation of argmin and argmax when a
+# dimension of the input is too large. However, it is not categorizable as it
+# seems that the converter fails before reaching the actual primitive call. This
+# suggests that we may need to harden the converter to handle inputs this big.
+# + tuple( # Document limitation in case of too large axis
+#  _make_argminmax_harness("overflow_axis", prim=prim,
+#                          arr=np.ones((2**31,), dtype=np.uint8))
+#  for prim in [lax.argmin_p, lax.argmax_p]
+#)
```
bchetioui, line 268: @gnecula I don't know how relevant this is, but I suspect that such dimensions are not unheard of (?). In which case, I plan to investigate this soon.

gnecula: This may be nice to fix, but I think it is not a very high priority. I wonder though if handling large constants is a general problem, not just for argmax. In that case, I think it becomes a higher priority.

bchetioui: To be accurate, this seems to be a problem for any op right now; if any dimension of the shape of the operand can not fit in an `int32`, we encounter problems before primitive conversions. I believe this can be fixed by carefully adding upcasts where needed.

PR #4886, opened 2020-11-12
jaxlib: Fix Python 3.9 build and drop CUDA 10.0 support

skye:
The TF/XLA build no longer works with CUDA 10.0. This could potentially be fixed, but isn't easy or officially supported.

### DIFF
--- a/build/build_wheel_docker_entrypoint.sh
+++ b/build/build_wheel_docker_entrypoint.sh
@@ -62,7 +62,9 @@ esac
 
 export JAX_CUDA_VERSION=$3
 python setup.py bdist_wheel --python-tag "$PY_TAG" --plat-name "$PLAT_NAME"
-if python -m auditwheel show dist/jaxlib-*.whl  | grep -v 'platform tag: "manylinux2010_x86_64"' > /dev/null; then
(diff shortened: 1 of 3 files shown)


build/build_wheel_docker_entrypoint.sh
```diff
@@ -62,7 +62,9 @@ esac
 
 export JAX_CUDA_VERSION=$3
 python setup.py bdist_wheel --python-tag "$PY_TAG" --plat-name "$PLAT_NAME"
-if python -m auditwheel show dist/jaxlib-*.whl  | grep -v 'platform tag: "manylinux2010_x86_64"' > /dev/null; then
+if ! python -m auditwheel show dist/jaxlib-*.whl  | grep 'platform tag: "manylinux2010_x86_64"' > /dev/null; then
```
hawkinsp, line 65: Please verify the wheel you get out passes the test when done manually!
(I don't really parse the shell here.)

hawkinsp (approved):

build/build_wheel_docker_entrypoint.sh
skye: I used this to generate the just-released jaxlib 0.1.57 wheels, and the tests pass (both mac and linux).

PR #4893, opened 2020-11-13
More stable implementation of np.hypot

### LINKED ISSUE #4882: jnp.hypot is numerically unstable
`sqrt(x**2 + y**2)` is not a good implementation of hypot (e.g., see Wikipedia https://en.wikipedia.org/wiki/Hypot)

ashridh:
Fixes: #4882

This should ideally avoid overflows.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -986,7 +986,10 @@ def heaviside(x1, x2):
 def hypot(x1, x2):
   _check_arraylike("hypot", x1, x2)
   x1, x2 = _promote_dtypes_inexact(x1, x2)
-  return lax.sqrt(x1*x1 + x2*x2)
+  x1 = lax.abs(x1)
+  x2 = lax.abs(x2)
+  x1, x2 = maximum(x1, x2), minimum(x1, x2)
+  return lax.select(x1 == 0, x1, x1 * lax.sqrt(1 + lax.square(lax.div(x2, lax.select(x1 == 0, ones_like(x1), x1)))))
 
 
 @_wraps(np.reciprocal)


ayush-1506:
Ah, might have to make changes to get this to work for arrays.

jax/_src/numpy/lax_numpy.py
```diff
@@ -986,7 +986,12 @@ def heaviside(x1, x2):
 def hypot(x1, x2):
   _check_arraylike("hypot", x1, x2)
   x1, x2 = _promote_dtypes_inexact(x1, x2)
-  return lax.sqrt(x1*x1 + x2*x2)
+  x1 = lax.abs(x1)
+  x2 = lax.abs(x2)
+  if x2 > x1:
```
hawkinsp, line 991: You need to use `lax.select` or `jnp.where` here, instead of Python-level control flow; otherwise the function won't be `jit`-able.

hawkinsp (changes requested):
The other thing I am wondering about with this implementation is: do we end up with sensible gradients? I'd want to verify that; we may need a custom gradient.

One thing I would probably want to do is plot the gradient for a few fixed values of `y` for values of `x` in a small range about 0 using matplotlib.

ayush-1506:
@hawkinsp This is what plotting the `grad(jnp.hypot)` looks like for x starting from 32 and ending at `1.9551592726e-149` (dividing by 2 at every step) for the value of `y = 5.0` (constant).
hypot_grad

- For small values of x, this is almost linear (which makes sense, given the gradient can be approximated as `x/5`)

hawkinsp:
> @hawkinsp This is what plotting the `grad(jnp.hypot)` looks like for x starting from 32 and ending at `1.9551592726e-149` (dividing by 2 at every step) for the value of `y = 5.0` (constant).
> hypot_grad
> 
> * For small values of x, this is almost linear (which makes sense, given the gradient can be approximated as `x/5`)

I'd also check y = 0, but this looks good.

ayush-1506:
For y=0, I get a straight line with grad = 1 for any value of x, so it seems right. @hawkinsp

ayush-1506:
@hawkinsp The tests pass locally, but fail for a certain configuration on CI. Any idea what might be causing this?

jax/_src/numpy/lax_numpy.py
```diff
@@ -986,7 +986,11 @@ def heaviside(x1, x2):
 def hypot(x1, x2):
   _check_arraylike("hypot", x1, x2)
   x1, x2 = _promote_dtypes_inexact(x1, x2)
-  return lax.sqrt(x1*x1 + x2*x2)
+  x1 = lax.abs(x1)
+  x2 = lax.abs(x2)
+  x1, x2 = where(x1 > x2, x1, x2), where(x1 < x2, x1, x2)
+  yx = lax.div(x2, x1)
```
hawkinsp, line 992: This will produce `inf` if `x1` is exactly 0, which is why the CI build is failing.

You need to test and defend against that case. I suspect you need something like the following (aside: I think it may be slightly cleaner to use `max` and `min` rather than `select` or `where`):

```
x1 = lax.abs(x1)
x2 = lax.abs(x2)
x1, x2 = lax.max(x1, x2), lax.min(x1, x2)
return lax.select(x1 == 0, x1, x1 * lax.sqrt(1 + lax.square(lax.div(x2, x1)))
```

Note you should also verify that the gradient in your plots does something sensible at 0 as well: I think you might have omitted 0 before? I suspect the formulation above will produce NaN gradients at 0, so we will probably have to change it to:
```
x1 = lax.abs(x1)
x2 = lax.abs(x2)
x1, x2 = lax.max(x1, x2), lax.min(x1, x2)
return lax.select(x1 == 0, x1, x1 * lax.sqrt(1 + lax.square(lax.div(x2, lax.select(x1 == 0, ones_like(x1), x1))))
```

See https://github.com/google/jax/issues/1052 for an explanation of why the second one may be needed, but I caution you it's pretty complicated.

hawkinsp (changes requested):

jax/_src/numpy/lax_numpy.py
```diff
@@ -986,7 +986,10 @@ def heaviside(x1, x2):
 def hypot(x1, x2):
   _check_arraylike("hypot", x1, x2)
   x1, x2 = _promote_dtypes_inexact(x1, x2)
-  return lax.sqrt(x1*x1 + x2*x2)
+  x1 = lax.abs(x1)
+  x2 = lax.abs(x2)
+  x1, x2 = where(x1 > x2, x1, x2), where(x1 < x2, x1, x2)
```
hawkinsp, line 991: Can we make this:
```
x1, x2 = lax.max(x1, x2), lax.min(x1, x2)
```
?

hawkinsp (changes requested):
Looks good! I suggest one small improvement.

Sorry for the slow review, I was on vacation for the last week or so.

jax/_src/numpy/lax_numpy.py
ayush-1506: I tried that, but tests failed. I don't remember the exact reason, but I believe it had something to do with max and min not being broadcast-able.

hawkinsp: Ah. What about:
```
x1, x2 = maximum(x1, x2), mininimum(x1, x2)
```

?

That uses the `maximum` and `minimum` defined in this file, and they should both broadcast.

ayush-1506: Done, tests pass locally now!

hawkinsp (approved):

hawkinsp:
Thanks for the PR!

PR #4894, opened 2020-11-13
 Fix select_and_gather_add_transpose rule.

LenaMartens:
Fixes a  'jax.ad_util.Zero' is not a valid JAX type error.


mattjj (approved):
LGTM!

PR #4897, opened 2020-11-15
ci-build.yaml: mypy jax failing on the current version of Python

cclauss:
% `python3.9 -m mypy jax`
```
jax/_src/image/scale.py:137: error: syntax error in type comment
jax/_src/image/scale.py:138: error: syntax error in type comment
jax/_src/image/scale.py:140: error: syntax error in type comment
jax/_src/image/scale.py:236: error: syntax error in type comment
jax/_src/image/scale.py:261: error: syntax error in type comment
Found 5 errors in 1 file (checked 147 source files)
```

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -13,16 +13,9 @@ on:
 jobs:
   lint_and_typecheck:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [3.7]
-
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
-      with:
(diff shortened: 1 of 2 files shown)


hawkinsp:
I'm not sure what you're trying to achieve with this change.

* The `mypy` error appears to be related to mypy 0.770. Upgrading to 0.790 fixes it.
* We intentionally test against multiple Python versions, so upgrading to Python 3.9 in the CI builds isn't a great idea. We could add a Python 3.9 build or change one of the existing builds to 3.9.
* The jaxlib version is pinned to the *oldest* jaxlib we support. For Python 3.9, that is indeed 0.1.57, but for other Python versions older jaxlibs are supported. That said, we could just bump the minimum version number everywhere to 0.1.57; in that case we'd also need to update `version.py`.

hawkinsp:
Closing in lieu of #5125

PR #4912, opened 2020-11-16
[jax2tf] Add testing for conversion of comparison operators.

bchetioui:


jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -259,6 +259,10 @@ def _to_np_dtype(dtype) -> NpDType:
       # operations.
       tf_unimpl(np_dtype)
 
+  if prim in [lax.le_p, lax.lt_p, lax.ge_p, lax.gt_p]:
```
gnecula, line 262: I am wondering what we should do here. If we support `bool` by casting to `int8`, why not support `uint16` and `uint32` by casting to `int64`?

gnecula: Or maybe we can just as well keep `bool` unimplemented

bchetioui: I was actually also wondering about that. We can even take it further: why not upcast whenever there is a limitation and it is possible? This would trim the [list of primitives with limitations](https://github.com/google/jax/blob/master/jax/experimental/jax2tf/g3doc/primitives_with_limited_support.md) a lot; but maybe it is better to fail loudly than to silently upcast in most cases. We could also imagine adding an option to the converter to allow/disallow such upcasting. WDYT?

gnecula: I know that the original bool_to_int8 was added because somebody in DM needed it. Casting back and forth is likely going to cost something, and it may be tricky to debug performance discrepancies. Let's remove it from here and we add it later if we really need it? 

I am also wondering if this particular limitation goes away if we compile the TF.

PR #4896, opened 2020-11-14
use `typematch` rather than `==` in cond transpose assertion

froystig:
fixes #4696

PR #4916, opened 2020-11-17
fix grad-of-switch weak type mismatch test

froystig:
As suggested by @jakevdp in https://github.com/google/jax/pull/4896#discussion_r523432702


jakevdp (approved):
Looks great!

PR #4921, opened 2020-11-17
Filter out non-integer index_dtypes in argmin/max.

bchetioui:
Specifying non-integer index dtypes in the two primitives is supported on CPU/TPU, but not on GPU as a custom translation rule is used. This change makes the behavior consistent.

I am not sure where to add a test for it, since this is specifically a test on the primitive rather than the wrapper. The harnesses in #4883 will ensure that this passes once in.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -4863,6 +4863,9 @@ def _argminmax_shape_rule(operand, *, axes, index_dtype):
   return tuple(np.delete(operand.shape, axis))
 
 def _argminmax_dtype_rule(operand, *, axes, index_dtype):
+  if not dtypes.issubdtype(index_dtype, np.integer):
+    raise TypeError("index_dtype must be an integer type, but got {}"
+                    .format(np.dtype(index_dtype).name))
   return index_dtype
 
 def _argminmax_translation_rule(value_comparator, identity,
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -2348,6 +2348,16 @@ class LazyConstantTest(jtu.JaxTestCase):
     make_const = lambda: lax.broadcast_in_dim(arr, (2, 1, 3), (0, 2))
     self._Check(make_const, expected)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_fn={}_indexdtype={}"
+       .format(jax_fn.__name__, np.dtype(index_dtype).name),
+       "index_dtype": index_dtype, "jax_fn": jax_fn}
+      for index_dtype in jtu.dtypes.all_inexact + jtu.dtypes.boolean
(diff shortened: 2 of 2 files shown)


gnecula:
This looks like a good change to me, making the behavior consistent.

hawkinsp (commented):
This needs a test in `lax_test.py`.

jax/_src/lax/lax.py
```diff
@@ -4896,7 +4896,10 @@ def _argminmax_translation_rule(value_comparator, identity,
 def _argminmax_gpu_translation_rule(op, a, *, axes, index_dtype):
   axis, = axes
   idxs = tie_in(a, broadcasted_iota(index_dtype, a.shape, axis))
-  maxval = np.array(dtypes.iinfo(index_dtype).max, dtype=index_dtype)
+  maxval = (dtypes.finfo(index_dtype).max
```
hawkinsp, line 4899: Actually, on reflection: why do we support inexact index types at all? This seems like a bug. Perhaps a better change would be to disallow them in the type rules?

bchetioui: Yes, I was also wondering if it made more sense to add a check in the type rule. By default, since it is possible to align the behaviour on GPU with what was already possible on CPU/TPU, I decided to go that way – I don't know if anyone depends on it. But if you think it is wiser to instead restrict it beforehand, I am also fine with that :)

hawkinsp: I'm comfortable with disallowing it. I doubt anyone depends on it (it seems like an unintentional "feature") and it is untested.

bchetioui:
@hawkinsp thx, I filtered out the non-integer dtypes and added the test in the right place.

tests/lax_test.py
```diff
@@ -2348,6 +2348,16 @@ def testBroadcastInDim(self):
...
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_prim={}_indexdtype={}"
+       .format(prim, np.dtype(index_dtype).name),
+       "index_dtype": dtypes.canonicalize_dtype(index_dtype), "prim": prim}
+      for index_dtype in jtu.dtypes.all_inexact + jtu.dtypes.boolean
+      for prim in [lax.argmin_p, lax.argmax_p]))
+  def testArgMinMaxIndexDtypeError(self, prim, index_dtype):
+    with self.assertRaisesRegex(TypeError,
+                                "index_dtype must be an integer type"):
+      prim.bind(np.ones((2, 2)), axes=(0,), index_dtype=index_dtype)
```
gnecula, line 2360: Wouldn't it be better to test `lax.argmin` instead, since it is the public API?

bchetioui: You're right, this is better.

PR #4923, opened 2020-11-17
Improve pmap caching

apaszke:
Previously each result of `vmap(pmap(f))` would get a distinct
compilation cache, leading to potential unexpected recompilations.

Note that while this patch helps with that issue, it still doesn't
address the same problem in the case of `pmap(vmap(f))`. In general the
global caching mechanism seems very fragile and we might want to
reconsider this guarantee.


hawkinsp:
Closing as stale.

PR #4932, opened 2020-11-17
Support shape parameter in jax.numpy.*_like()

jakevdp:
Also added missing unit tests for `jnp.zeros_like` and `jnp.ones_like`.

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -2232,23 +2232,47 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     with self.assertRaises(TypeError):
       jnp.ones((-1, 1))
 
+  @unittest.skipIf(numpy_version < (1, 17), "shape parameter not supported in older numpy")
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_inshape={}_filldtype={}_outdtype={}".format(
+      {"testcase_name": "_inshape={}_filldtype={}_outdtype={}_outshape={}".format(
           jtu.format_shape_dtype_string(shape, in_dtype),
           np.dtype(fill_value_dtype).name,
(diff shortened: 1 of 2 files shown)


hawkinsp (approved):

hawkinsp:
It looks like you may need to guard the test case against numpy >= 1.17.0 .

jakevdp:
> It looks like you may need to guard the test case against numpy >= 1.17.0 .

Done.

tests/lax_numpy_test.py
```diff
@@ -2232,26 +2232,50 @@ def testOnesWithInvalidShape(self):
...
-    jnp_fun = lambda x, fill_value: jnp.full_like(x, fill_value, dtype=out_dtype)
+  def testFullLike(self, shape, in_dtype, fill_value_dtype, out_dtype, out_shape):
+    rng = jtu.rand_default(self.rng())
+    np_fun = lambda x, fill_value: np.full_like(x, fill_value, dtype=out_dtype, shape=out_shape)
+    jnp_fun = lambda x, fill_value: jnp.full_like(x, fill_value, dtype=out_dtype, shape=out_shape)
     args_maker = lambda: [rng(shape, in_dtype), rng((), fill_value_dtype)]
     self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker)
     self._CompileAndCheck(jnp_fun, args_maker)
 
+  @unittest.skipIf(numpy_version < (1, 17), "shape parameter not supported in older numpy")
```
hawkinsp, line 2258: This is fine. One possible refinement would be to allow the `None` case for earlier NumPy, but I don't think it's worth bothering about it, especially given Numpy 1.16 support is due to be dropped upstream in a few months.

hawkinsp (approved):

PR #4936, opened 2020-11-17
fix mypy error

jakevdp:


hawkinsp (approved):
I don't actually follow why this changes anything: `is_trivial` checks for None and is declare to  accept an optional argument...

jakevdp:
Because mypy is smart enough to see the None check in the function, but not smart enough to crawl other functions and see the None check.

Issue #2748, opened 2020-04-17
autodiff support for jax.numpy.linalg.eig


froystig:
Note that `eigh` is already taken care of.

shoyer:
I'm pretty sure we could achieve this with a minor variation of our existing JVP rule for `eigh`, replacing `U.T.conj()` -> `U**-1` (of course it should really use an LU solve rather than computing the inverse directly).

ianwilliamson:
Just wanted to throw in a +1 for wanting this to be implemented.

j-towns:
@shoyer do you have a reference for it? I've just been working through the math by hand and it seems what you said is correct, except that you have to do a slightly awkward correction to ensure that `dU.T @ U` has ones down the diagonal (which I think is required - this comes from the constraint that the eigenvectors are normalized). Anyway I think I will draft an implementation today.

j-towns:
Also @shoyer how should I do a solve (`inv(a) @ b` for square matrices matrices `a` and `b`)? I think I can't use jax.numpy.linalg.solve from jax.lax_linalg because of circular dependency.

shoyer:
Section 3.1 from this reference in a comment under `eigh_jvp_rule` (in `lax_linalg.py`) works through the general case of how to calculate eigenvector derivative:
https://people.maths.ox.ac.uk/gilesm/files/NA-08-01.pdf

shoyer:
> Also @shoyer how should I do a solve (`inv(a) @ b` for square matrices matrices `a` and `b`)? I think I can't use jax.numpy.linalg.solve from jax.lax_linalg because of circular dependency.

My suggestion would be to use a local import, in the JVP rule, e.g.,
```
def eig_jvp_rule(...):
  from jax.numpy.linalg import solve
  ...
```

You could also try refactoring, but this the usual hack for circular dependency challenges.

hawkinsp:
I'd be tempted to at least try the refactoring of moving the guts of `solve` into `lax_linalg.py`.

j-towns:
Cool, I've moved it in the draft pr, wasn't too bad to do. Still getting incorrect values for the eig derivatives though 😔.

j-towns:
The JVP seems to be correct now that I've relaxed the test tolerance slightly, but the VJPs are way out, I'm not sure why that is yet.

j-towns:
[...]

PR #4941, opened 2020-11-18
Add derivative for eigenvalues

j-towns:
We aren't supporting eigenvectors for now because eigenvectors are not uniquely determined by the input matrix, they're only determined up to 'gauge' (that is multiplication by a complex scalar with absolute value 1). This means that eigenvalue second (and higher) derivatives aren't supported either, because they involve differentiating the eigvals jvp, which itself depends on eigenvectors.

I had to move the body of `solve` from jnp.linalg into lax.linalg, because it's needed in the jvp rule.

This fixes partially https://github.com/google/jax/issues/2748.

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -237,6 +237,33 @@ def triangular_solve(a, b, left_side: bool = False, lower: bool = False,
 
 
 # utilities
+@partial(vectorize, signature='(n,m),(m)->(n)')
+def _matvec_multiply(a, b):
+  return lax.dot(a, b, precision=lax.Precision.HIGHEST)
+
+def _check_solve_shapes(a, b):
+  if not (a.ndim >= 2 and a.shape[-1] == a.shape[-2] and b.ndim >= 1):
+    msg = ("The arguments to solve must have shapes a=[..., m, m] and "
+           "b=[..., m, k] or b=[..., m]; got a={} and b={}")
+    raise ValueError(msg.format(a.shape, b.shape))
+
+def _solve(a, b):
+  _check_solve_shapes(a, b)
+
+  # With custom_linear_solve, we can reuse the same factorization when
+  # computing sensitivities. This is considerably faster.
+  lu_, _, permutation = lu(lax.stop_gradient(a))
+  custom_solve = partial(
+      lax.custom_linear_solve,
+      lambda x: _matvec_multiply(a, x),
+      solve=lambda _, x: lu_solve(lu_, permutation, x, trans=0),
+      transpose_solve=lambda _, x: lu_solve(lu_, permutation, x, trans=1))
+  if a.ndim == b.ndim + 1:
+    # b.shape == [..., m]
+    return custom_solve(b)
+  else:
+    # b.shape == [..., m, k]
+    return api.vmap(custom_solve, b.ndim - 1, max(a.ndim, b.ndim) - 1)(b)
 
 def _T(x): return jnp.swapaxes(x, -1, -2)
 def _H(x): return jnp.conj(_T(x))
@@ -380,6 +407,20 @@ def eig_batching_rule(batched_args, batch_dims, *, compute_left_eigenvectors,
                      compute_right_eigenvectors=compute_right_eigenvectors),
           (0,) * (1 + compute_left_eigenvectors + compute_right_eigenvectors))
 
+def eig_jvp_rule(primals, tangents, *, compute_left_eigenvectors,
+                 compute_right_eigenvectors):
+  if compute_left_eigenvectors or compute_right_eigenvectors:
+    raise NotImplementedError(
+        'The derivatives of eigenvectors are not implemented, only '
+        'eigenvalues. See '
+        'https://github.com/google/jax/issues/2748 for discussion.')
+  # Formula for derivative of eigenvalues w.r.t. a is eqn 4.60 in
+  # https://arxiv.org/abs/1701.00392
+  a, = primals
+  da, = tangents
+  l, v = eig(a, compute_left_eigenvectors=False)
+  return [l], [jnp.sum(_solve(v, da.astype(v.dtype)) * _T(v), -1)]
+
 eig_p = Primitive('eig')
 eig_p.multiple_results = True
 eig_p.def_impl(eig_impl)
@@ -387,6 +428,7 @@ eig_p.def_abstract_eval(eig_abstract_eval)
 xla.translations[eig_p] = eig_translation_rule
 xla.backend_specific_translations['cpu'][eig_p] = eig_cpu_translation_rule
 batching.primitive_batchers[eig_p] = eig_batching_rule
+ad.primitive_jvps[eig_p] = eig_jvp_rule
 
 
 # Symmetric/Hermitian eigendecomposition
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -272,6 +272,30 @@ class NumpyLinalgTest(jtu.JaxTestCase):
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_shape={}".format(
            jtu.format_shape_dtype_string(shape, dtype)),
+       "shape": shape, "dtype": dtype, "rng_factory": rng_factory
+       }
+      for shape in [(4, 4), (5, 5), (8, 8), (7, 6, 6)]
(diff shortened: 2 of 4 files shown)


j-towns:
xref https://github.com/google/jax/issues/2748

jax/_src/lax/linalg.py
```diff
@@ -381,13 +408,28 @@ def eig_batching_rule(batched_args, batch_dims, *, compute_left_eigenvectors,
...
+    raise NotImplementedError(
+        'The derivatives of eigenvectors are not implemented, only '
+        'eigenvalues. See '
+        'https://github.com/google/jax/issues/2748 for discussion.')
+  # Formula for derivative of eigenvalues w.r.t. a is eqn 4.60 in
+  # https://arxiv.org/abs/1701.00392
+  a, = primals
+  da, = tangents
+  l, v = eig(a, compute_left_eigenvectors=False)
+  return [l], [jnp.sum(_solve(v, da.astype(v.dtype)) * _T(v), -1)]
```
shoyer, line 423: Rather than multiply followed by sum, I wonder if we should use a batched vector dot-product here instead? i.e.,
`einsum('ij,ji->i', _solve(v, da.astype(v.dtype), v)`

(or the equivalent using the lower level `lax.dot_general`)

This gives slightly higher level information to XLA, which in principle provides more opportunities for optimization. That said, when I test it out performance seems to be slightly worse! I guess we should leave it alone, then :)

e.g., on CPU:
```python
import jax.numpy as jnp
import jax
import numpy as np

N = 512
x = jnp.array(np.ones((N, N)))
y = jnp.array(np.ones((N, N)))

@jax.jit
def f(x, y):
  return jnp.einsum('ij,ji->i', x, y)

@jax.jit
def g(x, y):
  return (x * y.T).sum(axis=-1)

%timeit f(x, y).block_until_ready()  # 1000 loops, best of 3: 969 µs per loop
%timeit g(x, y).block_until_ready()  # 1000 loops, best of 3: 660 µs per loop
```

jax/_src/lax/linalg.py
```diff
@@ -381,13 +408,28 @@ def eig_batching_rule(batched_args, batch_dims, *, compute_left_eigenvectors,
...
+  if compute_left_eigenvectors or compute_right_eigenvectors:
+    raise NotImplementedError(
+        'The derivatives of eigenvectors are not implemented, only '
+        'eigenvalues. See '
+        'https://github.com/google/jax/issues/2748 for discussion.')
+  # Formula for derivative of eigenvalues w.r.t. a is eqn 4.60 in
+  # https://arxiv.org/abs/1701.00392
+  a, = primals
+  da, = tangents
+  l, v = eig(a, compute_left_eigenvectors=False)
```
shoyer, line 422: Rather than doing another linear solve, why not compute the left-eigenvectors as part of the eigenvalue problem and use those directly? I suspect that could save a bit of work.

tests/linalg_test.py
```diff
@@ -269,6 +269,29 @@ def check_left_eigenvectors(a, w, vl):
...
+      for dtype in float_types + complex_types
+      for rng_factory in [jtu.rand_default]))
+  # TODO(phawkins): enable when there is an eigendecomposition implementation
+  # for GPU/TPU.
+  @jtu.skip_on_devices("gpu", "tpu")
+  def testEigvalsGrad(self, shape, dtype, rng_factory):
+    # This test sometimes fails for large matrices. I (@j-towns) suspect, but
+    # haven't checked, that might be because of perturbations causing the
+    # ordering of eigenvalues to change, which will trip up check_grads. So we
+    # just test on small-ish matrices.
```
shoyer, line 287: One way to solve this would be to explicitly create matrices without near degenerate eigenvalues, e.g., by constructing them via a matrix decomposition. But really, the checks you have here already seem plenty comprehensive to me.

jax/_src/lax/linalg.py
j-towns: This is the kind of thing that I imagine XLA can fuse, I'm slightly surprised that its worse with the user-fused code though. Maybe the jaxpr can tell us something...
```python
   ...: import jax
   ...: import numpy as np
   ...:
   ...: N = 512
   ...: x = jnp.array(np.ones((N, N)))
   ...: y = jnp.array(np.ones((N, N)))
   ...:
   ...: @jax.make_jaxpr
   ...: def f(x, y):
   ...:   return jnp.einsum('ij,ji->i', x, y)
   ...:
   ...: @jax.make_jaxpr
   ...: def g(x, y):
   ...:   return (x * y.T).sum(axis=-1)
   ...:
In [2]: f(x, y)
Out[2]:
{ lambda  ; a b.
  let c = xla_call[ backend=None
                    call_jaxpr={ lambda  ; a b.
                                 let c = dot_general[ dimension_numbers=(((0,), (1,)), ((1,), (0,)))
                                                      precision=None ] b a
                                 in (c,) }
                    device=None
                    donated_invars=(False, False)
                    name=_einsum ] a b
  in (c,) }
In [3]: g(x, y)
Out[3]:
{ lambda  ; a b.
  let c = transpose[ permutation=(1, 0) ] b
      d = mul a c
      e = reduce_sum[ axes=(1,) ] d
  in (e,) }
```
So it seems dot_general on CPU lowers to something slower than the alternative. I would guess this isn't the case on other backends, but I'm gonna leave it as it is for now anyway.

jax/_src/lax/linalg.py
j-towns: This is a really good idea but surprisingly doesn't seemt to work, because of the gauge subtlety rearing its ugly head again. `inv(v).T` is a valid left eigenvector matrix but it seems that it's not the matrix that eig produces (I just learned this from [here](https://math.stackexchange.com/questions/2071964/left-right-eigenvectors)). The tests fail with this approach; I tried doing `left_eigenvectors.T @ right_eigenvectors` in a debugger and it looked pretty far from identity.

tests/linalg_test.py
j-towns: Yeah I agree, but I'm not going to implement this now, just thought it worth putting a note there in case someone has the time and motivation to improve the situation in the future 🙂.

j-towns:
Thanks for the excellent review comments @shoyer!

jax/_src/lax/linalg.py
shoyer: Oh, how awkward! 0.8 is pretty far from 1, well outside what I would consider a reasonable tolerance.

tests/linalg_test.py
```diff
@@ -269,6 +269,29 @@ def check_left_eigenvectors(a, w, vl):
...
+    rng = rng_factory(self.rng())
+    jtu.skip_if_unsupported_type(dtype)
+    args_maker = lambda: [rng(shape, dtype)]
+    a, = args_maker()
+    jtu.check_grads(lambda x: jnp.linalg.eigvals(x), (a,), order=1,
+                    modes=['fwd'], rtol=1e-2, atol=1e-2)
```
shoyer, line 293: Could we check `rev` mode, too? It seems like that would be a worthwhile check to include, since people definitely care about it.

j-towns: Agreed, it appears to be breaking when I turn it on, which is interesting, I'll try and get to the bottom of it tomorrow. By the way, it looks like the derivatives of `solve` might not be be covered by the tests. Searching `linalg_test.py` I've found `testSolve` methods but they don't seem to check the derivatives. Apologies if I'm missing something obvious, it's a little late here.

j-towns: To be clear, it's probably best to avoid merging until this is fixed.

j-towns: Things seem to work fine for float64 and complex128, so I'm assuming this is a numerical precision issue rather than the high level implementation being incorrect. I've enabled reverse mode tests and raised the atol and rtol for float32 and complex64 to 1e-1, I was able to lower the rtol and atol to 1e-4 for the two high precision dtypes.

j-towns: FYI I also tried constructing the random input matrix to make sure it had eigenvalues which wouldn't get re-ordered by small perturbations, but that didn't fix the failures.

j-towns: Hmm I thought that eigenvalues came out of eig in some kind of sorted order, but it seems I was wrong. So maybe there's no way to prevent small perturbations from occasionally changing the order... 😟

shoyer (approved):

PR #4948, opened 2020-11-19
Add vector-valued gradient example to Autodiff Cookbook

gmittal:
@cghawthorne and I both had difficulty immediately finding an easy way to take vector-valued gradients (which `tf.gradients` provides) and felt it would be useful to provide an example. If there is a better place for this to live (or if you don't feel it's necessary in the docs at all) please let me know.


gnecula:
Thanks for this contribution. I think that the Autodiff cookbook is a better place for this example: https://jax.readthedocs.io/en/latest/notebooks/autodiff_cookbook.html

gmittal:
Great. I've moved the example there.

PR #1216, opened 2019-08-21
Add jax.devices() and friends, and add `devices` arg to pmap.

skye:
This change adds the following APIs:
* jax.devices(). This returns a list of available Device subclass instances.
* jax.host_id(). Currently always 0, but will be useful on multi-host platforms.
* jax.local_device_count(). Currently always equal to jax.device_count(), but
    will be useful on multi-host platforms.
* Optional `devices` argument to pmap. This can be used to specify which devices
    should be used in the replicated computation.

PR #4953, opened 2020-11-19
Remove devices from xla.AxisEnv

apaszke:
In most cases they were not even filled in, and they are not used
anywhere.


mattjj:
Looks like these were added in #1216, specifically [here](https://github.com/google/jax/commit/ae835b747e5b6689e3835ca238a10aea9517f770#diff-121975a1a88d61a259f8423655e23a1b389eb18c399b267395d2aba5a196bae9R316). I suspect they're not needed but just in case, @skye any extra context?

skye (approved):

skye:
@mattjj I somehow missed your comment. I honestly don't remember the reasoning, but it looks like we only used it for error checking, which is now removed.

PR #4959, opened 2020-11-19
Change round/round_p API and implement jax2tf conversion & tests

### LINKED ISSUE #4952: [jax2tf] Incorrect translation for jax.lax.round
```
import jax
import jax.experimental.jax2tf as jax2tf
import jax.numpy as jnp
import tensorflow as tf
tf.compat.v1.enable_eager_execution()

print("Jax:", jax.lax.round(0.5))
print("Jax2Tf:", jax2tf.convert(jax.lax.round)(0.5))
```

```
Jax: 1.0
Jax2Tf: tf.Tensor(0.0, shape=(), dtype=float32)
```

The issue is coming from the fact that jax.lax.round uses away from zero rounding based on what the corresponding XLA op is doing but then it maps to tf,round what does round towards even.

bchetioui:
Fixes google/jax#4952.

### DIFF
--- a/docs/jax.lax.rst
+++ b/docs/jax.lax.rst
@@ -196,3 +196,14 @@ Linear algebra operators (jax.lax.linalg)
     qr
     svd
     triangular_solve
+
+Argument classes
+----------------
+
+.. automodule:: jax.lax
+
+.. autoclass:: ConvDimensionNumbers
+.. autoclass:: GatherDimensionNumbers
+.. autoclass:: Precision
+.. autoclass:: RoundingMethod
+.. autoclass:: ScatterDimensionNumbers
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -16,10 +16,12 @@
 # pytype: skip-file
 
 import builtins
+from enum import IntEnum
 import functools
 import itertools
 import operator
-from typing import (Any, Callable, List, NamedTuple, Optional, Sequence, Union, Tuple)
+from typing import (Any, Callable, List, NamedTuple, Optional, Sequence,\
+                    Union, Tuple)
 import warnings
 
 import numpy as np
@@ -150,12 +152,28 @@ def ceil(x: Array) -> Array:
   r"""Elementwise ceiling: :math:`\left\lceil x \right\rceil`."""
   return ceil_p.bind(x)
 
-def round(x: Array) -> Array:
+class RoundingMethod(IntEnum):
+  AWAY_FROM_ZERO = 0
+  TO_NEAREST_EVEN = 1
+
+def round(x: Array,
+          rounding_method: RoundingMethod = RoundingMethod.AWAY_FROM_ZERO
+          ) -> Array:
   r"""Elementwise round.
 
-  Rounds values to the nearest integer. Halfway values (e.g., `0.5`) are rounded
-  away from zero."""
-  return round_p.bind(x)
+  Rounds values to the nearest integer.
+
+  Args:
+    x: an array or scalar value to round.
+    rounding_method: the method to use when rounding halfway values
+      (e.g., `0.5`). See ``lax.RoundingMethod`` for the list of possible
+      values.
+
+  Returns:
+    An array containing the elementwise rounding of x.
+  """
+  rounding_method = RoundingMethod(rounding_method)
+  return round_p.bind(x, rounding_method=rounding_method)
 
 def is_finite(x: Array) -> Array:
   r"""Elementwise :math:`\mathrm{isfinite}`.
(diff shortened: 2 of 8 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -854,7 +854,16 @@ def _unexpected_primitive(p: core.Primitive, *args, **kwargs):
 tf_impl[lax.sign_p] = tf.math.sign
 tf_impl[lax.floor_p] = tf.math.floor
 tf_impl[lax.ceil_p] = tf.math.ceil
-tf_impl[lax.round_p] = tf.math.round
+
+def _round(operand):
```
gnecula, line 858: I am a bit surprised that we need such a translation, especially the conditional. I am curious what is the bug in the naive translation before? I would have guessed that it is about the sign, but the conditional seems to be a different one.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -854,7 +854,16 @@ def _unexpected_primitive(p: core.Primitive, *args, **kwargs):
...
+  sign = tf.math.sign(operand)
+  operand *= sign
+  floor = tf.math.floor(operand)
+  operand -= floor
+  cond = tf.math.equal(operand, tf.constant(np.array(0.5), operand.dtype))
+  return sign * (tf.where(cond, 2 * operand, tf.math.round(operand)) + floor)
```
gnecula, line 864: I find the equality comparison with 0.5 suspicious. Should it be <= ? The 2 * operand also seems strange.

jax/experimental/jax2tf/jax2tf.py
bchetioui: The issue before is, as outline in #4952, JAX rounds away from 0 while TF rounds to the nearest even number.

jax/experimental/jax2tf/jax2tf.py
bchetioui: `2 * operand` is indeed weird (leftover from a previous attempt where `abs` was not used and the first condition could match several values). I've replaced it by `tf.constant(np.array(1), operand.dtype)` (which is still `2 * operand` in this case, for the relevant values, except that these values are always `0.5`).
The comparison is accurate, it is meant to make rounding consistent when the decimal part is equal to 0.5, when JAX and TF take different arbitrary rounding choices (the former rounds away from 0, and the latter to the nearest even number).

jax/experimental/jax2tf/jax2tf.py
tberghammer: More specifically `jnp.round` rounds towards even (and gets mapped to a soup of HLO and thus to a soup of TF ops instead of `tf.round`) while `lax.round` rounds towards 0.

bchetioui: Indeed `jnp.round` should mirror the behavior of `numpy.round`, which is to round towards even, like `tf.round`.
But we are not able to recognize that from the jax2tf side, because `jax2tf` implements conversion only at the primitive level in JAX (i.e. for `lax.round_p`, which is called by `lax.round`; there is no direct conversion for `jnp.round`).

Also to clarify, re. the `lax.round` behaviour, it indeed rounds away from 0 (which you correctly pointed out in the issue :)), see
```python
>>> from jax import lax
>>> import numpy as np
>>> lax.round(np.array([-0.5, -2.5, -1.5, 1.5, 2.5, 0.5], dtype=np.float32))
DeviceArray([-1., -3., -2.,  2.,  3.,  1.], dtype=float32)
```

hawkinsp: I don't think we'd object if you made the `round` primitive support multiple rounding modes (via a keyword arg) to make it easier to jax2tf to translate to `tf.round` directly in the cases that it applies.

gnecula: Thank you both for explaining. It seems that the round situation is complicated. `jnp.round` is actually implemented much like in this PR, without using `lax.round_p`, so the conversion to TF is irrelevant. Do you think it makes sense to reuse [_round_to_nearest_even](https://github.com/google/jax/blob/1e715121a2ab8020929d9550b7d24e7f36e9ad64/jax/_src/numpy/lax_numpy.py#L1699) for the TF conversion of `lax.round_p`?  

The only use of `lax.round_p` is by way of `lax.round`. I think it makes sense to add a docstring for `lax.round` to explain that this has different behavior than `numpy.round`. 

`lax.round` is used in `jnp.divmod`, which itself has some conditionals preceeding the use.  This means that a conversion of `jnp.divmod` will have both the conditionals from `jnp.divmod` and those from the translation to TF. I do not see an easy way to avoid this, except inventing a mechanism for `jax.custom_jax2tf` to attach to `jnp.divmod`. I do not think it is worth it at this moment.

bchetioui: I don't think we can reuse `_round_to_nearest_even` for the TF conversion of `lax.round_p`, since rounding to the nearest even in TF is what we are trying to circumvent (we want to instead round away from 0!).

I think that moving the logic for different rounding strategies inside `round_p` and parameterizing with a kwarg as @hawkinsp suggested is sensible and would circumvent the issue in this case :)

bchetioui: I implemented and pushed the suggested change. PTAL whenever you get the chance :)

jax/_src/lax/lax.py
```diff
@@ -150,12 +152,30 @@ def ceil(x: Array) -> Array:
   r"""Elementwise ceiling: :math:`\left\lceil x \right\rceil`."""
   return ceil_p.bind(x)
 
-def round(x: Array) -> Array:
+class RoundingMethod(Enum):
```
hawkinsp, line 155: I wonder if this should be `IntEnum`. That might remove the need for `RoundingMethodLike`?

hawkinsp (commented):
You might need to also do something to `lax.rst` to make `lax.RoundingMethod` show up in the docs?

jax/_src/lax/lax.py
bchetioui: Thanks! I did not know about `IntEnum`, it indeed seems very appropriate :)

bchetioui:
I modified `jax.lax.rst` to add an "Argument wrappers" section to the doc! I hope it is an appropriate name :)

docs/jax.lax.rst
```diff
@@ -196,3 +196,14 @@ Linear algebra operators (jax.lax.linalg)
     qr
     svd
     triangular_solve
+
+Argument wrappers
```
hawkinsp, line 200: I might call this "Types" or "Classes" or something like that. I don't find "wrappers" very descriptive.

hawkinsp (approved):

PR #4963, opened 2020-11-19
[jax2tf] Added testing code for div/rem

bchetioui:


jax/experimental/jax2tf/g3doc/primitives_with_limited_support.md
```diff
@@ -93,6 +93,8 @@ conversion to Tensorflow.
 | conv_general_dilated | Missing TF support | Primitive is unimplemented in TF; likely bug in the HLO -> LLVM IR lowering of XlaConv | complex128, complex64 | CPU, GPU, TPU |
 | cosh | Missing TF support | Primitive is unimplemented in TF | float16 | CPU, GPU, TPU |
 | digamma | Missing TF support | Primitive is unimplemented in TF | bfloat16 | CPU, GPU |
+| div | Missing TF support | Primitive is unimplemented in TF | int16, int8, uint16, uint32, uint64, uint8 | CPU, GPU, TPU |
+| div | Missing TF support | Primitive is unimplemented in TF; integer division fails if the divisor contains a 0 | int32, int64 | CPU, GPU, TPU |
```
gnecula, line 97: What happens in JAX for division by 0? Nan I Assume?

bchetioui: No! `nan` is not a valid value for integers, so JAX systematically returns `dtype(-1)` in this case (i.e. `-1` for `int8`, `255` for `uint8`, etc…)

gnecula: I see, thanks! I was wondering if this should be documented somewhere.

PR #4965, opened 2020-11-19
Fix build.py to work on Linux once again.

hawkinsp:
* strip DOS end-of-line characters from build.py for consistency with the rest of the source tree.
* use shutil.copy() instead of shutil.copyfile(). On Unix systems we must preserve execute permissions.
* add code to explicitly delete and recreate the target directory.
* Move build/jaxlib/__init_py to jaxlib/__init__.py and have the script move it into position, so the output directory for the jaxlib is an empty directory that the script creates.

### DIFF
--- a/build/install_xla_in_source_tree.py
+++ b/build/install_xla_in_source_tree.py
@@ -26,35 +26,37 @@ args = parser.parse_args()
 
 r = runfiles.Create()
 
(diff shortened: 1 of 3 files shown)


hawkinsp:
@cloudhan FYI. I needed to make a few changes to get things working on Linux again. I don't think this will break the Windows build but there is a chance that it may.

PR #4972, opened 2020-11-20
Add support for multi-host partitioning when using pmap(sharded_jit).

skye:
This extends the pmap logic in a way similar to https://github.com/google/jax/pull/4746. The new arguments to sharded_jit specifying the local partitioning can be reused by pmap, but with one wrinkle: the pmap implementation needs to trace its jaxpr to "see" the sharded_jit and get these values, but it needs to know the global aval shapes in order to correctly trace through the sharded_jit. For now, we simply add this information as a new "global_arg_shapes" argument to pmap. Ideally we'll replace this with a more elegant solution, e.g. global-view device arrays.

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -504,11 +504,13 @@ xla.canonicalize_dtype_handlers[ShardedDeviceArray] = identity
 ### the xla_pmap primitive and its rules are comparable to xla_call in xla.py
 
 def xla_pmap_impl(fun: lu.WrappedFun, *args, backend, axis_name, axis_size,
-                  global_axis_size, devices, name, in_axes, donated_invars):
+                  global_axis_size, devices, name, in_axes, donated_invars,
+                  global_arg_shapes):
   abstract_args = unsafe_map(xla.abstractify, args)
   compiled_fun = parallel_callable(fun, backend, axis_name, axis_size,
                                    global_axis_size, devices, name, in_axes,
-                                   donated_invars, *abstract_args)
+                                   donated_invars, global_arg_shapes,
+                                   *abstract_args)
   return compiled_fun(*args)
 
 @lu.cache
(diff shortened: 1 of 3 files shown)


jax/interpreters/pxla.py
```diff
@@ -650,16 +685,16 @@ def dynamic_fun(dummy, *args):
   logging.log(log_priority,
               f"Compiling {fun.__name__} for {num_global_shards} devices with "
               f"args {avals}. (num_replicas={num_global_replicas} "
-              f"num_partitions={num_partitions}")
+              f"num_partitions={num_partitions})")
```
jekbradbury, line 688:
```suggestion
              f"num_partitions={num_partitions}")
```

jax/interpreters/pxla.py
```diff
@@ -757,19 +793,27 @@ def dynamic_fun(dummy, *args):
...
     Optional[Tuple[PartitionsOrReplicated, ...]],
     Optional[Tuple[PartitionsOrReplicated, ...]],
-    int]:
-  """Returns (in_partitions, out_partitions, num_partitions)."""
+    int,
+    Optional[Tuple[PartitionsOrReplicated, ...]],
+    Optional[Tuple[PartitionsOrReplicated, ...]],
+    Optional[int]]:
+  """Returns (in_partitions, out_partitions, num_partitions, local_in_parts,
+              local_out_parts_thunk, local_num_partitions).
```
jekbradbury, line 801:
```suggestion
              local_out_partitions, local_num_partitions).
```

jekbradbury (approved):

jax/interpreters/pxla.py
skye: This closes a paren in the comment

jax/interpreters/pxla.py
skye: Done.

PR #4974, opened 2020-11-20
Add -Wno-stringop-truncation to build flags on Linux.

hawkinsp:
Works around https://github.com/tensorflow/tensorflow/issues/39467 for gcc 10+ builds.


aviallon:
Thanks

PR #4977, opened 2020-11-20
[jax2tf] Fix and test conversion of iota

bchetioui:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -918,6 +918,7 @@ def _conj(x, **kwargs):
 
 
 def _iota(*, dtype, shape, dimension):
+  dtype = to_tf_dtype(dtype)
```
bchetioui, line 921: This is needed for `dtypes.bfloat16` to be properly converted to `tf.bfloat16`, and makes it clear that our dtype is in the TF world :)

gnecula: Good catch!

gnecula (approved):

bchetioui:
Closed in favor of #4984

PR #4982, opened 2020-11-20
Change jaxlib build rules to build a wheel package.

hawkinsp:
Before this PR, the jaxlib build script (`build/build.py`) writes its output to `build/` subdirectory of the jax source tree. This is a bit ugly since it intermingles the jax source tree with the artifacts built from it (an unpacked wheel).

This PR changes the build scripts to build a `.whl` package as the output of running `build.py`, which is by default written to the `dist/` subdirectory of the current directory.

Also remove mentions of `cython` as a build dependency. We don't need the user to provide `cython`, bazel downloads it. `six` also seems unused.


hawkinsp:
@cloudhan

cloudhan:
Hmm, it take me time to figure out why `from wheel import pep425tags` fails, turns out it has been removed in 0.35 of wheel
See https://github.com/pypa/wheel/issues/370 and https://github.com/pypa/wheel/commit/a51977075740fda01b2c0e983a79bfe753567219

hawkinsp:
> Hmm, it take me time to figure out why `from wheel import pep425tags` fails, turns out it has been removed in 0.35 of wheel
> See [pypa/wheel#370](https://github.com/pypa/wheel/issues/370) and [pypa/wheel@a519770](https://github.com/pypa/wheel/commit/a51977075740fda01b2c0e983a79bfe753567219)

I just noticed that too. I already pushed a new version that fixes this.

cloudhan:
`tmpdir.cleanup()` is causing the same problem.

hawkinsp:
> `tmpdir.cleanup()` is causing the same problem.

Can you verify you are using commit c06ead6b04f3b1190e1c24589ed98d32231134a0 ?

In particular I was seeing that until I added an extra `os.chdir` to make sure the current directory wasn't inside the temp dir when it was deleted.

cloudhan:
I think I'll wait you merge this PR and fix that later.

cloudhan:
I am on that commit. But `cleanup` is calling into `rmtree` and `rmtree` will refuse to delete readonly pyd file. 😂

PR #4986, opened 2020-11-20
Add np.intc to the set of valid jaxtypes.

hawkinsp:
Fixes a number of test failures on Windows.

### DIFF
--- a/jax/abstract_arrays.py
+++ b/jax/abstract_arrays.py
@@ -47,7 +47,7 @@ array_types = {np.ndarray, np.bool_,
                np.uint8, np.uint16, np.uint32, np.uint64,
                dtypes.bfloat16, np.float16, np.float32, np.float64,
                np.complex64, np.complex128,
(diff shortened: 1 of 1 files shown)


hawkinsp:
@cloudhan

jakevdp:
Is there any reasonable regression test we can add for this that will run on our CI?

hawkinsp:
> Is there any reasonable regression test we can add for this that will run on our CI?

I don't think so. We'd need a windows CI build (which we should add, but not yet...)

hawkinsp:
> > Is there any reasonable regression test we can add for this that will run on our CI?
> 
> I don't think so. We'd need a windows CI build (which we should add, but not yet...)

To add some detail to this, `np.int32 is np.intc` is true on Mac and Linux but false on Windows. So there's no way to distinguish `np.int32` and `np.intc` on our current CI platforms as far as I can tell.

PR #4987, opened 2020-11-20
Add dynamically methods to _DeviceArray and PyBuffer.

jblespiau:
As you suggested during a review, calling `_DeviceArray` directly when LazyExpr are present is clearer.
We need to remove the weakref slot, otherwise we get:

TypeError: __weakref__ slot disallowed: either we already got one, or __itemsize__ != 0

### DIFF
--- a/jax/_src/dlpack.py
+++ b/jax/_src/dlpack.py
@@ -19,7 +19,7 @@ import jax.lib
 from jax.lib import xla_client
 from jax.lib import xla_bridge
 
-def to_dlpack(x: xla.DeviceArray, take_ownership: bool = False):
+def to_dlpack(x: xla.DeviceArrayProtocol, take_ownership: bool = False):
   """Returns a DLPack tensor that encapsulates a DeviceArray `x`.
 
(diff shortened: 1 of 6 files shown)


mattjj:
Looks like #1180 effectively added `__weakref__` to DeviceArray, though unfortunately it didn't add a test.

@brianwa84 do you still need DeviceArrays to be weakref-able?

jblespiau:
(I have note tested it, but it's is possible that they are still weak-ref friendly, because I think pybind11 classes, are weak-ref friendly by default). If it's still needed, I guess we should add a test.

brianwa84:
I expect we have tests in TFP that would break if weakref support were broken. I can think of a change that might allow us to get rid of this, but it's a user-visible API change to bijectors and transformed distributions.

jblespiau:
I have added a test in api_test.py:

```
  def test_devicearray_weakref_friendly(self):
    x = device_put(1.)
    y = weakref.ref(x)
    self.assertEqual(y(), 1.)
    del x
    self.assertIsNone(y())
```

mattjj (approved):

PR #4988, opened 2020-11-20
Fix lax_reference population_count implementation to avoid overflowin…

hawkinsp:
…g Python int -> np.uint32 casts.

These cause errors on Windows. Instead, start with np.uint64s and truncate using numpy.


hawkinsp:
@cloudhan

PR #4989, opened 2020-11-20
Make jnp.gradient() call signature match np.gradient

jakevdp:


tests/lax_numpy_test.py
```diff
@@ -4497,7 +4497,6 @@ def testWrappedSignaturesMatch(self):
       'eye': ['order'],
       'full': ['order'],
       'full_like': ['subok', 'order'],
-      'gradient': ['varargs', 'axis', 'edge_order'],
```
froystig, line 4500: Did `axis` become unsupported in this change?

froystig (approved):

tests/lax_numpy_test.py
jakevdp: axis is still supported, but now is supported as an explicit keyword, rather than via ``**kwargs``

froystig: I misread the diff! Got it now.

PR #4991, opened 2020-11-20
Support 'where' argument in jax.numpy reductions

jakevdp:


froystig (approved):
Nice!

domenicjdonato:
I think this change breaks jax.nn.softmax(tensor, axis=-1). Can you please double check this?

jakevdp:
> I think this change breaks jax.nn.softmax(tensor, axis=-1). Can you please double check this?

Perhaps... could you open an issue with an example of the error?

PR #4993, opened 2020-11-20
Fix random doc

### LINKED ISSUE #4992: JAX Random Docs Page is Blank
Seems the page for Jax.random is not generating correctly...?

https://jax.readthedocs.io/en/latest/jax.random.html

jakevdp:
Fixes #4992 

I don't totally understand why, but the sphinx automodule stopped working after #4964. I tried using `:private-members:`, but that didn't help. I finally landed on explicitly listing the functions as we do in `lax.numpy.rst`. Confirmed locally that this results in a generated page (Note: the readthedocs link below goes to the first version of this PR, which still does not have all functions).


docs/jax.random.rst
```diff
@@ -2,5 +2,45 @@ jax.random package
...
+    normal
+    pareto
+    permutation
+    poisson
+    rademacher
+    randint
+    shuffle
+    split
+    t
+    threefry_2x32
```
mattjj, line 42: Maybe this one is internal (?)

jax/random.py
```diff
@@ -12,6 +12,68 @@
...
 
+"""Utilities for pseudo-random number generation.
+
+The ``jax.random`` package provides a number of routines for deterministic
+generation of sequences of pseudorandom numbers.
+
+Basic usage
+-----------
+
+>>> rng = jax.random.PRNGKey(seed)
```
mattjj, line 23: Slight preference for calling this `key`, since `rng` implies it's more of a heavyweight object imo, but this is a minor preference so feel free to ignore it.

mattjj (approved):
Beautiful!

docs/jax.random.rst
jakevdp: Done

jax/random.py
jakevdp: Done

PR #5003, opened 2020-11-24
use strict action env on windows to avoid constant full rebuilding

cloudhan:
On windows, without `experimental_strict_action_env`, rebooting or maybe terminal restarting results in a full rebuild.


mattjj (approved):
Thank you!

cloudhan:
the doc build crashed?

mattjj (approved):

PR #5005, opened 2020-11-24
Make xmap into a primitive

apaszke:
Add a compilation cache. Also make sure that it raises a clear error
when you try to use it with other transforms.

Also, add a bunch of checks to make sure that the arguments are valid.

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -524,7 +524,7 @@ def jaxpr_replicas(jaxpr):
   For a eqn, multiply the `axis_size` with the `jaxpr_replicas` of the
   subjaxprs. For a list of eqns, take the maximum number of replicas.
   """
-  return max((eqn_replicas(eqn) for eqn in jaxpr.eqns), default=1)
+  return max(unsafe_map(eqn_replicas, jaxpr.eqns), default=1)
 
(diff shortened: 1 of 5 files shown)


jax/interpreters/xla.py
```diff
@@ -524,24 +524,21 @@ def jaxpr_replicas(jaxpr):
...
   """
-  return max((eqn_replicas(eqn) for eqn in jaxpr.eqns), default=1)
+  return max(unsafe_map(eqn_replicas, jaxpr.eqns), default=1)
 
 # TODO(mattjj): this function assumes that only pmap has a parameter named
 # axis_size, and that it corresponds to cross-replica mapping
 def eqn_replicas(eqn):
   call_jaxpr = eqn.params.get("call_jaxpr")
-  if call_jaxpr:
+  if eqn.primitive.map_primitive:
```
mattjj, line 533: Is this correct? It seems like if we have a call primitive that _contains_ a map primitive (even though it itself is not a map primitive) then we need to recurse. WDYT?

mattjj (approved):
LGTM modulo the one potential bug I mentioned in a comment!

jax/interpreters/xla.py
apaszke: Oh, yeah. You're right of course. Let me fix that!

PR #5007, opened 2020-11-24
jnp.array(): avoid unnecessary numpy array copy

jakevdp:
Currently calling `jnp.array(x)` for a numpy array `x` may create an unnecessary copy of `x` in numpy before pushing it to the device. Passing `copy=False` prevents this.

Related to #4486


mattjj (approved):
Good idea!

PR #2337, opened 2020-02-28
Change psum to perform a single tuple AllReduce, rather than one AllR…

hawkinsp:
…educe per PyTree element.

When performing AllReduces compiled on separate GPU or TPU hosts, it is important that the same AllReductions are performed in the same order. So far we have mostly been relying on compiler determinism to ensure that this is the case, but on GPU it is not a safe assumption because XLA may be nondeterministic. One example is that CuDNN autotuning may in principle cause completely different HLO schedules based on things like the temperature in the room.

Since XLA supports a tuple-at-a-time all-reduce, we can use it here to ensure that the correct behavior happens, provided there is only a single AllReduce in the program. If there are multiple AllReduces, ultimately we will need to force an order by threading a token value between them.

Tuple AllReduce is buggy on CPU, so we lower to the non-tuple variant there.

PR #5011, opened 2020-11-24
unify reduction collectives' implementations

mattjj:
The implementation of `psum` had diverged from that of `pmin` and `pmax` because the former had been adapted to to generate a single XLA collective ops with a tuple of inputs and outputs, rather than many separate collective ops. (See #2337 for the motivation to make `psum` operate on tuples.)

This PR updates the `pmin` and `pmax` implementations to follow that of `psum`, so that they can share code. This fixes a TODO from #4835.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -402,7 +402,7 @@ def _allreduce_translation_rule(prim, c, *args, axis_name, axis_index_groups,
     n = len(dtype_args)
     if is_complex and prim is lax.add_p:
       # TODO(b/141575627): we handle complex-dtype sum-reduction directly as a
-      # special case because it's not currently handled by XLA:GPU or XLA:CPU
+      # special case because it's not currently handled by XLA:GPU
       dtype_args = ([xops.Real(x) for x in dtype_args] +
                     [xops.Imag(x) for x in dtype_args])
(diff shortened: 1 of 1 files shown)


jax/_src/lax/parallel.py
```diff
@@ -420,14 +400,15 @@ def _psum_translation_rule(c, *args, axis_name, axis_index_groups, axis_env,
   for dtype, (indices, dtype_args) in sorted(args_by_type.items()):
     is_complex = dtypes.issubdtype(dtype, np.complexfloating)
     n = len(dtype_args)
-    if is_complex:
+    if is_complex and prim is lax.add_p:
+      # we handle complex-dtype sum-reduction directly as a special case
```
apaszke, line 404: It would be good to add some extra rationale for why is it done this way (if you remember).

apaszke (approved):
LGTM

jax/_src/lax/parallel.py
mattjj: Good point. I think the reason is/was that XLA did/didn't support collectives on complex values, but if that's the case then I should link a bug.

mattjj: Aha, it's b/141575627, where some folks commented: "The reason that the GPU doesn't support complex(AllReduce) is because nccl library doesn't support complex data types". Still, XLA:GPU could solve this for us.

I'll link the bug in a comment. Nice catch!

PR #5012, opened 2020-11-24
Call asarray() rather than array() to avoid host round-trips.

jakevdp:
Related to #4486, I was looking at some memory performance issues and realized that `jnp.array()`, unlike `jnp.asarray()`, will round-trip DeviceArray inputs to numpy arrays stored on the host before pushing them back to device: https://github.com/google/jax/blob/30bb4cf58676c5d173cbbe68b86c5a3b2a8966b0/jax/_src/numpy/lax_numpy.py#L2594-L2596

This makes it a very poor choice for library functions to call on inputs that may be device arrays: it adds an unnecessary round-trip to the host in order to copy the data.

This PR rectifies all the places in jax where I found this inadvertent round-trip.

### DIFF
--- a/jax/_src/nn/functions.py
+++ b/jax/_src/nn/functions.py
@@ -288,7 +288,7 @@ def one_hot(x, num_classes, *, dtype=jnp.float64):
   x = jnp.asarray(x)
   lhs = x[..., jnp.newaxis]
   rhs = lax.broadcast_to_rank(jnp.arange(num_classes, dtype=x.dtype), lhs.ndim)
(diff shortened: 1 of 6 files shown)


mattjj:
Should we instead revise `array` not to do the round-tripping?

jakevdp:
> Should we instead revise array not to do the round-tripping?

@hawkinsp put in a TODO for that at some point; I think that would be ideal but this is a quicker fix for now.

mattjj (approved):

PR #5014, opened 2020-11-25
Add support for nested xmaps

apaszke:
With the caveat that the outer xmap cannot use the `vectorize` resource
mapping, because the batching rule for xmap is not implemented yet (this
is going to be a follow up). But as long as the outer xmap only uses
real devices the nesting should work fine.

Note that we don't try to hard to handle perfectly nested `xmap`s
efficiently. But this shouldn't be an issue, because it is usually
trivial to flatten them on the user side.

Very similar logic can now be used to implement a `sharded_jit` version
that composes well with `xmap` (although it is unclear if it's an API
that we want to have in the long run).

### DIFF
--- a/jax/experimental/general_map.py
+++ b/jax/experimental/general_map.py
@@ -18,7 +18,8 @@ import contextlib
 import numpy as np
 import itertools as it
 from collections import namedtuple, OrderedDict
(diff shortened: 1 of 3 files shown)


mattjj (approved):
Wooo such fast progress!

PR #5018, opened 2020-11-25
X32 tests: fail on dtype warnings

jakevdp:
Any code in JAX that requests an unavailable dtype results in a warning. For example, in X32 mode:
```python
>>> import jax.numpy as jnp
>>> jnp.arange(10, dtype='float64')                                                                             
/Users/vanderplas/github/google/jax/jax/_src/numpy/lax_numpy.py:2744: UserWarning: Explicitly requested dtype float64 requested in arange is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
DeviceArray([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.], dtype=float32)
```
Currently, there are many places in the jax internals that generate this warning, which is not a great experience for users and contributes to warning fatigue. This issue is not caught because currently, our tests explicitly ignore these warnings.

This PR rectifies the situation: it removes the warning filter from the test suite, and fixes several remaining instances of this warning being generated by internal code & tests.

Related: #5019

### DIFF
--- a/jax/_src/image/scale.py
+++ b/jax/_src/image/scale.py
@@ -54,7 +54,7 @@ def compute_weight_mat(input_size: int, output_size: int, scale,
   # want to interpolate.
   kernel_scale = jnp.maximum(inv_scale, 1.) if antialias else 1.
 
-  sample_f = ((np.arange(output_size) + 0.5) * inv_scale -
+  sample_f = ((jnp.arange(output_size) + 0.5) * inv_scale -
               translation * inv_scale - 0.5)
   x = (
       jnp.abs(sample_f[jnp.newaxis, :] -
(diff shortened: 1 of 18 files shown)


froystig (approved):

hawkinsp:
Note that in the past it was at least somewhat intentional that we test 64-bit types for many operations in non-x64 mode. The expected behavior is that they work but produce the warning. Is a consequence of this change that we have lost all such testing?

jakevdp:
Yes, aside from places where the error is explicitly silenced, we no longer test some of those situations.

PR #5019, opened 2020-11-25
X32: don't warn when builtin type passed as dtype

jakevdp:
Currently in X32 mode, specifying `int`, `float`, or `complex` in place of a dtype leads to a warning:
```python
In [2]: jnp.arange(4, dtype=float)                                                                                  
/Users/vanderplas/github/google/jax/jax/_src/lax/lax.py:6272: UserWarning: Explicitly requested dtype <class 'float'>
requested in arange is not available, and will be truncated to dtype float32. To enable more dtypes, set the
jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See
https://github.com/google/jax#current-gotchas for more.
Out[2]: DeviceArray([0., 1., 2., 3.], dtype=float32)
```
This is a legacy of the way that builtin types are handled by numpy's dtype system (i.e. they're interpreted as 64-bit). In JAX, I believe it would make sense to silence these warnings for builtin scalar types, because users rarely mean to strictly specify `float64` when passing `dtype=float`.

More broadly, I think this is a useful part of a larger effort to remove unnecessary warnings in JAX, to avoid getting to where users become accustomed to ignoring them.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6256,6 +6256,9 @@ def _abstractify(x):
 
 
 def _check_user_dtype_supported(dtype, fun_name=None):
+  # Avoid using `dtype in [...]` becuase of numpy dtype equality overloading.
+  if isinstance(dtype, type) and dtype in {bool, int, float, complex}:
+    return
   np_dtype = np.dtype(dtype)
   if np_dtype.kind not in "biufc" and np_dtype.type != dtypes.bfloat16:
     msg = f"JAX only supports number and bool dtypes, got dtype {dtype}"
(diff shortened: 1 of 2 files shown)


mattjj (approved):
LGTM! Copied from chat:

I originally thought this kind of thing _should_ be a warning because in numpy .astype(float) means f64, but in jax numpy with jax_enable_x64=False it'd be f32. However, I like the point about not conditioning users to ignore warnings. So the new philosophy is to interpret .astype(float) as "give me a float dtype (per whatever x64 policy jax has)" and not "give me a float64 like numpy would"!

PR #5020, opened 2020-11-25
add experimental pdot primitive, basic tests

mattjj:
This PR adds `pdot_p` along with impl, vmap, and xla translation rules.

I also added a rudimentary implementation of `DynamicJaxprTrace.process_xmap` (defined in general_map.py for now).

The tests here are super basic at the moment, but on the other hand there's not much to the `pdot` implementation!

For posterity, some [xmap pdot examples](https://gist.github.com/mattjj/ba9b24df446a90902d7b41aeb0766a99).

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -337,6 +337,13 @@ def axis_index(axis_name):
   return axis_index_p.bind(axis_name=axis_name)
 
 
+def pdot(x, y, axis_name, pos_contract=((), ())):
+  if not isinstance(axis_name, (list, tuple)):
+    axis_name = (axis_name,)
+  return pdot_p.bind(x, y, axis_name=axis_name,
+                     pos_contract=pos_contract, pos_batch=[(), ()])
+
+
 ### parallel primitives
 
 def _allreduce_soft_pmap_rule(prim, reducer, vals, mapped, chunk_size,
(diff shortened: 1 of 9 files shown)


jax/_src/lax/parallel.py
```diff
@@ -749,6 +756,70 @@ def _process_axis_index(self, frame):
...
+                    pos_batch=pos_batch)
+  return out, result_batch_dim
+batching.primitive_batchers[pdot_p] = _pdot_vmap_batching_rule
+
+def _pdot_translation_rule(c, x, y, *, axis_name, pos_contract, pos_batch,
+                           axis_env, platform):
+  assert axis_name
+  local_out = lax._dot_general_translation_rule(
+      c, x, y, dimension_numbers=[pos_contract, pos_batch], precision=None)
+  out_tup = xla.parallel_translations[psum_p](
```
apaszke, line 805: Shouldn't we check if `axis_name` isn't empty, or does `psum` translation handle that?

jax/_src/lax/parallel.py
```diff
@@ -749,6 +756,70 @@ def _process_axis_index(self, frame):
...
+      c, local_out, axis_name=axis_name, axis_index_groups=None,
+      axis_env=axis_env, platform=platform)
+  out, = xla.xla_destructure(c, out_tup)
+  return out
+xla.parallel_translations[pdot_p] = _pdot_translation_rule
+
+def _pdot_transpose_lhs(g, y, *, axis_name, pos_contract, pos_batch):
+  # TODO: avals with names, call pbroadcast with axis_name
```
apaszke, line 813: I'm pretty sure that we need to assert that there were no `axis_name`s left. I don't think that our current AD system guarantees that a cotangent of an unmapped aval is also unmapped, so this would be incorrect.

apaszke (approved):
LGTM, except for the AD formulas (that might be correct, but I'm pretty sure that are broken).

jax/experimental/general_map.py
```diff
@@ -581,3 +586,27 @@ def subst_eqn_axis_names(eqn, axis_subst: Dict[AxisName, Tuple[AxisName]]):
...
+  constvars = map(trace.getvar, map(trace.instantiate_const, consts))
+  outvars = map(trace.makevar, out_tracers)
+  new_in_axes = (None,) * len(consts) + params['in_axes']
+  new_params = dict(params, in_axes=new_in_axes,
+                    call_jaxpr=pe.convert_constvars_jaxpr(jaxpr))
+  eqn = pe.new_jaxpr_eqn([*constvars, *invars], outvars, prim,
+                         new_params, source_info)
+  trace.frame.eqns.append(eqn)
+  return out_tracers
+pe.DynamicJaxprTrace.process_xmap = _dynamic_jaxpr_trace_process_xmap  # type: ignore
```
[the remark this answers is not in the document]
mattjj: I'll remove this then!

tests/batching_test.py
```diff
@@ -1111,6 +1111,69 @@ def testAxisIndex(self):
...
+
+    rng = np.random.RandomState(1)
+    x = rng.randn(3, 4)
+    y = rng.randn(4, 5)
+    z_bar = rng.randn(3, 5)
+
+    x_bar, y_bar = vmap(lambda x, y, z_bar: vjp(f, x, y)[1](z_bar),
+                        axis_name='i', in_axes=(1, 0, None), out_axes=(1, 0))(x, y, z_bar)
+    self.assertAllClose(x_bar, jnp.dot(z_bar, y.T))
+    self.assertAllClose(y_bar, jnp.dot(x.T, z_bar))
```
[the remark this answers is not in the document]
mattjj: I'm not sure how to call check_grads here, since we need to do a vmap-of-vjp to get the right callable, whereas check_grads applies vjp itself (and not vmap). If we call vmap ourselves to turn this into a positional-axis function, we could use check_grads to check vjp-of-vmap, but that wouldn't be testing the transpose rule of pdot.

WDYT? Does that sound right, or am I missing something?

jax/_src/lax/parallel.py
```diff
@@ -749,6 +756,70 @@ def _process_axis_index(self, frame):
...
+  return lax.dot_general_p.abstract_eval(
+      x, y, dimension_numbers=[pos_contract, pos_batch],
+      precision=None)
+
+def _pdot_vmap_collective_rule(frame, vals_in, dims_in, *, axis_name,
+                               pos_contract, pos_batch):
+  x, y = vals_in
+  x_dim, y_dim = dims_in
+  x_pos_contract, y_pos_contract = pos_contract
+  x_pos_contract = [x_dim] + [d + (d >= x_dim) for d in x_pos_contract]
```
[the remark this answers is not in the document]
mattjj: I think the order must agree with the order of `y_pos_contract`, as the elements of these two lists together tell us how to match axes up (i.e. which x axes to contract against which y axes). So I think it'd better to think of this as a list. An alternative API would be to have a set of pairs, but this API was chosen to match that of dot_general, which in term matches XLA HLO's DotGeneral.

jax/_src/lax/parallel.py
mattjj: I think `assert axis_name` a few lines above does that. WDYT?

jax/_src/lax/parallel.py
mattjj: Hmm not sure I follow. Can you give an example of the error you're worried about? Also, what do you mean exactly by checking that there were no axis_names left?

tests/gmap_test.py
```diff
@@ -256,6 +256,46 @@ def f(x):
...
+      return lax.pdot(x, y, 'i')
+
+    f_mapped = xmap(f, in_axes=[A({'i': 1}), A({'i': 0})], out_axes=A(),
+                    schedule=[('i', 'r1'), ('i', 'vectorize')])
+
+    rng = np.random.RandomState(0)
+    x = rng.randn(3, 8)
+    y = rng.randn(8, 5)
+
+    with fake_resources(r1=2):
```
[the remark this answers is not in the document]
mattjj: Done!

jax/_src/lax/parallel.py
mattjj: Since this code is moving fast (with both of us editing it), I'd like to get this change in to avoid rebasing. Moreover it might be easier to think through the AD rules when we can talk with lower latency, i.e. when we can video chat. For those reasons, and because I think I addressed the other review comments, I'm going to consider this merge-able, but I'll be happy to land fixes as we find need for them!

PR #5025, opened 2020-11-27
Add an experimental xmap lowering via the SPMD partitioner

apaszke:
As suggested by @jekbradbury. The SPMD partitioner might allow us to
avoid writing partitioning formulas on the JAX level and can result in
better performance by giving the compiler more flexibility.


mattjj (approved):
Woo!

Issue #3348, opened 2020-06-06
memory leak with grad and lax.scan


John-Boik:
I coded a GRU for time series analysis that uses lax.scan to scan through each time step. It appears that lax.scan when called by grad results in a GPU memory leak. This issue might be related to [282](https://github.com/google/jax/issues/282), but I am not sure how to fix it. As I understand it, a recurrent network built with Jax implies use of lax.scan (to avoid a for loop). Any ideas would be appreciated.  I tried clearing the xla cache, to no avail, and tried using remat, which reduced the memory leak only slightly (both efforts are commented out in the code below).

Simplified code to reproduce the problem is below. Each iteration of a call to grad produces a 370 MB memory leak.  I use cuda 10.1 with driver 435.21, Ununtu 19.10, Python 3.7, and Jax 0.1.62.
```

import time
import os
import jax.numpy as np

from functools import partial
from jax.experimental.stax import serial
from jax import remat
from jax import lax
from jax import random
from jax.nn.initializers import glorot_normal, normal
from jax.experimental import optimizers
from jax import grad
from jax.interpreters import xla


os.environ['XLA_PYTHON_CLIENT_PREALLOCATE']='false'
os.environ['XLA_PYTHON_CLIENT_ALLOCATOR'] = 'platform'

# --------------------------------------------------------------------------------------------------
# Gru
# --------------------------------------------------------------------------------------------------
def GRU(num_latents, W_init=glorot_normal(), b_init=normal()):
    # ----------------------------------------------------------------------------------------------
    def init_fun(rng, input_shape, **kwargs):
        k1, k2 = random.split(rng, num=2)
        p1 = b_init(k1, (input_shape[0], num_latents))  
        p2 = b_init(k2, (input_shape[0], num_latents))  
        return ((p1,p2),(None,))

        
    # ----------------------------------------------------------------------------------------------
    def apply_fun(params, inputs, **kwargs):
        p1 = params[0]  # general parameters
        p2 = params[1]  # the hidden state that evolves over time
        
        # -------------------------------
        def apply_fun_scan(p1_, p2_, inputs_):
            output = inputs_ * p1_ * p2_
            return output, output
        
        #fx = remat (partial(apply_fun_scan, p1), concrete=True)
        fx = partial(apply_fun_scan, p1)
        _, ht_new = lax.scan(fx, p2, inputs)  
        return ht_new
        
    return init_fun, apply_fun


rng = random.PRNGKey(1)
net_init, apply = serial(
[...]

Issue #3158, opened 2020-05-20
Jitting and memory leaks


xwinxu:
I am using a computing cluster to run my jobs. I notice that Slurm keeps killing my jobs within seconds of submission due to out of memory errors, which I believe is because JAX is asking for more memory than allocated since GPU usage seems to increase, potentially due to a memory leak. I am running `cuda10.1`.

This is the stack trace I get:
```
RuntimeError: Internal: Unable to launch cuBLAS gemm on stream 0x562d67df53e0
2020-05-18 19:33:00.557492: E external/org_tensorflow/tensorflow/stream_executor/cuda/cuda_driver.cc:940] could not synchronize on CUDA context: CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered :: *** Begin stack trace ***
        _PyDict_DelItem_KnownHash
        _PyEval_EvalFrameDefault
        _PyEval_EvalCodeWithName
        _PyFunction_FastCallDict
        PyObject_CallFunctionObjArgs
        PyObject_ClearWeakRefs
        PyDict_SetItem
        PyDict_SetItemString
        PyImport_Cleanup
        Py_FinalizeEx
        _Py_UnixMain
        __libc_start_main
*** End stack trace ***
2020-05-18 19:33:00.557560: F external/org_tensorflow/tensorflow/compiler/xla/service/gpu/gpu_executable.cc:88] Check failed: pair.first->SynchronizeAllActivity() 
slurmstepd: error: *** JOB 505887 ON gpu117 CANCELLED AT 2020-05-18T19:33:01 ***
```
I also get this sometimes as in this [issue](https://github.com/google/jax/issues/1491):
```
 File "...py", line 126, in apply_fun
    outputs = sdeint(augmented_drift, augmented_diffusion, aug_init(inputs), np.linspace(0, 1, 100), rng)
  File "...py", line 49, in sdeint
    _, states = lax.scan(f_scan, dict(curr_state=s_init), xs)
  File "...anaconda/lib/python3.7/site-packages/jax/lax/lax_control_flow.py", line 955, in scan
    linear=(False,) * (len(consts) + len(in_flat)))
  File "...anaconda/lib/python3.7/site-packages/jax/lax/lax_control_flow.py", line 1337, in scan_bind
    num_carry=num_carry, linear=linear)
  File ".../anaconda/lib/python3.7/site-packages/jax/core.py", line 214, in bind
    out_tracer = top_trace.process_primitive(self, tracers, kwargs)
[...]

PR #5029, opened 2020-11-27
Add note about lax.scan memory leaks

gmittal:


gmittal:
Couldn't replicate the issue in a simple script, though I have had similar problems as #3158 #3348 related to `lax.scan` and `jit` in my own code. Will close for now, but may re-open if I end up finding a more effective way to reproduce the issue.

PR #5030, opened 2020-11-27
clean up symbolic zero handling in transpose rules

mattjj:
We had a few transpose rules in which:
1. didn't correctly handle symbolic zero cotangent inputs, and/or
2. returned the ad_util.Zero class rather than an instance of it (and/or returned a singleton where they should return a list)

This solves a FIXME in the code left by @apaszke (because he first spotted these issues' effects on ad.backward_pass).


jekbradbury:
This was rolled back due to internal test failures; you'll need to resubmit.

mattjj:
That's what I get for not adding tests myself! I was being lazy and expecting a rollback :)

mattjj:
Thanks for the heads up on this @jekbradbury! I'll do a global presubmit check on the next one.

PR #5032, opened 2020-11-27
Transpose rule for imag

### LINKED ISSUE #979: Inconsistent convention for complex derivatives
The convention for the handling of complex derivatives in JAX seems to be mixed. Consider the following two implementations of the abolute value function.
```py
import jax.numpy as np
from jax import grad

def f(x):
    return np.abs(x)

def g(x):
    return np.sqrt(np.real(x)**2 + np.imag(x)**2)

df = grad(f)
dg = grad(g)

print('gradient by df: ',df(1.0+0.5j))
print('gradient by dg: ',dg(1.0+0.5j))
```
The respective results differ by complex conjugation:
```
gradient by df:  (0.8944272-0.4472136j)
gradient by dg:  (0.8944272+0.4472136j)
```

inailuig:
This PR is an attempt to bring back a corrected version of the transpose rule for imag.

Back in https://github.com/google/jax/pull/981 the (wrong) linear transpose rule for imag was replaced with a (correct) jvp rule in order to fix https://github.com/google/jax/issues/979. 
However ommitting a transpose rule also broke linear_transpose for imag:

```
import jax
import types
import numpy as np
complex_scalar = types.SimpleNamespace(shape=(), dtype=np.complex64)

def f(x):
    return (-1j*x).real

def g(x):
    return x.imag

print(jax.linear_transpose(f, complex_scalar)(1.))
print(jax.linear_transpose(g, complex_scalar)(1.))
```
before:
```
(DeviceArray(0.-1.j, dtype=complex64),)

...
NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'imag' not implemented
```
after the changes in this PR:
```
(DeviceArray(0.-1.j, dtype=complex64),)
(DeviceArray(0.-1.j, dtype=complex64),)
```

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2306,6 +2306,7 @@ real_p = unop(_complex_basetype, _complex, 'real')
 ad.deflinear(real_p, lambda t: [complex(t, np.zeros((), _dtype(t)))])
 
 imag_p = unop(_complex_basetype, _complex, 'imag')
+ad.deflinear(imag_p, lambda t: [complex(np.zeros((), _dtype(t)), mul(_const(t, -1), t))])
 ad.defjvp(imag_p, lambda g, _: real(mul(_const(g, -1j), g)))
 
 _complex_dtype = lambda dtype, *args: (np.zeros((), dtype) + np.zeros((), np.complex64)).dtype
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2306,7 +2306,7 @@ real_p = unop(_complex_basetype, _complex, 'real')
 ad.deflinear(real_p, lambda t: [complex(t, np.zeros((), _dtype(t)))])
 
 imag_p = unop(_complex_basetype, _complex, 'imag')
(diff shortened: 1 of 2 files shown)


shoyer:
Could you kindly add a new unit test for one of the cases this fixes? Perhaps the example you wrote for the PR description? :)

inailuig:
The only existing tests for linear transpose rules are in `api_test.py`.
Just before `test_linear_transpose_complex` the following could be added:

```
  def test_linear_transpose_imag(self):
    f = lambda x: x.imag
    transpose = api.linear_transpose(f, 1.j)
    actual, = transpose(1.)
    expected = -1.j
    self.assertEqual(actual, expected)
```
however I am not sure if this is what we want (essentialy only checking for the definition) and if `api_test.py` would be the right place?

The jvp rule which is automatically added by `deflinear` (which is exactly the same as explicitly defined before) is already being tested in `lax_autodiff_test.py`.

shoyer:
This feels like a more operation specific test, which suggests that it would make sense to add it to `lax_autodiff_test.py`. I would probably just add the test you wrote to the end of that file as a regression test.

inailuig:
For completeness I also added the same test for `real`

Issue #4322, opened 2020-09-17
Allow custom_linear_solve to return things besides the solution


alewis:
If I'm reading correctly `lax.custom_linear_solve` can presently only wrap `_solve` functions that return the computed solution `x` and nothing else. Particularly since in-place modification of arguments is of course not possible, this is a pretty severe limitation:

- Most solvers also return some kind of diagnostic information that the user will definitely want.
- Some solvers (e.g. LGMRES) additionally return auxiliary data that can be used to accelerate the solution of similar problems (the SciPy version of LGMRES does this with an in-place modification). This is quite useful in the fairly common case that the solve is one step in a larger iterative algorithm.

Is there a way around this?

shoyer:
I agree, this does seem like an unfortunate limitation of the current implementation. My original hope was to solve this with solve general system for side effects in JAX code, but that doesn't seem to be forthcoming.

In the meantime, I think a good solution would be to add an optional `has_aux` argument to `custom_linear_solve`, which could work like `has_aux` on `jax.grad`: if `hax_aux` is True, then the user provided `solve` function (and maybe `transpose_solve`, too?) should return a tuple of the solution and any auxiliary output instead.

Note also that `jax.experimental.host_callback` is a good solution for the diagnostic information use case (especially for solves that appear in the backwards pass).

PR #5034, opened 2020-11-28
GMRES cleanup/fixup

shoyer:
1. Switched the interface for picking the solver from `qr_mode=True`/`False` to `solve_method='incremental'`/`'direct'`. I'm open to adjustments here, but I don't think `qr_mode=True` was a good way to describe these algorithm since the "direct" method could also be calculated via a QR decomposition, at least in principle.
2. Fixed a bug in the "direct" calculation.
3. Fixed handling of complex numbers in the "iterative" calculation, by copying the handling of complex numbers in the Givens rotation from SciPy. I also updated the numerics for the Givens rotation to match SciPy, which I guess should be more numerically stable, since it avoids calculating `sqrt(a**2 + b**2)` in favor of calculations like `a * sqrt(1 + (b/a)**2` (xref https://github.com/google/jax/issues/4882).

With these changes, `gmres` passes the full test suite for compatibility with SciPy! If we're happy with the interface, we can do ahead and expose it publicly as `jax.scipy.sparse.linalg.gmres`.

### DIFF
--- a/jax/_src/scipy/sparse/linalg.py
+++ b/jax/_src/scipy/sparse/linalg.py
@@ -40,10 +40,9 @@ def _vdot_real_part(x, y):
   #  where M is positive definite and Hermitian, so the result is
   # real valued:
   # https://en.wikipedia.org/wiki/Definiteness_of_a_matrix#Definitions_for_complex_matrices
-  vdot = partial(jnp.vdot, precision=lax.Precision.HIGHEST)
-  result = vdot(x.real, y.real)
+  result = _vdot(x.real, y.real)
   if jnp.iscomplexobj(x) or jnp.iscomplexobj(y):
-    result += vdot(x.imag, y.imag)
+    result += _vdot(x.imag, y.imag)
   return result
 
 
@@ -51,12 +50,9 @@ def _vdot_real_tree(x, y):
   return sum(tree_leaves(tree_multimap(_vdot_real_part, x, y)))
 
 
-def _norm_tree(x):
-  return jnp.sqrt(_vdot_real_tree(x, x))
-
-
-def _vdot_tree(x, y):
-  return sum(tree_leaves(tree_multimap(_vdot, x, y)))
+def _norm(x):
+  xs = tree_leaves(x)
+  return jnp.sqrt(sum(map(_vdot_real_part, xs, xs)))
 
 
 def _mul(scalar, tree):
@@ -211,7 +207,7 @@ def _safe_normalize(x, thresh=None):
   which by default is the machine precision of x's dtype, it will be
   taken to be 0, and the normalized x to be the zero vector.
   """
-  norm = _norm_tree(x)
+  norm = _norm(x)
   dtype = jnp.result_type(*tree_leaves(x))
   if thresh is None:
     thresh = jnp.finfo(norm.dtype).eps
@@ -233,7 +229,7 @@ def _project_on_columns(A, v):
   return tree_reduce(operator.add, v_proj)
 
 
-def _iterative_classical_gram_schmidt(Q, x, max_iterations=2):
+def _iterative_classical_gram_schmidt(Q, x, xnorm, max_iterations=2):
   """
   Orthogonalize x against the columns of Q. The process is repeated
   up to `max_iterations` times, or fewer if the condition
@@ -247,6 +243,8 @@ def _iterative_classical_gram_schmidt(Q, x, max_iterations=2):
   x : array or tree of arrays
       A vector. It will be replaced with a new vector q which is orthonormal
       to the columns of Q, such that x in span(col(Q), q).
+  xnorm : float
+      Norm of x.
 
   Returns
   -------
@@ -259,17 +257,18 @@ def _iterative_classical_gram_schmidt(Q, x, max_iterations=2):
   # "twice is enough"
   # http://slepc.upv.es/documentation/reports/str1.pdf
 
+  # TODO(shoyer): consider switching to only one iteration, like SciPy?
+
   # This assumes that Q's leaves all have the same dimension in the last
   # axis.
   r = jnp.zeros((tree_leaves(Q)[0].shape[-1]))
   q = x
-  _, xnorm = _safe_normalize(x)
   xnorm_scaled = xnorm / jnp.sqrt(2)
 
   def body_function(carry):
     k, q, r, qnorm_scaled = carry
     h = _project_on_columns(Q, q)
-    Qh = tree_map(lambda X: _dot_tree(X, h), Q)
+    Qh = tree_map(lambda X: _dot(X, h), Q)
     q = _sub(q, Qh)
     r = _add(r, h)
 
@@ -298,7 +297,7 @@ def _iterative_classical_gram_schmidt(Q, x, max_iterations=2):
   return q, r
 
 
-def _kth_arnoldi_iteration(k, A, M, V, H, tol):
+def _kth_arnoldi_iteration(k, A, M, V, H):
   """
   Performs a single (the k'th) step of the Arnoldi process. Thus,
   adds a new orthonormalized Krylov vector A(M(V[:, k])) to V[:, k+1],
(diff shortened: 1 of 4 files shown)


alewis:
Hey, this is great!

alewis:
So the difference between the two methods as I understood them was that in the "direct" method, we build the matrix of Krylov vectors, while in the "incremental" method, we build a QR decomposition of that matrix. I'm happy changing the name from qr_mode, but what motivates this specific choice?

shoyer:
> So the difference between the two methods as I understood them was that in the "direct" method, we build the matrix of Krylov vectors, while in the "incremental" method, we build a QR decomposition of that matrix. I'm happy changing the name from qr_mode, but what motivates this specific choice?

My thinking was:

- `solve_method="direct"` solves the least squares problem "directly", with a single solve of the Hessenberg matrix at the end of a full GMRES loop.
- `solve_method="incremental"` solves the least squares "incrementally", by updating the QR decomposition of the Hessenberg matrix in each step inside a GMRES loop with a Givens rotation.

My concern was that the single solve could also arguably use a QR decomposition, so I didn't find `qr_mode=True` to be descriptive.

shoyer:
@alewis @gehring if either of you are up for giving this PR a code review, that would be greatly appreciated!

gehring (approved):
LGTM

@alewis is probably in a better position to catch any sneaky problems that might be hiding since the code has changed a fair bit since I last dived into it, but all the changes seem good to me.

jax/_src/scipy/sparse/linalg.py
```diff
@@ -483,13 +484,13 @@ def body_fun(value):
 
   initialization = (x0, 0, unit_residual, residual_norm)
   x_final, k, _, err = lax.while_loop(cond_fun, body_fun, initialization)
-  _ = k # Until we can pass this out
+  _ = k  # Until we can pass this out
```
gehring, line 487: This is completely tangental to this PR but is there an open issue for tracking this missing feature?

shoyer: Yes, see https://github.com/google/jax/issues/4322

gehring: Great, thanks for the pointer!

alewis (approved):
I can't find anything to complain about here - good work! 

Do we want to test half precision?

shoyer:
> Do we want to test half precision?

I doubt this is worthwhile, at least not without some careful thought. Linear algebra with half precision is quite tricky. I would guess it requires specialized algorithms to work well.

shoyer:
Throwing out some other options for the parameter name:

1. `qr_mode=True` vs `qr_mode=False` (current implementation)
2. `solve_method='incremental'` vs `solve_method='direct'` (this PR)
3. `lstsq_method='incremental'` vs `lstsq_method='direct'`
4. `allow_early_termination=True` vs `allow_early_termination=False`
5. `incremental_lstsq=True` vs `incremental_lstsq=False`
6. `lstsq_update=True` vs `lstsq_update=False`

Any opinions?

shoyer:
By the way, this notebook shows some simple benchmarking results with the new code, where the linear operator is a dense matrix of different sizes:
https://gist.github.com/shoyer/146902a92016d3d936ad0b224910a967

```
size=20, restart=10
SciPy CPU:
1000 loops, best of 3: 321 µs per loop
JAX incremental CPU:
10000 loops, best of 3: 93.4 µs per loop
JAX direct CPU:
10000 loops, best of 3: 98.1 µs per loop
JAX incremental GPU:
100 loops, best of 3: 4.52 ms per loop
JAX direct GPU:
100 loops, best of 3: 1.88 ms per loop

size=200, restart=50
SciPy CPU:
100 loops, best of 3: 2.4 ms per loop
JAX incremental CPU:
1000 loops, best of 3: 691 µs per loop
JAX direct CPU:
1000 loops, best of 3: 954 µs per loop
JAX incremental GPU:
10 loops, best of 3: 63.1 ms per loop
JAX direct GPU:
100 loops, best of 3: 8.38 ms per loop

size=2000, restart=200
SciPy CPU:
1 loop, best of 3: 329 ms per loop
JAX incremental CPU:
1 loop, best of 3: 362 ms per loop
JAX direct CPU:
1 loop, best of 3: 370 ms per loop
JAX incremental GPU:
1 loop, best of 3: 884 ms per loop
JAX direct GPU:
10 loops, best of 3: 79.7 ms per loop
```

To summarize:

- The "incremental" approach dominates on CPU, whereas "direct" dominates on GPU. I still need to test TPUs.
- JAX CPU can be considerably faster than SciPy on _small_ problems (which makes sense).
- JAX GPU becomes attractive for very large matrices / expensive linear operators.

gehring:
> Throwing out some other options for the parameter name:
> 
> 1. qr_mode=True vs qr_mode=False (current implementation)
> 2. solve_method='incremental' vs solve_method='direct' (this PR)
> 3. lstsq_method='incremental' vs lstsq_method='direct'
> 4. allow_early_termination=True vs allow_early_termination=False
> 5. incremental_lstsq=True vs incremental_lstsq=False
> 6. lstsq_update=True vs lstsq_update=False
> 
> Any opinions?

I like the `solve_method` one best. Otherwise, I think it's a great idea to include the word "incremental" here so I would eliminate 1, 4, and 6 just because of that. If you have a rough idea of the strength and weaknesses of GPUs, it does a good job biasing you towards using the right one for the right backend.

While totally good enough, "direct" doesn't feel as informative to me. Maybe something along the line of "vectorized" or "batched" might really drive the point in. I think the priority is to have names that make it easy to remember which option to use when. While some of the other options might better convey the fundamental differences in the methods, if you're not all that familiar with GMRES or linear solvers in general, it won't really mean much to you. As long as users have an easy time remembering and they feel compelled to look it up in the documentation, then I think it's accomplished its goal. We can always dive into the subtleties in the docs.

This is probably way overthinking it at this point, but I'm committed now. :P

jax/_src/scipy/sparse/linalg.py
```diff
@@ -483,13 +484,13 @@ def body_fun(value):
...
   _ = err
   return x_final  # , info
 
 
-def _gmres(A, b, x0=None, *, tol=1e-5, atol=0.0, restart=20, maxiter=None,
-           M=None, qr_mode=False):
+def gmres(A, b, x0=None, *, tol=1e-5, atol=0.0, restart=20, maxiter=None,
+          M=None, solve_method='incremental'):
```
gehring, line 493: Given the benchmark results you just shared, would it maybe be better to use the "direct" method as a default? It seems like the direct method is close enough on CPU while significantly faster on GPU. Granted those were very synthetic benchmarks but I think most use cases won't be all that different. Also, I suspect that the users most likely to care are also the ones most likely to override the default behavior.

shoyer:
> While totally good enough, "direct" doesn't feel as informative to me. Maybe something along the line of "vectorized" or "batched" might really drive the point in

Yes, "batched" is definitely a better name than "direct"! I will make that switch.

@gehring One important caveat for my benchmarks is that by design they never terminate the inner GMRES loop early. If the algorithm converges in fewer than `restart` iterations (e.g., because you have a good pre-conditioner), the "incremental" method could be much faster. In contrast, the "batched" method always calculates calculates `restart` matrix-vector products at a time.

gehring:
Ah! I thought this was with termination. Then it's not as clear cut, I agree. I think it would depend on how big of a difference hat makes in more "realistic" settings. 

Though for the users with a "good pre-conditioner" point, those are the exactly the kind of users I'm less worried about. The kind of user that uses a good pre-conditioner is also likely to be the kind of user that knows there isn't a 'one size fits all' solution for solving those kind of problems. It'd expect them to be drawn to figuring out what this enigmatic `solve_method` argument does before even considering using it.

As a result, I'm admittedly a bit biased towards thinking about simpler use cases here, like the ones you might see from a more novice user coming from numpy who only knows `np.linalg.solve` or worse! They might have only ever used `inv(A) @ b` for their linear systems (\*\*gasp\*\*) and now feel a bit lost working with a `vjp` while they look around for a matrix-free solver. I would expect those type of users to be much more likely to just go with the default values for everything.

shoyer:
This may be a reasonable case for making the "batched" solver the default. With the other default parameters (`restart=20` and no preconditioner, which likely implies a large number of required outer iterations), I would expect it to usually have similar or better performance.

alewis:
I actually think both methods support early termination now

alewis:
Yeah, they do. The "direct" method terminates when the Arnoldi process start spitting out zero vectors, while the "incremental" one directly computes the error at each step.

shoyer:
> Yeah, they do. The "direct" method terminates when the Arnoldi process start spitting out zero vectors, while the "incremental" one directly computes the error at each step.

Right, but the "direct" method doesn't compute the residual norm except after a full GMRES loop. We _could_ compute the error estimate more often, but that would add a significant number of matrix/vector products.

My guess is that it is more common for GMRES to terminate due to the residual norm going below the tolerance than due to the Arnoldi process terminating early (which should only happen when the number of GMRES iterations is larger than the rank of the linear operator).

alewis:
Yeah, fair enough. In this case the direct method probably makes more sense as the default. A common use case is also when the linear operator is a sparse representation of some very large matrix, so that the cost of evaluating matvecs vastly overwhelms that of the linear solve etc; in this case the direct method may be favourable even on accelerators.

gehring:
Yeah, I don't think there is a "right" answer since we don't have any representative benchmarks so it might not be worth spending too much time pondering about this. This can definitely be revisited later once we've accumulated a few real-world examples of jax + matrix-free solver.

Personally, I'm leaning towards "batched" as default because of the noice user argument and also because I feel like it will have more predictable performance (computation-wise) but that's mostly a gut feeling. In the end, I feel like I'm guestimating more than anything else so if either of you have stronger feelings about this than me, I'm happy with either choice!

shoyer:
> in this case the direct method may be favourable even on accelerators due to the better termination criterion.

Here you are referring to how with the incremental approach our error estimate is the norm of the _preconditioned_ residual? It's a little hard to tell from the convoluted Fortran/Python in SciPy, but it does appear that SciPy checks the exact residual norm at every GMRES restart:
https://github.com/scipy/scipy/blob/12892b9a5c3be11686ff14dddf58b553717b3303/scipy/sparse/linalg/isolve/iterative.py#L634

Side note: we definitely aren't quite matching how scipy handles error tolerance in the "incremental" case (we don't do the "inner loop tolerance control" yet), which I guess is another small factor in favor of defaulting to "direct"/"batched"

shoyer:
I made some adjustments to the handling of tolerance to make things more similar to scipy, but there are still minor differences. I also adjusted the default solver_method to "batched".

Any other feedback on the PR? :)

gehring:
LGTM! Great job with the fixups!

shoyer:
thanks @gehring and @alewis for reviews here!

Issue #5009, opened 2020-11-24
Add support for `prepend` and `append` in jax.numpy.diff


jakevdp:
These arguments are currently not supported; see NumPy's description of the behavior here: https://numpy.org/devdocs/reference/generated/numpy.diff.html, and JAX's current implementation here: https://github.com/google/jax/blob/c9579b85584ea7439202d3eab37eb59bd9e3d0b2/jax/_src/numpy/lax_numpy.py#L1099-L1122

soumya997:
hey, @jakevdp I want to contribute to this, but I am fairly new to JAX. If you can share what are the prerequisites for this issue then it would be very helpful. 
if I am not wrong then By this issue, you want to implement `prepend` and `append` in jax.numpy.diff which is not available currently and this feature is similar to `prepend` and `append` of numpy.

jakevdp:
Hi @soumya997 – there is already an effort underway at the PR linked above (#5035)

soumya997:
Is there any other issue where I can contribute? @jakevdp

PR #5035, opened 2020-11-28
added prepend and append to diff

saklani:
Addresses #5009.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1115,7 +1115,7 @@ def angle(z):
 
 
 @_wraps(np.diff)
-def diff(a, n=1, axis=-1):
+def diff(a, n=1, axis=-1, prepend=None, append=None):
   _check_arraylike("diff", a)
   if n == 0:
     return a
@@ -1125,6 +1125,28 @@ def diff(a, n=1, axis=-1):
     raise ValueError(f"diff requires input that is at least one dimensional; got {a}")
 
   nd = a.ndim
+
+  combined = []
+  if prepend is not None:
+    _check_arraylike("diff", prepend)
+    if isscalar(prepend):
+      shape = list(a.shape)
+      shape[axis] = 1
+      prepend = broadcast_to(prepend, tuple(shape))
+    combined.append(prepend)
+
+  combined.append(a)
+
+  if append is not None:
+    _check_arraylike("diff", append)
+    if isscalar(append):
+      shape = list(a.shape)
+      shape[axis] = 1
+      append = broadcast_to(append, tuple(shape))
+    combined.append(append)
+
+  if len(combined) > 1:
+    a = concatenate(combined, axis)
 
   slice1 = [slice(None)] * nd
   slice2 = [slice(None)] * nd
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -296,7 +296,6 @@ JAX_COMPOUND_OP_RECORDS = [
               check_dtypes=False),
     op_record("true_divide", 2, all_dtypes, all_shapes, jtu.rand_nonzero,
               ["rev"], inexact=True),
-    op_record("diff", 1, number_dtypes + bool_dtypes, nonzerodim_shapes, jtu.rand_default, ["rev"]),
     op_record("ediff1d", 3, [np.int32], all_shapes, jtu.rand_default, []),
     op_record("unwrap", 1, float_dtypes, nonempty_nonscalar_array_shapes,
               jtu.rand_default, ["rev"],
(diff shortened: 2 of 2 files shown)


saklani:
@googlebot I signed it!

jax/_src/numpy/lax_numpy.py
```diff
@@ -1108,6 +1108,28 @@ def diff(a, n=1, axis=-1):
 
   nd = a.ndim
 
+  combined = []
+  if prepend is not None:
+    _check_arraylike(prepend)
```
jakevdp, line 1113: The signature here is `_check_arraylike("diff", prepend)`; same below.

jakevdp (commented):
Looks great! Just some minor changes below to the additional code.

The last thing that will be needed is tests. Currently the default version of `jnp.diff` is tested using automatically-generated tests here: https://github.com/google/jax/blob/dd9526bd398caf3d0b92236e219f1e2682b61953/tests/lax_numpy_test.py#L299

Testing the four optional parameters will involve replacing that with a parametrized test that exercises all of these options. You could model it off of other tests in the test suite; for example this one: https://github.com/google/jax/blob/dd9526bd398caf3d0b92236e219f1e2682b61953/tests/lax_numpy_test.py#L2289-L2304

tests/lax_numpy_test.py
```diff
@@ -2303,6 +2302,31 @@ def testFull(self, shape, fill_value_dtype, out_dtype, rng_factory):
...
+       for shape, dtype in _shape_and_dtypes(nonempty_nonscalar_array_shapes, default_dtypes)
+       for n in [0, 1, 2]
+       for axis in list(range(-len(shape), max(1, len(shape))))
+       for prepend in [None, 1, jtu._rand_dtype(np.random.rand, shape, dtype)]
+       for append in [None, 1, jtu._rand_dtype(np.random.rand, shape, dtype)]
+       ))
+  def testDiff(self, shape, dtype, n, axis, prepend, append, rng_factory):
+    rng = rng_factory(self.rng())
+    args_maker = lambda: [rng(shape, dtype)]
+    if prepend is None and append is None: np_fun = lambda x: np.diff(x, n=n, axis=axis)
```
saklani, line 2322: @jakevdp I couldn't figure out a better way to call np diff because it checks for ```np._NoValue ``` instead of ```None```. So passing ```None``` gives an error

jakevdp:
Getting so close!

To address the last test failure, you can add something like `check_dtypes=(dtype != np.float16)` to the `CheckAgainstNumpy` call.

jakevdp:
Also, before we merge it would be great if you could squash all of these commits into one and force-push it to this branch. If you're not comfortable doing that, I could either walk you through it, or take care of it for you!

saklani:
>Getting so close!
>
>To address the last test failure, you can add something like check_dtypes=(dtype != np.float16) to the CheckAgainstNumpy call.

Yes!

I've added the check. Quick query: Would it make sense completely to ignore all datatype checks i.e. setting `check_dtypes=False` in CheckAgainstNumpy for this function? It would simplify the code a bit.

> Also, before we merge it would be great if you could squash all of these commits into one and force-push it to this branch. If you're not comfortable doing that, I could either walk you through it, or take care of it for you!
Sure. I can do it. Just to be clear, you mean rebase->force-push, correct?

jakevdp:
> Sure. I can do it. Just to be clear, you mean rebase->force-push, correct?

I mean make it so that the pull request contains a single commit with all the changes.

jakevdp (approved):

jakevdp:
I'm going to run some additional tests on this (including GPU & TPU backends) and let you know if anything else comes up

tests/lax_numpy_test.py
```diff
@@ -2303,6 +2302,44 @@ def testFull(self, shape, fill_value_dtype, out_dtype, rng_factory):
...
+      elif not np.isscalar(append) and append.dtype == jnp.bfloat16:
+        append = append.astype(np.float32)
+
+      if x.dtype == jnp.bfloat16:
+        return np.diff(x.astype(np.float32), n=n, axis=axis, prepend=prepend, append=append).astype(jnp.bfloat16)
+      else:
+        return np.diff(x, n=n, axis=axis, prepend=prepend, append=append)
+
+    jnp_fun = lambda x: jnp.diff(x, n=n, axis=axis, prepend=prepend, append=append)
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=(dtype != np.float16))
```
jakevdp, line 2340: Running a fuller set of tests revealed some more dtype issues here (for example, numpy tends to promote things aggressively to float64). The easiest thing would probably be to set `check_dtypes=False`

saklani: Ok. I've added `check_dtypes=False`.

saklani:
> > Sure. I can do it. Just to be clear, you mean rebase->force-push, correct?
> 
> I mean make it so that the pull request contains a single commit with all the changes.

Yeah, thats what I meant. Done.

saklani:
@jakevdp, thanks for the reviews. Unfortunately, it seems an internal test has failed, for which I don't have the view permission.

jakevdp:
Yep, I'm looking into that, and I'm not entirely sure what's going on. It might help for you to rebase this on master and force push - would you be able to do that?

saklani:
Yeah, I gave it a shot. Let's see what happens.

jakevdp:
Yeah, I'm not sure what's going on. The same errors are blocking #5060 as well.

jakevdp:
A sync & rebase ~an hour ago worked for #5060. Would you mind trying again?

saklani:
> A sync & rebase ~an hour ago worked for #5060. Would you mind trying again?

Thanks! It worked.

jakevdp:
Thanks for the contribution!

PR #5042, opened 2020-11-30
readme typo

redwrasse:
change 'multiplies' to 'multiplication' in readme


hawkinsp (approved):
I think it wasn't actually wrong before, but your version reads better!

PR #5046, opened 2020-11-30
Fix broken CI build due to NumPy 1.18 bug in full_like tests.

hawkinsp:


jakevdp (approved):

jakevdp:
Is this true for just numpy 1.18, or all numpy < 1.19? If so why did our numpy 1.16 tests pass? Did we just get lucky that `shape=()` was never run?

hawkinsp:
> Is this true for just numpy 1.18, or all numpy < 1.19? If so why did our numpy 1.16 tests pass? Did we just get lucky that `shape=()` was never run?

* the feature was added in NumPy 1.17 and you added a skipif line that skips the test for Numpy < 1.17.
* as far as I can tell, NumPy was buggy for all versions before 1.19.0.
* we have Github CI builds with an explicit 1.16.4 pin and without any pin at all. I assume the latter gives us the latest NumPy, presumably 1.19.something.
* Kokoro appears to be using 1.18 at the moment.

So Kokoro is simply testing a NumPy version which we don't otherwise cover, and it turned out that version was buggy.

jakevdp:
> the feature was added in NumPy 1.17 and you added a skipif line that skips the test for Numpy < 1.17.

Ah, I'd forgotten about that. Thanks.

PR #5051, opened 2020-11-30
Remove duplicate code to copy cusolver_kernels in wheel build.

hawkinsp:
May help with #4994


hawkinsp:
I submitted https://github.com/google/jax/pull/5053 instead of this PR (does the same thing, working around a CI problem).

PR #5055, opened 2020-11-30
Remove fake xmap resources, remove gmap

apaszke:
xmap can now handle real devices, so there's no point in maintaining the
simulated codepaths. Also, remove single-dimensional gmap as it will
have to be superseeded by a more xmap-friendly alternative.

### DIFF
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -30,7 +30,7 @@ import jax.numpy as jnp
 from jax import test_util as jtu
 from jax import vmap
 from jax import lax
-from jax.experimental.general_map import gmap, Mesh, mesh, xmap, A
+from jax.experimental.maps import Mesh, mesh, xmap, A
 from jax.lib import xla_bridge
 from jax.util import curry, unzip2
 from jax.interpreters import pxla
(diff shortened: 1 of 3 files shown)


tests/xmap_test.py
```diff
@@ -115,96 +90,14 @@ def new_f(*args, **kwargs):
...
-    b = jnp.arange(6*16).reshape((6, 16))
-    c, d = fm(a, b)
-    self.assertAllClose(c, (a + 2).transpose((1, 0, 2)))
-    self.assertAllClose(d, (b * 4).T)
-
-  @ignore_gmap_warning()
-  @skipIf(not config.omnistaging_enabled,
-          "vmap collectives only supported when omnistaging is enabled")
-  @with_mesh([('r1', 4), ('r2', 2), ('r3', 5)])
-  def testXMapCollectives(self):
```
apaszke, line 186: Note that those two tests were always skipped anyway, because none of our test environments has 40 devices available

mattjj (approved):

mattjj:
> xmap can now handle real devices, so there's no point in maintaining the simulated codepaths

Would it still be useful to have simulated devices so that even when you don't have physical devices available you can test executing your code (backed by vectorization but using the same syntax as if you used physical devices)? In addition to testing and exploring the scheduling API, I can imagine that being useful for demos, especially if we connected it to some kind of visualiziation. (If so, we can always add that kind of thing later, and adding it later might help us avoid friction now!)

apaszke:
Yeah I'm removing it mostly because I suspect that it might remove some friction in implementing new xmap features (e.g. it is quite annoying to have to handle those fake resources in the xmap translation rule). But on the other hand I am split about removing them too, mainly because the code necessary to handle sequential iteration might look similar to that.

For visualizations can't you just get XLA to simulate 500 CPU devices?

PR #5059, opened 2020-12-01
all_to_all: Add optional axis_index_groups argument

jpuigcerver:
Most XLA collective ops support a `replica_groups` argument, which Jax typically exposes as `axis_index_groups`. psum, pmean, all_gather, etc support this, but all_to_all doesn't even though XLA allows it as well (https://www.tensorflow.org/xla/operation_semantics#alltoall).

This PR adds this feature to jax.lax.all_to_all function. 

However, notice that this PR is not fully functional yet. Although the "forward" call to the function works as expected (check the new testAllToAllReplicaGroups test), when computing gradients Jax raises two mysterious exceptions:

- testGradOfAllToAllReplicaGroups:
```
[...]
  File "jax/util.py", line 35, in safe_map
    assert len(arg) == n, 'length mismatch: {}'.format(list(map(len, args)))
AssertionError: length mismatch: [1, 8]
```

- testGradOfPswapaxes:
```
[...]
  File "jax/interpreters/ad.py", line 180, in write_cotangent
    assert v.aval.strip_weak_type() == joined_aval, (prim, v.aval, ct_aval)
AssertionError: (all_to_all, ShapedArray(float32[1,8]), ShapedArray(float32[8]))
```

I say that they are "mysterious" because pswapaxes just calls all_to_all, so I don't understand why they are different. Also, pswaxes (in that test) uses axis_index_groups=None, which should yield the same results as before (and it does in the "forward" call).

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -233,14 +233,16 @@ def pshuffle(x, axis_name, perm):
   return ppermute(x, axis_name, list(zip(perm, range(len(perm)))))
 
 
-def pswapaxes(x, axis_name, axis):
+def pswapaxes(x, axis_name, axis, *, axis_index_groups=None):
   """Swap the pmapped axis ``axis_name`` with the unmapped axis ``axis``.
 
   If ``x`` is a pytree then the result is equivalent to mapping this function to
   each leaf in the tree.
 
-  The mapped axis size must be equal to the size of the unmapped axis; that is,
-  we must have ``lax.psum(1, axis_name) == x.shape[axis]``.
+  The group size of the mapped axis size must be equal to the size of the
+  unmapped axis; that is, we must have
+  ``lax.psum(1, axis_name, axis_index_groups=axis_index_groups) == x.shape[axis]``.
+  By default, when ``axis_index_groups=None``, this encompasses all the devices.
 
   This function is a special case of ``all_to_all`` where the pmapped axis of
   the input is placed at the position ``axis`` in the output. That is, it is
@@ -252,13 +254,17 @@ def pswapaxes(x, axis_name, axis):
       :func:`jax.pmap` documentation for more details).
     axis: int indicating the unmapped axis of ``x`` to map with the name
       ``axis_name``.
+    axis_index_groups: optional list of lists containing axis indices (e.g. for
+      an axis of size 4, [[0, 1], [2, 3]] would run pswapaxes over the first
+      two and last two replicas). Groups must cover all axis indices exactly
+      once, and all groups must be the same size.
 
   Returns:
     Array(s) with the same shape as ``x``.
   """
-  return all_to_all(x, axis_name, axis, axis)
+  return all_to_all(x, axis_name, axis, axis, axis_index_groups=axis_index_groups)
 
-def all_to_all(x, axis_name, split_axis, concat_axis):
+def all_to_all(x, axis_name, split_axis, concat_axis, *, axis_index_groups=None):
   """Materialize the mapped axis and map a different axis.
 
   If ``x`` is a pytree then the result is equivalent to mapping this function to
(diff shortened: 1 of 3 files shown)


jax/_src/lax/parallel.py
```diff
@@ -586,10 +599,16 @@ def _all_to_all_translation_rule(c, x, *, split_axis, concat_axis, axis_name,
...
 
-def _all_to_all_transpose_rule(cts, axis_name, split_axis, concat_axis):
-  return (all_to_all(cts, axis_name=axis_name, split_axis=concat_axis, concat_axis=split_axis),)
+def _all_to_all_transpose_rule(cts, axis_name, split_axis, concat_axis, axis_index_groups):
+  return all_to_all(
+      cts,
+      axis_name=axis_name,
+      split_axis=concat_axis,
+      concat_axis=split_axis,
+      axis_index_groups=axis_index_groups)
```
apaszke, line 608: Note that this used to return a tuple, but doesn't do that anymore. That might explain why you're seeing surprising failures in the grad tests.

jpuigcerver: I'm embarrassed by just a dumb mistake. I spent hours looking at my code without noticing it... (╯°□°）╯︵ ┻━┻

Good news is that everything works as expected now.

apaszke (commented):
Do you know what are the semantics of `all_to_all` in presence of `axis_index_groups`? I've been trying to find a good description, but can't find anything in the XLA docs. My main concern is that `axis_index_groups` usually makes the collective operate over some smaller subgroups of devices, but the expected shape of the input of `all_to_all` should match the number of those devices. 

However, your tests seems to be passing in values of shape equal to the total number of devices across all groups. This is a bit surprising to me, and I wonder if it should be our API. What are your thoughts? I don't fully understand the reference implementation in the test and would appreciate a more thorough description (might be worth extending the docstring with it as well!).

jax/_src/lax/parallel.py
apaszke: Well I'm glad it works now 😄

jpuigcerver:
I only have this: https://www.tensorflow.org/xla/operation_semantics#alltoall 
According to that doc, the dimension size along the split_axis has to be a multiple of split_count (i.e. the number of cores within a group). 

You are absolutely right that the API should allow dim_size = group_size. I will update this. The question is whether we want to allow "dim_size % group_size == 0", as XLA allows, or we want to restrict to "dim_size == group_size". To be honest, the former would avoid some nasty reshapes in my use case, but not sure if that's a legitimate reason.

apaszke:
Hmm, I kind of like the batched version to be explicit, but I do agree that eliminating reshapes is nice. In any case, I think that getting the `dim_size == group_size` case working (and tested) is definitely a requirement. cc @mattjj @jekbradbury for opinions on the batching behavior.

jpuigcerver:
I've updated the implementation to work with (and only with) split_axis_size = group_size, as you suggested.

jax/_src/lax/parallel.py
```diff
@@ -252,13 +252,17 @@ def pswapaxes(x, axis_name, axis):
       :func:`jax.pmap` documentation for more details).
     axis: int indicating the unmapped axis of ``x`` to map with the name
       ``axis_name``.
+    axis_index_groups: optional list of lists containing axis indices (e.g. for
```
apaszke, line 255: Can you please update the docstring above? The property that the axis size must be equal to the size of the mapped axis is not true. It should be equal to the group in which the collective is performed, which just happen to encompass all devices by default. I assume that the same change has to happen in the docstring of `all_to_all`

jax/_src/lax/parallel.py
```diff
@@ -288,13 +296,19 @@ def all_to_all(x, axis_name, split_axis, concat_axis):
     where ``axis_size`` is the size of the mapped axis named ``axis_name`` in
     the input ``x``, i.e. ``axis_size = lax.psum(1, axis_name)``.
   """
+  if axis_index_groups and split_axis != concat_axis:
+    raise ValueError("all_to_all only supports axis_index_groups if "
+                     "split_axis = concat_axis")
```
apaszke, line 301: It is 100% ok to have this for now, but I'm curious as to why did you have to add this restriction?

apaszke (approved):
LGTM, but please update the docstring!

jax/_src/lax/parallel.py
jpuigcerver: Done.

jax/_src/lax/parallel.py
jpuigcerver: It was actually a misconception from my side, it works perfectly with split_axis != concat_axis. I've removed this check and updated the test case.

jax/_src/lax/parallel.py
```diff
@@ -289,16 +293,13 @@ def all_to_all(x, axis_name, split_axis, concat_axis, *, axis_index_groups=None)
       once, and all groups must be the same size.
 
   Returns:
-    Array(s) with shape given by the expression::
```
apaszke, line 284: Oh this was not a typo. RST uses double colons to start code blocks 😕

apaszke (approved):

gnecula:
Please squash the commits before we merge them. Thanks!

apaszke:
I've squashed the commits, so no action necessary on your side @jpuigcerver!

jpuigcerver:
Thanks for your help!

Issue #5010, opened 2020-11-24
Add support for more padding modes in jax.numpy.pad


jakevdp:
Currently only a few of the available modes are implemented. See https://numpy.org/devdocs/reference/generated/numpy.pad.html#numpy.pad for details, and the current JAX implementation here: https://github.com/google/jax/blob/c9579b85584ea7439202d3eab37eb59bd9e3d0b2/jax/_src/numpy/lax_numpy.py#L2398-L2404

minoring:
Do we have to implement ‘empty’ and 'function' mode? 
Those are new in numpy version 1.17.

jakevdp:
I think it would make sense to implement these (with tests marked to skip in older numpy versions).

Note that it's not possible to create uninitialized data in JAX, but we could fill with zeros in this case, similar to how `jnp.empty` is implemented.

jakevdp:
Looking at it, the function version would have to be implemented differently from numpy, because numpy's version has the function to modify the array in-place, while JAX arrays are immutable.

minoring:
I see. Let me implement those :)
Thanks for the tips!

minoring:
By the way, unsupported keyword argumenets should raise an error.
I am not sure this is another issue, but I will handle it after implementing 'empty' and 'function' mode.
```python
>>> import numpy as np
>>> a = [1, 2, 3]
>>> np.pad(a, mode='maximum', pad_width=2, reflect_type='odd')
ValueError: unsupported keyword arguments for mode 'maximum': {'reflect_type'}

>>> import jax.numpy as jnp
>>> a = [1, 2, 3]
>>> jnp.pad(a, mode='maximum', pad_width=2, reflect_type='odd')
DeviceArray([3, 3, 1, 2, 3, 3, 3], dtype=int32)
```

minoring:
I believe `jax.numpy.pad` support every padding modes now :)

jakevdp:
Thanks for the work on this!

PR #5060, opened 2020-12-01
Implement statistic pad mode in jax.numpy.pad

minoring:
Implemented maximum, mean, median, and minimum padding modes.

Related #5010

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2347,35 +2347,79 @@ def _pad_edge(array, pad_width):
   return array
 
 
-@partial(jit, static_argnums=(1, 2))
-def _pad(array, pad_width, mode, constant_values):
+def _pad_stats(array, pad_width, stat_length, stat_func):
+  nd = ndim(array)
+  for i in range(nd):
+    if stat_length is None:
+      stat_before = stat_func(array, axis=i, keepdims=True)
+      stat_after = stat_before
+    else:
+      array_length = array.shape[i]
+      length_before, length_after = stat_length[i]
+      if length_before == 0 or length_after == 0:
+        raise ValueError("stat_length of 0 yields no value for padding")
+
+      # Limit stat_length to length of array.
+      length_before = _min(length_before, array_length)
+      length_after = _min(length_after, array_length)
+
+      slice_before = lax.slice_in_dim(array, 0, length_before, axis=i)
+      slice_after = lax.slice_in_dim(array, -length_after, None, axis=i)
+      stat_before = stat_func(slice_before, axis=i, keepdims=True)
+      stat_after = stat_func(slice_after, axis=i, keepdims=True)
+
+    if np.issubdtype(array.dtype, np.integer):
+      stat_before = round(stat_before)
+      stat_after = round(stat_after)
+
+    stat_before = stat_before.astype(array.dtype)
+    stat_after = stat_after.astype(array.dtype)
+
+    npad_before, npad_after = pad_width[i]
+    pad_before = repeat(stat_before, npad_before, axis=i)
+    pad_after = repeat(stat_after, npad_after, axis=i)
+
+    array = lax.concatenate([pad_before, array, pad_after], dimension=i)
+  return array
+
+
+def _broadcast_to_pairs(nvals, nd, name):
+  nvals_shape = np.shape(nvals)
+  if nvals_shape == (nd, 2):
+    # ((before_1, after_1), ..., (before_N, after_N))
+    pass
+  elif nvals_shape == (1, 2):
+    # ((before, after),)
+    nvals = nvals * nd
+  elif nvals_shape == (2,):
+    # (before, after)  (not in the numpy docstring but works anyway)
+    before, after = nvals
+    nvals = (nvals,) * nd
+  elif nvals_shape == (1,):
+    # (pad,)
+    nvals, = nvals
+    nvals = ((nvals, nvals),) * nd
+  elif nvals_shape == ():
+    # pad
+    nvals = ((nvals, nvals),) * nd
+  else:
+    raise ValueError(f"{name} given unexpected structure: {nvals}. "
+                     "See docstring for valid {name} formats.")
+  return nvals
+
+
+@partial(jit, static_argnums=(1, 2, 4))
+def _pad(array, pad_width, mode, constant_values, stat_length):
   array = asarray(array)
   nd = ndim(array)
 
   if nd == 0:
     return array
 
-  pad_width_shape = np.shape(pad_width)
-  if pad_width_shape == (nd, 2):
-    # ((before_1, after_1), ..., (before_N, after_N))
-    pass
-  elif pad_width_shape == (1, 2):
-    # ((before, after),)
-    pad_width = pad_width * nd
-  elif pad_width_shape == (2,):
-    # (before, after)  (not in the numpy docstring but works anyway)
-    before, after = pad_width
-    pad_width = (pad_width,) * nd
-  elif pad_width_shape == (1,):
-    # (pad,)
-    pad_width, = pad_width
-    pad_width = ((pad_width, pad_width),) * nd
-  elif pad_width_shape == ():
-    # pad
-    pad_width = ((pad_width, pad_width),) * nd
-  else:
-    raise ValueError(f"pad_width given unexpected structure: {pad_width}. "
-                     "See docstring for valid pad_width formats.")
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2342,6 +2342,23 @@ def _pad_edge(array, pad_width):
...
 
 
+def _pad_stats(array, pad_width, stat_func):
+  nd = ndim(array)
+  for i in range(nd):
+    npad_before, npad_after = pad_width[i]
+
+    padded_values = stat_func(array, axis=i, keepdims=True)
+    if np.issubdtype(array.dtype, np.integer):
+      padded_values = jax.numpy.round(padded_values)
```
jakevdp, line 2352: `round` is defined in this file, so you can drop `jax.numpy` here.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2350,6 +2367,9 @@ def _pad(array, pad_width, mode, constant_values):
   if nd == 0:
     return array
 
+  stat_funcs = {"maximum": jax.numpy.amax, "minimum": jax.numpy.amin,
+                "mean": jax.numpy.mean, "median": jax.numpy.median}
```
jakevdp, line 2371: Same with these: all these aggregations are defined directly in this file so you can drop `jax.numpy`.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2342,6 +2342,23 @@ def _pad_edge(array, pad_width):
   return array
 
 
+def _pad_stats(array, pad_width, stat_func):
```
jakevdp, line 2345: Have you thought about handling the optional `stat_length` argument? If you'd prefer, feel free to leave it as a TODO.

tests/lax_numpy_test.py
```diff
@@ -1306,6 +1306,10 @@ def testOperatorRound(self):
           ('reflect', nonempty_shapes),
           ('wrap', nonempty_shapes),
           ('edge', nonempty_shapes),
+          ('maximum', nonempty_shapes),
+          ('minimum', nonempty_shapes),
+          ('mean', nonempty_shapes),
+          ('median', nonempty_shapes),
```
jakevdp, line 1312: I'd lean toward splitting these into separate tests, because `constant_values` has no effect on these modes, and so this parametrization is generating ~10 identical tests for each. This would also allow separate parametrization over the `stat_length` argument, if/when it is implemented.

jakevdp (changes requested):
Looks great! Thanks for tackling this. A few comments below.

minoring:
@jakevdp Thank you for your kind review.
I updated PR. Please take a look :)

jax/_src/numpy/lax_numpy.py
```diff
@@ -2342,35 +2342,82 @@ def _pad_edge(array, pad_width):
...
+      array_length = array.shape[i]
+      length_before, length_after = stat_length[i]
+      if length_before == 0 or length_after == 0:
+        raise ValueError("stat_length of 0 yields no value for padding")
+
+      # Limit stat_length to length of array.
+      if length_before > array_length:
+        length_before = array_length
+      if length_after > array_length:
+        length_after = array_length
```
jakevdp, line 2361: nit: maybe use `length_before = mlin(length_before, array_length)`

jax/_src/numpy/lax_numpy.py
```diff
@@ -2342,35 +2342,82 @@ def _pad_edge(array, pad_width):
...
+
+      slice_before = lax.slice_in_dim(array, 0, length_before, axis=i)
+      slice_after = lax.slice_in_dim(array, array_length - length_after,
+                                     array_length, axis=i)
```
jakevdp, line 2365: More concise, I think this works: `slice_after = lax.slice_in_dim(array, -length_after, None, axis=i)`

tests/lax_numpy_test.py
```diff
@@ -1351,6 +1351,54 @@ def testPad(self, shape, dtype, mode, pad_width, constant_values, rng_factory):
...
+          jtu.format_shape_dtype_string(shape, dtype), mode, pad_width, stat_length),
+       "shape": shape, "dtype": dtype, "mode": mode, "pad_width": pad_width,
+       "stat_length": stat_length, "rng_factory": jtu.rand_default}
+      for mode, shapes in [
+          ('maximum', nonempty_shapes),
+          ('minimum', nonempty_shapes),
+          ('mean', nonempty_shapes),
+          ('median', nonempty_shapes),
+      ]
+      for shape, dtype in _shape_and_dtypes(shapes, all_dtypes)
```
jakevdp, line 1365: since all shapes here are the same, you could simplify this and pass `nonempty_shapes` directly here.

tests/lax_numpy_test.py
```diff
@@ -1351,6 +1351,54 @@ def testPad(self, shape, dtype, mode, pad_width, constant_values, rng_factory):
...
+          # ((before, after),)
+          ((1, 2),), ((2, 0),),
+          # (before, after)  (not in the docstring but works in numpy)
+          (2, 0), (0, 0),
+          # (pad,)
+          (1,), (2,),
+          # pad
+          0, 1,
+      ]
+      for stat_length in [
```
jakevdp, line 1378: We should probably test `None` as well.

tests/lax_numpy_test.py
```diff
@@ -1351,6 +1351,54 @@ def testPad(self, shape, dtype, mode, pad_width, constant_values, rng_factory):
...
+          # (before, after)  (not in the docstring but works in numpy)
+          (1, 1), (3, 4),
+          # (pad,)
+          (1,), (2,),
+          # pad
+          1, 2
+      ]
+      if pad_width != ()))
+  def testPadStatValues(self, shape, dtype, mode, pad_width, stat_length, rng_factory):
+    rng = rng_factory(self.rng())
```
jakevdp, line 1392: Minor: recently, we've started using `jtu.rand_default()` directly to avoid having to add `rng_factory` to all the parametrization boilerplate.

jakevdp (changes requested):
Great progress! Nice work on the updated test - it has great coverage of all the various valid inputs.

A few mostly minor comments to address, and then we can get this merged!

minoring:
Nice tips. Thank you!
I made some improvement. PTAL :)

jax/_src/numpy/lax_numpy.py
minoring: I tried `length_before = min(length_before, array_length)` and got an error.
I suspect `jax.numpy.min` shadows python function `min`

jakevdp: Sorry, yes you need to use `_min()` here to get the builtin `min()`.

jakevdp (approved):
Looks great, thanks! I'm going to run this through a fuller set of tests (including GPU & TPU) and see if we can get this merged.

jakevdp:
OK, so there are a number of issues that came up in more comprehensive testing:

- passing `stat_length=()` for mode in (median, min, max) results in `ValueError: zero-size array to reduction operation minimum which has no identity`. It might be that an error is expected in this case: I'm not sure. But we need to shore-up the tests for this either way.
- with `dtype=bool`, `stat_length=2` and `mode='mean'`, there are differences on TPU (possibly because of differences in how `0.5` is converted to boolean, but I'm not certain). I think it would be fine to leave boolean dtypes out of these tests... computing means of booleans is not particularly well defined anyway.

Would you mind taking a look at those? Thanks!

minoring:
- Removed `stat_length=()` from tests since numpy gave me the same error message.
- Removed calculating mean of booleans. I considered leaving whole boolean type out of tests, but computing others might be OK since boolean can act like integer. https://docs.python.org/3/library/stdtypes.html#boolean-values

jax/_src/numpy/lax_numpy.py
minoring: Oh I see! updated ;)

jakevdp:
OK, I'll check the tests again.

Last thing: can you please squash all the commits to a single commit and force-push to this branch? If that's not something you're comfortable doing, I can walk you through it or take care of it for you. Thanks!

minoring:
Squashed it into single commit. 👍

jakevdp:
Thanks! Tests look good – I'll work on getting this merged!

jakevdp:
We have some internal test failures that I believe would be remedied by syncing to master and force-pushing to this branch. Are you comfortable doing that?

minoring:
Syncing to master worked! Thanks.

PR #5064, opened 2020-12-01
Fix pmap compilation cache regressions from #4904.

apaszke:
AD didn't use `HashableFunction` enough, tripping up the compilation cache.
I've also used the occasion to make the hashing a little safer by tagging
each place that `HashableFunction` with a unique tag in its key.

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -1516,11 +1516,11 @@ def pmap(fun: Callable[..., T],
       #   out_axes == 0 path working as we look for a better solution.
       out_axes_thunk = HashableFunction(
         lambda: (0,) * out_tree().num_leaves,
-        key=out_axes)
+        closure=out_axes)
     else:
       out_axes_thunk = HashableFunction(
         lambda: tuple(flatten_axes("pmap out_axes", out_tree(), out_axes)),
-        key=out_axes)
+        closure=out_axes)
     out = pxla.xla_pmap(
         flat_fun, *args, backend=backend, axis_name=axis_name,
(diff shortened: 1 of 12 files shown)


mattjj (changes requested):
Can we add some tests here? In the review on #4904 I mentioned some ideas for tests we could add, and it sounds like there is a specific issue that prompted this change for which we should add a test too.

mattjj:
In particular, it seems like this fix may have been prompted by a grad-of-pmap caching bug. If so, such a test could be similar to the [`test_grad_of_jit_compilation_caching`](https://github.com/google/jax/blob/087b7b81ac356cb9c246354fcf51ddeef6387a3c/tests/api_test.py#L1766) test.

skye:
There's an even better way to test caching: `jtu.count_jit_and_pmap_compiles()`! See the test case I added in https://github.com/google/jax/pull/5037/files (I'll also update that PR to add the extra cache key assuming this goes in first).

apaszke:
So I thought about adding a test, but honestly I would like to make this a bit more structured.

This directly touches upon our discussion about #4923, which fixes the situation with vmap, but not AD. This patch should make AD better, but not vmap... I'd really like us to sketch out what kind of compositions do we expect to preserve the compilation cache, because without it we might just end up with a bunch of random cases in the tests, and that's not a good way to design API semantics. This is why I've skipped the test for now, but I can add it if you _really_ think that it's a good idea to evolve it in such an ad-hoc way.

mattjj (approved):

PR #5066, opened 2020-12-01
[jax2tf] Add testing for the conversion of (almost) all the remaining untested primitives

bchetioui:
This PR contains testing (and some fixes) for the conversion of:
- `integer_pow`
- `pow`
- `reshape`
- `rev`
- `device_put`
- `sub`
- `bitcast_convert_type`
- `add_any`
- `tie_in`
- `stop_gradient`

Three last primitives are completely missing from the tests:
- `convert_element_type` – which is waiting on (#5063 and #4959)
- `shape_as_value`
- `custom_lin`

Another missing thing from this PR might be the harnesses for the unary elementwise/binary elementwise operations as well, which I plan to rewrite as some previous examples seem to indicate they do not provide sufficient coverage.

### DIFF
--- a/jax/experimental/jax2tf/tests/primitive_harness.py
+++ b/jax/experimental/jax2tf/tests/primitive_harness.py
@@ -233,6 +233,27 @@ lax_convert_element_type = tuple( # Validate dtypes to dtypes
                    else set(jtu.dtypes.all) - set(jtu.dtypes.all_unsigned))
 )
 
+def _make_integer_pow_harness(name, *, shape=(20, 30), dtype=np.int32, y=3):
+  return Harness(f"{name}_shape={jtu.format_shape_dtype_string(shape, dtype)}_y={y}",
+                 lax.integer_pow,
+                 [RandArg(shape, dtype), StaticArg(y)],
+                 shape=shape,
+                 dtype=dtype,
+                 y=y)
+
+lax_integer_pow = tuple( # Validate dtypes and y values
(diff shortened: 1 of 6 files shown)


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -179,6 +180,212 @@ def parameterized(harness_group: Iterable[Harness],
...
+                 dtypes=dtypes)
+
+ad_util_add_jaxvals = tuple( # Validate dtypes
+  _make_add_jaxvals_harness("dtypes", dtype=dtype)
+  for dtype in set(jtu.dtypes.all) - set(jtu.dtypes.boolean)
+)
+
+lax_tie_in = tuple(
+  Harness(f"lhs={jtu.format_shape_dtype_string(lhs_shape, lhs_dtype)}_rhs={jtu.format_shape_dtype_string(rhs_shape, rhs_dtype)}",
+          lax.tie_in_p.bind,
```
gnecula, line 361: This primitive is deprecated and not used if omnistaging is on. It does not hurt to test it though.

gnecula (approved):
Thanks!

jax/experimental/jax2tf/tests/primitive_harness.py
bchetioui: The test is actually skipped if omnistaging is on. Since it is still possible to end up with a call to `tie_in_p` somewhere by disabling omnistaging, I figured it is best to ensure it works properly.

PR #5075, opened 2020-12-02
Ensure lax.convert_element_type preserves concrete inputs

jakevdp:
There are several places in the JAX library where it is implicitly assumed that `lax.convert_element_type` will return concrete results for concrete inputs. This doesn't always hold, though; for example:
```python
from jax.scipy.special import multigammaln
from jax import core, partial, jit, lax, grad
import numpy as np

jit(lambda x: multigammaln(x, 1))(1)  # executes
jit(lambda x: multigammaln(x, np.int32(1)))(1)  # fails
```
The second line fails with onmistaging because `lax.convert_element_type` does not ensure that concrete inputs always result in concrete outputs.

This PR fixes this issue, and adds a test to ensure this is the case.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -432,8 +432,6 @@ def convert_element_type(operand: Array, new_dtype: DType) -> Array:
   if type(operand) in dtypes.python_scalar_dtypes:
     operand = np.asarray(operand, new_dtype)
   old_dtype = dtypes.canonicalize_dtype(_dtype(operand))
-  if old_dtype == new_dtype:
-    return operand
   if (dtypes.issubdtype(old_dtype, np.complexfloating) and
       not dtypes.issubdtype(new_dtype, np.complexfloating)):
     msg = "Casting complex values to real discards the imaginary part"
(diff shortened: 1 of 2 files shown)


mattjj:
Interesting!

> There are several places in the JAX library where it is implicitly assumed that lax.convert_element_type will return concrete results for concrete inputs

 Could you unpack this more? That is, maybe these implicit assumptions are wrong, and we should be revising those places rather than revising convert_element_type.

jakevdp:
So I'm not sure about all the places this might arise, but any time we call utilities like `_promote_args`, it will call `lax.convert_element_type()` under the hood, and currently this will turn static dtype instances into dynamic traced values, which may break the assumptions in the code.

Before omnistaging, this was not the case: convert_element_type would not be staged when applied to static values.

With omnistaging, that broke in rare cases that happen to not be well covered in our test suite, so we didn't notice.

I came across this bug when refactoring `lax.convert_element_type` in a way that broke this assumption further, and it caused a number of test failures due to staging type conversions for concrete values.

I think the right way to go here is to restore the pre-omnistaging behavior, so that we can safely use `lax.convert_element_type` and utilities like `_promote_args` without having to consider whether the inputs are static.

What do you think?

jakevdp:
I'm going to close this because it's not clear that we want to always stage `lax.convert_element_type`.

PR #5077, opened 2020-12-02
improve sinc jvp at zero

### LINKED ISSUE #5054: Add custom JVP rule for jnp.sinc
The current implementation has incorrect even-ordered derivatives at `x=0.0`:
```python
import jax.numpy as jnp
from jax import grad, vmap

x = jnp.linspace(-5, 5, 101)
y = vmap(grad(grad(jnp.sinc)))(x)

plt.plot(x, y)
```
download

Related to #5039

mattjj:
Thanks to @dougalm for thinking this through with me!

The basic problem in #5054 is that we had for `sinc` an expression like

```python
  return where(x == 0, 1., sin(pi * x) / (pi * x))
```

One way to read that is at `x == 0`, we're replacing the function with its zeroth-order Taylor expansion. But that meant at zero we aren't computing its correct derivatives! (We were getting lucky with the first derivative at zero: it happens to be zero because that's the slope of the sinc function at zero, but we were computing zero because we were differentiating a constant function.)

One way to think about the fix here is that we're replacing the constant function at zero with a truncated Taylor series at zero. For any fixed Taylor series truncation level, we'd only be able to support some finite order of autodiff before incorrectly truncating the derivatives to zero. So we play a trick: we use a `custom_jvp` rule to in effect generate all the Taylor series coefficients we need, lazily as we differentiate.

fixes #5054

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1015,9 +1015,9 @@ def _sinc_maclaurin(k, x):
   # compute the kth derivative of x -> sin(x)/x evaluated at zero (since we
   # compute the monomial term in the jvp rule)
   if k % 2:
-    return lax._const(x, 0)
+    return lax.full_like(x, 0)
   else:
-    return lax._const(x, (-1) ** (k // 2) / (k + 1))
+    return lax.full_like(x, (-1) ** (k // 2) / (k + 1))
 
 @_sinc_maclaurin.defjvp
 def _sinc_maclaurin_jvp(k, primals, tangents):
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1005,10 +1005,24 @@ def sinc(x):
   _check_arraylike("sinc", x)
   x, = _promote_dtypes_inexact(x)
   eq_zero = lax.eq(x, lax._const(x, 0))
-  safe_x = where(eq_zero, lax._const(x, 0), x)
-  pi_x = lax.mul(lax._const(x, pi), safe_x)
-  return where(eq_zero,
-               lax._const(x, 1), lax.div(lax.sin(pi_x), pi_x))
+  pi_x = lax.mul(lax._const(x, pi), x)
+  safe_pi_x = where(eq_zero, lax._const(x, 0), pi_x)
+  return where(eq_zero, _sinc_maclaurin(0, pi_x),
+               lax.div(lax.sin(safe_pi_x), safe_pi_x))
+
+@partial(custom_jvp, nondiff_argnums=(0,))
+def _sinc_maclaurin(k, x):
+  # compute the kth derivative of x -> sin(x)/x evaluated at zero (since we
+  # compute the monomial term in the jvp rule)
+  if k % 2:
+    return lax._const(x, 0)
+  else:
+    return lax._const(x, (-1) ** (k // 2) / (k + 1))
+
+@_sinc_maclaurin.defjvp
+def _sinc_maclaurin_jvp(k, primals, tangents):
+  (x,), (t,) = primals, tangents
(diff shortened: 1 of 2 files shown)


apaszke (approved):

jax/_src/numpy/lax_numpy.py
```diff
@@ -1005,10 +1005,24 @@ def sinc(x):
...
+               lax.div(lax.sin(safe_pi_x), safe_pi_x))
+
+@partial(custom_jvp, nondiff_argnums=(0,))
+def _sinc_maclaurin(k, x):
+  # compute the kth derivative of x -> sin(x)/x evaluated at zero (since we
+  # compute the monomial term in the jvp rule)
+  if k % 2:
+    return lax._const(x, 0)
+  else:
+    return lax._const(x, (-1) ** (k // 2) / (k + 1))
```
apaszke, line 1020: Actually, isn't this missing a factorial in the denominator?

mattjj: Great thing to check! It's always confusing to me where the factorials live. The only way I'm ever confident is by checking numerically. (And perhaps the function name could be chosen better; see below.)

I _think_ the current expression, without a factorial, is right:

```python
In [1]: from jax import grad

In [3]: f = jnp.sinc

In [4]: for i in range(10):
   ...:     f = grad(f)
   ...:

In [5]: f(0.)  # slow! jax.experimental.jet would be much faster
Out[5]: DeviceArray(-8513.462, dtype=float32)
```

That agrees with [Wolfram Alpha's symbolic answer](https://www.wolframalpha.com/input/?i=10th+derivative+of+sin%28pi+*+x%29+%2F+%28pi+*+x%29+at+0) of `-jnp.pi ** 10 / 11`.

Another way to check is to make sure the derivatives are all continuous at zero:

```python
from jax import grad, vmap
import jax.numpy as jnp


xs = jnp.linspace(-5, 5, 101)
plt.plot(xs, vmap(grad(jnp.sinc))(xs))
plt.plot(xs, vmap(grad(grad(jnp.sinc)))(xs))
plt.plot(xs, vmap(grad(grad(grad(jnp.sinc))))(xs))
plt.plot(xs, vmap(grad(grad(grad(grad(jnp.sinc)))))(xs))
plt.savefig('sinc_4th.png')
```

sinc_4th

(Interestingly, we can't use `jnp.linspace(-5, 5, 1001)` because for small nonzero values our derivatives get numerically unstable... maybe I'll use the Maclaurin series approximation _near_ zero, not just _at_ zero.)

(Also, this test caught a bug where I wasn't broadcasting arrays correctly!)

About the function name, this function is really computing the kth _derivative_ of sinc at zero, not the kth Taylor/Maclaurin coefficient (which is the kth derivative divided by k factorial). The function might instead be named `_deriv_of_sinc_at_zero`, at least in terms of describing the primal value that gets computed. But I renamed it to `_sinc_maclaurin` because the way it behaves under autodiff makes it act to its caller more like an accurate-as-needed Taylor expansion of sinc at zero.

jakevdp:
This is a really cool PR. That's all I have to contribute :grin:

gnool:
Really neat trick!

mattjj:
Thanks for the encouragement! To be honest I don't feel like I fully understand this technique yet. I'd like to keep thinking about it.

I want to improve the numerical stability of the sinc derivative near zero (not just at zero), but I haven't found time to think about that yet. So instead of letting the perfect be the enemy of the good/improved, I'm going to submit this improvement and then hopefully follow up with numerical stability later.

nadukandi:
Dear @mattjj . You might find the section 5.2 of the following article (it a work I did some years ago) relevant to this discussion. 
https://nadukandi.es/preprints/p_2014_nadukandi_numerically_stable_formulas_for_xivas_scheme.pdf 
That section address the following question: should we opt to write piecewise formulas for divided differences of any function, then what is the best series approximation we can use?

Issue #5082, opened 2020-12-02
Convert_element_type fails at converting floating point numbers to unsigned integers


bchetioui:
Running the following code

```python
from jax import lax
from jax import lax_reference
import numpy as np

arg = np.array([ 1.482590026198537
               , 1.7933394827929403
               , -1.7620562498625687
               , 0.09710140511430804
               , -1.0999886315120282
               , -3.121078288456996
               , 2.9317790831943236
               , -0.07594804943413619
               , -1.1277297959665809
               , -8.642109006576892], dtype=np.float32)
print(lax.convert_element_type(arg, new_dtype=np.uint16))
print(lax_reference.convert_element_type(arg, dtype=np.uint16))
```
gives the following output
```
[    1     1     0     0     0     0     2     0 65535 65528]
[    1     1 65535     0 65535 65533     2     0 65535 65528]
```

I understand from [the documentation](https://www.tensorflow.org/xla/operation_semantics#convertelementtype), that `convert_element_type` is supposed to perform the equivalent of an elementwise C++ `static_cast`, i.e. the output should correspond to the output of

```cpp
#include <iostream>

int main() {
  float arr[10] = { 1.482590026198537
                  , 1.7933394827929403
                  , -1.7620562498625687
                  , 0.09710140511430804
                  , -1.0999886315120282
                  , -3.121078288456996
                  , 2.9317790831943236
                  , -0.07594804943413619
                  , -1.1277297959665809
                  , -8.642109006576892};
  for(size_t i = 0; i < 10; ++i) {
    std::cout << static_cast<unsigned short>(arr[i]) << " ";
  }
  std::cout << std::endl;
}
```
which is
```
1 1 65535 0 65535 65533 2 0 65535 65528
```
and corresponds to the output of the call to `lax_reference.convert_element_type`.

This seemingly occurs when converting any floating point type to `uint8` or `uint16`.

jakevdp:
Thanks! I'll take a look.

jakevdp:
This is a known issue; XLA:ConvertElementType is not well defined for some type pairs. See https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#miscellaneous-divergences-from-numpy for more

PR #5083, opened 2020-12-02
[jax2tf] Added testing for the conversion of convert_element_type.

bchetioui:
This PR currently fails tests for `uint` dtypes, which is why they are disabled.
I believe the bug is in the XLA implementation of `convert_element_type`, however (see #5082).


jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -572,6 +572,19 @@ def custom_assert(result_jax, result_tf):  # noqa: F811
   def test_round(self, harness: primitive_harness.Harness):
     self.ConvertAndCompare(harness.dyn_fun, *harness.dyn_args_maker(self.rng()))
 
+  @primitive_harness.parameterized(primitive_harness.lax_convert_element_type)
+  def test_convert_element_type(self, harness: primitive_harness.Harness):
+    old_dtype = harness.params["old_dtype"]
```
gnecula, line 577: I think it would be best to try to move this filtering in the construction of the harnesses, perhaps also with a comment about the nature of the bug.

PR #5085, opened 2020-12-02
lax.convert_element_type: always return DeviceArray

jakevdp:
This required disabling two tests which now fail because they were explicitly checking that a device push does *not* happen for some values.

The duplicated constants may have problematic implications in terms of memory usage - I'm open to ideas about that.


hawkinsp (approved):
It's probably worth running a global presubmit for this.

jakevdp:
> It's probably worth running a global presubmit for this.

Will do.

Do you have any thoughts on whether the duplicated constants should be a blocker for this? I think the main issue would be cases where someone is using a large numpy array in a jax calculation, this might lead to the array being pushed to device multiple times. But I'm not sure of all the ramifications of that...

Issue #5069, opened 2020-12-01
Tensorboard profiling fails with jax


tomweingarten:
I get the following error when I follow the jax documentation for profiling using tensorboard:
`Failed to capture profile: ; No such file or directory`

The console output from tensorboard is:
```
2020-12-01 11:16:21.437665: I tensorflow/stream_executor/platform/default/dso_loader.cc:49] Successfully opened dynamic library libcudart.so.11.0
TensorBoard 2.5.0a20201120 at http://0.0.0.0:6006/
```

One thing to note is that in order to make CUDA 11.1 work, libcudart.so.11.0 is a symbolic link to libcudart.so.11.1.74

I'm using CUDA 11.1 and CUDNN 8.0.5. I've tried this on using both the stable and nightly versions of tensorboard and the tensorboard profiling plugin.

bchetioui:
@tomweingarten did you try with the versions recommended by @zhangqiaorjc [here](https://github.com/google/jax/issues/4427#issuecomment-731341850)?

tomweingarten:
Thanks! Just gave those versions a try, unfortunately I'm still getting the same error. I should note I compiled jax and jaxlib from master so that I can use CUDA 11.1 and CUDNN 8.0.5

8bitmp3:
Hi @tomhennigan, will you be able to list the steps you've taken so far?

For example:

1. Update/uninstall CUDA (in Colab):

```shell
!apt-get --purge remove cuda nvidia* libnvidia-*
!dpkg -l | grep cuda- | awk '{print $2}' | xargs -n1 dpkg --purge
!apt-get remove cuda-*
!apt autoremove
!apt-get update
```

2. Install CUDA 11.1:

```shell
# Check that your version of Ubuntu on Colab is 18.04
!lsb_release -a
```

```shell
# Download CUDA 11.1 for Ubuntu 18.04
# (source: https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=1804&target_type=deblocal
!wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/cuda-ubuntu1804.pin
!sudo mv cuda-ubuntu1804.pin /etc/apt/preferences.d/cuda-repository-pin-600
[...]

Issue #4427, opened 2020-09-30
No tools in profiler


tancik:
Ever since I upgraded jax, the profiler no longer works for me. It is able to capture a trace but there are no tools to investigate the trace.

![image](https://user-images.githubusercontent.com/3310961/94741785-8384e800-0329-11eb-895e-4f5d73499f47.png)

I start the profiler with:
```
import jax.profiler
server = jax.profiler.start_server(9999)
```

jax - 0.2.0
jaxlib - 0.1.55
tensorboard - 2.3.0
tensorboard-plugin-profile - 2.3.0

DannyKong12:
I can confirm this happens with latest tensorboard as well as v2.2.0

DannyKong12:
Downgrading back to jax==0.1.75 and jaxlib==0.1.52 everything works as expected. Tools dropdown has (5) options including trace_viewer. Any idea on what the timeline is for bringing tensorboard support back/anything we can do to help?

tancik:
Interesting, I downgraded back to jax==0.1.75 and jaxlib==0.1.52 and I still have zero tools available.

DannyKong12:
I had to capture profile again, but the same code works as expected on 0.1.75 for me. Checking the logdir directory I also see different files, certain files are just not produced after upgrading.

shoyer:
@tfrerix is struggling with the same issue here

zhangqiaorjc:
I'm able to repro this issue. I may have found a root cause. This may require a fix from tensorflow/profiler folks.

AnimatedRNG:
I also ran into this issue with the latest jax/jaxlib; downgrading to jax==0.1.75 and jaxlib==0.1.52 (CUDA 10.1; cuDNN 7.6.5) and capturing a new profile fixed things on my end

qiuminxu:
Profiler recently moved the data processing from TensorFlow data collection side into the TensorBoard. Data collection will generate a single .xplane.pb file, and TensorBoard will process and visualize the trace. These all lands in the recent nightly, the reported issue is caused by old version of profile plugin not able to visualize the new .xplane.pb format.

Can you try: `pip3 install -U tf-nightly tbp-nightly` and see if this works?

zhangqiaorjc:
I just verified that by upgrading tf-nightly to 2.5.0-dev20201027, you will get a backwards compatible profiler client that handles existing jaxlib versions.

Could people try `pip3 install -U tf-nightly`? 

[...]

PR #5090, opened 2020-12-03
Update TF/TB versions for profiling in documentation.

bchetioui:


docs/profiling.md
```diff
@@ -13,8 +13,8 @@ GPU and TPU. The end result looks something like this:
 ### Installation
```
8bitmp3, line 14: Hi @bchetioui 👋  Thank you so much for updating the JAX profiler docs. We're stuck at https://github.com/google/jax/issues/5069 as you probably are aware. And workarounds, such as https://github.com/google/jax/issues/4427, have helped some folks here.

May I suggest a small thing here from the UX perspective for users who aren't as experienced with TB:

```suggestion

Install specific nightly versions of TensorBoard, TensorFlow, as well as the What-If Tool TensorBoard:
```

8bitmp3 (commented):
Hi @bchetioui 👋  Thank you so much for updating the JAX profiler docs. We're stuck at https://github.com/google/jax/issues/5069 as you probably are aware. And workarounds, such as https://github.com/google/jax/issues/4427, have helped some folks here.

May I suggest a small thing here from the UX perspective for users who aren't as experienced with TB:

```diff

+ Install specific nightly versions of TensorBoard, TensorBoard profiler, TensorFlow, as well as the What-If Tool TensorBoard, as follows:
```

PR #2410, opened 2020-03-13
remove input shapes from params of some primitives

mattjj:
Long, long ago, when JAX was first born, we realized that we couldn't transpose this jaxpr:

```
{ lambda  ; a.
  let b = reduce_sum[ axes=(0,) ] a
  in b }
```

The problem was that the transpose of a reduce-sum is a broadcast, but because jaxprs didn't have shape information available, we didn't know what input shape to broadcast to!

Our hack was to have the primitives that required shape information for transposition acquire it into their parameters, so that we'd produce jaxprs like this one:

```
{ lambda  ; a.
  let b = reduce_sum[ axes=(0,)
                      input_shape=(3,) ] a
  in b }
```

That's not only aesthetically unpleasant, but also it meant we were limiting an (unused) capability of the system: ideally we should be able to trace a reduce-sum jaxpr without specializing on shape information (e.g. at the Unshaped level) and only require shape specialization for transposition. (Good thing no one actually traces at Unshaped...)

But at long last @chr1sj0nes in #2299 added avals to jaxprs, so that shape information (or whatever information with which the jaxpr was specialized out of Python) is in the jaxpr itself. So we could finally remove these shapes-in-params warts!

That's exactly what this commit does!

PR #5095, opened 2020-12-03
make convert_element_type_p not require old_dtype

mattjj:
Previously we needed to add old_dtype to the primitive's parameters so that it could be transposed. However, now that avals information is available in more places (in particular, attached to the UndefinedPrimal instances we use to indicate inputs with respect to which we are transposing), we don't need that kind of a hack!

This is a follow-up to #2410.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -438,8 +438,7 @@ def convert_element_type(operand: Array, new_dtype: DType) -> Array:
       not dtypes.issubdtype(new_dtype, np.complexfloating)):
     msg = "Casting complex values to real discards the imaginary part"
     warnings.warn(msg, np.ComplexWarning, stacklevel=2)
-  return convert_element_type_p.bind(
-      operand, new_dtype=new_dtype, old_dtype=old_dtype)
+  return convert_element_type_p.bind(operand, new_dtype=new_dtype)
 
 def bitcast_convert_type(operand: Array, new_dtype: DType) -> Array:
   """Elementwise bitcast.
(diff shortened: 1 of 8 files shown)


jax/_src/lax/lax.py
```diff
@@ -2571,38 +2567,35 @@ def _minmax_translation_rule(c, x, y, *, minmax=None, cmp=None):
 ad.defjvp_zero(lt_p)
 
 
-def _convert_element_type_shape_rule(operand, *, new_dtype, old_dtype):
+def _convert_element_type_shape_rule(operand, *, new_dtype):
```
jakevdp, line 2570: What would you think about using `dtype` rather than `new_dtype` as the argument name?

tests/lax_test.py
```diff
@@ -2344,6 +2344,7 @@ def testBroadcastInDim(self):
     self._Check(make_const, expected)
 
   def testConvertElementTypeMismatchedDTypeOldDType(self):
```
jakevdp, line 2346: delete?

jakevdp (approved):

tests/lax_test.py
mattjj: Hm yeah, WDYT? I have a superstition against deleting tests. But this is a pretty clearly harmless one to delete.

jax/_src/lax/lax.py
mattjj: That makes a ton of sense!

tests/lax_test.py
jakevdp: I think this one is pretty clearly obsolete at this point! I’d be fine removing it.

jax/_src/lax/lax.py
mattjj: Actually, because the argument to the lax function takes argument `new_dtype`, I think it might be better to leave the primitive parameter to have the same name.

tests/lax_test.py
bchetioui: I added this test earlier this week, I indeed don't think it serves any purpose anymore :)

mattjj: Okay, I'm convinced, it's safe to delete! Looks like this change is failing a bunch of jax2tf tests anyway :/

PR #5099, opened 2020-12-04
Support mapping a single logical axis to multiple mesh axes

apaszke:
This requires further relaxations to `ShardingSpec`s, so that `Chunked` dimensions can actually introduce multiple `ShardedAxis` indices.

cc @jekbradbury @skye


jekbradbury (approved):
It should probably be documented/commented that the way the logical axis is mapped to resources is major-to-minor in the order the resources are listed in the schedule, since preserving this order is something `mk_sharding_spec` has to be careful to do.

apaszke:
@jekbradbury I've added a comment and some extra tests to make sure we catch any misuses.

PR #5101, opened 2020-12-04
device_put_sharded tweaks and docstring detail

mattjj:
I had some ideas for small improvements to the device_put_sharded docstring and error checking!

I turned some assertions into errors, since they corresponded to checking whether the user input was valid (rather than internal errors). I always think of assertions as for internal consistency checks only, since in principle they can be disabled with `python -O` (though it seems no one does that).

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -2040,20 +2040,24 @@ def device_put(x, device: Optional[xc.Device] = None):
   return tree_map(lambda y: xla.device_put_p.bind(y, device=device), x)
 
 
-def device_put_sharded(x: Sequence[Any], devices: Sequence[xc.Device]):
-  """Transfers pre-sharded input to the specified devices, returning ShardedDeviceArrays.
+def device_put_sharded(shards: Sequence[Any], devices: Sequence[xc.Device]):
+  """Transfer array shards to specified devices and form ShardedDeviceArray(s).
 
   Args:
-    x: A sequence of arrays, scalars, or (nested) standard Python containers thereof.
-    devices: A sequence of devices()
+    shards: A sequence of arrays, scalars, or (nested) standard Python
+      containers thereof representing the shards to be stacked together to form
+      the output. The length of ``shards`` must equal the length of ``devices``.
+    devices: A sequence of :py:class:`Device` instances representing the devices
+      to which corresponding shards in ``shards`` will be transferred.
 
   Returns:
-    A ShardedDeviceArray or (nested) Python container thereof containing a stacked
-    version of x sharded across the specified devices.
+    A ShardedDeviceArray or (nested) Python container thereof representing the
+    elements of ``shards`` stacked together, with each shard backed by physical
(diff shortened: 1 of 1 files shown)


jax/api.py
```diff
@@ -2081,20 +2085,27 @@ def device_put_sharded(x: Sequence[Any], devices: Sequence[xc.Device]):
...
-  # TODO(jakevdp): provide a default for devices that considers both local devices and pods
-  assert len(x) == len(devices), f"len(x) = {len(x)} must equal len(devices) = {len(devices)}."
+  # TODO(jakevdp): provide a default for devices that considers both local
+  # devices and pods
+  if not isinstance(shards, Sequence):
+    raise ValueError("device_put_sharded `shards` input must be a sequence; "
+                     f"got {type(shards)}")
+  if not len(shards) == len(devices):
+    raise ValueError(f"len(shards) = {len(shards)} must equal "
+                     f"len(devices) = {len(devices)}.")
```
jakevdp, line 2095: I'm looking forward to the day when we're on Python 3.8 and can write this as `f"{len(shards)=} must equal {len(devices)=}"` :grin:

jax/api.py
```diff
@@ -2081,20 +2085,27 @@ def device_put_sharded(x: Sequence[Any], devices: Sequence[xc.Device]):
...
     avals = [core.raise_to_shaped(core.get_aval(x)) for x in xs]
-    # We cannot check aval equality directly, because it may fail for ConcreteArray.
-    assert all(aval.shape == avals[0].shape and aval.dtype == avals[0].dtype for aval in avals),\
-      f"abstract values not compatible: {avals}"
-    x_aval = core.raise_to_shaped(avals[0])
-    aval = ShapedArray((len(devices),) + x_aval.shape, x_aval.dtype)
-    buffers = list(it.chain.from_iterable(xla.device_put(x, d) for x, d in zip(xs, devices)))
-    return pxla.ShardedDeviceArray(aval, buffers)
-  return tree_multimap(_device_put_sharded, *x)
+    if not all(a1 == a2 for a1, a2 in zip(avals[:-1], avals[1:])):
```
jakevdp, line 2099: Do we need to worry about the avals being concrete arrays here? If they are, the equality check will check value, rather than just shape/dtype.

jakevdp (approved):
Looks great, thanks!

jax/api.py
mattjj: Good thing to check! I don't think so, because of the `core.raise_to_shaped`. (I wondered if perhaps the concern about ConcreteDeviceArrays was left over from some point when `raise_to_shaped` wasn't called.)

jax/api.py
mattjj: image

jax/api.py
jakevdp: Ah, right... that makes sense.

PR #5104, opened 2020-12-04
Support ndarrays as arguments to cg and gmres

shoyer:
This is consistent with SciPy, and makes things a little bit less
surprising for users.

### DIFF
--- a/jax/_src/scipy/sparse/linalg.py
+++ b/jax/_src/scipy/sparse/linalg.py
@@ -73,6 +73,21 @@ def _identity(x):
   return x
 
 
+def _normalize_matvec(f):
+  """Normalize an argument for computing matrix-vector products."""
+  if callable(f):
+    return f
+  elif isinstance(f, (np.ndarray, jnp.ndarray)):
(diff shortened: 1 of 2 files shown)


jax/_src/scipy/sparse/linalg.py
```diff
@@ -73,6 +73,15 @@ def _identity(x):
   return x
 
 
+def _normalize_matvec(f):
+  """Normalize an argument for computing matrix-vector products."""
+  if callable(f):
+    return f
+  else:
```
mattjj, line 80: WDYT about something like

```python
if callable(f):
  return f
elif isinstance(f, jax.numpy.ndarray):
  return partial(_dot, f)
else:
  raise TypeError(...)
```

Basically, do a bit more input validation here, if it's easy. Could also check that it's 2D and square.

mattjj (approved):

jax/_src/scipy/sparse/linalg.py
shoyer: Good idea, done!

PR #5105, opened 2020-12-04
add jax.device_put_replicated

mattjj:
Add `jax.device_put_replicated`. Also move tests for `device_put_sharded` into pmap_test.py, since that file tests with multiple devices even in our OSS CI. Add both `device_put_replicated` and `device_put_sharded` to jax/__init__.py so they can be used as `jax.device_put_replicated` and `jax.device_put_sharded`.

DIFFBASE=#5101

### DIFF
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -2051,7 +2051,7 @@ class ShardedDeviceArrayTest(jtu.JaxTestCase):
     self.assertIsInstance(y, pxla.ShardedDeviceArray)
     self.assertEqual(len(y.device_buffers), len(devices))
     self.assertTrue(all(b.device() == d for b, d in zip(y.device_buffers, devices)))
-    self.assertArrayEqual(y, jnp.stack(x))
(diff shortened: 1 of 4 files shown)


tests/pmap_test.py
```diff
@@ -2043,6 +2043,56 @@ def testNoCopyIndexing1D(self):
...
+
+  def test_device_put_replicated_array(self):
+    devices = jax.local_devices()
+    n_devices = len(devices)
+    x = np.arange(1, 5)
+    y = jax.device_put_replicated(x, devices)
+    self.assertIsInstance(y, pxla.ShardedDeviceArray)
+    self.assertEqual(len(y.device_buffers), len(devices))
+    self.assertTrue(all(b.device() == d for b, d in zip(y.device_buffers, devices)))
+    self.assertAllClose(y, np.stack([x for _ in devices]))
```
jakevdp, line 2076: `self.assertArraysEqual` may be more appropriate here... similar elsewhere in the new tests.

jakevdp (approved):

PR #5108, opened 2020-12-05
[jax2tf] Fix bug in conversion of bfloat16 to tensor

### LINKED ISSUE #5106: Infinite recursion in jax2tf with bfloat16 input.
Short repro:
```python
import jax.numpy as jnp

jax2tf.convert(lambda x: x)(jnp.bfloat16(1.0))
```
```pytb
  File "tmp.py", line 7, in <module>
    f(jnp.bfloat16(1.0))
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 278, in converted_fun
    out_flat = converted_fun_flat_with_custom_gradient(*args_flat)
  File "/Users/vanderplas/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/tensorflow/python/ops/custom_gradient.py", line 258, in __call__
    return self._d(self._f, a, k)
  File "/Users/vanderplas/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/tensorflow/python/ops/custom_gradient.py", line 212, in decorated
    return _eager_mode_decorator(wrapped, args, kwargs)
  File "/Users/vanderplas/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/tensorflow/python/ops/custom_gradient.py", line 409, in _eager_mode_decorator
    result, grad_fn = f(*args, **kwargs)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 273, in converted_fun_flat_with_custom_gradient
    out_with_avals = _interpret_fun(flat_fun, args_flat, args_avals_flat)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 304, in _interpret_fun
    out_vals: Sequence[Tuple[TfVal, core.AbstractValue]] = fun.call_wrapped(*in_vals)
  File "/Users/vanderplas/github/google/jax/jax/linear_util.py", line 155, in call_wrapped
    args, kwargs = next(gen)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 341, in _interpret_subtrace
    in_tracers = tuple(TensorFlowTracer(trace, val, aval)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 341, in <genexpr>
    in_tracers = tuple(TensorFlowTracer(trace, val, aval)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 657, in __init__
    self.val = _safe_convert_to_tensor(val, dtype=self._aval.dtype)  # type: ignore[attr-defined]
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 95, in _safe_convert_to_tensor
    val = tf.bitcast(tf.convert_to_tensor(val.view(jnp.uint16),
  File "/Users/vanderplas/github/google/jax/jax/_src/numpy/lax_numpy.py", line 4909, in _view
    return lax.bitcast_convert_type(arr, dtype)

<snip>

  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 703, in pure
    return TensorFlowTracer(self, val, core.ShapedArray(shape, dtype))
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 657, in __init__
    self.val = _safe_convert_to_tensor(val, dtype=self._aval.dtype)  # type: ignore[attr-defined]
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 95, in _safe_convert_to_tensor
    val = tf.bitcast(tf.convert_to_tensor(val.view(jnp.uint16),
  File "/Users/vanderplas/github/google/jax/jax/_src/numpy/lax_numpy.py", line 4909, in _view
    return lax.bitcast_convert_type(arr, dtype)
  File "/Users/vanderplas/github/google/jax/jax/_src/lax/lax.py", line 463, in bitcast_convert_type
    return bitcast_convert_type_p.bind(operand, new_dtype=new_dtype)
  File "/Users/vanderplas/github/google/jax/jax/core.py", line 270, in bind
  File "/Users/vanderplas/github/google/jax/jax/util.py", line 36, in safe_map
    return list(map(f, *args))
  File "/Users/vanderplas/github/google/jax/jax/core.py", line 377, in full_raise
    return self.pure(val)
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 703, in pure
    return TensorFlowTracer(self, val, core.ShapedArray(shape, dtype))
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 657, in __init__
    self.val = _safe_convert_to_tensor(val, dtype=self._aval.dtype)  # type: ignore[attr-defined]
  File "/Users/vanderplas/github/google/jax/jax/experimental/jax2tf/jax2tf.py", line 92, in _safe_convert_to_tensor
    if not isinstance(val, jnp.ndarray):
  File "/Users/vanderplas/github/google/jax/jax/_src/numpy/lax_numpy.py", line 112, in __instancecheck__
    return isinstance(instance.aval, _arraylike_types)
RecursionError: maximum recursion depth exceeded in __instancecheck__
```
I believe this is the issue blocking #5085

jakevdp:
Fixes #5106

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -52,6 +52,7 @@ from jaxlib import xla_client
 _VALID_SCOPE_REGEX = re.compile("^[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*$")
 _INVALID_SCOPE_CHAR = re.compile("[^A-Za-z0-9_.\\/>-]")
 
+_dtype = np.result_type
 
 def _sanitize_scope_name(name):
   scope_name = _INVALID_SCOPE_CHAR.sub("_", name)
@@ -87,13 +88,15 @@ def _safe_convert_to_tensor(val, dtype=None) -> TfVal:
   `ndarray` with a `uint16` dtype, for which conversion is properly defined. Then, we
   simply bitcast it back to `bfloat16`.
   """
-  dtype = dtype if dtype else (val.dtype if hasattr(val, "dtype") else None)
-  if (dtype == jnp.bfloat16 or isinstance(val, jnp.bfloat16)):
-    if not isinstance(val, jnp.ndarray):
-      val = np.array(val, jnp.bfloat16)
+  dtype = dtype or _dtype(val)
 
-    val = tf.bitcast(tf.convert_to_tensor(val.view(jnp.uint16),
-                                          dtype=to_tf_dtype(jnp.uint16)),
+  if (dtype == jnp.bfloat16 or _dtype(val) == jnp.bfloat16):
+    # TODO: handle bfloat16 without going through numpy.
+    # We cannot call val.view() on a DeviceArray, because it passes val to
+    # bitcast_convert_type_p, which calls back here, leading to an infinite
+    # recursion.  Instead, copy to a numpy array and create the view there.
+    val = np.asarray(val, jnp.bfloat16).view(np.uint16)
+    val = tf.bitcast(tf.convert_to_tensor(val, dtype=to_tf_dtype(jnp.uint16)),
                      type=to_tf_dtype(jnp.bfloat16))
   else:
     conversion_type = to_tf_dtype(dtype) if dtype else None
--- a/jax/experimental/jax2tf/tests/primitives_test.py
+++ b/jax/experimental/jax2tf/tests/primitives_test.py
@@ -944,5 +944,13 @@ class JaxPrimitiveTest(tf_test_util.JaxToTfTestCase):
     self.assertAllClose(tf_fn_array(np.array([3, 4, 5])),
                         np.array([4.5, 10, 17.5], jnp.bfloat16))
 
+  # test_bfloat16_array checks that https://github.com/google/jax/issues/5106 is
+  # fixed
+  def test_bfloat16_array(self):
+    @jax2tf.convert
+    def f(x):
+      return x
+    f(jnp.bfloat16(1))
+
 if __name__ == "__main__":
   absltest.main(testLoader=jtu.JaxTestLoader())
--- a/jax/experimental/jax2tf/tests/jax2tf_test.py
+++ b/jax/experimental/jax2tf/tests/jax2tf_test.py
@@ -32,6 +32,7 @@ import tensorflow as tf  # type: ignore[import]
 
 config.parse_flags_with_absl()
 
+numpy_version = tuple(map(int, np.version.version.split('.')[:2]))
 
 class Jax2TfTest(tf_test_util.JaxToTfTestCase):
 
@@ -252,8 +253,11 @@ class Jax2TfTest(tf_test_util.JaxToTfTestCase):
       jax2tf.convert(5.)
 
   def test_convert_argument_non_tensor_error(self):
-    with self.assertRaisesRegex(TypeError,
-                                "Argument.*should be NumPy array"):
+    if numpy_version < (1, 19):
+      msg = "data type not understood"
+    else:
+      msg = "Cannot interpret .* as a data type"
+    with self.assertRaisesRegex(TypeError, msg):
       jax2tf.convert(lambda x: x)(lambda y: y)
 
   def test_argument_eager_tensor(self):


mattjj (approved):

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -52,6 +52,7 @@
 _VALID_SCOPE_REGEX = re.compile("^[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*$")
 _INVALID_SCOPE_CHAR = re.compile("[^A-Za-z0-9_.\\/>-]")
 
+_dtype = np.result_type
```
gnecula, line 55: Do we really need this abbreviation? It only saves a few characters.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -87,13 +88,15 @@ def _safe_convert_to_tensor(val, dtype=None) -> TfVal:
...
-    val = tf.bitcast(tf.convert_to_tensor(val.view(jnp.uint16),
-                                          dtype=to_tf_dtype(jnp.uint16)),
+  dtype = dtype or _dtype(val)
+
+  if (dtype == jnp.bfloat16 or _dtype(val) == jnp.bfloat16):
+    # TODO: handle bfloat16 without going through numpy.
+    # We cannot call val.view() on a DeviceArray, because it passes val to
+    # bitcast_convert_type_p, which calls back here, leading to an infinite
+    # recursion.  Instead, copy to a numpy array and create the view there.
+    val = np.asarray(val, jnp.bfloat16).view(np.uint16)
```
gnecula, line 98: Since the JAX and TF `bfloat16` actually have the same representation, even though they look like different types, is it possible to do the conversion without any array copy?

gnecula: @bchetioui PTAL

jax/experimental/jax2tf/jax2tf.py
jakevdp: This is what we use in `lax_numpy.py` and I thought it makes the intent clearer at the call-site. Happy to use `np.result_type` directly if you'd prefer.

gnecula: This is minor, but I do prefer `np.result_type`. This is one aspect of Google style guide that I like to use: identifiers from other modules use `module.id` style.

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -944,5 +944,13 @@ def jax_fn_array(x):
     self.assertAllClose(tf_fn_array(np.array([3, 4, 5])),
                         np.array([4.5, 10, 17.5], jnp.bfloat16))
 
+  # test_bfloat16_array checks that https://github.com/google/jax/issues/5106 is
+  # fixed
+  def test_bfloat16_array(self):
+    @jax2tf.convert
+    def f(x):
+      return x
+    f(jnp.bfloat16(1))
```
bchetioui, line 953: I am actually not able to pass this test on my machine. @jakevdp what versions of TF and NumPy are you running this with?

jax/experimental/jax2tf/jax2tf.py
bchetioui: I can not think of a way to do it from inside jax2tf. However, with a little more careful of handling of dtypes (i.e. adding calls to `to_jax_dtype` and `to_tf_dtype` where it is necessary to get the expected bfloat16 representation), we should be able to call to make a call like `jax2tf.convert(func_jax)(jnp_bfloat16_array.view(tf.bfloat16.as_numpy_dtype))`.

This is not ideal – as it puts a burden on the user, and adds some clutter to the code (which seems like it might be annoying to test thoroughly). The ideal solution would be for the TF and JAX bfloat16 to be merged together, and then all this logic can go away.

gnecula: This merging of bfloat16 is now in progress, perhaps we can wait with this PR until that merging lands in TF.

jakevdp: The bug this PR addresses is blocking work on weak types, so I’d prefer not to delay the fix.

jakevdp: Another option would be to wait, but disable the problematic tests, with the understanding that changes to `lax.convert_element_type` will break some uses of `bfloat16` in jax2tf.

jax/experimental/jax2tf/tests/primitives_test.py
jakevdp: Yes, it fails on my machine too, but no longer fails with an infinite recursion error.

I pushed it because ~half the jax2tf tests fail on my machine, so I was hoping it was one of the expected failures :grin:

bchetioui: Oh got it! I think tests fail because you are not using a recent enough version of `tf-nightly` :-) I think the tests should run properly if you simply use the latest one. Otherwise, for future references, the [jax2tf README](https://github.com/google/jax/tree/master/jax/experimental/jax2tf) should in principle document a version that works (actually it doesn't right now, because a breaking bfloat16 change happened in October, so I will update it!).

BTW, one reason we didn't catch the error you encountered sooner is that we perform all our jax2tf tests in the TF world, i.e. with the JAX arguments converted to TF tensors with the proper dtype. The failure you encounter is a similar kind of bfloat16 bug, which I think we can work around. But to be honest, if that logic has to be removed very soon (when bfloat16 on TF and JAX are unified), it might not be worth fixing here – because verifying that all these edge cases are handled properly is probably quite hard. I would be in favor of disabling whatever problematic tests there may be in the meantime (I actually do not expect that any of our tests will fail because of your change btw, since as I mentioned, all the calls they make to jax2tf conversions use `tf.Tensor`s as parameters).

jakevdp: OK, sounds good. For reference, this is the PR in question: https://github.com/google/jax/pull/5085

The jax2tf failures that are blocking the merge only show up on internal tests. I'll disable the relevant tests in order to get that change in.

bchetioui: SGTM! My bad, I can not see internal failures! :)

jakevdp:
OK - given review and offline conversations, I'm going to close this. Here's the plan as I understand it:

- disable the problematic `bfloat16_constant` test in #5085 to unblock it
- wait for the TF/JAX bfloat16 unification work to allow this special path for bfloat16 conversion to be removed
- at that point re-enable the test disabled in #5085

PR #5110, opened 2020-12-05
Implement linear_ramp pad mode in jax.numpy.pad

minoring:
Related #5010

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2347,6 +2347,34 @@ def _pad_edge(array, pad_width):
   return array
 
 
+def _pad_linear_ramp(array, pad_width, end_values):
+  for axis in range(ndim(array)):
+    edge_before = lax.slice_in_dim(array, 0, 1, axis=axis)
+    edge_after = lax.slice_in_dim(array, -1, None, axis=axis)
+    ramp_before = linspace(
+        start=end_values[axis][0],
+        stop=edge_before.squeeze(axis), # Dimension is replaced by linspace
+        num=pad_width[axis][0],
+        endpoint=False,
+        dtype=array.dtype,
+        axis=axis
+    )
+    ramp_after = linspace(
+        start=end_values[axis][1],
+        stop=edge_after.squeeze(axis), # Dimension is replaced by linspace
+        num=pad_width[axis][1],
+        endpoint=False,
+        dtype=array.dtype,
+        axis=axis
+    )
+
+    # Reverse linear space in appropriate dimension
+    ramp_after = flip(ramp_after, axis)
+
+    array = lax.concatenate([ramp_before, array, ramp_after], dimension=axis)
+  return array
+
+
 def _pad_stats(array, pad_width, stat_length, stat_func):
   nd = ndim(array)
   for i in range(nd):
@@ -2408,8 +2436,8 @@ def _broadcast_to_pairs(nvals, nd, name):
   return nvals
 
 
-@partial(jit, static_argnums=(1, 2, 4))
-def _pad(array, pad_width, mode, constant_values, stat_length):
+@partial(jit, static_argnums=(1, 2, 4, 5))
+def _pad(array, pad_width, mode, constant_values, stat_length, end_values):
   array = asarray(array)
   nd = ndim(array)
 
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2347,6 +2347,31 @@ def _pad_edge(array, pad_width):
...
+            stop=edge.squeeze(axis), # Dimension is replaced by linspace
+            num=width,
+            endpoint=False,
+            dtype=array.dtype,
+            axis=axis
+        )
+        for end_value, edge, width in zip(
+            end_values[axis], edges, pad_width[axis]
+        )
+    )
```
jakevdp, line 2366: Slight preference here for just writing two linspace operations; reading over this, it took me far longer to understand what the list comprehension was doing than it would have if the two statements were just written out.

minoring: Thanks for your review. I agree it was hard to understand.
I converted it into two statements. PTAL :)

jakevdp (approved):
Looks great, thanks!

gnecula:
When trying to import this we ran into test failures, on CPU/GPU/TPU. E.g., on CPU `LaxBackedNumpyTests.testPadLinearRamp_shape=int32[1,4]_mode=linear_ramp_pad_width=(2, 0)_end_values=10.0`
fails with 
```
Max relative difference: nan
 x: array([[10, 10, 10, 10, 10, 10],
       [10,  6,  3,  6,  5,  4],
       [10,  3, -4,  1,  0, -1]], dtype=int32)
 y: array([[10, 10, 10, 10, 10, 10],
       [10,  6,  3,  5,  5,  4],
       [10,  3, -4,  1,  0, -1]], dtype=int32)
```

I am wondering why this test did not fail in GitHub. It may be because google uses internally a different version of numpy (1.16.4)?

jakevdp:
I suspect what's happening here is that rounding is happening in a different way for integer values. For example, here are two equivalent ways of generating float64 ramps:
```python
>>> np.linspace(0, 2, num=5, dtype=float)[::-1]                                                              
array([2. , 1.5, 1. , 0.5, 0. ])

>>> 2 - np.linspace(0, 2, num=5, dtype=float)                                                                
array([2. , 1.5, 1. , 0.5, 0. ])
```
but for integer dtypes, these equivalent approaches lead to differences in rounding direction:
```python
>>> np.linspace(0, 2, num=5, dtype=int)[::-1]                                                                
array([2, 1, 1, 0, 0])

>>> 2 - np.linspace(0, 2, num=5, dtype=int)                                                                  
array([2, 2, 1, 1, 0])
```
I suspect the numpy versions are generating linspaces one way, and the jax version is generating them the other way.

minoring:
It seems that there was a bug in linear_ramp padding for numpy <= 1.15
https://github.com/numpy/numpy/commit/1c45e0df150b1f49982aaa3fc1a328407b5eff7e#

Here are test results I found.

"numpy 1.15.x"
```python
a = [[-4, 1, 0, -1]]
np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0)
array([[10, 10, 10, 10, 10, 10],
       [10,  6,  3,  6,  5,  4],
       [10,  3, -4,  1,  0, -1]])

np.linspace(start=10.0, stop=1, endpoint=False, num=2, dtype=int)
array([10,  5])
```

"numpy 1.16.x"
```python
a = [[-4, 1, 0, -1]]
np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0)
array([[10, 10, 10, 10, 10, 10],
       [10,  6,  3,  5,  5,  4],
       [10,  3, -4,  1,  0, -1]])

np.linspace(start=10.0, stop=1, endpoint=False, num=2, dtype=int)
array([10,  5])
```

`np.linspace` works in the same way, but not `np.pad`

I am going to remove numpy <= 1.15 when testing linear_ramp.

jakevdp:
I think the issue must be something different - the tests in question were run on numpy 1.16.4

jakevdp:
It looks like this issue was fixed in numpy version 1.16.6:
```python
>>> import numpy as np
>>> print(np.__version__)
1.16.6
>>> a = np.array([[-4, 1, 0, -1]], dtype='int32')
>>> print(np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0))
print(np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0))
[[10 10 10 10 10 10]
 [10  6  3  5  5  4]
 [10  3 -4  1  0 -1]]
```
```python
>>> import numpy as np
>>> print(np.__version__)
1.16.5
>>> a = np.array([[-4, 1, 0, -1]], dtype='int32')
>>> print(np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0))
print(np.pad(a, pad_width=(2, 0), mode='linear_ramp', end_values=10.0))
[[10 10 10 10 10 10]
 [10  6  3  6  5  4]
 [10  3 -4  1  0 -1]]
```

minoring:
Yes, that seems the answer :) thanks for the testing.
I will change numpy version of testing.

jakevdp:
Thanks for the contribution, and for working out all the testing issues!

PR #5114, opened 2020-12-05
Add support for AMD GPU's (ROCm)

### LINKED ISSUE #2012: Add support for other GPUs (than NVIDIA)
Is it possible to run JAX on other GPU architectures other than NVIDIA (ex.: Intel, AMD)?

inailuig:
Adds the necessary build options for XLA.

Requires building with upstream TF until XLA in jax is updated to include https://github.com/tensorflow/tensorflow/pull/45344 

For build instructions see: https://github.com/google/jax/issues/2012#issuecomment-731843904
partially resolves https://github.com/google/jax/issues/2012

this is essentially ready, however there are still quite a few tests which fail because of missing features and bugs (presumably in XLA)

@hawkinsp

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -193,6 +193,8 @@ build --action_env=PYENV_ROOT
 build --python_path="{python_bin_path}"
 build --repo_env TF_NEED_CUDA="{tf_need_cuda}"
 build --action_env TF_CUDA_COMPUTE_CAPABILITIES="{cuda_compute_capabilities}"
+build --repo_env TF_NEED_ROCM="{tf_need_rocm}"
+build --action_env TF_ROCM_AMDGPU_TARGETS="{rocm_amdgpu_targets}"
 build --distinct_host_configuration=false
 build:linux --copt=-Wno-sign-compare
 build:macos --copt=-Wno-sign-compare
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):
This looks good to me. Let's merge it and iterate on it from here. It might need one or two more changes; let's see what the CI build says.

jaxlib/BUILD
```diff
@@ -76,13 +76,14 @@ pyx_library(
...
     name = "jaxlib",
     srcs = [
-        "cuda_prng.py",
-        "cusolver.py",
         "init.py",
         "pocketfft.py",
         "setup.py",
         "version.py",
-    ],
+    ] + if_cuda([
```
hawkinsp, line 83: This change actually causes problems for our internal CI. I think it might work with `if_cuda_is_configured`, although you'll need to verify that keeps the ROCm version working.

But also see my comments on the next change: ideally I think we can just share these python files?

hawkinsp (changes requested):
Would you mind squashing your commits as well? Thanks!

jaxlib/BUILD
inailuig: As long as 'gpu' does not default to 'ROCM' in xla_client.py we can omit this whole commit

inailuig: @hawkinsp Lets just leave in cusolver.py and live with the `cusolver_kernels is not defined` errors until we fix it in https://github.com/google/jax/pull/5115

inailuig: Or if you deem it necessary we could also write a dummy version which returns NotImplementedError

hawkinsp: Resolving this in the next PR seems fine. This is all new hardware support so it doesn't break anything!

hawkinsp (approved):

PR #5115, opened 2020-12-05
Add support for linear algebra ops on ROCm using rocBLAS/rocSolver

inailuig:
sequel to https://github.com/google/jax/pull/5114

Also enables the use of the built-in prng for ROCm (instead of a custom kernel like with cuda) since we did not include that in the last one.

The accompanying changes in TF/XLA were merged with https://github.com/tensorflow/tensorflow/pull/45583.
 
partially solves https://github.com/google/jax/issues/2012.

@hawkinsp

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -32,7 +32,9 @@ from jax._src.lax.lax import (
     _input_dtype, _broadcasting_select)
 from jax._src.lax import lax as lax_internal
 from jax.lib import lapack
+
 from jax.lib import cusolver
+from jax.lib import rocsolver
 
 from jax.lib import xla_client
 from jax.lib import xla_bridge as xb
@@ -330,8 +332,13 @@ def _cholesky_cpu_gpu_translation_rule(potrf_impl, c, operand):
 xla.backend_specific_translations['cpu'][cholesky_p] = partial(
   _cholesky_cpu_gpu_translation_rule, lapack.potrf)
 
-xla.backend_specific_translations['gpu'][cholesky_p] = partial(
-  _cholesky_cpu_gpu_translation_rule, cusolver.potrf)
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][cholesky_p] = partial(
+    _cholesky_cpu_gpu_translation_rule, cusolver.potrf)
+
+if rocsolver is not None:
+  xla.backend_specific_translations['gpu'][cholesky_p] = partial(
+    _cholesky_cpu_gpu_translation_rule, rocsolver.potrf)
 
 # Asymmetric eigendecomposition
 
@@ -520,10 +527,13 @@ _cpu_syevd = lapack.syevd
 xla.backend_specific_translations['cpu'][eigh_p] = partial(
   _eigh_cpu_gpu_translation_rule, _cpu_syevd)
 
-xla.backend_specific_translations['gpu'][eigh_p] = partial(
-  _eigh_cpu_gpu_translation_rule, cusolver.syevd)
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][eigh_p] = partial(
+    _eigh_cpu_gpu_translation_rule, cusolver.syevd)
 
-
+if rocsolver is not None:
+  xla.backend_specific_translations['gpu'][eigh_p] = partial(
+    _eigh_cpu_gpu_translation_rule, rocsolver.syevd)
 
 
 triangular_solve_dtype_rule = partial(
@@ -671,7 +681,7 @@ def _triangular_solve_cpu_translation_rule(
 xla.backend_specific_translations['cpu'][triangular_solve_p] = \
   _triangular_solve_cpu_translation_rule
 
-def _triangular_solve_gpu_translation_rule(
+def _triangular_solve_gpu_translation_rule(trsm_impl,
     c, a, b, left_side, lower, transpose_a, conjugate_a, unit_diagonal):
   shape = c.get_shape(a)
   dims = shape.dimensions()
@@ -681,7 +691,7 @@ def _triangular_solve_gpu_translation_rule(
     a = xops.Conj(a)
     conjugate_a = False
   if batch > 1 and m <= 32 and n <= 32:
-    return cusolver.trsm(
+    return trsm_impl(
       c, a, b, left_side, lower, transpose_a,
       conjugate_a, unit_diagonal)
   else:
@@ -694,8 +704,13 @@ def _triangular_solve_gpu_translation_rule(
     return xops.TriangularSolve(a, b, left_side, lower, unit_diagonal,
                                 transpose)
 
-xla.backend_specific_translations['gpu'][triangular_solve_p] = \
-    _triangular_solve_gpu_translation_rule
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][triangular_solve_p] = \
+      partial(_triangular_solve_gpu_translation_rule, cusolver.trsm)
+
+if rocsolver is not None:
+  xla.backend_specific_translations['gpu'][triangular_solve_p] = \
+      partial(_triangular_solve_gpu_translation_rule, rocsolver.trsm)
 
 # LU decomposition
 
@@ -889,8 +904,13 @@ batching.primitive_batchers[lu_p] = _lu_batching_rule
 xla.backend_specific_translations['cpu'][lu_p] = partial(
   _lu_cpu_gpu_translation_rule, lapack.getrf)
 
-xla.backend_specific_translations['gpu'][lu_p] = partial(
-  _lu_cpu_gpu_translation_rule, cusolver.getrf)
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][lu_p] = partial(
+    _lu_cpu_gpu_translation_rule, cusolver.getrf)
+
+if rocsolver is not None:
+  xla.backend_specific_translations['gpu'][lu_p] = partial(
+    _lu_cpu_gpu_translation_rule, rocsolver.getrf)
 
 xla.backend_specific_translations['tpu'][lu_p] = _lu_tpu_translation_rule
 
@@ -1058,14 +1078,17 @@ def _qr_cpu_gpu_translation_rule(geqrf_impl, orgqr_impl, c, operand,
     q = xops.Pad(r, xops.Constant(c, np.array(0, dtype=shape.element_type())),
                  xla_client.make_padding_config(padding_config))
     q, info_orgqr = orgqr_impl(c, q, tau)
+  if info_geqrf is not None:
+    ok = xops.And(
+      xops.Eq(info_geqrf, xops.ConstantLiteral(c, np.array(0, np.int32))),
+      xops.Eq(info_orgqr, xops.ConstantLiteral(c, np.array(0, np.int32))))
+    q = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), q,
+                             _nan_like(c, q))
+    r = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), r,
+                             _nan_like(c, r))
+  else:
+    pass # rocsolver does not return info
 
-  ok = xops.And(
-    xops.Eq(info_geqrf, xops.ConstantLiteral(c, np.array(0, np.int32))),
-    xops.Eq(info_orgqr, xops.ConstantLiteral(c, np.array(0, np.int32))))
-  q = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), q,
-                           _nan_like(c, q))
-  r = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), r,
-                           _nan_like(c, r))
   r = xla.lower_fun(jnp.triu, multiple_results=False)(c, r)
   return xops.Tuple(c, [q, r])
 
@@ -1080,8 +1103,13 @@ batching.primitive_batchers[qr_p] = qr_batching_rule
 xla.backend_specific_translations['cpu'][qr_p] = partial(
   _qr_cpu_gpu_translation_rule, lapack.geqrf, lapack.orgqr)
 
-xla.backend_specific_translations['gpu'][qr_p] = partial(
-  _qr_cpu_gpu_translation_rule, cusolver.geqrf, cusolver.orgqr)
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][qr_p] = partial(
+    _qr_cpu_gpu_translation_rule, cusolver.geqrf, cusolver.orgqr)
+
+if rocsolver is not None:
+  xla.backend_specific_translations['gpu'][qr_p] = partial(
+    _qr_cpu_gpu_translation_rule, rocsolver.geqrf, rocsolver.orgqr)
 
 
 # Singular value decomposition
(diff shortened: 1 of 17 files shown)


jaxlib/handle_pool.h
```diff
@@ -0,0 +1,107 @@
...
+#include "absl/base/thread_annotations.h"
+#include "absl/synchronization/mutex.h"
+
+namespace jax {
+namespace {
+
+
+// To avoid creating cublas/cusolver contexts in the middle of execution, we maintain
+// a pool of them.
+template <typename handle_t, typename stream_t>
```
hawkinsp, line 30: Stylistically, I'd probably call this `Handle` and `Stream` or `HandleType` and `StreamType`.

(We mostly follow Google C++ style, in which types start with uppercase letters: https://google.github.io/styleguide/cppguide.html#Type_Names)

jaxlib/rocblas.cc
```diff
@@ -0,0 +1,724 @@
+/* Copyright 2019 Google LLC
```
hawkinsp, line 1: My main concern with this file is the amount of code duplication with the cusolver implementation.

I'm wondering if we would do well to introduce a thin interface abstraction over either solver library, but keep the kernel code otherwise shared. This would also allow us to avoid a number of the changes at higher levels: e.g., we wouldn't need to clone the python bindings or change many of the translation rules.

The only significant difference I noticed is that rocsolver doesn't return `info` values in some cases. Presumably we could handle that by returning an `info` value of `0` ("ok")?

What do you think?

inailuig: something like https://github.com/tensorflow/tensorflow/blob/6859f52a3fba6714b5360262f190c9649613ac5c/tensorflow/stream_executor/blas.h ?

I think it would be very nice, however there are quite a few other places where they differ: e.g. in cublas/rocblas there are two separate versions of Getrf. Also I added some more batched versions from rocsolver directly where the cuda verision still uses loops (although it might be possible to add them, I have not checked...). Also in gesvd the buffers work a bit differently.

But I think it should at least be possible to share the python part, adding a few if statements here and there.

jaxlib/handle_pool.h
inailuig: done. 700ac820e2b76b47a09250cc140392a8ebb8564b

jax/_src/lax/linalg.py
```diff
@@ -694,8 +704,14 @@ def _triangular_solve_gpu_translation_rule(
...
                                 transpose)
 
-xla.backend_specific_translations['gpu'][triangular_solve_p] = \
-    _triangular_solve_gpu_translation_rule
+if cusolver is not None:
+  xla.backend_specific_translations['gpu'][triangular_solve_p] = \
+      partial(_triangular_solve_gpu_translation_rule, cusolver.trsm)
+
+if rocsolver is not None:
+  # TODO (inailuig) remove always_use_gpu_kernel=True once complex trsm is added to XLA
```
hawkinsp, line 712: Unless I'm remembering wrong, the XLA triangular solve already fully supports complex arguments?

jaxlib/rocblas.cc
hawkinsp: On reflection, I think the right thing to do here would be to move these kernels into XLA itself, so, e.g., TF can share them. Let's therefore not spend too much time on adding better abstractions around them here.

jax/_src/lax/linalg.py
inailuig: Not yet on ROCm (see: https://github.com/tensorflow/tensorflow/blob/6859f52a3fba6714b5360262f190c9649613ac5c/tensorflow/stream_executor/rocm/rocm_blas.cc#L2432). I think @deven-amd might already be working on adding it (https://github.com/google/jax/issues/2012#issuecomment-742764789)

inailuig: https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/pull/1203/commits/3ceb49c107977b31a5a6dc7552886695ff82d9c8

deven-amd: this change is now in the ROCm fork ( https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/pull/1203 ) and 
a PR has been created to upstream it ( https://github.com/tensorflow/tensorflow/pull/45697 )

jax/_src/lax/linalg.py
```diff
@@ -1166,17 +1195,23 @@ def _svd_cpu_gpu_translation_rule(gesvd_impl, c, operand, full_matrices, compute
...
 
   result = [s]
 
   if compute_uv:
-    u = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), u,
-                             _nan_like(c, u))
-    vt = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), vt,
-                              _nan_like(c, vt))
+    if info is not None:
+        u = _broadcasting_select(c, xops.Reshape(ok, batch_dims + (1, 1)), u,
```
hawkinsp, line 1209: Small nit: 2-space indent. Same everywhere.

jaxlib/rocsolver.py
```diff
@@ -0,0 +1,304 @@
...
+
+# we have a single module for both rocsolver and rocblas functions
+rocsolver_kernels = rocblas_kernels
+
+
+_ops = xla_client.ops
+_Shape = xla_client.Shape
+
+# TODO(phawkins): remove after we no longer need to support old jax releases.
+def _unpack_builder(c):
```
hawkinsp, line 37: You don't need this compatibility code, since this module never existed in ancient `jax` versions.

hawkinsp (commented):
This seems fine.

Do you want to merge it as is or wait for the upstream TF/XLA changes to land?

I'm still a bit sad about the code duplication, but I don't want the perfect to be the enemy of the good, and the better fix would be to add more HLO operators to XLA and remove the need for backend-specific code here.

jaxlib/rocsolver.py
inailuig: https://github.com/google/jax/pull/5115/commits/e51a33c2b7d3dff8bab8baa208d8c1178728f274

jax/_src/lax/linalg.py
inailuig: d07f5c2af7b1daae305eb69443185795268d75ce

inailuig:
> Do you want to merge it as is or wait for the upstream TF/XLA changes to land?

I think we can merge it as is, and I'll just open an other PR for TRSM once that goes through.
do you want me to squash the commits?

hawkinsp:
Yes, please squash the commits.

jax/_src/lax/linalg.py
```diff
@@ -681,7 +691,7 @@ def _triangular_solve_gpu_translation_rule(
     a = xops.Conj(a)
     conjugate_a = False
   if batch > 1 and m <= 32 and n <= 32:
```
inailuig, line 693: @hawkinsp So we are only using the cusolver.trsm for small matrices when they are batched, is that correct?

hawkinsp: Yes. XLA lowers TRSM to the unbatched kernel always; this is just a small refinement over what XLA itself does:
https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/service/gpu/triangular_solve_thunk.cc;l=89

The right end state here would be to delete this code and have XLA emit the batched triangular solve where applicable. I just never quite got to that!

inailuig: And if the matrices are large it does not matter that XLA emits the unbatched version, since the performance is the same ?

hawkinsp: Yes, that's the intuition. It's possible the heuristic could use some tuning.

inailuig: Alright. Thanks for the clarification!

inailuig:
(just moved the formatting of handle_pool.h which ended up in the wrong commit)

hawkinsp (approved):

deven-amd:
just as an FYI,  here is the summary of the unit-test results I get on the ROCm platform (ROCm 3.10 + gfx900 GPU), when I point `org_tensorflow` to the ROCm TF fork, and `jax` to the branch for this PR

```
TESTCASE                                           :     PASSED     FAILED    SKIPPED       DURATION
----------------------------------------------------------------------------------------------------
tests/api_test.py                                  :        315                    18        109.25s
tests/api_util_test.py                             :         12                                1.70s
tests/array_interoperability_test.py               :         10                    51          1.89s
tests/batching_test.py                             :        149                    12        103.51s
tests/callback_test.py                             :          9                                3.91s
tests/core_test.py                                 :        228                               67.66s
tests/custom_object_test.py                        :         14                                7.41s
tests/debug_nans_test.py                           :          4          1                     2.97s
tests/doubledouble_test.py                         :         62                               19.16s
tests/dtypes_test.py                               :         52                               13.04s
tests/errors_test.py                               :          1                     4          1.26s
tests/fft_test.py                                  :         89          1                    22.84s
tests/generated_fun_test.py                        :         24                                8.77s
tests/host_callback_test.py                        :         69                     5         20.92s
tests/image_test.py                                :         42                    20         82.65s
tests/infeed_test.py                               :          3                                2.21s
tests/jax_jit_test.py                              :          4                                1.70s
tests/jax_to_hlo_test.py                           :          2                                1.71s
tests/jaxpr_util_test.py                           :          5                                1.75s
tests/jet_test.py                                  :         57                    14        131.04s
tests/lax_autodiff_test.py                         :        545                              435.36s
tests/lax_control_flow_test.py                     :        268                     1        459.50s
tests/lax_numpy_einsum_test.py                     :        142                               29.99s
tests/lax_numpy_indexing_test.py                   :        318                              129.09s
tests/lax_numpy_test.py                            :       3555          1        421       1531.17s
tests/lax_numpy_vectorize_test.py                  :         28                               14.76s
tests/lax_scipy_sparse_test.py                     :         27                    28         71.21s
tests/lax_scipy_test.py                            :        196                              261.09s
tests/lax_test.py                                  :       1912                     1        538.76s
tests/lax_vmap_test.py                             :        858                    20        861.31s
tests/linalg_test.py                               :        161         27        233        299.75s
tests/loops_test.py                                :         21                     2          7.62s
tests/masking_test.py                              :         77                    45         27.71s
tests/metadata_test.py                             :          3                     1          1.66s
tests/multi_device_test.py                         :                                7          1.65s
tests/multibackend_test.py                         :         17                     6          9.40s
tests/nn_test.py                                   :         56                     2         54.59s
tests/ode_test.py                                  :          2                    11         43.54s
tests/optimizers_test.py                           :         12                    11         14.39s
tests/pmap_test.py                                 :        156                    91         89.43s
tests/polynomial_test.py                           :                               40          1.81s
tests/profiler_test.py                             :          3                     2          1.88s
tests/random_test.py                               :        157                     5        164.49s
tests/scipy_ndimage_test.py                        :         17                                7.56s
tests/scipy_optimize_test.py                       :          5                               16.44s
tests/scipy_signal_test.py                         :         29          1                    38.05s
tests/scipy_stats_test.py                          :        210                              179.84s
tests/sharded_jit_test.py                          :          2                    36          1.79s
tests/stax_test.py                                 :         71                               53.64s
tests/util_test.py                                 :          1                                1.27s
tests/xla_bridge_test.py                           :          5                                1.78s
tests/xmap_test.py                                 :                               10          1.74s
tests/third_party/scipy/line_search_test.py        :          7                               66.47s
----------------------------------------------------------------------------------------------------
TOTAL                                              :      10012         31       1097

```

inailuig:
@deven-amd Thats great! I am assuming of those 27 linalg_test errors 26 are `NotImplementedError: Symmetric (Hermitian) eigendecomposition is not yet implemented in ROCSolver` is that the case? 
(I'll see when I get around to figuring out a nice way of detecting rocm in the tests and skipping these for now.)
On upstream TF (including the 2 pending PR's)  there are still more tests which fail on my gfx803.

deven-amd:
> I am assuming of those 27 linalg_test errors 26 are NotImplementedError: Symmetric (Hermitian) eigendecomposition is not yet implemented in ROCSolver is that the case?

Yes..and the 27th failure is in this subtest `self = <linalg_test.NumpyLinalgTest testMethod=testPinv_shape=float32[1,1]>`

Let me create a branch based on upstream TF + the 2 pending PRs and see what I get on my end.

hawkinsp:
> > I am assuming of those 27 linalg_test errors 26 are NotImplementedError: Symmetric (Hermitian) eigendecomposition is not yet implemented in ROCSolver is that the case?
> 
> Yes..and the 27th failure is in this subtest `self = <linalg_test.NumpyLinalgTest testMethod=testPinv_shape=float32[1,1]>`
> 
> Let me create a branch based on upstream TF + the 2 pending PRs and see what I get on my end.

The correct thing to do is to skip the test in this case. Not everything is supported everywhere, and that's fine.

`test_util.py` has a few utilities for doing this, but they may need changing a little for ROCm (because they conflate "gpu" with "CUDA" right now.)

jaxlib/rocblas.cc
```diff
@@ -0,0 +1,901 @@
...
+using rocBlasHandlePool = HandlePool<rocblas_handle, hipStream_t>;
+
+template <>
+/*static*/ rocBlasHandlePool::Handle rocBlasHandlePool::Borrow(
+    hipStream_t stream) {
+  rocBlasHandlePool* pool = Instance();
+  absl::MutexLock lock(&pool->mu_);
+  rocblas_handle handle;
+  if (pool->handles_.empty()) {
+    ThrowIfErrorStatus(rocblas_create_handle(&handle));
```
inailuig, line 62: @hawkinsp Here a handle is created, but we never destroy it (same for cuda).

hawkinsp: I think this code is correct: we pass ownership of the handle to the `::Handle` class. When the `Handle` is released, ownership is returned to the pool.

We never shrink the pool, so there's no need to release the underlying handles.

We could, I guess, delete the handles in the pool destructor, but note the pool is a global singleton that is itself never deallocated, so it wouldn't change anything.

hawkinsp:
FYI: I needed to make a couple of build changes to get this through our CI. Hopefully they didn't break anything! (in particular, I think I needed to use `rocm_headers` instead of `rocm` in one place.)

deven-amd:
> FYI: I needed to make a couple of build changes to get this through our CI. Hopefully they didn't break anything! (in particular, I think I needed to use `rocm_headers` instead of `rocm` in one place.)

let me do a quick run and verify nothing broke

inailuig:
Just building with the headers does not seem to be enough:
```
In [2]: from jaxlib import rocblas_kernels
ImportError                               Traceback (most recent call last)
<ipython-input-2-e7bd8cdf5c91> in <module>
----> 1 from jaxlib import rocblas_kernels

ImportError: /home/clemens/temp/new/vnewrocm/lib/python3.8/site-packages/jaxlib/rocblas_kernels.so: undefined symbol: rocblas_status_to_string
```

deven-amd:
hmm, build goes through fine, but I get the following error when I try to run any unit test
```
tests/optimizers_test.py:22: in <module>
    import jax.numpy as jnp
jax/__init__.py:22: in <module>
    from .api import (
jax/api.py:38: in <module>
    from . import core
jax/core.py:31: in <module>
    from . import dtypes
jax/dtypes.py:31: in <module>
    from .lib import xla_client
jax/lib/__init__.py:67: in <module>
    from jaxlib import rocsolver  # pytype: disable=import-error
/usr/local/lib/python3.6/dist-packages/jaxlib/rocsolver.py:30: in <module>
    rocsolver_kernels = rocblas_kernels
E   NameError: name 'rocblas_kernels' is not defined
```

inailuig:
I think it might work if we add 
`+        "@org_tensorflow//tensorflow/stream_executor/rocm:rocblas_plugin",`
(and also the same for rocsolver which is still missing from TF) to `rocblas_kernels`

PR #5130, opened 2020-12-07
Upgrade trigonometric functions to primitives.

hawkinsp:
This allows for slightly simpler derivatives than differentiating their implementations.


jakevdp (approved):
Looks good!

Do you know if there is any test coverage for these gradients? I can't find any existing unit tests for these gradients in either `lax_test.py` or `lax_numpy_test.py`.

jakevdp:
Ah, looks like gradients are tested in `lax_autodiff_test.py`.

hawkinsp:
> Ah, looks like gradients are tested in `lax_autodiff_test.py`.

Yup, that's right!

PR #5146, opened 2020-12-09
Implement the new xmap API

apaszke:
This is a stack of three changes, so I'm keeping them as separate commits. I didn't want to go through separate PRs, because subsequent changes would be blocked on LGTMs and that's just annoying.

#### Expose `is_leaf` predicate for `pytree.flatten`
and add tests for it. The change has already been landed in the TF code,
where the C++ pytree components live. This is why I needed to bump the
commit.

#### Make xmap in/out_axes take in native Python containers
instead of forcing the pytree leaves to be the custom container exported
by xmap. This makes the API a bit less verbose, and also relaxes it so
that the mapping can be both specified through a dict (mapping
positional axes to named axes), as well as a list.

All this thanks to the recent pytree changes that let us terminate the
flattening before reaching true leaves.

#### Make vectorization the default xmap schedule
It's been really annoying to have to spell out all the `vectorize`
schedule components, especially when dealing with lots of axes. And the
utility of checking that an axis is _not_ vectorized seems quite
limited.

Finally, one could argue that vectorization really is _the default_.
`xmap` can be seen as a generalization of `einsum`, or as a way to
expose programming with named axes instead of positional axes. In both
those cases, the focus is on shifting how the program is expressed,
without saying anything about the lowering or execution strategy. But,
vectorization is the default of the whole ecosystem, given the emphasis
most libraries put on broadcasting semantics. So it also makes sense to
adopt it here.

### DIFF
--- a/jax/tree_util.py
+++ b/jax/tree_util.py
@@ -39,6 +39,7 @@ for examples.
 import functools
 import collections
 import operator as op
+from typing import Optional, Callable, Any
 
 from .lib import pytree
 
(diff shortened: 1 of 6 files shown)


jax/tree_util.py
```diff
@@ -48,17 +49,23 @@
...
+      flattening step. It should return a boolean, which indicates whether
+      the flattening should traverse the current object, or if it should be
+      stopped immediately, with the whole subtree being treated as a leaf.
 
   Returns:
     A pair where the first element is a list of leaf values and the second
     element is a treedef representing the structure of the flattened tree.
   """
-  return pytree.flatten(tree)
+  # We skip the second argument in support of old jaxlibs
```
hawkinsp, line 66: Leave a TODO comment saying which jaxlib version must be the minimum to allow us to remove the fallback case.

tests/tree_util_tests.py
```diff
@@ -15,11 +15,13 @@
 
 import collections
 
+from unittest import skipIf
 from absl.testing import absltest
 from absl.testing import parameterized
 
 from jax import test_util as jtu
 from jax import tree_util
+from jax.lib import version as jaxlib_version
```
hawkinsp, line 24: Why not just `import jax` and then use `jax.lib.version`? One character longer, and doesn't require defining a new name.

hawkinsp (commented):
XLA version bump LGTM. A couple of quick comments on the pytree change, otherwise it LGTM.

I didn't review the xmap change.

mattjj (approved):
LGTM.

PR #5153, opened 2020-12-09
Expose jit args to accommodate serialization

### LINKED ISSUE #5043: [regression] Pickling a jitted function with cloudpickle
Hi there,

I just stumbled upon a regression between `jax==0.2.5` and `jax==0.2.6`. I make heavy use of cloudpickle (`cloudpickle==1.6.0`) for serializing objects. I do this either directly ([example](https://coax.readthedocs.io/en/latest/coax/utils.html#coax.utils.dump)) or indirectly ([example](https://github.com/ray-project/ray/blob/1d0ade1b936260fc6140be03ec94ace32cd60e27/python/ray/serialization.py#L6)).

I just noticed that I can no longer pickle jitted functions after upgrading from `jax==0.2.5` to `jax==0.2.6`.

Here's a minimal script to reproduce:

```
$ pip install cloudpickle==1.6.0 jax==0.2.5
$ python3 -c "
import jax
import cloudpickle as pickle
@jax.jit
def f(x):
  return x * 13
pickle.dumps(f)
print('OK')
"
OK
```

Same but with later version of jax:
```
$ pip install cloudpickle==1.6.0 jax==0.2.6
$ python3 -c "
import jax
import cloudpickle as pickle
@jax.jit
def f(x):
  return x * 13
pickle.dumps(f)
print('OK')
"

  File "<string>", line 7, in <module>
  File "/home/kris/.local/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/home/kris/.local/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 563, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle 'jaxlib.xla_extension.jax_jit.CompiledFunction' object
```

Related issue: https://github.com/google/jax/issues/679

FWIW This is a blocker for me.

### Some system information:
OS: Ubuntu 20.04.1 LTS
Python: 3.8.5
Other packages: `jaxlib==0.1.57+cuda110`

KristianHolsheimer:
This would expose the original args passed to `jax.jit` as `jitted_func._jit_args` to accommodate serialization of jit-compiled functions along the lines of:

```python
class FunctionApproximator:
  def __init__(self):
    def func(a, b, c, d):
      return a + b + c + d
    self.func = jax.jit(func, static_argnums=(1, 2), donate_argnums=3)

  def __getstate__(self):
    jit_args = self.func._jit_args
    return (jit_args.fun, jit_args.static_argnums, jit_args.donate_argnums)

  def __setstate__(self, state):
    f, s, d = state
    self.func = jax.jit(f, static_argnums=s, donate_argnums=d)

```

closes #5043

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -181,6 +181,14 @@ def jit(fun: Callable[..., T],
     return _python_jit(fun, static_argnums, device, backend, donate_argnums)
 
 
+class _JitArgs(NamedTuple):
+  fun: Callable
+  static_argnums: Iterable[int]
+  device: Optional[xc.Device]
+  backend: Optional[str]
+  donate_argnums: Iterable[int]
+
+
 def _python_jit(
     fun: Callable,
     static_argnums: Union[int, Iterable[int]] = (),
--- a/jax/api.py
+++ b/jax/api.py
@@ -192,6 +192,7 @@ def _python_jit(
   _check_callable(fun)
   static_argnums = _ensure_index_tuple(static_argnums)
   donate_argnums = _ensure_index_tuple(donate_argnums)
+  jit_args = _JitArgs(fun, static_argnums, device, backend, donate_argnums)
   donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)
 
   @wraps(fun)
@@ -225,6 +226,7 @@ def _python_jit(
         donated_invars=donated_invars)
     return tree_unflatten(out_tree(), out)
 
+  f_jitted._jit_args = jit_args  # expose jit args for serialization
   return f_jitted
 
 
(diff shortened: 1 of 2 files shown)


KristianHolsheimer:
@hawkinsp I just rebased on `master`. By the way, if you're not keen on merging this, don't worry about it. Just let me know. I can handle it downstream.

KristianHolsheimer:
Hi all, returning to this after a while, I think it might be better to handle this downstream instead.

In my case, I added an alternative `jit` util that supports serialization [here](https://github.com/microsoft/coax/blob/main/coax/utils/_jit.py).

Thanks for looking into this. Will close this PR now.

PR #5156, opened 2020-12-10
jnp.arccosh: use same complex branch as numpy

### LINKED ISSUE #5154: jnp.acosh and np.acosh follow different conventions
The following code outputs different results

```python
import numpy as np
from jax import numpy as jnp

x = -4.69577+2.1128223j

print(np.arccosh(x))
print(jnp.arccosh(x))
```

While there are infinitely many possible results, numpy defines a convention in its documentation for what the result should be:
https://numpy.org/doc/stable/reference/generated/numpy.arccosh.html

I do not know if it is that important, but figured I would report since we expect `jnp` and `np` to have the same behavior :)

jakevdp:
Fixes #5154

Confirmed that the modified arccosh test fails without the updated function; I modified the arcsinh test because it's potentially subject to the same issue and lacked test coverage.


tests/lax_numpy_test.py
```diff
@@ -180,10 +180,10 @@ def op_record(name, nargs, dtypes, shapes, rng_factory, diff_modes,
...
-    op_record("arcsinh", 1, number_dtypes, all_shapes, jtu.rand_positive, ["rev"],
-              inexact=True),
-    op_record("arccosh", 1, number_dtypes, all_shapes, jtu.rand_positive, ["rev"],
-              inexact=True),
+    op_record("arcsinh", 1, number_dtypes, all_shapes, jtu.rand_default, ["rev"],
+              inexact=True, tolerance={np.complex64: 2E-6, np.complex128: 2E-14}),
+    op_record("arccosh", 1, number_dtypes, all_shapes, jtu.rand_default, ["rev"],
+              inexact=True, tolerance={np.complex64: 2E-2, np.complex128: 2E-12}),
     op_record("arctanh", 1, number_dtypes, all_shapes, jtu.rand_small, ["rev"],
               inexact=True, tolerance={np.float64: 1e-9}),
```
bchetioui, line 188: Is `arctanh` free of that problem, or should we also perform a similar check?

bchetioui (approved):
LGTM!

tests/lax_numpy_test.py
jakevdp: arctanh may have that problem, but `jtu.rand_small` already tests all phase angles in the complex plane. The issue with the arcsinh and arccosh tests was that they were only testing inputs with positive real and imaginary parts.

jakevdp: And `rand_small` is more appropriate than `rand_default`, I think, because of the domain for which arctanh is defined for real inputs.

PR #5161, opened 2020-12-11
[jax2tf] Added a non-XLA conversion path for reduce_window_sum

qiuminxu:
PR#4589 added a conversion path for reduce_window_max use tf.nn.max_pool.
This PR make the function broader to also apply to reduce_window_sum (which is used by flax.nn.avg_pool).

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1596,24 +1596,30 @@ def _reduce_window(operand, init_value, *, jaxpr, consts, window_dimensions,
       base_dilation, window_dilation, _in_avals, _out_aval
   )
 
-# _try_tf_max_pool returns a Tensor when it succeeds, or a string describing why
-# it did not succeed otherwise.
-def _try_tf_max_pool(operand, window_dimensions, window_strides, padding,
-                     base_dilation, window_dilation) -> Union[str, TfVal]:
-  # TODO(bchetioui): this function is not exhaustive wrt which
-  # reduce_window_max cases can be translated into a call to max_pool. Further
-  # investigation is needed to fully flesh it out.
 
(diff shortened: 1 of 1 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1596,24 +1596,30 @@ def reducer(arg1: TfVal, arg2: TfVal) -> TfVal:
...
   # Contrarily to the main path, tf.int8 is actually a valid type for
   # tf.nn.max_pool.
-  if operand.dtype in [tf.bool, tf.uint32, tf.uint64, tf.complex64,
-                       tf.complex128]:
+  if op_name == "reduce_window_max" and operand.dtype in [
+      tf.bool, tf.uint32, tf.uint64, tf.complex64, tf.complex128
+  ]:
     return f"tf.nn.max_pool does not support operands of type {operand.dtype}"
+  if op_name == "reduce_window_sum" and operand.dtype not in [
+      tf.half, tf.float16, tf.float32, tf.float64
```
bchetioui, line 1615: I don't think there is a difference between `tf.half` and `tf.float16`:
```
>>> tf.half is tf.float16
True
```
I think we could just use `tf.float16`? :)

bchetioui (approved):
Passes the tests otherwise, LGTM once the `tf.half`/`tf.float16` thing is fixed :)

jax/experimental/jax2tf/jax2tf.py
gnecula: I would also prefer tf.float16; I do not think we have any uses of tf.half in the code base.

gnecula (approved):

jax/experimental/jax2tf/jax2tf.py
qiuminxu: Yep, I missed this. Done.

PR #5163, opened 2020-12-11
[jax2tf] Factorize and clean up test utils.

bchetioui:
This PR splits our jax2tf test utils in more reusable blocks for further reuse in other settings (e.g. benchmarking, shape polymorphism checks, etc…).

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -27,6 +27,7 @@ from jax.api_util import flatten_fun
 from jax.interpreters import ad, batching
 from jax.interpreters import masking
 from jax.interpreters import pxla
+from jax.interpreters import sharded_jit
 from jax.interpreters import xla
 from jax._src.lax import lax
 from jax._src.lax import linalg as lax_linalg
@@ -840,6 +841,8 @@ tf_not_yet_impl = [
   lax_parallel.axis_index_p, lax_parallel.pdot_p,
 
   pxla.xla_pmap_p,
+
+  sharded_jit.sharding_constraint_p
 ]
 
 try:
--- a/jax/experimental/jax2tf/tests/correctness_stats.py
+++ b/jax/experimental/jax2tf/tests/correctness_stats.py
@@ -40,7 +40,6 @@ Limitation = NamedTuple("Limitation", [ ("primitive_name", str)
                                       , ("devices", Tuple[str,...])
                                       ])
 
-CATEGORY_POSSIBLE_INCORRECT_RESULTS = "Possible incorrect results"
 CATEGORY_MISSING_TF_SUPPORT = "Missing TF support"
 
 def categorize(prim: core.Primitive, *args, **kwargs) \
@@ -76,11 +75,6 @@ def categorize(prim: core.Primitive, *args, **kwargs) \
     if additional_msg:
       msg += '; ' + additional_msg
     _report_failure(CATEGORY_MISSING_TF_SUPPORT, msg, np_dtype, devs=devs)
-
-  def tf_possible_incorrect(np_dtype: Optional[NpDType] = None,
-                            msg: str = "",
-                            devs: Sequence[str] = all_devices) -> None:
-    _report_failure(CATEGORY_POSSIBLE_INCORRECT_RESULTS, msg, np_dtype, devs=devs)
 
   def _to_np_dtype(dtype) -> NpDType:
     try:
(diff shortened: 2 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -840,6 +841,8 @@ def _unexpected_primitive(p: core.Primitive, *args, **kwargs):
   lax_parallel.axis_index_p, lax_parallel.pdot_p,
 
   pxla.xla_pmap_p,
+
+  sharded_jit.sharding_constraint_p
```
bchetioui, line 845: I think [this commit](https://github.com/google/jax/commit/a04f8197cae1434840f9badc5d931250bc80a153) broke my tests locally. I guess it does not happen in google3, but just to be sure… :)

jax/experimental/jax2tf/tests/correctness_stats.py
```diff
@@ -77,11 +76,6 @@ def tf_unimpl(np_dtype: Optional[NpDType] = None,
       msg += '; ' + additional_msg
     _report_failure(CATEGORY_MISSING_TF_SUPPORT, msg, np_dtype, devs=devs)
 
-  def tf_possible_incorrect(np_dtype: Optional[NpDType] = None,
-                            msg: str = "",
-                            devs: Sequence[str] = all_devices) -> None:
-    _report_failure(CATEGORY_POSSIBLE_INCORRECT_RESULTS, msg, np_dtype, devs=devs)
```
bchetioui, line 83: I don't think we need this anymore, we have not used it for any of the primitives, as we have documented this directly in the .md file.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -40,6 +40,40 @@
   atexit.register(correctness_stats.pprint_all_limitations,
                   output_file, template_file)
 
+
+def _make_tf_args(args):
+  def _convert_if_bfloat16(v):
```
bchetioui, line 45: These utils are not tied to ConvertAndCompare, or to the test class. It makes more sense to have them out here for reuse, I think.

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -73,10 +110,80 @@ def to_numpy_dtype(dt):
...
+
+    def restore_get_primitive_impl():
+      jax2tf.jax2tf.TensorFlowTrace.get_primitive_impl = original_impl
+
+    # Restore the original jax2tf.TensorFlowTrace.get_primitive_impl
+    # implementation at the end of the test.
+    self.addCleanup(restore_get_primitive_impl)
+
+  def RunAndDetectLimitations(self, func_tf: Callable, *tf_args, mode: str):
+    """
```
bchetioui, line 135: That way we can use this for other tests on jax2tf. I wanted to experiment more with this change before making the PR, but I am not sure if I will have the time, hence why I am making it now :)

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -57,6 +91,9 @@ def setUp(self):
       self.assertEqual(jtu.device_under_test().upper(),
                        self.tf_default_device.device_type)
 
+    self._collect_limitations = False
+    self._original_impl: Callable = None
```
gnecula, line 95: This does not seem to be used

jax/experimental/jax2tf/tests/tf_test_util.py
```diff
@@ -73,10 +110,80 @@ def to_numpy_dtype(dt):
       self.assertEqual(to_numpy_dtype(jtu._dtype(x)),
                        to_numpy_dtype(jtu._dtype(y)))
 
+  def _enable_limitations_collection(self):
+    # No need to monkey patch again
+    if self._collect_limitations:
```
gnecula, line 115: This test will not work if we have patched already but have not collected any limitations, right?

gnecula (approved):

jax/experimental/jax2tf/tests/tf_test_util.py
bchetioui: Hm, I think it will not work at all since I should set `self._collect_limitations` to True below! I fixed it in a new commit, good catch!

jax/experimental/jax2tf/tests/tf_test_util.py
bchetioui: Removed.

gnecula (approved):

PR #5165, opened 2020-12-11
[BUG] Associative scan regression

AdrienCorenflos:
This PR now only includes a test to reproduce a problem which arises when using linalg.solve or linalg.lu from within the associative scan operation.

### DIFF
--- a/tests/lax_control_flow_test.py
+++ b/tests/lax_control_flow_test.py
@@ -39,7 +39,6 @@ import jax.numpy as jnp  # scan tests use numpy
 import jax.scipy as jsp
 
 from jax.config import config
-
 config.parse_flags_with_absl()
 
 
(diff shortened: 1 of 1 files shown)


hawkinsp:
The issue should now be fixed; I added a more specific unit test. If you wanted to add this test as well as a regression test, that'd be great! You'd need to clean up the PR a little bit, e.g., currently it reformats the whole test file; it should only add a new test case and not touch unrelated lines.

AdrienCorenflos:
Thanks for the fix Peter. Indeed, I configured auto pep8 on commit in PyCharm, didn't think about the fact JAX doesn't follow it. I'll need to revert and deactivate it prior to committing again. Apologies!

AdrienCorenflos:
Merged and updated

hawkinsp:
I slightly improved this test case in https://github.com/google/jax/pull/6741, by (a) squashing the commits, and (b) reducing the number of test cases. ~100 test cases is too many.

(I'm assuming you won't mind me merging it via my own PR, it seemed easier than going back and forth on tiny improvements to a test.)

PR #5171, opened 2020-12-11
Check for non-tuples and not int subclasses in Chunked

apaszke:
Apparently some projects like to pass in instances of `numpy.int64`
where `int`s are expected, and those fail the subclass check. This
should be a hotfix for them, though it would be good to figure out where
does the NumPy scalar come from, and make them well typed.

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -83,7 +83,7 @@ class Chunked:
   chunks: Tuple[int, ...]
 
   def __init__(self, chunks: Union[int, Tuple[int, ...]]):
-    if isinstance(chunks, int):
+    if not isinstance(chunks, tuple):
       chunks = (chunks,)
     object.__setattr__(self, 'chunks', chunks)
 


jakevdp:
post-merge review: I think before this change, passing a list of chunks would work; after this change it would break. In other places we check `isinstance(object, typing.Sequence)` - that may be a better choice here. What do you think?

apaszke:
This was a big of a hotfix and it's already 10pm on Friday here, so this is why I skipped the review 😛 But I'm not sure if I follow your argument. I definitely never want this to be a list, and I do have a bunch of type annotations that assert that the number of chunks should always be a tuple, mostly to guarantee immutability. Also, this is purely an internal API, so it should be completely up to us. The problem is that we do tend to use internal APIs in internal projects and this is what tripped up my previous patch (that project didn't confirm to my type signature).

jakevdp:
Ah, thanks. If it's an internal API & we're sure no users of the internal API are instantiating `Chunked()` with a list, then this is probably fine.

PR #5174, opened 2020-12-11
handle weak types in all lax primitives

jakevdp:
This changes the default weak type rule such that outputs match the inputs; this seems to be the appropriate default behavior, and there are only a few standard primitives that require something different.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -463,11 +463,7 @@ def bitcast_convert_type(operand: Array, new_dtype: DType) -> Array:
     `new_dtype`.
   """
   new_dtype = dtypes.canonicalize_dtype(new_dtype)
-  old_dtype = _dtype(operand)
-  if old_dtype != new_dtype:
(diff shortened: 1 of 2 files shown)


mattjj (approved):
The code LGTM, but is there a way to add some test coverage for any new behaviors? Without test coverage it's hard to tell if these changes have the desired effect (and what that effect is), and also it risks accidentally regressing things in the future.

jakevdp:
We could add test coverage, but it would be a bespoke test for each lax function. There is already bulk coverage of desired behavior for standard unary and binary operations.

Maybe a new test suite class within `lax_test.py`?

mattjj:
Hmm... since AIUI there are only two operative functions (`_strip_weak_type` and `_argnum_weak_type`), how about we just test one primitive for each? That is, no need to be exhaustive, perhaps?

jakevdp:
I added some tests that, while not comprehensive, cover the bulk of the new code paths.

PR #5176, opened 2020-12-12
api: handle numpy integers for static argnums

jakevdp:
Currently, the JAX API doesn't play well with numpy ints. For example:
```python
In [1]: from jax import jit                                                                                                                                                    

In [2]: import numpy as np                                                                                                                                                     

In [3]: jit(lambda x: x, static_argnums=0)(1)                                                                                                                                  
Out[3]: DeviceArray(1, dtype=int32)

In [4]: jit(lambda x: x, static_argnums=np.int32(0))(1)                                                                                                                        
<ipython-input-5-dab63edd05c3> in <module>
----> 1 jit(lambda x: x, static_argnums=np.int32(0))(1)

~/github/google/jax/jax/api.py in jit(fun, static_argnums, device, backend, donate_argnums)
    176   """
    177   if FLAGS.experimental_cpp_jit and config.omnistaging_enabled:
--> 178     return _cpp_jit(fun, static_argnums, device, backend, donate_argnums)
    179   else:
    180     return _python_jit(fun, static_argnums, device, backend, donate_argnums)

~/github/google/jax/jax/api.py in _cpp_jit(fun, static_argnums, device, backend, donate_argnums)
    249   """
    250   _check_callable(fun)
--> 251   static_argnums = _ensure_tuple(static_argnums)
    252   donate_argnums = _ensure_tuple(donate_argnums)
    253   donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)

~/github/google/jax/jax/api.py in _ensure_tuple(x)
   2581 
   2582 def _ensure_tuple(x: Union[int, Iterable[int]]) -> Tuple[int, ...]:
-> 2583   return (x,) if isinstance(x, int) else tuple(x)
   2584 
   2585 def invertible(fun: Callable) -> Callable:

TypeError: 'numpy.int32' object is not iterable
```
This PR updates the api to handle any index-compatible scalar input for `static_argnums` or `dyn_argnums`.

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -26,6 +26,7 @@ import collections
 import functools
 import inspect
 import itertools as it
+import operator
 import threading
 import weakref
 from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Sequence, Tuple, TypeVar, Union
@@ -43,7 +44,7 @@ from .core import eval_jaxpr
 from .api_util import (flatten_fun, apply_flat_fun, flatten_fun_nokwargs,
                        flatten_fun_nokwargs2, argnums_partial,
                        argnums_partial_except, flatten_axes, donation_vector,
-                       rebase_donate_argnums)
+                       rebase_donate_argnums, _ensure_index, _ensure_index_tuple)
 from ._src import traceback_util
 from ._src.traceback_util import api_boundary
 from .tree_util import (tree_map, tree_flatten, tree_unflatten, tree_structure,
@@ -189,8 +190,8 @@ def _python_jit(
 ) -> Callable:
   """The Python implementation of `jax.jit`, being slowly replaced by _cpp_jit."""
   _check_callable(fun)
-  static_argnums = _ensure_tuple(static_argnums)
-  donate_argnums = _ensure_tuple(donate_argnums)
+  static_argnums = _ensure_index_tuple(static_argnums)
+  donate_argnums = _ensure_index_tuple(donate_argnums)
   donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)
 
   @wraps(fun)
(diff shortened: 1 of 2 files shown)


jax/api.py
```diff
@@ -762,11 +763,15 @@ def value_and_grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
             "same shape as the arguments at positions {argnums}.")
 
   _check_callable(fun)
+  if np.isscalar(argnums):
```
hawkinsp, line 766: I'm curious if this would be better written as:
```
try:
  argnums = (operator.index(argnums),)
except TypeError:
  argnums = tuple(map(operator.index, argnums))
```

i.e., if `operator.index` is ultimately the judge of whether we consider something an index or not, why not just call it first rather than first looking at `np.isscalar`, which may or may not coincide?

Same everywhere.

jax/api.py
```diff
@@ -868,6 +873,10 @@ def jacfwd(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
    [ 1.6209   0.       0.84147]]
   """
   _check_callable(fun)
+  if np.isscalar(argnums):
```
hawkinsp, line 876: Rather than repeating this code, put it in a helper? I'd rather not duplicate parsing code.

jax/api.py
hawkinsp: Why not just call `_ensure_index_tuple`?

jakevdp: Because in this instance, the code further down branches on whether the input was a tuple or a scalar.

jax/api.py
jakevdp: As above, the code furthur down branches on whether the input is a tuple or a scalar, so we can't convert it to a tuple here.,

jax/api.py
hawkinsp: But hang on, it doesn't do so in a meaningful way, as best I can tell? i.e., we could canonicalize it to a tuple and then remove the branch below? The only reason I can see not to do that might be to avoid a tuple allocation.

jakevdp:
Ready for another review... I made a second utility routine and shared both between `api.py` and `api_util.py` to avoid code duplication.

hawkinsp (approved):

PR #5180, opened 2020-12-13
Fix a typo on the dynamic definition of __hash__.

jblespiau:


hawkinsp (commented):
Seems fine, but are we missing a test?

jblespiau:
> Seems fine, but are we missing a test?

Indeed, done.

hawkinsp (approved):

hawkinsp:
There is a test failure.

jblespiau:
Thanks, I fixed it (it was because jaxlib is not released yet, so `Buffer` does not behave like a DeviceArray just yet.

Issue #5134, opened 2020-12-08
jax.checkpoint prevents id_tap from running


sbodenstein:
This should, but doesn't print:
```
import jax
from jax.experimental import host_callback

@jax.checkpoint
def f(x):
  host_callback.id_tap(print, x)
  return x

jax.grad(f)(2.)
```
It does print if either:
- `@jax.checkpoint` is removed
- the output of host_callback is used

gnecula:
The best way to use `id_tap` is to use its result. This fits best with the functional nature of JAX, because internally JAX is not aware that `id_tap` has side-effects. This can make JAX sometime eliminate the operation as it appears to not be needed. This seems to happen during the analysis that is part of implementing `jax.checkpoint`. If you add a `jax.make_jaxpr` you will see that the `jax.grad(f)` is the constant function `lambda _: 1.`. 

If you use the result of `id_tap` then you will also be able to print the cotangent of the output.

sbodenstein:
The [documentation states that](https://github.com/google/jax/blob/master/jax/experimental/host_callback.py#L45):
> As of September 2020, it is not necessary anymore for the results of the tap to be used in the rest of the computation. The tap function will execute based on program order. 
If this is correct, then surely JAX must be aware that `id_tap` has side-effects, and cannot simply eliminate it if it wants?

gnecula:
I will add a caveat to the documentation.

gnecula:
I have clarified the documentation: id_tap that does not use its results may be eliminated by partial evaluation; if in doubt, use the result. Perhaps a better solution is possible, to make JAX aware of side-effects, but right now that seems a low priority.

gnecula:
I have re-opened this issue, we should try to fix this.

hawkinsp:
id_tap is removed from JAX, so this issue should be moot.

PR #5182, opened 2020-12-13
[host_callback] Add support for pmap and for passing the device to tap

### LINKED ISSUE #5169: jax.pmap causes multiple id_tap callbacks to run
Consider the following:
```
@jax.pmap
def f(x):
  host_callback.id_tap(print, x)
  return x**2
f(jnp.arange(8))

# Prints:
0 ()
1 ()
2 ()
4 ()
5 ()
3 ()
7 ()
6 ()
```
So you get 8 seperate calls to id_tap, and no information is given that you are in a `pmap`. If I switch to `jax.vmap` in the above, you get a single call to id_tap and the print: `[0 1 2 3 4 5 6 7] (('batch', {'batch_dims': (0,)}),)`

There is no test for how `id_tap` should work with `pmap`, other than that the output should be correct. So not sure what the expected behaviour is.

gnecula:
* Adds support for jit of pmap and pmap of pmap.
* Also add a `tap_with_device` optional argument to `id_print` and
  `id_tap`, to have the tap function invoked with a device keyword argument.

In presence of `pmap` there will be multiple devices sending data to the host. Each such
tap will be processed separately (and there may be interleaving between devices). To make
the output more understandable, we added an option to pass the device to the tap function.
For backwards-compatibility, the device is passed as a keyword argument to `id_tap` and only if
the `tap_with_device` optional flag is passed to `id_tap` or `id_print`. 

Issue: #5134
Fixes: #5169

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -590,7 +590,7 @@ def _hashable_index(idx):
 # TODO(skye): is there a simpler way to rewrite this using sharding_spec?
 def _shard_sharded_device_array_slow_path(x, devices, indices):
   candidates = defaultdict(list)
-  for buf, idx in zip(x.device_buffers, x.indices):
+  for buf, idx in safe_zip(x.device_buffers, x.indices):
     candidates[_hashable_index(idx)].append(buf)
 
   bufs = []
@@ -690,7 +690,7 @@ def parallel_callable(fun: lu.WrappedFun,
 
   if config.omnistaging_enabled:
     sharded_avals = tuple(shard_aval(axis_size, axis, aval) if axis is not None else aval
-                          for axis, aval in zip(in_axes, avals))
+                          for axis, aval in safe_zip(in_axes, avals))
     if any(s is not None for s in global_arg_shapes):
       # TODO(skye): we could take this branch unconditionally if we handled
       # grad of global_arg_shapes correctly.
@@ -712,7 +712,7 @@ def parallel_callable(fun: lu.WrappedFun,
         return fun.call_wrapped(*args)
 
     sharded_avals = tuple(shard_aval(axis_size, axis, aval) if axis is not None else aval
-                          for axis, aval in zip(in_axes, avals))
+                          for axis, aval in safe_zip(in_axes, avals))
     pvals = [pe.PartialVal.unknown(aval) for aval in sharded_avals]
     # We add a dummy first invar, to carry the trace  details to `dynamic_fun`
     pval = pe.PartialVal.unknown(core.abstract_unit)  # dummy value for axis env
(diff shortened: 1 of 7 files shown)


jax/interpreters/pxla.py
```diff
@@ -590,7 +590,7 @@ def _hashable_index(idx):
 # TODO(skye): is there a simpler way to rewrite this using sharding_spec?
 def _shard_sharded_device_array_slow_path(x, devices, indices):
   candidates = defaultdict(list)
-  for buf, idx in zip(x.device_buffers, x.indices):
+  for buf, idx in safe_zip(x.device_buffers, x.indices):
```
gnecula, line 593: I made these changes of `zip` -> `safe_zip` because while debugging my changes `zip` was failing silently.

shoyer:
What do you think about treating `pmap` as just another transformation from the perspective of host_callback? E.g., adding an entry to the sequence of transformations, like `('pmap', {'device': ...})`?

From a user perspective, I guess this might be a little more expected and would avoid needing to change the signature of host callback. On the other hand, `device` is part of how a computation is run (you can't nest `pmap` with different devices).

gnecula:
I thought about adding pmap to the list of transforms but it is not easy at all, because pmap is not really a transformation. It is more like jit than like vmap. One of the test cases I have added is a jvp of pmap of vmap, specifically to explore this. By the time I can intervene the code is transformed (with jvp of vmap). (The same would happen if we had jvp + jit + vmap).

sbodenstein:
@gnecula: for the logging usecase, we would like the data in the form @shoyer mentioned. But there is not enough info to get it into this form with the current setup, even with the device information. Consider:
```
@jax.pmap
@jax.vmap
def f(x):
  x = host_callback.id_tap(print, x)
  return x**2
```
If we switch the `vmap` and `pmap`, the output of the function is the same. But in the one case the prints are 
```
[0 0] (('batch', {'batch_dims': (0,)}),)
[1 1] (('batch', {'batch_dims': (0,)}),)
```
and for the other it is 
```
[0 1] (('batch', {'batch_dims': (0,)}),)
[0 1] (('batch', {'batch_dims': (0,)}),)
```

gnecula:
Unfortunately, I cannot surface precisely in the printing that the user had a `vmap(pmap(f))`, because that information is lost by the time we get to execute the code. What is printed is what actually executes on each device. Probably most users would not understand that `vmap(pmap(f))` is identical to `pmap(vmap(f), in_axes=1)`. In other words, `pmap` is always moved to the top level of the transformation stack, taking care of adjusting `in_axes`. The new change to printing can be used to explore that. In fact, I have added a `test_pmap_vmap` and `test_vmap_pmap`. 

  

I suspect that most users would quite understand that what `vmap(pmap

shoyer:
Right, `pmap` is really just sugar for launching the same program on multiple devices at the same time. I guess this distinction is lost by the time host_callback runs?

tests/host_callback_test.py
```diff
@@ -103,6 +128,40 @@ def repl_func(match_group):
...
+      expected = expected[0:start_device_1]
+
+  def replace_device_name(m) -> str:
+    return str(api.devices()[int(m.group(1))])
+  expected = re.sub(r'cpu:(\d+)', replace_device_name, expected)
+  what = testing_stream.output_sorted_by_device
+  return assertMultiLineStrippedEqual(tst, expected, what)
+
+
+# Run all tests with 8 CPU devices.
```
shoyer, line 154: is this a typo? It looks like it's always 2 CPU devices?

shoyer (approved):

tests/host_callback_test.py
gnecula: A typo indeed, fixed. Thanks!

PR #5183, opened 2020-12-13
Typo ?

AntoinePlumerault:
I removed "-e" option from "pip install -e dist/*.whl  # installs jaxlib (includes XLA)" line 58. It is now coherent with lines 69-70. 
When I tried the command with the "-e" it threw an error, without "-e" it worked fine.


hawkinsp (approved):
Thanks! Yes, the `-e` was left over from an older release when we didn't build a `.whl` file.

PR #5187, opened 2020-12-14
[jax2tf] Expand coverage for unary elementwise ops.

bchetioui:
This fixes some conversion quirks and documents the limitations of the unary elementwise ops with previously insufficient coverage.


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -852,7 +852,18 @@ def _unexpected_primitive(p: core.Primitive, *args, **kwargs):
 tf_impl[xla.device_put_p] = lambda x, device=None: x
 
 tf_impl[lax.neg_p] = tf.math.negative
-tf_impl[lax.sign_p] = tf.math.sign
+
+def _sign(x):
```
gnecula, line 856: I think that for now we should avoid adding more casts. We need to decide how to handle more systematically the unimplemented dtypes

jax/experimental/jax2tf/tests/primitives_test.py
```diff
@@ -512,60 +512,240 @@ def test_linear_solve(self, harness: primitive_harness.Harness):
...
+                   lax.asinh_p: np.sinh,
+                   lax.atanh_p: np.tanh,
+                   lax.acos_p: np.cos,
+                   lax.asin_p: np.sin,
+                   lax.atan_p: np.tan}[harness.params["prim"]]
+    custom_assert = None
+    operand, = harness.dyn_args_maker(self.rng())
+    if harness.params["dtype"] in [np.complex64, np.complex128]:
+      def custom_assert(result_jax, result_tf):
+        # There are technically infinitely many possible results, so we check
```
gnecula, line 531: I think that we ought to document that JAX and TF may return different answers for these ops.

bchetioui: In that case, maybe we should do the same for `eig` and the likes, and every function that may return different but equally valid results. WDYT?

gnecula: Yes, that is a good idea. It would be good to document those ops for which we may get a different result. These should be exactly those that have custom asserts, no?

bchetioui: I think this is true, yes.

gnecula: Should we add those limitations in this PR, or a future one?

bchetioui: I just added it here! Sorry for the delay :)

PR #5189, opened 2020-12-14
Improve a few JVP rules with rewrites from Herbie.

hawkinsp:


jakevdp:
How do efforts like this interact with XLA's optimization of orders of operations & related rewrites?

jakevdp (approved):

hawkinsp:
It's hard to say without looking harder at XLA. However, I think these do no harm for the moment (even if XLA undoes them!) and we can look into any precision-reducing optimizations in XLA separately.

PR #5191, opened 2020-12-15
Fix colab_tpu.setup_tpu import in example Cloud TPU notebooks.

skye:


yashk2810:
Will this fix work outside of colab?

The kokoro builds/CI that test these notebooks don't run inside colab. They run inside their own environment.

skye:
I believe so. This fix isn't colab-specific, it's just a python path thing. Time will tell for sure though :)

PR #5194, opened 2020-12-15
[jax2tf] Added utils to generated stats on generated HLO IR (WIP)

bchetioui:
Opening as draft so we can discuss this.
Maybe measuring the size at the character level is not so helpful without deeper investigation, as shown by the difference in the sizes in the resulting file.
Part of it is simply that TF generated code seems to have more metadata, as you can see in the example below:

```
>>> print(tf.function(lambda x: x, experimental_compile=True).experimental_get_compiler_ir(1)())
HloModule a_inference__lambda__28__.5

ENTRY %a_inference__lambda__28__.5 () -> s32[] {
  %constant.1 = s32[] constant(1)
  %reshape.2 = s32[] reshape(s32[] %constant.1), metadata={op_name="XLA_Retvals"}
  %tuple.3 = (s32[]) tuple(s32[] %reshape.2), metadata={op_name="XLA_Retvals"}
  ROOT %get-tuple-element.4 = s32[] get-tuple-element((s32[]) %tuple.3), index=0, metadata={op_name="XLA_Retvals"}
}
```

vs

```
>>> print(jax.xla_computation(lambda x: x)(1).as_hlo_text())
HloModule xla_computation__lambda___6.4

ENTRY xla_computation__lambda___6.4 {
  parameter.1 = s32[] parameter(0)
  ROOT tuple.3 = (s32[]) tuple(parameter.1)
}
```

I haven't looked yet into whether it is possible to strip it. It might be enough to count the number of lines instead, which should be closer!


bchetioui:
It seems that TF constant folded the parameter when it was a scalar. Using an otherwise shaped array gives something like this:

```
>>> print(tf.function(lambda x: x, autograph=False, experimental_compile=True).experimental_get_compiler_ir(np.ones((2, 3)))())
HloModule a_inference__lambda__76__.6

ENTRY %a_inference__lambda__76__.6 (arg0.1: f64[2,3]) -> f64[2,3] {
  %arg0.1 = f64[2,3]{1,0} parameter(0), parameter_replication={false}, metadata={op_name="XLA_Args"}
  %reshape.2 = f64[2,3]{1,0} reshape(f64[2,3]{1,0} %arg0.1)
  %reshape.3 = f64[2,3]{1,0} reshape(f64[2,3]{1,0} %reshape.2), metadata={op_name="XLA_Retvals"}
  %tuple.4 = (f64[2,3]{1,0}) tuple(f64[2,3]{1,0} %reshape.3), metadata={op_name="XLA_Retvals"}
  ROOT %get-tuple-element.5 = f64[2,3]{1,0} get-tuple-element((f64[2,3]{1,0}) %tuple.4), index=0, metadata={op_name="XLA_Retvals"}
```

vs

```
>>> print(jax.xla_computation(lambda x: x)(np.ones((2, 3))).as_hlo_text())
HloModule xla_computation__lambda___7.4

ENTRY xla_computation__lambda___7.4 {
  parameter.1 = f32[2,3]{1,0} parameter(0)
  ROOT tuple.3 = (f32[2,3]{1,0}) tuple(parameter.1)
}
```

hawkinsp:
> I haven't looked yet into whether it is possible to strip it.

 TF is simply choosing to print more of its metadata by default. I don't know whether it's easy to make TF print without metadata, but you can make JAX print with metadata (also a semiprivate API):

```
In [43]: options = jax.lib.xla_client._xla.HloPrintOptions()

In [44]: options.print_metadata = True

In [45]: print(jax.xla_computation(lambda x: x + 2)(1).as_hlo_module().to_string(options))
HloModule xla_computation__lambda___12.6

ENTRY %xla_computation__lambda___12.6 (parameter.1: s32[]) -> (s32[]) {
  %constant.2 = pred[] constant(false)
  %parameter.1 = s32[] parameter(0)
  %constant.3 = s32[] constant(2), metadata={op_type="add" op_name="xla_computation(<lambda>)/add" source_file="<ipython-input-45-380200ef996d>" source_line=1}
  %add.4 = s32[] add(s32[] %parameter.1, s32[] %constant.3), metadata={op_type="add" op_name="xla_computation(<lambda>)/add" source_file="<ipython-input-45-380200ef996d>" source_line=1}
  ROOT %tuple.5 = (s32[]) tuple(s32[] %add.4)
}
```

Note I slightly change the example to contain an operator, not just a constant.

hawkinsp:
As a high level point, if your goal is to count the number of HLO operators, you should probably do that at the C++ level on the `HloModule` data structure rather than by looking at the string representation.

hawkinsp:
Closing as stale.

PR #5207, opened 2020-12-16
[jax2tf] Clean up and expand binary elementwise harnesses.

bchetioui:


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -826,54 +812,141 @@ def _make_div_rem_harness(name, *, shapes=((2,), (2,)), dtype=np.float32,
...
-          )
-  for f_lax in _LAX_BINARY_ELEMENTWISE
+lax_atan2 = tuple( # Validate dtypes
+  _make_binary_elementwise_harness("dtypes", prim=lax.atan2_p, dtype=dtype)
   for dtype in jtu.dtypes.all_floating
-  for arg1, arg2 in [
-    (np.array([-1.6, -1.4, -1.0, 0.0, 0.1, 0.2, 1., 1.4, 1.6], dtype=dtype),
-     np.array([-1.6, 1.4, 1.0, 0.0, 0.1, 0.2, 1., 1.4, -1.6], dtype=dtype))
+) + tuple( # Validate broadcasting
+  _make_binary_elementwise_harness("broadcasting", prim=lax.atan2_p,
```
gnecula, line 829: Is there a way to consolidate the broadcasting tests so we don't have to repeat this pattern?

bchetioui: Hm I believe yes. This can probably be factorised a bit better – I maybe overly used `sed` :)
I think maybe `_make_binary_elementwise_harness` could become a function that only takes a primitive and a set of dtypes, and returns the whole tuple for both dtypes/broadcasting checks.
WDYT?

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -826,54 +812,141 @@ def _make_div_rem_harness(name, *, shapes=((2,), (2,)), dtype=np.float32,
   ]
 )
 
-_LAX_BINARY_ELEMENTWISE = (
-  lax.add, lax.atan2, lax.div, lax.igamma, lax.igammac, lax.max, lax.min,
```
gnecula, line 830: Are there all the binary ops tested? I did not see below the harnesses for add and and sub. I think that div/rem and min/max are already tested separately.

bchetioui: Yes! `add` was already tested in `lax_add_mul` (line 721) and `sub` in `lax_sub` (line 516).

jax/experimental/jax2tf/tests/primitive_harness.py
gnecula: Sounds reasonable.

jax/experimental/jax2tf/tests/primitive_harness.py
gnecula: I see. Those don't have broadcasting. I wonder if it is important to add broadcasting checks everywhere. Maybe not.

bchetioui:
This last commit should unify things the way we want. Note that I did not change the implementation of `lax_div_rem` because it requires paying some special attention to the generation of the arguments.

gnecula (approved):
Nice!

PR #5211, opened 2020-12-16
Add xla translation rules that accept avals

jakevdp:
Currently, default xla translations assume a one-to-one mapping between XLA buffers and JAX abstract values; with the advent of multi-buffer JAX types, this is no longer sufficient.

~~We could address this by adding a new flavor of translation rules that accept avals, but we already have a proliferation of these (including `translations`, `initial_style_translations`, `parallel_translations`, and `call_translations`)~~

~~This PR introduces a `general_translations` table which contains translation rules with a syntax that is flexible enough to encompass *all* the above flavors of translation rule. This primarily solves the direct problem at hand (the ability to explicitly accept avals in lowered translation rules), and also allow us to migrate other translation forms to this single table and eventually remove the other translation tables (mentioned in new TODOs).~~

This PR adds a `translations_with_avals` table that passes the avals to translation rules; eventually we should be able to move all standard translations to this form, but for now this unblocks the multi-buffer lower_fun work.

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -320,6 +320,9 @@ def primitive_computation(prim, axis_env, backend, tuple_args, *avals, **params)
   elif prim in translations:
     rule = translations[prim]
     ans = rule(c, *xla_args, **params)
+  elif prim in translations_with_avals:
+    rule = translations_with_avals[prim]
+    ans = rule(c, avals, xla_args, params)
   elif prim in initial_style_translations:
     rule = initial_style_translations[prim]
     ans = rule(c, axis_env, extend_name_stack(prim.name), avals, backend,
@@ -430,11 +433,15 @@ def jaxpr_subcomp(c, jaxpr, backend, axis_env, consts, name_stack, *args):
         source_file=frame.file_name if frame else None,
         source_line=frame.line_num if frame else None))
     in_nodes = _flatmap(read, eqn.invars)
+    # TODO(jakevdp): migrate `translations` table to `translations_with_avals`
     if eqn.primitive in backend_specific_translations[platform]:
       rule = backend_specific_translations[platform][eqn.primitive]
       ans = rule(c, *in_nodes, **eqn.params)
     elif eqn.primitive in translations:
       ans = translations[eqn.primitive](c, *in_nodes, **eqn.params)
+    elif eqn.primitive in translations_with_avals:
+      rule = translations_with_avals[eqn.primitive]
+      ans = rule(c, map(aval, eqn.invars), in_nodes, eqn.params)
     elif eqn.primitive in initial_style_translations:
       new_params = check_backend_params(eqn.params, backend)
       rule = initial_style_translations[eqn.primitive]
@@ -897,6 +904,7 @@ ad.primitive_transposes[xla_call_p] = partial(ad.call_transpose, xla_call_p)
 ### translation tables
 
 translations: Dict[core.Primitive, Callable] = {}
+translations_with_avals: Dict[core.Primitive, Callable] = {}
 parallel_translations: Dict[core.Primitive, Callable] = {}
 initial_style_translations: Dict[core.Primitive, Callable] = {}
 call_translations: Dict[core.Primitive, Callable] = {}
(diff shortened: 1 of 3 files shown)


mattjj:
Nice start!

It's not clear to me that we want to unify _all_ translation rule variants by passing everything a rule could possibly need through a generic "translation_params". For example, initial-style translations, which might have subjaxprs, need more data (axis env, name stack, backend), whereas others don't. It's nice to be sparing in the caller (i.e. in `jaxpr_subcomp`) with what information is passed in, rather than always passing in everything and relying on the callee to drop things. That makes information flow clearer in the caller, and can keep callees simpler too.

So we may not want to unify all translation rules into one function signature, we could ask which categories might be worth unifying. I don't know the answer to that! Looking at the current ones, the signatures of call_translations and initial_style_translations look very similar, so perhaps those could be unified as higher_order_translations. We could also merge backend_specific_translations into other tables by passing the platform to every rule. On the flip side, distinguishing rules for higher-order primitives, parallel primitives, and primitives which are neither seems reasonable.

A somewhat orthogonal goal is generalizing all translation rules to take in_avals. I think all translation rules should get that information. As you say, a good incremental adaptation strategy would be to add a new table and move primitives from e.g. `xla.translations` to the new table one-by-one.

AIUI your current main objective is this latter goal. We could punt on the rule-table-coalescing for now.

If you want to focus on the latter goal in this PR, my concrete suggestion is: perhaps change the name of this new table from `general_translations` to `translations_with_avals` or something (not sure the ideal name) and adapt the TODOs and PR description accordingly. (We'd expect this rule table to be transient, in that it would eventually replace `xla.translations`.)

If we want to think about the former goal too, my suggestion is that we should first figure out which rule tables we want to keep distinguished and which we want to unify. (We might end up deciding we _do_ want to unify all of them, or that we want to keep the current distinguished tables, though my guess is we'll converge to some answer in the middle.)

One other concrete suggestion: don't forget to also update `xla.xla_primitive_callable` for any change to the rules tables logic that you make in `xla.jaxpr_subcomp`.

WDYT?

mattjj:
Can you say a bit about what tests/xla_test.py is meant to test? What's the justification for adding it?

jakevdp:
The new tests are unit tests that ensure that both flavors of lowered translation rules work as advertised, with and without multiple results. Since the new argument is not yet used anywhere in the package (that remains a TODO) it's useful to have unit test coverage of this.

jax/interpreters/xla.py
```diff
@@ -954,12 +955,12 @@ def f(c, *xla_args, **params):
                                           stage_out=True)  # type: ignore
       xla_consts = _xla_consts(c, consts)
       outs = jaxpr_subcomp(c, jaxpr, None, axis_env, xla_consts, '', *xla_args)
-    if multiple_results:
+    if multiple_results or len(outs) > 1:
```
mattjj, line 958: This seems surprising. Don't we want `multiple_results` to indicate exactly when outputs should be tupled, and for it to be an error if multiple outputs are produced by the traced function when `multiple_results` is not indicated? The change on this line, together with deleting the assertion that was just below, seems like information is becoming more disorganized/complicated, e.g. in that `multiple_results` doesn't seem to mean what it did previously. I might just be missing something...

mattjj:
I'm concerned unit tests like the ones in xla_test.py may be too low-level; they seem closer to testing an implementation detail rather than a higher-level behavior we care about. The problem with tests on low-level implementation details is that they can make those implementation details harder to revise later (as we'd also need to update the tests). I'd say that's why we don't want unit tests for each function in each module; we choose the appropriate granularity. Where to draw the line is a case-by-case judgement call, though.

> Since the new argument is not yet used anywhere in the package (that remains a TODO) it's useful to have unit test coverage of this.

If you mean the `lower_fun` argument of `with_avals=True`, is it not fully covered by custom_object_test.py?

Whether or not custom_object_test.py already covers it, an alternative to adding the tests in the new xla_test.py test file may be to just make use of `lower_fun(..., with_avals=True)` and/or the `translations_with_avals` table for an existing primitive that we care about. Then existing tests of that primitive would automatically cover this new path.

I'll leave it to you whether to add the new tests that are in xla_test.py. Given my current understanding of what it covers (which could be mistaken!), I suggest against doing it, but we can always try it and see if it proves useful or onerous. If you do add it, don't forget to update the BUILD files so it gets run internally!

jakevdp:
All makes sense – I'll remove the new tests and try converting one or two existing primitives to the new mechanism.

jax/interpreters/xla.py
jakevdp: Perhaps the confusion here is that this change does two things: (1) make xla translations that accept avals, and (2) allow `lower_fun` to be used for functions that return multi-buffered objects, in which case a tuple is required regardless of `multiple_results`.

Since the output avals are not available to inspect here, I don't know of any other way to support this than to assume that multiple buffers should be returned as a tuple. What do you think?

mattjj: Hm I see.

Actually, we might be able to inspect the output avals: `out_avals = [v.aval for v in jaxpr.outvars]`. Would that work?

jakevdp: Good idea - done!

mattjj (approved):
Thanks for being patient and iterating! It was good to think through.

PR #5213, opened 2020-12-17
Add support for 'preferred_element_type' arg in `dot` and `dot_general`.

### LINKED ISSUE #4465: Integer matrix multiplication with int8 input and int32 accumulation
I'm trying to do a hardware-accelerated matmul with int8 input tensors and get back an int32 tensor, but I don't see any way to do that:

```
x=jnp.array(100).reshape((1,1)).astype(jnp.int8)
jnp.matmul(x,x)
```

gives back an int8 result, [[16]]. I don't see any options to specify an accumulator type separate from an input type to `jnp.matmul` or `lax.dot_general`. 

I could cast `x` to `jnp.int32` before the multiply, but I want to use a hardware-accelerated int8 matmul, if available.

Ultimately, I'd like to be able to specify an arbitrary integer bit width input and desired accumulator as well (eg, 4-bit input with 16-bit accumulation), even if current hardware can't accelerate that.

Thanks!

malmaud:
XLA recently added support for this parameter to xops.DotGeneral. It's an optional parameter that controls the accumulation type used by the dot operation.

This is useful for eg quantized ANNs, where you might want to do matrix multiples with int8 tensors and get back an int32 tensor instead of an int8 tensor that suffers from severe overflow. Note it's not sufficient in this case to cast the inputs to 'dot' to int32 beforehand and rely on the default output dtype inference, since backend devices might have an accelerated path for int8*int8->int32 matmuls and we want that explicitly represented in the HLO.

Note because XLA still doesn't support integer dots on the CPU backend, that use case can't tested with a CPU-only test at the moment. I was hoping the experts could weigh on the proper place to add a test for this.

This closes https://github.com/google/jax/issues/4465.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -609,7 +609,8 @@ def conv_general_dilated(
       lhs_shape=lhs.shape, rhs_shape=rhs.shape,
       precision=_canonicalize_precision(precision))
 
-def dot(lhs: Array, rhs: Array, precision: PrecisionLike = None) -> Array:
+def dot(lhs: Array, rhs: Array, precision: PrecisionLike = None,
+        preferred_element_type: Optional[DType] = None) -> Array:
   """Vector/vector, matrix/vector, and matrix/matrix multiplication.
 
   Wraps XLA's `Dot
@@ -625,13 +626,16 @@ def dot(lhs: Array, rhs: Array, precision: PrecisionLike = None) -> Array:
       the backend, a ``lax.Precision`` enum value (``Precision.DEFAULT``,
       ``Precision.HIGH`` or ``Precision.HIGHEST``) or a tuple of two
       ``lax.Precision`` enums indicating precision of ``lhs``` and ``rhs``.
+    preferred_element_type: Optional. Either ``None``, which means the default
+      accumulation type for the input types, or a datatype, indicating to
+      accumulate results to and return a result with that datatype.
 
   Returns:
     An array containing the product.
   """
   if 1 <= lhs.ndim <= 2 and 1 <= rhs.ndim <= 2 and lhs.shape[-1] == rhs.shape[0]:
     return dot_general(lhs, rhs, (((lhs.ndim - 1,), (0,)), ((), ())),
-                       precision=precision)
+                       precision=precision, preferred_element_type=preferred_element_type)
   else:
     raise TypeError("Incompatible shapes for dot: got {} and {}.".format(
         lhs.shape, rhs.shape))
@@ -641,7 +645,8 @@ DotDimensionNumbers = Tuple[Tuple[Sequence[int], Sequence[int]],
                             Tuple[Sequence[int], Sequence[int]]]
 
 def dot_general(lhs: Array, rhs: Array, dimension_numbers: DotDimensionNumbers,
-                precision: PrecisionLike = None) -> Array:
+                precision: PrecisionLike = None,
+                preferred_element_type: Optional[DType] = None) -> Array:
   """More general contraction operator.
 
   Wraps XLA's `DotGeneral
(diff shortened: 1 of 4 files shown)


jax/_src/lax/lax.py
```diff
@@ -640,7 +641,8 @@ def dot(lhs: Array, rhs: Array, precision: PrecisionLike = None) -> Array:
                             Tuple[Sequence[int], Sequence[int]]]
 
 def dot_general(lhs: Array, rhs: Array, dimension_numbers: DotDimensionNumbers,
-                precision: PrecisionLike = None) -> Array:
+                precision: PrecisionLike = None,
+                preferred_element_type: Optional[DType] = None) -> Array:
```
jekbradbury, line 645: Please add this to the args section in the docstring.

jekbradbury (approved):
Do you also need this wired through to any of the wrappers that call dot_general (like `lax.dot` or `np.dot`)? If not it's probably fine to leave it like this.

jax/_src/lax/lax.py
malmaud: Done

malmaud:
The PR already did wire through `lax.dot`, but I added `np.dot` as well.

shoyer:
Can we call this argument `dtype` in the `jax.numpy` functions? That would be consistent with the argument name that NumPy uses for this sort of functionality, e.g., see `dtype` in `np.sum`:
https://numpy.org/doc/stable/reference/generated/numpy.sum.html

malmaud:
> Can we call this argument `dtype` in the `jax.numpy` functions? That would be consistent with the argument name that NumPy uses for this sort of functionality, e.g., see `dtype` in `np.sum`:
> https://numpy.org/doc/stable/reference/generated/numpy.sum.html

Done.

shoyer:
Thanks for adding this to all the NumPy dot functions :).

Currently we summarize the JAX specific arguments in `_PRECISION_DOC`. Maybe you could add a brief note about `dtype` here as well? Most of these functions don't take a `dtype` argument in NumPy, so it doesn't already have documentation.

malmaud:
Done

jax/_src/numpy/lax_numpy.py
```diff
@@ -3439,7 +3445,7 @@ def apply_over_axes(func, a, axes):
 
 
 @_wraps(np.dot, lax_description=_PRECISION_DOC)
```
shoyer, line 3447: Maybe also rename `_PRECISION_DOC` to `_PRECISION_AND_DTYPE_DOC`? :)

shoyer (commented):
Looks great to me on the implementation! It would probably be ideal to also add a few tests to verify that the argument gets plumbed down into XLA correctly, e.g., to verify that the result has the correct `dtype`/values?

malmaud:
I was finding in my tests that the numeric behavioral and return type rules of the `dtype` parameter in numpy seem to differ quite substantially from the `preferred_element_type` in XLA. I can dive deeper into it once I have the cycles, but for now I think it's easiest to have this PR be purely about modifying the `lax` functions to thread `preferred_element_type` through to XLA and have a potential follow-up PR for adding a `dtype` argument to the `numpy` functions that might take advantage of that. 

I also changed the return type inference for lax.dot_general to be correct when the input datatype is more precise than `preferred_element_type` and adding a corresponding test to `lax_tests.py`.

malmaud:
This PR relies on changes to XLA that haven't been upstreamed yet. Once those changes make their way into jaxlib and Jax correspondingly bumps the min jaxilb version, the CI on this PR will pass.

malmaud:
@jekbradbury CI passes now. I had to make a few tweaks since you last looked at this, so you might want to review again.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1287,9 +1287,10 @@ def _conv_general_dilated(lhs, rhs, window_strides, padding, lhs_dilation,
 tf_impl_with_avals[lax.conv_general_dilated_p] = _conv_general_dilated
 
 
-def _dot_general(lhs, rhs, dimension_numbers, precision):
+def _dot_general(lhs, rhs, dimension_numbers, precision, preferred_element_type):
   """Implementation of lax.dot_general_p in terms of tf.linalg.einsum."""
   del precision
+  del preferred_element_type
```
gnecula, line 1293: I assume that `tf.linalg.einsum`, which we use to convert the `lax.dot_general_p`, does not support this capability?

malmaud: It unfortunately doesn't at the moment. We're starting to look into it what it would take to add this on the Tensorflow side.

PR #5216, opened 2020-12-17
Fix UnexpectedTracer omnistaging error.

LenaMartens:
Before: 

```
jax/core.py", line 1220, in call_bind                                                                                                                                                  
jax/util.py", line 36, in safe_map                                                                                                                                                     
    return list(map(f, *args))
jax/core.py", line 377, in full_raise                                                                                                                                                  
    return self.pure(val)                                                                                                                   
jax/interpreters/partial_eval.py", line 1034, in new_const                                                                                                                             
    self.frame.tracers.append(tracer)                                                                                                       
jax/interpreters/partial_eval.py", line 1022, in frame                                                                                                                                 
    return self.main.jaxpr_stack[-1]  # pytype: disable=attribute-error                                                                                                                                                                                                                 
IndexError: tuple index out of range
```

After:
```
jax/core.py", line 1219, in call_bind
    top_trace = find_top_trace(args)
jax/core.py", line 785, in find_top_trace
    top_tracer._assert_live()
jax/interpreters/partial_eval.py", line 950, in _assert_live
    raise core.escaped_tracer_error(msg)
jax.core.UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.
The functions being transformed should not save traced values to global state. Detail: tracer created on line (...).
```

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -779,8 +779,13 @@ def full_lower(val):
     return val
 
 def find_top_trace(xs) -> Trace:
-  top_main = max((x._trace.main for x in xs if isinstance(x, Tracer)),
-                 default=None, key=attrgetter('level'))
+  top_tracer = max((x for x in xs if isinstance(x, Tracer)),
+                    default=None, key=attrgetter('_trace.level'))
+  if top_tracer is not None:
+    top_tracer._assert_live()
+    top_main = top_tracer._trace.main  # type: Optional[MainTrace]
+  else:
+    top_main = None
   dynamic = thread_local_state.trace_state.trace_stack.dynamic
   top_main = (dynamic if top_main is None or dynamic.level > top_main.level
               else top_main)
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -1960,6 +1960,25 @@ class APITest(jtu.JaxTestCase):
                                 "tracer created on line"):
       g()
 
+  def test_escaped_tracer_omnistaging_top_trace(self):
+    if not config.omnistaging_enabled:
+      raise unittest.SkipTest("test is omnistaging-specific")
+
+    count = 1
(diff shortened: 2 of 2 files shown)


jax/core.py
```diff
@@ -779,8 +779,11 @@ def full_lower(val):
     return val
 
 def find_top_trace(xs) -> Trace:
-  top_main = max((x._trace.main for x in xs if isinstance(x, Tracer)),
-                 default=None, key=attrgetter('level'))
+  top_tracer, top_main = max(((x, x._trace.main) for x in xs if isinstance(x, Tracer)),
+                             default=(None, None),
+                             key=lambda x: x[1].level)
+  if top_tracer is not None:
+    top_tracer._assert_live()
```
LenaMartens, line 786: happy to put this check somewhere else if there is a better place for it :)

mattjj: This seems like the perfect place!

Riffing on this code, WDYT about writing it like this?

```python
def find_top_trace(xs) -> Trace:
  top_tracer = max((x for x in xs if isinstance(x, Tracer)),
                    default=None, key=attrgetter('_trace.level'))
  if top_tracer is not None:
    top_tracer._assert_live()
    top_main = top_tracer._trace.main
  else:
    top_main = None
  dynamic = thread_local_state.trace_state.trace_stack.dynamic
  top_main = (dynamic if top_main is None or dynamic.level > top_main.level
              else top_main)
  return top_main and top_main.with_cur_sublevel()  # type: ignore
```

mattjj (approved):
Woohoo, much better error message!

jax/core.py
LenaMartens: Done.

PR #5223, opened 2020-12-18
optional tie-in argument to lax.create_token when omnistaging is on

### LINKED ISSUE #5218: remove tie-in operands under omnistaging for primitives that don't need them
With omnistaging enabled, any primitive such as `create_token`, and its corresponding `lax` function, no longer needs to take a tie-in argument for data dependence.

Removing `create_token`'s operands also enforces that the primitive can't be linear, so it can't be transposed, and the system shouldn't try to transpose it. As @mattjj points out, a nullary primitive wouldn't appear in a jaxpr to be transposed because partial evaluation will never pick it up. This ought to address [this comment in #4292](https://github.com/google/jax/issues/4292#issuecomment-746289386).

froystig:
fixes #5218

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -5898,21 +5898,25 @@ ad.primitive_jvps[ad_util.stop_gradient_p] = _stop_gradient_jvp_rule
 batching.primitive_batchers[ad_util.stop_gradient_p] = _stop_gradient_batch_rule
 
 
-def create_token(x):
+def create_token(_=None):
   """Creates an XLA token value with no preconditions for sequencing effects.
 
   Experimental.
 
-  Args:
-    x: a dummy argument used to tie the CreateToken operator into a trace. The
-       value of `x` is ignored.
+  The argument is ignored. It exists for backward compatibility.
   """
-  # x is a dummy argument used to tie the operator into a trace.
-  return create_token_p.bind(stop_gradient(x))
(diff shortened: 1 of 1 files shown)


jax/_src/lax/lax.py
```diff
@@ -5898,21 +5898,27 @@ def _stop_gradient_batch_rule(batched_args, batch_dims):
...
-def create_token(x):
+def create_token(x=None):
   """Creates an XLA token value with no preconditions for sequencing effects.
 
   Experimental.
 
   Args:
-    x: a dummy argument used to tie the CreateToken operator into a trace. The
-       value of `x` is ignored.
+    x: Optional dummy argument used to tie the CreateToken operator into a
```
mattjj, line 5907: Let's be even more bold here: just say the argument is ignored and nothing else. Consider the docstring of `lax.tie_in` itself!

WDYT about a docstring of:

```python
def create_token(_=None):
  """Creates an XLA token value with no preconditions for sequencing effects.

  Experimental.

  The argument is ignored. It exists for backward compatibility.
  """
```

mattjj (approved):
LGTM! One suggestion on the docstring.

jax/_src/lax/lax.py
froystig: Good call.

froystig:
Also fixes #4292

PR #5224, opened 2020-12-18
Add flag for inf checking

MikeInnes:
The `nan` checker from #482 is really useful. Since `inf` is often just as deadly (and indeed in many cases the cause of `nan` results), I thought it might be useful to have a check for that value too.

I see one other place where there's a `nan` check, [here](https://github.com/google/jax/blob/36f91261099b00194922bd93ed1286fe1c199724/jax/api.py#L375), but since I don't know how this code gets invoked I left it alone. I did test the flag with regular and jitted execution:

```python
>>> jax.config.update("jax_debug_infs", True)
>>> np.exp(100)
  [...]
FloatingPointError: invalid value (inf) encountered in exp
```

If the idea looks good to you let me know anything else I'm missing (I'm very much not familiar with python codebases).


jax/interpreters/xla.py
```diff
@@ -62,6 +62,9 @@
 flags.DEFINE_bool('jax_debug_nans',
                   bool_env('JAX_DEBUG_NANS', False),
                   'Add nan checks to every operation.')
+flags.DEFINE_bool('jax_debug_infs',
+                  bool_env('JAX_DEBUG_infs', False),
```
gnecula, line 66: I would use all-caps for the name of the environment variable to match the syntax of the others

jax/interpreters/xla.py
```diff
@@ -372,6 +376,18 @@ def _check_nans(name, xla_shape, buf):
     if np.any(np.isnan(buf.to_py())):
       raise FloatingPointError(f"invalid value (nan) encountered in {name}")
 
+def check_infs(prim, bufs):
```
gnecula, line 379: I am wondering if instead of duplicating the code for NaN, we can extend it to check for any combination of NaN and Inf based on the flags?

gnecula (changes requested):
Thank you for your contribution!

Please add some tests, perhaps extending debug_nans_test.py. 
Also, please add an entry to docs/changelog.md

jax/interpreters/xla.py
MikeInnes: Whoops, yes, just slipped through – fixed.

jax/interpreters/xla.py
MikeInnes: I agree on the duplication, but wasn't confident doing major surgery on a first PR. I'll try out a little refactor and see what it looks like though.

MikeInnes:
Thanks a lot for pointing me to the changelog and the right place in the tests, I've updated both of those.

On duplication: Functions that used to call `check_nans` conditionally now call `check_special` unconditionally, and `check_special` checks the flags before checking for either `inf` or `nan` as before. Let me know if you'd like any changes there.

gnecula (approved):

gnecula:
I am making small updates to fix test failures.

MikeInnes:
Great, I had trouble running tests locally so was going to check the CI logs, but you got there first. Thanks a lot for getting those and merging.

Issue #2161, opened 2020-02-04
Matrix multiplication precision API


shoyer:
Operations that do matrix multiplication in JAX accept a `precison` for controlling precision when executed on TPUs.

The current API seems non-ideal to me:

1. You have to pass an enum value (e.g., `np.dot(x, y, precision=lax.Precison.HIGHEST)`). This is a little cumbersome and inconsistent with most NumPy/SciPy APIs which use strings (e.g., `np.dot(x, y, precision='highest')`).
2. The current names for precision levels ("highest", "high" and "default") are not very descriptive. In my ideal world we would use some direct indication of the corresponding precision (e.g., bfloat16 multiplication with float32 accumulation), but as the very least can we switch "default" to "low"?
3. The default low precision is a bit of a footgun, at least when doing anything that isn't implementing a neural net layer. In my opinion, it would be much safer to use "highest" precision by default (which isn't _that_ much slower) on float32 data. Neural net libraries, of course, can default to lower precision, so this really only effects users who directly use NumPy APIs or the `@` infix operator.

shoyer:
On TPUs, "HIGH" precision corresponds to 3 passes of bfloat16 and "HIGHEST" precision corresponds to 6 passes, which is effectively full float32 (dropping denormals), as explained in this [Intel paper](https://arxiv.org/pdf/1904.06376.pdf).

With that in mind, and considering that ideally we would retain some flexibility for alternative `matmul` optimizations that might appears on other platforms, what naming scheme makes sense for the `precision` API?

Some ideas:

1. `'low', 'high', 'highest'`: description of precision level
2. `'fastest', 'fast', 'slow'`: description of speed
3. `'fastest', 'fast', 'accurate'`: mixed description, using only positive words
4. `'bfloat16', 'float24', 'float32'`: rough precision of the underlying arithmetic (but what is "float24"??)

I think I lean towards option 3?

shoyer:
Notes from offline discussion:

- This is really a "minimum precision" configuration, so perhaps a name like `min_precision` would be more appropriate.
- The other way to configure matmul precision (maybe more obvious) is by explicitly setting `dtype`. XLA will use lowest precision on `bfloat16` data regardless of the precision option.
- We want an API that also can support new matmul precision options as they arise on different platforms (GPU, CPU, etc), e.g., `precision={'tpu': 'bfloat16', 'gpu': 'float16'}`.
[...]

PR #5225, opened 2020-12-18
Complex SVD JVP rule

frederikwilde:
The current implementation of the singular value decomposition is only differentiable for real-valued input.
I derived the JVP rule [to appear on arxiv.org soon] for complex matrices and implemented it. I also tested it on various random matrices. Happy holidays!

Here is some minimal example to demonstrate that the current implementation does not work on complex matrices.
```Python
from jax import grad
import jax.numpy as jnp

A = jnp.array([[1.j, .5j], 
               [0.,  0. ]], dtype=jnp.complex64)
B = jnp.array([[1.,  0.],
               [0.,  0.]], dtype=jnp.complex64)
def f(x):
    U, S, V = jnp.linalg.svd(A + x * B, full_matrices=False)
    return U.dot(jnp.diag(S).dot(V))[0, 0].real

print(f'The gradient should give 1 but gives: {grad(f)(1.)}')
```

### DIFF
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -559,11 +559,34 @@ class NumpyLinalgTest(jtu.JaxTestCase):
 
     self._CompileAndCheck(partial(jnp.linalg.svd, full_matrices=full_matrices, compute_uv=compute_uv),
                           args_maker)
-    if not (compute_uv and full_matrices):
+    if not compute_uv:
       svd = partial(jnp.linalg.svd, full_matrices=full_matrices,
                     compute_uv=compute_uv)
       # TODO(phawkins): these tolerances seem very loose.
-      jtu.check_jvp(svd, partial(jvp, svd), (a,), rtol=5e-2, atol=2e-1)
+      if dtype == np.complex128:
+        jtu.check_jvp(svd, partial(jvp, svd), (a,), rtol=1e-4, atol=1e-4, eps=1e-8)
+      else:
+        jtu.check_jvp(svd, partial(jvp, svd), (a,), rtol=5e-2, atol=2e-1)
+
+    if jtu.device_under_test() == "tpu":
+      raise unittest.SkipTest("TPU matmul does not have enough precision")
+    # TODO(frederikwilde): Find the appropriate precision to use for this test on TPUs.
+
+    if compute_uv and (not full_matrices):
+      b, = args_maker()
+      def f(x):
+        u, s, v = jnp.linalg.svd(
+          a + x * b,
+          full_matrices=full_matrices,
+          compute_uv=compute_uv)
+        vdiag = jnp.vectorize(jnp.diag, signature='(k)->(k,k)')
+        return jnp.matmul(jnp.matmul(u, vdiag(s)), v).real
+      _, t_out = jvp(f, (1.,), (1.,))
+      if dtype == np.complex128:
+        atol = 1e-13
+      else:
+        atol = 5e-4
+      self.assertArraysAllClose(t_out, b.real, atol=atol)
 
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_shape={}_fullmatrices={}".format(
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -1171,7 +1171,6 @@ def svd_jvp_rule(primals, tangents, full_matrices, compute_uv):
     raise NotImplementedError(
       "Singular value decomposition JVP not implemented for full matrices")
 
-  k = s.shape[-1]
   Ut, V = _H(U), _H(Vt)
   s_dim = s[..., None, :]
   dS = jnp.matmul(jnp.matmul(Ut, dA), V)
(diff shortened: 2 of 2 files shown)


mattjj:
Wow, what a gift for the holidays! Amazing!

shoyer:
This is really fantastic!

It would be great to add some unit tests for this to verify this fix. The specific example you give above would be a good example to add as a regression test.

Here are our current tests for SVD gradients. But as noted by @hawkinsp, "these tolerances seem very loose":
https://cs.opensource.google/jax/jax/+/master:tests/linalg_test.py;l=567;drc=f74235cdae037c686f1b05bb9608fc48cd849664

frederikwilde:
It seems like these tolerances need to be so loose because the ```eps``` used for the [numerical JVP](https://github.com/google/jax/blob/master/jax/test_util.py#L215) is ```1e-4```, which is the lowest possible value for ```np.complex64``` numbers. For ```np.complex128``` numbers one can set ```eps = 1e-8``` and ```atol = 1e-4``` and ```rtol = 1e-4```, possibly even lower.

I will add this and some test for the entire SVD, as @shoyer mentioned.

frederikwilde:
I just noticed the [JVP test](https://github.com/google/jax/blob/master/tests/linalg_test.py#L563) is also run when ```compute_uv``` is true. I don't know how the test cases are picked from all possible parameters, but only one configuration, namely 3 matrices of shape (53, 7), is run and here the test passes, probably by luck (or bad luck).

When I run it manually for a 2 by 2 matrix it fails, which is not surprising, since the (numerical) derivative of ```U``` and ```Vt``` is not well defined.

I will fix this as well.

frederikwilde:
I hope these tests look good to you. I also have a question:

In my note I want to motivate a bit why one would want to use the JVP rule instead of the VJP.  As far as I understand, for backward-mode differentiation the JVP rule is "translated" into the VJP rule via transposition. Could you point me to the relevant piece of code, or are there any references on this technique?

shoyer:
I don't think there's a formal write up of how JAX does backwards mode
autodiff, but perhaps there should be! Another way of thinking of a
vector-Jacobian product is that is it transpose(Jacobian)-vector product.
So JAX first derives how to calculate a JVP, which is a linear map, and
then transposes it operation by operation using the rule for transposing
matrix products: (AB)^T = B^T A^T.

hawkinsp:
@shoyer LGTY?

shoyer:
Just to confirm -- your changed unit-tests would fail with the original (unmodified) JVP rule?

frederikwilde:
Yes. Here are the tests on the unmodified JVP rule, which is exactly what one would expect.
```
[ RUN      ] NumpyLinalgTest.testSVD_n=complex128[2,3,29,7]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=complex128[2,3,29,7]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=complex128[53,3]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=complex128[53,3]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=complex64[2,3,3,53]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=complex64[2,3,3,53]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=complex64[3,53,7]_full_matrices=False_compute_uv=True
[  FAILED  ] NumpyLinalgTest.testSVD_n=complex64[3,53,7]_full_matrices=False_compute_uv=True
[ RUN      ] NumpyLinalgTest.testSVD_n=float32[2,3,29,53]_full_matrices=True_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=float32[2,3,29,53]_full_matrices=True_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=float32[2,3,7,53]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=float32[2,3,7,53]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=float32[3,7,7]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=float32[3,7,7]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=float32[7,29]_full_matrices=False_compute_uv=True
[       OK ] NumpyLinalgTest.testSVD_n=float32[7,29]_full_matrices=False_compute_uv=True
[ RUN      ] NumpyLinalgTest.testSVD_n=float64[29,7]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=float64[29,7]_full_matrices=False_compute_uv=False
[ RUN      ] NumpyLinalgTest.testSVD_n=float64[3,7,3]_full_matrices=False_compute_uv=False
[       OK ] NumpyLinalgTest.testSVD_n=float64[3,7,3]_full_matrices=False_compute_uv=False
```

shoyer (approved):
This is great, thank you!

Please be sure to come back and add share link to your arXiv paper when it comes out :)

tests/linalg_test.py
```diff
@@ -560,11 +560,30 @@ def norm(x):
...
+          full_matrices=full_matrices,
+          compute_uv=compute_uv)
+        vdiag = jnp.vectorize(jnp.diag, signature='(k)->(k,k)')
+        return jnp.matmul(jnp.matmul(u, vdiag(s)), v).real
+      _, t_out = jvp(f, (1.,), (1.,))
+      if dtype == np.complex128:
+          tol = 1e-13
+      else:
+          tol = 5e-4
+      self.assertTrue(np.all(np.linalg.norm(t_out - b.real, axis=(-2, -1)) < tol))
```
jakevdp, line 586: This test is failing on GPU for `complex64` and on TPU for `float32`, which is blocking merge. Because of the use of `assertTrue` the error does not report what the computed tolerance is. It would be better to chnge this to
```python
self.assertLess(np.all(np.linalg.norm(t_out - b.real, axis=(-2, -1)), tol))
```
so the error message reports the norm.

Relevant tolerances will then have to be adjusted in order to pass tests & merge.

jakevdp: Actually, that was a silly suggestion. I think this would be the way to do it:
```python
self.assertArraysAllClose(np.linalg.norm(t_out - b.real, axis=(-2, -1)), 0, atol=atol)
```

jakevdp: .. or perhaps compare the arrays directly, unless there's good reason to do it via a manual norm:
```python
self.assertArraysAllClose(t_out, b.real, atol=atol)
```

IvanYashchuk:
Hi, just wanted to leave relevant links here where differentiation rules for complex-valued SVD are derived.
https://giggleliu.github.io/2019/04/02/einsumbp.html
https://arxiv.org/abs/1909.02659

shoyer:
The failing TPU tests are almost certainly yet another manifestation of the surprisingly low default precision for TPU matmul (https://github.com/google/jax/issues/2161). To fix them, either try setting `precision=lax.Precison.HIGHEST` inside `jnp.matmul`, or feel free to skip the test entirely on TPU, e.g.,
```python
    if jtu.device_under_test() == "tpu":
      raise unittest.SkipTest("TPU matmul does not have enough precision")
```

tests/linalg_test.py
frederikwilde: I wouldn't say there is any reason to use the norm and I will switch to ```assertArraysAllClose```. However the tests don't fail on the GPU I am using, so I don't really know which tolerance to choose. Hopefully the copybara checks reveal useful insights.

jakevdp:
Thanks! GPU tests are passing, but TPU tests are still failing. Example failure:
```
NumpyLinalgTest.testSVD_n=float32[2,29]_full_matrices=False_compute_uv=True

AssertionError: 
Not equal to tolerance rtol=0.001, atol=0.0005

Mismatch: 86.2%
Max absolute difference: 0.03964618
Max relative difference: 0.22800188
 x: array([[-6.51828 ,  1.907845,  1.439919, -1.211914,  3.241142,  0.213531,
        -0.334927,  2.037017, -0.959   , -5.248371,  3.44313 , -0.882515,
        -5.530334,  0.243767, -5.050095,  0.480526,  6.139389, -1.660263,...
 y: array([[-6.503403,  1.909549,  1.441591, -1.208022,  3.229221,  0.173885,
        -0.340035,  2.025751, -0.968591, -5.264877,  3.430215, -0.867395,
        -5.54072 ,  0.249359, -5.049953,  0.474082,  6.128053, -1.67023 ,...
```
 I'd recommend following @shoyer's advice above and skip the TPU tests for now, with a TODO to re-enable them; someone with more direct access to TPU runtimes can iterate faster to fix the issue.

tests/linalg_test.py
jakevdp: Thanks!

jakevdp:
Failures are unrelated, due to the issue in #5374

Can you sqash your commits and rebase on master? Thanks!

frederikwilde:
I hope this looks good. There are actually still a few code comments that are obsolete. Should I remove the [TODO](https://github.com/frederikwilde/jax/blob/complex-svd-jvp-rule/tests/linalg_test.py#L565) by phawkins?

hawkinsp:
We merged this PR. If you want to send another doing any cleanup, that'd be welcome.

jakevdp:
Thanks @frederikwilde, this is an awesome contribution!

mattjj:
Yeah, thanks! Amazing work!

frederikwilde:
Thanks everyone! Also for maintaining this great package. It's been an educating experience all the way :-)

PR #5229, opened 2020-12-19
Convert arguments of pad into kwargs and handle invalid kwargs

minoring:
Unsupported keyword argumenets for jax.numpy.pad should raise an error.
In numpy,

```python
>>> import numpy as np
>>> a = [1, 2, 3]
>>> np.pad(a, mode='maximum', pad_width=2, reflect_type='odd')
ValueError: unsupported keyword arguments for mode 'maximum': {'reflect_type'}
```

In jax,
Before
```python
>>> import jax.numpy as jnp
>>> a = [1, 2, 3]
>>> jnp.pad(a, mode='maximum', pad_width=2, reflect_type='odd')
DeviceArray([3, 3, 1, 2, 3, 3, 3], dtype=int32)
```

After
```python
>>> import jax.numpy as jnp
>>> a = [1, 2, 3]
>>> jnp.pad(a, mode='maximum', pad_width=2, reflect_type='odd')
ValueError: unsupported keyword arguments for mode 'maximum': {'reflect_type'}
```

Related #5010

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2521,39 +2521,62 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
   if mode == "constant":
     return _pad_constant(array, pad_width, constant_values)
 
-  elif mode == "wrap":
+  if mode == "wrap":
     return _pad_wrap(array, pad_width)
 
-  elif mode in ("symmetric", "reflect"):
+  if mode in ("symmetric", "reflect"):
     return _pad_symmetric_or_reflect(array, pad_width, mode, reflect_type)
 
-  elif mode == "edge":
+  if mode == "edge":
     return _pad_edge(array, pad_width)
 
-  elif mode == "linear_ramp":
+  if mode == "linear_ramp":
     end_values = _broadcast_to_pairs(end_values, nd, "end_values")
     return _pad_linear_ramp(array, pad_width, end_values)
 
-  elif mode in stat_funcs:
+  if mode in stat_funcs:
     if stat_length is not None:
       stat_length = _broadcast_to_pairs(stat_length, nd, "stat_length")
     return _pad_stats(array, pad_width, stat_length, stat_funcs[mode])
 
-  elif mode == "empty":
+  if mode == "empty":
     return _pad_empty(array, pad_width)
 
-  else:
-    msg = "Unimplemented padding mode '{}' for np.pad."
-    raise NotImplementedError(msg.format(mode))
+  assert False, ("Should not be reached since pad already handled unsupported and"
+                 "not implemented modes")
 
 
 @_wraps(np.pad)
-def pad(array, pad_width, mode="constant", constant_values=0, stat_length=None,
-        end_values=0, reflect_type="even"):
+def pad(array, pad_width, mode="constant", **kwargs):
   if isinstance(pad_width, Iterable):
     pad_width = tuple(
         tuple(int(i) for i in x) if isinstance(x, Iterable) else x
         for x in pad_width)
+  allowed_kwargs = {
+      'empty': [], 'edge': [], 'wrap': [],
+      'constant': ['constant_values'],
+      'linear_ramp': ['end_values'],
+      'maximum': ['stat_length'],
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2521,39 +2521,62 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
   if mode == "constant":
     return _pad_constant(array, pad_width, constant_values)
 
-  elif mode == "wrap":
```
gnecula, line 2524: I have a slight preference for the sequence of `elif` because it more clearly expresses a case statement, without having to read all the branches to ensure that there is no fall-through. WDYT?

minoring: Thanks for the review:)
I agree with your idea considering it is a fairly long case statement.
Learn a new thing!

jax/_src/numpy/lax_numpy.py
```diff
@@ -2543,17 +2543,41 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
...
+      'mean': ['stat_length'],
+      'median': ['stat_length'],
+      'minimum': ['stat_length'],
+      'reflect': ['reflect_type'],
+      'symmetric': ['reflect_type'],
+  }
+  try:
+    unsupported_kwargs = set(kwargs) - set(allowed_kwargs[mode])
+  except KeyError:
+    msg = "Unimplemented padding mode '{}' for np.pad."
```
gnecula, line 2570: It seems that this branch is not tested

jax/_src/numpy/lax_numpy.py
```diff
@@ -2543,17 +2543,41 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
     return _pad_empty(array, pad_width)
 
   else:
-    msg = "Unimplemented padding mode '{}' for np.pad."
```
gnecula, line 2546: Shouldn't this be an "Unimplemented padding mode" error? I believe that the previous version of your changes had the right `else` clause here, followed by the `assert False`

tests/lax_numpy_test.py
```diff
@@ -1477,6 +1477,38 @@ def testPadEmpty(self):
...
+        'minimum': {'stat_length': None},
+        'reflect': {'reflect_type': 'even'},
+        'symmetric': {'reflect_type': 'even'},
+        'wrap': {},
+        'empty': {}
+    }
+    arr = [1, 2, 3]
+    pad_width = 1
+
+    for mode in modes.keys():
```
gnecula, line 1497: Maybe add an invocation with an invalid `mode`?

jax/_src/numpy/lax_numpy.py
minoring: Add tests for this PTAL:)

jax/_src/numpy/lax_numpy.py
minoring: I am not mistaken,
https://github.com/google/jax/pull/5229/files#diff-13c884892e20a9f28e3bd980ee444f3b0d5d90258fe9aa6dcfa84a2089d34844R2570
I moved "Unimplemented padding mode" to above.

Every errors of unsupported modes and invalid keyword arguments are handled in `pad` function,
I thought `_pad` function should never reach its end.

tests/lax_numpy_test.py
minoring: Added test for this.

jax/_src/numpy/lax_numpy.py
gnecula: Indeed, sorry!

gnecula: Right

gnecula:
It turns out that this change breaks internal Google tests, for code that was using `jnp.pad` without kwargs. I am going to try to fix those failures, but can you please make another PR with an entry in the docs/changelog.rst with the API change?

minoring:
@gnecula https://github.com/google/jax/pull/5249 I made a new PR. PTAL:)
And will you revert these commits?

Issue #5230, opened 2020-12-20
Issue in readme for vmap


MiladInk:
Hi great Jax team!

I just noticed in the readme file of the repository, in the explanation of vmap, there may be an unintended typo:

```
def predict(params, input_vec):
  assert input_vec.ndim == 1
  for W, b in params:
    output_vec = jnp.dot(W, input_vec) + b  # `input_vec` on the right-hand side!
    input_vec = jnp.tanh(output_vec)
  return output_vec
```

I guess it should be:

```
def predict(params, input_vec):
  assert input_vec.ndim == 1
  for W, b in params:
    output_vec = jnp.dot(W, input_vec) + b  # `input_vec` on the right-hand side!
    output_vec = jnp.tanh(output_vec) # I think this line was intended to be like this
  return output_vec
```

Thanks for Jax!

8bitmp3:
Yup, makes sense: `prediction_output = jnp.tanh(jnp.dot(W, input_vec) + b)` 👍

> Thanks for Jax!

Same here!

![image](https://user-images.githubusercontent.com/19637339/102720546-e82e5e00-42ec-11eb-9446-4b9e3c4ed412.png)

mattjj:
Hey, thanks for calling attention to this, and for unpacking it so clearly!

Actually, it is written as intended. The value of `input_vec` is used on the next iteration of the loop, so that one layer's output is the next layer's input.

Perhaps it looks surprising at first because the last iteration's updated value of `input_vec` is not consumed. But writing it this way lets us avoid special-casing the last layer (to be linear, i.e. not to have an activation). That is, it avoids us having to write it more like:

```python
def predict(params, input_vec):
  assert input_vec.ndim == 1
  for W, b in params[:-1]:
    output_vec = jnp.dot(W, input_vec) + b
    input_vec = jnp.tanh(output_vec)
  W_final, b_final = params[-1]
  output_vec = jnp.dot(W_final, input_vec) + b_final
  return output_vec
```

We chose to make the example code more concise. Moreover, any dead code is eliminated automatically under a `jit`, so we don't usually worry about things like an extra `jnp.tanh` evaluation!

Hope that makes sense. Thanks for the kind words, and for surfacing any possible issue you spot. It's really helpful to have extra checks; false positives are better than missed detections!

MiladInk:
Thanks a lot for the detailed explanation!

PR #5236, opened 2020-12-21
(docfix) Clarify variable names in README.md snippit

bileschi:
`predict` function should apply nonlinearity to output

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,10 +217,11 @@ function:
 ```python
 def predict(params, input_vec):
   assert input_vec.ndim == 1
+  activations = inputs
   for W, b in params:
-    output_vec = jnp.dot(W, input_vec) + b  # `input_vec` on the right-hand side!
-    output_vec = jnp.tanh(output_vec)
-  return output_vec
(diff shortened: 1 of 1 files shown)


jakevdp:
Hi - thanks for looking into this! This is actually correct as currently written: the idea is that the output vector from one iteration becomes the input vector for the next. I agree it's a bit confusing... We discussed a bit how to improve this in #4048, but didn't land on any good options.

gnecula:
I propose to change the example along the lines discussed in #4048:

```
def predict(params, inputs):
  activations=inputs
  for W, b in params:
    outputs = jnp.dot(activations, W) + b
    activations = jnp.tanh(outputs)
  return outputs
```

@bileschi do you want to do that instead?

8bitmp3:
(Naming is hard. Linking this issue https://github.com/google/jax/issues/5230 just in case)

bileschi:
Updated to fix / reflect `activations` / `outputs` naming.

gnecula (approved):

PR #5243, opened 2020-12-22
[host_callback] Added API to call host computations and return values to the accelerator computation

gnecula:
This PR extends the `host_callback` module with one new function (`call`) to call a Python function on the host and feed the result back to the device computation. 

The PR contains the documentation of the API at the top of `host_callback.py`, along with relevant examples of use cases: 
calling from JAX to NumPy CPU custom kernels, to TensorFlow functions, or to JAX running on another device. In the latter
two cases we show how we can support reverse-mode AD for the host callbacks, by deferring to the reverse-mode AD on the target platform. The relevant examples are in `host_callback_test.py` and `host_callback_to_tf_test.py`.

In addition to the user documentation, here are some design considerations to discuss. 

Choice of API
--------------

* We have both `id_tap` and `call` (for now). This is because `id_tap` and `id_print` support all JAX transformations, while `call` does not have support for any transformation (yet).
* The `call` function takes only a single argument (like the `id_tap` and `id_print`). If the user wants to pass multiple arguments, they can pass a pytree. 
* The `call` function takes an optional parameter `call_with_device`, that will ensure that a `device=dev` kwarg is passed to the host callback function. This is useful in presence of execution on multiple devices, e.g., inside pmap. In many other cases it is not interesting. Alternative designs would be:
    * Always pass the device to all callback functions. I think that most functions will not need that argument, but perhaps this option makes sense for uniformity. 
    * Or, use the `inspect` module to see if the host function expects a `device` kwarg and pass it. I opted for the more explicit option. 

* The `call` function takes a `result_shape` parameter, which can be an example value from which the shape and dtype are determined, or any object with `shape` and `dtype` properties, e.g.,  `np.ndarray`, or `jax.ShapeDtypeStruct`. The compilation of `call` needs to know the expected shape and dtype of the result. There are several alternative designs:
   * Always require an object with `shape` and `dtype`. This is annoying when the argument is a scalar and the user needs to make some jax.ShapeDtypeStruct from it. So, I have decided to allow scalars to be passed as example values. 
   * JAX could obtain the result shape by actually running the callback function when the `call` is encountered during tracing. This works very well outside of staging contexts (jit, pmap, control-flow primitives), because it effectively inlines the host function. This can in fact be the implementation of `call` outside of staging contexts. But in staging contexts, it may be confusing to execute the callback function once during tracing, with actual arguments manufactured from the JAX abstract values and perhaps triggering side-effects. For the cases when the programmer wishes to, the proposed API allows the programmer to construct the result shape by calling the actual host function to get an example result. 

Unpleasant ergnonomics in presence of errors
----------------------------------------------

For `call` (but not for `id_tap`, nor for `call` that does not have results), if there is an exception in the Python user-defined function, or if the function returns a result that does not have the expected shape, we can log an error and we can throw an exception but by this point the accelerator computation will likely block waiting for the result. 

I have tried several things here. The only safe solution seems to be to abort the device computation in a way that surfaces a Python exception. Returning anything that allows the computation to continue seems unsafe. 

I will continue to investigate how to abort the computation.

The best option I have at the moment is to send something with a shape that is different than what the device expects, counting on a shape check failure. I picked a distinctive shape s8[12345] so that we can recognize it in the error
messages. This works for CPU (runtime crash), for GPU (a nice Python error message), but not for TPU. On TPU there is no shape check, the computation will just go ahead with whatever garbage it thinks it got. So, for TPU I am not sending anything and let the computation hang.

This is especially bad for Colab uses because the user may not see the error logs. 

Implementation mechanism
----------------------------

Currently, the implementation uses `outfeed` to send the arguments to the host, and (optionally) `infeed` if there are expected results (for `call` when the results are not empty). Two tokens are threaded through the computation (after all the transformations and before compilation to XLA): one token for outfeed and one for infeed, carried through the nested control-flow primitives to ensure that the infeed/outfeed operations are happening in the proper order. The outfeed operations are sequenced among themselves, and the infeed operation corresponding to an outfeed is sequenced after the outfeed, and after previous infeed operations.

Consider the compilation of the following code that computes `sin(cos(x))` but passing the result of `cos(x)` to the host. 
```
def func(x):
  return jnp.sin(hcb.call(lambda x_host: x_host, 
                                        jnp.cos(x), result_shape=x))
```

The following shows the compilation, with comments
```
ENTRY %xla_computation_fun.24 (parameter.1: f32[2]) -> (f32[2]) {
  %after-all.4 = token[] after-all()    # create an infeed token, if none is passed in 
  %parameter.1 = f32[2]{0} parameter(0)
  %cosine.5 = f32[2]{0} cosine(f32[2]{0} %parameter.1)       # jnp.cos(x)
  %tuple.6 = (f32[2]{0}) tuple(f32[2]{0} %cosine.5)    # prepare for outfeed
  %constant.7 = u32[2]{0} constant({271828, 12253941})   # outfeed first a descriptor identifying the callback function and argument shape
  %after-all.3 = token[] after-all()    # create an outfeed token, if none is passed in
  %outfeed.8 = token[] outfeed(u32[2]{0} %constant.7, token[] %after-all.3)   # outfeed the descriptor
  %outfeed.9 = token[] outfeed((f32[2]{0}) %tuple.6, token[] %outfeed.8)  # outfeed the arguments, after the descriptor
  %after-all.10 = token[] after-all(token[] %after-all.4, token[] %outfeed.9)     # a token after the outfeed and the previous infeeds
  %infeed.11 = ((f32[2]{0}), token[]) infeed(token[] %after-all.10)   # infeed the result. The compiler may schedule other computation before
  %get-tuple-element.12 = (f32[2]{0}) get-tuple-element(((f32[2]{0}), token[]) %infeed.11), index=0
  %get-tuple-element.14 = f32[2]{0} get-tuple-element((f32[2]{0}) %get-tuple-element.12), index=0
  %sine.22 = f32[2]{0} sine(f32[2]{0} %get-tuple-element.14)   # the jnp.sin operation
  ROOT %tuple.23 = (f32[2]{0}) tuple(f32[2]{0} %sine.22)
}
```
(The computation is optimized if the `call` does not have expected result: the `infeed` is skipped)

The outfeed/infeed mechanism has the advantage of working on all platforms. It has the drawback of using one global communication channel for each device and thus requires the descriptor mechanism shown above to pick the right callback function in presence of control flow in the device computation. (We need to do two separate outfeed operations due to a limitation/bug in the GPU implementation of outfeed; it would be nice to have a single outfeed that includes the descriptor and payload in the same operation.)

We are now exploring an alternative implementation on CPU and GPU that would use XLA CustomCall to call directly to the callback function.

### DIFF
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -247,7 +247,7 @@ are processed asyncronously with the device computation.
 Exceptions from the user-defined callback functions are logged along with their
 stack traces, but the receiving threads are not stopped. Instead the last
 exception is recorded and the subsequent :func:`barrier_wait` will
-raise :exc:`TapFunctionException` if any exception had occurred
+raise :exc:`CallbackException` if any exception had occurred
 in one of the tap functions. This exception will include the text and the
 stack trace of the last exception encountered.
 
@@ -447,7 +447,7 @@ def id_tap(tap_func, arg, *, result=None, tap_with_device=False, **kwargs):
         "Support for **kwargs in ``id_tap`` has been removed. Instead, "
         "pre-apply keyword arguments, either by using a closure or by passing "
         "``functools.partial(tap_func, **kwargs)``.")
-    raise ValueError(msg)
+    raise TypeError(msg)
 
   _initialize_outfeed_receiver()  # Lazy initialization
   api._check_callable(tap_func)
@@ -1469,20 +1469,22 @@ xla.translations[id_p] = lambda c, *args: xops.Tuple(c, args)
 xla.outfeed_rewriter = lambda j: _rewrite_jaxpr(j, False, False)
 
 
-class TapFunctionException(Exception):
-  """Signals that some tap function had exceptions.
+class CallbackException(Exception):
+  """Signals that some callback function had exceptions.
 
   Raised by :func:`barrier_wait`.
+  See module documentation for details.
   """
   pass
 
+TapFunctionException = CallbackException  # For backwards compatibility
 
 # For now we keep a single outfeed receiver
 class _OutfeedReceiverData:
   """Keep track of the outfeed receiver data."""
   receiver: Any
   lock: threading.Lock
-  last_tap_exception: Optional[str]
+  last_callback_exception: Optional[Tuple[Exception, str]]
   clients: Tuple[XlaLocalClient, ...]
   devices: Tuple[XlaDevice, ...]
   consumer_registry: Dict[_ConsumerCallable, int]
@@ -1491,7 +1493,7 @@ class _OutfeedReceiverData:
   def __init__(self):
     self.receiver = None  # Initialize lazily, when first needed
     self.lock = threading.Lock()
-    self.last_tap_exception = None
+    self.last_callback_exception = None
     self.clients = ()
     self.devices = ()
     # The consumer registries must be live for the lifetime of the program,
@@ -1520,11 +1522,10 @@ def _outfeed_receiver_callback(device, consumer_id, arrays):
   assert consumer is not None, "We should have crashed in the runtime"
   try:
     consumer.invoke(arrays, device)
-  except Exception:
-    edata = traceback.format_exc()
-    logging.error("Postponing exception raised in tap function: %s", edata)
-    _outfeed_receiver.last_tap_exception = edata
-    return
+  except Exception as e:
+    formatted_e = traceback.format_exc()
+    logging.error("Postponing exception raised in callback function: %s", formatted_e)
+    _outfeed_receiver.last_callback_exception = (e, formatted_e)
 
 
 def _initialize_outfeed_receiver(
@@ -1578,7 +1579,7 @@ def barrier_wait(logging_name: Optional[str] = None):
 
   Waits until all outfeed from computations already running on all devices
   has been received and processed by the Python callbacks. Raises
-  TapFunctionException if there were exceptions while processing the callbacks.
+  Callback if there were exceptions while processing the callbacks.
 
   This works by enqueueing a special tap computation to all devices to which
   we are listening for outfeed. Once all those tap computations are done, we
@@ -1619,12 +1620,12 @@ def barrier_wait(logging_name: Optional[str] = None):
   with lock:
     cv.wait_for(lambda: num_at_large == 0)
   logging.vlog(2, f"barrier_wait[{logging_name}]: done")
-  if _outfeed_receiver.last_tap_exception is not None:
-    last_exception = _outfeed_receiver.last_tap_exception
-    _outfeed_receiver.last_tap_exception = None
-    raise TapFunctionException(
-        "There were exceptions during id_tap processing. "
-        f"Last one was: {last_exception}")
+  if _outfeed_receiver.last_callback_exception is not None:
+    last_exception, formatted_last_exception = _outfeed_receiver.last_callback_exception
+    _outfeed_receiver.last_callback_exception = None
+    raise CallbackException(
+        "There were exceptions during callback processing. "
+        f"Last one was: {formatted_last_exception}") from last_exception
 
 
 def stop_outfeed_receiver():
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -875,25 +875,20 @@ def _add_transform(params: Dict, name: str, *transform_params) -> Dict:
 # TODO(necula): there must be a better way to do this.
 # The AttributeError is for regular values, the KeyError is for ConcreteArray
 def _instantiate_zeros(arg, tan):
-  """Turn special ad.zero tangents into arrays of 0s."""
+  """Turn special ad.zero tangents into arrays of 0s for sending to host."""
+  # return ad.instantiate_zeros(tan)
   if type(tan) is not ad.Zero:
     return tan
 
-  try:
-    aval = arg.aval
-    return ad.instantiate_zeros_aval(aval, tan)
-  except (AttributeError, KeyError):
-    # We get here for regular Python values
-    return ad.zeros_like_jaxval(arg)
-
-
-def _instantiate_zeros_aval(aval, tan):
-  """Turn special ad.zero tangents into arrays of 0s."""
-  if type(tan) is not ad.Zero:
-    return tan
-
-  return ad.instantiate_zeros_aval(aval, tan)
-
+  if tan.aval is core.abstract_unit:
+    if ad.is_undefined_primal(arg):
+      aval = arg.aval
+    else:
+      aval = core.raise_to_shaped(core.get_aval(arg))
+  else:
+    aval = tan.aval
+  res = ad.instantiate_zeros_aval(aval, tan)
+  return res
 
 def _id_tap_jvp_rule(primals, tangents, **params):
   tangent_instantiated = tuple(map(_instantiate_zeros, primals, tangents))
@@ -1579,7 +1574,7 @@ def barrier_wait(logging_name: Optional[str] = None):
 
   Waits until all outfeed from computations already running on all devices
   has been received and processed by the Python callbacks. Raises
-  Callback if there were exceptions while processing the callbacks.
+  CallbackException if there were exceptions while processing the callbacks.
 
   This works by enqueueing a special tap computation to all devices to which
   we are listening for outfeed. Once all those tap computations are done, we
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -40,7 +40,7 @@ NumPy arrays to the device computation.
 Host computation is useful, e.g., when a device computation needs some data
 that requires I/O on the host, or it needs a library that is available on the
 host and you do not want to code it in JAX.
-For example, eigen decomposition in JAX does not work on TPU.
+For example, eigen decomposition for general matrices in JAX does not work on TPU.
 We can call the Numpy implementation from any JAX accelerator computation,
 using a host computation::
 
@@ -447,7 +447,7 @@ def id_tap(tap_func, arg, *, result=None, tap_with_device=False, **kwargs):
         "Support for **kwargs in ``id_tap`` has been removed. Instead, "
         "pre-apply keyword arguments, either by using a closure or by passing "
         "``functools.partial(tap_func, **kwargs)``.")
-    raise ValueError(msg)
+    raise TypeError(msg)
 
   _initialize_outfeed_receiver()  # Lazy initialization
   api._check_callable(tap_func)
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -11,12 +11,118 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Primitives for calling from accelerators to Python functions on the host.
+"""Primitives for calling from JAX accelerator code to Python functions on the host.
 
 **Experimental: please give feedback, and expect changes.**
 
-This module introduces the host callback functions :func:`id_tap` and
-:func:`id_print`, which behave like the identity function but have the
+This module introduces the host callback functions :func:`call`,
+:func:`id_tap`, and :func:`id_print`, that send their arguments from the device
+to the host and invoke user-defined Python functions on the host, optionally
+returning results back to the device computation.
+
+We show below how these functions can be used. We start with :func:`call`,
+and we discuss examples of calling from JAX to NumPy CPU custom kernels,
+or to TensorFlow functions, or to JAX running on another device. In the latter
+two cases we show how we can support JAX autodiff for the host callbacks,
+by deferring to the reverse-mode AD on the target platform. Then we
+show uses of :func:`id_tap` and :func:`id_print`, which have the restriction
+that they cannot return values from the host to the device.
+These primitives are generally faster
+because they are executed asynchronously with the device code and they also
+support the whole spectrum of JAX transformations. In particular, they can be
+used to tap into and to debug JAX-transformed code.
+
+Using :func:`call` to call a host function and return results to device
+-----------------------------------------------------------------------
+
+Use :func:`call` to invoke a computation on the host and return
+NumPy arrays to the device computation.
+Host computation is useful, e.g., when a device computation needs some data
+that requires I/O on the host, or it needs a library that is available on the
+host and you do not want to code it in JAX.
+For example, eigen decomposition in JAX does not work on TPU.
+We can call the Numpy implementation from any JAX accelerator computation,
+using a host computation::
+
+  # This function runs on the host
+  def host_eig(m: np.ndarray) -> np.ndarray:
+    return np.linalg.eigvals(m)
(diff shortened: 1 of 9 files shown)


tests/infeed_test.py
```diff
@@ -48,6 +48,25 @@ def f(x):
...
+    to_infeed_shape = dict(a=jax.ShapedArray((), dtype=np.float32),
+                           b=jax.ShapedArray((3, 4), dtype=np.int16))
+    @jax.jit
+    def f(x):
+      token = lax.create_token(x)
+      res, token = lax.infeed(token, shape=to_infeed_shape)
+      return res
+
+    device = jax.local_devices()[0]
+    # We must transfer the flattened data, as a tuple!!!
```
gnecula, line 65: I added this test because it was not clear to me how to use `lax.infeed` with pytrees.

tomhennigan:
This is very cool!

Re when the users function throws an exception, two suggestions:

1. Return NaN's of the appropriate shape, many users catch and throw on NaN in their `main()` already.
2. In Colab it is possible to locate and print to the output stream shown in the current cell. IIRC TF team (Tomer?) implemented support for this in `tf.print`.

Please let me know if you want any help testing this on GPU/TPU.

gnecula:
Returning NaN seems dangerous

gnecula:
> Re when the users function throws an exception, two suggestions:
> 
> 1. Return NaN's of the appropriate shape, many users catch and throw on NaN in their `main()` already.
> 2. In Colab it is possible to locate and print to the output stream shown in the current cell. IIRC TF team (Tomer?) implemented support for this in `tf.print`.
> 

I have tried several things here. The only safe solution seems to be to abort the device computation in a way that surfaces a Python exception. Returning anything that allows the computation to continue seems unsafe. Plus, sending NaN is not an option when the computation expects integers. 

I will continue to investigate how to abort the computation. 

The best option I have at the moment is to send something with a shape that is different than what the device expects, counting on a shape check failure. I picked a distinctive shape `s8[12345]` so that we can recognize it in the error
messages. This works for CPU (runtime crash), for GPU (a nice Python error message), but not for TPU. On TPU there is no shape check, the computation will just go ahead with whatever garbage it thinks it got. So, for TPU I am not sending anything and let the computation hang.

jax/experimental/host_callback.py
```diff
@@ -148,25 +220,89 @@ def power3(x):
...
+are processed asyncronously with the device computation.
+
+Exceptions from the user-defined callback functions are logged along with their
+stack traces, but the receiving threads are not stopped. Instead the last
+exception is recorded and the subsequent :func:`barrier_wait` will
+raise :exc:`TapFunctionException` if any exception had occurred
+in one of the tap functions. This exception will include the text and the
+stack trace of the last exception encountered.
+
+One further complication arises for callback functions that must return
```
gnecula, line 254: I'd be curious if anybody has a better idea for how to handle errors in the callback functions

shoyer:
Python's `conurrent.futures` modules may be a useful model for error handling. Error from execution inside a separate thread get saved, so they can be re-raised when `Future.result()` is called:
https://github.com/python/cpython/blob/def7dc3b715d92ccc3e8a6a58ac36779435e41c9/Lib/concurrent/futures/thread.py#L51-L58

There are still the questions of:

1. How to tell XLA to terminate early
2. Checking for errors to re-raise from `host_callback.call()` when finishing execution of XLA compiled code.

jax/experimental/host_callback.py
```diff
@@ -11,12 +11,118 @@
...
+
+Using :func:`call` to call a host function and return results to device
+-----------------------------------------------------------------------
+
+Use :func:`call` to invoke a computation on the host and return
+NumPy arrays to the device computation.
+Host computation is useful, e.g., when a device computation needs some data
+that requires I/O on the host, or it needs a library that is available on the
+host and you do not want to code it in JAX.
+For example, eigen decomposition in JAX does not work on TPU.
```
shoyer, line 43: This is a little misleading -- TPUs support eigen-decomposition for Hermitian matrices only:
```suggestion
For example, eigen decomposition for general matrices in JAX does not work on TPU.
```

jax/experimental/host_callback.py
```diff
@@ -148,25 +220,89 @@ def power3(x):
...
+
+In order to pause the execution until all data from computations already
+started on devices has arrived and has been processed, use :func:`barrier_wait`.
+Note that this is needed only for :func:`id_tap` and :func:`id_print`, which
+are processed asyncronously with the device computation.
+
+Exceptions from the user-defined callback functions are logged along with their
+stack traces, but the receiving threads are not stopped. Instead the last
+exception is recorded and the subsequent :func:`barrier_wait` will
+raise :exc:`TapFunctionException` if any exception had occurred
```
shoyer, line 250: I wonder if it would be better to re-raise the original error instead? That would preserve the full traceback.

jax/experimental/host_callback.py
shoyer: It seems like this is something that should be explicitly supported in XLA.

jax/experimental/host_callback.py
```diff
@@ -278,16 +440,15 @@ def id_tap(tap_func, arg, *, result=None, tap_with_device=False, **kwargs):
...
-        "in the future. Instead, pre-apply keyword arguments, either by using "
-        "a closure or by passing ``functools.partial(tap_func, **kwargs)`` "
-        "instead.",
-        FutureWarning, stacklevel=2)
-    tap_func = functools.partial(tap_func, **kwargs)
+    msg = (
+        "Support for **kwargs in ``id_tap`` has been removed. Instead, "
+        "pre-apply keyword arguments, either by using a closure or by passing "
+        "``functools.partial(tap_func, **kwargs)``.")
+    raise ValueError(msg)
```
shoyer, line 450: `TypeError` would be consistent with what Python would use if you simply omitted `**kwargs` (which would probably be a better idea, eventually):
```suggestion
    raise TypeError(msg)
```

jax/experimental/host_callback.py
```diff
@@ -1066,10 +1619,13 @@ def barrier_tap(dev_idx, _):
...
   logging.vlog(2, f"barrier_wait[{logging_name}]: done")
-  if _outfeed_receiver.num_tap_exceptions > 0:
-    _outfeed_receiver.num_tap_exceptions = 0
+  if _outfeed_receiver.last_tap_exception is not None:
+    last_exception = _outfeed_receiver.last_tap_exception
+    _outfeed_receiver.last_tap_exception = None
     raise TapFunctionException(
-        "There were exceptions during id_tap processing.")
+        "There were exceptions during id_tap processing. "
+        f"Last one was: {last_exception}")
```
shoyer, line 1627: I would suggest either re-raising the original error:
```python
    raise last_exception
```

Or maybe raising `TapFunctionException` _from_ the original error:
```python
    raise TapFunctionException(
        "There were exceptions during id_tap processing.") from last_exception
```
This ends up with an error message like:
```
(original traceback)
(original error)

(barrier_tap traceback)
(TapFunctionException error)
```

tests/host_callback_test.py
```diff
@@ -1466,6 +1605,379 @@ def step(acc, step_nr):
...
+      return 2. * x * x * x
+
+    def f_outside(x):
+      return 2. * call_jax_other_device(lambda x: x * x * x, x, device=self.outside_device)
+
+    res_jax = api.grad(api.grad(f_jax))(5.)
+    # print(f"res_jax = {res_jax}")
+    # print(f"jax_jaxpr = {api.make_jaxpr(api.grad(api.grad(f_jax)))(5.)}")
+    # print(f"outside_jaxpr = {api.make_jaxpr(api.grad(api.grad(f_outside)))(5.)}")
+    # print(f"outside xla = {api.xla_computation(api.grad(api.grad(f_outside)))(5.).as_hlo_text()}")
```
shoyer, line 1976: delete?

tests/host_callback_to_tf_test.py
```diff
@@ -0,0 +1,256 @@
...
+    def tf_vjp_fun(arg_and_ct_res):
+      """Invoke TF gradient; used with hcb.call."""
+      arg, ct_res = arg_and_ct_res
+      arg_var = tf.Variable(arg)
+      with tf.GradientTape(persistent=True) as tape:
+        res = tf_fun(arg_var)
+
+      dres_darg = tape.gradient(res, sources=arg_var,
+                                unconnected_gradients=tf.UnconnectedGradients.ZERO)
+      return dres_darg * ct_res
```
shoyer, line 87: You can let TensorFlow do this multiplication for you:
```suggestion
      dres_darg = tape.gradient(res, sources=arg_var,
                                output_gradients=ct_res,
                                unconnected_gradients=tf.UnconnectedGradients.ZERO)
      return dres_darg
```

tests/host_callback_to_tf_test.py
```diff
@@ -0,0 +1,256 @@
...
+                                  unconnected_gradients=tf.UnconnectedGradients.ZERO)
+        tf.nest.assert_same_structure(dres_darg, arg)
+        scaled_dres_darg = tf.nest.map_structure(lambda d: d * ct_res_, dres_darg)
+        nonlocal accumulator
+        accumulator = (scaled_dres_darg if accumulator is None
+                       else tf.nest.map_structure(lambda x, y: x + y,
+                                                  accumulator, scaled_dres_darg))
+
+      tf.nest.map_structure(acc_ct, res, ct_res)
+      return accumulator
```
shoyer, line 140: This looks harder than it needs to be :). I wonder if setting `output_gradients` in a single call to `tape.gradient()` would suffice?

tests/host_callback_to_tf_test.py
```diff
@@ -0,0 +1,256 @@
...
+    arg = res  # residual is the primal argument
+
+    def tf_vjp_fun(arg_and_ct_res):
+      """Invoke TF gradient; used with hcb.call."""
+      arg, ct_res = arg_and_ct_res
+
+      def make_var(a):
+        return a if isinstance(a, tf.Variable) else tf.Variable(a)
+
+      arg_var = tf.nest.map_structure(make_var, arg)
```
shoyer, line 121: I think another viable option would be to simply use `tf.nest.map_structure(tf.convert_to_tensor, arg)`. Tensors also work fine in GradientTape.

tests/host_callback_to_tf_test.py
```diff
@@ -0,0 +1,256 @@
...
+      return 3. * jnp.sin(2. * x)
+
+    def f_outside(x):
+      return 3. * call_tf(tf.math.sin, 2. * x, result_shape=x)
+
+    x = 4.
+    self.assertAllClose(f_jax(x), f_outside(x))
+
+    # print(f"outside_jaxpr = {api.make_jaxpr(f_outside)(xy)}")
+    # print(api.xla_computation(f_outside)(4.).as_hlo_text())
```
shoyer, line 200: delete? or add to the message field of `assertAllClose`?

tests/host_callback_to_tf_test.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def call_tf_no_ad(tf_fun: Callable, arg, *, result_shape):
+  """The simplest implementation of calling to TF, without AD support.
+
+  We must use hcb.call because the TF invocation must happen outside the
+  JAX staged computation."""
+
+  def tf_to_numpy(t):
+    return t.numpy() if isinstance(t, tf.Tensor) else t
```
shoyer, line 49: Maybe add a comment noting that ideally we would convert directly on device (e.g., with DLpack) rather than using NumPy?

tests/host_callback_to_tf_test.py
gnecula: I agree, and it took me a while to arrive at this, but with my limited knowledge of TF this is how I managed to make it work. The difficulty is that we have pytree args and results, which arise naturally when you start doing higher-order differentiation.

tests/host_callback_test.py
gnecula: Done

tests/host_callback_to_tf_test.py
gnecula: I did not realize that `t.numpy()` would copy the array. I have changed the code to use np.array(memoryview(t)) instead, and added a comment.

tests/host_callback_to_tf_test.py
gnecula: Actually this does not seem to work for higher-order differentiation, some dependencies get lost and I end up with 0 gradients. Perhaps you can try to clean this up once I merge this branch?

tests/host_callback_to_tf_test.py
gnecula: Done

gnecula:
Here is [the API documentation generated from this PR](https://jax.readthedocs.io/en/print_refactor/jax.experimental.host_callback.html)

jax/experimental/host_callback.py
gnecula: I have implemented the latter suggestion (`raise ... from last_exception`), and it works as you predicted only for `id_tap`. For errors during `call`, at least on CPU, the XLA runtime crashes (because we infeed to it a bad shape) and in that case the `from last_exception` does not seem to work.

jax/experimental/host_callback.py
gnecula: I am going to resolve this one, continuing the discussing in the similar thread below.

jax/experimental/host_callback.py
```diff
@@ -1578,7 +1579,7 @@ def barrier_wait(logging_name: Optional[str] = None):
 
   Waits until all outfeed from computations already running on all devices
   has been received and processed by the Python callbacks. Raises
-  TapFunctionException if there were exceptions while processing the callbacks.
+  Callback if there were exceptions while processing the callbacks.
```
shoyer, line 1582:
```suggestion
  CallbackException if there were exceptions while processing the callbacks.
```

tests/host_callback_to_tf_test.py
```diff
@@ -46,7 +48,8 @@ def call_tf_no_ad(tf_fun: Callable, arg, *, result_shape):
   JAX staged computation."""
 
   def tf_to_numpy(t):
-    return t.numpy() if isinstance(t, tf.Tensor) else t
+    # Turn the Tensor to NumPy array without copying.
+    return np.array(memoryview(t)) if isinstance(t, tf.Tensor) else t
```
shoyer, line 52: I was actually more concerned about copying from GPU -> CPU, but indeed CPU copies are also a concern.

Unfortunately `np.array(memoryview(t))` still makes a copy (`asarray` copies by default). You would need `np.array(memoryview(t), copy=False)` or `np.asarray(memoryview(t))`. The `memoryview()` part is also likely optional, given that tensorflow implements NumPy's `__array__` conversion protocol.

shoyer (approved):

jax/experimental/host_callback.py
gnecula: Done

tests/host_callback_to_tf_test.py
gnecula: Understood. The example that I tried to copy was using `np.asarray` indeed, but I mis-copied it :-) I changed to asarray

PR #5244, opened 2020-12-22
factor convert_closure from ode to custom_derivatives

### LINKED ISSUE #5222: utility for closure-conversion in higher-order functions with custom derivatives
When a higher-order python function is associated with a custom derivative (e.g. via `custom_vjp`), differentiation doesn't handle the closures of its function arguments. A workaround is to accept auxiliary arguments and thread them back into input functions that otherwise would have closed over them. An example sketch:

```python
@custom_vjp
def minimize(objective_fn, x0, objective_aux_args=()):
  # ...
  t = objective_fn(..., *objective_aux_args)
  # ...
```

An alternative approach taken by `jax.experimental.ode.odeint` is to stage function arguments out to jaxprs, converting closures in the process. We could try and factor [this setup](https://github.com/google/jax/blob/943c7794f92f32c4796cf980eedd5a11c86eea66/jax/experimental/ode.py#L50) into a common utility for use in other similar applications.

froystig:
Adds a closure-conversion utility to custom derivatives, and in turn to our API, useful when setting up custom derivatives for higher-order functions. This is based on a subroutine in `ode`, with minor modifications to generalize it for wider use.

Fixes #5222

### DIFF
--- a/jax/custom_derivatives.py
+++ b/jax/custom_derivatives.py
@@ -841,6 +841,61 @@ class Residuals:
 
 
 def closure_convert(fun, example_args):
+  """Closure conversion utility, for use with higher-order custom derivatives.
+
+  To define custom derivatives such as with ``jax.custom_vjp(f)``, the target
+  function ``f`` must take, as formal arguments, all values involved in
+  differentiation. If ``f`` is a higher-order function, in that it accepts as an
+  argument a Python function ``g``, then values stored away in ``g``'s closure
+  will not be visible to the custom derivative rules, and attempts at AD
+  involving these values will fail. One way around this is to convert the
+  closure by extracting these values, and to pass them as explicit formal
+  arguments across the custom derivative boundary. This utility carries out that
+  conversion. More precisely, it closure-converts the function ``fun``
+  specialized to the types of the arguments given in ``example_args``.
+
+  When we refer here to "values in the closure" of ``fun``, we do not mean the
+  values that are captured by Python directly when ``fun`` is defined (e.g. the
+  Python objects in ``fun.__closure__``, if the attribute exists). Rather, we
+  mean values encountered during the execution of ``fun`` on ``example_args``
+  that determine its output. This may include, for instance, arrays captured
+  transitively in Python closures, i.e. in the Python closure of functions
+  called by ``fun``, the closures of the functions that they call, and so forth.
+
+  The function ``fun`` must be a pure function.
+
+  Example usage::
+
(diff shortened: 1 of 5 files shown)


shoyer:
This is great! One thing I might add to the docs is that you can get a pytree compatible _function_ with `tree_util.Partial(*closure_convert(objective_fn, x0))`.

Another good use case for this helper would be `lax.custom_root`, which with `closure_convert` (replacing `_initial_style_jaxpr`) can now be implemented entirely with JAX public APIs.

jax/custom_derivatives.py
```diff
@@ -837,3 +838,107 @@ def tree_flatten(self):
...
+  that determine its output. This may include, for instance, arrays captured
+  transitively in Python closures, i.e. in the Python closure of functions
+  called by ``fun``, the closures of the functions that they call, and so forth.
+
+  The function ``fun`` must be a pure function.
+
+  Example usage::
+
+    def minimize(objective_fn, x0):
+      converted_fn, *aux_args = closure_convert(objective_fn, x0)
```
mattjj, line 870:
```suggestion
      converted_fn, aux_args = closure_convert(objective_fn, x0)
```

jax/custom_derivatives.py
```diff
@@ -837,3 +838,107 @@ def tree_flatten(self):
...
+  values that are captured by Python directly when ``fun`` is defined (e.g. the
+  Python objects in ``fun.__closure__``, if the attribute exists). Rather, we
+  mean values encountered during the execution of ``fun`` on ``example_args``
+  that determine its output. This may include, for instance, arrays captured
+  transitively in Python closures, i.e. in the Python closure of functions
+  called by ``fun``, the closures of the functions that they call, and so forth.
+
+  The function ``fun`` must be a pure function.
+
+  Example usage::
```
mattjj, line 867: This is a great example. Could you add a test that basically copy-pastes it? That way we get a test almost for free, and we're confident the docstring is accurate! (The test need not do any actual optimization; its implementation could be trivial perhaps.)

mattjj (approved):
Nice! Idea for a test mentioned below.

mattjj:
@shoyer I haven't thought through this, but I think that might not interact well with `nondiff_argnums`. Perhaps that can be sorted out but we might have to give it some thought and try out some examples.

jax/custom_derivatives.py
froystig: Good call. Done.

PR #5249, opened 2020-12-23
Add breaking changes about jax.numpy.pad to change log

minoring:
https://github.com/google/jax/pull/5229#issuecomment-749028088

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -24,6 +24,11 @@ jax 0.2.8 (Unreleased)
   * ``host_callback.outfeed_receiver`` has been removed (it is not necessary,
     and was deprecated a few months ago).
 
+* Breaking changes:
+
+  * ``jax.numpy.pad`` now take keyword arguments. Positional argument ``constant_values``
+    has been removed. In addition, passing unsupported keyword arguments raises an error.
+
 jax 0.2.7 (Dec 4 2020)
 ----------------------
 * `GitHub commits <https://github.com/google/jax/compare/jax-v0.2.6...jax-v0.2.7>`_.


docs/CHANGELOG.rst
```diff
@@ -24,6 +24,11 @@ jax 0.2.8 (Unreleased)
   * ``host_callback.outfeed_receiver`` has been removed (it is not necessary,
     and was deprecated a few months ago).
 
+* Breaking changes:
+
+  * ``jax.numpy.pad`` now take keyword arguments. Extra parameter (``constant_values``)
```
gnecula, line 29: I would propose "positional argument ``constant_values`` has been removed. ..."

jax/_src/numpy/lax_numpy.py
```diff
@@ -2543,17 +2543,41 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
     return _pad_empty(array, pad_width)
 
   else:
```
gnecula, line 2545: I do not think that we need these changes, they were merged already in #5229.

gnecula (changes requested):
In this particular case we were able to fix the breaking changes in the users' code so we did not revert your changes. If we did, the revert would have been visible in github also.

docs/CHANGELOG.rst
minoring: Thanks for the proposal.
Updated! PTAL.

jax/_src/numpy/lax_numpy.py
minoring: Yes, I misunderstood.
I thought #5229 will be reverted and this new PR is merged after fixing the breaking changes.

I updated PR only modifying change log. PTAL:)

gnecula (approved):

PR #5251, opened 2020-12-23
Fix header guard typo.

zhangqiaorjc:


mattjj (approved):
Looks like just a typo IIUC!

inailuig:
@zhangqiaorjc Thanks for catching this! 
Looks like I also forgot to rename the header guards in `rocm_gpu_kernel_helpers.h`: https://github.com/google/jax/blob/6cddc2fa779844522a3220825eef6bff7d472976/jaxlib/rocm_gpu_kernel_helpers.h#L16

PR #5260, opened 2020-12-24
Use XLA AllGather op for GPU (when supported).

chr1sj0nes:
My benchmarks suggest a speed-up of ~2-2.5x for larger inputs.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -670,10 +670,6 @@ def _expand(dim, size, index, x):
   out = lax.full(shape, lax._const(x, 0))
   return lax.dynamic_update_index_in_dim(out, x, index, dim)
 
-def _allgather(x, dim, size, index, axis_name, axis_index_groups=None):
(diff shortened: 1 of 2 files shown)


mattjj:
Woohoo, awesome!

How come the tests need to be skipped? (Sorry if this is obvious.)

chr1sj0nes:
Same issue as the `all_to_all` tests - the fallback implementation calls `axis_index`, which causes an error when omni-staging is disabled. I don't exactly understand why but, as discussed internally, probably not worth spending the time to investigate with omni-staging becoming mandatory in the near future.

mattjj:
We had an idea (copying here from chat): we can switch on `config.omnistaging_enabled` in the `all_gather` function, and basically call into the new primitive-based code only if omnistaging is enabled. That way we don't have to disable any tests (until we get rid of disable_omnistaging entirely).

chr1sj0nes:
I've changed it so we fall back to the old behaviour when omni-staging is disabled and re-enabled the tests.

mattjj (approved):
Awesome work, and a welcome speedup!

PR #5277, opened 2020-12-29
Implement "function" padding mode in jax.numpy.pad

minoring:
Implemented "function" padding mode in jax.numpy.pad

There was an issue about this implementation which was
> the function version would have to be implemented differently from numpy, because numpy's version has the function to modify the array in-place, while JAX arrays are immutable. https://github.com/google/jax/issues/5010#issuecomment-744621850

I think we can circumvent this problem by expecting different argument from numpy that `func` should return the modified array (while numpy `func` modify array in-place).

Related #5010

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2470,9 +2470,6 @@ def _pad_empty(array, pad_width):
 
 
 def _pad_func(array, pad_width, func, **kwargs):
-  # Note: In numpy, `func` should modify a rank 1 array in-place.
-  # However, since Jax arrays are immutable, `func` should return
-  # the modified array.
   pad_width = _broadcast_to_pairs(pad_width, ndim(array), "pad_width")
   padded = _pad_constant(array, np.array(pad_width), 0)
   for axis in range(ndim(padded)):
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2469,6 +2469,17 @@ def _pad_empty(array, pad_width):
   return array
 
 
+def _pad_func(array, pad_width, func, **kwargs):
+  # Note: In numpy, `func` should modify a rank 1 array in-place.
+  # However, since Jax arrays are immutable, `func` should return
+  # the modified array.
```
jakevdp, line 2475: The explanation of this difference should be added to the `jnp.pad()` docstring via the `lax_description` argument; search for `lax_description` in this file to see examples.

tests/lax_numpy_test.py
```diff
@@ -1516,6 +1516,31 @@ def testPadKwargs(self):
...
+
+    arr = np.arange(6).reshape(2, 3)
+    np_res = np.pad(arr, 2, np_pad_with)
+    jnp_res = jnp.pad(arr, 2, jnp_pad_with)
+    np.testing.assert_equal(np_res, jnp_res)
+
+    arr = np.arange(24).reshape(2, 3, 4)
+    np_res = np.pad(arr, 1, np_pad_with, padder=100)
+    jnp_res = jnp.pad(arr, 1, jnp_pad_with, padder=100)
+    np.testing.assert_equal(np_res, jnp_res)
```
jakevdp, line 1542: It would be worth adding a jit-compiled test as well; perhaps use `self._CompileAndCheck` for convenience.

jakevdp (changes requested):
Looks good! Just a couple minor changes.

jax/_src/numpy/lax_numpy.py
minoring: Thanks for the review :)
I added `lax_description` and removed the comment above.

tests/lax_numpy_test.py
minoring: I added jit-compiled test :) PTAL.

jax/_src/numpy/lax_numpy.py
jakevdp: Thanks!

jakevdp (approved):
Looks good! I'll run this through some additional tests and then get it merged. Thanks for the contribution!

PR #5279, opened 2020-12-29
Nit: fix typo in custom_derivatives.py ("separate")

8bitmp3:
Noticed a small thing while reading the `jit` compile docs (https://jax.readthedocs.io/en/latest/jax.html#just-in-time-compilation-jit). 

Custom JVPs docstring ("seperate" -> "separate"):

```
class custom_jvp:
  """Set up a JAX-transformable function for a custom JVP rule definition.
...
  :py:func:`~jax.custom_jvp.defjvp`, and allows you to provide separate
  definitions for the partial derivatives of the function w.r.t. each of its
  arguments.
...


jakevdp:
Looks good, thanks!

PR #5281, opened 2020-12-29
Nit: lint (grammar) JAX F.A.Q.

8bitmp3:
Stumbled on a few tiny tiny things when reading the JAX F.A.Q. docs (I need to get a life).

Indefinite article ("a [symptom]", not "an"):

```
This is usually a symptom of calling your function generating a large amount of
of code in JAX’s internal representation, typically because it makes heavy use of
Python control flow such as for loop.
```

No comma (I think) after "another example" (beginning of sentence):
```
Another example is when explicitly
when explicitly casting a concrete tracer value to a regular type, e.g., int(x) or 
x.astype(float). 
```


jakevdp:
Proof-reading is always helpful - thanks!

PR #5282, opened 2020-12-29
Fix broadcast_in_dim transpose issue.

### LINKED ISSUE #5276: lax.broadcast_in_dim transposition rule
I noticed that the transposition rule of lax.broadcast_in_dim assumes that the input operand has a shape that matches the broadcast dimensions in the target shape. That is, we cannot use the following
```
x = np.ones((1, 3, 1))
lax.broadcast_in_dim(x, (2, 3, 4), broadcast_dimensions=(0, 1, 2))
```
The above would compute correctly but will fail in transpose (only if core.skip_checks == False). 

Instead one must do:
```
x = np.ones((1, 3, 1))
lax.broadcast_in_dim(x.reshape((3,)), (2, 3, 4), broadcast_dimensions=(1,))
```

Either we must add a new parameter to broadcast_in_dim with the input shape, or at the very least we ought to change the shape rule to reject unsupported uses.

jakevdp:
Fixes #5276

The `broadcast_in_dim_p` transpose rule currently does not correctly handle broadcasted explicit dimensions.

To handle this correctly, the transpose rule needs information about the input shapes, which is not forwarded by `ad.deflinear()`, but is forwarded by `ad.deflinear2()`.

This change modifies the transpose rule to use `deflinear2`, and return an array of the correct shape as confirmed by [this check](https://github.com/google/jax/blob/2ebd5285c8056989a92d49812ae4f8bfef022a26/jax/interpreters/ad.py#L180) when `core.skip_checks = False`.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -3302,8 +3302,6 @@ def _broadcast_in_dim_shape_rule(operand, *, shape, broadcast_dimensions):
   return shape
 
 def _broadcast_in_dim_transpose_rule(ct, operand, *, shape, broadcast_dimensions):
-  if type(ct) is ad_util.Zero:
-    return ad_util.Zero
   shape_in = operand.aval.shape
   unit_dimensions = tuple(i for i, s in enumerate(shape_in) if s == 1)
   bdims = tuple(np.delete(broadcast_dimensions, unit_dimensions))
@@ -3323,8 +3321,7 @@ def _broadcast_in_dim_batch_rule(batched_args, batch_dims, *, shape,
 broadcast_in_dim_p = standard_primitive(
     _broadcast_in_dim_shape_rule, _input_dtype, 'broadcast_in_dim')
 broadcast_in_dim_p.def_impl(_broadcast_in_dim_impl)
-ad.primitive_jvps[broadcast_in_dim_p] = partial(ad.linear_jvp, broadcast_in_dim_p)
-ad.primitive_transposes[broadcast_in_dim_p] = _broadcast_in_dim_transpose_rule
+ad.deflinear2(broadcast_in_dim_p, _broadcast_in_dim_transpose_rule)
 batching.primitive_batchers[broadcast_in_dim_p] = _broadcast_in_dim_batch_rule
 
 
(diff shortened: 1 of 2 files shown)


jax/_src/lax/lax.py
```diff
@@ -3301,9 +3301,14 @@ def _broadcast_in_dim_shape_rule(operand, *, shape, broadcast_dimensions):
 
   return shape
 
-def _broadcast_in_dim_transpose_rule(t, *, shape, broadcast_dimensions):
-  axes = tuple(np.delete(range(len(shape)), broadcast_dimensions))
-  return [_reduce_sum(t, axes)]
+def _broadcast_in_dim_transpose_rule(ct, operand, *, shape, broadcast_dimensions):
+  if type(ct) is ad_util.Zero:
+    return ad_util.Zero
```
mattjj, line 3306: We need to return an instance of `ad_util.Zero` here, not the class itself.

```suggestion
    return ad_util.Zero(operand.aval)
```

mattjj (approved):
This looks okay to me, but how hard would it be just to adapt the `deflinear` wrapper to forward the right input aval information once and for all? If we don't do that in this PR, then how about leaving a TODO so we don't forget to put this back to an `ad.deflinear` when we do make that upgrade?

jax/_src/lax/lax.py
jakevdp: I was following `linear_transpose`, which returns the class rather than an instance: https://github.com/google/jax/blob/19b3cf9a7222768aecfb606b54796367427c5fa8/jax/interpreters/ad.py#L440

Should that be changed as well?

jakevdp:
It turns out that `ad.deflinear2` does exactly what we need; I switched to using that approach.

jax/_src/lax/lax.py
jakevdp: This is no longer relevant to this PR; I'll open a separate issue regarding returning the `Zero` class in `ad.py`.

mattjj: Good catch! Yeah, I think that's a bug. (We used to do something like that, where we had a singleton class for zero, but when we changed zeros to carry avals we should've replaced those things.)

I'm surprised it wasn't caught by [this assertion](https://github.com/google/jax/blob/62f943435f20345aa2cafd6e0201de77ae459352/jax/interpreters/ad.py#L169). We must not be hitting that `linear_transpose` case somehow...

mattjj:
Good find! Now that you mention it, I think we added `deflinear2` in #2410 with the goal of replacing `deflinear` with it, but then didn't make progress on that.

gnecula:
Thanks!

PR #5285, opened 2020-12-29
Re-land #4850 weak types change

jakevdp:
This re-lands the changes reverted in #5250

We need to make certain the downstream issues mentioned there are fixed before merging this again. *Update, 9-Feb 2021 worked with downstream users and everything is good to go!*

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -28,6 +28,13 @@ jax 0.2.10 (Unreleased)
 
 * Breaking changes:
 
+  * JAX's promotion rules were adjusted to make promotion more consistent and
+    invariant to JIT. In particular, binary operations can now result in weakly-typed
+    values when appropriate. The main user-visible effect of the change is that
+    some operations result in outputs of different precision than before; for
+    example the expression ``jnp.bfloat16(1) + 0.1 * jnp.arange(10)``
+    previously returned a `float64` array, and now returns a `bfloat16` array.
+    JAX's type promotion behavior is described at :ref:`type-promotion`.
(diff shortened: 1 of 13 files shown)


mattjj (approved):
Thanks for figuring this out. It's extremely tricky and subtle.

tberghammer:
The `_standard_weak_type_rule` looks incorrect for `multiple_results=True` because `standard_primitive` would expect it to return a tuple of bool's with the same number of results as the primitive but it only returns a single bool value.

Not sure if it worth the complexity for the very few multi-result primitives we have but if we decide to not support that case then we should report an error from standard_primitive directly as the current error message is quite confusing.

Can provide an example if it is useful but will require a bit of work.

jakevdp:
Thanks for pointing that out - I think it's definitely worth fixing. Is that something you're interested in doing? If not, I can put it on my TODO list.

mattjj:
It may be a good idea to make `standard_primitive` only generate single-result primitives. After all, it's only meant to handle the common case, and the logic for handling both single- and multi-result primitives makes it pretty complicated. Complicated logic makes bugs more likely.

If we need a helper function for multi-result primitives, then maybe we could split that out as a separate function rather than having it as a flag. That'd even cut down on the runtime logic! But also there are few enough multi-result primitives that we generate with this helper that it's not clear we need one.

@jakevdp I can take this, and I'll fix the weak type rule at the same time.

mattjj:
(We already removed the `multiple_results` option in `standard_primitive` in #5702, and IIRC it was only used for `reduce_p`. But I figure it may be a good idea to land that change separately, as #5702 may not land immediately.)

PR #5286, opened 2020-12-30
[shapecheck] Improve the shape computations to work better with the n…

gnecula:
[shapecheck] Increase coverage of shapecheck, to work with the new sound Poly implementation

This is a follow-up to #4878, which fixed unsoundness in the implementation of Poly (equality and inequality operators
now may return UndefinedPoly). As a side-effect, a number of JAX shape checking rules that depended on the
unsound behavior have stopped working with non-concrete shapes. In order to land #4878, I had to disable
some tests in masking_test.py and jax2tf/tests/shape_poly_test.py.

A summary of the changes here:

  * fixed the shape rules for broadcast, pad, reshape, dot_general, conv, squeeze slice. Added more tests to `masking_test.py`. See below for more details. 
  * removed all the `SkipTest` introduced in #4878. In addition to changes to the shape rules, I had to change the input shape specifications to ensure that the input dimensions are "big enough" for the test. E.g., instead of `in_shape="n"`, I had to
    use `in_shape="n + 1"` if the test was doing negative padding.
  * indirectly this restores the coverage of the jax2tf batch polymorphism where it was before #4878 (most Flax models can be converted batch polymorphically)
  * added better error messages for when Poly inequality comparisons are indeterminate
  * fixed `masking.parse_spec("()")`, was failing before
  * added a missing check in the padding shape rule to prevent cases when the resulting shape has negative-sized axes.
    Previously, these were caught as Internal Errors by XLA. Added new tests to lax_test.py

Next, I discuss a bit the philosophy behind the changes proposed here.

Definition of **soundness of shape polymorhism (and `shapecheck`)**: given a function `f` and `input_shape`/`output_shape` two shape polynomials over a set of shape variables, 
if `shapecheck(in_shape, out_shape)(f)` succeeds, then for any valuation of the shape variables with **non-negative** integers,
let `s_in` and `s_out` be the values of `in_shape` and `out_shape` respectively for the given valuation. 
Then, for any input `x` whose shape is `s_in` we have that `f(x)` **succeeds with an output of shape `s_out`**. 

There are many complex computations with shapes in JAX, , e.g., in the process of lowering
`jax.numpy` to JAX primitives, in the abstract evaluation rules for the primitives, and even in the implementation of the
transformation rules. We want to reuse these with minimal changes, for the case when the shapes are
polymorphic. Adding escape hatches such as `if shape_is_polymorphic(..): ...` makes it hard to ensure soundness.

The current design is to introduce a class `Poly` that duck-types integers and generalizes all shape computations 
in JAX to work over tuples of `Poly`.
The change in #4878 ensures that shape equality and inequality comparisons throw `UndefinedPoly` if the comparison cannot be expressed soundly as a boolean. 
For example, if the Python code contains `shape[0] == 1` and the value of `shape[0]` is the polynomial `m + n`, then it would be unsound to assume that the comparison should be represented as `False`, because there are values for `n` and `m` that would return `True`, and would thus make the Python code that different paths.

**If we avoid all explicit polymorphism checks in the shape computation code, then we have soundness, as long as
the Poly implementation of the arithmetic operators is sound.**

Unfortunately, there are some cases when the ergonomics would suffer (unfriendly error messages or cases that would raise errors too aggressively). We propose to add a limited number of
mechanisms to use in JAX shape computation code to improve the usability, while making soundness easy to review.
The following use cases show a few common instances:

1. Shape computation for error reporting.

There are many instances in JAX of the following form:
```
if shape_comp(): raise ValueError(msg)
```

We want to report the intended error message even if `shape_comp()` raises. We propose the following:
```
with masking.shape_error(ValueError(msg)) as e:
  if shape_comp(): raise e
```

This constructs the exception early, and if the body of the `with` throws `UndefinedPoly`, then
we still raise the intended exception, with an expanded error message:
```
raise type(e)(e.message + ". Polynomial comparison 'n == 1' is inconclusive")
```

2. A common instance of the error checking code is for simple equality comparisons. We propose a simpler form:

```
if not masking.must_equal(v1, v2):
   raise ValueError(msg)
```

The `must_equal` function returns False if the comparison is indeterminate, e.g., `must_equal(1, 1)`, `must_equal(n + 1, n + 1)`, 
but not for `must_equal(1, 2)` nor for `must_equal(n, 1)`.

The programmer must ensure that the code is sound even in the case when `must_equal` returns False with polymorphic shapes but returns True for concrete shapes. 
In the above usage this is easy to see, because if `must_equals` returns False, then we abort, which is always sound during
polymorphic execution.

3. There are a few instances when we need to check an either-or equality, e.g., for broadcasting:
```
if op_dim != 1 and op_dim != target_dim:
   raise ValueError(msg)
```

The problem here is that if `op_dim` and `target_dim` are both `n`, then `op_dim != 1` will raise before we get a chance to see that `op_dim == target_dim`.

For this pattern we propose
```
if not masking.must_equal_or(op_dim, 1, target_dim):
  raise ValueError(msg)
```

4. The most complicated cases are when we have a sequence of nested conditionals on shapes. A good example are the
dilation shape computation, to compute "1 + (0 if shape == 0 else dilation * (shape - 1))",
which was before written as:

```
return np.where(shape == 0, 0,
                         np.multiply(dilation, np.subtract(shape, 1)) + 1)
```

The problem here is that the `shape == 0` array comparison may throw `UndefinedPoly` for some of
the dimensions even if the `dilation` is 1 on that dimension. An alternative way to write this is:

```
[shape[i] if masking.must_equal(dilation[i], 1) else (0 if shape[i] == 0 else 1 + dilation[i] * (shape[i] - 1))
 for i in range(len(shape))]]
```

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -38,7 +38,7 @@ from jax import tree_util
 from jax.config import flags, config
 from jax.core import (Primitive, _canonicalize_dimension, UnshapedArray,
                       ShapedArray, ConcreteArray, raise_to_shaped,
-                      abstract_token, canonicalize_shape)
+                      abstract_token, canonicalize_shape, Shape)
 from jax.abstract_arrays import array_types
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
@@ -48,7 +48,7 @@ from jax.interpreters import invertible_ad as iad
 from jax.interpreters import batching
 from jax.interpreters import masking
 from jax.util import (cache, safe_zip, partial, prod, safe_map, canonicalize_axis,
-                      split_list)
+                      split_list, unzip3)
 from jax.tree_util import tree_map
 from jax.lib import pytree
 from jax.lib import xla_bridge
@@ -66,7 +66,6 @@ _reduce = functools.reduce
 
 Array = Any
 DType = Any
-Shape = Sequence[int]
 
 def _try_broadcast_shapes(shapes):
   assert shapes
@@ -76,14 +75,14 @@ def _try_broadcast_shapes(shapes):
   if not rank: return ()  # scalar case
   result_shape = [None] * rank
   for i, sizes in enumerate(zip(*shapes)):
-    if sizes[:-1] == sizes[1:]:
-      result_shape[i] = sizes[0]  # all equal sizes for this dimension
+    # Dim[i]: Must have equal sizes other than 1-sized axes
+    not_ones = [s for s in sizes if not masking.must_equal(s, 1)]
+    if masking.must_equal(not_ones[:-1], not_ones[1:]):
+      result_shape[i] = not_ones[0] if not_ones else 1
     else:
-      sizes = [d for d in sizes if d != 1]
-      if sizes[:-1] != sizes[1:]:
-        return None  # must have equal sizes other than 1-sized axes
-      result_shape[i] = sizes[0] if sizes else 1
+      return None
   return tuple(result_shape)
+
 
 @cache()
 def broadcast_shapes(*shapes):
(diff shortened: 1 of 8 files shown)


mattjj (changes requested):
Revising core lax.py functions for the needs of masking seems like a step in the wrong direction. Masking is experimental, and it's more likely we want to decrease rather than increase its use right now so as to revise or replace it.

In other words, I think it's better to leave disabled the masking tests that #4878 had to disable, rather than adapting more of lax.py to masking needs.

If there's another goal here, other than re-enabling tests that #4878 had to disable, maybe we can figure out a way to accomplish it without growing the use of masking logic in lax.py?

PR #5288, opened 2020-12-30
Use f-strings for string formatting rather than .format(...).

aslanides:
This should tidy up/improve readability somewhat.

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -200,9 +200,9 @@ def _python_jit(
     if _jit_is_disabled():
       return fun(*args, **kwargs)
     if max(static_argnums + donate_argnums, default=-1) >= len(args):
-      msg = ("jitted function has static_argnums={}, donate_argnums={} but "
-             "was called with only {} positional arguments.")
-      raise ValueError(msg.format(static_argnums, donate_argnums, len(args)))
+      raise ValueError(f"jitted function has static_argnums={static_argnums}, "
+                       f"donate_argnums={donate_argnums} but "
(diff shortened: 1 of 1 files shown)


jax/api.py
```diff
@@ -1198,12 +1197,12 @@ def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None) -> F:
...
-           "those types as leaves, but got {}.")
-    raise TypeError(msg.format(in_axes))
+           f"those types as leaves, but got {in_axes}.")
+    raise TypeError(msg)
   if not all(isinstance(l, (type(None), int)) for l in out_axes_):
     msg = ("vmap out_axes must be an int, None, or (nested) container with "
-           "those types as leaves, but got {}.")
-    raise TypeError(msg.format(out_axes))
+           f"those types as leaves, but got {out_axes}.")
+    raise TypeError(msg)
```
jakevdp, line 1205: nit: remove the `msg` variable here, as you did elsewhere?

jakevdp (commented):
Looks good, this is something that I've been meaning to do some time :grin:

Only comment: there are still a few places that use the pattern `msg = "..."; raise TypeError(msg)`. I have a slight preference toward avoiding the `msg` variable and passing the f-string literal to the TypeError. What do you think?

jax/api.py
aslanides: Done.

aslanides:
Thanks Jake! PTAL.

jakevdp (approved):

jakevdp:
Thanks!

PR #5291, opened 2020-12-30
Add setup classifiers fields for PyPI

aslanides:
This is following https://packaging.python.org/tutorials/packaging-projects/.

We might want to add other fields in future, e.g. license, etc.

This metadata shows up conveniently in the left-hand column on PyPI, e.g. https://pypi.org/project/dm-haiku/


jakevdp (approved):
Looks good, thanks!

I do all local development in Python 3.8, but was surprised to see we have no CI coverage for that version!

PR #5294, opened 2020-12-30
avoid cache misses when evaluating cond primitives

### LINKED ISSUE #5210: jax.lax.switch host memory leak
I'm encountering a memory leak in code that repeatedly calls `jax.lax.switch`. Here's a small example that reproduces the issue in colab:

```python
import resource, gc
import numpy as np
import jax, jax.numpy as jp

d = 100

def do_thing():
  i = jp.array(np.random.rand() > 0.5, dtype="int32")
  x = jp.array(np.random.randn(d))
  y = jp.array(np.random.randn(d))
  def fn(i, x, y):
    return jax.lax.switch(i, [(lambda _: x), (lambda _: y)], None)
  fn(i, x, y)

stats = np.zeros(1000)
for i in range(len(stats)):
  gc.collect()
  do_thing()
  gc.collect()
  stats[i] = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss # (kilobytes)

plt.plot(stats)
plt.gca().ticklabel_format(style="plain", useOffset=False)
```

jax_switch_memory_leak

froystig:
Cache the jaxprs that we form for branches so that the branch params we pass when binding the cond primitive hit the cache for primitive computations.

Simplify the jaxpr branch setup code along the way.

fixes #5210

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -44,7 +44,7 @@ from jax.interpreters import batching
 from jax.interpreters import masking
 from jax.lib import xla_bridge as xb
 from jax.lib import xla_client
-from jax.util import (partial, unzip2, unzip4, safe_map, safe_zip, split_list,
+from jax.util import (partial, unzip2, unzip3, safe_map, safe_zip, split_list,
                       cache, extend_name_stack)
 from jax.tree_util import (tree_flatten, tree_unflatten, treedef_is_leaf,
                            treedef_children, treedef_tuple, tree_multimap,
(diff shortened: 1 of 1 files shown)


froystig:
This is a focused fix that also motivates thinking about related improvements across the codebase, such as overriding jaxpr identity (hash, eq, etc., for caching purposes), when and where to cache, whether we can have size-based rather than count-based caches, whether to conditionally toggle caching based on flags for internal debugging purposes, and so on.

jax/_src/lax/control_flow.py
```diff
@@ -2606,6 +2607,8 @@ def omnistaging_disabler() -> None:
   global _initial_style_open_jaxpr, _initial_style_jaxpr, \
       _initial_style_jaxprs_with_common_consts
 
+  from jax.util import unzip4
```
mattjj, line 2610:
```suggestion
```

mattjj (approved):
Nice! Thanks for this deep digging.

jax/_src/lax/control_flow.py
mattjj: Oops, this was mistaken!

PR #5298, opened 2020-12-30
Switch internal code from deflinear to deflinear2

jakevdp:
As discussed with @mattjj


mattjj (approved):
Woo, thanks for doing this!

PR #3796, opened 2020-07-19
WIP: Add bicgstab implementation

sunilkpai:
I am working on the `bicgstab` implementation (using the current pytree convention in `cg`); I am most of the way there as the implementation was very similar to cg and the tests generally seem to work. This addresses one of the goals of #1531.

Before finalizing this PR, there are some points of confusion on the tests that have to do with numerical issues in iterative solvers. I would like some advice on this before proceeding.

1. If you use a iterative solver, you only get the correct answer up to some tolerance. This makes gradient checking harder and different from a lot of the other more predictable behaviors in `jax`. I noticed the biggest shape in the gradient tests `*_solve` is `(2, 2)`, and when I checked larger shapes, there was an error (depending on the type, either array not equal or grad check failure)! What is the general consensus on testing with respect to gradient checks, especially for these approximate solvers?
2. Mid-iteration convergence and breakdown checks may require some special implementation of loops. To truly match `scipy` for `cg` and `bicgstab` more generally, I think this will need to be implemented somehow. I have a TODO comment somewhere in this PR explaining how it would be implemented if `jit` were not a requirement.

I also want to make sure that the changes I made to `cg` to avoid code duplication are kosher.

PR #5299, opened 2020-12-31
Feature/bicgstab

sunilkpai:
This is an update on #3796 for an implementation of BICGSTAB for the new repo and is no longer a WIP. I used the same tests as GMRES and they seem to work fine.

### DIFF
--- a/jax/_src/scipy/sparse/linalg.py
+++ b/jax/_src/scipy/sparse/linalg.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 from functools import partial
 import operator
 
@@ -48,6 +49,11 @@ def _vdot_real_part(x, y):
 
 def _vdot_real_tree(x, y):
   return sum(tree_leaves(tree_multimap(_vdot_real_part, x, y)))
+
+
+def _vdot_tree(x, y):
+  return sum(tree_leaves(tree_multimap(partial(
+    jnp.vdot, precision=lax.Precision.HIGHEST), x, y)))
 
 
 def _norm(x):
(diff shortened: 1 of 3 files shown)


jax/_src/scipy/sparse/linalg.py
```diff
@@ -123,10 +129,107 @@ def body_fun(value):
...
+      lambda _: _sub(s, _mul(omega_, t)),
+      None
+    )
+    k_ = lax.cond((omega_ == 0) | (alpha_ == 0),
+                  lambda _: -11, lambda _: k + 1, None)
+    k_ = lax.cond((rho_ == 0), lambda _: -10, lambda _: k_, None)
+    return x_, r_, rhat, alpha_, omega_, rho_, p_, q_, k_
+
+  r0 = _sub(b, A(x0))
+  # need to use _vdot_tree to match dtype (hacky...)
```
ianwilliamson, line 179: You could tree map `ones_like` I think?

sunilkpai: Here I am just trying to get the number 1 to match the dtype (complex or float, 32 or 64 bit). How would that work?

ianwilliamson: Sorry... thought I was being helpful with my drive by comment... so you actually want a scalar `1` . Perhaps https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.result_type.html#jax.numpy.result_type would be helpful, but I'll defer to someone else who knows for sure!

tests/lax_scipy_sparse_test.py
```diff
@@ -193,6 +196,110 @@ def tree_unflatten(cls, aux_data, children):
...
+    A = rng(shape, dtype)
+    solution = rng(shape[1:], dtype)
+    M = self._fetch_preconditioner(preconditioner, A, rng=rng)
+    b = matmul_high_precision(A, solution)
+    tol = shape[0] * jnp.finfo(A.dtype).eps
+    x, info = jax.scipy.sparse.linalg.bicgstab(A, b, tol=tol, atol=tol, M=M)
+    using_x64 = solution.dtype.kind in {np.float64, np.complex128}
+    solution_tol = 1e-8 if using_x64 else 1e-4
+    self.assertAllClose(x, solution, atol=solution_tol, rtol=solution_tol)
+
```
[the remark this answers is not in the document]
sunilkpai: @shoyer By just adding this line, both GMRES and BICGSTAB seem to fail this test, moreso for the `complex64` and `float32` dtypes. But since `scipy` matches, I expect the same thing would happen if we had used `scipy` for the iterative solves.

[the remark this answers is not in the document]
shoyer: Try relaxing the error tolerance a little bit?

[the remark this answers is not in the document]
sunilkpai: Hmm, ok it's already 0.2 right now, and I saw this case for GMRES on my local tests:

```
Mismatch: 100%
Max absolute difference: 0.6610913
Max relative difference: 161.79356
 x: array(0.657005, dtype=float32)
 y: array(-0.004086, dtype=float32)
```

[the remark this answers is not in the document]
shoyer: Hmm! My guess is that there is a genuine bug in the derivative rule for GMRES. I would suggest commenting out this test for GMRES for now (opening an issue so we don't forget about it), and we can try to fix that in a follow-on PR.

sunilkpai:
@shoyer It's ready now. I needed to relax the gradient check a bit more to 0.3 for BICGSTAB, and I think we can address GMRES gradient issues in a future PR as you mentioned so I commented it out for now.

shoyer:
Unfortunately some of these gradient checks (both GMRES and BICGSTAB) are failing on Google's internal CI, which also runs things on GPU/TPU and Google's slightly different linear algebra libraries for CPU.

This is rather annoying for non-Googlers to debug, but I suspect it's entirely spurious because it mostly turns up for low precision. If you want to give this one more try, perhaps set `order=1` in `check_grads()` (under the theory that higher order derivatives have larger numerical error). Otherwise, I guess we can comment out those checks, too, for now.

sunilkpai:
@shoyer what do you think is the best course of action here?

shoyer:
One of the GitHub CI tests seems to be timing out. It used to run in about 30 minutes and is now taking over an hour. Can you check to see if the tests this PR introduces are much slower than expected? 

The timing out test sets `JAX_NUM_GENERATED_CASES=25`:
https://jax.readthedocs.io/en/latest/developer.html#running-the-tests

shoyer:
It might be worth squashing your changes and rebasing on master to see if that helps the slow tests.

sunilkpai:
Hi @shoyer hmm no dice on the squashing. I couldn't immediately find which tests took so long (it didn't seem to matter if I increase the number of cases), but the longest tests seemed to be whereever grads are checked for CG. Is there a way I can see which tests took so long during CI?

shoyer:
I thought you commented out the gradient tests? If you can't reproduce this locally I'm honestly baffled here. Let's see if other JAX maintainers have ideas...

shoyer:
If you could sync in the latest changes from the JAX repo, the unit tests will run in verbose mode. That should hopefully make it more obvious where the test suite is getting stuck.

sunilkpai:
@shoyer re: grad tests, I only kept the grad tests that didn't pose an issue in master! And will do!

sunilkpai:
@shoyer I'm wondering if the issue was because of the x64 enable I accidentally kept?

tests/lax_scipy_sparse_test.py
```diff
@@ -30,6 +30,7 @@
 
 from jax.config import config
 config.parse_flags_with_absl()
+# config.update("jax_enable_x64", True)
```
shoyer, line 33: Good catch! This was probably it. Can you remove this commented out line entirely?

sunilkpai: Done!

sunilkpai:
@shoyer I think this should be good to go, let me know if there's anything else you'd like me to change!

shoyer (approved):

shoyer:
Thanks for your perseverance here, @sunilkpai !

We are really excited to have this feature.

sunilkpai:
Awesome, thanks @shoyer, happy to contribute!

tetterl:
Great work! For visibility https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.sparse.linalg should be updated as well.

jakevdp:
Thanks! Done in https://github.com/google/jax/pull/6647

PR #5305, opened 2021-01-04
Add custom JVP for scipy.special.gammaln

NeilGirdhar:


NeilGirdhar:
This gives a 50x speedup on my machine for `grad(jac.scipy.special.gammaln)(4.0)`

jax/_src/scipy/special.py
```diff
@@ -29,6 +29,7 @@
 def gammaln(x):
   x, = _promote_args_inexact("gammaln", x)
   return lax.lgamma(x)
+ad.defjvp(lax.lgamma_p, lambda g, x: lax.mul(g, digamma(x)))
```
shoyer, line 32: It looks like this line is already written here?
https://github.com/google/jax/blob/afe612217b6d956ee02d924ec286ebdb6fabdbb7/jax/_src/lax/lax.py#L2315

NeilGirdhar: @shoyer Ah, I didn't see that!  Sorry for the noise.  I just ran it with %timeit and got different results, but I should have just tried it with `make_jaxpr` to begin with.

Issue #5206, opened 2020-12-16
The side-effect omnistaging example is broken


sbodenstein:
The example illustrating a possible issue with omnistaging and global state (https://github.com/google/jax/blob/master/design_notes/omnistaging.md#example-1) no longer leads to an escaped tracer:
```
from jax import jit
from jax import random

print('omnistaging enabled', jax.config.omnistaging_enabled)
print(jax.__version__)

key = random.PRNGKey(0)
def init():
  global key
  key, subkey = random.split(key)
  print(key)
  return random.normal(subkey, ())
print(init())  # -1.2515389
print(init())  # -0.58665067
init = jit(init)
print(init())  # 0.48648298
print(init())  # 0.48648298  !!
print(key) # Traced<ShapedArray(uint32[2])>with<DynamicJaxprTrace(level=0/1)>
print(random.normal(key, ()))

# omnistaging enabled True
# 0.2.7
# [4146024105  967050713]
# -1.2515285
# [2384771982 3928867769]
# -0.5866531
# [3382499631 3878610767]
# 0.48647928
# 0.48647928
# [3382499631 3878610767]
# Buffer(-0.4876217, dtype=float32)
```
Has this omnistaging behaviour been changed?

jakevdp:
Thanks for flagging this - I tried back as far as jax 0.2.0 when omnistaging was first turned on by default, and could not reproduce the expected escaped tracer error. I'm going to assign to @mattjj because he might know what's going on.

mattjj:
Hrm, curious. This repros for me:

```python
import jax
from jax import jit
from jax import random

key = random.PRNGKey(0)
def init():
  global key
  key, subkey = random.split(key)
  return random.normal(subkey, ())
init = jit(init)
print(init())  # -1.2515389
print(init())  # -1.2515389
print(key) # Traced<ShapedArray(uint32[2])>with<DynamicJaxprTrace(level=0/1)>
```

If we put a single `print(init())` before jitting, then we also get the escaped tracer. But if we put two, we don't! Very curious.

I think the reason why @jakevdp wasn't able to reproduce is that he didn't rewind jaxlib as well as jax. The funny actor here is the C++ jit. If we run the original version with `JAX_CPP_JIT=0` then we get the behavior as expected:

```
$ cat 5206.py
import jax
from jax import jit
from jax import random

key = random.PRNGKey(0)
def init():
  global key
  key, subkey = random.split(key)
  return random.normal(subkey, ())
print(init())
print(init())
[...]

PR #5314, opened 2021-01-05
make c++ jit dispatch sensitive to required global context

mattjj:
See #5206. The C++ jit dispatch path was not sensitive to the global state introduced by omnistaging, which led to the bug in #5206. In the edge cases where it could occur, the bug had the effect of disabling omnistaging in certain ways, which could lead to higher memory fragmentation. With the C++ jit dispatch path ignoring necessary global context, other bugs could have arisen in the future as it grew to handle more cases.

This was a collaboration with @jblespiau !


mattjj:
@googlebot I fixed it. @jblespiau works at DeepMind.

Issue #5195, opened 2020-12-15
Older jaxlib versions not available for Python 3.9


lumip:
It seems that only the latest `jaxlib` version (`0.1.57`) is available on the Python package index for Python 3.9. This breaks installs for projects depending on older versions (such as, e.g., [numpyro](https://github.com/pyro-ppl/numpyro/issues/838)). Would it be possible to make at least `jaxlib 0.1.56` available for Python 3.9?

hawkinsp:
`jaxlib` 0.1.57 is the first version for which we made a Python 3.9 wheel release. I think it's unlikely we're going to make retroactive releases of old jaxlib minor versions; instead, I think we should focus on getting numpyro onto a newer jaxlib.

I think jaxlib 0.1.56 should also build from source, but no promises... You should be able to check out the git tag and build it as normal.

Does that work for you?

lumip:
I've been in contact with the folks at numpyro and they don't want to upgrade their jax/jaxlib dependency while #5162 is not sorted out. But I guess I'll just have to wait then..

on another note: you mentioned "it's unlikely we're going to make retroactive releases of old jaxlib **minor** versions". Could you shed some light on what constitutes a minor version for jax/jaxlib (and is that the middle or rightmost number in your version designation, i.e., do you follow major.minor.micro as in PEP 440 or some other designation?), because I haven't seen a clear pattern.. so far it just seems like the micro version number is increased no matter if there are incompatible changes or not, which makes having jax/jaxlib as dependencies somewhat cumbersome to manage.

hawkinsp:
Currently we're not aiming to provide any particular compatibility guarantees across versions. I expect we will in the future, but we aren't right now. I'd suggest that pinning both `jax` and `jaxlib` is the right course of action if you want a stable API, but it does create problems if you want features that were added in newer `jax` releases (e.g., here, Python 3.9).

#5162 is due to an LLVM change, but we still haven't bisected it to exactly which change. Given the holidays for many of us it might not be fixed until January.

As I mentioned, the short-term fix would be to use an older Python version, or to build `jaxlib` 0.1.56 yourself with Python 3.9, which should work.

hawkinsp:
Jaxlib 0.1.58 fixes the issue that was blocking the NumPyro folks.

PR #5315, opened 2021-01-05
CI: use python 3.9 for numpy dispatch tests

jakevdp:
We currently have no Python 3.9 test coverage, and this is a good candidate because it is testing behavior of future numpy versions.


jakevdp:
Looks like we'll need to wait until the oldest supported jaxlib version is 0.1.57 (see #5195)

jakevdp:
Looks like we can try again after #5438

froystig (approved):

PR #5319, opened 2021-01-05
add source line info to more escaped tracer errors

mattjj:
The main goal here is to generate better error messages, particularly related to escaped tracers (e.g. due to side effects in user code). In addition to the regular traceback, it's helpful to show information about where the problematic tracer was created. That is, while the error is being raised on the line where a problematic tracer is _encountered_, we also want to explain how this problematic tracer came into being.

This extra source info is still only on jaxpr staging tracers, but those seem to be the most common culprits. I moved the `_line_info` attribute to the base Tracer class in core.py in anticipation of populating it for more traces than just DynamicJaxprTrace, but I'll leave that extension to follow-up.

I adapted the main escaped tracer error messages in core.py, and also slightly generalized and debugged source_info_util functions (thanks for explaining the path prefix bug, @froystig !).

Now an error message that used to look like this:

> The functions being transformed should not save traced values to global state. Detail: Incompatible sublevel: DynamicJaxprTrace(level=1/1), (2, 0). Offending tracer created on /foo/bar/baz.py:104 (__call__)..

Turns into something like this:

> jax.core.UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.
> The functions being transformed should not save traced values to global state. Detail: Incompatible sublevel: DynamicJaxprTrace(level=1/1), (2, 0).
> The tracer that caused this error was created on line /foo/bar/baz.py:104 (__call__).
> When the tracer was created, the final 10 stack frames (most recent last) were:
> /foo/bar/baz.py:111 (__call__)
> ... suppressed a few lines in this example ...
> /foo/bar/baz.py:104 (__call__)

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -384,19 +384,19 @@ class Trace:
       elif val._trace.sublevel < sublevel:
         return self.sublift(val)
       else:
-        raise escaped_tracer_error("Can't lift sublevels {} to {}"
-                                   .format(val._trace.sublevel, sublevel))
+        raise escaped_tracer_error(
+            val, f"Can't lift sublevels {val._trace.sublevel} to {sublevel}.")
     elif val._trace.level < level:
       if val._trace.sublevel > sublevel:
-        raise escaped_tracer_error("Incompatible sublevel: {}, {}"
-                                   .format(val._trace, (level, sublevel)))
+        raise escaped_tracer_error(
+            val, f"Incompatible sublevel: {val._trace}, {(level, sublevel)}.")
       return self.lift(val)
     elif val._trace.level > level:
-      raise escaped_tracer_error("Can't lift level {} to {}"
-                                 .format(val, self))
+      raise escaped_tracer_error(
+          val, f"Can't lift level {val} to {self}.")
     else:  # val._trace.level == self.level:
-      raise escaped_tracer_error("Different traces at same level: {}, {}"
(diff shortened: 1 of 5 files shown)


jax/core.py
```diff
@@ -384,19 +384,22 @@ def full_raise(self, val) -> 'Tracer':
...
-      raise escaped_tracer_error("Can't lift level {} to {}"
-                                 .format(val, self))
+      raise escaped_tracer_error(f"Can't lift level {val} to {self}."
+                                 + tracer_line_info_detail(val))
     else:  # val._trace.level == self.level:
-      raise escaped_tracer_error("Different traces at same level: {}, {}"
-                                 .format(val, self))
+      raise escaped_tracer_error(
+          f"Different traces at same level: {val}, {self}."
+          + tracer_line_info_detail(val))
```
froystig, line 402: Factor the common call to `tracer_line_info_detail` into `escaped_tracer_error`? The latter is a helper function already. Up to you.

froystig (approved):
Nice.

jax/core.py
mattjj: It's a good idea, though `escaped_tracer_error` is called in a couple other places. Those error messages are somewhat related, so I'll take a look and see if updating them makes sense.

Issue #5296, opened 2020-12-30
Add SVD to jax.random.multivariate_normal


SetarehAr:
I would like to request an update on the jax.random.multivariate_normal function to accommodate SVD in addition to Cholesky decomposition. This request is made based on two reasons: (1) To match the recent api on Numpy in https://numpy.org/devdocs/reference/random/generated/numpy.random.Generator.multivariate_normal.html (2) The numerical issues I've faced using Cholesky method for dimensions larger than 10, where the issues were fixed with using SVD.

minoring:
Hello @mattjj !
I took a look at history commits and believe `jax.random.multivariate_normal` used Cholesky decomposition only to simplify API.
I would like to get your opinion about this issue :)

minoring:
Related code:
https://github.com/google/jax/blob/6c8fc1b031275c85b02cb819c6caa5afa002fa1d/jax/_src/random.py#L693-L695

mattjj:
This seems like a good idea to me! Thanks for the super clear explanation @SetarehAr. I didn't know this could help with numerical issues.

@minoring are you interested in working on this? If not, I can take a stab at it.

minoring:
Yes, let me handle this issue ;)

minoring:
Hi @mattjj I believe this issue have been solved!

mattjj:
Thank you, @minoring !

PR #5323, opened 2021-01-06
Add method parameter to jax.random.multivariate_normal

minoring:
Add `method` parameter to `jax.random.multivariate_normal` to support 'svd' and 'eigh'.
Numpy's default method is 'svd' https://numpy.org/devdocs/reference/random/generated/numpy.random.Generator.multivariate_normal.html
but for JAX, I set default method to 'cholesky' for backward compatibility.

Related #5296

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -28,7 +28,7 @@ from jax.lib import xla_bridge
 from jax.lib import xla_client
 from jax.lib import cuda_prng
 from jax import core
-from jax.numpy.linalg import cholesky
+from jax.numpy.linalg import cholesky, svd, eigh
 from jax.interpreters import ad
 from jax.interpreters import batching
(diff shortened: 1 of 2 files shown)


jax/_src/random.py
```diff
@@ -657,22 +658,25 @@ def multivariate_normal(key: jnp.ndarray,
       broadcasting together the batch shapes of ``mean`` and ``cov``.
     dtype: optional, a float dtype for the returned values (default float64 if
       jax_enable_x64 is true, otherwise float32).
-
+    method: optinoal, a method to compute the factor of ``cov``.
```
mattjj, line 661:
```suggestion
    method: optional, a method to compute the factor of ``cov``.
```

mattjj (approved):
Very nice work! Thanks for doing this.

PR #5324, opened 2021-01-06
Skip tests that are failing without omnistaging (which will not be an…

chr1sj0nes:
… option shortly).

Enable tests on all platforms that were unnecessarily skipped.

### DIFF
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -159,6 +159,9 @@ class PmapTest(jtu.JaxTestCase):
 
   @ignore_slow_all_to_all_warning()
   def testTrees(self):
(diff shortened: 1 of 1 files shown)


mattjj (approved):
Thanks!

chr1sj0nes:
Thanks Matt. Is there anything blocking this from being merged?

mattjj:
No, sorry!

mattjj:
I'm not sure, but I think you can start merges too, by applying the 'pull ready' label and then sending the resulting CL to jax-merges@. If not, ping me on chat and we'll sort out the permissions so that you can.

PR #5330, opened 2021-01-06
clear some caches when setting up jaxpr typecheck tests

froystig:
In order to test that the typechecker identifies invalid jaxprs, some tests modify jaxprs in place. This is typically not allowed, since jaxprs are assumed immutable, and may be cached. As a workaround, this change clears the relevant caches before every test. This ought to prevent some order-dependent test failures.

### DIFF
--- a/tests/core_test.py
+++ b/tests/core_test.py
@@ -22,6 +22,7 @@ import numpy as np
 from absl.testing import absltest
 from absl.testing import parameterized
 
+import jax
 from jax import core
 from jax import lax
 from jax import numpy as jnp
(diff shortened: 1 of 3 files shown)


mattjj (approved):
Wow, good find!

tests/core_test.py
```diff
@@ -318,6 +319,12 @@ def f(x, y):
 
 class JaxprTypeChecks(jtu.JaxTestCase):
 
+  def setUp(self):
+    super().setUp()
+    jax._src.lax.control_flow._initial_style_open_jaxpr.cache_clear()
```
gnecula, line 324: Wouldn't it be safer to clear the caches on tearDown instead?

gnecula:
I do not understand why the cache was the problem. The jaxpr are cached based on the function pointer for the traced function. All the tests that change jaxpr trace lambda's which must always be different for each run, so there should be no cache hits for them.

froystig:
You might be thinking about the top-level jaxpr. What's being shared across tests here are the branch jaxprs of a cond ("subjaxprs"). In these tests, these jaxprs are associated with global functions (`jnp.sin` and `jnp.cos`), and will live on in caches across tests.

jakevdp (approved):
Nice catch!

PR #5333, opened 2021-01-06
Support empty inputs for jnp.linalg.pinv and jnp.linalg.svd

### LINKED ISSUE #5331: linalg.svd and linalg.pinv fail for zero-size matrices
Hi all,

This issue follows from #5313 with PR #5316

`jnp.linalg.svd(jnp.zeros((0,0)))` and `jnp.linalg.pinv(jnp.zeros((0,0)))` both fail with runtime errors.

I will try and take a look if I find time, but I'm not sure how soon that will be I'm afraid :(

jakevdp:
Fixes #5331

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -1128,6 +1128,10 @@ def svd_impl(operand, full_matrices, compute_uv):
 def svd_translation_rule(c, operand, full_matrices, compute_uv):
   shape = c.get_shape(operand).dimensions()
   m, n = shape[-2:]
+  if m == 0 or n == 0:
+    return xla.lower_fun(_empty_svd, multiple_results=True)(
+      c, operand, full_matrices=full_matrices, compute_uv=compute_uv)
+
   u, s, v = xops.SVD(operand)
   permutation = list(range(len(shape)))
   permutation[-1], permutation[-2] = permutation[-2], permutation[-1]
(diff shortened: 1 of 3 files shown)


hawkinsp:
Would a better fix be to make `lax.svd` do something sensible for `(0, 0)` arrays? Would this code then work unmodified?

jakevdp:
Maybe - I chose this approach because it's what `np.linalg.pinv` does: https://github.com/numpy/numpy/blob/31647f1b3e56c2f2a471cd2c3a583311534173f8/numpy/linalg/linalg.py#L1997-L2000

jakevdp:
But I'm looking at fixing the svd case; I'll see if this fix comes along for free.

hawkinsp:
Yes, this is fine, but I have a preference for avoiding special cases when they can fall out of a general case...

jakevdp:
Turns out this was easy to check, becuase svd of an empty array works on GPU. That leads to:
```python
>>> jnp.linalg.pinv(jnp.zeros((0, 0))
ValueError: zero-size array to reduction operation max which has no identity
```
So a special case will be necessary one way or another.

jakevdp:
Hmm, on second thought, specifying `initial=-jnp.inf` in the reduction is enough to fix `pinv` as long as SVD works. I think I'll go that route.

jakevdp:
OK, fixed the SVD as well.

jax/_src/lax/linalg.py
```diff
@@ -1188,10 +1188,30 @@ def svd_jvp_rule(primals, tangents, full_matrices, compute_uv):
...
+  v = jnp.empty(batch_shape + (0, 0), dtype=a.dtype)
+  if m < n:
+    u, v = v, u
+  return s, u, v
 
+def _svd_cpu_gpu_translation_rule(gesvd_impl, empty_ok, c, operand, full_matrices, compute_uv):
   shape = c.get_shape(operand)
   batch_dims = shape.dimensions()[:-2]
+
+  if not empty_ok and prod(shape.dimensions()[-2:]) == 0:
```
hawkinsp, line 1211: I'd actually be tempted to unconditionally use this code path for empty arrays, even on GPU. It would simplify the code a little.

hawkinsp (approved):

jax/_src/lax/linalg.py
jakevdp: Thanks - that makes sense. I'll make that change.

jakevdp:
Complex JVP test is failing, but this should be fixed by #5225

jakevdp:
Pushed a few minor tweaks to fix TPU errors

PR #5334, opened 2021-01-06
Add types to axes in lax_numpy.

hawkinsp:
I have mixed feelings about adding annotations in a couple of cases where the code needed changes to work around mypy.


jakevdp:
Looks good, although I agree that it's a bit unfortunate that satisfying mypy forces changes that arguably make the code less clear...

jakevdp (approved):

PR #5348, opened 2021-01-07
Use tiny for gumbel sampler

fehiepsi:
Previously, `gumbel` only gives samples in the range (-2.7689812183380127, 16). After this, `gumbel` will give samples in the range (-4.46976900100708, 16).


jakevdp:
Looks great, thanks!

jakevdp (approved):

jakevdp:
Interestingly enough, the fraction of the distribution that is inaccessible is exactly given by the minimum value we plug in here, so previously it was 10^-7, and now is 1-^-38. Nice improvement!

PR #3048, opened 2020-05-11
Update random.logistic() to prevent infinities

jakevdp:
Fixes #3043

PR #5349, opened 2021-01-08
Simplify jax.random.logistic algorithm

jakevdp:
Previous algorithm was overly complicated, and potentially biased. This is a much more straightforward approach to computing logistic samples, and changes the results by less than 1 part in 10^7 for float32. Test script:
```python
from jax import random
x = random.logistic(random.PRNGKey(1701), shape=(10000000,))
print(x[:5])
print(x.min(), x.max())
```
Before this PR:
```
[-0.4922756 -1.1400323  1.9611039 -1.1750274  1.8708186]
-15.942385 15.942385
```
After this PR:
```
[-0.4922756 -1.1400323  1.9611038 -1.1750274  1.8708185]
-15.942385 15.942385
```

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1299,21 +1299,9 @@ def logistic(key, shape=(), dtype=dtypes.float_):
 
 @partial(jit, static_argnums=(1, 2))
 def _logistic(key, shape, dtype):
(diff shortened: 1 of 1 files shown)


mattjj (approved):
LGTM! Cross-ref #3048.

Bonus points for saving so much vertical space ❤️

PR #5362, opened 2021-01-11
Make xmapped functions with no axis resources callable more than once

apaszke:
When no XLA lowering was involved (i.e. when using a purely vectorized
schedule), we used to return `f.call_wrapped` as the xmapped function.
This obviously didn't end well when someone tried to call it multiple
times, which this patch should fix.

### DIFF
--- a/jax/experimental/maps.py
+++ b/jax/experimental/maps.py
@@ -304,7 +304,9 @@ def make_xmap_callable(fun: lu.WrappedFun,
                                     EXPERIMENTAL_SPMD_LOWERING,
                                     *in_avals)
   else:
-    return f.call_wrapped
+    # We have to trace again, because `f` is a linear function, so we can't just return it.
+    final_jaxpr, _, final_consts = pe.trace_to_jaxpr_final(f, mapped_in_avals)
+    return core.jaxpr_as_fun(core.ClosedJaxpr(jaxpr, consts))
 
 class EvaluationPlan(NamedTuple):
   """Encapsulates preprocessing common to top-level xmap invocations and its translation rule."""
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -25,6 +25,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 from functools import partial
 
+import jaxlib
 import jax
 import jax.numpy as jnp
 from jax import test_util as jtu
@@ -65,6 +66,8 @@ def tearDownModule():
 
 @curry
 def with_mesh(named_shape, f):
+  if not named_shape:
+    return f
   def new_f(*args, **kwargs):
     axis_names, shape = unzip2(named_shape)
     size = np.prod(shape)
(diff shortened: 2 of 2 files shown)


apaszke:
cc @levskaya

mattjj (approved):
LGTM! (Not sure what's up with the CI failure; I saw it locally too, and assumed jaxlib needed an update...)

levskaya (commented):
Thanks for the fix, Adam!

jax/experimental/maps.py
```diff
@@ -304,7 +304,9 @@ def make_xmap_callable(fun: lu.WrappedFun,
                                     EXPERIMENTAL_SPMD_LOWERING,
                                     *in_avals)
   else:
-    return f.call_wrapped
+    # We have to trace again, because `f` is a linear function, so we can't just return it.
+    final_jaxpr, _, final_consts = pe.trace_to_jaxpr_final(f, mapped_in_avals)
+    return core.jaxpr_as_fun(core.ClosedJaxpr(jaxpr, consts))
```
levskaya, line 309: shouldn't these be 'final_jaxpr' and 'final_consts' from right above?

tests/xmap_test.py
```diff
@@ -302,6 +307,28 @@ def f(x, y):
...
+  def testMultipleCalls(self, mesh, axis_resources):
+    def f(x, y):
+      return x * y
+
+    @with_mesh(mesh)
+    def run_test():
+      f_mapped = xmap(f,
+                      in_axes=(['i', ...], ['j', ...]),
+                      out_axes=['i', 'j', ...],
+                      axis_resources=dict(axis_resources))
```
levskaya, line 326: If the test is using axis_resources / mesh then it's not actually going to exercise the problematic branch, is it?

jax/experimental/maps.py
```diff
@@ -304,7 +304,9 @@ def make_xmap_callable(fun: lu.WrappedFun,
                                     EXPERIMENTAL_SPMD_LOWERING,
                                     *in_avals)
   else:
-    return f.call_wrapped
+    # We have to trace again, because `f` is a linear function, so we can't just return it.
+    final_jaxpr, _, final_consts = pe.trace_to_jaxpr_final(f, mapped_in_avals)
```
levskaya, line 308: Correct me if I'm wrong - shouldn't this be using in_avals?  Some tests suggest this is the case.
```python
final_jaxpr, _, final_consts = pe.trace_to_jaxpr_final(f, in_avals)
return core.jaxpr_as_fun(core.ClosedJaxpr(final_jaxpr, final_consts))
```

levskaya (changes requested):

tests/xmap_test.py
levskaya: Adam - sorry I just misread the conditions the first time!

jax/experimental/maps.py
levskaya: @apaszke - sorry - I'm not sure the correct fix went in with this merge?

PR #5371, opened 2021-01-11
Add support for 'static_broadcasted_args' in 'sharded_jit'.

adarob:


jekbradbury:
The argument should just be called `static_argnums`, since `sharded_jit` is more like `jit` than `pmap`.

jekbradbury (approved):
LGTM other than the arg name.

adarob:
Done.

Issue #5374, opened 2021-01-12
jaxlib pip package 0.1.58 regression: ImportError: cannot import name 'cusolver' from 'jaxlib'


nfelt:
The most recent jaxlib CPU release (0.1.58), [new to PyPI as of earlier today](https://pypi.org/project/jaxlib/0.1.58/), leads to an ImportError at `import jax` (using jax 0.2.7) in a fresh venv on my machine: `ImportError: cannot import name 'cusolver' from 'jaxlib'`.  Reverting to jaxlib 0.1.57 eliminates the error.

Machine is macOS Cataline 10.15.7; virtualenv is python 3.8.6.

Repro script:

```
~$ virtualenv jaxtest
created virtual environment CPython3.8.6.final.0-64 in 142ms
  creator CPython3Posix(dest=~/jaxtest, clear=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=~/.local/share/virtualenv)
    added seed packages: pip==20.1.1, pkg_resources==0.0.0, setuptools==44.0.0, wheel==0.34.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
~$ source jaxtest/bin/activate
(jaxtest) ~$ pip install --upgrade pip
Collecting pip
  Downloading pip-20.3.3-py2.py3-none-any.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 15.2 MB/s
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.1.1
    Uninstalling pip-20.1.1:
      Successfully uninstalled pip-20.1.1
Successfully installed pip-20.3.3
(jaxtest) ~$ pip install --upgrade jax jaxlib
Collecting jax
  Using cached jax-0.2.7-py3-none-any.whl
Collecting numpy>=1.12
  Downloading numpy-1.19.5-cp38-cp38-manylinux2010_x86_64.whl (14.9 MB)
     |████████████████████████████████| 14.9 MB 13.2 MB/s
Collecting jaxlib
  Using cached jaxlib-0.1.58-cp38-none-manylinux2010_x86_64.whl (34.1 MB)
Collecting absl-py
  Using cached absl_py-0.11.0-py3-none-any.whl (127 kB)
Collecting flatbuffers
  Using cached flatbuffers-1.12-py2.py3-none-any.whl (15 kB)
[...]

PR #5376, opened 2021-01-12
[jax2tf] Remove legacy bfloat16 conversion code.

bchetioui:
Now that `tf.bfloat16` and `jnp.bfloat16` are the same, there should not be a need for specific conversion code anymore between the two.

### DIFF
--- a/jax/lib/__init__.py
+++ b/jax/lib/__init__.py
@@ -23,7 +23,7 @@ __all__ = [
 import jaxlib
 
 # Must be kept in sync with the jaxlib version in build/test-requirements.txt
-_minimum_jaxlib_version = (0, 1, 58)
+_minimum_jaxlib_version = (0, 1, 59)
 try:
   from jaxlib import version as jaxlib_version
 except Exception as err:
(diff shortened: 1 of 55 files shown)


bchetioui:
@googlebot I signed it!

jax/lib/__init__.py
```diff
@@ -23,7 +23,7 @@
 import jaxlib
 
 # Must be kept in sync with the jaxlib version in build/test-requirements.txt
-_minimum_jaxlib_version = (0, 1, 55)
+_minimum_jaxlib_version = (0, 1, 58)
```
gnecula, line 26: Do we really need to change this, or is it enough to use a new version of tf-nightly? In other words, does the jnp.bfloat16 from jaxlib 0.1.55 match the tf.bfloat16 from tf-nightly?

gnecula (commented):
Thanks

jax/lib/__init__.py
bchetioui: No, you're right. I thought I recalled we mentioned this was needed before, but it seems it is not. Thanks!

gnecula: There have been some problems with 0.1.58 (#5374). I am reluctant to force people to upgrade at the moment.

gnecula:
I am proposing that we hold off on merging this until jaxlib minimum version needs to be upgraded for other reasons as well.

bchetioui:
Oops, I meant to answer sooner! Sounds good to me, thanks :+1:

jax/lib/__init__.py
gnecula: I am going to be merging now, since minimum jaxlib was bumped to 1.59

Issue #2109, opened 2020-01-29
value_and_grad method for jacobian


aidanscannell:
Hi, I am fairly new to JAX (I've been loving it so far!) and there is some functionality that I would like to implement. I would like to use ```jacfwd``` in a similar way to ```value_and_grad``` i.e. return both the value and the jacobian. 

Currently I have implemented something like this,
```
jac_my_func = jacfwd(my_func, 0)
jacobian = jac_my_func(x, y)
value = my_func(x, y)
```
where the argument to differentiate wrt is the array x with shape (1, 2) and ```my_func``` returns an array (4,). ```my_func``` calculates a matrix (2, 2) and then vectorizes it to shape (4,). As such, the derivative/jacobian should be of shape (4,2).

Is there a function or a best practice to calculate the value and the jacobian at the same time?  Cheers.

matt-graham:
There are not currently dedicated `value_and_jac*` functions equivalent to the `value_and_grad` function however I submitted a pull-request (#762) to add `value_and_jacrev` and `value_and_jacfwd` functions which if merged should then in your example allow you to do

```
value, jacobian = value_and_jacfwd(my_func, 0)(x, y)
```

AdrienCorenflos:
:eyes:

jakevdp:
See https://github.com/google/jax/pull/762#issuecomment-1001885437 for some discussion of why this has not been added, and see https://github.com/google/jax/pull/762#issuecomment-1002267121 for a minimal implementation of this kind of functionality.

PR #762, opened 2019-05-23
Adding value_and_jac* functions

matt-graham:
At the moment while the Jacobian functions returned by both `jacfwd` and `jacrev` calculate the value of the function evaluated at the passed arguments the Jacobian is being computed for, there is no way to access this value. Often the value of the function will be needed in conjunction with the Jacobian - for example in a when applying Newton's method to find a root of a vector-valued function we require both the Jacobian and value of the function on each iteration.

This pull-request adds an optional `return_value` argument to both `jacfwd` and `jacrev` which if True makes the returned function return a tuple `(jacobian, value)` where `jacobian` is the Jacobian evaluated at the passed arguments and `value` is the value of the function evaluated at the passed arguments. If `return_value` is False (the default), `jacfwd` and `jacrev` retain their current behaviour - i.e. they return a function which returns only the Jacobian.

There already exists a `value_and_grad` function which implements an analogous operator for the specific case of scalar-valued functions. It would be more consistent with this existing function to define `value_and_jacrev` and `value_and_jacfwd` functions rather than using an additional argument, however this seemed like it could end up producing a lot of code and documentation duplication. The design proposed here is also more in keeping with the `has_aux` optional argument to `grad` and `value_and_grad` which alter the return signature of the returned function.

The same idea could also be applied to `hessian` function to optionally return both the Jacobian and value of the function being differentiated in addition to the Hessian itself. This would be useful for example when using Newton's method in an optimisation setting, where the objective function would be scalar valued and we would need both the Hessian and Jacobian (gradient) for the Newton iteration, and the value objective function to monitor convergence. The current pull-request implements part of the necessary functionality to do this as `hessian` internally calls `jacfwd` on the output of a `jacrev` call, with respectively the Jacobian and value being computed as intermediate values during these calls. It seems the best way to plumb everything together however would be to add a similar `has_aux` optional argument to `jacfwd` to allow it to be applied to a function returned by `jacrev` which returns both the Jacobian and value of a function. Adding `has_aux` to `jacfwd` and `jacrev` was mentioned as a todo in PR #499 however it looks like that PR was closed without merging and the related issue #497 remains open, so I'm not sure if there is some issue with adding `has_aux` for forward-mode?

PR #5377, opened 2021-01-12
Adding value_and_jacrev and value_and_jacfwd

tomcharnock:
In almost all uses (that I have come across), the value of the function is also wanted when calculating the Jacobian. To be able to obtain both of these without running the function and the Jacobian separately (which does two computations of the function) I have moved `jacrev` and `jacfwd` to `_jacrev` and `_jacfwd` which now take a parameter `return_value: bool = False`. I have then written thin wrappers,  `jacrev`, `value_and_jacrev`, `jacfwd` and `value_and_jacfwd` around `_jacrev` and `_jacfwd`. This means `jacrev` and `jacfwd` act exactly as before but we now can return an evaluatable function which returns the value and the Jacobian exactly like `value_and_grad`.


tomcharnock:
@googlebot I fixed it.

tomcharnock:
Hi, I was just wondering if there was anything in particular stopping this merge?

mattjj (commented):
Thanks for the contribution! We were slow to get to it only because we are really backlogged.

A downside to the current change is the main docstrings of interest are on `_jacfwd` and `_jacrev`, which because of the underscore prefix are not public API functions.

I suggest reorganizing the change so that:
1. `jacfwd` and `jacrev` keep their current docstrings, but their implementations change to be calls to new `value_and_jacfwd` and `value_and_jacrev` functions respectively (just discarding the value);
2. the new `value_and_jacfwd` and `value_and_jacrev` functions contain the meat of the implementation, with no `return_value` argument (because in effect it's always `True`).

See also #762 and #2109. The first thread in particular has some relevant discussion, but mainly I wanted to link them here for discoverability.

mattjj:
@tomcharnock let me know if you're keen to follow up on this; otherwise I can make some revisions and add commits to this PR!

tomcharnock:
@googlebot I fixed it.

tomcharnock:
> @tomcharnock let me know if you're keen to follow up on this; otherwise I can make some revisions and add commits to this PR!

Hi @mattjj,
I've just done an update this morning to make `value_and_jacrev` and `value_and_jacfwd` follow the stylings of `value_and_grad` and `jacrev` and `jacfwd` to follow grad. I agree that this clears up the code a bit and removes the non-public `_jacrev` and `_jacfwd`.

What do you think?

PR #5383, opened 2021-01-12
Add example code to save JAX program and run using C++ runtime.

### LINKED ISSUE #5337: Examples of Running a JAX function in C++
Are there any examples available of running `jit` functions defined in python from C++? I see that there is an interface for generating [something usable by XLA](https://jax.readthedocs.io/en/latest/jax.html#jax.xla_computation) but it is a bit unclear how to use the result of this when the function is dependent on variables/weights (e.g. a flax module).

@zhangqiaorjc I am told you have some knowledge of this?

Thanks

zhangqiaorjc:
This is example code for

https://github.com/google/jax/issues/5337

### DIFF
--- a/examples/jax_cpp/BUILD
+++ b/examples/jax_cpp/BUILD
@@ -1,0 +1,39 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


mattjj (approved):
Wow, beautiful! This is so concise and clean.

PR #5393, opened 2021-01-13
Fix SVD on ROCm

inailuig:
Fixes the order of the buffers and the size of the buffer for `E` when batched (which I got wrong in https://github.com/google/jax/pull/5115).
Also info is returned, so that workaround is not necessary for gesvd.

@hawkinsp

### DIFF
--- a/jaxlib/rocsolver.py
+++ b/jaxlib/rocsolver.py
@@ -292,22 +292,22 @@ def gesvd(c, a, full_matrices=True, compute_uv=True):
             _Shape.array_shape(dtype, batch_dims + (n, n), matrix_layout),
             # buffers[4] (vt; actually u)
             _Shape.array_shape(dtype, batch_dims + (m, m), matrix_layout),
-            _Shape.array_shape(singular_vals_dtype, (min(m, n) - 1,),
-                               (0,)),  # buffers[5] (e)
             _Shape.array_shape(np.dtype(np.int32), batch_dims,
-                               scalar_layout),  # buffers[6] (info)
+                               scalar_layout),  # buffers[5] (info)
+            _Shape.array_shape(singular_vals_dtype, batch_dims + (min(m, n) - 1,),
+                               vector_layout),  # buffers[6] (e)
             _Shape.array_shape(np.dtype(np.int8), (lwork,),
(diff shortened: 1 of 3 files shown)


jaxlib/rocsolver.py
```diff
@@ -292,22 +292,22 @@ def gesvd(c, a, full_matrices=True, compute_uv=True):
             _Shape.array_shape(dtype, batch_dims + (n, n), matrix_layout),
             # buffers[4] (vt; actually u)
             _Shape.array_shape(dtype, batch_dims + (m, m), matrix_layout),
-            _Shape.array_shape(singular_vals_dtype, (min(m, n) - 1,),
-                               (0,)),  # buffers[5] (e)
             _Shape.array_shape(np.dtype(np.int32), batch_dims,
-                               scalar_layout),  # buffers[6] (info)
+                               scalar_layout),  # buffers[5] (info)
+            _Shape.array_shape(singular_vals_dtype, batch_dims+(min(m, n) - 1,),
```
hawkinsp, line 297: Nit: add spaces around `+`

jaxlib/rocsolver.py
```diff
@@ -327,10 +327,10 @@ def gesvd(c, a, full_matrices=True, compute_uv=True):
                                matrix_layout),  # buffers[3] (u)
             _Shape.array_shape(dtype, batch_dims + (n, n),
                                matrix_layout),  # buffers[4] (vt)
-            _Shape.array_shape(singular_vals_dtype, (min(m, n) - 1,),
-                               (0,)),  # buffers[5] (e)
             _Shape.array_shape(np.dtype(np.int32), batch_dims,
-                               scalar_layout),  # buffers[6] (info)
+                               scalar_layout),  # buffers[5] (info)
+            _Shape.array_shape(singular_vals_dtype, batch_dims+(min(m, n) - 1,),
```
hawkinsp, line 332: Same here.

hawkinsp (changes requested):

inailuig:
I added the spaces, and while I was at it also rebased on master and fixed the rocm_gpu_kernel_helpers header guard

hawkinsp (approved):

PR #5394, opened 2021-01-13
Update CUDA installation instructions.

### LINKED ISSUE #5231: Jax doesn't see my GPU, even though Pytorch does
Jax sounds like an impressive project, thanks for working on it.

That said: on Ubuntu 18.04, this happens

```
➜  python
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> import torch, jax; print(torch.cuda.is_available()); print(jax.devices())
True
[CpuDevice(id=0)]
```

I first tried to pip install jax and got various errors; the error messages said it was common with old versions of pip that didn't support newer kinds of wheel and directed me to upgrade pip, which I did (from 9 to 20). Now Jax seems to be installed (at least various numpy-compatible functions do), but not to the point where it appears to see my GPU, a laptop "Geforce" card by Nvidia.

I'm not sure what systems diagnostics I can bring to help. This is the name of my card:

```
✗  lspci | grep NVIDIA
01:00.0 3D controller: NVIDIA Corporation GP108M [GeForce MX150] (rev a1)
```

As best as I understand it, these are drivers:

```
✗  lsmod | grep nvidia
nvidia_uvm            970752  0
nvidia_drm             53248  3
nvidia_modeset       1212416  2 nvidia_drm
nvidia              27643904  103 nvidia_uvm,nvidia_modeset
drm_kms_helper        184320  2 nvidia_drm,i915
drm                   491520  8 drm_kms_helper,nvidia_drm,i915
```

Thanks for reading this anyway.

hawkinsp:
Fixes https://github.com/google/jax/issues/5231

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,9 +411,12 @@ pip install --upgrade jax jaxlib  # CPU-only version
 On Linux, it is often necessary to first update `pip` to a version that supports
 `manylinux2010` wheels.
 
-If you want to install JAX with both CPU and GPU support, using existing CUDA
-and CUDNN7 installations on your machine (for example, preinstalled on your
-cloud VM), you can run
+If you want to install JAX with both CPU and NVidia GPU support, you must first
+install [CUDA](https://developer.nvidia.com/cuda-downloads) and
+[CuDNN](https://developer.nvidia.com/CUDNN),
+if they have not already been installed. Unlike some other popular deep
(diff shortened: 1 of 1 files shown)


skye (approved):
Beautiful

PR #5400, opened 2021-01-13
Fix pdot translation rule.

levskaya:
This concerns the direct pdot translation rule, which is not used
during spmd lowering.


mattjj (approved):
Nice catch, thank you!

PR #5405, opened 2021-01-14
Better error messages for test_util.check_grads()

shoyer:
Rather than merely reporting a failure in check_grads(), we now
report the *specific* check that failed, e.g., "JVP tangent" or
"VJP of JVP cotangent projection". Gradient tests often fail for
spurious reasons (e.g., due to insufficient precision), so this should
be helpful for debugging.

I tested this manually by relaxing the tolerance for a test in
`linalg_test.py`.

### DIFF
--- a/jax/test_util.py
+++ b/jax/test_util.py
@@ -124,13 +124,13 @@ default_gradient_tolerance = {
   np.dtype(np.complex128): 1e-5,
 }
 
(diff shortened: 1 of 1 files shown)


mattjj (approved):
LGTM! Would be extra useful to include example error messages in the PR comment, but that's optional.

jakevdp (approved):
This is great - thanks!

PR #5419, opened 2021-01-14
jnp.power: use integer_power path in all applicable cases

jakevdp:
Previously the integer power path was chosen only if `x2` was a Python integer; after this change, the integer power path is chosen for any concrete integer scalar.

As a consequence, some previously incorrect results are now loud failures. For example:

Before:
```python
>>> jnp.power(2, jnp.int32(-1))
DeviceArray(0, dtype=int32)
```
After:
```python
>>> jnp.power(2, jnp.int32(-1))
<...>
TypeError: Integers cannot be raised to negative powers, got integer_pow(ShapedArray(int32[], weak_type=True), -1)
```
Compare to Numpy:
```python
np.power(2, np.int32(-1))                                                                                                             
<ipython-input-5-b3a3e57b684d> in <module>
----> 1 np.power(2, np.int32(-1))

ValueError: Integers to negative integer powers are not allowed.
```

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -579,11 +579,15 @@ def _float_divmod(x1, x2):
 
 @_wraps(np.power)
 def power(x1, x2):
-  # Special case for small positive integer scalars: use binary exponentiation.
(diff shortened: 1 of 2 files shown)


froystig (approved):
Nice.

PR #5425, opened 2021-01-14
Remove type overloads for tree_map and tree_multimap.

hawkinsp:
These overloads lead pytype to infer incorrect types for code like this:

```
from typing import Any
import jax.tree_util

def bar(x: Any) -> str:
  return repr(x)
reveal_type(jax.tree_util.tree_map(bar, [2, 3, 4]))
```

which deduces `str` when the output is `List[str]`.

### DIFF
--- a/jax/tree_util.py
+++ b/jax/tree_util.py
@@ -39,7 +39,7 @@ for examples.
 import functools
 import collections
 import operator as op
-from typing import Any, Callable, Optional, Sequence, Tuple, Type, TypeVar, Union, overload
+from typing import Any, Callable, Optional, Sequence, Tuple, Type, TypeVar, overload
 
 from .lib import pytree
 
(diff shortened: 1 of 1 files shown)


hawkinsp:
@NeilGirdhar FYI. See the problematic example in the description.

NeilGirdhar:
@hawkinsp 

Yeah, okay, sorry about missing this case!  You're right to remove the overloads then since this problem seems to apply to all of them.

It's really unfortunate.  I guess if JAX really catches on (like I hope/think it will) then someone will eventually write the mypy plugin that understands pytrees.

Thanks for pinging me :smile:

hawkinsp:
Interestingly `mypy` seems to do something sensible with this example. `pytype` does not. We have users who use both.

I'm still not sure about the remaining overloaded annotations: we might need to double check them also.

hawkinsp:
It's not clear to me whether it's a bug or not. I don't know if these behaviors are specified anywhere. It really depends on details of the type rules. e.g., is `Any` a valid instantiation of a polymorphic type variable. It appears that pytype thinks it is.

NeilGirdhar:
https://github.com/python/mypy/issues/9913

Issue #4643, opened 2020-10-19
Add a "How to think in JAX" doc


jakevdp:
Over the past year, I think there has been a bit of a transition in the usage of JAX-flavored numpy from *JAX is a drop-in replacement for numpy* to *JAX is a tool to use beside numpy*. This has manifested in many different ways (the switch from `import jax.numpy as np` to `import jax.numpy as jnp`, the deprecation of silent conversions to array within aggregates, some aspects of omnistaging, etc.) but I think we've now gotten to the point where using JAX effectively requires thinking about which operations should be staged (via `jnp` routines) and which should not be (via `np` routines).

I think it's important that we provide some entry-point into developing the correct mental model for using JAX effectively. I think this probably should take the form of a new narrative doc, which would absorb some aspect of the existing Sharp Bits doc.

A quick brainstorm of key points we should cover:

- JAX is not a drop-in replacement for numpy, but is a numpy-like interface for *staged* computations.
- Use `jnp` for operations that you want to be staged/optimized, and `np` for "compile-time" operations
- Common example: `jnp.prod(shape)` vs. `np.prod(shape)`. Why should you use the latter?
- Key difference: JAX arrays are immutable, so `x[idx] = y` does not work. Instead `x = x.at[idx].set(y)` & related
- JAX and dynamic shapes: can be staged but not compiled.
- Other points?

lukepfister:
I think this would be great.

Do you have any preliminary pointers on when to use `jnp` vs `np`?

jakevdp:
Briefly: use `jnp` when you want the calculation to be compiled / to be performed on the accelerator. Use `np` when you want the calculation to happen on the CPU / at compile time.

Joshuaalbert:
Another point to touch on is the idea of fixed size iterative algorithms over recursive. It's the main challenge in coding complicated algorithms with JAX. It's what hinders many people I presume, and I know I have had to redesign an algorithm several times to make it work with JAX. But the results are amazing in terms of speed improvments when you do this.

Another point might be about reducing how many `cond` operations you use to let accelerators work as effectively as possible. I.e. somethings it's better to use a `where` and compute both branchs of a switch rather than use a `cond`. Knowing when to do that typically requires profiling, but there are some good rules of thumb.

NeilGirdhar:
It might also be interesting to discuss
- static vs nonstatic attributes in pytrees, jitted functions, and custom derivatives, and
- imagining the propagation through a single code path of both primals and cotangents.

[...]

PR #5426, opened 2021-01-14
[DOC] add Thinking In JAX doc

jakevdp:
Addresses #4643

Rendered page: https://jax--5426.org.readthedocs.build/en/5426/notebooks/thinking_in_jax.html


docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+-------------
+**Key Concepts:**
+
+- JAX provides a numpy-inspired interface for convenience.
+- Through duck-typing, JAX arrays can often be used as drop-in replacements of numpy arrays.
+- Unlike numpy arrays, JAX arrays are always immutable.
+
+NumPy provides a well-known, powerful API for working with numerical data.
+For convenience, JAX includes the :mod:`jax.numpy` module, which closely mirrors the
+numpy API and provides easy entry into JAX. Here are some operations you might do in ``numpy``:
```
8bitmp3, line 20: This doc is awesome. 

May I suggest a tiny little thing: `NumPy` <- `numpy`

```suggestion
- JAX provides a Numpy-inspired interface for convenience.
- Through duck-typing, JAX arrays can often be used as drop-in replacements of NumPy arrays.
- Unlike NumPy arrays, JAX arrays are always immutable.

NumPy provides a well-known, powerful API for working with numerical data.
For convenience, JAX includes the :mod:`jax.numpy` module, which closely mirrors the
NumPy API and provides easy entry into JAX. Here are some operations you might do in ``numpy``:
```

([Here](https://jax.readthedocs.io/en/latest/_modules/numpy.html?highlight=numpy#) it's also mostly consistent)

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+   @doctest
+   In [1]: type(np.arange(10))
+   Out[1]: numpy.ndarray
+   
+   @doctest
+   In [2]: type(jnp.arange(10))
+   Out[2]: jax.interpreters.xla._DeviceArray
+   
+
+Python's `duck-typing <https://en.wikipedia.org/wiki/Duck_typing>`_ allows JAX arrays and Numpy arrays
```
8bitmp3, line 61: Nit:

```suggestion
Python's `duck-typing <https://en.wikipedia.org/wiki/Duck_typing>`_ allows JAX arrays and NumPy arrays
```

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+to be used interchangeably in many places.
+
+However, there is one important difference between JAX and numpy arrays: JAX arrays are immutable,
```
8bitmp3, line 64: Nit:

```suggestion
However, there is one important difference between JAX and NumPy arrays: JAX arrays are immutable,
```

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+`indexed update syntax <https://jax.readthedocs.io/en/latest/jax.ops.html#syntactic-sugar-for-indexed-update-operators>`_
+that returns an updated copy:
+
+.. ipython::
+
+    @doctest
+    In [1]: x.at[0].set(10)
+    Out[1]: DeviceArray([10,  1,  2,  3,  4,  5,  6,  7,  8,  9], dtype=int32)
+
+Numpy, lax & XLA: JAX's layers of API
```
8bitmp3, line 106: Nit:

```suggestion
NumPy, lax & XLA: JAX's layers of API
```

docs/thinking_in_jax.rst
jakevdp: Absolutely - thanks!

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
+How to think in JAX
+===================
+
+
+JAX provides a simple and powerful API for writing accelerated numerical code,
+but working effectively in JAX sometimes requires extra consideration.
+This document is meant to help build a ground-up understanding of how JAX operates,
+so that you can use it more effectively.
+
+JAX vs. Numpy
```
8bitmp3, line 10: Nit:

```suggestion
JAX vs. NumPy
```

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+.. ipython::
+
+   In [1]: x = jnp.array([1, 2, 1])
+
+   In [2]: y = jnp.ones(10)
+
+   In [3]: jnp.convolve(x, y)
+   Out[3]: DeviceArray([1., 3., 4., 4., 4., 4., 4., 4., 4., 4., 3., 1.], dtype=float32)
+
+Under the hood, this numpy operation is translated to a much more general convolution implemented by
```
8bitmp3, line 171: Nit:

```suggestion
Under the hood, this NumPy operation is translated to a much more general convolution implemented by
```

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+
+Static vs Traced Operations
+---------------------------
+
+**Key Concepts:**
+
+- Just as values can be either static or traced, operations can be static or traced.
+
+- Static operations are evaluated at compile-time in Python; traced operations are compiled
+  & evaluated at run-time in XLA.
```
8bitmp3, line 414: Nit:

```suggestion
  and evaluated at run-time in XLA.
```

(`&` -> `and` like in `JIT mechanics: tracing and static variables`)

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+for more detail on JAX convolutions).
+
+To JIT or not to JIT
+--------------------
+**Key Concepts:**
+
+- By default JAX executes operations one at a time, in sequence.
+- Using a just-in-time (JIT) compilation decorator, sequences of operations can be optimized
+  together and run at once.
+- Not all JAX code can be JIT compiled, as it requires array shapes to be static & known at
```
8bitmp3, line 199: Nit:

```suggestion
- Not all JAX code can be JIT compiled, as it requires array shapes to be static and known at
```

(`&` -> `and` like in `JIT mechanics: tracing and static variables`)

8bitmp3:
@jakevdp Hope you didn't mind me taking a look. I've got some ML editing experience (a book, docs). Thanks for writing this **Thinking 🤔 in JAX** guide 👍

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,484 @@
...
+JAX provides a simple and powerful API for writing accelerated numerical code,
+but working effectively in JAX sometimes requires extra consideration.
+This document is meant to help build a ground-up understanding of how JAX operates,
+so that you can use it more effectively.
+
+JAX vs. NumPy
+-------------
+**Key Concepts:**
+
+- JAX provides a Numpy-inspired interface for convenience.
```
8bitmp3, line 14: Nit (maybe I missed it earlier)

```suggestion
- JAX provides a NumPy-inspired interface for convenience.
```

jakevdp:
@levskaya pointed out in chat that the ipython-sphinx approach here has a real downside: code blocks are intermingled with IPython prompts, which makes them less convenient for copy/pasteing from the doc.

I'm going to mark this WIP for now and explore other approaches.

8bitmp3:
@jakevdp I was wondering if there a particular reason why the team favors RST with iPython here? I understand that Sphinx works with Markdown and I noticed the following JAX guides were created as MD files: [Device Memory Profiling](https://github.com/google/jax/blob/master/docs/device_memory_profiling.md), [`custom_vjp` and `nondiff_argnums` update guide](https://github.com/google/jax/blob/master/docs/custom_vjp_update.md), and [Profiling JAX programs](https://github.com/google/jax/blob/master/docs/profiling.md).

To follow the lessons from the _How to think in JAX_ guide, I manually converted the RST file into [a Jupyter notebook](https://github.com/8bitmp3/thinking-in-jax-jakevdp-notebook/blob/main/How_to_think_in_JAX_(by_jakevdp)_notebook_version.ipynb) paragraph by paragraph. Then, I formatted RTS-flavored code text (`functions` etc), so there's no `::stuff::` (but you can add links to JAX APIs manually in Markdown, I guess), ran the cells in Colab, downloaded the IPYNB file, used [Jupytext](https://pypi.org/project/jupytext/) to convert it into Markdown, and left it there in a raw form (with a lot of spaces that can be cleaned up by a Markdown linter): https://github.com/8bitmp3/thinking-in-jax-jakevdp-notebook/blob/main/How_to_think_in_JAX_(by_jakevdp)_notebook_version.md - if you're interested. The only thing that stands out is that cell outputs aren't clearly marked - adding `Output:` before e.g. ` jax.interpreters.xla._DeviceArray` would help. (The IPYNB to Markdown process is manual but, for example, the NumPy.org docs team has integrated it into GH Actions, I think.)

Again, thanks for putting together _How to think in JAX_.

jakevdp:
We're discussing this elsewhere... the main reason I chose rst is because of the better support for doctests and cross referencing... we've had issues with our notebook-based documentation getting out-of-date. I actually wrote this content originally as a notebook, then translated to RST for that reason.

I'm not sure we're going to go the ipython extension route, though, because it makes it difficult to copy code blocks (as mentioned above). I'm looking into myst-nb, but the doctesting story there is not particularly clear.

dhirschfeld:
> code blocks are intermingled with IPython prompts, which makes them less convenient for copy/pasteing from the doc

...unless you paste them in `ipython` itself which ignores them. You could also use the [`sphinx-copybutton`](https://sphinx-copybutton.readthedocs.io/en/latest/#using-regexp-prompt-identifiers) extension which (IIUC) can strip the prompts.

docs/thinking_in_jax.rst
```diff
@@ -0,0 +1,479 @@
+How to think in JAX
```
gnecula, line 1: I really like this format instead of notebooks. It should make it easier to comment on PRs and see diffs.

jakevdp:
OK, I pushed an update where I use `myst-nb` rather than `RST` for the thinking in jax content. This required removing `nbsphinx` in favor of `myst-nb` for parsing notebooks; this is pretty nice because it does so without a pandoc requirement.

I think myst-nb is a pretty good system that ticks a lot of our boxes. WDYT?

docs/thinking_in_jax.rst
jakevdp: Take a look at the update, which uses myst-nb rather than RST. I think you'll like this as well :grin:

8bitmp3:
> This required removing nbsphinx in favor of myst-nb for parsing notebooks; this is pretty nice because it does so without a pandoc requirement.

👍👍 

@jakevdp does `myst-nb` play well with the popular JAX Quickstart and other notebooks?

docs/thinking_in_jax.md
```diff
@@ -0,0 +1,401 @@
...
+Python's [duck-typing](https://en.wikipedia.org/wiki/Duck_typing) allows JAX arrays and NumPy arrays
+to be used interchangeably in many places.
+
+However, there is one important difference between JAX and NumPy arrays: JAX arrays are immutable,
+meaning that once created their contents cannot be changed.
+
+Here is an example of mutating an array in Numpy:
+
+```{code-cell}
+x = np.arange(10)  # Numpy: mutable arrays
```
8bitmp3, line 85: Sorry, I missed these two during the previous check:

```suggestion
Here is an example of mutating an array in NumPy:

```{code-cell}
x = np.arange(10)  # NumPy: mutable arrays
```

8bitmp3 (commented):
Found the last couple of `NumPy`s

jakevdp:
> @jakevdp does myst-nb play well with the popular JAX Quickstart and other notebooks?

Yes, you can take a look at the rendered version here: https://jax--5426.org.readthedocs.build/en/5426/notebooks/quickstart.html

docs/thinking_in_jax.md
jakevdp: Done, thanks

jakevdp:
Unfortunately, it looks like execution errors in myst-nb don't lead sphinx to fail, but only to a warning.

This is unfortunate because it means that failures will silently result in unexpected errors being published to our docs via cell outputs.

This issue was reported last year in https://github.com/executablebooks/MyST-NB/issues/248; the suggestion there is to use the `-W` tag in sphinx to globally turn all warnings into errors; unfortunately that is not a good option for JAX, because documentation format issues are also reported as warnings, and we inherit many such warnings from NumPy and SciPy in our wrapped docstring.

I consider this a hard blocker: if we can't find a way to surface notebook execution errors as build failures rather than warnings, I don't think myst-nb is going to be a good solution for JAX.

jakevdp:
Given the failed execution issue, I think it's worth pursuing other options. I'm going to revisit the idea of using RST directly.

marcvanzee:
Interesting discussion! For our HOWTO system in Flax we are also exploring options on how to explain things with code that is tested, while still being easily usable with git. Avital suggested to use Sphinx's `doctest` (see https://github.com/google/flax/pull/925), which is an option you mention in this thread as well.

@jakevdp: Looking at the issue you linked in myst-nb, it seems that they are planning to fix it in the coming weeks. Supposing that they fix the issue, do you think myst-nb is prefable over doctest?

jakevdp:
I've decided to punt on the myst-nb question for now, in favor of just getting this content onto the website. I converted the notebook back to ipynb, reverted the old commits, and pushed the ipynb here.

jakevdp:
Ready for review: see the final rendered doc page here: https://jax--5426.org.readthedocs.build/en/5426/notebooks/thinking_in_jax.html

docs/thinking_in_jax.rst
jakevdp: Update... I decided to punt on myst-nb in this PR, and focus on the content instead. I'll revisit translating our docs to a text-based source in a followup.

mattjj:
Btw the 'rendered page' link in the OP didn't work for me, but [this does](https://github.com/google/jax/blob/8e29ebb5b1056f4dbb086f184b6e62e7067e80ef/docs/notebooks/thinking_in_jax.ipynb).

mattjj (approved):
Wow this is **phenomenal**. Fantastic work.

Only suggestion I could come up with: maybe rephrase "JAX's layers of API" to "JAX API layering".

jakevdp:
> Only suggestion I could come up with: maybe rephrase "JAX's layers of API" to "JAX API layering".

Done

jakevdp:
I fixed the rendered page link (it was from an earlier iteration and was missing the `notebooks` subdirectory)

PR #5427, opened 2021-01-14
Fix wrapped doc

jakevdp:
This improves the formatting of some docstrings for the HTML doc generation. Example:

Before: 
74kWmwrYZHQn7pG

After: 
8Gcwwzmmxevk3MZ


mattjj (approved):
So much better!

PR #5434, opened 2021-01-15
Make jax.random.categorical compatible with jax_numpy_rank_promotion='raise'

CreativeEntropy:
Without this, jax.random.categorical raises an exception when using a non-empty sample_shape and jax_numpy_rank_promotion is set to 'raise'.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1246,7 +1246,10 @@ def categorical(key, logits, axis=-1, shape=None):
     _check_shape("categorical", shape, batch_shape)
 
   sample_shape = shape[:len(shape)-len(batch_shape)]
-  return jnp.argmax(gumbel(key, sample_shape + logits.shape, logits.dtype) + logits, axis=axis)
+  return jnp.argmax(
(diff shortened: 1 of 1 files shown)


skye:
Can you add a unit test?

jax/_src/random.py
```diff
@@ -1246,7 +1246,10 @@ def categorical(key, logits, axis=-1, shape=None):
     _check_shape("categorical", shape, batch_shape)
 
   sample_shape = shape[:len(shape)-len(batch_shape)]
-  return jnp.argmax(gumbel(key, sample_shape + logits.shape, logits.dtype) + logits, axis=axis)
+  return jnp.argmax(
+      gumbel(key, sample_shape + logits.shape, logits.dtype) +
+      logits.reshape((1,) * len(sample_shape) + logits.shape),
```
hawkinsp, line 1251: Slight nit: I'd prefer `lax.expand_dims` here, since you are inserting size-1 dimensions. `reshape` is a big hammer and we'd prefer to avoid it where its power is not needed.

hawkinsp (changes requested):
+1 to a unit test, please.

PR #5435, opened 2021-01-15
Clean up xmap tests

apaszke:
Move `pdot` tests to a separate class. Automatically run all regular
xmap tests with both the `pmap`-style lowering and `sharded_jit`-style
lowering.


mattjj (approved):
LGTM!

PR #5443, opened 2021-01-16
Extend the pad batching rule for when the padding values are batched

### LINKED ISSUE #5326: incomplete implementation for batching rule for pad
For example:

```
def fun(x, pad_val):  # x: f32[3], pad_val: f32[]
  return lax.pad(x, pad_val, [(1, 1, 0)])

res = fun(np.ones((3,)), 7.)
# Pad the first row with 7 and the second with 8
res_batch = jax.vmap(fun)(np.ones((2, 3)), np.array([7, 8.]))
```

Raises NotImplementedError in the _pad_batch_rule:
```
  File "/Users/necula/Source/jax/jax/experimental/jax2tf/tests/batch_poly_test.py", line 119, in fun
    return lax.pad(x, pad_val, [(1, 1, 0)])
  File "/Users/necula/Source/jax/jax/_src/lax/lax.py", line 750, in pad
    return pad_p.bind(operand, padding_value, padding_config=tuple(padding_config))
  File "/Users/necula/Source/jax/jax/core.py", line 271, in bind
    out = top_trace.process_primitive(self, tracers, params)
  File "/Users/necula/Source/jax/jax/interpreters/batching.py", line 149, in process_primitive
    val_out, dim_out = batched_primitive(vals_in, dims_in, **params)
  File "/Users/necula/Source/jax/jax/_src/lax/lax.py", line 3470, in _pad_batch_rule
    raise NotImplementedError  # loop and stack
```

gnecula:
Fixes: #5326

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -1936,6 +1936,30 @@ def _memcpy(axis, num, src, dst, offset):
 masking.masking_rules[lax.concatenate_p] = _concat_masking_rule  # type: ignore
 
 
+def _pad_batch_rule(batched_args, batch_dims, *, padding_config):
+  operand, padding_value = batched_args
+  operand_bdim, padding_value_bdim = batch_dims
+  padding_config = list(padding_config)
+  if padding_value_bdim is None:
+    assert operand_bdim is not None
+    padding_config.insert(operand_bdim, (0, 0, 0))
+    return lax.pad(operand, padding_value, padding_config), operand_bdim
+  else:
+    # Different padding values for each batch element. Loop and stack with scan
+    def scan_body(_, i: int):
+      to_pad = (operand if operand_bdim is None
+                else lax.squeeze(lax.dynamic_index_in_dim(operand, i, axis=operand_bdim),
+                                 (operand_bdim,)))
+      return (None, lax.pad(to_pad,
+                            lax.dynamic_index_in_dim(padding_value, i).reshape(),
+                            padding_config))
+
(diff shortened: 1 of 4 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -1936,6 +1936,30 @@ def body(i, dst):
...
+def _pad_batch_rule(batched_args, batch_dims, *, padding_config):
+  operand, padding_value = batched_args
+  operand_bdim, padding_value_bdim = batch_dims
+  padding_config = list(padding_config)
+  if padding_value_bdim is None:
+    assert operand_bdim is not None
+    padding_config.insert(operand_bdim, (0, 0, 0))
+    return lax.pad(operand, padding_value, padding_config), operand_bdim
+  else:
+    # Different padding values for each batch element. Loop and stack with scan
```
gnecula, line 1948: Is there a better way of padding different rows with different values, without a loop?

hawkinsp: I think there is. In the absence of interior padding, you can broadcast the padding values to a suitable shape (i.e., with 1s in all the other dimensions of the LHS), replicate them as many times as the pad config says,  and concatenate them on the left and right.

To handle interior padding, I think you could pad an array of `True`s padded with `False` values to get a mask saying which elements are padding. You can then select either the appropriate padding value or data value (which you would also need to dilate using `pad` so it lines up, with any padding value) depending on the mask.

hawkinsp: https://github.com/google/jax/pull/5762 implements an idea close to the second paragraph here. I think it's much simpler so we can close this PR.

Issue #5226, opened 2020-12-18
Confusion about jvp input shapes


KeAWang:
I have a question about `jax.jvp`. From my understanding, given `f: R^n -> R^m`, `jax.jvp(f, x, v)` will compute the pushforward at `x` evaluated on tangent vector `v`, i.e. `df(x)[v]`. In this case we can think about this as a matrix multiplicatoin where `df(x)` as a `m x n` matrix and `v` to be the same shape as `x`.

However, it seems that `jax.jvp` allow `v` to be of a different shape than `x` which confuses me:
```python
import jax
def f(x):
  # Note that f: R^n -> R^(n-1)
  return x[:-1]

n = 10
key = jax.random.PRNGKey(0)
x = jax.random.normal(key, (n,))
v = jax.random.normal(key, (2*n,))
jax.jvp(f, (x,), (v,))
```
will run properly and return
```python
(DeviceArray([-0.372111  ,  0.26423106, -0.18252774, -0.7368198 ,
              -0.44030386, -0.15214427, -0.6713536 , -0.5908642 ,
               0.73168874], dtype=float32),
 DeviceArray([ 1.054516  , -0.9692889 , -0.59460217, -0.0318858 ,
               2.410932  , -1.8784496 , -0.7847697 , -0.3137085 ,
               0.33370885], dtype=float32))
```
despite the fact that `v` is not the same shape as `x`. Shouldn't this fail since `x` and `v` are not the same shape?

mattjj:
Great question!

It does indeed seem that we should add a check that `v` is the same shape as `x`. We have such a [check for `jax.linearize`](https://github.com/google/jax/blob/36f91261099b00194922bd93ed1286fe1c199724/jax/api.py#L1794-L1801) and [`jax.vjp`](https://github.com/google/jax/blob/36f91261099b00194922bd93ed1286fe1c199724/jax/api.py#L1815-L1821). We seem to have a [check for matching dtypes in `jax.jvp`](https://github.com/google/jax/blob/36f91261099b00194922bd93ed1286fe1c199724/jax/api.py#L1704-L1712); I wonder if there was some reason not to check the shapes as well...

LuiGiovanni:
Hi, I will take this issue. It's my first time working on an issue here in Jax

LuiGiovanni:
Hello @mattjj @KeAWang, I've been trying to replicate this issue and when I run `jax.vjp` I run into this error message:

[...]

PR #5446, opened 2021-01-17
Added check for shapes of arguments in jvp, resolves issue #5226

gsp-27:
* Added check for the array shape.
* For handling cases where the inputs are ```ints```, ```floats``` I am using ```try/except``` loop. Not sure if this is the best way to go about it

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -1721,7 +1721,7 @@ def _jvp(fun: lu.WrappedFun, primals, tangents):
                       f"{core.primal_dtype_to_tangent_dtype(_dtype(p))}, but got "
                       f"tangent dtype {_dtype(t)} instead.")
     if np.shape(p) != np.shape(t):
-      raise ValueError("jvp called with inconsistent primal and tangent shapes;"
+      raise ValueError("jvp called with different primal and tangent shapes;"
                        f"Got primal shape {np.shape(p)} and tangent shape as {np.shape(t)}")
 
   flat_fun, out_tree = flatten_fun_nokwargs(fun, tree_def)
--- a/jax/api.py
+++ b/jax/api.py
@@ -1720,6 +1720,12 @@ def _jvp(fun: lu.WrappedFun, primals, tangents):
                       f"Got primal dtype {_dtype(p)} and so expected tangent dtype "
                       f"{core.primal_dtype_to_tangent_dtype(_dtype(p))}, but got "
                       f"tangent dtype {_dtype(t)} instead.")
+    try:
+      if p.shape != t.shape:
+        raise ValueError("jvp called with inconsistent primal and tangent shapes;"
+                         f"Got primal shape {p.shape} and tangent shape as {t.shape}")
+    except AttributeError:
+      pass
   flat_fun, out_tree = flatten_fun_nokwargs(fun, tree_def)
   out_primals, out_tangents = ad.jvp(flat_fun).call_wrapped(ps_flat, ts_flat)
   return (tree_unflatten(out_tree(), out_primals),
--- a/jax/api.py
+++ b/jax/api.py
@@ -1720,12 +1720,10 @@ def _jvp(fun: lu.WrappedFun, primals, tangents):
                       f"Got primal dtype {_dtype(p)} and so expected tangent dtype "
                       f"{core.primal_dtype_to_tangent_dtype(_dtype(p))}, but got "
                       f"tangent dtype {_dtype(t)} instead.")
-    try:
-      if p.shape != t.shape:
-        raise ValueError("jvp called with inconsistent primal and tangent shapes;"
-                         f"Got primal shape {p.shape} and tangent shape as {t.shape}")
-    except AttributeError:
-      pass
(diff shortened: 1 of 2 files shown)


jax/api.py
```diff
@@ -1720,6 +1720,12 @@ def _jvp(fun: lu.WrappedFun, primals, tangents):
                       f"Got primal dtype {_dtype(p)} and so expected tangent dtype "
                       f"{core.primal_dtype_to_tangent_dtype(_dtype(p))}, but got "
                       f"tangent dtype {_dtype(t)} instead.")
+    try:
+      if p.shape != t.shape:
+        raise ValueError("jvp called with inconsistent primal and tangent shapes;"
+                         f"Got primal shape {p.shape} and tangent shape as {t.shape}")
+    except AttributeError:
```
gnecula, line 1727: Ignoring exceptions like this is suspicious and dangerous. I assume you are trying to work around the case when `p` or `t` are Python scalars? The most general way to get a shape is with `np.shape`, which works for scalars too.

tests/api_test.py
```diff
@@ -947,6 +947,12 @@ def test_jvp_mismatched_arguments(self):
       TypeError,
       "primal and tangent arguments to jax.jvp do not match.",
       lambda: api.jvp(lambda x: -x, (np.float16(2),), (np.float32(4),)))
+    # If primals and tangents are not of the same shape then raise error
+    self.assertRaisesRegex(
```
gnecula, line 951: We are trying to transition to the new style of checking for exceptions, using `with self.assertRaisesRegex(...)`.

tests/api_test.py
```diff
@@ -947,6 +947,12 @@ def test_jvp_mismatched_arguments(self):
...
+      ValueError,
+      "jvp called with inconsistent primal and tangent shapes",
+      lambda: api.jvp(lambda x: x+1, (np.random.randn(10,),), (np.random.randn(20,),))
+    )
```
gnecula, line 955: I would also add a test case when the primal and/or tangent are Python scalar.

gnecula (changes requested):
Thank you for your first contribution!

jax/api.py
gsp-27: Hi,
I was exactly trying to work around that. Thanks! for the recommendation.

tests/api_test.py
gsp-27: I have tried to address all of this changes in the latest commit.

Thanks,
Gaurav

tests/api_test.py
```diff
@@ -947,6 +947,12 @@ def test_jvp_mismatched_arguments(self):
       TypeError,
       "primal and tangent arguments to jax.jvp do not match.",
       lambda: api.jvp(lambda x: -x, (np.float16(2),), (np.float32(4),)))
+    # If primals and tangents are not of the same shape then raise error
+    fun = lambda x: x+1
+    with self.assertRaisesRegex(ValueError, "jvp called with inconsistent primal and tangent shapes"):
+      api.jvp(fun, (jnp.array([1.,2.,3.]),), (jnp.array([1.,2.,3.,4.]),))
```
gnecula, line 953: When want to test exceptions, each statement that you want to test must be in their own `self.assertRaisesRegex`, otherwise you are only testing the first one.

jax/api.py
```diff
@@ -1720,6 +1720,10 @@ def _jvp(fun: lu.WrappedFun, primals, tangents):
                       f"Got primal dtype {_dtype(p)} and so expected tangent dtype "
                       f"{core.primal_dtype_to_tangent_dtype(_dtype(p))}, but got "
                       f"tangent dtype {_dtype(t)} instead.")
+    if np.shape(p) != np.shape(t):
+      raise ValueError("jvp called with inconsistent primal and tangent shapes;"
```
gnecula, line 1724: One small suggestion: instead of using "inconsistent" we can use the more precise term: "different", because that is what you are actually checking for.

tests/api_test.py
gsp-27: I see, I will make the change. Thanks

jax/api.py
gsp-27: Hi,
I have tried to address all the requested changes

Thanks,
Gaurav

tests/api_test.py
gnecula: I do not actually see a test that uses a Python scalar for the primals and tangents. You can replace one of `np.float32(20)` with `20.`.

gsp-27: Sorry, I am not very clear on this. Do you mean make a test case like so
```jax.jvp(fun, (10.,), (20.,))```

Thanks,
Gaurav Pathak

gsp-27: I think I understand what you meant to say and have tried to add the test case.

Thanks,
Gaurav Pathak

gnecula (approved):

tests/api_test.py
gnecula: That is what I meant, thanks!

PR #5450, opened 2021-01-18
Support for the k-means algorithm.

henrykmichalewski:


gnecula:
Hi Henryk, are these implementations of scipy standard APIs? If so, then they deserve to be under `jax._src.scipy` with an export in `jax.scipy`. I will assign this to Jake to review, since he understands better that area of the code.

henrykmichalewski:
> Hi Henryk, are these implementations of scipy standard APIs? If so, then they deserve to be under `jax._src.scipy` with an export in `jax.scipy`. I will assign this to Jake to review, since he understands better that area of the code.

There are three main functions in this implementation:
- py_vq (very similar API to scipy, can be completely aligned; the difference consists in optional parameters, which have minimal impact on the implementation).
- kmeans_single_iter (close to https://github.com/scipy/scipy/blob/master/scipy/cluster/vq.py#L268 by not identical - I am commenting inside of the implementation how to express in jax the original behavior, though it seems to cause problems with jitting)
- kmeans - a generic high-level wrapper around kmeans_single_iter, can be aligned with scipy APIs.

henrykmichalewski:
> Will assign this to Jake to review, since he understands better that area of the code.

In another pull request (to flax_models, not yet visible on github) I am importing kmeans_single_iter function from this module.

lkhphuc:
It's been a year since the last comment so I just want to ping this again.
With the growing in popularity of Vector Quantize model (Dalle, VQGAN etc) native support for `scipy.cluster.vq` module would be great.

For reference, this is another notebook implementing vq and kmean in Jax https://colab.research.google.com/drive/1AwS4haUx6swF82w3nXr6QKhajdF8aSvA#scrollTo=N95UH5aULeye

jakevdp:
Hi @lkhphuc - this looks like it would be a useful addition to `jax.scipy`. Is it something you'd be willing to contribute via a pull request?

henrykmichalewski:
Hi, I will revisit this contribution next week and try to update it.

YouJiacheng:
👍ping this again, it has been a month~

jakevdp:
In the time since this was first opened, we've better-defined the scope of jax scipy & numpy wrappers (see https://github.com/google/jax/discussions/18383) and k-means is probably out-of-scope. Thanks for putting this together, and I'm sorry it didn't work to merge it into JAX!

PR #5453, opened 2021-01-18
Make it possible to vmap xmapped functions

apaszke:
Or perhaps more importantly make it possible to nest xmaps that don't
specify any `axis_resources`. The math is a little tricky, so I've added
a fairly strong test that enumerates a wide range of potential ways of
interleaving vmapped and xmapped axes in both inputs and the output.
Thanks to that, I've actually caught one very subtle bug in the dynamic
tracing rule for xmap (sorting by dimension names instead of positional
axes).

### DIFF
--- a/jax/experimental/maps.py
+++ b/jax/experimental/maps.py
@@ -31,9 +31,10 @@ from ..api_util import flatten_fun_nokwargs, flatten_axes
 from ..interpreters import partial_eval as pe
 from ..interpreters import pxla
 from ..interpreters import xla
+from ..interpreters import batching
 from ..lib import xla_bridge as xb
 from ..lib import xla_client as xc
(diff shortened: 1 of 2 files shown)


jax/experimental/maps.py
```diff
@@ -305,7 +306,7 @@ def make_xmap_callable(fun: lu.WrappedFun,
                                     *in_avals)
   else:
     # We have to trace again, because `f` is a linear function, so we can't just return it.
```
mattjj, line 308: Maybe this tweak would make the comment clearer? (This isn't part of this PR, but I just noticed it.)

```suggestion
    # We have to trace again, because `f` is an lu.WrappedFun, so we can't just return it.
```

mattjj (approved):
LGTM!

PR #5455, opened 2021-01-18
Use XLA AllGather op for GPU (attempt 2).

chr1sj0nes:
This is an expansion of the first, rolled-back attempt (https://github.com/google/jax/pull/5260), this time with auto-diff and batching rules that some users are relying on.

My benchmarks suggest a speed-up of ~2-2.5x for larger inputs.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -670,10 +670,6 @@ def _expand(dim, size, index, x):
   out = lax.full(shape, lax._const(x, 0))
   return lax.dynamic_update_index_in_dim(out, x, index, dim)
 
-def _allgather(x, dim, size, index, axis_name, axis_index_groups=None):
-  outs = tree_util.tree_map(partial(_expand, dim, size, index), x)
-  return psum(outs, axis_name, axis_index_groups=axis_index_groups)
-
 def all_gather(x, axis_name, *, axis_index_groups=None):
   """Gather values of x across all replicas.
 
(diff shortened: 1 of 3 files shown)


tests/pmap_test.py
```diff
@@ -610,6 +610,23 @@ def testGatherReplicaGroupsInterleaved(self):
 
     self.assertAllClose(ans, expected, check_dtypes=False)
 
+  def testGradOfGather(self):
```
mattjj, line 613: Would it be possible to use the `jtu.check_grads` function, perhaps patterning off of `testGradOfPsum`? That way we'll get slightly more general testing (e.g. we don't need the `jnp.sum` in the definition of `g`, and so we'll be testing the transpose rule with more than just all-equal cotangent vector entries.)

tests/pmap_test.py
```diff
@@ -1827,6 +1844,21 @@ def f(x):
     self.assertAllClose(pmap(pmap(f, axis_name='j'), axis_name='i')(x),
                         vmap(vmap(f, axis_name='j'), axis_name='i')(x))
 
+  @skipIf(not jax.config.omnistaging_enabled,
```
mattjj, line 1847: This test seems pretty limited, but I think the same is true for many other collectives' vmap rules. We don't have to act on it in this PR, but we need more systematic tests of collectives, like we have in lax_vmap_test.py.

mattjj (commented):
Nice, thanks for figuring out these details!

I have a suggestion for beefing up the autodiff test a bit.

jax/_src/lax/parallel.py
```diff
@@ -726,17 +722,96 @@ def all_gather(x, axis_name, *, axis_index_groups=None):
...
     axis_index_to_group_index = indices.argsort() % len(axis_index_groups[0])
     index = lax_numpy.array(axis_index_to_group_index)[index]
+  outs = tree_util.tree_map(partial(_expand, all_gather_dimension, axis_size, index), x)
+  return psum(outs, axis_name, axis_index_groups=axis_index_groups)
 
-  axis_size = psum(1, axis_name, axis_index_groups=axis_index_groups)
-
-  return _allgather(x, 0, axis_size, index, axis_name, axis_index_groups)
+def _all_gather_translation_rule(c, x, *, all_gather_dimension, axis_name, axis_index_groups, axis_size, axis_env, platform):
+  # TODO(cjfj): Enable this for TPU also?
```
majnemer, line 747: Any reason not to make this part of this change?

tests/pmap_test.py
chr1sj0nes: Done. Thanks for the tip.

tests/pmap_test.py
chr1sj0nes: Yeah, it's a pretty unsatisfactory test. As you say, I think it's probably best to address all of these vmap collective tests in a follow-up.

jax/_src/lax/parallel.py
chr1sj0nes: Only that I haven't run any benchmarks on TPU. If you're confident that it won't introduce a performance regression, I'm happy to enable it on TPU also. Do you happen to know if the TPU op supports any `all_gather_dimension`?

majnemer: TPU supports any all_gather_dimension, I believe you are working around an XLA:GPU bug. XLA:GPU should choose a layout which makes any all_gather_dimension possible.

majnemer: Alternatively, XLA:GPU could use https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/all_gather_decomposer.h#L25-L27

mattjj (approved):
LGTM!

jax/_src/lax/parallel.py
majnemer: I went ahead and did this with [this CL](https://github.com/tensorflow/tensorflow/commit/dd4835f906f925c0465eb1f304d451dba1eff585).

It should not be needed to check either the platform or the all_gather_dimension.

chr1sj0nes: Thanks David! I'll enable it on all platforms in a follow-up PR.

PR #5459, opened 2021-01-18
add systematic pdot tests, utility functions

mattjj:
Run lots of tests with e.g.

```
env JAX_NUM_GENERATED_CASES=1000 python tests/xmap_test.py PDotTests
```

Right now the new pdot tests are numbered rather than named, just because the names where we just put together the strings of all the test parameters were long and unwieldy (though I checked that there are no duplicates generated!). I may try to generate a nicer string name for tests...

#5491 is a copy of this PR to try to work around some weird copybara issues.

### DIFF
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -14,11 +14,14 @@
 
 # flake8: noqa
 
+from contextlib import contextmanager
 import functools
-import itertools
+import itertools as it
 import os
 import unittest
 from itertools import product, permutations
+from typing import (Tuple, List, NamedTuple, Dict, Generator, Sequence, Set,
+                    Any, Hashable, Iterable, Iterator, Union)
 from unittest import SkipTest, skip, skipIf
 
 import numpy as np
@@ -33,7 +36,8 @@ from jax import vmap
 from jax import lax
 from jax.experimental.maps import Mesh, mesh, xmap
 from jax.lib import xla_bridge
-from jax._src.util import curry, unzip2
+from jax._src.util import curry, unzip2, split_list, prod
+from jax._src.lax.lax import DotDimensionNumbers
 from jax.interpreters import pxla
 
 from jax.config import config
@@ -63,20 +67,20 @@ def tearDownModule():
     os.environ["XLA_FLAGS"] = prev_xla_flags
   xla_bridge.get_backend.cache_clear()
 
-@curry
-def with_mesh(named_shape, f):
-  if not named_shape:
-    return f
-  def new_f(*args, **kwargs):
-    axis_names, shape = unzip2(named_shape)
-    size = np.prod(shape)
-    local_devices = list(jax.local_devices())
-    if len(local_devices) < size:
-      raise SkipTest(f"Test requires {size} local devices")
-    mesh_devices = np.array(local_devices[:size]).reshape(shape)
-    with mesh(mesh_devices, axis_names):
-      return f(*args, **kwargs)
-  return new_f
+MeshSpec = List[Tuple[str, int]]
+
+@contextmanager
+def with_mesh(named_shape: MeshSpec) -> Generator[None, None, None]:
+  """Test utility for setting up meshes given mesh data from `schedules`."""
+  # This is similar to the `with_mesh` function above, but isn't a decorator.
+  axis_names, shape = unzip2(named_shape)
+  size = prod(shape)
+  local_devices = list(jax.local_devices())
+  if len(local_devices) < size:
+    raise SkipTest(f"Test requires {size} local devices")
+  mesh_devices = np.array(local_devices[:size]).reshape(shape)
+  with mesh(mesh_devices, axis_names):
+    yield
 
 
 class XMapTest(jtu.JaxTestCase):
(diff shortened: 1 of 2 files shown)


tests/xmap_test.py
```diff
@@ -335,6 +532,47 @@ def f(x, y):
 
     self.assertAllClose(z, jnp.einsum('nij,njk->nik', x, y))
 
+  @parameterized.named_parameters(jtu.cases_from_list(
```
gnecula, line 535: I am curious how many raw test cases are here? If there are many, then we pay the cost to generate them and then we throw most of them out. When testing jax2tf with this kind of combinatorial test case generation I have run several times into cases when the interesting use cases where systematically ignored.  An alternative strategy is to generate a sequence of families of test cases, each with shallow-nested for-loops.

However, I do not understand the code being tested, and how many interesting paths through it there are. Perhaps there are really lots of paths.

mattjj: That's a good point to raise! It seems the relevant figure of merit is test generation time. On my machine it takes less than 1.5 seconds to generate all the test cases here (computed by comparing pytest's reported time running tests to timing the pytest command itself from my shell).

Each test case here is a valid pdot call and a valid xmap scheduling for that call, and there are no duplicates. Moreover this seems like the place where we most want exhaustive tests: any of these calls could reasonably appear in the wild, and we want xmap + pdot to be rock solid since we expect them to get a lot of use.

Maybe we can revisit if the test generation time becomes unreasonable, like tens of seconds?

apaszke: It's a common theme in many of our tests. There's the same problem in other pmap and xmap tests. I've been briefly entertaining an idea of switching the parametrization from generating a full list only to downsample it, to taking a method that generates a single case, parametrized by some PRNG/oracle that determines which choices to make. So instead of saying
```py
jtu.cases_form_list(
  {"x": x, "y": y, "z": z}
  for x in range(10000000)
  for y in range(10000000)
  for z in range(10000000))
```
we would instead say
```py
jtu.cases_from_sampler(
  lambda sampler: {"x": sampler.sample(10000000),
                   "y": sampler.sample(10000000),
                   "z": sampler.sample(10000000)})
```
Technically the latter could still be used to generate an exhaustive list of tests, but we also would have the option to only evaluate its body a few times to generate a small subset of the space of potential test space.

apaszke: 1.5s is not a lot if we have a single huge parametrized test, but I expect us to have more of those in the future. And waiting for 20s just to generate a subset of test cases seems unreasonable.

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
...
+def schedules_from_pdot_spec(
+    spec: PdotSpec, lhs_shape: Tuple[int], rhs_shape: Tuple[int]
+    ) -> Generator[Tuple[AxisResources, MeshSpec], None, None]:
+  names = gen_axis_names(spec.num_mapped)
+  lhs_spec = dict(zip(spec.lhs_mapped, names))
+  rhs_spec = dict(zip(spec.rhs_mapped, names))
+  logical_sizes = {
+      name: shape[ax]
+      for shape, spec in [(lhs_shape, lhs_spec), (rhs_shape, rhs_spec)]
+      for ax, name in spec.items()}
```
apaszke, line 467: Ugh this part was really confusing to me, because I kept wondering what does `PdotSpec.items()` do. I've only realized that you've shadowed the name a while later 🤦  Can you please change that?

tests/xmap_test.py
```diff
@@ -335,6 +532,47 @@ def f(x, y):
 
     self.assertAllClose(z, jnp.einsum('nij,njk->nik', x, y))
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": f"_{next(test_counter)}",
+       "lhs_shape": lhs_shape, "rhs_shape": rhs_shape, "spec": spec,
+       "axis_resources": axis_resources, "mesh_data": mesh_data}
+      for test_counter in [it.count()]
```
apaszke, line 539: lol, another ninja-level feature. Let-bindings in comprehensions!

tests/xmap_test.py
```diff
@@ -335,6 +532,47 @@ def f(x, y):
...
+      for spec in all_pdot_specs(lhs_shape, rhs_shape)
+      for axis_resources, mesh_data in schedules_from_pdot_spec(
+          spec, lhs_shape, rhs_shape)))
+  @ignore_xmap_warning()
+  def testPdotSystematic(self, lhs_shape, rhs_shape, spec, axis_resources,
+                         mesh_data):
+    names = gen_axis_names(spec.num_mapped)
+    contract_names, batch_names = split_list(names, [spec.num_mapped_contract])
+    lhs_spec = dict(zip(spec.lhs_mapped, names))
+    rhs_spec = dict(zip(spec.rhs_mapped, names))
```
apaszke, line 552: This part is a little confusing, because it depends on some unsaid invariants of `PdotSpec` (in this case that `lhs_mapped` returns `(*map_contract, *map_batch)`. It would be nicer if we could put this spec generation in the `PdotSpec` itself, so that one doesn't have to remember about those properties to understand this bit of code.

apaszke (approved):
This is awesome! I think there are some simplifications/unifications that might make this a bit easier to understand, but I'm so glad we have those exhaustive tests now!

tests/xmap_test.py
gnecula: Besides the testing slowdown, I had run into another problem with such tests. If this pattern generates 1000+ test cases from which we sample 10 by default or 25 in CI (and we pick always the same ones), it is quite possible that some important use cases are never tested until somebody runs the test manually with a very large NUM_GENERATED_CASES. That is in part why I asked how many tests are generated here. It looks likely to be in the thousands, and if it takes more than 1sec to generate them then it is possibly many thousands.

apaszke: One of my parametrizations generates ~250k cases and finishes in ~1.5s, so I'd expect the number to be O(100k) here too

mattjj: > it is quite possible that some important use cases are never tested until somebody runs the test manually with a very large NUM_GENERATED_CASES

That seems like a pretty good process to me. It's worked well for us in the past, e.g. for the lazy sublanguage and for our early development of the more complex lax primitives (like conv_general_dilated, dot_general, etc). During development (or revision) we manually run lots of them, then we rely on a subset for CI for regressions. (Regressions seem less likely in these cases.)

We took an alternative route for einsum testing, where we just added cases that we thought of. It didn't work well: the result was we missed lots of edge cases that had to be cleaned up as users ran into them! The space of einsum configurations was just too tricky to cover thoroughly with a priori thinking. I concluded systematic tests can be really useful.

Luckily we don't have to choose between systematic coverage and tests for specific edge cases: we can do both! Perhaps a useful pattern (maybe even one that can be automated?) is to add explicit cases whenever we see a failure in the manually-run exhaustive tests. That way, we could have the exhaustive tests to run manually when we want to (during development or revision to the code in question), but perhaps by default their generation is skipped for speed (i.e. CI could skip generating them entirely, or we could figure out a more efficient way to generate a subsample). At the same time we can also have good explicit edge case coverage, both from whatever hand-crafting we come up with and from saving past failures of our systematic search.

WDYT?

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
     jax.experimental.maps.EXPERIMENTAL_SPMD_LOWERING = self.old_lowering_flag
 
 
+LHSSpec = RHSSpec = Tuple[int, ...]
+DimSpec = Tuple[LHSSpec, RHSSpec]
+
+class PdotSpec(NamedTuple):
+  """Specifies positional axes to contract or batch via mapping or position."""
+  map_contract: DimSpec
+  pos_contract: DimSpec
```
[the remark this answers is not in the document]
mattjj: Yes, it's about mapped axes and uses positions. All the info in PdotSpec uses positional axes, in this case to identify which positional axes xmap should map. That is, the fields `map_contract` and `map_batch` are accessed (indirectly, via `lhs_mapped` and rhs_mapped` when setting up the `xmap` call itself, rather than the `pdot` call.

The properties `pos_contract_after_mapping` and `pos_batch_after_mapping` compute the positions _after_ mapping has eliminated some.

Does that make sense?

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
...
+  # not to assign any additional axes to be batched or contracted, while the
+  # recursive cases in the latter two yields come from choosing equal-sized axes
+  # to batch or contract (recursing to possibly choose more axes).
+  def helper(excluded1: Set[int], excluded2: Set[int]):
+    yield ((), ()), ((), ())
+    for i, d1 in enumerate(lhs_shape):
+      if i not in excluded1:
+        for j, d2 in enumerate(rhs_shape):
+          if j not in excluded2 and d1 == d2 and j <= i:
+            for contract, batch in helper(excluded1 | {i}, excluded2 | {j}):
```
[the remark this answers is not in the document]
mattjj: I think that makes the position computations more complicated. That is, when the recursive call returns axis positions, then the caller would need to increment them depending on whether it had eliminated an axis.

I think it's better to keep things 'flat' so as to avoid as much axis incrementing bookkeeping as possible.

WDYT? LMK if that's unclear and I can explain better.

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
...
+  # while excluding certain axis positions. The base case in the first yield is
+  # not to assign any additional axes to be batched or contracted, while the
+  # recursive cases in the latter two yields come from choosing equal-sized axes
+  # to batch or contract (recursing to possibly choose more axes).
+  def helper(excluded1: Set[int], excluded2: Set[int]):
+    yield ((), ()), ((), ())
+    for i, d1 in enumerate(lhs_shape):
+      if i not in excluded1:
+        for j, d2 in enumerate(rhs_shape):
+          if j not in excluded2 and d1 == d2 and j <= i:
```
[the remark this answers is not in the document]
mattjj: It avoids redundant test cases. That's necessary to, for example, use named parameterized tests rather than numbered ones.

tests/xmap_test.py
```diff
@@ -335,6 +532,47 @@ def f(x, y):
...
+
+    rng = jtu.rand_default(self.rng())
+    lhs = rng(lhs_shape, np.float32)
+    rhs = rng(rhs_shape, np.float32)
+
+    def pdot_fun(x, y):
+      print(f'pdot(x:{x.aval.str_short()}, y:{y.aval.str_short()},\n'
+            f'     axis_name={contract_names},\n'
+            f'     pos_contract={spec.pos_contract_after_mapping}\n'
+            f'     pos_batch={spec.pos_batch_after_mapping})')
```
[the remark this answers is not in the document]
mattjj: Done.

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
...
+    names = 'ijkl'
+    for n in it.count(1):
+      for chars in it.product(names, repeat=n):
+        yield ''.join(chars)
+  return list(it.islice(_axis_names(), n))
+
+AxisResources = List[Dict[str, str]]
+MeshSpec = List[Tuple[str, int]]
+
+def schedules(sizes: Dict[str, int]
```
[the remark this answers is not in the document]
mattjj: Oh great idea, I forgot about that! Will do.

tests/xmap_test.py
mattjj: Oops. That was an oversight. Thanks for the catch!

tests/xmap_test.py
mattjj: I can't help myself. I love Python golf.

tests/xmap_test.py
mattjj: Hmm, well right now `PdotSpec` doesn't include any name information. But we could factor the names into it, which might help simplify this calling code as well as other things. I'll give that a shot.

tests/xmap_test.py
mattjj: By the way, as per my earlier comment in this thread, I think the best option here is to keep figuring out a faster approach to (subsampled) systematic testing while also not blocking this particular PR on it, since this PR doesn't itself introduce any problematic test generation times. That is, we should solve the test generation timing problem before it becomes an issue with these xmap tests, but it doesn't need to be solved in this PR.

tests/xmap_test.py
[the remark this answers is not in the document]
mattjj: Ah, a bug! This code was generating duplicate cases. Will fix...

tests/xmap_test.py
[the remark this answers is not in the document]
mattjj: On the other hand, doing it this way this was a source of a bug that caused duplicate cases to be generated :)

tests/xmap_test.py
mattjj: It seems there are 15295 tests generated in my current file, though I've been changing things so it may not reflect the same number as an earlier version. I did a more accurate timing measure and found it required took less than 0.09 seconds to generate them, and 0.96 seconds to generate and subsample them. So we're okay on timing for now I think!

tests/xmap_test.py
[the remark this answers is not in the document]
mattjj: I revised the data structure which might things clearer, though I still stuck with specifying things in terms of positional axes.

tests/xmap_test.py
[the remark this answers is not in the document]
mattjj: It turned out it wasn't this approach that caused the duplication bug (I tried it recursing on sub-shapes and it was just as hard!). I had just messed up.

tests/xmap_test.py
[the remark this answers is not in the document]
mattjj: Haven't done this yet, but I will...

tests/xmap_test.py
```diff
@@ -272,11 +276,204 @@ def tearDown(self):
...
+    for mesh_sizes in it.product(*(divisors(sizes[n]) for n in names)):
+      axis_resources = dict((name, 'r' + name) for name in names)
+      mesh_data = [('r' + name, size) for name, size in zip(names, mesh_sizes)]
+      yield axis_resources, mesh_data
+
+@contextmanager
+def generate_mesh(named_shape: MeshSpec) -> Generator[None, None, None]:
+  """Test utility for setting up meshes given mesh data from `schedules`."""
+  # This is similar to the `with_mesh` function above, but isn't a decorator.
+  # Could de-duplicate.
```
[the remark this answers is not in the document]
mattjj: Haven't done this yet, but I will...

tests/xmap_test.py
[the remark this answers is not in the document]
apaszke: Yeah, I think that what's there now is great. Thanks!

tests/xmap_test.py
gnecula: I like the idea of keeping the option to run exhaustive tests, and also manually adding cases that we think are important to cover. Without the latter, we may not spot regressions until users report them, and then we manually run with many generated cases. I, for one, do not have the habit to routinely run with many generated cases, nor it is on my mind to ask contributors to do so when I review their code. Perhaps we can change this habit.

mattjj: (Now it's at 0.3 seconds to generate and subsample them, based on generating more possible schedules as per Adam's comment below.)

mattjj:
This was merged as part of #5491.

PR #5466, opened 2021-01-19
bug: always return DeviceArray from lax.integer_pow

jakevdp:
Previously this function could return a non-DeviceArray value in some cases; e.g.
```python
>>> type(lax.integer_pow(2.0, 1))
float
```
Always binding the primitive makes the result more consistent, and also makes the jaxpr easier to understand (e.g `lax.integer_pow(x, 0)` results in a jaxpr containing `integer_pow`, rather than a jaxpr containing `broadcast_in_dim`).

After doing this, it became clear that the `integer_pow` translation rule and jvp rule were incorrect for `y=0`, so I corrected these as well.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2433,7 +2433,8 @@ def _integer_pow_dtype_rule(x, *, y):
 def _integer_pow_translation_rule(c, x, *, y):
   if y == 0:
     shape = c.get_shape(x)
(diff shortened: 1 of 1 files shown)


mattjj:
Seems worth a try just to always bind the primitive. I bet XLA can simplify this away for us.

Either way, LGTM!

mattjj (approved):

mattjj (approved):
LGTM yet again!

PR #5470, opened 2021-01-20
[ROCm] fix build

inailuig:
This fixes the issues described in https://github.com/google/jax/pull/5115#issuecomment-747737291 https://github.com/google/jax/pull/5115#issuecomment-747737669 and https://github.com/google/jax/issues/2012#issuecomment-763411418.

however I am not sure it would go through your CI? @hawkinsp 

cc @deven-amd


hawkinsp:
There's one way to find out!

(Feel free to send PRs fixing things: I'll figure out how to make them go in. The main problem is I can't test the result, so it might take an iteration or two to get things right.)

PR #5471, opened 2021-01-20
Always use XLA AllGather op for JAX all-gather primitive.

chr1sj0nes:


hawkinsp:
If the CPU implementation was recently fixed, you might need to keep the old implementation around (gated on the jaxlib version).

PR #5478, opened 2021-01-20
util._wraps: correctly handle initial line indentation

jakevdp:
Some docstrings have this form:
```python
def foo():
    """Summary string
    
    Other description
    """
```
while others have this form:
```python
def foo():
    """
    Summary string
    
    Other description
    """
```
In both cases, the dedented output should be
```python
"""
Summary string

Other description
"""
```
The previous approach of dedenting within sections caused issues with sphinx directives embedded in source docstrings; i.e. cases like this:
```
def foo():
    """
    The docs

    .. equation::

       y = \int_0^N f(x) dx

    More text
    """
```

This PR modifies our docstring parsing to handle both cases correctly, which leads to better HTML rendering of docstrings.


mattjj (approved):
Nice attention to detail!

PR #5488, opened 2021-01-21
Add experimental x64 context manager

jakevdp:
This is an experiment in allowing the X64 flag to be temporarily set via a context manager. For example:
```python
import numpy as np
import jax.numpy as jnp
from jax import config, jit

from jax.experimental import enable_x64, disable_x64

# Works with and without experimental_cpp_jit
# config.update('experimental_cpp_jit', False)
config.update('jax_enable_x64', False)

x = np.random.randn(100, 100)

print(jnp.array(x).dtype)
# float32

u, s, vt = jnp.linalg.svd(x)
print(u.dtype)
# float32

j_svd = jit(jnp.linalg.svd)

u, s, vt = j_svd(x)
print(u.dtype)
# float32

with enable_x64():
    print(jnp.array(x).dtype)
    # float64

    u, s, vt = jnp.linalg.svd(x)
    print(u.dtype)
    # float64

    with disable_x64(): # Nested context managers!
        print(jnp.array(x).dtype)
        # float32

    u, s, vt = j_svd(x)
    print(u.dtype)
    # float64
```

~I think the CPP jit cache is the one big remaining issue, though there may be others. I'm curious to hear thoughts about this approach!~

Small change made it so this works even with CPP jit (thanks @mattjj!)

Update: added some simple CI tests.

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -22,6 +22,7 @@ from typing import Any, Callable
 import numpy as np
 
 import jax
+from jax.config import FLAGS
 
 partial = functools.partial
 
@@ -185,22 +186,34 @@ def split_merge(predicate, xs):
 
 def cache(max_size=4096):
   def wrap(f):
-    cached = functools.lru_cache(maxsize=max_size)(f)
+    @functools.lru_cache(max_size)
+    def cached(_, *args, **kwargs):
+      return f(*args, **kwargs)
 
     @functools.wraps(f)
     def wrapper(*args, **kwargs):
       if jax.core.debug_state.check_leaks:
         return f(*args, **kwargs)
       else:
-        return cached(*args, **kwargs)
+        return cached(bool(FLAGS.jax_enable_x64), *args, **kwargs)
 
     wrapper.cache_clear = cached.cache_clear
     wrapper.cache_info = cached.cache_info
     return wrapper
   return wrap
 
+def memoize(f):
+  @functools.lru_cache(None)
+  def memoized(_, *args, **kwargs):
+    return f(*args, **kwargs)
 
-memoize: Callable[[Callable], Any] = functools.lru_cache(maxsize=None)
+  @functools.wraps(f)
+  def wrapper(*args, **kwargs):
+    return memoized(bool(FLAGS.jax_enable_x64), *args, **kwargs)
+
+  wrapper.cache_clear = memoized.cache_clear
+  wrapper.cache_info = memoized.cache_info
+  return wrapper
 
 def prod(xs):
   out = 1
--- a/jax/linear_util.py
+++ b/jax/linear_util.py
@@ -72,6 +72,9 @@ from ._src.util import curry
 from .tree_util import tree_map
 
 from ._src import traceback_util
+
+from .config import FLAGS
+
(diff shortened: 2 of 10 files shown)


jax/linear_util.py
```diff
@@ -237,12 +237,14 @@ def cache(call: Callable):
   Returns:
      A memoized version of ``call``.
   """
+  from jax.config import FLAGS
```
mattjj, line 240: Can we move this import to the top of the file?

jakevdp: will do

mattjj:
Awesome!

I'm not sure yet but I think we might need to update the `util.cache` decorator too, like the way `util.memoize` was updated. Those are used, for example, on lax control flow, caching the jaxprs that get formed. We don't want to reuse jaxprs that were cached with x64=False if we now have x64=True.

Maybe we could:
1. instead of adding a new separate `memoize_x64`, change the existing `util.cache` and `util.memoize` (and `linear_util.cache`) always to be sensitive to the x64 flag status
2. have an eye towards making this generalizable, in case we need to make the caches sensitive to other context (I'm working on a leak-checker flag with @jekbradbury right now which would need this!)

WDYT?

jakevdp:
I initially made `memoize` itself sensitive to the flag status, but off the bat that led to the `WARNING:absl:No GPU/TPU found, falling back to CPU.` warning being printed twice.

I'll take a look at `util.cache` and add a control flow test.

mattjj:
> I initially made memoize itself sensitive to the flag status, but off the bat that led to the WARNING:absl:No GPU/TPU found, falling back to CPU. warning being printed twice.

Printing a warning twice doesn't seem too bad :P

I guess there are only two only callers of `util.memoize`, in dtypes.py and in xla_bridge.py. Still, it might make sense for the "default" function `util.memoize` to have the right context sensitivity, and for the `get_backend` function in xla_bridge.py to opt out of that sensitivity (just by calling `functools.lru_cache(maxsize=None)` directly; we don't really need an alias for that).

jax/_src/util.py
```diff
@@ -185,6 +187,18 @@ def cache(max_size=4096):
 
 memoize = functools.lru_cache(maxsize=None)
 
+def memoize_x64(f):
+  """Memoization which returns different results depending on the X64 flag."""
+  f_memoized_x32 = memoize(f)
+  f_memoized_x64 = memoize(f)
```
mattjj, line 193: Instead of having separate caches, another way to do this would be to add dummy args for the things we want to be sensitive to, maybe like

```
def memoize(f):
  @functools.lru_cache(maxsize=None)
  def f_memoized(_, *args, **kwargs):
    return f(*args, **kwargs)

  @functools.wraps(f)
  def wrapper(*args, **kwargs):
    return f_memoized(FLAGS.jax_enable_x64, *args, **kwargs)

  return wrapper   
```

This might work nicely for `cache`, where we want to control the max size. It also might scale better to more context, e.g. if we have multiple flags of context we want to include.

jakevdp: Done - please take a look!

mattjj (approved):
Wow awesome. Great tests.

PR #5490, opened 2021-01-21
add xeinsum, an einsum for xmap (& jnp.einsum easter egg)

mattjj:
We upgraded `jnp.einsum` to work with `xmap`'s named axes, lowering to `pdot`!

There's still a lot more to do, including:
* handling more than two operands
* handling more positional axis cases
* maybe de-duplicating some positional axis logic with the rest of `jnp.einsum`
* lots more tests
* checks once we have avals-with-names
* docs!

But already this is a convenient wrapper around `pdot`.

Because this is a strict generalization of the existing `einsum` API, and in particular a generalization of the axis specification string language, we don't need a new user-facing API function: we can just extend `jnp.einsum` to handle mapped axes! For now, this path is like an undocumented easter egg in `jnp.einsum`, but it may already be useful for those playing with `xmap`.

Co-authored w/ @apaszke.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -16,6 +16,7 @@ Parallelization primitives.
 """
 
 import collections
+import string
 import warnings
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


apaszke (approved):

jekbradbury (approved):
Anything we're giving up by not using opt_einsum in xeinsum?

mattjj:
Yes: support for more than two operands, for one!

This is just a first cut. We should look at using opt_einsum for supporting multiple operands, and more generally look at merging/de-duplicating with jnp.einsum itself.

In short, we plan to do that, we just want to get something in first.

PR #5492, opened 2021-01-22
revive the leak checker, as a debug mode

mattjj:
Revive the tracer leak-checker! It's best to think of this feature as part of a 'debug mode', as it disables some caches and adds some data structure traversals/copying on jit cache misses, which might affect performance (but perhaps in unimportant ways).

Coauthored with @jekbradbury.

The idea is to provide a flag and corresponding context manager for enabling leak-checking, raising an error as soon as a trace finishes where one of its corresponding tracers was stashed in global context by a user side-effect:

```python
lst = []

@jit
def f(x):
  lst.append(x)
  return x

with jax.core.checking_leaks():
  f(3.)  # raises exception!
```

One can alternatively enable the leak checker globally using the env var `JAX_CHECK_TRACER_LEAKS` and the corresponding flag.

This should help us (and eventually users) catch one of the main sources of confusing bugs with JAX, for new users and old ones alike. (It also should make it easier to squash the trickiest omnistaging issues.)

The implementation mechanism is one that @dougalm devised long ago, before we even had this version of JAX hooked up to XLA: use Python reference counting! More specifically:
1. when we kick off a transformation's tracing, we create exactly one MainTrace instance;
2. all Tracer instances (the things that box values, which propagate through user code) have a reference to that MainTrace instance (indirectly, by having references to Trace instances which in turn have references to the MainTrace instance);
3. users never get references to MainTrace or Trace instances, and instead only get references to Tracers, which they might stash with side-effect bugs;
4. so once we've completed a trace, if we delete our reference to the MainTrace instance, then it should have zero references, unless the user code leaked a Tracer;
5. we can check whether there are any remaining references using the `weakref` API, relying on Python's reference counting.

On that last point, the basic mechanism looks like this, after we've completed the trace:

```python
t = ref(main)
del main
if t() is not None:
  raise Exception("leaked tracer!")
```

The trouble with this approach, and the reason why we had to disable it when we hooked up XLA, is caching. Our caches tend to hold onto MainTraces or Tracers. In particular:
1. the `jit` and `pmap` caching mechanisms, namely `linear_util.cache` and the C++-jit-dispatch equivalent, hold references to MainTrace instances in their key (via the `linear_util.WrappedFun` `transforms` attribute);
2. the caching we do with the `util.cache` decorator for control flow can hold references to Tracers in their output (e.g. if we have closed-over traced constants) as well as (I think) Tracers in their input (if the Python callable closes over Tracers).

We solve those two issues in this PR by, respectively,:
1. when we're checking leaks, copy all MainTrace instances when forming the `linear_util.cache` key, since we only need equality not identity;
2. when we're checking leaks, disable all `util.cache` caching.

~There was one other hitch: the `test_leak_checker_avoids_false_positives_scan_vmap_2` test case shows something that kept giving false positives, even though if we swapped the `vmap` transform for a `jvp` transform things would work. I couldn't figure out what was persisting a reference to the MainTrace! Using `gc.get_referrers`, I found that as far as the garbage collector could tell, there _were_ no remaining references: there was a MainTrace referred to by a BatchTrace, which was referred to by a BatchTracer (corresponding to one of the inputs to the vmapped function), which was referred to by a single-element list containing that BatchTracer (the `in_tracers` list for the function), which was referenced by... nothing! I wonder if it's a CPython bug or something. In any case, I just added a [special check for this pattern, where `gc` thinks there are no live references to the tracer](https://github.com/google/jax/pull/5492/files#diff-6bc6fb90ca348a8206e5beaccb59ed850154ace9a3683019641e5e690008281bR765-R772), to rule out this false positive. I think this hack is okay because (a) this is just for a debug mode so performance doesn't matter, and (b) this hack might buy us a debugging superpower!~

^- Of course, @hawkinsp figured this issue out! What a baller. Here's what was up: it had to do with how `batch_jaxpr` was factored differently from `jvp_jaxpr`, and in particular how in `jvp_jaxpr` the main trace is created in a different lu.transformation than the one that creates/destroys the tracers ([1](https://github.com/google/jax/blob/bc3cd1286b506803b7d4eafd313be162565ee2d2/jax/interpreters/ad.py#L53-L54), [2](https://github.com/google/jax/blob/bc3cd1286b506803b7d4eafd313be162565ee2d2/jax/interpreters/ad.py#L68)) whereas with `batch_jaxpr` those two things were fused [3](https://github.com/google/jax/blob/bc3cd1286b506803b7d4eafd313be162565ee2d2/jax/interpreters/batching.py#L419-L422). In the batching case, that led to `lu.WrappedFun.call_wrapped` holding onto a tracer reference in its `ans`, and then `gen.send(ans)` triggering the leak check before that `ans` ref had been dropped (by assigning `ans = gen.send(ans)`, i.e. clobbering the lhs) ([4](https://github.com/google/jax/blob/bc3cd1286b506803b7d4eafd313be162565ee2d2/jax/linear_util.py#L173)). By separating the two, the tracer creation/elimination and the main trace creation, the jvp approach didn't have this issue, as there were two separate evaluations of the `ans = gen.send(ans)` line, one for each of the transformations: the first one served to drop all references to the tracers and the second triggered the check.

(I had _thought_ I had ruled out this explanation, by doing `ans = [ans]; ans = gen.send(ans.pop())`, but apparently that doesn't drop the reference to the arg in the caller frame.)

Ultimately the solution was just to refactor the batching.py logic to be more like the ad.py logic, which gave the opportunity to clean it up in other ways too. We think an alternative would be to follow the convention everywhere that what gets sent into the generator is some unpackable object, like a list you pop from or maybe a single-element iterator, so that the callee can destroy the remaining references, but we haven't explored that yet.

Some follow-up work:
* make the leaked tracer error message better, in particular including source provenance for the leaked tracer (maybe by following the `gc.get_referrers` trail up from the MainTrace instance we have in hand to find any Tracers that still live)

### DIFF
--- a/jax/config.py
+++ b/jax/config.py
@@ -170,3 +170,9 @@ flags.DEFINE_integer(
     int_env('JAX_TRACER_ERROR_NUM_TRACEBACK_FRAMES', 5),
     help='Set the number of stack frames in JAX tracer error messages.'
 )
+
+flags.DEFINE_bool(
+    'jax_check_tracer_leaks',
+    bool_env('JAX_CHECK_TRACER_LEAKS', False),
+    help='Turn on checking for leaked tracers as soon as a trace completes.'
+)
--- a/jax/config.py
+++ b/jax/config.py
@@ -174,5 +174,7 @@ flags.DEFINE_integer(
 flags.DEFINE_bool(
     'jax_check_tracer_leaks',
     bool_env('JAX_CHECK_TRACER_LEAKS', False),
-    help='Turn on checking for leaked tracers as soon as a trace completes.'
+    help=('Turn on checking for leaked tracers as soon as a trace completes. '
+          'Enabling leak checking may have performance impacts: some caching '
+          'is disabled, and other overheads may be added.'),
 )
--- a/jax/linear_util.py
+++ b/jax/linear_util.py
@@ -63,10 +63,13 @@ data must be immutable, because it will be stored in function memoization tables
 """
 
 import threading
+from functools import partial
 from typing import Any, Tuple, Callable
 import weakref
 
+from . import core
 from ._src.util import curry
+from .tree_util import tree_map
 
 from ._src import traceback_util
 traceback_util.register_exclusion(__file__)
@@ -154,7 +157,7 @@ class WrappedFun(object):
       gen = gen(*(gen_static_args + tuple(args)), **kwargs)
       args, kwargs = next(gen)
       stack.append((gen, out_store))
-    gen = None
+    gen = gen_static_args = out_store = None
 
     try:
       ans = self.f(*args, **dict(self.params, **kwargs))
@@ -167,7 +170,7 @@ class WrappedFun(object):
         stack.pop()[0].close()
       raise
 
-    del args
+    args = kwargs = None
     while stack:
       gen, out_store = stack.pop()
       ans = gen.send(ans)
@@ -242,7 +245,10 @@ def cache(call: Callable):
 
   def memoized_fun(fun: WrappedFun, *args):
     cache = fun_caches.setdefault(fun.f, {})
-    key = (fun.transforms, fun.params, args)
+    if core.debug_state.check_leaks:
+      key = (_copy_main_traces(fun.transforms), fun.params, args)
+    else:
+      key = (fun.transforms, fun.params, args)
     result = cache.get(key, None)
     if result is not None:
       ans, stores = result
@@ -265,6 +271,13 @@ def cache(call: Callable):
   memoized_fun.cache_clear = fun_caches.clear  # type: ignore
 
   return memoized_fun
+
+@partial(partial, tree_map)
+def _copy_main_traces(x):
+  if isinstance(x, core.MainTrace):
+    return core.MainTrace(x.level, x.trace_type, **x.payload)
+  else:
+    return x
 
 
 @transformation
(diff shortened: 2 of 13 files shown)


mattjj:
We pushed cb584a4, which intentionally breaks tests so that we can try to debug the unexpected leaking / false positive behavior.

jax/config.py
```diff
@@ -170,3 +170,9 @@ def __getattr__(self, name):
     int_env('JAX_TRACER_ERROR_NUM_TRACEBACK_FRAMES', 5),
     help='Set the number of stack frames in JAX tracer error messages.'
 )
+
+flags.DEFINE_bool(
+    'jax_check_tracer_leaks',
+    bool_env('JAX_CHECK_TRACER_LEAKS', False),
+    help='Turn on checking for leaked tracers as soon as a trace completes.'
```
jekbradbury, line 177: If this flag is also going to disable caching/shouldn't be left on outside actual debugging, it should probably say so here.

jax/linear_util.py
```diff
@@ -63,10 +63,13 @@ def trans1(static_arg, *dynamic_args, **kwargs):
 """
 
 import threading
+from functools import partial
 from typing import Any, Tuple, Callable
 import weakref
 
+from . import core
 from ._src.util import curry
+from .tree_util import tree_map
```
jekbradbury, line 72: Seems like you don't use this?

jekbradbury (approved):

jax/config.py
mattjj: Gooooood idea!

jax/linear_util.py
mattjj: Actually it is used: on a line that says `@partial(partial, tree_map)`, which I believe you and I wrote together like 6 months ago, probably while joking about how funny it was :)

PR #5502, opened 2021-01-25
fixed testNorm

jackd:
Current version of `testNorm` verifies that `jnp.linalg.norm` is consistent with itself rather than the `np.linalg.norm`.


mattjj (approved):
Thanks for catching this, and for the fix!

PR #5504, opened 2021-01-25
Switched lax.linalg.eigh return order

jackd:
`eigh_p` implementations return `(vectors, values)`. This is propagated into `lax.linalg.eigh`, which is inconsistent both with it's documentation and `jax.[numpy,scipy].linalg.eigh` return order.

This PR reverses the order of the returned arguments in `jax.linalg.eigh`, making it consistent with `numpy`/`scipy` variants and existing documentation, though not `eigh_p` implementations.

This is a breaking change. A non-breaking version of this PR could just fix the documentation and be non-breaking. It won't break code using `jax.[numpy,scipy]` variants.

The inconsistency still exists between `eigh_p` implementations and `lax.linalg.eigh`, but I'm afraid I don't know enough about `xla` internals (e.g. `eigh_translation_rule` and `xops.Eigh`) to fix that.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -35,6 +35,8 @@ PLEASE REMEMBER TO CHANGE THE '..master' WITH AN ACTUAL TAG in GITHUB LINK.
     of array arguments: {func}`jax.numpy.pad`, :func`jax.numpy.ravel`,
     {func}`jax.numpy.repeat`, {func}`jax.numpy.reshape`.
     In general, {mod}`jax.numpy` functions should be used with scalars or array arguments.
+  * Reversed the order of {func}`jax.lax.linalg.eigh` and underlying primitives to be
+    consistent with documentation and `numpy` and `scipy` wrappers.
 
 ## jaxlib 0.1.62 (Unreleased)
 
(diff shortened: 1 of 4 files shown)


hawkinsp:
In general making breaking changes where we don't need to isn't wise, but I don't think this one is a big deal and I agree it improves consistency. Can you add a sentence to the release notes (`docs/CHANGELOG.rst`) mentioning the breaking change?

Note if you want you can freely change the argument order to the `eigh_p` primitive in most of the rules, and perhaps that's a good idea: the translation rules are then the things that would need to swap the arguments. Perhaps that's better?

You'll also need to squash your commits.

jackd:
@hawkinsp I agree changing primitive order would be cleaner (arguably we could go further and change the output order of `lapack.pyx` functions and CUDA implementations, but that's a lot more effort than I'm willing ot go do). Can you confirm I can leave [xops.Eigh](https://github.com/google/jax/pull/5504/files#diff-3d21165a07286e88ea6aa0277a6ee519992bf0e6f20a50b17628cc0a59a1e69fR462) alone? I have no idea where that's defined and it doesn't seem to get triggered in any of the code for me to check...

hawkinsp:
Sorry this PR dropped off my radar for a while. Would you mind syncing on top of head?

jackd:
@hawkinsp how's that?

PR #5509, opened 2021-01-25
Improve batched collective rule for all_gather_p

danieldjohnson:
When an `all_gather` references a `vmap`ped axis, there is a particularly simple way of implementing it: simply "forget" that the axis was mapped, and return the full array. Conveniently, this doesn't require any explicit broadcasting, and makes it possible to use `out_axes=None` with the results.

Also adds an evaluation rule for `all_gather`, which lets you gather constants under a `vmap`. I think I'm missing something here, though, because it seems to not work under a `jit` of `vmap`. Happy to fix that if it's easy, or remove the evaluation rule if it's not the right approach.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -743,6 +743,13 @@ def _all_gather_via_psum(x, *, all_gather_dimension, axis_name, axis_index_group
   outs = tree_util.tree_map(partial(_expand, all_gather_dimension, axis_size, index), x)
   return psum(outs, axis_name, axis_index_groups=axis_index_groups)
 
+def _all_gather_impl(x, *, all_gather_dimension, axis_name, axis_index_groups, axis_size):
+  # Only called when the argument is not mapped.
+  out_shape = list(x.shape)
+  out_shape.insert(all_gather_dimension, axis_size)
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Thanks, Daniel!

mattjj:
Can you give a failing test case with jit-of-vmap? What's the error?

mattjj:
Ah, I think I see the issue. It's that `vmap` works by data dependence (like all transformations other than `jit`/`pmap`/`scan`/... staging). But that's a bit awkward with collectives applied to constants, as the `vmap` doesn't process the collective (since there's no data dependence) yet the `jit` sees it (because of omnistaging).

We basically need a custom bind operation that will lift a constant into the trace, based on the axis name provided (which is associated with the trace into which we want to lift). [That's what we do with psum.](https://github.com/google/jax/blob/84f55f87ecd4b1a69ed7eac48b144b57df9d3f11/jax/_src/lax/parallel.py#L485-L488). We might want to abstract the machinery a bit. I'm thinking we should make a `Primitive` subclass for collectives with its own `bind` method that performs this lifting.

I think this issue is independent of this PR, because it applies to e.g. `pmin`/`pmax`/`pdot` as well. I'll file an issue to follow up.

PR #5510, opened 2021-01-25
Wrap api_benchmark in absltest for internal convenience.

zhangqiaorjc:


zhangqiaorjc:
@jblespiau fyi, this is necessary for internal bm. See more internal changes as a followup.

zhangqiaorjc:
Maybe we don't need this. Going to discard it for now

PR #5511, opened 2021-01-26
Clarify tracing documentation.

lukaszlew:
Clarify tracing a bit and use wording that does not suggest that JAX executed python program.


mattjj (approved):
Thank you!

PR #5521, opened 2021-01-26
[JAX-GPU] Use XLA AllToAll op on GPU (for supported configurations).

chr1sj0nes:
With NVSwitch, this can give ~15x speed-up on 16 GPUs.


mattjj (approved):
Awesome!

PR #5523, opened 2021-01-26
Allow `jax.custom_gradient` to return vjp with singleton return value

sharadmv:
Currently, `jax.custom_gradient` requires the VJP function return a tuple of input cotangent values, even if there's only one.

Example:
```
@jax.custom_gradient
def f(x)
  return x ** 2., lambda g: (g * x,)
```

This PR enables returning a singleton non-tuple value if there's only one primal.

Example:
```
@jax.custom_gradient
def f(x)
  return x ** 2., lambda g: g * x
```

### DIFF
--- a/jax/custom_derivatives.py
+++ b/jax/custom_derivatives.py
@@ -22,7 +22,7 @@ from . import core
 from . import dtypes
 from . import linear_util as lu
 from .tree_util import (tree_flatten, tree_unflatten, tree_map, tree_multimap,
-                        register_pytree_node_class)
+                        treedef_is_leaf, register_pytree_node_class)
 from ._src.util import cache, safe_zip, safe_map, split_list
 from .api_util import flatten_fun_nokwargs, argnums_partial, wrap_hashably
 from .core import raise_to_shaped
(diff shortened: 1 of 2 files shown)


jax/custom_derivatives.py
```diff
@@ -818,7 +818,10 @@ def bwd(res, cts):
     cts_flat, out_tree_ = tree_flatten((cts,))
     if out_tree != out_tree_: raise TypeError(f'{out_tree}\n!=\n{out_tree_}')
     cts_out = core.eval_jaxpr(jaxpr, consts, *cts_flat)
-    return tree_unflatten(in_tree, cts_out)
+    cts_out = tree_unflatten(in_tree, cts_out)
+    if not isinstance(cts_out, tuple):
```
mattjj, line 822:
```suggestion
    if tree_util.treedef_is_leaf(cts_out)
```

I'm thinking we could check whether the user returned a singleton, rather than just checking that it's not a tuple. WDYT?

sharadmv: Good, idea, done.

mattjj (approved):
LGTM! Thanks.

PR #5525, opened 2021-01-27
add systematic pdot vjp tests

mattjj:
I again expected some of these to fail and didn't get any failures...

### DIFF
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -650,6 +650,54 @@ class PDotTests(jtu.JaxTestCase):
     tol = 1e-1 if jtu.device_under_test() == "tpu" else None
     self.assertAllClose(result, expected, check_dtypes=False,
                         atol=tol, rtol=tol)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": f"_{next(test_counter)}",
+       "lhs_shape": lhs_shape, "rhs_shape": rhs_shape, "pdot_spec": pdot_spec,
+       "axis_resources": axis_resources, "mesh_data": mesh_data}
+      for test_counter in [it.count()]
+      for lhs_shape, rhs_shape in product(
+          [(2,), (2, 4, 2, 1)],
+          repeat=2)
+      for pdot_spec in all_pdot_specs(lhs_shape, rhs_shape)
+      for axis_resources, mesh_data in schedules_from_pdot_spec(
+          pdot_spec, lhs_shape, rhs_shape)))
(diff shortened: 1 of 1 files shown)


tests/xmap_test.py
```diff
@@ -650,6 +650,54 @@ def pdot_fun(x, y):
...
+               axis_resources=axis_resources)
+
+    with with_mesh(mesh_data):
+      lhs_bar, rhs_bar = fun(lhs, rhs, out_bar)
+
+    tol = 1e-1 if jtu.device_under_test() == "tpu" else None
+    self.assertAllClose(lhs_bar, expected_lhs, check_dtypes=False,
+                        atol=tol, rtol=tol)
+    self.assertAllClose(rhs_bar, expected_rhs, check_dtypes=False,
+                        atol=tol, rtol=tol)
```
apaszke, line 699: Can't we use some AD testing utilities instead of rolling out the comparison ourselves?

apaszke (approved):
LGTM, but I wonder why can't we use some AD testing helpers

tests/xmap_test.py
```diff
@@ -650,6 +650,54 @@ def pdot_fun(x, y):
...
+    fun = xmap(pdot_fun,
+               in_axes=[pdot_spec.lhs_in_axes, pdot_spec.rhs_in_axes,
+                        [*pdot_spec.batch_names, ...]],
+               out_axes=(pdot_spec.lhs_in_axes, pdot_spec.rhs_in_axes),
+               axis_resources=axis_resources)
+
+    with with_mesh(mesh_data):
+      lhs_bar, rhs_bar = fun(lhs, rhs, out_bar)
+
+    tol = 1e-1 if jtu.device_under_test() == "tpu" else None
```
apaszke, line 695: Wow, this is _really_ inaccurate

tests/xmap_test.py
mattjj: I could be wrong, but in this test I think we want to do a check-xmap-of-vjp. With `check_grads` and similar, the check is 'fused' with the vjp, so we can do check-of-vjp-of-X, like check-vjp-of-xmap, but it's not clear to me how to stick an xmap in the middle there. (We should also check-vjp-of-xmap when we add autodiff support to xmap!)

LMK if there is a way to use existing utilities that I'm missing though, b/c I agree it would be better!

mattjj: (And I think xmap-of-check-vjp doesn't make sense because the check is a Python side-effect and xmap (currently) stages out execution.)

PR #5527, opened 2021-01-27
implement soft_pmap in terms of xmap

mattjj:
We can kill `soft_pmap`, now that `xmap` subsumes it!

~However, there are still two tests that don't work.~ There were two tests that were tricky. They failed because [on this line in `pxla.vtile`, the value of `tile_size: Optional[int]` may be `None`](https://github.com/google/jax/blob/1fd1faa06ca12fed2332ec52b943b4a76bd99da1/jax/interpreters/pxla.py#L1562), but that leads to populating an [`core.AxisEnvFrame`](https://github.com/google/jax/blob/1fd1faa06ca12fed2332ec52b943b4a76bd99da1/jax/core.py#L698) with a `None` in its `size` field (which seems maybe invalid?), and that ultimately leads to things like [`axis_index`](https://github.com/google/jax/blob/1fd1faa06ca12fed2332ec52b943b4a76bd99da1/jax/_src/lax/parallel.py#L983) and [`psum`-of-constant](https://github.com/google/jax/blob/1fd1faa06ca12fed2332ec52b943b4a76bd99da1/jax/_src/lax/parallel.py#L618) failing because they rely on having an integer value of `frame.size`.

I introduced this issue in #5492 because that added `axis_size` to the `batching.batch` function that `pxla.vtile` uses, yet at the call site in `pxla.vtile` the axis size could be `None` because it needs to be inferred from the arguments. That is, after #5492 we assumed the caller of `batching.batch` knows the axis size, but that's not true for `pxla.vtile`. (Our `xmap` tests had not caught the issue, for whatever reason!)

In 71e820a I revised `batching.batchfun` (and `batching._match_axes`) to again allow their `axis_size` arguments to be optional. But to simplify them I'm curious if we can instead adapt `pxla.vtile` to have the axis size on hand.

cc @apaszke

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -484,15 +484,6 @@ class XeinsumSpecParser:
 
 ### parallel primitives
 
-def _allreduce_soft_pmap_rule(prim, reducer, vals, mapped, chunk_size,
-                              *, axis_name, axis_index_groups):
-  if axis_index_groups is not None:
-    raise NotImplementedError("soft_pmap does not yet support axis_index_groups")
-  reduced_vals = [reducer(x, [0]) if m else x for x, m in zip(vals, mapped)]
-  outs = prim.bind(*reduced_vals, axis_name=axis_name,
-                   axis_index_groups=axis_index_groups)
-  return outs, (False,) * len(vals)
-
 # This is only used for collectives that do not include the vmapped axis name,
 # which is why the rule is so simple.
 def _collective_batcher(prim, args, dims, **params):
@@ -593,8 +584,6 @@ def _psum_transpose_rule(cts, *args, axis_name, axis_index_groups):
 psum_p = core.Primitive('psum')
 psum_p.multiple_results = True
 psum_p.def_abstract_eval(lambda *args, **params: map(raise_to_shaped, args))
-pxla.soft_pmap_rules[psum_p] = \
-    partial(_allreduce_soft_pmap_rule, psum_p, lax._reduce_sum)
 xla.parallel_translations[psum_p] = partial(_allreduce_translation_rule, lax.add_p)  # type: ignore
 ad.deflinear2(psum_p, _psum_transpose_rule)
 pxla.multi_host_supported_collectives.add(psum_p)
@@ -957,14 +946,8 @@ def _axis_index_translation_rule(c, *, axis_name, axis_env, platform):
   unsigned_index = xops.Rem(xops.Div(xops.ReplicaId(c), div), mod)
   return xops.ConvertElementType(unsigned_index, xb.dtype_to_etype(np.int32))
 
-def _axis_index_soft_pmap_rule(vals, mapped, chunk_size, *, axis_name):
-  assert not vals and not mapped
-  idx = axis_index(axis_name)  # type: ignore
-  return idx * chunk_size + np.arange(chunk_size, dtype=np.int32), True
-
 axis_index_p = core.Primitive('axis_index')
 xla.parallel_translations[axis_index_p] = _axis_index_translation_rule
-pxla.soft_pmap_rules[axis_index_p] = _axis_index_soft_pmap_rule  # type: ignore
 axis_index_p.def_abstract_eval(
     lambda *args, **params: ShapedArray((), np.int32))
 pxla.multi_host_supported_collectives.add(axis_index_p)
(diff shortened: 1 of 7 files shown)


jax/_src/lax/parallel.py
```diff
@@ -980,6 +963,7 @@ def _axis_index_bind(*, axis_name):
 axis_index_p.def_custom_bind(_axis_index_bind)
 
 def _process_axis_index(self, frame):
+  assert frame.size is not None
```
apaszke, line 966: When can that happen?

jax/experimental/maps.py
```diff
@@ -668,3 +669,27 @@ def subst_eqn_axis_names(eqn, axis_subst: Dict[AxisName, Tuple[AxisName]]):
     axis_names = (axis_names,)
   new_axis_names = sum((axis_subst.get(name, (name,)) for name in axis_names), ())
   return eqn._replace(params=dict(eqn.params, axis_name=new_axis_names))
+
+
+# -------- soft_pmap --------
+
+def soft_pmap(fun: Callable, axis_name: Optional[AxisName] = None, in_axes=0
```
apaszke, line 676: If we're deleting `soft_pmap` from API (which is a breaking change), why don't we remove it completely? Is it because there are still some users internally and you'll need to fix their code?

jax/interpreters/batching.py
```diff
@@ -33,11 +33,13 @@ def batch(fun: lu.WrappedFun, axis_name, axis_size, in_dims, out_dim_dests,
...
   fun, out_dims_thunk = batch_subtrace(fun)
   return _match_axes(batchfun(fun, axis_name, axis_size, in_dims),
-                     axis_size, out_dims_thunk, out_dim_dests)
+                     axis_size, in_dims, out_dims_thunk, out_dim_dests)
 
 @lu.transformation
 def batchfun(axis_name, axis_size, in_dims, *in_vals):
   # analogue of `jvpfun` in ad.py
+  if axis_size is None:
+    axis_size, = {x.shape[d] for x, d in zip(in_vals, in_dims) if d is not not_mapped}
```
apaszke, line 42: Can't we do this check once in `def batch` instead of both in `batchfun` and `_match_axes`?

apaszke (approved):
LGTM!

apaszke:
About the `vtile` issue, it definitely should be possible to compute it at the call sites. It was just convenient to not have to do it, since the batching interpreter would take care of that anyway. I'm ok with going either way, let me know what you prefer.

apaszke:
Also, adding type annotations to the batching functions might have helped us catch this one!

jax/_src/lax/parallel.py
mattjj: This is related to what I mentioned in the PR message: namely, that in #5492 I changed `batching.batch` not to compute the axis size and instead take it as an argument, which worked with all call sites _except_ `pxla.vtile`, where it could be a `None`. The result was a `None` in the core axis env, and it was caught pretty late (only when it was consumed by a rule like this one). This particular assert doesn't catch it much earlier but the error message is slightly cleaner.

jax/experimental/maps.py
mattjj: Good point. Yes, the motivation was to keep downstream users working without having to fix their code, but (1) there are currently only two users (though I remember seeing a few more before so that's suspicious), and (2) we could just add this `xmap` wrapper to _their_ code rather than keeping it in ours.

How about this: I'll follow up with fixing the downstream users and deleting `soft_pmap` from the API entirely. But by landing this PR first we get to remove almost all the `soft_pmap` implementation without having to block on those downstream changes.

jax/interpreters/batching.py
mattjj: Unfortunately, `batch` doesn't have access to `in_vals`. We could factor out another `lu.transformation_with_aux` that we call from `batch`, and pass in the result to both `batchfun` and `_match_axes`, but that seemed a bit more complicated to me.

The fix I'd _really_ like is if we can assume all callers have an int `axis_size` available, and never pass in `None` (so these functions never have to compute the axis size from the args). I think `pxla.vtile` is the only caller which doesn't meet that requirement right now, but I figured it'd be easier just to make this work again now and then I could check with you as to whether `pxla.vtile` could easily be factored so as to have the axis size available.

WDYT? I can explain better if that's unclear.

mattjj:
Great idea on the type annotations; I added some to `batching.batch` as a start.

PR #5533, opened 2021-01-27
Add support for axis names in jnp.{sum,min,max}

apaszke:
#### Generalize the access to axis names embedded in primitives

Previously, a few places in our code assumed that all collectives (i.e.
primitives that operate over named axes) keep all of their axes in the
`axis_name` attribute. This was fine for a few simple use cases, but we
are now considering allowing named axes in many more primitives which
can have semantically different attributes where axis names can appear.

#### Add support for axis names in jnp.{sum,min,max}

Similarly to `jnp.einsum`, whenever we encounter an extension to the
positional NumPy API (in the case of reductions, the extension is
whenever a non-integer axis is specified), we reroute the call to a
parallel primitive instead of the standard lax reductions.

Note that this makes the parallel primitives implement a strict subset
of functionality of the lax reductions so in the future (when we decide
that we want axes to be truly first class) we can always swap out the
implementation for the parallel version. But, it makes sense to keep
them separate for the ease of prototyping in the near future.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -484,6 +484,12 @@ class XeinsumSpecParser:
 
 ### parallel primitives
 
+def _subst_all_names_in_axis_name(params: core.ParamDict, subst: core.AxisSubst) -> core.ParamDict:
+  axis_name = params['axis_name']
+  if not isinstance(axis_name, (tuple, list)):
+    axis_name = (axis_name,)
+  return dict(params, axis_name=sum((subst(name) for name in axis_name), ()))
+
 # This is only used for collectives that do not include the vmapped axis name,
 # which is why the rule is so simple.
 def _collective_batcher(prim, args, dims, **params):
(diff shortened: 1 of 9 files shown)


jax/_src/lax/parallel.py
```diff
@@ -361,34 +367,80 @@ def pdot(x, y, axis_name, pos_contract=((), ()), pos_batch=((), ())):
...
+  outs = prim.bind(*reduced_vals, axes=axes,
                    axis_index_groups=axis_index_groups)
   return outs, (False,) * len(vals)
 
+def _reduction_with_positional_batcher(prim, vals_in, dims_in, axis_index_groups,
+    transform_unmapped, transform_mapped):
+  if axis_index_groups is not None:
+    raise NotImplementedError("axis_index_groups not supported in vmap collectives. "
+                              "Please open a feature request!")
+  # TODO: Transpose all dims to 0, increment all axes
```
mattjj, line 395: I assume this is follow-up work.

mattjj (approved):
LGTM! Looking forward to using this, both the internal utilities and the user-facing API!

mattjj:
Looks like the test failures are just:
1. new tests under non-omnistaging (just skip new tests w/o omnistaging?)
2. jaxprs in docs being sensitive to the "axis_name" -> "axes" rename

jax/_src/lax/parallel.py
apaszke: This is actually done. I just forgot to remove the TODO!

PR #5534, opened 2021-01-27
Update jax2tf_limitations to update the list of integral dtypes for tf.pow

cheshire:
Update jax2tf_limitations to update the list of integral dtypes for tf.pow


cheshire:
@gnecula

PR #5537, opened 2021-01-27
Use myst-nb to build docs

jakevdp:
This PR does two main things:

- Changes our documentation build process from one based on [nbsphinx](https://nbsphinx.readthedocs.io/) to one based on [myst-nb](https://myst-nb.readthedocs.io/)
- Updates `quickstart.md`/`quickstart.ipynb` to use myst-md features & to show how this affects our documentation source (see the generated page [here](https://jax--5537.org.readthedocs.build/en/5537/notebooks/quickstart.html); compare to the version in the [current documentation](https://jax.readthedocs.io/en/latest/notebooks/quickstart.html)).
- Adds several unexecuted notebooks to the docs. Previously in the nbsphinx workflow these were left out of the HTML build.

Advantages of this approach:
- myst-nb can use sphinx cross-referencing directives to make our documentation more dynamic and links more stable.
- myst-nb has a lighter dependency chain (no pandoc!)
- Currently with `nbsphinx` we ignore slow notebooks, removing them from the HTML documentation. MyST-NB lets us mark these as not to be executed,  but still include them in the docs.
 
Disadvantages of this approach:
- the "Open in Colab" links now must be added manually to each notebook, rather than being inserted automatically by nbsphinx.
- minor: myst-nb does not have stored output cells, so we must treat the ipynb sources as primary in order to render unexecuted notebooks with code outputs
 
One feature we need is not yet implemented in myst-nb; I have a pull request for it at https://github.com/executablebooks/MyST-NB/pull/296. In the meantime, a temporary monkeypatch in `conf.py` gives us the behavior we want.

On the balance, I think the `myst-nb` approach will give us more flexibility in creating & maintaining JAX documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -66,9 +66,8 @@ extensions = [
     'sphinx.ext.napoleon',
     'sphinx.ext.viewcode',
     'matplotlib.sphinxext.plot_directive',
-    'nbsphinx',
     'sphinx_autodoc_typehints',
-    'myst_parser',
+    'myst_nb',
 ]
 
 intersphinx_mapping = {
@@ -81,9 +80,11 @@ intersphinx_mapping = {
 templates_path = ['_templates']
 
 # The suffix(es) of source filenames.
-# You can specify multiple suffix as a list of string:
-#
-source_suffix = '.rst'
+# Note: important to list ipynb before md here: we have both md and ipynb
+# copies of each notebook, and myst will choose which to convert based on
+# the order in the source_suffix list. Notebooks which are not executed have
+# outputs stored in ipynb but not in md, so we must convert the ipynb.
+source_suffix = ['.rst', '.ipynb', '.md']
 
 # The master toctree document.
 main_doc = 'index'
@@ -99,18 +100,12 @@ language = None
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = [
-    # Slow notebook: long time to load tf.ds
-    'notebooks/neural_network_with_tfds_data.ipynb',
-    # Slow notebook
-    'notebooks/Neural_Network_and_Data_Loading.ipynb',
-    'notebooks/score_matching.ipynb',
-    'notebooks/maml.ipynb',
-    # Fails with shape error in XL
-    'notebooks/XLA_in_Python.ipynb',
     # Sometimes sphinx reads its own outputs as inputs!
     'build/html',
+    'build/jupyter_execute',
+    'notebooks/README.md',
     'README.md',
-    # Ignore markdown source for notebooks; nbsphinx builds from the ipynb
+    # Ignore markdown source for notebooks; myst-nb builds from the ipynb
     # These are kept in sync using the jupytext pre-commit hook.
     'notebooks/*.md'
 ]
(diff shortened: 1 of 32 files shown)


jakevdp:
Just pushed a commit that includes a monkeypatched logger that turns execution failure warnings into errors. I don't think this should be merged, but it's the behavior we need 😁

jakevdp:
Removed the monkey patch... I've been testing this with the myst-nb branch at https://github.com/executablebooks/MyST-NB/pull/296 and it works as expected. If the myst developers choose to accept that patch, we'll be ready to start converting notebooks to myst-nb

gnecula:
The monkey patch is reasonable in my opinion. If it is the only blocked for us to start experimenting with some text-based notebooks, I would be in favor of having the monkey patch in our conf.py until myst-nb fixes this issue.

docs/conf.py
```diff
@@ -120,74 +121,6 @@
 autosummary_generate = True
 napolean_use_rtype = False
 
-# -- Options for nbsphinx -----------------------------------------------------
-
-# Execute notebooks before conversion: 'always', 'never', 'auto' (default)
-# We execute all notebooks, exclude the slow ones using 'exclude_patterns'
-nbsphinx_execute = 'always'
```
gnecula, line 127: We still have notebooks, until we transition them. Should we drop these options already?

gnecula: I think I understand, myst-nb will execute both .md and .ipynb files.

docs/notebooks/quickstart.md
```diff
@@ -0,0 +1,206 @@
+---
+jupytext:
```
gnecula, line 2: You probably don't want this table here.

docs/notebooks/quickstart.md
```diff
@@ -0,0 +1,206 @@
...
+
+JAX NumPy functions work on regular NumPy arrays. 
+
+```{code-cell}
+import numpy as np
+x = np.random.normal(size=(size, size)).astype(np.float32)
+%timeit jnp.dot(x, x.T).block_until_ready()
+```
+
+That's slower because it has to transfer data to the GPU every time. You can ensure that an NDArray is backed by device memory using {func}`~jax.device_put`.
```
gnecula, line 73: The {func}`~jax.device_put` does not seem to produce a link

gnecula: I am judging this by the GitHub view feature, which is why I think that I am not seeing the links

gnecula: I found a [link](https://jax--5537.org.readthedocs.build/en/5537/notebooks/quickstart.html) to the rendered documentation.

gnecula (commented):
Very nice!

docs/conf.py
jakevdp: Right - nbsphinx and myst-nb cannot be used together. This PR removes nbsphinx & changes it so that myst-nb is in charge of *all* notebooks, so the nbsphinx settings are no longer relevant.

docs/notebooks/quickstart.md
jakevdp: This table is what marks the file as a myst-nb document. It does not appear in the rendered docs.

docs/notebooks/quickstart.md
jakevdp: Yes, all of these files are text source (similar to RST) and cross-references & links will not render except in the sphinx build.

jakevdp:
Rebased on master and description updated; ready for another look.

gnecula:
I think that keeping two identical copies of the notebooks is not great. It creates confusion, it complicates the process for changing the documentation. I am thinking that somebody who just wants to fix a typo will be put off by the process, e.g., Googlers who would like to submit a small internal change.

If the only reason we keep the ipynb in the source tree is for colab to be able to read it easily, can we find some alternative solution, e.g., (a) generate the ipynb on the fly, (b) generate and store it automatically somewhere outside the code tree, e.g., readthedocs, (c) provide a simple manual/tooled recipe for importing the .md in Colab.

jakevdp:
The reason to keep the ipynb source in the github tree is so that the notebooks can be opened in Colab. If we're OK with removing that option, then we can have a single textual source.

Note that the two-source decision is not part of this PR; that was dome previously and is live in master.

mattjj (approved):
I'd like to unblock Jake on this, so I hope no one minds if I throw my 2¢ into the ring.

@gnecula raises good points about tradeoffs (e.g. having one extra step in the process versus not being able to 'open in colab'). We can't be 100% sure a priori if any given tradeoff will be worthwhile, but if I had to guess, based on the frequency with which docs are read vs written, it seems much more valuable to keep the 'open in colab' button.

On top of that, many of the doc updates in the near future may be from @jakevdp himself, so his opinion/workflow matters a lot and he should experiment with docs setups as he sees fit. In other words, even if my a priori guess were different than it is, I'd defer to Jake.

In any case, this seems like a [type 2 decision](https://www.businessinsider.com/jeff-bezos-on-type-1-and-type-2-decisions-2016-4) which we can easily revise if we need to. If the double-source-file solution ends up being onerous, or we realize we don't care about 'open in colab' buttons, we can easily switch back!

So I suggest let's merge this!

gnecula:
I did not mean to block this particular PR, especially since the 2 copies of notebooks is already in master. So, yes, let's merge.

PR #5562, opened 2021-01-30
Bugfix to allow `pxla.replicate` to generate Replicated buffers.

pschuh:


gnecula:
Thank you for the fix. Is it easy to add a test that shows this bug is fixed?

skye (approved):

PR #5567, opened 2021-01-30
update out of date comment in Pmap_Cookbook.ipynb

matpalm:
update comment regarding `pmap` not supporting `in_axes`, because it does now.


gnecula:
Thanks for the fix.

PR #5573, opened 2021-01-31
Changes to support numpy 1.20

gnecula:
One of the changes in numpy 1.20 is to add more type annotations.
However, this sometimes make mypy give errors. A common example is
numpy.take, which with the new type annotation does not appear to
mypy as indexable.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -591,10 +591,10 @@ def conv_general_dilated(
     rhs_dilation = (1,) * (rhs.ndim - 2)
   if isinstance(padding, str):
     lhs_perm, rhs_perm, _ = dnums
-    rhs_shape = np.take(rhs.shape, rhs_perm)[2:]
+    rhs_shape = np.take(rhs.shape, rhs_perm)[2:]  # type: ignore[index]
     effective_rhs_shape = [(k-1) * r + 1 for k, r in zip(rhs_shape, rhs_dilation)]
     padding = padtype_to_pads(
-        np.take(lhs.shape, lhs_perm)[2:], effective_rhs_shape,
+        np.take(lhs.shape, lhs_perm)[2:], effective_rhs_shape,  # type: ignore[index]
         window_strides, padding)
   return conv_general_dilated_p.bind(
       lhs, rhs, window_strides=tuple(window_strides), padding=tuple(padding),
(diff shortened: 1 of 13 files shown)


gnecula:
It seems that on top of the previous changes, all we need for numpy 1.20 is to fix some mypy annotations. This is ready to submit.

build/test-requirements.txt
```diff
@@ -1,6 +1,6 @@
 flake8
 # For now, we pin the numpy version here
-numpy>=1.16,<1.20
+numpy>=1.16
```
hawkinsp, line 3: Unfortunately you will need to leave this pin here, because the released jaxlib has a < 1.20 constraint. I have removed it but I only noticed after the build was complete.

jax/_src/lax/lax.py
```diff
@@ -591,10 +591,10 @@ def conv_general_dilated(
     rhs_dilation = (1,) * (rhs.ndim - 2)
   if isinstance(padding, str):
     lhs_perm, rhs_perm, _ = dnums
-    rhs_shape = np.take(rhs.shape, rhs_perm)[2:]
+    rhs_shape = np.take(rhs.shape, rhs_perm)[2:]  # type: ignore[index]
```
hawkinsp, line 594: I'm concerned that most of these look like either NumPy bugs or mypy bugs, in the sense that valid code is being rejected by inaccurate type annotations or inadequate type inference. File upstream bugs? The NumPy folks say in their release notes they want feedback on type annotations.

build/test-requirements.txt
hawkinsp: Strangely the CI actually passes. Somehow the package installation doesn't notice the conflict. However, it does show up if I install locally with pip.

gnecula: So pip will install 1.20 and then it will just abort when trying to install jaxlib? I will change and leave a comment.

jax/_src/lax/lax.py
gnecula: you are right, but there was just too much going on with this numpy upgrade. Now that this has calmed down I will try to file a bug upstream. I am not sure yet if it is a mypy only problem or a pytype also.

build/test-requirements.txt
gnecula: I left <1.20 in and I added a comment in CHANGELOG.md in a place that I thought the next person to release jaxlib will see. That is when we should relax the test-requirements pin.

PR #5580, opened 2021-02-01
Try to download bazel as the last resort

daskol:
In some sense this pull request reverts changes suggested in #2217. That changes force to download bazel. Now resolution order of paths to bazel binary is as follows.

1. Use `--bazel_path` command line option.
2. Search bazel binary in `PATH` environment variable.
3. Download required bazel release.

It is worth to note that there are excessive version checks. The first is done by bazel with `.bazelversion` (from [version 2.0.0][1]). The second is done manually in build script. Which one is really mandatory?

[1]: https://blog.bazel.build/2019/12/19/bazel-2.0.html#other-important-changes

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -178,9 +178,7 @@ def check_bazel_version(bazel_path):
   if match is None:
     return False
   actual_ints = [int(x) for x in match.group(1).split(".")]
-  if (2, 0, 0) > actual_ints:
-    return False
-  return True
+  return actual_ints >= (2, 0, 0)
 
 
 BAZELRC_TEMPLATE = """
--- a/build/build.py
+++ b/build/build.py
@@ -145,7 +145,7 @@ def download_and_verify_bazel():
 
 
 def get_bazel_paths(bazel_path_flag):
-  """Yields a sequence of a guess about bazel path. Some of sequence elements
+  """Yields a sequence of guesses about bazel path. Some of sequence elements
   can be None. The resulting iterator is lazy and potentially has a side
   effects."""
   yield bazel_path_flag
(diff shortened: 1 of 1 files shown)


daskol:
@googlebot I signed it!

build/build.py
```diff
@@ -432,8 +437,7 @@ def main():
   os.chdir(os.path.dirname(__file__ or args.prog) or '.')
 
   # Find a working Bazel.
-  bazel_path = get_bazel_path(args.bazel_path)
-  check_bazel_version(bazel_path, min_version="2.0.0", max_version=None)
+  bazel_path = get_bazel_path(args.bazel_path, min_version="2.0.0", max_version=None)
```
skye, line 440: My understanding is that the min_version is 2.0.0 here because that the min version that compares against .bazelversion. If so, maybe remove the min_version and max_version arguments here and just check directly in `check_bazel_version` that the version is >= 2?

skye (changes requested):
This a nice improvement, thanks for contributing! Just one small comment.

Also, @hawkinsp notes that https://github.com/bazelbuild/bazelisk may be able to replace some of this functionality. I like this change as-is, but we'd also be open to using bazelisk if you're familiar with it or wanted to look into it.

build/build.py
daskol: Yes, this is exactly what I wanted to do but I wasn't sure whether it was appropriate with you.

To be honest I would like to remove this check completely and update installation guide with requirements to bazel version to be >=2. This version check will become obsolete sometime and will be removed. Maybe it is out of date? I known nothing about bazel 2.0.0 adoption rate but it was released 19 December 2019.

skye: I also have no idea what bazel versions most people are running. I think we should keep the 2.0.0 check in this file just in case -- if someone is using an ancient bazel version, it'll really help with debugging, since using an old bazel usually produces a confusing error message. I can also imagine people having really old bazel versions laying around since relatively few projects use bazel.

I'm fine with this as-is, but WDYT of at least removing the min_version/max_version arguments here and hardcoding the 2.0.0 version check? I think that'll make this file less confusing.

build/build.py
```diff
@@ -144,44 +144,43 @@ def progress(block_count, block_size, total_size):
...
-    sys.exit(-1)
-  if max_version is not None:
-    max_ints = [int(x) for x in max_version.split(".")]
-    if actual_ints >= max_ints:
-      print("Please downgrade your bazel revision to build JAX (>= {} and < {}"
-            " required, found {})".format(min_version, max_version, version))
-      sys.exit(-1)
+  if (2, 0, 0) > actual_ints:
+    return False
+  return True
```
skye, line 183:
```suggestion
  return actual_ints >= (2, 0, 0)
```

skye (approved):
Thanks!

PR #5583, opened 2021-02-01
Add support for axis names in jax.scipy.special.logsumexp

apaszke:


mattjj (approved):
LGTM!

Will we need a `pbroadcast` call here when `keepdims=True` and reducing over a named axis? If so we might want to leave a comment or something for our future selves.

apaszke:
I replied in chat. We don't need any `pbroadcast`s here! `expand_dims` is only necessary to avoid shape errors in positional broadcasting, because removing dimensions can cause misalignment. But there's no misalignment when you have names!

mattjj:
Ah of course, you're right! It just keeps singleton dimensions, because positional axes are bad.

PR #5586, opened 2021-02-01
Fix a number of NumPy 1.20.0 test failures.

hawkinsp:


shoyer (approved):

tests/lax_numpy_test.py
```diff
@@ -4340,9 +4341,16 @@ def testLinspace(self, start_shape, stop_shape, num, endpoint, retstep, dtype):
...
+      # NumPy 1.20.0 changed the semantics of linspace to floor for integer
+      # dtypes.
+      if numpy_version >= (1, 20) and np.issubdtype(dtype, np.integer):
+        np_op = lambda start, stop: np.linspace(
+          start, stop, num,
+          endpoint=endpoint, retstep=retstep, dtype=dtype, axis=axis)
+      else:
+        np_op = lambda start, stop: np.linspace(
+          start, stop, num,
+          endpoint=endpoint, retstep=retstep, axis=axis).astype(dtype)
```
jakevdp, line 4353: Is this condition backward?

jakevdp (approved):

tests/lax_numpy_test.py
hawkinsp: Actually no, but it was slightly incorrect. The behavior of `jnp` is now the NumPy 1.20 behavior, so for old NumPy we need to imitate that behavior.

PR #5589, opened 2021-02-02
Initialize variables in LAPACK work size queries

ahoenselaar:
Prevents false positives in in memory sanitizers.

Related prior art in SciPy: https://github.com/scipy/scipy/pull/9054


gnecula:
Thanks!

PR #5593, opened 2021-02-02
Make eigh work for bfloat16 inputs.

hawkinsp:


hawkinsp:
@shoyer convinced me this is not the right fix; I am going to abandon this change and instead disable bfloat16 in the eigh test harness for jax2tf.

PR #5594, opened 2021-02-02
Disable bfloat16 in jax2tf eigh test.

hawkinsp:
We don't have any bfloat16 implementations, so we shouldn't be testing it.

@gnecula

### DIFF
--- a/jax/experimental/jax2tf/tests/primitive_harness.py
+++ b/jax/experimental/jax2tf/tests/primitive_harness.py
@@ -1659,9 +1659,14 @@ for dtype in jtu.dtypes.all_inexact:
             devices="tpu",
             dtypes=[np.complex64, np.complex128]),
           Limitation(
-            "unimplemented", devices="cpu", dtypes=[np.float16]),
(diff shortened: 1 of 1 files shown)


jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1659,9 +1659,14 @@ def _make_triangular_eigh_operand(shape, dtype, lower: bool, rng: Rng):
             devices="tpu",
             dtypes=[np.complex64, np.complex128]),
           Limitation(
-            "unimplemented", devices="cpu", dtypes=[np.float16]),
+            "unimplemented", devices="cpu",
```
gnecula, line 1662: This is a fine change, but it can be done with a single `Limitation` that applies to all devices. 

I suspect that you figured it out, but this does more than disable the `jax2tf` tests (as the title of the PR says). This says that `eigh` is not implemented in JAX, indirectly this will disable also the jax2tf tests for eigh and other tests built on primitive harnesses.

PR #5599, opened 2021-02-02
Fixes for numpy 1.20

gnecula:
It seems that `np.ndim` and `np.shape` fail in version 1.20
when used on sequences of Tracers. We avoid invoking these
functions in places where this may happen.

Instead of generalizing `np.ndim` we plan to tighten the API
of `jax.numpy` functions to not accept tuples or lists in lieu
or arrays. 

It is likely that this change will break users. Let's see how the tests are going.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -662,9 +662,11 @@ def cond(*args, **kwargs):
   return _cond(*args, **kwargs)
 
 def _cond(pred, true_fun: Callable, false_fun: Callable, operand):
-  if not isinstance(pred, Sequence) and len(np.shape(pred)) != 0:
+  if isinstance(pred, Sequence) or np.ndim(pred) != 0:
     raise TypeError(
-        f"Pred must be a scalar, got {pred} of shape {np.shape(pred)}.")
+        f"Pred must be a scalar, got {pred} of " +
+        (f"type {type(pred)}" if isinstance(pred, Sequence)
+         else f"shape {np.shape(pred)}."))
 
   try:
     pred_dtype = dtypes.result_type(pred)
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -662,7 +662,7 @@ def cond(*args, **kwargs):
   return _cond(*args, **kwargs)
 
 def _cond(pred, true_fun: Callable, false_fun: Callable, operand):
-  if len(np.shape(pred)) != 0:
+  if not isinstance(pred, Sequence) and len(np.shape(pred)) != 0:
     raise TypeError(
         f"Pred must be a scalar, got {pred} of shape {np.shape(pred)}.")
 
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -131,6 +131,20 @@ ndim = _ndim = np.ndim
 size = np.size
 _dtype = dtypes.result_type
 
+def _ndim_sequence(xs):
+  """A version of np.ndim that works on sequences that may include Tracers.
+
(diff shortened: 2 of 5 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2586,7 +2600,7 @@ def _pad(array, pad_width, mode, constant_values, stat_length, end_values, refle
 (In numpy, "function" mode's argument should modify a rank 1 array in-place.)
 """)
 def pad(array, pad_width, mode="constant", **kwargs):
-  pad_width = _broadcast_to_pairs(pad_width, ndim(array), "pad_width")
+  pad_width = _broadcast_to_pairs(pad_width, _ndim_sequence(array), "pad_width")
```
hawkinsp, line 2603: Isn't this the wrong fix? As a rule, don't we require numpy functions like `jnp.pad` to accept only arrays, not lists?

I think I would add a call to `_check_arraylike(array)` and fix any callers that are passing non-arrays here.

jax/_src/numpy/lax_numpy.py
```diff
@@ -4624,6 +4638,15 @@ def _is_int_arraylike(x):
           or issubdtype(getattr(x, "dtype", None), np.integer)
           or isinstance(x, (list, tuple)) and _all(_is_int_arraylike(e) for e in x))
 
+def _is_int_scalar_like(x):
```
hawkinsp, line 4641: Why not separate the "is scalar" aspect from the "type" aspect? i.e., revert the part of the change about checking types, and only have the part of the change about checking scalar-ness.

I'm wondering if you could just define scalarness something like this:
```
def _is_scalar(x):
  return  np.isscalar(x) or (isinstance(x, ndarray) and np.ndim(x) == 0)
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -131,6 +131,20 @@ def __init__(shape, dtype=None, buffer=None, offset=0, strides=None,
 size = np.size
 _dtype = dtypes.result_type
 
+def _ndim_sequence(xs):
```
hawkinsp, line 134: I don't think you need this, see below.

jax/_src/lax/control_flow.py
```diff
@@ -662,7 +662,7 @@ def cond(pred, true_fun, false_fun, operand):
   return _cond(*args, **kwargs)
 
 def _cond(pred, true_fun: Callable, false_fun: Callable, operand):
-  if len(np.shape(pred)) != 0:
+  if not isinstance(pred, Sequence) and len(np.shape(pred)) != 0:
```
hawkinsp, line 665: I find this fix a little confusing. Wouldn't a better fix be:
```
if isinstance(pred, Sequence) or np.ndim(pred) != 0:
...
```

hawkinsp (changes requested):

jax/_src/lax/control_flow.py
gnecula: Done

jax/_src/numpy/lax_numpy.py
gnecula: Sounds good, I will fix the callers.

jax/_src/numpy/lax_numpy.py
gnecula: Deleted

jax/_src/numpy/lax_numpy.py
gnecula: I replaced these with
```
_is_scalar(x) and issubdtype(_dtype(x), np.integer)
```

hawkinsp (approved):

PR #5603, opened 2021-02-02
Improve some xmap error messages

apaszke:
We used to raise really bad errors when:
- we failed to infer an axis size
- axis sizes were not divisible by the number of resources assigned to them
- axis sizes were inconsistent between arguments
- in/out_axes referring to out-of-bounds positional axes
- in/out_axes using negative indices for positional axes (this is not implemented yet)

### DIFF
--- a/jax/experimental/maps.py
+++ b/jax/experimental/maps.py
@@ -166,20 +166,33 @@ def fresh_resource_name(tag=None):
   finally:
     _next_resource_id += 1
 
+
 # This is really a Dict[AxisName, int], but we don't define a
 # pytree instance for it, so that it is treated as a leaf.
 class AxisNamePos(FrozenDict):
-  pass
+  user_repr: Any
+
+  def __init__(self, *args, user_repr, **kwargs):
+    super().__init__(*args, **kwargs)
+    self.user_repr = user_repr
+
+
+# str(...) == 'Ellipsis' which is really annoying
+class DotDotDotRepr:
+  def __repr__(self): return '...'
+
 
 def _parse_entry(arg_name, entry):
   # Dictionaries mapping axis names to positional axes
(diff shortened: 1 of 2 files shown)


jax/experimental/maps.py
```diff
@@ -166,20 +166,33 @@ def fresh_resource_name(tag=None):
...
+  user_repr: Any
+
+  def __init__(self, *args, user_repr, **kwargs):
+    super().__init__(*args, **kwargs)
+    self.user_repr = user_repr
+
+
+# str(...) == 'Ellipsis' which is really annoying
+class DotDotDotRepr:
+  def __repr__(self): return '...'
```
mattjj, line 182: If `user_repr` above is always a string (as it is at the only two current call sites), instead of having this class we could just have a literal `'...'` where the class constructor is currently called. Maybe you have some generalization in mind though, where it'd be handy to have these objects in tuples or lists (and maybe `user_repr` isn't always a string)?

mattjj (approved):
Great!

jax/experimental/maps.py
apaszke: Replacing `user_repr=str(entry[:-1] + [DotDotDotRepr()])` with `user_repr=str(entry[:-1] + ['...'])`would change `['i', 'j', ...]` (which is what the user writes) with `['i', 'j', '...']` which might be a little confusing. I use the dummy class just to avoid the extra quotes.

mattjj: I was thinking that it'd be outside the `str` call rather than adding it to the list, like `user_repr=f"[{', '.join(map(repr, entry[:-1]))}, ...]"`. Now that I write it out it's not obviously simpler than having the class, though it is a bit more local.

Anyway, I don't mean to bikeshed! Anything LGTM.

Issue #5532, opened 2021-01-27
experimental.enable_x64 fails for random.uniform


jakevdp:
Short repro:
```python
from jax import random
from jax.experimental import disable_x64, enable_x64

f = lambda: random.uniform(random.PRNGKey(0), (10,), 'float64', -1, 1)
with disable_x64():
    f()
with enable_x64():
    f()
    f()  # <--- fails
```
Failure:
```pytb
Traceback (most recent call last):
  File "tmp.py", line 11, in <module>
    f()  # <--- fails
  File "tmp.py", line 4, in <lambda>
    f = lambda: random.uniform(random.PRNGKey(0), (10,), 'float64', -1, 1)
  File "/Users/vanderplas/github/google/jax/jax/_src/random.py", line 365, in uniform
    return _uniform(key, shape, dtype, minval, maxval)  # type: ignore
jax._src.traceback_util.FilteredStackTrace: RuntimeError: Invalid argument: Argument does not match host shape or layout of computation parameter 1: want s64[], got s32[]: while running replica 0 and partition 0 of a replicated computation (other replicas may have failed as well).

The stack trace above excludes JAX-internal frames.
The following is the original exception that occurred, unmodified.

--------------------

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "tmp.py", line 11, in <module>
    f()  # <--- fails
  File "tmp.py", line 4, in <lambda>
    f = lambda: random.uniform(random.PRNGKey(0), (10,), 'float64', -1, 1)
  File "/Users/vanderplas/github/google/jax/jax/_src/random.py", line 365, in uniform
    return _uniform(key, shape, dtype, minval, maxval)  # type: ignore
  File "/Users/vanderplas/github/google/jax/jax/_src/traceback_util.py", line 139, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/Users/vanderplas/github/google/jax/jax/api.py", line 401, in f_jitted
    return cpp_jitted_f(context, *args, **kwargs)
RuntimeError: Invalid argument: Argument does not match host shape or layout of computation parameter 1: want s64[], got s32[]: while running replica 0 and partition 0 of a replicated computation (other replicas may have failed as well).
```

jakevdp:
This is not a problem with the python JIT path, but is a problem with the CPP jit. I think the issue is here: https://github.com/google/jax/blob/f4b5ff9d466c3f93d850efd772eee0ef03601a05/jax/api.py#L370-L372

[...]

PR #5604, opened 2021-02-02
Make jax_enable_x64 a thread-local value.

jakevdp:
Related to #5532

This adds `jax.config.x64_enabled` as the preferred way to check the X64 state, and makes this property thread-safe (confirmed by a new test that fails on master).

This change fixes the python side of #5532, but the CPP side of the issue remains. This change also includes a test of the CPP cache issue marked as an expected failure for now.

### DIFF
--- a/jax/config.py
+++ b/jax/config.py
@@ -14,6 +14,8 @@
 
 import os
 import sys
+import threading
+from typing import Optional
 
 def bool_env(varname: str, default: bool) -> bool:
   """Read an environment variable and interpret it as a boolean.
@@ -40,7 +42,15 @@ def int_env(varname: str, default: int) -> int:
   return int(os.getenv(varname, default))
 
 
+class _ThreadLocalState(threading.local):
+
+  def __init__(self):
+    self.enable_x64: Optional[bool] = None
+
+
 class Config:
+  _thread_local_state = _ThreadLocalState()
+
   def __init__(self):
     self.values = {}
     self.meta = {}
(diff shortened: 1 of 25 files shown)


jax/config.py
```diff
@@ -138,6 +148,17 @@ def disable_omnistaging(self):
         disabler()
       self.omnistaging_enabled = False
 
+  @property
+  def x64_enabled(self):
+    if _thread_local_state.enable_x64 is not None:
+      return bool(_thread_local_state.enable_x64)
+    else:
+      return bool(self.read('jax_enable_x64'))
```
hawkinsp, line 156: Should we set the thread-local state on first read if it doesn't exist?

jax/config.py
```diff
@@ -40,6 +42,15 @@ def int_env(varname: str, default: int) -> int:
   return int(os.getenv(varname, default))
 
 
+class _ThreadLocalState(threading.local):
```
hawkinsp, line 45: I'm wondering if it would be preferable to merge this state into the same ThreadLocalState object in api.py, where we store the jit is disabled flag.

jakevdp: Originally I had it there, but it caused circular dependencies.

jax/config.py
jakevdp: My reason for not setting it is to avoid breaking current behavior in cases where the user is setting the flag at runtime. But I suppose that pattern is broken to begin with...

I'll make the change.

jax/config.py
jakevdp: I added a TODO to move the thread local state from api to config

hawkinsp: Another option would be somewhere like `core`.

PR #5608, opened 2021-02-03
Add chi2 to jax.scipy.stats

Dpananos:
Hello,

I realized the chisquare distribution was missing from `jax.scipy.stats` so I went ahead and added it.  I've also added some tests as well (they all pass when I run `python tests/scipy_stats_test.py --test_targets="LaxBackedScipyStatsTests.testChi2" --num_generated_cases=100`).

PR #5619, opened 2021-02-03
Add betabinomial logpmf/pmf and tests

Dpananos:
Similar to  #5608, this PR adds logpmf/pmf for betabinomial in jax.scipy.stats

Tests included.

### DIFF
--- a/jax/_src/scipy/stats/betabinom.py
+++ b/jax/_src/scipy/stats/betabinom.py
@@ -1,0 +1,40 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+
+import scipy.stats as osp_stats
+
+from jax import lax
+from jax._src.numpy.util import _wraps
+from jax._src.numpy.lax_numpy import _promote_args_inexact, _constant_like, where, inf, logical_or, nan
+from jax._src.scipy.special import betaln
+
+
+@_wraps(osp_stats.betabinom.logpmf, update_doc=False)
+def logpmf(k, n, a, b, loc=0):
+    k, n, a, b, loc = _promote_args_inexact("betabinom.logpmf", k, n, a, b, loc)
+    y = lax.sub(lax.floor(k), loc)
+    one = _constant_like(y, 1)
+    zero = _constant_like(y, 0)
+    combiln = lax.neg(lax.add(lax.log1p(n), betaln(lax.add(lax.sub(n,y), one), lax.add(y,one))))
+    beta_lns = lax.sub(betaln(lax.add(y,a), lax.add(lax.sub(n,y),b)), betaln(a,b))
+    log_probs = lax.add(combiln, beta_lns)
+    y_cond = logical_or(lax.lt(y, lax.neg(loc)), lax.gt(y, lax.sub(n, loc)))
+    log_probs = where(y_cond, -inf, log_probs)
+    n_a_b_cond = logical_or(logical_or(lax.lt(n, one), lax.lt(a, zero)), lax.lt(b, zero))
+    return where(n_a_b_cond, nan, log_probs)
+
+@_wraps(osp_stats.betabinom.pmf, update_doc=False)
+def pmf(k, n, a, b, loc=0):
+    return lax.exp(logpmf(k, n, a, b, loc))
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -420,6 +420,25 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
                             tol=5e-4)
     self._CompileAndCheck(lax_fun, args_maker)
 
+  @genNamedParametersNArgs(5)
+  def testBetaBinomLogPmf(self, shapes, dtypes):
+    rng = jtu.rand_positive(self.rng())
+    scipy_fun = osp_stats.betabinom.logpmf
+    lax_fun = lsp_stats.betabinom.logpmf
+
+    def args_maker():
+      k, n, a, b, loc = map(rng, shapes, dtypes)
+      k = np.floor(k)
+      n = np.ceil(n)
+      a = np.clip(a, a_min = 0.1, a_max = None)
+      b = np.clip(a, a_min = 0.1, a_max = None)
+      loc = np.floor(loc)
+      return [k, n, a, b, loc]
+
+    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                            tol=5e-4)
+    self._CompileAndCheck(lax_fun, args_maker, rtol=1e-5, atol=1e-5)
+
   def testIssue972(self):
     self.assertAllClose(
       np.ones((4,), np.float32),
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -12,6 +12,7 @@ jax 0.2.10 (Unreleased)
 * `GitHub commits <https://github.com/google/jax/compare/jax-v0.2.9...master>`__.
 * New features:
   * :func:`jax.scipy.stats.chi2` is now available as a distribution with logpdf and pdf methods.
+  * :func:`jax.scipy.stats.betabinom` is now available as a distribution with logpmf and pmf methods.
 
 * Bug fixes:
 
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -129,6 +129,16 @@ jax.scipy.stats.beta
    logpdf
    pdf
 
+jax.scipy.stats.betabinom
+~~~~~~~~~~~~~~~~~~~~
+.. automodule:: jax.scipy.stats.betabinom
+
+.. autosummary::
+  :toctree: _autosummary
+
+   logpmf
+   pmf
+
 jax.scipy.stats.cauchy
 ~~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: jax.scipy.stats.cauchy
--- a/jax/scipy/stats/betabinom.py
+++ b/jax/scipy/stats/betabinom.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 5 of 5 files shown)


tests/scipy_stats_test.py
```diff
@@ -420,6 +420,23 @@ def args_maker():
                             tol=5e-4)
     self._CompileAndCheck(lax_fun, args_maker)
 
+  @genNamedParametersNArgs(5)
+  def testBetaBinomLogPmf(self, shapes, dtypes):
+    rng = jtu.rand_positive(self.rng())
```
jakevdp, line 425: Would it be worth testing negative inputs to make sure they correctly return infinity?

Dpananos: Ah, good catch.  Are you thinking maybe `rng = jtu.rand_default(self.rng())`?  That is what is used to test the bernoulli and Poisson.

jakevdp: Yeah something like that – particularly if the return value in this range is something explicit like `inf`

jax/_src/scipy/stats/betabinom.py
```diff
@@ -0,0 +1,50 @@
...
+def logpmf(k, n, a, b, loc=0):
+    k, n, a, b, loc = _promote_args_inexact("betabinom.logpmf", k, n, a, b, loc)
+    y = lax.sub(lax.floor(k), loc)
+    one = _constant_like(y, 1)
+    zero = _constant_like(y, 0)
+
+    
+    combiln = lax.neg(
+        lax.add(
+            lax.log(lax.add(n, one)),
```
majnemer, line 34: Should `lax.log(lax.add(n, one))` be replaced with `lax.log1p(n)`?

Dpananos: Yes!  Still learning some of the `lax` functions, so if you see more of that let me know.

Dpananos:
I'm runningn into a small problem with this PR re: testing as per @jakevdp 's suggestion.  When I use `jtu.rand_default` in the test, the parameters generated look something like the following:

```python

k = [1., -0.38,  1.36, -0.84]
n = [[ 2.0,  4.23,  0.7,  -1.07],
     [ 0.01,  0.7,   2.08,  2.5 ],
     [-0.92,  0.74, -3.81, -0.54]]

a = [[ 2.77, -0.7,   3.07, -4.9 ],
     [-2.61, -3.69, -3.24, -0.5 ],
     [ 1.78,  5.4,  -1.03,  1.2 ]]

b = [[-0.42],
     [-0.68],
     [-0.04]]

loc = [[0.],
       [ 3.06],
       [-4.31]]
```

If I compare my implementation of the logpmf (without the `where` in the return call), I get something like...

```
import jax.numpy as jnp
from jax import lax
import jax.scipy as lsp

import scipy as osp
import numpy as np

k, n, a, b, loc = map(jnp.array, [k, n, a, b, loc])
lsp.stats.betabinom.logpmf(k=k, n=n, a=a, b=b, loc=loc)

>>>DeviceArray([[ -1.2188828,        -inf,   0.3208853,         nan],
                             [ -7.949671 , -13.266653 , -10.89188  ,  -6.595767 ],
                             [ -1.5104122,  -1.5108628,         nan,  -3.7848833]],            dtype=float32)
```

Comparing this directly with scipy's [implementation of the logpmf](https://github.com/scipy/scipy/blob/v1.6.0/scipy/stats/_discrete_distns.py#L208)

```python
k = np.floor(k) - loc
combiln = -np.log(n + 1) - osp.special.betaln(n - k + 1, k + 1)
combiln + osp.special.betaln(k + a, n - k + b) - osp.special.betaln(a, b)
>>>array([[ -1.2188818,        -inf,   0.3208878,         nan],
                 [ -7.949671 , -13.266655 , -10.891884 ,  -6.595765 ],
                 [ -1.5104126,  -1.5108645,         nan,  -3.784882 ]],
                dtype=float32)
```

Nice, looks similar.  However, calling scipy's betabinom logpmf directly yeilds `nan`

```python
osp.stats.betabinom.logpmf(k=k, n=n, a=a, b=b, loc=loc)
>>>array([[nan, nan, nan, nan],
                 [nan, nan, nan, nan],
                 [nan, nan, nan, nan]])
```

Which is problematic because this is what is compared in the tests.

```
Running tests under Python 3.8.5: /Users/demetri/opt/miniconda3/envs/jax/bin/python
[ RUN      ] LaxBackedScipyStatsTests.testBetaBinomLogPmf_float32[4]_float32[3,4]_float32[3,4]_float32[3,1]_float32[3,1]
W0204 14:08:25.821909 4461870528 xla_bridge.py:135] No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
[  FAILED  ] LaxBackedScipyStatsTests.testBetaBinomLogPmf_float32[4]_float32[3,4]_float32[3,4]_float32[3,1]_float32[3,1]
======================================================================
FAIL: testBetaBinomLogPmf_float32[4]_float32[3,4]_float32[3,4]_float32[3,1]_float32[3,1] (__main__.LaxBackedScipyStatsTests)
LaxBackedScipyStatsTests.testBetaBinomLogPmf_float32[4]_float32[3,4]_float32[3,4]_float32[3,1]_float32[3,1]
testBetaBinomLogPmf_float32[4]_float32[3,4]_float32[3,4]_float32[3,1]_float32[3,1](shapes=((4,), (3, 4), (3, 4), (3, 1), (3, 1)), dtypes=(<class 'numpy.float32'>, <class 'numpy.float32'>, <class 'numpy.float32'>, <class 'numpy.float32'>, <class 'numpy.float32'>))
----------------------------------------------------------------------
  File "/Users/demetri/opt/miniconda3/envs/jax/lib/python3.8/site-packages/absl_py-0.11.0-py3.8.egg/absl/testing/parameterized.py", line 282, in bound_param_test
  File "tests/scipy_stats_test.py", line 433, in testBetaBinomLogPmf
    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
  File "/Users/demetri/gitrepos/jax/jax/test_util.py", line 915, in _CheckAgainstNumpy
    self.assertAllClose(numpy_ans, lax_ans, check_dtypes=check_dtypes,
  File "/Users/demetri/gitrepos/jax/jax/test_util.py", line 847, in assertAllClose
    self.assertArraysAllClose(x, y, check_dtypes=False, atol=atol, rtol=rtol)
  File "/Users/demetri/gitrepos/jax/jax/test_util.py", line 814, in assertArraysAllClose
    _assert_numpy_allclose(x, y, atol=atol, rtol=rtol)
  File "/Users/demetri/gitrepos/jax/jax/test_util.py", line 133, in _assert_numpy_allclose
    np.testing.assert_allclose(a, b, **kw)
  File "/Users/demetri/opt/miniconda3/envs/jax/lib/python3.8/site-packages/numpy-1.19.5-py3.8-macosx-10.9-x86_64.egg/numpy/testing/_private/utils.py", line 1527, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/Users/demetri/opt/miniconda3/envs/jax/lib/python3.8/site-packages/numpy-1.19.5-py3.8-macosx-10.9-x86_64.egg/numpy/testing/_private/utils.py", line 764, in assert_array_compare
    flagged = func_assert_same_pos(x, y, func=isnan, hasval='nan')
  File "/Users/demetri/opt/miniconda3/envs/jax/lib/python3.8/site-packages/numpy-1.19.5-py3.8-macosx-10.9-x86_64.egg/numpy/testing/_private/utils.py", line 740, in func_assert_same_pos
    raise AssertionError(msg)
AssertionError:
Not equal to tolerance rtol=0.0005, atol=0.0005

x and y nan location mismatch:
 x: array([[nan, nan, nan, nan],
       [nan, nan, nan, nan],
       [nan, nan, nan, nan]])
 y: array([[ -2.0639  ,   4.959385,  -4.269639,        nan],
       [-11.134994, -13.355612, -10.916795,  -6.697731],
       [ -1.443631,  -1.582576,        nan,  -3.863479]], dtype=float32)
```

It looks like scipy [does a check to see if [all of `n`, `a`, and `b`](https://github.com/scipy/scipy/blob/v1.6.0/scipy/stats/_discrete_distns.py#L205) are positive.  If not, I think it returns nan.  This might be to prevent singularities in the `betaln` call since the beta function is defined for non negative integers.

So, what should I do in this case?  Should I implement the same sort of argcheck in the logpmf?  If I don't the tests won't pass with `jtu.rand_default`.

Dpananos:
I'll also add that the tests pass if I do as other tests do and clip some of the variables.

```
    def args_maker():
      k, n, a, b, loc = map(rng, shapes, dtypes)
      k = np.floor(k)
      a = np.clip(a, a_min = 0.1, a_max = None)
      b = np.clip(b, a_min = 0.1, a_max = None)
      n = np.ceil(n)
      loc = np.floor(loc)
      return [k, n, a, b, loc]
```

jakevdp:
Hmm... I don't think scipy's behavior here is necessarily something we want to emulate here.

I think it's fine to only test positive values against scipy and add a separate test that explicitly checks that out-of-range inputs lead to inf or nan where appropriate. What do you think?

Dpananos:
Yep, I agree.  Let me see if I can write something up for this then.

Dpananos:
I don't see any other examples of testing distributions to see if poor parameterizations result in `inf` or `nan`.  For instance, the `testBernoulliLogPmf` test forces `p` to be between 0 and 1 through application of `expit`.  Additionally, `mu` in `testPoissonLogPmf` is forced to be positive as well.  

If we want to test that the distributions fail in predictable ways:

a) What would be the best way of doing that?  Should I just do something like `self.assertAlClose( inf, lax_fun(k,n,a,b,loc)` (there doesn't seem to be an `assertAllEqual`).

b) Should other distributions (i.e. all bounded distributions like the poisson and bernoulli) be checked in this way?

jakevdp:
You're right that we don't have great test coverage for out-of-bound inputs... if you want to ignore that in this case, it's probably OK.

If you do want to tackle it, I'd probably use `self.assertArraysEqual()` (which is a thin wrapper around `np.testing.assert_array_equal`).

Adding such tests to other distributions would be worthwhile as well; it might be worth using something like the `op_record` approach in `lax_numpy_test.py` in order to have a single test of invalid values for multiple distributions. But that doesn't have to be part of this PR if you don't want to go down that route!

Dpananos:
OK, this is at a place where I am happy with it.  I'm going to open another issue re: some differences with original scipy and lax scipy.

jakevdp (approved):
Looks great. It appears that a few test tolerances need to be adjusted, but once tests pass we can get this merged!

Also, please squash all changes into a single commit once you're finished. Thanks!

jakevdp:
I'm going to run some additional tests on GPU & TPU backends – I'll let you know if those uncover any issues.

Dpananos:
Sweet. Once you run the GPU/TPU checks then I will squash.  Just gimme the thumbs up when ready

jakevdp:
Everything looks good! We'll get this merged.

jakevdp:
Thanks for the contribution! Just FYI I ended up having to do a minor patch to make the new function work with older scipy, which is currently not tested via github CI (I have a PR at #5654 to rectify this).

Dpananos:
Cool.  Is there anything I should know for future PRs?

jakevdp:
Just that we need compatibility back to scipy 1.2... `betabinom` was added in 1.4. The updated test in #5654 should catch that for future PRs.

PR #5620, opened 2021-02-04
DOC: add examples to jax.tree_util.Partial

jakevdp:
Clarify how the function is intended to work in traced contexts.

### DIFF
--- a/jax/tree_util.py
+++ b/jax/tree_util.py
@@ -327,6 +327,15 @@ class Partial(functools.partial):
   >>> call_func(add_one, 2)
   DeviceArray(3, dtype=int32)
 
+  Passing zero arguments to ``Partial`` effectively wraps the original function,
+  making it a valid argument in JAX transformed functions:
+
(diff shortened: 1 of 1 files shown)


jax/tree_util.py
```diff
@@ -306,6 +306,36 @@ class Partial(functools.partial):
...
+  as an argument within transformed JAX functions, which is not possible with a
+  standard ``functools.partial`` function:
+
+  >>> from jax import jit
+  >>> @jit
+  ... def call_func(f, *args):
+  ...   return f(*args)
+  ...
+  >>> call_func(add_one, 2)
+  DeviceArray(3, dtype=int32)
```
shoyer, line 328: This might be a place to mention that for this reason it can make sense just to write `Partial(jnp.add)` sometimes, without any arguments.

shoyer (approved):

froystig (approved):

jax/tree_util.py
jakevdp: Done

PR #5618, opened 2021-02-03
Add tests for pytree dataclass support

NathanHowell:
These are the tests for https://github.com/tensorflow/tensorflow/pull/46894. I'm not sure what the best way to support conditional tests for Python 3.6 so I just took a stab at it.

Related to https://github.com/google/jax/issues/2371

PR #5621, opened 2021-02-04
Rename {tree_util_tests.py->tree_util_test.py}

NathanHowell:
`pytest` discovery is missing these tests due to the filename not matching it's expected pattern. this is causing erroneous tests to pass in #5618.

/cc @jakevdp


jakevdp (approved):
Nice catch!

NathanHowell:
doh sorry, fat fingered and accidentally closed this.

PR #5627, opened 2021-02-04
Add an experimental mechanism for calling TensorFlow functions from JAX.

gnecula:
The TF function is called in eager mode if called outside a JAX staging
context, or is compiled using TF (``tf.function(jit_compile=True)``) if
called in a staging context. Reverse-mode AD is supported, by
calling TensorFlow's ``tf.GradientTape``.

In a staging context, the implementation uses TF to compile the function
to XLA, and then adds a call from JAX's computation to the compiled
TF function. All metadata that TF puts in the generated XLA is carried
over.

See the `jax.experimental.jax2tf.README` for details.


jax/experimental/jax2tf/call_tf.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+# The impl will be used in op-by-op mode and calls func_tf in TF eager mode.
+def _call_tf_impl(*args, args_treedef, func_tf, **_):
+  # TODO(necula): does this copy to the host?
+  args_tf_flat = [tf.constant(np.asarray(a)) for a in args]
+  res_tf = func_tf(*args_treedef.unflatten(args_tf_flat))
+  # TODO(necula): check the result for tree and aval
+  # TODO(necula): does this copy to the host?
+  return [jnp.asarray(np.asarray(r)) for r in tree_util.tree_flatten(res_tf)[0]]
```
tomhennigan, line 156: nit: `tree_util.tree_flatten(res_tf)[0]` could be `tree_util.tree_leaves(res_tf)`

jax/experimental/jax2tf/call_tf.py
```diff
@@ -0,0 +1,232 @@
...
+call_tf_p.multiple_results = True
+
+
+# The impl will be used in op-by-op mode and calls func_tf in TF eager mode.
+def _call_tf_impl(*args, args_treedef, func_tf, **_):
+  # TODO(necula): does this copy to the host?
+  args_tf_flat = [tf.constant(np.asarray(a)) for a in args]
+  res_tf = func_tf(*args_treedef.unflatten(args_tf_flat))
+  # TODO(necula): check the result for tree and aval
+  # TODO(necula): does this copy to the host?
```
tomhennigan, line 155: Yes.

jax/experimental/jax2tf/call_tf.py
```diff
@@ -0,0 +1,232 @@
...
+        # TODO(necula): better checking that we are picking the right variable
+        var = func_tf_concrete.variables[next_var_idx]
+        assert inp.shape == var.shape
+        next_var_idx += 1
+        inp_const = np.asarray(var)
+      else:
+        inp_const = np.asarray(inp)
+      captured_ops.append(xops.ConstantLiteral(builder, np.asarray(inp_const)))
+
+  # For unoptimized HLO, does not make a different which device we use.
```
tomhennigan, line 207: Are we sure TF won't do similar things to JAX and stage out platform specific implementations of certain things?

jax/experimental/jax2tf/tests/call_tf_test.py
```diff
@@ -0,0 +1,284 @@
...
+
+class CallTfTest(jtu.JaxTestCase):
+
+  def setUp(self):
+    if tf is None:
+      raise unittest.SkipTest("Test requires tensorflow")
+    super().setUp()
+
+  @parameterized_jit
+  def test_eval(self, with_jit=True):
```
tomhennigan, line 59: nit: Why provide a default value for with_jit?

tomhennigan (approved):
Nice tests 😄

jax/experimental/jax2tf/call_tf.py
shoyer: Indeed, it definitely does copy to the host. In principle, the copy could be avoided using DLPack.

jax/experimental/jax2tf/README.md
```diff
@@ -9,6 +9,9 @@ TensorFlow eager mode, or stage it out as a TensorFlow graph, even save it
 as a SavedModel for use with TensorFlow tools such as serving stack,
 or TensorFlow Hub.
 
+The package also contains an experimental mechanism to call TensorFlow functions
+from JAX. See [below](#calling-tensorflow-functions-from-jax).
```
shoyer, line 13: It does probably make sense to keep the TensorFlow inter-op code all in one place, though this sort of feels like something that should live in a module called `tf2jax` instead of `jax2tf` :)

shoyer (approved):
Looks great, though

jax/experimental/jax2tf/tests/call_tf_test.py
gnecula: There is a reason. When I run the test in my IDE, I comment out the @parameterized_jit, and then with one click I can run `test_eval`.

jax/experimental/jax2tf/README.md
gnecula: You are right and I went back and forth there. I considered having this a separate experimental module, call_tf. I did not want to call it tf2jax because it does not really turn TF code into first-class JAX code, all you can do is call it and reverse-mode AD. 

For now I thought I keep it part of jax2tf, to make it easier to keep the documentation and tests together. The tests are different from the rest of JAX because they depend on TF, which is not a requirement for the JAX package.

jax/experimental/jax2tf/call_tf.py
```diff
@@ -0,0 +1,232 @@
...
+
+  # Define the fwd and bwd custom_vjp functions
+  def make_call_vjp_fwd(*args_jax):
+    # Return the primal argument as the residual
+    return make_call(*args_jax), args_jax
+
+  def make_call_vjp_bwd(residual, ct_res):
+    args_jax = residual  # residual is the primal argument
+
+    def tf_vjp_fun(arg, ct_res):
```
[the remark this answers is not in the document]
gnecula: Done

jax/experimental/jax2tf/call_tf.py
gnecula: I will work on this as a separate PR

jax/experimental/jax2tf/call_tf.py
gnecula: Not yet sure. I will turn it into a TODO. I plan to run the 2300+ primitive harness through jax2tf and then back with call_tf to see if that uncovers some surprises.

PR #5628, opened 2021-02-04
Add `jax.default_backend()` which returns the default platform name.

tomhennigan:
This can be useful when you need backend specific behaviour, e.g.:

    if jax.default_backend() == 'gpu':
      dataset = double_buffer(dataset)

Or if you want to assert a given backend is the default:

    assert jax.default_backend() == 'tpu'

I am a bit conflicted by the naming, "backend" is consistent with other APIs in
JAX (e.g. jit, local_devices etc) which accept a "backend" string which is used
to lookup an XLA backend by platform name.

### DIFF
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -32,6 +32,7 @@ Just-in-time compilation (:code:`jit`)
     make_jaxpr
     eval_shape
     device_put
+    default_backend
     named_call
 
 Automatic differentiation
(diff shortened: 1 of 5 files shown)


gnecula:
Maybe `default_backend` might be better? I would have preferred `default_platform`, because we also have `get_backend` that gets a backend object (with a `platform` attribute), but you are right that we use `backend` elsewhere to refer to the platform.

tomhennigan:
I'm happy to change to default_platform if preferred, but afaik all public API symbols use "backend" to refer to platform name.

gnecula (approved):

PR #5629, opened 2021-02-04
Add support for axis names in jax.nn.initializers.variance_scaling

apaszke:
... as well as in a few random functions that it needs (`uniform`,
`normal` and `truncated_normal`). The interface itself doesn't change to
much with the exception of the `shape` arguments of all those functions
now accepting `jax.core.NamedShape` (I didn't move it to be part of the
API just yet, but we can do that any time), which makes it possible to
generate sharded random arrays (in particular the random bits are
different on different shards). I also haven't updated the docstrings,
because I don't know if we're ready to go fully public with this
feature.


jax/core.py
```diff
@@ -1217,6 +1217,79 @@ def canonicalize_shape(shape):
...
+  def __getitem__(self, idx):
+    try:
+      idx = operator.index(idx)
+      return self.__positional[idx]
+    except TypeError:
+      pass
+    return self.__named[idx]
+
+  @property
+  def total(self):
```
mattjj, line 1266: `total_size`? This is fine too though.

tests/xmap_test.py
```diff
@@ -402,6 +403,43 @@ def testReductions(self, reduction, axes, mapped_axis):
...
+  @parameterized.named_parameters(
+    {"testcase_name": name, "distr_sample": sample}
+    for name, sample in [
+      ("Uniform", jax.random.uniform),
+      ("Normal", jax.random.normal),
+      ("Bernoulli", partial(jax.random.bernoulli, p=0.5)),
+      ("TruncatedNormal", partial(jax.random.truncated_normal, lower=-2, upper=2)),
+    ])
+  @ignore_xmap_warning()
+  def testSample(self, distr_sample):
```
mattjj, line 423: It would be nice to test that the values we get out don't depend on the schedule, i.e. whether vectorized or parallelized. Maybe leave that as a TODO?

mattjj (approved):
LGTM!

tests/xmap_test.py
apaszke: This is a great idea! I thought it is important enough that I've added another commit that adds this test. Fortunately, my previous implementation worked ok!

PR #5633, opened 2021-02-04
Don't require ellipsis at the end of each list axis spec in xmap

apaszke:
Skipping the ellipsis will make xmap assert that your inputs are of
ranks matching the length of the list.


mattjj (approved):
This is the best PR of all of them! I can't wait to delete my ellipses.

PR #5634, opened 2021-02-04
Use the C++ objects for the Sharding specification.

jblespiau:
One of 3 Cls to enable some C++ pmap.

I can remove `make_chunked` if you prefer.

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -35,7 +35,8 @@ import itertools as it
 import operator as op
 import threading
 from typing import (Any, Callable, Dict, List, Optional, Sequence, Set, Tuple,
-                    Type, Union, Iterable, no_type_check, NamedTuple, TYPE_CHECKING)
+                    Type, Union, Iterable, no_type_check, NamedTuple,
+                    TYPE_CHECKING)
 
 from absl import logging
 import numpy as np
@@ -48,9 +49,10 @@ from ..abstract_arrays import array_types
 from ..core import ConcreteArray, ShapedArray
 from .._src.util import (partial, unzip2, unzip3, prod, safe_map, safe_zip,
                          extend_name_stack, wrap_name, assert_unreachable,
-                         tuple_insert, tuple_delete, taggedtuple, curry)
+                         tuple_insert, tuple_delete, curry)
 from ..lib import xla_bridge as xb
 from ..lib import xla_client as xc
+from ..lib import pmap_lib
 from ..tree_util import tree_flatten, tree_map
 from . import batching
 from . import partial_eval as pe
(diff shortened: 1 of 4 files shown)


jax/interpreters/pxla.py
```diff
@@ -71,77 +72,26 @@
...
-# mypy is very unhappy about taggedtuple
-if TYPE_CHECKING:
-  class Unstacked(NamedTuple):
-    size: int
-else:
-  Unstacked = taggedtuple('Unstacked', ('size',))
 
-class Chunked:
-  chunks: Tuple[int, ...]
+def make_chunked(chunks: Union[int, Sequence[int]]):
```
apaszke, line 83: Can't you subclass `pmap_lib.Chunked` and override `__init__`?

jax/interpreters/pxla.py
```diff
@@ -71,77 +72,26 @@
...
-    return f'Chunked({self.chunks})'
-
-"""
-Represents all the ways we can shard a dimension.
-- `None` means no sharding;
-- `Chunked` means that the dimension is split into the specified number of chunks,
-  but the split dimension itself is preserved inside the map;
-- `Unstacked` means that the dimension is split into chunks of size 1, and doesn't
-  appear inside the map.
-"""
```
apaszke, line 122: Lost comment, which is still accurate

jax/interpreters/pxla.py
```diff
@@ -71,77 +72,26 @@
...
-
-"""
-Assigns sharded axes to mesh dimensions.
-
-When no axis is assigned, the data is replicated.
-Note that `ShardedAxis(2)` refers to the second actually sharded axis (i.e.
-counting as if the None dimensions of sharding were filtered out). For example,
-given the sharding `[Unstacked(n), None, Chunked(m)]`, an entry of `ShardedAxis(1)`
-refers to the `Chunked(m)` axis, not the `None`.
-"""
```
apaszke, line 143: Lost comment again

jax/interpreters/pxla.py
```diff
@@ -299,7 +249,9 @@ def indices(self, shape: Tuple[int, ...]) -> np.ndarray:
               .transpose(perm))
 
   def __eq__(self, other):
-    return (self.sharding, self.mesh_mapping) == (other.sharding, other.mesh_mapping)
+    return (self.sharding, self.mesh_mapping) == (other.sharding,
+                                                  other.mesh_mapping)
+
```
apaszke, line 255: nit: lots of whitespace between methods

jax/interpreters/pxla.py
apaszke: Alternatively, I think we can just replace all uses of this with a more explicit list comprehension that constructs `Chunked` with singleton lists

jax/interpreters/pxla.py
jblespiau: Replied in the hereabove comment.

jax/interpreters/pxla.py
jblespiau: The comments has been moved to the C++ (it's said on line 75 with
# See the C++ code for comments.
)

See https://github.com/tensorflow/tensorflow/blob/a1f732f7b04a152fc1b2d493cab59a686a0bdb7a/tensorflow/compiler/xla/python/pmap_lib.h#L54

Given the Python only reference the C++ code, and to prevent having to duplicate documentation that increases the chances that one of the 2 versions of the same doc get stale, I think it's a better idea to keep only once place for the comments. I think that, given it's defined in C++, it makes sense to have the doc there.

WDYT?

jax/interpreters/pxla.py
jblespiau: I guess it depends on the conventions there:
- I am working from google3, the required spacing is 2 new lines between functions (1 between methods).
- In Jax, it seems more frequent to have one line between functions, but it depends.

I have been conforming to google3 style when submitting code, and in practice, I adapt to each reviewer (but I do have a small preference for google3 style, given my day to day work, deals with these conventions), but I can understand too if JAX wants to enforce a different style -- in which case, I would encourage the team to configure some linter, that will say what it should be, and does not let each individual decides what they prefer.

Given that, what do you prefer?

jblespiau:
Updated, from my google3 CL, applying some of the comments (mainly Chunked), and answering to some of the comments.

jax/interpreters/pxla.py
jblespiau: This was a method, removed one new line (the next CL is turning it into a function).

jblespiau:
The follow up is https://github.com/google/jax/pull/5712

jax/interpreters/pxla.py
apaszke: Sounds good to me.

jax/interpreters/pxla.py
```diff
@@ -71,77 +72,20 @@
...
-  ShardedAxis = taggedtuple('ShardedAxis', ('axis',))
-  Replicated = taggedtuple('Replicated', ('replicas',))
+ShardedAxis = pmap_lib.ShardedAxis
+Replicated = pmap_lib.Replicated
 
-"""
-Assigns sharded axes to mesh dimensions.
+# mypy cannot deal with the C++ types.
+AvalDimSharding = Union[Unstacked, Chunked, NoSharding]  # type: ignore
+MeshDimAssignment = Union[ShardedAxis, Replicated]  # type: ignore
```
apaszke, line 87: Instead of `# type: ignore`, you could use the `if TYPE_CHECKING` trick I used. That would be a bit better because it would at least give us better coverage.

jax/interpreters/pxla.py
```diff
@@ -1172,8 +1117,13 @@ def partitioned_sharding_spec(num_partitions: int,
         mesh_mapping=maybe_replicate)
   else:
     assert len(partitions) == len(aval.shape)
-    return ShardingSpec(sharding=map(Chunked, partitions),
-                        mesh_mapping=map(ShardedAxis, range(len(partitions))))
+    # Chunked expects a list of integers and never an integer.
+    partitions_for_chunked = [
+        [x] if isinstance(x, int) else x for x in partitions
+    ]
```
apaszke, line 1123: nit: the whitespace in here is a bit excessive. We don't forcefully follow Google conventions for line length, and IMO it is more readable if you fold it. But not a hard constraint

apaszke: Also, how would `x` be anything else than an `int`? The type annotation explicitly rules it out.

apaszke (commented):
Should be good to merge once the TYPE_CHECKING conditionals are fixed and `partitions_for_chunked` is simplified.

jax/interpreters/pxla.py
jblespiau: Done. (it's just that I am not really sure the type annotation is correctly enforced here.).

jax/interpreters/pxla.py
jblespiau: Done.

apaszke (approved):

PR #5637, opened 2021-02-04
Cleanup: remove obsolete jaxlib version checks

jakevdp:
#5622 bumps the minimum jaxlib version to 0.1.60; this allows us to remove a number of checks for jaxlib versions lower than this.

### DIFF
--- a/jax/lib/__init__.py
+++ b/jax/lib/__init__.py
@@ -17,7 +17,7 @@
 
 __all__ = [
   'cuda_prng', 'cusolver', 'rocsolver', 'jaxlib', 'lapack',
-  'pytree', 'tpu_client', 'version', 'xla_client'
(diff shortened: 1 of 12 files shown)


jax/lib/__init__.py
```diff
@@ -38,20 +38,14 @@
 def _check_jaxlib_version():
   if version < _minimum_jaxlib_version:
     msg = 'jaxlib is version {}, but this version of jax requires version {}.'
-
-    if version == (0, 1, 23):
```
hawkinsp, line 42: This one is not yet obsolete, I think. This is the error you get if you `pip install jax jaxlib` and your `pip` is too old and doesn't know about `manylinux2010`.

We could however fix this by deleting all wheels <= 0.1.23 from pypi.

jakevdp: Good catch - I reverted this part.

hawkinsp (approved):

PR #5641, opened 2021-02-05
[call_tf] Add support for DLPack to avoid copying arrays in eager mode

gnecula:


jax/experimental/jax2tf/call_tf.py
```diff
@@ -147,13 +149,34 @@ def acc_ct(res_, ct_res_):
...
-  return [jnp.asarray(np.asarray(r)) for r in tree_util.tree_flatten(res_tf)[0]]
+
+  def _res_tf_to_jax(res_tf):
+    if isinstance(res_tf, tf.Tensor) and res_tf.dtype in dlpack.SUPPORTED_DTYPES:
+      res_tf_platform = tf.DeviceSpec.from_string(res_tf.backing_device).device_type
+      res_jax_platform = res_tf_platform.lower()
+      if res_jax_platform != "tpu":
+        res_dlpack = tf.experimental.dlpack.to_dlpack(res_tf)
+        return jax.dlpack.from_dlpack(
+            res_dlpack, backend=xla_bridge.get_backend(res_jax_platform))
```
gnecula, line 175: I guess it may be possible here for JAX to be configured with the GPU backend and encounter a Tensor on GPU? Perhaps we should fallback to CPU in that case?

tomhennigan: Do TF and JAX share the same allocator when both are running on the same GPU? I wonder if we can actually safely zero-copy on anything other than CPU?

jax/experimental/jax2tf/call_tf.py
```diff
@@ -147,13 +149,34 @@ def acc_ct(res_, ct_res_):
...
 # The impl will be used in op-by-op mode and calls func_tf in TF eager mode.
-def _call_tf_impl(*args, args_treedef, func_tf, **_):
-  # TODO(necula): does this copy to the host?
-  args_tf_flat = [tf.constant(np.asarray(a)) for a in args]
+def _call_tf_impl(*args_jax_flat, args_treedef, func_tf, **_):
+  def _arg_jax_to_tf(arg_jax):
+    # We try to avoid making a copy of JAX arrays
+    if (isinstance(arg_jax, xla.DeviceArray) and
+        arg_jax.dtype in dlpack.SUPPORTED_DTYPES and
+        arg_jax.device_buffer.client.platform != "tpu"):
```
tomhennigan, line 157: I'd suggest an allow list here, I think most of our backends do not support DLPack (e.g. tpu_driver or some internal only backends).

tomhennigan (approved):
I'm not very familiar with dlpack so it would probably be worth blocking on review from the Peter or Stephan.

jax/experimental/jax2tf/call_tf.py
gnecula: @hawkinsp Do you know if it is safe to use DLPack between TF and JAX on the GPU?

hawkinsp: On CPU, I wouldn't bother with DLPack; I would just share data via a numpy array. This can be done in a zero-copy way on both the TF and JAX sides, I believe.

On GPU, you either need DLPack or __cuda_array_interface__. TensorFlow doesn't support the latter at all. DLPack sharing works fine; we even test it:
https://cs.opensource.google/jax/jax/+/master:tests/array_interoperability_test.py;drc=2fd682ef2ab1a4c4439179e5386ec2a84bffa139;l=96

The problem you will have is that both TF and JAX like to preallocate all of GPU memory. However, that is something that both can be configured not to do, and I'm hopeful that CUDA 11.2's async allocator may be enough to fix the problem for good.

gnecula: I changed the code to only use DLPack on GPU. It was tricky to get TF to avoid copies when exporting numpy, but I managed to do it, and I added a test.

jax/experimental/jax2tf/call_tf.py
gnecula: Good point. It turns out we only use DLPack on GPU now.

PR #5658, opened 2021-02-06
make vmap-of-cond preserve nan safety

mattjj:
We need to apply a "double-where trick" (cf. #1052) when we convert a `cond` to a `select` in the cond vmap rule, since otherwise we change the nan semantics.

(Tangentially, we should add batching in cond itself, but this is a smaller first change!)

Thanks to John Jumper and Jonas Adler for the idea.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -754,47 +754,58 @@ def _select_tree(indices, branch_vals):
   assert len(branch_vals) > 0
   if len(branch_vals) == 1:
     return branch_vals[0]
-  mid = len(branch_vals) // 2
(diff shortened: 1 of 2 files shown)


froystig (approved):

john-m-jumper:
One slightly obscure issue to this (if I am reading it correctly, which I might not be) is that you assume the branch not taken is okay will all zero inputs.  In cases where there is control flow or logging in this branch, that could be a surprising assumption for the user.  It might be slightly safer to use a trick of the form `inputs = select(condition, inputs, stop_gradient(inputs))` to get the same effect.

mattjj:
Great idea, @john-m-jumper ! Indeed I was assuming no side-effects. But this way is better.

PR #5659, opened 2021-02-06
take the docstring and signature of cond from its implementation

froystig:
This better aligns the signature with the current docstring in autogenerated documentation.


mattjj (approved):
Aw yiss.

PR #5664, opened 2021-02-06
Add support for union1d

ashutoshvarma:
This is the implementation for np.union1d from #70 (and #2078 ).

Try to closely follow `np.union1d` implemetation.

Issue #2078, opened 2020-01-26
Implement NumPy set routines


shoyer:
Extracted from #70:

 - [x] np.unique
 - [x] np.ediff1d
 - [x] np.in1d
 - [x] np.intersect1d
 - [x] np.isin
 - [x] np.setdiff1d
 - [x] np.setxor1d
 - [x] np.union1d

Implementation wise, we could use either sorting (like NumPy) or hash-tables (of some sort) for most of these fucntions.

Some of these could be challenging to implement (would not work under `jit`), because the output has a dynamic shape. `in1d` and `isin` don't have dynamic shapes, but that they probably require `searchsorted` from https://github.com/google/jax/issues/2080 first (at least if using sorting).

jakevdp:
#5665 finishes this work!

PR #4635, opened 2020-10-17
[impl] Add support for setdiff1d

aldragan0:
This is the implementation for [np.setdiff1d](https://numpy.org/doc/stable/reference/generated/numpy.setdiff1d.html) from [#70 ](https://github.com/google/jax/issues/70) (and [#2078 ](https://github.com/google/jax/issues/2078)). I was unable to make the assume_unique work due to in1d not supporting assume_unique as well.

PR #5665, opened 2021-02-06
Add support for setxor1d

ashutoshvarma:
This is the implementation for np.setxor1d from #70 (and #2078 ).

Including this and #5664, all the NumPy set routines listed in #2078 are implemented.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1525,6 +1525,30 @@ def union1d(ar1, ar2):
   return unique(conc)
 
 
+@_wraps(np.setxor1d, lax_description="""
+In the JAX version, the input arrays are explicilty flattened regardless
+of assume_unique value.
+""")
+def setxor1d(ar1, ar2, assume_unique=False):
+  ar1 = core.concrete_or_error(asarray, ar1, "The error arose in setxor1d()")
+  ar2 = core.concrete_or_error(asarray, ar2, "The error arose in setxor1d()")
+
+  ar1 = ravel(ar1)
+  ar2 = ravel(ar2)
+
+  if not assume_unique:
+    ar1 = unique(ar1)
+    ar2 = unique(ar2)
+
+  aux = concatenate((ar1, ar2))
+  if aux.size == 0:
+    return aux
+
+  aux = sort(aux)
+  flag = concatenate((array([True]), aux[1:] != aux[:-1], array([True])))
+  return aux[flag[1:] & flag[:-1]]
+
+
 @partial(jit, static_argnums=2)
 def _intersect1d_sorted_mask(ar1, ar2, return_indices=False):
   """
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1180,6 +1180,30 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     self._CheckAgainstNumpy(np_fun, jnp.union1d, args_maker)
 
   @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_{}_{}_assume_unique={}".format(
+       jtu.format_shape_dtype_string(shape1, dtype1),
+       jtu.format_shape_dtype_string(shape2, dtype2),
+       assume_unique),
+       "shape1": shape1, "dtype1": dtype1, "shape2": shape2, "dtype2": dtype2,
+       "assume_unique": assume_unique}
+      for dtype1 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for dtype2 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for shape1 in all_shapes
+      for shape2 in all_shapes
+      for assume_unique in [False, True]))
+  def testSetxor1d(self, shape1, dtype1, shape2, dtype2, assume_unique):
+    rng = jtu.rand_default(self.rng())
+    args_maker = lambda: [rng(shape1, dtype1), rng(shape2, dtype2)]
+    jnp_fun = lambda ar1, ar2: jnp.setxor1d(ar1, ar2, assume_unique=assume_unique)
+    def np_fun(ar1, ar2):
+      if assume_unique:
+        # pre-flatten the arrays to match with jax implementation
+        ar1 = np.ravel(ar1)
+        ar2 = np.ravel(ar2)
+      return np.setxor1d(ar1, ar2, assume_unique)
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_{}_{}_assume_unique={}_return_indices={}".format(
        jtu.format_shape_dtype_string(shape1, dtype1),
        jtu.format_shape_dtype_string(shape2, dtype2),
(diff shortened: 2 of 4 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1525,6 +1525,27 @@ def union1d(ar1, ar2):
   return unique(conc)
 
 
+@_wraps(np.setxor1d)
+def setxor1d(ar1, ar2, assume_unique=False):
+  ar1 = core.concrete_or_error(asarray, ar1, "The error arose in setxor1d()")
+  ar2 = core.concrete_or_error(asarray, ar2, "The error arose in setxor1d()")
+
+  if ar1.ndim != 1 or ar2.ndim != 1:
+    raise ValueError("only 1-dimensional input supported.")
```
jakevdp, line 1534: It looks like numpy's `setxor1d` will ravel inputs:
```python
>>> import numpy as np
>>> np.setxor1d(np.ones((2, 2)), np.zeros((2, 2)))
array([0., 1.])
```
I think we should match those semantics here.

tests/lax_numpy_test.py
```diff
@@ -1179,6 +1179,25 @@ def np_fun(arg1, arg2):
...
+       jtu.format_shape_dtype_string(shape1, dtype1),
+       jtu.format_shape_dtype_string(shape2, dtype2),
+       assume_unique),
+       "shape1": shape1, "dtype1": dtype1, "shape2": shape2, "dtype2": dtype2,
+       "assume_unique": assume_unique}
+      for dtype1 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for dtype2 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for shape1 in one_dim_array_shapes
+      for shape2 in one_dim_array_shapes
+      for assume_unique in [False, True]))
```
jakevdp, line 1193: When `assume_unique` is True, we should make certain the inputs contain unique values, otherwise the output is ill-defined and it could cause non-meaningful test errors if either numpy or jax were to switch to a different implementation.

tests/lax_numpy_test.py
```diff
@@ -1179,6 +1179,25 @@ def np_fun(arg1, arg2):
...
+      {"testcase_name": "_{}_{}_assume_unique={}".format(
+       jtu.format_shape_dtype_string(shape1, dtype1),
+       jtu.format_shape_dtype_string(shape2, dtype2),
+       assume_unique),
+       "shape1": shape1, "dtype1": dtype1, "shape2": shape2, "dtype2": dtype2,
+       "assume_unique": assume_unique}
+      for dtype1 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for dtype2 in [s for s in default_dtypes if s != jnp.bfloat16]
+      for shape1 in one_dim_array_shapes
+      for shape2 in one_dim_array_shapes
```
jakevdp, line 1192: Per the above comment, we should be able to test a wider range of shapes here

jakevdp (changes requested):
Looks good, thanks for tackling this! Just a couple minor changes below

jax/_src/numpy/lax_numpy.py
ashutoshvarma: One thing to note is that `np.setxor1d` does not explicitly ravel the inputs, it just happens as a by-product from call to `np.unique`. 
```python
>>> import numpy as np
>>> np.setxor1d(np.ones((2, 2)), np.zeros((2, 2)))
array([0., 1.])

>>> np.setxor1d(np.ones((2, 2)), np.zeros((2, 2)), assume_unique=True)
<ipython-input-4-1b7ce8d6b603> in <module>
----> 1 np.setxor1d(np.ones((2, 2)), np.zeros((2, 2)), assume_unique=True)

<__array_function__ internals> in setxor1d(*args, **kwargs)

~/miniconda3/envs/jaxenv/lib/python3.8/site-packages/numpy/lib/arraysetops.py in setxor1d(ar1, ar2, assume_unique)
    485 
    486     aux.sort()
--> 487     flag = np.concatenate(([True], aux[1:] != aux[:-1], [True]))
    488     return aux[flag[1:] & flag[:-1]]
    489 

<__array_function__ internals> in concatenate(*args, **kwargs)

ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 1 dimension(s) and the array at index 1 has 2 dimension(s)
```

So should we replicate this as well or explicitly flatten the input arrays?

jakevdp: I'd lean toward explicitly flattening the input arrays: this matches what is done by `np.setdiff1d`, `np.union1d` and `np.intersect1d`.

jakevdp:
Hi - are you interested in continuing to work on this? Thanks!

ashutoshvarma:
Yes, I still am.
Will try to finish it by this week.

ashutoshvarma:
Since we decided to explicitly flatten the input arrays to match `setdiff1d`, `union1d` Jax implementations, I faced problems similar to https://github.com/google/jax/pull/4635#discussion_r509419926 as the behavior of jax `setxor1d` is no longer guaranteed to match `np.setxor1d` (`_CheckAgainstNumpy` assertions fails for few test cases).
So following `setdiff1d`, now `unique()` is called irrespective of `assume_uniqe` value.

jakevdp:
Thanks - I'm not certain I understand the problem re: uniqueness and raveling. What is an example of a case that fails?

ashutoshvarma:
Test cases with - non-1d shapes and `assume_unique=True` fails ( `np.setxor1d` does not support non-1d with assume_unique=True, https://github.com/google/jax/pull/5665#discussion_r571649983)
```
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_float16[3,1]_int16[1,4]_assume_unique=True
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_int16[1,4]_int8[2,3,4]_assume_unique=True
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_int32[4]_int16[3,4]_assume_unique=True
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_int32[4]_int8[1,4]_assume_unique=True
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_int8[4]_int16[1,4]_assume_unique=True
[  FAILED  ] LaxBackedNumpyTests.testSetxor1d_int8[4]_int8[3,1]_assume_unique=True
```

On the second thought, we can just flatten arrays before passing them to `np.setxor1d` when `assume_unique=True` in tests. This way it will match with jax `setxor1d`.

jakevdp:
I see, thanks! Agreed that pre-flattening arrays (with a comment indicating why) is probably the best.

It might also be worth adding a `lax_description` to the `_wraps` to mention this difference in behavior.

tests/lax_numpy_test.py
ashutoshvarma: Regarding this, I thought of using `rand_unique_int` instead of `rand_default` for rng when `assume_unique=True`   https://github.com/google/jax/blob/f63106c8485496750b9663b09813d11269c17c2a/jax/test_util.py#L718-L722
But using `rand_unique_int` will limit the test inputs to `int`s and also it throws exceptions when `shapes` is scalar due to the use of `prod(shapes)` in it. 

PS: Also while we're at it, this might be applicable to `testIntersect1d` as well.
https://github.com/google/jax/blob/f63106c8485496750b9663b09813d11269c17c2a/tests/lax_numpy_test.py#L1206-L1211

ashutoshvarma:
comment and `lax_description` added 👍

jakevdp (approved):
Looks great, thanks for all the followup work!

PR #5672, opened 2021-02-08
First draft of pjit

skye:
Note that this is split up into two commits, one for `pjit` and one for `with_sharding_constraint` (which includes the `pjit` translation rule for testing). It may be easier to review them one at a time.

Overall this works and was super easy thanks to Adam's xmap work! I'd still like to get some more thoughts on the API, i.e. how the positional axis -> mesh axis name mapping should look and what it should be called.

### DIFF
--- a/jax/experimental/pjit.py
+++ b/jax/experimental/pjit.py
@@ -1,0 +1,212 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import OrderedDict
+from typing import (Callable, Optional, Sequence, Tuple, Union)
+from warnings import warn
+
+from . import maps
+from .. import core
+from .. import linear_util as lu
+from ..api import _check_callable, _check_arg
+from ..api_util import (argnums_partial_except, flatten_axes,
+                        flatten_fun_nokwargs, _ensure_index_tuple,
+                        donation_vector, rebase_donate_argnums)
+from ..interpreters import ad
+from ..interpreters import pxla
+from ..interpreters import xla
+from ..lib import xla_bridge as xb
+from ..lib import xla_client as xc
+from ..tree_util import tree_flatten, tree_unflatten
+from .._src.util import (extend_name_stack, HashableFunction, safe_zip,
+                         wrap_name, wraps)
+xops = xc._xla.ops
+
+def pjit(fun: Callable,
+         in_axis_resources,
+         out_axis_resources,
+         static_argnums: Union[int, Sequence[int]] = (),
+         donate_argnums: Union[int, Sequence[int]] = ()):
+  warn("pjit is an experimental feature and probably has bugs!")
+  _check_callable(fun)
+
+  # To be a tree prefix of the positional args tuple, in_axes can never be a
+  # list: if in_axes is not a leaf, it must be a tuple of trees. However,
+  # in cases like these users expect tuples and lists to be treated
+  # essentially interchangeably, so we canonicalize lists to tuples here
+  # rather than raising an error. https://github.com/google/jax/issues/2367
+  if isinstance(in_axis_resources, list):
+    in_axis_resources = tuple(in_axis_resources)
+  if isinstance(out_axis_resources, list):
+    out_axis_resources = tuple(out_axis_resources)
+
+  static_argnums = _ensure_index_tuple(static_argnums)
+  donate_argnums = _ensure_index_tuple(donate_argnums)
+  donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)
+
+  @wraps(fun)
+  def wrapped(*args, **kwargs):
+    if kwargs:
+      raise NotImplementedError("pjit over kwargs not yet supported")
+    if max(static_argnums + donate_argnums, default=-1) >= len(args):
+      raise ValueError(f"jitted function has static_argnums={static_argnums}, "
+                       f"donate_argnums={donate_argnums} but "
+                       f"was called with only {len(args)} positional arguments.")
+
+    # Putting this outside of wrapped would make resources lexically scoped
+    resource_env = maps.thread_resources.env
+
+    f = lu.wrap_init(fun)
+    if static_argnums:
+      f, dyn_args = argnums_partial_except(f, static_argnums, args)
+    else:
+      dyn_args = args
+
+    args_flat, in_tree = tree_flatten(args)
+    for arg in args_flat: _check_arg(arg)
+    flat_fun, out_tree = flatten_fun_nokwargs(f, in_tree)
+    if donate_argnums:
+      donated_invars = donation_vector(donate_argnums, dyn_args, ())
+    else:
+      donated_invars = (False,) * len(args_flat)
+
+    in_axis_resources_flat = tuple(flatten_axes("pjit in_axis_resources",
+                                                in_tree, in_axis_resources))
+    out_axis_resources_thunk = HashableFunction(
+        lambda: tuple(flatten_axes("pjit out_axis_resources", out_tree(),
+                                   out_axis_resources)),
+        closure=out_axis_resources)
+
+    out = pjit_call_p.bind(
+        flat_fun,
+        *args_flat,
+        in_axis_resources=in_axis_resources_flat,
+        out_axis_resources_thunk=out_axis_resources_thunk,
+        resource_env=resource_env,
+        donated_invars=donated_invars,
+        name=flat_fun.__name__)
+    return tree_unflatten(out_tree(), out)
+
+  return wrapped
+
+def _pjit_call_impl(fun: lu.WrappedFun, *args, in_axis_resources,
+                    out_axis_resources_thunk, resource_env, donated_invars,
+                    name):
+  in_avals = [core.raise_to_shaped(core.get_aval(arg)) for arg in args]
+  return _pjit_callable(
+      fun, in_axis_resources, out_axis_resources_thunk, resource_env,
+      donated_invars, name, *in_avals)(*args)
+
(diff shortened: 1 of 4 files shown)


jax/experimental/pjit.py
```diff
@@ -0,0 +1,211 @@
...
+               for out, mesh_axes in safe_zip(out_nodes, out_mesh_axes)]
+
+  subc = subc.build(xops.Tuple(subc, out_nodes))
+  return xops.Call(c, subc, list(in_nodes))
+
+pjit_call_p = core.CallPrimitive("pjit_call")
+pjit_call_p.def_impl(_pjit_call_impl)
+xla.call_translations[pjit_call_p] = _pjit_translation_rule
+
+# None indicates replicated axis (partial replication)
```
apaszke, line 139: I don't think that this is the same thing as partial replication. You have to have at least one `None` in your axis partitioning to even have a chance of triggering partial replication, but there are partitions that do use Nones and are sharded over the whole mesh. What matters is whether you use all mesh axis names or not!

jax/experimental/pjit.py
```diff
@@ -0,0 +1,211 @@
...
+from ..lib import xla_bridge as xb
+from ..lib import xla_client as xc
+from ..tree_util import tree_flatten, tree_unflatten
+from .._src.util import (extend_name_stack, HashableFunction, safe_zip,
+                         wrap_name, wraps)
+xops = xc._xla.ops
+
+# TODO(skye): better name than mesh_axes
+def pjit(fun: Callable,
+         in_mesh_axes,
```
apaszke, line 37: Those are supposed to be like `xmap`'s `in_axes`, but the leaves are `PartitionSpec`s of mesh axis names and `None`s, right?

jax/experimental/pjit.py
```diff
@@ -0,0 +1,211 @@
...
+  if mesh_axes is None:
+    return OrderedDict()
+  return OrderedDict((axis, i)
+                     for i, axis_or_axes in enumerate(mesh_axes)
+                     for (axis, i) in _array_mapping_entries(axis_or_axes, i))
+
+def _array_mapping_entries(partitioning: ArrayAxisPartitioning, i: int):
+  if partitioning is None:
+    return
+  # TODO(skye): do axis names have to be strings?
```
apaszke, line 197: They don't. They can be anything and we only recurse when they're a `list` or a `tuple`.

jax/experimental/pjit.py
```diff
@@ -0,0 +1,211 @@
...
+sharding_constraint_p = core.Primitive("sharding_constraint")
+sharding_constraint_p.def_impl(_sharding_constraint_impl)
+sharding_constraint_p.def_abstract_eval(lambda x, **unused: x)
+ad.deflinear2(sharding_constraint_p,
+              lambda ct, _, mesh_axes, resource_env: (
+                  with_sharding_constraint(ct, mesh_axes),))
+xla.translations[sharding_constraint_p] = _sharding_constraint_translation_rule
+
+
+def get_array_mapping(mesh_axes: ArrayPartitioning) -> pxla.ArrayMapping:
```
apaszke, line 187: This gets called all over the place. Why can't we just use the same API as in xmap and avoid converting back and forth between partition specs and the dicts?

jax/experimental/pjit.py
```diff
@@ -0,0 +1,211 @@
...
+  in_avals = [core.raise_to_shaped(core.get_aval(arg)) for arg in args]
+  return _pjit_callable(
+      fun, in_mesh_axes, out_mesh_axes_thunk, resource_env, name,
+      *in_avals)(*args)
+
+def _pjit_translation_rule(c, axis_env, in_nodes, name_stack, backend, name,
+                           call_jaxpr, in_mesh_axes, out_mesh_axes_thunk,
+                           resource_env):
+  # TODO(skye): this should arguably be part of axis_env
+  mesh = resource_env.physical_mesh
```
apaszke, line 107: I don't think it should. Adding those axes to `axis_env` would mean that the collective lowering rules could pick them up, but `pjit` doesn't introduce any axis names into scope!

apaszke (commented):
Nice!

skye (commented):
Forgot to publish these replies earlier :(

skye:
@apaszke if you're happy with the pxla.py change, I propose we get this in as-is and iterate. I wanna continue iterating on the API and other changes but I think it'll be easier if we can experiment with checked-in code and review smaller chunks.

jax/experimental/pjit.py
apaszke: We currently do have an invariant that you can't switch the `resource_env` on any primitive wrapped in `xmap`/`pjit`, so I think it's fine. I'd just rather keep those mesh axes out of `axis_env`, because if we ever end up accidentally calling into e.g. a `psum` rule then it might succeed for no good reason instead of throwing an unbound axis error.

apaszke (approved):

jax/experimental/pjit.py
jekbradbury: I would maybe call these arguments `in_axis_resources` and `out_axis_resources` for max consistency with `xmap`?

jekbradbury (approved):

jax/experimental/pjit.py
skye: Yeah ok. I actually find the term `axis_resources` kinda confusing, but I'll use it here for consistency and maybe we can reconsider it for both `xmap` and `pjit` at some point.

[bikeshedding on naming below]
After studying the xmap docstring, I think the "axis" in `axis_resources` refers to the named array axes? Elsewhere in the docstring, the mesh axes are called "resource axes". So I think `axis_resources` is short for `axis_resource_axes`? If so, IMO renaming to `resource_axes` is slightly clearer.

jax/experimental/pjit.py
```diff
@@ -0,0 +1,210 @@
...
+    out_mesh_axes = tuple(out_mesh_axes)
+
+  static_argnums = _ensure_index_tuple(static_argnums)
+
+  @wraps(fun)
+  def wrapped(*args, **kwargs):
+    if kwargs:
+      raise NotImplementedError("pjit over kwargs not yet supported")
+
+    # Putting this outside of fun_mapped would make resources lexically scoped
```
apaszke, line 61: nit: there's no `fun_mapped` here

skye: Oops, changed to `wrapped`

jax/experimental/pjit.py
skye: After talking to Adam, we concluded it should be pretty easy to use the xmap API and get the best of all worlds. I'm gonna do this in a follow-up to get this in while I figure out how xmap does it.

tests/pjit_test.py
```diff
@@ -0,0 +1,173 @@
...
+    x = jnp.arange(np.prod(x_shape)).reshape(x_shape)
+    y = jnp.arange(np.prod(y_shape)).reshape(y_shape)
+    actual = f(x, y)
+    expected = x @ y
+    self.assertAllClose(actual, expected, check_dtypes=False)
+    self.assertIsInstance(actual, pxla.ShardedDeviceArray)
+    self.assertLen(actual.device_buffers, 4)
+
+    split0, split1 = np.split(expected, 2)
+    self.assertAllClose(actual.device_buffers[0].to_py(), split0,
```
gnecula, line 92: I am wondering if it is specified that the device_buffers will be assigned as tested here?

tests/pjit_test.py
```diff
@@ -0,0 +1,173 @@
...
+    self.assertIsInstance(actual, pxla.ShardedDeviceArray)
+    self.assertLen(actual.device_buffers, 2)
+    self.assertAllClose(actual.device_buffers[0].to_py(), expected,
+                        check_dtypes=False)
+
+  @with_mesh([('x', 2), ('y', 2)])
+  def testBasic2D(self):
+    @partial(pjit,
+             in_axis_resources=(P(None, 'x', 'y'), P('y')),
+             out_axis_resources=P('x'))
```
gnecula, line 77: Is this an example of the partial replication feature, where the output is sharded on the x resource axis and replicated on the y?

jax/experimental/pjit.py
```diff
@@ -0,0 +1,212 @@
...
+  return xb.set_sharding_proto(c, x_node,
+                               get_sharding_proto(c, x_node, axis_resources, mesh))
+
+sharding_constraint_p = core.Primitive("sharding_constraint")
+sharding_constraint_p.def_impl(_sharding_constraint_impl)
+sharding_constraint_p.def_abstract_eval(lambda x, **unused: x)
+ad.deflinear2(sharding_constraint_p,
+              lambda ct, _, axis_resources, resource_env: (
+                  with_sharding_constraint(ct, axis_resources),))
+xla.translations[sharding_constraint_p] = _sharding_constraint_translation_rule
```
gnecula, line 186: Does it make sense to add this only to the "tpu" translation rules? Or, throw an error from the translation rule for non-TPU

jax/experimental/pjit.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def with_sharding_constraint(x, axis_resources):
+  resource_env = maps.thread_resources.env
+  return sharding_constraint_p.bind(x, axis_resources=axis_resources,
+                                    resource_env=resource_env)
+
+def _sharding_constraint_impl(x, axis_resources, resource_env):
+  # TODO(skye): can we also prevent this from being called in other
+  # non-pjit contexts? (e.g. pmap, control flow)
```
gnecula, line 171: Is pjit going to work eventually inside pmap, or xmap?

tests/pjit_test.py
skye: Yah, so easy with the mesh API :)

tests/pjit_test.py
skye: I don't think this is part of the public API, no. A less implementation-specific way to do this might be to look at each device_buffer's device to determine which split it should have, although that might make the test hard to reason about. I will probably focus on other things first before revisiting this test, but something to keep in mind if/when we do revisit it (e.g. if it breaks due to this detail :wink: ).

jax/experimental/pjit.py
skye: Good question! I thought so, but after our discussion with James and Anselm, maybe pjit-of-vmap is enough. We should think more about this.

jax/experimental/pjit.py
skye: Agh, I realize I forgot to copy the TPU check from `sharded_jit`'s impl to `pjit`'s impl. I think a top-level check will be better than a TPU-only translation rule, since this rule will theoretically work if/when XLA supports GPU sharding. I'll add  the impl check next time I touch this code.

PR #5673, opened 2021-02-08
Add a pgather primitive, making it possible to index into mapped axes

apaszke:
cc @jekbradbury

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -18,6 +18,7 @@ Parallelization primitives.
 import collections
 import string
 import warnings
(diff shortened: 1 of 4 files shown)


mattjj:
The failure looks like a trivial jax2tf issue, where we need to add `pgather` to [this list](https://github.com/google/jax/blob/3df58b9729e3991b64cecbd25c48f101ae92e99b/jax/experimental/jax2tf/jax2tf.py#L832).

mattjj (approved):
(I haven't thought about a good systematic test strategy yet!)

PR #5674, opened 2021-02-08
Adds missing reduction in xmap() pydoc example.

andsteing:


jax/experimental/maps.py
```diff
@@ -385,7 +385,8 @@ def regression_loss(x, y, w, b):
       # Contract over in_features. Batch and out_features are present in
       # both inputs and output, so they don't need to be mentioned
       y_pred = jnp.einsum('{in_features},{in_features}->{}') + b
```
mattjj, line 387:
```suggestion
      y_pred = jnp.einsum('{in_features},{in_features}->{}', w, x) + b
```

This is a pre-existing bug but I just noticed it.

We could probably do the reduce-sum as part of the einsum, like

```python
      y_pred = jnp.einsum('{out_features,in_features},{in_features}->{}', w, x) + b
```

Not sure which seems nicer. Maybe the version in this PR with the explicit `jnp.sum` reduction.

WDYT @apaszke @andsteing ?

andsteing: Adding both in the `einsum` LGTM.
(I ran it locally using `pdot` because `einsum` was not available yet)

andsteing: Oh, wait - we would still have the `out_features` dimension in `y`, though, so I think it's better to keep the two separate reductions.

mattjj: I think the `out_features` axis would be summed over. Here's a positional axis analogy:

```python
In [1]: import numpy as np

In [2]: w = np.random.randn(4, 3)

In [3]: x = np.random.randn(3)

In [4]: np.einsum('oi,i->', w, x).shape
Out[4]: ()
```

or, with a batch axis:

```python
In [1]: import numpy as np

In [2]: w = np.random.randn(4, 3)

In [3]: x = np.random.randn(10, 3)

In [4]: np.einsum('oi,bi->b', w, x).shape
Out[4]: (10,)
```

That is, by analogy only the `w` input has the `out_features` axis, and we're summing it out.

(Unfortunately we're still landing the change where JAX will track axis names for us, so I can't easily show this working on named axes. And our named-axis einsum implementation may have bugs too!)

andsteing: But then in the next step we would have `loss = (y - y_pred)**2` and `y` has the named axis `batch` *and* `out_features` - so if I understand `xmap()` correctly this would result in `y_pred` being broadcast on `out_features` again and then an error when `jnp.sum(loss, axis="batch")` is returned because the return value would still have the named axis `out_features`.

(BTW : looking forward to have named axis in avals !)

mattjj: Oh, of course you're right, sorry! Thanks for explaining. I'm doing too much context switching :)

apaszke (approved):

andsteing:
This was submitted as #5688 - discarding this PR.

PR #5682, opened 2021-02-09
add pargmax, pargmin wrappers

mattjj:
We didn't add these to the `lax` namespace because they may end up being redundant with `lax.argmin` / `lax.argmax` if/when those functions (and their underlying primitives) grow named-axis support. (On the other hand, it might be useful to have a named-axis-only API call!)

#5694 is a follow-up that introduces primitives for these for vectorization efficiency.

The XLA:CPU bug described on #5694 is also present here.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -180,6 +180,23 @@ def pmin(x, axis_name, *, axis_index_groups=None):
                          axis_index_groups=axis_index_groups)
   return tree_util.tree_unflatten(treedef, out_flat)
 
+# TODO(mattjj): add a pargmin_p, or add named axis support to lax.argmin_p
+def pargmin(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmin only accepts a single axis, got {axis_name}")
+  return _axis_index_of_val(x, pmin(x, axis_name), axis_name)
+
+# TODO(mattjj): add a pargmax_p, or add named axis support to lax.argmax_p
+def pargmax(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmin only accepts a single axis, got {axis_name}")
+  return _axis_index_of_val(x, pmax(x, axis_name), axis_name)
+
+def _axis_index_of_val(x, val, axis_name):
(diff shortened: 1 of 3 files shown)


docs/jax.lax.rst
```diff
@@ -175,6 +175,8 @@ Parallelism support is experimental.
     psum
     pmax
     pmin
+    pargmax
+    pargmin
```
apaszke, line 179: I think we should stop exposing all those new functions. Let's keep them in `lax.parallel` for the time being, but ultimate `jnp` should be the interface to them.

apaszke (approved):
LGTM, but I would consider not making the new functions public and adding a TODO for making this fast in the vmap case (then it should just become a single variadic reduce).

docs/jax.lax.rst
mattjj: Well, our typical pattern is for `lax` to expose all the basic functionality (like primitives and small canonical wrappers) and then use `jnp` to set up a familiar NumPy API and add conveniences like shape and dtype promotion. That is, we always have redundancy between the `lax` API and the `jnp` API. It's often useful because `lax` is more strict (i.e. `lax` is less lax!).

I agree that this case is less clear because we could say that `lax.pmin`/`lax.pmax` constitute the basic functionality and `pargmin`/`pargmax` are just convenience wrappers. But perhaps we will want to make these into primitives for performance reasons.

To me, the main question is about redundancy _within_ `lax`, rather than between `jnp` and `lax`. That is, should we have just one `lax.argmin` which accepts either a positional or named axis? Probably so. In fact, I think I should revise this PR to do that now rather than wait until later. That'll resolve the API question too, in that we won't be adding a new API function to `lax` at all.

mattjj: To summarize: you're right, let's not add `pargmin`/`pargmax`, though I think the reason is to avoid redundancy in the `lax` API (rather than being about `jnp`).

mattjj: I started work on a primitive version, but I might get this PR merged first just to avoid blocking anything.

In the meantime, I removed `pargmax`/`pargmin` from the `lax` namespace and deleted their docstrings, since it seems likely named axis support will arise in `lax.argmax`/`lax.argmin` pretty soon.

Issue #5685, opened 2021-02-09
map_coordinates should fail with one dimensional coordinates


agudallago:
jax.scipy.ndimage.map_coordinates behaves differently than scipy.ndimage.map_coordinates:

```python
src = jnp.array([5, 1, 6])  # Fails!
# src = jnp.array([5, 1, 5]) # Doesn't fail

# This fails cause SCIPY map_coordinates doesn't support a list as args. Needs a rank 2 input.
coords = [2]      # Fails on scipy! Should also fail on JAX
#coords = [[2]] # Works!

def scipy_map_coordinates():
  return ndimage.map_coordinates(src, coords, order=1, mode="constant")

def jax_map_coordinates():
  return jax.scipy.ndimage.map_coordinates(
      src, coords, order=1, mode="constant")

print(jax_map_coordinates())
print(scipy_map_coordinates())

assert jnp.array_equal(scipy_map_coordinates(), jax_map_coordinates())
```

Jax's map_coordinates should fail in the same way the scipy version does if we want to keep the API as similar as possible. 

CC: @claudiofantacci

shoyer:
To me, this looks like a SciPy bug. Logically, the behavior of `map_coordinates` is perfectly well defined for one dimensional coordinates -- returning a scalar, just like indexing -- and indeed that is exactly what JAX does, without any special case logic.

I'm not sure why it is important to replicate this behavior _exactly_ in JAX. We do try to make JAX drop-in compatible with NumPy/SciPy, but we also add additional functionality in some cases where NumPy/SciPy would raise an error (e.g., allowing for new dtypes and new optional arguments).

agudallago:
I agree with you in this one. I don't know how is the process here, should we wait for somebody else to pitch in and give more opinions, or should we close this?

shoyer:
If we're agreed, I think we can close this. Thanks for bringing up the issue!

PR #5686, opened 2021-02-09
Make scipy.ndimage.map_coordinates fail when coordinates is rank 1

agudallago:
This will fix the just raised issue:
https://github.com/google/jax/issues/5685


zhangqiaorjc:
Based on https://github.com/google/jax/issues/5685#issuecomment-776883676, let's close this for now.

PR #5688, opened 2021-02-09
Minor fixes for xmap docstring, xeinsum parser

apaszke:
The regression loss example from the xmap docstring was broken and
the xeinsum parser didn't accept empty parens while it should.

### DIFF
--- a/jax/experimental/maps.py
+++ b/jax/experimental/maps.py
@@ -367,15 +367,15 @@ def xmap(fun: Callable,
   while named axes are just a convenient way to achieve batching. While this
   might seem like a silly example at first, it might turn out to be useful in
   practice, since with conjuction with ``axis_resources`` this makes it possible
-  to implement a distributed matrix-multiplication in just a few lines of code:
+  to implement a distributed matrix-multiplication in just a few lines of code::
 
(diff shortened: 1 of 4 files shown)


mattjj (approved):
LGTM modulo one docstring fix.

jax/experimental/maps.py
```diff
@@ -384,8 +384,8 @@ def xmap(fun: Callable,
     def regression_loss(x, y, w, b):
       # Contract over in_features. Batch and out_features are present in
       # both inputs and output, so they don't need to be mentioned
-      y_pred = jnp.einsum('{in_features},{in_features}->{}') + b
-      return jnp.mean((y - y_pred) ** 2, axis='batch')
+      y_pred = jnp.einsum('{in_features},{in_features}->{}', x, w) + b
+      return jnp.mean((y - y_pred) ** 2, axis=('batch', 'out_features'))
```
mattjj, line 388:
```suggestion
      error = jnp.sum((y - y_pred) ** 2, axis='out_features')
      return jnp.mean(error, axis='batch')
```

I think we want a sum to compute the l2 norm, then a mean over the batch, right?

apaszke: Right, of course! 🤦‍♂️

PR #5690, opened 2021-02-09
make jax.eval_shape duck-typing more robust

### LINKED ISSUE #5683: jax.eval_shape does not understand arbitrary object dtype
```python
import objax
import jax
m = objax.nn.Sequential([objax.nn.Linear(3, 4)])

v = objax.random.uniform(((32, 3)))
jax.eval_shape(m, v)  # ShapeDtypeStruct(shape=(32, 4), dtype=float32)

d = objax.util.EasyDict(shape=(32, 3), dtype=v.dtype)
print(repr(d.shape), repr(d.dtype)) # (32, 3) dtype('float32')
print(hasattr(d, 'shape'), hasattr(d, 'dtype')) # True True

jax.eval_shape(m, d)
# 
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3417, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-24-d48e201250f2>", line 1, in <module>
    jax.eval_shape(m, d)
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/jax/api.py", line 2283, in eval_shape
    *map(abstractify, args_flat))
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/jax/util.py", line 36, in safe_map
    return list(map(f, *args))
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/jax/api.py", line 2279, in abstractify
    return ShapedArray(np.shape(x), dtypes.result_type(x))
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/jax/dtypes.py", line 294, in result_type
    return canonicalize_dtype(dtype(args[0]))
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/jax/dtypes.py", line 288, in dtype
    return np.result_type(x)
  File "<__array_function__ internals>", line 5, in result_type
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/numpy/core/_internal.py", line 64, in _usefields
    names, formats, offsets, titles = _makenames_list(adict, align)
  File "/usr/local/google/home/dberth/jax3/lib/python3.8/site-packages/numpy/core/_internal.py", line 40, in _makenames_list
    format = dtype(obj[0], align=align)
TypeError: data type not understood
```

The class EasyDict is simply this:
```python
class EasyDict(dict):
    """Custom dictionary that allows to access dict values as attributes."""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.__dict__ = self
```

mattjj:
fixes #5683

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -2318,8 +2318,14 @@ def eval_shape(fun: Callable, *args, **kwargs):
   >>> print(out.dtype)
   float32
   """
+  def dtype(x):
+    try:
+      return dtypes.result_type(x)
+    except ValueError:
+      return dtypes.result_type(getattr(x, 'dtype'))
+
   def abstractify(x):
-    return ShapedArray(np.shape(x), dtypes.result_type(x))
+    return ShapedArray(np.shape(x), dtype(x))
   args_flat, in_tree = tree_flatten((args, kwargs))
   wrapped_fun, out_tree = flatten_fun(lu.wrap_init(fun), in_tree)
   out = pe.abstract_eval_fun(wrapped_fun.call_wrapped,
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -324,7 +324,9 @@ def dtype(x):
 def result_type(*args):
   """Convenience function to apply Numpy argument dtype promotion."""
    # TODO(jakevdp): propagate weak_type to the result.
-  if len(args) < 2:
+  if len(args) == 0:
+    raise ValueError("at least one array or dtype is required")
+  if len(args) == 1:
     return canonicalize_dtype(dtype(args[0]))
   # TODO(jakevdp): propagate weak_type to the result when necessary.
   return canonicalize_dtype(_least_upper_bound(*{_jax_type(arg) for arg in args}))
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -1222,6 +1222,17 @@ class APITest(jtu.JaxTestCase):
 
     self.assertEqual(out_shape.shape, (3, 5))
 
+  def test_eval_shape_duck_typing2(self):
+    # https://github.com/google/jax/issues/5683
+    class EasyDict(dict):
+      def __init__(self, *args, **kwargs):
+          super().__init__(*args, **kwargs)
+          self.__dict__ = self
+
+    x = EasyDict(shape=(3,), dtype=np.dtype('float32'))
+    out_shape = api.eval_shape(lambda x: x, x)  # doesn't crash
+    self.assertEqual(out_shape.shape, (3,))
+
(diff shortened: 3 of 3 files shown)


jax/api.py
```diff
@@ -2318,8 +2318,14 @@ def __init__(self, shape, dtype):
   >>> print(out.dtype)
   float32
   """
+  def dtype(x):
+    try:
+      return dtypes.result_type(x)
+    except ValueError:
+      return dtypes.result_type(getattr(x, 'dtype'))
```
jakevdp, line 2325: I'm not sure this is the place for this kind of logic. JAX already handles some of these cases by delegating to `np.result_type`, which checks for the `dtype` attribute:
```python
import numpy as np
import jax.numpy as jnp

class MyArray:
  def __init__(self, dtype=np.int32):
    self.dtype = dtype

print(jnp.result_type(MyArray()))
# int32
```
This doesn't work for `EasyDict` because `numpy.result_type` fails to recognize it.

I think we should try to understand why numpy fails this test before working around it in the jax wrapper. And if the workaround is warranted, IMO it would be better to do it at the location that we fall-back to `np.result_type`, which I believe is here: https://github.com/google/jax/blob/c2455d97e95f5c3c31a96571a1236b1f786ad93d/jax/dtypes.py#L319-L322

jakevdp: ...to fully support this we'd probably also have to add similar logic to `jnp.dtype`, which currently is just an alias to `np.dtype`, which fails in the same way.

mattjj: Actually, I already tried something like that: the earlier commit d99cba put this logic in dtypes.py, but that led to failures in existing tests for subtle reasons.

The reason the logic makes sense here is that `jax.eval_shape` explicitly says in its docstring that it supports its inputs being duck-typed this way:

> Since only the ``shape`` and ``dtype`` attributes are accessed, only values that duck-type arrays are required, rather than real ndarrays.

Other parts of our API, like `jnp.dtype`, don't make that promise, so we don't actually need to generalize the functions in dtypes.py. (If anything, it'd be surprising to make `jnp` API functions differ from their `np` cousins in this way!)

mattjj: In other words, the goal of this PR is not to change `jnp.result_type` or the logic in dtypes.py at all. We just want to fix `jax.eval_shape`.

jakevdp: OK, that makes sense!

jakevdp (approved):

PR #5691, opened 2021-02-09
DOC: Add automatically-synced markdown versions of notebooks.

jakevdp:
This PR uses [jupytext](https://jupytext.readthedocs.io/) to create parallel sources for our notebook documentation:

- the existing ipynb files in `docs/notebooks/` are largely unchanged apart from standardizing the JSON format in the files
- myst markdown files are in the same directory with the equivalent content

If a source of either format is updated, the companion can be synced via `jupytext` or `pre-commit`, as outlined in the updated developer documentation. I also added a new github action that fails if any of the sources are out of sync (see the failing CI result on the second commit).

For the time being, the sphinx documentation build is still done via nbsphinx, but this will make it easier to transition to a fully text-based source, as is being explored in #5537

The main benefit of this change is that the markdown versions of these documentations interact more seamlessly with version control, and can be more easily edited using a text editor.


mattjj (approved):
This seems great to me! I'm really impressed with how carefully you set this up, covering all angles and making the process easy.

gnecula:
Very nice, I am looking forward to trying this. Thanks!

jakevdp:
Attempting to fix the import errors in #5697

PR #5694, opened 2021-02-10
add pargmin_p / pargmax_p primitives

mattjj:
Follow up from #5682, which added these functions without defining new primitives.

By defining new primitives, we make `vmap`-of-`pargmin` more efficient, avoiding the "vectorized efficiency problem", which is the same reason why we're defining `psum`, `pdot`, etc. as primitives that support both named and positional axis arguments.

Because these primitives support both named and positional arguments, they can replace `argmin_p` and `argmax_p`, though that's left as follow-up work.

---

There's some weird failure with 2D float32 pmap test cases on CPU. It seems to be CPU-specific because the failures don't happen on TPU. The same issue does not arise on CPU in the int32 test cases with the same shape. (The failures are **not** due to the changes in `_axis_index_bind` or `_axis_index_translation_rule`; I made those changes to simplify the XLA HLO computation I was looking at while trying to debug this issue.)

An example of this failure is `PmapTest.testArgAllReduce_shape=float32[4,2]_axis=1_collective=pargmax`. The expected output is `[0, 1, 1, 0]` but the actual output is `[         0, 2147483647, 2147483647,          0]`. The HLO looks correct though:

```
HloModule pmap__lambda_.26

primitive_computation_max.9 {
  parameter.10 = f32[] parameter(0)
  parameter.11 = f32[] parameter(1)
  ROOT maximum.12 = f32[] maximum(parameter.10, parameter.11)
}

primitive_computation_min.18 {
  parameter.19 = s32[] parameter(0)
  parameter.20 = s32[] parameter(1)
  ROOT minimum.21 = s32[] minimum(parameter.19, parameter.20)
}

ENTRY pmap__lambda_.26 {
  constant.3 = pred[] constant(false)
  parameter.1 = f32[4]{0} parameter(0), parameter_replication={false}
  all-reduce.13 = f32[4]{0} all-reduce(parameter.1), replica_groups={{0,1}}, to_apply=primitive_computation_max.9
  tuple.14 = (f32[4]{0}) tuple(all-reduce.13)
  get-tuple-element.15 = f32[4]{0} get-tuple-element(tuple.14), index=0
  compare.16 = pred[4]{0} compare(get-tuple-element.15, parameter.1), direction=EQ
  replica-id.4 = u32[] replica-id()
  convert.5 = s32[] convert(replica-id.4)
  broadcast.6 = s32[4]{0} broadcast(convert.5), dimensions={}
  constant.7 = s32[] constant(2147483647)
  broadcast.8 = s32[4]{0} broadcast(constant.7), dimensions={}
  select.17 = s32[4]{0} select(compare.16, broadcast.6, broadcast.8)
  all-reduce.22 = s32[4]{0} all-reduce(select.17), replica_groups={{0,1}}, to_apply=primitive_computation_min.18
  tuple.23 = (s32[4]{0}) tuple(all-reduce.22)
  get-tuple-element.24 = s32[4]{0} get-tuple-element(tuple.23), index=0
  ROOT tuple.25 = (s32[4]{0}) tuple(get-tuple-element.24)
}
```

I think I may just file an XLA:CPU bug about this...

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -180,6 +180,16 @@ def pmin(x, axis_name, *, axis_index_groups=None):
                          axis_index_groups=axis_index_groups)
   return tree_util.tree_unflatten(treedef, out_flat)
 
+def pargmax(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmax only accepts a single axis, got {axis_name}")
+  return pargmax_p.bind(x, axis=axis_name)
+
+def pargmin(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmin only accepts a single axis, got {axis_name}")
+  return pargmin_p.bind(x, axis=axis_name)
+
 def _validate_axis_index_groups(axis_index_groups):
   if axis_index_groups is None:
     return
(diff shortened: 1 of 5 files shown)


jax/_src/lax/parallel.py
```diff
@@ -730,6 +740,82 @@ def pos_reduce(x):
...
+
+def _pargminmax_abstract_eval(x, *, axis):
+  if isinstance(axis, int):
+    if not 0 <= axis < np.ndim(x):
+      raise ValueError(f"axis index {axis} for operand shape {np.shape(x)}")
+    shape = tuple(np.delete(np.shape(x), axis))
+  else:
+    # TODO(mattjj): avals with names, check and eliminate axis name
+    shape = np.shape(x)
+  return ShapedArray(shape, np.dtype('int32'))
```
apaszke, line 761: I thought that `argmin` and `argmax` always return `int64`? Or they do at least when used from the NumPy API. The primitives themselves are parameterized by the integer type.

jax/_src/lax/parallel.py
```diff
@@ -730,6 +740,82 @@ def pos_reduce(x):
...
+pargmax_p.def_abstract_eval(_pargminmax_abstract_eval)
+pargmin_p.def_abstract_eval(_pargminmax_abstract_eval)
+
+def _pargminmax_vmap_collective_rule(reducer, frame, vals_in, dims_in, *, axis):
+  del frame, axis  # Unused.
+  (x,), (d,) = vals_in, dims_in
+  return reducer(x, d, np.dtype('int32')), None
```
apaszke, line 768: I think that you don't have a guarantee that `d` is not `batching.not_mapped` here, so we might have to handle that case explicitly.

jax/_src/lax/parallel.py
```diff
@@ -730,6 +740,82 @@ def pos_reduce(x):
...
+batching.collective_rules[pargmin_p] = partial(_pargminmax_vmap_collective_rule,
+                                               lax.argmin)
+
+def _pargminmax_vmap_batching_rule(prim, vals_in, dims_in, *, axis):
+  (x,), (d,) = vals_in, dims_in
+  if isinstance(axis, int):
+    new_axis = axis + (d <= axis)
+    return prim.bind(x, axis=new_axis)
+  else:
+    return prim.bind(x, axis)
```
apaszke, line 780: Doesn't this also have to return a second component for the batch axis in the output? How come aren't the tests catching this?

jax/_src/lax/parallel.py
```diff
@@ -730,6 +740,82 @@ def pos_reduce(x):
...
+
+def _pargminmax_axis_subst_rule(params: core.ParamDict, subst: core.AxisSubst
+                             ) -> core.ParamDict:
+  axis = params['axis']
+  if isinstance(axis, int):
+    return params
+  else:
+    return dict(params, axis=subst(axis))
+core.axis_substitution_rules[pargmax_p] = _pargminmax_axis_subst_rule
+core.axis_substitution_rules[pargmin_p] = _pargminmax_axis_subst_rule
```
apaszke, line 794: You can just use `partial(_subst_all_names_in_param, 'axis')` here

jax/_src/lax/parallel.py
```diff
@@ -180,6 +180,16 @@ def pmin(x, axis_name, *, axis_index_groups=None):
...
   return tree_util.tree_unflatten(treedef, out_flat)
 
+def pargmax(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmax only accepts a single axis, got {axis_name}")
+  return pargmax_p.bind(x, axis=axis_name)
+
+def pargmin(x, axis_name):
+  if isinstance(axis_name, (tuple, list)):
+    raise TypeError(f"pargmin only accepts a single axis, got {axis_name}")
```
apaszke, line 190: While we can definitely get away with primitives not supporting multiple axes in the user-facing API, we definitely cannot do that internally. `xmap` usually replaces each user-facing axis with multiple internal axes.

jax/_src/lax/parallel.py
apaszke: Also, your current implementation is ill-typed, because `subst(axis)` might return a tuple of axes, which will break all other rules. We need those primitives to have multi-axis support.

gnecula:
To fix the jax2tf/primitives_test failures, you need to add the new primitives [here](https://github.com/google/jax/blob/c2455d97e95f5c3c31a96571a1236b1f786ad93d/jax/experimental/jax2tf/jax2tf.py#L841)

jax/_src/lax/parallel.py
mattjj: `np.argmin` / `np.argmax` don't support tuples of axes, so I don't think these primitives should either. Why do you think we need these primitives to support tuples of axes? The issue might be in the substitution code if it requires that.

mattjj: Oh I see, for the 'soft_pmap case'. I'm not sure what should be returned as an argmin over two axes though... it could be the flattened index, or it could be a tuple of indices.

jax/_src/lax/parallel.py
mattjj: I think that should be guaranteed by [this code](https://github.com/google/jax/blob/5c14dafefef1ea891bab4fafd2d8d61afa927e8e/jax/interpreters/batching.py#L153-L154).

jax/_src/lax/parallel.py
mattjj: Indeed we'd want that for the `jnp` API, though we don't necessarily need it on the lax API here.

`lax.argmin` and `lax.argmax` are parameterized by the dtype, but because our `lax.axis_index` implementation only returns 32bit values (in turn because XLA's `ReplicaId` produces a `uint32` value) 64bit support here won't be complete. But it's easy to add!

jax/_src/lax/parallel.py
mattjj: Good catch, I forgot to add a test for this case! I should add an xmap test that uses systematic schedules.

jax/_src/lax/parallel.py
mattjj: Well, we could if we change the `xmap` lowering strategy not to use the split-axis approach :P

But point taken, I forgot about this constraint.

jakevdp:
Closing as obsolete.

PR #5701, opened 2021-02-10
Move flake8 & mypy checks to pre-commit framework

jakevdp:
This serves several purposes:

- combines two existing github actions into one
- pins `mypy` & `flake8` to specified versions to prevent errors being introduced by external releases
- when run locally via `pre-commit run --all`, flake8 and mypy checks will have identical invocations as when run by github actions CI
- optionally allows setting flake8 & mypy to be run automatically as a git precommit hook on your local clone via the [pre-commit](https://pre-commit.com/) framework (just run `pre-commit install`)

Note that I kept the workflow name `lint_and_typecheck` so that no changes are necessary for copybara CI validation.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -11,49 +11,15 @@ on:
       - master
 
 jobs:
-  pre-commit:
+  lint_and_typecheck:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


froystig (approved):
Nice!

PR #5702, opened 2021-02-10
[avals with names] abstract evaluation with named shapes, including rules for standard primitives and a few custom primitives

mattjj:
This is just one commit past #5626

cc @froystig @jekbradbury

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -602,6 +602,7 @@ def _allreduce_impl(pos_reducer, *args, axes, axis_index_groups):
   return [pos_reducer(arg, axes) for arg in args]
 
 def _allreduce_abstract_eval(*args, axes, axis_index_groups):
+  # TODO(frostig,mattjj,jekbradbury): maybe check aval names here
   pos_axes = tuple(axis for axis in axes if isinstance(axis, int))
   named_shapes = [arg.named_shape for arg in args]
   if axis_index_groups is None:
@@ -1142,6 +1143,7 @@ def _pdot_impl(x, y, *, axis_name, pos_contract, pos_batch):
 
 @pdot_p.def_abstract_eval
 def _pdot_abstract_eval(x, y, *, axis_name, pos_contract, pos_batch):
+  # TODO(frostig,mattjj,jekbradbury): check inputs have given axis names?
   if not len(set(axis_name)) == len(axis_name): raise ValueError
   pos_aval = lax.dot_general_p.abstract_eval(
       x, y, dimension_numbers=[pos_contract, pos_batch],
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -603,9 +603,16 @@ def _allreduce_impl(pos_reducer, *args, axes, axis_index_groups):
 
 def _allreduce_abstract_eval(*args, axes, axis_index_groups):
   pos_axes = tuple(axis for axis in axes if isinstance(axis, int))
+  named_shapes = [arg.named_shape for arg in args]
+  if axis_index_groups is None:
+    named_axes = set(axis for axis in axes if not isinstance(axis, int))
+    named_shapes = [{name: size for name, size in arg.named_shape.items()
+                     if name not in named_axes} for arg in args]
+  else:
+    assert len(pos_axes) == 0
   return [ShapedArray(lax._reduce_op_shape_rule(raise_to_shaped(arg), axes=pos_axes),
-                      arg.dtype)
-          for arg in args]
+                      arg.dtype, named_shape=named_shape)
+          for arg, named_shape in zip(args, named_shapes)]
 
 def _allreduce_translation_rule(prim, pos_prim, c, *args, axes, axis_index_groups,
                                 axis_env, platform):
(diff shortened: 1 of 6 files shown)


jekbradbury (approved):

jekbradbury:
So the status here is that control flow still wouldn't work, right? Maybe we should even add a broken test to that effect.

mattjj:
Yes. We don't think it's any less functional than in #5626, but indeed it's still a TODO to add tests and to work through trickier cases like control flow.

Maybe we should land it in pieces, as you say.

froystig:
That's right. This replaces the standard primitive named shape rule in #5626 (based on partially evaluating today's batching rules) with a directly defined one. It also adds a custom rule for `reduce`.

jax/_src/lax/parallel.py
```diff
@@ -586,9 +586,16 @@ def _allreduce_impl(pos_reducer, *args, axes, axis_index_groups):
 
 def _allreduce_abstract_eval(*args, axes, axis_index_groups):
   pos_axes = tuple(axis for axis in axes if isinstance(axis, int))
+  named_shapes = [arg.named_shape for arg in args]
+  if axis_index_groups is None:
+    named_axes = set(axis for axis in axes if not isinstance(axis, int))
+    named_shapes = [{name: size for name, size in arg.named_shape.items()
+                     if name not in named_axes} for arg in args]
```
mattjj, line 593: It seems like we should check here that the `named_axes` over which we're reducing are indeed all present in the input. (I'm assuming `psum(1, 'i')` in the "surface language" would generate a `pbroadcast`, maybe via the traceable. An alternative would be to say `psum [ axis_name='i' ] x:int32[]` is well-typed in jaxprs but IIRC we didn't want to go that route. I'm thinking of the abstract eval rule as our typechecking rule, so this is our chance to catch any type errors!)

jax/_src/lax/parallel.py
```diff
@@ -1014,7 +1021,9 @@ def _all_gather_abstract_eval(x, *, all_gather_dimension, axis_name, axis_index_
   x_aval = raise_to_shaped(x)
   new_shape = list(x_aval.shape)
   new_shape.insert(all_gather_dimension, axis_size)
-  return x_aval.update(shape=new_shape)
+  new_named_shape = {name: size for name, size in x_aval.named_shape.items()
+                     if name != axis_name}
```
mattjj, line 1025: Similarly to the above, should we check that `axis_name` is present to be gathered over?

jax/_src/lax/parallel.py
```diff
@@ -1113,11 +1125,14 @@ def _pdot_impl(x, y, *, axis_name, pos_contract, pos_batch):
...
 def _pdot_abstract_eval(x, y, *, axis_name, pos_contract, pos_batch):
-  # TODO: avals with names, check inputs are mapped along axis_name, eliminate
   if not len(set(axis_name)) == len(axis_name): raise ValueError
-  return lax.dot_general_p.abstract_eval(
+  pos_aval = lax.dot_general_p.abstract_eval(
       x, y, dimension_numbers=[pos_contract, pos_batch],
       precision=None, preferred_element_type=None)
+  named_shape = {name: size
+                 for aval in (x, y) for name, size in aval.named_shape.items()
+                 if name not in axis_name}
```
mattjj, line 1134: In line with the comments above, we want to check here that the inputs indeed have the names present to be contracted over, right? (As per the TODO that was deleted!)

Since this PR doesn't yet add tests for `pdot`, I suggest keeping some form of TODO here, mentioning that it's untested logic.

jax/_src/lax/parallel.py
```diff
@@ -1269,7 +1284,8 @@ def _axis_index_bind(*, axis_name):
     nreps = dynamic_axis_env.nreps
     trace = frame.pmap_trace
 
-    out_aval = ShapedArray((), np.int32)
+    out_aval = _axis_index_abstract_eval(
```
mattjj, line 1287: Do we actually need to update this non-omnistaging code? I'm curious if tests fail without it.

jax/interpreters/partial_eval.py
```diff
@@ -1110,8 +1110,9 @@ def process_map(self, map_primitive, f, tracers, params):
     constvars = map(self.getvar, map(self.instantiate_const, consts))
     outvars = map(self.makevar, out_tracers)
     new_in_axes = (None,) * len(consts) + params['in_axes']
-    new_params = dict(params, in_axes=new_in_axes, out_axes=out_axes,
-                      call_jaxpr=convert_constvars_jaxpr(jaxpr))
+    with core.extend_axis_env(axis_name, axis_size, None):  # type: ignore
```
mattjj, line 1113: Leave a comment explaining why this is necessary?

tests/lax_test.py
```diff
@@ -2452,5 +2452,28 @@ def testCumsumLengthOne(self):
...
+    self.assertEqual(out, expected)
+
+  def test_abstract_eval_collective(self):
+    if not config.omnistaging_enabled:
+      raise SkipTest("test requires omnistaging")
+    with core.extend_axis_env('i', 10, None):
+      aval1 = core.ShapedArray((2, 3), np.float32, False, {'i': 10, 'j': 5})
+      expected = core.ShapedArray((2, 3), np.float32, False, {'j': 5})
+      out, = lax.psum_p.abstract_eval(aval1, axes=('i',), axis_index_groups=None)
+      self.assertEqual(out, expected)
```
mattjj, line 2476: Can we add a test for `lax.psum_p.abstract_eval(aval2, axes=('i',), axis_index_groups=None)` raising an error when `aval2 = core.ShapedArray((2, 3), np.float32, False, {'j': 5})`? (Consistent with comment above.)

jax/_src/lax/parallel.py
jekbradbury: No, I think we do want to go the looser route here. I'm too tired to make the case right now, though :(
(And I disagree with Adam, at least, and probably also the rest of you!)

mattjj: We have since talked extensively about this among the three of us! We haven't figured out exactly what to do, but we have figured out that we can separate concerns: we can land this PR thinking of names as "metadata" (rather than types), in which case it's just tracking more information and not changing any behaviors (or ruling out any future behaviors), and figure out the type-level story separately.

jax/interpreters/partial_eval.py
mattjj: I added a comment; will push a commit in a sec.

mattjj:
This LGTM to land!

PR #5704, opened 2021-02-11
accept any arguments with shape/dtype attributes after make_jaxpr

froystig:
Also factor out common argument conversion logic from several `api` functions.


mattjj (approved):
Nice!

PR #5705, opened 2021-02-11
Cleanup unnecessary metadata from documentation notebooks

jakevdp:
This makes the markdown source more readable. I accomplished this using a series of regex find/replace in the markdown versions, then ran `jupytext --sync`.

The only downside here is that cell IDs in the Colab versions of the notebooks will change, so if anyone was linking to a particular cell via that mechanism, the links will no longer find the correct cell. My judgment is that it's better to have a cleaner documentation source than to try to maintain consistent Colab cell IDs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ print(normalize(jnp.arange(4.)))
 # prints [0.         0.16666667 0.33333334 0.5       ]
 ```
 
-You can even [nest `pmap` functions](https://colab.sandbox.google.com/github/google/jax/blob/master/cloud_tpu_colabs/Pmap_Cookbook.ipynb#scrollTo=MdRscR5MONuN) for more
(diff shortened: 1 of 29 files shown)


gnecula:
Is it possible to manually add a cell id to those cells to which one wants to link? Most often we link to sections in the notebook. Do they have stable ids (perhaps based on the title)?

jakevdp:
Note that the caveat about cell IDs only applies to notebooks opened in Colab. The documentation generated from these notebooks is not affected by the cell ID.

If we wish to be able to link persistently to any cell in the Colab view of the notebook, we'll have to add a metadata specification with a cell ID.

The reason I think this is the right call is because I think of these notebooks as primarily a source from which our HTML docs are built, and only secondarily as executable documents that can be opened in Colab.

jakevdp:
On second thought I think we should keep the metadata.

PR #5718, opened 2021-02-12
Add :jax-issue: sphinx role for use in CHANGELOG.

jakevdp:
Rather than writing something like
```
`#5244 <https://github.com/google/jax/pull/5244>`_
```
this allows us to write
```
:jax-issue:`5244`
```
equivalently, in markdown docs we can use
```
{jax-issue}`5244`
```
This will make it easier and less error-prone to add github issue links in the changelog and elsewhere.

A disadvantage is that the the links will only be rendered correctly in the HTML documentation, and not in the github view of the file, but the github view is already filled with a bunch of ugly sphinx directives (see the github view of [CHANGELOG.md](https://github.com/google/jax/blob/master/docs/CHANGELOG.md)), so I don't think this is too much of a down side.


mattjj (approved):
I'm so happy for this! Thanks for the crystal-clear explanation too.

PR #5720, opened 2021-02-12
BUG: fix empty bincount

### LINKED ISSUE #5719: bincount of empty array raises
Running
`jnp.bincount(jnp.int32([]))`
raises
`ValueError: zero-size array to reduction operation max which has no identity`

In contrast, `np.bincount([])` produces `array([], dtype=int64)`

jakevdp:
Fixes #5719


brianwa84 (approved):
Thanks for the quick fix!

PR #5736, opened 2021-02-16
Add new lax.rng_bit_generator primitive

tberghammer:
The new primitive provides access to the RngBitGenerator HLO
(https://www.tensorflow.org/xla/operation_semantics#rngbitgenerator)


jax/_src/lax/lax.py
```diff
@@ -6089,6 +6089,50 @@ def _rng_uniform_translation_rule(c, a, b, *, shape):
...
+
+RandomAlgorithm = xops.RandomAlgorithm
+RandomAlgorithm.__str__ = lambda algorithm: algorithm.name
+
+
+def rng_bit_generator(key,
+                      shape,
+                      dtype=np.uint32,
+                      algorithm=RandomAlgorithm.RNG_DEFAULT):
+  shape = jax.core.canonicalize_shape(shape)
```
hawkinsp, line 6130: Document?

In particular, I'd probably mention that most users should be looking for `jax.random` here, not this function.

tests/lax_test.py
```diff
@@ -2284,6 +2284,22 @@ def test_select_jvp_complexity(self):
                                              (x,), (1.,)))(1.)
     self.assertLen(jaxpr.jaxpr.eqns, 2)
 
+  def testRngBitGenerator(self):
```
hawkinsp, line 2287: You need to gate this test on having a sufficiently new jaxlib version. The C++ parts of jax are on a separate release schedule to the Python parts.

Perhaps
```
@unittest.skipIf(jax.lib.version < (0, 1, 62), "Needs jaxlib 0.1.62 or newer")
```
or similar?

hawkinsp (changes requested):
I'm still unsure about this as an API, but I agree it's in XLA and we should at least provide the option to use it.

jax/_src/lax/lax.py
tberghammer: Done.

tests/lax_test.py
tberghammer: I had to condition not only the test but the logic in `lax.py` and in `lax/__init__.py` because `RandomAlgorithm = xops.RandomAlgorithm` requires the op to be present.

I am not too happy with the added complexity so it might be nicer if we wait with this until a new version of jaxlib gets released.

tberghammer: I thought #5714 or #6001 which released jaxlib version 0.1.61 should have solved the issue but I might be still missing something.

Is there something I can do to make this work without having to add a version dependent import into `lax/__init__.py`?

hawkinsp: I don't think you need to gate the logic in `lax.py`. Remember that Python is dynamic, so if you don't execute `rng_bit_generator` it won't fail due to missing symbols.

If you want to be sure that the rng bit generator is present, we would need to bump the minimum jaxlib version. We should probably do that soon regardless, but I think gating the test (only) on a jaxlib version check should suffice to get this submitted.

jax/_src/lax/lax.py
```diff
@@ -6083,6 +6083,67 @@ def _rng_uniform_translation_rule(c, a, b, *, shape):
...
+xla.translations[rng_bit_generator_p] = _rng_bit_generator_translation_rule
+
+RandomAlgorithm = xops.RandomAlgorithm
+RandomAlgorithm.__str__ = lambda algorithm: algorithm.name
+
+
+def rng_bit_generator(key,
+                      shape,
+                      dtype=np.uint32,
+                      algorithm=RandomAlgorithm.RNG_DEFAULT):
```
hawkinsp, line 6127: Oh, I guess the default is problematic here. Use `None` for now, and add
```
algorithm = algorithm or RandomAlgorithm.RNG_DEFAULT here
```

jax/_src/lax/lax.py
```diff
@@ -6083,6 +6083,67 @@ def _rng_uniform_translation_rule(c, a, b, *, shape):
...
+rng_bit_generator_p.multiple_results = True
+rng_bit_generator_p.def_impl(partial(xla.apply_primitive, rng_bit_generator_p))
+rng_bit_generator_p.def_abstract_eval(
+    partial(standard_multi_result_abstract_eval, rng_bit_generator_p,
+            _rng_bit_generator_shape_rule, _rng_bit_generator_dtype_rule,
+            _rng_bit_generator_weak_type_rule,
+            _rng_bit_generator_named_shape_rule))
+xla.translations[rng_bit_generator_p] = _rng_bit_generator_translation_rule
+
+RandomAlgorithm = xops.RandomAlgorithm
```
hawkinsp, line 6120: Actually I guess you would need to gate these with a version check. I think it's no big deal though:
```
if jax.lib.version >= (...):
  RandomAlgorithm = ....
```

jax/lax/__init__.py
```diff
@@ -19,6 +19,7 @@
   DotDimensionNumbers,
   GatherDimensionNumbers,
   Precision,
+  RandomAlgorithm,
```
tberghammer, line 22: My main problem is what to do here? I don't think I can (reasonably) condition it on `jax.lib.version >= (...)` because  importing `jax` in this `__init__` file feels pretty wrong.

One solution I can imagine is to define a default version for these functions in `lax.py` what just throws an error so they exists (so importing them will work fine) but don't do anything in runtime. What do you think?

hawkinsp: I think it's fine to just set `RandomAlgorithm` to
```
class RandomAlgorithm: pass
```
if the version is too old.

(This is only a temporary state, we can bump the minimum, but I'd get this PR in first.)

jax/_src/lax/lax.py
tberghammer: I think I managed to condition all of the logic on jax.lib.version. It is not the prettiest but should be easy to read and we can get rid of the extra complexity when the minimum jaxlib version is bumped

hawkinsp (approved):

Issue #5730, opened 2021-02-15
Major efficiency improvements from building from source (not mentioned in docs)


DiffeoInvariant:
Hi JAX devs,

One thing I've observed recently is massive performance gains (2x speed increase or more) from building `jaxlib` from source, likely as a result of XLA's ops being compiled with all the vector instructions my machine has (as opposed to the binary that `pip` downloads). I first noticed this after building the (ROCm) GPU version and noticing that it was a lot faster than the pip-installed CPU only version even though my CPUs have higher floating-point bandwidth than my GPU, so I tried comparing the CPU-only pip version and the build-from-source CPU-only version, and found the massive performance gains I'm talking about here. 

This doesn't seem to be mentioned anywhere in the JAX documentation, and it would probably be useful for a lot of people, so I figured I'd bring it up here. I'm happy to add a note in the docs myself and make a PR if that's easier for y'all than writing it yourselves and you'd want that.

I suppose you could also compile multiple variants of `jaxlib` and include those with the pip package if you think that's a better solution (to just make the pip package fast on more systems)...

hawkinsp:
This is actually a little bit surprising to me!

The reason I say this is that JAX (via XLA) is a just-in-time compiler. That is, it generates code for your particular hardware, tailored for your particular hardware, every time it is run. For example, if you have an Intel Skylake CPU, JAX would generate AVX512 instructions via JIT compilation, even though the `jaxlib` wheel itself does not require or use such instructions. The build flags should only really affect things like the speed of the compiler itself, and should not benefit significantly from, say, vectorization.

Can you share a benchmark and instructions to reproduce the issue? I'd like to reproduce this so I can dig into it a little bit deeper and see what is going on.

(As it happens, I was thinking of at the very least requiring AVX instructions for JAX shortly for a host kernel I'm planning to add, but I'm surprised there's any meaningful difference at the moment!)

DiffeoInvariant:
Sure, I'll put together a good benchmark for this (the code I observed the effect on isn't a good benchmark, but I should be able to extract a good one from it) and drop it here later today. 

[...]

PR #5738, opened 2021-02-16
Enable AVX by default in release builds.

hawkinsp:
TensorFlow also enables AVX by default and has done since v1.6, so this seems like a reasonable baseline CPU feature to expect.

This mostly affects a small number of host kernels, notably matrix multiplication and convolution, especially of `float64` point types.

Also refactor the Linux/Mac feature flag handling in the `.bazelrc` file.

May help with https://github.com/google/jax/issues/5730


skye (approved):
Nice!

PR #5745, opened 2021-02-16
move convolutions from common gotchas to its own file

jakevdp:
When the section on convolutions was first added to the **Sharp Bits** documentation, the `jax.numpy.convolve` and `jax.scipy.signal.convolve` interfaces did not exist. Now that those simpler interfaces are available, this no longer makes as much sense as a "sharp bit". This PR moves that section to its own notebook, and adds some introductory examples of simple convolutions with `jax.numpy.convolve` and `jax.scipy.signal.convolve`.

Generated docs can be viewed here: https://jax--5745.org.readthedocs.build/en/5745/notebooks/convolutions.html


docs/notebooks/convolutions.md
```diff
@@ -0,0 +1,506 @@
...
+id: Jk5qdnbv6QgT
+outputId: 292205eb-aa09-446f-eec2-af8c23cfc718
+---
+from scipy import misc
+import jax.scipy as jsp
+
+fig, ax = plt.subplots(1, 3, figsize=(12, 5))
+
+# Load a sample image; compute mean() to convert from RGB to grayscale.
+image = jnp.array(misc.face().mean(-1))
```
skye, line 90: I love `misc.face`!!

docs/notebooks/convolutions.md
```diff
@@ -0,0 +1,506 @@
...
+ax[2].set_title('smoothed');
+```
+
++++ {"id": "Op-NhXy39z2U"}
+
+Like in the one-dimensional case, we use `mode='same'` to specify how we would like edges to be handled. For more information on available options in *N*-dimensional convolutions, see the {func}`jax.scipy.signal.convole` documentation.
+
++++ {"id": "bxuUjFVG-v1h"}
+
+## General Convolutions
```
skye, line 114: These are copy/pasted from the Sharp Bits notebook right? As someone who doesn't know much about convolutions, I find these examples confusing, but maybe I'm not the target audience and/or we should land this as a copy/paste and make improvements from there? If you do wanna fix this section up, lemme know and I can leave more detailed review comments.

skye (approved):
This is really great!

docs/notebooks/convolutions.md
jakevdp: Yes, this is just directly copy-pasted. I might make a pass at these in a later PR, but I'd prefer to leave them unchanged for now.

PR #5747, opened 2021-02-17
docs fix: correct overwriting of step function in optimizers example

theosanderson:
`step` is used in this example to refer to both step number and the step function, which leads to an error if implemented naively

### DIFF
--- a/jax/experimental/optimizers.py
+++ b/jax/experimental/optimizers.py
@@ -78,7 +78,7 @@ Example Usage:
     return value, opt_state
 
   for i in range(num_steps):
-    value, opt_state = step(step, opt_state)
+    value, opt_state = step(i, opt_state)
 """
 
 from typing import Any, Callable, NamedTuple, Tuple, Union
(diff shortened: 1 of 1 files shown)


jax/experimental/optimizers.py
```diff
@@ -77,7 +77,7 @@ def step(step, opt_state):
     opt_state = opt_update(step, grads, opt_state)
     return value, opt_state
 
-  for step in range(num_steps):
+  for i in range(num_steps):
     value, opt_state = step(step, opt_state)
```
jakevdp, line 81: should this then be `step(i, opt_state)`?

jakevdp:
Good catch :grin:

theosanderson:
Indeed it should -- thanks! Because this didn't throw an error I didn't see the second occurrence :)

jakevdp (approved):

jakevdp:
Thanks - would you mind squashing the changes into a single commit?

PR #5748, opened 2021-02-17
Set qualname in jax.numpy.util._wraps

jakevdp:
Previous behavior:
```python
import jax.numpy as jnp
print(jnp.cumsum.__qualname__)
# _make_cumulative_reduction.<locals>.cumulative_reduction
```
With this fix:
```python
import jax.numpy as jnp
print(jnp.cumsum.__qualname__)
# cumsum
```
Some editors and code inspectors rely on `__qualname__`, so this should make things cleaner.

### DIFF
--- a/jax/_src/numpy/util.py
+++ b/jax/_src/numpy/util.py
@@ -103,6 +103,7 @@ def _wraps(fun, update_doc=True, lax_description=""):
           f"{body}")
 
       op.__name__ = fun.__name__
+      op.__qualname__ = fun.__qualname__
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):
This seems fine, although some quick googling doesn't really answer the question for me: *should* decorators propagate `__qualname__`? I'm happy with the change, though.

jakevdp:
Thanks - `__qualname__` is one of the attributes set by `functools.wraps`, so I think it does make sense in this context.

PR #5761, opened 2021-02-17
Use lax.pad in jnp.packbits.

hawkinsp:
jnp.pad() is overkill for this situation.

Additional context: using `packbits` inside a `vmap`-ed cond fails (b/180481587). This PR works around the issue, and produces a simpler jaxpr so seems like a good change regardless.


jakevdp (approved):
Looks good!

PR #5765, opened 2021-02-17
Regularize notebook metadata

jakevdp:
Also ensure notebooks execute.

This should have no visible effect on the built docs or the Colab views of the notebooks; The main benefit here is reduction of boilerplate and silencing of a few extraneous sphinx build warnings.


mattjj (approved):
Feels cleaner already!

PR #5781, opened 2021-02-19
introduce `linear_call` for custom transposition

froystig:
This adds a primitive with a corresponding traceable function in `custom_derivatives` that takes a callee and its transpose, both functions. When the primitive is encountered during transposition, the given transpose function is invoked instead of transpose-transforming the callee. The invocation of the custom transposition is itself done via a `linear_call`, with the original callee set as the transpose. This maintains, in particular, that transposing twice is an identity.

The `linear_call` docstring gives examples. Here's the IR view of an example.
```python
$ cat test.py
from jax import linear_transpose, make_jaxpr
from jax.custom_derivatives import linear_call

def transpose(f, x_example):
  def transposed(t):
    x, = linear_transpose(f, x_example)(t)
    return x
  return transposed

def halve(x):
  def f(r, x): return x / r
  def t(r, t): return t * r    # custom!
  return linear_call(f, t, 2., x)

x, t = 1., 1.
print(make_jaxpr(halve)(x))
print(make_jaxpr(transpose(halve, x))(t))
```
```
$ python test.py 
{ lambda  ; a.
  let b = linear_call[ callee={ lambda  ; a b.                                  
                                let c = div b a                                 
                                in (c,) }
                       num_callee_consts=0
                       num_res=1
                       num_transpose_consts=0
                       transpose={ lambda  ; a b.                               
                                   let c = mul b a                              
                                   in (c,) } ] 2.0 a                            
  in (b,) }
{ lambda  ; a.
  let b = linear_call[ callee={ lambda  ; a b.
                                let c = mul b a
                                in (c,) }
                       num_callee_consts=0
                       num_res=1
                       num_transpose_consts=0
                       transpose={ lambda  ; a b.
                                   let c = div b a
                                   in (c,) } ] 2.0 a
  in (b,) }
```

Notes:

* We can later revisit whether to tie the transpose's transpose back to the original function via `linear_call`. There might be a good case for "inlining" the custom transpose function instead, as soon as the first transposition happens.
* Unlike many other call primitives, all arguments are fully staged out to jaxpr in advance here.
* This implementation materializes all symbolic zeros encountered when transposing `linear_call`. Representing the custom transpose function as a jaxpr essentially prevents it from accepting and processing symbolic zeros.

Up next: primitive rules for various transformations (batching, etc.).

Further out, I'd like to relate this to custom derivatives. Does custom transposition let us concurrently set up a custom JVP and VJP?

For a function `a -> b`, the current `custom_jvp` setup takes a custom JVP rule of the form:
```haskell
jvp :: (a, T a) -> (b, T b)
```
and `custom_vjp` takes custom forward and reverse functions of the form:
```haskell
fwd :: a -> (b, r)
rev :: (r, T b) -> T a
```
Suppose instead of `jvp` we were given:
```haskell
lin :: a -> T a -> T b
```
mirroring linearization, but without requiring structural linearity. Then we could generate a custom JVP along the lines of:
```python
def jvp(x, tx):
  y, res = fwd(x)
  tf = lin(x)
  ty = linear_call(tf, rev, res, tx)
  return y, ty
```
This satisfies the JVP signature and picks up _some_ custom JVP logic via `lin`, but it also carries out the custom VJP's forward rule. If linearized, it stages out a function that, when later transposed, carries out the custom VJP's reverse rule as well.

There are some potential efficiency losses with this approach, because accepting only `lin` means writing expressions to compute primals separately from tangents within `jvp`. This is perhaps not an issue when both expressions consist entirely of first-order primitives, especially after compiler optimizations. But sometimes primals and tangents can be jointly expressed, say, with one `scan`, while separate expressions require separate `scan` calls. I'm not sure that we can rely on a compiler to always fuse the two loops back into one.

### DIFF
--- a/jax/custom_derivatives.py
+++ b/jax/custom_derivatives.py
@@ -21,8 +21,9 @@ from typing import Callable, Sequence, Tuple, Any
 from . import core
 from . import dtypes
 from . import linear_util as lu
-from .tree_util import (tree_flatten, tree_unflatten, tree_map, tree_multimap,
-                        treedef_is_leaf, register_pytree_node_class)
+from .tree_util import (tree_flatten, tree_unflatten, tree_map,
+                        tree_multimap, treedef_is_leaf, treedef_tuple,
+                        register_pytree_node_class)
 from ._src.util import cache, safe_zip, safe_map, split_list
 from .api_util import flatten_fun_nokwargs, argnums_partial, wrap_hashably
 from .core import raise_to_shaped
@@ -54,6 +55,9 @@ def _resolve_kwargs(fun, args, kwargs):
 def _initial_style_jaxpr(fun, in_avals):
   jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(fun, in_avals)
   return jaxpr, consts
+
+def _close_jaxpr(jaxpr):
+  return core.ClosedJaxpr(pe.convert_constvars_jaxpr(jaxpr), ())
 
 def _initial_style_staging() -> bool:
   return core.thread_local_state.trace_state.initial_style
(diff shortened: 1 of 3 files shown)


jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
 
 def abstractify(x):
   return core.raise_to_shaped(core.get_aval(x))
+
+
+### Custom transposition
+
+def linear_call(fun: Callable, fun_transpose: Callable, residual_args,
+                linear_args):
+  """Call a function, indicating a custom implementation for its transpose.
```
mattjj, line 964:
```suggestion
  """Call a linear function, indicating a custom implementation for its transpose.
```

Would it make sense to mention 'linear' up-front here?

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+  .. code-block:: haskell
+
+    fun           :: r -> a -o b
+    fun_transpose :: r -> b -o a
+
+  where the ``-o`` arrow indicates a linear function, ``r`` is the
+  residual input type and ``a`` is the linear input type.
+
+  The functions ``fun`` and ``fun_transpose`` are coupled as
+  transposes of one another across transposition. Specifically, the
```
mattjj, line 977:
```suggestion
  transposes of one another. Specifically, the
```

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+  1.
+  >>> transpose(custom_id, 1.)(1.)
+  7.
+  >>> transpose(transpose(custom_id, 1.), 1.)(1.)
+  1.
+  >>> transpose(transpose(transpose(custom_id, 1.), 1.), 1.)(1.)
+  7.
+
+  Args:
+    fun: a Python callable specifying a structurally linear
```
mattjj, line 1026: In earlier parts of the docstring, it doesn't say "structurally". That made me think about whether we actually require it. I think because we're not automatically transposing `fun` then it should just correspond to a mathematically linear function, not necessarily a structurally linear one. So maybe we can strike that term here. WDYT?

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+    def div_add(denom, x):
+      return x + linear_call(f, t, denom, x)
+
+    def transpose(f, x_example):
+      def transposed(y):
+        x, = jax.linear_transpose(f, x_example)(y)
+        return x
+      return transposed
+
+  Then:
```
mattjj, line 998: Might be good to call out how this is a good example of what "residual" means. That is, divide is linear in one of its inputs but not the other.

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+      (type ``r``), i.e. inputs in which the function is not
+      necessarly linear, and one of "linear" inputs (type ``a``).  It
+      should return output whose components are linear in the linear
+      input (type ``b``).
+    fun_transpose: a Python callable specifying a structurally linear
+      function that is the transpose of ``fun`` with respect to its
+      linear inputs. Its first argument is the same residual inputs
+      (``r``) as ``fun``. Its second argument is of type
+      ``b``. Finally, its output is of type ``a`` and each of its
+      component are linear in its second argument (the ``b`` inputs).
```
mattjj, line 1037: Document `residual_args` and `linear_args`? The API may change but we'll probably have to document them somewhere, so might be good to cut a draft now. Also, I suspect that "residual" might be one of the things most in need of documentation.

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+  operands_lin, lin_tree = tree_flatten(linear_args)
+
+  f_in_tree = treedef_tuple((res_tree, lin_tree))
+  f, out_tree = flatten_fun_nokwargs(lu.wrap_init(fun), f_in_tree)
+
+  res_avals = map(abstractify, operands_res)
+  lin_avals = map(abstractify, operands_lin)
+  f_jaxpr, f_consts = _initial_style_jaxpr(f, (*res_avals, *lin_avals))
+  f_jaxpr = _close_jaxpr(f_jaxpr)
+  out_avals = map(core.raise_to_shaped, f_jaxpr.out_avals)
```
mattjj, line 1053: I think you and I have a TODO somewhere in the codebase to ensure these are always at shaped level, as "jaxpr types" = "abstract values at shaped level". That's not actionable for this PR; I was just reminded of it!

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+
+  t_in_tree = treedef_tuple((res_tree, out_tree()))
+  t, t_out_tree = flatten_fun_nokwargs(lu.wrap_init(fun_transpose), t_in_tree)
+
+  t_jaxpr, t_consts = _initial_style_jaxpr(t, (*res_avals, *out_avals))
+  t_jaxpr = _close_jaxpr(t_jaxpr)
+
+  if t_out_tree() != lin_tree:
+    raise TypeError(
+        f'transpose output pytree structure must match that of linear inputs, '
```
mattjj, line 1063:
```suggestion
        'transpose output pytree structure must match that of linear inputs, '
```

jax/custom_derivatives.py
```diff
@@ -951,3 +955,162 @@ def merge(l1, l2):
...
+         for ct, a in zip(cts, cts_avals)]
+
+  cts_out = linear_call_p.bind(*t_consts, *f_consts, *operands_res, *cts,
+                               callee=transpose,
+                               transpose=callee,
+                               num_callee_consts=len(t_consts),
+                               num_transpose_consts=len(f_consts),
+                               num_res=len(operands_res))
+
+  return ([None] * num_callee_consts +
```
mattjj, line 1104: Could do `[None] * (num_callee_consts + num_transpose_consts + num_res)`

mattjj (approved):
Masterfully done! Very clean.

shoyer:
Very cool!

Thinking about this from the perspective of the similar `custom_linear_solve`:

- Would this obviate the need for a custom primitive for `custom_linear_solve`? It would be interesting to see what this would look like, presumably in combination with `custom_jvp`.
- It seems like `linear_call` could probably also use a batching rule. This might look something similar to the batching rule for `custom_linear_solve`.

jax/custom_derivatives.py
froystig: Good point!

jax/custom_derivatives.py
froystig: Correct. It's in [partial eval of cond](https://github.com/google/jax/blob/b21341d0adb26fe2f384703a327dde8379396c64/jax/_src/lax/control_flow.py#L877).

froystig:
> * Would this obviate the need for a custom primitive for `custom_linear_solve`? It would be interesting to see what this would look like, presumably in combination with `custom_jvp`.

Good point!

> * It seems like `linear_call` could probably also use a batching rule. This might look something similar to the batching rule for `custom_linear_solve`.

Yes, support for various transformations will come next. Good idea looking at `custom_linear_solve` for guidance.

froystig:
I've updated the PR description with some notes on this implementation and thoughts on what's next.

PR #5794, opened 2021-02-20
Fix error message string interpolation

lucasb-eyer:


jakevdp:
Good catch, thank you!

PR #5797, opened 2021-02-20
Fix typo in thinking_in_jax notebook

jamesr66a:
The commented code attempted to call `prot()`, which looked like a typo in trying to call `prod()`.


jakevdp:
Good catch, thanks!

Issue #5786, opened 2021-02-19
[feature request] NotImplementedError: jax.numpy.cov not implemented for nontrivial y.


cccntu:
I got this error when using `jnp.corrcoef()`
> NotImplementedError: jax.numpy.cov not implemented for nontrivial y. Open a feature request at https://github.com/google/jax/issues !

minoring:
Can I handle this issue? :)

jakevdp:
@minoring - that would be great! I'll assign this to you, please feel free to open a work-in-progress PR and reach out if you need any help or advice.

PR #5799, opened 2021-02-21
Implement jax.numpy.cov for nontrivial y

minoring:
Related to  #5786

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4820,19 +4820,27 @@ def compress(condition, a, axis: Optional[int] = None, out=None):
 @_wraps(np.cov)
 def cov(m, y=None, rowvar=True, bias=False, ddof=None, fweights=None,
         aweights=None):
-  if y is not None: raise NotImplementedError(
-    "jax.numpy.cov not implemented for nontrivial y. "
-    "Open a feature request at https://github.com/google/jax/issues !")
-
   m, = _promote_args_inexact("cov", m)
+  if y is not None:
+    y, = _promote_args_inexact("cov", y)
 
   if m.ndim > 2:
     raise ValueError("m has more than 2 dimensions")  # same as numpy error
+  if y is not None:
+    if y.ndim > 2:
+      raise ValueError("y has more than 2 dimensions")
+
   X = atleast_2d(m)
   if not rowvar and X.shape[0] != 1:
     X = X.T
   if X.shape[0] == 0:
     return array([]).reshape(0, 0)
+
+  if y is not None:
+    y = atleast_2d(y)
+    if not rowvar and y.shape[0] != 1:
+      y = y.T
+    X = concatenate((X, y), axis=0)
   if ddof is None:
     ddof = 1 if bias == 0 else 0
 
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -4820,19 +4820,27 @@ def compress(condition, a, axis: Optional[int] = None, out=None):
 @_wraps(np.cov)
 def cov(m, y=None, rowvar=True, bias=False, ddof=None, fweights=None,
         aweights=None):
-  if y is not None: raise NotImplementedError(
-    "jax.numpy.cov not implemented for nontrivial y. "
-    "Open a feature request at https://github.com/google/jax/issues !")
-
   m, = _promote_args_inexact("cov", m)
+  if y is not None:
+    y, = _promote_args_inexact("cov", y)
```
jakevdp, line 4825: It would be better to promote both inputs to the same type; I think something like this would work:
```python
  if y is not None:
    m, y = _promote_args_inexact("cov", m, y)
  else:
    m, = _promote_args_inexact("cov", m)
```
You could also put the check on y's shape within this block to make things more succinct.

tests/lax_numpy_test.py
```diff
@@ -4267,28 +4267,30 @@ def np_fun(x):
   @parameterized.named_parameters(
       jtu.cases_from_list(
         {"testcase_name":
-          "_shape={}_dtype={}_rowvar={}_ddof={}_bias={}_fweights={}_aweights={}".format(
-            shape, dtype, rowvar, ddof, bias, fweights, aweights),
-         "shape": shape, "dtype": dtype, "rowvar": rowvar, "ddof": ddof,
+          "_shape={}_dtype={}_y={}_rowvar={}_ddof={}_bias={}_fweights={}_aweights={}".format(
```
jakevdp, line 4270: Rather than naming this parameter `y`, how about `y_dtype` for more clarity?

Also, I noticed that this test only checks cases where `m` and `y` have the same shape, where the algorithm is designed for more general cases. It would be useful to test those more general cases, though it would involve adding some logic to choose compatible shapes of `y`.

jakevdp (changes requested):
Looks great, thanks! Just a couple minor comments below.

jax/_src/numpy/lax_numpy.py
minoring: Thanks for your kind review :)
I agree that promoting both m and y make code looks cleaner!

tests/lax_numpy_test.py
minoring: Chagned `y` into `y_dtype` ;)

Also added testcases where `m` and `y` have different shape but compatible.

PTAL!

tests/lax_numpy_test.py
```diff
@@ -4267,25 +4267,34 @@ def np_fun(x):
...
     rng = jtu.rand_default(self.rng())
     wrng = jtu.rand_positive(self.rng())
     wdtype = np.real(dtype(0)).dtype
     wshape = shape[-1:] if rowvar or shape[0] == 1 else shape[:1]
+
+    import random
+    if len(shape) == 1:
+      y_shape = shape
+    elif rowvar or shape[0] == 1:
+      y_shape = (random.randint(1, 10), shape[-1])
```
jakevdp, line 4292: Using `random` like this makes the tests non-deterministic. It would be better if `y_shape` were passed as an argument in the parametrization, perhaps using a helper function to return a sequence of valid `y_shape` values given the `shape` value.

jakevdp (approved):
Looks great, thanks!

minoring:
Thanks :)
Always learn new things from you!

jakevdp:
Thanks again for the contribution!

PR #5824, opened 2021-02-23
readthedocs: set fail_on_warning=true

jakevdp:
Over the past six weeks, I've been slowly chipping away at our documentation build warnings: we had 300+, and are now down to zero! Warnings are non-fatal by default, but generally indicate that mis-formatted HTML will be generated.

Now that we're at zero, let's keep it that way!

This PR sets readthedocs so that any new warnings will be treated as errors; this will help us prevent introducing new formatting issues into the documentation. As a bonus, this obviates the myst-nb monkeypatch that we had been using to turn cell execution failure warnings into errors.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,7 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Woohoo!

PR #5827, opened 2021-02-24
Add first two sections of Autodidax

mattjj:
This is meant to be core documentation as well as a model system of JAX. The overarching goal is to make JAX developers more effective, specifically by
1. documenting the JAX core design;
2. identifying simplifications to be made to the core implementation;
3. maintaining a model system on which we can more easily prototype new designs.

**It's probably easiest to read the .ipynb, clicking 'view file' on GitHub to render it.**

This first PR includes the first two sections. The first section builds up primitives, the core tracing logic and `Trace`/`Tracer` classes, and both `jvp` and `vmap`. The second section introduces jaxprs. (The third will then use jaxprs to introduce both `jit` and `vjp`/`grad`. The fourth will have control flow.)

It's currently just ~530 lines, according to `cloc`:
```
Python                           1            217            289            530
```

Some simplifications that I'm already keen to make in main-line JAX (these are sort of notes to self):
* Let's make `jit` 'initial-style', meaning make it look more like the control flow primitives in that it'll form a jaxpr up-front. The main reason to have a fancy 'final-style' primitive is to support Python control flow, but `jit` doesn't even make use of that because its purpose is to delay / stage-out computation. A secondary reason is that it effectively fuses transformations and avoids multiple round-trips through jaxprs, but we don't have evidence that's useful. Once we make `jit` initial-style (and `pmap` too, though there's more to be said there), only `custom_jvp` and `custom_vjp` will need the fancy 'final-style' call behavior.
* We can probably get rid of `post_process_call` by having call bind lift into all traces on the trace stack.
* I think we can simplify `jax.remat`'s implementation, because it was written in a pre-omnistaging world where we only had partial eval to build jaxprs.


froystig (approved):
Excellent overall! Comments:

* `(a, T a) -> (b, T b)` comes up without context that `T a` stands for the tangent type of `a`.
* The term "jaxpr" is introduced without explicit definition. Calling it a language or an IR might help. I sometimes like to give a summary: that it is an explicitly typed, functional, first-order language. In describing syntax, it may be a helpful guide to say that `<jaxpr>` is what corresponds to a function. Otherwise it is soon suggested by the `lambda` keyword.

These are only suggestions for consideration. No need to take them.

mattjj:
Great comments. I attempted improvements along both those lines:
* I took out the JVP type stuff and just linked to the autodiff cookbook instead.
* I added some of that language you suggested. However, I actually expect the reader to already have some vague sense of what jaxprs are, since these are pretty internal-facing docs.

jakevdp:
I think this broke the CI, because it caused a warning in the doc build (not caught because it wasn't synced when merged).

The fix would be to add the new file to a toctree. You'll probably also have to ignore the md file similar to this line to avoid warnings about duplicate files: https://github.com/google/jax/blob/9b9cea86b9aa6b105c5a03d6d6110069d0857708/docs/conf.py#L117

PR #5829, opened 2021-02-24
DOC: Add more detail to the sharded_jit docstring

j-towns:


j-towns:
Btw I was delighted by how easy sharded_jit was!

skye (approved):
Thanks Jamie!

PR #5835, opened 2021-02-24
[avals with names] abstract eval with while_loop fix

jekbradbury:
#5702 broke internal researchers who were doing complicated things with pmapped while loops because it required the loop condition to be scalar even in named axes. That would affect people doing other kinds of batched control flow, too, as soon as we enable named shapes in vmap.

This re-lands that change without the additional requirement.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -271,7 +271,9 @@ def while_loop(cond_fun: Callable[[T], bool],
     if not treedef_is_leaf(cond_tree) or len(cond_jaxpr.out_avals) != 1:
       msg = "cond_fun must return a boolean scalar, but got pytree {}."
       raise TypeError(msg.format(cond_tree))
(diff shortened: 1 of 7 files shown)


rrnjbrr:
Hi dears I know my mistakes create very problem thank you for every things My estress give my learning chance

mattjj:
For convenience to posterity, 63455e1 was the change needed.

@jekbradbury is there a test for that change? If not, might be good to add one so we know we're covering it in our test suite (not just relying on downstream users to cover it).

rrnjbrr:
hi dear I need to small credit or balance for do anything I have not I have
not money now like every single moment in my life when chance with me but I
have no credit to buy and run and trust myself someone no let me to give
little first action I do not know why my all try for do not risky in 1 or 2
moment doing risky thins but not continue just for maybe ....

rrnjbrr:
hi dear My body is not normally work I wining in cancer war few mounts ago
so I want to create team please if possible guid me I tried 1000 way but or
my knoledge is low or fake mail but thank you for your everythings do I
tring to travel for work on time try to create a time

PR #5844, opened 2021-02-25
Lower remat to while(int(rng(1, 2))) on GPU.

chr1sj0nes:
The existing remat lowering to a conditional doesn't work correctly on GPU. Here, we add an alternative lowering to a single iteration while loop.


trevorcai:
Looks like the failed internal check was a pre-existing error - rerunning the presubmits will probably pass.

jekbradbury (approved):
I think the XLA team has been trying to fix the buffer assignment of the cond gadget, but it seems like what they've tried isn't enough for some cases of nested remat. So this seems like a good idea for now.

Issue #5839, opened 2021-02-25
TypeError: list indices must be integers or slices, not DynamicJaxprTracer


Joy-Preetha:
```python
@jit
def match_faces3(desc):
        distances = np.empty((len(desc), len(database)))
        time1 = time.time()
        for i, descr in enumerate(desc):
            for j, identity in enumerate(database):
                dist = []
                for k, id_desc in enumerate(identity[1]):
                    dist.append(jax_cosine(descr, id_desc))
                distances[i][j] = dist[jnp.argmin(jnp.asarray(dist))]
        time2 = time.time() - time1
        print("time2",time2)
        distances=distances.tolist()
        return jnp.asarray(distances)
```
<ipython-input-24-84d11a42c358> in match_faces3(desc)
     13                 for k, id_desc in enumerate(identity[1]):
     14                     dist.append(jax_cosine(descr, id_desc))
---> 15                 distances[i][j] = dist[jnp.argmin(jnp.asarray(dist))]
     16         time2 = time.time() - time1
     17         print("time2",time2)

TypeError: list indices must be integers or slices, not DynamicJaxprTracer

I understood using vectorization, it can be solved, but I dont know how we can implement vmap here.

mattjj:
Thanks for raising this!

The issue is that the expression `dist[jnp.argmin(jnp.asarray(dist))]` is indexing the list referred to by `dist` with a JAX tracer. I wonder if we can make this error better by overloading `__index__`:

```python
In [1]: class A:
   ...:     def __index__(self):
   ...:         raise Exception("hi")
   ...:

In [2]: [1,2,3][A()]
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-2-543bfcd5aeff> in <module>
----> 1 [1,2,3][A()]

<ipython-input-1-cfb1eaf3bf37> in __index__(self)
      1 class A:
      2     def __index__(self):
----> 3         raise Exception("hi")
      4

Exception: hi
```

As for this particular code, there are actually a few issues with it we need to amend. One is this use of list indexing, but another is the in-place mutation of `distances`; JAX in general doesn't work with side-effects like in-place mutation of arrays (in this case a `numpy.ndarray`). The printing and timing are also side-effects which won't work in a jitted function.

I can provide some concrete advice if you extend it to be a runnable example (including example inputs).

Joy-Preetha:
```python
database = [('8174',
  [array([ 9.2,  9.69,  7.8, -2.6,
[...]

PR #5853, opened 2021-02-25
Improve error when tracer is used as a list index

jakevdp:
Related to #5839.

Before:
```python
>>> jax.jit(lambda i: [1, 2, 3][i])(0)    
FilteredStackTrace                        Traceback (most recent call last)
<ipython-input-2-a6ab99490a5a> in <module>
----> 1 jax.jit(lambda i: [1, 2, 3][i])(0)

<ipython-input-2-a6ab99490a5a> in <lambda>(i)
----> 1 jax.jit(lambda i: [1, 2, 3][i])(0)

FilteredStackTrace: TypeError: list indices must be integers or slices, not DynamicJaxprTracer
```
After:
```python
>>> jax.jit(lambda i: [1, 2, 3][i])(0)
FilteredStackTrace                        Traceback (most recent call last)
<ipython-input-2-a6ab99490a5a> in <module>
----> 1 jax.jit(lambda i: [1, 2, 3][i])(0)

<ipython-input-2-a6ab99490a5a> in <lambda>(i)
----> 1 jax.jit(lambda i: [1, 2, 3][i])(0)

FilteredStackTrace: TypeError: The __index__ method was called on the JAX Tracer object
Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>.

This error can occur when a JAX Tracer object is used as an index for a non-JIT-compatible object, 
such as a Python list or tuple. This can often be fixed by converting the indexed object to a JAX array, 
for example by changing `obj[idx]` to `jnp.asarray(obj)[idx]`.
```

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -493,8 +493,18 @@ class Tracer:
            "array indexing, like `x[idx]`, it may be that the array being "
            "indexed `x` is a raw numpy.ndarray while the indices `idx` are a "
            "JAX Tracer instance; in that case, you can instead write "
-           "`jax.device_put(x)[idx]`.")
+           "`jnp.asarray(x)[idx]`.")
     raise Exception(msg)
+
+  def __index__(self):
+    msg = (f"The __index__ method was called on the JAX Tracer object {self}.\n\n"
+           "This error can occur when a JAX Tracer object is used in a context where "
+           "a Python integer is expected, such as an argument to the range() function, "
+           "or in index to a Python list. In the latter case, this can often be fixed "
+           "by converting the indexed object to a JAX array, for example by changing "
+           "`obj[idx]` to `jnp.asarray(obj)[idx]`."
+           )
+    raise TypeError(msg)
 
   def __init__(self, trace: Trace):
     self._trace = trace
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -650,6 +650,17 @@ class APITest(jtu.JaxTestCase):
                                 "Abstract tracer value"):
       jit(f)(1)
 
+  def test_list_index_err(self):
+    L = [1, 2, 3]
+    def f(n):
+      return L[n]
+
+    assert jit(f, static_argnums=(0,))(0) == L[0]
+    self.assertRaisesRegex(
+        TypeError,
+        "The __index__ method was called on the JAX Tracer object.*",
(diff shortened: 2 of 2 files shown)


jax/core.py
```diff
@@ -493,9 +493,18 @@ def __array__(self, *args, **kw):
...
            "indexed `x` is a raw numpy.ndarray while the indices `idx` are a "
            "JAX Tracer instance; in that case, you can instead write "
-           "`jax.device_put(x)[idx]`.")
+           "`jnp.asarray(x)[idx]`.")
     raise Exception(msg)
 
+  def __index__(self):
+    msg = (f"The __index__ method was called on the JAX Tracer object {self}.\n\n"
+           "This error can occur when a JAX Tracer object is used as an index for "
+           "a non-JIT-compatible object, such as a Python list or tuple. "
```
mattjj, line 502: I'm not sure if we should imply that lists and tuples are not jit-compatible, since they do work in many ways, just not with indexing.

How about something like:
"This error can occur when a JAX Tracer object is used as an index for a list or tuple."

mattjj (approved):

jax/core.py
jakevdp: Sounds good. Turns out this also comes up when a tracer is used in `range()`, so I have to make the message a bit more contextual anyway!

mattjj: Oh great idea! That'll make those errors better too!

jax/core.py
```diff
@@ -493,9 +493,19 @@ def __array__(self, *args, **kw):
...
+           "`jnp.asarray(x)[idx]`.")
     raise Exception(msg)
 
+  def __index__(self):
+    msg = (f"The __index__ method was called on the JAX Tracer object {self}.\n\n"
+           "This error can occur when a JAX Tracer object is used in a context where "
+           "a Python integer is expected, such as an argument to the range() function, "
+           "or in index to a Python list. In the latter case, this can often be fixed "
+           "by converting the indexed object to a JAX array, for example by changing "
+           "`obj[idx]` to `jnp.asarray(obj)[idx]`."
```
shoyer, line 505: The other common situation would be if the argument is passed into `jit`. In which case `static_argnums` is probably the recommended fix.

PR #5854, opened 2021-02-25
Decouple ShardedDeviceArray from _DeviceArray

pschuh:
Make `ShardedDeviceArray` inherit directly from `DeviceArray` instead of `_DeviceArray` in order to allow a c++ version of `ShardedDeviceArray`.


mattjj (approved):
LGTM, though it might be worth running a global presubmit before submitting.

PR #5823, opened 2021-02-23
DOC: add first JAX-101 notebook

jakevdp:
This adds the first of the series of notebooks in the JAX 101 tutorial. The only changes I made were to remove unnecessary ipynb metadata, adjust formats, and change internal links to (currently broken) external links.

You can view the docs built from this PR here: https://jax--5823.org.readthedocs.build/en/5823/

PR #5863, opened 2021-02-26
Fix build warning in autodidax.md

jakevdp:
This change should be entirely invisible to users. Originally part of #5823


mattjj (approved):
Thanks for finding and fixing this!

PR #5864, opened 2021-02-26
DOC: add first JAX-101 notebook

jakevdp:
Re-land the rolled-back changes in #5823. I re-executed the notebook via colab.research, which should fix the previous errors.


mattjj (approved):
Thanks!

PR #5888, opened 2021-03-01
Add jax.errors submodule & error troubleshooting docs

jakevdp:
This is the first step toward adding specific debugging information for common errors in JAX, and moving some of the verbose error boilerplate from the error strings to the documentation. Summary:

- Add `jax.errors` submodule with a few exception classes containing detailed debugging information in the docstring
- Add `docs/errors.rst` which will render to `http://jax.readthedocs.io/en/latest/errors.html` (preview [here](https://jax--5888.org.readthedocs.build/en/5888/errors.html))
- Add `jax.errors` to the doctest CI to ensure the docs don't go stale

### DIFF
--- a/jax/errors.py
+++ b/jax/errors.py
@@ -1,0 +1,19 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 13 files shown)


jax/errors.py
```diff
@@ -0,0 +1,292 @@
+# Copyright 2021 Google LLC
```
hawkinsp, line 1: Can you follow the `_src` pattern? This module should be a trivial forwarding module.

(This saves having to move this code later.)

jakevdp: Done

mattjj (approved):

PR #5890, opened 2021-03-02
Remove test that verifies complex eigh is unimplemented on TPU, in pr…

hawkinsp:
…eparation for implementing complex eigh on TPU.


jakevdp:
Should we also remove the translation rule `NotImplementedError` that this test is designed to remind us to remove?

hawkinsp:
> Should we also remove the translation rule `NotImplementedError` that this test is designed to remind us to remove?

No, not until I've checked in the implementation. But I can't do that right now because the test would fail.

PR #5899, opened 2021-03-02
Enable Hermitian Eigendecompositions on TPU.

hawkinsp:


alewis:
Will this enable the new fast TPU eigh?

PR #5905, opened 2021-03-02
Initial commit of second JAX-101 notebook

jakevdp:
Main TODO is to add authors

### DIFF
--- a/docs/jax-101/02-jitting.ipynb
+++ b/docs/jax-101/02-jitting.ipynb
@@ -1,0 +1,577 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "O-SkdlPxvETZ"
+   },
+   "source": [
+    "# Just In Time Compilation with JAX\n",
+    "\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/jax/blob/master/docs/jax-101/02-jitting.ipynb)\n",
+    "\n",
+    "*Authors: TODO*\n",
+    "\n",
+    "In this section, we will further explore how JAX works, and how we can make it performant.\n",
+    "We will discuss the `jax.jit()` transform, which will perform *Just In Time* (JIT) compilation\n",
+    "of a JAX Python function so it can be executed efficiently in XLA.\n",
+    "\n",
+    "## How JAX transforms work\n",
+    "\n",
+    "In the previous section, we discussed that JAX allows us to transform Python functions. This is done by first converting the Python function into a simple intermediate language called jaxpr. The transformations then work on the jaxpr representation. \n",
+    "\n",
+    "We can show a representation of the jaxpr of a function by using `jax.make_jaxpr`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "id": "P9Xj77Wx3Z2P",
+    "outputId": "77d28eec-bfbb-487b-b10f-e47bc8f7dd0d"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "{ lambda  ; a.\n",
+      "  let b = log a\n",
+      "      c = log 2.0\n",
+      "      d = div b c\n",
+      "  in (d,) }\n"
+     ]
+    }
+   ],
+   "source": [
+    "import jax\n",
+    "import jax.numpy as jnp\n",
+    "\n",
+    "global_list = []\n",
+    "\n",
+    "def log2(x):\n",
+    "  global_list.append(x)\n",
+    "  ln_x = jnp.log(x)\n",
+    "  ln_2 = jnp.log(2.0)\n",
+    "  return ln_x / ln_2\n",
+    "\n",
+    "print(jax.make_jaxpr(log2)(3.0))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "jiDsT7y0RwIp"
+   },
+   "source": [
+    "The [Understanding Jaxprs](https://jax.readthedocs.io/en/latest/jaxpr.html) section of the documentation provides more information on the meaning of the above output.\n",
+    "\n",
+    "Importantly, note how the jaxpr does not capture the side-effect of the function: there is nothing in it corresponding to `global_list.append(x)`. This is a feature, not a bug: JAX is designed to understand side-effect-free (a.k.a. functionally pure) code. If *pure function* and *side-effect* are unfamiliar terms, this is explained in a little more detail in [🔪 JAX - The Sharp Bits 🔪: Pure Functions]](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Pure-functions).\n",
+    "\n",
+    "Of course, impure functions can still be written and even run, but JAX gives no guarantees about their behaviour once converted to jaxpr. However, as a rule of thumb, you can expect (but shouldn't rely on) the side-effects of a JAX-transformed function to run once (during the first call), and never again. This is because of the way that JAX generates jaxpr, using a process called 'tracing'.\n",
+    "\n",
+    "When tracing, JAX wraps each argument by a *tracer* object. These tracers then record all JAX operations performed on them during the function call (which happens in regular Python). Then, JAX uses the tracer records to reconstruct the entire function. The output of that reconstruction is the jaxpr. Since the tracers do not record the Python side-effects, they do not appear in the jaxpr. However, the side-effects still happen during the trace itself.\n",
+    "\n",
+    "Note: the Python `print()` function is not pure: the text output is a side-effect of the function. Therefore, any `print()` calls will only happen during tracing, and will not appear in the jaxpr:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "id": "JxV2p7e2RawC",
+    "outputId": "af7472de-df57-412f-83cb-e468c4b02d64"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "printed x: Traced<ShapedArray(float32[], weak_type=True)>with<DynamicJaxprTrace(level=1/0)>\n",
+      "{ lambda  ; a.\n",
+      "  let b = log a\n",
+      "      c = log 2.0\n",
+      "      d = div b c\n",
+      "  in (d,) }\n"
+     ]
+    }
+   ],
+   "source": [
+    "def log2_with_print(x):\n",
+    "  print(\"printed x:\", x)\n",
+    "  ln_x = jnp.log(x)\n",
+    "  ln_2 = jnp.log(2.0)\n",
+    "  return ln_x / ln_2\n",
+    "\n",
+    "print(jax.make_jaxpr(log2_with_print)(3.))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "f6W_YYwRRwGp"
+   },
+   "source": [
+    "See how the printed `x` is a `Traced` object? That's the JAX internals at work.\n",
+    "\n",
+    "The fact that the Python code runs at least once is strictly an implementation detail, and so shouldn't be relied upon. However, it's useful to understand as you can use it when debugging to print out intermediate values of a computation."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "PgVqi6NlRdWZ"
+   },
+   "source": [
+    "A key thing to understand is that jaxpr captures the function as executed on the parameters given to it. For example, if we have a conditional, jaxpr will only know about the branch we take:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "id": "hn0CuphEZKZm",
+    "outputId": "669c5880-790d-4e5c-b635-f3acc21a0e02"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "{ lambda  ; a.\n",
+      "  let \n",
+      "  in (a,) }\n"
+     ]
+    }
+   ],
+   "source": [
+    "def log2_if_rank_2(x):\n",
+    "  if x.ndim == 2:\n",
+    "    ln_x = jnp.log(x)\n",
+    "    ln_2 = jnp.log(2.0)\n",
+    "    return ln_x / ln_2\n",
+    "  else:\n",
+    "    return x\n",
+    "\n",
+    "print(jax.make_jaxpr(log2_if_rank_2)(jax.numpy.array([1, 2, 3])))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "Qp3WhqaqvHyD"
+   },
+   "source": [
+    "## JIT compiling a function\n",
+    "\n",
+    "As explained before, JAX enables operations to execute on CPU/GPU/TPU using the same code.\n",
+    "Let's look at an example of computing a *Scaled Exponential Linear Unit*\n",
+    "([SELU](https://proceedings.neurips.cc/paper/6698-self-normalizing-neural-networks.pdf)), an\n",
+    "operation commonly used in deep learning:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {
+    "id": "JAXFYtlRvD6p",
+    "outputId": "e850862d-c9b4-4cfe-85a2-fb707178a957"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The slowest run took 21.33 times longer than the fastest. This could mean that an intermediate result is being cached.\n",
+      "100 loops, best of 5: 7.9 ms per loop\n"
+     ]
+    }
+   ],
+   "source": [
+    "import jax\n",
+    "import jax.numpy as jnp\n",
+    "\n",
+    "def selu(x, alpha=1.67, lambda_=1.05):\n",
+    "  return lambda_ * jnp.where(x > 0, x, alpha * jnp.exp(x) - alpha)\n",
+    "\n",
+    "x = jnp.arange(1000000)\n",
+    "%timeit selu(x).block_until_ready()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "ecN5lEXe6ncy"
+   },
+   "source": [
+    "The code above is sending one operation at a time to the accelerator. This limits the ability of the XLA compiler to optimize our functions.\n",
+    "\n",
+    "Naturally, what we want to do is give the XLA compiler as much code as possible, so it can fully optimize it. For this purpose, JAX provides the `jax.jit` transformation, which will JIT compile a JAX-compatible function. The example below shows how to use JIT to speed up the previous function."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "id": "nJVEwPcH6bQX",
+    "outputId": "64b5eeab-0861-4cac-9c48-34586aa14fb6"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "1000 loops, best of 5: 1.6 ms per loop\n"
+     ]
+    }
+   ],
+   "source": [
+    "selu_jit = jax.jit(selu)\n",
+    "\n",
+    "# Warm up\n",
+    "selu_jit(x).block_until_ready()\n",
+    "\n",
+    "%timeit selu_jit(x).block_until_ready()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "hMNKi1mYXQg5"
+   },
+   "source": [
+    "Here's what just happened:\n",
+    "\n",
+    "1) We defined `selu_jit` as the compiled version of `selu`.\n",
+    "\n",
+    "2) We ran `selu_jit` once on `x`. This is where JAX does its tracing -- it needs to have some inputs to wrap in tracers, after all. The jaxpr is then compiled using XLA into very efficient code optimized for your GPU or TPU. Subsequent calls to `selu_jit` will now use that code, skipping our old Python implementation entirely. \n",
+    "\n",
+    "(If we didn't include the warm-up call separately, everything would still work, but then the compilation time would be included in the benchmark. It would still be faster, because we run many loops in the benchmark, but it wouldn't be a fair comparison.)\n",
+    "\n",
+    "3) We timed the execution speed of the compiled version. (Note the use of `block_until_ready()`, which is required due to JAX's [Asynchronous execution](https://jax.readthedocs.io/en/latest/async_dispatch.html) model)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "DRJ6R6-d9Q_U"
+   },
+   "source": [
+    "## Why can't we just JIT everything?\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "DRJ6R6-d9Q_U"
+   },
+   "source": [
+    "## Why can't we just jit everything?\n",
+    "\n",
+    "After seeing the example above, you might be wondering whether we should simply apply `jax.jit` to every function. To understand why this is not the case, and when we should/shouldn't apply jit, let's first check some cases where jit doesn't work."
```
8bitmp3, line 253: Suggestion: (again, "apply jit" -> "apply `jit`" as in a method/fn `jit`. Using  https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit as a reference)

```suggestion
    "After going through the example above, you might be wondering whether we should simply apply `jax.jit` to every function. To understand why this is not the case, and when we should/shouldn't apply `jit`, let's first check some cases where `jit` doesn't work."
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "g_jit(10, 20)  # Should raise an error. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "isz2U_XX_wH2"
+   },
+   "source": [
+    "As you see, running the examples above will give an error. \n",
```
8bitmp3, line 336: Suggestion: (for accessibility/a11y, maybe "As you see" -> "As demonstrated above")

```suggestion
    "As demonstrated above, running the examples above will give an error. \n",
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "\n",
+    "The problem is that we tried to condition on the *value* of an input to the function being jitted. The reason we can't do this is related to the fact mentioned above that jaxpr depends on the actual values used to trace it. \n",
```
8bitmp3, line 338: Suggestion: again, maybe use "JIT" instead of "jit" - same as in https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit 

```suggestion
    "The problem is that we tried to condition on the *value* of an input to the function being JITted. The reason we can't do this is related to the fact mentioned above that jaxpr depends on the actual values used to trace it. \n",
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "\n",
+    "For `jax.jit`, the default level is `ShapedArray` -- that is, each tracer has a concrete shape (which we're allowed to condition on), but no concrete value. This allows the compiled function to work on all possible inputs with the same shape -- the standard use case in machine learning. However, because the tracers have no concrete value, if we attempt to condition on one, we get the error above.\n",
+    "\n",
+    "In `jax.grad`, the constraints are more relaxed, so you can do more. If you compose several transformations, however, you must satisfy the constraints of the most strict one. So, if you `jit(grad(f))`, `f` mustn't condition on value. For more detail on the interaction between Python control flow and JAX, see [this section](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Control-Flow) of the JAX docs.\n",
```
8bitmp3, line 344: Suggestion:

- a11y
- "this section... of the JAX docs" -> "this section... of 🔪 JAX - The Sharp Bits 🔪 (it's quite a popular guide AFAIU! 👏 

```suggestion
    "In `jax.grad`, the constraints are more relaxed, so you can do more. If you compose several transformations, however, you must satisfy the constraints of the most strict one. So, if you `jit(grad(f))`, `f` mustn't condition on value. For more detail on the interaction between Python control flow and JAX, check [this section](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Control-Flow) in 🔪 JAX - The Sharp Bits 🔪.\n",
```

(Emojis should be supported in notebooks 🤔)

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "g_inner_jitted(10, 20)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "5XUT2acoHBz-"
+   },
+   "source": [
+    "If we really need to jit a function that has a condition on the value of an input, we can tell JAX to help itself to a less abstract tracer for a particular input by specifing `static_argnums`. The cost of this is that the resulting jaxpr is less flexible, so JAX will have to re-compile the function for every new value of the specified input. It is only a good strategy if the function is guaranteed to get limited different values."
```
8bitmp3, line 392: Suggestion:

- Another one, based on https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit 
which you can totally ignore, if you have an established way to write JIT/`jit` 😆 

```suggestion
    "If we really need to JIT a function that has a condition on the value of an input, we can tell JAX to help itself to a less abstract tracer for a particular input by specifing `static_argnums`. The cost of this is that the resulting jaxpr is less flexible, so JAX will have to re-compile the function for every new value of the specified input. It is only a good strategy if the function is guaranteed to get limited different values."
```

8bitmp3, line 392: Suggestion: fix a small typo ("specifying")

```suggestion
    "If we really need to jit a function that has a condition on the value of an input, we can tell JAX to help itself to a less abstract tracer for a particular input by specifying `static_argnums`. The cost of this is that the resulting jaxpr is less flexible, so JAX will have to re-compile the function for every new value of the specified input. It is only a good strategy if the function is guaranteed to get limited different values."
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "print(jax.make_jaxpr(log2_if_rank_2)(jax.numpy.array([1, 2, 3])))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "Qp3WhqaqvHyD"
+   },
+   "source": [
+    "## Jitting a function\n",
```
8bitmp3, line 159: Suggestion: write "jit" as "JIT" (similar to https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit)


```suggestion
    "## JITting a function\n",
```

(as in Just-In-Time-ing... 🤔 )

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "3) We timed the execution speed of the compiled version. (Note the use of `block_until_ready()`, which is required due to JAX's [Asynchronous execution](https://jax.readthedocs.io/en/latest/async_dispatch.html) model)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "DRJ6R6-d9Q_U"
+   },
+   "source": [
+    "## Why can't we just jit everything?\n",
```
8bitmp3, line 251: Suggestion:

- "jit" -> "JIT" (similar to https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit ("To JIT or not to JIT")

```suggestion
    "## Why can't we just JIT everything?\n",
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "\n",
+    "In `jax.grad`, the constraints are more relaxed, so you can do more. If you compose several transformations, however, you must satisfy the constraints of the most strict one. So, if you `jit(grad(f))`, `f` mustn't condition on value. For more detail on the interaction between Python control flow and JAX, see [this section](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Control-Flow) of the JAX docs.\n",
+    "\n",
+    "One way to deal with this problem is to rewrite the code to avoid conditionals on value. Another is to use special [control flow operators](https://jax.readthedocs.io/en/latest/jax.lax.html#control-flow-operators) like `jax.lax.cond`. However, sometimes that is impossible. In that case, you can consider jitting only part of the function. For example, if the most computationally expensive part of the function is inside the loop, we can jit just that inner part (though make sure to check the next section on caching to avoid shooting yourself in the foot):"
```
8bitmp3, line 346: Suggestion:

- "JITting" since you wrote "to JIT" in https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit

```suggestion
    "One way to deal with this problem is to rewrite the code to avoid conditionals on value. Another is to use special [control flow operators](https://jax.readthedocs.io/en/latest/jax.lax.html#control-flow-operators) like `jax.lax.cond`. However, sometimes that is impossible. In that case, you can consider JITting only part of the function. For example, if the most computationally expensive part of the function is inside the loop, we can JIT just that inner part (though make sure to check the next section on caching to avoid shooting yourself in the foot):"
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+     },
+     "execution_count": 8,
+     "metadata": {
+      "tags": []
+     },
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# While loop conditioned on x and n with a jitted body.\n",
```
8bitmp3, line 371: Suggestion:

```suggestion
    "# While loop conditioned on x and n with a JITted body.\n",
```

https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "LczjIBt2X2Ms"
+   },
+   "source": [
+    "## When to use jit\n",
+    "\n",
+    "In many of the the examples above, jitting is not worth it:"
```
8bitmp3, line 445: Suggestion:

- Again, based on https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit

```suggestion
    "In many of the the examples above, JITting is not worth it:"
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "\n",
+    "Of course, impure functions can still be written and even run, but JAX gives no guarantees about their behaviour once converted to jaxpr. However, as a rule of thumb, you can expect (but shouldn't rely on) the side-effects of a JAX-transformed function to run once (during the first call), and never again. This is because of the way that JAX generates jaxpr, using a process called 'tracing'.\n",
+    "\n",
+    "When tracing, JAX wraps each argument by a *tracer* object. These tracers then record all JAX operations performed on them during the function call (which happens in regular python). Then, JAX uses the tracer records to reconstruct the entire function. The output of that reconstruction is the jaxpr. Since the tracers do not record the python side-effects, they do not appear in the jaxpr. However, the side-effects still happen during the trace itself.\n",
```
8bitmp3, line 67: Suggestion: nit ("python" 🐍  -> "Python")

```suggestion
    "When tracing, JAX wraps each argument by a *tracer* object. These tracers then record all JAX operations performed on them during the function call (which happens in regular Python). Then, JAX uses the tracer records to reconstruct the entire function. The output of that reconstruction is the jaxpr. Since the tracers do not record the Python side-effects, they do not appear in the jaxpr. However, the side-effects still happen during the trace itself.\n",
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "\n",
+    "Note: the python `print()` function is not pure: the text output is a side-effect of the function. Therefore, any `print()` calls will only happen during tracing, and will not appear in the jaxpr:"
```
8bitmp3, line 69: Suggestion: nit ("python" 🐍 -> "Python")

```suggestion
    "Note: the Python `print()` function is not pure: the text output is a side-effect of the function. Therefore, any `print()` calls will only happen during tracing, and will not appear in the jaxpr:"
```

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "f6W_YYwRRwGp"
+   },
+   "source": [
+    "See how the printed `x` is a `Traced` object? That's the JAX internals at work.\n",
+    "\n",
+    "The fact that the python code runs at least once is strictly an implementation detail, and so shouldn't be relied upon. However, it's useful to understand as you can use it when debugging to print out intermediate values of a computation."
```
8bitmp3, line 111: Suggestion: nit ("python" 🐍 -> "Python")

```suggestion
    "The fact that the Python code runs at least once is strictly an implementation detail, and so shouldn't be relied upon. However, it's useful to understand as you can use it when debugging to print out intermediate values of a computation."
```

8bitmp3 (commented):
I scanned through the Jupyter notebook only and left a few suggestions. Hope you don't mind @jakevdp 👍 Great doc, I'm still learning this JAX thing so this helps 💯.

Regarding whether to write "jit", "JIT" or `jit`, I initially used the Thinking in Jax doc as a reference (https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit) where you say "to JIT" (from Just-In-Time) 🤷‍♂️  but I'm sure the team has an established way of writing `jax.jit`. And I think "JITting" doesn't look as elegant as "jitting" 🤔. (UPDATE: found an answer in JAX-101 part 3: _"...you can wrap a vmapped function with JIT, or a JITted function with vmap, and everything will work correctly:_")

Re: SELU - for non ML folks, I suggested to provide a quick definition and an "arXiv"/paper link (to minimize googling):

```
1) We defined `selu_jit` as the compiled version of `selu`
([scaled exponential linear unit](https://proceedings.neurips.cc/paper/6698-self-normalizing-neural-networks.pdf)).
```

I've also proposed to spell out the name of the one and only 🔪 JAX - The Sharp Bits 🔪 guide, instead of referring to it as "JAX docs":

```
So, if you `jit(grad(f))`, `f` mustn't condition on value. For more detail on the interaction 
between Python control flow and JAX, check 
[this section](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Control-Flow)
in 🔪 JAX - The Sharp Bits 🔪.
```

Thanks again for teaching this stuff 👍

docs/jax-101/02-jitting.ipynb
```diff
@@ -0,0 +1,570 @@
...
+    "id": "O-SkdlPxvETZ"
+   },
+   "source": [
+    "# How does JAX work?\n",
+    "\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/jax/blob/master/docs/jax-101/02-jitting.ipynb)\n",
+    "\n",
+    "*Authors: TODO*\n",
+    "\n",
+    "In this section, we will take a further look at how JAX works, and how we can make it performant. In the previous lesson, we discussed that JAX allows us to transform Python functions. This is done by first converting the Python function into a simple intermediate language called jaxpr. The transformations then work on the jaxpr representation. \n",
```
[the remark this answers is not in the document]
jakevdp: The problem with the link is that there's no way to know the context: if someone is reading the doc on the website, linking to the notebook is the wrong thing to do. If someone is reading the notebook, linking to the website is the wrong thing to do. In the absence of a good option, I chose no link for now.

jakevdp:
Thanks @8bitmp3 for the comments! I've updated the notebook to reflect all these changes.

skye:
nit: the commit message says this is the third installment, even though it's the second (I'm guessing this was an accident of bash history).

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+
+g_jit_correct = jax.jit(g, static_argnums=1)
+print(g_jit_correct(10, 20))
+```
+
++++ {"id": "LczjIBt2X2Ms"}
+
+## When to use JIT
+
+In many of the the examples above, JITting is not worth it:
```
skye, line 243: And here

skye (approved):
Just a few minor comments, take 'em or leave 'em. Thanks!

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+    format_version: 0.13
+    jupytext_version: 1.10.0
+kernelspec:
+  display_name: Python 3
+  name: python3
+---
+
++++ {"id": "O-SkdlPxvETZ"}
+
+# How does JAX work?
```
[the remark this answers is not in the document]
skye: "Just-In-Time compiling with JAX" ?

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+  return ln_x / ln_2
+
+print(jax.make_jaxpr(log2)(3.0))
+```
+
++++ {"id": "jiDsT7y0RwIp"}
+
+The [Understanding Jaxprs](https://jax.readthedocs.io/en/latest/jaxpr.html) section of the documentation provides more information on the meaning of the above output.
+
+Importantly, note how the jaxpr does not capture the side-effect of the function: there is nothing in it corresponding to `global_list.append(x)`. This is a feature, not a bug: JAX is designed to understand side-effect-free (a.k.a. functionally pure) code. If *pure function* and *side-effect* are unfamiliar terms, this is explained in a [little more detail](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-Pure-functions) in the JAX docs.
```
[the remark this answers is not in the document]
8bitmp3: ... @skye the number is 4? (2 x 2) 😄

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+    i += 1
+  return x + i
+
+g_jit = jax.jit(g)
+g_jit(10, 20)  # Should raise an error. 
+```
+
++++ {"id": "isz2U_XX_wH2"}
+
+The problem is that we tried to condition on the *value* of an input to the function being JITted. The reason we can't do this is related to the fact mentioned above that jaxpr depends on the actual values used to trace it.
```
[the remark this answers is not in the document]
8bitmp3: @skye I mentioned to @jakevdp in an earlier suggestion that in _Thinking in JAX_ you use "to JIT" for the acronym used as a verb https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit 

Meanwhile, the Chicago Manual of Style and the AP Stylebook (according to https://www.quickanddirtytips.com/education/grammar/do-you-use-apostrophe-make-“ok”-past-tense) would likely write this as "JIT'd" (they apparently recommend "OK'd" for "to OK").

I'm a follower of good style guides, like the Google/TensorFlow/Microsoft ones (which provide guidelines, not rules, as they say), so I'd probably go with "JIT'd" but it does look weird, as @skye pointed out. However, an acronym is an acronym. `jit`-ed? I don't know.

image

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+
+(If we didn't include the warm-up call separately, everything would still work, but then the compilation time would be included in the benchmark. It would still be faster, because we run many loops in the benchmark, but it wouldn't be a fair comparison.)
+
+3) We timed the execution speed of the compiled version. (Note the use of `block_until_ready()`, which is required due to JAX's [Asynchronous execution](https://jax.readthedocs.io/en/latest/async_dispatch.html) model).
+
++++ {"id": "DRJ6R6-d9Q_U"}
+
+## Why can't we just JIT everything?
+
+After going through the example above, you might be wondering whether we should simply apply `jax.jit` to every function. To understand why this is not the case, and when we should/shouldn't apply `jit`, let's first check some cases where jit doesn't work.
```
8bitmp3, line 152: Suggestion: nit

```suggestion
After going through the example above, you might be wondering whether we should simply apply `jax.jit` to every function. To understand why this is not the case, and when we should/shouldn't apply `jit`, let's first check some cases where JIT doesn't work.
```

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+selu_jit(x).block_until_ready()
+
+%timeit selu_jit(x).block_until_ready()
+```
+
++++ {"id": "hMNKi1mYXQg5"}
+
+Here's what just happened:
+
+1) We defined `selu_jit` as the compiled version of `selu`
```
8bitmp3, line 140: Suggestion: (a full stop maybe)

```suggestion
1) We defined `selu_jit` as the compiled version of `selu`.
```

docs/jax-101/02-jitting.md
```diff
@@ -0,0 +1,293 @@
...
+
+x = jnp.arange(1000000)
+%timeit selu(x).block_until_ready()
+```
+
++++ {"id": "ecN5lEXe6ncy"}
+
+The code above is sending one operation at a time to the accelerator. This limits the ability of the XLA compiler to optimize our functions.
+
+Naturally, what we want to do is give the XLA compiler as much code as possible, so it can fully optimize it. For this purpose, JAX provides the `jax.jit` transformation, which will Just In Time (JIT) compile a JAX-compatible function. The example below shows how to use JIT to speed up the previous function.
```
8bitmp3, line 122: Suggestion:

- Similar to _"Every JAX operation is eventually expressed in terms of these fundamental XLA operations, which is what enables just-in-time (JIT) compilation."_ in https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html

```suggestion
Naturally, what we want to do is give the XLA compiler as much code as possible, so it can fully optimize it. For this purpose, JAX provides the `jax.jit` transformation, which will Just-In-Time (JIT) compile a JAX-compatible function. The example below shows how to use JIT to speed up the previous function.
```

🤷  cc @skye

docs/jax-101/02-jitting.md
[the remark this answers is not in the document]
8bitmp3: Similar to https://github.com/google/jax/pull/5905#discussion_r585944171 (comment for the notebook version @skye )

```diff
- in the JAX docs.
+ in 🔪 JAX - The Sharp Bits 🔪.
```

docs/jax-101/02-jitting.md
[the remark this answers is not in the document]
skye: Ok no joke, I was gonna suggest "JIT'd" but thought it was maybe a little much! But if two English style guides recommend it... I'll let Jake make the final call :) I'm fine with anything.

docs/jax-101/02-jitting.md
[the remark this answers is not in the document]
8bitmp3: @skye I noticed that there's a _"Using `jax.jit`"_-like section title in the [JAX QuickStart](https://jax.readthedocs.io/en/latest/notebooks/quickstart.html#using-jit-to-speed-up-functions) called _"Using `jit()` to speed up functions"_. 

Also, that JAX QuickStart refers to all the cool APIs with round brackets (`jit()`), which is consistent with the official [Style Guide for Python](https://www.python.org/dev/peps/pep-0008/) (e.g. _"For example, the os.stat() function returns a tuple"_).

docs/jax-101/02-jitting.md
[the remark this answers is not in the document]
jakevdp: We just have to decide which way to do it... my first round of revisions this afternoon was to change all the "jitted" to "JITted".

docs/jax-101/02-jitting.md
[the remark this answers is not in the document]
jakevdp: That comes from using RST directives to refer to functions. We could do that here (e.g. ```{func}`jax.jit` ``` in markdown) and it would render the text and links correctly in the HTML docs, but you'd still see the raw markdown in the notebooks.

It's another case of difficulty introduced by the fact that we want this single source to operate in two ways: as a viewable notebook, and as a built documentation page.

jakevdp:
I'm going to merge with authorship left as a TODO.

PR #5908, opened 2021-03-03
DOC: add transformations doc to HTML & reorganize contents

jakevdp:
This seems to fit better in the actual documentation, where the code can be doctested, which actually caught some existing mistakes.

This is a first-step toward making the github README a bit lighter and making the HTML docs a bit more organized and useful. As part of that, I re-organized our documentation to separate API docs from the advanced tutorials. I think some further reorganization based on this will be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,7 @@
 
 ## What is JAX?
 
-JAX is [Autograd](https://github.com/hips/autograd) and
-[XLA](https://www.tensorflow.org/xla),
+JAX is [Autograd](https://github.com/hips/autograd) and [XLA](https://www.tensorflow.org/xla),
 brought together for high-performance machine learning research.
 
 With its updated version of [Autograd](https://github.com/hips/autograd),
@@ -32,8 +31,7 @@ derivatives. It supports reverse-mode differentiation (a.k.a. backpropagation)
 via [`grad`](#automatic-differentiation-with-grad) as well as forward-mode differentiation,
 and the two can be composed arbitrarily to any order.
 
-What’s new is that JAX uses
-[XLA](https://www.tensorflow.org/xla)
+What’s new is that JAX uses [XLA](https://www.tensorflow.org/xla)
 to compile and run your NumPy programs on GPUs and TPUs. Compilation happens
 under the hood by default, with library calls getting just-in-time compiled and
 executed. But JAX also lets you just-in-time compile your own Python functions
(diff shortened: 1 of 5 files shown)


docs/transformations.md
```diff
@@ -0,0 +1,280 @@
...
+For parallel programming of multiple accelerators, like multiple GPUs, use
+{func}`jax.pmap`.
+With {func}`pmap` you write single-program multiple-data (SPMD) programs, including
+fast parallel collective communication operations. Applying {func}`pmap` will mean
+that the function you write is compiled by XLA (similarly to {func}`jit`), then
+replicated and executed in parallel across devices.
+
+Here's an example on an 8-GPU machine:
+
+```python
```
jakevdp, line 217: Note `python` here rather than `{code-cell}` which means that this code does not get executed or tested when the docs are built. The RTD build is on a single-core machine, so pmap fails.

Is there any way we could emulate a multi-core machine to test pmap in that environment?

jakevdp: Managed to do it using `--xla_force_host_platform_device_count=8` in a hidden setup cell.

docs/index.rst
```diff
@@ -1,11 +1,9 @@
 JAX reference documentation
 ===========================
 
-Composable transformations of Python+NumPy programs: differentiate, vectorize,
-JIT to GPU/TPU, and more.
-
-For an introduction to JAX, start at the
-`JAX GitHub page <https://github.com/google/jax>`_.
+JAX is Autograd_ and XLA_, brought together for high-performance machine learning research.
```
8bitmp3, line 4: Silly question: maybe "numerical computing research" > "machine learning computing research"? I saw the former somewhere maybe in the earlier JAX docs 🤔 It may also be more inclusive, given the use cases popping up potentially even outside of "ML for X".

jakevdp: Done!

README.md
```diff
@@ -220,14 +218,14 @@ function:
 ```python
 def predict(params, input_vec):
   assert input_vec.ndim == 1
-  activations = inputs
+  activations = input_vec
   for W, b in params:
     outputs = jnp.dot(W, activations) + b  # `input_vec` on the right-hand side!
```
mattjj, line 223:
```suggestion
    outputs = jnp.dot(W, activations) + b  # `activations` on the right-hand side!
```

README.md
```diff
@@ -220,14 +218,14 @@ function:
...
     activations = jnp.tanh(outputs)
   return outputs
 ```
 
-We often instead write `jnp.dot(inputs, W)` to allow for a batch dimension on the
+We often instead write `jnp.dot(activations, W)` to allow for a batch dimension on the
 left side of `inputs`, but we’ve written this particular prediction function to
```
mattjj, line 229:
```suggestion
left side of `activations`, but we’ve written this particular prediction function to
```

docs/index.rst
```diff
@@ -77,3 +80,7 @@ Indices and tables
 * :ref:`genindex`
 * :ref:`modindex`
 * :ref:`search`
+
+
+.. _Autograd: https://github.com/hips/autograd)
```
mattjj, line 85:
```suggestion
.. _Autograd: https://github.com/hips/autograd
```

mattjj (approved):
LGTM!

docs/index.rst
jakevdp: Done!

PR #5917, opened 2021-03-03
Use None as a trivial lazy expression, rather than lazy.array().

hawkinsp:
Improves a 1000-argument version of the jit_simple_many_args by around 20%.

It is faster for C++ to pattern-match a None than to pattern-match a complicated Python data structure. And we know that `None` should be handled correctly, because the C++ buffer object reports `None` as its lazy expression already.

```
Before:
name                       cpu/op
jit_simple_many_args_1000  3.12ms ± 3%

After:
name                       cpu/op
jit_simple_many_args_1000  2.60ms ± 5%
```


mattjj (approved):
Smart!

PR #5919, opened 2021-03-03
Make host_callback_test.py safe to run in environments with > 2 devices.

skye:
For example, Cloud TPU VMs expose 8 local devices.

This is achieved by only using `jax.devices()[:2]` with `pmap`, etc. since the test cases assume no more than 2 devices will be used.


gnecula (approved):
Thanks

PR #5928, opened 2021-03-04
Only vectorize xmap axes that have only one element per resource

apaszke:
To make the jaxpr much less noisy. size-1 vmaps are quite pointless.


mattjj (approved):
Good idea!

jekbradbury (approved):

PR #5936, opened 2021-03-04
Fix batching formula of xmap

apaszke:
Turns out that once you insert multiple dimensions things become much
more tricky than in the case of batching a one-dimensional map. Also
strenghten our tests to make sure we don't depend too much on the
semantics of the einsum batching rule.


mattjj (approved):
LGTM, though a test would help us sleep better at night!

apaszke:
Agreed, just wanted to unblock @hawkinsp. Added the test now (and verified that it would catch this issue on master).

PR #5943, opened 2021-03-04
Update suggested `pip install` command in build_wheel.py

skye:
Adding `--force-reinstall --no-deps` makes pip always install the new jaxlib, even if a jaxlib with the same version is already installed. This change will save me many seconds of typing in the future.


skye:
This is actually a bad idea, because `--no-deps` is not what you want if the deps aren't already installed, and `--force-reinstall` by itself is annoying because it reinstalls all the deps too. I'm gonna close this and continue wasting valuable seconds typing.

Issue #5706, opened 2021-02-11
JAX TypeError disagrees with documentation?


dfarmer:
The documentation for stable (and latest of course), https://jax.readthedocs.io/en/stable/_autosummary/jax.numpy.array.html says:

```
>>> x = np.array([(1,2),(3,4)],dtype=[('a','<i4'),('b','<i4')])
>>> x['a']
array([1, 3])
```

but when I try to do that in my local jupyterlab notebook (with JAX v0.2.9)  I get 

TypeError: JAX only supports number and bool dtypes, got dtype [('a', '<i4'), ('b', '<i4')] in array

Is the documentation wrong?

mattjj:
The documentation has lied to you!

JAX docstrings include the original NumPy docstrings. There's a critical sentence in there, "Original docstring below.", which indicates that it's just a reproduction of the NumPy docstring. Perhaps it should be clearer that the NumPy docstring may describe features that aren't available in JAX...

@shoyer @jakevdp any ideas?

hawkinsp:
I think in some cases where our behavior differs in significant ways, we should perhaps not simply reference the numpy strings. In this case I think `jnp.array` is significantly different that we should consider it...

dfarmer:
I'm not sure how much work it would be but even if y'all could just get the numpy doctests to actually run instead of having the copy-pasted version then it would at least be honest (your documentation might end up riddled with exceptions -- I don't know). 

Anyway, thanks for clearing that up -- that's useful for interpreting anything I reference on there.

jakevdp:
Fixed by https://github.com/google/jax/pull/5945; docstrings now remove all examples.

Issue #5539, opened 2021-01-28
update_numpydoc is too brittle


jakevdp:
There are docstrings in numpy/scipy that are incorrect; for example `scipy.special.erfinv`. Compare the scipy doc:
```python
In [1]: from scipy.special import erfinv; print(erfinv.__doc__)                                                                                                                                                                                                                         
erfinv(x, /, out=None, *, where=True, casting='same_kind', order='K', dtype=None, subok=True[, signature, extobj])

Inverse of the error function.

    Computes the inverse of the error function.

    In the complex domain, there is no unique complex number w satisfying
    erf(w)=z. This indicates a true inverse function would have multi-value.
    When the domain restricts to the real, -1 < x < 1, there is a unique real
    number satisfying erf(erfinv(x)) = x.

    Parameters
    ----------
    y : ndarray
        Argument at which to evaluate. Domain: [-1, 1]

    Returns
    -------
    erfinv : ndarray
        The inverse of erf of y, element-wise)
```
to the wrapped jax doc:
```python
In [2]: from jax.scipy.special import erfinv; print(erfinv.__doc__)                                                                                                                                                                                                                     
Inverse of the error function.

LAX-backend implementation of :func:`erfinv`.
Original docstring below.

erfinv(x, /, out=None, *, where=True, casting='same_kind', order='K', dtype=None, subok=True[, signature, extobj])

    Computes the inverse of the error function.

    In the complex domain, there is no unique complex number w satisfying
    erf(w)=z. This indicates a true inverse function would have multi-value.
    When the domain restricts to the real, -1 < x < 1, there is a unique real
    number satisfying erf(erfinv(x)) = x.

    Parameters
    ----------
  Returns
    -------
    erfinv : ndarray
        The inverse of erf of y, element-wise)
```
Perhaps we should limit the parameter removal to non-required parameters?

jakevdp:
Sure - the issue is in this line in the JAX documentation wrapper: https://github.com/google/jax/blob/3af76ff013c924b7aac0aa83508f928cbb9bee54/jax/_src/numpy/util.py#L130-L131

This only keeps parameter descriptions for parameters that appear in the function's argument list. Because of this, if functions list the wrong parameter name (as is the case in `scipy.special.erfinv`) the description is dropped.

I'm not sure what is the best way to fix this.

a-ma-n:
@jakevdp I understand the issue but why exactly would a function return the wrong name?
And if the function returns the correct name we would not have to worry about the docstring.
[...]

Issue #4199, opened 2020-09-03
Inaccuracy of jax.numpy.sum


nickmcgreivy:
So when I run this code:
```python
import numpy as np
import jax.numpy as jnp
from jax.config import config
config.update("jax_enable_x64", True)

N = int(1e7)
matrix = np.asarray([np.zeros(N), np.zeros(N), np.ones(N) / N])
jmatrix = jnp.asarray([jnp.zeros(N), jnp.zeros(N), jnp.ones(N) / N])

print(np.sum(matrix) - 1.0)
print(jnp.sum(jmatrix) - 1.0)
print(np.sum(matrix[2, :]) - 1.0)
print(jnp.sum(jmatrix[2, :]) - 1.0)
print(np.sum(matrix, axis=1)[2] - 1.0)
print(jnp.sum(jmatrix, axis=1)[2] - 1.0)
```
I get
```
1.7319479184152442e-14
1.2276624161700056e-11
1.7319479184152442e-14
-9.992007221626409e-16
1.7319479184152442e-14
-9.992007221626409e-16
```
It seems this has to do with [partial pairwise summation](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.sum.html). I see #3517, but that appears to be a slightly different issue. I have two questions:

1. The docs say 

> This improved precision is always provided when no axis is given. 

But for `jnp`, if I give no axis, the precision is terrible (1e-11). If I give the axis, or I map over a vector, Jax does better. Do you agree the docs seem to contradict this example?

2. Is there some way I can predict when these errors might arise in advance? I'd love to be able to use `jnp.sum` without worrying about the whether the result will be sufficiently numerically precise.

jakevdp:
Hi – note that the details of the docstring at the link you provided have to do with numpy, not JAX (note at the top where it says "Original docstring below"), and so those details don't actually apply to JAX's implementation .

The precise order of floating point operations performed by JAX will generally depend on the backend implementation, whether the code is JIT-compiled or not, whether you're running code on a CPU, GPU, TPU, etc.

Others may be able to chime in with more detailed information about what precision you should expect in each situation.

nickmcgreivy:
Ah! So this isn't an error in the docs. Thanks for clarifying.

I'm running this on my CPU for reference. Running it on a GPU, I get 

1.7319479184152442e-14
1.7319479184152442e-14
1.7319479184152442e-14
1.7319479184152442e-14
1.7319479184152442e-14
[...]

PR #5945, opened 2021-03-04
Remove extraneous information from wrapped docstrings

jakevdp:
This is an overhaul of our numpy docstring rewriting code.

The main issue with the current approach is that we include the full numpy docstring in the docstring of wrapped functions; via sphinx autodoc this gets propagated to our HTML documentation. This is problematic because the numpy examples don't always work with their associated JAX functions; for example the [`jnp.sum` docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.sum.html) include this set of examples:
```python
>>> np.sum([0.5, 1.5])
2.0
>>> np.sum([0.5, 0.7, 0.2, 1.5], dtype=np.int32)
1
>>> np.sum([[0, 1], [0, 5]])
6
>>> np.sum([[0, 1], [0, 5]], axis=0)
array([0, 6])
>>> np.sum([[0, 1], [0, 5]], axis=1)
array([1, 5])
>>> np.sum([[0, 1], [np.nan, 5]], where=[False, True], axis=1)
array([1., 5.])
```
All of these error with `jnp.sum()`, because lists are not valid inputs to JAX reductions. This has caused some confusion for users in the past (e.g. #4199, #5706)

The fix here is to remove all sections except the opening summary, `Parameters`, `Returns`, and `References`. This may remove relevant information from some docstrings, but I think that's preferable to having incorrect information in the documentation.

Note that the sphinx documentation build provides a good set of tests that validate the correctness of the rewritten docstrings: if any are misformatted, the documentation build will fail.

Related to #5539

### DIFF
--- a/jax/_src/numpy/util.py
+++ b/jax/_src/numpy/util.py
@@ -12,104 +12,143 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import OrderedDict
 import re
 import textwrap
+from typing import Callable, NamedTuple, Optional, Dict, Sequence
 
 _parameter_break = re.compile("\n(?=[A-Za-z_])")
+_section_break = re.compile(r"\n(?=[^\n]{3,15}\n-{3,15})", re.MULTILINE)
+_numpy_signature_re = re.compile(r'^([\w., ]+=)?\s*[\w\.]+\([\w\W]*?\)$', re.MULTILINE)
 
-def update_numpydoc(docstr, fun, op):
-  '''Transforms the numpy docstring to remove references of
-     parameters that are supported by the numpy version but not the JAX version'''
-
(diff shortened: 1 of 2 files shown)


jakevdp:
This is failing in Python 3.6 because the ability to split a string on an empty regex was added in Python 3.7: https://docs.python.org/3/library/re.html#re.split

mattjj (approved):
Wow, phenomenal! This is a great idea.

mattjj:
Should we prune `out` from `Parameters`? And/or mention in our docstring prefix that not all NumPy features are supported, and give `out` as an example?

jakevdp:
> Should we prune out from Parameters?

That would happen automatically, except we include `out` in the signature of JAX reductions (erroring if it's anything other than `None`). If we left it out of the signature, then the current approach would leave it out of the parameter list as well.

PR #5951, opened 2021-03-05
Revive all to all

apaszke:
#### Make all_to_all primitive match XLA semantics

This has the benefit of limiting the insane axis arithmetic (with some
axes getting removed, and others introduced with their positions offset
by the removals) to the all_to_all user-facing function, but all the
collective rules should now be simpler to write. This should be a no-op
from the point of view of the users, but should make enabling all_to_all
splitting easier.


mattjj (approved):
This was tricky!

PR #5952, opened 2021-03-05
Raise OverflowError in X32 mode for large Python integers

jakevdp:
Python integers are arbitrary width, meaning that they can represent values that are not representable by numpy's `int64`. In these cases, numpy raises an overflow error. For example:
```python
>>> import numpy as np
>>> np.array(1 << 65, dtype=int)
OverflowError                             Traceback (most recent call last)
<ipython-input-36-bc194570c3ad> in <module>
----> 1 np.array(1 << 65, dtype=int)

OverflowError: Python int too large to convert to C long
```
In X64 mode, JAX inherits this property via numpy:
```python
>>> import jax.numpy as jnp
>>> from jax import config; config.update('jax_enable_x64', True)
>>> np.array(1 << 65, dtype=int)
OverflowError                             Traceback (most recent call last)
<ipython-input-3-e4d8a251ed7e> in <module>
----> 1 jnp.array(1 << 65, dtype=int)

~/github/google/jax/jax/_src/numpy/lax_numpy.py in array(object, dtype, copy, order, ndmin)
   2874 
   2875   if _can_call_numpy_array(object):
-> 2876     object = _np_array(object, dtype=dtype, ndmin=ndmin, copy=False)
   2877 
   2878   assert type(object) not in dtypes.python_scalar_dtypes

~/github/google/jax/jax/_src/numpy/lax_numpy.py in _np_array(obj, dtype, **kwargs)
    228   uses Jax's default dtypes.
    229   """
--> 230   arr = np.array(obj, dtype=dtype, **kwargs)
    231   obj_dtype = getattr(obj, 'dtype', None)
    232   arr_dtype = np.dtype(arr.dtype).type

OverflowError: Python int too large to convert to C long
```
With X64 disabled, however, values that are too large to be stored in `int32` are silently truncated to zero:
```python
>>> import jax.numpy as jnp
>>> jnp.array(1 << 33)
DeviceArray(0, dtype=int32)
```
Such silent data loss is problematic, particularly because X32 is the default mode.

This PR adds a value check when Python integers are coerced to arrays, so that an `OverflowError` is raised as appropriate for the default integer type:
```python
>>> import jax.numpy as jnp
>>> jnp.array(1 << 33)
OverflowError                             Traceback (most recent call last)
<ipython-input-2-50cb794a933c> in <module>
----> 1 jnp.array(1 << 33)

~/github/google/jax/jax/_src/numpy/lax_numpy.py in array(object, dtype, copy, order, ndmin)
   2872 
   2873   if _can_call_numpy_array(object):
-> 2874     object = dtypes.coerce_to_array(object, dtype=dtype, ndmin=ndmin, copy=False)
   2875 
   2876   assert type(object) not in dtypes.python_scalar_dtypes

~/github/google/jax/jax/dtypes.py in coerce_to_array(x, dtype, ndmin, copy)
    125     info = np.iinfo(dtype)
    126     if not info.min <= x <= info.max:
--> 127       raise OverflowError(f"Python int {x} too large to convert to {dtype}")
    128   return np.array(x, dtype=dtype, ndmin=ndmin, copy=copy)
    129 

OverflowError: Python int 8589934592 too large to convert to int32
```
This also required a minor change to `random.PRNGseed` as well, as the previous behavior was built on the assumption that out-of-bounds inputs were truncated rather than leading to an error.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2873,7 +2873,7 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
   dtype = dtype and dtypes.canonicalize_dtype(dtype)
 
   if _can_call_numpy_array(object):
-    object = _np_array(object, dtype=dtype, ndmin=ndmin, copy=False)
+    object = dtypes.coerce_to_array(object, dtype=dtype, ndmin=ndmin, copy=False)
 
   assert type(object) not in dtypes.python_scalar_dtypes
 
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -55,11 +55,10 @@ def PRNGKey(seed: int) -> jnp.ndarray:
   if not np.issubdtype(np.result_type(seed), np.integer):
     raise TypeError(f"PRNGKey seed must be an integer; got {seed!r}")
 
-  # Explicitly cast to int64 for JIT invariance of behavior on large ints.
+  # Explicitly cast Python integers to np.int64 to prevent overflow error for
+  # large Python ints in X32 mode.
   if isinstance(seed, int):
     seed = np.int64(seed)  # type: ignore[assignment]
-  # Converting to jnp.array may truncate bits when jax_enable_x64=False, but this
-  # is necessary for the sake of JIT invariance of the result for such values.
   seed = jnp.asarray(seed)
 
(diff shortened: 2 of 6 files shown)


jakevdp:
@hawkinsp – I requested your review even though this still has a number of downstream impacts. If this change is something we want, I can start fixing/mitigating those. Thanks!

jakevdp:
Replaced by #6269

PR #5953, opened 2021-03-05
move changelog to top level

froystig:


docs/changelog.rst
```diff
@@ -0,0 +1 @@
+.. include:: ../CHANGELOG.md
```
jakevdp, line 1: I think this will fail because it is a literal include of markdown syntax in an RST document. You may need to rename it to `changelog.md` and use the markdown version of the syntax:

    ```{include} ../CHANGELOG.md
    ```

CHANGELOG.md
```diff
@@ -1,3 +1,596 @@
-# Change Log
+# Change log
+
+Also rendered [here](https://jax.readthedocs.io/en/latest/CHANGELOG.html).
```
jakevdp, line 3: I think URLs are case-sensitive, so this would now be `changelog.html`

mattjj (approved):

PR #5962, opened 2021-03-06
Automatically initialize Cloud TPU topology env vars if running on a Cloud TPU VM.

skye:
This removes the need to manually set these env vars when running on a Cloud TPU pod slice.

I manually tested this on a v2-8 and v2-32.

### DIFF
--- a/jax/_src/cloud_tpu_init.py
+++ b/jax/_src/cloud_tpu_init.py
@@ -1,0 +1,89 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+def cloud_tpu_init():
+  """Automatically sets Cloud TPU topology env vars.
+
+  **This must be called before the TPU runtime is loaded, which happens as soon
+  as JAX's C++ backend is loaded! I.e. call this before xla_bridge or xla_client
+  is imported.**
+
+  These environment variables are used to tell the TPU runtime what kind of mesh
+  topology to use. It assumes a single-host topology by default, so we manually
+  set them here to default to the full pod slice if applicable.
+
+  This will not set any env vars if a single topology-related env var is already
+  set.
+  """
+
+  if not _running_in_cloud_tpu_vm():
+    return
+
+  # If the user has set any topology-related env vars, don't set any
+  # automatically.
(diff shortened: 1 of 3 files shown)


jax/_src/cloud_tpu_init.py
```diff
@@ -0,0 +1,86 @@
...
+      'v3-512': '8,8,1',
+      'v3-1024': '8,16,1',
+      'v3-2048': '16,16,1',
+  }
+
+  os.environ['CLOUD_TPU_TASK_ID'] = worker_id
+  os.environ['TPU_CHIPS_PER_HOST_BOUNDS'] = '2,2,1'
+  os.environ['TPU_HOST_BOUNDS'] = accelerator_type_to_host_bounds[
+      accelerator_type]
+  os.environ['TPU_MESH_CONTROLLER_ADDRESS'] = worker_network_endpoints.split(
```
jekbradbury, line 80: Any particular reason why the address includes the port but we also need to specify the port separately?

jax/_src/cloud_tpu_init.py
```diff
@@ -0,0 +1,86 @@
...
+  set them here to default to the full pod slice if applicable.
+
+  This will not set any env vars if a single topology-related env var is already
+  set.
+  """
+
+  if not _running_in_cloud_tpu_vm():
+    return
+
+  # If the user has set any topology-related env vars, don't set any
```
jekbradbury, line 35: Strictly speaking, `CLOUD_TPU_TASK_ID` is one of these too (maybe it should be `TPU_HOST_ID` for consistency with the others, but it predates them). I don't think it makes any sense to set that manually without also setting either chips per host or visible devices, so it probably doesn't matter.

jekbradbury (approved):

jax/_src/cloud_tpu_init.py
skye: I don't know, but inside libtpu they're separate flags you pass to the runtime, so I guess it likes to have both.

jax/_src/cloud_tpu_init.py
skye: I'll add it just in case, I just forgot it here, good catch. Agreed it should be TPU_HOST_ID but that probably won't be fixed for a while.

skye:
@hawkinsp does doing something like this in __init__ look ok to you, and did I get the _src idiom right?

jax/_src/cloud_tpu_init.py
```diff
@@ -0,0 +1,88 @@
...
+  ]):
+    return
+
+  # pylint: disable=import-outside-toplevel
+  # Don't assume non-Cloud TPU environments have requests installed
+  import requests
+  # pylint: enable=import-outside-toplevel
+  def get_metadata(key):
+    return requests.get(
+        'http://metadata.google.internal/computeMetadata'
```
jekbradbury, line 53: This URL is configurable in TF-TPU because Shawn wanted to do something weird; maybe we could recognize the same env var: https://github.com/tensorflow/tensorflow/pull/40317

skye: Done

skye:
I talked to Peter offline and he's cool with this.

Issue #5842, opened 2021-02-25
where argument for jax.numpy.{mean, var, std, all, any}


n2cholas:
As of NumPy 1.20.0 `numpy.{mean, var, std, all, any}` support the `where` keyword argument ([release notes](https://numpy.org/doc/stable/release/1.20.0-notes.html#where-keyword-argument-for-numpy-all-and-numpy-any-functions)). `jax.numpy.{sum, max, min, ...}` already support this (the NumPy versions had it since 1.17.0). For API consistency and parity with NumPy, I think this argument should be implemented. Thanks!

jakevdp:
Great idea! This should be pretty straightforward; compare the definition of `min` and `all`: https://github.com/google/jax/blob/bb51a58084caa1537ba7970fb5669f5fe6fe69be/jax/_src/numpy/lax_numpy.py#L2007-L2018

Then for tests, we'd need to remove this condition: https://github.com/google/jax/blob/bb51a58084caa1537ba7970fb5669f5fe6fe69be/tests/lax_numpy_test.py#L832-L835

Although we would still need this skip for older numpy versions (our CI tests as far back as version 1.16). There's an example of skipping based on numpy version here: https://github.com/google/jax/blob/bb51a58084caa1537ba7970fb5669f5fe6fe69be/tests/lax_numpy_test.py#L1500-L1501

mtsokol:
Hi! I would like to try solving it as a first issue if it's still free to take 🙂

jakevdp:
@mtsokol - this is indeed still free to take. Please let me know if you have any questions, and feel free to open a work in progress pull request if that would be convenient!

PR #5966, opened 2021-03-06
Added 'where' keyword to 'jnp.{mean, var, std}'

mtsokol:
Addresses #5842 

Hi @jakevdp!

Here's WiP PR for adding `where` keyword support to `jnp.{mean, var, std}` as added in numpy `1.20`.

While working on it I encountered a bug in numpy's `np.mean`: support for `where` keyword when `axis` param is present results in internal error. I filed an issue: https://github.com/numpy/numpy/issues/18552 and fix PR: https://github.com/numpy/numpy/pull/18560 and I hope it will be included in `1.20.2` but for now this is blocked (impressive `jax`s test case generators cached it!).

I followed numpy API when adding `where` keyword (also its source code). I also added a separate test suite as those methods do not take `initial` argument (as they have their own identity defined). Also when generating those `where` masks it often masks the whole row that results in division by zero (but is legal operation, as in numpy) so I included additional warning suppressors. Is it correct to do?

Thank you for any help!

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -351,6 +351,17 @@ JAX_REDUCER_INITIAL_RECORDS = [
     op_record("min", 1, all_dtypes, all_shapes, jtu.rand_default, []),
 ]
 
+JAX_REDUCER_WHERE_NO_INITIAL_RECORDS = [
+    op_record("all", 1, bool_dtypes, all_shapes, jtu.rand_some_zero, []),
+    op_record("any", 1, bool_dtypes, all_shapes, jtu.rand_some_zero, []),
+    op_record("mean", 1, all_dtypes, nonempty_shapes, jtu.rand_default, [],
+              inexact=True),
+    op_record("var", 1, all_dtypes, nonempty_shapes, jtu.rand_default, [],
+              inexact=True),
+    op_record("std", 1, all_dtypes, nonempty_shapes, jtu.rand_default, [],
+              inexact=True),
+]
+
 JAX_REDUCER_NO_DTYPE_RECORDS = [
     op_record("all", 1, all_dtypes, all_shapes, jtu.rand_some_zero, []),
     op_record("any", 1, all_dtypes, all_shapes, jtu.rand_some_zero, []),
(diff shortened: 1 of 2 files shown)


tests/lax_numpy_test.py
```diff
@@ -853,6 +864,51 @@ def np_fun(x):
...
+      for whereshape in _compatible_shapes(shape)
+      for axis in list(range(-len(shape), len(shape))) + [None]
+      for keepdims in [False, True])
+    for rec in JAX_REDUCER_WHERE_NO_INITIAL_RECORDS))
+  def testReducerWhere(self, np_op, jnp_op, rng_factory, shape, dtype, axis,
+                              keepdims, inexact, whereshape):
+    if (shape in [()] + scalar_shapes and
+            dtype in [jnp.int16, jnp.uint16] and
+            jnp_op in [jnp.min, jnp.max]):
+      self.skipTest("Known XLA failure; see https://github.com/google/jax/issues/4971.")
```
jakevdp, line 887: The referenced PR is closed - can we remove this skip?

tests/lax_numpy_test.py
```diff
@@ -853,6 +864,51 @@ def np_fun(x):
...
+    rng = rng_factory(self.rng())
+    is_bf16_nan_test = dtype == jnp.bfloat16 and rng_factory.__name__ == 'rand_some_nan'
```
jakevdp, line 889: It looks like the inputs never use `rand_some_nan`, so this condition could be removed

tests/lax_numpy_test.py
```diff
@@ -853,6 +864,51 @@ def np_fun(x):
...
+    # Do not pass where via args_maker as that is incompatible with _promote_like_jnp.
+    where = jtu.rand_bool(self.rng())(whereshape, np.bool_)
```
jakevdp, line 891: Another effect of passing this outside args_maker is that it is treated as static – i.e. the test does not cover the case that `where` is a traced value. I see this comes from the existing `where` test, but it's a bit worrisome in terms of test coverage.

tests/lax_numpy_test.py
```diff
@@ -853,6 +864,51 @@ def np_fun(x):
...
+    @jtu.ignore_warning(category=RuntimeWarning,
+                        message="Degrees of freedom <= 0 for slice.*")
+    @jtu.ignore_warning(category=RuntimeWarning,
+                        message="Mean of empty slice.*")
+    @jtu.ignore_warning(category=RuntimeWarning,
+                        message="invalid value encountered in true_divide*")
```
jakevdp, line 897: These ignore_warning statements look like a good way to address this! 👍

jakevdp (commented):
Looks good! Regarding the numpy incompatibility you mentioned... we should gate those cases with a numpy version check

tests/lax_numpy_test.py
mtsokol: So `args_maker = lambda: [rng(shape, dtype)]` should also contain `[..., rng(whereshape, np.bool_)]` and `jnp_fun = lambda x, where: ...`? Should I rewrite it?

mtsokol:
> Looks good! Regarding the numpy incompatibility you mentioned... we should gate those cases with a numpy version check

I assumed it will be 1.20.2 so I added (...,2) to required version.

jakevdp:
> I assumed it will be 1.20.2 so I added (...,2) to required version.

Oh, I see... but you disabled the entire test in this case, right? It would be better to just skip the `CheckAgainstNumpy`, because the JAX code by itself should work regardless of what numpy version is installed. And try to only skip the actual cases that are broken, because if I understand correctly, some of them should work correctly with 1.20.0.

We can't merge this code if all tests that cover it are skipped.

tests/lax_numpy_test.py
```diff
@@ -853,6 +864,48 @@ def np_fun(x):
...
+      res = np_op(x_cast, axis, keepdims=keepdims, where=where)
+      res = res if not is_bf16_nan_test else res.astype(jnp.bfloat16)
+      return res
+
+    np_fun = _promote_like_jnp(np_fun, inexact)
+    np_fun = jtu.ignore_warning(category=np.ComplexWarning)(np_fun)
+    jnp_fun = lambda x: jnp_op(x, axis, keepdims=keepdims, where=where)
+    jnp_fun = jtu.ignore_warning(category=jnp.ComplexWarning)(jnp_fun)
+    args_maker = lambda: [rng(shape, dtype)]
+    if numpy_version >= (1, 20, 2):
```
jakevdp, line 905: From the numpy issue, it looks like this works correctly for `sum` but not for `mean`. Can we only skip the tests in specific cases where numpy returns the wrong result?

mtsokol: I've just pushed changes that includes `any` and `all` from this test suite - the rest is failing. `sum` is not present in this test suite as it's already tested with `where` in `testReducerInitialWhere`

jakevdp: Sounds good, thanks!

tests/lax_numpy_test.py
jakevdp: That sounds right, though I think it might make the type promotion wrapper not work correctly, so we'd have to use `check_dtypes=False` in the numpy check.

I could go either way on this - it's up to you.

mtsokol:
I marked this PR so it's not a draft anymore. I think it's ready but in CI one test is failing that wasn't affected here. One of the `tests/lax_numpy_test.py::LaxBackedNumpyTests::testReducerInitialWhereSum` fails with being out of tolerance. I haven't modified `sum` or `reduce` methods here and I also can't reproduce it locally (with `pytest -n auto -k "testReducerInitialWhere" tests/lax_numpy_test.py`) so I don't know where it comes from.

jakevdp:
>  also can't reproduce it locally (with pytest -n auto -k "testReducerInitialWhere" tests/lax_numpy_test.py) so I don't know where it comes from.

Try with:
```
$ JAX_NUM_GENERATED_CASES=25 pytest -n auto -k "testReducerInitialWhere" tests/lax_numpy_test.py
```

mtsokol:
@jakevdp Hi! Thanks for a guide - it's all green now (changing existing test's name caused this failure - no idea why). Also https://github.com/numpy/numpy/pull/18560 was merged yesterday so I guess it will be present in 1.20.2. Is it ready to merge or should I add anything?

jakevdp (approved):
Looks great!

jakevdp:
Regarding the test name change causing failures: I believe it's related to the random seed being changed.

One last request: could you squash all your commits into one and force-push to the branch?

mtsokol:
Sure - squashed and pushed, all green now.

jakevdp:
Thanks for all the work on this!

mtsokol:
Thank you for assistance! I guess #5842 can be closed.
Also if there's a next not-so-hard issue to take please let me know! I can try sth in coming weeks. If there isn't right now I will check issues page somewhere in the future.

PR #5975, opened 2021-03-07
Simplify lazy expressions

hawkinsp:
Replace _lazy_expr on xla._DeviceArray with a new _lazy_broadcast_dims field that corresponds to the dims of the previous lazy expression.

Add new `_DeviceArray._broadcast` and `_DeviceArray._transpose` methods that take the place of `lazy.broadcast` and `lazy.transpose`. Change in preparation for implementing the same functionality in C++.


hawkinsp:
This PR is moot now we have removed lazy expressions.

PR #5980, opened 2021-03-08
Initial commit of seventh JAX-101 notebook

jakevdp:


docs/jax-101/07-state.md
```diff
@@ -0,0 +1,283 @@
...
+
++++ {"id": "1wq3L6Xg1UHP"}
+
+## Taking it further
+
+The strategy described above is how any (jitted) JAX program must handle state. 
+
+Handling parameters manually seems fine if you're dealing with two parameters, but what if it's a neural net with dozens of layers? You might already be getting worried about two things:
+
+1) Are we supposed to initialize them all manually, essentialy repeating what we already write in the forward pass definition?
```
8bitmp3, line 278: Suggestion: nit ("essentially")

```suggestion
1) Are we supposed to initialize them all manually, essentially repeating what we already write in the forward pass definition?
```

docs/jax-101/07-state.md
```diff
@@ -0,0 +1,283 @@
...
+  print(fast_count())
+```
+
++++ {"id": "weiI0V7_pKGv"}
+
+Oh no! Our counter isn't working. This is because the line
+```
+self.n += 1
+```
+in `count` was only called once, when JAX was compiling the method call. Moreover, since the return value doesn't depend on the arguments to `count`, once it returned the first 1, subsequent calls to `fast_count` will always return 1. This won't do. So, how do we fix it?
```
8bitmp3, line 104: Suggestion: nit (present tense, maybe better formatting)

```suggestion
in `count` was only called once, when JAX was compiling the method call. Moreover, since the return value doesn't depend on the arguments to `count`, once it returns the first `1`, subsequent calls to `fast_count` will always return `1`. This won't do. So, how do we fix it?
```

docs/jax-101/07-state.md
```diff
@@ -0,0 +1,283 @@
...
+This section explores how JAX constrains the implementation of stateful programs.
+
++++ {"id": "Avjnyrjojo8z"}
+
+## Motivation
+
+In machine learning, program state most often comes in the form of
+* model parameters,
+* optimizer state, and
+* stateful layers, such as [BatchNorm](https://en.wikipedia.org/wiki/Batch_normalization).
```
8bitmp3, line 31: Suggestion: nit

```suggestion
In machine learning, program state most often comes in the form of:
* model parameters,
* optimizer state, and
* stateful layers, such as [BatchNorm](https://en.wikipedia.org/wiki/Batch_normalization).
```

docs/jax-101/07-state.md
```diff
@@ -0,0 +1,283 @@
...
+state = counter.reset()
+
+for _ in range(3):
+  value, state = counter.count(state)
+  print(value)
+```
+
++++ {"id": "PrBjmgZtq89b"}
+
+In this new version of Counter, we moved `n` to be an argument of `count`, and added another return value that represents the new, updated, state. To use this counter, we now need to keep track of the state explicitly. But in return, we can now safely `jax.jit` this counter:
```
8bitmp3, line 139: Suggestion: nit (formatting)

```suggestion
In this new version of `Counter`, we moved `n` to be an argument of `count`, and added another return value that represents the new, updated, state. To use this counter, we now need to keep track of the state explicitly. But in return, we can now safely `jax.jit` this counter:
```

(similar to  "In our case, the `CounterV2` class is nothin..." further below.)

docs/jax-101/07-state.ipynb
```diff
@@ -0,0 +1,440 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "SQ-RNLfdiw04"
+   },
+   "source": [
+    "The `n` attribute maintains the counter's _state_ between successive calls of `count`. It is modified as a side effect of calling `count`.\n",
+    "\n",
+    "Let's say we want to count fast, so we `jax.jit` the `count` method. (In this example, this wouldn't actually help speed anyway, for many reasons, but treat this as a toy model of wanting to jit-compile the update of model parameters, where `jax.jit` makes an enormous difference)."
```
8bitmp3, line 108: Suggestion:

```suggestion
    "Let's say we want to count fast, so we `jax.jit` the `count` method. (In this example, this wouldn't actually help speed anyway, for many reasons, but treat this as a toy model of wanting to JIT-compile the update of model parameters, where `jax.jit` makes an enormous difference)."
```

8bitmp3 (commented):
Minor tiny suggestions @jakevdp Thanks to VM and you for putting this together

docs/jax-101/07-state.md
```diff
@@ -0,0 +1,275 @@
...
+The strategy described above is how any (jitted) JAX program must handle state. 
+
+Handling parameters manually seems fine if you're dealing with two parameters, but what if it's a neural net with dozens of layers? You might already be getting worried about two things:
+
+1) Are we supposed to initialize them all manually, essentially repeating what we already write in the forward pass definition?
+
+2) Are we supposed to pipe all these things around manually?
+
+The details can be tricky to handle, but there are examples of libraries that take care of this for you. One example
+is [Haiku](https://github.com/deepmind/dm-haiku), which allows you to write OOP-style neural nets without explicitly tracking their parameters, and then to transform them into functional programs.
```
skye, line 275: Instead of calling out Haiku here, how about we link to https://github.com/google/jax#neural-network-libraries instead.

skye (approved):

docs/jax-101/07-state.md
jakevdp: Great idea - done!

PR #5984, opened 2021-03-09
Automatically initialize Cloud TPU topology env vars if running on a Cloud TPU VM.

skye:
This removes the need to manually set these env vars when running on a Cloud TPU pod slice.

### DIFF
--- a/jax/__init__.py
+++ b/jax/__init__.py
@@ -17,6 +17,18 @@ import os as _os
 _os.environ.setdefault('TF_CPP_MIN_LOG_LEVEL', '1')
 del _os
 
+# Set Cloud TPU env vars if necessary before transitively loading C++ backend
+from .cloud_tpu_init import cloud_tpu_init as _cloud_tpu_init
+try:
+  _cloud_tpu_init()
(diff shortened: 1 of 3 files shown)


jax/__init__.py
```diff
@@ -17,6 +17,11 @@
 _os.environ.setdefault('TF_CPP_MIN_LOG_LEVEL', '1')
 del _os
 
+# Set Cloud TPU env vars if necessary before transitively loading C++ backend
+from .cloud_tpu_init import cloud_tpu_init as _cloud_tpu_init
+_cloud_tpu_init()
```
jakevdp, line 22: Would it be worth wrapping this in a try/except? I'm worried this could make jax unimportable if something unanticipated goes wrong.

skye: Good idea, done. I made it raise a warning with the exception, because in theory this should never fail (like so much of jax!). Lemme know if you think it should be silent instead.

jakevdp: Warning looks great 👍

jakevdp (approved):

PR #5990, opened 2021-03-09
jnp.power: fix overflow case for x1=0

### LINKED ISSUE #5987: Exponentiation on int32 data with base 0 gives different results when jitted
Example

```python
from jax import numpy as jnp
import jax

EXP_MULTIPLIER = 1


def power(a, b):
    return a**b


a = jnp.int32(0)
b = jnp.int32(EXP_MULTIPLIER * 64)

result = power(a, b)
result_jit = jax.jit(power)(a, b)

assert result == result_jit, f'{result} != {result_jit}'
```

```
AssertionError: 0 != 1
```

jakevdp:
Fixes #5987

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -607,12 +607,14 @@ def power(x1, x2):
 
   # TODO(phawkins): add integer pow support to XLA.
   bits = 6  # Anything more would overflow for any x1 > 1
-  acc = ones(shape(x1), dtype=dtype)
+  zero = _constant_like(x2, 0)
+  one = _constant_like(x2, 1)
+  # Initialize acc carefully such that pow(0, x2) is zero for x2 != 0
+  acc = where(lax.bitwise_and(lax.eq(x1, zero), lax.ne(x2, zero)), zero, one)
   for _ in range(bits):
-    acc = where(lax.bitwise_and(x2, _constant_like(x2, 1)),
-                lax.mul(acc, x1), acc)
+    acc = where(lax.bitwise_and(x2, one), lax.mul(acc, x1), acc)
     x1 = lax.mul(x1, x1)
-    x2 = lax.shift_right_logical(x2, _constant_like(x2, 1))
+    x2 = lax.shift_right_logical(x2, one)
   return acc
 
 
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1755,6 +1755,16 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     self.assertEqual(eqns[0].primitive, lax.integer_pow_p)
 
   @parameterized.named_parameters(jtu.cases_from_list(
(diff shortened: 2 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -607,12 +607,14 @@ def power(x1, x2):
 
   # TODO(phawkins): add integer pow support to XLA.
   bits = 6  # Anything more would overflow for any x1 > 1
-  acc = ones(shape(x1), dtype=dtype)
+  zero = _constant_like(x2, 0)
+  one = _constant_like(x2, 1)
+  # Initialize acc carefully such that pow(0, x2) is zero for x2 != 0
```
hawkinsp, line 612: I'm not 100% sure it's necessary that 0**0 is 1, but it's not the worst idea, so let's keep it.

tests/lax_numpy_test.py
```diff
@@ -1754,6 +1754,16 @@ def testIntegerPower(self, ptype):
     self.assertLen(eqns, 1)
     self.assertEqual(eqns[0].primitive, lax.integer_pow_p)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_x={}_y={}".format(x, y), "x": x, "y": y}
+      for x in [-1, 0, 1]
```
hawkinsp, line 1759: It would not be an error for JAX to produce anything at all if the value overflows. So this test may be too strict. Maybe add a comment to that effect?

hawkinsp (approved):

tests/lax_numpy_test.py
hawkinsp: Actually, never mind, I see that the values of `x` can't cause overflows.

jax/_src/numpy/lax_numpy.py
jakevdp: It's necessary if we want to match numpy's output in this case

hawkinsp: That's my question: is this an intentional behavior of NumPy or an accidental behavior of its implementation? We certainly do not want to match NumPy for all inputs, e.g., if integer overflow occurs.  That said, matching numpy for this specific input seems fine.

jakevdp: I think there are good arguments to define `0^0 = 1`, particularly in the context of integers. Searching around, Knuth argues this way:

> Some textbooks leave the quantity 0^0 undefined, because the functions x^0 and 0^x have different limiting values when x decreases to 0. But this is a mistake. We must define x0 = 1, for all x, if the binomial theorem is to be valid when x = 0, y = 0, and/or x = −y. The binomial theorem is too important to be arbitrarily restricted! By contrast, the function 0^x is quite unimportant"

hawkinsp: Another good argument is that IEEE 754-2008 section 9.2.1 specifies pow(0, 0) == 1.

PR #5996, opened 2021-03-09
Update cuDNN versions to highest available.

skye:
The TF build depends on newer versions than we previously used.


hawkinsp:
Seems fine, provided these versions exist in NVidia's repository.

hawkinsp (approved):

skye:
Yeah, I found these versions via apt in the build docker, and just built the latest jaxlibs using them.

PR #5998, opened 2021-03-09
Add jtu.count_device_put for tests to count device_put.

zhangqiaorjc:


jax/_src/lax/lax.py
```diff
@@ -435,10 +435,8 @@ def convert_element_type(operand: Array, new_dtype: DType = None,
     msg = "Casting complex values to real discards the imaginary part"
     warnings.warn(msg, np.ComplexWarning, stacklevel=2)
 
-  if not isinstance(operand, (core.Tracer, xla.DeviceArray)):
-    return _device_put_raw(np.asarray(operand, dtype=new_dtype),
-                           weak_type=new_weak_type)
-  elif (old_dtype, old_weak_type) == (new_dtype, new_weak_type):
+  if ((old_dtype, old_weak_type) == (new_dtype, new_weak_type)
+      and isinstance(operand, (core.Tracer, xla.DeviceArray))):
     return operand
```
jakevdp, line 440: One issue with this is that e.g. `lax.convert_element_type(np.arange(4))` will now return a numpy array rather than a JAX array, which is why we had originally called `device_put` in that case. That breaks the normal assumption that lax functions will return Device arrays.

mattjj: I'm not sure if that's true, as this branch is only triggered if the input is a Tracer or DeviceArray. For example:

```python
In [1]: from jax import lax
In [2]: import numpy as np
In [3]: lax.convert_element_type(np.arange(4), np.int32)
Out[3]: DeviceArray([0, 1, 2, 3], dtype=int32)
In [4]: lax.convert_element_type(np.arange(4), np.int64)
Out[4]: DeviceArray([0, 1, 2, 3], dtype=int32)
```

Do you have a particular example in mind?

mattjj (approved):

jax/_src/lax/lax.py
jakevdp: Oh, I see I misread the condition, thanks!

I think the side-effect of this change will be that every onstant value used in a function will result in a `convert_element_type` inserted into the jaxpr: i.e. each `x + 1` will insert a `convert_element_type` for the `1`.

It's probably the right thing to do functionally, but when we discussed a similar change a few months ago we decided the jaxprs were too messy, and that's when we landed on using `device_put` instead.

zhangqiaorjc: Does `convert_element_type` have runtime cost or does it just add visual clutter to jaxpr?

mattjj:
I had pushed some commits on top of @zhangqiaorjc's, but those are trickier to land because they involve some changes to jaxprs. So I'm going to try splitting those off into a separate PR!

PR #5999, opened 2021-03-09
Add scan and while rule for jax.experimental.callback transformation

sharadmv:


jax/experimental/callback.py
```diff
@@ -169,3 +175,70 @@ def process_custom_vjp_call(self, primitive, fun, fwd, bwd, tracers,
...
     del primitive, fwd, bwd, out_trees  # Unused.
     return fun.call_wrapped(*tracers)
+
+
+custom_callback_rules: Dict[Any, Any] = {}
+
+def _scan_callback_rule(trace, *tracers, reverse, length, num_consts, num_carry,
+                        jaxpr, linear, unroll):
+  const_tracers, carry_tracers, xs_tracers = split_list(tracers, [num_consts, num_carry])
+  carry_avals, xs_avals = tree_map(lambda x: x.aval, (carry_tracers, xs_tracers))
```
mattjj, line 185:
```suggestion
  carry_avals, xs_avals = tree_map(op.attrgetter('aval'), (carry_tracers, xs_tracers))
```

Just a stylistic thing I like :) Same elsewhere.

Not worth blocking the merge though!

jax/experimental/callback.py
```diff
@@ -169,3 +175,70 @@ def process_custom_vjp_call(self, primitive, fun, fwd, bwd, tracers,
...
+  const_vals, carry_vals, xs_vals = tree_map(lambda x: x.val, (const_tracers, carry_tracers, xs_tracers))
+
+  x_tracers = [t[0] for t in xs_tracers]
```
mattjj, line 188: I'm not sure this always works because `t` might not support indexing; in particular it might be a traced `core.unit` value.

Better to use `core.mapped_aval` here, something like

```python
x_avals = [core.mapped_aval(length, 0, core.get_aval(t)) for t in x_tracers]
```

mattjj (approved):

mattjj:
I suggested some tweaks but since this is in `jax.experimental` let's just land and iterate.

PR #6003, opened 2021-03-10
Add support for custom derivatives in jax.experimental.callback

sharadmv:
This PR enables `callback` to rewrite the bodies of `custom_derivatives` primitives. 

Any suggestions on how to test the custom_call_jaxpr primitives? I'm not sure how/why they are created normally.

### DIFF
--- a/jax/experimental/callback.py
+++ b/jax/experimental/callback.py
@@ -21,13 +21,15 @@ import jax.numpy as jnp
 from jax import core
 from jax.core import Trace, Tracer, jaxpr_as_fun
 from jax import lax
+from jax import custom_derivatives as cd
+from jax.interpreters import partial_eval as pe
 from jax import linear_util as lu
 from jax._src.util import partial, safe_map, wraps, split_list
 from jax._src.lax import control_flow as lcf
 
 import inspect
 from jax.api_util import flatten_fun_nokwargs
-from jax.tree_util import tree_flatten, tree_unflatten, tree_structure, tree_leaves, tree_map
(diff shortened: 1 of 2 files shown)


mattjj:
> Any suggestions on how to test the custom_call_jaxpr primitives? I'm not sure how/why they are created normally.

They're created when the custom derivative calls are staged out. That is, `custom_jvp_call` never appears in a jaxpr and is just a Python tracing primitive; when tracing a `custom_jvp` function to a jaxpr by e.g. `lax.scan`, we [bind a `custom_jvp_call_jaxpr`](https://github.com/google/jax/blob/8d3b4ac2f36761513ae132b59180e9770fe34922/jax/custom_derivatives.py#L221). So unlike `xla_call_p`, which is both a Python tracing primitive and a jaxpr primitive, we have pairs of primitives `custom_jvp_call_p`/`custom_jvp_call_jaxpr_p` and `custom_vjp_call_p`/`custom_vjp_call_jaxpr_p`.

TL;DR test them by using `scan`, like in [these tests](https://github.com/google/jax/blob/8d3b4ac2f36761513ae132b59180e9770fe34922/tests/api_test.py#L3041-L3128).

It might be worth adding a test like that, just for basic functionality, before we land-and-iterate!

sharadmv:
I added a test, but left a TODO to enable rewriting thunks (because I'm not sure of the best way to do that). If it's easy, I can add it to the PR, but if not, we can iterate on it later. WDYT?

sharadmv:
Thunks should be working now!

mattjj (approved):
Awesome! Thanks for doing this.

PR #6013, opened 2021-03-10
Add a test that fails due to unsupported strides.

zhangqiaorjc:
- test fails during dlpack tensor to buffer conversion
- XLA layout does not support arbitrary dlpack strides
- users should explicit materialze such tensors by making a copy

### DIFF
--- a/tests/array_interoperability_test.py
+++ b/tests/array_interoperability_test.py
@@ -19,6 +19,7 @@ from absl.testing import absltest, parameterized
 import jax
 from jax.config import config
 import jax.dlpack
+from jax.lib import xla_bridge, xla_client
 import jax.numpy as jnp
 from jax import test_util as jtu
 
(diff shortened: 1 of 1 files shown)


zhangqiaorjc:
This is the same PR as https://github.com/google/jax/pull/4230

except for bumping version guard to 0.1.60 --the current min version of jaxlib

and also rebasing to HEAD

tests/array_interoperability_test.py
```diff
@@ -152,6 +153,21 @@ def testTorchToJax(self, shape, dtype):
     y = jax.dlpack.from_dlpack(dlpack)
     self.assertAllClose(np, y)
 
+  @unittest.skipIf(not torch or jax.lib.version < (0, 1, 60),
```
hawkinsp, line 156: You can just drop the version check. 0.1.60 is the minimum.

build/test-requirements.txt
```diff
@@ -8,3 +8,4 @@ pillow
 pytest-benchmark
 pytest-xdist
 wheel
+torch
```
hawkinsp, line 11: I'm not sure we should start adding torch as a test dep.

zhangqiaorjc: removed.

PR #6015, opened 2021-03-10
Improve Actions hygiene

sethvargo:
👋 hello there! I'm a fellow Googler who works on projects that leverage GitHub Actions for CI/CD. Recently I noticed a large increase in our queue time, and I've tracked it down to the [limit of 180 concurrent jobs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) for an organization. To help be better citizens, I'm proposing changes across a few repositories that will reduce GitHub Actions hours and consumption. I hope these changes are reasonable and I'm happy to talk through them in more detail.

- **(you were already doing this, thank you!**) Only run GitHub Actions for pushes and PRs against the main branch of the repository. If your team uses a forking model, this change will not affect you. If your team pushes branches to the repository directly, this changes actions to only run against the primary branches or if you open a Pull Request against a primary branch.

- For long-running jobs (especially tests), I added the "Cancel previous" workflow. This is very helpful to prevent a large queue backlog when you are doing rapid development and pushing multiple commits. Without this, GitHub Actions' default behavior is to run all actions on all commits.

There are other changes you could make, depending on your project (but I'm not an expert):

- If you have tests that should only run when a subset of code changes, consider gating your workflow to particular file paths. For example, we have some jobs that do Terraform linting, but [they only run when Terraform files are changed](https://github.com/google/exposure-notifications-verification-server/blob/c4f59fee71042cf668747e599e7c769fca736554/.github/workflows/terraform.yml#L3-L11).

Hopefully these changes are not too controversial and also hopefully you can see how this would reduce actions consumption to be good citizens to fellow Googlers. If you have any questions, feel free to respond here or ping me on chat. Thank you!

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Cancel previous
-        uses: styfle/cancel-workflow-action@0.7.0
+        uses: styfle/cancel-workflow-action@0.8.0
         with:
           access_token: ${{ github.token }}
-        if: ${{github.ref != 'refs/head/main'}}
+        if: ${{github.ref != 'refs/head/master'}}
       - uses: actions/checkout@v2
       - name: Set up Python 3.8
         uses: actions/setup-python@v2
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -68,7 +68,7 @@ jobs:
       uses: styfle/cancel-workflow-action@0.7.0
       with:
         access_token: ${{ github.token }}
-      if: ${{github.ref != 'refs/head/main'}}
+      if: ${{github.ref != 'refs/head/master'}}
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
@@ -116,7 +116,7 @@ jobs:
       uses: styfle/cancel-workflow-action@0.7.0
       with:
         access_token: ${{ github.token }}
-      if: ${{github.ref != 'refs/head/main'}}
+      if: ${{github.ref != 'refs/head/master'}}
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -14,6 +14,11 @@ jobs:
   lint_and_typecheck:
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel previous
+        uses: styfle/cancel-workflow-action@0.7.0
+        with:
+          access_token: ${{ github.token }}
+        if: ${{github.ref != 'refs/head/main'}}
       - uses: actions/checkout@v2
       - name: Set up Python 3.8
         uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -14,6 +14,11 @@ jobs:
   lint_and_typecheck:
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel previous
+        uses: styfle/cancel-workflow-action@0.7.0
+        with:
+          access_token: ${{ github.token }}
+        if: ${{github.ref != 'refs/head/main'}}
```
jakevdp, line 21: Does `refs/head/main` refer to a branch name? If so I think this should probably be `refs/head/master` for this repository. Am I understanding that correctly?

jakevdp (commented):
Thanks! Overall this looks good. One question below

.github/workflows/ci-build.yaml
sethvargo: oop - good catch. Fixed!

jakevdp: It looks like there are two other locations as well. Thanks!

sethvargo: Sorry - I'm working across 50 repos and E_TOO_MANY_TABS. Updated.

jakevdp:
I'm looking at the docs for this... am I correct in inferring that this will only work for pull requests not opened from a fork? I think most PRs in JAX come from forks of the repository, so this approach might be more relevant: https://github.com/marketplace/actions/cancel-workflow-action#advanced-pull-requests-from-forks

sethvargo:
Hmm. The problem that that approach is that Actions is a FIFO queue, so you can get into a thundering herd problem where a bunch of Actions sit between your cancel workflow, meaning it won't run until other jobs have finished, defeating the point :/

jakevdp:
OK, thanks - this looks like a good start. Let's get it merged (as soon as the workflow gets unqueued :grin:).

Would you mind squashing the changes into a single commit?

sethvargo:
Well I want to be sure that this doesn't break your existing workflows. Let's try to coordinate a merge and then open a PR from a fork to test. Then we can easily revert if it fails. I have other ideas for this problem, but this is the easiest :). 

I'm not at my computer now, but I'll squash and rebase tomorrow.

sethvargo:
Hey @jakevdp - I'm around again if you want to try and coordinate a test.

jakevdp:
What sort of test do you have in mind?

Side note: I'm noticing that github actions are still Queued for the commit from 19 hours ago... that doesn't seem like it would be due to organic growth in actions usage by our org. Do you know what the root cause of yesterday's sharply increased usage might be?

sethvargo:
Yes, there are other things at play (namely a massive decrease in the number of parallel runners for the google org), and that's being addressed separately. The reason the queue keeps _growing_ is due to the lack of good hygiene on some of the projects though (like not canceling previous builds when new commits are pushed).

jakevdp (approved):

taquitos:
The Google org has A LOT of projects, and they all share the same set of free runners with a concurrency of 20. Even just one repo misconfiguration (like not setting the default action timeout - which is 6 hours by default 😵, or not cancelling runs when new pushes happen) can eat up all of the shared resources.
Another thing that might be worth investigating is using hosted runners https://github.blog/2020-08-04-github-actions-self-hosted-runners-on-google-cloud/

hawkinsp:
@taquitos We should have this discussion internally, but our understanding is that self-hosted (or GCP-hosted) runners aren't allowable for security reasons because we are a public repo and accept public PRs. Otherwise we almost certainly *would* use our own runners so we can assign them more CPU.

Issue #5982, opened 2021-03-08
jax.experimental.enable_x64 fails in JIT context


jakevdp:
minimal repro:
```python
from jax import jit, experimental

@jit
def f(a):
  with experimental.enable_x64():
    return 1 + a
f(1)
```

jakevdp:
Happily, it looks like #6014 mostly fixes this! The issue came from the fact that convert_element_type was not being bound at trace time (because the constant is 64-bit in that context), though it was required at runtime (because the constant is 32-bit in that context).

jakevdp:
Here's a repro that's closer to the metal:
```python
import jax.numpy as jnp
from jax import experimental, jit, lax

# Create an int64 in X32 mode
with experimental.enable_x64():
  x = jnp.int64(1)

y = lax.convert_element_type(x, jnp.int32)
assert y.dtype == jnp.int32  # fails

z = jit(lax.convert_element_type, static_argnums=1)(x, jnp.int32)
assert z.dtype == jnp.int32  # fails
```

btanner:
Bumping this issue with another simple example:
```
import jax
import jax.experimental
import jax.numpy as jnp

def foo(x):
  with jax.experimental.enable_x64():
    clipped = x.astype(jnp.uint64)
    clipped = clipped % 2

    # Notably, the line below gives the identical error
    # clipped = clipped % jnp.asarray(2, jnp.uint64)

    return clipped

x = jnp.asarray(.5)

print(foo(x))
j_foo = jax.jit(foo)
print(j_foo(x))
```
The non-jitted call works fine.

Here is the error:
```
google3/third_party/py/jax/interpreters/xla.py in jaxpr_subcomp(c, jaxpr, backend, axis_env, consts, name_stack, *args)
    462 
    463     assert isinstance(ans, xe.XlaOp)
--> 464     c.get_shape(ans)  # force xla to do shape error checking
    465     if eqn.primitive.multiple_results or any(v.aval._num_buffers > 1 for v in eqn.outvars):
    466       out_nodes = xla_destructure(c, ans)

RuntimeError: Invalid argument: Binary op remainder with different element types: u64[] and u32[].
```

mattjj:
Copying some comments from a chat with @btanner , for posterity:

[...]

PR #6018, opened 2021-03-10
convert_element_type: don't canonicalize old_dtype

jakevdp:
Related to #5982 and #6014. Canonicalizing the input type can lead to the primitive not being bound, because `old_dtype` doesn't reflect the actual value of the input dtype.

The added test fails on master, and passes on this branch. This is orthogonal to the change in #6014; they both fix similar issues.


mattjj (approved):
LGTM! Nice fix. I was at first worried that this would change how we transpose things, but as Jake explained to me offline, this `old_dtype` value isn't passed into the primitive bind (though it used to be); it's just used in this traceable wrapper.

PR #6019, opened 2021-03-11
Adjust tolerance for TPU on ExamplesTest.testKernelRegressionTrainAndPredict

skye:
@mattjj does this look ok? This is a pretty loose tolerance but not sure how else to address it...

### DIFF
--- a/examples/examples_test.py
+++ b/examples/examples_test.py
@@ -21,6 +21,7 @@ from absl.testing import parameterized
 
 import numpy as np
 
+from jax import lax
 from jax import test_util as jtu
 from jax import random
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


examples/examples_test.py
```diff
@@ -90,7 +90,7 @@ def testKernelRegressionTrainAndPredict(self):
     ys = jnp.dot(xs, truth)
     kernel = lambda x, y: jnp.dot(x, y)
```
hawkinsp, line 91: You might try using highest precision for the matmuls. On TPU, this may be important.

mattjj:
Is this a TPU-only failure? If so we could do `0.2 if jtu.device_under_test() == 'tpu' else 1e-3`.

But also +1 to Peter's suggestion of trying high-precision matmuls. Maybe this is a good example of why the default precision should be different...

examples/examples_test.py
skye: Nice, Precision.HIGH does the trick! (On TPU, I'm assuming this won't hurt GPU/CPU)

PR #6025, opened 2021-03-11
Fix import of datasets in differentially private SGD example

bastings:
Python 3 doesn't like the relative import of `datasets`


mattjj:
Thank you, Jasmijn! Sorry for being slow to merge :)

PR #6029, opened 2021-03-11
Actions hygiene: set reasonable timeouts for github actions jobs

jakevdp:
The default timeout is six hours.

I set these new timeouts by looking at the time taken for a recent set of completed tests: https://github.com/google/jax/actions/runs/640842836

These timeouts are at least two times the expected job completion time.


froystig (approved):

jakevdp:
This will not actually affect our current queue, so I'm not going to force a merge until we actually see it run successfully.

PR #6034, opened 2021-03-12
ahead-of-time compilation

froystig:
```python
$ cat test.py
import jax
import jax.numpy as jnp

def f(x):
  return x + 2.

x = jnp.ones((2, 3))
g = jax.api.compile_for_args(f, x)
print(g(x + 1.))
```

```
$ python test.py
[[4. 4. 4.]
 [4. 4. 4.]]
```

### DIFF
--- a/jax/api.py
+++ b/jax/api.py
@@ -92,8 +92,8 @@ F = TypeVar("F", bound=Callable)
 T = TypeVar("T")
 U = TypeVar("U")
 
-map = safe_map
-zip = safe_zip
+map, unsafe_map = safe_map, map
+zip, unsafe_zip = safe_zip, zip
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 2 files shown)


Joshuaalbert:
Just to clarify, does `g = jax.api.compile_for_args(f, x)` block until `f` is compiled for dummy variable `x` and then `g` can be used later?

froystig:
That's correct for the current draft. Ultimately we'd rather not require an actual array for `x` when it isn't needed, and we still need to strengthen the checks that `g` is called in a compatible way.

litanlitudan:
This is very cool!!! Also very useful for some DSA compiler projects as a lot of lower-level compilation is very slow. Having an AOT interface would make a lot of sense. If we can also specify the path of the compiled binary file, that would be even better.

hawkinsp:
Closing because it is most likely subsumed by the version of the API that is checked in.

Issue #6045, opened 2021-03-12
JIT leads to overflow error for large integers in X64


jakevdp:
```python
import jax.numpy as jnp
import numpy as np
from jax import config, jit
config.update('jax_enable_x64', True)

big = np.iinfo(np.int64).max

print(np.array(big).dtype)  # int64
print(np.array(big + 1).dtype)  # uint64

print(jnp.array(big).dtype)  # int64
print(jnp.array(big + 1).dtype)  # uint64

print(jit(jnp.array)(big).dtype)  # int64
print(jit(jnp.array)(big + 1).dtype)  # OverflowError: Python int too large to convert to C long
```

hawkinsp:
Seems like the bug is that
```
print(jnp.array(big + 1).dtype)  # uint64
```

should error as well?

hawkinsp:
I think I prefer the error to silently overflowing, which is what NumPy does. (And if you do want NumPy's behavior, you can first cast to a NumPy array.)

jakevdp:
Just to clarify, when you say "silently overflowing" are you referrring to the behavior of switching to `uint64` if the value is too large for `int64`, and swittching to `dtype('O')` if the value is too large for `uint64`?

hawkinsp:
Yeah I guess "overflowing" isn't a correct description, on further thought.

Maybe we should follow NumPy here. We could certainly have the same `uint64` behavior for `jit` of integers outside the `int64` range. We can't use `dtype('O')`, of course, so we'd have to error in that case.

jakevdp:
OK, that sounds like a reasonable approach.

The reason I was going down this route is related to #5952: I need to figure out what to do in X32 mode with Python integers between `int32.max` and `uint32.max`. If we choose to promote to uint64 here, we should probably choose to promote to uint32 there.

PR #6047, opened 2021-03-12
Make JAX treatment of large integers consistent

jakevdp:
Addresses part of #6045.

The goal here is to make JAX treatment of large Python integers consistent: regardless of how the integer is passed to JAX, the behavior should be the same.

After discussion, we decided on this behavior for X64 mode:
```python
import pytest
from jax import api, config
import jax.numpy as jnp

functions = [
  jnp.array,
  api.device_put,
  api._python_jit(lambda x: x),
  api._cpp_jit(lambda x: x)
]

big = jnp.iinfo('int64').max
for func in functions:
  assert func(big).dtype == 'int64'
  with pytest.raises(OverflowError):
    func(big + 1)
```
Note that this differs from the current behavior in one case: `jnp.array(big + 1)` returns a `uint64` array rather than raising an OverflowError.

Similar changes for X32 mode will require jaxlib C++ changes, and will be part of a followup PR.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2891,6 +2891,10 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
   dtype = dtype and dtypes.canonicalize_dtype(dtype)
 
   if _can_call_numpy_array(object):
+    if dtypes.is_python_scalar(object):
+      object = dtypes.coerce_to_array(object)
+    # TODO(jakevdp): falling back to numpy here fails to overflow for lists containing
+    # large integers; see discussion in https://github.com/google/jax/pull/6047.
     object = _np_array(object, dtype=dtype, ndmin=ndmin, copy=False)
 
   assert type(object) not in dtypes.python_scalar_dtypes
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -59,21 +59,15 @@ def PRNGKey(seed: int) -> jnp.ndarray:
     key is constructed from a 64-bit seed by effectively bit-casting to a pair
     of uint32 values (or from a 32-bit seed by first padding out with zeros).
   """
-  if np.shape(seed):
+  seed_arr = jnp.asarray(seed)
+  if seed_arr.shape:
     raise TypeError(f"PRNGKey seed must be a scalar; got {seed!r}.")
-  if not np.issubdtype(np.result_type(seed), np.integer):
+  if not np.issubdtype(seed_arr.dtype, np.integer):
     raise TypeError(f"PRNGKey seed must be an integer; got {seed!r}")
 
-  # Explicitly cast to int64 for JIT invariance of behavior on large ints.
-  if isinstance(seed, int):
-    seed = np.int64(seed)  # type: ignore[assignment]
-  # Converting to jnp.array may truncate bits when jax_enable_x64=False, but this
-  # is necessary for the sake of JIT invariance of the result for such values.
-  seed = jnp.asarray(seed)
-
   convert = lambda k: lax.reshape(lax.convert_element_type(k, np.uint32), [1])
-  k1 = convert(lax.shift_right_logical(seed, lax._const(seed, 32)))
-  k2 = convert(jnp.bitwise_and(seed, np.uint32(0xFFFFFFFF)))
+  k1 = convert(lax.shift_right_logical(seed_arr, lax._const(seed_arr, 32)))
+  k2 = convert(jnp.bitwise_and(seed_arr, np.uint32(0xFFFFFFFF)))
   return lax.concatenate([k1, k2], 0)
 
 def _is_prng_key(key: jnp.ndarray) -> bool:
--- a/jax/dtypes.py
+++ b/jax/dtypes.py
@@ -21,7 +21,7 @@
 
 
 import functools
-from typing import Dict
+from typing import Any, Dict
 
 import numpy as np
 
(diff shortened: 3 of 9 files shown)


jax/dtypes.py
```diff
@@ -105,14 +113,52 @@ def scalar_type_of(x):
...
+
 def coerce_to_array(x):
   """Coerces a scalar or NumPy array to an np.array.
 
   Handles Python scalar type promotion according to JAX's rules, not NumPy's
   rules.
   """
-  dtype = python_scalar_dtypes.get(type(x), None)
-  return np.array(x, dtype) if dtype else np.array(x)
+  if type(x) in python_scalar_dtypes:
```
hawkinsp, line 157: What happens in cases like:

```
jnp.asarray([1 << 63])
```

`asarray` accepts, e.g., lists of Python scalars and I think this code will not handle them correctly.

Would it perhaps be possible to first let NumPy build an array and then cast it if needed?

Test this case?

jax/dtypes.py
```diff
@@ -105,14 +113,52 @@ def scalar_type_of(x):
...
+  np.int64
+  >>> _scalar_dtype_to_dtype(int, 0)
+  np.int64
+  >>> _scalar_dtype_to_dtype(int, 1 << 63)
+  np.uint64
+  """
+  dtype = python_scalar_dtypes[typ]
+  # For integers, we return unsigned if required to store the value.
+  if typ is int and value is not None:
+    unsigned_dtype = _UINT_DTYPES[np.iinfo(dtype).bits]
```
hawkinsp, line 141: I'm surprised we don't consider x64 mode here. I guess this is intentional for now? If so leave a TODO.

jax/_src/random.py
```diff
@@ -52,15 +52,19 @@ def PRNGKey(seed: int) -> jnp.ndarray:
...
 
-  # Explicitly cast to int64 for JIT invariance of behavior on large ints.
-  if isinstance(seed, int):
-    seed = np.int64(seed)  # type: ignore[assignment]
   # Converting to jnp.array may truncate bits when jax_enable_x64=False, but this
   # is necessary for the sake of JIT invariance of the result for such values.
-  seed = jnp.asarray(seed)
+  seed = jnp.array(seed)
+
+  # Strip weak type to avoid lax typecasting issues.
```
hawkinsp, line 66: Elaborate. What goes wrong? Is this a temporary or permanent fix?

jax/dtypes.py
jakevdp: Good point... let me think on that a bit.

jax/_src/random.py
jakevdp: Actually, I think this can be removed now that #6103 is in.

jax/dtypes.py
jakevdp: One reason I didn't do that from the beginning is that `np.array(1 << 100)` returns dtype `object` rather than overflowing. From JAX's standpoint, in this case it's hard to know whether an array with object dtype implies an overflow error, or something different (e.g. `np.array([1, [1]])` also results in `dtype=object`)

jakevdp: Hmm... this actually has enough weird corners that I might step back and write a brief design doc

jax/dtypes.py
jakevdp: Yeah - the fact that we don't consider X64 mode in this utility I think has to do with this TODO: https://github.com/google/jax/blob/2947f562d53ae7d4f6ba9e87d9ccddca2f84a4d7/jax/dtypes.py#L56-L59

The reason I didn't canonicalize the type before checking the value is that I'm planning to address integer overflow in X32 mode separately (#5952). In fact, it was exploring that that led me to realize that the X64 mode behavior was inconsistent, so I decided to do this PR first.

jax/_src/random.py
jakevdp: Removed.

jax/dtypes.py
jakevdp: I fixed the list case here, and ensured that lists are no longer passed to `coerce_to_array`.

jax/_src/numpy/lax_numpy.py
```diff
@@ -2893,7 +2893,14 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
...
 
-  if _can_call_numpy_array(object):
+  leaves = tree_leaves(object)
+  can_call_numpy_array = not _any(
+      isinstance(leaf, (core.Tracer, DeviceArray)) for leaf in leaves)
+
+  if can_call_numpy_array:
+    # Construct array with numpy, but use JAX promotion rules to determine the output dtype.
+    if dtype is None and leaves:
+      dtype = dtypes.result_type(*map(dtypes.coerce_to_array, leaves))
```
hawkinsp, line 2903: My main concern here is that the reason for this path is so `jnp.array(...large Python value...)` is fast. We now do per-object Python work.

What does this PR do to the timing of, say, passing a large list to `jnp.array`?

If the effect is large we may ultimately need to implement our own version of this code in C++.

hawkinsp (commented):
These semantics should be documented.

jax/_src/numpy/lax_numpy.py
jakevdp: I haven't benchmarked it, but `_can_call_numpy_array` already does a `tree_map` across all inputs, so I don't think an additional `tree_map` will be significant.

hawkinsp: My concern isn't the `tree_map`, but instead the per-element call to `coerce_to_array`.

jakevdp: I see - `coerce_to_array` is much more costly than a type check, so yes this does come at a computational cost.

I suppose our tradeoff here, short of pushing array creation to C++, is speed vs. correctness.

hawkinsp: And knowing what the benchmark impact is on a simple microbenchmark would probably go a long way to answering how much we care.

jakevdp: On master:
```python
In [3]: L = list(range(1000))
In [4]: %timeit np.array(L)
55.8 µs ± 809 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
In [5]: %timeit jnp.array(L)
345 µs ± 4.89 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
on this branch:
```python
In [6]: %timeit jnp.array(L)                                                                                                                                                             
5.8 ms ± 36.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So about 10x slower to call `coerce_to_array` on each element.

jakevdp: What we're trying to protect against here is cases where numpy is applying its casting rules to list inputs, e.g.
```python
>>> np.array([0, 1 << 63]).dtype                                                                                                                                                     
dtype('float64')
```
Under the proposal here, this should be an overflow error. I don't know how to do that without adding some per-item Python logic... perhaps we could try for something less heavy-weight than `coerce_to_array`. Any thoughts?

jakevdp: We can reduce the time by aboud a factor of 3 if we only call `result_type` rather than `_coerce_to_array`, at which point numpy's handling of large integers will lead to an OverflowError if the result is int64. But if we want overflow logic to eventually be implemented in X32 mode, that solution will no longer work.

jakevdp: The other options I see are:
- implement list-to-array parsing logic into C++, which sounds pretty heavy
- accept and document the fact that `jnp.array()` called on list inputs will not have consistent behavior.

jakevdp: For now I restored the previous behavior and left a TODO.

jax/dtypes.py
```diff
@@ -60,6 +60,14 @@
     np.dtype('complex128'): np.dtype('complex64'),
 }
 
+_UINT_DTYPES = {
```
hawkinsp, line 63: This is no longer used?

hawkinsp (approved):

jax/dtypes.py
jakevdp: Good catch - fixed!

PR #6048, opened 2021-03-12
Make jax_debug_nans and jax_debug_infs work with pmap, xmap, and pjit.

### LINKED ISSUE #6044: Debugging nan when using 'pmap'
I was trying to set up error propagation by returning nan values as explained in this [issue](https://github.com/google/jax/issues/4257) and I noticed that to make it work for a _pmap_-ed function I also need to _jit_ it.

So something like this doesn't give me an error:
```python
from jax.config import config
config.update("jax_debug_nans", True)

@jax.pmap
def my_func(x):
  return jnp.nan

x = jnp.zeros((1,))
y = my_func(x)
```

While by adding `jax.jit` I get the error (which is what I want): 
```python
from jax.config import config
config.update("jax_debug_nans", True)

@jax.jit
@jax.pmap
def my_func(x):
  return jnp.nan

x = jnp.zeros((1,))
y = my_func(x)
```

So basically a _pmap_-ed function doesn't behave as a _jit_-ed function or a plain function in regards to debugging nans with `jax_debug_nans`. Adding `jax.jit` is an easy fix but I think it would help if this was documented, maybe [here](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#Debugging-NaNs) somewhere. Personally it took me quite a bit of time to figure out what was going on.

skye:
Note that unlike in the jit case, this doesn't rerun the function in
op-by-op mode when it finds a nan, since we don't have op-by-op
parallel execution yet :)

This change doesn't appear to regress performance:

```
---------Benchmark summary for pmap_shard_outputs---------
  nouts    nshards       mean      %std    relative    mean/baseline
-------  ---------  ---------  --------  ----------  ---------------
     10          8   0.105598  5.06671      1               1.00693
    100          8   0.287756  0.870751     2.72502         0.973204
    500          8   1.20119   0.823624    11.3752          0.955185
   1000          8   2.56071   0           24.2497          0.983063
   5000          8  12.909     0          122.247           0.965925
    100          2   0.173727  5.15115      1.64518         0.98918
    100          4   0.207774  3.71411      1.9676          0.955849
    100          8   0.286103  1.60243      2.70937         0.971869
    100        100   2.34168   0           22.1755          0.904475
    100        500  15.9558    0          151.1             1.00483
```

Fixes #6044

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -1165,6 +1165,9 @@ def partitioned_sharding_spec(num_partitions: int,
 def execute_replicated(compiled, backend, in_handler, out_handler, *args):
   input_bufs = in_handler(args)
   out_bufs = compiled.execute_sharded_on_local_devices(input_bufs)
+  if xla.needs_check_special():
+    for bufs in out_bufs:
+      xla.check_special("parallel computation", bufs)
   return out_handler(out_bufs)
 
 
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -357,7 +357,7 @@ def _execute_compiled_primitive(prim, compiled, result_handler, *args):
   device, = compiled.local_devices()
   input_bufs = list(it.chain.from_iterable(device_put(x, device) for x in args if x is not token))
   out_bufs = compiled.execute(input_bufs)
-  check_special(prim, out_bufs)
+  check_special(prim.name, out_bufs)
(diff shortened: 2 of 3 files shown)


mattjj (approved):
Thanks for adding this!

zhaopku:
Are there any updates on this issue? Looks like JAX_DEBUG_NANS is still not working for pjit.

PR #6049, opened 2021-03-12
autodidax: add linearize and vjp

mattjj:
Progress:
  - [x] `linearize` code _without_ `jit` support
  - [x] `vjp` and `grad` code
  - [x] `linearize` code _with_ `jit` support (i.e. partial eval of `xla_call`)
  - [x] `linearize` and partial-eval text explanation
  - [x] `vjp` and `grad` text explanation

```
Python                           1            417            654           1374
```

So, over the original goal of ~1000 lines! But this includes more features than originally planned, especially around typechecking and robustness, but also things like pytrees and pretty-printing.

### DIFF
--- a/docs/autodidax.ipynb
+++ b/docs/autodidax.ipynb
@@ -27,10 +27,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_next_cell": 0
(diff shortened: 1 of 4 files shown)


ljppro:
Do we have further update plan on this doc, for example, pmap? It would be very valuable!
Many thanks!

PR #6057, opened 2021-03-13
fix the precision of complex jax.random.normal

### LINKED ISSUE #6052: complex precision not respected in jax.random.normal
I am trying to generate random complex data of different precision and noticed the following:

```python
from jax.config import config
import jax

print(jax.random.normal(jax.random.PRNGKey(1), dtype=jax.numpy.complex128).dtype)
# complex128

print(jax.random.normal(jax.random.PRNGKey(1), dtype=jax.numpy.complex64).dtype)
# complex128
```
even though the second one supposed to generate `complex64`

With the corresponding real types (float32, float64) this works fine.

inailuig:
Fixes https://github.com/google/jax/issues/6052

example:
``` python
from jax.config import config
import jax
print(jax.random.normal(jax.random.PRNGKey(1), dtype=jax.numpy.complex64).dtype)
```
output before: `complex128`
output now: `complex64`

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -634,7 +634,7 @@ def _normal(key, shape, dtype) -> jnp.ndarray:
     dtype = dtypes.dtype_real(dtype)
     _re = _normal_real(key_re, shape, dtype)
     _im = _normal_real(key_im, shape, dtype)
-    return 1 / sqrt2 * (_re + 1j * _im)
+    return (_re + 1j * _im) / sqrt2
   else:
     return _normal_real(key, shape, dtype) # type: ignore
 
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -223,6 +223,7 @@ class LaxRandomTest(jtu.JaxTestCase):
     for samples in [uncompiled_samples, compiled_samples]:
       self._CheckKolmogorovSmirnovCDF(jnp.real(samples), scipy.stats.norm(scale=1/np.sqrt(2)).cdf)
       self._CheckKolmogorovSmirnovCDF(jnp.imag(samples), scipy.stats.norm(scale=1/np.sqrt(2)).cdf)
+      self.assertEqual(dtype, samples.dtype)
 
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_dtype={}".format(np.dtype(dtype).name), "dtype": dtype}


jax/_src/random.py
```diff
@@ -628,23 +628,21 @@ def normal(key: jnp.ndarray,
 @partial(jit, static_argnums=(1, 2))
 def _normal(key, shape, dtype) -> jnp.ndarray:
   if dtypes.issubdtype(dtype, np.complexfloating):
-    sqrt2 = np.array(np.sqrt(2), dtype)
-
     key_re, key_im = split(key)
     dtype = dtypes.dtype_real(dtype)
     _re = _normal_real(key_re, shape, dtype)
     _im = _normal_real(key_im, shape, dtype)
-    return 1 / sqrt2 * (_re + 1j * _im)
```
inailuig, line 637: the 1 here was the culprit.

jakevdp: I think it might be confusing to move the sqrt2 factor out of the helper function, because it no longer computes what it says it computes (`normal_real`). Instead, you could get around the numpy type promotion issue by making this line
```python
(_re + 1j * _im) / sqrt2
```
and keeping everything else the same. Since this is JIT-compiled, the multiplication and division by the same constant will be optimized out of the final computation.

tests/random_test.py
```diff
@@ -223,6 +223,8 @@ def testNormalComplex(self, dtype):
     for samples in [uncompiled_samples, compiled_samples]:
       self._CheckKolmogorovSmirnovCDF(jnp.real(samples), scipy.stats.norm(scale=1/np.sqrt(2)).cdf)
       self._CheckKolmogorovSmirnovCDF(jnp.imag(samples), scipy.stats.norm(scale=1/np.sqrt(2)).cdf)
+      if config.x64_enabled:
+          self.assertTrue(samples.dtype == dtype)
```
jakevdp, line 227: You may as well check the dtype in all cases, and please use `self.assertEqual(samples.dtype, dtype)`, as it results in better error messages on failure.

jax/_src/random.py
inailuig: done. I am amazed that these kinds of constants can be completely optimized away :D

tests/random_test.py
inailuig: done.

jakevdp (approved):
Looks great, thanks for tracking this down!

Issue #6058, opened 2021-03-13
Cross Product Error in 64bit mode


oracle3001:
When I try a simple cross product calculation using Jax in 64bit on a Windows 10 machine, I get the following error.
 
```python
import jax.numpy as jnp
from jax.config import config
config.update('jax_enable_x64', True)

x = jnp.array((1.0,1.0,1.0))
y =  jnp.array((0.0,0.0,0.0))
z = jnp.cross(x,y)
```

```
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm 2020.3.3\plugins\python\helpers\pydev\_pydevd_bundle\pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\_src\numpy\lax_numpy.py", line 3938, in cross
    return _cross(a, b, axisa, axisb, axisc)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\_src\traceback_util.py", line 139, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\api.py", line 416, in f_jitted
    return cpp_jitted_f(context, *args, **kwargs)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\api.py", line 297, in cache_miss
    out_flat = xla.xla_call(
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\core.py", line 1394, in bind
    return call_bind(self, fun, *args, **params)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\core.py", line 1385, in call_bind
    outs = primitive.process(top_trace, fun, tracers, params)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\core.py", line 1397, in process
    return trace.process_call(self, fun, tracers, params)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\core.py", line 625, in process_call
    return primitive.impl(f, *tracers, **params)
  File "C:\Users\Adam\anaconda3\envs\tensorflow\lib\site-packages\jax\interpreters\xla.py", line 586, in _xla_call_impl
[...]

PR #6068, opened 2021-03-15
BUG: fix jnp.result_type for non-canonical weak types

### LINKED ISSUE #6051: Non-canonical weak types can lead to failed promotion
```python
from jax import lax
x = lax.convert_element_type(1, 'int16', weak_type=True)
print(x)
# DeviceArray(1, dtype=int16)

print(x + 1)
# TypeError: add requires arguments to have the same dtypes, got int16, int32.
```
The reason this happens is due to something @mattjj noted in #6000; `dtypes.result_type` essentially canonicalizes all weak types before doing any promotion

jakevdp:
Fixes #6051 (and most likely also #6058)

To summarize, the problem here is that weak dtypes are currently canonicalized by `jnp.result_type`. Thus, with X64 disabled:
```python
>>> jnp.result_type(1)
dtype('int32')
```
The issue is that this can cause problems when the weak types are represented by non-canonical dtypes; for example:
```python
>>>  x = lax.convert_element_type(0, 'int16', weak_type=True)
>>> jnp.result_type(x)
dtype('int32')
>>> x.dtype == jnp.result_type(x)
False
```
This is particularly problematic because throughout the code base we use `result_type` as a shortcut to get the dtype of a value, which led to things like this in JAX 0.2.9:
```python
>>> x < 1
TypeError: lt requires arguments to have the same dtypes, got int16, int32.
```
It also comes up frequently with the experimental `enable_x64()`/`disable_x64()` context managers, because these lead to more non-cannonical weak types floating aroung.

To make `jnp.result_type` consistent, if all types in a promotion are weak, we promote the underlying type and then apply a weak type flag at the end. This preserves existing weak promotion behavior, but avoids the problematic canonicalization of the weak dtype's underlying representations.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -272,20 +272,18 @@ def _promote_dtypes(*args):
   if len(args) < 2:
     return args
   else:
-    to_dtype_raw = dtypes._result_type_raw(*args)
-    weak_type = to_dtype_raw in set(dtypes._weak_types)
-    to_dtype = dtypes.canonicalize_dtype(to_dtype_raw)
+    to_dtype, weak_type = dtypes._lattice_result_type(*args)
+    to_dtype = dtypes.canonicalize_dtype(to_dtype)
     return [lax.convert_element_type(x, to_dtype, weak_type) for x in args]
 
 def _promote_dtypes_inexact(*args):
   """Convenience function to apply Numpy argument dtype promotion.
 
   Promotes arguments to an inexact type."""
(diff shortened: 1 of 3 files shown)


jakevdp:
Note that this addresses the issue that @mattjj pointed out with the initial approach in https://github.com/google/jax/pull/6000#pullrequestreview-609138262:
```python
from jax import config, dtypes, lax

x = lax.convert_element_type(2, 'int32', weak_type=True)
print(dtypes.result_type(x))     # int32
print(dtypes.result_type(x, x))  # int32
```

mattjj (approved):
LGTM, thanks!

PR #6069, opened 2021-03-15
WIP: experiment with CodeCov integration

jakevdp:
Not sure whether this is something we want... but I was looking at the flax CI setup and thought I'd try it out.

Looks like our current test coverage baseline is 78%.


jakevdp:
I realized that the metrics reported here are not particularly meaningful, because our test coverage requires runs in both X32 and X64 mode and CodeCov can only report results of a single test run. I'm going to close this for now.

PR #6087, opened 2021-03-16
Add extras_require to setup.py

jakevdp:
This partially solves the problem of `jaxlib` not being listed as an explicit dependency for jax installation. The behavior of `pip install jax` does not change. However, once this PR is part of a release, it will be possible to install JAX along with a compatible CPU version of jaxlib using:
```
$ pip install jax[cpu]
```
and to install JAX along with the most recent GPU version of jaxlib, specifying the CUDA version, using, e.g.:
```
$ pip install jax[cuda-112] -f https://storage.googleapis.com/jax-releases/jax_releases.html
```
Unfortunately, there is no way to specify the alternate package repository URL within the setup.py file; this is a deliberate choice made in the `pip` API.

This does add another task to the release checklist (updating the listed versions in `setup.py`). In the case of `minimum_jaxlib_version`, we could read this from its current declared value with a minor refactoring.

If we decide to merge this, we can update installation instructions in the README once this change is part of a release.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -12,5 +12,4 @@ pytest-xdist
 # Packages used for notebook execution
 matplotlib
 sklearn
-jaxlib
-.  # install jax itself from current directory.
+.[cpu]  # Install jax from the current directory; jaxlib from pypi.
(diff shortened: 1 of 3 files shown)


skye (approved):
Cool! I think the extra release steps are fine, just be sure to add them to the playbook instructions.

PR #6088, opened 2021-03-16
Specify minimum jaxlib version in a single location

jakevdp:
This is a followup to #6087; this makes it so that `minimum_jaxlib_version` need only be specified in one place, `jax/version.py`. The other locations read the value from there. This avoids the potential for these values to become out of sync.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,13 +14,15 @@
 
 from setuptools import setup, find_packages
 
-__version__ = None
-_minimum_jaxlib_version = '0.1.62'
+# The following should be updated with each new jaxlib release.
 _current_jaxlib_version = '0.1.64'
 _available_cuda_versions = ['101', '102', '110', '111', '112']
 
+_dct = {}
 with open('jax/version.py') as f:
-  exec(f.read(), globals())
+  exec(f.read(), _dct)
+__version__ = _dct['__version__']
+_minimum_jaxlib_version = _dct['_minimum_jaxlib_version']
 
 setup(
     name='jax',
@@ -37,6 +39,9 @@ setup(
         'opt_einsum',
     ],
     extras_require={
+        # Minimum jaxlib version; used in testing.
+        'minimum-jaxlib': [f'jaxlib=={_minimum_jaxlib_version}'],
+
         # CPU-only jaxlib can be installed via:
         # $ pip install jax[cpu]
         'cpu': [f'jaxlib>={_minimum_jaxlib_version}'],
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -1,10 +1,10 @@
 flake8
 # For now, we pin the numpy version here
 numpy>=1.16
- # Must be kept in sync with the minimum jaxlib version in jax/lib/__init__.py
-jaxlib==0.1.62
 mypy==0.790
 pillow
 pytest-benchmark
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -14,13 +14,15 @@
 
 from setuptools import setup, find_packages
 
-__version__ = None
-_minimum_jaxlib_version = '0.1.62'
+# The following should be updated with each new jaxlib release.
```
hawkinsp, line 17: Can you please add a comment to `jax/lib/version.py` explaining that this must be done?

Or alternatively could we somehow avoid the redundancy?

jakevdp: I did add this as a step in our release playbook.

I'm not sure it's possible to remove any redundancy, because the current jaxlib version is not available anywhere else in the files associated with a release (the `jaxlib` directory is not included in the `jax` tarball), and the CUDA versions used to build the released wheels are a property of the repo at build time, not the repo at HEAD.

hawkinsp: A comment in the file is certainly possible. I'm more likely to see it if it's right next to the version number.

However, note that `jax/lib/` (N.B. not `jaxlib`) probably *is* in our `jax` tarball. It gets copied into `jaxlib` by the build script, but it's probably present in both.

hawkinsp: I misremembered slightly. The `version.py` file is in `jaxlib/`. But it doesn't have to be that way. We could easily include it in `jax` as well, you'd just have to modify the wheel build script to reflect the new location.

jakevdp: Added a comment to `jaxlib/version.py`. I think that's our best fix for now.

jakevdp:
Hey @hawkinsp - let me know if you have other thoughts on this.

hawkinsp (approved):

reneleonhardt:
Why is the variable called `minimum_jaxlib_version` but the operator is `==` instead of `>=`?

jakevdp:
@reneleonhardt could you be more specific about which line of code you're asking about?

setup.py
```diff
@@ -37,6 +39,9 @@
         'opt_einsum',
     ],
     extras_require={
+        # Minimum jaxlib version; used in testing.
+        'minimum-jaxlib': [f'jaxlib=={_minimum_jaxlib_version}'],
```
reneleonhardt, line 43: `== _minimum_jaxlib_version` instead of `>=`

jakevdp: This is where we specify what is installed if you run `pip install jax[minimum-jaxlib]`. The intent is that this installs exactly the specified minimum jaxlib, not that it will install any jaxlib larger than the minimum.

reneleonhardt: Interesting, I wouldn't have guessed when reading "minimum version", only after searching the code I found the equals operator.

jakevdp: The code you’re pointing to is mainly used for testing. For testing purposes it is useful to be able to install exactly the specified minimum version. In more general contexts, any version larger than the minimum is suitable, and the version requirements reflect that. Does that make sense?

Edit: here's the more general installation context, which uses `>=` as you expect: https://github.com/jax-ml/jax/blob/31de82bfe8f64d4c51873badda302d96ad9dad3c/setup.py#L64

reneleonhardt: It's okay :)

PR #6100, opened 2021-03-17
Make xmap tutorial public

apaszke:
Note that xmap is still experimental, so treat this more as a preview
than as documentation.


jakevdp:
This has two large PNGs embedded in the source, which causes them to be duplicated in the ipynb and md. Could we host them in `docs/_static` instead?

apaszke:
Hosting in `_static` sounds good, but I'm not sure how should I link to `_static` from the notebook? Any tips?

Also, I did check our other notebooks before sending this patch and plenty of them seem to embed pngs inline, so I assumed this is the way to go.

PR #6108, opened 2021-03-17
Annotate scatter and random

NeilGirdhar:


jax/_src/ops/scatter.py
```diff
@@ -15,11 +15,21 @@
...
+
+import numpy as np
+
 from jax import lax
 from jax._src.numpy import lax_numpy as jnp
 from jax._src import util
 
 
+Index = Union[int, slice, Sequence[int], np.ndarray, 'ellipsis',
+              Tuple[Union[int, slice, Sequence[int], np.ndarray, 'ellipsis'], ...]]
```
jakevdp, line 28: This also needs `None` to account for `np.newaxis`.

It might be more readable to express it as something like:
```python
SingleIndex = Union[int, slice, Sequence[int], np.ndarray, 'ellipsis', None]
Index = Union[SingleIndex, Tuple[SingleIndex, ...]]
```

jax/_src/ops/scatter.py
```diff
@@ -15,11 +15,21 @@
...
+Scalar = Union[float, int]
+Numeric = Union[jnp.ndarray, Scalar]
```
jakevdp, line 30: Type checking `jnp.ndarray` doesn't actually work with mypy at the moment, I think. We get around this elsewhere by aliasing `Array = Any` and using that instead.

jax/_src/ops/scatter.py
NeilGirdhar: I like the composite definition.  Very nice.

I'm not sure about `None` though.  Which scatter function supports adding axes?  It appears that not only do they not add axes, but they sliently ignore them!

```python
z = jnp.zeros((4, 4))
index_add(z, (None, ...), 1).shape  # (4, 4)!!
```
Looks like a bug to be honest.

jax/_src/ops/scatter.py
NeilGirdhar: Sure, let's do that.  I was just copying some of the other files that are annotating with `jnp.ndarray`.

In any case, it appears that there are circular imports because `_src/numpy/lax_numpy.py` import `ops`, which imports `_src/ops/scatter.py`, and something will have to be done about that.  For example, `ndarray` could be moved to its own file.

I'll update to Array = Any as you suggest.

jax/_src/ops/scatter.py
jakevdp: I think by analogy to numpy's index update, it makes sense that `None` is a valid input, and should be ignored:
```python
import numpy as np
x = np.arange(12).reshape(3, 4)
x[:, 1, None] += 10
print(x)
# array([[ 0, 11,  2,  3],
#        [ 4, 15,  6,  7],
#        [ 8, 19, 10, 11]])

import jax.numpy as jnp
x = jnp.arange(12).reshape(3, 4)
print(x.at[:, 1, None].add(10))
# [[ 0 11  2  3]
#  [ 4 15  6  7]
#  [ 8 19 10 11]]
```

jax/_src/ops/scatter.py
jakevdp: Hmm... I'm not sure. From what I understand, variables returned from a function annotated with `jnp.ndarray` will not typecheck correctly further in the program because, for example, jnp.ndarray doesn't define all the attributes and methods that are available on a JAX DeviceArray. It may be that nobody has hit this because nobody has tried using these functions in annotated code. But I suppose if it's not causing people issues it's not worth spending time to fix!

jax/_src/ops/scatter.py
NeilGirdhar: Okay, great!  I added `None` to `SingleIndex` to match the behavior.

jax/_src/ops/scatter.py
NeilGirdhar: Okay, we'll leave it then in the other file if that's what you think is best. :smile:

jax/_src/ops/scatter.py
jakevdp: It turns out the `None` is not ignored; for example, this fails without it:
```python
jnp.arange(4).at[:, None].max(jnp.ones((4, 1)))
```

NeilGirdhar:
@jakevdp Thanks for the quick review.  Since I made a few changes since yesterday, I'll just throw up some comments to make it easier for you to notice what I've done. :smile:

jax/_src/ops/scatter.py
```diff
@@ -15,11 +15,27 @@
...
+
+import numpy as np
+
 from jax import lax
 from jax._src.numpy import lax_numpy as jnp
 from jax._src import util
 
 
+Array = Any
+if sys.version_info >= (3, 10):
```
NeilGirdhar, line 29: I couldn't get ellipsis type annotating, and [it appears](https://bugs.python.org/issue41810) that it won't really work until Python 3.10.  This check switch doesn't do anything yet, but in a year when JAX supports 3.10, it should save the trouble of an extra PR.

jax/_src/ops/scatter.py
```diff
@@ -15,11 +15,27 @@
...
+    from typing import EllipsisType
+    SingleIndex = Union[None, int, slice, Sequence[int], np.ndarray, EllipsisType]
+else:
+    SingleIndex = Union[None, int, slice, Sequence[int], np.ndarray]
+Index = Union[SingleIndex, Tuple[SingleIndex, ...]]
+Scalar = Union[complex, float, int, np.number]
+Numeric = Union[Array, Scalar]
```
NeilGirdhar, line 36: Often the right-hand side can be scalar, so I defined an appropriate scalar type.

jax/_src/random.py
```diff
@@ -36,6 +36,15 @@
 from jax._src.util import prod
 
 
+IntOrFloat = Union[float, int, np.integer, np.floating]
+IntFloatOrArray = Union[jnp.ndarray, IntOrFloat]
+IntOrArray = Union[jnp.ndarray, int, np.integer]
```
NeilGirdhar, line 41: I used `IntFloatOrArray` and `IntOrArray` in many places, which was sometimes broader than the existing annotation.  I tested to make sure that the functions support it.

jax/_src/random.py
```diff
@@ -36,6 +36,15 @@
...
+# TODO: Import or define these to match
+# https://github.com/numpy/numpy/blob/main/numpy/typing/_dtype_like.py.
+DTypeLikeInt = Any
+DTypeLikeFloat = Any
```
NeilGirdhar, line 45: I replaced some of the `np.dtype` annotations with these aliases currently set to `Any`.  Numpy added these types internally six weeks ago, which cover all the bases.  The most common thing to pass as dtype is probably `np.float32` or `'f'` and neither of those are even an `np.dtype`.

jax/_src/random.py
```diff
@@ -407,9 +416,9 @@ def _uniform(key, shape, dtype, minval, maxval) -> jnp.ndarray:
 
 def randint(key: jnp.ndarray,
             shape: Sequence[int],
-            minval: Union[int, jnp.ndarray],
-            maxval: Union[int, jnp.ndarray],
-            dtype: np.dtype = dtypes.int_):
+            minval: IntOrArray,
+            maxval: IntOrArray,
```
NeilGirdhar, line 420: Broaden to support `np.int32`, etc.

jax/_src/random.py
```diff
@@ -572,7 +586,7 @@ def choice(key, a, shape=(), replace=True, p=None):
     a = int(a)
   else:
     a = _asarray(a)
-  n_inputs = a if np.ndim(a) == 0 else len(a)
+  n_inputs = int(a) if np.ndim(a) == 0 else len(a)  # type: ignore[arg-type]
```
NeilGirdhar, line 589: This cast ensures that from here on `n_inputs` is an `int` (not an `np.int32` or a `jnp.array` of shape `()`)

jax/_src/random.py
```diff
@@ -1254,13 +1286,16 @@ def categorical(key, logits, axis=-1, shape=None):
   if shape is None:
     shape = batch_shape
   else:
+    shape = tuple(shape)
```
NeilGirdhar, line 1289: All of the other functions in this file support shapes of any sequence type.  `categorical` only supported `tuple`.   This cast repairs that.

jax/_src/random.py
```diff
@@ -1315,12 +1352,15 @@ def _logistic(key, shape, dtype):
...
+def pareto(key: jnp.ndarray,
+           b: IntFloatOrArray,
+           shape: Optional[Sequence[int]] = None,
+           dtype: DTypeLikeFloat = dtypes.float_) -> jnp.ndarray:
   """Sample Pareto random values with given shape and float dtype.
 
   Args:
     key: a PRNGKey used as the random key.
-    a: a float or array of floats broadcast-compatible with ``shape``
+    b: a float or array of floats broadcast-compatible with ``shape``
```
NeilGirdhar, line 1363: Fix docstring typo.

jax/_src/ops/scatter.py
NeilGirdhar: Yes, good point!

hawkinsp: I think the key is that the update value is broadcast to the index shape. Adding a leading `None` to the index doesn't  do a whole lot because broadcasting implicitly does that.

NeilGirdhar:
> Even in the future when Array is not aliased to Any, scalar inputs are in most places treated as identical to zero-dimensional arrays, 

Yeah, I see how scalar inputs are treated as zero-dimensional arrays by the code.  We want the type annotation to accept such scalar inputs, which is why I added the various scalar types.  This union is an improved version of [the one from `chex`](https://github.com/deepmind/chex/blob/master/chex/_src/pytypes.py).  Don't we need things like the following to work?
```python
def uniform(key, minval=0, maxval=1)
def uniform(key, minval=0.0, maxval=1.0)
def uniform(key, minval=np.int32(0), maxval=np.int32(1))
def uniform(key, minval=some_tracer, maxval=some_other_tracer)
```

jakevdp:
> Don't we need things like the following to work?

Yes - I was trying to point out that currently those do work, because `Array` is aliased to `Any`. Also, If sometime in the future `Array` is more narrowly specified, it's not clear to me that we wouldn't want `Array` itself to be this unioned type, becaue virtually anywhere that arrays are accepted, scalars are as well.

Additionally, it seems a bit strange to have a union like `IntFloatOrArray`, when the array type does not specify any dtype. Perhaps once the type system can handle specifying arrays with particular dtypes, it would also make sense to specify particular scalar types as well. Or perhaps those dtyped array unions will already include the scalar dtypes.

I don't know what a future fully-fleshed-out array type system will look like, but given all of this, it seems to me that adding various flavors of `ArrayOrScalar` unions currently adds complication without any apparent advantage.

NeilGirdhar:
> Yes - I was trying to point out that currently those do work, 

Okay, I think I see better what you're saying.  In the case of `uniform`, which I was showing above, I broadened the existing definition of `uniform`, which currently is annotated to only accept `Union[float, jnp.ndarray]`.  You're right that it could be broadened all the way to `All`.

> becaue virtually anywhere that arrays are accepted, scalars are as well.

Right.  However, a few examples of where we know that we're returning `jnp.ndarray` (and not one of the scalars) are the return values of the functions in `random.py` and the first parameter of the scatter functions.  Also, I tried to be careful about when it's an integer array, a float or integer array, or a general scalar array (float, integer, or complex).  The scatter functions work with complex arrays, but none of the curent probability functions do.

> Perhaps once the type system can handle specifying arrays with particular dtypes, it would also make sense to specify particular scalar types as well

Yes, I see your point.  I was just following the existing annotations like `Union[float, jnp.ndarray]`.  It won't be long before Numpy has its array generic on the dtype (I think 1.21 has it?)  And then presumably Jax will follow.  I thought it might make the future easier if it's already done and all you need to do is update `IntOrArray = Union[int, np.integer, jnp.ndarray[DTypeLikeInt]]` or however it will be done.

> I don't know what a future fully-fleshed-out array type system will look like, but given all of this, it seems to me that adding various flavors of ArrayOrScalar unions currently adds complication without any apparent advantage.

I got rid of `IntOrFloat` since we didn't need that.

I guess I could collapse `IntFloatOrArray` and `IntOrArray` into `Numeric` (which accepts complex scalars) then?  But the benefit of keeping them separate is that we block floats from being passed where only integers are allowed.  Are you sure you want to lose that?

jakevdp:
There's another complication when it comes to JAX, though: any time one of these functions is traced (e.g. when used in `vmap`, `jit`, etc.) any of the inputs, including some of the things here marked as `Numeric`, may also be a `Tracer` object.

On the whole, we haven't yet figured out a good policy for delineating such things in JAX, which is another reason for the `Array = Any` catch-all for annotating quantities that may be traced, but this includes traced integers and traced floats... I think that's one reason it feels over-complicated to call out unions of Python scalars and array objects, because during tracing those things all look the same. I don't see a lot to be gained from adding a bunch of type-union boilerplate.

I think the cleanest thing for now would be to treat `Array` as a catch-all covering arrays, scalars, and tracers thereof.

I think the `Index` union is probably reasonable, though I'd probably lean toward something simpler, like `Union[None, slice, Array, EllipsisType]` (with `Array` covering arrays, scalars, and traced objects)

We could then replace things like `IntFloatOrArray` and `IntOrArray` with `Array`, to reduce the boilerplate.

What do you think?

NeilGirdhar:
@jakevdp Sorry for the late reply.  I'm really slow at context switches.

> I think the cleanest thing for now would be to treat Array as a catch-all covering arrays, scalars, and tracers thereof.

Yes, let's go ahead and do what you want.  I appreciate your taking the time to convince me :smile:  I want to just go over some of your points just so we're on the same page.

> On the whole, we haven't yet figured out a good policy for delineating such things in JAX

Fair enough.  I'm sure given the great design in the rest of JAX, your type annotation will be great.  I was just trying to mitigate some of your future work.

> Python's type annotation ecosystem has not yet grown a mature way to specify what type those tracers contain… it feels over-complicated to call out unions of Python scalars and array objects, because during tracing those things all look the same. 

Yup, I agree.  Can't wait for numpy and ultimately JAX to have annotate dtypes.

> I think the Index union is probably reasonable, though I'd probably lean toward something simpler, like Union[None, slice, Array, EllipsisType] (with Array covering arrays, scalars, and traced objects)

Oh!  I didn't realize you could slice with tracers.  Let's replace `np.array` with `Array`.  I don't think we should consider `Array` to subsume `int` or `Sequence[int]` since `Array` is used in other places where these types wouldn't fit, e.g., the first argument to `index_add`.

> We could then replace things like IntFloatOrArray and IntOrArray with Array, to reduce the boilerplate.

I'll do it your way, but I personally think that the definitions:
```python
IntFloatOrArray = Union[int, float, np.integer, np.floating, jnp.ndarray]
IntOrArray = Union[int, np.integer, jnp.ndarray]
```
have value right now in preventing many type errors even though I agree with your above point that "during tracing [tracers] all look the same."  I guess it's a question of how much you catch.  How about a compromise solution of
```python
RealArray = Array
IntegerArray = Array
```
No worries if that feels too complicated, I don't mind just going back to `Array` :smile:

jakevdp:
Looks good - thanks for iterating on this. I think if you rebase on master you should get the fix to the flake8 error.

PR #6109, opened 2021-03-17
Add programmatic profiling APIs.

skye:
Also bumps the minimum jaxlib version, since these APIs depend on newly-exposed xla_client functionality.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -18,7 +18,69 @@ Install specific nightly versions of TensorBoard, TensorBoard profiler, TensorFl
 pip install --upgrade tb-nightly==2.5.0a20201203 tbp-nightly==2.4.0a20201203 tf-nightly==2.5.0.dev20201203 tensorboard-plugin-wit==1.7.0
 ```
 
-### Usage
+### Programmatic capture
+
+You can instrument your code to capture a profiler trace via the
+{func}`jax.profiler.start_trace` and {func}`jax.profiler.stop_trace`
+methods. Call {func}`~jax.profiler.start_trace` with the directory to write
+trace files to. This should be the same `--logdir` directory used to start
+TensorBoard. Then, you can use TensorBoard to view the traces.
+
+For example, to take a profiler trace:
+
+```python
+import jax
+
+jax.profiler.start_trace("/tmp/tensorboard")
+
+# Run the operations to be profiled
+key = jax.random.PRNGKey(0)
+x = jax.random.normal(key, (5000, 5000))
+y = x @ x
+y.block_until_ready()
+
+jax.profiler.stop_trace()
+```
+
+Note the {func}`block_until_ready` call. We use this to make sure on-device
+execution is captured by the trace. See {ref}`async-dispatch` for details on why
+this is necessary.
+
+You can also use the {func}`jax.profiler.trace` context manager as an
+alternative to `start_trace` and `stop_trace`:
+
+```python
+import jax
+
+with jax.profiler.trace():
+  key = jax.random.PRNGKey(0)
+  x = jax.random.normal(key, (5000, 5000))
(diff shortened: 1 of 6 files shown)


docs/profiling.md
```diff
@@ -18,7 +18,69 @@ Install specific nightly versions of TensorBoard, TensorBoard profiler, TensorFl
 pip install --upgrade tb-nightly==2.5.0a20201203 tbp-nightly==2.4.0a20201203 tf-nightly==2.5.0.dev20201203 tensorboard-plugin-wit==1.7.0
 ```
 
-### Usage
+### Programmatic capture
+
+You can instrument your code to capture a profiler trace via the
+{func}`jax.profiler.start_trace` and {func}`jax.profiler.stop_trace`
+methods. Call `start_trace` with the directory to write trace files to. This
```
jakevdp, line 25: If you want, you can make this ``` {func}`~jax.profiler.start_trace` ``` which will render as **`start_trace()`** with a hyperlink to the source (the tilde removes the module path).

docs/profiling.md
```diff
@@ -18,7 +18,69 @@ Install specific nightly versions of TensorBoard, TensorBoard profiler, TensorFl
...
+# Run the operations to be profiled
+key = jax.random.PRNGKey(0)
+x = jax.random.normal(key, (5000, 5000))
+y = x @ x
+y.block_until_ready()
+
+jax.profiler.stop_trace()
+```
+
+Note the {func}`block_until_ready` call. We use this to make sure on-device
```
jakevdp, line 45: I think this has to be something like ``` {meth}`~jax.interpreters.xla._DeviceArray.block_until_ready` ``` in order to generate the link to the method docs.

jakevdp: ... though we may not even have `DeviceArray` methods in our docs

docs/profiling.md
```diff
@@ -76,10 +138,12 @@ from a running program.
...
 
-1. By default, the events in the trace viewer are mostly low-level internal JAX
-   functions. You can add your own events and functions by using
-   {func}`jax.profiler.TraceContext` and {func}`jax.profiler.trace_function` in
-   your code and capturing a new profile.
+### Adding custom trace events
+
+By default, the events in the trace viewer are mostly low-level internal JAX
+functions. You can add your own events and functions by using
+{func}`jax.profiler.TraceContext` and {func}`jax.profiler.trace_function` in
```
jakevdp, line 145: I think this should be ``` {class}`jax.profiler.TraceContext` ```

jakevdp (commented):
A few comments on the docs... I didn't look closely at the implementation

jax/_src/profiler.py
```diff
@@ -24,15 +26,87 @@ def start_server(port: int):
...
+  with _profile_state.lock:
+    if _profile_state.profile_session is None:
+      raise RuntimeError("No profile started")
+    _profile_state.profile_session.stop_and_export(_profile_state.log_dir)
+    _profile_state.profile_session = None
+    _profile_state.log_dir = None
+
+
+@contextmanager
+def take_trace(log_dir):
```
hawkinsp, line 87: I note that TF calls this `tf.profiler.Profile`. I think I like the form of the name more, but I find `Profile` misleading.

Perhaps `Trace`?

and maybe we should rename `TraceContext` to `TraceMe` or `TraceAnnotation`?

hawkinsp (commented):
Seems fine to me.

jax/_src/profiler.py
skye: As discussed offline, I went with the oh-so-tasteful `trace`.

I also renamed TraceContext --> TraceAnnotation and other similar renames.

hawkinsp (approved):

PR #6124, opened 2021-03-18
Make sublevel weak-referable, and enable the leak checker on sublevels.

LenaMartens:
Reimplement Sublevel to not inherit from `int`.
See [docs on weakref](https://docs.python.org/3/library/weakref.html#:~:text=Other%20built-in%20types%20such%20as%20tuple%20and%20int): "CPython implementation detail: Other built-in types
such as tuple and int do not support weak references even when subclassed."

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -688,7 +688,23 @@ class TraceStack:
     new.dynamic = self.dynamic
     return new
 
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Beautiful work!

PR #6127, opened 2021-03-18
Fix error when running a jvp of a jit of a custom_vjp.

LenaMartens:
Doing `jvp(jit(f), (3.,), (0.,))` where `f` has a `custom_vjp`.

Error before:
`NotImplementedError: XLA translation rule for primitive 'custom_lin' not found`

Error after:
`TypeError: can't apply forward-mode autodiff (jvp) to a custom_vjp function.`


mattjj (approved):
Thanks for catching this!

Issue #190, opened 2019-01-03
Flattening function like in Autograd


sscardapane:
I was reimplementing my old Autograd code in JAX, but I can't seem to find an equivalent of the "flatten" utility:
https://github.com/HIPS/autograd/tree/master/autograd/misc

Basically, given a list of the network's parameters `w,` I would need something like:

```
w_flat, unflattener = flatten(w)
```

where `w_flat` is a one-dimensional vector containing all parameters, and `unflattener` reverts this operation. If I understand correctly, this was used in Autograd at the beginning for the implementation of the optimizers, but it is not needed here. However, it is useful in many cases and also goes well with the idea of chainable transformations on data.

mattjj:
I added some simple utilities in #201, so you should be able to write

```python
from jax.flatten_util import ravel_pytree

w_flat, unflattener = ravel_pytree(w)
```

I changed the name from "flatten" to "ravel_pytree" because I think it's more descriptive and in-line with other names in JAX. At one point we had too many uses of "flatten" and "unflatten" to mean slightly different things.

Please re-open this issue if you have issues with that function, or if I missed the mark somehow.

jermwatt:
Is there a [flatten_func](https://github.com/HIPS/autograd/blob/master/autograd/misc/flatten.py) analog?

jermwatt:
@mattjj - I'm probably missing it, is a [flatten_func](https://github.com/HIPS/autograd/blob/master/autograd/misc/flatten.py) analog baked into JAX somewhere?  I may be missing it in the [flatten_util](https://github.com/google/jax/blob/master/jax/flatten_util.py) (?) or [tree_util](https://github.com/google/jax/blob/master/jax/tree_util.py) files?

mattjj:
Thanks for the ping. My notification setup isn't great. When in doubt, open a new issue, since those are much more visible. It also makes things like feature requests easier to track.

We haven't added at flatten_func to JAX, but it's probably a good idea to put it in flatten_util.py and make it call ravel_pytree. I think it should look about the same as Autograd's. We could call it ravel_func or flatten_func, I don't have a strong opinion.

Want to send us a PR with that change?

jermwatt:
Great! - will do.

PR #6136, opened 2021-03-19
generalize ravel_pytree to handle int types

### LINKED ISSUE #1028: `ravel_pytree` does not work with int input
After https://github.com/google/jax/pull/897, `vjp` does not accept int input. Because `ravel_pytree` uses `vjp` for ravel/unravel logic, something like `ravel_pytree([1])` does not work anymore. I would be nice if we are able to disable this check in `vjp`. Probably we can provide an optional kwargs `disable_check_inexact_input` in `vjp` function to maintain this behaviour.

cc @hawkinsp @neerajprad

### LINKED ISSUE #6135: unravel_pytree not functioning on integer input
Hi Jax-team,

There is a bug related to vjp of the flatten operation and integer input:
```
pt = {'foo': jnp.array([1,2,3]), 'bar': {'baz': jnp.array([5,6]), 'bork': jnp.array([7,8])}}
flat, unflatten = jax.flatten_util.ravel_pytree(pt)
unflatten(flat)
```

returns 

```
{'bar': {'baz': array([(b'',), (b'',)], dtype=[('float0', 'V')]),
  'bork': array([(b'',), (b'',)], dtype=[('float0', 'V')])},
 'foo': array([(b'',), (b'',), (b'',)], dtype=[('float0', 'V')])}
```

This is only if I provide integer input. A proposed fix by @mattjj would be to recast to float and raise a warning, we agreed I would file a bug here.

All the best!

Gabe

mattjj:
fixes #6135, fixes #1028. See also #190, #2056

### DIFF
--- a/jax/flatten_util.py
+++ b/jax/flatten_util.py
@@ -12,12 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
+
+import numpy as np
 
 from .tree_util import tree_flatten, tree_unflatten
-from ._src.util import safe_zip
+from ._src.util import safe_zip, unzip2
 
 import jax.numpy as jnp
-from jax.api import vjp
+from jax import dtypes
+from jax import lax
 
 zip = safe_zip
 
@@ -26,18 +30,40 @@ def ravel_pytree(pytree):
   """Ravel (i.e. flatten) a pytree of arrays down to a 1D array.
 
   Args:
-    pytree: a pytree to ravel.
+    pytree: a pytree of arrays and scalars to ravel.
 
   Returns:
     A pair where the first element is a 1D array representing the flattened and
-    concatenated leaf values, and the second element is a callable for
-    unflattening a 1D vector of the same length back to a pytree of of the same
-    structure as the input ``pytree``.
+    concatenated leaf values, with dtype determined by promoting the dtypes of
+    leaf values, and the second element is a callable for unflattening a 1D
+    vector of the same length back to a pytree of of the same structure as the
+    input ``pytree``. If the input pytree is empty (i.e. has no leaves) then as
+    a convention a 1D empty array of dtype float32 is returned in the first
+    component of the output.
+
+  For details on dtype promotion, see
+  https://jax.readthedocs.io/en/latest/type_promotion.html.
+
   """
   leaves, treedef = tree_flatten(pytree)
-  flat, unravel_list = vjp(_ravel_list, *leaves)
+  flat, unravel_list = _ravel_list(leaves)
   unravel_pytree = lambda flat: tree_unflatten(treedef, unravel_list(flat))
   return flat, unravel_pytree
 
-def _ravel_list(*lst):
-  return jnp.concatenate([jnp.ravel(elt) for elt in lst]) if lst else jnp.array([])
+def _ravel_list(lst):
+  if not lst: return jnp.array([], jnp.float32), lambda _: []
+  from_dtypes = [dtypes.dtype(l) for l in lst]
+  to_dtype = dtypes.result_type(*from_dtypes)
+  sizes, shapes = unzip2((jnp.size(x), jnp.shape(x)) for x in lst)
+  indices = np.cumsum(sizes)
+
+  def unravel(arr):
+    chunks = jnp.split(arr, indices[:-1])
+    with warnings.catch_warnings():
+      warnings.simplefilter("ignore")  # ignore complex-to-real cast warning
+      return [lax.convert_element_type(chunk.reshape(shape), dtype)
+              for chunk, shape, dtype in zip(chunks, shapes, from_dtypes)]
+
+  ravel = lambda e: jnp.ravel(lax.convert_element_type(e, to_dtype))
+  raveled = jnp.concatenate([ravel(e) for e in lst])
+  return raveled, unravel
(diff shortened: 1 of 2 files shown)


jax/flatten_util.py
```diff
@@ -26,18 +31,42 @@ def ravel_pytree(pytree):
...
+    vector of the same length back to a pytree of of the same structure as the
+    input ``pytree``.
+
+  For details on dtype promotion, see
+  https://jax.readthedocs.io/en/latest/type_promotion.html.
   """
   leaves, treedef = tree_flatten(pytree)
-  flat, unravel_list = vjp(_ravel_list, *leaves)
+  dts = [dtypes.dtype(l) for l in leaves]
+  dtype = functools.reduce(dtypes.promote_types, dts)
```
jakevdp, line 49: It would be more efficient here to call `dtypes.promote_types(*dts)`, and the lattice property ensures that it will give the same results 😎

jakevdp (commented):
Looks good - a couple minor things.

jax/flatten_util.py
mattjj: That would be nice, but iiuc [`dtypes.promote_types` only accepts two arguments](https://github.com/google/jax/blob/d75becbf676e392bcf08e91594e8495bbc39317d/jax/dtypes.py#L299), though `_least_upper_bound` accepts many.

Maybe I should call `_least_upper_bound` directly? What do you advise?

tests/tree_util_test.py
```diff
@@ -274,5 +277,49 @@ def testTransposeWithCustomObject(self):
...
+    tree_ = unravel(raveled)
+    self.assertAllClose(tree, tree_)
+
+  def testMixedFloatComplex(self):
+    tree = [jnp.array([1.], jnp.float32),
+            jnp.array([[1, 2 + 3j], [3, 4]], jnp.complex64)]
+    raveled, unravel = flatten_util.ravel_pytree(tree)
+    self.assertEqual(raveled.dtype, dtypes.promote_types(jnp.float32, jnp.complex64))
+    tree_ = unravel(raveled)
+    self.assertAllClose(tree, tree_)
```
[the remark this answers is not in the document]
mattjj: That's true but unfortunately `assertArraysEqual` only works on arrays, whereas here we're asserting that two trees are equal. I could flatten the trees and map `assertArraysEqual`, but given that extra complexity paired with the uncertainty about mixed float dtypes, I think I'd rather stick with just asserting allclose. Luckily we know the values in these tests are such that no errors will be introduced anyway.

I could set the tolerances to zero I suppose... WDYT?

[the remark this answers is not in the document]
mattjj: I set the tolerances to zero!

jax/flatten_util.py
jakevdp: Oh, sorry, I meant to say `dtypes.result_type(*dts)`

jakevdp: I'm not sure why `promote_types` is a separate implementation...

mattjj: Okay, I'll try that!

jax/flatten_util.py
```diff
@@ -26,18 +30,36 @@ def ravel_pytree(pytree):
...
   leaves, treedef = tree_flatten(pytree)
-  flat, unravel_list = vjp(_ravel_list, *leaves)
+  flat, unravel_list = _ravel_list(leaves)
   unravel_pytree = lambda flat: tree_unflatten(treedef, unravel_list(flat))
   return flat, unravel_pytree
 
-def _ravel_list(*lst):
-  return jnp.concatenate([jnp.ravel(elt) for elt in lst]) if lst else jnp.array([])
+def _ravel_list(lst):
+  from_dtypes = [dtypes.dtype(l) for l in lst]
```
jakevdp, line 51: Is it important to preserve weak type information in the outputs? If so, you could use `dtype._dtype_and_weaktype` for each entry, and use `to_dtype = jnp.result_type(*leaves)`

mattjj: I'm not sure. Actually you should tell me, since you are the weak type expert :)

My guess is it is not important.

mattjj: My guess is it's not important because AIUI weak types are primarily useful for handling Python scalar literals in user code in a reasonable way. But we don't have to think about the round-tripped output as a Python scalar literal.

jakevdp: It sounds reasonable to ignore weak types in unravelling. But for the ravelled dtype, it may be important. For example, should we expect `ravel_pytree(jnp.bfloat16(0.0), 1.0))` ravel to `bfloat16` or to `float32`?

mattjj: Good thinking! What do you recommend as the answer to that question?

mattjj: Actually, I'd really appreciate if you could recommend a specific change. Otherwise I'm happy with landing it as-is and not worrying about it too much :)

jakevdp: I'm not entirely familiar with where this is intended to be used, but it looks like weak types are ignored in the current implementation, so it seems reasonable to keep doing that.

mattjj: SG!

Thanks for the advice. Anything else blocking this, or LGTY?

jakevdp (approved):
Looks great!

PR #6138, opened 2021-03-19
Handle jax.numpy.dtype('bfloat16') properly.

adarob:


jakevdp:
Looks good, thanks!

jakevdp (approved):

mattjj:
Can we add a test so we don't accidentally regress this?

jakevdp:
@adarob are you interested in adding a test for this? Thanks!

adarob:
Sure

adarob:
Oddly, the tests passes without this change when I run it locally, but not in Colab. Any ideas about that @mattjj ?

```
assert jax.numpy.dtype('bfloat16') == jax.numpy.bfloat16
```

jakevdp:
That test fails on the master branch. I suspect you somehow installed JAX from your branch somewhere in your Python path, so the tests are hitting the modified code.

adarob:
test added

jakevdp:
Thanks!

jax/dtypes.py
```diff
@@ -325,6 +325,8 @@ def is_python_scalar(x):
     return type(x) in python_scalar_dtypes
 
 def dtype(x):
+  if isinstance(dtype, str) and dtype == "bfloat16":
```
jakevdp, line 328: I see the issue... this should be `isinstance(x, str) and x == "bfloat16":`

jakevdp: After you fix this, could you squash all the changes into a single commit? Thanks!

jax/dtypes.py
```diff
@@ -325,6 +325,8 @@ def is_python_scalar(x):
...
+    return bfloat16
```
jakevdp, line 329: Also, I just noticed that this should be `np.dtype(bfloat16)` (`result_type` returns a dtype object; `bfloat16` is a type not a dtype)

hawkinsp (commented):
FYI: you might also want to know about this PR:
https://github.com/tensorflow/tensorflow/commit/16671ca46eda906e822f61a3979f1f68d8bf603c

which added `np.dtype("bfloat16")` support for classic NumPy. That code is used by both JAX and TF, so it will show up with a jaxlib built from head.

jakevdp:
Shoot, looks like this got merged with failing tests.

I'll send a fix

jakevdp:
Fixed in #6176

PR #6112, opened 2021-03-18
unify configuration state handling

mattjj:
**Reviewer:** The main action is in config.py. The other changed files are downstream of those changes.

This change arose from working on adding a `jax_tpu_dot_precision` flag / context manager (see #6143).

We have a few configurable bits of global state. These bits of configurable state are managed by flags and sometimes dynamically by context managers, but neither the APIs nor implementations are uniform. Moreover the state is spread out through a few different files.

The bits of global state we have in mind are:

1. `core.skip_checks` (state in core.py, flag defined in config.py, not properly thread-local, has a context manager, does *not* affect `jit` dispatch since it's all about trace-time errors)
2. `core.debug_state.check_leaks` (state in core.py, flag defined in config.py, thread-local, has a context manager, does *not* affect `jit` dispatch since it's all about trace-time errors)
3. `jax_debug_nans` / `jax_debug_infs` (flag defined in xla.py, not properly thread-local, no context manager, affects `jit` dispatch in that it adds checks to every execution on the Python side) ([affects `jit` dispatch](https://github.com/google/jax/blob/5e88ed290359c301aa20d57b6c40521b43d4a7bf/jax/api.py#L361-L374))
4. `jax_log_compiles` (flag defined in xla.py, not properly thread-local, no context manager, does *not* affect `jit` dispatch since it's all about trace-time logging)
4. `jax_enable_x64` (work-in-progress, state in jax_jit.cc, context manager being developed in `jax.experimental`, thread-local, [affects `jit` dispatch](https://github.com/google/jax/blob/5e88ed290359c301aa20d57b6c40521b43d4a7bf/jax/api.py#L359) in that it's part of the compilation cache key and affects how input arguments are handled in the c++ code)
5. `jax_default_dot_precision` (work-in-progress, not present yet, affects `jit` dispatch analogously to `jax_enable_x64`)
6. `disable_jit` (state in jax_jit.cc, [affects `jit` dispatch](https://github.com/google/jax/blob/5e88ed290359c301aa20d57b6c40521b43d4a7bf/jax/api.py#L375) in that it's part of the compilation cache key)
7. `jax_numpy_rank_promotion` (flag defined in lax_numpy.py, not thread-local and no context manager, does *not* affect `jit` dispatch in that it's all about trace-time errors)

[...]

PR #6143, opened 2021-03-19
add jax_default_matmul_precision flag and context manager

### LINKED ISSUE #2161: Matrix multiplication precision API

mattjj:
~**Notice the "diffbase" on the flag-cleanup branch, i.e. #6112**.~ See the comment on #6112 for more context.

fixes #2161

This PR adds a configuration option `jax_default_matmul_precision` and a context manager `jax.default_matmul_precision(...)` to control the default precision of internal computations used in matrix multiplies and convolutions on float32 inputs for supported backends (currently just TPU but likely soon A100 GPUs as well).

For example, say we have a function `foo` which includes a `jnp.dot` call:

```python
def foo(...):
  ...
  z = jnp.dot(x, y)  # x and y are float32 arrays
  ...
```

We can ensure that `dot` is computed at the highest (or lowest) precision using any of these methods:
1. we can set the shell environment variable `JAX_DEFAULT_MATMUL_PRECISION=float32` (or `JAX_DEFAULT_MATMUL_PRECISION=bfloat16`);
2. if our main script parses flags with [`absl`](https://github.com/abseil/abseil-py), then we can use the command-line flag `--jax_default_matmul_precision=float32` (or `--jax_default_matmul_precision=bfloat16`);
3. at the top of our main script, we can put `jax.config.update('jax_default_matmul_precision', 'float32')` (or `jax.config.update('jax_default_matmul_precision', 'bfloat16')`);
4. when we call `foo`, we can use the `jax.default_matmul_precision` context manager:
```python
with jax.default_matmul_precision('float32'):   # or 'bfloat16' for lowest
  ... = foo(...)
```

This configuration option controls the _default_ precision in the sense that convolution operations like `lax.conv_general_dilated` and matrix multiply operations like `lax.dot` take an optional `precision` argument. This configuration option does not change the behaviors of such calls with explicit precision arguments; it only changes the behaviors of calls with no such argument provided.

This PR does **not** change the default default dot precision; that remains `'bfloat16'`. It only adds new ways to control the default dot precision. We might change the default in follow-up work.

In follow-up work we may add an analogous bit of enum state for controlling the default device. But I think we should land this part first.

TODO:
- [x] write PR message
- [x] add tests

cc @rohan-anil @sharadmv @shoyer @SiegeLordEx @jonbarron

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -506,8 +506,17 @@ def concatenate(operands: Sequence[Array], dimension: int) -> Array:
 Precision = xla_client.PrecisionConfig.Precision
 Precision.__str__ = lambda precision: precision.name
 PrecisionType = Any
-PrecisionLike = Union[None, PrecisionType, Tuple[PrecisionType, PrecisionType]]
-
+PrecisionLike = Union[None, str, PrecisionType, Tuple[str, str],
+                      Tuple[PrecisionType, PrecisionType]]
+_precision_strings = {
+    'highest':       Precision.HIGHEST,
+    'float32':       Precision.HIGHEST,
+    'bfloat16_3x':   Precision.HIGH,
+    'tensorfloat32': Precision.HIGH,
+    'bfloat16':      Precision.DEFAULT,
+    'fastest':       Precision.DEFAULT,
+    None:            Precision.DEFAULT,
+}
 
 class ConvDimensionNumbers(NamedTuple):
   """Describes batch, spatial, and feature dimensions of a convolution.
@@ -555,23 +564,25 @@ def conv_general_dilated(
     rhs_dilation: `None`, or a sequence of `n` integers, giving the
       dilation factor to apply in each spatial dimension of `rhs`. RHS dilation
       is also known as atrous convolution.
-    dimension_numbers: either `None`, a `ConvDimensionNumbers` object, or
-      a 3-tuple `(lhs_spec, rhs_spec, out_spec)`, where each element is a string
-      of length `n+2`.
+    dimension_numbers: either `None`, a ``ConvDimensionNumbers`` object, or
+      a 3-tuple ``(lhs_spec, rhs_spec, out_spec)``, where each element is a
+      string of length `n+2`.
     feature_group_count: integer, default 1. See XLA HLO docs.
     batch_group_count: integer, default 1. See XLA HLO docs.
     precision: Optional. Either ``None``, which means the default precision for
       the backend, a ``lax.Precision`` enum value (``Precision.DEFAULT``,
-      ``Precision.HIGH`` or ``Precision.HIGHEST``) or a tuple of two
-      ``lax.Precision`` enums indicating precision of ``lhs``` and ``rhs``.
+      ``Precision.HIGH`` or ``Precision.HIGHEST``), a string (e.g. 'highest' or
+      'fastest', see the ``jax.default_matmul_precision`` context manager), or a
+      tuple of two ``lax.Precision`` enums or strings indicating precision of
+      ``lhs`` and ``rhs``.
 
   Returns:
     An array containing the convolution result.
 
-  In the string case of `dimension_numbers`, each character identifies by
+  In the string case of ``dimension_numbers``, each character identifies by
   position:
 
-  - the batch dimensions in `lhs`, `rhs`, and the output with the character
+  - the batch dimensions in ``lhs``, ``rhs``, and the output with the character
     'N',
   - the feature dimensions in `lhs` and the output with the character 'C',
   - the input and output feature dimensions in rhs with the characters 'I'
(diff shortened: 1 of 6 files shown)


jax/_src/lax/lax.py
```diff
@@ -551,9 +557,10 @@ def conv_general_dilated(
     feature_group_count: integer, default 1. See XLA HLO docs.
     batch_group_count: integer, default 1. See XLA HLO docs.
     precision: Optional. Either ``None``, which means the default precision for
-      the backend, a ``lax.Precision`` enum value (``Precision.DEFAULT``,
-      ``Precision.HIGH`` or ``Precision.HIGHEST``) or a tuple of two
-      ``lax.Precision`` enums indicating precision of ``lhs``` and ``rhs``.
+      the backend, a string ('bfloat16', 'tensorfloat32', or 'float32'), a
+      ``lax.Precision`` enum value (``Precision.DEFAULT``, ``Precision.HIGH`` or
+      ``Precision.HIGHEST``) or a tuple of two ``lax.Precision`` enums
+      indicating precision of ``lhs``` and ``rhs``.
```
shoyer, line 563: Could we also support a tuple of two strings for this case? E.g., `precision=('bfloat16', 'float32')` would be nice than requiring a pair of enums.

jax/config.py
```diff
@@ -316,3 +372,32 @@ def __getattr__(self, name):
...
+          'and :func:`jax.lax.dot` docstrings. But it can be useful to control '
+          'the default behavior obtained when an operation is not given a '
+          'specific precision.\n\n'
+
+          'This option can be used to control the default precision '
+          'level for computations involved in matrix multiplication and '
+          'convolution on 32bit inputs. The levels roughly describe the '
+          "precision at which scalar products are computed. The 'bfloat16' "
+          "option is the fastest and least precise; 'float32' is similar to "
+          "full 'float32 precision; 'tensorfloat32' is intermediate.\n\n"))
```
shoyer, line 403:
```suggestion
          "full float32 precision; 'tensorfloat32' is intermediate.\n\n"))
```

tests/api_test.py
```diff
@@ -2393,6 +2393,58 @@ def f(_):
...
+
+    with jax.default_dot_precision("bfloat16"):
+      jnp.dot(x, x)  # doesn't crash
+      jaxpr = jax.make_jaxpr(jnp.dot)(x, x)
+    self.assertIn('precision=DEFAULT', str(jaxpr))
+
+    with jax.default_dot_precision("tensorfloat32"):
+      jnp.dot(x, x)  # doesn't crash
+      jaxpr = jax.make_jaxpr(jnp.dot)(x, x)
+    self.assertIn('precision=HIGH', str(jaxpr))
```
shoyer, line 2412: note that `precision=HIGHEST` would also pass this test.

tests/api_test.py
```diff
@@ -2393,6 +2393,58 @@ def f(_):
     expected = jnp.arange(1) + 1
     self.assertAllClose(ans, expected)
 
+  def test_dot_precision_context_manager(self):
```
shoyer, line 2396: A test or two specifically for `@` as matrix multiplication (perhaps via `operator.matmul`) might also be worthwhile.

shoyer (approved):

tests/api_test.py
mattjj: Hah! Good catch.

jax/_src/lax/lax.py
```diff
@@ -496,7 +496,13 @@ def concatenate(operands: Sequence[Array], dimension: int) -> Array:
 Precision = xla_client.PrecisionConfig.Precision
 Precision.__str__ = lambda precision: precision.name
 PrecisionType = Any
-PrecisionLike = Union[None, PrecisionType, Tuple[PrecisionType, PrecisionType]]
+PrecisionLike = Union[None, str, PrecisionType,
+                      Tuple[PrecisionType, PrecisionType]]
+_precision_strings = {
+    'bfloat16':      Precision.DEFAULT,
+    'tensorfloat32': Precision.HIGH,
```
hawkinsp, line 503: I'm wondering whether `tensorfloat32` is an good description. On A100 I believe CUDA `tensorfloat32` gives you  10 bits of mantissa on the input. And I'm not sure that `HIGH` in XLA's naming actually gives you `tensorfloat32` on GPU.

On TPU it means something different, namely a multipass algorithm on bfloat16 inputs. Is conflating the two is wise? I suspect we should just have two different names here.

shoyer: I guess the nice thing about strings instead of enums is that we can support multiple redundant names -- they don't have to be unique, and we can extend it over time.

So perhaps we could support all of `bfloat16`/`bfloat16_1x`, `bfloat16_3x` and `float32`/`bfloat16_6x` for now, and extend that list later when XLA adds true support for `tensorfloat32` on GPUs, in whatever form that takes.

shoyer:
"matmul_precision" might be a more descriptive name than "dot_precision", as it only effects matrix-matrix multiplication and convolutions, which are arguably a special case of matrix/matrix multiplication.

jax/_src/lax/lax.py
hawkinsp: I suggest you want a couple of generic options:
* fastest
* most precise

and after that point you want to be completely specific about which algorithm you mean.

shoyer: OK, so for the current state of things, how about:

- `precision='highest'` -> `lax.Precision.HIGHEST`
- `precision='float32'` -> `lax.Precision.HIGHEST`
- `precision='bfloat16_3x'` -> `lax.Precision.HIGH`
- `precision='bfloat16'` -> `lax.Precision.DEFAULT`
- `precision='fastest'` -> `lax.Precision.DEFAULT`

(We could add the `bfloat16_1x` and `bfloat16_6x` aliases as well, if desired)

mattjj: cc @jekbradbury 

Maybe we should discuss on chat for lower latency?

hawkinsp (approved):

mattjj:
Ping for robots...

Issue #3813, opened 2020-07-21
`regularized_incomplete_beta_p` incorrectly broadcasts when `vmap`-ed


sharadmv:
Repro:
```python
from functools import partial
from jax import lax
import jax.numpy as jnp
import jax

reg_inc_beta = partial(lax.regularized_incomplete_beta_p.bind, jnp.ones(1), 0.5)
print(reg_inc_beta(jnp.array(1.))) # works
print(jax.vmap(reg_inc_beta)(jnp.ones([4]))) # ERRORS
```

Traceback:
```
[1.]
/home/sharadmv/workspace/jax/jax/lax/lax.py:1758: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  shapes = onp.array([aval.shape for aval in avals if aval.shape])
Traceback (most recent call last):
  File "test.py", line 8, in <module>
    print(jax.vmap(reg_inc_beta)(jnp.ones([4]))) # ERRORS
  File "/home/sharadmv/workspace/jax/jax/api.py", line 770, in batched_fun
    lambda: flatten_axes(out_tree(), out_axes))
  File "/home/sharadmv/workspace/jax/jax/interpreters/batching.py", line 34, in batch
    return batched_fun.call_wrapped(*in_vals)
  File "/home/sharadmv/workspace/jax/jax/linear_util.py", line 150, in call_wrapped
    ans = self.f(*args, **dict(self.params, **kwargs))
  File "/home/sharadmv/workspace/jax/jax/core.py", line 214, in bind
    out_tracer = top_trace.process_primitive(self, tracers, kwargs)
  File "/home/sharadmv/workspace/jax/jax/interpreters/batching.py", line 134, in process_primitive
    val_out, dim_out = batched_primitive(vals_in, dims_in, **params)
  File "/home/sharadmv/workspace/jax/jax/interpreters/batching.py", line 249, in broadcast_batcher
    out = prim.bind(*args, **params)
  File "/home/sharadmv/workspace/jax/jax/core.py", line 211, in bind
    return self.impl(*args, **kwargs)
  File "/home/sharadmv/workspace/jax/jax/interpreters/xla.py", line 217, in apply_primitive
    compiled_fun = xla_primitive_callable(prim, *map(arg_spec, args), **params)
[...]

PR #6145, opened 2021-03-20
skip scalars when broadcasting for batch dimension agreement

### LINKED ISSUE #6096: betainc batching rule is not quite correct
This issue can be replicated using the following code

```python
import jax
import jax.numpy as jnp
from jax.scipy.special import betainc

def f(x):
    return betainc(jnp.ones(3), 1., x)

assert f(jnp.ones(3)).shape == (3,)
assert jax.vmap(f)(jnp.ones((2, 3))).shape == (2, 3)
# TypeError: regularized_incomplete_beta got arrays of different rank: (2, 3), (2,), (2, 3).
```

froystig:
fixes #6096

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2113,6 +2113,7 @@ def _naryop_weak_type_rule(name, *avals, **kwargs):
   return all(aval.weak_type for aval in avals)
 
 def naryop(result_dtype, accepted_dtypes, name, translation_rule=None):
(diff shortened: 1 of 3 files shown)


mattjj (approved):
Gooood catch! And +1 to reverting `naryop` back to binary, and just special-casing `regularized_incomplete_beta`.

sharadmv:
Does this also fix #3813?

PR #6182, opened 2021-03-23
Add a JVP rule for the general case of lax.reduce.

hawkinsp:


hawkinsp:
@mattjj and @apaszke ping?

jax/_src/lax/lax.py
```diff
@@ -4968,6 +4968,49 @@ def _reduction_computation(c, jaxpr, consts, init_values, singleton=True):
...
+    return tuple(squeeze(x, (axis,)) for x in xs)
+
+  return api.jvp(_reduce_tree, primals, tangents)
+
+def _reduce_jvp_rule(primals, tangents, *, computation, jaxpr,
+                     consts, dimensions):
+  primal_xs, init_values = split_list(primals, [len(primals) // 2])
+  tangent_xs, tangent_init = split_list(tangents, [len(tangents) // 2])
+  # TODO(phawkins): perhaps verify the init values do not have tangents.
+  # Admittedly, we don't verify that the primal init values are monoid units
```
mattjj, line 5009: Do we / should we document the monoid assumption somewhere, e.g. in the docstring?

jax/_src/lax/lax.py
```diff
@@ -4968,6 +4968,49 @@ def _reduction_computation(c, jaxpr, consts, init_values, singleton=True):
...
+      return [full(input_shape[non_axes], i) for i in init_values]
+    return tuple(squeeze(x, (axis,)) for x in xs)
+
+  return api.jvp(_reduce_tree, primals, tangents)
+
+def _reduce_jvp_rule(primals, tangents, *, computation, jaxpr,
+                     consts, dimensions):
+  primal_xs, init_values = split_list(primals, [len(primals) // 2])
+  tangent_xs, tangent_init = split_list(tangents, [len(tangents) // 2])
+  # TODO(phawkins): perhaps verify the init values do not have tangents.
```
mattjj, line 5008:
```suggestion
if any(type(t) is not ad_util.Zero for t in tangent_init): raise Exception
```

mattjj (approved):
LGTM, thanks! Great tests :)

jax/_src/lax/lax.py

It might be too strict a test if someone has non-symbolic zeros for some reason, but presumably they'll let us know.

jax/_src/lax/lax.py
hawkinsp: We already do that in the `lax.reduce` docstring (https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.reduce.html)

PR #6184, opened 2021-03-23
Enable sharded_jit_test and xmap SPMD tests for GPU.

jurahul:


jurahul:
@hawkinsp or @skye, please review.

tests/sharded_jit_test.py
```diff
@@ -247,15 +250,21 @@ def f(x):
...
     z = jnp.array([3.], dtype=np.float32)
 
-    result = f(x)
     assert len(jax.local_devices()) == len(y_shards)
     for device, y_shard in zip(jax.local_devices(), y_shards):
       device.transfer_to_infeed((y_shard, z))
+    # Transfer data to infeed first before executing the function. For some
+    # devices, the infeed is blocking, so transferring data to infeed before
+    # executing ensures that the execution does not deadlock waiting for
+    # the infeed data.
```
skye, line 259: I'm not sure I understand this comment -- what does the infeed block exactly? Also when you say "some devices", I'm guessing this is GPUs, so I would just say that so future readers know.

jurahul: I can't say I understand the threading thats happening within JAX, but essentially result = f(x), which is the execution of the compiled function blocks the main thread here so the code never reaches `tranfer_to_infeed` when running on a GPU. TPU seems to have a different behavior where the execution of the compiled code does not block the main thread, so `transfer_to_infeed` is reached when running on TPU. @hawkinsp, please let me know if this is a correct characterization. The other JAX test that exercises infeed, `infeed_test.py` also seems to use this pattern, where `transfer_to_infeed` happens before function execution (though there it doesn't look like the code was intentionally changed at some point to do this, it has been like this from the start).

I'll update the comment to be more specific here.

skye: Thanks, the updated comment is much clearer! (It said "the infeed is blocking before", now it says "the execution of the compiled function is blocking" which makes more sense :slightly_smiling_face:) This sounds like a larger problem with infeed on GPU but seems fine for this test.

skye (approved):
GPU sharded_jit wooo

tests/sharded_jit_test.py
jurahul: Thanks. I have filed a bug to investigate the GPU infeed issue further on XLA side of things.

PR #6189, opened 2021-03-23
Add transform name to UnexpectedTracerError.

LenaMartens:
If we thread through the name of the transform that kicked off a jaxpr trace, we can add this information to the `UnexpectedTracerError`. This is useful to track down where the actual tracer side-effect occurred. 
Happy to be convinced of a different approach which uses some global tracer state :)

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2070,6 +2070,17 @@ class APITest(jtu.JaxTestCase):
       # level, which is no longer live.
       jax.jit(jnp.add)(jnp.ones(()), count)
 
+  def test_escaped_tracer_transform_name(self):
+    with self.assertRaisesRegex(core.UnexpectedTracerError,
+                                "transformed by jit"):
+      jax.jit(self.helper_save_tracer)(1)
+      _ = self._saved_tracer+1
+
+    with self.assertRaisesRegex(core.UnexpectedTracerError,
+                                "transformed by pmap"):
+      jax.pmap(self.helper_save_tracer)(jnp.ones((1, 2)))
+      _ = self._saved_tracer+1
+
   def test_pmap_static_kwarg_error_message(self):
     # https://github.com/google/jax/issues/3007
     def f(a, b):
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -63,14 +63,18 @@ T = TypeVar('T')
 Array = Any
 
 @cache()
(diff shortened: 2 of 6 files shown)


tests/api_test.py
```diff
@@ -2076,6 +2076,17 @@ def f(_, __):
       # level, which is no longer live.
       jax.jit(jnp.add)(jnp.ones(()), count)
 
+  def test_escaped_tracer_transform_name(self):
+    with self.assertRaisesRegex(core.UnexpectedTracerError,
+                                "transformed by jit"):
+      jax.jit(self.helper_save_tracer)(1)
+      _ = self._saved_tracer+1
```
jakevdp, line 2083: Here and throughout, I think the line after the error will not be executed, because the block stops execution when it sees the (expected) error. To make sure the cleanup step runs regardless of whether the expected error is raised, we might consider a `try...finally` statement

jakevdp: Wait, strike that: I see now this final line is actually where the error is occurring! Carry on... 😬

jakevdp (approved):
Looks good, I think propagating this info is a good idea.

The CI has been having issues today: I'd suggest rebase on master and force-push to trigger a new run.

jakevdp (approved):

LenaMartens:
the `build without omnistaging (py 3.7 on ubuntu-latest, x64=0)` seems to be consistently timing out on this one, not sure if that is a cause for concern? I did try running the tests locally with omnistaging disabled and they were fine (and also that shouldn't actually make any difference anymore 🤔)

jakevdp:
Yes, something looks strange here... but if you've synced past #6196, `disable_omnistaging` should be a no-op, so the only difference I think is that the CI would be running Python 3.7

jakevdp:
I triggered a re-run of the jobs... we'll see if the error persists.

jakevdp:
... no, it's failing again, and unfortunately failing so slowly that the CI times out before reporting the actual error.

LenaMartens:
aha, thanks! I'll take a look on Monday. (btw is there a better way for me to rerun the checks other than re-pushing the commit?)

jakevdp:
I think I figured out what's happening: #6196 made `disable_omnistaging()` a no-op, but the *test* still thinks omnistaging is enabled, and so is failing on any omnistaging-dependent tests. I'm going to send another PR to fix that.

jakevdp:
Should be fixed after #6227

PR #6190, opened 2021-03-23
DOC: remove unimplemneted parameters from lax.numpy docstrings

jakevdp:
Our docstring wrappers already remove parameters from the docstring that do not appear in the function's argument list. However, some functions have valid parameters for which a `NotImplementedError` is raised; this adds a hook to `_wraps` to specify which such parameters should be skipped in the wrapped docstring.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1243,7 +1243,7 @@ def _gradient(a, varargs, axis):
   return a_grad
 
 
-@_wraps(np.gradient)
+@_wraps(np.gradient, skip_params=['edge_order'])
 def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
              edge_order=None):
   if edge_order is not None:
@@ -1785,7 +1785,7 @@ dsplit = _split_on_axis(np.dsplit, axis=2)
 def array_split(ary, indices_or_sections, axis: int = 0):
   return _split("array_split", ary, indices_or_sections, axis=axis)
 
-@_wraps(np.clip)
+@_wraps(np.clip, skip_params=['out'])
 def clip(a, a_min=None, a_max=None, out=None):
   _check_arraylike("clip", a)
   if out is not None:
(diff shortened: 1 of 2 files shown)


froystig (commented):
Why do we accept these parameters if only to raise `NotImplementedError`, rather than never accepting them to begin with? Should this PR simply remove them?

jakevdp:
That would be a possibility. The difference would be that passing the argument would result in a `TypeError` rather than `NotImplementedError`. Maybe that's why the arguments were initially supported?

The only potential breakage due to removing the arguments would be in the case code that explicitly passes the default value; e.g. `jnp.sum(x, out=None)` is currently fine, but would raise a `TypeError` if we removed the `out` argument completely.

I'd be open to trying to remove unimplemented arguments – what do you think?

froystig:
> I'd be open to trying to remove unimplemented arguments – what do you think?

What we have now is a mixed convention between `TypeError` and `NotImplementedError`. I think it's not clear to us which is better, which is perhaps why we ended up doing both over time. If we're taking the opportunity to switch to a consistent convention, then removal at least buys us simpler code, and maybe that's reason enough to stick to it.

Good point about potential breakage. Want to try removing unimplemented arguments, and if that causes a lot of errors downstream then return to the original change (filtering docstrings)?

jakevdp:
> Want to try removing unimplemented arguments, and if that causes a lot of errors downstream then return to the original change (filtering docstrings)?

Sure, I'll give it a try and run the tests

jakevdp:
So, interesting thing I discovered: numpy will delegate some operations to JAX array methods. So, for example, calling `np.sum(jnp.arange(4))` will end up calling the JAX `sum()` method, passing `out=None`. That's probably why we originally implementeded it the way it is.

I'm certain there are probably some situations in which this is not required, but with that in mind, I'd propose going forward with adjusting the docstrings.

froystig (approved):
Ah, good to know, and thanks for trying it out. Adjusting docstrings sounds good then. Is there a good place to summarize this finding about `numpy` calling into `jax.numpy`, maybe in the vicinity of `_wraps` or the new `skip_params` description?

jakevdp:
I added a comment in the `_reduction` implementation. For posterity, here's the evidence:
```python
import jax.numpy as jnp
import numpy as np

def print_on_call(f):
  def wrapped(*args, **kwargs):
    print(f"called with {args} {kwargs}")
    return f(*args, **kwargs)
  return wrapped

x = jnp.arange(4)                                                                                                                                                               
type(x).sum = print_on_call(type(x).sum)
np.sum(x)
# called with (DeviceArray([0, 1, 2, 3], dtype=int64),) {'axis': None, 'out': None}
# DeviceArray(6, dtype=int64)
```

PR #6203, opened 2021-03-24
Change enable_x64 to use the common boolean configuration mechanism.

hawkinsp:
To keep C++ and Python state in synchronization, adds new update_..._hook callbacks to the boolean configuration objects that are called when global or thread-local state changes.

I intend to use this mechanism for other flags that have C++ state as well, for example, `disable_jit` in subsequent PRs.

### DIFF
--- a/jax/config.py
+++ b/jax/config.py
@@ -19,6 +19,7 @@ import sys
 import threading
 
 from jax import lib
+from typing import Callable, Optional
 
 def bool_env(varname: str, default: bool) -> bool:
   """Read an environment variable and interpret it as a boolean.
@@ -58,6 +59,7 @@ class Config:
     # TODO(mattjj): delete these when only omnistaging is available
     self.omnistaging_enabled = bool_env('JAX_OMNISTAGING', True)
     self._omnistaging_disablers = []
+    self._update_hooks = {}
 
   def update(self, name, val):
     if self.use_absl:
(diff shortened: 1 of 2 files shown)


jax/config.py
```diff
@@ -90,8 +94,12 @@ def _read(self, name):
   def add_option(self, name, default, opt_type, meta_args, meta_kwargs):
     if name in self.values:
       raise Exception("Config option {} already defined".format(name))
+    update_hook = meta_kwargs.pop("update_hook", None)
```
mattjj, line 97: Maybe best not to mutate the passed-in `meta_kwargs` argument, unless I'm missing some context for why it's necessary?

hawkinsp: As far as I can see the purpose of `meta_kwargs` is to pass kwargs to absl, but we don't want to pass this argument to absl.

jakevdp: Maybe shallow copy before modifying?

jakevdp: Looks like this always comes via `**kwargs` in `config.DEFINE_*`, so I think mutation is fine.

jakevdp (approved):

jax/config.py
mattjj: A function mutating passed-in args is an error-prone pattern even if it's not _currently_ a bug given the callers in this file. (If this were clearly an internal-only method, like if it were named `_add_options`, then it'd be less error-prone, but still not great.)

How about just adding a `meta_kwargs = meta_kwargs.copy()` or similar?

hawkinsp: Done.

hawkinsp:
@mattjj LGTY?

jax/config.py
hawkinsp: Actually I had to move the parsing of this argument into the callers to make pytype happy (fine, whatever...) which makes this moot, since the `kwargs` are now ephemeral.

mattjj (approved):
LGTM, thanks, this is nice!

PR #6208, opened 2021-03-24
Allow integer/boolean convolutions

majnemer:


hawkinsp:
You'll need to push again to trigger a new CI build; the CI was failing when you sent this PR out.

hawkinsp (commented):
Looks like some cases aren't implemented on XLA:CPU. That probably means we need logic like:

```
if jtu.device_under_test() == "cpu" and np.issubdtype(dtype, np.integer):
  raise unittest.SkipTest("...")
```

majnemer:
> Looks like some cases aren't implemented on XLA:CPU. That probably means we need logic like:
> 
> ```
> if jtu.device_under_test() == "cpu" and np.issubdtype(dtype, np.integer):
>   raise unittest.SkipTest("...")
> ```

I fixed XLA:CPU with https://github.com/tensorflow/tensorflow/commit/89fc1ddd96f25aa9d8eb6a0104ca1fdb11b26753, XLA:GPU requires b/183565702 to get solved.

hawkinsp (approved):

PR #6210, opened 2021-03-24
Improve nested pmap error message.

skye:
It currently gives a misleading error message in the case of nested pmaps without this change.


mattjj (approved):
Woo tests for errors!

PR #6214, opened 2021-03-24
auto-label PRs as "pull ready" when lint_and_typecheck passes

froystig:


hawkinsp:
I suspect we do not intend to merge this.

PR #6215, opened 2021-03-24
Replace `host_id` with `process_index` terminology.

skye:
We're switching to the new terminology to avoid confusion in cases
where multiple jax processes are running on a single host, and each
process has a unique process_index/host_id.
    
This keeps aliases for the old `host_id` APIs for now, but these will
eventually be removed.


skye:
This is ready for review (forgot to update the PR title earlier).

PR #2719, opened 2020-04-15
Delete cotangent references on their last use

apaszke:
Current implementation of transposition may add a factor of 2x to
peak memory usage in real cases and _potentially an unbounded factor_
in pathological programs. The reason why this happens is because the
cotangents computed by the `backward_pass` are never evicted from the
environment until the whole transposition is complete. Other systems
(e.g. PyTorch) generally make use of refcounting or liveness analysis
to remove unnecessary references as soon as they are known to no
longer be needed.

A simple example that showcases this issue is this:
```python
def f(x):
  for i in range(1000):
    x = x * 4
  return x

x = np.ones(4)
vjp(f, x)[1](x)
```

Adding `print(len(ct_env))` at the end of `backward_pass` reveals that
the dictionary actually holds a thousand `DeviceArray`s, while both
forward and backward can be computed in constant memory. Of course this
is the pathological example I mentioned above, but one can easily see
that keeping the cotangents alive for the whole duration of differentiation
causes the memory cost to be approximately `fwd_coefs + all_fwd_intermediates`
instead of `fwd_memory + program_pathwidth` where:
* `fwd_coefs` is the amount of memory necessary to store all constant
  coefficients of the linearized function
* `all_fwd_intermediates` is the amount of memory necessary to
  store _all intermediates appearing in the forward program_.
* `program_pathwidth` is the maximum over amounts of memory necessary
  to store the live values over all transposed program locations

Note that usually we have that `all_fwd_intermediates > fwd_coefs >> program_pathwidth` (`>>` meaning that the RHS is usually significantly smaller).

PR #6216, opened 2021-03-24
simplify cotangent dropping in ad.backward_pass

mattjj:
Since variables in jaxprs are only assigned-to exactly once, when we transpose a jaxpr we end up reading each variable's cotangent value exactly once. That means we can pop the cotangent environment's reference to a cotangent value in read_cotangent.

[We did this in Autograd](https://github.com/HIPS/autograd/blob/01eacff7a4f12e6f7aebde7c4cb4c1c2633f217d/autograd/core.py#L20) but forgot about it in JAX until now! We noticed when working on autodidax #6049.

cross-ref #2719 for some context

### DIFF
--- a/jax/interpreters/ad.py
+++ b/jax/interpreters/ad.py
@@ -15,7 +15,7 @@
 
 import functools
 import itertools as it
-from typing import Any, Callable, Dict, Set, List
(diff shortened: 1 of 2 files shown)


froystig (approved):
!

PR #6217, opened 2021-03-24
add more options for jax_default_matmul_precision

mattjj:
This PR adds three aliases to the jax_default_matmul_precision enum options.

Currently (and in #6143) 'tensorfloat32' means the same thing as 'bfloat16_3x', but I think @hawkinsp may not like that. Now's our chance to make it right! WDYT?


jakevdp:
Closing as obsolete.

PR #6219, opened 2021-03-25
Fix lax.all_gather inside xmap

apaszke:
The batching rule didn't properly handle tupled axis names.


mattjj (approved):
Nice! (Sorry, I missed this one until now.)

PR #6238, opened 2021-03-26
Instantiate (symbolic) zero outputs of linear_transpose

j-towns:
IIRC symbolic zeros are not meant to be user-facing API. This change instantiates symbolic zero outputs of linear_transpose, in the same way that they would be instantiated by vjp/grad.


shoyer (approved):
Thanks!

Issue #3402, opened 2020-06-11
vjp of dtype promotion silently truncates complex numbers


shoyer:
As noticed in https://github.com/google/jax/pull/3398:
```
In [9]: from jax import vjp

In [10]: out, f_vjp = vjp(lambda x: 1j * x, 1.0)

In [11]: f_vjp(1 + 0j)  # wrong!
Out[11]: (DeviceArray(0., dtype=float32),)

In [12]: out, f_vjp = vjp(lambda x: 1j * x, 1.0 + 0j)

In [13]: f_vjp(1 + 0j)  # correct
Out[13]: (DeviceArray(0.+1.j, dtype=complex64),)
```

Instead, it looks like the cotangent always gets cast to the input dtype.

The result from the first calculation of `f_vjp` should either be the complex number `1j` (preferred) or at the very least should raise an error. Right now it calculates the wrong result.

mattjj:
This might be working as intended: the cast from f32 to c64 is like the R->C function x -> x + 0i, which if we write as an R->R^2 function has the Jacobian matrix [[1] [0]], the transpose of which is [1 0] (corresponding to the C->R function which drops the imaginary part of a c64 value and produces an f32).

Perhaps it looks surprising in this case because someone might think of 1.0 and 1.+0j as "the same", even though they have different types?

shoyer:
I'm not sure if it makes sense to define "linear" functions from real to complex numbers like R -> C, e.g., [Wikipedia](https://en.wikipedia.org/wiki/Linear_map#Definition_and_first_consequences) suggests that the input and output vector space of a linear maps need to over the same field. (In some cases, we might be interested in [linear functions over modules](https://en.wikipedia.org/wiki/Module_homomorphism) but either way it's the _same_ module in both input and output modules over the same ring.)

This model would suggest that every linear function transposed by VJP is either a function R^n -> R^m or a function C^n -> C^m. In which case, dtype promotion would be a convenient shortcut for evaluating these functions, rather than an intrinsic part of their nature.

Intuitively, I think of `f_vjp` as describing how to perform a calculation based on the (already linear) function `lambda x: 1j * x`, with the initial value 1.0 (or whatever) just a matter of book-keeping used in `out and required for partial evaluation / transposing. I do agree that this would be less surprising if Python functions included argument types as part of their definition.

shoyer:
Well, regardless of whether it is valid mathematically or not, certainly JAX works this way for exact the reason you describe:
```python
[...]

PR #3398, opened 2020-06-10
Add jax.linear_transpose

shoyer:
This is intended as a relatively low-level building block to facilitate using functional transposes in user code.

Questions:

- are we happy with this name? Other options: `transpose_fun`,  `transpose` (but could be confused with `jax.numpy.transpose`)
- add a formal definition of what a "linear function" is? (Suggestions would be appreciated!)

Co-authored-by: Matthew Johnson <mattjj@google.com>

PR #6239, opened 2021-03-26
Allow integer inputs/outputs of linear_transpose

j-towns:
I don't think disallowing integers in linear_transpose should be necessary (at risk of coming across as a nerd: I think they're a special case of the transpose of a module homomorphism 🤓, so mathematically well defined, see the definition of 'transpose' here: https://en.wikipedia.org/wiki/Module_homomorphism#Operations). 

The transpose rules haven't been tested on integers and there might be situations where we get silent failures ... If this change is deemed to be mergeable in principle, then I can implement such tests and add them to this PR.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2005,14 +2005,18 @@ def linear_transpose(fun: Callable, *primals) -> Callable:
   flat_fun, out_tree = flatten_fun_nokwargs(lu.wrap_init(fun), in_tree)
   in_avals = map(shaped_abstractify, primals_flat)
   in_dtypes = map(dtypes.dtype, in_avals)
-  if any(not np.issubdtype(dtype, np.inexact) for dtype in in_dtypes):
-    raise TypeError("linear_transpose only supports float and complex inputs, "
-                    f"but got {in_dtypes}")
 
   in_pvals = map(pe.PartialVal.unknown, in_avals)
   jaxpr, out_pvals, consts = pe.trace_to_jaxpr(flat_fun, in_pvals,
                                                instantiate=True)
   out_avals, _ = unzip2(out_pvals)
+  out_dtypes = map(dtypes.dtype, out_avals)
+  if not (all(dtypes.issubdtype(d, np.inexact) for d in in_dtypes + out_dtypes)
+          or all(dtypes.issubdtype(d, np.integer)
+                 for d in in_dtypes + out_dtypes)):
+    raise TypeError("linear_transpose only supports [float or complex] -> "
+                    "[float or complex], and integer -> integer functions, "
+                    f"but got {in_dtypes} -> {out_dtypes}.")
 
   def transposed_fun(out_cotangent):
     out_cotangents, out_tree2 = tree_flatten(out_cotangent)
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -1126,11 +1126,17 @@ class APITest(jtu.JaxTestCase):
     z, = transpose_fun(y)
     self.assertArraysEqual(2 * y, z, check_dtypes=True)
 
+  def test_linear_transpose_integer(self):
+    f = lambda x: 2 * x
+    transpose = api.linear_transpose(f, 1)
+    actual, = transpose(3)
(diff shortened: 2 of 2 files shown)


shoyer:
My main concern was about unexpected behavior for transposing integer -> float functions:
https://github.com/google/jax/pull/3398#issuecomment-690620368
https://github.com/google/jax/issues/3402#issuecomment-643830344

So I think I'm OK with adding support for integer -> integer functions, but that needs some additional checks.

j-towns:
Great, I've added a refined check that ensures the function is either [float or complex] -> [float or complex] or integer -> integer. By the way, it looks like transposing an float -> integer function was allowed before. Was that intentional? If so I need to re-allow that because it's banned by the new check.

In case you're interested, I'm using this for computing a data structure 'update' function from a view-like function (the matrix representation of a view is effectively an orthogonal projection matrix, so you can get a pseudo-inverse by transposing it - kind of using a sledgehammer to squash a fly, but it works: https://github.com/j-towns/crayjax/blob/3d0a84137cf607877d4f20a19fb5ec50f59b8238/crayjax/rans8.py#L132-L138). I've been doing this trick with Autograd for a while but needed to upgrade some code to run on GPU recently.

j-towns:
I wonder whether we can think of a more efficient way to do `view_update`, because the dense `tree_multimap(jnp.add, data, diff)` will be O(|data|), and I guess it should be possible to do the update in O(|item|).
```python
def view_update(data, view_fun):
    item, view_transpose = view_fun(data), linear_transpose(view_fun, data)
    item_copy = tree_map(_copy, item)
    def update(new_item):
        diff, = view_transpose(tree_multimap(jnp.subtract, new_item, item_copy))
        return tree_multimap(jnp.add, data, diff)  # <- this is inefficient
    return item, update
```

shoyer:
> Great, I've added a refined check that ensures the function is either [float or complex] -> [float or complex] or integer -> integer. By the way, it looks like transposing an float -> integer function was allowed before. Was that intentional? If so I need to re-allow that because it's banned by the new check.

Interesting... I think transposing float -> integer _should_ be OK. At least you won't see any funny truncation in the transposed computation (integer -> float).

j-towns:
Yeah, I see what you mean, also I guess it shouldn't technically be possible to define a linear float -> integer function (is it possible in JAX?). Shall we leave it disabled for now until/unless there's a use case for it?

jax/api.py
```diff
@@ -1936,14 +1936,17 @@ def linear_transpose(fun: Callable, *primals) -> Callable:
...
-    raise TypeError("linear_transpose only supports float and complex inputs, "
-                    f"but got {in_dtypes}")
 
   in_pvals = map(pe.PartialVal.unknown, in_avals)
   jaxpr, out_pvals, consts = pe.trace_to_jaxpr(flat_fun, in_pvals,
                                                instantiate=True)
   out_avals, _ = unzip2(out_pvals)
+  out_dtypes = map(dtypes.dtype, out_avals)
+  if not (all(np.issubdtype(d, np.inexact) for d in in_dtypes + out_dtypes)
+          or all(np.issubdtype(d, np.integer) for d in in_dtypes + out_dtypes)):
```
shoyer, line 1946: Better to use `jnp.issubdtype()` here, so it works on bfloat16.

(I see this was an issue in the original code, too, but might as well fix it now!)

shoyer (approved):
LGTM. I agree with can add support for float -> integer later if desired. It seems pretty weird for transpose to result in a non-transposable function, though.

j-towns:
The docs build failure looks to be unrelated to this change.

j-towns:
@shoyer I think this should be fine to merge.

shoyer:
This is indeed good to merge, but unfortunately it needs a rebase now after https://github.com/google/jax/pull/6422 moved the contents of `api.py`.

shoyer (approved):

j-towns:
Done.

j-towns:
Ping @shoyer.

shoyer:
Thanks Jamie!

PR #6250, opened 2021-03-28
Speling

knighton:
100% minor commit having to do with English spelling fixes. Not positive about your desired usage of dependence vs dependency.

PR #6251, opened 2021-03-28
Fix typo

saeta:


mattjj:
Thanks, Brennan!

There's also a set of typo fixes in #6250, which might include this one. It seems I write a lot of typos...

There's a way to sync the .md, .ipynb, and .py files which [Jake mentioned here](https://github.com/google/jax/pull/6250#issuecomment-808902123). (Personally I like to edit the .py directly, but luckily with this system Jake set up we can edit any one of the files and then push the changes to the others!)

saeta:
Think of these as an accent; it adds extra mystique and/or allure! (Looks like #6250 includes this fix, so closing this out.)

PR #6255, opened 2021-03-28
simplify public lax.convert_element_type api

mattjj:
Specifically:
1. don't expose weak_type in the public api, as it's jax-internal
2. don't make new_dtype optional, which could make bugs easier to write

This change keeps the public API simpler, and also makes convert_element_type match the ConvertElementType HLO. As an internal API we can call `lax._convert_element_type` just like before.


jakevdp (approved):
Looks good!

mattjj:
Thanks, Jake!

PR #6266, opened 2021-03-29
Add support for padding and cropping to fft

JoostvDoorn:
This adds padding and cropping of fft based on `s`.

### DIFF
--- a/jax/_src/numpy/fft.py
+++ b/jax/_src/numpy/fft.py
@@ -13,20 +13,24 @@
 # limitations under the License.
 
 
+import operator
 import numpy as np
 
 from jax import lax
 from jax.lib import xla_client
+from jax._src.util import safe_zip
 from .util import _wraps
 from . import lax_numpy as jnp
 from jax import ops as jaxops
 
 
 def _fft_core(func_name, fft_type, a, s, axes, norm):
-  # TODO(skye): implement padding/cropping based on 's'.
   full_name = "jax.numpy.fft." + func_name
+
   if s is not None:
-    raise NotImplementedError("%s only supports s=None, got %s" % (full_name, s))
+    s = tuple(map(operator.index, s))
+    if np.any(np.less(s, 0)):
+      raise ValueError("Shape should be non-negative.")
   if norm is not None:
     raise NotImplementedError("%s only supports norm=None, got %s" % (full_name, norm))
   if s is not None and axes is not None and len(s) != len(axes):
@@ -55,7 +59,18 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
     axes = tuple(range(a.ndim - len(axes), a.ndim))
     a = jnp.moveaxis(a, orig_axes, axes)
 
-  if s is None:
+  if s is not None:
+    a = jnp.asarray(a)
+    in_s = list(a.shape)
+    for axis, x in safe_zip(axes, s):
+      in_s[axis] = x
+    if fft_type == xla_client.FftType.IRFFT:
+      in_s[-1] = (in_s[-1] // 2 + 1)
+    # Cropping
+    a = a[tuple(map(slice, in_s))]
+    # Padding
+    a = jnp.pad(a, [(0, x-y) for x, y in zip(in_s, a.shape)])
+  else:
     if fft_type == xla_client.FftType.IRFFT:
       s = [a.shape[axis] for axis in axes[:-1]]
       if axes:
@@ -98,30 +113,31 @@ def _axis_check_1d(func_name, axis):
         "Got axis = %r." % (full_name, full_name, axis)
     )
 
-def _fft_core_1d(func_name, fft_type, a, s, axis, norm):
+def _fft_core_1d(func_name, fft_type, a, n, axis, norm):
   _axis_check_1d(func_name, axis)
   axes = None if axis is None else [axis]
+  s = None if n is None else [n]
   return _fft_core(func_name, fft_type, a, s, axes, norm)
 
 
 @_wraps(np.fft.fft)
 def fft(a, n=None, axis=-1, norm=None):
-  return _fft_core_1d('fft', xla_client.FftType.FFT, a, s=n, axis=axis,
+  return _fft_core_1d('fft', xla_client.FftType.FFT, a, n=n, axis=axis,
                       norm=norm)
 
 @_wraps(np.fft.ifft)
 def ifft(a, n=None, axis=-1, norm=None):
-  return _fft_core_1d('ifft', xla_client.FftType.IFFT, a, s=n, axis=axis,
+  return _fft_core_1d('ifft', xla_client.FftType.IFFT, a, n=n, axis=axis,
                       norm=norm)
 
 @_wraps(np.fft.rfft)
 def rfft(a, n=None, axis=-1, norm=None):
-  return _fft_core_1d('rfft', xla_client.FftType.RFFT, a, s=n, axis=axis,
+  return _fft_core_1d('rfft', xla_client.FftType.RFFT, a, n=n, axis=axis,
                       norm=norm)
 
 @_wraps(np.fft.irfft)
 def irfft(a, n=None, axis=-1, norm=None):
-  return _fft_core_1d('irfft', xla_client.FftType.IRFFT, a, s=n, axis=axis,
+  return _fft_core_1d('irfft', xla_client.FftType.IRFFT, a, n=n, axis=axis,
                       norm=norm)
 
 @_wraps(np.fft.hfft)
@@ -129,14 +145,14 @@ def hfft(a, n=None, axis=-1, norm=None):
   conj_a = jnp.conj(a)
   _axis_check_1d('hfft', axis)
   nn = (a.shape[axis] - 1) * 2 if n is None else n
-  return _fft_core_1d('hfft', xla_client.FftType.IRFFT, conj_a, s=n, axis=axis,
+  return _fft_core_1d('hfft', xla_client.FftType.IRFFT, conj_a, n=n, axis=axis,
                       norm=norm) * nn
 
 @_wraps(np.fft.ihfft)
 def ihfft(a, n=None, axis=-1, norm=None):
   _axis_check_1d('ihfft', axis)
   nn = a.shape[axis] if n is None else n
-  output = _fft_core_1d('ihfft', xla_client.FftType.RFFT, a, s=n, axis=axis,
+  output = _fft_core_1d('ihfft', xla_client.FftType.RFFT, a, n=n, axis=axis,
                       norm=norm)
   return jnp.conj(output) * (1 / nn)
 
--- a/tests/fft_test.py
+++ b/tests/fft_test.py
@@ -47,6 +47,11 @@ def _get_fftn_test_axes(shape):
     axes.append((-index,))
   return axes
 
+def _get_fftn_test_s(shape, axes):
+  s_list = [None]
+  if axes is not None:
+    s_list.extend(itertools.product(*[[shape[ax]+i for i in range(-shape[ax]+1, shape[ax]+1)] for ax in axes]))
+  return s_list
 
 def _get_fftn_func(module, inverse, real):
   if inverse:
@@ -59,8 +64,8 @@ def _irfft_with_zeroed_inputs(irfft_fun):
   # irfft isn't defined on the full domain of inputs, so in order to have a
   # well defined derivative on the whole domain of the function, we zero-out
   # the imaginary part of the first and possibly the last elements.
-  def wrapper(z, axes):
-    return irfft_fun(_zero_for_irfft(z, axes), axes=axes)
+  def wrapper(z, axes, s=None):
+    return irfft_fun(_zero_for_irfft(z, axes), axes=axes, s=s)
   return wrapper
 
 
@@ -91,15 +96,16 @@ class FftTest(jtu.JaxTestCase):
         func()
 
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_inverse={}_real={}_shape={}_axes={}".format(
-          inverse, real, jtu.format_shape_dtype_string(shape, dtype), axes),
-       "axes": axes, "shape": shape, "dtype": dtype, "inverse": inverse, "real": real}
+      {"testcase_name": "_inverse={}_real={}_shape={}_axes={}_s={}".format(
+          inverse, real, jtu.format_shape_dtype_string(shape, dtype), axes, s),
+       "axes": axes, "shape": shape, "dtype": dtype, "inverse": inverse, "real": real, "s": s}
       for inverse in [False, True]
       for real in [False, True]
       for dtype in (real_dtypes if real and not inverse else all_dtypes)
       for shape in [(10,), (10, 10), (9,), (2, 3, 4), (2, 3, 4, 5)]
-      for axes in _get_fftn_test_axes(shape)))
-  def testFftn(self, inverse, real, shape, dtype, axes):
+      for axes in _get_fftn_test_axes(shape)
+      for s in _get_fftn_test_s(shape, axes)))
+  def testFftn(self, inverse, real, shape, dtype, axes, s):
     rng = jtu.rand_default(self.rng())
     args_maker = lambda: (rng(shape, dtype),)
     jnp_op = _get_fftn_func(jnp.fft, inverse, real)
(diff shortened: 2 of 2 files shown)


jax/_src/numpy/fft.py
```diff
@@ -23,16 +23,29 @@
 
 
 def _fft_core(func_name, fft_type, a, s, axes, norm):
-  # TODO(skye): implement padding/cropping based on 's'.
   full_name = "jax.numpy.fft." + func_name
-  if s is not None:
-    raise NotImplementedError("%s only supports s=None, got %s" % (full_name, s))
+
+  if s is not None and not isinstance(s, list):
```
jakevdp, line 28: It looks like `s` is required to be a static sequence of integers here. We could improve validation by using something similar to [`_ensure_index_tuple`](https://github.com/google/jax/blob/5eb07c1b4acd099ff51c29b129950139dfdaf1d9/jax/api_util.py#L41-L46)

jax/_src/numpy/fft.py
```diff
@@ -23,16 +23,29 @@
...
+    s = [s]
   if norm is not None:
     raise NotImplementedError("%s only supports norm=None, got %s" % (full_name, norm))
   if s is not None and axes is not None and len(s) != len(axes):
     # Same error as numpy.
     raise ValueError("Shape and axes have different lengths.")
 
+  if s is not None:
+    in_s = [x for x in s]
```
jakevdp, line 37: nit: `in_s = list(s)` seems clearer

jax/_src/numpy/fft.py
```diff
@@ -23,16 +23,29 @@
...
+    if fft_type == xla_client.FftType.IRFFT:
+      in_s[-1] = (in_s[-1] // 2 + 1)
+
+    if any([x != y for x, y in zip(a.shape, in_s)]):
+      # Cropping
+      a = a[tuple([slice(0, min(*x)) for x in zip(a.shape, in_s)])]
```
jakevdp, line 43: Given the semantics of slicing, I think this is a bit cleaner:
```python
a = a[tuple(map(slice, in_s))]
```
It seems we should also somewhere check that all entries in `s` are positive to avoid strange outputs here.

Also, we should call `jnp.asarray` before this line, to prevent slicing from happening statically.

jax/_src/numpy/fft.py
```diff
@@ -23,16 +23,29 @@
...
+      # Padding
+      a = jnp.asarray(a)
+      zeros = jnp.zeros(in_s, dtype=a.dtype)
+      a = jaxops.index_update(zeros, tuple([slice(0, x) for x in a.shape]), a)
```
jakevdp, line 47: more succinct version of these three lines:
```
a = jnp.zeros(in_s, a.dtype).at[tuple(map(slice, a.shape))].set(a)
```

tests/fft_test.py
```diff
@@ -164,18 +164,19 @@ def testFftEmpty(self):
     self.assertArraysEqual(jnp.zeros((0,), jnp.complex64), out)
 
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_inverse={}_real={}_hermitian={}_shape={}_axis={}".format(
-          inverse, real, hermitian, jtu.format_shape_dtype_string(shape, dtype), axis),
+      {"testcase_name": "_inverse={}_real={}_hermitian={}_shape={}n={}_axis={}".format(
```
jakevdp, line 167: Please add an underscore before the `n={}`

jakevdp (changes requested):
Looks great, thanks! A few comments below and we can get this merged.

jax/_src/numpy/fft.py
shoyer: There's also `jnp.pad`. Functionally I would guess they are very similar, though.

tests/fft_test.py
```diff
@@ -164,18 +164,19 @@ def testFftEmpty(self):
...
        "axis": axis, "shape": shape, "dtype": dtype, "inverse": inverse, "real": real,
-       "hermitian": hermitian}
+       "hermitian": hermitian, "n": n}
       for inverse in [False, True]
       for real in [False, True]
       for hermitian in [False, True]
       for dtype in (real_dtypes if (real and not inverse) or (hermitian and inverse)
                                 else all_dtypes)
       for shape in [(10,)]
+      for n in [None, 1, 20]
```
shoyer, line 177: I would suggest includes some odd shapes in the tests, given that some of the padding depends on whether the input has odd or even size.

jax/_src/numpy/fft.py
shoyer, line 47: I am a little surprised that you didn't have to touch anything in `jax/_src/lax/fft.py` to make this work. Maybe we are just lucky, but it makes me wonder if maybe the test suite is not comprehensive enough.

I would suggest writing a few couple of additional explicit tests to verify that this works. In particular, it would be good to test gradients of `rfft`/`irfft` with an explicit size.

shoyer: It would also be nice to put this logic close to the `if s is None:` block below, just so all that handling is in one place.

hawkinsp: I'd probably recommend using `lax.pad` or `jnp.pad` here, given I'm not completely confident XLA knows how to optimize a indexed-update of zeros into a `pad`, and we certainly prefer a `pad`.

jakevdp:
Hi @JoostvDoorn - let us know when this is ready for another round of review. Thanks!

JoostvDoorn:
> Hi @JoostvDoorn - let us know when this is ready for another round of review. Thanks!

@jakevdp It's now ready for another round. I wanted to test the changes with cuda as well, but hadn't had the chance till today, all seems to be working.

jax/_src/numpy/fft.py
```diff
@@ -55,7 +58,20 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
+    for axis, x in zip(axes, s):
+      in_s[axis] = x
+    if fft_type == xla_client.FftType.IRFFT:
+      in_s[-1] = (in_s[-1] // 2 + 1)
+
+    if any([x != y for x, y in zip(in_s, a.shape)]):
+      # Cropping
+      a = a[tuple(map(slice, in_s))]
+      # Padding
+      a = lax.pad(a, lax._const(a, 0), [(0, x-y, 0) for x, y in zip(in_s, a.shape)])
```
shoyer, line 73: I would suggest using the higher level `jnp.pad` here:
```suggestion
      a = jnp.pad(a, [(0, x-y) for x, y in zip(in_s, a.shape)])
```

jax/_src/numpy/fft.py
```diff
@@ -55,7 +58,20 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
-  if s is None:
+  if s is not None:
+    a = jnp.asarray(a)
+    in_s = list(a.shape)
+    for axis, x in zip(axes, s):
+      in_s[axis] = x
+    if fft_type == xla_client.FftType.IRFFT:
+      in_s[-1] = (in_s[-1] // 2 + 1)
+
+    if any([x != y for x, y in zip(in_s, a.shape)]):
```
shoyer, line 69: Nit: this optimization probably isn't needed, as long as fft is called inside jit compiled code. I would remove it for now.

jax/_src/numpy/fft.py
```diff
@@ -55,7 +58,20 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
+      # Cropping
+      a = a[tuple(map(slice, in_s))]
+      # Padding
+      a = lax.pad(a, lax._const(a, 0), [(0, x-y, 0) for x, y in zip(in_s, a.shape)])
+  else:
     if fft_type == xla_client.FftType.IRFFT:
```
shoyer, line 75: I wonder if we could reuse this logic for building up `in_s`, to avoid duplicated code?

E.g.,
```python
    if fft_type == xla_client.FftType.IRFFT:
      in_s = [a.shape[axis] for axis in axes[:-1]]
      if axes:
        in_s += [max(0, 2 * (a.shape[axes[-1]] - 1))]
    else:
      in_s = [a.shape[axis] for axis in axes]

    if s is None:
      s = in_s
    else:
      a = a[tuple(map(slice, in_s))]
      a = jnp.pad(a, [(0, x-y) for x, y in zip(in_s, a.shape)])
```

JoostvDoorn: It would be nice if this could be simplified. However `in_s` is different because it represents the full input shape, while s and axes could present less axes. E.g. if `a.shape = [5, 3, 4, 5]`, `s = [10]`, and `axes = [1]` then we have `in_s=[5,10,4,5]`.

Since `lax.pad`/`jnp.pad` require all axes to be supplied, I think it's necessary for `in_s` to contain all even if no padding/cropping is needed.

jax/_src/numpy/fft.py
```diff
@@ -55,7 +58,20 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
     a = jnp.moveaxis(a, orig_axes, axes)
 
-  if s is None:
+  if s is not None:
+    a = jnp.asarray(a)
+    in_s = list(a.shape)
+    for axis, x in zip(axes, s):
+      in_s[axis] = x
+    if fft_type == xla_client.FftType.IRFFT:
+      in_s[-1] = (in_s[-1] // 2 + 1)
```
shoyer, line 67: Does this correctly handle the where the FFT is not along the last axis?

Perhaps this should be something like:
```suggestion
      in_s[axes[-1]] = (in_s[axes[-1]] // 2 + 1)
```

jax/_src/numpy/fft.py
shoyer: Ah, good point.

jax/_src/numpy/fft.py
JoostvDoorn: This ensures that the last axis of the fft is also the last axis of `a`:
https://github.com/google/jax/blob/d5154a43c41dc5f6a56f2acf2ca590b4fdd873ca/jax/_src/numpy/fft.py#L56-L59

jax/_src/numpy/fft.py
JoostvDoorn: I will removed this. I can see with 0 padding that this will be included in the XLA code, but I am not familiar enough with the internals to know if this matters or not.
```python
import jax
import numpy as np
import jax.numpy as jnp
def f(sig):
    return jnp.pad(sig, [(0, 0)])
print(jax.xla_computation(f)(jnp.asarray(np.random.randn(5))).as_hlo_text())
```

<details>
<summary>Output</summary>


```
HloModule xla_computation_f__1.64

jit__pad__1.6 {
  constant.11 = pred[] constant(false)
  parameter.9 = f32[5]{0} parameter(2)
  parameter.10 = s32[] parameter(3)
  broadcast.12 = s32[1,2]{1,0} broadcast(parameter.10), dimensions={}
  convert.13 = f32[1,2]{1,0} convert(broadcast.12)
  parameter.7 = s32[0]{0} parameter(0)
  convert.23 = s32[0]{0} convert(parameter.7)
  constant.14 = s32[] constant(0)
  constant.15 = s32[] constant(0)
  compare.16 = pred[] compare(constant.14, constant.15), direction=LT
  constant.17 = s32[] constant(0)
  constant.18 = s32[] constant(1)
  add.19 = s32[] add(constant.17, constant.18)
  constant.20 = s32[] constant(0)
  select.21 = s32[] select(compare.16, add.19, constant.20)
  broadcast.22 = s32[1]{0} broadcast(select.21), dimensions={}
  concatenate.24 = s32[1]{0} concatenate(convert.23, broadcast.22), dimensions={0}
  constant.25 = s32[] constant(0)
  constant.26 = s32[] constant(0)
  compare.27 = pred[] compare(constant.25, constant.26), direction=LT
  constant.28 = s32[] constant(0)
  constant.29 = s32[] constant(2)
  add.30 = s32[] add(constant.28, constant.29)
  constant.31 = s32[] constant(0)
  select.32 = s32[] select(compare.27, add.30, constant.31)
  broadcast.33 = s32[1]{0} broadcast(select.32), dimensions={}
  concatenate.34 = s32[2]{0} concatenate(concatenate.24, broadcast.33), dimensions={0}
  gather.35 = f32[] gather(convert.13, concatenate.34), offset_dims={}, collapsed_slice_dims={0,1}, start_index_map={0,1}, index_vector_dim=0, slice_sizes={1,1}
  pad.36 = f32[5]{0} pad(parameter.9, gather.35), padding=0_0
  parameter.8 = s32[0]{0} parameter(1)
  convert.46 = s32[0]{0} convert(parameter.8)
  constant.37 = s32[] constant(0)
  constant.38 = s32[] constant(0)
  compare.39 = pred[] compare(constant.37, constant.38), direction=LT
  constant.40 = s32[] constant(0)
  constant.41 = s32[] constant(1)
  add.42 = s32[] add(constant.40, constant.41)
  constant.43 = s32[] constant(0)
  select.44 = s32[] select(compare.39, add.42, constant.43)
  broadcast.45 = s32[1]{0} broadcast(select.44), dimensions={}
  concatenate.47 = s32[1]{0} concatenate(convert.46, broadcast.45), dimensions={0}
  constant.48 = s32[] constant(1)
  constant.49 = s32[] constant(0)
  compare.50 = pred[] compare(constant.48, constant.49), direction=LT
  constant.51 = s32[] constant(1)
  constant.52 = s32[] constant(2)
  add.53 = s32[] add(constant.51, constant.52)
  constant.54 = s32[] constant(1)
  select.55 = s32[] select(compare.50, add.53, constant.54)
  broadcast.56 = s32[1]{0} broadcast(select.55), dimensions={}
  concatenate.57 = s32[2]{0} concatenate(concatenate.47, broadcast.56), dimensions={0}
  gather.58 = f32[] gather(convert.13, concatenate.57), offset_dims={}, collapsed_slice_dims={0,1}, start_index_map={0,1}, index_vector_dim=0, slice_sizes={1,1}
  pad.59 = f32[5]{0} pad(pad.36, gather.58), padding=0_0
  ROOT tuple.60 = (f32[5]{0}) tuple(pad.59)
}

ENTRY xla_computation_f__1.64 {
  constant.4 = pred[] constant(false)
  constant.1 = s32[0]{0} constant({})
  constant.2 = s32[0]{0} constant({})
  parameter.3 = f32[5]{0} parameter(0)
  constant.5 = s32[] constant(0)
  call.61 = (f32[5]{0}) call(constant.1, constant.2, parameter.3, constant.5), to_apply=jit__pad__1.6
  get-tuple-element.62 = f32[5]{0} get-tuple-element(call.61), index=0
  ROOT tuple.63 = (f32[5]{0}) tuple(get-tuple-element.62)
}
```
</details>

jax/_src/numpy/fft.py
```diff
@@ -17,16 +17,19 @@
...
 from jax import ops as jaxops
 
 
 def _fft_core(func_name, fft_type, a, s, axes, norm):
-  # TODO(skye): implement padding/cropping based on 's'.
   full_name = "jax.numpy.fft." + func_name
+
   if s is not None:
-    raise NotImplementedError("%s only supports s=None, got %s" % (full_name, s))
+    s = _ensure_index_tuple(s)
```
jakevdp, line 30: For `fftn`, I think we need to be more careful: numpy's version raises an error if `s` is not a tuple.

jax/_src/numpy/fft.py
```diff
@@ -55,7 +58,18 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
     axes = tuple(range(a.ndim - len(axes), a.ndim))
     a = jnp.moveaxis(a, orig_axes, axes)
 
-  if s is None:
+  if s is not None:
+    a = jnp.asarray(a)
+    in_s = list(a.shape)
+    for axis, x in zip(axes, s):
+      in_s[axis] = x
```
jakevdp, line 65: Use `safe_zip` to raise an error if `axes` and `s` have different sizes. I *think* this is handled by the above logic, but it's not obvious from a quick read so `safe_zip` would be safer :grin:

jakevdp (commented):
Looking good! Just a couple more minor comments.

jax/_src/numpy/fft.py
JoostvDoorn: Numpy just calls `list(s)`, seems sensible here as well.

jax/_src/numpy/fft.py
JoostvDoorn: It will throw "Shape and axes have different lengths.", but no problem I can add this :grinning:

jax/_src/numpy/fft.py
```diff
@@ -17,16 +17,20 @@
...
 from jax import ops as jaxops
 
 
 def _fft_core(func_name, fft_type, a, s, axes, norm):
-  # TODO(skye): implement padding/cropping based on 's'.
   full_name = "jax.numpy.fft." + func_name
+
   if s is not None:
-    raise NotImplementedError("%s only supports s=None, got %s" % (full_name, s))
+    s = _ensure_index_tuple(list(s))
```
jakevdp, line 31: SInce we're disallowing the `isinstance(s, int)` case, I think `s = tuple(map(operator.index, s))` or `tuple(operator.index(n) for n in s)` would be clearer.

jakevdp (approved):

shoyer (approved):

tests/fft_test.py
```diff
@@ -47,6 +47,15 @@ def _get_fftn_test_axes(shape):
     axes.append((-index,))
   return axes
 
+def _get_fftn_test_s(shape, axes):
+  s_list = [None]
+  if axes is None or len(axes) == 0:
+    return s_list
+  s_list.append(tuple(shape[i] for i in axes))
+  s_list.append(tuple(shape[::-1][i]+1 for i in axes))
+  s_list.append(tuple(shape[i]+i for i in axes))
```
jakevdp, line 56: This logic results in repeated entries, which leads to an error when running a wider suite of tests:
```
raise DuplicateTestNameError(test_class_name, new_name, original_name)
absl.testing.parameterized.DuplicateTestNameError: Duplicate parameterized test name in FftTest: generated test name 'testFftn_inverse=True_real=True_shape=int32[2,3,4]_axes=(-3,)_s=(5,)' (generated from 'testFftn_inverse=True_real=True_shape=int32[2,3,4]_axes=(-3,)_s=(5,)') already exists. Consider using named_parameters() to give your tests unique names and/or renaming the conflicting test method.
```
We'll need to fix this before we can merge.

peterroelants: @JoostvDoorn You can have a look at https://github.com/google/jax/pull/6343/commits/9291632ea331a4b149ebd6e798505b4c50b64a67 where I tried adding the same as this PR.

The `testFftnWithShape` specifically tests fftn with explicit shapes.

JoostvDoorn: @peterroelants Thanks, I used your version of this function this is much simpler.

jakevdp:
I think the test failures are unrelated to this PR – if you rebase on master, they should go away.

PR #6272, opened 2021-03-29
Test: add test of random.fold_in for large integers

jakevdp:
Why? This is the [documented behavior](https://github.com/google/jax/blob/5eb07c1b4acd099ff51c29b129950139dfdaf1d9/jax/_src/random.py#L276) and is depended on by downstream libraries, but I inadvertently broke this in while working on #6269 and none of our tests caught the breakage.


mattjj (approved):
Awesome, thank you!

PR #3532, opened 2020-06-23
Static kwargs for jit

shoyer:
Fixes https://github.com/google/jax/issues/1936

This PR adds supports for static keyword arguments in `jit`:

- The new `static_argnames` allows for explicitly indicating keyword arguments that should be treated as static. This eliminates the need for separate helper functions to support keyword arguments (https://github.com/google/jax/issues/1936)
- If either `static_argnums` or `static_argnames` is not provided, the other argument is set by introspecting the wrapped function for positional-and-keyword arguments with `inspect.signature()`.
- It is now allowed to not include all positional argument referenced in `static_argnums`, because keyword arguments are identified automatically. It's still _possible_ to mess this up (e.g., by not using `functools.wraps` or explicitly setting static_argnums/static_argnames inconsistently with the function signature), but it should be much less likely to happen inadvertently.

Example usage:
```python
@partial(jax.jit, static_argnames=['precision'])
def dot(x, y, *, precision=None):
  ...
```

The overhead of `static_argnames` seems to be comparable to `static_argnums` (about 6 us each):
```
In [1]: import jax

In [2]: def f(x, *args, **kwargs):
   ...:     print(args, kwargs)
   ...:     return x
   ...:

In [3]: f_jit = jax.jit(f)

In [4]: f_args = jax.jit(f, static_argnums=(1,))

In [5]: f_kwargs = jax.jit(f, static_argnames=('y',))

In [6]: %timeit f_jit(0).block_until_ready()
() {}
82.2 µs ± 2.88 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: %timeit f_args(0, 'something static').block_until_ready()
('something static',) {}
88.3 µs ± 1.44 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: %timeit f_kwargs(0, y='something static').block_until_ready()
() {'y': 'something static'}
88 µs ± 2.19 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

TODOs:

- [x] ~~combine `static_argnums` and `static_argnames` into `static_args`? This
      can be done without any overhead in the jitted function by creating two
      separate variables.~~ I think it's better to keep the arguments separate. That allows for avoiding `inspect` if you prefer to be fully explicit. 
- [x] allow omitting static args when calling f_jitted? It is already OK to
      omit `static_kwargs`
- [x] validation of static_argnames
[...]

PR #6273, opened 2021-03-29
Progress towards static kwargs for jit

shoyer:
Based on #3532, but not exposed via `jax.jit` yet, while we work on the C++ implementation.

See #3532 for the description. To test it, use `jax.api._python_jit_with_static_argnames` in place of `jax.jit`.

### DIFF
--- a/jax/api_util.py
+++ b/jax/api_util.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import operator
-from typing import Any, Tuple, Union
+from typing import Any, Dict, Iterable, Tuple, Union
 
 import numpy as np
 
@@ -23,7 +23,7 @@ from .tree_util import (tree_flatten, tree_unflatten, tree_multimap,
                         tree_structure, treedef_children, treedef_is_leaf)
 from ._src.tree_util import _replace_nones
 from . import linear_util as lu
-from ._src.util import safe_map, WrapHashably, Hashable
+from ._src.util import safe_map, WrapHashably, WrapKwArgs, Hashable
 from .core import unit
 
 from ._src import traceback_util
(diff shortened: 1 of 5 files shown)


jax/api_util.py
```diff
@@ -124,6 +142,49 @@ def _argnums_partial(dyn_argnums, fixed_args, *dyn_args, **kwargs):
...
+  dyn_kwargs = {k: kwargs[k] for k in dyn_argnames}
+  return _argnames_partial(f, WrapKwArgs(fixed_kwargs)), dyn_kwargs
+
+
+def argnames_partial_except(f: lu.WrappedFun, static_argnames: Tuple[str, ...],
+                            kwargs: Dict[str, Any]):
+  if not static_argnames:
+    return f, kwargs
+  dyn_argnames = tuple(k for k in kwargs if k not in static_argnames)
+  dyn_kwargs = {k: kwargs[k] for k in dyn_argnames}
```
jakevdp, line 159: Combine these comprehensions and iterate over `kwargs.items()`?

jakevdp (approved):
LGTM

My only question here is in the behavior of `inspect.signature` - is the order of the keyword arguments deterministic in all supported Python versions? I'm worried it may rely on the dict implementation and therefore not preserve order before Python 3.7. Have you checked on that?

shoyer:
> My only question here is in the behavior of `inspect.signature` - is the order of the keyword arguments deterministic in all supported Python versions? I'm worried it may rely on the dict implementation and therefore not preserve order before Python 3.7. Have you checked on that?

Under `inspect.Signature.parameters`, the Python docs say:

> An ordered mapping of parameters’ names to the corresponding Parameter objects. Parameters appear in strict definition order, including keyword-only parameters.
>
> Changed in version 3.7: Python only explicitly guaranteed that it preserved the declaration order of keyword-only parameters as of version 3.7, although in practice this order had always been preserved in Python 3.

So it looks like we're good here: `parameters` has always preserved definition order, even for keyword-only arguments! And we only needed a guarantee of order "keyword or positional" arguments.

jax/api_util.py
shoyer: Done!

PR #6274, opened 2021-03-29
Re-exported tensorflow...xla_extension type stubs in jaxlib

superbobry:
The type stubs allow using precise types for XLA primitives instead
of aliasing them to Any.

This commit does not change any type annotations within JAX. That will
be done in a followup. I have manually verified that type stubs are
discoverable by mypy once the new jaxlib is installed by type "checking"

    from jaxlib import xla_extension as xe
    d: xe._Dtype

The type stubs were only exported in tensorflow/tensorflow@451befaa,
so this commit also bumps the TensorFlow version.

### DIFF
--- a/jaxlib/setup.py
+++ b/jaxlib/setup.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from setuptools import setup
-from glob import glob
 import os
 
 __version__ = None
@@ -25,19 +24,20 @@ cuda_version = os.environ.get("JAX_CUDA_VERSION")
 if cuda_version:
   __version__ += "+cuda" + cuda_version.replace(".", "")
 
-binary_libs = [os.path.basename(f) for f in glob('jaxlib/*.so*')]
-binary_libs += [os.path.basename(f) for f in glob('jaxlib/*.pyd*')]
-
 setup(
     name='jaxlib',
     version=__version__,
     description='XLA library for JAX',
     author='JAX team',
     author_email='jax-dev@google.com',
-    packages=['jaxlib'],
+    packages=['jaxlib', 'jaxlib.xla_extension-stubs'],
     python_requires='>=3.6',
     install_requires=['scipy', 'numpy>=1.16', 'absl-py', 'flatbuffers'],
     url='https://github.com/google/jax',
     license='Apache-2.0',
-    package_data={'jaxlib': binary_libs},
+    package_data={
+        'jaxlib': ['*.so', '*.pyd*', 'py.typed'],
+        'jaxlib.xla_extension-stubs': ['*.pyi'],
+    },
+    zip_safe=False,
 )
--- a/build/build_wheel.py
+++ b/build/build_wheel.py
@@ -78,6 +78,38 @@ def copy_file(src_file, dst_dir, dst_filename=None):
   else:
     _copy_normal(src_file, dst_dir, dst_filename=dst_filename)
 
+
+_XLA_EXTENSION_STUBS = [
+    "__init__.pyi",
+    "jax_jit.pyi",
+    "ops.pyi",
+    "outfeed_receiver.pyi",
+    "pmap_lib.pyi",
+    "profiler.pyi",
+    "pytree.pyi",
+]
+
+
+def patch_copy_xla_extension_stubs(dst_dir):
+  # This file is required by PEP-561. It marks jaxlib as package containing
+  # type stubs.
+  with open(os.path.join(dst_dir, "py.typed"), "w") as f:
+    pass
+  # The -stubs suffix is required by PEP-561.
+  xla_extension_dir = os.path.join(dst_dir, "xla_extension-stubs")
+  os.makedirs(xla_extension_dir)
+  for stub_name in _XLA_EXTENSION_STUBS:
+    with open(r.Rlocation(
+        "org_tensorflow/tensorflow/compiler/xla/python/xla_extension/" + stub_name)) as f:
+      src = f.read()
+    src = src.replace(
+        "from tensorflow.compiler.xla.python import xla_extension",
+        "from .. import xla_extension"
+    )
+    with open(os.path.join(xla_extension_dir, stub_name), "w") as f:
+      f.write(src)
+
+
 def patch_copy_xla_client_py(dst_dir):
   with open(r.Rlocation("org_tensorflow/tensorflow/compiler/xla/python/xla_client.py")) as f:
     src = f.read()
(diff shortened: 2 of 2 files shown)


jaxlib/setup.py
```diff
@@ -25,19 +24,25 @@
...
-    packages=['jaxlib'],
+    packages=['jaxlib', 'jaxlib.xla_extension-stubs'],
     python_requires='>=3.6',
     install_requires=['scipy', 'numpy>=1.16', 'absl-py', 'flatbuffers'],
     url='https://github.com/google/jax',
     license='Apache-2.0',
-    package_data={'jaxlib': binary_libs},
+    package_data={
+        'jaxlib': ['*.so', '*.pyd*', 'py.typed'],
+        'jaxlib.xla_extension-stubs': ["*.pyi"],
```
hawkinsp, line 45: I'm not familiar with the Python packaging requirements for stubs. This creates a new package that we'd have to distribute on Pypi, right?

Given we are providing first-party type stubs, is it possible to keep the `.pyi` files in the `jaxlib` package proper? The answer wasn't clear to me on a quick read of PEP 561.

jaxlib/setup.py
```diff
@@ -25,19 +24,25 @@
 if cuda_version:
   __version__ += "+cuda" + cuda_version.replace(".", "")
 
-binary_libs = [os.path.basename(f) for f in glob('jaxlib/*.so*')]
-binary_libs += [os.path.basename(f) for f in glob('jaxlib/*.pyd*')]
+# This file is required by PEP-561. It marks jaxlib as package containing
+# type stubs.
+with open('jaxlib/py.typed', 'w') as f:
```
hawkinsp, line 29: I'd be tempted to do this as part of the build_wheel script, just to keep all the package assembly code together.

hawkinsp (commented):
Thanks for the PR! A quick question.

jaxlib/setup.py
superbobry: My understanding is that `package_data` is just a way to define which files get included into the distribution (source archive, wheel, egg etc). See https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html. Running `build.py` at this PR produced a single `jaxlib` wheel with typestubs inside:

```
$ unzip -l dist/jaxlib-0.1.65-cp38-none-manylinux2010_x86_64.whl
Archive: dist/jaxlib-0.1.65-cp38-none-manylinux2010_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
      630  2021-03-29 20:32   jaxlib/__init__.py
   470208  2021-03-29 20:32   jaxlib/_pocketfft.so
   905728  2021-03-29 20:32   jaxlib/lapack.so
     5217  2021-03-29 20:32   jaxlib/pocketfft.py
    12485  2021-03-29 20:32   jaxlib/pocketfft_flatbuffers_py_generated.py
        0  2021-03-29 20:32   jaxlib/py.typed
     2359  2021-03-29 20:32   jaxlib/tpu_client.py
 13247744  2021-03-29 20:32   jaxlib/tpu_client_extension.so
      927  2021-03-29 20:32   jaxlib/version.py
    24362  2021-03-29 20:32   jaxlib/xla_client.py
145570784  2021-03-29 20:32   jaxlib/xla_extension.so
    11942  2021-03-29 20:32   jaxlib/xla_extension-stubs/__init__.pyi
     2357  2021-03-29 20:32   jaxlib/xla_extension-stubs/jax_jit.pyi
    13822  2021-03-29 20:32   jaxlib/xla_extension-stubs/ops.pyi
     1212  2021-03-29 20:32   jaxlib/xla_extension-stubs/outfeed_receiver.pyi
     2347  2021-03-29 20:32   jaxlib/xla_extension-stubs/pmap_lib.pyi
     1595  2021-03-29 20:32   jaxlib/xla_extension-stubs/profiler.pyi
     1848  2021-03-29 20:32   jaxlib/xla_extension-stubs/pytree.pyi
   240909  2021-03-29 20:32   jaxlib-0.1.65.dist-info/LICENSE.txt
      344  2021-03-29 20:32   jaxlib-0.1.65.dist-info/METADATA
      110  2021-03-29 20:32   jaxlib-0.1.65.dist-info/WHEEL
        7  2021-03-29 20:32   jaxlib-0.1.65.dist-info/top_level.txt
     1944  2021-03-29 20:33   jaxlib-0.1.65.dist-info/RECORD
---------                     ------- 
```

hawkinsp: Fair enough. I'm still not entirely understanding, say, how the period in the name becomes a directory in the package, but it looks like the output is sensible.

hawkinsp (changes requested):
Please fix the conflict in the WORKSPACE. I think you may just be able to drop that file from the change.

jaxlib/setup.py
superbobry: The convention is indeed a bit strange, but the keys in `package_data` are package (and subpackages) and the values are the corresponding inclusion patterns.

From https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html:

```python
from setuptools import setup, find_packages
setup(
    ...
    package_data={
        # If any package contains *.txt or *.rst files, include them:
        "": ["*.txt", "*.rst"],
        # And include any *.msg files found in the "hello" package, too:
        "hello": ["*.msg"],
    }
)
```

hawkinsp: Yeah, but how does `jaxlib.xla_extension-stubs` become `jaxlib/xla_extension-stubs` in the package? I don't see documentation for how this works.

(I also don't see documentation for the convention of how to package stubs for a binary extension, but this seems plausible to me if you say it works!)

jaxlib/setup.py
superbobry: Done.

hawkinsp (approved):

jaxlib/setup.py
superbobry: We manually copy `xla_extension/*.pyi` from TF to `jaxlib/xla_extension-stubs` in `patch_copy_xla_extension_stubs`. AFAICT `setuptools` does not change the directory structure, it merely selects a subtree of the build directory to be included into the distribution.

> (I also don't see documentation for the convention of how to package stubs for a binary extension, but this seems plausible to me if you say it works!)

There is indeed no discussion of native extensions in PEP-0561, but since type checkers cannot import them, I guess it's sufficient to just provide `.pyi` files alongside native extensions. That said, I've taken a slightly more conservative approach here by putting type stubs in a `-stubs` package which is intended for type [stub-only packages](https://www.python.org/dev/peps/pep-0561/#stub-only-packages ). This allows to avoid accidental clashes when (and if) `xla_extension` becomes a pure Python facade package for a bunch of native extensions.

PR #6275, opened 2021-03-29
remove disable_omnistaging mechanism

mattjj:


jax/config.py
```diff
@@ -59,11 +59,10 @@ def __init__(self):
     self.FLAGS = NameSpace(self.read, self.update)
     self.use_absl = False
     self._contextmanager_flags = set()
+    self._update_hooks = {}
 
     # TODO(mattjj): delete these when only omnistaging is available
     self.omnistaging_enabled = bool_env('JAX_OMNISTAGING', True)
```
jakevdp, line 65: Should we delete this flag as well?

jax/config.py
```diff
@@ -160,25 +159,14 @@ def parse_flags_with_absl(self):
...
           "To remove this warning, unset the JAX_OMNISTAGING environment variable.")
 
-  def register_omnistaging_disabler(self, disabler):
-    if self.omnistaging_enabled:
-      self._omnistaging_disablers.append(disabler)
-
   def enable_omnistaging(self):
-    if not self.omnistaging_enabled:
-      raise Exception("can't re-enable omnistaging after it's been disabled")
+    return  # TODO(mattjj): remove all callers
```
jakevdp, line 163: I think a warning here would be appropriate.

tests/api_test.py
```diff
@@ -2118,6 +2103,7 @@ def f(x):
     self.assertNotIn('constant.2 = f32[2]{0} constant({7, 14})', hlo_lines)
 
   def test_omnistaging_flag(self):
```
jakevdp, line 2105: delete?

tests/api_test.py
```diff
@@ -2806,8 +2750,7 @@ def scanned_f(x, _):
 
   def test_remat_jit_static_argnum(self):
```
jakevdp, line 2751: delete test?

jakevdp (approved):
LGTM - a couple minor comments. So many sweet, sweet red-highlighted lines...

tests/api_test.py
mattjj: I have a superstition against deleting tests, in part because they leave a breadcrumb trail. But in this case I can delete them if you prefer.

jakevdp: Keeping it around is fine

jax/config.py
mattjj: I hesitate to lump that in here because we might have to change downstream (e.g. google-internal) callers that rely on the flag existing, including perhaps its value being set by this environment variable and then read by their own code. I'd rather leave that for follow-up work. (If we were somehow 100% sure we could delete this without having to make downstream changes as a result, then I'd like to as well!)

jakevdp: OK, makes sense!

mattjj: To be clear, I agree with your suggestion, just that I might sequence it in an immediate follow-up PR!

tests/api_test.py
mattjj: I will delete this one!

mattjj:
Thanks for the thorough review of all these red lines!

mattjj:
@jakevdp should I make disable_omnistaging an error in this PR? I think there's no reason not to, unless I'm missing something.

tests/api_test.py
mattjj: You're right though: in this case it seems not so useful. I deleted it.

jax/config.py
mattjj: I'm going to attempt the change you suggest in an immediate follow-up PR. Specifically, I'm thinking something like:
1. get this PR in
2. make a pypi release
3. make disable_omnistaging an error, make enable_omnistaging at least a warning, possibly also removing the flag
4. make another pypi release

PR #6229, opened 2021-03-25
pre-release omnistaging cleanup

jakevdp:
Main changes:

- make `disable_omnistaging()` raise a warning
- add note in changelog
- modify docs to reflect this

PR #6276, opened 2021-03-29
disable_omnistaging error, enable_omnistaging warn

mattjj:


jax/config.py
```diff
@@ -339,7 +332,7 @@ def __setattr__(self, name, val):
 flags.DEFINE_bool(
     'jax_omnistaging',
     bool_env('JAX_OMNISTAGING', True),
-    help='Enable staging based on dynamic context rather than data dependence.'
+    help='Deprecated. The flag does nothing!',
```
jakevdp, line 335: "Will result in an error if set to False".

jakevdp (approved):

jax/config.py
mattjj: Oooh good thinking!

Will it though? I can't figure out how `disable_omnistaging` is called anymore...

mattjj: I think [this line](https://github.com/google/jax/pull/6229/files#diff-c429b93641dd4c7499c46f6e777cbe8982f5785545abac46aed8f8ce22e31e3cL157) removed in #6229 might mean it doesn't get called anymore.

mattjj: Nevermind, you're totally right, because of the second place the error is raised!

PR #6287, opened 2021-03-30
Reduce verbosity of jax-101 outline and fix pytree tutorial headers

skye:


jakevdp:
Looking at the [current version](https://jax.readthedocs.io/en/latest/jax-101/index.html) it looks like another thing that needs to be changes is the fact that the `pytrees` notebook uses top-level headings, so its subsections are treated as tutorial sections.

jakevdp (approved):
Looks great!

Issue #5785, opened 2021-02-19
jax.random.randint cannot generate values equal to DTYPE_MAX


altche:
According to randint docstring: `Sample uniform random values in [minval, maxval) with given shape/dtype.`

However because of the conversion here https://github.com/google/jax/blob/5bbb449ae5849c508194c8eb5b10c101f1fa22ae/jax/_src/random.py#L439 passing `maxval={dtype max value + 1}` will generate an all-zero array.
The behavior in numpy is different (and correct):
```
jax.random.randint(jax.random.PRNGKey(0), shape=(10,), minval=0, maxval=256, dtype=jnp.uint8)
>> DeviceArray([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=uint8)

np.random.randint(size=(10,), low=0, high=256, dtype=np.uint8)
>> array([ 87, 136, 195, 254,  95,  94, 113,   7,  66,  13], dtype=uint8)
```

A workaround is to not pass the dtype to randint (or use a dtype with a higher maximum value), and cast the output:
```
jax.random.randint(jax.random.PRNGKey(0), shape=(10,), minval=0, maxval=256).astype(jnp.uint8)
>> DeviceArray([178, 233,  99, 238,  39, 149,  46, 139,  68, 142], dtype=uint8)
```

jakevdp:
Thinking about this a bit... this is really tricky for JAX to handle in a general way. Consider the case of `max = 1 << 64`, which is one greater than the maximum uint64. Numpy handles it as expected:
```python
>>> np.random.randint(0, 1<<64, dtype='uint64')
17392714807819977315
```
But if you increase this bound by one, numpy returns an error:
```
>>> np.random.randint(0, 1<<64 + 1, dtype='uint64')                                                                                                                         
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-30-e02298560197> in <module>
----> 1 np.random.randint(0, 1<<64 + 1, dtype='uint64')

mtrand.pyx in numpy.random.mtrand.RandomState.randint()

_bounded_integers.pyx in numpy.random._bounded_integers._rand_uint64()

ValueError: high is out of bounds for uint64
```
[...]

PR #6293, opened 2021-03-30
random.randint: clip rather than wrap out-of-bounds min/max

jakevdp:
Related to #5785

Currently `random.randint` can fail in unexpected ways when the minimum/maximum value are out of the representable range of the specified dtype. One example:
```python
>>> from jax import random
>>> random.randint(random.PRNGKey(0), (10,), 0, 256, 'uint8')
DeviceArray([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=uint8)
```
This is because when `256` is converted to `uint8`, it wraps to zero.

With this PR, the result is this:
```python
>>> from jax import random
>>> random.randint(random.PRNGKey(0), (10,), 0, 256, 'uint8')
DeviceArray([112, 140,  86, 223,   7, 224, 109,  28,  21, 192], dtype=uint8)
```
The value 255 will never be generated (becuase `max` is exclusive and clipped to 255), but this is a step in the right direction, and will avoid the kinds of catastrophic failures seen in the above example.

### DIFF
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -26,6 +26,7 @@ import scipy.stats
 
 from jax import api
 from jax import core
+from jax import dtypes
 from jax import grad
 from jax import lax
 from jax import numpy as jnp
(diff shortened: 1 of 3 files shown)


tests/random_test.py
```diff
@@ -191,6 +192,8 @@ def testRngRandint(self, dtype):
     uncompiled_samples = rand(key)
     compiled_samples = crand(key)
 
+    print(uncompiled_samples)
+    print(compiled_samples)
```
zhangqiaorjc, line 196: Remove the debug prints?

zhangqiaorjc (approved):
Thanks Jake!

PR #6297, opened 2021-03-31
remove python jit path

mattjj:
cc @hawkinsp @jblespiau


jakevdp:
Note that #6273 is proposing new JIT logic only on the Python side for now. I'm not sure how these two PRs will interact.

mattjj:
Thanks for the heads-up!

Actually this PR doesn't present very much savings; less than 100 lines. Maybe we should keep the Python path around for prototyping, with #6273 as just one example.

Based on that thinking I'm leaning towards closing this PR. WDYT, @jakevdp and @shoyer ?

PR #6304, opened 2021-03-31
Draft: Add PRNGKey64 function

jakevdp:
Draft of a PRNGKey64 function, which would require key inputs to be 64-bit


hawkinsp:
Stale?

PR #6305, opened 2021-03-31
Adjust JetTest.test_scatter_add tolerance for TPU

skye:


skye:
Not sure if this is the right fix, should the precision be adjusted somewhere instead?

mattjj (approved):
This looks great, thank you.

Issue #6288, opened 2021-03-30
Case where jax.linear_transpose fails, but jax.vjp works 


lukepfister:
I'm running into an error when I try to use `jax.linear_transpose` on a function that involves `jax.ops.index_update`.  I get an error stating the transpose rule (for reverse-mode differentiation) is not implemented.  However, I don't get this error if I use `jax.vjp` to compute the adjoint.  

Using:
- jax  0.2.11
- jaxlib 0.1.64

I'm happy to contribute a fix, but could use some guidance on where to look.

```python
import numpy as np
import jax

def mosaic(x):
    y = jnp.zeros(x.shape)
    y = jax.ops.index_update(y, jax.ops.index[1::2], x[1::2])
    return y

x = jax.device_put(np.r_[:10].astype(np.float32))
fx = mosaic(x)

vjp_fun = jax.vjp(mosaic, x)[1]
vjp_fun(fx)  # works fine!

trans_fun = jax.linear_transpose(mosaic, x)
trans_fun(fx)  # Raises exception:

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/interpreters/ad.py in get_primitive_transpose(p)
    253   try:
--> 254     return primitive_transposes[p]
    255   except KeyError as err:

KeyError: scatter

The above exception was the direct cause of the following exception:

NotImplementedError                       Traceback (most recent call last)
~/scico/vjp_test.py in <module>
     18 
     19     trans_fun = jax.linear_transpose(mosaic, x)
---> 20     trans_fun(fx)

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/api.py in transposed_fun(out_cotangent)
   1965                       f"expected {out_avals} but got {out_cotangents}")
   1966     dummies = [ad.UndefinedPrimal(a) for a in in_avals]
-> 1967     in_cotangents = ad.backward_pass(jaxpr, consts, dummies, out_cotangents)
   1968     return tree_unflatten(in_tree, in_cotangents)
   1969 

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/interpreters/ad.py in backward_pass(jaxpr, consts, primals_in, cotangents_in)
    225             params, call_jaxpr, invals, cts_in, cts_in_avals)
    226       else:
--> 227         cts_out = get_primitive_transpose(eqn.primitive)(cts_in, *invals,
    228                                                          **eqn.params)
[...]

PR #6306, opened 2021-03-31
WIP: scatter via scatter_add

shoyer:
xref https://github.com/google/jax/issues/6288

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -4831,85 +4831,33 @@ batching.primitive_batchers[scatter_max_p] = (
   partial(_scatter_batching_rule, scatter_max))
 ad.primitive_jvps[scatter_max_p] = partial(_scatter_extremal_jvp, scatter_max_p)
 
-def _scatter_jvp(primals, tangents, *, update_jaxpr, update_consts,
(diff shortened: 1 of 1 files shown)


jax/_src/lax/lax.py
```diff
@@ -4831,85 +4831,33 @@ def _scatter_extremal_jvp(scatter_op, primals, tangents, update_jaxpr,
...
-  # If there are overlapping indices in the scatter, it is unspecified which
-  # update "wins". So we use the following perhaps surprising scheme:
-  # a) attach a positive ID to each update in updates, and perform the scatter
-  #    on the IDs
-  # b) perform the inverse gather on the scattered IDs (similar to
-  #    _scatter_add_transpose).
-  # c) use the gathered IDs to mask the primal and tangent values.
-  # d) perform a scatter-add on the masked primal and tangent values. A benefit
-  #    of using scatter-add here is that we don't need a `scatter` transpose
-  #    rule.
```
shoyer, line 4859: We probably shouldn't merge this change, because it breaks this guarantee.

hawkinsp:
Closing this PR, since, as Stephan suggests, it would break some of the semantics of `scatter`.

PR #6314, opened 2021-04-01
Lower lu_pivots_to_permutation using a custom CUDA kernel.

### LINKED ISSUE #5880: LU decomposition runtime is dominated by lu_pivots_to_permutation on GPU
The actual decomposition in handled by cusolver. However, the pivots returned by `getrf` are transformed into a permutation matrix via `lu_pivots_to_permutation`. For an `(n, n)` matrix this transformation is implemented as a loop with `n` iterations. Even though the amount of data involved is small (`int32` vector of length `n`), this loop performs poorly and its runtime by far exceeds that of `getrf`.

Trace for a 75x75 doube-precision matrix on a V100.
image

Batching does not improve the situation much. Trace for a batch of 64 double-precision matrices with shape (75, 75):
image

ahoenselaar:
Fixes google/jax#5880


jaxlib/cuda_lu_pivot.py
```diff
@@ -0,0 +1,68 @@
...
+  from . import cuda_lu_pivot_kernels
+  for _name, _value in cuda_lu_pivot_kernels.registrations().items():
+    xla_client.register_custom_call_target(_name, _value, platform="CUDA")
+except ImportError:
+  pass
+
+_prod = lambda xs: functools.reduce(operator.mul, xs, 1)
+
+
+# TODO(phawkins): remove after we no longer need to support old jax releases.
```
hawkinsp, line 32: You don't need this backward compatibility code. This function is always the identity in sufficiently new `jax` releases.

jaxlib/cuda_lu_pivot_kernels.cu.cc
```diff
@@ -0,0 +1,95 @@
...
+                                   const std::int32_t num_permutation_rows) {
+  for (int i = 0; i < num_permutation_rows; ++i) {
+    permutation_out[i] = i;
+  }
+
+  // Compute the permutation from a sequence of transpositions encoded in the
+  // pivot array by applying the transpositions in order on the identity
+  // permutation.
+  for (int i = 0; i < num_pivots; ++i) {
+    if (pivots[i] < 0) {
```
hawkinsp, line 39: I'd be tempted to skip out of bounds pivots on the high end also. This should never happen, but it's probably better to avoid corrupting GPU memory in that case.

jax/_src/lax/linalg.py
```diff
@@ -888,7 +889,9 @@ def _lu_batching_rule(batched_args, batch_dims):
   x = batching.moveaxis(x, bd, 0)
   return lu_p.bind(x), (0, 0, 0)
 
-def _lu_cpu_gpu_translation_rule(getrf_impl, c, operand):
+
+def _lu_cpu_gpu_translation_rule(getrf_impl, pivots_to_permutation_impl, c,
```
hawkinsp, line 893: This is fine, but I think we can do even better. I suggest we make `lu_pivots_to_permutation` into a primitive, which allows it in turn to have per-backend lowering rules. You can then always use the `lower_fun` implementation in this translation rule.

The reason I suggest this is that some folks are calling `lu_pivots_to_permutation` explicitly:
https://cs.opensource.google/search?q=lu_pivots_to_permutation&sq=

and right now they would still get the slow implementation on GPU. Indeed the `scipy` API almost forces that upon us.

Happily we would not need, say, autodiff rules for such a primitive, just abstract eval, shape, and batching rules.

jaxlib/cuda_lu_pivot.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2021 Google LLC
```
hawkinsp, line 1: This is fine, but I have to wonder if this is the last hand-written linear algebra CUDA kernel we will have to add... I might have been tempted to name the module something like `cuda_linalg.py`, but we can always do that if and when we have a second kernel.

hawkinsp (changes requested):

hawkinsp:
I should add: this is awesome! Thanks for the PR!

PR #6317, opened 2021-04-01
Correct jax.random.poisson docstring

NeilGirdhar:


jakevdp:
Thanks for looking into this! I think there are contexts in which referring to lambda by the name "rate" is more typical than referring to it by the name "scale", so it's not clear to me that this change should be made. For example, the Wikipedia article on the [Poisson distribution](https://en.wikipedia.org/wiki/Poisson_distribution) calls this parameter the "expected rate of occurrances".

NeilGirdhar:
@jakevdp Yeah, I actually thought of that when I suggested the change.  You are right that the rate of Poisson process (lambda) is modeled by a Poisson distribution whose mean is lambda.  However, the terms [_scale parameter_](https://en.wikipedia.org/wiki/Scale_parameter) and [_rate parameter_](https://en.wikipedia.org/wiki/Scale_parameter#Rate_parameter) and I don't think they should change meaning in this context.

On the one hand, the Wikipedia article for the Poisson distribution does call lambda, the "rate".  On the other hand, a Google scholar search for "Poisson distribution" and "scale parameter" turns up more hits than "Poisson distribution" and "rate paramter".

Up to you, I don't mind either way.  I just found it confusing.

NeilGirdhar:
Meh, I should have read more carefully. :)

PR #6321, opened 2021-04-01
Py38 docs fix

jakevdp:
Fixes #6252, and updates our readthedocs build to use Python 3.8

To be merged after #6318

PR #6322, opened 2021-04-01
DOC: clean up JAX 101 notebooks and remove WIP warning

jakevdp:
Mostly cleaning up formatting & metadata, fixing some links, truncating tracebacks, rearranging imports... also removed mentions of this being a work in progress.

I think this will have to wait for #6321 to avoid doc build warnings.


skye (approved):
LGTM modulo doc build warnings

PR #6330, opened 2021-04-02
Run doctest on all source files except jax2tf

jakevdp:
We had previously not been testing most docstrings; turns out a few of them were stale and/or contained typos.

The documentation test CI job now tests everything in jax except `jax/experimental/jax2tf` (this one has a number of errors that don't seem trivial to fix).


froystig (approved):
good catch!

PR #6336, opened 2021-04-03
Make safe_zip return iterator to match Python 3 builtin zip

lgeiger:
This PR changes `utils.safe_zip` to return an `Iterator` instead of a list to match the signature of the builtin [Python 3 `zip`](https://docs.python.org/3/library/functions.html#zip) function.
Since `utils.safe_zip` is mostly iterated over only once or passed directly to other builtin functions like `enumerate` that support iterators, this PR removes the need to allocate temporary lists during these iterations.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -3498,8 +3498,8 @@ def _pad_transpose(t, operand, padding_value, *, padding_config):
     total = lambda x: _reduce_sum(x, list(range(t.ndim)))
 
     def t_op():
-      unpad_config = safe_zip(np.negative(lo), np.negative(hi),
-                              np.zeros_like(interior))
+      unpad_config = tuple(safe_zip(np.negative(lo), np.negative(hi),
+                                    np.zeros_like(interior)))
       unpadded = pad(t, np.array(0., t.dtype), unpad_config)
       return slice(unpadded, np.zeros_like(lo), unpadded.shape, np.add(interior, 1))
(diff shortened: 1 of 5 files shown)


jekbradbury:
I think it's intentional that safe_zip/safe_map return lists; it makes things like print-debugging easier and avoids the possible bugs that can come from iterating multiple times (they're "safe" both from zip-length-mismatches and from that). Are there any benchmarks that this speeds up for you?

lgeiger:
Thanks for the fast review and sorry for the late reply.

> I think it's intentional that safe_zip/safe_map return lists; it makes things like print-debugging easier and avoids the possible bugs that can come from iterating multiple times (they're "safe" both from zip-length-mismatches and from that). Are there any benchmarks that this speeds up for you?

I don't expect any measurable performance benefits from this change.

The main motivation was here was only consistency with the builtin `zip` since it is often aliased, e.g.:
https://github.com/google/jax/blob/f3bc86d91c2d6b465f872950918a3880f541d6a4/jax/interpreters/xla.py#L46
Which is then often casted to a list when needed anyway, e.g.:
https://github.com/google/jax/blob/f3bc86d91c2d6b465f872950918a3880f541d6a4/jax/interpreters/xla.py#L359

But I don't have strong feelings on this one, so happy to close the PR 👍

hawkinsp:
Sounds like we're not planning to merge this. Thanks for the PR though!

Issue #5850, opened 2021-02-25
Implement np.mgrid, np.ogrid, & np.r_, np.c_


Conchylicultor:
Numpy has a very convenient `np.mgrid` which allow to easily create coordinate grids: https://numpy.org/doc/stable/reference/generated/numpy.mgrid.html.

I was expecting jax to have the same, but it seems to be missing:

```python
jnp.mgrid  # raise AttributeError
```

---
Tracking implementation of these:

- [x] `jnp.mgrid` #6248
- [x] `jnp.ogrid` #6342
- [x] `jnp.r_` #6593
- [x] `jnp.c_` #6593
- [x] `jnp.ix_`
- [x] `jnp.s_`

jakevdp:
Thanks! Agreed it would be useful to have a jax-native version of this.

VikasHanumegowda:
I'd love to solve this. Anyone with a good direction suggestion for me to point to, to solve this issue? @jakevdp @Conchylicultor

Conchylicultor:
To add to the list, `np.c_` is missing too: https://numpy.org/doc/stable/reference/generated/numpy.c_.html#numpy.c_

jakevdp:
The way to implement these woulc be to create a private class, and override the `__getitem__` method to mimic what numpy does. Something like this:
```python
import jax.numpy as jnp

class _Mgrid:
  def __getitem__(self, key):
    if isinstance(key, slice):
      return jnp.arange(key.start or 0, key.stop, key.step or 1)
    else:
      raise NotImplementedError()

mgrid = _Mgrid()
print(mgrid[:5])
# [0 1 2 3 4]
```

Conchylicultor:
You could also look at the numpy implementation to see how they are implementing it

jakevdp:
> You could also look at the numpy implementation to see how they are implementing it

But note that if you copy from numpy's implementation rather than writing your own version, the code should go in https://github.com/google/jax/tree/master/jax/_src/third_party for licensing reasons.

hawkinsp:
And in this case, where there's probably no need to look at NumPy's implementation to implement the API faithfully and well, it's much better to implement from scratch so we don't have to split up the code based on license. In general:

* please try not to use code from NumPy
* if you do use code from NumPy, it *must* go in `third_party/`, and ultimately may need to be rewritten anyway.

minoring:
Is there an any update for this issue? If not, I would like to progress it :)

[...]

PR #6342, opened 2021-04-05
Implement jnp.ogrid

minoring:
Related to #5850


tests/lax_numpy_test.py
```diff
@@ -4670,6 +4670,33 @@ def testMgrid(self):
                         atol=atol,
                         rtol=rtol)
 
+  def testOgrid(self):
+    assertAllEqual = partial(self.assertAllClose, atol=0, rtol=0)
```
jakevdp, line 4674: I believe `self.assertArraysEqual` will do what you want here.

jakevdp (changes requested):
Looks great!

tests/lax_numpy_test.py
minoring: Thanks for the review!

I tried to use `self.assertArraysEqual` instead of custom `assertAllEqual`.
However, in the case of `self.assertArraysEqual(np.ogrid[0:5, 1:3], jnp.ogrid[0:5, 1:3])`, it raised an error, which is `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`.
Is there a good work around for this exception? Thanks :)

jakevdp: Oh, I see: the issue is you're not comparing arrays, you're comparing tuples of arrays. I think you'll need a custom comparator... the `assertAllClose` thing happens to work, but was not designed for this situation, so it's probably best not to rely on an accident of its implementation. Maybe define a custom function like `assertTuplesOfArraysEqual` and use that?

minoring: Thanks. Now I see the problem!

I defined `assertListOfArraysEqual` (I believe ogrid returns list of arrays. Please correct me if I am wrong :) and use assertArraysEqual instead.

PTAL!

tests/lax_numpy_test.py
```diff
@@ -4671,15 +4671,21 @@ def testMgrid(self):
...
   def testOgrid(self):
-    assertAllEqual = partial(self.assertAllClose, atol=0, rtol=0)
-    assertAllEqual(np.ogrid[:5], jnp.ogrid[:5])
-    assertAllEqual(np.ogrid[:5,], jnp.ogrid[:5,])
-    assertAllEqual(np.ogrid[:5], jax.jit(lambda: jnp.ogrid[:5])())
-    assertAllEqual(np.ogrid[0:5, 1:3], jnp.ogrid[0:5, 1:3])
-    assertAllEqual(np.ogrid[1:7:2], jnp.ogrid[1:7:2])
-    assertAllEqual(np.ogrid[:5, :9, :11], jnp.ogrid[:5, :9, :11])
+    def assertListOfArraysEqual(xs, ys):
+      self.assertEqual(len(xs), len(ys))
```
jakevdp, line 4675: It would be worth asserting that the type is a list, otherwise an N-dimensional array would pass this test.

tests/lax_numpy_test.py
```diff
@@ -4671,15 +4671,21 @@ def testMgrid(self):
...
+      for x, y in zip(xs, ys):
+        self.assertArraysEqual(x, y)
+
+    self.assertArraysEqual(np.ogrid[:5], jnp.ogrid[:5])
+    self.assertArraysEqual(np.ogrid[:5,], jnp.ogrid[:5,])
```
jakevdp, line 4680: This is technically a list of arrays, isn't it?

tests/lax_numpy_test.py
minoring: Yes that's true.
Updated commit.

tests/lax_numpy_test.py
minoring: Indeed.
Updated PR!

tests/lax_numpy_test.py
```diff
@@ -4670,6 +4670,41 @@ def testMgrid(self):
                         atol=atol,
                         rtol=rtol)
 
+  def testOgrid(self):
+    def assertListOfArraysEqual(xs, ys):
+      self.assertTrue(isinstance(xs, list))
+      self.assertTrue(isinstance(ys, list))
```
jakevdp, line 4676: I think we actually have `assertIsInstance`, which will give more informative error messages on failure

jakevdp:
Looks great aside from one last minor comment!

Could you also squash all changes into a single commit before we merge? Thanks!

tests/lax_numpy_test.py
minoring: Yes that's a better way :)

minoring:
Squashed all commits.
Thanks for kind review!

jakevdp:
Thanks!

PR #6345, opened 2021-04-05
[jax2tf]  Add support for shape polymorphism for jax2tf; some refactoring of shape computations in JAX core

gnecula:
Add back support for shape polymorphism to jax2tf, aiming for both soundness and clean interactions with JAX core. 

The main goal is to allow saving of a batch-polymorphic SavedModel for a JAX function, which is a very important feature for users of jax2tf.

**This support is highly experimental. We expect that when the conversion succeeds, it is correct (see below for definition of soundness), but we also expect that there will be valid shape-polymorphic JAX programs for which the conversion will fail, hopefully with a helpful error message. We will aim to increase coverage, without complicating the JAX core internals.**

Unlike previous attempts, we do not use `masking` (see below for a history of this feature). This means that the code should be easier to see it is sound, and it should result in cleaner `lax.py` and `lax_numpy.py`.

We aim to support the conversion of JAX programs for which one or more dimensions of the inputs have sizes that are specified as *dimension variables*. 

**Soundness**: The conversion should succeed only if the converted TF program **faithfully represents all the executions of the original JAX function for any non-zero value of the dimension variables**. This means that we must reject programs that have conditionals dependent on a value of a dimension variable (e.g., `if x.shape[0] == 16: ...`), whether such conditionals are in the user JAX program, in the `jax.numpy` implementation, or in JAX's internal abstract evaluation and transformation rules. We must also reject conditionals `if x.shape[0] == x.shape[1]` if the two dimension are distinct dimension variables. 

In order to ensure soundness, we want to avoid polluting JAX internals with `if shape_is_polymorphic ...`. Instead, we have identified in the JAX internal code a small number of basic shape and dimension computations (e.g., `greater_equal_dim`, `dilate_dim`, `stride_dim`) and have introduced some functions in `core.py` to implement these computations. All that we should see in `lax.py` and `lax_numpy.py` should be references to these helper functions. These helper functions are extensible. They have a base implementation in `core.DimensionHandlerBase` and are customized by sub-classes `masking.DimensionHandlerPoly` and  `shape_poly.DimensionHandlerDimVar`. 

Unlike `masking.Poly`, the dimension variables do not support arbitrary arithmetic and in particular do not duck-type integers. E.g., you must use explicit `core.sum_dim` to add dimensions, and when one of the dimensions is a variable, then only 0 can be added. There is no way to represent the result of other additions.

Completeness is best effort: there will be JAX programs that are shape polymorphic but the jax2tf converter won't recognize as such. One of the goals is to ensure that we can convert with shape polymorphism the output of `jax.vmap` for all primitives. We have tested this fairly exhaustively, by taking all primitive harnesses, picking those for which `jax.vmap` works in JAX, and then converting them shape-polymorphically to TF. All 440 such harnesses work, except for `random_gamma`, `lu`, `custom_linear_solve` and `conv_general_dilated` (for which the batching rule does some shape computation that I have not yet ported). Most cases of `conv_general_dilated` actually work. 

The only know source of unsoundness is that there is no check that the TF function resulting from conversion is applied to arguments whose shapes can be obtained by some replacement of dimension variables with non-zero integers. E.g., `jax2tdf.convert(f_jax, polymorphic_shapes_experimental=["(b, b)"])` should only be applied to a square 2D matrix with non-zero dimensions. Applying it anything else is not guaranteed to produce the same result as in JAX. It *may* result in a TF error, or may not.  

This is a follow-up to a series of previous PRs:
* #4360: add a first version of shape polymorphism, based on `masking.Poly` and `jax.shapecheck`.
* #4878: roll-back the jax2tf shape polymorphism, once we discovered that the underlying `masking.Poly` comparison operations were unsound
* #5286: an attempt to fix jax2tf shape polymorphism (not merged, will abandon)
* #6081: remove more mentions of jax2tf shape polymorphism

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -129,7 +129,7 @@ sets up symbolic links from site-packages into the repository.
 
 To run all the JAX tests, we recommend using `pytest-xdist`, which can run tests in
 parallel. First, install `pytest-xdist` and `pytest-benchmark` by running
-`pip install pytest-xdist pytest-benchmark`.
+`ip install -r build/test-requirements.txt`.
 Then, from the repository root directory run:
 
 ```
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -16,7 +16,6 @@
 # pytype: skip-file
 
 import builtins
-import collections
 from enum import IntEnum
 import functools
 import itertools
@@ -76,7 +75,7 @@ def _try_broadcast_shapes(shapes):
   if not rank: return ()  # scalar case
   result_shape = [None] * rank
   for i, sizes in enumerate(zip(*shapes)):
-    non_1s = set([d for d in sizes if not core.dim_symbolic_equal(d, 1)])
+    non_1s = set([d for d in sizes if not core.symbolic_equal_dim(d, 1)])
     if len(non_1s) > 1:
       return None  # must have equal sizes other than 1-sized axes
     result_shape[i] = next(iter(non_1s), 1)
@@ -3338,7 +3337,7 @@ def _broadcast_in_dim_shape_rule(operand, *, shape, broadcast_dimensions):
     msg = ('broadcast_in_dim broadcast_dimensions must be a subset of output '
            'dimensions, got {} for operand ndim {} and shape {}.')
     raise TypeError(msg.format(broadcast_dimensions, operand_ndim, shape))
-  if not all(core.dim_symbolic_equal_one_of(operand.shape[i],
+  if not all(core.symbolic_equal_one_of_dim(operand.shape[i],
                                             [1, shape[broadcast_dimensions[i]]])
              for i in range(operand_ndim)):
     msg = (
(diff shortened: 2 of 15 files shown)


docs/developer.md
```diff
@@ -129,7 +129,7 @@ sets up symbolic links from site-packages into the repository.
 
 To run all the JAX tests, we recommend using `pytest-xdist`, which can run tests in
 parallel. First, install `pytest-xdist` and `pytest-benchmark` by running
-`pip install pytest-xdist pytest-benchmark`.
+`ip install -r build/test-requirements.txt`.
```
mattjj, line 132:
```suggestion
`pip install -r build/test-requirements.txt`.
```

jax/_src/lax/lax.py
```diff
@@ -75,13 +75,11 @@ def _try_broadcast_shapes(shapes):
...
   result_shape = [None] * rank
   for i, sizes in enumerate(zip(*shapes)):
-    if sizes[:-1] == sizes[1:]:
-      result_shape[i] = sizes[0]  # all equal sizes for this dimension
-    else:
-      sizes = [d for d in sizes if d != 1]
-      if sizes[:-1] != sizes[1:]:
-        return None  # must have equal sizes other than 1-sized axes
-      result_shape[i] = sizes[0] if sizes else 1
+    non_1s = set([d for d in sizes if not core.symbolic_equal_dim(d, 1)])
```
mattjj, line 78:
```suggestion
    non_1s = {d for d in sizes if not core.symbolic_equal_dim(d, 1)}
```

jax/core.py
```diff
@@ -1207,16 +1207,165 @@ def raise_to_shaped(aval: AbstractValue, weak_type=None):
...
+  return _get_dim_handler(d1, d2).symbolic_equal(d1, d2)
+
+def symbolic_equal_one_of_dim(d1: DimSize, dlist: Sequence[DimSize]) -> bool:
+  handler = _get_dim_handler(d1, *dlist)
+  return any([handler.symbolic_equal(d1, d2) for d2 in dlist])
+
+def symbolic_equal_shape(s1: Shape, s2: Shape) -> bool:
+  """See DimensionHandler.symbolic_equal."""
+  return (len(s1) == len(s2) and
+          all(safe_map(symbolic_equal_dim, s1, s2)))
```
mattjj, line 1319: IMO it's better to write this as just `map`, since `safe_map` should be the default (and `unsafe_map`, aka builtin `map`, should be opt-in), but it's not important.

mattjj (approved):
This is great! Thank you.

jax/core.py
gnecula: Done

PR #6354, opened 2021-04-06
Fix jnp.flip for axis tuples

### LINKED ISSUE #6353: jnp.flip crashes when given a tuple of axes
Please:

```python
import jax.numpy as jnp
jnp.flip(jnp.zeros((3, 3, 3)), axis=(1,2))
```

```
  File "/home/buehrle/dev/polyssimo/.venv/lib/python3.6/site-packages/jax/_src/numpy/lax_numpy.py", line 1114, in flip
    return lax.rev(m, [_canonicalize_axis(axis, ndim(m))])
  File "/home/buehrle/dev/polyssimo/.venv/lib/python3.6/site-packages/jax/_src/util.py", line 262, in canonicalize_axis
    axis = operator.index(axis)
TypeError: 'tuple' object cannot be interpreted as an integer
```

ebuehrle:
Fixes #6353. `jnp.flip` would crash when given a tuple of axes. I also tried to add a test for this.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1111,7 +1111,8 @@ def flip(m, axis: Optional[Union[int, Tuple[int, ...]]] = None):
   _check_arraylike("flip", m)
   if axis is None:
     return lax.rev(m, list(range(len(shape(m)))))
-  return lax.rev(m, [_canonicalize_axis(axis, ndim(m))])
+  axis = _ensure_index_tuple(axis)
+  return lax.rev(m, [_canonicalize_axis(ax, ndim(m)) for ax in axis])
 
 
 @_wraps(np.fliplr)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1111,7 +1111,9 @@ def flip(m, axis: Optional[Union[int, Tuple[int, ...]]] = None):
   _check_arraylike("flip", m)
   if axis is None:
     return lax.rev(m, list(range(len(shape(m)))))
-  return lax.rev(m, [_canonicalize_axis(axis, ndim(m))])
+  if isinstance(axis, int):
+    axis = (axis,)
```
jakevdp, line 1115: Instead of this isinstance check, you can use `axis = _ensure_index_tuple(axis)`. This will handle a wider range of input types.

jakevdp (commented):
Looks good, thanks! Just one minor change.

jax/_src/numpy/lax_numpy.py
ebuehrle: Cool, thanks! Changing that.

jakevdp (approved):
Looks great, thanks!

PR #6371, opened 2021-04-08
Fix xmap cache under mesh context manager.

LenaMartens:
Before this change:
```
device_mesh = mesh(..)
with device_mesh:
  xmap(f, ..)(..) # compiling
with device_mesh:
  xmap(f, ..)(..) # recompiling (cache miss)
```

After this change:
```
device_mesh = mesh(..)
with device_mesh:
  xmap(f, ..)(..) # compiling
with device_mesh:
  xmap(f, ..)(..) # cache hit
```

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -1315,6 +1315,16 @@ class Mesh:
   def __repr__(self):
     return f"Mesh({self.devices!r}, {self.axis_names!r})"
 
+  def __eq__(self, other) -> bool:
+    return (type(other) is Mesh and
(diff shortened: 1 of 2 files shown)


LenaMartens:
just realized there might be a bug, let me add another test..

LenaMartens:
I think this is working now, I was worried it wouldn't recompile when the dimensions of the mesh changed (everything else remaining the same), so tested for that.

LenaMartens:
Closing because this was fixed in https://github.com/google/jax/pull/6492.

PR #6375, opened 2021-04-08
[masking] Remove references to masking.Poly from lax and lax_numpy

gnecula:
Previously, in order to increase the coverage of masking we added special
cases in lax.py and lax_numpy.py to avoid exceptions in presence of
masking.Poly.

For example:
```
if not isinstance(d, masking.Poly):
   if some_check(d):
      raise ValueError
```

All such conditionals make the code behave potentially differently when
tracing with masking.Poly than when tracing with concrete shapes, which
makes it hard to ensure soundness.

Perhaps the most egregious was:
```
if type(i) is Poly:
  # dummy index if i is polynomial, doesn't matter for shape inference
  i = 0
```


gnecula:
I checked and internally we have only one test failing, which I am fixing separately.

mattjj (approved):
Thanks so much, George!

PR #6377, opened 2021-04-08
Pass axis name to _match_axes and add to error message.

boyentenbi:
This makes it easier to understand which axis the output is being mapped over in the error.

### DIFF
--- a/jax/interpreters/batching.py
+++ b/jax/interpreters/batching.py
@@ -57,7 +57,8 @@ def batch_subtrace(main, in_dims, *in_vals):
   yield out_vals, out_dims
 
 @lu.transformation
-def _match_axes(axis_size, in_dims, out_dims_thunk, out_dim_dests, *in_vals):
+def _match_axes(axis_size, axis_name, in_dims, out_dims_thunk, out_dim_dests,
+                *in_vals):
   if axis_size is None:
     axis_size, = {x.shape[d] for x, d in zip(in_vals, in_dims) if d is not not_mapped}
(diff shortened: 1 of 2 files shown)


boyentenbi:
Fixed APITest.test_vmap_mismatched_axis_sizes_error_message_issue_705

jax/interpreters/batching.py
```diff
@@ -68,15 +70,16 @@ def batch_subtrace(main, in_dims, *in_vals):
...
+                *in_vals):
   if axis_size is None:
     axis_size, = {x.shape[d] for x, d in zip(in_vals, in_dims) if d is not not_mapped}
   out_vals = yield in_vals, {}
   out_dim_dests = out_dim_dests() if callable(out_dim_dests) else out_dim_dests
   out_dims = out_dims_thunk()
   for od, od_dest in zip(out_dims, out_dim_dests):
     if od is not None and not isinstance(od_dest, int):
-      msg = f"vmap has mapped output but out_axes is {od_dest}"
+      msg = f"vmap has mapped output (axis_name={axis_name}) but out_axes is {od_dest}"
```
apaszke, line 82: Can you please make sure that we omit `axis_name` from the message when the user didn't specify it? You can use `isinstance(core._TempAxisName, axis_name)` to detect that.

boyentenbi: Done, and added another test for that case.

apaszke (approved):
LGTM, but there's a merge conflict

PR #6384, opened 2021-04-08
Implement segment_prod, segment_max, segment_min

jakevdp:
We already had most of the bits in place with `segment_sum`; this implements the rest of the reductions already provided by other parts of `jax.ops`


zhangqiaorjc (approved):
Nice generalization!

PR #6391, opened 2021-04-09
Some updates to jax2tf

marcvanzee:
This PR makes the following changes:
- Replaces `jit_compile` with `experimental_compile` in a call to `tf.function`, since the most recent version of TF (2.4.1) uses this (see [tf.function documentation](https://www.tensorflow.org/api_docs/python/tf/function))
- Not running eval on the entire train set when training the `quickdraw` model reduces training time from 30min to 10min
- Some minor improvements to the quickdraw README
- Using flags in quickdraw for some settings

### DIFF
--- a/jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
+++ b/jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
@@ -12,24 +12,28 @@ and is a part of the code featured in the aforementioned blog post.
 ## Training the model
 
 You can train and export the model yourself by running
+
 ```bash
 $ python3 quickdraw.py
 ```
 
-The training itself lasts roughly 30 minutes, assuming the dataset has already
-been downloaded. You can also skip to the [next section](#interacting-with-the-model)
-to see instructions for playing with a pre-trained model.
+This will first download the dataset if it is not downloaded yet, which is about
+11Gb in total. Assuming the dataset has been downloaded already, training for 5
+epochs takes roughly 10 minutes on a CPU (3,5 GHz Dual-Core Intel Core i7). You
+can also skip to the [next section](#interacting-with-the-model) to see
+instructions for playing with a pre-trained model.
 
 The dataset will be downloaded directly into a `data/` directory in the
 `/tmp/jax2tf/tf_js_quickdraw` directory; by default, the model is configured to
-classify inputs into 100 different classes, which corresponds to a dataset of
-roughly 11 Gb. This can be tweaked by modifying the value of the `NB_CLASSES`
-global variable in `quickdraw.py` (max number of classes: 100). Only the files
-corresponding to the first `NB_CLASSES` classes in the dataset will be
+classify inputs into 100 different classes. This can be tweaked using the
+command-line argument `--num_classes` (max number of classes: 100). Only the
+files corresponding to the first `num_classes` classes in the dataset will be
 downloaded.
 
-The training loop runs for 5 epochs, and the model as well as its equivalent
-TF.js-loadable model are subsequently saved into `/tmp/jax2tf/tf_js_quickdraw`.
+The training loop runs for 5 epochs by default (this can be changed using
+the command-line argument `--num_epochs`), and the model as well as its
+equivalent TF.js-loadable model are subsequently saved into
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
```diff
@@ -12,24 +12,28 @@ and is a part of the code featured in the aforementioned blog post.
...
-corresponding to the first `NB_CLASSES` classes in the dataset will be
+classify inputs into 100 different classes. This can be tweaked using the
+command-line argument `--num_classes` (max number of classes: 100). Only the
+files corresponding to the first `num_classes` classes in the dataset will be
 downloaded.
 
-The training loop runs for 5 epochs, and the model as well as its equivalent
-TF.js-loadable model are subsequently saved into `/tmp/jax2tf/tf_js_quickdraw`.
+The training loop runs for 5 epochs by default (this can be changed using
+the command-line argumnet `--num_epochs`), and the model as well as its
```
bchetioui, line 34: nit: `argumnet` should be `argument` :)

marcvanzee: Thanks, I'll fix that! By the way, did you train the model that is stored in the tarball? I notice there are a lot of different version of the "bin" files, but it seems we only need the one called `group1-shard1of1.bin`, right?

bchetioui: I did train it. Good catch, I gave it a try, I think you are right – the other files are not fetched. Not actually sure now why that is (or why these files were generated), but I think they could be removed!

jax/experimental/jax2tf/examples/saved_model_lib.py
```diff
@@ -116,7 +116,7 @@ def convert_and_save_model(
     params)
   tf_graph = tf.function(lambda inputs: tf_fn(param_vars, inputs),
                          autograph=False,
-                         jit_compile=compile_model)
+                         experimental_compile=compile_model)
```
gnecula, line 119: We normally use tf_nightly for testing this, jax2tf in principle depends on relatively new features of TF. It may work with the 2.4.1 release, but I suspect that in some corners cases is going to fail.

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
marcvanzee: Ok, @gnecula gave me access and I removed the redundant bin files.

jax/experimental/jax2tf/examples/saved_model_lib.py
marcvanzee: I believe it is called `experimental_compile` from TF 2.4 onwards ([see first comment here](https://www.tensorflow.org/xla/tutorials/jit_compile)). I just installed `tf-nightly` (`TF2.6.0-dev20210429`) and there the parameter is also called `experimental_compile`. Or perhaps I am misunderstanding something?

gnecula: It used to be called experimental_compile but in November it was declared non-experimental, and was renamed to jit_compile. See [here](https://github.com/tensorflow/tensorflow/commit/9df9d06e27882504b75e16e4c444ee22405340b3#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23c). It [seems](https://github.com/tensorflow/tensorflow/releases) that this will be released with 2.5.0. I think that experimental_compile will be kept for backwards compatibility.

marcvanzee: I see, that makes sense. Indeed, I just tried it and `jit_compile` also works in tf-nightly. I've now changed it back to `jit_compile`.

jax/experimental/jax2tf/examples/tf_js/quickdraw/README.md
gnecula: @marcvanzee Can you please try to push some changes to trigger another CI test? I suspect that the failures are flaky.

gnecula:
Actually, I was able to trigger a re-run myself.

PR #6392, opened 2021-04-09
Use the axis names attached to a primitive when selecting the top trace

apaszke:
This is useful e.g. for handling psums of values that are not sharded,
but are also not statically known constants that we can fold.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -242,6 +242,7 @@ class Primitive:
   multiple_results = False  # set for multi-output primitives
   call_primitive = False    # set for call primitives processed in final style
   map_primitive = False     # set for map primitives processed in final style
+  _dispatch_on_params = False  # whether to include axis names form params in dispatch
 
   def __init__(self, name: str):
     self.name = name
@@ -253,7 +254,8 @@ class Primitive:
   def bind(self, *args, **params):
     assert (not config.jax_enable_checks or
             all(isinstance(arg, Tracer) or valid_jaxtype(arg) for arg in args)), args
-    top_trace = find_top_trace(args)
+    top_trace = find_top_trace(
+        args, used_axis_names(self, params) if self._dispatch_on_params else None)
     tracers = map(top_trace.full_raise, args)
     out = top_trace.process_primitive(self, tracers, params)
     return map(full_lower, out) if self.multiple_results else full_lower(out)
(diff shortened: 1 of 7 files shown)


tests/pmap_test.py
```diff
@@ -1324,12 +1324,6 @@ def testJitPmapComposition(self):
     ans = pmap(jit(f), 'i')(x)
     self.assertAllClose(ans, expected, check_dtypes=False)
 
-  def testMakeJaxprOfOpenSpmd(self):
-    f = lambda x: x - lax.psum(x, 'i')
-    shape = (xla_bridge.device_count(), 4)
-    x = np.arange(prod(shape), dtype=np.float32).reshape(shape)
-    make_jaxpr(f)(x)  # doesn't crash
```
apaszke, line 1331: I removed this test, because it was already quite fictional. For example, replacing `lax.psum(x, 'i')` with `lax.psum(1, 'i')` makes it fail with an unbound axis error. If we really want to give people the ability to trace jaxprs with names, then we should ask them to provide the environments to `make_jaxpr`.

mattjj (commented):
Instead of loading this behavior into the standard `bind`, WDYT about having a separate `bind` just for primitives that accept axis names? That is, we can just subclass `Primitive` with something like `AxisPrimitive` with its own `bind` method, and specialize the logic there (i.e. call a special `find_top_trace`). That might also be a step towards not relying on string names of parameters so much.

apaszke:
Yeah I think that sounds reasonable. I was a little worried about dispatch times with this change, but making it more specific should make it less important

jax/core.py
```diff
@@ -811,14 +813,17 @@ def full_lower(val):
   else:
     return val
 
-def find_top_trace(xs) -> Trace:
+def find_top_trace(xs, axis_names=None) -> Trace:
+  top_main: Optional[MainTrace] = None
+  if axis_names:
```
mattjj, line 818: This feels a bit like the style of having two bits of functionality underneath one function, gated by configuration arguments, rather than just separating out two functions.

How about defining a new `bind` method on `AxisPrimitive`? That is, don't make any changes to `Primitive.bind` or to `core.find_top_trace`. Instead just have something like

```python
class AxisPrimitive(Primitive):
  def bind(self, *args, **params):
    top_trace = find_top_trace_with_axis_names(...)
    ...

def find_top_trace_with_axis_names(...):
  ...
```

We could even adapt the `bind` signature here to look like `def bind(self, *args, axis_name, **params):` or something, to be more explicit.

WDYT?

mattjj: (This also avoids the need for the `_dispatch_on_params` flag, as the type tag of `AxisPrimitive` has all the info we need.)

apaszke: I thought about it, but I disliked the code duplication. I am afraid that we'll end up changing those bits (e.g. because of your autodidax-inspired simplifications) and we'll just miss something important. Are you worried about the performance implications of this or is it something else?

mattjj: I'm not worried about performance; I just think it's easier to read if we don't pack functionality together like this. I don't think duplicating a few lines is a problem, and tests will easily catch any issues during a refactor.

mattjj: We can land this for now; no need to block. But I might revise it later!

tests/pmap_test.py
mattjj: Maybe do a global presubmit to see if someone is relying on this behavior, and fix downstream?

mattjj (approved):
LGTM modulo the style disagreement which we can iterate on later.

PR #3350, opened 2020-06-07
make jnp.array faster

mattjj:
fixes #2919

This turned out not to be so hard!

BEFORE (CPU):

```
In [1]: import numpy as np
   ...: import jax.numpy as jnp

In [2]: %timeit np.array([0] * int(1e5))
3.96 ms ± 2.46 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [2]: %timeit jnp.array([0] * int(1e5))
10.3 s ± 49.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

AFTER (CPU):

```
In [3]: %timeit jnp.array([0] * int(1e5))
21.9 ms ± 54.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

The example in #2919 actually used `1e6` rather than `1e5`:

BEFORE (CPU, copied from #2919):

```
import numpy as np
import jax.numpy as jnp

%%timeit
np.array([0] * int(1e6))
> 60.1 ms ± 578 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
jnp.array([0] * int(1e6))
> 3min 12s ± 794 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

AFTER (CPU):

```
In [2]: %timeit jnp.array([0] * int(1e6))
224 ms ± 16.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Seems faster, I dno.

PR #6400, opened 2021-04-09
Use convert_element_type instead of device_put_raw.

pschuh:


jakevdp:
Hi - out of curiosity, what was the impetus for this change? I ask because this kind of change tends to have subtle unintented consequences, so I'd prefer to avoid it unless there is good reason.

mattjj:
Also, this code was written this way for performance, so we'd want to check some benchmarks for any change! #3350 has some basic microbenchmarks for `jnp.array`. (Now that we have some basic benchmark infrastructure, we should add this kind of stuff! )

jax/_src/numpy/lax_numpy.py
```diff
@@ -2944,20 +2944,20 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
   if type(object) is np.ndarray:
     _inferred_dtype = object.dtype and dtypes.canonicalize_dtype(object.dtype)
     lax._check_user_dtype_supported(_inferred_dtype, "array")
-    out = _device_put_raw(object, weak_type=weak_type)
+    out = np.array(object, copy=copy, dtype=dtype)
```
pschuh, line 2947: I think this should maybe be _np_array ?

mattjj (approved):
Great find, Parker! (See jax-dev chat thread for more discussion.)

mattjj:
I ran the simple `timeit jnp.array([0] * int(1e6))` benchmark from #3350 and there was no change (`214 ms ± 1.95` on master, `213 ms ± 1.88` on this branch).

mattjj:
@jakevdp the main motivation is to avoid host-to-device (H2D) transfers during jit tracing, very much analogous to #6014. See the jax-dev chat thread for more discussion.

mattjj:
@googlebot I'm a Googler, and I consent!

PR #6413, opened 2021-04-12
Simplify all-reduce translation rule.

chr1sj0nes:
XLA should now be deterministic on GPU, so we don't need to special-case the GPU backend.


hawkinsp (approved):
This looks fine, but it looks like `flake8` has flagged a now-unused import. Please fix?

hawkinsp:
(The non-flake8 error is not from your PR, so ignore that one.)

PR #6416, opened 2021-04-12
Avoid forming unnecessary constants when lowering NumPy indexing expr…

### LINKED ISSUE #6404: Indexing into arrays hoists empty array constants out of Jaxprs
Repro:
```
def test_fun(x):
  return x[0] + x[1] + x[2] + x[3]
```

Jaxpr:
```
{ lambda a b c d ; e.
  let f = lt 0 0
      g = add 0 4
      h = select f g 0
      i = convert_element_type[ new_dtype=int32
                                weak_type=False ] h
      j = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(1,) ] i
      k = concatenate[ dimension=0 ] a j
      l = gather[ dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
                  slice_sizes=(1,) ] e k
  [...]
}
```
where `a`, `b`, `c`, and `d` are empty int32 arrays - one per indexing operation: `[array([], dtype=int32), array([], dtype=int32), array([], dtype=int32), array([], dtype=int32)]`

In severe cases, thousands of these constants inflate the size of the unoptimized HLO and result in slow compilation. In code with higher-order functions, there is evidence that these constants and the associated `lt`,`add`, and `select` operations do not get optimized away.

Workarounds: `jnp.take` does not create empty constants and `lax.slice` results in even more compact code in situations where its semantics are applicable.

hawkinsp:
…essions.

Rather than progressively forming the indices for gather() by repeated concatentation, instead use a single large concatenation of all indices. This removes the need for a size 0 constant array in most cases.

Fixes #6404

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4714,7 +4714,13 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
 
   use_64bit_index = _any([not core.is_constant_dim(d) or d >= (1 << 31) for d in x_shape])
   index_dtype = int64 if use_64bit_index else int32
-  gather_indices = np.zeros((0,), dtype=index_dtype)  # use np to save a compilation
+
+  # Gather indices.
+  # Pairs of (array, start_dim) values. These will be broadcast into
+  # gather_indices_shape, with the array dimensions aligned to start_dim, and
+  # then concatenated.
+  gather_indices = []
+  gather_indices_shape = []
 
   # We perform three transformations to y before the scatter op, in order:
   # First, y is broadcast to slice_shape. In general `y` only need broadcast to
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -4740,15 +4746,12 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
...
-
-      # Broadcast gather_indices from [..., k] to [..., 1, 1, ..., 1, k].
-      gather_indices = lax.broadcast_in_dim(
-        gather_indices, np.insert(gather_indices.shape, -1, shape),
-        tuple(range(gather_indices.ndim - 1)) + (gather_indices.ndim + ndim - 1,))
-      gather_indices = concatenate([gather_indices] + advanced_indexes, -1)
+
+      start_dim = len(gather_indices_shape)
+      gather_indices += ((lax.convert_element_type(a, index_dtype), start_dim)
+                         for a in advanced_indexes)
```
jakevdp, line 4752: I didn't know you could concatenate a generator expression to a list like this... but it works!

jax/_src/numpy/lax_numpy.py
```diff
@@ -4846,6 +4838,20 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
       msg = "Indexing mode not yet supported. Open a feature request!\n{}"
       raise IndexError(msg.format(idx))
 
+  if len(gather_indices) == 0:
+    gather_indices_array = np.zeros((0,), dtype=index_dtype)
+  elif len(gather_indices) == 1:
+    g, _ = gather_indices[0]
+    gather_indices_array = lax.expand_dims(g, (g.ndim,))
+  else:
+    xs = []
```
jakevdp, line 4847: unused?

jakevdp (approved):

jax/_src/numpy/lax_numpy.py
mattjj: TIL

jax/_src/numpy/lax_numpy.py
hawkinsp: Done.

jax/_src/numpy/lax_numpy.py
hawkinsp: Yeah, I am a little surprised too. But it seems to work, so why create an intermediate list!

PR #6420, opened 2021-04-13
Speed up test generation 7x

apaszke:
There are a few test cases that generate millions of configurations,
only to have a handful of them selected by `cases_form_list`. I've
found all tests that spend over 100ms in case generation and
converted them to a new "test sampler" approach. The result: test
generation time drops from 15s to around 2s. Doesn't sound like much,
but I expect that we all run tests many times daily, so it seems like a
useful thing to have.

The rough idea is that the sampling generators get parameterized by a
sampler function that should be applied to the range of every `for` loop.
This allows us to sample runs of the generator through different
configurations by restricting each loop to a smaller subset. Right now
we always narrow it down to a single randomly selected instance. But,
we still retain the possibility of adding exhaustive testing in the
future, which can be achieved by passing in an identity sampling
function that wouldn't modify any loop ranges.

### DIFF
--- a/jax/test_util.py
+++ b/jax/test_util.py
@@ -52,6 +52,14 @@ flags.DEFINE_integer(
   int(os.getenv('JAX_NUM_GENERATED_CASES', 10)),
   help='Number of generated cases to test')
 
+flags.DEFINE_integer(
+  'max_cases_sampling_retries',
+  int(os.getenv('JAX_MAX_CASES_SAMPLING_RETRIES', 100)),
+  'Number of times a failed test sample should be retried. '
+  'When an unseen case cannot be generated in this many trials, the '
+  'sampling process is terminated.'
+)
+
 flags.DEFINE_bool(
     'jax_skip_slow_tests',
     bool_env('JAX_SKIP_SLOW_TESTS', False),
@@ -793,6 +801,29 @@ def cases_from_gens(*gens):
   for size in sizes:
     for i in range(cases_per_size):
       yield ('_{}_{}'.format(size, i),) + tuple(gen(size) for gen in gens)
+
+def named_cases_from_sampler(gen):
+  seen = set()
+  retries = 0
+  rng = npr.RandomState(42)
+  def choose_one(x):
+    if not isinstance(x, (list, tuple)):
+      x = list(x)
+    return [x[rng.randint(len(x))]]
+  while (len(seen) < FLAGS.num_generated_cases and
+         retries < FLAGS.max_cases_sampling_retries):
+    retries += 1
+    cases = list(gen(choose_one))
+    if not cases:
+      continue
+    if len(cases) > 1:
+      raise RuntimeError("Generator is expected to only return a single case when sampling")
+    case = cases[0]
+    if case["testcase_name"] in seen:
+      continue
+    retries = 0
+    seen.add(case["testcase_name"])
+    yield case
 
 
 class JaxTestLoader(absltest.TestLoader):
--- a/tests/lax_numpy_indexing_test.py
+++ b/tests/lax_numpy_indexing_test.py
@@ -889,20 +889,20 @@ class UpdateOps(enum.Enum):
 
 class IndexedUpdateTest(jtu.JaxTestCase):
 
-  @parameterized.named_parameters(jtu.cases_from_list({
+  @parameterized.named_parameters(jtu.named_cases_from_sampler(lambda s: ({
       "testcase_name": "{}_inshape={}_indexer={}_update={}_sugared={}_op={}".format(
           name, jtu.format_shape_dtype_string(shape, dtype), indexer,
           jtu.format_shape_dtype_string(update_shape, update_dtype), sugared, op.name),
        "shape": shape, "dtype": dtype, "indexer": indexer,
        "update_shape": update_shape, "update_dtype": update_dtype,
        "op": op, "sugared": sugared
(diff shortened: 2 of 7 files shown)


jakevdp:
Thanks for taking a look at this – the test startup time has become increasingly noticeable in recent months. I wonder if it might be better to move toward a more declarative specification of test cases, maybe something that would look like this:
```python
@parametrized_test(
  # positional arg is a function to build the title from kwds below
  lambda **kw: "{shape}_{dtype}_axis={axis}".format(**kw)
  # each kwarg is an iterable; space of test cases is the outer product of all iterables.
  shape=all_shapes,
  dtype=all_dtypes,
  axis=[(0,), (1,)])
def my_test(self, shape, dtype, axis):
  ...
```
I don't think the implementation would be too complicated, and we could quickly extract any number of unique combinations from a flattened index without having to construct the full outer product of possibilities. It also reduces some of the boilerplate in our current approach, and avoids collisions from repeated random draws. What do you think?

hawkinsp:
> Thanks for taking a look at this – the test startup time has become increasingly noticeable in recent months. I wonder if it might be better to move toward a more declarative specification of test cases, maybe something that would look like this:

It's probably worth taking a look at the `jax2tf` test cases here. @gnecula

gnecula:
What we have done for `primitive_harness.py` was to try to avoid the cartesian product of parameter spaces as the default way to generate use cases. A simple example was `reshape`. We test all dtypes with one set of old and new shapes. Then we fix the dtype and test various combinations of new and old shapes. This has reduced dramatically the number of test cases for jax2tf. 

This is more costly in terms of thinking about the problem, but the cartesian product cannot scale too much. 

One other problem we faced with random sampling from a very large cartesian product was that some corner cases were never picked. If we have a space of 1000 raw cases (which happens) and we pick deterministically 50 of them, then many important cases will never be picked.

apaszke:
@jakevdp Thanks for the suggestion. I'm down for adding more sugar, but I didn't want to make this into a huge refactor just yet. Also, it's not like our test spaces are pure cartesian products. They often include some conditionals that filter out certain invalid or redundant parts of the search space and I'm not sure how your solution would express that (this also includes the case where ranges of some parameters might depend on other params). I also don't see how this proposal rules out duplicates when sampling the configurations, but would be curious to hear more if you could elaborate.

@gnecula Sure, this works when you expect your ops to be quite independent in their behavior over multiple configuration parameters, but many of the worst offenders are tests that combine different permutations of multiple axes (e.g. vmap of pmap) and I don't think there's a straightforward way to compress the search space. Maybe we could come up with some equivalence classes based on the ordering or equality of different axes, but it's not as simple as "vary one parameter at a time" in that case.

In any case, I'm not saying this is a perfect solution and I'm 100% ok with someone taking over and implementing another fix if we consider it to be better. But the current startup time has just started to get unbearable for me, so I thought that I'll go ahead and propose something.

jax/test_util.py
```diff
@@ -52,6 +52,14 @@
   int(os.getenv('JAX_NUM_GENERATED_CASES', 10)),
   help='Number of generated cases to test')
 
+flags.DEFINE_integer(
+  'max_retries',
+  int(os.getenv('JAX_MAX_RETRIES', 100)),
```
gnecula, line 57: Perhaps the name could be less generic, e.g., 'JAX_MAX_CASES_SAMPLING_RETRIES'?

gnecula:
Aren't you worried that if you have many test cases from which you sample deterministically a small number, you will systematically miss important ones? Is the plan that occasionally we will run with a very large number of NUM_GENERATED_CASES?

apaszke:
I am worried, but I don't see what alternative we have for the main CI without making it excessively slow. I think that it would be a good idea to run exhaustive tests daily or weekly

gnecula:
Thanks, LGTM!

mattjj (approved):
Great work, and clever mechanism! +1 to George's suggestion on the flag name, but otherwise LGTM.

jax/test_util.py
apaszke: Good point. Updated with your suggestion.

jakevdp:
>  I also don't see how this proposal rules out duplicates when sampling the configurations

I was thinking of doing this by avoiding repeated sampling. For example, something like this:
```
A = [1, 2, 3]
B = ['a', 'b']
M = len(A) * len(B)
for i in random.choice(M, n_trials, replace=False):
  Ai = A[i % 3]
  Bi = B[i // 3]
  result = func(A=Ai, B=Bi)
```
and we could generalize similar logic for N parameters.

jakevdp:
But this is a clever solution that addresses the immediate problem, so I'm fine with mergining it.

My main concerns are
(1) repeated sampling with collision avoidance not being particularly clean / efficient
(2) adding even *more* boilerplate to our test parametrizations, which are already really heavy on boilerplate (e.g. each parameter is generally included five times!)

apaszke:
@jakevdp Oh I see, that's a clever trick! However, our test configurations are pure cartesian products only in a few cases and adding any filtering criterion would necessitate adding some repeated sampling to that scheme as well. So it sadly still does not resolve the full issue.

In any case, I agree that this is far from perfect, but I think the general consensus is that it's still an improvement, so I'm going to go ahead and merge this. About issue (1) I think that we should just have a dedicated flag for exhaustive testing, which is easy to add with my approach, while about (2) this is an orthogonal concern that I wasn't planning to address here.

PR #6428, opened 2021-04-13
Make Weibull, Maxwell sampling parameters nonstatic

NeilGirdhar:
These parameters cannot be static since numpy arrays are not hashable.


jakevdp (approved):
Looks good, thanks! This wasn't caught because the tests only pass scalars to these parameters. It would be useful to have better test coverage here, but as this is a strict improvement I think it's fine as-is.

NeilGirdhar:
Thanks!

Issue #4818, opened 2020-11-06
Support DCE on `make_jaxpr`


tomhennigan:
It would be useful to support dead code elimination on staged out jaxprs as part of `make_jaxpr`, I would suggest making this a new optional argument that would run some simple passes like dead code elimination and perhaps could prune unused arguments from subjaxprs etc.

As a motivating use case, from access to an optimized/minimized jaxpr we would be able to determine which arguments to a given function were "live" and prune any unused inputs.

froystig:
I believe that we _could_ do something like this. It seems easy to start with something sound but not comprehensive, then gradually advance the latter. For example, how do we handle higher-order primitives like scan/loops? A simple analysis might dictate that every value read by a loop primitive is "live". A more thorough attempt might analyze the loop body jaxpr for DCE, maybe calling for some fixpoint procedure.

So, before we proceed with something like this, it'd be good to have an idea of how far an implementation would need to go. A more complete motivating use case would help there.

A related implementation question is whether we can reuse any DCE machinery from our underlying compiler (XLA). I'm not entirely clear on how that would look. Jaxpr and HLO are not identical (notably around control flow, via scan). Again the relevance of this may depend on the use case. For example, is this feature only useful to have under `jit`?

etarassov:
I'd also love to have this.
In my case it's a library that calls a `jit`ed function and passes it all the available data as inputs. The function itself only uses a subset of inputs. I'd love to be able to prune those inputs automatically to avoid the waste of uploading the unused inputs on-device just to drop them immediately.

hamzamerzic:
In my case we build modular networks using Haiku, and the networks can compute multiple outputs. Very often a lot of the outputs aren't used in the outer jitted function (e.g. only used for debugging purposes), hence many of the params Haiku creates end up not being used, but are still expected to be provided. The problem comes into play once we need to sync params across devices, e.g. send from learners to actors in a distributed RL setup, since then we need to send all of the unused parameters as well.

I would like to create a helper function that wraps Haiku init_apply functions such that they produce/expect only the used params.

hamzamerzic:
For completeness, @mattjj posted a snippet that should give us argnums of unused inputs, but it does not perform DCE:

```In [1]: from jax import make_jaxpr

In [2]: def foo(x, y, z):
   ...:     return x + z  # y unused
   ...:

In [3]: make_jaxpr(foo)(1, 2, 3)

Out[3]:
{ lambda  ; a b c.
  let d = add a c
[...]

PR #6430, opened 2021-04-14
Add jaxpr DCE to autodidax.

zhangqiaorjc:
This PR demonstrates how to add a jaxpr DCE feature as requested in #4818.

Will attempt to land in regular JAX instead of Autodidax.

Paired with @mattjj


zhangqiaorjc:
this is already upstreamed in JAX

https://github.com/google/jax/pull/6582

PR #6437, opened 2021-04-14
Simplify named shape handling in jax.core

apaszke:
This makes the named shape joining function significantly simpler and
also has the added benefit of removing the requirement of having a
global total order on named axes, which we definitely shouldn't require!
After all, they might be classes defined by users who are unaware of
e.g. the classes we use internally.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -37,7 +37,7 @@ from . import linear_util as lu
 from .lib import jax_jit
 from jax._src import source_info_util
 from ._src.util import (safe_zip, safe_map, partial, curry, prod, partialmethod,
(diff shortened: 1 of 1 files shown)


mattjj (approved):

jekbradbury:
I think this is fine semantically, but it might still be better to prettyprint named shapes with a canonicalized order (since the order isn't significant). This might just mean sorting any alphabetic named axes.

jekbradbury (approved):

apaszke:
That's a good point. I'll merge this to unblock follow-up patches, and I'll add the sorting in repr tomorrow!

PR #6439, opened 2021-04-14
Fix nanquantile handling of negative nans

### LINKED ISSUE #6431: nanmedian (sometimes) incorrectly returns nan
`nanmedian` of a logged array containing `nan`s incorrectly returns `nan`

```python
import jax
import jax.numpy as np
import itertools

M = np.e # or w/e
a = np.array([0, M, 0])

# log of nans
la1 = np.log(np.array([np.nan, M, np.nan]))
la2 = np.log(np.where(a > 0, a, np.nan))
la3 = np.log(jax.ops.index_update(a, a <= 0, np.nan))

# nans inserted after log
la4 = np.array([np.nan, np.log(M), np.nan])
la5 = np.log(a); la5 = np.where(np.isinf(la5), np.nan, la5)
la6 = np.log(a); la6 = jax.ops.index_update(la5, np.isinf(la6), np.nan)

for pair in itertools.combinations((la1,la2,la3,la4,la5,la6), 2):
    assert np.array_equal(*pair, equal_nan=True) # all True

print(np.nanmedian(la1)) # nan
print(np.nanmedian(la2)) # nan
print(np.nanmedian(la3)) # nan
print(np.nanmedian(la4)) # log(M)
print(np.nanmedian(la5)) # log(M)
print(np.nanmedian(la6)) # log(M)
```

Other nan functions seem to work as expected, e.g.
```python
print(np.nanmean(la1)) # log(M)
print(np.nanmean(la4)) # log(M)
```

jakevdp:
Fixes #6431

The issue was than `nanmedian`, via `nanquantile`, relies on the assumption that nans sort to the end of an array, but lax sorts negative NaNs to the start of the array.

Rather than create a new purpose-built test, I opted to modify the test utils to catch this kind of issue more broadly.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5170,9 +5170,10 @@ def _quantile(a, q, axis, interpolation, keepdims, squash_nans):
     raise ValueError("q must be have rank <= 1, got shape {}".format(shape(q)))
 
   a_shape = shape(a)
-  a = lax.sort(a, dimension=axis)
 
   if squash_nans:
+    a = where(isnan(a), nan, a) # Ensure nans are positive so they sort to the end.
+    a = lax.sort(a, dimension=axis)
     counts = sum(logical_not(isnan(a)), axis=axis, dtype=q.dtype,
                  keepdims=keepdims)
     shape_after_reduction = counts.shape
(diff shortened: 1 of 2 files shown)


hawkinsp:
Is it possible we need to fix https://cs.opensource.google/jax/jax/+/master:jax/_src/lax/lax.py;drc=26e9ebcdae5e610d1173fac1bc3504a9c3f2011f;l=5870 instead?

jakevdp:
Yeah - it looks like that is not behaving as documented. That would be a better fix.

jakevdp:
I removed the custom test and opted to adjust `rand_some_nan` to catch issues like this more widely.

jakevdp:
See also #6443

froystig (approved):

PR #6441, opened 2021-04-14
update djax to use XLA dynamic shapes

mattjj:
Now that jaxlib==0.1.65 supports XLA dynamic shapes on CPU, we can avoid complexity in our own lowerings, at least for the cases that XLA dynamic shapes support (not involving DimIndexingExprs).

### DIFF
--- a/jax/experimental/djax.py
+++ b/jax/experimental/djax.py
@@ -19,8 +19,8 @@ from typing import (Tuple, List, Sequence, Set, Dict, Any, Callable, Union,
                     Optional)
 
 from jax import core
+from jax.core import Var, Literal, Atom, Tracer, AbstractValue
 from jax._src import dtypes
-from jax.core import Var, Literal, Atom, Tracer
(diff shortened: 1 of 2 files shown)


machineko:
> Now that jaxlib==0.1.65 supports XLA dynamic shapes on CPU, we can avoid complexity in our own lowerings, at least for the cases that XLA dynamic shapes support (not involving DimIndexingExprs).

Hey Matthew is dynamic shapes on new XLA implementation have performance hit over fully static shapes? (I can't find information anywhere in docs 😄 and I can rember lax.dynamic_slice give me a hard performance drop few months ago should i give it a try again?)

mattjj:
Closing this because we're going to land pieces in real JAX, rather than djax.

@machineko I didn't notice your comment until now, sorry! Yes, when the compiler isn't allowed to specialize its generated code on shape information we should expect a loss of performance, though exactly how much will depend on a lot of details (of the particular workload, of the backend, etc).

PR #6450, opened 2021-04-15
Improve annotation in docs for TypeVar in jit/vmap

shoyer:


gnecula:
I think that it would be useful for posterity to explain the description what problem this solves. I copied this from chats:


our docstring type formatting for readthedocs looks kind of weird right now, with function argument types showing up as `~F`

shoyer:
Thanks for the feedback, @gnecula. I probably should have marked this PR as a "work in progress" -- I mostly made this into a PR so I could test the doc build without running this locally -- and unfortunately what I tried here doesn't actually work!

PR #6453, opened 2021-04-15
autodidax: add cond, start while_loop

mattjj:
Not only does staged-out control flow give a more complete picture of JAX internals, but also they'll be critical to using Autodidax as a prototyping platform (e.g. for side-effect models) because staged-out control flow often puts the test to design ideas.

Interestingly, `cond` and `while_loop` were not much easier than mainline JAX. One idea Dougal had, which might simplify the binder manipulation, is not to closure-convert everything and instead allow open jaxprs. That's worth trying! But I thought we could get this version in first (especially since it'll unblock other experimentation on autodidax).

After this PR, the next two big steps are:
* finish `while_loop` (or decide it doesn't add enough to include, and we should remove it?)
* tests
* final-style call primitives (maybe modeling `pmap`, or `custom_jvp`)

I'd like to prototype a cleaner version of a batched `while_loop` here, hence my interest in improving `while_loop`.

### DIFF
--- a/docs/autodidax.ipynb
+++ b/docs/autodidax.ipynb
@@ -3077,7 +3077,7 @@
     "\n",
     "There are actually two rules to write: one for trace-time partial evaluation,\n",
     "which we'll call `xla_call_partial_eval`, and one for partial evaluation of\n",
-    "jaxprs, whicch we'll call `xla_call_peval_eqn`."
+    "jaxprs, which we'll call `xla_call_peval_eqn`."
    ]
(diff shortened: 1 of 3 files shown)


8bitmp3:
This is very cool @mattjj

froystig (approved):
Nice. Maybe remove the partially complete `while_loop` and add `cond` in isolation, so that this PR itself can be a reference for adding a single higher-order primitive?

PR #6460, opened 2021-04-15
Fix Zero handling in select_jvp.

LenaMartens:
This fixes an actual bug seen in the wild, but not sure if it's the right fix.

An error occurs calling `_zeros(on_false_dot)`, when `on_false_dot` is a Zero with a `float0` dtype (because it's calling `np.result_type(x)`). 
I forget, do we expect full_like to work on such a Zero? (Maybe that is the true fix)


mattjj (approved):
Totally the right fix!

PR #6461, opened 2021-04-15
Enable avals-with-names in xmap

apaszke:
Starting from this change, we start introducing xmapped names when
tracing the xmap jaxpr and eliminating them from avals when the values
are returned. This lets us enable two long-awaited checks:
1. Returning values that are mapped along more axes than `out_axes`
   declare now results in a readable error, instead of an internal
   vmap assertion.
2. We catch the resource-overlap error triggered by making two axes
   mapped to the same resources coincide in a single value.

One thing that I'm not sure about is whether those should raise a `TypeError` or a `ValueError`.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -1705,14 +1705,48 @@ def subst_axis_names(primitive: Primitive, params: ParamDict, subst: AxisSubst) 
     new_params[name] = subst_axis_names_jaxpr(jaxpr, shadowed_subst)
   return new_params
 
+class DuplicateAxisNameError(Exception):
+  def __init__(self, var):
+    self.var = var
+    self.eqn = None
+
+def subst_axis_names_var(v: Var, subst: AxisSubst, var_map: Dict[Var, Var]) -> Var:
+  # Var identity is load-bearing, so we can't have duplicates!
+  if v is unitvar: return v
+  if v is dropvar: return v
+  assert v not in var_map
+  if not hasattr(v.aval, 'named_shape'):
+    var_map[v] = v
+    return v
+  names = tuple(it.chain.from_iterable(subst(name) for name in v.aval.named_shape))
+  named_shape = {name: axis_frame(name).size for name in names}
+  if len(named_shape) != len(names):
+    raise DuplicateAxisNameError(v)
+  new_v = Var(v.count, v.suffix, v.aval.update(named_shape=named_shape))
+  var_map[v] = new_v
+  return new_v
+
+def subst_axis_names_eqn(eqn: JaxprEqn, subst: AxisSubst, var_map: Dict[Var, Var]) -> JaxprEqn:
+  invars: List[Atom] = [v if isinstance(v, Literal) else var_map[v] for v in eqn.invars]
+  try:
+    outvars = [subst_axis_names_var(v, subst, var_map) for v in eqn.outvars]
+  except DuplicateAxisNameError as e:
+    e.eqn = eqn
+    raise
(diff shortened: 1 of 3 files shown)


jax/core.py
```diff
@@ -1705,14 +1705,48 @@ def shadowed_subst(name):
...
+
+def subst_axis_names_var(v: Var, subst: AxisSubst, var_map: Dict[Var, Var]):
+  # Var identity is load-bearing, so we can't have duplicates!
+  if v is unitvar: return v
+  if v is dropvar: return v
+  assert v not in var_map
+  if not hasattr(v.aval, 'named_shape'):
+    var_map[v] = v
+    return v
+  names = sum([subst(name) for name in v.aval.named_shape], ())
```
mattjj, line 1721:
```suggestion
  names = tuple(it.chain.from_iterable(subst(name) for name in v.aval.named_shape))
```

jax/core.py
```diff
@@ -1705,14 +1705,48 @@ def shadowed_subst(name):
     new_params[name] = subst_axis_names_jaxpr(jaxpr, shadowed_subst)
   return new_params
 
+class DuplicateAxisNameError(Exception):
+  def __init__(self, var):
+    self.var = var
+    self.eqn = None
+
+def subst_axis_names_var(v: Var, subst: AxisSubst, var_map: Dict[Var, Var]):
```
mattjj, line 1713:
```suggestion
def subst_axis_names_var(v: Var, subst: AxisSubst, var_map: Dict[Var, Var]
                         ) -> Var:
```

jax/experimental/maps.py
```diff
@@ -636,6 +667,23 @@ def to_mesh(axes):
     return (tuple(unsafe_map(to_mesh, in_axes)),
             tuple(unsafe_map(to_mesh, out_axes)))
 
+def _check_out_avals_vs_out_axes(out_avals: Sequence[core.AbstractValue],
+                                 out_axes: Sequence[AxisNamePos],
+                                 global_axis_sizes: Dict[AxisName, int]):
+  defined_axes = set(global_axis_sizes.keys())
```
mattjj, line 673:
```suggestion
  defined_axes = set(global_axis_sizes)
```

😛

jax/core.py
```diff
@@ -1705,14 +1705,48 @@ def shadowed_subst(name):
...
+    return v
+  names = sum([subst(name) for name in v.aval.named_shape], ())
+  named_shape = {name: axis_frame(name).size for name in names}
+  if len(named_shape) != len(names):
+    raise DuplicateAxisNameError(v)
+  new_v = Var(v.count, v.suffix, v.aval.update(named_shape=named_shape))
+  var_map[v] = new_v
+  return new_v
+
+def subst_axis_names_eqn(eqn: JaxprEqn, subst: AxisSubst, var_map: Dict[Var, Var]):
```
mattjj, line 1729:
```suggestion
def subst_axis_names_eqn(eqn: JaxprEqn, subst: AxisSubst, var_map: Dict[Var, Var]
                         ) -> JaxprEqn:
```

mattjj (approved):
Nice!

jekbradbury (approved):

jax/core.py
apaszke: What kind of code golf is that? My version was shorter 😛

mattjj: Ah true! But it's efficiency golf: I think the `sum` creates more temporary tuples (one for each pair of tuples added).

PR #6466, opened 2021-04-15
sparse: add initial higher-level sparse objects

jakevdp:
This PR proposes some higher-level wrappers of the low-level sparse primitives added in #6429

These are specifically marked as not being the final API, but will be generally useful for building benchmarks of sparse operations.

### DIFF
--- a/jax/experimental/sparse_ops.py
+++ b/jax/experimental/sparse_ops.py
@@ -26,9 +26,19 @@ These routines have reference implementations defined via XLA scatter/gather
 operations that will work on any backend, although they are not particularly
 performant. On GPU runtimes with jaxlib 0.1.66 or newer built against CUDA 11.0
 or newer, each operation is computed efficiently via cusparse.
-"""
 
+Further down are some examples of potential high-level wrappers for sparse objects.
+(API should be considered unstable and subject to change).
+"""
+import functools
+import operator
+
+from typing import Any, Tuple
+
+from jax import api
 from jax import core
+from jax import jit
+from jax import tree_util
 from jax.interpreters import xla
 from jax.lib import cusparse
 from jax.lib import xla_bridge
@@ -38,6 +48,19 @@ import numpy as np
 
 xb = xla_bridge
 xops = xla_client.ops
+
+Dtype = Any
+
+#--------------------------------------------------------------------
+# utilities
+@functools.partial(jit, static_argnums=1)
+def _csr_to_coo(indptr, nnz):
+  return jnp.cumsum(jnp.zeros_like(indptr, shape=nnz).at[indptr].add(1)) - 1
+
+@functools.partial(jit, static_argnums=1)
+def _coo_to_csr(row, nrows):
+  indptr = jnp.zeros(nrows + 1, row.dtype)
+  return indptr.at[1:].set(jnp.cumsum(jnp.bincount(row, length=nrows)))
 
 #--------------------------------------------------------------------
 # csr_todense
@@ -60,8 +83,7 @@ def csr_todense(data, indices, indptr, *, shape):
 
 @csr_todense_p.def_impl
 def _csr_todense_impl(data, indices, indptr, *, shape):
-  row = jnp.zeros_like(indices).at[indptr].add(1).cumsum() - 1
-  return _coo_todense_impl(data, row, indices, shape=shape)
+  return _coo_todense_impl(data, _csr_to_coo(indptr, len(indices)), indices, shape=shape)
 
 @csr_todense_p.def_abstract_eval
 def _csr_todense_abstract_eval(data, indices, indptr, *, shape):
@@ -99,10 +121,9 @@ def csr_fromdense(mat, *, nnz, index_dtype=np.int32):
     indices : array of shape ``(nnz,)`` and dtype ``index_dtype``
     indptr : array of shape ``(mat.shape[0] + 1,)`` and dtype ``index_dtype``
   """
-  return csr_fromdense_p.bind(
-      mat,
-      nnz=nnz,
-      index_dtype=np.dtype(index_dtype))
+  mat = jnp.asarray(mat)
+  nnz = core.concrete_or_error(operator.index, nnz, "nnz argument of csr_fromdense()")
+  return csr_fromdense_p.bind(mat, nnz=nnz, index_dtype=np.dtype(index_dtype))
 
 @csr_fromdense_p.def_impl
 def _csr_fromdense_impl(mat, *, nnz, index_dtype):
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse_ops.py
```diff
@@ -440,3 +447,129 @@ def _coo_matmat_gpu_translation_rule(c, data, row, col, B, *, shape, transpose):
 if cusparse and cusparse.is_supported:
   xla.backend_specific_translations['gpu'][
       coo_matmat_p] = _coo_matmat_gpu_translation_rule
+
+#----------------------------------------------------------------------
+# Sparse objects (APIs subject to change)
+class JAXSparse:
+  """Base class for high-level JAX sparse objects."""
+  _data: Tuple[jnp.ndarray, jnp.ndarray, jnp.ndarray]
```
shoyer, line 455: nit: This feels a little out of place to me in a baseclass. It is not obvious that there will always be exactly three arrays backing each sparse matrix. I would lean towards leaving `_data` and `__init__` up to subclasses to implement. I guess it works for now, though.

jakevdp: Thanks for the suggestion - I rearranged things a bit, I think it's cleaner now.

GJBoth:
I'd be happy to contribute here! It'll be my first time contributing, so a little help would be appreciated. What is the exact aim? Is is to recreate the sparse set of [scipy](https://docs.scipy.org/doc/scipy/reference/sparse.html)? Support seems to exist in XLA, so that means mostly writing wrappers for Jax, right? If so - any pointers on where to get started? Thanks!

jakevdp:
@GJBoth - good question! At this point I don't think the goal is to build-out a full-featured high-level sparse object. My hope with this is to explore how JAX can most effectively support the kinds of sparse operations the users want. A few thoughts in no particular order:

- 2D CSC/CSR/COO may not be the abstraction we want in the end... [pydata-sparse](https://github.com/pydata/sparse/) has N-dimensional sparse arrays implemented using COO or a multi-dimensional generalization of CSR. Similarly [taco](http://tensor-compiler.org/) has routines for general N-dimensional sparse computation. That route may be better, particularly for some deep learning workflows that e.g. have 4D batched arrays of sparse image sequences
- XLA scatter/gather implementations only go so far... any operations we implement should have efficient lowerings on as many backends as possible - I've started with cusparse on GPU because the low-level routines are already available in our GPU build process. This means that every non-trivial operation should be either implemented as a primitive or composed of other primitives
- JAX is about composability, and in particular pmap/vmap, jit, and grad are central to the API. I want to spend some time exploring how we can best implement gradients of sparse operations or batched sparse operations, and also think about how ongoing work on dynamic shapes can be utilized to make jit of sparse operations more efficient and usable. With all that up in the air I want to keep the design minimal for now in case we need to re-think our approach based on what we find
- all of the above should be motivated by real-world workflows that people have. I'm starting to gather a few example workflows so we can have that in mind.
 
Anyway, I hope that helps frame things here - I'd love to hear any thoughts you have!

GJBoth:
Thanks for the extensive reply. If I'm understanding you correctly, the four directions would basically be 1) higher dimensional sparse arrays, 2) backend / performance and 3) "Jaxization". I'm summarizing some thoughts about each one below:

**Higher dimensional arrays**
What are the usecases for this? Is it often sparse image sequences? Couldn't this be implemented as stacked sparse 2D arrays int his case? My first response was that if we could get vmap / pmap to work properly with sparse arrays, this could become a relatively simple highlevel implementation.

**Backend / performance**
This is something I'm a little hesitant about. It'll be my first time contributing low-level stuff and going beyond XLA calls seems a bit too much for now. Additionally, I think that once we have a decent structure / API we can always work on this. It could be useful to get a few baselines on pytorch and/or tensorflow to see what we're aiming for. If we can get similar performance using relatively high level XLA I'd personally be content. 

Additional question - am I understanding correctly that your current low-level sparse ops are implemented using cusparse and not XLA? Or are these two not mutually exclusive?

**Jaxization**
For me this feels like the most interesting. I think being able to integrate sparse matrices into the jax workflow would be really cool, even if they were limited to 2D, only COO type etc. We could try to get something minimal (only 2D, only COO format) going with pmap / vmap / jit / grad to get some experience and feel for the problem, and then evaluate.

Would it be a good idea if I do some research into what's available already and what they implement? Both tensorflow and pytorch have sparse support and are differentiable. pydata-sparse and taco aren't at first look (I'm not familiar with both so I could be wrong) and I think that's a key issue. What do you think?

shoyer (approved):

jakevdp:
@GJBoth – eventually it may be better to move this discussion to a dedicated issue, but a few answers for now:

> **Higher dimensional arrays**
> What are the usecases for this?

My main motivation here is that MLIR in the future may be adding general N-dimensional sparse support (see https://llvm.discourse.group/t/rfc-introduce-a-sparse-tensor-type-to-core-mlir/2944) and it would be useful to not bake-in assumptions of two-dimensional storage.

> **Backend / performance**
> am I understanding correctly that your current low-level sparse ops are implemented using cusparse and not XLA? Or are these two not mutually exclusive?

Currently the operations in `sparse_ops.py` are implemented in terms of XLA, with GPU-only translation rules that compute things via cusparse. The XLA implementation is not particularly efficient, so these purpuse-built translation rules will make things faster. In the future we might target MLIR on CPU, and similar general sparse compilers on GPU/TPU when available.

This is similar to how JAX treats many linear algebra routines: for example, SVD has translation rules that use appropriate LAPACK implementations on CPU/GPU

> **Jaxization**
> Both tensorflow and pytorch have sparse support and are differentiable... Would it be a good idea if I do some research into what's available already and what they implement?

Yes, that would be interesting! In particular I'm curious if they are essentially just doing autodiff over the low-level array representations, or if there's some logic around sparse matrices as an entity. e.g. does the gradient of a matrix-vector product look the same regardless of whether that matrix is in CSR or COO form? In what cases might it make sense to represent gradients as sparse matrices?

GJBoth:
Alright, that makes sense. So basically we're looking for pointers / prior work on:
1) How to handle n-D sparse arrays. This seems to be only implemented in taco and pydata-sparse, not pytorch/tensorflow at first sight.
2) Possible lower-level efficient ops for various backends (e.g. cusparse for gpu, MLIR for cpu, something for tpu?)
3) How to efficiently do VJPs/JVPs with sparse matrices. Pytorch seems to have support for sparse gradients on some operations (see supported linear algebra on the [pytorch documentation](https://pytorch.org/docs/stable/sparse.html)

I'll also try and define few simple benchmarks on which we can compare performance of your preliminary implementation with pytorch and tensorflow. I'll make a separate issue with these results and summarizing what we discussed above.

Did I capture our discussion above correctly with this approach?

jackd:
@jakevdp I've always thought of `CSX` indices as ragged arrays. Would it be worth implementing a separate class (similar to `tf.RaggedTensor`) and have `n`d `CSX` arrays just a class wrapper around `(values, ragged_indices)`? There are many uses for ragged arrays outside of `CSX` implementations - e.g. batching of NLP/graph networks.

jakevdp:
> Would it be worth implementing a separate class (similar to tf.RaggedTensor) and have nd CSX arrays just a class wrapper around (values, ragged_indices)? 

That may be a worthwhile exploration in itself, but I'm not sure the sparse representation here stands to gain anything from that added layer of abstraction.

Issue #4012, opened 2020-08-10
Please support boolean types in `conv_general_dilated`


romanngg:
Example:
```
x = lax.conv_general_dilated(
    lhs=np.ones((1, 1, 1), np.bool_),
    rhs=np.ones((1, 1, 1), np.bool_),
    window_strides=(1,),
    padding='SAME',
    dimension_numbers=('NCH', 'HIO', 'NCH')
)
```
Gives
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-25-7cb941fb15ac> in <module>()
      4     window_strides=(1,),
      5     padding='SAME',
----> 6     dimension_numbers=('NCH', 'HIO', 'NCH')
      7 )
      8 

6 frames
google3/third_party/py/jax/lax/lax.py in naryop_dtype_rule(result_dtype, accepted_dtypes, name, *avals, **kwargs)
   1924       typename = str(np.dtype(aval_dtype).name)
   1925       typenames = ', '.join(t.__name__ for t in types)
-> 1926       raise TypeError(msg.format(name, typename, i, i, typenames))
   1927   _check_same_dtypes(name, False, *aval_dtypes)
   1928   return result_dtype(*avals)

TypeError: conv_general_dilated does not accept dtype bool at position 0. Accepted dtypes at position 0 are subtypes of floating, complexfloating.
```
Thanks!

shoyer:
Would you be happy with how NumPy defines boolean arithmetic? I.e., `True + True = True`?

Would you also be interested in integer convolutions? That feels a little better defined to me. At least you wouldn't be effectively hitting integer overflow all the time.

romanngg:
Yes, I was thinking exactly about the (R, +, *) -> ({True, False}, OR, AND) correspondence.

Integer convolutions would be great too, but perhaps for my specific usecase I was hoping to squeeze more speed / use less memory by doing these kind of operations on boolean types. But admittedly I don't know how easy it is / what accelerators support efficient boolean convolutions.

jekbradbury:
I don't think any major accelerators except the A100 support efficient boolean (they call it 1-bit) convolutions, and the A100's support isn't yet wired through cuDNN (let alone XLA and JAX).

romanngg:
Thanks James. Just to clarify, do you mean that on other platforms it will not let me save neither space, nor time compared to float32?

michaelklachko:
[...]

PR #6467, opened 2021-04-15
Add support for `preferred_element_type` arg in convolutions

lgeiger:
XLA recently added support for integer convolutions on CPU in https://github.com/tensorflow/tensorflow/commit/89fc1ddd96f25aa9d8eb6a0104ca1fdb11b26753.

This PR adds support for the `preferred_element_type` parameter in the `conv`, `conv_general_dilated`, `conv_with_general_padding`, `conv_general_dilated_patches` and `conv_transpose` which is passed through to XLA.

This PR is analogous to #5213 which added support for `preferred_element_type` in `lax.dot`.

Please note that I don't know if this has any implications for GPU since GPUs don't have support integer convolution yet, but hopefully CI will catch any potential problems related to GPUs. Please let me know if I am missing something.

Do you know if there is a timeline integer GPU convolution support in XLA which would be very useful for research on quantized neural networks?

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -541,7 +541,8 @@ def conv_general_dilated(
   rhs_dilation: Optional[Sequence[int]] = None,
   dimension_numbers: ConvGeneralDilatedDimensionNumbers  = None,
   feature_group_count: int = 1, batch_group_count: int = 1,
-  precision: PrecisionLike = None) -> Array:
+  precision: PrecisionLike = None,
+  preferred_element_type: Optional[DType] = None) -> Array:
   """General n-dimensional convolution operator, with optional dilation.
 
   Wraps XLA's `Conv
@@ -573,6 +574,9 @@ def conv_general_dilated(
       'fastest', see the ``jax.default_matmul_precision`` context manager), or a
       tuple of two ``lax.Precision`` enums or strings indicating precision of
       ``lhs`` and ``rhs``.
+    preferred_element_type: Optional. Either ``None``, which means the default
+      accumulation type for the input types, or a datatype, indicating to
+      accumulate results to and return a result with that datatype.
 
   Returns:
     An array containing the convolution result.
@@ -625,7 +629,8 @@ def conv_general_dilated(
       feature_group_count=feature_group_count,
       batch_group_count=batch_group_count,
       lhs_shape=lhs.shape, rhs_shape=rhs.shape,
-      precision=_canonicalize_precision(precision))
+      precision=_canonicalize_precision(precision),
+      preferred_element_type=preferred_element_type)
 
 def dot(lhs: Array, rhs: Array, precision: PrecisionLike = None,
         preferred_element_type: Optional[DType] = None) -> Array:
@@ -1585,7 +1590,8 @@ def stop_gradient(x):
 
 
 def conv(lhs: Array, rhs: Array, window_strides: Sequence[int],
-         padding: str, precision: PrecisionLike = None) -> Array:
+         padding: str, precision: PrecisionLike = None,
+         preferred_element_type: Optional[DType] = None) -> Array:
   """Convenience wrapper around `conv_general_dilated`.
 
   Args:
@@ -1598,19 +1604,24 @@ def conv(lhs: Array, rhs: Array, window_strides: Sequence[int],
       the backend, a ``lax.Precision`` enum value (``Precision.DEFAULT``,
       ``Precision.HIGH`` or ``Precision.HIGHEST``) or a tuple of two
       ``lax.Precision`` enums indicating precision of ``lhs``` and ``rhs``.
+    preferred_element_type: Optional. Either ``None``, which means the default
+      accumulation type for the input types, or a datatype, indicating to
+      accumulate results to and return a result with that datatype.
 
   Returns:
     An array containing the convolution result.
   """
   return conv_general_dilated(lhs, rhs, window_strides, padding,
-                              precision=precision)
+                              precision=precision,
+                              preferred_element_type=preferred_element_type)
 
 def conv_with_general_padding(lhs: Array, rhs: Array,
                               window_strides: Sequence[int],
                               padding: Union[str, Sequence[Tuple[int, int]]],
                               lhs_dilation: Optional[Sequence[int]],
                               rhs_dilation: Optional[Sequence[int]],
-                              precision: PrecisionLike = None) -> Array:
+                              precision: PrecisionLike = None,
+                              preferred_element_type: Optional[DType] = None) -> Array:
   """Convenience wrapper around `conv_general_dilated`.
 
   Args:
@@ -1631,13 +1642,17 @@ def conv_with_general_padding(lhs: Array, rhs: Array,
       the backend, a ``lax.Precision`` enum value (``Precision.DEFAULT``,
       ``Precision.HIGH`` or ``Precision.HIGHEST``) or a tuple of two
       ``lax.Precision`` enums indicating precision of ``lhs``` and ``rhs``.
+    preferred_element_type: Optional. Either ``None``, which means the default
+      accumulation type for the input types, or a datatype, indicating to
+      accumulate results to and return a result with that datatype.
 
   Returns:
     An array containing the convolution result.
   """
   return conv_general_dilated(
       lhs, rhs, window_strides, padding, lhs_dilation=lhs_dilation,
-      rhs_dilation=rhs_dilation, precision=precision)
+      rhs_dilation=rhs_dilation, precision=precision,
+      preferred_element_type=preferred_element_type)
 
 
 def _conv_transpose_padding(k, s, padding):
@@ -1678,7 +1693,8 @@ def conv_transpose(lhs: Array, rhs: Array, strides: Sequence[int],
                    rhs_dilation: Optional[Sequence[int]] = None,
                    dimension_numbers: ConvGeneralDilatedDimensionNumbers = None,
                    transpose_kernel: bool = False,
-                   precision: PrecisionLike = None) -> Array:
+                   precision: PrecisionLike = None,
+                   preferred_element_type: Optional[DType] = None) -> Array:
   """Convenience wrapper for calculating the N-d convolution "transpose".
 
   This function directly calculates a fractionally strided conv rather than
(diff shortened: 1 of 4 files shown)


jax/_src/lax/lax.py
```diff
@@ -2810,11 +2830,25 @@ def _conv_general_dilated_shape_rule(
                                batch_group_count)
   return tuple(np.take(out_trans, np.argsort(out_perm)))  # type: ignore[arg-type]
 
+def _validate_preferred_elemet_type(input_dtype, preferred_element_type):
+  if dtypes.issubdtype(input_dtype, np.integer) and not dtypes.issubdtype(preferred_element_type, np.integer):
+    raise TypeError("`preferred_element_type` and the original type must both be integral or both be floating point.")
```
jakevdp, line 2835: It looks like this doesn't catch the case when `input_dtype` is a float and `preferred_element_type` is an integer.

Do we need a similar check for complex?

jakevdp (commented):
Looks great, thanks! Just one minor comment.

jax/_src/lax/lax.py
lgeiger: > It looks like this doesn't catch the case when input_dtype is a float and preferred_element_type is an integer.

Good catch! This checking was just copied from `lax.dot` The interpreter throws a sensible error message, but I agree it's better to catch it already here.

lgeiger:
Thanks for the fast review 🚀 

I improved the `dtype` checking in 2ec096d08f5220fc299ba807b767f36777a4a66e and extended it to also handle complex types.

lgeiger:
@jakevdp Out of curiosity, do you know whether there is support for integer convolutions on GPUs planned in XLA? I think this could be incredibly useful for research on quantized neural networks if XLA is able generate fast integer kernels.

jakevdp:
> do you know whether there is support for integer convolutions on GPUs planned in XLA?

I'm not certain... I see you're already following #4012 which is probably the relevant issue for that.

jakevdp (approved):

gnecula:
@lgeiger I have not seen integer convolutions in the plans for GPU for the next quarter. Have you filed a feature request with XLA?

lgeiger:
I wasn't aware of the complex number support so 7870ecb0c48f9d2753f322270b56c811cdfd4bf1 fixes the handling of complex numbers when `expand_complex_convolutions=True` and extends the unittests.

> I have not seen integer convolutions in the plans for GPU for the next quarter. Have you filed a feature request with XLA?

Thanks for the info. No, I haven't filled an issue yet. Would `tensorflow/tensorflow` be the right place for this?
I've only seen a reference to an internal issue so far:
https://github.com/google/jax/blob/0f1ddf6dc39ea29d39b256d4009406c6c1c5c844/tests/lax_test.py#L496

jax/_src/lax/lax.py
```diff
@@ -2940,18 +2975,38 @@ def _conv_general_dilated_translation_rule(
     # error bound (e.g. 1p-24 in case of float32) would be relative to the
     # maximum of real and imaginary parts of the result instead of being
     # satisfied by the real and imaginary parts independently of each other.
+    if preferred_element_type is not None:
+      # Convert complex dtype to types used for real and imaginary parts
+      assert np.issubdtype(preferred_element_type, np.complexfloating)
+      preferred_element_type = np.dtype(f"f{np.dtype(preferred_element_type).itemsize // 2}")
```
lgeiger, line 2981: Not sure if there is a nicer way of doing this. I'm very open to suggestions.

jax/_src/lax/lax.py
```diff
@@ -2940,18 +2975,38 @@ def _conv_general_dilated_translation_rule(
     # error bound (e.g. 1p-24 in case of float32) would be relative to the
     # maximum of real and imaginary parts of the result instead of being
     # satisfied by the real and imaginary parts independently of each other.
+    if preferred_element_type is not None:
+      # Convert complex dtype to types used for real and imaginary parts
+      assert np.issubdtype(preferred_element_type, np.complexfloating)
+      preferred_element_type = xla_client.dtype_to_etype(
+          np.dtype(f"f{np.dtype(preferred_element_type).itemsize // 2}"))
```
lgeiger, line 2982: Not sure if there is a nicer way of doing this. I'm very open to suggestions :)

jakevdp: I'd use a helper function, similar to this: https://github.com/google/jax/blob/dbe9ff5f2657f9c124f60e4183d05423d913af8a/jaxlib/rocsolver.py#L36-L47

jakevdp: or, since you know this is complex and there are only two possibilities, an if/else statement would probably suffice :grin:

lgeiger: Great idea! I wasn't aware that there are only two complex types. Fixed in c8f4173b6ec3093d8a3130a26fe31a0e764dc963

jakevdp (approved):
Looks great - thanks!

Final request: can you squash all the changes into a single commit and force-push to the branch? (see https://github.com/google/jax/blob/master/CONTRIBUTING.md)

lgeiger:
> Final request: can you squash all the changes into a single commit and force-push to the branch?

Ah sorry missed that, I thought that would happen during merge via GitHubs "Squash and merge" option. I rebased the PR and squashed everything into one commit.

jakevdp:
It looks like this is causing some failures in `jax2tf`, because it doesn't handle the new `preferred_element_type` argument (these tests are not currently run in by the github CI).

You can run them locally by making sure you have the most recent version of tensorflow installed, and run `pytest -n auto jax/experimental/jax2tf/tests`

lgeiger:
> It looks like this is causing some failures in jax2tf, because it doesn't handle the new preferred_element_type argument (these tests are not currently run in by the github CI).

:+1: I updated `jax2tf` and rebased the PR.  Let's see what CI thinks

tests/lax_test.py
```diff
@@ -62,6 +62,18 @@
...
+# type and where both are either both floating-point or both integral,
+# which are the only supported configurations.
+preferred_type_combinations = [
+  (np.float16, np.float16), (np.float16, np.float32), (np.float16, np.float64),
+  (dtypes.bfloat16, dtypes.bfloat16), (dtypes.bfloat16, np.float32),
+  (dtypes.bfloat16, np.float64), (np.float32, np.float32), (np.float32, np.float64),
+  (np.float64, np.float64), (np.int8, np.int8), (np.int8, np.int16), (np.int8, np.int32),
+  (np.int8, np.int64), (np.int16, np.int16), (np.int16, np.int32), (np.int16, np.int64),
+  (np.int32, np.int32), (np.int32, np.int64), (np.int64, np.int64),
+  (np.complex64, np.complex64), (np.complex64, np.complex128), (np.complex128, np.complex128)]
```
jakevdp, line 75: `np.complex128` is currently not supported on TPU, so we need to avoid this test case for that device (you can check via `jtu.device_under_test() == 'tpu'`)

jakevdp: Also on CPU we're seeing some failures due to insufficient precision:
```
LaxTest.testConvPreferredElement_lhs_shape=complex64[2,2,9,10]_rhs_shape=complex64[3,2,4,5]_preferred_element_type=<class 'numpy.complex128'>
```
You can probably reproduce this locally with
```
$ JAX_NUM_GENERATED_CASES=50 pytest -n auto tests/lax_test.py -k PreferredElement
```
Also worth testing with X64 enabled:
```
$ JAX_ENABLE_X64=1 JAX_NUM_GENERATED_CASES=50 pytest -n auto tests/lax_test.py -k PreferredElement
```

jakevdp (commented):
Thanks! Just a couple more test failures to address on the more comprehensive non-github CI run.

tests/lax_test.py
lgeiger: Thanks for checking internal CI. I skipped the test on TPU and reduced the precision on CPU for the mentioned cases which should fix the failures.

tests/lax_test.py
```diff
@@ -378,6 +390,53 @@ def fun(lhs, rhs):
...
+        or dtype == np.complex128 or preferred_element_type == np.complex128)):
+      raise SkipTest("64-bit mode disabled")
+    if jtu.device_under_test() == "gpu" and np.issubdtype(dtype, np.integer):
+      # TODO(b/183565702): Support integer convolutions on CPU/GPU.
+      raise SkipTest("Integer convolution not yet supported on GPU")
+    if (jtu.device_under_test() == "tpu" and
+       (dtype == np.complex128 or preferred_element_type == np.complex128)):
+      raise SkipTest("np.complex128 is not yet supported on TPU")
+    tol = None
+    if (jtu.device_under_test() == "cpu" and
```
jakevdp, line 418: I'm actually seeing the same issue on GPU as well (sorry I didn't notice that earlier) so we may as well set this tolerance for all devices

tests/lax_test.py
```diff
@@ -921,16 +980,7 @@ def testDot(self, lhs_shape, rhs_shape, dtype, precision):
...
-      # which are the only supported configurations.
-      for dtype, preferred_element_type in [
-        (np.float16, np.float16), (np.float16, np.float32), (np.float16, np.float64),
-        (dtypes.bfloat16, dtypes.bfloat16), (dtypes.bfloat16, np.float32),
-        (dtypes.bfloat16, np.float64), (np.float32, np.float32), (np.float32, np.float64),
-        (np.float64, np.float64), (np.int8, np.int8), (np.int8, np.int16), (np.int8, np.int32),
-        (np.int8, np.int64), (np.int16, np.int16), (np.int16, np.int32), (np.int16, np.int64),
-        (np.int32, np.int32), (np.int32, np.int64), (np.int64, np.int64)]))
+      for dtype, preferred_element_type in preferred_type_combinations))
   def testDotPreferredElement(self, lhs_shape, rhs_shape, dtype, preferred_element_type):
```
jakevdp, line 984: complex128 on TPU needs to be skipped here as well

jakevdp (commented):
So close! Just a couple more CI issues.

lgeiger:
Thanks for the help. I updated the tests to handle the complex cases on GPU and TPU.

jakevdp (approved):

jakevdp:
Thanks for sticking with this! Looks like we can get this merged now.

lgeiger:
> Thanks for sticking with this! Looks like we can get this merged now.

Sure, thanks for the help and the fast review. Really appreciate it 👍

lgeiger:
> I have not seen integer convolutions in the plans for GPU for the next quarter. Have you filed a feature request with XLA?

For reference, I now opened an issue at https://github.com/tensorflow/tensorflow/issues/49140

PR #6470, opened 2021-04-15
support implicit broadcasting in transpose rules

### LINKED ISSUE #1459: Consider revising broadcasting JVP strategy
We should think about reversing the decision documented here, and locking shapes into more primitives at trace time: https://github.com/google/jax/blob/064014b53cd5cffb1ea44f6bc6f9e4e9074c4752/jax/lax/lax.py#L1518-L1521

(It appears that XLA isn't always able to optimize away the resulting cruft once it's been transposed, perhaps because it can't fuse two `reduce_sum`s -- the transpose of `broadcast_in_dim` -- separated by a `reshape`.)

### LINKED ISSUE #3056: No 'unbroadcasting' when using custom_jvp
```python
>>> import jax.numpy as np
>>> from jax import custom_jvp, grad
>>>
>>> a = np.array([1., 1.])
>>>
>>> @custom_jvp
... def f(x):
...   return a * x
...
>>> @f.defjvp
... def f_jvp(primals, tangents):
...   x, = primals
...   dx, = tangents
...   return a * x, a * dx
...
>>> print(grad(lambda x: np.sum(f(x)))(np.array(1.)))
[1. 1.]  # This should be a scalar
```
It looks like this is because the multiplication `a * x` broadcasts `x`, and normally that would induce a `broadcast_in_dim` in the jaxpr for the jvp, but it doesn't. With the custom_jvp we get
```python
>>> from jax import make_jaxpr, linearize
>>> make_jaxpr(linearize(f, 1.)[1])(1.)
{ lambda b ; a.
  let c = mul b a
  in (c,) }
```
If we linearize an instance of `f` without the custom_jvp we get
```python
>>> make_jaxpr(linearize(lambda x: a * x, 1.)[1])(1.)
{ lambda b ; a.
  let c = broadcast_in_dim[ broadcast_dimensions=()
                            shape=(2,) ] a
      d = mul b c
  in (d,) }
```
Not sure if this is fixable or whether we just need to require users to do explicit broadcasting in their custom_jvp definitions?

### LINKED ISSUE #3201: Using vmap inside defjvp
I have my own defined Jacobian-vector product function. It requires vertorizing function inside its computation. It seems that this will result in shape dismatch if I then use `jax.grad` or `jax.jacrev`. Although I can replace `vmap` result by my manually designed vectorized function, is it possible to allow `vmap` inside defvjp for convenience?
My code is long, so I put a brief description. (I can try to make a simple example if this is unclear):
```python
import jax

@jax.custom_jvp
def f(args):
    # do something
    pass

@f.defjvp
def f_jvp(primals, tangents):
    # do something
    pass

    # an inside function
    def g(args):
        # do something
        pass

    # vectorize it
    # vjp_py WILL RAISE SHAPE EXCEPTION AFTER f_jvp
    bat_g = jax.vmap(g)

    # do something with bat_g
    pass

# they will raise shape error at vjp_py
jax.grad(f)(something)
jax.jacrev(f)(something)
```

### LINKED ISSUE #5849: Shape error combining `custom_jvp` with `vmap`.
Recently I've been using `custom_jvp` with `vmap` and I'm seeing something that looks like a shape error that might involve the multiply operation. The code involves computing the pairwise displacement between points that are parameterized by a position in the unit square, $u_i \in [0,1]^2$, along with an affine transformation $T$ so that $r_i = Tu_i$. We use a `custom_jvp` because we want to compute gradients wrt $r_i$ rather than $u_i$. In traditional JAX fashion, we compute the displacement between two points and then use `vmap` to vectorize. When we compute the gradient of a function wrt the affine transformation we observe (I think) erroneous shapes of the gradient that appear to ignore the vmap.

Here is a repro. Let me know if anything is unclear, if you need more information, or if it turns out I'm just confused.

```python
import jax.numpy as np
from jax.api import vmap, grad, custom_jvp

from jax import lax
from jax import random

from functools import partial

# Transformation Code.

@custom_jvp
def transform(box, R):
  if np.isscalar(box) or box.size == 1:
    return R * box
  elif box.ndim == 2:
    return np.einsum(f'ij,j->i', box, R)
  raise ValueError()

@transform.defjvp
def transform_jvp(primals, tangents):
  box, R = primals
  dbox, dR = tangents
  return (transform(box, R), dR + transform(dbox, R))
  
def periodic_general(box):
  def displacement_fn(Ra, Rb, **kwargs):
    _box = kwargs.get('box', box)
    return transform(_box, Ra - Rb) 

  return displacement_fn

# Repro:

N = 250

# Box can either be specified as a scalar or a 2x2
# Affine transformation.
scalar_box = 1.0
matrix_box = np.eye(2)

displacement = periodic_general(scalar_box)

# Create some randomly positioned points.
key = random.PRNGKey(0)
R = random.uniform(key, (N, 2))

# A simple energy function as the total square of the
# pairwise distances.
# 
# If `box` is a scalar then
# energy_fn :: R -> R
#
# If `box` is a matrix then
# energy_fn :: R^{2\times2} -> R
def energy_fn(box):
  d = partial(displacement, box=box)
  d = vmap(vmap(d, (None, 0)), (0, None))
  return np.sum(d(R, R) ** 2)

# We can verify that the shapes of the primal are correct.
print(energy_fn(scalar_box))
print(energy_fn(matrix_box))

# However, the shape of the gradient is incorrect only for
# the scalar box case.
print(grad(energy_fn)(scalar_box).shape)  # Prints (250, 250, 2)
print(grad(energy_fn)(matrix_box).shape)  # Prints (2, 2)
```

### LINKED ISSUE #6357: Broadcasting error within custom JVP
With the following code
```python

import jax
import jax.numpy as jnp

import numpy as onp
from numpy.testing import assert_array_almost_equal, assert_allclose


@jax.custom_jvp
def projection_unit_simplex(x: jnp.ndarray) -> jnp.ndarray:
  """Projection onto the unit simplex."""
  s = 1.0
  n_features = x.shape[0]
  u = jnp.sort(x)[::-1]
  cssv = jnp.cumsum(u) - s
  ind = jnp.arange(n_features) + 1
  cond = u - cssv / ind > 0
  idx = jnp.count_nonzero(cond)
  threshold = cssv[idx - 1] / idx.astype(x.dtype)
  return jax.nn.relu(x - threshold)


@projection_unit_simplex.defjvp
def projection_unit_simplex_jvp(primals, tangents):
  x, = primals
  x_dot, = tangents
  primal_out = projection_unit_simplex(x)
  supp = primal_out > 0
  card = jnp.count_nonzero(supp)

  # This works:
  #Jacobian = jnp.diag(supp) - jnp.outer(supp, supp) / card
  #tangent_out = jnp.dot(Jacobian, x_dot)

  # This doesn't:
  tangent_out = supp * x_dot - (jnp.dot(supp, x_dot) / card) * supp

  # This doesn't either:
  # supp_x_dot = supp * x_dot
  # tangent_out = supp_x_dot - (jnp.sum(supp_x_dot) / card) * supp

  return primal_out, tangent_out


def test_jacobian():
  rng = onp.random.RandomState(0)
  x = rng.rand(5).astype(onp.float32)
  v = rng.randn(5).astype(onp.float32)

  J_rev = jax.jacrev(projection_unit_simplex)(x)
  J_fwd = jax.jacfwd(projection_unit_simplex)(x)

  p = projection_unit_simplex(x)
  support = (p > 0).astype(jnp.int32)
  cardinality = jnp.count_nonzero(support)
  J_true = jnp.diag(support) - jnp.outer(support, support) / cardinality
  assert_array_almost_equal(J_true, J_fwd)
  assert_array_almost_equal(J_true, J_rev)


def test_vmap():
  proj = jax.vmap(projection_unit_simplex)

  def fun(X):
    return jnp.sum(proj(X) ** 2)

  rng = onp.random.RandomState(0)
  X = rng.rand(4, 5).astype(onp.float32)
  U = rng.rand(4, 5)
  U /= onp.sqrt(onp.sum(U ** 2))
  U = U.astype(onp.float32)

  eps = 1e-3
  dir_deriv_num = (fun(X + eps * U) - fun(X - eps * U)) / (2 * eps)
  dir_deriv = jnp.vdot(jax.grad(fun)(X), U)
  assert_allclose(dir_deriv, dir_deriv_num, atol=1e-3)


if __name__ == '__main__':
  test_jacobian()
  test_vmap()
```
I get the following error
```
  File "/Users/mblondel/opt/anaconda3/lib/python3.7/site-packages/jax/_src/lax/lax.py", line 2115, in _broadcasting_shape_rule
    raise TypeError(msg.format(name, ', '.join(map(str, map(tuple, shapes)))))
TypeError: add got arrays of different rank: (5, 5, 5), (5, 5).
```
even though I don't see any shape errors in my code.

### LINKED ISSUE #6452: Case where grad shape is wrong
I have come across a case where `grad` of a scalar with respect to another scalar gives a vector. I have noticed at least one place where if you change the code you get a scalar (indicated by the fail clause).

```python
from jax import numpy as jnp, vmap, value_and_grad

FAIL = True


def f2(y, z):
    v1 = z
    v2 = jnp.sum(y) + z

    if FAIL:
        return jnp.logaddexp(v1, v2)
    else:
        return v1 + v2


def f1(y, z):
    v = vmap(lambda _y: f2(_y, z))(y)
    return jnp.sum(v)


if __name__ == '__main__':
    y = jnp.ones((3, 2))
    f = lambda z: f1(y, z)
    z = 0.1
    val, g = value_and_grad(f)(z)
    print(val.shape, g.shape)
    assert val.shape == ()
    assert g.shape == ()

```

``` python
jax.__version__ = 0.2.9
jaxlib.__version__ = 0.1.61
```

mattjj:
fixes #6452, fixes #6357, fixes #5849, fixes #3201, fixes #3056, fixes #1459

We took a shortcut in JAX's transposition rules! We didn't support transposing primitive applications that involved implicit broadcasting. Like this:

```
{ lambda c:f32[5] x:f32[] .
  let y:f32[5] = mul c x
  in (y,) }
```

Think of `c` as a constant. That jaxpr represents a linear function on `x`. But we couldn't transpose it with respect to `x` because the `mul` application involves an _implicit_ broadcast (i.e. one built into the `mul` application itself, rather than an explicit `broadcast` primitive application) to make its two argument shapes match. We didn't track enough shape information to recognize it and generate the correct `reduce_sum` in the transpose:

```
{ lambda c:f32[5] ybar:f32[5] .
  let wbar:f32[5] = mul c ybar
      xbar:f32[] = reduce_sum [ axes=[0] ] wbar
  in (xbar,) }
```

Instead, we just set up our JVP rules to avoid generating any linear jaxprs involving implicit broadcasts. That is, we put explicit broadcasts in our JVP rules, so that we'd end up with jaxprs like:

```
{ lambda c:f32[5] x:f32[] .
  let w:f32[5] = broadcast [ sizes=[5] ] x
      y:f32[5] = mul c w
  in (y,) }
```

But if we ever forgot to write those explicit broadcasts in JVP rules, we'd get weird shape errors. At least these JVP rules were all JAX-internal code, right?

Wrong! Once we added `jax.custom_jvp`, it meant that users would have to obey this funny no-implicit-broadcasting mandate in their custom rules. But it's worse: #6452 showed us that `vmap` of `custom_jvp` rules, even those in `jax.numpy`, could violate the mandate.

Lucky for us, in the intervening period, we started propagating shape information much more thoroughly in JAX. In particular, by adding `UndefinedPrimal`s in #2410, we propagated to transpose rules all the shape information needed to support transposing implicit broadcasts.

This PR leverages those upgrades and finally adds support for transposing jaxprs with implicit broadcasting. As a result, we can remove all the explicit broadcasts from our JVP rules, and it fixes multiple issues to boot!

TODO
- [x] write PR description

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -141,7 +141,7 @@ def nextafter(x1: Array, x2: Array) -> Array:
 
   For the smallest usable (i.e. normal) float, use ``tiny`` of ``jnp.finfo``.
   """
-  return nextafter_p.bind(_brcast(x1, x2), _brcast(x2, x1))
+  return nextafter_p.bind(x1, x2)
 
 def floor(x: Array) -> Array:
   r"""Elementwise floor: :math:`\left\lfloor x \right\rfloor`."""
@@ -287,7 +287,7 @@ def complex(x: Array, y: Array) -> Array:
 
   Builds a complex number from real and imaginary parts.
   """
-  return complex_p.bind(_brcast(x, y), _brcast(y, x))
+  return complex_p.bind(x, y)
 
 def conj(x: Array) -> Array:
   r"""Elementwise complex conjugate function: :math:`\overline{x}`."""
@@ -2146,55 +2146,48 @@ def naryop(result_dtype, accepted_dtypes, name, translation_rule=None):
   return prim
 standard_naryop = partial(naryop, _input_dtype)
 
-
+# Decorator for translation rules which adds explicit broadcasting of positional
+# arguments. This is necessary only for a handful of primitives whose XLA
+# implementations do not support broadcasting.
 def _broadcast_translate(translate: Callable):
-  # Decorator for translation rules which adds explicit broadcasting of
-  # positional arguments. This is necessary only for a handful of primitives
-  # whose XLA implementations do not support broadcasting.
-  def _broadcast_array(array, array_shape, result_shape):
-    if array_shape == result_shape:
-      return array
-    bcast_dims = tuple(range(len(result_shape) - len(array_shape),
-                             len(result_shape)))
-    result = xops.BroadcastInDim(array, result_shape, bcast_dims)
+  def _broadcast_array(x, shape, result_shape):
+    if shape == result_shape:
+      return x
+    bcast_dims = tuple(range(len(result_shape) - len(shape), len(result_shape)))
+    result = xops.BroadcastInDim(x, result_shape, bcast_dims)
     return result
 
   def _broadcasted_translation_rule(c, *args, **kwargs):
-    shapes = [c.get_shape(arg).dimensions() for arg in args]
+    shapes = [c.get_shape(x).dimensions() for x in args]
     result_shape = broadcast_shapes(*shapes)
-    args = [_broadcast_array(arg, arg_shape, result_shape)
-            for arg, arg_shape in zip(args, shapes)]
+    args = [_broadcast_array(x, s, result_shape) for x, s in zip(args, shapes)]
     return translate(c, *args, **kwargs)
   return _broadcasted_translation_rule
 
-# NOTE(mattjj): this isn't great for orchestrate fwd mode because it means JVPs
-# get two extra ops in them: a reshape and a broadcast_in_dim (or sometimes just
-# a broadcast). but saving the shape info with the primitives isn't great either
-# because then we can't trace these ops without shape data.
-def _brcast(x, *others):
-  # Used in jvprules to make naryop broadcasting explicit for transposability.
-  # Requires shape info during jvp tracing, which isn't strictly necessary.
-  # We don't need full numpy broadcasting, but otherwise the logic is the same
-  # so we reuse the broadcast_shapes function after filtering out scalars.
-  shapes = tuple(filter(None, map(np.shape, (x,) + others)))
-  shape = shapes and broadcast_shapes(*shapes)
-  if np.shape(x) != shape:
-    return _brcast_to(x, shape)
-  else:
-    return x
-
-
-def _brcast_to(x, shape):
+# Like autograd.numpy.numpy_vjps.unbroadcast, this utility handles transposition
+# involving linear primitives with implicit broadcasting.
+def _unbroadcast(aval, x):
+  if not isinstance(aval, ShapedArray):
+    raise TypeError("transpose with implicit broadcasting of unshaped values")
   x_shape = np.shape(x)
-  assert x_shape != shape
-  if x_shape:
-    assert len(x_shape) == len(shape)
-    broadcast_dimensions, = np.where(np.equal(x_shape, shape))
-    squeezed_dimensions, = np.where(np.not_equal(x_shape, shape))
-    squeezed = squeeze(x, squeezed_dimensions)
-    return broadcast_in_dim(squeezed, shape, broadcast_dimensions)
+  if aval.shape == x_shape:
+    return x
+  assert not aval.shape or len(x_shape) == len(aval.shape)
+  if not aval.shape:
+    return _reduce_sum(x, list(range(len(x_shape))))
   else:
-    return broadcast(x, shape)
+    dims = [i for i, (a, b) in enumerate(zip(x_shape, aval.shape)) if a != b]
+    if config.jax_enable_checks: assert all(aval.shape[i] == 1 for i in dims)
+    return reshape(_reduce_sum(x, dims), aval.shape)
+
+def _maybe_broadcast(target_shape, x):
+  x_shape = np.shape(x)
+  if x_shape == target_shape:
+    return x
+  else:
+    dims = [i for i, (a, b) in enumerate(zip(x_shape, target_shape)) if a == b]
+    squeeze_shape = [x_shape[i] for i in dims]
+    return broadcast_in_dim(reshape(x, squeeze_shape), target_shape, dims)
 
 
 _float = {np.floating}
@@ -2224,9 +2217,10 @@ def _sign_translation_rule(c, x):
 sign_p = standard_unop(_num, 'sign', translation_rule=_sign_translation_rule)
 ad.defjvp_zero(sign_p)
 
-nextafter_p = standard_naryop(
-  [_float, _float], 'nextafter',
-  translation_rule=_broadcast_translate(partial(standard_translate, 'next_after')))
+_nextafter_translation_rule = \
+    _broadcast_translate(partial(standard_translate, 'next_after'))
+nextafter_p = standard_naryop([_float, _float], 'nextafter',
+                              translation_rule=_nextafter_translation_rule)
 
 floor_p = standard_unop(_float, 'floor')
 ad.defjvp_zero(floor_p)
@@ -2346,8 +2340,8 @@ ad.defjvp(atan_p, lambda g, x: div(g, _const(x, 1) + square(x)))
 
 atan2_p = standard_naryop([_float, _float], 'atan2')
 ad.defjvp(atan2_p,
-  lambda g, x, y: _brcast(g, y) * (y / (square(x) + square(y))),
-  lambda g, x, y: _brcast(g, x) * -x / (square(x) + square(y)))
+          lambda g, x, y: g * (y / (square(x) + square(y))),
+          lambda g, x, y: g * -x / (square(x) + square(y)))
 
 sinh_p = standard_unop(_float | _complex, 'sinh')
 ad.defjvp(sinh_p, lambda g, x: mul(g, cosh(x)))
@@ -2398,10 +2392,10 @@ igamma_grad_a_p = standard_naryop([_float, _float], 'igamma_grad_a',
                                                'igamma_grad_a')))
 
 def igamma_gradx(g, a, x):
-  return _brcast(g, a, x) * exp(-x + (a - _ones(a)) * log(x) - lgamma(a))
+  return g * exp(-x + (a - _ones(a)) * log(x) - lgamma(a))
 
 def igamma_grada(g, a, x):
-  return _brcast(g, a, x) * igamma_grad_a(a, x)
+  return g * igamma_grad_a(a, x)
 
 ad.defjvp(igamma_p, igamma_grada, igamma_gradx)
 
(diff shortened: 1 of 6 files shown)


froystig (approved):
nice!!

jax/_src/lax/lax.py
```diff
@@ -2452,10 +2446,29 @@ def _bessel_i1e_jvp(g, y, x):
 imag_p = unop(_complex_basetype, _complex, 'imag')
 ad.deflinear2(imag_p, lambda t, _: [complex(np.zeros((), _dtype(t)), neg(t))])
 
+
+def _complex_transpose_rule(t, x, y):
```
jekbradbury, line 2450: It kind of feels like some of this logic should be able to live in primitive dispatch (i.e., that primitives applied to Zeros should return Zeros)?

jax/_src/lax/lax.py
```diff
@@ -2554,61 +2567,108 @@ def _integer_pow_jvp(g, x, *, y):
...
+  xdot, ydot = tangents
+  primal_out = add(x, y)
+  if type(xdot) is type(ydot) is ad_util.Zero:
+    return primal_out, ad_util.Zero.from_value(primal_out)
+  if type(xdot) is ad_util.Zero:
+    return primal_out, _maybe_broadcast(primal_out.shape, ydot)
+  elif type(ydot) is ad_util.Zero:
+    return primal_out, _maybe_broadcast(primal_out.shape, xdot)
+  else:
+    return primal_out, add(xdot, ydot)
```
jekbradbury, line 2581: It's (IMO) even clearer here that this logic should live in dispatch...

jekbradbury (approved):
This is great! Unsurprisingly, it's _exactly_ analogous to what I need to add for transpose of implicit named-axis broadcasting.

gnecula:
I am curious why not add explicit broadcasting so that the JAX primitives do not have to worry about implicit broadcast? This would keep their semantics and the associated transformation rules simpler. It seems that we have a mix of primitives, some that do support broadcast and some that do not, but I am not sure it is clear for those writing rules which primitives are of which kind. 

I am imagining that we don't add explicit broadcast to make XLA's job's easier? Would XLA produce different results with explicit broadcasts than with implicit ones?

jax/_src/lax/lax.py
mattjj: Well, zeros currently aren't jax types, so we can't handle this in primitive application as primitives don't know about zeros. Instead, zeros are internal to the AD machinery, and have to be handled in the AD rules. That has some nice advantages but maybe some disadvantages too?

Are you +1ing the general idea of thinking about adding symbolic zeros as jax types (rather than having them internal to the AD machinery), or are you suggesting a different change for this PR? (Or maybe I've misunderstood!)

mattjj:
> I am curious why not add explicit broadcasting so that the JAX primitives do not have to worry about implicit broadcast?

That's certainly an option! The historical reason is that for op-by-op (eager) evaluation, it's more efficient to have broadcasts fused into the ops/computations we dispatch, since it means fewer overall operation executions and fewer temporaries (plus a bit less Python overhead from deciding whether to insert a broadcast). I think that's why implicit broadcasting is built into e.g. NumPy, PyTorch, and TF kernels.

With JAX we have the advantage that we can decouple what an op/computation is from what a language primitive is, since we can just wrap a jit around lax functions and/or jax.numpy functions. That way we'd get the efficiency benefits of broadcasts built into ops as well as the simplification benefits of rules that don't have to handle primitives' implicit broadcasts. We didn't follow that approach originally because our jit dispatch overheads were much higher than our primitive dispatch overheads, but now that our jit dispatch overheads have come way down, it's a good longer-term option. (Some confirmatory benchmarking would be helpful but I believe that's how things would work out.)

In the meantime, we should merge this PR, since it's already a simplification for our JVP rules as well as a fix for bugs that are affecting users now. (Actually I suspect there are several open issues this change resolves, though I haven't gathered them up yet.) Then once we're comfortable putting jit everywhere, e.g. on lax and jax.numpy functions (something @hawkinsp is actively working on), we can try removing implicit broadcasting behavior from our primitives, though that might take a more thorough updating of our rules (e.g. any rules which directly bind primitives, like IIRC some vmap rules, may need to be updated).

WDYT?

mattjj:
> I am imagining that we don't add explicit broadcast to make XLA's job's easier? Would XLA produce different results with explicit broadcasts than with implicit ones?

No, I don't think so. The implicit broadcasts in lax primitives are exactly those built into the corresponding HLO primitives. (I believe an early HLO pass just inserts explicit broadcasts.)

(Removing implicit broadcasts from our primitives would make jaxprs deviate more from HLO, and would make them a bit longer for human readers too. But I don't think those are issues we care too much about.)

mattjj:
By the way, the reason this PR had to change some JVP rules (e.g. for `add` and `sub`) is that our [`add_jaxvals_p` primitive](https://github.com/google/jax/blob/d9605c162730907d99377fb3f118ffd343a6b552/jax/ad_util.py#L39) doesn't support implicit broadcasts. But now that we've removed explicit broadcasts from some JVP rules, that means we can't use `add_jaxvals_p` directly in forward-mode (or the `ad.defjvp` convenience wrapper which calls `add_jaxvals_p`). Instead we just call `add` or `sub` in the corresponding JVP rules. We can still use `add_jaxvals_p` everywhere in the transpose machinery, which is its main reason for existing (namely to be a generic mechanism to accumulate cotangents corresponding to fan-out).

mattjj:
I added a list of fixed issues to the top of the OP. I'm not sure I got them all...

tests/api_test.py
```diff
@@ -3782,6 +3782,24 @@ def foo(x):
     expected = 2. * jnp.ones(3)
     self.assertAllClose(ans, expected, check_dtypes=False)
 
+  def test_custom_jvp_vmap_broadcasting_interaction(self):
+    # https://github.com/google/jax/issues/6452
```
mblondel, line 3786: Would it be useful to add the code snippet in #6357 as a non-regression test or is it already indirectly covered by this test? In #6357, the broadcasting error occurs even without vmap.

mattjj: Great idea. I'll do that.

(Actually, running #5849 showed a minor bug in this code, not covered by current tests. I've fixed it in my local branch and will push a regression test for #5849 as well as for #6357.)

jax/_src/lax/lax.py
jekbradbury: Just the former!

PR #6477, opened 2021-04-16
Disable concurrent test_tap_multiple on CPU.

zhangqiaorjc:
New TFRT CPU backend will have concurrent program execution which
can cause outfeed from different programs to interleave.


gnecula:
I agree that the test with `concurrent=True` is a bad test today. There is no fundamental reason why that test should work with today's implementation (separate outfeed for tag and payload). That test is still exercising a behaviour that we want, but to make it pass we need to change the implementation to either use CustomCall (will be done for CPU and GPU) or pack the tag and payload into the same outfeed (will be done for TPU). 

Meanwhile, it is fine to disable the test for GPU and CPU.

gnecula (approved):

PR #6485, opened 2021-04-18
DOC: Modify JAX 101 hyperlinks

minoring:
While I am reading JAX_101 tutorial (which I think is really informative), I found some hyperlinks link to JAX documentation and some links to notebooks.

Therefore, modify hyperlinks to link to JAX documentation instead of notebooks.
Also modify link text to use the title of the tutorials.

### DIFF
--- a/docs/jax-101/01-jax-basics.ipynb
+++ b/docs/jax-101/01-jax-basics.ipynb
@@ -290,7 +290,7 @@
     "id": "yQAMTnZSqo-t"
    },
    "source": [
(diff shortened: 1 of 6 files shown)


jakevdp:
Thanks for taking a look at this! I'm not sure we should make this change. The original authors of the notebooks intended them to be read as notebooks, so linking directly to other notebooks makes sense in that context.

Further, links to the HTML docs should generally use sphinx references so that they don't break if and when the documentation structure is re-arranged. However, if you use sphinx references in the notebook, those references will not work if you view the notebook directly, and so the links will cease to function in that context.

Given these constraints, and the fact that we want the same source to be useable both when viewed on Colab and via the generated doc, we chose to use notebook links when linking between sections of the tutorial. What do you think?

minoring:
Thanks for the review!
I think the reason you gave is fair enough :)
Let me close this issue.

Issue #3614, opened 2020-06-30
shape hint arguments in lax numpy functions to enable compilation


froystig:
Several functions in `jax.numpy` are not `jit`-compilable for shape-level arguments. Specifically, we can't decide their output shape from input shapes alone.

We've given [`jax.numpy.bincount`](https://github.com/google/jax/blob/db8f66d508d536e1312291ba966905ddf910582d/jax/numpy/lax_numpy.py#L1309) an optional `length` argument that's not in standard numpy because it determines the output shape (when it is static/concrete). It's not uncommon for callers to have this information on hand.

We could consider extending other functions similarly to accept optional shape hints:
- [x] argwhere (#6915)
- [x] compress (#21090)
- [x] extract (#21090)
- [x] nonzero (#6501 #7592)
- [x] flatnonzero (#6913)
- [x] 1-argument where (#6910)
- [x] repeat (#3670)
- [x] unique (#6912, #8121, #8186)
- [x] setdiff1d (#8144)
- [x] union1d (#6930, #8143)
- [x] intersect1d (#22979)
- [ ] setxor1d (#23020)

shoyer:
For something like `unique`, what would the other elements of the result be filled with if not unique values? It seems like we probably would need to add another parameter for indicating the `fill_value`.

PR #6501, opened 2021-04-19
Add optional `size` argument to jnp.nonzero to allow use in JIT

jakevdp:
This changes the implementation of `jnp.nonzero` to allow it to be used within JIT by optionally passing a static `size` argument:
```python
>>> import jax.numpy as jnp
>>> from jax import jit
>>> x = jnp.array([0, 1, 1, 0, 0, 1])

# normal calling pattern still works
>>> jnp.nonzero(x)
(DeviceArray([1, 2, 5], dtype=int32),)

# pass a static size and it works with jit()
>>> jit(jnp.nonzero, static_argnums=1)(x, 3)
(DeviceArray([1, 2, 5], dtype=int32),)

# if the size is too large, the result is zero-padded
>>> jit(jnp.nonzero, static_argnums=1)(x, 5)
(DeviceArray([1, 2, 5, 0, 0], dtype=int32),)
```
Performance-wise, this appears to be comparable to the previous approach; using the following test code:
```python
import jax.numpy as jnp
import numpy as np
x = np.random.randn(200, 300, 400)
x[x < 0] = 0
x = jnp.asarray(x)
%timeit [i.block_until_ready() for i in jnp.nonzero(x)]
```
on master gives:
```
760 ms ± 4.71 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
and in this branch gives:
```
772 ms ± 7.67 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ PLEASE REMEMBER TO CHANGE THE '..master' WITH AN ACTUAL TAG in GITHUB LINK.
   * When combined with jaxlib 0.1.66, {func}`jax.jit` now supports static
     keyword arguments. A new `static_argnames` option has been added to specify
     keyword arguments as static.
+  * {func}`jax.nonzero` has a new optional `size` argument that allows it to
+    be used within `jit` ({jax-issue}`6501`)
 * Breaking changes:
   * Arguments to {func}`jax.jit` other than the function are now marked as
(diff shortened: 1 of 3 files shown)


froystig:
this works toward #3614

froystig (approved):
nice!

PR #6503, opened 2021-04-20
Raise error in case of xla_computation(pmap) foot gun.

tomhennigan:


jax/_src/api.py
```diff
@@ -675,6 +675,15 @@ def xla_computation(fun: Callable,
...
   fun_name = getattr(fun, "__name__", "unknown")
 
+  if getattr(fun, "_is_pmap", False):
+    # TODO(tomhennigan): Unbox the pmap and set axis_env rather than throwing?
+    raise ValueError(
+        "`xla_computation(pmap(f, 'i'))(x)` is equivalent to adding an outer "
+        "jit to your pmap'd computation making it hard to interpret the "
+        "resulting xla computation. To get the xla_computation of a pmap'd "
+        "function you should instead use: "
+        "`xla_computation(f, axis_env=[('i', x.shape[0])])(x[0])`")
```
apaszke, line 685: I'm not sure how I feel about making this an error instead of a warning 😕 @mattjj any preferences?

One other minor issue here is that usually `xla_computation` accepts anything that has `.shape` and `.dtype` as an arg, but it doesn't require it to define `__getitem__`, so the recommendation might not always work.

jax/_src/api.py
```diff
@@ -1640,6 +1649,8 @@ def f_pmapped(*args, **kwargs):
         global_arg_shapes=tuple(global_arg_shapes_flat))
     return tree_unflatten(out_tree(), out)
 
+  f_pmapped._is_pmap = True
```
apaszke, line 1652: I think that this is not a pmap-only issue. sharded_jit, xmap and pjit will end up doing the same thing (though e.g. xmap might just crash, because I don't think it works inside a jit).

tomhennigan: As discussed on chat, perhaps a better solution is that these compiling functions (jit, pmap, sharded_jit, pjit) should return Callable objects which have an `xla_computation(..)` endpoint accepting the same args as the wrapped function but returning an XlaComputation instance equivalent to what they would use when called regularly. I think then they are able to do their own argument handling and the user facing API is trivial:

    f = pmap(f, axis)  # could swap out for jit/sharded_jit/xmap etc
    y = f(x)
    xla_comp = f.xla_computation(x)

@mattjj wdyt?

apaszke: I would definitely prefer that and I think that some folks have actually requested that in the past. Then we could even start deprecating `jax.xla_computation`

jax/_src/api.py
mattjj: I am also unsure, and could easily be convinced otherwise, but I lean towards warning instead of error because
1. people will see the warning and so be alerted to the possibly undesirable behavior and be able to rewrite the code to get what they want, disarming the footgun;
2. some folks might _want_ to do xla_computation-of-pmap, perhaps in some nested pmap case (there could even be checked-in tests that already do this...).

I'm not sure if that first premise is true. Maybe some peoples' eyeballs automatically filter out warnings?

Do you folks buy those premises, and further that the premises suggest we should warn instead of error? Are there other premises worth considering?

jax/_src/api.py
mattjj: That sounds like a good idea, though to find out for sure we'd need to work out details. For the short term it might make sense to still land some kind of error check like in this PR (assuming @tomhennigan still wants to pursue it rather than pursuing this new idea).

I agree tagging with `_is_pmap` is likely to be pretty limited/brittle, and not just for the reason cited: it would miss things like `xla_computation(lambda x: pmap(f)(x))`. I think a better implementation mechanism would be just to check the jaxpr that gets produced. After all, it's readily available in this function (i.e. in `computation_maker`). One option is we could check whether it's a single-equation jaxpr, and whether that eqn is the application of a parallel primitive, but maybe there are better versions of this.

tomhennigan: I'm in favour of implementing the `pmap(f).xla_computation(..)` solution, it feels much better to me to have a solution that doesn't allow you to express the footgun than either warning or raising.

As for whether we need to first make progress on this PR, I think we can drop it. I think xla_computation is obscure enough that the urgency of disarming the footgun is quite low.

tomhennigan:
Thanks for the feedback!

PR #6519, opened 2021-04-20
Disable HostCallbackIdTapTest.test_tap_multiple_barriers

zhangqiaorjc:
for TPU too since it runs on CPU.

### DIFF
--- a/tests/host_callback_test.py
+++ b/tests/host_callback_test.py
@@ -534,7 +534,7 @@ class HostCallbackIdTapTest(jtu.JaxTestCase):
           for concurrent in [True, False]))
   def test_tap_multiple(self, concurrent=False):
     """Call id_tap multiple times, concurrently or in sequence. """
-    if concurrent and jtu.device_under_test() in ["cpu", "gpu"]:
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):
We determined it's actually the `barrier_wait` call that happens on all devices (including CPU).

zhangqiaorjc:
@gnecula fyi

gnecula:
I note that this PR actually disables `test_tap_concurrent_True` (like #6477) and not `test_tap_multiple_barrier` as the title suggest. 

I am not sure I understand why this test needs to be disabled on TPU. (In general, I would hope that all tests should work on TPU.) The TPU test will try to run multiple `id_tap` concurrently, and from what I know that is working properly now. Is it not? At the end of the TPU test there is a `barrier_wait` in the test, and another one in `tearDown`, and each `barrier_wait` will outfeed one value on each device, including the CPU. But these `barrier_wait` should be properly sequenced, there should be no concurrency. 

Perhaps the issue that concurrent `id_tap` does not work on TPU (hence nothing to do with `barrier_wait`)? This is actually possible, but I'd like to understand it. I could try to figure this out if I see a log. (See "Debugging" comments at the top of the host_callback.py file).

PR #6534, opened 2021-04-21
Use pytest's filterwarnings feature instead of filtering each test case.

skye:
We often forget to put the per-test-case decorators, resulting in test
failures in cases not covered by github CI (e.g. Cloud TPU
tests). This change filters the "experimental feature" warnings by
default.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -11,6 +11,6 @@ filterwarnings =
     ignore:Using or importing the ABCs.*:DeprecationWarning
     # jax2tf tests due to mix of JAX and TF
     ignore:numpy.ufunc size changed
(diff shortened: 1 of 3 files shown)


jakevdp:
We have similar global filters in `pytest.ini`; maybe put this one there as well? https://github.com/google/jax/blob/90d606fe25297f6adb675391c54d68f77a7a37b7/pytest.ini#L1-L16

apaszke (approved):
LGTM, although I would use the filter Jake has proposed

skye:
Ooh agreed Jake's proposal is better! Done.

apaszke (approved):

PR #6541, opened 2021-04-22
Update linalg.py

cdfreeman-google:
Adding special casing for 3x3 matrices, which speeds up det calculation by ~8x

### DIFF
--- a/jax/_src/numpy/linalg.py
+++ b/jax/_src/numpy/linalg.py
@@ -237,12 +237,35 @@ def _cofactor_solve(a, b):
 
   return partial_det[..., -1], x
 
+def _det_2x2(a):
+  return (a[..., 0, 0] * a[..., 1, 1] -
+           a[..., 0, 1] * a[..., 1, 0])
+
+def _det_3x3(a):
+  return (a[..., 0, 0] * a[..., 1, 1] * a[..., 2, 2] +
+          a[..., 0, 1] * a[..., 1, 2] * a[..., 2, 0] +
+          a[..., 0, 2] * a[..., 1, 0] * a[..., 2, 1] -
+          a[..., 0, 2] * a[..., 1, 1] * a[..., 2, 0] -
+          a[..., 0, 0] * a[..., 1, 2] * a[..., 2, 1] -
+          a[..., 0, 1] * a[..., 1, 0] * a[..., 2, 2])
+
 
 @custom_jvp
 @_wraps(np.linalg.det)
+@jit
 def det(a):
-  sign, logdet = slogdet(a)
-  return sign * jnp.exp(logdet)
+  a = _promote_arg_dtypes(jnp.asarray(a))
+  a_shape = jnp.shape(a)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/linalg.py
```diff
@@ -238,11 +238,32 @@ def _cofactor_solve(a, b):
...
+def _det_3x3(a):
+  return (a[..., 0, 0] * a[..., 1, 1] * a[..., 2, 2] +
+          a[..., 0, 1] * a[..., 1, 2] * a[..., 2, 0] +
+          a[..., 0, 2] * a[..., 1, 0] * a[..., 2, 1] -
+          a[..., 0, 2] * a[..., 1, 1] * a[..., 2, 0] -
+          a[..., 0, 0] * a[..., 1, 2] * a[..., 2, 1] -
+          a[..., 0, 1] * a[..., 1, 0] * a[..., 2, 2])
+
+@jit
+def _det(a):
```
hawkinsp, line 250: I think you can just stick a `@jit` around `det`, no need for another `_det` function.

(Sometimes we use that pattern when we want to handle static arguments specially, but `det` doesn't have any.)

jax/_src/numpy/linalg.py
```diff
@@ -238,11 +238,32 @@ def _cofactor_solve(a, b):
   return partial_det[..., -1], x
 
 
+def _det_3x3(a):
```
hawkinsp, line 241: Is it worth adding a 2x2 case also?

hawkinsp (changes requested):

jax/_src/numpy/linalg.py
cdfreeman-google: ooh good call!

jax/_src/numpy/linalg.py
cdfreeman-google: kk--was about to ask about this actually :D

hawkinsp: BTW the jit should probably be innermost (i.e., closest to the `def`)

jax/_src/numpy/linalg.py
```diff
@@ -237,12 +237,35 @@ def _cofactor_solve(a, b):
 
   return partial_det[..., -1], x
 
+def _det_2x2(a):
+  return ( a[..., 0, 0] * a[..., 1, 1] -
```
hawkinsp, line 241: Remove extra space after `(`

jax/_src/numpy/linalg.py
cdfreeman-google: ++

hawkinsp (commented):
Please squash your commits.

jax/_src/numpy/linalg.py
cdfreeman-google: done, I think!

jax/_src/numpy/linalg.py
cdfreeman-google: also done!

hawkinsp:
A version of this PR was already merged as https://github.com/google/jax/pull/6543

PR #6543, opened 2021-04-22
Added special cases for 2x2 and 3x3 determinant

cdfreeman-google:
Added special cases for 2x2, 3x3 linalg dets, and added test coverage for these cases.


hawkinsp (approved):

cdfreeman-google:
@hawkinsp Would you mind clicking the merge button? It won't let me 😃

Issue #1364, opened 2019-09-17
jax.numpy.log(2., dtype=np.float32) does not accept dtype arg


brianwa84:

hawkinsp:
Actually the issue is a bit more widespread than that: all ufuncs an optional `dtype` argument, not just `log`. https://docs.scipy.org/doc/numpy/reference/ufuncs.html

PR #6545, opened 2021-04-22
Support `dtype` kwarg in `jnp.matmul`

lgeiger:
This PR adds support for the `dtype` keyword argument in `jnp.matmul` to more closely match [`np.matmul`](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html#numpy.matmul) which accepts ufunc arguments since version 1.16.

I didn't add a new test for this PR since these changes only pass through the `dtype` parameter to `dot` which is tests this behaviour [here](https://github.com/google/jax/blob/594309c12b39ccc79df142b85af40c53fa780468/tests/lax_test.py#L971-L1002), but let me know if you would prefer to do the same for `jnp.matmul`.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -47,7 +47,7 @@ from jax.core import UnshapedArray, ShapedArray, ConcreteArray, canonicalize_sha
 from jax.config import config
 from jax.interpreters.xla import DeviceArray, _DeviceArray, _CppDeviceArray
 from jax import lax
-from jax._src.lax.lax import _device_put_raw
+from jax._src.lax.lax import _device_put_raw, _validate_preferred_element_type
 from jax import ops
 from jax._src.util import (partial, unzip2, prod as _prod, subvals, safe_zip,
                            canonicalize_axis as _canonicalize_axis, maybe_named_axis)
@@ -4026,7 +4026,7 @@ def dot(a, b, *, precision=None):  # pylint: disable=missing-docstring
 
 
 @_wraps(np.matmul, lax_description=_PRECISION_DOC)
-def matmul(a, b, *, precision=None):  # pylint: disable=missing-docstring
+def matmul(a, b, *, casting="same_kind", dtype=None, precision=None):  # pylint: disable=missing-docstring
   _check_arraylike("matmul", a, b)
   for i, x in enumerate((a, b)):
     if ndim(x) < 1:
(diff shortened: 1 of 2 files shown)


jakevdp:
There seems to be one difference here that I can see: in Numpy, this works:
```python
x = np.ones((4, 4)).astype('float64')
np.matmul(x, x, dtype='float32')
```
The equivalent operation in JAX fails with the error
```
TypeError: `preferred_element_type` must not be narrower than the original type.
```
There's a similar issue when changing the kind of type via `dtype`:
```python
x = np.ones((4, 4)).astype('int64') 
np.matmul(x, x, dtype='float64')  # <-- no error

x = jnp.ones((4, 4)).astype('int64') 
jnp.matmul(x, x, dtype='float64') 
# TypeError: `preferred_element_type` and the original type must both be integral, both be floating point, or both complex.
```
I think this keyword needs a bit more handling to match numpy's semantics.

lgeiger:
> I think this keyword needs a bit more handling to match numpy's semantics.

@jakevdp Good point, sorry for missing that.

I updated the PR to handle numpy semantics.
This now adds quite a bit of additional code to handle type casting.
So I think in a broader context this PR only makes sense if you plan to add support for the `dtype` keyword argument to other `ufuncs` as well in the future (see #1364). I'll let you decide if support for this argument is worth the added maintenance burden or whether this is something that will be deliberately not supported by JAX in which case I am happy to close this PR.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3821,15 +3821,30 @@ def dot(a, b, *, precision=None):  # pylint: disable=missing-docstring
...
 
-  a, b = _promote_dtypes(a, b)
+  preferred_element_type = None
+  if dtype is None:
+    a, b = _promote_dtypes(a, b)
+  else:
+    lax._check_user_dtype_supported(dtype, "matmul")
+    dtype = dtypes.canonicalize_dtype(dtype)
+    promoted_dtype, _ = dtypes._lattice_result_type(a, b)
+    promoted_dtype = dtypes.canonicalize_dtype(promoted_dtype)
```
jakevdp, line 3839: You can replace these two lines with `promoted_dtype = result_type(a, b)`

jakevdp (commented):
Sorry, I somehow lost track of this PR.

Small comment below, then we should be able to get this merged!

jax/_src/numpy/lax_numpy.py
lgeiger: Great! That's much more readable.

jakevdp:
Sorry, I looked a bit closer at this and I think it needs a bit more thought. In particular, the handling of the `casting` argument seems different than numpy: here it is only referenced if `dtype is None`. I think (though I may be wrong?) that in numpy the `casting` argument applies to casting of the inputs as well.

Back to your previous comment, we may want to step back and decide whether it is a goal to handle optional ufunc keywords consistently across jax; I'm not certain what the answer is.

lgeiger:
>  I think (though I may be wrong?) that in numpy the casting argument applies to casting of the inputs as well.

Good catch. Yes, I think you are right. I would need to look into it a bit more closely, but I think numpy uses something like the following logic:
```python
result_dtype = dtype or np.result_type(a, b)
if not (np.can_cast(a.dtype, result_dtype) and np.can_cast(b.dtype, result_dtype))
    raise
```

> Back to your previous comment, we may want to step back and decide whether it is a goal to handle optional ufunc keywords consistently across jax; I'm not certain what the answer is.

I agree, this is something that should be handled for all ufuncs consistently. I think a first step would be to decide whether it is within the scope of JAX to support the `casting` argument in all ufuncs. Once JAX follows the same ufunc casting rules as numpy, supporting user definable result dtypes should be pretty easy. But then again the question whether the added maintenance burden for supporting all ufunc arguments is a worthwhile investment is still not obvious.

jakevdp:
Sorry to have let this go stale. In the meantime, `jax.numpy.matmul` has added a `preferred_element_type` argument, which gets passed directly to `lax.dot_general`.

I'm going to close this for now; feel free to re-open if you'd like to work on adding the `dtype` argument here. Thanks!

PR #6548, opened 2021-04-23
Add eval_shape to the UnexpectedTracerError too.

LenaMartens:
One side-effect of filtering out JAX internal frames from the UnexpectedTracerError is that we can't see what transform kicked of the trace which leaked the value. There is a trade-off here, because including the JAX frames means there is probably going to be no user code in this stack trace, which is unhelpful too (maybe we could re-include `api.py` specifically?)

We previously added some transform_names to the error to make up for this missing info in #6189, here's one more.


mattjj (approved):
Good improvement!

PR #2414, opened 2020-03-13
lower fori_loop to scan when possible

mattjj:
When a fori_loop specialized on trip count is to be evaluated, it's preferable to generate a scan rather than a while_loop because the former is reverse-mode differentiable while the latter is not. Otherwise they're essentially the same; in particular, no extensive inputs/outputs arise unless reverse-mode autodiff is applied.

Also fixes #2412.

PR #2026, opened 2020-01-19
add custom_jvp / vjp, delete custom_transforms

mattjj:
fixes #116, #1097, #1249, #1275, #1366, #1723, #1670, #1875, #1938, #2345, #2346

See the [design doc](https://github.com/google/jax/blob/custom-transforms3/design_notes/custom_derivatives.md) and [user tutorial](https://github.com/google/jax/blob/custom-transforms3/docs/notebooks/Custom_derivative_rules_for_Python_code.ipynb) for details.

TODO:
- [x] improve error messages
- [x] write a PR message / design doc
- [x] add docstrings and reference docs
- [ ] fix user code (Google code, NumPyro, jaxnet, anything else we think of...)

PR #6549, opened 2021-04-23
enable fori-to-scan lowering

mattjj:
When a `lax.fori_loop` has static bounds, we can lower it to a `scan` and thus support efficient reverse-mode autodiff.

We had attempted to add this in #2414, but had to disable it for [some weird `custom_transforms` reason](https://github.com/google/jax/pull/2414#issuecomment-600183624). That means #2026 unblocked it (and maybe #3370 helped too)!.


mblondel:
Awesome, thanks @mattjj! The docstring needs to updated to mention the lowering to scan.

jekbradbury (approved):

PR #6556, opened 2021-04-24
bug fix for BatchTrace.post_process_custom_jvp

mattjj:
Thanks to @danieldjohnson for spotting this and helping to debug it!


mattjj:
See #8915 for the follow-up.

PR #6565, opened 2021-04-26
setup.py: make cuda extras specifier more consistent

jakevdp:
I went to document this feature and realized that the dash in the specifier is confusing, because its not used when specifying the wheel directly.

This pip extras API has not yet been advertised or documented, so I don't think this change will affect many users.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,8 +47,8 @@ setup(
         'cpu': [f'jaxlib>={_minimum_jaxlib_version}'],
 
         # CUDA installations require adding jax releases URL; e.g.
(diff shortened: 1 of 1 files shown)


skye (approved):

skye:
Maybe we should switch the README to this? I would love to stop specifying the jaxlib version when installing for cuda :slightly_smiling_face:

jakevdp:
Yes! I was starting to switch the readme to this when I realized the dash would likely cause confusion. Once this PR is into a release, I'll plan on updating the README to reflect it.

PR #6575, opened 2021-04-28
Fix cusparse kernel build.

hawkinsp:
The build_wheel.py script was copying the wrong module.

In addition the CUDA stubs from the TF repo were missing a number of cusparse symbols. The updated TF includes the correct stubs.


hawkinsp:
(Just going to merge this one: I'm fairly confident it's the right fix!)

PR #6581, opened 2021-04-28
Bundle libdevice.10.bc with jaxlib wheels.

hawkinsp:
libdevice.10.bc is a redistributable part of the CUDA SDK.

This avoids problems trying to locate a copy of libdevice inside the user's CUDA installation.

### DIFF
--- a/build/BUILD.bazel
+++ b/build/BUILD.bazel
@@ -42,6 +42,7 @@ py_binary(
         "//jaxlib:cusparse_kernels",
         "//jaxlib:cuda_lu_pivot_kernels",
         "//jaxlib:cuda_prng_kernels",
+        "@local_config_cuda//cuda:cuda-nvvm",
     ]) + if_rocm([
         "//jaxlib:rocblas_kernels",
     ]),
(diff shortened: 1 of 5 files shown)


build/BUILD.bazel
```diff
@@ -42,6 +42,7 @@ py_binary(
         "//jaxlib:cusparse_kernels",
         "//jaxlib:cuda_lu_pivot_kernels",
         "//jaxlib:cuda_prng_kernels",
+	"@local_config_cuda//cuda:cuda-nvvm",
```
skye, line 45: Fix indentation

skye (approved):
Just to clarify, this works with all cuda versions?

Also does this significantly increase the jaxlib binary size?

skye:
This is awesome btw

build/BUILD.bazel
hawkinsp: Done.

hawkinsp:
> Just to clarify, this works with all cuda versions?
> 
> Also does this significantly increase the jaxlib binary size?

This should work across CUDA 11 minor releases. For other versions, we are using the same version that came with the CUDA release the user is using.

The file is  ~450k uncompressed, so yes it makes jaxlib bigger but not that much bigger.

PR #6584, opened 2021-04-28
add 'inline' option to xla_call for jaxpr inlining

mattjj:
The basic idea here is to add a boolean 'inline' option to `jit` (i.e. to the `xla_call` primitive). It doesn't affect the behavior of calling a jitted function from Python, i.e. it doesn't affect anything at the "top level", but instead it only affects nested `jit` calls.

With `inline=False` (the default value) we see the previous behavior:

```python
from functools import partial
import jax

@partial(jax.jit, inline=False)
def f(x):
  return x * 2

print(jax.make_jaxpr(f)(3))
```

```
{ lambda  ; a.
  let b = xla_call[ backend=None
                    call_jaxpr={ lambda  ; a.
                                 let b = mul a 2
                                 in (b,) }
                    device=None
                    donated_invars=(False,)
                    inline=False
                    name=f ] a
  in (b,) }
```

But any `jit` call with `inline=True` will get inlined into an outer jaxpr:

```python
@partial(jax.jit, inline=True)
def f(x):
  return x * 2

print(jax.make_jaxpr(f)(3))
```

```
{ lambda  ; a.
  let b = mul a 2
  in (b,) }
```

This jaxpr is much cleaner!

The main motivation for this change is to put `jit` in lots more places within JAX, in particular on lots of `jax.numpy` functions, without cluttering up jaxprs (or affecting tests that check jaxprs, even though we discourage such tests). cc @hawkinsp 

---

The basic implementation mechanism is to tweak the jaxpr-building tracer used for staging out computation (i.e. used for jit, pmap, make_jaxpr, control flow, rather than the jaxpr-building tracer for reverse-mode autodiff), and in particular to tweak the behavior when it encounters a call primitive with an 'inline' parameter. That is, we add two lines to `DynamicJaxprTrace.process_call`, so that if `inline=True` then immediately after forming the jaxpr we apply it to the tracer arguments using `core.eval_jaxpr`:

```python
  # in DynamicJaxprTrace.process_call
  with core.new_sublevel():
    jaxpr, out_avals, consts = trace_to_subjaxpr_dynamic(f, self.main, in_avals)
+ if params.get('inline', False):
+   return core.eval_jaxpr(jaxpr, consts, *tracers)
```

(One might wonder why we bother building the jaxpr first, instead of just applying `fun.call_wrapped` to the tracers immediately. Read on to see why...)

So... what's with all the other changed lines? And what the heck are sublevels anyway!?

The sublevel, a global integer, indicates at trace time how many nested `jit`s we're inside. More precisely, it indicates how many levels of final-style higher-order primitives we're inside, so `pmap`, `xmap`, `custom_jvp`, and `custom_vjp` count too. Levels, on the other hand, count how many transformations we're inside.

Sublevels are only used for catching errors. Each Tracer instance is tagged with a sublevel, just like it's tagged with a level, and we check that sublevels make sense in core.py, particularly in the `full_raise` function. If a Tracer is leaked, e.g. via a side-effect, it might end up interacting with a Tracer of a different sublevel which doesn't make sense.

For example, while this is a _level_ error, to do with transformations like `grad` rather than final-style higher-order primitives like `jit`:

```python
from jax import grad

def f(x):
  y = None
  def g(x):
    nonlocal y
    y = x
    return x
  return grad(g)(x) + y
grad(f)(1.)
```

```
jax.core.UnexpectedTracerError: Encountered an unexpected tracer. [...]
Detail: Can't lift level Traced<ConcreteArray(2.0)>with<JVPTrace(level=4/0)>
```

If we do something similar with `jit`, we might get a _sublevel_ error:

```python
from jax import jit

def f(x):
  y = None
  def g(x):
    nonlocal y
    y = x
    return x
  return jit(g)(x) + y
jit(f)(1.)
```

```
jax.core.UnexpectedTracerError: Encountered an unexpected tracer. [...]
Detail: Can't lift sublevels 2 to 1.
```

Or this one using both `jit` and `grad`:

```python
from jax import jit, grad

def f(x):
  y = None
  def g(x):
    nonlocal y
    y = x
    return x
  return jit(g)(x) + y

grad(f)(1.)
```

```
jax.core.UnexpectedTracerError: Encountered an unexpected tracer. [...]
Detail: Can't lift sublevels 1 to 0.
```

See [these api tests](https://github.com/google/jax/blob/4fcfaeb8a9ce55098269a40bfb45565db71bf344/tests/api_test.py#L2109-L2156) for more examples of exercising these error paths.

We increment the sublevel when we run the callable argument to a final-style higher-order primitive, and we decrement it again when the callable exits. That ensures that the primitive binds we encountered within the callable would see an incremented sublevel, and binds that happened after the callable finishes would be back at the previous sublevel. We do that with the context manager `core.new_sublevel`. There are a few different places where we'd run the callable: for example, when we evaluate the application of a call primitive, i.e. in `call_impl`, and also when we stage out the application, i.e. in `DynamicJaxprTrace.process_call`.

Before this PR, we actually used a helper function `core.maybe_new_sublevel`. That helper function checked whether the trace for which we were about to call `process_call` was the 'dynamic' trace (introduced in #3370, see also [Autodidax](https://jax.readthedocs.io/en/latest/autodidax.html)), leveraging the rule that a trace would run the call primitive's callable argument if and only if it is the 'dynamic' trace (which represents the bottom of the interpreter stack).

But we can't use `maybe_new_sublevel` anymore with this inlining logic! The reason is that we need `DynamicJaxprTrace.process_call` to run the jaxpr to be inlined, i.e. to call `eval_jaxpr` and thus run primitive binds, but running those operations should _not_ happen with an incremented sublevel. That is, we need `DynamicJaxprTrace.process_call` to manage the sublevel itself, and be able to run binds at two different sublevels (an incremeneted one while it traces the Python callable, and a non-incremented one when it inlines the jaxpr). (I'd wanted to make this change earlier, for some other reason, which I can't remember.)

So that's why this PR includes a refactor which removes `maybe_new_sublevel` and instead puts explicit `new_sublevel` calls in several places.

Now that we have that refactoring, why do we have to bother forming the jaxpr inside `DynamicJaxprTrace.process_call`, rather than just calling `fun.call_wrapped(*tracers)` directly? It has to do with the assumptions that `trace_to_subjaxpr_dynamic` makes about whether the sublevel has been incremented. That is, we've already set up the Python callable (in a `WrappedFun`) to expect the sublevel to be incremented; we can't run it without raising the sublevel, and yet we can't inline it with an incremented sublevel. We could probably perform a larger refactor to avoid the need to form the jaxpr, but I decided it was better to keep this change relatively small.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -263,7 +263,10 @@ def jit(
       buffers to reduce the amount of memory needed to perform a computation,
       for example recycling one of your input buffers to store a result. You
       should not reuse buffers that you donate to a computation, JAX will raise
-      an error if you try to.
+      an error if you try to. By default, no arguments are donated.
+    inline: Specify whether this function should be inlined into enclosing
+      jaxprs (rather than being represented as an application of the xla_call
+      primitive with its own subjaxpr). Default False.
 
   Returns:
     A wrapped version of ``fun``, set up for just-in-time compilation.
@@ -304,7 +307,7 @@ def _python_jit(
     donate_argnums: Union[int, Iterable[int]] = (),
     inline: bool = False,
 ) -> F:
-  """The Python implementation of `jax.jit`, being slowly replaced by _cpp_jit."""
+  # The Python implementation of `jax.jit`, being slowly replaced by _cpp_jit.
   _check_callable(fun)
   static_argnums, static_argnames = _infer_argnums_and_argnames(
       fun, static_argnums, static_argnames)
@@ -365,15 +368,13 @@ def _cpp_jit(
     donate_argnums: Union[int, Iterable[int]] = (),
     inline: bool = False,
 ) -> F:
-  """An implementation of `jit` that tries to do as much as possible in C++.
-
-  The goal of this function is to speed up the time it takes to process the
-  arguments, find the correct C++ executable, start the transfer of arguments
-  and schedule the computation.
-  As long as it does not support all features of the Python implementation
-  the C++ code will fallback to `_python_jit` when it faces some unsupported
-  feature.
-  """
+  # An implementation of `jit` that tries to do as much as possible in C++.
+  # The goal of this function is to speed up the time it takes to process the
+  # arguments, find the correct C++ executable, start the transfer of arguments
+  # and schedule the computation.
+  # As long as it does not support all features of the Python implementation
+  # the C++ code will fallback to `_python_jit` when it faces some unsupported
+  # feature.
   _check_callable(fun)
   static_argnums, static_argnames = _infer_argnums_and_argnames(
       fun, static_argnums, static_argnames)
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -212,6 +212,7 @@ def jit(
   device: Optional[xc.Device] = None,
   backend: Optional[str] = None,
   donate_argnums: Union[int, Iterable[int]] = (),
+  inline: bool = False,
 ) -> F:
   """Sets up ``fun`` for just-in-time compilation with XLA.
 
@@ -288,10 +289,10 @@ def jit(
     static_argnames = ()
   if FLAGS.experimental_cpp_jit:
     return _cpp_jit(fun, static_argnums, static_argnames, device, backend,
-                    donate_argnums)
+                    donate_argnums, inline)
   else:
     return _python_jit(fun, static_argnums, static_argnames, device, backend,
-                       donate_argnums)
+                       donate_argnums, inline)
 
 
 def _python_jit(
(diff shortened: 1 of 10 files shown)


jax/_src/api.py
```diff
@@ -212,6 +212,7 @@ def jit(
   device: Optional[xc.Device] = None,
   backend: Optional[str] = None,
   donate_argnums: Union[int, Iterable[int]] = (),
+  inline: bool = False,
 ) -> F:
   """Sets up ``fun`` for just-in-time compilation with XLA.
```
hawkinsp, line 217: I'd document inline.

hawkinsp (approved):

jax/_src/api.py
mattjj: Do you like it as a public API? (Seems pretty reasonable to me! But I wanted to double check.)

hawkinsp: I think so.

The only caveat is at some point in the future we might also want users to provide a hint to, say, XLA on whether it should inline functions. Perhaps that could be solved by `inline="xla"` or something at that time.

mattjj: For that case I propose an `inline2` option.

tomweingarten:
Thanks for this excellent writeup! Is there any performance implication to using inline? Are there cases where using inline will allow the compiler to optimize further? This might be useful to affirm or deny in the documentation.

PR #6593, opened 2021-04-29
Implement jnp.r_ and jnp.c_

nicholasjng:
This PR contains the `jax.numpy` equivalents of `np.r_` and `np.c_`. This is my first attempt at a contribution to JAX, and these two routines were the ones left open in issue #5850, so I decided to take them on. I added some tests as well.

I am not sure if I handled the dtype promotion correctly, and whether I should have dtype-checked in the tests as well in this case. Stylistically, I based my implementation here on that of `jnp.ogrid` (or rather, the private base class of it) from the same issue as above. I would love to get some feedback on this and improve it so that it can be merged into Jax in the end.


jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+  def _maybe_process_directive(self, key):
+    """Adjust parameters when detecting a directive."""
+    directive = key[0]
+    # two special cases: matrix directives
+    mats = ("r", "c")
+    matrix = -1 if str(directive) not in mats else mats.index(directive)
+    params = [self.axis, self.ndmin, self.trans1d, matrix]
+
+    if not isinstance(directive, str) or matrix != -1:
+      return params
```
jakevdp, line 3349: This seems strange that we'd accept any possible value here if the directive is not a string. Shouldn't we raise an error if it's not an expected value?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+        k = len(vec)
+        if k < 4:
+          vec += params[k:]
+        else:
+          vec = vec[:3] + [matrix]
+        return list(map(int, vec))
+      except ValueError:
+        raise ValueError(
+          "could not understand directive {!r}".format(directive)
+        )
```
jakevdp, line 3362: It looks like this `ValueError` catch is meant to apply to `list(map(int, vec))`; if so I'd put it just around that statement. Also a useful pattern here for more informative tracebacks is
```python
except ValueError as err:
  raise ValueError(msg) from err
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+
+  def _axes_shift(self, ndmin, trans1d):
+    """Shifts beginning of 1D axes if necessary."""
+    shape_obj = list(range(ndmin))
+    # Calculate number of shifts, with overflow protection by mod
+    num_rshifts = abs(ndmin + trans1d + 1) % ndmin
```
jakevdp, line 3368: `abs` in this context is `jnp.abs`. I think you probably want `_abs`, which is the local alias for the built-in `abs`.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+    for i in range(num_rshifts):
+      shape_obj.insert(0, shape_obj.pop())
+    return tuple(shape_obj)
```
jakevdp, line 3371: Maybe more succintly:
```
return tuple(shape_obj[num_rshifts - 1:] + shape_obj[:num_rshifts - 1])
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+
+  def __getitem__(self, key):
+    if not isinstance(key, tuple):
+      key = (key,)
+
+    axis, ndmin, trans1d, matrix = self._maybe_process_directive(key)
+
+    # cut out the directive if there is one
+    if isinstance(key[0], str):
+      key = key[1:]
```
jakevdp, line 3381: Ah, I see, the "directive is not a string" case I asked about above is handled here.

I think it would be better to put all of that logic in one place, so that someone reading the code could more easily understand the logic of how `key` is handled. Could you put the logic from `_maybe_process_directive` inline here, under `if isinstance(key[0], str)`?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+      if trans1d != -1 and ndmin - ndim(item) > 0:
+        newobj = transpose(newobj, self._axes_shift(ndmin, trans1d))
+
+      output.append(newobj)
+
+    # Final dtype promotion
+    res_dtypes = [dtypes.dtype(arr) for arr in output]
+    final_dtype = jax.dtypes.result_type(*res_dtypes)
+    for arr in output:
+      arr = arr.astype(final_dtype)
```
jakevdp, line 3408: I think `output = _promote_dtypes(*output)` can replace these four lines.

But actually, I think `concatenate` below calls `_promote_dtypes` on the input, so it may not even be necessary.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+    return res
+
+  def __len__(self):
+    return 0
+
+
+class RClass(_AxisConcat):
+  """Concatenate slices, scalars and array-like objects along the first axis.
+
+  LAX-backend implementation of :obj:`numpy.r_`."""
```
jakevdp, line 3425: It would be useful to add a few examples to this docstring.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,115 @@ class _Ogrid(_IndexGrid):
...
+  trans1d = -1
+
+
+r_ = RClass()
+
+
+class CClass(_AxisConcat):
+  """Concatenate slices, scalars and array-like objects along the last axis.
+
+  LAX-backend implementation of :obj:`numpy.c_`."""
```
jakevdp, line 3437: Same here - add some examples to the docstring.

tests/lax_numpy_test.py
```diff
@@ -4709,6 +4709,80 @@ def assertListOfArraysEqual(xs, ys):
...
+    self.assertArraysEqual(np.r_['0,2', [1,2,3], [4,5,6]], jnp.r_['0,2', [1,2,3], [4,5,6]])
+    self.assertArraysEqual(np.r_['0,2,0', [1,2,3], [4,5,6]], jnp.r_['0,2,0', [1,2,3], [4,5,6]])
+    self.assertArraysEqual(np.r_['1,2,0', [1,2,3], [4,5,6]], jnp.r_['1,2,0', [1,2,3], [4,5,6]])
+    # negative 1d axis start
+    self.assertArraysEqual(np.r_['0,4,-1', [1,2,3], [4,5,6]], jnp.r_['0,4,-1', [1,2,3], [4,5,6]])
+    self.assertArraysEqual(np.r_['0,4,-2', [1,2,3], [4,5,6]], jnp.r_['0,4,-2', [1,2,3], [4,5,6]])
+
+    # matrix directives
+    with warnings.catch_warnings():
+      warnings.filterwarnings("ignore", category=PendingDeprecationWarning)
```
jakevdp, line 4726: What warning does this catch? If numpy is going to be deprecating something, I think we should avoid adding that deprecated API to JAX.

Same below.

tests/lax_numpy_test.py
```diff
@@ -4709,6 +4709,80 @@ def assertListOfArraysEqual(xs, ys):
...
+      self.assertArraysEqual(np.r_['r',[1,2,3], [4,5,6]], jnp.r_['r',[1,2,3], [4,5,6]])
+      self.assertArraysEqual(np.r_['c', [1, 2, 3], [4, 5, 6]], jnp.r_['c', [1, 2, 3], [4, 5, 6]])
+
+    # bad directive
+    self.assertRaises(ValueError, lambda: jnp.r_["asdfgh",[1,2,3]])
```
jakevdp, line 4731: I'd lean toward using something like
```python
with self.assertRaisesRegex(ValueError, "could not understand directive.*"):
  jnp.r_["asdfgh",[1,2,3]]
```
to ensure the expected error message applies.

Same below.

jakevdp (changes requested):
Thanks for taking a look at this! I have a number of comments below; the main thing is that the splitting of the logic for handling inputs between different methods is a bit confusing: I think the code would be easier to follow if it were all in one place.

One more thing: to ensure these appear in the documentation, you should add `c_` and `r_` to the list in https://github.com/google/jax/blob/master/docs/jax.numpy.rst

njunge94:
Thank you for all of your comments @jakevdp! I'll start tackling them tomorrow with a fresh mind.

Regarding the `PendingDeprecationWarning`: This was raised in my local test session when calling e.g. `np.r_["r",...]`, causing the tests to fail.
`jnp.r_/c_` returns a JAX array, but I still felt it was necessary to test equality for the two special directive options `"r"/"c"` explicitly, and for those, `numpy.r_/c_` returns a `numpy.matrix` instead of an `array`, as per the [numpy.r_ docs](https://numpy.org/doc/stable/reference/generated/numpy.r_.html). This is a deprecated class in numpy; quoting [this StackOverflow answer](https://stackoverflow.com/questions/53254738/deprecation-status-of-the-numpy-matrix-class):

> And at the same time a `PendingDeprecationWarning` has been added to `matrix.__new__`. 

Is there a better way to go about this? I tried to use the `jtu.ignore_warning` decorator, but failed (I think because the class has no `__call__` operator).

I have another question: The `isinstance(item, slice)` portion in the main if-else branch is basically a duplicate of the code in `_IndexGrid.__getitem__`. I did not want to reuse that though, because of the nested for loop. Do you think it makes sense to refactor this part into its own function and use it in both places?

And lastly: Should I address each of your points in separate commits, or should I just put everything into a single commit and iterate the review procedure?

jakevdp:
> Thank you for all of your comments @jakevdp! I'll start tackling them tomorrow with a fresh mind.

Great! I'll look forward to seeing it

> Regarding the PendingDeprecationWarning...

Thanks for the clarification. In that case, I think this is fine. I didn't see any indication as to whether numpy will remove this option when matrices go away, or just return arrays instead. I suspect it's the latter.

> I have another question: The isinstance(item, slice) portion in the main if-else branch is basically a duplicate of the code in _IndexGrid.__getitem__.

Moving that to a common utility would be a good approach.

> And lastly: Should I address each of your points in separate commits, or should I just put everything into a single commit and iterate the review procedure?

Do whatever is most convenient for you; once everything is finished I'll be asking you to [squash all the changes](https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) into a single commit, but that can easily done retroactively.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,207 @@ class _Ogrid(_IndexGrid):
...
+      key = (key,)
+
+    params = [self.axis, self.ndmin, self.trans1d, -1]
+
+    if isinstance(key[0], str):
+      directive = key[0]
+      # check two special cases: matrix directives
+      mats = ("r", "c")
+      if directive in mats:
+        params[-1] = mats.index(directive)
```
jakevdp, line 3362: Slight preference here for the more direct
```
if directive == "r":
  params[-1] = 0
elif directive == "c":
  params[-1] = 1
else:
  ...
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -3334,6 +3334,207 @@ class _Ogrid(_IndexGrid):
...
+      else:
+        vec = directive.split(",")
```
jakevdp, line 3364: What about when `directive` is an empty string?

jakevdp (changes requested):
Looks good.

I'm still a bit unsure about corner cases and handling of unexpected inputs.

jax/_src/numpy/lax_numpy.py
njunge94: Alright! Before I push this, do you think that the 

```
      # cut out the directive
      key = key[1:]
```
part at the very end of that if-else branch should move up top (just below the `directive` assignment) to improve legibility/flow?

jakevdp: Sure, that's a good idea. You could even use something like this to be more concise:
```
directive, *key = key
```

jax/_src/numpy/lax_numpy.py
njunge94: The empty string specifically ends in the ValueError at the `int` conversion attempt, with the correct traceback:
```
  File "/Users/nicholasjunge/workspaces/python_projects/jax/jax/_src/numpy/lax_numpy.py", line 3372, in __getitem__
    try:
ValueError: invalid literal for int() with base 10: ''
  File "<input>", line 1, in <module>
  File "/Users/nicholasjunge/workspaces/python_projects/jax/jax/_src/numpy/lax_numpy.py", line 3374, in __getitem__
    except ValueError as err:
ValueError: could not understand directive ''
```

I definitely concede that this (relatively opaque) integer-literal conversion is doing very heavy lifting here. Are there any unexpected int-convertible literals that can screw things up in this case? I came up empty-handed on my search attempts on that so far.

The other option I could think of is to convert this into an `elif` with a suitable condition and then make an `else` branch which just raises a ValueError indicating a bad directive. What do you think?

jakevdp:
Looks good! A couple more things:

- looks like the documentation is failing because of the `jnp.r_` and `jnp.c_` under `See Also:`. You can try just writing `r_` and `c_` instead.
- There are conflicts that must be resolve by merging or rebasing on the current master
- Before we can finish, we need all the changes squashed into a single commit.

Let me know if you need advice on any of that!

njunge94:
> Looks good! A couple more things:
> 
>     * looks like the documentation is failing because of the `jnp.r_` and `jnp.c_` under `See Also:`. You can try just writing `r_` and `c_` instead.
> 
>     * There are conflicts that must be resolve by merging or rebasing on the current master
> 
>     * Before we can finish, we need all the changes squashed into a single commit.
> 
> 
> Let me know if you need advice on any of that!

Thank you for offering advice, I really appreciate it! So far I have only contributed PRs in projects that did not require rebasing/squashing. I have of course read the contributing section and a few SO threads / git tutorials on rebasing, but I think it will take an example for me to really get the hang of it. So if you could help me a bit with steps 2 and 3, I would really appreciate it.

Regarding the documentation: I actually just went and reused the template I found in the `_Ogrid` class directly above, which also specifies See Also items with the prefix `jnp`. Does it work there because the private class is not picked up by the documentation?

jakevdp:
No problem; this is how I generally do it (note that I usually have `upstream` point to the main repository, and `origin` point to my own fork).

First, merge in new changes and fix conflicts if there are any:
```
$ git remote -v
origin	git@github.com:jakevdp/jax.git (fetch)
origin	git@github.com:jakevdp/jax.git (push)
upstream	git@github.com:google/jax.git (fetch)
upstream	git@github.com:google/jax.git (push)

$ git checkout master
$ git pull upstream master
$ git checkout axis_concatenation
$ git merge master  # may result in conflicts that you'll have to manually fix
```
There's some info on resolving merge conflicts here: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-using-the-command-line

Now we're going to squash everything into a single commit:
```
$ git reset --soft master  # rewinds the commit to master; soft means that your changes will still be in the working directory
$ git commit -m "message for my single commit"
$ git push -f origin axis_concatenation  # Force push to change the commits on the branch linked to this PR
```
Hopefully that helps!

jakevdp:
Regarding the doc failure: it looks like the issue is that `jnp.r_` is being parsed by RST, where a trailing underscore indicates a link, and the link `jnp.r` does not exist.

You could probably put the function name within double back-ticks to get around this:
```
``jnp.r_``: Concatenates slices, scalars and array-like objects along the first axis.
```

njunge94:
Nice! I succeeded in building the JAX docs locally as described [here](https://jax.readthedocs.io/en/latest/developer.html#update-documentation) with the double backticks that you suggested. So I will push that and then give rebasing a go the way you outlined. 
Also, thank you for writing up your workflow there, it's really awesome that you went out of your way to help, and it helps me improve for the next time as well!

jakevdp (approved):
Looks great!

jakevdp:
It looks like there are some doctest failures for the new docstrings. You can run the relevant tests locally with
```
pytest --doctest-modules jax/_src/numpy/lax_numpy.py
```

njunge94:
Apologies! Looks like two things are happening at the same time: 
1. A stupid missing closing square bracket that my Pycharm did catch, that one is easy
2. My "comments" on output shapes are being stripped away from the examples. Removing them makes the tests succeed. Should I take them out then?

jakevdp:
Yes, I'd say take the comments out to make the doctests pass

jakevdp:
Thanks, looks great! We'll get this merged as soon as tests pass :grin:

PR #6596, opened 2021-04-29
Remove docs fallback for numpy < 1.16

lgeiger:
This PR removes docs wrapper fallbacks for functions that where added in numpy 1.15.
Jax already requires numpy 1.16 or newer which removes the need for these fallbacks.


jakevdp (approved):
Looks good, Thanks!

PR #6597, opened 2021-04-29
Export jnp.broadcast_shapes as user facing function

lgeiger:
Since numpy 1.20 [`np.broadcast_shapes`](https://numpy.org/doc/stable/reference/generated/numpy.broadcast_shapes.html) is part of the public API. This PR adds an alias to `lax.broadcast_shapes` it to `jax.numpy` in order to more closely match numpy.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1733,6 +1733,12 @@ def bincount(x, weights=None, minlength=0, *, length=None):
     raise ValueError("shape of weights must match shape of x.")
   return zeros(length, _dtype(weights)).at[clip(x, 0)].add(weights)
 
+@_wraps(getattr(np, "broadcast_shapes", None))
+def broadcast_shapes(*shapes):
+  if not shapes:
+    return ()
+  shapes = [(shape,) if np.ndim(shape) == 0 else tuple(shape) for shape in shapes]
+  return lax.broadcast_shapes(*shapes)
 
 def broadcast_arrays(*args):
   """Like Numpy's broadcast_arrays but doesn't return views."""
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -87,6 +87,7 @@ Not every function in NumPy is implemented; contributions are welcome!
     block
     bool_
     broadcast_arrays
+    broadcast_shapes
     broadcast_to
     can_cast
     cbrt
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -25,7 +25,7 @@ from jax._src.numpy.lax_numpy import (
     arcsinh, arctan, arctan2, arctanh, argmax, argmin, argsort, argwhere, around,
     array, array_equal, array_equiv, array_repr, array_split, array_str, asarray, atleast_1d, atleast_2d,
     atleast_3d, average, bartlett, bfloat16, bincount, bitwise_and, bitwise_not,
-    bitwise_or, bitwise_xor, blackman, block, bool_, broadcast_arrays,
+    bitwise_or, bitwise_xor, blackman, block, bool_, broadcast_arrays, broadcast_shapes,
     broadcast_to, can_cast, cbrt, cdouble, ceil, character, choose, clip, column_stack,
     complex128, complex64, complex_, complexfloating, compress, concatenate,
     conj, conjugate, convolve, copysign, corrcoef, correlate, cos, cosh,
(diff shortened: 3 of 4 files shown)


jakevdp (approved):
Good call - thanks for catching this!

jax/_src/numpy/lax_numpy.py
```diff
@@ -1733,6 +1733,7 @@ def bincount(x, weights=None, minlength=0, *, length=None):
     raise ValueError("shape of weights must match shape of x.")
   return zeros(length, _dtype(weights)).at[clip(x, 0)].add(weights)
 
+broadcast_shapes = _wraps(getattr(np, "broadcast_shapes", None))(lax.broadcast_shapes)
```
jakevdp, line 1736: I think we'll need a bit more of a wrapper than this to match numpy's behavior: `lax.broadcast_shapes` does not accept integers, while numpy's does:
```python
In [1]: import numpy as np                                                                                                                              

In [2]: np.broadcast_shapes(3, (2, 1))                                                                                                                  
Out[2]: (2, 3)

In [3]: from jax import lax                                                                                                                             

In [4]: lax.broadcast_shapes(3, (2, 1))                                                                                                                 
<ipython-input-4-0981db21d119> in <module>
----> 1 lax.broadcast_shapes(3, (2, 1))

~/github/google/jax/jax/_src/util.py in wrapper(*args, **kwargs)
    184         return f(*args, **kwargs)
    185       else:
--> 186         return cached(config._trace_context(), *args, **kwargs)
    187 
    188     wrapper.cache_clear = cached.cache_clear

~/github/google/jax/jax/_src/util.py in cached(_, *args, **kwargs)
    177     @functools.lru_cache(max_size)
    178     def cached(_, *args, **kwargs):
--> 179       return f(*args, **kwargs)
    180 
    181     @functools.wraps(f)

~/github/google/jax/jax/_src/lax/lax.py in broadcast_shapes(*shapes)
     86   if len(shapes) == 1:
     87     return shapes[0]
---> 88   ndim = _max(len(shape) for shape in shapes)
     89   shapes = [(1,) * (ndim - len(shape)) + shape for shape in shapes]
     90   result_shape = _try_broadcast_shapes(shapes)

~/github/google/jax/jax/_src/lax/lax.py in <genexpr>(.0)
     86   if len(shapes) == 1:
     87     return shapes[0]
---> 88   ndim = _max(len(shape) for shape in shapes)
     89   shapes = [(1,) * (ndim - len(shape)) + shape for shape in shapes]
     90   result_shape = _try_broadcast_shapes(shapes)

TypeError: object of type 'int' has no len()
```

jakevdp (changes requested):
One quick comment below

jax/_src/numpy/lax_numpy.py
lgeiger: Good catch! I forgot about that.
I updated the PR and adapted `lax.broadcast_shapes` to handle integers, lists and numpy arrays as shapes. Let me know if you prefer to move this handling to the numpy function instead.

jax/_src/lax/lax.py
```diff
@@ -81,13 +81,19 @@ def _try_broadcast_shapes(shapes):
...
 def broadcast_shapes(*shapes):
   """Returns the shape that results from NumPy broadcasting of `shapes`."""
+  if not shapes:
+    return ()
   if len(shapes) == 1:
     return shapes[0]
-  ndim = _max(len(shape) for shape in shapes)
-  shapes = [(1,) * (ndim - len(shape)) + shape for shape in shapes]
-  result_shape = _try_broadcast_shapes(shapes)
+  shapes = [(shape,) if np.ndim(shape) == 0 else tuple(shape) for shape in shapes]
```
lgeiger, line 95: We need to normalize shapes to tuples, in order to support integers as well as lists and 1D numpy arrays which would otherwise break the caching decorator since they are not hashable.

jakevdp:
Thanks for the update! I think it's best to leave the lax function alone and put the broadened input handling in the `jax.numpy` wrapper. The `lax` interface is designed deliberately to be strict about these kinds of things (most functions fail for non-iterable shapes), whereas `jax.numpy` follows the looser convensions of `numpy` itself. As an example of this, take a look at the handling of `shape` in `lax.full` vs `jnp.full`.

lgeiger:
Sounds good. I moved the handling to the numpy function itself.

jakevdp (approved):

jakevdp:
Thanks for the PR!

PR #6604, opened 2021-04-30
Add `jnp.real_if_close`

lgeiger:
This PR adds `jnp.real_if_close` to match [`np.real_if_close`](https://numpy.org/doc/stable/reference/generated/numpy.real_if_close).

### DIFF
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -303,6 +303,7 @@ Not every function in NumPy is implemented; contributions are welcome!
     ravel
     ravel_multi_index
     real
+    real_if_close
     reciprocal
     remainder
     repeat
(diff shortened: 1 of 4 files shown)


lgeiger:
That is a fair point. I don't see any way of easily solving this since XLA requires matching dtypes in control flow.

> I think the only option then would be to unconditionally return complex values, but zero-out the imaginary part if it's small. That departs from numpy's behavior (and arguably from the intent of the function), but it would then be compatible with JAX transforms. What do you think?

I think this would be a sensible behaviour. But ultimately I agree that the function with this behaviour would not be very useful in practice.

So I think the best way would probably to throw in cases where the functions is jited, or to just keep it unimplemented. Thinking about this again, I'd personally be in favour of closing this PR unless someone has a strong use case for this function.

jakevdp:
Makes sense - thanks for looking into this!

PR #6608, opened 2021-05-01
Add associated Legendre functions of the first kind

tlu7:
Uses three recurrence relations to compute normalized associated Legendre functions.  The maximum order and maximum degree are the same.

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -18,12 +18,16 @@ import numpy as np
 import scipy.special as osp_special
 
 from jax._src import api
+from jax import jit
 from jax import lax, core
+from jax import ops
 from jax.interpreters import ad
 from jax._src.numpy import lax_numpy as jnp
 from jax._src.numpy.lax_numpy import (asarray, _reduction_dims, _constant_like,
                                       _promote_args_inexact)
 from jax._src.numpy.util import _wraps
+
+from typing import Tuple
 
 
 @_wraps(osp_special.gammaln)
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/special.py
```diff
@@ -657,3 +659,92 @@ def i1e(x):
...
+    raise ValueError('x must be a 1D array.')
+
+  if m != n:
+    raise NotImplementedError('Computations for m!=n are not yet supported.')
+
+  l_max = n
+  p = np.zeros((l_max + 1, l_max + 1, z.shape[0]))
+
+  # The initial value p(0,0).
+  initial_value = 0.5 / np.sqrt(math.pi)
```
jakevdp, line 724: nit: use `np.pi` to avoid additional import

jax/_src/scipy/special.py
```diff
@@ -657,3 +659,92 @@ def i1e(x):
...
+  """
+  dtype = lax.dtype(z)
+  if dtype not in (jnp.float32, jnp.float64):
+    raise TypeError(
+        'x.dtype={} is not supported, see docstring for supported types.'
+        .format(dtype))
+
+  if z.ndim != 1:
+    raise ValueError('x must be a 1D array.')
+
```
jakevdp, line 716: Before comparing `m != n`, we should call `core.concrete_or_error()` on each value (it leads to better error messages in the case that `m` or `n` are traced).

jax/_src/scipy/special.py
```diff
@@ -657,3 +659,92 @@ def i1e(x):
...
+def lpmn(m, n, z):
+  r"""The normalized associated Legendre functions of the first kind.
+
+  In the descriptions of the associated Legendre functions (ALFs), we use `l`
+  to denote the degree (`l_max` is the maximum degree) and `m` to denote
+  the order. The ALFs of the first kind are often used in spherical harmonics.
+  The spherical harmonic of degree `l` and order `m` can be written as
+  .. math::
+    \begin{equation}
```
jakevdp, line 671: I don't think the `\begin{equation}` blocks are necessary here & below

jax/_src/scipy/special.py
```diff
@@ -657,3 +659,92 @@ def i1e(x):
...
+
+  # The initial value p(0,0).
+  initial_value = 0.5 / np.sqrt(math.pi)
+  p[0, 0] = initial_value
+
+  # Compute the diagonal entries p(l,l) with recurrence.
+  y = np.sqrt(1.0 - z * z)
+  for l_idx in range(1, l_max + 1):
+    a = -1.0 * np.sqrt(1.0 + 0.5 / l_idx)
+    p[l_idx, l_idx] = a * y * p[l_idx - 1, l_idx - 1]
```
jakevdp, line 731: Something is strange here... `p` is a numpy array, but this statement populates it with a potentially traced value (`y` derives from `z`). I don't think this will work correctly with JAX transforms.

tests/lax_scipy_test.py
```diff
@@ -237,6 +238,45 @@ def testXlog1pyShouldReturnZero(self):
...
+          c1 = (4.0 * math.pi) * osp_special.factorial(l + m)
+          c2 = np.sqrt(c0 / c1)
+          a_normalized[m, l] = c2 * a[m, l]
+      return a_normalized
+
+    # The results from scipy are not normalized and the comparison requires
+    # normalizing the results.
+    expected_p_normalized = _apply_normalization(expected_p)
+
+    self.assertAllClose(actual_p, expected_p_normalized, rtol=7e-14, atol=4e-15)
```
jakevdp, line 279: I'd recommend using `self._CheckAgainstNumpy` and `self._CompileAndCheck`, which will reduce boilerplate and ensure that the function behaves the way we want it to.

jakevdp (changes requested):
Looks good! A few comments below.

One larger question: from the test, it looks like this does not return the same values as the scipy function, but rather returns a normalized version. Our aim is for functions in `jax.scipy.special` to be JAX-compatible drop-in replacements for their counterparts in `scipy.special`. Can we adjust the implementation to return the same results as `scipy`?

jax/_src/scipy/special.py
tlu7: Done.

jax/_src/scipy/special.py
tlu7: Done. Could you explain the concept of trace? Or any reference/document will help me understand. Thanks!

jax/_src/scipy/special.py
tlu7: Removed the \begin{equation} blocks here and after.

jax/_src/scipy/special.py
jakevdp: This doc might be useful: https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html

jax/_src/scipy/special.py
shoyer: I have a similar concern to @jakevdp. If this function is returning a NumPy array, why would a user use it rather than the SciPy version?

jax/_src/scipy/special.py
```diff
@@ -657,3 +659,92 @@ def i1e(x):
...
+    for l_idx in range(m_idx + 2, l_max + 1):
+      c0 = l_idx * l_idx
+      c1 = m_idx * m_idx
+      c2 = 2.0 * l_idx
+      c3 = (l_idx - 1.0) * (l_idx - 1.0)
+      d0 = np.sqrt((4.0 * c0 - 1.0) / (c0 - c1))
+      d1 = np.sqrt(((c2 + 1.0) * (c3 - c1)) / ((c2 - 3.0) * (c0 - c1)))
+      p[m_idx, l_idx] = d0 * z * p[m_idx, l_idx - 1] - d1 * p[m_idx, l_idx - 2]
+
+  return p
```
shoyer, line 750: The return value here is different from SciPy, which returns two arrays.

jax/_src/scipy/special.py
tlu7: Thanks for the comments! The proposed function has two advantages over the scipy version (1) performance and (2) numerical stability. 

The following numbers come from numerical experiments
(1) performance: 

Maximum degree | Number of spatial samplings | proposed (seconds) | scipy (seconds) | speed-up
                       128 |                                         1024 |                          0.164 |                   0.619 |         3.78x
                      256 |                                          1024 |                           0.57 |                     2.53 |         4.41x
                       512 |                                         1024 |                            2.33 |                   10.91 |         4.68x
                    1024 |                                          1024 |                          10.18 |                  48.39 |          4.75x

(2) numerical stability
setting maximum degree to 512 and number of sampling points to 1024

order = 128, degree = 256, sampling point index = 1024

proposed = mode 0.00047938111886318223
scipy = 5.430959668839146e+301

tlu7:
> Looks good! A few comments below.
> 
> One larger question: from the test, it looks like this does not return the same values as the scipy function, but rather returns a normalized version. Our aim is for functions in `jax.scipy.special` to be JAX-compatible drop-in replacements for their counterparts in `scipy.special`. Can we adjust the implementation to return the same results as `scipy`?

Thanks for the explanations! 

The normalization is to avoid overeflow/underflow and achieve better numerical stability. This is especially helpful considering the large-scale problems running on and the precision offered by accelerators. 

If you do not feel strongly, I would like to keep the implementation. Results of the numerical experiments are also attached inline with the comments.

jakevdp:
Thanks - I think it's important that we match the API of the scipy routine. One way we could move forward is to add a new keyword argument, e.g. `normalize=False`, such that the default would return the same result as the scipy function, but you could set it to `True` to return the more stable unnormalized version that you're interested in. What do you think?

tlu7:
> One larger question: from the test, it looks like this does not return the same values as the scipy function, but rather returns a normalized version. Our aim is for functions in `jax.scipy.special` to be JAX-compatible drop-in replacements for their counterparts in `scipy.special`. Can we adjust the implementation to return the same results as `scipy`?

Thanks for the comment! Normalization avoids overflow/underflow and achieves better numerical stability. For example, 


> Thanks - I think it's important that we match the API of the scipy routine. One way we could move forward is to add a new keyword argument, e.g. `normalize=False`, such that the default would return the same result as the scipy function, but you could set it to `True` to return the more stable unnormalized version that you're interested in. What do you think?

Thanks for the suggestion. Following the idea, 

> Thanks - I think it's important that we match the API of the scipy routine. One way we could move forward is to add a new keyword argument, e.g. `normalize=False`, such that the default would return the same result as the scipy function, but you could set it to `True` to return the more stable unnormalized version that you're interested in. What do you think?

Thanks for the suggestion! Following your idea about the new keyword argument, I actually added a check on the input `z`. If `z` is a scalar, use scipy.special.lpmn; if `z` is an array, branch to the proposed implementation.  I think through this way it matches the API of the scipy routine. Any suggestions?

jakevdp:
> If z is a scalar, use `scipy.special.lpmn`

This is not really a viable option for JAX. We need to implement JAX functions in terms of JAX operations, so that the resulting function will be compatible with transforms like `jit`, `grad`, `vmap`, etc.

shoyer:
> This is not really a viable option for JAX. We need to implement JAX functions in terms of JAX operations, so that the resulting function will be compatible with transforms like `jit`, `grad`, `vmap`, etc.

This comes back to my other concern for this PR, which currently does not seem to be compatible with `jit`, grad`, `vmap` etc and it's unclear that it ever could be. To me, this suggests that the improved numerical methods would probably fit better in a library such as SciPy or a stand-alone project.

tlu7 (commented):
> > This is not really a viable option for JAX. We need to implement JAX functions in terms of JAX operations, so that the resulting function will be compatible with transforms like `jit`, `grad`, `vmap`, etc.
> 
> This comes back to my other concern for this PR, which currently does not seem to be compatible with `jit`, grad`, `vmap` etc and it's unclear that it ever could be. To me, this suggests that the improved numerical methods would probably fit better in a library such as SciPy or a stand-alone project.

Thank you both for the comments and the offline discussions! 

The following four items are suggested by @jakevdp as the first step to make this PR compatible with JAX
1. Use jnp.numpy
2. Return (values, derivatives) in same PR and comply with scipy API
3. Replace for-loops with
    3a.Optimally, vectorized codes, e.g. cumulative sums
    3b.Or Lax control flow, jax._src.lax.control_flow
4. Use index update

I will keep you updated on the progress.

shoyer:
> 3\. Replace for-loops with
>     3a.Optimally, vectorized codes, e.g. cumulative sums
>     3b.Or Lax control flow, jax._src.lax.control_flow
> 4\. Use index update

The only argument that could support non-static values would be `z` -- would vectorizing or doing auto-diff through this argument be useful to you? If not, then I would suggest not going to the trouble here and considering putting the code elsewhere.

tlu7:
> > 3. Replace for-loops with
> > 3a.Optimally, vectorized codes, e.g. cumulative sums
> > 3b.Or Lax control flow, jax._src.lax.control_flow
> > 4. Use index update
> 
> The only argument that could support non-static values would be `z` -- would vectorizing or doing auto-diff through this argument be useful to you? If not, then I would suggest not going to the trouble here and considering putting the code elsewhere.

Yes, vectorization reduces the size of the graph, which solves the scaling issue. Otherwise, one has to stamp a large node of constant on a graph.

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+  m_mat, l_mat = jnp.mgrid[:l_max + 1, :l_max + 1]
+  if is_normalized:
+    c0 = l_mat * l_mat
+    c1 = m_mat * m_mat
+    c2 = 2.0 * l_mat
+    c3 = (l_mat - 1.0) * (l_mat - 1.0)
+    d0 = jnp.sqrt((4.0 * c0 - 1.0) / (c0 - c1))
+    d1 = jnp.sqrt(((c2 + 1.0) * (c3 - c1)) / ((c2 - 3.0) * (c0 - c1)))
+  else:
+    with np.errstate(divide='ignore', invalid='ignore'):
```
jakevdp, line 693: I don't think this should be necessary as long as `m_mat` is a `jax.numpy` array, which appears to be the case.

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+  d0_mask_indices = jnp.triu_indices(l_max + 1, 1)
+  d1_mask_indices = jnp.triu_indices(l_max + 1, 2)
+  d_zeros = jnp.zeros((l_max + 1, l_max + 1))
+  d0_mask = d_zeros.at[d0_mask_indices].set(d0[d0_mask_indices])
+  d1_mask = d_zeros.at[d1_mask_indices].set(d1[d1_mask_indices])
+
+  # Creates a 3D mask that contains 1s on the diagonal plane and 0s elsewhere.
+  i = jnp.arange(l_max + 1)[:, None, None]
+  j = jnp.arange(l_max + 1)[None, :, None]
+  k = jnp.arange(l_max + 1)[None, None, :]
```
jakevdp, line 706: You could do this more concisely with `jnp.ogrid`:
```python
i, j, k = jnp.ogrid[:l_max + 1, :l_max + 1, :l_max + 1]
```

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+  mask = 1.0 * (i + j - k == 0)
+
+  d0_mask_3d = jnp.einsum('jk,ijk->ijk', d0_mask, mask)
+  d1_mask_3d = jnp.einsum('jk,ijk->ijk', d1_mask, mask)
+
+  return (d0_mask_3d, d1_mask_3d)
+
+
+@partial(jit, static_argnums=(2))
+def gen_derivatives(p: jnp.ndarray,
```
jakevdp, line 716: nit: this is a private function, so its name should start with an underscore.

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+      update_p_p2 = jnp.einsum('i,ij->ij', coeff, p_p2)
+      p_mm2_lm1 = p_mm2_lm1.at[ops.index[0, 3:num_l, :]].set(update_p_p2)
+
+  m_mat, l_mat = jnp.mgrid[:num_m, :num_l]
+
+  coeff_zeros = jnp.zeros((num_m, num_l))
+  upper_0_indices = jnp.triu_indices(num_m, 0, num_l)
+  zero_vec = jnp.zeros((num_l,))
+
+  with np.errstate(divide='ignore', invalid='ignore'):
```
jakevdp, line 768: `errstate` context shouldn't be necessary, because `m_mat` is a jax array, not a numpy array.

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+    a0 = -0.5 / (m_mat - 1.0)
+  a0_masked = coeff_zeros.at[upper_0_indices].set(a0[upper_0_indices])
+  a0_masked = a0_masked.at[1, :].set(zero_vec)
+
+  b0 = l_mat + m_mat
+  with np.errstate(invalid='ignore'):
```
jakevdp, line 774: `errstate` context shouldn't be necessary for jax arrays.

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+      g0 = g0 -  p[2, :, :]
+    p_derivative_m0 = jnp.einsum('j,ij->ij', 0.5 / jnp.sqrt(1 - x * x), g0)
+    p_derivative = p_derivative.at[1, :, :].set(p_derivative_m0)
+    p_derivative = p_derivative.at[1, 0, :].set(jnp.zeros((num_x,)))
+
+  return p_derivative
+
+
+@partial(jit, static_argnums=(0, 2))
+def gen_associated_legendre(l_max: int,
```
jakevdp, line 810: preceding underscore for private function

jax/_src/scipy/special.py
```diff
@@ -657,3 +661,290 @@ def i1e(x):
...
+         jnp.einsum('ij,ijk->ijk', coeff_1, jnp.roll(p_val, shift=2, axis=1)))
+    p_val = p_val + h
+    return p_val
+
+  p = lax.fori_loop(lower=2, upper=l_max+1, body_fun=body_fun, init_val=p)
+
+  return p
+
+
+@_wraps(osp_special.lpmn)
```
jakevdp, line 908: Note that using `_wraps` will overwrite the function's docstring with a modified version of the docstring from the wrapped function.

We generally use `_wraps()` for convenience, along with an optional `lax_description` argument to note departures from the wrapped API.

If you'd prefer to keep your docstring, you can remove the `_wraps` decorator.

jakevdp (changes requested):
Looks great! This is getting really close.

Along with the (minor) comments below, please also add your function to the list in https://github.com/google/jax/blob/master/docs/jax.scipy.rst so that it will appear in the HTML documentation.

jax/_src/scipy/special.py
tlu7: Thanks! Done.

jax/_src/scipy/special.py
tlu7: Done. Thanks!

jax/_src/scipy/special.py
tlu7: Done. Thanks!

jax/_src/scipy/special.py
tlu7: Done. Thanks!

jax/_src/scipy/special.py
tlu7: Done. Thanks!

jax/_src/scipy/special.py
tlu7: Done. Thanks!

jax/_src/scipy/special.py
tlu7: Thanks for the explanations. Removed decorator `_wraps`.

jakevdp (approved):
Looks great!

The last thing we need before merging is to squash all the changes into a single commit – let me know if you need pointers on how to do that.

tlu7:
Thanks for teaching me and walking me through the rebasing and squashing
@jakevdp!

Issue #6610, opened 2021-05-01
Local fork pip install instruction fails


njunge94:
I came across this when I started working on my first PR the other day, using the new [contributing document](https://jax.readthedocs.io/en/latest/contributing.html) in the JAX documentation. This was merged in #6573 according to the history.

Current instructions (under the section "Contributing code using pull requests"):

```
git clone https://github.com/YOUR_USERNAME/jax
cd jax
pip install -r requirements.txt  # Installs all testing requirements. <-- this fails because it points to the wrong file
pip install -e .  # Installs JAX from the current directory in editable mode.

# results in ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```
fail because of a wrong requirements file path. For me, adjusting the third line to 
```
pip install -r build/test-requirements.txt
```
instead does the job. This saves a few seconds / minutes and prevents possible early-stage frustration trying to locate the requirements file :)

jakevdp:
Thanks for the report! That was a typo on my part. Feel free to send a fix, otherwise I'll get to it soon :grin:

njunge94:
Just sent one in #6611.

PR #6611, opened 2021-05-01
Fix pip install instruction for forks in contributing section

nicholasjng:
In reference to issue #6610. Fixes the `pip install` instruction in the contributing document by putting the right path to the requirements file.

Maybe it could also be worthwhile to upgrade the "add JAX as an upstream remote" instruction directly below this block to use SSH or at least https, what do you think?


jakevdp (approved):
Looks good, thanks!

PR #6622, opened 2021-05-03
Add a tridiagonal eigh solver.

hawkinsp:
This code was written by Rasmus for TensorFlow; this PR ports it to JAX.

### DIFF
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -15,6 +15,7 @@
 
 from functools import partial
 
+import numpy as np
 import scipy.linalg
 import textwrap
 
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/linalg.py
```diff
@@ -434,3 +435,141 @@ def block_diag(*arrs):
...
+    raise TypeError("Only float32 and float64 inputs are supported as inputs "
+                    "to jax.scipy.linalg.eigh_tridiagonal, got "
+                    f"{alpha.dtype} and {beta.dtype}")
+  n = alpha.shape[0]
+  if n == 1:
+    return alpha
+  beta_abs = jnp.abs(beta)
+  beta_sq = beta * beta
+
+  # Compute an interval containing the eigenvalue of T using the Gerschgorin
```
mattjj, line 515:
```suggestion
  # Compute an interval containing the eigenvalue of T using the Gershgorin
```

hawkinsp: https://en.wikipedia.org/wiki/Semyon_Aranovich_Gershgorin says this is an accepted transliteration!

mattjj: Okay you win, though the spelling is inconsistent with a comment a few lines down. (Maybe that's good, as all possible transliterations are represented.)

mattjj (approved):
(Since GitHub already marked me as a reviewer, given my drive-by comment...)

Beautiful code as always!

hawkinsp:
> (Since GitHub already marked me as a reviewer, given my drive-by comment...)
> 
> Beautiful code as always!

To be clear, Rasmus gets all the credit for this code. I just ported it, which happily at the level of NumPy-like ops is almost completely mechanical to do.

(The autodiff stuff, when we have the pieces for it, will be a bit different.)

mattjj:
Copying Rasmus code is a recipe for success in any endeavor! :D

PR #6627, opened 2021-05-03
JVPs for sparse ops.

GJBoth:
This PR adds JVP rules to the `jax.experimental.sparse_ops` for `coo_todense, coo_fromdense, coo_matvec` and `coo_matmat`.  It builds on the work by @jakevdp in #6466 and #6249 and adds forward-mode differentiation to these ops.

The main rule I followed in implementing was that a sparse primal must be matched by a sparse tangent. In practice, this means that functions taking in a dense matrix and returning a sparse matrix yield a sparse tangent tuple `(tangent_in, Zero(rows), Zero(col))`, and vice versa. Under the hood I worked as much as possible with sparse arrays: for example, the jvp of `coo_matmat` is implemented in terms of sparse-dense matmul, instead of creating a dense tangent and using normal matmul. 

It's still a WIP, mostly on the front of tests. The main things I'm looking for input for:
* How to properly handle `ad_utils.Zero`. The current `coo_matmat` works but it's not exactly a michelangelo...
* What tests I would need to implement to check if the JVP is correct. The current `matmat` and `matvec` tests turn the sparse matrix into a scipy `coo_matrix` object (blocking a jvp check, see line 242 - 244), and I'm not sure why. My own tests showed it was correct, though.
* And of course whats needs to be done for it to be accepted :-)
 
The rules for the CSR and CSC format are exactly the same, so once I get the go ahead I'll copy paste (and change coo -> csr/csc) so we have those as well!

### DIFF
--- a/jax/experimental/sparse_ops.py
+++ b/jax/experimental/sparse_ops.py
@@ -39,16 +39,17 @@ from jax import api
 from jax import core
 from jax import jit
 from jax import tree_util
+from jax import lax
 from jax.interpreters import xla
 from jax.lib import cusparse
 from jax.lib import xla_bridge
 from jax.lib import xla_client
 import jax.numpy as jnp
 import numpy as np
-
+from jax.interpreters import ad
+from jax import ad_util
 xb = xla_bridge
 xops = xla_client.ops
-
 Dtype = Any
 
 #--------------------------------------------------------------------
@@ -298,6 +299,17 @@ if cusparse and cusparse.is_supported:
   xla.backend_specific_translations['gpu'][
       coo_todense_p] = _coo_todense_gpu_translation_rule
 
+def _coo_todense_jvp_rule(primals_in, tangents_in, **params):
+  vals, rows, cols,  = primals_in
+  mat_dot, rows_dot, cols_dot = tangents_in
+  assert type(rows_dot) is ad_util.Zero
+  assert type(cols_dot) is ad_util.Zero
+
+  primals_out = coo_todense(vals, rows, cols, **params)
+  tangents_out = ad_util.Zero.from_value(primals_out) if type(mat_dot) is ad_util.Zero else coo_todense(mat_dot, rows, cols, **params)
+  return primals_out, tangents_out
+ad.primitive_jvps[coo_todense_p] = _coo_todense_jvp_rule
+
 #--------------------------------------------------------------------
 # coo_fromdense
 
@@ -351,6 +363,14 @@ if cusparse and cusparse.is_supported:
   xla.backend_specific_translations['gpu'][
       coo_fromdense_p] = _coo_fromdense_gpu_translation_rule
 
+def _coo_fromdense_jvp_rule(primals_in, tangents_in, **params):
+  mat,  = primals_in
+  mat_dot, = tangents_in
+  data, row, col = coo_fromdense(mat, **params)
+  tangents_out = ad_util.Zero.from_value(data) if type(mat_dot) is ad_util.Zero else coo_fromdense(mat_dot, **params)[0]
+  return (data, row, col), (tangents_out, ad_util.Zero.from_value(row), ad_util.Zero.from_value(col))
+ad.primitive_jvps[coo_fromdense_p] = _coo_fromdense_jvp_rule
+
 #--------------------------------------------------------------------
 # coo_matvec
 
@@ -403,6 +423,22 @@ if cusparse and cusparse.is_supported:
   xla.backend_specific_translations['gpu'][
       coo_matvec_p] = _coo_matvec_gpu_translation_rule
 
+def _coo_matvec_jvp_rule(primals_in, tangents_in, **params):
+  vals, rows, cols, vec = primals_in
+  sparse_mat_dot, rows_dot, cols_dot, vec_dot = tangents_in
+  assert type(rows_dot) is ad_util.Zero
+  assert type(cols_dot) is ad_util.Zero
+
+  primals_out = coo_matvec(vals, rows, cols, vec, **params)
+  _zero = lambda p, t: lax.zeros_like_array(p) if isinstance(t, ad_util.Zero) else t
+
+  _sparse_mat_dot = _zero(vals, sparse_mat_dot)
+  _vec_dot = _zero(vec, vec_dot)
+
+  tangents_out = coo_matvec(_sparse_mat_dot, rows, cols, vec, **params) + coo_matvec(vals, rows, cols, _vec_dot, **params)
+  return primals_out, tangents_out
+ad.primitive_jvps[coo_matvec_p] = _coo_matvec_jvp_rule
+
 #--------------------------------------------------------------------
 # coo_matmat
 
@@ -453,6 +489,22 @@ xla.translations[coo_matmat_p] = xla.lower_fun(
 if cusparse and cusparse.is_supported:
   xla.backend_specific_translations['gpu'][
       coo_matmat_p] = _coo_matmat_gpu_translation_rule
+
+def _coo_matmat_jvp_rule(primals_in, tangents_in, **params):
+  vals, rows, cols, mat = primals_in
+  sparse_mat_dot, rows_dot, cols_dot, mat_dot = tangents_in
+  assert type(rows_dot) is ad_util.Zero
+  assert type(cols_dot) is ad_util.Zero
+
+  primals_out = coo_matmat(vals, rows, cols, mat, **params)
+  _zero = lambda p, t: lax.zeros_like_array(p) if isinstance(t, ad_util.Zero) else t
+  _sparse_mat_dot = _zero(vals, sparse_mat_dot)
+  _mat_dot = _zero(mat, mat_dot)
+
+  tangents_out = coo_matmat(_sparse_mat_dot, rows, cols, mat, **params) + coo_matmat(vals, rows, cols, _mat_dot, **params)
+  return primals_out, tangents_out
+ad.primitive_jvps[coo_matmat_p] = _coo_matmat_jvp_rule
+
 
 #----------------------------------------------------------------------
 # Sparse objects (APIs subject to change)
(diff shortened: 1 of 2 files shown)


GJBoth:
@googlebot I signed it!

jax/experimental/sparse_ops.py
```diff
@@ -351,6 +360,14 @@ def _coo_fromdense_gpu_translation_rule(c, mat, *, nnz, index_dtype):
   xla.backend_specific_translations['gpu'][
       coo_fromdense_p] = _coo_fromdense_gpu_translation_rule
 
+def _coo_fromdense_jvp_rule(primals_in, tangents_in, **params):
+  mat,  = primals_in
+  mat_dot, = tangents_in
+  data, row, col = coo_fromdense(mat, **params)
+  tangents_out = ad_util.Zero.from_value(data) if type(mat_dot) is ad_util.Zero else coo_fromdense(mat_dot, **params)[0]
```
jakevdp, line 367: There is a corner case here we need to worry about, I think: what if `mat_dot` has a different sparsity pattern than `mat`?

jax/experimental/sparse_ops.py
```diff
@@ -403,6 +420,23 @@ def _coo_matvec_gpu_translation_rule(c, data, row, col, v, *, shape, transpose):
   xla.backend_specific_translations['gpu'][
       coo_matvec_p] = _coo_matvec_gpu_translation_rule
 
+def _coo_matvec_jvp_rule(primals_in, tangents_in, **params):
+  vals, rows, cols, vec = primals_in
+  sparse_mat_dot, _, _, vec_dot = tangents_in
```
jakevdp, line 425: As above, I think we should assert that the underscored values are `ad.Zero`

jax/experimental/sparse_ops.py
```diff
@@ -403,6 +420,23 @@ def _coo_matvec_gpu_translation_rule(c, data, row, col, v, *, shape, transpose):
...
+  is_zero = lambda x: type(x) is ad_util.Zero
+
+  if is_zero(sparse_mat_dot) and is_zero(vec_dot):
+    tangents_out = ad_util.Zero.from_value(primals_out)
+  elif not is_zero(sparse_mat_dot) and is_zero(vec_dot):
+    tangents_out = coo_matvec(sparse_mat_dot, rows, cols, vec, **params)
+  elif is_zero(sparse_mat_dot) and not is_zero(vec_dot):
+    tangents_out = coo_matvec(vals, rows, cols, vec_dot, **params)
+  else:
+    tangents_out = coo_matvec(sparse_mat_dot, rows, cols, vec, **params) + coo_matvec(vals, rows, cols, vec_dot, **params)
```
jakevdp, line 437: I think we could simplify this with something like:
```python
_zero = lambda p, t: lax.zeros_like_array(p) if isinstance(t, ad.Zero) else t
data_dot = _zero(data, data_dot)
v_dot = _zero(v, v_dot)

tangents_out = (
  coo_matvec(data_dot, row, col, v, shape=shape, transpose=transpose) +
  coo_matvec(data, row, col, v_dot, shape=shape, transpose=transpose)
)
```

jakevdp (changes requested):
Looks good!

I've been thinking about this in parallel... one thing we'll have to face soon is that the autodiff for these operations is not complete until the relevant transpose rules are defined. The transpose rule for `coo_matvec` is fairly straightforward, but I'm not certain that it's possible in the transpose rules for `coo_todense` & `coo_fromdense` to easily "re-use" the sparsity patterns from the primals, because the sparsity patterns don't enter the expression!

This might indicate that the pytree approach to representing sparse matrices here is not viable. I'd like to hash that out and make certain there's a path forward before we merge these JVP rules.

jakevdp:
Hey - FYI in #6692 I have an alternate high-level implementation of sparse objects based on multi-buffers rather than pytrees. It might make implementing JVPs for sparse a little more natural... still, I think both approaches ending up hitting problems with transpose rules. It might be that we need the sparsity pattern encoded by `indices` to be an abstract property like `shape` and `dtype` in order to fully implement auto-diff over sparse inputs, although that has downsides of its own...

jax/experimental/sparse_ops.py
GJBoth: Good point. This is not just an issue here, though, but in general. I assume in all my implementations that the primal and tangent share the same sparsity pattern - and I'm not sure how to handle the case where this isn't the point? Do you think this is something we can assume for now and work on later, or not?

jax/experimental/sparse_ops.py
```diff
@@ -298,6 +299,14 @@ def _coo_todense_gpu_translation_rule(c, data, row, col, *, shape):
   xla.backend_specific_translations['gpu'][
       coo_todense_p] = _coo_todense_gpu_translation_rule
 
+def _coo_todense_jvp_rule(primals_in, tangents_in, **params):
+  vals, rows, cols,  = primals_in
+  mat_dot, _, _  = tangents_in
```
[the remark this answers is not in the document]
GJBoth: Done!

jax/experimental/sparse_ops.py
GJBoth: Done. I also added the assertions to `coo_matmat`.

jax/experimental/sparse_ops.py
GJBoth: I had something like this originally, but don't you run the risk of doing unnecessary ops if one of the tangents is zero? I changed it anyway - it's much prettier :-).

jakevdp:
Hey - I spent a bit of time chatting with @froystig today, who has a much better understanding of autodiff than I do. It turns out a couple of my review comments here are not exactly correct. In particular, I think we do not need to validate sparsity pattern equality, and with this idea in place, the transpose rules are fairly easy to derive.

I think the best way forward is to limit this PR to the JVP rules, which I think are more or less defined how we would like. I think that means reverting just the last commit, which adds transpose rules.

After that, if you squash all your JVP work into one commit (as mentioned in the [developer guide](https://jax.readthedocs.io/en/latest/contributing.html#contributing-code-using-pull-requests)) then we can merge this, and then add the transpose rules in a susequent PR.

What do you think?

jakevdp:
Let me know if you want to continue working on this - I'm eager to get these changes in so we can start looking into transposes, batching, etc.

GJBoth:
> I think the best way forward is to limit this PR to the JVP rules, which I think are more or less defined how we would like. I think that means reverting just the last commit, which adds transpose rules.

Works for me! I have some time tomorrow to fix this- I'll also add in the functions for the CSR and CSC versions. That good?

jakevdp (approved):
Very nice, thanks!

jakevdp:
As for next steps...

- take a look at #6739, which I will merge soon. @froystig was able to re-express some of the JVP & transpose rules in terms of existing utilities, and did so for all except the `coo_matmat`. A good next step would be to apply the same shortcuts to that, and add a suitable transpose rule.
- before adding jvps and transposes for the other primitives, it would be nice to develop a couple sparse test utilities that would reduce the boilerplate of all the COO tests. Maybe one similar to `_CheckAgainstNumpy` that compares a sparse op to the equivalent dense op, one similar to `_CompileAndCheck` that checks JIT compilation, and one similar to `check_grads` that checks gradients of a sparse operation. 
- once we have good testing infrastructure in place, we can add JVP and transpose rules for other primitives

Once that is done, I'd love to think a bit about batching rules: it would be natural in the gradient tests to use `jax.jacfwd` and `jax.jacref` to explore the entire basis space of gradients... but those use vmap, which is not yet implemented for the sparse primitives. I have some ideas about that - let me know if you want to discuss.

And in the meantime, @froystig and I have been discussing the larger strategy of how we marry sparse to JAX: for example, how do we make an operation like `jnp.dot(mat, vec)` do the right thing for sparse matrices? There are a few possible approaches, and I think the final form of the higher-level objects will depend on that answer. All that to say, I don't think right now is the best time to invest effort in trying to make `COO`/`CSC`/`CSR` feature-complete... think of those as skeletal prototypes for now, with the focus of the effort on developing this set of sparse primitives that play well with the rest of JAX.

How does that sound?

GJBoth:
That sounds good, #6739 looks really nice!

> and did so for all the JVPs here except coo_matmat. A good next step would be to apply the same shortcuts to that, and add a suitable transpose rule & tests.

I'll try and update `coo_matmat` in the same line. Hopefully this week, but probably next week as there's a neurips deadline coming which I've been putting off... Is that okay? It's not blocking anything right?

> Maybe one similar to _CheckAgainstNumpy that compares a sparse op to the equivalent dense op, one similar to _CompileAndCheck that checks JIT compilation, and one similar to check_grads that checks gradients of a sparse operation.

More checks would be great. I had some issues implementing some tests for the jvps, and the tests in #6739 already seem much nicer than what I came up with. I think getting even more up could speed up things a lot.

> but those use vmap, which is not yet implemented for the sparse primitives. I have some ideas about that - let me know if you want to discuss.

Yeah that'd be great - I'll read up a bit on how `vmap` works exactly, and then I'd be happy to discuss.

One thing I'm not seeing in your list is how to deal with the containers - should we use the pytrees or the buffers? Did you make any progress on this, or do you want to drop this for now?

>I'd love to work with you on any parts of this that interest you – if there are pieces that don't seem that exciting, let me know and I can add them to my own TODO list.

Happy to do anything really - I'm enjoying it and learning a ton. I'm still learning a lot about how jax works internally, so I'll follow your lead; just let me know where you think I can contribute and happy to do the work.

jakevdp:
> One thing I'm not seeing in your list is how to deal with the containers - should we use the pytrees or the buffers? Did you make any progress on this, or do you want to drop this for now?

I think the answer to this question will be more clear as we explore how sparse primitives interact with JAX transforms. For example, batching transforms might have implications in terms how how to generalize sparse representations to N dimensions. With that in mind, I think it would be premature to focus on high-level wrappers at the moment

PR #6639, opened 2021-05-04
Remat extensions for scan-over-layers

jekbradbury:
We would like to express a stack of repeated neural network layers
(e.g. a ResNet or Transformer) using lax.scan, in order to reduce
compile times and obtain better control over memory usage.
Unlike grad of straight-line code, XLA compilation of grad of scan won't
touch JAX's residual choices (including those made through jax.remat).
But in order to express the residual choices we want, without blocking
other needed optimizations, we need a few extensions to remat. This
change adds the following params to remat_call and kwargs to jax.remat:

(1) allows disabling the "gadget" that blocks XLA optimizations across a
    remat block (since an outer scan already prevents undoing remat,
    and the extra gadget prevents important optimizations)
(2) adds an "extend" param that enables a remat block to be
    automatically extended to encompass certain primitives that are
    direct consumers of outputs of the remat (and would otherwise cause
    those outputs to be unnecessarily saved as residuals)

The intent of the latter option is to enable "one residual per matmul"
by wrapping nonlinearities and normalizers in jax.remat.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2397,7 +2397,8 @@ def eval_shape(fun: Callable, *args, **kwargs):
   return tree_unflatten(out_tree(), out)
 
 
-def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
+def checkpoint(fun: Callable, concrete: bool = False, extend: bool = False,
+               gadget: bool = True) -> Callable:
   """Make ``fun`` recompute internal linearization points when differentiated.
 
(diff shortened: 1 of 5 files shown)


shoyer:
These tricks look pretty handy, but also low-level -- it's not obvious how to use them correctly. I wonder if they could also be encapsulated in a higher level "scan with remat" helper?

jakevdp:
Closing as obsolete.

PR #6640, opened 2021-05-04
Do not perform second call to config_with_absl

qstanczyk:
Avoid double-defined flags errors as reported in https://github.com/deepmind/launchpad/issues/5 by not executing logic in config_with_absl if it has been already called.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -123,6 +123,8 @@ class Config:
                     update_hook=update_hook)
 
   def config_with_absl(self):
+    if self.use_absl:
+      return
     # Run this before calling `app.run(main)` etc
     import absl.flags as absl_FLAGS  # noqa: F401
     from absl import app, flags as absl_flags


jakevdp (commented):
Thanks for the fix - this looks like a reasonable approach to avoid double initialization.

I wonder though whether silently doing nothing on a second call is the right approach, as it might hide more subtle bugs (e.g. a user may expect that flags will be parsed again after an update). Perhaps we should fail with a more explicit error in this case? (e.g. `raise ValueError("config_with_absl() cannot be called twice in the same script")`)

qstanczyk:
In our use case this function is currently called twice (from the user code and from one of the imported libraries). We could add a logic to not call this method for the seconds time in the liobrary, but basically it would be equivalent to this proposal. Maybe we could add a parameter "fail_on_double_call" and default to true?

jakevdp:
Sorry to lose the thread on this.

Thinking about it more, in general calling `config_with_absl` twice is indicative of a larger problem with the script: it means the caller is expecting explicit configuration initialization in a place where it is not allowed. I think it would be better to continue raising an error in this case rather than to silence it. For your application, if you're certain that calling it twice is not problematic, perhaps you can check the state at the call site.

What do you think?

qstanczyk:
Sounds good, thanks.

PR #6656, opened 2021-05-05
Update profiling instructions.

skye:


skye:
Here's the rendered doc: https://jax--6656.org.readthedocs.build/en/6656/profiling.html

docs/profiling.md
```diff
@@ -12,12 +12,22 @@ GPU and TPU. The end result looks something like this:
...
 ### Installation
 
-Install specific nightly versions of TensorBoard, TensorBoard profiler, TensorFlow, as well as the What-If Tool TensorBoard, as follows:
+The TensorBoard profiler is only available with the version of TensorBoard
+bundled with TensorFlow.
+
 ```shell
-# Profiling is known to work with the following packages.
-pip install --upgrade tb-nightly==2.5.0a20201203 tbp-nightly==2.4.0a20201203 tf-nightly==2.5.0.dev20201203 tensorboard-plugin-wit==1.7.0
+pip install tensorflow tbp-nightly
```
hawkinsp, line 19: Perhaps it's still worth identifying a known-good tbp-nightly release?

docs/profiling.md
```diff
@@ -166,6 +177,11 @@ Add the path to `libcupti.so` to the environment variable `LD_LIBRARY_PATH`.
 export LD_LIBRARY_PATH=/usr/local/cuda-10.1/extras/CUPTI/lib64/:$LD_LIBRARY_PATH
 ```
 
+If you still get the `Could not load dynamic library` message after doing this,
+check if the GPU trace shows up in the trace viewer anyway. This message
+sometimes occurs even when everything is working, since it looks for the
```
hawkinsp, line 182: We should fix this to at least hide the warning in the case it finds it in the second place it looks.

hawkinsp (approved):

docs/profiling.md
skye: Good point. I noted the tbp-nightly and tensorflow versions in the parenthetical statement below. I'm also trying to put together end-to-end Cloud TPU VM testing for this, although I don't think I can test that the trace viewer loads correct, just that TB can start. I also wanna change these instructions to use tensorflow-plugin-profile once 2.5.0 is released.

docs/profiling.md
skye: Agreed. I'm not gonna gate these docs on doing that, but I'll delete this paragraph if/when we fix it. (I can take a quick look, but don't wanna spend too much time on it right now.)

PR #6658, opened 2021-05-06
Generalize constant handlers for multi-buffer objects

jakevdp:
Paired with @mattjj


mattjj (approved):
LGTM! Though the test failures suggest we're not flattening somewhere we need to...

jakevdp:
I'll take a look tomorrow, but I suspect the issue is here: https://github.com/google/jax/blob/8ded7ad895760699b325fee138f63edbd1826b62/jax/_src/lax/control_flow.py#L96-L97

mattjj:
Hmm that's all avals and jaxprs though, should be unrelated to the xla lowering...

PR #6662, opened 2021-05-06
Make jnp.unravel_index rank promotion explicit.

mkunesch:
There was a silent rank promotion in jax.numpy.unravel_index which caused the
code to fail with jax_numpy_rank_promotion='raise'. This commit makes the rank
promotion explicit. Furthermore, it fixes an outdated comment in the same
function.


mkunesch:
Should I add a `jax.numpy_rank_promotion("raise"):` context to the existing test for unravel_index to guarantee the behavior in the future? I was unsure since the pattern isn't used in any of the other tests. Thanks a lot!

jakevdp (approved):
Thanks! This looks like a reasonable fix.

Regarding the test, I don't think we currently have any policy of test coverage in strict mode. That may be worth thinking about, but for now I'm happy to leave this untested.

mkunesch:
Cool, thanks a lot for the review!

PR #6682, opened 2021-05-07
Update code example in custom VJP documentation.

bdamoc:
Update example code to call `skip_app` function from VJP `skip_app_fwd` function.


jakevdp (approved):
Good catch, thanks!

PR #6698, opened 2021-05-09
Add support for non-numeric dtypes in associative_scan

danieldjohnson:
`associative_scan` uses an interleave method, which was implemented
using `lax.add`, meaning that it only worked for numeric dtypes.
Instead, we can use dynamic updating to set the elements without having
to add zeros to them.


hawkinsp:
I think https://github.com/google/jax/pull/6738 is a more conservative fix, in the sense that it fixes only the boolean problem without otherwise changing the performance of the associative scan. Closing this PR in favor of that one.

PR #6709, opened 2021-05-10
Minor type annotation correction

NeilGirdhar:
* `jax.random.categorical(key, np.ones(4))` is allowed.
* Similarly `jax.random.shuffle(key, np.arange(4))` is allowed with any
  dtype.
* Made `jax.random.permutation` consistent with `shuffle`.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -501,7 +501,7 @@ def _randint(key, shape, minval, maxval, dtype):
   return lax.add(minval, lax.convert_element_type(random_offset, dtype))
 
 
-def shuffle(key: jnp.ndarray, x: jnp.ndarray, axis: int = 0) -> jnp.ndarray:
+def shuffle(key: jnp.ndarray, x: np.ndarray, axis: int = 0) -> jnp.ndarray:
   """Shuffle the elements of an array uniformly at random along an axis.
 
   Args:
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks for looking at this – I think in general we indend that anything annotated `jnp.ndarray` also accepts `np.ndarray`, but not vice-versa; you can see this via `isinstance` checks:
```python
>>> import numpy as np
>>> import jax.numpy as jnp
>>> isinstance(np.ones(4), jnp.ndarray)
True
>>> isinstance(jnp.ones(4), np.ndarray)
False
```
Given this, I think that for functions that accept both numpy and jax.numpy arrays as input `jnp.array` is the right type annotation.

Caveat: array annotations in general currently have very poor support in mypy & pytype; this is why in most places you'll see us annotate array inpyts with `Array` which is aliased to `Any`.

jakevdp:
Ah, I see. There's also the issue that when tracing functions, jax passes `Tracer` objects in place of arrays, so the `np.ndarray` type will fail there I suspect. We've not yet landed on any way to have meaningful mypy-compatible type annotations for traceable arguements to JAX functions.

If you want to fix this, I'd suggest doing what we do elsewhere: set `Array = Any` and annotate it with type `Array`. This is useful for people reading & understanding the source code, and mypy will never complain :grin:

NeilGirdhar:
@jakevdp Okay, sounds great!  Would it be possible to change these two `jnp.ndarray` to `Array` (`Any`) so that MyPy doesn't complain?

PR #6717, opened 2021-05-11
[jax2tf] Change the conversion of dot_general to use XLA op.

gnecula:
Instead of converting the dot_general to a sea of TF ops, when
we enable_xla we just use the XLA op. This has the advantage
that it also supports the preferred_element_type.

Also, it turns out that `tf.linalg.matmul` now supports tf.bfloat16 and tf.int32, so change the 
conversion for `not _enable_xla` to use `tf.linag.matmul` for those cases as well.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6653,6 +6653,11 @@ def remaining(original, *removed_lists):
 
 
 def _canonicalize_precision(precision: PrecisionLike) -> Optional[Tuple[PrecisionType, PrecisionType]]:
+  """Turns an API precision specification, into a pair of enumeration values.
+
+  The API can take the precision as a string, or int, and either as a single
+  value to apply to both operands, or as a sequence of two values.
+  """
   if precision is None:
     if config.jax_default_matmul_precision is None:
       return None
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6652,28 +6652,30 @@ def remaining(original, *removed_lists):
   return [i for i in original if i not in removed]
 
 
-def _canonicalize_precision(precision):
+def _canonicalize_precision(precision: PrecisionLike) -> Optional[Tuple[PrecisionType, PrecisionType]]:
   if precision is None:
     if config.jax_default_matmul_precision is None:
       return None
     try:
-      return _precision_strings[config.jax_default_matmul_precision]
+      precision = _precision_strings[config.jax_default_matmul_precision]
+      return (precision, precision)
     except KeyError:
       raise ValueError(
           "jax_default_matmul_precision flag must be set to None or a value in "
           f"{_precision_strings}, but got {config.jax_default_matmul_precision}"
       ) from None
   elif isinstance(precision, str) and precision in _precision_strings:
-    return _precision_strings.get(precision)
+    precision = _precision_strings.get(precision)
+    return (precision, precision)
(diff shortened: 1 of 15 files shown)


gnecula:
Hold off reviewing, I found more problems.

jax/experimental/jax2tf/examples/serving/model_server_request.py
```diff
@@ -15,7 +15,7 @@
 
 See README.md for instructions.
 """
-import grpc
+import grpc  # type: ignore[import]
```
saeta, line 18: For my own edification, could you explain why this is the right approach here?

gnecula:
This is ready now. I will likely try to merge very soon, because it is blocking other changes. Please do send comments and I will incorporate them in a future CL.

jax/_src/lax/lax.py
```diff
@@ -6652,28 +6652,30 @@ def remaining(original, *removed_lists):
   return [i for i in original if i not in removed]
 
 
-def _canonicalize_precision(precision):
+def _canonicalize_precision(precision: PrecisionLike) -> Optional[Tuple[PrecisionType, PrecisionType]]:
```
saeta, line 6655: Doc string? (e.g. to explain the meaning of the tuple values)

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1226,24 +1236,38 @@ def _conv_general_dilated(lhs, rhs, window_strides, padding, lhs_dilation,
 tf_impl_with_avals[lax.conv_general_dilated_p] = _conv_general_dilated
 
 
-def _dot_general(lhs, rhs, dimension_numbers, precision, preferred_element_type):
+def _dot_general(lhs, rhs, *,
```
saeta, line 1239: The third argument is surprising to me.

saeta (approved):
I feel somewhat less qualified to review this PR. I'm preemptively approving, but if you would like a more substantive review, I recommend either a video call, or someone else if you're more pressed for time.

jax/experimental/jax2tf/jax2tf.py
gnecula: The `*`? It says that the rest of the arguments must be passed as kwargs, not positional.

jax/experimental/jax2tf/examples/serving/model_server_request.py
gnecula: We run `mypy` in OSS, and it complains if it cannot find one of the modules. `grpc` is not a standard requirement for JAX, and we don't want it to be. You need it only to run this serving example. For myself, when `mypy` complains, if I do not see the type error, I just `type: ignore`. My thinking is that we want the type annotations for documentation first, and the actual type checking is best effort.

gnecula:
> I feel somewhat less qualified to review this PR. I'm preemptively approving, but if you would like a more substantive review, I recommend either a video call, or someone else if you're more pressed for time.

Indeed, there are many details here. I sent this to you mostly as education. I do think it is worth trying to digest it, but indeed it may be easier over a video call.

jax/_src/lax/lax.py
gnecula: I added a docstring. I wish we did have more docstrings. For now, what we are trying to do is to have docstrings for all API functions, then ideally for all module exports, and probably not add them for module locals. I think that we should
add type annotations for all functions though.

PR #6720, opened 2021-05-11
[jax2tf] Fix the scoping of the enable_xla conversion parameter

gnecula:
Previously, the global enable_xla flag was set upon entry to
`jax.convert`. It should instead be set only for the duration
of the just-in-time conversion, which may happen later when
the converted function is invoked.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -190,8 +190,6 @@ def convert(fun: Callable, *,
     A version of `fun` that expects TfVals as arguments (or
     tuple/lists/dicts) thereof, and returns TfVals as outputs.
   """
-  global _enable_xla
-  _enable_xla = enable_xla
   api._check_callable(fun)
 
   def converted_fun(*args: TfVal) -> TfVal:
@@ -276,6 +274,9 @@ def convert(fun: Callable, *,
     try:
       global _shape_env
       assert not _shape_env, f"Unexpected shape environment {_shape_env}"
+      global _enable_xla
+      prev_enable_xla = _enable_xla
+      _enable_xla = enable_xla
       _shape_env = shapeenv
 
       if with_gradient:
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -293,6 +294,7 @@ def converted_fun_flat_with_custom_gradient(*args_flat: TfVal) -> TfVal:
                     for o, _ in out_flat_raw]
     finally:
       _shape_env = {}
+      _enable_xla = _default_enable_xla
```
saeta, line 298: Sorry, I'm a little confused about the logic here, but perhaps I don't properly understand the behavior you're going for. I would have assumed the implementation would have stored the original value of `_enable_xla` in a local variable at the beginning of the function, and reset it back in the finally block here, instead of setting it to the original default. But I'm guessing you intend some different behavior?

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -516,6 +517,21 @@ def jax_fn_array(x):
...
+      return lax.pad(x, np.float32(0), [(-1, 0, 0), (0, 0, 0)])
+
+    tf_fun_with_xla = jax2tf.convert(fun, enable_xla=True)
+    tf_fun_without_xla = jax2tf.convert(fun, enable_xla=False)
+    x = np.ones((2, 3), dtype=np.float32)
+
+    self.assertAllClose(fun(x), tf_fun_with_xla(x))
+    with self.assertRaisesRegex(NotImplementedError,
+                                "Call to pad can only be converted through TFXLA"):
+      tf_fun_without_xla(x)
```
saeta, line 533: For completeness / pedantry, consider converting a different function in the opposite order (for values of `_enable_xla`).

saeta (approved):
Nice fix! Eagerly approving to not block you given the time zone differential.

jax/experimental/jax2tf/jax2tf.py
gnecula: Either way works in this case because the conversions can never be nested. I thought that this logic makes is cleared that you always restore the default. But maybe it is confusing as you point out. I have changed it to restore the previous value.

jax/experimental/jax2tf/tests/jax2tf_test.py
gnecula: Done.

saeta (approved):

PR #6722, opened 2021-05-11
[jax2tf] Improve conversion of pad when enable_xla=True, add tests.

gnecula:
If enable_xla is True we should directly use the XLA conversion, without
trying to see whether the primitive can actually be converted
to non XLA TF ops. This is happening elsewhere, but it was not
happening for pad.

Also discovered that we were never turning enable_xla=False in
tests, even when we had special harnesses for it.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -114,11 +114,13 @@ tf_impl_with_avals: Dict[core.Primitive,
 # requiring a TFXLA operation, an exception is thrown instead.
 _enable_xla = True
 
-def _xla_path_disabled_error(primitive_name: str) -> Exception:
+def _xla_disabled_error(primitive_name: str,
+                        extra_msg: Optional[str] = None) -> Exception:
   assert not _enable_xla
-  return NotImplementedError(
-    f"Call to {primitive_name} can only be converted through TFXLA, but "
-     "XLA is disabled")
+  msg = f"Call to {primitive_name} cannot be converted with enable_xla=False."
+  if extra_msg:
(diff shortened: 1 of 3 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1350,15 +1350,17 @@ def _pad(operand, padding_value, *, padding_config,
...
   if all(lo >= 0 and hi >= 0 and i == 0 for lo, hi, i in padding_config):
     return tf.pad(operand, util.safe_zip(low, high),
                   mode="CONSTANT", constant_values=padding_value)
-  if not _enable_xla:
-    raise _xla_path_disabled_error("pad")
-  out = tfxla.pad(operand, padding_value, low, high, interior)
-  # TODO(b/184499027): improve shape inference for XlaPad
-  out.set_shape(_aval_to_tf_shape(_out_aval))
-  return out
+  raise _xla_path_disabled_error("pad")
```
saeta, line 1362: Would it make sense to add to the error message more information about why the op can't be converted? (As it looks like it can be converted in some cases.)

saeta (approved):
LGTM, with one question that can be discussed asynchronously. Thanks for sending my this PR!

jax/experimental/jax2tf/jax2tf.py
gnecula: In this particular case it does, because there are cases when we can convert pad, and they can be described succinctly.    For convolutions there are also cases, but would be too hard to describe. For the other ops, we simply don't have a fallback. 

I have added the additional text to the error for pad.

saeta (approved):
Thanks!

PR #6724, opened 2021-05-11
Reenable numpy dispatch tests.

### LINKED ISSUE #5663: numpy-dispatch tests disabled for numpy 1.20
I am filing this issue to remember to turn back on the numpy-dispatch tests (in [ci-build.yaml](https://github.com/google/jax/blob/master/.github/workflows/ci-build.yaml#L61-L69)). I turned them off when trying to fix the numpy 1.20 issues. Now the rest of JAX works with numpy 1.20, but we are still keeping a pin to [numpy<1.20](https://github.com/google/jax/blob/master/build/test-requirements.txt#L3) until we release the next jaxlib. At that point we should try to re-enable the numpy-dispatch tests.

hawkinsp:
Fixes #5663

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -40,14 +40,13 @@ jobs:
             enable-x64: 0
             package-overrides: "none"
             num_generated_cases: 25
-# TODO: re-enable this for numpy 1.20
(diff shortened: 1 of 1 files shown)


jakevdp (approved):

gnecula (approved):
Thanks

PR #6725, opened 2021-05-11
Fix broken FFT following `flatbuffers` 2.0 release

nrontsis:
Certain `jax` functions are broken as of today due to a [new `flatbuffers` version](https://github.com/google/flatbuffers/releases/tag/v2.0.0). This PR fixes that by restricting `jaxlib`'s requirements.

### Minimal example
Install latest `jax`, `jaxlib` on a new environment in `Debian GNU/Linux 10 (buster)` and run
```python
import jax.numpy as np
np.fft.rfft(np.zeros(10), axis=0)

```
the above raises `TypeError: EndVector() takes 1 positional argument but 2 were given in jax`.

### NB
I have just realised that this is almost a duplicate of #6712.

The difference is that this PR suggests not to use `flatbuffers 2.0`. Furthermore, this PR suggests a potentially high priority as certain functionality of `jax` is broken right now.

### DIFF
--- a/jaxlib/setup.py
+++ b/jaxlib/setup.py
@@ -32,7 +32,7 @@ setup(
     author_email='jax-dev@google.com',
     packages=['jaxlib', 'jaxlib.xla_extension-stubs'],
     python_requires='>=3.6',
-    install_requires=['scipy', 'numpy>=1.16', 'absl-py', 'flatbuffers<2.0'],
+    install_requires=['scipy', 'numpy>=1.16', 'absl-py', 'flatbuffers>=1.12,<2.0'],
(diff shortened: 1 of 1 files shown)


hawkinsp:
I think https://github.com/google/jax/pull/6712 is a better fix for this.

We are about to deploy a new jaxlib (0.1.66) that does support flatbuffers 2.0, which will fix the immediate problem. We can't retroactively add the `flatbuffers` version constraints to existing jaxlibs, but #6712 will fix the issue in case there are any future breaking changes.

hawkinsp:
For additional context #6710 added flatbuffers 2.0 support.

Thanks for the PR!

PR #6726, opened 2021-05-11
Add solver diagnostics support in custom_linear_solve

nicholasjng:
This work is directed towards issue #4322 on returning additional solver data from `custom_linear_solve`. I emulated the idea by @shoyer from that thread of adding a `has_aux` flag (similar to that in `grad`) that indicates whether an auxiliary object holding solver data (like diagnostics) is returned by `solve` (i.e., is present in the output pytree). 

This is very raw as of now, so I opened it as a draft first. However, I did manage to pipe through a mock/dummy diagnostics object in a CG run by slightly modifying `jax.scipy.sparse.linalg.cg`. I wrote a small working demo script for that, which I can share if that is of interest. Also, I did not include the aforementioned modification of `cg` in this commit to limit the scope of this PR to `custom_linear_solve`.

Things that still definitely need improvement here include:
1. Salvaging the `_check_tree` call to handle the case of `has_aux == True`,
2. Maybe improving the `_shape_checked` logic to still use only a single nested `f` with some clever tuple unpacking,
3. Deciding whether `has_aux` should also apply to `transpose_solve` if specified (it currently does),
4. Adding tests with aux data.

I would be excited about some feedback!

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -49,7 +49,7 @@ from jax._src.util import (partial, unzip2, unzip3, safe_map, safe_zip,
                            split_list, cache, extend_name_stack)
 from jax.tree_util import (tree_flatten, tree_unflatten, treedef_is_leaf,
                            treedef_children, treedef_tuple, tree_multimap,
-                           tree_leaves)
+                           tree_leaves, tree_structure)
 from jax import ad_util
 from jax.config import config
 
@@ -1954,7 +1954,19 @@ def _check_tree_and_avals(what, tree1, avals1, tree2, avals2):
         f"{tree_unflatten(tree2, avals2)}.")
 
 
-def _check_tree(func_name, expected_name, actual_tree, expected_tree):
+def _check_tree(func_name, expected_name, actual_tree, expected_tree, has_aux=False):
+  if has_aux:
+    actual_tree_children = actual_tree.children()
+
+    if len(actual_tree_children) == 2:
+      # select first child as result tree
+      actual_tree = tree_structure(actual_tree_children[0])
+    else:
+      raise ValueError(
+        f"{func_name}() produced a pytree with structure "
+        f"{actual_tree}, but a pytree tuple with auxiliary "
+        f"output was expected because has_aux was set to True.")
+
   if actual_tree != expected_tree:
     raise TypeError(
         f"{func_name}() output pytree structure must match {expected_name}, "
@@ -2141,7 +2153,7 @@ def _check_shapes(func_name, expected_name, actual, expected):
 
 @api_boundary
 def custom_linear_solve(
-    matvec, b, solve, transpose_solve=None, symmetric=False):
+    matvec, b, solve, transpose_solve=None, symmetric=False, has_aux=False):
   """Perform a matrix-free linear solve with implicitly defined gradients.
 
   This function allows for overriding or defining gradients for a linear
@@ -2160,7 +2172,7 @@ def custom_linear_solve(
     b: constant right handle side of the equation. May be any nested structure
       of arrays.
     solve: higher level function that solves for solution to the linear
-      equation, i.e., ``solve(matvec, x)) == x`` for all ``x`` of the same form
+      equation, i.e., ``solve(matvec, x) == x`` for all ``x`` of the same form
       as ``b``. This function need not be differentiable.
     transpose_solve: higher level function for solving the transpose linear
       equation, i.e., ``transpose_solve(vecmat, x) == x``, where ``vecmat`` is
@@ -2169,10 +2181,12 @@ def custom_linear_solve(
       ``symmetric=True``, in which case ``solve`` provides the default value.
     symmetric: bool indicating if it is safe to assume the linear map
       corresponds to a symmetric matrix, i.e., ``matvec == vecmat``.
+    has_aux: bool indicating whether the ``solve`` and ``transpose_solve`` functions
+      return auxiliary data like solver diagnostics as a second argument.
 
   Returns:
     Result of ``solve(matvec, b)``, with gradients defined assuming that the
-    solution ``x`` satisfies the linear equation ``matvec(x) == b``.
+      solution ``x`` satisfies the linear equation ``matvec(x) == b``.
   """
   if transpose_solve is None and symmetric:
     transpose_solve = solve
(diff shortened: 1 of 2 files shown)


shoyer:
> AFAIK, L-BFGS returns an optimizer result, is that compatible with autodiff?

We haven't written a custom gradient rule for `jax.scipy.optimize.minimize` yet, but indeed it is exactly the same issue. I would only expect gradients with respect to the optimum point and objective value to be well defined -- everything else would have a symbolic zero tangent.

njunge94:
I made some progress! I redid the internal pytree check to accomodate auxiliaries (this can later be reused to upgrade `lax.custom_root` for auxiliaries as well), and after finally understanding the JVP I was able to get the upgraded auxiliary-ready JVP to run.

Is it fine to set all of the `aux` tree leaves to `lax.zeros_like_array`s there, or do I need to use another API?

I also tried to add a parametrized test with `has_aux`, but what I did does not work with `jtu.check_grads` at the moment. I'm working on fixing that.

shoyer:
> Is it fine to set all of the `aux` tree leaves to `lax.zeros_like_array`s there, or do I need to use another API?

We `ad_util.Zero` for symbolic zeros that don't get evaluated. You can find a few examples using `ad_util.Zero.from_value()` elsewhere in `control_flow.py`.

njunge94:
Thanks for the hint. I implemented the eval rule for `jit`. Now I am working on the batching rule, and I can get the batched aux values out, but then there is a problem:

Consider a scalar aux value, like a `converged` bool indicating whether the solve function converged or not. For a batch of solves of size `n`, `vmap` batches this to a DeviceArray with shape `(n,)` - so far, so good. Now say I would like to vmap over a matrix of left-hand sides `b` stacked column-wise:

```py
# this is based on a test snippet from jax

a = rng.randn(3, 3)
b = rng.randn(3, 2)
expected = jnp.linalg.solve(a, b)

# linear_solve_aux returns a bool like e.g. {"converged": True}
actual_vmap = api.vmap(linear_solve_aux, (None, 1), 1)(a, b) # [3, 3] x [3, 2] -> [3, 2]
```
This fails because `out_axis =  1`, which a batched array of scalars does not have as `ndim == 1`. Changing the `out_axis` to 0 makes it work. My solution was therefore to promote every aux value array shape to at least the same number of dimensions as given by the out axis, and then put the batch axis up front. Do you have any idea how to do this? I saw `batching.broadcast` being used, but that only adds one dimension at a time (if I understood correctly), where multiple may be required.

I'm guessing `jax.numpy` cannot be used here either, since we are operating on abstract values / shaped arrays.

shoyer:
I'm not sure you need special handling here. There are two ways you could
do this sort of thing with JAX out of the box:
1. Write out_axis=-1
2. Specify per output axes using a pytreees matching the output, e.g., with
out_axis=(1,0)

njunge94:
Thank you! In hindsight, that should have been obvious, but I'm glad you clarified. I think it's pretty much done, so I opened it up for review.

tests/lax_control_flow_test.py
```diff
@@ -2067,6 +2067,62 @@ def linear_solve(a, b):
...
+
+    # vmap test
+    c = rng.randn(3, 2)
+    expected = jnp.linalg.solve(a, c)
+    expected_aux = tree_util.tree_map(partial(np.repeat, repeats=2), orig_aux)
+    actual_vmap, vmap_aux = api.vmap(linear_solve_aux, (None, 1), -1)(a, c)
+
+    self.assertAllClose(expected, actual_vmap)
+    jtu.check_eq(expected_aux, vmap_aux)
+
```
shoyer, line 2124: Maybe add `grad` test, too, just to make sure things don't break?

njunge94:
Thank you for the review. I am almost set to push the fixes later in the evening, I just need your input for a small thing that came up:

I gave no attention to the VJP machinery yet. When I try to evaluate the VJP of an augmented solve function `linear_solve_aux`, computing the cotangent fails due to a length mismatch between the `eqn.invars` and the `cts_out` in the call to `ad.backward_pass`; the auxiliary values do not appear here as `invars`, but as `DropVar` objects in the `eqn.outvars` of the transpose solve computation, leading to them not being recognized as cotangents by `ad.backward_pass`.

I then end up with too many cotangents for the given list of `invars`. Once I cut off the aux values returned by the primitive in the transpose rule entirely, everything is fine.

Is it consistent to not return any aux data from the cotangent evaluation? In theory, just the values with no aux represents the value of `f`'s pullback. But in practice, it involves a call to the primitive, so the aux metrics will be computed. What do you think?

shoyer:
> Is it consistent to not return any aux data from the cotangent evaluation? In theory, just the values with no aux represents the value of `f`'s pullback. But in practice, it involves a call to the primitive, so the aux metrics will be computed. What do you think?

I think we do have to drop `aux` data from the cotangent solve, just like how we have to drop it from the tangent solve -- there is nowhere for the information to be stored.

njunge94:
Good point. I guess what put me off is that in the JVP case, at least a zeros-like object matching the shape of aux is actually returned, which is not going to be the case with the VJP then.

Just so I get you correctly: Did you mean I should add a test combining `vmap` and `jvp`, or similar? Thanks in advance for clarification.

shoyer:
> Just so I get you correctly: Did you mean I should add a test combining `vmap` and `jvp`, or similar? Thanks in advance for clarification.

This might not be a bad idea, but I'm not sure it's needed. I mostly wanted at least one test that goes through each of the linear solve transformation rules, e.g., `vmap`, `jit`, `jvp`, `vjp`.

jax/_src/lax/control_flow.py
```diff
@@ -2318,7 +2367,12 @@ def _linear_solve_batching_rule(args, dims, axis_name, main_type, const_lengths,
       vecmat_jaxpr_batched, vecmat_x_bat = batching.batch_jaxpr(
           vecmat, size, vecmat_bat + b_bat, instantiate=x_bat,
           axis_name=axis_name, main_type=main_type)
-      x_bat_out = _map(operator.or_, vecmat_x_bat, solve_x_bat)
+
+      if num_aux > 0:
+        # batch all aux data by default
+        x_bat_out = _map(operator.or_, vecmat_x_bat + [True] * num_aux, solve_x_bat)
+      else:
+        x_bat_out = _map(operator.or_, vecmat_x_bat, solve_x_bat)
```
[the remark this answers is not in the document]
njunge94: Thanks! I took out the if-branch there.

tests/lax_control_flow_test.py
njunge94: I now opted to use array-like aux arguments, which enables the use of `jtu.check_grads` - I follow the same procedure now as in the test directly above.

froystig:
In case this is relevant and saves some work, or makes a test that much easier to write: calling `jacfwd` and `jacrev` will exercise `vmap` of `jvp` and of `vjp` respectively.

njunge94:
So, what's the status? I played around a little bit with `jacrev` and `jacfwd`, and it looks like everything is well-behaved there, too. I can add some tests still if that's of interest. I planned to tackle auxiliary data in `custom_root` next, so if my strategy here is OK, I would just go and emulate it there as well. What do you think?

jax/_src/lax/control_flow.py
```diff
@@ -1954,7 +1954,23 @@ def _check_tree_and_avals(what, tree1, avals1, tree2, avals2):
...
+  if has_aux:
+    actual_tree_children = actual_tree.children()
+    num_children = len(actual_tree_children)
+
+    if num_children == 2:
+      # select first child as result tree if b is a singleton
+      actual_tree = tree_structure(actual_tree_children[0])
+    elif num_children > 2:
+      # prune away only the last child (aux)
+      actual_tree = tree_structure(actual_tree_children[:-1])
```
shoyer, line 1967: When could this happen? I thought the return value should always be a tuple of two elements if `has_aux=True`.

njunge94: You are absolutely correct. I was not aware that the tuple is a pytree itself, and thought you had to operate on the tree structure of `b` there. I just cleaned this up in 0308527.

shoyer (approved):

shoyer:
Thanks @njunge94 !

PR #6739, opened 2021-05-12
Sparse AD: add transpose rules for COO primitives

### LINKED ISSUE #6544: Jax support for sparse matrices
Hi all,

I was discussing with @jakevdp in #6466 about adding support for sparse matrices in Jax. He already did quite some groundwork, but there's a lot of open questions on the what and how. Quoting Jake here:

> * 2D CSC/CSR/COO may not be the abstraction we want in the end... pydata-sparse has N-dimensional sparse arrays implemented using COO or a multi-dimensional generalization of CSR. Similarly taco has routines for general N-dimensional sparse computation. That route may be better, particularly for some deep learning workflows that e.g. have 4D batched arrays of sparse image sequences.
>* XLA scatter/gather implementations only go so far... any operations we implement should have efficient lowerings on as many backends as possible - I've started with cusparse on GPU because the low-level routines are already available in our GPU build process. This means that every non-trivial operation should be either implemented as a primitive or composed of other primitives
> * JAX is about composability, and in particular pmap/vmap, jit, and grad are central to the API. I want to spend some time exploring how we can best implement gradients of sparse operations or batched sparse operations, and also think about how ongoing work on dynamic shapes can be utilized to make jit of sparse operations more efficient and usable. With all that up in the air I want to keep the design minimal for now in case we need to re-think our approach based on what we find
> * all of the above should be motivated by real-world workflows that people have. I'm starting to gather a few example workflows so we can have that in mind.

So I set out to compare the implementations in pytorch and tensorflow to get a feel for what's out there, and considering the points above. Specifically, we're interested in how they implement it low-level: do they secretly turn everything dense under the surface, or are they given the full treatment with specific ops? 

I'm summarizing what I found in the table below  - TL;DR Pytorch offers better support. Since documentation and support are all over the place, in this [notebook](https://colab.research.google.com/drive/1dyfhD4SaSAfKPhx-aG7lF8BY88WlsWWx?usp=sharing) you can find a bunch of common operations with sparse tensor to see if they are supported. It's been a while since I did tensorflow, so let me know if I made a mistake!

| Feature                | Pytorch                                              | Tensorflow                                                   |
| :--------------------- | ---------------------------------------------------- | ------------------------------------------------------------ |
| Documentation          | [here](https://pytorch.org/docs/stable/sparse.html#) | [here](https://www.tensorflow.org/guide/sparse_tensor)       |
| Formats                | COO                                                  | COO                                                          |
| Supported ops          | variations of matmul                                 | sums, min, matmul, element-wise ops through `tf.sparse.map_values` |
| Specialized ops        | Probably[^5]                                                  | Seems so [^4]                                                         |
| Grad of sparse[^1]     | Limited to specific matmul                           | No.                                                          |
| Sparse grad[^2]        | Yes                                                  | No.                                                          |
| Dimensions             | 2D, Hybrid                                           | 2D - tensor supports higher-D, but operations not            |
| Uncoalesced[^3]allowed | Yes                                                  | No mention                                                   |
| Sparse - sparse matmul | No                                                   | No                                                           |
| Extra goodies | Has an adam optimizer for sparse tensor | Nope |

[^1]: Meaning gradient w.r.t to the sparse tensor
[^2]: Meaning calculating the gradient without turning it into a dense tensor
[^3]: Uncoalesced means duplicate entries in the coordinates - total value of the element is then sum of duplicate elements.
[^4]: Sparse ops are imported from gen_sparse_ops.py, which doesnt exist in the repo?
[^5]: They mention Torchcudasparse is now defunct [here](https://github.com/pytorch/pytorch/tree/master/aten/src) but I can't find any other mention in the source code.

To do
* Dig in the source code a little further (although if someone who actually is familiar with the code could chime in, that'd be great.)
* Check how pydata-sparse implement n-D sparse arrays.

Proposal:
* Get a very minimal differentiable sparse matmul working for COO format using XLA ops and one GPU-specific backend in cusparse (some help and guidance here would be nice)

I figured we can use this issue to track progress and discuss, so ... thoughts?

jakevdp:


jax/experimental/sparse_ops.py
```diff
@@ -292,6 +306,30 @@ def _coo_todense_abstract_eval(data, row, col, *, shape):
...
+
+  primals_out = coo_todense(data, row, col, shape=shape)
+  tangents_out = coo_todense(data_dot, row, col, shape=shape)
+
+  return primals_out, tangents_out
+
+def _coo_todense_transpose(ct, data, row, col, *, shape):
+  assert ad.is_undefined_primal(data)
+  if ad.is_undefined_primal(row) or ad.is_undefined_primal(col):
+    raise ValueError("Cannot differentiate with respect to sparse indices")
```
froystig, line 325:
```suggestion
    raise ValueError("Cannot transpose with respect to sparse indices")
```

jax/experimental/sparse_ops.py
```diff
@@ -345,6 +383,29 @@ def _coo_fromdense_gpu_translation_rule(c, mat, *, nnz, index_dtype):
...
+  _, row, col = primals_out
+  tangents_out = _coo_extract(row, col, Mdot), ad.Zero(row.aval), ad.Zero(col.aval)
+  return primals_out, tangents_out
+
+def _coo_fromdense_transpose(ct, M, *, nnz, index_dtype):
+  data, row, col = ct
+  assert len(data) == nnz
+  assert row.dtype == col.dtype == index_dtype
+  if isinstance(row, ad.Zero) or isinstance(col, ad.Zero):
+    raise ValueError("Cannot differentiate with respect to sparse indices")
```
froystig, line 402:
```suggestion
    raise ValueError("Cannot transpose with respect to sparse indices")
```

tests/sparse_ops_test.py
```diff
@@ -315,6 +319,107 @@ def test_matmul(self, shape, dtype, Obj, bshape):
...
+      for shape in [(5, 8), (8, 5), (5, 5), (8, 8)]
+      for dtype in jtu.dtypes.floating + jtu.dtypes.complex))
+  def test_coo_todense_ad(self, shape, dtype):
+    rng = rand_sparse(self.rng(), post=jnp.array)
+    M = rng(shape, dtype)
+    data, row, col = sparse_ops.coo_fromdense(M, nnz=(M != 0).sum())
+    f = lambda data: sparse_ops.coo_todense(data, row, col, shape=M.shape)
+
+    # Forward-mode
+    primals, tangents = api.jvp(f, [data], [jnp.ones_like(data)])
```
froystig, line 334: Might be easier to use `jacfwd` and `jacrev` in place of `jvp` and `vjp`, then compare all resulting entries.

Perhaps useful for all of the sparse work (AD aside) is comparing against a non-sparse reference computation (on the non-zero entries).

We can also test higher-order AD (e.g. forward-on-reverse), for instance with `jax.hessian`.

jax/experimental/sparse_ops.py
```diff
@@ -292,6 +306,30 @@ def _coo_todense_abstract_eval(data, row, col, *, shape):
 def _coo_todense_gpu_translation_rule(c, data, row, col, *, shape):
   return cusparse.coo_todense(c, data, row, col, shape=shape)
 
+def _coo_todense_jvp(primals, tangents, *, shape):
+  data, row, col = primals
+  data_dot, row_dot, col_dot = tangents
+
+  assert isinstance(row_dot, ad.Zero)
+  assert isinstance(col_dot, ad.Zero)
+  data_dot = lax.zeros_like_array(data) if isinstance(data_dot, ad.Zero) else data_dot
```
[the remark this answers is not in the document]
froystig: `ad.Zero.from_value` might be good for this (we'll also want a test).

PR #6740, opened 2021-05-12
Improve JVP rule for scatters with non-overlapping indices.

hawkinsp:
If the scattered values don't overlap, we don't need complicated masking logic to work out which of the two overlapping values "win".

Also fix spurious error for certain empty scatters.


mattjj (approved):
Nice!

PR #6746, opened 2021-05-13
Fixed Tuple_replace bug: index not incremented to replace

### LINKED ISSUE #6745: tuple_replace function in util.py seems incorrect
In the file jax._src.util.py

Current definition
```python
def tuple_replace(t, idx, val):
  assert 0 <= idx < len(t), (idx, len(t))
  return t[:idx] + (val,) + t[idx:]
```
looks like an almost copy of tuple_insert. I think the correct definition should be:

```python
def tuple_replace(t, idx, val):
  assert 0 <= idx < len(t), (idx, len(t))
  return t[:idx] + (val,) + t[idx+1:]
```

joshuagornall:
closes #6745 

### Description:

In the file:  `jax/_src./util.py`

The `tuple_replace` function is identical to the `tuple_insert` function already defined in the same file location (named above). In order to replace the tuple as the function intends the below code should be implemented. 

### Before

``` py
def tuple_replace(t, idx, val):
  assert 0 <= idx < len(t), (idx, len(t))
  return t[:idx] + (val,) + t[idx:]
```
### After

```py
def tuple_replace(t, idx, val):
  assert 0 <= idx < len(t), (idx, len(t))
  return t[:idx] + (val,) + t[idx+1:]
```
```diff
- return t[:idx] + (val,) + t[idx:]
+ return t[:idx] + (val,) + t[idx+1:]
```

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -332,7 +332,7 @@ def tuple_delete(t, idx):
 
 def tuple_replace(t, idx, val):
   assert 0 <= idx < len(t), (idx, len(t))
-  return t[:idx] + (val,) + t[idx:]
+  return t[:idx] + (val,) + t[idx+1:]
 
 # TODO(mattjj): replace with dataclass when Python 2 support is removed
 def taggedtuple(name, fields) -> Callable[..., Any]:


joshuagornall:
@googlebot I signed it!

jax/_src/util.py
```diff
@@ -332,7 +332,7 @@ def tuple_delete(t, idx):
 
 def tuple_replace(t, idx, val):
   assert 0 <= idx < len(t), (idx, len(t))
-  return t[:idx] + (val,) + t[idx:]
+  return t[:idx] + (val,) + t[idx+1:]
```
hawkinsp, line 335: In general, you should follow the style of the surrounding code. Put spaces around binary operators like `+`.

joshuagornall: Thanks for the feedback! 

Best wishes
Josh

zhangqiaorjc: Hi Josh, would you like to make the change and close this PR? thanks!

hawkinsp:
Closing in favor of #6793 which deletes this function instead.

Thanks for finding the bug!

PR #6757, opened 2021-05-14
Fix `jsp.linalg.lu` translation rule to pass backend arg to `lower_fun`.

LenaMartens:
If it doesn't, trying to run `lu` with a custom CPU backend when a GPU is present results in a `Unable to resolve runtime symbol: 'cuda_lu_pivots_to_permutation'` fatal error.

eg. fixes following repro from a user:
```
a = jnp.eye(2)
jax.jit(jnp.linalg.inv, backend='cpu')(a)
```
which before this fix throws:
```
E0513 20:00:46.341652 2455650 simple_orc_jit.cc:197] Unable to resolve runtime symbol:
`cuda_lu_pivots_to_permutation'.  
Hint: if the symbol a custom call target, make sure you've registered it with the JIT 
using XLA_CPU_REGISTER_CUSTOM_CALL_TARGET.
JIT session error: Symbols not found: [ cuda_lu_pivots_to_permutation ]
```

I didn't investigate deeply if there's other `lower_fun` translation rules that need to be updated in the same way, maybe that could be done as a follow-up?

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -989,7 +989,7 @@ def _tuple_output(*args, **kwargs):
   ans = yield args, kwargs
   yield (ans,)
 
-def lower_fun(fun, multiple_results, parallel=False, with_avals=False):
+def lower_fun(fun, multiple_results, parallel=False, with_avals=False, backend=None):
   # TODO(jakevdp): migrate dependent code & always use the with_avals=True.
   def f(c, *xla_args, **params):
     avals = [_array_aval_from_xla_shape(c.get_shape(x)) for x in xla_args]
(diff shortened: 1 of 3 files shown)


jax/interpreters/xla.py
```diff
@@ -989,7 +989,7 @@ def _tuple_output(*args, **kwargs):
   ans = yield args, kwargs
   yield (ans,)
 
-def lower_fun(fun, multiple_results, parallel=False, with_avals=False):
+def lower_fun(fun, multiple_results, parallel=False, with_avals=False, backend=None):
```
hawkinsp, line 992: In a follow-up PR, I'd be tempted to make `backend` mandatory, given that it seems highly likely that anyone not providing a `backend` might hit similar misbehavior.

hawkinsp (approved):
Do you want to take the follow-up PR as well?

jax/interpreters/xla.py
LenaMartens: sgtm, wasn't sure how wide-spread this buggy behavior would be. I'll send a follow-up.

hawkinsp: Well, I'd guess it's rare! But this way we need never hit this bug again...

PR #6758, opened 2021-05-14
Simplify `unzip2` and `unzip3` utility functions

lgeiger:
`zip()` is its own inverse when unpacking the iterator with `*`. This PR uses this behaviour to remove the specialised `unzip2` and `unzip3` implementations. Performance wise this is even preferable since no intermediate lists need to be constructed.

I am also happy to replace all uses of `unzip2` and `unzip3` with either `unzip` or just with `xs, ys, zs = zip(*xyzs)` since when directly unpacking, the implementation is almost as readable as calling `unzip`. Let me know if you would prefer that.

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -28,6 +28,8 @@ partial = functools.partial
 
 
 def safe_zip(*args):
+  if not args:
+    return []
   n = len(args[0])
   for arg in args[1:]:
     assert len(arg) == n, 'length mismatch: {}'.format(list(map(len, args)))
@@ -41,22 +43,10 @@ def safe_map(f, *args):
   return list(map(f, *args))
 
 def unzip2(xys):
-  xs = []
-  ys = []
-  for x, y in xys:
-    xs.append(x)
-    ys.append(y)
-  return tuple(xs), tuple(ys)
+  return tuple(safe_zip(*xys)) or ((), ())
 
 def unzip3(xyzs):
-  xs = []
-  ys = []
-  zs = []
-  for x, y, z in xyzs:
-    xs.append(x)
-    ys.append(y)
-    zs.append(z)
-  return tuple(xs), tuple(ys), tuple(zs)
+  return tuple(safe_zip(*xyzs)) or ((), (), ())
 
 def subvals(lst, replace):
   lst = list(lst)


jakevdp:
I think the test failures point to the reason these utilities are defined the way they are: for empty iterators, `unzip2` and `unzip3` always return tuples of length 2 and 3, respectively. `zip(*x)` returns an empty tuple when `x` is empty.

lgeiger:
> for empty iterators, unzip2 and unzip3 always return tuples of length 2 and 3, respectively. zip(*x) returns an empty tuple when x is empty.

Good point, sorry for exercising CI instead of checking this locally. I changed the PR to handle empty iterators correctly.

mattjj:
Thanks for looking for ways to improve JAX internals!

But I think there's some misunderstanding here: we're well aware of how `zip*` can be used, but `unzip2` and `unzip3` are specialized for good reasons, including whether they raise clear errors or fail silently:

```python
In [1]: from jax._src.util import unzip2

In [2]: def new_unzip2(xys):
   ...:     return tuple(zip(*xys)) or ((), ())
   ...:

In [3]: unzip2([(1, 2), (3,)])
<ipython-input-3-40905abf4923> in <module>
----> 1 unzip2([(1, 2), (3,)])

~/packages/jax/jax/_src/util.py in unzip2(xys)
     44   xs = []
     45   ys = []
---> 46   for x, y in xys:
     47     xs.append(x)
     48     ys.append(y)

ValueError: not enough values to unpack (expected 2, got 1)

In [4]: new_unzip2([(1, 2), (3,)])
Out[4]: ((1, 3),)
```

You wouldn't believe how many times that's saved us!

So in other words, having `unzip2` and `unzip3` with explicit, specialized implementations is quite intentional.

In any case, thanks again for looking for ways to improve! This isn't a spot to improve, but I'm sure there are many others.

mattjj:
Maybe we should add a comment in the code here, like "avoid using `zip(*...)` for better errors"...

lgeiger:
Yeah, sorry about that. This makes sense. I just stumbled upon this while trying to debug an unrelate issue on my side and though this might be simpler.

I'm happy to change the code to
```python
def new_unzip2(xys):
     return tuple(safe_zip(*xys)) or ((), ())
```
which throws a nice `AssertionError: length mismatch: [2, 1]`. Otherwise we can just keep the PR closed.

mattjj:
That might work, actually! That's a good idea. We should convince ourselves there are no other differences with the current `unzip2`, but maybe this covers it (edge cases and errors).

mattjj:
(Sorry, I was wrong to close the issue so soon! As you showed, it turns out we just needed one more tweak to get what we needed with fewer lines.)

lgeiger:
No worries at all, thanks for the blazing fast reviews 🚀

jakevdp:
Quick thought: one option would be to add an optional `n=None` keyword parameter to `safe_zip`; if we define it appropriately then we could write
```
def unzip2(vals):
  return safe_zip(*vals, n=2)
```
and as a bonus, it would allow us to explicitly specify the number of expected elements in each argument to `safe_zip` where it would be useful.

mattjj:
I think I prefer `unzip2` and `unzip3` to be spelled out over adding a kwarg to `safe_zip`.

I think Jake is making the excellent point that this behavior isn't what we want:

```python
In [1]: from jax._src.util import unzip2, safe_zip

In [2]: def new_unzip2(xys):
   ...:     return tuple(safe_zip(*xys)) or ((), ())
   ...:

In [3]: new_unzip2([(1,), (2,)])
Out[3]: ((1, 2),)

In [4]: unzip2([(1,), (2,)])
<ipython-input-4-ee726b410613> in <module>
----> 1 unzip2([(1,), (2,)])

~/packages/jax/jax/_src/util.py in unzip2(xys)
     44   xs = []
     45   ys = []
---> 46   for x, y in xys:
     47     xs.append(x)
     48     ys.append(y)

ValueError: not enough values to unpack (expected 2, got 1)
```

I think I'm back to believing we are probably better off keeping things how they are (and closing this PR). WDYT?

mattjj:
(Achievement unlocked: start every sentence with "I think".)

mattjj:
Based on the most recent thinking, namely about erroring on `unzip2([(1,), (2,)])`, I think it's best just to leave these functions as they are.

Thanks again for the PR and the discussion!

PR #6759, opened 2021-05-14
[sparse] add batching rules to COO primitives

### LINKED ISSUE #6544: Jax support for sparse matrices

jakevdp:
This is a work in progress.

The goal is to have functioning batching rules, though they're not necessarily particularly efficient. Long-term the primitives themselves should probably handle batching; this would effectively be a version of a higher-dimensional sparse representation.

I'm thinking of how to best codify that.

In the meantime, these basic batching rules would let us do things like use things like `jacfwd`, `jacrev`, and `hessian` to more efficiently test autodiff rules for these primitives.

### DIFF
--- a/jax/experimental/sparse_ops.py
+++ b/jax/experimental/sparse_ops.py
@@ -31,8 +31,8 @@ Further down are some examples of potential high-level wrappers for sparse objec
 (API should be considered unstable and subject to change).
 """
 import functools
+import itertools
 import operator
-
 from typing import Any, Tuple
 
 from jax import api
@@ -41,6 +41,7 @@ from jax import jit
 from jax import lax
 from jax import tree_util
 from jax.interpreters import xla
+from jax.interpreters import batching
 from jax.lib import cusparse
 from jax.lib import xla_bridge
 from jax.lib import xla_client
@@ -302,6 +303,16 @@ def _coo_todense_impl(data, row, col, *, shape):
 def _coo_todense_abstract_eval(data, row, col, *, shape):
   return core.ShapedArray(shape, data.dtype)
 
+def _coo_todense_batching_rule(batched_args, batch_dims, *, shape):
+  def _batched(x, bdim):
+    if bdim is None:
+      return itertools.repeat(x)
+    else:
+      return list(jnp.rollaxis(x, bdim))
+  batched = tuple(_batched(x, bdim) for x, bdim in zip(batched_args, batch_dims))
+  results = [coo_todense(*args, shape=shape) for args in zip(*batched)]
+  return jnp.stack(results), 0
+
 def _coo_todense_gpu_translation_rule(c, data, row, col, *, shape):
   return cusparse.coo_todense(c, data, row, col, shape=shape)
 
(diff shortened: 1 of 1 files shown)


jakevdp:
Closing in favor of #6824

PR #6763, opened 2021-05-17
Improves jax2tf error message for conv_general_dilated

marcvanzee:
When using `jax2tf` with XLA disabled, `conv_general_dilated` may still convert successfully in some special cases. In those cases where it does not, the variable `info_or_result` contains useful information for debugging why. However, currently this information isn't output. This PR makes sure it is output.

For instance, when trying to convert a [ResNet18](https://github.com/google/flax/blob/master/examples/imagenet/models.py#L121) model, the original error message was:

```
NotImplementedError: Call to conv_general_dilated can only be converted through TFXLA, but XLA is disabled
```

After this PR, the error message is:

```
NotImplementedError: Could not convert conv_general_dilated without using XLA, reason: tf.nn.convolution is not supported for dtype <dtype: 'bfloat16'>
```

This PR also simplifies some of the error raising logic in `_try_tr_conv`.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1209,7 +1209,8 @@ def _conv_general_dilated(lhs, rhs, window_strides, padding, lhs_dilation,
     if not isinstance(info_or_result, str):
       return info_or_result
     else:
-      raise _xla_path_disabled_error("conv_general_dilated")
+      raise NotImplementedError("Could not convert conv_general_dilated "
+                                f"without XLA, reason: {info_or_result}")
 
   dnums_proto = _conv_general_dimension_numbers_proto(dimension_numbers)
   precision_config_proto = _conv_general_precision_config_proto(precision)
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1240,7 +1240,8 @@ def _conv_general_dilated(lhs, rhs, *,
     if not isinstance(info_or_result, str):
       return info_or_result
     else:
-      raise _xla_path_disabled_error("conv_general_dilated")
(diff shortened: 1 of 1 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1120,40 +1120,41 @@ def _precision_config_proto(precision: Optional[Tuple[PrecisionType, PrecisionTy
...
-# _try_tf_conv returns a Tensor when it succeeds, or a string describing why
-# it did not succeed otherwise.
 def _try_tf_conv(lhs, rhs, window_strides, padding, lhs_dilation, rhs_dilation,
                  dimension_numbers, feature_group_count, batch_group_count,
-                 out_shape) -> Union[str, TfVal]:
+                 out_shape) -> TfVal:
+
+  class ConvError(Exception):
+    def __init__(self, e):
+      super().__init__("Couldn't convert conv_general_dilated without XLA: "
```
gnecula, line 1129: I would align this with _xla_path_disabled_error, so that all primitives that cannot be converted with enable_xla=False use the same message pattern.

In #6722 I was going to change the error message globally, and add the ability to add additional messages. That PR got rolled back, because of some issues to XlaPad. I expect to be able to re-land it in a couple of day.

marcvanzee: Alright, perhaps it then makes sense to wait for your PR to land before adding this one?

gnecula: #6755 got merged. Inspired by your change I have added to it the surfacing of the precise reason why convolution cannot be converted. PTAL.

(I hope this won't get rolled back again)

gnecula:
@marcvanzee Is this change still applicable?

PR #6770, opened 2021-05-17
no more post_process, passing core_test

mattjj:
TODO
- [x] fix remat
- [ ] fix scan
- [ ] write PR message
- [ ] (probably in follow-up PR) remove post_process from custom_jvp/vjp


hawkinsp:
Stale, surely?

mattjj:
Thanks, Peter!

PR #6776, opened 2021-05-18
[jax2tf] Added documentation explaining how to handle undefined TF ops

gnecula:
Added a test case showing how to mix compileable and non-compileable code.

### DIFF
--- a/jax/experimental/jax2tf/README.md
+++ b/jax/experimental/jax2tf/README.md
@@ -77,6 +77,14 @@ The Autograph feature of `tf.function` cannot be expected to work on
 functions converted from JAX as above, so it is recommended to
 set `autograph=False` in order to avoid warnings or outright errors.
 
+It is a good idea to use XLA to compile the converted function; that is
+the scenario for which we are optimizing for numerical and performance
+accuracy w.r.t. the JAX execution:
+
+```python
+tf.function(jax2tf.convert(f_jax), autograph=False, jit_compile=True)(x)
+```
+
 ## Usage: saved model
 
 Since jax2tf provides a regular TensorFlow function using it with SavedModel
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/README.md
```diff
@@ -386,16 +394,86 @@ jax2tf.convert(lambda x: jnp.sum(x, axis=0) / x.shape[0],
...
+```python
+jax2tf.convert(lambda x: lax.min(x, x))(np.array([True]))
+
+>>> InvalidArgumentError: Value for attr 'T' of bool is not in the list of allowed values:
+>>>    bfloat16, half, float, double, uint8, int16, int32, int64;
+>>>    NodeDef: {{node Minimum}};
+>>>    Op<name=Minimum; signature=x:T, y:T -> z:T; attr=T:type,allowed=[DT_BFLOAT16, DT_HALF, DT_FLOAT, DT_DOUBLE, DT_UINT8, DT_INT16, DT_INT32, DT_INT64]> [Op:Minimum]
+```
+
+In the above cases, you should file a bug with JAX or TensorFlow, or consider
```
saeta, line 422: Small suggestion: consider being a bit more specific in the recommendation. I think for cases where the TF dtype is not registered, the bug should likely go to TF. Or are you suggesting that JAX should implement a workaround where JAX will lower to a different dtype in the converter?

jax/experimental/jax2tf/README.md
```diff
@@ -386,16 +394,86 @@ jax2tf.convert(lambda x: jnp.sum(x, axis=0) / x.shape[0],
...
+>>>  device='CPU'; T in [DT_FLOAT]
+>>>  device='CPU'; T in [DT_DOUBLE]
+>>>  ...
+```
+
+In this situation, you can still run the converted program if you compile it with
+XLA:
+```python
+tf.function(jax2tf.convert(lax.sign),
+            autograph=True, jit_compile=True)(np.array([5], dtype=np.uint32))
```
saeta, line 445: Here, you have `autograph=True`, whereas before it was set to false. Could you describe (for my own learning) the nuances here?

saeta (approved):
Thanks for sending this along!

jax/experimental/jax2tf/README.md
gnecula: This is a mistake. It never makes sense to set `autograph=True`, because it would ask autograph to make sense of the jax2tf converter. It occasionally gives warnings, and in very rare occasions it fails (this happened last a while ago).

jax/experimental/jax2tf/README.md
gnecula: I think that the solution depends on the case. In some cases the bug should clearly go to TF, e.g., `lax.min` working on `int32` but not `int8`. However, in practice it is still us who fix those bugs, and we are actively working on it. Before long, the remaining cases are those when it is not so clear that the TF op should be registered at a given type. Examples are: comparisons of complex numbers, min/max for booleans. In those cases XLA does work, and has picked a semantics, but TF may not want to go that far. In those cases we might have to add workarounds in the converter.

PR #6777, opened 2021-05-18
Initial implementation of jax.numpy.poly

paul-tqh-nguyen:
This is an initial `jax.numpy.poly` implementation intended to help with https://github.com/google/jax/issues/70. 

There are some differences with the `np.poly` implementation (I've described them in my commit messages), so I anticipate I will need to make changes.

Please let me know if there are any changes I can make to improve this PR.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3847,6 +3847,43 @@ def diagflat(v, k=0):
   res = res.reshape(adj_length,adj_length)
   return res
 
+_POLY_DOC="""\
+This differs from np.poly when an integer array is given.
+np.poly returns a result with dtype float64 in this case.
+jax returns a result with an inexact type, but not necessarily
+float64.
+
+This also differs from np.poly when the input array strictly
+contains pairs of complex conjugates, e.g. [1j, -1j, 1-1j, 1+1j].
+np.poly returns an array with a real dtype in such cases.
+jax returns an array with a complex dtype in such cases.
+"""
+
+@_wraps(np.poly, lax_description=_POLY_DOC)
+def poly(seq_of_zeros):
+  _check_arraylike('poly', seq_of_zeros)
+  seq_of_zeros, = _promote_dtypes_inexact(seq_of_zeros)
+  seq_of_zeros = atleast_1d(seq_of_zeros)
+
+  sh = seq_of_zeros.shape
+  if len(sh) == 2 and sh[0] == sh[1] and sh[0] != 0:
+    # import at runtime to avoid circular import
+    from . import linalg
+    seq_of_zeros = linalg.eigvals(seq_of_zeros)
+
+  if seq_of_zeros.ndim != 1:
+    raise ValueError("input must be 1d or non-empty square 2d array.")
+
+  dt = seq_of_zeros.dtype
+  if len(seq_of_zeros) == 0:
+    return ones((), dtype=dt)
+
+  a = ones((1,), dtype=dt)
+  for k in range(len(seq_of_zeros)):
+    a = convolve(a, array([1, -seq_of_zeros[k]], dtype=dt), mode='full')
+
+  return a
+
 
 @_wraps(np.polyval)
 def polyval(p, x):
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1753,6 +1753,27 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
 
 
   @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_shape={}_rank{}".format(
+        jtu.format_shape_dtype_string(a_shape, dtype), rank),
+       "dtype": dtype, "a_shape": a_shape, "rank": rank}
+      for rank in (1, 2)
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes))
+  def testPoly(self, a_shape, dtype, rank):
+    if dtype in (np.float16, jnp.bfloat16, np.int16):
+      self.skipTest(f"{dtype} gets promoted to {np.float16}, which is not supported.")
+    elif rank == 2 and jtu.device_under_test() in ("tpu", "gpu"):
+      self.skipTest("Nonsymmetric eigendecomposition is only implemented on the CPU backend.")
+    rng = jtu.rand_default(self.rng())
+    tol = { np.int8: 1e-3, np.int32: 1e-3, np.float32: 1e-3, np.float64: 1e-6 }
+    if jtu.device_under_test() == "tpu":
+      tol[np.int32] = tol[np.float32] = 1e-1
+    tol = jtu.tolerance(dtype, tol)
+    args_maker = lambda: [rng(a_shape * rank, dtype)]
+    self._CheckAgainstNumpy(np.poly, jnp.poly, args_maker, check_dtypes=False, tol=tol)
+    self._CompileAndCheck(jnp.poly, args_maker, check_dtypes=True, rtol=tol, atol=tol)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "a_shape={} , b_shape={}".format(
           jtu.format_shape_dtype_string(a_shape, dtype),
           jtu.format_shape_dtype_string(b_shape, dtype)),
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -286,6 +286,7 @@ Not every function in NumPy is implemented; contributions are welcome!
     pad
     percentile
     piecewise
+    poly
     polyadd
     polyder
     polyint
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -51,7 +51,7 @@ from jax._src.numpy.lax_numpy import (
     nanmax, nanmean, nanmin, nanprod, nanstd, nansum, nanvar, ndarray, ndim,
     negative, newaxis, nextafter, nonzero, not_equal, number,
     object_, ogrid, ones, ones_like, operator_name, outer, packbits, pad, percentile,
-    pi, piecewise, polyadd, polyder, polyint, polymul, polysub, polyval, positive, power,
+    pi, piecewise, poly, polyadd, polyder, polyint, polymul, polysub, polyval, positive, power,
     prod, product, promote_types, ptp, quantile,
     r_, rad2deg, radians, ravel, ravel_multi_index, real, reciprocal, remainder, repeat, reshape,
     result_type, right_shift, rint, roll, rollaxis, rot90, round, row_stack,


paul-tqh-nguyen:
@googlebot I fixed it.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
   res = res.reshape(adj_length,adj_length)
   return res
 
+@_wraps(np.poly)
+def poly(seq_of_zeros):
+  seq_of_zeros = atleast_1d(seq_of_zeros)
+
+  dt = seq_of_zeros.dtype
+  seq_of_zeros = seq_of_zeros.astype(np.mintypecode(dt.char))
```
jakevdp, line 3855: Is this line necessary? In what contexts will it result in the type of the array being changed?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
   res = res.reshape(adj_length,adj_length)
   return res
 
+@_wraps(np.poly)
+def poly(seq_of_zeros):
```
jakevdp, line 3851: Please add a call to `_check_arraylike` on the input (we generally don't want to accept Python lists as inputs to JAX functions)

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+  seq_of_zeros = atleast_1d(seq_of_zeros)
+
+  dt = seq_of_zeros.dtype
+  seq_of_zeros = seq_of_zeros.astype(np.mintypecode(dt.char))
+
+  sh = seq_of_zeros.shape
+  if len(sh) == 2 and sh[0] == sh[1] != 0:
```
jakevdp, line 3858: I find this kind of chaining hard to read... I'd lean toward using another `and`

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+    # import at runtime to avoid circular import
+    from . import linalg
+    seq_of_zeros = linalg.eigvals(seq_of_zeros)
+
+  if len(seq_of_zeros.shape) != 1:
```
jakevdp, line 3863: `if seq_of_zeros.ndim != 1`

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+    raise ValueError("input must be 1d or non-empty square 2d array.")
+
+  if len(seq_of_zeros) == 0:
+    return 1.0
```
jakevdp, line 3867: consider returning `jnp.ones((), dtype=dt)` so the return type is consistent

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+  dt = seq_of_zeros.dtype
+  a = ones((1,), dtype=dt)
+  for k in range(len(seq_of_zeros)):
+    a = convolve(a, array([1, -seq_of_zeros[k]], dtype=dt), mode='full')
+
+  if issubclass(a.dtype.type, complexfloating):
```
jakevdp, line 3873: use `np.issubdtype(a.dtype, np.complexfloating)`

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+    roots = asarray(seq_of_zeros, complex)
```
jakevdp, line 3874: I believe passing `dtype=complex` here will surface a warning in X32 mode. Is there any reason not to use `a.dtype` here instead?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3847,6 +3847,36 @@ def diagflat(v, k=0):
...
+    roots_are_real = all(sort(roots) == sort(roots.conjugate()))
```
jakevdp, line 3875: I believe this check is not doing what you intend: the dtype of the returned value cannot depend on traced quantities, so the output will have a complex dtype regardless. I think this is probably a break from numpy's API that should be documented in the function's docstring via `lax_description` (search in the file to see how it's used)

Also, I think the logic of this check may be broken. e.g. if `roots=[1j, -1j]`, then `roots_are_real` is True.

tests/lax_numpy_test.py
```diff
@@ -1752,6 +1752,27 @@ def testTrimZeros(self, a_shape, dtype, trim):
...
+       "dtype": dtype, "a_shape": a_shape, "rank": rank}
+      for rank in (1, 2)
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes))
+  def testPoly(self, a_shape, dtype, rank):
+    if np.mintypecode(dtype(0).dtype.char) == 'd' and not config.x64_enabled:
+      self.skipTest(f"Only run {dtype} testcase when float64 is enabled.")
+    elif dtype == np.float32 and not config.x64_enabled and rank == 2:
+      self.skipTest(f"Only run {dtype} testcase with rank 2 when "
+                    "float64 is enabled (due to the need for complex64).")
```
jakevdp, line 1767: `complex64` is available with X64 disabled. You should be able to remove this skip (I suspect the reason you saw it fail is because of the warning addressed in the comment above)

tests/lax_numpy_test.py
```diff
@@ -1752,6 +1752,27 @@ def testTrimZeros(self, a_shape, dtype, trim):
...
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_shape={}_rank{}".format(
+        jtu.format_shape_dtype_string(a_shape, dtype), rank),
+       "dtype": dtype, "a_shape": a_shape, "rank": rank}
+      for rank in (1, 2)
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes))
+  def testPoly(self, a_shape, dtype, rank):
+    if np.mintypecode(dtype(0).dtype.char) == 'd' and not config.x64_enabled:
+      self.skipTest(f"Only run {dtype} testcase when float64 is enabled.")
```
jakevdp, line 1764: I don't think this check should be needed... `default_dtypes` already skips unavailable types.

tests/lax_numpy_test.py
```diff
@@ -1752,6 +1752,27 @@ def testTrimZeros(self, a_shape, dtype, trim):
...
+    elif dtype == np.float32 and not config.x64_enabled and rank == 2:
+      self.skipTest(f"Only run {dtype} testcase with rank 2 when "
+                    "float64 is enabled (due to the need for complex64).")
+    elif dtype == np.float16:
+      self.skipTest(f"{dtype} not supported in numpy.")
```
jakevdp, line 1769: Perhaps skip `CheckAgainstNumpy`, but still run `CompileAndCheck`?

jakevdp (changes requested):
Looks great, thanks for tackling this! A number of comments below – please let me know if anything is unclear!

jax/_src/numpy/lax_numpy.py
jakevdp: Seems like it's probably not something we want; e.g.
```python
>>> dtype = np.dtype(jnp.bfloat16)                                                                                                                           
>>> np.dtype(np.mintypecode(np.dtype(jnp.bfloat16).char))                                                                                                    
dtype('float64')
```

paul-tqh-nguyen: In https://github.com/numpy/numpy/issues/5096, there's an issue where `np.poly` doesn't work as expected for integer arrays, so the solution that they implemented to fix it was to turn them into `float64` arrays.

I wrote this implementation to try and be similar to numpy's implementation (even if it might not be the best). I agree that this line is probably not-ideal behavior, so I can change that if it's okay for there to be a tiny difference between jax's behavior and numpy's behavior.

jakevdp: I see, that makes sense. In JAX we generally handle that using `_promote_dtypes_inexact`, which will do the promotion to the appropriate float type following JAX's type promotion logic.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: My changes in https://github.com/google/jax/pull/6777/commits/30633e3839cef7c7d29a72a90d1aeee7b8d9c701 implement this suggestion.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: My changes in https://github.com/google/jax/pull/6777/commits/30633e3839cef7c7d29a72a90d1aeee7b8d9c701 implement this suggestion.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: My changes in https://github.com/google/jax/pull/6777/commits/30633e3839cef7c7d29a72a90d1aeee7b8d9c701 implement this suggestion.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: Upon investigating this further, I think we can remove this whole `if` block. 

The short explanation for this is that all this `if` block does is this: when all the imaginary parts are zero, throw away the imaginary parts of the answer by casting it to a real dtype.

I added this in my initial commit because I ported it over without looking at it in deeply. Since we can't change the type of an array based on its contents in JAX (please correct me if I'm wrong), we can't do what numpy does. Thus, we should just remove the whole block. 

I apologize for not finding this earlier. This was an oversight on my part. I should've figured out whether this was necessary when doing my first commit. I removed it in https://github.com/google/jax/commit/30633e3839cef7c7d29a72a90d1aeee7b8d9c701.

Here's the long explanation (feel free to ignore this as it's just for extra clarity):
<details>

<ul>
<li><code>np.poly</code> returns the coefficients of a polynomial whose zeros/roots are the inputs. </li>
<li>Let&#39;s call the returned result <code>c</code>. <code>c</code> represents the polynomial <code>c[0] * x**(N) + c[1] * x**(N-1) + ... + c[N-1] * x + c[N]</code>.</li>
<li>Because <code>c[0]</code> is always 1 by <a href="https://numpy.org/doc/stable/reference/generated/numpy.poly.html">the definition of <code>np.poly</code></a>, the polynomial can be written in the form: <ul>
<li><code>(x-seq_of_zeros[0])*(x-seq_of_zeros[1])...</code></li>
<li>Note how the coefficient of all the <code>x</code> terms is 1.</li>
</ul>
</li>
<li><code>roots_are_real</code> tells us if <code>seq_of_zeros</code> are all complex conjugates.</li>
<li>If <code>roots_are_real</code> is <code>True</code>, then the polynomial can be rewritten as:<ul>
<li><code>(x-seq_of_zeros[A])*(x-seq_of_zeros[A_CONJ])*(x-seq_of_zeros[B])*(x-seq_of_zeros[B_CONJ])*...</code></li>
<li>where <code>seq_of_zeros[A]</code> and <code>seq_of_zeros[A_CONJ]</code> are conjugates (and similarly for <code>B</code> and all the rest).</li>
</ul>
</li>
<li>Since <code>(x-seq_of_zeros[A])*(x-seq_of_zeros[A_CONJ])</code> when expanded out will have all real coefficients (and similarly for <code>B</code> and all the rest), the whole polynomial will have all real coefficients. </li>
<li>Thus, <code>roots_are_real</code> just tells us whether or not the polynomial has all real coefficients. </li>
<li>In such cases, we can safely throw away all the imaginary parts of our answer and can safely cast our answer to have a real type.</li>
</ul>
</details>

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: I think that this whole `if` block is not needed; I removed it in https://github.com/google/jax/commit/30633e3839cef7c7d29a72a90d1aeee7b8d9c701. I provide an explanation at https://github.com/google/jax/pull/6777#discussion_r634993656 about why we can remove this `if` block.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: https://github.com/google/jax/commit/30633e3839cef7c7d29a72a90d1aeee7b8d9c701 implements this suggestion.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: I think that this whole `if` block is not needed; I removed it in https://github.com/google/jax/commit/30633e3839cef7c7d29a72a90d1aeee7b8d9c701. I provide an explanation at https://github.com/google/jax/pull/6777#discussion_r634993656 about why we can remove this `if` block.

tests/lax_numpy_test.py
paul-tqh-nguyen: Yes, that sounds right. I removed this skip in https://github.com/google/jax/commit/d7a0190b545475e4140feebf411082c4b2b10245.

tests/lax_numpy_test.py
jakevdp: Clarification: this check should not be needed once the `mintypecode` logic in the function is replaced with `_promote_dtypes_inexact`.

paul-tqh-nguyen: Yes, that makes sense. https://github.com/google/jax/commit/79d331db2fa4275f0cea4c53131c087398b9fb56 implements this suggestion.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: https://github.com/google/jax/commit/79d331db2fa4275f0cea4c53131c087398b9fb56 uses `_promote_dtypes_inexact`, which removes the need for `np.mintypecode`.

tests/lax_numpy_test.py
paul-tqh-nguyen: It seems that we cannot run `CompileAndCheck` since `jnp.linalg.eigvals` is currently not implemented for float16. 
```

>>> jnp.linalg.eigvals(jnp.array([[1,2],[3,4]], dtype=jnp.float16))
a DeviceArray([[1., 2.],
             [3., 4.]], dtype=float16)
  File "/Users/pnguyen/code/jax/jax/_src/numpy/linalg.py", line 294, in eigvals
    compute_right_eigenvectors=False)[0]
  File "/Users/pnguyen/code/jax/jax/_src/lax/linalg.py", line 82, in eig
    return eig_p.bind(x, compute_left_eigenvectors=compute_left_eigenvectors,
  File "/Users/pnguyen/code/jax/jax/core.py", line 264, in bind
    out = top_trace.process_primitive(self, tracers, params)
  File "/Users/pnguyen/code/jax/jax/core.py", line 603, in process_primitive
    return primitive.impl(*tracers, **params)
  File "/Users/pnguyen/code/jax/jax/_src/lax/linalg.py", line 374, in eig_impl
    xla.apply_primitive(eig_p, operand,
  File "/Users/pnguyen/code/jax/jax/interpreters/xla.py", line 234, in apply_primitive
    compiled_fun = xla_primitive_callable(prim, *unsafe_map(arg_spec, args), **params)
  File "/Users/pnguyen/code/jax/jax/_src/util.py", line 186, in wrapper
    return cached(config._trace_context(), *args, **kwargs)
  File "/Users/pnguyen/code/jax/jax/_src/util.py", line 179, in cached
    return f(*args, **kwargs)
  File "/Users/pnguyen/code/jax/jax/interpreters/xla.py", line 276, in xla_primitive_callable
    built_c = primitive_computation(prim, AxisEnv(nreps, (), ()), backend,
  File "/Users/pnguyen/code/jax/jax/_src/util.py", line 186, in wrapper
    return cached(config._trace_context(), *args, **kwargs)
  File "/Users/pnguyen/code/jax/jax/_src/util.py", line 179, in cached
    return f(*args, **kwargs)
  File "/Users/pnguyen/code/jax/jax/interpreters/xla.py", line 317, in primitive_computation
    ans = rule(c, *xla_args, **params)
  File "/Users/pnguyen/code/jax/jax/_src/lax/linalg.py", line 414, in eig_cpu_translation_rule
    w, vl, vr, info = _cpu_geev(c, operand, jobvl=compute_left_eigenvectors,
  File "jaxlib/lapack.pyx", line 1785, in lapack.geev
    raise NotImplementedError("Unsupported dtype {}".format(dtype))
NotImplementedError: Unsupported dtype float16
>>> 
```
Since this is a `NotImplementedError`, should I just skip `CheckAgainstNumpy` and call`pytest.xfail` right before running `CompileAndCheck` when the dtype is float16?

jakevdp: If float16 is not compatible with either compilation or numpy, it's probably fine to skip the test (or ensure that float16 is left out of the parametrization above)

paul-tqh-nguyen: That makes sense. I simply skip the test in https://github.com/google/jax/commit/d3c3cfb43eeaa9bf30944b6550be0097ceefafd5.

jax/_src/numpy/lax_numpy.py
paul-tqh-nguyen: https://github.com/google/jax/commit/d3c3cfb43eeaa9bf30944b6550be0097ceefafd5 documents the differences with numpy's behavior via `lax_description`.

jakevdp (approved):
Looks good, thanks for tackling this!

Let's make sure tests pass, and then we can get this merged.

jakevdp:
It looks like some tolerances need to be adjusted.

Additionally, we need to skip the rank=2 tests on GPU and TPU, becuase non-symmetric eigenvalue decomposition is only supported on CPU:
```pytb
"/build/work/737bb96d2bd1cba8ca8fa60e671f42d64a4c/google3/runfiles/google3/third_party/py/jax/_src/lax/linalg.py", line 381, in eig_translation_rule
    "Nonsymmetric eigendecomposition is only implemented on the CPU backend")
NotImplementedError: Nonsymmetric eigendecomposition is only implemented on the CPU backend
```

tests/lax_numpy_test.py
```diff
@@ -1752,6 +1752,25 @@ def testTrimZeros(self, a_shape, dtype, trim):
...
+      for rank in (1, 2)
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes))
+  def testPoly(self, a_shape, dtype, rank):
+    if dtype in (np.float16, jnp.bfloat16, np.int16):
+      self.skipTest(f"{dtype} gets promoted to {np.float16}, which is not supported.")
+    elif rank == 2 and jtu.device_under_test() in ('tpu', 'gpu'):
+      self.skipTest("Nonsymmetric eigendecomposition is only implemented on the CPU backend.")
+    rng = jtu.rand_default(self.rng())
+    tol = { np.int8: 1e-4, np.int32: 1e-3, np.float32: 1e-3, np.float64: 1e-6 }
```
jakevdp, line 1768: A couple more tolernace updates are needed for the larger test suite:
- change the int8 tolerance to 1e-3
- on TPU, it looks like int32 & float32 tolerances need to be `0.1` for the tests to pass

paul-tqh-nguyen: I've updated the tolerances in https://github.com/google/jax/commit/e985c135764c65f831235d9a4a6f16bc368670bb. I made the int32 and float32 tolerances specific to the TPU in [this line](https://github.com/google/jax/commit/e985c135764c65f831235d9a4a6f16bc368670bb#diff-6413a73e50360a19ffcf338449b206c057e47f6d66eb0a40be5610e6baff6420R1769).

jakevdp:
Alright, it's looking good! We should be able to merge this soon.

Thanks for the contribution, and for sticking with this through all the back and forth!

PR #6785, opened 2021-05-19
Make Jax linesearch implementation closer to Scipy

### LINKED ISSUE #5284: JAX (vs Autograd) convergence and runtime problems?
I built a toy optimization script comparing JAX with Autograd in the minimization of a function using scipy.optimize.BFGS. Two issues:

1. with a start point x0 further from the minimum BFGS with JAX gradients failed to converge, but it converged with Autograd gradients.

2. with a start point x1 closer to the minimum BFGS converged with both gradients. However its runtime was 2.2 times slower with JAX gradients, despite using jit.

Why could this be happening?

Using the code below I get the following output:

> JAX converged with x0?: False
> JAX converged with x1?: True
> JAX elapsed time with x1: 1.44 sec
> 
> Autograd converged with x0?: True
> Autograd converged with x1?: True
> Autograd lapsed time with x1: 0.65 sec

```
import time
import scipy.optimize
import jax.numpy as jnp
import autograd.numpy as anp
import jax
import autograd

def rosenbrock(x):
    answer = sum(100.0*(x[1:]-x[:-1]**2.0)**2.0 + (1-x[:-1])**2.0)
    return answer

jEvalFunc = jax.jit(rosenbrock)
jGradFunc= jax.jit(jax.grad(rosenbrock))
aEvalFunc = rosenbrock
aGradFunc = autograd.grad(rosenbrock)

x0 = [76.0, 97.0, 20.0, 120.0, 0.01, 1e4]
x1 = [76.0, 97.0, 20.0, 120.0, 0.01, 1e2]
jx0 = jnp.array(x0)
jx1 = jnp.array(x1)
ax0 = anp.array(x0)
ax1 = anp.array(x1)

jOptimRes_x0 = scipy.optimize.minimize(fun=jEvalFunc, x0=jx0, method='BFGS', jac=jGradFunc)
startTime = time.time()
jOptimRes_x1 = scipy.optimize.minimize(fun=jEvalFunc, x0=jx1, method='BFGS', jac=jGradFunc)
jElapsedTime = time.time()-startTime

aOptimRes_x0 = scipy.optimize.minimize(fun=aEvalFunc, x0=ax0, method='BFGS', jac=aGradFunc)
startTime = time.time()
aOptimRes_x1 = scipy.optimize.minimize(fun=aEvalFunc, x0=ax1, method='BFGS', jac=aGradFunc)
aElapsedTime = time.time()-startTime

print("JAX converged with x0?: {}".format(jOptimRes_x0.success))
print("JAX converged with x1?: {}".format(jOptimRes_x1.success))
print("JAX elapsed time with x1: {:0.2f} sec".format(jElapsedTime))
print("")
print("Autograd converged with x0?: {}".format(aOptimRes_x0.success))
print("Autograd converged with x1?: {}".format(aOptimRes_x1.success))
print("Autograd lapsed time with x1: {:0.2f} sec".format(aElapsedTime))
```

### LINKED ISSUE #5292: jax.scipy.optimize.minimize convergence problem
I attempted to minimize a simple test function with scipy.optimize.minimize and with jax.scipy.optimize.minimize. I used identical parameters and start point. scipy.optimize.minimize converged to the function minimum but jax.scipy.optimize.minimize did not.

```python
import time
import scipy.optimize
import jax
import jax.numpy as jnp
import jax.scipy.optimize
import jax.config
import autograd.numpy as anp
import autograd

def zakharovFromIndices(x, ii):
    sum1 = (x**2).sum()
    sum2 = (0.5*ii*x).sum()
    answer = sum1+sum2**2+sum2**4
    return answer

def zakharov_jaxNumpy(x):
    ii = jnp.arange(1, len(x)+1, step=1)
    answer = zakharovFromIndices(x=x, ii=ii)
    return answer

def zakharov_autogradNumpy(x):
    ii = anp.arange(1, len(x)+1, step=1)
    answer = zakharovFromIndices(x=x, ii=ii)
    return answer

jEvalFunc = jax.jit(zakharov_jaxNumpy)
aEvalFunc = zakharov_autogradNumpy
aGradFunc = autograd.grad(aEvalFunc)

x0 = [600.0, 700.0, 200.0, 100.0, 90.0, 1e4]
toleranceChange = 1e-9
maxIter = 10000
jx0 = jnp.array(x0)
ax0 = anp.array(x0)

aOptimRes_x0 = scipy.optimize.minimize(fun=aEvalFunc, x0=ax0, method='BFGS', jac=aGradFunc)
jOptimRes_x0 = jax.scipy.optimize.minimize(fun=jEvalFunc, x0=jx0, method='BFGS')

print("scipy.optimize converged?: {}".format(aOptimRes_x0.fun<1e-6))
print("jax.scipy.optimize converged?: {}".format(jOptimRes_x0.fun<1e-6))
```

### LINKED ISSUE #6756: jax.scipy.optimize hangs
I'm comparing performance for the scipy BFGS implementation and the replica in jax, but the one in jax keeps hanging and/or being 100x slower. To make sure this is not only my system, I have verified that the code below hangs on Colab.
```python
import jax
import jax.numpy as jnp
import jax.scipy.optimize
import jax.random

jX = jnp.arange(-2, 8, 0.03) # A 1D evaluation grid
targetp = jnp.array(
    [1.3615158,  3.1700504,  0.3901142,  0.8582838,
     -0.24486497,  3.078153,  1.2610698,  2.276195,
     0.84759176,  1.563921, -0.91339713,  0.23969328]) # The "true" parameters

def fun(p): # This function was once auto-generated, so it looks silly
    arg1 = p[2] + p[3] * jnp.tanh(p[4]  + p[5] * jX)
    arg2 = p[6] + p[7] * jnp.log(jnp.abs(p[8] + p[9] * jnp.arctan(p[10] + p[11] * jX)))
    return p[0] + p[1] * arg1 * arg2

targety = fun(targetp)
def mse_loss(p):
    return ((fun(p) - targety) ** 2).mean()
```

I can optimize `mse_loss` just fine in scipy.optimize.minimize, but when I run the code below, the interpreter hangs on the second iteration:
```python
key = jax.random.PRNGKey(1337)
keys = jax.random.split(key, 10)

for k in keys: # Try optimize from different starting points
    x0 = targetp + jax.random.normal(k)
    optx = jax.scipy.optimize.minimize(mse_loss, x0, method='BFGS')
    print(optx.fun, optx.nfev)
```

I have verified this with and without GPU and with and without jitted functions.

[This](https://colab.research.google.com/drive/1sBwLh5w_XRDABHJB2SrmwZtOTXu33EOD) is a link to the above code in Colab.

GregCT:
Hi,

This contains a couple of changes to make Jax's BFGS implementation closer to Scipy's.
Specifically:
- Adds a maxiters to the _zoom function, which [matches Scipy](https://github.com/scipy/scipy/blob/2a4942d3ff21459e5bd66de7db9adc48fe571855/scipy/optimize/linesearch.py#L616) and will prevent optimization getting stuck in an infinite loop (should close #6756)
- Updates the line search to use [Scipy's heuristic](https://github.com/scipy/scipy/blob/2a4942d3ff21459e5bd66de7db9adc48fe571855/scipy/optimize/linesearch.py#L403) for the starting point, rather than just starting from 1
- Updates line search so that `dalpha` is always calculated as hi - lo [(as in Scipy)](https://github.com/scipy/scipy/blob/2a4942d3ff21459e5bd66de7db9adc48fe571855/scipy/optimize/linesearch.py#L565), regardless of min / max functions
- Small fixes so that internal state is propagated correctly in the line search
- Lowers the threshold for the saddle point, I think in practice this will probably now never fire and could just be removed

The various line search tweaks should also resolve #5284 and resolve #5292, and the change includes a test to ensure that this is the case.

Thanks,
Greg

### DIFF
--- a/jax/_src/scipy/optimize/bfgs.py
+++ b/jax/_src/scipy/optimize/bfgs.py
@@ -55,6 +55,7 @@ class _BFGSResults(NamedTuple):
   f_k: jnp.ndarray
   g_k: jnp.ndarray
   H_k: jnp.ndarray
+  old_old_fval: jnp.ndarray
   status: Union[int, jnp.ndarray]
   line_search_status: Union[int, jnp.ndarray]
 
@@ -108,6 +109,7 @@ def minimize_bfgs(
       f_k=f_0,
       g_k=g_0,
       H_k=initial_H,
+      old_old_fval=f_0 + jnp.linalg.norm(g_0) / 2,
       status=0,
       line_search_status=0,
   )
@@ -124,6 +126,7 @@ def minimize_bfgs(
         state.x_k,
         p_k,
         old_fval=state.f_k,
+        old_old_fval=state.old_old_fval,
         gfk=state.g_k,
         maxiter=line_search_maxiter,
     )
@@ -153,7 +156,8 @@ def minimize_bfgs(
         x_k=x_kp1,
         f_k=f_kp1,
         g_k=g_kp1,
-        H_k=H_kp1
+        H_k=H_kp1,
+        old_old_fval=state.f_k,
     )
     return state
 
--- a/jax/_src/scipy/optimize/line_search.py
+++ b/jax/_src/scipy/optimize/line_search.py
@@ -104,15 +104,16 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
   def body(state):
     # Body of zoom algorithm. We use boolean arithmetic to avoid using jax.cond
     # so that it works on GPU/TPU.
+    dalpha = (state.a_hi - state.a_lo)
     a = jnp.minimum(state.a_hi, state.a_lo)
     b = jnp.maximum(state.a_hi, state.a_lo)
-    dalpha = (b - a)
     cchk = delta1 * dalpha
     qchk = delta2 * dalpha
 
     # This will cause the line search to stop, and since the Wolfe conditions
     # are not satisfied the minimization should stop too.
-    state = state._replace(failed=state.failed | (dalpha <= 1e-10))
+    threshold = jnp.where((jnp.finfo(dalpha).bits < 64), 1e-5, 1e-10)
+    state = state._replace(failed=state.failed | (dalpha <= threshold))
 
     # Cubmin is sometimes nan, though in this case the bounds check will fail.
     a_j_cubic = _cubicmin(state.a_lo, state.phi_lo, state.dphi_lo, state.a_hi,
@@ -169,9 +170,9 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
             hi_to_lo,
             state._asdict(),
             dict(
-                a_hi=a_lo,
-                phi_hi=phi_lo,
-                dphi_hi=dphi_lo,
+                a_hi=state.a_lo,
+                phi_hi=state.phi_lo,
+                dphi_hi=state.dphi_lo,
                 a_rec=state.a_hi,
                 phi_rec=state.phi_hi,
             ),
@@ -191,6 +192,9 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
         ),
     )
     state = state._replace(j=state.j + 1)
+    # Choose higher cutoff for maxiter than Scipy as Jax takes longer to find
+    # the same value - possibly floating point issues?
+    state = state._replace(failed= state.failed | state.j >= 30)
     return state
 
   state = while_loop(lambda state: (~state.done) & (~pass_through) & (~state.failed),
@@ -213,7 +217,6 @@ class _LineSearchState(NamedTuple):
   phi_star: Union[float, jnp.ndarray]
   dphi_star: Union[float, jnp.ndarray]
   g_star: jnp.ndarray
-  saddle_point: Union[bool, jnp.ndarray]
 
 
 class _LineSearchResults(NamedTuple):
@@ -269,6 +272,11 @@ def line_search(f, xk, pk, old_fval=None, old_old_fval=None, gfk=None, c1=1e-4,
   else:
     phi_0 = old_fval
     dphi_0 = jnp.dot(gfk, pk)
+  if old_old_fval is not None:
+    candidate_start_value = 1.01 * 2 * (phi_0 - old_old_fval) / dphi_0
+    start_value = jnp.where(candidate_start_value > 1, 1.0, candidate_start_value)
+  else:
+    start_value = 1
 
   def wolfe_one(a_i, phi_i):
     # actually negation of W1
@@ -292,18 +300,12 @@ def line_search(f, xk, pk, old_fval=None, old_old_fval=None, gfk=None, c1=1e-4,
       phi_star=phi_0,
       dphi_star=dphi_0,
       g_star=gfk,
-      saddle_point=False,
   )
 
   def body(state):
     # no amax in this version, we just double as in scipy.
     # unlike original algorithm we do our next choice at the start of this loop
-    a_i = jnp.where(state.i == 1, 1., state.a_i1 * 2.)
-    # if a_i <= 0 then something went wrong. In practice any really small step
-    # length is a failure. Likely means the search pk is not good, perhaps we
-    # are at a saddle point.
-    saddle_point = a_i < 1e-5
-    state = state._replace(failed=saddle_point, saddle_point=saddle_point)
+    a_i = jnp.where(state.i == 1, start_value, state.a_i1 * 2.)
 
     phi_i, dphi_i, g_i = restricted_func_and_grad(a_i)
     state = state._replace(nfev=state.nfev + 1,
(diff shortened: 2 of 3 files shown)


hawkinsp:
It looks like a number of tests fail with this change.

jax/_src/scipy/optimize/bfgs.py
```diff
@@ -133,7 +136,15 @@ def body_fun(state):
...
         line_search_status=line_search_results.status,
     )
-    s_k = line_search_results.a_k * p_k
+    alpha_k = line_search_results.a_k
+    # If the step is too small in 32 bit mode then the optimizer can get stuck
+    # with a direction of zero - ensure that doesn't happen by setting a
+    # minimum step size.
+    if not jax.config.x64_enabled:
+      alpha_k = jnp.where(jnp.abs(alpha_k) < 1e-8,
+                          jnp.sign(alpha_k) * 1e-8,
```
shoyer, line 145: These thresholds looks suspiciously close to zero for float32. As a general rule of thumb, floating point arithmetic is accuracy up to some multiple of `np.finfo(dtype).eps`, which is roughly 1e-7 for float32 and 2e-16 for float64:
https://en.wikipedia.org/wiki/Machine_epsilon

jax/_src/scipy/optimize/line_search.py
```diff
@@ -298,11 +306,11 @@ def wolfe_two(dphi_i):
   def body(state):
     # no amax in this version, we just double as in scipy.
     # unlike original algorithm we do our next choice at the start of this loop
-    a_i = jnp.where(state.i == 1, 1., state.a_i1 * 2.)
+    a_i = jnp.where(state.i == 1, start_value, state.a_i1 * 2.)
     # if a_i <= 0 then something went wrong. In practice any really small step
     # length is a failure. Likely means the search pk is not good, perhaps we
     # are at a saddle point.
-    saddle_point = a_i < 1e-5
+    saddle_point = a_i < 1e-20
```
shoyer, line 313: 1e-20 seems suspiciously small for float32

jax/_src/scipy/optimize/bfgs.py
```diff
@@ -133,7 +136,15 @@ def body_fun(state):
         failed=line_search_results.failed,
         line_search_status=line_search_results.status,
     )
-    s_k = line_search_results.a_k * p_k
+    alpha_k = line_search_results.a_k
+    # If the step is too small in 32 bit mode then the optimizer can get stuck
+    # with a direction of zero - ensure that doesn't happen by setting a
+    # minimum step size.
+    if not jax.config.x64_enabled:
```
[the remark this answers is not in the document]
GregCT: I'm not 100% sure if the input type matters - because the jax arrays which have precision issues are all created inside the function, rather than the input arrays.

As another datapoint the test I wrote to test passing in a float32 input didn't compile, because the output type from the loop (float64 array created inside the function) was different to the input type (float32 array created outside).

Based on that I'd be inclined to leave it as it is, but obviously defer to you if you would prefer testing the input type?

jax/_src/scipy/optimize/line_search.py
```diff
@@ -104,9 +104,9 @@ def _zoom(restricted_func_and_grad, wolfe_one, wolfe_two, a_lo, phi_lo,
   def body(state):
     # Body of zoom algorithm. We use boolean arithmetic to avoid using jax.cond
     # so that it works on GPU/TPU.
+    dalpha = (state.a_hi - state.a_lo)
     a = jnp.minimum(state.a_hi, state.a_lo)
     b = jnp.maximum(state.a_hi, state.a_lo)
-    dalpha = (b - a)
     cchk = delta1 * dalpha
     qchk = delta2 * dalpha
```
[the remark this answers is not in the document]
GregCT: Makes sense - done 👍

jax/_src/scipy/optimize/bfgs.py
```diff
@@ -133,7 +136,15 @@ def body_fun(state):
...
+      alpha_k = jnp.where(jnp.abs(alpha_k) < 1e-8,
+                          jnp.sign(alpha_k) * 1e-8,
+                          alpha_k)
```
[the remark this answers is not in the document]
GregCT: Done

jax/_src/scipy/optimize/line_search.py
GregCT: Yep - higher thresholds were causing the optimizer to fail where it would have been able to converge given more time. However I was a bit wary of taking it out entirely (Chesterton's fence). I've deleted it in the updated commit, and can always bring it back if removing it does cause issues.

jax/_src/scipy/optimize/line_search.py
```diff
@@ -384,25 +386,28 @@ def body(state):
...
               state.i > maxiter,
               jnp.array(3),  # maxiter reached
               jnp.array(0),  # passed (should be)
           ),
-      ),
   )
+  # Step sizes which are too small causes the optimizer to get stuck with a
+  # direction of zero in  <64 bit mode - avoid this with  a minimum step size.
+  alpha_k = state.a_star
+  if not jax.config.x64_enabled:
```
shoyer, line 400: Is there any way we can detect this based on the `dtype` of some arrays? Or perhaps we can just do it in float64 mode too (with a much smaller threshold)?

We'd really like JAX to treat float32 values in a fully consistent way even if x64 mode is enabled, at least if the function is carefully written to only use float32 numbers internally and all the inputs are float32.

Even if that isn't true of the current `jax.scipy.optimize.minimize` routines, I would prefer not to introduce this explicit dependency on `jax.config`, which would be a step in the wrong direction.

shoyer (commented):
PR is looking like a great improvement! I just have my one persistent complaint about checking for x64 mode :)

jax/_src/scipy/optimize/line_search.py
GregCT: That makes sense - I've updated it to be dependent on the type of the array.

I used the number of bits from `jnp.finfo` so that it would also work for bfloat16, etc.

Happy to change that if there's a more standard way of checking the precision of `dtype` for arrays - I couldn't find an obvious helper but may have been looking in the wrong place.

hawkinsp:
@shoyer @froystig Is this ready to merge?

shoyer:
> @shoyer @froystig Is this ready to merge?

Yes, I think so!

shoyer (approved):

PR #6798, opened 2021-05-20
[jax2tf] Initial support for kwargs.

### LINKED ISSUE #6791: jax2tf does not preserve argument names
When converting a function from JAX to TensorFlow using experimental `jax2tf.convert`, the function arguments are not preserved, thus it is not possible to call the code with named instead of positional arguments.

For example, the following snippet will break:

```python
    def f_jax(x):
        return jnp.sum(x)
    
    f_tf = jax2tf.convert(f_jax)

    x = jnp.zeros(3)
    print(f_tf(x=x))
```

```
  File "test.py", line 8, in <module>
    print(f_tf(x=x))
TypeError: converted_fun() got an unexpected keyword argument 'x'
```

My workaround so far has been to wrap the converted function around another function, and then proceed to export the saved model. 

```python
    def f_jax(x):
        return jnp.sum(x)
    
    f_tf = jax2tf.convert(f_jax)
    f_wrap = lambda x: f_tf(x)

    x = jnp.zeros(3)
    print(f_wrap(x=x))
```

Since jax2tf is experimental, I expected to find some issues. However, this one is not listed as a [known issue](https://github.com/google/jax/tree/master/jax/experimental/jax2tf#known-issues). An important use case for this is to allow models with multiple inputs to be called via Tensorflow Serving.

saeta:
Fixes #6791


saeta:
@gnecula PTAL? I'm sure there's lots I'm doing wrong; feedback welcome! :-)

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -206,19 +206,23 @@ def check_arg(a):
...
 
     # This function may take pytrees of TfVals. We can only set
     # tf.custom_gradient on functions that take a flat argument list.
-    args_flat, in_tree = tree_util.tree_flatten((args, {}))
+    args_flat, in_tree = tree_util.tree_flatten((args, kwargs))
 
     if polymorphic_shapes is None:
       polymorphic_shapes_ = (None,) * len(args)
     else:
+      if kwargs:
```
gnecula, line 224: I think that what may be better to do here is to say that we support polymorphic_shapes only for the positional arguments, and do not raise this error. The check in line 226 should still work, and your changes below in line 236 should take care of setting the kwargs to be shape monomorphic. So, the only thing to do is to drop this check here, and update the docstring.

BTW, this suggestion follows the model of what we do for vmap, where we have an `in_axes` (they apply only to positional arguments).

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -516,6 +516,15 @@ def jax_fn_array(x):
         tf_fn_array(np.array([3, 4, 5])), np.array([4.5, 10, 17.5],
                                                    jnp.bfloat16))
 
+  def test_kwargs(self):
+    # Re: https://github.com/google/jax/issues/6791
+    def f_jax(x):
```
gnecula, line 521: Would this test fail if we replace `x` with `y`? Probably not. Maybe the test would be clearer if `x` is forced to be a kwarg here?

jax/experimental/jax2tf/jax2tf.py
saeta: Excellent. That's great.

jax/experimental/jax2tf/tests/jax2tf_test.py
saeta: I've now forced `x` to be a `kwarg` by adding a `*`.

saeta:
PTAL? In particular, feedback regarding the text in the docstring would be appreciated!

gnecula (approved):
Thank you!

jax/experimental/jax2tf/tests/jax2tf_test.py
```diff
@@ -516,6 +516,15 @@ def jax_fn_array(x):
         tf_fn_array(np.array([3, 4, 5])), np.array([4.5, 10, 17.5],
                                                    jnp.bfloat16))
 
+  def test_kwargs(self):
+    # Re: https://github.com/google/jax/issues/6791
+    def f_jax(*, x):
+      return jnp.sum(x)
+    f_tf = jax2tf.convert(f_jax)
+    self.assertAllClose(
+      f_tf(x=jnp.zeros(3)),  # Call with kwargs.
```
gnecula, line 525: This test fails with JAX_ENABLE_X64=True. I think that you may need to force `jnp.zeros(3, dtype=jnp.float32)`. 

Note that unfortunately the jax2tf test do not run in GitHub, only internally, due to JAX not being ready to take a dependency on tf_nightly, which is needed for jax2tf. This means that test failures only pop up when we try to import the change in Google.

PR #6807, opened 2021-05-21
Reduce redundant calculations of `tan`, `erfc` and `rsqrt` jvp

lgeiger:
This is a minor change to the `jvp`s of `tan`, `erfc` and `rsqrt` to reuses some already computed values.
When jited I guess XLA will already do some of these optimizations automatically, so this shouldn't have a big performance impact, but it might help a bit in eager mode.


gnecula (approved):
LGTM, thanks!

mattjj:
Wow, nicely done, thanks for improving these.

PR #6819, opened 2021-05-24
Replace `pow()` with `sqrt()` or `square()` where possible

lgeiger:
This PR replaces `lax.pow()` with `lax.square()` or `lax.sqrt()` where possible. `lax.pow()` may be imprecise for floating-point values so switching to `lax.square()` should be preferred since it uses `lax.integer_power()` under the hood and might even be slightly faster.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1019,7 +1019,7 @@ def _gamma_one(key, alpha):
   alpha = lax.select(lax.ge(alpha, one), alpha, lax.add(alpha, one))
 
   d = lax.sub(alpha, one_over_three)
-  c = lax.div(one_over_three, lax.pow(d, one_over_two))
+  c = lax.div(one_over_three, lax.sqrt(d))
 
   def _cond_fn(kXVU):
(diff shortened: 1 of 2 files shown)


gnecula (approved):
Nice! Thank you.

hawkinsp:
I'll note that XLA makes a similar transformation on `jit`-compiled programs, but of course that only applies under a jit: this is  certainly a good change! In addition to helping the non-`jit` case, it may also improve, say, autodiff a little bit.

lgeiger:
> I'll note that XLA makes a similar transformation on jit-compiled programs, but of course that only applies under a jit: this is certainly a good change! In addition to helping the non-jit case, it may also improve, say, autodiff a little bit.

Fair point! I think LLVM applies `pow(x, 2) -> square(x)` optimizations as well, but I wasn't sure about `pow(x, 0.5) -> sqrt(x)` since this is usually disabled, unless `-ffast-math` flag is used, since it might violate IEEE floating point spec.

PR #6825, opened 2021-05-24
add fingerprint to debugging log

colemanliyah:
Example Code and Output:

```
In [1]: import jax
   ...: jax.config.update("jax_distributed_debug", True)
   ...: def f(x):
xb.register_constant_handler(ShardedDeviceArray, _sharded_device_array_constant_handler)
   ...:     return x + 1
   ...: jax.pmap(f)(jax.numpy.ones(1))
```

```
WARNING:absl:
DISTRIBUTED_DEBUG_BEGIN
Initialized backend: cpu
  process_index: 0
  device_count: 1
  local_devices: [CpuDevice(id=0)]
DISTRIBUTED_DEBUG_END
WARNING:absl:
DISTRIBUTED_DEBUG_BEGIN
Initialized backend: cpu
  process_index: 0
  device_count: 1
  local_devices: [CpuDevice(id=0)]
DISTRIBUTED_DEBUG_END
WARNING:absl:
DISTRIBUTED_DEBUG_BEGIN
Running pmapped function: f
  python function: <function f at 0x7ff6d3785820>
  devices: None
  abstract args: [ShapedArray(float32[1])]
  fingerprint: None
DISTRIBUTED_DEBUG_END
Out[1]: ShardedDeviceArray([2.], dtype=float32)
```

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -59,6 +59,10 @@ from . import partial_eval as pe
 from . import xla
 from . import ad
 
+# Built in Python lists don't support weak refs but subclasses of lists do.
+class WeakRefList(list):
+  pass
+
 if sys.version_info >= (3, 8):
   from functools import cached_property as maybe_cached_property
 else:
@@ -616,17 +620,19 @@ def xla_pmap_impl(fun: lu.WrappedFun, *args, backend, axis_name, axis_size,
                   global_axis_size, devices, name, in_axes, out_axes_thunk,
                   donated_invars, global_arg_shapes):
   abstract_args = unsafe_map(xla.abstractify, args)
-  compiled_fun = parallel_callable(fun, backend, axis_name, axis_size,
+  compiled_fun, xla_executable = parallel_callable(fun, backend, axis_name, axis_size,
                                    global_axis_size, devices, name,
                                    in_axes, out_axes_thunk,
                                    donated_invars, global_arg_shapes,
                                    *abstract_args)
+
   # Don't re-abstractify args unless logging is enabled for performance.
   if config.jax_distributed_debug:
     distributed_debug_log(("Running pmapped function", name),
                           ("python function", fun.f),
                           ("devices", devices),
-                          ("abstract args", map(xla.abstractify, args)))
+                          ("abstract args", map(xla.abstractify, args)),
+                          ("fingerprint", xla_executable.fingerprint))
   return compiled_fun(*args)
 
 @lu.cache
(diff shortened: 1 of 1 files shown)


jax/interpreters/pxla.py
```diff
@@ -621,13 +624,15 @@ def xla_pmap_impl(fun: lu.WrappedFun, *args, backend, axis_name, axis_size,
...
   # Don't re-abstractify args unless logging is enabled for performance.
   if config.jax_distributed_debug:
     distributed_debug_log(("Running pmapped function", name),
                           ("python function", fun.f),
                           ("devices", devices),
-                          ("abstract args", map(xla.abstractify, args)))
-  return compiled_fun(*args)
+                          ("abstract args", map(xla.abstractify, args)),
+                          ("Fingerprint", compiled_fun[1].fingerprint))
+  return compiled_fun[0](*args)
```
skye, line 635: Instead of using `compiled_fun[0]` and `compiled_fun[1]`, unpack these into two separate variables immediately at the `parallel_callable` call site with more descriptive names. Something like:
```
compiled_fun, xla_executable = parallel_callable(...)
```
(These still aren't the most descriptive names, but it's the best I can come up with)

jax/interpreters/pxla.py
```diff
@@ -621,13 +624,15 @@ def xla_pmap_impl(fun: lu.WrappedFun, *args, backend, axis_name, axis_size,
...
+
   # Don't re-abstractify args unless logging is enabled for performance.
   if config.jax_distributed_debug:
     distributed_debug_log(("Running pmapped function", name),
                           ("python function", fun.f),
                           ("devices", devices),
-                          ("abstract args", map(xla.abstractify, args)))
-  return compiled_fun(*args)
+                          ("abstract args", map(xla.abstractify, args)),
+                          ("Fingerprint", compiled_fun[1].fingerprint))
```
skye, line 634: Nit pick: the other logged fields are all lowercase, so make "fingerprint" all lowercase too :)

jax/interpreters/pxla.py
```diff
@@ -890,7 +895,8 @@ def parallel_callable(fun: lu.WrappedFun,
                                       handle_outs)
   compiled = xla.backend_compile(backend, built, compile_options)
   handle_args = partial(shard_args, compiled.local_devices(), input_indices)
-  return partial(execute_replicated, compiled, backend, handle_args, handle_outs)
+  returnable_list = list([partial(execute_replicated, compiled, backend, handle_args, handle_outs), compiled])
```
skye, line 898: Nit: instead of creating a variable `returnable_list`, I think it'll be easier to read if you do it like:
```python
execute_fun = partial(...)
return WeakRefList([execute_fun, compiled])
```
(This is also incorporating the list rename comment from above)

skye (changes requested):
Nice work!

skye:
Can you also update the PR description to include example output of the debug log with this change? This helps verify the logging actually looks like what we expect. You can copy/paste the whole DISTRIBUTED_DEBUG_BEGIN block from your test command.

skye (changes requested):
Just a few more tiny style comments! I also forgot that fingerprint returns None on non-TPU platforms. You'll be able to see more interesting fingerprints once we get you set up with Cloud TPU VMs (we can submit this change before then though).

skye (approved):
Looks good!

PR #6834, opened 2021-05-25
Enable custom gradients SavedModel option in test

k-w-w:
TensorFlow SavedModel added the option to save custom gradients in [e6c7ba](https://github.com/tensorflow/tensorflow/commit/e6c7ba3875634d7c142d787c2c5012cea9ad8daa). This change updates the JAX SavedModel test to use this option.

### DIFF
--- a/jax/experimental/jax2tf/tests/savedmodel_test.py
+++ b/jax/experimental/jax2tf/tests/savedmodel_test.py
@@ -35,7 +35,10 @@ class SavedModelTest(tf_test_util.JaxToTfTestCase):
   def save_and_load_model(self, model: tf.Module) -> tf.Module:
     # Roundtrip through saved model on disk.
     model_dir = os.path.join(absltest.get_default_test_tmpdir(), str(id(model)))
-    tf.saved_model.save(model, model_dir)
+    tf.saved_model.save(
+        model,
+        model_dir,
+        options=tf.saved_model.SaveOptions(custom_gradients=True))
     restored_model = tf.saved_model.load(model_dir)
     return restored_model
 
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/tests/savedmodel_test.py
```diff
@@ -94,13 +97,8 @@ def f_jax_jvp(primals, tangents):
...
       y = restored_model.f(xv)
-
-    # TODO: at the moment TF does not fully-support custom_gradient in a
-    # savedmodel (b/123499169), but at least a warning is printed and an
-    # exception is thrown when gradients are taken. The exception, however,
-    # is a very strange one, for now.
-    with self.assertRaisesRegex(TypeError, "An op outside of the function building code is being passed"):
-      _ = tape.gradient(y, xv)
+    self.assertAllClose(tape.gradient(y, xv).numpy(),
+                        jax.grad(f_jax)(0.7).astype(np.float32))
```
gnecula, line 101: We should use here `x` in lieu of `0.7`.

gnecula (commented):
Thank you so much for fixing the custom_gradient support. It has been a long-standing issue for us. 

I would recommend revising also the text in https://github.com/google/jax/blob/3ec5bc503139985c8d099f2ba8bdb71fc059d051/jax/experimental/jax2tf/README.md#no-savedmodel-fine-tuning. I would say instead that we support saving only for the first-order gradient computation.

jax/experimental/jax2tf/tests/savedmodel_test.py
k-w-w: Updated the `x` and documentation

gnecula (approved):
LGTM. Thanks!

PR #6836, opened 2021-05-25
DOC: add note about array views in numpy docs

### LINKED ISSUE #6831: Functions claim to return a view but return a copy
My understanding is the immutable-array policy means that `jnp` functions (eg, `transpose`) return a copy of an array, rather than a view. 

There are a few places in the docs where a function claims to return a view:  [atleast_2d](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.atleast_2d.html#jax.numpy.atleast_2d) and [atleast_3d](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.atleast_3d.html#jax.numpy.atleast_3d) state it in the top line of the docstring.   

Other functions, like [ravel](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ravel.html) or [transpose](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.transpose.html), have some verbiage about returning a view where possible in the "original docstring below" portion.  I suppose this is inevitable if the docstring is copied directly from the numpy functions.  

Perhaps the view/copy distinction could be added to the bulleted list in the `jax.numpy` [API doc page ](https://jax.readthedocs.io/en/latest/jax.numpy.html)?

jakevdp:
Fixes #6831

Rendered version: https://jax--6836.org.readthedocs.build/en/6836/jax.numpy.html#module-jax.numpy

### DIFF
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -17,8 +17,13 @@ cannot follow NumPy exactly.
   array updates (:code:`x[i] = y`), JAX provides an alternative pure indexed
   update function :func:`jax.ops.index_update`.
 
+* Relatedly, some NumPy functions return views of arrays when possible (examples
+  are :func:`numpy.transpose` and :func:`numpy.reshape`). JAX versions of such
+  functions will return copies instead, although such copies can often be optimized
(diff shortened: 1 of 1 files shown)


gnecula (approved):
Thanks

PR #6839, opened 2021-05-26
DOC: add notes to jax.numpy docstrings about returning copies...

### LINKED ISSUE #6831: Functions claim to return a view but return a copy

jakevdp:
…rather than views

Fixes #6831

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1082,15 +1082,18 @@ def _sinc_maclaurin_jvp(k, primals, tangents):
   (x,), (t,) = primals, tangents
   return _sinc_maclaurin(k, x), _sinc_maclaurin(k + 1, x) * t
 
+_ARRAY_VIEW_DOC = """
+The JAX version of this function will return a copy rather than a view of the input.
+"""
 
-@_wraps(np.transpose)
+@_wraps(np.transpose, lax_description=_ARRAY_VIEW_DOC)
 def transpose(a, axes=None):
   _check_arraylike("transpose", a)
   axes = np.arange(ndim(a))[::-1] if axes is None else axes
   return lax.transpose(a, axes)
 
 
-@_wraps(np.rot90)
+@_wraps(np.rot90, lax_description=_ARRAY_VIEW_DOC)
 def rot90(m, k=1, axes=(0, 1)):
   _check_arraylike("rot90", m)
   ax1, ax2 = axes
@@ -1112,7 +1115,7 @@ def rot90(m, k=1, axes=(0, 1)):
       return flip(transpose(m, perm), ax2)
 
 
-@_wraps(np.flip)
+@_wraps(np.flip, lax_description=_ARRAY_VIEW_DOC)
 def flip(m, axis: Optional[Union[int, Tuple[int, ...]]] = None):
   _check_arraylike("flip", m)
   if axis is None:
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1082,15 +1082,18 @@ def _sinc_maclaurin_jvp(k, primals, tangents):
   (x,), (t,) = primals, tangents
   return _sinc_maclaurin(k, x), _sinc_maclaurin(k + 1, x) * t
 
+_ARRAY_VIEW_DOC = """
+The JAX version will return a copy rather than a view of the input.
```
8bitmp3, line 1086: Would it be okay to specify the "version of x" here to help less experienced users? 

e.g.

```suggestion
The JAX version of NumPy will return a copy rather than a view of the input.
```

(assuming it's that, based on what I read in _NumPy, lax & XLA: JAX API layering_
 https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#numpy-lax-xla-jax-api-layering)

8bitmp3 (commented):
Would it be okay to specify the "version of x" here to help less experienced users? 

e.g.

```suggestion
The JAX version of NumPy will return a copy rather than a view of the input.
```

(assuming it's that, based on what I read in _NumPy, lax & XLA: JAX API layering_
 https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#numpy-lax-xla-jax-api-layering)

jax/_src/numpy/lax_numpy.py
jakevdp: In context, this is at the top of a function's doc string, so it's meant to imply "the jax version [of this function]".

It require a bit more boilerplate to include the name of the function in each of these, but I judged it to be clear enough in context. What do you think?

jakevdp: To be concrete, here's how the current version looks in context for the docstring of reshape:
```
Signature: jnp.reshape(a, newshape, order='C')
Docstring:
Gives a new shape to an array without changing its data.

LAX-backend implementation of :func:`reshape`.

The JAX version will return a copy rather than a view of the input.

*Original docstring below.*
```

8bitmp3: How about:

```suggestion
The JAX version of this function will return a copy rather than a view of the input.
```

🤷‍♂️

jakevdp: Sure, done.

froystig (approved):

PR #6841, opened 2021-05-26
fixed fingerprint debugging message to be compatible with current min jaxlib version

colemanliyah:


jakevdp (approved):
Looks great, thanks!

PR #6842, opened 2021-05-26
QDWH

adam-lewis-gcorp:


jakevdp:
Is this meant as a replacement for #6751?

adam-lewis-gcorp:
@googlebot I signed it!

adam-lewis-gcorp:
@hawkinsp Added the eigh stuff - I can put some more tests in, but for now let's see what we can do about the Jitting

jax/_src/scipy/test_eigh.py
```diff
@@ -0,0 +1,49 @@
+"""Test for eigh.py."""
```
hawkinsp, line 1: Would you mind moving the tests into the jax/tests directory?

jax/_src/scipy/test_eigh.py
```diff
@@ -0,0 +1,49 @@
...
+
+import jax.numpy as jnp
+from jax import lax
+import numpy as np
+import pytest
+
+Ns = [16, 256]
+precisions = [lax.Precision.HIGHEST,]
+
+@pytest.mark.parametrize("N", Ns)
```
hawkinsp, line 12: I'd avoid using pytest marks. We don't always use pytest to run tests (e.g., in our internal CI), so the tests need to work without pytest as the test runner, which means you can't use marks.

The test should work if you just run `python test_eigh.py` without pytest; you'd probably need to add a `__main__`  and move the tests into a class for this to work.

Typically we use absl's `parameterized` module for parametric tests.

Take a look at, say, `lax_numpy_test.py` for some examples.

jax/_src/scipy/test_qdwh.py
```diff
@@ -0,0 +1,95 @@
...
+  svs[n_zero_svs:] = nonzero_svs
+  svs = svs[::-1]
+
+  result = np.dot(left_vecs * svs, right_vecs.conj().T).astype(np.float32)
+  result = jnp.array(result)
+  spectrum = jnp.array(svs.astype(np.float32))
+  return result, spectrum
+
+
+@pytest.mark.parametrize("n_zero_svs", n_zero_svs)
```
hawkinsp, line 54: Same here.

jax/_src/scipy/eigh.py
```diff
@@ -0,0 +1,223 @@
...
+    return jnp.linalg.eigh(H)
+
+  evals, evecs = _eigh_work(H, precision=precision)
+  sort_idxs = jnp.argsort(evals)
+  evals = evals[sort_idxs]
+  evecs = evecs[:, sort_idxs]
+  return evals, evecs
+
+
+def svd(A, precision=lax.Precision.HIGHEST):
```
hawkinsp, line 208: It looks to me like the polar decomposition is jittable, correct?

One thing we could probably do is hook up the polar decomposition to the API (because why not, it's useful) and use the existing `eigh` implementation as the basis of an SVD on TPU, which is simply missing at the moment.

We can separately work on making the `eigh` above jittable.

jax/_src/scipy/qdwh.py
```diff
@@ -0,0 +1,235 @@
+"""
```
hawkinsp, line 1: I'd be tempted to move this into `jax/_src/lax/qdwh.py`. This is mostly because we want to be able to call it from `jax.lax.linalg` primitives like SVD. Optional, though, I can do that in another PR.

hawkinsp (changes requested):

jax/_src/scipy/eigh.py
adam-lewis-gcorp: Yes, it should be jittable

jax/_src/scipy/test_eigh.py
adam-lewis-gcorp: Done

jax/_src/scipy/test_eigh.py
adam-lewis-gcorp: Done

tests/lax_scipy_test.py
```diff
@@ -405,6 +454,137 @@ def testSphHarmCornerCaseWithWrongNmax(self):
...
+    matrix, _ = _initialize_polar_test(
+      shape, n_zero_sv, degeneracy, geometric_spectrum, max_sv,
+      nonzero_condition_number, dtype)
+    if dtype in (jnp.bfloat16, np.float16):
+      self.assertRaises(
+        NotImplementedError, jsp.linalg.polar, matrix, method=method,
+        side=side)
+      return
+
+    unitary, posdef, info = jsp.linalg.polar(matrix, method=method, side=side)
```
hawkinsp, line 495: I'm actually missing where we export `polar` in the `jax.scipy.linalg` namespace. Don't we need to add it to the list in `jax/scipy/linalg.py`

adam-lewis-gcorp: isn't it in there?

jax/scipy/linalg.py
```diff
@@ -35,3 +35,8 @@
   tril,
   triu,
 )
+
+from jax._src.lax.polar import (
+  polar,
+  polar_unitary
+)
```
adam-lewis-gcorp, line 42: The namespace adding is up here

hawkinsp: Hmm I must have missed this.

jax/_src/scipy/qdwh.py
adam-lewis-gcorp: Moved

hawkinsp (changes requested):
You need to add the Apache license header to the new files.

hawkinsp:
Merged #7335 which is a squashed version of this PR.

PR #6848, opened 2021-05-27
Fix all_to_all fallback implementation when axis_index_groups are used.

chr1sj0nes:


apaszke:
Thanks! This looks like a duplicate of #6818, which additionally includes a test, so let's merge the other PR. But feel free to reopen if I've missed something!

apaszke:
But I did end up stealing your neat `argsort()` trick to invert a permutation :)

PR #6851, opened 2021-05-27
Handle negative values for list-like sections in `jnp.split`

### LINKED ISSUE #6599: Support negative numpy.split indices
```python
import numpy as np
np.split([1, 2, 3], [-1])
```

works, while

```python
import jax.numpy as jnp
jnp.split(jnp.array([1, 2, 3]), [-1])
```

fails.

nicholasjng:
Fixes #6599. I added the repro from that same issue as a test. 

I guess this is technically abuse of notation as `i_s` is not an axis there but rather an index. I did it this way to prefer an existing core functionality over inserting technical debt from my own logic.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1805,7 +1805,8 @@ def _split(op, ary, indices_or_sections, axis=0):
   size = ary.shape[axis]
   if isinstance(indices_or_sections, (tuple, list) + _arraylike_types):
     indices_or_sections = np.array(
-        [core.concrete_or_error(np.int64, i_s, f"in jax.numpy.{op} argument 1")
+        [_canonicalize_axis(core.concrete_or_error(np.int64, i_s,
+                                                   f"in jax.numpy.{op} argument 1"), size)
          for i_s in indices_or_sections], np.int64)
     split_indices = np.concatenate([[np.int64(0)], indices_or_sections,
                                     [np.int64(size)]])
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1805,7 +1805,8 @@ def _split(op, ary, indices_or_sections, axis=0):
   size = ary.shape[axis]
   if isinstance(indices_or_sections, (tuple, list) + _arraylike_types):
     indices_or_sections = np.array(
-        [core.concrete_or_error(np.int64, i_s, f"in jax.numpy.{op} argument 1")
+        [_canonicalize_axis(core.concrete_or_error(np.int64, i_s,
+                                                   f"in jax.numpy.{op} argument 1"), size)
```
jakevdp, line 1809: Unrelated to your change, but the explicit cast to int64 here allows non-integer floats to be passed as indices. We should probably use `operator.index` rather than `np.int64`.

njunge94: I see. Looking at `testSplitTypeError` in `lax_numpy_test.py`, which calls `jnp.split` on some floating point values, I assumed that was intended. Indeed, changing to `operator.index` breaks that test for me locally (the last JVP evaluation, specifically). Can you even evaluate a JVP of an integer-domain function?

jakevdp: Thanks for looking into it! It's not clear to me that the JVP over split indices is meaningful, but we can probably punt on that for now.

jakevdp (approved):
Looks good, thanks!

PR #6856, opened 2021-05-28
Typo.

lkhphuc:


jakevdp:
Good catch - thanks! The companion markdown will have to be updated as well. See https://jax.readthedocs.io/en/latest/developer.html#update-notebooks for details; in this case the change is small enough that you could probably get away with just applying the same fix to the markdown file:  https://github.com/google/jax/blob/master/docs/notebooks/Common_Gotchas_in_JAX.md

lkhphuc:
I updated the md file as well. Hope that will work. Let me know if the jupytext need to be run again.

jakevdp (approved):
Looks great! I should have also mentioned: can you please squash your changes into a single commit? (see [Single-Change commits](https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)) Thanks!

lkhphuc:
@jakevdp sorry forgot to ping you earlier. I squashed the commit and force push this branch.

jakevdp (approved):
Looks great, thanks!

PR #6860, opened 2021-05-31
[jax2tf] Add support for generating HLO OpMetadata in the TF graph

gnecula:
The goal is to ensure that the HLO that
jax2tf->TF/XLA generates has the same metadata as what JAX generates.
This includes `op_type`, `op_name`, and source information, which are
used for debugging and profiling.

In order to ensure that this metadata is carried from the JAX tracing
time to TF/XLA, we save the metadata in custom TF op attributes. These
attributes are automatically preserved through SavedModel. This relies
on a separate change in TF/XLA to look for these custom attributes
and override its default.

For the source information, we use pretty much the same code that
xla.py uses. HLO OpMetadata has room for only one source location.
JAX (xla.py) picks the top-most user frame, which is obtained by
filtering out the stack frames in the JAX source tree. When used
with jax2tf we also need to filter out stack frames in the
TensorFlow source tree.

The hardest part is to generate the `op_name`, which is a hierarchical
name with components separated by '/', e.g., `jax2tf(top_func)/while/cond/le`.
We carry the current `name_stack` in thread-local state. Unfortunately, there
is no easy way to share the exact code that achieves this in xla.py. At the
same time it is not crucial that we have exactly identical name stacks as in
JAX.

I attempted to also carry this state in the JAX `MainTrace`, but could not
fully control the name stack. E.g., when calling a jitted-function we
have to reuse the current `MainTrace` although we want to push an element
on the name stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ PLEASE REMEMBER TO CHANGE THE '..master' WITH AN ACTUAL TAG in GITHUB LINK.
     unknown dimensions are used in arithmetic operations, e.g., `jnp.reshape(-1)`
     ({jax-issue}`#6827`).    
 
+  * The {func}`jax2tf.convert` generates custom attributes with location information
+   in TF ops. The code that XLA generates after jax2tf
+   has the same location information as JAX/XLA.
+
 * Breaking changes:
 
 * Bug fixes:
@@ -29,8 +33,6 @@ PLEASE REMEMBER TO CHANGE THE '..master' WITH AN ACTUAL TAG in GITHUB LINK.
   * The {func}`jax2tf.convert` now scopes the `enable_xla` conversion parameter
     properly to apply only during the just-in-time conversion
     ({jax-issue}`#6720`).
-  * Fixed assertion failure in {func}`jax2tf.call_tf` when used with captured
-    `tf.Variable` ({jax-issue}`#6572`).
   * The {func}`jax2tf.convert` now converts `lax.dot_general` using the
     `XlaDot` TensorFlow op, for better fidelity w.r.t. JAX numerical precision
     ({jax-issue}`#6717`).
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -13,16 +13,16 @@ PLEASE REMEMBER TO CHANGE THE '..master' WITH AN ACTUAL TAG in GITHUB LINK.
...
+   in TF ops. The code that XLA generates after jax2tf
+   has the same location information as JAX/XLA.
+
 * Breaking changes:
 
 * Bug fixes:
   * The {func}`jax2tf.convert` now scopes the `enable_xla` conversion parameter
     properly to apply only during the just-in-time conversion
     ({jax-issue}`#6720`).
-  * Fixed assertion failure in {func}`jax2tf.call_tf` when used with captured
```
jekbradbury, line 22: Accidentally dropped?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -186,13 +211,19 @@ def convert(fun: Callable,
     enable_xla: if unset, the converter will try harder to use pure TF ops to
       convert the function, and raise an error if it can not be converted
       without resorting to XLA ops (default: True).
+    include_xla_op_metadata: if set, the generated TF ops will include
```
jekbradbury, line 214: If unset, will the TF ops have the same XLA op metadata (after the action of the bridge) that they'd have if they were written in raw TF, or will they have no metadata? And does this one flag also control non-XLA metadata in the form of the TF graph name scope?

jekbradbury (approved):

CHANGELOG.md
gnecula: Actually, it was a duplicate. This was already includes in 0.2.13

jax/experimental/jax2tf/jax2tf.py
gnecula: I have refined the comment to say that if unset then TF will use default OpMetadata (as if written in raw TF). Those will not contain correct source location, because TF will pick the top-most non-TF stack trace, and that would
point to some place in the jax2tf converter. Furthermore, without these OpMetadata, when loading from SavedModel the source location will point to the line where `tf.saved_model.load()` is called. 

The name scopes are not affected by this flag.

PR #6861, opened 2021-05-31
rename v to vh

ho-oto:
`svd` returns `u`, `s` and `vh`, not `v`.
This patch fixes the confusing naming of the local variable.


ho-oto:
@googlebot I signed it!

hawkinsp (approved):

PR #6869, opened 2021-06-01
Completed the implementation for File system cache 

colemanliyah:


jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+import os
+import tempfile
+
+class FileSystemCache:
+
+  def __init__(self, path: str):
+    """Sets up a cache at 'path'. Cached values may already be present."""
+    if not os.path.isdir(path):
+      print("HERE\n")
```
colemanliyah, line 29: I'll get rid of this print statement with the other changes I need to make

jax/experimental/compilation_cache/compilation_cache_test.py
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from absl.testing import absltest
+import jax.test_util as jtu
+import file_system_cache as fsc
+
+class CompilationClassTest(jtu.JaxTestCase):
```
skye, line 19: Call this FileSystemCacheTest, to better describe what it's testing

jax/experimental/compilation_cache/compilation_cache_test.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from absl.testing import absltest
+import jax.test_util as jtu
+import file_system_cache as fsc
```
skye, line 17: To be even more concise, I would use: `from file_system_cache import FileSystemCache`. Then you don't need the `fsc` abbreviation at all.

jax/experimental/compilation_cache/compilation_cache_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+class CompilationClassTest(jtu.JaxTestCase):
+
+  def test_new_directory_and_path(self):
+   fileSystemCacheInstance = fsc.FileSystemCache("/tmp/new_cache_directory")
```
skye, line 22: Now that you have this working in `/tmp`, let's switch to using [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) using a `with` statement. This will make sure the cache directory is deleted after the test runs, which will prevent tests from leaving a bunch of cache files around and possibly interfering with other tests.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,46 @@
...
+      os.mkdir(path)
+    self.path  = path
```
skye, line 31: Use `self._path` instead of `self.path` (here and below). Using a leading underscore is a common way to mark private variables in Python.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,46 @@
...
+
+  def get(self, key: str) -> Optional[bytes]:
+    """Returns None if 'key' isn't present."""
+    path_to_key = os.path.join(self.path, key)
+    if(os.path.exists(path_to_key)):
+      with open(path_to_key, "rb") as file:
+        file_contents = file.read()
+      return file_contents
```
skye, line 39: You can combine this into one line: `return file.read()`

skye (changes requested):

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,48 @@
...
+
+  def __init__(self, path: str):
+    """Sets up a cache at 'path'. Cached values may already be present."""
+    if not os.path.isdir(path):
+      os.mkdir(path)
+    self._path  = path
+
+  def get(self, key: str) -> Optional[bytes]:
+    """Returns None if 'key' isn't present."""
+    if(key==""):
```
skye, line 34: Style thing: you don't need the parentheses, and put spaces after `if` and around `==`, like this: `if key == ""`

In general, see https://engdoc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head#s3.6-whitespace

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,48 @@
...
+      raise ValueError("key cannot be empty, you cannot create a directory path for a new file from the home directory which has already been created.")
```
skye, line 35: I don't really understand the part about creating a new file from the home directory. Let's just leave it at "key cannot be empty".

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,48 @@
...
+    path_to_key = os.path.join(self._path, key)
+    if(os.path.exists(path_to_key)):
```
skye, line 37: Same as above, put space after `if` and you don't need the parens

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,48 @@
...
+      with open(path_to_key, "rb") as file:
+        return file.read()
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry, possibly evicting older entries."""
+    if(key==""):
```
skye, line 45: And here

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+from absl.testing import absltest
+from file_system_cache import FileSystemCache
+import jax.test_util as jtu
+import tempfile
+
+class FileSystemCacheTest(jtu.JaxTestCase):
+
+  def test_new_directory_and_path(self):
+   with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
```
skye, line 24: Style thing: `file_system_cache_instance` is very descriptive, but it's too long in my opinion :) You can just use `cache`.

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+  def test_new_directory_existing_path(self):
+   with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
+     file_system_cache_instance.put("foo", b"bar")
+     self.assertEqual(file_system_cache_instance.get("foo"), b"bar")
+
+  def test_existing_cache_path(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
```
skye, line 35: Same as above, you can just call these `cache1` and `cache2`

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+from file_system_cache import FileSystemCache
+import jax.test_util as jtu
+import tempfile
+
+class FileSystemCacheTest(jtu.JaxTestCase):
+
+  def test_new_directory_and_path(self):
+   with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
+     self.assertEqual(file_system_cache_instance.get("nonExistantFile"), None)
```
skye, line 25: This is super minor, but instead of using "nonExistentFile", use "nonExistentKey". Even though it doesn't make a difference to the test which nonexistent key is used, I think it's better to use cache-related terms instead of filesystem-related terms when testing `get` and `put`, since this test file doesn't really deal with files directly (it's handled in the FileSystemCache code instead). The only filesystem-related thing this test file deals with is the path used to create the cache(s), not the files themselves.

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+# limitations under the License.
+
+from absl.testing import absltest
+from file_system_cache import FileSystemCache
+import jax.test_util as jtu
+import tempfile
+
+class FileSystemCacheTest(jtu.JaxTestCase):
+
+  def test_new_directory_and_path(self):
```
skye, line 22: Rename to `test_get_nonexistent_key`, I think that's more descriptive

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+   with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
+     self.assertEqual(file_system_cache_instance.get("nonExistantFile"), None)
+
+  def test_new_directory_existing_path(self):
```
skye, line 27: Rename to `test_put_and_get_key`

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,59 @@
...
+   with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
+     file_system_cache_instance.put("foo", b"bar")
+     self.assertEqual(file_system_cache_instance.get("foo"), b"bar")
+
+  def test_existing_cache_path(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+     file_system_cache_instance = FileSystemCache(tmpdir)
+     file_system_cache_instance.put("foo", b"bar")
```
skye, line 36: Let's actually delete the first cache here (I know I said it's unnecessary before, but I changed my mind :)). Something like `del cache1`. Then we can be sure that the cache files still exist even after the cache is deleted.

skye (changes requested):
Looking great! Just some final style comments.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+#
+## Core functionality:
+##create my own directory in experimental folder in jax
+##file for this class then seperate file for tests
+
+
+from typing import Optional
+import os
+import tempfile
```
skye, line 22: Remove unnecessary import (it's only used in the test file)

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,42 @@
...
+
+  def __init__(self, path: str):
+    """Sets up a cache at 'path'. Cached values may already be present."""
+    if not os.path.isdir(path):
+      os.mkdir(path)
+    self._path  = path
+
+  def get(self, key: str) -> Optional[bytes]:
+    """Returns None if 'key' isn't present."""
+    if key == "":
```
skye, line 28: Ah I made a mistake earlier. This should actually be `if not key`, see https://google.github.io/styleguide/pyguide.html#2144-decision

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,42 @@
...
+    path_to_key = os.path.join(self._path, key)
+    if os.path.exists(path_to_key):
+      with open(path_to_key, "rb") as file:
+        return file.read()
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry, possibly evicting older entries."""
+    if key == "":
```
skye, line 39: Same here

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2021 Google LLC
```
skye, line 1: Another thing I just realized: this test file should actually go in [jax/tests](https://github.com/google/jax/tree/master/tests). This is where we put all test files, and this will make sure this file is included when we run all tests.

jax/experimental/compilation_cache/file_system_cache_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+  def test_put_and_get_key(self):
+   with tempfile.TemporaryDirectory() as tmpdir:
+     cache = FileSystemCache(tmpdir)
+     cache.put("foo", b"bar")
+     self.assertEqual(cache.get("foo"), b"bar")
+
+  def test_existing_cache_path(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache_one = FileSystemCache(tmpdir)
```
skye, line 35: Style nit: you can use `cache1` and `cache2` instead of spelling out the numbers. Saves characters!

skye (changes requested):
Just a few more comments I should have made earlier :)

@zhangqiaorjc I think this is ready for you to review as well

jax/experimental/compilation_cache/file_system_cache_test.py
colemanliyah: If I move it to that directory it looks, it will not let me run this test file in the current directory where I have file_system_cache. Should I just make a copy or move it over at the end?

skye: You'll have to run the test from a different directory. I usually do everything from the base `jax` directory, so it'd be something like: `python3 tests/file_system_cache_test.py`

You'll also have to change the file_system_cache import on L16. I think `from jax.experimental.file_system_cache import FileSystemCache` should work, but you might have to experiment...

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,42 @@
...
+# limitations under the License.
+
+from typing import Optional
+import os
+
+class FileSystemCache:
+
+  def __init__(self, path: str):
+    """Sets up a cache at 'path'. Cached values may already be present."""
+    if not os.path.isdir(path):
```
zhangqiaorjc, line 22: I think you meant to have
```
if not os.path.exists(path):
    os.makedirs(path)
```
An even better version might be

```
os.makedirs(path, exist_ok=True)
```

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,42 @@
...
+      raise ValueError("key cannot be empty")
+    path_to_key = os.path.join(self._path, key)
+    if os.path.exists(path_to_key):
+      with open(path_to_key, "rb") as file:
+        return file.read()
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry, possibly evicting older entries."""
```
zhangqiaorjc, line 38: Adds a TODO for eviction?

zhangqiaorjc (changes requested):
Great job Liyah!

jax/experimental/compilation_cache/file_system_cache.py
colemanliyah: I made this change, but actually could it pose a problem such as replicating a cache directory? Since I don't have an if statement now it says a directory will be made no matter what the path is but don't I just want it to make a directory if the path given isn't already one?

zhangqiaorjc: os.makedirs("path/to/directory", exist_ok=True)  # succeeds even if directory exists.

see
https://stackoverflow.com/questions/273192/how-can-i-safely-create-a-nested-directory

this is equivalent do "mkdir -p"

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -0,0 +1,42 @@
...
+    if not key:
+      raise ValueError("key cannot be empty")
+    #TODO(colemanliyah):implement eviction
```
zhangqiaorjc, line 40: minor comment: add a space here

zhangqiaorjc: ```# TODO(colemanliyah): implement eviction```

skye (approved):

jax/experimental/compilation_cache/file_system_cache.py
skye: This is still not quite right, but I explained to Liyah offline how to format TODOs moving forward. This particular TODO is going away in the next PR, so I think we can leave as-is for now.

zhangqiaorjc (approved):

PR #6873, opened 2021-06-02
add gpu to the rocsolver backend for ROCm

reza-amd:
This PR fixes the missing backend argument for ROCm rocsolver


hawkinsp (approved):
Thanks for the fix!

Issue #6874, opened 2021-06-02
Confusing differences between numpy and JAX type promotion


gnecula:
The following computes an int16, as expected
```
np.int16(1) + jnp.int16(2) + 3
```

But the following computes an int32, because `3 + np.int16(1)` results in `np.int64(1)` (and if JAX_ENABLE_X64=0 this gets canonicalized to a `int32`).
```
3 + np.int16(1) + jnp.int16(2)
```

For now I will just change the documentation, but I think it is fair to classify this as a surprise.

jakevdp:
We don’t have any control over numpy type promotion semantics, so I’m not sure what we can do here?

gnecula:
I understand that it is not clear how to solve this. We can try to document it better, but most people would not get to that piece of documentation.

How bad would it be to treat `nd.array` the same as Python scalars? I think that then we would not have this particular problem. 

One problem that I can imagine is that it would be hard to truly control the type of the inputs, e.g., `jnp.sin(np.float16(1.))` would be the same as `jnp.sin(1.)`, and the only way to control the type of the argument would be to use `jnp.float16(1.)` or `jnp.array(1., np.float16)`.

jakevdp:
Just to say it out loud, the issue is not that type promotion is non-associative. If you replace `a + b` with `jnp.add(a, b)` in the examples above, the results will be equivalent. Rather, the issue is that ``+`` is overloaded based on the types of the operands, and we're seeing the difference between numpy type promotion (in `np.add`) and JAX type promotion (in `jax.add`).

Fundamentally, then, this inconsistency comes from the design decision of making JAX functions polymorphic to numpy and JAX arrays. We could think about treating numpy arrays as weakly-typed, but this would still result in inconsistencies, particularly at JIT boundaries. If we wanted to be totally safe, we could make implicit casts of numpy objects to JAX objects an error – that would fix these inconsistencies, but at some cost to user convenience.

gnecula:
> the issue is not that type promotion is non-associative.

You are right, I changed the title.

gnecula:
I am closing this as won't fix. I mostly wanted this discussion recorded.

PR #6875, opened 2021-06-02
Expanded the type promotion documentation with a confusing case

gnecula:
Bug: #6874

I filed this as a bug, but I am assuming that it is not easy to fix, so
I also change the documentation.

### DIFF
--- a/docs/type_promotion.rst
+++ b/docs/type_promotion.rst
@@ -130,11 +130,14 @@ Jax's type promotion rules differ from those of NumPy, as given by
 :func:`numpy.promote_types`, in those cells highlighted with a green background
 in the table above. There are three key differences:
 
-* when promoting a Python scalar value against a typed JAX value of the same category,
+* When promoting a Python scalar value against a typed JAX value of the same category,
   JAX always prefers the precision of the JAX value. For example, ``jnp.int16(1) + 1``
-  will return ``int16`` rather than promoting to ``int64`` as in Numpy.
+  will return ``int16`` rather than promoting to ``int64`` as in NumPy.
+  Note that this applies only to Python scalar values; if the constant is a NumPy
+  array then the above lattice is used for type promotion.
+  For example, ``jnp.int16(1) + np.array(1)`` will return ``int64``.
 
-* when promoting an integer or boolean type against a floating-point or complex
+* When promoting an integer or boolean type against a floating-point or complex
   type, JAX always prefers the type of the floating-point or complex type.
 
 * JAX supports the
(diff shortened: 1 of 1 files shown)


jakevdp:
I think I'd instead say something like "This only applies to JAX operations; numpy has different type promotion semantics. Note that operators like `+` will dispatch based on the Python type of the two values being added. This means that `np.int64(1) + 1` will promote using numpy rules, whereas `jnp.int64(1) + 1` will promote using JAX rules."

jakevdp:
(As an aside, it may be worth mentioning that NumPy also has different promotion semantics for scalars and arrays, so `np.int16(1) + 1` is of type int64, but `np.int16([1]) + 1` is of type int16)

docs/type_promotion.rst
```diff
@@ -133,6 +133,13 @@ in the table above. There are three key differences:
...
   will return ``int16`` rather than promoting to ``int64`` as in Numpy.
+  Note that this applies only to Python scalar values; if the constant is a NumPy
+  array then the above lattice is used for type promotion.
+  For example, ``jnp.int16(1) + np.array(1)``
+  will return ``int64``. This may result in confusing non-associative behavior:
+  ``np.int16(1) +jnp.int16(2) + 3`` results in a ``int16``, while
+  ``3 + np.int16(1) + jnp.int16(2)`` results in an ``int64``, because the result
+  of ``3 + np.int16(1)`` is a NumPy array of dtype ``int64``.
 
 * when promoting an integer or boolean type against a floating-point or complex
```
8bitmp3, line 144: Nit suggestion for readability:

```suggestion
* When promoting a Python scalar value against a typed JAX value of the same category,
  JAX always prefers the precision of the JAX value. For example, ``jnp.int16(1) + 1``
  will return ``int16`` rather than promoting to ``int64`` as
  in NumPy. Note that this applies only to Python scalar values; if the constant is a NumPy
  array then the above lattice is used for type promotion.
  For example, ``jnp.int16(1) + np.array(1)``
  will return ``int64``. This may result in confusing non-associative behavior:
  ``np.int16(1) +jnp.int16(2) + 3`` results in a ``int16``, while
  ``3 + np.int16(1) + jnp.int16(2)`` results in an ``int64``, because the result
  of ``3 + np.int16(1)`` is a NumPy array of dtype ``int64``.

* When promoting an integer or boolean type against a floating-point or complex
```

Maybe there's a way to highlight that _JAX always prefers the precision of the JAX value_  and/or _this applies only to Python scalar values_ (**bolden then?**) since, for example, 3 + int16s is an int64 (!) and that's a JAX Gotcha 🔪 that may catch some NumPy users off-guard.

gnecula: @8bitmp3 Since I have more changes to make on my local copy, I have just incorporated your suggestions into my changes, rather than committing them (I do not know of an easy process by which to merge the committed suggestions and my subsequent local changes).

gnecula:
> I think I'd instead say something like "This only applies to JAX operations; NumPy has different type promotion semantics. Note that operators like `+` will dispatch based on the Python type of the two values being added. This means that, for example, `np.int16(1) + 1` will promote using numpy rules, whereas `jnp.int16(1) + 1` will promote using JAX rules. This can lead to potentially confusing non-associative promotion semantics when the two types of promotion are combined; for example with `np.int16(1) + 1 + jnp.int16(1)`"

I have added this as a separate paragraph further below. I have kept "Note that this applies only to Python scalar values; if the constant is a NumPy  array then the above lattice is used for type promotion." in the same place, because this is a separate issue.

jakevdp (approved):
Looks great - thanks!

PR #6877, opened 2021-06-02
Add a --jax_traceback_filtering flag to control the traceback filteri…

### LINKED ISSUE #6715: Add configuration option for disabling filtered stack traces
Some users report filtered stack traces interact badly with `pdb`.

We should add an `jax.config` option for disabling filtered stack traces, and hint at its existence in the filtered stack trace messages.

hawkinsp:
…ng mode.

Add a new traceback filtering mode that uses __tracebackhide__, and use it in IPython.

Validate enum configuration options.

Fixes #6715


froystig (approved):
Very nice! Thank you for picking up #6715 along the way, too.

PR #3821, opened 2020-07-22
DeviceArray.__iter__ returns DeviceArrays, without host sync

mattjj:
cross-ref #1583 #3330 

The current (before this PR) implementation of `DeviceArray.__iter__` fom #965 pulls the array back to the host and does the slicing on the resulting NumPy array, effectively:

```python
# current implementation
class DeviceArray:
  def __iter__(self):
    return iter(self._value)  # self._value is the cached ndarray version of the array
```

There are three issues we want to address here:
1. `list(device_array)` shouldn't return a list of `numpy.ndaray`s, and instead should return a list of `DeviceArray`s, regardless of how those are computed (this issue came up in #1583 and in some JAX user chats recently);
2. `key, subkey = random.split(key)` in op-by-op mode shouldn't incur device synchronization, but it does curently because the we effectively call `random.split(key).__iter__()` (this issue came up in #3330 and in JAX chats);
3. from #965, we don't want expressions like `list(device_array)` or `list(jnp.arange(10000))` to be slow to evaluate, where before #965 there could be a several-second compilation time and relatively slow (~100ms) execution time.

To address (1) we could either keep the bounce via NumPy and add some device_puts, or keep all the operations on the device. To address (2) we want to keep all the operations on the device. Currently that means we need to compile and execute XLA computations (rather than just doing something in the runtime).

To address (3), i.e. to keep things performant, most importantly we don't want to incur big compilation times, which basically means we don't want to compile computations with large output arity. We also don't want to dispatch lots of XLA computations, to keep execution time reasonable. We can balance the two by chunking things, so that for chunk size C and array size N we end up compiling O(1) computations, none having output arity more than C, and dispatching about 2 N / C computations (for each chunk, one to slice it out of the original array and one to explode it).

I timed the current and new implementation this way:

```python
import time
import jax.numpy as jnp
from jax.interpreters import xla

def clear_caches():
  xla.xla_primitive_callable.cache_clear()
  xla._xla_callable.cache_clear()

def time_things(x):
  # time to grab the first element, nothing compiled
  times = []
  for _ in range(100):
    clear_caches()
    tic = time.time()
    next(iter(x))
    times.append(time.time() - tic)
  print(f"time to grab first element, uncompiled: {sum(times) / len(times) * 1000} ms")

  # time to grab first element, compiled
  times = []
  for _ in range(100):
    tic = time.time()
    next(iter(x))
    times.append(time.time() - tic)
[...]

PR #6879, opened 2021-06-02
BUG: return JAX arrays rather than NumPy arrays in jnp.unravel_index

jakevdp:
`tuple(device_array)` returns a tuple of numpy arrays outside tracing, so this should be avoided here.


hawkinsp (approved):
I think we need to fix `tuple(device_array)` to return a tuple of device arrays at some point. I think there's a github issue open for this already. Leave a TODO to revert this change when that is done?

jakevdp:
Good call - I'll add the TODO and also start exploring making `DeviceArray.__iter__` avoid the round-trip to host. When traced it calls `lax.index_in_dim`, which might be a suitable approach in eager execution as well, particularly if we return a lazy iterator.

mattjj:
There's #3821, which last I checked was just blocked on a handfull of internal test failures.

PR #6889, opened 2021-06-03
Delay formatting of the name stack in translation rules

apaszke:
Now the name stack can store the stack of primitives inside which a
given translation rule is called. We can use this to provide more
informative errors and warnings for unsupported configurations of
primitives that cause XLA compilation (think `pmap` in `xmap`,
`with_sharing_constraint` in `jit`, etc.)

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -46,7 +46,7 @@ from jax.lib import xla_bridge as xb
 from jax.lib import xla_client
 from jax._src.traceback_util import api_boundary
 from jax._src.util import (partial, unzip2, unzip3, safe_map, safe_zip,
-                           split_list, cache, extend_name_stack)
+                           split_list, cache)
 from jax.tree_util import (tree_flatten, tree_unflatten, treedef_is_leaf,
                            treedef_children, treedef_tuple, tree_multimap,
                            tree_leaves, tree_structure)
@@ -336,7 +336,7 @@ def _while_loop_translation_rule(c, axis_env, name_stack, avals, backend, *args,
   x, _, z = split_list(cond_carry_elts, [cond_nconsts, body_nconsts])
   pred, = xla.jaxpr_subcomp(cond_c, cond_jaxpr.jaxpr, backend, axis_env,
                             _map(partial(xb.constant, cond_c), cond_jaxpr.consts),
-                            extend_name_stack(name_stack, 'cond'), *(x + z))
+                            name_stack.extend(while_p, 'cond'), *(x + z))
   if batched:
     scalar = ShapedArray((), np.bool_)
     or_ = xla.primitive_subcomputation(lax.or_p, scalar, scalar)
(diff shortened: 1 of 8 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -336,7 +336,7 @@ def _while_loop_translation_rule(c, axis_env, name_stack, avals, backend, *args,
   x, _, z = split_list(cond_carry_elts, [cond_nconsts, body_nconsts])
   pred, = xla.jaxpr_subcomp(cond_c, cond_jaxpr.jaxpr, backend, axis_env,
                             _map(partial(xb.constant, cond_c), cond_jaxpr.consts),
-                            extend_name_stack(name_stack, 'cond'), *(x + z))
+                            name_stack.extend(while_p, 'cond'), *(x + z))
```
gnecula, line 339: As I am reading through this linearly, I notice that this is one case where `extend_name_stack` was called without `xla.wrap_name`. This means that we are changing the way the XLA op_names will look. Is `while(cond)` better than `while/cond`? Perhaps.

jax/interpreters/xla.py
```diff
@@ -953,7 +979,7 @@ def _xla_call_translation_rule(c, axis_env, in_nodes, name_stack, backend, name,
   subc = xb.make_computation_builder(f"jit_{name}")
   args = [xb.parameter(subc, i, c.get_shape(n)) for i, n in enumerate(in_nodes)]
   out_nodes = jaxpr_subcomp(subc, call_jaxpr, backend, axis_env, (),
-                            extend_name_stack(name_stack, wrap_name(name, 'jit')), *args)
+                            name_stack.extend(xla_call_p, name), *args)
```
gnecula, line 982: Here I think that this will result in a XLA op_name that includes `xla_call(foo)`, which I think is less familiar to our users than `jit(foo)`. Maybe we can change the name of the xla_call primitive?

jax/interpreters/xla.py
```diff
@@ -1474,7 +1500,7 @@ def _named_call_translation_rule(c, axis_env, in_nodes, name_stack, *,
   subc = xb.make_computation_builder(name)
   args = [xb.parameter(subc, i, c.GetShape(n)) for i, n in enumerate(in_nodes)]
   out_nodes = jaxpr_subcomp(subc, call_jaxpr, backend, axis_env, (),
-                            extend_name_stack(name_stack, name), *args)
+                            name_stack.extend(name), *args)
```
gnecula, line 1503: It seems that this is the only place where we pass a string instead of a primitive. Even here I would pass the named_call primitive, if we want the stack to represent the stack of primitives.

tests/metadata_test.py
```diff
@@ -82,9 +82,9 @@ def f(x):
     self.assertRegex(hlo, 'op_type="cond"')
     self.assertRegex(hlo, 'op_name=".*cond\\[ linear=\\(False, False\\) \\]"')
     self.assertRegex(hlo, 'op_type="cos"')
-    self.assertRegex(hlo, 'op_name=".*cond/branch_0_fun/cos"')
+    self.assertRegex(hlo, 'op_name=".*cond\\(branch_0_fun\\)/cos"')
```
gnecula, line 85: nit: I would use `re.escape` instead of escaping by hand.

jax/interpreters/xla.py
```diff
@@ -59,6 +59,32 @@
 XlaComputationBuilder = Any  # xla_bridge._JaxComputationBuilder
 XlaExecutable = Any # xla_extension.LocalExecutable
 
+
```
gnecula, line 62: I think that the idea of having the context of a stack of nested primitives is a very good one. I would then require that `kind` is always a `Primitive`, and I would perhaps call it `PrimitiveStack`. Pushing this even further, why not include the primitive params in the stack, or have it be a `EqnStack`.

jax/interpreters/xla.py
```diff
@@ -59,6 +59,32 @@
...
+class NameStackFrame(NamedTuple):
```
gnecula, line 63: I think it is worth considering separating this mechanism from the generation of XlaOpMetadata. The latter can be based on this, but maybe not quite so literally. For example, we may prefer `jit(foo)` rather than `xla_call(foo)`. For `named_call` primitive we may prefer to just say `foo` rather than `named_call(foo)`. In essence, we should not require that we surface the internal names of the primitives in the XlaOpMetadata.

jax/_src/lax/control_flow.py
jekbradbury: My original choice to use `while/cond` and `while/body` was pretty arbitrary, and I don't think it's load-bearing.

jax/interpreters/xla.py
apaszke: I am open to that, but I think we should do it in a subsequent change.

jax/interpreters/xla.py
apaszke: Ok, but how would you suggest we do that? Implement an XLA specific pretty-printer of those objects instead of relying on their `__str__` methods?

gnecula: The cleanest way would be indeed to implement a NameStack pretty-printer for XLA. Or, if the only customization is the name of the primitive, we can go half-way there by passing an optional "op_metadata_name" field.  If the field is empty then we drop the "()" also, so that we can print `foo` instead of `named_call(foo)`.

hawkinsp:
Closing as stale.

PR #6910, opened 2021-06-07
jnp.where: add optional size argument

jakevdp:
Part of #3614; uses #6501

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1704,14 +1704,19 @@ _WHERE_DOC = """\
 At present, JAX does not support JIT-compilation of the single-argument form
 of :py:func:`jax.numpy.where` because its output shape is data-dependent. The
 three-argument form does not have a data-dependent shape and can be JIT-compiled
-successfully.
+successfully. Alternatively, you can specify the optional ``size`` keyword:
+if specified, the first ``size`` True elements will be returned; if there
+are fewer True elements than ``size`` indicates, the index arrays will be
+padded with zeros.
 """
 
 @_wraps(np.where, update_doc=False, lax_description=_WHERE_DOC)
-def where(condition, x=None, y=None):
+def where(condition, x=None, y=None, *, size=None):
   if x is None and y is None:
-    return nonzero(asarray(condition))
+    return nonzero(asarray(condition), size=size)
   else:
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1704,14 +1704,19 @@ def _where(condition, x=None, y=None):
 At present, JAX does not support JIT-compilation of the single-argument form
 of :py:func:`jax.numpy.where` because its output shape is data-dependent. The
 three-argument form does not have a data-dependent shape and can be JIT-compiled
-successfully.
+successfully. Alternatively, you can specify the optional ``size`` keyword.
+If specified, the first ``size`` True elements will be returned; if there
+are fewer True elements than ``size`` indicates, the index arrays will be
+padded with zeros.
```
8bitmp3, line 1710: Small suggestion - link the last sentences maybe? They may look more cohesive if together? 🤷‍♂️ 

```suggestion
successfully. Alternatively, you can specify the optional ``size`` keyword:
if specified, the first ``size`` True elements will be returned; and if there
are fewer True elements than ``size`` indicates, the index arrays will be
padded with zeros.
```

jakevdp: Done, thanks

tests/lax_numpy_test.py
```diff
@@ -4189,6 +4189,10 @@ def testWhereOneArgument(self, shape, dtype):
     args_maker = lambda: [rng(shape, dtype)]
     self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False)
 
+    # JIT compilation requires specifying the size statically:
+    jnp_fun = lambda x: jnp.where(x, size=np.size(x) // 2)
```
froystig, line 4193: Is `size(x) // 2` guaranteed sufficient under all random draws of the argument? If I understand correctly, [that's only the expected size](https://github.com/google/jax/blob/648b5d3265c3f1e28be39195a194981239ef2c6d/jax/test_util.py#L728), and there could be fewer zeros.

jakevdp: In general `size` may be larger or smaller than the true number, and `jnp.where` should be able to handle that.

froystig (approved):

PR #6914, opened 2021-06-07
Add auxiliary data support to `lax.custom_root`

nicholasjng:
Adds support for auxiliary data such as solver diagnostics in `lax.custom_root`. It builds on my previous pull request for the same feature in `lax.custom_linear_solve`. I added a test covering all of `jit, vmap, grad, jvp` transformations.

Notable differences in this case, at least from my point of view:
* It is not as clear whether `has_aux` should apply for both `solve` and `tangent_solve`; I opted for this in the first draft, though I can clearly see objections to that design choice.
* `custom_root`'s JVP is implemented with the `defjvp` decorator, which prevents me from padding the auxiliary tangents with `ad_util.Zero`s; since this is due to a detail in the custom derivatives workings of JAX, I wanted to get feedback on this and use `lax.zeros_like_array` in the meantime.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -2018,7 +2018,7 @@ def _split_root_args(args, const_lengths):
 
 
 @api_boundary
-def custom_root(f, initial_guess, solve, tangent_solve):
+def custom_root(f, initial_guess, solve, tangent_solve, has_aux=False):
   """Differentiably solve for a roots of a function.
 
   This is a low-level routine, mostly intended for internal use in JAX.
@@ -2049,6 +2049,8 @@ def custom_root(f, initial_guess, solve, tangent_solve):
       - For vector ``y``, you could use a linear solve with the Jacobian, if
         dimensionality of ``y`` is not too large:
         ``lambda g, y: np.linalg.solve(jacobian(g)(y), y)``.
+    has_aux: bool indicating whether the ``solve`` function returns
+      auxiliary data like solver diagnostics as a second argument.
 
   Returns:
     The result of calling solve(f, initial_guess) with gradients defined via
@@ -2060,11 +2062,11 @@ def custom_root(f, initial_guess, solve, tangent_solve):
       f, in_args_tree, guess_avals)
 
   in_tree, = treedef_children(in_args_tree)
-  _check_tree("f", "initial_guess", out_tree, in_tree)
+  _check_tree("f", "initial_guess", out_tree, in_tree, False)
 
   solve_jaxpr, solve_consts, solution_tree = _initial_style_jaxpr(
       partial(solve, _stop_gradient_fun(f)), in_args_tree, guess_avals)
-  _check_tree("solve", "initial_guess", solution_tree, in_tree)
+  _check_tree("solve", "initial_guess", solution_tree, in_tree, has_aux)
 
   def linearize_and_solve(x, b):
     unchecked_zeros, f_jvp = jax.linearize(f, x)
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -2049,6 +2049,8 @@ def custom_root(f, initial_guess, solve, tangent_solve):
       - For vector ``y``, you could use a linear solve with the Jacobian, if
         dimensionality of ``y`` is not too large:
         ``lambda g, y: np.linalg.solve(jacobian(g)(y), y)``.
+    has_aux: bool indicating whether the ``solve`` and ``tangent_solve`` functions
+      return auxiliary data like solver diagnostics as a second argument.
```
shoyer, line 2053: I don't see a good reason for including `has_aux` in `tangent_solve`, given that the aux values from tangent solve are always discarded. People might be surprised by how they are always replaced by zeros.

tests/lax_control_flow_test.py
```diff
@@ -2019,6 +2019,45 @@ def pos_def_solve(g, b):
...
+
+    # test vmap and jvp combined by jacfwd
+    fwd = jax.jacfwd(lambda x, y: root_aux(high_precision_dot(x, x.T), y), argnums=(0, 1))
+    expected_fwd = jax.jacfwd(lambda x, y: jnp.linalg.solve(high_precision_dot(x, x.T), y), argnums=(0, 1))
+
+    fwd_val, fwd_aux = fwd(a, b)
+    expected_fwd_val = expected_fwd(a, b)
+    self.assertAllClose(fwd_val, expected_fwd_val)
+    self.assertTrue(tree_util.tree_all(tree_util.tree_map(
+      lambda x: jtu.check_close(x, lax.zeros_like_array(x)), fwd_aux)))
```
shoyer, line 2059: I believe `jtu.check_close()` already contains an assert statement, via `np.testing.assert_allclose`.

So I think this might work:
```suggestion
    jtu.check_close(fwd_aux, tree_util.tree_map(jnp.zeros_like, fwd_aux))
```

tests/lax_control_flow_test.py
```diff
@@ -2019,6 +2019,45 @@ def pos_def_solve(g, b):
     jtu.check_grads(lambda x, y: linear_solve(high_precision_dot(x, x.T), y),
                     (a, b), order=2, rtol={jnp.float32: 1e-2})
 
+  def test_custom_root_with_aux(self):
+    def root_aux(a, b):
+      f = lambda x: high_precision_dot(a, x) - b
+      factors = jsp.linalg.cho_factor(a)
+      cho_solve = lambda f, b: tuple([jsp.linalg.cho_solve(factors, b), orig_aux])
```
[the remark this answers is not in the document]
njunge94: Fixed, thanks!

jax/_src/lax/control_flow.py
njunge94: Makes sense, I took it out and changed the docstring to mention only `solve`. 

As a small counterpoint: What do you think about giving users the option of outputting the aux values instead of symbolic zero tangents, e.g. by setting a flag?
It would technically break the JVP contract (which is why it's not great) but in turn, it would not inadvertently destroy the information from tangent_solve (and also `solve` in `custom_linear_solve`'s JVP, I guess), and we're operating on constants here anyhow, strictly speaking.

tests/lax_control_flow_test.py
njunge94: That did work indeed. Thanks for the suggestion!

jax/_src/lax/control_flow.py
shoyer: > As a small counterpoint: What do you think about giving users the option of outputting the aux values instead of symbolic zero tangents, e.g. by setting a flag?
> It would technically break the JVP contract (which is why it's not great) but in turn, it would not inadvertently destroy the information from tangent_solve (and also `solve` in `custom_linear_solve`'s JVP, I guess), and we're operating on constants here anyhow, strictly speaking.

Sure, in theory we could do this, which would allow for extracting aux information from explicit JVP calls.

But do we really care enough about forward mode autodiff for this to be worth the trouble? I'm not sure. There's no way we can use this trick for reverse mode, unfortunately.

njunge94: Good point. I guess in the tangent vs cotangent space thinking, currently the aux values would naturally be interpreted as tangents, since they are missing in reverse mode. Unless some higher-level aux handling machinery gets introduced, this is probably the most natural solution.

jax/_src/lax/control_flow.py
```diff
@@ -2114,6 +2119,9 @@ def _root_jvp(const_lengths, jaxprs, primals, tangents):
       params.f, params_dot.f)
   solution_dot = _map(
       operator.neg, linearize_and_solve(*itertools.chain(solution, rhs)))
+  # append aux and symbolic tangents to the solutions
+  solution += aux
+  solution_dot[f_out_vals:] = _map(lax.zeros_like_array, aux)
```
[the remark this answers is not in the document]
shoyer: Maybe slightly clearer, to indicate that we aren't overriding any values:
```suggestion
  solution_dot += _map(lax.zeros_like_array, aux)
```

shoyer (approved):

jax/_src/lax/control_flow.py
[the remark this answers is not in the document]
njunge94: Thanks for the suggestion! I did not look at the code for a few days and thought that an override was actually happening. Fixed.

PR #6915, opened 2021-06-07
jnp.argwhere: add optional size parameter for JIT compatibility

jakevdp:
Part of #3614

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4467,10 +4467,18 @@ def vander(x, N=None, increasing=False):
 
 ### Misc
 
+_ARGWHERE_DOC = """\
+Because the size of the output of ``argwhere`` is data-dependent, the function is not
+typically compatible with JIT. The JAX version adds the optional ``size`` argument, which
+specifies the size of the leading dimension of the output - it must be specified statically
+for ``jnp.argwhere`` to be traced. If ``size`` is specified, the indices of the first ``size``
+True elements will be returned; if there are fewer nonzero elements than `size` indicates,
+the index arrays will be zero-padded.
+"""
 
-@_wraps(np.argwhere)
-def argwhere(a):
-  result = transpose(vstack(nonzero(a)))
+@_wraps(np.argwhere, lax_description=_ARGWHERE_DOC)
+def argwhere(a, *, size=None):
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -4462,10 +4462,18 @@ def vander(x, N=None, increasing=False):
 
 ### Misc
 
+_ARGWHERE_DOC = """\
+Because the size of the output of ``argwhere`` is data-dependent, the function is not
+typically compatible with JIT. The JAX version adds the optional `size` argument which
+specifies the size of the leading dimension of the output; it must be specified statically
+for ``jnp.argwhere`` to be traced. If specified, the indices of the first `size` True elements
+will be returned; if there are fewer nonzero elements than `size` indicates, the index arrays
+will be zero-padded.
```
8bitmp3, line 4471: Just an idea - "If specified" on its own here may require some users to reread the whole thing, so this may just ease the cognitive overload a little:

```suggestion
typically compatible with JIT. The JAX version adds the optional `size` argument, which
specifies the size of the leading dimension of the output - it must be specified statically
for ``jnp.argwhere`` to be traced. If `size` is specified, the indices of the first `size` True
elements will be returned; and if there are fewer nonzero elements than `size` indicates, the index
arrays will be zero-padded.
```

Is the max char limit 80 or 100 btw

jakevdp: JAX conforms to no style guide – make the lines as long as you wish :grin:

jakevdp: Done, thanks

tests/lax_numpy_test.py
```diff
@@ -966,6 +966,10 @@ def testArgWhere(self, shape, dtype):
       self.skipTest("np.argwhere() result for scalar input changed in numpy 1.18.")
     self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False)
 
+    # JIT compilation requires specifying the size statically:
+    jnp_fun = lambda x: jnp.argwhere(x, size=np.size(x) // 2)
```
froystig, line 970: Same comment as [here](https://github.com/google/jax/pull/6910#discussion_r647714910).

froystig (approved):

PR #6918, opened 2021-06-07
Add LRU eviction policy to persistent compilation cache

colemanliyah:


jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
     if not key:
       raise ValueError("key cannot be empty")
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+
+    directory_size = self.get_size_of_directory()
+    new_file_size = 0
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+      new_file_size = self.creating_a_file(tmpdir, key, value)
```
skye, line 47: This is a clever way to get the size of the file before writing it to the cache directory! I think it's a little overkill here though, and we can just assume the new file size is equal to `len(value)` (i.e. the byte size of `value`). Writing to the temporary directory will also take extra time and potentially use too much disk space if `value` is large too.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+
+    if(new_file_size + directory_size > self._max_cache_size):
```
skye, line 49: Style nit: space after `if`, and you don't need parentheses

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+      last_time = 0
+      file_to_delete = ""
+
+      for files in os.listdir(self._path):
```
skye, line 53: Rename `files` to `file` or `filename` (since it's the just the one filename)

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+        file_to_inspect = os.path.join(self._path, files)
+
```
skye, line 55: nit: remove this blank line

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+        if os.path.getmtime(file_to_inspect)  > last_time:
+          last_time = os.path.getmtime(file_to_inspect)
```
skye, line 57: This is minor, but you should create an `mtime` variable so you don't have to call `getmtime` twice (here and one line above). This will save a tiny, probably unnoticeable bit of performance, so it doesn't really matter, but is a good practice in general.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+          file_to_delete = file_to_inspect
+
+      os.remove(file_to_delete)
```
skye, line 60: You might have to remove multiple files if `value` is larger than any existing file. You should add a test case for this, and fix this function to handle this case (I think right now it will only remove a maximum of 1 file).

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,42 @@ def put(self, key: str, value: bytes):
...
+
+    self.creating_a_file(self._path, key, value)
+
+  def creating_a_file(self, directory: str, key: str, value: bytes):
```
skye, line 64: Rename this to `_create_file`. Function names generally use the imperative mood and not present continuous (I had to look these grammar terms up lol -- basically `do_something` instead of `doing_something`). The leading underscore indicates this is a private method, since we don't expect users of the cache class to call this directly.

skye (changes requested):

jax/experimental/compilation_cache/file_system_cache.py
colemanliyah: I added a while loop here but is there a more efficient way to do this?

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
     """Adds new cache entry, possibly evicting older entries."""
     if not key:
       raise ValueError("key cannot be empty")
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+
```
skye, line 42: Remove blank line

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
...
+    self._evict_entries_if_necessary(key, value)
+
```
skye, line 44: Remove blank line

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
...
+    self._create_a_file(self._path, key, value)
+
+  def _evict_entries_if_necessary(self, key:str, value:bytes):
+
```
skye, line 48: Remove blank line. Where to use blanks lines is generally a subjective stylistic preference, but it's pretty standard to not put a blank line at the beginning of a function.

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+      del cache1
+      cache2 = FileSystemCache(tmpdir)
+      self.assertEqual(cache2._get_cache_directory_size(), 3)
+
+  def test_cache_is_full(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes = 5)
+      cache.put("foo", b"bar")
+      cache.put("second", b"file")
+      self.assertEqual(cache.get("foo"), None)
```
skye, line 83: Also test that "second" is present, to make sure that only "foo" was deleted and "second" was successfully added.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -12,15 +12,17 @@
...
 
-from typing import Optional
 import os
+import tempfile
+from typing import Optional
 
 class FileSystemCache:
 
-  def __init__(self, path: str):
+  def __init__(self, path: str, max_cache_size_bytes = 32000000000):
```
skye, line 21: For some reason we don't put spaces around the `=` sign when using keyword argument values. See https://google.github.io/styleguide/pyguide.html#36-whitespace (paragraph starting with "Never use spaces around").

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache1 = FileSystemCache(tmpdir)
+      cache1.put("foo", b"bar")
+      del cache1
+      cache2 = FileSystemCache(tmpdir)
+      self.assertEqual(cache2._get_cache_directory_size(), 3)
+
+  def test_cache_is_full(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes = 5)
```
skye, line 80: Same as above with the keyword argument `=`, should be `max_cache_size_bytes=5` with no spaces

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+      cache.put("foo", b"bar")
+      cache.put("second", b"file")
+      self.assertEqual(cache.get("foo"), None)
+
+  def test_delete_multiple_files(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes = 6)
```
skye, line 87: No spaces around keyword arg `=`

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("snd"), None)
+
+  def test_directory_same_after_deletion_from_cache(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache1 = FileSystemCache(tmpdir, max_cache_size_bytes = 5)
```
skye, line 95: No spaces around keyword arg `=`

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+      cache.put("second", b"file")
+      self.assertEqual(cache.get("foo"), None)
+
+  def test_delete_multiple_files(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes = 6)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("snd"), None)
```
skye, line 91: Test that "second" has been deleted and "third" is present

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,48 @@ def test_empty_key_get(self):
...
+
```
skye, line 92: Another edge case you should test: inserting a value into the cache that's above the size threshold, even if everything else is deleted. I'm not totally sure what the behavior should be actually! I think it shouldn't insert the value into the cache (to avoid going over the threshold), but it should also _not_ delete everything that's already there.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
...
+
+  def _create_a_file(self, directory: str, key: str, value: bytes):
+    """Creates a new file, returns the size of the new file"""
+    path_to_new_file = os.path.join(directory, key)
+    with open(path_to_new_file, "wb") as file:
       file.write(value)
+
+  def _get_cache_directory_size(self):
+    """Retrieves the current size of the directory, self.path"""
+    return sum(os.path.getsize(file_in_directory) for file_in_directory in os.scandir(self._path) if file_in_directory.is_file())
```
skye, line 74: This line is very long. You can just use `file` or even `f` for `file_in_directory`. A general principle is that variable names can be shorter if they're only used in a very small amount of code, since it's easy to see where the variable is coming from so the name is less important to understanding what it is. You might also put a line break before the `if`.

skye (changes requested):

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,69 @@ def test_empty_key_get(self):
...
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      self.assertEqual(cache.get("foo"), None)
+
+  def test_file_removed(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=5)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      self.assertEqual(cache.get("second"), b"two")
```
skye, line 90: You can add this assert to the end of `test_cache_is_full`, and delete this `test_file_removed` case (basically combine the two tests into one). It's common to have multiple asserts in a single test, you don't need to create a new test case for each one.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +38,33 @@ def put(self, key: str, value: bytes):
...
+      last_time = 0.0
+      file_to_delete = ""
+      for file_name in os.listdir(self._path):
+        file_to_inspect = os.path.join(self._path, file_name)
+        mtime = os.path.getmtime(file_to_inspect)
+        if mtime > last_time:
+          last_time = mtime
+          file_to_delete = file_to_inspect
+      os.remove(file_to_delete)
+      directory_size = self._get_cache_directory_size()
```
skye, line 60: You can just call `self._get_cache_directory_size()` directly in the while loop condition on L50, and remove the directory_size variable (here and L45)

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,69 @@ def test_empty_key_get(self):
...
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("second"), None)
+
+  def test_big_file_exists(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=6)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("third"), b"three")
```
skye, line 106: Same here, you can put this line in `test_delete_multiple_files`, and delete this test case

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,69 @@ def test_empty_key_get(self):
...
+      cache.put("second", b"two")
+      self.assertEqual(cache.get("second"), b"two")
+
+  def test_delete_multiple_files(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=6)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("second"), None)
```
skye, line 98: Add another assert for `cache.get("foo")`

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,69 @@ def test_empty_key_get(self):
...
+
+  def test_big_file_exists(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=6)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      cache.put("third", b"three")
+      self.assertEqual(cache.get("third"), b"three")
+
```
skye, line 107: Also add tests to make sure that when there are multiple files that could be removed to make room, the least-recently-accessed file is removed. Create a test case for each of these:

1. Adding at least 3 files until one needs to be removed. Make sure that the first one added is removed, and that the other ones are still there.
2. Adding at least 2 files, `get` the first one added, and then add more file(s) until one needs to be removed. Assert that the least-recently-accessed one is deleted (i.e. not the one you called `get` on), and that the others are still there.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +38,33 @@ def put(self, key: str, value: bytes):
...
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+    self._evict_entries_if_necessary(key, value)
+    self._create_a_file(self._path, key, value)
+
+  def _evict_entries_if_necessary(self, key:str, value:bytes):
+    directory_size = self._get_cache_directory_size()
+    new_file_size = len(value)
+
+    assert (new_file_size <= self._max_cache_size_bytes), "This compilation file is larger than the size of the cache"
```
skye, line 48: I don't think we should raise an error here, because I'm guessing people will want their program to keep running even if the cache is too small (I'm not 100% sure about this, but seems better to not risk crashing people's programs when they don't expect it). Instead, you can use `warnings.warn` (https://docs.python.org/3/library/warnings.html#warnings.warn) to print the message as a warning, without crashing the program.

Unlike raising an exception or failing an assert, calling `warnings.warn` doesn't cause the current function to stop running (since it's not a crash), so you'll also have to change this function + `put` to return without creating the new file. I suggest making `_evict_entries_if_necessary` return True if there's enough room to write the file, and False if not, and then using the return value in `get` to decide whether to call `_create_a_file`.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +38,33 @@ def put(self, key: str, value: bytes):
     """Adds new cache entry, possibly evicting older entries."""
     if not key:
       raise ValueError("key cannot be empty")
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+    self._evict_entries_if_necessary(key, value)
+    self._create_a_file(self._path, key, value)
+
+  def _evict_entries_if_necessary(self, key:str, value:bytes):
```
skye, line 44: Style thing: space after the `:` for type annotations (e.g. `key: str`) (why is there a space here and not for `=`?? I don't know)

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +38,33 @@ def put(self, key: str, value: bytes):
...
+    directory_size = self._get_cache_directory_size()
+    new_file_size = len(value)
+
+    assert (new_file_size <= self._max_cache_size_bytes), "This compilation file is larger than the size of the cache"
+
+    while new_file_size + directory_size > self._max_cache_size_bytes:
+      last_time = 0.0
+      file_to_delete = ""
```
skye, line 52: Style nit: use `None` instead of `""` to indicate that file_to_delete hasn't been set yet

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +38,33 @@ def put(self, key: str, value: bytes):
...
+      for file_name in os.listdir(self._path):
+        file_to_inspect = os.path.join(self._path, file_name)
+        mtime = os.path.getmtime(file_to_inspect)
+        if mtime > last_time:
+          last_time = mtime
+          file_to_delete = file_to_inspect
+      os.remove(file_to_delete)
```
skye, line 59: Right before calling `remove`, add `assert file_to_delete`. This is to make sure the above logic is successfully finding a file to delete (which we always expect to be true).

tests/file_system_cache_test.py
```diff
@@ -56,5 +56,69 @@ def test_empty_key_get(self):
...
+  def test_file_bigger_than_cache(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=2)
+      with self.assertRaisesRegex(AssertionError, r"This compilation file is larger than the size of the cache"):
+        cache.put("foo", b"bar")
+
+  def test_directory_same_after_deletion_from_cache(self):
```
skye, line 114: I actually think you can delete this test case :D You already test in other places that the cache directory doesn't change across `FileSystemCache` instances, so I don't think you need to test every possible variation of this pattern. In general more test cases are better, but having too many redundant test cases can make it hard to read the test file later to understand what's being tested and what's not.

skye (changes requested):

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
...
     if not key:
       raise ValueError("key cannot be empty")
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+    if self._evict_entries_if_necessary(key, value):
+      self._create_a_file(self._path, key, value)
+    else:
+      warnings.warn("This compilation file is larger than the size of the cache")
+
+  def _evict_entries_if_necessary(self, key: str, value: bytes) -> bool:
```
skye, line 47: Add a docstring explaining what the return value means. Something like:
`"""Returns True if there's enough space to add 'value', False otherwise."""`

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
     """Adds new cache entry, possibly evicting older entries."""
     if not key:
       raise ValueError("key cannot be empty")
-      #TODO(colemanliyah):implement eviction
-    with open(os.path.join(self._path, key), "wb") as file:
+    if self._evict_entries_if_necessary(key, value):
+      self._create_a_file(self._path, key, value)
+    else:
+      warnings.warn("This compilation file is larger than the size of the cache")
```
skye, line 45: It's nice to include `value`'s size and the max size in the warning, so people can understand exactly what's going wrong. Users may not understand what "compilation file" means either. I suggest something like:
`f"Cache value of size {len(value)} is larger than the max cache size of {self._max_cache_size_bytes}"`
(If you haven't seen it before, the above is a [formatted string literal](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals). It's the best way to include variable values in a string in Python.)

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,36 @@ def put(self, key: str, value: bytes):
...
+        file_to_inspect = os.path.join(self._path, file_name)
+        atime = os.stat(file_to_inspect).st_atime
+        if atime < last_time:
+          last_time = atime
+          file_to_delete = file_to_inspect
+      assert file_to_delete
+      os.remove(file_to_delete)
+    return True
+
+  def _create_a_file(self, directory: str, key: str, value: bytes):
```
skye, line 66: This function is so small and simple now, and only called in one place, that I would copy/paste it directly where you call it and get rid of this function. (Just FYI, I think `_get_cache_directory_size` should still be its own helper function because even though it's only one line, it's pretty fancy logic that merits its own function to describe what its doing.)

tests/file_system_cache_test.py
```diff
@@ -56,5 +57,63 @@ def test_empty_key_get(self):
...
+      self.assertEqual(cache.get("foo"), None)
+      self.assertEqual(cache.get("second"), None)
+      self.assertEqual(cache.get("third"), b"three")
+
+  def test_least_recently_accessed_file(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=6)
+      cache.put("foo", b"bar")
+      cache.put("second", b"two")
+      #Added sleeps because these operations were executing too fast that the access time could not be captured properly
```
skye, line 104: Minor nits on verb tense and line length, please change to:
```python
      # Sleep because otherwise these operations execute too fast and
      # the access time isn't captured properly.
```

tests/file_system_cache_test.py
```diff
@@ -56,5 +57,63 @@ def test_empty_key_get(self):
...
+      time.sleep(1)
+      cache.get("foo")
+      cache.put("third", b"the")
+      self.assertEqual(cache.get("foo"), b"bar")
+      self.assertEqual(cache.get("second"), None)
+
+  @jtu.ignore_warning(message="This compilation file is larger than the size of the cache")
```
skye, line 111: You'll have to change this too when you change the warning message.

skye (changes requested):
Looking good!

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -40,11 +40,14 @@ def put(self, key: str, value: bytes):
     if not key:
       raise ValueError("key cannot be empty")
     if self._evict_entries_if_necessary(key, value):
-      self._create_a_file(self._path, key, value)
+      path_to_new_file = os.path.join(self._path, key)
+      with open(path_to_new_file, "wb") as file:
+        file.write(value)
     else:
-      warnings.warn("This compilation file is larger than the size of the cache")
+      warnings.warn(("Cache value of size {} is larger than the max cache size of {}").format(len(value), self._max_cache_size_bytes))
```
skye, line 47: You should use a formatted string literal (https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals) instead of the `format` function. Also add a line break to keep lines less than about 80 characters long. Like this:

```python
      warnings.warn(f"Cache value of size {len(value)} is larger than the "
                    "max cache size of {self._max_cache_size_bytes}")
```

tests/file_system_cache_test.py
```diff
@@ -101,14 +101,15 @@ def test_least_recently_accessed_file(self):
...
+      # Sleep because otherwise these operations execute too fast and
+      # the access time isn't captured properly.
       time.sleep(1)
       cache.get("foo")
       cache.put("third", b"the")
       self.assertEqual(cache.get("foo"), b"bar")
       self.assertEqual(cache.get("second"), None)
 
-  @jtu.ignore_warning(message="This compilation file is larger than the size of the cache")
+  @jtu.ignore_warning(message=("Cache value of size {} is larger than the max cache size of {}").format(len(b"bar"), 2))
```
skye, line 112: You don't need to use a format string here at all, you can just hardcode the actual warning that the test produces (something like "Cache value of size 3 is larger than the max cache size of 2"). This makes it easier to read.

skye (changes requested):
Almost there! @zhangqiaorjc I think this is ready for you to review as well.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -12,15 +12,17 @@
...
 
-from typing import Optional
 import os
+from typing import Optional
+import warnings
 
 class FileSystemCache:
 
-  def __init__(self, path: str):
+  def __init__(self, path: str, max_cache_size_bytes=32000000000):
```
zhangqiaorjc, line 21: I wonder if 32 * 2**30 bytes (or 32 GiB) instead of 32 GB is better?

For example, GCP is using GiB for disk capacity

"For Compute Engine, disk size, machine type memory, and network usage are calculated in gigabytes (GB), where 1 GB is 230 bytes. This unit of measurement is also known as a gibibyte (GiB)."

https://cloud.google.com/compute/disks-image-pricing

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -37,6 +39,34 @@ def put(self, key: str, value: bytes):
...
+                    f" the max cache size of {self._max_cache_size_bytes}")
+
+  def _evict_entries_if_necessary(self, key: str, value: bytes) -> bool:
+    """Returns True if there's enough space to add 'value', False otherwise."""
+    new_file_size = len(value)
+
+    if new_file_size >= self._max_cache_size_bytes:
+      return False
+
+    while new_file_size + self._get_cache_directory_size() > self._max_cache_size_bytes:
```
zhangqiaorjc, line 57: this walks the entire directory on every while loop condition..it's a bit expensive

tests/file_system_cache_test.py
```diff
@@ -56,5 +57,64 @@ def test_empty_key_get(self):
...
+      cache1 = FileSystemCache(tmpdir)
+      cache1.put("foo", b"bar")
+      del cache1
+      cache2 = FileSystemCache(tmpdir)
+      self.assertEqual(cache2._get_cache_directory_size(), 3)
+
+  def test_cache_is_full(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir, max_cache_size_bytes=6)
+      cache.put("foo", b"bar")
```
zhangqiaorjc, line 82: it seems easier to read if you change "foo" to "first" and "bar" to "one" here and below

zhangqiaorjc (commented):
Great job, Liyah!

jax/experimental/compilation_cache/file_system_cache.py
skye: I agree this is expensive, but I think we should improve it after we have things working end-to-end and have microbenchmarks (so we can see how much it helps!). Maybe leave a TODO here though.

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -54,6 +54,7 @@ def _evict_entries_if_necessary(self, key: str, value: bytes) -> bool:
     if new_file_size >= self._max_cache_size_bytes:
       return False
 
+    #TODO: Refactor this section so in the future the whole directory doesn't need to be checked
```
skye, line 57: TODO formatting: use `# TODO(colemanliyah): blah blah` (space after `#`, put your username in the TODO to indicate you're likely to address it, or can at least answer questions about it)

Also, you can remove the "in the future", all TODOs are in the future :D

skye (approved):
One more tiny comment, otherwise LGTM!

skye:
Oh except a test is failing too

zhangqiaorjc (approved):

PR #6933, opened 2021-06-09
Simplify multi slice implementation

lgeiger:
This PR simplifies the multi slice implementation by switching from `lax.reshape` to `lax.squeeze` for removing dimensions.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5954,7 +5954,7 @@ def _multi_slice(arr,
   for starts, limits, removed in safe_zip(start_indices, limit_indices, removed_dims):
     sliced = lax.slice(arr, starts, limits)
     if removed:
-      sliced = sliced.reshape(np.delete(sliced.shape, removed_dims))
+      sliced = lax.squeeze(sliced, removed)
     results.append(sliced)
   return results
 setattr(_DeviceArray, "_multi_slice", _multi_slice)
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -348,8 +348,8 @@ for _t in array_types:
   shard_arg_handlers[_t] = _shard_array
 
 def _shard_device_array(x, devices, indices):
-  start_indices, limit_indices, removed_dims = map(tuple, unzip3(
-      _as_slice_indices(x, idx) for idx in indices))
+  start_indices, limit_indices, removed_dims = unzip3(
+      _as_slice_indices(x, idx) for idx in indices)
   shards = x._multi_slice(start_indices, limit_indices, removed_dims)
   return device_put(shards, devices)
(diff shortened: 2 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -5954,7 +5954,7 @@ def _multi_slice(arr,
   for starts, limits, removed in safe_zip(start_indices, limit_indices, removed_dims):
     sliced = lax.slice(arr, starts, limits)
     if removed:
-      sliced = sliced.reshape(np.delete(sliced.shape, removed_dims))
```
lgeiger, line 5957: From reading the code, I think this should be using `removed` instead of `removed_dims`, but please let me know if I am missing something.

jax/interpreters/pxla.py
```diff
@@ -348,8 +348,8 @@ def _shard_array(x, devices, indices):
   shard_arg_handlers[_t] = _shard_array
 
 def _shard_device_array(x, devices, indices):
-  start_indices, limit_indices, removed_dims = map(tuple, unzip3(
```
lgeiger, line 351: `unzip3` already returns 3 tuples. No need to convert again.

jakevdp (commented):
Thanks for the PR - sorry this fell through the cracks.

I'm a bit concerned about the multi-slice change, not because it looks incorrect (your changes look reasonable to me!) but because I don't fully understand what the code is doing, and because this is only used in sharded device puts, which I don't think have particularly good test coverage.

Can you come up with an example of where using `removed` vs `removed_dims` here changes the result, and confirm that your version produces the expected result?

jakevdp:
So, it looks like in every test case we have, each entry of `removed_dims` is identical, such that `np.delete(shape, removed_dims)` returns the same thing as `np.delete(shape, removed_dims[i])`

So I'm fairly certain this PR corrects a bug, I'd just feel better about it if we could find a case that your change fixes!

jakevdp:
I'm going to assign to @skye because I think she has the most context on this part of the code.

lgeiger:
> Can you come up with an example of where using removed vs removed_dims here changes the result, and confirm that your version produces the expected result?

I tried to find a test case for that as well, but didn't find any obvious one. But I will take another look.
Although to me it seems that it is just due to the behaviour of `np.delete` that this works with the higher dimensional `removed_dims` as long as all values are the same.

skye:
I think we have the constraint that `removed_dims` always has the same values! My reasoning: `_multi_slice` is used to extract the per-device slices from an input array to run a distributed computation (e.g. a pmap), and all devices get identically-shaped slices. So `_multi_slice`, and all of the index-based functionality leading up to it

skye:
Oops accidentally hit the close button

skye (approved):
Thanks for the fixes!

skye:
Ah sorry, forgot to merge this! Doing so now.

PR #6948, opened 2021-06-11
Support value computation of associated Legendre functions.

tlu7:
@jakevdp 

Exposed the value computation of the associated Legendre functions as a stand-alone function as discussed.  Made normalization an option for users.

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -914,7 +914,7 @@ def _gen_associated_legendre(l_max: int,
   return p
 
 
-def lpmn(m, n, z):
+def lpmn(m: int, n: int, z: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
   """The associated Legendre functions (ALFs) of the first kind.
 
   Args:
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/special.py
```diff
@@ -956,3 +956,56 @@ def lpmn(m, n, z):
...
   return (p_vals, p_derivatives)
+
+
+def lpmn_values(m, n, z, is_normalized):
+  """The associated Legendre functions (ALFs) of the first kind.
+
+  Unlike `lpmn`, this function only computes the values of ALFs.
+  The ALFs of the first kind can be used in spherical harmonics. The
+  spherical harmonic of degree `l` and order `m` can be written as
+  `Y_l^m(θ, φ) = N_l^m * P_l^m(cos(θ)) * exp(i m φ)`, where `N_l^m` is the
```
jakevdp, line 967: You could use a `:math:` directive here to make this expression render more cleanly in the HTML docs.

tests/lax_scipy_test.py
```diff
@@ -266,6 +266,42 @@ def testLpmn(self, l_max, num_z):
...
+          a_normalized[m, l] = c2 * a[m, l]
+
+        return a_normalized
+
+    # The results from scipy are not normalized and the comparison requires
+    # normalizing the results.
+    expected_p_vals_normalized = apply_normalization(expected_p_vals)
+
+    self.assertAllClose(actual_p_vals, expected_p_vals_normalized, rtol=0.00093, atol=0.72)
+
```
jakevdp, line 304: Also please add a `self._CompileAndCheck` test for this function (this will ensure that it works correctly with JIT).

jakevdp (changes requested):
Looks good! Have you checked to see whether this is compatible as-written with autodiff? Since you alredy have an analytic expression for the gradient in `lpmn`, it might be nice to add that as a custom JVP rule for the `lpmn_values` function.

Also, we could use `jtu.check_grads()` within the unit tests to make sure the analytic gradient approximately matches the numerical gradient.

jax/_src/scipy/special.py
tlu7: Done. Thanks!

tests/lax_scipy_test.py
tlu7: Thanks for the suggestion. Add  `self._CompileAndCheck` test for both `lpmn` and `lpmn_values`.

tlu7:
> Looks good! Have you checked to see whether this is compatible as-written with autodiff? Since you alredy have an analytic expression for the gradient in `lpmn`, it might be nice to add that as a custom JVP rule for the `lpmn_values` function.
> 
> Also, we could use `jtu.check_grads()` within the unit tests to make sure the analytic gradient approximately matches the numerical gradient.

Thanks for the comments! As discussed, I will add the custom JVP rule in the next PR, which requires binding the normalization factor to the analytical gradient computation.  PTAL.

jax/_src/scipy/special.py
```diff
@@ -914,6 +914,7 @@ def body_fun(i, p_val):
   return p
 
 
+@partial(jit, static_argnums=(0, 1))
```
jakevdp, line 917: It's better to avoid `jit` with `static_argnums` when defining user-facing functions, becuase this currently causes the function to fail when called with keyword arguments. In this case, I'd probably leave off the `JIT`: the user can always wrap with JIT if they'd like it to be compiled, and since `_gen_associated_legendre` is JIT compiled, this annotation does not add much.

jax/_src/scipy/special.py
```diff
@@ -956,3 +957,61 @@ def lpmn(m, n, z):
   p_derivatives = _gen_derivatives(p_vals, z, is_normalized)
 
   return (p_vals, p_derivatives)
+
+
+@partial(jit, static_argnums=(0, 1, 3))
```
jakevdp, line 962: Same here: probably better not to add the JIT annotation at this level.

tests/lax_scipy_test.py
```diff
@@ -266,6 +266,54 @@ def testLpmn(self, l_max, num_z):
...
+          a_normalized[m, l] = c2 * a[m, l]
+
+        return a_normalized
+
+    # The results from scipy are not normalized and the comparison requires
+    # normalizing the results.
+    expected_p_vals_normalized = apply_normalization(expected_p_vals)
+
+    with self.subTest('Test accuracy.'):
+      self.assertAllClose(actual_p_vals, expected_p_vals_normalized, rtol=0.00093, atol=0.72)
```
jakevdp, line 310: This seems like a pretty large `atol`. Is it necessary? Also, multiple significant figures in a test tolerance seems strange

jakevdp (changes requested):

tests/lax_scipy_test.py
tlu7: Thanks for guiding me into the details here. It is a bug.  The indentation of `return a_normalized` in `apply_normalization` is wrong such that `expectred_p_vals_normalized` only contains the first row.

tests/lax_scipy_test.py
```diff
@@ -291,23 +291,21 @@ def testNormalizedLpmnValues(self, l_max, num_z):
...
       a_normalized = np.zeros_like(a)
       for m in range(num_m):
         for l in range(num_l):
           c0 = (2.0 * l + 1.0) * osp_special.factorial(l - m)
           c1 = (4.0 * np.pi) * osp_special.factorial(l + m)
           c2 = np.sqrt(c0 / c1)
           a_normalized[m, l] = c2 * a[m, l]
-
-        return a_normalized
+      return a_normalized
```
tlu7, line 301: @jakevdp here is the bug fixed.

jax/_src/scipy/special.py
tlu7: Thanks! Done.

jax/_src/scipy/special.py
tlu7: Thanks for the explanations! Removed `JIT`. Done.

tests/lax_scipy_test.py
jakevdp: looks good!

jakevdp (approved):
Looks good!

Please squash the commits into one, and then we can merge this. Thanks for working on this functionality!

PR #4030, opened 2020-08-12
[jax2tf] Clean up code for XlaGather, experimental_compile not necessary

gnecula:
Now that XlaGather has been fixed in XLA, we do not need to use
experimental_compile workaround (which was not sorking anyway when
put in a SavedModel).

This fix requires a recent tf-nightly installation.

PR #6964, opened 2021-06-14
[jax2tf] Add a best-effort conversion for gather with enable_xla=False

gnecula:
In an attempt to support lax.gather in absence of XLA, e.g., for
TFjs converter, we add an alternative conversion for the case
when enable_xla=False. Previously, the conversion would fail in this case.

The important parts of the code are lifeted from an old PR #3486, which was
then removed in #4030 (to avoid having to think about out-of-bounds accesses).

Now we added clipping of the indices so the behavior should match XLA.

### DIFF
--- a/jax/experimental/jax2tf/g3doc/primitives_with_limited_support.md
+++ b/jax/experimental/jax2tf/g3doc/primitives_with_limited_support.md
@@ -1,6 +1,6 @@
 # Primitives with limited support for jax2tf
 
-*Last generated on (YYYY-MM-DD): 2021-06-14*
+*Last generated on (YYYY-MM-DD): 2021-06-15*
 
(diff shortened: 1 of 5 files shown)


gnecula:
@marcvanzee PTAL. The out-of-bounds index access is not resolved here. With TF eager we would get a loud error, with TF compiled it works. I am not sure what TF.js does with this (whether it likes the tf.gather and how it behaves for OOB accesses).

gnecula:
> @marcvanzee PTAL. The out-of-bounds index access is not resolved here. With TF eager we would get a loud error, with TF compiled it works. I am not sure what TF.js does with this (whether it likes the tf.gather and how it behaves for OOB accesses).

I have fixed also the OOB issues.

PR #6978, opened 2021-06-15
get_cache_key implementation with unit tests

colemanliyah:


skye (changes requested):
I didn't get to the test file yet, will do so later today!

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -0,0 +1,75 @@
...
+              compile_options_obj.argument_layouts)
+    _hash_int(hash_obj, compile_options_obj.parameter_is_tupled_arguments)
+    _hash_executable_build_options(hash_obj, compile_options_obj.executable_build_options)
+    _hash_bool(hash_obj, compile_options_obj.tuple_arguments)
+    _hash_int(hash_obj, compile_options_obj.num_replicas)
+    _hash_int(hash_obj, compile_options_obj.num_partitions)
+    if compile_options_obj.device_assignment is not None:
+        hash_obj.update(compile_options_obj.device_assignment.serialize())
+
+def _hash_executable_build_options(hash_obj, executable_obj):
```
skye, line 41: Put a similar assert like in `_hash_compile_options`?

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    _hash_bool(hash_obj, debug_obj.xla_cpu_enable_xprof_traceme)
+    _hash_bool(hash_obj, debug_obj.xla_llvm_disable_expensive_passes)
+    _hash_bool(hash_obj, debug_obj.xla_test_all_input_layouts)
+
+def _hash_libtpu(hash_obj, libtpu_obj):
+    _hash_string(hash_obj, libtpu_obj.platform)
+    _hash_string(hash_obj, libtpu_obj.platform_version)
+    _hash_string(hash_obj, libtpu_obj.runtime_type)
+
+def _hash_int(hash_obj, int_var):
```
skye, line 68: You should add unit tests for these new helper function (int, bool, string)

skye (changes requested):

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -0,0 +1,75 @@
...
+def get_cache_key(xla_computation, compile_options) -> str:
+    hash_obj = hashlib.sha256()
+    hash_obj.update(xla_computation.as_serialized_hlo_module_proto())
+    _hash_compile_options(hash_obj, compile_options)
+    hash_obj.update(bytes(jax.lib.version))
+    _hash_libtpu(hash_obj, jax.lib.xla_bridge.get_backend())
+    return hash_obj.digest().hex()
+
+def _hash_compile_options(hash_obj, compile_options_obj):
+    assert (len(dir(compile_options_obj)) == 31), "The compile options are not consistent with the version you ran with for previously compiled programs."
```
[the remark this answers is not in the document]
colemanliyah: I took out the outer parenthesis too because mypy wasn't happy with them...I was getting the error "Assert is always True"...likewise I added parenthesis around the string part in order for python to be happy with the line breaks.

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -0,0 +1,89 @@
...
+    _hash_bool(hash_obj, debug_obj.xla_cpu_fast_math_honor_nans)
+    _hash_bool(hash_obj, debug_obj.xla_cpu_fast_math_honor_division)
+    _hash_bool(hash_obj, debug_obj.xla_cpu_fast_math_honor_functions)
+    _hash_bool(hash_obj, debug_obj.xla_gpu_enable_fast_min_max)
+    _hash_int(hash_obj, debug_obj.xla_backend_optimization_level)
+    _hash_bool(hash_obj, debug_obj.xla_cpu_enable_xprof_traceme)
+    _hash_bool(hash_obj, debug_obj.xla_llvm_disable_expensive_passes)
+    _hash_bool(hash_obj, debug_obj.xla_test_all_input_layouts)
+
+def _hash_libtpu(hash_obj, libtpu_obj):
```
skye, line 77: Rename this function to `_hash_platform`, and `libtpu_obj` to `backend`.

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,120 @@
...
+    hash_obj2 = self.create_sudo_hash_obj(cc._hash_compile_options, compile_options_filled)
+    hash_obj3 = self.create_sudo_hash_obj(cc._hash_compile_options, compile_options_not_filled)
+    self.assertEqual(hash_obj1.hex(), hash_obj2.hex())
+    self.assertNotEqual(hash_obj1.hex(), hash_obj3.hex())
+
+  def test_executable_build_options(self):
+    compile_options_not_filled = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    compile_options_filled = self.fill_compile_options()
+    hash_obj1 = self.create_sudo_hash_obj(cc._hash_executable_build_options, compile_options_filled.executable_build_options)
```
skye, line 44: Add a line break to make this line shorter (here and similarly-long lines below)

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,120 @@
...
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    hash_obj1 = self.create_sudo_hash_obj(cc._hash_debug_options, compile_options.executable_build_options.debug_options)
+    hash_obj2 = self.create_sudo_hash_obj(cc._hash_debug_options, compile_options.executable_build_options.debug_options)
+    self.assertEqual(hash_obj1.hex(), hash_obj2.hex())
+
+  def test_libtpu(self):
+    hash_obj1 = self.create_sudo_hash_obj(cc._hash_libtpu, jax.lib.xla_bridge.get_backend())
+    hash_obj2 = self.create_sudo_hash_obj(cc._hash_libtpu, jax.lib.xla_bridge.get_backend())
+    self.assertEqual(hash_obj1.hex(), hash_obj2.hex())
```
skye, line 60: If this test is running on a non-CPU backend, you can get the CPU backend to check the non-equal case! Something like this:
```python
if jax.lib.xla_bridge.get_backend().platform != "cpu":
  cpu_backend = jax.lib.xla_bridge.get_backend("cpu")
  <create hash of cpu backend>
  <assertNotEqual to non-cpu backend hash>
```
You need the if statement to avoid accidentally comparing the same backend and asserting it's not equal when it is.

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,120 @@
...
+    self.assertEqual(cc.get_cache_key(computation, compile_options),
+                     cc.get_cache_key(computation, compile_options))
+
+  def tets_different_hash_key(self):
+    computation = jax.xla_computation(lambda x, y: x + y)(1, 1)
+    compile_options_not_filled = jax.lib.xla_bridge.get_compile_options(
+                       num_replicas=1, num_partitions=1)
+    compile_options_filled = self.fill_compile_options()
+    self.assertNotEqual(cc.get_cache_key(computation, compile_options_not_filled),
+                        cc.get_cache_key(computation, compile_options_filled))
```
skye, line 99: Also test that two different computations with the same compile options produce different hashes

skye (changes requested):

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,120 @@
...
+    self.assertEqual(type(cc.get_cache_key(computation, compile_options)), str)
+
+  def test_compile_options(self):
+    compile_options_not_filled = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    compile_options_filled = self.fill_compile_options()
+    hash_obj1 = self.create_sudo_hash_obj(cc._hash_compile_options, compile_options_filled)
+    hash_obj2 = self.create_sudo_hash_obj(cc._hash_compile_options, compile_options_filled)
+    hash_obj3 = self.create_sudo_hash_obj(cc._hash_compile_options, compile_options_not_filled)
+    self.assertEqual(hash_obj1.hex(), hash_obj2.hex())
```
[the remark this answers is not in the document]
colemanliyah: Yes, however I am using this helper function for tests like "test_hash_bool" and I just want the digest part, not .hex()

tests/compilation_cache_test.py
colemanliyah: I'm seeing now that the project doc only mentions hashing the platform version value as part of the key...would you like me to take out the other things I am hashing for the libtpu in my main file?

tests/compilation_cache_test.py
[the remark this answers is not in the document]
skye: Can you use just the digest here, or use hex() in the other tests? I guess, when do you need one vs. the other?

tests/compilation_cache_test.py
skye: No I think it's good! I only had the platform_version in mind initially, but the other things don't hurt, and will be helpful if we ever extend the cache to work on non-TPU backends.

tests/compilation_cache_test.py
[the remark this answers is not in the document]
colemanliyah: No you're right I don't know why I thought that wouldn't work for all the tests...I went ahead and changed that helper function.

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    self.assertEqual(type(cc.get_cache_key(computation, compile_options)), str)
+
+  @unittest.skipIf(jax.lib.version < (0, 1, 68), "fails with earlier jaxlibs")
+  def test_compile_options(self):
+    compile_options_not_filled = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    compile_options_filled = self.filled_compile_options()
+    hash_obj1 = self.get_hashed_value(cc._hash_compile_options, compile_options_filled)
```
skye, line 37: Style int: maybe rename these variables from "hash_obj" (and it's actually a string at this point, not the hash object thing) to things like: `filled_hash1`, `filled_hash2`, etc. (and similarly for below tests)

skye (commented):
@zhangqiaorjc you can start reviewing

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+import unittest
+
+class CompilationCacheTest(jtu.JaxTestCase):
+
+  @unittest.skipIf(jax.lib.version < (0, 1, 68), "fails with earlier jaxlibs")
+  def test_cache_key_type(self):
+    computation = jax.xla_computation(lambda x, y: x + y)(1, 1)
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    self.assertEqual(type(cc.get_cache_key(computation, compile_options)), str)
```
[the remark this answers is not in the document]
colemanliyah: Yeah, this test was put in just as an extra check really but I can take it out if its redundant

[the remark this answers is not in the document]
skye: Agree we should rely on pytype instead of testing here.

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+  @unittest.skipIf(jax.lib.version < (0, 1, 68), "fails with earlier jaxlibs")
+  def test_hash_platform(self):
+    filled_hash1 = self.get_hashed_value(cc._hash_platform, jax.lib.xla_bridge.get_backend())
+    filled_hash2 = self.get_hashed_value(cc._hash_platform, jax.lib.xla_bridge.get_backend())
+    if jax.lib.xla_bridge.get_backend().platform != "cpu":
+        cpu_backend = jax.lib.xla_bridge.get_backend("cpu")
+        filled_hash3 = self.get_hashed_value(cc._hash_platform, cpu_backend)
+        self.assertNotEqual(filled_hash1, filled_hash3)
+    self.assertEqual(filled_hash1, filled_hash2)
```
skye, line 67: nit: move this above the if statement

tests/compilation_cache_test.py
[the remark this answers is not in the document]
zhangqiaorjc: yeah, maybe remove this?

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+    self.assertEqual(type(cc.get_cache_key(computation, compile_options)), str)
+
+  @unittest.skipIf(jax.lib.version < (0, 1, 68), "fails with earlier jaxlibs")
+  def test_compile_options(self):
+    compile_options_not_filled = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    compile_options_filled = self.filled_compile_options()
+    filled_hash1 = self.get_hashed_value(cc._hash_compile_options, compile_options_filled)
+    filled_hash2 = self.get_hashed_value(cc._hash_compile_options, compile_options_filled)
+    filled_hash3 = self.get_hashed_value(cc._hash_compile_options, compile_options_not_filled)
```
skye, line 32: You should named this `not_filled_hash`, to match `compile_options_not_filled`. Please check your other tests too and make sure the hash variable names reflect what they're a hash of.

zhangqiaorjc (approved):

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+    computation1 = jax.xla_computation(lambda x, y: x + y)(1, 1)
+    computation2 = jax.xla_computation(lambda x, y: x * y)(2, 2)
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                       num_replicas=1, num_partitions=1)
+    self.assertNotEqual(cc.get_cache_key(computation1, compile_options),
+                        cc.get_cache_key(computation2, compile_options))
+
+  def filled_compile_options(self):
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                       num_replicas=1, num_partitions=1)
```
skye, line 118: Can you actually manually create a CompileOptions here instead of calling `get_compile_options`? Like this: https://github.com/google/jax/blob/main/jax/lib/xla_bridge.py#L86-L88

This will make it easier to see which fields are set.

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,142 @@
...
+    self.assertNotEqual(filled_hash1, filled_hash3)
+
+  def test_debug_options(self):
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    filled_hash1 = self.get_hashed_value(cc._hash_debug_options,
+                compile_options.executable_build_options.debug_options)
+    filled_hash2 = self.get_hashed_value(cc._hash_debug_options,
+                compile_options.executable_build_options.debug_options)
+    self.assertEqual(filled_hash1, filled_hash2)
```
skye, line 57: Given that you don't actually test different DebugOptions anywhere else, I think you should manually create a different DebugOptions object here and test that it gets a different hash. You can do so by modifying the fields of an existing DebugOptions.

skye (changes requested):

tests/compilation_cache_test.py
```diff
@@ -0,0 +1,153 @@
...
+                   compile_options_filled.executable_build_options)
+    not_filled_hash3 = self.get_hashed_value(cc._hash_executable_build_options,
+                       compile_options_not_filled.executable_build_options)
+    self.assertEqual(filled_hash1, filled_hash2)
+    self.assertNotEqual(filled_hash1, not_filled_hash3)
+
+  def test_debug_options(self):
+    compile_options = jax.lib.xla_bridge.get_compile_options(
+                      num_replicas=1, num_partitions=1)
+    filled_hash1 = self.get_hashed_value(cc._hash_debug_options,
```
skye, line 54: `filled_hash` doesn't really make sense here, since there's no concept of filled vs. not filled for debug options. Please rename the variables in this test + below tests that also don't have a "filled" concept (I think all of the below ones). You can just use `hash1`, `hash2`, etc., or something more descriptive if you feel inspired.

skye (commented):
Almost there!!

skye (approved):
Nice work!

PR #6983, opened 2021-06-15
Add jaxlib version check on newly-added NumpyLinalgTest.testEighZeroDiagonal test.

skye:
This is testing behavior recently added to jaxlib, so fails on older jaxlibs.


hawkinsp (approved):
Thanks!

PR #6993, opened 2021-06-16
Model effects in Jaxpr with an effectful tag.

zhangqiaorjc:


jax/_src/random.py
```diff
@@ -207,6 +207,7 @@ def _broadcast(x):
 
 threefry2x32_p = core.Primitive("threefry2x32")
 threefry2x32_p.multiple_results = True
+threefry2x32_p.effectful = True
```
jekbradbury, line 210: I think `rng_uniform_p` is effectful, but `threefry2x32_p` should be pure?

PR #6996, opened 2021-06-16
[sparse] add sparse grad wrapper

jakevdp:
This is a wrapper for `jax.grad` that correctly handles gradients with respect to a sparse matrix (i.e. does not differentiate the indices).

### DIFF
--- a/jax/experimental/sparse/ad.py
+++ b/jax/experimental/sparse/ad.py
@@ -1,0 +1,101 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Callable, Sequence, Tuple, Union
+
+import numpy as np
+
+import jax
+from jax import core
+from jax import tree_util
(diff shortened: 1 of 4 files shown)


jax/experimental/sparse_ops.py
```diff
@@ -51,11 +51,52 @@
...
 from jax.util import safe_zip
 from jax._src.lax.lax import ranges_like, remaining, _dot_general_batch_dim_nums
+from jax.api_util import _ensure_index, _ensure_index_tuple
 
 xb = xla_bridge
 xops = xla_client.ops
 Dtype = Any
 
+
+def grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0, **kwargs) -> Callable:
```
shoyer, line 61: Maybe worth doing `value_and_grad`, and making `grad` a wrapper of that, as is the case in `api.py`? In my experience, it is pretty common to want to pull out the `loss` when doing gradient based optimization. (and it doesn't add any additional cost)

jakevdp: Good idea - Done!

jax/experimental/sparse/ad.py
```diff
@@ -0,0 +1,97 @@
...
+from jax import core
+from jax import tree_util
+from jax.api_util import _ensure_index, _ensure_index_tuple
+from jax.util import safe_zip
+from jax._src.util import wraps
+from jax._src.traceback_util import api_boundary
+from .ops import BCOO
+
+
+def value_and_grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0, **kwargs) -> Callable[..., Tuple[Any, Any]]:
```
mattjj, line 29: Could you try to wrap to 80 characters? It's not a strict rule but it helps some of our colleagues. This line seems super long. (I have a command in my editor which hard-wraps to 80 chars. Maybe yours can do that too!)

mattjj (approved):
LGTM!

jax/experimental/sparse/ad.py
jakevdp: Done

PR #7008, opened 2021-06-17
Improve concretization errors for jnp indexing routines

nicholasjng:
A small refactor of `jnp.mgrid, jnp.ogrid, jnp.r_, jnp.c_` for handling slice arguments. I refactored the logic for slice creation to apply to all four routines, and proper concretization errors are now raised in `jnp.r_` and `jnp.c_` when trying to pass tracers as slice parameters. This is checked by some added tests.


jakevdp (approved):
Beautiful, thanks!

PR #7012, opened 2021-06-17
random: generate large sequences via vmap rather than Python loops

### LINKED ISSUE #7011: eval_shape appears to hang when tracing a large random sample
Calling

```python
import jax
from jax import random
jax.eval_shape(lambda: random.normal(key=random.PRNGKey(0), shape=[int(1e11)]))
```

returns `ShapeDtypeStruct(shape=(100000000000,), dtype=float32)`. I'd expect that calling

```python
jax.eval_shape(lambda: random.normal(key=random.PRNGKey(0), shape=[int(1e11), int(1e11)]))
```

would similarly return `ShapeDtypeStruct(shape=(100000000000,100000000000), dtype=float32)` (of course, if we *actually* ran the function we'd get an OOM error). 

But instead, this call seems to end up in an ~infinite loop (at least, it hangs for > 2 minutes). Reproduction here:
https://colab.research.google.com/drive/14IZF3WUTaq976jaUC5tehKTPwt3YhNTe?usp=sharing

Is this expected?

jakevdp:
Fixes #7011

Testing that this does not change random results requires generating >4GB of random bits. Instead, I confirmed this is correct by temporarily setting the maximum to a small number to generate multiple small blocks, and compared the output of the old version to the output of the new version.

With the new version, the size of the jaxpr no longer scales with the size of the requested random bits.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -304,14 +304,15 @@ def _random_bits(key, bit_width, shape):
   max_count = int(np.ceil(bit_width * size / 32))
 
   nblocks, rem = divmod(max_count, jnp.iinfo(np.uint32).max)
+
   if not nblocks:
     bits = threefry_2x32(key, lax.iota(np.uint32, rem))
   else:
-    *subkeys, last_key = split(key, nblocks + 1)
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):

hawkinsp:
Nice use of `vmap`!

PR #7034, opened 2021-06-21
Add vmap rule for lax.clamp.

### LINKED ISSUE #5325: vmap failing for lax.clamp
For example:
```
# Works if min/max are scalars and x is an array 
lax.clamp(1., np.arange(4, dtype=np.float32), 3.)
# We vmap it with a batch size of 3
jax.vmap(lax.clamp)(np.arange(3.),
                                   np.arange(12, dtype=np.float32).reshape((3, -1)),
                                   np.arange(3.))
```
gives the error:

```
  File "/Users/necula/Source/jax/jax/experimental/jax2tf/tests/batch_poly_test.py", line 116, in test_batch_clamp
    np.arange(3.))
  File "/Users/necula/Source/jax/jax/_src/lax/lax.py", line 478, in clamp
    return clamp_p.bind(min, x, max)
  File "/Users/necula/Source/jax/jax/_src/lax/lax.py", line 1988, in standard_abstract_eval
    shapes, dtypes = shape_rule(*args, **kwargs), dtype_rule(*args, **kwargs)
  File "/Users/necula/Source/jax/jax/_src/lax/lax.py", line 3331, in _clamp_shape_rule
    raise TypeError(m.format(min.shape))
jax._src.traceback_util.FilteredStackTrace: TypeError: clamp requires min.shape == operand.shape or min.shape == (), got (3, 1).
```

It looks like the clamp needs a more elaborate batching rule.

zhangqiaorjc:


tests/lax_vmap_test.py
```diff
@@ -220,7 +220,6 @@ def testBitcastElementType(self, shape, from_dtype, to_dtype, bdims,):
       for bdims in all_bdims(min_shape, operand_shape, max_shape)))
   def testClamp(self, min_shape, operand_shape, max_shape, dtype, bdims):
     rng = jtu.rand_default(self.rng())
-    raise SkipTest("batching rule for clamp not implemented")  # TODO(mattj)
```
mattjj, line 223: Oh man, high five to my past self for these tests being here :D

mattjj (approved):
Nice!

mattjj:
I wonder if the test failures are just numerical jitter? WDYT?

tests/lax_vmap_test.py
zhangqiaorjc: Yeah, great tests :)

PR #7040, opened 2021-06-21
Fix random test for scipy>=1.17.0

### LINKED ISSUE #7039: CI fails due to scipy 1.7.0
The new scipy release causes test failures: https://github.com/google/jax/runs/2875466140

jakevdp:
Fixes #7039

This is a test-suite only change, and does not affect any JAX user code.

The context here is that we are using a chi-square test in our test suite to confirm that our random samples match the expected distribution.

The issue is that scipy 1.17 added a restriction on its chi-square test inputs, because the chi-square test only produces correct results if the expected and actual counts are equal.  Previous to scipy 1.17, this criterion was not checked, but now it is (see https://github.com/scipy/scipy/issues/13362). Since our test suite currently uses the chi-square test incorrectly, it is flagged by this new check.

I initially considered artificially normalizing the expected frequencies to side-step this requirement, but this is a flawed approach: the reason the JAX test suite does not meet this criterion is because we were only considering the expected counts **for observed values**, instead we need to consider expected counts across all possible values. The failure mode is if (for example) we observe zero counts for a value where the distribution predicts 100: by only considering observed values, the test would miss this divergence from expectation.

Rather, the solution I landed on is to consider expected counts for a wider range of values. This makes the test more correct, and as a side-effect, passes scipy's new input validation.

### DIFF
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -61,20 +61,30 @@ class LaxRandomTest(jtu.JaxTestCase):
 
   def _CheckChiSquared(self, samples, pmf):
     alpha = 0.01  # significance level, threshold for p-value
-    values, actual_freq = np.unique(samples, return_counts=True)
+
+    # scipy.stats.chisquare requires the sum of expected and actual to
+    # match; this is only the case if we compute the expected frequency
+    # at *all* nonzero values of the pmf. We don't know this a priori,
+    # so we add extra values past the largest observed value. The number
+    # below is empirically enough to get full coverage for the current set
+    # of tests. If a new test is added where this is not enough, chisquare()
+    # below will error due to the sums of the inputs not matching.
+    extra_values = 100
+    actual_freq = np.bincount(samples, minlength=samples.max() + extra_values)
+    values = np.arange(len(actual_freq))
+
     expected_freq = pmf(values) * samples.size
-    # per scipy: "A typical rule is that all of the observed and expected
-    # frequencies should be at least 5."
-    valid = (actual_freq > 5) & (expected_freq > 5)
-    self.assertGreater(valid.sum(), 1,
-                       msg='not enough valid frequencies for chi-squared test')
-    _, p_value = scipy.stats.chisquare(
-        actual_freq[valid], expected_freq[valid])
+
+    valid = expected_freq > 0
+    actual_freq = actual_freq[valid]
+    expected_freq = expected_freq[valid]
+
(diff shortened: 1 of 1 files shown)


tests/random_test.py
```diff
@@ -539,10 +545,10 @@ def testPoisson(self, lam, dtype):
 
   def testPoissonBatched(self):
     key = random.PRNGKey(0)
-    lam = jnp.concatenate([2 * jnp.ones(10000), 20 * jnp.ones(10000)])
+    lam = jnp.concatenate([2 * jnp.ones(10000), 11 * jnp.ones(10000)])
```
srvasude, line 548: Any reason these are changed to 11? I would imagine the test should work with 20

srvasude (approved):

tests/random_test.py
```diff
@@ -61,20 +61,24 @@ def _CheckKolmogorovSmirnovCDF(self, samples, cdf):
 
   def _CheckChiSquared(self, samples, pmf):
     alpha = 0.01  # significance level, threshold for p-value
-    values, actual_freq = np.unique(samples, return_counts=True)
+
+    # chi-square test requires sums to be the same, so we over-sample
+    # the range to pick up the full expected distribution.
+    actual_freq = np.bincount(samples, minlength=samples.max() + 100)
```
mattjj, line 67: Maybe brief comment on the magic 100 number (we talked in 1:1 and it sounds reasonable).

mattjj (approved):
LGTM!

tests/random_test.py
jakevdp: It fails with the combination of PRNG Key 0 and lam=20 because an unlikely value happens to be produced. It wasn't caught previously because we removed all values with counts less than 5.

tests/random_test.py
jakevdp: Done

tests/random_test.py
jakevdp: The docs seem to suggest that this can produce false negatives when counts are less than 5, but I don't know how to avoid that in long-tailed distributions while still maintaining the requirement that the expected and actual distribution add up to the same number.

mattjj: Was it that a rare value was produced, or just that the pmf has too much nonzero mass beyond ~100 when the parameter is 20?

jakevdp: I don’t think it’s from the pig having power beyond the cutoff, because that would lead to a RuntimeError form scipy, where the failure here was a p value that was out of range.

PR #7045, opened 2021-06-21
Catch CalledProcessError in build script.

zhangqiaorjc:
print(e.output) should tell us the build errors for https://github.com/google/jax/issues/6637

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -48,7 +48,11 @@ def is_windows():
 
 
 def shell(cmd):
-  output = subprocess.check_output(cmd)
+  try:
+    output = subprocess.check_output(cmd)
+  except subprocess.CalledProcessError as e:
+    print(e.output)
+    raise
   return output.decode("UTF-8").strip()
 
 


build/build.py
```diff
@@ -48,7 +48,10 @@ def is_windows():
 
 
 def shell(cmd):
-  output = subprocess.check_output(cmd)
+  try:
+    output = subprocess.check_output(cmd)
+  except subprocess.CalledProcessError as e:
+    print(e.output)
```
hawkinsp, line 54: I think we should reraise the exception here?

build/build.py
```diff
@@ -206,7 +209,8 @@ def get_bazel_path(bazel_path_flag):
 def check_bazel_version(bazel_path):
   try:
     version_output = shell([bazel_path, "--bazelrc=/dev/null", "version"])
-  except subprocess.CalledProcessError:
+  except subprocess.CalledProcessError as e:
```
hawkinsp, line 212: Isn't this one redudant if you rethrow the exception above?

build/build.py
zhangqiaorjc: Done

build/build.py
zhangqiaorjc: Reverted

build/build.py
hawkinsp: I think to reraise you can just write `raise` with no `e`.

build/build.py
```diff
@@ -48,7 +48,11 @@ def is_windows():
 
 
 def shell(cmd):
-  output = subprocess.check_output(cmd)
+  try:
+    output = subprocess.check_output(cmd)
+  except subprocess.CalledProcessError as e:
+    print(e.output)
```
hawkinsp, line 54: Shouldn't this be `output` not `e.output`?

Verify this works on linux?

hawkinsp: Actually I guess this is fine as is.

build/build.py
zhangqiaorjc: done

hawkinsp (approved):

Issue #7052, opened 2021-06-22
Current JAX release incompatible with numpy 1.21


jakevdp:
Numpy 1.21 was released a few hours ago, and appears incompatible with the current jaxlib release.

shoyer:
https://github.com/numpy/numpy/issues/19305 is the source of one of the remaining errors.

hawkinsp:
I've done all I plan to do on this for now: remaining is to release a new jaxlib (in progress) and to do something to fix or work around the numpy bug.

jakevdp:
Alright, I'll self-assign so I can close the issue after the next jaxlib release.

hawkinsp:
Note that we can't close it until we do something to work around the numpy problem. It could just be skipping those tests under NumPy 1.21 for now.

If you want to try running the tests before the new jaxlib lands, apply the following patch:
```
diff --git a/jax/__init__.py b/jax/__init__.py
index 17629466..caca87a4 100644
--- a/jax/__init__.py
+++ b/jax/__init__.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.

+import numpy as np
+np.typeDict = np.sctypeDict
+
 # Set default logging level before any logging happens.
 import os as _os
 _os.environ.setdefault('TF_CPP_MIN_LOG_LEVEL', '1')
```

jakevdp:
Part of the action item here may be to add a CI job to detect these kinds of breakages before release. @shoyer pointed out the approach that xarray uses, which is a github action that runs tests against upstream dev versions and opens a github issue in case of failure: https://github.com/pydata/xarray/blob/main/.github/workflows/upstream-dev-ci.yaml

PR #7055, opened 2021-06-22
Fix most test failures under NumPy 1.21.

hawkinsp:
NumPy is stricter about scalar -> bfloat16 promotions in 1.21, which we can work around by casting explicitly.

NumPy is also stricter about distinguishing between dtype objects and scalar objects for .dtype attributes.

np.unwrap takes a new `period` argument which is not correctly promoted for a bfloat16 input.

Also relax a tree_util test that was failing for me.

https://github.com/google/jax/issues/7052

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -302,7 +302,13 @@ JAX_COMPOUND_OP_RECORDS = [
     op_record("true_divide", 2, all_dtypes, all_shapes, jtu.rand_nonzero,
               ["rev"], inexact=True),
     op_record("ediff1d", 3, [np.int32], all_shapes, jtu.rand_default, []),
-    op_record("unwrap", 1, float_dtypes, nonempty_nonscalar_array_shapes,
+    # TODO(phawkins): np.unwrap does not correctly promote its default period
+    # argument under NumPy 1.21 for bfloat16 inputs. It works fine if we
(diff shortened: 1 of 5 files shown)


tests/lax_numpy_test.py
```diff
@@ -302,7 +302,8 @@ def op_record(name, nargs, dtypes, shapes, rng_factory, diff_modes,
     op_record("true_divide", 2, all_dtypes, all_shapes, jtu.rand_nonzero,
               ["rev"], inexact=True),
     op_record("ediff1d", 3, [np.int32], all_shapes, jtu.rand_default, []),
-    op_record("unwrap", 1, float_dtypes, nonempty_nonscalar_array_shapes,
+    op_record("unwrap", 1, [t for t in float_dtypes if t != dtypes.bfloat16],
+              nonempty_nonscalar_array_shapes,
```
jakevdp, line 306: Should we leave a TODO here, or are we fine no longer supporting `bfloat16` in `jnp.unwrap`?

hawkinsp: Added a TODO. `np.unwrap` itself actually works, it just cannot promote 2*pi to `bfloat16` since scalar -> bfloat16 float promotions work in fewer places in numpy 1.21.

jakevdp (approved):

PR #7062, opened 2021-06-23
Add `pip install jax[tpu]` configuration.

skye:
This can be used on Cloud TPU VMs to automatically install compatible
versions of jax, jaxlib, and libtpu (the low-level library JAX uses to
access the TPU on Cloud TPU VMs).

The new install command requires a new jax release (`>=0.2.15`) and
jaxlib release (`>=0.1.68`) to work, since it requires both
https://github.com/google/jax/commit/cdfbd9dde17dd97f62b4bae3d972ee001c119e18
and
https://github.com/tensorflow/tensorflow/commit/ce2bc249968344c57e91c45799ad5ec91aad2566
to pick up the pip-installed libtpu. I'll update the README and Cloud
TPU VM documentation once these releases are out.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,11 @@ with open('jax/version.py') as f:
 __version__ = _dct['__version__']
 _minimum_jaxlib_version = _dct['_minimum_jaxlib_version']
 
+_libtpu_version = '20210615'
+_libtpu_url = (
+    f'https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/'
+    f'libtpu-nightly/libtpu_nightly-0.1.dev{_libtpu_version}-py3-none-any.whl')
+
 setup(
     name='jax',
     version=__version__,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,6 +51,11 @@
         # $ pip install jax[cpu]
         'cpu': [f'jaxlib>={_minimum_jaxlib_version}'],
 
+        # Cloud TPU VM jaxlib can be installed via:
+        # $ pip install jax[tpu]
+        'tpu': [f'jaxlib=={_current_jaxlib_version}',
+                f'libtpu-nightly @ {_libtpu_url}'],
```
jakevdp, line 57: I haven't seen this `<libary> @ <url>` syntax before. Could we use the same thing to avoid requiring the -f addition for GPU installs?

skye: @yashk2810 pointed me to it! https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py/54216163#54216163

And yes, I think we could use it for the GPU builds, great idea! It would require hardcoding the wheel URL, but that doesn't seem bad given we already hardcode the jaxlib version. Let's do it in a follow-up change though, as I'd like to get this in ASAP to avert impending jaxlib/libtpu incompatibility.

jakevdp (approved):

PR #7064, opened 2021-06-23
[jax2tf] Fix conversion of gradients for shape polymorphic functions.

gnecula:
This fixes the case when the primal shape polymorphic function has
output shapes that are polynomials of the input shapes (not just
dimension variables).


gnecula:
Copybara configuration error. Closing and retrying.

PR #7073, opened 2021-06-23
Implementing thread safety check

colemanliyah:


jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -41,8 +44,14 @@ def put(self, key: str, value: bytes):
       raise ValueError("key cannot be empty")
     if self._evict_entries_if_necessary(key, value):
       path_to_new_file = os.path.join(self._path, key)
-      with open(path_to_new_file, "wb") as file:
-        file.write(value)
+      with tempfile.TemporaryDirectory() as tmpdir:
```
skye, line 47: Good idea to use a temporary directory! You don't need to include the process and thread ID in the filename with this, since different threads/processes will get different temp directories.

skye: Also, leave a comment explaining the rationale for using the temporary directory + move.

tests/file_system_cache_test.py
```diff
@@ -122,5 +123,19 @@ def test_file_bigger_than_cache(self):
...
       self.assertEqual(cache.get("foo"), None)
 
+  def test_threads(self):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir)
+      for i in range(100):
+          threading.Thread(target=self.call_multiple_puts_and_gets(cache))
+          self.assertEqual(cache.get("foo"), b"bar")
+
+  def call_multiple_puts_and_gets(self, cache):
```
skye, line 133: Style nit: since this is a one-off function needed for the threads in the above test, you can actually put this function definition inside the `test_threads` function. I put it as the first thing in `test_threads`, like before creating the tmpdir.

skye (changes requested):
Nice work!

jax/experimental/compilation_cache/file_system_cache.py
skye: I think you still need to add the comment?

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -43,6 +43,9 @@ def put(self, key: str, value: bytes):
       raise ValueError("key cannot be empty")
     if self._evict_entries_if_necessary(key, value):
       path_to_new_file = os.path.join(self._path, key)
+      #Creating the path for the file in a temporary directory so we can use the
+      #atomic move function to ensure that the file and its contents are properly
+      #stored in the cache
```
skye, line 48: A few nits:
* Put a space between `#` and rest of comment
* Say "Create the path" instead of "Creating the path"
* You should mention something about thread safety, since the tmpdir + move is only needed in the case of concurrent `put`s and `get`s across multiple threads or processes. Maybe: "Create the path for the file in a temporary directory so we can use the atomic move function to ensure that the file is properly stored and read in the case of concurrent access across multiple threads or processes" (it's a mouthful, but hopefully gets the point across)

skye (approved):

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -41,8 +43,14 @@ def put(self, key: str, value: bytes):
...
       path_to_new_file = os.path.join(self._path, key)
-      with open(path_to_new_file, "wb") as file:
-        file.write(value)
+      # Create the path for the file in a temporary directory so we can use the
+      # atomic move function to ensure that the file is properl stored and read
+      # in the case of concurrent access across multiple threads or processes
+      with tempfile.TemporaryDirectory() as tmpdir:
+          temp_path_to_file = os.path.join(tmpdir, key)
+          with open(temp_path_to_file, "wb") as file:
+              file.write(value)
```
zhangqiaorjc, line 52: To ensure data is actually written to disk,

```
file.write(value)
file.flush() # flush from app buffer to os buffer
os.fsync(file.fileno())  # flush from os buffer to disk
```
For explanations of why flush and fsync is needed, see
https://stackoverflow.com/questions/7127075/what-exactly-is-pythons-file-flush-doing

For background on the perils of filesystem async writes, see
- Slide 70 onwards
https://www.slideshare.net/nan1nan1/eat-my-data 

Shorter version
- https://stackoverflow.com/questions/2333872/how-to-make-file-creation-an-atomic-operation

jax/experimental/compilation_cache/file_system_cache.py
```diff
@@ -41,8 +43,14 @@ def put(self, key: str, value: bytes):
...
+          shutil.move(temp_path_to_file, path_to_new_file)
```
zhangqiaorjc, line 53: hmm, maybe we should use `os.rename` instead of `shutil.move` here?

- POSIX requires rename in the same file system to be atomic
https://docs.python.org/3/library/os.html#os.rename

- shutil.move seems to be more general and work on different file system, but i don't think that kind of file rename is atomic anymore and therefore defeats the purpose of writing to a temp file first and rename. In fact, I don't think we want to support renaming to a file in a different filesystem?

tests/file_system_cache_test.py
```diff
@@ -122,5 +123,19 @@ def test_file_bigger_than_cache(self):
...
+    def call_multiple_puts_and_gets(cache):
+      for i in range(100):
+        cache.put("foo", b"bar")
+        cache.put("foo", b"bar")
+        cache.get("foo")
+        self.assertEqual(cache.get("foo"), b"bar")
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+      cache = FileSystemCache(tmpdir)
+      for i in range(100):
```
zhangqiaorjc, line 136: I think you want to wait for the threads to finish doing stuff first...
```
threads = []
for i in range(100):
   t = threading.Thread(target=call_multiple_puts_and_gets(cache))
   t.start()
   threads.append(t)
for t in threads:
  t.join()

self.assertEqual(cache.get("foo"), b"bar")
```

tests/file_system_cache_test.py
```diff
@@ -122,5 +123,19 @@ def test_file_bigger_than_cache(self):
       cache.put("foo", b"bar")
       self.assertEqual(cache.get("foo"), None)
 
+  def test_threads(self):
+    def call_multiple_puts_and_gets(cache):
+      for i in range(100):
+        cache.put("foo", b"bar")
```
zhangqiaorjc, line 129: "bar" is only 3 bytes, and it's most likely too small to be not atomic...so we won't be testing the limits of cache write.. we may need a cache write bigger than 64KB or PIPE_MAX size, see

https://stackoverflow.com/questions/7236475/what-happens-if-a-write-system-call-is-called-on-same-file-by-2-different-proces

also, since all the threads write "bar", it's hard to get corruption too...Maybe we should have a set of larger-than-64KB values that each thread can choose from to write, and then at the end of test, just hash contents from each cache read and make sure that the md5sum is one of the expected values

zhangqiaorjc (changes requested):

jax/experimental/compilation_cache/file_system_cache.py
colemanliyah: Could you expand on what you mean by last question in your second bullet point, I wasn't sure if this was something I needed to address or something that was referring to the shutil.move method

zhangqiaorjc (approved):
Great job!

Issue #7081, opened 2021-06-23
Can't import jax; module 'jaxlib.cusparse' has no attribute 'gtsv2'


lukepfister:
After updating to 0.2.15 and 0.1.68, I'm unable to import jax.
Using CUDA 11.0, Python 3.8, and jax versions:
```
jax                       0.2.15                   pypi_0    pypi
jaxlib                    0.1.68+cuda110           pypi_0    pypi
```

Here's the full traceback:
```python

In [1]: import jax
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-cb15c4215ef7> in <module>
----> 1 import jax

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/__init__.py in <module>
    106   xla_computation,
    107 )
--> 108 from .experimental.maps import soft_pmap
    109 from .version import __version__
    110

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/experimental/maps.py in <module>
     23 from functools import wraps, partial, partialmethod
     24
---> 25 from .. import numpy as jnp
     26 from .. import core
     27 from .. import linear_util as lu

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/numpy/__init__.py in <module>
     14
     15 # flake8: noqa: F401
---> 16 from . import fft
     17 from . import linalg
     18

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/numpy/fft.py in <module>
     15 # flake8: noqa: F401
     16
---> 17 from jax._src.numpy.fft import (
     18   ifft,
     19   ifft2,

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/_src/numpy/fft.py in <module>
     17 import numpy as np
     18
---> 19 from jax import lax
     20 from jax.lib import xla_client
     21 from jax._src.util import safe_zip

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/lax/__init__.py in <module>
    356 )
    357 from jax._src.ad_util import stop_gradient_p
--> 358 from . import linalg

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/lax/linalg.py in <module>
     14
     15 # flake8: noqa: F401
---> 16 from jax._src.lax.linalg import (
     17   cholesky,
     18   cholesky_p,

~/miniconda3/envs/scico/lib/python3.8/site-packages/jax/_src/lax/linalg.py in <module>
[...]

PR #7082, opened 2021-06-23
Test for gtsv2 attr on cusparse.

zhangqiaorjc:


hawkinsp (approved):

zhangqiaorjc:
This PR will fix the issue in https://github.com/google/jax/issues/7081

We will still need to release a new JAX after that.

PR #7091, opened 2021-06-24
Implement jnp.unique with axis and size set.

### LINKED ISSUE #7090: Support size and axis argument for `jnp.unique`
Please:

Hi,
at the moment it is not possible to have `jnp.unique` functioning in jit with `axis` being something else than `None`. While the `size` argument is supposed to make `jnp.unique` jit compatible, it doesn't work with also the `axis` argument being set.

A simple use case for this would be to find the unique vectors in a set.

n-gao:
Fixes #7090

Part of #3614

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4923,7 +4923,9 @@ def _unique_axis(ar, axis, return_index=False, return_inverse=False,
     new_shape = (mask.sum() or aux.shape[1], *out_shape)
   else:
     ind = nonzero(mask, size=size)[0]
-    new_shape = (size, *out_shape)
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks for taking a look at this! Once the implementation is finished, we'll also need to add tests for this new behavior; you can probably model them off of the existing tests (including the one added in #6912)

Please let me know if you run into any difficulties

jakevdp:
Hi - I'd love to get this feature into JAX. Please let me know if there's any way I can help!

jakevdp:
Hi - I'm going to close this PR for now, since it looks like work on it has stalled. Feel free to re-open if you want to continue work on this feature!

PR #7100, opened 2021-06-24
remat: don't apply cse-foiling widget to primal

### LINKED ISSUE #6098: remat_call results in unnecessary CSE foiling when used without partial_eval.
With remat, we want the forward pass to be lowered unmodified, and for the backwards to include a forwards + backwards. However, XLA does common subexpression elimination (CSE), which threatens this - so the `remat_call` primitive lowers with a CSE foiling gadget.

However, foiling CSE has a cost, e.g. missed fusions and other optimizations, so we don't want to do it unless necessary. So, at the moment, the partial-eval-of-remat machinery removes the `remat_call` from the forward pass jaxpr, but keeps it in the backwards pass.

Lost in this logic is that sometimes we just... don't partial eval the `remat_call`, so it sticks around into the final jaxpr and is lowered with CSE-foiling even though it's not needed.

```python
import functools
import jax
import jax.numpy as jnp

@jax.remat
def f(x):
  return jnp.sin(x)

print('Without partial_eval.')
print(jax.make_jaxpr(f)(jnp.zeros(4)))
print('\nWith partial_eval.')
print(jax.make_jaxpr(functools.partial(jax.vjp, f))(jnp.zeros(4)))
```

```
Without partial_eval.
{ lambda  ; a.
  let b = remat_call[ call_jaxpr={ lambda  ; a.
                                   let b = sin a
                                   in (b,) }
                      concrete=False
                      name=f ] a
  in (b,) }
With partial_eval.
{ lambda  ; a.
  let b = sin a
  in (b, a) }
```

The conceptual error here is that one primitive is taking care of two concepts at once:
- Indicating that something is meant to be rematerialized.
- Indicating that we should foil XLA CSE around a certain block. 

One solution: separate this into two calls, one is `remat_call` and the other is `atomic_call`. `remat_call` is inert when lowering; `atomic_call` isn't. partial-evaling `remat_call` produces a n`atomic_call` in the backwards pass.

mattjj:
fixes #6098 (finally)

### The problem

Consider this `jax.remat` (aka `jax.checkpoint`) code

```python
import jax

@jax.remat
def f(x):
  return x ** 2
```

Evaluating `f(3.)`, whether inside a `jax.jit` or not, is made more expensive than just evaluating `x ** 2` by the fact that `jax.remat` always adds a `while`- or `conditional`-based "widget" during XLA lowering to foil CSE optimizations:

```python
print(jax.xla_computation(f)(3.).as_hlo_text())
```

```
HloModule xla_computation_f.23

remat_call_subcomputation.9 {
  constant.12 = pred[] constant(false)
  parameter.10 = (f32[]) parameter(0)
  get-tuple-element.11 = f32[] get-tuple-element(parameter.10), index=0
  multiply.13 = f32[] multiply(get-tuple-element.11, get-tuple-element.11)
  ROOT tuple.14 = (f32[]) tuple(multiply.13)
}

remat_call_dummy_subcomputation.15 {
  parameter.16 = (f32[]) parameter(0)
  constant.17 = f32[] constant(0)
  broadcast.18 = f32[] broadcast(constant.17), dimensions={}
  ROOT tuple.19 = (f32[]) tuple(broadcast.18)
}

ENTRY xla_computation_f.23 {
  constant.3 = f32[] constant(0)
  constant.4 = f32[] constant(1)
  rng.5 = f32[]invalid{} rng(constant.3, constant.4), distribution=rng_uniform
  constant.6 = f32[] constant(2)
  compare.7 = pred[]invalid{} compare(rng.5, constant.6), direction=LT
  parameter.1 = f32[] parameter(0)
  tuple.8 = (f32[]) tuple(parameter.1)
  conditional.20 = (f32[]) conditional(compare.7, tuple.8, tuple.8), true_computation=remat_call_subcomputation.9, false_computation=remat_call_dummy_subcomputation.15
  get-tuple-element.21 = f32[] get-tuple-element(conditional.20), index=0
  ROOT tuple.22 = (f32[]) tuple(get-tuple-element.21)
}
```

This cse-foiling widget exists because when we apply reverse-mode AD (or `jax.linearize`) to a `jax.remat`-decorated function, we _want_ some primal values to be recomputed during the backward pass of the computation, and we don't want XLA to perform CSE, which would have the effect of removing the recomputation.

But notice that this widget appears even though no autodiff is happening here! That is, `jax.remat` is incurring a cost even for primal (aka "inference") computations. The cost is made worse because this widget blocks more XLA optimizations than just CSE, including fusion. And when AD isn't involved, the cost is totally unnecessary!

### The solution

The fix here is to add two boolean parameters, "`prevent_cse`" and "`differentiated`", to the underlying `remat_call` primitive which together indicate whether the widget should be generated when lowering a `remat_call` to XLA HLO. The  "`differentiated`" parameter is managed internally: it starts out as False, and is set True when partial evaluation unzips one `remat_call` into two. The "`prevent_cse`" parameter is an API option which allows expert users to disable the CSE-prevention widget entirely (since in some cases, like in a `scan`, it may not be necessary).

This solution makes sense because `remat_call` is all about special partial evaluation behavior. That is, both for `jax.vjp` and `jax.linearize`, the intended behavior for `jax.remat` is to avoid saving any residuals (i.e. intermediates) in the decorated function during partial evaluation, and instead to recompute those values in the backward pass or linearized pass, respectively. Thus when [`remat_call`'s partial eval rule](https://github.com/google/jax/blob/df3cc0d9801cce71a8ce66306c816b1a7813b12b/jax/interpreters/partial_eval.py#L754-L824) unzips one `remat_call` into two, separating the primal and tangent parts of the computation, it copies the nonlinear operations into the tangent part of the computation. (For reverse-mode, its transpose rule handles evaluating these nonlinear constants.) The widget only needs to apply to the copy of the nonlinear operations; those are the ones for which we need to prevent CSE. We don't need to prevent CSE for the original not-generated-by-AD computation!

This solution makes sense for higher-order AD as well: if we differentiate a `remat_call` application which itself was generated by an application of AD, so the boolean parameter is already set to `differentiated=True`, then both the primal (which is itself part of the derivative for the first-order differentiation) and tangent sides of the unzipped computation should inherit `differentiated=True` values.

### The tests

The tests just check the HLO generated by applying a function like `f` above.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2408,7 +2408,8 @@ def eval_shape(fun: Callable, *args, **kwargs):
   return tree_unflatten(out_tree(), out)
 
 
-def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
+def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
+               ) -> Callable:
   """Make ``fun`` recompute internal linearization points when differentiated.
 
   The :func:`jax.checkpoint` decorator, aliased to ``jax.remat``, provides a
@@ -2446,6 +2447,14 @@ def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
       control flow is optional, and disabled by default, because in some
       edge-case compositions with :func:`jax.jit` it can lead to some extra
       computation.
+    prevent_cse: Optional, boolean indicating whether to prevent common
+      subexpression elimination (CSE) optimizations in the HLO generated from
+      differentiation. This CSE prevention has costs because it can foil other
+      optimizations, and because it can incur high overheads on some backends,
+      especially GPU. The default is True because otherwise, under a ``jit`` or
+      ``pmap``, CSE can defeat the purpose of this decorator. But in some
+      settings, like when used inside a ``scan``, this CSE prevention mechanism
+      is unnecessary, in which case ``prevent_cse`` can be set to False.
 
   Returns:
     A function (callable) with the same input/output behavior as ``fun`` but
@@ -2496,7 +2505,8 @@ def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
     args_flat, in_tree = tree_flatten((args, kwargs))
     flat_fun, out_tree = flatten_fun(lu.wrap_init(fun), in_tree)
     out_flat = pe.remat_call(flat_fun, *args_flat, name=flat_fun.__name__,
-                             concrete=concrete)
+                             concrete=concrete, prevent_cse=prevent_cse,
+                             differentiated=False)
     return tree_unflatten(out_tree(), out_flat)
   return fun_remat
 remat = checkpoint
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -814,12 +814,12 @@ def _remat_partial_eval(trace, _, f, tracers, params):
 
   # dce jaxpr outputs
   new_jaxpr = _dce_jaxpr(closed_jaxpr, out_unknowns, drop_outputs=True).jaxpr
-  new_params = dict(params, call_jaxpr=new_jaxpr)
+  new_params = dict(params, call_jaxpr=new_jaxpr, differentiated=True)
 
   # set up eqn for unknown outputs
   in_tracers = (*const_tracers, *env_tracers, *instantiated_tracers)
-  eqn = new_eqn_recipe(in_tracers, unknown_output_tracers, remat_call_p, new_params,
-                       source_info_util.current())
+  eqn = new_eqn_recipe(in_tracers, unknown_output_tracers, remat_call_p,
+                       new_params, source_info_util.current())
   for t in unknown_output_tracers: t.recipe = eqn
   return _zip_knowns(known_output_tracers, unknown_output_tracers, out_unknowns)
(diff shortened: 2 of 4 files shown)


froystig (approved):

jax/_src/api.py
```diff
@@ -2407,7 +2407,8 @@ def __init__(self, shape, dtype):
   return tree_unflatten(out_tree(), out)
 
 
-def checkpoint(fun: Callable, concrete: bool = False) -> Callable:
+def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = False,
```
gnecula, line 2410: Do we need to change the API here? How JAX tells XLA to remat should not be a concern for the user. 
If we do need this flag now, perhaps we can name it or document it explicitly as experimental.

mattjj: Ah, you're right that this could be internal only. I forget why I added it to the API; it may be that @jekbradbury suggested adding it from experience on some of his work. But I'll change this so that it's internal-only until we need it!

jekbradbury: I wanted this so that I could set it to False when using remat inside scan (which already prevents CSE between the forward and backward passes).

mattjj: Ah, then I've done this wrong. I think I know how to fix it.

PR #7102, opened 2021-06-25
Ensure zeros from AD are generated on device.

### LINKED ISSUE #7093: Array constants are generated on the host during AD
I noticed that the function `abstract_arrays.zeros_like_shaped_array` has code like `np.broadcast_to(np.array(0, aval.dtype), aval.shape)` (using regular NumPy). These values will be computed on the host during tracing and then may be passed to the device. Is this the desired behavior? 

For example, when we take gradients of constant functions:
```
x = np.ones((10, 10), dtype=np.float32)
jax.make_jaxpr(jax.grad(lambda x: 42.))(x)
```

we get
```
{ lambda a ; b.
  let 
  in (a,) }
```

where a 10x10 array of 0s is built during tracing and passed to the function as the array `a`.
Wouldn't it be better to compute the array of 0s directly on the device?

```
{ lambda  ; a.
   let  b = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(10, 10) ] 0.0
  in (b,) }
```

Incidentally, this affects things like jax2tf where we may end up with large constants embedded in the resulting graph, when instead their computation can be embedded in the graph.

gnecula:
Fixes: #7093

### DIFF
--- a/jax/_src/abstract_arrays.py
+++ b/jax/_src/abstract_arrays.py
@@ -38,7 +38,8 @@ def make_shaped_array(x):
 
 def zeros_like_array(x):
   dtype = dtypes.canonicalize_dtype(dtypes.result_type(x))
-  return zeros_like_shaped_array(ShapedArray(np.shape(x), dtype))
+  aval = ShapedArray(np.shape(x), dtype)
+  return ad_util.zeros_like_aval(aval)
 
 array_types = {np.ndarray, np.bool_,
                np.int8, np.int16, np.int32, np.int64,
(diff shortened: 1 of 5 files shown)


gnecula:
@hawkinsp I wonder if it is easy for you to spot what the problem is here. It fails on Python 3.6.13, but it works internally. 

The error is when calling `dtype(val)` where `val` is an array of `float0`.

https://github.com/google/jax/pull/7102/checks?check_run_id=2911939065

mattjj (approved):
Awesome, thanks!

Issue #6805, opened 2021-05-21
Disable warning for No GPU/TPU found


altosaar:
Minor idea - would be nice to be able to disable this warning for debugging ease:

`WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)`

The only place I found a reference to this was in this colab: https://colab.research.google.com/github/google/jax/blob/master/docs/notebooks/quickstart.ipynb#scrollTo=ehUS7s8xKkxK

```
# Execute this to consume & hide the GPU warning.
jnp.arange(10)
```

Which doesn't work outside of a notebook.

I tried using `export TF_CPP_MIN_LOG_LEVEL=3`, `import os; os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'` - any other ideas how to suppress? 

Thanks so much!

mattjj:
If you intend to use the CPU, it is a bit annoying!

Here are two ways to suppress it:
1. set the shell environment variable `JAX_PLATFORM_NAME=cpu`
2. near the top of your main file, write `import jax; jax.config.update('jax_platform_name', 'cpu')`

Either of those things will tell JAX that you intend to use the CPU, so it won't warn about only having a CPU available.

altosaar:
Awesome thanks so much Matt! That worked :)

wookayin:
> 1. set the shell environment variable `JAX_PLATFORM_NAME=cpu`
> 2. near the top of your main file, write `import jax; jax.config.update('jax_platform_name', 'cpu')`
> 
> Either of those things will tell JAX that you intend to use the CPU, so it won't warn about only having a CPU available.

The name of config has changed from `jax_platform_name` to `jax_platforms`, so this will not work as of now (jax 0.3+)

The new way:

1. set the shell environment variable `JAX_PLATFORMS=cpu`
2. near the top of your main file, write `import jax; jax.config.update('jax_platforms', 'cpu')`

But many thanks @mattjj for the workaround!

hawkinsp:
We deprecated `jax_platform_name` in place of `jax_platforms`. So I recommend using the new name.

mattjj:
Indeed! I'm just clarifying that it does still work as of jax 0.3+. (Also it doesn't have a deprecation message...)

wookayin:
[...]

PR #7105, opened 2021-06-25
Allow suppression of GPU warning via jax_platform_name

jakevdp:
Related to #6805

At some point, suppressing the GPU/TPU warning via `jax.config.update('jax_platform_name', 'cpu')` stopped working.

This makes it work again, and adds a regression test.


mattjj (approved):
Thanks for catching this!

PR #7114, opened 2021-06-26
Adding np.s_ functionality and associated test

charmerDark:
Adds the np.s_ function requested in issue #5850 . I've added a basic test as well. Please do look let me know if there are any modifications to be made


jax/_src/numpy/lax_numpy.py
```diff
@@ -3617,6 +3617,18 @@ class CClass(_AxisConcat):
...
+
+  Examples
+    >>> jnp.s_[1:2:10]
+    slice(1, 2, 10)
+  """
+  def __getitem__(self,slc):
+    return slc
+
+s_ = S_class()
+
```
jakevdp, line 3631: In this case, since the return value does not contain any arrays, it may be sufficient to write `s_ = np.s_` rather than re-implementing it. What do you think?

charmerDark: That does sound easier. The issue discouraged using numpy code as it is, so I just implemented it all over again. I was wondering whether I missed I something because this function seemed rather easy but was still the last one left

jakevdp: Sorry for the lack of clarity there: there is no problem with using numpy functionality via import in the JAX source; the problem is with copying numpy source code and pasting it into the JAX source code.

charmerDark: Yep..it would be cleaner to just use the numpy implementation then!

jax/_src/numpy/lax_numpy.py
```diff
@@ -3616,6 +3616,19 @@ class CClass(_AxisConcat):
...
+  """Handy tool that returns `slice` objects from standard array indexing syntax.
+
+  Examples
+    >>> jnp.s_[1:2:10]
+    slice(1, 2, 10)
+  """
+  def __getitem__(self,slc):
+    return slc
+
+s_ = S_class()
```
jakevdp, line 3631: I think we can delete `S_class` if we're just aliasing `jnp.s_` to `np.s_`.

Also, we may as well define `index_exp = np.index_exp` at the same time.

tests/lax_numpy_test.py
```diff
@@ -4945,6 +4945,9 @@ def testC_(self):
                         jnp.c_[1.2:4.8:0.24],
                         atol=atol,
                         rtol=rtol)
```
jakevdp, line 4947: Nit: add a blank line between test methods

charmerDark: Do we really need the test if we are just using the numpy implementation? Maybe I should just remove them? And thank you so much for the support

jakevdp: I think a simple test like this is worthwhile just to confirm the object is properly exposed in the `jnp` namespace, which can come in handy in case of future refectors, etc.

Can you add `index_exp` to the same test?

jakevdp:
Lastly, before we can merge we'll need all the changes squashed into a single commit (See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Thanks!

jakevdp (approved):

jakevdp:
Thanks! We'll get this merged 😁

tests/lax_numpy_test.py
```diff
@@ -4945,6 +4945,12 @@ def testC_(self):
...
+
```
jakevdp, line 4948: Linter is complaining that this line has stray whitespace

jakevdp:
Thanks - can you please squash the changes into a single commit?

tests/lax_numpy_test.py
charmerDark: Wouldn't the precommit hooks catch on to these? Mine only showed a lot of errors with some other files

PR #7130, opened 2021-06-29
custom_jvp named shape hotfix

jekbradbury:
When we added "avals with names", we intended to start by making the distinction between types with and without named axes load-bearing only in specific parts of the system, while (continuing to) ignore it elsewhere. This fixes a spot I missed, and that a user ran into.

Most likely, we'll want to restore something like this typecheck after vmap and pmap use avals with names; for now, the typecheck won't always be satisfied in those contexts and needs to be loosened.

CC @dpfau


dpfau:
Thanks for the quick fix!

PR #7132, opened 2021-06-29
Fix broken link in 06-parallelism.ipynb

shawwn:
It looks like the notebook `07-state-in-jax.ipynb` was renamed to `07-state.ipynb`, causing a link to it to break. This commit fixes the link by using the correct name.

### DIFF
--- a/docs/jax-101/06-parallelism.ipynb
+++ b/docs/jax-101/06-parallelism.ipynb
@@ -544,7 +544,7 @@
     "* the `update()` function\n",
     "* the replication of parameters and splitting of data across devices.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


jakevdp:
Thanks! We'll need the equivalent change in the associated markdown file as well.

shawwn:
> Thanks! We'll need the equivalent change in the associated markdown file as well.

Done!

jakevdp (changes requested):
Looks good, thanks! The last thing we need is to squash these changes into a single commit (See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Thanks!

shawwn:
> Looks good, thanks! The last thing we need is to squash these changes into a single commit (See [jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests](https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)) Thanks!

Sure thing! I squashed the commits.

jakevdp (approved):
Thanks!

PR #7146, opened 2021-06-30
Giving installation source for gpu/tpu no longer required.

lumip:
Installing jax for GPU/TPU previously required to provide the jaxlib release storage url to pip as an additional argument, making this more complicated than it needs to be. This PR hides this from the user by adding a small code snippet to `setup.py` that figures out the exact source URL for the required jaxlib wheel and uses PEP 508 URL notation to remove the need for adding an additional package index by the user.

tl;dr:
`pip install --upgrade "jax[cudaXXX]" -f https://storage.googleapis.com/jax-releases/jax_releases.html`
becomes
`pip install --upgrade "jax[cudaXXX]"`.
Similar for TPU installation.

I also took the liberty of completing the instructions for dealing with non-standard CUDA installation paths that seemed to be incomplete in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,7 @@ Next, run
 
 ```bash
 pip install --upgrade pip
-pip install --upgrade "jax[cuda111]" -f https://storage.googleapis.com/jax-releases/jax_releases.html
+pip install --upgrade "jax[cuda111]"
 ```
 
 The jaxlib version must correspond to the version of the existing CUDA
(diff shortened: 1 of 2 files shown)


lumip:
It seems now like this will run into problems with `pip`s "Packages installed from PyPI cannot depend on packages which are not also hosted on PyPI." error, I'll have to check that.

jakevdp:
Thanks for exploring this – we tried this approach and came to the same conclusion as you (though unfortunately not before releasing JAX v0.2.15, which contains a setup.py file similar to the one in this PR!) I wish there were a cleaner way to do this, but I also understand the security concerns that led to this check on PyPI.

lumip:
I frankly don't see how their policy actually alleviates security concerns, but that's a moot point here. Do you perhaps have plans on putting the cuda jaxlib releases at PyPI at some point?

PR #7147, opened 2021-06-30
Broadcast arrays manually in categorical sampling.

amol-mandhane:
This enables sampling with jax_numpy_rank_promotion=raise.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1306,7 +1306,10 @@ def categorical(key: jnp.ndarray,
     _check_shape("categorical", shape, batch_shape)
 
   sample_shape = shape[:len(shape)-len(batch_shape)]
-  return jnp.argmax(gumbel(key, sample_shape + logits.shape, logits.dtype) + logits, axis=axis)
+  return jnp.argmax(
+      gumbel(key, sample_shape + logits.shape, logits.dtype) +
+      lax.expand_dims(logits, tuple(range(len(sample_shape)))),
+      axis=axis)
 
 
 def laplace(key: jnp.ndarray,
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -379,6 +379,7 @@ class LaxRandomTest(jtu.JaxTestCase):
     ]
     for sample_shape in [(10000,), (5000, 2)]
     for dtype in jtu.dtypes.floating))
+  @jtu.disable_implicit_rank_promotion
   def testCategorical(self, p, axis, dtype, sample_shape):
     key = random.PRNGKey(0)
     p = np.array(p, dtype=dtype)


jakevdp (commented):
Looks good, thanks! Can you add a short regression test?

amol-mandhane:
Thanks.  I modified random_test.py:testCategorical to run with jax_numpy_rank_promotion=raise.

jakevdp:
Thanks, this looks good.

I wonder, though: should we expect JAX API functions to avoid implicit rank promotion? If so, we should probably test this more globally (as you found, it's largely untested). If not, then it's not clear to me that this change is necessary.

Assuming that we want JAX internals to avoid implicit rank promotion, it might motivate a more re-usable approach here; for example:
- on the testing side, we might use a decorator on the test method to temporarily enable rank promotion rather than doing it manually
- Alternatively, if we want this behavior globally we could have a separate test run with rank promotion warnings enabled via a flag on the invocation (I suspect there would be lots of failures currently)
- on the implementation side, we might instead have a context manager that temporarily disables the rank promotion warning in places where we explicitly choose. That might be a better fix than doing explicit broadcasting as in this PR.

What do you think?

amol-mandhane:
Locally, I need this change because we run our experiments with jax_numpy_rank_promotion=raise and we want to do categorical sampling.  We find this setting very helpful in catching errors early at the cost of manually broadcasting every time.

Globally, I think JAX API should guarantee full compatibility with jax_numpy_rank_promotion=raise since this setting would be pointless if JAX API itself can't support it.  The config is quite important as it can allow users to build robust setups.

Personally, I think the best option in long term is to run all tests with jax_numpy_rank_promotion=raise as it increases confidence in the tests further.  A reasonable testing solution would be to subclass JaxTestCase and set the config in setUpClass.  However, this would involve broadcasting fixes across the entire framework and would be a large change.  If this feels opinionated and disruptive, I think the next best solution would be a decorator the disallow rank promotion in selected test.

jakevdp:
Thanks - that makes sense to me. With that in mind, would you be willing to create such a decorator in `jax.test_util` and use it here in your test?

amol-mandhane:
Sending a separate PR for it.

jax/_src/random.py
```diff
@@ -1306,7 +1306,10 @@ def categorical(key: jnp.ndarray,
     _check_shape("categorical", shape, batch_shape)
 
   sample_shape = shape[:len(shape)-len(batch_shape)]
-  return jnp.argmax(gumbel(key, sample_shape + logits.shape, logits.dtype) + logits, axis=axis)
+  return jnp.argmax(
+      gumbel(key, sample_shape + logits.shape, logits.dtype) +
+      jnp.broadcast_to(logits, sample_shape + logits.shape),
```
jakevdp, line 1311: I just realized this is a full `broadcast_to`, which actually materializes a larger array through duplication of the entries. I think we could avoid the rank promotion warning more efficiently by using `lax.expand_dims` instead to insert leading dimension of size 1 up to the desired rank.

amol-mandhane:
Will wait for #7192 before resolving this.

jakevdp:
#7192 is in! Once you update this PR to utilize that new utility, I'll take another look. Thanks!

jax/_src/random.py
amol-mandhane: Done

amol-mandhane:
PTAL.

amol-mandhane:
CI failure appears to be unrelated.

jakevdp (approved):
Very nice!

PR #7160, opened 2021-07-01
Add partitioning support to outfeed.

pschuh:
This is needed to support outfeed from inside pjit.


jax/_src/lax/lax.py
```diff
@@ -6434,23 +6434,31 @@ def _infeed_translation_rule(c, token, *, shapes, partitions):
 infeed_p.def_abstract_eval(_infeed_abstract_eval)
 xla.translations[infeed_p] = _infeed_translation_rule
 
-def outfeed(token, xs):
+def outfeed(token, xs, partitions = None):
   """Outfeeds value `xs` to the host. Experimental.
 
   `token` is used to sequence infeed and outfeed effects.
```
gnecula, line 6440: Maybe extend the docstring with how `partitions` works?

pschuh: Done

PR #7177, opened 2021-07-02
setup.py: change jax[cpu] to target the current jaxlib version

### LINKED ISSUE #7174: Constraints for jaxlib version during jax installation
In v0.2.11 jax added the nice optional installation targets `cpu`, `cudaXXX` and `tpu` which install jaxlib for the desired platform. Currently, the setup script guarantees that the installed version of jaxlib is compatible by constraining the version number from below to the earliest known compatible version. However, no similar constraint is put for a maximum compatible version number. Given that jaxlib does not appear to necessarily maintain backward compatibility between updates, this is an issue for installing older versions of jax (as these will always pull the latest available jaxlib version, leading to crashes due to incompatibility).

Jax installation should always result in a working setup, i.e., only the latest compatible version of jaxlib should be installed when installing jax. This either requires version constraints in jax's setup.py or backward compatibility guarantees in jaxlib (or a combination of both).

Please:

jakevdp:
Fixes #7174

Because jaxlib is not always backward compatible with old versions of JAX, it makes sense to pin the current version of the CPU wheel so that in the future it's easier to install a working jax/jaxlib combination for old versions of JAX. Note that we already do this for GPU and TPU.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
 
         # CPU-only jaxlib can be installed via:
         # $ pip install jax[cpu]
-        'cpu': [f'jaxlib>={_minimum_jaxlib_version}'],
+        'cpu': [f'jaxlib=={_current_jaxlib_version}'],
 
         # Cloud TPU VM jaxlib can be installed via:
         # $ pip install jax[tpu] -f https://storage.googleapis.com/jax-releases/jax_releases.html


skye (approved):
Thanks Jake! Agree this makes more sense, and will also probably make it easier to pull in newer jaxlibs (because why not?).

PR #7179, opened 2021-07-02
[sparse] Implement several BCOO methods via sparsify

jakevdp:
I kept expecting these to work, but I hadn't implemented them yet...

This just gets the functionality in & leaves refactoring to avoid circular imports as a TODO.


froystig (approved):
nice!!

PR #7180, opened 2021-07-02
[sparse] add support for sparse-sparse matrix products

jakevdp:
This implements the equivalent of `lax.dot_general` between two BCOO sparse matrices, for matrices with a single sparse dimension.

Still TODO:
- multiple sparse dimensions (will take some thought)
- batching (should be easy)
- autodiff (should be excruciating)

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -1176,6 +1176,140 @@ xla.translations[bcoo_dot_general_sampled_p] = xla.lower_fun(
     _bcoo_dot_general_sampled_impl, multiple_results=False)
 
 #----------------------------------------------------------------------
+# bcoo_spdot_general
+# (batched) general dot product of two BCOO sparse arrays returning a
+# Dense ND array.
+
+bcoo_spdot_general_p = core.Primitive('bcoo_spdot_general')
+
+def bcoo_spdot_general(lhs_data, lhs_indices, rhs_data, rhs_indices, *, lhs_shape, rhs_shape, dimension_numbers):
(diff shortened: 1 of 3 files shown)


sunilkpai:
This looks great. Is there any intention to implement the vjp/jvp for sparse-sparse as well?

jakevdp:
Yeah, I wouldn't consider this ready to merge until all the autodiff & batching rules are implemented (see the TODO in the code)

jakevdp:
By request, I think I'll get this impl merged, and leave batching and autodiff as a TODO.

jax/experimental/sparse/ops.py
```diff
@@ -1175,6 +1175,139 @@ def impl(A, B, indices):
...
+    batch_dims[1] = 0
+  assert batch_dims[0] == batch_dims[1] == 0
+  if batch_dims[2] is None:
+    rhs_data = rhs_data[None]
+    batch_dims[2] = 0
+  if batch_dims[3] is None:
+    rhs_indices = rhs_indices[None]
+    batch_dims[3] = 0
+  # TODO: handle batchings along different axes?
+  assert all(dim == 0 for dim in batch_dims)
```
mattjj, line 1296: Maybe a `raise NotImplementedError` instead? I always suggest this kind of thing because in theory someone could run with `python -O` and elide all assertions.

mattjj (approved):
Damn fine code!

jax/experimental/sparse/ops.py
jakevdp: Oh, good call. Done!

PR #7188, opened 2021-07-04
Add documentation for jax.device_get

### LINKED ISSUE #6445: Document jax.device_get
The function `jax.device_get` is not documented. Preferably, it should appear on this page
https://jax.readthedocs.io/en/latest/jax.html

The code for the function also does not provide any doc-string.
https://github.com/google/jax/blob/26e9ebcdae5e610d1173fac1bc3504a9c3f2011f/jax/_src/api.py#L2258

changjonathanc:
Fixes #6445 

Will readthedocs be generated automatically or do I need to edit elsewhere?

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2328,7 +2328,35 @@ def _device_get(x):
   else:
     return copy()
 
-def device_get(x):
+def device_get(x: Any):
+  """Transfer ``x`` to host.
+
+  Args:
+    x: An array, scalar, DeviceArray or (nested) standard Python container thereof
+      representing the array to be transferred to host.
+
+  Returns:
+    An array or (nested) Python container thereof representing the
+    value of ``x``.
+
+  Examples:
+    Passing a DeviceArray:
+
+    >>> import jax
+    >>> x = jax.numpy.array([1., 2., 3.])
+    >>> jax.device_get(x)
+    array([1., 2., 3.], dtype=float32)
+
+    Passing a scalar (has no effect):
+
+    >>> jax.device_get(1)
+    1
+
+  See Also:
+    - device_put
+    - device_put_sharded
+    - device_put_replicated
+  """
   for y in tree_leaves(x):
     try:
       y.copy_to_host_async()
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -36,6 +36,7 @@ Just-in-time compilation (:code:`jit`)
     device_put
     device_put_replicated
     device_put_sharded
+    device_get
     default_backend
     named_call
 
@@ -91,6 +92,7 @@ Parallelization (:code:`pmap`)
 .. autofunction:: device_put
 .. autofunction:: device_put_replicated
 .. autofunction:: device_put_sharded
+.. autofunction:: device_get
 .. autofunction:: default_backend
 .. autofunction:: named_call
 


cccntu:
I tried `sphinx-build -b html -D jupyter_execute_notebooks=off docs docs/build/html` successfully, so I rebased and force-pushed.

jakevdp (changes requested):
Thanks for the contribution - sorry we missed this previously.

Regarding your question: readthedocs will build from the PR automatically, and the link can be found under "details" in the readthedocs CI job (The link for this PR is https://jax--7188.org.readthedocs.build/en/7188/)

A couple comments on this change:

- `device_get()` works for standard DeviceArray objects as well as ShardedDeviceArray objects; the current version makes it sound like it's *only* sharded arrays
- I don't think these new docs will appear anywhere on the docs website currently; I think `device_get` would have to be added to the docs here: https://github.com/google/jax/blob/main/docs/jax.rst#just-in-time-compilation-jit

cccntu:
@jakevdp 
Thanks for the comments, I have updated it accordingly.

jax/_src/api.py
```diff
@@ -2328,7 +2328,47 @@ def _device_get(x):
...
+def device_get(x: Any):
+  """Transfer ``x`` to host.
+
+  Args:
+    x: An array, scalar, DeviceArray or (nested) standard Python container thereof
+      representing the array to be transferred to host.
+
+  Returns:
+    An array or (nested) Python container thereof representing the
+    value of ``x``.
```
jakevdp, line 2340: I believe a line break is required here for the docs to format correctly

jax/_src/api.py
```diff
@@ -2328,7 +2328,47 @@ def _device_get(x):
...
+  Examples:
+    Passing a DeviceArray:
+
+    >>> import jax
+    >>> devices = jax.local_devices()
+    >>> x = jax.numpy.array([1., 2., 3.])
+    >>> y = jax.device_put(x, device=devices[0])
```
jakevdp, line 2347: There is no need to call `device_put` on a jax array: it already lives on a device. You could skip `device_put`, or you could define `x` as a standard numpy array, in which case `device_put` puts it on the device.

jakevdp (changes requested):
Looks good! Sorry for the delay: I was on holiday for a bit.

A couple last comments - also we need to confirm that the CI passes.

Once that's all set, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) and then we can get this merged!

jax/_src/api.py
cccntu: I am not sure what you mean, the docs looks correct to me. But I added an extra line after line 2340.

jakevdp (approved):

cccntu:
Thank you for your patient reviews!

The test failed because the example is slightly off, so I pushed another fix, then squashed and forced pushed.

```diff
     >>> import jax
     >>> x = jax.numpy.array([1., 2., 3.])
-    >>> x
-    DeviceArray([[1., 2., 3.]], dtype=float32)
     >>> jax.device_get(x)
-    array([[1., 2., 3.]], dtype=float32)
+    array([1., 2., 3.], dtype=float32)
```

Would it be better if I keep all the logs and create another PR with squashed version, or you could somehow see what I pushed before?

jakevdp:
For a small change like this, squashing & force-pushing is fine. Thanks!

cccntu:
I found the command for the failing test and ran it locally, should be fine this time. 😅 
note: I deleted the example with `ShardedDeviceArray` for simplicity.

jakevdp:
Thanks for the contribution!

Issue #7063, opened 2021-06-23
Loops don't track axis names correctly


apaszke:
As reported by @sharadmv:
```py
import jax
from jax import lax
from jax.experimental import maps

def f(x):
  z = x + lax.axis_index('a')
  y = x + lax.axis_index('b')
  def cond(carry):
    i, x = carry
    return x < 5
  def body(carry):
    i, x = carry
    return i + 1, x + lax.psum(y, 'b')
  return lax.while_loop(cond, body, (0, z))[1]
maps.xmap(f, axis_sizes=dict(a=2, b=10), out_axes=(['a']), in_axes={})(1.)
```

PR #7206, opened 2021-07-07
Fix while loop fixed point condition

sharadmv:
Adjust the while loop fixed point to address the error in #7063. 

Here's the relevant repro:
```python
import jax
from jax import lax
from jax.experimental import maps

def f(x):
  z = x + lax.axis_index('a')
  y = x + lax.axis_index('b')
  def cond(carry):
    i, x = carry
    return x < 5
  def body(carry):
    i, x = carry
    return i + 1, x + lax.psum(y, 'b')
  return lax.while_loop(cond, body, (0, z))[1]
maps.xmap(f, axis_sizes=dict(a=2, b=10), out_axes=(['a']), in_axes={})(1.)
```

The translation rule for while loop requires that the predicate shape is a prefix for the carry shapes (if predicate is scalar, this is trivially true). However, when predicate is *not* a scalar, we need to make sure that this is the case. Previously, if the predicate was not a scalar, we'd batch every value in the carry and batch the predicate. This would ensure that the predicate shape remains a prefix of the carry shape, satisfying the condition in the while loop. However, this overbatching can cause problems.

In the repro, the batching rule for `while` is triggered twice, once for axis `'a'` and once for axis `'b'`. The body has the `x` value in the carry, which mapped across `'a'`, and closes over `y` (which has a `'b`' axis). However, the input and output carry axes do not include `'b'`, because it is summed out with the `psum`.

With the current batching rule, for `'a'` we properly batch the body and cond of the while loop. In the batching rule for `'b'`, specifically in the fixed point, we instantiate the output for the cond_jaxpr because `bool(cond_jaxpr.avals.out_shape[0])` is `True` (this is because we have batched the cond for `'a'`). This causes us to map the entire `carry`, which adds an axis to the carry for `'b'` when none needs to exist.

What we'd instead like to do is not `instantiate` if the predicate is is non-scalar. We check the `pred_bat` produced by `batching.batch_jaxpr` (which tells us if the predicate is batched or not). If it is, we do need to batch everything and reproduce the `instantiate=True` behavior. If it is not, we need to make sure to have the batched predicate shape still be a prefix of the carry shapes. We accomplish this with some extra bookkeeping and modifications to `batching.batch_jaxpr` to allow us to thread axes in as opposed to bools indicating whether something is batched on the leading dimension.

Co-authored-by: Adam Paszke <apaszke@google.com>

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -373,7 +373,7 @@ def _pred_bcast_select(c, pred, x, y, x_y_aval: core.AbstractValue):
   elif x_y_aval is core.abstract_token:
     return xops.AfterAll(c, [x, y])
   else:
-    assert pred_shape == x_shape[:len(pred_shape)] == y_shape[:len(pred_shape)]
+    assert pred_shape == x_shape[:len(pred_shape)] == y_shape[:len(pred_shape)], (pred_shape, x_shape, y_shape)
     bcast_pred = xops.BroadcastInDim(pred, x_shape, list(range(len(pred_shape))))
     return xops.Select(bcast_pred, x, y)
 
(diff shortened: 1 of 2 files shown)


jax/core.py
```diff
@@ -1748,7 +1748,9 @@ def subst_axis_names_jaxpr(jaxpr: Union[Jaxpr, ClosedJaxpr], subst: AxisSubst):
   if isinstance(jaxpr, ClosedJaxpr):
     consts = jaxpr.consts
     jaxpr = jaxpr.jaxpr
-  var_map: Dict[Var, Var] = {}
+  var_map: Dict[Var, Var] = {
+      unitvar: unitvar
```
apaszke, line 1752: Why is this necessary? unitvar is specially handled in `subst_axis_names_var`

jax/_src/lax/control_flow.py
```diff
@@ -383,43 +383,77 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
...
   cconst_bat, bconst_bat, init_bat = split_list(orig_batched, [cond_nconsts, body_nconsts])
+  cconsts, bconsts, init = split_list(args, [cond_nconsts, body_nconsts])
+  cconst_dims, bconst_dims, init_dims = split_list(dims, [cond_nconsts, body_nconsts])
 
+  carry_bat = init_bat
   # Fixpoint computation of which carry are batched: either
   # batched from init, or the carry out is batched. Each iteration promotes
   # at least one carry to batched. We need at most len(carry) iterations,
   # but we need one last iteration to prepare the jaxpr based on the final
   # carry_bat.
```
apaszke, line 394: This comment needs an update (e.g. the details of last iteration no longer hold)

jax/_src/lax/control_flow.py
```diff
@@ -383,43 +383,77 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
...
-    carry_bat_out = _map(partial(operator.or_, pred_bat), carry_bat_out)
-    if carry_bat_out == carry_bat:
+    if carry_bat == carry_bat_out:
       break
-    else:
-      carry_bat = _map(operator.or_, carry_bat, carry_bat_out)
+    carry_bat = safe_map(operator.or_, carry_bat, carry_bat_out)
   else:
-    assert False, "Fixpoint not reached"
+    raise ValueError("Fixpoint not reached")
```
apaszke, line 403: Can you please make this `AssertionError`? `ValueError` is for invalid argument values, but this looks more like an internal logic error

apaszke (approved):
Left some minor comments, but LGTM

jax/core.py
sharadmv: Done.

jax/_src/lax/control_flow.py

jax/_src/lax/control_flow.py
sharadmv: Oops, forgot to change this back. Thanks for the catch.

sharadmv:
Addressed the comments, thanks for the review!

PR #7207, opened 2021-07-07
Complete compilation_cache implementation

colemanliyah:
implemented initialize_cache, get_executable, and put_executable functions

### DIFF
--- a/jax/experimental/compilation_cache/compilation_cache.py
+++ b/jax/experimental/compilation_cache/compilation_cache.py
@@ -14,7 +14,41 @@
 
 import hashlib
 import jax
+from jax.experimental.compilation_cache.file_system_cache import FileSystemCache
+from jax.lib import xla_client
+from typing import Optional
 
+_cache = None
+
+def initialize_cache(path):
+    """Creates a global cache object. Should only be called once per process."""
+    global _cache
+    assert _cache == None, f"The cache path has already been initialized to {_cache}"
+    _cache = FileSystemCache(path)
+
+def get_executable(xla_computation, compile_options) -> Optional[xla_client.Executable]:
+    """Returns the cached executable if present, or None otherwise."""
+    assert _cache is not None, "initialize_cache must be called before calling this function."
+    cache_key = get_cache_key(xla_computation, compile_options)
+    xla_executable_serialized = _cache.get(cache_key)
+    if not xla_executable_serialized:
+        return None
+    backend = jax.lib.xla_bridge.get_backend()
+    # TODO(skye): xla_computation.get_hlo_module() is the unoptimized HLO but it should
+     #be optimized
+    xla_executable_deserialized = backend.deserialize_executable(
+                                  xla_executable_serialized,
+                                  xla_computation.get_hlo_module(),
+                                  compile_options)
+    return xla_executable_deserialized
+
+def put_executable(xla_computation, compile_options, executable: xla_client.Executable):
+    """Adds 'executable' to the cache, possibly evicting older entries."""
+    assert _cache is not None, "initialize_cache must be called before calling this function."
+    cache_key = get_cache_key(xla_computation, compile_options)
+    backend = jax.lib.xla_bridge.get_backend()
+    serialized_executable = backend.serialize_executable(executable)
+    _cache.put(cache_key, serialized_executable)
 
 def get_cache_key(xla_computation, compile_options) -> str:
     """Creates a hashed string to use as a key to the compilation cache.
--- a/tests/compilation_cache_test.py
+++ b/tests/compilation_cache_test.py
@@ -19,7 +19,9 @@ import jax
 import jax.test_util as jtu
 import numpy as np
 import random
+import tempfile
 import unittest
+from unittest import SkipTest
 
 class CompilationCacheTest(jtu.JaxTestCase):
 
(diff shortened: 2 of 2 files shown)


tests/compilation_cache_test.py
```diff
@@ -117,6 +118,44 @@ def test_different_computations(self):
...
+          self.assertEqual(cc.get_executable(computation, compile_options), None)
+
+  def test_diff_executables(self):
+      with tempfile.TemporaryDirectory() as tmpdir:
+          cc.initialize_cache(tmpdir)
+          computation1 = jax.xla_computation(lambda x, y: x + y)(1, 1)
+          computation2 = jax.xla_computation(lambda x, y: x * y)(2, 2)
+          compile_options = jax.lib.xla_bridge.get_compile_options(
+                                num_replicas=1, num_partitions=1)
+          lib_tpu = jax.lib.xla_bridge.get_backend()
```
skye, line 136: Rename to `backend`

tests/compilation_cache_test.py
```diff
@@ -117,6 +118,44 @@ def test_different_computations(self):
...
+          self.assertNotEqual(cc.get_executable(computation1, compile_options),
+                              cc.get_executable(computation2, compile_options))
+
+  def test_put_executable(self):
+      with tempfile.TemporaryDirectory() as tmpdir:
+          cc.initialize_cache(tmpdir)
+          computation = jax.xla_computation(lambda x, y: x + y)(1, 1)
+          compile_options = jax.lib.xla_bridge.get_compile_options(
+                               num_replicas=1, num_partitions=1)
+          lib_tpu = jax.lib.xla_bridge.get_backend()
```
skye, line 150: Rename to `backend`

skye (changes requested):
Looks great! Just a few small comments.

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -14,7 +14,44 @@
...
+
+def initialize_cache(path):
+    """Creates a global cache object. Should only be called once per process."""
+    global cache
+    cache = FileSystemCache(path)
+
+def get_executable(xla_computation, compile_options) -> Optional[xla_client.Executable]:
+    """Returns the cached executable if present, or None otherwise."""
+    if not cache:
+        return None
```
skye, line 32: Let's actually change this to `assert cache is not None`, meaning the caller is responsible for calling `initialize_cache` before calling `get_executable` (or not calling `get_executable` at all if they don't wanna initialize the cache). Also note this requirement in the docstring, something like: "`initialize_cache` must be called before calling this function."

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -14,7 +14,44 @@
...
+                                  xla_executable_serialized,
+                                  xla_computation.get_hlo_module(),
+                                  compile_options)
+    return xla_executable_deserialized
+
+def put_executable(xla_computation, compile_options, executable: xla_client.Executable):
+    """Adds 'executable' to the cache, possibly evicting older entries."""
+    if not cache:
+        warnings.warn("file path for cache has not been specified")
+        return
```
skye, line 50: Same here, make this `assert cache is not None`, and add a note to the docstring.

skye (changes requested):
A few more minor comments. I think this is ready for @zhangqiaorjc to look at.

skye:
Also, don't forget to fix the CLA.

jax/experimental/compilation_cache/compilation_cache.py
colemanliyah: Is there a way we can do both? I do want to return None if there is not a cache because my logic for the pmap integration handles this but if I just use an assert statement then I  the logic wont go through with the pmap integration it will just print out this message...although maybe that is preferable...I'm not sure, what do you think?

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -14,7 +14,41 @@
...
 import jax
+from jax.experimental.compilation_cache.file_system_cache import FileSystemCache
+from jax.lib import xla_client
+from typing import Optional
 
+_cache = None
+
+def initialize_cache(path):
+    """Creates a global cache object. Should only be called once per process."""
+    global _cache
```
zhangqiaorjc, line 25: if _cache already initialized, raise an exception?

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -14,7 +14,41 @@
...
+                                  xla_executable_serialized,
+                                  xla_computation.get_hlo_module(),
+                                  compile_options)
+    return xla_executable_deserialized
+
+def put_executable(xla_computation, compile_options, executable: xla_client.Executable):
+    """Adds 'executable' to the cache, possibly evicting older entries."""
+    assert _cache is not None, "initialize_cache must be called before calling this function."
+    cache_key = get_cache_key(xla_computation, compile_options)
+    cache_key = get_cache_key(xla_computation, compile_options)
```
zhangqiaorjc, line 48: remove this redundant line?

zhangqiaorjc (approved):

skye (approved):

PR #7210, opened 2021-07-08
Add config.define_string_state.

skye:
Also consolidates some shared config logic into a new _StateContextManager class.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -228,7 +228,7 @@ class Config:
       return val if val is not unset else self._read(name)
     setattr(Config, name, property(get_state))
 
-    return _BoolStateContextManager(name, help, update_thread_local_hook)
+    return _StateContextManager(name, help, update_thread_local_hook)
 
   def define_enum_state(
       self, name: str, enum_values: List[str], default: Optional[str],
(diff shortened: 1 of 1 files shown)


jax/_src/config.py
```diff
@@ -295,33 +324,40 @@ def _trace_context(self):
     return (self.x64_enabled, self.jax_numpy_rank_promotion,
             self.jax_default_matmul_precision)
 
-class _BoolStateContextManager:
-  def __init__(self, name, help, update_thread_local_hook):
+class _StateContextManager:
+  def __init__(self, name, help, update_thread_local_hook,
+               validate_new_val_hook: Callable[[Any], None] = None):
```
mattjj, line 329:
```suggestion
               validate_new_val_hook: Optional[Callable[[Any], None]] = None):
```

jax/_src/config.py
```diff
@@ -263,27 +263,56 @@ def get_state(self):
...
+    See docstring for ``define_bool_state``.
+
+    Args:
+      name: string, converted to lowercase to define the name of the config
+        option (and absl flag). It is converted to uppercase to define the
+        corresponding shell environment variable.
+      default: boolean, a default value for the option.
+      help: string, used to populate the flag help information as well as the
+        docstring of the returned context manager.
+      update_global_hook: a optional callback that is called with the updated
```
mattjj, line 290:
```suggestion
      update_global_hook: an optional callback that is called with the updated
```

jax/_src/config.py
```diff
@@ -263,27 +263,56 @@ def get_state(self):
...
+        value of the global state when it is altered or set initially.
+      update_thread_local_hook: a optional callback that is called with the
```
mattjj, line 292:
```suggestion
      update_thread_local_hook: an optional callback that is called with the
```

jax/_src/config.py
```diff
@@ -263,27 +263,56 @@ def get_state(self):
...
+      = None):
+    """Set up thread-local state and return a contextmanager for managing it.
+
+    See docstring for ``define_bool_state``.
+
+    Args:
+      name: string, converted to lowercase to define the name of the config
+        option (and absl flag). It is converted to uppercase to define the
+        corresponding shell environment variable.
+      default: boolean, a default value for the option.
```
mattjj, line 287:
```suggestion
      default: string, a default value for the option.
```

jax/_src/config.py
```diff
@@ -263,27 +263,56 @@ def get_state(self):
...
-    set_state.__doc__ = f"Context manager for `{name}` config option.\n\n{help}"
-    return set_state
+
+    return _StateContextManager(name, help, update_thread_local_hook, validate)
+
+  def define_string_state(
+      self, name: str, default: Optional[str], help: str,
+      update_global_hook: Optional[Callable[[str], None]] = None,
+      update_thread_local_hook: Optional[Callable[[Optional[str]], None]]
+      = None):
```
mattjj, line 278:
```suggestion
      = None) -> ContextManager[None]:
```

(I _think_ that annotation works...)

And maybe join the `= None` part onto the previous line, followed by a newline? I dno. It looks lonely.

jax/_src/config.py
```diff
@@ -295,33 +324,40 @@ def _trace_context(self):
...
     self.__name__ = name[4:] if name.startswith('jax_') else name
     self.__doc__ = f"Context manager for `{name}` config option.\n\n{help}"
-    self._hook = update_thread_local_hook
+    self._update_thread_local_hook = update_thread_local_hook
+    self._validate_new_val_hook = validate_new_val_hook
 
   @contextlib.contextmanager
-  def __call__(self, new_val: bool):
+  def __call__(self, new_val):
+    if self._validate_new_val_hook:
```
mattjj, line 338: Do you want `is not None` for consistency with the other lines?

(Actually I prefer the way this line is written, since we know it's either a callable or None. Whatever floats your boat though, I just wanted to flag the inconsistency!)

mattjj (approved):
Awesome! Some optional comments.

jax/_src/config.py
skye: Seems to work! I put `= None` with its friends.

jax/_src/config.py
skye: Good catch. I changed the others to not have `is not None` for brevity.

jax/_src/config.py
skye: Actually the ContextManager annotation doesn't work (and I apparently don't know how to run mypy). It says:
`Incompatible return value type (got "_StateContextManager", expected "ContextManager[None]")`
I'm guessing because we're using a weird context manager pattern. I'm just gonna omit it, unless you have ideas.

Issue #7208, opened 2021-07-07
Run test suite with jax_numpy_rank_promotion=raise


jakevdp:
This setting is currently not always useful because the JAX source itself often generates rank promotion errors. We should fix existing issues and change the test suite to run with this setting enabled.

xref #7147

PR #7223, opened 2021-07-08
Fix rank promotion warning in jnp.trapz

jakevdp:
Part of #7208


jakevdp:
So I'm actually a bit torn on this one... unlike some of the previous fixes, here we are actually reshaping user inputs to make their ranks match. Should we be doing this? Or is this a case where a rank promotion warning/error would be expected?

froystig:
Do many current uses, starting with the JAX source itself (per #7208), expect the reshape-to-promote behavior? If not then maybe that's a sign that we can still choose to err?

jakevdp:
I'm going to close this: I think since it's ranks of user-inputs that generate the rank promotion warning, the warning is expected here.

PR #7234, opened 2021-07-09
[sparse] add sparse transform rule for lax.while_p

jakevdp:
May need some more extensive test cases...

### DIFF
--- a/jax/experimental/sparse/transform.py
+++ b/jax/experimental/sparse/transform.py
@@ -58,7 +58,7 @@ from jax import linear_util as lu
 from jax.api_util import flatten_fun_nokwargs
 from jax.interpreters import partial_eval as pe
 from jax.tree_util import tree_flatten, tree_unflatten
-from jax.util import safe_map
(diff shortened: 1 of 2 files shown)


froystig (commented):
Judging by its callers, does the current `_sparsify_jaxpr` have a more general interface than we need, and would less indirection be useful? Instead of accepting `AnyArray`s, flattening, and creating new indices/data mappings, could it accept the entire environment from the caller? This might also help address the TODO comment about converting indices/data mappings to arrays and back.

jakevdp:
Thanks - I think refactoring `_sparsify_jaxpr` will help, but it doesn't get us all the way to addressing the TODO because we're still having to convert the argspecs to arrays in order to pass them to `lax.while_p.bind`, which then calls our transformed jaxprs, which then re-construct the argspecs from their inputs.

It's not obvious to me that there's any way to get around that.

froystig (approved):
looks good with the new test and comment!

PR #7251, opened 2021-07-12
efficiency: use lax.broadcast_to_rank instead of jnp.broadcast_to in shape promotion

jakevdp:
Why? This is a frequently-used utility (called in all broadcasted numpy binary ops!) and `jnp.broadcast_to` includes a bunch of validation overhead that is not relevant to this case where we know a priori that we're just adding leading ones to the shape.


jekbradbury (approved):

jakevdp:
Turns out the overhead is necessary to correctly handle dynamic shapes.

PR #7264, opened 2021-07-13
Cleanup try_broadcast_shapes.

gnecula:
This code got more complicated to handle sound comparisons of dimension polynomials. See [diff](https://github.com/google/jax/commit/799325bbba352842a56ca225f02b6eb970f45712#diff-a81a5cc54c9a9c2797bcca38f5f1c13d5c9883763ee1e14ae5f0e215aa115bcfL65-R80).

Since then we have added some utility functions, e.g., `core.symbolic_equal_dim`, so we can revert the code to something simpler, and close to what it was originally.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -67,18 +67,13 @@ Shape = core.Shape
 
 def _try_broadcast_shapes(shapes):
   assert shapes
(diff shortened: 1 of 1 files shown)


gnecula:
Actually, this is not much of a simplication, dropping.

PR #7273, opened 2021-07-13
Compute gtsv2 buffer size ahead of time and pass in to kernel.

tomhennigan:
(Apologies for the large diff due to improving the error messages. The only important functional change is at the end of the file with `s/cuda{Free,Malloc}Async/cuda{Free,Malloc}/`)

A user reported that with their Quadro M4000 GPU (Driver: 460.56) tridiagonal_solve was throwing `operation not supported`. I improved the logging (also included in this patch) and tracked it down to:

    jax._src.traceback_util.UnfilteredStackTrace: RuntimeError: third_party/py/jax/jaxlib/cusparse.cc:902: CUDA operation cudaMallocAsync(&buffer, bufferSize, stream) failed: operation not supported

It's unclear why this is not supported from [the docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html#group__CUDART__MEMORY__POOLS_1gbbf70065888d61853c047513baa14081), I was able to run the original code on an A100 machine (same driver version, same binary etc) with no issue. For now I'll revert so we're using the sync APIs such that this works.

### DIFF
--- a/jaxlib/cusparse.cc
+++ b/jaxlib/cusparse.cc
@@ -841,9 +841,9 @@ py::bytes BuildGtsv2Descriptor(int m, int n, int ldb) {
   return PackDescriptor(Gtsv2Descriptor{m, n, ldb});
 }
 
-template <typename T, typename F1, typename F2>
-void gtsv2(F1 computeGtsv2BufSize, F2 computeGtsv2, cudaStream_t stream,
-           void** buffers, const char* opaque, std::size_t opaque_len) {
+template <typename T, typename F>
+void gtsv2(F computeGtsv2, cudaStream_t stream, void** buffers,
+           const char* opaque, std::size_t opaque_len) {
   auto handle = SparseHandlePool::Borrow();
 
   const Gtsv2Descriptor& descriptor =
@@ -857,6 +857,7 @@ void gtsv2(F1 computeGtsv2BufSize, F2 computeGtsv2, cudaStream_t stream,
   const T* du = (const T*)(buffers[2]);
   const T* B = (T*)(buffers[3]);
   T* X = (T*)(buffers[4]);
+  void* buffer = buffers[5];
 
   // The solution X is written in place to B. We need to therefore copy the
   // contents of B into the output buffer X and pass that into the kernel as B.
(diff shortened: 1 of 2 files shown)


tomhennigan:
I've been debugging an unrelated cuSolver issue with another user, and decided to update this PR to consistently apply the `JAX_THROW_IF_ERROR` macro to all of the CUDA custom calls (cusparse, cusolver, cublas, prng kernels etc) so we have useful context in error messages.

This PR is now quite large, and combines both the error macro change, and the removal of `cudaMallocAsync`. If you have a strong preference for me to split this into two PRs please let me know.

apaszke (commented):
While the NVIDIA docs are terrible in this regard, couldn't we gate on the compute capability and disable this on only older GPUs? We shouldn't be regressing the new GPUs

tomhennigan:
> While the NVIDIA docs are terrible in this regard, couldn't we gate on the compute capability and disable this on only older GPUs? We shouldn't be regressing the new GPUs

Sounds good, I'll check a few of the GPUs we have access to and see if I can find a suitable compute version to anchor on.

In the mean time I've split out the error macro change into #7280 given that the change to alloc/free will be a bit more involved and deserves independent review.

tomhennigan:
OK so I've tested on P100 (`sm60`), V100 (`sm70`) and A100 (`sm80`) with driver 460.56 and async malloc works. On M4000 (`sm52`) with the same driver version it fails.

I also tested on V100 (`sm70`) with driver 450.80.02 and it fails (with `RuntimeError: API call is not supported in the installed CUDA driver`).

Given that, I think it's reasonable to gate this on driver version supports CUDA >= 11.2 and compute capability > 5.

jaxlib/cusparse.cc
```diff
@@ -854,6 +854,36 @@ py::bytes BuildGtsv2Descriptor(int m, int n, int ldb) {
...
+
+int cachedAsyncMallocSupported = -1;
+
+bool AsyncMallocSupported() {
+  if (cachedAsyncMallocSupported == -1) {
+    // This may get computed by multiple threads and overriden in paralllel,
+    // but we expect all threads to compute the same value (XLA requires
+    // replicated computations to be launched on equivalent devices) so it does
+    // not matter which threads result is written.
+    cachedAsyncMallocSupported = ComputeAsyncMallocSupported();
```
apaszke, line 882: Benign UB like this one might not be terrible, but it's still UB. But this would be perfectly thread safe:
```cpp
bool AsyncMallocSupported() {
  static bool cached = ComputeAsyncMallocSupported();
  return cached;
}
```

apaszke: Also, you're basing your decision on the compute capability which might vary between devices, so we might want to use e.g. a static array of booleans (32 is probably more than enough).

tomhennigan: Fair points, your solution was much simpler (I was not aware that this was thread safe). Luckily it is no longer needed since we can actually determine the buffer size ahead of time and get XLA to allocate it. Much simpler 😄

tomhennigan:
OK the code is now simpler and actually ready for review. Thank you for the discussion @apaszke! It turns out the cuSPARSE buffer size computation does not require the data, and can provide a buffer size given just the array sizes. As such we can compute it ahead of time, and ask XLA to allocate the buffer for us. This removes all calls to malloc.

I've tested and this version works nicely on V100 and A100.

PR #7277, opened 2021-07-13
Pmap integration

colemanliyah:


tests/compilation_cache_test.py
```diff
@@ -169,6 +174,42 @@ def test_put_executable(self):
...
+      if jtu.device_under_test() != "tpu":
+          raise SkipTest("serialize executable only works on TPU")
+      cc._cache = None
+      with tempfile.TemporaryDirectory() as tmpdir:
+          cc.initialize_cache(tmpdir)
+          f = pmap(lambda x: x - lax.psum(x, 'i'), axis_name='i')
+          shape = (jax.lib.xla_bridge.device_count(), 4)
+          x = np.arange(prod(shape), dtype=np.float32).reshape(shape)
+          f(x)
+          files_in_directory = sum(1 for f in os.listdir(tmpdir))
```
skye, line 187: I think you can just do `len(os.listdir(tmpdir))`

tests/compilation_cache_test.py
```diff
@@ -169,6 +174,42 @@ def test_put_executable(self):
...
+          self.assertEqual(files_in_directory, 1)
+          x = np.arange(prod(shape), dtype=np.int32).reshape(shape)
+          f(x)
+          files_in_directory = sum(1 for f in os.listdir(tmpdir))
+          self.assertEqual(files_in_directory, 2)
```
skye, line 192: Move your TODO about calling pmap more than once here -- I forgot about the in-memory cache, so I got to the end of this function was like, why aren't we calling the same pmap to test that the cache works?? You can lso remove the bit about different processes, since I think there are ways to eventually test this in the same process.

tests/compilation_cache_test.py
```diff
@@ -169,6 +174,42 @@ def test_put_executable(self):
...
+
+  def test_if_executable_exists(self):
```
skye, line 194: I think you can remove this test actually. You get enough coverage with the pmap test.

skye (changes requested):

tests/compilation_cache_test.py
```diff
@@ -179,6 +181,22 @@ def test_put_executable(self):
...
+          shape = (jax.lib.xla_bridge.device_count(), 4)
+          x = np.arange(prod(shape), dtype=np.float32).reshape(shape)
+          f(x)
+          files_in_directory = len(os.listdir(tmpdir))
+          self.assertEqual(files_in_directory, 1)
+          x = np.arange(prod(shape), dtype=np.int32).reshape(shape)
+          f(x)
+          files_in_directory = len(os.listdir(tmpdir))
+          self.assertEqual(files_in_directory, 2)
+          #TODO: create a test for calling pmap more than once
```
skye, line 198: nit: say "for calling pmap with the same input more than once", and specify that we can't test this now due to the in-memory cache.

tests/compilation_cache_test.py
```diff
@@ -179,6 +181,22 @@ def test_put_executable(self):
           actual = jax.lib.xla_client.execute_with_python_values(deserialized_executable, inputs_to_executable, backend)
           self.assertEqual(expected, actual)
 
+  def test_pmap(self):
+      cc._cache = None
+      with tempfile.TemporaryDirectory() as tmpdir:
+          cc.initialize_cache(tmpdir)
+          f = pmap(lambda x: x - lax.psum(x, 'i'), axis_name='i')
+          shape = (jax.lib.xla_bridge.device_count(), 4)
+          x = np.arange(prod(shape), dtype=np.float32).reshape(shape)
```
skye, line 190: Similar to the integration test, you can use a simpler input here: 
`x = np.arange(jax.device_count(), dtype=np.int64)`
(and then do same thing except with int32 for the second one)

skye (approved):
Looks good! This is ready for @zhangqiaorjc to look at.

zhangqiaorjc (approved):

PR #7280, opened 2021-07-14
Add source location and expression to error messages for CUDA API calls.

tomhennigan:
Before:

    jax._src.traceback_util.UnfilteredStackTrace: RuntimeError: operation not supported

After:

    jax._src.traceback_util.UnfilteredStackTrace: RuntimeError: third_party/py/jax/jaxlib/cusparse.cc:902: CUDA operation cudaMallocAsync(&buffer, bufferSize, stream) failed: operation not supported

### DIFF
--- a/jaxlib/BUILD
+++ b/jaxlib/BUILD
@@ -83,6 +83,7 @@ cc_library(
         "@com_google_absl//absl/base",
         "@com_google_absl//absl/memory",
         "@com_google_absl//absl/strings",
+        "@com_google_absl//absl/strings:str_format",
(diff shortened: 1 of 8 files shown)


apaszke (approved):
Looks good to me, although I can't claim any ownership over those files, so might be worth asking @hawkinsp if the overall plan makes sense. Also the tests seem to be segfaulting?

hawkinsp (approved):

hawkinsp:
The segfaults are unrelated to this PR.

Issue #7010, opened 2021-06-17
Make the default matmul precision float32 even on TPUs


shoyer:
Follow-up from https://github.com/google/jax/issues/2161:

The default low precision is a bit of a footgun, at least when doing anything that isn't implementing a neural net layer. In my opinion, it would be much safer to use "highest" precision by default (which isn't that much slower) on float32 data. Neural net libraries, of course, can default to lower precision, so this really only effects users who directly use NumPy APIs or the @ infix operator.

j-towns:
I would argue this is also a footgun for some neural net use cases. Both of the two times that I’ve carefully re-implemented a model in JAX, I’ve found that performance is worse than expected with the default precision and it took me some time to realise that precision was the cause.

As a (temporary) improvement on the current situation, we could at least add some information on this issue to point 5 of the ‘Current Gotchas’ in the readme.

shoyer:
https://github.com/google/jax/pull/6143 added a config flag, so this should in principle be easier to change now.

j-towns:
As @juliuskunze just pointed out to me, this is the only difference in semantics (that we can think of) between the GPU, CPU and TPU backends. 'Backend transparency' is an valuable property in mine and Julius's opinion, and given how close JAX is to achieving it (if this really is the only case where semantics differ significantly), it's surely worth changing the TPU default for all matmul-like ops (including conv) to closely approximate GPU and CPU behaviour. I understand this will harm 'default' speed, but I think we can mitigate that by making clear the availability of the bfloat16 option.

jonbarron:
My team shot itself in the foot last week for the ~fourth time due to matmult defaulting to bfloat16. This issue continues to be my biggest/only grievance with JAX.

mattjj:
Just got another +1 for this issue. Another foot lost.

inoryy:
+1 to setting default to f32. 

Having had the pleasure of shooting both my feet with either performance gotchas and numerics gotchas, I'd wholeheartedly prefer debugging the performance ones as they're much easier to spot.

ppwwyyxx:
Similar discussion in pytorch: https://github.com/pytorch/pytorch/issues/67384. Pytorch once enabled tfloat32 by default for a few ops, and then had to revert the decision due to similar complains. Enabling bfloat16 by default is presumably be even worse.

nouiz:
[...]

PR #7304, opened 2021-07-16
Add TPU precision details to README gotchas

j-towns:
See https://github.com/google/jax/issues/7010#issuecomment-865076339.


hawkinsp (approved):
Fair enough. I think we should just switch the default sometime soon...

PR #7306, opened 2021-07-16
Avoid sharing handles across streams.

tomhennigan:
When running across 8xV100 GPUs we observed the following error:

    libc++abi: terminating with uncaught exception of type std::runtime_error: third_party/py/jax/jaxlib/cusolver.cc:171: operation cusolverDnSpotrf(handle.get(), d.uplo, d.n, a, d.n, static_cast<float*>(workspace), d.lwork, info) failed: cuSolver execution failed

I cannot find documentation to this effect, but I believe that it is unsafe to share cuSolver handles across streams, since keeping the handle pool stream local does solve the issue.

### DIFF
--- a/jaxlib/cublas.cc
+++ b/jaxlib/cublas.cc
@@ -45,16 +45,16 @@ template <>
   BlasHandlePool* pool = Instance();
   absl::MutexLock lock(&pool->mu_);
   cublasHandle_t handle;
-  if (pool->handles_.empty()) {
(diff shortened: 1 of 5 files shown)


hawkinsp:
As best I can tell the code is correct and this is NVidia's bug. Can you make some sort of small test case that reproduces this from JAX? I'd like to follow up with NVidia about this.

We could merge this as a temporary fix I guess, but I'm a little unhappy with it.

PR #7317, opened 2021-07-18
[jax2tf] Expand shape polymorphism support to use dimension polynomials as values

gnecula:
The goal of this change is to support shape polymorphism for operations
such as average (which needs to divide by the size of a dimension) or
indexing (which needs to normalize indices by comparing them with 0 and
adding dimension size for negative indices). In both of these cases
the size of a dimenion needs to be used as a value in the array
computation. In general, the size of a dimension is used only to
customize primitives.

This change introduces `core.dim_as_value` which must be used on
a dimension size before using it as a value in the array computation.
E.g.,

```
def average(x):
   return jnp.sum(x, axis=0) / core.dim_as_value(x.shape[0])
```

This function is the identity function if the dimension size is
constant, otherwise it uses a new primitive `shape_poly.dim_as_value_p`.

Note that this does not change fundamentally the flavor of shape
polymorphism supported in jax2tf: intermediate shapes and their values
may depend on the input shapes, but never does a shape depend on the
input values. In fact, one could have expressed the `dim_as_value`
already:

```
def dim_as_value(d):
   jnp.sum(jnp.broadcast_to(jnp.array(1), shape=(d,)))
```

We were able to suppot `jnp.mean` and `jnp.take` by using
`core.dim_as_value` internally, but to fully roll-up the solution
we need to make `core.dim_as_value` a public API and teach the
users how to use it when they want to use shape polymorphism.
Alternatively, perhaps there is a way to automatically convert
dimension polynomials to values when passed to the lax primitives.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -1283,6 +1283,9 @@ class DimensionHandler:
     """Implements `0 if d == 0 else 1 + dilation * (d - 1))`"""
     return 0 if d == 0 else 1 + dilation * (d - 1)
 
+  def as_value(self, d: DimSize):
+    """Turns a dimension size into a JAX value that we can compute with."""
+    return d
 
 _dimension_handler_int = DimensionHandler()
 _SPECIAL_DIMENSION_HANDLERS: Dict[type, DimensionHandler] = {}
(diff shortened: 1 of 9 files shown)


jax/core.py
```diff
@@ -1283,6 +1283,9 @@ def dilate(self, d: DimSize, dilation: int) -> DimSize:
     """Implements `0 if d == 0 else 1 + dilation * (d - 1))`"""
     return 0 if d == 0 else 1 + dilation * (d - 1)
 
+  def as_value(self, d: DimSize):
+    """Turns a dimension size into a Jax value that we can compute with."""
```
mattjj, line 1287:
```suggestion
    """Turns a dimension size into a JAX value that we can compute with."""
```

jax/core.py
```diff
@@ -1378,6 +1381,11 @@ def stride_shape(s: Shape, window_size: Shape, window_stride: Shape) -> Shape:
   """(s - window_size) // window_stride + 1"""
   return tuple(map(stride_dim, s, window_size, window_stride))
 
+def dim_as_value(d: DimSize):
+  """Turns a dimension size into a Jax value that we can compute with.
```
mattjj, line 1385:
```suggestion
  """Turns a dimension size into a JAX value that we can compute with.
```

mattjj (approved):
LGTM!

PR #7330, opened 2021-07-19
jit integration with persistent compilation cache

colemanliyah:


skye (approved):
Nice! Let's have @zhangqiaorjc take a look too

jax/interpreters/xla.py
```diff
@@ -61,6 +61,21 @@
...
 
+
+def compile_or_get_cached(backend, computation, compile_options):
+    # Avoid import cycle between jax and jax.experimental
+    from jax.experimental.compilation_cache import compilation_cache as cc
+    if cc.is_initialized():
+        cached_executable = cc.get_executable(computation, compile_options)
+        if cached_executable is not None:
+            return cached_executable
+        else:
```
zhangqiaorjc, line 72: nit: you can remove the two `else` here and below to make the code more compact and have less indent
```
if A:
  return B
else:
  return C
```
is shorter with
```
if A:
  return B
return C
```

zhangqiaorjc (approved):

jax/interpreters/xla.py
skye: Ha, I think I explicitly requested the second `else` because I think it makes it easier to read (makes it clear that the two paths are separate without having to follow the control flow of the first path, plus the `else` case is the exceptional case so I personally don't like using fall-through there). Maybe as a compromise we can remove this `else` and keep the second one :)

PR #7341, opened 2021-07-21
Improve wording for multi-process docs

PSilvestre:
Currently, there is no direct implication that device-to-device links are a requirement.

Though it is a very small change, it would have saved me a couple of hours in trying to leverage NCCL's ability to create rings over ethernet to run JAX.

I'm sure others will also appreciate this.

### DIFF
--- a/docs/multi_process.md
+++ b/docs/multi_process.md
@@ -12,8 +12,8 @@ operations (e.g. {func}`jax.lax.psum`) in multi-process settings, although other
 communication methods may be useful too depending on your use case (e.g. RPC,
 [mpi4jax](https://github.com/mpi4jax/mpi4jax)). If you’re not already familiar
 with JAX’s collective operations, we recommend starting with the
-{doc}`/jax-101/06-parallelism` notebook. An important feature of multi-process
-environments is direct communication links between accelerators, e.g. the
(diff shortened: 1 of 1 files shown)


PSilvestre:
@googlebot I signed it!

skye (approved):

skye:
This change LGTM, thanks for the contribution and sorry for the confusion!

However, it may be possible to use NCCL over ethernet with JAX -- my understanding is that JAX doesn't care if devices communicate over ethernet or dedicated links, as long as there's some kind of protocol that manages the communication and can be targeted by JAX's backend. I think the trick with NCCL is initializing it correctly. @hawkinsp can you say more?

PSilvestre:
@skye If that is the case, I would be very interested in having documentation covering this. I could help with that.

I also wonder if there is a way to retrigger the build of the documentation (which seems to have failed due to timeout?).

hawkinsp:
JAX *does* support cross-host (i.e., ethernet or infiniband) GPU communication via NCCL already. However, we don't advertise the feature yet or provide a public API for it.

skye:
The docs build was unrelatedly broken when you created your PR (not your fault!). It's fixed now though, and your change is live: https://jax.readthedocs.io/en/latest/multi_process.html

PR #7343, opened 2021-07-21
Add shape and dtype of leaked tracer to UnexpectedTracerError.

LenaMartens:
Adding some extra info which helped me track down a leaked tracer. Looks like:
```
...
The tracer that caused this error was created on line (...)
The tracer has shape (4, 3) and dtype int32.
...
```


hawkinsp (approved):

hawkinsp:
Looks like flake8 is unhappy: you may need to quote some characters in your regex or use a raw string.

PR #7345, opened 2021-07-21
Add tracers to LeakChecker error, and filter out false positives this way

### LINKED ISSUE #5636: Spurious leak checker error with custom JVP
I'd love to be able to use the (great-looking) new leaked tracer tools, but I'm running into a problem. After I set
```
export JAX_CHECK_TRACER_LEAKS=1
```
and then run:
```python
import jax

@jax.custom_jvp
def t(y):
    return y

def t_jvp(primals, tangents):
    assert False

t.defjvp(t_jvp)

@jax.jit
def s(y):
    return t(y)

s(0.0)
```
I unfortunately get
```
Exception: Leaked trace MainTrace(0,DynamicJaxprTrace)
```
There's possibly a deeper bug here because I'm getting leaked tracers in my code wtihout the flag, so I'm hoping this is a spurious error.

LenaMartens:
Currently when using `jax.checking_leaks()`, the error message looks like: `Leaked sublevel 1.` or `Leaked trace MainTrace(1,DynamicJaxprTrace)`. This tells you the level of the leak, but nothing about the Tracer which caused the leak. 
To get to the leaked Tracer, we can use `gc.get_referrers`. This relies on the assumption that each `Tracer` keeps a reference to its `Trace`, which keeps a reference to the `MainTrace`. 
First we find the `traces` which hold references to the `MainTrace` (or `SubLevel`), then we find the `tracers` which hold references to those `traces`. 

I propose that if we can't find any tracers this way, we don't throw an error, which fixes false positives caused by the JAX internals **while also not introducing any false negatives**. The claim I'm making is that user code will only ever have a handle on `Tracer` objects, and never directly on `Trace` objects. So the only possible way for `MainTrace` leak to be the fault of the user is through a `Tracer->Trace->MainTrace` chain (keeping the `MainTrace` alive by keeping a `Tracer` alive). A chain like `Trace->MainTrace` can only be the fault of JAX (or libraries using JAX internals which do have a handle on `Trace` objects).

For example: as mentioned in #5636, when using a custom_jvp, we get a false positive in the leak checker because of some lambdas in the JAX internals closing over the `MainTrace`.  The reference chain in this case will not contain a Tracer, but rather a python cell object which is holding onto the `MainTrace`. So by skipping the error we remove that false positive.

Another approach I was considering is adding a `WeakSet` global variable which keeps track of _all_ tracers which are created throughout the program (if config.checking_leaks is true), and checking that list for any tracers which refer to the MainTrace we care about. That might be cleaner than relying on `gc.get_referrers` (but more memory intensive?).

Fixes #5636.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -18,6 +18,7 @@ from operator import attrgetter
 from contextlib import contextmanager
 from collections import namedtuple
 from functools import total_ordering
+import gc
 import itertools as it
 from weakref import ref
 import threading
@@ -736,6 +737,17 @@ def reset_trace_state() -> bool:
 def cur_sublevel() -> Sublevel:
   return thread_local_state.trace_state.substack[-1]
 
+def maybe_find_leaked_tracers(x: Optional[Union[MainTrace, Sublevel]]):
+  """Find the leaked tracers holding a reference to the MainTrace or SubLevel.
+
+  It's possible there's none! eg. there's some cases where JAX itself holds a
+  reference to `x` inside of a lambda closure, and no tracers were leaked
+  by the user. In this case an empty list is returned.
+  """
+  traces = list(filter(lambda x: isinstance(x, Trace), gc.get_referrers(x)))
+  tracers = list(filter(lambda x: isinstance(x, Tracer), gc.get_referrers(*traces)))
+  return tracers
+
 @contextmanager
 def new_main(trace_type: Type[Trace],
              dynamic: bool = False,
@@ -761,7 +773,9 @@ def new_main(trace_type: Type[Trace],
     t = ref(main)
     del main
     if t() is not None:
-      raise Exception(f'Leaked trace {t()}')
+      leaked_tracers = maybe_find_leaked_tracers(t())
+      if leaked_tracers:
+        raise Exception(f'Leaked level {t()}. Leaked tracer(s): {leaked_tracers}.')
 
 @contextmanager
 def new_base_main(trace_type: Type[Trace]) -> Generator[MainTrace, None, None]:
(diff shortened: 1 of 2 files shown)


mattjj:
Could you mention in the PR message the idea you explained to me that user code only gets direct references to `Tracer` instances, and so only have indirect references to `Trace`s, i.e. only JAX-internal code (and libraries using JAX internals) get direct references to `Trace`s? That seems like solid reasoning for why we can ignore `Trace`s that are still alive if there are no `Tracer`s pointing to them.

mattjj (approved):
Brilliant! Thanks for explaining this.

Also, very practical! This will make the leak checker actually useful.

jax/interpreters/partial_eval.py
```diff
@@ -1169,6 +1169,9 @@ def post_process_custom_vjp_call(self, out_tracers, params):
     assert False  # unreachable
 
 def _memoize(thunk):
+  if config.jax_check_tracer_leaks:
```
mattjj, line 1172: Actually, I'm concerned about this change. It seems like it might lead to infinite recursions when leak checking is on. Is it necessary for this PR?

mattjj: Hrm or maybe this memoization only has to do with efficiency, not whether there are infinite recursions. I forget! I'll try a test...

mattjj: Yeah, basically my concern is that I think `python tests/api_test.py CustomJVPTest` no longer passes when we make this function return `thunk`.

mattjj: I think this is a repro, tested against this branch:

```python
from jax import custom_jvp, jvp, lax, check_tracer_leaks

@custom_jvp
def f(x):
  return x

@f.defjvp
def f_jvp(primals, tangents):
  (x,), (xdot,) = primals, tangents
  return f(x), xdot

with check_tracer_leaks(True):
  jvp(lambda x: lax.cond(True, f, f, x), (3.,), (1.,))
```

But if we delete these two lines, the above repro still works and the new test you added passes too.

LenaMartens: sorry, this was a left-over from me trying to fix the actual false positive leak in `custom_jvp` (described here: https://github.com/google/jax/issues/5636#issuecomment-774290760). I gave up on fixing that one (couldn't figure out how to remove the main trace from the `jvp_jaxpr_thunk` closure), and I thought it couldn't hurt to include.
I'll remove it!

LenaMartens:
Thanks for the review! I've removed the changes to `_memoize` and updated the PR description.

PR #7347, opened 2021-07-21
Remove custom __hash__ implementation

zafarali:
Currently Jax arrays pass the check isinstance(x, collections.Hashable) despite not being hashable.  The [python datamodel](https://docs.python.org/3/reference/datamodel.html#object.__hash__) says that 

> A class that overrides __eq__() and does not define __hash__() will have its __hash__() implicitly set to None. When the __hash__() method of a class is None, instances of the class will raise an appropriate TypeError when a program attempts to retrieve their hash value, and will also be correctly identified as unhashable when checking isinstance(obj, collections.abc.Hashable).

This PR removes the custom implementation of __hash__.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2077,10 +2077,7 @@ class APITest(jtu.JaxTestCase):
     rep = jnp.ones(()) + 1.
     self.assertIsInstance(rep, jax.interpreters.xla.DeviceArray)
     msg = "JAX DeviceArray, like numpy.ndarray, is not hashable."
-    with self.assertRaisesRegex(TypeError, msg):
-      hash(rep)
-    with self.assertRaisesRegex(TypeError, msg):
-      hash(rep.device_buffer)
+    self.assertNotIsInstance(rep, collections.Hashable)
 
   def test_grad_without_enough_args_error_message(self):
     # https://github.com/google/jax/issues/1696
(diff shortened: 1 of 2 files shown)


tests/api_test.py
```diff
@@ -2076,11 +2076,7 @@ def test_device_array_repr(self):
   def test_device_array_hash(self):
     rep = jnp.ones(()) + 1.
     self.assertIsInstance(rep, jax.interpreters.xla.DeviceArray)
-    msg = "JAX DeviceArray, like numpy.ndarray, is not hashable."
-    with self.assertRaisesRegex(TypeError, msg):
-      hash(rep)
```
mattjj, line 2081: Can you keep this line, but with a different `msg` string? That way we test both that `hash(rep)` fails and that `isinstance(rep, Hashable)` is False.

mattjj (approved):
LGTM, one suggestion!

mattjj:
I think the readthedocs failure is a pre-existing issue.

shoyer:
LGTM, though perhaps explicitly setting `__hash__ = None` would be clearer

zafarali:
Had to eventually set to None as @shoyer suggested. I think this is because DeviceArray is actually an empty class and this piece of code is adding attributes after the class is defined.

PR #7350, opened 2021-07-22
fix xla.lower_fun axis env issue

mattjj:
We weren't using the translation-time axis env to populate the tracing-time axis env when calling `xla.lower_fun`, as done in the all_gather translation on non-gpu backends.

Usually we think of translation-to-XLA as happening after all the tracing is done. But actually it's convenient to write translation rules by writing out some Python which calls into `lax`/`jnp` and then tracing it, rather than writing explicit calls into the XLA Builder API. So our translation rules sometimes turn around and do more tracing, using the `xla.lower_fun` utility!

That was the setting for this issue. When we define axis names, we have a (global) trace-time data structure which tracks their sizes, which is how e.g. axis size data is available at trace time (e.g. via `psum(1, axis_name)`, which evaluates to a regular Python integer that you can print and branch on). We also need that information at translation-to-xla time, so there's a separate (local) translation-time environment.

But when our translation-to-XLA process jumps back into tracing Python using `xla.lower_fun`, that means we again might use the tracing-time axis environment. Hence we need the translation-time axis environment to populate the tracing-time axis environment! That part was missing, but was a one-liner fix.

The test uses `all_gather` which on non-GPU backends calls into `xla.lower_fun`.

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -1021,7 +1021,8 @@ def lower_fun(fun, multiple_results, parallel=False, with_avals=False, backend=N
     wrapped_fun = lu.wrap_init(fun, params)
     if not multiple_results:
       wrapped_fun = _tuple_output(wrapped_fun)
-    jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, avals)
+    with core.extend_axis_env_nd(zip(axis_env.names, axis_env.sizes)):
+      jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, avals)
     outs = jaxpr_subcomp(c, jaxpr, backend, axis_env, _xla_consts(c, consts), '',
                          *xla_args)
     if multiple_results or any(v.aval._num_buffers > 1 for v in jaxpr.outvars):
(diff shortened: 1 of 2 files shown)


jax/interpreters/xla.py
```diff
@@ -1021,7 +1021,8 @@ def f_with_avals(c, avals, xla_args, params):
     wrapped_fun = lu.wrap_init(fun, params)
     if not multiple_results:
       wrapped_fun = _tuple_output(wrapped_fun)
-    jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, avals)
+    with core.extend_axis_env_nd(zip(axis_env.names, axis_env.sizes)):
```
apaszke, line 1024: Shouldn't we extend the env in `api.xla_computation` instead?

mattjj: Good question. Thanks for checking this.

[We already extend the env while forming the jaxpr](https://github.com/google/jax/blob/fe35680d1844f063d396e3e376c3dad500b922d7/jax/_src/api.py#L683-L686), and there's no need to populate the env [when doing jaxpr-to-hlo lowering](https://github.com/google/jax/blob/fe35680d1844f063d396e3e376c3dad500b922d7/jax/_src/api.py#L699) (since we're no longer tracing / calling bind, so nothing should need the trace-time env...) _unless_ we call into `xla.lower_fun`. For that reason I think this is the right place to extend the axis env.

WDYT?

mattjj: I'll copy some explanatory comments I made in JAXers chat into the PR message above.

mattjj: (I speculatively merged, but if we fiugre out that this is the wrong approach I will revise it!)

PR #7362, opened 2021-07-23
Add pure JAX implementation of tridiagonal_solve for non-GPU platforms.

tomhennigan:
Implementation by @dpfau and team.


mattjj:
Awesome!

I think we might be able to make this even better (at least in some regimes) using `associative_scan`, as per [Blelloch's Prefix Sums and Their Applications](https://www.cs.cmu.edu/~guyb/papers/Ble93.pdf). I long ago promised @dpfau that I'd make an attempt at this, since in principle I've written similar code before (for Gaussian graphical models), but I still haven't gotten around to it. I just wanted to mention it here so we don't forget!

mattjj (approved):
Very clean, thanks for submitting this!

Issue #7340, opened 2021-07-20
concrete_aval cannot handle ad_util.Zero causing crash in id_tap


NeilGirdhar:
Sorry for not having a MWE.  I spent about half an hour on trying to build one, but I'll need more time to figure it out if it's necessary.  In the off chance that a stack trace is enough, here's what I have:

```
Traceback (most recent call last):
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/bin/cmm_demo", line 5, in <module>
    cli()
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/neil/.cache/pypoetry/virtualenvs/cmm-tspD8tmv-py3.9/lib/python3.9/site-packages/typer/main.py", line 497, in wrapper
    return callback(**use_params)  # type: ignore
  File "/home/neil/src/cmm/cmm/demo/geometry.py", line 127, in geometry
    run_demo(mode, demo)
  File "/home/neil/src/cmm/cmm/visualization/run.py", line 22, in run_demo
    show_gui(demo)
  File "/home/neil/src/cmm/cmm/visualization/gui.py", line 64, in show_gui
[...]

PR #7366, opened 2021-07-23
Fix crash in concretization of ad_util.Zero

NeilGirdhar:


NeilGirdhar:
I wasn't sure how to resolve the circular dependency.  Please advise if the non-top-level import is okay.

jax/_src/ad_util.py
```diff
@@ -75,8 +75,11 @@ def __repr__(self):
   @staticmethod
   def from_value(val):
     return Zero(raise_to_shaped(get_aval(val)))
+  def to_concrete(self):
+    return zeros_like_aval(self.aval)
 
 register_pytree_node(Zero, lambda z: ((), z.aval), lambda aval, _: Zero(aval))
+core.pytype_aval_mappings[Zero] = lambda z: z.to_concrete()
```
gnecula, line 82: This does not look right.  `pytype_aval_mappings` should contain functions that return `core.AbstractValue`. 

I think it would be helpful to come up with a JAX test case that demonstrates this. The test case in #7340 is too complicated.

NeilGirdhar: > This does not look right. pytype_aval_mappings should contain functions that return core.AbstractValue

Oh, I see.  How do we do that from an `aval`?  Do we just return `z.aval`?

> I think it would be helpful to come up with a JAX test case that demonstrates this.

I totally agree.  The issue is that it's incredibly time consuming for me to create MWEs since I have to basically strip out chunks of the giant program until I have one.  Is it possible to look at the tracer types and know how to produce such tracers with a minimal program?

gnecula:
I propose that we close this PR. I think that the fix is in #7378

PR #7369, opened 2021-07-24
refine constant-hoisting heuristic in closure_convert

### LINKED ISSUE #6415: Hiding constants from closure_convert and the AD system
Use case: Higher order functions (like ODE solvers and optimizers) with custom VJP rules.

Consider this example form the JAX documentation for `closure_convert`:

```
def minimize(objective_fn, x0):
  converted_fn, aux_args = closure_convert(objective_fn, x0)
  return _minimize(converted_fn, x0, *aux_args)

@partial(custom_vjp, nondiff_argnums=(0,))
def _minimize(objective_fn, x0, *args):
  z = objective_fn(x0, *args)
  # ... find minimizer x_opt ...
  return x_opt

def fwd(objective_fn, x0, *args):
  y = _minimize(objective_fn, x0, *args)
  return y, (y, args)

def rev(objective_fn, res, g):
  y, args = res
  y_bar = g
  # ... custom reverse-mode AD ...
  return x0_bar, *args_bars

_minimize.defvjp(fwd, rev)
```

Imagine now that `objective_fn` depends on a large matrix of constants `c`,  e.g. captured via closure. Then `closure_convert` pulls this matrix out and makes it an explicit argument of `converted_fn`.
During the reverse pass, `c_bar` has to be calculated as part of `args_bar`. This computation can be expensive and complex but it is not needed if differentiation with respect to `c` is never required or does not make sense.
There should be a way for the author of `objective_fn` to annotate `c` accordingly. This cannot be done at the level of the minimizer, where it is not known which `args` are needed in the VJP and which ones are not.

In simple cases, it is conceivable that XLA manages to optimize out the unused `c_bar`. That does not work for more complex cases, e.g. [the backward pass of an ODE integrator that has to integrate an augmented system of ODEs](https://github.com/google/jax/blob/3ac809ede30d961b873ca62f9828090709ced6be/jax/experimental/ode.py#L221).

In our concrete application (ODE integration where the dynamics function captures a large matrix), the impact of calculating `c_bar` is such that XLA consumes >44 GB of RAM during the compilation stage while attempting to generate code for the Jacobian of the dynamics function with respect to each and every element of `c`.

Can `closure_convert` be modified to not extract certain captured variables? At the moment, it ignores integer arrays, for example.
I also see parallels to https://github.com/google/jax/issues/5913, which is more focused on the JVP case.

froystig:
Instead of hoisting all float-type arrays during closure conversion, only hoist values of type `JVPTracer` (or tracers carrying such tracers indirectly). Doing so better approximates the subset of closure-captured values that participate in automatic differentiation.

Potentially fixes #6415

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -19,7 +19,6 @@ import operator as op
 from typing import Callable, Generic, Optional, Sequence, Tuple, TypeVar, Any
 
 from jax import core
-from jax._src import dtypes
 from jax import linear_util as lu
 from jax.tree_util import (tree_flatten, tree_unflatten, tree_map,
                         tree_multimap, treedef_is_leaf, treedef_tuple,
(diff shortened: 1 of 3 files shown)


jax/_src/custom_derivatives.py
```diff
@@ -831,18 +830,21 @@ def rev(objective_fn, res, g):
   else:
     return _closure_convert_for_avals(fun, in_tree, in_avals)
 
+def _is_perturbed(x: Any) -> bool:
+  if isinstance(x, (pe.JaxprTracer, ad.JVPTracer)):
```
mattjj, line 834:
```suggestion
  if isinstance(x, ad.JVPTracer):
```

If the tests pass with just this, it's... cleaner.

tests/api_test.py
```diff
@@ -5211,40 +5211,81 @@ def f(x):
...
     g_c, g_x = api.grad(solve, argnums=(0, 1))(c, x)
-    self.assertAllClose(g_c, 42. * jnp.ones(2), check_dtypes=False)
-    self.assertAllClose(g_x, 17. * jnp.ones(2), check_dtypes=False)
+    self.assertAllClose(g_c, 42. * c, check_dtypes=False)
+    self.assertAllClose(g_x, 17. * x, check_dtypes=False)
+
+  def test_closure_convert_mixed_consts(self):
+    # like test_closure_convert, but close over values that participate
+    # in AD as well as values that do not.
+
```
mattjj, line 5253:
```suggestion
# See https://github.com/google/jax/issues/6415

```

mattjj (approved):
Nice! ~Can we check if it fixes #6415~ Nevermind, #6415 doesn't have a repro on it.

PR #7372, opened 2021-07-24
Add scipy.special.erfcx function

### LINKED ISSUE #1987: Support erfcx
Jax is missing support for scipy.special.erfcx (https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.erfcx.html). Please add this function.

k4tta:
Add scipy.special.erfcx

Scaled complementary error function, `erfcx(x) = exp(x**2) * erfc(x)`.
Implementation is based on [this](https://stackoverflow.com/a/39777361) stackoverflow answer.

Resolves: #1987

### DIFF
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -87,6 +87,7 @@ jax.scipy.special
    entr
    erf
    erfc
+   erfcx
    erfinv
    expit
    gammainc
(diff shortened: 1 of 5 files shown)


obkyrush:
@googlebot I signed it!

jakevdp:
Hi - thanks for the contribution. Unfortunately, I don't think we can accept it: StackOverflow code is released under a CC-BY-SA license, which is not compatible with the Apache 2.0 license used by the JAX repository. If you're able to find a similar algorithm with a compatible license (such as BSD, MIT, etc.) then please feel free to open another pull request. Thanks, and sorry we're not able to include this.

PR #7386, opened 2021-07-26
Replaced Any with concrete types for APIs in jaxlib.xla_client

superbobry:
`jaxlib.xla_client` is a native extension module and is therefore opaque to mypy. To overcome that JAX used `Any` for classes defined in `jaxlib.xla_client`. This PR replaces `Any` with concrete types and updates/corrects existing usages to make mypy happy.

### DIFF
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -15,7 +15,8 @@
 import functools
 import collections
 import operator as op
-from typing import Any, Callable, Optional, Sequence, Tuple, Type, TypeVar, overload
+from typing import (Any, Callable, Hashable, Iterable, Optional, Tuple, Type,
+                    TypeVar, overload, TYPE_CHECKING)
 
 from ..lib import pytree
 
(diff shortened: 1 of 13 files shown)


jax/_src/tree_util.py
```diff
@@ -27,6 +28,12 @@
 T = TypeVar("T")
 U = TypeVar("U")
 
+if TYPE_CHECKING:
```
hawkinsp, line 31: Any reason not to do this unconditionally?

hawkinsp (approved):

jax/_src/tree_util.py
superbobry: Yep, `PyTreeDef` is not exported by pybind11. I think this is because because it has no exported `__init__`.

```
>>> from jaxlib.xla_extension import pytree
>>> dir(pytree)
['__doc__', '__loader__', '__name__', '__package__', '__spec__', 'all_leaves', 'flatten', 'register_node', 'tuple']
```

hawkinsp: Seems like something we should fix, right?

hawkinsp:
mypy is now sad.

superbobry:
Yeah, we will need a new jaxlib release to make mypy happy again :)

jax/_src/tree_util.py
superbobry: Sure, I'll look into this in a separate PR.

LenaMartens (approved):

jax/_src/tree_util.py
jakevdp: Addressing this in #11455

PR #7399, opened 2021-07-28
Log a warning after 60 secs to remind the user to run code on all hos…

yashk2810:
…ts for Cloud TPU 1VM

### DIFF
--- a/jax/lib/xla_bridge.py
+++ b/jax/lib/xla_bridge.py
@@ -147,7 +147,7 @@ def _make_tpu_driver_client():
 
 def tpu_client_timer_callback(timer: float):
   def _log_warning():
-    logging.warning('Did you run your code on all the hosts?')
+    warnings.warn('Did you run your code on all the hosts?')
 
   # Will log a warning after `timer` secs.
   t = threading.Timer(timer, _log_warning)
--- a/jax/lib/xla_bridge.py
+++ b/jax/lib/xla_bridge.py
@@ -145,12 +145,15 @@ def _make_tpu_driver_client():
   return tpu_driver_client.TpuBackend.create(worker=FLAGS.jax_backend_target)
 
 
-def tpu_client_timer_callback(timer: float):
+def tpu_client_timer_callback(timer_secs: float):
   def _log_warning():
-    warnings.warn('Did you run your code on all the hosts?')
+    warnings.warn(
+      (f'TPU backend initialization is taking more than {timer_secs} seconds. '
+      'Did you run your code on all TPU hosts? '
+      'See https://jax.readthedocs.io/en/latest/multi_process.html for more information.'))
 
-  # Will log a warning after `timer` secs.
-  t = threading.Timer(timer, _log_warning)
+  # Will log a warning after `timer_secs`.
+  t = threading.Timer(timer_secs, _log_warning)
   t.start()
 
   try:
--- a/jax/lib/xla_bridge.py
+++ b/jax/lib/xla_bridge.py
@@ -145,6 +145,22 @@ def _make_tpu_driver_client():
   return tpu_driver_client.TpuBackend.create(worker=FLAGS.jax_backend_target)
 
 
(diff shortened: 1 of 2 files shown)


jax/lib/xla_bridge.py
```diff
@@ -145,6 +145,22 @@ def _make_tpu_driver_client():
   return tpu_driver_client.TpuBackend.create(worker=FLAGS.jax_backend_target)
 
 
+def tpu_client_timer_callback(timer: float):
+  def _log_warning():
+    warnings.warn('Did you run your code on all the hosts?')
```
skye, line 150: Let's say something about the TPU runtime taking a while to start, and I also think we should link to https://jax.readthedocs.io/en/latest/multi_process.html. Here's my suggestion, feel free to reword:
`TPU backend initialization is taking more than {timer} seconds. Did you run your code on all TPU hosts? See https://jax.readthedocs.io/en/latest/multi_process.html for more information.`

jax/lib/xla_bridge.py
```diff
@@ -145,6 +145,22 @@ def _make_tpu_driver_client():
   return tpu_driver_client.TpuBackend.create(worker=FLAGS.jax_backend_target)
 
 
+def tpu_client_timer_callback(timer: float):
```
skye, line 148: nit: rename `timer` to `timer_secs` or `warning_secs` or something.

tests/xla_bridge_test.py
```diff
@@ -56,6 +64,15 @@ def test_local_devices(self):
     with self.assertRaisesRegex(RuntimeError, "Unknown backend foo"):
       xb.local_devices(backend="foo")
 
+  @mock.patch('jax.lib.xla_client.make_tpu_client', side_effect=mock_tpu_client)
+  def test_timer_tpu_warning_1vm(self, _):
+    with warnings.catch_warnings(record=True) as w:
+      warnings.simplefilter('always')
+      xb.tpu_client_timer_callback(0.01)
```
skye, line 71: Maybe put a sleep here just in case? Not sure if this is necessary.

tests/xla_bridge_test.py
```diff
@@ -56,6 +64,15 @@ def test_local_devices(self):
     with self.assertRaisesRegex(RuntimeError, "Unknown backend foo"):
       xb.local_devices(backend="foo")
 
+  @mock.patch('jax.lib.xla_client.make_tpu_client', side_effect=mock_tpu_client)
```
skye, line 67: I think this is the first use of mocks in jax, so cool! :D

tests/xla_bridge_test.py
```diff
@@ -56,6 +64,15 @@ def test_local_devices(self):
...
+  def test_timer_tpu_warning_1vm(self, _):
+    with warnings.catch_warnings(record=True) as w:
+      warnings.simplefilter('always')
+      xb.tpu_client_timer_callback(0.01)
+      msg = str(w[-1].message)
+      self.assertLen(w, 1)
```
skye, line 73: nit: put this above the `w[-1]` access

skye (changes requested):
Beautiful!

tests/xla_bridge_test.py
yashk2810: Done

tests/xla_bridge_test.py
yashk2810: Its not necessary. The time is so small that it runs perfectly everytime. I tested it a lot of times.

tests/xla_bridge_test.py
yashk2810: Woohoo!

jax/lib/xla_bridge.py
yashk2810: Done

jax/lib/xla_bridge.py
yashk2810: Done

skye (approved):

PR #7409, opened 2021-07-29
Add some clarification to the async dispatch docs

yashk2810:


docs/async_dispatch.rst
```diff
@@ -10,6 +10,8 @@ program:
 >>> import jax.numpy as jnp
 >>> from jax import random
 >>> x = random.uniform(random.PRNGKey(0), (1000, 1000))
+>>> # Printing the result or repr(result) or str(result) will block until the
```
mattjj, line 13:
```suggestion
>>> # Printing the result (i.e. evaluating `repr(result)` or `str(result)`) will block until the
```

Maybe also tweak this to wrap at 80 character lines.

docs/async_dispatch.rst
```diff
@@ -10,6 +10,8 @@ program:
...
+>>> # value is ready. That's why the value is present in the output below.
```
mattjj, line 14:
```suggestion
>>> # value is ready.
```

I'm not sure but I suspect this second sentence isn't clear without more context.

mattjj (approved):
Thanks for the attention to detail! I suggested some tweaks; WDYT?

docs/async_dispatch.rst
yashk2810: Done

docs/async_dispatch.rst
yashk2810: Done

PR #7410, opened 2021-07-29
added debugging logs to compilation cache

colemanliyah:


jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -67,9 +69,13 @@ def get_cache_key(xla_computation, compile_options) -> str:
     """
     hash_obj = hashlib.sha256()
     hash_obj.update(xla_computation.as_serialized_hlo_module_proto())
+    logging.vlog(1, "get_cache_key function, hash after serializing computation")
```
skye, line 72: You should actually log the hash :) You can remove the "function" to make it a big shorter too. I think this should work:
```python
logging.vlog(1, f"get_cache_key hash after serializing computation: {hash_obj.digest().hex()}")
```

Also update the vlogs below.

skye (changes requested):
I forgot to mention earlier, you can enable the vlogs by adding `-v=1` to a python3 command that uses the cache. I suggest running your tests and making sure the vlogs are there and look good:
```
python3 tests/compilation_cache_test.py -v=1
```

skye (approved):
Looks great!

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -67,9 +69,13 @@ def get_cache_key(xla_computation, compile_options) -> str:
     """
     hash_obj = hashlib.sha256()
     hash_obj.update(xla_computation.as_serialized_hlo_module_proto())
+    logging.vlog(1, f"get_cache_key hash after serializing computation: {hash_obj.digest().hex()}")
```
zhangqiaorjc, line 72: Use the following idiom:

```
if logging.vlog_is_on(1):
  logging.vlog(1, blah)
```

Python evaluates the arguments eagerly and you don't want to calculate digest and hex when vlogging is *not* enabled

This is a common misuage of vlog, read
https://github.com/abseil/abseil-py/blob/master/absl/logging/__init__.py#L53-L57

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -28,6 +29,7 @@ def initialize_cache(path, max_cache_size_bytes=32 * 2**30):
     global _cache
     assert _cache == None, f"The cache path has already been initialized to {_cache}"
     _cache = FileSystemCache(path, max_cache_size_bytes)
+    logging.warning(f"Initialized compilation cache at {path}")
```
zhangqiaorjc, line 32: add "persistent"

zhangqiaorjc (approved):

PR #7417, opened 2021-07-30
Abstracts the evaluation of a single jaxpr equation

botev:
I have been working quite extensively with Jax internals as I need to write custom function transformations. As such one thing that always ends up just copy-pasting from `jax.core` is the evaluation of a single jaxpr equation. The code for that though seems non-trivial and I notice it does change over time. As such I think it would be very useful to have this as a separate function that is possible to call into.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -316,6 +316,26 @@ def traverse_jaxpr_params(f, params):
           if type(p) in (Jaxpr, ClosedJaxpr)}
 
 
+def eval_jaxpr_eqn(eqn, in_vals):
+  """Evaluates the jaxpr equation with the provided input values."""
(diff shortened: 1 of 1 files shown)


jax/core.py
```diff
@@ -316,6 +316,26 @@ def traverse_jaxpr_params(f, params):
           if type(p) in (Jaxpr, ClosedJaxpr)}
 
 
+def evaluate_jaxpr_eqn(eqn, in_vals):
```
froystig, line 319: nit: `eval_jaxpr_eqn`, for consistency with `eval_jaxpr`.

froystig (approved):

PR #7419, opened 2021-07-30
Windows build related updates

cloudhan:
Disable various s3 like fs support in windows build. Originally enabled due to compiling error. Now it seems upstream have addressed them.

Workaround weird issue with non-standard CUDA path. See https://github.com/tensorflow/tensorflow/issues/51040 for more information.

bazel auto download failed due to wrong machine name.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -126,7 +126,7 @@ bazel_packages = {
             file="bazel-4.1.0-darwin-arm64",
             sha256=
             "c372d39ab9dac96f7fdfc2dd649e88b05ee4c94ce3d6cf2313438ef0ca6d5ac1"),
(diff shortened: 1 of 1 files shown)


cloudhan:
I added a windows builder which you can find it at https://github.com/cloudhan/jax-windows-builder

cloudhan:
It only make the build pass, but has some problems with DLL loading.

zhangqiaorjc:
@hawkinsp

cloudhan:
@hawkinsp Could you take time to look into this?

hawkinsp (approved):

hawkinsp:
Thanks for the PR!

PR #7422, opened 2021-07-30
[ROCm] Change skip_on_devices to handle device tags

reza-amd:
**Please do not merge this PR before merging https://github.com/tensorflow/tensorflow/pull/50972. Its functionality depends on the change in platform_version in TF/PJRT**

This PR changes the skip_in_devices to compare the input list of skipped devices to a set of tags assigned based on the underlying device. 

How does it work?
When a test case decorated with skip_on_devices is executed, we find all the associated tags with the current device and compare it with the skip_on_devices list. If there is a common item between these two, we skip the test.

How can it be helpful?
At runtime, we specify all the tags associated with the device and it is more flexible than using a single tag for the device. For example, we have adopted this to skip on some FFT tests only for ROCm.

@hawkinsp , @deven-amd

### DIFF
--- a/jax/test_util.py
+++ b/jax/test_util.py
@@ -423,16 +423,32 @@ def skip_if_unsupported_type(dtype):
     raise unittest.SkipTest(
       f"Type {dtype.name} not supported on {device_under_test()}")
 
+def is_device_rocm():
+  return xla_bridge.get_backend().platform_version.startswith('rocm')
+
+def is_device_cuda():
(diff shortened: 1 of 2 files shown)


reza-amd:
https://github.com/tensorflow/tensorflow/pull/50972 has been merged. This PR is ready to be reviewed.

zhangqiaorjc:
@hawkinsp

hawkinsp (approved):

PR #7423, opened 2021-07-30
Add `jax_hlo_source_file_canonicalization_regex` config.

skye:
This is meant to be used with @colemanliyah's persistent compilation cache, since the serialized HLO computation (including the source_file metadata) is used in the cache key. The config can be used to remove bits of the source file path that vary between program invocations, to avoid spurious cache misses.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -490,11 +490,21 @@ log_compiles = config.define_bool_state(
           'DEBUG.'))
 
 distributed_debug = config.define_bool_state(
-    name="jax_distributed_debug",
+    name='jax_distributed_debug',
     default=False,
     help=('Enable logging useful for debugging multi-process distributed '
           'computations. Logging is performed with `absl.logging` at WARNING '
           'level.'))
+
+hlo_source_file_canonicalization_regex = config.define_string_state(
+    name='jax_hlo_source_file_canonicalization_regex',
+    default=None,
+    help=('Used to canonicalize the source_path metadata of HLO instructions '
+          'by removing the given regex. If set, re.sub() is called on each '
+          'source_file with the given regex, and all matches are removed. '
+          'This can be used to avoid spurious cache misses when using the '
+          'persistent compilation cache, which includes HLO metadata in the '
+          'cache key.'))
 
 def _update_x64_global(val):
   lib.jax_jit.global_state().enable_x64 = val
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -16,6 +16,7 @@
 from collections import defaultdict, deque
 import itertools as it
 import operator as op
+import re
 from typing import (Any, Callable, Dict, List, Optional, Sequence, Set, Type,
                     Tuple, Union, NamedTuple)
 from warnings import warn
@@ -93,6 +94,13 @@ def _make_array_shape(a):
   else:
     return (xc.Shape.array_shape(a.dtype, a.shape),)
 
+def _get_canonical_source_file(frame: source_info_util.Frame):
+  source_file = frame.file_name
+  if config.jax_hlo_source_file_canonicalization_regex:
+    source_file = re.sub(config.jax_hlo_source_file_canonicalization_regex,
+                         '', source_file)
+  return source_file
+
 tracebacks = {}
 def make_op_metadata(primitive: core.Primitive,
                      params: Dict, *,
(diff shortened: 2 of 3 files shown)


skye:
If anyone has suggestions for the config name let me know. I erred on the side of descriptiveness, since this is a pretty obscure feature, but I'm open to less java-y suggestions...

jax/__init__.py
```diff
@@ -40,7 +40,7 @@
 from ._src.config import (
   config, enable_checks, check_tracer_leaks, checking_leaks,
   debug_nans, debug_infs, log_compiles, default_matmul_precision,
-  numpy_rank_promotion
+  numpy_rank_promotion, hlo_metadata_source_file_prefix_removal_regex
```
hawkinsp, line 43: I don't think I'd export this in the jax namespace. It's not something users should consume.

jax/_src/config.py
```diff
@@ -490,12 +490,20 @@ def update_thread_local_jit_state(**kw):
...
     help=('Enable logging useful for debugging multi-process distributed '
           'computations. Logging is performed with `absl.logging` at WARNING '
           'level.'))
 
+hlo_metadata_source_file_prefix_removal_regex = config.define_string_state(
+    name='jax_hlo_metadata_source_file_prefix_removal_regex',
+    default=None,
+    help=('Used to strip a prefix from the source_path metadata of HLO '
+          'instructions. If set, will be searched as a regex in each '
+          'source_file, and if found, everything up to and including the '
```
hawkinsp, line 504: Perhaps it would be cleaner to remove *exactly* things that match? If you want to remove up to the match prefix it with `.*`.

jax/interpreters/xla.py
```diff
@@ -93,6 +94,15 @@ def _make_array_shape(a):
   else:
     return (xc.Shape.array_shape(a.dtype, a.shape),)
 
+def _get_source_file(frame: source_info_util.Frame):
+  source_file = frame.file_name
+  if config.jax_hlo_metadata_source_file_prefix_removal_regex:
+    match = re.search(
```
hawkinsp, line 100: Just call `re.sub` here?

jax/interpreters/xla.py
```diff
@@ -93,6 +94,15 @@ def _make_array_shape(a):
   else:
     return (xc.Shape.array_shape(a.dtype, a.shape),)
 
+def _get_source_file(frame: source_info_util.Frame):
```
hawkinsp, line 97: Perhaps `_get_canonical_source_path` ?

jax/_src/config.py
```diff
@@ -490,12 +490,20 @@ def update_thread_local_jit_state(**kw):
...
 distributed_debug = config.define_bool_state(
-    name="jax_distributed_debug",
+    name='jax_distributed_debug',
     default=False,
     help=('Enable logging useful for debugging multi-process distributed '
           'computations. Logging is performed with `absl.logging` at WARNING '
           'level.'))
 
+hlo_metadata_source_file_prefix_removal_regex = config.define_string_state(
+    name='jax_hlo_metadata_source_file_prefix_removal_regex',
```
hawkinsp, line 500: Maybe `jax_metadata_source_path_canonicalization_regex` ?

Perhaps also mention that this relates to the cache key in the flag description?

jax/__init__.py
skye: I'm not sure how else to access the context manager. We set jax.config to be the Config singleton object inside config.py. Is there a way to access the config module itself?

I can also add the try/finally logic to the test itself, it just seems silly given we have this nice context manager.

hawkinsp: `import jax._src.config` ?

skye: Oh, duh. I forgot we can do that. That works!

jax/interpreters/xla.py
skye: Done.

jax/interpreters/xla.py
skye: I did `_get_canonical_source_file` to match the existing terminology. It's more accurately a path, but it's called a source_file.

jax/_src/config.py
skye: I did `jax_hlo_source_file_canonicalization_regex`. `hlo` instead of `metadata` because I think that's the higher-order bit of where this takes effect, and `source_file` instead of `source_path` to be consistent with existing nomenclature.

Mentioned the cache key.

jax/_src/config.py
skye: I changed to re.sub() as suggested, which effective does this.

skye (commented):
This is much better now, thanks.

hawkinsp (approved):

skye:
Thanks for the quick reviews!

PR #7425, opened 2021-07-30
Initial commit of jax2iree

stellaraccident:


doug:
This is awesome! ❤️

jax/experimental/jax2iree/examples/simple_convert.py
```diff
@@ -0,0 +1,19 @@
+from jax import numpy as jnp
+from jax.experimental import jax2iree
+
+
+def f(x, y):
```
jpienaar, line 5: Dead? Or did you intend to call this from fabs?

jax/experimental/jax2iree/ir_builder.py
```diff
@@ -0,0 +1,102 @@
...
+        return ir.RankedTensorType.get(shape, element_type)
+      elif isinstance(aval, core.UnshapedArray):
+        element_type = self.convert_dtype_to_ir_type(aval.dtype)
+        return ir.UnrankedTensorType.get(element_type)
+
+      raise NotImplementedError(f"Unsupported AbstractValue conversion: {aval}")
+
+  def convert_dtype_to_ir_type(self, dtype) -> ir.Type:
+    """Convert a dtype to an ir type."""
+    # TODO: Terrible.
```
jpienaar, line 65: Haha, perhaps TODO: Switch on dtype and build appropriate type rather than just uniformly returning f32. f32 is purely for initial commit to create e2e thin slice.

jax/experimental/jax2iree/ir_builder.py
```diff
@@ -0,0 +1,102 @@
...
+    return ir.F32Type.get(self.context)
+
+  def convert_ir_type_to_dtype(self, element_type: ir.Type):
+    """Convert an IR type to a dtype."""
+    # TODO: Terrible.
+    return np.float32
+
+  def get_shaped_type_dims_list(self,
+                                t: ir.ShapedType) -> Sequence[Union[None, int]]:
+    # TODO: Ugh. Has anyone tried to use this before?
```
jpienaar, line 75: Do you need to guard here against unranked?

jax/experimental/jax2iree/ir_builder.py
```diff
@@ -0,0 +1,102 @@
...
+    self.b = builder
+    self.context = self.b.context
+    self.func_op = func_op
+    self.entry_block = self.func_op.add_entry_block()
+    self.ip = ir.InsertionPoint(self.entry_block)
+
+  def emit_return(self, values: Sequence[ir.Value], update_type: bool = True):
+    """Emits a return op, also updating the containing function type."""
+    std.ReturnOp(values, loc=self.b.loc, ip=self.ip)
+    if update_type:
```
jpienaar, line 95: OOC when would this not be needed in this flow? E.g., with tracing it would seem that we'd always reach return type last and so would not know what type we return until end [exception I could see is control flow]

jax/experimental/jax2iree/primitives.py
```diff
@@ -0,0 +1,98 @@
...
+                               Sequence[core.AbstractValue]], params: dict):
+    self.primitive = primitive
+    self.vals = vals
+    self.avals = avals
+    self.out_aval = out_aval
+    self.params = params
+
+  def emit_fallback(self, fb: FunctionBuilder):
+    # TODO: This is super fragile: just to see what is going on.
+    op_name = f"jax_unrecognized.{self.primitive.name}"
```
jpienaar, line 45: What is this used for? Oh yes I think you mentioned this is just for testing to avoid triggering any optimizations etc while bootstrapping. Is that correct?

jax/experimental/jax2iree/primitives.py
```diff
@@ -0,0 +1,98 @@
...
+
+  return decorator
+
+
+HLO_HANDLERS: PrimitiveHandlerTable = {}
+
+
+@_add_handler(HLO_HANDLERS, lax.add_p)
+def hlo_add(fb: FunctionBuilder, inv: PrimitiveInvocation):
+  # TODO: Most of this can be generically meta-programmed.
```
jpienaar, line 82: Tablegen the "old friend" :) (mlir-hs this has worked quite well for the simple cases)

jpienaar (commented):
Drive by :)

stellaraccident:
> Drive by :)

Thanks - most of these items will go away as part of finishing this PR and flipping it out of Draft :)

jax/experimental/jax2iree/primitives.py
stellaraccident: Done right, it would be a real debugging aid: For unrepresented ops, actually emit them for inspection (with location info) vs hard-erroring on the first such and making someone parse python exception logs.

jax/experimental/jax2iree/primitives.py
stellaraccident: I mean, I was thinking some Python goo. There will end up being a few types of ops that dominate and we should have helpers for those.

stellaraccident:
This work prompted me to finally get IREE publishing a first class C/Python API to its compiler (we were previously doing a hoaky, shell out to a compiler executable thing, which isn't the best for JIT/in-process): https://github.com/google/iree/pull/6683

I'll combine that with the existing IREE runtime API early next week and should be able to land this POC spike in a couple of days. It'd be good to chat then about next steps.

stellaraccident:
Ok, the most recent push pulls in our new in-process compiler API and demonstrates basic runtime integration for e2e execution. Still plenty to do, but this represents a complete POC.

gnecula:
Thanks for getting this started!

I was expecting that an IREE integration will be more lightweight from a JAX point of view: export the HLO module, invoke the IREE compiler, plus some runtime integration to run the executables. The current PR seems to use a custom trace to convert each JAX primitive separately. I think that this is very costly: a lot of conversion rules some of which will be very tricky. We do follow this strategy for jax2tf, but there the target language is very far from HLO. Do we really need to do the same for IREE?

stellaraccident:
Also, I'm about to go out on vacation for a few weeks, and I'm in no hurry to move this forward right this minute. What's here is literally a few hours of coding and if it serves to solidify a plan that I/someone can pick up and start working on in September, it will have done its job, from my perspective.

hawkinsp:
On Tue, Aug 10, 2021 at 9:05 AM Stella Laurenzo ***@***.***>
wrote:

> Assuming that such a converter table API existed (or could graduate out of
> this work), this is a pretty lightweight PR and what is left is most likely
> needed. IREE's IR has more features than XLA, especially for things related
> to deployment, and we will want to introduce new primitives for python
> defined kernels, control the way the function boundaries are emitted,
> handle access to globals, etc. All of those reasons are why we find the
> existing HLO proto path unsuitable and I think are worth the extra code.
>


However what you could do is bootstrap your translation rules from the
existing translation rules. For every primitive, at least for those not
parameterized by other computations, you can create a translation rule that
calls the existing HLO translation rule and translate its output. You can
then replace the rules one-by-one as you need. This should get you fairly
complete coverage for a fairly low initial cost.

Some of the translation rules aren't trivial. Not all JAX primitives are
one-to-one with HLO constructs. You can find the most interesting examples
of this by searching for xla.lower_fun() which allows you to call back into
the tracing machinery in a translation rule: no-one wants to write
non-trivial code in the builder API so the non-trivial code typically uses
this facility to embed JAX code in a translation rule.  For an example of
this, see, for example, how `cumsum` is implemented:
https://github.com/google/jax/blob/a93eaf3c9ef7db7bfaf419953adcd37e81a7b5f7/jax/_src/lax/control_flow.py#L2644
Now admittedly that one probably should have an HLO primitive but it
doesn't.

I mention this because you'd get this for free by calling into the existing
translation rules but you'd have to do some work to make it work without
that.


> I don't relish building out the primitive conversion tables, but if that is
> the price of a high quality implementation, it's not *that* hard. It would
> be nice to share most of it, though.
>

Another suggestion:

I'd suggest actually implementing the JAX runtime API so IREE can
transparently integrate with the rest of JAX. (e.g., things like `jit(f)`
will just work but backed by IREE instead of XLA).

You can either implement something that duck-types as the Python API:
https://github.com/tensorflow/tensorflow/blob/76d68c3f2ce78168ffb59f43c39e1ee718c24be1/tensorflow/compiler/xla/python/xla_extension/__init__.pyi#L295
or implement its C++ version:
https://github.com/tensorflow/tensorflow/blob/76d68c3f2ce78168ffb59f43c39e1ee718c24be1/tensorflow/compiler/xla/pjrt/pjrt_client.h

You can implement either, although I'd probably recommend the C++ version
so you can get access to low-latency JIT dispatch. In essence you would
need to implement four classes (Client, Device, Buffer, Executable). Many
methods don't need to be implemented (e.g., infeed/outfeed, channels,
cross-device transfers, etc.) so don't get dissuaded by the fact some APIs
look awfully TPU-specific.

There are a few places where the API may be a little too XLA-specific.
That's fine, we can fix that. The most notable one is that Compile() takes
an XLA computation, but we could, say, make that some sort of opaque object
or add an overload that takes mHLO.

Peter

stellaraccident:
> On Tue, Aug 10, 2021 at 9:05 AM Stella Laurenzo ***@***.***> wrote:
> Assuming that such a converter table API existed (or could graduate out of this work), this is a pretty lightweight PR and what is left is most likely needed. IREE's IR has more features than XLA, especially for things related to deployment, and we will want to introduce new primitives for python defined kernels, control the way the function boundaries are emitted, handle access to globals, etc. All of those reasons are why we find the existing HLO proto path unsuitable and I think are worth the extra code.
> However what you could do is bootstrap your translation rules from the existing translation rules. For every primitive, at least for those not parameterized by other computations, you can create a translation rule that calls the existing HLO translation rule and translate its output. You can then replace the rules one-by-one as you need. This should get you fairly complete coverage for a fairly low initial cost. Some of the translation rules aren't trivial. 

Were you thinking of literally calling the existing machinery, producing an XlaComputation and then converting that to  equivalent MHLO ops? Some of that code does exist but it is quite far in terms of build systems and deps (and is old/brittle in other ways that add tax)... and then we are in the space of evaluating how much to invest in bootstrap paths, etc (esp. for OSS builds where dependency boundaries are strict and high cost). One possible simplification: we implement a duck-typed implementation of XlaBuilder in the MHLO repo and have a way to inject that JAX-side. There'd still be typing involved but it would probably be pretty formulaic and relatively easy to rip through and/or meta-program. We could either implement that in Python or as a pybind module against the C-API.

> Not all JAX primitives are one-to-one with HLO constructs. You can find the most interesting examples of this by searching for xla.lower_fun() which allows you to call back into the tracing machinery in a translation rule: no-one wants to write non-trivial code in the builder API so the non-trivial code typically uses this facility to embed JAX code in a translation rule. For an example of this, see, for example, how `cumsum` is implemented: https://github.com/google/jax/blob/a93eaf3c9ef7db7bfaf419953adcd37e81a7b5f7/jax/_src/lax/control_flow.py#L2644 Now admittedly that one probably should have an HLO primitive but it doesn't. I mention this because you'd get this for free by calling into the existing translation rules but you'd have to do some work to make it work without that.

Nice - that is how it should be (imo).

> I don't relish building out the primitive conversion tables, but if that is the price of a high quality implementation, it's not *that* hard. It would be nice to share most of it, though.
> Another suggestion: I'd suggest actually implementing the JAX runtime API so IREE can transparently integrate with the rest of JAX. (e.g., things like `jit(f)` will just work but backed by IREE instead of XLA). You can either implement something that duck-types as the Python API: https://github.com/tensorflow/tensorflow/blob/76d68c3f2ce78168ffb59f43c39e1ee718c24be1/tensorflow/compiler/xla/python/xla_extension/__init__.pyi#L295 or implement its C++ version: https://github.com/tensorflow/tensorflow/blob/76d68c3f2ce78168ffb59f43c39e1ee718c24be1/tensorflow/compiler/xla/pjrt/pjrt_client.h You can implement either, although I'd probably recommend the C++ version so you can get access to low-latency JIT dispatch. In essence you would need to implement four classes (Client, Device, Buffer, Executable). 

Thanks for these references. A C++ implementation is likely where we need to land, but I'm tempted to start with an iteration in pure Python just for the value of seeing something connect a bit more easily (to my eyes). IREE's Python runtime API is really just a toy for functional evaluation, not for a serious integration with a production stack (it takes shortcuts with its modeling of host/device memory, has limited support for multi-device, doesn't expose most of the concurrency/synchronization primitives, etc) -- however, it makes up for that with a lot of convenience :)  IREE's C-API models everything but is, of course, harder to use. There are also real design questions that come up when looking at bridging IREE-C+PJRT in the most efficient way possible, and I would feel better about having something/anything working to compare against while working those out.

Conversely, the IREE/MLIR-based compilers are easier to integrate with from Python if you are already largely homed in Python (as JAX is). There is also a C-API, and we did take that approach [with PyTorch](https://github.com/llvm/mlir-npcomp/tree/main/frontends/pytorch/csrc/builder), although, in that case, it is because PyTorch's IR is implemented in C++ and that is where you get the most unfettered access to it. The complexity and code cost is high, however. For JAX, I think we can largely stay at the Python level for the main compiler and then just use that from the runtime APIs related to compilation. That also makes it easier to do some more interesting things. Do you see any issues with that path? 

> Many methods don't need to be implemented (e.g., infeed/outfeed, channels, cross-device transfers, etc.) so don't get dissuaded by the fact some APIs look awfully TPU-specific. There are a few places where the API may be a little too XLA-specific. That's fine, we can fix that. The most notable one is that Compile() takes an XLA computation, but we could, say, make that some sort of opaque object or add an overload that takes mHLO. Peter
> […](#)

Since I'm new to your codebase, could you provide some code pointers to what on the JAX side invokes this runtime layer? Reading that code would help my mental model and maybe make me useful in offering concrete suggestions for generalizing.

jpienaar:
On Tue, Aug 10, 2021, 08:06 Stella Laurenzo ***@***.***>
wrote:

> On Tue, Aug 10, 2021 at 9:05 AM Stella Laurenzo *@*.***> wrote:
> Assuming that such a converter table API existed (or could graduate out of
> this work), this is a pretty lightweight PR and what is left is most likely
> needed. IREE's IR has more features than XLA, especially for things related
> to deployment, and we will want to introduce new primitives for python
> defined kernels, control the way the function boundaries are emitted,
> handle access to globals, etc. All of those reasons are why we find the
> existing HLO proto path unsuitable and I think are worth the extra code.
> However what you could do is bootstrap your translation rules from the
> existing translation rules. For every primitive, at least for those not
> parameterized by other computations, you can create a translation rule that
> calls the existing HLO translation rule and translate its output. You can
> then replace the rules one-by-one as you need. This should get you fairly
> complete coverage for a fairly low initial cost. Some of the translation
> rules aren't trivial.
>
> Were you thinking of literally calling the existing machinery, producing
> an XlaComputation and then converting that to equivalent MHLO ops? Some of
> that code does exist but it is quite far in terms of build systems and deps
> (and is old/brittle in other ways that add tax)... and then we are in the
> space of evaluating how much to invest in bootstrap paths, etc (esp. for
> OSS builds where dependency boundaries are strict and high cost). One
> possible simplification: we implement a duck-typed implementation of
> XlaBuilder in the MHLO repo and have a way to inject that JAX-side.
>
Not sure if this is what you mean, but we have an XlaBuilder variant that
produces MHLO already. I think it is just a change in constructor of the
builder (of course there may be a caveat with nested builders as we haven't
used that much, but easy to add). We could effectively move that variant
into MHLO side, it could get tricky with helper functions (but not too
bad/those should perhaps be MHLOs and can expand later, sort of as Peter
mentioned some of them should be ops).

stellaraccident:
> On Tue, Aug 10, 2021, 08:06 Stella Laurenzo ***@***.***> wrote:
> On Tue, Aug 10, 2021 at 9:05 AM Stella Laurenzo *@*.***> wrote: Assuming that such a converter table API existed (or could graduate out of this work), this is a pretty lightweight PR and what is left is most likely needed. IREE's IR has more features than XLA, especially for things related to deployment, and we will want to introduce new primitives for python defined kernels, control the way the function boundaries are emitted, handle access to globals, etc. All of those reasons are why we find the existing HLO proto path unsuitable and I think are worth the extra code. However what you could do is bootstrap your translation rules from the existing translation rules. For every primitive, at least for those not parameterized by other computations, you can create a translation rule that calls the existing HLO translation rule and translate its output. You can then replace the rules one-by-one as you need. This should get you fairly complete coverage for a fairly low initial cost. Some of the translation rules aren't trivial. Were you thinking of literally calling the existing machinery, producing an XlaComputation and then converting that to equivalent MHLO ops? Some of that code does exist but it is quite far in terms of build systems and deps (and is old/brittle in other ways that add tax)... and then we are in the space of evaluating how much to invest in bootstrap paths, etc (esp. for OSS builds where dependency boundaries are strict and high cost). One possible simplification: we implement a duck-typed implementation of XlaBuilder in the MHLO repo and have a way to inject that JAX-side.
> Not sure if this is what you mean, but we have an XlaBuilder variant that produces MHLO already. I think it is just a change in constructor of the builder (of course there may be a caveat with nested builders as we haven't used that much, but easy to add). We could effectively move that variant into MHLO side, it could get tricky with helper functions (but not too bad/those should perhaps be MHLOs and can expand later, sort of as Peter mentioned some of them should be ops).

Ah, I didn't know we'd made progress on that (I know it had been discussed for a long time). I'll ping you and see if some code organization could make it easier/possible to leverage.

hawkinsp:
> Since I'm new to your codebase, could you provide some code pointers to what on the JAX side invokes this runtime layer? Reading that code would help my mental model and maybe make me useful in offering concrete suggestions for generalizing.

This may help: https://gist.github.com/hawkinsp/bcbd8375177a5249f2bf8de6a7ef0cda

This is a toy backend that doesn't actually compile anything but instead hardwires in a particular computation. However it shows you how to register a backend and use it to run a trivial computation. If you have a utility to convert an XLA HLO proto into IREE, you could probably use this to plug in IREE without too much further complication as a first step.

(This is mostly orthogonal to what you have done in this PR, which is about avoiding the HLO step.)

stellaraccident:
Thanks, this is the piece I was missing! Let me give this a shot with our
current mechanisms for bridging HLO and see where we go.

stellaraccident:
FYI - I'll be on vacation for a couple of weeks and will pick this up when back.

jax/experimental/jax2iree/jax2iree.py
```diff
@@ -0,0 +1,168 @@
...
+
+  @property
+  def aval(self):
+    return self._aval
+
+  def full_lower(self):
+    return self
+
+
+class IreeTrace(core.Trace):
```
apaszke, line 124: This is a fine way of constructing the trace, but it might be simpler to first trace the function to jaxpr and then write a jaxpr -> IREE module translation (really just a small interpreter). Traces are ok for simple cases but will get more hairy once you get to handling call/map primitives

stellaraccident:
I'm back from vacation and am getting caught back up on this. I've gone ahead and landed what is here so far into IREE's samples repository so I can iterate on it a bit more: https://github.com/google/iree-samples/tree/main/pre-release/python

I'll rework it now to add the discussed runtime interop and a JAXPR interpreter, then we can discuss where to take it.

stellaraccident:
> > Since I'm new to your codebase, could you provide some code pointers to what on the JAX side invokes this runtime layer? Reading that code would help my mental model and maybe make me useful in offering concrete suggestions for generalizing.
> 
> This may help: https://gist.github.com/hawkinsp/bcbd8375177a5249f2bf8de6a7ef0cda
> 
> This is a toy backend that doesn't actually compile anything but instead hardwires in a particular computation. However it shows you how to register a backend and use it to run a trivial computation. If you have a utility to convert an XLA HLO proto into IREE, you could probably use this to plug in IREE without too much further complication as a first step.
> 
> (This is mostly orthogonal to what you have done in this PR, which is about avoiding the HLO step.)

Thanks for this sample. Here is our iree-samples commit with it working: https://github.com/google/iree-samples/commit/d0f266a2bf0432e4ea71302c8c3b03b43c72cc9d

This is a toy implementation for a few reasons:

* Just marshalling through numpy ndarray vs using lower level buffers and avoiding copies.
* Uses our HLO->MLIR utility which involves a lot of ASM munging and a process jump.
* Hard-codes device setup and such (just for a single CPU).
* Uses the synchronous Python API vs the lower level C-API, which allows for async dispatch and corresponding buffer lifetime, etc: If ever making this more than a quick way to evaluate IREE, probably want to drop down to C.

I need to dive in and see how to redirect this away from its reliance on an XlaComputation. Maybe some kind of higher level hook that delegates to the normal Xla way (haven't grok'd the code at all).

hawkinsp:
Closing this PR because I think the IREE and MHLO integration that are checked in replace it (e.g,.  https://github.com/google/jax/pull/8529 )

Issue #5000, opened 2020-11-23
Working with data of different lengths using automasking


noashin:
I am working with multiple time series of different lengths, that should all undergo the same operations. 
Following the [tweet](https://twitter.com/marcvanzee/status/1305516283941867521) from @marcvanzee I attempted to use mask and vmap for these manipulations. Unfortunately due to the lack of documentation, I am not sure I can fully decipher Marc's example. 
The simplest operation I am attempting is concatenating two padded arrays. Is it possible to parallelize it using mask and vmap?
```python
import jax.numpy as np

# each row is a data_set of different length. Zero padding is used to put it all in one array.
data_sets_a = np.array([[1,2,0,0], [4,5,6,7]])
data_lengths_a = np.array([2, 4])

data_sets_b = np.array([[1,0], [1,2]])
data_lengths_b = np.array([1, 2])

num_data_sets = data_sets_a.shape[0]

# implementation using list comprehension
data_sets_concatenated_list = [np.concatenate([data_sets_a[i, :data_lengths_a[i]], data_sets_b[i, :data_lengths_b[i]]]) \
                               for i in range(num_data_sets)]
data_sets_concatenated = np.array(list(itertools.zip_longest(*data_sets_concatenated_list, fillvalue=0))).T
```

Is it possible to define something like:
```python
def concatenate(array_1, array_2):
    return np.concatenate([array_1, array_2])

input_mask_1 = data_lengths_a
input_mask_2 = data_lengths_b
output_mask = data_lengths_a + data_sets_b
```
and use vmap and mask to get the same result as in the code above?

JossWhittle:
Hi noashin, I'm looking at similar stuff and I found this comment with an example of the masking feature https://github.com/google/jax/issues/2521#issuecomment-604759386

It's from march 2020 so I don't know if they've changed anything about the API since then. I'm trying to get it working myself too so please say if you manage to get it working :)

noashin:
So far no luck. For the more basic operations I need I am using padded arrays, but for the more complicated ones, I am currently looping.

myagues:
Not an *exact* answer, but looking at the masking tests and that comment:
```python
@functools.partial(jax.mask, in_shapes=["n", "m"], out_shape="_")
def mask_concat(x, y):
    return jnp.pad(jnp.concatenate([x, y]), (0, 1))

[...]

PR #7428, opened 2021-07-31
Implement squeeze masking rule

shawwn:
This PR implements a masking rule for `lax.squeeze`.

## Why?

`lax.matmul` calls `lax.squeeze`:

https://github.com/google/jax/blob/655a3e79c23060e87595d4a52dc595352fa29c9b/jax/_src/numpy/lax_numpy.py#L4212-L4213

... which was throwing an error when multiplying masked matrices `a` and `b` shaped like:

```
a.polymorphic_shape = (1, 12, 1, n + 1)
b.polymorphic_shape = (1, 12, n + 1, 64)
```

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -3899,10 +3899,15 @@ def _squeeze_batch_rule(batched_args, batch_dims, *, dimensions):
   dimensions = tuple(np.add(1, dimensions))
   return squeeze(operand, dimensions=dimensions), 0
 
+def _squeeze_masking_rule(padded_args, logical_shapes, *, dimensions):
+  operand, = padded_args
+  return squeeze(operand, dimensions=dimensions)
+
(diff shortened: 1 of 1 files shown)


shawwn:
A few other useful masking rules are missing too, like broadcast_in_dims and einsum. I could create a more comprehensive PR to implement all of these, or just submit PRs for each of them separately.

I got mask() working with my GPT sampling loop — it’s great!

shoyer:
Interestingly, masking was part of the original motivation for the `lax.squeeze()` primitive, but note that the implementation of `mask` is currently being revisited: https://github.com/google/jax/issues/5000#issuecomment-757354298

Given the current status, I'm not sure if we want accept these sort of PRs.

mattjj:
Thanks! But we deleted `mask`, so this PR is definitely stale. Closing!

PR #7443, opened 2021-08-02
Type stubs for jaxlib.xla_extension no longer use -stubs suffix

superbobry:
PEP-561 does not specify whether subpackages of a non-stub-only-package
could use the -stubs suffix. setuptools seems to allow that, yet mypy fails
to resolve the subpackage with a -stubs suffix.

This PR makes jaxlib.xla_extension a ~normal package with a toplevel `__init__.pyi`.

Unblocks #7386.


hawkinsp (approved):
I assume you've tried this by building and installing a jaxlib?

superbobry:
Yep, I've built&installed a local jaxlib and ran both pytest and mypy with #7386 checked out.

PR #7456, opened 2021-08-02
Validate shapes for boolean indices

### LINKED ISSUE #7329: Boolean indexing/masking with the wrong shape should fail
This example results in an error with NumPy but not jax.numpy:
```python
import jax.numpy as jnp

x = jnp.arange(4)
y = jnp.array([True, False])
x[y]   # DeviceArray([0], dtype=int32)
```

It looks like JAX converts the boolean indexer array into an integer array for indexing without checking the shape.

We should raise an error in this case: indexing an array of shape (4,) with a boolean array of shape (2,) is never valid.

jakevdp:
Fixes #7329

This potentially breaks existing code (because invalid shapes previously passed through silently) so I'm going to run an internal TGP before attempting to merge.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4922,6 +4922,7 @@ def _unique_axis_sorted_mask(ar, axis):
   size, *out_shape = aux.shape
   aux = aux.reshape(size, _prod(out_shape)).T
   if aux.shape[0] == 0:
+    size = 1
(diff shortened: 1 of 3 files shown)


jakevdp:
Test failures are real – I'll take a look at it.

jakevdp:
I fixed a couple places where we were using invalid boolean indices. TGP tests were clean aside from the failures seen here.

zhangqiaorjc (approved):

Issue #3124, opened 2020-05-17
Getting all values in a pytree


joaogui1:
Hi, sorry if this is a question with an obvious answer, but how do I get all the values in a pytree?
I tried using `tree_flatten`, but it gives me the device arrays, not the values they contain
(For extra context I want to implement regularizers using jax and need to sum all the squares/absolute values of the neural network parameters)

jacobjinkelly:
Does work `flat_x, unravel = ravel_pytree(x)`? IIUC you want all your parameters in one long vector?

joaogui1:
That's exactly it! Thanks @jacobjinkelly , didn't know about the `flatten_util` module

mattjj:
It works to ravel things, but that might actually be less efficient because it probably involves concatenating all the values into a single contiguous buffer. It might be better to do something like

```python
def l2_normsq(x):
  leaves, _ = tree_util.tree_flatten(x)
  return sum([np.sum(leaf ** 2) for leaf in leaves])
```

(where `sum` is just Python's builtin.)

joaogui1:
Thanks for the tip @mattjj

samuela:
> Does `flat_x, unravel = ravel_pytree(x)` work? IIUC you want all your parameters in one long vector?

Does `ravel_pytree` still exist? I'm not able to find it in the docs: https://jax.readthedocs.io/en/latest/search.html?q=ravel_pytree&check_keywords=yes&area=default

jakevdp:
It does still exist:
```
from jax.flatten_util import ravel_pytree
```
It looks like it's not listed in JAX's documentation

PR #7458, opened 2021-08-03
add ravel_pytree to generated docs

jakevdp:
Addresses https://github.com/google/jax/issues/3124#issuecomment-891409507

### DIFF
--- a/docs/jax.flatten_util.rst
+++ b/docs/jax.flatten_util.rst
@@ -1,0 +1,14 @@
+jax.flatten_util package
+========================
+
+.. currentmodule:: jax.flatten_util
+
+.. automodule:: jax.flatten_util
+
+List of Functions
+-----------------
+
+.. autosummary::
+   :toctree: _autosummary
+
+   ravel_pytree
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -18,6 +18,7 @@ Subpackages
     jax.ops
     jax.random
     jax.tree_util
+    jax.flatten_util
     jax.dlpack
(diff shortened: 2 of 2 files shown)


froystig (approved):

carlosgmartin:
@jakevdp Out of curiosity, is there any reason why [`ravel_pytree`](https://jax.readthedocs.io/en/latest/_autosummary/jax.flatten_util.ravel_pytree.html) is under [`flatten_util`](https://jax.readthedocs.io/en/latest/jax.flatten_util.html) (where it is the only function) rather than [`tree_util`](https://jax.readthedocs.io/en/latest/jax.tree_util.html) with the other tree functions?

mattjj:
I think it's because tree_util doesn't depend on any array stuff (array data types, jax.numpy, etc) whereas the ravel stuff does.

carlosgmartin:
@mattjj Another question: Is excluding `flatten_util` from [`jax.__init__`](https://github.com/google/jax/blob/main/jax/__init__.py) (unlike, say, `tree_util`) intentional?

froystig:
Yeah, that is intentional.

carlosgmartin:
@froystig Thanks. Do you know why it's excluded? (Just trying to understand the architectural decisions a bit better.)

froystig:
We haven't seen a strong case for considering it a public API (in turn taking on the cost of maintainenance according to the [compatibility policy](https://jax.readthedocs.io/en/latest/api_compatibility.html)). We generally try to keep a minimal API surface.

PR #7477, opened 2021-08-03
Jittable eigh

adam-lewis-gcorp:
@hawkinsp @tlu7 @rmlarson

This commit evades the varying blocksize issue associated with QDWH-SVD by using a different polynomial iteration that computes a rank-k orthogonal projector for given k directly. The specific algorithm is described here: https://core.ac.uk/download/pdf/79508571.pdf.

I'd been trying to avoid this because the performance and error behaviour, while still not awful, tends not to be quite so good as obtained by QDWH. But the solutions I came up with turned out to be generally even worse.

### DIFF
--- a/jax/_src/scipy/eigh.py
+++ b/jax/_src/scipy/eigh.py
@@ -21,6 +21,32 @@ import jax.scipy as jsp
 from jax import lax
 
 
+def _fill_diagonal(X, vals):
+  return jax.ops.index_update(X, jnp.diag_indices(X.shape[0]), vals)
+
+
+def _shift_diagonal(X, val):
+  return _fill_diagonal(X, X.diagonal() + val)
+
+
+def _gershgorin(H):
+  """ Computes bounds on the smalles and largest magnitude eigenvalues
+  of H using the "Gershgorin circle theorem".
+  """
+  H_diag = jnp.diag(H)
+  diag_elements = jnp.diag_indices_from(H)
+  abs_H_diag0 = jnp.abs(H.at[diag_elements].set(0.))
+  col_sums = jnp.sum(abs_H_diag0, axis=0)
+  row_sums = jnp.sum(abs_H_diag0, axis=1)
+  row_min = jnp.min(H_diag - row_sums)
+  row_max = jnp.max(H_diag + row_sums)
+  col_min = jnp.min(H_diag - col_sums)
+  col_max = jnp.max(H_diag + col_sums)
+  min_est = jnp.max(jnp.array([row_min, col_min]))
+  max_est = jnp.min(jnp.array([row_max, col_max]))
+  return min_est, max_est
+
+
 def _similarity_transform(
     matrix_in, matrix_out, precision=lax.Precision.HIGHEST):
   """ Returns matrix_out.conj().T @ matrix_in @ matrix_out, done in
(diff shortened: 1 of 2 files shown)


jax/_src/scipy/eigh.py
```diff
@@ -21,6 +21,32 @@
 from jax import lax
 
 
+def _fill_diagonal(X, vals):
+  return jax.ops.index_update(X, jnp.diag_indices(X.shape[0]), vals)
```
hawkinsp, line 25: JAX style thing: I'd always prefer `X.at[index].update(y)` over `index_update` these days. They are synonyms but I think the terser syntax is more readable.

tests/lax_scipy_test.py
```diff
@@ -559,8 +554,13 @@ def test_spectral_dac_eigh(self, linear_size, seed, dtype):
       self.assertRaises(
         NotImplementedError, jax._src.scipy.eigh.eigh, H)
       return
-    evs, V = jax._src.scipy.eigh.eigh(H)
+    evs, V = jax._src.scipy.eigh.eigh(H, termination_size=2)
     ev_exp, eV_exp = jnp.linalg.eigh(H)
+    #guess = jax._src.scipy.eigh._canonically_purify_initial_guess(H, linear_size // 2)
```
rmlarsen, line 559: delete?

tests/lax_scipy_test.py
```diff
@@ -576,12 +576,8 @@ def test_spectral_dac_eigh(self, linear_size, seed, dtype):
       'linear_size': linear_size, 'seed': seed, 'dtype': dtype}
     for linear_size in linear_sizes
     for seed in seeds
-    for dtype in jtu.dtypes.floating))
+    for dtype in [jnp.float32, ]))
```
rmlarsen, line 579: revert?

tests/lax_scipy_test.py
```diff
@@ -576,12 +576,8 @@ def test_spectral_dac_eigh(self, linear_size, seed, dtype):
...
   def test_spectral_dac_svd(self, linear_size, seed, dtype):
```
rmlarsen, line 580: let's use "bisection" instead of "dac" (which is hardcoded to translate to "digital to analog converter" in my brain ;-) )

tests/lax_scipy_test.py
```diff
@@ -59,7 +59,7 @@
 methods = ["qdwh", "svd"]
 seeds = [1, 10]
 
-linear_sizes = [16, 128, 256]
+linear_sizes = [8,] #128, 256]
```
rmlarsen, line 62: revert? We should test for larger sizes.

jax/_src/scipy/eigh.py
```diff
@@ -190,6 +238,24 @@ def _eigh_work(
...
+    raise TypeError(f"Input H of shape {H.shape} must be square.")
+  if symmetrize:
+    H = 0.5 * (H + H.conj().T)
+
+  if ncols <= termination_size:
+    return jnp.linalg.eigh(H)
+
+  evals, evecs = _eigh_work(H, None, precision, termination_size)
+  sort_idxs = jnp.argsort(evals)
+  evals = evals[sort_idxs]
```
rmlarsen, line 254: out of curiosity: Does JAX have a fast O(n) in-place permute implementation?

jax/_src/scipy/eigh.py
```diff
@@ -30,6 +56,62 @@ def _similarity_transform(
   return jnp.dot(out, matrix_out, precision=precision)
 
 
+def _canonically_purify_initial_guess(H, rank):
+  lambda_min, lambda_max = _gershgorin(H)
+  N = H.shape[0]
+  mu = jnp.trace(H) / N
+  theta = rank / N
+  beta = theta / (lambda_max - mu)
```
rmlarsen, line 64: We should probably guard against division by zero here. This would blow up for H=eye(n), I think.

rmlarsen (changes requested):

hawkinsp:
Closing. We've had a `jit`-table `eigh` for some time by using a bucketing and padding strategy.

PR #7484, opened 2021-08-04
Add jax.invertible to the JAX documentation

shawwn:
This PR adds `jax.invertible` to the official JAX documentation at https://jax.readthedocs.io/en/latest/jax.html

I noticed that the `jax/api.py` function defines `invertible`, which seems like a useful-but-undocumented function. So I decided to try to PR it. Is this the correct way to add something to the docs? If anything else needs to be done, let me know.

(If `jax.invertible` isn't intended to be a public API, feel free to close this PR.)

### DIFF
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -39,6 +39,7 @@ Just-in-time compilation (:code:`jit`)
     device_put_sharded
     default_backend
     named_call
+    invertible
 
 .. _jax-grad:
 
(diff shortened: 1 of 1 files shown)


hawkinsp:
I think it might be wise to keep `invertible` undocumented for the moment, since it only has support for a very limited set of JAX primitives.

hawkinsp:
We ended up deleting `jax.invertible`.

PR #7493, opened 2021-08-04
Add cublas headers. cublas_kernels BUILD target has that dependency

yashk2810:


jaxlib/BUILD
```diff
@@ -81,11 +81,13 @@ cc_library(
     features = ["-use_header_modules"],
     deps = [
         "@org_tensorflow//tensorflow/stream_executor/cuda:cusparse_lib",
+        "@org_tensorflow//tensorflow/stream_executor/cuda:cusolver_lib",
```
hawkinsp, line 84: This probably should really be the headers, not the lib, but fine for now.

hawkinsp (approved):

PR #7494, opened 2021-08-04
Use R2C FFTs to implement jnp.fft.fft() for real inputs.

### LINKED ISSUE #7490: FFT on CPU noticeably slower than SciPy's FFT
The FFT implementation in JAX seems to be noticeably slower than the one in SciPy even though both use some flavor of the PocketFFT FFT implementation.

```python
from jax.config import config

import timeit
import numpy as np
from scipy import fft as sp_fft
from jax import numpy as jnp
from jax import jit, random

key = random.PRNGKey(42)
jr = random.normal(key, (2**26, ))
r = np.array(jr)

N_IT = 7

timing = timeit.timeit(lambda: np.fft.fft(r), number=N_IT) / N_IT
print(f"NumPy: {timing} s")
timing = timeit.timeit(lambda: sp_fft.fft(r), number=N_IT) / N_IT
print(f"SciPy: {timing} s")

jax_fft = jnp.fft.fft
timing = timeit.timeit(
    lambda: jax_fft(jr).block_until_ready(), number=N_IT
) / N_IT
print(f"JAX (unjitted): {timing} s")

jax_fft_jit = jit(jax_fft)
jax_fft_jit(jr)  # Warm-up
timing = timeit.timeit(
    lambda: jax_fft_jit(jr).block_until_ready(), number=N_IT
) / N_IT
print(f"JAX (jitted): {timing} s")
```

On an AMD Ryzen 7 4800H (with Radeon Graphics) with JAX 0.2.18 and jaxlib 0.1.69 installed from PyPI, I get the following timings:

```txt
NumPy: 3.5420487681424544 s
SciPy: 1.9076589474290293 s
JAX (unjitted): 4.5806920641433475 s
JAX (jitted): 4.63735637514245 s
```

The timings improve if I compile JAX(lib) myself though it still compares unfavorably to SciPy (compiled JAX (unjitted): 3.506302507857202 s).

My hypothesis is that JAX is distributing binaries that are suboptimal for recent AMD CPUs and much more importantly JAX is probably using some outdated version of PocketFFT.

hawkinsp:
Fixes #7490 for 1-dimensional FFTs.

I couldn't quite figure out the right rule for N-dimensional FFTs, so I leave that for a future PR from someone.

### DIFF
--- a/jax/_src/lax/fft.py
+++ b/jax/_src/lax/fft.py
@@ -14,6 +14,7 @@
 
 
 from functools import partial
+from typing import Union, Sequence
 
 import numpy as np
 
@@ -43,9 +44,28 @@ def _promote_to_real(arg):
   dtype = dtypes.result_type(arg, np.float32)
   return lax.convert_element_type(arg, dtype)
 
+def _str_to_fft_type(s: str):
+  if s == "FFT":
+    return xla_client.FftType.FFT
+  elif s == "IFFT":
+    return xla_client.FftType.IFFT
+  elif s == "RFFT":
+    return xla_client.FftType.RFFT
+  elif s == "IRFFT":
+    return xla_client.FftType.IRFFT
+  else:
+    raise ValueError(f"Unknown FFT type '{s}'")
+
 @partial(jit, static_argnums=(1, 2))
-def fft(x, fft_type, fft_lengths):
-  if fft_type == xla_client.FftType.RFFT:
+def fft(x, fft_type: Union[xla_client.FftType, str], fft_lengths: Sequence[int]):
+  if isinstance(fft_type, str):
+    typ = _str_to_fft_type(fft_type)
+  elif isinstance(fft_type, xla_client.FftType):
+    typ = fft_type
+  else:
+    raise TypeError(f"Unknown FFT type value '{fft_type}'")
+
+  if typ == xla_client.FftType.RFFT:
     if np.iscomplexobj(x):
       raise ValueError("only real valued inputs supported for rfft")
     x = _promote_to_real(x)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/fft.py
```diff
@@ -78,7 +78,18 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
+  # It's faster to compute a real->complex FFT and form the second half by
+  # complex conjugation than to promote to complex and perform a
+  # complex->complex FFT.
+  if (fft_type == xla_client.FftType.FFT
+      and not jnp.issubdtype(a.dtype, np.complexfloating)
+      and len(axes) == 1):
+    # TODO(phawkins): extend this logic to n > 1 dimensions.
+    t = lax.fft(a, xla_client.FftType.RFFT, s)
+    start = -2 + (a.shape[-1] % 2)
+    transformed = jnp.concatenate([t, jnp.conj(t[..., start:0:-1])], axis=-1)
```
shoyer, line 90: are you sure this logic is right for both even and odd sizes? I think there might be an off-by-one error, e.g., when the last axis has length 8

jax/_src/numpy/fft.py
```diff
@@ -78,7 +78,18 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
     else:
       s = [a.shape[axis] for axis in axes]
 
-  transformed = lax.fft(a, fft_type, s)
+  # It's faster to compute a real->complex FFT and form the second half by
+  # complex conjugation than to promote to complex and perform a
+  # complex->complex FFT.
```
shoyer, line 83: Doing it this way also has implications for autodiff, e.g., due to the fact that `fft`/`ifft` and `rfft`/`irfft` have very different transpose rules.

So I wonder: would be better to put this inside the `lax.fft` primitive instead?

At the very least, it would be good to make sure that this short-cut doesn't slow down transposes.

jax/_src/numpy/fft.py
```diff
@@ -78,7 +78,18 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
...
+  if (fft_type == xla_client.FftType.FFT
+      and not jnp.issubdtype(a.dtype, np.complexfloating)
+      and len(axes) == 1):
+    # TODO(phawkins): extend this logic to n > 1 dimensions.
```
shoyer, line 87: I will continue to ponder this...

PR #7514, opened 2021-08-05
Scrub function pointers in op_name metadata in persistent compilation cache key

skye:
The functions pointers were causing spurious cache misses.

I wasn't able to write a unit test for this, due to the existing issue
where identical computations in the same process will have unique
names, making it hard to test cache hits. I manually verified by
logging the cache key with `-v=1` and running the following test case:

```python
with tempfile.TemporaryDirectory() as tmpdir:
    cc.initialize_cache(tmpdir)
    @jax.custom_jvp
    def t(y):
      return y

    def t_jvp(p, t):
      pass

    t.defjvp(t_jvp)

    @jit
    def s(y):
      return t(y)

    s(3)
```

### DIFF
--- a/jax/experimental/compilation_cache/compilation_cache.py
+++ b/jax/experimental/compilation_cache/compilation_cache.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import hashlib
+import re
+
 import jax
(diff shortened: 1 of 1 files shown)


colemanliyah (approved):
LGTM

zhangqiaorjc (approved):

Issue #7510, opened 2021-08-05
jax_debug_nans


manuel-delverme:
The [current implementation of quantile](https://github.com/google/jax/blob/606cbe036af2d18201709a7a33214d9246d1c495/jax/_src/numpy/lax_numpy.py#L5645) function generates NaNs that are detected when jax_debug_nans is activated.

```python
from jax.config import config
config.update("jax_debug_nans", True)
import jax.numpy as jnp
jnp.quantile(jnp.ones((3, 3)), 0.5)
```
```python
from jax.config import config
config.update("jax_debug_nans", True)
import jax.numpy as jnp
jnp.where(1, float('nan'), 1)
```

Traceback
```
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 5560, in percentile
    return quantile(a, q, axis=axis, out=out, overwrite_input=overwrite_input,
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 5370, in quantile
    return _quantile(a, q, axis, interpolation, keepdims, False)
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 5436, in _quantile
    a = where(any(isnan(a), axis=axis, keepdims=True), nan, a)
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 1686, in where
    return _where(condition, x, y)
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 1669, in _where
    x, y = _promote_dtypes(x, y)
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 268, in _promote_dtypes
    return [lax._convert_element_type(x, to_dtype, weak_type) for x in args]
  File "/home/esac/projects/venv/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 268, in <listcomp>
    return [lax._convert_element_type(x, to_dtype, weak_type) for x in args]
FloatingPointError: invalid value (nan) encountered in convert_element_type
python-BaseException
Invalid value encountered in the output of a jit function. Calling the de-optimized version.

```

jakevdp:
[...]

PR #7515, opened 2021-08-05
test_util: provide hook to easily set config values in test cases

jakevdp:
Why? Reduces boilerplate associated with #7208. I made the solution general enough to handle similar future issues (e.g. testing with `debug_nans` enabled; #7510)

### DIFF
--- a/jax/test_util.py
+++ b/jax/test_util.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from contextlib import contextmanager
+import inspect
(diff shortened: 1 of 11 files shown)


jakevdp:
Errors look to be from the fact that the config is not reset after the test.

froystig (approved):

froystig:
This looks good as is, or with a class decorator (to set `_config`) so things look more explicit.

PR #7516, opened 2021-08-05
Use bazel --version. 

### LINKED ISSUE #7507: Use bazel --version to check the bazel version
Please use `bazel --version` to check the version of Bazel. The currently used variant ignores the bazel-startup-options and actually starts a full bazel server which is a) overkill and b) wrong in case of options like `--output-user-root` and no or readonly home folders

Potential change:
```
try:
  version_output = shell([bazel_path, "--version"])
except subprocess.CalledProcessError:
  return False
match = re.search("(Build label:|bazel) *([0-9\\.]+)", version_output)
```

I kept the regexp generic enough so that you can still fall back to the current `bazel version` call and parse both kinds of output

yashk2810:
Fixes #7507

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -209,7 +209,7 @@ def get_bazel_path(bazel_path_flag):
 
 def check_bazel_version(bazel_path):
   try:
-    version_output = shell([bazel_path, "--bazelrc=/dev/null", "version"])
+    version_output = shell([bazel_path, "--version"])
   except subprocess.CalledProcessError:
     return False
   match = re.search("Build label: *([0-9\\.]+)[^0-9\\.]", version_output)


mattjj (approved):

secondspass:
Hi there, this PR breaks the bazel version check because the regex search wasn't appropriately changed to match the output of `bazel --version`. `bazel --version` outputs something like:

```
root@8babaa20f59b:/tmp/jax# bazel --version
bazel 3.7.2- (@non-git)
```
whereas the [regex search](https://github.com/google/jax/pull/7516/files#diff-4a0bef7f4b92e71c6bfdea2a3cc9aeb7335a73541de5c14fbef1a5d6004e852eR215) that checks the bazel version looks like this: 
```
match = re.search("Build label: *([0-9\\.]+)[^0-9\\.]", version_output)
```
That regex search works fine for the output of 
```
root@8babaa20f59b:/tmp/jax# bazel --bazelrc=/dev/null version
Build label: 3.7.2- (@non-git)
Build target: bazel-out/ppc-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Aug 5 18:20:40 2021 (1628187640)
Build timestamp: 1628187640
Build timestamp as int: 1628187640
``` 
which was what was previously used for the version check in the build.py. 

It would be a good idea to revert this commit or fix the regex search to correctly match the output of `bazel --version`.

yashk2810:
https://xkcd.com/1171/

Let me see what I can do!

yashk2810:
Fix here; https://github.com/google/jax/pull/7523/files

Issue #2521, opened 2020-03-26
Avoid recompilation caused by variably shaped arrays?


john-heyer:
I'm working with a function that computes a likelihood given a 1D array. The problem is that this array has variable length, and so the `jit`'ed function triggers recompilation when the length of this array changes, which happens frequently in my program, causing a significant hit in performance.

I know that there are various levels of abstraction, and see the existence of an `Unshaped` arrays in the docs: "JAX can also trace at higher levels of abstraction, like Unshaped, but that’s not currently the default for any transformation".

First, is it even possible with JAX to compile without declaring the shape a compile-time constant? Would using `Unshaped` arrays solve this? If so, how can I tell JAX to trace with this level of abstraction?

Thank you!

mattjj:
Another great question!

There are two separate issues here worth disentangling:
1. staging programs out of Python by trace-specializing, and
2. compiling those staged-out programs using XLA.

The levels of abstraction in JAX pertain to the former. It's true that JAX can trace-specialize at the Unshaped level, but semi-tangentially there's an interesting tradeoff between amount of abstraction and the restrictions required for Python code to be traceable. For example, we couldn't trace `mean0 = lambda x: x.sum(0) / x.shape[0]` at the Unshaped level, while if we specialize on shape that Python function is traceable. (We could trace `lambda x: np.sin(np.cos(x))` at Unshaped just fine though.) There's a relationship to reverse-mode autodiff too: we can transpose `lambda x: x.sum(0)` specialized on shape, but not unshaped, because we need to know what size to broadcast to. (We could transpose `lambda x: x * 5.` though.) Just from the point of view of this tradeoff, it's been a nice sweet spot to specialize on shapes.

But you asked about compilation in particular. Let's set issue 1 aside (say we have a Python function that we can specialize and transform at the Unshaped level, like `lambda x: np.sin(np.cos(x))`). Once we have a staged-out program (i.e. a jaxpr) that is only specialized at the Unshaped level, could we compile one XLA program for it, and thus not have to recompile for different shapes?

[...]

PR #7517, opened 2021-08-05
autodidax prototype of user-customizable partial eval strategies

mattjj:
**This PR comment is a copy-and-paste of a design doc (or most of it) that we wrote to accompany this prototype implementation.** If/when this implementation goes through we'll add a separate design doc in a .md file, so take this PR comment as a draft.

# User-customizable gradient checkpointing policies

_mattjj@, jekbradbury@_

started 2021.07.07, updated 2021.08.05

### Intro and motivation {#intro-and-motivation}

By default, `jax.grad` (and really the underlying `jax.linearize`) computes and stores as much as possible in the forward pass (i.e. the primal computation). For example:

image

Notice that the cosines are evaluated in the primal computation, with the resulting values stored as "residual" outputs to be used as constants in the tangent or cotangent jaxpr.

`jax.checkpoint` (aliased as `jax.remat`) lets users switch this strategy to the opposite extreme: don't store anything except inputs to the function, and instead recompute (i.e rematerialize) all the values needed for the linear computation. At its heart `jax.checkpoint` is a call-like primitive with a special partial evaluation rule, which is why it works for `jax.linearize` as well as `jax.vjp`: both are based on partial evaluation. For example:

image

Notice that now the cosines, and even some sines, are evaluated in the tangent and cotangent jaxprs, with no residuals stored beyond the input to the function. (In addition to this custom partial evaluation behavior, the primitive underlying `jax.checkpoint` also has [a special transpose rule](https://github.com/google/jax/blob/78a689bb09799d8b13b28fe90bee6a04bddb7968/jax/interpreters/ad.py#L547-L570) for evaluating these nonlinear expressions first, as well as [a special XLA HLO translation rule](https://github.com/google/jax/blob/78a689bb09799d8b13b28fe90bee6a04bddb7968/jax/interpreters/xla.py#L1465-L1478) to prevent CSE optimizations from foiling its intended behavior. These additional rules are why it makes sense to model `jax.checkpoint` as a call-like primitive.)

Even if neither the “store everything” nor the “store nothing except inputs” policies is right for a particular application, in principle users can apply `jax.checkpoint` to sub-functions as needed to control exactly what to store and what to recompute. But this API can require refactoring model code just to express the kind of checkpointing desired. Not only does the current API make expressing the right thing hard, but it also makes revising the strategy, for example to experiment with new ones or switch based on hardware backend, untenable.

In particular, there’s an intermediate activation-storage policy that’s a good fit for many memory-sensitive neural net applications on accelerators: storing the outputs of dots and convolutions, but recomputing everything else. (The motivation for this policy is that fused element-wise nonlinear operations are generally much cheaper than dots and convolutions on GPUs and TPUs, or, when fused into those operations on TPUs, completely free.) That’s easy to say in words, but it was previously impossible to express at the top level, without diving into neural network layer code that might have been written by others as part of a library. 

The situation was roughly:

```
# in library code
def apply_layer(W, x):
  return jnp.sin(jnp.dot(W, x))

# in top-level user code
def predict(params, x):
  for W in params[:-1]:
    x = apply_layer(W, x)
  return jnp.dot(params[-1], x)
```

We can't get what we want by applying `jax.checkpoint` to `apply_layer`; that would actually recompute the outputs of dots while saving the outputs of the nonlinear operations in the nonlinearity's JVP rule. It's the opposite of what we want! We'd need something more like `jax.checkpoint(lambda W, x: jnp.dot(W, jnp.sin(x)))`, but that's a strange way to organize layers in neural networks. And do we really want to require users to refactor libraries into non-idiomatic organizations just for checkpointing strategies? No way!

Until recently, this issue was somewhat hidden because XLA:TPU optimizations often ended up producing an optimized graph that used this intermediate strategy (specifically, XLA:TPU is able to duplicate these free nonlinear operations into multiple dot/conv fusions). But it sprang up recently in two contexts. First, users reported significantly higher memory usage for the same JAX program on GPU versus TPU, and we traced part of this back to XLA:GPU not performing the same duplication-into-fusions. 

Second, the issue also sprang up recently because of one more element in the large model application: control flow. XLA:TPU can often figure out great rematerialization strategies, regardless of what JAX's AD produces, so manual control over gradient checkpointing isn't always necessary in the toy code above. But in the real example, we couldn't rely on XLA to sort out what should be stored and what shouldn't be because we were actually doing a scan-over-layers. The presence of that control flow primitive foiled XLA’s rematerialization- related optimizations, and thus required JAX's AD to produce more efficient programs.

Our aim here is **to enable easy-to-revise remat strategies** by improving JAX's AD.

##### Basic idea {#basic-idea}

**As an internal API**, the basic idea is to parameterize `jax.checkpoint` with an optional callback function which represents the checkpointing strategy. The callback takes as input a type-level specification of a primitive application, and returns a boolean indicating whether the value of the application should be saved as a residual:

```
def checkpoint_policy(prim: Primitive, *avals: AbstractValue,
                      **params: Any) -> bool:
  ...

@partial(jax.checkpoint, policy=checkpoint_policy)
def f(...):
  ...
```

This way policies can depend not only on which primitives are being applied but also on input and output sizes. The default `jax.checkpoint` strategy would then correspond to a callback which always returns False, indicating that no values of any applications should be saved. Moreover, the default strategy of AD without `jax.checkpoint` corresponds to a callback which always returns True, indicating that values of every nonlinear application (which is ultimately used in the application of a partially-linear primitive, i.e. is not dead code) should be saved:

```
def default_checkpoint_strategy(*_, **__):
  return False  # no intermediates stored as residuals

def default_ad_strategy(*_, **__):
  return True  # store the value of every nonlinear computation as a residual
```

This general API must be an internal one because it depends on internal details like primitives and abstract values.

**To provide a public API**, we can offer some pre-baked strategies. The clearest current candidate is the `checkpoint_dots` strategy described above, where we save the outputs only of dots and convolutions. Internally to JAX, it could be implemented as

```
def checkpoint_dots(prim: Primitive, *_, **__) -> bool:
  return prim in (lax.dot_general_p, lax.conv_general_dilated_p)
```

To employ this strategy, the user would write something like

```
@partial(jax.checkpoint, policy=jax.checkpoint_policies.checkpoint_dots)
def f(...):
  ...
```

##### Example {#example}

Here's an example of how this policy would work on our toy example (residuals and recomputed operations, representing extra memory and extra compute respectively, are highlighted):

image

Some points to notice:

*   For clarity, where outer products of vectors are needed in cotangent computations, we inlined some transpose expressions into primitive applications and elided insertion of singleton dimensions. These aren't allowed in real jaxprs but keep the example simpler.
*   The sine JVP rule only involves one nonlinear operation (a cosine), but for other nonlinearities there could be greater storage savings relative to the checkpoint-free baseline shown in (a).
*   Since the `checkpoint_dots` strategy used in (c) indicates that we should only save `a1`, it's the only residual output by the primal jaxpr (beyond the inputs to the function that we need, namely `W1`, `W2`, and `x1`). Correspondingly in the cotangent jaxpr we recompute nonlinear values which depend on the value of a1, namely `x2` and `r1`.

##### Nesting: the inner policy wins {#nesting-the-inner-policy-wins}

One detail is how these policies behave under nesting. That is, now that each `jax.checkpoint` decorator can have a different policy, if one `jax.checkpoint`-decorated function calls another, which policy is used for the called function?

```
@partial(jax.checkpoint, policy=policy1)
def f(x):
  ...
  @partial(jax.checkpoint, policy=policy2)
  def g(y):
    ...  # policy1 or policy2 for these primitive applications?
  return g(...)
```

To enable more local control over how autodiff behaves, and to enable more expressive mixing-and-matching of one policy within another, we choose to let the inner policy win. If a top-level power user really wants to control checkpointing policy, and override whatever the library author wanted, they can always resort to a simple transformation to prune inner `jax.checkpoint` primitives.

An alternative, suggested by dougalm@, is that we allow policies to be tagged with a priority score, and when nesting we follow the policy with the higher priority score. That approach gives strictly more control to users and library writers alike, and works well for e.g. NumPy array API priorities. We'll defer on this approach for now because it can be added later as a strict generalization.

### Implementation proposal {#implementation-proposal}

See [the full Autodidax prototype in #7517](https://github.com/google/jax/pull/7517).

##### Fundamentals and jaxpr-to-jaxpr version {#fundamentals-and-jaxpr-to-jaxpr-version}

To explain the implementation of the proposed API, we'll start by thinking in terms of jaxpr-to-jaxpr transformations. The implementation of `jax.checkpoint` would in its partial evaluation rule form a full jaxpr ([like it does currently](https://github.com/google/jax/blob/1646ddaa805625c8acb1385c13de5f111db0aed6/jax/interpreters/partial_eval.py#L754), before this proposal) and then apply this jaxpr-to-jaxpr transformation to build both the eagerly-evaluated jaxpr1 and the staged-out jaxpr2. We'll return to the possibility of doing some transformation work while tracing in a [subsection to follow](#performing-the-non-dce-work-while-tracing-rather-than-jaxpr-to-jaxpr).

Recall from `partial_eval_jaxpr` ([Autodidax reference](https://github.com/google/jax/blob/806899f81d2a2be76e3a25bd0d543bf05138b724/docs/autodidax.py#L2224-L2272)) that partial evaluation 'unzips' one jaxpr into two. Given a jaxpr and a list of booleans indicating which input binders correspond to unknowns, it:

1. for each variable in the jaxpr, associates a boolean indicating whether it is unknown;
2. for each first-order primitive application adds it either into the resulting jaxpr1 (if all its inputs are known, in which case all outputs are known) or else puts it in the resulting jaxpr2 (if any input is unknown, in which case all outputs are unknown); and
3. for each known variable used as an input to a jaxpr2 primitive application, adds it as a residual output to jaxpr1 and residual input to jaxpr2.

The behavior for higher-order primitives is fixed by desiderata like `jit` invariance, which is why we focus on first-order primitives in the descriptions here.

By allowing the user to make some values unsaveable, we need to make some changes:

1. for each variable in the jaxpr, we associate a pair of booleans: one indicating whether the variable is known, and another indicating whether it is instantiated in jaxpr2 (as opposed to being saveable from jaxpr1);
2. some first-order primitive applications may need to be staged into both jaxpr1 and jaxpr2 (as shown in previous examples), and whether a primitive application is ultimately necessary in jaxpr1 or respectively jaxpr2 cannot be decided based solely on the known/unknown and saveable/unsaveable labels of its inputs but instead depends on downstream data dependence, namely whether an output of jaxpr1 or respectively jaxpr2 depends on it (see [Examples](#examples-related-to-dead-code) below);
3. only saveable known inputs to jaxpr2 applications are added as residuals, and the others are made available as outputs of primitive applications in jaxpr2 rather than as top-level residual inputs.

Our implementation is a fairly direct consequence of these new requirements. However, there are a few remaining decisions.

One decision is how much of this work to perform at trace time (i.e. how much to adapt JaxprTrace) versus as a jaxpr-to-jaxpr transformation; we'll defer that discussion to a later subsection.

Another decision is how to handle the dead code issue described in change 2 above, and in particular whether we should generate dead code and then clean it up, or else try to avoid generating it in the first place. We choose the former: we handle 2 by defensively adding some primitive applications to both jaxpr1 and jaxpr2, though they may not be needed downstream; then we rely on a new dead-code elimination (DCE) pass to clean things up. (Notice we can't apply DCE until the very end, i.e. we can't apply it to subjaxprs within a jit since we don't know how its outputs will be used!) An alternative to relying on DCE would be to form jaxpr1 and jaxpr2 in a reverse-dataflow pass; that's viable, but it still requires a backward pass with its own rules for higher-order primitives (i.e. it doesn't offer much of a simplification), and DCE is a useful tool on its own (whereas this alternative backward pass would only be useful for partial evaluation).

Here's [the resulting Autodidax version of partial\_eval\_jaxpr\_custom](https://github.com/google/jax/pull/7517/files#diff-bda36e5567d146410fb8b82f722eba347827bb6faae605b5e87f3a32130f3c99R2905-R2959), as well as [its rule for xla\_call as a higher-order primitive](https://github.com/google/jax/pull/7517/files#diff-bda36e5567d146410fb8b82f722eba347827bb6faae605b5e87f3a32130f3c99R2977-R2994). Here's the implementation of [dce\_jaxpr](https://github.com/google/jax/pull/7517/files#diff-bda36e5567d146410fb8b82f722eba347827bb6faae605b5e87f3a32130f3c99R3006-R3036) and [its rule for xla\_call](https://github.com/google/jax/pull/7517/files#diff-bda36e5567d146410fb8b82f722eba347827bb6faae605b5e87f3a32130f3c99R3040-R3049).

##### Examples related to dead code {#examples-related-to-dead-code}

**Example 1: dead code in jaxpr1 and jaxpr2**

Without DCE, the `partial_eval_jaxpr_custom` implementation described above can produce primitive applications that are unused on either of its output jaxprs. For example, given this input jaxpr to be partially evaluated with a save-nothing policy:

```
{ lambda a:f64[], b:f64[] .
  let c:f64[] = sin a
      d:f64[] = cos a
      e:f64[] = mul d b
  in ( f, e ) }
```

With the inputs labeled as known and unknown, respectively, the cosine is added to jaxpr1 defensively because all its inputs are known, and so based on forward-flow information alone its value could be needed in the primal computation.

In addition, the sine is added to jaxpr2 because the policy does not allow saving its value as a residual, and its value may later be needed in jaxpr2 (as an input to a first-order primitive application with some unknown inputs).

**Example 2: non-dead code in a sub-jaxpr of jaxpr2**

This example is meant to illustrate that we can't perform DCE on jaxpr2 when we form subjaxprs (e.g. for nested `jit` calls), and instead we must defer it until the end. Consider this jaxpr to be partially evaluated with a save-nothing policy:

```
{ lambda a:f64[], b:f64[] .
  let c:f64[] d:f64[] = xla_call [ num_consts=0 ] a
        { lambda a:f64[] .
          let b:f64[] = sin a
              c:f64[] = cos a
          in ( b, c ) }
      e:f64[] = mul d b
  in ( c, e ) }
```

With top-level inputs labeled known and unknown, respectively, both outputs to the xla_call application are known. Yet the value of the cosine application (and not the sine application) is needed in jaxpr2, downstream of the xla_call application, since it is used as an input to a first-order primitive with some unknown inputs.

### DIFF
--- a/docs/autodidax.ipynb
+++ b/docs/autodidax.ipynb
@@ -335,7 +335,10 @@
     "    raise Exception(\"ShapedArray can't be unambiguously converted to bool\")\n",
     "\n",
     "  def str_short(self):\n",
-    "    return f'{self.dtype.name}[{\",\".join(str(d) for d in self.shape)}]'\n",
+    "    dtype_name = self.dtype.name\n",
+    "    for long_name, short_name in [('float', 'f'), ('int', 'i'), ('bool', 'b')]:\n",
+    "      dtype_name = dtype_name.replace(long_name, short_name)\n",
+    "    return f'{dtype_name}[{\",\".join(str(d) for d in self.shape)}]'\n",
     "\n",
     "  def __hash__(self):\n",
     "    return hash((self.shape, self.dtype))\n",
@@ -1093,9 +1096,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from typing import Union\n",
@@ -1348,10 +1349,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_end_of_cell_marker": 0,
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from typing import Set\n",
@@ -1758,8 +1756,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
-    "lines_to_next_cell": 1,
     "tags": [
      "hide-input"
     ]
@@ -1884,9 +1880,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "@contextmanager\n",
@@ -2457,7 +2451,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -2521,7 +2514,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -2559,6 +2551,38 @@
     "  return vcat([lhs >> pp(' = ') >> rhs,\n",
     "               pp_jaxpr(eqn.params['jaxpr']).indent(2)])\n",
     "pp_rules[xla_call_p] = pprint_xla_call"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Finally, we'll tweak how xla_call is pretty-printed in jaxprs.\n",
+    "+ tags=[\"hide-input\"]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def pprint_xla_call(names: DefaultDict[Var, str], eqn: JaxprEqn) -> PPrint:\n",
+    "  lhs = pp(' '.join(var_str(names, v) for v in eqn.out_binders))\n",
+    "  params_without_jaxpr = {k:v for k, v in eqn.params.items() if k != 'jaxpr'}\n",
+    "  rhs = (pp(eqn.primitive.name) >> pp_params(params_without_jaxpr) >>\n",
+    "         pp(' '.join(names[x] if isinstance(x, Var) else str(x.val)\n",
+    "                     for x in eqn.inputs)))\n",
+    "  return vcat([lhs >> pp(' = ') >> rhs,\n",
+    "               pp_jaxpr(eqn.params['jaxpr']).indent(2)])\n",
+    "pp_rules[xla_call_p] = pprint_xla_call"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "-"
    ]
   },
   {
@@ -2840,9 +2864,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from weakref import ref, ReferenceType\n",
@@ -2867,7 +2889,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -3459,7 +3480,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -3612,10 +3632,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_end_of_cell_marker": 0,
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "def cond(pred, true_fn, false_fn, *operands):\n",
@@ -3668,9 +3685,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "def cond_impl(pred, *operands, true_jaxpr, false_jaxpr):\n",
@@ -3759,9 +3774,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "xs = np.array([1., 2., 3])\n",
@@ -3820,9 +3833,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "lines_to_next_cell": 1
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "def cond_abstract_eval(pred_type, *in_types, true_jaxpr, false_jaxpr):\n",
@@ -3863,7 +3874,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -3955,7 +3965,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -3996,7 +4005,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "lines_to_end_of_cell_marker": 0,
     "lines_to_next_cell": 1
    },
    "outputs": [],
@@ -4049,6 +4057,8 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
+    "lines_to_end_of_cell_marker": 0,
+    "lines_to_next_cell": 1,
     "tags": [
      "hide-input"
     ]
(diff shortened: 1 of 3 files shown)


shawwn:
HELL YEAH!

Go @mattjj! Thanks!

(I was just thinking how useful https://github.com/google/jax/issues/2521#issuecomment-604759386 has been, unrelatedly, so thanks for all your work.)

My only regret is that I can't read this PR in github, and I have an interview today so I'll have to wait. But the title is exciting.

shawwn:
Oh wait, there's an .md along with the notebook. I can read it on github!

shawwn:
Hmm. Did you just remove the three examples from the .md file? They were very nice; I would keep them.

It also seems like you may be force-pushing to the branch rather than writing separate commits, so I'm not quite sure how to go back and get the examples (at least from github's UI).

shawwn:
Hmm. I think I'll add comments here rather than https://github.com/google/jax/commit/4d1c0ff40aa87e1daa2d28d166cdf8c27726a94e#r54476265 since they seem to vanish with each force push:

> Before we turn to DCE, let's look at some examples.
> ... 3 examples cut ...

I think the examples are quite lovely, and I'd hate to see them get cut. But perhaps you could expand on "examples of what?" here -- for a casual reader who's skimming, I find it helpful to have a little bit of intro before diving in.

Also, I'm not sure if I was supposed to "start a review" or just leave comments here -- if me opening a review caused problems (I'm not sure what), feel free to ignore/close it.

mattjj:
> Hmm. Did you just remove the three examples from the .md file? They were very nice; I would keep them.

Ah yes, I did remove them, but really I _moved_ them, to a Google doc serving as the design doc (shared internally with the team). But that's not useful for external readers like you!

I'll see if I can copy the doc into a .md file without too much trouble...

mattjj:
Now there is a long PR message here! The examples from the autodidax.py file are now distilled at the end of it.

mattjj:
I don't think starting a review would cause any problems!

mattjj:
Now that the real implementation is up in #7631, I'll close this PR.

shoyer:
> An alternative, suggested by dougalm@, is that we allow policies to be tagged with a priority score, and when nesting we follow the policy with the higher priority score. That approach gives strictly more control to users and library writers alike, and works well for e.g. NumPy array API priorities.

Nit: I don't think this really worked out great for NumPy's `__array_priority__`. Users basically picked priorities > 1 at random without any sense of how they could interact.

NumPy's newer mechanism for overriding operations with NumPy arrays `__array_ufunc__` got rid of array priorities in favor of forcing users to explicitly choose how to handle every pair of array types. That would be a better model to emulate in my opinion (though I also think it's fine to just say "inner policy wins" in this case, since there is an ambiguous ordering).

shawwn (commented):
> > Hmm. Did you just remove the three examples from the .md file? They were very nice; I would keep them.
> 
> Ah yes, I did remove them, but really I _moved_ them, to a Google doc serving as the design doc (shared internally with the team). But that's not useful for external readers like you!
> 
> I'll see if I can copy the doc into a .md file without too much trouble...

Were you able to copy the design doc into a .md file somewhere? The examples were quite informative.

(As an aside, I don't think this autodidax PR should be closed. Autodidax is separate from JAX. It's much easier to read and understand, due to its simplicity. I'm a big fan of it, and I link people to it a lot when trying to explain the essence of JAX.)

PR #7519, opened 2021-08-05
Use trace_to_jaxpr_dynamic in variadic_reduction_jaxpr.

hawkinsp:
We want jit(..., inline=True) inlining to apply in reduction jaxprs.

Fixes some test failures when adding `jit` decorators to `jax.numpy` unary and binary operators.


mattjj (approved):
Thanks!

PR #7525, opened 2021-08-05
Use MainTrace payload mechanism in experimental callback tracer.

hawkinsp:
A change that adds jit() decorators on a number of standard library functions was triggering incorrect cache hits for these tests. This is because the payload fields of the MainTrace were not being included in __hash__() and __eq__().


hawkinsp:
@sschoenholz as an FYI

mattjj (approved):
Thanks for figuring this out. It was probably my bug to begin with :P

PR #7530, opened 2021-08-06
Update maml.ipynb

tigerneil:
typo fixed.


gnecula:
Thank you for the fix. The same fix must be applied to `maml.md`, there are instructions in the error message for the line_and_typecheck test for how to ensure that. You may also check the [documentation](https://jax.readthedocs.io/en/latest/developer.html#update-notebooks)

PR #7536, opened 2021-08-06
Don't try to use persistent compilation cache when running CPU computations.

skye:


jax/interpreters/xla.py
```diff
@@ -67,14 +67,15 @@
 def compile_or_get_cached(backend, computation, compile_options):
     # Avoid import cycle between jax and jax.experimental
     from jax.experimental.compilation_cache import compilation_cache as cc
-    if cc.is_initialized():
-        cached_executable = cc.get_executable(computation, compile_options)
+    # Persistent compilation cache only implemented on TPU. Avoid errors on CPU.
+    if cc.is_initialized() and backend.platform != 'cpu':
```
zhangqiaorjc, line 71: what about gpu?

can we narrow this to only supported platform ?

trevorcai: GPU definitely fails to serialize the executable.

skye: I intentionally did it this way so it does raise an error when using GPU, so people know it doesn't work. I think we may eventually wanna intercept all cache exceptions and print a warning but continue running, but for now I think it's better to fail. I could be convinced to just check for TPU though.

colemanliyah: + 1 could we do backend.platform == 'tpu' ?

colemanliyah (approved):

jax/interpreters/xla.py
skye: Changed to platform == tpu.

zhangqiaorjc (approved):

Issue #7533, opened 2021-08-06
jit changes behaviour of sparse.coo_matmat with transposed argument


dimasad:
JIT gives erroneous results for the multiplication of a sparse matrix with a transposed dense matrix.

```python

import jax
import jax.numpy as jnp
import jax.experimental.sparse as jsparse

d = jnp.r_[0.1, 2., 3., -2]
x = jax.random.normal(jax.random.PRNGKey(0), (3, 3))


def ok(d, x):
    i = jnp.r_[0, 0, 1, 2]
    j = jnp.r_[0, 2, 0, 0]
    shape = (3, 3)
    return jsparse.coo_matmat(d, i, j, x, shape=shape)


def NOT_ok(d, x):
    i = jnp.r_[0, 0, 1, 2]
    j = jnp.r_[0, 2, 0, 0]
    shape = (3, 3)
    return jsparse.coo_matmat(d, i, j, x.T, shape=shape)

e_ok = ok(d, x) - jax.jit(ok)(d, x)
e_not_ok = NOT_ok(d, x) - jax.jit(NOT_ok)(d, x)

print('ok: error = \n', e_ok)
print('\nNOT ok: error = \n',e_not_ok)
## OUTPUT
#ok: error = 
# [[0. 0. 0.]
# [0. 0. 0.]
# [0. 0. 0.]]
#
#NOT ok: error = 
# [[ 0.97765154  0.7773348  -0.0488826 ]
# [ 0.         -3.0031524  -1.4664774 ]
# [ 0.          2.0021017   0.9776516 ]]
```

jakevdp:
Are you seeing this only on a GPU backend?

jakevdp:
I don't reproduce the issue on CPU; I suspect it has to do with the lowering to cusparse. We need to change the layout definition here for a transposed matrix: https://github.com/google/jax/blob/c94f41290cc5c28fba059e172279994107c3103e/jaxlib/cusparse.py#L285

But I'm not certain how to tell at this level whether the input buffer is transposed... perhaps XLA operators somehow expose their strides?

dimasad:
Yes, I'm seeing this on a GPU backend only.

jakevdp:
I took a closer look at this and realized that in the transposed version, you're computing the matrix product of a (3,3) matrix and a (4, 3) matrix, which should be an error! The issue is poor input validation, which should be fixed by #7557.

[...]

PR #7540, opened 2021-08-06
[sparse] implementing transposition rule for `sparse.coo_matmat`.

dimasad:
The transposition rule for `sparse.coo_matmat` is very straightforward, and the code is very similar to `_coo_matvec_transpose`, already implemented in `jax.experimental.sparse.ops.py`.

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -549,6 +549,18 @@ def _coo_matmat_jvp_rule(primals_in, tangents_in, **params):
   return primals_out, tangents_out
 ad.primitive_jvps[coo_matmat_p] = _coo_matmat_jvp_rule
 
+def _coo_matmat_transpose(ct, data, row, col, B, *, shape, transpose):
+  assert not ad.is_undefined_primal(row)
+  assert not ad.is_undefined_primal(col)
+    
+  if ad.is_undefined_primal(B):
+    B_ct = coo_matmat(data, row, col, ct, shape=shape, transpose=not transpose)
+    return data, row, col, B_ct
+  else:
+    data_ct = jnp.sum(ct[row]*B[col], axis=1)
+    return data_ct, row, col, B
+
+ad.primitive_transposes[coo_matmat_p] = _coo_matmat_transpose
 
 #----------------------------------------------------------------------
 # BCOO primitives: batched extension of COO.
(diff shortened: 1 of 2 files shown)


dimasad:
@googlebot I signed it!

jakevdp:
Thanks for this! Just FYI, the transpose rule is a different sense of the word "transpose" than in the issue in #7533: transpose rules are used to implement reverse-mode autodiff (vjp). The reason I stopped before implementing it is because it became clear that batching was more important (e.g. you can't use `jacfwd` or `jacrev` to test autodiff without batching being implemented) so I changed my focus to developing the batched COO format (BCOO in that file) which now has batching & full autodiff implemented.

If you're interested in contributing this transpose rule, we'd also need to add a test that exercises it similar to coo_matvec's autodiff test here: https://github.com/google/jax/blob/38cb2b530aaecc74443de373f27a0ee143b6166d/tests/sparse_test.py#L319-L363

dimasad:
> Thanks for this! Just FYI, the transpose rule is a different sense of the word "transpose" than in the issue in #7533: transpose rules are used to implement reverse-mode autodiff (vjp). 

Yes, I'm aware, of this, I implemented this code because I was using `jax.grad` to perform reverse differentiation of a cost function that used a matrix-matrix multiplication like in #7533. I was able to work around the bug by simply transposing the whole expression twice, but for reverse differentiation I used this code I just contributed. I opened the pull request because I think it might be useful to more people.

I'll implement a few tests like the `matvec` counterparts, perhaps Monday.

dimasad:
> If you're interested in contributing this transpose rule, we'd also need to add a test that exercises it similar to coo_matvec's autodiff test here:
> 
> https://github.com/google/jax/blob/38cb2b530aaecc74443de373f27a0ee143b6166d/tests/sparse_test.py#L319-L363
> 
> The "reverse mode" sections there are where the transpose rule will be called.

Added test coverage of the implemented code with minimal changes to the existing code. By what was written in the TODO comments I just updated the `matvec_ad` test to test `matmul` in a way that covers both `coo_matvec` and `coo_matmat`. If separate tests are wanted I can just change it. All updated tests passed here.

jakevdp:
Looking good - you can address the test failures by modifying the tolerances.

Once you have that fixed, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Then we can get this in!

jakevdp:
It looks like `coo_matvec_transpose` has been added in the meantime. Sorry we weren't able to merge this contribution!

PR #7541, opened 2021-08-06
[sparse] correctly handle units in sparsify argspecs

jakevdp:
With #7537, this is the last change required to seamlessly sparsify a jaxopt logistic regression

### DIFF
--- a/jax/experimental/sparse/transform.py
+++ b/jax/experimental/sparse/transform.py
@@ -68,7 +68,6 @@ from jax.experimental.sparse import BCOO
 sparse_rules : Dict[core.Primitive, Callable] = {}
 
 Array = Any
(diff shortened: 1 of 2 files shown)


froystig (approved):
One way we end up with unit values in the wild is by partial evaluation of higher-order primitives like `cond` and `scan`, but a test that relies on linearizing `scan` only in order to test unit handling seems needlessly indirect. The simple test in this PR seems good.

What we could consider separately (for a later PR) is testing the composition of `sparsify` with several other transformations, and in particular sparsifying with both AD and `scan`. I would guess that we're encountering something along those lines when sparsifying jaxopt.

jakevdp:
Good ideas! I'll take a look at that next week

Issue #7546, opened 2021-08-08
NoneType coming from gradient function when using transformed parameters


peterroelants:
I have been trying to create constrained parameters, for example a scale parameter that is strictly positive, to be used for example by a normal distribution. Inspired slightly by [`tfp.util.TransformedVariable`](https://www.tensorflow.org/probability/api_docs/python/tfp/util/TransformedVariable).

I have been trying to use `PyTrees`(https://jax.readthedocs.io/en/latest/pytrees.html#extending-pytrees) to create parameter classes that store some unconstrained representation, and return the constraint parameter upon access (see the example below). To create the parameter object, I want to pass in the desired constrained value, transform it to an unconstrained representation and store the latter. However, for some reason, this inverse conversion, from constrained to unconstrained is causing issues.
I hope the example below will provide more clarity on the issue:

```python
import numpy as np
import jax
import jax.numpy as jnp

from jax.tree_util import register_pytree_node_class


def softplus_inv(x):
    """Inverse to softplus"""
    return jnp.log(jnp.exp(x) - 1.)


@register_pytree_node_class
class NormParams:
    """
    Represent parameters of a normal distribution. Ensures scale is strictly positive.
    """
    def __init__(self, loc, scale):
        self.loc = loc
        self._scale = float(softplus_inv(scale))  # ! Seems to be the origin of the issue !
#         self._scale = softplus_inv(scale)  # ! Same issue !
        
    @property
    def scale(self):
        """Make sure `scale` parameter is positive."""
        return jax.nn.softplus(self._scale)
    
    def __repr__(self):
        return "NormParams(loc={}, scale={})".format(self.loc, self.scale)
    
    def tree_flatten(self):
        children = (self.loc, self.scale)
        aux_data = None
        return (children, aux_data)

    @classmethod
    def tree_unflatten(cls, aux_data, children):
        return cls(*children)
    

# Define parameters and inputs for testing
params_test = NormParams(
    loc=1.,
    scale=1.,
)
print(f"params_test: {params_test!r}")
x_test = np.random.randn(9)



def loss(params: NormParams, x):
    """Negative log-likelihood for the normal distribution."""
    return -jax.scipy.stats.norm.logpdf(x, loc=params.loc, scale=params.scale).sum()


print("loss: ", loss(params_test, x_test))


# Get the grad function of the loss wrt the parameters
grad_fun = jax.grad(loss, argnums=0)
g = grad_fun(params_test, x_test)  # ! FAILS HERE !
[...]

PR #7551, opened 2021-08-09
Do not unflatten trees with None values in grad.

mkunesch:
When checking the data type of the dynamic arguments in jax.value_and_grad the
PyTree is unflattened with `None` (the output of `_check_input_dtype_grad`) as
value for each leaf. This causes an issue if a custom PyTree does not accept
None as a value for the leaves (issue #7546) even though the tree that is
returned from the data type check is never used.

This commit solves this issue by making `_check_input_dtype_grad` an identity
function if the checks pass. That way it is guaranteed that the PyTree can be
unflattened.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -865,7 +865,8 @@ def value_and_grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
 
     f = lu.wrap_init(fun, kwargs)
     f_partial, dyn_args = argnums_partial(f, argnums, args)
-    tree_map(partial(_check_input_dtype_grad, holomorphic, allow_int), dyn_args)
+    for leaf in tree_leaves(dyn_args):
+      _check_input_dtype_grad(holomorphic, allow_int, leaf)
     if not has_aux:
       ans, vjp_py = _vjp(f_partial, *dyn_args, reduce_axes=reduce_axes)
     else:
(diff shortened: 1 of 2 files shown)


jakevdp:
Thanks! Could you add a simple test that fails without this change?

mkunesch:
Totally, happy to do that! Just to check though: is jax even meant to support custom nodes that cannot be unflattened with arbitrary children (e.g. `None`s in the case of issue #7546)? If PyTrees are not allow to care about which elements they hold (as would usually be the case) this PR isn't necessary. Thanks a lot!

mkunesch:
I've added a minimal test now which breaks without the change. The error only arises if a custom node has an unflatten function that does not work if one of the children is `None` (like in issue #7546). As I said, as far as I know this is not the intended us of custom nodes so if jax isn't meant to support nodes that don't accept `None` the PR isn't necessary (and an extra line in the docs could help instead since a user might not expect `grad` to unflatten a tree with `None`).

mkunesch:
Thanks a lot!
(I just changed to `tree_leaves` following your suggestion - thanks a lot!)

jax/_src/api.py
```diff
@@ -865,7 +865,8 @@ def value_and_grad_f(*args, **kwargs):
 
     f = lu.wrap_init(fun, kwargs)
     f_partial, dyn_args = argnums_partial(f, argnums, args)
-    tree_map(partial(_check_input_dtype_grad, holomorphic, allow_int), dyn_args)
+    for leaf in tree_leaves(dyn_args):
+      _check_input_dtype_grad(holomorphic, allow_int, leaf)
```
jakevdp, line 869: Nice!

jakevdp (approved):

PR #7555, opened 2021-08-09
Strip debug_info and jaxpr_stack from MainTrace instances used as C++…

hawkinsp:
… JIT cache keys.

Without this fix, we created reference count cycles via the C++ JIT cache.


mattjj (approved):
LGTM, this is a bit of a workaround and it's on me to submit a cleanup after.

Issue #7175, opened 2021-07-02
Common Loss Functions written natively in JAX?


derrickxli:
Could we create a new subpackage (or find somewhere to put this) for native loss functions written in JAX? The issue is that when I try to use loss functions from say `sk_metrics`, the function will not be able to take advantage of autodiff from `jax.grad` because of the incompatibility of tracing device arrays through vanilla numpy functions.

If there were a suite of common native written loss functions that could use `jax.loss`, then we could have the ability to use loss functions written from a standardized, robust library while also taking advantage of autodiff. This way, users will not create their own error prone loss functions.

jakevdp:
Thanks for the question! I don't know of any centralized effort at that yet, but one example where something like this exists is within the optax project: https://github.com/deepmind/optax/blob/master/optax/_src/loss.py

mblondel:
We also have some here https://github.com/google/jaxopt/blob/main/jaxopt/_src/loss.py

froystig:
As @tomhennigan put it in https://github.com/google/jax/pull/7575#issuecomment-896556986:

> some folks in the community have discussed this before, and felt that losses should be provided by libraries on top of JAX rather than be a part of JAX itself (to keep JAX focussed and lightweight).

There are loss definitions available in optax, jaxopt, RLAX, and other tested and maintained libraries, all of which are compatible with JAX's autodiff transformations.

I'll close this issue since it is out of jax's scope for the time being. We can still continue the discussion, of course.

mwitiderrick:
Check this JAX loss functions article https://www.machinelearningnuggets.com/jax-loss-functions/

PR #7575, opened 2021-08-10
Implement L1 loss and L2 loss

MaanasArora:
Starting to implement #7175.

I will commit to add more losses.

Thank you for reviewing!

### DIFF
--- a/jax/_src/ops/losses.py
+++ b/jax/_src/ops/losses.py
@@ -1,0 +1,33 @@
+from jax._src.numpy import lax_numpy as jnp
+
+def l2_loss(pred, targets=None):
+    '''L2 (mean squared) loss.
+    
+    Args:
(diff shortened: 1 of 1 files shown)


tomhennigan:
Hi @MaanasArora , I think some folks in the community have discussed this before, and felt that losses should be provided by libraries on top of JAX rather than be a part of JAX itself (to keep JAX focussed and lightweight).

One popular implementation I am aware of is in [RLAX `losses.py`](https://github.com/deepmind/rlax/blob/75784078e77a611674f5298875112c66d57c93d2/rlax/_src/losses.py) which supports l2 and some other losses. Perhaps it would be worth discussing the other losses you want to add to that library?

MaanasArora:
I see, thanks for informing! I will close the pull request then and look around in other libraries.

PR #7595, opened 2021-08-11
Improve semantics of out-of-bounds indices in gather and scatter

hawkinsp:
Previously the behavior of out-of-bounds indexed accesses was not well-defined. Gathers clamped indices to be in-bounds, whereas out-of-bounds scatters were dropped. Additionally, the gradients of out-of-bounds indices were incorrect, since scatter (the transpose of gather) had different semantics to gather.

This PR adds a new `mode` option to `gather` that allows the user to select one of three modes:
* `"clip"`: clamp out-of-bounds indices into range
* `"fill"`: fill out-of-bounds gathered windows with a constant value, typically NaN or 0.
* `"promise_in_bounds"`: preserve the current behavior, which is `"clip"` for gathers.

The PR also adds the same `mode` option to `scatter`. The modes for `scatter` are identical, but `"fill"` is an alias for `"drop"`, meaning "drop out of bounds updates".

The PR gives two main benefits:
* if we change the default for `gather` to `"fill"`, then we can help users catch accidental out-of-bounds indexing. If no `fill_value` is provided, `gather` defaults to one of NaN, or an extremal integer value. This should indicate the presence of out-of-bounds indices.
*  Since we now have both `"clip"` and `"fill"`/`"drop"` modes for both `gather` and `scatter`, we can now define a correct transpose for out-of-bounds indices. The transpose of a `"clip"` gather is a `"clip"` scatter, and the transpose of a `"fill"` gather is a `"drop"` scatter.

We probably need to add the same semantics to `dynamic_slice` and `dynamic_update_slice` in a future PR.

This PR does not yet change the default mode, since that is likely to break users. For now, it is set to `"promise_in_bounds"`, which preserves the current behavior.

### DIFF
--- a/docs/jax.ops.rst
+++ b/docs/jax.ops.rst
@@ -21,6 +21,7 @@ follows (where ``idx`` is a NumPy index expression).
 =========================  ===================================================
 Alternate syntax           Equivalent in-place expression
 =========================  ===================================================
+``x.at[idx].get(y)``       ``x[idx]``
 ``x.at[idx].set(y)``       ``x[idx] = y``
 ``x.at[idx].add(y)``       ``x[idx] += y``
 ``x.at[idx].multiply(y)``  ``x[idx] *= y``
@@ -32,14 +33,22 @@ Alternate syntax           Equivalent in-place expression
 
 None of these expressions modify the original `x`; instead they return
 a modified copy of `x`. However, inside a :py:func:`jit` compiled function,
-expressions like ``x = x.at[idx].set(y)`` are guaranteed to be applied inplace.
+expressions like ``x = x.at[idx].set(y)`` are guaranteed to be applied in-place.
+
+By default, JAX assumes that all indices are in-bounds. There is experimental
+support for giving more precise semantics to out-of-bounds indexed accesses,
+via the ``mode`` parameter to functions such as ``get`` and ``set``. Valid
+values for ``mode`` include ``"clip"``, which means that out-of-bounds indices
+will be clamped into range, and ``"fill"``/``"drop"``, which are aliases and
+mean that out-of-bounds reads will be filled with a scalar ``fill_value``,
+and out-of-bounds writes will be discarded.
 
 
 Indexed update functions (deprecated)
 -------------------------------------
 
 The following functions are aliases for the ``x.at[idx].set(y)``
-style operators. Prefer to use the ``x.at[idx]`` operators instead.
+style operators. Use the ``x.at[idx]`` operators instead.
 
 .. autosummary::
   :toctree: _autosummary
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -16,7 +16,8 @@
 # pytype: skip-file
 
 import builtins
-from enum import IntEnum
+import enum
+from enum import Enum, IntEnum
 import functools
 from functools import partial
 import itertools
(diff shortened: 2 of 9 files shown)


jax/_src/lax/lax.py
```diff
@@ -966,12 +966,45 @@ class GatherDimensionNumbers(NamedTuple):
...
+
+  @staticmethod
+  def from_any(s: Optional[Union[str, 'GatherScatterMode']]):
+    if s == "clip" or s == GatherScatterMode.CLIP:
+      return GatherScatterMode.CLIP
+    if s == "fill" or s == "drop" or s == GatherScatterMode.FILL_OR_DROP:
+      return GatherScatterMode.FILL_OR_DROP
+    if (s is None or s == "promise_in_bounds" or
+        s == GatherScatterMode.PROMISE_IN_BOUNDS):
+      return GatherScatterMode.PROMISE_IN_BOUNDS
```
shoyer, line 995: `isintance` on the enum might help simplify this:
```suggestion
    if isinstance(s, GatherScatterMode):
      return s
    if s == "clip":
      return GatherScatterMode.CLIP
    if s == "fill" or s == "drop":
      return GatherScatterMode.FILL_OR_DROP
    if s is None or s == "promise_in_bounds":
      return GatherScatterMode.PROMISE_IN_BOUNDS
```

jax/_src/lax/lax.py
```diff
@@ -966,12 +966,45 @@ class GatherDimensionNumbers(NamedTuple):
   start_index_map: Sequence[int]
 
 
+class GatherScatterMode(IntEnum):
```
shoyer, line 969: Is there any particular reason for using `IntEnum` here? Generally `Enum` is a better choice for non-legacy code, which does not need to support integer values.

shoyer (commented):
partial review: I did not look at the implemenation, only the inferface

jax/_src/lax/lax.py

jax/_src/lax/lax.py
hawkinsp: Changed to `Enum`. Thanks for the tip!

jax/_src/lax/lax.py
```diff
@@ -966,12 +966,46 @@ class GatherDimensionNumbers(NamedTuple):
   start_index_map: Sequence[int]
 
 
+class GatherScatterMode(Enum):
+  # Indices will be clamped to the nearest in-range value, i.e., such that the
+  # entire window to be gathered is in-range.
+  CLIP = 0
```
mattjj, line 972: Do you like using `auto`, as in `from enum import auto` and `CLIP = auto()` etc?

(My favorite way to write enums is like namedtuples, as in `GatherScatterMode = enum.Enum('GatherScatterMode', ['CLIP', 'FILL_OR_DROP', 'PROMISE_IN_BOUNDS'])`. But then we don't have space for the comments, or for adding the `from_any` method, so probably not the best option here. Just thought I'd share!)

mattjj (approved):
Great change, and writeup!

docs/jax.ops.rst
```diff
@@ -21,6 +21,7 @@ follows (where ``idx`` is a NumPy index expression).
 =========================  ===================================================
 Alternate syntax           Equivalent in-place expression
 =========================  ===================================================
+``x.at[idx].get(y)``       ``x[idx]``
```
soraros, line 24: I think this should be `x.at[idx].get()`?

hawkinsp: Thanks! Yes. #8000 fixes it.

PR #7597, opened 2021-08-11
Add a bot that adds kokoro:run label when the PR is approved automati…

yashk2810:
…cally.

We can also add pull-ready label via this bot. But I want to test kokoro:run first.

If this doesn't work, I'll revert the changes.

### DIFF
--- a/.github/workflows/bot-label.yaml
+++ b/.github/workflows/bot-label.yaml
@@ -10,17 +10,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      # - uses: actions/github-script@v4
-      #   with:
-      #     script: |
-      #       github.issues.addLabels({
-      #         issue_number: context.issue.number,
-      #         owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


froystig (commented):
Could this be simplified if we used github's built-in [labeler action](https://github.com/marketplace/actions/labeler)?

yashk2810:
There's this too: https://github.com/marketplace/actions/label-approved-pull-requests. Can you use this?

yashk2810:
So I used the github script action and modified the bot. I think this should work.

froystig (approved):
Approving to test.

Separately: can we rename the workflow config file to something along the lines of `pull-request-review.yaml`?

froystig (approved):
approving just to test, again.

yashk2810:
Test comment

froystig (changes requested):
It seems review approval is a tricky event to listen for. I also think that the `pull ready` label is a more appropriate event. Let's try gating on that instead?

PR #7605, opened 2021-08-12
Remove opensource CS link since JAX is not indexed

yashk2810:


hawkinsp (approved):
Hopefully we can revive it soon.

PR #7606, opened 2021-08-12
PEP484-compatible export of names in jax & its public submodules

### LINKED ISSUE #7570: Consider defining public API explicitly
Mypy requires public API of a package being exported using either `__all__` or `import ... as ...` syntax.

```bash
python3 -m venv env
source env/bin/activate
pip install -U pip mypy jax
python -m mypy --install-types --non-interactive --strict -c "from jax import numpy as jnp; xs = jnp.zeros(1)"

<string>:1: error: Module has no attribute "zeros"
Found 1 error in 1 file (checked 1 source file)
```

Currently, projects that are using jax need to set`--no-implicit-reexport=False`.

There are some references on that matter.

- [PEP-0484](https://www.python.org/dev/peps/pep-0484/#stub-files)
  > - Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form. 
  > - However, as an exception to the previous bullet, all objects imported into a stub using from ... import * are considered exported. 
- [MyPy docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport)
  > `--no-implicit-reexport` always treated as enabled for stub files.

Is it possible to add a little boilerplate to `__init__.py` to make it working? :-)

jakevdp:
Packages like mypy, pylance, vscode, etc. require explicit export of names in order to work correctly for type checking, autocomplete, and other features. According to PEP 484 there are two ways to do this:

1. use `__all__ = ["name1", "name2", ...]` to specify exported names
2. use `import xyz as xyz` (where the two names must be identical) to specify imports to be considered part of the namespace

Given the fact that our stub files like https://github.com/google/jax/blob/main/jax/numpy/__init__.py already explicitly import the names that should be in the namespace, I think option 2 is probably cleaner in most cases (it avoids the problem of having to keep two separate lists in sync).

This PR adopts this explicit export strategy throughout JAX's public-facing submodules.

Fixes #7570 & #7762


hawkinsp (approved):
I agree, this seems like the better of the two options.

PR #7607, opened 2021-08-12
Add mypy export of jax.numpy namespace

### LINKED ISSUE #7570: Consider defining public API explicitly

manifest:
Another option to fix #7570


jakevdp:
I prefer the approach in  #7570, because in that version you can immediately tell that the two lists match.

Issue #5364, opened 2021-01-11
Possible bug in jax.linearize()


DiffeoInvariant:
I'm using `jax.linearize()` to compute pushforwards, and I'm getting the following error:
```
ValueError: linearized function called on tangent values inconsistent with the original primal values: got ConcreteArray([[-0.65073081 -0.0755682 ]
        [ 0.07366651 -1.6313327 ]]) for primal aval ConcreteArray([[ 1.2067288  -0.5176796 ]
        [ 0.4400338   0.39490446]])
```
As far as I can tell, this exception is only documented in the code that raises it (in the form of the exception itself, not any comments or any other documentation), and since the tangent and primal avals are the same shape, I'm not sure what's going wrong here. Is this a bug in jax.linearize(), or have I just made a mistake in my code somewhere? If this isn't a bug in jax.linearize(), then why is this exception being raised? (The code indicates that it comes from a `TypeError` when doing `jax.core.lattice_join(primal_aval.at_least_vspace(),tangent_aval)`)

froystig:
It's possible that the array types differ in ways not shown in the exception string formatting. In that case, it might be a bug in the system, or it might be a reason to improve the error message. Do you have a minimal code example that reproduces the issue?

DiffeoInvariant:
Unfortunately no, it's not possible to provide a minimal example*, but I can answer any questions you may have, like "what type are the inputs" or the such. The functions I'm differentiating can accept a wrapper around `DeviceArray`s (using `register_pytree_node_class()`) as well as a regular `DeviceArray`, but in this case, I verified that all of the inputs are `jax.interpreters.xla.DeviceArray`s. I suspect that there may be an issue with `jax.linearize()` as opposed to a bad error message because the shape appears to match the expected shape (and, at least with other JAX functions, an incorrect shape usually produces a more direct error message), but that definitely doesn't rule anything out.

*I'm integrating JAX with Pyadjoint, a library for computing derivatives of finite element code, and this error only happens in one of the more complicated tests I have, so there would probably just be too much code to share to give you a good idea what's going on in a reasonable amount of time. If you'd like access to the code to see what's going on, just ping me at zane(dot)jakobs(at)colorado(dot)edu.

DiffeoInvariant:
One thing that may be useful: the I'm linearizing about these inputs:
```
 [DeviceArray([[ 1.2067288 , -0.5176796 ],
[...]

PR #7619, opened 2021-08-13
Add dtype to string representation of ConcreteArray.

mkunesch:
The string representation of ConcreteArray did not include the data type of the
wrapped value. This makes it harder to spot the reason for errors arising from
inconsistent values (issue #5364). This commit adds the data type to the string
representation of ConcreteArray.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -1190,7 +1190,7 @@ class ConcreteArray(ShapedArray):
       raise TypeError(self, other)
 
   def str_short(self) -> str:
-    return str(self.val)
+    return f'{self.val}, dtype={self.dtype.name}'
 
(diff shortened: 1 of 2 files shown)


mkunesch:
(an alternative fix for issue #5364 would be to add more information to that specific error but I thought that this solution would be more general and would also be consistent with other array types in jax whose string representation includes dtype)

LenaMartens:
LGTM!

LenaMartens (approved):

yashk2810:
FYI: You just need to approve the change :)

The google-ml-butler will apply the labels (pull ready and kokoro:run) for you :)

PR #7620, opened 2021-08-13
Add nocuda jaxlib wheels to the index as well

yashk2810:


yashk2810:
https://storage.googleapis.com/jax-releases/jax_releases.html is updated to included nocuda wheels as well.

skye (approved):

PR #7626, opened 2021-08-13
always use same object for vmap temp axis name

### LINKED ISSUE #7621: Memory leak when computing the hessian of a jitted function
Please:

```python
import jax
import jax.numpy as jnp

f = lambda x: jnp.square(x).mean()
jf = jax.jit(f)

x = jax.random.uniform(jax.random.PRNGKey(0), shape=(8, 4))

# Start monitoring memory with htop
while True:
    try:
        # let this run for a bit, notice memory usage doesn't increase
        y = jax.hessian(f)(x)
    except KeyboardInterrupt:
        pass

while True:
    # let this run, notice memory usage starts increasing immediately
    # also, the memory usage doesn't appear to stop increasing ever
    y = jax.hessian(jf)(x)
```
Also, note that the equivalent code with `jax.grad` (or `jax.jacfwd` for multidimensional functions) replacing `jax.hessian` *does not* cause a memory leak.

As a result of this leak, my python process ends up getting killed by my OS (due to OOM), so I don't ever see any python or jax traces.

mattjj:
Fixes #7621.

We don't need to create a fresh temporary axis name for every `vmap` call which is not given an axis name, since this axis name can't be used in collectives. Instead, we can always use the same object for the name. In particular that means we don't get the cache misses which caused #7621.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2949,6 +2949,20 @@ class APITest(jtu.JaxTestCase):
     with jax.checking_leaks():
       _ = jax.grad(loss)(A, x)  # doesn't crash
 
+  def test_vmap_caching(self):
+    # https://github.com/google/jax/issues/7621
+
+    f = lambda x: jnp.square(x).mean()
+    jf = jax.jit(f)
+    x = jax.random.uniform(jax.random.PRNGKey(0), shape=(8, 4))
+
+    with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
+      jax.hessian(jf)(x).block_until_ready()
+      jax.hessian(jf)(x).block_until_ready()
+      jax.hessian(jf)(x).block_until_ready()
+
+    self.assertEqual(count[0], 2)
+
 
 class RematTest(jtu.JaxTestCase):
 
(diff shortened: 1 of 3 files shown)


tests/api_test.py
```diff
@@ -2949,6 +2949,20 @@ def loss(A, x):
...
+    f = lambda x: jnp.square(x).mean()
+    jf = jax.jit(f)
+    x = jax.random.uniform(jax.random.PRNGKey(0), shape=(8, 4))
+
+    with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
+      jax.hessian(jf)(x).block_until_ready()
+      jax.hessian(jf)(x).block_until_ready()
+      jax.hessian(jf)(x).block_until_ready()
+
+    self.assertEqual(count[0], 2)
```
mattjj, line 2964: I should probably have left a comment about this: the reason why the expected value is 2 is that this is a grad-of-jit type situation (really hessian-of-jit), so we have one compiled computation for the forward pass and one for the backward pass. If this were jit-of-grad we'd only get one compilation.

Issue #7622, opened 2021-08-13
Trying to get a PRNGKey on cpu results in strange gpu memory allocation


ysngshn:
I am trying to get a `jax.random.PRNGKey` on cpu to manage the source of randomness for some host function. If I understand correctly, one should do something like:
```python
import jax
key = jax.jit(jax.random.PRNGKey, backend='cpu')(0)
```
This indeed returns a key on cpu, and subsequent operations like `jax.random.normal` also stays on cpu, so far so good!

However, I just realized that when executing the line `key = jax.jit(jax.random.PRNGKey, backend='cpu')(0)` in a python interpreter, it results in allocation of ~140 MB memory on GPU, which is unexpected. This is not the 90% memory pre-allocation, which would occupy > 10 GB GPU memory.

Any ideas on why this happens? Is this somehow an expected behavior, or maybe there are better ways to achieve what I need? (I also need to run functions on the gpu so setting JAX to use cpu globally does not solve the issue.)

mattjj:
Hrm, I don't know why this would happen, though I would guess it's an effect of backend-initialization. That is, when you run the first JAX operation (_not_ when JAX is imported), all available backends are initialized. That means that if you haven't already initialized the backends, the first operation on CPU will cause the GPU backend (if available) to be initialized.

Is this 140MB allocation a problem somehow, or are you just curious about it?

ysngshn:
Thanks for the quick reply! This is mainly out of curiosity. Backend initialization indeed seems a plausible explanation. Ideally I guess one wouldn't expect gpu backend be initialized when working only on the cpu backend. Also, the default 90% allocation is not triggered, wouldn't this potentially generate some issues?

mattjj:
> Ideally I guess one wouldn't expect gpu backend be initialized when working only on the cpu backend.

We might be able to make this initialization lazier... @skye do you know if there are any reasons not to do that?

> Also, the default 90% allocation is not triggered, wouldn't this potentially generate some issues?

Maybe, but I don't know of any it's caused yet! Any guesses?

ysngshn:
> Maybe, but I don't know of any it's caused yet! Any guesses?
I do remember having to disable gpu preallocation to avoid some cuda oom error, which is also mentioned in the JAX documentation, but that was for some other code (which also uses jax.random for some dataloader function on cpu) which might be totally unrelated. 

Anyway I feel like this is something interesting to point out. Thanks a lot for the quick replies!

skye:
[...]

PR #7627, opened 2021-08-14
Add xla_bridge_test.GetBackendTest

skye:
This is in preparation for refactoring the get_backend logic, see the TODOs in the test.

PR #7628, opened 2021-08-14
Don't initialize all backends when get_backend() is called with a specific backend.

skye:
This change refactors the get_backend() logic to also handle some edge cases more clearly, see the unit test changes.

One side-effect of this change not reflected in the test is that there'll be INFO logging for backend factories that return None or backends with no devices. I'm not sure if this is a good or bad thing, but I don't think it should make a huge difference either way.

Partially addresses https://github.com/google/jax/issues/7622.

### DIFF
--- a/jax/lib/xla_bridge.py
+++ b/jax/lib/xla_bridge.py
@@ -193,53 +193,68 @@ register_backend_factory(
   'tpu', partial(tpu_client_timer_callback, timer_secs=60.0), priority=300)
 
 _default_backend = None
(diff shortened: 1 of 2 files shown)


skye:
This is based on https://github.com/google/jax/pull/7627, so I recommend looking at just the second commit until that goes in.

mattjj (approved):
This is great!

hawkinsp:
Close this PR? I think we agreed not to do this.

PR #7631, opened 2021-08-15
user-customizable gradient checkpointing strategies

mattjj:
This PR is the real implementation of the Autodidax prototype in #7517. It actually stayed quite close to the prototype!

See the PR message for #7517 for a draft design doc. We'll add that to the design notes in a follow-up PR.

This more general remat code effectively sets up a parallel implementation for standard remat (corresponding to a policy of `lambda *_, **__: False)`). We can delete the previous implementation, but for simplicity we'll leave that to a follow-up PR.

The DCE implementation here is only applied to jaxprs formed via partial evaluation, i.e. those which are formed by AD and assumed to be effect-free. For that reason we can punt on thinking about effects for now. (If/when we apply it to computations we lower to XLA, [for the purpose of pruning inputs](https://github.com/google/jax/blob/9b62fd80d0e6068c0bf05d8fbb184f507b51a087/jax/interpreters/xla.py#L813), we shouldn't prune effectful primitive applications, cc @zhangqiaorjc.)

The remat implementation also pretends effectful primitives don't exist (like a lot of JAX does...).

Some last things to do:
- [x] deflake
- [ ] maybe delete the old DCE implementation and just use the new one (could leave this for follow-up)

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -29,6 +29,7 @@ import itertools as it
 import sys
 import threading
 import weakref
+import types
 from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
                     Sequence, Tuple, TypeVar, Union, overload)
 from warnings import warn
@@ -36,6 +37,7 @@ from warnings import warn
 import numpy as np
 from contextlib import contextmanager, ExitStack
 
+import jax
 from .. import core
 from .. import lib
 from .. import linear_util as lu
@@ -2686,11 +2688,12 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
       ``pmap``, CSE can defeat the purpose of this decorator. But in some
       settings, like when used inside a ``scan``, this CSE prevention mechanism
       is unnecessary, in which case ``prevent_cse`` can be set to False.
-    policy: Optional, callable which takes as input a type-level specification
-      of a first-order primitive application and returns a boolean indicating
-      whether the corresponding output value(s) can be saved as a residual (or,
-      if not, instead must be recomputed in the (co)tangent computation).
-      Experimental.
+    policy: This is an experimental feature and the API is likely to change.
+      Optional callable, one of the attributes of ``jax.checkpoint_policies``,
+      which takes as input a type-level specification of a first-order primitive
+      application and returns a boolean indicating whether the corresponding
+      output value(s) can be saved as a residual (or, if not, instead must be
+      recomputed in the (co)tangent computation).
 
   Returns:
     A function (callable) with the same input/output behavior as ``fun`` but
@@ -2747,6 +2750,12 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
     return tree_unflatten(out_tree(), out_flat)
   return fun_remat
 remat = checkpoint
+
+checkpoint_policies = types.SimpleNamespace(
+    checkpoint_dots=
+    lambda prim, *_, **__: prim in {jax._src.lax.lax.dot_general_p,
+                                    jax._src.lax.lax.conv_general_dilated_p},
+)
 
 
 def named_call(
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2639,7 +2639,7 @@ def eval_shape(fun: Callable, *args, **kwargs):
 
 
 def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
-               saveable_policy: Optional[Callable[..., bool]] = None,
+               policy: Optional[Callable[..., bool]] = None,
                ) -> Callable:
   """Make ``fun`` recompute internal linearization points when differentiated.
 
@@ -2686,11 +2686,11 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
       ``pmap``, CSE can defeat the purpose of this decorator. But in some
       settings, like when used inside a ``scan``, this CSE prevention mechanism
       is unnecessary, in which case ``prevent_cse`` can be set to False.
-    saveable_policy: Optional, callable which takes as input a type-level
-      specification of a first-order primitive application and returns a boolean
-      indicating whether the corresponding output value(s) can be saved as a
-      residual (or, if not, instead must be recomputed in the (co)tangent
-      computation). Experimental.
+    policy: Optional, callable which takes as input a type-level specification
+      of a first-order primitive application and returns a boolean indicating
+      whether the corresponding output value(s) can be saved as a residual (or,
+      if not, instead must be recomputed in the (co)tangent computation).
+      Experimental.
 
   Returns:
     A function (callable) with the same input/output behavior as ``fun`` but
@@ -2743,7 +2743,7 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
     out_flat = pe.remat_call(flat_fun, *args_flat, name=flat_fun.__name__,
                              concrete=concrete, prevent_cse=prevent_cse,
                              differentiated=False,
-                             saveable_policy=saveable_policy)
+                             policy=policy)
     return tree_unflatten(out_tree(), out_flat)
   return fun_remat
 remat = checkpoint
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2639,6 +2639,7 @@ def eval_shape(fun: Callable, *args, **kwargs):
 
 
 def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
+               saveable_policy: Optional[Callable[..., bool]] = None,
                ) -> Callable:
   """Make ``fun`` recompute internal linearization points when differentiated.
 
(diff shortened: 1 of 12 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -772,67 +773,265 @@ def _remat_partial_eval(trace, _, f, tracers, params):
...
+  jaxpr1 = Jaxpr((), ins1, [*outs1, *residuals], eqns1)
+  config.jax_enable_checks and core.check_jaxpr(jaxpr1)
+
+  _, outs2 = partition_list(out_inst, jaxpr.outvars)
+  jaxpr2 = Jaxpr((), [*residuals, *jaxpr.invars], outs2, eqns2)
+  config.jax_enable_checks and core.check_jaxpr(jaxpr2)
+
+  return jaxpr1, jaxpr2, out_unknowns, out_inst, len(residuals)
+
+PartialEvalCustomResult = Tuple[Optional[JaxprEqn], Optional[JaxprEqn],
```
jekbradbury, line 938: Document what each of these tuple elements means? (or use a dataclass)

jax/_src/custom_derivatives.py
```diff
@@ -377,6 +378,23 @@ def _custom_jvp_call_jaxpr_transpose(reduce_axes, cts, *args, fun_jaxpr,
       fun_jaxpr.jaxpr, reduce_axes, fun_jaxpr.consts, args, cts)
 ad.reducing_transposes[custom_jvp_call_jaxpr_p] = _custom_jvp_call_jaxpr_transpose
 
+def custom_jvp_jaxpr_custom_partial_eval_rule(
+    saveable: Callable[..., bool], unks_in: List[bool], inst_in: List[bool],
+    eqn: core.JaxprEqn
+  ) -> Tuple[core.JaxprEqn, core.JaxprEqn, List[bool], List[bool], List[core.Var]]:
+  # It doesn't make sense to unzip (i.e. break up) a custom_jvp function into
+  # constituient parts, so we always perform full remat. An alternative would be
```
jekbradbury, line 386: a few typos in this paragraph

jax/_src/api.py
```diff
@@ -2611,6 +2611,7 @@ def __init__(self, shape, dtype):
 
 
 def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
+               saveable_policy: Optional[Callable[..., bool]] = None,
```
jekbradbury, line 2614: I don't really like saveable_policy; would be okay with just policy (but no strong preference)

jax/interpreters/partial_eval.py
```diff
@@ -772,67 +773,265 @@ def _remat_partial_eval(trace, _, f, tracers, params):
...
+
+def partial_eval_jaxpr_custom_rule_not_implemented(
+    saveable: Callable[..., bool], unks_in: List[bool], inst_in: List[bool],
+    eqn: JaxprEqn) -> PartialEvalCustomResult:
+  raise NotImplementedError
+
+
+ParamsUpdater = Callable[[List[bool], int, dict, dict], Tuple[dict, dict]]
+
+def partial_eval_jaxpr_custom_call_rule(
```
jekbradbury, line 953: Aren't rules more commonly called primitivename_ruletype_rule rather than ruletype_primitivename_rule?

jekbradbury: this one in particular sounds like it's a "custom call rule"

tests/api_test.py
```diff
@@ -2966,8 +2966,14 @@ def test_vmap_caching(self):
 
 class RematTest(jtu.JaxTestCase):
 
-  def test_remat_basic(self):
-    @api.remat
+  @parameterized.named_parameters(
+      {"testcase_name": f"{suffix}", "remat": remat}
+      for suffix, remat in [
```
jekbradbury, line 2971: Is it possible to factor this out?

jekbradbury (commented):
Haven't deeply reviewed yet, but definitely looks good and I trust the tests.

tests/api_test.py
mattjj: As in, stop repeating this same decorator text verbatim? Yeah, good idea, let me try...

jax/interpreters/partial_eval.py
mattjj: Good idea. Adding this comment:

```
# A primitive rule for policy-driven partial evaluation returns a 5-tuple
# with the components representing, respectively:
#  * the JaxprEqn for the 'known' side (or None if there is no known component),
#  * the JaxprEqn for the 'unknown' side (or None),
#  * a list of booleans indicating which of the original outputs are unknown,
#  * a list of booleans indicating which of the original outputs are
#    instantiated (i.e. available) in the 'unknown' side,
#  * a list of Var instances representing residuals to be added (i.e. to be
#    plumbed as outputs of the 'known' side jaxpr and added as input binders to
#    the 'unknown' jaxpr).
```

WDYT?

jax/_src/custom_derivatives.py
mattjj: Thanks, done!

jax/interpreters/partial_eval.py
mattjj: You're right. I switched to `call_partial_eval_jaxpr_custom_rule`. Still kinda bad but an improvement!

tests/api_test.py
mattjj: Done!

jax/_src/api.py
mattjj: Done.

tests/api_test.py
mattjj: Aw man, looks like there's a bad interaction with pytest-xdist or absl:

image

Not sure what's up with that, but I don't really feel like debugging it...

jekbradbury (approved):

jax/_src/api.py
```diff
@@ -2611,6 +2611,7 @@ def __init__(self, shape, dtype):
 
 
 def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
+               policy: Optional[Callable[..., bool]] = None,
```
apaszke, line 2614: Wasn't the plan to expose a few pre-packaged policies for now? Exposing primitives and their names might make future internal changes very breaking

jax/_src/util.py
```diff
@@ -74,6 +74,13 @@ def split_list(args, ns):
   lists.append(args)
   return lists
 
+def partition_list(bs: List[bool], l: List[Any]) -> Tuple[List[Any], List[Any]]:
+  assert len(bs) == len(l)
+  lists = lst1, lst2 = [], []  # type: ignore
+  for b, x in zip(bs, l):
+    lists[b].append(x)
+  return lst1, lst2
```
apaszke, line 82: code golf nit: this just `lists`, so you don't need the `lst1` and `lst2` binders :)

jax/interpreters/partial_eval.py
```diff
@@ -772,67 +773,275 @@ def _remat_partial_eval(trace, _, f, tracers, params):
...
-
-  # set up eqn for unknown outputs
-  in_tracers = (*const_tracers, *env_tracers, *instantiated_tracers)
-  eqn = new_eqn_recipe(in_tracers, unknown_output_tracers, remat_call_p,
-                       new_params, source_info_util.current())
-  for t in unknown_output_tracers: t.recipe = eqn
-  return _zip_knowns(known_output_tracers, unknown_output_tracers, out_unknowns)
+  if params['policy']:
+    # unzip into jaxpr1 and jaxpr2
+    jaxpr1_, jaxpr2_, out_unknowns, out_inst, _ = _partial_eval_jaxpr_custom(
```
apaszke, line 784: can we please call them jaxpr_known and jaxpr_unknown? The 1, 2 naming scheme is a great joke, but doesn't make it easy on the reader 😕  I can't even follow what happens next without this info

jax/interpreters/partial_eval.py
```diff
@@ -772,67 +773,275 @@ def _remat_partial_eval(trace, _, f, tracers, params):
...
+  assert len(known_list) + len(unknown_list) == len(which_unknown)
   known_iter, unknown_iter = iter(known_list), iter(unknown_list)
   return [next(unknown_iter) if uk else next(known_iter) for uk in which_unknown]
 
 
+def _partial_eval_jaxpr_custom(
+    jaxpr: Jaxpr, in_unknowns: List[bool], saveable: Callable[..., bool],
+  ) -> Tuple[Jaxpr, Jaxpr, List[bool], List[bool], int]:
+  if jaxpr.constvars: raise NotImplementedError  # TODO
+  env: Dict[Var, Tuple[bool, bool]] = {}
```
apaszke, line 881: perhaps use namedtuples instead of pairs of booleans? what does "inst" mean in the context of the function? we add the equations if inst is true of if it's false?

jax/interpreters/partial_eval.py
```diff
@@ -772,67 +773,275 @@ def _remat_partial_eval(trace, _, f, tracers, params):
...
+        eqns2.append(new_jaxpr_eqn(inputs, eqn.outvars, eqn.primitive,
+                                   eqn.params, eqn.source_info))
+        map(partial(write, False, True), eqn.outvars)
+  out_unknowns, out_inst = unzip2(map(read, jaxpr.outvars))
+  assert all(type(v) is Var for v in residuals), residuals
+
+  ins1, _ = partition_list(in_unknowns, jaxpr.invars)
+  outs1_, _ = partition_list(out_unknowns, jaxpr.outvars)
+  outs1 = [x for x in outs1_ if x.aval is not abstract_unit]
+  jaxpr1 = Jaxpr((), ins1, [*outs1, *residuals], eqns1)
```
apaszke, line 929: Should we sort the residuals (perhaps by names)? Otherwise we might be generating different jaxprs in different processes and that sounds like a recipe for trouble if one's running a distributed computation

apaszke (commented):
I did a quick skim and it looks fine, but pleeeeease stop using jaxpr1 and jaxpr2 because this is impossible to follow.

jax/interpreters/partial_eval.py
apaszke: known and staged sound good to me!

PR #7633, opened 2021-08-16
fix custom_call_status for rocm

reza-amd:
This PR fixes the issue introduced in PR #7480 for ROCm.

@deven-amd


yashk2810 (approved):

yashk2810:
There was a similar fix for GPU kernels, so this is good to go.

PR #7649, opened 2021-08-17
fix rocm_amdgpu_targets for rocm

reza-amd:
This PR updates the list of supported ```rocm_amdgpu_targets```.  In addition, it fixes a bug that the ```cuda_compute_capabilities and rocm_amdgpu_targets``` arguments in ```build.py ``` were not being reflected to .bazelrc

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -46,7 +46,7 @@ build:native_arch_posix --host_copt=-march=native
 build:mkl_open_source_only --define=tensorflow_mkldnn_contraction_kernel=1
 
 build:cuda --repo_env TF_NEED_CUDA=1
-build:cuda --action_env=TF_CUDA_COMPUTE_CAPABILITIES="3.5,5.2,6.0,6.1,7.0"
+build:cuda --action_env TF_CUDA_COMPUTE_CAPABILITIES="3.5,5.2,6.0,6.1,7.0"
 build:cuda --crosstool_top=@local_config_cuda//crosstool:toolchain
 build:cuda --@local_config_cuda//:enable_cuda
 build:cuda --define=xla_python_enable_gpu=true
@@ -55,7 +55,7 @@ build:rocm --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm --define=using_rocm=true --define=using_rocm_hipcc=true
 build:rocm --define=xla_python_enable_gpu=true
 build:rocm --repo_env TF_NEED_ROCM=1
-build:rocm --action_env=TF_ROCM_AMDGPU_TARGETS="gfx803,gfx900,gfx906,gfx1010"
+build:rocm --action_env TF_ROCM_AMDGPU_TARGETS="gfx900,gfx906,gfx908"
 
 build:nonccl --define=no_nccl_support=true
 
--- a/build/build.py
+++ b/build/build.py
@@ -375,17 +375,9 @@ def main():
       default=None,
       help="CUDNN version, e.g., 8")
   parser.add_argument(
-      "--cuda_compute_capabilities",
-      default="3.5,5.2,6.0,6.1,7.0",
-      help="A comma-separated list of CUDA compute capabilities to support.")
-  parser.add_argument(
       "--rocm_path",
       default=None,
       help="Path to the ROCm toolkit.")
-  parser.add_argument(
-      "--rocm_amdgpu_targets",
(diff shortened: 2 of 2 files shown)


.bazelrc
```diff
@@ -55,7 +55,7 @@ build:rocm --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm --define=using_rocm=true --define=using_rocm_hipcc=true
 build:rocm --define=xla_python_enable_gpu=true
 build:rocm --repo_env TF_NEED_ROCM=1
-build:rocm --action_env=TF_ROCM_AMDGPU_TARGETS="gfx803,gfx900,gfx906,gfx1010"
+build:rocm --action_env=TF_ROCM_AMDGPU_TARGETS="gfx900,gfx906,gfx908"
```
reza-amd, line 58: @yashk2810 , I think it should be ```--action_env TF_ROCM_AMDGPU_TARGETS``` (drop =). right?

yashk2810: Yeah, you are right. Can you change `TF_CUDA_COMPUTE_CAPABILITIES` too?

build/build.py
```diff
@@ -244,6 +245,9 @@ def write_bazelrc(python_bin_path=None, remote_build=None,
         tf_cuda_paths.append(cudnn_install_path)
       f.write("build --action_env CUDNN_INSTALL_PATH=\"{cudnn_install_path}\"\n"
               .format(cudnn_install_path=cudnn_install_path))
+    if cuda_compute_capabilities:
```
yashk2810, line 248: Why is this needed?

Just use `--config=cuda` when you are invoking `bazel`.

build/build.py
```diff
@@ -256,6 +260,9 @@ def write_bazelrc(python_bin_path=None, remote_build=None,
     if rocm_toolkit_path:
       f.write("build --action_env ROCM_PATH=\"{rocm_toolkit_path}\"\n"
               .format(rocm_toolkit_path=rocm_toolkit_path))
+    if rocm_amdgpu_targets:
+      f.write("build --action_env TF_ROCM_AMDGPU_TARGETS=\"{rocm_targets}\"\n"
```
yashk2810, line 264: Same comment. Just use `--config=rocm`.

Is this because you may need different targets that are not in the static .bazelrc flie?

.bazelrc
reza-amd: Yes.

build/build.py
reza-amd: The main reason I added this was because the ```parser.add_argument("--rocm_amdgpu_targets", .... ``` were still there, I am ok to drop this from the build.py and keep it in the static .bazelrc, but we should also update the arguments list of scripts as well (same for cuda_compute_capabilities).
Please let me know your decision to update the PR.

yashk2810: Yeah, let's do it. I would prefer to keep that in the static .bazelrc.

.bazelrc
```diff
@@ -55,7 +55,7 @@ build:rocm --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm --define=using_rocm=true --define=using_rocm_hipcc=true
 build:rocm --define=xla_python_enable_gpu=true
 build:rocm --repo_env TF_NEED_ROCM=1
-build:rocm --action_env=TF_ROCM_AMDGPU_TARGETS="gfx803,gfx900,gfx906,gfx1010"
```
yashk2810, line 58: 1010 is not a valid target?

reza-amd: No, it is not for now.

yashk2810 (approved):

reza-amd:
Thanks @yashk2810 for reviewing this PR. Just a quick note, we have forked JAX repo in https://github.com/ROCmSoftwarePlatform/jax. When I first submitted this PR in our forked repo (just for sharing it with my colleague before submitting it here), your CI jobs also started even though I did not submitted the PR here. I am not sure if this is intentional or not, I just wanted to share it with you in case you are not aware of.

Here is the link to the dummy PR:
https://github.com/ROCmSoftwarePlatform/jax/pull/5

yashk2810:
I think those tests are getting triggered since you have the workflows in your fork too: https://github.com/ROCmSoftwarePlatform/jax/blob/main-rocm/.github/workflows/ci-build.yaml

PR #7650, opened 2021-08-17
Adjust wording for errors involving array rank

jakevdp:
Why? A user pointed out that the term *rank* is probably not as well known/clear as *number of dimensions*

cc/ @patrickvonplaten

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2235,7 +2235,7 @@ def _broadcasting_shape_rule(name, *avals):
   if not shapes:
     return ()
   if len({len(shape) for shape in shapes}) != 1:
-    msg = '{} got arrays of different rank: {}.'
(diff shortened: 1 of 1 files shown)


jax/_src/lax/lax.py
```diff
@@ -2235,7 +2235,7 @@ def _broadcasting_shape_rule(name, *avals):
   if not shapes:
     return ()
   if len({len(shape) for shape in shapes}) != 1:
-    msg = '{} got arrays of different rank: {}.'
+    msg = '{}: arrays must have same number of dimenions, got {}.'
```
skye, line 2238:
```suggestion
    msg = '{}: arrays must have same number of dimensions, got {}.'
```

skye (approved):
Optional: add a unit test that shows the full error message (it's hard to tell from the code how the "got {}"s will be formatted).

PR #7651, opened 2021-08-17
custom_vjp: automatically handle float0 cotangents

mattjj:
The main issue here was that when a custom_vjp function took a primal input with integer dtype, we would demand (via an internal assertion) that the user's bwd function would produce a value with float0 dtype (i.e. the tangent type corresponding to an integer primal). But users don't know about float0s, and shouldn't have to learn about them! So instead we just automatically handle producing zeros here (with the right dtype).

Also misc cleanups (I'll mark them below in review comments).

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -242,6 +242,7 @@ def _flatten_jvp(in_tree, *args):
     msg = ("Custom JVP rule must produce primal and tangent outputs with equal "
            "container (pytree) structures, but got {} and {} respectively.")
     raise TypeError(msg.format(out_tree, out_tree2))
+  # TODO(mattjj): compare primals' tangent types to tangent objects' types
   primal_avals_out = [
       raise_to_shaped(core.get_aval(x), weak_type=False).strip_named_shape()
       for x in primals_out]
@@ -553,8 +554,10 @@ def _flatten_bwd(in_tree, in_avals, out_trees, *args):
            "number of arguments to the primal function, but got VJP output "
            "structure {} for primal input structure {}.")
     raise TypeError(msg.format(in_tree2, in_tree)) from None
-  yield [zeros_like_aval(aval.at_least_vspace()) if ct is zero else ct
-         for aval, ct in zip(in_avals, cts_in_flat)]
+  # Ignore any None cotangents, and any corresponding to inputs for which the
+  # type doesn't equal the tangent type (i.e. float0s)
+  yield [Zero(a.at_least_vspace()) if ct is zero or a != a.at_least_vspace()
+         else ct for a, ct in zip(in_avals, cts_in_flat)]
 
 
 class CustomVJPCallPrimitive(core.CallPrimitive):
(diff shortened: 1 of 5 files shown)


jax/_src/custom_derivatives.py
```diff
@@ -242,6 +242,7 @@ def _flatten_jvp(in_tree, *args):
     msg = ("Custom JVP rule must produce primal and tangent outputs with equal "
            "container (pytree) structures, but got {} and {} respectively.")
     raise TypeError(msg.format(out_tree, out_tree2))
+  # TODO(mattjj): compare primals' tangent types to tangent objects' types
```
mattjj, line 245: This is an unrelated bug I happened to notice. It's baked in the assumption that primal type equals tangent type, i.e. that `T a == a`, which is probably true for all inputs to custom_jvp/vjp functions given existing JAX datatypes but won't always be true.

jax/_src/custom_derivatives.py
```diff
@@ -553,8 +554,10 @@ def _flatten_bwd(in_tree, in_avals, out_trees, *args):
            "number of arguments to the primal function, but got VJP output "
            "structure {} for primal input structure {}.")
     raise TypeError(msg.format(in_tree2, in_tree)) from None
-  yield [zeros_like_aval(aval.at_least_vspace()) if ct is zero else ct
-         for aval, ct in zip(in_avals, cts_in_flat)]
+  # Ignore any None cotangents, and any corresponding to inputs for which the
+  # type doesn't equal the tangent type (i.e. float0s)
+  yield [Zero(a.at_least_vspace()) if ct is zero or a != a.at_least_vspace()
```
mattjj, line 559: The `or a != a.at_least_vspace()` is the only thing we needed to fix the bug at hand.

Another change on this line was to produce symbolic zeros using `Zero` rather than `zeros_like_aval`. I couldn't resist making this optimization, even though it is unrelated. (I hope it doesn't cause unexpected breakages...)

jax/core.py
```diff
@@ -2076,9 +2076,3 @@ def pp_kv_pairs(kv_pairs):
     return pp('[ ') >> vcat([pp_kv_pair(k, v) for k, v in kv_pairs]) >> pp(' ]')
   else:
     return pp('')
-
-# Casting float0 array to a float-valued zero array.
-def zeros_like_float0(array, dtype=None):
```
mattjj, line 2081: I moved this function to the only file where it's used, and then I realized that we could just inline it.

jax/interpreters/batching.py
```diff
@@ -275,9 +276,24 @@ def batch_custom_vjp_bwd(bwd, axis_name, axis_size, in_dims, out_dim_dests, main
 def _match_axes_and_sum(axis_size, out_dims_thunk, out_dim_dests, *in_vals):
   # this is like _match_axes, but we do reduce-sums as needed
   out_vals = yield in_vals, {}
-  yield map(partial(matchaxis, axis_size, sum_match=True),
+  yield map(partial(_matchaxis_symbolic_zeros, axis_size, sum_match=True),
             out_dims_thunk(), out_dim_dests, out_vals)
 
+def _matchaxis_symbolic_zeros(sz, src, dst, x, sum_match=False):
```
mattjj, line 282: This change is unrelated to the bug at hand, but it is related to producing symbolic Zeros as the output of a custom_vjp bwd function (mentioned above). The reason we weren't producing symbolic zeros before is that we had to make this function, which is only called in the vmap rule for custom_vjp, handle them.

mattjj:
On second thought, I'm going to be more responsible and break this PR up into two.

Issue #7535, opened 2021-08-06
Expose jax.core.eval_context() as part of the public API


FedericoV:
Currently, this works:

```from functools import partial
import jax
import jax.numpy as jnp
from jax._src.custom_derivatives import closure_convert

idx = jnp.array([True, False])

def test_fcn(x):
  return x[idx]

@jax.jit
def f(x):
  g, args = closure_convert(test_fcn, x)
  return g(x, *args)

f(jnp.array([0., 1.]))
```

But this does not work unless you use jax.core.eval_context():

```
from functools import partial
import jax
import jax.numpy as jnp
from jax._src.custom_derivatives import closure_convert

idx = jnp.array([True, False])

def test_fcn(x):
  with jax.core.eval_context():
     bad_idx = idx == 0
  return x[bad_idx]

@jax.jit
def f(x):
  g, args = closure_convert(test_fcn, x)
  return g(x, *args)

f(jnp.array([0., 1.]))
```

The API should probably be renamed to something like - fold_at_compile_time or something similar, but, it's useful functionality to end users.

mattjj:
Naming suggestions from @froystig:

> yeah, along those lines: const_{eval,fold}, const_{eval,fold}_only, {ensure,force}_compile_time, {ensure,force}_compile_time_eval

froystig:
closed in #7987

PR #7659, opened 2021-08-18
logsumexp: fix issue with debug_nans

jakevdp:
Similar issue to #7648


jakevdp:
Lint error is #7660

froystig (approved):
Nice. This is tricky. Maybe reference an issue number or extend the inline comment?

@mattjj, could #7535 have helped here?

mattjj:
> @mattjj, could #7535 have helped here?

I'm not sure but it doesn't seem likely to me.

Issue #3038, opened 2020-05-11
`jax.numpy.sum` and `jax.numpy.mean` inconsistency on python lists


yujiali:
This works:
```
import jax.numpy as jnp

x = [jnp.ones((), dtype=jnp.float32), jnp.ones((), dtype=jnp.float32)]
jnp.sum(x)
```
and gives me `DeviceArray(2., dtype=float32)`.  But this doesn't:

```
import jax.numpy as jnp

x = [jnp.ones((), dtype=jnp.float32), jnp.ones((), dtype=jnp.float32)]
jnp.mean(x)
```
and I got `TypeError: data type not understood`.

In numpy both would work:
```
import numpy as np

x = [np.ones((), dtype=np.float32), np.ones((), dtype=np.float32)]
np.mean(x)  # 1.0
np.sum(x)  # 2.0
```

Is this intended?

jakevdp:
Generally jax does not automatically upgrade lists to device arrays within its numpy API functions, though this behavior is by no means uniform across the package. The difference in behavior here is due to an implementation diffrence between ``mean`` and ``sum``; you can get the expected behavior by always passing a jax array explicitly:
```
In [1]: import jax.numpy as jnp                                                 

In [2]: x = jnp.array([jnp.ones(()), jnp.ones(())])                             

In [3]: jnp.sum(x)                                                              
Out[3]: DeviceArray(2., dtype=float32)

In [4]: jnp.mean(x)                                                             
Out[4]: DeviceArray(1., dtype=float32)
```

yujiali:
Thanks Jake.  I know how to work around this issue, was just asking about the difference in behavior between `jnp.sum` and `jnp.mean`.  The implementations are different, but if the behaviors are consistent across all the reduce operators then the users will have one less thing to think about.

jakevdp:
I'm going to mark this as a bug, because the package should have a consistent API on this one way or the other. Thanks for the report.

jakevdp:
From discussions with team members, I think the preferred appoach here would be to *not* accept lists as inputs to JAX functions (in particular, converting lists to arrays may lead to strange interactions with how we handle pytrees internally).

So to make the API more consistent, we should ensure that ``sum()`` and other routines raise an appropriate ``TypeError`` when passed a Python list.

hawkinsp:
Both `sum` and `mean` now raise errors at head!

PR #7662, opened 2021-08-18
Require array-like inputs for jax.numpy set-like operations

jakevdp:
This is a breaking change for users passing lists to these functions, but disallowing list inputs has been a goal for a while (cf. #3038). Part of #7737.


froystig (approved):
good catch!

PR #7672, opened 2021-08-19
chore(ci) : setup dependabot for pip and github-actions

mmorel-35:
Hi! This allows you to receive pr to keep your dependencies up-to-date.
See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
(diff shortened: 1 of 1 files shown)


jakevdp:
I think this is probably not a good fit for JAX, because all contributions require a CLA to be signed by the commit author: because of that, I don't think dependabot-authored commits will be able to be merged.

mmorel-35:
Take a look at go-github, here https://github.com/google/go-github/pull/2053, is there a different configuration on Jax?

jakevdp:
I think that's the same CLA configuration, so it might work after all.

One question, since I couldn't find any info in the docs: for pip mode, what files does dependabot look at to find outdated dependencies? Does it parse `setup.py`?

mmorel-35:
When I go in Insight > Dependency  graph, it is able to work with "setup.py, docs/requirements.txt, and build/test-requirements.txt"  (concerning JAX)

jakevdp:
Sounds good - let's give it a try!

PR #7692, opened 2021-08-21
add support for scipy.special.{expn,expi,exp1}

terhorst:
This PR adds support for some exponential integral functions. The code is based off of Cephes, so there shouldn't be any licensing issues. ~I believe the underlying code is all correct, as it validates against the scipy version as well as check_grad. However, I am not able to get it to pass the jax test suite--I get a cryptic error about recompilation that I have not been able to solve. I think it may be related to use of `jnp.piecewise`, but I cannot see why. I'm submitting this PR in hopes that someone else may understand what's going on.~ One issue is that I have to `jit` the added functions in order to avoid test failures. See #7693.


terhorst:
jitting solves the issue described above for exp1 and expn. For expi, now I get a leaked tracer.

jax/_src/scipy/special.py
```diff
@@ -1089,3 +1089,379 @@ def sph_harm(m: jnp.ndarray,
...
 
   return _sph_harm(m, n, theta, phi, n_max)
+
+
+# exponential integrals
+
+
+def _expint1(x):
+    # 0 < x <= 2
+    A = [
```
jakevdp, line 1099: Please add a comment that indicates the source of the algorithm used here

jax/_src/scipy/special.py
```diff
@@ -1089,3 +1089,379 @@ def sph_harm(m: jnp.ndarray,
       'be statically specified to use `sph_harm` within JAX transformations.')
 
   return _sph_harm(m, n, theta, phi, n_max)
+
+
+# exponential integrals
+
+
+def _expint1(x):
+    # 0 < x <= 2
```
jakevdp, line 1098: Please use two spaces for indentation

jax/_src/scipy/special.py
```diff
@@ -1089,3 +1089,379 @@ def sph_harm(m: jnp.ndarray,
...
+
+
+def _eval_expint_k(A, B, x):
+    # helper function for all subsequent intervals
+    A, B = [jnp.array(U, dtype=x.dtype) for U in [A, B]]
+    one = _constant_like(x, 1.0)
+    w = one / x
+    f = jnp.polyval(A, w) / jnp.polyval(B, w)
+    f = w * f + one
+    # w, f, x = id_print((w, f, x), what="eval_expint")
```
jakevdp, line 1128: Remove?

jax/_src/scipy/special.py
```diff
@@ -1089,3 +1089,379 @@ def sph_harm(m: jnp.ndarray,
...
+
+
+def _expn1(n, x):
+    # exponential integral En
+    _c = _constant_like
+    x = jnp.array(x)
+    MACHEP = jnp.finfo(x.dtype).eps
+
+    def gamma(x):
+        from jax.scipy.special import gammaln
```
jakevdp, line 1324: I think this import can be removed, because gammaln is defined above in this file

jax/_src/scipy/special.py
```diff
@@ -1089,3 +1089,379 @@ def sph_harm(m: jnp.ndarray,
...
+        d["t"] = jnp.where(d["ans"] != zero, abs(d["yk"] / d["ans"]), one)
+        return d
+
+    def cond(d):
+        return (d["x"] > _c(d["x"], 0.0)) & (d["t"] > MACHEP)
+
+    d = lax.while_loop(cond, body, init)
+    t = n
+    r = n - _c(n, 1)
+    return d["z"] ** r * psi / gamma(t) - d["ans"]
```
jakevdp, line 1357: use `jnp.exp(gammaln(t))` inline here?

jakevdp (changes requested):
Looks great! Thanks for digging-in on this. A couple minor changes below; also I think we can get #7693 fixed today to allow full testing here.

terhorst:
All done. I removed the jits in anticipation of resolving #7693.

jakevdp:
I'm not sure how soon #7693 will be resolved (in my judgment, the fix seems more complicated than is warranted for the issue) so if you have a way to make these tests pass without it, we should probably go with that.

terhorst:
Okay, I added them back.

jakevdp (approved):

jakevdp:
Hi - just FYI, the delay in merging this is due to the fact that the exp1 tests fail on GPU with an error that I'm having difficulty debugging.

jakevdp:
My working theory is that might somehow be due to the inefficiency of the implementation (the GPU abort looks like it may come from resources being exhausted somehow) – I wonder if rather than using `piecewise` over multiple functions containing different constants, it might be more efficient to put all the constants in a 2D array and use `jnp.where` to select the applicable row?

jakevdp:
To be more concrete, this is the failure we're seeing:
```
[ RUN      ] LaxBackedScipyTests.testScipySpecialFunExp1_float64[]
W0824 09:08:23.276341    5286 symbolize_elf.inc:1312] RAW: /build/cas/3fe/3fec575d73fe9a7bdb835176c490dd0b8690e3f8aaeef028ac72e1df44121122_02005d5b2df8: too many executable LOAD segments
F0824 09:08:23.275103    5286 nvptx_compiler.cc:480] ptxas returned an error during compilation of ptx to sass: 'INTERNAL: ptxas exited with non-zero error code 11, output: '  If the error message indicates that a file could not be written, please verify that sufficient filesystem space is provided.
```

terhorst:
Thanks for the update. The tests pass on my GPU and `jaxlib[cuda11]`--what is the error you're getting. Re: `jnp.where`, that should be possible by zero-padding all the A and B arrays (from the front) to have equal length.

terhorst:
Oh thanks, just saw your message. Yeah I would have no idea how to debug that :-) Let me try your suggestion, I'll push in a sec.

terhorst:
Actually if it's failing on `exp1` then it's unrelated to the use of `piecewise` to select the correct rational function approximation. Those only get invoked for `expi`. `expi` does end up calling `expn` to handle negative arguments, but not the other way around (confusing I know). 

Can you check if the same failure occurs for `exp1` specifically? `exp1`/`expn` also employ `piecewise`, but in a way that's harder to eliminate.

jakevdp:
It looks like all three functions cause a similar error, but only for float64. So I think for this PR, maybe the best option is in the tests to change `float_dtypes` to `[np.float32]` and leave a TODO to run tests for float64 as well.

jakevdp:
Given that float64 is skipped by default, you may be able to reproduce this using
```
$ JAX_ENABLE_X64=1 pytest -n auto tests/lax_scipy_test.py -k SpecialFunExp
```

terhorst:
Even enabling x64, everything passes locally for me, so it's going to be hard to debug further. Restricting the tests to float32 seems like a fine option.

jakevdp:
Would you be willing to make those test changes, and then also squash all the changes into a single commit? Thanks!

jakevdp:
Looks like it's in! Thanks for the work on this!

PR #7699, opened 2021-08-23
jnp.where: provide fill_value for single-argument form

noahtren:
Extends #6910 & #7592. Padding with 0 can collide with actual zero indices and cause problems. We could specify the pad value, i.e. as -1.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1777,16 +1777,16 @@ three-argument form does not have a data-dependent shape and can be JIT-compiled
 successfully. Alternatively, you can specify the optional ``size`` keyword:
 if specified, the first ``size`` True elements will be returned; if there
 are fewer True elements than ``size`` indicates, the index arrays will be
-padded with zeros.
+padded with ``fill_value`` (default is 0.)
 """
 
 @_wraps(np.where, update_doc=False, lax_description=_WHERE_DOC)
-def where(condition, x=None, y=None, *, size=None):
+def where(condition, x=None, y=None, *, size=None, fill_value=None):
   if x is None and y is None:
-    return nonzero(asarray(condition), size=size)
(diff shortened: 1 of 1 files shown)


jakevdp (approved):
Looks great, thanks!

jax/_src/numpy/lax_numpy.py
```diff
@@ -1777,13 +1777,13 @@ def _where(condition, x=None, y=None):
...
 """
 
 @_wraps(np.where, update_doc=False, lax_description=_WHERE_DOC)
-def where(condition, x=None, y=None, *, size=None):
+def where(condition, x=None, y=None, *, size=None, fill_value=None):
   if x is None and y is None:
-    return nonzero(asarray(condition), size=size)
+    return nonzero(asarray(condition), size=size, fill_value=fill_value)
   else:
     if size is not None:
```
jakevdp, line 1788: Perhaps check if `fill_value is not None` here as well?

noahtren: How's this? The semantics are the same but it only passes down `fill_value` if provided.

jakevdp: Sorry to be unclear: I meant below, where it raises an error `if size is not None`. We should change that to `if size is not None or fill_value is not None` and change the error text appropriately.

I think it's fine to pass `fill_value` to `nonzero` regardless of its value.

noahtren: Got it!

jakevdp (commented):
Looks great! The last thing we need is to squash all the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Then we can get this merged!

jakevdp (approved):

jakevdp:
Thanks for working on this!

PR #7715, opened 2021-08-25
jnp.histogramdd: fix support for range argument

### LINKED ISSUE #7714: histogram2d error when specifying `range`
```python
from jax import numpy as jnp
a = [-2,2]
b=jnp.array([a,a])
print(b)
jnp.histogram2d(jnp.ones(10), jnp.ones(10), bins=10, range=b)
```

Suggested fix from the user: [this line](https://github.com/google/jax/blob/7a40aa0114e46a7a8975e04b175dfbc7b0b388ef/jax/_src/numpy/lax_numpy.py#L1054) should probably have `range[i]`. (That seems to fix it, but I didn't try writing tests!)

@jakevdp could you take a look?

jakevdp:
Fixes #7714

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1037,6 +1037,11 @@ def histogramdd(sample, bins=10, range=None, weights=None, density=None):
   if weights is not None and weights.shape != (N,):
     raise ValueError("should have one weight for each sample.")
 
+  if range is not None and (
+      len(range) != D or _any(r is not None and len(r) != 2 for r in range)):
+    raise ValueError(f"For sample.shape={(N, D)}, range must be a sequence "
+                     f"of {D} pairs or Nones; got range={range}")
+
   try:
     num_bins = len(bins)
     if num_bins != D:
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1037,6 +1037,10 @@ def histogramdd(sample, bins=10, range=None, weights=None, density=None):
   if weights is not None and weights.shape != (N,):
     raise ValueError("should have one weight for each sample.")
 
+  if range is not None and len(range) != D:
```
mattjj, line 1040: In addition to checking that it's a sequence of length D, would it make sense to check that every element is itself a sequence of length 2 (of integers)? I wonder if `try: np.shape(range) == (N, D)` would work...

mattjj (approved):
Awesome! LGTM and feel free to merge as-is, though I have one question/suggestion inline, and one here: would it make sense to add an 'error test', i.e. a test that, when given some malformed `range` values, the function raises an exception? (I'm sure our test coverage of errors is uneven, but we could chip away at it!)

jax/_src/numpy/lax_numpy.py
jakevdp: Sure, I'll add that validation.

PR #7717, opened 2021-08-25
Make `input_shape` explicitly integer-typed in `_reduce_jvp`

romanngg:
This is to avoid accidental float shapes when primal shape is `()`. Example where this fails: `grad(jnp.prod)(jnp.ones(()))`. Add a test.


jakevdp (approved):
Looks good, thanks!

PR #7719, opened 2021-08-25
Add jax.debugging.make_hlo()

jakevdp:
Based on code by @levskaya

See the generated docs here: https://jax--7719.org.readthedocs.build/en/7719/_autosummary/jax.debugging.make_hlo.html#jax.debugging.make_hlo

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -1,0 +1,88 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


soraros:
If this PR only deals with the printing part(as it returns a string), I think having an extra `print_option` variable would be helpful. Something along the line of
```python
def make_hlo(..., print_option: str | HloPrintOptions = 'short_parsable'):
  ...
  if isinstance(print_option, str):
    match print_option:
      case 'short_parsable': ...
      case 'canonical': ...
	  case 'fingerprint': ...
      case _: raise
  else:
    ...
 ...
```

jax/_src/debugging.py
```diff
@@ -0,0 +1,88 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jax
+
+
+def make_hlo(f, optimize=False, metadata=False, platform=None):
```
zhangqiaorjc, line 18: I agree that adding a print option would be nice

but we could ask  @soraros to contribute an improvement PR if there's interest

zhangqiaorjc (approved):

hawkinsp:
Closing because I think @froystig 's work on the new AOT APIs replaces it.

PR #7729, opened 2021-08-26
Add np.polydiv numpy function

milind-soni:
added the numpy polydiv core function (part of #70)

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4071,19 +4071,16 @@ def polyint(p, m=1, k=None):
 def polydiv(u,v):
   _check_arraylike("polydiv", u, v)
   u, v = _promote_dtypes_inexact(u, v)
-
-  w = u[0] + v[0]
   m = len(u) - 1
   n = len(v) - 1
   scale = 1. / v[0]
-  q = zeros((_max(m - n + 1, 1),), w.dtype)
-  r = u.astype(w.dtype)
+  q = zeros(_max(m - n + 1, 1),)
   for k in range(0, m-n+1):
-    d = scale * r[k]   
+    d = scale * u[k]   
     q = q.at[k].set(d)
-    r = r.at[k:k+n+1].add(-d*v)
-  r = r.at[~(abs(r) > 1E-14).cumsum().astype(bool)].set(0)
-  return q, r
+    u = u.at[k:k+n+1].add(-d*v)
+  u = u.at[~(abs(u) > 1E-14).cumsum().astype(bool)].set(0)
+  return q, u
 
 
 @_wraps(np.polyder)
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4067,6 +4067,26 @@ def polyint(p, m=1, k=None):
     return true_divide(concatenate((p, k)), coeff)
 
 
+@_wraps(np.polydiv)
+def polydiv(u,v):
+  u = asarray(u) + 0.0
+  v = asarray(v) + 0.0
+
+  w = u[0] + v[0]
+  m = len(u) - 1
+  n = len(v) - 1
+  scale = 1. / v[0]
+  q = zeros((max(m - n + 1, 1),), w.dtype)
+  r = u.astype(w.dtype)
+  for k in range(0, m-n+1):
+    d = scale * r[k]
+    q[k] = d
+    r[k:k+n+1] -= d*v
+  while allclose(r[0], 0, rtol=1e-14) and (r.shape[-1] > 1):
+    r = r[1:]
+  return q, r
+
+
 @_wraps(np.polyder)
 def polyder(p, m=1):
   m = core.concrete_or_error(operator.index, m, "'m' argument of jnp.polyder")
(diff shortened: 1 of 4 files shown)


jakevdp (changes requested):
Thanks for tackling this! A few comments below.

jax/_src/numpy/lax_numpy.py
```diff
@@ -4067,6 +4067,26 @@ def polyint(p, m=1, k=None):
...
+  n = len(v) - 1
+  scale = 1. / v[0]
+  q = zeros((max(m - n + 1, 1),), w.dtype)
+  r = u.astype(w.dtype)
+  for k in range(0, m-n+1):
+    d = scale * r[k]
+    q[k] = d
+    r[k:k+n+1] -= d*v
+  while allclose(r[0], 0, rtol=1e-14) and (r.shape[-1] > 1):
+    r = r[1:]
```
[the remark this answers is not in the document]
milind-soni: I am facing issues with this line 
```
q = zeros((max(m - n + 1, 1),), w.dtype)
```
It is showing me a Value error 

also what other way would you suggest for assignment in JAX with reference to  r[k:k+n+1] -= d*v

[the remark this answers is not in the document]
jakevdp: `jax.numpy` overrides `max` to match the behavior of `numpy.max`. If you want the builtin `max` function, use `_max` here.

`r[k:k+n+1] -= d * v` will not work because JAX arrays are immutable. You'll need to use [index update])https://jax.readthedocs.io/en/latest/jax.ops.html#indexed-update-operators) operations instead; e.g.
```python
r = r.at[k:k+n+1].add(-d*v)
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -4067,6 +4067,25 @@ def polyint(p, m=1, k=None):
     return true_divide(concatenate((p, k)), coeff)
 
 
+@_wraps(np.polydiv)
+def polydiv(u,v):
+  _check_arraylike("polydiv", u, v)
+  u, v = _promote_dtypes_inexact(u, v)
+
+  w = u[0] + v[0]
```
soraros, line 4078: I think this line can be removed, also the type cast `r = u.astype(...)`.

jax/_src/numpy/lax_numpy.py
```diff
@@ -4067,6 +4067,25 @@ def polyint(p, m=1, k=None):
...
+  m = len(u) - 1
+  n = len(v) - 1
+  scale = 1. / v[0]
```
soraros, line 4081: Some major care still needs to be taken here, since in compiled code we generally don't have the guarantee that `v` is leading-zero-free.

jax/_src/numpy/lax_numpy.py
```diff
@@ -4067,6 +4067,25 @@ def polyint(p, m=1, k=None):
...
+  q = zeros((_max(m - n + 1, 1),), w.dtype)
+  r = u.astype(w.dtype)
+  for k in range(0, m-n+1):
+    d = scale * r[k]   
+    q = q.at[k].set(d)
+    r = r.at[k:k+n+1].add(-d*v)
+  r = r.at[~(abs(r) > 1E-14).cumsum().astype(bool)].set(0)
```
soraros, line 4088: Maybe replace `1E-14` with `something * jnp.finfo(r.dtype).eps`?

jax/_src/numpy/lax_numpy.py
milind-soni: oh okay just saw

jax/_src/numpy/lax_numpy.py
milind-soni: got it will update

milind-soni: also, what will you suggest to make this code not outdated

soraros: > what will you suggest to make this code not outdated

That I have no idea. 🤷

jax/_src/numpy/lax_numpy.py
```diff
@@ -4070,6 +4070,22 @@ def polyint(p, m=1, k=None):
...
+  u, v = _promote_dtypes_inexact(u, v)
+  m = len(u) - 1
+  n = len(v) - 1
+  scale = 1. / v[0]
+  q = zeros(_max(m - n + 1, 1),)
+  for k in range(0, m-n+1):
+    d = scale * u[k]   
+    q = q.at[k].set(d)
+    u = u.at[k:k+n+1].add(-d*v)
+  u = u.at[~(abs(u) > 1E-14).cumsum().astype(bool)].set(0)
```
jakevdp, line 4085: Since this like is pretty non-trivial, I might factor out a helper function for clarity. Something like
```
def _zero_out_leading_elements(u, tol=1E-14):
  return u.at[~(abs(u) > 1E-14).cumsum().astype(bool)].set(0)
```
Also, the tolerance may have to be adjusted depending on the dtype of `u`. In numpy, I believe it's always cast to float64, so hard-coding 1E-14 is probably OK. In JAX, where possible we try to keep computations in 16 or 32 bit, matching the inputs, because these types have better support on some accelerators.

jakevdp:
Superseded by #10226 - thanks!

PR #7731, opened 2021-08-26
Check for aval weak_type attr before accessing it.

zhangqiaorjc:


jakevdp (approved):
Looks good: this fixes the issue, but I'm a bit confused how this came up in the first place. How does a `DeviceArray` end up with an `aval` that does not have `weak_type` defined? Shouldn't all abstract values for `DeviceArray` derive from `UnshapedArray`, which defines a `weak_type` slot?

mattjj (approved):

zhangqiaorjc:
fixed in an internal cl

PR #7732, opened 2021-08-26
Require array-like inputs for jnp.poly functions

jakevdp:
These functions had previously been designed to recognize `np.poly1d` objects; in retrospect I think this was a mistake, because such objects are incompatible with JAX tracing/transformations.

This is a breaking change for users passing `list` or `poly1d` objects to these functions, but disallowing non-array inputs has been a goal for a while (cf. #3038) Part of #7737


froystig (approved):
Add a changelog entry as part of this PR?

PR #7734, opened 2021-08-26
Combine initial_style_batchers with collective_rules

### LINKED ISSUE #7047: custom_vjp (and jvp) doesn't support collectives
The following crashes (reported by @sharadmv):
```py
@jax.custom_vjp
def f(x):
  return lax.psum(x, 'foo')

def f_fwd(x):
  return lax.psum(x, 'foo'), None

def f_bwd(res, dx):
  return dx
f.defvjp(f_fwd, f_bwd)

maps.xmap(f, axis_sizes={'foo': 4}, in_axes={}, out_axes={})(2.)
```

Most likely this is because the `custom_vjp` gets no batching tracers as inputs, entirely skipping the batching interpreter, while it really needs to get triggered to desugar the psum over the axis name into a positional one.

sharadmv:
Initial style primitives (control flow primitives, custom derivative primitives) currently do not work with collectives under certain conditions. Suppose we have an initial style primitive with a call to `axis_index` in it. At some point, the `vmap` transformation needs to turn that call into the actual index. When it's inside of an initial style primitive though, the `BatchTrace` can often miss the staged jaxpr entirely. In order for it to trigger, several things need to happen:

1. The initial style primitive takes batched tracers as inputs. This is the usual route by which JAX dispatches primitives to traces in `bind`. 
2. The primitive is an `AxisPrimitive`, leading to a special `bind`. `AxisPrimitive.bind` checks if the primitive uses an axis names and dispatches to the appropriate `BatchTrace`.

Since initial style primitives are generally *not* `AxisPrimitive`s, we only really have the 1st option. Unfortunately that means if we have an initial style primitive that doesn't take in a batched value but calls `axis_index`, the `BatchTrace` will miss it entirely, leading to errors when we hit the `impl` rule for `axis_index`. This leads to the issue found in #7047 for custom derivative primitives (and there is an analogous case for control flow primitives).

## Solution

We make all initial style primitives `AxisPrimitive`s. This will use the special `bind` that will dispatch to `BatchTrace` if the initial style primitive uses axis names in its body. 

Once we do this, the `BatchTrace` is triggered but we run into yet another issue. The `BatchTrace` short circuits to `primitive.bind` if it sees none of the inputs have batch dimensions (which is the case we're talking about). Before it does this, however, it checks to see if the primitive has a `collective_rule` (generally reserved for `psum`, `pmax`, etc.). It turns out that `initial_style_batchers` (the usual way of batching an initial style primitive) and `collective_rules` share a lot of similarities and since they essentially serve the same purpose, we unify them. All previously `initial_style_batcher`s rules will be converted into `collective_rules` since the code path for triggering an initial style batcher via `AxisPrimitive.bind` matches the path for the other collectives.

Also as suggested by @apaszke, we add a `no_axis_name` sentinel instead of `None`.

Fixes: #7047

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -377,10 +377,9 @@ def _pred_bcast_select(c, pred, x, y, x_y_aval: core.AbstractValue):
     bcast_pred = xops.BroadcastInDim(pred, x_shape, list(range(len(pred_shape))))
     return xops.Select(bcast_pred, x, y)
 
-def _while_loop_batching_rule(args, dims, axis_name, main_type,
+def _while_loop_batching_rule(axis_size, axis_name, main_type, args, dims,
                               cond_nconsts, cond_jaxpr,
                               body_nconsts, body_jaxpr):
-  size, = {x.shape[d] for x, d in zip(args, dims) if d is not batching.not_mapped}
   orig_batched = [d is not batching.not_mapped for d in dims]
   cconst_bat, bconst_bat, init_bat = split_list(orig_batched, [cond_nconsts, body_nconsts])
   cconsts, bconsts, init = split_list(args, [cond_nconsts, body_nconsts])
@@ -393,7 +392,7 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
   # reach a fixpoint.
   for _ in range(1 + len(carry_bat)):
     _, carry_bat_out = batching.batch_jaxpr(
-        body_jaxpr, size, bconst_bat + carry_bat, instantiate=False,
+        body_jaxpr, axis_size, bconst_bat + carry_bat, instantiate=False,
         axis_name=axis_name, main_type=main_type)
     if carry_bat == carry_bat_out:
       break
@@ -404,7 +403,7 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
   # Knowing how the carry is batched now, we can determine if the predicate is
   # batched.
   _, (pred_bat,) = batching.batch_jaxpr(
-      cond_jaxpr, size, cconst_bat + carry_bat, instantiate=False,
+      cond_jaxpr, axis_size, cconst_bat + carry_bat, instantiate=False,
       axis_name=axis_name, main_type=main_type)
 
   if pred_bat:
@@ -413,10 +412,10 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
     carry_bat = [True] * len(carry_bat)
     carry_dims = [0] * len(carry_bat)
     body_jaxpr_batched, _ = batching.batch_jaxpr_axes(
-        body_jaxpr, size, bconst_dims + carry_dims,
+        body_jaxpr, axis_size, bconst_dims + carry_dims,
         carry_dims, axis_name=axis_name, main_type=main_type)
     cond_jaxpr_batched, _ = batching.batch_jaxpr_axes(
-        cond_jaxpr, size, cconst_dims + carry_dims, [0],
+        cond_jaxpr, axis_size, cconst_dims + carry_dims, [0],
         axis_name=axis_name, main_type=main_type)
   else:
     # If the predicate is not batched, we can look at the `cond_jaxpr`'s out
@@ -427,12 +426,12 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
     cond_rank = len(cond_jaxpr.out_avals[0].shape)
     carry_dims = [cond_rank if b else None for b in carry_bat]
     body_jaxpr_batched, _ = batching.batch_jaxpr_axes(
-        body_jaxpr, size, bconst_dims + carry_dims, carry_dims,
+        body_jaxpr, axis_size, bconst_dims + carry_dims, carry_dims,
         axis_name=axis_name, main_type=main_type)
     # Now we need to rebatch the `cond_jaxpr` according to the new dims of the
     # carry.
     cond_jaxpr_batched, _ = batching.batch_jaxpr_axes(
-        cond_jaxpr, size, cconst_dims + carry_dims, (None,),
+        cond_jaxpr, axis_size, cconst_dims + carry_dims, (None,),
         axis_name=axis_name, main_type=main_type)
 
   # To prepare the `init` to the `while_p`, we broadcast values if they are
@@ -442,7 +441,7 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
   new_init = []
   for x, old_axis, new_axis in zip(init, init_dims, carry_dims):
     if old_axis is batching.not_mapped and new_axis is not batching.not_mapped:
-      new_init.append(batching.broadcast(x, size, new_axis))
+      new_init.append(batching.broadcast(x, axis_size, new_axis))
     elif old_axis is batching.not_mapped and new_axis is batching.not_mapped:
       new_init.append(x)
     else:
@@ -584,7 +583,7 @@ def _while_transpose_error(*_, **kwargs):
                    "lax.while_loop or lax.fori_loop. "
                    "Try using lax.scan instead.")
 
-while_p = core.Primitive('while')
+while_p = core.AxisPrimitive('while')
 while_p.multiple_results = True
 while_p.def_impl(partial(xla.apply_primitive, while_p))
 while_p.def_abstract_eval(_while_loop_abstract_eval)
@@ -592,7 +591,7 @@ ad.primitive_jvps[while_p] = _while_loop_jvp
 pe.custom_partial_eval_rules[while_p] = _while_partial_eval
 xla.initial_style_translations[while_p] = _while_loop_translation_rule
 ad.primitive_transposes[while_p] = _while_transpose_error
-batching.initial_style_batchers[while_p] = _while_loop_batching_rule
+batching.axis_primitive_batchers[while_p] = _while_loop_batching_rule
 pe.partial_eval_jaxpr_custom_rules[while_p] = pe.partial_eval_jaxpr_custom_rule_not_implemented
 
 
@@ -817,8 +816,7 @@ def _bcast_select(pred, on_true, on_false):
     pred = lax.broadcast_in_dim(pred, np.shape(on_true), idx)
   return lax.select(pred, on_true, on_false)
 
-def _cond_batching_rule(args, dims, axis_name, main_type, branches, linear):
-  size, = {x.shape[d] for x, d in zip(args, dims) if d is not batching.not_mapped}
+def _cond_batching_rule(axis_size, axis_name, main_type, args, dims, branches, linear):
   index, *ops = args
   index_dim, *op_dims = dims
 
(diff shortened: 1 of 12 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -2425,15 +2431,15 @@ def _linear_solve_batching_rule(args, dims, axis_name, main_type, const_lengths,
...
+linear_solve_p = core.AxisPrimitive('custom_linear_solve')
 linear_solve_p.multiple_results = True
 linear_solve_p.def_impl(_custom_linear_solve_impl)
 linear_solve_p.def_abstract_eval(_linear_solve_abstract_eval)
 ad.primitive_jvps[linear_solve_p] = _custom_linear_solve_jvp
 xla.initial_style_translations[linear_solve_p] = \
     xla.lower_fun_initial_style(_custom_linear_solve_impl)
 ad.primitive_transposes[linear_solve_p] = _linear_solve_transpose_rule
-batching.initial_style_batchers[linear_solve_p] = _linear_solve_batching_rule
+batching.collective_rules[linear_solve_p] = _linear_solve_batching_rule
```
apaszke, line 2442: Can we rename `collective_rules` to something else? Perhaps `axis_primitive_batchers`? It's not just about collectives anymore

jax/core.py
```diff
@@ -1802,7 +1804,7 @@ def subst_axis_names_jaxpr(jaxpr: Union[Jaxpr, ClosedJaxpr], subst: AxisSubst):
   if isinstance(jaxpr, ClosedJaxpr):
     consts = jaxpr.consts
     jaxpr = jaxpr.jaxpr
-  var_map: Dict[Var, Var] = {}
+  var_map: Dict[Var, Var] = {unitvar: unitvar}
```
apaszke, line 1807: Why this change?

jax/interpreters/batching.py
```diff
@@ -122,27 +122,34 @@ def lift(self, val):
...
+      # reconstruct it from the information we have available
+      axis_sizes = {x.shape[d] for x, d in zip(vals, dims) if d is not not_mapped}
+      assert len(axis_sizes) == 1
+      axis_size, = axis_sizes
+      return core.AxisEnvFrame(self.axis_name, axis_size, self.main)
+    return core.axis_frame(self.axis_name)
 
   def process_primitive(self, primitive, tracers, params):
     vals_in, dims_in = unzip2((t.val, t.batch_dim) for t in tracers)
+    frame = self.get_frame(vals_in, dims_in)
```
apaszke, line 145: I think that this might fail if one has a batching trace with `no_axis_name` and a primitive that has no batched inputs (the short-circuit case below).

jax/interpreters/batching.py
```diff
@@ -122,27 +122,34 @@ def lift(self, val):
...
     if (primitive in collective_rules and
           _main_trace_for_axis_names(self.main, core.used_axis_names(primitive, params))):
-      frame = core.axis_frame(self.axis_name)
       val_out, dim_out = collective_rules[primitive](frame, vals_in, dims_in, **params)
     elif all(bdim is not_mapped for bdim in dims_in):
       return primitive.bind(*vals_in, **params)
     else:
-      batched_primitive = self.get_primitive_batcher(primitive)
+      batched_primitive = self.get_primitive_batcher(primitive, frame)
       val_out, dim_out = batched_primitive(vals_in, dims_in, **params)
```
apaszke, line 153: Instead of adding `frame` to `get_primitive_batcher`, couldn't we just remove the `_main_trace_for_axis_name` condition from the first `if` case?

tests/api_test.py
```diff
@@ -5018,6 +5032,23 @@ def foo(x):
     expected = 2. * jnp.cos(jnp.arange(3.))
     self.assertAllClose(ans, expected, check_dtypes=False)
 
+  def test_initial_style_vmap_with_collective(self):
+
+    @api.custom_vjp
```
apaszke, line 5037: Is `custom_vjp` always initial style? I thought it only becomes that after tracing

jax/experimental/pjit.py
```diff
@@ -457,11 +457,14 @@ def _pjit_translation_rule(c, axis_env, in_nodes, name_stack, backend, name,
...
 def _pjit_batcher(insert_axis,
+                  frame,
                   vals_in, dims_in,
                   axis_name, main_type,
                   jaxpr, in_axis_resources, out_axis_resources,
                   resource_env, donated_invars, name):
-  axis_size, = {x.shape[d] for x, d in zip(vals_in, dims_in) if d is not batching.not_mapped}
+  axis_size = frame.axis_size
+  axis_name = frame.name
+  main_type = frame.main_trace.trace_type
```
apaszke, line 467: Honestly I think it would be better to make `collective_rules` have the same signature as `initial_style_batchers` used to have.  Taking in `frame` causes a lot of code duplication, as can be seen here, and exposes lots of implementation details from `core`. It would be better to do all of those unpacks in `batching.py`.

jax/core.py
sharadmv: This PR enables calling `subst_axis_names_jaxpr` on control flow primitive JAXprs, which apparently have unitvars in them. This function previously could not handle unitvars even though the substitution is trivial.

jax/_src/lax/control_flow.py

jax/interpreters/batching.py
sharadmv: Good catch. Fixed.

jax/interpreters/batching.py
sharadmv: I don't think so. There are axis primitives (e.g. psum I believe) that have a standard primitive batcher in addition to an axis primitive rule. In this case, I think the logic needs to follow:
1. If it's not an axis primitive, use the standard primitive batcher
2. If it is an axis primitive (collective or initial style):
  * If the axis name matches, use the axis primitive batcher (overlapping case)
  * If the axis name does not match:
     * If there is a standard batcher, use it (collectives case)
     * Otherwise, use the axis primitive batcher (initial style primitives case)

I've made this more explicit in `process_primitive`, PTAL.

tests/api_test.py
sharadmv: Right, I wrapped it in a eval_jaxpr(make_jaxpr(f)).

jax/interpreters/batching.py
```diff
@@ -123,22 +123,36 @@ def sublift(self, val):
...
-      return partial(initial_style_batchers[primitive],
-                    axis_name=self.axis_name,
-                    main_type=self.main.trace_type)
     try:
       return primitive_batchers[primitive]
     except KeyError as err:
       msg = "Batching rule for '{}' not implemented"
       raise NotImplementedError(msg.format(primitive)) from err
 
+  def get_axis_primitive_batcher(self, primitive, frame):
```
sharadmv, line 132: The reason I had to add this method is because it needs to be overridden by `SPMDBatchTrace` in to allow `spmd_primitive_batchers` to have the correct behavior.

jax/core.py
apaszke: I'm pretty sure it did handle unit vars. Check out the `if` in `subst_axis_names_var`. I'd rather avoid adding this here if possible.

jax/interpreters/batching.py
```diff
@@ -123,22 +123,36 @@ def sublift(self, val):
...
+    return core.axis_frame(self.axis_name)
+
   def process_primitive(self, primitive, tracers, params):
     vals_in, dims_in = unzip2((t.val, t.batch_dim) for t in tracers)
-    if (primitive in collective_rules and
-          _main_trace_for_axis_names(self.main, core.used_axis_names(primitive, params))):
-      frame = core.axis_frame(self.axis_name)
-      val_out, dim_out = collective_rules[primitive](frame, vals_in, dims_in, **params)
+    is_axis_primitive = primitive in axis_primitive_batchers
+    is_initial_style = is_axis_primitive and primitive not in primitive_batchers
```
apaszke, line 149: I don't think this is correct (it might be right now, but it doesn't look like a generalizable principle). Why do we need it? It's not like we have to actually call the initial style batching rules if there are no batched inputs _and_ the axis is not used.

apaszke (commented):
Mostly LGTM, but there are two things I'd like to revisit before we merge this. We also definitely need to cache the results of `core.used_axis_names` but I can add that later if you'd prefer.

jax/core.py
sharadmv: It handles `unitvars` in the inputs but not the outputs. This line
```
outvars: List[Atom] = [v if isinstance(v, Literal) else var_map[v] for v in jaxpr.outvars]
```
fails if there are any unitvars in `jaxpr.outvars`. I wasn't sure of a simpler solution than this.

jax/interpreters/batching.py
sharadmv: It seems like although "collectives" and "initial style primitives" can share the same batching rule, there are still some differences between how they operate. 

1. collectives - have both standard batchers and axis batchers. Axis batcher is triggered when the frame name matches the axis name in the params. If there are batch dims, standard batcher is triggered and otherwise short circuit.
2. initial style - always trigger axis batcher if there are batch dims, short circuit otherwise (there is no ordinary batching rule so we have to avoid taking that code path).

The solutions I see are either:
1. combine standard and axis batchers into `get_primitive_batcher`. This involves passing `frame` into `get_primitive_batcher`.
2. Check if the primitive has both a `standard_batcher` and an `axis_primitive_batcher`. If it does, it's a collective otherwise it's an initial style primitive. That is the current solution.

I've added in the short-circuiting condition to address the issue you pointed out, but as you said, I'm not sure how to make this into a generalizable principle while there is still this difference between collectives and initial style primities.

jax/experimental/pjit.py
apaszke: Looks good to me 🤷‍♂️

jax/core.py
apaszke: Ohh I see, good point. Is that line necessary for the tests for this PR to pass? If no then let's submit it as a separate change.

jax/interpreters/batching.py
apaszke: I see the problem now. Would the following workflow work?

1. If the primitive has an axis_primitive_batcher *and* mentions the name then call the axis primitive rule. 
2. If there are no BatchTracers in inputs, short circuit
3. Call the primitive_batcher if available, falling back to an axis_primitive_batcher if unavailable

jax/core.py
sharadmv: It is necessary unfortunately.

jax/interpreters/batching.py
sharadmv: Right so this is the previous solution I implemented. The main issue is the 3rd point; we need the frame to fall back to the `axis_primitive_batcher` so I had previously augmented the `get_primitive_batcher` method to take in the `frame`. Do you think there's a better option than that?

apaszke: I think it's ok for `get_primitive_batcher` to take in `frame` if it's also supposed to handle `axis_primitive_batchers`.

sharadmv: Done.

Issue #7737, opened 2021-08-27
Regularize jax.numpy API


jakevdp:
I've been picking away at this for the past year or so (since  #3038), but I wanted to track it a bit more formally here.

We want the jax.numpy API to have two properties:

- static arguments should generally be checked with `core.concrete_or_error()`. This helps localize concretization errors and provides a more uniform user experience.

- dynamic arguments should generally be validated with `_check_arraylike`. This is because passing lists to jax functions can be a quiet source of performance degradation, because lists are treated as pytrees, for example:
```python
In [1]: import jax.numpy as jnp                                                                                                                              

In [2]: from jax import jit                                                                                                                                  

In [3]: f = jit(lambda x: jnp.mean(jnp.asarray(x)))

In [4]: %timeit f([float(i) for i in range(1000)])                                                                                                           
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
1.79 ms ± 556 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit f(jnp.array([float(i) for i in range(1000)]))                                                                                                
375 µs ± 4.76 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
This change may be problematic because it breaks users who may be passing lists to `jax.numpy` functions, but this input type restriction has [long been documented](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#non-array-inputs-numpy-vs-jax). Still, with each change I plan to run a full set of tests and fix downstream packages if necessary.

shoyer:
Another example of this would be being strict about requiring tuples (rather than lists) for all static arguments like `axes`. This came up for `jax.numpy.fft` in https://github.com/google/jax/pull/7758

shailesh1729:
This breaking change in 0.2.24 actually helped my code.

I had a piece of code as:

```python
 coeffs = jnp.take(coeffs, indices=tuple(range(start, start+n)), axis=axis)
```
which broke with the upgrade last night in CI. 

I replaced it with 

```python
coeffs = jnp.take(coeffs, indices=jnp.arange(start, start+n), axis=axis)
```
and then JAX was happy. And I realize, this version is actually more appropriate. I shouldn't have used Python `range` in numerical code in the first place.

Being more restrictive in JAX API indeed is helpful.

jakevdp:
[...]

PR #7747, opened 2021-08-27
Tighten input validation in jax.numpy API

jakevdp:
Part of #7737


jakevdp:
I'm finding that nearly all downstream breakages resulting from this change come from two functions: `jnp.take`, to which people often pass lists of indices, and `jnp.allclose`, which is used primarily within tests comparing arrays to lists.

I'm tempted to push those two changes into their own PRs and merge the other less impactful changes in bulk.

jakevdp:
OK, I moved `allclose`/`isclose` and `take`/`take_along_axis` to #7802 and #7803, respectively.

PR #7750, opened 2021-08-28
README: mention that CUDA installation is for linux only

jakevdp:
Trying this on other operating systems causes pip to cycle through past JAX versions until it hits JAX 0.2.15 and then hit the error `Packages installed from PyPI cannot depend on packages which are not also hosted on PyPI.` from #7083.

I don't know of any way we can improve that error path, but at least we can document it.


mattjj (approved):
Just saw this! Sorry for the holdup.

Thanks for the fix!

PR #7751, opened 2021-08-29
Implement np.polyfit

avani17101:
Implementation of [```np.polyfit``` ](https://numpy.org/doc/stable/reference/generated/numpy.polyfit.html) from https://github.com/google/jax/issues/70

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1368,6 +1368,76 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
 def isrealobj(x):
   return not iscomplexobj(x)
 
+_POLYFIT_DOC = """\
+Unlike NumPy's implementation of polyfit, :py:func:`jax.numpy.polyfit` will not warn on rank reduction, which indicates an ill conditioned matrix
+Also, it works best on rcond <= 10e-3 values.
+"""
+@_wraps(np.polyfit, lax_description=_POLYFIT_DOC )
+def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
+  _check_arraylike("polyfit", x, y)
+  deg = core.concrete_or_error(int, deg, "deg must be int")
+  order = deg + 1
+  # check arguments
+  if deg < 0:
+    raise ValueError("expected deg >= 0")
+  if x.ndim != 1:
+    raise TypeError("expected 1D vector for x")
+  if x.size == 0:
+    raise TypeError("expected non-empty vector for x")
+  if y.ndim < 1 or y.ndim > 2:
+    raise TypeError("expected 1D or 2D array for y")
+  if x.shape[0] != y.shape[0]:
+    raise TypeError("expected x and y to have same length")
+
+  # set rcond
+  if rcond is None:
+    rcond = len(x)*finfo(x.dtype).eps
+  rcond = core.concrete_or_error(float, rcond, "rcond must be float")
+  # set up least squares equation for powers of x
+  lhs = vander(x, order)
+  rhs = y
+
+  # apply weighting
+  if w is not None:
+    _check_arraylike("polyfit", w)
+    w, = _promote_dtypes_inexact(w)
+    if w.ndim != 1:
+      raise TypeError("expected a 1-d array for weights")
+    if w.shape[0] != y.shape[0]:
+      raise TypeError("expected w and y to have the same length")
+    lhs *= w[:, newaxis]
+    if rhs.ndim == 2:
+      rhs *= w[:, newaxis]
+    else:
+      rhs *= w
+
+  # scale lhs to improve condition number and solve
+  scale = sqrt((lhs*lhs).sum(axis=0))
+  lhs /= scale[newaxis,:]
+  from . import linalg
+  c, resids, rank, s = linalg.lstsq(lhs, rhs, rcond)
+  c = (c.T/scale).T  # broadcast scale coefficients
+
+  if full:
+    return c, resids, rank, s, rcond
+  elif cov:
+    Vbase = linalg.inv(dot(lhs.T, lhs))
+    Vbase /= outer(scale, scale)
+    if cov == "unscaled":
+      fac = 1
+    else:
+      if len(x) <= order:
+        raise ValueError("the number of data points must exceed order "
+                            "to scale the covariance matrix")
+      fac = resids / (len(x) - order)
+      fac = fac[0] #making np.array() of shape (1,) to int
+    if y.ndim == 1:
+      return c, Vbase * fac
+    else:
+      return c, Vbase[:,:, newaxis] * fac
+  else:
+    return c
+
 
 @_wraps(np.reshape, lax_description=_ARRAY_VIEW_DOC)
 def reshape(a, newshape, order="C"):
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1379,10 +1379,41 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     tol = max(jtu.tolerance(lhs_dtype, tol_spec),
               jtu.tolerance(rhs_dtype, tol_spec))
     # TODO(phawkins): there are float32/float64 disagreements for some inputs.
-    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False,
-                            tol=tol)
-    self._CompileAndCheck(jnp_fun, args_maker, check_dtypes=False, atol=tol,
-                          rtol=tol)
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False, tol=tol)
+    self._CompileAndCheck(jnp_fun, args_maker, check_dtypes=False, atol=tol, rtol=tol)
+
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_{}_deg={}_rcond={}_full={}_w={}_cov={}".format(
+          jtu.format_shape_dtype_string(shape, dtype),
+          deg,
+          rcond,
+          full,
+          w,
+          cov),
+      "shape": shape, "dtype": dtype, "deg": deg,
+      "rcond": rcond, "full": full, "w":w, "cov":cov}
+      for dtype in [dt for dt in float_dtypes if dt not in [jnp.float16, jnp.bfloat16]]
+      for shape in [shape for shape in one_dim_array_shapes if shape != (1,)]
+      for deg in [1, 2, 3]
+      for rcond in [None, -1, 10e-3, 10e-5, 10e-10]
+      for full in [False, True]
+      for w in [False, True]
+      for cov in [False, True, "unscaled"]))
+  def testPolyfit(self, shape, dtype, deg, rcond, full, w, cov):
+    rng = jtu.rand_default(self.rng())
+    tol_spec = {np.float32: 1e-4, np.float64: 1e-13, np.complex64: 1e-5}
+    if jtu.device_under_test() == "tpu":
+      tol_spec[np.float32] = tol_spec[np.complex64] = 2e-1
+    tol = jtu.tolerance(dtype, tol_spec)
+    _w = lambda a: abs(a) if w else None
+    args_maker = lambda: [rng(shape, dtype), rng(shape, dtype), rng(shape, dtype)]
+    jnp_fun = lambda x, y, a: jnp.polyfit(x, y, deg=deg, rcond=rcond, full=full, w=_w(a), cov=cov)
+    np_fun = jtu.ignore_warning(
+      message="Polyfit may be poorly conditioned*")(lambda x, y, a: np.polyfit(x, y, deg=deg, rcond=rcond, full=full, w=_w(a), cov=cov))
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False, tol=tol)
+    self._CompileAndCheck(jnp_fun, args_maker, check_dtypes=False, atol=tol, rtol=tol)
+
 
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_{}_amin={}_amax={}".format(
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -294,6 +294,7 @@ Not every function in NumPy is implemented; contributions are welcome!
     poly
     polyadd
     polyder
+    polyfit
     polyint
     polymul
     polysub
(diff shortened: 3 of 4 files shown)


avani17101:
@googlebot I signed it!

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+    """
+    Issued by `polyfit` when the Vandermonde matrix is rank deficient.
+    For more information, a way to suppress the warning, and an example of
+    `RankWarning` being issued, see `polyfit`.
+    """
+    pass
+
+@_wraps(np.polyfit)
+def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
+  _check_arraylike("polyfit", x, y, deg)
```
jakevdp, line 1382: I don't think `deg` needs to be `arraylike` – it's an integer, correct?

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+  x = core.concrete_or_error(None, x, "The error arose in polyfit()")
+  y = core.concrete_or_error(None, y, "The error arose in polyfit()")
```
jakevdp, line 1384: I don't think there's any need to ensure `x` and `y` are concrete. Looking through the code below, it seems like they could be traced quantites, except perhaps for the `rank` check (which I would consider removing in order to allow JIT compilation)

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+  order = int(deg) + 1
```
jakevdp, line 1385: Should use
```
deg = core.concrete_or_error(int, deg, msg)
order = deg + 1
```
`deg` and `order` must be concrete, because their values are used in logical statements and to control the shapes of arrays.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+  if rcond is None:
+      rcond = len(x)*finfo(x.dtype).eps
+
+  # set up least squares equation for powers of x
+  lhs = vander(x, order)
+  rhs = y
+
+  # apply weighting
+  if w is not None:
+      w = core.concrete_or_error(None, w, "The error arose in polyfit()") + 0.0
```
jakevdp, line 1408: I don't think `w` needs to be concrete, so I would remove this.

It looks like it does need to be arraylike, so `_check_arraylike` would be applicable.

If you need to ensure it's floating point, use `_promote_dtypes_inexact` rather than adding 0.0.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+          rhs *= w
+
+  # scale lhs to improve condition number and solve
+  scale = lax.sqrt((lhs*lhs).sum(axis=0))
+  lhs /= scale
+  c, resids, rank, s = lax.lstsq(lhs, rhs, rcond)
+  c = (c.T/scale).T  # broadcast scale coefficients
+
+  # warn on rank reduction, which indicates an ill conditioned matrix
+  if rank != order and not full:
```
jakevdp, line 1426: This statement will have to be removed for this function to be compatible with JIT (`rank` is derived from traced quantities, so we can't have logic based on its value). I'd consider removing this block and noting in the updated doc string that unlike numpy, the function will not warn if the input is poorly conditioned.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,87 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+      if cov == "unscaled":
+          fac = 1
+      else:
+          if len(x) <= order:
+              raise ValueError("the number of data points must exceed order "
+                                "to scale the covariance matrix")
+          # note, this used to be: fac = resids / (len(x) - order - 2.0)
+          # it was deciced that the "- 2" (originally justified by "Bayesian
+          # uncertainty analysis") is not was the user expects
+          # (see gh-11196 and gh-11197)
```
jakevdp, line 1444: This comment seems to be about the numpy implementation; it's probably not relevant here.

tests/lax_numpy_test.py
```diff
@@ -1384,6 +1384,34 @@ def np_fun(lhs, rhs):
...
+    for full in [False, True]
+    for w in [None]
+    for cov in [False, True]))
+  def testpolyfit(self, shape1, dtype1, shape2, dtype2, deg, rcond, full, w, cov):
+    rng = jtu.rand_default(self.rng())
+    args_maker = lambda: [rng(shape1, dtype1), rng(shape2, dtype2)]
+    jnp_fun = lambda x, y: jnp.polyfit(x, y, deg=deg, rcond=rcond, full=full, w=w, cov=cov)
+    np_fun = lambda x, y: np.polyfit(x, y, deg=deg, rcond=rcond, full=full, w=w, cov=cov)
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False)
+
```
jakevdp, line 1413: We should add a `self._CompileAndCheck` as well. You'll have to make the other changes suggested above in order for it to pass.

jakevdp (changes requested):
Looks great, thanks for the contribution!

I've suggested a few little modifications around which variables to treat as concrete & arraylike in order to make this implementation compatible with JAX transforms like JIT. Give it a try and let me know what you think

avani17101:
> Looks great, thanks for the contribution!
> 
> I've suggested a few little modifications around which variables to treat as concrete & arraylike in order to make this implementation compatible with JAX transforms like JIT. Give it a try and let me know what you think

thanks for reviewing, made the changes, can you please review.

jakevdp:
You can run the lint checks locally as well; see https://jax.readthedocs.io/en/latest/contributing.html?highlight=pre-commit#linting-and-type-checking

avani17101:
> You can run the lint checks locally as well; see https://jax.readthedocs.io/en/latest/contributing.html?highlight=pre-commit#linting-and-type-checking

thanks a lot, fixed it :)

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,82 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
 def isrealobj(x):
   return not iscomplexobj(x)
 
+@partial(jit, static_argnums=1)
+class _RankWarning(UserWarning):
```
jakevdp, line 1372: Looks like this is no longer used – I think we can remove it.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,82 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+    For more information, a way to suppress the warning, and an example of
+    `RankWarning` being issued, see `polyfit`.
+    """
+    pass
+
+@_wraps(np.polyfit)
+def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
+  """
+  Note: Unlike numpy, the function will not warn on rank reduction, which indicates an ill conditioned matrix
+  """
```
jakevdp, line 1384: This docstring will be overwritten by `_wraps`. You can put this info in the `lax_description` argument of `_wraps`; there are a few examples elsewhere in this file.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,82 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+  if rcond is None:
+      rcond = len(x)*finfo(x.dtype).eps
+
+  # set up least squares equation for powers of x
+  lhs = vander(x, order)
+  rhs = y
+
+  # apply weighting
+  if w is not None:
+      w = _promote_dtypes_inexact(_check_arraylike("polyfit", w))
```
jakevdp, line 1410: I think `_check_arraylike` returns `None`, so this will have to be done in two lines.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1368,6 +1368,82 @@ def gradient(f, *varargs, axis: Optional[Union[int, Tuple[int, ...]]] = None,
...
+def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
+  """
+  Note: Unlike numpy, the function will not warn on rank reduction, which indicates an ill conditioned matrix
+  """
+  _check_arraylike("polyfit", x, y)
+  deg = core.concrete_or_error(int, deg, "deg must be int")
+  order = deg + 1
+  # check arguments.
+  if deg < 0:
+      raise ValueError("expected deg >= 0")
```
jakevdp, line 1390: Here and throughout, please use 2-space indentation for consistency.

tests/lax_numpy_test.py
```diff
@@ -1379,10 +1379,40 @@ def np_fun(lhs, rhs):
...
+          w,
+          cov),
+      "shape": shape, "dtype": dtype, "deg": deg,
+      "rcond": rcond, "full": full, "w":w, "cov":cov}
+      for dtype in [dt for dt in float_dtypes if dt not in [jnp.float16, jnp.bfloat16]]
+      for shape in [shape for shape in one_dim_array_shapes if shape != (1,)]
+      for deg in jnp.array([1, 2, 3])
+      for rcond in [None]
+      for full in [False, True]
+      for w in [None]
```
jakevdp, line 1400: We should test with and without weights for completeness – for one idea about how to approach this, see how weights are passed in `testHistogram`

jakevdp (changes requested):
Looks great, thanks for responding to all those changes! Just a couple more comments below - thanks!

jax/_src/numpy/lax_numpy.py
avani17101: To make it consistent with other functions in the file, changing it to one space indentation

avani17101:
For rcond >= 1, it is failing test(```tol``` goes way up), I see ```lingalg.lstsq``` being responsible. It has some differences from ```numpy.lstsq``` implementation

avani17101:
> Looks great, thanks for responding to all those changes! Just a couple more comments below - thanks!

Thanks a lot for reviewing, made the suggested changes. Also added test for ```rcond``` values other than None.

tests/lax_numpy_test.py
```diff
@@ -1379,10 +1379,42 @@ def np_fun(lhs, rhs):
...
+      for dtype in [dt for dt in float_dtypes if dt not in [jnp.float16, jnp.bfloat16]]
+      for shape in [shape for shape in one_dim_array_shapes if shape != (1,)]
+      for deg in jnp.array([1, 2, 3])
+      for rcond in [None, -1, 0.5, 10e-2, 10e-5, 10e-10]
+      for full in [False, True]
+      for w in [False, True]
+      for cov in [False, True]))
+  def testPolyfit(self, shape, dtype, deg, rcond, full, w, cov):
+    rng = jtu.rand_default(self.rng())
+    jnp_fun = lambda lhs, rhs: jnp.inner(lhs, rhs)
```
jakevdp, line 1405: Delete? This `jnp_fun` appears to be unused.

tests/lax_numpy_test.py
```diff
@@ -1379,10 +1379,42 @@ def np_fun(lhs, rhs):
...
+          cov),
+      "shape": shape, "dtype": dtype, "deg": deg,
+      "rcond": rcond, "full": full, "w":w, "cov":cov}
+      for dtype in [dt for dt in float_dtypes if dt not in [jnp.float16, jnp.bfloat16]]
+      for shape in [shape for shape in one_dim_array_shapes if shape != (1,)]
+      for deg in jnp.array([1, 2, 3])
+      for rcond in [None, -1, 0.5, 10e-2, 10e-5, 10e-10]
+      for full in [False, True]
+      for w in [False, True]
+      for cov in [False, True]))
```
jakevdp, line 1402: We should also test `cov="unscaled"` to get full test coverage

jakevdp (changes requested):
Getting very close!

After fixing these last two things, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests). Let me know if you need tips on the best way to do that.

avani17101:
> Getting very close!
> 
> After fixing these last two things, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests). Let me know if you need tips on the best way to do that.

Squashed the commit to one, does it seem fine?

jakevdp (approved):
Looks great! Thanks for all the great work here during code review 😁 

I'm going to run some additional tests, and then hopefully we can get this merged soon.

jakevdp:
It looks like there's an error due to numpy outputting a warning (warnings are treated as errors in the JAX test suite):
```
        # warn on rank reduction, which indicates an ill conditioned matrix
        if rank != order and not full:
            msg = "Polyfit may be poorly conditioned"
>           warnings.warn(msg, RankWarning, stacklevel=4)
E           numpy.RankWarning: Polyfit may be poorly conditioned
```
The best way to deal with this is to use the `jtu.ignore_warning()` decorator for your `np_fun`; you can see examples of this elsewhere in `lax_numpy_test.py`.

If you want to test this locally, you can run the test case with more generated cases. For example:
```bash
JAX_NUM_GENERATED_CASES=50 pytest -n auto tests/lax_numpy_test.py -k testPolyfit
```

jakevdp:
And please be sure to rebase so the updated test is part of the single commit - thanks!

avani17101:
> And please be sure to rebase so the updated test is part of the single commit - thanks!

made the changes. Thanks a lot for the valuable suggestions.

tests/lax_numpy_test.py
```diff
@@ -1379,10 +1379,41 @@ def np_fun(lhs, rhs):
...
+          deg,
+          rcond,
+          full,
+          w,
+          cov),
+      "shape": shape, "dtype": dtype, "deg": deg,
+      "rcond": rcond, "full": full, "w":w, "cov":cov}
+      for dtype in [dt for dt in float_dtypes if dt not in [jnp.float16, jnp.bfloat16]]
+      for shape in [shape for shape in one_dim_array_shapes if shape != (1,)]
+      for deg in jnp.array([1, 2, 3])
```
jakevdp, line 1398: This should be `for deg in [1, 2, 3]`. Trying to use `jnp.array` in the initial parsing of the files leads to errors in some backends because XLA has not yet been initialized.

avani17101: made the change

jakevdp (approved):

jakevdp:
Thanks for sticking with this through all the changes!

avani17101:
> Thanks for sticking with this through all the changes!

thanks for your valuable feedback throughout :)

PR #7761, opened 2021-08-30
fix PEP 561 problems in jax.numpy init file

machineko:
Add simple fix for problem with PEP 561 considering this document [document](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md)
"In particular, imported symbols are not assumed to be re-exported unless they appear in __all__ or use a redundant form of import"

autocompletion before fix

jaxautocompl

after fix
after


machineko:
#7762

jakevdp:
Thanks! We were discussing this in #7570, but I think the preferred fix would be explicit imports, to avoid ambiguity in whether the lists are synced; see https://github.com/google/jax/pull/7606.

machineko:
Ahh didnt see that was looking for autocomplete and stuff like that 😄

jakevdp:
I've stalled on merging that because I hate how redundant it is... but I guess that's required by PEP 484, so I think there's probably no way around it.

I think I'll update that PR to make similar changes to other namespaces.

machineko:
Hey np. Take it over I didn't see u guys were working on that so I create a fast PR as I was only searching phrases like auto completion and vscode.

I'll be happy with working vscode autocompletion for jax and can agree that all export method looks ugly but it works 😅

PR #7772, opened 2021-09-01
Add unary xeinsum & allow named axis reductions for unary/binary xeinsums

bloops:
The main functional change apart from accepting unary xeinsum is using psum on the named axes which should be reduced. There were some unimplemented/failing cases in binary xeinsums with only subscripts which I have added as tests. The axes to contract/reduce are computed slightly differently now.

In this change I have also extended to XeinsumSpecParser to be less restrictive so that it accepts named axes and subscripts in any order.

It still remains open to support diagonals and traces for unary xeinsums. ('ii->i' and 'ii->').

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -36,7 +36,7 @@ from jax._src.lax import lax
 from jax._src.lax import slicing
 from jax._src.numpy import lax_numpy
 import jax._src.util as util
-from jax._src.util import unzip2, prod, canonicalize_axis, safe_map, moveaxis
+from jax._src.util import unzip2, prod, canonicalize_axis, safe_map, safe_zip, moveaxis
 from jax._src.lib.mlir import ir
 from jax._src.lib.mlir.dialects import mhlo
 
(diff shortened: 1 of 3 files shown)


jax/_src/lax/parallel.py
```diff
@@ -405,50 +405,94 @@ def pdot(x, y, axis_name, pos_contract=((), ()), pos_batch=((), ())):
...
+    subs_reduce = list(set(in_subs) - set(out_subs))
+    subs_reduce_axes = [in_subs.index(n) for n in subs_reduce]
+    x = lax._reduce_sum(x, subs_reduce_axes)
+
+    # reduce named axes.
+    named_reduce_axes = list(set(in_named) - set(out_named))
+    for ax in named_reduce_axes:
+      x = psum(x, ax)
+    return x
+  elif len(operands) == 2:
```
apaszke, line 434: Is this casing at the top level really necessary? It would be better to stop assuming that the output of `XeinsumSpecParser` has only one or two output but process it as a list instead. At the moment we will effectively have to review and maintain two parallel implementations that could be handled by a single algorithm.

bloops: Yeah -- I think that is a good idea, let me take a look.

BTW is there any plan to use opt_einsum to find a contraction path like in regular einsum (defined in lax_numpy.py)? That way we'd only ever have to worry about unary or binary einsums. Or is it that depending on 'axis_resources' you'd need a different cost metric for the path contraction that perhaps opt_einsum cannot handle?

apaszke: Great question! In the future we definitely should be using opt_einsum. We might need to do a little munging on the einsum strings to get them to a form that's compatible with that library, but it should be doable. But there's no need to include it in this PR. For now we can encode any (possibly silly) strategy for n-ary contractions.

bloops:
Sorry, I think I have messed up the rebasing/merging process (perhaps I should abandon this PR and start another?) @apaszke

jax/_src/lax/parallel.py
bloops: Thanks! For this PR I am not adding n-ary contractions to keep the scope limited. (Maybe in a future PR)

This should be ready for another review. Please take another look!

jax/_src/lax/parallel.py
```diff
@@ -405,50 +405,77 @@ def pdot(x, y, axis_name, pos_contract=((), ()), pos_batch=((), ()),
...
+    subs_reduce_axes = [in_subs.index(n) for n in subs_reduce]
+    if subs_reduce_axes:
+      x = lax._reduce_sum(x, subs_reduce_axes)
+      for i in sorted(subs_reduce_axes, reverse=True):
+        del all_in_subs[idx][i]
+
+    # reduce named axes.
+    named_reduce_axes = list(set(in_named) - {*out_named, *other_named})
+    for ax in named_reduce_axes:
+      x = psum(x, ax)
```
apaszke, line 452: By the way you can do a `psum` with a tuple of both named and positional axes and it should work. That could simplify the code.

apaszke: Also, you delete the reduced axes from `all_in_subs` above, but you don't remove it from `all_in_named` here. Is that intentional?

jax/_src/lax/parallel.py
```diff
@@ -500,26 +526,32 @@ def maybe_take(self, char: str, on_eof: bool = False):
...
+            raise ValueError("Unterminated named axis brace")
+          axis_name = self.parse_axis_name()
+          names.append(axis_name)
+      else:
+        # we're at a subscript
+        subscript = self.parse_subscript()
+        subscripts.append(subscript)
+
+    # we hit eof, so we don't expect a subsequent argument.
+    return False, (subscripts, names)
```
apaszke, line 554: Was the previous logic buggy somehow?

apaszke (approved):
Looks very good now, thanks! I have one question about removing names from `all_in_named` as we're `psum`ming them, but LGTM otherwise.

jax/_src/lax/parallel.py
bloops: Yes, that was intentional. Removing from all_in_subs is necessary because we need to find the correct position for the subscript later on when contracting. But for all_in_named that's not necessary as we can just refer to the axis name in pdot.

It's another reason why named axes are cool (they're oblivious to other axes being removed)

bloops: Re: doing psum with both named and positional axes -- thats a good point!

However, I thought `psum` can only take one named axis at a time? (While it can accept multiple positional axes in the same call) If that is true, we can't simplify much here.

jax/_src/lax/parallel.py
bloops: I don't think it was buggy. It just didn't support subscripts and named axes in arbitrary order (just supported named axes at the end) and also it was not raising a ValueError for invalid characters (like having '0' as a subscript would raise the ValueError now)

Specifically, the logic before was:

1. keep consuming subscripts until either eof, an axis brace, a comma or an invalid character is hit.
2. if eof was hit, return the subscripts and (empty) named axes, and `cont = False` (which means no more arguments).
3. ...or if comma was hit, return the subscripts and (empty) named axes, and `cont = True`.
4. ...or if axis brace was hit, parse all the named axes inside `{...}`. And `cont = True/False` depending on whether the next character is a comma.

The change to the parser can be summarized as:
 (a) Expand the original while loop (where each iteration consumed a subscript) to either consume a subscript or a `{...}` block containing multiple axis names in each iteration. This means we now support subscripts, followed by groups of named axes interspersed in any order (so `ab{c,d}e{f,g}h` is also valid input.)
 (b) Remove the responsibility of detecting axis brace or comma from `parse_subscript` (the while loop body takes care of that) so that it can raise a ValueError if it doesn't see a lowercase ascii character.

The other important side-change I can think of is: because the earlier method checked axis braces last, it could do an assert if it failed to parse a subscript or detect a comma. (`assert self.maybe_take('{')`). But now it doesn't do that anymore, as it would go into the `parse_subscript... raise ValueError` code path, and raise an error because it's expecting a subscript.

jax/_src/lax/parallel.py
bloops: Sounds good. I'll try this out.

bloops: Updated with a single psum and also now removing the reduced named axes. Thanks for the timely reviews!

jax/_src/lax/parallel.py
apaszke: Is there any good argument for allowing named axes to be interspersed with subscripts?

bloops: The main argument I have is ergonomics / readability. It's easier to switch to a version without named axes to one with _some_ named axes, but not imposing further constraints on the ordering of subscripts/named axes in the einsum spec. 

It can hurt readability if the user is forced to write `yz{x}` when the natural ordering is `{x}yz`. Of course, here the user might eventually use only named axes, i.e. `{x,y,z}` in which case there is no problem. But I think what might arise in practice is a transition period from no-named-axes to some-named-axes (maybe only eventually to all-named-axes). This is also because, currently, many ops already work with one 'batch' named axis, but very few work with all named axes.

That said, this is not a strong reason - so if you prefer the simpler API contract of named axes appearing together after all the subscripts, I can revert this part of the change.

apaszke (approved):

PR #7773, opened 2021-09-01
Drop compatibility with flatbuffers 1.12.

hawkinsp:
Remove backwards compatibility code.


hawkinsp:
It turns out we can't drop flatbuffers 1.12 support yet because Google uses it internally still.

PR #7775, opened 2021-09-01
Use higher-level API to build FFT flatbuffer descriptor.

hawkinsp:
The higher-level API is not only simpler, it also has the additional advantage of being unchanged between 1.12 and 2.0.


hawkinsp:
Closing, because this PR didn't actually solve the version compatibility problem. I suspect a better fix is to build the descriptor in C++, and then we can drop the Python `flatbuffer` dependency completely.

PR #7781, opened 2021-09-01
change while loop batching fixed point condition

sharadmv:
Adjust the while loop fixed point to address the error in #7063. 

Here's the relevant repro:
```python
import jax
from jax import lax
from jax.experimental import maps

def f(x):
  z = x + lax.axis_index('a')
  y = x + lax.axis_index('b')
  def cond(carry):
    i, x = carry
    return x < 5
  def body(carry):
    i, x = carry
    return i + 1, x + lax.psum(y, 'b')
  return lax.while_loop(cond, body, (0, z))[1]
maps.xmap(f, axis_sizes=dict(a=2, b=10), out_axes=(['a']), in_axes={})(1.)
```

The translation rule for while loop requires that the predicate shape is a prefix for the carry shapes (if predicate is scalar, this is trivially true). However, when predicate is *not* a scalar, we need to make sure that this is the case. Previously, if the predicate was not a scalar, we'd batch every value in the carry and batch the predicate. This would ensure that the predicate shape remains a prefix of the carry shape, satisfying the condition in the while loop. However, this overbatching can cause problems.

In the repro, the batching rule for `while` is triggered twice, once for axis `'a'` and once for axis `'b'`. The body has the `x` value in the carry, which mapped across `'a'`, and closes over `y` (which has a `'b`' axis). However, the input and output carry axes do not include `'b'`, because it is summed out with the `psum`.

With the current batching rule, for `'a'` we properly batch the body and cond of the while loop. In the batching rule for `'b'`, specifically in the fixed point, we instantiate the output for the cond_jaxpr because `bool(cond_jaxpr.avals.out_shape[0])` is `True` (this is because we have batched the cond for `'a'`). This causes us to map the entire `carry`, which adds an axis to the carry for `'b'` when none needs to exist.

What we'd instead like to do is not `instantiate` if the predicate is is non-scalar. We check the `pred_bat` produced by `batching.batch_jaxpr` (which tells us if the predicate is batched or not). If it is, we do need to batch everything and reproduce the `instantiate=True` behavior. If it is not, we need to make sure to have the batched predicate shape still be a prefix of the carry shapes. We accomplish this with some extra bookkeeping and modifications to `batching.batch_jaxpr` to allow us to thread axes in as opposed to bools indicating whether something is batched on the leading dimension.

Co-authored-by: Adam Paszke <apaszke@google.com>

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -373,7 +373,7 @@ def _pred_bcast_select(c, pred, x, y, x_y_aval: core.AbstractValue):
   elif x_y_aval is core.abstract_token:
     return xops.AfterAll(c, [x, y])
   else:
-    assert pred_shape == x_shape[:len(pred_shape)] == y_shape[:len(pred_shape)]
+    assert pred_shape == x_shape[:len(pred_shape)] == y_shape[:len(pred_shape)], (pred_shape, x_shape, y_shape)
     bcast_pred = xops.BroadcastInDim(pred, x_shape, list(range(len(pred_shape))))
     return xops.Select(bcast_pred, x, y)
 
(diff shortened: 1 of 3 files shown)


sharadmv:
This is an updated version of #7206 that fixes some issues that popped up in internal tests.

sharadmv:
PTAL @apaszke

jax/_src/lax/control_flow.py
```diff
@@ -383,43 +383,76 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
...
+    carry_dims = [0] * len(carry_bat)
+    body_jaxpr_batched, _ = batching.batch_jaxpr_axes(
+        body_jaxpr, size, bconst_dims + carry_dims,
+        carry_dims, axis_name=axis_name, main_type=main_type)
+    cond_jaxpr_batched, _ = batching.batch_jaxpr_axes(
+        cond_jaxpr, size, cconst_dims + carry_dims, [0],
+        axis_name=axis_name, main_type=main_type)
+  else:
+    # If the predicate is not batched, we need to make sure that the in axes
+    # match the out axes in the batched `body_jaxpr`. Otherwise the predicate's
```
apaszke, line 423: I'm not sure if I follow this comment. What are in axes and out axes here? Can you please reword this?

jax/_src/lax/control_flow.py
```diff
@@ -383,43 +383,76 @@ def _while_loop_batching_rule(args, dims, axis_name, main_type,
...
+    body_jaxpr_batched, _ = batching.batch_jaxpr_axes(
+        body_jaxpr, size, bconst_dims + carry_dims, carry_dims,
+        axis_name=axis_name, main_type=main_type)
+    # Now we need to rebatch the `cond_jaxpr` according to the dims of the
+    # carry.
+    cond_jaxpr_batched, _ = batching.batch_jaxpr_axes(
+        cond_jaxpr, size, cconst_dims + carry_dims, (None,),
+        axis_name=axis_name, main_type=main_type)
+
+
```
apaszke, line 437: nit: can you please avoid putting two empty lines within a function body

tests/lax_control_flow_test.py
```diff
@@ -2739,5 +2740,30 @@ def side_effecting_scan(carry, val):
...
+    def f(i, x):
+      def cond(carry):
+        i, x = carry
+        return i < 5
+      def body(carry):
+        i, z = carry
+        # Close over const with batch dim = 1
+        return i + 1, z + x
+      return lax.while_loop(cond, body, (i, jnp.ones(3)))[1]
+    jax.vmap(f, in_axes=(0, 1))(jnp.arange(4), jnp.ones((3, 4)))
```
apaszke, line 2766: Does one of those tests catch the breakage that the previous version has caused?

apaszke (approved):

tests/lax_control_flow_test.py
sharadmv: Yes, this test with the closed over value does.

jax/_src/lax/control_flow.py
sharadmv: Done.

jax/_src/lax/control_flow.py

PR #7783, opened 2021-09-02
Don't refer to deprecated jax.ops.index_update in error messages

shoyer:
This came up when a user asked about _how_ exactly they were supposed to use `jax.ops.index_update` :)

I've also updated the docs for ``jax.ops`` to note that ``at[].set()`` is guaranteed to be performed in-place under JIT. Someone who knows XLA well should double check that fact!

### DIFF
--- a/docs/jax.ops.rst
+++ b/docs/jax.ops.rst
@@ -31,7 +31,8 @@ Alternate syntax           Equivalent in-place expression
 =========================  ===================================================
 
 None of these expressions modify the original `x`; instead they return
-a modified copy of `x`.
+a modified copy of `x`. However, inside a :py:func:`jit` compiled function,
+expressions like ``x = x.at[idx].set(y)`` are guaranteed to be applied inplace.
 
 
 Indexed update functions (deprecated)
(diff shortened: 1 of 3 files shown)


jakevdp (approved):
Good catch!

lfaucheux:
> [...] guaranteed to be performed in-place under JIT [...]

Sorry folks, but when doing 

```
@jit
def f(
    a, idx, b
):
    a.at[idx].set(b)
```
and

```
>>> a = jnp.ones((3, 3))
>>> idx = ((1, 1, 2, 2), (1, 2, 1, 2),)
>>> f(a, idx, -1) or a
DeviceArray([[1., 1., 1.],
             [1., 1., 1.],
             [1., 1., 1.]], dtype=float64)  # no `-1`s
```

I must admit that I do not get what is meant by  *guaranteed to be performed in-place under JIT*. After reading this, I had something like the following in mind:

```
>>> oa = onp.ones((3, 3))
>>> oa.__setitem__(idx, -1) or oa
array([[ 1.,  1.,  1.],
       [ 1., -1., -1.],
       [ 1., -1., -1.]])
```

Is such thing doable, i.e. "in-place fancy-update", in a near future ?

***[jax==0.3.13][jaxlib==0.3.7]***<sub>JAX is amazing</sub>

jakevdp:
No, jax arrays are immutable by design, so it is not possible to update/mutate an existing jax array in-place. What the documentation refers to is the face that when these index update operations are used within a sequence of operations and compiled by jit, the compiler will ensure that in execution the updates happen in-place rather than generating copies of the input within the compiled function.

PR #7790, opened 2021-09-02
[sparse] add BCOO._unbatch() utility

jakevdp:
Builds on #7779

This will be important for MLIR-backed computation, because batched sparse formats are not supported.

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -597,6 +597,19 @@ def _dedupe_bcoo_one(data, indices):
   return data_unique, indices_unique
 
 
+def _unbatch_bcoo(data, indices, shape):
+  n_batch, _, _ = _validate_bcoo(data, indices, shape)
+  if n_batch == 0:
+    return data, indices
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/ops.py
```diff
@@ -1429,6 +1442,10 @@ def from_scipy_sparse(cls, mat, *, index_dtype=None, n_dense=0, n_batch=0):
     indices = jnp.column_stack((mat.row, mat.col)).astype(index_dtype)
     return cls((data, indices), shape=mat.shape)
 
+  def _unbatch(self):
+    """Return an unbatched representation of the BCOO matrix."""
+    return self.__class__(_unbatch_bcoo(self.data, self.indices, self.shape), shape=self.shape)
```
froystig, line 1447: Any reason not to construct `BCOO(...)` explicitly? I don't think we're expecting subclasses, and if anyone were to create one then this makes an unchecked assumption about its constructor.

froystig (approved):

jax/experimental/sparse/ops.py
jakevdp: Good call, Done.

PR #7792, opened 2021-09-02
DOC: add documentation of jax.experimental.sparse

jakevdp:


froystig (approved):

jakevdp:
Need to wait until #7779 is merged, as the `indices` output will have to be updated.

PR #7811, opened 2021-09-03
WIP: try making Python strings pytrees

shoyer:
This is just to see if anything breaks!

If we do this for real, it should be done on the C++ side, so the repr
looks better. E.g., this is what it currently looks like:

    >>> tree_util.tree_flatten('foo')
    ([], PyTreeDef(CustomNode(<class 'str'>[foo], [])))


hawkinsp:
Closing as stale.

PR #7812, opened 2021-09-03
handle linear custom_jvp functions

mattjj:
I added a TODO expressing that we could offer more control over how custom_jvp functions interact with custom checkpointing policies, in the policy function (making more info about the custom_jvp function readily visible) and/or when the custom_jvp decorator is applied (some kind of annotation).

Another observation we made, relevant to follow-up work but not really requiring any particular follow-up work itself, is that this change affects more than just linear custom_jvp functions because of how jax.checkpoint relies on partial evaluation tracing (i.e. on JaxprTrace). That is, jax.checkpoint may be the only place in the system where partial evaluation 'unknowns' _don't_ correspond to linear inputs.

cc @jekbradbury

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -381,17 +381,25 @@ ad.reducing_transposes[custom_jvp_call_jaxpr_p] = _custom_jvp_call_jaxpr_transpo
 def custom_jvp_jaxpr_custom_partial_eval_rule(
     saveable: Callable[..., bool], unks_in: List[bool], inst_in: List[bool],
     eqn: core.JaxprEqn
-  ) -> Tuple[core.JaxprEqn, core.JaxprEqn, List[bool], List[bool], List[core.Var]]:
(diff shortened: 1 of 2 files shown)


jax/_src/custom_derivatives.py
```diff
@@ -386,12 +386,15 @@ def custom_jvp_jaxpr_custom_partial_eval_rule(
   # constituent parts, so we always perform full remat. An alternative would be
```
levskaya, line 386: Return type `-> Tuple[core.JaxprEqn, core.JaxprEqn, List[bool], List[bool], List[core.Var]]` needs to be `-> Tuple[Optional[core.JaxprEqn], core.JaxprEqn, List[bool], List[bool], List[core.Var]]` with this change to avoid the type errors we're seeing here.

mattjj: You're the best, Anselm! Thanks.

PR #7824, opened 2021-09-06
Fix argument names for jnp.ndarray.clip

### LINKED ISSUE #7823: Keyword arguments to jnp.ndarray.clip() do not follow the documentation
The following should work according to the docs (https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.html), and it does work when using `numpy` instead of `jax.numpy`:

```python
import jax.numpy as jnp
jnp.zeros(3).clip(min=1)
```

However, in contrast to the docs, `jnp.ndarray.clip` actually expects the arguments to be called `a_min` and `a_max` respectively, just like `jnp.clip` does – i.e. `jnp.zeros(3).clip(a_min=1)` works.

khdlr:
Fixes #7823 by implementing `jnp.ndarray.clip` instead of directly delegating to `jnp.clip` (the implementation calls `jnp.clip` with renamed keywords).

Not sure where exactly the wrapping function `_clip` should go within the file `jax/_src/numpy/lax_numpy.py`.

Also not sure whether something needs to be done to not break the autodocs for `jnp.ndarray.clip`, since I could not get the docs to build locally.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5946,6 +5946,20 @@ def _nbytes(arr):
   return size(arr) * _dtype(arr).itemsize
 
 
+def _clip(number, min=None, max=None, out=None, *, a_min=None, a_max=None):
+  # ndarray.clip has a slightly different API from clip (min -> a_min, max -> a_max)
+  # TODO: remove after deprecation window
+  if a_min is not None or a_max is not None:
+    warnings.warn('`a_min` and `a_max` keyword arguments to ndarray.clip are deprecated '
+                  'in favor of `min` and `max` for compatibility with numpy. '
+                  'They will be removed in JAX 0.22.2', FutureWarning)
+  if min is None and a_min is not None:
+    min = a_min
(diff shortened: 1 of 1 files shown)


jakevdp:
Looks great, thanks!

jakevdp (approved):

jakevdp:
This may take a bit of time to merge, because I found a few downstream uses of the `a_min` and `a_max` keywords in the `clip` method.

khdlr:
I thought that might happen. But it's not a critical issue, and I guess that people who are confused by the current behaviour will come across #7824 which explains how to work around it for now :upside_down_face:

jakevdp:
Maybe the best approach would be a deprecation cycle: accept `a_min` and `a_max` as keyword-only arguments in the new version of the function, and if they are used raise a warning that they're going away in JAX version 0.2.22 (the release after next)

khdlr:
Sounds good, pushed the corresponding changes just now :+1:

jakevdp (approved):

PR #7847, opened 2021-09-07
add random key array type to public module

froystig:
This is meant to allow for downstream type annotations and checking (e.g. via `isinstance`).

### DIFF
--- a/jax/random.py
+++ b/jax/random.py
@@ -78,6 +78,10 @@ for the design and its motivation.
 
 # flake8: noqa: F401
 
+# TODO(frostig): replace with KeyArray from jax._src.random once we
+# always enable_custom_prng
+from jax._src.prng import PRNGKeyArray as KeyArray
+
 from jax._src.random import (
   PRNGKey,
   bernoulli,


LenaMartens (approved):

tomhennigan:
Is there a way we can make the name of type `PRNGKey` rather than `PRNGKeyArray` (purely for cosmetic reasons)?

For example we could implement `__new__` to support the old behaviour of creating an ndarray or the new type?

```python
class PRNGKey:
  def __new__(self, ..):
    if legacy_behaviour:
      # construct and return an ndarray
    else:
      # construct and return a PRNGKey
```

froystig:
> Is there a way we can make the name of type `PRNGKey` rather than `PRNGKeyArray` (purely for cosmetic reasons)?

I had in mind:

```python
from jax import vmap, random, numpy as jnp
seeds = jnp.arange(4)
keys = vmap(random.PRNGKey)(seeds)
```

Here `keys.shape` is `(4,)` i.e. it serves as an array of keys. I found it confusing to say that `keys` is a PRNGKey and not an array of keys. But I can see the back-compat rationale...

> For example we could implement `__new__` to support the old behaviour of creating an ndarray or the new type?
> 
> ```python
> class PRNGKey:
>   def __new__(self, ..):
>     if legacy_behaviour:
>       # construct and return an ndarray
>     else:
>       # construct and return a PRNGKey
> ```

Clever!

This all would have been easier if we had not, for so long, exposed the function `random.PRNGKey`, which is capitalized to suggest a type. I think it would be better called `random.seed` simply. A question is whether to "correct" for that now, continue to roll with it entirely, or do something in between.

tomhennigan:
Sorry for the delay getting back to you! I'd suggest we roll forward with the PR as is.

Definitely agree it is unfortunate that PRNGKey is named as it is (given that it is not a type). Renaming that to `seed` as you suggest and keeping `PRNGKeyArray` sounds like a nice improvement that we could think about doing later on.

PR #7869, opened 2021-09-09
Add `jnp.get_printoptions` and `jnp.printoptions`.

dan-zheng:
`jnp.set_printoptions` already exists (579665b1b0ee8caa0f154d6677cd4b13be6adc3f) but other printing utilities do not.

`jnp.get_printoptions` forwards [`numpy.get_printoptions`](https://numpy.org/doc/stable/reference/generated/numpy.get_printoptions.html).
`jnp.printoptions` forwards [`numpy.printoptions`](https://numpy.org/doc/stable/reference/generated/numpy.printoptions.html).

---

Context: I wanted to use the `jnp.printoptions` context manager in my codebase but found that it does not exist.

Perhaps I could use `np.printoptions` instead – but since `jnp.set_printoptions` exists, it was not obviously clear to me that `np.printoptions` would work out of the box for JAX arrays.

Supporting (forwarding) all three NumPy printing utilities seems reasonable.


dan-zheng:
Alternatively: perhaps it is a goal to remove forwarding APIs from JAX?

If so, and if the NumPy printing utilities (e.g. `np.printoptions`) work for printing JAX arrays, an option is to remove all of the forwarded printing utilities from `jax.numpy` to reduce API surface.

If this is done, a use of `jnp.set_printoptions` in https://github.com/google/jax/blob/main/cloud_tpu_colabs/JAX_NeurIPS_2020_demo.ipynb needs to be updated to `np.set_printoptions`.

jakevdp (changes requested):
Looks reasonable, thanks! I think you'll need to rebase on the main branch and update the imports in `jax/numpy/__init__.py`, and then we can get this merged.

Thanks!

dan-zheng:
> Looks reasonable, thanks! I think you'll need to rebase on the main branch and update the imports in `jax/numpy/__init__.py`, and then we can get this merged.

Thanks, done!

jakevdp (approved):
Thanks!

PR #7901, opened 2021-09-13
Test pre-commit config for large files

yashk2810:


jakevdp:
Looks like we have a couple existing files that would break this were we to ever send a PR that modifies them:
```
bash-3.2$ git ls-tree -r -l --abbrev --full-name HEAD | sort -n -r -k 4 | head -n 5
100644 blob f866c53f 5913790	cloud_tpu_colabs/images/wave_movie.gif
100644 blob 42dc2b4e 2546436	docs/notebooks/score_matching.ipynb
100644 blob 66059c0b  574004	cloud_tpu_colabs/images/lorentz.png
100644 blob 5f5cc605  526374	docs/notebooks/convolutions.ipynb
100644 blob 1ac0f3de  432019	docs/notebooks/xmap_tutorial.ipynb
```

yashk2810:
I am testing if I modify one of those files, will the pre-commit tests.

yashk2810:
Looks like pre-commit for large files is passing right now. Interesting!

hawkinsp:
Closing as stale.

PR #7906, opened 2021-09-13
Add precision param for pdot

sharadmv:
Currently `pdot` has no precision parameter. When it is lowered we lose information about the `default_matmul_precision` decorator and the precision is always default. This PR adds in the `precision` parameter and plumbs it to the lowered `dot_general`.


mattjj (commented):
Can you add a test?

sharadmv:
> Can you add a test?

Added a test that looks at the jaxpr and the vmapped jaxpr.

mattjj:
Instead of checking the jaxpr, can we just check the behavior (of whether the precision is preserved)?

sharadmv:
Updated to check jaxpr text instead of looking through the data structure.

mattjj (approved):
Thanks!

PR #7907, opened 2021-09-13
jnp.take: require array argument (but not indices) to be arraylike

jakevdp:
Part of #7737

I'm leaving out the `indices` argument for now, because it's quite common downstream for users to pass lists of indices to `jnp.take`. I will likely revisit the indices later.


soraros:
How about add a warning for `indices`, so when/if you revisit this, the situation downstream will be better.

mattjj (approved):

PR #7908, opened 2021-09-14
Switch internal users of jax.util.partial to use functools.partial.

hawkinsp:


soraros:
There is still `jax/lax/__init__.py` and `jax/interpreters/partial_eval.py`.
`

hawkinsp:
> There is still `jax/lax/__init__.py` and `jax/interpreters/partial_eval.py`.

Thanks! I left a TODO for the `lax/__init__.py` one for now.

Issue #7910, opened 2021-09-14
Should the predict function in README.md return inputs rather than outputs?


juliangall:
The example in the readme file has:

```python
def predict(params, inputs):
  for W, b in params:
    outputs = jnp.dot(inputs, W) + b
    inputs = jnp.tanh(outputs)
  return outputs
```

The return value should surely be the **inputs** variable (not **outputs**), as this is the output of the final tanh activation in the loop.

zhangqiaorjc:
Thanks for finding a typo. Would you like to contribute a PR to fix this?

jakevdp:
Actually, I believe this is correct as written. The variable names are a bit confusing, but the intent is to return the result of the final dot product, not the result of the final `tanh` activation function.

zhangqiaorjc:
@jakevdp  Oops..I read the snippet in isolation. Could we maybe add a comment at least? This is confusing on a cursory look...

jakevdp:
We've discussed this previously and ended up renaming variables, and talked about adding a comment but never got consensus on what to add (a sampling of similar issues with relevant discussions: #31, #119, #4048, #4560, #5230, #5236, #7888)

juliangall:
Thanks everyone. A comment will be helpful. My preference would be to have the code do what is necessary to express its intent, and not to "throw away" the final ```tanh``` activation. However, I am sure others will feel differently.

jakevdp:
Thanks! I'm going to close this, since I think the issue should be addressed by #7915

PR #7915, opened 2021-09-14
DOC: add brief comments to example in README

jakevdp:
Why? This example has repeatedly cause confusion (see e.g. #31, #119, #2421, #2422, #3900, #4048, #4560, #5230, #5236, #7888, #7910)


mattjj (approved):
Best code I've ever seen.

Issue #7918, opened 2021-09-14
Boolean indexing inconsistent with numpy in the case of a scalar index 


ajwittmond:
Scalar boolean applied to jax arrays behave differently than when applied to numpy arrays, as the following example demonstrates:
```
>>> import numpy as np
>>> import jax.numpy as jnp
>>> nx = np.array([[1,2],[3,4]])
>>> jx = jnp.array([[1,2],[3,4]])
>>> nx[True]
array([[[1, 2],
        [3, 4]]])
>>> jx[True]
DeviceArray([[1, 2]], dtype=int32)
>>> nx[True] = 1
>>> nx
array([[1, 1],
       [1, 1]])
>>> jx.at[True].set(1)
DeviceArray([[1, 1],
             [3, 4]], dtype=int32)
```
In particular, this is surprising since one might expect the last expression to have the same output as `jnp.where(True, jx, 1)` as `nx[True] = 1` results in `nx` being identical to `np.where(True, nx, 1)`.

jakevdp:
Thanks for the report! I vaguely remember us looking into this before, but I can't find any discussion of it. Scalar boolean indices are kind of weird though....
```python
>>> np.arange(4)[True].shape                                                                                                
(1, 4)

>>> np.arange(4)[False].shape                                                                                               
(0, 4)

>>> np.arange(4)[:, True].shape                                                                                             
(4, 1)

>>> np.arange(4)[:, False].shape                                                                                            
(4, 0)

>>> np.arange(4)[None, :, True].shape                                                                                       
(1, 4, 1)
```
OK, so they act kind of like `np.newaxis`, with size 0 an 1. But wait:
```python
>>> np.arange(4)[None, :, None].shape                                                                                       
(1, 4, 1)

>>> np.arange(4)[True, :, True].shape                                                                                       
(1, 4)
```
it only seems to care about the first boolean scalar passed, and throws out the rest. But...
```python
>>> np.arange(4)[True, :, False].shape                                                                                      
(0, 4)
```
No, it doesn't throw out the last one, it does an `and` reduction across all scalar boolean indices, and puts them in the place of the first. But what if there are boolean masks also?
```
>>> x = jnp.arange(4)

>>> x[True, x < 3]
array([0, 1, 2])
```
huh, it ignores the boolean scalar. What if there's a False?
```python
>>> x[False, x < 3]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
[...]

PR #7925, opened 2021-09-15
jnp.array: raise TypeError for scalar boolean indexing

jakevdp:
Addresses #7918

Rather than implement numpy's behavior here, we will now raise an explicit `TypeError`. The reason is that numpy's behavior for scalar boolean indices is poorly-documented, not well-defined, and difficult to understand.

See, e.g. https://github.com/google/jax/issues/7918#issuecomment-920260618 and https://stackoverflow.com/a/65598320


mattjj (approved):
Excellent analysis! I like your thinking.

PR #7927, opened 2021-09-15
Turn jnp.ndarray into a true abstract base class.

hawkinsp:
Make all JAX array types instances of jnp.ndarray.
Remove np.ndarray from jnp.ndarray.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -24,9 +24,9 @@ transformations for NumPy primitives can be derived from the transformation
 rules for the underlying :code:`lax` primitives.
 """
 
+import abc
 import builtins
 import collections
-import collections.abc
 from functools import partial
 import operator
 import types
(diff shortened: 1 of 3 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -95,24 +95,26 @@
...
 
   def __instancecheck__(self, instance):
+    # Allow tracer instances that avals that are instances of UnshapedArray.
+    # We could instead just declare Tracer an instance of the ndarray type, but
+    # there can be traced values that are not arrays. The main downside here is
+    # that isinstance(x, ndarray) might return true but
+    # issubclass(type(x), ndarray) might return false for an array tracer.
     try:
-      return isinstance(instance.aval, _arraylike_types)
+      return isinstance(instance.aval, UnshapedArray)
```
jakevdp, line 110: We've run into a few situations where `DeviceArray` objects are created with `aval = None` (there were some `repr`-related bugs in the last week or two). We might still want to handle the `isinstance(DeviceArray)` directly, or perhaps assert that `aval is not None` to avoid difficult-to-debug errors.

jax/_src/numpy/lax_numpy.py
```diff
@@ -95,24 +95,26 @@
...
-class _ArrayMeta(type(np.ndarray)):  # type: ignore
+class ArrayMeta(abc.ABCMeta):
   """Metaclass for overriding ndarray isinstance checks."""
 
   def __instancecheck__(self, instance):
+    # Allow tracer instances that avals that are instances of UnshapedArray.
+    # We could instead just declare Tracer an instance of the ndarray type, but
+    # there can be traced values that are not arrays. The main downside here is
+    # that isinstance(x, ndarray) might return true but
+    # issubclass(type(x), ndarray) might return false for an array tracer.
```
jakevdp, line 108: Should we also overload `__subclasscheck__` to address this concern?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3106,7 +3359,7 @@ def _concatenate_array(arr, axis: int):
 
 @_wraps(np.concatenate)
 def concatenate(arrays, axis: int = 0):
-  if isinstance(arrays, ndarray):
+  if isinstance(arrays, (np.ndarray, ndarray)):
```
jakevdp, line 3362: Here and elsewhere where `isinstance` is used, it might be more correct to use `if _arraylike(arrays)`, though perhaps that should be a separate PR because it does slightly change the behavior.

jax/_src/numpy/lax_numpy.py
```diff
@@ -6285,11 +6539,12 @@ def wrapped(*args, **kwargs):
 # functions, which can themselves handle instances from any of these classes.
 
 _scalar_types = (int, float, complex, np.generic)
+_accepted_binop_types = (int, float, complex, np.generic, np.ndarray, ndarray)
 
 def _defer_to_unrecognized_arg(binary_op):
   # Ensure that other array types have the chance to override arithmetic.
   def deferring_binary_op(self, other):
-    if not isinstance(other, _scalar_types + _arraylike_types + (core.Tracer,)):
+    if not isinstance(other, _accepted_binop_types):
```
jakevdp, line 6547: This is a change of behavior, no? Because some tracers may not have an arraylike aval?

jakevdp:
I think this is a great change, by the way 😁

jax/_src/numpy/lax_numpy.py
hawkinsp: Yes, this would now reject Tracers that are not arrays. But does anyone pass such a thing to one of these operators — wouldn't these all only work for arrays anyway?

In practice, there are no test failures in our internal tests from this cause.

jax/_src/numpy/lax_numpy.py
hawkinsp: I don't think it can be done.  `__subclasscheck__` receives a type, not a value. Whether a tracer is an array or not is a property of a value, not a type. We could rectify this by defining a new `ArrayTracer` subclass of `Tracer`, perhaps.

jax/_src/numpy/lax_numpy.py
hawkinsp: I added a `hasattr(instance, "aval")` guard for now.

jax/_src/numpy/lax_numpy.py
hawkinsp: Yes, I think I'd rather keep this change mechanical and think about that in a future change.

jakevdp (approved):

PR #7943, opened 2021-09-17
Check presence of __jax_array__ in _arraylike before calling it in is…

rsepassi:
…scalar

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -301,7 +301,7 @@ def _result_dtype(op, *args):
 
 
 def _arraylike(x):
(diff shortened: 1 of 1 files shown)


jakevdp:
Hi - can you say more about the motivation for this change?

rsepassi:
Sure. This ensures that the narrower property of whether the value has the `__jax_array__` property is checked before actually calling it within `isscalar`. The motivation is that I'm working on a small `LazyArray` library that attempts to delay array materialization as late as possible when interacting with JAX functions, and the current order of calls here forces materialization earlier than is optimal.

jakevdp:
Makes sense, thanks!

jakevdp (approved):

rsepassi:
Thanks!

Issue #7912, opened 2021-09-14
Build from source instruction for window in doc has error in using pacman


nonconvexopt:
As you can see in [additional-notes-for-building-jaxlib-from-source-on-windows](https://jax.readthedocs.io/en/latest/developer.html#additional-notes-for-building-jaxlib-from-source-on-windows) section in doc, 

`pacman -S patch realpath`

should be changed to 

`pacman -S patch coreutils`

Reference: [https://command-not-found.com/realpath](https://command-not-found.com/realpath)

zhangqiaorjc:
Would you like to contribute a PR to fix the doc?

nonconvexopt:
Sure, it is my pleasure.

I also found another error that in Windows:

In the example command to build JAX from source
`python .\build\build.py `
  --enable_cuda `
  --cuda_path='C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.1' `
  --cudnn_path='C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.1' `
  --cuda_compute_capabilities='x.x' `
  --cuda_version='x.x' `
  --cudnn_version='x.x.x'`
'--cuda_compute_capabilities' raises an error and we should get rid of it.

hawkinsp:
Thanks for the PR!

PR #7947, opened 2021-09-17
Wrong instruction in building from source for Windows #7912

nonconvexopt:
Modified trivial errors in Google JAX docs about building from source instruction in Windows.
[#7912](https://github.com/google/jax/issues/7912)

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -94,7 +94,7 @@ for more details. Install the following packages:
 ```
 pacman -S patch coreutils
 ```
-`realpath` command should be found in shell command line.
+Once coreutils is installed, the realpath command should be present in your shell's path.
 
 Once everything is installed. Open PowerShell, and make sure MSYS2 is in the
 path of the current session. Ensure `bazel`, `patch` and `realpath` are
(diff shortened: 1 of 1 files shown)


docs/developer.md
```diff
@@ -109,6 +110,7 @@ python .\build\build.py `
   --cuda_version='10.1' `
   --cudnn_version='7.6.5'
 ```
+Discard `--cuda_compute_capabilities` flag in case of argument exception.
```
hawkinsp, line 113: I'd remove this comment. But instead I'd just remove `--cuda_compute_capabilities` from the command above. The defaults should work for most people, and even if they do not, it isn't a Windows-specific thing.

(It turns out we also accidentally removed the flag, but we'll add it back again.)

docs/developer.md
```diff
@@ -92,8 +92,9 @@ is needed. See [Installing Bazel on Windows](https://docs.bazel.build/versions/m
 for more details. Install the following packages:
 
 ```
-pacman -S patch realpath
+pacman -S patch coreutils
 ```
+`realpath` command should be found in shell command line.
```
hawkinsp, line 97: Nit: perhaps:
"Once `coreutils` is installed, the `realpath` command should be present in your shell's path."

Or maybe just omit this? If we are installing the right package above, won't all the right things happen already?

hawkinsp (changes requested):

docs/developer.md
nonconvexopt: That would be more straightforward. Thank you for your comment.

docs/developer.md
nonconvexopt: I revised it according to your comment.

hawkinsp (approved):

Issue #7884, opened 2021-09-11
`jax2tf` is failing with `__call__() missing 1 required keyword-only argument: 'train'`


sayakpaul:
Hi.

I am trying to convert a Vision Transformer model to TensorFlow and serialize it as a SavedModel following the guidelines from here: https://github.com/google/jax/blob/main/jax/experimental/jax2tf/examples/README.md.

I am able to generate all the artifacts needed to run the conversion but it seems like the `tf_graph` (out of which one would extract the concrete function) is not getting constructed properly. 

Short code snippet:

```python
predict_fn = lambda params, inputs: model.apply(dict(params=params), inputs)

tf_fn = jax2tf.convert(
    predict_fn,
    with_gradient=False,
    enable_xla=True)

param_vars = tf.nest.map_structure(
    lambda param: tf.Variable(param, trainable=False),
    params)
tf_graph = tf.function(lambda inputs: tf_fn(param_vars, inputs),
                        autograph=False,
                        jit_compile=True)
```

While running `signatures[tf.saved_model.DEFAULT_SERVING_SIGNATURE_DEF_KEY] = \
    tf_graph.get_concrete_function(input_signatures[0])` I get into:

```
/usr/local/lib/python3.7/dist-packages/flax/linen/module.py in wrapped_module_method(*args, **kwargs)
    269     _context.module_stack.append(self)
    270     try:
--> 271       y = fun(self, *args, **kwargs)
    272       if _context.capture_stack:
    273         filter_fn = _context.capture_stack[-1]

TypeError: __call__() missing 1 required keyword-only argument: 'train'
```

Here's the Colab Notebook for reproducing [this error](https://colab.research.google.com/gist/sayakpaul/def1050fc334f6cc8b139000586e54f3/scratchpad.ipynb).

sayakpaul:
@gnecula if we update the TensorFlow version to `tf-nightly` (`2.7.0-dev20210911`) then the error goes away. 

Another question I wanted to ask. Is there any workaround to enable fine-tuning of the converted models? [This](https://github.com/google/jax/blob/main/jax/experimental/jax2tf/examples/saved_model_lib.py#L108-L113) tells me it is NOT supported but I still wanted to know.

gnecula:
I am glad you found that the `tf-nightly` fix. We only test with tf-nightly, but for reasons having to do with some details of how certain TF ops are working. I did not realize that the handling of `kwargs` may have changed in TF. 

[...]

PR #7978, opened 2021-09-22
[jax2tf] Minor updates to the jax2tf examples documentation

gnecula:


gnecula:
@sayakpaul please take a look. 

Re: #7884

jax/experimental/jax2tf/examples/README.md
```diff
@@ -11,6 +11,10 @@ This directory contains a number of examples of using the
     in a larger TensorFlow Keras model.
   * use Flax models with TensorFlow Serving, TensorFlow JavaScript, and TensorFlow Lite.
 
+You can also find usage examples in other projects:
+
+  * A [Vision Transformer model converted to TensorFlow](https://git.io/Jz8pj) and integrated with TensorFlow Hub.
```
sayakpaul, line 16: "[Vision Transformer models converted to TensorFlow](https://git.io/Jz8pj) and integrated with TensorFlow Hub"

might be better. You might also add this: https://git.io/JzR68. This shows the same thing but for another architecture called MLP-Mixer. 

This way the community will have ready access to examples where `jax2tf` is being used for different architectures.

gnecula: Why not put both examples in the same repo?

sayakpaul: Both are different architectures hence I decided to prepare separate repositories to keep them clean and segregated.

gnecula: I think then that we can update the wording as you suggest, but keep a reference only to the ViT repo (there is small additional benefit to the documentation of jax2tf to have another link for a very similar conversion).

PR #7982, opened 2021-09-22
Add atol and rtol arguments to jtu._CheckAgainstNumpy().

hawkinsp:
Prefer atol and rtol if they are provided.


tlu7:
Thanks for the help!

PR #2982, opened 2020-05-06
Use a whitelist to restrict visibility in top-level jax namespace.

hawkinsp:

PR #7984, opened 2021-09-22
Import numpy explicitly in the jax namespace

jakevdp:
Why? We already do import it by default, so
```python
import jax
print(jax.numpy)
```
will work (because `jax.numpy` falls back to a `sys.modules` lookup), and I suspect we have users that depend on this.

Currently, though, static analyzers will flag this as an error, because `numpy` is not part of the declared `jax` namespace.

Since we already import it for initialization purposes, there is no cost to making `jax.numpy` explicitly part of the `jax` namespace at import.

Note also that the import in line 115... https://github.com/google/jax/blob/484b5af5c915fb6356f2cd1de8c9c826eb7d1938/jax/__init__.py#L115 ...already results in `jax.numpy` being imported: https://github.com/google/jax/blob/484b5af5c915fb6356f2cd1de8c9c826eb7d1938/jax/experimental/maps.py#L25 ...so moving the explicit import to its alphabetical order in this PR is essentially a no-op.

### DIFF
--- a/jax/__init__.py
+++ b/jax/__init__.py
@@ -124,6 +124,7 @@ from . import errors as errors
 from . import image as image
 from . import lax as lax
 from . import nn as nn
+from . import numpy as numpy
 from . import ops as ops
 from . import profiler as profiler
 from . import random as random
@@ -131,9 +132,3 @@ from . import tree_util as tree_util
 from . import util as util
 
 import jax.lib  # TODO(phawkins): remove this export.
-
-def _init():
-  from . import numpy as numpy # side-effecting import sets up operator overloads
-
-_init()
-del _init


mattjj:
I'm not sure I have all the context to evaluate any changes, though the explanation sounds reasonable!

Seems like #2982 set up this `_init` thing, but for flag-setting purposes. Then #3010 put the `jax.numpy` import under there.

AIUI this change sort of undoes #3010, though back then it was the `jax.np` name getting exported and now it's the `jax.numpy` one.

Can we get @hawkinsp to check this?

hawkinsp:
I think in the past we were trying not to implicitly import submodules of `jax`. But at some point that changed, and now we import many submodules by default. This PR seems consistent with that, and in practice I doubt you want to write a JAX program without importing `jax.numpy`.

hawkinsp (approved):

PR #7987, opened 2021-09-23
add jax.ensure_compile_time_eval to public api

### LINKED ISSUE #7535: Expose jax.core.eval_context() as part of the public API

mattjj:
alias for `jax.core.eval_context`

fixes #7535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
   * Added a new debugging flag/environment variable `JAX_DUMP_IR_TO=/path`.
     If set, JAX dumps the MHLO/HLO IR it generates for each computation to a
     file under the given path.
+  * Added `jax.ensure_compile_time_eval` to the public api ({jax-issue}`#7987`).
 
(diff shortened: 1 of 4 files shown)


froystig (commented):
Add docstring, add to docs index, and add a changelog item?

froystig (approved):

mattjj:
This got submitted but weirdly wasn't marked as such. Closing!

PR #7989, opened 2021-09-23
Expand the docstring for jax.checkpoint.

gnecula:
I found the old docstring to be a bit inaccurate, but mostly
not explanatory enough. I think that this feature deserves
a tutorial, in addition to an expanded docstring.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2721,20 +2721,12 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
   At that time, the value ``jnp.sin(2.0)`` is recomputed, along with the values
   ``jnp.cos(2.0)`` and ``jnp.cos(jnp.sin(2.0))``.
 
-  Note that in some cases the XLA compiler may do its optimization that affect
-  the meory usage, e.g., common-subexpression optimization, fusion, or
-  even rematerialization. For example, if the code uses only element-wise
-  operations, like in our example, XLA may fuse both the forward and backward
-  pass, or may rematerialize aggressively, resulting in low memory usage even
-  without ``jax.checkpoint``. In fact, in some such cases ``jax.checkpoint`` may
-  hinder the compiler and you may see larger memory usage. For complex examples,
-  the effect of ``jax.checkpoint`` is likely to be more significant than the
-  XLA optimizations.
-
-  The best way to use ``jax.checkpoint`` is to experiment with its placement
-  on sub-computations. For example, ``lambda x: f(jax.checkpoint(g)(x))`` is
-  likely to have lower memory usage under ``jax.grad`` than when not using
-  ``jax.checkpoint``, or than when using it on ``f`` or the whole function.
+  While ``jax.checkpoint`` controls what values are stored from the forward-pass
+  to be used on the backward pass, the total amount of memory required to
+  evaluate a function or its VJP depends on many additional internal details of
+  that function. Those details include which numerical primitives are used,
+  how they're composed, where jit and control flow primitives like scan
+  are used, and other factors.
 
   The :func:`jax.checkpoint` decorator can be applied recursively to express
   sophisticated autodiff rematerialization strategies. For example:
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2708,15 +2708,33 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
   ...   z = jnp.sin(y)
   ...   return z
   ...
-  >>> jax.grad(g)(2.0)
-  DeviceArray(-0.25563914, dtype=float32)
+  >>> jax.value_and_grad(g)(2.0)
+  (DeviceArray(0.78907233, dtype=float32, weak_type=True), DeviceArray(-0.2556391, dtype=float32))
 
   Here, the same value is produced whether or not the :func:`jax.checkpoint`
-  decorator is present. But when using :func:`jax.checkpoint`, the value
-  ``jnp.sin(2.0)`` is computed twice: once on the forward pass, and once on the
-  backward pass. The values ``jnp.cos(2.0)`` and ``jnp.cos(jnp.sin(2.0))`` are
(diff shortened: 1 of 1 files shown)


gnecula:
@jheek PTAL

jax/_src/api.py
```diff
@@ -2708,15 +2708,33 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
...
+  ``jnp.cos(2.0)`` and ``jnp.cos(jnp.sin(2.0))`` are computed on the forward
+  pass and are stored for use in the backward pass, because they are needed
+  on the backward pass and depend only on the primal inputs. When using
+  :func:`jax.checkpoint`, the forward pass will compute only the primal outputs
+  and only the primal inputs (``2.0``) will be stored for the backward pass.
+  At that time, the value ``jnp.sin(2.0)`` is recomputed, along with the values
+  ``jnp.cos(2.0)`` and ``jnp.cos(jnp.sin(2.0))``.
+
+  Note that in some cases the XLA compiler may do its optimization that affect
+  the meory usage, e.g., common-subexpression optimization, fusion, or
```
jheek, line 2725: meory -> memory

jax/_src/api.py
```diff
@@ -2708,15 +2708,33 @@ def checkpoint(fun: Callable, concrete: bool = False, prevent_cse: bool = True,
...
+  pass, or may rematerialize aggressively, resulting in low memory usage even
+  without ``jax.checkpoint``. In fact, in some such cases ``jax.checkpoint`` may
+  hinder the compiler and you may see larger memory usage. For complex examples,
+  the effect of ``jax.checkpoint`` is likely to be more significant than the
+  XLA optimizations.
+
+  The best way to use ``jax.checkpoint`` is to experiment with its placement
+  on sub-computations. For example, ``lambda x: f(jax.checkpoint(g)(x))`` is
+  likely to have lower memory usage under ``jax.grad`` than when not using
+  ``jax.checkpoint``, or than when using it on ``f`` or the whole function.
```
mattjj, line 2737: I don't think this last part is accurate; using `jax.checkpoint(lambda x: f(g(x)))` will have lower memory because it will prevent any intermediate residuals from `f` or `g` being stored, whereas `lambda x: f(jax.checkpoint(g)(x))` will only prevent intermediate residuals from `g` (and in both cases we have to store the input).

Here's an example:

```python
import jax
from jax.tree_util import tree_leaves
import jax.numpy as jnp

def f(x):
  return jnp.sin(jnp.sin(x))

def g(x):
  return jnp.sin(jnp.sin(x))

num_res = len(tree_leaves(jax.vjp(lambda x: f(jax.checkpoint(g)(x)), 2.0)[1]))
print(f'number of residuals: {num_res}')

num_res = len(tree_leaves(jax.vjp(jax.checkpoint(lambda x: f(g(x))), 2.0)[1]))
print(f'number of residuals: {num_res}')
```

```
number of residuals: 3
number of residuals: 1
```

I suggest just deleting this paragraph. Not only is the example given inaccurate in general, but also I don't think the best way to use `jax.checkpoint` is experimentation: its behavior is not too hard to predict for someone familiar with reverse-mode AD (the intended user of the function), as the updated docstring now articulates! (In short: save no intermediate values for the decorated function.) It could be that XLA optimizations for `jit`ted functions can be unpredictable, but that's not `jax.checkpoint`-specific.

WDYT?

mattjj (changes requested):
Thanks! Great catch!

It looks like the buggy thing in the docstring was the comment about cosines being evaluated in the example; indeed that only applies when the `concrete=True` parameter is passed! (I checked and indeed it is the case with`concrete=True`.) I suspect that sentence about cosines was meant to start with something like "When `concrete=True`, then ...". Anyway, no one should use `concrete=True`, so I don't think that's worth commenting on.

I do have one change requested below though...

jax/_src/api.py
mattjj: Let me know if there was a particular example `f` and `g` you had in mind! I might be missing something.

jheek: The problem with this example that it only focuses on the residuals that are stored during the forward pass but not the once that are stored during remat. So effectively the memory usage will dependent on max(res_before_remat, res_inside_remat). And placing remat correctly depends on balancing this out. I think this example is deceptive because it only talks about memory stored during the forward pass and users might easily think that a single remat around the whole function results in O(1) memory consumption.

mattjj: Indeed the example only focuses on what residuals are stored from the forward pass for the backward pass, but that's what remat directly controls. I agree it has downstream implications about the total memory required to evaluate a function's vjp, but that depends on many other factors as well, as your formula shows, including e.g. many internal details of the function being differentiated, details about nesting of remats and jits, etc.

In other words, the memory required to evaluate a function's vjp (or even just the function itself!) is always pretty complex, independent of whether remat is present. To minimize memory usage might require experimenting not just with remat but also with jit, scan unrolling, etc.

So I suggest:
1. focusing only on the direct effect of remat, on saved residuals, rather than talking about total memory usage to evaluate a function or its vjp (i.e. let's certainly not say that putting a remat around a function results in O(1) memory consumption, which would be imposible);
2. if we want to comment on how total memory consumption required to evaluate a vjp (or just a function) depends in a complex way on that function's internal implementation as well as transformations like jit, remat, control flow, etc, then let's make a general comment about all those things rather than singling out remat.

WDYT?

mattjj: Another rephrase of the same idea: how much memory is required to evaluate a function (whether that function is a vjp of another function or not) depends on compiler details, so it's XLA's turf and not JAX's. How much memory is stored in residuals when you form a VJP is all about AD, and that's JAX's turf not XLA's.

mattjj: Here's a concrete proposal for how a revised comment on this could look:

> While `jax.checkpoint` controls what values are stored from the forward-pass to be used on the backward pass, the total amount of memory required to evaluate a function or its VJP depends on internal details of that function. Those details include which numerical primitives are used, how they're composed, where `jit` and control flow primitives like `scan` are used, and other factors.

PR #7996, opened 2021-09-23
Remove `.[minimum-jaxlib]` from test-requirements.txt

skye:
This means that jax and its dependencies (e.g. jaxlib) must be
manually installed before running the tests. This is useful for
testing an existing jax install, e.g. a later version of jaxlib, GPU
jaxlib, etc.


jakevdp (approved):
LGTM - I think the test failure is related to kokoro needing jaxlib to be installed

Issue #7733, opened 2021-08-26
ahead-of-time lowering and compilation


froystig:
Following the experiment in #6034, plus other thoughts and motivation accrued since, we should support a notion of lowering and compilation "ahead of time," i.e. upon explicit request. Some things to note:

* Lowering means producing XLA HLO. Compiling means running it through the XLA compiler
* Compiling produces optimized HLO and renders a computation ready for execution.
* Lowering and compilation happen for a fixed type signature.
* A function compiled ahead of time and can be called with arguments of the fixed type signature only.
* A function compiled ahead of time will intentionally fail to transform, e.g. if called under `jvp` or `vmap`, since it will be compiled for a particular type signature, and transformations modify signatures.
* We will want this for every compiling transformation: `jit`, `pjit`, `pmap`, `xmap`.
* Although a function compiled ahead of time is bound to correspond to some XLA executable, it's out of scope to provide supported access to, or any serialization of, that executable.

froystig:
[f12af93](https://github.com/google/jax/commit/f12af93258410fb1165335a7266818987e2d4b6f) reworks some of the AOT API, and attempts to set expectations around the (in)consistency of outputs.

Joshuaalbert:
@froystig What is the timeline on this?

froystig:
The API and implementation are essentially in place, and already in use. I'm working on documentation.

froystig:
With documentation now in place (#12204) and plenty of usage already, it's time to declare this complete! We will return with any thoughts and work on cross-AOT when the time comes for that.

PR #7997, opened 2021-09-24
ahead-of-time lowering and compilation for jit

froystig:
This is an initial take on #7733 corresponding to `jax.jit`

Example usage:

```python
>>> import jax; from jax import numpy as jnp
>>> def f(x): return jnp.sqrt(x ** 2) + 1.
>>> f_jit = jax.jit(f)

>>> f_low = f_jit.lower(1.)
>>> print(f_low._xla_computation().as_hlo_text())
HloModule jit_f__1.8

ENTRY jit_f__1.8 {
  parameter.1 = f32[] parameter(0)
  multiply.3 = f32[] multiply(parameter.1, parameter.1)
  sqrt.4 = f32[] sqrt(multiply.3)
  constant.5 = f32[] constant(1)
  add.6 = f32[] add(sqrt.4, constant.5)
  ROOT tuple.7 = (f32[]) tuple(add.6)
}

>>> f_exe = f_low.compile()
>>> print(f_exe._xla_executable().hlo_modules()[0].to_string())
HloModule jit_f__1.8

%fused_computation (param_0.2: f32[]) -> f32[] {
  %param_0.2 = f32[] parameter(0)
  %abs.1 = f32[] abs(f32[] %param_0.2), metadata={op_type="sqrt" op_name="jit(f)/sqrt" source_file="<ipython-input-2-f5f7990cdb84>" source_line=2}
  %constant.0 = f32[] constant(1), metadata={op_type="add" op_name="jit(f)/add" source_file="<ipython-input-2-f5f7990cdb84>" source_line=2}
  ROOT %add.0 = f32[] add(f32[] %abs.1, f32[] %constant.0), metadata={op_type="add" op_name="jit(f)/add" source_file="<ipython-input-2-f5f7990cdb84>" source_line=2}
}

ENTRY %jit_f__1.8 (parameter.1: f32[]) -> (f32[]) {
  %parameter.1 = f32[] parameter(0)
  %fusion = f32[] fusion(f32[] %parameter.1), kind=kLoop, calls=%fused_computation, metadata={op_type="add" op_name="jit(f)/add" source_file="<ipython-input-2-f5f7990cdb84>" source_line=2}
  ROOT %tuple.7 = (f32[]) tuple(f32[] %fusion)
}

>>> f_exe(3.)
DeviceArray(4., dtype=float32, weak_type=True)

>>> f_exe([3.])
<ipython-input-7-048b0cc42154> in <module>
----> 1 f_exe([3.])

~/ws/jax/jax/_src/api.py in __call__(self, *args, **kwargs)
    540     if in_tree != self.in_tree:
    541       # TODO(frostig): provide more info about the source function
--> 542       raise TypeError(
    543           f'function compiled for {self.in_tree}, called with {in_tree}')
    544     out_flat = self.computation.call(*args_flat)

TypeError: function compiled for PyTreeDef(((*,), {})), called with PyTreeDef((([*],), {}))
```

Note the compiler optimization between lowering and compilation, from XLA.

We have yet to determine exactly how to expose (functions of) the lowered computation and compiled executable, so for now the example above reaches for the underlying computation and executable directly, via a hidden method (leading underscore).

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -52,8 +52,8 @@ from .api_util import (flatten_fun, apply_flat_fun, flatten_fun_nokwargs,
 from . import traceback_util
 from .traceback_util import api_boundary
 from ..tree_util import (tree_map, tree_flatten, tree_unflatten, tree_structure,
-                        tree_transpose, tree_leaves, tree_multimap,
-                        treedef_is_leaf, treedef_children, Partial)
+                         tree_transpose, tree_leaves, tree_multimap,
+                         treedef_is_leaf, treedef_children, Partial, PyTreeDef)
 from .util import (unzip2, curry, safe_map, safe_zip, prod, split_list,
                    extend_name_stack, wrap_name, cache, wraps, HashableFunction)
 from jax._src.lib import jax_jit
@@ -97,8 +97,8 @@ F = TypeVar("F", bound=Callable)
 T = TypeVar("T")
 U = TypeVar("U")
 
-map = safe_map
-zip = safe_zip
+map, unsafe_map = safe_map, map
+zip, unsafe_zip = safe_zip, zip
 
 FLAGS = flags.FLAGS
 
@@ -322,9 +322,8 @@ def _prepare_jit(fun, static_argnums, static_argnames, donate_argnums,
 
   for arg in args_flat:
     _check_arg(arg)
-  flat_fun, out_tree = flatten_fun(f, in_tree)
 
-  return flat_fun, out_tree, args_flat, donated_invars
+  return f, in_tree, args_flat, donated_invars
 
 
 def _python_jit(
@@ -349,14 +348,17 @@ def _python_jit(
   def f_jitted(*args, **kwargs):
     if config.jax_disable_jit:
       return fun(*args, **kwargs)
-    flat_fun, out_tree, args_flat, donated_invars = _prepare_jit(
+    closed_fun, in_tree, args_flat, donated_invars = _prepare_jit(
         fun, static_argnums, static_argnames, donate_argnums, args, kwargs)
+    flat_fun, out_tree = flatten_fun(closed_fun, in_tree)
     out_flat = xla.xla_call(
         flat_fun, *args_flat,
         device=device, backend=backend, name=flat_fun.__name__,
         donated_invars=donated_invars, inline=inline)
     return tree_unflatten(out_tree(), out_flat)
 
+  f_jitted.lower = _make_lower(fun, static_argnums, static_argnames, device,
+                               backend, donate_argnums, inline)
   return f_jitted
 
 
@@ -407,8 +409,9 @@ def _cpp_jit(
     # An alternative would be for cache_miss to accept from C++ the arguments
     # (dyn_args, donated_invars, args_flat, in_tree), since otherwise we have
     # work/code that is redundant between C++ and Python. We can try that later.
-    flat_fun, out_tree, args_flat, donated_invars = _prepare_jit(
+    closed_fun, in_tree, args_flat, donated_invars = _prepare_jit(
         fun, static_argnums, static_argnames, donate_argnums, args, kwargs)
+    flat_fun, out_tree = flatten_fun(closed_fun, in_tree)
     out_flat = xla.xla_call(
         flat_fun, *args_flat,
         device=device, backend=backend, name=flat_fun.__name__,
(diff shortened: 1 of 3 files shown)


apaszke (commented):
Should this have more caching? I would expect calling `jit(f).lower(*args).compile()` to cause cache hits when I `jit(f)(*args)` later.

froystig:
Yep, that's part of the plan. I might scope this PR to a cache-free version of things first, to get the refactoring and basic functionality in place.

jax/_src/api.py
```diff
@@ -321,27 +343,18 @@ def _python_jit(
...
-      donated_invars = donation_vector(donate_argnums, args, kwargs)
-    else:
-      donated_invars = (False,) * len(args_flat)
-    for arg in args_flat:
-      _check_arg(arg)
-    flat_fun, out_tree = flatten_fun(f, in_tree)
-    out = xla.xla_call(flat_fun, *args_flat, device=device, backend=backend,
-        name=flat_fun.__name__, donated_invars=donated_invars, inline=inline)
-    return tree_unflatten(out_tree(), out)
+    closed_fun, in_tree, args_flat, donated_invars = _prepare_jit(
```
apaszke, line 346: consider returning a dict you later splat as kwargs? that's what I do with `xmap` and `pjit` to avoid messing up the order of returns/binders

apaszke: I also called it `_infer_params`, which I think is a better description for this function than `_prepare_jit`, but I don't have a super strong preference

jax/_src/api.py
```diff
@@ -467,9 +467,100 @@ def get_device_info():
...
+    # executable directly via this method.
+    if self.computation.is_trivial():
+      raise ValueError('A trivial computation has no executable')
+    return self.computation.xla_executable
+
+  def __call__(self, *args, **kwargs):
+    _, in_tree, args_flat, _ = _prepare_jit(
+        self.source_fun, self.options.static_argnums,
+        self.options.static_argnames, self.options.donate_argnums,
+        args, kwargs)
```
apaszke, line 539: Do we really need to be passing `source_fun` and all those other things here? It seems to me that the only thing you want to do here is flatten the args. Can't you just `tree_flatten` and then compare `self.in_tree` with the tree you get by doing that?

jax/interpreters/xla.py
```diff
@@ -809,23 +813,29 @@ def from_xla_computation(
...
   @staticmethod
   def from_trivial_jaxpr(jaxpr, consts, device, out_avals, kept_var_idx):
     result_handlers = map(partial(aval_to_result_handler, device), out_avals)
-    return XlaCompiledComputation(partial(
+    return XlaCompiledComputation(None, partial(
         _execute_trivial, jaxpr, device, consts, out_avals,
         result_handlers, kept_var_idx))
 
-  def __call__(self, *args):
+  def call(self, *args):
```
apaszke, line 832: Why isn't `__call__` sufficient?

jax/_src/api.py
```diff
@@ -467,9 +467,100 @@ def get_device_info():
...
+    if in_tree != self.in_tree:
+      # TODO(frostig): provide more info about the source function
+      raise TypeError(
+          f'function compiled for {self.in_tree}, called with {in_tree}')
+    out_flat = self.computation.call(*args_flat)
+    return tree_unflatten(self.out_tree, out_flat)
+
+
+def _lower(fun, static_argnums, static_argnames, device, backend,
```
apaszke, line 548: Wouldn't it make sense to use `_lower` on the regular execution path too? That's what I do with `xmap` and `pjit` (though `jit` has the added difficult of having a C++ path, which I don't know too much about).

jax/_src/api.py
```diff
@@ -467,9 +467,100 @@ def get_device_info():
...
   f_jitted = wraps(fun)(cpp_jitted_f)
 
+  f_jitted.lower = partial(_lower,
+                           fun, static_argnums, static_argnames, device,
+                           backend, donate_argnums, inline)
+
   return f_jitted
 
 
+class JitOptions(NamedTuple):
```
apaszke, line 477: Perhaps instead of a `NamedTuple` we could use [`TypedDict`](https://www.python.org/dev/peps/pep-0589/)? That has all the same typing benefits, but you can also easily splat it into params.

apaszke (approved):
AFAIK this is only supposed to be a first step and to that end LGTM. I left some comments that you might want to address, but it's fine by me to do it later.

jax/_src/api.py
hawkinsp: I also wonder if you can/should just precompute the bits you need in `lower`. Why do this work at call time?

hawkinsp (changes requested):

PR #7999, opened 2021-09-24
PRNGKeys can have dtype float0 if they are a tangent.

LenaMartens:
Testing question: I've added a test for this in `api_test`, and added a `skipTest` conditioned on `config.jax_enable_custom_prng`, same as the tests in `LaxRandomWithCustomPRNGTest`. Are these tests currently being run? I couldn't find a test configuration which runs these tests with `config.jax_enable_custom_prng=True` so I think they are always skipped?

### DIFF
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -24,6 +24,7 @@ from jax import core
 from jax import numpy as jnp
 from jax import tree_util
 from jax.config import config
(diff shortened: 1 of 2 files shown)


LenaMartens:
Added the test to `random_test` instead because that one is tested with `jax_enable_custom_prng=True`.

froystig (approved):

PR #8003, opened 2021-09-24
Add @jit decorators to jax.numpy.linalg and jax.scipy.linalg.

hawkinsp:


jax/_src/lax/linalg.py
```diff
@@ -1254,8 +1254,8 @@ def svd_jvp_rule(primals, tangents, full_matrices, compute_uv):
   if not compute_uv:
     return (s,), (ds,)
 
-  s_diffs = jnp.square(s_dim) - jnp.square(_T(s_dim))
-  s_diffs_zeros = jnp.eye(s.shape[-1], dtype=A.dtype)  # jnp.ones((), dtype=A.dtype) * (s_diffs == 0.)  # is 1. where s_diffs is 0. and is 0. everywhere else
+  s_diffs = (s_dim + _T(s_dim)) * (s_dim - _T(s_dim))
+  s_diffs_zeros = jnp.eye(s.shape[-1], dtype=s.dtype)  # jnp.ones((), dtype=A.dtype) * (s_diffs == 0.)  # is 1. where s_diffs is 0. and is 0. everywhere else
```
jakevdp, line 1258: Unrelated to main change? Seems reasonable though.

jakevdp (approved):
LGTM!

jax/_src/lax/linalg.py
hawkinsp: Actually, no, it isn't!

The `jit` on `jax.numpy.linalg.svd` did bad things to the numerics of this code, because LLVM optimized `x**2 - y **2` to `fma(x, x, -y**2)` which has bad cancellation behavior. This caused one of the tests to fail. This avoids the problem and is strictly better anyway (at least as accurate and perhaps cheaper).

PR #8014, opened 2021-09-27
Schur decomposition lapack wrapper

### LINKED ISSUE #6478: scipy.linalg.schur decomposition
Looks like schur decomposition isn't implemented at the moment. Are there any plans to support this? Also, anything specific that might block me from implementing this?

SaturdayGenfo:
This is an implementation of a wrapper for LAPACK's gees routine (in order to close #6478). 

The schur primitive calls this wrapper on CPU. Tests for this primitive to make sure it matches scipy's schur decomposition and batching tests are added in `tests/linalg_tests.py`.


jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,120 @@ def tridiagonal_solve(dl, d, du, b):
...
+          sort_eig_vals=False,
+          select_callable=None):
+  return schur_p.bind(
+      x,
+      compute_schur_vectors=compute_schur_vectors,
+      sort_eig_vals=sort_eig_vals,
+      select_callable=select_callable)
+
+
+def schur_impl(operand, *, compute_schur_vectors, sort_eig_vals,
```
jakevdp, line 1460: Slight preference for all of these implementation details to be marked private with a leading underscore, e.g.
```
def _schur_impl(...)
```
Same below.

jakevdp:
Thanks for the contribution!

I believe that the reason tests are failing is because github CI tests are run against released jaxlib versions, and testing this code requires jaxlib to be built from this branch.

The way we work around this generally is to gate the tests on the value of `jax._src.lib._xla_extension_version`, which is defined within the tensorflow source and is [currently at `38`](https://github.com/tensorflow/tensorflow/blob/18cd0a566088ec6aa03ff931bdbc421f0fb1e394/tensorflow/compiler/xla/python/xla_client.py#L51-L53).

For this PR, you could specify the tests to be skipped if the version is `_xla_extension_version <= 38`, which will make the tests green for the PR. Then we can do some internal testing against tensorflow with `_xla_extension_version` bumped to `39` before merging. What do you think?

jax/_src/lax/linalg.py
SaturdayGenfo: Thank you for reviewing ! I have marked them private

SaturdayGenfo:
There was a reference to the gees wrapper that was just in the body of the `linalg.py` primitives file, I put it in the scope of the schur primitive so it doesn't cause any issues without calls to the primitive. I also added an `if _xla_extension_version > 38` condition that skips the test for older version.

tests/linalg_test.py
```diff
@@ -1450,12 +1450,13 @@ def test_tridiagonal_solve(self, dtype):
                             for dtype in float_types + complex_types))
   @jtu.skip_on_devices("gpu", "tpu")
   def testSchur(self, shape, dtype):
-      rng = jtu.rand_default(self.rng())
-      n = shape[-1]
-      args_maker = lambda: [rng(shape, dtype)]
+      if jax._src.lib._xla_extension_version > 38:
```
jakevdp, line 1453: Rather than blocking the test code on an `iff` statement, it would be better to use `self.skipTest(reason)` so that the test is reported as skipped.

jakevdp:
Also, once you address that final change, please squash the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Thanks!

hawkinsp:
> There was a reference to the gees wrapper that was just in the body of the `linalg.py` primitives file, I put it in the scope of the schur primitive so it doesn't cause any issues without calls to the primitive. I also added an `if _xla_extension_version > 38` condition that skips the test for older version.

Thinking about this, I think it's probably even better to use the condition `jax._src.lib.version >= (0, 1, 72)`.

You can think about `_xla_extension_version` as the version of the XLA/TF source trees. In this case, this code is in the JAX tree, so I think your best bet is to use a version number defined in that tree also. (Ideally we'd have a monotonic version number in the JAX tree as well, but the jaxlib version number will do fine I think.)

hawkinsp:
You have some flake8 lint failures as well.

jakevdp:
@hawkinsp - the reason I suggested checking `xla_extension_version` is because I was planning to patch the CL to run internal tests for the new code. If we gate on jaxlib release, we won't see any potential failures in CI until the next version of jaxlib is released.

hawkinsp:
Kokoro *is* doing a jaxlib build from source for at least two of the builds. So we should see the failures right here in Github.

jakevdp:
Thanks, that makes sense. Agreed that checking on `jax._src.lib.version` is the better option here.

SaturdayGenfo:
I think I really messed up my git repo, I am going to do another pull request from a fresh repository. I really apologize.

PR #8019, opened 2021-09-27
Replace Jaxpr pretty printer

hawkinsp:
Add a Wadler-Lindig pretty printer.

Change jaxpr printing to use it.


mattjj (approved):
The changes in printed results are amazing! I didn't review the code in detail, but it's Peter code so it's the best in the biz.

PR #8029, opened 2021-09-28
Change absl flag documentation to mention `re.search`

nicholasjng:
The absl help texts for test target discovery mention that targets
will be discovered by `re.match` use. However, in the subsequent
implementation, actually `re.search` is used. This commit changes the
help texts for the `test_targets` and `exclude_test_targets` flags
to correctly mention `re.search` as the discovery algorithm.


nicholasjng:
@googlebot I fixed it.

jakevdp (approved):
Looks great, thanks!

PR #8033, opened 2021-09-28
Schur lapack wrapper

### LINKED ISSUE #6478: scipy.linalg.schur decomposition

SaturdayGenfo:
This is an implementation of a wrapper for LAPACK's gees routine (in order to close #6478).

The schur primitive calls this wrapper on CPU. Tests for this primitive to make sure it matches scipy's schur decomposition and batching tests are added in tests/linalg_tests.py. 

The tests are only run if `jax._src.lib.version >= (0, 1, 72)`, they are skipped otherwise.

(This is a continuation #8014, where I really messed up my branch, I'm very sorry for the spam it may have sent out)

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -42,6 +42,7 @@ from jax._src.lib import rocsolver
 
 from jax._src.lib import xla_client
 from jax._src.lib import xla_bridge as xb
+from jax._src.lib import version as jaxlib_version
 
 xops = xla_client.ops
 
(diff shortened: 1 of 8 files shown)


jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+  return schur_p.bind(
+      x,
+      compute_schur_vectors=compute_schur_vectors,
+      sort_eig_vals=sort_eig_vals,
+      select_callable=select_callable)
+
+
+def _schur_impl(operand, *, compute_schur_vectors, sort_eig_vals,
+               select_callable):
+  return (xla.apply_primitive(
```
jakevdp, line 1462: outer parentheses not necessary

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+          select_callable=None):
+  return schur_p.bind(
+      x,
+      compute_schur_vectors=compute_schur_vectors,
+      sort_eig_vals=sort_eig_vals,
+      select_callable=select_callable)
+
+
+def _schur_impl(operand, *, compute_schur_vectors, sort_eig_vals,
+               select_callable):
```
jakevdp, line 1461: match indentation; here and below

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+
+
+def _schur_translation_rule(c, operand, *, compute_schur_vectors, sort_eig_vals):
+  raise NotImplementedError(
+      "Schur decomposition is only implemented on the CPU backend.")
+
+
+def _schur_abstract_eval(operand, *, compute_schur_vectors, sort_eig_vals,
+                        select_callable):
+  if isinstance(operand, ShapedArray):
```
jakevdp, line 1477: I don't think this check is necessary

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+    T = operand.update(shape=batch_dims + (n, n), dtype=dtype)
+    vs = operand.update(shape=batch_dims + (n, n), dtype=dtype)
+  else:
+    raise NotImplementedError
+
+  output = [T]
+  if compute_schur_vectors:
+    output.append(vs)
+
+  return tuple(output)
```
jakevdp, line 1495: nit; maybe cleaner:
```
return (T, vs) if compute_schur_vectors else (T,)
```

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+
+
+def _schur_cpu_translation_rule(c, operand, *, compute_schur_vectors,
+                               sort_eig_vals, select_callable):
+  shape = c.get_shape(operand)
+  batch_dims = shape.dimensions()[:-2]
+
+  _cpu_gees = lapack.gees
```
jakevdp, line 1503: Before doing this, perhaps we should check the jaxlib version and raise an informative error if it is too old. As currently written, if the user has too old a version of jaxlib it will be an unadorned AttributeError.

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1441,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+  return xops.Tuple(c, output)
+
+
+def _schur_batching_rule(batched_args, batch_dims, *, compute_schur_vectors,
+                        sort_eig_vals, select_callable):
+  x, = batched_args
+  bd, = batch_dims
+  x = batching.moveaxis(x, bd, 0)
+
+  return (schur_p.bind(
```
jakevdp, line 1539: Outer parenthesis is not necessary.

tests/linalg_test.py
```diff
@@ -1440,5 +1440,43 @@ def test_tridiagonal_solve(self, dtype):
...
+          self.skipTest("Schur LAPACK wrapper only implemented for jaxlib versions >= (0, 1, 72)")
+      rng = jtu.rand_default(self.rng())
+      batch_size = 10
+      shape = (batch_size, ) + shape
+      args = rng(shape, dtype)
+      Ts, Ss = vmap(lax.linalg.schur)(args)
+      for i in range(batch_size):
+          self.assertAllClose(
+              (jnp.matmul(jnp.matmul(Ss[i], Ts[i]), jnp.conjugate(Ss[i].T))),
+              args[i], atol=1e-4)
```
jakevdp, line 1479: Perhaps it would be more clear to do something like:
```python
f = vmap(lambda S, T: S @ T @ jnp.conj(S.T))
self.assertAllClose(f(Ss, Ts), args)
```

jakevdp (changes requested):
Looking good! A few more comments

jax/_src/lax/linalg.py
SaturdayGenfo: I imported the jaxlib version `from jax._src.lib import version as jaxlib_version` and added a condition that raises a NotImplemented error before the reference to `lapack.gees`.

jax/_src/lax/linalg.py
SaturdayGenfo: I'm sorry I couldn't figure out which two lines "here" and "below" referred to

jakevdp: the indentation of the parameters on the line after the function definition should match the indentation of the parameters on the first line. i.e. this:
```python
def _schur_impl(operand, *, compute_schur_vectors, sort_eig_vals,
                select_callable):
```
rather than this:
```python
def _schur_impl(operand, *, compute_schur_vectors, sort_eig_vals,
               select_callable):
```
There are several other places where this comment applies in the code below this.

jax/_src/lax/linalg.py
```diff
@@ -1441,3 +1442,119 @@ def tridiagonal_solve(dl, d, du, b):
...
+
+
+def _schur_cpu_translation_rule(c, operand, *, compute_schur_vectors,
+                                sort_eig_vals, select_callable):
+  shape = c.get_shape(operand)
+  batch_dims = shape.dimensions()[:-2]
+
+  if jaxlib_version < (0, 1, 72):
+    raise NotImplementedError(
+        "The Schur primitive is only implemented for jaxlib versions >= (0, 1, 72)"
```
jakevdp, line 1501: For user-facing code, I would say "versions >= 0.1.72", because this is how versions are formatted when installing packages.

jax/_src/lax/linalg.py
SaturdayGenfo: Got it, thank you. I think the current commit fixes this.

jakevdp (approved):

PR #8035, opened 2021-09-28
Add `--jax_platforms` flag to replace `--jax_platform_name`.

skye:
The motivation for this change is to make it possible to avoid
initializing unused backends, which may have undesirable side effects
(e.g. GPU memory allocation, only one process can use a Cloud TPU at a
time). It also provides a new and more flexible mechanism for
configuring the default backend, in order to minimize the number of
configs we have.

### DIFF
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -53,14 +53,22 @@ flags.DEFINE_string(
 flags.DEFINE_string(
     'jax_backend_target', '',
     'Either "local" or "rpc:address" to connect to a remote service target.')
+# TODO(skye): warn when this is used once we test out --jax_platforms a bit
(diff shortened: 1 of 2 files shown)


skye:
This is rebased on https://github.com/google/jax/pull/8034, I suggest looking at the second commit only or waiting for that to go in.

hawkinsp (approved):

skye:
This is now rebased directly on main.

jakevdp:
Should this also deprecate `jax_xla_backend`, which advertises itself as an alias for `jax_platform_name`?

skye:
Oh good catch. I'll change the description of `jax_xla_backend`, either in the PR or a follow-up.

skye:
Didn't refresh in time to see the merge, I guess a follow-up :)

Issue #3330, opened 2020-06-05
Unexpected / unintuitive behaviour of tuple(DeviceArray)


szepi1991:
`tuple(DeviceArray)` evaluates the DeviceArray (at least when using xla) and returns them as np arrays. 

Example:
```python
ar = jnp.ones(shape=3, dtype=jnp.float32)    # type: jax.interpreters.xla.DeviceArray,
tuple(ar)
```
produces a tuple of `numpy.float32: (1.0, 1.0, 1.0)`, while
```python
tuple(ar[i] for i in range(ar.shape[0]))
```
produces a tuple of DeviceArrays.

It feels like `tuple(ar)` returning a tuple of DeviceArrays would make sense. Is there some drawback to this?

Furthermore, when jitting vs not-jitting the following:
```python
def tuple_split(ar):
  return tuple(ar)
```
I do get different return types like so:
```python
print("orig:", tuple_split(jnp.ones(shape=3, dtype=jnp.float32)))
> orig: (1.0, 1.0, 1.0)
print("jit:", jit_tuple_split(jnp.ones(shape=3, dtype=jnp.float32)))
> jit: (DeviceArray(1., dtype=float32), DeviceArray(1., dtype=float32), DeviceArray(1., dtype=float32))
```
which I also didn't expect.

xla tuple implementation: https://github.com/google/jax/blob/master/jax/interpreters/xla.py#L1044

mattjj:
Great question! Take a look at #965 for an explanation of why things are currently implemented this way.

I suppose we could make `tuple(device_array)` return a tuple of DeviceArrays _even if_ we kept `DeviceArray.__iter__` implemented via bouncing to the host, i.e. by writing it like `map(device_put, self._value.__iter__())`.

@hawkinsp any thoughts to add?

szepi1991:
I see, thank you for sharing that, that makes sense.

My understanding is that your suggestion would copy contents to the host, iterate there, then copy back onto the device. Is my intuition correct that copying back and forth may be slow, but the idea is that it's still faster than taking slices on the device, at least when there are many slices?

I'm now somewhat torn since I could also reasonably expect that __iter__ and tuple() do the same thing.

Is it fairly usual that the original and the jitted function have different return types like in my above example? Can you simply construct another example (other than using __iter__ or tuple)?

shoyer:
From an API consistency perspective, I think it would be best to always return DeviceArray objects.

[...]

PR #8043, opened 2021-09-29
DeviceArray.__iter__ returns DeviceArrays, without host sync

### LINKED ISSUE #1583: List comprehensions and for-loops on xla.DeviceArrays return numpy.ndarrays
Looping through a `DeviceArray` like in a list comprehension, for loop, or enumerate returns `numpy.ndarray`. Is this expected behaviour?

```python
import jax.numpy as np

X = np.array(range(3)).reshape(3,1)
print(type(X))
#<class 'jax.interpreters.xla.DeviceArray'>

x0,x1,x2 = [xi for xi in X]
print(type(x0))
#<type 'numpy.ndarray'>

for xi in X:
  print(type(xi))
#<type 'numpy.ndarray'>
#<type 'numpy.ndarray'>
#<type 'numpy.ndarray'>
```

This is unexpected to me. Why are these not also `DeviceArray`s?

hawkinsp:
This is a clone of
https://github.com/google/jax/pull/3821

rebased on top of current `main`.

Original description:

cross-ref #1583 #3330 

The current (before this PR) implementation of `DeviceArray.__iter__` fom #965 pulls the array back to the host and does the slicing on the resulting NumPy array, effectively:

```python
# current implementation
class DeviceArray:
  def __iter__(self):
    return iter(self._value)  # self._value is the cached ndarray version of the array
```

There are three issues we want to address here:
1. `list(device_array)` shouldn't return a list of `numpy.ndaray`s, and instead should return a list of `DeviceArray`s, regardless of how those are computed (this issue came up in #1583 and in some JAX user chats recently);
2. `key, subkey = random.split(key)` in op-by-op mode shouldn't incur device synchronization, but it does curently because the we effectively call `random.split(key).__iter__()` (this issue came up in #3330 and in JAX chats);
3. from #965, we don't want expressions like `list(device_array)` or `list(jnp.arange(10000))` to be slow to evaluate, where before #965 there could be a several-second compilation time and relatively slow (~100ms) execution time.

To address (1) we could either keep the bounce via NumPy and add some device_puts, or keep all the operations on the device. To address (2) we want to keep all the operations on the device. Currently that means we need to compile and execute XLA computations (rather than just doing something in the runtime).

To address (3), i.e. to keep things performant, most importantly we don't want to incur big compilation times, which basically means we don't want to compile computations with large output arity. We also don't want to dispatch lots of XLA computations, to keep execution time reasonable. We can balance the two by chunking things, so that for chunk size C and array size N we end up compiling O(1) computations, none having output arity more than C, and dispatching about 2 N / C computations (for each chunk, one to slice it out of the original array and one to explode it).

I timed the current and new implementation this way:

```python
import time
import jax.numpy as jnp
from jax.interpreters import xla

def clear_caches():
  xla.xla_primitive_callable.cache_clear()
  xla._xla_callable.cache_clear()

def time_things(x):
  # time to grab the first element, nothing compiled
  times = []
  for _ in range(100):
    clear_caches()
    tic = time.time()
    next(iter(x))
    times.append(time.time() - tic)
  print(f"time to grab first element, uncompiled: {sum(times) / len(times) * 1000} ms")

  # time to grab first element, compiled
  times = []
  for _ in range(100):
    tic = time.time()
    next(iter(x))
    times.append(time.time() - tic)
  print(f"time to grab first element, compiled: {sum(times) / len(times) * 1000} ms")

  # time to grab the whole thing, nothing compiled
  times = []
  for _ in range(100):
    clear_caches()
    tic = time.time()
    list(x)
    times.append(time.time() - tic)
  print(f"total time to grab all elements, uncompiled: {sum(times) / len(times) * 1000} ms")
  print(f"average time to grab each element, uncompiled: {sum(times) / len(times) / x.shape[0] * 1000} ms")

  # time to grab the whole thing, compiled
  times = []
  for _ in range(100):
    tic = time.time()
    list(x)
    times.append(time.time() - tic)
  print(f"total time to grab all elements, compiled: {sum(times) / len(times) * 1000} ms")
  print(f"average time to grab each element, compiled: {sum(times) / len(times) / x.shape[0] * 1000} ms")

small = jnp.arange(10) + 1
big = jnp.arange(10000) + 1
```

Notice we are *not* calling `block_until_ready` and instead just looking at dispatch times. The numbers that come out are pretty noisy.

For the *current* implementation, on a TPUv3 internal colab we see this:

```
time_things(small)
time to grab first element, uncompiled: 0.001392364501953125 ms
time to grab first element, compiled: 0.001277923583984375 ms
total time to grab all elements, uncompiled: 0.003528594970703125 ms
average time to grab each element, uncompiled: 0.0003528594970703125 ms
total time to grab all elements, compiled: 0.0027894973754882812 ms
average time to grab each element, compiled: 0.0002789497375488281 ms

time_things(big)
time to grab first element, uncompiled: 0.0012826919555664062 ms
time to grab first element, compiled: 0.0011134147644042969 ms
total time to grab all elements, uncompiled: 0.7117271423339844 ms
average time to grab each element, uncompiled: 7.117271423339843e-05 ms
total time to grab all elements, compiled: 0.7156133651733398 ms
average time to grab each element, compiled: 7.156133651733398e-05 ms
```

For the *new* implementation with chunk_size=100, on a TPUv3 internal colab we see this:

```
time_things(small)
time to grab first element, uncompiled: 12.47842788696289 ms
time to grab first element, compiled: 0.1662755012512207 ms
total time to grab all elements, uncompiled: 12.541697025299072 ms
average time to grab each element, uncompiled: 1.2541697025299074 ms
total time to grab all elements, compiled: 0.16188859939575195 ms
average time to grab each element, compiled: 0.016188859939575195 ms

time_things(big)
time to grab first element, uncompiled: 88.96037578582764 ms
time to grab first element, compiled: 0.809943675994873 ms
total time to grab all elements, uncompiled: 182.55109786987305 ms
average time to grab each element, uncompiled: 0.018255109786987307 ms
total time to grab all elements, compiled: 80.00826835632324 ms
average time to grab each element, compiled: 0.008000826835632323 ms
```

I _think_ the ballparks here seem acceptable, and so meet desideratum (3) while avoiding the host sync per (2). Comparing to #965, we avoid any multi-second compile times even though we end up paying 80ms to compute `list(big)` rather than <1ms. I suspect these times will get better when we improve `jit` dispatch time, since a significant fraction of the trace is spent on Python overhead (anything that's not an Execute bar):

image

Here's one other benchmark:

```python
from jax import random 
import time

def time_things2():
  key = random.PRNGKey(0)
  key, _ = random.split(key)
  
  tic = time.time()
  for _ in range(1000):
  toc = time.time()
  print(f"{(toc - tic)} ms")
```

With the *current* implementation, on TPU we get:

```
time_things2()
0.27765631675720215 ms
```

With the *new* implementation, on TPU we get:

```
time_things2()
0.20100140571594238 ms
```

So perhaps we're saving something from avoiding the sync here, even though there's no other work going on here.

fixes #1583

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -919,7 +919,7 @@ def dynamic_slice(operand: Array, start_indices: Sequence[Array],
   """
   start_indices = _dynamic_slice_indices(operand, start_indices)
   return dynamic_slice_p.bind(operand, *start_indices,
-                              slice_sizes=tuple(slice_sizes))
+                              slice_sizes=core.canonicalize_shape(slice_sizes))
 
 def dynamic_update_slice(operand: Array, update: Array,
                          start_indices: Array) -> Array:
@@ -1355,7 +1355,7 @@ def transpose(operand: Array, permutation: Sequence[int]) -> Array:
   <https://www.tensorflow.org/xla/operation_semantics#transpose>`_
   operator.
   """
-  permutation = tuple(permutation)
+  permutation = tuple(operator.index(d) for d in permutation)
   if (permutation == tuple(range(np.ndim(operand)))
       and isinstance(operand, (core.Tracer, xla.DeviceArray))):
     return operand
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -6695,6 +6695,22 @@ def _multi_slice(arr,
     results.append(sliced)
   return results
 
+# The next two functions are related to iter(device_array), implemented here to
+# avoid circular imports.
+@jit
+def _unstack(x):
+  return [lax.index_in_dim(x, i, keepdims=False) for i in range(x.shape[0])]
+setattr(DeviceArray, "_unstack", _unstack)
+def _chunk_iter(x, size):
+  if size > x.shape[0]:
+    yield x
+  else:
+    num_chunks, tail = divmod(x.shape[0], size)
+    for i in range(num_chunks):
+      yield lax.dynamic_slice_in_dim(x, i * size, size)
+    if tail:
+      yield lax.dynamic_slice_in_dim(x, num_chunks * size, tail)
+setattr(DeviceArray, "_chunk_iter", _chunk_iter)
 
 # Syntactic sugar for scatter operations.
 class _IndexUpdateHelper:
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -622,21 +622,34 @@ def _sda_value(self):
 
 
 def _sda__getitem__(self, idx):
+  self._check_if_deleted()
   if not isinstance(idx, tuple):
     cidx = (idx,) + (slice(None),) * (len(self.aval.shape) - 1)
   else:
     cidx = idx + (slice(None),) * (len(self.aval.shape) - len(idx))
-  try:
-    buf_idx = self.indices.index(cidx)
-  except ValueError:
-    # NOTE: Slow path, this will materialize the sharded array on a single
-    # device and use XLA's Gather to index into the resulting array.
-    return xla.DeviceArray.__getitem__(self, idx)
+  if self._npy_value is None:
+    try:
+      buf_idx = self.indices.index(cidx)
+    except ValueError:
+      buf_idx = None
+    if buf_idx is not None:
+      buf = self.device_buffers[buf_idx]
+      aval = ShapedArray(buf.xla_shape().dimensions(), self.aval.dtype)
+      return xla.make_device_array(aval, None, buf)
+  return super(self.__class__, self).__getitem__(idx)
+
+
+def _sda__iter__(self):
+  if self.ndim == 0:
+    raise TypeError("iteration over a 0-d array")  # same as numpy error
   else:
-    self._check_if_deleted()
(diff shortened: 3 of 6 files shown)


mattjj (approved):
Thanks, Peter!

hawkinsp:
I'm going to hold off merging this PR until after the ICLR deadline (Oct 5), since I think there's some chance of performance changes if nothing else.

jakevdp:
A note: this change seems to have broken matplotlib's histogram function when a `DeviceArray` is passed to it (with most recent matplotlib and JAX releases):
```python
import jax.numpy as jnp
import numpy as np

fix, ax = plt.subplots(1, 2, figsize=(10, 5))
ax[0].hist(np.arange(10))
ax[1].hist(jnp.arange(10));
```
download-3

I believe this should be considered a matplotlib bug (see discussion in https://github.com/google/jax/issues/8899#issuecomment-991286825) but the new behavior here is what resulted in the problematic code path.

soraros:
Maybe this deserves a sharp-edges entry?

jakevdp:
matplotlib issue fixed in https://github.com/matplotlib/matplotlib/issues/22016

PR #8047, opened 2021-09-30
Canonicalize shapes in jax.image.resize().

hawkinsp:
This makes jax.image.resize() robust to having jnp arrays passed as sizes. It also turns out some users were passing floating point values here, and this means they are correctly flagged as errors.


hawkinsp:
Closing in favor of https://github.com/google/jax/pull/8048

PR #8054, opened 2021-09-30
Mark --jax_xla_backend as deprecated.

skye:


skye:
cc @jakevdp

hawkinsp (approved):
I wonder if we can start issuing warnings if these are set.

skye:
+1 to warnings. I'm hoping to get a few "alpha testers" to confirm this is a reasonable API, and then I can add warnings once we're more confident we won't need to make further changes.

PR #8055, opened 2021-09-30
Uniquify variable names globally within a jaxpr.

hawkinsp:
It is confusing when the same name is shadowed within an inner lambda expression. Use globally unique variable names in each pretty-printed jaxpr.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import collections
 from collections import namedtuple
 from contextlib import contextmanager
 from functools import partial, total_ordering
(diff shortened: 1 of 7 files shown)


jax/core.py
```diff
@@ -2050,104 +2053,139 @@ def check_map(prim, in_avals, params):
 
 
 # ------------------- Jaxpr printed representation -------------------
-def pp_vars(vs: Sequence[Any], *, print_shapes: bool = False) -> pp.Doc:
+
+# A JaxprPpContext allow us to uniquify variable names within nested Jaxprs.
+class JaxprPpContext:
```
mattjj, line 2058: We could consider replacing `JaxprPpContext` with a `defaultdict(itertools.count().__next__)`. That would mean we don't need the logic about incrementing `num_vars` etc, and we could just use indexing syntax in `pp_var`.

mattjj (approved):
Nice!

I left three comments but really they just amount to one implementation suggestion.

jax/core.py
hawkinsp: Done, that's a neat trick I didn't know.

PR #8062, opened 2021-10-01
[sparse]: use namedtuple for bcoo validation

jakevdp:
This makes it more convenient to use the output of validation.

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -32,7 +32,7 @@ Further down are some examples of potential high-level wrappers for sparse objec
 import functools
 import operator
 
-from typing import Any, Sequence, Tuple
+from typing import Any, NamedTuple, Sequence, Tuple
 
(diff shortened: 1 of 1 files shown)


jax/experimental/sparse/ops.py
```diff
@@ -610,7 +611,11 @@ def _unbatch_bcoo(data, indices, shape):
   return data, jnp.hstack([batch_indices, indices])
 
 
-def _validate_bcoo(data, indices, shape):
+BCOOProperties = collections.namedtuple(
```
yashk2810, line 614: Maybe use typing.NamedTuple?

yashk2810 (approved):

jax/experimental/sparse/ops.py
jakevdp: Good idea - Done!

yashk2810 (approved):

PR #8065, opened 2021-10-01
DOC: add deprecation message to index_update and friends

### LINKED ISSUE #8064: Update jax.ops.index_* documentation to include current preferred index notation
If you search on Google for how to update array indices in JAX, you end up on [this documentation page](https://jax.readthedocs.io/en/latest/_autosummary/jax.ops.index_update.html#jax-ops-index-update) for `jax.ops.index_update`.

`jax.ops.index_update` is no longer the best way to update indices -- in most cases users should instead write code like `x.at[idx].set(y)`, which is a really great improvement!

The documentation for the `jax.ops.index_*` should also describe and/or link to the alternate, simpler, and preferred notation for working with indices.

jakevdp:
Fixes #8064

Preview the new docs here: https://jax--8065.org.readthedocs.build/en/8065/_autosummary/jax.ops.index_update.html#jax.ops.index_update

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -6690,22 +6690,43 @@ class _IndexUpdateHelper:
 
   The ``at`` property is syntactic sugar for calling the indexed update functions
   defined in :mod:`jax.ops`, and acts as a pure equivalent of in-place
-  modificatons. For further information, see `Indexed Update Operators
-  <https://jax.readthedocs.io/en/latest/jax.ops.html#indexed-update-operators>`_.
+  modificatons.
 
   In particular:
 
-  - ``x = x.at[idx].set(y)`` is a pure equivalent of ``x[idx] = y``.
-  - ``x = x.at[idx].add(y)`` is a pure equivalent of ``x[idx] += y``.
-  - ``x = x.at[idx].multiply(y)`` (aka ``mul``) is a pure equivalent of
-    ``x[idx] *= y``.
-  - ``x = x.at[idx].divide(y)`` is a pure equivalent of ``x[idx] /= y``.
-  - ``x = x.at[idx].power(y)`` is a pure equivalent of ``x[idx] **= y``.
-  - ``x = x.at[idx].min(y)`` is a pure equivalent of
-    ``x[idx] = minimum(x[idx], y)``.
-  - ``x = x.at[idx].max(y)`` is a pure equivalent of
-    ``x[idx] = maximum(x[idx], y)``.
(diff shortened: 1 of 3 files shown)


hawkinsp:
This looks good.

There is one further step we probably should do. The `index_mul` etc docs have an additional and necessary paragraph:

"
Unlike the NumPy code x[idx] *= y, if multiple indices refer to the same location the updates will be multiplied. (NumPy would only apply the last update, rather than multiplying the updates.) The order in which conflicting updates are applied is implementation-defined and may be nondeterministic (e.g., due to concurrency on some hardware platforms).
"

We should probaby add that to the `.at[...]` documentation, suitably modified.

hawkinsp (approved):

hawkinsp:
> This looks good.
> 
> There is one further step we probably should do. The `index_mul` etc docs have an additional and necessary paragraph:
> 
> " Unlike the NumPy code x[idx] *= y, if multiple indices refer to the same location the updates will be multiplied. (NumPy would only apply the last update, rather than multiplying the updates.) The order in which conflicting updates are applied is implementation-defined and may be nondeterministic (e.g., due to concurrency on some hardware platforms). "
> 
> We should probaby add that to the `.at[...]` documentation, suitably modified.

There is also some useful parameter documentation on those functions that needs to be transferred. Maybe for a future PR, though.

jakevdp:
Added the extra paragraph, as well as a TODO to document additional keyword arguments.

jax/_src/numpy/lax_numpy.py
```diff
@@ -6690,22 +6690,43 @@ class _IndexUpdateHelper:
...
+  ``x = x.at[idx].get()``         ``x = x[idx]``
+  ==============================  ================================
+
+  None of these expressions modify the original ``x``; instead they return
+  a modified copy of ``x``. However, inside a :py:func:`jax.jit` compiled function,
+  expressions like ``x = x.at[idx].set(y)`` are guaranteed to be applied in-place.
+
+  Unlike NumPy in-place operations such as :code:`x[idx] += y`, if multiple
+  indices refer to the same location, all updates will be applied (NumPy would
+  only apply the last update, rather than multiplying the updates.) The order
```
hawkinsp, line 6716: "multiplying" only makes sense for `index_mul` perhaps. But instead "applying" might work.

hawkinsp (approved):

jax/_src/numpy/lax_numpy.py
jakevdp: Done.

PR #8066, opened 2021-10-01
[sparse] Cleanup: add dot_general shape utility

jakevdp:
This is a cleanup on the way to implementing full sparse-sparse dot product.

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -51,11 +51,14 @@ import jax.numpy as jnp
 import numpy as np
 from jax.interpreters import ad
 from jax.util import safe_zip
-from jax._src.lax.lax import ranges_like, remaining, _dot_general_batch_dim_nums, _dot_general_shape_computation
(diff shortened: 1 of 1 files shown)


jax/experimental/sparse/ops.py
```diff
@@ -979,6 +979,14 @@ def _bcoo_transpose_batch_rule(batched_args, batch_dims, *, permutation, shape):
 
 bcoo_dot_general_p = core.Primitive('bcoo_dot_general')
 
+def _dot_general_validated_shape(lhs_shape, rhs_shape, dimension_numbers):
```
yashk2810, line 982: From my experience reading jax code. Since the variable names are not descriptive, maybe add type annotations here?

(The variable names here are descriptive, but was saying in general type annotations help a lot)

yashk2810 (approved):

jax/experimental/sparse/ops.py
jakevdp: Done, thanks!

PR #8067, opened 2021-10-02
add experimental RngBitGenerator ("RBG") PRNG

mattjj:
Builds on the awesome #6899 upgrade.

Not only is this an experimental API, but also because the RngBitGenerator is not guaranteed to be stable across compiler versions (or backends or shardings), let's assert that this JAX PRNG implementation may not be stable across JAX versions.

Even without that kind of stability, this PRNG is still useful because unlike the effectful RNG primitives, such as lax.rng_uniform, at a given jax/jaxlib version this RBG PRNG will still work correctly with lax.scan and jax.checkpoint (while still potentially being more performant on some platforms than JAX's standard PRNG).

Thanks to #6899, it's super easy to use, just by adapting key creation to use `prng.seed_with_impl` and then using the `jax.random` API as usual:

```python
In [1]: from jax import prng
In [2]: key = prng.seed_with_impl(prng.rbg_prng_impl, 87)

In [3]: import jax

In [4]: key, subkey = jax.random.split(key)

In [5]: jax.random.uniform(subkey)
Out[5]: DeviceArray(0.62572646, dtype=float32)
```

cc @levskaya @jekbradbury @zhangqiaorjc 

Here's an illustration of how things can go horribly wrong with the effectful `lax.rng_uniform`:

```python
import jax
from jax import lax
import jax.numpy as jnp

def scanned_fun(_, x):
  r = lax.rng_uniform(0., 1., ())
  return None, x * r

# without differentiation we're okay
_, ys = lax.scan(scanned_fun, None, jnp.ones(5))
print(ys)
# [0.59189475 0.00664103 0.17746115 0.23565137 0.69601536]

# grad-of-scan-of-effectful-rng = danger
gs = jax.grad(lambda xs: lax.scan(scanned_fun, None, xs)[1].sum())(jnp.ones(5))
print(gs)
# [0.90172756 0.90172756 0.90172756 0.90172756 0.90172756]

# remat doesn't save you
gs = jax.grad(lambda xs: lax.scan(
    jax.remat(scanned_fun), None, xs)[1].sum())(jnp.ones(5))
print(gs)
# [0.8394314 0.8394314 0.8394314 0.8394314 0.8394314]

# forward pass is messed up too (though only after differentiation)
ys, vjp = jax.vjp(lambda xs: lax.scan(scanned_fun, None, xs)[1], jnp.ones(5))
gs, = vjp(jnp.ones(5))
print(ys)
print(gs)
# [0.32732797 0.32732797 0.32732797 0.32732797 0.32732797]
# [0.32732797 0.32732797 0.32732797 0.32732797 0.32732797]
```

Here's RBG helping us out:

```python
import jax
from jax import lax
from jax import prng
from jax import random
import jax.numpy as jnp

def scanned_fun(key, x):
  r = random.uniform(subkey, ())
  return key, x * r

key = prng.seed_with_impl(prng.rbg_prng_impl, 87)
_, ys = lax.scan(scanned_fun, key, jnp.ones(5))
print(ys)
# [0.62572646 0.34407544 0.94847405 0.15406036 0.3797592 ]

gs = jax.grad(lambda xs: lax.scan(scanned_fun, key, xs)[1].sum())(jnp.ones(5))
print(gs)
# [0.62572646 0.34407544 0.94847405 0.15406036 0.3797592 ]

gs = jax.grad(lambda xs: lax.scan(jax.remat(scanned_fun), key, xs)[1].sum())(jnp.ones(5))
print(gs)
# [0.62572646 0.34407544 0.94847405 0.15406036 0.3797592 ]


ys, vjp = jax.vjp(lambda xs: lax.scan(scanned_fun, key, xs)[1], jnp.ones(5))
gs, = vjp(jnp.ones(5))
print(ys)
print(gs)
# [0.62572646 0.34407544 0.94847405 0.15406036 0.3797592 ]
# [0.62572646 0.34407544 0.94847405 0.15406036 0.3797592 ]
```

Ah, precisely the kind of reproducibility we want: different across iterations, the same across the forward and backward passes (and repeatable too!).

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6741,24 +6741,44 @@ xla.translations[rng_uniform_p] = _rng_uniform_translation_rule
 
 
 def _rng_bit_generator_shape_rule(key, *, shape, dtype, algorithm):
-  _ = dtype, algorithm
+  del dtype, algorithm
   return (key.shape, tuple(shape))
 
 
 def _rng_bit_generator_dtype_rule(key, *, shape, dtype, algorithm):
-  _ = key, shape, algorithm
+  del shape, algorithm
   return (key.dtype, dtype)
 
 
 def _rng_bit_generator_weak_type_rule(key, *, shape, dtype, algorithm):
-  _ = shape, dtype, algorithm
+  del shape, dtype, algorithm
   return (key.weak_type, False)
 
 
 def _rng_bit_generator_translation_rule(c, key, *, shape, dtype, algorithm):
-  _ = c
+  key_shape, key_dtype = c.get_shape(key).dimensions(), c.get_shape(key).numpy_dtype()
+  # While the RngBitGenerator HLO accepts a u64[2] key on all backends, we
+  # typically represent the key argument to this primitive as a u32[4] so as to
+  # sidestep issues with the jax_enable_x64=False configuration. As a result, we
+  # need to convert u32[4] -> u64[2] here in the translation rule. However, we
+  # also polymorphically allow a u64[2] for backward compatibility.
+  assert ((key_shape == (4,) and key_dtype == dtypes.dtype('uint32')) or
+          (key_shape == (2,) and key_dtype == dtypes.dtype('uint64')))
   xla_shape = xc.Shape.array_shape(np.dtype(dtype), shape)
-  return xops.RngBitGenerator(algorithm, key, xla_shape)
+  if key_dtype == dtypes.dtype('uint32'):
+    u64_etype = xla_client.dtype_to_etype(dtypes.dtype('uint64'))
+    # TODO(mattjj): use BitcastConvertType implementation with newer jaxlib
+    # new_key = xops.BitcastConvertType(xops.Reshape(key, (2, 2)), u64_etype)
+    new_key = xla_bridge.constant(c, np.zeros(2, dtype=np.dtype('uint64')),
+                                  canonicalize_types=False)
+    for i in range(4):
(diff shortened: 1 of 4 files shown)


jax/_src/lax/lax.py
```diff
@@ -6755,12 +6755,28 @@ def _rng_bit_generator_weak_type_rule(key, *, shape, dtype, algorithm):
...
-def _rng_bit_generator_translation_rule(c, key, *, shape, dtype, algorithm):
-  _ = c
+def _rng_bit_generator_translation_rule_tpu(c, key, *, shape, dtype, algorithm):
   xla_shape = xc.Shape.array_shape(np.dtype(dtype), shape)
   return xops.RngBitGenerator(algorithm, key, xla_shape)
 
 
+def _rng_bit_generator_translation_rule_cpu_gpu(c, key, *, shape, dtype, algorithm):
+  assert c.get_shape(key).dimensions() == (4,)
+  # convert the u32[4] key to a u64[2] key
```
mattjj, line 6765: This is necessary because of how uneven RngBitGenerator is across backends. On CPU and GPU it requires a uint64[2] key or it will crash, while on TPU it will accept a uint32[2] or uint64[2].

We make things more uniform in JAX by always tracking the key as a uint32[4] (for the same number of bits), then on CPU/GPU manually converting it to a uint64[2]. 

When the next jaxlib is released, we should be able to replace most of this implementation with a single BitcastConvertType.

froystig (approved):
nice!!

PR #8081, opened 2021-10-04
[sparse] add support for sparse matrix-vector product

jakevdp:
Previously this implementation only supported sparse dot products with one sparse dimension on each side. This version supports multiple sparse dimensions on the left hand side, but still requires a single sparse dimension on the right.

This also contains one breaking change: previously the (batched) sparse vector-vector product returned a dense matrix; now any sparse-sparse dot product will return a sparse BCOO matrix.

### DIFF
--- a/jax/experimental/sparse/ops.py
+++ b/jax/experimental/sparse/ops.py
@@ -1207,57 +1207,63 @@ def bcoo_spdot_general(lhs_data, lhs_indices, rhs_data, rhs_indices, *, lhs_shap
   return bcoo_spdot_general_p.bind(lhs_data, lhs_indices, rhs_data, rhs_indices,
                                    lhs_shape=lhs_shape, rhs_shape=rhs_shape, dimension_numbers=dimension_numbers)
 
-def _bcoo_vv(lhs_data, lhs_indices, rhs_data, rhs_indices, *, size, dtype):
-  # Compute the dot product of two sparse vectors, returning a scalar
-  assert lhs_data.ndim == rhs_data.ndim == 1
-  assert lhs_indices.shape == (lhs_data.size, 1)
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/ops.py
```diff
@@ -1207,82 +1207,100 @@ def bcoo_spdot_general(lhs_data, lhs_indices, rhs_data, rhs_indices, *, lhs_shap
...
-  assert lhs_indices.shape == (lhs_data.size, 1)
-  assert rhs_indices.shape == (rhs_data.size, 1)
-  indices = jnp.vstack([lhs_indices, rhs_indices])
-  size = min(indices.shape[0], size)
-  _, inv = jnp.unique(indices, size=size, return_inverse=True)
-  lhs_inv, rhs_inv = inv[:len(lhs_data)], inv[len(lhs_data):]
-  lhs = jnp.zeros(size, dtype).at[lhs_inv].add(lhs_data)
-  rhs = jnp.zeros(size, dtype).at[rhs_inv].add(rhs_data)
-  return lax.dot_general(lhs, rhs, dimension_numbers=(([0], [0]), ([], [])))
+def _bcoo_Mv(lhs_data, lhs_indices, rhs_data, rhs_indices, *, lhs_shape, rhs_shape, dtype, lhs_contract):
```
yashk2810, line 1210: I usually prefer NamedTuples as input when its so many arguments because its really easy to pass something wrong or maybe just put `*` at the start so everything becomes a kwarg?

But this is fine too!

yashk2810 (approved):

jax/experimental/sparse/ops.py
jakevdp: Thanks - I chose this approach because the intent of this function is to be used with`vmap`, so positional arguments are more direct

PR #8090, opened 2021-10-05
Include XLA_FLAGS in persistent compilation cache key.

skye:
This is to prevent false cache hits when the compiler behavior is
changed via flags. Flags known to not affect the compiled executable
(e.g. dumping HLO) are excluded from the key.

### DIFF
--- a/jax/experimental/compilation_cache/compilation_cache.py
+++ b/jax/experimental/compilation_cache/compilation_cache.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import hashlib
+import os
 import re
+import sys
 
 import jax
 from jax.experimental.compilation_cache.file_system_cache import FileSystemCache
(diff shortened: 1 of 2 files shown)


skye:
cc @trevorcai

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -145,6 +154,40 @@ def _hash_platform(hash_obj, backend):
...
+    "--xla_dump_hlo_as_text",
+    "--xla_dump_to",
+    "--xla_force_host_platform_device_count",
+    "--xla_dump_disable_metadata",
+    "--xla_dump_hlo_pipeline_re",
+]
+
+def _hash_xla_flags(hash_obj):
+  # All XLA flags must be specified via the XLA_FLAGS env var, both inside and
+  # outside of google (I think).
```
trevorcai, line 177: While this may be true externally, it's certainly not true internally, especially for TPUs. I'll send you an email internally with some references.

skye: I added logic to also look at sys.argv.

trevorcai: LGTM, but add explicit note wrt XLA absl flags must be passed in the form of `--flag_name=flag_value`, not with spaces?
(And confirmatory action sent internally.)

zhangqiaorjc (approved):

PR #8091, opened 2021-10-05
Add float0 support to equality and closeness check

eelregit:
`float0` arrays, as derivative of integer inputs and often part of pytrees, should be able to be compared by `check_eq` and `check_close`.

I don't know if adding float0 to the default tolerance dict has any downside.


eelregit:
@googlebot I signed it!

apaszke (approved):

PR #8094, opened 2021-10-05
Update doc for eigh.

sracaniere:
Mention `eigenvectors` before `eigenvalues` in doc to match the order of returned values.


jakevdp (approved):
Looks good, thanks!

PR #8103, opened 2021-10-05
Add DeprecationWarnings to jax.ops.index_... operators.

hawkinsp:
Remove uses of index_... in Common Gotchas notebook.

### DIFF
--- a/jax/_src/ops/scatter.py
+++ b/jax/_src/ops/scatter.py
@@ -14,7 +14,7 @@
 
 # Helpers for indexed updates.
 
-
+import warnings
 import sys
 from typing import Any, Callable, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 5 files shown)


jax/_src/ops/scatter.py
```diff
@@ -178,6 +178,8 @@ def index_add(x: Array,
                [1., 1., 1., 7., 7., 7.],
                [1., 1., 1., 1., 1., 1.]], dtype=float32)
   """
+  warnings.warn("index_add is deprecated. Use x.at[idx].add(y) instead.",
+                DeprecationWarning)
```
jakevdp, line 182: I would lean toward using `FutureWarning` instead, because I believe `DeprecationWarning` is ignored by default unless the function is called directly in `__main__`.

hawkinsp:
> I think this may fail due to the version of jupytext you have installed... I've been meaning to update our pre-commit hook to the latest version and re-generate the markdown

Which CI build will fail?

jax/_src/ops/scatter.py
jakevdp: OK, that makes sense.

jakevdp (approved):

jakevdp:
> Which CI build will fail?

I thought the `lint_and_typecheck` step would fail because of the updated jupytext version, but it looks like I was wrong.

PR #8107, opened 2021-10-05
jnp.array: replace host round-trip with on-device copy

jakevdp:


jax/_src/lax/lax.py
```diff
@@ -69,6 +69,23 @@
 DType = Any
 Shape = core.Shape
 
+def _array_copy(arr):
```
mattjj, line 72: Could you put a `jit` on this? That way XLA will optimize away the addition.

mattjj (approved):

jax/_src/lax/lax.py
jakevdp: Good call - done!

PR #8117, opened 2021-10-06
Add vmap support for rng_bit_generator.

LenaMartens:
Adds a naive batching rule which loops over the primitive, then stacks the results. This has the advantage that the vmapped result will be the same as calling the primitive n times, but is not very performant.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6826,6 +6826,25 @@ def _convert_2xU64_to_4xU32_without_bitcast(c, key):
 def _rng_bit_generator_named_shape_rule(key, *, shape, dtype, algorithm):
   return [key.named_shape, key.named_shape]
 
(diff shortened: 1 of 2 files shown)


froystig (approved):
This is an "inline" loop that will emit N individual primitive calls. Is there a batch dimension length at which `jax.lax.map` would make sense instead? We can follow up after this PR if there's an answer to that question.

froystig:
Should we give some kind of warning in the meantime, especially if the batch dimension is large?

froystig:
I'm not sure what's "too large" but it might depend on the backend/device. Conceptually we want a `map` and that alone might be reason enough to switch. One thing we could do is always use an explicit `scan` for the map and then tune its `unroll` parameter for performance. This has the additional advantage of not increasing the jaxpr size, only the HLO.

PR #8129, opened 2021-10-07
ShardedDeviceArray: lazily set avals for device_buffers

jakevdp:
Previously the `aval` could be set to `None`, which causes issues when people access the device buffers directly.

A prior eager fix had to be rolled-back for performance reasons.


jakevdp:
Looks like this will require a jaxlib C++ change as well...

hawkinsp (approved):
Looks good, but you'll need the jaxlib change for the test to pass.

jakevdp:
Replaced by #8312

PR #8134, opened 2021-10-08
jnp.array: handle raw device buffers

jakevdp:
For the sake of efficiency, `ShardedDeviceArray` buffers are not fully initialized, and do not have `aval` defined on each entry of `device_buffers`. This means that using them directly in jax operations can lead to errors.

This change updates `jax.numpy.array()` to handle this case, and properly create fully initialized DeviceArray objects when passed a raw buffer.


mattjj (approved):
LGTM modulo a question about `object.weak_type` that I sent in chat (probably just me not understanding...)

PR #8137, opened 2021-10-08
remove ShapedArray.__len__

mattjj:
It was confusing to overload, since we sometimes think of avals like shapes paired with dtypes, and in that case len(aval) should perhaps be like len(aval.shape). The only place where this behavior was relied on was sparse/ops.py.

(It bit @zhangqiaorjc recently!)


zhangqiaorjc (approved):
Thanks!

PR #8145, opened 2021-10-09
ahead-of-time lowering and compilation frontend for pjit

froystig:
Initial take on #7733 corresponding to `jax.pjit`. Generalizes some of the machinery introduced in #7997.


apaszke (approved):
LGTM. The comments I left on the `Lowered` class also apply to `Compiled` since they look very similar.

PR #3295, opened 2020-06-02
skip fori_loops with zero trip count

mattjj:
fixes #3285

See especially [this comment](https://github.com/google/jax/issues/3285#issuecomment-637655722) for an explanation.

Interestingly, we _can't_ make an analogous change to `lax.scan` because we need to infer the types of its scanned-over outputs (to produce outputs of the correct shape, even in the zero-iteration case), and to do that we need to run abstract interpretation on the scanned function anyway.

An alternative to this change is to close #3285 as "working as intended".

PR #8160, opened 2021-10-11
fix #8152

### LINKED ISSUE #8152: fori_loop with disable_jit fails when no actual iterations happen
In a trivial call of `fori_loop`, i.e. when `upper == lower`, a normal call works just fine,
but a call with `disable_jit` causes a `TypeError`.

Can be reproduced by 
```python
from jax import lax, disable_jit


def test():
    non_trivial = lax.fori_loop(lower=0, upper=3, body_fun=lambda x, i: x + i, init_val=1)
    print(f'non-trivial result: {non_trivial}')
    trivial = lax.fori_loop(lower=0, upper=0, body_fun=lambda x, i: x + i, init_val=1)
    print(f'trivial result: {trivial}')


test()
# non-trivial result: 4
# trivial result: 1

with disable_jit():
    test()
# non-trivial result: 4
# TypeError: tree_map() missing 1 required positional argument: 'tree'
```

This is because the [non-jit implementation of `lax.scan`](https://github.com/google/jax/blob/cd3d37f4d14bdce8501795c759be7a5b9f79e157/jax/_src/lax/control_flow.py#L1286)
calls `tree_multimap` to stack the outputs, but there aren't any outputs to stack.

I suggest to check for this special case (of a trivial array to scan over) in the non-jit branch of scan and directly return the initial carry and a trivial stacked_y.

Happy to make a PR if this is the way you want to do it.

Jakob-Unfried:
As discussed in #8152 , this is supposed to fix `lax.fori_loop` in `disable_jit()` context when no actual iterations
take place (i.e. when `upper == lower`).

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -210,6 +210,10 @@ def fori_loop(lower, upper, body_fun, init_val):
     use_scan = False
 
   if use_scan:
+    if config.jax_disable_jit and upper_ == lower_:
+      # non-jit implementation of scan does not support length=0
+      return init_val
+
     (_, result), _ = scan(_fori_scan_body_fun(body_fun), (lower_, init_val),
                           None, length=upper_ - lower_)
   else:
@@ -1284,6 +1288,8 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
       length, = unique_lengths
 
   if config.jax_disable_jit:
+    if length == 0:
+      raise ValueError("zero-length scan is not supported in disable_jit() mode because the output type is unknown.")
     carry = init
     ys = []
     maybe_reversed = reversed if reverse else lambda x: x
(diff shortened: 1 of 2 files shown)


jakevdp (changes requested):
Looks great - last thing we should add here is a regression test. You can use something like your snippet from the issue, and put it somewhere in `lax_control_flow_test.py`: https://github.com/google/jax/blob/d17633413deb0122fc8dedb334f70233a2c583c4/tests/lax_control_flow_test.py#L98

Probably ensuring that the results are the same with and without JIT is the best approach.

jakevdp (changes requested):
Looks good - I think we're landing on the right fix here. A couple comments below.

jakevdp (approved):

jakevdp:
> Are the concerns raised in #3295 relevant?

Thanks for pointing that out. I think this change is OK because it essentially does the following:
- replace a non-explicit error with an explicit error in the case of zero-length scan with disable_jit
- make fori_loop behavior consistent whether JIT is disabled or not.

And in particular, the concern in https://github.com/google/jax/pull/3295#discussion_r435925982 is in the spirit of my most recent review. But perhaps @gnecula would like to weigh-in here?

PR #8161, opened 2021-10-11
Allow `random.choice` and `random.permutation` on multidimensional arrays

### LINKED ISSUE #8158: `random.choice` on multidimensional arrays
Can we model `jax.random.choice` after the newer [`numpy.random.Generator.choice`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html) instead of [`numpy.random.choice`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html)? See [numpy discussion](https://github.com/numpy/numpy/issues/2724) for reference.

This requires adding an `axis=0` parameter which would generalize the current implementation and keep it backward-compatible.

juliuskunze:
Fixes https://github.com/google/jax/issues/8158.


jakevdp (changes requested):
Looks great!

One minor comment (feel free to ignore). Also, please add an entry to the changelog, and squash all changes into a single commit. Thanks!

jakevdp (approved):

juliuskunze:
@jakevdp Done. Let me know if I can help to fix the Google internal error.

jakevdp:
Thanks for your work on this!

PR #8180, opened 2021-10-12
WIP: [x64] change dtype default to 32-bit

jakevdp:
Step 1 of the process outlined in #8178


jakevdp:
I'm going to close this because it's out-of-date. It's not clear that we're going to continue with this approach.

PR #8191, opened 2021-10-13
rewrite jax.checkpoint (aka jax.remat), leave old implementation for now

mattjj:
The new `jax.checkpoint` is currently engaged whenever we pass an explicit policy. To get a "remat everything" decorator, one should write `@partial(jax.checkpoint, policy=lambda *_, **__: False)`.

Because the new code path is used whenever an explicit policy is passed, that means that all our parameterized remat tests already cover this new code path. (And I made sure it's covered, by putting an `assert False` in the new implementation and running the old tests!)

Fundamentally `jax.checkpoint` is about customizing the partial evaluation involved in autodiff. So our previous implementation was just a call primitive with a custom partial evaluation rule, inheriting most of its behaviors from the other call-processing machinery. (We also need a custom XLA translation rule, to foil CSE, and a custom transpose rule, to handle the nonlinear terms during transposition.)

But because JAX's partial evaluation works by data dependence only, that meant constants were "invisible" to `jax.checkpoint`, in the sense that they would unavoidably be evaluated / constant-folded and not recomputed.

It turns out that this behavior could be both surprising and costly. Here's a toy example where a potentially-large constant can be created behind the scenes:

```python
import jax
import jax.numpy as jnp

# utility for inspecting the types of residuals
def saved_residuals(f, *args, **kwargs):
  linearize = lambda *args: jax.linearize(f, *args, **kwargs)[1]
  jaxpr = jax.make_jaxpr(linearize)(*args).jaxpr
  return [v.aval for v in jaxpr.outvars]

# example function. no residuals should be stored, right? ...right?
@jax.checkpoint
def f(x):
  return jnp.einsum('ii->i', x)  # extract the diagonal

print(saved_residuals(f, jnp.ones((2, 2))))  # [ShapedArray(float32[2,2]), ShapedArray(float32[2,2])]
```

The reason residuals are being saved here is that to extract diagonals `jnp.einsum` basically creates an identity matrix, more or less evaluating `jnp.einsum('ij,ij->i', jnp.eye(N), x)`. And `jax.checkpoint` would save that identity matrix as a residual!

This toy example doesn't look bad, but actually in large transformer layers this could lead to saving constants which scaled like the square of the sequence length: `logits: f32[B, N, T, T] = einsum('BTNH,BTNH->BNTT', q_proj, k_proj) / sqrt(H)`. Moreover, these constants need not be materialized at all: XLA can often fuse their creation into consumers!

To fix this, we want `jax.checkpoint` to be able to see constant creation, and not just see operations from data dependence. Sounds like omnistaging, #3370! Given JAX's current tracing machinery, I couldn't think of a way to tweak the existing partial evaluation rule to get what we wanted. But it was straightforward just to adapt `jax.checkpoint` into an "initial-style primitive" like the control flow primitives. (See Autodidax for some explanation of that jargon.) That "initial-style" implementation meant we have to write a bit more explicit boilerplate-y rules, but they weren't so bad. And we gain a big upside!

Another limitation of the new implementation is that it can't handle `concrete=True`. I think that we can just get rid of that option; it seems no one is using it inside Google. (If some OSS people need it, we can figure out some kind of workaround when the need arises...)

This enables us to fix several other issues, so I've wanted to do it for a while:
1. it gives us the tools to properly handle effectful primitives in `jax.checkpoint` (which might mean just raising an error, rather than silently failing like today), since otherwise those lack data dependence and hence have the same 'constant folding' issue;
2. it lets `jax.checkpoint` avoid a [`scan` optimization](https://github.com/google/jax/blob/d703def10406097abd80853a0b64a991e752739d/jax/_src/lax/control_flow.py#L1625) which can hoist creation of constants outside a loop, defeating consumer fusions (whereas previously using `jax.checkpoint` would not help there).

I left the old implementation in place mainly so that we can land this in stages.

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -1,0 +1,308 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from functools import partial
+from typing import Callable, Optional
+import types
+
+import jax
+from jax import core
+from jax import linear_util as lu
+from jax.interpreters import ad
+from jax.interpreters import batching
+from jax.interpreters import partial_eval as pe
+from jax.interpreters import xla
+from jax.tree_util import tree_flatten, tree_unflatten
+from jax._src import ad_util
+from jax._src import source_info_util
+from jax._src.api_util import flatten_fun
+from jax._src.traceback_util import api_boundary
+from jax._src.util import (unzip2, wraps, split_list, partition_list, safe_map,
+                           safe_zip)
+
+# TODO(mattjj): before this can be the standard remat implementation, we must:
+#   [ ] fix up callers who use the 'concrete' option (now removed)
+#   [ ] implement remat-of-control-flow-primitives (passing through the policy)
+
+map = safe_map
+zip = safe_zip
+
+
+def _checkpoint_dots(prim, *_, **__) -> bool:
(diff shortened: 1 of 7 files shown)


zhangqiaorjc (approved):
Thanks Matt! Will be immensely helpful for our use case.

jekbradbury (approved):

shoyer:
Sounds good, I'll ping you about this when it comes up! We will likely be getting started on this in a month or two.

PR #8195, opened 2021-10-13
Remove unused variable from notebooks

bsaf:
Both the "Training a Simple Neural Network" notebooks include a `param_scale` variable that is not used.


jakevdp (approved):
Looks good, thanks!

PR #8197, opened 2021-10-13
Consolidate primitive and jit lowering paths.

hawkinsp:
Before this change, primitives have a special case dispatch path that attempts
to avoid building a jaxpr in the cache miss case. However, there's no good
reason for this: it makes the code more complicated, and we're not particularly
optimizing for fast cache misses anyway (we care mostly about cache hits).

Make the primitive lowering path trace a small function using the xla_callable
lowering path instead.

As a side effect, this makes the jit-of-pmap warning trigger when using, say, `scan(pmap(...))`. This is a desirable side effect: that case is indeed slow, we just weren't warning about it before.


mattjj (approved):
Awesome!

PR #8223, opened 2021-10-14
jnp.unique: improve efficiency & consolidate implementation

jakevdp:
We previously had two copies of the `jnp.unique` logic. This was useful because the ND version was previously about 50% slower than the 1D version. This PR merges the two implementations & reworks the logic to avoid extra array reshapes, which brings the speed of the two approaches (measured on CPU) to within a few percent.

Note that currently on the main branch, `axis=None` routes to the 1D implementation, and `axis=0` does the same computation via the ND implementation; the new implementation results in slightly slower timings for 1D, and ~30% faster timings for ND.

Some simple benchmarks:
```python
import jax.numpy as jnp
from jax import jit
from functools import partial
x = jnp.arange(1000) % 50

f_1D = partial(jnp.unique, axis=None)
f_ND = partial(jnp.unique, axis=0)
jit_f_1D = jit(partial(jnp.unique, axis=None, size=50))
jit_f_ND = jit(partial(jnp.unique, axis=0, size=50))
_ = jit_f_1D(x)
_ = jit_f_ND(x)

%timeit f_1D(x).block_until_ready()
%timeit f_ND(x).block_until_ready()
%timeit jit_f_1D(x).block_until_ready()
%timeit jit_f_ND(x).block_until_ready()
```
main branch:
```
1.04 ms ± 97.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.5 ms ± 48.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
122 µs ± 2.52 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
178 µs ± 2.76 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
this branch:
```
1.15 ms ± 5.05 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.08 ms ± 7.09 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
166 µs ± 2.02 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
163 µs ± 902 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
The 1D versions are slightly slower with this PR, but I think the win in terms of code maintainability is worth it.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5544,89 +5544,34 @@ def take_along_axis(arr, indices, axis: Optional[int]):
 
 
 ### SetOps
-
 @partial(jit, static_argnums=1)
-def _unique1d_sorted_mask(ar, optional_indices=False):
-  """
-  Helper function for unique which is jit-able
-  """
-  ar = ar.flatten()
-
-  if optional_indices:
-    aux, perm = lax.sort_key_val(ar, lax.iota(int, len(ar)))
-  else:
-    perm = np.empty(0, dtype=int)
-    aux = ar.sort()
-
-  mask = ones(aux.shape, dtype=bool_).at[1:].set(aux[1:] != aux[:-1])
-
-  return aux, mask, perm
-
-def _unique1d(ar, return_index=False, return_inverse=False,
-              return_counts=False, size=None, fill_value=None):
-  """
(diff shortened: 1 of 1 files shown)


froystig (approved):
Nice! Good to have a unique implementation of unique once again.

PR #8225, opened 2021-10-15
[sparse]: change bcoo pad values to use OOB indices

jakevdp:
At times, the BCOO representation needs to pad indices and data with null values. Currently, that padding is done with zeros:
```python
import jax.numpy as jnp
from jax.experimental import sparse

mat = jnp.array([[1, 0, 2, 0],
                 [3, 4, 5, 6]])

M = sparse.BCOO.fromdense(mat, n_batch=1)
print(M.data)
# [[1 2 0 0]
#  [3 4 5 6]]
print(M.indices[..., 0])
# [[0 2 0 0]
#  [0 1 2 3]]
```
Note that `M` has two specified elements in the first row, and four in the second row; the first row pads these extra indices and data with zeros. This approach works in most cases because repeated entries are implicitly added, and adding zero doesn't change the data

This zero-padding becomes problematic when we use the indices to extract values from another matrix, for example:
```python
new_mat = jnp.ones(M.shape)
print(new_mat)
# [[1. 1. 1. 1.]
#  [1. 1. 1. 1.]]
new_data = sparse.bcoo_extract(M.indices, jnp.ones(M.shape))
new_M = sparse.BCOO((new_data, M.indices), shape=M.shape)
print(new_M.todense())
# [[3. 0. 1. 0.]
#  [1. 1. 1. 1.]]
```
The algorithm has no way of knowing that these zeros are meant to be treated as padding, so three entries are extracted and added. This is particularly problematic because several jvp/transpose rules rely on `bcoo_extract`, potentially leading to difficult-to-debug wrong outputs.

This PR updates the representation to pad missing entries with out-of-range indices, here using the size of the axis for consistency:
```python
print(M.data) 
# [[1 2 0 0] 
#  [3 4 5 6]] 
print(M.indices[..., 0]) 
# [[0 2 4 4] 
#  [0 1 2 3]]  
```
With updates to the implementation of `bcoo_extract` and related functions to appropriately handle these out-of-bounds values, things work much more cleanly:
```python
new_M = sparse.BCOO((new_data, M.indices), shape=M.shape) 
print(new_M.todense())                                                                                                                    
[[1. 0. 1. 0.]
 [1. 1. 1. 1.]]
```

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -55,35 +55,23 @@ def _dedupe_bcoo(data, indices, shape):
   if indices.shape[:props.n_batch] != data.shape[:props.n_batch]:
     # TODO: handle broadcasted dimensions.
     raise NotImplementedError("dedupe_bcoo for broadcasted dimensions.")
-  f = _dedupe_bcoo_one
+  f = functools.partial(_dedupe_bcoo_one,
+      shape=shape[props.n_batch:props.n_batch + props.n_sparse])
   for _ in range(props.n_batch):
     f = vmap(f)
   return f(data, indices)
 
-def _dedupe_bcoo_one(data, indices):
-  assert indices.ndim == 2
-  assert data.shape[:1] == indices.shape[:1]
-
+def _dedupe_bcoo_one(data, indices, *, shape):
+  nse, = data.shape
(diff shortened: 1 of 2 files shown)


froystig (approved):
Nice! Seems easier to manually inspect/debug with this approach too.

PR #8234, opened 2021-10-15
Enable batcher and batched collective rules for tiled all gathers

### LINKED ISSUE #8221: Implement tiled=True for all_gather inside a vmap  (to address NotImplementedError exception)
My scenario:
   I use an all_gather with tiled=True inside a pmap.  I hit an error which is nearly impossible to understand because pmap JITs its code.  So I tried the recommended strategy of substituting "vmap" for "pmap".  I then hit the following exception:     raise NotImplementedError("Please open a feature request!")

This exception is raised because "tiled" is set to True for my all_gather.  The exception is raised by the second line of _all_gather_batched_collective() in parallel.py

apaszke:
Fixes #8221.


mattjj (approved):
Nice!

PR #8236, opened 2021-10-15
jnp.bincount: fix corner cases & improve tests

jakevdp:
- fix the case of `jnp.bincount(jnp.array([-1]))` which previously returned an empty array, but according to the documentation, should return an array of length 1
- clarify the behavior when `minlength > length`. Now, when `length` is supplied, `minlength` is explicitly ignored.
- improve test coverage: previously, tests did not assess correctness of results when `length` was supplied


mattjj (approved):
Nice!

PR #8237, opened 2021-10-15
Change default kind for jnp.argsort to `stable`. Warn if anything oth…

hawkinsp:
…er than `stable` is passed.


jakevdp (approved):
Looks good! The title worried me at first because I've written a bunch of code that relies on argsort being stable! Turns out this just makes that more explicit 😁

PR #8240, opened 2021-10-15
Fix inaccurate type annotations

jakevdp:
Each `dtype` has two associated python types, the type class (e.g. `np.float64`) and the dtype class (e.g. `np.dtype('float64')`). The annotations used in `dtypes.py` are currently assuming the wrong object:
```python
>>> isinstance(np.float64, np.dtype)
False
>>> type(np.float64)
type
>>> isinstance(np.dtype(np.float64), np.dtype)
True
```

### DIFF
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -40,10 +40,10 @@ _bfloat16_dtype: np.dtype = np.dtype(bfloat16)
 
 # Default types.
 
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):
Yes, this is right, they are subclasses of `np.generic` not dtypes.

An alternative would be to make them true dtypes. Up to you.

PR #8251, opened 2021-10-17
Added AdamW, AdaBelief Optimizers and tests

TimSYQQX:
Added two variants of Adam optimizers, AdamW and AdaBelief, I've tested the optimizers on my test bench using MLP and resnet18 on MNIST,  and mini-ImageNet. They seem to be working as expected. 

Thanks.

### DIFF
--- a/jax/experimental/optimizers.py
+++ b/jax/experimental/optimizers.py
@@ -500,6 +500,80 @@ def sm3(step_size, momentum=0.9):
 
   return init, update, get_params
 
+@optimizer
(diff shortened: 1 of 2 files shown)


levskaya (commented):
There's a bug in your AdamW implementation.  At a higher-level: I don't think these belong in JAX's built-in optimizer examples.  The ones that are there are "grandfathered" in as elementary pedagogical examples of how to write optimizers in JAX, but we don't want to add any more and would prefer all optimization-focused work be done between the Optax and JaxOpt optimizer libraries.  However, thank you for your efforts!  We just want to unify all work on these topics into standard places in the ecosystem.

PR #8288, opened 2021-10-19
stats.multivariate_normal: support broadcasted inputs

jakevdp:
Fixes #8282

Note that with this change, `jax.scipy.stats.multivariate_normal` handles a much wider set of inputs than `scipy.stats.multivariate_normal` – for example, JAX's version now handles batched covariance matrices and batched means. ~This is untested for now, because numpy's ground-truth is not available for such cases.~ This is tested using `np.vectorize` along with scipy's `multivariate_normal`.

### DIFF
--- a/jax/_src/scipy/stats/multivariate_normal.py
+++ b/jax/_src/scipy/stats/multivariate_normal.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from functools import partial
 
 import numpy as np
 import scipy.stats as osp_stats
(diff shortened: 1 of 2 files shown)


mattjj:
> This is untested for now, because numpy's ground-truth is not available for such cases.

Could we write some test cases with vmap? That is, assuming vmap is correct, it seems we have two ways to express things like multivariate_normal batched over means, covariances, or both: one using higher-rank inputs, the other using vmap.

jakevdp:
Added a broadcasting test - thanks!

mattjj (approved):
Nice!

PR #8317, opened 2021-10-20
dynamic_slice: ensure start_indices dtypes match

jakevdp:
The types of the `start_indices` arguments to `lax.dynamic_slice` must match, but in the presence of dynamic shapes this may not be preserved. This bug came up in #8180.

I think it can only arise when `dtypes.int_` is different than `dtypes.result_type(int)`. Currently that is true by happenstance, but since we will be changing that soon we should not rely on that fact.


hawkinsp (approved):
Seems fine if it doesn't break any tests (internal or external).

PR #8346, opened 2021-10-23
Add truncated_normal initializer

anuragarnab:
Adds a truncated_normal initializer that does not use variance scaling depending on the input dimensions.

### DIFF
--- a/jax/_src/nn/initializers.py
+++ b/jax/_src/nn/initializers.py
@@ -44,6 +44,12 @@ def normal(stddev=1e-2, dtype=jnp.float_):
     return random.normal(key, shape, dtype) * stddev
   return init
 
+def truncated_normal(stddev=1e-2, lower=-2, upper=2, dtype=jnp.float_):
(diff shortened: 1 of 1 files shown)


jax/_src/nn/initializers.py
```diff
@@ -44,6 +44,13 @@ def init(key, shape, dtype=dtype):
     return random.normal(key, shape, dtype) * stddev
   return init
 
+def truncated_normal(stddev=1e-2, lower=-2, upper=2, dtype=jnp.float_):
```
jekbradbury, line 47: I think you can factor https://github.com/google/jax/pull/8346/files#diff-dfe359c36c6d226fc1df3447c75b7d39f409e4d145221a2d7597147f45c51b1fR126-R133 into a helper and then call it here? That way this will also work for complex dtypes.

jekbradbury (changes requested):

hawkinsp:
Closing as stale, `jax.nn.initializers.truncated_normal` is a thing these days.

PR #3217, opened 2020-05-27
Prefer using broadcast_in_dim/squeeze instead of reshape

shoyer:
`reshape()` is quite powerful, but does not necessarily preserve a notion of axis identity (particularly for axes of length 1). This is problematic for transformation rules that need to preserve a notion of axis identity, such as for masking and a new transformation rule I'm exploring for unraveling pytrees.

This PR adds a new `lax.squeeze` primitive, and rewrite functions from `lax_numpy.py` in terms of `lax.squeeze`/`lax.broadcast_in_dim`, whenever possible, to ensure a well-defined mapping between input and output axes. In particular, it updates: `matmul`, various `stack` functions, the `array` constructor, broadcasting arithmetic, array indexing, `squeeze` and reductions with `keepdims=True`.

I also implemented support for multiple axes in `jnp.expand_dims` (added in NumPy 1.18), since it was convenient for some of these other functions.

I considered trying to write a masking rule for broadcast_in_dim as well, but it was trickier than I expected and @JuliusKunze has probably already thought about it :)

PR #8381, opened 2021-10-27
Support reshape/concatenate/broadcast on PRNGKeyArrays.

LenaMartens:
`PRNGKeyArrays` (which are currently behind the `enable_custom_prng` flag) do not support any `jnp` operations. This makes sense for operations which act on the value of a key (we shouldn't support operations like `key0 + key1`), but we want to keep supporting operations which are agnostic about the value of a key  (eg. a `broadcast`). 

This PR adds a way to overload `jnp.reshape`, `jnp.concatenate` and `jnp.broadcast`, then overloads them for `PRNGKeyArrays`. The claim is that overloading these three operations is sufficient to support most `jnp` functions users care about.

Given we support  `reshape`, `concatenate` and `broadcast`, there are some `jnp` functions we get for free (eg. `tile`), but there are some which are implemented in terms of `lax` primitives like `lax.expand_dims` instead. The motivation for this is described in https://github.com/google/jax/pull/3217 (preserve the notion of axis identity for reshapes which add/remove singleton dimensions)
 
I see two options to support these functions:
- overload `expand_dims` and `squeeze` for `PRNGKeyArrays`
- rewrite `jnp.expand_dims/squeeze` to use `reshape`/`broadcast` instead

Happy to discuss what operations should be supported here, and how best to go about it :)

----
For completeness, here's an enumeration of `jnp` functions which we _could_ support for PRNGKeyArrays, and what needs to be overloaded to support them.  

I see 5 categories of `jnp` functions, in terms of the base ops they are implemented with:
1. operations which are implemented in terms of reshape/concatenate/broadcast, and we get for free (eg. `append`)
2. **operations which use expand_dims/squeeze, but could be implemented with reshapes (eg. `stack`)**
3. operations which use indexing (which we probably don't want to support?) (eg. `delete`)
4. operations which create/cast to arrays (should we support `jnp.array(PRNGKeyArray)`?)
5. transposes (shuffling of elements)

(Incomplete) list of `jnp` operations which we could support for PRNGKeyArrays and their categories (see above):
jnp op | implemented with.. | category
-- | -- | --
atleast_1d | reshape+array | 4
atleast_2d | expand_dims+array | 4
append(arr, values[, axis]) | concatenate | 1
array(object[, dtype, copy, order, ndmin, …]) | array | 4
array_split(ary, indices_or_sections[, axis]) | split=slice | 3
asarray(a[, dtype, order]) | array | 4
broadcast_arrays(*args) | broadcast_to | 1
broadcast_to(arr, shape) | broadcast | 1
choose(a, choices[, out, mode]) | broadcast+index | 3
compress(condition, a[, axis, out]) | index | 3
concatenate(arrays[, axis]) | concatenate | 1
delete(arr, obj[, axis]) | ravel+index | 3
empty_like(a[, dtype, shape]) | zeros_like | 4
expand_dims(a, axis) | expand_dims (could be broadcast/reshape?) | 2
flip(m[, axis]) | lax.rev | 5
fliplr(m) | flip=rev | 5
flipud(m) | flip=rev | 5
hsplit(ary, indices_or_sections) | split=slice | 3
hstack(tup) | concatenate+atleast_1dim | 2
insert(arr, obj, values[, axis]) | indexing | 3
moveaxis(a, source, destination) | lax.transpose | 5
ones_like(a[, dtype, shape]) | lax.full_like | 4
ravel(a[, order]) | reshape | 1
repeat(a, repeats[, axis, total_repeat_length]) | take | 3
reshape(a, newshape[, order]) | reshape | 1
resize(a, new_shape) | reshape+tile+indexing | 3
roll(a, shift[, axis]) | indexing | 3
rollaxis(a, axis[, start]) | move_axis | 5
row_stack(tup) | vstack | 2
split(ary, indices_or_sections[, axis]) | lax.slice | 3
squeeze(a[, axis]) | lax.squeeze | 2
stack(arrays[, axis, out]) | concatenate+expand_dims | 2
swapaxes(a, axis1, axis2) | lax.transpose | 5
take(a, indices[, axis, out, mode]) | index | 3
take_along_axis(arr, indices, axis) | index | 3
transpose(a[, axes]) | lax.transpose | 5
tile | broadcast_to+reshape | 1
vsplit(ary, indices_or_sections) | split | 3
vstack(tup) | concatenate+expand_dims | 2
zeros_like(a[, dtype, shape]) | lax.full_like | 4

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -30,7 +30,7 @@ import collections
 from functools import partial
 import operator
 import types
-from typing import Sequence, FrozenSet, Optional, Tuple, Union
+from typing import Sequence, FrozenSet, Optional, Tuple, Union, Set, Type, Callable
 from textwrap import dedent as _dedent
 import warnings
 
@@ -552,6 +552,12 @@ def _result_dtype(op, *args):
 def _arraylike(x):
   return (isinstance(x, np.ndarray) or isinstance(x, ndarray) or
           hasattr(x, '__jax_array__') or isscalar(x))
+
+
+def _stackable(*args):
+  return _all(type(arg) in stackables for arg in args)
+stackables: Set[Type] = set()
+_register_stackable: Callable[[Type], None] = stackables.add
 
 def _check_arraylike(fun_name, *args):
   """Check if all args fit JAX's definition of arraylike."""
@@ -1718,7 +1724,7 @@ def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
 
 @_wraps(np.reshape, lax_description=_ARRAY_VIEW_DOC)
 def reshape(a, newshape, order="C"):
-  _check_arraylike("reshape", a)
+  _stackable(a) or _check_arraylike("reshape", a)
   try:
     return a.reshape(newshape, order=order)  # forward to method for ndarrays
   except AttributeError:
@@ -1761,7 +1767,7 @@ def _transpose(a, *args):
 @_wraps(np.ravel, lax_description=_ARRAY_VIEW_DOC)
 @partial(jit, static_argnames=('order',), inline=True)
 def ravel(a, order="C"):
-  _check_arraylike("ravel", a)
+  _stackable(a) or _check_arraylike("ravel", a)
   if order == "K":
     raise NotImplementedError("Ravel not implemented for order='K'.")
   return reshape(a, (size(a),), order)
@@ -2224,6 +2230,8 @@ def broadcast_arrays(*args):
 The JAX version does not necessarily return a view of the input.
 """)
 def broadcast_to(arr, shape):
+  if hasattr(arr, "broadcast_to"):
+    return arr.broadcast_to(shape)
   arr = arr if isinstance(arr, ndarray) else array(arr)
   shape = (shape,) if ndim(shape) == 0 else shape
   shape = canonicalize_shape(shape)  # check that shape is concrete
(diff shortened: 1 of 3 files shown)


jax/numpy/__init__.py
```diff
@@ -304,6 +304,7 @@
     ravel_multi_index as ravel_multi_index,
     real as real,
     reciprocal as reciprocal,
+    _register_stackable as _register_stackable,
```
mattjj, line 307: My only nit with this PR is that maybe we could consider not including `_register_stackable` in this __init__.py file, and instead importing it differently where it's needed (e.g. instead of `jnp._register_stackable(PRNGKeyArray)` maybe we'd have to do `from jax._src.numpy.lax_numpy import _register_stackable` instead). That'd make extra clear it's an internal API.

mattjj (approved):
Great work! The code is nice and clean. But even better is the PR message! That taxonomy is a fantastic leap forward for the broader effort here.

froystig (approved):
@mattjj said it well. Really nice work on this, and thanks for laying things out so clearly in the PR description!

shoyer:
Just a note: if `PRNGKeyArray` was in NumPy, we could imagine implementing it as an array with a new data type, i.e., with data stored in a vector of two elements like `np.dtype('u4,u4')`.

It may not be worth building in a new high-level dtype abstraction into JAX to handle this at the primitive level, but for what it's worth this is exactly the sort of thing we would need to properly support strings.

jax/numpy/__init__.py
LenaMartens: yep, that makes a lot of sense!

LenaMartens:
Thanks for the review! (and sorry for the delay)

I think there is still an open question about which set of operations to eventually support, but we can do this in a follow-up.

PR #8395, opened 2021-10-29
Add context manager implementation for name stack handling

sharadmv:
TLDR; This PR adds a parallel context manager implementation for name stacks, gated by the `jax_experimental_name_stack` config flag. `named_call` uses this new context manager rather than a call primitive. Another public API to access and manipulate the name stack will be in a forthcoming PR along with more detailed documentation.

# Approach

The name stack is a data structure (a stack) that keeps track of both name scopes and transformations. Both scopes and transformations are pushed and popped off the stack. A stack that looks like `[foo, vmap, bar]` will produce the following OP metadata: `'foo/vmap(bar)'`, indicating that the bar stack is transformed by a `vmap` transformation. The `jvp` and `transpose` transformations are also example transformations that can be on the name stack.

## Part 1: Annotations in jaxprs

The name stack mechanism needs to somehow record information into jaxprs, to ensure it's preserved through transformations that stage out Python code into a jaxpr and reinterpret that jaxpr.

How do we use a context manager to annotate jaxprs? Let’s say that each jaxpr equation has a new `name_stack` field that records the desired XLA HLO annotations we want when the jaxpr is lowered. How do we populate the `name_stack` field with a context manager?

First let’s implement a simple context manager API:
```python
@contextmanager
def extend_name_stack(name):
  # appends `name` to the active name stack

def current_name_stack(name):
  # returns the current name stack
```

A simple use case for this function is annotating primitives within a function.

```python
def f(x):
  with extend_name_stack('bar'):
    return x * 2.
```

We hope that when this function is staged out to a jaxpr we get the following annotations (represented in brackets after the equation).
```
{ lambda a:float64[] .
let b:float64[] = mul a 2.0 # [bar]
in ( b ) }
```

We can modify the trace that builds jaxprs to call `current_name_stack()` when it’s creating an equation to populate the `name_stack` field.

This is a simple strategy for propagating information from a context manager into the staged out jaxpr. In practice, we utilize the `source_info` tracking already present in JAX and augment the `SourceInfo` with a name stack field.

## Part 2: Final style transformations

What’s the desired behavior when functions with `extend_name_stack`s in them are transformed? When a particular primitive is reinterpreted under a transformation we’d like to modify the name stack to include information about the transformation.

Take the following vmap example:
```python
@vmap
def f(x):
  with append_name_stack('bar'):
    return x * 2.
```

We’d like the following annotated jaxpr:
```
{ lambda a:float64[2] .
let b:float64[2] = broadcast [ axes=[0] shape=[2] ] 2.0 # [vmap(bar))]
    c:float64[2] = mul a b # [vmap(bar)]
in ( c ) }
```

See how we have `vmap(bar)` annotations for both the `broadcast` and the `mul`. How do we modify our context manager for this use case? Let’s support adding “transformations” to the name stack.
```python
@contextmanager
def transform_name_stack(name):
  # appends `name` to the active name stack as a transformation
```

In our vmap implementation we can do something like this:
```python
def vmap(f):
  def wrapped(*args, **kwargs):
    with transform_name_stack('vmap'):
      out = …
    …
  return wrapped
```

When we vmap a function we’ll have a transform on the name stack which we can then postprocess to produce the `vmap(bar)` annotation we desire. We can do the same for all final-style transformations, i.e. the transformations implemented as Traces vs jaxpr interpreters (e.g. jvp).

## Part 3: Initial style transformations

Initial style transformations like grad or the control flow primitives first trace a function out into a jaxpr, then reinterpret it in some interesting way. `grad`, for example, traces the partially evaluated `jvp` of a function into a jaxpr and interprets the resulting jaxpr backwards to obtain gradients.

This brings up a potential issue with adding annotations using a dynamically scoped name stack. Recall that we add annotations to jaxpr eqns via the `DynamicJaxprTrace` looking up the currently active name stack. If we’re staging out a jaxpr while inside of another transformation (which is the case in initial style transformations), we’ll include that transform’s annotation in the stack. For example, consider a no-op initial style transformation that stages out and immediately evaluates a jaxpr.
```python
def reinterpret(f):
  def wrapped(*args):
    jaxpr = make_jaxpr(f)(*args)
    return eval_jaxpr(jaxpr, *args)
  return wrapped
```

We could imagine that if there's a `foo` name stack active outside of the call to `reinterpret`, we'll bake `foo` into the name stacks in the `jaxpr` and then add yet another `foo` when calling `eval_jaxpr`. Furthermore, what do we do with the name stacks backed into the jaxpr themselves? We have to be careful about how we treat name stacks in this setting.

#### `eval_jaxpr`

How does eval_jaxpr handle annotations? We have a couple choices. We could have it just inherit the current active name stack, i.e.:
```python
def eval_jaxpr(jaxpr, *args):
  …
  for eqn in jaxpr.eqns:
    …
    ans = eval_eqn(eqn, …)
  …
```
However, if the equations we’re evaluating have name stacks, they will be discarded and we will lose information. We need to somehow preserve the name stacks in the jaxpr we’re evaluating. We could instead re-use that information, throwing away the current name stack info using a `set_name_stack` context manager.
```python
def eval_jaxpr(jaxpr, *args):
  …
  for eqn in jaxpr.eqns:
    …
    with set_name_stack(eqn.name_stack): # overrides current name stack
      ans = eval_eqn(eqn, …)
  …
```
This is also incorrect, however. We want to use the current active name stack as well since we might be transforming eval_jaxpr from the outside. We thus need to prepend the current active name stack to the one from the equation.
```python
def eval_jaxpr(jaxpr, *args):
  …
  for eqn in jaxpr.eqns:
    …
    with set_name_stack(get_current_stack() + eqn.name_stack):       
      ans = eval_eqn(eqn, …)
    …
```

#### `make_jaxpr`
Now consider the following example.
```python
@vmap
@reinterpret
def f(x):
  with append_name_stack('bar'):
    return x * 2.
f(jnp.ones(2))
```

In principle, we should obtain the same annotations as before (`vmap(bar)`). However, in our currently implementation we first obtain a jaxpr in reinterpret. Since that jaxpr is using dynamically scoped annotations, it will annotate the mul with a `vmap(bar)` even though we’re not actually doing a vmap – we’re just staging out an unbatched function. The actual vmapping happens when we call `eval_jaxpr`. If we’re staging out the entirety of `f` into jaxpr, the vmap will still be on the stack when we call `eval_jaxpr` and we will get a doubly-annotated `vmap(vmap(bar))`, an incorrect annotation. This is a huge problem!

This pattern of transforming an initial style transformation occurs everywhere (`vmap(grad)` is an example of this). How do we modify our implementation to get the right annotations?

First, while building the jaxpr, we need to start with an empty active name stack rather than the current active name stack. We can accomplish this with another context manager that essentially clears the active name stack:
```python
@contextmanager
def reset_name_stack():
  # sets the current name stack to be empty
```

Right before we build a jaxpr, we make sure to clear the active name stack.
```python
def make_jaxpr(f):
  def wrapped(*args):
    with reset_name_stack():
      jaxpr = …
    return jaxpr
  return wrapped
```

The jaxpr we now stage from the example will no longer have a `vmap(bar)` annotation (it will just have a bar annotation). When we eval_jaxpr, the vmap on the name stack will be prepended to the bar to produce the appropriate vmap(bar).

This strategy of disabling the name stack when staging out to a jaxpr and prepending the active name stack to the one from the jaxpr equation generalizes to other instances of staging out and reinterpreting, for example with control flow primitives. We need to make sure that when tracing out a body function to a jaxpr, for example, it is staged out without the active name stack. When the body is reinterpreted in the context of a loop, eval_jaxpr will add the active name stack back in.

### Part 4: `partial_eval`
Partial evaluation is a particularly tricky case because it both evaluates and stages at the same time. We thus need to be careful about how we annotate both the evaluated primitives and the staged out ones. We can use the same principles from before though: disable the name stack when we stage out and enable when we are evaluating. 

How do we emulate this when we’re in a final style transformation though? In an initial style transformation, the distinction between staging and evaluation is very clear and there are obvious places to disable and enable the name stack. In a partial eval trace, when we evaluate a primitive there will be names on the stack from outside of the partial eval trace and from inside. When we stage out, we want to annotate the jaxpr equation with the names from the “inside”, not the “outside”. Here’s an example:
```python
@vmap
@partial_eval
def f(x, y):
  with append_name_stack('foo'):
    x = x + 1
  with append_name_stack('bar'):
    y = y + 1
  return x, y
```

Let’s say this partial_eval transformation treats x as known and y as unknown. We thus want to evaluate `x + 1` and stage `y + 1` into a jaxpr. When we evaluate `x + 1` we’d like it to be annotated with `vmap(foo)` but when we stage out `y + 1` we want it to be staged out with just `bar`, not `vmap(bar)`. 

When we’re inside of the partial eval trace, we need to distinguish between the names that are added outside of the partial eval (`vmap` in this case) and the ones inside (`foo` and `bar`). We can accomplish this by recording the name stack before the partial eval trace and removing it when staging out a jaxpr eqn.

```python
def partial_eval_flat(f, pvals_in):                                                                                                                     
                                                                                                           
  outer_stack = current_name_stack()                                                                                                                                                   
  with new_main(PartialEvalTrace, outer_stack) as main:                                                                                                                                                                                                                           
    outs = …                                                                                                                                                         
  return jaxpr, pvals_out, consts
```

Any names added to the stack while doing the partial eval will be the inside names. Inside of the `process_primitive` for partial eval, we can do the following:
```python
def process_primitive(self, primitive, tracers, params):                                                                                                                                       
  if all(t.pval.is_known for t in tracers):      
      return bind(primitive, *map(full_lower, tracers), **params)           
  …
  # grab the portion of the name stack that extends beyond the outer name stack
  name_stack = current_name_stack()[len(self.outer_name_stack):]                                                                                               
  eqn = JaxprEqnRecipe(..., name_stack)
  …                                                                                                                  
```
If we evaluate inside of partial eval, we use the full stack (outer + inner). If we stage out, then we just use the inner names.

# Implementation details

The goal of this implementation is to preserve the previous mechanism and enable the context-manager behind a `config` flag. To accomplish this, we have two parallel mechanisms keeping track of the name stack and decide to use one or the other when lowering.

* To add a name scope to the name stack, we use `source_info_util.extend_name_stack`, a context manager that pushes a name onto the stack and pops it off after the context manager exits. There is also `source_info_util.transform_name_stack` that pushes a transformation onto the stack.
* When we lower a jaxpr to HLO (`xla_subcomp`), we, by default, use the current mechanism for adding OP metadata to the HLO. If `jax_experimental_name_stack` is enabled, however, we use the name stack from the `SourceInfo` (i.e the jaxpr equation). We make sure to thread an equation's name stack into any sub contexts use to generate HLO for say call primitive jaxprs.
* To inspect and debug name stacks, you can now do `jaxpr.pretty_print(name_stack=True)` to have name stacks print for each equation.

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -97,6 +97,11 @@ class PartialVal(tuple):
 
 
 class JaxprTrace(Trace):
+
+  def __init__(self, *args, name_stack: source_info_util.NameStack):
+    super().__init__(*args)
+    self.name_stack = name_stack
+
   def pure(self, val) -> 'JaxprTracer':
     return self.new_const(val)
 
@@ -163,7 +168,8 @@ class JaxprTrace(Trace):
     tracers = map(self.instantiate_const, tracers)
     avals = [t.aval for t in tracers]
     out_aval = primitive.abstract_eval(*avals, **params)
-    source = source_info_util.current()
+    name_stack = self._current_truncated_name_stack()
+    source = source_info_util.current().replace(name_stack=name_stack)
     if primitive.multiple_results:
       out_tracers = [JaxprTracer(self, PartialVal.unknown(aval), None)
                      for aval in out_aval]
@@ -213,11 +219,11 @@ class JaxprTrace(Trace):
     # The outputs of the staged-out call are Tracers with the new eqn as recipe.
     out_tracers = [JaxprTracer(self, PartialVal.unknown(a), None)
                    for a in out_avals]
+    name_stack = self._current_truncated_name_stack()
+    source = source_info_util.current().replace(name_stack=name_stack)
     eqn = new_eqn_recipe((*const_tracers, *env_tracers, *unknown_arg_tracers),
-                         out_tracers, primitive, staged_params,
-                         source_info_util.current())
+                         out_tracers, primitive, staged_params, source)
     for t in out_tracers: t.recipe = eqn
-
     return merge_lists(out_knowns, out_tracers, out_consts)
 
   def process_map(self, primitive, f: lu.WrappedFun, tracers, params):
@@ -305,8 +311,9 @@ class JaxprTrace(Trace):
       update_params = call_param_updaters.get(primitive) or (lambda p, _, __: p)
       new_params = update_params(params, [], len(in_tracers))
       new_params = dict(new_params, call_jaxpr=convert_constvars_jaxpr(jaxpr))
-      eqn = new_eqn_recipe(in_tracers, out_tracers, primitive, new_params,
-                           source_info_util.current())
+      name_stack = self._current_truncated_name_stack()
+      source = source_info_util.current().replace(name_stack=name_stack)
+      eqn = new_eqn_recipe(in_tracers, out_tracers, primitive, new_params, source)
       for t in out_tracers: t.recipe = eqn
       return merge_lists(out_knowns, out_tracers, out_consts)
 
@@ -342,8 +349,10 @@ class JaxprTrace(Trace):
                    for d, a in zip(staged_out_axes, out_avals_mapped)]
       out_tracers = [JaxprTracer(trace, PartialVal.unknown(a), None)
                      for a in out_avals]
+      name_stack = self._current_truncated_name_stack()
+      source = source_info_util.current().replace(name_stack=name_stack)
       eqn = new_eqn_recipe((*const_tracers, *env_tracers), out_tracers,
-                           primitive, staged_params, source_info_util.current())
+                           primitive, staged_params, source)
       for t in out_tracers: t.recipe = eqn
       return merge_lists(out_knowns, out_tracers, out_consts)
 
@@ -354,6 +363,9 @@ class JaxprTrace(Trace):
     out_axes_unknown: Optional[list] = None
 
     return out, (todo, out_axes_transform)
+
+  def _current_truncated_name_stack(self):
+    return source_info_util.current_name_stack()[len(self.name_stack):]
 
   def partial_eval(self, f: lu.WrappedFun, pvals: Sequence[PartialVal],
                    app: Callable[[lu.WrappedFun, Tuple[core.Value, ...]], Tuple[core.Value]],
@@ -394,11 +406,13 @@ class JaxprTrace(Trace):
       converted_jaxpr = convert_envvars_to_constvars(jaxpr, len(env))
       return converted_jaxpr, (*consts, *env)
 
+    name_stack = self._current_truncated_name_stack()
+    source = source_info_util.current().replace(name_stack=name_stack)
     eqn = new_eqn_recipe(in_tracers, out_tracers, prim.initial_style,
                          dict(fun_jaxpr=closed_jaxpr,
                               jvp_jaxpr_thunk=jvp_jaxpr_thunk,
                               num_consts=len(consts) + len(env)),
-                         source_info_util.current())
+                         source)
     for t in out_tracers: t.recipe = eqn
     return out_tracers
 
@@ -434,12 +448,14 @@ class JaxprTrace(Trace):
       converted_jaxpr = convert_envvars_to_constvars(jaxpr, len(env))
       return converted_jaxpr, (*consts, *env)
 
+    name_stack = self._current_truncated_name_stack()
+    source = source_info_util.current().replace(name_stack=name_stack)
     eqn = new_eqn_recipe(in_tracers, out_tracers, prim.initial_style,
                          dict(fun_jaxpr=closed_jaxpr,
                               fwd_jaxpr_thunk=fwd_jaxpr_thunk,
                               num_consts=len(consts) + len(env),
                               bwd=bwd, out_trees=out_trees),
-                         source_info_util.current())
+                         source)
     for t in out_tracers: t.recipe = eqn
     return out_tracers
 
@@ -551,7 +567,8 @@ def trace_to_jaxpr(
     returned jaxpr takes as inputs the known residual values followed by values
     of the originally unknown inputs.
   """
-  with core.new_main(JaxprTrace) as main:
+  current_name_stack = source_info_util.current_name_stack()
+  with core.new_main(JaxprTrace, name_stack=current_name_stack) as main:
     fun = trace_to_subjaxpr(fun, main, instantiate)
     jaxpr, (out_pvals, consts, env) = fun.call_wrapped(pvals)
     assert not env
@@ -565,7 +582,7 @@ def trace_to_subjaxpr_nounits(
     main: core.MainTrace, instantiate: Union[bool, Sequence[bool]],
     in_pvals: Sequence[PartialVal]):
   assert all([isinstance(pv, PartialVal) for pv in in_pvals]), in_pvals
-  trace = JaxprTrace(main, core.cur_sublevel())
+  trace = main.with_cur_sublevel()
   in_knowns  = [pval.is_known()     for pval in in_pvals]
   in_consts  = [pval.get_known()    for pval in in_pvals if     pval.is_known()]
   in_tracers = [trace.new_arg(pval) for pval in in_pvals if not pval.is_known()]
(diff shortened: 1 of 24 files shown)


tomhennigan:
This is really nice! One thought is that we are very keen in Haiku to turn on names by default (this will help a lot of our users) but it would probably not be appropriate for us (as a library) to force (or set) the new global config flag to be on. I see a couple of options:

1. If the flag is going to be flipped on by default soon (O(weeks) after submitting), then we'll be patient and just wait.
2. If the rollout will be slower than this, can we have an option in `named_call` (or similar) to opt into the new behaviour at the call site (perhaps as well as globally?). This would allow Haiku (and other libraries) to opt our internal usages into the new behaviour while letting users control the behaviour of the rest of their program.

jax/interpreters/partial_eval.py
```diff
@@ -200,8 +206,8 @@ def new_out_axes_thunk():
       ctx = contextlib.suppress()  # This is a no-op
       app = partial(primitive.bind, **params)
     with ctx:
-      jaxpr, out_pvals, consts, env_tracers = self.partial_eval(
-          f, in_pvals, app, instantiate=False)
+      jaxpr, out_pvals, consts, env_tracers = self.partial_eval(f, in_pvals,
+          app, False)
```
mattjj, line 210: put all args on same line, keep `instantiate=False`?

jax/interpreters/partial_eval.py
```diff
@@ -271,7 +278,7 @@ def post_process_call(self, primitive, out_tracers, params):
     def todo(x):
       n = len(jaxpr.outvars)
       out_pv_consts, consts = x[:n], x[n:]
-      trace = JaxprTrace(main, core.cur_sublevel())
+      trace = JaxprTrace(main, core.cur_sublevel(), **main.payload)
```
mattjj, line 281: Replace with `main.with_cur_sublevel()`?

jax/interpreters/partial_eval.py
```diff
@@ -523,7 +539,7 @@ def fun(ki, ui):  # ki will be a known input in this example
 def trace_to_subjaxpr(main: core.MainTrace, instantiate: Union[bool, Sequence[bool]],
                       pvals: Sequence[PartialVal]):
   assert all([isinstance(pv, PartialVal) for pv in pvals]), pvals
-  trace = JaxprTrace(main, core.cur_sublevel())
+  trace = JaxprTrace(main, core.cur_sublevel(), **main.payload)
```
mattjj, line 542: Replace with `main.with_cur_sublevel()`?

jax/interpreters/partial_eval.py
```diff
@@ -510,7 +525,8 @@ def fun(ki, ui):  # ki will be a known input in this example
     out_pvals = [PartialVal.unknown(ConcreteArray(6)), PartialVal.unknown(ShapedArray)]
     consts = [3, 6]  # values for `ka` and `kb` constvars
   """
-  with core.new_main(JaxprTrace) as main:
+  current_name_stack = source_info_util.current_name_stack()
+  with core.new_main(JaxprTrace, name_stack=current_name_stack) as main:
```
mattjj, line 529: Why is this attached as a payload on the MainTrace here, but done via context manager in other places?

mattjj (commented):
Overall LGTM but I admit I've lost some context on this (my fault) so I feel like I'm forgetting some things. Let's discuss in next meeting to put the final touches?

tomhennigan:
Hi all, any ETA on this change landing? We would love to take advantage of it in Haiku.

sharadmv:
Sorry for the delay, I've rebased it and will soon investigate the internal failures that blocked it previously.

jax/interpreters/partial_eval.py
sharadmv: Done.

jax/interpreters/partial_eval.py
sharadmv: Done.

jax/interpreters/partial_eval.py
sharadmv: Other main traces don't need to "remember" the name stack when they were created. The reason we need this functionality for `partial_eval` is because it does both evaluation and staging, so we need to remember the name stack before the trace started so we stage out the jaxpr, we remove the part of the name stack we use in evaluation.

I suppose this could be done via a context manager / store that maps `MainTrace -> NameStack` but since the payload mechanism already exists, I figure that would do. WDYT?

mattjj (approved):

PR #8403, opened 2021-10-29
Generate a better error message if [pv]map receives no array arguments

shoyer:
The current error message for `jax.vmap(lambda x: x)({})` is:
`ValueError: vmap must have at least one non-None value in in_axes`

With this PR, it becomes:
`ValueError: vmap wrapped function must be passed at least one argument
containing an array, got empty *args=({},) and **kwargs={}`


jakevdp (approved):
Looks good! Sorry for the slow review

PR #499, opened 2019-03-11
add has_aux to jvp (fixes #497)

mattjj:
TODO:
- [ ] plumb `has_aux` through `jacfwd` and `jacrev` too

PR #8420, opened 2021-11-01
Support has_aux for jacfwd and jacrev

### LINKED ISSUE #497: has_aux for forward-mode differentiation functions
Related to PR https://github.com/google/jax/pull/484

We should have corresponding `has_aux` implementation of passing side-information for the forward-mode functions like `jvp`, `jacfwd` etc.

My use case is under a scenario that forward-mode is much more efficient than the reverse-mode.

Huizerd:
The project I'm working on would benefit from easy switching between backward-mode and forward-mode AD, so I implemented the convenience method `value_and_grad_fwd` based on existing code (`value_and_grad`, `jacfwd`). To allow auxiliary data, I built on #499, which adds `has_aux` to `jvp`, and adapted it to the current codebase. Note that this is my first PR for JAX, so I might've misunderstood some things:

- Not sure about the `out_axes` value on line 1072 of `api.py`
- Not sure about the added tests, there probably should be more

Also, if you think `value_and_grad_fwd` is unnecessary because the same can be achieved easily (and as efficiently as for `value_and_grad`) with existing code, I'd love to learn more :)

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -964,7 +964,8 @@ def value_and_grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
     tuple). If ``argnums`` is an integer then the gradient has the same shape
     and type as the positional argument indicated by that integer. If argnums is
     a sequence of integers, the gradient is a tuple of values with the same
-    shapes and types as the corresponding arguments.
+    shapes and types as the corresponding arguments. If ``has_aux`` is True
+    then a tuple of ((value, auxiliary_data), gradient) is returned.
   """
 
   docstr = ("Value and gradient of {fun} with respect to positional "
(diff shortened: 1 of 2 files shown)


shoyer:
Do you have an actual use-case here, other than benchmarking the speed of forward vs. reverse mode?

This will be very slightly more efficient than simply evaluating a function `f(x)` and its Jacobian `jax.jacfwd(f)(x)`, but only very slightly and its much worse than using reverse mode `value_and_grad`.

Huizerd:
We're working on comparing different ways of learning in an online fashion, for networks of a size where even RTRL is tractable. But you're right, the added benefit of having `value_and_grad_fwd` is very small, and now that I think of it, it will probably not allow us the flexibility to make approximations to RTRL anyway. Instead, we'd benefit more from having an auxiliary return from `jacfwd`/`jacrev`, which will prevent duplicate code on our side, and maybe a tiny bump in speed. This is actually asked for in #497, so I can adapt this PR to close that.

froystig (approved):
Thanks for the contribution – this looks good! Apologies that my review took a while.

The 3-commit history here introduces and then reverts a large change to api.py. Would you mind [squashing the commits](https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) into one?

froystig:
fixes #497

Huizerd:
Done!

PR #8421, opened 2021-11-01
Implement hermitian=... option on jax.numpy.linalg.svd.

### LINKED ISSUE #8414: Incorrect documentation for jax.numpy.linalg.svd
The [documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.svd.html) shows that `jax.numpy.linalg.svd` accepts the keyword argument `hermitian`, like the numpy.linalg version. But the signature `jax.numpy.linalg.svd(a, full_matrices=True, compute_uv=True)` does not list this, and when one tries to use it an error is thrown.

hawkinsp:
Fixes #8414

### DIFF
--- a/jax/_src/numpy/linalg.py
+++ b/jax/_src/numpy/linalg.py
@@ -53,9 +53,26 @@ def cholesky(a):
 
 
 @_wraps(np.linalg.svd)
-@partial(jit, static_argnames=('full_matrices', 'compute_uv'))
-def svd(a, full_matrices=True, compute_uv=True):
+@partial(jit, static_argnames=('full_matrices', 'compute_uv', 'hermitian'))
+def svd(a, full_matrices: bool = True, compute_uv: bool = True,
+        hermitian: bool = False):
   a = _promote_arg_dtypes(jnp.asarray(a))
+  if hermitian:
+    w, v = lax_linalg.eigh(a)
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks good. Have you benchmarked this? My read is that the purpose of the `hermitian` argument is to allow opting in to a more efficient code path. Is this code path more efficient? If not, perhaps the best way to support this would be via `del hermitian` 😁

hawkinsp:
I haven't benchmarked this extensively, but the assumption is that computing `eigh` is faster than computing the SVD of a general matrix, which I think is a safe assumption. It certainly appears true on CPU:

```
In [1]: import numpy as np

In [2]: x = np.random.randn(1000, 1000)

In [3]: y = x + x.T

In [8]: import jax

In [9]: z = jax.device_put(y)

In [11]: %timeit jnp.linalg.svd(z)[0].block_until_ready()
164 ms ± 5.99 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [12]: %timeit jnp.linalg.svd(z, hermitian=True)[0].block_until_ready()
74.6 ms ± 1.85 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

PR #8438, opened 2021-11-02
[sparse]: add tracer-based implementation of sparsify

jakevdp:
Co-authored by: @mattjj

Eventually I think we'll choose one implementation to stick with; for now we'll keep both in parallel.

### DIFF
--- a/jax/experimental/sparse/transform.py
+++ b/jax/experimental/sparse/transform.py
@@ -70,6 +70,7 @@ from jax.experimental.sparse import BCOO
 sparse_rules : Dict[core.Primitive, Callable] = {}
 
 Array = Any
+ArrayOrSparse = Any
 
 
 class SparseEnv:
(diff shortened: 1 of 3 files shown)


jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+    out_tracers = tuple(SparseTracer(self, argspec=argspec) for argspec in out_argspecs)
+    return out_tracers if primitive.multiple_results else out_tracers[0]
+
+  def process_call(self, call_primitive, f: lu.WrappedFun, tracers, params):
+    spenv = self.main.spenv
+    argspecs = tuple(t._argspec for t in tracers)
+    in_bufs = spenv._buffers
+    fun, out_argspecs = sparsify_subtrace(f, self.main, argspecs)
+    if any(params['donated_invars']):
+      raise ValueError("sparsify does not support donated_invars")
```
mattjj, line 216:
```suggestion
      raise NotImplementedError("sparsify does not support donated_invars")
```

Just an idea. I think one way to support this would be to track how dense argument positions correspond to buffers underlying the sparse representations. Maybe we could implement that later, and hence a `NotImplementedError` could make sense.

jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+    params['donated_invars'] = tuple(False for buf in in_bufs)
```
mattjj, line 217:
```suggestion
    params = dict(params, donated_invars=tuple(False for buf in in_bufs))
```

Better not to mutate the caller's dict.

jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+        raise NotImplementedError(f"sparse rule for {primitive}")
+      out_argspecs = sparse_rules[primitive](self.main.spenv, *(t._argspec for t in tracers), **params)
+    else:
+      out_bufs = primitive.bind(*(argspec.data(spenv) for argspec in argspecs), **params)
+      out_argspecs = arrays_to_argspecs(spenv, out_bufs if primitive.multiple_results else [out_bufs])
+    out_tracers = tuple(SparseTracer(self, argspec=argspec) for argspec in out_argspecs)
+    return out_tracers if primitive.multiple_results else out_tracers[0]
+
+  def process_call(self, call_primitive, f: lu.WrappedFun, tracers, params):
+    spenv = self.main.spenv
```
mattjj, line 211: We could save this for later, but maybe it'd be cleanest to pop the spenv off the main trace here, like

```python
spenv = popattr(self.main, 'spenv')
```

where

```python
def popattr(obj, name):
  val = getattr(obj, name)
  delattr(obj, name)
  return val
```

This might prevent some head-scratching later. (Hopefully it doesn't crash now!)

jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+    argspecs = tuple(t._argspec for t in tracers)
+    in_bufs = spenv._buffers
+    fun, out_argspecs = sparsify_subtrace(f, self.main, argspecs)
+    if any(params['donated_invars']):
+      raise ValueError("sparsify does not support donated_invars")
+    params['donated_invars'] = tuple(False for buf in in_bufs)
+    bufs_out = call_primitive.bind(fun, *in_bufs, **params)
+    self.main.spenv = SparseEnv(bufs_out)
```
mattjj, line 219: Along the lines of the above comment about `popattr`, here we could change the line to

```python
setnewattr(self.main, 'spenv', SparseEnv(bufs_out))
```

where

```python
def setnewattr(obj, name, val):
  sentinel = object()
  assert getattr(obj, name, sentinel) is sentinel
  setattr(obj, name, val)
```

Again, I suspect this might help catch bugs earlier in the future.

jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+    return [SparseTracer(self, argspec=argspec) for argspec in out_argspecs()]
+
+@lu.transformation_with_aux
+def sparsify_subtrace(main, argspecs, *bufs):
+  main.spenv = SparseEnv(bufs)
```
mattjj, line 224: Here's another place for `setnewattr`.

jax/experimental/sparse/transform.py
```diff
@@ -162,6 +163,97 @@ def argspec_to_aval(argspec):
...
+  trace = main.with_cur_sublevel()
+  in_tracers = [SparseTracer(trace, argspec=argspec) for argspec in argspecs]
+  outs = yield in_tracers, {}
+  out_traces = [trace.full_raise(out) for out in outs]
+  buffers = main.spenv._buffers
+  main.spenv = SparseEnv()
```
mattjj, line 230: I think we could change these lines to be more like

```python
spenv = popattr(main, 'spenv')
buffers = spenv._buffers
```

and then leave `main.spenv` deleted rather than setting `main.spenv = SparseEnv` here.

mattjj (approved):
Nice!

I have some optional suggestions about how to keep things a bit cleaner and thus catch potential bugs earlier.

jakevdp:
I spent a half hour trying to do the `popattr` and `setnewattr` solution, and ran into two issues:

1. because `Tracer` delegates `getattr()` to `aval`, and `SparseTracer` needs to reference `spenv` to compute its `aval`, doing this directly leads to infinite recursion errors.
2. Even controlling for that (by gating things on `hasattr`) the code errors if `main.spenv` is removed in the places you indicated, I think because of this `aval` issue.

Perhaps the issue lies in the fact that we're computing avals lazily, and we should instead compute them eagerly when `SparseTracer` is created?

jakevdp:
No, even that doesn't work... we still end up trying to access `main.spenv` during instantiation of `SparseTracer` at moments when it is not defined.

jakevdp:
OK, I think I managed to do roughly what you had in mind. PTAL!

PR #8445, opened 2021-11-03
Print the bazel version from build.py.

hawkinsp:
The main goal of the change is to make sure the version number is printed in failed builds.

In passing, increment the minimum version that build.py checks for to 3.7.2. This is redundant and less restrictive than the `.bazelversion` check, but the additional check does no harm.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -185,30 +185,32 @@ def get_bazel_paths(bazel_path_flag):
 
 def get_bazel_path(bazel_path_flag):
   """Returns the path to a Bazel binary, downloading Bazel if not found. Also,
-  it checks Bazel's version at lease newer than 2.0.0.
+  checks Bazel's version is at least newer than 4.1.0.
 
-  NOTE Manual version check is reasonably only for bazel < 2.0.0. Newer bazel
-  releases performs version check against .bazelversion (see for details
+  A manual version check is needed only for really old bazel versions.
+  Newer bazel releases perform their own version check against .bazelversion
+  (see for details
   https://blog.bazel.build/2019/12/19/bazel-2.0.html#other-important-changes).
   """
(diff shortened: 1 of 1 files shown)


build/build.py
```diff
@@ -185,30 +185,31 @@ def get_bazel_paths(bazel_path_flag):
 
 def get_bazel_path(bazel_path_flag):
   """Returns the path to a Bazel binary, downloading Bazel if not found. Also,
-  it checks Bazel's version at lease newer than 2.0.0.
+  checks Bazel's version is at least newer than 3.7.2.
```
yashk2810, line 188: If .bazelversion is checking for 4.1.0, can we make it the same?

Having 2 checks that are different is confusing.

hawkinsp: Sure, why not. Done.

(I chose 3.7.2 since that is probably an acceptable version on most platforms, but you're probably right that picking one version is fine.)

build/build.py
```diff
@@ -185,30 +185,31 @@ def get_bazel_paths(bazel_path_flag):
...
   """
   for path in filter(None, get_bazel_paths(bazel_path_flag)):
-    if check_bazel_version(path):
-      return path
+    version = get_bazel_version(path)
+    if version is not None and version >= (4, 1, 0):
+      return path, ".".join(map(str, version))
 
-  print("Cannot find or download bazel. Please install bazel.")
+  print("Cannot find or download bazel >= 4.1.0. Please install bazel.")
```
yashk2810, line 200: nit: Please install bazel >= 4.1.0?

I know you say it before but someone might skip it, so its good to add the version that you are asking to install.

yashk2810 (approved):

build/build.py
hawkinsp: I reworded it a bit to avoid needing to repeat the version number.

PR #8449, opened 2021-11-03
ahead-of-time lowering and compilation for pmap

froystig:
Implements the basics of #7733 for `jax.pmap`.

Along the way, factors the internal parallel compilation path into separate lowering and compilation stages.

cc @tamaranorman – a few minor API changes to mesh computations and executables.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1792,11 +1792,11 @@ class PmapCallInfo(NamedTuple):
   flat_fun: lu.WrappedFun
   in_tree: PyTreeDef
   out_tree: PyTreeDef
-  flat_args: Iterable[Any]
-  donated_invars: Iterable[bool]
+  flat_args: Sequence[Any]
+  donated_invars: Sequence[bool]
   in_axes_flat: Sequence[Optional[int]]
   local_axis_size: int
-  global_arg_shapes_flat: Any
+  global_arg_shapes_flat: Sequence[Optional[Tuple[int, ...]]]
   out_axes_thunk: HashableFunction
 
 
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -496,7 +496,9 @@ class Lowered:
   in_tree: PyTreeDef
   out_tree: PyTreeDef
   donate_argnums: Tuple[int]
-  _lowering: Union[dispatch.XlaComputation, pxla.MeshComputation]
+  _lowering: Union[dispatch.XlaComputation,
+                   pxla.MeshComputation,
+                   pxla.PmapComputation]
   _no_kwargs: bool
 
   def __init__(self, lowering, in_tree, out_tree, donate_argnums,
@@ -532,7 +534,9 @@ class Compiled:
   in_tree: PyTreeDef
   out_tree: PyTreeDef
   donate_argnums: Tuple[int]
-  _executable: Union[dispatch.XlaCompiledComputation, pxla.MeshExecutable]
+  _executable: Union[dispatch.XlaCompiledComputation,
+                     pxla.MeshExecutable,
+                     pxla.PmapExecutable]
   _no_kwargs: bool
 
   def __init__(self, executable, in_tree, out_tree, donate_argnums,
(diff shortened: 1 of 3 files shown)


jax/interpreters/pxla.py
```diff
@@ -913,112 +936,193 @@ def parallel_callable(fun: lu.WrappedFun,
...
+                         local_arg_parts,
+                         local_out_parts,
+                         out_parts,
+                         num_local_replicas,
+                         num_local_shards,
+                         num_global_shards,
+                         num_global_replicas,
+                         avals,
+                         sharded_avals,
+                         out_sharded_avals)
```
mattjj, line 964: This is a huge struct. Maybe some of the information is redundant. To cut it down in size, maybe some of these values can be computed in terms of others, and don't have to be stored?

(And maybe join these lines instead of having one argument per line, here and elsewhere?)

jax/_src/api.py
```diff
@@ -1750,6 +1754,95 @@ def pmap(
       global_arg_shapes=global_arg_shapes)
 
 
+class PmapCallInfo(NamedTuple):
+  flat_fun: lu.WrappedFun
+  in_tree: PyTreeDef
+  out_tree: PyTreeDef
+  flat_args: Iterable[Any]
+  donated_invars: Iterable[bool]
+  in_axes_flat: Any
```
apaszke, line 1763: Don't we have a better annotation for this? I would expect it's a `Sequence[Optional[int]]`

jax/interpreters/pxla.py
```diff
@@ -743,9 +742,10 @@ def xla_pmap_impl(fun: lu.WrappedFun, *args, backend, axis_name, axis_size,
                           ("fingerprint", fingerprint))
   return compiled_fun(*args)
 
+
 @lu.cache
 def parallel_callable(fun: lu.WrappedFun,
-                      backend_name: Optional[str],
+                      backend: Optional[str],
```
apaszke, line 748: `backend_name` is a slightly better name IMO, because `backend` sounds more like a real backend object from the XLA client.

jax/interpreters/pxla.py
```diff
@@ -769,15 +788,15 @@ def parallel_callable(fun: lu.WrappedFun,
         f"Specified axis_size {global_axis_size} doesn't match received "
         f"axis_size {axis_size}.")
 
-  if devices is not None and backend_name is None:
-    backend = xb.get_device_backend(devices[0])
+  if devices is not None and backend is None:
+    be = xb.get_device_backend(devices[0])
   else:
-    backend = xb.get_backend(backend_name)
+    be = xb.get_backend(backend)
```
apaszke, line 794: Sorry but this naming change is really weird and IMO significantly lowers the readability of this function

apaszke (approved):
This change is pretty large so I haven't done a super careful review, but it generally looks good. A couple minor points around renaming which IMO makes the code much harder to follow.

I also agree that it would be good to avoid passing in a million values to `PmapComputation`. Lots of them do seem redundant.

jax/interpreters/pxla.py
froystig: I've done some refactoring. Thoughts?

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+  shards = ShardInfo(
+      sharded_avals, out_sharded_avals, global_sharded_avals,
+      num_local_shards, num_global_shards)
+
+  return jaxpr, consts, replicas, parts, shards
+
+
+def lower_parallel_callable(fun: lu.WrappedFun,
+                            backend_name: Optional[str],
+                            axis_name,
```
mattjj, line 887:
```suggestion
                            axis_name: core.AxisName,
```

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+                            axis_size: int,
+                            global_axis_size: Optional[int],
+                            devices: Optional[Sequence[Any]],
```
mattjj, line 890:
```suggestion
                            devices: Optional[Sequence[xla.Device]],
```

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+                            name: str,
+                            in_axes: Iterable[Optional[int]],
+                            out_axes_thunk: Callable[[], Sequence[Optional[int]]],
+                            donated_invars: Iterable[bool],
```
mattjj, line 894:
```suggestion
                            donated_invars: Sequence[bool],
```

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+                            global_arg_shapes,
```
mattjj, line 895:
```suggestion
                            global_arg_shapes: Sequence[Optional[Tuple[int, ...]]],
```

Should double check that.

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+                            avals: Sequence[Any]):
```
mattjj, line 896:
```suggestion
                            avals: Sequence[core.AbstractValue]):
```

mattjj (approved):
Nice!

jax/interpreters/pxla.py
```diff
@@ -767,6 +778,122 @@ def parallel_callable(fun: lu.WrappedFun,
...
+  pmap_computation = lower_parallel_callable(
+      fun, backend_name, axis_name, axis_size, global_axis_size, devices, name,
+      in_axes, out_axes_thunk, donated_invars, global_arg_shapes, avals)
+  pmap_executable = pmap_computation.compile()
+  return WeakRefList([pmap_executable.unsafe_call, pmap_executable.fingerprint])
+
+
+@dataclasses.dataclass(frozen=True)
+class ParallelCallableInfo:
+  backend: Any
```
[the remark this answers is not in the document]
froystig: It's `xla.Backend`. Annotating it as such seems to require changes throughout `xla_bridge`, which houses functions that work when given an `xla.Backend` but are annotated only for `str | None`. I left a comment and can try upgrading `xla_bridge` in a separate change, so that we can fix this up.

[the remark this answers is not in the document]
froystig: sent as #8656

PR #8466, opened 2021-11-04
[sparse] register BCOO as a vmappable

jakevdp:
Cleaning up `sparse.BCOO` based on the new vmappable registry from #8451. This allows us to simply pass `shape` when flattening and unflattening, rather than having to construct and deconstruct the sparse part of the shape.

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -981,10 +981,6 @@ class BCOO(ops.JAXSparse):
   n_sparse = property(lambda self: self.indices.shape[-1])
   n_dense = property(lambda self: self.data.ndim - 1 - self.n_batch)
 
-  @property
-  def _sparse_shape(self):
-    return tuple(self.shape[self.n_batch:self.n_batch + self.n_sparse])
-
   def __init__(self, args, *, shape):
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -1111,3 +1089,27 @@ def sum(self, *args, **kwargs):
...
     from jax.experimental.sparse import sparsify
     return sparsify(lambda x: x.sum(*args, **kwargs))(self)
+
+Array = Any
+ArrayElt = Any
+Int = Union[int, core.Tracer]
+
+
+def _bcoo_to_elt(cont: Callable[[Array, Optional[int]], ArrayElt],
+                 _: int, val: BCOO, axis: Optional[int]) -> BCOO:
```
mattjj, line 1099:
```suggestion
                 _, val: BCOO, axis: Optional[int]) -> BCOO:
```

(I think this was meant to be `Int` instead of `int`, but I had to change this type at the last minute so maybe just don't annotate it for now, or call it a `Callable` if you want to convey some information.)

jakevdp: Done

mattjj (approved):
Woo!

PR #8468, opened 2021-11-05
add checkify transform and error-checking evaluator

mattjj:
Improved debugging is a top priority for us.

This PR is a proof-of-concept for a functional error-checking transformation, together with a higher-level convenience API which applies the transformation to all JAX operations and raises a Python exception when it spots an error.

First, some motivation. When staging out computations with `jit` or `pmap`, we lose the ability to detect and locate runtime errors which might occur in the middle of a computation. These runtime errors include things like unexpected nans being generated, or out-of-bounds indexing, or overflow.

The `jax.debug_nans` mechanism is one tool which can help, but it has big limitations, including:
1. because there's no eager `pmap`, the `debug_nans` mechanism can't recurse into `pmap`ped computations like it can recurse into `jit`ted ones, ultimately meaning there are no automatic ways to find where a nan is occurring in `pmap` computation (standard advice: try replacing `pmap` with `vmap`);
2. it only applies to nans and not other kinds of runtime errors.

Still, the basic idea is a nice one: since nans are _values_ which indicate errors, they fit nicely in a functional system like JAX.

This PR aims to generalize the debug_nans mechanism to overcome these limitations.

(See #7595 for a related error-values-based approach to detecting out-of-bounds indexing, though it requires changing expressions like `x[idx]` to `x.at[idx].get(mode='fill')`, and moreover [the error values for integer- or boolean-valued `x`](https://github.com/google/jax/blob/c90fbadd4718070604b4fd1ebb3b656d89a821fa/jax/_src/lax/lax.py#L1079-L1090) could be mistaken to be valid.)

So far this is joint work with @hawkinsp.

---

Here's a toy example:

```python
import jax
import jax.numpy as jnp

@jax.pmap         # line 4
def f(x, i):      # line 5
  y = x[i]        # line 6
  z = jnp.cos(y)  # line 7
  return z
```

```python
zs = f(jnp.array([[3., 4., 5.],
                  [6., 7., 8.]]),
       jnp.array([2, 0]))
print(zs)  # [0.2836622  0.96017027]
```

All's well and good here, but what if we call this function on arguments which result in a nan being produced?

```python
zs = f(jnp.array([[3., 4., 5.],
                  [jnp.inf, 7., 8.]]),
       jnp.array([2, 0]))
print(zs)  # [0.2836622       nan]
```

Unfortunately, `jax.debug_nans` is not much help with `pmap` because there's no eager `pmap` (yet) and thus we can't get a more precise error message than "something in the pmapped function produced a nan" (thanks, we knew that already!):

```python
with jax.debug_nans(True):
  zs = f(jnp.array([[3., 4., 5.],
                    [jnp.inf, 7., 8.]]),
        jnp.array([2, 0]))
```

```
  File "/usr/local/google/home/mattjj/packages/jax/example.py", line 11, in <module>
    zs = f(jnp.array([[3., 4., 5.],
  File "/usr/local/google/home/mattjj/packages/jax/jax/interpreters/xla.py", line 489, in _check_special
    raise FloatingPointError(f"invalid value (nan) encountered in {name}")
FloatingPointError: invalid value (nan) encountered in parallel computation
```

And this is just for nans: what about other runtime errors, like out-of-bounds indexing?

```python
zs = f(jnp.array([[3., 4., 5.],
                  [jnp.inf, 7., 8.]]),
        jnp.array([5, 0]))  # note: 5 is out-of-bounds!
```

```
[0.2836622       nan]
```

JAX silently clips the index to be in-bounds, rather than reporting any error. Hm!

---

Okay, here's the new content of this PR:

```python
from jax.experimental import checkify

with checkify.runtime_checks():
  zs = f(jnp.array([[3., 4., 5.],
                  [jnp.inf, 7., 8.]]),
        jnp.array([5, 0]))
```

```
  File "/usr/local/google/home/mattjj/packages/jax/example.py", line 11, in <module>
    zs = f(jnp.array([[3., 4., 5.],
  File "/usr/local/google/home/mattjj/packages/jax/jax/experimental/checkify.py", line 306, in process_call
    raise Exception(f'inside {name}-decorated function {fun_info}:'
Exception: inside pmap-decorated function f at ~/packages/jax/example.py:4:
at mapped index 0: out-of-bounds indexing at ~/packages/jax/example.py:6 (f)
at mapped index 1: nan generated by primitive cos at ~/packages/jax/example.py:7 (f)
```

Since this is a proof-of-concept, any detail is subject to change. But the big idea is that we'll at least get line numbers (or stack traces if we want them!) for runtime errors, even inside a `pmap`.

---

This `runtime_checks` context manager just applies a `checkify` transformation to every JAX primitive that gets evaluated (including `pmap` primitives). The `checkify` transformation adds an extra output to any function it transforms, and that extra output is just an error value. (We could say something like `checkify : (a -> b) -> a -> Err b` where, essentially, `data Err a = Err Bool Int a`.)

For example:

```python
@jax.jit
def f(x1, x2):
  y1 = jnp.sin(x1)
  y2 = jnp.sin(x2)
  return y1 + y2

err, z = checkify.checkify(f)(3., jnp.inf)
print(z)  # nan
print(err.get())  # nan generated by primitive sin at ~/packages/jax/debugging_example.py:20 (f)
```

Take a look at:
1. [the `Error` class](https://github.com/google/jax/pull/8468/files#diff-248334448f17d11421b048052ce19398f60f4054c31f6683cb3effe296fb008aR54-R68) (a simple dataclass [registered as a pytree](https://github.com/google/jax/pull/8468/files#diff-248334448f17d11421b048052ce19398f60f4054c31f6683cb3effe296fb008aR70-R72)),
2. the form of [the `checkify` rules](https://github.com/google/jax/pull/8468/files#diff-248334448f17d11421b048052ce19398f60f4054c31f6683cb3effe296fb008aR225-R254) for each primitive, and
3. [the `assert_func`](https://github.com/google/jax/pull/8468/files#diff-248334448f17d11421b048052ce19398f60f4054c31f6683cb3effe296fb008aR81-R85) those rules call (notice it's not even a primitive, and the `next_code()` global could be avoided with `uuid4`, so this implementation could be completely functionally pure if we wanted!).

---

There are some known issues we need to sort out, both in terms of API design and in terms of implementation:
* [ ] How can we make it easy to collect _all_ errors? (We need something that looks like a set of integer error codes rather than a single integer.)
* [ ] Should the public API be the context manager, or should we also surface the functional `checkify` API? Should we add an effectful 'assert' API in addition, for convenience?
* [ ] `checkify` only works by data dependence at present (and it turns out that we don't currently have a good way to make a _transformation_, rather than a bottom-of-the-stack interpreter, work in any other way), but the technique in the no-more-post-process core revision branch should let us solve that.
* [ ] Relatedly, the trick I pulled to make the context manager / `CheckifyEvaluator` work was shady and required abusing `post_process_call`, so that should be fixed (I think by the no-more-post-process branch technique).
* [ ] Add more rules! Especially for the other higher-order primitives (there's a `cond` rule to prove it's possible).


patrick-kidger:
Having been pointed at this via #8572 (thanks @jakevdp!)

- Is this able to catch normal runtime errors, like those discussed in #8572? At least over there, what's really desired is the ability to flag that we should raise a standard Python exception after a JIT block has been executed.
- It sounds like "producing a NaN" is special-cased as an error, alongside "proper" errors like out-of-bounds indexing. Is this optional? I frequently use NaNs to represent missing data.

mattjj:
We've started landing this in other PRs.

PR #8491, opened 2021-11-09
[sparse] bcoo_dot_general: support contraction along batch dimensions

jakevdp:
Addresses request in #8454

I was hoping I could make the implementation more intuitive to follow but... it's complicated. I have to admit, without the tests I'd be very uncertain of its correctness.

As part of this, I refactored the associated test methods to programmatically generate cases from the whole range of supported inputs.


froystig (approved):
Really nice work!

PR #8514, opened 2021-11-11
Update bcoo_dot_general GPU translation rule.

tlu7:
Co-authored-by: vanderplas@google.com

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -33,7 +33,11 @@ from jax._src.api_util import flatten_axes
 from jax._src.lax.lax import (
   ranges_like, remaining, _dot_general_batch_dim_nums, _dot_general_shape_rule,
   DotDimensionNumbers)
+from jax._src.lib import cusparse
 from . import ops
+from jax._src.lib import xla_client as xc
+
+xops = xc._xla.ops
 
 Dtype = Any
 Shape = Tuple[int, ...]
@@ -581,6 +585,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
   out_dtype = jnp.promote_types(lhs_data.dtype, rhs.dtype)
   return core.ShapedArray(out_shape, out_dtype)
 
+_bcoo_dot_general_default_translation_rule = xla.lower_fun(
+    _bcoo_dot_general_impl, multiple_results=False, new_style=True)
+
+def _bcoo_dot_general_cuda_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), _ = dimension_numbers
+  _, lhs_indices_aval, _, = avals_in
+  props = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  lhs_transpose = False
+  if props.n_sparse == 1:
+    # Converts lhs to a row vector.
+    col = xops.Collapse(lhs_indices, dimensions=[0, 1])
+    row = xops.Broadcast(xops.Constant(ctx.builder, np.array(0, dtype=int)),
+                         ctx.builder.get_shape(col).dimensions())
+    lhs_shape = (1, lhs_shape[0])
+    lhs_transpose = True
+  elif props.n_sparse == 2:
+    row = xops.Collapse(
+        xops.SliceInDim(lhs_indices, start_index=0, limit_index=1, stride=1, dimno=1),
+        dimensions=[0, 1])
+    col = xops.Collapse(
+        xops.SliceInDim(lhs_indices, start_index=1, limit_index=2, stride=1, dimno=1),
+        dimensions=[0, 1])
+    if lhs_contract[0] == 0:
+      lhs_transpose = True
+
+  rhs_ndim = len(ctx.builder.get_shape(rhs).dimensions())
+  if rhs_ndim == 1:
+    bcoo_dot_general_fn = cusparse.coo_matvec
+  elif rhs_ndim == 2:
+    bcoo_dot_general_fn = cusparse.coo_matmat
+    if rhs_contract[0] == 1:
+      rhs = xops.Transpose(rhs, permutation=[1, 0])
+
+  return [bcoo_dot_general_fn(ctx.builder, lhs_data, row, col, rhs,
+                              shape=lhs_shape, transpose=lhs_transpose)]
+
+def _bcoo_dot_general_gpu_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), (lhs_batch, rhs_batch) = dimension_numbers
+  _, lhs_indices_aval, rhs_aval, = avals_in
+  n_batch, n_sparse, n_dense, _ = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  if n_batch or n_dense or n_sparse not in [1, 2] or rhs_aval.ndim not in [1, 2] or lhs_batch or rhs_batch or len(lhs_contract) != 1:
+    return _bcoo_dot_general_default_translation_rule(
+      ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs,
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
+  else:
+      return _bcoo_dot_general_cuda_translation_rule(
+      ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs,
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
+
 def _bcoo_dot_general_jvp_lhs(lhs_data_dot, lhs_data, lhs_indices, rhs, *, dimension_numbers, lhs_shape):
   return bcoo_dot_general(lhs_data_dot, lhs_indices, rhs, dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
 
@@ -646,8 +707,11 @@ def _bcoo_dot_general_batch_rule(batched_args, batch_dims, *, dimension_numbers,
 ad.defjvp(bcoo_dot_general_p, _bcoo_dot_general_jvp_lhs, None, _bcoo_dot_general_jvp_rhs)
 ad.primitive_transposes[bcoo_dot_general_p] = _bcoo_dot_general_transpose
 batching.primitive_batchers[bcoo_dot_general_p] = _bcoo_dot_general_batch_rule
-xla.register_translation(bcoo_dot_general_p, xla.lower_fun(
-    _bcoo_dot_general_impl, multiple_results=False, new_style=True))
+xla.register_translation(bcoo_dot_general_p, _bcoo_dot_general_default_translation_rule)
+if cusparse and cusparse.is_supported:
+  xla.register_translation(bcoo_dot_general_p,
+                           _bcoo_dot_general_gpu_translation_rule,
+                           platform='gpu')
 
 #----------------------------------------------------------------------
 # bcoo_dot_general_sampled
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -592,27 +592,12 @@ def _bcoo_dot_general_cuda_translation_rule(
     ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
     lhs_shape):
 
+  lhs_indices = xops.ConvertElementType(
+      lhs_indices, xla.dtype_to_primitive_type(np.dtype(np.int)))
+
   (lhs_contract, rhs_contract), _ = dimension_numbers
   _, lhs_indices_aval, _, = avals_in
   props = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
-
-  lhs_transpose = False
-  if props.n_sparse == 1:
-    # Converts lhs to a row vector.
-    col = xops.Collapse(lhs_indices, dimensions=[0, 1])
-    row = xops.Broadcast(xops.Constant(ctx.builder, np.array(0, dtype=int)),
-                         ctx.builder.get_shape(col).dimensions())
-    lhs_shape = (1, lhs_shape[0])
-    lhs_transpose = True
-  elif props.n_sparse == 2:
-    row = xops.Collapse(
-        xops.SliceInDim(lhs_indices, start_index=0, limit_index=1, stride=1, dimno=1),
-        dimensions=[0, 1])
-    col = xops.Collapse(
-        xops.SliceInDim(lhs_indices, start_index=1, limit_index=2, stride=1, dimno=1),
-        dimensions=[0, 1])
-    if lhs_contract[0] == 0:
-      lhs_transpose = True
 
   rhs_ndim = len(ctx.builder.get_shape(rhs).dimensions())
   if rhs_ndim == 1:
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,52 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
   out_dtype = jnp.promote_types(lhs_data.dtype, rhs.dtype)
   return core.ShapedArray(out_shape, out_dtype)
 
+
+_bcoo_dot_general_gpu_translation_rule_default = xla.lower_fun(
+    _bcoo_dot_general_impl, multiple_results=False, new_style=True)
```
jakevdp, line 581: I would call this something like `_bcoo_dot_general_default_translation_rule`, and then refer to it again when doing `xla.register_translation(bcoo_dot_general_p, ...)` rather than creating a second copy of the same lowered function.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,52 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+
+def _bcoo_dot_general_gpu_translation_rule(
```
jakevdp, line 583: It looks like this function is not being referenced anywhere. We'll need to add it to the device-specific translation table, in a similar manner to this: https://github.com/google/jax/blob/7f3609f039513360335b2f11ec16ffc2a9519dd0/jax/experimental/sparse/ops.py#L577-L579

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,52 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+
+  if n_batch or n_dense or n_sparse != 2 or rhs_aval.ndim > 2 or lhs_batch or rhs_batch or not lhs_contract:
+    return _bcoo_dot_general_gpu_translation_rule_default(
+      ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs,
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
+
+  lhs_transpose = False
+  if n_sparse == 1:
+    # Converts lhs to a row vector.
+    col = lhs_indices
```
jakevdp, line 599: The shape of `col` here is `(nse, 1)`. Does it need to be collapsed?

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,52 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+
+def _bcoo_dot_general_gpu_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), (lhs_batch, rhs_batch) = dimension_numbers
+  _, lhs_indices_aval, rhs_aval, = avals_in
+  n_batch, n_sparse, n_dense, _ = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  if n_batch or n_dense or n_sparse != 2 or rhs_aval.ndim > 2 or lhs_batch or rhs_batch or not lhs_contract:
```
jakevdp, line 591: another thing to check here: the cusparse lowering requires `len(lhs_contract) == 1`

tests/sparse_test.py
```diff
@@ -731,9 +731,13 @@ def test_bcoo_todense_partial_batch(self, shape, dtype, n_batch, n_dense):
       for lhs_shape, rhs_shape, lhs_contracting, rhs_contracting in [
           [(5,), (6,), [], []],
           [(5,), (5,), [0], [0]],
+          [(5,), (5, 7), [0], [0]],
+          [(5,), (7, 5), [0], [1]],
           [(5, 7), (5,), [0], [0]],
           [(7, 5), (5,), [1], [0]],
           [(3, 5), (2, 5), [1], [1]],
+          [(3, 5), (5, 2), [1], [0]],
+          [(5, 3), (2, 5), [0], [1]],
```
jakevdp, line 740: It wouldn't hurt I think to cover these in a new GPU-only test method specifically to test cases that should be lowered to GPU. If something changes with cusparse and leads to those failing, it would be a much clearer signal of the root issue.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,52 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+    return _bcoo_dot_general_gpu_translation_rule_default(
+      ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs,
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
```
jakevdp, line 594: Design-wise, it occurred to me that it might be cleaner to put the cusparse logic in another function, so then this translation rule would just contain something like
```
if <conditions>:
  return _bcoo_dot_general_cuda_translation_rule(...)
else:
  return _bcoo_dot_general_default_translation_rule(...)
```
What do you think?

jakevdp (changes requested):
Wow, this looks great! A couple comments below

jax/experimental/sparse/bcoo.py
tlu7: Done.

jax/experimental/sparse/bcoo.py
tlu7: Done.

jax/experimental/sparse/bcoo.py
tlu7: Fall back to default rule when `len(lhs_contract) != 1` is true.

jakevdp:
Hi @tlu7 - I saw your responses to some of the suggestions, but not the updates. Do you have some commits to push to the branch?

tests/sparse_test.py
tlu7: Separate the test cases into a standalone test method. Is there a way to only pick up the method on GPU?

tlu7:
> Hi @tlu7 - I saw your responses to some of the suggestions, but not the updates. Do you have some commits to push to the branch?

Thanks for checking! Just commit the changes. PTAL.

tests/sparse_test.py
jakevdp: Yes, you can do something similar to this: https://github.com/google/jax/blob/1388dbe25c3b1c003955a9f4743dca72668d2d8d/tests/sparse_test.py#L135

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+      return data, indices, lhs, rhs
+    dimension_numbers = ((lhs_contracting, rhs_contracting), ([], []))
+
+    def f_dense(data, indices, lhs, rhs):
+      return lax.dot_general(lhs, rhs, dimension_numbers=dimension_numbers)
+
+    def f_sparse(data, indices, lhs, rhs):
+      return sparse.bcoo_dot_general(data, indices, rhs,
+                                         lhs_shape=lhs.shape,
+                                         dimension_numbers=dimension_numbers)
```
jakevdp, line 810: nit: match indentation

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+
+    self._CheckAgainstNumpy(f_dense, f_sparse, args_maker)
+    self._CheckAgainstNumpy(f_dense, jit(f_sparse), args_maker)
```
jakevdp, line 813: we could probably replace the second one with `self._CompileAndCheck(f_sparse, args_maker)` here

This was causing issues with the default impl rule, but I suspect it will be fine in this case.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+  _, lhs_indices_aval, _, = avals_in
+  _, n_sparse, _, _ = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  lhs_transpose = False
+  if n_sparse == 1:
+    # Converts lhs to a row vector.
+    col = xops.Collapse(lhs_indices, dimensions=[0, 1])
+    row = xops.Broadcast(xops.Constant(ctx.builder, np.array(0, dtype=int)),
+                         ctx.builder.get_shape(col).dimensions())
+    lhs_shape = (1, lhs_shape[0]) if len(lhs_shape) == 1 else sorted(lhs_shape)
```
jakevdp, line 596: Under what circumstance is `len(lhs_shape) != 1`? I think `n_dense == n_batch == 0` and `n_sparse == 1` is sufficient to guarantee that `len(lhs_shape) == `.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+
+def _bcoo_dot_general_gpu_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), (lhs_batch, rhs_batch) = dimension_numbers
+  _, lhs_indices_aval, rhs_aval, = avals_in
+  n_batch, n_sparse, n_dense, _ = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  if n_batch or n_dense or n_sparse != 2 or rhs_aval.ndim > 2 or lhs_batch or rhs_batch or len(lhs_contract) != 1:
```
jakevdp, line 627: It looks like the cuda rule as implemented  handles `n_sparse == 1` as well. Also, I think it requires `rhs_aval.ndim in [1, 2]`?

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+          [(5,), (5, 7), [0], [0]],
+          [(5,), (7, 5), [0], [1]],
+          [(5, 7), (5,), [0], [0]],
+          [(7, 5), (5,), [1], [0]],
+          [(3, 5), (2, 5), [1], [1]],
+          [(3, 5), (5, 2), [1], [0]],
+          [(5, 3), (2, 5), [0], [1]],
+          [(5, 3), (5, 2), [0], [0]],
+      ]
+      for n_dense in range(len(lhs_shape) - max(lhs_contracting, default=0))
```
jakevdp, line 791: Probably we can remove this `n_dense` logic and just set it to zero, because GPU lowering is not applicable for `n_dense != 0`

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+      for dtype in jtu.dtypes.floating + jtu.dtypes.complex))
+  def test_bcoo_dot_general_no_dense_no_batch_only_contract_on_gpu(
```
jakevdp, line 793: Maybe `test_bcoo_dot_general_cusparse`?

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+_bcoo_dot_general_default_translation_rule = xla.lower_fun(
+    _bcoo_dot_general_impl, multiple_results=False, new_style=True)
+
+def _bcoo_dot_general_cuda_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), _ = dimension_numbers
+  _, lhs_indices_aval, _, = avals_in
+  _, n_sparse, _, _ = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
```
jakevdp, line 588: minor: it's slightly cleaner I think to do `props = _validate_bcoo_indices(...)` and then refer to `props.n_sparse` below.

jakevdp (changes requested):

tlu7:
Thanks for the review! PTAL.
@jakevdp

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+          [(5, 7), (5,), [0], [0]],
+          [(7, 5), (5,), [1], [0]],
+          [(3, 5), (2, 5), [1], [1]],
+          [(3, 5), (5, 2), [1], [0]],
+          [(5, 3), (2, 5), [0], [1]],
+          [(5, 3), (5, 2), [0], [0]],
+      ]
+      for dtype in jtu.dtypes.floating + jtu.dtypes.complex))
+  def test_bcoo_dot_general_cusparse(
+    self, lhs_shape, rhs_shape, dtype, lhs_contracting, rhs_contracting, n_dense):
```
jakevdp, line 793: remove `n_dense` from argument list

tests/sparse_test.py
```diff
@@ -767,6 +767,51 @@ def f_sparse(data, indices, lhs, rhs):
...
+    rng = jtu.rand_small(self.rng())
+    rng_sparse = rand_sparse(self.rng())
+    def args_maker():
+      lhs = rng_sparse(lhs_shape, dtype)
+      rhs = rng(rhs_shape, dtype)
+      data, indices = sparse.bcoo_fromdense(lhs, n_dense=n_dense)
```
jakevdp, line 799: remove `n_dense=n_dense` here

jakevdp (commented):
A couple small things, then we can trigger the GPU CI jobs

tlu7:
> A couple small things, then we can trigger the GPU CI jobs

Thanks again for the comments!

jakevdp (approved):

jakevdp (approved):

jakevdp:
The GPU tests are failing with an illegal memory address error. I'm not certain what may be causing it.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+  props = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  lhs_transpose = False
+  if props.n_sparse == 1:
+    # Converts lhs to a row vector.
+    col = xops.Collapse(lhs_indices, dimensions=[0, 1])
+    row = xops.Broadcast(xops.Constant(ctx.builder, np.array(0, dtype=int)),
+                         ctx.builder.get_shape(col).dimensions())
+    lhs_shape = (1, lhs_shape[0])
+    lhs_transpose = True
```
jakevdp, line 597: The tests are failing in this block (the lhs matrix is 1-dimensional). I think this should be `lhs_transpose = False`, yes? We've constructed a `[1, N]` matrix, so it does not need to be transposed.

jax/experimental/sparse/bcoo.py
```diff
@@ -572,6 +576,63 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+def _bcoo_dot_general_cuda_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
+
+  (lhs_contract, rhs_contract), _ = dimension_numbers
+  _, lhs_indices_aval, _, = avals_in
+  props = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  lhs_transpose = False
+  if props.n_sparse == 1:
```
jakevdp, line 591: I would add `assert lhs_contracting == [0]` here just to be sure the input is valid.

jax/experimental/sparse/bcoo.py
jakevdp: Also, in this case the cusparse matmat routine will return a `[1, M]` matrix, but we want it to return a length-M vector. So we'll need to do some post-processing of the result if `lhs.ndim == 1`.

tlu7:
> The GPU tests are failing with an illegal memory address error. I'm not certain what may be causing it.

The failure occurs in the case where `n_sparse = 1`.  The reason for extending the 1D lhs to 2D is that in `cusparse.py`, both `coo_matvec` and `coo_matmat` require `rows, cols = shape`.  
https://github.dev/google/jax/blob/8381a4ade0bdff7869e95e1c0ab6b326bf7aca3a/jaxlib/cusparse.py#L276-L277

I think instead of going from 1D to 2D and then back to 1D , we should add  `coo_vecmat` in `cusparse.py`. This will be helpful for cases where n_batch and n_dense are not zeros.

What do you think?

jakevdp:
(summarizing offline convo: cusparse doesn't have a sparse vec-mat operation, but it does have a sparse vec-vec operation. So for the latter it would probably make sense to expose it in `cusparse.py`, but for the former we'll have to do reshapes at some level anyway)

jakevdp (approved):

Issue #8367, opened 2021-10-26
Jax2TF shape polymorphism not compatible with tf.image.resize


ksagoog:
I am trying to export a model with Jax2TF but am running into some issues with polymorphic shapes and their interaction with jax.image.resize ops. The following snippet raises an error for me:

```
from jax.experimental import jax2tf

polymorphic_image_shape = jax2tf.shape_poly.PolyShape(
  1, 'height', 'width', 3
)

def apply_fn(image):
  b,h,w,c = image.shape
  return jax.image.resize(image, (b,h*2,w*2,c), method="linear")

apply_fn_jitted = jax.jit(apply_fn, backend='cpu')
apply_fn_tf = tf.function(
    jax2tf.convert(
        apply_fn_jitted,
        with_gradient=False,
        enable_xla=True,
        polymorphic_shapes=[polymorphic_image_shape]),
    autograph=False,
    jit_compile=True)

image_f = np.random.uniform(size=(1, 480, 480, 3))

apply_fn_tf(image_f)
```

It would be good for my JAX models with resize methods in them to be exportable to TF. Please let me know if I can provide any additional details. Thank you!

gnecula:
Closed by #8516

PR #8516, opened 2021-11-11
[shape_poly] Update arange and image_resize to support shape polymorphism

gnecula:
Bug: #8367

Small refactoring to jax.image.resize to make it compatible with
shape polymorphismin jax2tf. In the process added also support for
jnp.arange([dim_poly]). Note that the underlying lax.iota already
supported shape polymorphism.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3755,20 +3755,22 @@ def identity(n, dtype=None):
 
 
 @_wraps(np.arange)
-def arange(start, stop=None, step=None, dtype=None):
+def arange(start: core.DimSize, stop: Optional[core.DimSize]=None,
+           step: Optional[core.DimSize]=None, dtype=None):
   lax._check_user_dtype_supported(dtype, "arange")
   require = partial(core.concrete_or_error, _np_asarray)
   msg = "It arose in jax.numpy.arange argument `{}`.".format
+  dtype = dtype or _dtype(start, *(x for x in [stop, step] if x is not None))
   if stop is None and step is None:
-    start = require(start, msg("stop"))
-    dtype = dtype or _dtype(start)
-    return lax.iota(dtype, np.ceil(start).astype(int)) # avoids materializing
+    if not core.is_dim_size(start):
+      start = require(start, msg("stop"))
+      start = np.ceil(start).astype(int)
+
+    return lax.iota(dtype, start)
   else:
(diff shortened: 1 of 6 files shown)


hawkinsp (commented):
The change looks fine, but the test failure looks concerning. Why did we get a numerical difference?

gnecula:
I agree the numerical change is a problem, will look into it.

gnecula:
Drop in favor of #8569

gnecula:
> The change looks fine, but the test failure looks concerning. Why did we get a numerical difference?

The change was correct but triggered an XLA compilation error (bad result). I filed XLA bug b/206898375. Previously this was not an issue because we were doing the computations in question using numpy, now we do them with jax.numpy. I added a workaround here.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3755,20 +3755,22 @@ def identity(n, dtype=None):
 
 
 @_wraps(np.arange)
-def arange(start, stop=None, step=None, dtype=None):
+def arange(start: core.DimSize, stop: Optional[core.DimSize]=None,
+           step: Optional[core.DimSize]=None, dtype=None):
   lax._check_user_dtype_supported(dtype, "arange")
   require = partial(core.concrete_or_error, _np_asarray)
   msg = "It arose in jax.numpy.arange argument `{}`.".format
+  dtype = dtype or _dtype(start, *(x for x in [stop, step] if x is not None))
```
jakevdp, line 3763: Post-submit: I was trying to incorporate this into the X64 changes, and I'm a bit confused by this line. Below, there is special logic in case `start` is a special dimension handler. But this line (with `_dtype = dtypes.result_type`) fails in this case because `start` cannot be interpreted as a dtype.

What is the correct return type in case `start` is a special dimension handler and `dtype` is not specified?

gnecula: `start` should not be of type `core.DimensionHandler`, it may be though of type `shape_poly._DimPolynomial`.
This should work due to https://github.com/google/jax/blob/639517f97f155fa5c74b449f8c177ddf7b5b9ef1/jax/experimental/jax2tf/shape_poly.py#L453

This case is tested by shape_poly_tests: 

```
pytest --durations=10 -r a --verbosity 1 -s jax/experimental/jax2tf/tests/shape_poly_test.py -k "arange_start_no_dtype"
```

PR #8522, opened 2021-11-11
Add numerical comparison for jax2tf -> TFLite

marcvanzee:
This PR adds numerical comparisons for all JAX functions that are converted to TFLite. 

The PR compares the outputs using `np.allclose` with relative tolerance set to `1e-05`. If the tolerances are too big, an error is raised, which is then catched by the eval framework and written to the results table. The comparison is done 5 times for random inputs with different seeds and max_values.

The numerical comparison is done in three stages:

- Between the JAX function and the obtained TF function from jax2tf with `enable_xla=True`. If there are failures here, it means we have a problem in the jax2tf emitter (`jax2tf.py`).
- Between the JAX function and the obtained TF function from jax2tf with `enable_xla=False`. If there are failures here, it means we have a problem in the special no xla code (`impl_no_xla.py`).
- Between TF and the TFLite after using the TFLite converter. Errors here indicate problems with the TFLite converter.

The PR uncovers a number of interesting failure cases, which all occur in the first stage, indicating bugs in jax2tf. I am planning to creating a doc with "Known limitations" and writing them all down, since otherwise we may lose track of things.

### Analysis of failures

There are currently 2 failing examples for `enable_xla=True`, 4 remaining failing Flax examples due to numerical differences (`lm1b`, `pixelcnn++`, `ppo`, `vae`)., and one other error (`sst2`). Overall, I've uncovered three issues that should be fixed / investigated further:

1. `conv_general_dilated` returns only zeros when `enabled_xla=False` and padding has some specific values. This indicates a bug in the `impl_no_xla.py` code. (See TODO in `all_examples.py`, in the definition of `ModuleSpec` for the `ppo` example.)

2. We sometimes get a concatenation error on TFLite, but not in JAX. This also seems like a bug or a limitation of TFLite. (See `TODO` in `all_examples.py`, in the definition of `ModuleSpec` for the `seq2seq` example.)

3. `flax.linen.Dense` behaves unexpectedly depending on the input values. This is the strangest case and seems most difficult to debug. As a result, `vae` and `lm1b` are either failing or not depending on the value of the inputs.


jax/experimental/jax2tf/examples_eval/converters.py
```diff
@@ -66,4 +82,39 @@ def jax2tf_to_tflite(module: examples_converter.ModuleToConvert):
       tf.lite.OpsSet.TFLITE_BUILTINS,  # enable TensorFlow Lite ops.
       tf.lite.OpsSet.SELECT_TF_OPS  # enable TensorFlow ops.
   ]
-  converter.convert()
+  # Convert the model.
+  tflite_model = converter.convert()
```
gnecula, line 86: I suggest running also the TF version on the same data, to identify if the discrepancy is due to jax2tf or TFLite.

marcvanzee: I've added the numerical comparison, and indeed the failures are due to jax2tf. This means we have some bugs in our conversion code.

jax/experimental/jax2tf/examples_eval/all_examples.py
```diff
@@ -117,8 +117,12 @@ def _flax_examples():
       'ppo':
           ModuleSpec(
               module_path='ppo.models.ActorCritic',
-              input_shape=(1, 8, 8, 4),
-              module_kwargs=dict(num_outputs=2)),
+              # TODO(marcvanzee): We get numerical differences if we run this
```
gnecula, line 120: I am wondering why not let it fail? I.e., why adjust the inputs so that it passes?

I am also wondering if we should try to run by hand multiple valid combinations of parameters, maybe there are more failures?

jax/experimental/jax2tf/examples_eval/converters.py
```diff
@@ -46,6 +51,35 @@ def jax2tf_to_tfjs(module: examples_converter.ModuleToConvert):
...
+    return np.random.randint(0, 100, size=shape, dtype=dtype)
+  elif np.issubdtype(dtype, np.floating):
+    return np.array(np.random.uniform(size=shape), dtype=dtype) * 100
+  elif dtype == np.bool:
+    return np.random.choice(a=[False, True], size=shape)
+  else:
+    raise ValueError(f"Unsupported dtype for numerical comparison: {dtype}")
+
+
+def _all_close(jax_results: Tuple[Array, ...], tf_results: Tuple[Array, ...],
```
gnecula, line 66: We should not need to write this. Perhaps all that is needed is to do a tree_map over the tf_results to apply `x.numpy()` to the results, and then use np.allclose?

jax/experimental/jax2tf/examples_eval/converters_results.md
```diff
@@ -16,11 +16,11 @@ These exampls are representative for what the average ML researcher is intereste
 | lm1b | SUCCESS |
 | mnist | SUCCESS |
 | nlp_seq | SUCCESS |
-| pixelcnn++ | SUCCESS |
+| pixelcnn++ | FAIL | ValueError('Numerical difference JAX vs TF: JAX result=[[[-1.42588705e-01  4.43906128e-01 -4.43267524e-01 ...  2.41633713e-01\n    4.30841953e-01 -3.73984545e-01]\n  [-5.50750673e-01  8.27922702e-01 -5.37640810e-01 ...  7.13608935e-02\n    5.00771403... (CROPPED)
```
gnecula, line 19: I am curious if these discrepancies arise also with enable_xla=True.

jax/experimental/jax2tf/examples_eval/all_examples.py
marcvanzee: I changed it to these input shape since these were the ones used by the example. But I agree we can just let them fail, which will make it easier to fix them. I have also reverted the other change in shapes (in SST2) so that one is failing now as well.

As a follow-up, I think it would be useful to change `input_shape` to a list, where we could add more shapes that we want to test.

jax/experimental/jax2tf/examples_eval/converters_results.md
marcvanzee: I added a new converter that just does jax2tf with `enable_xla=True`, PTAL.

jax/experimental/jax2tf/examples_eval/all_examples.py
marcvanzee: I agree, I changed it so it is now failing again.

Multiple combinations of inputs is a good idea, I can do that as a follow-up. In the meantime I've now extended it so that it runs 5 times with different input data, which did lead to another example failing, which is a good sign.

jax/experimental/jax2tf/examples_eval/converters.py
marcvanzee: We need this because a Flax Module might return a tuple that has multiple arrays with different shapes, and if we just do `np.allclose` over the tuple this fails because it will try to wrap everything in one array, requiring all shapes are the same. So that is why I iterate over the tuples explicitely.

gnecula (approved):

PR #8524, opened 2021-11-12
fix for translation to cudaDataType

reza-amd:
fix for missing CUDA_C_16F in jaxlib/cusparse.cc type mapping


jakevdp:
Thanks, good catch!

jakevdp (approved):

PR #8525, opened 2021-11-12
Use disable_jit in named_call

rsepassi:


mattjj:
I think we actually don't want disable_jit to affect named_call. At least, since disable_jit means "run things in eager mode, without staging", and named_call doesn't do any staging, it doesn't seem like the right fit.

PR #8528, opened 2021-11-12
Make MLIR bindings build work under Bazel.

hawkinsp:
Tested on Linux and Mac, but not Windows.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -26,8 +26,12 @@ build --define=grpc_no_ares=true
 build -c opt
 
 build --config=short_logs
+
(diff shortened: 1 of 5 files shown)


.bazelrc
```diff
@@ -26,6 +26,10 @@ build --define=grpc_no_ares=true
 build -c opt
 
 build --config=short_logs
+
+build --copt=-DMLIR_PYTHON_PACKAGE_PREFIX=jaxlib.mlir.
+build:posix --copt=-fvisibility=hidden
```
yashk2810, line 31: Nit: can you move the build:posix below on line 33?

yashk2810 (approved):

PR #8535, opened 2021-11-13
[x64] clean up usage of dtypes.dtype

jakevdp:
Part of #8178

Why this change? In addressing dtype consistency across the JAX package, we have to be more careful about the use of `dtypes.dtype` and `dtypes.result_type`. This PR replaces unambiguous uses of `dtypes.dtype()` with `np.dtype()` where applicable, in order to streamline that upcoming change.

This PR should have no user-visible effect.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -6925,29 +6925,29 @@ def _rng_bit_generator_translation_rule(
   # sidestep issues with the jax_enable_x64=False configuration. As a result, we
   # need to convert u32[4] -> u64[2] here in the translation rule. However, we
   # also polymorphically allow a u64[2] for backward compatibility.
-  assert ((key_shape == (4,) and key_dtype == dtypes.dtype('uint32')) or
-          (key_shape == (2,) and key_dtype == dtypes.dtype('uint64'))), (key_shape, key_dtype)
+  assert ((key_shape == (4,) and key_dtype == np.dtype('uint32')) or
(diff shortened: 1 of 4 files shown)


mattjj:
In what way is `dtypes.dtype` different from `np.dtype`, and is that difference intentional/desirable?

IIUC `dtypes.dtype` calls `np.result_type` under the hood, but maybe that should be changed, instead of changing code not to use `dtypes.dtype`... wdyt?

jakevdp:
`dtypes.dtype` currently is a used for three purposes:

1. getting the dtype of an object (e.g. `x = np.arange(4); dt = dtypes.dtype(x)`)
2. creating a dtype from a dtype-like object (e.g. `dtypes.dtype('uint64')`)
3. getting the numpy-style promotion result of multiple objects or types (e.g. `dtypes.dtype(x, float)`)

The function's creation predates my time on the project, but my guess is that it was created for the purpose of (1), and used `np.result_type` because that generally does the right thing. I think (2) was probably an inadvertent use (because `dtypes.dtype` resembles `np.dtype`) and I suspect (3) was as well.

I think we should only use this function for (1), because (2) is better accomplished by the more direct method, and (3) should generally use jax's promotion rules rather than numpy's promotion rules.

This PR is the first step: removing all uses of (1) within the package.

jakevdp:
I'm deliberately doing these changes in small chunks, starting with things that have very little user impact, rather than one giant PR that touches every aspect of the JAX package... but maybe this chunk is too small and loses the context?

mattjj (approved):
Thanks so much for the clear explanation in your comment!

PR #8546, opened 2021-11-15
update readme to focus on most active libraries

mattjj:


README.md
```diff
@@ -497,19 +497,16 @@ source](https://jax.readthedocs.io/en/latest/developer.html#building-from-source
...
-
-DeepMind has open-sourced an
-[ecosystem of libraries around JAX](https://deepmind.com/blog/article/using-jax-to-accelerate-our-research) including [Haiku](https://github.com/deepmind/dm-haiku) for neural
-network modules, [Optax](https://github.com/deepmind/optax) for gradient
-processing and optimization, [RLax](https://github.com/deepmind/rlax) for RL
-algorithms, and [chex](https://github.com/deepmind/chex) for reliable code and
-testing. (Watch the NeurIPS 2020 JAX Ecosystem at DeepMind talk [here](https://www.youtube.com/watch?v=iDxJxIyzSiM))
+[Flax](https://github.com/google/flax).
+
+DeepMind has open-sourced an [ecosystem of libraries around
```
skye, line 502: Maybe add "In addition, DeepMind has..."? This seems a little sudden.

README.md
```diff
@@ -497,19 +497,16 @@ source](https://jax.readthedocs.io/en/latest/developer.html#building-from-source
...
+JAX](https://deepmind.com/blog/article/using-jax-to-accelerate-our-research)
+including [Haiku](https://github.com/deepmind/dm-haiku) for neural network
+modules, [Optax](https://github.com/deepmind/optax) for gradient processing and
```
skye, line 505: Is Optax used by flax yet?

skye (approved):

README.md
mattjj: Done.

README.md
mattjj: I'm not sure the status... if so, would you suggest rearranging this?

I might leave that for a follow up PR :D

avital: Yes Flax uses Optax only now for all but one of our examples that hasn't been fully ported yet (PixelCNN). But that's because of some numerical equivalence that's off. Training PixelCNN takes 4 days so it's hard to diagnose. But zooming out, we explicitly tell all of our users to use Optax and have a thin pattern that we call `TrainState` that ergonomically glues them together: https://github.com/google/flax/blob/main/docs/flip/1009-optimizer-api.md

PR #8558, opened 2021-11-16
[shape_poly] Refactor shape_poly to be independent of TF.

gnecula:
Separated out into shape_poly_tf.py the TF-specific parts of shape polymorphism (essentially using tf.shape to get
the actual shape, along with tf.{add,subtract,multiply} to evaluate shape polynomials.

This opens up the possibility of using shape polymorphism independent of TF.


gnecula:
Drop in favor of #8569

PR #8560, opened 2021-11-16
[sparse] add BCOO.sum_duplicates() with nse option

jakevdp:
This replaces the existing private `_dedupe()` method with a public `sum_duplicates()` method (the name of this function follows precedent in scipy and pydata/sparse).

The default behavior is not JIT-compatible, so it adds an `nse` parameter that allows the user to specify the desired size of the output representation.

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -16,6 +16,7 @@
 import functools
 import operator
 from typing import Any, NamedTuple, Sequence, Tuple
+import warnings
 
 import numpy as np
 
@@ -23,6 +24,7 @@ from jax import core
 from jax import lax
 from jax import tree_util
 from jax import vmap
+from jax.errors import NonConcreteBooleanIndexError
 from jax.interpreters import batching
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -69,25 +71,26 @@ def _bcoo_nse(mat, n_batch=0, n_dense=0):
...
     return data, indices
-  indices_unique, inv_idx = jnp.unique(indices, axis=0, return_inverse=True,
-                                       size=nse, fill_value=jnp.array(shape))
-  data_unique = jnp.zeros_like(data).at[inv_idx].add(data)
+  try:
+    indices_unique, inv_idx = jnp.unique(indices, axis=0, return_inverse=True,
+                                         size=nse, fill_value=jnp.array(shape))
+  except NonConcreteBooleanIndexError:
+    raise ValueError("In a traced context, calling BCOO.sum_duplicates() "
+                     "requires passing a non-None value for the nse argument.")
```
froystig, line 91: Echoing @mattjj's comment in https://github.com/google/jax/pull/8305/files#r733105418, this is another case where ideally we would avoid referring to tracing, although I'm not sure what to say instead.

froystig (approved):

jax/experimental/sparse/bcoo.py
jakevdp: Maybe "When used with JIT, vmap, or another transform..."

PR #8565, opened 2021-11-16
[sparse] spdot_general: implement more cases

jakevdp:
This PR implements more cases of dot_general between sparse matrices. It is built on commits from #8567 and #8584; I'll rebase once those are merged.


froystig (approved):
Really nice work on this!

PR #8594, opened 2021-11-18
Delete the XLA in Python notebook.

hawkinsp:
Its tests are failing, and it describes a non-public API that we are phasing out.


llCurious:
Hey, just wonder has this issue been fixed that XLA custom_call can be used in Python. If so, could your please give the link to official docs, thanks!

hawkinsp:
XLA custom calls can be used in Python: 
e.g.,:
https://github.com/google/jax/blob/4dd1f001c626eb15f1a8deac58d97b578a1bd85c/jaxlib/lapack.py#L74

That's an example of Python code building a CustomCall that calls a C++ function
(https://github.com/google/jax/blob/4dd1f001c626eb15f1a8deac58d97b578a1bd85c/jaxlib/lapack.cc#L132).

If you want a CustomCall that calls back to a Python function, I recommend looking at this test case:
https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/python/xla_client_test.py;l=388
That currently only works on CPU but we intend to make it work on other platforms also.

llCurious:
I have just tried the CustomCallWithLayout as you said. However, when i try to register this customcall as XLA translation rules using `xla.register_translation(xla_pim, xla_translation, platform='cpu')`, there is the below error msg `TypeError: xla_translation() takes 2 positional arguments but 4 were given`


I follow the doc in https://jax.readthedocs.io/en/latest/notebooks/How_JAX_primitives_work.html#xla-compilation-rules to build custom call for jax. And the problem is the XLA custom call fails.

```Python
def xla_translation(c, x):
  return xla_client.ops.CustomCallWithLayout(
      builder=c,
      call_target_name=b'custom',
      # The inputs:
      operands=(xc,),
      # The input shapes:
      operand_shapes_with_layout=(
          xla_client.Shape.array_shape(jnp.dtype(jnp.float32), (), ()),
      ),
      # The output shapes:
      shape_with_layout=xla_client.Shape.array_shape(
          jnp.dtype(jnp.float32), (), ()),
  ) 
xla.register_translation(xla_pim, xla_translation, platform='cpu')

# use jit to compute 
print(jax.jit(prim_func)(2.))

```

BTW, if in this way, the code works.
```Python
xla.backend_specific_translations['cpu'][xla_pim] = xla_translation
print(jax.jit(prim_func)(2.))

```

I wonder if my xla register is wrong. Sometimes the input c shall be XlaBulider, sometimes shall be TranslationContext?

ayaka14732:
It's 2024 and I am trying to figure out the same thing that is done in this notebook

PR #8601, opened 2021-11-18
make vmap ppermute consistent with pmap/docstring

mattjj:
This was a bad bug! Unfortunately our tests didn't catch it, in part because permutations on size-two axes are special. The simplest test might have a size-three axis.

I briefly audited our other primitives for similar lack of vmap test coverage, by looking at `jax.interpreters.batching.axis_index_batchers.keys()` and comparing to our tests. I think this was the only uncovered spot.

### DIFF
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -1890,6 +1890,26 @@ class VmapPmapCollectivesTest(jtu.JaxTestCase):
     self.assertAllClose(f(jax.pmap, jax.vmap)(x, x), y)
     self.assertAllClose(f(jax.vmap, jax.pmap)(x, x), y)
 
+  @parameterized.named_parameters(
+      {"testcase_name": "_collective={}".format(collective.__name__).replace(" ", ""),
+       "collective": collective}
+      for collective in [lax.psum, lax.pmean, lax.pmax, lax.pmin])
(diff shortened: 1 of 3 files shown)


apaszke (approved):

tests/pmap_test.py
```diff
@@ -1890,6 +1890,26 @@ def f(x, y):
...
+        return x + collective(x.dot(y), ('i', 'j'))
+      return f
+
+    if jax.device_count() < 8:
+      raise SkipTest("test requires at least eight devices")
+    x = jnp.arange(4*2*64*64).reshape(4, 2, 64, 64)
+    y = f(jax.pmap, jax.pmap)(x, x)
+    self.assertAllClose(f(jax.vmap, jax.vmap)(x, x), y)
+    self.assertAllClose(f(jax.pmap, jax.vmap)(x, x), y)
+    self.assertAllClose(f(jax.vmap, jax.pmap)(x, x), y)
```
apaszke, line 1911: How is this test different to the one above?

mattjj:
Re: how the new test is different, I meant to add this comment when I requested the review, but I messed up and didn't actually post it:

image

So the difference is just that we're mapping an axis of size different from 2. It's not really relevant; just something I felt like adding in case axis size 2 ever acts special.

PR #8625, opened 2021-11-19
Implement SciPy's RegularGridInterpolator

### LINKED ISSUE #8572: Implement SciPy's RegularGridInterpolator
I recently required a fast i.e. jit-able version of SciPy's RegularGridInterpolator and adapted the original code in SciPy to work with JAX, see [this RegularGridInterpolator gist](https://gist.github.com/Edenhofer/f248f0de5a1dce54a246375d53345bc5). Is there any interest in putting a tidier version of this into JAX? If so I would be more than happy to file a merge request.

If there is interest in having this in JAX what should be done about sanity checks within the code. Code that performs such checks needs to run during the initialization of the class. However, JAX's omni-staging staging might disallow probing the truthfullness of any such checks if jitted. Should code blocks such as the following thus simply remain in numpy?
```python
if not np.all(np.diff(p) > 0.):
    raise ValueError(f"the points in dimension {i} must be strictly ascending")
```

Edenhofer:
Resolves #8572 .

### DIFF
--- a/jax/_src/third_party/scipy/interpolate.py
+++ b/jax/_src/third_party/scipy/interpolate.py
@@ -1,0 +1,158 @@
+from itertools import product
+import scipy.interpolate as osp_interpolate
+
+from jax._src.tree_util import register_pytree_node
+from jax._src.numpy.lax_numpy import (_check_arraylike, _promote_dtypes_inexact,
+                                      asarray, broadcast_arrays, can_cast,
+                                      empty, nan, searchsorted, where, zeros)
+from jax._src.numpy.util import _wraps
+
+
+def _ndim_coords_from_arrays(points, ndim=None):
+  """Convert a tuple of coordinate arrays to a (..., ndim)-shaped array."""
+  if isinstance(points, tuple) and len(points) == 1:
+    # handle argument tuple
+    points = points[0]
+  if isinstance(points, tuple):
+    p = broadcast_arrays(*points)
+    for p_other in p[1:]:
+      if p_other.shape != p[0].shape:
+        raise ValueError("coordinate arrays do not have the same shape")
+    points = empty(p[0].shape + (len(points),), dtype=float)
+    for j, item in enumerate(p):
+      points = points.at[..., j].set(item)
+  else:
+    _check_arraylike("_ndim_coords_from_arrays", points)
+    points = asarray(points)  # SciPy: asanyarray(points)
+    if points.ndim == 1:
+      if ndim is None:
+        points = points.reshape(-1, 1)
+      else:
+        points = points.reshape(-1, ndim)
+  return points
+
+
+@_wraps(
+    osp_interpolate.RegularGridInterpolator,
+    lax_description="""
+In the JAX version, `bounds_error` defaults to and must always be `False` since no
+bound error may be raised under JIT.
+
+Furthermore, in contrast to SciPy no input validation is performed.
+""")
+class RegularGridInterpolator:
+  # Based on SciPy's implementation which in turn is originally based on an
+  # implementation by Johannes Buchner
+
+  def __init__(self,
+               points,
+               values,
+               method="linear",
+               bounds_error=False,
+               fill_value=nan):
+    if method not in ("linear", "nearest"):
+      raise ValueError(f"method {method!r} is not defined")
+    self.method = method
+    self.bounds_error = bounds_error
+    if self.bounds_error:
+      raise NotImplementedError("`bounds_error` takes no effect under JIT")
+
+    _check_arraylike("RegularGridInterpolator", values)
+    if len(points) > values.ndim:
+      ve = f"there are {len(points)} point arrays, but values has {values.ndim} dimensions"
+      raise ValueError(ve)
+
+    values, = _promote_dtypes_inexact(values)
+
+    if fill_value is not None:
+      _check_arraylike("RegularGridInterpolator", fill_value)
+      fill_value = asarray(fill_value)
+      if not can_cast(fill_value.dtype, values.dtype, casting='same_kind'):
+        ve = "fill_value must be either 'None' or of a type compatible with values"
+        raise ValueError(ve)
+    self.fill_value = fill_value
+
+    # TODO: assert sanity of `points` similar to SciPy but in a JIT-able way
+    _check_arraylike("RegularGridInterpolator", *points)
+    self.grid = tuple(asarray(p) for p in points)
+    self.values = values
+
+  @_wraps(osp_interpolate.RegularGridInterpolator.__call__, update_doc=False)
+  def __call__(self, xi, method=None):
+    method = self.method if method is None else method
+    if method not in ("linear", "nearest"):
+      raise ValueError(f"method {method!r} is not defined")
+
+    ndim = len(self.grid)
+    xi = _ndim_coords_from_arrays(xi, ndim=ndim)
+    if xi.shape[-1] != len(self.grid):
+      raise ValueError("the requested sample points xi have dimension"
+                       f" {xi.shape[1]}, but this RegularGridInterpolator has"
+                       f" dimension {ndim}")
+
+    xi_shape = xi.shape
+    xi = xi.reshape(-1, xi_shape[-1])
+
+    indices, norm_distances, out_of_bounds = self._find_indices(xi.T)
+    if method == "linear":
+      result = self._evaluate_linear(indices, norm_distances)
+    elif method == "nearest":
+      result = self._evaluate_nearest(indices, norm_distances)
+    else:
+      raise AssertionError("method must be bound")
+    if not self.bounds_error and self.fill_value is not None:
+      bc_shp = result.shape[:1] + (1,) * (result.ndim - 1)
+      result = where(out_of_bounds.reshape(bc_shp), self.fill_value, result)
+
+    return result.reshape(xi_shape[:-1] + self.values.shape[ndim:])
+
+  def _evaluate_linear(self, indices, norm_distances):
+    # slice for broadcasting over trailing dimensions in self.values
+    vslice = (slice(None),) + (None,) * (self.values.ndim - len(indices))
+
+    # find relevant values
+    # each i and i+1 represents a edge
+    edges = product(*[[i, i + 1] for i in indices])
+    values = asarray(0.)
+    for edge_indices in edges:
+      weight = asarray(1.)
+      for ei, i, yi in zip(edge_indices, indices, norm_distances):
+        weight *= where(ei == i, 1 - yi, yi)
+      values += self.values[edge_indices] * weight[vslice]
+    return values
+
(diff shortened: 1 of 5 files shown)


jakevdp:
Hi - sorry we missed this PR. For third_party code like this with license restrictions, we generally put it in the third party directory: https://github.com/google/jax/tree/main/jax/_src/third_party

That said, we like to avoid this if at all possible, so if this functionality can be re-implemented from scratch without reference to the scipy source code, that would be a better solution.

Edenhofer:
To behave exactly the same as SciPy, the only viable option - at least to me - seems to be to re-implement their code while looking at it which probably does not get rid of the license.

Going merely by their documentation, the function would probably fail for certain input types for which SciPy would do something. To be specific I am thinking e.g. about [this casting here](https://github.com/google/jax/pull/8625/files#diff-861d74811ab9bbd0f32dda8436e87d44a5f149e74d8102955cca6645fdc5577aR94). In my first attempt to re-implement a N-dimensional interpolator I did not implement such a casting scheme. I would now only do so because I looked at their code and know that they do such a thing.

I think the underlying question here is what JAX wants: Is there a preference for Google having the copyright but potentially not being 1:1 compatible with SciPy or being fully compatible but not having the copyright (SciPy is licensed under the BSD license so there should be no friction).

jakevdp:
That makes sense - I think using the current implementation (and moving it into third_party) would be the best approach here.

Edenhofer:
Ok, I will probably resort to that then.

On a related note, some of these object oriented interfaces feel a little bit odd in JAX. For the very specific case of `RegularGridInterpolator` the SciPy design perfectly fits my implementation but I can easily imagine someone wanting to jit-compile or differentiate w.r.t. the interpolated values. With the current approach doing so feels awkward.

Hence my question: What do you think about implementing a call named e.g. `regular_grid_interpolator` as a purely functional implementation of `RegularGridInterpolator.__call__` and letting `RegularGridInterpolator` be a simple partial evaluation of said function? Partial evaluation should work since the SciPy API does not specify any additional properties of `RegularGridInterpolator`. If I would write such a function from scratch it would probably also be sufficiently distinct to the one in SciPy that we do not have to put it into the third_party directory. All of this of course only makes sense if `regular_grid_interpolator` would be exposed in JAX somewhere (is `jax.lax` the right place for that?).

More generally, what do you think about this as a general scheme to implement some of SciPy's object oriented interfaces?

jakevdp:
Another option would be to make `RegularGridInterpreter` a pytree, then it could be JIT-compiled directly (see https://jax.readthedocs.io/en/latest/pytrees.html#extending-pytrees).

Edenhofer:
One would get the same benefits from a partial evaluation as well (using `jax.tree_util.Partial`). However, if you prefer the pytree approach, I am happy to register `RegularGridInterpreter`.

jakevdp:
Well, I think we want the class-based API for ease of conversion between scipy and jax, but it would be fine with me to also provide a functional API as a thin wrapper (or vice versa)

Edenhofer:
I finally had some time to tackle the remaining tasks. Most importantly I added a simple test to assert that the SciPy version yields the same result as the re-implementation in JAX and registered the interpolator as pytree.

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+  if isinstance(points, tuple):
+    p = broadcast_arrays(*points)
+    for p_other in p[1:]:
+      if p_other.shape != p[0].shape:
+        raise ValueError("coordinate arrays do not have the same shape")
+    points = empty(p[0].shape + (len(points),), dtype=float)
+    for j, item in enumerate(p):
+      points = points.at[..., j].set(item)
+  else:
+    points = asarray(points)  # SciPy: asanyarray(points)
```
jakevdp, line 24: Please call `_check_arraylike()` on `points` before `asarray`

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+class RegularGridInterpolator:
+  # Based on SciPy's implementation which in turn is originally based on an
+  # implementation by Johannes Buchner
+
+  def __init__(self,
+               points,
+               values,
+               method="linear",
+               bounds_error=False,
+               fill_value=nan):
```
jakevdp, line 50: Call `_check_arraylike()` on `values` here; from its use below it seems that it must be an array.

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+    if len(points) > values.ndim:
+      ve = f"there are {len(points)} point arrays, but values has {values.ndim} dimensions"
+      raise ValueError(ve)
+
+    # Promote `values` such that RegularGridInterpolator.__call__ is always
+    # jit-able
+    values, = _promote_dtypes_inexact(values)
+
+    if fill_value is not None:
+      fill_value = asarray(fill_value)
```
jakevdp, line 67: Call `_check_arraylike` before `asarray`.

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+      if not can_cast(fill_value.dtype, values.dtype, casting='same_kind'):
+        ve = "fill_value must be either 'None' or of a type compatible with values"
+        raise ValueError()
```
jakevdp, line 70: empty ValueError?

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+      raise AssertionError("method must be bound")
+    if not self.bounds_error and self.fill_value is not None:
+      bc_shp = result.shape[:1] + (1,) * (result.ndim - 1)
+      result = where(out_of_bounds.reshape(bc_shp), self.fill_value, result)
+
+    return result.reshape(xi_shape[:-1] + self.values.shape[ndim:])
+
+  def _evaluate_linear(self, indices, norm_distances):
+    from itertools import product
```
jakevdp, line 107: Move import to top of module

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,155 @@
...
+      norm_distances.append((x - g[i]) / (g[i + 1] - g[i]))
+      if not self.bounds_error:
+        out_of_bounds += x < g[0]
+        out_of_bounds += x > g[-1]
+    return indices, norm_distances, out_of_bounds
+
+
+register_pytree_node(
+    RegularGridInterpolator,
+    lambda obj: ((obj.grid, obj.values, obj.bounds_error, obj.fill_value),
```
jakevdp, line 151: `bounds_error` and `method` must be static (because `__init__()` has logic that branches on their value), so they should be part of `aux_data`.

jakevdp (changes requested):
Looking good! I especially like the test.

A few requested changes below...

jax/_src/third_party/scipy/interpolate.py
Edenhofer: Will do!

Just out of curiosity since I wasn't able to quickly find the reason in the code: Is this for better error messages or are there cases where `asarray` works but does not yield the expected array?

jax/_src/third_party/scipy/interpolate.py
Edenhofer: arghhh; there is a beautiful error message (`ve`) just above but I did not use it :( Thanks!

jax/_src/third_party/scipy/interpolate.py
Edenhofer: `method` should already be in `aux_data` and I can move `bounds_error` there as well. Currently, one could even drop `bounds_error` because it must always be `False`.

jax/_src/third_party/scipy/interpolate.py
jakevdp: It's mainly to prevent silent slowdowns if users pass Python sequences to JIT-compiled functions; some context in #7737

Edenhofer:
I incorporated all requested changes. Is there anything else that need some polishing?

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,157 @@
...
+               method="linear",
+               bounds_error=False,
+               fill_value=nan):
+    if method not in ("linear", "nearest"):
+      raise ValueError(f"method {method!r} is not defined")
+    self.method = method
+    self.bounds_error = bounds_error
+    if self.bounds_error:  # TODO after #8468 this might be feasible
+      raise NotImplementedError("`bounds_error` takes no effect under JIT")
+
```
jakevdp, line 59: I'd also call `_check_arraylike("RegularGridInterpolator", *points)` to check that each entry in points is an array.

jax/_src/third_party/scipy/interpolate.py
```diff
@@ -0,0 +1,157 @@
...
+    _check_arraylike("RegularGridInterpolator", values)
+    if len(points) > values.ndim:
+      ve = f"there are {len(points)} point arrays, but values has {values.ndim} dimensions"
+      raise ValueError(ve)
+
+    # Promote `values` such that RegularGridInterpolator.__call__ is always
+    # jit-able
+    values, = _promote_dtypes_inexact(values)
```
jakevdp, line 67: I don't understand this comment – how does converting to an inexact type relate to JIT-compatibility?

jax/scipy/interpolate/__init__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2018 Google LLC
```
jakevdp, line 1: 2022

tests/scipy_interpolate.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2018 Google LLC
```
jakevdp, line 1: 2022

jakevdp (changes requested):
Thanks! Just a couple more things...

jax/_src/third_party/scipy/interpolate.py
Edenhofer: I think this comment is just a remnant from an older version; I'll remove it.

jakevdp (approved):

Edenhofer:
Is there anything I can or still need to do to get this merged soon?

jakevdp:
Sorry - the merge was halted by an internal lint error, and it fell off my radar. I'll work to remedy that. Thanks for your patience!

Edenhofer:
No problem! Hope that everything went smoothly now :)

ro0mquy:
`jax.scipy.interpolate` is not documented on the website:

https://jax.readthedocs.io/en/latest/jax.scipy.html

PS: Thank you for your work on this everyone.

PR #8629, opened 2021-11-20
[x64] regularize dtype helpers

jakevdp:
Part of #8178; followup to #8535.

Background: we currently have two functions related to getting dtypes from values, that have overlapping but sublely different uses:

- `jax._src.dtypes.dtype` is an **internal** function that currently takes a single value or dtype-like object, and returns the **non-canonicalized** dtype, defaulting to numpy types for scalar inputs.
- `jax.dtypes.result_type` is a **public** function (also available via `jax.numpy.result_type`) which takes any number of input values or dtype-like objects and returns the **canonicalized** dtype (for a single input) or the **canonicalized** jax type promotion result (for multiple inputs)

Frequently in the JAX package, we alias a local helper called `dtype` or `_dtype` to `result_type`, which introduces implicit type promotion, and canonicalization in places where it's not necessarily desired.

This PR improves the situation by changing the API of `jax._src.dtypes.dtype` to explicitly use JAX defaults for scalar values (it already "accidentally" does this by virtue of its call to `np.result_type`), and to take an optional `canonicalize` argument. We then can alias local dtype helpers to this rather than to `result_type`. In cases where type promotion of multiple arguments is desired, we call `result_type` explicitly.

These changes should have no user-visible effects currently in either X32 or X64 mode, but cleaning this up makes the X64 deprecation process in  #8178 more straightforward as it can be done orthogonally to these changes.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -88,6 +88,8 @@ from .._src.config import (flags, config, bool_env, disable_jit as _disable_jit,
 
 traceback_util.register_exclusion(__file__)
 
+_dtype = partial(dtypes.dtype, canonicalize=True)
+
 AxisName = Any
 
 # These TypeVars are used below to express the fact that function types
@@ -1018,7 +1020,7 @@ def value_and_grad(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
       ans, vjp_py, aux = _vjp(
           f_partial, *dyn_args, has_aux=True, reduce_axes=reduce_axes)
     _check_scalar(ans)
-    dtype = dtypes.result_type(ans)
+    dtype = _dtype(ans)
     tree_map(partial(_check_output_dtype_grad, holomorphic), ans)
     g = vjp_py(np.ones((), dtype=dtype))
     g = g[0] if isinstance(argnums, int) else g
@@ -1308,9 +1310,6 @@ def _split(x, indices, axis):
     return np.split(x, indices, axis)
   else:
     return x.split(indices, axis)
-
-def _dtype(x):
-  return dtypes.canonicalize_dtype(dtypes.result_type(x))
 
 
 def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -341,14 +341,18 @@ def is_python_scalar(x):
   except AttributeError:
     return type(x) in python_scalar_dtypes
 
-def dtype(x):
-  if type(x) in python_scalar_dtypes:
-    return python_scalar_dtypes[type(x)]
-  dt = np.result_type(x)
+def dtype(x, *, canonicalize=False):
+  """Return the dtype object for a value or type, optionally canonicalized based on X64 mode."""
+  if isinstance(x, type) and x in python_scalar_dtypes:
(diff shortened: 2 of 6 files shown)


jax/_src/api.py
```diff
@@ -85,6 +85,8 @@
 
 traceback_util.register_exclusion(__file__)
 
+_dtype = partial(dtypes.dtype, canonicalize=True)
```
gnecula, line 88: Is it really worth having this alias. It seems that there is a single use in this file and I assume no uses in other files.

jax/_src/dtypes.py
```diff
@@ -341,14 +341,18 @@ def is_python_scalar(x):
   except AttributeError:
     return type(x) in python_scalar_dtypes
 
-def dtype(x):
-  if type(x) in python_scalar_dtypes:
-    return python_scalar_dtypes[type(x)]
-  dt = np.result_type(x)
+def dtype(x, *, canonicalize=False):
+  """Return the dtype of a value, optionally canonicalized based on X64 mode."""
```
gnecula, line 345: It seems that the function allows passing of types also, not just values.

jax/_src/lax/lax.py
```diff
@@ -6471,8 +6471,8 @@ def _const(example, val):
 _twos: Callable = partial(full_like, fill_value=2)
 _two: Callable = partial(full_like, shape=(), fill_value=2)
 
-dtype: Callable = dtypes.result_type
-_dtype: Callable = dtypes.result_type
+dtype: Callable = partial(dtypes.dtype, canonicalize=True)
```
gnecula, line 6474: Maybe add a TODO to clean up these aliases.

gnecula: Would it be too much to do the cleanup here also? Perhaps we can introduce `dtypes.dtype_canonical` which is `partial(dtypes.dtype, canonicalize=True)` and if we need aliases elsewhere we can use `from dtypes import dtype, dtype_canonical`

I suspect that you considered this already and perhaps it leads to too many changes?

jax/_src/api.py
jakevdp: I count 13 uses in this file, 12 of them existing before this PR... I could do that refactor if you think it would be better?

jax/_src/dtypes.py
jakevdp: Changed the docstring.

jax/_src/lax/lax.py
jakevdp: The problem with importing `dtype` is that in many cases `dtype` is the name of a parameter in the function. So we'd have to import dtype as something other name... which is not so different than the current approach.

jax/_src/api.py
jakevdp: (Followup: this comment seems similar to below - I'm not sure how much there is to be gained from removing these helpers)

jax/_src/lax/lax.py
gnecula: Do we need both `dtype` and `_dtype`? Can we keep just `_dtype`?

gnecula (approved):

jax/_src/lax/lax.py
jakevdp: We could, except `lax.dtype` is an exported name, so it has to be defined somewhere... may as well be here.

PR #8650, opened 2021-11-22
Enable QDWH TPU tests.

tlu7:
TODO for  `testQdwhWithRandomMatrix` indicates TPU failure, whereas both `gpu` and `tpu` are disabled.  Please let me know if I should disable `gpu` if there is any A100 failue.

>  TODO(tianjianlu): Fails on TPU.
  @jtu.skip_on_devices("gpu", "tpu")
  def testQdwhWithRandomMatrix(self, m, n, log_cond):

### DIFF
--- a/tests/qdwh_test.py
+++ b/tests/qdwh_test.py
@@ -13,6 +13,7 @@
 # limitations under the License
 
 """Tests for the library of QDWH-based polar decomposition."""
+import functools
 
 from jax import test_util as jtu
 from jax.config import config
@@ -28,14 +29,14 @@ from absl.testing import parameterized
 config.parse_flags_with_absl()
 _JAX_ENABLE_X64 = config.x64_enabled
 
-# Input matrix data type for PolarTest.
-_POLAR_TEST_DTYPE = np.float64 if _JAX_ENABLE_X64 else np.float32
+# Input matrix data type for QdwhTest.
+_QDWH_TEST_DTYPE = np.float64 if _JAX_ENABLE_X64 else np.float32
 
-# Machine epsilon used by PolarTest.
-_POLAR_TEST_EPS = jnp.finfo(_POLAR_TEST_DTYPE).eps
+# Machine epsilon used by QdwhTest.
+_QDWH_TEST_EPS = jnp.finfo(_QDWH_TEST_DTYPE).eps
 
-# Largest log10 value of condition numbers used by PolarTest.
-_MAX_LOG_CONDITION_NUM = np.log10(int(1 / _POLAR_TEST_EPS))
+# Largest log10 value of condition numbers used by QdwhTest.
+_MAX_LOG_CONDITION_NUM = np.log10(int(1 / _QDWH_TEST_EPS))
 
 
 def _check_symmetry(x: jnp.ndarray) -> bool:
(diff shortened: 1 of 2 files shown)


tlu7:
@hawkinsp

tests/qdwh_test.py
```diff
@@ -103,30 +106,31 @@ def testQdwhWithUpperTriangularInputAllOnes(self, m, n, log_cond):
...
+      relative_diff_u = _compute_relative_diff(actual_u, expected_u)
+      np.testing.assert_almost_equal(relative_diff_u, 1E-4, decimal=2)
 
     with self.subTest('Test h.'):
-      self.assertAllClose(actual_h, expected_h, rtol=rtol)
+      relative_diff_h = _compute_relative_diff(actual_h, expected_h)
+      np.testing.assert_almost_equal(relative_diff_h, 1E-4, decimal=2)
 
     with self.subTest('Test u.dot(h).'):
       a_round_trip = actual_u.dot(actual_h)
```
hawkinsp, line 117: Is it possible that the TPU precision problems occur because JAX currently defaults to low precision for matmuls on TPU? I think calling `.dot` will do that.

Try using the scope:
```
with jax.default_matmul_precision('float32'):
```

or passing the `precision=` argument to methods like `dot`.

tlu7:
Thanks for the offline discussion. Turned on highest precision globally in qdwh.py and at the matmul in qdwh_test.py; and saw two orders of magnitude higher accuracy in the test.

tlu7:
The test failure is related to the precision check. I am looking at it. 
[https://github.dev/google/jax/blob/b0d334a8811e5fc38a5e875e42d598c06e229c83/tests/lax_numpy_test.py#L5506-L5507](url)

tlu7:
> The test failure is related to the precision check. I am looking at it.
> 
> https://github.com/google/jax/blob/b0d334a8811e5fc38a5e875e42d598c06e229c83/tests/lax_numpy_test.py#L5506

Here are the findings:

The test failure in `lax_numpy_test` is due to `jax.config.update('jax_default_matmul_precision', 'float32')` in qdwh.py.

Removing `jax.config.update('jax_default_matmul_precision', 'float32')` in qdwh.py does not affect the test accuracy (this is most likely due to the function calls are all within the scope of `with jax.default_matmul_precision('float32'):`.

tests/qdwh_test.py
```diff
@@ -85,38 +89,45 @@ def testQdwhUnconvergedAfterMaxNumberIterations(
...
 
     with self.subTest('Test h.'):
-      self.assertAllClose(actual_h, expected_h, rtol=rtol)
+      relative_diff_h = _compute_relative_diff(actual_h, expected_h)
+      np.testing.assert_almost_equal(relative_diff_h, 1E-6, decimal=5)
 
     with self.subTest('Test u.dot(h).'):
-      a_round_trip = actual_u.dot(actual_h)
-      self.assertAllClose(a_round_trip, a, rtol=rtol)
+      with jax.default_matmul_precision('float32'):
```
hawkinsp, line 120: You might find it cleaner to simply define a helper function:
```
_dot = partial(jnp.dot, precision="highest")
```
and use it everywhere in this test file, rather than sticking the scope in many places?

tests/qdwh_test.py
```diff
@@ -85,38 +89,45 @@ def testQdwhUnconvergedAfterMaxNumberIterations(
...
+    a = jnp.triu(jnp.ones((m, n))).astype(_QDWH_TEST_DTYPE)
     u, s, v = jnp.linalg.svd(a, full_matrices=False)
     cond = 10**log_cond
     s = jnp.linspace(cond, 1, min(m, n))
     a = (u * s) @ v
     is_symmetric = _check_symmetry(a)
     max_iterations = 10
-    actual_u, actual_h, _, _ = qdwh.qdwh(a, is_symmetric, max_iterations)
+
+    with jax.default_matmul_precision('float32'):
```
hawkinsp, line 104: I would not add the scope around the call to qdwh. I would add the necessary precision overrides inside the qdwh implementation, since I'm not sure we have any evidence that QDWH works well with low matmul precision.

tlu7:
Thanks for the comments! PTAL.

hawkinsp (approved):

PR #8652, opened 2021-11-22
Document lax.GatherScatterMode.

hawkinsp:
Recommend the .at[...] property in the docstrings for lax.scatter_ operators.

Add several missing lax.scatter_ operators to the index.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1043,21 +1043,28 @@ class GatherDimensionNumbers(NamedTuple):
 
 
 class GatherScatterMode(Enum):
-  # Indices will be clamped to the nearest in-range value, i.e., such that the
-  # entire window to be gathered is in-range.
+  """
+  Describes how to handle out-of-bounds indices in a gather or scatter.
+
+  Possible values are:
+
+  CLIP:
+    Indices will be clamped to the nearest in-range value, i.e., such that the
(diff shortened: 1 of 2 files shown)


jax/_src/lax/lax.py
```diff
@@ -1222,7 +1232,10 @@ def scatter_mul(
   <https://www.tensorflow.org/xla/operation_semantics#scatter>`_, where
   multiplication is used to combine updates and values from `operand`.
 
-  The semantics of scatter are complicated and its API is subject to change.
+  The semantics of scatter are complicated, and its API might change in the
+  future. For most use cases, you should prefer the
+  :attr:`jaxlib.xla_extension.DeviceArray.at` property on JAX arrays which uses
```
jakevdp, line 1237: Elsewhere we have linked to ``` :attr:`jax.numpy.ndarray.at` ```. I think either will work (it's https://jax.readthedocs.io/en/latest/jax.numpy.html#jaxlib.xla_extension.DeviceArray.at vs. https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html#jax.numpy.ndarray.at) But let's choose one to standardize to.

Same comment elsewhere.

hawkinsp: Changed to `jax.numpy.ndarray.at`.

jakevdp (approved):

PR #8658, opened 2021-11-23
Improve error when jnp.arange is used with non-constant arguments

gnecula:


jakevdp (approved):
Looks good!

jax/_src/numpy/lax_numpy.py
```diff
@@ -3761,12 +3761,16 @@ def arange(start: core.DimSize, stop: Optional[core.DimSize]=None,
   lax._check_user_dtype_supported(dtype, "arange")
   require = partial(core.concrete_or_error, None)
   msg = "It arose in jax.numpy.arange argument `{}`.".format
+  if _any([core.is_special_dim_size(d) for d in (start, stop, step)]):
```
jakevdp, line 3764: nit, remove the square brackets and call `_any` on the raw iterator

gnecula: Done

PR #8661, opened 2021-11-23
[x64] Make autodiff respect weak types

jakevdp:
Came up in the process of working on #8178.

The issue this solves is the fact that, currently, autodiff results in functions that treat inputs as strongly-typed:
```python

In [2]: from jax import grad

In [3]: jnp.sin(0.0)
Out[3]: DeviceArray(0., dtype=float32, weak_type=True)

In [4]: jnp.cos(0.0)
Out[4]: DeviceArray(1., dtype=float32, weak_type=True)

In [5]: grad(jnp.sin)(0.0)
Out[5]: DeviceArray(1., dtype=float32)  # <------- not weakly typed!
```
This makes it easy for 64-bit values to slip into otherwise 32-bit computations in a post-X64 world.

### DIFF
--- a/jax/_src/abstract_arrays.py
+++ b/jax/_src/abstract_arrays.py
@@ -37,8 +37,9 @@ def make_shaped_array(x):
   return ShapedArray(np.shape(x), dtype)
 
 def zeros_like_array(x):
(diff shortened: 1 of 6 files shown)


mattjj (approved):
Nice!

PR #8665, opened 2021-11-23
tweak implementation of lax._array_copy (underlying jnp.array(x, copy=True))

mattjj:
The first commit made a small change to simplify the `lax._array_copy` implementation to use a nested call.

The second (at @hawkinsp's wise suggestion) switched the implementation to be a primitive.


jax/_src/lax/lax.py
```diff
@@ -5594,6 +5576,20 @@ def _rng_bit_generator_named_shape_rule(key, *, shape, dtype, algorithm):
...
+def _array_copy(arr):
+  return copy_p.bind(arr)
+
+# The copy_p primitive exists for expressing making copies of runtime arrays.
+# For that reason we don't simplify it out of jaxprs (e.g. for jit invariance).
+# It's used in jnp.array(x, copy=True), which is the user-facing API.
+copy_p = core.Primitive('copy_p')
+copy_p.def_impl(partial(xla.apply_primitive, copy_p))
+copy_p.def_abstract_eval(lambda x: x)
+xla.register_translation(copy_p, lambda ctx, avals_in, avals_out, x: [x])
```
hawkinsp, line 5588: mlir.register_lowering as well.

hawkinsp (approved):

jax/_src/lax/lax.py
mattjj: DOne.

PR #8667, opened 2021-11-23
custom pp_eqn rules, simpler xla_call print

mattjj:
Add a simple internal mechanism for tweaking `pp_eqn` rules on a primitive-by-primitive basis. Add one such rule for `xla_call_p` so as to suppress printing of default arg values.

```python
import jax
jaxpr = jax.make_jaxpr(jax.jit(lambda x: x))(1.)
print(jaxpr)
```

Before:

```
{ lambda ; a:f32[]. let
    b:f32[] = xla_call[
      backend=None
      call_jaxpr={ lambda ; c:f32[]. let  in (c,) }
      device=None
      donated_invars=(False,)
      inline=False
      name=<lambda>
    ] a
  in (b,) }
```

After:

```
{ lambda ; a:f32[]. let
    b:f32[] = xla_call[
      call_jaxpr={ lambda ; c:f32[]. let  in (c,) }
      name=<lambda>
    ] a
  in (b,) }
```

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -76,12 +76,14 @@ class Jaxpr:
     self.eqns = list(eqns)
 
   def __str__(self):
-    return str(pp_jaxpr(self, JaxprPpContext()))
+    return str(pp_jaxpr(self, JaxprPpContext(), custom_pp_eqn_rules=True))
   __repr__ = __str__
 
-  def pretty_print(self, *, source_info=False, print_shapes=True, **kw):
+  def pretty_print(self, *, source_info=False, print_shapes=True,
+                   custom_pp_eqn_rules=True, **kw):
     doc = pp_jaxpr(self, JaxprPpContext(), source_info=source_info,
-                   print_shapes=print_shapes)
+                   print_shapes=print_shapes,
+                   custom_pp_eqn_rules=custom_pp_eqn_rules)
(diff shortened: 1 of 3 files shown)


jax/core.py
```diff
@@ -2122,12 +2122,16 @@ def pp_eqn(eqn, context: JaxprPpContext, *, print_shapes=True, source_info=False
...
   annotation = (source_info_util.summarize(eqn.source_info)
                 if source_info else None)
-  return pp.concat([
-    lhs, pp.text(" = ", annotation=annotation), pp.text(eqn.primitive.name),
-    pp_kv_pairs(sorted(eqn.params.items()), context),
-    pp.text(" ") + pp_vars(eqn.invars, context)
-  ])
-
+  rule = pp_eqn_rules.get(eqn.primitive)
+  if rule:
```
hawkinsp, line 2126: I would be tempted to add a flag to turn off special pretty-printing to the printer. It may be useful sometimes to see the unperturbed generic form.

hawkinsp (approved):

jax/core.py
mattjj: Good idea.

mattjj: Done.

PR #8678, opened 2021-11-24
[hcb] Simplifications to the host_calback API

gnecula:
* dropping support for special AD handling for hcb.id_tap and id_print.
  From now on, only the primals are tapped.

This allows us to make some significant cleanup in the internals.


shoyer:
I guess the idea is that users should host_callback inside a custom_vjp if they want to capture values from the reverse pass?

gnecula:
> I guess the idea is that users should host_callback inside a custom_vjp if they want to capture values from the reverse pass?

Indeed, that is the idea. In fact, some folks already do this because they want to pre-process the gradients on the device before sending them to the host.

PR #8681, opened 2021-11-24
DOC: add FAQ section on JAX vs. Numpy performance

jakevdp:
This is a question I've seen come up a number of times (most recently in #8673); I thought it would be useful to have a definitive answer to point to.

View the rendered version here: https://jax--8681.org.readthedocs.build/en/8681/faq.html#is-jax-faster-than-numpy

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -214,6 +214,39 @@ JAX/accelerators vs NumPy/CPU. For example, if we switch this example to use
 .. _`%time and %timeit magics`: https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-time
 .. _Colab: https://colab.research.google.com/
 
+.. _faq-jax-vs-numpy:
+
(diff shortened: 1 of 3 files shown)


mattjj (commented):
I think less is more here. Can we try to cut down on the specifics, since they may change (and I'm not sure if they're telling the right story even now)?

For example, what if we just kept the first paragraph and the last paragraph, and excised the examples and detailed claims?

jakevdp:
Thanks - I made the discussion a bit more high level to address your comments - PTAL!

mattjj (approved):
I love it, beautifully explained!

mcleantom:
Is the error message after this section in the docs a mistake?

image

jakevdp:
No, that’s the heading for the next section. The content was moved several months ago but we kept the heading in order to not break existing links

jakevdp:
Addressing this strange formatting in #8968

PR #8694, opened 2021-11-25
[x64] make scan masking rule more type safe

jakevdp:
Part of #8178; this issue was revealed by `masking_test.py` when default dtypes were changed.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -2017,7 +2017,8 @@ def _scan_masking_rule(padded_vals, logical_shapes, reverse, length,
   consts, init, xs = split_list(padded_vals, [num_consts, num_carry])
   max_length, = {x.shape[0] for x in xs}
   const_linear, init_linear, xs_linear = split_list(linear, [num_consts, num_carry])
-  out_vals = scan_p.bind(dynamic_length, *consts, 0, *init, *xs,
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -2017,7 +2017,8 @@ def _scan_masking_rule(padded_vals, logical_shapes, reverse, length,
   consts, init, xs = split_list(padded_vals, [num_consts, num_carry])
   max_length, = {x.shape[0] for x in xs}
   const_linear, init_linear, xs_linear = split_list(linear, [num_consts, num_carry])
-  out_vals = scan_p.bind(dynamic_length, *consts, 0, *init, *xs,
+  dynamic_length = lax.convert_element_type(dynamic_length, dtypes.int_)
+  out_vals = scan_p.bind(dynamic_length, *consts, dtypes.int_(0), *init, *xs,
```
jakevdp, line 2021: These two values must have dtype `int_` because of the aval here: https://github.com/google/jax/pull/8694/files#diff-a8bc4c84f5715589c52d4cfbd1dd803e3f38314cf11c8b042efc5c59a8e16ab1R2041

mattjj (approved):
Nice catch! (Tangentially, this is basically dead prototype code, so if you ever need to, you can just disable tests around it.)

PR #8700, opened 2021-11-25
Fix MLIR CAPI DLL build

cloudhan:
Recent added MLIR CAPI library breaks windows build, this PR fixes it.

### DIFF
--- a/jaxlib/jax.bzl
+++ b/jaxlib/jax.bzl
@@ -33,3 +33,69 @@ flatbuffer_py_library = _flatbuffer_py_library
 
 def py_extension(name, srcs, copts, deps):
     pybind_extension(name, srcs = srcs, copts = copts, deps = deps, module_name = name)
+
+def windows_cc_shared_mlir_library(name, out, deps = [], srcs = []):
+    """Workaround DLL building issue.
+
+    1. cc_binary with linkshared enabled cannot produce DLL with symbol
+       correctly exported.
+    2. Even if the DLL is correctly built, the resulting target cannot be
+       correctly consumed by other targets.
+
+    Args:
+      name: the name of the output target
+      out: the name of the output DLL filename
+      deps: deps
+      srcs: srcs
+    """
+
+    # create a dummy library to get the *.def file
+    dummy_library_name = name + ".dummy.dll"
+    native.cc_binary(
+        name = dummy_library_name,
+        linkshared = 1,
+        linkstatic = 1,
+        deps = deps,
+    )
+
+    # .def file with all symbols, not usable
+    full_def_name = name + ".full.def"
+    native.filegroup(
+        name = full_def_name,
+        srcs = [dummy_library_name],
+        output_group = "def_file",
+    )
+
+    # filtered def_file, only the needed symbols are included
+    filtered_def_name = name + ".filtered.def"
+    filtered_def_file = out + ".def"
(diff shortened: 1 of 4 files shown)


cloudhan:
@hawkinsp The story of building shared library on Windows with Bazel is totally s**tshow nowadays. The workaround is pretty weird, pls dont try to understand it 🤣

cloudhan:
linking error in [4208386597](https://github.com/cloudhan/jax-windows-builder/runs/4208386597) 
xutility C2440 error in [4314467138](https://github.com/cloudhan/jax-windows-builder/runs/4314467138)

Built successfully in [1501825747](https://github.com/cloudhan/jax-windows-builder/actions/runs/1501825747)

hawkinsp:
Wonderful! I had a sneaking suspicion that the Windows support wouldn't be that easy. Thanks so much for the fix!

hawkinsp (approved):

jaxlib/jax.bzl
```diff
@@ -33,3 +33,69 @@ flatbuffer_py_library = _flatbuffer_py_library
...
+    )
+
+    # filtered def_file, only the needed symbols are included
+    filtered_def_name = name + ".filtered.def"
+    filtered_def_file = out + ".def"
+    native.genrule(
+        name = filtered_def_name,
+        srcs = [full_def_name],
+        outs = [filtered_def_file],
+        cmd = """echo 'LIBRARY {}\nEXPORTS ' > $@ && grep '^\\W*mlir' $(location :{}) >> $@""".format(out, full_def_name),
```
hawkinsp, line 76: I am curious though. I would have thought that the existing symbol visibility logic in the LLVM code base would have done the right thing here: only symbols marked as exported should have been exported, so why do we need the `.def` filtering stuff?

cloudhan: LLVM has `MLIR_CAPI_EXPORTED` used on [include/mlir-c/*.h](https://github.com/llvm/llvm-project/blob/7e73c2a66a8bb73b80aedc94c4e58598ac87e9d5/mlir/include/mlir-c/IR.h#L85) CAPIs. And there is a [MLIR_CAPI_BUILDING_LIBRARY](https://github.com/llvm/llvm-project/blob/30d61893fb7bbe364bf25074feaf0b178dac64e6/mlir/include/mlir-c/Support.h#L39-L43) define to control its semantics, which is defined at [AddMLIR.cmake](https://github.com/llvm/llvm-project/blob/68e2231f8724eec26c9fcf5c5330fae897f9435f/mlir/cmake/modules/AddMLIR.cmake#L409). However, in [build_defs.bzl](https://github.com/llvm/llvm-project/blob/7e73c2a66a8bb73b80aedc94c4e58598ac87e9d5/utils/bazel/llvm-project-overlay/mlir/build_defs.bzl#L28), it is not correctly defined.

In my local experiment, I indeed added the define to build_defs.bzl, with no luck. Maybe it's the missing of `cc_import` invalidate the fix. If it is the case, then the def filtering can be omitted if I patch LLVM's `build_defs.bzl` along with fixing toward [tensorflow/compiler/mlir/hlo/BUILD](https://github.com/tensorflow/tensorflow/blob/84af5b08335d9099de320bd0409e250289865507/tensorflow/compiler/mlir/hlo/BUILD#L1896-L1926).

PR #8716, opened 2021-11-29
Refactor the lowering of remat

gnecula:
Replace the XLA-specific lowering with the same logic written
with LAX primitives. This will allow reusing the logic for
different lowerings, e.g., jax2tf or MLIR.


hawkinsp:
There's another copy of this logic here: https://github.com/google/jax/blob/c1ce85203c65cd816e98422122b25320eb33539c/jax/interpreters/mlir.py#L929
which you should probably change to use the new version. It will look very similar to the copy in `xla.py`.

gnecula:
> There's another copy of this logic here:
> 
> https://github.com/google/jax/blob/c1ce85203c65cd816e98422122b25320eb33539c/jax/interpreters/mlir.py#L929
> 
> 
> which you should probably change to use the new version. It will look very similar to the copy in `xla.py`.

I saw that, but I could not yet figure out how to call the xla lowering. Is there an example? It seemed to me that all the primitives where this would be needed are done by way of fallback, instead of just interpreting the LAX function (a la `lower_fun`).

gnecula:
> You shouldn't be calling the XLA lowering, you should be using `mlir.lower_fun` to lower the traceable JAX version to MLIR directly.

Done.

jax/_src/lax/control_flow.py
```diff
@@ -2924,3 +2925,92 @@ def _cumulative_jvp_rule(primals, tangents, *, axis: int, reverse: bool,
...
+
+
+def _dummy_remat_result(aval: core.AbstractValue):
+  """A result that will be discarded"""
+  if aval is core.AbstractToken:
+    return lax.create_token()
+  elif aval is core.abstract_unit:
+    return core.unit
+  else:
+    return lax.broadcast(np.array(0, dtype=aval.dtype), aval.shape)  # type: ignore
```
hawkinsp, line 2937: lax.full_like ?

hawkinsp (approved):
Verify this works with MLIR enabled. I'm about to enable internal CI tests for this.

jax/_src/lax/control_flow.py
gnecula: For `lax.full_like` I need a value x to use the shape or dtype for. I could write `lax.full_like(None, 0, dtype=aval.dtype, shape=aval.shape)` but I think it is less clear that this is just a broadcast.

PR #8718, opened 2021-11-29
DOC: add documentation for configuration functionality

jakevdp:
Preview the new doc here: https://jax--8718.org.readthedocs.build/en/8718/jax.config.html

### DIFF
--- a/docs/jax.config.rst
+++ b/docs/jax.config.rst
@@ -1,0 +1,20 @@
+.. currentmodule:: jax
+
+JAX configuration
+=================
+
+.. autosummary::
+   :toctree: _autosummary
+
+   config
+   enable_checks
+   check_tracer_leaks
+   checking_leaks
+   enable_custom_prng
+   debug_nans
+   debug_infs
+   log_compiles
+   default_matmul_precision
+   default_prng_impl
+   numpy_rank_promotion
+   
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -22,6 +22,7 @@ Subpackages
     jax.flatten_util
     jax.dlpack
     jax.profiler
+    jax.config
 
 .. toctree::
    :hidden:


docs/jax.config.rst
```diff
@@ -0,0 +1,20 @@
...
+JAX configuration
+=================
+
+.. autosummary::
+   :toctree: _autosummary
+
+   config
+   enable_checks
+   check_tracer_leaks
+   checking_leaks
```
mattjj, line 12: This one is redundant with `check_tracer_leaks`. Maybe just have one or the other? (I'm not sure which though! @LenaMartens WDYT?)

docs/jax.config.rst
```diff
@@ -0,0 +1,20 @@
...
+   enable_custom_prng
```
mattjj, line 13: This one and `default_prng_impl` might be transient so that we don't want to document them... I'm not sure. @froystig wdyt?

docs/jax.config.rst
jakevdp: This is the full list of functionality imported into the `jax` namespace, so I'd lean toward keeping both here (unless we want to clean up the namespace imports).

docs/jax.config.rst
jakevdp: If that's the case, we should also avoid declaring them in the JAX namespace (accomplished by importing `foo` rather than importing `foo as foo`) https://github.com/google/jax/blob/5084a12034e6e88cd1fe274f78a9f58e2d109eeb/jax/__init__.py#L40-L52

docs/jax.config.rst
LenaMartens: I think I like `check_tracer_leaks` better because it's more in-line with other debug names (like `debug_nans` and `enable_checks`)? Which one can be used as a context manager again?

jakevdp: Both can be used as a context manager; `checking_leaks` is just a partial with the target set to `True`: https://github.com/google/jax/blob/5084a12034e6e88cd1fe274f78a9f58e2d109eeb/jax/_src/config.py#L475

docs/jax.config.rst
froystig: I expect `default_prng_impl` to remain indefinitely. It indicates which PRNG implementation to use if none are specified, and we plan to continue to have multiple implementations. I'd say we keep that one and document it.

`enable_custom_prng` is explicitly intended to be transient (it is an upgrade flag). It is nonetheless public, so the `jax` namespace seems appropriate. Its description text is clear regarding our intent to remove it eventually:

https://github.com/google/jax/blob/3f2a4b10f63c398d5a30a9dd11ffba704de9baf2/jax/_src/config.py#L517-L521

froystig: Can we reflect the flag's description text in the auto-generated docs view, even if manually?

jakevdp: Is this what you have in mind? https://jax--8718.org.readthedocs.build/en/8718/_autosummary/jax.enable_custom_prng.html#jax.enable_custom_prng

froystig: There it is! Yes.

I was also thinking about the the [`jax.config` page](https://jax--8718.org.readthedocs.build/en/8718/jax.config.html). If we'd like to further distinguish a setting like this one from the others, we could try to add to the first docstring line, so that it shows up on the `jax.config` summary as, say:

> Context manager for _jax_enable_custom_prng_ config option. **[transient]**

rather than the current:

> Context manager for _jax_enable_custom_prng_ config option.

But I'm also happy with things as they are in this PR.

froystig (approved):

docs/jax.config.rst
jakevdp: That summary is defined here: https://github.com/google/jax/blob/610187cd332ee18b40bc9caf90e54db50f4d0659/jax/_src/config.py#L338

We could modify that for the class in question. Alternatively, we could use a [custom template](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#customizing-templates) for the autosummary directive, but that might be a too heavy-handed here.

jakevdp: We could thread a boolean `transient` flag through all the config definition boilerplate, and then display `[transient]` in the docstring if it's set to True. What do you think?

froystig: Sounds like a fine plan. Do we have any other transient settings at the moment? @mattjj – any thoughts, given your question that started this thread?

jakevdp: Maybe the most flexible approach would be to add an `extra_description` setting that defaults to `""`.

jakevdp: Done in #8768

PR #8719, opened 2021-11-29
[x64] jnp.pad: preserve weak types of inputs

jakevdp:
Part of #8178; this was causing problems when default dtypes were changed.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3049,7 +3049,7 @@ def _check_no_padding(axis_padding, mode):
 def _pad_constant(array, pad_width, constant_values):
   nd = ndim(array)
   constant_values = broadcast_to(asarray(constant_values), (nd, 2))
-  constant_values = lax.convert_element_type(constant_values, array.dtype)
+  constant_values = lax._convert_element_type(constant_values, array.dtype, dtypes.is_weakly_typed(array))
   for i in range(nd):
     widths = [(0, 0, 0)] * nd
     widths[i] = (pad_width[i, 0], 0, 0)
@@ -3160,6 +3160,7 @@ def _pad_linear_ramp(array, pad_width, end_values):
         dtype=array.dtype,
         axis=axis
     )
+    ramp_before = lax._convert_element_type(ramp_before, weak_type=dtypes.is_weakly_typed(array))
     ramp_after = linspace(
         start=end_values[axis][1],
         stop=edge_after.squeeze(axis), # Dimension is replaced by linspace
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -3049,7 +3049,7 @@ def _check_no_padding(axis_padding, mode):
 def _pad_constant(array, pad_width, constant_values):
   nd = ndim(array)
   constant_values = broadcast_to(asarray(constant_values), (nd, 2))
-  constant_values = lax.convert_element_type(constant_values, array.dtype)
+  constant_values = lax._convert_element_type(constant_values, array.dtype, dtypes.is_weakly_typed(array))
```
mattjj, line 3052: Would it work to rewrite this to `constant_values.astype(...)` or similar? If instead we have to use `lax._convert_element_type` (with the underscore prefix) in our NumPy implementation, it makes me think we need it to be a public lax API after all... WDYT?

jakevdp: `astype` does not allow specification of weak types. The only way to convert to weak type is to use the underscored method - it previously was possible via `lax.convert_element_type`, but you removed that in #6255 😁

mattjj: Maybe we should put it back then!

But do we actually need to specify a weak type here, or can the weak type just be inherited from the input? That is, maybe `convert_element_type` shouldn't always set `weak_type=False`, if we want to be able to convert element types w/o making the type strong (as here)?

Whatever the fix is, I think lax_numpy.py is for the most part something of a model for user code consuming the lowest-level JAX APIs, so the fact that code as simple as `jnp.pad` might need to call an internal API makes me think the lax API isn't quite right, and that's where the fix should be...

mattjj: In #6255 I indeed used `lax._convert_element_type` in lax_numpy.py, though only in the `_promote_dtypes` helper functions and in `jnp.array`. I thought those were isolated edge cases which needed the `weak_type` logic to be exposed. But maybe that was just mistaken, and all of lax_numpy.py may need to access explicit `weak_type` control, as in this `jnp.pad` (unless it can be written differently so as to avoid this need)...?

I was hoping that weak types could be mostly a JAX-internal API, but it's sounding more like user code might need them...

jakevdp: Further down I was able to get around this by using `empty_like`.

We could do something similar here by doing
```python
constant_values = jnp.empty_like(array, shape=constant_values.shape).at[].set(constant_values)
```
but I think the scatter operation would be a performance hit compared to the current version, which is basically just a metadata update.

jakevdp: Oh, for the `jnp.eye` below, numpy 1.20 added a `like` argument to `np.identity` which we could use to do weak_type casting indirectly: https://numpy.org/doc/stable/reference/generated/numpy.identity.html

I think this would still have to be implemented via the underscored convert_element_type though.

mattjj (approved):
Let's merge whatever you think is best right now, though I'm still thinking that weak_type might need to be an option on lax.convert_element_type.

jakevdp:
Thanks - #8722 fixes a similar issue. I'll think more about the best way to expose an API for setting weak types.

PR #8722, opened 2021-11-29
[x64] preserve weak types in jax.scipy.sparse solvers

jakevdp:
Part of #8178; this was causing problems when default dtypes were changed.


mattjj (approved):
Beautiful!

PR #8739, opened 2021-11-30
add more detail about aux_data in pytree docs

### LINKED ISSUE #8726: Doc for register_pytree_node_class doesn't explain aux_data
The documentation [here](https://jax.readthedocs.io/en/latest/pytrees.html?highlight=register_pytree_node_class#extending-pytrees) and [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.tree_util.register_pytree_node_class.html) does not explain what `aux_data` is useful for (`aux_data` is set to `None`). Some more explanations and an example would be welcome.

mattjj:
fixes #8726

### DIFF
--- a/docs/pytrees.md
+++ b/docs/pytrees.md
@@ -208,9 +208,19 @@ show_example(Special(1., 2.))
 ```
 
 The set of Python types that are considered internal pytree nodes is extensible,
-through a global registry of types, and values of registered types are traversed
-recursively. To register a new type, you can use
-{func}`~jax.tree_util.register_pytree_node`:
+through a global registry of types. Values of registered types are traversed
+recursively, so a custom pytree type can itself contain pytrees.
+
+To register a new type, you can use {func}`~jax.tree_util.register_pytree_node`
+to specify how to flatten and unflatten instances of the new type. That is, you
+register two functions: the first function accepts instances of the new type and
+returns a pair where the first element is an iterable of children (for example,
+the values of a dictionary) and the second element is auxiliary metadata needed
+to reconstruct the instance (for example, the sorted keys of a dictionary). The
+second function goes the other way: given the auxiliary data and children, it
+constructs an instance of the new type.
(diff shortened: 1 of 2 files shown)


docs/pytrees.md
```diff
@@ -208,9 +208,19 @@ show_example(Special(1., 2.))
...
-recursively. To register a new type, you can use
-{func}`~jax.tree_util.register_pytree_node`:
+through a global registry of types. Values of registered types are traversed
+recursively, so a custom pytree type can itself contain pytrees.
+
+To register a new type, you can use {func}`~jax.tree_util.register_pytree_node`
+to specify how to flatten and unflatten instances of the new type. That is, you
+register two functions: the first function accepts instances of the new type and
+returns a pair where the first element is an iterable of children (for example,
+the values of a dictionary) and the second element is auxiliary metadata needed
```
jakevdp, line 218: Could we add some words about how you might decide what parts of an object are "children" and what are "aux_data"? My understanding is that children should contain all the traceable/dynamic/JAX-array components (not sure the best language to use there), and aux_data should be all the static components (and must be hashable, I believe)

mattjj: Good idea. Though I don't like the word "static", I'll figure out a way to express that.

mattjj: (The reason I don't like the word "static" here is that I think of that as jit- or staging-specific jargon, which doesn't cover all cases here, e.g. in how pytree metadata interacts with vmap or AD. It's hard to think of a better term though...)

jakevdp: Maybe something like "generally all array components should go into `children`, and all non-array metadata should go into `aux_data`"

mattjj: Nice! That seems like a good rule of thumb.

jakevdp:
Ping on this - it would be a nice addition to the docs

PR #8741, opened 2021-11-30
[x64] improve consistency in handling dtype=None

jakevdp:
This makes JAX's `astype()` method match the behavior of numpy's with `dtype=None`.

Numpy behavior treats `None` as its default float type:
```python
>>> import numpy as np
>>> np.arange(4).astype(None)
array([0., 1., 2., 3.])
```
JAX behavior before:
```python
>>> import jax.numpy as jnp
>>> jnp.arange(4).astype(None)
DeviceArray([0, 1, 2, 3], dtype=int32)
```
JAX behavior after:
```python
>>> import jax.numpy as jnp
>>> jnp.arange(4).astype(None)
DeviceArray([0., 1., 2., 3.], dtype=float32)
```
Also clean up and test handling of `None` within `jax.dtypes` helpers, ensuring that `dtypes.dtype(None)` returns the default floating point type. This does not change behavior currently, but will make things more consistent for #8178

### DIFF
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -343,7 +343,9 @@ def is_python_scalar(x):
 
 def dtype(x, *, canonicalize=False):
   """Return the dtype object for a value or type, optionally canonicalized based on X64 mode."""
-  if isinstance(x, type) and x in python_scalar_dtypes:
+  if x is None:
+    return float_
+  elif isinstance(x, type) and x in python_scalar_dtypes:
     dt = python_scalar_dtypes[x]
   elif type(x) in python_scalar_dtypes:
     dt = python_scalar_dtypes[type(x)]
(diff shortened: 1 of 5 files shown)


jax/_src/dtypes.py
```diff
@@ -343,7 +343,9 @@ def is_python_scalar(x):
 
 def dtype(x, *, canonicalize=False):
   """Return the dtype object for a value or type, optionally canonicalized based on X64 mode."""
-  if isinstance(x, type) and x in python_scalar_dtypes:
+  if x is None:
```
hawkinsp, line 346: I'm wondering if callers expecting this behavior should be calling an API from `jnp` instead. Should `dtypes` encode this particular numpy quirk?

hawkinsp (approved):

jax/_src/dtypes.py
jakevdp: `dtypes` already encodes this particular quirk, because if `x = None` then it hits `np.result_type(x)`, which returns `float64`.

I made this change because I thought it would be better for `None` to explicitly return JAX's default float dtype, rather than implicitly returning numpy's default float dtype (currently the two defaults match, which is why this does not change the behavior of the function, but for the sake of #8178 returning the JAX default is useful)

I think it would be useful to make `x=None` result in a `ValueError` within `lax.dtype` and `lax.convert_element_type`, but since that would be a breaking change I think it would make sense to do in a separate PR.

PR #8753, opened 2021-12-01
add jax.experimental.checkify transformation

mattjj:
cf. #8468


jax/experimental/checkify.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @as_hashable_function(closure=params['out_axes_thunk'])
+    def new_out_axes_thunk():
+      return (0, 0, *params['out_axes_thunk']())
+
+    params_ = dict(params, in_axes=(None, None, *params['in_axes']),
+                   out_axes_thunk=new_out_axes_thunk,
+                   donated_invars=(False, False, *params['donated_invars']))
+    errs, codes, *outs = primitive.bind(f, e.err, e.code, *in_vals, **params_)
+    err, code = _reduce_any_error(errs, codes)
```
LenaMartens, line 126: I don't understand what _reduce_any_error does: sorts the boolean errors and returns the last one? So it would return a True error if there is one, but if there's multiple it would return an arbitrary one?

jax/experimental/checkify.py
```diff
@@ -0,0 +1,269 @@
...
+    else:
+      out = primitive.bind(*in_vals, **params)
+    if primitive.multiple_results:
+      return [ErrorTracer(self, x) for x in out]
+    else:
+      return ErrorTracer(self, out)
+
+  def process_call(self, primitive, f, tracers, params):
+    in_vals = [t.val for t in tracers]
+    e = popattr(self.main, 'error')
```
LenaMartens, line 106: Why do we need to pop the attribute here (and later set it again?)

LenaMartens (approved):
LGTM! Mainly two questions for my understanding.

jax/experimental/checkify.py
mattjj: We want the error value (i.e. the boolean and code) to be threaded through the call in the sense that it should be passed as an argument to the call application (in the jaxpr) and returned as an output.

When not dealing with higher-order primitives like this, we just keep the error value stuck on the MainTrace instance, as global data. (A cleaner version of this is `global_data` in Autodidax, an explicit attribute on MainTrace, but it's the same idea.)

So at this line in the code we want to get the error value from the MainTrace instance so that we can pass it in as an argument to the called function. Then just on the "other side" of the call, we take that passed-in argument and glue it back on the MainTrace instance. In short, we take off the error value, pass it through the call as an explicit argument, then immediately glue it back on.

In that first step, we don't actually have to pop/delete the error attribute from the MainTrace instance, but I thought it was cleaner that way.

WDYT?

LenaMartens: yeah, I guess my confusion was about why we pop it instead of getting it, I guess to make sure that the error is not available as an attribute on the Maintrace within the call primitive? Just to be safe?

mattjj: Yeah exactly, just to be safe. (Sorry for mostly answering the wrong question!)

jax/experimental/checkify.py
LenaMartens: Will it not return the code with the highest index because it returns the last element? Or am I thinking in the wrong dimension?

And is this "code with the lowest index" the "first" error which occurred in all mapped computations, or the error which occurred at the lowest mapped element? I think it's the latter, in which case I guess it doesn't matter which of the True codes we return (if we're only going to return one).

PR #8754, opened 2021-12-01
lax.convert_element_type: stricter dtype validation

jakevdp:
Following up on https://github.com/google/jax/pull/8741#discussion_r759798787

This includes a couple changes to how lax handles dtype inputs. In particular:

- dtypes.dtype(None) now raises an error rather than returning `float64`. The previous behavior came from the fact that it fell back to `np.result_type` for general inputs.
- `lax.convert_element_type(arr, None)` previously was a no-op. This is in contrast to `jax.numpy` and `numpy`, for which `x.astype(None)` converts to float. This was confusing, so `lax.convert_element_type(arr, None)` now results in an error.
- The dtype argument to `lax.convert_element_type` is now validated as a dtype. Previously, the implementation allowed any object with a dtype to be passed, so for example `lax.convert_element_type(x, jnp.arange(3))` would convert `x` to the dtype of the second argument. This now results in an error.
- As part of this, `dtypes.dtype(None)` now returns an error rather than `np.float_`. To keep the behavior of `arr.astype(None)` consistent with numpy's behavior, we now explicitly enumerate this case and convert to float.

All these previous behaviors were unintentional (I believe) and the new behavior is more in line with lax's philosophy of requiring strict/explicit inputs.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -464,10 +464,14 @@ def _convert_element_type(operand: Array, new_dtype: Optional[DType] = None,
                           weak_type: bool = False):
   # Don't canonicalize old_dtype because x64 context might cause
   # un-canonicalized operands to be passed in.
-  old_dtype = np.result_type(operand)
+  old_dtype = dtypes.dtype(operand, canonicalize=False)
   old_weak_type = dtypes.is_weakly_typed(operand)
 
-  new_dtype = dtypes.canonicalize_dtype(new_dtype or old_dtype)
+  if new_dtype is None:
+    new_dtype = old_dtype
+  else:
+    new_dtype = np.dtype(new_dtype)
+  new_dtype = dtypes.dtype(new_dtype, canonicalize=True)
   new_weak_type = bool(weak_type)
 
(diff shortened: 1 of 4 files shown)


jakevdp:
(pulling in before review for additional testing)

jax/_src/lax/lax.py
```diff
@@ -464,10 +464,14 @@ def _convert_element_type(operand: Array, new_dtype: Optional[DType] = None,
                           weak_type: bool = False):
   # Don't canonicalize old_dtype because x64 context might cause
   # un-canonicalized operands to be passed in.
-  old_dtype = np.result_type(operand)
+  old_dtype = dtypes.dtype(operand, canonicalize=False)
   old_weak_type = dtypes.is_weakly_typed(operand)
 
-  new_dtype = dtypes.canonicalize_dtype(new_dtype or old_dtype)
+  if new_dtype is None:
```
hawkinsp, line 470: Doesn't the PR description say this case is now an error?

hawkinsp (approved):

jax/_src/lax/lax.py
jakevdp: Added - thanks

PR #8773, opened 2021-12-02
add an assert primitive

mattjj:
The assert primitive has an effectful API and so it can't be staged out; it's only a trace-time primitive. It can be discharged to the functional form.

We might want to have separate transforms for discharging errors and for adding error checks. But right now they're just bundled together in the checkify transform.


jax/experimental/checkify.py
```diff
@@ -212,6 +214,24 @@ def check_errors_traceable(msgs, err, code, *args):
   yield outs
 
 
+## assert primitive
+
+def assert_(pred: Bool, msg: str) -> None:
```
LenaMartens, line 219: nit: could probably use a nicer public symbol name, but we can think of that later? I guess you're not supposed to overwrite keywords, and we can't provide a `checkify.assert`?

LenaMartens (approved):
LGTM!

jax/experimental/checkify.py
mattjj: I think we can't overwrite `assert` actually; it's a syntax error. (I tried, of course!)

You're right that this name is bad. But what's a good one? `jassert`? `assert2`? `asssert`? `jax_assert`? `buttert`? (Brainstormed those with @skye...)

mattjj: `allege`?

PR #8774, opened 2021-12-02
Add `profiler.annotate_function` to some internals

rsepassi:


rsepassi:
Fixed the failures in `errors_test` by adding `_src/profiler.py` to the traceback exclusions.

jax/_src/device_array.py
```diff
@@ -125,6 +126,7 @@ def _check_if_deleted(self):
     if self.device_buffer is deleted_buffer:
       raise RuntimeError("DeviceArray has been deleted.")
 
+  @annotate_function
```
hawkinsp, line 129: I'd strongly prefer we use a qualified name. `jax.profiler.annotate_function` or `profiler.annotate_function` has an immediately clear and obvious meaning. `annotate_function` does not.

Same everywhere.

hawkinsp (changes requested):

jax/_src/device_array.py
rsepassi: No problem, done.

hawkinsp (approved):

PR #8780, opened 2021-12-02
[sparse] BCOO.sum_duplicates: fix handling of explicit zeros

jakevdp:
Also patch up some corner cases in `jnp.unique` that the new tests revealed.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5618,14 +5618,21 @@ def _unique(ar, axis, return_index=False, return_inverse=False, return_counts=Fa
   """
   Find the unique elements of an array along a particular axis.
   """
+  if ar.shape[axis] == 0 and size and fill_value is None:
+    raise ValueError(
+      "jnp.unique: for zero-sized input with nonzero size argument, fill_value must be specified")
+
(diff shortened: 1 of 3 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -5618,14 +5618,21 @@ def _unique(ar, axis, return_index=False, return_inverse=False, return_counts=Fa
...
   if fill_value is not None:
     fill_value = asarray(fill_value, dtype=result.dtype)
   if size is not None and fill_value is not None:
-    valid = lax.expand_dims(arange(size) < mask.sum(), tuple(range(1, result.ndim)))
-    result = where(valid, result, fill_value)
+    if result.shape[0]:
+      valid = lax.expand_dims(arange(size) < mask.sum(), tuple(range(1, result.ndim)))
+      result = where(valid, result, fill_value)
+    else:
+      result = empty_like(result, shape=(size, *result.shape[1:])).at[:].set(fill_value)
```
soraros, line 5635: Wouldn't `full_like` work in this case?

jakevdp: Good catch - Done!

froystig (approved):

PR #8795, opened 2021-12-03
[sparse] initial implementation of sparse-sparse multiply

jakevdp:
This leaves some cases unimplemented, but is a good start.


froystig (approved):
Good questions in the inline comments. This is indeed a good start!

PR #8798, opened 2021-12-03
methods for retrieving IRs from each AOT stage

froystig:
part of #7733

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -514,15 +514,17 @@ class Lowered:
     self.donate_argnums = donate_argnums
     self._no_kwargs = no_kwargs
 
-  def _xla_computation(self):
-    # TODO(frostig): finalize API. For now, return the underlying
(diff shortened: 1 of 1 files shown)


jax/_src/api.py
```diff
@@ -552,6 +554,12 @@ def __init__(self, executable, in_tree, out_tree, donate_argnums,
     self.donate_argnums = donate_argnums
     self._no_kwargs = no_kwargs
 
+  def compiler_ir(self):
+    return self._executable.xla_executable().hlo_modules()
```
hawkinsp, line 558: Comment to explain why this exists? Is it supposed to be the optimized IR?

hawkinsp (approved):

jax/_src/api.py
froystig: Good idea. Yes it is. Added a comment.

PR #8829, opened 2021-12-06
Add checkify rule for scan.

LenaMartens:


tests/checkify_test.py
```diff
@@ -147,6 +147,64 @@ def f(x):
     self.assertIs(err.get(), None)
 
 
+  def test_scan_map(self):
+    def scan_body(_, x):
+      return None, jnp.sin(x)
```
mattjj, line 152: I think `jnp.sin(jnp.inf)` fails to generate a nan on TPU. That's why I made some of the tests in this file skip TPU. We might have to do that here. Or we could switch to a different test function+input, like jnp.inf * 0. (I'm not totally sure if this test will fail, but just wanted to give a heads up in case it happens!)

mattjj (approved):
Lovely!

tests/checkify_test.py
LenaMartens: Yep, they need to be skipped for TPU too (forgot to run the TPU tests). Thanks for the heads-up!

PR #8834, opened 2021-12-06
[x64] add new jax_default_dtype_bits flag

jakevdp:
This is part of landing step 1 of #8178.

This adds a new `jax_default_dtype` flag, which defaults to `'x64'`, and can be set to `'x32'`. It controls the precision of `dtypes.int_`, `dtypes.float_`, `jnp.int_`, `jnp.float_`, etc. and by extension, controls the default precision of arrays created by functions like `jnp.arange`, `jnp.linspace`, `jnp.argsort`, etc.

This makes it such that the core change of #8180 will simply be to flip the default flag value to `'32'`. Having the flag available during the transition to this new default will allow a softer landing for teams affected by this change.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -543,6 +543,14 @@ hlo_source_file_canonicalization_regex = config.define_string_state(
           'persistent compilation cache, which includes HLO metadata in the '
           'cache key.'))
 
+config.define_enum_state(
+    name='jax_default_dtype_bits',
+    enum_values=['32', '64'],
+    default='64',
+    help=('Specify bit width of default dtypes, either 32-bit or 64-bit. '
+          'This is a temporary flag that will be used during the process '
+          'of deprecating the ``jax_enable_x64`` flag.'))
+
 def _update_x64_global(val):
   lib.jax_jit.global_state().enable_x64 = val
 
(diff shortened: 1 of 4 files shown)


mattjj:
WDYT about calling it `jax_default_dtype` or `jax_default_dtype_precision`?  I'm just a little concerned with it getting confused with `jax_default_matmul_preicision`, in the sense that someone looking for the thing `jax_default_matmult_precision` controls might see `jax_default_precision` and assume it's what they're after. Both are about precision but it seems like mentioning `dtype` might make it a little more intuitive.

jax/_src/config.py
```diff
@@ -543,6 +543,17 @@ def update_thread_local_jit_state(**kw):
           'persistent compilation cache, which includes HLO metadata in the '
           'cache key.'))
 
+default_precision = config.define_enum_state(
```
mattjj, line 546:
```suggestion
config.define_enum_state(
```

`config.define_enum_state` returns a function to be used as a contextmanager, so if you don't want a contextmanager to be an option then we shouldn't define one here.

jakevdp: Done

jakevdp:
Thanks - renamed the flag to `jax_default_dtype`, though it's maybe a bit confusing (`"32"` isn't a dtype, after all). Maybe `jax_default_typewidth`? `jax_default_dtype_bits`?

jakevdp:
Or maybe keep it as `jax_default_dtype`, but make the enumeration `['x32', 'x64']` to stick with our current nomenclature

mattjj:
You make good points! `jax_default_dtype_bits` seems kinda nice and thorough; no one's going to confuse that with matmul precision! I trust whatever you settle on.

mattjj (approved):

PR #8841, opened 2021-12-07
[host_callback] Fix usage of host_callback together with lower_fun.

gnecula:
lower_fun is not quite public API, but people do use it when defining
new primitives. lower_fun was missing a call to the outfeed_rewriter.
In order to make the apply_outfeed_rewriter visible from xla.py, I
moved it from dispatch.py to core.py.

Bug: #8853

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -1019,6 +1019,7 @@ def lower_fun(fun: Callable, *, multiple_results: bool, backend=None,
       wrapped_fun = _tuple_output(wrapped_fun)
     with core.extend_axis_env_nd(zip(axis_env.names, axis_env.sizes)):
       jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, avals)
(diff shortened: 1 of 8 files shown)


jax/interpreters/xla.py
```diff
@@ -1019,6 +1019,7 @@ def f_with_avals(c, avals, xla_args, params):
       wrapped_fun = _tuple_output(wrapped_fun)
     with core.extend_axis_env_nd(zip(axis_env.names, axis_env.sizes)):
       jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, avals)
+    jaxpr = core.apply_outfeed_rewriter(jaxpr)
```
hawkinsp, line 1022: If this needs to go here, it also needs to go in the other variants within this function, and in the similar function in mlir.py.

But I'm a bit confused by, e.g., what `avals` means here now, because this function changes the jaxpr's signature. How can this be type correct? The caller will not have passed enough arguments to make this work.

hawkinsp (changes requested):

PR #8843, opened 2021-12-07
[x64]: improve dtype mismatch errors for lax functions

jakevdp:
This improves the type mismatch error message for `lax` operations, pointing to a similar numpy function that does automated type promotion as part of the API. This is a common type of error that arises in the transition to different default dtypes (see #8178), so hopefully this will improve user experience.

I couldn't think of any way to extract this mapping automatically, but the `lax` and `jax.numpy` APIs are stable enough at this point that I think hard-coding it is OK.

Before:
```pytb
$ python -c "from jax import lax; lax.add(1, 1.0)"
<...>
TypeError: add requires arguments to have the same dtypes, got int32, float32.
```
After:
```pytb
$ python -c "from jax import lax; lax.add(1, 1.0)"
<...>
TypeError: lax.add requires arguments to have the same dtypes, got int32, float32. (Tip: jnp.add is a similar function that does automatic type promotion on inputs).
```

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -4225,6 +4225,62 @@ def padtype_to_pads(in_shape, window_shape, window_strides, padding):
     raise TypeError(msg.format(padding))
 
 
+# Map of lax function to equivalent jax.numpy function for use in error string below.
+_JNP_FUNCTION_EQUIVALENTS = {
+  'abs': 'fabs',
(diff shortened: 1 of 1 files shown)


mattjj (approved):
Nice!

(Not for this PR, but maybe a future one, just thinking about error messages: if one argument is a Python literal and that causes an exception, we could mention it in the error message.)

PR #8848, opened 2021-12-08
[sparse] add jvp rule for bcoo_spdot_general

jakevdp:
I'll do the transpose rule in a later PR.

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -886,8 +886,23 @@ def _bcoo_spdot_general_batch_rule(batched_args, batch_dims, *, dimension_number
                                  lhs_shape=new_lhs_shape, rhs_shape=new_rhs_shape)
   return batched_out, (result_batch_dim, result_batch_dim)
 
-# TODO(JVP): jvp, transpose
+
+def _bcoo_spdot_general_jvp(primals, tangents, **kwds):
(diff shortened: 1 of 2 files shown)


froystig (approved):

lucagrementieri:
Will the vjp rule for bcoo_spdot_general implemented soon?

jakevdp:
I'm not currently working on it - is it something you need?

lucagrementieri:
I'm writing an optimization software that has to back-propagate through a product of large sparse matrices.
I would like to use JAXopt, but the program crashes because JAXopt uses the gradient computed using reverse mode differentiation. I can change the inner working of JAXopt but it would be better to have the vjp rule for the product of sparse matrices.

I'm trying to implement the vjp rule myself right now, but I have no previous experience with the inner workings of JAX, so I'm not sure I can succeed.

jakevdp:
The challenge with the spdot_general transpose rule is that it requires the sparse equivalent of bcoo_extract, which is not yet implemented, and is an operation that I'm not sure how to implement efficiently. You can see roughly what it would have to do by looking at `_bcoo_dot_general_transpose`: https://github.com/google/jax/blob/be64d8ba7c0ebffc935cd9d88bb6e927b6ee491c/jax/experimental/sparse/bcoo.py#L874

jakevdp:
That said, it seems like there may be some problem with the forward rule as well: #10163

PR #8852, opened 2021-12-08
[sparse] Update bcoo_dot_general GPU translation rule.

tlu7:
Co-authored-by: vanderplas@google.com

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -25,7 +25,7 @@ from jax import lax
 from jax import tree_util
 from jax import vmap
 from jax.experimental.sparse._base import JAXSparse
-from jax.experimental.sparse.util import _safe_asarray
+from jax.experimental.sparse.util import _safe_asarray, CuSparseEfficiencyWarning
 from jax.interpreters import batching
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
@@ -36,7 +36,11 @@ from jax._src.api_util import flatten_axes
 from jax._src.lax.lax import (
   ranges_like, remaining, _dot_general_batch_dim_nums, _dot_general_shape_rule,
   DotDimensionNumbers)
+from jax._src.lib import cusparse
+from jax._src.lib import xla_client as xc
 from jax._src.numpy.lax_numpy import _unique
+
+xops = xc._xla.ops
 
 Dtype = Any
 Shape = Tuple[int, ...]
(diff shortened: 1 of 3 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,136 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+  assert len(data_shape) <= 2
+  assert len(bcoo_indices.shape) <= 2
+
+  nse, n_sparse = bcoo_indices.shape
+  assert n_sparse <= 2
+
+  mask = (jnp.broadcast_to(data_shape, (nse, n_sparse)) -
+          bcoo_indices).astype(bool).astype(jnp.int32)
+
+  bcoo_indices = jnp.multiply(bcoo_indices, mask)
```
jakevdp, line 647: In order to catch *any* out-of-bound index, we could do something like this:
```python
bcoo_indices = jnp.where(bcoo_indices < jnp.array(data_shape), bcoo_indices, 0)
```

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,136 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+
+  return bcoo_indices
+
+def _bcoo_padding_indices_correction_fn(
+    lhs_data, lhs_indices, rhs, *, dimension_numbers, lhs_shape):
+  return _bcoo_padding_indices_correction_impl(lhs_indices, lhs_shape)
+
+_bcoo_padding_indices_correction = xla.lower_fun(
+      _bcoo_padding_indices_correction_fn, multiple_results=False,
+      new_style=True)
```
jakevdp, line 657: I think we can avoid the extra function wrapper by doing something like this:
```python
_bcoo_padding_indices_correction_translation = xla.lower_fun(
    _bcoo_padding_indices_correction_impl, multiple_results=False, new_style=True)
```
Then when we call it, use:
```python
aval = avals_in[1]
lhs_indices = _bcoo_padding_indices_correction(
      ctx, [aval], [aval], data_shape=lhs_shape)
```
something like that... note the input aval is the same as the output aval, becuase the function is returning an array of the same shape and dtype as the input.

jakevdp (commented):
Looking good - we're getting there!

jax/experimental/sparse/bcoo.py
tlu7: I also thought about not using the function wrapper. It reports the error below
`    unflattened = [[next(xs_iter) for _ in range(n)] for n in ns]
jax._src.traceback_util.UnfilteredStackTrace: StopIteration`

at 

`_CompileAndCheck`
 
for 

` _bcoo_dot_general_gpu_translation_rule
    ctx, [aval], [aval], data_shape=lhs_shape)`.

The reason for using function wrapper is to keep the same set of arguments (both positional and keyward). Do you have any insights on the reported error?

jax/experimental/sparse/bcoo.py
tlu7: Changed to `jnp.where`.  I think `jnp.where` is based on `lax.select`, `broadcast_arrays`, and complex logic like element-wise not equal `lax.ne`, which may not be as efficient as the matrix multiplications on GPU/TPU.

tlu7:
> 

Thanks for the comments! PTAL.

jax/experimental/sparse/bcoo.py
jakevdp: Sure, but the matrix multiplication approach only catches out-of-bound indices that are exactly equal to the shape, and will lead to a segfault otherwise. I think we should focus on avoiding segfaults before focusing on optimization!

jax/experimental/sparse/bcoo.py
jakevdp: I see - there might be some interplay between `ctx` and the avals.

Still, it seems wrong to me to pass `avals_out` to this function, since they have nothing to do with the outputs of this function!

jakevdp: I just tried a simpler version of this locally and it seems to work fine the way I indicated in my initial comment (lowering the utility directly). The error you're seeing makes me think that the avals you passed do not have a one-to-one mapping with the arguments you passed.

jakevdp: Here's the code I wrote to try it out:
```python
_bcoo_dot_general_default_translation_rule = xla.lower_fun(
    _bcoo_dot_general_impl, multiple_results=False, new_style=True)

@functools.partial(xla.lower_fun, multiple_results=True, new_style=True)
def _bcoo_correct_oob_indices(data, indices, *, shape):
  props = _validate_bcoo(data, indices, shape)
  if props.n_batch or props.n_dense:
    raise NotImplementedError("n_batch & n_dense not yet supported")
  mask = (indices >= jnp.array(shape)).any(-1)
  return jnp.where(mask, 0, data), jnp.where(mask[:, None], 0, indices)

def _bcoo_dot_general_custom_translation_rule(ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers, lhs_shape):
  lhs_data, lhs_indices = _bcoo_correct_oob_indices(ctx, avals_in[:2], avals_in[:2], lhs_data, lhs_indices, shape=lhs_shape)
  return _bcoo_dot_general_default_translation_rule(ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)

# ...

xla.register_translation(bcoo_dot_general_p, _bcoo_dot_general_custom_translation_rule)
```

jakevdp: Then tested with
```python
import operator
import jax
import jax.numpy as jnp
from jax.experimental import sparse

x = sparse.BCOO.fromdense(jnp.arange(6).reshape(2, 3))
y = jnp.arange(3)
print(jax.jit(operator.matmul)(x, y))
# [5, 14]
```

tlu7: Thanks for the offline discussion! Love the concise and precise way you implemented it! Copied it and update the PR.

tlu7: Extended the test coverage for the returned data for data with nonzero entries at the corrected indices.

jax/experimental/sparse/bcoo.py
tlu7: Agree!

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,121 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
 
   return core.ShapedArray(out_shape, lhs_data.dtype)
 
+def _bcoo_correct_oob_indices_impl(data, indices, shape):
+  """Replaces out-of-bound indices in a padded representation with zeros.
+
+  Here is an example of padded representation:
```
jakevdp, line 620: There is a standard docstring formatting to use for examples which can be used by autodoc (for generating HTML docs) and doctest (for automatically testing the examples and assuring they don't go stale). You can see en example of it here: https://github.com/google/jax/blob/f6f6b4f04cb1cfb64d531eb41172e1783c57fe47/jax/experimental/sparse/bcoo.py#L1083-L1098

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,121 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
+
+  if (n_batch or n_dense or
+      n_sparse not in [1, 2] or rhs_aval.ndim not in [1, 2] or
+      lhs_batch or rhs_batch or len(lhs_contract) != 1):
+    return _bcoo_dot_general_default_translation_rule(
+      ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs,
+      dimension_numbers=dimension_numbers, lhs_shape=lhs_shape)
+  else:
+    lhs_indices = _bcoo_correct_oob_indices(
```
jakevdp, line 725: I think tis should be `lhs_data, lhs_indices = ...`

jax/experimental/sparse/bcoo.py
tlu7: Updated the dosctring.

tlu7: Learned the new thing of doctest
`python3 -m pytest --doctest-modules jax/experimental/sparse/sparse_test.py
`
Thanks for teaching me @jakevdp 
Sorry to spam people's git history log.

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,147 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
 
   return core.ShapedArray(out_shape, lhs_data.dtype)
 
+def _bcoo_correct_oob_indices_impl(data, indices, shape):
+  """Replaces out-of-bound indices in a padded representation with zeros.
+
```
jakevdp, line 619: Maybe add some context like:
```suggestion
The BCOO format at times must pad indices to indicate unused entries, and it does this
with out-of-bounds indices. This can cause problems when passing the data to other
backends (like cuSparse) that don't handle this natively. This function converts out-of-bounds
indices to zeros, and also zeros-out associated values in the data array.
```

jakevdp:
Looks good! Let's pull it in and run the GPU tests to be sure.

jakevdp (approved):

jax/experimental/sparse/bcoo.py
tlu7: Thanks. Done.

tests/sparsify_test.py
```diff
@@ -373,16 +373,17 @@ def testWeakTypes(self):
...
+# class SparsifyTracerTest(SparsifyTest):
+#   @classmethod
+#   def sparsify(cls, f):
+#     return sparsify(f, use_tracer=True)
+#
+#   def testTracerIsInstanceCheck(self):
+#     @self.sparsify
+#     def f(x):
+#       self.assertIsInstance(x, SparseTracer)
+#     f(jnp.arange(5))
```
jakevdp, line 386: Have you looked at whether the tracer-based transform still causes issues now that the rest of the code is correct?

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,152 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+_bcoo_correct_oob_indices = xla.lower_fun(
+    _bcoo_correct_oob_indices_impl, multiple_results=True, new_style=True)
+
+_bcoo_dot_general_default_translation_rule = xla.lower_fun(
+    _bcoo_dot_general_impl, multiple_results=False, new_style=True)
+
+def _bcoo_dot_general_cuda_translation_rule(
+    ctx, avals_in, avals_out, lhs_data, lhs_indices, rhs, *, dimension_numbers,
+    lhs_shape):
```
jakevdp, line 681: I'm still a bit worried about the way this is structured. There are a number of conditions that will make this function fail very badly (e.g. matrices larger than 2 dimensions, wrong dtypes, etc.). You've checked for those conditions in `_bcoo_dot_general_gpu_translation_rule`, but having that validation so far separated from this code seems problematic to me. Not because the code is wrong, but because that kind of action at a distance is hard to understand and maintain.

Maybe we can add a list of assertions here?

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,152 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+  _, lhs_indices_aval, _, = avals_in
+  props = _validate_bcoo_indices(lhs_indices_aval, lhs_shape)
+
+  rhs_ndim = len(c.get_shape(rhs).dimensions())
+  if rhs_ndim == 1:
+    bcoo_dot_general_fn = cusparse.coo_matvec
+  elif rhs_ndim == 2:
+    bcoo_dot_general_fn = cusparse.coo_matmat
+    if rhs_contract[0] == 1:
+      rhs = xops.Transpose(rhs, permutation=[1, 0])
```
jakevdp, line 695: We should catch the `else` case here explicitly and raise an error... I know we've already checked this in another function, but that's one of those "actions at a distance" that will make this hard to follow when we come back to it in the future.

jax/experimental/sparse/bcoo.py
```diff
@@ -610,6 +614,152 @@ def _bcoo_dot_general_abstract_eval(lhs_data, lhs_indices, rhs, *, dimension_num
...
+                   start_indices=[0, 1],
+                   limit_indices=[c.get_shape(lhs_indices).dimensions()[0], 2],
+                   strides=[1, 1]),
+        dimensions=[0, 1])
+
+    if lhs_contract[0] == 0:
+      lhs_transpose = True
+
+    return [bcoo_dot_general_fn(
+        c, lhs_data, row, col, rhs, shape=lhs_shape, transpose=lhs_transpose)]
```
jakevdp, line 730: The function implicitly returns `None` if `props.n_sparse` is not `1` or `2`, which would result in a huge & incomprehensible traceback through the JAX machinery. Another action-at-distance 😁

jakevdp:
GPU tests seem to be passing! This is looking really nice.

tests/sparsify_test.py
tlu7: Yes, I am still having issue with the tracer-based sparsify transform. Need to revisit it and go deeper on the transform.

jakevdp: Maybe we can just skip the affected test; e.g.
```python
class SparsifyTracerTest(SparsifyTest):
  @classmethod
  def sparsify(cls, f):
    return sparsify(f, use_tracer=True)

  def testTracerIsInstanceCheck(self):
    @self.sparsify
    def f(x):
      self.assertIsInstance(x, SparseTracer)
    f(jnp.arange(5))

  def testSparseMatmul(self):
    if jtu.device_under_test() == "gpu":
      self.skipTest("fails on gpu with cusparse")
    super().testSparseMatmul()
```
(untested, but I think something like that should work)

tlu7: >

tlu7: It works! I think for the next step I will see how the tracer-based transform looks like in the dot general op with lhs.T. Added a TODO.

jax/experimental/sparse/bcoo.py
tlu7: Done.

tests/sparsify_test.py
tlu7: I still see the segmentation fault by skipping the gpu test.

```
jax/_src/dispatch.py", line 436 in _execute_compiled
jax/_src/dispatch.py", line 100 in <lambda>
jax/_src/dispatch.py", line 80 in apply_primitive
jax/core.py", line 630 in process_primitive
jax/core.py", line 278 in bind
jax/_src/lax/lax.py", line 495 in _convert_element_type
jax/_src/numpy/lax_numpy.py", line 3633 in array
jax/_src/numpy/lax_numpy.py", line 3652 in asarray
jax/experimental/sparse/transform.py", line 85 in push
jax/experimental/sparse/transform.py", line 133 in array_to_argspec
jax/_src/tree_util.py", line 178 in <genexpr>
jax/_src/tree_util.py", line 178 in tree_map
jax/experimental/sparse/transform.py", line 134 in arrays_to_argspecs
jax/experimental/sparse/transform.py", line 253 in sparsify_fun
jax/experimental/sparse/transform.py", line 267 in _wrapped
jax/tests/sparsify_test.py", line 297 in testSparsifyDenseXlaCall
```

jax/experimental/sparse/bcoo.py
tlu7: Done.

jax/experimental/sparse/bcoo.py
tlu7: Done.

jakevdp (approved):

PR #8860, opened 2021-12-08
Add checkify rule for while_loop.

LenaMartens:
This one is a bit more involved than `scan`, because an error can occur either in the cond or body function.

So, thread the error into the while_loop input, ignore the error argument in the `cond` function (because it needs to return a boolean and cannot do its own error threading), rewrite the `body` function to also call the `cond` function -> it adds any `cond` errors to the body error threading!

I'm sure I'm missing some jaxpr util functions, I wasn't sure if eg. I could just add two new (ignored) invars on a jaxpr without re-building the whole jaxpr.

One limitation: if an error occurs in the cond function and also the cond function then returns False, no error will be reported: because the loop stops and the body will not be executed.


mattjj:
> I wasn't sure if eg. I could just add two new (ignored) invars on a jaxpr without re-building the whole jaxpr.

I think that should be fine.

> One limitation: if an error occurs in the cond function and also the cond function then returns False, no error will be reported: because the loop stops and the body will not be executed.

I guess outside the loop we could check the cond function for an error on the init_val. Do you think that could work?

jax/experimental/checkify.py
```diff
@@ -298,6 +301,39 @@ def scan_error_check(error, *in_flat, reverse, length, jaxpr, num_consts, num_ca
...
+  cond_f = core.jaxpr_as_fun(cond_jaxpr)
+  body_f = core.jaxpr_as_fun(body_jaxpr)
+  def new_body_f(*vals):
+    _ = cond_f(*vals)
+    return body_f(*vals)
+  return checkify_fun_to_jaxpr(lu.wrap_init(new_body_f), error, body_jaxpr.in_avals)
+
+def ignore_errors_jaxpr(jaxpr, error):
+  """Constructs a jaxpr which takes two extra args but ignores them."""
+  f = core.jaxpr_as_fun(jaxpr)
```
mattjj, line 314: I think it might be safe to implement this function as roughly

```python
  err_aval = ...
  code_aval = ...
  new_jaxpr = core.Jaxpr(jaxpr.jaxpr.constvars, (core.Var(err_aval), core.Var(code_aval), *jaxpr.jaxpr.invars), ...)
  return core.ClosedJaxpr(jaxpr.consts, new_jaxpr)
```

That would save the tracing round trip.

mattjj (approved):
Awesome!

jax/experimental/checkify.py
LenaMartens: Cool, I've done that!

I used `core.gensym` to generate Vars which do not conflict with the existing Jaxpr Vars, not sure if that's the best way.

LenaMartens:
> I guess outside the loop we could check the cond function for an error on the init_val. Do you think that could work?

Yep, it would work for the initial iteration, but we'd still need a mechanism if this occurs in further iterations. Maybe check if the _next_ cond errors in the current body iteration, then if an error occurs and no body error has occured, log that error? Not sure how to thread that, but something like:

```
while(cond(val)):
  body_error, out_body = error_body(val)
  cond_error, _ =  cond(out_body)
  return body_error or cond_error  
```
Actually, probably if we just call the next cond function after the current body function, the errors will be threaded correctly!

Anyways, I've left in a TODO to fix this.

PR #8868, opened 2021-12-08
Testing: avoid hard-coding random seeds

jakevdp:
Hard-coding common random seeds in a test suite is poor practice, because we end up testing the same inputs over and over (particularly in parametrized tests). JAX's test runners have `self.rng()` for this purpose; we should use this everywhere.

Making this change revealed a few tests that are very fragile – some I was able to fix by adjusting test tolerances, but some are irredeemable, and so I went back to a fixed random seed with an explicit note about the issue (mostly as an indicator of existing problems in case someone finds functions behaving unexpectedly & digs into the code).

In other places, we had non-deterministic tests (ones that relied on the global numpy random seed). As part of this change, I also made these tests deterministic.

### DIFF
--- a/examples/control_test.py
+++ b/examples/control_test.py
@@ -15,9 +15,9 @@
 from functools import partial
 
 from absl.testing import absltest
-import numpy as np
 
 from jax import lax
+# TODO(jakevdp) avoid dependence on private test_util in examples
 from jax._src import test_util as jtu
 import jax.numpy as jnp
 
@@ -138,7 +138,7 @@ class ControlExampleTest(jtu.JaxTestCase):
 
 
   def testLqrPredict(self):
-    randn = np.random.RandomState(0).randn
+    randn = self.rng().randn
     dim, T = 2, 10
     p = one_step_lqr(dim, T)
     x0 = randn(dim)
@@ -153,7 +153,7 @@ class ControlExampleTest(jtu.JaxTestCase):
 
 
   def testIlqrWithLqrProblem(self):
-    randn = np.random.RandomState(0).randn
+    randn = self.rng().randn
     dim, T, num_iters = 2, 10, 3
     lqr = one_step_lqr(dim, T)
     p = control_from_lqr(lqr)
@@ -166,7 +166,7 @@ class ControlExampleTest(jtu.JaxTestCase):
 
 
   def testIlqrWithLqrProblemSpecifiedGenerally(self):
-    randn = np.random.RandomState(0).randn
+    randn = self.rng().randn
     dim, T, num_iters = 2, 10, 3
     p = one_step_control(dim, T)
     x0 = randn(dim)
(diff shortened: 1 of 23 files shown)


jakevdp:
Pulling-in pre-review for more test coverage.

tests/jet_test.py
```diff
@@ -329,9 +329,11 @@ def test_atan2(self):       self.binary_check(lax.atan2, lims=[-40, 40])
 
   @jtu.skip_on_devices("tpu")
   def test_clamp(self):
-    lims = [-2, 2]
+    lims = [-1, 1]
     order = 3
     dims = 2, 3
+    # rng = self.rng()
+    # This test is very sensitive to the inputs, so we use a known working seed.
```
hawkinsp, line 336: I think a better fix here would be to change the test to make sure that the primal points lie outside an epsilon ball of the two critical points in the function.

tests/lax_autodiff_test.py
```diff
@@ -1055,18 +1055,18 @@ def f2(x, y):
 
   # TODO(mattjj): make this a more systematic test
   def testRemainder(self):
-    rng = np.random.RandomState(0)
+    rng = self.rng()
     x = rng.uniform(-0.9, 9, size=(3, 4))
     y = rng.uniform(0.7, 1.9, size=(3, 1))
     assert not set(np.unique(x)) & set(np.unique(y))
-    tol = 1e-1 if jtu.num_float_bits(np.float64) == 32 else 1e-3
+    tol = 1e-1
```
hawkinsp, line 1062: This seems like a pretty loose tolerance for a 64-bit computation.

hawkinsp (approved):

tests/lax_autodiff_test.py
jakevdp: The test fails with the old tolerance.

Options:
- continue hard-coding the random seed
- keep this adjustment
- disable the test

Which do you think is best?

examples/control_test.py
```diff
@@ -138,7 +137,7 @@ def testLqrSolve(self):
 
 
   def testLqrPredict(self):
-    randn = np.random.RandomState(0).randn
```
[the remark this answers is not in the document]
jakevdp: I can make that change, but it's a much bigger one (e.g. we can't use `self.assertAllClose`). I'd prefer to do it in another PR.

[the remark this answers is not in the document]
hawkinsp: That's fine.

tests/lax_autodiff_test.py
hawkinsp: Bumping the tolerance and leaving a TODO seems fine.

examples/control_test.py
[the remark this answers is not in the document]
jakevdp: Added TODOs for this.

tests/jet_test.py
jakevdp: Added a TODO for this.

PR #8869, opened 2021-12-09
Remove all `stop_gradient` calls from `custom_root`.

### LINKED ISSUE #8744: `lax.custom_root` fails on root finders that require gradient
The `lax.custom_root` method does not work on methods that require gradient, e.g., the Newton-Raphson method. The core of the problem seems to be overly-aggressive application of `stop_gradient` [here](https://github.com/google/jax/blob/e6950dd63ed3c054067c53bd4806dfbb85e0eef7/jax/_src/lax/control_flow.py#L2310).

While I've included a minimal example below, [this notebook](https://colab.research.google.com/drive/1e5_0gGYc-v9lKKwvUS7rqSUhA0D-OpT3) does a deeper dive into the error and demonstrates that removing `_stop_gradient_fun` applications resolves the problem.

I'd be happy to make a PR to resolve the issue, but my naïve solution of removing all calls to `stop_gradient` in `custom_root` appears... well, naïve. I'm just not sure why `stop_gradient` is required at all here, but it seems related to not wanting to trace gradients through control structures like while-loops. I suppose I'm curious why custom JVP'ing a function doesn't automatiically shortcut the gradient computation...

### Related issues

 - #1448 discusses a `define_implicit_gradient` function that doesn't use `stop_gradient`. This function was never merged. Regardless, the fact that `custom_root` methods can't use gradients seems like a bug.
   - Additional reading of the discussion there suggests that one of the main reasons for `stop_gradient` was the non-differentiability of control structures (`lax.while_loop` in particular). Nevertheless, my tests don't seem to trigger the issue.

## Minimal example

This example follows [the JAX unit test](https://github.com/google/jax/blob/e6950dd63ed3c054067c53bd4806dfbb85e0eef7/tests/lax_control_flow_test.py#L2010) of the `custom_root` method, with the bisection method replaced with Newton's method.

```python
import jax
import numpy as np

def newton(f, x0):
  """Newton's method for root-finding."""
  initial_state = (0, x0)  # (iteration, x)

  def cond(state):
    it, x = state
    # We fix 10 iterations for simplicity, this is plenty for convergence in our tests.
    return (it < 10)

  def body(state):
    it, x = state
    fx, dfx = f(x), jax.grad(f)(x)
    step = fx / dfx
    new_state = it + 1, x - step
    return new_state

  return jax.lax.while_loop(
    cond,
    body,
    initial_state,
  )[1]
```

### Newton's method without `custom_root`

```python
initial_guess = 2.0
x_test = 5.0

def sqrt_cubed_newton(x):
  implicit = lambda y: y ** 2 - x ** 3
  return newton(implicit, initial_guess)

print("Without custom_root")
v = sqrt_cubed_newton(x_test)
print(f"Value = {v}")
if abs(v - x_test ** 1.5) < 1e-5:
  print("Pass")
else:
  print("Fail")
```

```
Without custom_root
Value=11.180339813232422
Pass
```

### With `custom_root`

```python

def sqrt_cubed_newton_custom_root(x):
  def scalar_solve(g, y):
    return y / g(1.0)
  implicit = lambda y: y ** 2 - x ** 3
  return jax.lax.custom_root(implicit, initial_guess, newton, scalar_solve)

print("With custom_root")
v = sqrt_cubed_newton_custom_root(x_test)
print(f"Value = {v}")
if abs(v - x_test ** 1.5) < 1e-5:
  print("Pass")
else:
  print("Fail")
```
```
With custom_root
Value = nan
Fail
```

The notebook referenced above shows how removing the call to `_stop_gradient_fun` fixes the problem.

Please:

- [x] Provide a complete example of how to reproduce the bug, wrapped in triple backticks like this

mbmccoy:
Resolves #8744 by removing all calls to `stop_gradient` in `jax.lax.control_flow.custom_root`.

### Context

The `custom_root` function allows users to create custom root finders, i.e., methods that solve `f(x, y) == 0` for `x`, and then automatically generates the gradient of the found `x` with respect to `y` using the implicit function theorem.

Unfortunately, `custom_root` does not support root-finding methods that require gradients, e.g., Newton's method and related approaches. The main issue was the application of `lax.stop_gradient` to all arguments of the root-finding method. This approach was needed in the past due to 

1. Implementation of `custom_root` as a primitive, and
2. The lack of differentiability of many jax control structures, such as `while_loop`.

Fortunately, both of these issues are no longer blocking, so we can remove the application of `stop_gradient`.

### Changes

1. Removes `_stop_gradient_fun` from `lax.control_flow`. (Key functional change.)
2. Adds tests, including an implementation of a simple newton `custom_root` solver. Extant tests updated to use this solver in addition to binary search.

### DIFF
--- a/tests/lax_control_flow_test.py
+++ b/tests/lax_control_flow_test.py
@@ -95,6 +95,52 @@ ignore_jit_of_pmap_warning = partial(
   jtu.ignore_warning, message=".*jit-of-pmap.*")
 
 
+# Simple optimization routine for testing custom_root
+def binary_search(func, x0, low=0.0, high=100.0):
+  del x0  # unused
+
+  def cond(state):
+    low, high = state
+    midpoint = 0.5 * (low + high)
+    return (low < midpoint) & (midpoint < high)
+
+  def body(state):
+    low, high = state
+    midpoint = 0.5 * (low + high)
+    update_upper = func(midpoint) > 0
+    low = jnp.where(update_upper, low, midpoint)
+    high = jnp.where(update_upper, midpoint, high)
+    return (low, high)
+
+  solution, _ = lax.while_loop(cond, body, (low, high))
+  return solution
+
+# Optimization routine for testing custom_root.
+def newton_raphson(func, x0):
+  tol = 1e-16
+  max_it = 20
+
+  fx0, dfx0 = func(x0), jax.jacobian(func)(x0)
+  initial_state = (0, x0, fx0, dfx0)  # (iteration, x, f(x), grad(f)(x))
+
+  def cond(state):
+    it, _, fx, _ = state
+    return (jnp.max(jnp.abs(fx)) > tol) & (it < max_it)
+
+  def body(state):
+    it, x, fx, dfx = state
+    step = jnp.linalg.solve(
+      dfx.reshape((-1, fx.size)), fx.ravel()
+    ).reshape(fx.shape)
+    x_next = x - step
+    fx, dfx = func(x_next), jax.jacobian(func)(x_next)
+    return (it + 1, x_next, fx, dfx)
+
+  _, x, _, _ = lax.while_loop(cond, body, initial_state)
+
+  return x
+
+
 class LaxControlFlowTest(jtu.JaxTestCase):
 
   def setUp(self):
@@ -2007,33 +2053,19 @@ class LaxControlFlowTest(jtu.JaxTestCase):
 
     jax.grad(lambda x: jit_run_scan(x))(0.)  # doesn't crash
 
-  def test_custom_root_scalar(self):
+  @parameterized.named_parameters(
+      {"testcase_name": "binary_search", "solve_method": binary_search},
+      {"testcase_name": "newton_raphson", "solve_method": newton_raphson},
+  )
+  def test_custom_root_scalar(self, solve_method):
 
     def scalar_solve(f, y):
       return y / f(1.0)
 
-    def binary_search(func, x0, low=0.0, high=100.0):
-      del x0  # unused
-
-      def cond(state):
-        low, high = state
-        midpoint = 0.5 * (low + high)
-        return (low < midpoint) & (midpoint < high)
-
-      def body(state):
-        low, high = state
-        midpoint = 0.5 * (low + high)
-        update_upper = func(midpoint) > 0
-        low = jnp.where(update_upper, low, midpoint)
-        high = jnp.where(update_upper, midpoint, high)
-        return (low, high)
-
-      solution, _ = lax.while_loop(cond, body, (low, high))
-      return solution
-
     def sqrt_cubed(x, tangent_solve=scalar_solve):
       f = lambda y: y ** 2 - x ** 3
-      return lax.custom_root(f, 0.0, binary_search, tangent_solve)
+      # Note: Nonzero derivative at x0 required for newton_raphson
+      return lax.custom_root(f, 1.0, solve_method, tangent_solve)
 
     value, grad = jax.value_and_grad(sqrt_cubed)(5.0)
     self.assertAllClose(value, 5 ** 1.5, check_dtypes=False, rtol=1e-6)
@@ -2044,7 +2076,11 @@ class LaxControlFlowTest(jtu.JaxTestCase):
 
     inputs = jnp.array([4.0, 5.0])
     results = jax.vmap(sqrt_cubed)(inputs)
-    self.assertAllClose(results, inputs ** 1.5, check_dtypes=False)
+    self.assertAllClose(
+      results, inputs ** 1.5, check_dtypes=False,
+      atol={jnp.float32: 1e-3, jnp.float64: 1e-6},
+      rtol={jnp.float32: 1e-3, jnp.float64: 1e-6},
+    )
 
     results = jax.jit(sqrt_cubed)(5.0)
     self.assertAllClose(
@@ -2072,6 +2108,30 @@ class LaxControlFlowTest(jtu.JaxTestCase):
     actual = jax.jit(linear_solve)(a, b)
     expected = jnp.linalg.solve(a, b)
     self.assertAllClose(expected, actual)
+
+  def test_custom_root_vector_nonlinear(self):
+
+    def nonlinear_func(x, y):
+      # func(x, y) == 0 if and only if x == y.
+      return (x - y) * (x**2 + y**2 + 1)
+
+    def tangent_solve(g, y):
+      return jnp.linalg.solve(
+        jax.jacobian(g)(y).reshape(-1, y.size),
+        y.ravel()
+      ).reshape(y.shape)
+
+    def nonlinear_solve(y):
+      f = lambda x: nonlinear_func(x, y)
+      x0 = -jnp.ones_like(y)
+      return lax.custom_root(f, x0, newton_raphson, tangent_solve)
+
+    y = self.rng().randn(3, 1)
+    jtu.check_grads(nonlinear_solve, (y,), order=2,
+                rtol={jnp.float32: 1e-2, jnp.float64: 1e-3})
+
+    actual = jax.jit(nonlinear_solve)(y)
+    self.assertAllClose(y, actual, rtol=1e-5, atol=1e-5)
 
   def test_custom_root_with_custom_linear_solve(self):
 
--- a/docs/notebooks/Custom_derivative_rules_for_Python_code.ipynb
+++ b/docs/notebooks/Custom_derivative_rules_for_Python_code.ipynb
@@ -1070,7 +1070,7 @@
     "id": "HowvqayEuy-H"
    },
    "source": [
-    "A limitation to this approach is that the argument `f` can't close over any values involved in differentiation. That is, you might notice that we kept the parameter `a` explicit in the argument list of `fixed_point`. While other JAX mechanisms can handle closed-over transformation-traced values in the arguments to higher-order functions (as is done for the control flow primitives like `lax.cond`, `lax.scan`, and `lax.while_loop` itself), `jax.custom_vjp` used as above cannot. A `fixed_point` routine that used a bit more of JAX's internals could have a more convenient and robust API."
+    "A limitation to this approach is that the argument `f` can't close over any values involved in differentiation. That is, you might notice that we kept the parameter `a` explicit in the argument list of `fixed_point`. For this use case, consider using the low-level primitive `lax.custom_root`, which allows for deriviatives in closed-over variables with custom root-finding functions."
    ]
   },
   {
--- a/docs/notebooks/Custom_derivative_rules_for_Python_code.md
+++ b/docs/notebooks/Custom_derivative_rules_for_Python_code.md
@@ -560,7 +560,7 @@ print(grad(grad(jnp.sqrt))(2.))
 
 +++ {"id": "HowvqayEuy-H"}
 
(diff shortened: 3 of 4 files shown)


tests/lax_control_flow_test.py
```diff
@@ -96,6 +96,76 @@ def posify(matrix):
   jtu.ignore_warning, message=".*jit-of-pmap.*")
 
 
+# Simple optimization routine for testing custom_root
+def binary_search(func, x0, low=0.0, high=100.0):
```
mbmccoy, line 100: Refactored so we can use in parameterized tests.

tests/lax_control_flow_test.py
```diff
@@ -96,6 +96,76 @@ def posify(matrix):
...
+    # The max() helps the JIT understand
+    # that we get the same output size for both
+    # branches.
+    lambda _: jnp.divide(b, A.max()).reshape(b.shape),
+    lambda _: jnp.linalg.solve(
+        A.reshape((-1,b.size)),
+        b.reshape((b.size, 1))
+      ).reshape(b.shape),
+    None,
+  )
```
shoyer, line 166: Yes -- you can just do a normal Python `if` here. `b.size` is available as static information, e.g.,
```python
  if b.size == 1:
    return b / A
  else:
    return jnp.linalg.solve(
        A.reshape((-1, b.size)), b.ravel()
    ).reshape(b.shape)
```

tests/lax_control_flow_test.py
```diff
@@ -2050,29 +2106,54 @@ def sqrt_cubed(x, tangent_solve=scalar_solve):
...
-      return lax.custom_root(f, x0, oracle, vector_solve)
+      return lax.custom_root(f, x0, oracle, solve_method)
 
     rng = np.random.RandomState(0)
     a = rng.randn(2, 2)
     b = rng.randn(2)
     jtu.check_grads(linear_solve, (a, b), order=2,
-                    atol={np.float32: 1e-2, np.float64: 1e-11})
+                    atol={np.float32: 1e-2, np.float64: 1e-11},
+                    modes=["fwd"],
```
shoyer, line 2128: Does `rev` mode no longer work? If so, that would be a concern...

tests/lax_control_flow_test.py
```diff
@@ -2161,7 +2242,7 @@ def dummy_root_usage(x):
   def test_custom_linear_solve(self, symmetric):
 
     def explicit_jacobian_solve(matvec, b):
-      return lax.stop_gradient(jnp.linalg.solve(jax.jacobian(matvec)(b), b))
+      return jnp.linalg.solve(jax.jacobian(matvec)(b), b)
```
shoyer, line 2245: I do actually like keeping stop_gradient here (and in the other `custom_linear_solve` tests), becauses it ensures that we're testing the custom gradient instead of the gradient that is defiend by `jnp.linalg.solve`.

mbmccoy: Will add back.

tests/lax_control_flow_test.py
mbmccoy: Spent some time looking in to this. There's no regression introduced here, but the behavior is definitely not ideal. To be explicit, here's the status of the `fwd` and `rev` mode:

 - For `vector_solve`, both `fwd` and `rev` work, as before. I'll make sure to add the `rev` mode back for that function.
 - For `newton_raphson`, only `fwd` mode works with vectors in this test (but see below about scalars vs. vectors).

The key problem is that JAX traces the function during reverse mode differentiation, and it hits a `while_loop`, which is not reverse-mode differentiable. I've verified that it's a problem in the original code as well—for example, you can replace `vector_solve` with this equivalent function and trigger the same error:

```python
def vector_solve(f, y):
  # This while_loop is an identity operation on y:
  y = lax.while_loop(lambda _: False, lambda _: _, y)
  return jnp.linalg.solve(jax.jacobian(f)(y), y)
```

I believe that the JVP definition provides all of the necessary information for the VJP, so tracing seems completely unnecessary. Nevertheless, I'm having difficulty tracking down where and why the tracer attempts to trace in `rev` mode. 

My suspicion is that reverse mode uses tracing as a shortcut for verifying array shapes.


### Scalars vs. Vectors

If we replace `y` in [line 2150](https://github.com/google/jax/pull/8869/files#diff-c3f68aab0e05d9d2de62fd88802c96c19d9060f6d2500a7eb4266ee4139e717cR2150) with a scalar, then reverse mode works on the newton method as well.

I suspect that there is some optimization that shortcuts tracing in the case of a scalar. This would explain why both `binary_search` and `newton_rasphson` are reverse-mode differentiable in `test_custom_root_scalar` despite their `while_loop`.

tests/lax_control_flow_test.py
```diff
@@ -2050,29 +2106,54 @@ def sqrt_cubed(x, tangent_solve=scalar_solve):
...
+    def tangent_solve(g, y):
+      return linear_equation_solve(jax.jacobian(g)(y), y)
+
+    def nonlinear_solve(y):
+      f = lambda x: nonlinear_func(x, y)
+      x0 = -jnp.ones_like(y)
+      return lax.custom_root(f, x0, newton_raphson, tangent_solve)
+
+    rng = np.random.RandomState(0)
+    y = rng.randn(3, 1)
```
mbmccoy, line 2150: Note: If we use a scalar instead of an array (e.g., `y = 3.0`), reverse mode works in this test. See comment above.

tests/lax_control_flow_test.py
mbmccoy: For completeness, the reverse mode test for newton_raphson fails with error:

```
E     ValueError: Reverse-mode differentiation does not work for lax.while_loop or lax.fori_loop. Try using lax.scan instead.
```

mbmccoy:
Went ahead and updated per the comments. We were seeing a failure on an earlier commit in the X64 tests. This was resolved by decreasing the tolerance in `newton_raphson`. I checked locally by running tests with `JAX_ENABLE_X64=1` set.

This should be good to go, unless we need to chase down the `rev`-mode differentiation issue.

shoyer:
I'm going to take a look into the reverse mode issue, enabling this is one
of the major reasons for custom_root, so it would be really nice if it
worked.

tests/lax_control_flow_test.py
```diff
@@ -2044,34 +2094,58 @@ def sqrt_cubed(x, tangent_solve=scalar_solve):
...
-    a = rng.randn(2, 2)
-    b = rng.randn(2)
-    jtu.check_grads(linear_solve, (a, b), order=2,
-                    atol={np.float32: 1e-2, np.float64: 1e-11})
-
-    actual = jax.jit(linear_solve)(a, b)
-    expected = jnp.linalg.solve(a, b)
-    self.assertAllClose(expected, actual)
+  def test_custom_root_vector_with_solve_closure(self, solve_method, modes):
+    with jax.disable_jit():
```
shoyer, line 2109: This should be removed. (Side note: It also doesn't make a difference for disabling `while_loop` here, because `custom_root` traces the given solver into a JAXpr, which uses JIT compilation no matter what.)

tests/lax_control_flow_test.py
```diff
@@ -2044,34 +2094,58 @@ def sqrt_cubed(x, tangent_solve=scalar_solve):
...
-    self.assertAllClose(results, inputs ** 1.5, check_dtypes=False)
+    self.assertAllClose(results, inputs ** 1.5, check_dtypes=False, atol=1e-6, rtol=1e-6)
 
     results = jax.jit(sqrt_cubed)(5.0)
     self.assertAllClose(
         results, 5.0**1.5, check_dtypes=False, rtol={np.float64: 1e-7})
 
+  @parameterized.named_parameters([
+    {"testcase_name": "vector_solve", "solve_method": vector_solve, "modes": ["fwd", "rev"]},
+    {"testcase_name": "newton_raphson", "solve_method": newton_raphson, "modes": ["fwd"]},
```
shoyer, line 2105: The reason why this is failing is because the unit test is checking second order gradients (e.g., VJP  ​of JVP), and `newton_raphson` itself is not differentiable. So this is expected behavior -- custom_root can only support higher order auto-diff is the supplied linear solver is also differentiable.

This could be fixed in a couple of ways:
1. make `newton_raphson` itself differentiable with a custom gradient rule via `custom_root` (this feels a little overly complex)
2. only check first order gradients for this test (`order=1` in `check_grads`).
3. drop this casse entirely, because Newton-Raphson is already being tested sufficiently elsewhere.

My preferred option would be (2) or (3), but I would lean slightly towards (3)

shoyer:
@mbmccoy anything I could do to help get this across the finish line?

mbmccoy:
@shoyer Thanks, just got back from break. I'll take a look at it today and let you know if I'm blocked.

mbmccoy:
@shoyer Should be good to go now. Note that I also updated a line in an example to refer the reader to `lax.custom_root`. The prior wording implies that the functionality isn't readily available.

tests/lax_control_flow_test.py
```diff
@@ -95,6 +95,65 @@ def posify(matrix):
...
+
+def _linear_equation_solve(A, b):
+  """Solve A*x = b for x.
+
+  Needed because linalg.solve does not work for scalar
+  equations.
+  """
+  if b.size == 1:
+    return b / A
+  else:
```
shoyer, line 151: can we delete this part and only keep the second clause? I think it should work for scalars, too.

mbmccoy: Done. I went ahead and inlined the function in the two places it was used.

shoyer (approved):

mbmccoy:
@shoyer I'm now blocked on an "internal CI infrastructure error". Looks like that target was cancelled. Mind taking a look?

hawkinsp:
I think that particular failure is benign and will resolve itself if we update the PR. However there's another problem, namely the tests now fail on TPU. The two failures are:

```
LaxControlFlowTest.test_custom_root_scalar_binary_search

...
  File "jax/tests/lax_control_flow_test.py", line 2079, in test_custom_root_scalar
    self.assertAllClose(results, inputs ** 1.5, check_dtypes=False, atol=1e-6, rtol=1e-6)
...

AssertionError: 
Not equal to tolerance rtol=1e-06, atol=1e-06

Mismatched elements: 1 / 2 (50%)
Max absolute difference: 0.00037384
Max relative difference: 3.343618e-05
 x: array([ 8.     , 11.18034], dtype=float32)
 y: array([ 8.      , 11.180714], dtype=float32)
```

and
```
LaxControlFlowTest.test_custom_root_scalar_newton_raphson

 ...
  File "jax/tests/lax_control_flow_test.py", line 2079, in test_custom_root_scalar
    self.assertAllClose(results, inputs ** 1.5, check_dtypes=False, atol=1e-6, rtol=1e-6)
...
AssertionError: 
Not equal to tolerance rtol=1e-06, atol=1e-06

Mismatched elements: 1 / 2 (50%)
Max absolute difference: 0.00037384
Max relative difference: 3.343618e-05
 x: array([ 8.     , 11.18034], dtype=float32)
 y: array([ 8.      , 11.180714], dtype=float32)
```

I suspect the easiest thing to do is just to relax the test tolerances a little bit?

(Making an update to the PR should also retrigger the CI build that failed.)

mbmccoy:
Just getting back to this. Relaxed the tolerance for 32-bit floats:

```python
      atol={jnp.float32: 1e-3, jnp.float64: 1e-6},
      rtol={jnp.float32: 1e-3, jnp.float64: 1e-6},
```

mbmccoy:
@shoyer @hawkinsp looks like a maintainer still needs to kick off the testing workflow since it's my first contribution.

hawkinsp (approved):

shoyer:
Thank you @mbmccoy !

PR #8873, opened 2021-12-09
Add `contiguous_submeshes` option to `mesh_utils.create_device_mesh()`.

skye:
Unless you're using GlobalDeviceArrays, the device mesh passed to pjit
must be composed of contiguous submeshes for each process (i.e. each
process's local devices must all be next to each other in the full
mesh and form a rectangular submesh). This change teaches
`create_device_mesh` how to output meshes that satisfy this
constraint in some common cases.

This isn't the default behavior because the resulting meshes are a
little awkward and magical, and eventually we'd like using
GlobalDeviceArrays to be the common use case.

### DIFF
--- a/jax/experimental/mesh_utils.py
+++ b/jax/experimental/mesh_utils.py
@@ -16,7 +16,7 @@
 """Utils for building a device mesh."""
 
 import itertools
-from typing import Any, List, Sequence, Tuple
+from typing import Any, Dict, List, Sequence, Tuple
 
 from absl import logging
 import jax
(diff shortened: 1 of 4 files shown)


jax/experimental/mesh_utils.py
```diff
@@ -25,6 +25,35 @@
...
+        (1, 16, 4): (0, 2, 1),
+        (16, 4): (0, 2, 1),
+    },
+    (4, 8, 8): {
+        (1, 64, 4): (0, 2, 1),
+        (1, 4, 64): (0, 2, 1),
+        (64, 4): (0, 2, 1),
+        (4, 64): (0, 2, 1),
+    },
+    (8, 8, 8): {
```
jekbradbury, line 51: Missing (8, 16, 16), which is a valid and useful slice size where the trick also works

jekbradbury (approved):

jax/experimental/mesh_utils.py
skye: Nice, added that one.

jax/experimental/mesh_utils.py
```diff
@@ -143,6 +179,7 @@ def _jax_devices_order_normalized(
   # h_x, h_y can be 2x2 or 1x1 depending on tasks_per_host=4 or 1
   h_x, h_y, _, cores_per_chip = local_topology
   assert cores_per_chip == 1
+  assert (h_x, h_y) in [(1, 1), (2, 2)]
```
zhangqiaorjc, line 182: can you remove this assertion? certain system requires more generality here

skye: Done.

jax/experimental/mesh_utils.py
```diff
@@ -171,21 +227,30 @@ def create_device_mesh(mesh_shape: Sequence[int]) -> np.ndarray:
   local_jax_devices_from_process_0 = jax.local_devices(process_index=0)
```
zhangqiaorjc, line 227: nit: maybe use `process_0_devices` here too? and s/jax_devices/global devices on line 228 to be consistent?

jax/experimental/mesh_utils.py
```diff
@@ -156,7 +192,27 @@ def _jax_devices_order_normalized(
...
+    raise ValueError(
+        f"create_device_mesh cannot create contiguous submeshes for "
+        f"mesh_shape {mesh_shape} and physical mesh topology {topology}. "
+        f"Available mesh_shapes: {list(_TRANSPOSE_TRICKS[topology].keys())}")
+
+  return physical_mesh.transpose(*_TRANSPOSE_TRICKS[topology][mesh_shape])
+
+
+def create_device_mesh(mesh_shape: Sequence[int],
+                       contiguous_submeshes: bool = False) -> np.ndarray:
```
zhangqiaorjc, line 215: Maybe update docstring for the newly added bool parameter `contiguous_submeshes`?

jax/experimental/mesh_utils.py
```diff
@@ -156,7 +192,27 @@ def _jax_devices_order_normalized(
...
   """Creates a performant device mesh for jax.experimental.maps.mesh.
```
zhangqiaorjc, line 217: In the comments for `Returns`, maybe say jax *global* devices? i like how you made it explicit elsewhere that global devices need to be passed in

jax/experimental/mesh_utils.py
```diff
@@ -25,6 +25,42 @@
 _TPU_V3 = 'TPU v3'
 _TPU_V4 = 'TPU v4'
 
+# Maps physical topology -> mesh shape -> transpose to use for jekbradbury's
+# famous contiguous mesh trick. The trick only works for certain topologies and
```
zhangqiaorjc, line 29: nit: line break for this comment?

zhangqiaorjc (approved):
Great extension to handle cases where jax.devices can be transposed to be continuous submesh! Thanks Skye!

jax/experimental/mesh_utils.py
skye: Done.

PR #8890, opened 2021-12-10
Add missing annotations to lax

NeilGirdhar:


jax/_src/lax/lax.py
```diff
@@ -113,16 +115,16 @@ def _try_broadcast_shapes(shapes):
...
     return shapes[0]
   ndim = _max(len(shape) for shape in shapes)
-  shapes = [(1,) * (ndim - len(shape)) + shape for shape in shapes]
-  result_shape = _try_broadcast_shapes(shapes)
+  shape_list = [(1,) * (ndim - len(shape)) + shape for shape in shapes]
+  result_shape = _try_broadcast_shapes(shape_list)
   if result_shape is None:
     raise ValueError("Incompatible shapes for broadcasting: {}"
-                     .format(tuple(map(tuple, shapes))))
+                     .format(tuple(shape_list)))
```
NeilGirdhar, line 127: I modified the error display here because MyPy was [having trouble](https://mypy-play.net/?mypy=latest&python=3.10&gist=ec51e527a58bd719a3bc049e443c9c90).  However, I don't think `map` is necessary since if `shape_list` contains anything other than tuples, the tuple addition on line 123 won't work.

jax/_src/lax/lax.py
```diff
@@ -836,8 +838,10 @@ def argmax(operand: Array, axis: int,
   return argmax_p.bind(operand, axes=(axis,),
                        index_dtype=dtypes.canonicalize_dtype(index_dtype))
 
-def reduce(operands, init_values, computation: Callable,
-           dimensions: Sequence[int]):
+def reduce(operands: T,
```
hawkinsp, line 841: I don't think the type variable is quite accurate unfortunately. For example, it's legal to pass a scalar integer as an argument but get back a JAX array.

jax/_src/lax/lax.py
```diff
@@ -1083,7 +1087,7 @@ def _tri(dtype: DType, shape: Shape, offset: int) -> Array:
                 broadcasted_iota(np.int32, shape, 1))
   return convert_element_type_p.bind(bool_tri, new_dtype=dtype, weak_type=False)
 
-def stop_gradient(x):
+def stop_gradient(x: T) -> T:
```
hawkinsp, line 1090: Same here.

jax/_src/lax/lax.py
```diff
@@ -1111,7 +1115,7 @@ def stop(x):
       return x  # only bind primitive on inexact dtypes, to avoid some staging
   return tree_map(stop, x)
 
-def reduce_precision(operand, exponent_bits, mantissa_bits):
+def reduce_precision(operand: T, exponent_bits: int, mantissa_bits: int) -> T:
```
hawkinsp, line 1118: Same here.

hawkinsp (changes requested):

jax/_src/lax/lax.py
NeilGirdhar: Yeah, that's true.  It's also true for practically every Jax function that scalar can be silently promoted to shape `()` arrays.  I think it's still better to annotate such cases using a type variable since:
- in the case of scalars, most people will be using an annotation like [this one](https://github.com/NeilGirdhar/tjax/blob/master/tjax/_src/annotations.py#L35), and there won't be a type error, and
- in the case of dataclass pytrees, the type variable will correctly constrain the type.

What do you think?

jax/_src/lax/lax.py
NeilGirdhar: It seems to me that the caller can force the type in the rare case they're pushing scalars in:
```
x: RealNumeric = stop_gradient(2.0)
```
Also, pushing concrete (non-tracer) scalars into to `stop_gradient` doesn't really make sense.

jax/_src/lax/lax.py
jakevdp: I think no type annotations are better than incorrect type annotations.

NeilGirdhar: I wouldn't call them "incorrect".  It's no more incorrect than annotating something as `int` when actually it should be annotated as `numbers.Integral`.  At the end of the day, the point of annotations is to help users find bugs.  Annotating as `int` everywhere catches plenty of bugs.

Similarly, I'm annotating my Jax arrays as `numpy.typing.NDArray[np.integer[Any]]` (etc.) since there is no corresponding Jax type yet.  But because there's a bijection between Jax array types and Numpy array types, this catches practically all of the typing errors.

I understand the desire to make everything perfect.  I'm also a perfectionist.

What if we can find a solution with overloads that makes scalar inputs produce `Any` outputs, would that be acceptable to you?

NeilGirdhar: Something like this:
```python
from typing import Any, TypeVar, overload, Union

T = TypeVar('T')
Scalar = Union[int, float]

@overload
def f(x: Scalar) -> Any:
    ...

@overload
def f(x: T) -> T:
    ...

def f(x: Any) -> Any:
    return x

x = 2
reveal_type(f(x))  # Any
reveal_type(f((x,)))  # Tuple[builtins.int]
```

hawkinsp: Wouldn't a simpler and more correct type look like this:
```
def f(x: Union[Array, Scalar]) -> Array:
```
?

NeilGirdhar: @hawkinsp Yes, you're right.  (I'm assuming that you're correcting my first overload.)

hawkinsp: No, I'm saying "you don't need the overloads at all: that type suffices".

NeilGirdhar: You do need the overload because you can have aggregate pytree types like a `flax.struct.dataclass`.

NeilGirdhar: This is what may be lost in this discussion.  All of the Jax code that I write extensively uses dataclasses as aggregate pytree types.  This makes code much easier to read than using, for example, tuples.  It would be nice to have type inference work for these types.  That means that if `x` is an instance of `SomeClass`, then `stop_gradient(x)` should also be `SomeClass`.

jax/_src/lax/lax.py
NeilGirdhar: On further inspection, it appears that `stop_gradient` seems to return concrete values unchanged, so this annotation should be fine?

jax/_src/lax/lax.py
NeilGirdhar: I added an overload for `float`.  This function doesn't accept integers or complex dtypes.

NeilGirdhar:
@hawkinsp Would you mind looking over this again?  I've adjusted the code in response to your comments.

jax/_src/lax/lax.py
NeilGirdhar: I removed the type variable for this function since as Jake points out "having no type annotations would be better than having incorrect type annotations".

hawkinsp (approved):

hawkinsp:
(Closing and reopening to force a CI rebuild.)

hawkinsp:
This PR appears to cause test failures under pytype, which we also check with (internally, although I would imagine you can reproduce by just running `pytype` on the PR). I found this patch fixes it:

```
--- a/google3/third_party/py/jax/_src/lax/lax.py
+++ b/google3/third_party/py/jax/_src/lax/lax.py
@@ -99,13 +99,14 @@ def _array_copy(arr):
     return bitwise_or(arr, _const(arr, False))
   return add(arr, _const(arr, 0))

-def _try_broadcast_shapes(shapes):
+def _try_broadcast_shapes(
+    shapes: Sequence[Tuple[int, ...]]) -> Optional[Tuple[int, ...]]:
   assert shapes
   if len(shapes) == 1: return shapes[0]
   rank, *others = {len(shape) for shape in shapes}
   if others: return None  # must have consistent rank
   if not rank: return ()  # scalar case
-  result_shape = [None] * rank
+  result_shape = [-1] * rank
   for i, sizes in enumerate(zip(*shapes)):
     non_1s = set([d for d in sizes if not core.symbolic_equal_dim(d, 1)])
     if len(non_1s) > 1:
@@ -125,7 +126,8 @@ def broadcast_shapes(*shapes: Tuple[int,
   if result_shape is None:
     raise ValueError("Incompatible shapes for broadcasting: {}"
                      .format(tuple(shape_list)))
-  return result_shape
+  else:
+    return result_shape

 def _identity(x): return x
```

Mind making those changes?

(Sorry for the slow review, I was out for the holidays.)

NeilGirdhar:
No problem, and done.

> -  return result_shape
> +  else:
> +    return result_shape

I didn't make this change though since I don't understand how it would change anything?  Most linters complain about the unnecessary else.

NeilGirdhar:
pytype passes on my machine

NeilGirdhar:
```shell
$ pytype jax/_src/lax/
Computing dependencies
Analyzing 13 sources with 138 local dependencies
ninja: Entering directory `.pytype'
[118/238] check jax._src.lax.lax
[231/238] check jax._src.lax.polar
FAILED: /Users/neil/src/jax/.pytype/pyi/jax/_src/lax/polar.pyi
/Users/neil/.pyenv/versions/3.7.12/bin/python3.7 -m pytype.single --imports_info /Users/neil/src/jax/.pytype/imports/jax._src.lax.polar.imports --module-name jax._src.lax.polar -V 3.7 -o /Users/neil/src/jax/.pytype/pyi/jax/_src/lax/polar.pyi --analyze-annotated --nofail --quick /Users/neil/src/jax/jax/_src/lax/polar.py
File "/Users/neil/src/jax/jax/_src/lax/polar.py", line 27, in <module>: Couldn't import pyi for 'jax.scipy' [pyi-error]
  Unreplaced NamedType: 'jax.scipy.fft', referenced from 'jax.scipy'

For more details, see https://google.github.io/pytype/errors.html#pyi-error
ninja: build stopped: subcommand failed.
Leaving directory '.pytype'
```

PR #8907, opened 2021-12-11
[hcb] Add support for remat2 to host_callback

gnecula:
A callback under ad_checkpoint.checkpoint will be invoked
twice when taking the gradient: once during the forward pass
and once again during the backward pass when the residuals
for the forward pass are rematerialized.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -1678,6 +1678,32 @@ call_p.def_impl(call_impl)
 named_call_p: CallPrimitive = CallPrimitive('named_call')
 named_call_p.def_impl(call_impl)
 
+outfeed_primitives: Set[Primitive] = set()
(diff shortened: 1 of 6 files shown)


jax/core.py
```diff
@@ -1678,6 +1678,32 @@ def call_impl(f: lu.WrappedFun, *args, **params):
 named_call_p: CallPrimitive = CallPrimitive('named_call')
 named_call_p.def_impl(call_impl)
 
+outfeed_primitives: Set[Primitive] = set()
```
gnecula, line 1681: I moved this stuff from dispatch to core so that we can access it from partial_eval.

mattjj (approved):

PR #8915, opened 2021-12-12
fix a custom_vjp post_process bug, related cleanups

### LINKED ISSUE #8783: grad + vmap + odeint AssertionError
### Discussed in https://github.com/google/jax/discussions/8782

<div type='discussions-op-text'>

<sup>Originally posted by **DanPuzzuoli** December  2, 2021</sup>
Hi,

I've seen a bunch of discussions and issues surrounding this so I apologize if i'm re-raising something that has already been addressed elsewhere. I don't understand the internals of JAX enough to understand if this is some version of an issue that's already raised, though a lot of what I'm seeing is in already closed issues so I assume are solved and hence this is different.

When I run the following code on the latest release versions of jax/jaxlib (a self-contained version of my actual code):
```
from jax import jit, value_and_grad, vmap
import jax.numpy as jnp
from jax.config import config
config.update('jax_platform_name', 'cpu')


T = 1.

X = -1j * jnp.array([[0., 1.], [1., 0.]], dtype=complex)
Y = -1j * jnp.array([[0., -1j], [1j, 0.]], dtype=complex)

def err_obj(a, b_vals):
    
    def err(b):
        def rhs(y, t):
            return (b * X + a * (t**2) * Y) @ y
        
        res = odeint(rhs, y0=jnp.eye(2, dtype=complex), t=jnp.array([0, T], dtype=complex), rtol=1e-6, atol=1e-6)
        
        return jnp.abs((X * res[-1]).sum())**2 / 4


    all_err = vmap(err)(b_vals)
    return all_err.sum()

b_vals = jnp.array([1., 2., 3., 4., 5.])
jit(value_and_grad(lambda a: err_obj(a, b_vals)))(1.)
``` 

I get the error:
```
AssertionError: length mismatch: [1, 4]
```

I've definitely reverse-mode differentiated this code in the past with success, though it was some time ago so would have been on a much older version of jax/jaxlib. Based on other similar issues I've seen it seems like this has something to do with the interaction of reverse-mode autodiff, vmap, and the control flow used in `odeint`, but again, the issues I've seen raising these kinds of errors seem to have been solved?

Thanks!</div>

mattjj:
related to #8783 (fixes one bug but another remains)

There are a few themes here, the first two being about cleanup and the latter being most directly related to the fix:
1. Some functions, like `core.process_env_traces` and `core.call_bind`, were reused for many similar but not identical applications: for final-style calls (like `core.call_p` and `xla_call_p`), final-style maps (like `pmap_p` and `xmap_p`), and for final-style custom differentiation rules (like `custom_jvp` and `custom_vjp`). That made sense at first because these various instances looked pretty similar, but over time they diverged, and as a result we were packing too much polymorphic functionality into functions like `call_bind` and `process_env_traces`, often under `if`/`else` switches. By splitting these into separate functions the logic of each becomes much simpler, and more importantly easier to follow in the source (without mentally modeling the different types and indirection which might be involved at runtime). We were also able to remove some of the indirection (like the `process` and `post_process` methods on call primitives), and clarify some function signatures (both on the input and output side). Finally, it made this fix easier because this fix involved introducing different signatures for the post-process stuff related to `custom_jvp` and `custom_vjp`.
2. Remove `_IgnoreElemList`. Unless I'm mistaken, I think it was just another mechanism for accomplishing what our `linear_util.Strore`s already do. Maybe this unification was obscured before because of the polymorphism/indirection mentioned above (i.e. having to support the same utility function signatures for both calls and maps).
3. Generalize `post_process_custom_jvp_call` and `post_process_custom_vjp_call` to depend on whether the primal function or the differentiation rule was run. At a high level, `custom_jvp` and `custom_vjp` are call-like primitives. But instead of being parameterized by a single callable, they're parameterized by two, only one of which is run: the primal function (if no differentiation happens) and the JVP or VJP rule (if differentiation does happen). The bug in #8783 arose because I had mistakenly reused the same `post_process` machinery for both cases, but the behaviors need to be different in the two cases (i.e. whether differentiated or not). The heart of this PR is just plumbing that information: for the JVP side I just added a new boolean argument to `post_process_custom_jvp_call` indicating which side was taken (`jvp_was_called`, which is True if differentiation happened and False otherwise), while for the VJP side the signatures for the two cases were different enough that I changed the API to have two different functions, namely `post_process_custom_vjp_call` and `post_process_custom_vjp_call_fwd` (where the former is called when no differentiation happened, and the latter otherwise).

As ever with final-style primitives, the control flow here can be a little tricky to follow.

There's more cleanup I'd like to do along these lines, especially for map primitives (like simplifying `JaxprTrace.process_call` not to handle both maps and calls). But I'll leave that for future work.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -4955,6 +4955,28 @@ class CustomJVPTest(jtu.JaxTestCase):
 
     api.vmap(sample)(jax.random.split(jax.random.PRNGKey(1), 3))  # don't crash
 
+  def test_closure_with_vmap2(self):
+    # https://github.com/google/jax/issues/8783
+    def h(z):
+      def f(x):
+        @jax.custom_jvp
+        def g(y):
+          return x * y
+
+        # NOTE: rule closes over vmap tracer
+        @g.defjvp
+        def g_jvp(primals, tangents):
+          (y,), (ydot,) = primals, tangents
+          return x * y, x * ydot
+
+        return g(z)  # NOTE: no vmapped arg
+
+      return jax.vmap(f)(jnp.arange(3., dtype='float32'))
+
+    primals, tangents = jax.jvp(h, (jnp.float32(1.),), (jnp.float32(2.),))
+    self.assertAllClose(primals ,     jnp.arange(3., dtype='float32'))
+    self.assertAllClose(tangents, 2 * jnp.arange(3., dtype='float32'))
+
   @unittest.skipIf(numpy_version == (1, 21, 0),
                    "https://github.com/numpy/numpy/issues/19305")
   def test_float0(self):
@@ -6198,6 +6220,27 @@ class CustomVJPTest(jtu.JaxTestCase):
 
     jtu.check_grads(batched_scan_over_mul, (x_batch, coeff), order=2,
                     modes=['rev'])
+
+  def test_closure_with_vmap2(self):
+    # https://github.com/google/jax/issues/8783
+    def h(z):
+      def f(x):
+        @jax.custom_vjp
+        def g(y):
+          return x * y
+
+        def g_fwd(y):
+          return x * y, (x, x * y, y)
+        def g_rev(res, w_bar):
+          x, *_ = res
+          return (x * w_bar,)
+        g.defvjp(g_fwd, g_rev)
+
+        return g(z)
+
+      return jax.vmap(f)(jnp.arange(3., dtype='float32')).sum()
+
+    jtu.check_grads(h, (jnp.float32(3.14),), order=1, modes=['rev'])
 
 
 def transpose_unary(f, x_example):
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -274,10 +274,10 @@ class CustomJVPCallPrimitive(core.CallPrimitive):
   def bind(self, fun, jvp, *args):
     args = map(core.full_lower, args)
     top_trace = core.find_top_trace(args)
-    fun, env_trace_todo1 = core.process_env_traces(
-        fun, self, top_trace and top_trace.level, (), None)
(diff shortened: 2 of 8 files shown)


tests/api_test.py
```diff
@@ -4882,6 +4882,28 @@ def f_jvp(primal, tangent):
 
     api.vmap(sample)(jax.random.split(jax.random.PRNGKey(1), 3))  # don't crash
 
+  def test_closure_with_vmap2(self):
+    # https://github.com/google/jax/issues/8783
+    def h(z):
+      def f(x):
+        @jax.custom_jvp
+        def g(y):
+          return x * y
```
froystig, line 4891: Maybe a test that returns only `x` here? Since the implementation of `*` might involve call primitives for transformations (namely `xla_call` for `jit`).

froystig (approved):
Looks good!

> 3. Generalize `post_process_custom_jvp_call` and `post_process_custom_vjp_call` to depend on whether the primal function or the differentiation rule was run. At a high level, `custom_jvp` and `custom_vjp` are call-like primitives. But instead of being parameterized by a single callable, they're parameterized by two, only one of which is run: the primal function (if no differentiation happens) and the JVP or VJP rule (if differentiation does happen).

I suspect that we'll need something similar in other incoming "custom transformation" work, such as custom batching and custom transposition. Our current sketches of these require a similar pattern: a call-like primitive with two call targets, e.g. a "primal" function and its custom-defined transpose. Similarly to custom JVP/VJP, the transformation context determines which target is evaluated, e.g. the primal function (if no transposition is taking place) and the transpose (under transposition).

> There's more cleanup I'd like to do along these lines, especially for map primitives (like simplifying `JaxprTrace.process_call` not to handle both maps and calls). But I'll leave that for future work.

That sounds nice.

PR #8926, opened 2021-12-13
fix: thread fill_value option

### LINKED ISSUE #8928: Jax options not threaded through to base implementation in jnp
**Bug**
`jnp.flatnonzero` and `jnp.argwhere` mention `fill_value` option in documentation (see image below), but calling with `fill_value` leads to 'unexpected keyword argument' error.

In the below code, the last two lines will raise errors due to unexpected argument `fill_value`
```python
import jax.numpy as jnp
d = jnp.array([True,False,True])
jnp.nonzero(d, size=20, fill_value=100)
jnp.flatnonzero(d, size=20, fill_value=100)
jnp.argwhere(d, size=20, fill_value=100)
```

**Why?**
The implementation of `jnp.flatnonzero` and `jnp.argwhere` link back to `jnp.nonzero` but the `fill_value` argument wasn't threaded through (just the `size` argument).

The threading is there in `jnp.where` and other cases, just those two have been missed as far as I've noticed.

**Fix**
Pull request:
#8926 
just threads the argument

Images

image

image

image

oliverdutton:
Fixes #8928

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2875,8 +2875,8 @@ def nonzero(a, *, size=None, fill_value=None):
   return out
 
 @_wraps(np.flatnonzero, lax_description=_NONZERO_DOC)
-def flatnonzero(a, *, size=None):
-  return nonzero(ravel(a), size=size)[0]
+def flatnonzero(a, *, size=None, fill_value=None):
+  return nonzero(ravel(a), size=size, fill_value=fill_value)[0]
 
 
 def _nan_reduction(a, name, jnp_reduction, init_val, nan_if_all_nan,
@@ -5212,8 +5212,8 @@ nonzero elements than `size` indicates, the index arrays will be zero-padded.
 """
 
 @_wraps(np.argwhere, lax_description=_ARGWHERE_DOC)
-def argwhere(a, *, size=None):
-  result = transpose(vstack(nonzero(a, size=size)))
+def argwhere(a, *, size=None, fill_value=None):
+  result = transpose(vstack(nonzero(a, size=size, fill_value=fill_value)))
   if ndim(a) == 0:
     return result[:0].reshape(result.shape[0], 0)
   return result.reshape(result.shape[0], ndim(a))
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -1058,8 +1058,8 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
         return result[:size]
       else:
         fillvals = fill_value if np.ndim(fill_value) else result.shape[-1] * [fill_value or 0]
-      return np.stack([np.concatenate([arg, np.full(size - len(arg), fval, arg.dtype)])
-                  for fval, arg in safe_zip(fillvals, result.T)]).T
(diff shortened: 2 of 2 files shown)


hawkinsp (commented):
Two comments:
(a) we try not to diverge from upstream NumPy. Here, if the change makes sense in jax.numpy, I think the same change would also make sense in upstream NumPy as well. Perhaps propose the same API change in NumPy as well?

(An alternative would be to fork these functions into your own code with your extension.)

(b) we'd need a test for the new feature!

hawkinsp:
Actually, I take it back. `fill_value` is a JAX extension. In which case, I think all we need is a test...

oliverdutton:
No worries, thanks for the quick reply.

It's the same as `size`, Jax is adding it in to define padding behaviour for jit'ing.

oliverdutton:
On the test front. The current test doc shrugs off testing to the underlying `jnp.nonzero` call for `size`.`fill_value` comes under the same scope?

image

hawkinsp:
I'd say it's worth adding a test, because apparently it didn't actually work before despite being documented (hence your PR!) A test is your best guarantee this will stay working.

oliverdutton:
Very good point

oliverdutton:
I've added tests similar to the ones that are there. 
 
Padding passes through as expected.

```python
x = jnp.arange(9).reshape(3,3)%5
jnp.argwhere(x, size=10, fill_value=(-1,-2)) # Works with both tuple and plain old int padding
jnp.flatnonzero(x, size=10, fill_value=-1) # Forces int padding, as it should
```

The numpy test padding functions apply in this way too.

image

hawkinsp (approved):

hawkinsp:
It looks like there are some test failures.

oliverdutton:
I've reviewed the fails in https://source.cloud.google.com/results/invocations/aee37748-1efc-4414-befe-32f1b95f5083/targets/jax%2Ftesting%2Finternal_numpy_scipy%2Fpresubmit_github/log

16 are argwhere:
- 11 are due to the shape inference being taken from the wrong dim, 0 rather  1 in the argwhere output.
- 5 are due to np.stack being passed a generator, not a sequence. This causes a `FutureWarning`. I've adjusted so the generator is evaluated to a list
- (There's also an edge case `shape=int32[]_size=1_fill_value=(1,)` where argwhere isn't appropriate. It spits out a (size,0) empty array. This seems reasonable behaviour, the numpy padding function now matches the jax function.)

7 are flatnonzero:
- All 7 are due to example with multidimensional fill values being given as a test case. i.e. fill_value=(3,4). The flatnonzero output is flat, so thats nonsensical in this context.

Commits that should fix those cases have been made. Please try again.

hawkinsp:
Looks like one more tiny fix is needed: `flake8` is unhappy.

oliverdutton:
I think I've fixed flake8's unhappiness, please try again.

hawkinsp (approved):

oliverdutton:
I'll install flake8 and actually check this time, rather than guess.

oliverdutton:
This time I've verified locally it passes flake8

hawkinsp (approved):

hawkinsp:
Looks like there's still one flake8 warning :-)

oliverdutton:
Trailing whitespace removed

hawkinsp:
Thanks for the PR!

PR #8934, opened 2021-12-14
Tweak documentation on error handling for host_callback.call()

juesato:
Delineate which aspects apply specifically to CPU/GPU vs TPU.


gnecula (approved):
Thanks

PR #8938, opened 2021-12-14
jnp.array: remove unused device argument

jakevdp:
It looks like this argument was inadvertently added as part of #8016; it is undocumented and not referenced within the function.


hawkinsp (approved):

hawkinsp:
There's a small risk this could break users, but I agree it's not large.

PR #8948, opened 2021-12-14
jnp.array: use jax-style promotion for list inputs

### LINKED ISSUE #8945: jnp.array breaks JIT invariance for some list inputs
Minimal repro:
```python
import numpy as np
import jax.numpy as jnp
from jax import jit

x = [0, np.int16(1)]

print(jnp.array(x).dtype)
# int32

print(jit(jnp.array)(x).dtype)
# int16
```
The reason for this is that, outside JIT, `jnp.array(x)` falls back to `np.array(x)`, presumably for performance reasons, which results in numpy-style type promotion being applied to the list contents in place of JAX-style type promotion: https://github.com/google/jax/blob/2f4bd118839b29bdf9aa84f8c5b351ce6d7bff71/jax/_src/numpy/lax_numpy.py#L3589-L3594

Within JIT, however, the list contents are traced and `_can_call_numpy_array` evaluates to False, meaning that the final dtype is determined by `jnp.stack`, which uses JAX type promotion to decide on the output type: https://github.com/google/jax/blob/2f4bd118839b29bdf9aa84f8c5b351ce6d7bff71/jax/_src/numpy/lax_numpy.py#L3614-L3616

I think the best fix here would be to use JAX-style type promotion in both cases.

jakevdp:
Fixes #8945

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -475,25 +475,6 @@ _INT_DTYPES = {
   64: np.int64,
 }
 
-def _np_array(obj, dtype=None, **kwargs):
-  """Return a properly-typed numpy array.
-
-  `_np_array(obj, **kwds)` is equivalent to `np.array(obj, **kwds)`, with the
-  exception that when obj.dtype is not defined and dtype is not specified, it
-  uses Jax's default dtypes.
-  """
-  arr = np.array(obj, dtype=dtype, **kwargs)
-  obj_dtype = getattr(obj, 'dtype', None)
-  arr_dtype = np.dtype(arr.dtype).type
-  if dtype is None and obj_dtype is None:
-    if dtypes.is_python_scalar(obj):
-      arr = arr.astype(result_type(obj))
-    elif arr_dtype in _DEFAULT_TYPEMAP:
-      arr = arr.astype(_DEFAULT_TYPEMAP[arr_dtype])
-  return arr
-
-_np_asarray = partial(_np_array, copy=False)
-
 def _promote_shapes(fun_name, *args):
   """Prepend implicit leading singleton dimensions for Numpy broadcasting."""
   if len(args) < 2:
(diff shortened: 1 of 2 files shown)


jax/_src/dtypes.py
```diff
@@ -343,7 +343,10 @@ def dtype(x, *, canonicalize=False):
   elif type(x) in python_scalar_dtypes:
     dt = python_scalar_dtypes[type(x)]
   else:
-    dt = np.result_type(x)
+    try:
+      dt = np.result_type(x)
+    except TypeError:
+      dt = np.asarray(x).dtype
```
mattjj, line 349: Is this change used by this PR? What's the case this line is meant to handle?

jax/_src/numpy/lax_numpy.py
```diff
@@ -3583,28 +3564,39 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0):
   # check if the given dtype is compatible with JAX
   lax._check_user_dtype_supported(dtype, "array")
 
+  # Here we make a judgment call: we only return a weakly-typed array when the
+  # input object itself is weakly typed. That ensures asarray(x) is a no-op whenever
+  # x is weak, but avoids introducing weak types with something like array([1, 2, 3])
```
mattjj, line 3569: Would it be feasible/desirable for `asarray(x)` always to return a strong type, even if its input is weakly typed? That is, not to ensure that `asarray(x)` is a no-op when `x` is weak? I ask in the spirit of having the simplest policy possible.

mattjj (approved):

jax/_src/dtypes.py
jakevdp: Yes - we now pass memoryview objects to `_lattice_result_type`, and causes a TypeError in `np.result_type`.

jax/_src/numpy/lax_numpy.py
jakevdp: I'm exploring that as well, but didn't include that in this PR. The behavior described in the comment is the current behavior (the comment is moved from further down)

jakevdp: We could go two ways here: either have `jnp.array` always return strong types, or have `jnp.array` always respect the weak type coming from the type promotion of the inputs. I'm still unsure which would be best, and I want to run the full internal test suite on both versions to help think through the consequences.

jax/_src/dtypes.py
jakevdp: I ended up reverting this change and handling it in the `array` function. If `np.result_type` raises a TypeError for an input, then `jnp.result_type` should probably do so as well.

PR #8953, opened 2021-12-14
Prototype of PRNG Key re-use checking - eager checks

LenaMartens:
### Introduction

This check is to address a common pitfall in JAX where it's easy to re-use a PRNG key and get the same random bits unintentionally (see https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#jax-prng).

This PR adds a way to check if you've re-used a key when executing code eagerly. It tracks the consumption of a PRNG key, and will throw an error when a key is used after is has already been consumed. A key is consumed if it is passed into a `jax.random.split` call, or any `jax.random` call which generates random_bits from the key (eg. `jax.random.normal`). The error will tell you where the key was consumed for the first time.

```
KeyReuseError: Re-used a key, the key was previously used at (...)file.py:function (73)
```

If you want to re-use a key, you can broadcast it to make copies of the key. The idea is that re-use should be avoided, unless you are explicit about copying the key. We could brain-storm about additional copying APIs, but currently a `keys = jnp.broadcast_to(key, (2,))` will work as a copy.  

### Coverage

This is an incomplete implementation of this kind of check, and does not cover all cases of re-use. As a way of documenting coverage, I've included currently failing cases in `tests/key_reuse_test.py` (all tests decorated with `@unittest.expectedFailure`). These tests can be used to encode the desired final behavior of this check (eg. `test_vmap_broadcast_key_cross_batch`, should we expect this case to pass the re-use check?)

With the current implementation, the check will work if you use a key twice while running eagerly, or if you use a key twice while tracing a function.
For example:

```python
# example 1 - eager
key = random.PRNGKey(123)
with jax.config.debug_prng_key_reuse(True):
  x0 = random.normal(key, ())
  x1 = random.normal(key, ())  # key re-use error!

# example 2 - re-use while tracing
@jax.jit
def f(k):
  x0 = random.normal(k, ())
  x1 = random.normal(k, ())  # key re-use error!

key = random.PRNGKey(456)
with jax.config.debug_prng_key_reuse(True):
  f(k)
```
However, re-use checking will **not** work if you need to track re-use across a staged-out boundary (ie. the key is _not_ re-used while tracing). These cases will need to be handled in a follow-up:
```python
@jax.jit
def f(k):
  x0 = random.normal(k, ())
  return x0

key = random.PRNGKey(123)
with jax.config.debug_prng_key_reuse(True):
  f(k)
  f(k)  # no error!
```
### Implementation details

Some details worth highlighting:
- `PRNGKeyArray._split` consumes itself, but `PRNGKeyArray._random_bits` does not. For `random_bits` the `consume` call is lifted to the wrapper functions in `jax.random`. This is because of the technicality that all `jax.random` calls are internally jitted. Because this PR does not yet handle carrying the consumption flag across the `jit` boundary, the consumption needs to happen outside of that boundary.
- The check is behind a flag called `jax_debug_prng_key_reuse`, which I see as analogous to a flag like `jax_debug_nans`. However, it also needs the new PRNGKeyArray type (enabled through `jax_enable_custom_prng`) to work.
-  If we allow for using the same key on both branches of a `cond` (because semantically only one branch is ever taken), then there is a false positive if those two branches _close over_ the key (see `test_cond_uses_on_both_branches_closed`). This seems undesirable, but maybe we should discourage closing over keys in any case,  or maybe we want to always conservatively reject using one key on both branches of the cond. This is still an open question for me.

---- 
Happy to discuss any of the assumptions made above, but also happy to land this preliminary check and refine in follow-up PRs while it's behind a flag. I think the most interesting point of discussion is the cases in `tests/key_reuse_test.py`, and what the expected behavior for them is (eg. wrt to scan/vmap/slicing into a key array).

### DIFF
--- a/jax/_src/errors.py
+++ b/jax/_src/errors.py
@@ -546,3 +546,43 @@ class UnexpectedTracerError(JAXTypeError):
 
   def __init__(self, msg: str):
     super().__init__(msg)
+
+
+class KeyReuseError(JAXTypeError):
+  """This error occurs when you re-use a PRNGKey.
+
+  .. note::
+    This feature is experimental (it does not catch all cases of key re-use,
+    most importantly around JAX control-flow and across JAX transformation
+    boundaries).
+    It needs the `jax_enable_custom_prng` flag to be set to True.
+
+  JAX's random API uses explicit PRNG keys which are passed into random calls.
+  This means it's possible to re-use a PRNG key, and without realizing it get
+  the same random bits out of two separate calls to a random function::
+
+    >>> from jax import random
+    >>> key = random.PRNGKey(123)
+    >>> random_bits0 = random.normal(key, ())
+    >>> random_bits1 = random.normal(key, ())
+    >>> random_bits0 == random_bits1
+    DeviceArray(True, dtype=bool)
+
+  See also the
+  `documentation here about JAX PRNG <https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#jax-prng>`_
+  on why this is a common pitfall.
+
+  With PRNG re-use checking enabled the above would fail. Currently this check
+  is behind a debug flag (by setting jax_debug_prng_key_reuse to True), or you
+  can use it as a context manager::
+
+    >>> from jax import random
+    >>> with jax.enable_custom_prng(True):
+    ...   with jax.debug_prng_key_reuse(True):
+    ...     key = random.PRNGKey(123)
+    ...     random_bits0 = random.uniform(key)
+    ...     random_bits1 = random.uniform(key)  # doctest: +IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+        ...
+    KeyReuseError: Re-used a key, the key was previously used at [...]
+  """
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -230,7 +230,9 @@ def uniform(key: KeyArray,
                      f"got {dtype}")
   dtype = dtypes.canonicalize_dtype(dtype)
   shape = core.as_named_shape(shape)
-  return _uniform(key, shape, dtype, minval, maxval)  # type: ignore
+  out = _uniform(key, shape, dtype, minval, maxval)  # type: ignore
+  key.consume()
+  return out
 
 @partial(jit, static_argnums=(1, 2), inline=True)
 def _uniform(key, shape, dtype, minval, maxval) -> jnp.ndarray:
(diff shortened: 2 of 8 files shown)


jax/_src/errors.py
```diff
@@ -546,3 +546,43 @@ class UnexpectedTracerError(JAXTypeError):
...
+  `documentation here about JAX PRNG <https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#jax-prng>`_
+  on why this is a common pitfall.
+
+  With PRNG re-use checking enabled the above would fail. Currently this check
+  is behind a debug flag (by setting jax_debug_prng_key_reuse to True), or you
+  can use it as a context manager::
+
+    >>> from jax import random
+    >>> with jax.enable_custom_prng(True):
+    ...   with jax.debug_prng_key_reuse(True):
```
LenaMartens, line 581: this need for a double context manager is not as nice as I would have liked, and makes me think maybe I should create this documentation _after_ this check is less experimental, WDYT?

froystig: How about having `debug_prng_key_reuse` require `enable_custom_prng`? It can either err if it isn't already enabled, or enable it and issue a warning that it has done so. That might be convenient anyway, but it would also allow this documentation to be written today the way that we expect it to look in the longer term.

froystig (approved):
Nice! And a very helpful PR description once again. Great idea writing out questions in the form of tests with TODOs.

jax/_src/random.py
```diff
@@ -230,7 +230,9 @@ def uniform(key: KeyArray,
                      f"got {dtype}")
   dtype = dtypes.canonicalize_dtype(dtype)
   shape = core.as_named_shape(shape)
-  return _uniform(key, shape, dtype, minval, maxval)  # type: ignore
+  out = _uniform(key, shape, dtype, minval, maxval)  # type: ignore
```
gnecula, line 233: Seeing all the use of `key.consume`, I am wondering if the code ends up more readable with a slightly different API where `key.consume()` is the identity function, and you write `return _uniform(key.consume(), shape, ...)`

tests/key_reuse_test.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # TODO(lenamartens): Is in_axes=(None,) an explicit broadcast?
+    # Should this behave similar to scan_reusing_inp?
+    jax.vmap(f, in_axes=(None, 0))(key, jnp.ones((2,)))
+
+  # TODO(lenamartens) fix false positives and false negatives.
+  # ---- false positives ----
+  @unittest.expectedFailure
+  def test_cond_uses_on_both_branches_closed(self):
+    # TODO(lenamartens): Is this a false positive? Yes, if we allow re-use on
```
gnecula, line 215: If it does not complicate the checking too much, I think that it is good to allow re-use on different branches of a conditional.

jakevdp:
Closing, since this is superseded by the work in `jax.experimental.key_reuse`

PR #8969, opened 2021-12-16
[JAX] Update sparse_bcoo_matvec benchmark test with indicies dtype.

tlu7:


tlu7:
@yashk2810 @jakevdp

yashk2810:
Can you show me the results? (best if you run it internally)

jakevdp:
One idea here: we could swap out this code for `sparse.random_bcoo`, which didn't exist when I wrote the benchmark: https://github.com/google/jax/blob/0d71bff7b974bc037dc873beb4bfd5be53350de9/jax/experimental/sparse/random.py#L25-L26

We could add an `indices_dtype` argument to that to control this dtype.

tlu7:
> Can you show me the results? (best if you run it internally)

name                                     cpu/op
sparse_bcoo_matvec                       8.12ms ± 1%
sparse_bcoo_matvec_jit                   37.7µs ±11%
sparse_bcoo_matvec_compile               46.7ms ±13%

tlu7:
Duplicate [[JAX] Requires indices to be sorted and of int32 in _sparse_bcoo_matvec test. #9021](https://github.com/google/jax/pull/9021)

Issue #8967, opened 2021-12-15
[sparse] add higher-order primitive for applying a function to specified elements


jakevdp:
I imagine an API that is something like this:
```python
>>> import jax.numpy as jnp
>>> from jax.experimental import sparse
>>> mat = sparse.BCOO.fromdense(jnp.arange(4))
>>> mat2 = sparse.apply_to_specified(mat, lambda x: x + 1)
>>> mat2.todense()
DeviceArray([0, 2, 3, 4], dtype=int32)
```
perhaps it could be a primitive that is defined for dense inputs, with an appropriate sparsification rule.

jakevdp:
Thinking more about this... really what we want is a way to efficiently sparsify a function that looks like this:
```python
def apply(f, x): 
  return jnp.where(x == 0, x, f(x)) 
```
The jaxpr for this function (disabling JIT to avoid call primitives) looks like this:
```python
x = jnp.arange(5.0)
with jax.disable_jit(): 
  print(jax.make_jaxpr(partial(apply, jnp.cos))(x))
# { lambda ; a:f32[5]. let
#     b:bool[5] = eq a 0.0
#     c:f32[5] = cos a
#     d:f32[5] = select b a c
#   in (d,) }
```
Our `sparsify` transform works one primitive at a time, so it would hit the `cos` primitive (which maps zeros to ones) and raise `NotImplemented`.

I'm not sure it's feasible with `sparsify()` to do multi-primitive compilation... but perhaps the MLIR compiler could handle something like this in the future.

That being said, the best way forward may be to define a `sparse.apply` primitive, which essentially does the above on dense matrices, but has a sparsification rule that does the appropriate thing on sparse matrices. #8991 is a draft of this idea.

jakevdp:
Closing, because we're not planning to further develop `jax.experimental.sparse` at the moment.

PR #8991, opened 2021-12-16
[sparse] add sparse.map_specified_elements primitive

jakevdp:
This PR addresses #8967 (more context in https://github.com/google/jax/issues/8967#issuecomment-996152905).  The idea here is to provide an API whereby a user can apply a unary function elementwise to the defined entries of a sparse matrix. like the `sparse.todense_p` primitive, it must also accept dense inputs so that it can be used within a function that is transformed by sparsify.

For example:
```python
>>> from jax.experimental import sparse
>>> import jax.numpy as jnp

>>> x = sparse.BCOO.fromdense(jnp.arange(4.0))
>>> print(x.todense())
[0. 1. 2. 3.]

>>> print(sparse.sparsify(jnp.cos)(x))
NotImplementedError: sparse rule for cos is not implemented because it would result in dense output.
If this is your intent, use sparse.todense() to convert your arguments to dense matrices.

>>> out = sparse.map_specified_elements(jnp.cos, x)
>>> print(out.todense())
[ 0.          0.5403023  -0.41614684 -0.9899925 ]
```

### DIFF
--- a/jax/experimental/sparse/api.py
+++ b/jax/experimental/sparse/api.py
@@ -31,13 +31,14 @@ Further down are some examples of potential high-level wrappers for sparse objec
 """
 from functools import partial
 import operator
-from typing import Optional, Union
+from typing import Callable, Optional, Union
+import numpy as np
 
 import jax
 from jax import core
 from jax import tree_util
 from jax.experimental.sparse._base import JAXSparse
-from jax.experimental.sparse.bcoo import BCOO
+from jax.experimental.sparse.bcoo import BCOO, _bcoo_extract
 from jax.experimental.sparse.bcsr import BCSR
 from jax.experimental.sparse.coo import COO
 from jax.experimental.sparse.csr import CSR, CSC
@@ -46,7 +47,9 @@ from jax.interpreters import ad
 from jax.interpreters import batching
 from jax.interpreters import mlir
 from jax._src import dtypes
-from jax._src.typing import Array, DTypeLike, Shape
+from jax._src.typing import Array, ArrayLike, DTypeLike, Shape
+from jax._src import util
+from jax._src.lax.control_flow import _initial_style_jaxpr
 
 
 #----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


soraros:
Just an idea, not sure how difficult/useful it would be, but is it possible to have an api looks more like `a.at[:].apply(f)` which could even be introduced for regular arrays.

jakevdp:
To me `a.at[:].apply(f)` implies that `f` will be applied to all elements of the array. What we really want is something like `a.at[a != 0].set(f(a[a != 0])`; I discussed this a bit in https://github.com/google/jax/issues/8967#issuecomment-996152905

jakevdp:
Test failures seem to be due to a PyPI outage: https://status.python.org/

jax/experimental/sparse/api.py
```diff
@@ -107,6 +108,95 @@ def _todense_batching_rule(batched_args, batch_dims, *, tree):
...
+#----------------------------------------------------------------------
+# map_specified_elements – function to map a scalar function to specified elements
+map_specified_elements_p = core.Primitive('map_specified_elements')
+map_specified_elements_p.multiple_results = True
+
+def map_specified_elements(fun: Callable[[ArrayLike], ArrayLike], arg: Union[ArrayLike, JAXSparse]) -> Union[ArrayLike, JAXSparse]:
+  """Map a given function across the specified elements of an array
+
+  If the input is a dense array, ``fun`` will be applied to every array element.
+  If the input is a sparse array, ``fun`` will be applied only to the data buffer.
```
froystig, line 120: > [...] it must also accept dense inputs so that it can be used within a function that is transformed by sparsify

I understand that we want to express this operation in "dense mode", but do we actually want it to execute and have semantics other than "error" in that case? As the comment here notes, the meaning is (possibly subtly) different for dense vs sparse operands.

jakevdp: My thought on this was that for dense inputs, every element is specified! So this lets you write one function that applies to both sparse and dense inputs (either with direct sparse inputs or implicitly via `sparsify`)

froystig: In this case, `sparsify(f)` may no longer be a semantic identity, for example if:
```python
f = partial(sparse.map_specified_elements(jnp.cos)
```
Right? It still won't be a semantic identity if it errs vs. runs too, but I'm offering that that might be better than silently-not-being-the-identity.

froystig: Note you can still `sparsify` a function that calls `map_specified_elements` and have `map_specified_elements` error on execution. It's a matter of a primitive having a valid sparsification rule while, say, its impl/lowering rule is an error.

PR #9000, opened 2021-12-16
Add NaN checkify rule to all lax primitives.

LenaMartens:
And when I say "all", I mean all primitives which also have rules in `experimental/jet.py` (I copied over that set), so maybe incomplete.

I think there are some primitives here which can never generate a NaN (if by generate we mean: produce a NaN when there's no NaN in its input args). eg. boolean functions like `not` and value-polymorphic functions like `broadcast`? Maybe I'll prune those out.


mattjj:
#9000 !!!!

mattjj (approved):
LGTM though from our convo maybe only add nan checks to float-output primitives.

(Also from our convo: wouldn't it be sick if we had a list of all primitives, maybe together with their types, so we could have a loop that says "add this check to all float-output primitives.")

LenaMartens:
nice, I've removed the primitives which seemed to obviously return booleans. Seems like there is already some mechanism for lax primitives to register their output type (eg. https://github.com/google/jax/blob/18baa6e93bbb2b6b45fcb4de4509e092ea970d76/jax/_src/lax/lax.py#L2142), so might be easy to construct a set of primitives which _could_ return floats?

Issue #8757, opened 2021-12-01
Bug in odeint for complex-valued functions


FlorianH-1QBit:
This is happening on the `main` branch and in 0.2.25, not in some older versions. 
I think it was introduced in https://github.com/google/jax/commit/de777c8d37fed5504e5b161a40f4fbdc3bb31026

Code to reproduce the issue: Modified from `tests/ode_test.py` by adding ` * jnp.exp(-t)` in line 11.
```python
import jax
import jax.numpy as jnp
import jax.test_util as jtu
from jax.experimental.ode import odeint


def test_complex_odeint():
    print(jax.__version__)

    def dy_dt(y, t, alpha):
        return alpha * y * jnp.exp(-t)

    def f(y0, ts, alpha):
        return odeint(dy_dt, y0, ts, alpha).real

    alpha = 3 + 4j
    y0 = 1 + 2j
    ts = jnp.linspace(0., 1., 11)
    tol = 1e-1

    jtu.check_grads(f, (y0, ts, alpha), modes=["rev"], order=2, atol=tol, rtol=tol)


test_complex_odeint()
```
The error that's thrown is `TypeError: mul requires arguments to have the same dtypes, got float32, complex64.`, full trace below.

It looks like in https://github.com/google/jax/blob/main/jax/experimental/ode.py#L95, the dtype of `alpha` is `complex64`, which leads to a complex time `ti` in `https://github.com/google/jax/blob/main/jax/experimental/ode.py#L112`.
So `func_` in https://github.com/google/jax/blob/main/jax/experimental/ode.py#L184 gets called sometimes with a time of dtype `float32`, and sometime `complex64`, which seems to cause this problem.

The fix should be to change the dtype of `alpha` in https://github.com/google/jax/blob/main/jax/experimental/ode.py#L95 from `f0.dtype` to `t0.dtype`.

I'm happy to put together a PR!

Here is the stack trace:
```
0.2.26
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
Traceback (most recent call last):
  File "/Users/florianhopfmueller/Library/Application Support/JetBrains/PyCharmCE2021.2/scratches/debug_complex_odeint.py", line 24, in <module>
    test_complex_odeint()
  File "/Users/florianhopfmueller/Library/Application Support/JetBrains/PyCharmCE2021.2/scratches/debug_complex_odeint.py", line 21, in test_complex_odeint
[...]

PR #9020, opened 2021-12-20
Fix a type issue in odeint

FlorianH-1QBit:
This PR fixes the issue described in https://github.com/google/jax/issues/8757, and modifies a test to catch it.

### DIFF
--- a/jax/experimental/ode.py
+++ b/jax/experimental/ode.py
@@ -92,7 +92,7 @@ def initial_step_size(fun, t0, y0, order, rtol, atol, f0):
 
 def runge_kutta_step(func, y0, f0, t0, dt):
   # Dopri5 Butcher tableaux
-  alpha = jnp.array([1 / 5, 3 / 10, 4 / 5, 8 / 9, 1., 1., 0], dtype=f0.dtype)
+  alpha = jnp.array([1 / 5, 3 / 10, 4 / 5, 8 / 9, 1., 1., 0], dtype=dt.dtype)
   beta = jnp.array(
       [[1 / 5, 0, 0, 0, 0, 0, 0], [3 / 40, 9 / 40, 0, 0, 0, 0, 0],
        [44 / 45, -56 / 15, 32 / 9, 0, 0, 0, 0],
@@ -165,9 +165,10 @@ def odeint(func, y0, t, *args, rtol=1.4e-8, atol=1.4e-8, mxstep=jnp.inf):
   """
   for arg in tree_leaves(args):
     if not isinstance(arg, core.Tracer) and not core.valid_jaxtype(arg):
-      msg = ("The contents of odeint *args must be arrays or scalars, but got "
-             "\n{}.")
-      raise TypeError(msg.format(arg))
+      raise TypeError(
+        f"The contents of odeint *args must be arrays or scalars, but got {arg}.")
+  if not jnp.issubdtype(t.dtype, jnp.floating):
+    raise TypeError(f"t must be an array of floats, but got {t}.")
 
   converted, consts = custom_derivatives.closure_convert(func, y0, t[0], *args)
   return _odeint_wrapper(converted, rtol, atol, mxstep, y0, t, *args, *consts)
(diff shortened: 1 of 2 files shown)


jax/experimental/ode.py
```diff
@@ -92,7 +92,7 @@ def initial_step_size(fun, t0, y0, order, rtol, atol, f0):
 
 def runge_kutta_step(func, y0, f0, t0, dt):
   # Dopri5 Butcher tableaux
-  alpha = jnp.array([1 / 5, 3 / 10, 4 / 5, 8 / 9, 1., 1., 0], dtype=f0.dtype)
+  alpha = jnp.array([1 / 5, 3 / 10, 4 / 5, 8 / 9, 1., 1., 0], dtype=t0.dtype)
```
FlorianH-1QBit, line 95: `alpha` is used only in l112, where it is combined with the two times `t0` and `dt`. It should have the same dtype as the times, rather than the return value `f0`.

jakevdp:
Thanks for the fix - once the CLA is resolved, I'll take a look.

FlorianH-1QBit:
Sounds good - my company signed the CLA, I believe it's waiting for approval from Google's end now. I'll let you know once we receive that so you can re-run the workflow!

FlorianH-1QBit:
Hi @jakevdp , happy new year! Could you please re-run the CLA workflow? Thanks.

jakevdp:
I re-ran it; the CLA check is still failing. You'll need to follow the instructions at https://cla.developers.google.com/ to register the email address you use to sign your commits.

FlorianH-1QBit:
Sorry about that, something went wrong on our end. I'll ping you again when it's fixed.

jax/experimental/ode.py
jakevdp: Good catch – though what happens if someone inadvertently passes in an integer sequence of times? Should we promote `t0` and `dt` to inexact somewhere?

jakevdp:
Thanks! Looks like the CLA check passes now.

jax/experimental/ode.py
FlorianH-1QBit: Good point - if an integer `t0` ever makes it here, `alpha` would be wrong if using `t0.dtype`!
Safer would be `f0.real.dtype` or `_complex_elem_type(f0)` from `jax._src.numpy.lax_numpy`?

Right now, `odeint(func=lambda x, t: x, y0=jnp.array(1.), t=jnp.array([0, 1]))` fails elsewhere, so if we want to make that run then `t` should probably be promoted further outside - in `_odeint` maybe? I don't know enough about jax types to do that unfortunately.

jakevdp: Makes sense – I think the easiest fix then would be to (somewhere) add `assert np.issubdtype(t, np.floating)`. The `odeint` docstring says that `t` is an array of floats, so there's probably not any good reason to handle other cases.

jax/experimental/ode.py
```diff
@@ -168,6 +168,9 @@ def odeint(func, y0, t, *args, rtol=1.4e-8, atol=1.4e-8, mxstep=jnp.inf):
       msg = ("The contents of odeint *args must be arrays or scalars, but got "
              "\n{}.")
       raise TypeError(msg.format(arg))
+  if not jnp.issubdtype(t.dtype, jnp.floating):
+    msg = "t must be an array of floats, but got \n{}."
+    raise TypeError(msg.format(t))
```
jakevdp, line 173:
```suggestion
    raise TypeError(f"t must be an array of floats, but got {t}.")
```

jakevdp:
I think the errors are ones that have been fixed on the main brach. Can you rebase?

jakevdp:
Also, we should squash all changes into a single commit before merging (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

FlorianH-1QBit:
> I think the errors are ones that have been fixed on the main brach. Can you rebase?

Do you mean that the main branch uses f-strings for the `TypeError` messages? 
Right now the `*args` TypeError message also uses `.format` so I'll change that as well.

jakevdp:
No, I meant the failed documentation build in the github CI. If you rebase on main, those errors should go away.

jakevdp:
The `.format()` calls are leftover from older versions of Python. We generally prefer f strings now.

jakevdp (approved):
Looks great!

PR #9029, opened 2021-12-21
Add `where=` arg to jax.nn.{softmax, log_softmax, normalize}.

rjagerman:
This change adds a `where=` argument (analogous to `jnp.sum`, `jnp.mean`, etc) to `jax.nn.{softmax, log_softmax, normalize}` that can be used to specify which elements to include in the calculation. This is useful for situations with padding where certain elements must be excluded when calling the `softmax`, `log_softmax` or `normalize` functions.

### DIFF
--- a/jax/_src/nn/functions.py
+++ b/jax/_src/nn/functions.py
@@ -261,7 +261,10 @@ def glu(x: Array, axis: int = -1) -> Array:
 logsumexp = _logsumexp
 
 
-def log_softmax(x: Array, axis: Optional[Union[int, Tuple[int, ...]]] = -1) -> Array:
+def log_softmax(x: Array,
+                axis: Optional[Union[int, Tuple[int, ...]]] = -1,
+                where: Optional[Array] = None,
+                initial: Optional[Array] = None) -> Array:
   r"""Log-Softmax function.
 
   Computes the logarithm of the :code:`softmax` function, which rescales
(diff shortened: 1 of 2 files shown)


jax/_src/nn/functions.py
```diff
@@ -275,11 +276,16 @@ def log_softmax(x: Array, axis: Optional[Union[int, Tuple[int, ...]]] = -1) -> A
     x : input array
     axis: the axis or axes along which the :code:`log_softmax` should be
       computed. Either an integer or a tuple of integers.
+    where: Elements to include in the log_softmax.
```
jakevdp, line 279: `jax.numpy.max` requires `initial` to be explicitly passed when `where` is used – I think it would make sense to do the same for `log_softmax` and `softmax`. For `normalize`, it's probably not necessary because it has a natural identity.

jakevdp (changes requested):

jax/_src/nn/functions.py
rjagerman: Sounds good. I added `initial` as an argument for `softmax` and `log_softmax`.

jakevdp (approved):
Looks good!

tests/nn_test.py
```diff
@@ -123,6 +123,28 @@ def testHardTanhMemory(self):
...
+
+    self.assertAllClose(out_masked, out_filtered)
+
+  def testNormalizeWhereMask(self):
+    x = jnp.array([5.5, 1.3, -4.2, 0.9])
+    m = jnp.array([True, False, True, True])
+    x_filtered = jnp.take(x, jnp.array([0, 2, 3]))
+
+    out_masked = jnp.take(nn.normalize(x, where=m), jnp.array([0, 2, 3]))
+    out_filtered = fn(x_filtered)
```
jakevdp, line 144: Looks like `fn` is undefined

rjagerman: Ah, missed that one. Should be fixed now.

PR #9033, opened 2021-12-21
Fix straight-through estimator example in docs (#9032)

### LINKED ISSUE #9032: Straight-through estimator example in docs is incorrect
In the [Advanced Autodiff docs](https://jax.readthedocs.io/en/latest/jax-101/04-advanced-autodiff.html), there's a numerically incorrect example of a straight-through estimator (STE) implemented with `jax.lax.stop_gradient`, excerpted below.

```
def f(x):
  return jnp.round(x)  # non-differentiable

def straight_through_f(x):
  return x + jax.lax.stop_gradient(f(x) - x)
```

In particular, an STE implementation should have the property that functionally `f == straight_through_f` but this is not the case due to floating point non-associativity above. At the end of this issue I describe this in detail.

Given that this function is only in documentation this might seem a bit pedantic, but I've seen people copy this code now for their own STE implementations, in places where numerical exactness matters (such as quantization).

I have a PR prepared with the quick fix.

## Details on why the given implementation is incorrect

```
import jax
import jax.numpy as jnp

def pass_through_incorrect(x: jnp.ndarray, fn) -> jnp.ndarray:
  # computes (x + (fn(x) - x))
  return x + jax.lax.stop_gradient(fn(x) - x)

def pass_through(x: jnp.ndarray, fn) -> jnp.ndarray:
  # computes (x - x) + fn(x)
  # By Sterbenz lemma, and ieee754 zero arithmatic, this is exactly fn(x).
  return x - jax.lax.stop_gradient(x) + jax.lax.stop_gradient(fn(x))

# Above two are NOT equivalent, but you need load-bearing tests to prove this.
eps = jnp.finfo(jnp.float32).eps

def evaluate_fn(x, name, true_fn):
  pass_through_fn = eval(name)

  actual = pass_through_fn(x, true_fn)
  expected = true_fn(x)

  print(name)
  # Widening our float32 to float64 for printing is OK, as the latter captures
  # the original precisely.
  print('  actual', str(float(actual)))
  print('  expected', str(float(expected)))
  print('  equal?', actual == expected)
  
  def precise_fn(x):
  return eps

inp = jnp.float32(100)

evaluate_fn(inp, 'pass_through_incorrect', precise_fn)
evaluate_fn(inp, 'pass_through', precise_fn)

# A less trivial example

def precise_fn(x):
  return x * eps + eps

inp = jnp.float32(100)

evaluate_fn(inp, 'pass_through_incorrect', precise_fn)
evaluate_fn(inp, 'pass_through', precise_fn)
```

vlad17:
In the Advanced Autodiff docs, there's a numerically incorrect example of a straight-through estimator (STE) implemented with `jax.lax.stop_gradient`, excerpted below.

```
def f(x):
  return jnp.round(x)  # non-differentiable

def straight_through_f(x):
  return x + jax.lax.stop_gradient(f(x) - x)
```

In particular, an STE implementation should have the property that functionally `f == straight_through_f` but this is not the case due to floating point non-associativity above. At the end of the corresponding issue to this PR I describe this in detail.

Given that this function is only in documentation this might seem a bit pedantic, but I've seen people copy this code now for their own STE implementations, in places where numerical exactness matters (such as quantization).

By rearranging the expression, we can recover the desired equality. See the original Github issue for more details.


jakevdp (changes requested):
Thanks -  this looks good!

A request:could you make the change in the `md` file, and then use `jupytext` to generate the `ipynb` file? That way we wouldn't have all new colab metadata and modified output IDs (the former is just keeping the source neat; the latter is problematic because cell ID is the way that colab does hyperlinks to specific cells)

vlad17:
Done.

jakevdp (approved):
Thanks!

jakevdp:
It looks like there's a jupytext version mismatch - can you update?

vlad17:
Just to check, you'd like me to install jupytertext==1.13.0 and re-run it?

vlad17:
For whatever reason, that isn't possible to install. I just manually removed that line from the diff.

jakevdp:
Thanks!

froystig:
Does having merged this resolve #9032?

froystig:
It would be good to have @mtthss confirm as well, as an author of this tutorial.

vlad17:
> Does having merged this resolve #9032?

Yes

mtthss:
Looks good thanks for the fix!

Issue #7242, opened 2021-07-11
jax.nn.initializers.constant implementation


isaaccorley:
Please:

- [X] Check for duplicate requests.
- [X] Describe your goal, and if possible provide a code snippet with a motivating example.

There doesn't seem to be a `jax.nn.initializers.constant` function. I would like to gauge the interest of adding this in a PR.

I have an example implementation below from another [project](https://github.com/isaaccorley/jax-enhance) of mine which simply initializes some ndarray using `jnp.ones` and then multiplies it by the constant value.

```python
import jax.numpy as jnp

def constant(key, shape, value, dtype=jnp.float32) -> jnp.ndarray:
    value = jnp.asarray(value, dtype)
    return jnp.ones(shape, dtype) * value
```

hawkinsp:
The obvious question is: does this function pay for itself? An alternative might be to simply call `jnp.full`.

isaaccorley:
Thanks for pointing this out. I wasn't aware of jnp.full and that seems to be what I was looking for.

isaaccorley:
I would still recommending adding this jax.nn.initializers. There are jax.nn.initializers.ones and jax.nn.initializers.zeros which simply wrap jnp.ones and jnp.zeros, respectively.

hawkinsp:
Looks fixed!

PR #9052, opened 2021-12-27
Add jax.nn.initializers.constant

jpuigcerver:
Sometimes it's very convenient to initialize parameters to a constant value other than one or zero. This function wraps jax.numpy.full to allow that. (e.g. all VIT-like architectures initialize the bias of the head with the value log(1/num_classes), rather than 0).


jpuigcerver:
Fixes #7242.

jekbradbury (approved):

Issue #9073, opened 2021-12-30
custom batching (vmap)


froystig:
Support custom batching, i.e. the ability to register a custom "vmap rule" for any given function. Example usage would look something like:

```python
from jax import vmap, custom_vmap, numpy as jnp

@custom_vmap
def vector_dot(u, v):
  assert u.ndim == v.ndim == 1
  return u @ v

@vector_dot.def_vmap
def vector_dot_vmap_rule(axis_size, in_batched, u, v):
  u_batched, v_batched = in_batched
  if u_batched:
    assert u.ndim == 2 and u.shape[0] == axis_size
    print('lhs batched')
  if v_batched:
    assert v.ndim == 2 and v.shape[0] == axis_size
    print('rhs batched')
  if u_batched and v_batched:
    out = jnp.sum(u * v, axis=1)
  else:
    out = u @ v if u_batched else v @ u
  return [out], [u_batched or v_batched]

def f(u, v):
  return jnp.exp(vector_dot(u, v))

x = lambda *shape: jnp.ones(shape)
vmap(f, in_axes=(0, None))(x(4, 3), x(3))  # -> lhs batched
vmap(f, in_axes=(1, None))(x(3, 4), x(3))  # -> lhs batched
vmap(f, in_axes=(None, 0))(x(3), x(4, 3))  # -> rhs batched
vmap(f, in_axes=(0, 0))(x(4, 3), x(4, 3))  # -> lhs batched, rhs batched
```

This would enable #7199 among other things.

froystig:
A rough update on where the implementation is at present:

* Batching, forward-mode AD (e.g. `jvp`, `jacfwd`), and compilation are all supported.
* Reverse-mode AD (specifically linearization by partial evaluation, and transposition) are a work in progress.
* The underlying primitive currently stages out the custom-batched function eagerly. We may want to move to a delayed tracing approach.

We're also thinking about whether to recommend the general `custom_vmap` function as a direct user-facing API, or whether instead to encourage more structured uses via functions like `sequential_vmap` or variations thereof.

froystig:
Let's look to fix/support #13283 as part of this as well.

patrick-kidger:
By the way, I notice that the above example doesn't include `axis_name`, which in general needs be passed in as well.

PR #9074, opened 2021-12-31
custom vmap: entry point, batching, and JVP

froystig:
Introduces `custom_vmap`, backed by a corresponding primitive, and implements batching and JVP rules for it.

First step of #9073.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -6460,6 +6460,180 @@ class CustomVmapTest(jtu.JaxTestCase):
         'custom vmap rule output values must be a sequence.*',
         lambda: api.vmap(f)(xs))
 
+  def test_jvp_basic(self):
+    @api.custom_vmap
+    def f(x): return jnp.sin(x)
+
+    @f.def_vmap
+    def rule(axis_size, in_batched, xs):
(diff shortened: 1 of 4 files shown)


mattjj (approved):

tests/api_test.py
```diff
@@ -6314,6 +6314,327 @@ def tp(r, t): return t / r
...
+    ys = api.vmap(f)(xs)
+    self.assertAllClose(ys, jnp.cos(xs))
+
+  def test_nary(self):
+    @api.custom_vmap
+    def f(x, y): return jnp.sin(x) + y ** 2.
+
+    @f.def_vmap
+    def rule(axis_size, in_batched, xs, ys):
+      self.assertEqual(in_batched, [True, True])
```
mattjj, line 6341: Maybe add a case where not all of these are true?

mattjj: NVM this is in the case below.

mattjj:
We should add a test which has a batching rule with a loop, since that models a use case we have in mind. But we can't make_jaxpr yet, as you pointed out, so we should add that later.

PR #9079, opened 2022-01-03
Annotate tree_util

NeilGirdhar:


jax/_src/tree_util.py
```diff
@@ -27,15 +27,19 @@
...
   PyTreeDef = pytree.PyTreeDef
 else:
   PyTreeDef = Any
 
 
-def tree_flatten(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
+def tree_flatten(tree: Any,
+                 is_leaf: Optional[Callable[[Any], bool]] = None) -> Tuple[
+                     Sequence[Array],
+                     PyTreeDef]:
```
mattjj, line 42:
```suggestion
                 is_leaf: Optional[Callable[[Any], bool]] = None
                 ) -> Tuple[Sequence[Array], PyTreeDef]:
```

jax/_src/tree_util.py
```diff
@@ -46,56 +50,59 @@ def tree_flatten(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
...
   Returns:
     The reconstructed pytree, containing the ``leaves`` placed in the structure
     described by ``treedef``.
   """
   return treedef.unflatten(leaves)
 
-def tree_leaves(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
+def tree_leaves(tree: Any,
+                is_leaf: Optional[Callable[[Any], bool]] = None) -> (
+                    Sequence[Array]):
```
mattjj, line 77:
```suggestion
                is_leaf: Optional[Callable[[Any], bool]] = None
                ) -> Sequence[Array]:
```

jax/_src/tree_util.py
```diff
@@ -46,56 +50,59 @@ def tree_flatten(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
       stopped immediately, with the whole subtree being treated as a leaf.
 
   Returns:
-    A pair where the first element is a list of leaf values and the second
+    A pair where the first element is a sequence of leaf values and the second
```
mattjj, line 53: How about we keep it as returning a list? In general I think it makes sense for the input specification to be loose and the output specification to be tight. (We would update the above annotation too.)

jax/_src/tree_util.py
mattjj: As mentioned above, I think we should annotate the return as `List[Array]` instead of `Sequence[Array]`.

jax/_src/tree_util.py
```diff
@@ -46,56 +50,59 @@ def tree_flatten(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
...
   return pytree.flatten(tree, is_leaf)[1]
 
-def treedef_tuple(treedefs):
-  """Makes a tuple treedef from a list of child treedefs."""
+def treedef_tuple(treedefs: Iterable[PyTreeDef]) -> PyTreeDef:
+  """Makes a tuple treedef from an iterable of child treedefs."""
   return pytree.tuple(list(treedefs))
 
-def treedef_children(treedef):
+def treedef_children(treedef: PyTreeDef) -> Sequence[PyTreeDef]:
```
mattjj, line 90:
```suggestion
def treedef_children(treedef: PyTreeDef) -> List[PyTreeDef]:
```

jax/_src/tree_util.py
```diff
@@ -46,56 +50,59 @@ def tree_flatten(tree, is_leaf: Optional[Callable[[Any], bool]] = None):
...
+  """Tests whether all of the elements in the given iterable are leaves.
 
   >>> tree = {"a": [1, 2, 3]}
   >>> assert all_leaves(jax.tree_leaves(tree))
   >>> assert not all_leaves([tree])
 
   This function is useful in advanced cases, for example if a library allows
-  arbitrary map operations on a flat list of leaves it may want to check if
-  the result is still a flat list of leaves.
+  arbitrary map operations on a flat iterable of leaves it may want to check
```
mattjj, line 104: Keep as list rather than iterable?

jax/_src/ad_checkpoint.py
```diff
@@ -224,7 +224,8 @@ def fun_remat(*args, **kwargs):
 ### Utilities
 
 def saved_residuals(f, *args, **kwargs) -> List[Tuple[core.AbstractValue, str]]:
-  args, in_tree = tree_flatten((args, kwargs))
+  args_seq, in_tree = tree_flatten((args, kwargs))
+  args = tuple(args_seq)
```
mattjj, line 228: I'm curious, was this change actually needed? It seems surprising. (It'd be cleaner not to add the extra line!)

mattjj (approved):
Nice! Thanks as always, Neil.

I have suggestions which fall into two main categories:
1. whitespace
2. having more specific output type annotations (lists rather than sequences).

I'm interested if there are any strong opinions against 2, but it seems preferable to be specific about output types (unless we thought they were likely to change).

NeilGirdhar:
> Nice! Thanks as always, Neil.
> 
> I have suggestions which fall into two main categories:
> 
> 1. whitespace

Of course, I'm happy to fix all whitespace to match your style 😄 

> 2. having more specific output type annotations (lists rather than sequences).
> 
> I'm interested if there are any strong opinions against 2, but it seems preferable to be specific about output types (unless we thought they were likely to change).

I don't have any strong opinion, but I should have commented on this in the change.   I initially wrote the annotations using `List`, but it turns out that the underlying `jaxlib` functions are already annotated using `Sequence`.  This caused typing errors in the `tree_util` code.

I feel like `jaxlib` and `tree_util` should match: either `jaxlib` should change its annotations to `List`, or else `tree_util` should use `Sequence` to match.  What do you think?  Do you agree?

Promising sequence means that you could one day choose a different sequence type than list.  For example, you might provide some alternative structure that supports easier tree traversal.  Maybe that was the motivation of whoever annotated `jaxlib`:  keep options open.  If I had to choose, I think promising less is usually better.

Please let me know what you prefer.

jax/_src/ad_checkpoint.py
NeilGirdhar: It was tripping a typing error on the commit hook.  We could get rid of the extra line by keeping
```python
args_seq, in_tree = tree_flatten((args, kwargs))
```
here, but then changing the two uses of it to `*args_seq` and `len(args_seq)` below.  Would that be preferable?  I was trying to be minimally invasive, but you're right that shorter is probably clearer.

NeilGirdhar: The reason that this is a type error by the way is because `args` changes type.  As an argument, it's a tuple, but the result of `tree_flatten` is either a sequence or a list—neither of which is a subclass of tuple.  To be honest, it might be better to give this a whole new name anyway since it doesn't correspond to just `args` anyway.  Maybe `in_leaves` to match `in_tree`?

mattjj: Ah, makes sense! `in_leaves` sgtm.

mattjj:
> I agree with the jaxlib author that promising less is usually better.

I'm not so sure about this in general. Why not promise even less and say it's just an Iterable? We want to promise the right amount, and IMO there's no harm in continuing to say it's a list, as our docs already did. Or at least, there wouldn't be except for the other issue you mentioned.

> but it turns out that the underlying jaxlib functions are already annotated using Sequence. This caused typing errors in the tree_util code.

Ah I see. Let's keep it how your PR does it then! I don't mean to add any extra work.

NeilGirdhar:
> I'm not so sure about this in general. Why not promise even less and say it's just an Iterable? We want to promise the right amount,

Yes, you're absolutely right of course.

> Ah I see. Let's keep it how your PR does it then! I don't mean to add any extra work.

Okay, thanks!  I'll make the change above, squash, and let you know when it's done.

mattjj:
You're so generous and easy to work with, Neil! It's always a pleasure, and I always learn something. You're the best.

NeilGirdhar:
Done, and thank you, you all are a pleasure to work with as well!

mattjj:
Unfortunately, internal typechecks (i.e. typechecks spanning all google code which uses these functions!) ran into some issues with this.

As one example, [jraph](https://github.com/deepmind/jraph) does [addition on the leaves outputs of `tree_flatten` calls](https://github.com/deepmind/jraph/blob/master/jraph/_src/utils.py#L816). But that doesn't typecheck against `Sequence`, since you can't add `Sequence`s together!

Maybe we could put these annotations [under `if TYPE_CHECKING`](https://github.com/google/jax/blob/main/jax/_src/tree_util.py#L32)? Got a better idea?

NeilGirdhar:
> As one example, jraph does addition on the leaves outputs of tree_flatten calls. But that doesn't typecheck against Sequence, since you can't add Sequences together!

That makes sense.  That's why I had to change a couple instances of concatenation in the Jax code to use generalized unpacking.

> Maybe we could put these annotations under if TYPE_CHECKING? Got a better idea?

Unfortunately, that won't solve the type checking errors since those run under TYPE_CHECKING too.  I can see a few options:

- Change tree-util annotations to use lists, and change jaxlib to match,
- Change tree-util annotations to use lists, and add casts throughout so that jaxlib doesn't complain, or
- Change the client code to cast results to list where necessary.

Option 1 means changing Jaxlib (is that hard?)  Option 2 adds extra casts that weren't there before, and is awkward having two mismatched interfaces.  Option 3 depends on how many type errors you ended up with.

Another question goes back to your point above.  What do you want in an ideal world?  Would you rather promise list and people can easily concatenate them or would you rather promise sequence and you're free to use another container down the road?

There may be other options, but those are the obvious ones to me.  Please let me know what you decide and I'll repair this change.

mattjj:
Thanks for that super clear analysis.

I think Option 3 would be pretty annoying.

I like Option 1 but I wonder if @hawkinsp has a different opinion. Peter, WDYT?

NeilGirdhar:
@hawkinsp If you like, I can make a corresponding pull request to tensorflow's jaxlib to make the interface promise lists?

hawkinsp:
Sure, that's fine. Just tag me in the PR so I see it.

NeilGirdhar:
This should pass when Tensorflow 2.10's version of jaxlib is used (whenever that is)

NeilGirdhar:
@mattjj Would it be possible to get this merged now that the tensorflow type annotations are in jaxlib?

jakevdp (approved):
Looks great, thanks!

jakevdp:
(taking over review/merge, since this is related to #12049)

PR #9089, opened 2022-01-04
Fix test failures under Numpy 1.22.

hawkinsp:


jax/_src/numpy/lax_numpy.py
```diff
@@ -2910,11 +2910,14 @@ def nanmax(a, axis: Optional[Union[int, Tuple[int, ...]]] = None, out=None,
...
 @partial(jit, static_argnames=('axis', 'dtype', 'keepdims'))
 def nansum(a, axis: Optional[Union[int, Tuple[int, ...]]] = None, dtype=None,
-           out=None, keepdims=None):
+           out=None, keepdims=None, initial=None, where=None):
   lax._check_user_dtype_supported(dtype, "nanprod")
   return _nan_reduction(a, 'nansum', sum, 0, nan_if_all_nan=False,
                         axis=axis, dtype=dtype, out=out, keepdims=keepdims)
 
+# Work around a sphinx documentation warning in NumPy 1.22.
+nansum.__doc__ = nansum.__doc__.replace("\n\n\n", "\n\n")
```
jakevdp, line 2919: Maybe better to do this as part of `_wraps`?

jax/_src/numpy/lax_numpy.py
```diff
@@ -6322,28 +6325,29 @@ def corrcoef(x, y=None, rowvar=True):
 
 @_wraps(np.quantile, skip_params=['out', 'overwrite_input'])
 @partial(jit, static_argnames=('axis', 'overwrite_input', 'interpolation',
-                               'keepdims'))
+                               'keepdims', 'method'))
 def quantile(a, q, axis: Optional[Union[int, Tuple[int, ...]]] = None, out=None,
-             overwrite_input=False, interpolation="linear", keepdims=False):
+             overwrite_input=False, method="linear", keepdims=False,
+             interpolation=None):
```
jakevdp, line 6331: Should we raise a deprecation warning if `interpolation` is not `None`?

jax/_src/numpy/lax_numpy.py
hawkinsp: I think it's a documentation bug in this particular docstring. The right fix is probably to send a PR upstream to remove the extra newline..

jax/_src/numpy/lax_numpy.py
hawkinsp: Yes, but I'm leaving that for a future PR: this is just to unbreak the CI.

hawkinsp: Changed my mind and did it.

jakevdp (approved):

Issue #8978, opened 2021-12-16
scipy.stats.mode


gianlucadetommaso:
It would be great if you could add the JAX equivalent of `scipy.stats.mode`, which is currently unavailable. 

A use case may be a ML classification task with ensembles, where multiple models give different predictions, and we are interested in finding the most common one.

As an example, consider `predictions` to be a two-dimensional DeviceArray of predictions, with 
shape = (number of models, number of data points).

It would be nice if one could compute

`mode_prediction  = jax.scipy.stats.mode(predictions, axis=0)`,

returning the most common prediction for every data point.

jakevdp:
Here's a quick jit-compatible implementation if you want to make use of it:
```python
import jax.numpy as jnp
from jax import vmap

def mode(x, axis=None):
  def _mode(x):
    vals, counts = jnp.unique(x, return_counts=True, size=x.size)
    return vals[jnp.argmax(counts)]
  if axis is None:
    return _mode(x)
  else:
    x = jnp.moveaxis(x, axis, 0)
    return vmap(_mode, in_axes=(1,))(x.reshape(x.shape[0], -1)).reshape(x.shape[1:])

x = jnp.array([1, 1, 2, 2, 2, 3, 4])
print(mode(x))
# 2

y = jnp.array([[1, 1, 2, 3],
               [4, 5, 6, 5]])
print(mode(y, 1))
# [1, 5]
```

gianlucadetommaso:
Great, thanks!

jakevdp:
It's definitely in scope to add this to `jax.scipy.stats`. If anyone wants to tackle it, you could take the example implementation from above, make sure it returns the same values as `scipy.stats`, and add an appropriate test in `tests/scipy_stats_test.py`

jayendra13:
Let me take this up, can anyone suggest a good location/file for this function to put ?
For Scipy it's defined [here](https://github.com/scipy/scipy/blob/v1.8.0rc2/scipy/stats/_mstats_basic.py#L285) and [here](https://github.com/scipy/scipy/blob/v1.8.0rc2/scipy/stats/_stats_py.py#L365).

I am starting this for now by putting it inside the [__init__.py](https://github.com/google/jax/blob/main/jax/scipy/stats/__init__.py)

Yash621:
@gianlucadetommaso  can i work on this ?

gianlucadetommaso:
@Yash621 please go ahead, I haven't had much time to take this up myself.

[...]

PR #9091, opened 2022-01-04
Implemented mode function

Yash621:
this PR is respect to #8978 , @gianlucadetommaso  please review it.

I have implemented the mode function ,and have also created the _core function and have placed the function there.

### DIFF
--- a/jax/_src/scipy/stats/_core.py
+++ b/jax/_src/scipy/stats/_core.py
@@ -1,3 +1,17 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import jax.numpy as jnp
 from jax import vmap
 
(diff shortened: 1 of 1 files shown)


jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,16 @@
+import jax.numpy as jnp
```
jakevdp, line 1: We'll need a license here; see the other files in the directory

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,16 @@
...
+    Return an array of the modal (most common) value in the passed array.    
+  """
+  def _mode(x):
+    vals, counts = jnp.unique(x, return_counts=True, size=x.size)
+    return vals[jnp.argmax(counts)]
+  if axis is None:
+    return _mode(x)
+  else:
+    x = jnp.moveaxis(x, axis, 0)
+    return vmap(_mode, in_axes=(1,))(x.reshape(x.shape[0], -1)).reshape(x.shape[1:])
```
jakevdp, line 16: I think `scipy.stats.mode` returns a tuple containing the counts; we'll want this function to do the same

jakevdp (changes requested):
Thanks for this! It's a good start, but we have quite a ways to go before we can merge this. A few things:

- we should make this function match the semantics of `scipy.stats.mode`, in particular it should return counts as well as the mode value
- we need to import this function in the appropriate place (`jax/scipy/stats/__init__.py`) in order for it to be in the public namespace.
- we will need thorough unit tests to ensure the function is returning the correct results. This could go in `tests/scipy_stats_test.py` and be modeled after something like this test: https://github.com/google/jax/blob/04f322e065a2cdbb988bf73f26b6ae4968bf1ea5/tests/lax_numpy_test.py#L4570

Let me know if you have any questions!

PR #9093, opened 2022-01-04
Checkify: add division by zero check.

LenaMartens:
Inspired by `np.seterr` having an option to raise an error on division by zero (https://numpy.org/doc/stable/reference/generated/numpy.seterr.html)


mattjj (approved):
So sick.

PR #9094, opened 2022-01-04
lax.dynamic_update_slice: fix batching rule

### LINKED ISSUE #9083: dynamic_update_slice start index adjustment fails when vmapped
The documentation for `dynamic_update_slice` explains "If the update slice is too large to fit in the array, the start index will be adjusted to make it fit" and this behavior does indeed seem to work normally but fails when I apply vmap. No update occurs if the update slice would not fit even if the update slice would still be partially in the array as in the example below.

```python
zeros = jnp.zeros(5)
ones = jnp.ones(3)

for i in range(1, 4):
    print(jax.lax.dynamic_update_slice(zeros, ones, (jnp.array(i, dtype=int),)))

print(jax.vmap(jax.lax.dynamic_update_slice, (None, None, 0))(zeros, ones, (jnp.arange(1, 4),)))

# [0. 1. 1. 1. 0.]
# [0. 0. 1. 1. 1.]
# [0. 0. 1. 1. 1.]
# [[0. 1. 1. 1. 0.]
#  [0. 0. 1. 1. 1.]
#  [0. 0. 0. 0. 0.]]
```

jakevdp:
Fixes #9083

### DIFF
--- a/jax/_src/lax/slicing.py
+++ b/jax/_src/lax/slicing.py
@@ -933,7 +933,7 @@ def _dynamic_update_slice_batching_rule(batched_args, batch_dims):
     scatter, (operand, index, update), (operand_bd, index_bdim, update_bd),
     update_jaxpr=None, update_consts=None, dimension_numbers=dnums,
     indices_are_sorted=True, unique_indices=True,
-    mode=GatherScatterMode.PROMISE_IN_BOUNDS)
+    mode=GatherScatterMode.CLIP)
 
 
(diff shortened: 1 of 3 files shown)


hawkinsp (approved):
This seems right to me. I think I misremembered that dynamic_update_slice had the same semantics as scatter ("drop"), but it doesn't (https://www.tensorflow.org/xla/operation_semantics#dynamicupdateslice). It's clearly documented to have "clip" behavior.

Issue #9107, opened 2022-01-05
`jnp.searchsorted` returns different values to `np.searchsorted`


patrick-kidger:
```python
import numpy as np
import jax.numpy as jnp

x = np.arange(10, dtype=float)
x[4:] = jnp.nan
np.searchsorted(x, 0)  # 0
jnp.searchsorted(x, 0)  # DeviceArray(10, dtype=int32)
```
It doesn't look like `jnp.searchsorted` handles NaNs correctly.

jakevdp:
I think this is expected: for `searchsorted` to work correctly, the array must be sorted in increasing order. But for your input I get this:
```python
>>> x[:-1] < x[1:]
array([ True,  True,  True, False, False, False, False, False, False])
```
which implies the array is not in a sorted order, which means the output is ill-defined. We have not made any effort to ensure that JAX's undefined outputs match numpy's undefined outputs, as it would be quite difficult to do so (side note: even for `sort()`, JAX treats NaNs differently than NumPy; see #6443)

What do you think? Is this something you can work around?

jakevdp:
Thinking about it... we could consider this defined behavior in that `sort()` is defined for `NaN` (and `inf`) entries. We could require that the index returned is that that would leave the array unchanged under JAX's definition of `sort()`, probably by first calling `_float_to_int_for_sort` on the entries, as is done when `sort()` is called on a float array: https://github.com/google/jax/blob/f96761cdc861ed6ecaa3bd1b84ca88b257d1b601/jax/_src/lax/lax.py#L3649

That still would not exactly match NumPy's behavior because of the handling of negative NaNs (cf. #6443) but it would definitely be more sensible than the current behavior. What do you think?

jakevdp:
(side-note: this is quite timely becaue I've been working on `searchsorted` all this afternoon: #9108)

patrick-kidger:
I think your proposal to handle NaNs as per `jnp.sort` sounds like a good one.

This is the choice NumPy has made as well: ["As of NumPy 1.4.0 searchsorted works with real/complex arrays containing nan values. The enhanced sort order is documented in sort."](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html) (And it would certainly handle my use case.)

The point about negative NaNs is certainly an oddity - since the concept of a negative NaN isn't technically defined - but agreed that this is still more sensible than the current behaviour.

jakevdp:
Complex version is fixed as part of #9108

PR #9108, opened 2022-01-05
Add lax.searchsorted_p primitive

jakevdp:
Make `searchsorted` a lax primitive, primarily to allow switching to a more efficient implementation on accelerators.

~Also addresses one of the TODOs left over from #9107 (correct results for complex inputs in the presence of NaNs)~

Still a few TODOs/things to think about:

- ~I previously ran some benchmarks on early versions of these implementations that convinced me this was worth doing; I plan to run some benchmarks on top of this change to better convince us that the backend-specific default implementations are suitable.~ Benchmarks here: https://colab.research.google.com/drive/1kt9dwAFWZIkClOrrKr3zZLmAyWBroxj3?usp=sharing
- ~The form of the `batch_dims` argument (a tuple of lists of batch indices) is the most general approach I could think of, but it may be simpler to specify `batch_dims` as an integer and require then to be at the front of both arrays. The downside is this would in general require transposes within the batch rule.~ Done: I like the simplified version better.
- ~Should the primitive be generalized further to take `N` sorted arrays and `N` query arrays? I can think of some applications, e.g. sparse-sparse matmul implementations.~ No, we'll punt on this for now. Variadic searchsorted can be added in a fully backward-compatible way, if desired.

~This needs to be rebased against #9127 once that is merged.~

### DIFF
--- a/jax/_src/lax/search.py
+++ b/jax/_src/lax/search.py
@@ -122,14 +122,15 @@ def _searchsorted_scan_unbatched(sorted_arr, query, *, side, dtype):
 def _searchsorted_sort_unbatched(sorted_arr, query, *, side, dtype):
   assert sorted_arr.ndim == 1
   assert side in ['left', 'right']
+  working_dtype = np.int32 if sorted_arr.size + query.size < np.iinfo(np.int32).max else np.int64
   def _rank(x):
-    idx = lax.iota(dtype, len(x))
+    idx = lax.iota(working_dtype, len(x))
     return lax._zeros(idx).at[lax.sort_key_val(x, idx)[1]].set(idx)
   if side == 'left':
     index = _rank(lax.concatenate([query.ravel(), sorted_arr], 0))[:query.size]
   else:
     index = _rank(lax.concatenate([sorted_arr, query.ravel()], 0))[sorted_arr.size:]
-  return lax.reshape(lax.sub(index, _rank(query.ravel())), np.shape(query))
+  return lax.reshape(lax.sub(index, _rank(query.ravel())), np.shape(query)).astype(dtype)
 
 def _searchsorted_batch_rule(batched_args, bdims, *, side, dimension, batch_dims, method):
   sorted_arr, query = batched_args
--- a/jax/_src/lax/search.py
+++ b/jax/_src/lax/search.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from functools import partial
+import operator
 
 import numpy as np
 
@@ -20,7 +21,6 @@ import jax
 from jax._src import ad_util
 from jax._src import api
 from jax._src import util
-from jax._src.api_util import _ensure_index_tuple
 from jax._src.lax import lax
 from jax._src.lax.control_flow import fori_loop
 from jax import core
@@ -29,7 +29,7 @@ from jax.interpreters import batching
 from jax.interpreters import xla
 
 
-def searchsorted(sorted_arr, query, *, side='left', dimension=0, batch_dims=None, method="default"):
+def searchsorted(sorted_arr, query, *, side='left', dimension=0, batch_dims=0, method="default"):
   """Find indices of query values within a sorted array.
 
   Args:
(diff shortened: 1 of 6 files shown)


jax/_src/lax/search.py
```diff
@@ -0,0 +1,179 @@
...
+  assert sorted_arr.ndim == 1
+  assert side in ['left', 'right']
+  def _rank(x):
+    idx = lax.iota(dtype, len(x))
+    return lax._zeros(idx).at[lax.sort_key_val(x, idx)[1]].set(idx)
+  if side == 'left':
+    index = _rank(lax.concatenate([query.ravel(), sorted_arr], 0))[:query.size]
+  else:
+    index = _rank(lax.concatenate([sorted_arr, query.ravel()], 0))[sorted_arr.size:]
+  return lax.reshape(lax.sub(index, _rank(index)), np.shape(query))
```
soraros, line 132: Will changing `_rank(index)` into `_rank(query)` be helpful, as then `index` and `_rank(query)` can be scheduled in parallel (with data dependency removed)?

jax/_src/lax/search.py
```diff
@@ -0,0 +1,179 @@
...
+    raise ValueError(f"invalid argument side={side!r}, expected 'left' or 'right'")
+  if not batch_dims <= dimension < sorted_arr.ndim:
+    raise ValueError(f"dimension={dimension} must be in range [{batch_dims}, {sorted_arr.ndim})")
+  if sorted_arr.shape[:batch_dims] != query.shape[:batch_dims]:
+    raise ValueError(f"batch dimension sizes must match; got {sorted_arr.shape[:batch_dims]} != "
+                     f"{query.shape[:batch_dims]}")
+  shape = (*sorted_arr.shape[:batch_dims],
+           *(s for d, s in enumerate(sorted_arr.shape) if d >= batch_dims and d != dimension),
+           *(s for d, s in enumerate(query.shape) if d >= batch_dims))
+  dtype = np.int32 if sorted_arr.shape[dimension] < np.iinfo(np.int32).max else np.int64
```
soraros, line 80: Considering the sort based approach, it should be the length of the concatenated array that is smaller than `1 << 31` right?

jax/_src/lax/search.py
jakevdp: Good call! Done.

jakevdp:
I ran some benchmarks on the final implementation: https://colab.research.google.com/drive/1kt9dwAFWZIkClOrrKr3zZLmAyWBroxj3?usp=sharing

To be honest, they're not totally convincing: The `scan` approach is clearly optimal on CPU, but neither implementation seems to have a clear edge on GPU or TPU.

soraros:
I guess you also need to benchmark larger for query size. It is true a direct call to `searchsorted` tend to have smaller query size, but for instance `vmap(jnp.digitize, in_axes=(0, None))` can generate a call with large query size.

jakevdp:
How large is "large" to you?

soraros:
> How large is "large" to you?

Indeed, that depends. Let's just say I run your benchmark code with `len(query) == 10000` and the result doesn't look good for sort.

jakevdp:
Thanks - I updated the notebook. Looks like TPU clearly benefits from the sort method as the query grows, but GPU does not.

jax/_src/lax/search.py
jakevdp: Good catch - done.

jakevdp:
I suspect performance tradeoffs here will likely depend on the size of the compute registers on the chip, which makes me think that achieving optimal performance will probably involve pushing the algorithmic choice here down to the XLA or MLIR compiler backend. That said, given the overall trend in these benchmarks we can probably default to `sort` on TPU and to `scan` on CPU/GPU. What do you think?

jax/_src/lax/search.py
```diff
@@ -0,0 +1,180 @@
...
+      and `query` to treat as batch dimensions.
+    method : {'default', 'scan', 'sort'} The method used to compute the result.
+      Outside JIT, this defaults to "scan". Within JIT, this defaults to "scan"
+      on CPU, and "sort" otherwise. Assume ``M = query.size`` and
+      ``N = sorted_arr.shape[dimension]``. ``method='scan'`` uses an *O[M log N]*
+      binary search algorithm, while ``method='sort'`` uses an *O[(M+N) log(M+N)]*
+      co-sorting approach. Although ``method='scan'`` may appear to have better
+      scaling overall, in practice it cannot take advantage of parallelism inherent
+      to accelerators, and so in most cases is slower than ``method='sort'`` on GPU
+      and TPU backends.
```
shoyer, line 52: At what point does it make sense to switch to parallel search (`(a < v).sum()`, vmapped over `v` is necessary) rather than bothering with a sort?

I don't know if this needs to be built in to searchsorted, but I'm sure this is faster on accelerators, at least for low dimensional inputs.

jakevdp: Good question - I think that's hard to answer in general (different hardware will have different behavior) which is why this is difficult to do well without pushing the logic into the compiler.

shoyer: I gave this a try. To my surprise, in my tests using this for many vmapped interpolations on TPU, it didn't seem to really help.

jakevdp:
I think I'm going to close this. The sort-based algorithm is now available in  #11576, and it's not clear to me that the choice of algorithm can be made effectively at the level of a translation rule, so the primitive is probably too heavy-handed a solution for now (unless XLA grows a binary search operator, which would be very cool)

PR #9113, opened 2022-01-06
Updated broken links and cell outputs in thinking_in_jax.ipynb to reflect the current JAX behavior

che-shr-cat:
I updated the broken links and cell outputs (including error messages and some related texts) to reflect the current JAX behavior

### DIFF
--- a/docs/notebooks/thinking_in_jax.ipynb
+++ b/docs/notebooks/thinking_in_jax.ipynb
@@ -264,7 +264,7 @@
     "id": "yRYF0YgO3F4H"
    },
    "source": [
-    "For updating individual elements, JAX provides an [indexed update syntax](https://jax.readthedocs.io/en/latest/jax.ops.html#syntactic-sugar-for-indexed-update-operators) that returns an updated copy:"
+    "For updating individual elements, JAX provides an [indexed update syntax](https://jax.readthedocs.io/en/latest/jax.ops.html#indexed-update-operators) that returns an updated copy:"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


che-shr-cat:
I didn't get it, why the check is failed with the message:

[jupytext] Error: the git index is outdated.
Please add the paired notebook with:
    git add docs/notebooks/thinking_in_jax.ipynb

jakevdp:
You can find information on our approach to notebooks at https://jax.readthedocs.io/en/latest/developer.html#update-notebooks; in short, you can `git add` your changes and then run `jupytext --sync` to fix the issue

That said, I'd prefer we avoid the large (mostly meaningless) metadata diffs in the notebook format here; I'd suggest reverting the change to the `ipynb`, and then re-generating it from the markdown with jupytext.

Sorry for the complication here – notebooks and git just don't mix all that well.

jakevdp (changes requested):
Looks good, thanks!

A couple minor remaining things:

- I think the CI failure is something that's already been fixed; if you rebase on the current main, it should work itself out
- Please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests for details)

Thanks for looking at this!

che-shr-cat:
Thanks for the help! Working on this

che-shr-cat:
Read the Docs build is still failing, and I cannot understand what's the issue with it. It looks like two cells cannot be parsed (there are warnings there) but I cannot identify which ones and what's exactly the reason :(

jakevdp:
I suspect the issue is the back-ticks within the traceback messages (e.g. line 238 of the ipynb file)

che-shr-cat:
Still cannot fix it, so I'll close the PR for now to understand better what's wrong

che-shr-cat:
I removed cell output changes from the notebook as I cannot identify what exactly makes tests fail. So I kept only text changes regarding the broken URLs and updated text to better align with the output of current versions of the JAX library.

jakevdp (approved):

jakevdp:
Thanks! Sorry for the confusion about this. For what it's worth, I think the issue is that myst-nb cannot handle error messages that contain back-tick characters, despite the fact that some of JAX's current error messages contain back-ticks. We could either fix this by patching myst-nb, or by modifying JAX's error messages to remove the back-ticks.

PR #9114, opened 2022-01-06
jnp.searchsorted: properly handle NaNs

### LINKED ISSUE #9107: `jnp.searchsorted` returns different values to `np.searchsorted`

jakevdp:
Fixes #9107

Still TODO is to properly handle complex inputs; this will be a bit, *ahem*, complex because `lax.sort` sorts complex numbers lexicographically, so we'd have to modify the binary search to handle lexicographic sorting.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -50,7 +50,7 @@ from jax.config import config
 from jax.interpreters import pxla
 from jax import lax
 from jax._src import device_array
-from jax._src.lax.lax import _array_copy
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Nice!

PR #9118, opened 2022-01-07
Add `jax.default_device` context manager

skye:
This currently only supports setting a specific Device object, not a
platform like "cpu". That should be added in the future.

Bumps the minimum jaxlib version in order to include
https://github.com/tensorflow/tensorflow/pull/53656

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -27,6 +27,7 @@ import warnings
 from jax._src import lib
 from jax._src.lib import jax_jit
 from jax._src.lib import transfer_guard_lib
+from jax._src.lib import xla_client
 
 def bool_env(varname: str, default: bool) -> bool:
   """Read an environment variable and interpret it as a boolean.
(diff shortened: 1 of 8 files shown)


skye:
Sorry this is so huge; I can break up into smaller changes if that'd be helpful. I'll also create a CL if you prefer to review that.

jax/_src/config.py
```diff
@@ -578,6 +616,33 @@ def _update_x64_thread_local(val):
...
+def _update_default_device_thread_local(val):
+  lib.jax_jit.thread_local_state().default_device = val
+
+def _validate_default_device(val):
+  if val is not None and not isinstance(val, xla_client.Device):
+    raise ValueError("jax.default_device must be passed a Device object (e.g. "
+                     f"`jax.devices('cpu')[0]`), got: {repr(val)}")
+
+# TODO(skye): default_device only accepts devices for now. Make it work with
+# platform names as well (e.g. "cpu").
```
mattjj, line 632:
```suggestion
# platform names as well (e.g. "cpu" to mean the same as jax.devices("cpu")[0]).
```

jax/_src/config.py
```diff
@@ -578,6 +616,33 @@ def _update_x64_thread_local(val):
...
+default_device = config.define_string_or_object_state(
+    name='jax_default_device',
+    default=None,
+    help=(
+        'Set to a Device object (e.g. ``jax.devices("cpu")[0]``) to use as the '
```
mattjj, line 637:
```suggestion
        'Configure the default device for JAX operations. '
        'Set to a Device object (e.g. ``jax.devices("cpu")[0]``) to use that Device as the '
```

(I thought it might be useful to start with a short description of what this option is, before saying more about how to set it.)

mattjj (approved):
Nice! This wasn't too long at all.

Left a couple optional comments.

cgarciae:
👀

PR #9121, opened 2022-01-07
add axis names to sharding_constraint_p

mattjj:
This allows annotating vmapped axes:

image (10)

### DIFF
--- a/jax/interpreters/sharded_jit.py
+++ b/jax/interpreters/sharded_jit.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from functools import partial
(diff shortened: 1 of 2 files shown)


pschuh (commented):
Looks good for sharded_jit, but I think you wanted the with_sharding_constraint here: jax/experimental/pjit.py ?

mattjj:
Hah, I didn't realize there were two versions of the function! Indeed I only want the pjit one. I'll basically move the changes over to that one...

mattjj:
Actually, the main intended user of this feature ended up not interested, so I killed the PR!

PR #9123, opened 2022-01-07
Fix host_callback docs

prazek:
There was a missing ':' causing invalid rendering of the docs.


mattjj (approved):

mattjj:
Thank you for the fix!

PR #9137, opened 2022-01-08
simplify custom_jvp_call_p, remove custom_jvp_call_jaxpr_p

mattjj:
Remove custom_jvp_call_jaxpr_p and its transformation rules. They were superfluous! Instead use the new mechanism for converting from jaxpr params to bind params (in #9136).

~This PR currently includes the commit from #9136, but it should be considered as a "diffbase".~

The simplification in JaxprTrace.process_custom_jvp_call was actually made possible by omnistaging #3370, though we didn't apply it until now.

In a follow-up PR we'll delete custom_vjp_call_jaxpr_p too (or die trying). I'd like to land this one fully first to make sure this approach works (and hence ensure doing the same to custom_vjp_call_p makes sense).

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -15,8 +15,8 @@
 from functools import update_wrapper, reduce, partial
 import inspect
 import operator as op
-from typing import (Any, Callable, Generic, List, Optional, Sequence, Set,
-                    Tuple, TypeVar)
+from typing import (Callable, Generic, Optional, Sequence, Tuple, TypeVar, Set,
+                    Any)
 
 from jax import core
 from jax import linear_util as lu
@@ -276,8 +276,6 @@ def _flatten_jvp(in_tree, *args):
   yield primals_out + tangents_out, out_tree
 
 class CustomJVPCallPrimitive(core.CallPrimitive):
-  initial_style: core.Primitive
-
   def bind(self, fun, jvp, *args):
     args = map(core.full_lower, args)
     top_trace = core.find_top_trace(args)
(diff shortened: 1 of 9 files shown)


tests/api_test.py
```diff
@@ -2666,8 +2666,7 @@ def test_escaped_tracer_transform_name(self):
       _ = self._saved_tracer+1
 
   def test_escaped_tracer_shape_dtype(self):
-    with self.assertRaisesRegex(core.UnexpectedTracerError,
-                                r"shape \(4, 3\) and dtype int32"):
+    with self.assertRaisesRegex(core.UnexpectedTracerError, r"int32\[4,3\]"):
```
mattjj, line 2669: I changed this test because I improved the error message. I changed the error message in this PR because previously the error message formatting always attempted to print the shape and dtype associated with a Tracer, but that Tracer could be on a core.unit (which has no shape or dtype).

jax/interpreters/partial_eval.py
```diff
@@ -316,38 +316,11 @@ def partial_eval(self, f: lu.WrappedFun, pvals: Sequence[PartialVal],
...
-
-    eqn = new_eqn_recipe(in_tracers, out_tracers, prim.initial_style,
-                         dict(fun_jaxpr=closed_jaxpr,
-                              jvp_jaxpr_thunk=jvp_jaxpr_thunk,
-                              num_consts=len(consts) + len(env)),
-                         source_info_util.current())
-    for t in out_tracers: t.recipe = eqn
-    return out_tracers
+    # We assume partial evaluation is only performed to build linear functions,
+    # and hence we don't need to keep the custom JVP rule around anymore.
```
mattjj, line 320: Notice that we already make this assumption, e.g. in `post_process_custom_jvp_call`. We just hadn't fully leveraged it because until #3370 we needed this approach for staging.

mattjj: Actually, it turns out that we still need it for `remat`, until we upgrade to the new implementation in ad_checkpoint.py

jax/_src/custom_derivatives.py
```diff
@@ -298,119 +311,31 @@ def _apply_todos(todos, outs):
 
 custom_jvp_call_p = CustomJVPCallPrimitive('custom_jvp_call')
 
+def _custom_jvp_call_typecheck(*in_avals, call_jaxpr, jvp_jaxpr_thunk, num_consts):
+  # TODO(mattjj): could do more checking here...
```
froystig, line 315: Why prevent `core.check_call`? If it's necessary, then can we still carry out some of the checks from `core.check_call`, if not all?

mattjj: `check_call` doesn't quite work for `custom_jvp_call`; for example, there's no `call_jaxpr` parameter here. Moreover there are other things we might want to check here, like `num_consts` and maybe some things about `jvp_jaxpr_thunk` too.

So while I suspect we can factor out some shared helper functions, in general `custom_jvp_call` will need a custom typecheck rule.

froystig (approved):
Looks good. I believe this tracks/reflects what we've been doing on the custom transpose front too, so this serves to confirm that as well in a sense.

PR #9143, opened 2022-01-09
fix jaxpr type checking error messages

mattjj:
The pretty-printing changes a few months ago defined variable names based on the state in `JaxprPpContext` instances. But that meant incorrect variable names could be printed in jaxpr type checking error messages. (This probably went unnoticed because users can't produce ill-typed jaxprs unless there's a bug in JAX. But JAX devs can notice!)

This commit correctly threads through the context so as to provide error messages with coherent variable names.

~I haven't added a test case yet, and I may ask @froystig to do that, but~
Here's an example which I saw in real life:

```
jax.core.JaxprTypeError: Variable 'b' not defined

in equation:

a:f32[] b:f32[] = scan[
    jaxpr={ lambda ; c:f32[] d:f32[]. let e:f32[] = add c 1.0 in (e, d) }
    length=1
    linear=(False, True)
    num_carry=2
    num_consts=0
    reverse=False
    unroll=1
  ] f g

from source: /usr/local/google/home/mattjj/packages/jax/tests/api_test.py:3688 (scanned_f)

while checking jaxpr:

{ lambda ; a:f32[]. let
    b:f32[] c:f32[] = scan[
      jaxpr={ lambda ; d:f32[] e:f32[]. let f:f32[] = add d 1.0 in (f, e) }
      length=1
      linear=(False, True)
      num_carry=2
      num_consts=0
      reverse=False
      unroll=1
    ] a g
  in (b, c) }
```

(Notice there is no undefined 'b' variable, and the undefined variable is in fact 'g'. Moreover notice how the output binders for the scan application change names between the two times they're printed. Those things are fixed by this PR!)

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -920,7 +920,7 @@ class AbstractValue:
     raise NotImplementedError("must override")
 
   def str_short(self, short_dtypes=False):
-    raise NotImplementedError("must override")
+    return str(self)
 
 class Bot(AbstractValue): pass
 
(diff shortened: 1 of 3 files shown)


jax/core.py
```diff
@@ -2022,34 +2022,38 @@ def check_jaxpr(jaxpr: Jaxpr):
...
 
   def write(v: Var, a: AbstractValue) -> None:
-    typecheck_assert(v not in env, f"Variable '{v}' already bound")
+    typecheck_assert(v not in env, f"Variable '{pp_var(v, ctx)}' already bound")
     if not isinstance(v, DropVar):
       typecheck_assert(typecompat(v.aval, a),
-                       f"Variable '{v}' inconsistently typed as {a}, "
-                       f"bound as {v.aval}")
+                       f"Variable '{pp_var(v, ctx)}' inconsistently typed as "
+                       f"{pp_aval(a, ctx)}, bound as {pp_aval(v.aval, ctx)}")
```
mattjj, line 2056: This change is useful for jaxprs involving dynamic shapes, which have variables in types (and hence need ctx to print the correct variable names in types).

tests/core_test.py
```diff
@@ -431,17 +431,17 @@ def new_jaxpr():
     jaxpr.eqns[0].outvars[0].aval = make_shaped_array(jnp.int32(2))
     self.assertRaisesRegex(
         core.JaxprTypeError,
-        r"Variable '.' inconsistently typed as ShapedArray(.*), "
-        r"bound as ShapedArray(.*)\n\nin equation:\n\n.:i32\[\] = sin .",
+        r"Variable '.' inconsistently typed as f32\[\], "
+        r"bound as i32\[\]\n\nin equation:\n\n.:i32\[\] = sin .",
```
mattjj, line 435: These checks are more specific than they were before. The formatting is also better IMO, since it uses the short-and-readable form of types that we prefer.

tests/core_test.py
```diff
@@ -431,17 +431,17 @@ def new_jaxpr():
     jaxpr.eqns[0].outvars[0].aval = make_shaped_array(jnp.int32(2))
     self.assertRaisesRegex(
         core.JaxprTypeError,
-        r"Variable '.' inconsistently typed as ShapedArray(.*), "
-        r"bound as ShapedArray(.*)\n\nin equation:\n\n.:i32\[\] = sin .",
+        r"Variable 'b' inconsistently typed as f32\[\], "
+        r"bound as i32\[\]\n\nin equation:\n\nb:i32\[\] = sin a",
```
mattjj, line 435: I added tests! Well actually I just added variable names to the existing test. This fails on the main branch because of exactly the issue that this PR fixes.

froystig (approved):

PR #9148, opened 2022-01-10
prepare to replace/delete old remat implementation

mattjj:
A few months ago we wrote a new version of remat to support custom rematerialization policies. The new version is also built on a simpler basic implementation. But instead of replacing the pre-existing remat implementation, we kept both around because the new remat did not yet support higher-order primitives like scan and cond. Tech debt!

I'd like to start moving on replacing the old remat implementation, since it will enable some additional simplifications throughout the system (like #9137).

To that end, I'd like to first land a version of the new remat *which only supports standard remat policies for scan and cond*. That is, we won't yet have nonstandard remat policies which apply underneath scan and cond.

This restricted version is useful because it's a drop-in replacement for the old remat (which only supported a standard remat policy) and so it'll let us delete the old implementation and remove the tech debt.

This PR updates all remat tests to check the new remat, and also adds rules for scan and cond which work for default policies (and raise a NotImplementedError for nonstandard policies).

TODO:
* [x] remat-of-scan #10576
* [x] remat-of-cond #11651
* [ ] remat-of-while_loop
* [ ] flag which switches between implementations (first default to old impl and land everything, then switch to default to new but leave the flag in place in case of emergency, then delete flag and old impl)

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -300,7 +300,7 @@ ad.primitive_jvps[remat_p] = remat_jvp
 
 def remat_partial_eval(trace, *tracers, jaxpr, **params):
   assert not jaxpr.constvars
-  policy = params['policy'] or (lambda *_, **__: False)
+  policy = params['policy'] or nothing_saveable
   # unzip into jaxpr_known and jaxpr_unknown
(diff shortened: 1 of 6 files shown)


mattjj:
We ended up going with a different code approach than the one started here, though the PR message is a useful description.

PR #9154, opened 2022-01-10
Fix DeviceArray class reference

che-shr-cat:
Does it look like the DeviceArray class reference changed?


jakevdp (approved):
Good catch, thanks!

PR #9164, opened 2022-01-11
checkify: tweak some organization and names

mattjj:
The main ideas were:
1. remove redundant function `check_errors_traceable`
2. make names consistent: use "checkify" everywhere
3. some code organization: keep all rules together, put the API at the bottom


LenaMartens (approved):
LGTM!

PR #9178, opened 2022-01-12
lax.sort: regularize handling of -0.0 and -NaN

### LINKED ISSUE #6443: jnp.sort and np.sort treat negative NaNs differently
```python
import jax.numpy as jnp
import numpy as np
values = jnp.array([-np.nan, np.nan, -np.inf, np.inf, -1, 1])

print(np.sort(values))
# [-inf  -1.   1.  inf  nan  nan]

print(jnp.sort(values))
# [ nan -inf  -1.   1.  inf  nan]
```
This appears to be deliberate behavior on the part of `lax.sort`, as seen in the code comment here: https://github.com/google/jax/blob/3c362ced9090fa790a5347ce47d4b210808415e2/jax/_src/lax/lax.py#L5845-L5849

I'm not sure whether the best fix here is to change the behavior of `jnp.sort`, or to simply document that this is a known semantic difference.

jakevdp:
Fixes #6443

previously, `lax.sort` (and thus `jnp.sort`) would treat `-NaN` as less than `-inf`, and treat `-0.0` as less than `0.0`.

With this change, `sort` treats all `NaN` values as equivalent (sorting them to the end, as numpy does) and treats `-0.0` as equivalent to `0.0`, maintaining their input order in stable sorts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,13 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
     environment variable, or the ```--flax_host_callback_ad_transforms``` flag.
     Additionally, added documentation for how to implement the old behavior
     using JAX custom AD APIs ({jax-issue}`#8678`).
+  * Sorting now matches the behavior of NumPy for ``0.0`` and ``NaN`` regardless of the
+    bit representation. In particular, ``0.0`` and ``-0.0`` are now treated as equivalent,
+    where previously ``-0.0`` was treated as less than ``0.0``. Additionally all ``NaN``
+    representations are now treated as equivalent and sorted to the end of the array.
+    Previously negative ``NaN`` values were sorted to the front of the array, and ``NaN``
+    values with different internal bit representations were not treated as equivalent, and
+    were sorted according to those bit patterns ({jax-issue}`#9178`).
 
 * Bug fixes:
   * host_callback now supports ad_checkpoint.checkpoint ({jax-issue}`#8907`).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,13 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
     environment variable, or the ```--flax_host_callback_ad_transforms``` flag.
     Additionally, added documentation for how to implement the old behavior
     using JAX custom AD APIs ({jax-issue}`#8678`).
+  * Sorting now matches the behavior of NumPy for ``0.0`` and ``NaN`` regardless of the
+    bit representation. In particular, ``0.0`` and ``-0.0`` are now treated as equivalent,
+    where previously ``-0.0`` was treated as less than ``0.0``. Additionally and all ``NaN``
```
hawkinsp, line 28: remove "and"

jax/_src/lax/lax.py
```diff
@@ -3681,7 +3686,13 @@ def _float_to_int_for_sort(x):
   signed_dtype = _INT_DTYPES[nbits]
   unsigned_dtype = _UINT_DTYPES[nbits]
 
+  # Standardize NaNs: NaN is the only float value that does not equal itself.
+  x = select(ne(x, x), _nans(x), x)
```
hawkinsp, line 3690: I don't know if we have `lax.isnan(x)` but it would probably make sense to define it (like this) if we don't.

hawkinsp (approved):
You will need to run global presubmits on this internally, given we know that in the past at least one person was depending on this behavior in their test.

jax/_src/lax/lax.py
jakevdp: Good idea - I made it a private utility for now. We can expose it in the `lax` namespace in a followup.

PR #9179, opened 2022-01-12
jax2tf: Small fix for backprop through functions using bfloat16

allenlavoie:
np.issubdtype excludes bfloat16, so I believe jnp.issubdtype is more appropriate here.


mattjj (approved):
Nice! Thanks, Allen!

PR #9181, opened 2022-01-12
cache tracing of (sub)calls when forming a jaxpr

### LINKED ISSUE #7155: inner `jit` functions are re-traced (and re-compiled)
It seems that `jit` decorators are ignored for nested functions. Consider the following example (simplified from an actual program, where `inner_fn` is slightly more complex):

```python
import jax
import jax.numpy as jnp
from functools import partial

@partial(jax.jit, inline=True)
def inner_fn(state):
    print("entering inner_fn now")
    return 2*state

@jax.jit
def outer_fn(x):
    print("entering outer_fn now")
    old_x = x
    for _ in range(10):
        x = inner_fn(x)
    x = x + old_x
    return x

state = jnp.arange(5, dtype=jnp.uint32)
outer_fn(state).block_until_ready()
```

**Expected behavior**: As both functions are decorated with jit, `inner_fn` should be compiled once and the result re-used by the plain for loop inside of `outer_fn`. Consequently, the output would show only one emission of "entering inner_fn now".

**Actual behavior**: The jit decorator for `inner_fn` seems to be ignored: "entering inner_fn now" is printed 10 times and setting env var `JAX_LOG_COMPILES=1` only prints one line for `Compiling outer_fn`. In short, we observe the same behavior as if the jit decorator were absent for `inner_fn`.

**Why a bug?**: The current behavior negates any of the caching done for jitted functions, resulting in repeated tracing of a function with identical tracers and thus inflated compilation times and can easily surprise the user.

**Other considerations**:
1. Using `fori_loop`: Using a `fori_loop` results in `inner_fn` being visited only once, so that will mitigate the issue for this particular example. However, we were actually having troubles with it as it prevents unrolling the loop and our `inner_fn` was small enough so that we actually were seeing performance hits from cuda kernel launches for `inner_fn` in this case.
2. No `inline=True` for jit decorator of  `inner_fn`: In this case the compilation will compile 10 separate sub-jaxpressions for `inner_fn` that are not inlined but all invoked via `xla_call` (when inspecting this via `jax.make_jaxpr`). We see that the jit decorator is therefore not completely without effect, but this is arguably the worst case (separate `xla_call` overhead as in the `fori_loop` AND  multiple compilation of `inner_fn`).

Please:

mattjj:
cc @hawkinsp 

currently includes changes from #9188 

The main motivations here are:
1. reduce trace times (by not retracing functions to a jaxpr when we can just look up the result in a cache), and
2. reduce compilation times by lowering outlined (as opposed to inlined) functions to our compiler targets.

This PR on its own tackles (1) but not (2). It sets us up for (2) because subjaxprs are now identical Python objects when we trace to a jaxpr, and so we can have have lowerings which are themselves memoized on (sub)jaxpr object id or alternatively we can have an early pass to produce a `JaxprModule` with explicit outlined functions. The latter seems more robust and clearer.

fixes #7155

TODO:
* [ ] traceback surgery for source info (that's the last failing test)

### DIFF
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -1880,8 +1880,9 @@ class PythonPmapTest(jtu.JaxTestCase):
     self.assertEqual(count[0], 2)  # one for fwd, one for bwd
 
     with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
-      _  = jax.vjp(f, x)
+      _, f_bwd2 = jax.vjp(f, x)
       _ = f_bwd(x)
+      _ = f_bwd2(x)
     self.assertEqual(count[0], 0)  # cache hits on fwd and bwd
 
   @unittest.skipIf(jax._src.lib._xla_extension_version < 44,
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -1877,7 +1877,7 @@ class PythonPmapTest(jtu.JaxTestCase):
     with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
       _, f_bwd  = jax.vjp(f, x)
       _ = f_bwd(x)
-    self.assertEqual(count[0], 2)  # one for fwd, one for bwd
+    self.assertEqual(count[0], 2)  # once for fwd, once for bwd
 
     with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
       _, f_bwd2 = jax.vjp(f, x)
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -21,7 +21,7 @@ import itertools
 import os
 import sys
 import threading
-from typing import Any, List, Callable, NamedTuple, Optional
+from typing import Any, List, Callable, NamedTuple, Optional, Hashable
 import warnings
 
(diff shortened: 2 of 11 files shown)


tests/pmap_test.py
```diff
@@ -1877,11 +1877,12 @@ def f(x):
     with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
       _, f_bwd  = jax.vjp(f, x)
       _ = f_bwd(x)
-    self.assertEqual(count[0], 2)  # one for fwd, one for bwd
+    self.assertEqual(count[0], 3)  # once each for fwd, bwd, and _multi_slice
```
mattjj, line 1880: `_multi_slice` (used for dispatch) now gets compiled again, despite the 'warmup' call to `f(x) ` a few lines up, because it's in a new axis env context... not sure what's up with that yet.

mattjj:
@patrick-kidger check out this PR if you haven't already! Steps towards a non-inlined and faster-compiling world.

mattjj:
I think this is subsumed by #10775 (for the subjaxpr stuff) and #11298 (for the axis env stuff).

PR #9188, opened 2022-01-13
add axis env state to cache keys, fixes #9187

### LINKED ISSUE #9187: caching not sensitive to bound axis names not present on inputs (args or closed over)
Here's correct behavior, without `jit`:

```python
In [3]: def f():
   ...:     return lax.psum(1, 'i')
   ...:

In [6]: jax.vmap(f, axis_name='i', axis_size=2, out_axes=None)()
Out[6]: 2

In [7]: jax.vmap(f, axis_name='i', axis_size=3, out_axes=None)()
Out[7]: 3
```

But we get incorrect `jit` cache jits for this kind of computation:

```python
In [8]: f = jax.jit(f)

In [9]: jax.vmap(f, axis_name='i', axis_size=3, out_axes=None)()
Out[9]: DeviceArray(3, dtype=int32, weak_type=True)

In [10]: jax.vmap(f, axis_name='i', axis_size=2, out_axes=None)()
Out[10]: DeviceArray(3, dtype=int32, weak_type=True)
```

The issue for C++ jit caching is that the [`ThreadLocalJitState`](https://github.com/google/jax/blob/f0e4f0472d0e83f2f3ecb3f0ab31148015b7c1cd/jax/_src/config.py#L417) on which caching depends isn't updated for [changes to the trace-time named axis environment](https://github.com/google/jax/blob/f0e4f0472d0e83f2f3ecb3f0ab31148015b7c1cd/jax/core.py#L1853-L1870).

For Python jit caching the same issue happens, since [the trace context used as part of the cache key](https://github.com/google/jax/blob/f0e4f0472d0e83f2f3ecb3f0ab31148015b7c1cd/jax/linear_util.py#L266) also [is not sensitive to the axis env](https://github.com/google/jax/blob/f0e4f0472d0e83f2f3ecb3f0ab31148015b7c1cd/jax/_src/config.py#L335).

The same problem exists for the control flow primitive caches:

```python
In [21]: scanned_f = lambda _, __: (lax.psum(1, 'i'), None)

In [22]: f = lambda: lax.scan(scanned_f, 0, None, length=1)[0]

In [23]: jax.vmap(f, axis_name='i', axis_size=2, out_axes=None)()
Out[23]: DeviceArray(2, dtype=int32, weak_type=True)

In [24]: jax.vmap(f, axis_name='i', axis_size=3, out_axes=None)()
Out[24]: DeviceArray(2, dtype=int32, weak_type=True)
```

The solution there is similarly to pass in a hashable version of the trace-time axis environment as an argument to [the memoized jaxpr-forming functions](https://github.com/google/jax/blob/f0e4f0472d0e83f2f3ecb3f0ab31148015b7c1cd/jax/_src/lax/control_flow.py#L71-L118).

How did we not notice until now? Well, any kind of data dependence (on a value mapped over the named axis in question) makes the caches work. So this seems to be primarily about things like jitting/scanning simple functions involving `axis_size('i')`, `axis_index('i')`, or `random.normal(key, shape=NamedShape(3, i=4))`. Tests involving the latter is what uncovered the issue!

mattjj:
fixes #9187 

This change could cause some new and even unnecessary cache misses. That is, while the cache hits in #9187 were buggy and should be cache misses, to correct them we may end up being overly defensive.

The reason for defensiveness is that in general, even for a fixed sequence of input abstract values to a jitted function, whenever the named axis environment (which represents the mapped-over names bound by `vmap`, `pmap`, and/or `xmap`) is different from one we've seen before with those input abstract values, we need to re-trace the Python callable. That's because the Python callable could bind a collective like `psum` or `axis_index`, and the result of such a collective depends on the axis environment. (When a mapped axis has no name, i.e. when its name is `core.no_axis_name`, this error can't happen. So for caching purposes unnamed axes can be ignored.)

But some Python callables may not involve collectives at all! It'd be a waste to re-trace those. One such instance arose in this PR: the `_multi_slice` function, which is used in `pmap` dispatch, was getting re-traced (and re-compiled) even though it binds no collective primitives. So I introduced a low-level way to, in effect, promise that no collectives are bound: use `core.stash_axis_env` to temporarily empty out the axis environment and thus effectively ignore its value.

Still, this issue could arise in user code... is the only option to provide such a "promise" API to users?

TODO:
* [ ] wrap (almost) all jax.numpy functions in `core.stash_axis_env` (tried this and ran into a weird `__doc__` issue...)

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -21,7 +21,7 @@ import itertools
 import os
 import sys
 import threading
-from typing import Any, List, Callable, NamedTuple, Optional
+from typing import Any, List, Callable, NamedTuple, Optional, Hashable
 import warnings
 
 from jax._src import lib
@@ -324,15 +324,8 @@ class Config:
 
     return _StateContextManager(name, help, update_thread_local_hook, validate)
 
-  def _trace_context(self):
-    """Returns a tuple of configuration values that affect tracing.
-
-    These values are included in the cache key for linear_util.cache.
-
-    Values included in this set should also most likely be included in
-    the C++ JIT state, which is handled separately."""
-    return (self.x64_enabled, self.jax_numpy_rank_promotion,
-            self.jax_default_matmul_precision)
+  def get_trace_state(self):
+    return get_trace_state()
 
 class _StateContextManager:
   def __init__(self, name, help, update_thread_local_hook,
@@ -405,7 +398,7 @@ already_configured_with_absl = False
 
 class GlobalJitState(NamedTuple):
   numpy_rank_promotion: Optional[str] = None
-  default_matmul_precision: Optional[Any] = None
+  default_matmul_precision: Optional[Hashable] = None
 
 
 def update_global_jit_state(**kw):
@@ -415,15 +408,27 @@ def update_global_jit_state(**kw):
 
 
 class ThreadLocalJitState(NamedTuple):
-  dynamic_trace_state: Optional[Any] = None
+  dynamic_trace_state: Optional[Hashable] = None
+  axis_env_state: Optional[Hashable] = None
   numpy_rank_promotion: Optional[str] = None
-  default_matmul_precision: Optional[Any] = None
+  default_matmul_precision: Optional[Hashable] = None
 
 
 def update_thread_local_jit_state(**kw):
   tls = jax_jit.thread_local_state()
   context = tls.extra_jit_context or ThreadLocalJitState()
   tls.extra_jit_context = context._replace(**kw)
+
+
+def get_trace_state() -> Hashable:
+  # This function is used for Python-based caching (as opposed to caching in C++
+  # dispatch code). It reads the axis_env_state from the C++ thread-local state,
+  # but gets everything else off the config object, which reflects the
+  # appropriate unification of global and thread-local settings.
+  tls = jax_jit.thread_local_state()
+  ctx = tls.extra_jit_context or ThreadLocalJitState()
+  return (ctx.axis_env_state, config.jax_enable_x64, config.jax_disable_jit,
+          config.jax_numpy_rank_promotion, config.jax_default_matmul_precision)
 
 
 # TODO(mattjj): remove all uses of this flag
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -21,7 +21,7 @@ import itertools
 import os
 import sys
 import threading
-from typing import Any, List, Callable, NamedTuple, Optional, Hashable
+from typing import Any, List, Callable, NamedTuple, Optional, Hashable, Tuple
 import warnings
 
 from jax._src import lib
@@ -409,7 +409,7 @@ def update_global_jit_state(**kw):
 
 class ThreadLocalJitState(NamedTuple):
   dynamic_trace_state: Optional[Hashable] = None
-  axis_env_state: Optional[Hashable] = None
+  axis_env_state: Tuple[Hashable, ...] = ()
   numpy_rank_promotion: Optional[str] = None
   default_matmul_precision: Optional[Hashable] = None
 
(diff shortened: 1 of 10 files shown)


tests/pmap_test.py
```diff
@@ -1877,11 +1877,12 @@ def f(x):
     with jtu.count_jit_and_pmap_compiles() as count:  # noqa: F841
       _, f_bwd  = jax.vjp(f, x)
       _ = f_bwd(x)
-    self.assertEqual(count[0], 2)  # one for fwd, one for bwd
+    self.assertNotEqual(count[0], 0)
```
mattjj, line 1880: _multi_slice (used for dispatch) now gets compiled again, despite the 'warmup' call to f(x) a few lines up, because it's in a new axis env context... not sure what's up with that yet. JAX_LOG_COMPILES=1 prints out a different object id for it!

mattjj: Indeed, on the first call the axis env is empty and on the second it's got something in it. That doesn't affect _multi_slice though.

More generally, it might be nice to promise that some functions don't depend on the axis environment (in the sense of #9187, like not using any collectives), so we don't have to cache miss when it changes... one way is we could temporarily set the axis env to empty when we trace such functions.

mattjj: I've added such a mechanism!

jax/_src/util.py
```diff
@@ -199,7 +199,7 @@ def wrapper(*args, **kwargs):
       if config.jax_check_tracer_leaks:
         return f(*args, **kwargs)
       else:
-        return cached(config._trace_context(), *args, **kwargs)
+        return cached(config.get_thread_local_trace_state(), *args, **kwargs)
```
hawkinsp, line 202: I don't think this part is correct. In general you need to merge the thread-local state with the global state to get a cache key, I think?

mattjj: Ah good, this is exactly the kind of thing I wanted to check!

LMK if you have a pointer to where this is already done somewhere (even if in C++) so I can check against it...

mattjj: We think this is actually correct: the other global parts of the state are already introduced in this function by it reading the relevant attributes from the `config` object. But the function name was misleading since it included "thread_local". I'll change the name!

Separately, I may look at making this faster...

jax/_src/config.py
```diff
@@ -426,6 +420,17 @@ def update_thread_local_jit_state(**kw):
   tls.extra_jit_context = context._replace(**kw)
 
 
+def get_trace_state() -> Hashable:
+  # This function is used for Python-based caching (as opposed to caching in C++
+  # dispatch code). It reads the axis_env_state from the C++ thread-local state,
```
hawkinsp, line 425: Perhaps comment why this is the case? (i.e., there's no such thing as a global axis environment.)

jax/core.py
```diff
@@ -1853,21 +1853,46 @@ def _unmap_shaped_array(size: int, axis_name, axis: int, aval: ShapedArray) -> S
...
+    jax_config.update_thread_local_jit_state(
+        axis_env_state=tuple(f for f in ts.axis_env
+                             if f.name is not no_axis_name))
+
+@contextmanager
+def stash_axis_env():
+  "Promise that a function or with-suite does not depend implicitly on axis env"
+  s = thread_local_state.trace_state
+  prev_axis_env, s.axis_env = s.axis_env, []
+  jax_config.update_thread_local_jit_state(axis_env_state=())
```
hawkinsp, line 1890: I wonder if it would be possible to put a dummy value that caused an error if you were to actually try to use it here?

hawkinsp (approved):

jax/_src/config.py
mattjj: Done!

jax/core.py
mattjj: Such things are possible...

The current approach would also lead to an error, just with a potentially misleading error message (unbound axis name).

mattjj: I added a comment to this effect but didn't add any fancy error mechanisms. Gotta leave some work for the future, right?

mattjj:
Subsumed by [11298](https://github.com/google/jax/pull/11298).

PR #9194, opened 2022-01-14
Update contributing.md with OS X instructions

Naqu6:
Hi! I'm following the instructions [here](https://jax.readthedocs.io/en/latest/contributing.html) to install JAX on my local machine for contributing (OS X), and I ran into an issue running one of the pip commands for installation. Specifically, running
`pip install -e .[cpu]` causes an issue because zsh, the default shell on OS X, uses square brackets as part of the [globbing syntax](https://zsh.sourceforge.io/Guide/zshguide05.html#l137).

So, I modified the docs with a note indicating a fix for that I've tested on my machine.

Feel free to reject this PR if contributing to JAX isn't supported on OS X for whatever reason.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -42,6 +42,15 @@ Follow these steps to contribute code:
    pip install -r build/test-requirements.txt  # Installs all testing requirements.
    pip install -e .[cpu]  # Installs JAX from the current directory in editable mode.
    ```
+   
+
+   *Note*: if you're using [zsh](https://en.wikipedia.org/wiki/Z_shell) on OS X and
+   the command `pip install -e .[cpu]` fails with the error
+   `zsh: no matches found: .[cpu]`, try running `pip install -e ."[cpu]"` instead.
+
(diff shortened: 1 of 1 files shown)


Naqu6:
My CLA should because I'm currently a Student Researcher at Google (ldap: ryanehrlich@). When I link my Google account to my github account (again) the CLA issue should disappper.

jakevdp:
Instead of adding a paragraph on zsh-specific instructions, perhaps it would be better to update the problematic line? I think ` pip install -e ".[cpu]"` should work on most shells.

Naqu6:
That sounds good to me! Thanks for the suggestion.

jakevdp:
I re-ran the CLA check and it's still failing. You'll need to make sure the email you used for the commit is the same as the one you used to register the CLA.

Naqu6:
Thanks, I'll redo the commit with my Google email. I'll close this an open it again when I've figured that out.

PR #9196, opened 2022-01-14
cache initial-style jaxpr transformations

### LINKED ISSUE #3847: slow jit-compilation with odeint inside tfp.optimizer.lbfgs_minimize
I want to optimize the initial condition of a dynamical system w.r.t. a criterion based on the system's trajectory. To this end, I use `jax.experimental.ode.odeint` inside of `tfp.optimizer.lbfgs_minimize` (cf. [TFP-on-JAX L-BFGS](https://github.com/google/jax/issues/1400#issuecomment-661889919)). This might sound very particular, but I suppose this will come up more often when using JAX in the context of physics models.

However, jit-compilation is very slow, which impedes scaling-up.
Is there a good reason for slow compilation or is there a way to speed this up?

Below you see a code example and two graphs that show the compile times and the length of the code generate during jit-compilation, i.e., `len(str(jax.make_jaxpr(...)(...)))`, as a function of `odeint` integration steps. The curves have similar shape and become somewhat flat.

```python
!pip install tfp-nightly==0.12.0.dev20200723

from functools import partial
import jax
import jax.numpy as jnp
from tensorflow_probability.substrates import jax as tfp


def ode_func(x, t):
  """
  Time derivative of a linear ODE that decays the state.
  """
  return -0.1 * x

def loss(x, n_integration_steps):
  """
  Minimizing this function essentially just minimizes the norm of x.
  """
  t = jnp.asarray([n*0.1 for n in range(n_integration_steps)])
  y = odeint(ode_func, x, t)
  return jnp.sum(jnp.square(y))  

@partial(jax.jit, static_argnums=(1,))
def optimize(start, n_integration_steps):
  """
  Optimizes the initial condition of the ODE system using L-BFGS.
  """
  loss_fn = partial(loss, n_integration_steps=n_integration_steps)
  optim_results = tfp.optimizer.lbfgs_minimize(
        jax.value_and_grad(loss_fn),
        initial_position=start,
        num_correction_pairs=10,
        tolerance=1e-8)
  return optim_results.converged, optim_results.position


ndims = 3
n_integration_steps = 10
start = jnp.arange(ndims, 0, -1.)
converged, results = optimize(start, n_integration_steps)
```

compile_times jaxpr_str_len

### LINKED ISSUE #9187: caching not sensitive to bound axis names not present on inputs (args or closed over)

mattjj:
These changes reduce the tracing time on [the shoyer version](https://github.com/google/jax/issues/3847#issuecomment-702845696) of the code in #3847 from ~50s to ~2.5s.

~I'm a little wary to add more caching until we land #9188 to fix #9187.~ But otherwise this is straightforward!

I think we can say this fixes #3847, though there's a lot more improvement we could make in this area.

### DIFF
--- a/jax/interpreters/ad.py
+++ b/jax/interpreters/ad.py
@@ -27,8 +27,8 @@ from jax.core import (Trace, Tracer, get_aval, call_p, Primitive, Literal,
                     raise_to_shaped)
 from jax._src.ad_util import (add_jaxvals, add_jaxvals_p, zeros_like_jaxval,
                               zeros_like_aval, zeros_like_p, Zero)
-from jax._src.util import (unzip2, safe_map, safe_zip, split_list,
-                         wrap_name, as_hashable_function)
+from jax._src.util import (unzip2, safe_map, safe_zip, split_list, wrap_name,
+                           as_hashable_function, cache)
 from jax.tree_util import register_pytree_node
 from jax import linear_util as lu
 from jax._src.api_util import flatten_fun, flatten_fun_nokwargs
@@ -637,6 +637,11 @@ def map_transpose(primitive, params, call_jaxpr, args, ct, _, reduce_axes):
 
 
 def jvp_jaxpr(jaxpr, nonzeros, instantiate):
+  inst = tuple(instantiate) if isinstance(instantiate, list) else instantiate
+  return _jvp_jaxpr(jaxpr, tuple(nonzeros), inst)
+
+@cache()
+def _jvp_jaxpr(jaxpr, nonzeros, instantiate):
   assert len(jaxpr.in_avals) == len(nonzeros)
   f = lu.wrap_init(core.jaxpr_as_fun(jaxpr))
   f_jvp, out_nonzeros = f_jvp_traceable(jvp(f, instantiate=instantiate), nonzeros)
--- a/jax/interpreters/batching.py
+++ b/jax/interpreters/batching.py
@@ -28,7 +28,7 @@ from jax._src.ad_util import (add_jaxvals, add_jaxvals_p, zeros_like_jaxval,
 from jax import linear_util as lu
 from jax._src.util import (unzip2, safe_map, safe_zip, wrap_name, split_list,
                            canonicalize_axis, moveaxis, as_hashable_function,
-                           curry, memoize)
+                           curry, memoize, cache)
 from jax.interpreters import partial_eval as pe
 
 map = safe_map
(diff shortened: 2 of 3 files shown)


froystig (approved):

mattjj:
This was merged, but not marked as merged...

PR #9201, opened 2022-01-14
Checkify: add way to disable categories of errors.

LenaMartens:
By default only user_asserts are lifted into the checked function.

Current error categories can be freely combined through set operations:
```
ErrorCategory = enum.Enum('ErrorCategory', ['NAN', 'OOB', 'DIV', 'ASSERT'])

float_errors = {ErrorCategory.NAN, ErrorCategory.DIV}
index_errors = {ErrorCategory.OOB}
automatic_errors = float_errors | index_errors
user_asserts = {ErrorCategory.ASSERT}
```


mattjj (approved):
Looks awesome!

PR #9204, opened 2022-01-14
Checkify: rename some symbols and add some docstrings.

LenaMartens:
Thought I'd start a draft PR with name-changes/docstrings to iterate on. Draft because I need to merge with changes in https://github.com/google/jax/pull/9201, and because the docstrings need some more work.

Main changes:
- `assert_` renamed to `assert_that` (other options: `assertion` which is not in imperative form,  `check` which maybe aligns nicely with `checkify`)
- `assert2_` renamed to `reassert`: want to avoid referring to discharging/recharging, but can't think of a good antonym for `functionalize` that still communicates that we're re-injecting the error. Maybe `recharge_error` is best, and we can commit to explaining what that means (+it seems like a more advanced API so maybe it's fine to refer to it as recharging)
- `reassert` now takes an error object instead of code/msgs (this seems more natural to me)
- add `Error.throw` for convenience


LenaMartens:
`assert_` and `assert2_` suggestion: `check` and `check_error` (from @tomhennigan)

jax/experimental/checkify.py
```diff
@@ -212,26 +217,71 @@ def checkify_fun_to_jaxpr(f, error, in_avals):
 
 ## assert primitive
 
-def assert_(pred: Bool, msg: str) -> None:
+def assert_that(pred: Bool, msg: str) -> None:
+  """Assert a predicate, add an error with msg if predicate evaluates to False.
+
+  The predicate needs to be a jnp boolean.
```
mattjj, line 223: I think Python booleans work too! Maybe 'needs to be a boolean or rank-0 array with boolean dtype'? (Or am I missing something?)

jax/experimental/checkify.py
```diff
@@ -212,26 +217,71 @@ def checkify_fun_to_jaxpr(f, error, in_avals):
...
+
+  Args:
+    pred: jnp bool
+    msg: msg to print if assertion fails.
+  Returns:
+    Nothing: this is an effectful operation, and needs to be functionalized
```
mattjj, line 229: Technically it returns `None`, since Python functions always have a return value.

Maybe instead of having the description about how to functionalize, we can just provide a link to a docs page which has all the info.

LenaMartens: yep, I thought it might be a nice place for discussing the fact that it's an effect that needs to be discharged (because my linter was complaining about the `Returns` section missing), but it's just an experiment. Are you thinking about a "checkify" doc to link to?

jax/experimental/checkify.py
LenaMartens: True! I wanted to make sure it mentions that you have to use `jnp` on traced values (to not get ConcretizationErrors) but that's not the same as requiring `jnp` bools.

LenaMartens:
Some progress in making valid docstrings, and a bit on the wording (but not really happy with it yet: https://jax--9204.org.readthedocs.build/en/9204/jax.experimental.html#jax.experimental.checkify.checkify)

I think we might have converged on `check` and `check_error` as good names for `assert_` and `assert2_`?

LenaMartens:
https://jax--9204.org.readthedocs.build/en/9204/_autosummary/jax.experimental.checkify.checkify.html#jax.experimental.checkify.checkify

jax/experimental/checkify.py
```diff
@@ -229,26 +237,86 @@ def checkify_fun_to_jaxpr(f, error, enabled_errors, in_avals):
...
+def check_error(error: Error) -> None:
+  """Check if an error has occurred.
+
+  When running in an un-staged function, this will throw the error
+  if an error has occured.
+  When running in a staged function, the error will be threaded back
+  into the function, so it can be functionalized later.
+
+  This is useful when you want to re-check an error after you've
+  functionalized a function.
```
mattjj, line 286:
```suggestion
  `check_error` is useful when you want to turn checks represented by an `Error` value
  (produced by functionalizing `check`s via `checkify`) back into Python Exceptions.
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+user_checks = frozenset({ErrorCategory.USER_CHECK})
 
 Out = TypeVar('Out')
-def checkify(fun: Callable[..., Out], errors: Set[ErrorCategory] = user_asserts
+
+
+def checkify(fun: Callable[..., Out],
+             errors: FrozenSet[ErrorCategory] = user_checks
              ) -> Callable[..., Tuple[Error, Out]]:
+  """Check for run-time errors in ``fun``.
```
mattjj, line 580:
```suggestion
  """Functionalize `check` calls in `fun`, and optionally add run-time error checks.
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+
+  Run-time errors are either user-added ``checkify.check`` assertions, or
+  automatically added checks like NaN checks, depending on the ``errors``
+  argument.
+
+  The returned function will return an Error object along with the output of
+  the function. ``err.get()`` will either return ``None`` (if no error occurred)
```
mattjj, line 587:
```suggestion
  the original function. ``err.get()`` will either return ``None`` (if no error occurred)
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+def checkify(fun: Callable[..., Out],
+             errors: FrozenSet[ErrorCategory] = user_checks
              ) -> Callable[..., Tuple[Error, Out]]:
+  """Check for run-time errors in ``fun``.
+
+  Run-time errors are either user-added ``checkify.check`` assertions, or
+  automatically added checks like NaN checks, depending on the ``errors``
+  argument.
+
+  The returned function will return an Error object along with the output of
```
mattjj, line 586:
```suggestion
  The returned function will return an Error object `err` along with the output of
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+  the function. ``err.get()`` will either return ``None`` (if no error occurred)
+  or a string containing an error message. This error message will correspond to
+  the first error which occurred.
+
+  The kinds of errors are:
+    - ErrorCategory.USER_CHECK: a ``checkify.check`` predicate evaluated
```
mattjj, line 592: Why make these all-caps, like `USER_CHECK` instead of `user_check`?

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+      to False.
+    - ErrorCategory.NAN: a floating-point operation generated a NaN value
+      as output.
+    - ErrorCategory.DIV: division by zero
+    - ErrorCategory.OOB: an indexing operation was out-of-bounds
+
+  Multiple categories can be enabled together (eg. ``errors={ErrorCategory.NAN,
```
mattjj, line 599:
```suggestion
  Multiple categories can be enabled together by creating a `Set` (eg. ``errors={ErrorCategory.NAN,
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
 automatic_errors = float_errors | index_errors
-user_asserts = {ErrorCategory.ASSERT}
+user_checks = frozenset({ErrorCategory.USER_CHECK})
 
 Out = TypeVar('Out')
-def checkify(fun: Callable[..., Out], errors: Set[ErrorCategory] = user_asserts
+
+
+def checkify(fun: Callable[..., Out],
+             errors: FrozenSet[ErrorCategory] = user_checks
```
mattjj, line 578:
```suggestion
             errors: Set[ErrorCategory] = user_checks
```

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+      as output.
+    - ErrorCategory.DIV: division by zero
+    - ErrorCategory.OOB: an indexing operation was out-of-bounds
+
+  Multiple categories can be enabled together (eg. ``errors={ErrorCategory.NAN,
+  ErrorCategory.OOB}``).
+
+  Args:
+    fun: Callable which can contain user checks (see ``check``).
+    errors: Enabled errors. Options are NAN, OOB, DIV. By default USER_CHECK is
```
mattjj, line 604: Mention the type is a `Set` containing `ErrorCategory` values, and how one selects options by building a set?

jax/experimental/checkify.py
```diff
@@ -495,16 +564,65 @@ def assert_discharge_rule(error, enabled_errors, pred, code, *, msgs):
...
+      enabled.
+  Returns:
+    A function with output signature (Error, out)
```
mattjj, line 607:
```suggestion
    A function which accepts the same arguments as ``fun`` and returns as output
    a pair where the first element is an ``Error`` value, representing any failed
    ``check``s, and the second element is the original output of ``fun``.
```

mattjj (approved):
LGTM! Some suggestions and questions below. But we can just land and iterate; no need to block any changes.

jax/experimental/checkify.py
```diff
@@ -229,26 +237,86 @@ def checkify_fun_to_jaxpr(f, error, enabled_errors, in_avals):
...
-  return assert_p.bind(pred, code, msgs=msgs)
-
-assert_p = core.Primitive('assert')
+def check_error(error: Error) -> None:
+  """Check if an error has occurred.
+
+  When running in an un-staged function, this will throw the error
+  if an error has occured.
+  When running in a staged function, the error will be threaded back
+  into the function, so it can be functionalized later.
```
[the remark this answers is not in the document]
LenaMartens: yay, nice! I especially like the "inverse of checkify" as a way of explaining this.

jax/experimental/checkify.py
LenaMartens: QQ: I made this a FrozenSet because of a "dangerous default value" warning, but I guess the Set type is a super-set of the FrozenSet type, and it can still accept Sets even though the default value will be a FrozenSet? (I think I answered my own question)

jax/experimental/checkify.py
LenaMartens: They are the enum values, which I think style-wise [are dictated to be upper case](https://mail.python.org/pipermail/python-ideas/2016-September/042340.html)? 

It probably makes more sense to refer to the sets here instead of the enum values, because those are the ones which need to be passed into the function. Maybe every enum value should have a corresponding (camel_case) set?

jax/experimental/checkify.py
LenaMartens: nit: representing any failed ``check``s -> representing the first failed ``check``?

LenaMartens:
Thanks for the suggestions, nice iteration! I've accepted the suggestions, we can iterate in a follow-up?

jax/experimental/checkify.py
mattjj: Ah thanks for teaching me about that style guidance. (I always feel like all-caps variables are yelling at me...)

Having a corresponding singleton set for each case makes sense... so does perhaps making this function flexible enough to handle both (i.e. a raw enum value as well as sets). We can fiddle with that after adding!

jax/experimental/checkify.py
LenaMartens: Mentioned the type and how to recombine sets/build sets. (still a bit awkward, but let's iterate)

jax/experimental/checkify.py
LenaMartens: ah whoops, flake is not happy with that: `Incompatible default for argument "errors" (default has type "FrozenSet[ErrorCategory]", argument has type "Set[ErrorCategory]") `

Issue #9135, opened 2022-01-08
`einsum` with more than 2 operands accepts only length 2 tuples as the `precision` argument


dwf:
```python
import jax.numpy as jnp
import jax.lax as lax
b = 11; c = 3; h = 5; v = 13; k = 2
x = jnp.ones((b, c, h), dtype=jnp.int8)
y = jnp.ones((c, h, k), dtype=jnp.float32)
z = jnp.ones((c, k, v), dtype=jnp.int8)
jnp.einsum('bch,chk,ckv->bv', x, y, z, precision=(lax.Precision.DEFAULT, lax.Precision.HIGHEST, lax.Precision.DEFAULT))
```
Based on my fledgling understanding of per-argument precision in XLA I expected this to work, but instead I get
```
ValueError: Precision argument must be None, a string in ['highest', 'float32', 'high', 'bfloat16_3x', 'tensorfloat32', 'default', 'bfloat16', 'fastest', None], a lax.Precision value or a tuple of two lax.Precision values or strings; got (<Precision.DEFAULT: 0>, <Precision.HIGHEST: 2>, <Precision.DEFAULT: 0>).
```
The docs do say that if it's a tuple, it should be length 2, but the semantics of that aren't super clear to me (and what I think the semantics are don't make a ton of sense). It seems like it's doing successive `dot_general`s on 2 arguments at a time, and reusing the same length 2 precision tuple on each, which seems like maybe a bit of an oversight.

However, it's not super clear what the right thing to do is for per-argument precision in the n>2 operand case, if you were to accept as many precision flags as there are operands. If that were allowed, I suppose you could define an inherit-higher or inherit-lower precision policy for intermediate results (perhaps switched between with a boolean flag), and suggest any more complicated policy be implemented by expanding the einsum into several explicitly. Alternatively, you could disallow tuple precision arguments in the n>2 case.

jakevdp:
Thanks for the report – agreed that the semantics of N>2 precision arguments is not obvious. I think your idea to disallow tuple precisions for N>2 is probably the best way forward.

PR #9205, opened 2022-01-14
einsum: clarify use of precision.

jakevdp:
Addresses #9135

I chose not to raise any errors here for tuple precisions, because despite the somewhat confusing semantics, the behavior is well-defined and there's no need to break existing code. Instead, I improved the documentation.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4935,7 +4935,13 @@ def tensordot(a, b, axes=2, *, precision=None):
                          precision=precision)
 
 
-@_wraps(np.einsum, lax_description=_PRECISION_DOC, skip_params=['out'])
+_EINSUM_DOC = _PRECISION_DOC + """\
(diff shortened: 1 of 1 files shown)


skye (approved):

shoyer:
I think there would be a case for breaking existing behavior if there are more than two arguments to `einsum`, if the two precisions are different.

I doubt anyone is using this functionality and we might want to change it going forward, e.g., to suppose supplying one precision per einsum argument.

jakevdp:
Agreed one precision per einsum argument would be more intuitive, but the semantics are not clear. e.g. if you have `A, B, C` with precision `lowest, default, highest`, this becomes two `dot_general` calls:
```python
tmp = dot_general(A, B, precision=['lowest', 'default'])
result = dot_general(tmp, C, precision=[X, 'highest'])
```
It's not clear to me what `X` should be in this case.

shoyer:
Fair enough, I guess this probably covers all the use-cases then :)

PR #9207, opened 2022-01-14
[sparse] change call signature of coo primitive wrappers

jakevdp:
The main change here is that the signature of lower-level functions is changed to accept higher-level objects rather than directly accepting buffers; for example, for `coo_matvec_p` we previously had a function that looks like this:
```python
def coo_matvec(data: Array, row: Array, col: Array, v: Array, *, shape: Tuple[int]):
  return coo_matvec_p.bind(data, row, col, v, *, shape)
```
whereas after this change we have a pair of functions, with the public version accepting a `sparse.COO` object directly, something like this:
```python
def coo_matvec(M: COO, v: Array):
    return _coo_matvec(M.data, M.row, M.col, v, shape=M.shape)

def _coo_matvec(data: Array, row: Array, col: Array, v: Array, *, shape: Tuple[int]):
  return coo_matvec_p.bind(data, row, col, v, *, shape)
```

Why this change? This has the goal of expanding the metadata that `COO` matrices can carry with them. For the sake of cusparse lowering, we need indices to be sorted, but sorting is an expensive operation. A solution to that is to carry an `is_sorted` flag with the matrix so that we need only sort the indices if necessary.

The problem is, every primitive will then need to be able to accept this flag for sparse inputs, and return this flag for sparse outputs. But primitives have no way of returning static metadata, so we'll need a layer around each primitive call to handle that metadata output.

So, for example, we have `coo_fromdense_p`, which currently just returns three buffers, and the `coo_fromdense` wrapper mirrors this API. This PR changes it so that `coo_fromdense` returns a `COO` object, which contains the three buffers as well as any necessary metadata.

This is a bit awkward because it's a departure from our current habit of making these low-level functions basically match the API of the primitives they wrap. I'm not entirely married to that idea, but we will need *something* between the high-level objects and the raw primitive calls, so this feels like a decent option.

### DIFF
--- a/jax/experimental/sparse/coo.py
+++ b/jax/experimental/sparse/coo.py
@@ -15,7 +15,7 @@
 """COO (coordinate format) matrix object and associated primitives."""
 
 import operator
-from typing import Tuple
+from typing import Any, NamedTuple, Tuple
 import warnings
 
 import numpy as np
@@ -39,6 +39,14 @@ try:
 except ImportError:
   hipsparse = None
 
+
+Dtype = Any
+Shape = Tuple[int, ...]
+
+class COOInfo(NamedTuple):
+  shape: Shape
+
+
 @tree_util.register_pytree_node_class
 class COO(JAXSparse):
   """Experimental COO matrix implemented in JAX; API subject to change."""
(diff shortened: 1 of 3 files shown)


jakevdp:
I can't request a review from a non-maintainer, but I'd love input from @tlu7 as well

froystig (approved):
Makes sense!

tlu7:
Thanks for making this change! LGTM. I really like the comments you added for the change. Super clear for me to understand the change.

PR #9209, opened 2022-01-14
[sparse] add bcoo_sort_indices

jakevdp:


froystig (approved):

jax/experimental/sparse/bcoo.py
```diff
@@ -119,6 +119,22 @@ def _bcoo_sum_duplicates_unbatched(data, indices, *, shape, nse, remove_zeros):
...
   return data_unique, indices_unique, nse
 
+def _bcoo_sort_indices(data, indices, shape):
+  props = _validate_bcoo(data, indices, shape)
+  def f(data, indices):
+    _, N = indices.shape
+    if N == 0:
+      return data, indices
+    elif data.ndim > 1:
+      i = jnp.lexsort(indices[:, ::-1].T)
```
soraros, line 129: Will unfolding `lexsort` here be helpful since that way we can save a `gather` in `indices[i]`.

jakevdp: Im not sure what you mean by unfolding

soraros: *inline

jakevdp: Thanks, but I'm not exactly sure how that applies here. Could you give an example of what you have in mind?

soraros: I mean inside `lexsort` the sorted indices are computed, and only to be thrown away in its return statement `lax.sort((*keys[::-1], iota), dimension=axis, num_keys=len(keys))[-1]`. So I've been thinking maybe if we inline `lexsort` here we can save a indexing which is linear wrt. `N`.

jakevdp: Thanks, Done.

PR #9222, opened 2022-01-18
Extend .compiler_ir() to accept kwargs controlling IR emission.

stellaraccident:
As a user of this API, is important to have access to 'binary=', 'enable_debug_info=', and 'print_generic_op_form=' as different tasks and integrations call for different things (i.e. 'binary=' saves time/space on large programs, 'print_generic_op_form=' is more compatible with tooling to ensure compatibility, etc).

I debated just having it return a live Module object, but I think it is still wise for the moment to keep this API narrow, and that would be a large expansion.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -521,10 +521,23 @@ class Lowered:
         self._lowering.compile(), self.in_tree, self.out_tree,
         self.donate_argnums, self._no_kwargs)
 
-  def compiler_ir(self, dialect: Optional[str] = None):
+  def compiler_ir(self, dialect: Optional[str] = None, **kwargs):
+    """Gets compiler IR of a specified 'dialect'
+
+    Args:
+      dialect: One of 'mhlo' or 'hlo' (or None for 'hlo').
+      **kwargs: Dialect specific keyword arguments:
+        'hlo': None
+        'mhlo': Keyword arguments match the MLIR Operation.print() method.
+
+    Returns:
+      A dialect specific representation.
+    """
     if dialect == "mhlo":
-      return self._lowering.mhlo()
+      return self._lowering.mhlo(**kwargs)
     elif dialect == "hlo" or dialect is None:
+      if kwargs:
+        raise TypeError(f"Unsupported keyword arguments: {kwargs.keys()}")
       return self._lowering.hlo()
     else:
       raise ValueError(f"Unknown dialect {dialect}")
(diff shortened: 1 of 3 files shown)


stellaraccident:
I wasn't able to find formatting options. Let me know if there is an auto-formatter I should be using.

jax/interpreters/mlir.py
```diff
@@ -405,10 +405,16 @@ def lower_jaxpr_to_module(
...
-  module.operation.print(file=output, enable_debug_info=True,
-                         print_generic_op_form=False)
+def module_to_string(module: ir.Module, *,
+    binary: bool = False, enable_debug_info: bool = True,
+    print_generic_op_form: bool = False) -> Union[bytes, str]:
+  if binary:
+    output = io.BytesIO()
+  else:
+    output = io.StringIO()
+  module.operation.print(file=output, binary=binary,
```
hawkinsp, line 415: OOC, what does "binary" do on the print function? Avoid escaping certain strings?

jax/_src/api.py
```diff
@@ -521,10 +521,23 @@ def compile(self) -> 'Compiled':
         self._lowering.compile(), self.in_tree, self.out_tree,
         self.donate_argnums, self._no_kwargs)
 
-  def compiler_ir(self, dialect: Optional[str] = None):
+  def compiler_ir(self, dialect: Optional[str] = None, **kwargs):
```
hawkinsp, line 524: I'm increasingly leaning towards making `compiler_ir` return an `ir.Module` object. This is mostly because MLIR printing is not free. In particular, I recently ran across a case where the printed IR might be disproportionately large compared to the C++ object graph (callsiteloc).

This would have a downside that the `ir.Module` object may not agree between different Python extensions. But perhaps that isn't so bad and it could be covered by documentation?

soraros: > This is mostly because MLIR printing is not free.

Something related, don't know if worth creating an issue, so I will just mention it here. Large constants are not hidden in MLIR printing, even a 10 by 10 matrix can bloat printed IR.

hawkinsp: @soraros is that causing a noticeable problem for you? If so, file a bug?

stellaraccident: I'm +1 on returning a Module. Code that crosses boundaries already needs to be aware of such things.

binary=True builds into a bytes based IO. For tool interop, the additional python str conversion is pure overhead (shouldn't change the size at rest much but python has to make multiple passes/copies in order to determine and use the minimal encoded string representation).

hawkinsp:
Closing in favor of https://github.com/google/jax/pull/9242

PR #9230, opened 2022-01-18
Correct lax.while_loop type annotation

NeilGirdhar:


NeilGirdhar:
This was my fault.  I forgot to consider Boolean arrays in the result.  I've checked the rest of that pull request and this was the only error of this type.

jax/_src/lax/control_flow.py
```diff
@@ -67,6 +67,7 @@
 
 T = TypeVar('T')
 Array = Any
+BooleanNumeric = Any  # A bool, or a Boolean array.
```
NeilGirdhar, line 70: One day, you can have something like `jax.NDArray[np.bool_] | bool`

jakevdp (approved):
Looks good, thanks!

PR #9236, opened 2022-01-19
checkify: fix and test post_process_call/map

mattjj:
This PR also wrapped some lines to 80chars, though those weren't necessary for correctness.

The only action is in `CheckifyTrace.post_process_call` and `post_process_map` and the corresponding tests. I think we just never tested them!


tests/checkify_test.py
```diff
@@ -400,6 +401,28 @@ def multi_errors(x):
     self.assertIsNotNone(err.get())
     self.assertStartsWith(err.get(), expected_error)
 
+  def test_post_process_call(self):
+    @partial(checkify.checkify, errors=checkify.float_errors)
+    def g(x):
+      @jax.jit
+      def f(y):
+        return jnp.sin(x * y)
+      return f(jnp.inf)
```
LenaMartens, line 410: for my understanding: what do we need to test a post_process_call? Looks like a `checkify(jit(f))` with a closed over value maybe?

LenaMartens (approved):

tests/checkify_test.py
mattjj: Yeah exactly: post_process_call handles the case where we have Tracers on a jitted function's output which correspond to Traces not present on the input. (If we had corresponding Tracers on the input then we'd have hit process_call instead.) It's like "uh oh, this output has Tracers on it that we didn't anticipate given the Tracers on the input!"

PR #9239, opened 2022-01-19
Eliminate redundancies in jax.scipy.linalg.expm

### LINKED ISSUE #5647: Redundant computation in expm?
Hi all:

It looks to me there are redundant calculations in the `expm` function, but wanted to ask to see if I'm missing something (perhaps something clever is happening in compilation). 

E.g., for `float64` and `complex128`, the numerator/denominator for the Pade approximant is generated as follows:

```
if A.dtype == 'float64' or A.dtype == 'complex128':
   U3, V3 = _pade3(A)
   U5, V5 = _pade5(A)
   U7, V7 = _pade7(A)
   U9, V9 = _pade9(A)
   maxnorm = 5.371920351148152
   n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
   A = A / 2**n_squarings
   U13, V13 = _pade13(A)
   conds=jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
                    9.504178996162932e-001, 2.097847961257068e+000])
   U = jnp.select((A_L1<conds), (U3, U5, U7, U9), U13)
   V = jnp.select((A_L1<conds), (V3, V5, V7, V9), V13)
```

The algorithm being used (I believe by Al-Mohy and Higham, same as in `scipy`) prescribes choosing between 1 of 5 different Pade approximations (of various orders) based on the norm of `A`. In the above, it appears all potential Pade approximations are being computed, and only afterwards is one chosen based on the norm of `A`. Hence, the computations for the others are discarded and ultimately unnecessary.

It seems this function could benefit from changing the above logic to use nested `lax.cond` statements (or perhaps `switch`?) to avoid computing all possible versions. To motivate this: I quickly tried modifying the `expm` function to simply always do the `U13, V13` version of of the computation (the most expensive) and not compute any of the lower order versions, and observed roughly a factor of 2 speed increase in compiled execution. My speed tests weren't rigorous - I didn't test it for a variety of norms - but I believe this should speed things up in all cases (again, unless I'm missing something with lax behaviour). Edit: This is on CPU.

Thoughts?

P.s. In some of my use cases of jax, the most expensive piece is a ton of matrix exponentials, so I'm very interested in potential speed gains for this function :D.

DanPuzzuoli:
## Summary

This PR introduces a minor change to the `expm` function:
- Currently, given the matrix `A`, `expm` computes all Pade approximant numerators/denominators, then selects which one to use based on the norm of the matrix.
- This PR modifies this to only compute the relevant Pade approximant numerator/denominator by using `switch` conditioned on the norm of the matrix `A`.

Closes #5647

## Speed testing

I unfortunately don't currently have access to a GPU, but have run some CPU speed tests comparing the code in this PR to the current release. Testing in 64 bit mode with complex matrices of dimensions `[5, 10, 50, 100, 500, 1000, 5000]`, the ratios by dimension of "release run time" / "PR run time" for `expm` (once compiled) are:
```
[1.0794702 , 3.63815789, 3.41229117, 3.38029568, 1.68847097, 1.54355993, 1.61012301]
```
These are not the most rigorous (I'm just generating random matrices and measuring time by sandwiching `time()` calls around a single exponentiation for each dimension), but seem pretty reasonable based on the number of matrix multiplications this is potentially saving (which varies based on matrix norm).

I've attached the jupyter notebook I've run the tests in [jax_expm_speed_test.ipynb.zip](https://github.com/google/jax/files/7898241/jax_expm_speed_test.ipynb.zip), which include a cell of the saved list of execution times for JAX version (release and this PR). 

## Going forward

I'm trying to finally get GPU access in my org and will reply with GPU speed comparisons here when possible. In the meantime, please let me know if there's anything you'd like to see in terms of speed comparisons.

### DIFF
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -294,15 +294,15 @@ def _calc_P_Q(A):
    A = A / 2**n_squarings
    conds = jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
                       9.504178996162932e-001, 2.097847961257068e+000, jnp.inf])
-   idx = jnp.argmax(A_L1 < conds)
-   U, V = lax.switch(idx, [_pade3, _pade5, _pade7, _pade9, _pade13])
+   idx = jnp.argwhere(A_L1 < conds, size=1).squeeze()
+   U, V = lax.switch(idx, [_pade3, _pade5, _pade7, _pade9, _pade13], A)
   elif A.dtype == 'float32' or A.dtype == 'complex64':
     maxnorm = 3.925724783138660
     n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
     A = A / 2**n_squarings
     conds = jnp.array([4.258730016922831e-001, 1.880152677804762e+000, jnp.inf])
-    idx = jnp.argmax(A_L1 < conds)
-    U, V = lax.switch(idx, [_pade3, _pade5, _pade7])
+    idx = jnp.argwhere(A_L1 < conds, size=1).squeeze()
+    U, V = lax.switch(idx, [_pade3, _pade5, _pade7], A)
   else:
     raise TypeError("A.dtype={} is not supported.".format(A.dtype))
   P = U + V  # p_m(A) : numerator
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -293,15 +293,15 @@ def _calc_P_Q(A):
    n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
    A = A / 2**n_squarings
    conds = jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
-                      9.504178996162932e-001, 2.097847961257068e+000, jnp.inf])
-   idx = jnp.argwhere(A_L1 < conds, size=1).squeeze()
+                      9.504178996162932e-001, 2.097847961257068e+000])
+   idx = jnp.digitize(A_L1, conds)
    U, V = lax.switch(idx, [_pade3, _pade5, _pade7, _pade9, _pade13], A)
   elif A.dtype == 'float32' or A.dtype == 'complex64':
     maxnorm = 3.925724783138660
     n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
     A = A / 2**n_squarings
-    conds = jnp.array([4.258730016922831e-001, 1.880152677804762e+000, jnp.inf])
-    idx = jnp.argwhere(A_L1 < conds, size=1).squeeze()
+    conds = jnp.array([4.258730016922831e-001, 1.880152677804762e+000])
+    idx = jnp.digitize(A_L1, conds)
     U, V = lax.switch(idx, [_pade3, _pade5, _pade7], A)
   else:
     raise TypeError("A.dtype={} is not supported.".format(A.dtype))
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -289,28 +289,20 @@ def _calc_P_Q(A):
   A_L1 = np_linalg.norm(A,1)
   n_squarings = 0
   if A.dtype == 'float64' or A.dtype == 'complex128':
-   U3, V3 = _pade3(A)
-   U5, V5 = _pade5(A)
-   U7, V7 = _pade7(A)
-   U9, V9 = _pade9(A)
    maxnorm = 5.371920351148152
    n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
    A = A / 2**n_squarings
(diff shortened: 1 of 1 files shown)


DanPuzzuoli:
I've just signed the CLA, so please rescan!

jakevdp:
You could use a public Colab runtime to run benchmarks on GPU if you wish

jax/_src/scipy/linalg.py
```diff
@@ -289,28 +289,20 @@ def _calc_P_Q(A):
...
    maxnorm = 5.371920351148152
    n_squarings = jnp.maximum(0, jnp.floor(jnp.log2(A_L1 / maxnorm)))
    A = A / 2**n_squarings
-   U13, V13 = _pade13(A)
    conds = jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
-                      9.504178996162932e-001, 2.097847961257068e+000])
-   U = jnp.select((A_L1 < conds), (U3, U5, U7, U9), U13)
-   V = jnp.select((A_L1 < conds), (V3, V5, V7, V9), V13)
+                      9.504178996162932e-001, 2.097847961257068e+000, jnp.inf])
+   idx = jnp.argwhere(A_L1 < conds, size=1).squeeze()
```
soraros, line 297: Ah, this is your use case. Then you might also want to try jnp.digitize.

DanPuzzuoli: Nice yes that is cleaner.

soraros: But be aware that `jnp.digitize` is `searchsorted`-based, which could have different performance characteristics from the `nonzero`-based `jnp.argwhere`. I have absolutely no idea which one should/would be faster in this particular case, only through benchmark can one tell I guess.

DanPuzzuoli: Ah I see - thanks. My naive expectation is that the matrix operations would dominate in this function anyway, especially given how few options either `digitize` or `argwhere` need to choose from.

DanPuzzuoli:
> You could use a public Colab runtime to run benchmarks on GPU if you wish

Thanks for this suggestion. I've just run some comparisons in colab on GPU using the latest release version and the version in this PR. Using the same code to generate the random arrays and measuring the speed after an initial run of `expm` that includes compilation, for dimensions as before `[5, 10, 50, 100, 500, 1000, 5000]`, we get:
- Execution times for latest release version: 
```
[0.0013790130615234375, 0.0016705989837646484, 0.004743814468383789, 0.00931692123413086, 0.137678861618042,  0.7950971126556396, 100.97727608680725]
```
- Execution times for PR version:
```
[0.0011055469512939453, 0.0013234615325927734, 0.0035729408264160156, 0.00675511360168457, 0.08099985122680664, 0.5249743461608887, 67.66769027709961]
```
- Ratio of "release version" / "PR version"
```
[1.24735821, 1.26229508, 1.32770586, 1.37923976, 1.69974215, 1.5145447 , 1.49225244]
```

Again this is all done with 64-bit. I expect that this change will have less of an impact on 32-bit execution as there were fewer redundant calculations being done.

jakevdp:
Keep in mind that `searchsorted` is particularly inefficient on GPU, so your switch to `digitize` may have an outsized impact there.

DanPuzzuoli:
Just switched back to using `argwhere` and reran the tests, the times on GPU after initial compilation are:
```
[0.0011105537414550781,
 0.0013575553894042969,
 0.003557443618774414,
 0.00667119026184082,
 0.08334660530090332,
 0.5242531299591064,
 68.71966004371643]
```
which are essentially the same as before.

jax/_src/scipy/linalg.py
jakevdp: I think this would give the same result, and be more efficient than either `argwhere` or `digitize`:
```
idx = jnp.argmax(A_L1 < conds)
```

DanPuzzuoli: Okay so I've just run the following code:

```
conds = jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
                      9.504178996162932e-001, 2.097847961257068e+000, jnp.inf])
conds2 = jnp.array([1.495585217958292e-002, 2.539398330063230e-001,
                      9.504178996162932e-001, 2.097847961257068e+000])
```
and in different cells, for various values of `x`, I'm seeing:
```
%timeit jnp.argmax(x < conds).block_until_ready()
```
CPU (local): 178 microseconds
GPU (colab): 610 microseconds
```
%timeit jnp.argwhere(x < conds, size=1).squeeze().block_until_ready()
```
CPU (local): 1.89 ms
GPU (colab): 5.09 ms
```
%timeit jnp.digitize(x, conds2).block_until_ready()
```
CPU (local): 5.8 microseconds
GPU (colab): 270 microseconds

The above values were for `x=3.` but the speed rankings seem to be the same regardless of `x`. Seems like for this case `digitize` is the clear winner.

Going to revert this branch back to the digitize version - let me know if you think there are any issues with this.

jakevdp: OK, sounds good!

jakevdp (approved):

PR #9245, opened 2022-01-19
remove cpu platform setting in quickstart

### LINKED ISSUE #9244: Jax Quickstart tutorial forces platform to CPU
*Super* minor thing, but this entry at the start of the JAX Quickstart notebook seems like a weird addition, and makes the next few cells rather confusing for the new user:

https://github.com/google/jax/blob/eddea68a9aebe54e45899a73e127897df43607d6/docs/notebooks/quickstart.ipynb?short_path=d35eb80#L55-L58

```
# Prevent GPU/TPU warning.
import jax; jax.config.update('jax_platform_name', 'cpu')
```
Its only function seems to be to suppress the warning that appears in the next cell if the notebook is run on a machine with no GPU/TPU:

`WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)`

.... but it's doing so by simply forcing the entire notebook to run on the CPU regardless of architecture. That in turn turns the following statements into a lie:

image

(i.e. the Numpy array is reproducibly slightly faster than the JAX one).

Anyway, I think it would be sensible to remove this cell entirely and perhaps add a note before the next one to the effect that the user will see a warning if they don't have a compatible GPU available, but the code will still work (just a bit more slowly).

mattjj:
fixes #9244

undoes #6806

### DIFF
--- a/docs/notebooks/quickstart.ipynb
+++ b/docs/notebooks/quickstart.ipynb
@@ -44,20 +44,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "tags": [
-     "remove-cell"
-    ]
-   },
-   "outputs": [],
-   "source": [
-    "# Prevent GPU/TPU warning.\n",
-    "import jax; jax.config.update('jax_platform_name', 'cpu')"
-   ]
-  },
-  {
    "cell_type": "markdown",
    "metadata": {
     "id": "FQ89jHCYfhpg"
--- a/docs/notebooks/quickstart.md
+++ b/docs/notebooks/quickstart.md
@@ -45,13 +45,6 @@ from jax import grad, jit, vmap
 from jax import random
 ```
 
-```{code-cell} ipython3
-:tags: [remove-cell]
-
-# Prevent GPU/TPU warning.
-import jax; jax.config.update('jax_platform_name', 'cpu')
-```
-
 +++ {"id": "FQ89jHCYfhpg"}
 
 ## Multiplying Matrices


froystig (approved):
Could we also suppress the warning, and would we want to? Merging this as is would be good either way.

jakevdp:
Note that as a result of this change, the generated docs will include the GPU/TPU warning: https://jax--9245.org.readthedocs.build/en/9245/notebooks/quickstart.html#multiplying-matrices

Getting rid of that is the reason I added the line in the first place.

jakevdp:
We could always go back to my original approach to suppressing this, which was a hidden cell that printed `2 * jnp.arange(10)` or something like that, though it looks a bit strange when someone views the notebook directly.

froystig:
I'm not sure that it's undesirable to display the warning, especially considering that timing results later in the notebook can be surprising on CPU (see comments in #9244). We can add a remark about it to reduce confusion, but it seems to actually be serving its intended purpose here on the whole.

Responding to my own comment above, I wouldn't try to suppress the warning for this reason.

jakevdp:
OK, sounds good.

There are other places in the docs where I've done the same thing to reduce distractions in our HTML documentation; if anyone feels strongly about this they can grep to find those other cases.

mattjj:
Yeah, I think having the real content be accurate and clear (e.g. the part about executing on GPU, as also mentioned in #9244) is more important than whether there's a warning at the top.For the other docs, it depends on whether they have

jakevdp:
Our other option here would be to avoid running `quickstart` as part of the doc build: we could run it on GPU and then have the outputs included as-is in the generated documentation. The up-side would be that timings would actually reflect what's happening on the GPU. Down-side would be that we lose doctest CI for the content, and that outputs may become stale.

PR #9258, opened 2022-01-20
Add more documentation for buffer donation

### LINKED ISSUE #9237: Add more documentation for how argument donation works
A user was asking what is the meaning of the warning
```
UserWarning: Some donated buffers were not usable: ShapedArray(float32[185,16,16,3]), ShapedArray(int32[185])
```

It seems that the documentation does not say much about argument donation, except what is written in the docstring for `jit` and `pmap`:

```
    donate_argnums: Specify which arguments are "donated" to the computation.
      It is safe to donate arguments if you no longer need them once the
      computation has finished. In some cases XLA can make use of donated
      buffers to reduce the amount of memory needed to perform a computation,
      for example recycling one of your input buffers to store a result. You
      should not reuse buffers that you donate to a computation, JAX will raise
      an error if you try to.
```

gnecula:
Fixes: #9237

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -366,6 +366,73 @@ Interestingly, ``jax.grad(divide)(3., 2.)``, works because :func:`jax.grad`
 uses concrete tracers, and resolves the conditional using the concrete
 value of ``y``.
 
+.. _faq-donation:
+
+Buffer donation
+---------------
+
+(This feature is implemented only for TPU and GPU.)
+
+When JAX executes a computation it reserves buffers on the device for all inputs and outputs.
+If you know than one of the inputs is not needed after the computation, and if it
+matches the shape and element type of one of the outputs, you can specify that you
+want the corresponding input buffer to be donated to hold an output. This will reduce
+the memory required for the execution by the size of the donated buffer.
+
+You achieve this by using the `donate_argnums` parameter to the functions :func:`jax.jit`,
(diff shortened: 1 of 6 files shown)


docs/faq.rst
```diff
@@ -366,6 +366,71 @@ Interestingly, ``jax.grad(divide)(3., 2.)``, works because :func:`jax.grad`
...
+
+You achieve this by using the `donate_argnums` parameter to the functions :func:`jax.jit`,
+:func:`jax.pjit`, and :func:`jax.pmap`. This parameter is a sequence of indices (0 based) into
+the positional argument list::
+
+    def add(x, y):
+      return x + y
+
+   x = jax.device_put(np.ones((2, 3))
+   y = jax.device_put(np.ones((2, 3))
```
jakevdp, line 388: These two lines are missing a closing paren

docs/faq.rst
```diff
@@ -366,6 +366,71 @@ Interestingly, ``jax.grad(divide)(3., 2.)``, works because :func:`jax.grad`
 uses concrete tracers, and resolves the conditional using the concrete
 value of ``y``.
 
+.. _faq-donation:
+
+Buffer donation
+---------------
+
```
jakevdp, line 373: Should we mention more prominently here (at the top) that buffer donation is not implemented on CPU?

gnecula: Good idea, I added a short sentence at the start.

PR #9271, opened 2022-01-20
jnp.[nan]argmin/max: implement keepdims

jakevdp:
Numpy added this in version 1.22, so I gated the tests against this version.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -30,7 +30,7 @@ import collections
 from functools import partial
 import operator
 import types
-from typing import Sequence, FrozenSet, Optional, Tuple, Union, Set, Type, Callable
+from typing import Any, Sequence, FrozenSet, Optional, Tuple, Union, Set, Type, Callable
 from textwrap import dedent as _dedent
 import warnings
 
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -5222,70 +5222,82 @@ def argwhere(a, *, size=None, fill_value=None):
...
 """
 
-@_wraps(np.nanargmax, lax_description=_NANARG_DOC.format("max"))
-def nanargmax(a, axis: Optional[int] = None):
-  return _nanargmax(a, None if axis is None else operator.index(axis))
+@_wraps(np.nanargmax, lax_description=_NANARG_DOC.format("max"), skip_params=['out'])
+def nanargmax(a, axis: Optional[int] = None, out : Any = None, keepdims : Optional[bool] = None):
+  if out is not None:
+    raise NotImplementedError("The 'out' argument to jnp.nanargmax is not supported.")
+  return _nanargmax(a, None if axis is None else operator.index(axis), keepdims=bool(keepdims))
```
soraros, line 5274: OOC, why is this `keepdims` wrapped in `bool` and not the other ones?
Not super important, but wrapping them in `bool` seems to be a good thing since it saves a jit when the outer function is called with `f(..., keepdims=False)`.

jakevdp: I added these to appease mypy... the other private methods are not annotated so it wasn't flagged.

jakevdp: I wrapped the others in bool as well!

mattjj (approved):

PR #9301, opened 2022-01-24
disable implicit rank promotion for api_test

jakevdp:
Part of #7208

I had to move one test into its own test class, because setting the rank promotion flag via the decorators was causing test failures (probably due to some interaction between those contexts and the direct modification of the flag used in the test). I chose to avoid the problem rather than debugging the cause of that.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -69,6 +69,7 @@ python_version = (sys.version_info[0], sys.version_info[1])
 numpy_version = tuple(map(int, np.__version__.split('.')[:3]))
 
 
+@jtu.with_config(jax_numpy_rank_promotion="raise")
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):
The change looks fine, but I'm wondering if the fact you had to move the test means something is suboptimal with the `with_config` decorator. Ideally shouldn't we reset that state on every test case?

jakevdp:
Yeah, something strange is definitely happening. Even if you decorate the test case with `@jax.numpy_rank_promotion('allow')` (which should be a no-op) the result is that the test fails.

I think it has to do with this test manually updating the `FLAGS`, but I'm not sure.

jakevdp:
Tracking the config issue in #9304

PR #9309, opened 2022-01-25
linalg_test: disable implicit rank promotion

jakevdp:
Related to #7208

In this case, rather than manually reshaping arrays to avoid rank promotion errors, I locally disabled the rank promotion instead, as I judged that manual reshapes obscured the logic too much (happy to go the other way if anyone feels strongly).

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -321,7 +321,8 @@ def cholesky_jvp_rule(primals, tangents):
   # Forward-mode rule from https://arxiv.org/pdf/1602.07527.pdf
   def phi(X):
     l = jnp.tril(X)
-    return l / (jnp._constant_like(X, 1) + jnp.eye(X.shape[-1], dtype=X.dtype))
+    return l / lax.expand_dims(jnp._constant_like(X, 1) + jnp.eye(X.shape[-1], dtype=X.dtype),
+                               range(l.ndim - 2))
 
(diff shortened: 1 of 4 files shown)


jax/_src/lax/linalg.py
```diff
@@ -321,7 +322,8 @@ def cholesky_jvp_rule(primals, tangents):
   # Forward-mode rule from https://arxiv.org/pdf/1602.07527.pdf
   def phi(X):
     l = jnp.tril(X)
-    return l / (jnp._constant_like(X, 1) + jnp.eye(X.shape[-1], dtype=X.dtype))
+    with jax.numpy_rank_promotion('allow'):
```
hawkinsp, line 325: Is it better to add allow-promotion scopes, or just to make the promotion explicit? Not sure how painful the latter would be.

jakevdp: I went back and forth on that – in most of these cases though I judged that doing it manually obscures the logic. For example, here, the manual approach would be something like
```python
return l / (jnp._constant_like(X, 1) + lax.expand_dims(jnp.eye(X.shape[-1], dtype=X.dtype), range(l.ndim - 2)))
```
And in the cases further below, there are multiple such casts that are required.

jakevdp: I gave it a shot - making promotion explicit isn't too bad. Thanks for the nudge!

hawkinsp (approved):
This looks better to me, thanks!

PR #9312, opened 2022-01-25
fix caching example in jit tutorial

### LINKED ISSUE #9288: unjitted_loop_body is not constantly re-compiled in the JIT tutorial
Hello there,

I could be misunderstanding something here(new to Jax still) but on the [JIT tutorial](https://jax.readthedocs.io/en/latest/jax-101/02-jitting.html#caching) it seems like the "bad" example is not actually that bad. 

More concretely, the given example is as follows:
```python
def g_inner_jitted_poorly(x, n):
  i = 0
  while i < n:
    # Don't do this!
    i = jax.jit(unjitted_loop_body)(i)
  return x + i
```
Contrary to the description of this example on the tutorial, it seems like the  unjitted_loop_body function is actually not being compiled every iteration, but cached somewhere. 
```python
def unjitted_loop_body(prev_i):
  print("i am being jitted")
  return prev_i + 1

def g_inner_jitted_poorly(x, n):
  i = 0
  while i < n:
    # Don't do this!
    i = jax.jit(unjitted_loop_body)(i)
  return x + i
g_inner_jitted_poorly(10, 3)
```
outputs
```
i am being jitted
```
However, if I switch this to a lambda function, it is indeed being compiled every iteration. Which seems like a more accurate example(if I'm not mistaken).

```python
def unjitted_loop_body(prev_i):
  print("i am being jitted")
  return prev_i + 1

def g_inner_jitted_poorly(x, n):
  i = 0
  while i < n:
    # Don't do this!
    i = jax.jit(lambda x: unjitted_loop_body(x))(i)
  return x + i
g_inner_jitted_poorly(10, 3)
```
outputs 
```
i am being jitted
i am being jitted
i am being jitted
```

Thanks!

GeraldCSC:
closes #9288 

This pull request fixes the problem as discussed in the above issue. To summarise, the caching in JAX can handle the original example, however it is not able to cache things like partial(f), lambda x: f(x) when calling `jax.jit` inside the loop. Which I've updated the paragraph for and showed the performance degradation. 

I ran the notebook on colab, and then ran `jupytext --sync` on the notebook.

Happy to make any modifications!

### DIFF
--- a/docs/jax-101/02-jitting.ipynb
+++ b/docs/jax-101/02-jitting.ipynb
@@ -28,17 +28,17 @@
    "execution_count": 1,
    "metadata": {
     "id": "P9Xj77Wx3Z2P",
-    "outputId": "77d28eec-bfbb-487b-b10f-e47bc8f7dd0d"
+    "outputId": "5a0597eb-86c9-4762-ce10-2811debbc732"
    },
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "{ lambda  ; a.\n",
-      "  let b = log a\n",
-      "      c = log 2.0\n",
-      "      d = div b c\n",
+      "{ lambda ; a:f32[]. let\n",
+      "    b:f32[] = log a\n",
+      "    c:f32[] = log 2.0\n",
+      "    d:f32[] = div b c\n",
       "  in (d,) }\n"
      ]
     }
@@ -80,7 +80,7 @@
    "execution_count": 2,
    "metadata": {
     "id": "JxV2p7e2RawC",
-    "outputId": "af7472de-df57-412f-83cb-e468c4b02d64"
+    "outputId": "9dfe8a56-e553-4640-a04e-5405aea7832d"
    },
    "outputs": [
     {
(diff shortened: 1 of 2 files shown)


jakevdp (commented):
Looks great! I love the examples you got together.

A couple things before we can merge:

- Your jupytext version does not match the one used by the CI. You can either manually install the right version, or you can run it via `pre-commit`, which will ensure the correct version: https://jax.readthedocs.io/en/latest/developer.html#syncing-notebooks
- The readthedocs failure is due to an error message that uses back-ticks, which the lexer is unable to parse. We need to remove those from our messages... but in the meantime you could land this by manually removing the back-ticks from the traceback in the ipynb source (sorry about that)
- Once you make these fixes, please squash the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Thanks!

GeraldCSC:
thanks for the feedback! I've incorporated the changes. Let me know if there's anything I can edit/add.

jakevdp (approved):
Looks good, thanks!

PR #9313, opened 2022-01-25
[WIP] migrate from pocketfft to ducc.fft

mreineck:
Related to #7490

This is still at a very early stage:
 - ~~one FFT test fails with a segfault (I'm looking into that)~~
 - I probably have to add BSD licensing headers to some of the ducc source files
 - if multi-threading should be used, we have to adjust the interface accordingly
 - I haven't done any proper formatting for the source changes in JAX

However I'm happy about any feedback already at this stage!

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,10 +31,8 @@ build --copt=-DMLIR_PYTHON_PACKAGE_PREFIX=jaxlib.mlir.
 
 ###########################################################################
 
-build:posix --copt=-fvisibility=hidden
-build:posix --copt=-Wno-sign-compare
-build:posix --cxxopt=-std=c++14
-build:posix --host_cxxopt=-std=c++14
+build:posix --cxxopt=-std=c++17
+build:posix --host_cxxopt=-std=c++17
 
 build:avx_posix --copt=-mavx
 build:avx_posix --host_copt=-mavx
@@ -75,8 +73,8 @@ build:windows --host_copt=-DNOGDI
 # https://devblogs.microsoft.com/cppblog/announcing-full-support-for-a-c-c-conformant-preprocessor-in-msvc/
 # otherwise, there will be some compiling error due to preprocessing.
 build:windows --copt=/Zc:preprocessor
-build:windows --cxxopt=/std:c++14
-build:windows --host_cxxopt=/std:c++14
+build:windows --cxxopt=/std:c++17
+build:windows --host_cxxopt=/std:c++17
 # Generate PDB files, to generate useful PDBs, in opt compilation_mode
 # --copt /Z7 is needed.
 build:windows --linkopt=/DEBUG
--- a/build/LICENSE.txt
+++ b/build/LICENSE.txt
@@ -4332,8 +4332,8 @@ Copyright 2019 The TensorFlow Authors.  All rights reserved.
    limitations under the License.
 
 --------------------------------------------------------------------------------
-License for pocketfft:
-Copyright (C) 2010-2018 Max-Planck-Society
+License for the FFT components of ducc0:
+Copyright (C) 2010-2022 Max-Planck-Society
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
--- a/jaxlib/pocketfft_kernels.cc
+++ b/jaxlib/pocketfft_kernels.cc
@@ -16,71 +16,106 @@ limitations under the License.
 #include <complex>
 
 #include "flatbuffers/flatbuffers.h"
-#include "pocketfft/pocketfft_hdronly.h"
+#include "pocketfft/src/ducc0/fft/fft.h"
 #include "jaxlib/pocketfft_generated.h"
 
 namespace jax {
 
+using shape_t = ducc0::fmav_info::shape_t;
+using stride_t = ducc0::fmav_info::stride_t;
+
+void fixstrides(stride_t &str, size_t size)
+  {
+  ptrdiff_t ssize = ptrdiff_t(size);
+  for (auto &s: str)
+    {
+    auto tmp = s/ssize;
+    if (tmp*ssize != s) throw "Bad stride";
+    s = tmp;
+    }
+  }
+
 void PocketFft(void* out, void** in) {
   const PocketFftDescriptor* descriptor = GetPocketFftDescriptor(in[0]);
-  pocketfft::shape_t shape(descriptor->shape()->begin(),
-                           descriptor->shape()->end());
-  pocketfft::stride_t stride_in(descriptor->strides_in()->begin(),
-                                descriptor->strides_in()->end());
-  pocketfft::stride_t stride_out(descriptor->strides_out()->begin(),
-                                 descriptor->strides_out()->end());
-  pocketfft::shape_t axes(descriptor->axes()->begin(),
-                          descriptor->axes()->end());
+  shape_t shape(descriptor->shape()->begin(),
+                descriptor->shape()->end());
+  stride_t stride_in(descriptor->strides_in()->begin(),
+                     descriptor->strides_in()->end());
+  stride_t stride_out(descriptor->strides_out()->begin(),
+                      descriptor->strides_out()->end());
+  shape_t axes(descriptor->axes()->begin(),
+               descriptor->axes()->end());
 
   switch (descriptor->fft_type()) {
     case PocketFftType_C2C:
       if (descriptor->dtype() == PocketFftDtype_COMPLEX64) {
-        std::complex<float>* a_in =
-            reinterpret_cast<std::complex<float>*>(in[1]);
-        std::complex<float>* a_out =
-            reinterpret_cast<std::complex<float>*>(out);
-        pocketfft::c2c(shape, stride_in, stride_out, axes,
-                       descriptor->forward(), a_in, a_out,
-                       static_cast<float>(descriptor->scale()));
+        fixstrides(stride_in, sizeof(std::complex<float>));
+        fixstrides(stride_out, sizeof(std::complex<float>));
+        ducc0::cfmav<std::complex<float>> m_in(reinterpret_cast<std::complex<float>*>(in[1]),
+          shape, stride_in);
+        ducc0::vfmav<std::complex<float>> m_out(reinterpret_cast<std::complex<float>*>(out),
+          shape, stride_out);
+        ducc0::c2c(m_in, m_out, axes, descriptor->forward(),
+                   static_cast<float>(descriptor->scale()));
       } else {
-        std::complex<double>* a_in =
-            reinterpret_cast<std::complex<double>*>(in[1]);
-        std::complex<double>* a_out =
-            reinterpret_cast<std::complex<double>*>(out);
-        pocketfft::c2c(shape, stride_in, stride_out, axes,
-                       descriptor->forward(), a_in, a_out, descriptor->scale());
+        fixstrides(stride_in, sizeof(std::complex<double>));
+        fixstrides(stride_out, sizeof(std::complex<double>));
+        ducc0::cfmav<std::complex<double>> m_in(reinterpret_cast<std::complex<double>*>(in[1]),
+          shape, stride_in);
+        ducc0::vfmav<std::complex<double>> m_out(reinterpret_cast<std::complex<double>*>(out),
+          shape, stride_out);
+        ducc0::c2c(m_in, m_out, axes, descriptor->forward(),
+                   static_cast<double>(descriptor->scale()));
       }
       break;
     case PocketFftType_C2R:
       if (descriptor->dtype() == PocketFftDtype_COMPLEX64) {
-        std::complex<float>* a_in =
-            reinterpret_cast<std::complex<float>*>(in[1]);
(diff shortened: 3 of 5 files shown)


mreineck:
I realized that the `copts` in `third_party/pocketfft/BUILD.bazel` most likely won't achieve anything, since most of the code isin headers and will be compiled as part of  `jaxlib/pocketfft_kernels.cc`. So we should probably adjust the compiler options for this file...

mreineck:
Technical question: the CLA only covers the actual content of this PR, not the library sources themselves, correct? If so, I probably won't have to discuss this with a lawyer ;-)

mreineck:
There are two failures in `lax_vmap_test.py` where the results are too inaccurate. I don't understand the tests well enough to judge whether this is really a problem.

hawkinsp:
Thanks! I'll look at the technical aspects of this shortly.

> Technical question: the CLA only covers the actual content of this PR, not the library sources themselves, correct? If so, I probably won't have to discuss this with a lawyer ;-)

I don't speak for Google's lawyers, but yes, my understanding is that the CLA applies only to your contributions, e.g., PRs.

As I mentioned before, we don't want to incorporate GPL-licensed code because it would require us to distribute JAX itself under a much more restrictive license than its current license (Apache 2). (Obviously third parties are welcome to plug in GPL code: it's just not something we ourselves want to distribute.)

To get this merged, we'd need ducc0's LICENSE information to be explicit about which files are under the BSD license and where the GPL license does not apply, i.e., a clear statement from you as the author in the project's license information that the GPL license doesn't apply to any of the files needed for FFT transformations.

Currently it looks like, for example, in ducc0 `fft/fft.h` is marked as BSD-licensed, but some of its dependencies (e.g., in `infra/`) are GPL-licensed and marked as such. We'd need everything in the transitive closure of the FFT library to fall under the more liberal license (e.g.., a BSD license, see https://opensource.google/docs/thirdparty/licenses/#notice) and for that to be clearly stated. Does that seem possible?

mreineck:
> Currently it looks like, for example, in ducc0 fft/fft.h is marked as BSD-licensed, but some of its dependencies (e.g., in infra/) are GPL-licensed and marked as such. We'd need everything in the transitive closure of the FFT library to fall under the more liberal license (e.g.., a BSD license, see https://opensource.google/docs/thirdparty/licenses/#notice) and for that to be clearly stated. Does that seem possible?

Absolutely, that's the plan. As I mentioned above, I just didn't get around to add the change to all affected files. Will do that soon.

third_party/pocketfft/workspace.bzl
```diff
@@ -19,11 +19,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 def repo():
     http_archive(
         name = "pocketfft",
```
hawkinsp, line 21: Would it make sense to add ducc0 as a parallel dependency to pocketfft rather than replacing pocketfft for the moment?

I'm concerned that it might take a while to enable C++17 by default, but I don't want that to hold up your PR. One way we could manage that would be to add ducc0 as an alternative to pocketfft and select between the two depending on whether a C++17 compiler is available.

third_party/pocketfft/BUILD.bazel
```diff
@@ -4,8 +4,21 @@ package(default_visibility = ["//visibility:public"])
 
 cc_library(
     name = "pocketfft",
-    hdrs = ["pocketfft_hdronly.h"],
-    copts = ["-fexceptions"],
```
hawkinsp, line 8: This would need to be reformatted, but I can do that as part of the merge.
(We just apply this tool: https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md if you're curious.)

third_party/pocketfft/BUILD.bazel
```diff
@@ -4,8 +4,21 @@ package(default_visibility = ["//visibility:public"])
...
+            "src/ducc0/infra/error_handling.h",
+            "src/ducc0/infra/mav.h",
+            "src/ducc0/infra/simd.h",
+          "src/ducc0/infra/threading.h",
+          "src/ducc0/infra/threading.cc",
+            "src/ducc0/infra/useful_macros.h",
+            "src/ducc0/math/cmplx.h",
+            "src/ducc0/math/unity_roots.h",
+],
+    copts = ["-fexceptions", "-ffast-math"],
```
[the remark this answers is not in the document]
mreineck: `-ffast-math`(or at least a subset of it) does indeed have a large impact, because without this flag the compiler is not allowed to use FMA instructions ... and FFTs consist of mostly FMAs. I know that there is a `#pragma` which is supposed to allow this, but `gcc` has a history of not dealing with this properly (I don't know about `clang`).
My idea would be to switch this on only for the  compilation of `jaxlib/pocketfft_kernels.cc`, if that's possible.

third_party/pocketfft/workspace.bzl
mreineck: I'm not in a hurry to have this merged; we can just wait until C++17 is acceptable. My main goal was to see how big the necessary adjustments will be (surprisingly small, it seems!).
I don't know how to add this as a parallel dependency, but if you help me do that, it's also perfectly fine with me.

third_party/pocketfft/BUILD.bazel
mreineck: Thanks, will have a look!

third_party/pocketfft/BUILD.bazel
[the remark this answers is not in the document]
mreineck: Maybe we can get away with `-ffp-contract=fast` I'd have to test that with several compilers though. In any case this is the most important component of `-ffast-math` for this code.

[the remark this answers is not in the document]
hawkinsp: Interesting.

For clang to use fma instructions I think you need two things:
a) to target an architecture with FMA instructions, and
b) to enable `fp-contract`: https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-contract
I assume gcc is similar.

(b) is no problem: we can just turn that flag on for this file: it's just a slightly more nuanced `-ffast-math`.

However, for (a) we don't distribute our Python wheels built for an architecture that has FMA instructions. We currently assume AVX but not AVX2 or FMA. So we'd want to start building for, say, Haswell machines, which introduced AVX2 and FMA, to get the full benefit.

(Function multi-versioning is technically possible but it's problematic in C++ code because it runs a risk of introducing [ODR problems](https://en.wikipedia.org/wiki/One_Definition_Rule).)

I think that's doable, but certainly even the AVX requirement has sometimes tripped people up! Apparently plenty of machines from older than 2013 are still out there!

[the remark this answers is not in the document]
mreineck: If I remember correctly, I also encountered some situations where `gcc` refused to vectorize certain expressions without `-ffast-math`; if that is true, it could have an effect on pure AVX architectures (without FMA) as well. But this is probably more a concern for the other `ducc` components like spherical harmonic transforms, and not so much for FFT.
Let's see how far we get without the sledgehammer `-ffast-math` :-)

mreineck:
I'm currently updating all source files required by the FFT functionality to include the following:

```
/* SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0-or-later */

/*
All rights reserved.

Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met:

* Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.
* Redistributions in binary form must reproduce the above copyright notice, this
  list of conditions and the following disclaimer in the documentation and/or
  other materials provided with the distribution.
* Neither the name of the copyright holder nor the names of its contributors may
  be used to endorse or promote products derived from this software without
  specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
*/

/*
 *  This code is free software; you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation; either version 2 of the License, or
 *  (at your option) any later version.
 *
 *  This code is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 *
 *  You should have received a copy of the GNU General Public License
 *  along with this code; if not, write to the Free Software
 *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
```

Would that be sufficiently clear?

I'm also updating the README.md file to mention that a subset of the ode is dual-licensed.
I'd rather not mess with the LICENSE file itself, since that is a verbatim copy of GPLv2. As far as I understand, BSD3-licensed code does not require an accompanying license file; just the licensing header in each source file should be enough.

hawkinsp:
I suspect that's fine, but for our use I'll need to get an ok from our compliance folks. But I think so long as it's clear and well-documented exactly what is covered by the BSD license they will be happy with it.

mreineck:
OK. In the extreme case, I could make a reduced package containing only the relevant files (not my favorite oprion, since it means more maintenance), or you can import the FFT subset into your sources directly. I don't expect frequent updates, roughly once per year or so.

.bazelrc
```diff
@@ -33,8 +33,8 @@ build --copt=-DMLIR_PYTHON_PACKAGE_PREFIX=jaxlib.mlir.
 
 build:posix --copt=-fvisibility=hidden
 build:posix --copt=-Wno-sign-compare
```
hawkinsp, line 35: We can revert these changes: they are already present at head.

build/LICENSE.txt
```diff
@@ -4333,7 +4333,7 @@ Copyright 2019 The TensorFlow Authors.  All rights reserved.
 
 --------------------------------------------------------------------------------
 License for pocketfft:
```
hawkinsp, line 4335: ducc instead of pocketfft.

third_party/pocketfft/workspace.bzl
```diff
@@ -19,11 +19,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
-        sha256 = "66eda977b195965d27aeb9d74f46e0029a6a02e75fbbc47bb554aad68615a260",
-        strip_prefix = "pocketfft-f800d91ba695b6e19ae2687dd60366900b928002",
+        sha256 = "f8d0fbb90c56be4bb168fd9c64d5e62dfa0f89085d3fc21b7a9f7fa08a190603",
```
hawkinsp, line 22: We can bump this to a newer release.

jaxlib/pocketfft_kernels.cc
```diff
@@ -16,71 +16,106 @@ limitations under the License.
 #include <complex>
```
hawkinsp, line 16: We should rename these files but it can wait until a subsequent PR.

third_party/pocketfft/BUILD.bazel
```diff
@@ -4,8 +4,21 @@ package(default_visibility = ["//visibility:public"])
...
     features = ["-use_header_modules"],
+    local_defines = [],
```
hawkinsp, line 21: Unused; remove.

hawkinsp (commented):
I think we can now merge this! I realize it's been a while, so I'm happy to do these fixes as part of merging the PR if you like.

third_party/pocketfft/BUILD.bazel
mreineck: Great to see this making progress! I think I made the necessary adjustments.

mreineck:
If I understood correctly, a Google account is required to sign the CLA. Since I don"t have one, I suggest that @Edenhofer re-submits the patch as his own. OK?

Edenhofer:
Since the CLA bot is still unhappy, I filed a new MR at https://github.com/google/jax/pull/12120 .

hawkinsp:
#12122 was merged! Thanks for your help!

PR #9314, opened 2022-01-25
Fix `max_squarings` in `jax.scipy.linalg.expm`

attila-i-szabo:
`max_squarings=16` is hard-coded in the upstream implementation of `jax.scipy.linalg.expm` at the point where the squarings are actually done. This PR fixes the issue by passing `max_squarings` as a non-jitting argument to `_squaring`.

It is hard to give a discriminative test for this issue because it's only relevant if you need more than 16 squarings, at which point you probably get an overflow.

### DIFF
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -275,7 +275,7 @@ def expm(A, *, upper_triangular=False, max_squarings=16):
   def _compute(args):
     A, P, Q = args
     R = _solve_P_Q(P, Q, upper_triangular)
-    R = _squaring(R, n_squarings)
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks for the contribution! If you could complete the CLA, then we'll be able to proceed with review. Thanks!

attila-i-szabo:
I've completed it just after making the PR, I guess the test didn't check for it again...

jakevdp:
OK, looks like the CLA is good now

jakevdp (approved):
Thanks - I think this looks like a reasonable change. Looking at the history of this function, it seems like threading the value through was overlooked when `max_squarings` was added in #4314

PR #9316, opened 2022-01-25
djax: let make_jaxpr build dyn shape jaxprs

mattjj:
Also make small adaptations to the numpy and lax layers so that we can stage out a shape-polymorphic mnist.

TODO:
* [x] flag because we want to gate whether `jax.make_jaxpr(lambda n: jnp.zeros(n))(3)` is an error (after this change it need not be, but we have some tests which check that this is a trace-time error, and trace-time errors can be nicer for users than lowering-time errors, so we need to figure out what to do here. We'll use a flag to unblock the PR though.)

Known issues:
* [ ] for dynamic (rather than just polymorphic) shapes, we need to be able to return arrays with let-bound axis sizes (splatted as a dependent tuple and type annotation)
* [x] with nested jits, closed-over tracers don't get enough substitution into their avals (i.e. avals may contain tracers which also need to be lifted)

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -32,7 +32,7 @@ import threading
 import weakref
 import types
 from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
-                    Sequence, Tuple, TypeVar, Union, overload)
+                    Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable)
 from warnings import warn
 
 import numpy as np
@@ -53,7 +53,8 @@ from jax._src.traceback_util import api_boundary
 from jax.tree_util import (tree_map, tree_flatten, tree_unflatten,
                            tree_structure, tree_transpose, tree_leaves,
                            tree_multimap, treedef_is_leaf, treedef_children,
-                           Partial, PyTreeDef)
+                           Partial, PyTreeDef, all_leaves)
+from jax._src.tree_util import broadcast_prefix
 from jax._src.util import (unzip2, curry, safe_map, safe_zip, prod, split_list,
                            extend_name_stack, wrap_name, cache, wraps,
                            HashableFunction)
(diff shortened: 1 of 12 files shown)


dougalm (approved):
Looks great. Can we add some tests (even currently failing ones) for closing over and passing in shapes with repeated variable names? We should probably also have tests that check we correctly throw an error when dynamic shapes don't match.

mattjj:
> closing over and passing in shapes with repeated variable names

added these:

`test_basic_staging_repeated` builds this jaxpr:
```
{ lambda ; a:i32[] b:i32[a] c:f32[a,a]. let  in (b,) }
```

`test_basic_staging_multiple_shape_vars` builds this jaxpr:
```
{ lambda ; a:i32[] b: i32[] c:i32[a] d:f32[b,a]. let  in (c,) }
```

`test_closing_over_repeated_shapes` builds this jaxpr:
```
{ lambda ; a:i32[]. let
    b:i32[] = mul a 2
    c:f32[b,b] = broadcast_in_dim[broadcast_dimensions=() shape=(None, None)] 0.0
      b b
    d:f32[b] = xla_call[
      call_jaxpr={ lambda ; e:i32[] f:f32[e,e]. let
          g:f32[e] = reduce_sum[axes=(0,)] f
        in (g,) }
      name=<lambda>
    ] b c
  in (d,) }
```

`test_staging_repeated_nested` builds this jaxpr:
```
{ lambda ; a:i32[]. let
    b:i32[] = mul a 2
    c:f32[b,a] = broadcast_in_dim[broadcast_dimensions=() shape=(None, None)] 0.0
      b a
    d:f32[b] = broadcast_in_dim[broadcast_dimensions=() shape=(None,)] 1.0 b
    e:f32[b] = xla_call[
      call_jaxpr={ lambda ; f:i32[] g:i32[] h:f32[f,g] i:f32[f]. let
          j:f32[f] = reduce_sum[axes=(1,)] h
          k:f32[f] = add j i
        in (k,) }
      name=<lambda>
    ] b a c d
  in (e,) }
```

> We should probably also have tests that check we correctly throw an error when dynamic shapes don't match

added `test_shape_errors_var_and_lit` and `test_shape_errors_distinct_vars`

PR #9331, opened 2022-01-26
Update `get_compile_options` to accept Device objects.

pschuh:


jax/_src/lib/xla_bridge.py
```diff
@@ -124,6 +124,9 @@ def get_compile_options(
       msg = 'device_assignment does not match num_partitions: {} vs {}.'
       raise ValueError(msg.format(device_assignment, num_partitions))
 
+    if device_assignment.dtype == object:
```
hawkinsp, line 127: Could we make this unconditional and require an 2d sequence of objects, fixing up any remaining callers?

Presumably the end state we want to get here is the C++ runtime receives a 2D array of device objects rather than IDs, although that's not necessary for this PR. i.e., everywhere we talk about devices, we do so by object.

jax/interpreters/pxla.py
```diff
@@ -2113,7 +2112,7 @@ def from_hlo(name: str,
       num_replicas, num_partitions = 1, mesh.size
     else:
       num_replicas, num_partitions = mesh.size, 1
-    device_assignment = mesh.device_ids.reshape((num_replicas, num_partitions))
```
hawkinsp, line 2116: I wonder if we could even just drop the `device_ids` method from mesh. It seems to be mostly used in the `__repr__` method. (I may be wrong about that.)

hawkinsp (approved):

jax/interpreters/pxla.py
pschuh: It looks like there were some usage of `device_ids` in the context of `axis_index_groups`.

jax/_src/lib/xla_bridge.py
pschuh: Sounds good, but note that it would become impossible to get_compile_options without a physical device available.

hawkinsp: Reflecting on this: I think that's fine. If you don't have physical devices, you should not need a device *assignment* to compile: i.e. which specific devices you want to use. You need some abstraction of what that device topology looks like.

Issue #9337, opened 2022-01-26
Cannot import jax.experimental.pjit in Colab with version 0.2.27


LeoLaugier:
Hi, 

I came across the following behavior today: 

In Colab, today, the current default jax version is 0.2.25 and I can import jax.experimental.pjit

```from jax.experimental import pjit```

Though, if I update jax[tpu] to the 0.2.27 version with 

```!pip install jax[tpu]==0.2.27 -f https://storage.googleapis.com/jax-releases/libtpu_releases.html```

I get the following error when importing jax.experimental.pjit: 

      AttributeError                            Traceback (most recent call last)
      <ipython-input-3-7739b5b56126> in <module>()
      ----> 1 from jax.experimental import pjit
      
      1 frames
      /usr/local/lib/python3.7/dist-packages/jax/experimental/pjit.py in <module>()
           22 
           23 from jax.experimental import maps
      ---> 24 from jax.experimental.global_device_array import GlobalDeviceArray as GDA
           25 from jax import core
           26 from jax import linear_util as lu
      
      /usr/local/lib/python3.7/dist-packages/jax/experimental/global_device_array.py in <module>()
          431   return lambda bufs: GlobalDeviceArray(
          432       global_aval.shape, global_mesh, out_axis_resources, bufs, fast_path_args)
      --> 433 pxla.global_result_handlers[core.ShapedArray] = _gda_array_result_handler
          434 pxla.global_result_handlers[core.ConcreteArray] = _gda_array_result_handler
      
      AttributeError: module 'jax.interpreters.pxla' has no attribute 'global_result_handlers'

I'm not sure what it means, and I'll work with previous jax version but I thought it would be interesting to point this out.

Thanks,

Leo

**Comment by jakevdp** [comment]
`pip install jax[tpu]` is designed for Google Cloud TPU, and is not appropriate for Colab TPU runtimes which use an older TPU architecture.

To update JAX on Google Colab TPU, you can do the following:
```python
!pip install jax[cuda11_cudnn805]==0.2.27 -f https://storage.googleapis.com/jax-releases/jax_releases.html
```
Then you can setup the TPU and import `pjit` like this:
```python
import jax.tools.colab_tpu
jax.tools.colab_tpu.setup_tpu()

from jax.experimental import pjit
```

jakevdp:
Hopefully the doc update in #9338 will clarify this.

LeoLaugier:
Thank you, it solves the issue 👍

PR #9338, opened 2022-01-26
README: clarify installation instructions for Colab TPU

jakevdp:
This is currently confusing as written (cf. #9337)

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,11 +480,15 @@ the following in your cloud TPU VM:
 pip install --upgrade pip
 pip install "jax[tpu]>=0.2.16" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
 ```
+
+### pip installation: Colab TPU
 Colab TPU runtimes come with JAX pre-installed, but before importing JAX you must run the following code to initialize the TPU:
 ```python
 import jax.tools.colab_tpu
 jax.tools.colab_tpu.setup_tpu()
 ```
+Colab TPU runtimes use an older TPU architecture than Cloud TPU VMs, so installing `jax[tpu]` should be avoided on Colab.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -480,11 +480,15 @@ the following in your cloud TPU VM:
...
 pip install "jax[tpu]>=0.2.16" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
 ```
+
+### pip installation: Colab TPU
 Colab TPU runtimes come with JAX pre-installed, but before importing JAX you must run the following code to initialize the TPU:
 ```python
 import jax.tools.colab_tpu
 jax.tools.colab_tpu.setup_tpu()
 ```
+Colab TPU runtimes use an older architecture than Google Cloud TPU, so installing `jax[tpu]` should be avoided on Colab.
```
skye, line 490:
```suggestion
Colab TPU runtimes use an older architecture than Cloud TPU VMs, so installing `jax[tpu]` should be avoided on Colab.
```

README.md
```diff
@@ -480,11 +480,15 @@ the following in your cloud TPU VM:
...
+If for any reason you would like to update the JAX version on a Colab TPU runtime, follow the GPU-specific instructions above to install `jax[cuda11_cudnn805]`.
```
skye, line 491: Why install the GPU version on a TPU Colab? Shouldn't they just upgrade jax?

skye (changes requested):

README.md
jakevdp: I don't know whether it's necessary, but I recommended this because `cuda11_cudnn805` is the jaxlib version that comes pre-installed on Colab TPU runtimes.

jakevdp: I tried it, and the CPU wheels work on TPU. I think GPU wheels are used by default because it's a one-size-fits-all solution for all accelerators (it works on CPU, GPU, and TPU).

PR #9341, opened 2022-01-26
disable implicit rank promotion in a number of remaining tests

jakevdp:
Part of #7208 (Contains a number of changes extracted from #9330). Also needs #9346 for tests to pass.

After this PR, we should be able to globally disable rank promotion in `JaxTestCase` without any additional changes to the package.

### DIFF
--- a/examples/examples_test.py
+++ b/examples/examples_test.py
@@ -23,6 +23,7 @@ from absl.testing import parameterized
 
 import numpy as np
 
+import jax
 from jax import lax
 from jax import random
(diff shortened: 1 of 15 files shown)


jakevdp:
The failing tests (`examples_test`) don't explicitly disable rank promotion, and the errors don't reproduce when they are run alone. I think it's probably a state thing... my suspicion is that the issue fixed by #9339 may be the culprit.

hawkinsp (approved):

PR #9342, opened 2022-01-26
make jit(inline=True) do trace-time specializing

### LINKED ISSUE #9298: A case in which `jax.jit` changes results, prevents backprop, and inhibits optimisations
The summary is that I've found a case in which:
- adding a `jax.jit` decorator can change the functional behaviour of the program;
- adding a `jax.jit` decorator can prohibit reverse-mode autodifferentiation.
- adding a `jax.jit` decorator can inhibit the use of compile-time optimisations;

In all cases it is due to the same single root cause: a nested `jax.jit` unnecessarily converts concrete values to tracers.

---

First a MWE of the root cause:
```python
import jax
import jax.numpy as jnp

@jax.jit
def f(x):
    print(x)

@jax.jit
def g():
    f(1)

g()
```
This program prints `Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/2)>`: the `1` is unnecessarily converted into a tracer.

At first glance that probably seems reasonable. Why is this undesirable? Consider the following case.
```python
import jax
import jax.lax as lax
import jax.numpy as jnp

@jax.jit
def f(init, lower):
    def _body_fun(_, _y):
        return _y + 1
    return lax.fori_loop(lower, 1, _body_fun, init)

@jax.jit
def g():
    return jax.grad(f)(0.0, 0)

g()
```

**adding a `jax.jit` can change the functional behaviour of the program**

If `f` is decorated with `jax.jit` then the above program crashes.
If `f` is undecorated then the program runs.

**adding a `jax.jit` can inhibit reverse-mode autodifferentiation**

(Same as above:
If `f` is decorated with `jax.jit` then the above program crashes.
If `f` is undecorated then the program runs.)

**adding a `jax.jit` can inhibit the use of compile-time optimisations**

As a library author interested in optimising compile time and run times, then I'd like to specialise behaviour based on compile-time values.

In particular I have a case in which if a value is known at compile time then I can produce code that is efficient to compile. If it is only known at run time then the extra generality requires a more complicated program -- that produces the same functional results -- but is several orders of magnitude slower to compile.

(So morally speaking something a little similar to `lax.fori_loop` in JAX.)

---

Resolution: I think the resolution should be that within the dynamic context of a `jax.jit` decorator, then all other `jax.jit` decorators should be disabled.

C.f. #7155: sub-jit-functions are retraced within the dynamic context of a `jax.jit` decorator.

Overall, sub-jit-functions enter a weird state of being only "partially disabled" within the dynamic context of a `jax.jit` decorator: they are retraced, but values are still unnecessarily promoted to tracers.

mattjj:
Fixes #9298. See that discussion thread.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -285,7 +285,8 @@ def jit(
       should not reuse buffers that you donate to a computation, JAX will raise
       an error if you try to. By default, no argument buffers are donated.
 
-      For more details on buffer donation see the [FAQ](https://jax.readthedocs.io/en/latest/faq.html#buffer-donation).
+      For more details on buffer donation see the
+      [FAQ](https://jax.readthedocs.io/en/latest/faq.html#buffer-donation).
 
     inline: Specify whether this function should be inlined into enclosing
       jaxprs (rather than being represented as an application of the xla_call
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -1377,14 +1377,17 @@ class DynamicJaxprTrace(core.Trace):
     return out_tracers if primitive.multiple_results else out_tracers.pop()
 
   def process_call(self, call_primitive, f, tracers, params):
+    if params.get('inline', False):
+      frame = self.frame
+      args = [frame.constvar_to_val.get(frame.tracer_to_var.get(id(t)), t)
+              for t in tracers]
+      return f.call_wrapped(*args)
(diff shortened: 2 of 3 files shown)


patrick-kidger:
I don't completely grok all the implications, but I think the answer to "why does `jit(inline=True)` build a jaxpr", raised in #9298, is given by the final paragraph here: #6584

hawkinsp:
Closing as stale.

Issue #8996, opened 2021-12-16
Better `pjit in_axis_resources specification must be a tree prefix of the corresponding value` error


co-kuba:
When using pjit with large, complicated pytrees of arguments and partition specification, the runtime error `ValueError: pjit in_axis_resources specification must be a tree prefix of the corresponding value, got specification` is hard to work with, because it outputs the pspec pytree, and the value pytree (with asterisks) as huge single-line blobs.

An easy first step towards making these errors easier to work with would be to print out the pytrees like JSONs, with linebreaks and multiple indentation levels.

An more involved improvement would be to point out the difference between the two trees to the user in an easy-to-work-with way. For example, the smallest differing sub-pytrees could be printed, together with their locations in the main pytree (as list of keys/indices `(k_i)` that one needs to select in order, to traverse from the root down to the differing sub-pytree: `sub_pytree == root[k_1][...][k_n]`).

Issue #5047, opened 2020-11-30
Improve __str__/__repr__ of PyTreeDefs


j-towns:
Currently calling `print` on a pytree definition (the second output of `tree_util.tree_flatten`), produces something like:
```python
>>> from jax import tree_util
>>>
>>> _, treedef = tree_util.tree_flatten([1, 2, {'a': 3}])
>>> print(treedef)
PyTreeDef(list, [*,*,PyTreeDef(dict[['a']], [*])])
```
That's ok, but I think it's easier to parse if instead of nested PyTreeDefs we produce a Pythonic representation of the pytree, with variables replaced by `*`, i.e.
```python
PyTreeDef([*, *, {'a': *}])
```
It's pretty easy to implement a custom print function that does this in Python:
```python
def pretty_repr(treedef):
  class star(): __repr__ = lambda _: "*"
  return "PyTreeDef({})".format(treedef.unflatten(treedef.num_leaves * [star()]))
```
Could we make this the default representation?

hawkinsp:
Seems reasonable. The relevant code is here: https://github.com/tensorflow/tensorflow/blob/3b2b12fb53e1e9b1a52038e0ceaf0b24c1c1872b/tensorflow/compiler/xla/python/pytree.cc#L568

If you wanted to send a PR that would speed up the process!

j-towns:
How does one build using a local branch of TF/XLA?

hawkinsp:
I'd do this by:
* checkout `tensorflow` as a subdirectory of your `jax` source tree
* uncomment https://github.com/google/jax/blob/945017225a4bf64252349b346b485df6c231950f/WORKSPACE#L38 and comment out https://github.com/google/jax/blob/945017225a4bf64252349b346b485df6c231950f/WORKSPACE#L29
* make whatever changes you want to the tensorflow tree.
* build a jaxlib wheel (`python build/build.py`), which will use your TF tree to get XLA, and install it.
* profit!

NeilGirdhar:
[...]

PR #9372, opened 2022-01-29
add pytree utilities for better errors

mattjj:
This PR adds internal-only utility functions. My plan is to use these utilities to raise better pytree disagreement errors, especially for the `in_axes` and `in_axis_resources` arguments for functions like `vmap`, `pmap`, `xmap`, and `pjit` (see e.g. #8996 for `pjit`). But I wanted to get feedback on them and check them in first before making those changes. That is, none of these functions are actually used yet (except in tests).

I also included a `broadcast_prefix` function which I think will replace `api_util.flatten_axes`.

For example, here's a complex `vmap` where we might get a tree prefix error:

```python
full = lambda x: jnp.full((3,), x)

x = {'a': [(full(1), full(2), 3)], 'b': [(full(4), full(5), 6)], 'c': [(full(7), full(8), 9)]}
in_axes = {'a': [(0, 0, None)], 'b': [(0,)], 'c': [(0, 0, None)]}
jax.vmap(lambda x: x, in_axes=in_axes)(x)
```

```
# BEFORE this PR
ValueError: vmap in_axes specification must be a tree prefix of the
corresponding value, got specification {'a': [(0, 0, None)], 'b': [(0,)], 'c':
[(0, 0, None)]} for value tree PyTreeDef(({'a': [(*, *, *)], 'b': [(*, *, *)],
'c': [(*, *, *)]},)).
```

In real world examples, often involving neural network parameter trees, these errors can be _much_ longer, with hundreds or thousands of tree leaves and significant nesting!

Here's a better kind of error that the new utility function lets us raise:

```python
from jax._src.tree_util import prefix_errors
all_errors = prefix_errors(in_axes, x)
e = all_errors[0]
raise e('in_axes')
```

```
# AFTER this PR
ValueError: pytree structure error: different numbers of pytree children at
in_axes['b'][0]: prefix pytree in_axes has 1 children where full pytree has 3
children.
```

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1404,11 +1404,12 @@ def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
       range ``[-ndim, ndim)`` for each array, where ``ndim`` is the number of
       dimensions (axes) of the corresponding input array.
 
-      If the positional arguments to ``fun`` are container types, the
+      If the positional arguments to ``fun`` are container (pytree) types, the
       corresponding element of ``in_axes`` can itself be a matching container,
       so that distinct array axes can be mapped for different container
       elements. ``in_axes`` must be a container tree prefix of the positional
-      argument tuple passed to ``fun``.
+      argument tuple passed to ``fun``. See this link for more detail:
+      https://jax.readthedocs.io/en/latest/pytrees.html#applying-optional-parameters-to-pytrees
 
       Either ``axis_size`` must be provided explicitly, or at least one
       positional argument must have ``in_axes`` not None. The sizes of the
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -17,7 +17,7 @@ import functools
 from functools import partial
 import operator as op
 from typing import (Any, Callable, Hashable, Iterable, Optional, Tuple, List,
-                    Type, TypeVar, overload, TYPE_CHECKING)
+                    Dict, Type, TypeVar, overload, TYPE_CHECKING, NamedTuple)
 
 from jax._src.lib import pytree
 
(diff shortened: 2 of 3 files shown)


jax/_src/api.py
```diff
@@ -1403,7 +1403,7 @@ def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
       range ``[-ndim, ndim)`` for each array, where ``ndim`` is the number of
       dimensions (axes) of the corresponding input array.
 
-      If the positional arguments to ``fun`` are container types, the
+      If the positional arguments to ``fun`` are container (pytree) types, the
       corresponding element of ``in_axes`` can itself be a matching container,
       so that distinct array axes can be mapped for different container
       elements. ``in_axes`` must be a container tree prefix of the positional
```
gnecula, line 1409: Since the mechanism of parameters that are container tree prefix is used in several places in the API, I think it is worth consolidating a more in-depth explanation and linking to it. E.g., [this](https://jax.readthedocs.io/en/latest/pytrees.html?highlight=arguments#applying-optional-parameters-to-pytrees) in the pytree docs.

jax/_src/tree_util.py
```diff
@@ -335,3 +335,126 @@ def __new__(klass, func, *args, **kw):
...
+  if handler:
+    return handler(pytree)
+  elif isinstance(pytree, tuple) and hasattr(pytree, '_fields'):
+    # handle namedtuple as a special case, based on heuristic
+    return [AttributeKeyPathEntry(s) for s in pytree._fields]
+  else:
+    num_children = len(treedef_children(tree_structure(pytree)))
+    return [FlattenedKeyPathEntry(i) for i in range(num_children)]
+
+_keypath_registry: Dict[Type, Callable[[Any], List[KeyPathEntry]]] = {}
```
gnecula, line 406: Is the plan that there will be other types registered than these? E.g., for custom pytrees?

tests/tree_util_test.py
```diff
@@ -425,5 +425,86 @@ def testDtypeMonomorphicUnravel(self):
       _ = unravel(y)
 
 
+class TreePrefixErrorsTest(jtu.JaxTestCase):
+
+  def test_different_types(self):
```
gnecula, line 430: Shouldn't there be some tests for no errors? 
E.g., (1, 2), ((11, 12, 13), 2)

gnecula (approved):

j-towns:
I made this [proposal](https://github.com/google/jax/issues/5047) a while ago that might be worth implementing to further improve the readability of the errors. IIRC I briefly tried to implement it myself by editing the pytree C++, but gave up after ~1 hour.

jax/_src/api.py
mattjj: Great idea!

jax/_src/tree_util.py
mattjj: Yes, I hope so, at least for library-writers like Flax and Haiku (if they want to use this). Then those parameter data structures can get better error messages. WDYT?

mattjj:
@j-towns great idea, thanks for surfacing that. I remember it now!

(IIUC that would be compatible with this PR, and we should do both, right? Correct me if I'm wrong!)

jax/_src/tree_util.py
gnecula: I am not sure I understand when would this be used? I can think of needing this for custom pytrees. It would be good to have a test/example.

mattjj: Well, libraries like Flax register lots of their own pytrees, e.g. for collections of parameters. Some examples are [`flax.struct`](https://flax.readthedocs.io/en/latest/flax.struct.html) and [`FrozenDict`](https://flax.readthedocs.io/en/latest/_modules/flax/core/frozen_dict.html#FrozenDict). In fact, that's the main motivating use case: several times users have gotten pytree errors when passing `PartitionSpec`s or similar for custom Flax pytrees, and the error messages are long and terrible!

From the point of view of this API, there's no difference between "custom pytrees" and pytrees which come registered with JAX. Even so, there are already tests which cover pytrees which don't come registered with JAX: see the use of `Special` in `test_fallback_keypath`. So I think the test coverage is sufficient.

I expect library developers like Flax folks will start using this API and add tests in their libraries for this functionality used with their own types.

mattjj: Googlers can check out go/bad-pytree-mismatch-error-message-example to see an example such message which came up recently.

tests/tree_util_test.py
mattjj: Done!

PR #9391, opened 2022-01-31
Add complex types to mlir constant handlers

### LINKED ISSUE #9380: Type specification/inferrence now fails in @jax.jit
Please:

The following [test case](https://github.com/ska-sa/codex-africanus/blob/master/africanus/rime/jax/tests/test_jax_phase_delay.py) has worked for a while, but has recently started failing. It seems that there isn't a constant handler for `complex64` anymore? The following reproducer demonstrates the issue, but replacing `out_dtype.type(1j)` with `1j` fixes the problem.

This works on jax 0.2.26 and jaxlib 0.1.75, but fails on jax 0.2.27 and 0.1.76.

```python
import numpy as np
import jax
import jax.numpy as jnp

@jax.jit
def phase_delay(lm, uvw, frequency):
    out_dtype = jnp.result_type(lm, uvw, frequency, np.complex64)

    one = lm.dtype.type(1.0)
    neg_two_pi_over_c = lm.dtype.type(-2*np.pi/3e8)

    l = lm[:, 0, None, None]  # noqa
    m = lm[:, 1, None, None]

    u = uvw[None, :, 0, None]
    v = uvw[None, :, 1, None]
    w = uvw[None, :, 2, None]

    n = jnp.sqrt(one - l**2 - m**2) - one

    real_phase = (neg_two_pi_over_c *
                  (l * u + m * v + n * w) *
                  frequency[None, None, :])
 
    # replacing out_dtype.type(1j) with 1j fixes this problem
    return jnp.exp(out_dtype.type(1j)*real_phase)

if __name__ == "__main__":
    uvw = np.random.random(size=(100, 3)).astype(np.float32)
    lm = np.random.random(size=(10, 2)).astype(np.float32)*0.001
    frequency = np.linspace(.856e9, .856e9*2, 64).astype(np.float32)

    complex_phase = phase_delay(lm, uvw, frequency)
```

- [x] #9390

```
$ python test_complex_constant_fail.py 
  File "test_complex_constant_fail.py", line 32, in <module>
    complex_phase = phase_delay(lm, uvw, frequency)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/traceback_util.py", line 165, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/api.py", line 429, in cache_miss
    donated_invars=donated_invars, inline=inline)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/core.py", line 1671, in bind
    return call_bind(self, fun, *args, **params)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/core.py", line 1683, in call_bind
    outs = top_trace.process_call(primitive, fun, tracers, params)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/core.py", line 596, in process_call
    return primitive.impl(f, *tracers, **params)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/dispatch.py", line 143, in _xla_call_impl
    *unsafe_map(arg_spec, args))
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/linear_util.py", line 272, in memoized_fun
    ans = call(fun, *args)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/dispatch.py", line 170, in _xla_callable_uncached
    *arg_specs).compile().unsafe_call
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/profiler.py", line 206, in wrapper
    return func(*args, **kwargs)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/dispatch.py", line 260, in lower_xla_callable
    donated_invars)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 403, in lower_jaxpr_to_module
    input_output_aliases=input_output_aliases)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 541, in lower_jaxpr_to_fun
    *args)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 606, in jaxpr_subcomp
    in_nodes = map(read, eqn.invars)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/_src/util.py", line 44, in safe_map
    return list(map(f, *args))
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 583, in read
    return ir_constants(v.val, canonicalize_types=True)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 171, in ir_constants
    raise TypeError("No constant handler for type: {}".format(type(val)))
jax._src.traceback_util.UnfilteredStackTrace: TypeError: No constant handler for type: <class 'numpy.complex64'>

--------------------

  File "test_complex_constant_fail.py", line 32, in <module>
    complex_phase = phase_delay(lm, uvw, frequency)
  File "/home/sperkins/venv/afr/lib/python3.7/site-packages/jax/interpreters/mlir.py", line 171, in ir_constants
    raise TypeError("No constant handler for type: {}".format(type(val)))
TypeError: No constant handler for type: <class 'numpy.complex64'>
```

jakevdp:
Fixes #9380

What's still unclear to me is why this only errors with the most recent jaxlib release.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -3120,6 +3120,16 @@ class APITest(jtu.JaxTestCase):
     expected = jnp.arange(1) + 1
     self.assertAllClose(ans, expected)
 
+  @parameterized.named_parameters([
+      {"testcase_name": f"{dtype.__name__}", "dtype": dtype}
+      for dtype in jtu.dtypes.all])
+  def test_constant_handlers(self, dtype):
+    # https://github.com/google/jax/issues/9380
+    @jax.jit
+    def f():
+      return jnp.exp(dtype(0))
+    f()  # doesn't error
+
   def test_large_python_ints(self):
     with self.assertRaises(OverflowError):
       jnp.multiply(2 ** 100, 3.)
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -241,9 +241,10 @@ def _ndarray_constant_handler(val: np.ndarray, canonicalize_types
 register_constant_handler(np.ndarray, _ndarray_constant_handler)
 
 for _scalar_type in [np.int8, np.int16, np.int32, np.int64,
-                    np.uint8, np.uint16, np.uint32, np.uint64,
-                    np.float16, np.float32, np.float64,
-                    np.bool_, np.longlong, dtypes.bfloat16]:
+                     np.uint8, np.uint16, np.uint32, np.uint64,
+                     np.float16, np.float32, np.float64,
+                     np.complex64, np.complex128,
+                     np.bool_, np.longlong, dtypes.bfloat16]:
   register_constant_handler(_scalar_type, _ndarray_constant_handler)
 
 def _python_scalar_handler(dtype, val, canonicalize_dtypes):


hawkinsp (approved):
This probably started failing on the latest jaxlib because MLIR was enabled gated on the jaxlib being new enough.

jakevdp:
Thanks @hawkinsp. As I was looking into the origin of this, I found that the associated constant handler is not defined in `xla.py` either: https://github.com/google/jax/blob/c12ca7f64cb4a0338939d84b7dc5c25ccc5b47be/jax/interpreters/xla.py#L356-L361

Any thoughts as to why this bug hasn't come up before?

hawkinsp:
> Thanks @hawkinsp. As I was looking into the origin of this, I found that the associated constant handler is not defined in `xla.py` either:
> 
> https://github.com/google/jax/blob/c12ca7f64cb4a0338939d84b7dc5c25ccc5b47be/jax/interpreters/xla.py#L356-L361
> 
> Any thoughts as to why this bug hasn't come up before? Should we add complex value handlers there as well?

It turns out to be because `xla.py` calls (redundantly) calls `canonicalize_dtype` on the value it passes to the function that constructs IR constants:
https://github.com/google/jax/blob/c12ca7f64cb4a0338939d84b7dc5c25ccc5b47be/jax/interpreters/xla.py#L564

This is pointless (mostly) because the callee would canonicalize it anyway. However, it does have the unintended effect of turning NumPy scalars into arrays, which means you never see the scalar value in the callee (via this route, anyway).

tests/api_test.py
```diff
@@ -3120,6 +3120,16 @@ def f(_):
     expected = jnp.arange(1) + 1
     self.assertAllClose(ans, expected)
 
+  @parameterized.named_parameters([
+      {"testcase_name": f"{dtype.__name__}", "dtype": dtype}
+      for dtype in jtu.dtypes.all])
```
hawkinsp, line 3125: The issue you have in the internal CI is that you cannot determine which backend we are using (via `jtu.dtypes`) until after absl initialization has run. This parameter gets instantiated too early.

This is why we usually use the idiom of skipping unsupported dtypes.

jakevdp: If that's the case, then why do these not cause the same issue? https://github.com/google/jax/blob/f80887e69cdac905a7c748622fd6bcd75bbc0cf9/tests/lax_numpy_test.py#L67-L75

hawkinsp: They do. But I suspect anyone calling that test is setting `--jax_test_dut`:
https://github.com/google/jax/blob/f80887e69cdac905a7c748622fd6bcd75bbc0cf9/jax/_src/test_util.py#L434

jakevdp: Can we require that for tests? (it's not our own internal tests that are failing, but someone else using our targets)

hawkinsp: Yes perhaps. Probably we'd do well to make the logic unconditionally get its "target device" from exactly one source: either flags or "what's registered" but not both. Not necessarily for this PR.

PR #9394, opened 2022-01-31
update pre-commit package versions & fix new mypy errors

jakevdp:


jax/_src/numpy/lax_numpy.py
```diff
@@ -1864,7 +1864,7 @@ def swapaxes(a, axis1: int, axis2: int):
   _check_arraylike("swapaxes", a)
   perm = np.arange(ndim(a))
   perm[axis1], perm[axis2] = perm[axis2], perm[axis1]
-  return lax.transpose(a, perm)
+  return lax.transpose(a, list(perm))
```
apaszke, line 1867: What's the error here? Are NumPy arrays not sequences?

jax/_src/numpy/util.py
```diff
@@ -92,7 +92,7 @@ def _parse_parameters(body: str) -> Dict[str, str]:
   return OrderedDict((p.partition(' : ')[0].partition(', ')[0], p) for p in parameters)
 
 
-def _wraps(fun: Callable, update_doc: bool = True, lax_description: str = "",
+def _wraps(fun: Optional[Callable], update_doc: bool = True, lax_description: str = "",
```
apaszke, line 95: Why is this optional now?

apaszke (approved):
LGTM, but it was surprising in two places?

jax/_src/numpy/util.py
jakevdp: We treat it as optional here: https://github.com/google/jax/blob/e3fe4a2c7c31ce56b6b99469b12a05e43a57727c/jax/_src/numpy/lax_numpy.py#L2207

PR #9396, opened 2022-01-31
type promotion: disable uint64->float

jakevdp:
In the current type promotion lattice, there is a promotion path from `uint64` to `float*`, which means that JAX follows numpy's behavior of promoting `np.uint64`  & `np.int64` to float.

This promotion behavior can be surprising, and there are good arguments that we should not do this kind of promotion implicitly.

This PR is a draft that removes this behavior, such that `np.uint64` promoting with any signed type results in an error.

Pulling in before review for additional testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 ## jax 0.2.29 (Unreleased)
 * [GitHub
   commits](https://github.com/google/jax/compare/jax-v0.2.28...main).
+  * In `jax.numpy` type promotion, values of type `np.uint64` now raise a `TypePromotionError`
+    in case of implicit type promotion to float. This affects its use in binary operations,
+    such as `2.0 * x`, which for `x` of type `uint64` previously returned `float64` and now
+    raises an error. It also affects any function that implicitly casts inputs to float, such
+    as `jnp.exp(x)`, which for `x` of type `uint64` previously returned `float64` and now
+    raises an error. The workaround in either case is to explicitly cast values to
+    `float64` when this behavior is desired; e.g. `2.0 * jnp.float64(x)`.
 
 ## jax 0.2.28 (Feb 1 2022)
 * [GitHub
--- a/tests/dtypes_test.py
+++ b/tests/dtypes_test.py
@@ -136,33 +136,52 @@ class DtypesTest(jtu.JaxTestCase):
       self.assertEqual(z.dtype, dtypes.canonicalize_dtype(dtype), msg=(x, y, z))
 
   def testPromoteDtypes(self):
-    for t1 in all_dtypes:
-      self.assertEqual(t1, dtypes.promote_types(t1, t1))
+    inexact_types = float_dtypes + complex_dtypes
+    exact_types = bool_dtypes + signed_dtypes + [
+      t for t in unsigned_dtypes if t != np.uint64]
 
+    for t1 in inexact_types + exact_types:
+      self.assertEqual(t1, dtypes.promote_types(t1, t1))
       self.assertEqual(t1, dtypes.promote_types(t1, np.bool_))
       self.assertEqual(np.dtype(np.complex128),
                        dtypes.promote_types(t1, np.complex128))
 
-      for t2 in all_dtypes:
+      for t2 in inexact_types + exact_types:
         # Symmetry
         self.assertEqual(dtypes.promote_types(t1, t2),
                          dtypes.promote_types(t2, t1))
 
-    self.assertEqual(np.dtype(np.float32),
-                     dtypes.promote_types(np.float16, dtypes.bfloat16))
+    # bfloat16 promotes like float16, but bfloat16 + float16 -> float32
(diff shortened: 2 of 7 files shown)


tests/lax_numpy_test.py
```diff
@@ -128,11 +126,11 @@ def op_record(name, nargs, dtypes, shapes, rng_factory, diff_modes,
                   test_name, check_dtypes, tolerance, inexact, kwargs)
 
 JAX_ONE_TO_ONE_OP_RECORDS = [
-    op_record("abs", 1, number_dtypes + unsigned_dtypes + bool_dtypes,
+    op_record("abs", 1, number_dtypes + unsigned_dtypes_no64 + bool_dtypes,
```
hawkinsp, line 129: This seems surprising. I would have expected `abs` not to promote.

tests/lax_numpy_test.py
```diff
@@ -235,11 +233,11 @@ def op_record(name, nargs, dtypes, shapes, rng_factory, diff_modes,
     op_record("expm1", 1, number_dtypes, all_shapes, jtu.rand_small_positive,
               [], tolerance={np.float64: 1e-8}, inexact=True),
     op_record("fix", 1, float_dtypes, all_shapes, jtu.rand_default, []),
-    op_record("fix", 1, int_dtypes + unsigned_dtypes, all_shapes,
+    op_record("fix", 1, int_dtypes + unsigned_dtypes_no64, all_shapes,
               jtu.rand_default, [], check_dtypes=False),
     op_record("floor_divide", 2, float_dtypes + int_dtypes,
               all_shapes, jtu.rand_nonzero, ["rev"]),
-    op_record("floor_divide", 2, unsigned_dtypes, all_shapes,
+    op_record("floor_divide", 2, unsigned_dtypes_no64, all_shapes,
```
hawkinsp, line 240: This does remove a useful test case, which is u64 // u64. Same for most of the binary operators changed by this PR.

I wonder if these test harness would be better expressed by providing the complete set of valid types for all arguments, built using some helpers.

Another option would be to add another record just for the u64/u64 case.

(Not to mention: we should probably be testing invalid types also and checking for errors.)

tests/lax_numpy_test.py
```diff
@@ -288,18 +286,18 @@ def op_record(name, nargs, dtypes, shapes, rng_factory, diff_modes,
...
-    op_record("modf", 1, int_dtypes + unsigned_dtypes, all_shapes,
+    op_record("modf", 1, int_dtypes + unsigned_dtypes_no64, all_shapes,
               jtu.rand_default, [], check_dtypes=False),
     op_record("rint", 1, inexact_dtypes, all_shapes, jtu.rand_some_inf_and_nan,
               []),
-    op_record("rint", 1, int_dtypes + unsigned_dtypes, all_shapes,
+    op_record("rint", 1, int_dtypes + unsigned_dtypes_no64, all_shapes,
               jtu.rand_default, [], check_dtypes=False),
-    op_record("sign", 1, number_dtypes + unsigned_dtypes,
+    op_record("sign", 1, number_dtypes + unsigned_dtypes_no64,
```
hawkinsp, line 295: This also seems surprising. In classic NumPy, numpy.sign of a `uint64` yields a `uint64`.

tests/dtypes_test.py
```diff
@@ -136,34 +136,53 @@ def testBinaryPromotion(self, swap, jit):
...
 
   def testPromoteDtypes(self):
-    for t1 in all_dtypes:
-      self.assertEqual(t1, dtypes.promote_types(t1, t1))
+    inexact_types = float_dtypes + complex_dtypes
+    exact_types = bool_dtypes + signed_dtypes + [
+      t for t in unsigned_dtypes if t != np.uint64]
 
+    for t1 in inexact_types + exact_types:
+      self.assertEqual(t1, dtypes.promote_types(t1, t1))
```
hawkinsp, line 144: Looks like you should add this test case for uint64 as well.

CHANGELOG.md
```diff
@@ -11,6 +11,13 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 ## jax 0.2.29 (Unreleased)
 * [GitHub
   commits](https://github.com/google/jax/compare/jax-v0.2.28...main).
+  * In `jax.numpy` type promotion, values of type `np.uint64` now raise a `TypePromotionError`
```
hawkinsp, line 14: I think it's probably also worth emphasizing that it's the (uint64, int64)-> float promotion that is particularly odd. (uint64, float) -> float seems sensible to me, but it's the consequence of that in the lattice rules for uint64 and int64 that is troubling.

hawkinsp (changes requested):
Update https://jax.readthedocs.io/en/latest/type_promotion.html also?

CHANGELOG.md
jakevdp: Yeah, this is strange... Semantically it might be nice if `uint64,float -> float` were allowed, but `uint64,int -> float` were not. But I don't think that's possible within a lattice-based promotion approach.

jakevdp:
Closing this, it's become far out-of-date.

PR #9407, opened 2022-02-02
Add type promotion design doc

jakevdp:
*Edit: result of this PR can be viewed here: https://jax.readthedocs.io/en/latest/design_notes/type_promotion.html*

---

I think we're ready to publish a draft of this notebook. We can still iterate, particularly in the discussion around uint64 promotion (I left ample room in the discussion to drop `u64->f*`, as explored #9396)

Preview the notebook here: https://colab.sandbox.google.com/github/jakevdp/jax/blob/type-promotion-design/docs/design_notes/type_promotion.ipynb

Preview of rendered doc: https://jax--9407.org.readthedocs.build/en/9407/design_notes/type_promotion.html


design_notes/type_promotion.ipynb
```diff
@@ -0,0 +1,1418 @@
...
+   "source": [
+    "### Properties of a Type Promotion Lattice\n",
+    "\n",
+    "Specifying type promotions in terms of a lattice ensures a number of useful properties. Denoting the lattice join as *J*, we have:\n",
+    "\n",
+    "**Existence:** A lattice by definition requires that a unique lattice join exists for every pair of elements: $\\forall (a, b): \\exists ! J(a, b)$\n",
+    "\n",
+    "**Commutativity:** The lattice join is commutative, i.e. for all *a* and *b* on the lattice, $\\forall (a, b): J(a, b) = J(b, a)$.\n",
+    "\n",
+    "**Associativity:** The lattice join is associative, i.e. for all *a*, *b*, and *c* on the lattice, $\\forall (a, b): J(a, J(b, c)) = J(J(a, b), c)$.\n",
```
soraros: I think you meant `\forall a, b, c`. I would also suggest using the common symbol `\vee` for join.

jakevdp: Good idea - Done

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+Specifying type promotions in terms of a lattice ensures a number of useful properties. Denoting the join on the lattice with the $\vee$ operator, we have:
+
+**Existence:** A lattice by definition requires that a unique lattice join exists for every pair of elements: $\forall (a, b): \exists !(a \vee b)$
+
+**Commutativity:** The lattice join is commutative: $\forall (a, b): a\vee b = b \vee a$.
+
+**Associativity:** The lattice join is associative: $\forall (a, b, c): a \vee (b \vee c) = (a \vee b) \vee c$.
+
+Both commutativity and associativity are useful properties for type promotion to satisfy, because they mirror the commutativity and associativity of basic binary arithmetic operations like addition and multiplication.
+
```
hawkinsp, line 130: Perhaps say highlight that these are good things for a type promotion system to have? You don't quite draw that logical connection. You say, in essence:
* type promotion can be represented as a lattice
* NumPy promotion is not representable as a lattice

but seem to be missing a chance to argue: the properties of a lattice are properties that any reasonable programmer would expect from type promotion and that anything else is surprising.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+```python
+np.int32(3) * (x + np.int32(1)) ** np.int32(2)
+```
+Although it is explicit, numerical code would become tedious to read or write. With the scalar promotion semantics described above, given an array `x` of type `int32`, the types in the second statement are implicit within the first.
+
++++ {"id": "HX7yeIf6jbjz"}
+
+## Combining Lattices
+
+Recall that we began our discussion by introducing the lattice representing type promotion within Python: `int -> float -> complex`. Let's rewrite this as `i* -> f* -> c*`, and lets further allow `i*` to subsume `u*` (after all, there is no unsigned integer scalar type in Python).
```
hawkinsp, line 322: typo "let's"

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
++++ {"id": "WVjEMlmm9RS5"}
+
+Notice that this is not (yet) a true lattice: there are many pairs of nodes for which a join does not exist. However, we can think of this as a *partial* lattice, in which some pairs of nodes do not have a defined promotion behavior, and the defined portion of this partial lattice does correctly describe NumPy's array promotion behavior (leaving aside value-dependent semantics mentioned above).
+
+This sets up a nice framework by which we can think about filling-out these undefined promotion rules, by adding connections on this graph. But which connections to add?
+Broadly speaking, we want any additional connections to satisfy a few properties:
+
+1. Connections should maintain the commutativity and associativity of type promotion: that is, the graph should remain a (partial) lattice.
+
+2. Connections should never allow for a loss of data: for example, we should never promote `float` to `int`, as it would discard everything beyond the decimal point.
```
hawkinsp, line 362: The phrasing of this could probably be improved, given as I read it we contradict it a couple of points later when pointing out that int->float promotions may lose mantissa bits. I'm not quite sure how to phrase it though. What do you mean by "loss of data"?

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+```
+
++++ {"id": "WVjEMlmm9RS5"}
+
+Notice that this is not (yet) a true lattice: there are many pairs of nodes for which a join does not exist. However, we can think of this as a *partial* lattice, in which some pairs of nodes do not have a defined promotion behavior, and the defined portion of this partial lattice does correctly describe NumPy's array promotion behavior (leaving aside value-dependent semantics mentioned above).
+
+This sets up a nice framework by which we can think about filling-out these undefined promotion rules, by adding connections on this graph. But which connections to add?
+Broadly speaking, we want any additional connections to satisfy a few properties:
+
+1. Connections should maintain the commutativity and associativity of type promotion: that is, the graph should remain a (partial) lattice.
```
hawkinsp, line 360: I might phrase these as "Promotion should" rather than "Connections should".

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+
+2. Connections should never allow for a loss of data: for example, we should never promote `float` to `int`, as it would discard everything beyond the decimal point.
+
+3. Connections should never allow for a loss of magnitude. For example, the maximum possible `uint32` is twice as large as the maximum possible `int32`, so we should not implicitly promote `uint32` to `int32`.
```
hawkinsp, line 364: Rephrase a bit? I know what you mean, but "loss of magnitude" isn't something I can define exactly. "Should not lead to overflow?"

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+
+4. Wherever possible, connections should avoid loss of precision. For example, an `int64` value may have 64 bits of mantissa, so promoting `int64` to `float64` represents a possible loss of precision. However, because the maximum representable float64 is larger than the maximum representable int64, the *magnitude* can still be correctly represented.
+
+5. Wherever possible, connections should avoid unnecessary promotion to larger bit-widths. The purpose of this is to ensure that JAX's implicit promotions remain friendly to accelerator-based workflows, in which users often want to restrict types to 32-bit (or in some cases 16-bit) values.
```
hawkinsp, line 368: Go one step further: promotion will not produce higher precision types than those the user used as input.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+
+Each new connection introduces some level of convenience to the user (a new set of types that can interact without explicit casting), but the convenience may becomes too costly if any of the above criteria are violated. Developing a full promotion lattice involves striking a balance between this convenience and this cost.
+
++++ {"id": "GSqwTTS8nYdn"}
+
+## Mixed Promotion: Float and Complex
+
+Let's begin with what is perhaps the easiest case, that of promotion between float and complex values.
+
+Complex numbers are made up of pairs of floating point numbers, and so we have a natural path of promotion between them: cast float to complex while maintaining the width of the real part. In terms of our lattice representation, it would look like this:
```
hawkinsp, line 378: You should probably note that this is in fact not a lattice: the supremum is not defined for all values. So you should probably note that this is a partial lattice and that in the future when you write "lattice" you mean partial lattice.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+
+This is effectively what Numpy type promotion does, but in doing so it breaks the lattice property of the graph: for example, the pair *{i8, u8}* no longer has a unique least upper bound: the possibilities are *i16* and *f16*, which are unorderable on the graph. This turns out to be the source of NumPy's non-associative type promotion highlighted above.
+
+Can we come up with a modification of NumPy's promotion rules, such that it will satisfy the lattice property, while also giving sensible results for mixed type promotion? There are a few approaches we could take here.
+
++++ {"id": "nLXKOk48lfY2"}
+
+### Option 0: Leave integer/floating mixed precision undefined
+
+To make behavior utterly predictable (at some cost to user convenience), a defensible choice would be to leave mixed integer/float type promotion undefined, stopping with the partial lattice from the previous section. The cost would be requiring the user to explicitly type-cast when operating between integer and floating-point quantities.
```
hawkinsp, line 512: One thing to note: the previous lattice would still allow Python scalars to be promoted to floats.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+}
+fig, ax = plt.subplots(figsize=(6, 5))
+nx.draw(graph, with_labels=True, node_size=1500, node_color='lightgray', pos=pos, ax=ax)
+```
+
++++ {"id": "BfHlmmF_GOo_"}
+
+While this does allow for precision-losing promotions between integers and floats, these promotions will not mis-represent the *magnitude* of the result: though the floating point mantissa is not wide enough to represent all values, the exponent is wide enough to approximate them.
+
+This approach also allows a natural promotion path from `int64` to `float64`, though `uint64` remains unpromotable. That said, a connection from `u64` to `f64` would be more defensible here than before.
```
hawkinsp, line 592: remains unpromotable ... if we wish to avoid i64/u64 promoting to f64.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+  'f16': [1.5, 2], 'f32': [2.5, 2], 'f64': [3.5, 2],
+  'c64': [3, 3], 'c128': [4, 3],
+}
+fig, ax = plt.subplots(figsize=(6, 5))
+nx.draw(graph, with_labels=True, node_size=1500, node_color='lightgray', pos=pos, ax=ax)
+```
+
++++ {"id": "xBDCy0AnGsbJ"}
+
+This involves a small slight-of-hand: previously we had used `f*` to refer to a scalar type. In this lattice, `f*` might be applied to the array output of a mixed computation. Instead of thinking of `f*` as a scalar, we could think of it as a special kind of `float64`, with promotion rules that differ from those of `f64`.
```
hawkinsp, line 632: sp. "sleight of hand".

(not "slight").

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+
+The advantage of this approach is that it avoids *all* wider-than-necessary promotions: you can never get an f64 output without a 64-bit input, and you can never get an f32 output without a 32-bit input: this results in convenient semantics for working on accelerators while avoiding inadvertent 64-bit values.
+
+This lattice also happens to very closely resembles JAX's original *ad hoc* type promotion table, which was not based on a lattice, but had this feature of giving primacy to floating point types.
+
+Additionally, this lattice offers a natural location to insert `bfloat16`, without the need to impose an ordering between `bf16` and `f16`.
```
hawkinsp, line 638: It's probably worth highlighting that bf16 and f16 are incomparable: one represents a larger range, but the other has more precision.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+  'i8': [0, 1], 'i16': [1, 1], 'i32': [2, 1], 'i64': [3, 1],
+  'f16': [5.75, 0.8], 'bf16': [5.75, 0.2], 'f32': [7, 0.5], 'f64': [8, 0.5],
+  'c64': [7.5, 1.5], 'c128': [8.5, 1.5],
+}
+fig, ax = plt.subplots(figsize=(10, 4))
+ax.set_ylim(-0.5, 2)
+nx.draw(graph, with_labels=True, node_size=1500, node_color='lightgray', pos=pos, ax=ax)
+# ax.patches[12].set_linestyle((0, (2, 4)))
+```
+
```
hawkinsp, line 719: Updating removing the u64->f* edge?

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
++++ {"id": "e2h9qOh2KgZy"}
+
+# Design of Type Promotion Semantics for JAX
+
+[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/jax/blob/main/docs/design_notes/type_promotion.ipynb)
+
+*Jake VanderPlas, December 2021*
+
+One of the challenges faced in the design of any numerical computing library is the choice of how to handle operations between values of different types. This document describes the reasoning behind the type promotion semantics used in [JAX](http://github.com/google/jax). This document outlines the thought process behind the promotion semantics used by JAX, summarized in [JAX Type Promotion Semantics](https://jax.readthedocs.io/en/latest/type_promotion.html).
```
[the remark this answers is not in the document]
jakevdp: Done

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Tried to rephrase... still not sure how to best put this.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done - I think you're right that referring to overflow is a bit more clear.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Done.

docs/design_notes/type_promotion.md
jakevdp: Yes, this was intentional. I discuss u64 more below.

docs/design_notes/type_promotion.md
```diff
@@ -0,0 +1,722 @@
...
+fig, ax = plt.subplots(figsize=(6, 5))
+nx.draw(graph, with_labels=True, node_size=1500, node_color='lightgray', pos=pos, ax=ax)
+```
+
++++ {"id": "FZ6ana2UooNh"}
+
+However, these advantages comes with a few tradeoffs:
+
+- mixed float/integer promotion is very prone to precision loss: for example, `int64` (with a maximum value of $9.2 \times 10^{18}$) can be promoted to `float16` (with a maximum value of $6.5 \times 10^4$), meaning most representable values will become `inf`.
+- as mentioned above, `f*` can no longer be thought of as a "scalar type", but whether as a different flavor of float64. In JAX's parlance, this is referred to as a *weak type*, in that it is represented as 64-bit, but only weakly holds to this bit width in promotion with a strongly-typed value.
```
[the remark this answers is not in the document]
jakevdp: I linked to our other discussion of weak types.

PR #9416, opened 2022-02-02
Move design notes into docs/, and render them as part of the document…

hawkinsp:
…ation.

It seems like a shame to me that these aren't rendered in the documentation tree!


mattjj (approved):
Great idea.

jakevdp:
Looks good – though I think you need to add these docs to the contents in https://github.com/google/jax/blob/main/docs/index.rst

jakevdp (approved):

PR #9417, opened 2022-02-02
Support string fft_type values in lax.fft.

hawkinsp:


jax/_src/lax/fft.py
```diff
@@ -43,9 +44,28 @@ def _promote_to_real(arg):
   dtype = dtypes.result_type(arg, np.float32)
   return lax.convert_element_type(arg, dtype)
 
+def _str_to_fft_type(s: str):
```
jakevdp, line 47:
```suggestion
def _str_to_fft_type(s: str) -> xla_client.FftType:
```
as long as we're doing type annotations...

jakevdp (approved):

PR #9419, opened 2022-02-02
Introduce a new jax/jaxlib versioning scheme.

hawkinsp:
Adds a design note that describes the scheme and how the jax and jaxlib versions
are related.

### DIFF
--- a/docs/design_notes/jax_versioning.md
+++ b/docs/design_notes/jax_versioning.md
@@ -1,0 +1,189 @@
+# Jax and Jaxlib versioning
+
+## Why are `jax` and `jaxlib` separate packages?
+
+We publish JAX as two separate Python wheels, namely `jax`, which is a pure
+Python wheel, and `jaxlib`, which is a mostly-C++ wheel that contains libraries
+such as:
+* XLA,
+* pieces of LLVM used by XLA,
+* MLIR infrastructure, such as the MHLO Python bindings.
+* JAX-specific C++ libraries for fast JIT and PyTree manipulation.
+
+We distribute separate `jax` and `jaxlib` packages because it makes it easy to
+work on the Python parts of JAX without having to build C++ code or even having
+a C++ toolchain installed. `jaxlib` is a large library that is not easy for
+many users to build, but most changes to JAX only touch Python code. By
+allowing the Python pieces to be updated independently of the C++ pieces, we
+improve the development velocity for Python changes.
+
+In addition `jaxlib` is not cheap to build, but we want to be able to iterate on
+and run the JAX tests in environments without a lot of CPU, for example in
+Github Actions or on a laptop. Many of our CI builds simply use a prebuilt
+`jaxlib`, rather than needing to rebuild the C++ pieces of JAX on each PR.
+
+As we will see, distributing `jax` and `jaxlib` separately comes with a cost, in
+that it requires that changes to `jaxlib` maintain a backward compatible API.
+However, we believe that on balance it is preferable to make Python changes
+easy, even if at the cost of making C++ changes slightly harder.
+
+
+## How are `jax` and `jaxlib` versioned?
+
+Summary: `jax`'s version must be greater than or equal to `jaxlib`'s version,
+and `jaxlib`'s version must be greater than or equal to the minimum `jaxlib`
+version specified by `jax`.
+
+Both `jax` and `jaxlib` releases are numbered `x.y.z`, where `x` is the major
+version, and `y` is the minor version, and `z` is an optional patch release.
+Version numbers must follow
+[PEP 440](https://www.python.org/dev/peps/pep-0440/). Version number comparisons
+are lexicographic comparisons on tuples of integers.
+
+Each `jax` release has an associated minimum `jaxlib` version `mx.my.mz`. The
+minimum `jaxlib` version for `jax` version `x.y.z` must be no greater than
+`x.y.z`.
+
+For `jax` version `x.y.z` and `jaxlib` version `lx.ly.lz` to be compatible,
+the following must hold:
+
+* The jaxlib version (`lx.ly.lz`) must be greater than or equal to the minimum
+  jaxlib version (`mx.my.mz`).
+* The jax version (`x.y.z`) must be greater than or equal to the jaxlib version
+  (`lx.ly.lz`).
+
+These constraints imply the following rules for releases:
+* `jax` may be released on its own at any time, without updating `jaxlib`.
+* If a new `jaxlib` is released, a `jax` release whose version is equal to or
+  greater than version the `jaxlib`'s version number must be made at the same
+  time.
+
+These
+[version constraints](https://github.com/google/jax/blob/main/jax/version.py)
+are currently checked by `jax` at import time, instead of being expressed as
+Python package version constraints. `jax` checks the `jaxlib` version at
+runtime rather than using a `pip` package version constraint because we
+[provide separate `jaxlib` wheels](https://github.com/google/jax#installation)
+for a variety of hardware and software versions (e.g, GPU, TPU, etc.). Since we
+do not know which is the right choice for any given user, we do not want `pip`
+to install a `jaxlib` package for us automatically.
+
+In the future, we hope to separate out the hardware-specific pieces of `jaxlib`
+into separate plugins, at which point the minimum version could be expressed as
+a Python package dependency. For now, we do provide
+platform-specific extra requirements that install a compatible jaxlib version,
+e.g., `jax[cuda]`.
+
+## How can I safely make changes to the API of `jaxlib`?
+
+* `jax` may drop compatibility with older `jaxlib` releases at any time, so long
+  as the minimum `jaxlib` version is increased to a compatible version. However,
+  note that the minimum `jaxlib`, even for unreleased versions of `jax`, must be
+  a released version! This allows us to use released `jaxlib` wheels in our CI
+  builds, and allows Python developers to work on `jax` at HEAD without ever
+  needing to build `jaxlib`.
+
+  For example, to remove an old backwards compatibility path in the `jax` Python
+  code, it is sufficient to bump the minimum jaxlib version and then delete the
+  compatibility path.
+
+* `jaxlib` may drop compatibility with older `jax` releases lower than
+  its own release version number. The version constraints enforced by `jax`
+  would forbid the use of an incompatible `jaxlib`.
+
+  For example, for `jaxlib` to drop a Python binding API used by an older `jax`
+  version, the `jaxlib` minor or major version number must be incremented.
+
+* If possible, changes to the `jaxlib` should be made in a backwards-compatible
+  way.
+
+  In general `jaxlib` may freely change its API, so long
+  as the rules about `jax` being compatible with all `jaxlib`s at least as new
+  as the minimum version are followed. This implies that
+  `jax` must always be compatible with at least two versions of `jaxlib`,
+  namely, the last release, and the tip-of-tree version, effectively
+  the next release. This is easier to do if compatibility is maintained,
+  although incompatible changes can be made using version tests from `jax`; see
+  below.
+
+  For example, it is usually safe to add a new function to `jaxlib`, but unsafe
+  to remove an existing function or to change its signature if current `jax` is
+  still using it. Changes to `jax` must work or degrade gracefully
+  for all `jaxlib` releases greater than the minimum up to HEAD.
+
+
+Note that the compatibility rules here only apply to *released* versions of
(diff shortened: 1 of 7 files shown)


jax/_src/lib/__init__.py
```diff
@@ -52,20 +52,25 @@
...
+    msg = (f'jaxlib is version {jaxlib_version.__version__}, but this version '
+           f'of jax requires version {_minimum_jaxlib_version_str}.')
     raise ValueError(msg)
 
+  jaxlib_epoch = getattr(jaxlib_version, "epoch", 1)
+  if jaxlib_epoch > _maximum_jaxlib_epoch:
+    raise ValueError(
+      f'The jaxlib version (version {jaxlib_version.__version__}) is too new '
+       'for this version of jax '
+      f'(epoch {jaxlib_epoch} > {_maximum_jaxlib_epoch})')
```
yashk2810, line 72: Add `Please update your jax version or downgrade your jaxlib version`?

yashk2810 (approved):

docs/design_notes/jaxlib_versioning.md
```diff
@@ -0,0 +1,153 @@
...
+`jaxlib` releases. 
+
+However, since the `jax` and `jaxlib` code is split across repositories that
+cannot be updated atomically at the same time, we need to manage compatibility
+at a finer granularity than our release cycle. To manage fine-grained
+compatibility, we have additional versioning that is independent of the `jaxlib`
+release version numbers.
+
+We maintain an additional version number (`_version`) in [`xla_client.py` in
+the XLA repository](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/python/xla_client.py#L45). The idea is that this version number, is defined in `xla/python`
```
jakevdp, line 111: Possibly link to a release tag rather than `master` here, to avoid the link becoming outdated?

hawkinsp: Removed the line number instead, which should keep the link reasonably fresh.

hawkinsp:
I revised this PR significantly. PTAL.

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,175 @@
...
+
+For `jax` version `x.y.z` and `jaxlib` version `lx.ly.lz` to be compatible,
+the following must hold:
+
+* The jaxlib version (`lx.ly.lz`) must be greater than or equal to the minimum
+  jaxlib version (`mx.my.mz`).
+* The jax version (`x.y.z`) must have major and minor versions (i.e., `x.y`)
+  greater than or equal to the jaxlib major and minor versions (i.e., `mx.my`).
+  For example, `jax` 3.4.5 is compatible with `jaxlib` 3.4.9, but not `jaxlib`
+  3.5.0 or 4.1.3.
```
skye, line 51: Given that we very rarely bump the jax major or minor version numbers, will guaranteeing jax 0.2.X compatibility with future jaxlibs indefinitely be a problem? E.g. what about when we change APIs?

skye (changes requested):
This is awesome, thanks!

I would LGTM except for my one question about old jax versions + new jaxlib versions.

jax/_src/lib/__init__.py
```diff
@@ -41,33 +42,46 @@
...
-           f'but this version of jax requires version {_minimum_jaxlib_version_str}.')
-
-    if version == (0, 1, 23):
-      msg += ('\n\nA common cause of this error is that you installed jaxlib '
-              'using pip, but your version of pip is too old to support '
-              'manylinux2010 wheels. Try running:\n\n'
-              'pip install --upgrade pip\n'
-              'pip install --upgrade jax jaxlib\n')
-    raise ValueError(msg)
+  version_regex = re.compile(r"[0-9]+\.[0-9]+(?:\.[0-9]+)*")
```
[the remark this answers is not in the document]
hawkinsp: Added a comment and simplified the regex. In general we allow any number of version components; no reason to forbid them.

docs/design_notes/jax_versioning.md
hawkinsp: Changed the design as discussed offline.

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,175 @@
...
+  In other words, it should always be sufficient to use `jaxlib` releases with
+  the same `major.minor` version as the `jax` release, if one exists, or the
+  most recent release before that, if not.
+
+  For example, to remove an old backwards compatibility path in the `jax` Python
+  code, it is sufficient to bump the minimum jaxlib version and then delete the
+  compatibility path.
+
+* For `jaxlib` to make a change that is not compatible with existing `jax`
+  releases, its `major.minor` version numbers must be incremented. The version
```
yashk2810, line 94: If we make let's say 4 or 5 incompatible changes then we'll be at jaxlib 4.xx, or 5.xx. Is that desirable?

jax/_src/lib/__init__.py
yashk2810, line 57: Can you add a comment as to what this regex does?

I can parse the first 2 things and the `*` at the  end is confusing. SHould it be `$`?

https://xkcd.com/1171/

jax/_src/lib/__init__.py
```diff
@@ -41,33 +42,46 @@
...
+  _minimum_jaxlib_version = _parse_version(jax.version._minimum_jaxlib_version)
+  _jaxlib_version = _parse_version(jaxlib.version.__version__)
+
+  if _jaxlib_version < _minimum_jaxlib_version:
+    msg = (f'jaxlib is version {jaxlib.version.__version__}, but this version '
+           f'of jax requires version >= {jax.version._minimum_jaxlib_version}.')
+    raise RuntimeError(msg)
+
+  if _jaxlib_version[:2] > _jax_version[:2]:
+    msg = (f'jaxlib version {jaxlib.version.__version__} is newer than and '
```
yashk2810, line 74: Why is this in a variable? WHy not put it directly inside raise RuntimeError(f'...')

jax/_src/lib/__init__.py
```diff
@@ -41,33 +42,46 @@
...
+  _jax_version = _parse_version(jax.version.__version__)
+  _minimum_jaxlib_version = _parse_version(jax.version._minimum_jaxlib_version)
+  _jaxlib_version = _parse_version(jaxlib.version.__version__)
+
+  if _jaxlib_version < _minimum_jaxlib_version:
+    msg = (f'jaxlib is version {jaxlib.version.__version__}, but this version '
+           f'of jax requires version >= {jax.version._minimum_jaxlib_version}.')
+    raise RuntimeError(msg)
+
+  if _jaxlib_version[:2] > _jax_version[:2]:
```
yashk2810, line 73: Are there tests for this code? Is it possible to add them?

docs/design_notes/jax_versioning.md
hawkinsp: I just pushed a new version of the PR, but note that *nothing* either past or previous requires that you bump the major version. Bumping the minor version would have been enough in the previous scheme. Now bumping *any* version is enough: you just have to release `jax` every time you release `jaxlib`.

jax/_src/lib/__init__.py
hawkinsp: It line breaks better like this because the message can start further to the left.

jax/_src/lib/__init__.py
hawkinsp: No, I don't see how you could without doing something like monkey patching the jax versions. And we wouldn't be able to run those tests in pytest because the imports are shared across tests.

jax/_src/lib/__init__.py
hawkinsp: Already done. It's intentionally not matching the whole string, see PEP 440.

jax/_src/lib/__init__.py
hawkinsp: There you go I added a test.

hawkinsp:
PTAL.

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,188 @@
...
+
+* The jaxlib version (`lx.ly.lz`) must be greater than or equal to the minimum
+  jaxlib version (`mx.my.mz`).
+* The jax version (`x.y.z`) must be greater than or equal to the jaxlib version
+  (`lx.ly.lz`).
+
+These constraints imply the following rules for releases:
+* `jax` may be released on its own at any time, without updating `jaxlib`.
+* If a new `jaxlib` is released, a `jax` release with equal or greater version
+  number must be made at the same time.
```
yashk2810, line 58: Can you add a clarifying statement about `equal or greater version number` to what exactly?

Here it should be the last jax version right?

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,188 @@
...
+```
+# 123 is the new version number for _version in xla_client.py
+if jax._src.lib.xla_extension_version >= 123:
+  # Use new code path
+  ...
+else:
+  # Use old code path.
+```
+
+Note that this version number is in *addition* to the constraints on the
```
yashk2810, line 184: Does this doc deserve a section on how we plan to release CUDA jaxlib's too?

I want that documented so we don't keep changing it every now and then. If not here, I can document it somewhere else.

docs/design_notes/jax_versioning.md
hawkinsp: No, it's the newly released jaxlib version. You now *cannot* upload jaxlib wheels without also uploading a new `jax wheel with an equal or greater version number. That is an intentional policy change.

yashk2810: Ok, got it!

Maybe add that as the clarifying statement or restructure so it says that? I got confused when I read it earlier. But other than that LGTM.

docs/design_notes/jax_versioning.md
yashk2810: Yeah, sounds good. I'll do it after you submit this.

yashk2810 (approved):

docs/design_notes/jax_versioning.md
hawkinsp: I reworded it already but possibly pushed it after you last looked.

hawkinsp (commented):
LG to everyone? @skye ? @jakevdp ?

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,189 @@
...
+
+As we will see, distributing `jax` and `jaxlib` separately comes with a cost, in
+that it requires that changes to `jaxlib` maintain a backward compatible API.
+However, we believe that on balance it is preferable to make Python changes
+easy, even if at the cost of making C++ changes slightly harder.
+
+
+## How are `jax` and `jaxlib` versioned?
+
+Summary: `jax`'s version must be greater than or equal to `jaxlib`'s version,
```
skye, line 33:
```suggestion
**Summary:** `jax`'s version must be greater than or equal to `jaxlib`'s version,
```

docs/design_notes/jax_versioning.md
```diff
@@ -0,0 +1,189 @@
...
+* The jaxlib version (`lx.ly.lz`) must be greater than or equal to the minimum
+  jaxlib version (`mx.my.mz`).
+* The jax version (`x.y.z`) must be greater than or equal to the jaxlib version
+  (`lx.ly.lz`).
+
+These constraints imply the following rules for releases:
+* `jax` may be released on its own at any time, without updating `jaxlib`.
+* If a new `jaxlib` is released, a `jax` release whose version is equal to or
+  greater than version the `jaxlib`'s version number must be made at the same
+  time.
```
skye, line 59: Add something about how we usually make the newly-released jax and jaxlib have the same version number, which may result in skipped jaxlib version numbers if interim jax's have been released.

skye (approved):
Thanks Peter!

jax/experimental/compilation_cache/compilation_cache.py
```diff
@@ -93,7 +93,7 @@ def get_cache_key(xla_computation, compile_options, backend) -> str:
   _hash_compile_options(hash_obj, compile_options)
   _log_cache_key_hash(hash_obj, "compile_options")
 
-  hash_obj.update(bytes(jax._src.lib.version))
+  hash_obj.update(jax._src.lib.version_str.encode('utf-8'))
```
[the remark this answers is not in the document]
hawkinsp: Actually I changed my mind and `version` is still present. I just chose to include the unaltered string in the cache key without any parsing.

jakevdp:
This is really great, Peter – thanks for putting it together!

PR #9422, opened 2022-02-03
Add scipy.signal.stft.

### LINKED ISSUE #9411: scipy.signal.stft (and istft)
STFT plays a vital role in audio/ speech signal processing, and it would be nice to have jax-version of `scipy.signal.stft` function.
As Jax already has auto-differentiable FFT and conv1d, it can be basically a utility function that combines those two.

It would be nice if we can have [SciPy-compatible](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.stft.html#scipy.signal.stft) STFT function in Jax.

yotarok:


jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+      if nperseg != win.shape[0]:
+        raise ValueError("value specified for nperseg is different"
+                          " from length of window")
+  return win, nperseg
+
+def _fft_helper(x, win, detrend_func, nperseg, noverlap, nfft, sides):
+  """Internal helper function for calculating windowed FFT.
+
+  This funciton is basically a jax re-implementation of
+  https://github.com/scipy/scipy/blob/47bb6febaa10658c72962b9615d5d5aa2513fa3a/scipy/signal/spectral.py#L1874
```
jakevdp, line 210: If the code is copied & adapted from there, we need to put it in https://github.com/google/jax/tree/main/jax/_src/third_party/scipy

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+      None: lambda x, *args, **kwargs: x
+  }
+
+  if boundary not in boundary_funcs:
+    raise ValueError(
+        f"Unknown boundary option '{boundary}', "
+        f"must be one of: {list(boundary_funcs.keys())}")
+
+  # If x and y are the same object we can save ourselves some computation.
+  same_data = y is x
```
jakevdp, line 298: This is problematic because it makes the function's behavior change when wrapped in JIT:
```python
from jax import jit
import jax.numpy as jnp

def f(x, y): 
  return x is y

x = jnp.arange(4)
print(f(x, x))
# True

print(jit(f)(x, x))
# False
```
We'll need to come up with another strategy for this.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+
+  if not same_data and mode != 'psd':
+    raise ValueError("x and y must be equal if mode is not 'psd'")
+
+  axis = int(axis)
+
+  # Ensure we have np.arrays, get outdtype
+  x = jnp.asarray(x)
```
jakevdp, line 306: We should call `_check_arraylike()` on inputs before converting them to array.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+
+  if not same_data:
+    # Check if we can broadcast the outer axes together
+    xouter = list(x.shape)
+    youter = list(y.shape)
+    xouter.pop(axis)
+    youter.pop(axis)
+    try:
+      outershape = jnp.broadcast_arrays(
+          np.empty(xouter), jnp.empty(youter))[0].shape
```
jakevdp, line 321: We can use `lax.broadcast_shapes` for this, which will avoid creating the broadcasted versions of the arrays.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+                         axis=axis)
+  return ext
+
+
+_spectral_helper_mode = ('psd', 'stft')
+
+def _spectral_helper(x, y, fs=1.0, window='hann', nperseg=None, noverlap=None,
+                     nfft=None, detrend_type='constant', return_onesided=True,
+                     scaling='density', axis=-1, mode='psd', boundary=None,
+                     padded=False):
```
jakevdp, line 272: Any numerical arguments that are required to be concrete should be checked here with `core._concrete_or_error`. I think this would include `nperseg`, `noverlap`, & `nfft` (maybe more?)

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+    except ValueError as e:
+      raise ValueError('x and y cannot be broadcast together.') from e
+
+  if same_data:
+    if x.size == 0:
+      return jnp.empty(x.shape), jnp.empty(x.shape), jnp.empty(x.shape)
+  else:
+    if x.size == 0 or y.size == 0:
+      outshape = outershape + (min([x.shape[axis], y.shape[axis]]),)
+      emptyout = np.rollaxis(np.empty(outshape), -1, axis)
```
jakevdp, line 331: `jnp`?

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  # All imaginary parts are zero anyways
+  if same_data and mode != 'stft':
+    result = result.real
+
+  # Output is going to have new last axis for time/window index, so a
+  # negative axis index shifts down one
+  if axis < 0:
+    axis -= 1
+
+  # TODO: Change to use moveaxis
```
jakevdp, line 482: Outdated TODO?

jakevdp (changes requested):
Thanks for working on this!

Looks pretty good, a couple of minor comments, but one that's more consequential: the `same_data` approach used here is not going to work for JAX. More details below.

jax/_src/scipy/signal.py
yotarok: I think this function itself is sufficiently different from the original one (Jax-reimplementation is non-trivial). However, the `_triage_segments` above is a literal copy, so it is moved to "third_party".

jax/_src/scipy/signal.py
yotarok: Changed `y` to accept None, so one can explicitly indicate to use `same_data` short-cut.

jax/_src/scipy/signal.py
yotarok: changed to use `jnp.broadcast_shapes`.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +148,351 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  elif not hasattr(detrend_type, '__call__'):
+    def detrend_func(d):
+      return detrend(d, type=detrend_type, axis=-1)
+  elif axis != -1:
+    # Wrap this function so that it receives a shape that it could
+    # reasonably expect to receive.
+    def detrend_func(d):
+      d = jnp.moveaxis(d, axis, -1)
+      d = detrend(d)
+      return jnp.moveaxis(d, -1, axis)
```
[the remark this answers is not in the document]
yotarok: It was actually a mistake that the function should call "detrend_type" (not standard `signal.detrend`), and SciPy API doesn't enforce this `detrend` function to have an `axis` argument. So, I think this should be as it is after fixing the above mistake.

jax/_src/scipy/signal.py
yotarok: Applied with adding a check for `axis`.

yotarok:
Primarily for improving the test coverage for the helper function `_spectral_helper`, I also implemented `scipy.signal.csd` in this branch.
If you feel it's out-of-scope for this PR, please let me know so I can exclude this change.

jakevdp:
Looks like you got some unrelated commits on the branch. Can you rebase on `main`?

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,341 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  result *= scale
+
+  if sides == 'onesided' and mode == 'psd':
+    bin_scale = np.ones((1,) * (result.ndim - 1) + (result.shape[-1],))
+    if nfft % 2:
+      bin_scale[..., 1:] = 2.0
+    else:
+      # Last point is unpaired Nyquist freq point, don't double
+      bin_scale[..., 1:-1] = 2.0
+    result *= bin_scale
```
soraros, line 430: Maybe
```suggestion
    end = None if nfft % 2 else -1
    result = result.at[..., 1:end].mul(2)
```

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,341 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+      raise ValueError('x and y cannot be broadcast together.') from e
+
+  # Special cases for size == 0
+  if y is None:
+    if x.size == 0:
+      return jnp.zeros(x.shape), jnp.zeros(x.shape), jnp.zeros(x.shape)
+  else:
+    if x.size == 0 or y.size == 0:
+      outshape = outershape + (min([x.shape[axis], y.shape[axis]]),)
+      emptyout = jnp.moveaxis(np.zeros(outshape), -1, axis)
```
soraros, line 297:
```suggestion
      emptyout = jnp.moveaxis(jnp.zeros(outshape), -1, axis)
```

yotarok: done.

jax/_src/scipy/signal.py
yotarok: Thanks,
I had a wrong impression that a partial update with `at` is slower on TPUs.
I ran some benchmarks on TPUs, and found that the proposed change realized ~20% faster scaling than the original.
So, fixed.

yotarok:
> Looks like you got some unrelated commits on the branch. Can you rebase on `main`?

Thanks for catching this.
Fixed.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+    nfft = nperseg
+  elif nfft < nperseg:
+    raise ValueError('nfft must be greater than or equal to nperseg.')
+  else:
+    nfft = int(nfft)
+
+  if noverlap is None:
+    noverlap = nperseg//2
+  else:
+    noverlap = int(noverlap)
```
jakevdp, line 337: Casting to int already done above?

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+    result = jnp.conjugate(result) * result
+
+  result *= scale
+
+  if sides == 'onesided' and mode == 'psd':
+    end = None if nfft % 2 else -1
+    result = result.at[..., 1:end].mul(2)
+
+  time = jnp.arange(nperseg/2, x.shape[-1] - nperseg/2 + 1,
+                    nperseg - noverlap)/float(fs)
```
jakevdp, line 428: This casting of `fs` to `float` means that `fs` cannot be a traced quantity. We can probably remove that? I suspect this is left over from Python 2.X, where integer division was treated as floor division, so you had to make certain one od the inputs was floating point, but that's no longer necessary in Python 3.

jakevdp (changes requested):

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  axis = operator.index(axis)
+  x = jnp.asarray(x)
+  if y is not None:
+    y = jnp.asarray(y)
+    outdtype = jnp.result_type(x, y, np.complex64)
+  else:
+    outdtype = jnp.result_type(x, np.complex64)
+
+  if y is not None and mode != 'psd':
+    raise ValueError("two-argument mode is available only when mode=='psd'")
```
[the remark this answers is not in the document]
yotarok: In CSD case, mode == 'psd'. Therefore, ValueError will not be raised.  On the other hand, if we call jit(csd)(x, x), we don't get a computation short-cut as we have in "csd(x, x)", but at least the results should be identical.
Might be better if I rephrase the condition statement to be `mode == "stft"`

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  # Apply window by multiplication
+  result = win.reshape((1,) * len(batch_shape) + (1, nperseg)) * result
+
+  # Perform the fft. Acts on last axis by default. Zero-pads automatically
+  if sides == 'twosided':
+    func = jax.numpy.fft.fft
+  else:
+    result = result.real
+    func = jax.numpy.fft.rfft
+  result = func(result, n=nfft)
```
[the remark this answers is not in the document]
yotarok: done.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  right_end = axis_slice(x, start=-1, axis=axis)
+  right_ext = axis_slice(x, start=-2, stop=-(n + 2), step=-1, axis=axis)
+  ext = jnp.concatenate((2 * left_end - left_ext,
+                         x,
+                         2 * right_end - right_ext),
+                         axis=axis)
+  return ext
+
+
+_spectral_helper_mode = ('psd', 'stft')
```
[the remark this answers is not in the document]
yotarok: done.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+                                      "nfft of windowed-FFT")
+  axis = jax.core.concrete_or_error(int, axis,
+                                    "axis of windowed-FFT")
+
+  _check_arraylike("_spectral_helper", x)
+  if x is y:
+    y = None  # For SciPy compatibility (but doesn't work in general with jit)
+  if y is not None:
+    _check_arraylike("_spectral_helper", y)
+  axis = operator.index(axis)
```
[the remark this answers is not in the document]
yotarok: done.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  # parse window; if array like, then set nperseg = win.shape
+  win, nperseg = signal_helper._triage_segments(
+      window, nperseg, input_length=x.shape[-1])
+
+  if nfft is None:
+    nfft = nperseg
+  elif nfft < nperseg:
+    raise ValueError('nfft must be greater than or equal to nperseg.')
+  else:
+    nfft = int(nfft)
```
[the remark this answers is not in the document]
yotarok: done. and moved None parameter completion above for simplifying type validation.

jax/_src/scipy/signal.py
yotarok: done.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+      sides = 'twosided'
+      warnings.warn('Input data is complex, switching to '
+                    'return_onesided=False')
+    else:
+      sides = 'onesided'
+      if y is not None:
+        if np.iscomplexobj(y):
+          sides = 'twosided'
+          warnings.warn('Input data is complex, switching to '
+                        'return_onesided=False')
```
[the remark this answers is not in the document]
yotarok: thanks,  done.

jax/_src/scipy/signal.py
yotarok: thanks for catching that. fixed.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+
+  return result
+
+@_wraps(osp_signal.signaltools.odd_ext)
+def odd_ext(x, n, axis=-1):
+  def axis_slice(a, start=None, stop=None, step=None, axis=-1):
+    a_slice = [slice(None)] * a.ndim
+    a_slice[axis] = slice(start, stop, step)
+    b = a[tuple(a_slice)]
+    return b
```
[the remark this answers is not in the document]
yotarok: done. might be better if we can enhance `lax.slice_in_dim` to support negative strides, so far it is done with combining `jnp.flip`.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,336 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  # Check/ normalize inputs
+  nperseg = jax.core.concrete_or_error(int, nperseg,
+                                       "nperseg of windowed-FFT")
+  noverlap = jax.core.concrete_or_error(int, noverlap,
+                                        "noverlap of windowed-FFT")
+  if nfft is not None:
+    nfft = jax.core.concrete_or_error(int, nfft,
+                                      "nfft of windowed-FFT")
+  axis = jax.core.concrete_or_error(int, axis,
+                                    "axis of windowed-FFT")
```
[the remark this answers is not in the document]
yotarok: `axis` is changed to use operator.index. I'm wondering if we should do the same for other parameters, e.g. nperseg. Unfortunatelly, SciPy API accepts floats, and I can imagine some of existing code is relying on this behavior.

jax/_src/third_party/scipy/signal_helper.py
```diff
@@ -0,0 +1,82 @@
...
+  bias : float
+      Calculated bias.
+  References
+  ----------
+  .. [1] B. Allen, W.G. Anderson, P.R. Brady, D.A. Brown, J.D.E. Creighton.
+          "FINDCHIRP: an algorithm for detection of gravitational waves from
+          inspiraling compact binaries", Physical Review D 85, 2012,
+          :arxiv:`gr-qc/0509116`
+  """
+  ii_2 = 2 * jnp.arange(1., (n-1) // 2 + 1)
```
soraros, line 81:
```suggestion
  ii_2 = jnp.arange(2, n, 2)
```

yotarok: thanks, done.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,341 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+    n : the number of points to be added to the both end
+    axis: the axis to be extended
+  """
+  if n < 1:
+    return x
+  if n > x.shape[axis] - 1:
+    raise ValueError(
+        f"The extension length n ({n}) is too big. "
+        f"It must not exceed x.shape[axis]-1, which is {x.shape[axis] - 1}.")
+  left_end = lax.slice_in_dim(x, start_index=0, limit_index=1, axis=axis)
```
soraros, line 203: Nit: looking at the JAX codebase, people usually call `lax.slice_in_dim` with `start_index` and `limit_index` as positional arguments. Maybe you could do the same here.

yotarok:
Thanks for the comments, I updated the branch according to your comments. I further added `signal.stft.welch` for testing `_spectral_helper` with `mode=='psd'` and `x == y` configuration.
In the new version, `odd_ext` no longer wraps `osp_signal.signaltools.odd_ext` because `signaltools` now looks deprecated, and currently there's no clean way to access this helper function in the original SciPy.

jax/_src/scipy/signal.py
yotarok: done.

jax/_src/scipy/signal.py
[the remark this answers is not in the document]
yotarok: Okay, I think now I understand it's more complicated than I originally thought.
The output dtype of `_spectral_helper` with `mode=='psd'  is different depending on whether `x is y` or not, because of L423 below:
https://github.com/google/jax/commit/43462d8c934bd63ade73aad81167071f38d09d4d#diff-70fbf1556869ea4771d5a99c679f20051a98d0d1dc15629440e95813b0abc468R423
and this issue breaks transparency of JIT.

In fact, the original scipy function has the similar type issue:
```
>>> sp.signal.csd(x, x)[1].dtype
dtype('float64')
>>> sp.signal.csd(x, x.copy())[1].dtype
dtype('complex128')
```
and JIT makes it visible even when `csd(x, y)` is called.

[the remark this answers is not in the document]
yotarok: I think computing CSD with the same variable is anyway an artificial corner case.  So, we can safely assume that `csd` returns complex array.  I'll document this behavior in the docstring.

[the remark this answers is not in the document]
yotarok: Added document, and users can now give `None` as the second parameter of `csd` for explicitly instructing to follow the behavior when `x is y`.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +152,343 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+  freqs, time, Zxx = _spectral_helper(x, None, fs, window, nperseg, noverlap,
+                                      nfft, detrend, return_onesided,
+                                      scaling='spectrum', axis=axis,
+                                      mode='stft', boundary=boundary,
+                                      padded=padded)
+
+  return freqs, time, Zxx
+
+
+_csd_description = textwrap.dedent("""
```
jakevdp, line 448: No need to call `dedent` here, since the string is not indented

tests/scipy_signal_test.py
```diff
@@ -106,5 +106,219 @@ def testDetrend(self, shape, dtype, axis, type, bp):
...
+_stft_test_shapes = [
+    # (input_shape, nperseg, noverlap, axis)
+    ((50,), 17, 5, -1),
+    ((2, 13), 7, 0, -1),
+    ((3, 17, 2), 9, 3, 1),
+    ((2, 3, 389, 5), 17, 13, 2),
+]
+
+
+@jtu.with_config(jax_numpy_rank_promotion="raise")
```
jakevdp, line 118: You can leave this out, as it will be set by default once #9330 is merged. Same below.

tests/scipy_signal_test.py
```diff
@@ -106,5 +106,219 @@ def testDetrend(self, shape, dtype, axis, type, bp):
...
+
+    self._CheckAgainstNumpy(osp_fun, jsp_fun, args_maker, tol=tol)
+    self._CompileAndCheck(jsp_fun, args_maker, rtol=tol, atol=tol)
+
+
+_welch_test_shapes = _stft_test_shapes
+
+
+@jtu.with_config(jax_numpy_rank_promotion="raise")
+class LaxBackedScipyWelchTests(jtu.JaxTestCase):
```
jakevdp, line 275: No need for a new class for each test case – you could just put all these cases within the existing `LaxBasedScipySignalTest` class

jakevdp (changes requested):
Looks great! A couple minor remaining comments. Once you address these, please also squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) and then we can get this merged!

jax/_src/scipy/signal.py
yotarok: done

tests/scipy_signal_test.py
yotarok: those test classes were merged into `LaxBackedScipySignalTests`.
So, done for `LaxBackedScipySignalTests`

yotarok:
Thanks for the review. The commits are squashed, and revised according to the remaining comments.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,343 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+      raise ValueError('x and y cannot be broadcast together.') from e
+
+  # Special cases for size == 0
+  if y is None:
+    if x.size == 0:
+      return jnp.zeros(x.shape), jnp.zeros(x.shape), jnp.zeros(x.shape)
+  else:
+    if x.size == 0 or y.size == 0:
+      outshape = outershape + (min([x.shape[axis], y.shape[axis]]),)
+      emptyout = jnp.moveaxis(jnp.zeros(outshape), -1, axis)
```
soraros, line 297: Could use `_src/util.tuple_insert` here.

yotarok: Same here.  Once `tuple_insert` supported negative indices we should use it here.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,343 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+    outdtype = jnp.result_type(x, y, np.complex64)
+  else:
+    outdtype = jnp.result_type(x, np.complex64)
+
+  if y is not None and mode != 'psd':
+    raise ValueError("two-argument mode is available only when mode=='psd'")
+
+  if y is not None:
+    # Check if we can broadcast the outer axes together
+    xouter = list(x.shape)
```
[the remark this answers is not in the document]
yotarok: `tuple_delete` doesn't support negative indices.  One may resolve negative `axis` in advance in this function, but I think ideally `tuple_delete` should be able to handle negative axes, and enhancing that function is currently out-of-scope.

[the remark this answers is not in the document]
soraros: You actually make a very good point here: we should check that `axis` is not out of bound somewhere.

jax/_src/scipy/signal.py
```diff
@@ -146,3 +151,343 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+
+_csd_description = """
+The original SciPy function exhibits slightly different behavior between
+``csd(x, x)``` and ```csd(x, x.copy())```.  The LAX-backend version is designed
+to follow the latter behavior.  For using the former behavior, call this
+function as `csd(x, None)`."""
+
+
+@_wraps(osp_signal.csd, lax_description=_csd_description)
+def csd(x, y, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
```
soraros, line 455: Currently, calling `cad(x, y)` without specifying `nperseg` gives very obscure error message.

jax/_src/scipy/signal.py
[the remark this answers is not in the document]
yotarok: Thanks for the pointer. Fixed to use `canonicalize_axis` and then `tuple_delete`.

jax/_src/scipy/signal.py
yotarok: done

jax/_src/scipy/signal.py
yotarok: Found that `_triage_segments` must occur first. Reordered and streamlined input verification codes.

yotarok:
Thanks for the comments.  A test case is added for ensuring osp-vs-jsp parity of the default step-size of `_spectral_helper` 
 when `nperseg is None` and/ or `noverlap is None`.

jakevdp (approved):

yotarok:
Update the branch to fix CI errors.

yotarok:
Thanks for the suggestion.
I fixed `csd` and some tests.
While working on it, I had to disable the test of `welch` with `window=='median'` because of a potential bug in the original SciPy function.
The bug is reported here: https://github.com/scipy/scipy/issues/15601

It's also easy to completely mimic the buggy behavior, but I think it's better to fix this bug already in jsp.

jakevdp:
Thanks - I'm still seeing the same failures on CPU; for example:
```
LaxBackedScipySignalTests.testCsdAgainstNumpy_xshape=float32[50]_yshape=float32[13]_average=median_scaling=density_nfft=17_fs=1.0_window=blackman_detrend=constant_nperseg=17_noverlap=5_axis=-1

  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/absl/testing/parameterized.py", [line 318](https://cs.corp.google.com/piper///depot/google3/third_party/py/absl/testing/parameterized.py?l=318&ws=tap-prod-presubmit/172261304&snapshot=2), in bound_param_test
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/tests/scipy_signal_test.py", [line 218](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/tests/scipy_signal_test.py?l=218&ws=tap-prod-presubmit/172261304&snapshot=2), in testCsdAgainstNumpy
    self._CheckAgainstNumpy(osp_fun, jsp_fun, args_maker, rtol=tol, atol=tol)
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/_src/test_util.py", [line 1074](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/test_util.py?l=1074&ws=tap-prod-presubmit/172261304&snapshot=2), in _CheckAgainstNumpy
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/_src/test_util.py", [line 992](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/test_util.py?l=992&ws=tap-prod-presubmit/172261304&snapshot=2), in assertAllClose
    err_msg=err_msg)
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/_src/test_util.py", [line 1000](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/test_util.py?l=1000&ws=tap-prod-presubmit/172261304&snapshot=2), in assertAllClose
    err_msg=err_msg)
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/_src/test_util.py", [line 964](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/test_util.py?l=964&ws=tap-prod-presubmit/172261304&snapshot=2), in assertArraysAllClose
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/jax/_src/test_util.py", [line 185](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/test_util.py?l=185&ws=tap-prod-presubmit/172261304&snapshot=2), in _assert_numpy_allclose
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/numpy/testing/_private/utils.py", [line 1531](https://cs.corp.google.com/piper///depot/google3/third_party/py/numpy/testing/_private/utils.py?l=1531&ws=tap-prod-presubmit/172261304&snapshot=2), in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)
  File "/build/work/a368ba8d79bc5137ccbf72ebbf6c382047d4/google3/runfiles/google3/third_party/py/numpy/testing/_private/utils.py", [line 844](https://cs.corp.google.com/piper///depot/google3/third_party/py/numpy/testing/_private/utils.py?l=844&ws=tap-prod-presubmit/172261304&snapshot=2), in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-05, atol=1e-05

Mismatched elements: 3 / 9 (33.3%)
Max absolute difference: 0.04104351
Max relative difference: 3.385951
 x: array([-2.317960e-01+0.000000e+00j, -1.326986e-01+2.587524e-01j,
        1.084604e-02+4.102368e-02j,  1.647000e-10+1.518764e-09j,
        0.000000e+00+0.000000e+00j,  0.000000e+00+0.000000e+00j,...
 y: array([-2.590589e-01+0.000000e+00j, -1.483060e-01+2.891857e-01j,
        1.212171e-02+0.000000e+00j,  0.000000e+00+0.000000e+00j,
        0.000000e+00+0.000000e+00j,  0.000000e+00+0.000000e+00j,...
```

I wonder if it might have something to do with the scipy version? These tests are run with scipy version 1.2.3

yotarok:
Oops, the last push was fixing this bug in the other way around.
Now, new version should fix the issue. I confirmed that the tests pass on the internal infrastructure.

The bug in question looks like introduced recently (and my code previously was written to simulate that).
Due to this bug, it's difficult to write a concise test that passes on both SciPy 1.2.X and the latest version.
So, workaround is applied that just excludes `mean=="average"` settings.

jakevdp (approved):
Nice - lets run the CI again!

jax/_src/scipy/signal.py
```diff
@@ -146,3 +152,343 @@ def detrend(data, axis=-1, type='linear', bp=0, overwrite_data=None):
...
+
+def _fft_helper(x, win, detrend_func, nperseg, noverlap, nfft, sides):
+  """Calculate windowed FFT in the same way the original SciPy does.
+  """
+  if x.dtype.kind == 'i':
+    x = x.astype(win.dtype)
+
+  # Created strided array of data segments
+  if nperseg == 1 and noverlap == 0:
+    result = x[..., np.newaxis]
```
sdenton4, line 165: FYI: This leaves batch_shape undefined, resulting in an exception at line 181.

yotarok: Thanks for reporting. Fixed.

PR #9426, opened 2022-02-03
[dynamic_shapes] Add support for iota, and reshape (experimental)

gnecula:
This work is to enable using the jax2tf shape polymorphic tests to run with the new support for dynamic shapes. 

For now, we need to add support for iota and reshape. For the latter, JAX includes static shape checks, which we don't yet know how to handle.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -840,7 +840,7 @@ bcoo_cusparse_lowering = config.define_bool_state(
 # if the intended backend can handle lowering the result
 config.define_bool_state(
     name='jax_dynamic_shapes',
-    default=False,
+    default=bool(os.getenv('JAX_DYNAMIC_SHAPES', '')),
     help=('Enables experimental features for staging out computations with '
           'dynamic shapes.'),
     update_global_hook=lambda val: \
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -302,7 +302,6 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
               shape=tuple(expected_shape), dtype=expected_type.dtype,
               weak_type=expected_type.weak_type)
           assert core.typematch(expected_aval, aval)
-
   with log_elapsed_time(f"Finished tracing + transforming {fun.__name__} "
                         "for jit in {elapsed_time} sec"):
     jaxpr, out_type, consts = pe.trace_to_jaxpr_final2(
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -302,6 +302,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
               shape=tuple(expected_shape), dtype=expected_type.dtype,
               weak_type=expected_type.weak_type)
           assert core.typematch(expected_aval, aval)
+
   with log_elapsed_time(f"Finished tracing + transforming {fun.__name__} "
                         "for jit in {elapsed_time} sec"):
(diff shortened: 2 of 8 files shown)


jax/experimental/jax2tf/shape_poly.py
```diff
@@ -649,17 +649,17 @@ def need_dim_var_msg():
 
 
 def _add(v1, v2):
-  if isinstance(v1, int) and v1 == 0:
```
gnecula, line 652: I suspect these changes are not needed anymore; will check

jax/interpreters/partial_eval.py
```diff
@@ -1921,7 +1921,9 @@ def trace_to_subjaxpr_dynamic2(
     out_tracers = map(trace.full_raise, ans)
     jaxpr, out_type, consts = frame.to_jaxpr2(out_tracers)
     del fun, main, trace, frame, in_tracers, out_tracers, ans
-  config.jax_enable_checks and core.check_jaxpr(jaxpr)
+  # TODO(necula): re-enable this, disabled because check_jaxpr does shape
```
gnecula, line 1924: I think that we need to add typechecking rules for some of the primitives

jax/interpreters/partial_eval.py
```diff
@@ -2040,7 +2042,7 @@ def _collect_implicit(
     for i, name in spec.items():
       if name not in idxs and id(x.shape[i]) not in explicit_tracers:
         idxs[name] = next(counter)
-        implicit_types.append(core.get_aval(x.shape[i]))
```
gnecula, line 2043: This is part of #11179

jax/core.py
```diff
@@ -1633,10 +1633,6 @@ def symbolic_equal_shape(s1: Shape, s2: Shape) -> bool:
           all(unsafe_map(symbolic_equal_dim, s1, s2)))
 
 def greater_equal_dim(d1: DimSize, d2: DimSize) -> bool:
-  # TODO(mattjj): revise this temporary workaround for dynamic shapes
```
gnecula, line 1636: I could not find a test that requires this. I believe that the right place to make this change is in partial_eval.DimensionHandlerTracer.greater_equal, and then the returned value should be False.

jax/_src/config.py
```diff
@@ -827,7 +827,7 @@ def _update_disable_jit_thread_local(val):
 # if the intended backend can handle lowering the result
 config.define_bool_state(
     name='jax_dynamic_shapes',
-    default=False,
+    default=bool(os.getenv('JAX_DYNAMIC_SHAPES', 'True')),  # TODO(necula): remove default
```
gnecula, line 830: Will remove

mattjj (approved):
This looks great, thanks so much!

jax/_src/dispatch.py
```diff
@@ -283,8 +283,28 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
...
+                     for d1, d2 in zip(expected_type.shape, aval.shape)
+                     if type(d1) is core.DBIdx)
+          # Check the type matches after substitution.
+          expected_shape = [axis_sizes.get(d, d) for d in expected_type.shape]  # type: ignore[call-overload]
+          expected_aval = core.ShapedArray(
+              shape=tuple(expected_shape), dtype=expected_type.dtype,
+              weak_type=expected_type.weak_type)
+          assert core.typematch(expected_aval, aval)
+
+
```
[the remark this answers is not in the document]
mattjj: Actually, let's just do this in follow-ups.

PR #9437, opened 2022-02-03
DOC: move functions in jax.html to their own pages

### LINKED ISSUE #9435: Make Jax's API searchable
Currently if I search for jax.vmap on google search, I am not directed directly to that point in the big page.

This is very inconvenient because I have to search for vmap again on the big page. It would be good to split each API endpoint into its own page (as @hawkinsp suggested) to make it more searchable.

jax.numpy packages have their own pages and if you search for `jax.numpy.argmax`, you are directed to that page!

jakevdp:
Fixes #9435

See the rendered docs here: https://jax--9437.org.readthedocs.build/en/9437/jax.html

### DIFF
--- a/docs/jax.experimental.ann.rst
+++ b/docs/jax.experimental.ann.rst
@@ -6,5 +6,8 @@ jax.experimental.ann module
 API
 ---
 
-.. autofunction:: approx_max_k
-.. autofunction:: approx_min_k
+.. autosummary::
+   :toctree: _autosummary
+
+   approx_max_k
+   approx_min_k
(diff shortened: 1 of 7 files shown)


yashk2810:
Can you do jax.experimental too? https://github.com/google/jax/blob/main/docs/jax.experimental.rst

Or will you have to do each file within jax.experimental.*rst?

jakevdp:
Ugh, I see why we didn't do this before.

jakevdp:
OK, I think I got it. PTAL – I like this much better!

yashk2810 (approved):

Issue #9434, opened 2022-02-03
jax.readthedocs.io API docs doesn't have right nav


yashk2810:
It would be really nice if we have a right nav like what TF has: https://www.tensorflow.org/api_docs/python/tf/Tensor

jakevdp:
I don't think this is possible unless we move away from using the readthedocs theme: https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html

yashk2810:
Which theme has that feature?

I have found a right nav super helpful when the page is long and you don't know what's on it.

You can just look at the right nav and see all the methods/attributes/args, etc

jakevdp:
I don't know if there are any themes that have that feature, but it doesn't appear that the readthedocs theme does.

jakevdp:
Just for clarification: what is it that the right nav gives you that you can't get from the left nav?

yashk2810:
Yeah that would work too. But then there is the point of having too deep of a left nav. You can't list every method/arg/attribute in a left nav (I mean you can, but there's a tradeoff).

PR #9440, opened 2022-02-03
Use the sphinx-book-theme for JAX documentation.

hawkinsp:
Improves #9434 (there may still be work to do on the page structure to have a usable right navigation).


jakevdp (approved):
Looks great!

PR #9456, opened 2022-02-05
add flag for jaxpr colorful syntax highlighting

mattjj:
@hawkinsp added all the cool stuff here, this PR is just turning it on by default in ipython and jupyter notebooks, and making it easier to switch on manually!

Also use real colors, not just intensity.

image

image

---

TODO:
* [x] detect if output is a terminal emulator or ipython session (and not e.g. `less`), and only print colors by default then

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -441,6 +441,12 @@ flags.DEFINE_integer(
 )
 
 flags.DEFINE_bool(
+    'jax_pprint_use_color',
+    bool_env('JAX_PPRINT_USE_COLOR', False),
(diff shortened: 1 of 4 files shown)


mattjj:
@soraros wow awesome, thanks for the tip!

zhangqiaorjc (approved):

carlosgmartin:
@mattjj Is there an equivalent for `jax._src.stages.Lowered` and `jax._src.stages.Compiled`? I see colorful syntax highlighting for `jax._src.core.ClosedJaxpr`, but not the former.

PR #9459, opened 2022-02-05
Added jax.blocking

chaserileyroberts:
This methods takes any JAX method and turns it into a blocking version. Can be thought of as a more functional version of `block_until_ready`


NeilGirdhar:
Why not use a type variable so that the type of the function the decorator applies to doesn't get lost?

Thenerdstation:
That's a good point. I will add that.

PR #9461, opened 2022-02-06
Expand implementation of lax_numpy._quantile to allow the input of a …

### LINKED ISSUE #9277: jax.numpy.nanpercentile with axis as tuple
currently, the following code results in a NotImplementedError:
```
jax.numpy.nanpercentile(jnp.ones((3,3,12,24)), 99.9, axis=(0,1,2))
```

```
 File "/home/clemens/.local/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 6359, in _quantile                                       
    raise NotImplementedError("Tuple values for axis are not implemented")   
```
However the docstring actually explicitly mentions tuple as a valid option for axis: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.nanpercentile.html

same for jax.numpy.percentile.html

MichaelMarien:
Expand implementation of lax_numpy._quantile to allow the input of a tuple as axis argument. Fixes #9277 

- introduced and changed corresponding numpy code to swap and reshape axis to be quantiled
- introduced code to accomodate the reintroduction of those axes if keepdims=True
- added testcases

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -6406,13 +6406,31 @@ def _quantile(a, q, axis, interpolation, keepdims, squash_nans):
     raise ValueError("interpolation can only be 'linear', 'lower', 'higher', "
                      "'midpoint', or 'nearest'")
   a, q = _promote_dtypes_inexact(a, q)
+  keepdim = []
   if issubdtype(a.dtype, np.complexfloating):
     raise ValueError("quantile does not support complex input, as the operation is poorly defined.")
   if axis is None:
     a = ravel(a)
     axis = 0
   elif isinstance(axis, tuple):
-    raise NotImplementedError("Tuple values for axis are not implemented")
+    keepdim = list(shape(a))
+    nd = ndim(a)
+    axis = tuple([_canonicalize_axis(ax, nd) for ax in axis])
+    if len(set(axis)) != len(axis):
+        raise ValueError('repeated axis')
+    for ax in axis:
+      keepdim[ax] = 1
+
+    keep = set(range(nd)) - set(axis)
+    # prepare permutation
+    dimensions = list(range(nd))
+    for i, s in enumerate(sorted(keep)):
+      dimensions[i], dimensions[s] = dimensions[s], dimensions[i]
+    do_not_touch_shape = tuple(x for idx,x in enumerate(shape(a)) if idx not in axis)
+    touch_shape = tuple(x for idx,x in enumerate(shape(a)) if idx in axis)
+    a = lax.reshape(a, do_not_touch_shape + (int(np.prod(touch_shape)),), dimensions)
+    keepdim = tuple(keepdim)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -6406,13 +6406,29 @@ def _quantile(a, q, axis, interpolation, keepdims, squash_nans):
...
   elif isinstance(axis, tuple):
-    raise NotImplementedError("Tuple values for axis are not implemented")
+    keepdim = list(shape(a))
+    nd = ndim(a)
+    axis = tuple([_canonicalize_axis(ax, nd) for ax in axis])
+
+    for ax in axis:
+      keepdim[ax] = 1
+
+    keep = set(range(nd)) - set(axis)
```
jakevdp, line 6423: Numpy errors on repeated axes. Perhaps JAX should do this too?

jax/_src/numpy/lax_numpy.py
```diff
@@ -6406,13 +6406,29 @@ def _quantile(a, q, axis, interpolation, keepdims, squash_nans):
...
+    nkeep = len(keep)
+    # swap axis that should not be reduced to front
+    for i, s in enumerate(sorted(keep)):
+      a = a.swapaxes(i, s)
+    # merge reduced axis
+    a = reshape(a, shape(a)[:nkeep] + (-1,))
```
jakevdp, line 6429: swapping axes and reshaping can be somewhat expensive. I think `lax.reshape` is general enough to do this axis swapping plus the reshape in a single XLA call: we should use that here.

jakevdp (changes requested):
Looks great, thanks for taking this on! Just one substantial comment below, regarding a more efficient implementation.

jax/_src/numpy/lax_numpy.py
MichaelMarien: Agreed!

jax/_src/numpy/lax_numpy.py
MichaelMarien: Totally, did not know about the existence of the `dimensions` arg in `lax.reshape`.

MichaelMarien:
@jakevdp tried to address your comments, let me know what you think and thanks for the review effort.

jakevdp (approved):
Looks awesome, thanks for all the quick responses on the feedback!

Last thing: could you please squash all the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests). Once you do, we can start the merge process

MichaelMarien:
> Last thing: could you please squash all the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests). Once you do, we can start the merge process

Made a bit of a git mess at first, but I think it's fine now.

jakevdp (approved):

jakevdp:
Thanks for the contribution!

PR #9469, opened 2022-02-07
Recommend optax in jax.experimental_libraries.optimizers documentation.

hawkinsp:


jax/example_libraries/optimizers.py
```diff
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Optimizers for use with JAX.
+"""Examples of how to write optimizers with JAX.
+
+This is probably not what you are looking for! The optimizers in this library
```
froystig, line 17: Thoughts on this tweak?
```suggestion
You likely do not mean to import this module! The optimizers in this library
```

froystig (approved):

jax/example_libraries/optimizers.py
hawkinsp: Done.

PR #9475, opened 2022-02-07
DOC: add ability to document extra parameters within _wraps

jakevdp:
This should allow us to improve documentation of jax-only parameters in wrapped functions.

Example:

<details>
<summary>argwhere docstring before:</summary>

```
$ python -c "import jax.numpy; print(jax.numpy.argwhere.__doc__)"
Find the indices of array elements that are non-zero, grouped by element.

LAX-backend implementation of :func:`argwhere`.

Because the size of the output of ``argwhere`` is data-dependent, the function is not
typically compatible with JIT. The JAX version adds the optional ``size`` argument, which
specifies the size of the leading dimension of the output - it must be specified statically
for ``jnp.argwhere`` to be compiled with non-static operands. If ``size`` is specified,
the indices of the first ``size`` True elements will be returned; if there are fewer
nonzero elements than `size` indicates, the index arrays will be zero-padded.

*Original docstring below.*

Parameters
----------
a : array_like
    Input data.

Returns
-------
index_array : (N, a.ndim) ndarray
    Indices of elements that are non-zero. Indices are grouped by element.
    This array will have shape ``(N, a.ndim)`` where ``N`` is the number of
    non-zero items.
```

</details>
<details>
<summary>argwhere docstring after:</summary>

```
$ python -c "import jax.numpy; print(jax.numpy.argwhere.__doc__)"
Find the indices of array elements that are non-zero, grouped by element.

LAX-backend implementation of :func:`argwhere`.

Because the size of the output of ``argwhere`` is data-dependent, the function is not
typically compatible with JIT. The JAX version adds the optional ``size`` argument which
must be specified statically for ``jnp.argwhere`` to be used within JAX transforms.

*Original docstring below.*

Parameters
----------
a : array_like
    Input data.
size : int, optional
    If specified, the first ``size`` unique elements will be returned. If there are fewer unique
    elements than ``size`` indicates, the return value will be padded with ``fill_value``.
fill_value : array_like, optional
    When ``size`` is specified and there are fewer than the indicated number of elements, the
    remaining elements will be filled with ``fill_value``, which defaults to zero.

Returns
-------
index_array : (N, a.ndim) ndarray
    Indices of elements that are non-zero. Indices are grouped by element.
    This array will have shape ``(N, a.ndim)`` where ``N`` is the number of
    non-zero items.
```

</details>

New rendered version: https://jax--9475.org.readthedocs.build/en/9475/_autosummary/jax.numpy.argwhere.html

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1987,15 +1987,19 @@ def in1d(ar1, ar2, assume_unique=False, invert=False):
   else:
     return (ar1[:, None] == ar2[None, :]).any(-1)
 
-_SETDIFF1D_DOC = """\
-Because the size of the output of ``setdiff1d`` is data-dependent, the function is not
-typically compatible with JIT. The JAX version adds the optional `size` argument which
-specifies the size of the output array: it must be specified statically for ``jnp.setdiff1d``
-to be compiled with non-static operands. If specified, the first `size` unique elements will
-be returned; if there are fewer unique elements than `size` indicates, the return value will
-be padded with the `fill_value`, which defaults to zero."""
-
-@_wraps(np.setdiff1d, lax_description=_SETDIFF1D_DOC)
+@_wraps(np.setdiff1d,
+  lax_description=_dedent("""
+    Because the size of the output of ``setdiff1d`` is data-dependent, the function is not
+    typically compatible with JIT. The JAX version adds the optional ``size`` argument which
+    must be specified statically for ``jnp.setdiff1d`` to be used within some of JAX's
+    transformations."""),
+  extra_params=_dedent("""
+    size : int, optional
+        If specified, the first ``size`` elements of the result will be returned. If there are
+        fewer elements than ``size`` indicates, the return value will be padded with ``fill_value``.
+    fill_value : array_like, optional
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1986,15 +1986,18 @@ def in1d(ar1, ar2, assume_unique=False, invert=False):
...
-to be compiled with non-static operands. If specified, the first `size` unique elements will
-be returned; if there are fewer unique elements than `size` indicates, the return value will
-be padded with the `fill_value`, which defaults to zero."""
-
-@_wraps(np.setdiff1d, lax_description=_SETDIFF1D_DOC)
+@_wraps(np.setdiff1d,
+  lax_description=_dedent("""
+    Because the size of the output of ``setdiff1d`` is data-dependent, the function is not
+    typically compatible with JIT. The JAX version adds the optional ``size`` argument which
+    must be specified statically for ``jnp.setdiff1d`` to be used within JAX transforms."""),
```
froystig, line 1993: How about:
```suggestion
    must be specified statically for ``jnp.setdiff1d`` to be used within some of JAX's transformations."""),
```
and the same for similar instances below? Transformations like JVP take no issue.

froystig (approved):

PR #9493, opened 2022-02-08
improve pjit in/out_axis_resources pytree errors

### LINKED ISSUE #8996: Better `pjit in_axis_resources specification must be a tree prefix of the corresponding value` error

mattjj:
Fixes #8996.

This is an application to `pjit` of the utilities added in #9372. We still want to apply them to `xmap`, `pmap`, and `vmap` too.

Have you ever tried to use `pjit`'s `in_axes_resources` or `out_axis_resources`, or `vmap`'s `in_axes` or `out_axes`, only to get a giant error message about a pytree mismatch?

And I mean really giant: **here's [a real example](https://gist.github.com/mattjj/bb7ba55f79789ccc6739f6c8116e6eaa)**. What are you supposed to do with an error message like that?

These error messages were gigantic because we just printed two full pytrees, even though pytrees for neural net models can be huge, and even though it might be just one buried pytree node that disagrees. #9372 added utilities to find disagreements between pytrees and report to them more precisely. By using those utilities with `pjit`, we can get more actionable errors.

Here's the new error message on that same real example:
```
ValueError: pytree structure error: different pytree metadata at key path
    pjit out_axis_resources.opt_states[0][2].stats.local_stats['lm']['final_ln']['bias'].diagonal_statistics
At that key path, the prefix pytree pjit out_axis_resources has a subtree of type
    <class 'scalable_shampoo.optax.distributed_shampoo.QuantizedValue'>
with metadata
    (<class 'jax._src.numpy.lax_numpy.float32'>, False, [32])
but at the same key path the full pytree has a subtree of the same type but with metadata
    (<class 'jax._src.numpy.lax_numpy.float32'>, False, (32,))
so the diff in the metadata at these pytree nodes is
    - (<class 'jax._src.numpy.lax_numpy.float32'>, False, [32])
    ?                                                     ^  ^

    + (<class 'jax._src.numpy.lax_numpy.float32'>, False, (32,))
    ?                                                     ^  ^ +
```

(Actually, to get that error message with the full key path we need to submit two-liner updates to Flax and Lingvo. I'll do those too!)

cc @skye 

**Notes for reviewer**:
* The changes in _src/tree_util.py are just (1) tweaks to whitespace in printed errors, and (2) plumbing an optional `is_leaf` function through `prefix_errors` (even though we don't use that argument in this PR);
* Most of the changes in tree_util_test.py are compensating for the whitespace changes mentioned above;
* The changes in pjit_test.py are (1) rephrasing of existing error messages and (2) a few new cases;
* The changes in pjit.py are mainly about calling the `prefix_errors` function, though they also (1) clarified the error-raising logic a bit and (2) don't rely on the `flatten_axes` utility to raise errors anymore (indeed I hope to remove all callers of that function as we move `vmap`, `xmap`, and `pmap` to a pattern more like the one in this PR).

### DIFF
--- a/jax/experimental/pjit.py
+++ b/jax/experimental/pjit.py
@@ -40,7 +40,9 @@ from jax.interpreters import batching
 from jax.interpreters import partial_eval as pe
 from jax.interpreters.sharded_jit import PartitionSpec
 from jax._src.lib import xla_client as xc
-from jax.tree_util import tree_map, tree_flatten, tree_unflatten
+from jax.tree_util import (tree_map, tree_flatten, tree_unflatten,
+                           treedef_is_leaf, tree_structure)
+from jax._src.tree_util import prefix_errors
 from jax._src.util import (extend_name_stack, HashableFunction, safe_zip,
                          wrap_name, wraps, distributed_debug_log,
                          split_list, cache, tuple_insert)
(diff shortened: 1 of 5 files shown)


jax/experimental/pjit.py
```diff
@@ -293,11 +295,60 @@ def flatten_axis_resources(what, tree, axis_resources, tupled_args):
...
+  errors = prefix_errors(axis_tree, dummy_tree)
+  if errors:
+    e = errors[0]  # Only show information about the first disagreement found.
+    raise e(what)
+
+  # At this point we've failed to find a tree prefix error.
+  assert False, "Please open a bug report!"  # This should be unreachable.
+
+class PytreeLeaf:
+  def __repr__(self): return "pytree leaf"
```
apaszke, line 350: Those things should never show up in the error messages, right? If there's a mismatch then it has to be a type mismatch that should be caught before we get to leaves. Or is my reasoning incorrect?

apaszke (approved):
Looks great! This is awesome!

jax/experimental/pjit.py
```diff
@@ -293,11 +295,60 @@ def flatten_axis_resources(what, tree, axis_resources, tupled_args):
...
-    pass
+    pass  # Raise a tree prefix error below
+
+  # Tree leaves are always valid prefixes, so if there was a prefix error as
+  # assumed here, axis_resources must not be a leaf.
+  assert not treedef_is_leaf(tree_structure(axis_resources))
+
+  # Check the type directly rather than using isinstance because of namedtuples.
+  if tupled_args and (type(axis_resources) is not tuple or
+                      len(axis_resources) != len(tree.children())):
```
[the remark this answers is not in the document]
mattjj: That's a good point. I'd like to revise this but I think I'll do it in follow-up work, the main reason being that I've got a lot of plates spinning and I'm worried I might delay merging this too long (like I already have!).

jax/experimental/pjit.py
mattjj: That's my understanding yes: these are internal-implementation-only dummies should never show up anywhere. I meant to test this more carefully...

mattjj:
Also, I realized this fix isn't quite done: it won't be as useful until I submit a fix to Flax as well. I'll make a PR for that now...

PR #1224, opened 2019-08-21
No more tuples

mattjj:
This PR removes tuples from the jaxpr language, and JaxTuples / AbstractTuples / DeviceTuples / ShardedDeviceTuples from the implementation. Instead of tuples, jaxprs and primitives have multiple outputs.

Here are the main reasons for removing tuples:
1. every tracer needed its own transparent tuple type (e.g. JaxprTracerTuple, TangentTuple) so that packing into a tuple with a traced value didn't irreversibly lift quantities into a trace (e.g. so that tuple pack/unpack didn't cause broadcasting of unmapped values for a vmap trace),
1. allowing jaxpr variables to be bound to tuples meant types (e.g. for linearity, both in the AD sense and the PRNG sense) and lattice representations (e.g. for the initial-style analog of the batched/unbatched problem above) were more complex and the logic dealing with them had recursions everywhere,
1. relatedly, to support `lax.scan`'s mixed linear/nonlinear extensive arguments we had to introduce `pat_fmap` pattern destructuring and `JaxprEqn.restructure`, which polluted all jaxprs and jaxpr interpreters,
1. tuple munging wasn't actually a good way to do bookkeeping in e.g. control flow primitives and was leading to extra jaxpr munging work,
1. tuples basically doubled the number of data types we had to support (e.g. with device-backed versions), including in things like dispatch logic,
1. no user ever wanted to see or learn about JaxTuples, yet it was one of the first things they would encounter in api.py.

So this PR gets rid of them! We also took the opportunity to:
* revise control flow (including a more conservative and efficient vmap-of-while rule, a simpler implementation of `cond`, a removal of all jaxpr eqn munging, better compilation caching, and more modular code)
* make xla.py and pxla.py support the introduction of user-defined types like the rest of the system (to be used in follow-up work on PRNG key linearity checking)
* a cleanup of batching.py utility functions
* de-register None as a pytree so that users can handle it as they like (and e.g. use it as their own sentinel value).

I checked for performance regressions on `pmap` using a benchmark from @hawkinsp and @ibab; in 2 out of my 3 runs it went from ~15.5ms to ~14.5ms, and in the third run they both were ~15ms, so I don't think this presents a significant performance regression (it may already be faster, and in any case I'm optimistic it opens up more opportunities for improvement).

Co-authored by @dougalm, particularly the hard control flow parts. @dougalm also got us started by de-tupling jaxprs and ad.py.

Two things that will probably come in smaller follow-up PRs:
[...]

PR #9498, opened 2022-02-09
[remove units] make JaxprTrace.process_call not introduce units

mattjj:
This is the first in a sequence of changes to remove units from jax and jaxprs.

Units are trivial values which inhabit a trivial (single-member) type. We mainly use them as placeholders during partial evaluation: for example, when we partially evaluate, we use units on the 'known' side of a computation to represent unknown values, and analogously units on the 'unknown' side of the computation stand in for known inputs:

```
# jaxpr to be partially evaluated
{ lambda a b c.
  let ...
  in (d, e, f, g) }

# if we evaluate with the input unknowns as [False, True, True], then
# we might get a known jaxpr like
{ lambda a * *.
  let ...
  in (d, e, *, *) }

# and an unknown jaxpr like
{ lambda * b c.
  let ...
  in (*, *, f, g) }
```

In the above jaxpr-pseudocode examples, the `*` symbols represent unit-typed input binders and unit-literal outputs. Notice that the pattern of units between the known and unknown sides, both on inputs and outputs, is complementary. (In general, the known jaxpr also gains additional outputs, and the unkonwn jaxpr gains corresponding additional inputs, representing 'residuals'. I didn't include residuals in the above examples just because they're not relevant to the discussion.)

By inserting trivial placeholder values this way, input and output lists stay the same length, which handles some bookkeeping for us in a simple and relatively error-avoiding way.

But over time we came to think units were not so convenient. In particular, everywhere in the JAX system had to handle them; even though handling them was usually simple, it was another case to consider. Moreover, if we were to design units from scratch, we'd make them more like base types, and allow arrays of units. But in the current implementation we can only have single unit values and not arrays of units; that leads to some weird behaviors like being able to scan or map over a single unit no matter the scanned axis size. (That behavior makes some sense if we think of units as empty tuples, and we think of scan as 'tree-fmapping over' tuple trees; indeed units were introduced around the time of #1224 specifically because we were removing tuples from jaxprs, and units took the place of empty tuples.)

And if you've ever wondered why JAX always adds a builds HLO computations with a `pred[] constant(false)` in them... it's because of units!

So we'd like to remove units from JAX. Since they have to be handled everywhere, it's not a small task. I'm going to attempt it incrementally, removing places that add units, one by one, until nothing in the system introduces them anymore. Then we should be able to just delete all the sites which consume them as well.

While we're touching these other parts of the code, I'm going to try to revise them for clarity, and even add a few comments where appropriate.

This first PR starts by adapting `JaxprTrace.process_call` to avoid ever introducing units. It so happens that `JaxprTrace.process_call` builds jaxprs with units _and then attempts to remove them via steps like DCE_. This change avoids introducing them in the first place.

Off the top of my head, the follow-up steps are:
1. prevent `JaxprTrace.process_map` (handled in #9500), `process_xmap`, `process_custom_jvp_call`, and `process_custom_vjp_call` from introducing any units;
2. prevent old `remat` from introducing units (the new `remat` in ad_checkpoint.py already avoids units), or delete it (though deleting it in favor of new `remat` would be easier if we didn't have units...);
3. prevent each initial-style higher-order primitive's partial evaluation rule from introducing units (`cond`, `while_loop`, `scan`, ...).

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -33,7 +33,7 @@ from jax._src.ad_util import Zero
 from jax._src.api_util import flattened_fun_in_tree
 from jax._src.tree_util import PyTreeDef, tree_unflatten, tree_leaves
 from jax._src.util import (unzip2, safe_zip, safe_map, toposort, split_list,
-                           partition_list, cache, OrderedSet,
+                           merge_lists, partition_list, cache, OrderedSet,
                            as_hashable_function)
 from jax.core import (Trace, Tracer, Jaxpr, Literal, get_aval, AbstractValue,
                       unit, unitvar, abstract_unit, ClosedJaxpr, new_jaxpr_eqn,
@@ -93,6 +93,7 @@ class PartialVal(tuple):
     """Either the stored known value, or the given 'val'."""
     known = self.get_known()
     return known if known is not None else val
+    breakpoint()
 
 
 class JaxprTrace(Trace):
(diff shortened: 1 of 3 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -1437,10 +1483,13 @@ def process_call(self, call_primitive, f, tracers, params):
     new_params = dict(params, call_jaxpr=convert_constvars_jaxpr(jaxpr))
     update_params = call_param_updaters.get(call_primitive)
     if update_params:
-      # TODO(mattjj): update donated_invars to work with axis size variables
-      new_params = update_params(new_params, [True] * len(tracers))
+      new_params = update_params(new_params, [True] * len(tracers),
+                                 len(consts) + len(dim_tracers))
     eqn = new_jaxpr_eqn([*constvars, *invars], outvars,
                         call_primitive, new_params, source_info)
+    # TODO TODO remove
```
froystig, line 1490: Is this comment intentional and meant to outlast this change?

froystig (approved):
Very nice!

jax/interpreters/partial_eval.py
mattjj: Whoops, thanks for the catch.

PR #9526, opened 2022-02-10
tests: use lax.broadcast_shapes in place of custom logic

jakevdp:
This is a minor cleanup that avoids allocation of temporary arrays in order to reason about their shapes


apaszke (approved):
Can't say I see anything wrong here, but I also wasn't even aware of the existence of `scalar_shapes`, so I might have missed something

Issue #1328, opened 2019-09-09
jax.numpy functions `out` argument fails


mgbukov:
It looks like the `out` argument in `jax.numpy` functions does not work.

Is there a deep reason why `out`-like arguments cannot be used on the GPU? If yes, it might be worth adding a note to the [jax documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.exp.html), to avoid confusion.  

Here's a minimum script in comparison to numpy:
```
import jax
import jax.numpy as jnp
import numpy as np

a_jnp=jnp.zeros(2)
a_np=np.zeros(2)

b=jnp.ones(2)

np.exp(b,out=a_np)
print('numpy result:', a_np)

jnp.exp(b,out=a_jnp)
print('jax result:', a_jnp)
```

which runs thru the numpy code but when it gets to jax.numpy it throws the error:
```
TypeError: <lambda>() got an unexpected keyword argument 'out'
```

mattjj:
Thanks for asking! It's mentioned in the [What's supported section of the readme](https://github.com/google/jax#whats-supported) under "Don't use" ("The `out` argument of NumPy functions").

The docstrings are just forwarded from NumPy, so without having some code to automatically edit them (which seems brittle) I'm not sure we can really fix up the docstrings unless we write our own by hand.

I think we might just consider this "working as intended", though this could be a place where contributions are welcome to improve the documentation.

What do you think?

mgbukov:
I see. Another possibility is to throw an error message which explicitly says that the `out` argument is not available, unlike the numpy version of the function, and refer to the What's supported sec online. 

On the technical level -- what's the underlying reason why `out`-arguments are not suitable for GPU computing?

rajasekharporeddy:
Hi @mgbukov 
I executed the mentioned code with the latest version of JAX(0.4.23) on. Google Colab. Now the error message indicates that `jax.numpy.exp() got an unexpected keyword argument 'out'`. Could you please verify this behavior with JAX 0.4.23 and confirm closure of this issue once resolved?
```python
import jax
import jax.numpy as jnp
import numpy as np

a_jnp=jnp.zeros(2)
a_np=np.zeros(2)

b=jnp.ones(2)

np.exp(b,out=a_np)
print('numpy result:', a_np)

jnp.exp(b,out=a_jnp)
print('jax result:', a_jnp)
```
Output:
```python
numpy result: [2.71828198 2.71828198]
---------------------------------------------------------------------------
[...]

PR #9529, opened 2022-02-10
jax.numpy: add ufunc-style APIs for several jnp functions

jakevdp:
This adds a `ufunc` object to the `jax.numpy` API, so that standard unary and binary functions have the extra methods exposed by numpy ufuncs. Example:
```python

In [2]: x = jnp.arange(5)

In [3]: indices = jnp.array([1, 1, 3])

In [4]: jnp.add(x, 1)  # Standard function call works
Out[4]: DeviceArray([1, 2, 3, 4, 5], dtype=int32)

In [5]: jnp.add.reduce(x)  # reduce() method applies binary reduction.
Out[5]: DeviceArray(10, dtype=int32)

In [6]: jnp.add.accumulate(x)  # accumulate() method is cumulative reduction
Out[6]: DeviceArray([ 0,  1,  3,  6, 10], dtype=int32)

In [7]: jnp.add.at(x, indices, 10, inplace=False)  # at() method is similar to JAX's ndarray.at
Out[7]: DeviceArray([ 0, 21,  2, 13,  4], dtype=int32)

In [8]: jnp.add.reduceat(x, indices)  # reduceat() is similar to JAX's segment_sum
Out[8]: DeviceArray([0, 3, 7], dtype=int32)
```

I'll still need to give the ufunc treatment to the rest of the functions in `jax._src.numpy.ufuncs`, but this gives us a good start.

### DIFF
--- a/tests/lax_numpy_ufunc_test.py
+++ b/tests/lax_numpy_ufunc_test.py
@@ -1,0 +1,312 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import collections
+import functools
+import itertools
+
+from absl.testing import absltest
+from absl.testing import parameterized
+import numpy as np
+
+import jax
+from jax import lax
+from jax import tree_util
+from jax._src import test_util as jtu
+from jax._src.numpy.lax_numpy import _promote_dtypes, _promote_dtypes_inexact
+import jax.numpy as jnp
+
+from jax.config import config
+config.parse_flags_with_absl()
+FLAGS = config.FLAGS
+
+nonempty_nonscalar_array_shapes = [(4,), (3, 4), (3, 1), (1, 4), (2, 1, 4), (2, 3, 4)]
+nonempty_array_shapes = [()] + nonempty_nonscalar_array_shapes
+one_dim_array_shapes = [(1,), (6,), (12,)]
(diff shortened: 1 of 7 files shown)


jakevdp:
Note to self: we could also probably raise better errors for unsupported function arguments, cf. #1328 and #1364

tests/lax_numpy_ufunc_test.py
```diff
@@ -0,0 +1,287 @@
...
+          itertools.combinations_with_replacement(rec.shapes, getattr(jnp, rec.name).nargs))
+        for dtypes in itertools.product(
+          *(_valid_dtypes_for_shape(s, rec.dtypes) for s in shapes)))
+      for rec in itertools.chain(UFUNC_RECORDS)))
+  @jax.numpy_rank_promotion('allow')  # This test explicitly exercises implicit rank promotion.
+  def testCall(self, rec, shapes, dtypes):
+    np_ufunc = getattr(np, rec.name)
+    jnp_ufunc = getattr(jnp, rec.name)
+
+    jnp_op = jnp_ufunc
```
mattjj, line 133:
```suggestion
    jnp_op = getattr(jnp, rec.name)
```

Any reason to bind the intermediate name `jnp_ufunc`? (I find superfluous intermediate names can make it easy to write bugs, so they stand out to me!)

mattjj (approved):
Nicely done! I think the main cost here is in code complexity and hence maintenance cost, to be weighed against the benefits of the features. It's always tricky to balance costs and benefits, but I trust your judgement on it, so if you think this is a net win to land, do it!

tests/lax_numpy_ufunc_test.py
jakevdp: Done, thanks!

jakevdp:
For some of these, we should probably compute accumulations via `associative_scan` rather than `scan` for better performance. I'll probably leave that for a followup.

jakevdp:
I was thinking about this more... maybe a more JAXy way to do this would be to define a `@ufunc` decorator that can turn *any* function into a ufunc instance with the appropriate attributes. That would reduce the complexity of this change, and we could introduce the feature in a more gradual manner. We could potentially wrap basic ufuncs by default, but we wouldn't have to do it right away.

jakevdp:
Closing in favor of https://github.com/google/jax/pull/17054 for now.

PR #9539, opened 2022-02-11
Respect __jax_array__ in jnp.ndarray operations

### LINKED ISSUE #9538: `DeviceArray.__mul__` etc. don't respect `__jax_array__`
```python
import jax.numpy as jnp

class M:
    def __jax_array__(self):
        return jnp.array([1, 1])

M() * jnp.array([1, 1])
```
produces `TypeError: unsupported operand type(s) for *: 'M' and 'DeviceArray'`

In contrast `jnp.multiply(M(), jnp.array([1, 1]))` handles this just fine. No error, works as expected.

Also, NumPy handles this fine:
```python
import numpy as np

class M:
    def __array__(self):
        return np.array([1, 1])

M() * np.array([1, 1])  # no error!
```

jakevdp:
Fixes #9538

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -6774,6 +6774,8 @@ _accepted_binop_types = (int, float, complex, np.generic, np.ndarray, ndarray)
 def _defer_to_unrecognized_arg(binary_op):
   # Ensure that other array types have the chance to override arithmetic.
   def deferring_binary_op(self, other):
+    if hasattr(other, '__jax_array__'):
+      other = other.__jax_array__()
     if not isinstance(other, _accepted_binop_types):
       return NotImplemented
     return binary_op(self, other)
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -3852,6 +3852,12 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     ans = jnp.array(a)
     self.assertEqual(ans, 3.)
 
+  def testJaxArrayOps(self):
+    class arraylike:
+      def __jax_array__(self):
+        return jnp.array(3.)
+    self.assertArraysEqual(arraylike() * jnp.arange(10), jnp.array(3.) * jnp.arange(10))
+
   def testMemoryView(self):
     self.assertAllClose(
         jnp.array(bytearray(b'\x2a')),


froystig (approved):

jax/_src/numpy/lax_numpy.py
```diff
@@ -6774,6 +6774,8 @@ def wrapped(*args, **kwargs):
 def _defer_to_unrecognized_arg(binary_op):
   # Ensure that other array types have the chance to override arithmetic.
   def deferring_binary_op(self, other):
+    if hasattr(other, '__jax_array__'):
+      other = other.__jax_array__()
```
shoyer, line 6778: This means means that JAX won't check for special methods (e.g., `__radd__` for overriding `+`) if `other` implements `__jax_array__`. In other words, `__jax_array__` is incompatible with having defining different semantics for arithmetic.

Are we sure this is what we what? Maybe this is consistent with how jax.numpy functions are already handled, but it might preclude some use-cases, e.g., perhaps someone already defines `jax_array + custom_array` to return a `CustomArray`, which would be broken by this change.

jakevdp: I don't have any strong opinion here – what do you think? Do you know of any code that this change would break, or is it just theoretical? @patrick-kidger since you raised the issue, do you have any opinion here?

patrick-kidger: Hmm. So FWIW my use case isn't affected by this, but I think I'd express a mild preference in favour of the current implementation.

Basically it comes down to how you feel about specificity in multiple dispatch. Switching to Julia notation: when calling `add(::DeviceArray, ::CustomArray)`, which of the following would you rate as "more specific"?
```julia
add(::DeviceArray, ::JAXArray)
```
or
```julia
add(::Any, ::CustomArray)
```
Where the type hierarchy goes `CustomArray <: JAXArray`, `DeviceArray <: JAXArray`.

(For clarity compared to Python: here `::JAXArray` is analogous to a `DeviceArray` or an object implementing `__jax_array__`.)

There's no set-in-stone answer here. For example Julia's multiple dispatch implementation is riddled with weird special cases and things. (e.g. [See here for quite a humourous/informative take on it.](https://www.youtube.com/watch?v=TPuJsgyu87U&t=1624s))

Personally I'm quite fond of the idea of what I refer to as "directional dispatch": iterate from left-to-right down the argument list, and choose the most specific method for each individual argument in turn. This is a simple rule that's easy to reason about. (Once you've internalised it.)

And in fact this is exactly what Python does -- and thus exactly what the current implementation does -- in the case of binary operations: first check `__add__`, then check `__radd__`. Hence my preference for the current implementation.

I would note that I'm sure not everyone would agree with me -- indeed this isn't what Julia does! -- and I can't claim to have studied specificity rules for multiple dispatch at any level of academic formalism.

/long answer to simple question.

shoyer: Python does allow you to do directional dispatch, but I've found users are almost always surprised when `x + y` differs from `y + x`. Usually that's considered a bug.

I think these semantics are probably fine, but when we get aroud to documenting `__jax_array__` we should be clear that it doesn't make sense for classes that define arithmetic without casting the outputs to a JAX array.

Issue #6089, opened 2021-03-16
scipy.linalg.solve_sylvester


shoyer:
This would be useful for differentiating through degenerate eigenvalue problems, among other use-cases: https://github.com/google/jax/issues/669#issuecomment-777052841

LoicRaillon:
I recently came across this [paper](https://arxiv.org/abs/2011.11430) if it helps.

kgaranger:
This feature would also be useful for implementing the matrix logarithm (https://github.com/google/jax/issues/5469#issuecomment-773567386) based on the algorithm introduced here: https://eprints.maths.manchester.ac.uk/2015.

Issue #2478, opened 2020-03-21
Implement matrix functions from scipy.linalg


oleg-kachan:
Requesting the implementation of the `scipy.linalg` matrix functions:

- [x] expm
- [ ] logm
- [ ] sqrtm
- [ ] fractional_matrix_power

khansel01:
Hi guys, 
I just wanted to ask if there is any progress on sqrtm implementation in Jax.
Thanks in advance!

PR #9544, opened 2022-02-12
Adds matrix sqrt

SaturdayGenfo:
This PR adds an implementation of the matrix square root (makes progress on #2478). The algorithm is a Schur decomposition based method and is entirely performed with complex types. (Recent improvements of this method use Sylvester equation solvers which aren't available in JAX for the moment as mentioned in #6089)

To make things jittable, the main routine, computing the square root of upper-triangular matrices, uses `lax.fori_loop` which makes things a bit obscure. Before the jit-friendly-rewrite, the function [here](https://github.com/SaturdayGenfo/jax/blob/173586e09f6e6a48c3ebedae1898fea683efe452/jax/_src/scipy/linalg.py#L616) looked like this:
```[python]
for j in range(n):
    for i in range(j-1, -1, -1):
        s = 0
        for k in range(i+1, j):
            s += U[i, k] * U[k,i]
        num = T[i, j] - s
        denom = diag[i] + diag[j]
        if num == 0.0:
            U = U.at[i,j].set(0.0)
        else:
            U = U.at[i,j].set(num/denom)  
```
(unsure if the proposed jit-friendly-rewrite is best)

I didn't squash the two commits for clarity. The first wraps the existing schur primitive into a `jax.scipy` function, the second adds `jax.scipy.sqrtm`.


shoyer:
How does performance compare to SciPy? I am a little concerned that iterating over individual matrix elements in JAX is going to be painfully slow. Usually we rely on compiled routines for this sort of thing, e.g., from Lapack or CuSolver.

SaturdayGenfo:
I ran a few tests (in double precision) and it performs roughly as well as the un-blocked version of `scipy.linalg.sqrtm`. But it is slower than the sylvester augmented version with default `blocksize=64`.

times

The execution times are the average of 5 runs of the following functions:
```python
def time_jax(n):
    x = (np.random.randn(n,n))
    tic = time.perf_counter()
    jsp.linalg.sqrtm(x).block_until_ready()
    return time.perf_counter() - tic

def time_scipy(n):
    x = (np.random.randn(n,n))
    tic = time.perf_counter()
    sp.linalg.sqrtm(x, blocksize=1)
    return time.perf_counter() - tic

def time_scipy_sylvester(n):
    x = (np.random.randn(n,n))
    tic = time.perf_counter()
    sp.linalg.sqrtm(x)
    return time.perf_counter() - tic
```

SaturdayGenfo:
I thought some more about this. It does feel un-jax like to have these nested impure loops and it does become slow for matrices of size exceeding `5000x5000`.  To avoid these loops, one option is to write a custom call to Eigen like they do in tensorflow. Only issue is that custom calls are not the easiest to write.  

This version is only on CPU and does not yet have a jvp defined (although the jvp is only a sylvester solver away). Because of these reasons, there might not be so much urgency to having it included in JAX.  So a possible choice is to table this PR and instead work towards writing the custom call or a sylvester solver to make this `sqrtm` worthwhile. What do you think ?

jax/_src/scipy/linalg.py
```diff
@@ -103,7 +103,23 @@ def eigh(a, b=None, lower=True, eigvals_only=False, overwrite_a=False,
   del overwrite_a, overwrite_b, turbo, check_finite
   return _eigh(a, b, lower, eigvals_only, eigvals, type)
 
+@partial(jit, static_argnames=('output',))
+def _schur(a, output):
+  if output != "real":
```
shoyer, line 108: This should raise an error if an invalid `output` is provided

jax/_src/scipy/linalg.py
```diff
@@ -103,7 +103,23 @@ def eigh(a, b=None, lower=True, eigvals_only=False, overwrite_a=False,
...
+    a *= complex(1.0, 0)
```
shoyer, line 109: I think this casting would be clearer (and maybe more efficient) with `astype`.

jax/_src/scipy/linalg.py
```diff
@@ -103,7 +103,23 @@ def eigh(a, b=None, lower=True, eigvals_only=False, overwrite_a=False,
...
+    T, S = lax_linalg.schur(a)
+    return T, S
+  else:
+    return lax_linalg.schur(a)
```
shoyer, line 113: both these code paths look the same to me? so they can probably be consolidated

jax/_src/scipy/linalg.py
```diff
@@ -595,3 +611,48 @@ def polar(a, side='right', method='qdwh', eps=None, maxiter=50):
...
+    j, U = data
+    i = j - 1 - l
+    s = lax.fori_loop(i + 1, j, lambda k, val: val + U[i, k] * U[k, j], 0.0)
+    value = jnp.where(T[i, j] == s, 0.0,
+                      (T[i, j] - s) / (diag[i] + diag[j]))
+    return j, U.at[i, j].set(value)
+
+  def j_loop(j, U):
+    _, U = lax.fori_loop(0, j, i_loop, (j, U))
+    return U
```
shoyer, line 631: do you have a reference for where this algorithm came from? this would be helpful to include in a comment

tests/linalg_test.py
```diff
@@ -1373,6 +1373,82 @@ def expm(x):
...
+  @jtu.skip_on_devices("gpu", "tpu")
+  def testSqrtmBatching(self, shape, dtype):
+    rng = jtu.rand_default(self.rng())
+    batch_size = 10
+    shape = (batch_size, ) + shape
+    args = rng(shape, dtype)
+    square = vmap(lambda mat: mat @ mat)
+
+    roots = vmap(jsp.linalg.sqrtm)(args)
+    self.assertAllClose(square(roots), args, atol=5e-3, check_dtypes=False)
```
shoyer, line 1451: I don't think we really need this test for the current implementation of `sqrtm`, because it's built out of other batchable JAX primitives.

(It's different for the tests that are checking linear algebra routines with custom batching rules.)

jax/_src/scipy/linalg.py
```diff
@@ -595,3 +611,48 @@ def polar(a, side='right', method='qdwh', eps=None, maxiter=50):
...
+  diag = jnp.sqrt(jnp.diag(T))
+  n = diag.size
+  U = jnp.diag(diag)
+
+  def i_loop(l, data):
+    j, U = data
+    i = j - 1 - l
+    s = lax.fori_loop(i + 1, j, lambda k, val: val + U[i, k] * U[k, j], 0.0)
+    value = jnp.where(T[i, j] == s, 0.0,
+                      (T[i, j] - s) / (diag[i] + diag[j]))
```
shoyer, line 626: Could we add a unit test that explicitly would check this code path? e.g., perhaps a matrix like `[[1, 0, 0], [0, 0,0], [0, 0, 0]]`

shoyer:
I think this implementation is actually probably OK for a first pass. Performance within a factor of 2-4x of SciPy is not terrible and I think matrices of size 5000x5000 are very rare. JVPs and even VJPs should work fine out of the box, because you wrote this in terms of existing JAX primitives, though it's possible there are more numerically stable implementations of gradients.

jax/_src/scipy/linalg.py
SaturdayGenfo: I added a test for diagonal matrices with zeros

jax/_src/scipy/linalg.py
SaturdayGenfo: I added a line to check the `output` argument `schur`.

jax/_src/scipy/linalg.py
```diff
@@ -595,3 +607,54 @@ def polar(a, side='right', method='qdwh', eps=None, maxiter=50):
...
+``jax.scipy.linalg.sqrtm`` is always ``complex64`` for 32-bit input,
+and ``complex128`` for 64-bit input.
+
+This function implements the complex Schur method described in [A]. It does not use recursive blocking
+to speed up computations as a Sylvester Equation solver is not available yet in JAX.
+
+[A] Björck, Å., & Hammarling, S. (1983).
+    "A Schur method for the square root of a matrix". Linear algebra and its applications, 52, 127-140.
+""")
+def sqrtm(A, disp=True, blocksize=1):
```
shoyer, line 656: Same as above: let's omit `disp=True`

jax/_src/scipy/linalg.py
```diff
@@ -103,7 +103,19 @@ def eigh(a, b=None, lower=True, eigvals_only=False, overwrite_a=False,
...
-
+@partial(jit, static_argnames=('output',))
+def _schur(a, output):
+  if output == "complex":
+    a = a.astype(jnp.complex64 if a.dtype <= jnp.complex64 else jnp.complex128)
+  return lax_linalg.schur(a)
+
+@_wraps(scipy.linalg.schur)
+def schur(a, output='real', lwork=None, overwrite_a=False, sort=None, check_finite=True):
+  del lwork, overwrite_a, sort, check_finite
```
shoyer, line 114: Can you omit unsupported arguments instead of deleting them? If someone explicitly writes `check_finite=True`, I would rather give them an `TypeError` about an invalid argument.

tests/linalg_test.py
```diff
@@ -1373,6 +1373,82 @@ def expm(x):
...
+  @jtu.skip_on_devices("gpu", "tpu")
+  def testSqrtmEdgeCase(self, diag, dtype):
+    """
+    Tests the zero numerator condition
+    """
+    mat = jnp.diag(jnp.array(diag)).astype(dtype)
+    root = jsp.linalg.sqrtm(mat)
+
+    self.assertAllClose(jnp.matmul(root, root, precision=lax.Precision.HIGHEST),
+                        mat, atol=1e-3, check_dtypes=False)
```
shoyer, line 1451: Why not directly test for the expected solution here (e.g., against the elementwise `jnp.sqrt(mat)`), rather than relying on squaring `root`? That seems a little more rigorous to me, unless `sqrtm` has some ambiguity about which root it returns.

jax/_src/scipy/linalg.py
```diff
@@ -595,3 +607,54 @@ def polar(a, side='right', method='qdwh', eps=None, maxiter=50):
...
+  T, Z = schur(A, output='complex')
+  sqrt_T = _sqrtm_triu(T)
+  return jnp.matmul(jnp.matmul(Z, sqrt_T, precision=lax.Precision.HIGHEST),
+                    jnp.conj(Z.T), precision=lax.Precision.HIGHEST)
+
+@_wraps(scipy.linalg.sqrtm,
+        lax_description="""
+This differs from ``scipy.linalg.sqrtm`` in that the return type of
+``jax.scipy.linalg.sqrtm`` is always ``complex64`` for 32-bit input,
+and ``complex128`` for 64-bit input.
```
shoyer, line 648: Why is this change from SciPy necessary for JAX? If possible, we try to maintain full compatibility.

jax/_src/scipy/linalg.py
SaturdayGenfo: Oh sorry misunderstood, I will fix it

jax/_src/scipy/linalg.py
SaturdayGenfo: It's because the SciPy version can operate entirely with real numbers because they have a few checks that avoid casting to complex types when unecessary. In particular, if the matrix is real, and has no negative or complex eigenvalues, all the types involved are made to stay real (see [here](https://github.com/scipy/scipy/blob/b5d8bab88af61d61de09641243848df63380a67f/scipy/linalg/_matfuncs_sqrtm.py#L54) and [here](https://github.com/scipy/scipy/blob/b5d8bab88af61d61de09641243848df63380a67f/scipy/linalg/_matfuncs_sqrtm.py#L167)). For JAX, it simply casts blindly to complex types, even if throughout the computations the imaginary part stays zero.  In cases where both the square root and the matrix are real, simply casting back to real types recovers the scipy result, but in my tests `jit` didn't appreciate having conditional casts.

shoyer: OK, that makes sense to me

tests/linalg_test.py
```diff
@@ -1373,6 +1373,84 @@ def expm(x):
...
       jtu.check_grads(expm, (a,), modes=["fwd", "rev"], order=1, atol=tol,
                       rtol=tol)
+  @parameterized.named_parameters(
+        jtu.cases_from_list({
+            "testcase_name":
+            "_shape={}".format(jtu.format_shape_dtype_string(shape, dtype)),
+            "shape": shape, "dtype": dtype
+        } for shape in [(4, 4), (15, 15), (50, 50), (100, 100)]
+                            for dtype in float_types + complex_types))
+  @jtu.skip_on_devices("gpu", "tpu")
```
shoyer, line 1383: Are these operations unsupported, slow or inaccurate for GPU/TPU? It would be good to understand why all these tests are skipped.

SaturdayGenfo: The bottleneck is the schur primitive which is only implemented for a CPU backend. What could be done is to test `_sqrtm_triu` on its own, on randomly generated upper triangular matrices (to skip the call to `shur`) so at least one test would be performed on GPU and TPU

shoyer: Ah, that makes complete sense. No need to change anything then. I don't think it's worth testing `_sqrt_triu` separately.

shoyer (approved):

Issue #3730, opened 2020-07-13
Document experimental.jet


thisiscam:
Hi all,

From a lot of discussion in issues, it seems like jet is able to "compute higher order derivatives efficiently". 
This functionality is particularly of my interest, but I can't seem to find any more documentation than those discussion and a few tests.
Could someone explain very shortly what is the arguments to `jet(f, primals, series)`, in particular, what `series` represents? 

Thanks!

jekbradbury:
I believe `series` is a list of arrays representing the first, second, third, etc.-order differentials to be propagated through the function (the generalization of the `tangents` argument in `jax.jvp`). So the equivalent of `jvp(f, p, t)` is `jet(f, p, [t])`.

thisiscam:
Thanks @jekbradbury. It worked for me, though I realized it was a bit lower level than I wanted (i.e. jet doesn't support pytrees yet), so I rolled my own hack for a function `value_jac_and_hessian`. 

My primary use case was for implementing a Newton-Raphson optimizer, so calculating just three orders works for me.

It will be great if at some point computing orders of derivatives all at once can get into main feature set of Jax.

mattjj:
cc @duvenaud @jacobjinkelly @jessebett 

I think @jessebett and @jacobjinkelly planned to write some kind of a post and/or docs for `jet`. In the meantime, there is the preprint [Learning Differential Equations that are Easy to Solve](https://arxiv.org/abs/2007.04504), though that's mostly about an application and doesn't offer a ton in the way of documentation. I also wrote up [some notes on the math](https://github.com/google/jax/files/4970131/jet.1.pdf) and Sec 3 might help explain the reason the `jet` API looks how it does, but actually the most important bit is Sec 4 and it's a bit hand-wavey.

I think `jet` only becomes more efficient at third and higher orders though, so it might not be relevant to the Newton-Raphson application.

mattjj:
@fedelebron noticed some typos in the "some notes on the math" note linked above, specifically a couple places where ∂g(x) was supposed to be ∂g(h(x)). Here's [an updated pdf with notes on the math](https://github.com/google/jax/files/6717197/jet.pdf). Thanks, @fedelebron!

pnkraemer:
Hi! I agree that it would be great to have some extra documentation for jet and would be willing to help. I started a corresponding discussion in #9536, because I wasn't sure how good a place this issue here is to discuss details :)

rajasekharporeddy:
Hi @thisiscam 

[...]

PR #9546, opened 2022-02-12
Include `experimental.jet` into docs

pnkraemer:
This PR introduces a module-level docstring for `jax.experimental.jet.py` and a function-level docstring for `jet.jet()`.

See the issue #3730 and the discussion #9536 for context.

The content of the docstrings is tightly coupled to how I use the API and how I understand the maths based on https://openreview.net/pdf?id=SkxEF3FNPH (as well as Griewank/Walther). Please let me know if I should rephrase some statements. I will leave comments in the diff, where I explain the thought process behind specific statements.

### DIFF
--- a/jax/experimental/jet.py
+++ b/jax/experimental/jet.py
@@ -18,19 +18,19 @@ r"""Jet is an experimental module for higher-order automatic differentiation
   How? Through the propagation of truncated Taylor polynomials.
   Let :math:`x: \mathbb{R} \rightarrow \mathbb{R}^n` be the Taylor polynomial
 
-  .. math:: x(t) := x_0 + x_1 t + \frac{x_2}{2!} t^2 + ... + \frac{1}{d!} x_d t^d.
+  .. math:: x(t) := x_0 + x_1 t + \frac{1}{2!} x_2 t^2 + ... + \frac{1}{d!} x_d t^d.
 
   For some function :math:`f: \mathbb{R}^n \rightarrow \mathbb{R}^m`,
   you can use :func:`jet` to compute the
   Taylor polynomial
 
-  .. math:: y(t) := y_0 + y_1 t + \frac{y_2}{2!} t^2 + ... + \frac{1}{d!} y_d t^d.
+  .. math:: y(t) := y_0 + y_1 t + \frac{1}{2!} y_2 t^2 + ... + \frac{1}{d!} y_d t^d.
 
   corresponding to :math:`f(x(\cdot)):\mathbb{R} \rightarrow \mathbb{R}^m`,
   and relates to :math:`x(t)` and :math:`f(z)` through
 
   .. math::
-    y_k = \frac{d^k}{dt^k}f(x(t_0)), \quad k=0,...,d.
+    y_k = \partial^k f(x(t_0)), \quad k=0,...,d.
 
   More specifically, :func:`jet` computes
 
@@ -40,7 +40,7 @@ r"""Jet is an experimental module for higher-order automatic differentiation
   and can thus be used for high-order
   automatic differentiation of :math:`f`.
   Details are explained in
-  `this paper <https://openreview.net/forum?id=SkxEF3FNPH>`__.
+  `these notes <https://github.com/google/jax/files/6717197/jet.pdf>`__.
 
   Note:
     Help improve :func:`jet` by contributing
--- a/jax/experimental/jet.py
+++ b/jax/experimental/jet.py
@@ -12,6 +12,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+r"""Jet is an experimental module for higher-order automatic differentiation
+  that does not rely on repeated first-order automatic differentiation.
+
+  How? Through the propagation of truncated Taylor polynomials.
+  Let :math:`x: \mathbb{R} \rightarrow \mathbb{R}^n` be the Taylor polynomial
+
+  .. math:: x(t) := x_0 + x_1 t + \frac{x_2}{2!} t^2 + ... + \frac{1}{d!} x_d t^d.
+
+  For some function :math:`f: \mathbb{R}^n \rightarrow \mathbb{R}^m`,
+  you can use :func:`jet` to compute the
+  Taylor polynomial
+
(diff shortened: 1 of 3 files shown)


jax/experimental/jet.py
```diff
@@ -12,6 +12,41 @@
...
+
+  More specifically, :func:`jet` computes
+
+  .. math::
+    y_0, (y_1, . . . , y_d) = \texttt{jet} (f, x_0, (x_1, . . . , x_d))
+
+  and can thus be used for high-order
+  automatic differentiation of :math:`f`.
+  Details are explained in
+  `this paper <https://openreview.net/forum?id=SkxEF3FNPH>`__.
```
pnkraemer, line 43: The content of this docstring closely follows the exposition in https://openreview.net/pdf?id=SkxEF3FNPH. It seems to be more maths than other, comparable module docstrings, but I figured it is useful to establish the variables for easier reference of what `jet` does.

jax/experimental/jet.py
```diff
@@ -12,6 +12,41 @@
...
+
+  Note:
+    Help improve :func:`jet` by contributing
+    `outstanding primitive rules <https://github.com/google/jax/issues/2431>`__.
```
pnkraemer, line 47: Other experimental modules had similar comments: `ann.py` lists todos, or `host_callback` states "please give feedback, and expect changes.". If I should remove this comment, though, please let me know.

jax/experimental/jet.py
```diff
@@ -32,6 +67,51 @@
...
+      array, scalar, or standard Python container of arrays or scalars.
+    primals: The primal values at which the Taylor approximation of ``fun`` should be
+      evaluated. Should be either a tuple or a list of arguments,
+      and its length should be equal to the number of positional parameters of
+      ``fun``.
+    series: Higher order Taylor-series-coefficients.
+      Together, `primals` and `series` make up a truncated Taylor polynomial.
+      Should be either a tuple or a list of tuples or lists,
+      and its length dictates the degree of the truncated Taylor polynomial.
+
```
pnkraemer, line 84: I did not document the exceptions that are raised, because I did not see this kind of doc in many other functions. But I am aware that it would go here, please let me know if I should add it.

jax/experimental/jet.py
```diff
@@ -32,6 +67,51 @@
...
+  :func:`jet` returns the Taylor coefficients of :math:`f(x(t)) = \sin(t^3)`:
+
+  >>> print(y0,  np.sin(x0))
+  0.12467473 0.12467473
+
+  >>> print(y1, np.cos(x0) * x1)
+  0.7441479 0.74414825
+
+  >>> print(y2, -np.sin(x0) * x1 +  x2)
+  2.9064622 2.906494
```
pnkraemer, line 113: In order to make, for instance, a `doctest` pass reliably, I would have to round these values. But I didnt see much rounding happening similar examples in other modules, and `pytest -n auto jax/experimental/jet.py` does not check these examples at least.

jax/experimental/jet.py
mattjj: Thanks for pushing on this! 

I don't recommend following the math or notation in that workshop paper. I recommend working from [these notes](https://github.com/google/jax/files/6717197/jet.pdf). In particular, the differentiation notation is different, with no \frac. We should link to that pdf as well (instead of openreview).

mattjj:
This is really awesome, thanks so much for contributing this!

jax/experimental/jet.py
mattjj: Not documenting exceptions sounds good to me.

jax/experimental/jet.py
mattjj: This seems like a good idea!

jax/experimental/jet.py
pnkraemer: Good to know, thanks! I updated the link as well as the notation to be more in line with the notes you linked.

jax/experimental/jet.py
```diff
@@ -12,6 +12,46 @@
...
+    (h(x), \partial h(x)[v], \partial^2 h(x)[v], ..., \partial^K h(x)[v,...,v]),
+
+  which represents a :math:`K`-th order Taylor approximation
+  of :math:`h` at :math:`x`, :func:`jet` returns a :math:`K`-th order
+  Taylor approximation of :math:`f` at :math:`x`,
+
+  .. math::
+    (f_0, ..., f_K) :=
+    (f(x), \partial f(x)[v], \partial^2 f(x)[v], ..., \partial^K f(x)[v,...,v]).
+
```
pnkraemer, line 39: the PDF you linked in the other comment discusses normalised Taylor coefficients (i.e. f_k / k!), but as far as I am aware, `jet` does not work with normalised coefficients. Right? 

Because if it does, I should update this docstring! :)

mattjj: > the PDF you linked in the other comment discusses normalised Taylor coefficients (i.e. f_k / k!), but as far as I am aware, jet does not work with normalised coefficients. Right?

Hah, I forget! Thanks for checking. It would be good to make sure the docstring agrees with the code. If you've looked at the code recently, I trust your evaluation as to which it uses!

YouJiacheng:
I would appreciate it if you could add some multivariate examples? For example, how to compute hessian of a MLP using `jet`? Or it is impossible to compute hessian accelerated by `jet`?🥺

pnkraemer:
> I would appreciate it if you could add some multivariate examples? For example, how to compute hessian of a MLP using `jet`? Or it is impossible to compute hessian accelerated by `jet`?pleading_face

No, I don't think it is impossible at all! :) 

But I am not sure the example in the documentation is the best place for it. I think it should also not be done as part of this pull request, but I will include it if @mattjj wants me to :) 

There seem to be a few more or less ongoing discussions about documentation of `jet`: e.g., #3730, #9536, and #9598 (which you raised -- see also my reply there for some more context). If those discussions keep attracting interest, perhaps someone volunteers to put together more elaborate examples of this type of functionality?

mattjj:
Hey, sorry for going silent on this. I got drawn into unrelated stuff.

mattjj (approved):
Looks great! Thanks for this contribution. I think it'll help rebuild momentum around `jet`!

PR #9549, opened 2022-02-13
Let random.choice mimic numpy behaviour and throw and error if more s…

### LINKED ISSUE #9548: jax.random.choice has undefined behavior when sampling without replacement and using probabilities
Summary: if enough weights/probabilities are zero such that sampling w/o replacement leads to too few possibilities to sample, jax should raise an error.
Simple case: 
```python
import jax
import jax.numpy as jnp

prob = jnp.zeros(10)
prob = prob.at[9].set(1)
jax.random.choice(jax.random.PRNGKey(5), jnp.arange(10), shape=(5,), replace=False, p=prob)
# DeviceArray([9, 0, 1, 2, 3], dtype=int32)
```
Numpy raises a 
```
ValueError: Fewer non-zero entries in p than size
```
with equivalent code.

Fix: 
https://github.com/google/jax/blob/main/jax/_src/random.py#L493
Here, do a check that the number of nonzero entries is >= desired number of entries.

MichaelMarien:
…ample are requested than provided non-zero probabilities and replace is False. Closes #9548 

* add check in random.choice on count_nonzero elements of p
* add tests to check new behavior

I believe that ideally, we would use jnp.count_nonzero instead of the numpy function. However, although using this seems to work for the functions (p and shape are already traced), it returns some errors on the current way the tests of these functions are executed. I'm not sure if I'm allowed to touch this and what would be the preferred way. Don't mind spending time on this if some direction is provided.

```
jax._src.traceback_util.UnfilteredStackTrace: jax._src.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: Traced<ShapedArray(bool[])>with<DynamicJaxprTrace(level=0/1)>
While tracing the function <lambda> at /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 for jit, this value became a tracer due to JAX operations on these lines:

  operation a:i32[] = xla_call[
  call_jaxpr={ lambda ; b:f32[1]. let
      c:bool[1] = ne b 0.0
      d:i32[1] = convert_element_type[new_dtype=int32 weak_type=False] c
      e:i32[] = reduce_sum[axes=(0,)] d
    in (e,) }
  name=count_nonzero
] f
    from line /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 (<lambda>)

  operation a:i32[] = convert_element_type[new_dtype=int32 weak_type=False] b
    from line /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 (<lambda>)

See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.ConcretizationTypeError

--------------------

  File "/Users/michaelmarien/anaconda3/envs/jax/lib/python3.9/site-packages/absl/testing/parameterized.py", line 314, in bound_param_test
  File "/Users/michaelmarien/PycharmProjects/jax/tests/random_test.py", line 521, in testChoice
    self.assertArraysEqual(sample, jax.jit(rand, static_argnames=
  File "/Users/michaelmarien/PycharmProjects/jax/tests/random_test.py", line 508, in <lambda>
    rand = lambda key, x: random.choice(key, x, shape, replace, p, axis)
  File "/Users/michaelmarien/PycharmProjects/jax/jax/_src/random.py", line 491, in choice
    if jnp.count_nonzero(p) < n_draws:
jax._src.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: Traced<ShapedArray(bool[])>with<DynamicJaxprTrace(level=0/1)>
While tracing the function <lambda> at /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 for jit, this value became a tracer due to JAX operations on these lines:

  operation a:i32[] = xla_call[
  call_jaxpr={ lambda ; b:f32[1]. let
      c:bool[1] = ne b 0.0
      d:i32[1] = convert_element_type[new_dtype=int32 weak_type=False] c
      e:i32[] = reduce_sum[axes=(0,)] d
    in (e,) }
  name=count_nonzero
] f
    from line /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 (<lambda>)

  operation a:i32[] = convert_element_type[new_dtype=int32 weak_type=False] b
    from line /Users/michaelmarien/PycharmProjects/jax/tests/random_test.py:508 (<lambda>)

See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.ConcretizationTypeError
```

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -488,6 +488,8 @@ def choice(key: KeyArray,
       r = p_cuml[-1] * (1 - uniform(key, shape))
       ind = jnp.searchsorted(p_cuml, r)
     else:
+      if np.count_nonzero(p) < n_draws:
+        raise ValueError("Fewer non-zero entries in p than size")
       # Gumbel top-k trick: https://timvieira.github.io/blog/post/2019/09/16/algorithms-for-sampling-without-replacement/
       g = -gumbel(key, (n_inputs,)) - jnp.log(p)
       ind = jnp.argsort(g)[:n_draws]
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -1128,6 +1128,14 @@ class LaxRandomTest(jtu.JaxTestCase):
     with self.assertRaises(TypeError):
       random.choice(key, 5, 2, replace=True)
 
+  def testMoreSamplesThanNonZeroProbabilitiesReplaceFalseError(self):
+    key = self.seed_prng(0)
+    prob = jnp.zeros(5)
+    prob = prob.at[4].set(1)
+    with self.assertRaises(ValueError):
+      random.choice(key, 5, shape=(2,), replace=False, p=prob)
+    self.assertEqual(random.choice(key, 5, shape=(1,), replace=False, p=prob), 4)
+
   def test_eval_shape_big_random_array(self):
     def f(x):
       return random.normal(self.seed_prng(x), (int(1e12),))


jax/_src/random.py
```diff
@@ -488,6 +488,8 @@ def choice(key: KeyArray,
       r = p_cuml[-1] * (1 - uniform(key, shape))
       ind = jnp.searchsorted(p_cuml, r)
     else:
+      if np.count_nonzero(p) < n_draws:
```
hawkinsp, line 491: I'm afraid this won't work in JAX in general: if this function is under a `jit` decorator, as it frequently will be, you cannot test the concrete values since at trace time you will only have abstract values.

We're working on adding some support for traced errors, though.

@LenaMartens

MichaelMarien: Thanks for the response @hawkinsp, still trying to grasp some of the basics of Jax and this definitely helps. Should I close this PR?

PR #9561, opened 2022-02-14
Add translation rule for optimization barrier.

pschuh:
Also adds a translation rule for remat that uses the new optimization barrier
op. If you find errors, consider disabling the remat lowering using
`jax_remat_opt_barrier` config flag.


jax/_src/lax/control_flow.py
```diff
@@ -3122,4 +3130,39 @@ def _remat_translation_rule(*args,
...
+  return args
+
+
+def _optimization_barrier_translation_rule(ctx, avals_in, avals_out, *args):
+  out = xops.OptimizationBarrier(xops.Tuple(ctx.builder, args))
+  return [xops.GetTupleElement(out, i) for i in range(len(args))]
+
+
+def _optimization_barrier_lowering_rule(ctx, *args):
+  loop_carry_types = _map(mlir.aval_to_ir_types, ctx.avals_in)
```
hawkinsp, line 3147: I suspect `loop_carry` is a poor choice of name here, given there are no longer any loops.

jax/_src/lax/control_flow.py
```diff
@@ -3122,4 +3130,39 @@ def _remat_translation_rule(*args,
...
+  flat_loop_carry_types = util.flatten(loop_carry_types)
+
+  flat_args = mlir.flatten_lowering_ir_args(args)
+  barrier_op = mhlo.OptimizationBarrierOp(flat_loop_carry_types, flat_args)
```
hawkinsp, line 3151: BTW, you can avoid the need to provide the output type if you implement a type inference rule on the MHLO op. You'd need to do that separately in another change, of course.

hawkinsp (approved):

jax/_src/lax/control_flow.py
pschuh: Done.

jax/_src/lax/control_flow.py
pschuh: Ah, good to know. I was curious why this didn't just work.

hawkinsp (approved):

PR #9571, opened 2022-02-15
Check input of broadcast_shapes:

### LINKED ISSUE #9527: `lax.broadcast_shapes` should error on negative shapes
```python
>>> lax.broadcast_shapes((-1,), (1,))                                                                                                                    
(-1,)
>>> np.broadcast_shapes((-1,), (1,))                                                                                                                     
<ipython-input-2-6e3eddf076bd> in <module>
----> 1 np.broadcast_shapes((-1,), (1,))

~/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/numpy/lib/stride_tricks.py in broadcast_shapes(*args)
    469     (5, 6, 7)
    470     """
--> 471     arrays = [np.empty(x, dtype=[]) for x in args]
    472     return _broadcast_shape(*arrays)
    473 

~/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/numpy/lib/stride_tricks.py in <listcomp>(.0)
    469     (5, 6, 7)
    470     """
--> 471     arrays = [np.empty(x, dtype=[]) for x in args]
    472     return _broadcast_shape(*arrays)
    473 

ValueError: negative dimensions are not allowed
```

MichaelMarien:
Check input of broadcast_shapes:
* canonicalize provided shapes to catch type errors et al
* do extra check on positivity of canonicalized dimensions and raise value error
* add test for type and value errors

This PR closes #9527

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -113,6 +113,9 @@ def _broadcast_shapes_cached(*shapes: Tuple[int, ...]) -> Tuple[int, ...]:
   return _broadcast_shapes_uncached(*shapes)
 
 def _broadcast_shapes_uncached(*shapes):
+  shapes = [canonicalize_shape(shape) for shape in shapes]
+  if not all([all(el >= 0 for el in shape) for shape in shapes]):
+    raise ValueError('negative dimensions are not allowed')
   fst, *rst = shapes
   if not rst: return fst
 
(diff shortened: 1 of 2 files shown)


MichaelMarien:
Similar as another PR I tried, I realised this won't work as the boolean conditioning doesn't just work on the shape but the actual elements. Guess this would be another use case for traced errors.

PR #9576, opened 2022-02-15
Validate `lax.broadcast_shape` inputs before control flow execution

nicholasjng:
Addresses #9527.

This commit addresses previously unvalidated inputs to `jax.lax.broadcast_shapes` by adding a small validation check before control flow execution. A legal input to `lax.broadcast_shapes` hereafter is defined as an input that
1) is a sequence (i.e., implements `for..in` iteration) of integers and
2) said integers are all non-negative.

In addition, two tests were added to `tests.lax_vmap_test` to check that proper errors are raised when attempting to use illegal inputs with `lax.broadcast_shapes`.

Two more things:
- Is the type check fine? I arrived at this after (successfully, at least locally) debugging some test failures, but this could be wrong for reasons that I did not see.
- The type hint there is probably wrong now, since I moved the validation check up from `_try_broadcast_shapes` to the main control flow? I guess inputs at that stage could also be `core.Tracer` objects. Please let me know what you think.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -83,9 +83,23 @@ Shape = core.Shape
 
 T = TypeVar("T")
 
+def _validate_shapes(shapes: Sequence[Shape]):
+  def _check_static_shape(shape: Shape):
+    checked = canonicalize_shape(shape)
+    if not all(idx >= 0 for idx in checked):
+      msg = f"Only non-negative indices are allowed when broadcasting" \
+            f" static shapes, but got shape {shape!r}."
+      raise TypeError(msg)
+
+  assert shapes
+  if config.jax_dynamic_shapes:
+    # pass dynamic shapes through unchecked
+    return
+  else:
+    _ = tuple(map(_check_static_shape, shapes))
+
 def _try_broadcast_shapes(
     shapes: Sequence[Tuple[int, ...]]) -> Optional[Tuple[int, ...]]:
-  assert shapes
   if len(shapes) == 1: return shapes[0]
   rank, *others = {len(shape) for shape in shapes}
   if others: return None  # must have consistent rank
@@ -113,6 +127,7 @@ def _broadcast_shapes_cached(*shapes: Tuple[int, ...]) -> Tuple[int, ...]:
   return _broadcast_shapes_uncached(*shapes)
 
 def _broadcast_shapes_uncached(*shapes):
+  _validate_shapes(shapes)
   fst, *rst = shapes
   if not rst: return fst
 
--- a/tests/lax_vmap_test.py
+++ b/tests/lax_vmap_test.py
@@ -723,6 +723,15 @@ class LaxVmapTest(jtu.JaxTestCase):
     out_shape = lax.broadcast_shapes(shape1, shape2)
     self.assertTrue(all(type(s) is int for s in out_shape))
 
+  def testBroadcastShapesFaultyInputs(self):
+    err_shape1, err_shape2 = (-1,), "hello"
+    # negative inputs should fail while informing about illegal negative indices...
+    with self.assertRaisesRegex(TypeError, "Only non-negative indices are allowed.*"):
+      lax.broadcast_shapes(err_shape1)
+    # ... while non-integers should error earlier, in the canonicalize_shape machinery.
+    with self.assertRaisesRegex(TypeError, "Shapes must be 1D sequences.*"):
+      lax.broadcast_shapes(err_shape2)
+
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_shape={}_k={}_bdims={}".format(
           jtu.format_shape_dtype_string(shape, dtype), k, bdims),


jax/_src/lax/lax.py
```diff
@@ -83,9 +83,15 @@
 
 T = TypeVar("T")
 
+def _validate_shapes(shapes: Sequence[Tuple[int, ...]]):
+  assert shapes
+  for shape in shapes:
+    if not all(np.issubdtype(type(ix), np.integer) and ix >= 0 for ix in shape):
```
jakevdp, line 89: This subdtype check is not correct – for example, dimensions can be instances of `jax.numpy.int32`, or they can be dynamic size objects. Better would be to call `_canonicalize_dimension` on each entry, because it takes care of these details (and raises an appropriate error if the dimension is invalid)

jakevdp (changes requested):
Thanks for addressing this! I think we need to use a different approach; see below

jax/_src/lax/lax.py
nicholasjng: Ah, I got you. Supposedly, this also handles the case that shape is a traced value? I'll try and fix it later in the day.

jakevdp:
So my confusion here is that I would have thought `broadcast_shapes` would require static inputs, in which case the check for `-1` would be fine. Do you know where the call to `broadcast_shapes` with traced values is happening?

jakevdp:
...oh I see it's from the dynamic shapes test. In that case we can check for negative shapes in the case that the inputs are static, but pass dynamic shapes through. What do you think?

nicholasjng:
> ...oh I see it's from the dynamic shapes test. In that case we can check for negative shapes in the case that the inputs are static, but pass dynamic shapes through. What do you think?

Sounds good to me. Do you know any check for dynamic shapes off the top of your head? 
(Maybe `if isinstance(shape, core.Tracer)`?) 
Thank you in advance, I'll try to resubmit tomorrow.

jakevdp:
I looked into it a bit, I think probably what we want is the `_canonicalize_dim` utility, and then we only check the value if the result is an int:
```python
shape = core.canonicalize_shape(shape)
```
And then we could check that all integer entires in the resulting shape are non-negative. Or perhaps that negativity check should be part of `canonicalize_shape` itself?

jakevdp:
Digging in it looks like `make_jaxpr` with `abstracted_axes` will in general produce traced shapes... this is not a public feature yet, so I'm checking with other team members regarding how it should be handled

nicholasjng:
I would agree on this - can you really expect to meaningfully canonicalize dynamic shapes? Also, since broadcasting requires some dimensions to match (which you cannot check in traced shapes), this might be impossible in general.

In the test itself, `api_test/test_basic_batch_poly`, it looks like the matrix-vector op `Wx + b` is the culprit. I guess `x` is being called with `input` here - which contains an "abstracted shape" from the `make_jaxpr` call (whatever that is).

The `abstracted_shapes` argument to `make_jaxpr` also seems to be undocumented:
https://github.com/google/jax/blob/d4c6f953312b47437b583de72e3f539081c9e833/jax/_src/api.py#L2622-L2646

Looks like we got some loose ends here, what are your thoughts?

nicholasjng:
> Digging in it looks like `make_jaxpr` with `abstracted_axes` will in general produce traced shapes... this is not a public feature yet, so I'm checking with other team members regarding how it should be handled. I think @mattjj would know

Okay, that's good to know. Let me know what comes out of it, happy to follow up with a fix once there's some direction.

jakevdp:
OK, so it sounds like in general handle traced shapes, but only if `config.jax_dynamic_shapes` is True. So as a first pass, I'd probably do the shape validation you have in mind, but do no validation if that flag is set. What do you think?

nicholasjng:
I incorporated your changes, and personally I'm much happier with the tracebacks now. Upon negative entries, the new shape validation function throws, while with non-integer inputs, `canonicalize_shape` throws an informative error. 

As you suggested, the whole validation machinery is bypassed when `config.jax_dynamic_shapes = True`. Let me know what you think.

jakevdp (approved):
Looks good to me! I'm going to request a review from @mattjj as well as he has more context on dynamic shapes

jax/_src/lax/lax.py
```diff
@@ -113,6 +127,7 @@ def _broadcast_shapes_cached(*shapes: Tuple[int, ...]) -> Tuple[int, ...]:
   return _broadcast_shapes_uncached(*shapes)
 
 def _broadcast_shapes_uncached(*shapes):
+  _validate_shapes(shapes)
```
mattjj, line 130: I'm not sure but I think an alternative here would be to only call `_validate_shapes(shapes)` in `_broadcast_shapes_cached` above (rather than here in `_broadcast_shapes_uncached`), and then we could even delete the `if config.jax_dynamic_shapes` logic in `_validate_shapes`. The idea is that the `try`/`except` in `broadcast_shapes` will fail (in `_validate_shapes`) only if dynamic shapes are involved.

But also this approach LGTM. Just thought I'd mention that possible alternative.

mattjj (approved):

jax/_src/lax/lax.py
nicholasjng: I just tried this locally, and unfortunately it restores the old behavior. But the dynamic shape test passes, just like you hinted. I think we still need to employ the validation machinery in the uncached code path to account for cache population / misses.

nicholasjng:
Thank you for the quick reviews!

PR #9584, opened 2022-02-15
Update JAX to use new math libraries in ROCm-5.0.

reza-amd:
This PR integrates new math libraries added in ROCm-5.0 in JAX.
A few unit tests have been marked to be skipped but they should be fixed in next ROCm release.


hawkinsp (commented):
Thanks for the PR!

I'm happy to see the CI/docker stuff: that promises to be useful if we ever need to debug a ROCm problem.

build/rocm/run_single_gpu.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
```
hawkinsp, line 1: All .py files have to have the apache license header. (Possibly other files also, but certainly the .py files.)

Same everywhere.

build/rocm/ci_build.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
reza-amd: Done.

build/rocm/run_single_gpu.py
reza-amd: Done.

PR #9594, opened 2022-02-16
fix: use boolean tf functions where possible instead of bool->int8->bool for jax2tf

oliverdutton:
Hello,

**Motivation**
I noticed jax2tf casts to int8 and back again to support the fact tf treats bool separately from int and float in its functions.
I propose using, where simple, the 'true' boolean tensorflow ops as these are more likely to be supported than the int8 by backends.

**Implementation**
I added an optional argument to `bool_to_int8` called `boolean_f` to match with `f` signature where the correct tf equivalent for booleans can be supplied. If available this is used in place of the bool->int8->bool pathway.

To avoid confusion with naming I updated the fn name `bool_to_int8` to `handle_boolean_args`.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1686,7 +1686,7 @@ tf_impl[lax.reduce_max_p] = handle_boolean_args(
   boolean_f=axes_to_axis(tf.reduce_any)) # Max is T if any one is T
 tf_impl[lax.reduce_min_p] = handle_boolean_args(
   axes_to_axis(tf.reduce_min), argnums=[0],
-  boolean_f=axes_to_axis(tf.reduce_all)) # Min is only F is all are T
+  boolean_f=axes_to_axis(tf.reduce_all)) # Min is only F if all are T
 tf_impl[lax.reduce_or_p] = axes_to_axis(tf.reduce_any)
 tf_impl[lax.reduce_and_p] = axes_to_axis(tf.reduce_all)
 
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -1376,7 +1376,7 @@ def handle_boolean_args(f, argnums: Sequence[int], boolean_f=None):
       # All argnums should be boolean
       assert len(argnum_types) == 1, argnum_types
       if boolean_f != None:
-        return boolean_f(*args **kwargs)
+        return boolean_f(*args, **kwargs)
       else:
         args_cast = [(tf.cast(a, tf.int8) if i in argnums else a)
                     for i, a in enumerate(args)]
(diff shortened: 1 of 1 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1373,38 +1375,46 @@ def wrapper(*args: TfVal, **kwargs):
...
 tf_impl[lax.eq_p] = tf.math.equal
 tf_impl[lax.ne_p] = tf.math.not_equal
 
-tf_impl[lax.ge_p] = bool_to_int8(tf.math.greater_equal, argnums=(0, 1))
-tf_impl[lax.gt_p] = bool_to_int8(tf.math.greater, argnums=(0, 1))
-tf_impl[lax.le_p] = bool_to_int8(tf.math.less_equal, argnums=(0, 1))
-tf_impl[lax.lt_p] = bool_to_int8(tf.math.less, argnums=(0, 1))
+bool_greater = lambda x,y: tf.logical_and(x, tf.logical_not(y)) # Only one combo: T,F -> T
+bool_less = lambda x,y: tf.logical_and(tf.logical_not(x), y) # Only one combo: F,T -> T
+def accept_equal(fn):
```
gnecula, line 1411: I would make the name of this function more specific, e.g., `boolean_compare_or_equal`.

gnecula (changes requested):
Thank you for contributing!

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -1671,10 +1681,20 @@ def _transpose(operand, *, permutation):
...
 tf_impl[lax.reduce_prod_p] = axes_to_axis(tf.reduce_prod)
-tf_impl[lax.reduce_max_p] = (
-    bool_to_int8(axes_to_axis(tf.reduce_max), argnums=[0]))
-tf_impl[lax.reduce_min_p] = (
-    bool_to_int8(axes_to_axis(tf.reduce_min), argnums=[0]))
+tf_impl[lax.reduce_max_p] = handle_boolean_args(
+  axes_to_axis(tf.reduce_max), argnums=[0],
+  boolean_f=axes_to_axis(tf.reduce_any))
+
+def bool_reduce_min(x, axes):
```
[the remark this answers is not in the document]
oliverdutton: If the hardware backend doesn't support int8 this becomes handy, boolean is virtually always supported. It took me a while to work out why int8 was causing me error messages.

I know its trivial that all hardware should support all types, but it still appears

jax/experimental/jax2tf/jax2tf.py
oliverdutton: It can also be greater_than_or_equal is not(less_than). This is shorter code, but possibly more confusing?

oliverdutton: I have split into two functions and used the >= is !(<) trick to simplify it

gnecula (approved):

PR #9596, opened 2022-02-16
Annotate nn.initializers

NeilGirdhar:
This was done to address concerns in this review: https://github.com/google/flax/pull/1803#discussion_r807895042, namely to expose an `Initializers` type annotation that can be used in other libraries.

### DIFF
--- a/jax/_src/nn/initializers.py
+++ b/jax/_src/nn/initializers.py
@@ -18,9 +18,10 @@ used in Keras and Sonnet.
 """
 
 
-from typing import Any, Callable, Sequence, Union
+from typing import Any, Sequence, Tuple, Union
 
 import numpy as np
+from typing_extensions import Literal, Protocol
 
 import jax.numpy as jnp
 from jax import lax
@@ -29,9 +30,25 @@ from jax import core
 from jax._src.util import prod
 from jax._src import dtypes
 
-DType = Any
+KeyArray = random.KeyArray
+Array = Any
+# TODO: Import or define these to match
+# https://github.com/numpy/numpy/blob/main/numpy/typing/_dtype_like.py.
+DTypeLikeFloat = Any
+DTypeLikeComplex = Any
+DTypeLikeInexact = Any  # DTypeLikeFloat | DTypeLikeComplex
+RealNumeric = Any  # Scalar jnp array or float
 
-def zeros(key, shape, dtype: DType = jnp.float_):
+class Initializer(Protocol):
+  @staticmethod
+  def __call__(key: KeyArray,
+               shape: core.Shape,
+               dtype: DTypeLikeInexact = jnp.float_) -> Array:
+    ...
+
+def zeros(key: KeyArray,
+          shape: core.Shape,
+          dtype: DTypeLikeInexact = jnp.float_) -> Array:
   """An initializer that returns a constant array full of zeros.
 
   The ``key`` argument is ignored.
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
@jakevdp Would you mind taking a look at this old pull request when you get a chance?  Thank you :)

jax/_src/nn/initializers.py
```diff
@@ -29,9 +30,26 @@
 from jax._src.util import prod
 from jax._src import dtypes
 
-DType = Any
-
-def zeros(key, shape, dtype: DType = jnp.float_):
+KeyArray = random.KeyArray
+Array = Any
+Shape = Tuple[int, ...]
```
jakevdp, line 35: I think restricting shapes to ints is too restrictive, because shapes can be dynamic. e.g. `zeros` below calls `jnp.zeros`, which in turn calls `core.canonicalize_shape`, which uses this annotation: https://github.com/google/jax/blob/bad9cd57f0ecbf125732252e172d80bdef30b5bb/jax/core.py#L1571-L1575

jakevdp (changes requested):
Seems reasonable, thanks. I have one concern regarding the definition of `Shape`; we'll also have to import to make sure the internal type checker (pytype) passes with these new annotations.

jax/_src/nn/initializers.py
NeilGirdhar: Thanks for catching this. I replaced all occurrences of `Shape` with `core.Shape`.

jakevdp (approved):
Looks good!

NeilGirdhar:
Thank you!

jakevdp:
It looks like this introduced some dtype errors in the test suite – please take a look!

jax/_src/nn/initializers.py
```diff
@@ -155,24 +189,33 @@ def _complex_uniform(key, shape, dtype):
...
-  whose modulus is truncated to `upper`, and the variance before the truncation is one.
+  whose modulus is truncated to `upper`, and the variance before the truncation
+  is one.
   """
   key_r, key_theta = random.split(key)
   real_dtype = np.array(0, dtype).real.dtype
   dtype = dtypes._to_complex_dtype(real_dtype)
-  t = (1 - jnp.exp(jnp.array(-(upper ** 2), dtype))) * random.uniform(key_r, shape, real_dtype).astype(dtype)
+  t = ((1 - jnp.exp(jnp.array(-(upper ** 2), dtype)))
+       * random.uniform(key_r, shape, real_dtype)).astype(dtype)
```
jakevdp, line 204: This seems to be the change that caused the issue – the new version is no longer compatible with strict dtype promotion.

NeilGirdhar: Very kind of you to track this down for me!  I wasn't careful enough when I did the rebase.  It is hopefully fixed now.

jakevdp:
Last thing - could you please squash your changes into a single commit? See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests for more. Thanks!

NeilGirdhar:
@jakevdp Yes, I'll squash them now!

I was told by some people not to squash during a code review to make it easy for you to verify that I'm making the requested changes.  When I'm submitting to Jax, would you prefer that I keep doing this going forward (not squashing until we're ready to push), or squash every time I make a change?

jakevdp:
FYI, this needed to be rolled back (#11735) because some of the annotations were incorrect (in particular, downstream callers were passing tuples of ints to some shapes annotated as `NamedShape`)

NeilGirdhar:
@jakevdp Sorry about that! The only things annotated `NamedShape` are private functions, so I guess I wrongly assumed that those wouldn't be called externally?

jakevdp:
In theory, that's true. In practice, folks use whatever they want 😁

NeilGirdhar:
@jakevdp Fair enough :smile: I'll try to be more careful of this in the future.  Do you think #11748 fixes that problem?

PR #9607, opened 2022-02-17
Index update operators: add scatter_apply()

jakevdp:
In the course of exploring a JAX API for numpy's ufunc methods (#9529), it became clear that there is one class of operation for which JAX doesn't currently have any high-level API: namely the equivalent of `np.ufunc.at(idx)` for a unary ufunc. For example:
```python
>>> import numpy as np

>>> x = np.array([2, 2, 2, 2])

>>> np.square.at(x, [1, 2, 2, 3, 3, 3])

>>> x
array([  2,   4,  16, 256])
```
The semantics are that the unary operation is repeatedly applied at the specified indices, something that is not currently possible with JAX (e.g. `x.at[idx].set(func(x[idx]))` only applies the function once in the case of repeated indices).

It turns out we can efficiently provide this functionality in JAX with a relatively thin wrapper around `scatter()`, accessed via `np.ndarray.at`, as is done in this PR:
```python
>>> import jax.numpy as jnp

>>> x = jnp.array([2, 2, 2, 2])

>>> idx = jnp.array([1, 2, 2, 3, 3, 3])

>>> x.at[idx].apply(jnp.square)
DeviceArray([  2,   4,  16, 256], dtype=int32)
```
This seems like it may be a useful API to make available, though admittedly I don't have any specific application in mind.

Why not allow an extra argument and extend this to binary ufuncs? Allowing this may cause confusion, because the semantics of `scatter()` are that updates can be applied in any order, and thus order of operations should not matter. This is not true for general binary functions (which is why we only implement specific binary ops in `scatter_add`, `scatter_mul`, etc.), but it is true of unary functions, so we can provide this more general implementation for the unary case.

### DIFF
--- a/jax/_src/lax/slicing.py
+++ b/jax/_src/lax/slicing.py
@@ -14,7 +14,8 @@
 
 import enum
 from functools import partial
-from typing import Any, NamedTuple, Optional, Sequence, Union
+from typing import Any, Callable, NamedTuple, Optional, Sequence, Union
+import weakref
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


jax/_src/lax/slicing.py
```diff
@@ -490,6 +491,63 @@ def scatter_max(
       indices_are_sorted=indices_are_sorted, unique_indices=unique_indices,
       mode=GatherScatterMode.from_any(mode))
 
+# To avoid recompilation, we store a dict of weak references to funcs.
+_scatter_apply_cache: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
+
+def scatter_apply(
```
hawkinsp, line 497: One catch with this is that scatter_p is not differentiable...

Document that, here and below?

jakevdp: Ah, thanks, will do. Maybe eventually a dedicated `scatter_apply` primitive would be useful then.

hawkinsp (approved):

soraros:
So now a boolean mask is casted to `int` right?
```
>>> x.at[jnp.array([True, False, False, True])].apply(jnp.square)
DeviceArray([4, 2, 2, 4], dtype=int32)
```
I guess when `idx.dtype == bool`, `at.apply` can be lowered into `jnp.where`?

Issue #9263, opened 2022-01-20
RNGs: key types and custom implementations


froystig:
This issue tracks the introduction of typed random key arrays with pluggable PRNG implementations.

*Update: see [JEP 9263](https://jax.readthedocs.io/en/latest/jep/9263-typed-keys.html) for an overview. Below is an earlier description.*

Work on this began with #6899, which mostly works out the "pluggable implementations" part, makes initial progress on the "typed" part, and sets things up throughout the codebase for backwards-compatibility and upgrading. The main configuration flag guarding the system upgrade is `config.jax_enable_custom_prng`.

The motivation for pluggability is straightforward: we'd like to set up various implementations for random bit generation, and still have all of `jax.random` work on top of any one of them. While we're at it, we could make this extensible.

The motivation for having keys reflected in types (somehow, at some level) is broadly to improve safety, and with it, to offer both users and the system some structural guarantees. Desiderata include:
* Moving to a user-facing representation of key arrays that reflects that they are indeed key arrays. Currently key arrays are plain `uint32` arrays to users, indistinguishable from any other data.
* Restricting operations on key arrays. The current plain array representation allows for key-invalidating operations (e.g. manual updating, addition, ...). We'd like to disallow these, or at least render mistakes unlikely.
* More/better opportunities to check key misuse, reuse, and so on.

Unsurprisingly, key types also help for mapping keys to RNG implementations (for the pluggability part above).

There are roughly two ways we've thought to approach endowing jax with key types, specifically, key-element-type arrays:

One is as a frontend component handled during staging, projected away to plain `u32` in staged IR, in analogy with how pytrees behave. To introduce a pytree type in Python that wraps an underlying key-data (say, `uint32`) array is not quite correct, since it would misbehave under `vmap`, `scan`, and even `jax.tree_map`. Instead, we might want to rely on something like the typeclass mechanisms still in development (e.g. `vmappable`, #8451) for this approach.

The other tack is to introduce key types into our IR and other internal machinery (with a corresponding lowering), and to map to that from a Python array-of-keys-like type during staging. This might confer some extra advantages downstream as well, such as more opportunities for checking throughout our front- and middle-end. This is roughly the approach we're taking (as of around summer '22).

[...]

PR #9609, opened 2022-02-17
implement `jnp.expand_dims` and `jnp.stack` for PRNGKeyArrays

froystig:
Also:
* fix `jnp.concatenate` and `jnp.append` for PRNGKeyArrays
* add `ndim` property to PRNGKeyArrays
* minor fix to `lax.expand_dims` with duplicate dimensions

Adding `expand_dims` is a step towards implementing other `jnp` operations as well (at least `vstack`). See the PR description for #8381 for more. We can follow up with any of those.

This might make the custom PRNG upgrade (#9263) easier too: the more `jnp` operations, the better.

cc @zhangqiaorjc

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1188,7 +1188,7 @@ def squeeze(array: Array, dimensions: Sequence[int]) -> Array:
 
 def expand_dims(array: Array, dimensions: Sequence[int]) -> Array:
   """Insert any number of size 1 dimensions into an array."""
-  ndim_out = np.ndim(array) + len(dimensions)
+  ndim_out = np.ndim(array) + len(set(dimensions))
   dims_set = frozenset(canonicalize_axis(i, ndim_out) for i in dimensions)
   result_shape = list(np.shape(array))
   for i in sorted(dims_set):
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1865,8 +1865,11 @@ def _squeeze(a, axis):
 
 @_wraps(np.expand_dims)
 def expand_dims(a, axis: Union[int, Sequence[int]]):
-  _check_arraylike("expand_dims", a)
-  return lax.expand_dims(a, _ensure_index_tuple(axis))
+  _stackable(a) or _check_arraylike("expand_dims", a)
+  axis = _ensure_index_tuple(axis)
+  if hasattr(a, "expand_dims"):
+    return a.expand_dims(axis)
+  return lax.expand_dims(a, axis)
 
 
 @_wraps(np.swapaxes, lax_description=_ARRAY_VIEW_DOC)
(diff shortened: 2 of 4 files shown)


LenaMartens (approved):
Thanks! (and thanks for fixing that concatenate bug!)

Just noting here that, ref. the issue description in https://github.com/google/jax/pull/8381: 

> I see two options to support these functions:
> - overload expand_dims and squeeze for PRNGKeyArrays
> - rewrite jnp.expand_dims/squeeze to use reshape/broadcast instead

we're going for option one (this has the advantage of preserving the information about singleton reshapes, at the cost of more overloads?)

jax/_src/lax/lax.py
```diff
@@ -1188,7 +1188,7 @@ def squeeze(array: Array, dimensions: Sequence[int]) -> Array:
 
 def expand_dims(array: Array, dimensions: Sequence[int]) -> Array:
   """Insert any number of size 1 dimensions into an array."""
-  ndim_out = np.ndim(array) + len(dimensions)
+  ndim_out = np.ndim(array) + len(set(dimensions))
```
jakevdp, line 1191: Numpy's version of this errors on repeated dimensions. Should we do the same?
```python
>>> x = np.arange(4)
>>> np.expand_dims(x, [1, 1, 1])                                                                                                                                                              
<ipython-input-4-3facbd77e287> in <module>
----> 1 np.expand_dims(x, [1, 1, 1])

~/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/numpy/core/overrides.py in expand_dims(*args, **kwargs)

~/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/numpy/lib/shape_base.py in expand_dims(a, axis)
    595 
    596     out_ndim = len(axis) + a.ndim
--> 597     axis = normalize_axis_tuple(axis, out_ndim)
    598 
    599     shape_it = iter(a.shape)

~/.local/share/virtualenvs/jax-LBbfM5ix/lib/python3.8/site-packages/numpy/core/numeric.py in normalize_axis_tuple(axis, ndim, argname, allow_duplicate)
   1394             raise ValueError('repeated axis in `{}` argument'.format(argname))
   1395         else:
-> 1396             raise ValueError('repeated axis')
   1397     return axis
   1398 

ValueError: repeated axis
```

froystig:
> we're going for option one (this has the advantage of preserving the information about singleton reshapes, at the cost of more overloads?)

Indeed. I'd like to try it out, at least.

jax/_src/lax/lax.py
froystig: Good observation. Let's try this is in a separate PR, in case it introduces new errors downstream?

jakevdp: Sounds good

jakevdp: Actually, I think it might be better to do it as part of this PR. It currently results in a less understandable error (because the output shape does not match), and if I'm reading it right, this change makes it so that it's no longer an error (by changing how the output shape is computed).

jakevdp: Oops, missed this before merge. Would you like to follow up or should I?

froystig: Done in #9615

PR #9617, opened 2022-02-17
in tests, compare jnp operations on PRNGKeyArrays to the same on jnp arrays

froystig:


LenaMartens (approved):
LGTM! although for a follow-up I feel some of these could be folded up into a parameterized test.

PR #9619, opened 2022-02-17
Remove jax.ops.index... functions.

hawkinsp:
These functions have been deprecated and have issued a DeprecationWarning since jax 0.2.22 in October 2021.


hawkinsp:
I suspect it will be a week or so before I can land this; I still need to make a number of cleanups.

jakevdp (changes requested):
Looks good! We'll also need to remove references from https://github.com/google/jax/blob/main/docs/jax.ops.rst and perhaps other places in the docs

hawkinsp:
> Looks good! We'll also need to remove references from https://github.com/google/jax/blob/main/docs/jax.ops.rst and perhaps other places in the docs

Done.

docs/jax.ops.rst
```diff
@@ -1,75 +1,15 @@
...
-=================
+===============
 
 .. currentmodule:: jax.ops
 
 .. automodule:: jax.ops
 
 .. _syntactic-sugar-for-ops:
 
-Indexed update operators
```
jakevdp, line 11: Keep lines 11-52?

hawkinsp: They have no reason to be in this particular module once the deprecation is complete. Is the reason to point people to the new names?

hawkinsp: I added a paragraph that points to jnp.ndarray.at.

gnecula:
It is rare to see a case when we simplify the API! Thanks!

PR #9622, opened 2022-02-17
Add an API compatibility policy.

hawkinsp:


docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
...
+* `jax.lax`
+* `jax.nn`
+* `jax.numpy`
+* `jax.ops`
+* `jax.profiler`
+* `jax.random`
+* `jax.scipy`
+* `jax.tree_util`
+
+Not everything in these modules is public. Over time, we are working to separate
```
yashk2810, line 31: Maybe move this after ## What is not covered list?

docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
...
+public and private APIs.
```
jakevdp, line 32: Mention that our aim is to follow the convention of hiding all non-public apis behind names that begin with an underscore (although we're not entirely compliant with that yet)

docs/api_compatibility.md
hawkinsp: This statement applies to the preceding list.

mattjj (approved):
Thanks for doing this. SGTM. I wonder if someone with experience in OSS deprecation policies could give any feedback, like @shoyer .

docs/api_compatibility.md
shoyer: I would also note that public APIs are documented in the JAX API docs

docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
+# API compatibility
+
+JAX is constantly evolving, and we are not yet ready to commit to a stable API.
+That said, we want to minimize churn for the JAX user community.
+
+JAX follows a 3 month deprecation policy. When an incompatible change is made
```
shoyer, line 6: 3 months is fast for an OSS project. The challenge is that you need to wait for downstream libraries to release fixes, and that can take time. It's not just about giving end-users time to upgrade.

6 months would be more in line with general norms. Definitely up to you, though, there are many tradeoffs here.

docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
...
+to an API, we will make our best effort to obey the following procedure:
+* the change will be announced in `CHANGELOG.md`, and the old API will issue a
+  `DeprecationWarning`.
```
shoyer, line 9: It's worth noting that DeprecationWarning is often silenced by default!

FutureWarning is a better choice for something you want end-users to see.

docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
+# API compatibility
+
+JAX is constantly evolving, and we are not yet ready to commit to a stable API.
+That said, we want to minimize churn for the JAX user community.
+
+JAX follows a 3 month deprecation policy. When an incompatible change is made
+to an API, we will make our best effort to obey the following procedure:
+* the change will be announced in `CHANGELOG.md`, and the old API will issue a
```
shoyer, line 8: Also in the docstring for the deprecated API

docs/api_compatibility.md
```diff
@@ -0,0 +1,47 @@
+# API compatibility
+
+JAX is constantly evolving, and we are not yet ready to commit to a stable API.
```
gnecula, line 3: This statement looks scarier than the reality. In reality, we remove APIs very rarely, and we think carefully before reducing functionality or making semantic changes. One can get a sense of the frequency of such changes by scanning the CHANGELOG. I think that this is worth pointing out.

I would also rewrite "not yet ready to commit to a stable API" to something like "and we want to be able to make improvements to the APIs".

docs/api_compatibility.md
gnecula: I also think that 3 months is fast. I expect that a lower bound on that interval will be our ability to migrate internal Google users off an API end point. Most likely a widely used API point will take much longer to remove.

docs/api_compatibility.md
hawkinsp: Both done.

docs/api_compatibility.md
hawkinsp: I left the deprecation period at three months with a comment that it is a lower bound and in practice deprecations may take longer. Also, nothing stops us from lengthening the period later. WDYT?

docs/api_compatibility.md
hawkinsp: PEP565 indicates that one should issue `DeprecationWarning` if the target of the warning is other python developers. Am I misunderstanding it or is it incorrect in practice?

hawkinsp:
PTAL.

docs/api_compatibility.md
gnecula: LGTM

gnecula (approved):

jakevdp (approved):

PR #9636, opened 2022-02-18
Checkify: address initial feedback.

LenaMartens:
- add a way to run checkify with no errors enabled
  - clarify "can't be staged" error message
  - export init_error: a way for users to set a default Error value


mattjj (approved):
Nice!

PR #9662, opened 2022-02-22
Adding a pytree registration function for dataclasses.

botev:
Adds a utility function to register a class as a dataclass and automatically register it is a pytree with all its fields as nodes.


froystig (commented):
The current PR relies on calling `dataclass` on the caller's behalf in a specific way (using all default options). If we're going to add to `tree_util`, I think we'd want to allow callers to configure a dataclass type however they like and supply it as input.

I'm not sure, but I suspect that handling general dataclasses would be more involved than what we'd want to maintain. Consider dataclasses with init-only variables, keyword-only variables, fields whose type is pytree-registered and not...

Do you have thoughts on this? Would you want to try generalizing this PR?

Note that we'd also need tests either way. The file `tests/tree_util_test.py` would be the place for them.

PR #9663, opened 2022-02-22
Checkify: fixing consts in scan and other control flow

mattjj:


LenaMartens:
@mattjj I've handled consts in the `cond` function as well (there was a complication with calling `cond` in the while_loop body, where we don't have the `cond_consts` available). PTAL.

jax/interpreters/partial_eval.py
```diff
@@ -1177,11 +1177,11 @@ def _reconstruct_pval(pval1: PartialVal, const2: core.Value):
 
 def move_binders_to_front(closed_jaxpr: ClosedJaxpr, to_move: Sequence[bool]) -> ClosedJaxpr:
   """Reorder the `invars` to move to front the ones for which `to_move` is True."""
-  assert not closed_jaxpr.jaxpr.constvars
+  # assert not closed_jaxpr.jaxpr.constvars
```
LenaMartens, line 1180: remove this?

Issue #9129, opened 2022-01-07
custom transposition


froystig:
Support custom transposition, i.e. the ability to register a custom "transpose rule" for any given function. Example usage would look something like:

```python
from functools import partial
from jax import custom_transpose, linear_transpose, numpy as jnp

def transpose_unary(f, x_example):
  def transposed(y):
    x, = linear_transpose(f, x_example)(y)
    return x
  return transposed

T = lambda f: transpose_unary(f, 0.)

# -- one degree of custom transposition

@custom_transpose
def f(z):
  return 2. * z

@f.def_transpose
def ft(z):
  return 3. * z

print(f(1.))              # 2.
print(T(f)(1.))           # 3.
print(T(T(f))(1.))        # 3.
print(T(T(T(f)))(1.))     # 3.
print(T(T(T(T(f))))(1.))  # 3. ...

# -- two degrees of custom transposition

@custom_transpose
def f(z):
  return 2. * z

@f.def_transpose
@custom_transpose
def ft(z):
  return 3. * z

@ft.def_transpose
def ftt(z):
  return 7. * z

print(f(1.))              # 2.
print(T(f)(1.))           # 3.
print(T(T(f))(1.))        # 7.
print(T(T(T(f)))(1.))     # 7.
print(T(T(T(T(f))))(1.))  # 7. ...

# -- symmetrically registered transposes (arbitrary degree)

@custom_transpose
def f(z):
  return 2. * z

@custom_transpose
def g(z):
  return 3. * z

f.def_transpose(g)
g.def_transpose(f)

print(f(1.))              # 2.
print(T(f)(1.))           # 3.
print(T(T(f))(1.))        # 2.
print(T(T(T(f)))(1.))     # 3.
print(T(T(T(T(f))))(1.))  # 2. ...

# recursively registered transposes (arbitrary degree)

@partial(custom_transpose, residual_argnums=[0])
def f(c, z):
  return c * z

@f.def_transpose
def ft(c, z):
  return f(c + 1., z)

g = partial(f, 1.)
print(g(1.))              # 1.
print(T(g)(1.))           # 2.
print(T(T(g))(1.))        # 3.
print(T(T(T(g)))(1.))     # 4.
print(T(T(T(T(g))))(1.))  # 5. ...
```

[...]

PR #9664, opened 2022-02-22
`custom_transpose` without staging

froystig:
Whereas the previous `custom_transpose` implementation staged its callable arguments upfront, this one preserves them as callables. For the time being, this requires callers to additionally supply the target function's output types at call time.

Part of #9129.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -6382,6 +6382,38 @@ def transpose_unary(f, x_example):
   return transposed
 
 
+# This class wraps api.custom_transpose in order to pass in a
+# particular tree of output type on each call. Otherwise it forwards
+# all attribute access.
+class _custom_transpose:
+  def __init__(self, out_types, fun):
+    self.out_types = out_types
+    self.fun = api.custom_transpose(fun)
+
+  def __getattr__(self, name):
+    return getattr(self.fun, name)
+
+  def __call__(self, *args):
+    return self.fun(self.out_types, *args)
+
+
+# This function is meant to be used as a decorator that delegates to
+# custom_transpose but makes it easy to specify output argument types
+# by example. If used directly a decorator (i.e. not invoked with
+# example arguments), assumes a scalar-valued function.
+#
+# TODO(frostig): remove this (and its uses) once custom_transpose offers
+# an option of inferring output types.
+def custom_transpose(example_out):
+  if isinstance(example_out, Callable):
+    out_type = core.get_aval(0.).at_least_vspace()
+    return _custom_transpose(out_type, example_out)
+  return partial(
+      _custom_transpose,
+      tree_util.tree_map(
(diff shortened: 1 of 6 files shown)


jax/_src/api_util.py
```diff
@@ -59,7 +58,18 @@ def _ensure_str_tuple(x: Union[str, Iterable[str]]) -> Tuple[str, ...]:
...
 
-@lu.transformation_with_aux
+# TODO(frostig,mattjj): consider "upstreaming" this to `linear_util`,
+# so that any transformation wrapped by `transformation_with_aux` can
+# choose to set the `pure` option (for an overwriting store), not only
+# the following flattening transformations. We have no such known uses
+# for now, though.
+def transformation_with_aux(xform):
+  def wrap_xform(*args, pure=False):
+    return lu.transformation_with_aux(xform, pure=pure)(*args)
```
mattjj, line 68: Can we change the 'pure' name? I'm not sure it's really descriptive, since these functions are always pure in the functional sense. How about "rerunnable" or "rerunnable_with_equal_aux" or something?

Actually might also be slightly cleaner not to change linear_util.py _or_ api_util.py in this PR, and just define these in custom_transpose.py:

```python
class StoreEqual:
  def store(self, val):
    if self._val is not lu_EMPTY_STORE_VALUE and val != self._val:
      raise lu.StoreException(f"Store assignment mismatch, from {self._val} to {val}")
    self._val = val

  @property
  def val(self):
    ...

def transformation_with_aux(gen, fun: lu.WrappedFun, *gen_static_args) -> Tuple[lu.WrappedFun, Any]:
  out_store = StoreEqual()
  out_thunk = lambda: out_store.val
  return fun.wrap(gen, gen_static_args, out_store), out_thunk
```

That way we don't need to decide on the name of a new kwarg, and we don't have to touch linear_util.py / api_util.py while this stuff is in flux.

WDYT? Anything I'm missing on why we'd want to change linear_util.py / api_util.py here?

(Also, since I've been slow to review, we can just land this and iterate if you prefer.)

tests/api_test.py
```diff
@@ -6377,6 +6377,38 @@ def transposed(y):
   return transposed
 
 
+# This class wraps api.custom_transpose in order to pass in a
```
mattjj, line 6380: Maybe not in this PR, but given that api_test.py is gigantic, I'm starting to think we should factor out CustomTransposeTest (and perhaps CustomJVPTest / CustomVJPTest).

mattjj (approved):
Nice! One suggestion to simplify this PR and not have it touch api_test.py / linear_util.py.

jax/_src/api_util.py
froystig: Great idea. Fixed, namely forked the three necessary functions over in custom_transpose.py.

PR #9683, opened 2022-02-24
refactoring code with List Comprehension

idiomaticrefactoring:
refactoring code with List Comprehension which is more pythonic, concise and efficient;how do think this change which has practical value？

```
static_idx=[i for i in range(1000)]
dynamic_idx=[i for i in range(1000)]
code='''
idx = []
for (s, d) in zip(static_idx, dynamic_idx):
    if d is not None:
        idx.append(d)
    elif isinstance(s, tuple):
        idx.append(slice(s[0], s[1], s[2]))
    else:
        idx.append(s)
'''
total_time_1 = min(
    timeit.repeat(code, globals={**globals(), **locals()},
                  repeat=1, number=1000))
pythonic_code='''
idx = [d if d is not None else slice(s[0], s[1], s[2]) if isinstance(s, tuple) else s for (s, d) in zip(static_idx, dynamic_idx)]
'''
total_time_2 = min(
    timeit.repeat(pythonic_code,
                  globals={**globals(), **locals()},
                  repeat=1, number=1000))
print("total time: ", total_time_2, total_time_1, total_time_1 / total_time_2)
```

output:
0.05620845900000049 0.08211149999999989 1.4608388392216762

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5880,14 +5880,11 @@ def _split_index_for_jit(idx, shape):
 
 def _merge_static_and_dynamic_indices(treedef, static_idx, dynamic_idx):
   """Recombines indices that were split by _split_index_for_jit."""
-  idx = []
-  for s, d in zip(static_idx, dynamic_idx):
-    if d is not None:
-      idx.append(d)
-    elif isinstance(s, tuple):
-      idx.append(slice(s[0], s[1], s[2]))
-    else:
-      idx.append(s)
+  idx = [d if d is not None 
+         else slice(s[0], s[1], s[2]) if isinstance(s, tuple) 
+         else s 
+         for (s, d) in zip(static_idx, dynamic_idx)]
+  
   return treedef.unflatten(idx)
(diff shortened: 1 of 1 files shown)


jakevdp:
Hi - thanks for looking into this. Improving JAX trace times is a great goal! That said, at first blush, I'm not convinced that microbenchmarks at this level are the best way to approach reducing JAX's overhead. Yes it's a ~30% improvement in runtime, but that improvement amounts to ~30µs per run when parsing 1000 static and 1000 dynamic indices, which I suspect is far more than are ever parsed in practice. If we assume the savings is linear in the number of parsed indices, then I expect the typical savings for a real use-case would be perhaps tens of nanoseconds for a typical gather operation. That's not zero, but I don't think it's something worth dwelling on in the overall computational cost of executing a JAX program.

Now I may be wrong... it would be interesting to see a line profiler output of a typical JAX computation, and see which operations are contributing the most. If `_merge_static_and_dynamic_indices` is high in that, then this may be a good change to consider.

hawkinsp:
Closing as stale.

PR #9689, opened 2022-02-25
Return consistent results for image.scale_and_translate with negative dims

fehiepsi:
Currently, when specified negative dimensions, we get some weird results, probably because `jnp.einsum` expected non-negative numbers for contractions. This PR fixes the issue and adds a regression test.

### DIFF
--- a/jax/_src/image/scale.py
+++ b/jax/_src/image/scale.py
@@ -90,9 +90,8 @@ def _scale_and_translate(x, output_shape: core.Shape,
   contractions = []
   in_indices = list(range(len(output_shape)))
   out_indices = list(range(len(output_shape)))
-  spatial_ndim = len(spatial_dims)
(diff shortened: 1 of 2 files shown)


jax/_src/image/scale.py
```diff
@@ -89,7 +89,10 @@ def _scale_and_translate(x, output_shape: core.Shape,
   contractions = []
   in_indices = list(range(len(output_shape)))
   out_indices = list(range(len(output_shape)))
+  spatial_ndim = len(spatial_dims)
   for i, d in enumerate(spatial_dims):
+    if d < 0:
```
hawkinsp, line 94: Use `jax._src.util.canonicalize_axis`

hawkinsp (approved):

hawkinsp:
Sorry for the slow review, I was on vacation.

PR #9711, opened 2022-02-28
Remove an unnecessary condition.

jblespiau:
This makes sure jit, xla_computation and pjit share the same logic for processing static arguments.

### DIFF
--- a/jax/experimental/pjit.py
+++ b/jax/experimental/pjit.py
@@ -218,11 +218,7 @@ def pjit(fun: Callable,
       raise RuntimeError("pjit only supports GPU and TPU devices")
 
     f = lu.wrap_init(fun)
-    if static_argnums:
-      f, dyn_args = argnums_partial_except(
-          f, static_argnums, args, allow_invalid=False)
-    else:
-      dyn_args = args
+    f, dyn_args = argnums_partial_except(f, static_argnums, args, allow_invalid=False)
     del args
 
     args_flat, in_tree = tree_flatten(dyn_args)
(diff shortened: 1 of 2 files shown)


jax/experimental/pjit.py
```diff
@@ -218,11 +218,7 @@ def infer_params(*args, **kwargs):
       raise RuntimeError("pjit only supports GPU and TPU devices")
 
     f = lu.wrap_init(fun)
-    if static_argnums:
-      f, dyn_args = argnums_partial_except(
-          f, static_argnums, args, allow_invalid=False)
-    else:
-      dyn_args = args
```
apaszke, line 225: It might be logically equivalent, but I'd expect that this path is quite a bit faster. This function runs every time a pjit'ed computation is executed.

jblespiau: Note that the first line of the `argnums_partial_except` is:

```
if not static_argnums:
  return f, args
```
which is exactly this condition (so we don't get through the heavy work). So it's the same, modulo: (a) an additional un-tupling of a tuple. (b) An additional function-call.

I used:

```
    times = []
    import time
    for _ in range(30):
      t = time.time()
      for _ in range(1000):
        f(x, x + 1).block_until_ready()
      times.append(time.time() - t)

    raise ValueError(np.mean(times), np.std(times))
 ```
 
 and I got (I ran 2 times):
 
Before: (mean, std)  
(1.2995330731074015, 0.15805728929291446)
(1.2651569207509359, 0.15750691481188572)

After:  
(1.2485803842544556, 0.13929473385469665)
ValueError: (1.301744771003723, 0.14164496885050082)

So there is probably no difference statistically measurable.

apaszke (approved):
Ah, I didn't realize this conditional was already there. LGTM!

PR #9723, opened 2022-03-01
Add simple effect types to Jaxprs (based on #8699)

sharadmv:
Adds simple effect types to Jaxprs (based on #8699).

This PR is adds a new fields to `Jaxpr`s and `JaxprEqn`s, `effects`, which is a `Set[Any]`. This PR adds the relevant threading of `effects` through the Jaxpr transformations and also adds a `.replace` convenience method for `JaxprEqn`s, `Jaxpr`s and `ClosedJaxpr`s.

Effects are added to Jaxpr at trace time via the `abstract_eval` rule, whose return values are augmented from `List[Aval]` to `Tuple[List[Aval], Set[Any]]`. Rather than rewriting every `abstract_eval` rule, we have `def_abstract_eval` wrap the provided function with one that additionally returns an empty set. Primitives now have a new `def_effectful_abstract_eval` that enables providing `abstract_eval` methods that return both avals and effects.

This PR also disallows effects in control flow primitives (i.e. `cond`, `while_loop`, and `scan`) by erroring if an effect is detected after tracing the body functions to jaxprs.

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -109,8 +109,7 @@ def _initial_style_jaxprs_with_common_consts(
     prefix = util.concatenate(unused_const_vars[:i])
     suffix = util.concatenate(unused_const_vars[i + 1:])
     constvars = [*prefix, *jaxpr.constvars, *suffix]
-    return core.Jaxpr(constvars=constvars, invars=jaxpr.invars,
-                      outvars=jaxpr.outvars, eqns=jaxpr.eqns)
+    return jaxpr.replace(constvars=constvars)
 
   consts = util.concatenate(all_consts)
   jaxprs = [pad_jaxpr_constvars(i, jaxpr) for i, jaxpr in enumerate(jaxprs)]
@@ -315,6 +314,9 @@ def while_loop(cond_fun: Callable[[T], BooleanNumeric],
     new_init_val, = tree_unflatten(in_tree, new_init_vals)
     init_vals, init_avals, body_jaxpr, in_tree, *rest = _create_jaxpr(new_init_val)
   cond_jaxpr, cond_consts, body_consts, body_tree = rest
+  joined_effects = core.join_effects(body_jaxpr.effects, cond_jaxpr.effects)
+  if joined_effects:
+    raise NotImplementedError('Effects not supported in `while`.')
 
   in_tree_children = in_tree.children()
   assert len(in_tree_children) == 1
@@ -326,8 +328,9 @@ def while_loop(cond_fun: Callable[[T], BooleanNumeric],
                       body_nconsts=len(body_consts), body_jaxpr=body_jaxpr)
   return tree_unflatten(body_tree, outs)
 
-def _while_loop_abstract_eval(*args, **kwargs):
-  return _map(raise_to_shaped, kwargs["body_jaxpr"].out_avals)
+def _while_loop_abstract_eval(*args, body_jaxpr, **kwargs):
+  del args, kwargs
+  return _map(raise_to_shaped, body_jaxpr.out_avals), core.no_effects
 
 def _while_loop_translation_rule(ctx, avals_in, avals_out, *args, cond_jaxpr,
                                  body_jaxpr, cond_nconsts, body_nconsts):
@@ -516,7 +519,8 @@ def _while_loop_jvp(primals, tangents, cond_nconsts, cond_jaxpr, body_nconsts,
   cond_jaxpr_augmented = core.Jaxpr(cond_jaxpr.jaxpr.constvars,
                                     invars_aug,
                                     cond_jaxpr.jaxpr.outvars,
-                                    cond_jaxpr.jaxpr.eqns)
+                                    cond_jaxpr.jaxpr.eqns,
+                                    cond_jaxpr.jaxpr.effects)
   cond_jaxpr_augmented = core.ClosedJaxpr(cond_jaxpr_augmented, cond_jaxpr.consts)
 
   out = while_p.bind(
@@ -611,7 +615,7 @@ def _while_transpose_error(*_, **kwargs):
 while_p = core.AxisPrimitive('while')
 while_p.multiple_results = True
 while_p.def_impl(partial(xla.apply_primitive, while_p))
-while_p.def_abstract_eval(_while_loop_abstract_eval)
+while_p.def_effectful_abstract_eval(_while_loop_abstract_eval)
 ad.primitive_jvps[while_p] = _while_loop_jvp
 pe.custom_partial_eval_rules[while_p] = _while_partial_eval
 xla.register_translation(while_p, _while_loop_translation_rule,
@@ -875,6 +879,9 @@ def _cond(pred, true_fun: Callable, false_fun: Callable, *operands,
 
   jaxprs, consts, out_trees = _initial_style_jaxprs_with_common_consts(
       (true_fun, false_fun), ops_tree, ops_avals, 'cond')
+  joined_effects = core.join_effects(*(jaxpr.effects for jaxpr in jaxprs))
+  if joined_effects:
+    raise NotImplementedError('Effects not supported in `cond`.')
   true_jaxpr, false_jaxpr = jaxprs
   out_tree, false_out_tree = out_trees
 
@@ -926,8 +933,8 @@ def _cond_with_per_branch_args(pred,
                lambda op: false_fun(op[1]),
                (true_operand, false_operand))
 
-def _cond_abstract_eval(*args, **kwargs):
-  return _map(raise_to_shaped, kwargs["branches"][0].out_avals)
+def _cond_abstract_eval(*args, branches, **kwargs):
+  return _map(raise_to_shaped, branches[0].out_avals), core.no_effects
 
 def _cond_translation_rule(ctx, avals_in, avals_out, index, *args, branches,
                            linear):
@@ -1121,11 +1128,13 @@ def _cond_partial_eval(trace, *tracers, branches, linear):
 
   linear_2 = (False,) * num_res + linear
   params = dict(branches=branches_2, linear=linear_2)
+  if any((branch.effects for branch in branches_2)):
+    raise NotImplementedError('Effects not supported in `cond`.')
   name_stack = source_info_util.current_name_stack()[len(trace.name_stack):]
   source = source_info_util.current().replace(name_stack=name_stack)
   eqn = pe.new_eqn_recipe(
       [index_tracer] + res_tracers + ops_tracers, out_tracers, cond_p, params,
-      source)
+      core.no_effects, source)
   for t in out_tracers: t.recipe = eqn
   return out_tracers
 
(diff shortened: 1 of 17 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -1385,6 +1396,8 @@ def _cond_typecheck(*avals, branches, linear):
     raise core.JaxprTypeError(
       f'cond branches take input types {jaxpr0_in_avals_str}, '
       f'called with operands of type {_avals_short(op_avals)}')
+  effects = core.join_effects((b.effects for b in branches))
```
mattjj, line 1399: I think we want to check that the effects are the same across branches here. That is, something like

```python
if not all(branches[0].effects == b.effects for b in branches[1:]):
  raise ...
```

jax/core.py
```diff
@@ -91,6 +96,18 @@ def pretty_print(self, *, source_info=False, print_shapes=True,
...
+    constvars = constvars or self.constvars
+    invars = invars or self.invars
+    outvars = outvars or self.outvars
+    eqns = eqns or self.eqns
+    effects = effects or self.effects
+    return Jaxpr(constvars=constvars, invars=invars, outvars=outvars, eqns=eqns,
+                 effects=effects)
+
+def join_effects(effects: Iterable[Effects]) -> Effects:
+  return functools.reduce(lambda a, b: a | b, effects, set())
```
mattjj, line 110:
```suggestion
  return set.union(*effects)
```

(Though TBH I kind of like not introducing a new abstraction in core.py and instead just calling `set.union` where `core.join_effects` is currently called. Up to you!)

jax/core.py
```diff
@@ -161,17 +187,22 @@ class JaxprEqn(NamedTuple):
...
   def __repr__(self):
     return str(pp_eqn(self, JaxprPpContext(), custom_pp_eqn_rules=False)
                ).rstrip()
 
-def new_jaxpr_eqn(invars, outvars, primitive, params, source_info=None):
+  def replace(self, *args, **kwargs):
+    return self._replace(*args, **kwargs)
+
+def new_jaxpr_eqn(invars, outvars, primitive, params, effects, source_info=None):
+  effects = effects or set()
```
mattjj, line 201: Do we want to support any falsey value to be passed to new_jaxpr_eqn? Maybe better not to add that polymorphism and require the caller to pass `set()`.

jax/experimental/djax.py
```diff
@@ -954,9 +955,10 @@ def new_res(v):
     unks = map(read, eqn.invars)
     if any(unks):
       invars = [v if unk else new_res(v) for unk, v in zip(unks, eqn.invars)]
-      eqns2.append(pe.new_jaxpr_eqn(invars, eqn.outvars, eqn.primitive,
-                                    eqn.params,
-                                    source_info_util.new_source_info()))
+      eqns2.append(core.new_jaxpr_eqn(invars, eqn.outvars, eqn.primitive,
+                                      eqn.params,
+                                      None,
```
mattjj, line 960:
```suggestion
                                      set(),
```

mattjj: An alternative to writing `set()` literals everywhere would be to define a `no_effects` variable and use that.

Also maybe join the lines here.

jax/_src/lax/control_flow.py
```diff
@@ -326,8 +330,9 @@ def _create_jaxpr(init_val):
                       body_nconsts=len(body_consts), body_jaxpr=body_jaxpr)
   return tree_unflatten(body_tree, outs)
 
-def _while_loop_abstract_eval(*args, **kwargs):
-  return _map(raise_to_shaped, kwargs["body_jaxpr"].out_avals)
+def _while_loop_abstract_eval(*args, body_jaxpr, **kwargs):
+  del args, kwargs
+  return _map(raise_to_shaped, body_jaxpr.out_avals), body_jaxpr.effects
```
mattjj, line 335: Can the cond_jaxpr have effects? The code above suggests yes. Then the effect of applying a while_loop isn't just the effect of the body.

I think for this PR we should just raise NotImplementedError in the `while_loop` traceable if the jaxprs formed have any effects at all. Then let's add support for effects to each control flow primitive one by one.

jax/_src/lax/control_flow.py
```diff
@@ -315,6 +316,9 @@ def _create_jaxpr(init_val):
     new_init_val, = tree_unflatten(in_tree, new_init_vals)
     init_vals, init_avals, body_jaxpr, in_tree, *rest = _create_jaxpr(new_init_val)
   cond_jaxpr, cond_consts, body_consts, body_tree = rest
+  joined_effects = core.join_effects([body_jaxpr.effects, cond_jaxpr.effects])
+  body_jaxpr = body_jaxpr.replace(jaxpr=body_jaxpr.jaxpr.replace(effects=joined_effects))
+  cond_jaxpr = cond_jaxpr.replace(jaxpr=cond_jaxpr.jaxpr.replace(effects=joined_effects))
```
mattjj, line 321: Also mentioned below: let's just raise a NotImplementedError if any control flow primitives get jaxprs with effects in them (in the traceables, maybe also in the typecheck rules). Getting these right will take some care so we should probably add support for control flow primitives one-by-one in follow-up PRs.

To put a fine point on it: I think the only changes this PR should make to control_flow.py should be to raise NotImplementedErrors (if this PR makes any changes at all).

jax/_src/lax/lax.py
```diff
@@ -2715,7 +2715,7 @@ def _broadcast_in_dim_staging_rule(
   out_tracer = pe.DynamicJaxprTracer(trace, aval, source_info)
   invars = [trace.getvar(x), *(trace.getvar(d) for d in dyn_shape)]
   eqn = pe.new_jaxpr_eqn(invars, [trace.makevar(out_tracer)],
-                         broadcast_in_dim_p, params, source_info)
+                         broadcast_in_dim_p, params, None, source_info)
```
mattjj, line 2718:
```suggestion
                         broadcast_in_dim_p, params, set(), source_info)
```

jax/experimental/djax.py
```diff
@@ -1326,8 +1328,8 @@ def _iota_staging_rule(trace, tracers, params):
     out_aval = core.ShapedArray((n,), np.dtype('int32'))
     out_tracer = pe.DynamicJaxprTracer(trace, out_aval, None)
     outvar = trace.makevar(out_tracer)
-    eqn = pe.new_jaxpr_eqn([], [outvar], iota_p, dict(size=n),
-        source_info_util.new_source_info())
+    eqn = core.new_jaxpr_eqn([], [outvar], iota_p, dict(size=n),
+        None, source_info_util.new_source_info())
```
mattjj, line 1332:
```suggestion
        set(), source_info_util.new_source_info())
```

jax/experimental/djax.py
```diff
@@ -1340,8 +1342,8 @@ def _iota_staging_rule(trace, tracers, params):
     out_tracer = pe.DynamicJaxprTracer(trace, out_aval, None)
     outvar = trace.makevar(out_tracer)
     invar = trace.getvar(tracer)
-    eqn = pe.new_jaxpr_eqn([invar], [outvar], iota_p, {},
-        source_info_util.new_source_info())
+    eqn = core.new_jaxpr_eqn([invar], [outvar], iota_p, {},
+        None, source_info_util.new_source_info())
```
mattjj, line 1346:
```suggestion
        set(), source_info_util.new_source_info())
```

jax/experimental/djax.py
```diff
@@ -1389,9 +1391,10 @@ def _broadcast_staging_rule(trace, tracers, params):
     dtype = aval._eltTy._dtype if isinstance(aval, AbsArray) else aval.dtype
     out_aval = AbsArray((d, *x.shape), BaseType(dtype))
     out_tracer = pe.DynamicJaxprTracer(trace, out_aval, None)
-    eqn = pe.new_jaxpr_eqn([trace.getvar(x), trace.getvar(d)],
-                           [trace.makevar(out_tracer)], broadcast_p, {},
-                           source_info_util.new_source_info())
+    eqn = core.new_jaxpr_eqn([trace.getvar(x), trace.getvar(d)],
+                             [trace.makevar(out_tracer)], broadcast_p, {},
+                             None,
```
mattjj, line 1396:
```suggestion
                             set(),
```

jax/experimental/host_callback.py
```diff
@@ -1420,10 +1420,10 @@ def _rewrite_jaxpr(jaxpr: core.Jaxpr, has_input_token: bool,
     # We need tokens but none is given in input; make one depending on all invars
     eqns.append(
         core.new_jaxpr_eqn(jaxpr.invars, [last_token_var],
-                           lax.create_token_p, {}, source_info_util.current()))
+                           lax.create_token_p, {}, None, source_info_util.current()))
```
mattjj, line 1423:
```suggestion
                           lax.create_token_p, {}, set(), source_info_util.current()))
```

jax/experimental/host_callback.py
```diff
@@ -1420,10 +1420,10 @@ def _rewrite_jaxpr(jaxpr: core.Jaxpr, has_input_token: bool,
...
     eqns.append(
         core.new_jaxpr_eqn(jaxpr.invars, [last_itoken_var],
-                           lax.create_token_p, {}, source_info_util.current()))
+                           lax.create_token_p, {}, None, source_info_util.current()))
```
mattjj, line 1426:
```suggestion
                           lax.create_token_p, {}, set(), source_info_util.current()))
```

mattjj (approved):
LGTM, my main suggestion is not to touch lax_control_flow.py in this PR (or just add NotImplementedErrors to it).

The eqn.replace / jaxpr.replace stuff is a great QOL improvement!

jax/_src/lax/control_flow.py

jax/core.py
sharadmv: Done.

jax/experimental/djax.py
sharadmv: Done.

jax/_src/lax/control_flow.py

jax/_src/lax/lax.py
sharadmv: Done.

jax/_src/lax/control_flow.py

jax/core.py
sharadmv: Done.

tests/jaxpr_effects_test.py
```diff
@@ -0,0 +1,133 @@
+# Copyright 2021 Google LLC
```
mattjj, line 1:
```suggestion
# Copyright 2022 Google LLC
```

mattjj (approved):

jax/experimental/host_callback.py
```diff
@@ -1476,11 +1476,9 @@ def _rewrite_eqn(eqn: core.JaxprEqn, eqns: List[core.JaxprEqn],
   """
```
gnecula, line 1476: This whole code ought to disappear once we have support for effects in Jaxpr.

sharadmv: Yes, we should be able to rewrite HCB once we can lower effects away with tokens.

jax/core.py
```diff
@@ -54,14 +54,21 @@
...
+Effects = Set[Effect]
+
+no_effects: Effects = set()
+
 class Jaxpr:
   constvars: List[Var]
   invars: List[Var]
   outvars: List[Atom]
   eqns: List[JaxprEqn]
+  effects: Effects
```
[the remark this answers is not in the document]
sharadmv: I've added some more detail to the constructor docstring.

jax/experimental/host_callback.py
sharadmv: Yes, that's the goal

PR #9737, opened 2022-03-01
cleanup: remove _constant_like in favor of lax._const

jakevdp:
`lax._const` differs slightly in that it will return a Python scalar value if the input is a scalar type, but this seems incidental. I don't think there's any reason to have multiple versions of this utility

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -321,7 +321,7 @@ def cholesky_jvp_rule(primals, tangents):
   # Forward-mode rule from https://arxiv.org/pdf/1602.07527.pdf
   def phi(X):
     l = jnp.tril(X)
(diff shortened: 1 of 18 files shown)


froystig (approved):
In modules where you're switching to using lax's `_const`, would it make sense to import it from `jax._src.lax.lax` instead of accessing it via the public `lax` module (as `lax._const`)? As I understand it, we may not want the symbol to remain in the latter over the long term (?).

jakevdp:
> As I understand it, we may not want the symbol to remain in the latter over the long term (?).

Maybe? I think I'd prefer to address that separately, as it will probably be a larger change.

froystig:
I didn't mean to suggest the larger change in this PR, but rather that we consider importing `_const` in the desired way going forward, i.e. at least for the calls introduced in this change. In other words, where this change calls `lax._const`, instead have it call `_lax_const` after issuing `from jax._src.lax import _const as _lax_const`, or something along those lines.

PR #9746, opened 2022-03-02
Call _check_arraylike on inputs to broadcast_to and broadcast_arrays

jakevdp:
Part of #7737

A list being passed to `broadcast_to` was the cause of the rollback of #9724


hawkinsp (approved):
Technically we might want to go through a deprecation period, but I think it's probably fine in this case.

PR #9763, opened 2022-03-03
Fix the display of GDA docs

yashk2810:


jakevdp:
This will result in the documentation for both classes being on the same page, rather than on separate pages. Is that your intent?

yashk2810:
I wanted to fix the from_callback and other methods that were not being displayed properly.

This fixes it. Is there a better way to do this?

yashk2810:
I am fine with both being on the same page (atleast for now) until I figure out how to fix the display problem.

But the display problem needs to be fixed first.

jakevdp:
There might be a way to configure this using [`autosummary`](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#directive-autosummary)

yashk2810:
There might be, but I don't have the bandwidth to figure that out right now. I can do it later and in the meantime this workaround should be fine.

PR #9772, opened 2022-03-04
custom_prng: better error messages for key validation

jakevdp:


froystig (changes requested):
`PRNGKeyArray` is not meant to be public-facing, in the sense that it is not expected that anyone outside of JAX's internals call its constructor. The checks it runs are necessary, but internal, and I'd rather we not draw outside attention to `PRNGKeyArray` for being what caught them. Instead of prefixing errors with its name (`PRNGKeyArray: expected ...`), how about something like the following?
```
JAX encountered invalid PRNG key data: missing ndim, shape, or dtype attributes. Got {key_data}
```
Either way, triggering these sort of failures from user code is much less likely to happen after #9263, since roughly no one will handle raw key data outside of our internals. Failing these checks would more likely indicate a bug in jax itself (think of these as internal assertions of our own intended invariants).

If we want more useful errors on invalid key data prior to #9263, we can look for places to add them closer to public functions in `jax.random`, where we have more context.

jakevdp:
Thanks for the context, so you think we should leave these errors as-is? If we want to go one level up we could check this in `_check_prng_key` which is called by the user-level functions - would that be better? (currently `_check_prng_key` calls `PRNGKeyArray` so it would just be moving the check up a single level)

froystig:
I think we want these particular checks to live in the `PRNGKeyArray` constructor, even after #9263, since we want to be sure they hold at that point. I wouldn't hoist them to `jax.random`, but we can add more specific heuristic checks there if we'd like. I'd suggest we:

1. take your proposed change with a different error message prefix, e.g. `JAX encountered invalid PRNG key data: ...`;
2. **[optional]** add any heuristic checks to `_check_prng_key` that specifically look for gotchas that can happen in the interim. For example, since my understanding is that we were motivated by this: if key data is scalar we can raise with something along the lines of "are you using `vmap` and did you mean to pass in a batch of keys?"

I say item 2 is optional since `_check_prng_key` will be drastically simplified after #9263 (to an `isinstance` check, if anything). Thoughts?

jakevdp:
OK, PTAL. I basically did (1), I think it makes sense to punt on (2) because things will be changing so significantly.

froystig (approved):
Seems like the right call. Thanks!

PR #9786, opened 2022-03-07
improve batch_jaxpr caching from #9196

mattjj:
In #9196 I missed a related utilty function which needed memoization.

### DIFF
--- a/jax/interpreters/batching.py
+++ b/jax/interpreters/batching.py
@@ -448,7 +448,6 @@ def batch_jaxpr(closed_jaxpr, axis_size, in_batched, instantiate, axis_name,
   return _batch_jaxpr(closed_jaxpr, axis_size, tuple(in_batched), inst,
                       axis_name, main_type)
 
-@cache()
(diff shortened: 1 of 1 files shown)


jax/interpreters/batching.py
```diff
@@ -463,6 +462,12 @@ def _batch_jaxpr(closed_jaxpr, axis_size, in_batched, instantiate, axis_name,
 
 def batch_jaxpr_axes(closed_jaxpr, axis_size, in_axes, out_axes_dest, axis_name,
                      main_type):
+  return _batch_jaxpr_axes(closed_jaxpr, axis_size, tuple(in_axes),
+                           tuple(out_axes_dest), axis_name, main_type)
```
apaszke, line 466: nit: it would be a bit faster to already make in_axes and out_axes_dest above tuples, but lgtm

apaszke (approved):

jax/interpreters/batching.py
mattjj: True but I thought I'd err on making things easier for the caller.

PR #9789, opened 2022-03-07
[sparse]: COO: check for sorted rows before cusparse lowering

jakevdp:
It's been brought to our attention that cusparse and hipsparse require COO inputs to be sorted row-wise, or else the output of gpu translations of functions like `coo_todense`, `coo_matvec`, `coo_matmat`, etc. may be incorrect or even lead to a segfault.

Unfortunately, the sorting opertaion is rather expensive, so pre-emptively sorting all inputs has significant performance impacts.

The solution we landed on here is to carry a static `rows_sorted` flag with the matrix, so that in the cases that we *know* rows are sorted, we can silently employ the GPU translations, and in cases where it's unclear we can fall back to the slower reference implementation, while warning the user that this is happening.

### DIFF
--- a/jax/experimental/sparse/coo.py
+++ b/jax/experimental/sparse/coo.py
@@ -21,6 +21,7 @@ import warnings
 import numpy as np
 
 from jax import core
+from jax import lax
 from jax.interpreters import ad
 from jax.interpreters import xla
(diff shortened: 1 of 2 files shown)


tlu7:
@tlu7

froystig (approved):

PR #9809, opened 2022-03-09
Switch gpu, cpu and jax2tf to use the new OptimizationBarrier op.

pschuh:
This change should unblock the removal of the cond and while widgets.


hawkinsp (commented):
The change looks ok, but there appear to be CI failures (crashes)?

hawkinsp (approved):
Please run global tests for this.

PR #9815, opened 2022-03-09
Fix fft dtype for norm='ortho'

eelregit:
FFT orthogonal normalization factor dtype was undefined.

### DIFF
--- a/jax/_src/numpy/fft.py
+++ b/jax/_src/numpy/fft.py
@@ -86,7 +86,8 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
         s += [max(0, 2 * (a.shape[axes[-1]] - 1))]
     else:
       s = [a.shape[axis] for axis in axes]
-  transformed = lax.fft(a, fft_type, tuple(s)) * _fft_norm(jnp.array(s), func_name, norm)
+  transformed = lax.fft(a, fft_type, tuple(s))
+  transformed *= _fft_norm(jnp.array(s, dtype=transformed.real.dtype), func_name, norm)
 
   if orig_axes is not None:
     transformed = jnp.moveaxis(transformed, axes, orig_axes)
--- a/tests/fft_test.py
+++ b/tests/fft_test.py
@@ -138,6 +138,11 @@ class FftTest(jtu.JaxTestCase):
       tol = 0.15
       jtu.check_grads(jnp_fn, args_maker(), order=2, atol=tol, rtol=tol)
 
+    # check dtypes
+    dtype = jnp_fn(rng(shape, dtype)).dtype
+    expected_dtype = jnp.promote_types(float if inverse and real else complex, dtype)
+    self.assertEqual(dtype, expected_dtype)
+
   def testIrfftTranspose(self):
     # regression test for https://github.com/google/jax/issues/6223
     def build_matrix(linear_func, size):


jakevdp (approved):
Looks good, thanks!

jakevdp:
Hmm... looks like there are some new test failures. Do you understand why that is?

eelregit:
Not really sure...

allclose failed with some quite big differences (8 vs 64), for a change of dtype. That's strange.

shoyer:
A new unit test that passes only with this change would probably also be a good idea, to ensure we don't have a regression.

jakevdp:
It looks like the failure has input of type `bool`, which suggests that the new dtype casting may be throwing away information. You should probably cast to the expected output type rather than the input type.

eelregit:
I added some checks as part of `testFftn` (with some duplicated computations...)

tests/fft_test.py
```diff
@@ -138,6 +138,12 @@ def testFftn(self, inverse, real, shape, dtype, axes, s, norm):
       tol = 0.15
       jtu.check_grads(jnp_fn, args_maker(), order=2, atol=tol, rtol=tol)
 
+    # check dtypes
+    a = rng(shape, dtype)
+    dtype = jnp_fn(a).dtype
+    expected_dtype = jnp.result_type(0. if inverse and real else 0j, dtype)
+    self.assertEqual(dtype, expected_dtype)
```
jakevdp, line 145: Could we cover this by setting `check_dtypes=True` in the Numpy check above?

eelregit: It seems that numpy always upcasts https://github.com/numpy/numpy/issues/17801

jax/_src/numpy/fft.py
```diff
@@ -86,7 +86,9 @@ def _fft_core(func_name, fft_type, a, s, axes, norm):
         s += [max(0, 2 * (a.shape[axes[-1]] - 1))]
     else:
       s = [a.shape[axis] for axis in axes]
-  transformed = lax.fft(a, fft_type, tuple(s)) * _fft_norm(jnp.array(s), func_name, norm)
+  norm_dtype = jnp.result_type(0., a.real.dtype)
```
jakevdp, line 89: This is not exactly the same as the type promotion used within `lax.fft`... perhaps it would be simpler to first compute `lax.fft` and then use this dtype to compute `_fft_norm` rather than doing the type promotion in two places?

eelregit: The normalization factor is always real though. Do you think it's worth converting it to complex value and do complex multiplications (for most of the cases)?

eelregit: Actually, I don't know if (complex x real) and (complex x complex) have the same cost

jakevdp: To be clear, what I meant is something like this:
```python
transformed = lax.fft(a, fft_type, tuple(s))
transformed *= _fft_norm(jnp.array(s, dtype=transformed.real.dtype), func_name, norm)
```

tests/fft_test.py
jakevdp: I see... perhaps we could use something like the `_promote_like_jnp` decorator to address this? (I think we'd have to add a `complex=False` argument to it similar to the existing `inexact=False` to ensure complex output)

jax/_src/numpy/fft.py
eelregit: I see. That's much cleaner. I have changed it.

tests/fft_test.py
eelregit: hmm... this is certainly beyond my knowledge of jax :)

jakevdp: Oh, this isn't really about JAX. I was just refering to a test utility that is used in some places, e.g. https://github.com/google/jax/blob/cf9a900d7879a1d8c7954c0687a5c764801d08df/tests/lax_numpy_test.py#L586

You could use this same test utility to modify `np.fft` so that its return type matches what is expected for JAX, and then we could set `check_dtypes=True` rather than having a separate block of code to check dtypes.

jakevdp: If that's too much to bite off, then we can merge this as-is and I can clean up the test in a followup commit.

eelregit: That sounds good. Very interested in see how it's done in the followup commit to learn more about jax!

jakevdp (changes requested):
OK, looks good.

The last thing we need is to squash all the changes into a single commit – see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests for details and let me know if you need assistance with that.

jakevdp (approved):

jakevdp:
Thanks!

PR #9830, opened 2022-03-10
Checkify: Fix empty enabled_errors case wrt user checks.

LenaMartens:
What looked like a quick win (short-cut on empty enabled_error, don't trace)
was actually a quick bug (user_checks throw error in eager).


mattjj (approved):
Quick bug fixes are the best!

PR #9838, opened 2022-03-10
Fix core.pytype_aval_mappings for _DeviceArray

shawwn:
Due to a typo, `core.pytype_aval_mappings` wasn't being created for `_DeviceArray`. This manifests itself when users (like me!) try to implement their own custom JAX backends, since we have to inherit from _DeviceArray rather than jaxlib's `DeviceArray`.

I wish I could come up with a unit test to expose this case, but for now I'll settle for fixing the bug.


jakevdp:
Wow, good catch! I think the reason this hasn't come up is because `DeviceArrayBase` is registered, and all the device array flavors inherit from that.

jakevdp (approved):

PR #9868, opened 2022-03-14
Fix _lax_max_taylor_rule for broadcastable inputs

YouJiacheng:
`lax.max(x, y)` allows x, y have different shape if they are mutually broadcastable. But `lax.select(p, x, y)` does not.


YouJiacheng:
@mattjj I'm sorry to trouble you. Could I have this PR be reviewed?

mattjj:
Thanks for the ping! (It's no trouble; on the contrary it's helpful so I don't let things slip.)

And thanks for the improvement!

mattjj (approved):

Issue #9863, opened 2022-03-13
Make the readthedocs download-able as a PDF


IanQS:
Is there a reason it's not? I'd like to download it to my tablet reader (a low-powered eink device) and be able to reference it from there. The web browsing feature (even if it's local) leaves a lot to be desired on some of those devices

jakevdp:
I'll take a look - I think this is configurable in readthedocs, though it may require some work to get the PDF formatting to look OK

IanQS:
Hey there! I'm just following up on this. It seems like hte build was successful :) ? Would this be merge-ready soon(ish)? 

Thanks for being so quick on this

jakevdp:
The build was successful, but as indicated at the link above, CI builds skip the PDF to avoid timeouts. So to actually test this, I need to merge it, which I strongly suspect will cause a timeout in our readthedocs build (which occasionally times out as it is), at which point I'll most likely have to revert the change.

I just haven't had a block of time yet where I can deal with the potential churn, but I'll try to give it a shot this week.

IanQS:
Absolutely no worries! Worst case scenario I can look into it and see if I can build it on my end

jakevdp:
OK, I gave it a try... good news is that it didn't time out. Bad news is that the PDF build failed because of SVG graphics used in the included jupyter notebooks (raw build log here: https://readthedocs.org/api/v2/build/16428989.txt). The problematic images are the "Open In Colab" badges at the top of each notebook rendered in the documentation.

It looks like the suggested fix is to use `image` directives in sphinx, but for us this is not possible because the source is in ipynb markdown and is automatically converted to rst via myst-nb, and the point of these badges is to be viewable in the ipynb itself.

We might be able to address this by switching these from SVG to PNG. Quite annoying that the only way to check is to merge a PR...

jakevdp:
There are a large number of other errors there as well, including unicode that's unsupported by LaTeX.

I think given the amount of trial-and-error that would be required here, it's probably not worth spending time on unless we get a lot more interest in PDF docs. I'm going to abandon this and revert the initial change.

I'll leave this open in case anyone else wants to attempt going down the road of generating PDF from our sphinx docs.

jakevdp:
Note that if your only goal is to be able to read the docs offline, you can always download the HTML zip version:
![8gBbG4JJiAQaL7p](https://user-images.githubusercontent.com/781659/159369841-624157b4-2b3e-422c-9292-a1822985d5a3.png)

[...]

PR #9869, opened 2022-03-14
Experiment: add PDF export to readthedocs build

jakevdp:
#9863

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,6 +18,7 @@ sphinx:
 # Optionally build your docs in additional formats such as PDF and ePub
 formats:
   - htmlzip
+  - pdf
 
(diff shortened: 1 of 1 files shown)


jakevdp:
Hmm... it turns out that the PDF is not built as part of the readthedocs CI: https://docs.readthedocs.io/en/stable/pull-requests.html#limitations

It looks like our only option to try out the PDF is to merge this PR and let readthedocs build off the main branch. My only worry is in regards to the readthedocs build time – I know we're pushing the timeout limit already, and I'm not sure how much time the PDF adds.

mattjj:
Let's give it a shot and be ready to undo it! If we wanted to extend the time limit maybe there's a way to pay to do that.

mattjj (approved):

Issue #9864, opened 2022-03-14
rocm docker image jax build failure


brettkoonce:
When running the rocm jax build script:

     sudo ./build/rocm/ci_build.sh --keep_image bash -c "./build/rocm/build_rocm.sh"

I get the following error:

```
ERROR: /workspace/jaxlib/BUILD:96:11: no such package '@org_tensorflow//tensorflow/third_party/
rocm/google': BUILD file not found in directory 'tensorflow/third_party/rocm/google' of externa
l repository @org_tensorflow. Add a BUILD file to a directory to mark it as a package. and refe
renced by '//jaxlib:hip_gpu_kernel_helpers'                   
``` 

ping @reza-amd

reza-amd:
Thanks for reporting the issue. Fix PR has been submitted.
https://github.com/google/jax/pull/9870

PR #9870, opened 2022-03-14
[ROCm] Fix jaxlib bazel file for ROCm

reza-amd:
Fix for #9864.

@jakevdp, was there any particular reason behind [this](https://github.com/google/jax/commit/616df55ad47862cd02d1a1131d4df692eb006d05)? For building jax alone, the gating is not necessary because ```hip_gpu_kernel_helpers``` is only used for rocm builds.

In other words, If we remove the gating logic and let the ```hip_gpu_kernel_helper``` remains similar to cuda_gpu_kernel_helpers, what would be the issue?

### DIFF
--- a/jaxlib/BUILD
+++ b/jaxlib/BUILD
@@ -107,7 +107,7 @@ cc_library(
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
     ] + if_rocm_is_configured([
-        "@org_tensorflow//tensorflow/third_party/rocm/google:rocm_headers",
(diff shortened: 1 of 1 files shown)


jakevdp:
I added that because the non-ROCM build failed without it. I did not test on an ROCM build - the pattern is used elsewhere (e.g. in tensorflow's bazel files) so I assumed it would be OK.

jakevdp (approved):

jakevdp:
For various reasons, it's easier to make sure everything is working if I replace this PR with an automatically generated one. See #9871. Thanks!

PR #1538, opened 2019-10-21
Fix NaNs in grad(jax.nn.elu) for large positive inputs.

jheek:
The gradient of np.where is not robust to NaNs in the unselected branch.
A double where trick makes sure that the `expm1` will not NaN.
I have added a check_grads test that NaNs with the original implementation.

PR #9876, opened 2022-03-14
Safe where in jax.nn.celu and nn.leaky_relu

### LINKED ISSUE #9873: celu activation wavefunction sometimes produces nans where it shouldn't
I have a fairly simple full forward neural network written in flax

```
from flax import linen as nn

class MLP(nn.Module):
    features: Tuple[int]

    @nn.compact
    def __call__(self, x):
        for f in self.features:
            x = nn.celu(nn.Dense(f)(x), alpha=2)
        return jnp.squeeze(nn.Dense(1)(x))
```
My problem is fairly simple: I standard scale a dataset of 5 features which one target, and I perform mean square error minimisation. Somehow, I get nans in the loss function, triggered by a call in `jnp.expm1` inside the `celu`.

Now, `nn.celu` calls `jax.nn.celu`. Checking the code I found out that `celu` is implemented like this:

```
def celu(x: Array, alpha: Array = 1.0) -> Array:
    return jnp.where(x > 0, x, alpha * jnp.expm1(x / alpha))
```

Fortunately, I checked the code of `elu`, which is implemented like this:
```
def elu(x: Array, alpha: Array = 1.0) -> Array:
    safe_x = jnp.where(x > 0, 0., x)
    return jnp.where(x > 0, x, alpha * jnp.expm1(safe_x))
```

I modified `celu` adding the `safe_x` line and the nan error completely disappeared.

My question is, why is there a need to have this `safe_x`? Shouldn't the `jnp.where` make `x` safe by default?
This looks like a bug in `jnp.where`, as the third argument should be executed only if `x > 0`. There should not be a need for the `safe_x`.

VolodyaCO:
Closes #9873

### DIFF
--- a/jax/_src/nn/functions.py
+++ b/jax/_src/nn/functions.py
@@ -159,7 +159,8 @@ def leaky_relu(x: Array, negative_slope: Array = 1e-2) -> Array:
     x : input array
     negative_slope : array or scalar specifying the negative slope (default: 0.01)
   """
-  return jnp.where(x >= 0, x, negative_slope * x)
+  safe_x = jnp.where(x > 0, 0., x)
+  return jnp.where(x >= 0, x, negative_slope * safe_x)
 
 @jax.jit
 def hard_tanh(x: Array) -> Array:
@@ -199,7 +200,8 @@ def celu(x: Array, alpha: Array = 1.0) -> Array:
     x : input array
     alpha : array or scalar (default: 1.0)
   """
-  return jnp.where(x > 0, x, alpha * jnp.expm1(x / alpha))
+  safe_x = jnp.where(x > 0, 0., x)
+  return jnp.where(x > 0, x, alpha * jnp.expm1(safe_x / alpha))
 
 @jax.jit
 def selu(x: Array) -> Array:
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks - once you take care of the CLA, I'll take a look!

YouJiacheng:
This problem is caused by that `exp' gradient is `inf`(not `nan`!!!) for large input.
Interestingly, only reverse-mode autodiff of `lax.select`(backend of `np.where`) will return `nan` when unselected branch gradient is `inf`
See https://github.com/google/jax/pull/1538 (but gradient actually is `inf` instead of `nan` in the unselected branch)
I think `leaky_relu` does not need such fix.

VolodyaCO:
Sorry @jakevdp. I'm waiting for my company to create a google group!

YouJiacheng:
Hi! I think we should not modify `leaky_relu`.
And are you ok for CLA now?

VolodyaCO:
@YouJiacheng So on Apr 4th we submitted the CLA, and we are waiting to receive the CLA signed from google, apparently. I just asked if the CLA step is ready.

VolodyaCO:
@jakevdp @YouJiacheng The CLA was finally sorted out! ☠️ sorry that it took so long!

VolodyaCO:
I see that the issue has been solved in another PR, do I close this one?

jakevdp:
Yes, it looks like the issue was addressed in a different way in https://github.com/google/jax/commit/b0805a8a318696cc39eb53a05dc10f2e9cee6a29.

Sorry about all the CLA troubles - the good news is now you're all set for your next contribution! 😀

PR #9891, opened 2022-03-15
[shape_poly] Improve error message and docs for shape polynomials used as JAX values

gnecula:


jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1074,6 +1074,27 @@ def test_errors(self):
...
+  def test_error_jax_value(self):
+
+    x = np.ones([3, 5], dtype=np.float32)
+
+    with self.assertRaisesRegex(
+        core.InconclusiveDimensionOperation,
+        "Dimension polynomial 'b' used in a context that requires a constant"):
+      self.CheckShapePolymorphism(
+          # Cannot use dimension variable as a JAX value
+          lambda x: jnp.array(x.shape[0]),
```
rxsang, line 1086: How about `jnp.shape(x)[0]`? Is it the same behavior as `x.shape[0]` in JAX, or different?

jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1074,6 +1074,27 @@ def test_errors(self):
...
+          input_signature=[tf.TensorSpec([None, 5], dtype=x.dtype)],
+          polymorphic_shapes=[("b, _")],
+          expected_output_signature=tf.TensorSpec([]))
+
+    self.CheckShapePolymorphism(
+        # We can convert a dimension variable to a JAX value
+        lambda x: jnp.array(core.dimension_as_value(x.shape[0])),
```
rxsang, line 1093: Why calling `core.dimension_as_value` could work? Is this a public API we expect JAX users to use?

rxsang (approved):
LGTM. Left 2 comments.

jax/experimental/jax2tf/tests/shape_poly_test.py
gnecula: `jnp.shape` is synonim for np.shape : https://github.com/google/jax/blob/3a949acccb47f6b6d22c71896569d047c0590a88/jax/_src/numpy/lax_numpy.py#L115

so `x.shape` and `jnp.shape` would produce the same result. In general, for computations that should happen at tracing time we use the regular `np`.

jax/experimental/jax2tf/tests/shape_poly_test.py
gnecula: `core.dimension_as_value` is right now an undocumented API, and it is used internally. We may remove the need for it. 

What it does is wrap a dimension as a JAX value that can then be used in JAX computations. The classic need for it is when implementing `jnp.mean` as `jnp.sum(x, axis=0) / x.shape[0]`. You need to wrap `x.shape[0]` with `core.dimension_as_value` or else you will get an error when the first dimension is polymorphic, because you cannot divide a JAX value by a dimension polynomial.

PR #9906, opened 2022-03-16
add random.loggamma and improve dirichlet & beta implementation

### LINKED ISSUE #9896: Distribution of jax.random.dirichlet samples is incorrect for small values of the concentration parameters
`jax.random.dirichlet` seems to produce incorrect distributions for small values of the concentration parameters (i.e. where most of the probability mass should be on the vertices of the simplex).

I've only tested this for 3D input (because this can easily be visualised in 2D), so I don't know if it happens for other dimensionalities.

I'm producing samples as follows:
```
samples = jax.random.dirichlet(key, 0.01 * np.array([1.0, 1.0, 1.0]), (100000,))
```

On CPU and TPU, there seems to be an additional vertex added to the simplex (i.e. it's actually 4D, projected into 2D), which lives at the barycenter of the three existing vertices.

On GPU, the problem is even worse, with additional extra vertices at the midpoints between each pair of vertices (so 4 spurious vertices in total).

Visualisations and more info at http://go/jax-dirichlet-simplex-issue

jakevdp:
Fixes #9896

This roughly follows the strategy of TFP, which computes the underlying gamma distribution in normal space and the boost offset in log space (see https://github.com/tensorflow/probability/blob/dc5b6b094cfc1f18f949e120c3491acb2e2af6eb/tensorflow_probability/python/distributions/gamma.py#L453-L459).

Tested by patching-in initial tests from #9919 to ensure that non-corner-case gamma, beta, and dirichlet outputs remain unchanged by this reimplementation.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -749,6 +749,7 @@ def beta(key: KeyArray,
     shape = core.canonicalize_shape(shape)
   return _beta(key, a, b, shape, dtype)
 
+
 def _beta(key, a, b, shape, dtype):
   if shape is None:
     shape = lax.broadcast_shapes(np.shape(a), np.shape(b))
@@ -760,9 +761,13 @@ def _beta(key, a, b, shape, dtype):
   key_a, key_b = _split(key)
   a = jnp.broadcast_to(a, shape)
   b = jnp.broadcast_to(b, shape)
-  gamma_a = gamma(key_a, a, shape, dtype)
-  gamma_b = gamma(key_b, b, shape, dtype)
-  return gamma_a / (gamma_a + gamma_b)
+  log_gamma_a = loggamma(key_a, a, shape, dtype)
+  log_gamma_b = loggamma(key_b, b, shape, dtype)
+  # Compute gamma_a / (gamma_a + gamma_b) without losing precision.
+  log_max = lax.max(log_gamma_a, log_gamma_b)
+  gamma_a_scaled = jnp.exp(log_gamma_a - log_max)
+  gamma_b_scaled = jnp.exp(log_gamma_b - log_max)
+  return gamma_a_scaled / (gamma_a_scaled + gamma_b_scaled)
 
 
 def cauchy(key: KeyArray,
@@ -840,8 +845,19 @@ def _dirichlet(key, alpha, shape, dtype):
     _check_shape("dirichlet", shape, np.shape(alpha)[:-1])
 
   alpha = lax.convert_element_type(alpha, dtype)
-  gamma_samples = gamma(key, alpha, shape + np.shape(alpha)[-1:], dtype)
-  return gamma_samples / jnp.sum(gamma_samples, axis=-1, keepdims=True)
+
+  # Compute gamma in log space, otherwise small alpha can lead to poor behavior.
+  log_gamma_samples = loggamma(key, alpha, shape + np.shape(alpha)[-1:], dtype)
+  return _softmax(log_gamma_samples, -1)
+
+
+def _softmax(x, axis):
+  """Utility to compute the softmax of x along a given axis."""
+  if not dtypes.issubdtype(x.dtype, np.floating):
+    raise TypeError(f"_softmax only accepts floating dtypes, got {x.dtype}")
+  x_max = jnp.max(x, axis, keepdims=True)
+  unnormalized = jnp.exp(x - lax.stop_gradient(x_max))
+  return unnormalized / unnormalized.sum(axis, keepdims=True)
 
 
 def exponential(key: KeyArray,
@@ -875,7 +891,7 @@ def _exponential(key, shape, dtype):
   return lax.neg(lax.log1p(lax.neg(u)))
 
 
-def _gamma_one(key: KeyArray, alpha):
+def _gamma_one(key: KeyArray, alpha, log_space):
   # Ref: A simple method for generating gamma variables, George Marsaglia and Wai Wan Tsang
   # The algorithm can also be founded in:
   # https://en.wikipedia.org/wiki/Gamma_distribution#Generating_gamma-distributed_random_variables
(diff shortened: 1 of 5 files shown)


jax/_src/random.py
```diff
@@ -991,6 +1045,7 @@ def gamma(key: KeyArray,
       produces a result shape equal to ``a.shape``.
     dtype: optional, a float dtype for the returned values (default float64 if
       jax_enable_x64 is true, otherwise float32).
+    log_spasce: optional, if True, return the samples in log space.
```
fehiepsi, line 1048: typo `log_spasce` :)
I think it is better to have [loggamma](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.loggamma.html) sampler rather than adding an argument here.

jax/_src/random.py
```diff
@@ -926,21 +961,39 @@ def _next_kxv(kxv):
...
 
   # initial state is chosen such that _cond_fn will return True
+  key, subkey = _split(key)
+  u_boost = uniform(subkey, (), dtype=dtype)
   _, _, V, _ = lax.while_loop(_cond_fn, _body_fn, (key, zero, one, _lax_const(alpha, 2)))
-  z = lax.mul(lax.mul(d, V), boost)
-  return lax.select(lax.eq(z, zero), jnp.finfo(z.dtype).tiny, z)
+  if log_space:
+    # TODO(jakevdp): there are negative infinities here due to issues mentioned above. How should
+    # we handle those?
```
fehiepsi, line 969: how about setting lower bound (through minval) for u_boost? `0` value is not expected for the algorithm I believe. It is even better than using `1 - u` (which has lower bound eps, rather than tiny)

jax/_src/random.py
```diff
@@ -840,8 +840,36 @@ def _dirichlet(key, alpha, shape, dtype):
     _check_shape("dirichlet", shape, np.shape(alpha)[:-1])
 
   alpha = lax.convert_element_type(alpha, dtype)
-  gamma_samples = gamma(key, alpha, shape + np.shape(alpha)[-1:], dtype)
-  return gamma_samples / jnp.sum(gamma_samples, axis=-1, keepdims=True)
+
+  # Compute gamma in log space, otherwise small alpha can lead to poor behavior.
+  log_gamma_samples = gamma(key, alpha, shape + np.shape(alpha)[-1:], dtype, log_space=True)
+  axes = (log_gamma_samples.ndim - 1,)
+  return lax.exp(log_gamma_samples - _logsumexp(log_gamma_samples, axes=axes, keepdims=True))
```
fehiepsi, line 847: Should we use `jax.nn.softmax` for this?

jakevdp: Probably not, because we want the log-softmax, and computing the softmax first and then taking the log would lead re-introduce the same kinds of precision issues we're seeking to avoid.

jax/_src/random.py
jakevdp: That's a good idea... what we should really do is base this on `exponential()`, and make those kinds of decisions there. But we can't because then the log samples would not be numerically equivalent to the non-log samples, and we can't change the non-log samples without breaking backward compatibilty for the sampler. So I think until we're comfortable breaking users, the algorithm I have here is the only option.

jax/_src/random.py
fehiepsi: Hmm, I thought loggamma generates log(gamma samples). Probable the name is confusing. For lognormal, we generate samples from exp(normal samples), but that pattern does not apply for loggamma. Probably statisticians should rename LogNormal to ExpNormal for a better interpretation. I searched the question and found this [stackexchange](https://math.stackexchange.com/questions/4268139/why-is-it-called-a-lognormal-distribution-instead-of-an-expnormal-distribution) thread.

jakevdp: It looks like (according to scipy's definition) log-gamma gives just what we need: https://github.com/scipy/scipy/blob/0ed66c76c010ec897a116e5a331941e43e3da562/scipy/stats/_continuous_distns.py#L4917

jakevdp: Still I'm not sure it would make sense to create a new `random_loggamma_p` primitive since the implementation would share so much with the existing `gamma_p` primitive. What if we add the keyword to the primitive, but expose it to the user via a `jax.random.loggamma()` which has the property `jnp.allclose(jnp.exp(loggamma(*args)), gamma(*args))`?

fehiepsi: > add the keyword to the gamma primitive, but expose it to the user via a jax.random.loggamma()

Agreed! I think we don't need to expose that argument to the user api. I just suggested removing `log_space` from this `gamma` function, not the primitive.

jakevdp: Done

jax/_src/random.py
jakevdp: Done.

jax/_src/random.py
```diff
@@ -760,9 +761,20 @@ def _beta(key, a, b, shape, dtype):
   key_a, key_b = _split(key)
   a = jnp.broadcast_to(a, shape)
   b = jnp.broadcast_to(b, shape)
-  gamma_a = gamma(key_a, a, shape, dtype)
-  gamma_b = gamma(key_b, b, shape, dtype)
-  return gamma_a / (gamma_a + gamma_b)
+  log_gamma_a = loggamma(key_a, a, shape, dtype)
+  log_gamma_b = loggamma(key_b, b, shape, dtype)
+  return lax.exp(log_gamma_a - _logaddexp(log_gamma_a, log_gamma_b))
```
jakevdp, line 766: I think a softmax-like approach may be more efficient here.

fehiepsi (commented):
Thanks Jake! The overall implementation looks reasonable to me. I'll play a bit with edge cases tomorrow to see how robust log-space implementation can provide.

jax/_src/random.py
```diff
@@ -926,21 +956,40 @@ def _next_kxv(kxv):
...
+  del prng_impl  # unused
   samples = jnp.reshape(sample, -1)
   alphas = jnp.reshape(a, -1)
+  if log_space:
+    samples = lax.exp(samples)
+    def gamma_grad(alpha, sample):
+      zero = lax_internal._const(sample, 0)
+      inf = lax_internal._const(sample, np.inf)
+      # TODO(jakevdp) add comment motivating grad->inf for sample->0
+      return lax.select(lax.eq(sample, zero), inf, lax.random_gamma_grad(alpha, sample) / sample)
```
jakevdp, line 983: I'm still not 100% sure what to return in the case that `sample` is zero.

jakevdp:
I did a bit of cleanup and improved the grad implementation to avoid infinities. PTAL!

jax/_src/random.py
```diff
@@ -926,21 +949,42 @@ def _next_kxv(kxv):
...
+  key, subkey = _split(key)
+  u_boost = uniform(subkey, (), dtype=dtype)
   _, _, V, _ = lax.while_loop(_cond_fn, _body_fn, (key, zero, one, _lax_const(alpha, 2)))
-  z = lax.mul(lax.mul(d, V), boost)
-  return lax.select(lax.eq(z, zero), jnp.finfo(z.dtype).tiny, z)
+  if log_space:
+    # TODO(jakevdp): there are negative infinities here due to issues mentioned above. How should
+    # we handle those?
+    log_boost = lax.select(boost_mask, zero, lax.mul(lax.log(u_boost), lax.div(one, alpha_orig)))
+    return lax.add(lax.add(lax.log(d), lax.log(V)), log_boost)
```
fehiepsi, line 959: I think it is a bit inconvenient to return loggamma samples that after taking exponent, we get 0. Maybe to be comparable with `gamma` sampler, we can do
```
z = lax.add(lax.add(lax.log(d), lax.log(V)), log_boost)
log_tiny = lax.log(jnp.finfo(z.dtype).tiny)
return lax.select(lax.lt(z, log_tiny), log_tiny, z)
```
Beside convenient for users, this way we might avoid -inf samples, avoid special treatment for small values in loggamma grad, avoid nan in grad test, and make gamma sampler = exp(loggamma sampler) even for small values.

fehiepsi (commented):
Thanks Jake! Looking like loggamma is robust for dirichlet/beta samplers. I just have a comment regarding small values of loggamma.

jax/_src/random.py
jakevdp: I'm going to push back here:  we don't want log-gamma to be comparable to gamma: we want it to be able to represent values that gamma cannot. The primary motivation for adding the log-gamma sampler is to be able to represent values that exponentiate to zero. If we truncate the output at `log(tiny)`, we will lose the ability to robustly sample `beta` and `dirichlet` values for small alpha.

fehiepsi: That's a good point! Thanks for the explanation.

fehiepsi (approved):
LGTM! I can confirm that using loggamma for dirichlet is robust for very small `alpha`
```
alpha = 1e-30 * jnp.arange(1, 8)
random.dirichlet(random.PRNGKey(0), alpha, (n,)).mean(0), alpha / alpha.sum()  # should be close
```
It is better than using `gamma(alpha + 1) / expon(concentration)` because with loggamma, we use exact same `alpha` to draw gamma samples without requiring the exponential term.

PR #9919, opened 2022-03-16
random: add values test cases for various distributions

jakevdp:
Why? In addressing https://github.com/google/jax/issues/9896, I plan to modify the implementation of `gamma`, `beta`, and `dirichlet`. For backward compatibility, we want to ensure that the sequence of values generated by these routines does not change in such refactorings; this PR adds a framework whereby we can test this for these and other distributions.

### DIFF
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -15,6 +15,8 @@
 
 from functools import partial
 from unittest import SkipTest, skipIf
+from typing import Any, Tuple, NamedTuple, Optional
+import zlib
 
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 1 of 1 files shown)


froystig (changes requested):
Let's test RNG implementations other than the default as well. The easiest way might be to do so under assumption that #9263 is complete. Here's an [example](https://github.com/google/jax/blob/4d966bb9ac9ae5aa058cebfc0e59dfa0f8bc64b4/tests/random_test.py#L701-L709).

jakevdp:
OK, updated to test all distributions on multiple PRNG backends. PTAL!

froystig (approved):
Looks good! Maybe add to the docstring when/how one might update these when making changes, or what policy to follow (in particular, do we need to version and deprecate?).

jakevdp:
> Maybe add to the docstring when/how one might update these when making changes, or what policy to follow

Done

PR #9923, opened 2022-03-16
revise internal helper _is_perturbed to _maybe_perturbed, fix logic

### LINKED ISSUE #9567: [Incorrect gradients] `_is_perturbed` produces the wrong value when used inside `lax.scan` etc.
This:
```python
import jax
import jax.lax as lax
from jax._src.custom_derivatives import _is_perturbed

def f(x):
    def g(y, _):
        z = y * x
        print(_is_perturbed(z))
        return y, None
    g(1, None)
    return lax.scan(g, 1, xs=None, length=1)[0]

jax.jvp(f, (1.0,), (1.0,))
```
produces:
```
True
False
```

The correct behaviour is for both to be `True`.

I imagine this is silently setting a few gradients to zero, somewhere out there.

mattjj:
@patrick-kidger FYI (and feel free to leave a review!)

fixes #9567

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -25,6 +25,7 @@ from jax.tree_util import (tree_flatten, tree_unflatten, tree_map,
                         tree_multimap, treedef_is_leaf, treedef_tuple,
                         register_pytree_node_class)
 from jax._src import custom_api_util
+from jax._src import dtypes
 from jax._src.util import cache, safe_zip, safe_map, split_list, Unhashable
 from jax._src.api_util import flatten_fun_nokwargs, argnums_partial
 from jax.core import raise_to_shaped
@@ -943,13 +944,25 @@ def closure_convert(fun, *example_args):
   else:
     return _closure_convert_for_avals(fun, in_tree, in_avals)
 
-def _is_perturbed(x: Any) -> bool:
-  if isinstance(x, ad.JVPTracer):
-    return True
-  elif isinstance(x, core.Tracer):
-    return any(_is_perturbed(attr) for name, attr in x._contents())
-  else:
+def _maybe_perturbed(x: Any) -> bool:
+  # False if x can't represent an AD-perturbed value (i.e. a value
+  # with a nontrivial tangent attached), up to heuristics, and True otherwise.
+  # See https://github.com/google/jax/issues/6415 for motivation.
+  x = core.full_lower(x)
+  if not isinstance(x, core.Tracer):
+    # If x is not a Tracer, it can't be perturbed.
     return False
+  elif isinstance(x, pe.DynamicJaxprTracer):
+    # If x is a DynamicJaxprTracer then we're staging out; differentiation could
+    # happen later, but some types always have trivial tangents.
+    vspace = x.aval.at_least_vspace()
+    return not (vspace is core.abstract_unit or vspace is core.abstract_token or
+                vspace is dtypes.float0)
+  elif not isinstance(x, ad.JVPTracer):
+    # If x is not a JVPTracer, recursively check its contents.
+    return any(_maybe_perturbed(attr) for name, attr in x._contents())
+  else:
+    return True  # We can't be sure!
 
 @cache()
 def _closure_convert_for_avals(fun, in_tree, in_avals):
@@ -957,7 +970,7 @@ def _closure_convert_for_avals(fun, in_tree, in_avals):
   jaxpr, out_pvals, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, in_avals)
   out_tree = out_tree()
 
-  (closure_consts, hoisted_consts), merge = partition_list(_is_perturbed, consts)
+  (closure_consts, hoisted_consts), merge = partition_list(_maybe_perturbed, consts)
   num_consts = len(hoisted_consts)
 
   def converted_fun(*args_hconsts):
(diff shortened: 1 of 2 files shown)


jax/_src/custom_derivatives.py
```diff
@@ -943,21 +944,33 @@ def rev(objective_fn, res, g):
...
+  # False if x can't represent an AD-perturbed value (i.e. a value
+  # with a nontrivial tangent attached), up to heuristics, and True otherwise.
+  # See https://github.com/google/jax/issues/6415 for motivation.
+  x = core.full_lower(x)
+  if not isinstance(x, core.Tracer):
+    # If x is not a Tracer, it can't be perturbed.
     return False
+  elif isinstance(x, pe.DynamicJaxprTracer):
+    # If x is a DynamicJaxprTracer then we're staging out; differentiation could
+    # happen later, but some types always have trivial tangents.
```
froystig, line 957: solid comment.

froystig (approved):
nice!

patrick-kidger:
As a fix for correctness this LGTM!

Since JAX sees a whole program then in principle it should be able to do this precisely without heuristics though. Correctness is good but correctness and efficiency is better.

I think I can see how this should be doable for all practical use cases. The fix is to adjust `custom_jvp` and `custom_vjp` and deprecate `closure_convert`. AFAIK it's only ever needed around these anyway.

We clearly can't have a non-heuristic `closure_convert` that works in arbitrary contexts, as we need to defer binary classification until autodiff time and I don't think there's any way to encode things appropriately in the jaxprs before then. But since we only ever (a) use it around custom autodiff, and (b) only ever use the hoisted constants in one way (as args back into the provided function), then we can handle everything at once just by pushing the whole lot into the custom autodiff operations.

Have a look at Equinox's `filter_custom_vjp`. This currently uses similar floating-point heuristics as it's constrained by its use of `jac.custom_vjp`, but it demonstrates an API in which nondifferentiability/differentiablility is a detail that can be hidden from the end user.

(I recognise this is a lot of work, and invites similar Equinox-filtering-style changes elsewhere, and that something like this is undoubtedly low priority.)

mattjj:
Thanks for taking a look @patrick-kidger !

> Since JAX sees a whole program then in principle it should be able to do this precisely without heuristics though.

> We clearly can't have a non-heuristic closure_convert that works in arbitrary contexts, as we need to defer binary classification until autodiff time and I don't think there's any way to encode things appropriately in the jaxprs before then.

I may not be following but aren't these two statements contradictions? ("should be able to do this precisely without heuristics" and "can't have a non-heuristic closure_convert") What's "this" mean in the first sentence?

Actually it's not quite true that JAX sees the whole program. Someone could always use `jax.make_jaxpr` to stage out a function, at which point the `_maybe_perturbed` function would be called, and later when interpreting that jaxpr differentiation may or may not be involved. For that reason we simply can't know for sure if differentiation is going to eventually be applied. (If we rule out the use of `jax.make_jaxpr` and jaxpr interpreters, and only think about e.g. higher-order primitives like `jax.lax.scan`, then indeed I believe we could have a complete and sound check.)

> The fix is to adjust custom_jvp and custom_vjp and deprecate closure_convert.

> Have a look at Equinox's filter_custom_vjp.

Could you say more about the fix you have in mind, and/or provide some code links? I can't tell what you mean :)

I looked at the docstring for `filter_custom_vjp` but since it also seems to work on heuristics (as you say) I don't see what the diff is.

patrick-kidger:
> I may not be following but aren't these two statements contradictions? ("should be able to do this precisely without heuristics" and "can't have a non-heuristic closure_convert") What's "this" mean in the first sentence?

Right, sorry for being unclear. "should be able to do this precisely without heuristics" -> 10,000 foot view that JAX should be able to figure out which arrays participate in autodiff, as it sees the whole program. [Yes it does! See my next point.]

"can't have a non-heuristic closure_convert" is then digging into the implementation details of how that might happen; at this point I'm making the comment that `closure_convert` is a bad abstraction because what is or isn't participating in autodiff isn't known at the time that it is parsed. As you say:

> Someone could always use jax.make_jaxpr to stage out a function, at which point the _maybe_perturbed function would be called, and later when interpreting that jaxpr differentiation may or may not be involved.

which is precisely my point as well. `_maybe_perturbed` is being called at the wrong time! So I agree with you ;)

> Could you say more about the fix you have in mind, and/or provide some code links? I can't tell what you mean :)
> 
> I looked at the docstring for filter_custom_vjp but since it also seems to work on heuristics (as you say) I don't see what the diff is.

So the problem is that `_maybe_perturbed` is being called at the wrong time. It's being called at first-trace-time; it should be called at autodiff time. Given the ubiquity of `closure_convert` around custom autodiff, I'm proposing that filtering differentiable from nondifferentiable be something that's handled inside `jax.custom_jvp` and `jax.custom_vjp`, rather than having `closure_convert` as a standalone function.

[Side note! `closure_convert` only filters differentiable from nondifferentiable for _implicit_ args -- but around custom autodiff we need to the same thing for _explicit_ args as well. Once again `closure_convert` isn't quite the right abstraction; using just the facilities of JAX we'd have to define a new function closing over the explicit args and then `closure_convert` _that_ in order to filter all arguments to the function.]

The implementation of `filter_custom_vjp` indeed has an implementation that depends on heuristics. I'm offering it just as an example of an API that doesn't have to depend on heuristics. Imagine `filter_custom_vjp` was actually a primitive that got preserved in the jaxpr across transformations (like `custom_vjp` already is). We could then use it to perform filtering at autodiff time rather than first-trace-time.

By the way, notice how the word "filter" keeps coming up in the above discussion. Filtering is just a really useful thing to be doing around transformations, and so much so that Equinox is built around this idea. (Along with the parameterised-functions-as-pytrees idea.)

PR #9926, opened 2022-03-16
elide trivial slices, cf. #9215

mattjj:


jax/_src/lax/slicing.py
```diff
@@ -55,6 +56,10 @@ def slice(operand: Array, start_indices: Sequence[int],
   <https://www.tensorflow.org/xla/operation_semantics#slice>`_
   operator.
   """
+  if (start_indices == (0,) * len(start_indices) and
+      limit_indices == tuple(np.shape(operand)) and
+      isinstance(operand, (device_array.DeviceArray, core.Tracer))):
```
jakevdp, line 61: `and strides is None`?

mattjj: Ah good catch! Thank you.

PR #9950, opened 2022-03-18
introduce a protocol for compilation wrappers

### LINKED ISSUE #9808: protocol/types for compilation wrappers
Let's set up a protocol for the return type of `jit`, `pmap`, `pjit`, and `xmap`. These essentially have `__call__` and `lower`.

This would:
* improve documentation;
* enable pytype annotations; and
* allow for some easier checks in internals that expect such a thing as input.

cc @hawkinsp

froystig:
Fixes #9808, as far as type annotations are concerned.

cc #7733

### DIFF
--- a/jax/_src/stages.py
+++ b/jax/_src/stages.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from typing import Any, Optional, Tuple, Union
+from typing_extensions import Protocol
 
 from jax import core
 from jax.interpreters import pxla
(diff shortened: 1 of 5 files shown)


jax/_src/stages.py
```diff
@@ -179,9 +110,94 @@ def __call__(self, *args, **kwargs):
...
+      return self._lowering.hlo()
+    else:
+      raise ValueError(f"Unknown dialect {dialect}")
+
+  # TODO(frostig): remove this in favor of `compiler_ir`
+  def _xla_computation(self):
+    return self._lowering.hlo()
+
+
+class Wrapped(Protocol):
```
mattjj, line 190: Maybe something more descriptive and less generic, like `WrappedForStaging` or `StagedCallable`?

mattjj (approved):

YouJiacheng:
Hi!
I think maybe there is something wrong with Protocol for `jit`.
`Protocol` method allows type variance, i.e. covariant return type & contravariant arguments type(unfortunately `Callable` allows these too.)
But compilation wrappers should be invariant, i.e. only allow exact match.

PR #9953, opened 2022-03-18
fix jnp.convolve int precision

mattjj:
We were promoting arguments to inexact dtypes (i.e. promoting ints to floats), possibly because this code was written before integer convolves were available in HLO, but that led to precision problems!


zhangqiaorjc (approved):

mattjj:
Uh oh:
```
E   RuntimeError: UNIMPLEMENTED:
E   Can't lower one or more integer convolutions to idioms supported by CuDNN.
E
E   CuDNN integer convolutions must have:
E
E     - s8 input and filter,
E     - f32 bias (if present),
E     - s8 or f32 output, and
E     - s8 side_input (if present) if output is s8.
```

Looks like on GPU we need to have an alternative lowering for these integer convolutions...

PR #3988, opened 2020-08-07
Add nbytes property to jax.numpy arrays.

jakevdp:
Fixes #3985

PR #9955, opened 2022-03-18
Add `itemsize` property to JAX arrays

### LINKED ISSUE #9593: JAX arrays should expose itemsize property (bug?)
Similar to issue #3985, in which JAX arrays were made to mimic NumPy's `numpy.ndarray.nbytes` property, they should also mimic the `numpy.ndarray.itemsize` property.  Note that JAX **does** already expose `itemsize` in its dtypes.   This unexpected lack of parity with NumPy might imply this is more like a bug than an enhancement.

```python
import numpy as onp
import jax.numpy as jnp

a = onp.ones(0, dtype='ushort')
ja = jnp.ones_like(a)

print(f'{a.itemsize = }')
print(f'{a.dtype.itemsize = }')

if hasattr(ja, 'itemsize'):
    print(f'{ja.itemsize = }')
else:
    print("jax array does not define itemsize.")
    
print(f'{ja.dtype.itemsize = }')
```
currently outputs

```
a.itemsize = 2
a.dtype.itemsize = 2
jax array does not define itemsize.
ja.dtype.itemsize = 2
```

Thanks in advance!

nicholasjng:
Fixes #9593.

This commit adds the `itemsize` property to the JAX Array and ShapedArray classes. Additionally, tests were added to check that the behavior exactly matches that of NumPy's `itemsize` property.

This change was directly modelled off of pull request #3988, which added the (related) `nbytes` property to JAX arrays.

Addendum: I added the last line to reflect how similar of a change this is to #3988, because it _really is that similar_. I am happy to give direct credit for this as well by updating the commit message (e.g. by adding a co-author) - just let me know.

cc @jakevdp , who authored #3988.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -5159,6 +5159,10 @@ def _nbytes(arr):
   return size(arr) * _dtype(arr).itemsize
 
 
+def _itemsize(arr):
+    return _dtype(arr).itemsize
+
+
 def _clip(number, min=None, max=None, out=None, *, a_min=None, a_max=None):
   # ndarray.clip has a slightly different API from clip (min -> a_min, max -> a_max)
   # TODO: remove after deprecation window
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks great, thanks!

PR #9966, opened 2022-03-19
Add weakreaf_lru_cache to prevent caches from pinning jaxprs.

pschuh:
To use this cache, the first argument must be some type that is
object identity hashed (like a jaxpr).

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -16,10 +16,13 @@
 import functools
 from functools import partial
 import itertools as it
+from collections import namedtuple
 import operator
 import types
-from typing import (Any, Callable, Iterable, List, Tuple, Generic, TypeVar, Set,
-                    Iterator, Sequence)
+import threading
+from typing import (Any, Callable, Dict, Iterable, List, Tuple, Generic,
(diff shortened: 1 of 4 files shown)


jax/_src/util.py
```diff
@@ -216,6 +219,55 @@ def wrapper(*args, **kwargs):
 
 memoize = cache(max_size=None)
 
+_CacheInfo = namedtuple("_CacheInfo", ["hits", "misses", "maxsize", "currsize"])
```
mattjj, line 222: Maybe remove the leading underscore? Mainly because the caller of `.cache_info()` will get one of these objects, so this object isn't module-internal.

mattjj (approved):
Wow, really nice code.

To fix cache leaks in control flow functions, not just `pjit`, I think we need to apply `weakref_lru_cache` instead of `cache` in a few other places, like near the top of control_flow.py. Want to do that in this PR, or leave for follow-up?

YouJiacheng:
"To use this cache, the first argument must be some type that is object identity hashed (like a jaxpr)."
Do you think it would be better if add an assertion `hash(weak_arg) == hash(id(weak_arg))`?

PR #9968, opened 2022-03-19
Rename jax.nn.normalize to standardize. Add normalize alias with DeprecationWarning

### LINKED ISSUE #9899: Rename jax.nn.normalize to jax.nn.standardize
As discussed: the jax.nn.normalize function actually standardizes by default, so we should name it accordingly.

Nobody reading `jax.nn.normalize(x)` expects the result not to have norm-1, and [PyTorch agrees with this too](https://pytorch.org/docs/stable/generated/torch.nn.functional.normalize.html).

The proposal here is just to rename to avoid user bugs. This would also open up the name `jax.nn.normalize` for a PyTorch-like implementation, but that should be as a follow-up, if desired.

dogeplusplus:
Fixes #9899 

Changes:
- `jax.nn.normalize` is renamed to `jax.nn.standardize`
- Add `normalize` as an alias with a `DeprecationWarning`
- Rename unit test to the new name along with references

Feel free to comment, change, or let me know if the deprecation should be in a separate PR.

### DIFF
--- a/jax/_src/nn/functions.py
+++ b/jax/_src/nn/functions.py
@@ -16,6 +16,7 @@
 
 from functools import partial
 import operator
+import warnings
 import numpy as np
 from typing import Any, Optional, Tuple, Union
 
@@ -333,7 +334,7 @@ def softmax(x: Array,
   return unnormalized / jnp.sum(unnormalized, axis, where=where, keepdims=True)
 
 @partial(jax.jit, static_argnames=("axis",))
-def normalize(x: Array,
+def standardize(x: Array,
               axis: Optional[Union[int, Tuple[int, ...]]] = -1,
               mean: Optional[Array] = None,
               variance: Optional[Array] = None,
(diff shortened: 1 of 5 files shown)


jax/_src/nn/functions.py
```diff
@@ -351,6 +352,16 @@ def normalize(x: Array,
         jnp.square(x), axis, keepdims=True, where=where) - jnp.square(mean)
   return (x - mean) * lax.rsqrt(variance + epsilon)
 
+@partial(jax.jit, static_argnames=("axis",))
```
froystig, line 355: We'll want to remove this `jit` decoration. It isn't needed since we call right into the jitted `standardize`, and it could interfere with issuing a warning (a side effect).

froystig (changes requested):
Thanks for the PR!

Following the [API compatibility](https://jax.readthedocs.io/en/latest/api_compatibility.html) policy, could you announce the deprecation in [the changelog](https://github.com/google/jax/blob/main/CHANGELOG.md)? The jaxlib 0.3.2 changelog entry offers an example. Including a link to #9899 would be helpful.

froystig (approved):
Looks good. Thanks again!

PR #9986, opened 2022-03-21
Fix for hipsparse in ROCm.

reza-amd:
This PR fixes sparse ops for ROCm.
@jakevdp

### DIFF
--- a/jax/_src/lib/__init__.py
+++ b/jax/_src/lib/__init__.py
@@ -126,6 +126,8 @@ try:
 except ImportError:
   hipsparse = None
 
+sparse_apis = cusparse or hipsparse or None
+
 try:
   import jaxlib.cuda_prng as cuda_prng  # pytype: disable=import-error
 except ImportError:
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -37,10 +37,9 @@ from jax._src.lib import lapack
 
 from jax._src.lib import cuda_linalg
 from jax._src.lib import cusolver
-from jax._src.lib import cusparse
 from jax._src.lib import hip_linalg
 from jax._src.lib import hipsolver
-from jax._src.lib import hipsparse
+from jax._src.lib import sparse_apis
 
 from jax._src.lib import xla_client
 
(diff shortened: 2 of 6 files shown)


jakevdp (commented):
Thanks, looks good.

Since this kind of logic appears in multiple files, I wonder if it wouldn't make sense to abstract it a bit? Could we make a `jax._src.lib.sparse_apis` module that would export the correct `coo_matvec`, `coo_matmat`, etc. to avoid having to put the cusparse/hipsparse switch logic in every file that uses them?

reza-amd:
This is a good idea. Let me work on it . I will update the PR with changes.

reza-amd:
@jakevdp could you please take a look at the recent changes. I will squash the commits after you agreed with the changes.

jax/_src/lib/__init__.py
```diff
@@ -117,14 +117,12 @@ def _parse_version(v: str) -> Tuple[int, ...]:
...
-  cusparse = None
-
-try:
-  import jaxlib.hipsparse as hipsparse  # pytype: disable=import-error
-except ImportError:
-  hipsparse = None
+  try:
+    import jaxlib.hipsparse as sparse_apis  # pytype: disable=import-error
+  except ImportError:
+    sparse_apis = None
```
jakevdp, line 125: Suggestion: it might be cleaner to do something like this, which has the side-effect of still exposing `cusparse` and `hipsparse` for anyone who might be using them directly:
```python
try:
  import jaxlib.cusparse as cusparse  # pytype: disable=import-error
except ImportError:
  cusparse = None

try:
  import jaxlib.hipsparse as hipsparse  # pytype: disable=import-error
except ImportError:
  hipsparse = None

sparse_apis = cusparse or hipsparse or None
```

jakevdp (changes requested):
Looks great! Only one minor comment

jakevdp (approved):
So much deleted code 😍

jakevdp:
We should fix the mypy errors here (`assert sparse_apis is not None` should do the trick I think)

Test failures are pre-existing on the main branch; I'll look into getting those fixed (tracked in https://github.com/google/jax/issues/9991)

reza-amd:
@hawkinsp do you agree to make similar changes to other hip/cuda modules (hipsolver/cusolver, cublas/hipblas, etc. )?

reza-amd:
> So much deleted code 😍

@jakevdp I can use my git skills to remove all jax repo if you like.:satisfied:
https://youtu.be/2uYeb_c8_bk

reza-amd:
Hi @jakevdp, does this PR need other changes from my side?
I think CI jobs were run before #9991 fix. If you restart the jobs, they all should be green.

reza-amd:
@rsanthanam-amd

jakevdp:
Hi, sorry I lost track of this. Can you rebase your commit on the main branch to see if it fixes the CI tests? Thanks

reza-amd:
Thanks Jake. Rebased the PR and all are green now.

jakevdp:
Thanks for working on this! If you're up for it, I think a similar treatment for hipsolver/cusolver and cublas/hipblas would be useful.

PR #9999, opened 2022-03-23
generic protocol for compilation wrappers

### LINKED ISSUE #9932: Add IDE friendly type annotation for jit
see https://github.com/google/jax/pull/9917

YouJiacheng:
Fix #9932 without breaking #9950

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -33,6 +33,7 @@ import weakref
 import types
 from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
                     Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable)
+from typing_extensions import ParamSpec
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


YouJiacheng:
@froystig WDYT?

YouJiacheng:
Emmm, it seems that type checking tool can't recognize……
error: Argument 1 to "partial" has incompatible type "Callable[[Callable[P, T], DefaultNamedArg(Union[int, Iterable[int], None], 'static_argnums'), DefaultNamedArg(Union[str, Iterable[str], None], 'static_argnames'), DefaultNamedArg(Optional[Device], 'device'), DefaultNamedArg(Optional[str], 'backend'), DefaultNamedArg(Union[int, Iterable[int]], 'donate_argnums'), DefaultNamedArg(bool, 'inline')], Wrapped[P, T]]"; expected "Callable[..., Wrapped[P, T]]"  [arg-type]

YouJiacheng:
Ah, I find that it is a `mypy` bug: it doesn't handle `partial` correctly, while `pylance` can.
https://github.com/python/mypy/issues/12593

PR #10017, opened 2022-03-23
Add upper version bound to readme pip install instructions

konstin:
Our notebooks (for [colabfold](https://github.com/sokrypton/ColabFold)) just [broke](https://github.com/sokrypton/ColabFold/issues/185) due to the release of jax 0.3.0. This pull requests adds upper version bounds to the readme pip install instructions to avoid downstream breakage through new jax releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ development on a laptop, you can run
 
 ```bash
 pip install --upgrade pip
-pip install --upgrade "jax[cpu]"
+pip install --upgrade "jax[cpu]<0.4.0"
 ```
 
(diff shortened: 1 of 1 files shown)


mattjj:
Thanks for reaching out! And sorry for the breakage. (Was it due to using JAX internal symbols/APIs?)

But I don't understand the reasoning behind this PR: if you want to pin your project to tested JAX versions, why would that mean changing the JAX install instructions? Perhaps instead the change should be in the ColabFold README.md's install instructions? That is, maybe the CoalbFold README should say something like `pip install --upgrade "jax[cuda]==<version> ..."`?

PR #10019, opened 2022-03-23
Add a documentation for the JAX transfer guard

hyeontaek:


docs/transfer_guard.rst
```diff
@@ -0,0 +1,71 @@
...
+  option.
+* A ``with jax.transfer_guard(GUARD_LEVEL): ...`` context manager will set the
+  thread-local option within the scope of the context manager.
+
+Note that similar to other JAX configuration options, a newly spawned thread
+will use the global option instead of any active thread-local option of the
+scope where the thread was spawned.
+
+The transfer guards also allow fine-control the transfer guard level of
+individual transfer directions. The flag and context manager name is suffixed
```
froystig, line 36: For the first sentence here, how about the following instead?

"The transfer guards can also be applied more selectively, based on the direction of transfer."

froystig (approved):
Thanks! This looks great.

docs/transfer_guard.rst
hyeontaek: Thanks, updated.

docs/transfer_guard.rst
```diff
@@ -0,0 +1,70 @@
...
+
+* Explicit transfers: ``jax.device_put*()`` and ``jax.device_get()`` calls.
+* Implicit transfers: Other transfers (e.g., printing a ``DeviceArray``).
+
+A transfer guard can take an action based on its guard level:
+
+* ``"allow"``: Silently allow all transfers (default).
+* ``"log"``: Log and allow implicit transfers. Silently allow explicit
+  transfers.
+* ``"disallow"``: Disallow implicit transfers. Silently allow explicit
```
soraros, line 18: IMO we should also make clear what does disallow means here, i.e. raise an error, and maybe also mention the error type.

hyeontaek: Thanks for the suggestion! I have updated the text.

PR #10033, opened 2022-03-25
Checkify: add way to embed runtime info in error messages.

LenaMartens:
Error messages in checkify are static (for technical reasons they need to be!), but we want to be able to embed a small amount of dynamic information in them to make them more useful. A clear use-case is in out-of-bounds errors: you can now report the value of the out-of-bounds index (which is only known at run-time) in the error message. This PR adds a `payload` field to `Error` which can contain a dynamic array.

After this change, the out-of-bounds error message looks like
```
out-of-bounds indexing at checkify_test.py:157 (<lambda>): index 100 is out of bounds for array of shape (2, 3, 7).'
```
For now, only the OOB error adds a payload, and the payload needs to always be an int32 scalar array.

Things to follow-up on:
- Relax the fixed-size constraint. This would require us to find a fix-point in the scan/while-loop rule and do some padding in the cond rule. (We could even make the amount of payloads variadic)
- Add the axis at which the index is out-of-bounds (ie. the payload becomes a (2,) int32 array)


mattjj:
The latter follow-up sounds super valuable!

jax/experimental/checkify/checkify_impl.py
```diff
@@ -82,19 +89,24 @@ def throw(self):
...
+                     lambda msgs, data: Error(data[0], data[1],  # type: ignore
+                                              dict(msgs), data[2]))  # type: ignore
 
 init_error = Error(False, 0, {})
 next_code = it.count(1).__next__  # globally unique ids, could be uuid4
 
 
-def assert_func(error: Error, pred: Bool, msg: str) -> Error:
+def assert_func(error: Error, pred: Bool, msg: str,
+                payload: Payload = np.ones((), np.int32)) -> Error:
```
mattjj, line 104:
```suggestion
                payload: Payload) -> Error:
```

Do we need a default arg here? Defaults are dangerous and this is an explicit-ish function anyway, so maybe we can drop it.

mattjj (approved):

jax/experimental/checkify/checkify_impl.py
mattjj: Nevermind, Lena explained to me why it's better to make this optional.

jax/experimental/checkify/checkify_impl.py
```diff
@@ -56,24 +56,31 @@ def setnewattr(obj, name, val):
 
 Bool = Union[bool, core.Tracer]
 Int = Union[int, core.Tracer]
+# For now, the payload needs to be a fixed-size array (int32 scalar).
+# TODO(lenamartens): Relax this fixed-size constraint.
+Payload = Union[np.ndarray, core.Tracer]
```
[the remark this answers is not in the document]
LenaMartens: If I define an `init_payload` at the top-level, it can't be a `jnp` array right?

PR #10036, opened 2022-03-25
use weakref_lru_cache from #9966 in more places

mattjj:
TODO: add tests

cc @pschuh


mattjj:
@pschuh pointed out that the WeakRefSequence stuff is broken:

> I think all of the weakrefsequence versions are wrong and will simply do no caching.
> The newly constructed obj will fall out of scope immediately and get gc'd

Two possible fixes:
* make `weakref_lru_cache` accept multiple weakref objects
* basically keep a reference to the sequence object somewhere, and create weakref callbacks for each fun so that after any one has fired we drop the reference to the sequence

YouJiacheng:
weakref lru cache will drop temporary object.
I think some control-flows simply don't need a cache for their typical usage, e.g. `lax.while_loop`.

PR #10038, opened 2022-03-26
allow recursive uses of `custom_transpose`

froystig:
[I sent this PR a second time as #10041 in order to work around pulldown issues.]

Previously, a `custom_transpose` function could not call itself from its transpose rule, because staging out the `custom_transpose` primitive would result in endless recursion. Instead of staging out the transpose rule when the enclosing `custom_transpose` rule is staged out, this change replaces it temporarily by a thunk (relieving the recursive staging issue). The thunk allows us to lazily stage out the original transpose rule, a process that needs to happen only when the enclosing `custom_transpose` primitive is bound or transposed.

Along the way: in order to write a test that involves staging without compilation, we relied on `lax.cond`. But to transpose `lax.cond` explicitly (i.e. using `jax.linear_transpose`), we need a way to indicate that the conditional branches are linear in their operands. We already have the means to do this internally; the `cond` primitive has a sequence of linear indicators (the JVP rule for `cond` sets these to a non-trivial pattern). This change simply allows one to control the same linear operand indicators when calling the `cond` traceable directly.

Part of #9129.

### DIFF
--- a/jax/_src/custom_transpose.py
+++ b/jax/_src/custom_transpose.py
@@ -19,6 +19,7 @@ from jax import core
 from jax import linear_util as lu
 from jax.interpreters import ad
 from jax.interpreters import mlir
+from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
 from jax.tree_util import (tree_flatten, tree_leaves, tree_map,
                            tree_structure, treedef_tuple, tree_unflatten)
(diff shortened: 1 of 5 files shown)


mattjj (approved):

froystig:
This PR wound up in a messy failed pulldown state. I remade it as #10041, which is now merged.

PR #10039, opened 2022-03-27
Add scipy.linalg.rsf2csf

ajcr:
This adds [`scipy.linalg.rsf2csf`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.rsf2csf.html#scipy.linalg.rsf2csf).

I followed the SciPy implementation as closely as I could, although operations such as 
```python
T[m-1:m+1, m-1:] = G @ T[m-1:m+1, m-1:]
```
(where `m` is variable) were quite fiddly to recreate in lax primitives (though what I have seems to work). Perhaps there's a more idiomatic Jax way I've missed?

For the tests, arrays of shape `(50, 50)` and larger, I had to set `tol=1e-05` to get passing tests. I wasn't 100% sure about how to mimic the dtype promotion that the SciPy implementation uses or how to maximise precision over larger numbers of loop iterations, so please let me know if I should look at improving the PR in this area too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
   * added array creation routines {func}`jax.numpy.frombuffer`, {func}`jax.numpy.fromfunction`,
     and {func}`jax.numpy.fromstring` ({jax-issue}`#10049`).
   * `DeviceArray.copy()` now returns a `DeviceArray` rather than a `np.ndarray` ({jax-issue}`#10069`)
+  * added {func}`jax.scipy.linalg.rsf2csf`
 * Deprecations:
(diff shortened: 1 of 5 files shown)


hawkinsp (approved):

hawkinsp:
It looks like there are some test failures:

```
______________ ScipyLinalgTest.testRsf2csf_shape=float64[100,100] ______________
[gw27] linux -- Python 3.7.13 /usr/bin/python3.7
tests/linalg_test.py:1407: in testRsf2csf
    args_maker, tol=tol)
E   UnboundLocalError: local variable 'tol' referenced before assignment
```

etc.

ajcr:
@hawkinsp: is there anything further that I can do to help get this PR merged?

I noticed that SciPy handled the `(1, 1)` shape case which my earlier JAX implementation didn't, so I've added the check
```python
  if N == 1:
    return T, Z
```
and an additional test case to cover this.

hawkinsp:
Thanks for the PR!

PR #10043, opened 2022-03-28
[shape_poly] Change internal representation of parametric shapes

### LINKED ISSUE #9975: jax2tf with certain reshapes fails with polymorphism
This feels like it should work:
```
import jax
import jax.numpy as jnp
import tensorflow as tf

@jax.jit
def f(x):
  n,z = x.shape
  padding_mask = jnp.ones((n + 1, z), dtype=bool)
  return jnp.reshape(padding_mask, ((n + 1) * z))

f_tf = tf.function(jax2tf.convert(f,
                                  polymorphic_shapes=["(b, 2)"]),
                                  autograph=False)
f_tf.get_concrete_function(tf.TensorSpec([None, 2], tf.int32))
```
gives
> TypeError: Shapes must be 1D sequences of concrete values of integer type, got 2*b + 2.

gnecula:
The parametric shapes are implemented as polynomials in terms of
dimension variables. In the previous implementation the representation
is as a `dict`. This happens to confuse a test in `jnp.reshape` because
the polynomial appears to be an iterable.

I removed the inheritance from `dict`, so that a dimension polynomial
is harder to confuse with other types.

Fixes: #9975

### DIFF
--- a/jax/experimental/jax2tf/shape_poly.py
+++ b/jax/experimental/jax2tf/shape_poly.py
@@ -36,7 +36,7 @@ import itertools
 import functools
 import operator as op
 import re
-from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union
 
 import jax
 from jax._src.numpy import lax_numpy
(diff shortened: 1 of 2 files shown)


gnecula:
@rxsang Please take a look

rxsang (approved):

PR #10048, opened 2022-03-28
custom VJP by custom JVP and transpose

froystig:
Adds an alternative implementation of `custom_vjp`, behind a flag, that reduces to `custom_jvp` + `custom_transpose`, as sketched out in #9129. Still experimental, not all tests pass yet.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -54,6 +54,14 @@ def int_env(varname: str, default: int) -> int:
   return int(os.getenv(varname, str(default)))
 
 
+UPGRADE_BOOL_HELP = (
+    " This will be enabled by default in future versions of JAX, at which "
+    "point all uses of the flag will be considered deprecated (following "
+    "https://jax.readthedocs.io/en/latest/api_compatibility.html).")
+
+UPGRADE_BOOL_EXTRA_DESC = " (transient)"
+
+
 class Config:
   _HAS_DYNAMIC_ATTRIBUTES = True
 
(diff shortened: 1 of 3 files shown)


jax/_src/config.py
```diff
@@ -54,6 +54,14 @@ def int_env(varname: str, default: int) -> int:
   return int(os.getenv(varname, str(default)))
 
 
+UPGRADE_BOOL_HELP = (
+    " This will be enabled by default in future versions of JAX, at which "
+    "point its reverse setting will be considered deprecated. In a version "
```
mattjj, line 59: Is "reverse setting" clear?

mattjj: I wonder if we can just link to a deprecation policy docs page here.

jax/_src/config.py
```diff
@@ -565,6 +578,12 @@ def update_thread_local_jit_state(**kw):
     help=('Select the default PRNG implementation, used when one is not '
           'explicitly provided at seeding time.'))
 
+enable_custom_vjp_by_jvp = config.define_bool_state(
+    name='jax_enable_custom_vjp_by_jvp',
```
mattjj, line 582:
```suggestion
    name='jax_enable_custom_vjp_by_custom_transpose',
```

mattjj (approved):

jax/_src/config.py
froystig: Good idea. Done.

PR #10049, opened 2022-03-28
Implement jnp.from* array creation functions

### LINKED ISSUE #10045: Support for `numpy.fromiter`
As far as I can tell there is not an existing issue or implementation for `jax.numpy.fromiter`, which currently raises:

```
[NotImplementedError: Numpy function <built-in function fromiter> not yet implemented]()
```

jakevdp:
Fixes #10045

Note that `np.fromregex` is not compatible with JAX because it can only create structured arrays, which JAX does not support.


froystig (changes requested):
See comment in #10045 about introducing only the pure functions among these?

jakevdp:
Thanks Roy - I removed `fromfile` and `fromiter` based on https://github.com/google/jax/issues/10045#issuecomment-1082153407. PTAL!

froystig (approved):
Nice!

PR #10058, opened 2022-03-29
Add istft to jax.scipy.signal.

yotarok:
Following the discussion in #9411, ISTFT is added to signal.py.
(ISTFT wasn't included in the first PR because it doesn't share a common utility function used for STFT.)

### DIFF
--- a/jax/_src/scipy/signal.py
+++ b/jax/_src/scipy/signal.py
@@ -492,3 +492,138 @@ def welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
                    axis=axis, average=average)
 
   return freqs, Pxx.real
+
+
+def _overlap_and_add(x, step_size):
+  """Utility function compatible with tf.signal.overlap_and_add.
+
+  Args:
+    x: An array with `(..., frames, frame_length)`-shape.
+    step_size: An integer denoting overlap offsets. Must be less than
+      `frame_length`.
+
+  Returns:
+    An array with `(..., output_size)`-shape containing overlapped signal.
+  """
+  _check_arraylike("_overlap_and_add", x)
+  step_size = jax.core.concrete_or_error(int, step_size,
+                                        "step_size for overlap_and_add")
+  if x.ndim < 2:
+    raise ValueError('Input must have (..., frames, frame_length) shape.')
+
+  *batch_shape, nframes, segment_len = x.shape
+  flat_batchsize = np.prod(batch_shape, dtype=np.int64)
+  x = x.reshape((flat_batchsize, nframes, segment_len))
+  output_size = step_size * (nframes - 1) + segment_len
+  nstep_per_segment = 1 + (segment_len - 1) // step_size
+
+  # Here, we use shorter notation for axes.
+  # B: batch_size, N: nframes, S: nstep_per_segment,
+  # T: segment_len divided by S
+
+  padded_segment_len = nstep_per_segment * step_size
+  x = jnp.pad(x, ((0, 0), (0, 0), (0, padded_segment_len - segment_len)))
+  x = x.reshape((flat_batchsize, nframes, nstep_per_segment, step_size))
+
+  # For obtaining shifted signals, this routine reinterprets flattened array
+  # with a shrinked axis.  With appropriate truncation/ padding, this operation
+  # pushes the last padded elements of the previous row to the head of the
+  # current row.
+  # See implementation of `overlap_and_add` in Tensorflow for details.
+  x = x.transpose((0, 2, 1, 3))  # x: (B, S, N, T)
+  x = jnp.pad(x, ((0, 0), (0, 0), (0, nframes), (0, 0)))  # x: (B, S, N*2, T)
+  shrinked = x.shape[2] - 1
+  x = x.reshape((flat_batchsize, -1))
+  x = x[:, :(nstep_per_segment * shrinked * step_size)]
+  x = x.reshape((flat_batchsize, nstep_per_segment, shrinked * step_size))
+
+  # Finally, sum shifted segments, and truncate results to the output_size.
+  x = x.sum(axis=1)[:, :output_size]
+  return x.reshape(tuple(batch_shape) + (-1,))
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/signal.py
```diff
@@ -492,3 +492,110 @@ def welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
...
+    shifted = jnp.pad(shifted, ((0, 0), (leftpad, rightpad)))[:, :out_length]
+
+    y += shifted
+  return y.reshape(tuple(batch_shape) + (-1,))
+
+
+@_wraps(osp_signal.istft)
+def istft(Zxx, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
+          input_onesided=True, boundary=True, time_axis=-1, freq_axis=-2):
+  # Input validation
```
jakevdp, line 522: Call `_check_arraylike` on `Zxx` to start with

jax/_src/scipy/signal.py
```diff
@@ -492,3 +492,110 @@ def welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
...
+  if Zxx.ndim < 2:
+    raise ValueError('Input stft must be at least 2d!')
+  freq_axis = canonicalize_axis(freq_axis, Zxx.ndim)
+  time_axis = canonicalize_axis(time_axis, Zxx.ndim)
+  if freq_axis == time_axis:
+    raise ValueError('Must specify differing time and frequency axes!')
+
+  Zxx = jnp.asarray(Zxx, dtype=jax.dtypes.canonicalize_dtype(
+      np.result_type(Zxx, np.complex64)))
```
jakevdp, line 531: Probably not for this PR, but this kind of thing is used frequently enough that we should probably define a `_promote_dtypes_complex` helper similar to our current `_promote_dtypes_inexact`.

jax/_src/scipy/signal.py
```diff
@@ -492,3 +492,110 @@ def welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
...
+    Zxx = jnp.transpose(Zxx, outer_idxs + (freq_axis, time_axis))
+
+  # Perform IFFT
+  ifunc = jax.numpy.fft.irfft if input_onesided else jax.numpy.fft.ifft
+  # xsubs: [..., T, N], N is the number of frames, T is the frame length.
+  xsubs = ifunc(Zxx, axis=-2, n=nfft)[..., :nperseg, :]
+
+  # Get window as array
+  if isinstance(window, (str, tuple)):
+    win = osp_signal.get_window(window, nperseg)
```
jakevdp, line 570: I think we should call `win = jnp.asarray(win)` to avoid doing the operations below at compile time.

jax/_src/scipy/signal.py
```diff
@@ -492,3 +492,110 @@ def welch(x, fs=1.0, window='hann', nperseg=None, noverlap=None, nfft=None,
...
+  x /= jnp.where(norm > 1e-10, norm, 1.0)
+
+  # Put axes back
+  if x.ndim > 1:
+    if time_axis != Zxx.ndim - 1:
+      if freq_axis < time_axis:
+        time_axis -= 1
+      x = jnp.moveaxis(x, -1, time_axis)
+
+  time = jnp.arange(x.shape[0]) / float(fs)
```
jakevdp, line 600: This cast to `float(fs)` will prevent `fs` from being traced by JIT, vmap, etc. - I don't think we need it?

tests/scipy_signal_test.py
```diff
@@ -376,6 +383,59 @@ def testWelchWithDefaultStepArgsAgainstNumpy(
...
+            x,
+            fs=fs, window=window, nperseg=nperseg, noverlap=noverlap,
+            nfft=nfft, input_onesided=onesided, boundary=boundary,
+            time_axis=timeaxis, freq_axis=freqaxis)
+      return result
+
+    jsp_fun = partial(jsp_signal.istft,
+        fs=fs, window=window, nperseg=nperseg, noverlap=noverlap, nfft=nfft,
+        input_onesided=onesided, boundary=boundary,
+        time_axis=timeaxis, freq_axis=freqaxis)
```
jakevdp, line 425: You might consider making `fs` one of the arguments to `osp_fun`/`jsp_fun` so that we test dynamic inputs for that value. If you remove the `float` cast in the implementation, I believe it will work.

jakevdp (changes requested):
Looks great! A few minor requests. One last comment, to ensure this is part of the documentation we should add it to the list in https://github.com/google/jax/blob/main/docs/jax.scipy.rst

jax/_src/scipy/signal.py
yotarok: Done

jax/_src/scipy/signal.py
yotarok: Done

jax/_src/scipy/signal.py
yotarok: No, we don't need it. Deleted.

tests/scipy_signal_test.py
yotarok: Done.

yotarok:
Thanks for the feedbacks.
I updated the branch reflecting your comments.
In addition to that, I modified the method for implmenting overlap-add following the suggestion given by Peter (https://github.com/google/jax/issues/9411#issuecomment-1082066504)

jax/_src/scipy/signal.py
yotarok: I agree, such utility function should be useful.

jakevdp (approved):

yotarok:
Fixed.
I first tried to fix this by simulating "osp" more accurately.  But, I found that the behavior is different between the version used in CI (probably 1.2.3?) and the newer version (1.8.0).
So, I gave up and disabled dtype checks.

jakevdp:
Sounds good - thanks!

Issue #2632, opened 2020-04-07
deepcopy of jax.numpy array generates regular numpy array


durian888:
While updating some existing code that used copy.deepcopy I got what was a rather unexpected result namely that deepcopy returns a regular numpy array when I passed  in a jax.numpy array.  Is this expected behaviour? If so perhaps a note in the docs to indicated this would be good. 

The following is a snippet that replicates the problem.

for `jax.version.__version__ == '0.1.62''`

import jax.numpy as np

import copy

array_in = np.array([1,2,3], dtype= np.int32)
array_out = copy.deepcopy(array_in)

print("type in:" + str(type(array_in))+ "type out:" +str(type(array_out)))

give 
type in:<class 'jax.interpreters.xla.DeviceArray'>type out:<class 'numpy.ndarray'>

mattjj:
Thanks for pointing this out!

AIUI `copy.deepcopy` works by pickling then unpickling an object. We [make JAX DeviceArrays pickle just like NumPy ndarrays](https://github.com/google/jax/blob/52e779c3c18250079d4ee022c36311fdc9531e9c/jax/interpreters/xla.py#L877-L878), and so this deepcopy behavior seems to be a consequence.

While this pickling behavior was the original intent, it might be worth revisiting! @shoyer any thoughts?

(For posterity: [here's](https://stackoverflow.com/a/26599346) how we could restore a DeviceArray.)

shoyer:
If I understand this correctly, that implementation of reduce means that DeviceArray objects unpickle to NumPy arrays, too?

I don't think we want either of these behaviors. Copy, deepcopy and pickle/unpickle should all restore an object with the same type as what you started with.

jaivardhankapoor:
[Sorry for the underspecified question, I am currently writing this on my mobile. Please let me know if you need more info on this query.]

@mattjj  are there any plans to follow-up  on this issue? I am tying to implement a sequential Monte Carlo algorithm with particle resampling, in which I use `deepcopy` in resampling the particle objects. Should I reconvert to `DeviceArray` after resampling the particles or wait for the proper restoration behaviours to be implemented in `JAX`?

tawe141:
I'm developing something that runs into this bug and requires gradient information to be preserved after `deepcopy`s. At the moment, I could probably re-cast everything back to `DeviceArray`s, but it would be great if I didn't have to!

shoyer:
There are at least two surprising behaviors here that should be fixed:

[...]

PR #10069, opened 2022-03-29
make device_array.copy() return a device array

jakevdp:
Addresses part of #2632

In https://github.com/google/jax/pull/9732 we made `jnp.copy` return a device array rather than a numpy array. This gives the same treatment to the `copy()` method of device arrays.

I think this is the behavior that we want – note that it will not chgne the behavior of `copy.copy()` or `copy.deepcopy()`, which current return numpy arrays due to their use of the `reduce()` method (similar to  `pickle`). We could address those separately via the `__copy__` and `__deepcopy__` methods if we wish.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2769,11 +2769,11 @@ def _device_get(x):
   if isinstance(x, core.Tracer):
     return x
   try:
-    copy = x.copy
+    toarray = x.__array__
   except AttributeError:
     return x
   else:
-    return copy()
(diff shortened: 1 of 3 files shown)


mattjj (approved):
This sounds right to me. It's reversing an ancient API decision, so it might break some very old (and hopefully easily fixable) code, but I think `jax.device_get` or `np.array(x)` is the right way to spell "give me a NumPy array".

PR #10097, opened 2022-03-30
Replace `reshape` with `expand_dims` if possible

lgeiger:
The [docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.reshape.html) recommend using `expand_dims` over `reshape` to preserve information about axis identity.

This PR switches to `expand_dims` in a couple of places where this is possible which also slightly simplifies the code.

### DIFF
--- a/jax/_src/lax/parallel.py
+++ b/jax/_src/lax/parallel.py
@@ -1487,7 +1487,7 @@ def _pgather_impl(src, idx, *, axes):
   non_axes_shape = src_axes_front.shape[len(axes):]
   src_one_axis_front = src_axes_front.reshape((-1,) + non_axes_shape)
   slice_sizes = (1,) + non_axes_shape
-  idx = lax.reshape(idx, idx.shape + (1,))
+  idx = lax.expand_dims(idx, (-1,))
   offset_dims = tuple(range(idx.ndim - 1, idx.ndim + src_one_axis_front.ndim - 2))
   dnums = slicing.GatherDimensionNumbers(
       offset_dims=offset_dims,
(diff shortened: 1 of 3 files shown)


jax/_src/lax/parallel.py
```diff
@@ -1487,7 +1487,7 @@ def _pgather_impl(src, idx, *, axes):
   non_axes_shape = src_axes_front.shape[len(axes):]
   src_one_axis_front = src_axes_front.reshape((-1,) + non_axes_shape)
   slice_sizes = (1,) + non_axes_shape
-  idx = lax.reshape(idx, idx.shape + (1,))
+  idx = lax_numpy.expand_dims(idx, -1)
```
jakevdp, line 1490: Let's use `lax.expand_dims` here instead, since this is a submodule of `lax`.

jax/interpreters/batching.py
```diff
@@ -621,7 +621,7 @@ def _handle_scalar_broadcasting(nd, x, d):
   if d is not_mapped or nd == np.ndim(x):
     return x
   else:
-    return x.reshape(x.shape + (1,) * (nd - np.ndim(x)))
+    return jax.numpy.expand_dims(x, tuple(range(np.ndim(x), nd)))
```
jakevdp, line 624: Similar here, let's prefer `lax.expand_dims`. It would be better to avoid dependence on `jax.numpy` within this file.

jakevdp (changes requested):
Thanks! Just a quick couple changes

jax/_src/lax/parallel.py
lgeiger: Cool, I didn't know that `lax.expand_dims` also supports negative `axis` arguments. I changed it and updated the PR.

jakevdp (approved):

PR #10102, opened 2022-03-31
[host_callback] Fix tests to ensure we use the correct platform

gnecula:
In host_callback_test, there are a few tests that inspect compiled HLO.
In some cases, we're explicitly creating a CPU XLA computation, but we're handing
it off the to the default backend. When we're on a TPU machine, we're asking a
TPU backend to compile a CPU XLA computation.

Fixes internal b/227521177.


sharadmv (approved):
Looks good!

PR #10103, opened 2022-03-31
Checkify: add axis and axis size to OOB error message.

LenaMartens:


jax/experimental/checkify/checkify_impl.py
```diff
@@ -490,12 +500,18 @@ def gather_error_check(error, enabled_errors, operand, start_indices, *,
...
-  msg = f'out-of-bounds indexing at {summary()}: '
-  msg += 'index {payload} is out of bounds for '
-  msg += f'array of shape {operand.shape}.'
-  start_indices, in_bounds = jnp.ravel(start_indices), jnp.ravel(in_bounds)
-  # Report first index which is out-of-bounds (in row-major order).
-  payload = start_indices[jnp.argsort(in_bounds, axis=0)[0]]
+  # Get first OOB index, axis and axis size so it can be added to the error msg.
+  flat_idx = jnp.argmin(in_bounds)
+  multi_idx = jnp.unravel_index(flat_idx, start_indices.shape)
+  oob_axis = jnp.array(dnums.start_index_map)[multi_idx[-1]]
```
LenaMartens, line 506: > `multi_idx[-1]`
I thought this needed to be `multi_idx[0]`, but from my tests this is actually needs to be the last index. I don't understand why so highlighting to understand if I made a mistake.

`multi_idx` is the multi-dimension index into the indexing array which points to the first OOB index in that array: I figured the _first_ element of this index corresponds to the axis or the array we're indexing into which is OOB, what do you think?

mattjj (approved):

PR #10110, opened 2022-03-31
Fix race condition for weakref destructor by catching rare exceptions.

### LINKED ISSUE #10107: "RuntimeError: dictionary changed size during iteration" inside util / weakref_lru_cache / wrapped
I'm getting `RuntimeError: dictionary changed size during iteration`, which looks like a bug in JAX. I don't have a minimally reproducible example, but can semi-reliable hit this in my training code after a few hours. Here is a stack trace:

```

     581
 --> 582     result = jax.grad(loss_fn)(differentiable_inputs)
     583

 /launcher_root/google3/third_party/py/jax/_src/traceback_util.py in reraise_with_filtered_traceback(*args, **kwargs)
     161     try:
 --> 162       return fun(*args, **kwargs)
     163     except Exception as e:

 /launcher_root/google3/third_party/py/jax/_src/api.py in grad_f(*args, **kwargs)
     875   def grad_f(*args, **kwargs):
 --> 876     _, g = value_and_grad_f(*args, **kwargs)
     877     return g

 /launcher_root/google3/third_party/py/jax/_src/traceback_util.py in reraise_with_filtered_traceback(*args, **kwargs)
     161     try:
 --> 162       return fun(*args, **kwargs)
     163     except Exception as e:

 /launcher_root/google3/third_party/py/jax/_src/api.py in value_and_grad_f(*args, **kwargs)
     957     tree_map(partial(_check_output_dtype_grad, holomorphic), ans)
 --> 958     g = vjp_py(lax_internal._one(ans))
     959     g = g[0] if isinstance(argnums, int) else g

 /launcher_root/google3/third_party/py/jax/_src/tree_util.py in __call__(self, *args, **kw)
     278   def __call__(self, *args, **kw):
 --> 279     return self.fun(*args, **kw)
     280

 /launcher_root/google3/third_party/py/jax/_src/api.py in _vjp_pullback_wrapper(cotangent_dtypes, cotangent_shapes, io_tree, fun, py_args)
    2317           f"{ct_shape}.")
 -> 2318   ans = fun(*args)
    2319   return tree_unflatten(out_tree, ans)

 /launcher_root/google3/third_party/py/jax/_src/tree_util.py in __call__(self, *args, **kw)
     278   def __call__(self, *args, **kw):
 --> 279     return self.fun(*args, **kw)
     280

 /launcher_root/google3/third_party/py/jax/interpreters/ad.py in unbound_vjp(pvals, jaxpr, consts, *cts)
     125     dummy_args = [UndefinedPrimal(v.aval) for v in jaxpr.invars]
 --> 126     arg_cts = backward_pass(jaxpr, reduce_axes, True, consts, dummy_args, cts)
     127     return map(instantiate_zeros, arg_cts)

 /launcher_root/google3/third_party/py/jax/interpreters/ad.py in backward_pass(jaxpr, reduce_axes, transform_stack, consts, primals_in, cotangents_in)
     235           cts_out = reducing_transposes[eqn.primitive](
 --> 236               reduce_axes, cts_in, *invals, **eqn.params)
     237         else:

 /launcher_root/google3/third_party/py/jax/_src/lax/control_flow.py in _scan_transpose(reduce_axes, cts, reverse, length, num_consts, num_carry, jaxpr, linear, unroll, *args)
    1935       num_carry=num_consts-num_ires+num_carry, linear=tuple(linear_trans),
 -> 1936       unroll=unroll)
    1937   ct_consts, ct_init, ct_xs = split_list(outs, [num_consts - num_ires, num_carry])

 /launcher_root/google3/third_party/py/jax/_src/lax/control_flow.py in scan_bind(*args, **params)
    2097     core.check_jaxpr(params['jaxpr'].jaxpr)
 -> 2098   return core.AxisPrimitive.bind(scan_p, *args, **params)
    2099

 /launcher_root/google3/third_party/py/jax/core.py in bind(self, *args, **params)
    2063                  else axis_main.with_cur_sublevel())
 -> 2064     return self.bind_with_trace(top_trace, args, params)
    2065

 /launcher_root/google3/third_party/py/jax/core.py in bind_with_trace(self, trace, args, params)
     290   def bind_with_trace(self, trace, args, params):
 --> 291     out = trace.process_primitive(self, map(trace.full_raise, args), params)
     292     return map(full_lower, out) if self.multiple_results else full_lower(out)

 /launcher_root/google3/third_party/py/jax/core.py in process_primitive(self, primitive, tracers, params)
     612   def process_primitive(self, primitive, tracers, params):
 --> 613     return primitive.impl(*tracers, **params)
     614

 /launcher_root/google3/third_party/py/jax/_src/dispatch.py in apply_primitive(prim, *args, **params)
      93   compiled_fun = xla_primitive_callable(prim, *unsafe_map(arg_spec, args),
 ---> 94                                         **params)
      95   return compiled_fun(*args)

 /launcher_root/google3/third_party/py/jax/_src/util.py in wrapper(*args, **kwargs)
     212       else:
 --> 213         return cached(config._trace_context(), *args, **kwargs)
     214

 /launcher_root/google3/third_party/py/jax/_src/util.py in cached(_, *args, **kwargs)
     205     def cached(_, *args, **kwargs):
 --> 206       return f(*args, **kwargs)
     207

 /launcher_root/google3/third_party/py/jax/_src/dispatch.py in xla_primitive_callable(prim, *arg_specs, **params)
     112   compiled = _xla_callable_uncached(lu.wrap_init(prim_fun), device, None,
 --> 113                                     prim.name, donated_invars, *arg_specs)
     114   if not prim.multiple_results:

 /launcher_root/google3/third_party/py/jax/_src/dispatch.py in _xla_callable_uncached(fun, device, backend, name, donated_invars, *arg_specs)
     170   return lower_xla_callable(fun, device, backend, name, donated_invars,
 --> 171                             *arg_specs).compile().unsafe_call
     172

 /launcher_root/google3/third_party/py/jax/_src/profiler.py in wrapper(*args, **kwargs)
     205     with TraceAnnotation(name, **decorator_kwargs):
 --> 206       return func(*args, **kwargs)
     207     return wrapper

 /launcher_root/google3/third_party/py/jax/_src/dispatch.py in lower_xla_callable(fun, device, backend, name, donated_invars, *arg_specs)
     260         module_name, closed_jaxpr, backend.platform,
 --> 261         mlir.ReplicaAxisContext(axis_env), name_stack, donated_invars)
     262   else:

 /launcher_root/google3/third_party/py/jax/interpreters/mlir.py in lower_jaxpr_to_module(module_name, jaxpr, platform, axis_context, name_stack, donated_args, replicated_args, arg_shardings, result_shardings)
1094 default_ThreadManager-3/51094 server.cc:1014] Scaffolding server shutting down. Disabling health check updater, waiting for running handlers to finish and disallowing new handlers from starting.
     504         arg_shardings=arg_shardings, result_shardings=result_shardings,
 --> 505         input_output_aliases=input_output_aliases)
     506

 /launcher_root/google3/third_party/py/jax/interpreters/mlir.py in lower_jaxpr_to_fun(ctx, name, jaxpr, public, replace_units_with_dummy, replace_tokens_with_dummy, replicated_args, arg_shardings, result_shardings, input_output_aliases)
     651                              jaxpr.jaxpr, map(ir_constants, jaxpr.consts),
 --> 652                              *args)
     653     outs = []

 /launcher_root/google3/third_party/py/jax/interpreters/mlir.py in jaxpr_subcomp(ctx, jaxpr, consts, *args)
     737       ans = rule(rule_ctx, *map(_unwrap_singleton_ir_values, in_nodes),
 --> 738                  **eqn.params)
     739

 /launcher_root/google3/third_party/py/jax/interpreters/mlir.py in f_lowered(ctx, *args, **params)
     770     with core.extend_axis_env_nd(zip(axis_env.names, axis_env.sizes)):
 --> 771       jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(wrapped_fun, ctx.avals_in)
     772     return jaxpr_subcomp(ctx.module_context, jaxpr, _ir_consts(consts),

 /launcher_root/google3/third_party/py/jax/_src/profiler.py in wrapper(*args, **kwargs)
     205     with TraceAnnotation(name, **decorator_kwargs):
 --> 206       return func(*args, **kwargs)
     207     return wrapper

 /launcher_root/google3/third_party/py/jax/interpreters/partial_eval.py in trace_to_jaxpr_dynamic(fun, in_avals, debug_info, keep_inputs)
    1739     jaxpr, out_avals, consts = trace_to_subjaxpr_dynamic(
 -> 1740       fun, main, in_avals, keep_inputs=keep_inputs)
    1741     del main, fun

 /launcher_root/google3/third_party/py/jax/interpreters/partial_eval.py in trace_to_subjaxpr_dynamic(fun, main, in_avals, keep_inputs)
    1775     in_tracers_ = [t for t, keep in zip(in_tracers, keep_inputs) if keep]
 -> 1776     ans = fun.call_wrapped(*in_tracers_)
    1777     out_tracers = map(trace.full_raise, ans)

 /launcher_root/google3/third_party/py/jax/linear_util.py in call_wrapped(self, *args, **kwargs)
     165     try:
 --> 166       ans = self.f(*args, **dict(self.params, **kwargs))
     167     except:

 /launcher_root/google3/third_party/py/jax/_src/lax/control_flow.py in _scan_impl(reverse, length, num_consts, num_carry, jaxpr, linear, unroll, *args)
    1607         num_carry=num_carry, linear=linear, f_impl=f_impl, x_avals=x_avals,
 -> 1608         y_avals=y_avals)
    1609

 /launcher_root/google3/third_party/py/jax/_src/lax/control_flow.py in _scan_impl_loop(reverse, length, num_consts, num_carry, linear, f_impl, x_avals, y_avals, *args)
    1565     init_val = [lax._const(length, 0)] + init + ys_init
 -> 1566     _, *outs = while_loop(cond_fun, body_fun, init_val)
    1567     return outs

 /launcher_root/google3/third_party/py/jax/_src/traceback_util.py in reraise_with_filtered_traceback(*args, **kwargs)
     161     try:
 --> 162       return fun(*args, **kwargs)
     163     except Exception as e:

 /launcher_root/google3/third_party/py/jax/_src/lax/control_flow.py in while_loop(cond_fun, body_fun, init_val)
     325                       cond_nconsts=len(cond_consts), cond_jaxpr=cond_jaxpr,
 --> 326                       body_nconsts=len(body_consts), body_jaxpr=body_jaxpr)
     327   return tree_unflatten(body_tree, outs)

 /launcher_root/google3/third_party/py/jax/core.py in bind(self, *args, **params)
    2059     top_trace = find_top_trace(args)
 -> 2060     axis_main = max((axis_frame(a).main_trace for a in used_axis_names(self, params)),
    2061                     default=None, key=lambda t: getattr(t, 'level', -1))

 /launcher_root/google3/third_party/py/jax/core.py in used_axis_names(primitive, params)
    1969   subst = NameGatheringSubst()
 -> 1970   subst_axis_names(primitive, params, subst)
    1971   return subst.axis_names

 /launcher_root/google3/third_party/py/jax/core.py in subst_axis_names(primitive, params, subst, traverse)
    1988   for name, jaxpr in jaxpr_params:
 -> 1989     new_params[name] = subst_axis_names_jaxpr(jaxpr, shadowed_subst)
    1990   return new_params

 /launcher_root/google3/third_party/py/jax/core.py in subst_axis_names_jaxpr(jaxpr, subst)
    2045   if isinstance(subst, NameGatheringSubst):  # This is a common case, so we optimize it!
 -> 2046     subst.axis_names |= used_axis_names_jaxpr(jaxpr)
    2047     return jaxpr

 /launcher_root/google3/third_party/py/jax/_src/util.py in wrapped(weak_arg, *args, **kwargs)
     253       while len(cache) > maxsize:
 --> 254         del cache[next(iter(cache))]
     255       return result

 UnfilteredStackTrace: RuntimeError: dictionary changed size during iteration
```

pschuh:
Note that the tests don't directly reproduce the error because it is a rare bug and it requires injecting GC calls in the right spot, but they provide reasonable coverage of the weakref part of the code. It wasn't even easy to trigger with multiple threads.

Fixes: https://github.com/google/jax/issues/10107

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -222,12 +222,28 @@ memoize = cache(max_size=None)
 CacheInfo = namedtuple("CacheInfo", ["hits", "misses", "maxsize", "currsize"])
 
 def weakref_lru_cache(call: Callable, maxsize=2048):
+  """
+  Least recently used cache decorator with weakref support.
+
+  The cache will take a weakref to the first argument of the wrapped function
+  and strong refs to all subsequent operations. In all other respects it should
+  behave similar to `functools.lru_cache`.
+  """
   cache: Dict[Any, Any] = {}
   hits = misses = 0
   lock = threading.Lock()
 
   def remove_key(tctx, args, kwargs, weak_arg):
-    del cache[(weak_arg, tctx, args, kwargs)]
+    k = (weak_arg, tctx, args, kwargs)
+    try:
+      # This has a chance to race with the iteration in next(iter(cache)),
+      # but we cannot lock because GC can get triggered synchronously inside
+      # a critical section and will not relinquish control until the callback
+      # has finished. This would lead to a deadlock between this weakref
+      # cleanup function and any function below which locks.
+      del cache[k]
+    except KeyError:
+      pass
 
   def wrapped(weak_arg, *args, **kwargs):
     nonlocal hits, misses
@@ -250,8 +266,22 @@ def weakref_lru_cache(call: Callable, maxsize=2048):
     result = call(weak_arg, *args, **kwargs)
     with lock:
       cache[k] = result
+      num_errors = 0
       while len(cache) > maxsize:
-        del cache[next(iter(cache))]
+        try:
+          del_k = next(iter(cache))
+          # This happens if a weakref callback happens between iter and
+          # next. Just ignore the error. WeakKeyDictionary handles this
+          # by deferring the deletes, but that has a chance at leaking,
+          # and this solution is easier.
+        except RuntimeError:
+          num_errors += 1
+          if num_errors > len(cache):
+            # This must be some other problem.
+            raise
+          else:
+            continue
+        del cache[del_k]
       return result
 
   def cache_info():
--- a/tests/util_test.py
+++ b/tests/util_test.py
@@ -18,6 +18,7 @@ from jax import linear_util as lu
 from jax._src import test_util as jtu
 
 from jax.config import config
+from jax._src.util import weakref_lru_cache
 config.parse_flags_with_absl()
 FLAGS = config.FLAGS
 
@@ -62,6 +63,21 @@ class UtilTest(jtu.JaxTestCase):
     self.assertEqual((2, 4), scaled_positional)
     self.assertEqual(dict(three=6, four=8), scaled_kwargs)
     self.assertEqual(2, out_thunk())
+
+  def test_weakref_lru_cache(self):
+    @weakref_lru_cache
+    def example_cached_fn(key):
+      return object()
+
+    class Key:
+      def __init__(self):
+        # Make a GC loop.
+        self.ref_loop = [self]
+
+    stable_keys = [Key() for _ in range(2049)]
+    for i in range(10000):
+      example_cached_fn(stable_keys[i % len(stable_keys)])
+      example_cached_fn(Key())
 
 
 if __name__ == "__main__":


jax/_src/util.py
```diff
@@ -250,8 +258,19 @@ def wrapped(weak_arg, *args, **kwargs):
     result = call(weak_arg, *args, **kwargs)
     with lock:
       cache[k] = result
+      num_errors = 0
       while len(cache) > maxsize:
-        del cache[next(iter(cache))]
+        try:
+          del cache[next(iter(cache))]
```
hawkinsp, line 264: Scope this exception guard to just the next(iter(...)), not the del ?

jax/_src/util.py
```diff
@@ -227,7 +227,15 @@ def weakref_lru_cache(call: Callable, maxsize=2048):
   lock = threading.Lock()
```
hawkinsp, line 227: Can you add a short class docstring? What exactly is weak in this cache? (Presumably the key, but... say so.)

hawkinsp, line 227: Can you add a class docstring?

jax/_src/util.py
```diff
@@ -227,7 +227,15 @@ def weakref_lru_cache(call: Callable, maxsize=2048):
...
 
   def remove_key(tctx, args, kwargs, weak_arg):
-    del cache[(weak_arg, tctx, args, kwargs)]
+    k = (weak_arg, tctx, args, kwargs)
+    try:
+      # This has a chance to race with
```
hawkinsp, line 232: Nit: Reflow onto fewer lines.

hawkinsp, line 232: Nit: Reflow onto fewer lines.

jax/_src/util.py
```diff
@@ -227,7 +227,15 @@ def weakref_lru_cache(call: Callable, maxsize=2048):
...
+      # other protected code next(iter(cache)),
+      # but we cannot lock because GC can
+      # get triggered synchronously inside a critical section.
```
hawkinsp, line 235: Explain the problem with locking in more detail?

hawkinsp, line 235: Explain the problem with locking in more detail? What goes wrong?

jax/_src/util.py
```diff
@@ -222,12 +222,28 @@ def wrapper(*args, **kwargs):
...
   cache: Dict[Any, Any] = {}
   hits = misses = 0
   lock = threading.Lock()
 
   def remove_key(tctx, args, kwargs, weak_arg):
-    del cache[(weak_arg, tctx, args, kwargs)]
+    k = (weak_arg, tctx, args, kwargs)
+    try:
+      # This has a chance to race with the iteration in next(iter(cache)),
+      # but we cannot lock because GC can get triggered synchronously inside
```
hawkinsp, line 240: You might be able to avoid gc if you use something like .pop() to extract the element inside the critical section but do not delete it until after? Just an idea, not mandatory.

tests/util_test.py
```diff
@@ -63,6 +64,21 @@ def kw_to_positional(factor, *args, **kwargs):
     self.assertEqual(dict(three=6, four=8), scaled_kwargs)
     self.assertEqual(2, out_thunk())
 
+  def test_weakref_lru_cache(self):
```
hawkinsp, line 67: Add a threaded test that exhibits the bug?

hawkinsp (approved):

tests/util_test.py
pschuh: I could not get the bug to arise in any reasonable amount of time without carefully delaying specific lines of code. (a sleep(0) between iter and next is necessary to trigger the bug). The bug is also unrelated to threading. This test, however, does deadlock if you put a lock in the weakref callback.

jax/_src/util.py
pschuh: This code runs like an interrupt on a microprocessor whenever the GC runs. This means that we cannot acquire any locks. We either have to make the code 'lock-free' where the atomic unit of operation is a single python op (pop isn't available on dictionaries, but we could add some sort of pop_front c++ function?), move all this code to c++ where the assumptions are different, or have some sort of new lock type which runs these callbacks 'at the end of the next critical section'.

tests/util_test.py
hawkinsp: It's unrelated to threading? Wasn't the original report only for threaded code, and presumably the easy way to make in between those two statements is to trigger GC of the weakref from another thread?

hawkinsp (approved):

tests/util_test.py
pschuh: I tried testing with threads and it didn't help reproduce the problem any more. The problem comes when GC is triggered which does not change based on the number of threads. All the other code is threadsafe where only one python thread can be in the critical section at a time, so more threads do not improve the chances of a problem happening. I suspect that the multiple threads in the user code just helped trigger this code because it was hitting the cache more.

jax/_src/util.py
YouJiacheng: I think callback should acquire `RLock` here, since [Google Styleguide says "Do not rely on the atomicity of built-in types."](https://google.github.io/styleguide/pyguide.html#218-threading)
It seems that `RLock` won't cause deadlock, see https://github.com/python/cpython/pull/5216.
Of course, `RLock` cannot fix the race condition(maybe "interleaving execution of weakref callback and next(iter(cache))" is more precise) here.

YouJiacheng: In addition, it seems that there is another race condition without `RLock`:
1. lock was acquired in Thread 1, `del_k` was successfully computed.
2. `gc.collect` was triggered in Thread 2 and callback running in Thread 2 remove `del_k` from `cache`
3. `del cache[del_k]` in Thread 1 raise `KeyError`.

YouJiacheng: To fix GC race condition, I think temporarily disable gc is better than catch runtime error.
```python
@contextmanager
def gc_disabled():
  import gc
  gc.disable()
  yield
  gc.enable()

with gc_disabled():
  del cache[next(iter(cache))]
```

pschuh: Disabling the gc doesn't prevent another python thread from dropping the final ref. It is also a global value which can race with other threads causing an indeterminate state for the gc. Libraries should not be randomly setting this value. SimpleQueue doesn't use RLock but instead is directly implemented via the python GIL.

This is probably the right fix for the race condition you mentioned (unless we want to move it to c so all the caching logic can be properly GIL protected and atomic):
```
try:
  del cache[next(iter(cache))]
except RuntimeError, KeyError:
  ...
```

YouJiacheng: If another python thread from dropping the final ref, the callback should be invoked in another thread, and RLock would ensure `cache` "owned" by current thread current control flow.
By referencing cpython#5216, I mean reentrant can avoid deadlock when gc. And SQLAlchemy use RLock successfully https://github.com/python/cpython/issues/59181#issuecomment-1093581337.
Alternatively, we could do cleanup in another thread(using threadpool) and acquire nonreentrant lock.

YouJiacheng: disabling gc won't cause intermediate state, at least if there is no gc callback that release gil, since gc is totally under gil. The only concern is that will set gc to enable regardless the old state of gc and we dont have a reliable way to save the old state.

pschuh: Not a big fan of the thread(pool?) idea either. While it is technically correct, it is very heavyweight and makes it so that the weakref callback can happen at some arbitrary time in the future.

The ideal would be LockWithWorkQueue (or a scoped protection against weakref callbacks), which would allow:
```
lock = LockWithWorkQueue()
with lock:
   # ... locks...
   
lock.call_locked(lambda: ...) # either runs immediately if the lock is available, or runs at the end of the current lock scope.
```

Without this, you still have the problem (with rlock) that the gc call can happen at any point. In which case, you might as well do what I've done and catch the potential errors that come up. Again, without moving code to c, it is hard to make anything perfect here.

YouJiacheng:
I investigated how to implement `LockWithWorkQueue` in python, and found it is hard to make it perfect.

```python
from collections import namedtuple
import functools
from queue import Empty, SimpleQueue
import threading
from typing import Any, Callable
import weakref

class LockWithWorkQueue:
  def __init__(self):
    self._block = threading.Lock()
    self._queue = SimpleQueue()
  
  def __enter__(self):
    return self._block.acquire()
  
  def __exit__(self):
    # we assume __enter__ and __exit__ must be paired
    # thus there is no race with __enter__ or other __exit__
    while True:
      try:
        work = self._queue.get_nowait()
      except Empty:
        break
      else:
        work()
    # if control flow is interrupted here in the same thread by call_locked
    # there will be intermediate state that _block is unlocked while _queue is not empty
    # that should be not too bad
    self._block.release()

  def call_locked(self, work: Callable[[], Any]):
    if self._block.acquire(blocking=False):
      # caller should ensure control flow won't be interrupted here in the same thread
      # or possible interrupt-er in the same thread won't acquire self._block
      # otherwise there will be deadlock
      # gc won't be triggered when invoking weakref callbacks
      # https://github.com/python/cpython/pull/2695
      try:
        work()
      finally:
        self._block.release()
    else:
      self._queue.put_nowait(work)

CacheInfo = namedtuple("CacheInfo", ["hits", "misses", "maxsize", "currsize"])
def weakref_lru_cache(call: Callable, maxsize=2048):
  """
  Least recently used cache decorator with weakref support.
  The cache will take a weakref to the first argument of the wrapped function
  and strong refs to all subsequent operations. In all other respects it should
  behave similar to `functools.lru_cache`.
  """
  cache = {}
  hits = misses = 0
  lock = LockWithWorkQueue()

  def remove_key(args, kwargs, weak_arg):
    k = (weak_arg, args, kwargs)
    lock.call_locked(lambda: cache.pop(k, None))
    

  def wrapped(weak_arg, *args, **kwargs):
    nonlocal hits, misses
    kwargs_key = tuple(kwargs.items())
    k = (weakref.ref(weak_arg,
        functools.partial(remove_key, args, kwargs_key)),
        args, kwargs_key)
    with lock:
      if k in cache:
        hits += 1
        result = cache[k]
        # del and reinsert to bump key in the insertion order.
        del cache[k]
        cache[k] = result
        return result
      misses += 1
    result = call(weak_arg, *args, **kwargs)
    with lock:
      cache[k] = result
      while len(cache) > maxsize:
        del cache[next(iter(cache))]
    return result

  def cache_info():
    with lock:
      return CacheInfo(hits, misses, maxsize, len(cache))

  def cache_clear():
    nonlocal hits, misses
    with lock:
      hits = misses = 0
      cache.clear()

  wrapped.cache_info = cache_info
  wrapped.cache_clear = cache_clear
  return wrapped
```

YouJiacheng:
After further investigating, it seems that moving `next(iter(cache))` code to C (i.e. call `tp_iter` and `tp_iternext` in c) cannot prevent gc callback , since `PyObject_GC_New` can still trigger gc, and calling stack of `tp_iter` on `dict` will contain `dict_iter` -> `dictiter_new` -> `PyObject_GC_New`(macro), and it SHOULD do that since `iter(some_dict)` should hold a reference to `some_dict`.

Moreover, since `__del__` and weakref callback can execute arbitrary code, while `Py_DECREF` that could dealloc the object is widely used in C extension, C code is not automatically immune to reentrant. `__eq__` used by dict access can execute arbitrary code as well.(this could be avoided by adding an identity hash shim, use identity match of wrapped object as its `__eq__`).

(BTW, `dict.popitem()` will allocate a tuple, so it is not "atomic" as well.)

Special care was taken in https://github.com/python/cpython/pull/11623 for `lru_cache`.
But there is still something wrong in current implementation of `lru_cache`, if I understand correctly.
https://github.com/python/cpython/blob/6cbb57f62d345d7a5d6aeb1b3b5d37a845344d5e/Modules/_functoolsmodule.c#L975-L976 said The __eq__ method call is never made for the deletion (dict access #3) because it is an identity match.
But as
https://github.com/python/cpython/blob/6cbb57f62d345d7a5d6aeb1b3b5d37a845344d5e/Objects/dictobject.c#L2207-L2224
and
https://github.com/python/cpython/blob/6cbb57f62d345d7a5d6aeb1b3b5d37a845344d5e/Objects/dictobject.c#L970-L989
show, if there is hash collision, only the first inserted key can avoid `__eq__` call.

In "conclusion", it is not easy to make operation "atomic". I think `non-reentrant Lock + background thread` or `non-reentrant Lock + deferred callback queue` or ignoring error are reasonable options.

YouJiacheng:
Not mutating cache dict is satisfactory to me. WDYT?
```python
import functools
import weakref

class _ResultWrapper:
  def __init__(self, data):
    self.data = data

  def clean(self):
    self.data = None

def weakref_lru_cache(max_size=4096):
  def wrap(f):
    @functools.lru_cache(max_size)
    def cached(_, weak_arg_ref, *args, **kwargs):
      return _ResultWrapper(f(weak_arg_ref(), *args, **kwargs))

    @functools.wraps(f)
    def wrapper(weak_arg, *args, **kwargs):
      assert type(weak_arg).__hash__ is object.__hash__
      assert type(weak_arg).__eq__ is object.__eq__
      tctx = None
      weak_arg_ref = weakref.ref(weak_arg, lambda _: result_wrapper.clean())
      # callback won't be triggered before name 'result_wrapper' definition because we hold weak_arg here
      result_wrapper = cached(tctx, weak_arg_ref, *args, **kwargs)
      return result_wrapper.data

    wrapper.cache_clear = cached.cache_clear
    wrapper.cache_info = cached.cache_info
    return wrapper
  return wrap
```

PR #10120, opened 2022-04-01
prototyping dynamic shapes

mattjj:
Co-authored-by: Dougal Maclaurin <dougalm@google.com>

As a result of previous PRs, we could build dynamic shape jaxprs with `jax.make_jaxpr`, but we couldn't do much else.

The main goal here was to hook up dynamic shape fundamentals to `jit`, in two ways:
1. on the front end, add a way to annotate `jit`-decorated functions to indicate when we wish the jaxpr(s) built during `jit` tracing to abstract over particular axis sizes; and
2. on the back end, add a way to lower such jaxprs to dynamic-shape MHLO and compile/execute them (e.g. using IREE).

To accomplish that goal, the main changes here are:
1. **`WrappedFun.in_type`**: We added an optional `in_type` attribute to `WrappedFun` to represent an optional input type annotation for the wrapped Python callable, which (when present) determines how the callable should be traced to a jaxpr (e.g. on what Tracers). The `in_type` attribute has type `Tuple[Tuple[AbstractValue, bool], ...]`, i.e. tuples of (aval, bool) pairs. Each aval represents the type of an input binder, while the boolean indicates whether the argument is _explicit_, in the sense that it should be passed as an argument to the Python callable (otherwise it is _implicit_ to the Python callable, so that its value is inferred from e.g. axis sizes of other explicit arguments). That is, at the jaxpr level all arguments are explicit, but for the Python callable some arguments are implicit. The `in_type` annotation must be adapted by transformations (e.g. `jvp` needs to expand it to accommodate tangent types).
2. **`DBIdx`**: In `in_type`, the shape tuples of some avals may now contain de Bruijn indices which refer to binders at positions in `in_type`. That is, we represent a type like `a:i32[] x:f32[a]` as `a:i32[] x:f32[DBIdx(0)]`. This representation is cache-friendly.
3. **Add `abstracted_axes` to `jit`**: While `in_type` is a nice cleanup on its own, the main reason we put `in_type` on `WrappedFun` is so that we could take user-specified annotations on `jit` "at the api.py level" and plumb them to where they are consumed. Correspondingly we added the `abstracted_axes` argument, which was already on `make_jaxpr`, to `jit`.

All the old tests pass. We've manually tested simple IREE examples; we may make internal-only tests for those.

There are more tests to add but we may land-and-iterate.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -32,7 +32,8 @@ import threading
 import weakref
 import types
 from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
-                    Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable)
+                    Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable,
+                    List)
 from warnings import warn
 
 import numpy as np
@@ -43,9 +44,9 @@ from jax import core
 from jax import linear_util as lu
 from jax import stages
 from jax.core import eval_jaxpr
-from jax.tree_util import (tree_map, tree_multimap, tree_flatten, tree_unflatten,
+from jax.tree_util import (tree_map, tree_flatten, tree_unflatten,
                            tree_structure, tree_transpose, tree_leaves,
-                           tree_map, treedef_is_leaf, treedef_children,
+                           tree_multimap, treedef_is_leaf, treedef_children,
                            Partial, PyTreeDef, all_leaves, treedef_tuple)
 
 from jax._src import device_array
@@ -191,7 +192,7 @@ def _infer_argnums_and_argnames(
     fun: Callable,
     argnums: Union[int, Iterable[int], None],
     argnames: Union[str, Iterable[str], None],
-) -> Tuple[Tuple[int, ...], Tuple[str, ...]]:
+  ) -> Tuple[Tuple[int, ...], Tuple[str, ...]]:
   """Infer missing argnums and argnames for a function with inspect."""
   if argnums is None and argnames is None:
     argnums = ()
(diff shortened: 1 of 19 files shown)


jax/_src/api.py
```diff
@@ -228,15 +229,16 @@ def _infer_argnums_and_argnames(
 
 
 def jit(
-  fun: Callable,
```
mattjj, line 231: These are just indentation changes, not relevant to the main content of the PR. (Sorry for including them...)

jax/interpreters/partial_eval.py
```diff
@@ -1811,57 +1820,95 @@ def partial_eval_to_jaxpr_dynamic(fun: lu.WrappedFun, in_pvals: Sequence[Partial
...
+               for i, name in spec.items())   # TODO(mattjj): better shape error
+    s = [DBIdx(env[spec[i]]) if i in spec else d for i, d in enumerate(x.shape)]
+    return DShapedArray(tuple(s), x.dtype, False)
+  explicit_args = [arg_type(x, canon_spec(len(x.shape), s))
+                   for x, s in zip(flat_args, axes_specs)]
+  implicit_args = [core.ShapedArray((), dtypes.dtype('int32'))] * len(env)
+  which_explicit = [False] * len(implicit_args) + [True] * len(explicit_args)
+  return tuple(zip(implicit_args + explicit_args, which_explicit))
+
+def _input_type_from_tracers(
```
mattjj, line 1855: This is a special case and needs to be fixed...

jax/_src/api.py
```diff
@@ -228,15 +229,16 @@ def _infer_argnums_and_argnames(
...
-) -> stages.Wrapped:
+    fun: Callable,
+    *,
+    static_argnums: Union[int, Iterable[int], None] = None,
+    static_argnames: Union[str, Iterable[str], None] = None,
+    device: Optional[xc.Device] = None,
+    backend: Optional[str] = None,
+    donate_argnums: Union[int, Iterable[int]] = (),
+    inline: bool = False,
+    abstracted_axes: Optional[Any] = None,
```
gnecula, line 240: What is your plan for landing some of this API change before the dynamic shapes support is ready? We can call this `experimental_abstracted_axes` and document it as not ready for use. Or, we can add a special private _jit that exposes this API.

gnecula, line 240: I can't quite tell how is the abstract_axes supposed to work. I am guessing that this can be a sequence of specifications, one for each argument. Each argument specification can be a pytree matching the argument type. And at the leaves, we have a sequence of strings?

E.g., how to we specify that the 2nd and 3rd dimensions have abstract size and are equal for a function that takes an array with 3 dimensions?
```
abstract_axes=[(?, 'n', 'n')]
```

mattjj: I'm comfortable just with leaving it undocumented. Having a private `_jit` would be fine too though; we could probably reuse `_python_jit` for that purpose.

mattjj: > I can't quite tell how is the abstract_axes supposed to work.

Sounds like the lack of documentation is working as intended! 😆

mattjj: (We talked about this in a 1:1 meeting earlier today. The API is just a placeholder now.)

Issue #9970, opened 2022-03-20
Memory leak when repeatedly saving a variable during a host_callback call.


denizokt:
I have a use case where I want to use host_callback to save a python object (specifically a scipy SuperLU object) and update it over and over again within a jitted function. These should persist between jit calls. Unfortunately I am running into a memory leak. I have a minimal example below, where the memory consumptions blows up linearly as the code runs. Even if I insert gc.collect() in the host_callback function, it seems that the SuperLU objects are somehow being kept around.

```python
import numpy as np
import scipy.sparse
import scipy.sparse.linalg

import jax.experimental.host_callback as hcb
import jax
import jax.numpy as jnp

class SolveFnContainer:
    lu = None

def host_do_something(_):
    print('solving')

    n = 100000
    A = scipy.sparse.diags((-1*np.ones(n-1), -1*np.ones(n-1), 2*np.ones(n)), offsets=(-1, 1, 0), format='csc')
    SolveFnContainer.lu = scipy.sparse.linalg.spilu(A)
    return 0.0

def device_do_something():
    inputs = (0.0,)
    return hcb.call(host_do_something, inputs,
                    result_shape=jax.ShapeDtypeStruct((), np.float64))

def main():
    def body_fn(x):
        return device_do_something()

    def cond_fn(x):
        return x == 0.0

    return jax.lax.while_loop(cond_fn, body_fn, 0.0)

if __name__ == '__main__':
    jit_main = jax.jit(main)
    print(jit_main())
```

denizokt:
Further info:

I have reduced it to an even more minimal example.

```
from time import sleep

import numpy as np
import scipy.sparse
import scipy.sparse.linalg

import jax.experimental.host_callback as hcb
import jax

import os
import psutil


class Container:
    A = None


def host_do_something(_):
    n = 100000
    A = scipy.sparse.diags((-1*np.ones(n-1), -1*np.ones(n-1), 2*np.ones(n)), offsets=(-1, 1, 0), format='csc')
    Container.A = scipy.sparse.linalg.spilu(A)
    #Container.A = A
    return 0.0


def device_do_something(_):
    inputs = (0.0,)
    return hcb.call(host_do_something, inputs,
                    result_shape=jax.ShapeDtypeStruct((), np.float64))


if __name__ == '__main__':
    while True:
        device_do_something(0.0)
        #host_do_something(0.0)

        print(psutil.Process(os.getpid()).memory_info().rss / 1024 ** 2)
        sleep(0.2)
[...]

PR #10121, opened 2022-04-01
Avoid retracing when a host_callback.call is called multiple times wi…

hawkinsp:
…th the same function.

If we build a lambda in the host_callback.call() method, the identity of that lambda is different each time and will never lead to a primitive compilation cache hit. Instead, use a custom wrapper object with hash/equality.

This issue was found in passing while debugging #9970.


gnecula (approved):
Thanks

PR #10123, opened 2022-04-01
remove $ from commands

fabianp:
A few commands in this file were prefixed with $ which results in an invalid command when copied with the sphinx "copy" button (see https://jax.readthedocs.io/en/latest/developer.html#update-documentation)


jakevdp (approved):
Looks good, thanks!

Issue #5746, opened 2021-02-17
set tree_map = tree_multimap


mattjj:
Some users suggested that we should just set `tree_map = tree_multimap` in jax.tree_util. Seems reasonable! We can't think of a reason not to, so let's try it and see if anything breaks.

jakevdp:
Done in #6585

carlosgmartin:
@jakevdp @mattjj Since these functions have been unified, should [the docs](https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html) be updated to reference only `tree_map` to avoid confusion? Perhaps a deprecation warning should be added to `tree_multimap`?

jakevdp:
Yes, those are good ideas!

jakevdp:
Remaining TODO: remove deprecated `tree_multimap` function after July 1, 2022, in accordance with the [API Compatibility Policy](https://jax.readthedocs.io/en/latest/api_compatibility.html).

PR #10126, opened 2022-04-01
Deprecate jax.tree_util.tree_multimap

jakevdp:
Addresses #5746

`tree_multimap` has long been an alias for `tree_map`; it will be less confusing in the long run if we deprecate it.


soraros:
As per the comment [here](https://github.com/google/jax/blob/9693898e85c4ad6040ce8fbe2069f751114969e1/jax/__init__.py#L108-L117), maybe it's also nice to avoid using `jax.tree_map` and use `tree_util.tree_map` in the docs?

jakevdp:
> As per the comment [here](https://github.com/google/jax/blob/9693898e85c4ad6040ce8fbe2069f751114969e1/jax/__init__.py#L108-L117), maybe it's also nice to avoid using jax.tree_map and use tree_util.tree_map in the docs?

This change removes all uses of `tree_multimap` in the docs.

mattjj (approved):

jakevdp:
Ah, @soraros I see I misread your comment. That's a good point that we should start using the `tree_util` namespace explicitly... I think I'll leave that for a later change though, because it will touch many more files. Thanks!

PR #10140, opened 2022-04-05
jnp.diagonal: implement in terms of gather rather than sum

jakevdp:
The original implementation appears to be from #94. Perhaps indexing wasn't yet so mature?

Some limited benchmarks show that this is significantly faster on CPU & GPU; moderately faster on TPU (http://go/colabx-drive/1ICPK_5YFjxRS_-MZg9iZFGRXWgWrIH3F).

Existing tests are sufficient to cover this change: https://github.com/google/jax/blob/97d834feed8b5e454fc12fe182a4713741cb8ed1/tests/lax_numpy_test.py#L2877-L2894

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2478,16 +2478,11 @@ def diagonal(a, offset=0, axis1: int = 0, axis2: int = 1):
   perm = perm + [axis1, axis2]
   a = lax.transpose(a, perm)
 
-  # Mask out the diagonal and reduce over one of the axes
-  a = where(eye(a_shape[axis1], a_shape[axis2], k=offset, dtype=bool),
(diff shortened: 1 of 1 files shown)


hawkinsp:
If you are willing, try one more strategy: gather 128x128 blocks and then extract their diagonals with a sum, then concatenate. I'm most interested in how that performs on TPU in particular.

hawkinsp (approved):

jakevdp:
Do you mean 128x128 in all dimensions?

hawkinsp:
I meant "gather 128x128 blocks in the two most minor dimensions in the matrix case". The reasoning is that on TPU indexing in the two most minor dimensions is quite expensive, but indexing on other dimensions can be cheap. That said, you don't really know the tensor layout at the JAX level, so maybe it's not a good idea.

PR #10148, opened 2022-04-05
Add version int tuple `__version_info__` to JAX

sharadmv:


jakevdp (approved):
Looks good to me!

jax/version.py
```diff
@@ -13,5 +13,7 @@
 # limitations under the License.
 
 __version__ = "0.3.5"
+__version_info__ = tuple(map(int, __version__.split(".")))
```
yashk2810, line 16: Maybe use `.isdigit()`?

yashk2810: `tuple(int(i) for i in __version__.split(".") if i.isdigit())`

sharadmv: Done.

jax/version.py
```diff
@@ -13,5 +13,7 @@
 # limitations under the License.
 
 __version__ = "0.3.5"
+__version_info__ = tuple(int(i) for i in __version__.split(".") if i.isdigit())
 
 _minimum_jaxlib_version = "0.3.0"
+_minimum_jaxlib_version_info = tuple(int(i) for i in _minimum_jaxlib_version.split(".") if i.isdigit())
```
yashk2810, line 19: Instead of repeating this twice, maybe make a private function?

sharadmv: Done.

Issue #10144, opened 2022-04-05
Pascal, Hilbert and/or other special matrices in `jax.scipy.linalg`


pnkraemer:
Hi! 

`scipy.linalg` implements ![special matrices](https://docs.scipy.org/doc/scipy/reference/linalg.html#special-matrices). At the moment, `jax.scipy.linalg` does not do that, but it would be useful for me to have access to Pascal and Hilbert matrices via `jax`. (I need them for some state-space modelling stuff, if that is relevant information.)

I could make a pull request if it helps :)
As a starting point, I will explain what I am doing currently below, and I will also explain some difficulties with `jit`ting that might need to be resolved before a pull request? 

## Related issues 

* https://github.com/google/jax/issues/1646 (vaguely related I guess)
* https://github.com/google/jax/issues/3171 (see below for why this is related)

## Hilbert matrix

For example, a Hilbert matrix would be rather straightforward to implement, via 
```python
def hilbert(n):
    a = jnp.arange(n)
    return 1 / (a[:, None] + a[None, :] + 1)
```
A problem I could see with this implementation is appropriate `jit` and `vmap` behaviour: the size of the output depends on `n` :( . This is true for most special matrices, unfortunately.

Therefore, in my own code, I have split the function into something like
```python

def hilbert(n):
    a = jnp.arange(n)
    return _hilbert(a)

@jax.jit
def _hilbert(a):
    return 1 / (a[:, None] + a[None, :] + 1)
```
which at least allows to `jit` parts of the implementation. 
An alternative would be to provide a Hankel matrix (where the shape of the matrix is determined by the shapes of the input vectors, and which should be easier to `jit`), and then let users like me do the Hilbert-via-Hankel conversion depending on each application. Scipy's Hankel matrix, however, uses `np.lib.as_strided`, and I understand from issue #3171 that this is not intended to be supported.
But this can probably be resolved, if one decides to go down that road.

## Pascal matrix

Pascal matrices are a little less obvious to implement than Hankel/Hilbert matrices, because they use binomial coefficients. For my applications, I have so far computed those via `jax.lax.exp(jax.lax.lgamma(n+1.))` (because `jnp.prod(jnp.arange(1, n+1))` has `jit` issues again; if I remember correctly, I took the gamma-function-idea from `jet.py`) and then vmapped a factorial implementation via 
```python
def pascal(n):
    return _pascal(jnp.arange(n))

@jax.jit
def _pascal(a):
    return _binom(a[:, None], a[None, :])


[...]

PR #10161, opened 2022-04-06
Add Hilbert matrix to jax.scipy.linalg

pnkraemer:
This PR adds an implementation of Hilbert matrices of order `n` that mirrors the API of `scipy.linalg.hilbert(n)` (https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.hilbert.html#scipy.linalg.hilbert).

See #10144 for context.

Pinging @jakevdp, because he offered to review :)

### DIFF
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -29,6 +29,7 @@ jax.scipy.linalg
    eigh_tridiagonal
    expm
    expm_frechet
+   hilbert
    inv
    lu
    lu_factor
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -657,7 +657,7 @@ def sqrtm(A, blocksize=1):
       raise NotImplementedError("Blocked version is not implemented yet.")
   return _sqrtm(A)
 
-
+@_wraps(scipy.linalg.hilbert)
 @partial(jit, static_argnames=("n",))
 def hilbert(n):
   a = jnp.arange(n)
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -656,3 +656,13 @@ def sqrtm(A, blocksize=1):
   if blocksize > 1:
       raise NotImplementedError("Blocked version is not implemented yet.")
   return _sqrtm(A)
+
+
+@partial(jit, static_argnames=("n",))
+def hilbert(n):
+  a = jnp.arange(n)
+  return _arange_to_hilbert(a)
+
+@jit
+def _arange_to_hilbert(a):
+  return 1 / (a[:, None] + a[None, :] + 1)
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -661,8 +661,4 @@ def sqrtm(A, blocksize=1):
 @partial(jit, static_argnames=("n",))
 def hilbert(n):
   a = jnp.arange(n)
-  return _arange_to_hilbert(a)
-
-@jit
-def _arange_to_hilbert(a):
   return 1 / (a[:, None] + a[None, :] + 1)
--- a/jax/scipy/linalg.py
+++ b/jax/scipy/linalg.py
@@ -37,6 +37,7 @@ from jax._src.scipy.linalg import (
   svd as svd,
   tril as tril,
   triu as triu,
+  hilbert as hilbert,
 )
 
 from jax._src.lax.polar import (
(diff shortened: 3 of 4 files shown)


tests/linalg_test.py
```diff
@@ -1564,5 +1564,11 @@ def testSchurBatching(self, shape, dtype):
       Ts, Ss = vmap(lax.linalg.schur)(args)
       self.assertAllClose(reconstruct(Ss, Ts), args, atol=1e-4)
 
+  @parameterized.parameters(0, 1, 5)
```
pnkraemer, line 1567: `n=0` and `n=1` as corner cases, and if it works for `n=5` then it must be correct

jax/_src/scipy/linalg.py
```diff
@@ -656,3 +656,13 @@ def sqrtm(A, blocksize=1):
...
+
+@_wraps(scipy.linalg.hilbert)
+@partial(jit, static_argnames=("n",))
+def hilbert(n):
+  a = jnp.arange(n)
+  return _arange_to_hilbert(a)
+
+@jit
+def _arange_to_hilbert(a):
+  return 1 / (a[:, None] + a[None, :] + 1)
```
pnkraemer, line 668: As detailed in the comment in #10144, through this splitting of the function, the transformation part (`_arange_to_hilbert`) can be jitted completely, whereas `hilbert` must be decorated with `static_argnames`. If this is too verbose, I can change it. :)

docs/jax.scipy.rst
```diff
@@ -29,6 +29,7 @@ jax.scipy.linalg
    eigh_tridiagonal
    expm
    expm_frechet
+   hilbert
```
pnkraemer, line 32: Just to be sure: I added this manually. If there is some mechanism that makes this change automatically, which I have missed, please let me know :)

jax/_src/scipy/linalg.py
jakevdp: It's not necessary to split this into a separate helper function. Regarding JIT compilation, you've already decorated the main function wtih jit, so there is no benefit to separately jitting the implementation. Regarding code design, the two functions are so short that in my judgment delegating to a helper obscures rather than aids in understanding the code.

I'd suggest deleting lines 664-667

jax/scipy/linalg.py
```diff
@@ -37,6 +37,7 @@
   svd as svd,
   tril as tril,
   triu as triu,
+  hilbert as hilbert,
```
jakevdp, line 40: please alphabetize the imports

jax/_src/scipy/linalg.py
```diff
@@ -656,3 +656,13 @@ def sqrtm(A, blocksize=1):
   if blocksize > 1:
       raise NotImplementedError("Blocked version is not implemented yet.")
   return _sqrtm(A)
+
+@_wraps(scipy.linalg.hilbert)
+@partial(jit, static_argnames=("n",))
+def hilbert(n):
```
jakevdp, line 662: It surprises me that there's no `dtype` option. I know scipy doesn't provide it, but given the computational characteristics of accelerators, maybe the JAX version should? What do you think?

jakevdp (changes requested):
Thanks! A couple comments

jax/_src/scipy/linalg.py
pnkraemer: Good question. Would the `dtype` be passed on to `arange`, or would the resulting matrix be cast to the `dtype`, or both?

I am not as familiar with the computational characteristics of accelerators, but here are some thoughts:
* Other array-creation routines (e.g. https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.fft.fftfreq.html?highlight=fftfreq) don't do that, even though a similar case could be made here, too?
* Which types would even be relevant? Roughly those that are allowed by `arange` (Mostly `float`- and `complex`-style types)? 
* I am also wondering whether the `1/(a + aT)` computation might introduce some implicit type promotion that introduces subtle issues which would be hard to catch in tests. But I am just speculating...

So intuitively, I would not have added that option. Though I don't have a strong opinion, and if you think it is a good idea I will add a dtype argument :) Thoughts?

jakevdp: `1/(a + aT)` would promote to inexact, so any inexact dtype should be fine.

We can probably punt on this now and add it if/when someone requests it.

jakevdp (commented):
Looks good! Last thing - could you please squash all the changes into a single commit? (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

jakevdp:
(We'll also need to rebase on an updated main branch now to resolve conflicts)

jakevdp:
I'm going to close this since it's gone stale – feel free to re-open if you'd like to work on this again!

aaronatp:
Hi @jakevdp hope you're doing well. Can I open a PR for this issue?

PR #10165, opened 2022-04-06
Add protocol for ndarray.at

YouJiacheng:
interface and docstr are copied from lax_numpy.py _IndexUpdateHelper and _IndexUpdateRef

Note docstr are outdated, refering to deprecated `jax.ops`.

### DIFF
--- a/jax/_src/numpy/ndarray.py
+++ b/jax/_src/numpy/ndarray.py
@@ -14,12 +14,216 @@
 
 import abc
 from typing import Any, Tuple, Optional, Union
+from typing_extensions import Protocol
 
 from jax import core
 from jax.interpreters import pxla
 from jax._src import device_array
 import numpy as np
 
+class Indexed(Protocol):
+  """Helper object to call indexed update functions for an (advanced) index.
+
+  This object references a source array and a specific indexer into that array.
+  Methods on this object return copies of the source array that have been
+  modified at the positions specified by the indexer.
+  """
+
+  def get(self, indices_are_sorted=False, unique_indices=False,
+          mode=None, fill_value=None):
+    """Equivalent to ``x[idx]``.
+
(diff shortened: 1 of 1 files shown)


jakevdp:
What's the benefit of creating a protocol rather than typing with the actual object type returned by the function?

YouJiacheng:
1. the actual object type name starts with underscore. 
2. decoupling interface and implementation may be better, just like why there is _src.
3. prevent import cycle (though str annotation can be used)

YouJiacheng:
@jakevdp what do you think?

jakevdp:
Sorry, we've been chatting about this. I'm a bit uncomfortable with the level of duplication here. I think rather than creating new protocols that mirror the existing classes, we should just annotate the existing classes as the return type. They're in underscored paths, but I think there are fewer downsides in that route than in the duplicative protocol route. What do you think?

jakevdp:
(The downsides I'm thinking of mainly are obfuscation of the true types, the lack of testability of the correct mapping between the protocol and the class, the potential for duplicated docstrings to become out-of-sync, etc.)

YouJiacheng:
I agree with you. Non-duplication and testability is important, while I think we can remove duplicated docstring. 
But if we want to annotate the existing classes as the return type, how to resolve the cyclic import?
lax_numpy.py import ndarray.py, while existing classes are in lax_numpy.py

jakevdp:
Reading around, it seems that the cyclic import issue is a common problem for which there aren't any great solutions. Maybe the best is the idea of creating an abstract base class for the API that both the implementation and the protocol derive from, though to be honest I don't love that.

YouJiacheng:
@jakevdp Okay I have a nice solution avoiding cyclic import in #10322

PR #10176, opened 2022-04-06
Simplify `jnp.diagonal` and `jnp.trace`

lgeiger:
This PR simplifies `jnp.diagonal` and `jnp.trace` by using `jnp.moveaxis` instead of inlining its implementation . This doesn't change the behaviour as [`jnp.moveaxis` has the same implementation internally and uses `partial(jit, ..., inline=True)`](https://github.com/google/jax/blob/7ee6adb1a5363851dcc4feb5e106b16375f83366/jax/_src/numpy/lax_numpy.py#L868-L885).

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2391,9 +2391,6 @@ def trace(a, offset=0, axis1: int = 0, axis2: int = 1, dtype=None, out=None):
     raise NotImplementedError("The 'out' argument to jnp.trace is not supported.")
   lax_internal._check_user_dtype_supported(dtype, "trace")
 
-  axis1 = _canonicalize_axis(axis1, ndim(a))
-  axis2 = _canonicalize_axis(axis2, ndim(a))
-
   a_shape = shape(a)
   if dtype is None:
     dtype = _dtype(a)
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -2467,21 +2467,15 @@ def diag_indices_from(arr):
...
-  axis2 = _canonicalize_axis(axis2, a_ndims)
-  perm = [i for i in range(a_ndims) if i != axis1 and i != axis2]
-  perm = perm + [axis1, axis2]
-  a = lax.transpose(a, perm)
+  a = moveaxis(a, (axis1, axis2), (-2, -1))
 
+  a_shape = shape(a)
   diag_size = _max(0, _min(a_shape[axis1] + _min(offset, 0),
                            a_shape[axis2] - _max(offset, 0)))
-  i = arange(diag_size)
```
hawkinsp, line 2483: I would stick with arange here. It will generate an `iota` HLO op, rather than a constant containing the same values. This is preferable because `iota` is more compact to represent.

hawkinsp (changes requested):

jax/_src/numpy/lax_numpy.py
lgeiger: > It will generate an iota HLO op, rather than a constant containing the same values.

Makes sense, sorry I didn't catch this since `arange(_abs(offset), _abs(offset) + diag_size)` currently always generates a constant so I missed it when looking at the HLO.

:+1: I reverted the change and updated the PR to also use `jnp.moveaxis` in `jnp.trace`.

hawkinsp (approved):

PR #10190, opened 2022-04-08
Make `unsafe_call` a method of `XlaCompiledComputation`

sharadmv:
This is so if we store keepalives in the compiled computation, they won't be garbage-collected if we cache `unsafe_call`.


sharadmv:
Actually, I think this solution will unnecessarily increase dispatch times. We can solve the original keepalive problem by attaching them to `unsafe_call` directly.

PR #10204, opened 2022-04-08
Use `lax.iota` also for `jnp.arange(0, stop)`

lgeiger:
This PR updates `jnp.arange` to also use `lax.iota` when calling it with `jnp.arange(0, stop)`. `jnp.arange(0, stop)` and `jnp.arange(stop)` will now return the same Jaxpr.

This was inspired by https://github.com/google/jax/pull/10176#discussion_r846166229 as `jnp.arange` is often called with an explicit `start` argument that equals zero.

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -4309,6 +4309,8 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     # from https://github.com/google/jax/issues/3450
     self.assertAllClose(np_arange(2.5),
                         jnp.arange(2.5))
+    self.assertAllClose(np_arange(0., 2.5),
+                        jnp.arange(0., 2.5))
 
   def testArangeTypes(self):
     # Test that arange() output type is equal to the default types.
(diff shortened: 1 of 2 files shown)


mattjj:
Want to add a test using `jax.make_jaxpr`? I can advise more if that's unclear.
image

lgeiger:
> Want to add a test using `jax.make_jaxpr`?

Good idea! I added a small test that verifies the Jaxpr and a test that checks for the correct floating behaviour.

lgeiger:
@mattjj or @jakevdp looks like this PR slipped through the cracks, do you mind taking a quick look at it?

tests/lax_numpy_test.py
```diff
@@ -4965,6 +4967,11 @@ def testArangeJit(self):
     expected = jtu.with_jax_dtype_defaults(np.arange)(5)
     self.assertAllClose(ans, expected)
 
+  def testArangeJaxpr(self):
+    jaxpr = jax.make_jaxpr(lambda: jnp.arange(5))()
+    jaxpr_zero = jax.make_jaxpr(lambda: jnp.arange(0, 5))()
+    self.assertEqual(str(jaxpr_zero), str(jaxpr))
```
hawkinsp, line 4973: This may end up being an overly strict test, since it depends on the exact string representation. I don't think there's a significantly better way to write it for now.

hawkinsp (approved):

tests/lax_numpy_test.py
lgeiger: Good point! I changed the test to not rely on the exact string representation.

PR #10213, opened 2022-04-09
add is_leaf to tree_util.all_leaves

mattjj:


jax/_src/tree_util.py
```diff
@@ -106,7 +106,11 @@ def all_leaves(iterable):
   Returns:
```
zhangqiaorjc, line 106: would it be useful to also document `is_leaf` argument?

zhangqiaorjc (approved):

jax/_src/tree_util.py
mattjj: image

mattjj: (Done!)

froystig (approved):

hawkinsp:
Closing as stale.

PR #10216, opened 2022-04-09
Do not generate trivial gathers when indexing entire axis

lgeiger:
Array indexing has a fast path that doesn't use gather in cases when the original array isn't modified or only additional axes are added.
This PR extends this path to also handle cases where users pass explicit start and stop values that are equivalent to `slice(None)`. E.g. `np.arange(4)[0:6:1]` is equivalent to `np.arange(4)[:]`.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3706,11 +3706,15 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
       # Normalize the slice to use None when possible
       start, stop, step = i.start, i.stop, i.step
       try:
-        if ((step is None or core.symbolic_equal_dim(step, 1)) and
-            stop is not None and core.symbolic_equal_dim(stop, x_shape[x_axis])):
-          # The following is a useful special case with shape polymorphism
-          stop = None
-      except TypeError:
+        if step is None or core.symbolic_equal_dim(step, 1):
+          step = None
+        if step is None:
+          if start is None or core.symbolic_equal_dim(start, 0):
+            start = None
+          if stop is None or (not isinstance(stop, core.Tracer) and
+              core.greater_equal_dim(stop, x_shape[x_axis])):
+            stop = None
+      except (TypeError, core.InconclusiveDimensionOperation):
         pass
 
       # Handle slice(None)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -3706,11 +3706,14 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
...
-            stop is not None and core.symbolic_equal_dim(stop, x_shape[x_axis])):
-          # The following is a useful special case with shape polymorphism
-          stop = None
-      except TypeError:
+        if step is None or core.symbolic_equal_dim(step, 1):
+          step = None
+        if step is None:
+          if start is None or core.symbolic_equal_dim(start, 0):
+            start = None
+          if stop is None or core.greater_equal_dim(stop, x_shape[x_axis]):
```
lgeiger, line 3714: I am not 100% sure that it is OK to use `greater_equal_dim` here because it includes this special case:
https://github.com/google/jax/blob/35b32eef96e579e5f565c48f5962463f8d82192d/jax/core.py#L1515-L1521

If this could cause problems I am happy switch back to `symbolic_equal_dim` which would mean that indices larger than the size of the axis wouldn't use this fast path.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3706,11 +3706,14 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
...
+            stop = None
+      except (TypeError, core.InconclusiveDimensionOperation):
```
lgeiger, line 3716: Handling `InconclusiveDimensionOperation` errors was added since according to the docs `greater_equal_dim` might throw that which isn't the case for `symbolic_equal_dim`.

tests/lax_numpy_indexing_test.py
```diff
@@ -849,6 +849,11 @@ def testTrivialGatherIsntGenerated(self):
     self.assertEqual(len(jaxpr.jaxpr.eqns), 1)
     self.assertNotIn('gather', str(jaxpr))
 
+    jaxpr = jax.make_jaxpr(lambda x: x[0:6:1])(np.arange(4))
+    self.assertEqual(len(jaxpr.jaxpr.eqns), 0)
+    jaxpr = jax.make_jaxpr(lambda x: x[:4])(np.arange(4))
+    self.assertEqual(len(jaxpr.jaxpr.eqns), 0)
```
lgeiger, line 855: I thought about parametrizing the test but it turned out to be much more readable this way with some slight duplication.

jax/_src/numpy/lax_numpy.py
lgeiger: I added an additional check to make sure that `stop` isn't a `Tracer` which should make this work correctly even for a possible dynamic case.

jakevdp (approved):
Sorry for the delay on this - this looks really nice! Thanks

Issue #10219, opened 2022-04-10
jax.scipy.stats.logistic.logpdf overflows easily with negative input


kluopaja:
Running
```python
import jax
print(jax.scipy.stats.logistic.logpdf(-100.0))
print(jax.scipy.stats.logistic.logpdf(100.0))
```
outputs
```python
-inf
-100.0
```
which is unexpected since the logistic distribution should be symmetric.

The reason for this is the overflowing `exp` in the implementation:
https://github.com/google/jax/blob/35b32eef96e579e5f565c48f5962463f8d82192d/jax/_src/scipy/stats/logistic.py#L22-L24

An easy fix would be taking the absolute value of `x` at the start of the function but I don't know if this could introduce some problems with `jax.grad`.

YouJiacheng:
What about
```python
import jax
def logpdf(x): 
    return - 2. * jax.numpy.logaddexp(x / 2, lax.neg(x) / 2)
```

froystig:
As I understand it, this was fixed by #10220.

PR #10220, opened 2022-04-11
Fix#10219

YouJiacheng:
#10219

### DIFF
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -322,6 +322,13 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
                             tol=1e-3)
     self._CompileAndCheck(lax_fun, args_maker)
 
+  def testLogisticLogpdfOverflow(self):
+    # Regression test for https://github.com/google/jax/issues/10219
+    self.assertAllClose(
+      np.array([-100, -100], np.float32),
+      lsp_stats.logistic.logpdf(np.array([-100, 100], np.float32)),
+      check_dtypes=False)
+
   @genNamedParametersNArgs(1)
   def testLogisticPpf(self, shapes, dtypes):
     rng = jtu.rand_default(self.rng())
--- a/jax/_src/scipy/stats/logistic.py
+++ b/jax/_src/scipy/stats/logistic.py
@@ -16,12 +16,18 @@ import scipy.stats as osp_stats
 from jax.scipy.special import expit, logit
 
 from jax import lax
+from jax._src.lax.lax import _const as _lax_const
 from jax._src.numpy.util import _wraps
+from jax._src.numpy.lax_numpy import _promote_args_inexact
+from jax._src.numpy import lax_numpy as jnp
 
 
 @_wraps(osp_stats.logistic.logpdf, update_doc=False)
 def logpdf(x):
-  return lax.neg(x) - 2. * lax.log1p(lax.exp(lax.neg(x)))
+  x, = _promote_args_inexact("logistic.logpdf", x)
+  two = _lax_const(x, 2)
+  half_x = lax.div(x, two)
+  return lax.mul(lax.neg(two), jnp.logaddexp(half_x, lax.neg(half_x)))
 
 @_wraps(osp_stats.logistic.pdf, update_doc=False)
 def pdf(x):


tests/scipy_stats_test.py
```diff
@@ -322,6 +322,12 @@ def args_maker():
                             tol=1e-3)
     self._CompileAndCheck(lax_fun, args_maker)
 
+  def testIssue10219(self):
```
jakevdp, line 325: We've stopped naming tests this way because they're less descriptive than we'd like them to be. Can you do something like this instead? https://github.com/google/jax/blob/902fc0c3d2b3ec9b6034c66074984386ec35606f/tests/lax_numpy_test.py#L2081-L2082

jakevdp (changes requested):
Looks great! Thanks for the fix. Just one minor comment.

tests/scipy_stats_test.py
YouJiacheng: Done.

jakevdp (approved):
Thanks!

jakevdp:
The new implementation is causing a dtype mismatch in the `CompileAndCheck` test when X64=True, which I don't entirely understand. It may be indicative of some larger type promotion bug.

jakevdp:
The new implementation is causing a dtype mismatch in the `CompileAndCheck` test when X64=True, which I don't entirely understand. It may be indicative of some larger type promotion bug.

YouJiacheng:
Would you investigate this type promotion problem?

jakevdp:
Ah, I see, the issue is that you use the raw input `x` in the expression `x / 2`, so when `x` is a numpy array it follows numpy-style promotion and promotes to float64. You can fix this by calling `x = jnp.asarray(x)` at the start of the function, or alternatively by calling `lax` functions directly in place of Python arithmetic operators.

YouJiacheng:
Ah, it is really subtle! Calling `lax` functions can be not suitable since `lax` won't do type promotion, and need a manually promoted literal `2` (or `0.5` if use `mul`).  In addition,`lax.div(1, 2) == 0` since `1` and `2` are `int32`.
I have fixed it with
```python
def logpdf(x):
  # convert to JAX before using python arithmetic op
  # ensure correct type promotion behavior
  # see https://github.com/google/jax/pull/10220#issuecomment-1095444492
  half_x = jnp.multiply(x, 0.5)
  return -2 * jnp.logaddexp(half_x, -half_x)
```

jakevdp:
> Calling lax functions can be not suitable

Right, you'll see elsewhere that we use `lax._const` for this reason: https://github.com/google/jax/blob/9455254b9fd456fe0ca00851f5fd4ebfdf04aa5b/jax/_src/scipy/stats/poisson.py#L26-L31

I think that would be preferred here over introducing a `jnp` dependence.

YouJiacheng:
Right, but
```python
from jax._src.lax.lax import _const
assert _const(1, 0.5) == 0
```

YouJiacheng:
Do we need a `_promote_args_inexact` here?

YouJiacheng:
Done.
```python
def logpdf(x):
  x = _promote_args_inexact("logistic.logpdf", x)
  two = _lax_const(x, 2)
  half_x = lax.div(x, two)
  return lax.mul(lax.neg(two), jnp.logaddexp(half_x, lax.neg(half_x)))
```

YouJiacheng:
I make an empty force-push to re-run lint and typecheck which seems to fail with juptext internal bug
```
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/bin/jupytext", line 8, in <module>
    sys.exit(jupytext())
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/cli.py", line [489](https://github.com/google/jax/runs/5996732829?check_suite_focus=true#step:5:489), in jupytext
    exit_code += jupytext_single_file(nb_file, args, log)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/cli.py", line 557, in jupytext_single_file
    notebook = read(nb_file, fmt=fmt, config=config)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/jupytext.py", line 409, in read
    return read(stream, as_version=as_version, fmt=fmt, config=config, **kwargs)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/jupytext.py", line 418, in read
    return reads(fp.read(), fmt, config=config, **kwargs)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/jupytext.py", line 370, in reads
    notebook = reader.reads(text, **kwargs)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/jupytext/jupytext.py", line 178, in reads
    return new_notebook(cells=cells, metadata=metadata)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/nbformat/v4/nbbase.py", line 164, in new_notebook
    validate(nb)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/nbformat/v4/nbbase.py", line 39, in validate
    return validate(node, ref=ref, version=nbformat)
  File "/home/runner/.cache/pre-commit/repobvvyoahl/py_env-python3.8/lib/python3.8/site-packages/nbformat/validator.py", line 281, in validate
    raise ValidationError("Non-unique cell id '{}' detected. Corrected to '{}'.".format(cell_id, cell['id']))
jsonschema.exceptions.ValidationError: Non-unique cell id '1022fbd7' detected. Corrected to 'b1d2593d'.

Error: The process '/opt/hostedtoolcache/Python/3.8.12/x64/bin/pre-commit' failed with exit code 1
```

jakevdp (approved):

PR #10221, opened 2022-04-11
Prefer `jnp.tile` over `concatenate`

lgeiger:
`jnp.tile` only uses broadcasting which tends to be preferred so this PR replaces `concatenate` with `jnp.tile` in the cases where it is possible.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3267,7 +3267,9 @@ def _roll(a, shift, axis):
                   np.broadcast_to(axis, b_shape)):
     i = _canonicalize_axis(i, a_ndim)
     x = remainder(x, (a_shape[i] or 1))
-    a = lax.concatenate((a, a), i)
+    reps = [1] * a_ndim
+    reps[i] = 2
+    a = tile(a, reps)
     a = lax.dynamic_slice_in_dim(a, a_shape[i] - x, a_shape[i], axis=i)
   return a
 
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -582,7 +582,7 @@ threefry_prng_impl = PRNGImpl(
 
 def _rbg_seed(seed: int) -> jnp.ndarray:
   halfkey = threefry_seed(seed)
-  return jnp.concatenate([halfkey, halfkey])
+  return jnp.tile(halfkey, 2)
 
 def _rbg_split(key: jnp.ndarray, num: int) -> jnp.ndarray:
   return vmap(_threefry_split, (0, None), 1)(key.reshape(2, 2), num).reshape(num, 4)


YouJiacheng:
I think we can use `jax.lax.broadcast_in_dim` + `jax.lax.reshape` for faster tracing time. There are lots of python code in `jnp.tile` and `jnp.broadcast_to` (used by `jnp.tile`).

jax/_src/numpy/lax_numpy.py
```diff
@@ -3245,7 +3245,9 @@ def _roll(a, shift, axis):
                   np.broadcast_to(axis, b_shape)):
     i = _canonicalize_axis(i, a_ndim)
     x = remainder(x, (a_shape[i] or 1))
-    a = lax.concatenate((a, a), i)
+    reps = [1] * a_ndim
+    reps[i] = 2
+    a = tile(a, reps)
```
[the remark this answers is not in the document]
YouJiacheng: Or maybe we can simplify implementation of `jnp.tile`?

YouJiacheng:
Or, for simplicity and better semantic we should use `jax.numpy.resize` in the first case.

jax/_src/prng.py
```diff
@@ -580,7 +580,7 @@ def threefry_random_bits(key: jnp.ndarray, bit_width, shape):
 
 def _rbg_seed(seed: int) -> jnp.ndarray:
   halfkey = threefry_seed(seed)
-  return jnp.concatenate([halfkey, halfkey])
+  return jnp.tile(halfkey, 2)
```
YouJiacheng, line 583: concatenate repeat axis 0, while tile repeat axis - 1. Are you sure halfkey is 1d array?

lgeiger: Yes, this seems to be the case:
https://github.com/google/jax/blob/3bfa6af2c81ac3c08b05d49b70c1bf01f09d28c5/jax/_src/prng.py#L246-L256

jakevdp: I think this will probably cause issues with custom PRNG keys, which don't concatenate like regular arrays. (`jnp.concatenate` checks for this, while `jnp.tile` does not). I'm going to request a review from @froystig, who would probably have recommendations here.

froystig: Good thinking @jakevdp. Actually, in this case, `tile` is valid use and is already supported! For more on how and why see #8381.

froystig (approved):

froystig:
@lgeiger, just to make sure you noticed: the CI checks show that one unit test (`LaxBackedNumpyTests.testNonScalarRepeats_fixed_size=False`) fails on python 3.7 with our oldest supported version of numpy (1.19.5).

lgeiger:
> just to make sure you noticed: the CI checks show that one unit test (`LaxBackedNumpyTests.testNonScalarRepeats_fixed_size=False`) fails on python 3.7 with our oldest supported version of numpy (1.19.5)

@froystig Sorry for the delay, but unfortunately I am unable to reproduce the test failure with Numpy 1.19.5 on macOS or even in a Python 3.7 Docker container. I am not sure what could cause this problem, since it doesn't seem to reproduce on all system.
For now I rebased my PR onto the latest main, maybe this will fix it? But this could also be indicative of some underlying problem that is platform specific.

froystig:
Thanks for taking a look and rebasing. Seems fine now...

PR #10226, opened 2022-04-11
Add np.polydiv numpy function

ntlm1686:
This PR is related to #7729 and #70.

Because of the error, the function can't pass the below test.
```python
self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False, tol=tol)
```

Different from the ```polymul```, the size of the residual is not fixed, and it's also hard to distinguish if a value in the result is an error or it's supposed to be small. If you use ```trim_zeros``` to remove the leading zeros in the residual polynomial, because of the error, the shape of jnp result may not match the result from np, so it won't pass ```_CheckAgainstNumpy```.

If we want to remove the values which is supposed to be zero(but not zero because of the error), we may use ```where(x<some_value)``` to make it zero. However, if the other values in the result are big(e.g. 1e8), the error can also be big(e.g. 1). Then we have to use ```where(x < max(some_value, some_value*max(x)))``` to remove those values, then the function won't pass ```_CompileAndCheck```.

Hi @jakevdp - could you please take a look and give some advice?

### DIFF
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -305,6 +305,7 @@ namespace; they are listed below.
     poly
     polyadd
     polyder
+    polydiv
     polyfit
     polyint
     polymul
--- a/jax/_src/numpy/polynomial.py
+++ b/jax/_src/numpy/polynomial.py
@@ -21,7 +21,7 @@ from jax import jit
 from jax import lax
 from jax._src.numpy.lax_numpy import (
     all, arange, argmin, array, asarray, atleast_1d, concatenate, convolve, diag, dot, finfo,
-    full, hstack, maximum, ones, outer, sqrt, trim_zeros, true_divide, vander, zeros)
+    full, hstack, maximum, ones, outer, sqrt, trim_zeros, trim_zeros_tol, true_divide, vander, zeros)
 from jax._src.numpy import linalg
 from jax._src.numpy.util import _check_arraylike, _promote_dtypes, _promote_dtypes_inexact, _wraps
 import numpy as np
@@ -277,6 +277,9 @@ def polyder(p, m=1):
 _LEADING_ZEROS_DOC = """\
 Setting trim_leading_zeros=True makes the output match that of numpy.
 But prevents the function from being able to be used in compiled code.
+Due to differences in accumulation of floating point arithmetic errors, the cutoff for values to be
+considered zero may lead to inconsistent results between NumPy and JAX, and even between different
+JAX backends. The result may lead to inconsistent output shapes when trim_leading_zeros=True.
 """
 
 @_wraps(np.polymul, lax_description=_LEADING_ZEROS_DOC)
@@ -292,6 +295,23 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
+def polydiv(u, v, *, trim_leading_zeros=False):
+  _check_arraylike("polydiv", u, v)
+  u, v = _promote_dtypes_inexact(u, v)
+  m = len(u) - 1
+  n = len(v) - 1
+  scale = 1. / v[0]
+  q = zeros(max(m - n + 1, 1), dtype = u.dtype) # force same dtype
+  for k in range(0, m-n+1):
+    d = scale * u[k]
+    q = q.at[k].set(d)
+    u = u.at[k:k+n+1].add(-d*v)
+  if trim_leading_zeros:
+    # use the square root of finfo(dtype) to approximate the absolute tolerance used in numpy
+    return q, trim_zeros_tol(u, tol=sqrt(finfo(u.dtype).eps), trim='f')
+  else:
+    return q, u
 
 @_wraps(np.polysub)
 @jit
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -298,6 +298,7 @@ from jax._src.numpy.polynomial import (
     poly as poly,
     polyadd as polyadd,
     polyder as polyder,
+    polydiv as polydiv,
     polyfit as polyfit,
     polyint as polyint,
     polymul as polymul,
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -2875,6 +2875,37 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     self._CompileAndCheck(jnp_fun_co, args_maker, check_dtypes=False)
 
   @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "a_shape={} , b_shape={}".format(
+          jtu.format_shape_dtype_string(a_shape, dtype),
+          jtu.format_shape_dtype_string(b_shape, dtype)),
+       "dtype": dtype, "a_shape": a_shape, "b_shape" : b_shape}
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes
+      for b_shape in one_dim_array_shapes))
+  def testPolyDiv(self, a_shape, b_shape, dtype):
+    rng = jtu.rand_default(self.rng())
+
+    def np_fun(arg1, arg2):
+      q, r = np.polydiv(arg1, arg2)
+      while r.size < max(arg1.size, arg2.size): # Pad residual to same size
+        r = np.pad(r, (1, 0), 'constant')
+      return q, r
+
+    def jnp_fun(arg1, arg2):
+      q, r = jnp.polydiv(arg1, arg2, trim_leading_zeros=True)
+      while r.size < max(arg1.size, arg2.size): # Pad residual to same size
+        r = jnp.pad(r, (1, 0), 'constant')
+      return q, r
+
+    args_maker = lambda: [rng(a_shape, dtype), rng(b_shape, dtype)]
+    tol = {np.float16: 2e-1, np.float32: 5e-2, np.float64: 1e-13}
+
+    jnp_compile = jnp.polydiv # Without trim_leading_zeros (trim_zeros make it unable to be compiled by XLA)
+    self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False, tol=tol)
+    self._CompileAndCheck(jnp_compile, args_maker, check_dtypes=True, atol=tol, rtol=tol)
+
+
+  @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_shape={}_offset={}_axis1={}_axis2={}".format(
           jtu.format_shape_dtype_string(shape, dtype), offset, axis1, axis2),
        "dtype": dtype, "shape": shape, "offset": offset, "axis1": axis1,
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2543,6 +2543,17 @@ def trim_zeros(filt, trim='fb'):
   return filt[start:len(filt) - end]
 
 
+def trim_zeros_tol(filt, tol, trim='fb'):
+  filt = core.concrete_or_error(asarray, filt,
+    "Error arose in the `filt` argument of trim_zeros_tol()")
+  nz = (abs(filt) < tol)
+  if all(nz):
+    return empty(0, _dtype(filt))
+  start = argmin(nz) if 'f' in trim.lower() else 0
+  end = argmin(nz[::-1]) if 'b' in trim.lower() else 0
+  return filt[start:len(filt) - end]
+
+
 @_wraps(np.append)
 @partial(jit, static_argnames=('axis',))
 def append(arr, values, axis: Optional[int] = None):


jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +293,18 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
+def polydiv(a1, a2):
+  a1, a2 = _promote_dtypes_inexact(a1, a2)
+  m = len(a1) - 1
+  n = len(a2) - 1
+  scale = 1. / a2[0]
+  q = zeros(builtins.max(m - n + 1, 1))
```
jakevdp, line 302: I don't think `builtins` here is required, as `max` is not redefined in this file.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +293,18 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
+def polydiv(a1, a2):
+  a1, a2 = _promote_dtypes_inexact(a1, a2)
```
jakevdp, line 298: Please call `_check_arraylike` before promoting dtypes

docs/jax.numpy.rst
```diff
@@ -308,6 +308,7 @@ namespace; they are listed below.
     polyfit
     polyint
     polymul
+    polydiv
```
jakevdp, line 311: Please alphabetize

jax/_src/numpy/polynomial.py
```diff
@@ -16,12 +16,13 @@
...
 import operator
 
+import builtins
 from jax import core
 from jax import jit
 from jax import lax
 from jax._src.numpy.lax_numpy import (
     all, arange, argmin, array, asarray, atleast_1d, concatenate, convolve, diag, dot, finfo,
-    full, hstack, maximum, ones, outer, sqrt, trim_zeros, true_divide, vander, zeros)
+    full, hstack, maximum, ones, outer, sqrt, trim_zeros, true_divide, vander, zeros, where)
```
jakevdp, line 25: Import is unused?

tests/lax_numpy_test.py
```diff
@@ -2874,6 +2874,24 @@ def testPolyMul(self, a1_shape, a2_shape, dtype):
...
+          jtu.format_shape_dtype_string(a_shape, dtype),
+          jtu.format_shape_dtype_string(b_shape, dtype)),
+       "dtype": dtype, "a_shape": a_shape, "b_shape" : b_shape}
+      for dtype in default_dtypes
+      for a_shape in one_dim_array_shapes
+      for b_shape in one_dim_array_shapes))
+  def testPolyDiv(self, a_shape, b_shape, dtype):
+    rng = jtu.rand_default(self.rng())
+    np_fun = lambda arg1, arg2: np.polydiv(arg1, arg2)
+    jnp_fun = lambda arg1, arg2: jnp.polydiv(arg1, arg2)
```
jakevdp, line 2888: No need for lambdas, we can write `np_fun = np.polydiv` and `jnp_fun = jnp.polydiv`, or just use `np.polydiv` and `jnp.polydiv` directly below.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +293,18 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
+def polydiv(a1, a2):
```
jakevdp, line 297: Numpy names its arguments `u` and `v`; we should match that in the jax implementation.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +293,18 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
```
jakevdp, line 296: This does not have a `trim_leading_zeros` argument, so I don't think `_LEADING_ZEROS_DOC` applies.

jax/numpy/__init__.py
```diff
@@ -301,6 +301,7 @@
     polyfit as polyfit,
     polyint as polyint,
     polymul as polymul,
+    polydiv as polydiv,
```
jakevdp, line 304: Please alphabetize imports

tests/lax_numpy_test.py
```diff
@@ -2874,6 +2874,24 @@ def testPolyMul(self, a1_shape, a2_shape, dtype):
...
+    args_maker = lambda: [rng(a_shape, dtype), rng(b_shape, dtype)]
+    tol = {np.float16: 2e-1, np.float32: 5e-2, np.float64: 1e-13}
+    # self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker, check_dtypes=False, tol=tol)
```
jakevdp, line 2891: We need to check this against NumPy to ensure the implementation is correct.

jakevdp (changes requested):
Looks like a good start - a few comments below

jakevdp:
Look at how `polymul` addresses the trimming of leading zeros - we'll need to do the same here to allow it to match numpy's output and to be compatible with JIT.

ljjsalt:
> Look at how `polymul` addresses the trimming of leading zeros - we'll need to do the same here to allow it to match numpy's output and to be compatible with JIT.

Hi, I am afraid it's impossible to match the output as numpy, since there is error in the result, and it effects the shape of the output.

jakevdp:
Can we match numpy when computing in float64?

ljjsalt:
> Can we match numpy when computing in float64?

As long as there is error, and we want to trim the zero, we won't be able to match the result from numpy.
The precision here doesn't matter. It's different from ```polymul``` where the output shape is fixed, and we can use `tolerance`. In ```polydiv```, the error effects the output shape.

jakevdp:
Just to make sure I'm understanding the issue - is it that we need to trim zeros based on some cutoff, and the exact floating point roundoffs involved differ between JAX and numpy so that it's difficult to match the correct cutoff exactly between them?

ljjsalt:
> Just to make sure I'm understanding the issue - is it that we need to trim zeros based on some cutoff, and the exact floating point roundoffs involved differ between JAX and numpy so that it's difficult to match the correct cutoff exactly between them?

Yes, we need to trim zeros based on some cutoff. And the cutoff is hard to determine(a constant like 1e-5 won't work).

jakevdp:
OK - well, at the very least we need to add a `trim_leading_zeros` parameter similar to that used in `polymul`, and then perhaps we can add a warning in the docstring regarding the fact that if its set to True the output may not match numpy exactly due to floating point roundoff (I suspect the same issue probably exists in `poymul` as well, it's just rarer to hit it due to multiplication being more stable than division)

Then we can rewrite the tests in such a way that they account for this difference. What do you think?

ljjsalt:
> OK - well, at the very least we need to add a `trim_leading_zeros` parameter similar to that used in `polymul`, and then perhaps we can add a warning in the docstring regarding the fact that if its set to True the output may not match numpy exactly due to floating point roundoff (I suspect the same issue probably exists in `poymul` as well, it's just rarer to hit it due to multiplication being more stable than division)
> 
> Then we can rewrite the tests in such a way that they account for this difference. What do you think?

Sure, I can do that. Thank you.

jakevdp (changes requested):
Looks great - thanks for taking the comments into account!

A couple more minor comments - also, once you address them please squash your changes into a single commit and force-push (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests for information).

Thanks!

ljjsalt:
> Looks great - thanks for taking the comments into account!
> 
> A couple more minor comments - also, once you address them please squash your changes into a single commit and force-push (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests for information).
> 
> Thanks!

Hi, could you please take a look why does others commit appear in my PR?

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +295,22 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
...
+  m = len(u) - 1
+  n = len(v) - 1
+  scale = 1. / v[0]
+  q = zeros(max(m - n + 1, 1))
+  for k in range(0, m-n+1):
+    d = scale * u[k]
+    q = q.at[k].set(d)
+    u = u.at[k:k+n+1].add(-d*v)
+  if trim_leading_zeros:
+    return q, trim_zeros(where(abs(u)<1e-3, 0, u), trim='f')
```
jakevdp, line 311: I think this is probably fine, but we may need to revisit this cutoff. I think numpy always converts to float64 and uses an `rtol=1E-14` cutoff. In JAX we may be computing in float32, bfloat16, etc. and perhaps we should change the cutoff depending on the width of the dtype.

jakevdp (approved):

jax/_src/numpy/polynomial.py
jakevdp: Since the PR needs a flake fix, can you also add a TODO here to revisit the tolerance depending on the dtype?

ljjsalt: No problem. Since I will use the ```jnp.polydiv``` myself, I will probably do some grid search to find the best tolerance here.

jakevdp: I would probably choose some multiple of `jnp.finfo(dtype).eps` as a baseline.

ljjsalt: Just used the ```jnp.finfo(dtype).eps``` for cutoff and also added some code for ```dtype```.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +295,25 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
   val = convolve(a1, a2, mode='full')
   return val
 
+@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
+def polydiv(u, v, *, trim_leading_zeros=False):
+  _check_arraylike("polydiv", u, v)
+  u, v = _promote_dtypes_inexact(u, v)
+  w = u[0] + v[0] # calculate type
```
jakevdp, line 302: This shouldn't be necessary, because the outputs of `_promote_dtypes_inexact` will all have the same dtype.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +295,25 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
...
+  scale = 1. / v[0]
+  q = zeros(max(m - n + 1, 1), dtype = w.dtype) # force same dtype
+  r = u.astype(w.dtype)
+  for k in range(0, m-n+1):
+    d = scale * r[k]
+    q = q.at[k].set(d)
+    r = r.at[k:k+n+1].add(-d*v)
+  if trim_leading_zeros:
+    return q, trim_zeros(
+      where((~(abs(r)<finfo(w.dtype).eps)).cumsum().astype(bool), r, 0), trim='f')
```
jakevdp, line 314: Numpy's version of this is here: https://github.com/numpy/numpy/blob/d410c8da4b9d7ef7e7be355f6804bc0f2ee9a2d5/numpy/lib/polynomial.py#L1042-L1043

Interestingly, the `rtol=1E-14` has no effect since it's multiplied by the second argument, which is zero. Since `atol` defaults to `1E-8`, this means it's effectively equivalent to using `1E-8` in place of `finfo(w.dtype).eps`, suggesting thta 1E-8 is the correct choice for `float64`.

It suggests that using `eps` directly as the cutoff is too tight a constraint. So maybe `np.sqrt(finfo(w.dtype).eps)` would be closer to the spirit of what numpy does here?

jakevdp: Also, I think the implementation here is slightly off, because this will zero-out *every* small entry in the result. The elements that are not trimmed should be left unmodified. It may be better to define an alternate `trim_zeros` method that accounts for the tolernace.

jakevdp (changes requested):

jax/_src/numpy/polynomial.py
ljjsalt: I changed the code, so the zero-out where stop when it found the first big element. ```trim_zeros``` does the same, it stops when it found the first non-zero element. In this case, both zero-out and trim_zeros visit from left to right. When they stop, rest of the elements are left unmodified.

jakevdp: Looks like the new changes didn't make it into the push?

ljjsalt: It's the ```cumsum```,
A mask like ```[False, False, True, False, True]``` becomes ```[False, False, True, True, True]```.

jakevdp: Oh, I see, thanks. I also had a comment about the eps needing to be adjusted as well – what do you think?

ljjsalt: Yeah, I added the ```sqrt``` in the latest code. Sorry, I squashed all the commits so you didn't notice.

jax/_src/numpy/polynomial.py
```diff
@@ -292,6 +295,25 @@ def polymul(a1, a2, *, trim_leading_zeros=False):
...
+  scale = 1. / v[0]
+  q = zeros(max(m - n + 1, 1), dtype = w.dtype) # force same dtype
+  r = u.astype(w.dtype)
+  for k in range(0, m-n+1):
+    d = scale * r[k]
+    q = q.at[k].set(d)
+    r = r.at[k:k+n+1].add(-d*v)
+  if trim_leading_zeros:
+    return q, trim_zeros(
+      where((~(abs(r)<sqrt(finfo(w.dtype).eps))).cumsum().astype(bool), r, 0), trim='f')
```
jakevdp, line 314: I imagine looking at this line in a year and having a lot of trouble understanding the logic. I would factor this out into a `trim_zeros_tol(filt, tol)` utility function that accepts a tolerance, and then add a comment describing why you are passing in the square root of eps

ljjsalt: Could you please tell me where I should put the ```trim_zeros_tol(filt, tol)```?

jakevdp (approved):
This looks great, thanks. One last thing: could you squash the changes into a single commit? Then we should be able to get this merged.

ljjsalt:
> This looks great, thanks. One last thing: could you squash the changes into a single commit? Then we should be able to get this merged.

No problem.

jakevdp:
Please fix the flake8 import issue - thanks!

ljjsalt:
> Please fix the flake8 import issue - thanks!

Sorry, it's fixed.

jakevdp:
Additional testing has revealed a few mismatches between `np.polydiv` and `jnp.polydiv`. In particular, there seems to be an issue when the output has a zero in the largest element (revealed by `LaxBackedNumpyTests.testPolyDiva_shape=int8[6] , b_shape=int8[12]`)

I think you can reproduce this locally with
```
$ JAX_NUM_GENERATED_CASES=90 pytest "LaxBackedNumpyTests.testPolyDiva_shape=int8[6] , b_shape=int8[12]"
```

ljjsalt:
> Additional testing has revealed a few mismatches between `np.polydiv` and `jnp.polydiv`. In particular, there seems to be an issue when the output has a zero in the largest element (revealed by `LaxBackedNumpyTests.testPolyDiva_shape=int8[6] , b_shape=int8[12]`)
> 
> I think you can reproduce this locally with
> 
> ```
> $ JAX_NUM_GENERATED_CASES=90 pytest "LaxBackedNumpyTests.testPolyDiva_shape=int8[6] , b_shape=int8[12]"
> ```

Fixed

jakevdp:
Thanks for sticking with this through all the back-and-forth!

ljjsalt:
> Thanks for sticking with this through all the back-and-forth!

Thank you for all the help!

Issue #10163, opened 2022-04-06
[sparse] Wrong jacobian computation for bcoo_spdot_general


lucagrementieri:
It seems that the Jacobian computed by `bcoo_spdot_general`, invoked through the operator `@` does not organize the values in the correct order.

In order to show the bug, I have written the following code.
```python
import jax
import jax.numpy as jnp
from jax.experimental.sparse.bcoo import BCOO

A_indices = jnp.array([[0, 1], [0, 2], [1, 1], [1, 2], [1, 0]])
A_values = jnp.array([-2.0, 1.0, -1.0, 0.5, 2.0])
A_shape = (2, 3)

B_indices = jnp.array([[0, 2], [2, 1], [0, 3], [1, 3], [1, 0], [0, 0]])
B_values = jnp.array([10.0, 100.0, 1000.0, -5.0, -50.0, -500.0])
B_shape = (3, 4)

def sp_sp_product(v1, v2):
    sparse1 = BCOO((v1, A_indices), shape=A_shape)
    sparse2 = BCOO((v2, B_indices), shape=B_shape)
    return (sparse1 @ sparse2).todense()

def sp_dense_product(v1, v2):
    sparse1 = BCOO((v1, A_indices), shape=A_shape)
    dense2 = BCOO((v2, B_indices), shape=B_shape).todense()
    return sparse1 @ dense2

sp_sp_product_jac = jax.jacfwd(sp_sp_product, argnums=1)(A_values, B_values)
sparse_dense_product_jac = jax.jacfwd(sp_dense_product, argnums=1)(A_values, B_values)

assert jnp.array_equal(
    sp_sp_product_jac, sparse_dense_product_jac
), "The two Jacobian should be the same"
```

`sparse_dense_product_jac` is correct (as it can be verified analitically), while `sp_sp_product_jac` is incorrect since it should be identical to `sparse_dense_product_jac` but it is not.
The shape of `sp_sp_product_jac` in the expected one and the values in `sp_sp_product_jac` seems correct but they are not placed in the expected location. Probably there is a problem in the indices of the sparse jacobian matrix.

lucagrementieri:
@jakevdp As the author of https://github.com/google/jax/pull/8848, have you any thought about it?

jakevdp:
I'll take a look

jakevdp:
Still working on debugging this... it's a tricky one.

lucagrementieri:
[...]

PR #10239, opened 2022-04-11
[sparse] make bcoo_sort_indices a primitive

jakevdp:
This started as an effort to make `sum_duplicates` a primitive to address the autodiff issues revealed by #10163. I thought about rolling `sum_duplicates` and `sort_indices` into the same primitive, but this ended up adding a level of complexity that I was not happy with. Since `sort_indices` is simpler, I decided to start with that on its own; I'll add a separate `sum_duplicates` primitive in a followup.

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -132,29 +132,6 @@ def _bcoo_sum_duplicates_unbatched(data, indices, *, shape, nse, remove_zeros):
   data_unique = jnp.where(oob_mask[(...,) + props.n_dense * (None,)], 0, data_unique)
   return data_unique, indices_unique, nse
 
-def _bcoo_sort_indices(data, indices, shape):
-  props = _validate_bcoo(data, indices, shape)
-  if props.n_sparse == 0:
-    return data, indices
-  def f(data, indices):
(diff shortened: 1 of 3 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -1375,6 +1351,101 @@ def _bcoo_spdot_general_jvp(primals, tangents, **kwds):
...
+  for _ in range(props.n_batch):
+    f = broadcasting_vmap(f)
+  indices_out, perm = f(indices)
+  permute = lambda d, p: d[p]
+  for _ in range(props.n_batch):
+    permute = broadcasting_vmap(permute)
+  data_out = permute(data, perm)
+
+  indices_dot_out = ad.Zero.from_value(indices)
+  data_dot_out = ad.Zero.from_value(data_out) if type(data_dot) is ad.Zero else permute(data_dot, perm)
```
froystig, line 1435: Looks like this idea worked out!

froystig (approved):

jax/experimental/sparse/bcoo.py
jakevdp: Yeah, returning the permutation was the key

Issue #10247, opened 2022-04-12
Surprising behavior of `jax.scipy.special.expit` and `jax.scipy.special.logit`


YouJiacheng:
```python
import jax
from jax.scipy.special import expit
expit(1) # error
expit([1.0, 2.0]) # DeviceArray([0.7310586, 0.880797 ], dtype=float32)
jax.jacrev(expit)([1.0, 2.0]) # confusing error: only integer scalar arrays can be converted to a scalar index
```
I think
```python
def expit(x):
  x = asarray(x)
  one = _lax_const(x, 1)
  return lax.div(one, lax.add(one, lax.exp(lax.neg(x))))
```
should be
```python
def expit(x):
  x, = _promote_args_inexact("expit", x)
  one = _lax_const(x, 1)
  return lax.div(one, lax.add(one, lax.exp(lax.neg(x))))
```

PR #10248, opened 2022-04-12
Fix#10247

YouJiacheng:
#10247

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -24,7 +24,7 @@ from jax import lax, core
 from jax.interpreters import ad
 import jax.numpy as jnp
 from jax._src.lax.lax import _const as _lax_const
-from jax._src.numpy.lax_numpy import asarray, _reduction_dims, _promote_args_inexact
+from jax._src.numpy.lax_numpy import _reduction_dims, _promote_args_inexact
 from jax._src.numpy.util import _wraps
 
 from typing import Optional, Tuple
@@ -89,7 +89,7 @@ def erfinv(x):
 @api.custom_jvp
 @_wraps(osp_special.logit, update_doc=False)
 def logit(x):
-  x = asarray(x)
+  x, = _promote_args_inexact("logit", x)
   return lax.log(lax.div(x, lax.sub(_lax_const(x, 1), x)))
 logit.defjvps(
     lambda g, ans, x: lax.div(g, lax.mul(x, lax.sub(_lax_const(x, 1), x))))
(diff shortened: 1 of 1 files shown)


jax/_src/scipy/special.py
```diff
@@ -89,7 +89,7 @@ def erfinv(x):
 @api.custom_jvp
 @_wraps(osp_special.logit, update_doc=False)
 def logit(x):
-  x = asarray(x)
```
jakevdp, line 92: What should really happen here is to call `_check_arraylike` on the input, so that list inputs will lead to an error. The attached bug is one of the reasons we started enforcing that.

jakevdp: See #7737 for details

YouJiacheng: IIUC, you want to add a `_check_arraylike` at the beginning of function?
But other functions in this file doesn't have this check now.
Maybe we can add it for all functions in this file in another PR?

jakevdp: Why not this PR? If I understand the other issue correctly, it's that passing a list to the gradient of the function returns a confusing error – the reason that error is confusing is because we haven't called `_check_arraylike` at the start of the function.

jakevdp: Making it accept a list is not the long-term behavior we want, so I think we should avoid a change that allows that input.

jakevdp: The reason we haven't done this wholesale thorugh the package yet is that (1) we don't have any mechanism to automatically detect where the check is missing, and (2) each change introduces a backward-incompatibility for some small percentage of users, so it often requires some mitigation to land the change.

YouJiacheng: I think the proposed implementation will reject (though not clearly) a list input. Since `lax_internal._convert_element_type` in `_promote_args_inexact` won't accept a list.
Almost all functions in this file use this promotion, maybe we can add the check for them all in this PR without breaking backward-compatibility?
Could you please check above claim?

jakevdp: Ah, yes you're right, `_promote_args_inexact` does call `_check_arraylike`, so this change will work. Thanks!

jakevdp (approved):

jakevdp (commented):
The reason that this was never caught is because we only test `scipy.special` functions for floating point inputs. Perhaps some of these should be changed to `float_dtypes + int_dtypes`: https://github.com/google/jax/blob/c06eff8cd80da2bad6149cad306b2f74f12f0386/tests/lax_scipy_test.py#L118-L125

PR #10249, opened 2022-04-12
Add Jaxpr effects tests for higher order primitives

sharadmv:


jax/_src/ad_checkpoint.py
```diff
@@ -276,10 +276,10 @@ def remat_impl(*args, jaxpr, prevent_cse, differentiated, policy):
   del prevent_cse, differentiated, policy  # Unused.
   return core.eval_jaxpr(jaxpr, (), *args)
 
-@remat_p.def_abstract_eval
+@remat_p.def_effectful_abstract_eval
 def remat_abstract_eval(*args, jaxpr, prevent_cse, differentiated, policy):
   del args, prevent_cse, differentiated, policy  # Unused.
-  return [v.aval for v in jaxpr.outvars]
+  return [v.aval for v in jaxpr.outvars], jaxpr.effects
```
mattjj, line 282: I think this should be an error whenever `jaxpr.effects` is nonempty, until we work out the semantics of effects in remat, effect-by-effect. (It could be that allowing the debug effect is okay already, essentially by definition, but even there we should write down what we want the semantics to be.)

Similar for `custom_vjp`: until we work out the semantics we should type error when effects are present.

jax/core.py
```diff
@@ -2310,8 +2310,6 @@ def check_map(ctx_factory, prim, in_avals, params):
   if "call_jaxpr" not in params:
     raise JaxprTypeError(f"Map primitive {prim} missing 'call_jaxpr' parameter")
   call_jaxpr = params["call_jaxpr"]
-  if call_jaxpr.effects:
```
mattjj, line 2313: I don't think we should delete this check entirely. Maybe you want to allow certain effects in maps, but others are still a type error?

mattjj (changes requested):

mattjj (approved):

PR #10264, opened 2022-04-13
Remove mention of six package from jaxlib build instructions.

hawkinsp:
I verified that six is no longer needed.


jakevdp (approved):
LGTM!

PR #10265, opened 2022-04-13
Add libstdc++ workaround for conda users.

hawkinsp:


jakevdp (approved):
Looks good - this seems like a relatively lightweight way to help users with a common issue.

PR #663, opened 2019-04-30
np.tile implementation

sharadmv:
Adds a `numpy.tile` implementation for `lax_numpy`.

PR #10266, opened 2022-04-13
Deprecate DeviceArray.tile method

jakevdp:
We shouldn't add this method to DeviceArray objects, because numpy arrays do not have such a method. I checked that all other methods are present:
```python
In [1]: from jax._src.numpy.lax_numpy import _diff_methods, _nondiff_methods

In [2]: import numpy as np

In [3]: x = np.arange(10)

In [4]: for method in _diff_methods + _nondiff_methods: 
   ...:   if not hasattr(x, method): 
   ...:     print(method) 
   ...:                                                                                               
tile
```
According to https://jax.readthedocs.io/en/latest/api_compatibility.html, we should keep this deprecated method for at least 3 months.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 -->
 
 
+## jax 0.3.8 (Unreleased)
+* [GitHub
+  commits](https://github.com/google/jax/compare/jax-v0.3.7...main).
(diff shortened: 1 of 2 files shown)


hawkinsp (approved):
I don't object to the change, but it might be worth checking that it wasn't added for some reason in particular (e.g., a user request). When was it added?

jakevdp:
It looks like it was added here: https://github.com/google/jax/pull/663 I suspect the `tile()` method was an inadvertent addition.

PR #10270, opened 2022-04-13
add some simple iree tests

mattjj:
The following command passes (for dynamic shapes testing), though two of the more interesting tests fail with what might be IREE bugs (and so are currently skipped):

```shell
JAX_PLATFORMS='iree' pytest -n auto tests/core_test.py tests/api_test.py -k Dynamic
```

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -228,7 +228,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
   backend = xb.get_device_backend(device) if device else xb.get_backend(backend)
 
   if (config.jax_dynamic_shapes and jaxpr_has_bints(jaxpr) and
-      backend.platform != 'iree'):
+      not _backend_supports_unbounded_dynamic_shapes(backend)):
     jaxpr, consts = pe.pad_jaxpr(jaxpr, consts)
 
   # Computations that only produce constants and/or only rearrange their inputs,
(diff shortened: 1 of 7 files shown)


jax/_src/dispatch.py
```diff
@@ -228,7 +228,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
   backend = xb.get_device_backend(device) if device else xb.get_backend(backend)
 
   if (config.jax_dynamic_shapes and jaxpr_has_bints(jaxpr) and
-      backend.platform != 'iree'):
+      not backend.platform == 'iree'):
```
hawkinsp, line 231: Maybe give this a more declarative name via, e.g.

```
def _backend_supports_unbounded_dynamic_shapes(b):
 ...
```

(Probably even better to make it a property on the backend, but not today.)

tests/api_test.py
```diff
@@ -8116,21 +8116,45 @@ def test_jit_abstracted_axes_return_polymorphic_shape2(self):
...
+  @unittest.skipIf(jtu.device_under_test() == 'iree', "fails on iree")
   def test_slicing_basic(self):
     f = jax.jit(lambda x, n: jnp.sum(x[:n]))
     ans = f(jnp.arange(10), 3)
     expected = jnp.sum(jnp.arange(10)[:3])
     self.assertAllClose(ans, expected, check_dtypes=True)
 
+  # TODO(mattjj,dougalm,phawkins): debug iree failure, "failed to legalize
+  # operation 'mhlo.while' that was explicitly marked illegal"
+  @unittest.skipIf(jtu.device_under_test() == 'iree', "fails on iree")
```
hawkinsp, line 8157: `@jtu.skip_on_devices(...)`

hawkinsp (approved):

PR #10274, opened 2022-04-14
Make `safe_zip` and `safe_map` work with empty lists. 

romanngg:
Helps avoid error when calling e.g. `lax.reduce_window` on scalar inputs.


tests/lax_test.py
```diff
@@ -1784,6 +1784,14 @@ def testReduceWeakType(self, op_namespace, op, arr_weak_type, init_weak_type):
...
       for shape, dims, strides, padding, base_dilation, window_dilation in (
         itertools.chain(
+          itertools.product(
+            [()],
+            [()],
+            [()],
+            [()],
+            [()],
+            [()]
+          ),
```
jakevdp, line 1794: Maybe just directly write `[((), (), (), (), (), ())]` for clarity?

jax/_src/util.py
```diff
@@ -34,16 +34,18 @@
...
 
 def safe_map(f, *args):
   args = list(map(list, args))
-  n = len(args[0])
-  for arg in args[1:]:
-    assert len(arg) == n, 'length mismatch: {}'.format(list(map(len, args)))
+  if len(args) != 0:
+    n = len(args[0])
+    for arg in args[1:]:
+      assert len(arg) == n, 'length mismatch: {}'.format(list(map(len, args)))
```
jakevdp, line 48: Is this change necessary for this PR? It doesn't look like `safe_map` with empty args is used anywhere. Do we need to support that?

jax/_src/util.py
```diff
@@ -34,16 +34,18 @@
 T = TypeVar("T")
 
 def safe_zip(*args):
-  n = len(args[0])
-  for arg in args[1:]:
-    assert len(arg) == n, 'length mismatch: {}'.format(list(map(len, args)))
+  if len(args) != 0:
```
jakevdp, line 37: A benefit of the previous behavior was that `safe_zip` errored with empty arguments, which is sometimes an indication of something unexpected happening. What if we handle the empty padding case without modifying `safe_zip`?

jakevdp (changes requested):

tests/lax_test.py
romanngg: Good point! I added `VALID` and `SAME` padding values though, to test that these work on empty inputs too though, so kept the product structure.

jax/_src/util.py
romanngg: It's not, happy to revert it, but I thought it made sense to have it behave consistently with `safe_zip` (and the vanilla python `map`) - lmk if you still would like me to revert it.

jax/_src/util.py
romanngg: I personally think that the new behavior would be a bit better in that it would be consistent with the behavior of `map` and `zip`, as well as logically correct (an empty set satisfies any condition), so I think it would be better if empty arguments, if they are unexpected, would cause errors elsewhere, but not inside these methods. But I'm not that familiar with JAX codebase, so if you still prefer these to fail on empty entires, happy to revert the changes to both `safe_map` and `safe_zip`, let me know!

romanngg:
Looks like the change causes some failures though, will look into it.

jax/_src/util.py
jakevdp: Looping in @mattjj because I think he has opinions...

jax/_src/lax/windowed_reductions.py
```diff
@@ -457,8 +457,8 @@ def reduce_window_shape_tuple(operand_shape, window_dimensions, window_strides,
     operand_shape = lax._dilate_shape(operand_shape, base_dilation)
   if window_dilation is not None:
     window_dimensions = lax._dilate_shape(window_dimensions, window_dilation)
-  pads_lo, pads_hi = zip(*padding)
-  operand_padded = core.sum_shapes(operand_shape, pads_lo, pads_hi)
+  pads = zip(*padding)
```
jakevdp, line 460: My worry here is that now padding of arbitrary shape is accepted. Correct me if I'm wrong, but we only want to accept empty paddings, or paddings of length 2, correct?

jakevdp: Actually, it seems like the best fix here is not to make `zip` behave differently, but to replace `zip(*padding)` with `util.unzip2(padding)`, which I think will handle all the corner cases exactly as we need it to.

romanngg: Thanks Jake, this is indeed better, changed!

However, I noticed that having `safe_zip` to work with 0-length inputs is also used in the transpose rule of `reduce_window_sum`:
https://github.com/google/jax/blob/f0112ccda5afe5114855d67845db60fb14907de9/jax/_src/lax/windowed_reductions.py#L365

So I kept the `safe_map` and `safe_zip` changes for now.

I then added the identical test case to the respective autodiff test https://github.com/google/jax/pull/10274/files#diff-22c1f2ce24cf39ea41d2fbba031630102a443aa243435c054e0cbafb6b8183ee

But now I am getting a rather mysterious `fatal Python error`:
https://source.cloud.google.com/results/invocations/768551e1-a1dc-4f3e-9f3f-a35649a3fa0d/targets/jax%2Ftesting%2Finternal_numpy_scipy%2Fpresubmit_github/log
Do you have an idea of what might be causing it?

jax/_src/util.py
jakevdp: I think @mattjj had opinions on this - I'm going to request a review from him

jax/_src/lax/windowed_reductions.py
romanngg: @mattjj do you have ideas how to fix these errors?

The goal is to make calling `lax.reduce_window` and the respective forward/reverse-mode AD on scalars work, e.g.
```python
def f(x):
  return jax.lax.reduce_window(x, 0., jax.lax.add, (), (), ())

f(1.)
grad(f)(1.)
jvp(f (1.,), (1.,))
```

But my suggested fix seems to lead to a `fatal Python error` (https://source.cloud.google.com/results/invocations/605f2ece-dadd-4073-a048-6f366be90a3f/targets/jax%2Ftesting%2Fcpu%2Fpresubmit_github/log) - any suggestions on how to figure out this error, or fix the issue in a different way?

PR #10285, opened 2022-04-14
Simplify util.py

YouJiacheng:
And cover #10274

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -34,22 +34,18 @@ Seq = Sequence
 T = TypeVar("T")
 
 def safe_zip(*args):
-  n = len(args[0])
-  for arg in args[1:]:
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks for the contribution! For reasons why we don't want to simplify some of these, see the discussion in https://github.com/google/jax/pull/6758

YouJiacheng:
Thanks for nice explanation! But maybe some of changes doesn't break error message.

YouJiacheng:
@jakevdp could you please have a look again? I revert change on `unzip2` and `unzip3` and add a comment.

jakevdp:
Since this overlaps work in #10274, let's wait for a review there before making any decisions here.

jakevdp:
I think we want to keep these utilities as-is. Sorry we weren't able to merge this!

PR #10288, opened 2022-04-14
Fix PRNGKeyArray.broadcast_to with scalar shape

YouJiacheng:


jakevdp (changes requested):
Looks good - add a test?

YouJiacheng:
Should I add a test directly use PRNGKeyArray? Or just test with PRNGKey(will not use PRNGKeyArray unless config set custom prng key, while CI doesn't set this.)

jakevdp:
I would probably add the test somewhere near here: https://github.com/google/jax/blob/f8d796939212fd2b3651c10c5e00f2f7503453c2/tests/random_test.py#L1550-L1555

I don't think we test custom_prng on github CI, but we have other test coverage for that that will catch it (and eventually custom_prng will be on by default)

tests/random_test.py
```diff
@@ -1553,6 +1553,9 @@ def test_broadcast_to(self):
     ref = jnp.broadcast_to(like(key), (3,))
     self.assertEqual(out.shape, ref.shape)
     self.assertEqual(out.shape, (3,))
+    out = jnp.broadcast_to(key,        3)
```
jakevdp, line 1556: Use a single space please

jakevdp (approved):
Thanks!

PR #10290, opened 2022-04-14
Comment on implementation of unzip2 & unzip3

jakevdp:
Hopefully will prevent the perrennial PRs fixing these 😀

### DIFF
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -47,6 +47,9 @@ def safe_map(f, *args):
   return list(map(f, *args))
 
 def unzip2(xys):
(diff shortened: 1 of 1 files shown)


mattjj (approved):
LGTM, though maybe `safe_map(f, [])` should be allowed?

YouJiacheng:
Ah, #10274 want to make `safe_map` and `safe_zip` work with empty lists?

jakevdp:
I'm not sure whether we want that; see my comments on #10274

In any case, I chose not to touch `safe_zip` and `safe_map` here since there are two other PRs in-flight with different ideas about them.

PR #10295, opened 2022-04-14
Add in runtime tokens for effectful jaxprs

sharadmv:
This PR adds runtime tokens for effectful jaxprs. 

We want to support enforcing ordering for side-effects in JAX. JAX can execute compiled functions asychronously (for performance reasons), which is usually okay to do when there are no side-effects present. However, when there are side-effects, they could potentially be reordered. Take the following example:
```python
@partial(jit, device=<device 0>)
def f():
  <something expensive>
  do_effect()

@partial(jit, device=<device 1>)
def g():
  do_effect()

f()
g()
```
If `f()` and `g()` are executed asynchronously (e.g. we are using GPUs), the effect in `g()` will probably happen before the effect in `f()` because these two functions are happening in parallel. The single-threaded Python programming model, however, would enforce that `f()` is finished before we start executing `g()`.

If users want that sort of ordering for effects (the order in the code matches the order of execution), we need to sequence `f()` and `g()` in the JAX runtime. We can accomplish this by first passing a dummy value into `f()` which is then returned and passed into `g()`, resulting in the following transformed code:
```python
@partial(jit, device=<device 0>)
def f(dummy):
  <something expensive>
  do_effect()
  return dummy

@partial(jit, device=<device 1>)
def g(dummy):
  do_effect()
  return dummy

dummy = <some dummy value>
dummy = f(dummy)
dummy = g(dummy)
```
Now, despite `f` and `g` executing asynchronously, we can't call `g()` until we receive the return dummy value of `f()`, forcing the two calls to be sequenced.

We may not want to do this for all side-effects though, since now we lose out on the performance gains from parallelism by sequencing `f()` and `g()`. Instead, we enable this behavior via an allow-list, `core.ordered_effects`. These dummy values will be automatically threaded in and out of functions that have side-effects in `core.ordered_effects`.

As part of this PR, each side-effect will be provided a different dummy value so functions with the same effects will be sequenced.

This sequencing is only within a single Python thread. If a user would like to run the same ordered effect in parallel, they can do so via Python threads.

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -15,6 +15,7 @@
 # Primitive dispatch and jit dispatch.
 from __future__ import annotations
 
+import atexit
 import contextlib
 from functools import partial
 import itertools
@@ -24,6 +25,7 @@ from typing import (
 from typing_extensions import Protocol
 import os
 import re
+import threading
 import warnings
 
 from absl import logging
@@ -100,6 +102,37 @@ def apply_primitive(prim, *args, **params):
 # TODO(phawkins): update code referring to xla.apply_primitive to point here.
 xla.apply_primitive = apply_primitive
 
+RuntimeToken = Any
+
+class RuntimeTokenSet(threading.local):
+  tokens: Dict[core.Effect, Tuple[RuntimeToken, Device]]
+
+  def __init__(self):
+    self.tokens = {}
+
+  def get_token(self, eff: core.Effect, device: Device) -> RuntimeToken:
+    if eff not in self.tokens:
+      self.tokens[eff] = device_put(np.zeros(0, np.bool_), device), device
+    elif self.tokens[eff][1] != device:
+      (old_token,), _ = self.tokens[eff]
+      old_token.aval = core.ShapedArray((0,), np.bool_)
+      self.tokens[eff] = device_put(old_token, device), device
+    return self.tokens[eff][0]
+
+  def update_token(self, eff: core.Effect, token: RuntimeToken):
+    self.tokens[eff] = token, self.tokens[eff][1]
+
+  def clear(self):
+    self.tokens = {}
+
+  def block_until_ready(self):
+    [t[0].block_until_ready() for t, _ in self.tokens.values()]
+
+runtime_tokens: RuntimeTokenSet = RuntimeTokenSet()
+
+@atexit.register
+def wait_for_tokens():
+  runtime_tokens.block_until_ready()
 
 @util.cache()
 def xla_primitive_callable(prim, *arg_specs: ArgSpec, **params):
@@ -256,7 +289,8 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
   if not jaxpr.eqns and not always_lower:
     return XlaComputation(
         name, None, True, None, None, jaxpr=jaxpr, consts=consts, device=device,
-        in_avals=abstract_args, out_avals=out_avals, kept_var_idx=kept_var_idx)
+        in_avals=abstract_args, out_avals=out_avals, effects=jaxpr.effects,
+        kept_var_idx=kept_var_idx)
 
   if not _on_exit:
     log_priority = logging.WARNING if config.jax_log_compiles else logging.DEBUG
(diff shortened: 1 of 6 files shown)


jax/_src/dispatch.py
```diff
@@ -548,23 +584,42 @@ def _check_special(name, xla_shape, buf):
...
                       output_buffer_counts: Optional[Sequence[int]],
-                      result_handlers, kept_var_idx, *args):
+                      result_handlers,
+                      effects: List[core.Effect],
+                      kept_var_idx, *args):
   device, = compiled.local_devices()
   args = input_handler(args) if input_handler else args
   input_bufs_flat = flatten(device_put(x, device) for i, x in enumerate(args)
                             if i in kept_var_idx)
+  if effects:
```
mattjj, line 594: Think we could abstract this into a function? Maybe something like:

```python
input_bufs_flat, output_buffer_counts, token_handler = add_tokens(input_bufs_flat, effects)
... execute and unflatten here ...
out_bufs = token_handler(out_bufs)
return tuple(...)
```

mattjj: Actually, we could compute the update to `output_buffer_counts` earlier, in the place where we `partial(_execute_compiled, ...)`, so it need not be an input and output here IIUC.

jax/interpreters/mlir.py
```diff
@@ -674,10 +685,10 @@ def aval_to_types(aval):
 
   input_types = map(aval_to_types, jaxpr.in_avals)
   output_types = map(aval_to_types, jaxpr.out_avals)
+  num_tokens = len(effects)
```
mattjj, line 688: Does one effect always lead to one token?

mattjj (approved):

mattjj:
Awesome PR message!

jax/interpreters/mlir.py
sharadmv: Yes.

jax/_src/dispatch.py
sharadmv: Done!

jax/_src/dispatch.py
```diff
@@ -544,27 +580,48 @@ def _check_special(name, xla_shape, buf):
...
```
hawkinsp, line 605: A slight catch is this won't do great things for performance (e.g., building a new closure), but it probably doesn't matter a whole lot since effects aren't yet common. And if we did want them to be fast we should move this logic to C++.

jax/interpreters/mlir.py
```diff
@@ -650,6 +660,7 @@ def lower_jaxpr_to_fun(
     jaxpr: the jaxpr to lower.
     effects: a sequence of `core.Effect`s corresponding to an ordering of tokens
       that will be created in or used by the lowered function.
+    create_tokens: if true, the MHLO will create tokens and ignore dummy input tokens.
```
hawkinsp, line 663: I wouldn't say "ignore", but rather something like "the computation will not expect input tokens nor emit output tokens for effects".

Implicitly, is there a TODO to remove this?

hawkinsp (approved):

jax/_src/dispatch.py
```diff
@@ -100,6 +102,37 @@ def apply_primitive(prim, *args, **params):
...
+class RuntimeTokenSet(threading.local):
+  tokens: Dict[core.Effect, Tuple[RuntimeToken, Device]]
+
+  def __init__(self):
+    self.tokens = {}
+
+  def get_token(self, eff: core.Effect, device: Device) -> RuntimeToken:
+    if eff not in self.tokens:
+      self.tokens[eff] = device_put(np.zeros(0, np.bool_), device), device
+    elif self.tokens[eff][1] != device:
```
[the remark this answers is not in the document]
sharadmv: Won't this result in a per-device ordering? Here we want a single global ordering so we need to make sure there's an explicit dependency between computations across devices.

jax/_src/dispatch.py
sharadmv: Yep I left a TODO to enable this in the fast path.

jax/interpreters/mlir.py
sharadmv: I don't think so. Maybe this boolean should be renamed, but its purpose is to distinguish between the two possible lowerings of an effectful jaxpr:
1. It's a top-level jaxpr, meaning we need to create MHLO tokens and return the dummy tokens as output
2. It's a jaxpr in a higher-order-primitive, meaning we are directly passed MHLO tokens and just return them directly.

hawkinsp: Even for (1), isn't the intent that you pass runtime tokens eventually? (Absent some change to the runtime like another mechanism for ordering.)

Issue #10311, opened 2022-04-15
Generic Protocol for transformed function


YouJiacheng:
With `ParamSpec` we can completely annotate `jit` signature.
Thus type checking can be performed with transformed function. (and `lower`)
```python
T = TypeVar("T")
P = ParamSpec("P")
def jit(fun: Callable[P, T], ...) -> stages.Wrapped[P, T]:
  ...

# stages.py
T_co = TypeVar("T_co", covariant=True)
P_contra = ParamSpec("P_contra", contravariant=True)
class Wrapped(Protocol[P_contra, T_co]):
  def __call__(self, *args: P_contra.args, **kwargs: P_contra.kwargs) -> T_co:
    ...

  def lower(self, *args: P_contra.args, **kwargs: P_contra.kwargs) -> Lowered:
    ...
```
Just like #9999 , and it works almost perfectly with Visual Studio Code + pylance, signature(including parameters name) of function is fully preserved.
However, `mypy` currently doesn't handle `partial` + function-transform correctly. https://github.com/python/mypy/issues/12593
We can wait `mypy` fix this bug, or make some workaround. 
Some workaround option:
1. make `jit` a polymorphism decorator + decorator factory.
toy example
```python
from typing import TypeVar, Callable, overload
from typing_extensions import ParamSpec, Protocol, Literal

T_co = TypeVar('T_co', covariant=True)
T = TypeVar('T')
P = ParamSpec('P')

class Wrapped(Protocol[P, T_co]):
  def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T_co:
    pass

@overload
def jit(f: Callable[P, T], *, a: bool) -> Wrapped[P, T]:
    ...

@overload
def jit(f: Literal[None] = None, *, a: bool) -> Callable[[Callable[P, T]], Wrapped[P, T]]:
    ...

def jit(f=None, *, a):
    if f is None:
        return lambda f: jit(f, a=a)
    # original code here
    return f

@jit(a=True)
def f(x: int, y: int):
    return x + y

f(0, 1) #mypy ok
f(0, 'a') # mypy error  Argument 2 to "__call__" of "Wrapped" has incompatible type "str"; expected "int"
```
This break "Our general philosophy is to make the JAX core API as simple and explicit as possible" https://github.com/google/jax/issues/184#issuecomment-582029342
I personally like this option, but it break the philosophy of JAX.
2. add a dedicated decorator factory function and use it internally
toy example
```python
from typing import TypeVar, Callable
[...]

PR #10321, opened 2022-04-16
Attempt to add Generic Wrapped

YouJiacheng:
#10311 Option 2


YouJiacheng:
Sorry for exercising CI 10 time, but I finally get it works!

YouJiacheng:
@froystig WDYT?

froystig (commented):
`partial(jit, ...)` is a very common pattern, not only for us but also in other codebases. Is there a way to make this option work together with `partial(jit, ...)`?

YouJiacheng:
I think this is current version `mypy`'s limitation. I haven't used `pytype`(could you check it with `pytype`?), but I think `pyright` can work well with `partial(jit, ...)`.

PR #10322, opened 2022-04-16
add annotation for ndarray.at and other functions.

YouJiacheng:
see discussion in #10165


YouJiacheng:
Maybe we can remove the underscore of `_IndexUpdateHelper` and `_IndexUpdateRef`?
And its docs need to update.

YouJiacheng:
Nice semantic highlight&auto-completion without manually type annotation
image

jax/_src/numpy/lax_numpy.py
```diff
@@ -1593,6 +1593,10 @@ def stack(arrays, axis: int = 0, out=None):
       new_arrays.append(expand_dims(a, axis))
     return concatenate(new_arrays, axis=axis)
 
+@overload
+def tile(A: ndarray, reps: Union[Iterable[int], int]) -> ndarray: ...
+@overload
+def tile(A: Any, reps: Any) -> Any: ...
```
jakevdp, line 1599: What's the benefit of having a strictly-typed version of this overloaded by a declaration that has no information?

Given the issues mypy has with correctly interpreting annotations of JAX code (e.g. #10321), I'm hesitant to add the maintenance burden of these types multiply-overloaded function annotations, particularly if this is meant to be exploratory work toward the goal of doing this treatment to the full jax.numpy API.

These overloadings look to me like a lot of untested/poorly-tested code that will easily become out of sync with the main implementations, and whose benefits may not outweight the costs of maintenance. For example, the `type: ignore` statements required in this PR make clear the extent of the false positives introduced by these overloadings.

Given that, I'd lean toward not including these annotations in JAX. What do you think?

jakevdp: Maybe another more concrete question, since I've not used typing extensively before: is there any best practice for comprehensively unit-testing these sorts of type annotations? Every other line of code in this file is covered by extensive unit tests in `lax_numpy_test.py`. Is there any way to provide that same level of test coverage for these complicated overloaded annotations, so that, for example, if the return type of the implementation is changed, we are alerted by the test suite that the annotations must be changed to match?

YouJiacheng: I have to explain, the `type: ignore` required in this PR has nothing to do with these overloading. IIUC, the reason that original code doesn't require `type: ignore` is that `mypy` won't check type inside a function without any annotation, i.e. the problem have existed before this PR, but `mypy` doesn't check this function.(`array`)

YouJiacheng: Actually it isn't a real "multiple overload", it actually a return type annotation which  must be satisfied when argument type meet some condition, and a catch-all default for corner case.
IIUC, mypy will try to match overloads in their order of declaration. So if user call `tile` with `ndarray`, `mypy` should infer that the return type must be `ndarray`.
Clearly, type annotation in this manner doesn't need a test, it is a part of API interface definition and constraints on implementation. Nobody will expect that calling `tile` on a `ndarray` will get an object isn't an instance of `ndarray`.
If you want to test if the implementation meet this constraint, then this is what type checking do. (But it seems that `mypy` won't check overloads, I will take a look.)

jakevdp: Is there something we could change about `ndarray` that would make it so we wouldn't have to overload an annotation that returns `Any`? What will break if we leave out the version with `Any` as a return type?

YouJiacheng: Ah, I personally think that `tile` will always return `ndarray`. But I am not sure, thus I add an `Any` version.

YouJiacheng: Annotate overload while leave implementation untyped is a common pattern to intentionally let type checker not to check the implementation
> `mypy`: By default the bodies of functions without annotations are not type checked.

another way to do so is providing separate stub files, which used by `pytorch` for c/cpp implemented functions, but can also be applied to python implemented functions.

jakevdp: Well, to be precise, `tile` will return whatever `broadcast_to` returns. And because of [this line](https://github.com/google/jax/blob/a48752a5786e5debda219cfaccb8aa694266e6d5/jax/_src/numpy/util.py#L343-L344) `broadcast_to` can return anything so long as the input object has a `broadcast_to` method. So I think the best we can do for `tile` is to annotate it as returning `Any`.

jakevdp: Of course, this may all change, and it might not be clear to the person changing the behavior of `util._broadcast_to` that the changes to that function require synchronous changes in the overloaded type annotations of `jnp.tile`; thus my question about unit testing.

YouJiacheng: Good catch. The `Any` argument `tile` return type should be `Any`.
If you want to prevent any future error, I suggest we can always have a catch-all fallback to `Any`.
For "What test would have told us that our annotations are now wrong? Or would we not have caught that until we released the package and gotten bug reports from users?". IMO, type annotation in python can be viewed as documentation, and doesn't have any runtime effect. Thus, if there is something wrong with type annotation, it won't block any user, but only make them confused, which is the same as the situation when there is something wrong with documentation. Do you think we should have unit-testing for documentation?

jakevdp: > Do you think we should have unit-testing for documentation?

We do run doctests on all code samples, but it's true that not all docs can be tested. The question is: does the benefit of having untested docs outweight the burden of maintaining those docs? Arguably the answer is yes.

Back to type annotations: does the benefit of having incomplete type annotations with a fundamentally flawed type checking system (`mypy`) outweight the burden of maintaining those type annotations? It's not clear to me that the answer is yes, and particularly the more complicated those type annotations must be (protocols, overloads, etc.) I think the weight goew toward "no".

YouJiacheng: Actually there are type annotations in documentation, though they are docstring from numpy.
image

And again, if we only annotate the trivial case and always let others fallback to `Any`, I don't think we will really have the burden of maintaining those type annotation: Given that `jax.numpy` will be compatible(to some extent) with `numpy`, I can't imagine that the return-type is not `ndarray` when arg-type is.

jakevdp: OK, so the idea now is that because return types are not well tested, and refactoring such as `broadcast_to` may change things, we need every function we annotate to have multiple overloads, one being the common case of returning an ndarray, and the other being a generic where every input and output is annotated as `Any`. Is that correct?

If so, I really don't like that. If the best we can do is say is that our functions accept `Any` and return `Any`, why not simplify things and avoid any type annotations at all?

Alternatively, this may suggest that the real goal is not static type checking (having a generic `Any` fallback for all functions can never catch errors in your code), but rather documentation. If that's the case, why not continue using `Array = Any`, which serves the documentation purpose very well without having to add gratuitous overloads throughout the codebase?

YouJiacheng: Ah, If I replace return-type `Any` with `object`(I misunderstanding `Any` before), it can catch error in some case.
```python
from typing import Any, overload

@overload
def f(x: int) -> int:...
@overload
def f(x: Any) -> object:...
def f(x):
    if type(x) is int:
        return x + 1
    else:
        return 'a'

def g(x: int) -> int:
    return x + 1

def h(x: Any):
    return x

g(f(1)) # pass
h(f(None)) # pass
g(f(None)) # mypy report error: Argument 1 to "g" has incompatible type "object"; expected "int"  [arg-type]
```

YouJiacheng: In addition, type annotation can help IDE to do auto-completion and semantic highlighting. If the return-type is annotated with `ndarray`, then IDE can do auto-completion for properties and methods of return value.
```python
jnp.tile(x, 2).reshape(...)
# I only need to type "re", and don't need to worry about typo
# if there is a typo, then semantic highlighting won't highlight the wrong method name
# thus I can fix it without run the code
```

YouJiacheng: I think the type annotation will be useful if I replace `Any` with `object`! WDYT?

jakevdp: Thanks for clarifying - so it seems your main motivation here is not necessarily static type checking or documentation, but rather giving auto-complete tools more information.

I'll bring it back to my previous question: does the benefit of this outweigh the maintenance burden? Testing is still important (potentially incorrect autocomplete is worse than no autocomplete, in my opinion). What IDE tooling is required to take advantage of this? Can we get any sense of how much of our user-base is likely to benefit? And are there downsides to annotations motivated by autocomplete? e.g. if we tune our annotations to that application, does it have any negative impact on the static type checking use-case?

I think these issues need to be explored before we start adding significant numbers of non-trivial type annotations to the codebase.

YouJiacheng: Not only auto-complete, but also semantic highlight, which works similarly to type checker when `Any` replaced by `object`. i.e. if I expected the return typed is `ndarray`, but it fallbacks to `object`(`Any` works the same as `object` in semantic highlighting), I can immediately find out this error.
Again, I can replace `Any` with `object`, so that it can help static type checking as well.

YouJiacheng:
After replace `Any` with `object`, the intention is clear: we provide return-type guarantee with `ndarray` arguments(which may need some modifications to ensure user can not inherit `ndarray` and exploit the special dispatch.), while leave others case typed `object`(NOT `Any`)

YouJiacheng:
But the downside of `object` is that if user use type checker and write something like `f(x) + 1`, where `x` is not a `ndarray`, then type checker will complain that `object` not support `+`. BUT I think this is intentioned, for example `PRNGKeyArray` doesn't support arithmetic. If I add `np.array` annotation, the only false postive is `__jax_array__`. (not sure it really works with `tile`)

jakevdp:
So I'm still kind of hesitant on this. It seems like it's opening the door to add a lot of extra code (two overloads for every user-facing API function) with very little testing. I know in theory these annotations should be covered by running mypy on our source files, and/or perhaps our test suite, but my main experience with mypy in JAX is having to add `# type: ignore` in seemingly random places across the package in order to fix new false positives that arise  every time we update the mypy version used by our CI. It doesn't give me much confidence in the test coverage provided by mypy's static type checking.

So what this PR looks like to me is a proposal to eventually add hundreds of lines of untested type annotations to the API in a manner that makes it very likely the annotations will silently become out-of-date at some point in the future. It's not obvious to me that the benefits here outweigh the increased maintenance burden.

YouJiacheng:
Okay, I can remove the "overload for type annotation" part in this PR because of the maintenance concern. What do you think about the rest of this PR?
Nevertheless, Pytorch heavily uses stubs(`pyi`) for type annotation and has won lots of endorsement from users, which is almost the same as I do. The only difference is that they place the type annotation in a separate file.

jakevdp:
Hi - thanks so much for sticking with a tough review process here – your comments & contribution here have been really helpful for me to learn more about some of the potential benefits of type annotations.

We've been having some offline conversation among the core team spurred by this, by #10321, and by other similar requests/contributions in the past. I think before we make any decision on this kind of thing we need to step back and do a more comprehensive look at the benefits & tradeoffs surrounding type annotations. With that in mind I think it would be best to table this contribution for now, until we can come up with a better roadmap for annotations within JAX as a whole.

How does that sound?

YouJiacheng:
Thanks for your nice review&insight. I'd like to wait for a comprehensive roadmap.
Nevertheless, I'd like to create a new PR, which only adds annotation for `ndarray.at` while advoid cyclic import. Do you think such PR can be merged without waiting for the roadmap?

jakevdp:
I would hold off on that – I think it will be easier to review once we've thought more about the broader question.

YouJiacheng:
Hi - friendly ping. Is there anything new about the roadmap?

jakevdp:
I think this is all done now via work related to https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html. Thanks for getting the conversation started here!

PR #10364, opened 2022-04-19
Add quotes to install commands in README

dbisk:
The commands in the README to install wheels compatible with specific versions of Cuda/cudnn are missing required quotation marks around the group name. This small fix adds those required quotation marks.


jakevdp (approved):
Thanks for catching this!  I think the unquoted version works on some shells, but I don't know of any shell where quotes *don't* work!

PR #10367, opened 2022-04-19
relax requirements on pytrees in custom_jvp

### LINKED ISSUE #10366: `tree_flatten` gets `None` when used in `defvjp`
Please:

I wanted to define a `custom_vjp` for a function that takes a string to indicate what operations are to be conducted. Since `str` is not a validate type for JAX whereas any objects that can be flatten and unflatten as a `pytree` is, I thought of this hacky way to define my own string object: 

```python
import jax
import jax.numpy as jnp
from functools import partial

OP_TO_IDX, IDX_TO_OP = {"add": 0, "mul": 1}, {0: "add", 1: "mul"}

@jax.tree_util.register_pytree_node_class
class OpStr(object):
    def __init__(self, op):
        assert op in OP_TO_IDX, "can only be one of the ops"
        self.op = op

    def __eq__(self, other):
        return self.op == other

    def __hash__(self):
        return hash(self.op)

    def tree_flatten(self):
        return ([OP_TO_IDX[self.op]]), None

    @classmethod
    def tree_unflatten(cls, aux_data, children):
        idx = int(children[0])
        return cls(op=IDX_TO_OP[idx])

@partial(jax.custom_vjp, nondiff_argnums=(2,))
def op(x, y, op):
    if op == "add": z = x + y
    if op == "mul": z = x * y
    return z

def op_fwd(x, y, op):
    cache = (x, y, op)
    if op == "add": z = x + y
    if op == "mul": z = x * y
    return z, cache

def op_bwd(cache, dz):
    x, y, op = cache
    if op == "add": dz_dx = dz_dy = 1.0
    if op == "mul": dz_dx, dz_dy = y, x
    return dz_dx, dz_dy

op.defvjp(op_fwd, op_bwd)
grad_op = jax.grad(op)
grad_op(1.0, 1.0, OpStr("add"))
```

But this gives me 
```
  File "/Users/wangy1/Documents/GitHub/dgl/tests/jax/test_op_str.py", line 47, in <module>
    grad_op(1.0, 1.0, OpStr("add"))
  File "/Users/wangy1/Documents/GitHub/dgl/tests/jax/test_op_str.py", line 24, in tree_unflatten
    idx = int(children[0])
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Which is not expected since the flattened tree shouldn't be `None`.

mattjj:
fixes #10366

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -22,8 +22,9 @@ from typing import (Callable, Generic, Optional, Sequence, Tuple, List, TypeVar,
 from jax import core
 from jax import linear_util as lu
 from jax.custom_transpose import custom_transpose
-from jax.tree_util import (tree_flatten, tree_unflatten, tree_map, treedef_is_leaf,
-                           treedef_tuple, register_pytree_node_class)
+from jax.tree_util import (tree_flatten, tree_unflatten, tree_map,
+                           treedef_is_leaf, treedef_tuple,
+                           register_pytree_node_class, tree_leaves)
 from jax._src import custom_api_util
 from jax._src import dtypes
 from jax._src.util import cache, safe_zip, safe_map, split_list, Unhashable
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Nice!

YouJiacheng:
What about add a dedicated `tree_apply` for `Callable[..., None]`? Or more precisely, `tree_preorder_apply`. IIUC, the terminology "map" doesn't guarantee the execution order. (And usually doesn''t allow side-effect).

mattjj:
True, but JAX internals are allowed to rely on JAX-internal implementation details, like the execution order ;)

jax/_src/custom_derivatives.py
```diff
@@ -564,21 +565,17 @@ def __call__(self, *args: Any, **kwargs: Any) -> ReturnValue:  # pytype: disable
...
-           "only be used for arguments that can't be or contain JAX tracers, "
-           "e.g. function-valued arguments. In particular, array-valued "
-           "arguments should typically not be indicated as nondiff_argnums. "
-           "\n\n"
-           "This behavior recently changed in JAX. "
-           "See https://github.com/google/jax/blob/main/docs/custom_vjp_update.md "
-           "for more information.")
-    raise UnexpectedTracerError(msg)
+  for leaf in tree_leaves(x):
+    if isinstance(x, core.Tracer):
```
froystig, line 570:
```suggestion
    if isinstance(leaf, core.Tracer):
```
Did we mean to write this instead here?

mattjj: Yes!

froystig: fixed in https://github.com/google/jax/commit/c8a7d5990daecdbf3ff5a25d67ad4979a55e2182

PR #10371, opened 2022-04-19
Adds MLIR token lowering for effectful jaxprs

sharadmv:
This PR enables lowering effectful jaxprs in a limited capacity. Specifically, the effects in the jaxpr have to be "lowerable" by adding them to the `mlir.lowerable_effects` allowlist. 

For effects that are in `core.ordered_effects`, we enable threading MLIR tokens to sequence the effects. We do this by creating an `mlir.TokenSet` that will be passed in and out of lowering rules.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -3390,8 +3390,10 @@ def _reduce_lower(ctx, *values, computation, jaxpr, consts, dimensions):
   reducer = op.regions[0].blocks.append(*(ir_types + ir_types))
   with ir.InsertionPoint(reducer):
     reducer_ctx = ctx.module_context.replace(name_stack=util.new_name_stack())
-    out_nodes = mlir.jaxpr_subcomp(reducer_ctx, jaxpr, consts,
-                                   *([a] for a in reducer.arguments))
+    if jaxpr.effects:
+      raise NotImplementedError('Cannot lower effectful `reduce`.')
+    out_nodes, _ = mlir.jaxpr_subcomp(reducer_ctx, jaxpr, mlir.TokenSet(), consts,
+                                      *([a] for a in reducer.arguments))
     mhlo.ReturnOp(util.flatten(out_nodes))
   return op.results
 
@@ -3782,7 +3784,9 @@ def _sort_lower(ctx, *operands, dimension, is_stable, num_keys):
         module_context = ctx.module_context,
         primitive=None,
         avals_in=util.flatten(zip(scalar_avals, scalar_avals)),
-        avals_out=[core.ShapedArray((), np.bool_)])
+        avals_out=[core.ShapedArray((), np.bool_)],
+        tokens_in=ctx.tokens_in,
+        tokens_out=ctx.tokens_out)
 
     out = lower_comparator(sub_ctx, *[[a] for a in comparator.arguments],
                            num_keys=num_keys)
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -14,6 +14,7 @@
 
 # Lowering and execution path that converts jaxprs into the MLIR MHLO/CHLO
 # dialects.
+from __future__ import annotations
 
 import collections
 import dataclasses
@@ -53,6 +54,8 @@ T = typing.TypeVar("T")
 
 # mypy implicitly sets this variable to true when type checking.
 MYPY = False
+
+lowerable_effects: Set[core.Effect] = set()
 
 
 # IR Helpers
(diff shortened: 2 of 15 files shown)


jax/interpreters/mlir.py
```diff
@@ -510,8 +519,15 @@ def lower_jaxpr_to_module(
     ctx.module.operation.attributes["sym_name"] = ir.StringAttr.get(
         f"{module_name}.{next(_module_unique_id)}")
     # TODO(phawkins): represent units with zero buffers at the runtime level.
+    unlowerable_effects = set(eff for eff in
+        jaxpr.effects if eff not in lowerable_effects)
```
mattjj, line 523:
```suggestion
    unlowerable_effects = {eff for eff in jaxpr.effects
                           if eff not in lowerable_effects)
```

i.e. use set comprehension, and indent.

jax/interpreters/mlir.py
```diff
@@ -547,11 +563,55 @@ def _set_up_aliases(avals_in, avals_out, donated_args):
...
+def token_type() -> Sequence[ir.Type]:
+  return [mhlo.TokenType.get()]
+
+def token() -> Token:
+  return wrap_singleton_ir_values(
+      mhlo.CreateTokenOp(mhlo.TokenType.get()).result)
+
+Token = Sequence[ir.Value]
+
+class TokenSet(collections.OrderedDict):
```
mattjj: Maybe delegate instead of inherit (mentioned in chat).

jax/_src/lax/lax.py
```diff
@@ -3390,8 +3390,8 @@ def _reduce_lower(ctx, *values, computation, jaxpr, consts, dimensions):
   reducer = op.regions[0].blocks.append(*(ir_types + ir_types))
   with ir.InsertionPoint(reducer):
     reducer_ctx = ctx.module_context.replace(name_stack=util.new_name_stack())
-    out_nodes = mlir.jaxpr_subcomp(reducer_ctx, jaxpr, consts,
-                                   *([a] for a in reducer.arguments))
+    out_nodes, _ = mlir.jaxpr_subcomp(reducer_ctx, jaxpr, mlir.TokenSet(), consts,
```
mattjj, line 3393: Should we also be raising a NotImplementedError in this case if the jaxpr has effects? (Basically, whenever we have an underscore dropping the token coming out of `mlir.jaxpr_subcomp`, it seems like we need a corresponding error if the jaxpr is effectful.)

mattjj (approved):
Nice!

jax/_src/lax/lax.py
mattjj: Maybe we could even assert that the token going in is same object as the token coming out. That could be an interesting check!

sharadmv: I'll add in the `NotImplementedError`s.

There probably won't be a token going in because it will be filtered at an earlier call site.

gnecula:
I'd like to ask to rebase this PR into fewer (or just 1) commit, for a cleaner history

sharadmv:
Yes, I'll make sure to squash before I submit. I'm doing this so it's easier for me to rebase other branches on top of this one.

tests/jaxpr_effects_test.py
```diff
@@ -175,14 +205,153 @@ def f(x):
...
+
+    @jax.jit
+    def f(x):
+      effect_p.bind(effect='foo')
+      return x + 1.
+    with self.assertRaisesRegex(ValueError, 'Lowering rule for `effect` returns '
+        'incorrect set of output token.'):
+      f.lower(2.)
+
+  def test_lowering_ordered_effect_should_create_tokens(self):
```
hawkinsp, line 265: Another option for testing the MHLO that is generated would be to use a FileCheck test. I am not saying you must do this but it may be preferable and perhaps more readable.

See
https://llvm.org/docs/CommandGuide/FileCheck.html
and some of the existing tests for other properties:
https://github.com/google/jax/tree/main/tests/filecheck

(These tests aren't routinely tested in our OSS CI, though. We do test them internally in CI, mostly because they aren't things you can easily hook up to pytest but one can easily hook up to Bazel.)

The advantage of doing this is that you don't have to roll your own pattern matching code. You may or may not prefer that, depending on the pattern being matched.

hawkinsp (approved):

tests/jaxpr_effects_test.py
sharadmv: Thanks for the pointer, FileCheck seems super useful. I think the current tests are simple enough (patterns are not very advanced) but I'll work on some more advanced tests using filecheck as well. These I'll submit internally since it's easier to test that way as you mentioned.

PR #10380, opened 2022-04-20
simplify promote_types

YouJiacheng:


mattjj:
Thanks for looking at this! I think it was written this way intentionally, to use identity checks instead of equality checks (because `in` uses equality checking). I can't remember why though...

jakevdp:
Yes, Matt's memory is correct: we want this to use identity rather than equality, because `numpy.dtype` overloads `__eq__` in ways that cause this code to break.

jakevdp:
To be more specific, this is the difference:
```python
In [7]: import numpy as np                                                                            

In [8]: dtype = np.dtype('float64')                                                                   

In [9]: dtype in [int, float, complex]                                                                
Out[9]: True

In [10]: any(dtype is t for t in [int, float, complex])                                               
Out[10]: False
```

YouJiacheng:
However, I think we can use `set` to perform `__contains__` check, which use hash equality check, WDYT?

jakevdp:
Yes, `set` would be an option, except membership checks require the argument to be hashable, and fail otherwise, so that's not an option here.

jakevdp:
```python
In [14]: np.arange(5) in [int, float, complex]                                                        
<ipython-input-14-0e4eaa6dc6c0> in <module>
----> 1 np.arange(5) in [int, float, complex]

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

YouJiacheng:
What about use a set of `id` and use `id(a) in id_set`? This is more explicit than `any(a is t for t in ...)` in my opinion.

jakevdp:
That seems less explicit and more brittle to me.

mattjj:
I kinda like the way it is written now, but IIUC @YouJiacheng may be observing that using the `in` syntax is extra clear for testing membership.

I'm not necessarily advocating for this but we could make a `jax.util.SetById` like

```python
class SetById:
  _elts: Set
  def __init__(self, elts):
    self._elts = elts
  def __contains__(self, elt):
    return any(elt is e for e in self._elts)
```

(We could also back it by a set of object id integers if we wanted.)

I'm not sure but I think this pattern has come up a couple times (needing to test membership by id), so it could make sense to have a utility class like this.

All that said, by default I probably wouldn't change anything unless/until we notice reusing this kind of idiom in a few different places...

YouJiacheng:
FYI, the performance difference is negligible
```python
import numpy as np
_weak_types = [int, float, complex]
id_set = set(id(t) for t in _weak_types)
from timeit import timeit
print(timeit('any(np.float32 is t for t in _weak_types)', globals=globals(), number=1000000)) # 0.480
print(timeit('id(np.float32) in id_set', globals=globals(), number=1000000)) # 0.126
print(timeit('any(int is t for t in _weak_types)', globals=globals(), number=1000000)) # 0.375
print(timeit('id(int) in id_set', globals=globals(), number=1000000)) # 0.116
```
only 0.2~0.3 μs. The explicity is more important.

mattjj:
Or we could have a utility function, like `contained_by_id(e: Any, s Sequence) -> bool`.

PR #10381, opened 2022-04-20
remove numpy.linalg._promote_arg_dtypes

YouJiacheng:
in favor of numpy.util._promote_dtypes_inexact

### DIFF
--- a/jax/_src/numpy/linalg.py
+++ b/jax/_src/numpy/linalg.py
@@ -23,35 +23,20 @@ from typing import Tuple, Union, cast
 from jax import jit, custom_jvp
 from jax import lax
 
-from jax._src import dtypes
 from jax._src.lax import lax as lax_internal
 from jax._src.lax import linalg as lax_linalg
 from jax._src.numpy import lax_numpy as jnp
-from jax._src.numpy.util import _wraps
+from jax._src.numpy.util import _wraps, _promote_dtypes_inexact
 from jax._src.util import canonicalize_axis
 
 _T = lambda x: jnp.swapaxes(x, -1, -2)
 _H = lambda x: jnp.conjugate(jnp.swapaxes(x, -1, -2))
 
 
-def _promote_arg_dtypes(*args):
-  """Promotes `args` to a common inexact type."""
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/linalg.py
```diff
@@ -23,43 +23,28 @@
...
-    return args[0]
-  else:
-    return args
-
-
 @_wraps(np.linalg.cholesky)
 @jit
 def cholesky(a):
-  a = _promote_arg_dtypes(jnp.asarray(a))
+  a = _promote_dtypes_inexact(jnp.asarray(a))[0]
```
jakevdp, line 39: I have a slight preference here for writing `a, = _promote_dtypes_inexact(jnp.asarray(a))` because it includes an implicit assertion about the length of the returned tuple.

jakevdp (changes requested):
Looks good, just a minor comment

jax/_src/numpy/linalg.py
YouJiacheng: Ah, I have seen that JAX use `a, =` in other file, but I have known the reason before. I use `...[0]` because `a, =` make `=` not align. But consider the reason your provide I think I should use `a, =`

jakevdp (approved):
Looks good, thanks!

jax/_src/scipy/linalg.py
```diff
@@ -63,7 +63,7 @@ def cho_solve(c_and_lower, b, overwrite_b=False, check_finite=True):
 
 @partial(jit, static_argnames=('full_matrices', 'compute_uv'))
 def _svd(a, *, full_matrices, compute_uv):
-  a = np_linalg._promote_arg_dtypes(jnp.asarray(a))
+  a = _promote_dtypes_inexact(jnp.asarray(a))
```
soraros, line 66: I think this is wrong. Curious how come it's not caught by the test suite.

jakevdp: Thanks for checking on this - is that something you'd be willing to add?

YouJiacheng: Fixed in #10387

Issue #2891, opened 2020-04-29
Improve performance of DeviceArray indexing


skye:
`DeviceArray.__getitem__` is implemented by [_rewriting_take](https://github.com/google/jax/blob/master/jax/numpy/lax_numpy.py#L3184) in lax_numpy.py. This function can be sped up via `jit`, but it's tricky to avoid excessive recompilations. @hawkinsp has laid the foundation for this with the `_split_index_for_jit` logic and the commented-out `jit` annotation on `_gather`, but it still needs some more work as described in the TODO and in https://github.com/google/jax/pull/2865. In the thread in #2865, Peter says:

> The slightly better idea sketched out here already is to split the indices into a static part and a dynamic part (_split_index_for_jit). We then call a helper function (_gather) that is jit-ted with only the static part being a static_argnum. Unfortunately I had to comment out the jit because of the pattern shown in the comment led to a lot of recompilation because it was considered static.
> 
> My plan was to work around that case by recognizing that pattern outside the call to _gather and removing the specific indices from the static part of the index. Note this is a little subtle because of cases like np.arange(10)[5:15] mean the shape may depend on the index value, but note we only support concrete indices here anyway.
> 
> I suspect another blocker here is that there is, subsequent to the jit attempt above, Matt added a fast-path to generate a reshape in some common cases that don't actually do any slicing. We must not break that because the lazy constant stuff depends on it working.

danielsuo:
Registering interest in this issue! We're using CPU backends for now, which are understandably not a priority, but if `jax` really is a "drop-in" for `numpy`, it'd be nice to have reasonable indexing performance.

For reference, here are some timing comparisons ([colab](https://colab.sandbox.google.com/drive/1VcBUN_DLxrEHiWcf6r0wA2AXw-WYdKZv#scrollTo=VCZw1xbOJXJM)).

ProfFan:
+1 for this issue. It's really a showstopper for first-time users like myself, where JAX shows 100x slower performance than numpy...

Issue #2878, opened 2020-04-29
_index_to_gather produces DeviceArrays for int / onp.ndarray inputs


mattjj:
The `_index_to_gather` function and some subfunctions, like `_normalize_index`, perform operations with `lax` functions even when dealing with integer or onp.ndarray inputs. That's unfortunate because it means, for example, doing scalar integer operations on the default accelerator device! We should probably revise.

mattjj:
Alternatively, we could re-enable `jit` as per [this TODO](https://github.com/google/jax/blob/790d92965c881e0316876242953009ca6fc8f3eb/jax/numpy/lax_numpy.py#L3192-L3194). Of course, I completely forget why it as hard... can we just make slices into pytrees?

skye:
https://github.com/google/jax/issues/2891 attempts to summarize why re-enabling the `jit` is hard (by quoting Peter).

mattjj:
Yeah, though I realized this issue is somewhat orthogonal because we likely want to keep static index munging in Python / onp anyway.

mattjj:
I think this is stale!

Issue #7426, opened 2021-07-31
issue composing jax.checkpoint with jax.jit on function with (concrete) bool parameter


deasmhumhna:
This last line presents a workaround but I encountered this error originally using `flax.linen` where the user doesn't have transparent access to the (sub)function that is being checkpointed i.e. when using a lifted transformation on a submodule constructed in a `__call__` method with the `compact` decorator. The error happens while trying to construct a `ConcretizationTypeError`. 

```python
import jax
from jax import numpy as jnp

from functools import partial

def bool_fn(x, flag: Optional[bool]=None):
    if flag:
        return x
    return x - 1

jax.jit(lambda x: bool_fn(x, flag=False))(jnp.ones((10,), jnp.float32)) # succeeds 
(lambda x: jax.checkpoint(bool_fn)(x, flag=False))(jnp.ones((10,), jnp.float32)) # succeeds
jax.jit(lambda x: jax.checkpoint(bool_fn)(x, flag=False))(jnp.ones((10,), jnp.float32)) # fails
# jax.jit(lambda x: jax.checkpoint(bool_fn, concrete=True)(x, flag=False))(jnp.ones((10,), jnp.float32)) # also fails
# jax.jit(lambda x: jax.checkpoint(bool_fn)(x, flag=jnp.array(False)))(jnp.ones((10,), jnp.float32)) # also fails
jax.jit(jax.checkpoint(partial(bool_fn, flag=False)))(jnp.ones((10,), jnp.float32)) # succeeds
```

```
---------------------------------------------------------------------------
UnfilteredStackTrace                      Traceback (most recent call last)
<ipython-input-32-db5b88a05ea3> in <module>()
      7 (lambda x: jax.checkpoint(bool_fn)(x, flag=False))(jnp.ones((10,), jnp.float32)) # succeeds
----> 8 jax.jit(lambda x: jax.checkpoint(bool_fn)(x, flag=False))(jnp.ones((10,), jnp.float32)) # fails
      9 # jax.jit(lambda x: jax.checkpoint(bool_fn, concrete=True)(x, flag=False))(jnp.ones((10,), jnp.float32)) # fails

30 frames
/usr/local/lib/python3.7/dist-packages/jax/_src/traceback_util.py in reraise_with_filtered_traceback(*args, **kwargs)
    182     try:
--> 183       return fun(*args, **kwargs)
    184     except Exception as e:

[...]

PR #10393, opened 2022-04-20
Improve performance of array integer indexing

lgeiger:
This PR tries to improve performance of array integer indexing as a first step towards addressing #2878.

Since it seems like using `jit` as suggested in #2891 is difficult I've limited this PR to the case of indexing using a single Python integer outside of `jit` to keep the diff somewhat manageable. Some of the changes should also benefit other cases.

I split the PR into multiple commits that are best reviewed one by one. Also please let me know if these changes would prevent future planned optimisation that I am not aware of. If so I'm happy to close this, or split out some of the optimisation into separate PR's.

## Changes
- 09134b36bd0a3a6fc10609b8a7a28d93c69b2e93 prevents the conversion of Python integers to DeviceArrays and uses plain Python code in favour of `lax` operations for index normalisation when possible. This moves the scalar index computations to the CPU instead of running on an accelerator. I currently can't easily measure the performance impact of this but for me this seems to be preferred for statically known indices.
- 75a092e027b26bcecf832c8cbfe088e98ec8c08d prevents unnecessary calls to `lax.expand_dims`
- b8d70d31a5187c3c4cf72743b3f96a9026627904 removes explicit conversion using `asarray`
- cf8e8ed46062fce095ce76b001626a4d3c248a3f and 19ac51e10339c465f7c40ea3b9acb966320c07a0 remove expensive calls to `core.get_aval` to speedup checks

## Latency impact

I've measured indexing performance using a simple toy example on CPU:
```python
from jax import numpy as jnp

a = jnp.arange(4000)
a[5]
%timeit a[6].block_until_ready()
```

This PR improves performance by a factor of **4.8x**. Jax is still a lot slower than plain numpy but this already seems like a decent improvement.

| commit                                   |   time | reduction |
| ---------------------------------------- | -----: | --------: |
| 88ac4edf63625db10f92d5368a598205aac1deb4 (main baseline) | 468 µs |           |
| 09134b36bd0a3a6fc10609b8a7a28d93c69b2e93 | 225 µs | - 52 %    |
| 75a092e027b26bcecf832c8cbfe088e98ec8c08d | 168 µs | - 25 %    |
| b8d70d31a5187c3c4cf72743b3f96a9026627904 | 143 µs | - 15 %    |
| cf8e8ed46062fce095ce76b001626a4d3c248a3f | 125 µs | - 13 %    |
| 19ac51e10339c465f7c40ea3b9acb966320c07a0 |  98 µs | - 22 %    |

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3535,7 +3535,6 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
   # Computes arr[idx].
   # All supported cases of indexing can be implemented as an XLA gather,
   # followed by an optional reverse and broadcast_in_dim.
-  arr = asarray(arr)
 
   # TODO(mattjj,dougalm): expand dynamic shape indexing support
   if (jax.config.jax_dynamic_shapes and type(idx) is slice and idx.step is None
@@ -3555,7 +3554,7 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
 def _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
             unique_indices, mode, fill_value):
   idx = _merge_static_and_dynamic_indices(treedef, static_idx, dynamic_idx)
-  indexer = _index_to_gather(shape(arr), idx)  # shared with _scatter_update
+  indexer = _index_to_gather(arr.shape, idx)  # shared with _scatter_update
   y = arr
 
   if fill_value is not None:
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3417,21 +3417,22 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
 def _normalize_index(index, axis_size):
   """Normalizes an index value in the range [-N, N) to the range [0, N)."""
   try:
-    idx = core.concrete_or_error(operator.index, index, "normalize index")
-    ndim = core.concrete_or_error(operator.index, axis_size, "normalize axis_size")
-    return _canonicalize_axis(idx, ndim)
-  except (TypeError, ValueError, IndexError, core.ConcretizationTypeError):
-    if issubdtype(_dtype(index), np.unsignedinteger):
-      return index
-    if core.is_constant_dim(axis_size):
-      axis_size_val = _lax_const(index, axis_size)
-    else:
-      axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
-                                              _dtype(index))
-    return lax.select(
-      lax.lt(index, _lax_const(index, 0)),
-      lax.add(index, axis_size_val),
-      index)
+    idx, ndim = operator.index(index), operator.index(axis_size)
+  except TypeError:
+    pass
+  else:
+    return idx + ndim if idx < 0 else idx
+  if issubdtype(_dtype(index), np.unsignedinteger):
+    return index
+  if core.is_constant_dim(axis_size):
+    axis_size_val = _lax_const(index, axis_size)
+  else:
+    axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
+                                             _dtype(index))
+  return lax.select(
+    lax.lt(index, _lax_const(index, 0)),
+    lax.add(index, axis_size_val),
+    index)
 
 
 TAKE_ALONG_AXIS_DOC = """
@@ -3660,9 +3661,10 @@ def _is_basic_int_index(x):
             not aval.shape and issubdtype(aval.dtype, integer))
   try:
     operator.index(x)
-    return not isinstance(x, bool)
   except TypeError:
     return False
+  else:
+    return not isinstance(x, bool)
 
 def _index_to_gather(x_shape, idx, normalize_indices=True):
   # Remove ellipses and add trailing slice(None)s.
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3657,8 +3657,8 @@ def _merge_static_and_dynamic_indices(treedef, static_idx, dynamic_idx):
 def _is_basic_int_index(x):
   if isinstance(x, core.Tracer):
     aval = x.aval
-    return (isinstance(aval, (ConcreteArray, ShapedArray)) and
-            not aval.shape and issubdtype(aval.dtype, integer))
+    return (isinstance(aval, ShapedArray) and not aval.shape
+            and issubdtype(aval.dtype, integer))
   try:
     operator.index(x)
   except TypeError:
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3664,7 +3664,7 @@ def _is_basic_int_index(x):
   except TypeError:
     return False
   else:
-    return not isinstance(x, bool)
+    return not isinstance(x, (bool, np.bool_))
 
 def _index_to_gather(x_shape, idx, normalize_indices=True):
   # Remove ellipses and add trailing slice(None)s.
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3416,17 +3416,22 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
 
 def _normalize_index(index, axis_size):
   """Normalizes an index value in the range [-N, N) to the range [0, N)."""
-  if issubdtype(_dtype(index), np.unsignedinteger):
-    return index
-  if core.is_constant_dim(axis_size):
-    axis_size_val = _lax_const(index, axis_size)
-  else:
-    axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
-                                             _dtype(index))
-  return lax.select(
-    lax.lt(index, _lax_const(index, 0)),
-    lax.add(index, axis_size_val),
-    index)
+  try:
+    idx = core.concrete_or_error(operator.index, index, "normalize index")
+    ndim = core.concrete_or_error(operator.index, axis_size, "normalize axis_size")
+    return _canonicalize_axis(idx, ndim)
+  except (TypeError, ValueError, IndexError, core.ConcretizationTypeError):
+    if issubdtype(_dtype(index), np.unsignedinteger):
+      return index
+    if core.is_constant_dim(axis_size):
+      axis_size_val = _lax_const(index, axis_size)
+    else:
+      axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
+                                              _dtype(index))
+    return lax.select(
+      lax.lt(index, _lax_const(index, 0)),
+      lax.add(index, axis_size_val),
+      index)
 
 
 TAKE_ALONG_AXIS_DOC = """
@@ -3725,8 +3730,7 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
       ndim = len(shape)
 
       start_dim = len(gather_indices_shape)
-      gather_indices += ((lax.convert_element_type(a, index_dtype), start_dim)
-                         for a in advanced_indexes)
+      gather_indices += ((a, start_dim) for a in advanced_indexes)
       gather_indices_shape += shape
 
       start_index_map.extend(x_advanced_axes)
@@ -3751,7 +3755,6 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
         # XLA gives error when indexing into an axis of size 0
         raise IndexError(f"index is out of bounds for axis {x_axis} with size 0")
       i = _normalize_index(i, x_shape[x_axis]) if normalize_indices else i
-      i = lax.convert_element_type(i, index_dtype)
       gather_indices.append((i, len(gather_indices_shape)))
       collapsed_slice_dims.append(x_axis)
       gather_slice_shape.append(1)
@@ -3808,8 +3811,7 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
         if needs_rev:
           reversed_y_dims.append(collapsed_y_axis)
         if stride == 1:
-          i = lax.convert_element_type(start, index_dtype)
-          gather_indices.append((i, len(gather_indices_shape)))
+          gather_indices.append((start, len(gather_indices_shape)))
           slice_shape.append(limit - start)
           gather_slice_shape.append(limit - start)
           offset_dims.append(collapsed_y_axis)
@@ -3842,14 +3844,18 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
     gather_indices_array = np.zeros((0,), dtype=index_dtype)
   elif len(gather_indices) == 1:
     g, _ = gather_indices[0]
-    gather_indices_array = lax.expand_dims(g, (g.ndim,))
+    try:
+      gather_indices_array = np.array([operator.index(g)], dtype=index_dtype)
+    except TypeError:
+      gather_indices_array = lax.expand_dims(lax.convert_element_type(g, index_dtype), (g.ndim,))
   else:
     last_dim = len(gather_indices_shape)
     gather_indices_shape.append(1)
     gather_indices_array = lax.concatenate([
-      lax.broadcast_in_dim(g, gather_indices_shape, tuple(range(i, i + g.ndim)))
-      for g, i in gather_indices],
-      last_dim)
+      lax.broadcast_in_dim(
+        lax.convert_element_type(g, index_dtype), gather_indices_shape, tuple(range(i, i + _ndim(g))))
+        for g, i in gather_indices],
+    last_dim)
 
   dnums = lax.GatherDimensionNumbers(
     offset_dims = tuple(offset_dims),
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -3417,15 +3417,20 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
...
-                                             _dtype(index))
-  return lax.select(
-    lax.lt(index, _lax_const(index, 0)),
-    lax.add(index, axis_size_val),
-    index)
+  try:
+    idx = core.concrete_or_error(operator.index, index, "normalize index")
+    ndim = core.concrete_or_error(operator.index, axis_size, "normalize axis_size")
+    return _canonicalize_axis(idx, ndim)
+  except (TypeError, ValueError, IndexError, core.ConcretizationTypeError):
```
lgeiger, line 3424: Not sure if this is the best way to check for scalar Python indices but it seems quite readable and is fast.
- `operator.index` might throw a `TypeError `
- `_canonicalize_axis` might throw a `ValueError`
- `core.concrete_or_error` might throw a `ConcretizationTypeError `
-  During construction of a `ConcretizationTypeError` an `IndexError` might be thrown similar to what was reported in #7426

jakevdp:
Related: #10150

jax/_src/numpy/lax_numpy.py
```diff
@@ -3519,7 +3524,6 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
   # Computes arr[idx].
   # All supported cases of indexing can be implemented as an XLA gather,
   # followed by an optional reverse and broadcast_in_dim.
-  arr = asarray(arr)
```
lgeiger, line 3522: `_rewriting_take` is only called from within `__getitem__` so I think it is fine to remove here. Let me know if `asarray` is still needed for some other reason here.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3866,6 +3882,8 @@ def _eliminate_deprecated_list_indexing(idx):
   # non-tuple sequence containing slice objects, [Ellipses, or newaxis
   # objects]". Detects this and raises a TypeError.
   if not isinstance(idx, tuple):
+    if isinstance(idx, (int, slice)):
+      return (idx,)
```
lgeiger, line 3886: This is a tiny improvement to early exit in case of integer indices. `isinstance(idx, typing.Sequence)` is quite slow since it calls `__instancecheck__`. `isinstance(idx, collections.abc.Sequence)` would be faster in Python 3.9, but early exiting for the simple case shouldn't complicate things too much here.

jax/_src/numpy/lax_numpy.py
YouJiacheng: Use `canonicalize_axis` for normalize index may cause confusing error:
`ValueError(f"axis {axis} is out of bounds for array of dimension {num_dims}")`

YouJiacheng:
Just curious, does this PR affect performance after JIT?

lgeiger:
> Related: #10150

@jakevdp Thanks for mentioning that PR. I rebased and added a small test in 8054f426e13346d8c09022704761824651ef983c to ensure that only a gather operation is present when using constant integer indices.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3417,17 +3417,22 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
...
-    lax.lt(index, _lax_const(index, 0)),
-    lax.add(index, axis_size_val),
-    index)
+  try:
+    idx = core.concrete_or_error(operator.index, index, "normalize index")
+    ndim = core.concrete_or_error(operator.index, axis_size, "normalize axis_size")
+    return _canonicalize_axis(idx, ndim)
+  except (TypeError, ValueError, IndexError, core.ConcretizationTypeError):
+    if issubdtype(_dtype(index), np.unsignedinteger):
+      return index
```
jakevdp, line 3426: This block can go before the `try` statement I think

jax/_src/numpy/lax_numpy.py
```diff
@@ -3417,17 +3417,22 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
...
-  else:
-    axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
-                                             _dtype(index))
-  return lax.select(
-    lax.lt(index, _lax_const(index, 0)),
-    lax.add(index, axis_size_val),
-    index)
+  try:
+    idx = core.concrete_or_error(operator.index, index, "normalize index")
+    ndim = core.concrete_or_error(operator.index, axis_size, "normalize axis_size")
```
jakevdp, line 3422: I think you can simplify this: concrete tracers implement the `__index__` method, so if the concretization error is not going to be user-facing, there's no benefit to calling `core.concrete_or_error`. Instead you could just call `operator.index` directly, and you'd only have to catch the `TypeError`.

jakevdp: Actually, I just realized that if `operator.index` is called on a non-concrete tracer, it will result in a `TracerIntegerConversionError`, so you'll have to also catch that.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3417,17 +3417,22 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
...
+    return _canonicalize_axis(idx, ndim)
+  except (TypeError, ValueError, IndexError, core.ConcretizationTypeError):
+    if issubdtype(_dtype(index), np.unsignedinteger):
+      return index
+    if core.is_constant_dim(axis_size):
+      axis_size_val = _lax_const(index, axis_size)
```
jakevdp, line 3428: The rest of this should probably not be in the `except` block, because it will cause confusing tracebacks if something in here raises an error. perhaps use `except: pass` followed by the rest?

jax/_src/numpy/lax_numpy.py
lgeiger: Great, this is much nicer! `TracerIntegerConversionError` is a subclass of `TypeError`, so this will already be caught together with errors from `operator.index`.
I changed it in the latest commit.

jax/_src/numpy/lax_numpy.py
lgeiger: 👍  Changed it. This reduces the diff of the PR.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3417,6 +3417,11 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
 
 def _normalize_index(index, axis_size):
   """Normalizes an index value in the range [-N, N) to the range [0, N)."""
+  try:
+    idx, ndim = operator.index(index), operator.index(axis_size)
+    return _canonicalize_axis(idx, ndim)
+  except (TypeError, ValueError):
```
lgeiger, line 3423: Not catching `ValueError`s here would bring the behaviour closer to numpy but would [change the current behaviour for out of bound indexing](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#out-of-bounds-indexing) resulting in inconsistent behaviour.

jax/_src/numpy/lax_numpy.py
YouJiacheng: Hi! - @jakevdp , concrete tracer will raise exception on `__index__` but not `int`!
```python
from jax import grad
grad(lambda x: print(x.aval) or print(int(x)) or print(x.__index__()) or x)(1.0)
# ConcreteArray(1.0, dtype=float32, weak_type=True)
# 1
# jax._src.errors.TracerIntegerConversionError: The __index__() method was called on the JAX Tracer object 
# Traced<ConcreteArray(1.0, dtype=float32, weak_type=True)>with<JVPTrace(level=2/0)
```

jax/_src/numpy/lax_numpy.py
jakevdp: In that case maybe it's cleaner to not call canonicalize axis, and rather do this:
```python
return idx + ndim if idx < 0 else idx
```
then we don't have to worry about the `ValueError`

jakevdp: Also, I think a more idiomatic way to write this may be
```python
try:
  idx, ndim = operator.index(index), operator.index(axis_size)
except TypeError:
  pass
else:
  return idx + ndim if idx < 0 else idx
```
It makes clear where the TypeError is expected, and avoids catching unexpected errors.

jax/_src/numpy/lax_numpy.py
jakevdp: Seems reasonable - I can't think of any reason why the argument would not already be an array

jax/_src/numpy/lax_numpy.py
```diff
@@ -3897,13 +3915,11 @@ def _eliminate_deprecated_list_indexing(idx):
...
 def _is_boolean_index(i):
-  try:
-    abstract_i = core.get_aval(i)
-  except TypeError:
-    abstract_i = None
-  return (isinstance(abstract_i, ShapedArray) and issubdtype(abstract_i.dtype, bool_)
-          or isinstance(i, list) and i and _all(_is_scalar(e)
-          and issubdtype(_dtype(e), np.bool_) for e in i))
+  if isinstance(i, core.Tracer):
+    i = i.aval
```
jakevdp, line 3919: Why did you switch to an instance check here but continue using `try: core.get_aval()` in other places?

tests/lax_numpy_indexing_test.py
```diff
@@ -871,6 +871,15 @@ def testTrivialGatherIsntGenerated(self):
...
     self.assertEqual(len(jaxpr.jaxpr.eqns), 0)
 
+  @parameterized.named_parameters(
+      {"testcase_name": f"_{idx}", "idx": idx}
+      for idx in (-3, 0, 5, np.array(6), jnp.array(7)))
+  def testConstantIndexing(self, idx):
+    x = jnp.arange(10)
+    jaxpr = jax.make_jaxpr(lambda: x[idx])()
+    self.assertEqual(len(jaxpr.jaxpr.eqns), 1)
+    self.assertEqual(jaxpr.jaxpr.eqns[0].primitive, lax.gather_p)
```
jakevdp, line 881: I might add another test here to confirm that if the index is a static array with `ndim > 0`, that we use traced operations rather than treating the index array as a constant. From your code I don't think there's any chance that would slip through, but it would be useful if there are any future refactorings.

jakevdp (changes requested):
Thanks! Looks good, a couple more comments

tests/lax_numpy_indexing_test.py
lgeiger: I think that's already covered in the following test, isn't it?
https://github.com/google/jax/blob/0ed29b63f08fa404b1e837b4ab8b96b481c4b78f/tests/lax_numpy_indexing_test.py#L623-L634

jax/_src/numpy/lax_numpy.py
lgeiger: `core.get_aval()` will always create a `DeviceArray` if the argument is not a `Tracer`. This is unnecessary when calling it with a scalar Python index and is also quite a bit slower than using an `isinstance` check:
```python
In [3]: %timeit core.get_aval(3)
8.94 µs ± 260 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit isinstance(3, core.Tracer)
67.2 ns ± 0.969 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Since these checks happen quite often this was noticeable in the performance profile, so I tried to remove it where needed. I left calls to `core.get_aval` untouched in all other code paths to keep this PR limited to indexing with a single Python indices, but I'm happy take a look at optimising other forms of constant indexing in a future PR if that would be appreciated.

jax/_src/numpy/lax_numpy.py
lgeiger: That's a lot cleaner. Done 👍

tests/lax_numpy_indexing_test.py
jakevdp: Ah, I missed that, thanks

jakevdp (approved):
Looks great, thanks!

tests/lax_numpy_indexing_test.py
```diff
@@ -871,6 +871,15 @@ def testTrivialGatherIsntGenerated(self):
     jaxpr = jax.make_jaxpr(lambda x: x[:4])(np.arange(4))
     self.assertEqual(len(jaxpr.jaxpr.eqns), 0)
 
+  @parameterized.named_parameters(
+      {"testcase_name": f"_{idx}", "idx": idx}
+      for idx in (-3, 0, 5, np.array(6), jnp.array(7)))
```
jakevdp, line 876: Creating a `jnp.array` here at the global scope is not supported in some of our test frameworks, because the backend has not yet been initialized. I'd suggest doing something like
```
for idx in (-3, 0, 5)
for idx_type in (int, np.array, jnp.ndarray)
```
and then convert idx as necessary.

jakevdp (changes requested):

tests/lax_numpy_indexing_test.py
lgeiger: :+1: changed it.

tests/lax_numpy_indexing_test.py
```diff
@@ -871,6 +871,18 @@ def testTrivialGatherIsntGenerated(self):
     jaxpr = jax.make_jaxpr(lambda x: x[:4])(np.arange(4))
     self.assertEqual(len(jaxpr.jaxpr.eqns), 0)
 
+  @parameterized.named_parameters(
+      {"testcase_name": f"_{idx_type_name}_{idx}", "idx": idx, "idx_type": idx_type}
+      for idx in (-3, 0, 5)
+      for idx_type_name, idx_type in (
+        ("int", int), ("np.array", np.array), ("jnp.array", jnp.array)))
```
lgeiger, line 878: Let me know if there is a better way to get the `idx_type` name automatically for the test name.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3647,8 +3655,17 @@ def _merge_static_and_dynamic_indices(treedef, static_idx, dynamic_idx):
       idx.append(s)
   return treedef.unflatten(idx)
 
-def _int(aval):
-  return not aval.shape and issubdtype(aval.dtype, integer)
+def _is_basic_int_index(x):
+  if isinstance(x, core.Tracer):
+    aval = x.aval
+    return (isinstance(aval, (ConcreteArray, ShapedArray)) and
```
jakevdp, line 3661: `ConcreteArray` is a subclass of `ShapedArray`, so `isinstance(aval, ShapedArray)` should be sufficient.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3647,8 +3655,17 @@ def _merge_static_and_dynamic_indices(treedef, static_idx, dynamic_idx):
...
+            not aval.shape and issubdtype(aval.dtype, integer))
+  try:
+    operator.index(x)
+  except TypeError:
+    return False
+  else:
+    return not isinstance(x, bool)
```
jakevdp, line 3668: I think the logic here misses one corner case, though the behavior is deprecated in the current version of numpy:
```python
>>> idx = np.array([True], dtype=bool)[0]
>>> operator.index(idx)
DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index
1
```

jakevdp (changes requested):

jax/_src/numpy/lax_numpy.py
lgeiger: Updated in bb518728a576a67f3e4ff9c48552dd58600b8df7

jax/_src/numpy/lax_numpy.py
lgeiger: Good catch, I wasn't aware of that. f23efb9f50880fa479323d35b8616a94e8fce37c should handle this case correctly.

jakevdp:
It looks like this breaks some jax2tf tests, namely `jax/experimental/jax2tf/tests/shape_poly_test.py:ShapePolyPrimitivesTest.test_prim_getitem_op=poly_idx=slice-poly_noxla_poly_axes=[0]` (this test is not run as part of the github CI)

lgeiger:
For now I reverted 8e501651d54e765b4b65907e7612015be5e0263e which should fix the broken test.

jakevdp:
Thanks for the updates.

So, overall my concern here is that this PR touches a *lot* of parts of the code, all of which have potential side-effects that are difficult for me to predict. The passing tests are a good sign, but I'm not confident that our test coverage is complete when it comes to things like all the various ways that arrays can be indexed. A PR like this seems ripe for unanticipated and unintended downstream consequences.

Unfortunately, I'm not really sure what to suggest – it's hard to check for unknown issues. I'll pull-in the changes and run a larger suite of internal tests in case that shows us something.

jakevdp:
Well, additional testing didn't find any failures. I think we're in good shape

lgeiger:
Thanks for the detailed review and the help!
In case any unforeseen problems arise, I'm happy to take another look or extract  10addfd040acbbd1972a6697bf9c8134a1248043 and 38cee9f68644cb27307513ad56f2dad9134b5259 into separate PRs.

PR #10395, opened 2022-04-21
prefer operator.index over int

YouJiacheng:
`int` allows `float`, while `operator.index` doesn't
```python
int(1.1)
int(np.float32(1.1))
```
I can't understand `interpreters.masking` so I don't change it.
Breaking change:
`int` allows class with `__index__` and `__int__`, while `operator.index` only allows `__index__`.
Let's see if this break anything.

### DIFF
--- a/jax/_src/nn/initializers.py
+++ b/jax/_src/nn/initializers.py
@@ -18,6 +18,7 @@ used in Keras and Sonnet.
 """
 
 
+import operator
 from typing import Any, Callable, Sequence, Union
 
 import numpy as np
@@ -129,15 +130,15 @@ def _compute_fans(shape: core.NamedShape, in_axis=-2, out_axis=-1,
   if isinstance(in_axis, int):
     in_size = shape[in_axis]
   else:
-    in_size = int(np.prod([shape[i] for i in in_axis]))
+    in_size = operator.index(prod(shape[i] for i in in_axis))
   if isinstance(out_axis, int):
     out_size = shape[out_axis]
   else:
-    out_size = int(np.prod([shape[i] for i in out_axis]))
+    out_size = operator.index(prod(shape[i] for i in out_axis))
   if isinstance(batch_axis, int):
     batch_size = shape[batch_axis]
   else:
-    batch_size = int(np.prod([shape[i] for i in batch_axis]))
+    batch_size = operator.index(prod(shape[i] for i in batch_axis))
   receptive_field_size = shape.total / in_size / out_size / batch_size
   fan_in = in_size * receptive_field_size
   fan_out = out_size * receptive_field_size
--- a/jax/_src/numpy/polynomial.py
+++ b/jax/_src/numpy/polynomial.py
@@ -109,7 +109,7 @@ Also, it works best on rcond <= 10e-3 values.
 @partial(jit, static_argnames=('deg', 'rcond', 'full', 'cov'))
 def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
   _check_arraylike("polyfit", x, y)
-  deg = core.concrete_or_error(int, deg, "deg must be int")
+  deg = core.concrete_or_error(operator.index, deg, "deg must be int")
   order = deg + 1
   # check arguments
   if deg < 0:
--- a/jax/core.py
+++ b/jax/core.py
@@ -1676,8 +1676,7 @@ def canonicalize_shape(shape: Shape, context: str="") -> Shape:
   try:
     return tuple(unsafe_map(_canonicalize_dimension, shape))
   except TypeError:
-    pass
-  raise _invalid_shape_error(shape, context)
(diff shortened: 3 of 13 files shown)


jax/core.py
```diff
@@ -1676,8 +1676,7 @@ def canonicalize_shape(shape: Shape, context: str="") -> Shape:
   try:
     return tuple(unsafe_map(_canonicalize_dimension, shape))
   except TypeError:
-    pass
-  raise _invalid_shape_error(shape, context)
+    raise _invalid_shape_error(shape, context)
```
YouJiacheng, line 1679: Not related to this PR, I just can't help simplifying this.

jax/_src/nn/initializers.py
```diff
@@ -129,15 +130,15 @@ def _compute_fans(shape: core.NamedShape, in_axis=-2, out_axis=-1,
...
   if isinstance(out_axis, int):
     out_size = shape[out_axis]
   else:
-    out_size = int(np.prod([shape[i] for i in out_axis]))
+    out_size = operator.index(prod(shape[i] for i in out_axis))
   if isinstance(batch_axis, int):
     batch_size = shape[batch_axis]
   else:
-    batch_size = int(np.prod([shape[i] for i in batch_axis]))
+    batch_size = operator.index(prod(shape[i] for i in batch_axis))
```
YouJiacheng, line 141: `np.prod([])` will return `1.0` of type `np.float64`, and `prod` is faster than `np.prod` in most cases.

jakevdp:
While I agree that in many cases this might be the right change to make, it is a backward incompatible breaking change, so it's probably not the right thing to do in bulk. For instance, for any `jax.numpy` or `jax.scipy` functions, we should ensure that the behavior matches that of the corresponding `numpy` or `scipy` function. Any such change that is found to be warranted should probably have its own deprecation cycle following the [API Compatibility Policy](https://jax.readthedocs.io/en/latest/api_compatibility.html).

YouJiacheng:
@jakevdp Do you remember if this test case use `float` for `fori_loop` deliberately?
https://github.com/google/jax/blob/0fc93a0a994204e79b97566d7fc7dff07462d2d8/tests/lax_control_flow_test.py#L1971-L1974
IIUC, the signature of `fori_loop` only allows Int.

YouJiacheng:
So, should I include the internal part(w/o public API compatibility change) in this PR, while defer others to some other PRs&only add deprecation warning?

YouJiacheng:
BTW, there are many functions are annotated with `int`. In these cases, do the changes break API compatibility, or fix the bug?

YouJiacheng:
@jakevdp Actually there are some functions don't match the `numpy` behavior.
> For instance, for any jax.numpy or jax.scipy functions, we should ensure that the behavior matches that of the corresponding numpy or scipy function.

For example.
```python
import numpy as np
np.ravel_multi_index([1], [2.0])
# File "<__array_function__ internals>", line 180, in ravel_multi_index
# TypeError: 'float' object cannot be interpreted as an integer
```
While
```
import jax.numpy as jnp
jnp.ravel_multi_index([1], [2.0]) # OK
```

jakevdp:
`grad` is only compatible with float arguments, so I believe using float there is intentional.

YouJiacheng:
@jakevdp I mean why not
```python
func = lambda x: lax.fori_loop(int(x), int(x + 2), lambda i, c: c, x) 
jax.grad(func)(1.)  # doesn't crash
jax.linearize(func, 1.)  # doesn't crash
```

jakevdp:
That may be reasonable – my concern with this PR is that it includes 13 files, each with multiple changes, and every change needs individual consideration to determine (1) whether it's correct, and (2) whether it needs a deprecation cycle. The fact that everything is changed in a single commit means that if a rollback is required for a single issue, then all will have to be rolled back. I just don't see any way for it to be feasible to accept this change.

YouJiacheng:
I definitely will split this PR into parts~👍🥰

jakevdp:
Please don't just split this into a PR per change – that will be too much.

What I'd like to see is, e.g. one change addressing all cases where numpy fails for integer input but jax does not; one change addressing cases where scipy fails for integer input but jax does not; etc.

jax/_src/numpy/polynomial.py
```diff
@@ -109,7 +109,7 @@ def roots(p, *, strip_zeros=True):
 @partial(jit, static_argnames=('deg', 'rcond', 'full', 'cov'))
 def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
   _check_arraylike("polyfit", x, y)
-  deg = core.concrete_or_error(int, deg, "deg must be int")
```
YouJiacheng, line 112: `numpy` use `int` here as well.

hawkinsp:
Closing as stale.

PR #10396, opened 2022-04-21
Some Simplification

YouJiacheng:
`list.index` already return `int`
`prod` is faster than `np.prod` in most cases, and `np.prod([])` returns `1.0` of type `np.float64`.

### DIFF
--- a/jax/_src/nn/initializers.py
+++ b/jax/_src/nn/initializers.py
@@ -129,15 +129,15 @@ def _compute_fans(shape: core.NamedShape, in_axis=-2, out_axis=-1,
   if isinstance(in_axis, int):
     in_size = shape[in_axis]
   else:
-    in_size = int(np.prod([shape[i] for i in in_axis]))
+    in_size = prod(shape[i] for i in in_axis)
   if isinstance(out_axis, int):
     out_size = shape[out_axis]
   else:
-    out_size = int(np.prod([shape[i] for i in out_axis]))
+    out_size = prod(shape[i] for i in out_axis)
   if isinstance(batch_axis, int):
     batch_size = shape[batch_axis]
   else:
-    batch_size = int(np.prod([shape[i] for i in batch_axis]))
+    batch_size = prod(shape[i] for i in batch_axis)
   receptive_field_size = shape.total / in_size / out_size / batch_size
   fan_in = in_size * receptive_field_size
   fan_out = out_size * receptive_field_size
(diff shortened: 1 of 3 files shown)


jax/_src/nn/initializers.py
```diff
@@ -129,15 +130,15 @@ def _compute_fans(shape: core.NamedShape, in_axis=-2, out_axis=-1,
   if isinstance(in_axis, int):
     in_size = shape[in_axis]
   else:
-    in_size = int(np.prod([shape[i] for i in in_axis]))
+    in_size = operator.index(prod(shape[i] for i in in_axis))
```
jakevdp, line 133: For each of these changes: it's not clear to me why it's warranted / desired. Can you say more about what use-case motivated the change?

YouJiacheng: Ah, it doesn't change any functionality, thus there isn't a use-case...
I think I can remove the `operator.index`, just use `prod`.

YouJiacheng: Now it is a chore simplification PR. WDYT?

jakevdp: I think the principle at play here is "if it's not broken, don't fix it". Every change to the codebase has a small but non-zero risk of creating an unintended consequence, so we should avoid making changes unless there is a good reason to do so.

jax/core.py
```diff
@@ -1676,8 +1676,7 @@ def canonicalize_shape(shape: Shape, context: str="") -> Shape:
   try:
     return tuple(unsafe_map(_canonicalize_dimension, shape))
   except TypeError:
-    pass
-  raise _invalid_shape_error(shape, context)
+    raise _invalid_shape_error(shape, context)
```
jakevdp, line 1679: For what it's worth, I think the reason to avoid raising the new error within the `except` statement is that it keeps the irrelevant TypeError out of the user-facing traceback.

jax/_src/nn/initializers.py
jakevdp: > Though in theory the `int(list.index())` to `list.index()` change is zero risk

Agreed, but if we're certain the new code does nothing, why add it?

I think this PR could probably be closed.

jax/core.py
YouJiacheng: Yes, you are right, I have just realized this point few minutes before.
Correct simplification should be `raise _invalid_shape_error(shape, context) from None`, and it is still not equivalent to the original code if `canonicalize_shape` is called inside a `try .. except ...`, in which the original implementation will include correct context information.

PR #10397, opened 2022-04-21
replace int with operator.index for numpy

YouJiacheng:
This change align the behavior (about integer argument) of `ravel_multi_index`, `split`, `indices` and `diag` to their `numpy` counterparts.
Also ensure size argument of `nonzero` should be integer.
The changes with `*space` are only simplification
If there is any deprecation cycle should be added, please point it out.
Note: `np.polyfit` use `int` as well, thus I don't change `jnp.polyfit`


jakevdp (changes requested):
Looks good!

This is one of those gray areas in terms of our API deprecation policy. It is technically backward-incompatible (code that passes a float in some cases will now fail) but I agree that we do want this kind of input validation.

I'd lean toward not doing a deprecation cycle for this. But it definitely needs a note in the CHANGELOG, mentioning the functions and arguments which have these stricter requirements.

jakevdp (approved):
Looks good!

PR #10398, opened 2022-04-21
replace int with operator.index for lax

YouJiacheng:
Note caller of _eye has ensured int.
Change with `jax.numpy.tri` is included in this PR for potential reverting

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -64,6 +64,7 @@ from jax._src.lax.utils import (
   standard_named_shape_rule,
   standard_primitive,
   standard_translate,
+  _check_integer,
 )
 from jax._src.lax import slicing
 
@@ -1062,7 +1063,7 @@ def sort_key_val(keys: Array, values: Array, dimension: int = -1,
 
 def top_k(operand: Array, k: int) -> Tuple[Array, Array]:
   """Returns top ``k`` values and their indices along the last axis of ``operand``."""
-  k = int(k)
+  k = _check_integer("lax.topk", "k", k)
   if k < 0:
     raise ValueError("k argument to top_k must be nonnegative, got {}".format(k))
   return top_k_p.bind(operand, k=k)
@@ -1118,7 +1119,7 @@ def broadcasted_iota(dtype: DType, shape: Shape, dimension: int) -> Array:
 
 def _eye(dtype: DType, shape: Shape, offset: int) -> Array:
   """Like numpy.eye, create a 2D array with ones on a diagonal."""
-  offset = int(offset)
+  assert type(offset) is int
   dtype = dtypes.canonicalize_dtype(dtype)
   bool_eye = eq(add(broadcasted_iota(np.int32, shape, 0), np.int32(offset)),
                 broadcasted_iota(np.int32, shape, 1))
@@ -1138,7 +1139,7 @@ def _delta(dtype: DType, shape: Shape, axes: Sequence[int]) -> Array:
 
 def _tri(dtype: DType, shape: Shape, offset: int) -> Array:
   """Like numpy.tri, create a 2D array with ones below a diagonal."""
-  offset = int(offset)
+  assert type(offset) is int
   dtype = dtypes.canonicalize_dtype(dtype)
   bool_tri = ge(add(broadcasted_iota(np.int32, shape, 0), np.int32(offset)),
                 broadcasted_iota(np.int32, shape, 1))
--- a/jax/_src/lax/utils.py
+++ b/jax/_src/lax/utils.py
@@ -20,6 +20,7 @@ import builtins
 from functools import partial
 import operator
 from typing import Callable
+import warnings
 
 from jax import core
 from jax._src import dtypes
@@ -115,3 +116,16 @@ def standard_named_shape_rule(*avals, **kwargs):
 
 def _standard_weak_type_rule(*avals, **kwargs):
   return all(aval.weak_type for aval in avals)
+
+def _check_integer(func_name: str, var_name: str, x):
+  try:
+    return operator.index(x)
+  except TypeError:
+    try:
+      x = int(x)
+    except Exception as e:
+      raise e from Exception(f"{func_name}: {var_name}")
+    msg = (f"{func_name}: {var_name} must be an integer."
+          "In future JAX versions non-integer values will result in an error.")
+    warnings.warn(msg)
+    return x
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2369,6 +2369,11 @@ def tri(N, M=None, k=0, dtype=None):
   lax_internal._check_user_dtype_supported(dtype, "tri")
   M = M if M is not None else N
   dtype = dtype or float32
+  try:
+    k = operator.index(k)
+  except TypeError:
+    warnings.warn("jnp.tri: k must be an integer. In future JAX versions non-integer values will result in an error.")
+    k = int(k)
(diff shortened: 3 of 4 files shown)


jax/_src/lax/lax.py
```diff
@@ -1138,7 +1139,7 @@ def _delta(dtype: DType, shape: Shape, axes: Sequence[int]) -> Array:
 
 def _tri(dtype: DType, shape: Shape, offset: int) -> Array:
   """Like numpy.tri, create a 2D array with ones below a diagonal."""
-  offset = int(offset)
+  assert type(offset) is int
```
jakevdp, line 1142: Here and elsewhere: we shouldn't assert that inputs are of type int, because (for example), this fails for numpy integer scalars.

jax/_src/lax/lax.py
```diff
@@ -1340,7 +1341,7 @@ def _iter(tracer):
   if tracer.ndim == 0:
     raise TypeError("iteration over a 0-d array")  # same as numpy error
   else:
-    n = int(tracer.shape[0])
+    n = operator.index(tracer.shape[0])
```
jakevdp, line 1344: Why is this change necessary? Is there any situation in which an element of a shape is not an integer?

jax/_src/numpy/lax_numpy.py
```diff
@@ -2369,6 +2369,11 @@ def tri(N, M=None, k=0, dtype=None):
   lax_internal._check_user_dtype_supported(dtype, "tri")
   M = M if M is not None else N
   dtype = dtype or float32
+  try:
+    k = operator.index(k)
+  except TypeError:
+    warnings.warn("using k supports int(k) but not k.__index__, e.g. float will be deprecated.", DeprecationWarning)
```
jakevdp, line 2375: I don't think the meaning or intent of this warning will be clear to users who see it. How about:
```
warnings.warn(f"jnp.tri: k must be an integer. In future JAX versions non-integer values will result in an error.")
```

jax/_src/lax/utils.py
```diff
@@ -115,3 +116,12 @@ def standard_named_shape_rule(*avals, **kwargs):
 
 def _standard_weak_type_rule(*avals, **kwargs):
   return all(aval.weak_type for aval in avals)
+
+def _int_replaced_by_index(x):
+  try:
+    return operator.index(x)
+  except TypeError:
+    msg = ("using types support int(x) but don't support x.__index__ as int, "
+          "e.g. float, np.float64 will be deprecated")
```
jakevdp, line 125: We should offer the user some clue here about the provenance of the error: imagine running a long & involved script and seeing several copies of this error, reported as coming from line 126 of `jax/_src/lax/utils.py`. What would you do to fix it?

Perhaps `_int_replaced_by_index` could accept the function name & variable name as strings, and put these in the warning to give context.

Alternatively, this may be an argument to inline this check in each place we use it, so that the warning is reported in the relevant context.

jax/_src/lax/control_flow.py
```diff
@@ -205,8 +206,8 @@ def fori_loop(lower, upper, body_fun, init_val):
   if (isinstance(core.get_aval(lower), ConcreteArray) and
       isinstance(core.get_aval(upper), ConcreteArray)):
     try:
-      lower_ = int(lower)
-      upper_ = int(upper)
+      lower_ = _int_replaced_by_index(lower)
+      upper_ = _int_replaced_by_index(upper)
```
jakevdp, line 210: The intent here is to check whether `lower` and `upper` are tracers. We should not raise deprecation warnings here, and we probably should not change this to `operator.index` in the long run.

tests/lax_control_flow_test.py
```diff
@@ -1969,7 +1969,7 @@ def loop(loop_impl, x):
     jtu.check_grads(loop_lax, (x,), order=2, modes=["fwd"])
 
   def testStaticForiGrad(self):
-    func = lambda x: lax.fori_loop(x, x + 2., lambda i, c: c, x)
+    func = lambda x: lax.fori_loop(int(x), int(x) + 2, lambda i, c: c, x)
```
jakevdp, line 1972: I don't think this test change is necessary once you address the comments within the `fori_loop` implementation.

jakevdp (changes requested):

jax/_src/lax/lax.py
YouJiacheng: I add a to `int` conversion in the only caller of `_tri` -- `jax.numpy.tri`, to be consistency with  `jax.numpy.eye`.

jax/_src/numpy/lax_numpy.py
YouJiacheng: Thanks! I am not familiar with how to write a clear warning.

jax/_src/lax/control_flow.py
YouJiacheng: Good catch! This is my mistake!

YouJiacheng: > The side effect here is that non-traced floating point array inputs would be switched from using the scan implementation to using the while_loop.implementation, which is not what we want.

What about using concrete check here?

jax/_src/lax/lax.py
YouJiacheng: Got it.

jax/_src/lax/control_flow.py
```diff
@@ -205,9 +206,9 @@ def fori_loop(lower, upper, body_fun, init_val):
   if (isinstance(core.get_aval(lower), ConcreteArray) and
       isinstance(core.get_aval(upper), ConcreteArray)):
     try:
-      lower_ = int(lower)
-      upper_ = int(upper)
-    except TypeError:
+      lower_ = _check_integer("lax.fori_loop", "lower", lower)
+      upper_ = _check_integer("lax.fori_loop", "upper", upper)
+    except TypeError: # change to TracerIntegerConversionError?
```
YouJiacheng, line 211: @jakevdp  Could you please give some advice?
Does `fori_loop` not allow `Tracer` with `ConcreteArray` type `aval`?

jakevdp: I believe that `int(x)` will return the concrete value if `x` is a tracer with a concrete aval.

jakevdp: Again, as far as I can tell this code works correctly and does exactly what we want it to do, so unless you have found a bug somewhere I'd say it's not broken, and there's no need to change it.

YouJiacheng: > Again, as far as I can tell this code works correctly and does exactly what we want it to do, so unless you have found a bug somewhere I'd say it's not broken, and there's no need to change it.

Sorry to be dense about this point, I have little experience in working for a real project. -- BTW, I am curious about how to get a tracer with a concrete aval.

jakevdp: It will mainly come in via autodiff. For example:
```python
from jax import grad

def f(x):
  print(x.aval)
  return x

_ = grad(f)(1.0)
# ConcreteArray(1.0, dtype=float32, weak_type=True)
```

YouJiacheng: Thanks for nice example! -- And I find Tracer define appropriate `__int__` method.

jax/_src/lax/lax.py
jakevdp: There's little or no downside to repeating `offset = operator.index(offset)` in place of an `isinstance` assertion; the upside is that if this low-level function is ever used elsewhere, it will be better behaved.

jakevdp:
Closing since this has bit-rotted – feel free to re-open if you wish to continue working on this!

PR #10406, opened 2022-04-21
CI: fix flake8 ignore declarations

### LINKED ISSUE #10014: flake8 ignore statements are incorrect
Throughout the package we use `flake8: noqa: F401` to attempt to ignore unused imports within a file. Unfortunately, I believe this has the effect of disabling flake8 *entirely* in such files.

According to https://flake8.pycqa.org/en/latest/user/violations.html, the way to accomplish this is to use the `per-file-ignorers` entry in our `setup.cfg`: https://github.com/google/jax/blob/4afd4b99d433378746fec0661edcdd0d744ed9b7/setup.cfg#L14-L15

jakevdp:
Fixes #10014 

The issue here is that we had been using the heading `# flake8: noqa: F411` to turn off unused import errors, but it turns out that this was parsed as `# flake8: noqa` which caused *all* flake8 checks to be disabled in each file where it appears. This has caused linting issues to slip into the codebase.

This PR instead uses the `setup.cfg` mechanism to disable this check in relevant files, and fixes some of the existing lint that this uncovered.


froystig (approved):
Good catch!

PR #10408, opened 2022-04-21
Apply flake8 checks to xmap_test.py

jakevdp:
They had previously been skipped via the `noqa` comment.

### DIFF
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -12,18 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# flake8: noqa
-
-from contextlib import contextmanager
 import functools
 import itertools as it
 import os
 import re
-import unittest
 from itertools import product, permutations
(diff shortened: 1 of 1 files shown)


tests/xmap_test.py
```diff
@@ -702,7 +696,7 @@ def testNestedMeshSPMD(self):
              axis_resources={'a': 'x', 'b': 'y'})
     xshape = (8, 2, 4, 5)
     x = jnp.arange(np.prod(xshape)).reshape(xshape)
-    y = f(x)
+    _ = f(x)
```
yashk2810, line 699: Just do `f(x)`?

yashk2810 (approved):

tests/xmap_test.py
jakevdp: Or maybe remove it? I wasn't clear if there was some required side-effect here for the test.

yashk2810: If you want to remove it then maybe test it internally first but running the tests on tpu tfrt.

PR #10417, opened 2022-04-22
Fix typo in _scatter_add_lower_gpu

YouJiacheng:
https://github.com/google/jax/commit/a87b21148c6d7eb9b46c751dde40b17ca0e7b03e doesn't notice `_scatter_add_lower_gpu` using `mlir.lower_fun` instead of `xla.lower_fun`.
I follow the change done in that commit for _scatter_lower.
https://github.com/google/jax/blob/0ed29b63f08fa404b1e837b4ab8b96b481c4b78f/jax/_src/lax/slicing.py#L1915-L1921

IIUC, `replace(avals_out=None)` is not necessary, just indicating that `avals_out` is not used.

I don't really understand what does the code do. Need review from @hawkinsp

### DIFF
--- a/jax/_src/lax/slicing.py
+++ b/jax/_src/lax/slicing.py
@@ -1965,7 +1965,7 @@ def _scatter_add_lower_gpu(ctx, operand, indices, updates,
 
   if mode == GatherScatterMode.CLIP:
     clip_fn = mlir.lower_fun(_clamp_scatter_indices, multiple_results=False)
-    (indices,), = clip_fn(ctx, ctx.avals_in, None, operand, indices, updates,
+    (indices,), = clip_fn(ctx.replace(avals_out=None), operand, indices, updates,
                           dnums=dimension_numbers)
 
   aval_out, = ctx.avals_out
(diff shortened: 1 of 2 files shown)


hawkinsp:
And just to explain what this code does:

`mlir.lower_fun` is a means to call back into JAX from an MHLO translation rule. It means "trace through this otherwise normal JAX code, and emit it as MHLO". It's sort of an internal inline `jit`: it's frequently much more convenient to write JAX user code than to write MHLO op builder code directly. If there's any non-trivial algorithm I need to write as part of a lowering rule I tend to reach for this feature. Here: it's logic for clamping gather/scatter indices into range if they are out-of-bounds.

One of the things the lowering code expects as part of its `LoweringRuleContext` is the output types, which we normally know because they are present on the jaxpr. Here I was feeling lazy and simply omitted them, because I happen to know that they aren't used. But I didn't update *how* they are passed for the difference between the old XLA translation rules and the newer MHLO lowering rules.

YouJiacheng:
I am working on adding a regression test, but I found the `dimension_numbers` of `scatter` or `scatter_add` is hard to understand, and doesn't notice that there has been a test for `scatter_add`, thanks for point that out! I will add it.

hawkinsp (approved):

PR #10424, opened 2022-04-22
Annotate vmap

NeilGirdhar:
Document the fact that both in_axes and out_axes support arbitrary
composite (not just container) types.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1384,7 +1384,11 @@ def _split(x, indices, axis):
     return x.split(indices, axis)
 
 
-def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
+def vmap(fun: F,
+         in_axes: Union[int, Sequence[Any]] = 0,
+         out_axes: Any = 0,
+         axis_name: Optional[Hashable] = None,
+         axis_size: Optional[int] = None) -> F:
   """Vectorizing map. Creates a function which maps ``fun`` over argument axes.
 
   Args:


jax/_src/api.py
```diff
@@ -1302,13 +1302,17 @@ def _split(x, indices, axis):
     return x.split(indices, axis)
 
 
-def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
+def vmap(fun: F,
+         in_axes: Union[None, int, Tuple[Any, ...]] = 0,
```
jakevdp, line 1306: Since `in_axes` can be an arbitrary pytree, I think the only viable type annotation is `Any`.

NeilGirdhar: @jakevdp Hmmm, I just tested it:
```python
from jax import vmap
import jax.numpy as jnp

def f(d):
    return d['blah']


vf = vmap(f, in_axes={'blah': 0})

try:
    vf({'blah': jnp.zeros(10)})
except:
    print("guess not?")

vf_tuple = vmap(f, in_axes=({'blah': 0},))
print(vf_tuple({'blah': jnp.zeros(10)}))
```
Am I doing something wrong here?  It seems that `in_axes` should be a tuple corresponding to the parameters of `f`?

jakevdp: I guess you're right, it needs to be a tuple of pytrees.

NeilGirdhar: Great :smile: Please let me know if the text can be improved at all.  I tried to keep it understandable for the most common cases, but it's hard when the underlying library is so powerful!

jakevdp: sorry, I'm just overly suspicious of type annotation PRs because annotations tend to have very poor test coverage in JAX, and these types of PRs have often needed to be rolled-back due to unexpected failures in downstream packages.

jakevdp: The benefit of `Any` is that you never have false positives 😀

jakevdp: Looks like the lint check caught an issue with the `in_axes` annotation

jax/_src/api.py
```diff
@@ -1302,47 +1302,53 @@ def _split(x, indices, axis):
...
-      range ``[-ndim, ndim)`` for each array, where ``ndim`` is the number of
-      dimensions (axes) of the corresponding input array.
-
-      If the positional arguments to ``fun`` are container (pytree) types, the
-      corresponding element of ``in_axes`` can itself be a matching container,
-      so that distinct array axes can be mapped for different container
-      elements. ``in_axes`` must be a container tree prefix of the positional
-      argument tuple passed to ``fun``. See this link for more detail:
-      https://jax.readthedocs.io/en/latest/pytrees.html#applying-optional-parameters-to-pytrees
+    in_axes: An optional integer--or a sequence of pytrees whose fields are
```
jakevdp, line 1314: So the benefit of the previous docstring is that it would be meaningful to users who are unfamiliar with JAX. Now it immediately jumps into a discussion that assumes an understanding of the term "pytree". I wonder if we could add the information about pytrees somewhere without making the entire discussion opaque to new users?

NeilGirdhar:
@jakevdp You're right of course.  I'll look at it tonight try to make it simpler for new users.

jax/_src/api.py
jakevdp: Actually, I think the previous version does this fairly well: the bulk of the description uses familiar terms, and it only discusses pytrees in the last paragraph.

jax/_src/api.py
```diff
@@ -1316,7 +1316,11 @@ def _split(x, indices, axis):
     return x.split(indices, axis)
 
 
-def vmap(fun: F, in_axes=0, out_axes=0, axis_name=None, axis_size=None) -> F:
+def vmap(fun: F,
+         in_axes: Union[None, int, Sequence[Any]] = 0,
```
jakevdp, line 1320: I don't think `None` is a viable option here. You can have `None` within a sequence of `in_axes`, but not by itself.

jax/_src/api.py
```diff
@@ -1316,7 +1316,11 @@ def _split(x, indices, axis):
...
+         out_axes: Any = 0,
```
jakevdp, line 1321: I think `out_axes` should probably be `Union[None, int, Sequence[Any]]` (unlike `in_axes`, it can be `None` by itself)

NeilGirdhar:
test

jax/_src/api.py
jakevdp: I see - so maybe `out_axes: Union[None, int, Any, Sequence[Any]]` would make sense? Or perhaps we could replace `Any` with `PyTree` as an alias of `Any`?

The fundamental question here is: what are we writing the type hints for? If it's for static type analysis only, than I agree that `out_axes: Any` is probably the best we can do (given the extent to which JAX relies on duck typing here). But if it's to help programmers reading the code to have a better idea of what inputs are valid, we could try to be more specific.

NeilGirdhar: >  But if it's to help programmers reading the code to have a better idea of what inputs are valid, 

I definitely agree that if we can help readers, we should.

How about:
```python
AxisPyTree = Any  # A PyTree whose leaves are None or integers.

def vmap(fun: F,
         in_axes: Union[int, Sequence[AxisPyTree]] = 0,
         out_axes: AxisPyTree = 0,
         ...
         ) -> F:
  ...
```

(I don't think `Sequence` should appear in `out_axes` since sequence isn't relevant there.  The reason that `Sequence` appears in `in_axes` is because they correspond to each parameter.)

jakevdp: For functions which return multiple values, `Sequence` is relevant in `out_axes`.

NeilGirdhar: But if the function returns multiple values as a tuple, then `out_axes` can't be a list, and vice versa.  This is different than with `in_axes`.  It always accepts any `abc.Sequence` type.

NeilGirdhar: (I updated the pull request.)

jakevdp: Thanks for the clarificaiton on Sequence vs Tuple.

I still find the current version misleading: it implies that the `out_axes` must be a pytree. But it can also be `None`, or an integer, or a tuple of any combination thereof. It seems confusing to annotate it with `AxisPyTree` alone, even if that's technically correct due to it being an alias of `Any`.

Wouldn't something like this be more correct/informative?
```python
out_axes: Union[int, None, AxisPyTree, Tuple[int, None, AxisPyTree]]
```

jakevdp: If that seems too complicated, then maybe removing the alias and just sticking to `Any` is better here. It is, after all, the simplest correct annotation in this case.

NeilGirdhar: I'm happy to revert to `Any` since that's more intuitive to you.  But correct me if I'm wrong, but aren't `None` and `int` "PyTrees whose leaves are None or integers"?  I guess I think of it as the base case.

> Wouldn't something like this be more correct/informative?

If you do that, then you'd want to do the same for `in_axes` and make it a sequence of what you have for `out_axes`.  Or more simply make `AxisPyTree` more complicated.  Or even more simply, just clarify the comment on `AxisPyTree`.  Maybe that's the best solution?  How about:
```python
# A PyTree whose leaves are None or integers.  For example, None, 1, (2, None), {'some_key': 0}, [(0, None), {'key': 1}], etc.
AxisPyTree = Any  
```

jakevdp: Yeah, the best solution would be to use recursively-defined types, but as far as I'm aware that't not yet possible in Python's type specification model.

But I'd still say that `out_axes: AxisPyTree` does not, on its face, suggest that (1) `int` and `None` are valid, and (2) if your function returns multiple outputs, you need to use a tuple rather than a list or other sequence. If we're leaving things that vague, we should just use `Any` to make it clear that the true type is un-annotatable.

NeilGirdhar: Done, as requested.

jakevdp: Sorry I may have been unclear: I think having a more specific type annotation would be a good thing. My point is that `AxisPyTree` by itself is uninformative, and no better (and probably worse) than `Any`. If we're attempting to add useful type annotations to the function, my preference would be something closer to `out_axes: Union[None, int, AxisPyTree, Tuple[AxisPyTree, ...]]`

NeilGirdhar: > Sorry I may have been unclear: I think having a more specific type annotation would be a good thing.

Oh great, I agree.  I thought it was a good idea.

> . If we're attempting to add useful type annotations to the function, my preference would be something closer to out_axes: Union[None, int, AxisPyTree, Tuple[AxisPyTree, ...]]

Why stop there then?  Why not take the pattern further:
```python
Union[None, int, AxisPyTree, Tuple[AxisPyTree, ...], List[AxisPyTree], Dict[str, AxisPyTree], SomeDataClass, SomeGenericDataclass[AxisPyTree]]  # and so on, and so on.
```
There is no way to annotate the broad variety of possible pytrees.  Correct me if I'm wrong, but nowhere else in Jax are general pytrees annotated in any way other than with simply `Any`.

The ideal solution is to write a MyPy extension that understands pytrees, and collects all of the pytree registration, and checks that the leaves are appropriate pytrees.  This would be very cool, but I'm not sure it would catch too many of my bugs, so it has limited upside compared with the work it would require.

> My point is that AxisPyTree by itself is uninformative, and no better (and probably worse) than Any

The benefit of `AxisPyTree` is that it has a useful comment that explains that the pytree leaves must be `None | int`.  It would also be useful in case a plugin ever does get written.

> `out_axes: Union[None, int, AxisPyTree, Tuple[AxisPyTree, ...]]`

Also, just for the record, `None` _is_ a pytree.  It is the simplest pytree.  So, I don't see the benefit of pulling it out as a special case.  It's not special.  Same goes for `int`.

This is not the same as `Sequence` that we see in `in_axes` because _that_ type is not part of the pytree.  That sequence is being interpreted by Jax when it makes a correspondence between the elements of the sequence and the parameters.  The fact that Jax accepts Sequence here is an arbitrary design choice.  Jax could have equivalently chosen to support `Iterable`, or a mapping from parameter index.  Thus, `in_axes` truly is `Sequence[AxisPyTree] | int`.  I think `out_axes` really is just `AxisPyTree`.

What do you think?

jakevdp: > Why stop there then? Why not take the pattern further:

Like I said before, what we really need are recursively-defined types, which are not supported. Without those, we **cannot** provide precisecly accurate annotations. So the question is, how far do we go? My argument is that
```
out_axes: AxisPyTree
```
is not far enough, because it obscures the fact that you can pass `None`, or `int`, or a tuple. The next step would be something like
```
out_axes: Union[None, int, Tuple[AxisPyTree]]
```
This is not a precisely correct type annotation (again, Python's lack of recursive types makes a precisely correct annotation impossible) but it's better. You could go another step deeper with something like
```
out_axes: Union[None, int, Tuple[Union[None, int, AxisPyTree]]]
```
but in my opinion, at this point the annotations are becoming complicated enough that they're no longer as useful for communicating to users what inputs are valid.

jakevdp: > Also, just for the record, `None` is a pytree. It is the simplest pytree.

I understand this. I don't think we should expect users glancing through the code to understand this.

jakevdp: >  I think `out_axes` really is just `AxisPyTree`.

Same here: you and I understand this, but I don't think it's clear to the average user who has not thought through this in detail. I think most users would not look at `out_axes: AxisPyTree` and think, "oh good, I can pass a tuple of PyTree objects".

jakevdp: That's why I think `out_axes: Any` is preferable to `out_axes: AxisPyTree`. The former implies that the user has to look at the docs to understand what to pass; the latter implies that tuples are not allowed.

NeilGirdhar: > Same here: you and I understand this, but I don't think it's clear to the average user who has not thought through this in detail. I think most users would not look at out_axes: AxisPyTree and think, "for functions with multiple results, I should pass a tuple of PyTree objects"

Okay!  I see what you're driving at.

This was the reason that we backed out my rewrite of the docstring, and I totally agree with you that a user won't appreciate what PyTree really means, and so they need a lot of examples, especially of the common cases.  Makes perfect sense to me.

However, I'm not convinced of the benefit of making the type annotations more complicated to cater to new users of Jax.

How about write a long, and clear explanation of PyTrees somewhere in the documentation, and making every function that accepts or returns PyTrees link to that documentation?  Then you can keep the type annotations simple, and you may even be able to simplify the docstrings.

And there is a benefit to simplifying docstrings.  It took me a long time to figure out from the docstring that vmap accepts arbitrary pytrees, and what that means.

> The former implies that the user has to look at the docs to understand what to pass; the latter implies that tuples are not allowed.

I don't think it's fair to say that the latter implies that tuples aren't allowed.  You could define
```python
AxisPyTree = None | int | tuple[None | int, ...]  # and so on.
```
It's fair to expect the user to look up `AxisPyTree` to see how it's defined.  And while they're looking up the definition, I think it's fair to expect them to read the comment that goes along with it.

I don't know.  I don't think you should have to pollute every usage of `AxisPyTree` with a disjunction of cases.  I think there should be a way to write a clear, concise annotation with a good comment for new users.  Is there no way to have documentation (comments, docstrings, or online) to make this clear?

jakevdp: I see - in that case I'd say let's go back to `out_axes: Any`, because it makes clear that (1) there is no static type constraint on `out_axes` (which is also true with `out_axes: AxisPyTree`, but less directly), and (2) if you want to understand what should be passed here, you should read the docstring.

jakevdp: We could even simplify things and use `in_axes: Union[None, int, Sequence[Any]]` which is I think as tight a type constraint as we can make there.

NeilGirdhar: > I see - in that case I'd say let's go back to out_axes: Any, b

Okay, perfect, so are there any changes you want me to make the pull request?

> We could even simplify things and use in_axes: Union[None, int, Sequence[Any]] which is I think as tight a type constraint as we can make there.

Yes, I agree.

jakevdp (approved):
Thanks for the discussion!

NeilGirdhar:
@jakevdp (Thank you for the discussion!  It really makes this a fun project to contribute to.)

PR #10429, opened 2022-04-23
remove useless&expensive core.get_aval in jnp.power

YouJiacheng:
useless: call operator.index on a concrete Tracer will raise TracerIntegerConversionError as well.
If anyone want to handle concrete Tracer, I can use
try: x2 = x2.aval.val
expensive: `core.get_aval(2)` cost 11us.

Improvement:
`x = jnp.full((2, 2), 2.0)`
`(x ** 2).block_until_ready()` on CPU
80us -> 48us
`(x ** 2.0).block_until_ready()` on CPU
22us -> 8us

### DIFF
--- a/jax/_src/numpy/ufuncs.py
+++ b/jax/_src/numpy/ufuncs.py
@@ -317,13 +317,12 @@ def power(x1, x2):
   # Using lax.pow may be imprecise for floating-point values; the goal of this
   # code path is to make sure we end up with a precise output for the common
   # pattern ``x ** 2`` or similar.
-  if isinstance(core.get_aval(x2), core.ConcreteArray):
-    try:
-      x2 = operator.index(x2)
-    except TypeError:
-      pass
-    else:
-      return lax.integer_pow(x1, x2)
+  if isinstance(x2, int):
+    return lax.integer_pow(x1, x2)
+  if isinstance(x2, core.Tracer):
+    return _power(x1, x2)
+  if issubclass(np.result_type(x2).type, np.integer) and np.ndim(x2) == 0:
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/ufuncs.py
```diff
@@ -317,13 +317,12 @@ def power(x1, x2):
...
-    except TypeError:
-      pass
-    else:
-      return lax.integer_pow(x1, x2)
+  try:
+    x2 = operator.index(x2)
+  except TypeError:
+    pass
+  else:
+    return lax.integer_pow(x1, x2)
```
jakevdp, line 325: Thanks - I need to check this, but I think this change would have the side-effect of calling `device_get` on any input `x2` that enters this function. The `__index__` method is implemented via `_forward_to_value`, and `x2._value` is a property that returns the numpy version of the array.

So the benefit of the original `if` statement is that for non-concrete arrays, the array is not pulled from device in the course of the check.

YouJiacheng: Good catch, I should fix it(at the cost of code complexity can't be reduced). But IIUC the `device_get` call will only happen for a `DeviceArray`: call `__index__` on Tracer will immediately raise an error.

hawkinsp: I think maybe a compromise would be to simply make the condition something like:

```
if isinstance(x2, int) or isinstance(core.get_aval(x2), core.ConcreteArray):
```

That avoids the `get_aval` call in the common case. (We can and should make `get_aval` faster, but that's not today's problem.)

PR #10431, opened 2022-04-23
[remove units] add partial_eval_jaxpr_nounits

mattjj:
No callers yet, but this is to be used for initial-style higher-order primitives (HOPs, like in lax/control_flow.py) to remove units. Follow-up PRs will switch the HOPs one-by-one to use this new function.

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -803,6 +803,60 @@ def partial_eval_jaxpr_nounits(
     jaxpr: ClosedJaxpr, unknowns: Sequence[bool],
     instantiate: Union[bool, Sequence[bool]],
   ) -> Tuple[ClosedJaxpr, ClosedJaxpr, List[bool]]:
+  """Unzip a jaxpr in two by data dependence into 'known' and 'unknown' parts.
+
+  That is, given a jaxpr and a sequence of booleans indicating which jaxpr
+  inputs (i.e. invars) are considered unknown, produce two jaxprs and a list of
+  booleans representing which of the original jaxpr's outputs are unknown (i.e.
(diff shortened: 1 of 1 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -785,88 +798,65 @@ def convert_envvars_to_constvars(jaxpr: Jaxpr, num_env_vars: int) -> Jaxpr:
 def _split_aval(unknown: bool, aval: AbstractValue) -> Tuple[AbstractValue, AbstractValue]:
   return (abstract_unit, aval) if unknown else (aval, abstract_unit)
 
-def partial_eval_jaxpr(jaxpr: ClosedJaxpr, unknowns: Sequence[bool],
-                       instantiate: Union[bool, Sequence[bool]],
-                       ) -> Tuple[ClosedJaxpr, ClosedJaxpr, Sequence[bool]]:
-  """Specializes a Jaxpr given an indication of which inputs are known.
```
froystig, line 791: Keep (some of) this docstring around?

froystig (approved):

jax/interpreters/partial_eval.py
mattjj: Well, I'd rather not document the deprecated one, and most of it is incorrect for the new one. I can add a docstring for the new function though...

mattjj: Done!

froystig: Nice. I've found it to be one of the more useful comments in this file, because of the example.

mattjj: Good suggestion.

PR #10432, opened 2022-04-23
[remove units] remove units from cond partial eval

mattjj:
(Includes a commit from #10431, so review that one first as we'll land it first.)


froystig (approved):
Makes sense!

PR #10433, opened 2022-04-23
[remove units] remove units from while_loop partial eval

mattjj:
The first commit is from #10431 and the second is from #10432, so ignore those commits here and review those PRs first. Only the final commit is new here.


froystig (approved):
Great PR sequence.

PR #10448, opened 2022-04-26
[remove-units] prevent scan partial eval from introducing units

mattjj:
Scan's partial evaluation rule is tricky!

Units are a bookkeping mechanism for partial evaluation. Even though they're only needed for partial evaluation, they currently exist in JAX tracing and jaxprs, meaning that while they may make partial evaluation bookkeeping marginally simpler (only in some cases!) everything else has to deal with units. As we added more consumers of e.g. jaxprs, dealing with units everywhere started to outweigh any benefits.

Do units really make partial evaluation bookkeeping simpler? Well, look no further than this PR for an example: because `scan` has special 'slots' for inputs and outputs (hence `num_consts`, `num_carry` etc), inserting units as placeholders meant that when performing partial evaluation these values didn't really need to change. But once we prune units, we need to keep track of how many of each kind of input and output is being pruned.

It wasn't so bad though. And I think this is the last control flow primitive to use them. We might be home free to clear out any last simple places units are introduced, at which point we can delete them...

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -1681,119 +1681,143 @@ def _prune_zeros(ts):
 def _scan_partial_eval(trace, *tracers, reverse, length, num_consts, num_carry,
                        jaxpr, linear, unroll):
   num_ys = len(jaxpr.out_avals) - num_carry
-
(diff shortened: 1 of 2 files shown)


froystig (approved):
!!

Tricky, but looks good!

Issue #10450, opened 2022-04-26
Autodiff with `numpy.linalg.qr` differs from autodiff with `scipy.linalg.qr`


pnkraemer:
Hi all! 

I am a little confused by the different implementations of QR decompositions in `jax.scipy.linalg` und `jax.numpy.linalg`.
If called with specific inputs, one admits a JVP, and the other one does not.

It is not unlikely that I missed something, but I suspect the reason is the following.
Both `jax.scipy.linalg` and `jax.numpy.linalg` point to `jax.lax.linalg.qr`, but differently: if one uses for example `mode="r"`, then the numpy version leads to `full_matrix=False`

https://github.com/google/jax/blob/04b6f15cdbf26163aae1ca73cd3422e6e1912df9/jax/_src/numpy/linalg.py#L481

and the scipy version to `full_matrix=True`

https://github.com/google/jax/blob/04b6f15cdbf26163aae1ca73cd3422e6e1912df9/jax/_src/scipy/linalg.py#L171

(Admittedly, it is probably hard to see from the permalinks... )

This is an issue for me, because the Jacobian-vector product of `jax.lax.linalg.qr` raises an error whenever `full_matrices` is True
https://github.com/google/jax/blob/04b6f15cdbf26163aae1ca73cd3422e6e1912df9/jax/_src/lax/linalg.py#L1176
and thus I have to be really careful with Jacobians in my code, which does rely quite heavily on QR-decompositions. 

Minimal working example:

```python

import jax
import jax.numpy as jnp
import jax.scipy as jsp

key = jax.random.PRNGKey(seed=2)
key, subkey = jax.random.split(key)
x = jax.random.normal(key, shape=(2, 2))
dx = jax.random.normal(subkey, shape=(2, 2))

# The values coincide
r_np = jnp.linalg.qr(x, mode="r")
r_sp = jsp.linalg.qr(x, mode="r")
assert jnp.allclose(r_np, r_sp)

# The numpy jvp seems to work, at least kind of...
_, dr = jax.jvp(lambda s: jnp.linalg.qr(s, mode="r"), (x,), (dx,))
dt = 1e-6
dr_ = (jnp.linalg.qr(x + dt * dx, mode="r") - jnp.linalg.qr(x, mode="r")) / dt
assert jnp.allclose(dr, dr_, atol=1e-3, rtol=1e-3) , (dr-dr_)  # with tol=1e-5 it does not work, but at least there are values...

# The scipy jvp fails:
[...]

PR #10451, opened 2022-04-26
lax.linalg.qr: allow jvp when m == n and full_matrices=True

jakevdp:
The JVP is fine in this case because the computation is the same as for `full_matrices=False`. Addresses #10450

### DIFF
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -671,8 +671,9 @@ class NumpyLinalgTest(jtu.JaxTestCase):
     self.assertTrue(np.all(
         norm(np.eye(k) - np.matmul(np.conj(T(lq)), lq)) < 5))
 
-    if not full_matrices and m >= n:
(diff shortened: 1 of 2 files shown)


tests/linalg_test.py
```diff
@@ -671,8 +671,9 @@ def compare_orthogonal(q1, q2):
     self.assertTrue(np.all(
         norm(np.eye(k) - np.matmul(np.conj(T(lq)), lq)) < 5))
 
-    if not full_matrices and m >= n:
-        jtu.check_jvp(jnp.linalg.qr, partial(jvp, jnp.linalg.qr), (a,), atol=3e-3)
+    if (not full_matrices and m >= n) or (full_matrices and m == n):
```
hawkinsp, line 674: Shouldn't this be?

`if m  <= n or not full_matrices:`

jakevdp: No, autodiff wil error for `m < n`

hawkinsp (approved):

PR #10458, opened 2022-04-27
Added random.orthogonal and random.unitary.

carlosgmartin:


jax/_src/random.py
```diff
@@ -14,7 +14,7 @@
 
 
 from functools import partial
-from typing import Any, Optional, Sequence, Union
+from typing import Any, Optional, Sequence, Union, Tuple
```
jakevdp, line 17: Please alphabetize imports

jax/_src/random.py
```diff
@@ -1595,3 +1595,50 @@ def threefry_2x32(keypair, count):
...
+  Args:
+    key: a PRNG key used as the random key.
+    n: an integer indicating the resulting dimension.
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified shape and dtype.
+  """
```
jakevdp, line 1616: Please call `_check_shape()` here to validate the shape; same below.

Also we should validate `n` with `core.concrete_or_error(operator.index, ...)`

jax/_src/random.py
```diff
@@ -1595,3 +1595,50 @@ def threefry_2x32(keypair, count):
...
+  z = normal(key, shape + (n, n), dtype)
```
jakevdp, line 1617: Since `check_shape` lets through arbitrary sequences, it's a bit safer to replace `shape + (n, n)` with `(*shape, n, n)`

jax/_src/random.py
```diff
@@ -1595,3 +1595,50 @@ def threefry_2x32(keypair, count):
...
+    key: a PRNG key used as the random key.
+    n: an integer indicating the resulting dimension.
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified shape and dtype.
+  """
+  a, b = normal(key, (2,) + shape + (n, n), dtype)
```
jakevdp, line 1640: same comments as above: validate `shape` and `n`, and use `(2, *shape, n, n)`

jakevdp (changes requested):
This is a great start! A few comments on the implementations below. Also before we merge this we'll need a few more things:

- [x] We need tests of these functions. You can model them after existing tests in `tests/random_test.py`
- [x] To ensure the functions are available in the `jax.random` module, we'll need to import them in `jax/random.py`
- [x] To ensure the new functions are in the HTML docs, we'll need to add them to the list in `docs/jax.random.rst`
- [x] Finally, we should add an entry to `CHANGELOG.md` announcing this new feature.

Thanks for working on this!

carlosgmartin:
@jakevdp Not sure why `pytest -n auto tests/random_test.py` says `RuntimeError: INVALID_ARGUMENT: shape has invalid element type:` [blank].

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
     is not of an integer type, matching the behavior of
     {func}`numpy.diag`. Previously non-integer `k` was silently
     cast to integers.
+  * {func}Added `jax.random.orthogonal` and `jax.random.unitary`.
```
jakevdp, line 47:
```suggestion
  * Added {func}`jax.random.orthogonal` and {func}`jax.random.unitary`.
```

jax/_src/random.py
```diff
@@ -1614,15 +1614,17 @@ def orthogonal(
   Returns:
     A random array with the specified shape and dtype.
   """
-  z = normal(key, shape + (n, n), dtype)
+  _check_shape("orthogonal", shape)
+  n = core.concrete_or_error(int, n, "The error occurred in jax.random.orthogonal()")
```
jakevdp, line 1618: Let's use `operator.index` instead of `int` - it will error instead of round if the input is a float.

jax/_src/random.py
```diff
@@ -1637,7 +1639,9 @@ def unitary(
   Returns:
     A random array with the specified shape and dtype.
   """
-  a, b = normal(key, (2,) + shape + (n, n), dtype)
+  _check_shape("unitary", shape)
+  n = core.concrete_or_error(int, n, "The error occurred in jax.random.unitary()")
```
jakevdp, line 1643: `operator.index`

tests/random_test.py
```diff
@@ -1032,6 +1032,38 @@ def testLogistic(self, dtype):
     for samples in [uncompiled_samples, compiled_samples]:
       self._CheckKolmogorovSmirnovCDF(samples, scipy.stats.logistic().cdf)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_n={}_shape={}_dtype={}"\
```
jakevdp, line 1036: For better test names, we usually use `jtu.format_shape_dtype_string`; see e.g. https://github.com/google/jax/blob/ab37701ce3e6973ea7be1d549ae4d0d7ca9ee826/tests/lax_numpy_test.py#L979

tests/random_test.py
```diff
@@ -1032,6 +1032,38 @@ def testLogistic(self, dtype):
...
+  def testOrthogonal(self, n, shape, dtype):
+    key = self.seed_prng(0)
+    q = random.orthogonal(key, n, shape, dtype)
+    assert q.shape == (*shape, n, n)
+    with jax.numpy_rank_promotion('allow'):
+      assert jnp.allclose(jnp.einsum('...ij,...jk->...ik', q, q.swapaxes(-2, -1)), jnp.eye(n), atol=1e-6)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_n={}_shape={}_dtype={}"\
+       .format(n, shape, dtype),
```
jakevdp, line 1053: `format_shape_dtype_string` here as well, please.

jax/_src/random.py
```diff
@@ -1595,3 +1595,54 @@ def threefry_2x32(keypair, count):
...
+
+  Args:
+    key: a PRNG key used as the random key.
+    n: an integer indicating the resulting dimension.
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified shape and dtype.
```
jakevdp, line 1615: It may be useful to spell-out what the output shape is. From the docstring alone, I don't think I'd guess that the output is of shape `(*shape, n, n)`. Same with the `unitary` docstring.

jakevdp (changes requested):
Looks great! Just a few last minor comments.

Also, once you've made the additional changes, please squash all changes into a single commit; see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests

carlosgmartin:
@jakevdp Any idea how to fix the aforementioned test error?

jax/_src/random.py
```diff
@@ -1595,3 +1596,54 @@ def threefry_2x32(keypair, count):
...
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array of shape (*shape, n, n) and specified dtype.
+  """
+  _check_shape("unitary", shape)
+  n = core.concrete_or_error(index, n, "The error occurred in jax.random.unitary()")
+  a, b = normal(key, (2, *shape, n, n), dtype)
+  z = a + b * 1j
```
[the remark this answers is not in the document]
jakevdp: Sure, a single function sounds good.

tests/random_test.py
```diff
@@ -1032,6 +1032,22 @@ def testLogistic(self, dtype):
...
+       "n": n,
+       "shape": shape,
+       "dtype": dtype}
+      for n in range(5)
+      for shape in [(), (5,), (10, 5)]
+      for dtype in jtu.dtypes.floating + jtu.dtypes.complex))
+  def testOrthogonal(self, n, shape, dtype):
+    key = self.seed_prng(0)
+    q = random.orthogonal(key, n, shape, dtype)
+    assert q.shape == (*shape, n, n)
```
jakevdp, line 1047: Let's add an assertion that `q.dtype` is equal to the requested dtype.

jax/_src/random.py
```diff
@@ -1595,3 +1596,29 @@ def threefry_2x32(keypair, count):
...
   return prng.threefry_2x32(keypair, count)
+
+def orthogonal(
+  key: KeyArray,
+  n: int,
+  shape: Sequence[int] = (),
+  dtype: DTypeLikeFloat = dtypes.float_
+) -> jnp.ndarray:
+  """Sample uniformly from the orthogonal group O(n). If the specified
+    dtype is complex, sample uniformly from the unitary group U(n).
```
jakevdp, line 1607: Docstrings should start with a single summary line, and then use a new paragraph for any additional summary, aligned with the left side of the triple quote. You can see examples of this in other functions in this file, e.g.  https://github.com/google/jax/blob/0b470361dac51fb4f5ab2f720f1cf35e442db005/jax/_src/random.py#L109-L122

tests/random_test.py
```diff
@@ -1032,6 +1032,23 @@ def testLogistic(self, dtype):
...
+    assert q.dtype == dtype
```
jakevdp, line 1048: For both of these, please use `self.assertEqual` rather than a raw assert

tests/random_test.py
```diff
@@ -1032,6 +1032,23 @@ def testLogistic(self, dtype):
...
+    with jax.numpy_rank_promotion('allow'):
+      assert jnp.allclose(jnp.einsum('...ij,...jk->...ik', q, jnp.conj(q).swapaxes(-2, -1)), jnp.eye(n, dtype=dtype), atol=1e-6)
```
jakevdp, line 1050: Please use `self.assertAllClose`

jakevdp (changes requested):

jax/_src/random.py
```diff
@@ -1595,3 +1596,30 @@ def threefry_2x32(keypair, count):
...
+
+  Args:
+    key: a PRNG key used as the random key.
+    n: an integer indicating the resulting dimension.
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array of shape (*shape, n, n) and specified dtype.
```
jakevdp, line 1618: This line is the source of the readthedocs error
```suggestion
    A random array of shape `(*shape, n, n)` and specified dtype.
```

jakevdp:
I'm not sure how fast the empty qr bug is going to be fixed; if you want to land this sooner, you could catch the `n == 0` case and return the empty matrix directly.

jakevdp (approved):
Looks great, thanks!

jakevdp:
QR issue should be fixed in https://github.com/google/jax/pull/10495

tests/random_test.py
```diff
@@ -1032,6 +1032,27 @@ def testLogistic(self, dtype):
...
+  def testOrthogonal(self, n, shape, dtype):
+    key = self.seed_prng(0)
+    q = random.orthogonal(key, n, shape, dtype)
+    self.assertEqual(q.shape, (*shape, n, n))
+    self.assertEqual(q.dtype, dtype)
+    with jax.numpy_rank_promotion('allow'):
+      self.assertAllClose(
+        jnp.einsum('...ij,...jk->...ik', q, jnp.conj(q).swapaxes(-2, -1)),
+        jnp.broadcast_to(jnp.eye(n, dtype=dtype), (*shape, n, n)),
+        atol=1e-6,
```
jakevdp, line 1053: This test is failing on TPU backends; I suspect it might be because you set `atol` to a constant, which overrides the default behavior of adjusting the tolerance to fit the particular backend and dtype. Can we try removing this?

We may need to adjust the tolerance further for TPU backends, I'm not sure.

jakevdp:
Thanks!

PR #10466, opened 2022-04-27
[sparse] implement sparse rule for lax.concatenate_p

jakevdp:
The cool thing is that because we've already implemented `broadcast_in_dim`, this makes the sparsify transform able to handle a number of numpy APIs, e.g. `jnp.concatenate`, `jnp.stack`, `jnp.vstack`, `jnp.dstack`, `jnp.hstack`, etc.


froystig (approved):
nice!

PR #10468, opened 2022-04-27
[remove-units] prevent remat's partial eval from introducing units

mattjj:
Even though 'old' remat will someday soon be replaced by 'new' remat in ad_checkpoint.py, we want to get rid of units first so we need to update the old thing. (Almost paradoxically, one of the main reasons to get rid of units is to make upgrading to 'new' remat easier...)

Nothing surprising here: we just had to update remat's partial eval rule from using trace_to_jaxpr to use trace_to_jaxpr_nounits, and then follow up on all the consequences.


froystig (approved):

froystig:
Just add `_nounits`!

PR #10469, opened 2022-04-28
[remove-units] prevent pjit partial eval from dealing with units

mattjj:
`pjit`'s partial eval rule used `partial_eval_jaxpr`, which would produce jaxprs with units, even though it would then do some work to prune the units away.

This PR switches to using `partial_eval_jaxpr_nounits`, so the rule need not ever deal with units in the first place! It should solve the caching TODO as well.

### DIFF
--- a/jax/experimental/pjit.py
+++ b/jax/experimental/pjit.py
@@ -49,7 +49,8 @@ from jax._src.tree_util import prefix_errors
 from jax._src import util
 from jax._src.util import (
     HashableFunction, safe_map, safe_zip, wrap_name, wraps,
-    distributed_debug_log, split_list, cache, tuple_insert, weakref_lru_cache)
+    distributed_debug_log, split_list, cache, tuple_insert, weakref_lru_cache,
+    merge_lists)
 
 class _FromGdaSingleton:
   pass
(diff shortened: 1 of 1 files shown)


jax/experimental/pjit.py
```diff
@@ -803,21 +804,15 @@ def _pjit_partial_eval(trace, *in_tracers,
...
   unknown_outs = tuple(unknown_outs)
   known_outs = tuple(not uk for uk in unknown_outs)
-  num_residuals = len(raw_known_jaxpr.jaxpr.outvars) - len(unknown_outs)
+  num_residuals = len(res_avals)
 
   def keep_where(l, should_keep):
     return tuple(x for x, keep in zip(l, should_keep) if keep)
 
-  # Prepare the known jaxpr
-  # TODO(apaszke): map_jaxpr will break caching!
```
sharadmv, line 816: Did you kill this TODO as well?

jax/experimental/pjit.py
```diff
@@ -850,18 +845,16 @@ def keep_where(l, should_keep):
...
   residual_tracers = [trace.new_instantiated_const(residual) for residual in residual_vals]
 
-  # Prepare the unknown jaxpr
-  # TODO(apaszke): map_jaxpr will break caching!
-  unknown_jaxpr = raw_unknown_jaxpr.map_jaxpr(lambda jaxpr: pe._drop_vars(
-      jaxpr,
-      drop_ins=known_ins + (False,) * num_residuals,
-      drop_outs=known_outs))
+  # The convention of partial_eval_jaxpr_nounits is to place residual binders
+  # at the front of the jaxpr produced, so we move them here.
```
sharadmv, line 855: Maybe a tiny bit more here, e.g.:
```python
# The convention of partial_eval_jaxpr_nounits is to place residual binders
# at the front of the jaxpr produced, so we move them to the back since later
# we'll pass the residual tracers into the primitive after the unknown tracers.
```

sharadmv (approved):
LGTM

jax/experimental/pjit.py
mattjj: Yep! No more `map_jaxpr`, and the jaxpr munging functions here are all cached.

jax/experimental/pjit.py
mattjj: SG!

PR #10471, opened 2022-04-28
docs: replace all `git.io` links with their actual URLs

SukkaW:
All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

> We will be removing all existing link redirection from git.io on April 29, 2022.

Replace all git.io links with their actual URLs.


hawkinsp (changes requested):
I think I would revert the changes to the SVG files, because it breaks the rendering of the image. We would need to regenerate them instead.

Thanks!

PR #10489, opened 2022-04-29
[remove-units] don't use abstract_unit for dropvar avals

mattjj:
Partial evaluation using `JaxprTrace` constructs a jaxpr by first, during tracing, building a kind of bipartite graph between `JaxprTracer`s and [`Recipe`s](https://github.com/google/jax/blob/cdd11670959b930e01d23199fadb32b183f03ca5/jax/interpreters/partial_eval.py#L530). The `JaxprTracer`s have references to the `Recipe` which constructs them, and the common `JaxprEqnRecipe`s in turn have references to the `JaxprTracer`s corresponding to both their inputs and outputs. More precisely, to avoid strong reference cycles, the `JaxprEqnRecipe`s hold weak references to their output `JaxprTracer`s. That provides a memory optimization: when user code drops a reference to a `JaxprTracer`, e.g. by a variable going out of scope, its memory can be reclaimed just like a user may expect from Python's usual reference counting operational semantics, as if there were no AD going on.

Before this PR, that led to a funny typing issue: while unused `JaxprTracer`s could go out of scope before the trace completed, we still might have variables in the jaxpr being formed which correspond to those `JaxprTracer`s, and yet since we relied on the `JaxprTracer` instance to get the aval for use as a type annotation in the jaxpr, we wouldn't be able to annotate the variable's type! We used a catch-all fallback: we used a special variable `DropVar` (pretty-printed like `_`) and, since we didn't have a type available, used an `AbstractUnit` type, pretty-printed like `*`. So the jaxpr would get let-binders which pretty-printed as `_:*`.

This had a funny downstream effect on autodiff: transpose rules could get `ad_util.Zero(core.abstract_unit)` instances passed in as their cotangent arguments, despite that not matching the type of the tangent output!

Here we fixed the issue simply by tracking the output type separately on the `JaxprEqnRecipe`, rather than relying on reading it from the weakly referenced `JaxprTracer` instances.

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -554,7 +554,7 @@ class JaxprTracer(Tracer):
   @property
   def parents(self) -> Sequence[JaxprTracer]:
     if isinstance(self.recipe, JaxprEqnRecipe):
-      return self.recipe.invars
+      return self.recipe.in_tracers
     else:
       return []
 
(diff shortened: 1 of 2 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -655,53 +655,47 @@ def instantiate_const_at(trace: JaxprTrace, instantiate: bool, tracer):
...
                   recipe: JaxprEqnRecipe) -> core.JaxprEqn:
-  _, in_tracers, out_tracer_refs, primitive, params, effects, source_info = recipe
-  out_tracers = [t_ref() for t_ref in out_tracer_refs]
+  (_, in_tracers, out_tracer_refs, out_avals, prim, params, eff, src) = recipe
   invars  = [getvar(t) for t in in_tracers]
-  outvars = [DropVar(core.abstract_unit) if t is None
-             else cast(Var, getvar(t)) for t in out_tracers]
-  return new_jaxpr_eqn(invars, outvars, primitive, params, effects, source_info)
+  out_tracers = [t_ref() for t_ref in out_tracer_refs]
+  outvars = [DropVar(a) if t is None else getvar(t)  # type: ignore
```
sharadmv, line 696: Nice!

sharadmv (approved):
The less units the better!

PR #10490, opened 2022-04-29
Add jax.scipy.linalg.funm

ajcr:
This PR is to add `jax.scipy.linalg.funm`. It follows the same approach/algorithm as [`scipy.linalg.funm`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.funm.html) (including the same logic around dtype precision and utility methods).

### DIFF
--- a/jax/_src/third_party/scipy/linalg.py
+++ b/jax/_src/third_party/scipy/linalg.py
@@ -1,0 +1,77 @@
+import scipy.linalg
+
+from jax import jit, lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.linalg import norm
+from jax._src.numpy.util import _wraps
+from jax._src.scipy.linalg import rsf2csf, schur
+
+@jit
+def _algorithm_11_1_1(F, T):
+  # Algorithm 11.1.1 from Golub and Van Loan "Matrix Computations"
+  N = T.shape[0]
+  minden = jnp.abs(T[0, 0])
+
+  def _outer_loop(p, F_minden):
+    _, F, minden = lax.fori_loop(1, N-p+1, _inner_loop, (p, *F_minden))
+    return F, minden
+
+  def _inner_loop(i, p_F_minden):
+    p, F, minden = p_F_minden
+    j = i+p
+    s = T[i-1, j-1] * (F[j-1, j-1] - F[i-1, i-1])
+    T_row, T_col = T[i-1], T[:, j-1]
+    F_row, F_col = F[i-1], F[:, j-1]
+    ind = (jnp.arange(N) >= i) & (jnp.arange(N) < j-1)
+    val = (jnp.where(ind, T_row, 0) @ jnp.where(ind, F_col, 0) -
+            jnp.where(ind, F_row, 0) @ jnp.where(ind, T_col, 0))
+    s = s + val
+    den = T[j-1, j-1] - T[i-1, i-1]
+    s = jnp.where(den != 0, s / den, s)
+    F = F.at[i-1, j-1].set(s)
+    minden = jnp.minimum(minden, jnp.abs(den))
+    return p, F, minden
+
+  return lax.fori_loop(1, N, _outer_loop, (F, minden))
+
+_FUNM_LAX_DESCRIPTION = """\
+The array returned by :py:func:`jax.scipy.linalg.funm` may differ in dtype
+from the array returned by py:func:`scipy.linalg.funm`. Specifically, in cases
+where all imaginary parts of the array values are close to zero, the SciPy
+function may return a real-valued array, whereas the JAX implementation will
+return a complex-valued array.
+
+Additionally, unlike the SciPy implementation, when ``disp=True`` no warning
+will be printed if the error in the array output is estimated to be large.
+"""
+
+@_wraps(scipy.linalg.funm, lax_description=_FUNM_LAX_DESCRIPTION)
+def funm(A, func, disp=True):
+  A = jnp.asarray(A)
+  if A.ndim != 2 or A.shape[0] != A.shape[1]:
+    raise ValueError('expected square array_like input')
+
+  T, Z = schur(A)
+  T, Z = rsf2csf(T, Z)
+
+  F = jnp.diag(func(jnp.diag(T)))
+  F = F.astype(T.dtype.char)
+
+  F, minden = _algorithm_11_1_1(F, T)
+  F = Z @ F @ Z.conj().T
+
+  if disp:
+    return F
+
+  if F.dtype.char.lower() == 'e':
+    tol = jnp.finfo(jnp.float16).eps
+  if F.dtype.char.lower() == 'f':
+    tol = jnp.finfo(jnp.float32).eps
+  else:
+    tol = jnp.finfo(jnp.float64).eps
+
+  minden = jnp.where(minden == 0.0, tol, minden)
+  err = jnp.where(jnp.any(jnp.isinf(F)), jnp.inf, jnp.minimum(1, jnp.maximum(
+          tol, (tol / minden) * norm(jnp.triu(T, 1), 1))))
+
+  return F, err
--- a/jax/scipy/linalg.py
+++ b/jax/scipy/linalg.py
@@ -41,3 +41,7 @@ from jax._src.scipy.linalg import (
 from jax._src.lax.polar import (
   polar_unitary as polar_unitary,
 )
+
+from jax._src.third_party.scipy.linalg import (
+  funm as funm,
+)
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -1466,6 +1466,28 @@ class ScipyLinalgTest(jtu.JaxTestCase):
       self._CompileAndCheck(jsp.linalg.rsf2csf, args_maker)
 
   @parameterized.named_parameters(
+        jtu.cases_from_list({
+            "testcase_name":
+            "_shape={}_disp={}".format(jtu.format_shape_dtype_string(shape, dtype), disp),
+            "shape": shape, "dtype": dtype, "disp": disp
+        } for shape in [(1, 1), (5, 5), (20, 20), (50, 50)]
+          for dtype in float_types + complex_types
+          for disp in [True, False]))
+  # funm uses jax.scipy.linalg.schur which is implemented for a CPU
+  # backend only, so tests on GPU and TPU backends are skipped here
+  @jtu.skip_on_devices("gpu", "tpu")
+  def testFunm(self, shape, dtype, disp):
+      def func(x):
+        return x**-2.718
+      rng = jtu.rand_default(self.rng())
+      args_maker = lambda: [rng(shape, dtype)]
+      jnp_fun = lambda arr: jsp.linalg.funm(arr, func, disp=disp)
+      scp_fun = lambda arr: osp.linalg.funm(arr, func, disp=disp)
+      self._CheckAgainstNumpy(jnp_fun, scp_fun, args_maker, check_dtypes=False,
+                              tol={np.complex64: 1e-5, np.complex128: 1e-6})
+      self._CompileAndCheck(jnp_fun, args_maker)
+
+  @parameterized.named_parameters(
     jtu.cases_from_list({
         "testcase_name":
         "_shape={}".format(jtu.format_shape_dtype_string(shape, dtype)),
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -29,6 +29,7 @@ jax.scipy.linalg
    eigh_tridiagonal
    expm
    expm_frechet
+   funm
    inv
    lu
    lu_factor


ajcr:
Thanks for the review - I've put the utility methods that mirror the SciPy implementations in the third_party directory (I based this on the `signal_helper` module there).

The code that remains in `jax._src.scipy.linalg.py` still contains logic that is quite similar to that in SciPy (but expressed with JAX methods). For example, line 784 onwards where the correct `err` value is computed and printed. 

Do these parts of the implementation also need to go in the third party directory? Reading the license and code, I'm not quite sure where the line is.

jakevdp:
Any code derived from the scipy source should go in the third_party directory, even if it has been modified.

If you merely looked at the scipy function's API (i.e. which types of values it accepts and returns) and wrote your own implementation without referencing the scipy source code, then you can put it outside the third_party folder.

ajcr:
Makes sense - I've moved everything to the third_party directory.

On this topic, I should mention that I introduced `jax.scipy.linalg.rsf2csf` (which `funm` uses) in a [previous PR](https://github.com/google/jax/pull/10039) which was also based on the code in SciPy. I can move this to the third_party directory as well if that's appropriate.

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,127 @@
...
+  Returns
+  -------
+  out : real or complex array
+      Either the input array B or only the real part of the input array B.
+
+  """
+  if jnp.isrealobj(A) and jnp.iscomplexobj(B):
+    if tol is None:
+      tol = {0: feps*1e3, 1: eps*1e6}[_array_precision[B.dtype.char]]
+    if jnp.allclose(B.imag, 0.0, atol=tol):
```
jakevdp, line 43: This will not be compatible with jax transforms, so we should avoid it. We need to statically determine the output type. This will mean breaking with the scipy API, but that's necessary to make the API JAX-compatible in this case.

There are other examples of this, but I can't recall where they are right now.

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,127 @@
...
+  F, minden = _algorithm_11_1_1(F, T)
+  F = Z @ F @ Z.conj().T
+  F = _maybe_real(A, F)
+
+  tol = {0: feps, 1: eps}[_array_precision[F.dtype.char]]
+
+  if minden == 0.0:
+    minden = tol
+
+  if jnp.any(jnp.isinf(F)):
```
jakevdp, line 116: This is another statement that precludes transformation. Instead you can write it something like this:
```python
err = jnp.where(jnp.any(jnp.isinf(F)), jnp.inf, jnp.minimum(1, jnp.maximum(
            tol, (tol / minden) * norm(jnp.triu(T, 1), 1))))
```

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,127 @@
...
+    err = jnp.inf
+  else:
+    err = jnp.minimum(1, jnp.maximum(
+            tol, (tol / minden) * norm(jnp.triu(T, 1), 1)))
+
+  if disp:
+    if err > 1000*tol:
```
jakevdp, line 123: Again, we cannot branch on non-static values, so we should remove this block.

tests/linalg_test.py
```diff
@@ -1480,6 +1480,19 @@ def testRsf2csf(self, shape, dtype):
...
+            "testcase_name":
+            "_shape={}".format(jtu.format_shape_dtype_string(shape, dtype)),
+            "shape": shape, "dtype": dtype
+        } for shape in [(1, 1), (5, 5), (20, 20), (50, 50)]
+            for dtype in float_types + complex_types))
+  @jtu.skip_on_devices("gpu", "tpu")
+  def testFunm(self, shape, dtype):
+      rng = jtu.rand_default(self.rng())
+      args_maker = lambda: [rng(shape, dtype), lambda x: x**-2.718]
+      self._CheckAgainstNumpy(osp.linalg.funm, jsp.linalg.funm, args_maker)
```
jakevdp, line 1494: We should also call `_CompileAndCheck` once the above issues are addressed.

jakevdp (changes requested):
Looks good! There will be some work required here to make the implementation fully-compatible with JAX, but I think it's pretty close.

ajcr:
Thanks for the detailed feedback - I didn't fully appreciate at first that a JAX implementation of `funm` would be able to be compiled with its `func` argument (and that we'd be able to diverge from the SciPy API).

I've removed the `_maybe_real` call; as you say, the type of the output can't be determined through static analysis. I've added a note (`lax_description`) to the docstring to explain this and the fact that the error won't be printed when `disp=True`.

jax/_src/third_party/scipy/linalg.py
ajcr: Thanks - I've removed the `_maybe_real` logic.

tests/linalg_test.py
ajcr: I think I've now added this check in correctly (I followed the tests for `jnp.apply_along_axis` for inspiration).

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,96 @@
...
+from jax._src.numpy.linalg import norm
+from jax._src.numpy.util import _wraps
+from jax._src.scipy.linalg import rsf2csf, schur
+
+_array_precision = {'i': 1, 'l': 1, 'f': 0, 'd': 1, 'F': 0, 'D': 1}
+
+eps = jnp.finfo(float).eps
+feps = jnp.finfo(jnp.single).eps
+
+def _asarray_square(A):
```
jakevdp, line 14: Since this is used only once, and is fairly simple, maybe we should inline it?

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,96 @@
...
+
+_FUNM_LAX_DESCRIPTION = """\
+The array returned by :py:func:`jax.scipy.linalg.funm` may differ in dtype
+from the array returned by py:func:`scipy.linalg.funm`. Specifically, in cases
+where all imaginary parts of the array values are close to zero, the SciPy
+function may return a real-valued array, whereas the JAX implementation will
+return a complex-valued array.
+
+Additionally, unlike the SciPy implementation, when ``disp=True`` no warning
+will be printed if the error in the array output is estimated to be large.
```
jakevdp, line 73: We could possibly use the experimental [checkify](https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.checkify.check.html) work to implement something like this, but I'd probably leave that as a TODO.

jax/_src/third_party/scipy/linalg.py
```diff
@@ -0,0 +1,96 @@
...
+  F = jnp.diag(func(jnp.diag(T)))
+  F = F.astype(T.dtype.char)
+
+  F, minden = _algorithm_11_1_1(F, T)
+  F = Z @ F @ Z.conj().T
+
+  if disp:
+    return F
+
+  tol = {0: feps, 1: eps}[_array_precision[F.dtype.char]]
```
jakevdp, line 91: Won't this error for float16 and bfloat16?

Also, this seems needlessly abstracted. How about:
```
if F.dtype.char.lower() == 'f':
  tol = np.finfo(np.float32).eps
else:
  tol = np.finfo(np.float64).eps
```
with maybe another block for float16 types.

Having this logic defined in one place makes it much easier to understand the intent and effect of this line.

jax/scipy/linalg.py
```diff
@@ -41,3 +41,5 @@
 from jax._src.lax.polar import (
   polar_unitary as polar_unitary,
 )
+
+from jax._src.third_party.scipy.linalg import funm
```
jakevdp, line 45: We need to `import funm as funm` otherwise it will be treated as a private method by type checkers.

tests/linalg_test.py
```diff
@@ -1480,6 +1480,32 @@ def testRsf2csf(self, shape, dtype):
                               args_maker, tol=tol)
       self._CompileAndCheck(jsp.linalg.rsf2csf, args_maker)
 
+  @parameterized.named_parameters(
+        jtu.cases_from_list({
+            "testcase_name":
+            "_shape={}_disp={}".format(jtu.format_shape_dtype_string(shape, dtype), disp),
+            "shape": shape, "dtype": dtype, "disp": disp
+        } for shape in [(1, 1), (5, 5), (20, 20), (50, 50)]
+            for dtype in float_types + complex_types
```
jakevdp, line 1489: Keep all the `for` lines at the same indentation

tests/linalg_test.py
```diff
@@ -1480,6 +1480,32 @@ def testRsf2csf(self, shape, dtype):
...
+              for disp in [True, False]))
+  @jtu.skip_on_devices("gpu", "tpu")
```
jakevdp, line 1491: Why skip on gpu and tpu? Do we not expect this function to work correctly on those backends? If so we should mention this in the documentation. If not, we should test it.

tests/linalg_test.py
```diff
@@ -1480,6 +1480,32 @@ def testRsf2csf(self, shape, dtype):
...
+      def func(x):
+        return x**-2.718
+      rng = jtu.rand_default(self.rng())
+      args_maker = lambda: [rng(shape, dtype)]
+      if disp:
+        jnp_fun = lambda arr: jsp.linalg.funm(arr, func, disp=disp)
+        scp_fun = lambda arr: osp.linalg.funm(arr, func, disp=disp)
+      else:
+        # if disp is False then funm returns a tuple (array, error) and
+        # the error is ignored here given it is likely to differ
```
jakevdp, line 1502: We should not leave the error untested, particularly since it's not a straightforward computation. Why is it likely to differ? Can we test the value with an appropriate tolerance to make sure we're not returning entirely the wrong thing?

tests/linalg_test.py
```diff
@@ -1480,6 +1480,32 @@ def testRsf2csf(self, shape, dtype):
...
+        jnp_fun = lambda arr: jsp.linalg.funm(arr, func, disp=disp)[0]
+        scp_fun = lambda arr: osp.linalg.funm(arr, func, disp=disp)[0]
+      tol = 1e-06 if shape[0] < 50 else 1e-05
```
jakevdp, line 1505: Setting tol this way overrides the default, which adjusts tol based on the dtype of the output. See other examples where `tol` is set to a dict mapping dtype to tolerance.

tests/linalg_test.py
```diff
@@ -1480,6 +1480,32 @@ def testRsf2csf(self, shape, dtype):
...
+      self._CheckAgainstNumpy(jnp_fun, scp_fun, args_maker, check_dtypes=False, tol=tol)
+      self._CompileAndCheck(jnp_fun, args_maker, check_cache_misses=False)
```
jakevdp, line 1507: It would be best if we could pass this test with cache checks. To do this, the inner functions of the implementation would have to be defined at top level, rather than as closures.

jakevdp (changes requested):
Thanks for the updates - it's coming along! A few more comments below.

tests/linalg_test.py
ajcr: `funm` uses `jax.scipy.linalg.schur` and my understanding is that Schur decomposition is currently implemented only for a CPU backend.

This also affects other methods such as `jax.scipy.linalg.sqrtm` which currently rely on Schur decomposition (and could affect yet-to-be-implemented methods such as `logm` if the same algorithm as SciPy is followed). 

I can add a note to the docstring explaining this (perhaps we'd want something generic that can be attached to other methods that may rely on Schur decomposition?).

jakevdp: Makes sense. It would be worth adding a comment to that effect in the line where you skip the tests.

jakevdp:
The readthedocs failure is due to being based on an old `main` branch can you update `main` and rebase?

tests/linalg_test.py
ajcr: In my initial experimentatio, I thought I noticed a large discrepancy. I've checked again and the discrepancy is less than I feared - I'll make sure the test here checks the error as well as the array.

jax/_src/third_party/scipy/linalg.py
ajcr: Agreed - I followed the SciPy source too closely here and the abstraction is less useful in this JAX code.

ajcr:
I think I've addressed the most recent set of comments and rebased on main.

However there's one issue I'm still looking into which is puzzling me. I'm checking `funm` on a `float16` dtype array and I consistently see errors like the following:
```python
import jax
import numpy as np

a = np.array([[3,2,1],[4,2,1],[4,4,0]], dtype=np.float16)

jax.scipy.linalg.funm(a, np.exp)
```
results in
```
malloc_consolidate(): invalid chunk size
Aborted (core dumped)
```
If I add a `np.float16` test case and run the unit tests, I see
```
[...]
[ RUN      ] ScipyLinalgTest.testFunm_shape=float16[1,1]_disp=False
[  FAILED  ] ScipyLinalgTest.testFunm_shape=float16[1,1]_disp=False
[ RUN      ] ScipyLinalgTest.testFunm_shape=float16[20,20]_disp=True
corrupted size vs. prev_size
Fatal Python error: Aborted

Current thread 0x00007f6a7b9dc740 (most recent call first):
[...call stack for thread...]
Aborted (core dumped)
```
Perhaps this is just my local machine/setup at fault, but I'm trying understand what's causing it.

jakevdp:
Can you tell if the float16 error is coming from the JAX function or from the scipy function?

ajcr:
I think it might `jax.scipy.linalg.schur` that's problematic on `np.float16` dtypes (at least on my laptop).
```
Python 3.10.4 (main, Mar 24 2022, 16:14:25) [GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> import jax
>>> a = np.array([[3,2,1],[4,2,1],[4,4,0]], dtype=np.float16)
>>> jax.scipy.linalg.schur(a)
[DeviceArray([[-6.062e+00,  0.000e+00,  0.000e+00],
             [ 5.817e-05,  1.150e-05,  0.000e+00],
             [ 0.000e+00,  0.000e+00,  0.000e+00]], dtype=float16), DeviceArray([[0.   , 0.   , 0.   ],
             [0.   , 0.   , 1.984],
             [0.   , 0.   , 0.   ]], dtype=float16)]
>>> jax.scipy.linalg.schur(a)
free(): invalid next size (fast)
Aborted (core dumped)
```
(That Schur decomposition returned before the crash is very different to `scipy.linalg.schur(a)`.)

I've run the same code online in a colab notebook and it returned:
```
NotImplementedError: Unsupported dtype float16
```
(JAX version 0.3.4)

jakevdp:
Yeah, looks like a bug in the `schur` implementation for half-precision.

For now, I'd skip `float16` and `bfloat16` in the tests; I'll open a separate issue to report the schur error.

jakevdp:
The schur fix is in, though the fix is to raise a `NotImplementedError` rather than segfault. So for now you should limit the tests here to standard float & complex types.

jakevdp (approved):
Looks good!

PR #10495, opened 2022-04-29
Make lax.linalg.qr robust to zero-dimensional inputs

### LINKED ISSUE #10485: lax.linalg.qr fails for empty matrices
It should instead return empty `q` and `r` matrices. This arose in #10458

```python

In [2]: from jax import lax

In [3]: lax.linalg.qr(jnp.empty((0, 0)))
 ** On entry to SGEQRF parameter number  4 had an illegal value
 ** On entry to SORGQR parameter number  5 had an illegal value
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-ed0ed6237394> in <module>
----> 1 lax.linalg.qr(jnp.empty((0, 0)))

<snip>

~/github/google/jax/jax/_src/dispatch.py in backend_compile(backend, built_c, options)
    646   # we use a separate function call to ensure that XLA compilation appears
    647   # separately in Python profiling results
--> 648   return backend.compile(built_c, compile_options=options)
    649 
    650 # TODO(phawkins): update users.

RuntimeError: INVALID_ARGUMENT: shape has invalid element type: 
```

jakevdp:
Fixes #10485

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -1150,18 +1150,22 @@ def qr_impl(operand, full_matrices):
   return q, r
 
 def _qr_translation_rule(ctx, avals_in, avals_out, operand, *, full_matrices):
+  operand_aval, = avals_in
+  shape = operand_aval.shape
+  m, n = shape[-2:]
+  if m == 0 or n == 0:
+    return [_eye_like_xla(ctx.builder, avals_out[0]),
+            _zeros_like_xla(ctx.builder, avals_out[1])]
   return xops.QR(operand, full_matrices)
 
 def qr_abstract_eval(operand, full_matrices):
   if isinstance(operand, ShapedArray):
     if operand.ndim < 2:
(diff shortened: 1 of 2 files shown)


jax/_src/lax/linalg.py
```diff
@@ -1150,18 +1150,22 @@ def qr_impl(operand, full_matrices):
...
   return xops.QR(operand, full_matrices)
 
 def qr_abstract_eval(operand, full_matrices):
   if isinstance(operand, ShapedArray):
     if operand.ndim < 2:
       raise ValueError("Argument to QR decomposition must have ndims >= 2")
-    batch_dims = operand.shape[:-2]
-    m = operand.shape[-2]
-    n = operand.shape[-1]
+    *batch_dims, m, n = operand.shape
```
hawkinsp, line 1165: Cool I didn't know you could do this!

hawkinsp (approved):

PR #10497, opened 2022-04-29
fix scan dce rule

mattjj:
This fixes the scan DCE rule (there was a dumb bug!), but we disabled it anyway until we can write more tests.

We're not sure how we missed it, but to trigger it we just needed remat-of-scan possibly combined with higher-order AD.


froystig (approved):
We can turn the DCE rule back on once we have a couple more tests for it.

PR #10501, opened 2022-04-29
[remove-units] remove partial_eval_jaxpr (no callers!)

mattjj:


sharadmv (approved):
Wow it's really gone :O

PR #10522, opened 2022-05-02
[remove-units] remove units from partial_eval.py

mattjj:
After last week's changes, units are no longer traced or introduced into jaxprs in any way, so we don't need to use them in partial evaluation.

(Also there are some unrelated removals of dead code in maps.py.)

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -62,9 +62,9 @@ class PartialVal(tuple):
   """Partial value: either a known value or an unknown (abstract) value.
 
   Represented as a pair `(aval_opt, const)` of one of two kinds:
-  * `(None, <Constant>)` indicates a known value, either a Python regular
-    value, or a Tracer.
-  * `(<AbstractValue>, *)` indicates an unknown value characterized by an
+  * `(None, <Constant>)` indicates a known value, where the constant is either a
+    Tracer or satisfies `core.valid_jaxtype(const)`;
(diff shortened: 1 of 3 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -84,7 +84,7 @@ def known(cls, const: core.Value) -> PartialVal:
 
   @classmethod
   def unknown(cls, aval: AbstractValue) -> PartialVal:
-    return PartialVal((aval, core.unit))
+    return PartialVal((aval, None))
```
mattjj, line 87: This one line is the main intended change of this PR!

jax/interpreters/partial_eval.py
```diff
@@ -742,8 +742,8 @@ def getconstvar(c):
       consts[v] = recipe.val
     elif isinstance(recipe, Literal):
       t_to_var[id(t)] = recipe
-    elif recipe is core.unit:
-      t_to_var[id(t)] = core.unitvar
+    elif recipe is None:
+      assert False
```
mattjj, line 746: This case arose when we had a Tracer created via `JaxprTracer.new_const` but not "instantiated", i.e. not corresponding to an "unknown" value in the jaxpr being built, and that Tracer was then an output of the traced function. Hence why, before removing units, it would always correspond to a unit-valued output when the jaxpr was built. But given the remove-units changes from last week, we no longer have such placeholders for known value outputs.

sharadmv (approved):

PR #10546, opened 2022-05-03
jnp.unravel_index: avoid overflow for large dimension sizes

### LINKED ISSUE #10540: Unexpected behaviour in `jnp.unravel_index`
I'm getting erroneous values for `jnp.unravel_index`. For an index of `0`, some  shapes give different results than the expected `(0,0)`. Some examples:

```python
import jax.numpy as jnp

side = 10
jnp.unravel_index(0, (side, side))
# (DeviceArray(0, dtype=int32), DeviceArray(0, dtype=int32))

side = 201_995
jnp.unravel_index(0, (side, side))
# (DeviceArray(0, dtype=int32), DeviceArray(0, dtype=int32))

side = 201_996
jnp.unravel_index(0, (side, side))
# (DeviceArray(-1, dtype=int32), DeviceArray(130531, dtype=int32))

side = 205_000
jnp.unravel_index(0, (side, side))
# (DeviceArray(-1, dtype=int32), DeviceArray(82039, dtype=int32))

side = 500_000
jnp.unravel_index(0, (side, side))
# (DeviceArray(0, dtype=int32), DeviceArray(0, dtype=int32))
```

I'm using Jax version `0.3.6`.

jakevdp:
Fixes #10540

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -808,22 +808,22 @@ def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
 
 _UNRAVEL_INDEX_DOC = """\
 Unlike numpy's implementation of unravel_index, negative indices are accepted
-and out-of-bounds indices are clipped.
+and out-of-bounds indices are clipped into the valid range.
 """
 
 @_wraps(np.unravel_index, lax_description=_UNRAVEL_INDEX_DOC)
 def unravel_index(indices, shape):
   _check_arraylike("unravel_index", indices)
-  sizes = append(array(shape), 1)
-  cumulative_sizes = cumprod(sizes[::-1])[::-1]
-  total_size = cumulative_sizes[0]
-  # Clip so raveling and unraveling an oob index will not change the behavior
-  clipped_indices = clip(indices, -total_size, total_size - 1)
-  # Add enough trailing dims to avoid conflict with clipped_indices
-  cumulative_sizes = expand_dims(cumulative_sizes, range(1, 1 + _ndim(indices)))
-  clipped_indices = expand_dims(clipped_indices, axis=0)
-  idx = clipped_indices % cumulative_sizes[:-1] // cumulative_sizes[1:]
-  return tuple(idx)
+  shape = atleast_1d(shape)
+  if shape.ndim != 1:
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -808,22 +808,22 @@ def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
...
+  shape = atleast_1d(shape)
+  if shape.ndim != 1:
+    raise ValueError("unravel_index: shape should be a scalar or 1D sequence.")
+  out_indices = [None] * len(shape)
+  for i, s in reversed(list(enumerate(shape))):
+    indices, out_indices[i] = divmod(indices, s)
+  oob_pos = indices > 0
+  oob_neg = indices < -1
+  return tuple(where(oob_pos, s - 1, where(oob_neg, 0, i))
+               for s, i in zip(shape, out_indices))
```
apaszke, line 826: Didn't the old implementation support negative indices? It seems to be that way, based on the fact that clipping allowed values down to `-total_size`. We might not be testing that though!

apaszke (approved):
LGTM, though the old implementation was likely quite a bit faster. But more precision might be better!

In any case, index arithmetic in int32 is always a footgun, so we might want to think about alternative solutions anyway.

jax/_src/numpy/lax_numpy.py
YouJiacheng: `oob_neg` is true only when indices after all divmod are less than `-1`, which means that input indices are less than `-total_size`.

YouJiacheng: `oob_neg` is true only when indices after all divmod are less than -1, which means that input indices are less than `-total_size`.

jakevdp: The new implementation matches the handling of negative indices of the old implementation. We have some (limited) existing test coverage here: https://github.com/google/jax/blob/5c838d2f6ff26dbdf62055ad52ec84551957a3f1/tests/lax_numpy_test.py#L4231-L4234

jakevdp: I pushed an updated test in the second commit that has more complete coverage, replacing the two old tests. Keeping as two commits because it makes it clearer that the old tests also pass with the new changes.

PR #10548, opened 2022-05-03
If optimization_barrier is not available, disable the jax2tf tests.

pschuh:


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -2387,7 +2387,8 @@ def select_one_carry(new_c: TfVal, c: TfVal, c_aval: core.ShapedArray) -> TfVal:
                     multiple_results=True,
                     extra_name_stack="checkpoint")
 
-tf_impl[lax_control_flow.optimization_barrier_p] = tfxla.optimization_barrier
+if hasattr(tfxla, 'optimization_barrier'):
```
hawkinsp, line 2390: Perhaps add a TODO stating when we can remove this test? What's the minimum TF version for it? Same everywhere.

hawkinsp (approved):

PR #10549, opened 2022-05-03
Attach callback keepalive to JIT executable

sharadmv:
When staging out Python callbacks, we need to keep a reference to the Python callback alive to avoid it being garbage collected. Currently, we leak the callback into global list (this is also the approach of `host_callback`) but we can do a bit better by attaching a reference to the callback to the executable XLA computation. When the computation is gone, we no longer need the callback and it will be freed along with the executable.

This PR attaches callbacks to JIT executables and leaves a TODO for pmap executables.

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -290,7 +290,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
     return XlaComputation(
         name, None, True, None, None, jaxpr=jaxpr, consts=consts, device=device,
         in_avals=abstract_args, out_avals=out_avals, effects=jaxpr.effects,
-        kept_var_idx=kept_var_idx)
+        kept_var_idx=kept_var_idx, keepalive=None)
 
   if not _on_exit:
     log_priority = logging.WARNING if config.jax_log_compiles else logging.DEBUG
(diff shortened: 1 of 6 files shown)


jax/_src/dispatch.py
```diff
@@ -814,11 +820,12 @@ def xla_executable(self):
 
   @staticmethod
   def from_trivial_jaxpr(jaxpr, consts, device, in_avals, out_avals, effects,
-                         kept_var_idx) -> XlaCompiledComputation:
+      kept_var_idx, keepalive: Optional[Any]) -> XlaCompiledComputation:
```
froystig, line 823:
```suggestion
      kept_var_idx, keepalive: Optional[Any]) -> XlaCompiledComputation:
    assert keepalive is None
```
From what I can tell, we're accepting an optional `keepalive` here that will always be `None`, only in order to accommodate having all the compilation arguments ("`compile_args`") passed around in bulk in other parts in this file. If what I'm saying sounds right, what do you think about documenting this by adding this assertion?

froystig (approved):

jax/_src/dispatch.py
sharadmv: Yep that sounds right, adding the assertion.

PR #10553, opened 2022-05-04
Initial debug print implementation

sharadmv:
This PR adds an initial implementation of `debug_print`, a stage-able printing function. It uses the `emit_python_callback` functionality from the runtime, which currently only works on CPU.

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -1,0 +1,152 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Module for JAX debugging primitives and related functionality."""
+import enum
+import functools
+import sys
+
+from typing import Callable, Any
+
+from jax import core
+from jax import tree_util
+from jax.interpreters import ad
+from jax.interpreters import batching
+from jax.interpreters import mlir
+
+DebugEffect = enum.Enum('DebugEffect', ['PRINT', 'ORDERED_PRINT'])
+
+core.ordered_effects.add(DebugEffect.ORDERED_PRINT)
+mlir.lowerable_effects.add(DebugEffect.PRINT)
+mlir.lowerable_effects.add(DebugEffect.ORDERED_PRINT)
+
+# `debug_callback_p` is the main primitive for staging out Python callbacks.
+debug_callback_p = core.Primitive('debug_callback')
+debug_callback_p.multiple_results = True
+
+@debug_callback_p.def_impl
+def debug_callback_impl(*flat_args, callback: Callable[..., Any],
+    effect: DebugEffect, in_tree: tree_util.PyTreeDef):
+  del effect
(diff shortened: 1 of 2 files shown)


jax/_src/debugging.py
```diff
@@ -0,0 +1,149 @@
...
+
+from jax import core
+from jax import tree_util
+from jax.interpreters import ad
+from jax.interpreters import batching
+from jax.interpreters import mlir
+
+class DebugEffect(enum.Enum):
+  PRINT = enum.auto()
+  ORDERED_PRINT = enum.auto()
```
mattjj, line 29:
```suggestion
DebugEffect = enum.Enum('DebugEffect', ['PRINT', 'ORDERED_PRINT'])
```

sharadmv: That vertical space :rocket:

jax/_src/debugging.py
```diff
@@ -0,0 +1,147 @@
...
+  return result
+mlir.register_lowering(debug_callback_p, debug_callback_lowering,
+                       platform="cpu")
+
+def debug_callback(callback: Callable[..., Any],
+                   effect: DebugEffect, result_shapes: Any, *args, **kwargs):
+  """Calls a stageable Python callback.
+
+  `debug_callback` enables you to pass in a Python function that can be called
+  inside of a staged JAX program.
```
mattjj, line 110: Maybe say something about what the "debug" part of the name means, in contrast with a non-"debug" callback?

jax/_src/debugging.py
```diff
@@ -0,0 +1,147 @@
...
+
+  Args:
+    callback: A Python callable.
+    effect: A `DebugEffect`.
+    result_shapes: A Pytree of `jax.ShapeDtypeStruct`-like objects that match
```
mattjj, line 115: Instead of prescribing using `jax.ShapeDtypeStruct`, maybe suggest duck-typing, i.e. any leaf objects with `shape` and `dtype` attributes.

jax/_src/debugging.py
```diff
@@ -0,0 +1,147 @@
...
+      output of `callback`.
+      *args: The positional arguments to the callback.
+      **kwargs: The positional arguments to the callback.
+  Returns:
+    The results of `callback` applied to `*args` and `**kwargs`.
```
mattjj, line 120:
```suggestion
    The value of `callback(*args, **kwargs)`.
```

What are the use cases for a debug callback returning anything though?

jax/_src/debugging.py
```diff
@@ -0,0 +1,147 @@
...
+def _format_print_callback(fmt: str, *args, **kwargs):
+  sys.stdout.write(fmt.format(*args, **kwargs) + "\n")
+
+def debug_print(fmt: str, *args, ordered=False, **kwargs) -> None:
+  """Prints values and works in staged out JAX functions.
+
+  Args:
+    fmt: A format string, e.g. `"hello {x}"`, that will be used to format
+      input arguments.
+    *args: A list of positional JAX arguments.
```
mattjj, line 139: What's "JAX" mean here? Maybe arrays? Is there any difference with what's passed in `kwargs`? (I ask b/c the description of the kwargs doesn't mention "JAX".)

mattjj (approved):
Nice!

jax/_src/debugging.py
sharadmv: Done.

jax/_src/debugging.py
sharadmv: Fixed, removed use of "JAX".

jax/_src/debugging.py
sharadmv: None for now, but the callback API does support it.

jax/_src/debugging.py
sharadmv: Elaborated here a bit more.

PR #10566, opened 2022-05-04
[sparse] add bcoo_update_layout utility

jakevdp:
This is a more general utility that will replace things like `bcoo_add_batch_dim` and `_unbatch_bcoo`. `bcoo_update_layout` takes an arbitrary `BCOO` matrix and returns the same matrix represented with a different `n_batch`, `n_sparse`, and `n_dense` in its representation. In some cases such an update can result in very inefficient storage; for these cases the default is to raise a `SparseEfficiencyError`, but this can be silenced using the `on_inefficient` argument.

### DIFF
--- a/tests/sparse_test.py
+++ b/tests/sparse_test.py
@@ -72,7 +72,6 @@ class BcooDotGeneralProperties(NamedTuple):
       jtu.format_shape_dtype_string(self.rhs_shape, self.dtype),
       self.n_batch, self.n_dense, self.dimension_numbers)
 
-
 def _iter_subsets(s):
   return itertools.chain.from_iterable(itertools.combinations(s, n) for n in range(len(s) + 1))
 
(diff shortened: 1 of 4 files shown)


tests/sparse_test.py
```diff
@@ -1969,15 +1950,45 @@ def test_bcoo_unbatch(self, shape, dtype, n_batch, n_dense):
...
+      for n_batch_out in range(len(shape) + 1)
+      for n_dense_out in range(len(shape) + 1 - n_batch_out)))
+  def test_bcoo_update_layout(self, shape, dtype, n_batch, n_batch_out, n_dense, n_dense_out):
+    rng = rand_sparse(self.rng())
+    mat = sparse.BCOO.fromdense(rng(shape, dtype), n_batch=n_batch, n_dense=n_dense)
+    kwds = dict(n_batch=n_batch_out, n_dense=n_dense_out)
+    # TODO(jakevdp): in case of length-0 or length-1 shapes errors/warnings will not be raised.
+    if n_dense_out > n_dense or n_batch_out > n_batch:
+      with self.assertRaises(sparse.SparseEfficiencyError):
+        sparse.bcoo_update_layout(mat, **kwds)
```
froystig, line 1983: Minor: Also test explicit `on_inefficient='error'`? The current line more specifically tests that together with the fact that it's the default.

froystig (approved):

PR #10569, opened 2022-05-04
Enable effect lowering for while/for/scan

sharadmv:
This change adds a global `allowed_effects` allowlist to `lax._src.control_flow.py`. Scan and while check that the jaxpr effects are in the allowlist (TODO for cond). In the lowering, ordered effects have tokens threaded in and out of the HOP bodies.

Note that this doesn't allow effects in the `cond` of a while loop, which is a trickier case and will be implemented in a subsequent PR.

cc: @LenaMartens

### DIFF
--- a/jax/_src/lax/control_flow.py
+++ b/jax/_src/lax/control_flow.py
@@ -24,7 +24,7 @@ import inspect
 import itertools
 import operator
 import os
-from typing import Any, Callable, Optional, Sequence, Tuple, TypeVar, List
+from typing import Any, Callable, Optional, Sequence, Set, Tuple, TypeVar, List
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


jax/_src/lax/control_flow.py
```diff
@@ -321,18 +323,26 @@ def _create_jaxpr(init_val):
   _check_tree_and_avals("body_fun output and input",
                         body_tree, body_jaxpr.out_avals,
                         in_tree_children[0], init_avals)
-  if cond_jaxpr.effects or body_jaxpr.effects:
-    raise NotImplementedError('Effects not supported in `while`.')
+  disallowed_effects = set(
+      e for e in core.join_effects(cond_jaxpr.effects, body_jaxpr.effects)
+      if e not in allowed_effects)
```
mattjj, line 328: Does this work?

```suggestion
  effects = core.join_effects(cond_jaxpr.effects, body_jaxpr.effects)
  disallowed_effects = effects - allowed_effects
```

sharadmv: Nice yea.

mattjj (approved):
Beautiful!

PR #1949, opened 2020-01-06
Allow Var to be a key in a JAX tree

tomhennigan:

PR #10573, opened 2022-05-04
remove count attribute and total_ordering from core.Var

mattjj:
Originally we used the `Var.count` attribute to ensure Var instances were printed consistently regardless of context, even though only their object id was load-bearing. That is, `Var.count` was only used for pretty printing. (#1949 added a total_ordering on `Var` for reasons out of scope of JAX's core code. I'm going to figure out if that's still needed... Haiku tests all seem to pass without it.)

But #8019 revised our pretty-printing so as not to use `Var.count`. Instead it chose how to pretty-print `Var` instances based on their order of appearance in a jaxpr. That meant `Var.count` really wasn't useful anymore.

So this PR removes `Var.count`. Since we no longer have `Var.count`, we also don't need `core.gensym` to take an optional sequence of jaxprs, since that was just used to set the starting count index for new `Var`s.

In fact, `Var.__repr__` and `JaxprEqn.__repr__` were made confusing after #8019, since they could print variable names totally different from the names that would appear when the same `JaxprEqn` or `Var` objects were printed as part of a jaxpr. That is, before this PR we might have a jaxpr which printed like:

```python
import jax

def f(x):
  for _ in range(3):
    x = jax.numpy.sin(x)
  return x

jaxpr = jax.make_jaxpr(f)(3.)
print(jaxpr)
# { lambda ; a:f32[]. let
#     b:f32[] = sin a
#     c:f32[] = sin b
#     d:f32[] = sin c
#   in (d,) }

_, eqn, _ = jaxpr.jaxpr.eqns
print(eqn)
# a:f32[] = sin b
```

Notice the variable names in the equation pretty-print don't correspond to any in the jaxpr pretty-print!

So this PR changes `JaxprEqn.__repr__` and `Var.__repr__` to show `Var` object ids, and in general just do less formatting (which seems consistent with the spirit of `__repr__`):
```
JaxprEqn(invars=[Var(id=140202705341552):float32[]], outvars=[Var(id=140202705339584):float32[]], primitive=sin, params={}, effects=set(), source_info=SourceInfo(traceback=<jaxlib.xla_extension.Traceback object at 0x7f837c73d770>, name_stack=NameStack(stack=())))
```

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -200,9 +200,6 @@ class JaxprEqn(NamedTuple):
   effects: Effects
   source_info: source_info_util.SourceInfo
 
-  def __repr__(self):
-    return str(pp_eqn(self, JaxprPpContext(), JaxprPpSettings())).rstrip()
-
   def replace(self, *args, **kwargs):
     return self._replace(*args, **kwargs)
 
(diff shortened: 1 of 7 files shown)


froystig (approved):

jax/core.py
```diff
@@ -213,27 +210,18 @@ def new_jaxpr_eqn(invars, outvars, primitive, params, effects, source_info=None)
...
 
-  def __lt__(self, other):
-    if not isinstance(other, Var):
-      return NotImplemented
-    else:
-      return (self.count, self.suffix) < (other.count, other.suffix)
-
   def __repr__(self):
-    return _encode_digits_alphabetic(self.count) + self.suffix
+    return f'Var(id={id(self)}){self.suffix}:{self.aval.str_short()}'
```
froystig, line 224:
```suggestion
    suffix_fmt = f', suffix={self.suffix}' if self.suffix else ''
    return f'Var(id={id(self)}{suffix_fmt}):{self.aval.str_short()}'
```

I could see something like `Var(id=..., suffix=...)` being less surprising for nontrivial suffixes. One suffix we use in our cond internals is `_` and currently that would appear as `Var(id=...)_:aval`.

This is a minor comment and I leave it up to you.

froystig (approved):
Nice. This makes everything easier.

jax/core.py
mattjj: Oh great idea, will do.

mattjj:
Let's submit this after the NeurIPS deadline, on May 19th I believe.

mattjj:
This needs to get rolled back, in #10638. I'll try to land it again after the NeurIPS deadline.

PR #10536, opened 2022-05-03
add more scan dce rule tests, fix issues

mattjj:
* add an `instantiate` argument to `pe.dce_jaxpr`, needed for fixpoint in scan DCE rule (o/w can hit "fixpoint not reached")
* add more scan DCE testing, including checkin the jaxpr integrity and checking input/output behavior

~TODO: let's brainstorm more tests~

PR #10576, opened 2022-05-04
[new-remat] add scan rule for new remat

mattjj:
TODO:
* [x] extensive-to-intensive residuals optimization
* [x] ~forwarding optimization~ turns out it's not needed with new remat, b/c new remat doesn't pass inputs through jaxpr_known as residuals!
* [x] make sure we don't forward numpy.ndarrays though (i.e. make the fix analogous to #4526) **NOTE:** I only half-finished this, see [this TODO](https://github.com/google/jax/pull/10576/files#diff-c3f68aab0e05d9d2de62fd88802c96c19d9060f6d2500a7eb4266ee4139e717cR2374).
* [x] get lots of coverage by adapting existing tests
* [x] merge #10536 (this PR includes changes from that PR)
* [x] merge #10595 (this PR includes changes from that PR)
* [x] merge #10616 (this PR includes changes from that PR)
* [x] merge #10711 (this PR includes changes from that PR)
* [x] write new tests
* [ ] [eventually, probably not this PR] de-duplicate with `_scan_partial_eval` (the policy-less tracers-in-tracers-out analogue of the new `_scan_partial_eval_custom` function)

The main idea is that `_scan_partial_eval_custom` generalizes partial evaluation so that what gets staged out doesn't necessarily correspond to what is unknown. That is, in standard partial evaluation, we unzip into known and unknown parts, and the unknown parts get staged into a jaxpr. But with more general partial evaluation, parameterized by a policy, what gets staged out can in general be different from what's unknown: known things can get staged out too. The rule signature, accordingly, has distinct list-of-bool inputs and outputs to indicate known-ness and staged-ness.

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -388,6 +388,22 @@ def remat_vmap(axis_size, axis_name, main_type, args, dims, *, jaxpr, **params):
   return remat_p.bind(*consts, *args, jaxpr=jaxpr_batched, **params), out_dims
 batching.axis_primitive_batchers[remat_p] = remat_vmap
 
+# TODO(mattjj,sharadmv): test this more
+# TODO(mattjj,sharadmv): de-duplicate with pe.dce_jaxpr_call_rule
+def remat_dce(used_outputs: List[bool], eqn: core.JaxprEqn
+              ) -> Tuple[List[bool], Optional[core.JaxprEqn]]:
+  new_jaxpr, used_inputs = pe.dce_jaxpr(eqn.params['jaxpr'], used_outputs)
+  new_params = dict(eqn.params, jaxpr=new_jaxpr)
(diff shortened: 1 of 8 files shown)


jax/_src/ad_checkpoint.py
```diff
@@ -388,6 +388,21 @@ def remat_vmap(axis_size, axis_name, main_type, args, dims, *, jaxpr, **params):
   return remat_p.bind(*consts, *args, jaxpr=jaxpr_batched, **params), out_dims
 batching.axis_primitive_batchers[remat_p] = remat_vmap
 
+# TODO(mattjj,sharadmv): de-duplicate with pe.dce_jaxpr_call_rule
```
sharadmv, line 391: Also maybe add a TODO to test this

sharadmv (approved):

jax/_src/ad_checkpoint.py
mattjj: Done.

PR #10579, opened 2022-05-04
DOC: remove mentions of units

jakevdp:
Fixes readthedocs failure on main. I think this failure was missed in #10532


mattjj (approved):
Thank you!

PR #10581, opened 2022-05-05
Enable batching rule for `debug_print`

sharadmv:
The batching rule for `debug_print` unrolls the callback across the mapped axis to avoid enforcing rank-polymorphism for the callback and print.

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -20,10 +20,13 @@ from typing import Callable, Any
 
 from jax import core
 from jax import tree_util
+from jax import lax
+from jax._src import util
 from jax.interpreters import ad
 from jax.interpreters import batching
 from jax.interpreters import mlir
 from jax._src.lax import control_flow as lcf
+import jax.numpy as jnp
 
 DebugEffect = enum.Enum('DebugEffect', ['PRINT', 'ORDERED_PRINT'])
 
@@ -36,6 +39,8 @@ lcf.allowed_effects.add(DebugEffect.ORDERED_PRINT)
 # `debug_callback_p` is the main primitive for staging out Python callbacks.
 debug_callback_p = core.Primitive('debug_callback')
 debug_callback_p.multiple_results = True
+
+map, unsafe_map = util.safe_map, map
 
 @debug_callback_p.def_impl
 def debug_callback_impl(*flat_args, callback: Callable[..., Any],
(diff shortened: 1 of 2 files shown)


jax/_src/debugging.py
```diff
@@ -48,11 +53,21 @@ def debug_callback_abstract_eval(*flat_avals, callback: Callable[..., Any],
...
+  """Unrolls the debug callback across the mapped axis."""
+  axis_size = next(x.shape[i] for x, i in zip(args, dims)
+                   if i is not None)
+  def get_arg_at_dim(i, dim, arg):
+    if dim is None:
+      # Broadcast unmapped argument
+      return arg
+    return lax.index_in_dim(arg, i, axis=dim, keepdims=False)
+  outs = []
+  for i in range(axis_size):
```
mattjj, line 66: We could consider generating a rolled loop, i.e. a lax.map. That might avoid inducing large compile times. Can do that in a follow-up though.

jax/_src/debugging.py
```diff
@@ -48,11 +53,21 @@ def debug_callback_abstract_eval(*flat_avals, callback: Callable[..., Any],
...
-    effect: DebugEffect, in_tree: tree_util.PyTreeDef):
-  del flat_args, callback, effect, in_tree
-  # TODO(sharadmv): implement batching rule
-  raise NotImplementedError('Batching not supported for `debug_callback`.')
+def debug_callback_batching_rule(args, dims, **params):
+  """Unrolls the debug callback across the mapped axis."""
+  axis_size = next(x.shape[i] for x, i in zip(args, dims)
+                   if i is not None)
+  def get_arg_at_dim(i, dim, arg):
+    if dim is None:
```
mattjj, line 61:
```suggestion
    if dim is batching.not_mapped:
```

Doesn't matter much but someday we might want to replace the use of `None` with some other sentinel object.

jax/_src/debugging.py
sharadmv: Added a TODO

mattjj (approved):
Beautiful!

jax/_src/debugging.py
```diff
@@ -48,11 +53,23 @@ def debug_callback_abstract_eval(*flat_avals, callback: Callable[..., Any],
   del flat_avals, callback, in_tree
   return [], {effect}
 
-def debug_callback_batching_rule(*flat_args, callback: Callable[..., Any],
-    effect: DebugEffect, in_tree: tree_util.PyTreeDef):
-  del flat_args, callback, effect, in_tree
-  # TODO(sharadmv): implement batching rule
-  raise NotImplementedError('Batching not supported for `debug_callback`.')
+def debug_callback_batching_rule(args, dims, **params):
```
mattjj, line 56: James has made the point that this drops info about the axis name. But with custom batching rules around the corner, I think that gives the user a good way to change the default behavior if they want.

PR #10582, opened 2022-05-05
Enable AD rules for `debug_print`

sharadmv:
The AD rules for `debug_print` are just errors.

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -78,15 +78,16 @@ batching.primitive_batchers[debug_callback_p] = debug_callback_batching_rule
 def debug_callback_jvp_rule(*flat_args, callback: Callable[..., Any],
     effect: DebugEffect, in_tree: tree_util.PyTreeDef):
   del flat_args, callback, effect, in_tree
-  # TODO(sharadmv): implement jvp rule
-  raise NotImplementedError('JVP not supported for `debug_callback`.')
+  # TODO(sharadmv): link to relevant documentation when it exists
+  raise ValueError(
+      "JVP doesn't support debugging callbacks. "
+      "Instead, you can use them with `jax.custom_jvp` or `jax.custom_vjp`.")
 ad.primitive_jvps[debug_callback_p] = debug_callback_jvp_rule
 
 def debug_callback_transpose_rule(*flat_args, callback: Callable[..., Any],
     effect: DebugEffect, in_tree: tree_util.PyTreeDef):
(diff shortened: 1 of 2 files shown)


mattjj (approved):

jax/_src/debugging.py
```diff
@@ -51,25 +56,37 @@ def debug_callback_abstract_eval(*flat_avals, callback: Callable[..., Any],
...
+  outs = [jnp.stack(xs) for xs in zip(*outs)]
+  return outs, (0,) * len(outs)
 batching.primitive_batchers[debug_callback_p] = debug_callback_batching_rule
 
 def debug_callback_jvp_rule(*flat_args, callback: Callable[..., Any],
     effect: DebugEffect, in_tree: tree_util.PyTreeDef):
   del flat_args, callback, effect, in_tree
-  # TODO(sharadmv): implement jvp rule
-  raise NotImplementedError('JVP not supported for `debug_callback`.')
+  raise ValueError(
```
gnecula, line 81: Perhaps we can include in the error message a link to a documentation section, e.g., by using jax.errors

jax/_src/debugging.py
```diff
@@ -51,25 +56,37 @@ def debug_callback_abstract_eval(*flat_avals, callback: Callable[..., Any],
...
+      "JVP doesn't support debugging callbacks. "
+      "Instead, you can use them with `jax.custom_jvp` or `jax.custom_vjp`.")
 ad.primitive_jvps[debug_callback_p] = debug_callback_jvp_rule
 
 def debug_callback_transpose_rule(*flat_args, callback: Callable[..., Any],
     effect: DebugEffect, in_tree: tree_util.PyTreeDef):
   del flat_args, callback, effect, in_tree
-  # TODO(sharadmv): implement transpose rule
-  raise NotImplementedError('Transpose not supported for `debug_callback`.')
+  raise ValueError("Transpose doesn't support debugging callbacks.")
```
gnecula, line 89: I am wondering what are the cases when the user will see this message. For `vjp` and `grad` they will see the `jvp` message first. This error message makes sense if the user explicitly invokes transposition, e.g., `linear_transpose`. If there are other instances when this pops up, then maybe the error message should be more explanatory.

jax/_src/debugging.py
sharadmv: That's a good idea though we don't have the relevant documentation quite yet. I'll leave a TODO.

PR #10584, opened 2022-05-05
remove jnp.array case for handling buffers w/ aval=None

mattjj:
This functionality was added in #8134, but was superseded by later changes which ensured that we never produce DeviceArrays with their 'aval' property set to None (even when indexing ShardedDeviceArrays with integers, which used to be a problem case).

That is, I _think_ this code path is no longer needed.

### DIFF
--- a/tests/pmap_test.py
+++ b/tests/pmap_test.py
@@ -117,7 +117,12 @@ class PythonPmapTest(jtu.JaxTestCase):
 
   def testDeviceBufferToArray(self):
     sda = self.pmap(lambda x: x)(jnp.ones((jax.device_count(), 2)))
-    buf = sda.device_buffers[-1]
+
+    # Changed in https://github.com/google/jax/pull/10584 not to access
+    # sda.device_buffers, which isn't supported, and instead ensure fast slices
+    # of the arrays returned by pmap are set up correctly.
+    # buf = sda.device_buffers[-1]
+    buf = sda[-1]
 
     view = jnp.array(buf, copy=False)
     self.assertArraysEqual(sda[-1], view)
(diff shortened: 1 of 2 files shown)


tests/pmap_test.py
```diff
@@ -115,20 +115,6 @@ class PythonPmapTest(jtu.JaxTestCase):
...
-    view = jnp.array(buf, copy=False)
-    self.assertArraysEqual(sda[-1], view)
-    self.assertEqual(buf.device(), view.device())
-    self.assertEqual(buf.unsafe_buffer_pointer(), view.unsafe_buffer_pointer())
-
-    copy = jnp.array(buf, copy=True)
-    self.assertArraysEqual(sda[-1], copy)
-    self.assertEqual(buf.device(), copy.device())
-    self.assertNotEqual(buf.unsafe_buffer_pointer(), copy.unsafe_buffer_pointer())
-
```
jakevdp, line 131: Why not keep this test as a regression check? If it fails it still indicates a problem with the implementation, I think.

mattjj: I don't think we want to preserve this behavior which relies on `buf = sda.device_buffers[-1]`. That is, we don't want the objects returned by `sda.device_buffers[-1]` to be compatible with `jnp.array`, as we don't want people accessing `device_buffers` at all. (IIRC we used to return objects with `aval=None` from expressions like `sda[-1]`, and that was the problem, but that's now been fixed.)

WDYT? I could adapt the test to set `buf = sda[-1]` rather than `buf.device_buffers[-1]`.

jakevdp: Yes, I think that sounds good.

PR #10590, opened 2022-05-05
jax.scipy.sparse.linalg: support sparse matrices as operators

jakevdp:
Rather than explicitly testing for sparse matrix types, I chose to instead test for a `__matmul__` attribute indicating compatibility with the `@` operator. This will include any of the sparse matrix objects implemented in JAX.


shoyer (approved):
Looks great, thank you!

PR #10594, opened 2022-05-05
api_util: make shaped_abstractify respect raise_to_shaped

jakevdp:
This issue came up when tracing functions with sparse-aware avals for testing enhancements to `jax.experimental.sparse`.

The test change is necessary because the new code path will not canonicalize the dtype of the aval; I don't think this is necessary in practice, but the pjit test with an explicitly constructed aval was falling afoul of this.

### DIFF
--- a/jax/_src/api_util.py
+++ b/jax/_src/api_util.py
@@ -318,7 +318,8 @@ def _dtype(x):
 
 def shaped_abstractify(x):
   try:
-    return core.raise_to_shaped(core.get_aval(x))
+    return core.raise_to_shaped(
+      x if isinstance(x, core.AbstractValue) else core.get_aval(x))
   except TypeError:
     pass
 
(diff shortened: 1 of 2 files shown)


yashk2810 (approved):

jax/_src/api_util.py
```diff
@@ -322,6 +322,11 @@ def shaped_abstractify(x):
   except TypeError:
     pass
 
+  try:
```
yashk2810, line 325: Why not use isinstance here and above instead of try;except?

jakevdp: Just force of habit from years of Python duck-typing 😁

yashk2810: Can you change it to use isinstance? It'll be faster I think!

Without needing to execute and get an error, you can check up-front and then execute (that in my mind seems faster but we can measure that).

jakevdp: Sure I'll give that a try

PR #10595, opened 2022-05-05
make core_test.py pass with core.call

mattjj:
This PR just makes a trivial fix in jax/interpreters/ad.py to make transposition work with `core.call` again.

I didn't actually set up any tests with `core.call`, though I did leave in some commented-out code which may be useful to that end in the near future.

(I may need to use `core.call` for something soon!)

### DIFF
--- a/tests/core_test.py
+++ b/tests/core_test.py
@@ -32,8 +32,10 @@ from jax import jvp, linearize, vjp, jit, make_jaxpr
 from jax.core import UnshapedArray, ShapedArray
 from jax.tree_util import (tree_flatten, tree_unflatten, tree_map, tree_reduce,
                            tree_leaves)
+from jax.api_util import flatten_fun_nokwargs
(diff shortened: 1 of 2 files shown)


tests/core_test.py
```diff
@@ -32,6 +32,7 @@
 from jax.core import UnshapedArray, ShapedArray
 from jax.tree_util import (tree_flatten, tree_unflatten, tree_map, tree_reduce,
                            tree_leaves)
+from jax.api_util import flatten_fun_nokwargs
```
sharadmv, line 35: Can remove

sharadmv (approved):

tests/core_test.py
mattjj: Hm it's used in the commented-out code... maybe instead of commenting it out I'll just leave it as an unused function.

Issue #10476, opened 2022-04-28
[Proposal] Define static arguments using typing annotations


JeppeKlitgaard:
I am really enjoying `jax`, but have found the ergonomics of declaring function arguments as static for use in `jit`, `pjit`, and `pmap` to be a little cumbersome and error-prone when refactoring.

In many (but not necessarily all) cases it is useful and intuitive to declare some arguments as static at the point of function declaration rather than later when mapping or jitting the function. This contrasts axis-mapping, which is always most intuitively done at the point of transforming the function using `xmap`, `vmap`, or `pmap`. 

Using [PEP 593](https://peps.python.org/pep-0593/) (Python 3.9+, backported in `typing_extensions`) it should be possible to declare static arguments in the function signature.

# Example

We have some function, which has both dynamic and static variables. Here `c_func` is "embarrasingly static" in that we strictly require it to be static at compile-time.

```python
def make_hypercube(a, b, c, c_func: Callable[..., Any]):
	arr = jnp.empty((3, 3))

	# Some other implementation details here
	arr += a
	arr *= b
	
	arr **= c_func(c)
	...

	return arr
```

Currently to `pmap` over `a` and `b`, one would do:

```python
pmake_hypercube = pmap(make_hypercube, in_axes=(0, 0, None, None), static_broadcasted_argnums=(3,))
```

This is not ideal in that changing the signature of `make_hypercube` would require us to change our transformation, bearing in mind that these may be in very different parts of the code. Additionally, when functions have many arguments, changing the `argnums` tuple during refactoring or experimentation becomes needlessly cumbersome and error-prone.

I would propose adding a PEP 593 annotation, which would complement (not replace) the current method of marking arguments as static. In this case our implementation becomes:

```python
from jax import pmap, Static

def make_hypercube(a, b, c, c_func: Static[Callable[..., Any]]):
	arr = jnp.empty((3, 3))

	# Some other implementation details here
	arr += a
	arr *= b
	
	arr **= c_func(c)
	...

	return arr

pmake_hypercube = pmap(make_hypercube, in_axes=(0, 0, None, None))
```

# Implementation

While the PEP and Python documentation go into more details on this, such an annotation would be implemented as:

```python
from typing import Annotated, TypeVar,, get_type_hints, get_args

T = TypeVar("T")

StaticAnnotation = object()
Static = Annotated[T, StaticAnnotation] 


def f(a: str, b: Static[dict[str, float]]):
    ...

def is_static(func, arg) -> bool:
[...]

PR #10603, opened 2022-05-06
Additional input validation for transformations

### LINKED ISSUE #10046: Proposal - jax.jit static_argnums and static_argnames raises warning if parameter doesn't exist
Currently, if a static_argnums or static_argname doesn't exist for the function, nothing happens
This can be an issue if a function is updated but the jax.jit not being updated

```
import jax
import functools

@functools.partial(jax.jit, static_argnames=('b', 'd'))
def f(a, b, c):
    return a + b + c

print(f(1., 2., 3.))

@functools.partial(jax.jit, static_argnums=(1, 3))
def g(a, b, c):
    return a + b + c

print(g(1., 2., 3.))
```

I propose that a warning is raised if an arg num or arg name doesn't exist such that code will still run if the arg num or name doesn't exist but the programmer is made aware of the issue (and can silence if necessary)
Is there a good reason for this not existing already? 

[x] I can't find a previous issue on this

### LINKED ISSUE #10601: `jit` input validation can lead to silently dynamic variables
`jax.jit` (and presumable other similar transformations) currently does not validate `static_argnames` and `static_argnums`, which leads to silent failures in some cases. These can be particularly frustrating to debug and the additional input validation can be done with relatively little overhead.

A particularly headache inducing example of this could be when used in combination with positional-only (or keyword-only) arguments.

# Examples:
```python
def f(dyn, stat):
	print(dyn)
	print(stat)
    ...

def g(dyn, stat, /):
	print(dyn)
	print(stat)
    ...

jf = jit(f, static_argnames=("stat",)
jf(1, 2)
> Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
> 2

jg = jit(g, static_argnames=("stat",)
jg(1, 2)
> Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
> Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
# This is somewhat expected, but should at least raise an error

jit(f, static_argnames=("some_arg_that_does_not_exist",)
# No error!

jit(f, static_argnums=(9,)
# No error!
```

# Solution
Input validation on `jit` and similar functions may be made stricter. This may require using `inspect` where possible.

I would be happy to try undertaking this.

JeppeKlitgaard:
Adds additional input validation for `*_argnames` and `*_argnums` arguments to `jax.jit`.

Previously `static_argnums` and `static_argnames` could easily lead to silently dynamic arguments. In fact, two such cases were found in the `jax` source code using the new test coverage (c477ce3).

Additional test coverage has been added.

Includes a small refactoring in `api.py` which introduces `_jit` to reduce code duplication (this leads to a number of small changes in various tests). The refactoring has the additional benefit of making experimentation with #10476 easier.

For discussion see issue: #10601

Other places that could use additional validation (just add plumbing to validation logic introduced by this PR)
- `jax.experiment.pjit`
- `jax.pmap`
- `jax.value_and_grad`
- `jax.custom_vjp`
- `jax.custom_jvp`
- `jax.hessian`
- `jax.jacrev`
- `jax.jacfwd`
- `jax.grad`

I will add validation for the functions mentioned above in a separate PR once/if this one has been merged.

# Handling of `*args` and `**kwargs`

In cases where variable positional and/or keyword arguments are used, we (rightly) assume that the function will know how to deal with these and thus it is safe to assume that for `*args` any `argnum` is valid. Similarly for `**kwargs` any `argname` is valid.

Importantly we still do some validation where possible. As an example, consider:
```python
def f(a, /, b, *, c): ...

jit(f, static_argnames=("a",))   # This will fail since we know that `a` must be positional

jit(f, static_argnums=(2,))  # This will fail since `c` must be keyword
```

# Additional improvements

`argnums` could be patched to contain positional-only arguments given as `static_argnames` using `inspect`.

Currently:
```python
def f(a, /, b, *, c):
    print(a, b, c)

ff = jit(f, static_argnames=("a", "b", "c"))  # with validation disabled
ff(1, 2, c=3)
> Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 2 3
# Expected: 1 2 3

ff2 = jit(f, static_argnames=("b", "c"), static_argnums=(0,))
ff2(1, 2, c=3)
> 1 Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 3
# Expected 1 2 3
```

This is due to the way `_infer_argnums_and_argnames` works, which might be worthwhile to change in a separate PR. See further discussion in #10614

Fix: #10601
Fix: #10046

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -32,6 +32,7 @@ import types
 from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
                     Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable,
                     List)
+from typing_extensions import Literal
 from warnings import warn
 
 import numpy as np
@@ -56,7 +57,7 @@ from jax._src.api_util import (
     flatten_fun, apply_flat_fun, flatten_fun_nokwargs, flatten_fun_nokwargs2,
     argnums_partial, argnums_partial_except, flatten_axes, donation_vector,
     rebase_donate_argnums, _ensure_index, _ensure_index_tuple,
-    shaped_abstractify, _ensure_str_tuple, argnames_partial_except)
+    shaped_abstractify, _ensure_str_tuple, argnames_partial_except, validate_argnames, validate_argnums)
 from jax._src.lax import lax as lax_internal
 from jax._src.lib import jax_jit
 from jax._src.lib import xla_bridge as xb
@@ -178,6 +179,7 @@ def _check_callable(fun):
     raise TypeError(f"Expected a function, got a generator function: {fun}")
 
 def _isgeneratorfunction(fun):
+  # TODO 3.9+: remove
   # re-implemented here because of https://bugs.python.org/issue33261
   while inspect.ismethod(fun):
     fun = fun.__func__
@@ -188,40 +190,35 @@ def _isgeneratorfunction(fun):
 _POSITIONAL_OR_KEYWORD = inspect.Parameter.POSITIONAL_OR_KEYWORD
 
 def _infer_argnums_and_argnames(
-    fun: Callable,
+    sig: inspect.Signature,
     argnums: Union[int, Iterable[int], None],
     argnames: Union[str, Iterable[str], None],
   ) -> Tuple[Tuple[int, ...], Tuple[str, ...]]:
   """Infer missing argnums and argnames for a function with inspect."""
   if argnums is None and argnames is None:
-    argnums = ()
-    argnames = ()
-  elif argnums is not None and argnames is not None:
+    return (), ()
+
+  if argnums is not None and argnames is not None:
     argnums = _ensure_index_tuple(argnums)
     argnames = _ensure_str_tuple(argnames)
+
+    return argnums, argnames
+
+  parameters = sig.parameters
+  if argnums is None:
+    assert argnames is not None
+    argnames = _ensure_str_tuple(argnames)
+    argnums = tuple(
+        i for i, (k, param) in enumerate(parameters.items())
+        if param.kind == _POSITIONAL_OR_KEYWORD and k in argnames
+    )
   else:
-    try:
-      signature = inspect.signature(fun)
-    except ValueError:
-      # In rare cases, inspect can fail, e.g., on some builtin Python functions.
-      # In these cases, don't infer any parameters.
-      parameters: Mapping[str, inspect.Parameter] = {}
-    else:
-      parameters = signature.parameters
-    if argnums is None:
-      assert argnames is not None
-      argnames = _ensure_str_tuple(argnames)
-      argnums = tuple(
-          i for i, (k, param) in enumerate(parameters.items())
-          if param.kind == _POSITIONAL_OR_KEYWORD and k in argnames
-      )
-    else:
-      assert argnames is None
-      argnums = _ensure_index_tuple(argnums)
-      argnames = tuple(
-          k for i, (k, param) in enumerate(parameters.items())
-          if param.kind == _POSITIONAL_OR_KEYWORD and i in argnums
-      )
+    argnums = _ensure_index_tuple(argnums)
+    argnames = tuple(
+        k for i, (k, param) in enumerate(parameters.items())
+        if param.kind == _POSITIONAL_OR_KEYWORD and i in argnums
+    )
+
   return argnums, argnames
 
 
@@ -332,15 +329,63 @@ def jit(
     DeviceArray([   0,    1,  256, 6561], dtype=int32)
   """
   if FLAGS.experimental_cpp_jit and not config.jax_dynamic_shapes:
-    return _cpp_jit(fun, static_argnums, static_argnames, device, backend,
+    return _jit(True, fun, static_argnums, static_argnames, device, backend,
                     donate_argnums, inline, keep_unused)
-  else:
-    return _python_jit(fun, static_argnums, static_argnames, device, backend,
-                       donate_argnums, inline, keep_unused, abstracted_axes)
+  return _jit(False, fun, static_argnums, static_argnames, device, backend,
+                      donate_argnums, inline, keep_unused, abstracted_axes)
 
+def _jit(
+    use_cpp_jit: bool,
+    fun: Callable,
+    static_argnums: Union[int, Iterable[int], None] = None,
+    static_argnames: Union[str, Iterable[str], None] = None,
+    device: Optional[xc.Device] = None,
+    backend: Optional[str] = None,
+    donate_argnums: Union[int, Iterable[int]] = (),
+    inline: bool = False,
+    keep_unused: bool = False,
+    abstracted_axes: Optional[Any] = None,
+  ) -> stages.Wrapped:
+  # Implemements common logic between CPP and Python backends
+  _check_callable(fun)
+
+  # Coerce input
+  donate_argnums = _ensure_index_tuple(donate_argnums)
+
+  try:
+    sig = inspect.signature(fun)
+  except ValueError:
+    # Some built-in functions don't support signature.
+    # See: https://github.com/python/cpython/issues/73485
+    # In this case no validation is done
+    static_argnums = () if static_argnums is None else _ensure_index_tuple(static_argnums)
+    static_argnames = () if static_argnames is None else _ensure_str_tuple(static_argnames)
+  else:
+    # Infer argnums and argnames according to docstring
+    static_argnums, static_argnames = _infer_argnums_and_argnames(
+        sig, static_argnums, static_argnames)
+
+    # Validation
+    validate_argnums(sig, static_argnums, "static_argnums")
+    validate_argnums(sig, donate_argnums, "donate_argnums")
+
+    validate_argnames(sig, static_argnames, "static_argnames")
+
+  # Compensate for static argnums absorbing args
+  donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)
+
+  if use_cpp_jit:
+    return _cpp_jit(fun, static_argnums=static_argnums, static_argnames=static_argnames,
+                    device=device, backend=backend,
+                    donate_argnums=donate_argnums, inline=inline, keep_unused=keep_unused)
+
+  return _python_jit(fun, static_argnums=static_argnums, static_argnames=static_argnames,
+                      device=device, backend=backend, donate_argnums=donate_argnums,
+                      inline=inline, keep_unused=keep_unused, abstracted_axes=abstracted_axes)
 
 def _prepare_jit(fun, static_argnums, static_argnames, donate_argnums,
                  args, kwargs):
+  # Validate donate_argnums
   if max(donate_argnums, default=-1) >= len(args):
     raise ValueError(
         f"jitted function has donate_argnums={donate_argnums} but "
@@ -362,22 +407,16 @@ PytreeOfAbstractedAxesSpec = Any
 
 def _python_jit(
     fun: Callable,
-    static_argnums: Union[int, Iterable[int], None] = None,
-    static_argnames: Union[str, Iterable[str], None] = None,
-    device: Optional[xc.Device] = None,
-    backend: Optional[str] = None,
-    donate_argnums: Union[int, Iterable[int]] = (),
-    inline: bool = False,
-    keep_unused: bool = False,
-    abstracted_axes: Optional[PytreeOfAbstractedAxesSpec] = None,
+    *,
+    static_argnums: Tuple[int, ...],
+    static_argnames: Tuple[str, ...],
+    device: Optional[xc.Device],
+    backend: Optional[str],
+    donate_argnums: Tuple[int, ...],
+    inline: bool,
+    keep_unused: bool,
+    abstracted_axes: Optional[PytreeOfAbstractedAxesSpec],
   ) -> stages.Wrapped:
-  _check_callable(fun)
-  static_argnums, static_argnames = _infer_argnums_and_argnames(
-      fun, static_argnums, static_argnames)
-  static_argnums = _ensure_index_tuple(static_argnums)
-  donate_argnums = _ensure_index_tuple(donate_argnums)
-  donate_argnums = rebase_donate_argnums(donate_argnums, static_argnums)
-
   @wraps(fun)
   @api_boundary
   def f_jitted(*args, **kwargs):
@@ -429,13 +468,14 @@ _cpp_jit_cache = jax_jit.CompiledFunctionCache()
 
 def _cpp_jit(
     fun: Callable,
-    static_argnums: Union[int, Iterable[int], None] = None,
-    static_argnames: Union[str, Iterable[str], None] = None,
-    device: Optional[xc.Device] = None,
-    backend: Optional[str] = None,
-    donate_argnums: Union[int, Iterable[int]] = (),
-    inline: bool = False,
-    keep_unused: bool = False,
+    *,
+    static_argnums: Tuple[int, ...],
+    static_argnames: Tuple[str, ...],
+    device: Optional[xc.Device],
+    backend: Optional[str],
+    donate_argnums: Tuple[int, ...],
+    inline: bool,
+    keep_unused: bool,
   ) -> stages.Wrapped:
   # An implementation of `jit` that tries to do as much as possible in C++.
   # The goal of this function is to speed up the time it takes to process the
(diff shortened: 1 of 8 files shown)


jax/_src/api.py
```diff
@@ -333,15 +335,67 @@ def jit(
...
   if FLAGS.experimental_cpp_jit and not config.jax_dynamic_shapes:
-    return _cpp_jit(fun, static_argnums, static_argnames, device, backend,
+    return _jit("cpp", fun, static_argnums, static_argnames, device, backend,
                     donate_argnums, inline, keep_unused)
   else:
-    return _python_jit(fun, static_argnums, static_argnames, device, backend,
+    return _jit("python", fun, static_argnums, static_argnames, device, backend,
                        donate_argnums, inline, keep_unused, abstracted_axes)
 
+def _jit(
```
JeppeKlitgaard, line 344: This refactoring is done to reduce code duplication in `_jit_cpp` and `_jit_python`.

This does lead to some small changes in tests

jax/_src/api.py
```diff
@@ -2364,37 +2406,18 @@ def _vjp_pullback_wrapper(cotangent_dtypes, cotangent_shapes,
   ans = fun(*args)
   return tree_unflatten(out_tree, ans)
 
+@overload
```
JeppeKlitgaard, line 2409: Since we have access to `typing_extensions`, no need for this complicated if statement.

In Python 3.8+ we have access to `Literal` anyway.

jax/_src/lax/qdwh.py
```diff
@@ -65,7 +65,7 @@ def _use_cholesky(u, params):
   return u
 
 
-@functools.partial(jax.jit, static_argnums=(1, 2, 3))
+@functools.partial(jax.jit, static_argnums=(1, 2))
```
JeppeKlitgaard, line 68: Bug found by new tests

jax/_src/scipy/linalg.py
```diff
@@ -146,7 +146,7 @@ def lu_factor(a, overwrite_a=False, check_finite=True):
 
 @_wraps(scipy.linalg.lu_solve,
         lax_description=_no_overwrite_and_chkfinite_doc, skip_params=('overwrite_b', 'check_finite'))
-@partial(jit, static_argnames=('trans', 'overwrite_a', 'check_finite'))
+@partial(jit, static_argnames=('trans', 'overwrite_b', 'check_finite'))
```
JeppeKlitgaard, line 149: Bug found by new tests

jax/_src/api.py
```diff
@@ -333,15 +335,67 @@ def jit(
...
+
+  # Infer argnums and argnames according to docstring
+  static_argnums, static_argnames = _infer_argnums_and_argnames(
+      fun, static_argnums, static_argnames)
+
+  # Coerce input
+  donate_argnums = _ensure_index_tuple(donate_argnums)
+
+  # Validation
+  validate_argnums(fun, static_argnums, "static_argnums")
```
hawkinsp, line 367: Since making validation stricter is liable to break user code, we may have to start by reporting these as warnings not errors.

Let me try running our internal presubmits with this PR, which include a lot of downstream JAX code, to get a sense of how bad the problem will be.

hawkinsp (commented):
This looks promising, thanks.

jax/_src/api.py
JeppeKlitgaard: That might be a good idea. On the other hand, invalidation only happens when called with genuinely invalid arguments and is very easily fixed. Could be frustrating if a library causes the error.

I don't know what the timeline for iterating on minor versions. My opinion would be that given JAX is still young and experimental, it would be fine to implement this as an exception rather than a warning if you issue a new minor version (or wait with merging this until a new minor version is appropriate). Given JAX is still 0.x.y, this would comply with SemVer afaik.

jax/_src/api_util.py
```diff
@@ -132,6 +133,99 @@ def __hash__(self):
...
+  """
+  Validate that the argnums are sensible for a given function.
+
+  Sensibly handles cases of `f(..., *args).`
+  """
+  if min(argnums, default=0) < 0:
+    raise ValueError(f"Jitted function has {argnums_name}={argnums}, "
+                     "but negative argnums are not supported.")
+
+  try:
```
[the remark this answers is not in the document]
JeppeKlitgaard: Absolutely. Since `_infer_argnums_and_argnames` and the validators don't require the function, I think the best option might be to change the `fun` argument to be the function signature. I'll implement this

jax/_src/api.py
hawkinsp: It's not looking too bad. I found that many tests broke because of this bug:
https://github.com/google/trajax/blob/67fd5ed4867914f1ce1ab78819551c12206a7773/trajax/optimizers.py#L753
but that was the main cause of most failures. But there were a handful of other bugs here and there.

In general, our policy is: https://jax.readthedocs.io/en/latest/api_compatibility.html

JeppeKlitgaard: Happy to change over to warnings, if you'd prefer. Given `trajax` is Google too, maybe it would be sufficient to to fix [google/trajax@67fd5ed/trajax/optimizers.py#L753](https://github.com/google/trajax/blob/67fd5ed4867914f1ce1ab78819551c12206a7773/trajax/optimizers.py#L753) and push a `trajax` release and then not change to warnings?

jax/_src/api_util.py
```diff
@@ -132,6 +133,99 @@ def __hash__(self):
...
+_POSITIONAL_ARGUMENTS = (
+  inspect.Parameter.POSITIONAL_ONLY,
+  inspect.Parameter.POSITIONAL_OR_KEYWORD
+)
+
+def validate_argnums(fun: Callable, argnums: Tuple[int, ...], argnums_name: str) -> None:
+  """
+  Validate that the argnums are sensible for a given function.
+
+  Sensibly handles cases of `f(..., *args).`
```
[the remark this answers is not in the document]
JeppeKlitgaard: I vaguely recall writing this at night and thinking: "this is a very bad docstring" and then heading to bed. Fixed now

JeppeKlitgaard:
@mattjj is there any reason for `allow_invalid` in https://github.com/google/jax/blob/8e4cf253dc047d23cbbcb34a57a99fb82427c943/jax/_src/api.py#L351 except backwards compatibility?

This PR aims to improve validation of 'argument annotations' (`argnums, argnames`) and performs validation before `_prepare_jit` is called. From what I can tell, it should be safe to remove `allow_invalid` with the caveat that this could break existing user code that has bugs (see discussion: https://github.com/google/jax/pull/10603#discussion_r868115498)

JeppeKlitgaard:
@hawkinsp changes since last time are primarily support for negative `argnums` as per #10669

jax/_src/api.py
```diff
@@ -189,40 +191,35 @@ def _isgeneratorfunction(fun):
...
+
+  if argnums is not None and argnames is not None:
     argnums = _ensure_index_tuple(argnums)
     argnames = _ensure_str_tuple(argnames)
+
+    return argnums, argnames
+
+  parameters = sig.parameters
+  if argnums is None:
+    assert argnames is not None
```
soraros, line 210: Nit: I think the assert is redundant at this point

JeppeKlitgaard: The assert is to aid `mypy`, as we get an error otherwise. Alternatively one could use a `typing.cast`, but I prefer the assert.

Further, this was already in place prior to this PR.

jax/_src/api.py
```diff
@@ -333,15 +330,78 @@ def jit(
...
+  ) -> stages.Wrapped:
+  # Implemements common logic between CPP and Python backends
+  _check_callable(fun)
+
+  # Coerce input
+  donate_argnums = _ensure_index_tuple(donate_argnums)
+
+  try:
+    sig = inspect.signature(fun)
+
```
hawkinsp, line 359: Nit: remove newline.

(In general we tend to like a pretty compact style.)

jax/_src/api.py
```diff
@@ -333,15 +330,78 @@ def jit(
...
+    return _cpp_jit(fun, static_argnums=static_argnums, static_argnames=static_argnames,
+                    device=device, backend=backend,
+                    donate_argnums=donate_argnums, inline=inline, keep_unused=keep_unused)
+
+  elif jit_implementation == "python":
+    return _python_jit(fun, static_argnums=static_argnums, static_argnames=static_argnames,
+                       device=device, backend=backend, donate_argnums=donate_argnums,
+                       inline=inline, keep_unused=keep_unused, abstracted_axes=abstracted_axes)
+
+  # Used only in in testing.
```
hawkinsp, line 391: I'm not sure which tests use this case? I'd rather not add it.

jax/_src/api.py
```diff
@@ -333,15 +335,67 @@ def jit(
...
-    return _cpp_jit(fun, static_argnums, static_argnames, device, backend,
+    return _jit("cpp", fun, static_argnums, static_argnames, device, backend,
                     donate_argnums, inline, keep_unused)
   else:
-    return _python_jit(fun, static_argnums, static_argnames, device, backend,
+    return _jit("python", fun, static_argnums, static_argnames, device, backend,
                        donate_argnums, inline, keep_unused, abstracted_axes)
 
+def _jit(
+    jit_implementation: Literal["cpp", "python", None],
```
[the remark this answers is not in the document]
hawkinsp: Yes, I'd prefer either the boolean or higher-order function approach.

(I have a dislike of string-based case statements where we can avoid them.)

jax/_src/lax/qdwh.py
hawkinsp, line 68: I think this may be fixed at head already; please sync up.

jax/_src/api.py
JeppeKlitgaard: This has now been changed over. Instead of using `jit_type` for parameterised tests, I have now implemented a tuple of 'jitters' in `test_utils` to still be able to easily parameterise tests

jax/_src/api.py
JeppeKlitgaard: Done

jax/_src/api.py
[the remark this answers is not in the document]
JeppeKlitgaard: This has now been fixed.

jax/_src/api.py
JeppeKlitgaard: Will do this later tonight

JeppeKlitgaard: Done.

jax/_src/api_util.py
```diff
@@ -132,6 +134,96 @@ def __hash__(self):
...
+  for param in sig.parameters.values():
+    if param.kind in _POSITIONAL_ARGUMENTS:
+      n_pos_args += 1
+
+    elif param.kind is inspect.Parameter.VAR_POSITIONAL:
+      # We can have any number of positional arguments
+      return
+
+  if (
+    -min(argnums, default=1) >= n_pos_args
```
YouJiacheng, line 159: should this be `-min(argnums, default=1) > n_pos_args`?
IIUC, `-n_pos_args` is the first argument.

YouJiacheng (changes requested):

jax/_src/api_util.py
YouJiacheng: And, I personally think that
```python
if not all(-n_pos_args <= argnum < n_pos_args for argnum in argnums):
```
looks better than
```python
if -min(argnums, default=1) > n_pos_args or max(argnums, default=-1) >= n_pos_args:
```

JeppeKlitgaard: I agree with your first comment.

With regards to the second comment: I don't like the introduction of a loop, even if this could be argued to be more readable. Further, I don't actually think it is more readable. To me, the original code reads "Are we below lower bound or above upper bound", which IMO reads quite nicely.

TODO
- [ ] Fix bound check (first comment)

YouJiacheng: Agreed, my suggestions is somewhat nit-pick. -- Another point: considering #10669 , is it safe to raise an error right now? @hawkinsp

JeppeKlitgaard: The short-circuit if check to get around default is quite neat. Definitely like that!

YouJiacheng:
Some idea:
If we do `arg` and `kwargs` inference using inspect tools following https://github.com/google/jax/issues/10614#issuecomment-1122688252 (there are some bugs in that comment), we might only need to catch error from bind?

JeppeKlitgaard:
> only

If I understand correctly, that would fall outside the scope of this PR. Further, it only addresses cases where bind is used (static) and not other argument annotations (donate).

This PR strictly deals with argument annotation validation at transformation time. I believe (but can't recall, and am traveling atm) that the binding process happens when the transformed function is called, not when the transformation is performed.

I am working on another PR that builds on top of this one which goes a little further, but this PR aims to be compatible and uncontroversial.

YouJiacheng:
I don't mean that the bind for call, I mean that if we use `Signature.bind_partial` to inference `argnums` and `argnames`.

jax/_src/test_util.py
```diff
@@ -845,6 +845,18 @@ def _CheckAgainstNumpy(self, numpy_reference_op, lax_op, args_maker,
                         atol=atol or tol, rtol=rtol or tol,
                         canonicalize_dtypes=canonicalize_dtypes)
 
+_CPP_JITTER = functools.partial(api._jit, True)
+_CPP_JITTER._name = "cpp"
+_PYTHON_JITTER = functools.partial(api._jit, False)
+_PYTHON_JITTER._name = "python"
+_NOOP_JITTER = lambda x: x
```
hawkinsp, line 852: I think I'd leave this out of the list and add it locally in the test cases that explicitly want to test "without jit"? After all, it isn't a jit implementation.

jax/_src/test_util.py
```diff
@@ -845,6 +845,18 @@ def _CheckAgainstNumpy(self, numpy_reference_op, lax_op, args_maker,
                         atol=atol or tol, rtol=rtol or tol,
                         canonicalize_dtypes=canonicalize_dtypes)
 
+_CPP_JITTER = functools.partial(api._jit, True)
```
hawkinsp, line 848: Perhaps JIT_IMPLEMENTATION instead of JITTER?

Same everywhere.

jax/_src/api_util.py
hawkinsp: I don't quite follow the question about #10669.

hawkinsp (commented):
This is looking good to me, but I have a final few nits.

jax/_src/api_util.py
JeppeKlitgaard: Changed over to short-circuit if statement:
`if argnums and (-min(argnums) > n_pos_args or max(argnums) >= n_pos_args)`

jax/_src/test_util.py
JeppeKlitgaard: All the parameterised tests use the `NOOP` jitter, so for now I think it should remain. Adding it in locally for every parameterised test would just lead to a lot of code duplication without any benefit, I think. Can always refactor this later if there are any cases where the no-op test case is not used.

jax/_src/test_util.py
JeppeKlitgaard: Will do

JeppeKlitgaard:
Implemented the requested changes.

Outstanding nit: https://github.com/google/jax/pull/10603#pullrequestreview-976051804

Let me know what you think @hawkinsp

JeppeKlitgaard:
Squashed to one commit and fixed a test.

tests/x64_context_test.py
```diff
@@ -81,12 +69,12 @@ def test_correctly_capture_default(self, jit, enable_or_disable):
...
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_jit={}".format(jit), "jit": jit}
-      for jit in ["python", "cpp", None]))
+      {"testcase_name": f"_jit={jit._name}", "jit": jit}
+      for jit in jtu.JIT_IMPLEMENTATION))
   def test_near_singular_inverse(self, jit):
     rng = jtu.rand_default(self.rng())
 
-    @partial(_maybe_jit, jit, static_argnums=1)
+    @partial(jit, static_argnums=1)
```
hawkinsp, line 77: This test case fails because static_argnums isn't a valid argument to `lambda x:x`.

tests/jax_jit_test.py
```diff
@@ -199,10 +200,11 @@ def test_arg_signature_of_value(self):
     self.assertTrue(signature.weak_type)
 
   def test_signature_support(self):
+    jit = partial(api._jit, "cpp")
```
hawkinsp, line 203: This looks incorrect — `api._jit` now takes a bool as its first argument.

hawkinsp (changes requested):

tests/jax_jit_test.py
JeppeKlitgaard: You're absolutely right. Apologies for missing this!

tests/x64_context_test.py
JeppeKlitgaard: Fixed

PR #10616, opened 2022-05-07
improve partial_eval_jaxpr_custom

mattjj:
No change in behavior here; just minor new features, cleanup, and more tests. The new features will be exercised by the `partial_eval_jaxpr_custom` rule for `scan` to come in a follow-up PR.

* add caching via `weakref_lru_cache`
* add `ensure_out_unknowns` and `ensure_out_inst` arguments, analogues of `instantiate` on e.g. `partial_eval_jaxpr`, `jvp_jaxpr`, etc (needed for fixpoints of loop primitives)
* add `inst_in` argument, which isn't strictly necessary because (as before this PR) we could always instantiate (i.e. stage out) all inputs and rely on DCE to prune unused ones, which nicely avoids ever having to forward an input through the known jaxpr via residuals, BUT we include because it gives the caller more flexibility
* better dce in `remat_partial_eval` (e.g. prune unused residuals)

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -938,8 +938,9 @@ def _remat_partial_eval(trace, _, f, tracers, params):
 
   if params['policy']:
     # unzip into jaxpr_known and jaxpr_unknown
-    jaxpr_known, jaxpr_unknown, out_unknowns, out_inst, _ = _partial_eval_jaxpr_custom(
-        jaxpr, in_unknowns, params['policy'])
+    jaxpr_known, jaxpr_unknown, out_unknowns, out_inst, _ = \
+        partial_eval_jaxpr_custom(jaxpr, in_unknowns, [True] * len(in_unknowns),
+                                  False, False, params['policy'])
     jaxpr_known, in_used_known = dce_jaxpr(jaxpr_known, [True] * len(jaxpr_known.outvars))
     _, used_outs_unknown = partition_list(out_inst, out_unknowns)
     jaxpr_unknown, in_used_unknown = dce_jaxpr(jaxpr_unknown, used_outs_unknown)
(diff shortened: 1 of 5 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -1064,17 +1076,23 @@ def ensure_instantiated(inst: bool, x: Atom) -> Atom:
...
+    if ensure_inst: ensure_instantiated(inst, x)
+  out_unknowns = map(op.or_, out_unknowns, ensure_out_unknowns)
+  out_inst     = map(op.or_, out_inst,     ensure_out_inst)
+
   ins_known, _ = partition_list(in_unknowns, jaxpr.invars)
   outs_known, _ = partition_list(out_unknowns, jaxpr.outvars)
   known_effects = core.join_effects(*(eqn.effects for eqn in known_eqns))
-  jaxpr_known = Jaxpr((), ins_known, [*outs_known, *residuals], known_eqns,
-                      known_effects)
+  jaxpr_known = Jaxpr(jaxpr.constvars, ins_known, [*outs_known, *residuals],
```
sharadmv, line 1087: Are the constvars just forwarded through into the known/staged jaxprs? I'm assuming the caller needs to DCE them away.

sharadmv (approved):
LGTM!

jax/interpreters/partial_eval.py
mattjj: Yeah, indeed I took the approach that both sides get constvars. If the caller wants to partition them between 'known' and 'staged' sides, they can be converted to invars (via `convert_constvars_jaxpr`). Similarly for DCE, constvars aren't pruned by `dce_jaxpr`, though the caller can just convert to invars. In both of these cases constvars just aren't affected.

We may want to revise that in the future, but this behavior was most consistent with the callers today.

Issue #10614, opened 2022-05-06
[Proposal] Consistent `argnums` and `argnames` parameters for transformations


JeppeKlitgaard:
Hey JAX team,

I have been trying to wrap my head around 'argument annotation` in JAX for a bit in the hopes of finding a more intuitive/consistent implementation, which has lead me to the big block of text below. I would be super keen to hear your thoughts as I try to dive deeper into the inner workings of JAX.

Lately there have been a number of issues requesting improvements to `*_argnums` and `*_argnames` parameters used in transformations in addition to other ergonomics improvements related to declaring which function arguments should be annotated with a given property. I figured it might be helpful to make an over-arching issue with the end goal of having a consistent, ergonomic way of specifying these parameters. Managing argument 'annotations' in transformations has definitely been one of the more frustrating experiences of learning JAX (which is otherwise entirely amazing, of course)

Related issues:
- #10539
- #10601 (#10603)
- #9452
- #10476 
- #9904
- #6871
- #1159

`jax.jit` correctly implements `static_argnames` even for cases with keyword-only arguments, which would suggest that it should be possible to add `argnames` equivalents to any function that currently only implements `argnums`.

An easier but less robust fix could be to map `argnames` to `argnums` using `inspect` (see discussion: #1159). This would likely not work for keyword-only arguments (though it might for things like `donate_arg...`?)

# Current shortcomings
Currently even the most robust implementation of the 'argument annotation' mechanism behaves in a very counter-intuitive way:
```python
def f(a, /, b, *, c):
    print(a, b, c)

jf = jit(f, static_argnames=("a", "b", "c"))
jf(1, 2, c=3)
> Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 2 3
# Expected: 1 2 3

jf2 = jit(f, static_argnames=("b", "c"), static_argnums=(0,))
jf2(1, 2, c=3)
> 1 Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 3
# Expected: 1 2 3

jf2(1, b=2, c=3)
> 1 2 3
# As expected

jf3 = jit(f, static_argnums=(0, 1, 2))
jf3(1, 2, c=3)
> 1 2 Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
# Expected: 1 2 3

jf3(1, b=2, c=3)
[...]

PR #10619, opened 2022-05-07
Better `argnums`/`argnames` inference

### LINKED ISSUE #10618: `static_argnames` of `jax.jit` does not correctly infer `argnums`
This is most easily shown by accessing `_infer_argnums_and_argnames` directly:

```python
def f(a, /, b, *, c):
	...

static_argnames = ("a", "b", "c")
_infer_argnums_and_argnames(f, tuple(), static_argnames)
> ((1,), ("a", "b", "c"))
# Expected: ((0, 1, 2,), ("a", "b", "c")
```

JeppeKlitgaard:
This PR adds additional tests for `infer_argnums_and_argnames` which has been moved from `api.py` to `api_utils.py`.

This is still a WIP and notably it changes the inference logic.

The previous logic is described in detail in [the documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html#jax.jit), but seems to be needlessly restrictive, since I cannot foresee a use-case where you would want:
> if both static_argnums and static_argnames are provided, inspect.signature is not used, and only actual parameters listed in either static_argnums or static_argnames will be treated as static.

Particularly since this gives somewhat unexpected/counter-intuitive behaviour when switching from a positional argument to a keyword argument:

```python
def f(a): ...

# We expect f(1) == f(a=1) always, but this is not the case with previous logic
```

Similarly there doesn't seem to be a reason to only look up `POSITIONAL_OR_KEYWORD` type parameters, which is what leads to the bug described in #10618.

TODO if this PR is deemed viable:
- [ ] change docstring of `jax.jit` to reflect new logic - will do this after first review

# Description of new logic
`infer_argnums_and_argnames` always fills `argnums` with parameters listed in `argnames` AND vice-versa. Only in the exceedingly rare event that signature inspection is not possible is this behaviour not true, in which case we return `argnums` and `argnames` unaltered as expected.

I am curious if there is a reason why one would prefer the old behaviour over the new?

# Other
Similarly to #10603, two cases of invalid use of `argnames`/`argnums` was found. They are also fixed by this PR in order to make tests pass.

This ties in with #10614

Fix: #10618

### DIFF
--- a/jax/_src/api_util.py
+++ b/jax/_src/api_util.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 import operator
 from functools import partial
-from typing import Any, Dict, Iterable, Tuple, Union, Optional
+from typing import Any, Callable, Dict, Iterable, Mapping, Tuple, Union, Optional
 
 import numpy as np
 
@@ -132,6 +133,45 @@ class _HashableWithStrictTypeEquality:
   def __eq__(self, other):
     return type(self.val) is type(other.val) and self.val == other.val
 
+
+_VARIABLE_PARAMETERS = (
+  inspect.Parameter.VAR_POSITIONAL,
+  inspect.Parameter.VAR_KEYWORD,
+)
+def infer_argnums_and_argnames(
+    fun: Callable,
+    argnums: Union[int, Iterable[int], None],
+    argnames: Union[str, Iterable[str], None],
+  ) -> Tuple[Tuple[int, ...], Tuple[str, ...]]:
+  """Infer missing argnums and argnames for a function with inspect."""
+  argnums = _ensure_index_tuple(argnums) if argnums is not None else ()
+  argnames = _ensure_str_tuple(argnames) if argnames is not None else ()
+
+  try:
+    signature = inspect.signature(fun)
+  except ValueError:
+    # In rare cases, inspect can fail, e.g., on some builtin Python functions.
+    # In these cases, don't infer any parameters.
+    return argnums, argnames
+
+  parameters = signature.parameters
+
+  # Resolve argnums from argnames
+  additional_argnums = tuple(
+    i for i, (k, param) in enumerate(parameters.items())
+    if param.kind not in _VARIABLE_PARAMETERS and k in argnames
+  )
+
+  # Resolve argnames from argnums
+  additional_argnames = tuple(
+    k for i, (k, param) in enumerate(parameters.items())
+    if param.kind not in _VARIABLE_PARAMETERS and i in argnums
+  )
+
+  argnums = tuple(set(argnums + additional_argnums))
+  argnames = tuple(set(argnames + additional_argnames))
+
+  return argnums, argnames
 
 def argnums_partial(f, dyn_argnums, args, require_static_args_hashable=True):
   dyn_argnums = _ensure_index_tuple(dyn_argnums)
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -55,7 +55,7 @@ from jax._src import source_info_util
 from jax._src import traceback_util
 from jax._src.api_util import (
     flatten_fun, apply_flat_fun, flatten_fun_nokwargs, flatten_fun_nokwargs2,
-    argnums_partial, argnums_partial_except, flatten_axes, donation_vector,
+    argnums_partial, argnums_partial_except, flatten_axes, donation_vector, infer_argnums_and_argnames,
     rebase_donate_argnums, _ensure_index, _ensure_index_tuple,
     shaped_abstractify, _ensure_str_tuple, argnames_partial_except)
 from jax._src.lax import lax as lax_internal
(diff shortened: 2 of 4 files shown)


jax/_src/api_util.py
```diff
@@ -133,6 +134,45 @@ def __eq__(self, other):
     return type(self.val) is type(other.val) and self.val == other.val
 
 
+_VARIABLE_PARAMETERS = (
+  inspect.Parameter.VAR_POSITIONAL,
+  inspect.Parameter.VAR_KEYWORD,
+)
+def infer_argnums_and_argnames(
```
JeppeKlitgaard, line 141: Will need tweaking once #10603 is merged

shoyer:
> Similarly there doesn't seem to be a reason to only look up POSITIONAL_OR_KEYWORD type parameters, which is what leads to the bug described in https://github.com/google/jax/issues/10618.

I agree, this seems worth fixing. I didn't really think about positional-only arguments when writing this function (we were not using a new enough version of Python to use them).

> `infer_argnums_and_argnames` always fills `argnums` with parameters listed in `argnames` AND vice-versa. Only in the exceedingly rare event that signature inspection is not possible is this behaviour not true, in which case we return `argnums` and `argnames` unaltered as expected.
> 
> I am curious if there is a reason why one would prefer the old behaviour over the new?

This original behavior was intentional. I think it's better because it's more explicit -- we only infer these arguments if they are not provided at all.

This is potentially a little more manual work, but is less magical.

JeppeKlitgaard:
> This original behavior was intentional. I think it's better because it's more explicit -- we only infer these arguments if they are not provided at all.
> 
> This is potentially a little more manual work, but is less magical.

@shoyer I found it super confusing not to have

JeppeKlitgaard:
> This original behavior was intentional. I think it's better because it's more explicit -- we only infer these arguments if they are not provided at all.
> 
> This is potentially a little more manual work, but is less magical.

@shoyer When I first ran into the following, I found the existing behaviour to be quite confusing:

```python
@partial(jit, static_argnames=("arg2", "arg3",))
def fun(arg1, arg2, arg3):
	...

fun(1, 2, arg3=3)  # here arg2 is not static!
```

Partly, this is because I haven't seen other places in the Python eco-system where behaviour changes based on whether an argument is given as positional or through a keyword.

shoyer:
> @shoyer When I first ran into the following, I found the existing behaviour to be quite confusing:
> 
> ```python
> @partial(jit, static_argnames=("arg2", "arg3",))
> def fun(arg1, arg2, arg3):
> 	...
> 
> fun(1, 2, arg3=3)  # here arg2 is not static!
> ```
> 
> Partly, this is because I haven't seen other places in the Python eco-system where behaviour changes based on whether an argument is given as positional or through a keyword.

What version of JAX are you running this on?

I tried to reproduce this:
```python
from jax import jit
from functools import partial

@partial(jit, static_argnames=("arg2", "arg3",))
def fun(arg1, arg2, arg3):
  print(arg1, arg2, arg3)
  return arg1

fun(1, 2, arg3=3)
```

This prints `Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 2 3`. Which indicates that arg2 is indeed static here.

The more surprising case might be something like:
```python
@partial(jit, static_argnames=("arg2",), static_argnums=(2,))
def fun(arg1, arg2, arg3):
  print(arg1, arg2, arg3)
  return arg1

fun(1, 2, arg3=3)
```
In which case none of the arguments are static. But I'm not sure this is really so surprising, given that it's inconsistent with the explicit list of static arguments.

JeppeKlitgaard:
> This prints Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)> 2 3. Which indicates that arg2 is indeed static here.

@shoyer you're absolutely right, this is what I get as well. I can't recall where I had an example that behaved like this – it seems like I must've misremembered it – it probably involved both `argnums` and `argnames` which is, as you pointed out, disallowed under the current implementation. Apologies for the confusion!

I still like the extra magic that this PR provides, but I respect that this might not be the direction that JAX want's to take. I'll make a new PR with the fixes discussed below if you confirm that it's a no on the magic inference

>> Similarly there doesn't seem to be a reason to only look up POSITIONAL_OR_KEYWORD type parameters, which is what leads to the bug described in https://github.com/google/jax/issues/10618.

> I agree, this seems worth fixing. I didn't really think about positional-only arguments when writing this function (we were not using a new enough version of Python to use them).

shoyer:
> I still like the extra magic that this PR provides, but I respect that this might not be the direction that JAX want's to take. I'll make a new PR with the fixes discussed below if you confirm that it's a no on the magic inference

Other JAX team members may have opinions here, but personally I don't like magic inference.

JeppeKlitgaard:
What initially lead me down this path was wanting to experiment with type annotations for 'argument annotation' (see: #10476). This would add a (in my opinion) significantly more ergonomic way of annotating arguments.

That would only make sense if the end user is allowed to 'mix' argument annotation styles such that the final annotation becomes the union of annotations given by type annotation, `argnums` and `argnames`. Perhaps this is a better description rather than calling it magic inference, which has a bit of a voodoo ring to it.

Example:

```python
# Assuming there is appetite for type annotations approach
def f(a, b, c: Static[Any]):
	...

jitted = jit(f)  # Just c is marked static
jitted = jit(f, static_argnames=('b',))  # c and b are marked static
jitted = jit(f, static_argnums=(0,), static_argnames=('b',))  # a, b, c marked static
```

JeppeKlitgaard:
I am currently working on a different PR that addresses some of these problems using a different approach – feel free to disregard this for now, although hopefully the discussion is interesting.

YouJiacheng:
As a user, I think current inference behavior is surprising: no inference done if both `argnums` and `argnames` are provided.
Especially, the absence of inference will cause the behavior depends on how user pass the arguments.

PR #10620, opened 2022-05-07
Officially support Python 3.10

JeppeKlitgaard:
This adds the appropriate classifiers to the `setup.py` package.

Additionally it adds Python 3.10 to the CI build matrix. This will use more build minutes. If this is not desirable, it would probably be better to drop Python 3.8 from the matrix since incompatibility is most likely to occur at oldest supported version (3.7) or latest supported version (3.10)

I have been using JAX with Python 3.10 for a while without any issues, but it would be nice to make the support 'official.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -1,5 +1,11 @@
 name: CI
 
+# We test all supported Python versions as follows:
+# - 3.7  : Documentation build
+# - 3.8  : Part of Matrix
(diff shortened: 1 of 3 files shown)


hawkinsp (approved):
Thanks, yes it is certainly our intention that Python 3.10 is fully supported.

(We may want to remove one of the existing CI builds because we also get some Python version coverage from our "Internal CI builds" which complement the Github Actions builds by, e.g., testing GPU support and building jaxlib from source.)

PR #10629, opened 2022-05-09
[ROCm] Fix unit test failures on ROCm.

rsanthanam-amd:
/cc @hawkinsp

### DIFF
--- a/jaxlib/gpu_prng.py
+++ b/jaxlib/gpu_prng.py
@@ -65,4 +65,4 @@ def _threefry2x32_lowering(prng, platform, keys, data):
       result_layouts=[layout] * 2)
 
 cuda_threefry2x32 = partial(_threefry2x32_lowering, _cuda_prng, "cuda")
-rocm_threefry2x32 = partial(_threefry2x32_lowering, _hip_prng, "rocm")
+rocm_threefry2x32 = partial(_threefry2x32_lowering, _hip_prng, "hip")


jax/_src/dlpack.py
```diff
@@ -57,6 +57,10 @@ def from_dlpack(dlpack):
     gpu_backend = xla_bridge.get_backend("cuda")
   except RuntimeError:
     gpu_backend = None
+    try:
```
hawkinsp, line 60: In this case is it actually correct to import the rocm backend?

We haven't implemented the relevant DLPack protocol:
https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/python/dlpack.cc;drc=4fca4330e69a333c4d6e78656a4897e92c9b5af6;l=252
so any tests involving DLPack *should* fail until we implement it.

Note `kDLROCM` is no the same as `kDLCUDA`:
https://github.com/dmlc/dlpack/blob/9351cf542ab478499294864ff3acfdab5c8c5f3d/include/dlpack/dlpack.h#L66

jaxlib/gpu_prng.py
```diff
@@ -65,4 +65,4 @@ def _threefry2x32_lowering(prng, platform, keys, data):
       result_layouts=[layout] * 2)
 
 cuda_threefry2x32 = partial(_threefry2x32_lowering, _cuda_prng, "cuda")
-rocm_threefry2x32 = partial(_threefry2x32_lowering, _hip_prng, "rocm")
+rocm_threefry2x32 = partial(_threefry2x32_lowering, _hip_prng, "hip")
```
hawkinsp, line 68: Thanks.

(I find it hard to keep "hip" and "rocm" straight: as I understand it: "ROCM" is the wider platform, and "HIP" is the CUDA-language compatibility part?)

rsanthanam-amd: Yes, that is correct.

jax/_src/dlpack.py
rsanthanam-amd: Actually, I am not sure.  But without this change, I was getting the following error:

 jaxlib.xla_extension.XlaRuntimeError: INVALID_ARGUMENT: DLPack tensor is on GPU, but no GPU backend was provided.

In this case, can you suggest an alternative resolution?

rsanthanam-amd: I reverted this change and will wait for the proper DLPack implementation.

Sorry for the misunderstanding.

hawkinsp: You didn't say exactly which test failed. If it's the test that roundtrips from JAX to itself via DLPack, that test would pass with this change, but shouldn't — we are lying about the hardware platform! However if you have TF or PyTorch installed in ROCm form then I would expect those tests to fail because we would report ourselves as CUDA.

hawkinsp (approved):

jax/_src/dlpack.py
rsanthanam-amd: It was array_interoperability_test.py that failed.

hawkinsp: Yes, but that test tests interoperability via DLPack with, e.g., JAX itself, TF, and PyTorch. I assume in your CI you only have JAX installed, so it's only testing JAX interoperating with itself via DLPack, which wouldn't reflect this kind of bug.

rsanthanam-amd: gotcha.  i will wait until dlpack is implemented for rocm.

Thanks for the clarification!

Issue #10596, opened 2022-05-06
Extremely slow when I jit compile a function on GPU.


xinyuxiao113:
When I use jit for a function with fft operation on GPU  as follows (Note that I use a global variable pulse here). It takes more than 10s to jit compile on GPU while 0.1s on CPU. I can not figure out the reason. Can you explain what happend ?

```python
import time
import jax
import numpy as np
import jax.numpy as jnp

pulse = jax.device_put(np.random.rand(8000))

def f():
    sigTx = jnp.fft.fft(pulse)
    return sigTx

t0 = time.time()
a = jax.jit(f, backend='gpu')()
t1 = time.time()
print('time cost on gpu:',t1-t0)


t0 = time.time()
a = jax.jit(f, backend='cpu')()
t1 = time.time()
print('time cost on cpu:',t1-t0)
```

![截屏2022-05-06 上午10 06 15](https://user-images.githubusercontent.com/55077744/167055137-aaff804f-05a6-45f1-997c-9f2a82d07912.png)

Surprisingly,  when I pass the global variable pulse as an argument into the function, the compile time is reduce to a normal level
```python
import time
import jax
import numpy as np
import jax.numpy as jnp

pulse = jax.device_put(np.random.rand(10000))

def f(pulse):
    return jnp.fft.fft(pulse)

t0 = time.time()
a = jax.jit(f, backend='gpu')(pulse)
t1 = time.time()
print('time cost on gpu:',t1-t0)


t0 = time.time()
a = jax.jit(f, backend='cpu')(pulse)
t1 = time.time()
print('time cost on cpu:',t1-t0)
```
![截屏2022-05-06 上午10 08 59](https://user-images.githubusercontent.com/55077744/167055138-ed452d66-94ec-4b2d-b09f-66a6dfa121d3.png)

xinyuxiao113:
Here is my environments.
cuda-11.4.2-gcc-11.2.0-rxy4qhm   
cudnn-8.2.4.15-11.4-gcc-11.2.0-a6q32ad   
gcc/11.2.0  
jax                                0.2.17  
jaxlib                             0.1.68

YouJiacheng:
This is related to constant-folding IMO.
```
2022-05-06 12:57:16.669751: E external/org_tensorflow/tensorflow/compiler/xla/service/slow_operation_alarm.cc:61] Constant folding an instruction is taking > 1s:

[...]

PR #10639, opened 2022-05-09
Don't create `DeviceArrays` for `onp.ndarray` indices

### LINKED ISSUE #2878: _index_to_gather produces DeviceArrays for int / onp.ndarray inputs

lgeiger:
This PR is a follow up to #10393 and prevents the creation of `DeviceArrays` when indexing with `onp.ndarray`. This prevents index computations to happen on accelerators if the index is an `onp.ndarray`.

```python
import numpy as np
from jax import make_jaxpr, numpy as jnp

x = jnp.arange(10)
idx = np.array([1, 2, 3])
print(make_jaxpr(lambda: x[idx])())
```
would previously generate:
```
{ lambda a:i32[3] b:i32[10]; . let
    c:bool[3] = lt a 0
    d:i32[3] = add a 10
    e:i32[3] = select_n c a d
    f:i32[3,1] = broadcast_in_dim[broadcast_dimensions=(0,) shape=(3, 1)] e
    g:i32[3] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
      fill_value=None
      indices_are_sorted=False
      slice_sizes=(1,)
      unique_indices=False
    ] b f
  in (g,) }
```
With this PR index normalisations happens in native Numpy so the Jaxpr becomes:
```
{ lambda a:i32[10] b:i32[3,1]; . let
    c:i32[3] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
      fill_value=None
      indices_are_sorted=False
      slice_sizes=(1,)
      unique_indices=False
    ] a b
  in (c,) }
```

If I haven't missed anything this should fix #2878.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3425,6 +3425,8 @@ def _normalize_index(index, axis_size):
   if issubdtype(_dtype(index), np.unsignedinteger):
     return index
   if core.is_constant_dim(axis_size):
+    if isinstance(index, np.ndarray):
+      return np.where(index < 0, index + axis_size, index)
     axis_size_val = _lax_const(index, axis_size)
   else:
     axis_size_val = lax.convert_element_type(core.dimension_as_value(axis_size),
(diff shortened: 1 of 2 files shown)


jakevdp:
I don't think we should do this, because this is a recipe for inadvertently generating very large compile-time constants that can slow down tracing a lot. Similar behavior that we'd like to get rid of can be seen here: #10596

YouJiacheng:
@jakevdp Hmmm, this PR can slow down tracing since it normalize idx during tracing.
But IIUC it won't generate larger compile-time constant than before:
only from un-normalized compile-time constant `np.ndarray` to normalized compile-time constant `np.ndarray` with same size.

YouJiacheng:
Nice explanation, I think it is good to keep this convention(avoid trace-time op on np array)!!👍😊

PR #10640, opened 2022-05-09
DOC: update myst-nb to v0.15.0

jakevdp:
This update changes a few configuration arguments and makes some metadata checks more strict, which is why this has to touch so many files.


docs/notebooks/Custom_derivative_rules_for_Python_code.ipynb
```diff
@@ -200,7 +200,7 @@
    "source": [
     "## Example problems\n",
     "\n",
-    "To get an idea of what problems `jax.custom_jvp` and `jax.custom_vjp` are meant to solve, let's go over a few examples. A more thorough introduction to the `jax.custom_jvp` and `jax.custom_vjp` APIs is in [the next section](#scrollTo=Dr0aNkBslfQf)."
+    "To get an idea of what problems `jax.custom_jvp` and `jax.custom_vjp` are meant to solve, let's go over a few examples. A more thorough introduction to the `jax.custom_jvp` and `jax.custom_vjp` APIs is in the next section."
```
sharadmv, line 203: Any reason this was changed?

sharadmv (approved):
Thanks!

docs/notebooks/Custom_derivative_rules_for_Python_code.ipynb
jakevdp: Yes, because myst-nb now errors for unrecognized hyperlink targets, and this is a Colab-specific target that is not recognized by myst-nb. It doesn't seem particularly important, so I thought dropping it would be fine.

PR #10642, opened 2022-05-09
Avoid generating trivial gathers when reversing array

lgeiger:
This PR avoids generating trivial gathers when reversing arrays via numpy indexing.
XLA will already remove the trivial gather. However, this change improves performance outside of a jit without hurting readability much.

```python
from jax import make_jaxpr, numpy as jnp

make_jaxpr(lambda x: x[::-1])(jnp.arange(10))
```
**before:**
```
{ lambda a:i32[1]; b:i32[10]. let
    c:i32[10] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(0,), collapsed_slice_dims=(), start_index_map=(0,))
      fill_value=None
      indices_are_sorted=True
      slice_sizes=(10,)
      unique_indices=True
    ] b a
    d:i32[10] = rev[dimensions=(0,)] c
  in (d,) }
```
**after:**
```
{ lambda ; a:i32[10]. let b:i32[10] = rev[dimensions=(0,)] a in (b,) }
```

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3775,11 +3775,16 @@ def _index_to_gather(x_shape, idx, normalize_indices=True):
           if stop is None or (not isinstance(stop, core.Tracer) and
               core.greater_equal_dim(stop, x_shape[x_axis])):
             stop = None
+        elif core.symbolic_equal_dim(step, -1):
+          step = -1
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Thanks, I think this change is probably safe! I'll run some internal tests then we can get this merged.

Issue #9659, opened 2022-02-22
[jax2tf] NotImplementedError: Call to scatter add cannot be converted with enable_xla=False


josephrocca:
Conversions for XlaScatter [are currently unsupported](https://github.com/google/jax/blob/main/jax/experimental/jax2tf/g3doc/no_xla_limitations.md#summary-table) when using `enabled_xla=False`. I'm wondering if support could be added?

Here's the full error that I'm seeing:
```
NotImplementedError                       Traceback (most recent call last)
/usr/local/lib/python3.7/dist-packages/jax/experimental/jax2tf/impl_no_xla.py in op(*arg, **kwargs)
     52 
     53   def op(*arg, **kwargs):
---> 54     raise _xla_disabled_error(name)
     55 
     56   return op

NotImplementedError: Call to scatter add cannot be converted with enable_xla=False.
```

Here is some code that reproduces this error:
```
!pip install --upgrade flax
!pip install git+https://github.com/josephrocca/transformers.git@patch-2
```
```python
import jax
from jax.experimental import jax2tf
from jax import numpy as jnp

import numpy as np
import tensorflow as tf

from transformers import FlaxCLIPModel

clip = FlaxCLIPModel.from_pretrained("openai/clip-vit-base-patch32")

def score(pixel_values, input_ids, attention_mask):
    pixel_values = jax.image.resize(pixel_values, (3, 224, 224), "nearest")
    inputs = {"pixel_values":jnp.array([pixel_values]), "input_ids":input_ids, "attention_mask":attention_mask}
    outputs = clip(**inputs)
    return outputs.logits_per_image[0][0][0]

score_tf = jax2tf.convert(jax.grad(score), enable_xla=False)

my_model = tf.Module()
my_model.f = tf.function(score_tf, autograph=False, jit_compile=True, input_signature=[
  tf.TensorSpec([3, 40, 40], tf.float32),
  tf.TensorSpec([1, 30], tf.int32),
  tf.TensorSpec([1, 30], tf.int32),
])

model_name = 'pixel_text_score_grad'
tf.saved_model.save(my_model, model_name, options=tf.saved_model.SaveOptions(experimental_custom_gradients=True))
```
Here's a public colab with that code: https://colab.research.google.com/drive/1HjpRXsa8Ue9KWiKbVVWUX6DlXoIYx2r8?usp=sharing You can click "Runtime > Run all" to see the error.

Thanks!

josephrocca:
[...]

PR #10653, opened 2022-05-10
[jax2tf] add rudimentary scatter support for enable_xla=False

oliverdutton:
I think this solves #9659.

Just checking now. It should work for simple indices where you index in only 1 dimension with your update. I've added quite a few guarding asserts. It still needs unit tests to be sure.

Checking #9659 test case now.

### DIFF
--- a/jax/experimental/jax2tf/impl_no_xla.py
+++ b/jax/experimental/jax2tf/impl_no_xla.py
@@ -792,10 +792,128 @@ def _dynamic_update_slice(operand, update, *start_indices,
 
 tf_impl_no_xla[lax.dynamic_update_slice_p] = _dynamic_update_slice
 
-tf_impl_no_xla[lax.scatter_p] = _unimplemented("scatter")
-tf_impl_no_xla[lax.scatter_min_p] = _unimplemented("scatter min")
-tf_impl_no_xla[lax.scatter_max_p] = _unimplemented("scatter max")
-tf_impl_no_xla[lax.scatter_mul_p] = _unimplemented("scatter mul")
-tf_impl_no_xla[lax.scatter_add_p] = _unimplemented("scatter add")
+
+def shift_axes_forward(operand, axes: tuple, inverse: bool=False,
+                       forward: bool=True):
+  """Shifts the tuple of axes to the front of an array"""
+  other_axes = tuple([i for i in range(len(operand.shape)) if i not in axes])
+  fwd_order = axes + other_axes if forward else other_axes + axes
+  order = fwd_order if not inverse else tf.math.invert_permutation(fwd_order)
+  return tf.transpose(operand, order)
+
+def convert_scatter_jax_to_tf(update_op, unsorted_segment_op=None):
+  def error(msg):
+    suffix = ("See source code for the precise conditions under which "
+              "scatter_(update/add/multiply/min/max) ops can be converted without XLA.")
+    return _xla_disabled_error("scatter_(update/add/multiply/min/max)", f"{msg} - {suffix}")
+
+  def _sparse_scatter(
+    operand,
+    scatter_indices,
+    updates,
+    update_jaxpr,
+    update_consts,
+    dimension_numbers,
+    indices_are_sorted: bool,
+    unique_indices: bool,
+    mode,
+    _in_avals: Sequence[core.ShapedArray],
+    _out_aval: core.ShapedArray):
+    """
+    Implementation of scatter specialised to indexing from the
+    front axes. This covers unique indices and non-unique indices
+    of single depth.
+
+    Note on unique indices: `tf.tensor_scatter_nd_update` interprets
+    indices thusly: every axis except the final one encodes a batch
+    dimension, the final axis encoding the actual indices to scatter in to.
+    It enforces, at least one, batch dimension so we add an empty
+    dimension to indices and updates if lacking.
+
+    Note on non-unique indices: There is no tf op for non single depth
+    indexing. But if indexing is single depth, this can be viewed as a
+    segment op.
+    """
+    # Infer unique indices from lack of batch dimension
+    unique_indices = unique_indices or (len(scatter_indices.shape) == 1)
+    if unique_indices:
+      suboperand = tf.gather_nd(operand, scatter_indices)
+      updated_suboperand = update_op(suboperand, updates)
+      # add a batch dim if none exist
+      if len(scatter_indices.shape) == 1:
+        scatter_indices = scatter_indices[None]
+        updated_suboperand = updated_suboperand[None]
+      y = tf.tensor_scatter_nd_update(operand, scatter_indices, updated_suboperand)
+    else:
+      if (scatter_indices.shape[-1] == 1) and (unsorted_segment_op != None):
+        # If only indexing into the first dimension, it's a segment op
+        operand_update = unsorted_segment_op(updates, tf.squeeze(scatter_indices, -1), operand.shape[0])
+        y = update_op(operand, operand_update)
+      else:
+        raise error("Scatter supports unique indices. Scatter also supports non-unique indices with indexing into only one dimension for (add, mul, min, max)")
+    return y
+
+  def sparse_scatter(
+    operand,
+    scatter_indices,
+    updates,
+    update_jaxpr,
+    update_consts,
+    dimension_numbers,
+    indices_are_sorted: bool,
+    unique_indices: bool,
+    mode,
+    _in_avals: Sequence[core.ShapedArray],
+    _out_aval: core.ShapedArray):
+    """
+    Wrapper around the scatter function.
+    The underlying tf ops `tf.tensor_scatter_nd_update` and
+    `tf.math.unsorted_segment_*` index from the front dimensions.
+    `tf.math.unsorted_segment_*` indexs to a depth 1 from the front.
+    `tf.tensor_scatter_nd_update` indexs from the front dimensions onwards
+    , with no ability to skip a dimension. This function
+    shifts the axes to be indexed to the front then calls a front-specific
+    implementation, then inverse-shifts the output.
+
+    scatter_dims_to_operand_dims: dimensions which the scatter indexes in to.
+      We shift these to the front to match tf syntax. All other dims are batch
+    update_window_dims: dimensions which are not batch dimensions. We shift
+      these to the back as the remaining dimensions are batch dimensions.
+    """
+    ud = dimension_numbers.update_window_dims
+    wd = dimension_numbers.inserted_window_dims
+    sd = dimension_numbers.scatter_dims_to_operand_dims
+    dtype = operand.dtype # assume updates has same dtype as operand
+    if dtype in [tf.bool, tf.complex64]:
+      raise error(f"Scatter does not support operands of type {dtype}")
+    if not (wd == sd):
+      raise error("Complex scatters are not supported")
+    if not (mode == lax.GatherScatterMode.PROMISE_IN_BOUNDS):
+      raise error("Only scatter mode `PROMISE_IN_BOUNDS` is supported")
+    # Shift axes to the front to match tf syntax, inverse afterwards
+    fwd = partial(shift_axes_forward, axes=sd)
+    inv = partial(fwd, inverse=True)
+    # shift update value axes to the back, so batch are at the front
+    updates_shifted = shift_axes_forward(updates, axes=ud, forward=False)
+    return inv(_sparse_scatter(
+      fwd(operand),
+      scatter_indices,
+      updates_shifted,
+      update_jaxpr,
+      update_consts,
+      dimension_numbers,
+      indices_are_sorted,
+      unique_indices,
+      mode,
+      _in_avals,
+      _out_aval,
+    ))
+  return sparse_scatter
+
+tf_impl_no_xla[lax.scatter_p] = convert_scatter_jax_to_tf(lambda x,y: y) # just replace with the update
+tf_impl_no_xla[lax.scatter_add_p] = convert_scatter_jax_to_tf(tf.add,      tf.math.unsorted_segment_sum)
+tf_impl_no_xla[lax.scatter_mul_p] = convert_scatter_jax_to_tf(tf.multiply, tf.math.unsorted_segment_prod)
+tf_impl_no_xla[lax.scatter_min_p] = convert_scatter_jax_to_tf(tf.minimum,  tf.math.unsorted_segment_min)
+tf_impl_no_xla[lax.scatter_max_p] = convert_scatter_jax_to_tf(tf.maximum,  tf.math.unsorted_segment_max)
 
 tf_impl_no_xla[lax.sort_p] = _unimplemented("sort")
--- a/jax/experimental/jax2tf/tests/primitive_harness.py
+++ b/jax/experimental/jax2tf/tests/primitive_harness.py
@@ -1272,7 +1272,7 @@ def _make_scatter_harness(name,
 # Validate dtypes
 for dtype in jtu.dtypes.all:
   for f_lax in [
-      lax.scatter_add, lax.scatter_mul, lax.scatter_max, lax.scatter_min
+      lax.scatter, lax.scatter_add, lax.scatter_mul, lax.scatter_max, lax.scatter_min
   ]:
     _make_scatter_harness("dtypes", dtype=dtype, f_lax=f_lax)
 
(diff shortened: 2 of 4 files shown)


oliverdutton:
Minor bug in my asserts, fixing now

oliverdutton:
Asserts added back in, with tf.assert when tensors involved.

josephrocca:
Out of curiosity, I tried to use this fix to get the derivative of Flax CLIP exporting correctly, and before you added your asserts back in it failed on the unique index assertion:
```
AssertionError: Indices must be unique for this implementation. For other cases look at using something with tf.math.unsorted_segment_*
```
But with your asserts added back in I get this error:
```
    854     tf.assert_equal(tf.rank(operand), tf.rank(updates)+1, message='Updates must be 1 dimension less than operand')
    855     tf.assert_rank(scatter_indices, 1, message='Scatter indices must be 1 dimensional')
--> 856     assert (len(wd)==len(sd)==1) & (wd==sd) & (ud==tuple(range(max(ud)+1))), 'Unsupported scatter op'
    857     tf.assert_rank(operand, len(ud)+1, message='Unsupported scatter op')
    858     axis = wd[0] # axis the `scatter_indices` index into within `operand`

ValueError: max() arg is an empty sequence
```
So it looks like an extra assert might be needed above that line? Here's a minimal notebook: https://colab.research.google.com/drive/1SPeT3bBBSzZpAz0kqgfzvvYCsTXIPJIb

(It would be really cool if this PR added support for CLIP too, in case it's not too hard to support non-unique indices, but you've already been absurdly helpful so please ignore this if you don't have the bandwidth right now to do this!)

gnecula (commented):
Thank you very much for contributing this. It is one of the most complex cases.

Can you please add some tests?

You would do this in [primitives_harness.py](https://github.com/google/jax/blob/8e4cf253dc047d23cbbcb34a57a99fb82427c943/jax/experimental/jax2tf/tests/primitive_harness.py#L1229), add a `enable_xla` parameter to the harnesses for scatter. Without that parameter set, it defaults to `True`. We want to create a harness with both True/False.

A good example to follow is [`gather`](https://github.com/google/jax/blob/8e4cf253dc047d23cbbcb34a57a99fb82427c943/jax/experimental/jax2tf/tests/primitive_harness.py#L1103). Note that not all `gather` tests work with `enable_xla=False`, see how the `needs_xla` variable is used. I assume something similar can be done to the existing tests for scatter.

To run the tests, I do something like
```
pytest --durations=10 -r a --verbosity 1 -s jax/experimental/jax2tf/tests/primitives_test.py -k "scatter"
```

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,91 @@ def _dynamic_update_slice(operand, update, *start_indices,
...
+    _out_aval: core.ShapedArray):
+    '''
+    This wraps around the scatter function
+    This works for functions which index only in one of the axes, it shifts the axis to
+    be operated on to the lead so a lead specific implementation can be used
+    '''
+    wd = dimension_numbers.inserted_window_dims
+    sd = dimension_numbers.scatter_dims_to_operand_dims
+    ud = dimension_numbers.update_window_dims
+    tf.assert_equal(tf.rank(operand), tf.rank(updates)+1, message='Updates must be 1 dimension less than operand')
```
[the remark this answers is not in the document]
oliverdutton: Shifted in latest commits

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -851,10 +855,14 @@ def sparse_scatter(
...
-    tf.assert_equal(tf.rank(operand), tf.rank(updates)+1, message='Updates must be 1 dimension less than operand')
-    tf.assert_rank(scatter_indices, 1, message='Scatter indices must be 1 dimensional')
-    assert (len(wd) == len(sd) == 1) & (wd == sd) & (ud == tuple(range(max(ud)+1))), 'Unsupported scatter op'
-    tf.assert_rank(operand, len(ud)+1, message='Unsupported scatter op')
+    dtype = operand.dtype # assume updates has same dtype as operand
+    if dtype in [tf.bool, tf.complex64]:
+      raise error(f"Scatter does not support operands of type {dtype}")
+    if not ((scatter_indices.shape==(1,)) & (len(operand.shape) == len(updates.shape)+1) & (len(updates.shape) == len(ud)) &
+      (len(wd) == len(sd) == 1) & (wd == sd) & (ud == tuple(range(max(ud)+1)))):
+      raise error('Scatters of more than one row in one dimension are unsupported')
```
oliverdutton, line 863: I think as long as the `lax.scatter_*` is being used correctly the first check `(scatter_indices.shape==(1,)` is all that's needed. The other inputs will only be wrong if someone is calling via lax.scatter_* directly rather than the .at[].* or numpy syntax which is rare and discouraged.

Should I make it just the first bit, or keep all these checks which ask if the scatter is valid in general.

jax/experimental/jax2tf/tests/primitive_harness.py
```diff
@@ -1315,6 +1317,25 @@ def _make_scatter_harness(name,
                         scatter_indices=np.array([[4], [77]]),
                         update_shape=(2,))
 
+# Validate no XLA scatters
+for dtype in set(jtu.dtypes.all) - set(jtu.dtypes.complex) - set(jtu.dtypes.boolean):
+  for f_lax in [
```
oliverdutton, line 1322: I would add `lax.scatter` here but I get
```python
E       AssertionError: Harness group 'scatter' must be either part of 'harness_groups_no_limitations' or must have a custom Jax2TfLimitation.classmethod defined (see module docstring)
```
when I try to. 

Next commit fixes this.

jax/experimental/jax2tf/impl_no_xla.py
gnecula: I would keep all the checks, this makes it more locally-obvious that the code is correct

oliverdutton:
`scatter` was overlooked in the list `scatter_add,...`. I've added it in and expanded the tests over it.

gnecula (approved):
Looks good, thanks!

josephrocca:
What should this code do if `dimension_numbers.update_window_dims` is an empty sequence? I'm still seeing the `ValueError: max() arg is an empty sequence` error [in this notebook](https://colab.research.google.com/drive/1SPeT3bBBSzZpAz0kqgfzvvYCsTXIPJIb) which tries to port CLIP with jax2tf. If it's not supported yet, then a more informative error could be thrown in this case?

oliverdutton:
Ah yes, I’d forgotten. I’ll add that in before it’s merged.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,99 @@ def _dynamic_update_slice(operand, update, *start_indices,
 
 tf_impl_no_xla[lax.dynamic_update_slice_p] = _dynamic_update_slice
 
-tf_impl_no_xla[lax.scatter_p] = _unimplemented("scatter")
-tf_impl_no_xla[lax.scatter_min_p] = _unimplemented("scatter min")
-tf_impl_no_xla[lax.scatter_max_p] = _unimplemented("scatter max")
-tf_impl_no_xla[lax.scatter_mul_p] = _unimplemented("scatter mul")
-tf_impl_no_xla[lax.scatter_add_p] = _unimplemented("scatter add")
+def get_move_to_leading_axis_bijector(axis):
+  '''
```
marcvanzee, line 796: Nit: we usually use """ rather than '''.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,99 @@ def _dynamic_update_slice(operand, update, *start_indices,
 
 tf_impl_no_xla[lax.dynamic_update_slice_p] = _dynamic_update_slice
 
-tf_impl_no_xla[lax.scatter_p] = _unimplemented("scatter")
-tf_impl_no_xla[lax.scatter_min_p] = _unimplemented("scatter min")
-tf_impl_no_xla[lax.scatter_max_p] = _unimplemented("scatter max")
-tf_impl_no_xla[lax.scatter_mul_p] = _unimplemented("scatter mul")
-tf_impl_no_xla[lax.scatter_add_p] = _unimplemented("scatter add")
+def get_move_to_leading_axis_bijector(axis):
```
marcvanzee, line 795: I am a bit confused by this name and the docstring. Given that it is only used once, perhaps it is easier to understand if you simply define the `shift` function and then define `fwd = shift` and `bwd = partial(shift, inverse=True)` below where you need it?

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,99 @@ def _dynamic_update_slice(operand, update, *start_indices,
...
+  '''
+  Bijector taking an array x, of rank `operand_rank`.
+  Forward shift the axis `axis` to the leading axis
+
+  so for array `operand_rank`=4 and `axis=4`. The last axis is
+  moved for the front: permutation - [3,0,1,2]
```
marcvanzee, line 801: This docstring is a bit confusing, because this function actually returns two functions, which is not really clear. Also the name `get_move_to_leading_axis_bijector` is maybe not very enlightening.

One option is to just define a function `shift_axis_forward(operation, axis, inverse=False)` and define `_fwd = partial(shift_axis, axis=axis)` and `_bwd = partial(_forward, inverse=True)` when you need it below (you only use them once anyway).

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,99 @@ def _dynamic_update_slice(operand, update, *start_indices,
...
+    mode,
+    _in_avals: Sequence[core.ShapedArray],
+    _out_aval: core.ShapedArray):
+    '''
+    Implementation of sparse scatter that works for cases when
+    operand (N dimensional tensor) is indexed in the leading dimension by 1D array `scatter_indices` with updates, an (N-1) dimensional tensor
+    '''
+    suboperand = tf.gather_nd(operand, scatter_indices)
+    updated_suboperand = update_op(suboperand, updates)
+    return tf.tensor_scatter_nd_update(operand, scatter_indices[None], updated_suboperand[None])
```
marcvanzee, line 836: Perhaps it is useful to explain in a comment why you need to add an empty dimension to `scatter_indices` and `updated_suboperand`.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -792,10 +792,99 @@ def _dynamic_update_slice(operand, update, *start_indices,
...
+    indices_are_sorted: bool,
+    unique_indices: bool,
+    mode,
+    _in_avals: Sequence[core.ShapedArray],
+    _out_aval: core.ShapedArray):
+    '''
+    This wraps around the scatter function
+    This works for functions which index only in one of the axes, it shifts the axis to
+    be operated on to the lead so a lead specific implementation can be used
+    '''
```
marcvanzee, line 854:
```suggestion
    Wrapper around the scatter function.
    This works for functions that index in only one of the axes, it shifts the axis to
    be operated on to the front so a front-specific implementation can be used.
    '''
```

marcvanzee (changes requested):
Looks really great, thanks for adding this! Added a few small comments.

jax/experimental/jax2tf/impl_no_xla.py
oliverdutton: Agreed, it's a terrible name. Will alter.

jax/experimental/jax2tf/impl_no_xla.py
oliverdutton: Syntax improvements added

jax/experimental/jax2tf/impl_no_xla.py
oliverdutton: Altered doc string. It comes down to:
XLAGather has dimension numbers which deal with specifying batch dimensions and which dims the scatters go in to. tf.tensor_scatter_nd_update does not have dimension numbers so has to infer everything from the shapes, it indexs from the front only and forcibly has at least one batch dimension.

For example, If you have 4D array, but are indexing into the last 2 dims e.g. x[:,:,0,0] you need to shift the last two dims to the front to use a tf equivalent, you can't just 'pass' on a dim.

oliverdutton:
My dim asserts are actually a bit off now I have time to think about it, I'll fix those

oliverdutton:
Expanded support up to  .at[:, ((3,4),(5,5)), 3] style scatters

oliverdutton:
I expanded support to non-unique indices where the index is to depth 1. i.e `x.at[:, ((0,1,1),)].`. I've added tests for a number of complex cases. These all pass.

I ran @josephrocca's [FlaxCLIPModel example](https://colab.research.google.com/drive/1SPeT3bBBSzZpAz0kqgfzvvYCsTXIPJIb#scrollTo=BU6ySvOGFmFt) and got the wrong answer, I got zeros everywhere. Pure jax gets happy gradients everywhere. It contains 6 scatter-add's I can see in the jax_pr. Possibly one of them has failed but I have no visibility on it failing, it does not error. It may be nothing to do with the scatters, but it seems likely.


```python
fft:f32[1] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
          ffw:f32[] = add ffu ffv
        in (ffw,) }
    ] ffs ffr 1.0
    fga:f32[1,1] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0,), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
          ffw:f32[] = add ffu ffv
        in (ffw,) }
    ] ffz ffl ffy
    fge:f32[1,1,1] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 1), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
          ffw:f32[] = add ffu ffv
        in (ffw,) }
    ] fgd fff fgc
    fic:f32[1,50,768] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 1), inserted_window_dims=(1,), scatter_dims_to_operand_dims=(1,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
          ffw:f32[] = add ffu ffv
        in (ffw,) }
    ] fib dbe fia
          hcs:f32[3,224,40] = scatter-add[
            dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 1), inserted_window_dims=(2,), scatter_dims_to_operand_dims=(2,))
            indices_are_sorted=False
            unique_indices=False
            update_consts=()
            update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
                ffw:f32[] = add ffu ffv
              in (ffw,) }
          ] hcr hco hcq
          hcv:f32[3,40,40] = scatter-add[
            dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 2), inserted_window_dims=(1,), scatter_dims_to_operand_dims=(1,))
            indices_are_sorted=False
            unique_indices=False
            update_consts=()
            update_jaxpr={ lambda ; ffu:f32[] ffv:f32[]. let
                ffw:f32[] = add ffu ffv
              in (ffw,) }
          ] hcu hcn hct
```

oliverdutton:
I ran through the jax_pr's extracted the scatter-add's and added tests. None fail. Looks like the problem in that FlaxCLIP model gradients going to zero in jax2tf.convert aren't down to the scatter ops. 

I am done with adding functionality to this pull request. Please advise.

gnecula:
Thank you @oliverdutton. Can you please squash the many commits into one, or a small number of meaningful ones?

For the error that remains, can either you or @josephrocca update #9659 (which we'll keep open)?

Issue #8505, opened 2021-11-10
bfloat16 arrays are incompatible with pickle


jakevdp:
```python
import pickle
import jax.numpy as jnp
pickle.dumps(jnp.bfloat16(0))
```
```pytb
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
<ipython-input-13-9fa7e9015eb7> in <module>()
      1 import pickle
      2 import jax.numpy as jnp
----> 3 pickle.dumps(jnp.bfloat16(0))

PicklingError: Can't pickle <class 'bfloat16'>: attribute lookup bfloat16 on builtins failed
```
I think this is due to the fact that picle the `bfloat16` custom type does not correctly define its full class name. Compare these::
```python
from jax._src.lib import xla_client
print(type(xla_client.bfloat16(0.0)))
# <class 'bfloat16'>

print(type(jnp.array(0)))
# <class 'jaxlib.xla_extension.DeviceArray'>

print(type(np.float16(0))
# <class 'numpy.float16'>
```

jakevdp:
I think for this to work correctly, we'd need the `tp_name` attribute [here](https://github.com/tensorflow/tensorflow/blob/50bf694f9b917c67e389aaf8e270cc377db4c5d9/tensorflow/python/lib/core/bfloat16.cc#L340) to correspond to its actual import location within JAX, i.e. `jaxlib.xla_client.bfloat16`.

That's not straightforward, though, because this object is shared with tensorflow and exported in a different path within that library.

jakevdp:
Indeed the tensorflow dtype causes this issue as well:
```python
import tensorflow as tf
import pickle
import numpy as np
pickle.dumps(np.ones(10, dtype=tf.bfloat16.as_numpy_dtype))
```
```pytb
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
<ipython-input-24-272a81341603> in <module>()
      2 import pickle
      3 import numpy as np
----> 4 pickle.dumps(np.ones(10, dtype=tf.bfloat16.as_numpy_dtype))

PicklingError: Can't pickle <class 'bfloat16'>: attribute lookup bfloat16 on builtins failed
```

jakevdp:
I think once this path issue is resolved in pickle, numpy's pickle serialization/deserialization should work. For example:
```python
import jax.numpy as jnp
x = jnp.arange(4, dtype='bfloat16')
constructor, cargs, state = x.__reduce__()
[...]

PR #10659, opened 2022-05-10
DeviceArray: Improve support for copy, deepcopy, and pickle

jakevdp:
Currently deep-copying or pickling of jax DeviceArrays is implemented by forwarding `arr.__reduce__` to the ndarray value, meaning that the copied/unpickled result is a normal numpy array, and calling `copy.copy` is a no-op.

This change implements custom `__copy__`, `__deepcopy__`, and `__reduce__` methods in order to properly copy and/or serialize DeviceArray objects in both traced and untraced contexts.

For `__reduce__` (i.e. pickle), deserialization is done on the default device – I'm not sure whether it would make sense to try to persist the device; in any case it's not straightforward because `jaxlib.xla_extension.Device` is not serializable. Further, because deserialization may happen in a different runtime than the one where serialization took place, it's not clear how the non-default device would be identified.

Addresses #2632

Note that pickling of bfloat16 arrays is still broken due to #8505

### DIFF
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -474,6 +474,30 @@ instantiate :class:`DeviceArray` objects manually, but rather will create them v
 :mod:`jax.numpy` functions like :func:`~jax.numpy.array`, :func:`~jax.numpy.arange`,
 :func:`~jax.numpy.linspace`, and others listed above.
 
+Copying and Serialization
+~~~~~~~~~~~~~~~~~~~~~~~~~
+:class:`~jax.numpy.DeviceArray`` objects are designed to work seamlessly with Python
+standard library tools where appropriate.
+
+With the built-in :mod:`copy` module, when :func:`copy.copy` or :func:`copy.deepcopy`
+encounder a :class:`~jax.numpy.DeviceArray`, it is equivalent to calling the
+:meth:`~jaxlib.xla_extension.DeviceArray.copy` method, which will create a copy of
+the buffer on the same device as the original array. This will work correctly within
+traced/JIT-compiled code, though copy operations may be elided by the compiler
+in this context.
+
+When the built-in :mod:`pickle` module encounters a :class:`~jax.numpy.DeviceArray`,
+it will be serialized via a compact bit representation in a similar manner to pickled
+:class:`numpy.ndarray` objects. When unpickled, the result will be a new
+:class:`~jax.numpy.DeviceArray` object *on the default device.*
+This is because in general, pickling and unpickling may take place in different runtime
+environments, and there is no general way to map the device IDs of one runtime
+to the device IDs of another. If :mod:`pickle` is used in traced/JIT-compiled code,
+it will result in a :class:`~jax.errors.ConcretizationTypeError`.
+
+Class Reference
+~~~~~~~~~~~~~~~
+
 .. autoclass:: jax.numpy.DeviceArray
 
 .. autoclass:: jaxlib.xla_extension.DeviceArrayBase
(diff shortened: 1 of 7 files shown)


jakevdp:
pulling in for more testing - not sure whether there will be unintended side-effects.

jakevdp:
Turns out there are a number of downstream users implicitly relying on pickling converting device arrays to numpy arrays. Perhaps we can start with a deprecation warning.

jakevdp:
I'm working on fixing internal breakages; probably will not be able to merge this for a week or two. But requesting review new because I think this will be close to the final state of the PR. Thanks!

jakevdp:
Ready for a review: PTAL

jax/_src/device_array.py
```diff
@@ -298,6 +303,15 @@ def raise_not_implemented():
 # pylint: enable=protected-access
 
 
+def reconstruct_device_array(fun, args, arr_state, aval_state):
+  """Method to reconstruct a device array from a serialized state."""
+  np_value = fun(*args)
+  np_value.__setstate__(arr_state)
+  jnp_value = jax.device_put(np_value)
```
hawkinsp, line 310: My only slight hesitation here is: which device should we put it on?

But maybe this is fine: you get a device array on the default device, and if you want it somewhere else, put it somewhere else.

This deserves documentation somewhere.

hawkinsp (approved):

jax/_src/device_array.py
jakevdp: I discussed this issue in the PR description

jakevdp: I added a note about the behavior to the CHANGELOG. I'm not sure what other obvious place there is to document this, what do you think?

hawkinsp: I'd suggest either on or near documentation for `DeviceArray`, or as a separate note somewhere.

jakevdp: Done, thanks

docs/jax.numpy.rst
```diff
@@ -474,6 +474,30 @@ instantiate :class:`DeviceArray` objects manually, but rather will create them v
 :mod:`jax.numpy` functions like :func:`~jax.numpy.array`, :func:`~jax.numpy.arange`,
 :func:`~jax.numpy.linspace`, and others listed above.
 
+Copying and Serialization
+~~~~~~~~~~~~~~~~~~~~~~~~~
+DeviceArray objects are designed to work seamlessly with Python standard library
```
hawkinsp, line 479: Quote DeviceArray

docs/jax.numpy.rst
```diff
@@ -474,6 +474,30 @@ instantiate :class:`DeviceArray` objects manually, but rather will create them v
...
+tools where appropriate.
+
+With the built-in :mod:`copy` module, when :func:`copy.copy` or :func:`copy.deepcopy`
+encounder a :class:`~jax.numpy.DeviceArray`, it is equivalent to calling the
+:meth:`~jaxlib.xla_extension.DeviceArray.copy` method, which will create a copy of
```
hawkinsp, line 484: A note, not necessarily for this change: I don't consider anything in `jaxlib` to be a public API. We should have public `jax.` names for anything we need to refer to.

hawkinsp (approved):

docs/jax.numpy.rst
jakevdp: Yeah, unfortunately array methods are only defined on the xla extension version, so we can only cross-reference them that way: https://jax.readthedocs.io/en/latest/jax.numpy.html#jax.numpy.DeviceArray

docs/jax.numpy.rst
jakevdp: Done

PR #10678, opened 2022-05-12
Upgrade legacy Python code

JeppeKlitgaard:


benchmarks/benchmark.py
```diff
@@ -126,7 +126,7 @@ def benchmark_suite(prepare: Callable[..., Callable], params_list: List[Dict],
 
   if FLAGS.export_dir:
     filename = _export_results(data_header, data, FLAGS.export_dir, name)
-    print("Wrote %s results to %s" % (name, filename))
+    print("Wrote {} results to {}".format(name, filename))
```
jakevdp, line 129: Is there any option to make this use f-strings instead? That seems like a clearer choice here.

jax/_src/dispatch.py
```diff
@@ -105,7 +105,7 @@ def apply_primitive(prim, *args, **params):
 RuntimeToken = Any
 
 class RuntimeTokenSet(threading.local):
-  tokens: Dict[core.Effect, Tuple[RuntimeToken, Device]]
+  tokens: dict[core.Effect, tuple[RuntimeToken, Device]]
```
jakevdp, line 108: Subscriptable types requires Python 3.9, if I'm not mistaken. We still require compatibility with Python 3.7

jax/_src/dispatch.py
```diff
@@ -153,7 +153,7 @@ def prim_fun(*args):
     return compiled
 
 
-def _device_from_arg_devices(devices: Sequence[Optional[Device]]) -> Optional[Device]:
+def _device_from_arg_devices(devices: Sequence[Device | None]) -> Device | None:
```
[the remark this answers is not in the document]
JeppeKlitgaard: Absolutely - sorry, this was meant to just be a draft for now!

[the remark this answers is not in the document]
jakevdp: Yeah, I think it could be useful!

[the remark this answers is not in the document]
JeppeKlitgaard: I believe the cases where it has upgraded past Python 3.7+ syntax is for files with `from __future__ import annotations`, in which cases there wouldn't be any runtime errors. I'll disable this upgrade path though, as I agree that it is a bit confusing.

benchmarks/benchmark.py
JeppeKlitgaard: This should be fixed now – it is simply because it can only do one "upgrade" at a time (%s -> {}, then next time pyupgrade is run {} -> f-string).

jax/_src/dispatch.py
JeppeKlitgaard: [See below](https://github.com/google/jax/pull/10678#discussion_r871665062)

tests/third_party/scipy/line_search_test.py
```diff
@@ -22,7 +22,7 @@ def assert_wolfe(self, s, phi, derphi, c1=1e-4, c2=0.9, err_msg=""):
     phi0 = phi(0)
     derphi0 = derphi(0)
     derphi1 = derphi(s)
-    msg = "s = %s; phi(0) = %s; phi(s) = %s; phi'(0) = %s; phi'(s) = %s; %s" % (
+    msg = "s = {}; phi(0) = {}; phi(s) = {}; phi'(0) = {}; phi'(s) = {}; {}".format(
```
JeppeKlitgaard, line 25: This could be improved, but should be done manually.

pyupgrade does not do very large `f-string` upgrades because they can be less readable in those cases.

I'd prefer for the conversion of this into an f-string to happen in a separate PR, but happy to add a commit here.

JeppeKlitgaard (commented):
Ready for review @jakevdp

JeppeKlitgaard:
Oh no, maybe not. Something broke!

JeppeKlitgaard:
Should be fixed now.

jakevdp:
Thanks - after discussing this with some team members, I don't think that a pyupgrade hook is something that we want to make part of the CI checks. The main concern is how "opinionated" it is (it seems to be close kin to `black` in that way). For example, if there's a line that you don't want pyupgrade to touch, it appears that there's no way to annotate that (compare this to `flake8` which allows `# noqa`, `pylint` which allows `# pylint: disable`, `mypy`, which allows `# type: ignore`, etc.) In addition, the general feeling is that requiring this formatter will add friction to development without offering enough benefit.

Thanks for proposing and prototyping this!

JeppeKlitgaard:
@jakevdp No worries – it certainly has a strong preference for f-strings, but most of the other 'upgrades' don't seem too controversial.

Let me know if there is appetite for picking through some of the changes introduced here and just disabling the CI. If not, I'll close the PR.

jakevdp:
I think we'd be interested in some of the upgrades here that make the code more clean and modern, but probably not a CI job that errors when those choices are not used.

Another idea proposed by @apaszke was that we might have a review bot that suggests optional changes to new code within pull requests. I'm not sure whether any framework exists for that.

mattjj:
The changes here look good! @jakevdp do you want to check over them in detail or should I?

jakevdp:
I'm happy to go through and review, thanks!

.pre-commit-config.yaml
```diff
@@ -25,3 +25,9 @@ repos:
   hooks:
   - id: jupytext
     args: [--sync]
+
+- repo: https://github.com/asottile/pyupgrade
+  rev: v2.32.1
+  hooks:
+  - id: pyupgrade
+    args: ["--py37-plus", "--keep-runtime-typing"]
```
jakevdp, line 33: As mentioned before, I think we can remove this pre-commit hook. I was looking through the docs to see if there's any way to mark a hook as optional or non-default in the yaml file, but I didn't find anything

jakevdp (changes requested):
I looked through all the diffs, everything looks great to me! Let's remove or disable the pre-commit hook, then get this merged.

JeppeKlitgaard:
Fantastic! Will see if I can do this tonight. We can always do another similar PR when py37 is dropped, etc.

JeppeKlitgaard:
@jakevdp 

Apologies for the delay – I have removed the CI, squashed into 1 commit and rebased off main. Should be ready to merge.

.pre-commit-config.yaml
JeppeKlitgaard: I have removed the hook now, so we should be ready to go 🚀

jakevdp (approved):
Looks great! I'm going to pull it in for some additional testing, and then we can hopefully get it merged.

Issue #6872, opened 2021-06-01
VJP of array slicing is inefficient


shoyer:
When benchmarking gradient calculation through the TPU "Wave Equation" example (https://github.com/google/jax/blob/master/cloud_tpu_colabs/Wave_Equation.ipynb), I noticed that the VJP pass was much slower than I expected (2.4x slower than the forward pass).

The bottleneck seems to be the VJP of array slicing, which ends up using `scatter-add`. When I switch to using `lax.slice` explicitly, or using a custom VJP using `pad`, the backwards pass is just as fast as the forward pass. This is not too surprising, given that the scatter-add call in the VJP doesn't know that indices are sorted or unique.

Here are what the JAXprs look like:
```python
import jax
import jax.numpy as jnp
import numpy as np

def make_grad(f):
  def g(x):
    return f(x).sum()
  return jax.grad(g)

def slicing_original(array):
  return array[1:]

def slicing_slice(array):
  return jax.lax.slice(array, (1,), array.shape)

def slicing_custom_vjp(array):
  @jax.custom_vjp
  def f(array):
    return array[1:]
  def f_fwd(array):
    return f(array), None
  def f_bwd(_, g):
    return (jnp.pad(g, [(1, 0)]),)
  f.defvjp(f_fwd, f_bwd)
  return f(array)

print("VJP of original slicing:")
print(jax.make_jaxpr(make_grad(slicing_original))(np.zeros(10)))

print()
print("VJP of lax.slice slicing:")
print(jax.make_jaxpr(make_grad(slicing_slice))(np.zeros(10)))

print()
print("Custom VJP of slicing:")
print(jax.make_jaxpr(make_grad(slicing_custom_vjp))(np.zeros(10)))
```
Prints:
```
VJP of original slicing:
{ lambda  ; a.
  let b = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(1,) ] 1
      c = gather[ dimension_numbers=GatherDimensionNumbers(offset_dims=(0,), collapsed_slice_dims=(), start_index_map=(0,))
                  slice_sizes=(9,) ] a b
      d = broadcast_in_dim[ broadcast_dimensions=(0,)
                            shape=(9,) ] c
      _ = reduce_sum[ axes=(0,) ] d
      e = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(9,) ] 1.0
      f = reduce_sum[ axes=(  ) ] e
[...]

PR #10682, opened 2022-05-12
[jax2tf] Expand gathers and scatters to slices and dynamic slices where possible

oliverdutton:
When looking into #9659. They'd hit a conversion problem because of a scatter-add in their jaxpr. The root cause comes from handy numpy indexing:
https://github.com/google-research/vision_transformer/blob/main/vit_jax/models_vit.py, line 283
```python
 x = x[:, 0]
```
This is traced to a gather. The grad of a gather is a scatter-add. This can be slower than dynamic_slice on some accelerators and has patchy support in both hardware and converters due to the complex syntax. 

I have rewritten some functions into lax.dynamic_slice to get performance boosts before, it'd be better to avoid the far uglier code it generates. I've had similar issues in scatter where to do `x.at[0].set(y)` I would `concat(y, slice(x,1,n))` to avoid poor scatter implementations. I've written an exit path which catches 'single' gathers into the numpy indexing. 

This can be applied to gathers and scatters in general, is there any simple way to do an optimisation sweep on the jax trace at the end? This would avoid me having to cover all the entry routes into gather/scatter. Though I see that this means the user loses the fine grained control over if they want a gather/scatter instead of dynamic slice or not due to their own reasons.

Note: 
I know that (nearly) all (commonly used) gather/scatters are just loops around: slice array - calculate update - apply update to slice. I would be solving just for loops of length 1. Though it could be handy occassionally to convert short loops as a fix in special cases JAX should not officially support it. Also, I'm only considering unique indices at the moment.

```python
import jax
import jax.numpy as jnp
from jax.lax import GatherScatterMode
# simplify jax_pr's
def wrap(f):
    def _wrapped(*args, **kwargs):
        with jax.ensure_compile_time_eval():
            return f(*args, **kwargs)
    return _wrapped

def check(f, x, eval=True):
    wrap_fn = wrap if eval else lambda f: f
    pr = lambda f: jax.make_jaxpr(wrap_fn(f))
    print(f'val\n{pr(f)(x)}\ngrad\n{pr(jax.grad(f))(x)}')

x = jnp.arange(6, dtype=float).reshape(2,3)

f = lambda x: x[:,0].max()
check(f, x)
# Output (without pull request)
val
{ lambda a:i32[1]; b:f32[2,3]. let
    c:f32[2] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(0,), collapsed_slice_dims=(1,), start_index_map=(1,))
      fill_value=None
      indices_are_sorted=True
      slice_sizes=(2, 1)
      unique_indices=True
    ] b a
    d:f32[] = reduce_max[axes=(0,)] c
  in (d,) }
grad
{ lambda a:i32[1] b:f32[2,3]; c:f32[2,3]. let
    d:f32[2] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(0,), collapsed_slice_dims=(1,), start_index_map=(1,))
      fill_value=None
      indices_are_sorted=True
      mode=GatherScatterMode.CLIP
      slice_sizes=(2, 1)
      unique_indices=True
    ] c a
    e:f32[] = reduce_max[axes=(0,)] d
    f:f32[1] = reshape[dimensions=None new_sizes=(1,)] e
    g:bool[2] = eq d f
    h:f32[2] = convert_element_type[new_dtype=float32 weak_type=False] g
    i:f32[] = reduce_sum[axes=(0,)] h
    j:f32[] = div 1.0 i
    k:f32[2] = broadcast_in_dim[broadcast_dimensions=() shape=(2,)] j
    l:f32[2] = mul k h
    m:f32[2] = reduce_sum[axes=()] l
    n:f32[2,3] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0,), inserted_window_dims=(1,), scatter_dims_to_operand_dims=(1,))
      indices_are_sorted=True
      mode=GatherScatterMode.CLIP
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; o:f32[] p:f32[]. let q:f32[] = add o p in (q,) }
    ] b a m
  in (n,) }
```

Output with pull request
```python
val
{ lambda ; a:f32[2,3]. let
    b:f32[2,1] = dynamic_slice[slice_sizes=(2, 1)] a 0 0
    c:f32[2] = squeeze[dimensions=(1,)] b
    d:f32[] = reduce_max[axes=(0,)] c
  in (d,) }
grad
{ lambda a:f32[2,3]; b:f32[2,3]. let
    c:f32[2,1] = dynamic_slice[slice_sizes=(2, 1)] b 0 0
    d:f32[2] = squeeze[dimensions=(1,)] c
    e:f32[] = reduce_max[axes=(0,)] d
    f:f32[1] = reshape[dimensions=None new_sizes=(1,)] e
    g:bool[2] = eq d f
    h:f32[2] = convert_element_type[new_dtype=float32 weak_type=False] g
    i:f32[] = reduce_sum[axes=(0,)] h
    j:f32[] = div 1.0 i
    k:f32[2] = broadcast_in_dim[broadcast_dimensions=() shape=(2,)] j
    l:f32[2] = mul k h
    m:f32[2] = reduce_sum[axes=()] l
    n:f32[2,1] = broadcast_in_dim[broadcast_dimensions=(0,) shape=(2, 1)] m
    o:f32[2,3] = dynamic_update_slice a n 0 0
  in (o,) }
```

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3527,75 +3527,6 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
   return _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
                  unique_indices, mode, fill_value)
 
-invert_permutation = lambda seq: sorted(range(len(seq)), key=seq.__getitem__)
-def _get_order_with_axes_shift_forward(ndim: int, axes: tuple, inverse: bool=False, forward: bool=True):
-  other_axes = tuple([i for i in range(ndim) if i not in axes])
-  fwd_order = axes + other_axes if forward else other_axes + axes
-  order = fwd_order if not inverse else invert_permutation(fwd_order)
-  return order
-
-def shift_axes_forward(operand: core.Tracer, axes: tuple, inverse: bool=False,
-                       forward: bool=True):
-  """Shifts the tuple of axes to the front of array"""
-  order = _get_order_with_axes_shift_forward(len(operand.shape), axes, inverse, forward)
-  return lax.transpose(operand, order)
-
-def shift_axes_forward_tuple(operand: tuple, axes: tuple, inverse: bool=False,
-                       forward: bool=True):
-  """Shifts the tuple of axes to the front of tuple"""
-  order = _get_order_with_axes_shift_forward(len(operand), axes, inverse, forward)
-  return [operand[k] for k in order]
-
-def get_slice_indices(start_indices: tuple, gather_slice_shape: tuple, array_shape: tuple):
-  """
-  Calculates the slice shapes for PROMISE_IN_BOUNDS
-  """
-  # inbuilt python functions like min are overwritten by np.min so we explicitly define
-  _add = lambda x, y: x + y
-  _min = lambda x, y: x if x < y else y
-  _max = lambda x, y: x if x > y else y
-  # shift if start_indices are lower-out-of-bounds
-  start_indices = tuple(map(lambda i: _max(0, i), start_indices))
-  # calculate limit_indices
-  limit_indices = tuple(map(_add, start_indices, gather_slice_shape))
-  # shift if limit_indices are upper-out-of-bounds
-  bounds_shift = tuple(map(lambda x, y: _min(0, x-y), array_shape, limit_indices))
-  start_indices, limit_indices = [tuple(map(_add, indices, bounds_shift)) for indices in [start_indices, limit_indices]]
-  return start_indices, limit_indices
-
-def _gather_subarray(arr, indexer):
-  """
-  To simplify construction of indices, we shift the dimensions which are
-  indexed to the front. Where the indices are static we use a slice in
-  preference to dynamic slice.
-  """
-  fwd = partial(shift_axes_forward, axes=indexer.dnums.start_index_map)
-  inv = partial(fwd, inverse=True)
-  # Move axes we want to index in to the front to simplify implementation
-  arr = fwd(arr)
-  gather_slice_shape = shift_axes_forward_tuple(indexer.gather_slice_shape, indexer.dnums.start_index_map)
-  # Now the axes beings indexed in to are at the front, building the start indices is simple
-  n_dim = len(arr.shape)
-  n_dim_indexed = len(indexer.gather_indices)
-  start_indices = concatenate([indexer.gather_indices, zeros(n_dim-n_dim_indexed, dtype=indexer.gather_indices.dtype)])
-
-  static_indices = _is_slice_element_none_or_constant(start_indices)
-  if static_indices:
-    # Resolve static indices DeviceArray->tuple
-    start_indices = tuple(map(int, start_indices))
-    start_indices, limit_indices = get_slice_indices(start_indices, gather_slice_shape, arr.shape)
-    subarr = lax.slice(arr, start_indices, limit_indices)
-  else:
-    # dynamic slice
-    subarr = lax.dynamic_slice(arr, start_indices, gather_slice_shape)
-  return subarr, start_indices, fwd, inv
-
-def _lower_gather(arr, indexer):
-  """Implementation covers cases of an unbatched, unstrided gather"""
-  subarr, _, _, inv = _gather_subarray(arr, indexer)
-  y = inv(subarr)
-  return y.squeeze(indexer.dnums.collapsed_slice_dims)
-
 # TODO(phawkins): re-enable jit after fixing excessive recompilation for
 # slice indexes (e.g., slice(0, 5, None), slice(10, 15, None), etc.).
 # @partial(jit, static_argnums=(1, 2))
@@ -3620,15 +3551,11 @@ def _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
 
   # We avoid generating a gather when indexer.gather_indices.size is empty.
   if not core.is_empty_shape(indexer.gather_indices.shape):
-    strides = (len(indexer.gather_indices.shape) != 1)
-    if (not strides) & ((mode == lax.GatherScatterMode.PROMISE_IN_BOUNDS) or (mode == None)) & (fill_value == None):
-      y = _lower_gather(y, indexer)
-    else:
-      y = lax.gather(
-        y, indexer.gather_indices, indexer.dnums, indexer.gather_slice_shape,
-        unique_indices=unique_indices or indexer.unique_indices,
-        indices_are_sorted=indices_are_sorted or indexer.indices_are_sorted,
-        mode=mode, fill_value=fill_value)
+    y = lax.gather(
+      y, indexer.gather_indices, indexer.dnums, indexer.gather_slice_shape,
+      unique_indices=unique_indices or indexer.unique_indices,
+      indices_are_sorted=indices_are_sorted or indexer.indices_are_sorted,
+      mode=mode, fill_value=fill_value)
 
   # Reverses axes with negative strides.
   if indexer.reversed_y_dims:
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3563,17 +3563,25 @@ def get_slice_indices(start_indices: tuple, gather_slice_shape: tuple, array_sha
   start_indices, limit_indices = [tuple(map(_add, indices, bounds_shift)) for indices in [start_indices, limit_indices]]
   return start_indices, limit_indices
 
-def _gather_slice(arr, indexer):
+def _lower_gather_scatter(arr, indexer, update=None, op=None):
   """
-  Implementation covers cases of an unbatched, unstrided gather. To
-  simplify construction of indices, we shift the dimensions which are
+  Implementation covers cases of an unbatched, unstrided gather/scatter op
+
+  If just arr and indexer are specified a gather is performed. It an update and
+  op are specified a scatter is performed.
+
+  To simplify construction of indices, we shift the dimensions which are
   indexed to the front. Where the indices are static we use a slice in
   preference to dynamic slice.
+
+  After extracting the subarray the indexing refers to in gather, we return this
+  ater collapsing singleton dimensions. In scatter, we use the update and subarray
+  to produce the updated subarray which we emplace back in the array.
   """
   fwd = partial(shift_axes_forward, axes=indexer.dnums.start_index_map)
   bwd = partial(fwd, inverse=True)
   # Move axes we want to index in to the front to simplify implementation
-  y = fwd(arr)
+  arr = fwd(arr)
   gather_slice_shape = shift_axes_forward_tuple(indexer.gather_slice_shape, indexer.dnums.start_index_map)
   # Now the axes beings indexed in to are at the front, building the start indices is simple
   n_dim = len(arr.shape)
@@ -3584,13 +3592,27 @@ def _gather_slice(arr, indexer):
   if static_indices:
     # Resolve static indices DeviceArray->tuple
     start_indices = tuple(map(int, start_indices))
-    start_indices, limit_indices = get_slice_indices(start_indices, gather_slice_shape, y.shape)
-    y = jax.lax.slice(y, start_indices, limit_indices)
+    start_indices, limit_indices = get_slice_indices(start_indices, gather_slice_shape, arr.shape)
+    subarr = jax.lax.slice(arr, start_indices, limit_indices)
   else:
     # dynamic slice
-    y = jax.lax.dynamic_slice(y, start_indices, gather_slice_shape)
-  y = bwd(y)
-  y = y.squeeze(indexer.dnums.collapsed_slice_dims)
+    subarr = jax.lax.dynamic_slice(arr, start_indices, gather_slice_shape)
+
+  if not ((update == op == None) or ((update != None) & (op != None))):
+    raise ValueError("Function called incorrectly. Either both op and update must be specified, or neither. Scatter is when update and op are set, Gather is when neither are set")
+
+  if (update == op == None):
+    # Gather, we return the subarray, with empty dimensions collapsed
+    y = subarr
+    y = bwd(y)
+    y = y.squeeze(indexer.dnums.collapsed_slice_dims)
+  else:
+    # Scatter, we reshape the update to match subarr then calculate the update, then emplace into original array
+    update = jax.lax.expand_dims(update, indexer.dnums.collapsed_slice_dims)
+    update =  fwd(update)
+    updated_subarr = op(subarr, update)
+    y = jax.lax.dynamic_update_slice(arr, updated_subarr, start_indices)
+    y = bwd(y)
   return y
 
 # TODO(phawkins): re-enable jit after fixing excessive recompilation for
@@ -3619,7 +3641,7 @@ def _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
   if not core.is_empty_shape(indexer.gather_indices.shape):
     strides = (len(indexer.gather_indices.shape) != 1)
     if (not strides) & ((mode == lax.GatherScatterMode.PROMISE_IN_BOUNDS) or (mode == None)) & (fill_value == None):
-      y = _gather_slice(y, indexer)
+      y = _lower_gather_scatter(y, indexer)
     else:
       y = lax.gather(
         y, indexer.gather_indices, indexer.dnums, indexer.gather_slice_shape,
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3527,6 +3527,72 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
   return _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
                  unique_indices, mode, fill_value)
 
+invert_permutation = lambda seq: sorted(range(len(seq)), key=seq.__getitem__)
+def _get_order_with_axes_shift_forward(ndim: int, axes: tuple, inverse: bool=False, forward: bool=True):
+  other_axes = tuple([i for i in range(ndim) if i not in axes])
+  fwd_order = axes + other_axes if forward else other_axes + axes
+  order = fwd_order if not inverse else invert_permutation(fwd_order)
+  return order
+
+def shift_axes_forward(operand: core.Tracer, axes: tuple, inverse: bool=False,
+                       forward: bool=True):
+  """Shifts the tuple of axes to the front of array"""
+  order = _get_order_with_axes_shift_forward(len(operand.shape), axes, inverse, forward)
+  return lax.transpose(operand, order)
+
+def shift_axes_forward_tuple(operand: tuple, axes: tuple, inverse: bool=False,
+                       forward: bool=True):
+  """Shifts the tuple of axes to the front of tuple"""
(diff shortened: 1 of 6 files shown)


shoyer:
Related discussion around the inefficiency of differentiating gathers:
- https://github.com/google/jax/issues/6872

On the whole I think this sort of change is probably a good idea, given the major potential performance gains. That said, I appreciate the simplicity of lowering indexing to a single XLA operation.

I have not thought about the relative merits of lowering to `lax.dynamic_slice` vs `lax.slice`.

oliverdutton:
I would need to hit:
- numpy indexing
- .at[] indexing + ops
- autobatching rules (I’ve noted a dynamic slice vmaps to a gather - will show jaxpr later)

While I like the idea of ‘fixing’ later on, I think just hitting these convenience cases like x[0] would get 99% of cases so we get pretty code and optimal performance.  

I will add a check if concrete array to catch slice opportunities where available over dynamic_slice.

sharadmv:
Could this be done in the MLIR lowering pass? That would cover the use cases you want, I believe. @hawkinsp

mattjj:
I don't remember for sure but I think there are cases that are hard to turn from general Gathers back to DynamicSlices (and similar for scatters). So +1 to adding something like this optimization! Maybe it wouldn't be hard to handle tuple of ints and slices too?

oliverdutton:
> Could this be done in the MLIR lowering? That would cover the use cases you want, I believe. @hawkinsp

Use cases like jax2tf use the jaxpr’s to convert functions, not MLIR, so you wouldn’t cover it. One of the motivators was scatter-add being unnecessarily present in the grad of models using numpy syntax which then couldn’t convert out of the ecosystem to tflite due to no XLA-free scatter conversion to tensorflow.

oliverdutton:
> I don't remember for sure but I think there are cases that are hard to turn from general Gathers back to DynamicSlices once they've been lowered from NumPy-style indexing to (M)HLO Gather (and similar for scatters). So +1 to adding something like this optimization! Maybe it wouldn't be hard to handle tuple of ints and slices too?

Currently handles
x[:,None,3:5,2]
I think it handles reversing the dim order because the outer of the parser deals with it. I haven’t looked at strides. Slice supports strides, but dynamic slice does not. 

I think x[(0,1)] already gets converted by the handy functions around into x[0,1], cases when someone does x[((0,1),)] however are not.

hawkinsp:
The jax2tf lowering issue is the most compelling reason I can see for doing this change. For compilation via XLA, my argument has always been "XLA can do this" and we should optimize for simpler Python code. But I don't feel that strongly about it.

oliverdutton:
I agree with the "XLA can do this" sentiment, but it's not there yet. Once it is we can remove the dynamic slice and slice primitives as they are just specialisations of gather. Until then, I think this change will make code more readable (and hence maintainable) as people can use numpy syntax rather than work out the particular slice/dynamic slice rules. Also, consider someone writing their own new backend for an FPGA/ASIC, this expands the number of models which will run out of the box on them without full gather semantics being supported.

oliverdutton:
Scatter implemented too. Benchmarks needed to show advantage.

Compare `x.at[3:10].add(1.)` for example
```python
# new jaxpr
{ lambda a:f32[2,7,10]; b:f32[5,7,10] c:i32[]. let
    d:f32[2,7,10] = slice[
      limit_indices=(5, 7, 10)
      start_indices=(3, 0, 0)
      strides=None
    ] b
    e:f32[2,7,10] = add d a
    f:f32[5,7,10] = dynamic_update_slice b e 3 0 0
  in (f,) }
#old jaxpr
{ lambda a:i32[1] b:f32[2,7,10]; c:f32[5,7,10] d:i32[]. let
    e:f32[5,7,10] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 1, 2), inserted_window_dims=(), scatter_dims_to_operand_dims=(0,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; f:f32[] g:f32[]. let h:f32[] = add f g in (h,) }
    ] c a b
  in (e,) }
```

jax/_src/numpy/lax_numpy.py
```diff
@@ -3527,6 +3527,94 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
...
+  # shift if start_indices are lower-out-of-bounds
+  start_indices = tuple(map(lambda i: _max(0, i), start_indices))
+  # calculate limit_indices
+  limit_indices = tuple(map(_add, start_indices, gather_slice_shape))
+  # shift if limit_indices are upper-out-of-bounds
+  bounds_shift = tuple(map(lambda x, y: _min(0, x-y), array_shape, limit_indices))
+  start_indices, limit_indices = [tuple(map(_add, indices, bounds_shift)) for indices in [start_indices, limit_indices]]
+  return start_indices, limit_indices
+
+def _lower_gather_scatter(arr, indexer, update=None, op=None):
```
oliverdutton, line 3566: This now has ugly semantics to support both scatter and gather as they share alot in common. Not sure of best way to proceed here.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3527,6 +3527,94 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
...
+  if static_indices:
+    # Resolve static indices DeviceArray->tuple
+    start_indices = tuple(map(int, start_indices))
+    start_indices, limit_indices = get_slice_indices(start_indices, gather_slice_shape, arr.shape)
+    subarr = jax.lax.slice(arr, start_indices, limit_indices)
+  else:
+    # dynamic slice
+    subarr = jax.lax.dynamic_slice(arr, start_indices, gather_slice_shape)
+
+  if not ((update == op == None) or ((update != None) & (op != None))):
```
oliverdutton, line 3601: Clunky check

jax/_src/numpy/lax_numpy.py
oliverdutton: Splitting methods. Also will abandon the reorder of indices approach in favour of a concat as it removes the tricky ordering of indices back and forth

jax/_src/numpy/lax_numpy.py
oliverdutton: Splitting method removes this in next commits

oliverdutton:
Comparing the scatter I see
```python
x = jax.random.uniform(key, (1_000,1_000))
f = lambda mode: wrap(lambda x: x.at[1:4, 2:20].add(1., mode=mode))
f_slice = f(mode=lax.GatherScatterMode.PROMISE_IN_BOUNDS)
f_scatter = f(mode=lax.GatherScatterMode.CLIP)

display(jax.make_jaxpr(f_slice)(x))
display(jax.make_jaxpr(f_scatter)(x))

f_slice(x).block_until_ready()
%timeit f_slice(x).block_until_ready()

f_scatter(x).block_until_ready()
%timeit f_scatter(x).block_until_ready() 
# Output
{ lambda a:f32[3,18]; b:f32[1000,1000]. let
    c:f32[3,18] = slice[limit_indices=(4, 20) start_indices=(1, 2) strides=None] b
    d:f32[3,18] = add c a
    e:f32[1000,1000] = dynamic_update_slice b d 1 2
  in (e,) }
{ lambda a:i32[2] b:f32[3,18]; c:f32[1000,1000]. let
    d:f32[1000,1000] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(0, 1), inserted_window_dims=(), scatter_dims_to_operand_dims=(0, 1))
      indices_are_sorted=True
      mode=GatherScatterMode.CLIP
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; e:f32[] f:f32[]. let g:f32[] = add e f in (g,) }
    ] c a b
  in (d,) }
2.92 ms ± 286 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
1.99 ms ± 235 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
The scatter-add is faster. This muddies the decision in which to use considerably.

oliverdutton:
Things have gotten a little messy. There are 4 representation layers I can easily access: jaxpr, HLO, MHLO and compiled HLO. (This analysis is with normal JAX, without this PR altering the gathers at the jaxpr level).

The jaxpr, HLO and MHLO all contain the scatter op. But once compiled the HLO module shows its altered down to a `dynamic-update-slice`. 
```python
import jax
from jax import numpy as jnp
x = jnp.zeros((2,))
f = lambda x: x.at[0].add(1.)

jaxpr = jax.make_jaxpr(f)(x)
l = jax.jit(f).lower(x)
mhlo = l.compiler_ir('mhlo')
hlo = l.compiler_ir('hlo').as_hlo_text()
c = l.compile()
compiled_hlo = c.compiler_ir()[0].to_string()

print("# jaxpr")
display(jaxpr)
print('# MHLO\n', mhlo)
print('# HLO\n', hlo)
print('# Compiled HLO\n', compiled_hlo)
# jaxpr
{ lambda ; a:f32[2]. let
    b:bool[] = lt 0 0
    c:i32[] = add 0 2
    d:i32[] = select_n b 0 c
    e:i32[] = convert_element_type[new_dtype=int32 weak_type=False] d
    f:i32[1] = broadcast_in_dim[broadcast_dimensions=() shape=(1,)] e
    g:f32[2] = scatter-add[
      dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
      indices_are_sorted=True
      unique_indices=True
      update_consts=()
      update_jaxpr={ lambda ; h:f32[] i:f32[]. let j:f32[] = add h i in (j,) }
    ] a f 1.0
  in (g,) }
# MHLO
 module @jit__lambda_.22 {
  func public @main(%arg0: tensor<2xf32>) -> tensor<2xf32> {
    %0 = mhlo.constant dense<0> : tensor<i32>
    %1 = mhlo.constant dense<0> : tensor<i32>
    %2 = "mhlo.compare"(%0, %1) {compare_type = #mhlo<"comparison_type SIGNED">, comparison_direction = #mhlo<"comparison_direction LT">} : (tensor<i32>, tensor<i32>) -> tensor<i1>
    %3 = mhlo.constant dense<0> : tensor<i32>
    %4 = mhlo.constant dense<2> : tensor<i32>
    %5 = mhlo.add %3, %4 : tensor<i32>
    %6 = mhlo.constant dense<0> : tensor<i32>
    %7 = "mhlo.select"(%2, %5, %6) : (tensor<i1>, tensor<i32>, tensor<i32>) -> tensor<i32>
    %8 = "mhlo.convert"(%7) : (tensor<i32>) -> tensor<i32>
    %9 = "mhlo.broadcast_in_dim"(%8) {broadcast_dimensions = dense<> : tensor<0xi64>} : (tensor<i32>) -> tensor<1xi32>
    %10 = mhlo.constant dense<1.000000e+00> : tensor<f32>
    %11 = "mhlo.scatter"(%arg0, %9, %10) ({
    ^bb0(%arg1: tensor<f32>, %arg2: tensor<f32>):
      %12 = mhlo.add %arg1, %arg2 : tensor<f32>
      "mhlo.return"(%12) : (tensor<f32>) -> ()
    }) {indices_are_sorted = true, scatter_dimension_numbers = #mhlo.scatter<inserted_window_dims = [0], scatter_dims_to_operand_dims = [0]>, unique_indices = true} : (tensor<2xf32>, tensor<1xi32>, tensor<f32>) -> tensor<2xf32>
    return %11 : tensor<2xf32>
  }
}

# HLO
 HloModule jit__lambda_.22

region_0.4 {
  Arg_0.5 = f32[] parameter(0)
  Arg_1.6 = f32[] parameter(1)
  ROOT add.7 = f32[] add(Arg_0.5, Arg_1.6)
}

ENTRY main.9 {
  Arg_0.1 = f32[2]{0} parameter(0)
  constant.3 = s32[1]{0} constant({0})
  constant.2 = f32[] constant(1)
  ROOT scatter.8 = f32[2]{0} scatter(Arg_0.1, constant.3, constant.2), update_window_dims={}, inserted_window_dims={0}, scatter_dims_to_operand_dims={0}, index_vector_dim=0, indices_are_sorted=true, unique_indices=true, to_apply=region_0.4
}


# Compiled HLO
 HloModule jit__lambda_.22

%fused_computation (param_0: f32[2]) -> f32[2] {
  %param_0 = f32[2]{0} parameter(0)
  %slice.6 = f32[1]{0} slice(f32[2]{0} %param_0), slice={[0:1]}
  %constant_47 = f32[1]{0} constant({1})
  %add.4 = f32[1]{0} add(f32[1]{0} %slice.6, f32[1]{0} %constant_47), metadata={op_name="add" source_file="<ipython-input-1-3e73024d5aba>" source_line=11}
  %constant_46 = s32[] constant(0)
  ROOT %dynamic-update-slice.2 = f32[2]{0} dynamic-update-slice(f32[2]{0} %param_0, f32[1]{0} %add.4, s32[] %constant_46), metadata={op_name="jit(<lambda>)/jit(main)/scatter-add[update_consts=() dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,)) indices_are_sorted=True unique_indices=True mode=GatherScatterMode.PROMISE_IN_BOUNDS]" source_file="<ipython-input-49-b1c3afa2bd3e>" source_line=4}
}

ENTRY %main.9 (Arg_0.1: f32[2]) -> f32[2] {
  %Arg_0.1 = f32[2]{0} parameter(0)
  %copy.1 = f32[2]{0} copy(f32[2]{0} %Arg_0.1)
  ROOT %fusion = f32[2]{0} fusion(f32[2]{0} %copy.1), kind=kLoop, calls=%fused_computation, metadata={op_name="jit(<lambda>)/jit(main)/scatter-add[update_consts=() dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,)) indices_are_sorted=True unique_indices=True mode=GatherScatterMode.PROMISE_IN_BOUNDS]" source_file="<ipython-input-49-b1c3afa2bd3e>" source_line=4}
}
```
In #9659 they wanted a tflite of the grad of a model. Using the handy through HLO converter of tflite in this [notebook](https://colab.research.google.com/drive/1cX2cRC5yNbuPTgGdFEEbPFJTVWgyW-Z9#scrollTo=meaningful-survivor)

Without this PR, it errors with
```
ConverterError: /usr/local/lib/python3.7/dist-packages/transformers/models/clip/modeling_flax_clip.py:569:0: error: 'mhlo.scatter' op is not any of a builtin TFLite op, a flex TensorFlow op or a custom TensorFlow op
<unknown>:0: error: failed while converting: 'main': 
```
because it uses the MHLO (which is pre-compilation). 

With this PR the gather was converted at the jaxpr (top) level, so does not error. 

I will look into whether tflite can use the jaxpr at compiled HLO level where there's no scatter. That would be the best as it means we don't have many levels all of which playing with different rules to take out the scatters/gathers where possible.

Similarly for jax2tf, it'd be greater if we could convert the compiled HLO rather than jaxpr/HLO/MHLO.

oliverdutton:
### Abbreviated view
Gather/Scatter may have poor support/poor implementations on some hardware backends. They may also have poor support for some converters, e.g. jax2tf no XLA which is commonly used to export to ONNX. 

Gather/Scatter are by post-compilation transformed where possible in all cases I've found. Looking at when gather/scatter are transformed into slice/dynamic_slice/dynamic_update_slice paths:
| Op | Simplified to slices |
| ------------- | ------------- |
| Gather  | Uncompiled HLO  |
| Scatter  | Compiled HLO  |

Looking at some ways Jax can be used:
| Conversion path | Exit point | End result |
| ------------- | ------------- | ------------- |
| jax2tf  | jaxpr  | jax2tf always has the gather/scatter, poor performance/support |
| tflite.experimental_from_jax  | Uncompiled HLO  | Gets a simplified gather but trips on scatters which aren't supported currently |
| Compiled  | Compiled HLO  | gather/scatters transformed, possibly there's some overhead in the transform, but reasonably marginal |

### Questions
- What stage to transform at: jaxpr, uncompiled or compiled?
- Can the jax2tf pathway be through HLO rather than jaxpr? and if so how horrific would it be to do so.

oliverdutton:
I need to rewrite my summary in the light of https://github.com/google/jax/issues/10715#issuecomment-1127582189. 

> The compilation flow of JAX looks like this, roughly: Python -> jaxpr -> MHLO -> optimized MHLO -> HLO -> optimized HLO.
> 
> mhlo in your script above corresponds to "MHLO" above, i.e., before any optimization has been done.
> hlo in your script corresponds to "HLO". We still go via MHLO and MHLO optimizations to get here! And any optimization that has been done was done by an MHLO rewrite: the conversion to HLO is almost mechanical beyond those rewrites.
> compiled_hlo corresponds to "optimized HLO", which includes many other rewrites that XLA does during compilation.
> 
> Note that it's entirely expected that optimizations haven't yet been done on mhlo! We do a handful of optimizations on the MHLO before converting it to HLO:
> https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/pjrt/mlir_to_hlo.cc;drc=74f728173a6a3b013f428668e1269b775c0a1e64;l=37
> The key rewrite is canonicalize, which applies a number of simplifications, I believe including the ones you are noting.
> 
> I'd suggest for the tflite conversion probably should be applying MHLO rewrites to simplify the gather also.
> 
> Hope that helps!

oliverdutton:
Looking at compiler passes, until tflite can handle mhlo.scatter the 'scatter_expander' pass just needs to be applied to the HLO, then passed into the converter. This has the benefit of already supporting batched scatter which this PR does not (and should not).

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/scatter_expander.h
> This pass rewrites scatter operations into (roughly) while loops of
> // dynamic-update-slices.
> //
> // This pass can be used in two ways:
> //
> //   - kEliminateAllScatters: For backends that don't support scatter, this pass
> //     can convert every scatter into a loop.
> //
> //   - kEliminateSimpleScatters: For backends that *do* support scatter, this
> //     pass can strength-reduce "simple" scatters -- specifically, scatters that
> //     can be represented without a loop -- to dynamic-update-slices.
> //
> // Note that even in kEliminateSimpleScatters mode, this pass may still expand a
> // scatter into a loop (with a trip-count of 1).  It's up to other
> // simplification passes to remove the loop.

oliverdutton:
This is now a companion PR request to #10653. But instead of mapping scatter to tf-scatter ops as the other PR does, this one expands them into slices. 

It covers a subset of the cases that #10653 does so I'm aiming to land both.

I've also placed the expanded gather, which may be replicating some of the rules already implemented.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -698,8 +699,84 @@ def _gather_with_batch_dims(args: GatherArgs):
...
+  return subarr, start_indices, fwd, inv
+
+def _expand_gather(arr, start_indices, slice_sizes, dnums):
+  """Implementation covers cases of an unbatched, unstrided gather"""
+  subarr, _, _, inv = _gather_subarray(arr, start_indices, slice_sizes, dnums)
+  y = inv(subarr)
+  return y.squeeze(dnums.collapsed_slice_dims)
 
-def _gather(operand, start_indices, *, dimension_numbers,
+def _gather(operand, start_indices, *extra_args, dimension_numbers,
```
oliverdutton, line 779: Accidental, needs reverting

PR #10683, opened 2022-05-12
Run scalar integer index normalization at trace time

### LINKED ISSUE #2878: _index_to_gather produces DeviceArrays for int / onp.ndarray inputs

lgeiger:
This PR speeds up scalar integer indexing outside of jit by doing index computations in plain Python.

I am seeing a ~3x speedup of following toy example on CPU without negatively affecting trace time:
```python
from jax import numpy as jnp

a = jnp.arange(100)
a[5]
%timeit a[6].block_until_ready()
```

**before**:
```
{ lambda ; a:i32[100]. let
    b:bool[] = lt 5 0
    c:i32[] = add 5 100
    d:i32[] = select_n b 5 c
    e:i32[] = convert_element_type[new_dtype=int32 weak_type=False] d
    f:i32[1] = broadcast_in_dim[broadcast_dimensions=() shape=(1,)] e
    g:i32[] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
      fill_value=None
      indices_are_sorted=True
      slice_sizes=(1,)
      unique_indices=True
    ] a f
  in (g,) }
```

**after:**
```
{ lambda a:i32[1]; b:i32[100]. let
    c:i32[] = gather[
      dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
      fill_value=None
      indices_are_sorted=True
      slice_sizes=(1,)
      unique_indices=True
    ] b a
  in (c,) }
```

#10393 was reverted since it included many changes that had unintended side effects which made it hard to review and ended up breaking user code, sorry about that. This PR is a stripped down version that tries to limit the the diff to not reducing code readability.

Fixes #2878

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3585,7 +3585,9 @@ def _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
     y = lax.rev(y, indexer.reversed_y_dims)
 
   # This adds np.newaxis/None dimensions.
-  return expand_dims(y, indexer.newaxis_dims)
+  if indexer.newaxis_dims:
+    y = lax.expand_dims(y, indexer.newaxis_dims)
+  return y
 
 _Indexer = collections.namedtuple("_Indexer", [
   # The expected shape of the slice output.
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3416,6 +3416,12 @@ def _take(a, indices, axis: Optional[int] = None, out=None, mode=None):
 
 def _normalize_index(index, axis_size):
   """Normalizes an index value in the range [-N, N) to the range [0, N)."""
+  try:
+    idx, ndim = operator.index(index), operator.index(axis_size)
+  except TypeError:
+    pass
+  else:
+    return idx + ndim if idx < 0 else idx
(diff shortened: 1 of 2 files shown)


jakevdp:
With all these kinds of speedups, a question that comes to mind is this: should `a.__getitem__(i)` and `jit(a.__getitem__)(i)` perform different operations? If you're detecting when `i` is a static value, then effectively what you're doing is adding a special code-path that only executes when the function is not jit-compiled.

I don't doubt this may be faster, but I'm not certain a 3x speedup of an already fast operation is worth the added complexity in implementation and in understanding how a particular operation is maps to core operations.

jakevdp:
Since I just saw this issue and it's at the top of my mind, here's an example of an optimization we did where a different code-path is chosen for static integer inputs, and it leads to a confusing divergence of behavior in a corner case: #10692

The more such optimizations the add, the more potential there is for these kinds of sharp corners.

YouJiacheng:
I think it is a rare case to perform millions of scalar indexing without `jit`.

PR #10700, opened 2022-05-13
DOC: add references for haskell-style signatures

### LINKED ISSUE #8355: non-Python type signatures in docs are confusing for a Python package
The docs include confusing type signatures, using what I assume is Haskell syntax (please correct me if I'm wrong---I never learned Haskell). Two examples I found are:
https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html#jax.lax.scan
https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html#jax.lax.fori_loop

PS: sorry, I didn't see an option to report an issue with the docs, so just used the "bugs" issue template.

jakevdp:
Fixes #8355

See the rendered version e.g. here: https://jax--10700.org.readthedocs.build/en/10700/_autosummary/jax.lax.scan.html


ngam:
This is beautifully done!!! 😻

ngam:
(but why is it not applied everywhere...?)

jakevdp:
I thought I was applying it everywhere... if you notice any places I've missed, please point them out to me.

ngam:
Oh, sorry I was only noticing the highlighting which I thought was a pretty good way of doing this stuff. 

Comparing 

https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html#jax.lax.scan 

against

https://jax--10700.org.readthedocs.build/en/10700/_autosummary/jax.lax.scan.html?highlight=scan#jax.lax.scan

and I liked how `scan` was yellow-highlighted throughout the whole page:

Screen Shot 2022-05-13 at 17 48 19

ngam:
Now looking more closely, I am not certain how this was produced in this PR!

jakevdp:
Your URL has `?highlight=scan#jax.lax.scan`

jakevdp:
Sorry about that - removed it from my pasted link.

jakevdp:
If you use the search box on the doc page, it will highlight the term you searched for in the result. I found the correct page to link to by searching for `"scan"`, and then I didn't remove this metadata from the URL

ngam:
Ah, wow, I am so old school, I didn't even think that was possible! Thanks for taking the time to explain that :)

mattjj (approved):
Great idea!

I wonder if we should add Python type signatures (which weren't so popular when these docstrings were first written). The syntax may be a bit uglier and that may obscure the meaning... but at least it'd be familiar. We can ponder on that later!

In the meantime this is a great idea for an improvement.

PR #10711, opened 2022-05-14
add core.closed_call_p

mattjj:
This PR adds a variant of core.call_p called core.closed_call_p. The only difference is that in its 'jaxpr form' its call_jaxpr parameter is a core.ClosedJaxpr rather than a core.Jaxpr.

Some background:

* `core.call_p` is the most vanilla call primitive possible: unlike, say, `xla_call_p` (the primitive underlying `jax.jit`), its impl rule isn't to compile an XLA computation, but instead it's just to interpret its jaxpr (staying in Python, using `core.eval_jaxpr`). Correspondingly, it doesn't need to raise the abstraction level of its arguments. It's basically a model for other "final-style" call primitives, each of which is interesting in precisely how it deviates from `core.call_p` (e.g. `xla_call_p`'s impl rule stages out for compilation; `remat_call_p` has a special partial evaluation rule; `custom_jvp_call_p` has a special JVP rule; etc). Historically it    was the first call primitive we introduced, just to test the system; `core.call_p` is not really used anywhere.

* `core.ClosedJaxpr` is a data type which would be better named as `PartiallyAppliedJaxpr`. When we form jaxprs, they usually get paired with "constants" (e.g. `trace_to_jaxpr_nounits` and `trace_to_jaxpr_dynamic` output a list of constants), which are values that are not arguments and that we don't want to turn into literals (e.g. because we want to de-duplicate them, or even just avoid inlining them in pretty-prints). In some cases, these "constants" can be `core.Tracer`s, like when we form the jaxprs for `jax.lax.scan` and the body function closes over some `Tracer`; when that's possible, because `Tracer`s have to be handled with `core.Primitive.bind`, we  typically just convert them to arguments (via `pe.convert_constvars_jaxpr`). But in other cases the constants that come out can't be `Tracer`s (e.g. in the JVP rule of an initial-style primitive, when we run `ad.jvp_jaxpr`, we can get new constants out which can't be `Tracer`s and must be raw array values). That's when `core.ClosedJaxpr` comes in handy: it lets us pair a `jaxpr` with some array constants so that the caller, e.g. a JVP rule for an initial-style higher-order primitive, doesn't need to deal with handling new constant values and their input binders. In other words, primitives which are parameterized by `ClosedJaxpr`s can have simpler rules, especially jaxpr-to-jaxpr rules, since those rules don't need to worry about handling new constants/binders introduced by the rule.

On that last point, when working on #10576 we ran into a situation where

* the current signature for "custom-policy partial eval rules" didn't allow a custom partial evaluation rule to introduce new constants (because such rules just get to output a pair of `Optional[JaxprEqn]`s and have no output for "new constants for the caller to handle appropriately");

* but to perform an optimization, namely hoisting loop-invariant residual computations out of a `scan` body, we might need such a rule to introduce multiple equations as well as new constants.

To proceed, there were at least two options:

1. make the signature for custom-policy partial evaluation rules even more complex (to support outputting multiple equations, new variable names being introduced, new constants, etc)

2. just use a call primitive to handle the "multiple equations with new variables" problem, and as long as it was a call primitive with a `ClosedJaxpr` it would handle the constants problem too.

I chose the second approach, which led to this PR.

For simplicity, we could delete `core.call_p` in favor of this `core.closed_call_p`; after all, the former is not used at all. Going further, we might want to make all higher-order primitives (i.e. even the final-style ones, not just the initial style ones as at present) take `ClosedJaxpr`s rather than `Jaxpr`s; futher still, at    that point we could de-duplicate `Jaxpr` and `ClosedJaxpr` so that we only have one such type. Those simplifications sound reasonable, but they're out of scope for this PR. Here I just want to land a change for enabling the new `remat` implementation with `scan` inside!

Finally, some notes on the changes here. Final-style primitives (like the new `closed_call_p`) have two forms, with different parameters: the 'bind form' used during tracing which takes a Python callable as a parameter representing the function to be called (really a `linear_util.WrappedFun`), and the 'jaxpr form' which appears in  a jaxpr which itself takes a `Jaxpr` (or after this PR alternatively a `ClosedJaxpr`). Since we're introducing a primitive which is like `core.call_p` except that it takes a `ClosedJaxpr` parameter, we need to

* update places where the bind-form primitive is converted to the jaxpr-form primitive (i.e. `JaxprTrace.process_call` and `DynamicJaxprTrace.process_call` in partial_eval.py, both of which can be handled by using the existing "call param updater" hook) to actually produce a `ClosedJaxpr` parameter;

* update places where the jaxpr-form is converted to the bind-form (namely `ClosedCallPrimitive.get_bind_params` in core.py)

* update rules which consume the jaxpr-form to handle the `ClosedJaxpr` parameter (namely the MLIR lowering rule in mlir.py, the transpose rule in ad.py, the typecheck rule in core.py, the DCE rule in partial_eval.py, and (once it exists for any calls) the forwarding rule in partial_eval.py); note that we do _not_ need to update    rules which consume the bind form (e.g. `JVPTrace.process_call` or `BatchTrace.process_call`) since the bind forms of `call_p` and `closed_call_p` are identical;

* update core_test.py to cover the new call primitive.

Only the second-to-last bullet seems burdensome. That would be mitigated by moving to make all call primitives take `ClosedJaxpr` parameters, which I _think_ was already a good idea. But again that's out of scope!

### DIFF
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -986,13 +986,13 @@ def lower_fun(fun: Callable, multiple_results: bool = True) -> Callable:
 
 def _call_lowering(fn_name, stack_name, call_jaxpr, backend, ctx, avals_in,
                    avals_out, tokens_in, *args):
+  if isinstance(call_jaxpr, core.Jaxpr):
+    call_jaxpr = core.ClosedJaxpr(call_jaxpr, ())
   xla.check_backend_matches(backend, ctx.platform)
   output_types = map(aval_to_ir_types, avals_out)
   flat_output_types = util.flatten(output_types)
   effects = tokens_in.effects()
-  symbol_name = lower_jaxpr_to_fun(ctx, fn_name,
-                                   core.ClosedJaxpr(call_jaxpr, ()),
-                                   effects).name.value
+  symbol_name = lower_jaxpr_to_fun(ctx, fn_name, call_jaxpr, effects).name.value
   args = [*tokens_in.tokens(), *args]
   call = func_dialect.CallOp(flat_output_types,
                              ir.FlatSymbolRefAttr.get(symbol_name),
@@ -1024,6 +1024,8 @@ def _named_call_lowering(ctx, *args, name, backend=None,
 
 register_lowering(core.named_call_p, _named_call_lowering)
 register_lowering(core.call_p, partial(_named_call_lowering, name="core_call"))
+register_lowering(core.closed_call_p,
+                  partial(_named_call_lowering, name="core_closed_call"))
 
 
 def full_like_aval(value, aval: core.ShapedArray) -> ir.Value:
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -222,15 +222,17 @@ class JaxprTrace(Trace):
     unknown_arg_tracers = [t for t in tracers if not t.is_known()]
     # Adjust parameters (e.g. donated_invars) for the staged-out call's args.
     num_new_args = len(const_tracers) + len(env_tracers)
-    staged_params = update_params(params, map(op.not_, in_knowns), num_new_args)
-    staged_params = dict(staged_params, call_jaxpr=convert_constvars_jaxpr(jaxpr))
+    staged_params = dict(params, call_jaxpr=convert_constvars_jaxpr(jaxpr))
+    staged_params = update_params(staged_params, map(op.not_, in_knowns),
+                                  num_new_args)
     # The outputs of the staged-out call are Tracers with the new eqn as recipe.
     out_tracers = [JaxprTracer(self, PartialVal.unknown(a), None)
                    for a in out_avals]
     name_stack = self._current_truncated_name_stack()
     source = source_info_util.current().replace(name_stack=name_stack)
     eqn = new_eqn_recipe((*const_tracers, *env_tracers, *unknown_arg_tracers),
-                         out_tracers, primitive, staged_params, jaxpr.effects, source)
+                         out_tracers, primitive, staged_params, jaxpr.effects,
+                         source)
     for t in out_tracers: t.recipe = eqn
     return merge_lists(out_knowns, out_tracers, out_consts)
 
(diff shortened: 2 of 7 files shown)


jax/interpreters/mlir.py
```diff
@@ -986,13 +986,13 @@ def f_lowered(ctx, *args, **params):
 
 def _call_lowering(fn_name, stack_name, call_jaxpr, backend, ctx, avals_in,
                    avals_out, tokens_in, *args):
+  if isinstance(call_jaxpr, core.Jaxpr):
```
sharadmv, line 989: To confirm, this check exists because not all call primitives use closed jaxprs yet. When they do, we can delete this.

jax/interpreters/partial_eval.py
```diff
@@ -511,6 +513,12 @@ def partial_eval_wrapper_nounits(
 call_partial_eval_rules: Dict[Primitive, Callable] = {}
 call_param_updaters: Dict[Primitive, Callable] = {}
 
+def _closed_call_param_updater(params, _, __):
+  jaxpr = params.get('call_jaxpr')
+  if jaxpr is None: return params
```
sharadmv, line 518: When is jaxpr None here?

sharadmv (approved):
LGTM. Thanks for the detailed PR description.

jax/interpreters/partial_eval.py
mattjj: Yeah good question, I forget... let me see if I can exercise this.

mattjj: Ah, it's because in `JaxprTrace.process_call` we actually call [the same call_param_updater](https://github.com/google/jax/blob/ba0a2b3a984858162c6465366ca9fdcc1c97a1c8/jax/interpreters/partial_eval.py#L196) for both the [bind-form](https://github.com/google/jax/blob/ba0a2b3a984858162c6465366ca9fdcc1c97a1c8/jax/interpreters/partial_eval.py#L210) and [jaxpr-form](https://github.com/google/jax/blob/ba0a2b3a984858162c6465366ca9fdcc1c97a1c8/jax/interpreters/partial_eval.py#L225) parameter versions. Usually it's just used to update params like `donated_invars`, and it doesn't matter whether we're working with the bind-form or the jaxpr-form (e.g. [for xla_call](https://github.com/google/jax/blob/main/jax/interpreters/xla.py#L403-L415)).

sharadmv: Makes sense, thanks!

PR #10723, opened 2022-05-16
Include mlir.transforms and mlir.passmanager in jaxlib BUILD.

hawkinsp:
These increase the binary size of jaxlib only a negligible amount, and allow running passes like canonicalization from Python.

These will also be necessary for running the MLIR sparse compiler from JAX.


jakevdp (approved):
Looks good! I'll take your word for it on the binary size 😀

PR #10724, opened 2022-05-16
normalize/rearrange (kw)args by signature in jit

YouJiacheng:
Normalize/Rearrange positional arguments and keyword arguments according to function signature.
Thus the static arguments and donate arguments behavior will not depend on the way arguments passed.
This is a breaking change, deprecation cycle might be needed.
Cache behavior is changed as well, cause `test_jit_kwargs` failure.
Since the type of argument is determined by signature, `test_jit_with_mismatched_static_argnames` fail and need some modifications.


JeppeKlitgaard:
I have a different approach that I would prefer over this in my WIP on argument annotation. I will release this soon, but IMO it would be best to avoid changing the jit transformation behaviour without first completing the discussion in #10614

PR #10732, opened 2022-05-16
custom transpose: nearly rewrite `custom_vjp`

froystig:
This change nearly supports a reimplementation of `custom_vjp` on top of `custom_transpose` and `custom_jvp`, with the former still in an intermediate state. All `custom_vjp` tests pass except for those that exercise "old" remat, and this upgrade continues to be guarded by a flag that is off by default.

part of #9129


froystig:
I overlooked that we need our non-`custom_vjp` version of `custom_transpose` working still for this PR to go in. I'll finish that up and include it as another change in this same PR.

hawkinsp:
Closing as stale.

PR #10752, opened 2022-05-18
Fix windows copy

cloudhan:
Together with https://github.com/tensorflow/tensorflow/pull/56154 this should fix latest jax problem.


hawkinsp (approved):
Thanks for the cleanup!

Issue #10750, opened 2022-05-18
Using ```grad``` on ```vmap``` on ```map``` on function containing ```sinc``` results in error


Ph03n1xdust:
Hi,

I ran into an error while trying to take the ```grad``` of a ```vmap``` on a ```map``` on a function, which contains ```sinc```.  The following code reproduces the error:

```
import jax
import jax.numpy as jnp

N_batch = 5
N_elem = 3

batch_data = jnp.reshape(jnp.arange(N_batch*N_elem),(N_batch,N_elem))

def fail(x):
    return jnp.sinc(x)

def mapfail(x):
    return jax.lax.map(fail, x)

vmapmapfail = jax.vmap(mapfail)

def loss(param):
    pos = param*batch_data
    desc = vmapmapfail(pos)
    return jnp.mean(desc)

jax.value_and_grad(loss)(0.1)
```

- Error: ```JaxStackTraceBeforeTransformation: TypeError: broadcast_in_dim broadcast_dimensions must have length equal to operand ndim; got broadcast_dimensions () for operand ndim 1.```
- Jax version: 0.3.13

I suspect this is related to the custom derivatives of the ```_sinc_maclaurin```, since commenting out both ```@partial(custom_jvp, nondiff_argnums=(0,))``` and ```@_sinc_maclaurin.defjvp``` in the jax source "solves" the problem. It is also specific to ```map```, since using a vmap instead results in no error. 

Ps.: A colleague of mine also reminded me, that the documentation of sinc, does not describe the x=0 behaviour correctly.

jakevdp:
Thanks for the report. I'll look into the issue – Can you say more about the issue you see in the `sinc` documentation?

sousaw:
I am afraid I am the nitpicking colleague (unexpectedly) mentioned in the report :-). My comment was actually more closely related to the part of the docstring inherited from numpy, specifically the sentence "the sinc function is sin(pi*x)/(pi*x)". Since the analytic character of sinc is important both for its applications and for the actual implementation at x=0, I would suggest defining it, for instance, as the limit when t->x of sin(pi*t)/(pi*t).

Thank you for your work on this fantastic project!

jakevdp:
Thanks - as you said, that part of the docstring is drawn directly from `numpy.sinc`, so the tweak would have to be contributed to the numpy package itself.

mattjj:
This is a fantastic bug! Thanks for raising it. (And thanks for the tip about the docstring too.)

[...]

PR #10756, opened 2022-05-18
quick fix for #10750, add checks and todo

mattjj:
Quick fix for #10750. See that comment thread for a description of cleaner fixes.

Also add checks so that we can catch such errors earlier (making them less confusing).

### DIFF
--- a/jax/_src/numpy/ufuncs.py
+++ b/jax/_src/numpy/ufuncs.py
@@ -700,10 +700,11 @@ def sinc(x):
 def _sinc_maclaurin(k, x):
   # compute the kth derivative of x -> sin(x)/x evaluated at zero (since we
   # compute the monomial term in the jvp rule)
+  # TODO(mattjj): see https://github.com/google/jax/issues/10750
   if k % 2:
-    return lax.full_like(x, 0)
+    return x * 0
   else:
-    return lax.full_like(x, (-1) ** (k // 2) / (k + 1))
+    return x * 0 + _lax_const(x, (-1) ** (k // 2) / (k + 1))
 
 @_sinc_maclaurin.defjvp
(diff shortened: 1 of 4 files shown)


jax/_src/numpy/ufuncs.py
```diff
@@ -700,10 +700,12 @@ def sinc(x):
...
   # compute the kth derivative of x -> sin(x)/x evaluated at zero (since we
   # compute the monomial term in the jvp rule)
+  # TODO(mattjj): use a full_like primitive here rather than multiplying by
+  # zero, see https://github.com/google/jax/issues/10750
   if k % 2:
-    return lax.full_like(x, 0)
+    return x * 0
   else:
-    return lax.full_like(x, (-1) ** (k // 2) / (k + 1))
+    return x * 0 + lax.full_like(x, (-1) ** (k // 2) / (k + 1))
```
jakevdp, line 708: We could remove the `full_like` in this line, and just rely on broadcasting, yes?

jakevdp (approved):

jakevdp:
Also, is it worth adding a regression test?

jax/_src/numpy/ufuncs.py
mattjj: I think it might insert a `convert_element_type` for us in some cases but I'm not sure. That is, I didn't want `(-1) ** (k // 2) / (k + 1)` to cause dtype promotion, and the `full_like` will ensure it's cast down to be like `x`. But I guess it's always weak?

jakevdp: Maybe use `_lax_const(x, (-1) ** (k // 2) / (k + 1))` then?

PR #10759, opened 2022-05-19
add support for lambda-bound dynamic shape output (iree only)

mattjj:
The point here is to make this work:

```python
@jax.jit
def f(n):
  print('tracing!')
  return jnp.ones(n)

print(f(3))
print(f(4))
```

```
tracing!
[1. 1. 1.]
[1. 1. 1. 1.]
```

The changes are just:
1. in dispatch.py, change result handlers to take a new "environment" argument representing lambda-bound values needed for output types,
2. delete some automatic-argument-padding stuff we decided to move away from.


froystig (approved):

dougalm (commented):
Looks good!

PR #10771, opened 2022-05-19
Using GFile for persistent compilation caching in JAX

sshahrokhi:
The filesystem method that was used for persistent compilation caching, was chosen because it was able to implement Least Recently Used, but it didn't support gcs. GFile supports gcs which can be used between machines but unfortunately it is not following the LRU policy. Therefore, there is no maximum memory we allow for the cache as well. 

[TESTING] I tested this method by using [this test](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/jax/compilation-cache.libsonnet), tested for both local file and a gcs bucket to make sure both works fine.


jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,56 @@
...
+        return file.read()
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry."""
+    if not key:
+      raise ValueError("key cannot be empty")
+    path_to_new_file = str(self._path / key)
+    logging.info(path_to_new_file)
```
skye, line 44: Remove debugging log?

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,56 @@
...
+    with gfile.GFile(path_to_new_file, "w") as f:
+      f.write(value)
+
+  def initialize_gfilecache(path, max_cache_size_bytes=None):
+    """Creates GFileCache Instance."""
+    assert cc._cache is None, f"The cache path has already been initialized to {cc._cache._path}"
+    if max_cache_size_bytes is not None:
```
skye, line 51: Let's just remove the max_cache_size_bytes argument. This will also require updating the tests.

skye (changes requested):
A few high-level comments:
* Please update the PR title and description to be more descriptive (see https://cbea.ms/git-commit/ for a long description of how to write a good commit message). This can also be the commit message once you squash everything before commit.
* Also describe how you tested this change in the PR description, since we don't have automatic Cloud TPU presubmits :(
* Delete the old implementation
* Add this as a breaking change to the changelog: https://github.com/google/jax/blob/main/CHANGELOG.md

sshahrokhi:
Thanks for the reviews! I am working on them! Meanwhile:
> * Add this as a breaking change to the changelog: https://github.com/google/jax/blob/main/CHANGELOG.md
 I didn't remove the max_cache_size_bytes so that this change would not be breaking(only the logic behind the file system is changing)! So do you suggest removing it and also add this to changelog?

skye:
> I didn't remove the max_cache_size_bytes so that this change would not be breaking(only the logic behind the file system is changing)! So do you suggest removing it and also add this to changelog?

Yes. This is a breaking change, but I think it's better to raise an error than to silently stop working.

sshahrokhi:
Thanks for all the comments, could you please take a look now? @skye

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
       previously were no-ops. Now they use the same mechanism as `DeviceArray.copy()`.
     - Calling `pickle` on a traced array now results in an explicit 
       `ConcretizationTypeError`.
+  * {func}`jax.experimental.compilation_cache.initialize_cache` now supports gcs 
+    bucket path as input and it does not support `max_cache_size_bytes` anymore and
+    will not get that as an input.
```
skye, line 30: For the `max_cache_size_bytes` bit, you can actually create a `* Breaking changes` section above the `* Changes` sections. (We haven't done this in a while, but there should be examples somewhere in this file.) You can keep the bit about GCS bucket paths here, since it's not a breaking change. The idea is that it's extra important to highlight breaking changes.

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry."""
+    if not key:
+      raise ValueError("key cannot be empty")
+    path_to_new_file = str(self._path / key)
+    with gfile.GFile(path_to_new_file, "w") as f:
+      f.write(value)
```
skye, line 45: We need to atomically create the file here. I think GCS renames are atomic, so I think it would work to write a temporary version of the file, maybe something like `tmp_path = str(self._path / "_" + key)` and then move that to `path_to_new_file`.

skye (changes requested):

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pathlib
+
+from jax.experimental.compilation_cache.cache_interface import CacheInterface
+import tensorflow.io.gfile as gfile
```
yashk2810, line 18: Please use `from etils import epath` instead of TF.

jax/experimental/compilation_cache/gfile_cache.py
yashk2810: GCS renames are NOT atomic.

GCS does a copy and then delete.

image

skye: Oh wow. If they're immutable, hopefully this means the copy is effectively atomic though? And maybe the write itself? Like it seems like it should be impossible to read a half-written file.

yashk2810: micban@ told us to not depend on rename in GCS at all. I'll tag you in that thread.

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,48 @@
...
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry."""
+    if not key:
+      raise ValueError("key cannot be empty")
+    path_to_new_file = self._path / key
+    if 'gs://' in str(path_to_new_file):
+      path_to_new_file.write_bytes(value)
+    else:
+      tmp_path = epath.Path(str(self._path / "_") + key)
```
yashk2810, line 46: Use f-strings here instead of `+`?

yashk2810: Also maybe you don't need another epath use?

`tmp_path = self._path / f"_temp_{key}"`?

yashk2810: Can you try using this please? What you have right now is very confusing.

sshahrokhi: aaah, I see, I didn't see your second comment! :D yeah sure.

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,48 @@
...
+    else:
+      return None
+
+  def put(self, key: str, value: bytes):
+    """Adds new cache entry."""
+    if not key:
+      raise ValueError("key cannot be empty")
+    path_to_new_file = self._path / key
+    if 'gs://' in str(path_to_new_file):
+      path_to_new_file.write_bytes(value)
```
yashk2810, line 44: Is there a reason to use write_bytes instead of write_text()?

Can you please add a comment?

sshahrokhi: Because the compiled stuff are not texts but they are bytes?

yashk2810: Ohhh.. ok!

Thanks!

jax/experimental/compilation_cache/gfile_cache.py
```diff
@@ -0,0 +1,48 @@
...
+    else:
+      tmp_path = self._path / f"_temp_{key}"
+      tmp_path.write_bytes(value)
```
skye, line 47: We should ideally be using the logic from the deleted FileSystemCache to atomically write local files:
```python
          with open(temp_path_to_file, "wb") as file:
              file.write(value)
              file.flush()
              os.fsync(file.fileno())
```

See https://stackoverflow.com/questions/2333872/how-to-make-file-creation-an-atomic-operation. However, this is only valid if we know that self._path refers to a local filesystem -- it could still be `s3://` or something. I think we basically need a `elif <self._path is a local path>`, and then `else` with the current fallback logic, but I can't find a way to determine if self._path is a local path... maybe for now use `'://' not in str(path_to_new_file)`, and we can file an issue against etils to add a way to get the filesystem more robustly?

skye (changes requested):
Almost there!!

jax/experimental/compilation_cache/gfile_cache.py
sshahrokhi: we can have `file:///tmp/x` as the input, which has that `://` in it?  
and are you suggesting I don't use epath for the local file?

yashk2810: epath handles local files too :)

skye: epath handles local files, but I don't think it does the `flush` + `fsync` that's required for atomicity (there's a performance cost). So I'm saying we use `epath` for everything except writing the file, where we use raw filesystem operations.

Good point about `file://` being valid in addition to not providing a filesystem prefix at all... maybe use `if '://' not in str(path_to_new_file) or 'file://' in str(path_to_new_file)`?

skye: We should also make a FR for epath to handle atomic writes :)

skye (approved):
Looks great! Can you squash all your changes to a single commit and then I'll get this in?

skye (approved):

tests/gfile_cache_test.py
```diff
@@ -0,0 +1,86 @@
...
+# limitations under the License.
+
+from absl.testing import absltest
+from jax.experimental.compilation_cache.gfile_cache import GFileCache
+import jax._src.test_util as jtu
+import tempfile
+import threading
+import time
+
+class FileSystemCacheTest(jtu.JaxTestCase):
```
skye, line 22: Rename to GFileCacheTest

skye (approved):

skye:
Please squash your changes again too, and then we can get this in :crossed_fingers:

sshahrokhi:
> 

done! thanks!

skye (approved):

PR #10789, opened 2022-05-21
[ROCm] Fix hip_lu_pivots_to_permutation.

rsanthanam-amd:
/cc @hawkinsp 

Sorry, I realized that this needed to be fixed also.

In addition, It looked like cuda_kernels.cc was a vestigial artifact, so I removed it.

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -981,7 +981,7 @@ if gpu_linalg:
   mlir.register_lowering(
       lu_pivots_to_permutation_p,
       partial(_lu_pivots_to_permutation_gpu_lowering,
-              gpu_linalg.rocm_lu_pivots_to_permutation),
+              gpu_linalg.hip_lu_pivots_to_permutation),
       platform='rocm')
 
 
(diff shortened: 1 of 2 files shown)


jaxlib/cuda/cuda_kernels.cc
```diff
@@ -1,70 +0,0 @@
-/* Copyright 2021 Google LLC
```
hawkinsp, line 1: Please revert this change.

This is actually not a vestigal artifact, but it's not used by JAX itself directly.

It exists for the convenience of some users of JAX who stage out HLO IR from JAX and run it separately from JAX itself. Of course, HLO staged out by JAX may contain calls to these kernels, so this is an additional library that statically links in the necessary CUDA kernels to run JAX HLO without any of the Python mechanisms.

We could add something similar to the ROCm side although I might wait until there's a user asking for it.

hawkinsp (changes requested):

jaxlib/cuda/cuda_kernels.cc
rsanthanam-amd: > Please revert this change.
> 
> This is actually not a vestigal artifact, but it's not used by JAX itself directly.
> 
> It exists for the convenience of some users of JAX who stage out HLO IR from JAX and run it separately from JAX itself. Of course, HLO staged out by JAX may contain calls to these kernels, so this is an additional library that statically links in the necessary CUDA kernels to run JAX HLO without any of the Python mechanisms.
> 
> We could add something similar to the ROCm side although I might wait until there's a user asking for it.

Done.  Sorry about that.

hawkinsp (approved):

PR #10799, opened 2022-05-23
ENH _wraps() in jax._src.numpy.util is now returns a generic function

superbobry:
This frees type checkers from the need to explicitly infer the return type of _wraps() at each call site.


jakevdp (approved):
Looks reasonable, thanks

Issue #2731, opened 2020-04-15
Support multinode training on GPU


py4:
I don't have a node with 8 gpus. I have two nodes each with 4 gpus. So is it possible to train a model on multiple nodes?

hawkinsp:
This *is* actually something that does work right now but it's still experimental. There's also no real public-facing API for it yet; you have to type in some obscure and fairly magical things to set it all up correctly.

We should polish it off and document it!

hawkinsp:
Can you say a bit more about your model, though? Would gradient all-reductions across multiple nodes suffice?

py4:
@hawkinsp Technically, I'm training a reformer model using Trax library.

hawkinsp:
And I assume you're just looking for data parallelism, i.e., partitioning a minibatch across GPUs, not partitioning in any other way (e.g., model parallelism)?

py4:
@hawkinsp yeah my concern is data parallelism

powderluv:
@hawkinsp Can you please share your notes on this (don't need a stable api) ? We are trying some hybrid data/model/pipeline parallelism  so it is a little different from @py4 but would love to get started with data parallelism

brettkoonce:
Data parallelism would of value to other projects that use XLA as well (eg https://www.tensorflow.org/swift).  Exposing this functionality in a standardized way would help drive progress in the broader ecosystem!

yxd886:
> I don't have a node with 8 gpus. I have two nodes each with 4 gpus. So is it possible to train a model on multiple nodes?

Hello py4, I am meeting the same problem, have you found some solutions?

yxd886:
> This _is_ actually something that does work right now but it's still experimental. There's also no real public-facing API for it yet; you have to type in some obscure and fairly magical things to set it all up correctly.
> 
> We should polish it off and document it!

Hello hawkinsp, Could you please provide more details about how to run data parallel with multi node GPUs?

connection-on-fiber-bundles:
@hawkinsp We are also interested in running JAX code on multiple nodes. Anything (hacky or not) that you can share would be appreciated. Thanks!

jramapuram:
[...]

PR #10822, opened 2022-05-25
Fix distributed system initialization.

hawkinsp:
There was a version compatibility problem that meant that the distributed system did not correctly initialize when mixing the current released versions of jax and jaxlib. We don't have tests for multiple GPUs in our CI, so this was not caught.

Issue https://github.com/google/jax/issues/2731#issuecomment-1133443849

### DIFF
--- a/jax/_src/distributed.py
+++ b/jax/_src/distributed.py
@@ -114,15 +114,22 @@ def initialize(coordinator_address: Optional[str] = None,
   logging.info('Connecting to JAX distributed service on %s', coordinator_address)
   distributed_client.connect()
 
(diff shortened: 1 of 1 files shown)


yashk2810 (approved):

yashk2810:
Should the version name be xla_extension_version?

hawkinsp:
> Should the version name be xla_extension_version?

It doesn't matter a whole lot: `jax._src_lib.xla_extension_version` is an alias for `jaxlib.xla_client._version`. In general I tend to prefer the first one inside JAX, though, but not enough to bother changing it here...

PR #10824, opened 2022-05-25
[x64] add jax_numpy_dtype_promotion flag

jakevdp:
This adds a `jax_numpy_dtype_promotion` flag, along with a `strict` mode that will raise an error on any implicit promotion between differing types. For example:
```python
import jax
import jax.numpy as jnp

x = jnp.float32(4)
y = jnp.int32(4)

# promotion='standard' is JAX's current default promotion scheme
with jax.numpy_dtype_promotion('standard'):
  print(x + 1)  # 5.0
  print(y + 1)  # 5
  print(y + 1.0)  # 5.0
  print(x + y)  # 8.0

# promotion='strict' disallows binary operations between differing dtypes, but still allows operations between a
# strong type and a compatible weak type.
with jax.numpy_dtype_promotion('strict'):
  print(x + 1)  # 5.0
  print(y + 1)  # 5
  print(y + 1.0)  # TypePromotionError: Input dtypes ('int32', "<class 'float'>") have no available implicit dtype promotion path
  print(x + y)  # TypePromotionError: Input dtypes ('float32', 'int32') have no available implicit dtype promotion path
```

I'm open to input on the details of the flag name & values.

After landing this, we should probably change `JaxTestCase` to run with `jax_numpy_dtype_promotion('strict')` by default, simlar to how we run with `jax_numpy_rank_promotion='raise'` currently. A quick tests shows that this will take a fair amount of work, as JAX library code relies on implicit promotion in many places.

### DIFF
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -21,7 +21,7 @@
 
 
 import functools
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


jax/_src/dtypes.py
```diff
@@ -270,23 +270,34 @@ def _dtype_and_weaktype(value):
...
+    return {
+      b1: [i_],
+      u1: [i2, u2], u2: [i4, u4], u4: [i8, u8], u8: [f_],
+      i_: [u1, i1], i1: [i2], i2: [i4], i4: [i8], i8: [f_],
+      f_: [bf, f2, c_], bf: [f4], f2: [f4], f4: [f8, c4], f8: [c8],
+      c_: [c4], c4: [c8], c8: [],
+    }
+  elif jax_numpy_dtype_promotion == 'strict':
+    return {
+      i_: [f_] + _jax_types[1:9],
```
mattjj, line 290: I wonder if it'd be more readable if we gave a variable name to `_jax_types[1:9]` and friends, like above if we split up things like

```python
_exact_types = [...]
_float_types = [...]
_complex_types = [...]
_jax_types = [*_exact_types, *_float_types, *_complex_types]
```

WDYT?

mattjj (approved):

jax/_src/dtypes.py
jakevdp: Good idea - Done!

PR #10827, opened 2022-05-25
Add a public facing `named_scope` function

sharadmv:
Adds `jax.named_scope`, a context-manager that adds annotations into XLA metadata for the purposes of profiling/debugging.

Example usage:
```python
@jax.jit
def layer(w, x):
  with jax.named_scope("dot_product"):
    logits = w.dot(x)
  with jax.named_scope("activation"):
    return jax.nn.relu(logits)
```

Note that the metadata from this contextmanager will only be used when the `'jax_experimental_name_stack'` flag is set to `True`.

For more examples, see `tests/name_stack_test.py`.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -29,9 +29,9 @@ import sys
 import threading
 import weakref
 import types
-from typing import (Any, Callable, Iterable, NamedTuple, Mapping, Optional,
-                    Sequence, Tuple, TypeVar, Union, overload, Dict, Hashable,
-                    List)
+from typing import (Any, Callable, Generator, Iterable, NamedTuple, Mapping,
+                    Optional, Sequence, Tuple, TypeVar, Union, overload, Dict,
+                    Hashable, List)
 from typing_extensions import Literal
 from warnings import warn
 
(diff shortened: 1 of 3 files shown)


jax/_src/api.py
```diff
@@ -3159,6 +3159,57 @@ def named_call_f(*args, **kwargs):
...
 
+@contextmanager
+def named_scope(
+    name: str,
+  ) -> Generator[None, None, None]:
+  """A context manager that adds a user specified name to the JAX name stack.
+
+  When staging out computations for just-in-time compilation to XLA (or other
+  backends such as TensorFlow) JAX runs your Python program but by default does
+  not preserve any of the function names or other metadata associated with it.
```
froystig, line 3170:
```suggestion
  backends such as TensorFlow) JAX does not, by default, preserve the names
  (or other source metadata) of Python functions it encounters.
```
Trying to avoid an unnecessary "JAX runs your program", to skip any incidental confusion regarding (abstract vs. concrete) evaluation.

froystig (approved):

PR #10852, opened 2022-05-27
Checkify: support discharging checks from control-flow through effects.

LenaMartens:
Currently supports scan and while-loop.

Two questions (@sharadmv):
  - Currently the `test_assert_primitive_staging_but_no_translation` test fails: you're able to abstractly evaluate an effectful `check` through `eval_shape` without an error (because no lowering happens). Do you see a way around this? Because I'm now allowing abstract_eval but disallowing lowering, I'm missing this error case (maybe that's fine?)
  - I can't seem to discharge from the `cond` in the `while_loop`, is this expected? If not, it might be a bug in my  while_loop_rule.

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -25,9 +25,11 @@ import jax.numpy as jnp
 from jax import core
 from jax import linear_util as lu
 from jax.api_util import flatten_fun
+from jax.interpreters import mlir
 from jax.interpreters import partial_eval as pe
 from jax.tree_util import tree_flatten, tree_unflatten, register_pytree_node
 from jax._src import source_info_util, traceback_util
+from jax._src.lax import control_flow as cf
 from jax import lax
 from jax._src.util import (as_hashable_function, unzip2, split_list, safe_map,
                            safe_zip)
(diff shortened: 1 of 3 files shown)


LenaMartens:
After thinking about it, maybe abstractly evaluating a `check` is fine because I wouldn't expect an assertion error to ever trigger on abstract values. The real issue is when _staging out_ the check, which I think means it has to go through a lowering step first, so that makes it ok to defer the error to lowering time.
So, something like `eval_shape-of-check` should be ok to allow, WDYT?

LenaMartens:
> to inline the cond logic inside of the body and discharge effects in the body as usual
Ah yes, cool! We do this trick in the checkify while_loop rule to add automatic checks to the `cond` (inline cond into body), but I now realize this leaves an undischarged `check`. I'll need to find a way to execute the cond without any checks, maybe by functionalizing the cond, then discarding the error value?

I'll try it out and get back to you :)

LenaMartens:
Sorry for the delay! Discharging the check in the `cond` seems to work nicely and actually simplifies some of my earlier code.

For future ref: we used to generate a "compatible cond" jaxpr which was equivalent to:

```
def compat_cond(error_value, *args):
  del error_value  # ignore error input
  return cond(*args)
```
This worked in the case of adding automatic checks, but did not actually discharge any check primitives from the cond. We already have a discharged version of the `cond` which we run once to capture any failing check in the first evaluation of cond, we can use that jaxpr instead. It's also easier to make compatible! It will do some spurious checks that we don't need though.

```
def compat_cond(error_value, *args):
  error_value, out = checked_cond(error_value, *args)
  del error_value  # ignore error output
  return out
```

jax/_src/checkify.py
```diff
@@ -457,14 +459,23 @@ def assert_impl(pred, code, payload, *, msgs):
   Error(~pred, code, msgs, payload).throw()
   return []
 
-@assert_p.def_abstract_eval
+CheckEffect = enum.Enum('CheckEffect', ['CHECK'])
```
LenaMartens, line 462: do you see this as the preferred type for an effect? I wasn't sure what type to give it (in this case it will be an enum with only one value)

sharadmv: It doesn't necessarily have to be an enum -- the main requirement is that it has to be hashable. If you only have one effect (unlike debug printing, which uses two), you can just use a sentinel object.

sharadmv (approved):
Looks great!

PR #10865, opened 2022-05-27
[x64] make lax_numpy_test compatible with strict dtype promotion

jakevdp:
Part of landing #10840

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -118,6 +118,13 @@ def _get_y_shapes(y_dtype, shape, rowvar):
     return [(1, shape[-1]), (2, shape[-1]), (5, shape[-1])]
   return [(shape[0], 1), (shape[0], 2), (shape[0], 5)]
 
+
+def _strict_promotion_if_dtypes_match(dtypes):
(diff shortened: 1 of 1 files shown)


mattjj (approved):
Is the `_strict_promotion_if_dtypes_match` utility a temporary shim, to test new and old functionality at the same time, to be removed when some new behavior is the standard? Or is it permanent?

jakevdp:
I think it will be permanent. The issue is that in order for strict promotion mode to be useful, we need our library code to respect it. The best way to maintain that property is to run our tests with `jax_numpy_dtype_promotion=strict`. That said, a lot of our `jax.numpy` API & associated tests explicitly exercise non-strict promotion rules, so we need to disable strict promotion for these tests.

But if we disable standard promotion for *all* such tests, then we might miss cases where e.g. `jnp_func(int32, int32)` is implemented in a way that is incompatible with strict mode.

So what I landed on was to have this context manager for these cases, such that we set promotion to strict in cases where we know it's supported, and set promotion to standard in cases where we know the test exercises standard promotion rules.

What do you think?

Issue #10375, opened 2022-04-20
Implement `jax.scipy.stats.bootstrap`


JeppeKlitgaard:
I think it would be super neat to have a JAX version of the `scipy.stats.bootstrap` function: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bootstrap.html

riven314:
Hi @JeppeKlitgaard 
I would love to work on this ticket but I am a beginner contributor. Would you mind giving me some guidance to get started? 

As I looked through the related source code in scipy, my plan is to break down the ticket into 2 steps, see if you think it makes sense:
1. implement helper functions `_bootstrap_resample`, `_bca_interval` and `_percentile_along_axis` in JAX (`scipy.stats.bootstrap` internally makes use of them)
2. together with (1), I can translate the main logic of `scipy.stats.bootstrap` into JAX(not sure if it can be straight-forwardly translated into JAX atm, constrained by its statical compilation property)

cc @jakevdp as well to see if you could kindly shed me some light to get started

jakevdp:
Hi - that sounds like a good plan for implementation. Regarding `BootstrapResult` – I think it's just a simple dataclass or named tuple, you could create a similar container for the results in a JAX implementation, I'd lean toward NamedTuple because you get pytree flattening for free.

Regarding the return shape: it looks like it only depends on static properties of the inputs, so it should be fine for JAX.

JeppeKlitgaard:
Would be super neat if @riven314 would implement this. Otherwise I have a WIP that I will probably get around to finishing eventually.

riven314:
@JeppeKlitgaard 
I am working on this
if you have some WIP code for this ticket, would also be great to share this with me!

JeppeKlitgaard:
@riven314 I never got to a point that would be of much use to you, I'm afraid. I dislike the the way the original `scipy` code is written to such an extent that I wouldn't want to mirror that implementation, but rather see if it is possible to get the same API using a different approach behind the scenes.

riven314:
hi @jakevdp @JeppeKlitgaard 
I have made a draft in my fork repo but I found that I couldn't raise any PR here
so I just post my draft here: https://github.com/google/jax/compare/main...riven314:impl-jax-bootstrap

Would you mind helping me to do a prior review on my draft to see if it is in the right direction? (e.g. approach, coding style)
Pending work for me is to do inputs validation and unit tests. I also left a few questions in the code.

riven314:
[...]

PR #10871, opened 2022-05-28
WIP: Implementation of Scipy Bootstrap

riven314:
address https://github.com/google/jax/issues/10375

### DIFF
--- a/jax/_src/scipy/stats/bootstrap.py
+++ b/jax/_src/scipy/stats/bootstrap.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2020 Google LLC
```
jakevdp, line 1: 2022

jakevdp (commented):
Thanks, this looks like a good start! Mostly small comments, but the larger change I'd consider is to express the resamplings in terms of `vmap` rather than `scan` if possible.

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,170 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import NamedTuple
+
+import scipy.stats as osp_stats
```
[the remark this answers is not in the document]
riven314: updated

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,170 @@
...
+    2. `batch` arg is not needed because lax.scan auto parallelize the resampling
+    3. to save memory, it computes statistic right after resampling
+    """
+    n = sample[0].shape[-1]
+
+    def _resample_and_compute_once(rng, x):
+        idxs = random.randint(rng, shape=(n,), minval=0, maxval=n)
+        # `sample` is a tuple of sample sets, we need to apply same indexing on each sample set
+        resample = jax.tree_map(lambda data: data[..., idxs], sample)
+        next_rng = jax.random.split(rng, 1)[0]
```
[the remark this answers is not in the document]
riven314: As changed to use `vmap`, so now I do the splitting in the vmap argument: `jax.random.split(key, n_resamples)`

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,170 @@
...
+    alpha_2 = ndtr(z0_hat + num2 / (1 - a_hat * num2))
+    return alpha_1, alpha_2
+
+
+def _percentile_along_axis(theta_hat_b, alpha):
+    shape = theta_hat_b.shape[:-1]
+    alpha = jnp.broadcast_to(alpha, shape)
+    # QUESTION: is it good practice to use vmap here?
+    # TODO: may need to handle nan
+    # TODO: handle numeric discrepancy against scipy's _percentile_along_axis
```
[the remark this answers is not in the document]
riven314: should I use a control flow here to apply multiple `vmap` if I find that `theta_hat_b` have more than 2 dimensions?

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,170 @@
...
+    idxs = jnp.arange(n - 1)
+    miss_first_sample = sample[1:]
+    miss_last_sample = sample[:-1]
+
+    def _jackknife_resample_and_compute(idx, resample):
+        resample = jnp.where(idxs >= idx, miss_first_sample, miss_last_sample)
+        return idx + 1, statistic(resample)
+
+    # TODO: check if it can handle `statistic` that return multiple scalars
+    _, theta_hat_i = lax.scan(_jackknife_resample_and_compute, 0, jnp.ones(n))
```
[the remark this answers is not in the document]
riven314: for `_jackknife_resample_and_compute_statistic`, I weirdly find that `vmap` runs SLOWER than `scan`.
Here is the colab notebook that I did my benchmarking:
https://colab.research.google.com/drive/1abKv-zyI-CZ4BEtXc3oh5Ebx_rcFq0dp?usp=sharing

I am not sure if its because my implementation is inefficient, or I have made some errrors.

jax/_src/scipy/stats/bootstrap.py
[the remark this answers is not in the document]
jakevdp: You can use Python `for` loops to apply multiple `vmap` transforms; something like this:
```python
import jax
import jax.numpy as jnp

def f(x):
  return x.sum()

x = jnp.ones((2, 3, 4, 5))
for i in range(x.ndim - 1):
  f = jax.vmap(f)
print(f(x).shape)
# (2, 3, 4)
```

jax/_src/scipy/stats/bootstrap.py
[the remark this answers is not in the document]
jakevdp: Interesting! On CPU `scan` and friends are generally not too bad, but I suspect if you benchmarked on GPU/TPU you'd find `scan` to be much slower than `vmap`.

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,240 @@
...
+  low: jnp.ndarray
+  high: jnp.ndarray
+
+
+class BootstrapResult(NamedTuple):
+  confidence_interval: ConfidenceInterval
+  standard_error: jnp.ndarray
+
+
+def _bootstrap_resample_and_compute_statistic(sample, statistic, n_resamples, key):
```
jakevdp, line 49: This would be more consistent with JAX's typical approach if (1) `key` were the first argument, and (2) batching were handled via `vmap` at the call-site, rather than via an argument passed to the function. So the API would be something like this:
```python
def _bootstrap_resample_and_compute_statistic(key, sample, statistic):
  ...
```
and rather than calling it like
```python
_bootstrap_resample_and_compute_statistic(sample, statistic, n_resamples, key)
```
you could instead call it like
```python
keys = random.split(key, n_resamples)
vmap(_bootstrap_resample_and_compute_statistic, (0, None, None))(keys, sample, statistic)
```
The benefits would be (1) more explicit handling of key splitting by the user of the function, and (2) `vmap` at the outer levels may be somewhat more efficient (I'm not entirely sure on that, but I think it is the case) and (3) it's more maintainable, because it makes use of JAX's composable transforms in the way they're intended to be used, rather than hiding them behind less flexible batch-aware APIs.

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,240 @@
...
+    1. `batch` arg is not needed because lax.scan auto parallelize the resampling
+    2. to save memory, it computes statistic right after resampling
+    """
+  n = sample.shape[-1]
+  # these assume arg `sample` has > 1 entries
+  idxs = jnp.arange(n - 1)
+  miss_first_sample = sample[1:]
+  miss_last_sample = sample[:-1]
+
+  @vmap
```
jakevdp, line 83: Same comment here. Can we define `_jackknife_resample_and_compute_statistic` so it natively handles a single batch, and then use `vmap` as appropriate at the call-site?

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,240 @@
...
+  alpha_2 = ndtr(z0_hat + num2 / (1 - a_hat * num2))
+  return alpha_1, alpha_2
+
+
+def _percentile_along_axis(theta_hat_b, alpha):
+  shape = theta_hat_b.shape[:-1]
+  alpha = jnp.broadcast_to(alpha, shape)
+  vmap_percentile = jnp.percentile
+  for i in range(theta_hat_b.ndim - 1):
+    vmap_percentile = vmap(vmap_percentile)
```
jakevdp, line 125: Rather than vmapping, can we use the `axis` argument to `jnp.percentile`?

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,240 @@
...
+  percentiles = vmap_percentile(theta_hat_b, alpha)
+  return percentiles[()]
```
jakevdp, line 127: I don't understand the purpose of empty indexing here.

jax/_src/scipy/stats/bootstrap.py
```diff
@@ -0,0 +1,240 @@
...
+  key, _ = _check_prng_key(key)
+
+  return (key, data_iv, statistic, vectorized, paired, axis,
+          confidence_level, n_resamples, method)
+
+
+@_wraps(
+  scipy.stats.bootstrap,
+  lax_description=_replace_random_state_by_key_no_batch_jnp_statistic_doc,
+  skip_params=("batch",),
```
jakevdp, line 177: You can use `extra_params` here to document the `key` argument.

That said, I'm starting to wonder if this should really be considered a wrapper of `scipy.stats.bootstrap`, because it's API is now substantially different. In numpy's case, we don't provide any wrappers for `numpy.random` functionality, instead using a different key-based API in `jax.random`. I'm starting to think that the same treatment would make sense here, because as written `jax.scipy.bootstrap` must be called with a different signature than `scipy.bootstrap`.

It also would solve the issue of how to handle irrelevant params like `vectorized`, and we could write the API in a way that is more typical of JAX library functions (i.e. keep batching orthogonal to the implementation, rather than calling `vmap` within.

What do you think?

jakevdp (changes requested):

wonhyeongseo:
Hello, @riven314!
I'm interested in this PR, may I please address @jakevdp's reviews?
Thank you, and I hope you have a wonderful day!

riven314:
Hi @wonhyeongseo 
Thanks for your interest! Currently I don't have bandwidth on this ticket at least for these 2 months, so feel free to work on it! Otherwise, I can continue working on it whenever I retain my bandwidth.

A few notes about my PR (you can further discuss with @jakevdp to see if those points are valid to address):
1. the numerical discrepancy between my JAX implementation v.s. scipy's is NOT immaterial, despite ramping up the resampling number. Can't give you a precise statistics (discrepancy of ~0.01?) because it has been a few months ago, but such discrepancy would render unit test to fail
2. my JAX implementation is a bit slower than scipy's in some settings (CPU).
3. Scipy bootstrap supports a lot of different scenarios (e.g. multi-statistics, one-set/ multi-set/ paired samples, different bootstrap methods). For now I am not confident if my implementation works as expected under various scenarios.

Hope forward to your contribution to the ticket.

carlosgmartin:
@wonhyeongseo Are you still interested in tackling this?

wonhyeongseo:
Hello @carlosgmartin ! Not at the moment because I don't know how to tackle this. Would love to see your implementation! 😊

carlosgmartin:
@wonhyeongseo Thanks for letting us know.

@riven314 Would you like to continue working on your PR, or would you prefer someone else take over?

riven314:
@carlosgmartin 
Hi! I don't foresee I have bandwidth in the short run so I would appreciate if someone is willing to take over.
I am glad to explain my code if any help is needed!

jax/_src/scipy/stats/bootstrap.py
carlosgmartin: @jakevdp Is there a decision/consensus on whether to adhere to the original API?

jakevdp:
Thanks for working on this – I think given the discussion in https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html this would now be considered out-of-scope for JAX itself. Sorry we weren't able to merge your contribution!

PR #10872, opened 2022-05-28
Fix mesh -> Mesh in tutorials and docs

fehiepsi:
The context manager is renamed in [jax 0.3.8](https://jax.readthedocs.io/en/latest/changelog.html#jax-0-3-8-april-29-2022)


sharadmv (approved):
Thanks!

sharadmv:
Could you squash your commits?

PR #10901, opened 2022-05-31
CI: pin traitlets to <5.2  in doc build

jakevdp:
We can unpin when https://github.com/ipython/traitlets/issues/741 is fixed.


yashk2810 (approved):

JanB1:
Traitlets error is fixed, newest version of traitlets is `5.2.2.post1` @jakevdp!

PR #10906, opened 2022-05-31
Store profiler server as a global variable and add a `stop_server` function

sharadmv:
Currently, the user needs to store around the return value of `jax.profiler.start_server` or the server will immediately be destroyed. This change stores the server in a global variable and exposes a way of destroying it via `stop_server`.


hawkinsp (approved):
I might note the semantic change in the CHANGELOG.

sharadmv:
Thank you for the review!

PR #10913, opened 2022-06-01
[x64] make jnp.histogram and related functions work with strict promotion

jakevdp:
Re-lands #10907 after rollback in #10909


yashk2810:
My rollback just went in. I have applied the kokoro label. If all tests pass, I'll approve or you can apply the pull ready label directly (if I am offline)

yashk2810 (approved):

PR #10924, opened 2022-06-01
[x64] deprecate unsafe type casting in scatter-update operations

### LINKED ISSUE #10892: [x64] scatter implementation silently generates 64-bit computations
In cases where `scatter` operations are called on mismatched types, it can silently lead to 64-bit operations that are then cast back to the input types.

For example, this code:
```python
import jax.numpy as jnp
from jax.config import config
x = jnp.arange(5000, dtype='int32')
def f(x):
  return x.at[0].add(1.0)
```
will return an array of type int32, but within the implementation converts the array to float64 before converting it back:
```python
print(jax.make_jaxpr(f)(x))
# { lambda ; a:i32[5000]. let
#     b:bool[] = lt 0 0
#     c:i64[] = add 0 5000
#     d:i64[] = select_n b 0 c
#     e:i32[] = convert_element_type[new_dtype=int32 weak_type=False] d
#     f:i32[1] = broadcast_in_dim[broadcast_dimensions=() shape=(1,)] e
#     g:f64[5000] = convert_element_type[new_dtype=float64 weak_type=True] a
#     h:f64[5000] = scatter-add[
#       dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
#       indices_are_sorted=True
#       mode=GatherScatterMode.FILL_OR_DROP
#       unique_indices=True
#       update_consts=()
#       update_jaxpr={ lambda ; i:f64[] j:f64[]. let k:f64[] = add i j in (k,) }
#     ] g f 1.0
#     l:i32[5000] = convert_element_type[new_dtype=int32 weak_type=False] h
#   in (l,) }
```
The reason for this is the conversion here: https://github.com/google/jax/blob/fd61aca5a10c06c90f553460edf0efc1e8268845/jax/_src/ops/scatter.py#L93

followed by the cast here: https://github.com/google/jax/blob/fd61aca5a10c06c90f553460edf0efc1e8268845/jax/_src/ops/scatter.py#L114

The fix for this is probably relatively straightforward (rather than calling `promote_dtypes`, just cast `y` to the type of `x`) but it would probably change the behavior of `scatter` in some corner cases.

jakevdp:
Fixes #10892

This is a breaking change – at the very least, we'll need to fix some internal tests and some downstream users. Also, perhaps this should involve a deprecation cycle?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,11 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
     traces as an alternative to the Tensorboard UI.
   * Added a `jax.named_scope` context manager that adds profiler metadata to
     Python programs (similar to `jax.named_call`).
+  * In scatter-update operations (i.e. :attr:`jax.numpy.ndarray.at`), unsafe implicit
+    dtype casts are deprecated, and now result in a `FutureWarning`.
+    In a future release, this will become an error. An example of an unsafe implicit
+    cast is `jnp.zeros(4, dtype=int).at[0].set(1.5)`, in which `1.5` previously was
+    silently truncated to `1`.
 
 ## jaxlib 0.3.11 (Unreleased)
 * [GitHub commits](https://github.com/google/jax/compare/jaxlib-v0.3.10...main).
--- a/jax/_src/ops/scatter.py
+++ b/jax/_src/ops/scatter.py
@@ -16,6 +16,7 @@
 
 import sys
 from typing import Any, Callable, Optional, Sequence, Tuple, Union
+import warnings
 
 import numpy as np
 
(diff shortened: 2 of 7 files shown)


hawkinsp (approved):

hawkinsp:
I'm not sure about the backward compatibility part: perhaps it's worth deprecating this slowly. But the change itself seems fine.

jakevdp:
Test failures raised a question here. Should this be a failure, or should it be allowed?
```python
x = jnp.arange(5, dtype='int32')
y = jnp.array(1, dtype='int64')
x.at[0].set(y)
```
In the current version of this PR, this fails because `(int32, int64)` promotes to `int64`. But perhaps for same-kind dtypes, we should allow casing `y` to `y.astype(x.dtype)`? What do you think?

jakevdp:
I changed the error to a `FutureWarning` for now, with a message that it will become an error in a future release.

PR #10934, opened 2022-06-01
Disallow complex dtypes for lax ops rem, lt, le, gt, and ge.

### LINKED ISSUE #10932: Abort trap in jax.numpy code
```python
import jax
import jaxlib

import jax.numpy as jnp

print(f"{jax.__version__=}")
print(f"{jaxlib.__version__=}")

x = jnp.arange(4)
y = jnp.arange(20)
z = jnp.ones(20, dtype='complex64')

print(jnp.interp(x, y, z, period=0.59))
```
output:
```
jax.__version__='0.3.14'
jaxlib.__version__='0.3.10'
2022-06-01 13:06:02.826868: F external/org_tensorflow/tensorflow/compiler/xla/service/hlo_evaluator.cc:137] unhandled direction for conversion to Comparison: LT
Abort trap: 6
```

hawkinsp:
In none of these cases did the compiler do anything sensible with complex values anyway.

Note that complex comparisons are still allowed by the jax.numpy layer (and are defined to have lexicographic comparison semantics).

Fixes #10932

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1552,6 +1552,7 @@ _bool = {np.bool_}
 _num = _int | _float | _complex
 _any = _int | _float | _complex | _bool
 _bool_or_int = _int | _bool
+_ordered = _int | _float | _bool
 
 neg_p = standard_unop(_num, 'neg')
 ad.deflinear2(neg_p, lambda t, operand: [neg(t)])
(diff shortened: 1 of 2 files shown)


jax/_src/lax/lax.py
```diff
@@ -2171,19 +2171,21 @@ def _compare_lower_mhlo(direction: str, ctx, x, y):
 ad.defjvp_zero(ne_p)
 mlir.register_lowering(ne_p, partial(_compare_lower_mhlo, "NE"))
 
-ge_p = naryop(_fixed_dtype(np.bool_), [_any, _any], 'ge')
+_ordered = _int | _float | _bool
```
jakevdp, line 2174: Perhaps define this above, where `_num`, `_any`, etc. are defined

jakevdp (approved):

jax/_src/lax/lax.py

PR #10937, opened 2022-06-01
Added random.gnormal and random.ball.

carlosgmartin:


jax/_src/random.py
```diff
@@ -1623,3 +1623,58 @@ def orthogonal(
...
+    p: a float representing the shape parameter.
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified shape and dtype.
+  """
+  _check_shape("gnormal", shape)
+  p = core.concrete_or_error(float, p, "The error occurred in jax.random.gnormal()")
```
jakevdp, line 1646: I don't think we need this to be concrete; I believe `gamma` allows dynamic values for its shape parameter.

jax/_src/random.py
```diff
@@ -1623,3 +1623,58 @@ def orthogonal(
...
+    shape: optional, the batch dimensions of the result. Default ().
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array of shape (*shape, d) and specified dtype.
+  """
+  _check_shape("ball", shape)
+  d = core.concrete_or_error(index, d, "The error occurred in jax.random.ball()")
+  p = core.concrete_or_error(float, p, "The error occurred in jax.random.ball()")
```
jakevdp, line 1676: Same here, no need for `p` to be concrete.

tests/random_test.py
```diff
@@ -1056,6 +1056,40 @@ def testOrthogonal(self, n, shape, dtype):
...
+       .format(p, jtu.format_shape_dtype_string(shape, dtype)),
+       "p": p,
+       "shape": shape,
+       "dtype": dtype}
+      for p in [.5, 1., 1.5, 2., 2.5]
+      for shape in [(), (5,), (10, 5)]
+      for dtype in jtu.dtypes.floating))
+  def testGnormal(self, p, shape, dtype):
+    key = self.seed_prng(0)
+    samples = random.gnormal(key, p, shape, dtype)
```
jakevdp, line 1070: We should test with and without `jit` (see the `rand`, `crand` pattern used elsewhere in this file)

tests/random_test.py
```diff
@@ -1056,6 +1056,40 @@ def testOrthogonal(self, n, shape, dtype):
...
+      for d in range(4)
+      for p in [.5, 1., 1.5, 2., 2.5]
+      for shape in [(), (5,), (10, 5)]
+      for dtype in jtu.dtypes.floating))
+  def testBall(self, d, p, shape, dtype):
+    key = self.seed_prng(0)
+    samples = random.ball(key, d, p, shape, dtype)
+    self.assertEqual(samples.shape, (*shape, d))
+    self.assertEqual(samples.dtype, dtype)
+    self.assertTrue(((jnp.abs(samples) ** p).sum(-1) <= 1).all())
```
jakevdp, line 1091: Is there any statistic we can use to gauge the correctness of the samples? I've not looked closely at the algorithm, and this test doesn't provide a strong constraint. 

For example, am I right in thinking the radial CDF should equal the hyperspherical volume formula as a function of radius? We could compute that using gamma functions and compare it with a KS test.

jakevdp (changes requested):
Looks good! A few comments below.

tests/random_test.py
```diff
@@ -1056,6 +1056,46 @@ def testOrthogonal(self, n, shape, dtype):
...
+       "p": p,
+       "shape": shape,
+       "dtype": dtype}
+      for d in range(1, 5)
+      for p in [.5, 1., 1.5, 2., 2.5]
+      for shape in [(), (5,), (10, 5)]
+      for dtype in jtu.dtypes.floating))
+  def testBall(self, d, p, shape, dtype):
+    key = self.seed_prng(0)
+    samples = random.ball(key, d, p, shape, dtype)
```
jakevdp, line 1092: Can we test both compiled and uncompiled versions here as well?

jakevdp (changes requested):
Looks great! One last comment

jakevdp (approved):
Looks great, thanks!

jakevdp:
Actually, looks like the readthedocs failure needs to be addressed. This is the error, which comes from a mis-formatted dosctring:
```
/home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/10937/jax/_src/random.py:docstring of jax._src.random.ball:17: WARNING: Inline emphasis start-string without end-string.
```

carlosgmartin:
> Actually, looks like the readthedocs failure needs to be addressed. This is the error, which comes from a mis-formatted dosctring:
> 
> ```
> /home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/10937/jax/_src/random.py:docstring of jax._src.random.ball:17: WARNING: Inline emphasis start-string without end-string.
> ```

carlosgmartin:
@jakevdp Fixed.

jakevdp (approved):
Looks great!

PR #10948, opened 2022-06-02
add `scipy.special.factorial` and `scipy.special.comb`

froystig:
I found a need for `comb` in a Threefry PRNG change that I'm working on...

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -1465,3 +1465,22 @@ def expn_jvp(n, primals, tangents):
 def exp1(x):
   (x,) = _promote_args_inexact("exp1", x)
   return expn(1, x)
+
+
+@_wraps(osp_special.factorial)
+def factorial(n, exact=False):
+  if exact:
+    raise NotImplementedError('scipy.special.factorial with exact=True')
+  return jnp.where(n > 0., jnp.exp(gammaln(n + 1.)), 0.)
+
+
+@_wraps(osp_special.comb)
+def comb(N, k, exact=False, repetition=False):
+  if exact:
+    raise NotImplementedError('scipy.special.comb with exact=True')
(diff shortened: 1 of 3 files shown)


jax/_src/scipy/special.py
```diff
@@ -1465,3 +1465,22 @@ def expn_jvp(n, primals, tangents):
 def exp1(x):
   (x,) = _promote_args_inexact("exp1", x)
   return expn(1, x)
+
+
+@_wraps(osp_special.factorial)
```
jakevdp, line 1470: I would use the `lax_description` argument to `_wraps` to mention that `exact` is not implemented.

jax/_src/scipy/special.py
```diff
@@ -1465,3 +1465,22 @@ def expn_jvp(n, primals, tangents):
...
+def factorial(n, exact=False):
+  if exact:
+    raise NotImplementedError('scipy.special.factorial with exact=True')
+  return jnp.where(n > 0., jnp.exp(gammaln(n + 1.)), 0.)
+
+
+@_wraps(osp_special.comb)
```
jakevdp, line 1477: Same here

jakevdp (approved):
Looks good. To make these appear in the docs, they should be added to the list here: https://github.com/google/jax/blob/main/docs/jax.scipy.rst#jaxscipyspecial

Also, it would be worth adding a changelog entry for these as well

quantshah:
Hi, how is this PR moving along? Is there something I can do to help? I came here as I was looking for a way to make double_factorial work with JIT/VMAP. Is it possible to do so with something similar by approximating in the following way using a Stirling approximation for the factorial?

(2n-1)!! = (2n)!/((2^n)n!)

hawkinsp:
Closing as stale.

PR #10950, opened 2022-06-02
[jax2tf] Added a test for SavedModel with pytrees

gnecula:
This was in response to an external user running into problems
with converting and saving a SavedModel for a function where
one of the arguments is a dict. This test demonstrates that
this works, both for the conversion, and for the saving/restoring.


gnecula:
@rxsang please take a look

gnecula:
@voutcn PTAL



[voutcn](https://github.com/voutcn)

rxsang (approved):

PR #10960, opened 2022-06-03
Refactor `control_flow.py` into several smaller pieces

sharadmv:
`jax/_src/lax/control_flow.py` is quite a big file (~3.5k LOC) and contains all sorts of primitives, ranging from `scan` to `cumsum` and a `remat` impl rule. We're planning on exploring a dedicated `for` primitive so rather than concatenate another complex implementation to this file, I figured it might be time to do a refactor of sorts. 

This change moves the various components of `control_flow.py` into `loops.py`, `conditionals.py`, `solves.py`, `remat_impl.py`, and `common.py`. It turns `control_flow` into a directory and exposes the same imports as the original `control_flow.py` so the rest of JAX can remain unchanged.


froystig (approved):
Nice work!

I tried to think of a name other than `remat_impl`, since the module contains other rules as well (translation and abstract eval), but I couldn't think of anything great.

Issue #3112, opened 2020-05-15
sparse solvers


mganahl:
Hi!

Our team is currently developing sparse symmetric and non-symmetric solvers in JAX (implicitly restarted arnoldi and lanzcos). I have it from hearsay that there are efforts to support this natively in JAX. I wanted to ask what the status there is. It would be actually great if JAX would provide those.

shoyer:
By "solvers" are you referring specifically to eigenvalue solvers?

Yes, I think we would be interested in adding some of these to JAX, but they would need to be reasonably robust and scalable, ideally comparable to the SciPy algorithms. The "basic" Lanczos method would probably not qualify, but restarted Arnoldi (used by SciPy in ARPACK) probably would.

It would also be nice to include auto-diff rules for these operations, but that could come later.

jakevdp:
Since scipy comes bundled with ARPACK, it may be possible to write a CPU translation rule to use the bundled ARPACK version for JAX (fun fact: my first scipy contribution involved the ARPACK wrappers :smile:). GPU & TPU would have to use another approach, though.

shoyer:
I played around with writing a thick-restart Lanczos solver. As a point of reference, you can find it in https://github.com/google/jax/pull/3114 but I'm not sure we actually want to merge it.

Thenerdstation:
We have an implementation at https://github.com/google/TensorNetwork/blob/d4ec0a381dbf1a7d453d97525ccc857fc416b575/tensornetwork/backends/jax/jax_backend.py#L230

Not sure if it would useful to have in JAX proper.

mganahl:
This version does not support restarts right now, and is in general geared towards tensornetwork applications. It could serve as a starting point for an implicitly restarted lanczos.

mganahl:
I had a look at #3114, and this would indeed be interesting for us. The three most prevalent problems that pop up frequently in our applications are sparse symmetric (mostly SA or LA eigenvalues, usually only one to a few), sparse non-symmetric (with LR eigenvalues), and linear system solvers (symmetric and non-symmetric, i.e. gmres or lgmres), all of which we are currently working on.
The requirement in terms of robustness and accuracy are often less rigorous for tensor networks than for other applications, hence our implementations are more mundane. But if the JAX team was interested as well in supporting those, it could make sense to join efforts.

shoyer:
GMRES and L-GMRES for linear solves would also be quite welcome in JAX. These would fit in quite naturally alongside our new CG solver.

[...]

PR #10962, opened 2022-06-03
Add initial LOBPCG top-k eigenvalue solver (#3112)

vlad17:
This initial version is f32-only for accelerators, since it relies on an eigh call (which itself is f32 at most) in its inner loop.

For details, see jax.experimental.linalg.standard_lobpcg documentation.

### DIFF
--- a/jax/experimental/sparse/linalg.py
+++ b/jax/experimental/sparse/linalg.py
@@ -1,0 +1,504 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sparse linear algebra routines."""
+
+from typing import Union, Callable
+import functools
+
+import jax
+import jax.numpy as jnp
+
+import numpy as np
+
+def lobpcg_standard(
+    A: Union[jnp.ndarray, Callable[[jnp.ndarray], jnp.ndarray]],
+    X: jnp.ndarray,
+    m: int = 100,
+    tol: Union[jnp.ndarray, float, None] = None):
+  """Compute the top-k standard eigenvalues using the LOBPCG routine.
+
+  LOBPCG [1] stands for Locally Optimal Block Preconditioned Conjugate Gradient.
+  The method enables finding top-k eigenvectors in an accelerator-friendly
+  manner.
+
+  This initial experimental version has several caveats.
+
+    - Only the standard eigenvalue problem `A U = lambda U` is supported,
+      general eigenvalues are not.
+    - Gradient code is not available.
+    - f64 will only work where jnp.linalg.eigh is supported for that type.
+    - Finding the smallest eigenvectors is not yet supported. As a result,
+      we don't yet support preconditioning, which is mostly needed for this
+      case.
+
+  The implementation is based on [2] and [3]; however, we deviate from these
+  sources in several ways to improve robustness or facilitate implementation:
+
+    - Despite increased iteration cost, we always maintain an orthonormal basis
+      for the block search directions.
+    - We change the convergence criterion; see the `tol` argument.
+    - Soft locking [4] is intentionally not implemented; it relies on
+      choosing an appropriate problem-specific tolerance to prevent
+      blow-up near convergence from catastrophic cancellation of
+      near-0 residuals. Instead, the approach implemented favors
+      truncating the iteration basis.
+
+  [1]: http://ccm.ucdenver.edu/reports/rep149.pdf
+  [2]: https://arxiv.org/abs/1704.07458
+  [3]: https://arxiv.org/abs/0705.2626
+  [4]: DOI 10.13140/RG.2.2.11794.48327
+
+  Args:
+    A : An `(n, n)` array representing a square Hermitian matrix or a
+        callable with its action.
+    X : An `(n, k)` array representing the initial search directions for the `k`
+        desired top eigenvectors. This need not be orthogonal, but must be
+        numerically linearly independent (`X` will be orthonormalized).
+        Note that we must have `0 < k * 5 < n`.
+    m : Maximum integer iteration count; LOBPCG will only ever explore (a
+        subspace of) the Krylov basis `{X, A X, A^2 X, ..., A^m X}`.
+    tol : A float convergence tolerance; an eigenpair `(lambda, v)` is converged
+          when its residual L2 norm `r = |A v - lambda v|` is below
+          `tol * 10 * n * (lambda + |A v|)`, which
+          roughly estimates the worst-case floating point error for an ideal
+          eigenvector. If all `k` eigenvectors satisfy the tolerance
+          comparison, then LOBPCG exits early. If left as None, then this is set
+          to the float epsilon of `A.dtype`.
+
+  Returns:
+    `theta, U, i [, diagnostics]`, where `theta` is a `(k,)` array
+    of eigenvalues, `U` is a `(n, k)` array of eigenvectors, `i` is the
+    number of iterations performed, and `diagnostics` is a dictionary with debug
+    information, which is only returned if `debug` is set to true.
+
+  Raises:
+    ValueError : if `A,X` dtypes or `n` dimensions do not match, or `k` is too
+                 large (only `k * 5 < n` supported), or `k == 0`.
+  """
+  # Jit-compile once per matrix shape if possible.
+  if isinstance(A, (jnp.ndarray, np.ndarray)):
+    return _lobpcg_standard_matrix(A, X, m, tol, debug=False)
+  return _lobpcg_standard_callable(A, X, m, tol, debug=False)
+
+@functools.partial(jax.jit, static_argnames=['m', 'debug'])
+def _lobpcg_standard_matrix(
+    A: jnp.ndarray,
+    X: jnp.ndarray,
+    m: int,
+    tol: Union[jnp.ndarray, float, None],
+    debug: bool = False):
+  """Computes lobpcg_standard(), possibly with debug diagnostics."""
+  return _lobpcg_standard_callable(
+      functools.partial(_mm, A), X, m, tol, debug)
+
+@functools.partial(jax.jit, static_argnames=['A', 'm', 'debug'])
+def _lobpcg_standard_callable(
+    A: Callable[[jnp.ndarray], jnp.ndarray],
+    X: jnp.ndarray,
+    m: int,
+    tol: Union[jnp.ndarray, float, None],
+    debug: bool = False):
+  """Supports generic lobpcg_standard() callable interface."""
+
+  # TODO(vladf): support mixed_precision flag, which allows f64 Rayleigh-Ritz
+  # with f32 inputs.
+
+  n, k = X.shape
+  dt = X.dtype
+
+  _check_inputs(A, X)
+
+  if tol is None:
+    tol = jnp.finfo(dt).eps
+
+  X = _orthonormalize(X)
+  P = _extend_basis(X, X.shape[1])
+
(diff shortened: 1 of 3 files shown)


vlad17:
@shoyer @tabakg @lobpcg would any of you be interested in reviewing?

tests/lobpcg_test.py
```diff
@@ -0,0 +1,238 @@
...
+class LobpcgTest(jtu.JaxTestCase):
+  # Override base class default config.
+  _default_config = {
+      'jax_enable_checks': True,
+      'jax_debug_nans': True,
+      'jax_numpy_rank_promotion': 'raise',
+      'jax_traceback_filtering': 'off',
+  }
+
+  # TODO(vladf): add f64 tests just to verify it compiles?
```
tabakg, line 102: I think that would be good -- you could probably check if it's running on CPU and disable otherwise?

shoyer:
A couple high level thoughts:

1. It would be nice to support matrix-free linear operators defined on pytrees, like the solvers in jax.scipy.sparse.linalg. For example, imagine solving for the largest eigenvalue pairs of a function that evaluates a linearized neural network.
2. It would be nice to support an iterative interface with init and update functions, similar to the optimizers in Optax. This sort of inversion of control provides valuable flexibility in cases where a full eigenvalue solve from scratch is prohibitively expensive. For example, imagine performing a single LOBPCG step to re-estimate the top eigenvalue after each gradient descent step when training a neural net, as is done in spectral normalization.

Neither of these are deal breakers for the first iteration, though.

vlad17:
> It would be nice to support matrix-free linear operators defined on pytrees, like the solvers in jax.scipy.sparse.linalg. For example, imagine solving for the largest eigenvalue pairs of a function that evaluates a linearized neural network.

Great idea, this should be easy to add and now that I think about it, and I think Shankar (skrishnan@google.com) was another user who'd immediately benefit from an interface like that.

> It would be nice to support an iterative interface with init and update functions, similar to the optimizers in Optax. This sort of inversion of control provides valuable flexibility in cases where a full eigenvalue solve from scratch is prohibitively expensive. For example, imagine performing a single LOBPCG step to re-estimate the top eigenvalue after each gradient descent step when training a neural net, as is done in spectral normalization.

Could you elaborate on this? I'd be very willing to do this as a follow-on if I had a user to work with on their ideal API for this, but as-stated I don't quite see how setting the initial `X` to the previous value and maximum iteration `m=1` wouldn't work.

shoyer:
> > It would be nice to support an iterative interface with init and update functions, similar to the optimizers in Optax. This sort of inversion of control provides valuable flexibility in cases where a full eigenvalue solve from scratch is prohibitively expensive. For example, imagine performing a single LOBPCG step to re-estimate the top eigenvalue after each gradient descent step when training a neural net, as is done in spectral normalization.
> 
> Could you elaborate on this? I'd be very willing to do this as a follow-on if I had a user to work with on their ideal API for this, but as-stated I don't quite see how setting the initial `X` to the previous value and maximum iteration `m=1` wouldn't work.

Wouldn't we want to calculate the matrix P from the previous iteration?

vlad17:
Ah, gotcha. I guess that makes sense, and it seems like it'd be a matter of exposing the `body` function in the current method. Maybe it'd be best to wait until the API solidifies a little though, since `largest=False` option is incoming, and it'd be good to figure out how that iterative interface would work with a preconditioner.

tests/lobpcg_test.py
vlad17: It looks like there are `f64` tests auto-triggered by the github action matrix, so I think all I should  really need to do here is adjust the epsilons.

vlad17:
@shoyer @tabakg just updated the pr on a flight home, let me know what you think of the new interface and tests.

vlad17:
Here are some cool curves (which can be generated from the tests) of jax vs scipy f32 on 1000x1000 versions of the test matrices for top-10 eigs:

linear, geom, and ring laplacian

clustered eigenvals

vlad17:
@lobpcg I used 1.8.0. The examples are all the same as the unit tests in the PR but size 1000.

The actual colab to make the viz depends on some Google-internal features for unrelated things, but I can try to find some time to clean up the notebook for a public-facing version. Would it be more appropriate to post that as a scipy github issue?

lobpcg:
> @lobpcg I used 1.8.0 (did anything change in 1.8.1?). The examples are all the same as the unit tests in the PR but size 1000.
> 
> The actual colab to make the viz depends on some Google-internal features for unrelated things, but I can try to find some time to clean up the notebook for a public-facing version. Would it be more appropriate to post that as a scipy github issue? Just to keep this thread focussed on jax.

1.8.0 is representative. I made changes in lobpcg specifically for float32, but that was before.

Of course if you create the reproducible issue in SciPy that would be ideal. No need to include the code to make the actual  plots, just please add a reference to your post with the plots above. 

If SciPy fails on smaller sizes like 100, all those would be good examples to include into SciPy as unit tests, if I find a fix so that they all run.

vlad17:
@lobpcg filled out https://github.com/scipy/scipy/issues/16408 with just the cases, no viz.

jax/experimental/sparse/linalg.py
```diff
@@ -0,0 +1,388 @@
...
+    tol: Union[jnp.ndarray, float, None] = None):
+  """Compute the top-k standard eigenvalues using the LOBPCG routine.
+
+  LOBPCG [1] stands for Locally Optimal Block Preconditioned Conjugate Gradient.
+  The method enables finding top-k eigenvectors in an accelerator-friendly
+  manner.
+
+  This initial experimental version has several caveats.
+
+    - Only the standard eigenvalue problem `A U = lambda U` is supported,
```
rmlarsen, line 38: This MR seems a little odd to me. The power of LOBPCG is primarily on finding the smallest eigenvalues when a preconditioner for A is known. For finding the largest, the (block) Lanczos algorithm is provably optimal and well-understood.

jax/experimental/sparse/linalg.py
```diff
@@ -0,0 +1,388 @@
...
+  # such that (2) V is (S.T S)-orthonormal.
+  # https://www.netlib.org/lapack/lug/node54.html
+  #
+  # Usually it requires solving the complicated standard eigensystem
+  # U^-T S^T A S U^-1 @ Q = w * Q and then backsolving V = U^-1 Q,
+  # but if S is standard orthonormal then we just need to find
+  # eigenvalues of S.T A S.
+
+  SAS = _mm(S.T, A(S))
+  SAS = (SAS + SAS.T) / 2
```
rmlarsen, line 364: Explain why we need to symmetrize. I've seen cases where it is actually detrimental to accuracy (assuming the eigensolver only uses one triangle of the input. Did you try to do it without?

jax/experimental/sparse/linalg.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def _extend_basis(X, m):
+  # Use a block householder reflector to generate orthogonal extension
+  # to X. There's nothing too special about this, and we could choose
+  # any random extension to X's basis, but this is a deterministic choice.
+  #
+  # https://epubs.siam.org/doi/abs/10.1137/0725014
+  # https://www.jstage.jst.go.jp/article/ipsjdc/2/0/2_0_298/_article
+  n, k = X.shape
```
rmlarsen, line 378: Please write out the math here instead of just citing the papers. The code is not particularly readable with all the concatenations etc.

rmlarsen (changes requested):
I

jax/experimental/sparse/linalg.py
vlad17: Discussed offline: this PR adds just top-k to make the first change as small as we can get.

The extra preconditioner param and support for k-smallest eigenvals will come in a follow up PR(s) which will help reduce reviewer load for testing/locs etc.

jax/experimental/sparse/linalg.py
vlad17: done

jax/experimental/sparse/linalg.py
```diff
@@ -291,20 +294,51 @@ def _eigh_ascending(A):
...
-  # a degenerate, non-orthogonal one.
+  """Derives a truncated orthonormal basis for `X`.
 
+  SVQB [1] is an accelerator-friendly orthonormalization procedure, which
+  squares the matrix `C = X.T @ X` and computes an eigenbasis for a smaller
+  `(k, k)` system; this offloads most of the work in orthonormalization
+  to the first multiply when `n` is large.
+
+  Importantly, if diagonalizing the squared matrix `C` reveals numerical
+  degeneracy in its spectrum (which would be evidenced by near-0 eigenvalues),
```
rmlarsen, line 305: s/degeneracy in its spectrum/rank deficiency of X/

rmlarsen (approved):

jax/experimental/sparse/linalg.py
```diff
@@ -291,20 +294,51 @@ def _eigh_ascending(A):
 
 
 def _svqb(X):
-  # https://sdm.lbl.gov/~kewu/ps/45577.html
-  # May choose to truncate the basis rather than return
-  # a degenerate, non-orthogonal one.
+  """Derives a truncated orthonormal basis for `X`.
 
+  SVQB [1] is an accelerator-friendly orthonormalization procedure, which
+  squares the matrix `C = X.T @ X` and computes an eigenbasis for a smaller
```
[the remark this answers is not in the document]
rmlarsen: Rank revealing QR would be a nice options if it was available in JAX. It might be worth adding. A version based on the randomized projection approach could be made accelerator-friendly.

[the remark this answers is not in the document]
vlad17: By QR here, you mean QR-ing `X` itself not its square right? At one point early in the algorithm development process I tried that but decided against it b/c the speed wasn't satisfactory. Cholesky on XTX would be faster but rank deficiency is specifically a case that needs to be handled here.

That said, I'll put trying the QR approach out again on my TODO list. It's worth revisiting and I don't have hard numbers saying it's too slow. And it avoids squaring.

I'm really intrigued by the randomized projection you're mentioning. Do you have a reference?

jax/experimental/sparse/linalg.py
```diff
@@ -291,20 +294,51 @@ def _eigh_ascending(A):
...
 
   inner = _mm(X.T, X)
 
   w, V = _eigh_ascending(inner)
 
+  # All mask logic is used to avoid divide-by-zeros when input columns
+  # may have been zero or new zero columns introduced from truncation.
+  #
+  # If an eigenvalue is less than max eigvalue * eps, then consider
+  # that direction degenerate.
```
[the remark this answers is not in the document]
vlad17: My usage here is bleeding in from optimization where you'd refer to non-full-rank constraints as a degenerate problem. Regardless, I put it in quotes.

jax/experimental/sparse/linalg.py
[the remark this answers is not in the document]
rmlarsen: Here is a reference for the RRQR with randomized projection: https://arxiv.org/abs/2008.04447

You are right that computing the QR decomposition of X is likely slower, since `X.T@X` is so fast on an accelerator. But avoiding the squaring would be nice. Just some things to experiment with, I guess.

jax/experimental/sparse/linalg.py
```diff
@@ -0,0 +1,388 @@
...
+
+def _project_out(basis, U):
+  # "twice is enough" from shoyer's reference:
+  # http://slepc.upv.es/documentation/reports/str1.pdf
+
+  for _ in range(2):
+    U -= _mm(basis, _mm(basis.T, U))
+    U = _orthonormalize(U)
+
+  # It's crucial to end on a subtraction of the original basis.
```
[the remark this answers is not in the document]
rmlarsen: Hmm this is very interesting. If you allow these small non-zero components in the complement of the span of the large column to be normalized, it can act as a kind on randomized restarting, but I agree that it is difficult to control. I used it in Lanczos (PROPACK) to recover from computing an invariant subspace before convergence, but it is tricky.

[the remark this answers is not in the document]
vlad17: Indeed! I hope to find time to at least dig into it to do an internal presentation on it, it seemed like a nice new learning from implementing this.

One approach that also worked, but had slightly worse quality, was to leave these "random restarts" in and do another svqb() to orthonormalize the Ritz basis (this is also more expensive b/c the whole 3k x 3k system needs to be mutually orthonormalized).

PR #10964, opened 2022-06-03
Add `python -m jax.collect_profile` for manual trace capture

sharadmv:
This is an alternative to the Tensorboard UI for manually capturing traces from the JAX profiling server.

Basic usage:
1. Start the profiler server in your script: `jax.profiler.start_server(<port>)`
2. Run the `collect_profile` script, pointing to your selected port:
```bash
python -m jax.collect_profile <port> <duration_in_ms>
```

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -41,6 +41,29 @@ or if you're using Google Cloud:
 $ gcloud compute ssh <machine-name> -- -L 9001:127.0.0.1:9001
 ```
 
+### Manual capture
+
+Instead of capturing traces programmatically using `jax.profiler.trace`, you can
+instead start a profiling server in the script of interest by calling
+`jax.profiler.start_server(<port>)`. If you only need the profiler server to be
+active for a portion of your script, you can shut it down by calling
+`jax.profiler.stop_server()`.
+
+Once the script is running and after the profiler server has started, we can
+manually capture an trace by running:
+```bash
+$ python -m jax.collect_profile <port> <duration_in_ms>
+```
+
+By default, the resulting trace information is dumped into a temporary directory
+but this can be overridden by passing in `--log_dir=<directory of choice>`. 
+Also, by default, the program will prompt you to open a link to
+`ui.perfetto.dev`. When you open the link, the Perfetto UI will load the trace
+file and open a visualizer. This feature is disabled by passing in
+`--no_perfetto_link` into the command. Alternatively, you can also point
+Tensorboard to the `log_dir` to analyze the trace (see the
+"Tensorboard Profiling" section below).
+
 ## TensorBoard profiling
 
 [TensorBoard's
--- a/jax/collect_profile.py
+++ b/jax/collect_profile.py
@@ -1,0 +1,107 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import gzip
+import os
+import pathlib
+import tempfile
+
+from typing import Optional
+
+# pytype: disable=import-error
+import jax
+try:
+  from tensorflow.python.profiler import profiler_v2 as profiler
+  from tensorflow.python.profiler import profiler_client
+except ImportError:
+  raise ImportError("This script requires `tensorflow` to be installed.")
+try:
+  from tensorboard_plugin_profile.convert import raw_to_tool_data as convert
+except ImportError:
+  raise ImportError(
+      "This script requires `tensorboard_plugin_profile` to be installed.")
+# pytype: enable=import-error
+
+
+_DESCRIPTION = """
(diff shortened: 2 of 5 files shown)


docs/profiling.md
```diff
@@ -41,6 +41,27 @@ or if you're using Google Cloud:
...
+`jax.profiler.stop_server()`.
+
+Once the script is running and after the profiler server has started, we can
+manually capture an trace by running:
+```bash
+$ python -m jax.collect_profile <port> <duration_in_ms>
+```
+
+By default, the resulting trace information is dumped into a temporary directory
+but this can be overridden by passing in `--log_dir=<directory of choice>`.
```
skye, line 59: Mention how to open the trace in TensorBoard using the log directory, or link to the TensorBoard section.

jax/collect_profile.py
```diff
@@ -0,0 +1,90 @@
...
+      args.duration_in_ms,
+      options=options)
+  print(f"Dumped profiling information in: {log_dir}")
+  curr_path = log_dir.resolve()
+  root_trace_folder = curr_path / "plugins" / "profile"
+  trace_folders = [root_trace_folder / trace_folder for trace_folder
+                   in root_trace_folder.iterdir()]
+  latest_folder = max(trace_folders, key=os.path.getmtime)
+  xplane = next(latest_folder.glob("*.xplane.pb"))
+  result = convert.xspace_to_tool_data([xplane], "trace_viewer^", None)
```
skye, line 81: What is this doing?

jax/collect_profile.py
```diff
@@ -0,0 +1,90 @@
...
+parser.add_argument("--no_perfetto_link",
+                    help="Disable creating a perfetto link", action="store_true")
+parser.add_argument("--host", default="127.0.0.1",
+                    help="Host to collect trace. Defaults to 127.0.0.1",
+                    type=str)
+parser.add_argument("--host_tracer_level", default=2,
+                    help="Profiler host tracer level", type=int)
+parser.add_argument("--device_tracer_level", default=1,
+                    help="Profiler device tracer level", type=int)
+parser.add_argument("--python_tracer_level", default=0,
```
skye, line 59: JAX defaults to python_tracer_level=1: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/python/profiler.cc#L41

I would also `set_enable_hlo_proto(true)` or create an option

jax/collect_profile.py
```diff
@@ -0,0 +1,90 @@
...
+                    help="Profiler Python tracer level", type=int)
+
+def main(args):
```
skye, line 62: To test this, could you pull this out to a function that doesn't require flags, and write a unit test that starts a long-running computation and manually calls this? It might be a big heavy duty.

skye (changes requested):

jax/collect_profile.py
sharadmv: The `profiler.trace(...)` outputs an `xplane.pb` file that needs to be converted into `trace.json` for use w/ Perfetto UI. I'll add a comment explaining what we are doing here.

skye:
Did you test if this works with a Cloud TPU VM? I'm a bit worried that the profiler won't be able to initialize without libtpu, and that the process you wanna profile will already have opened libtpu. There may be workarounds though!

sharadmv:
Yes it seems to work. I profiled the Flax WMT training example on a Cloud TPU v3-8: https://ui.perfetto.dev/#!/?s=e2c034d053cdb38c66215f8f424834a678112f74e0dddea56d524097aefa2b

jax/collect_profile.py
sharadmv: Thanks!

I can't see to set `enable_hlo_proto`. The struct (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/profiler/profiler_v2.py#L46) doesn't see to have it.

jax/collect_profile.py
sharadmv: Done.

docs/profiling.md
sharadmv: Done.

jax/collect_profile.py
sharadmv: Done, though the test may not actually be run w/o the proper deps: cc @yashk2810

tests/profiler_test.py
```diff
@@ -205,5 +213,30 @@ def on_profile(port, logdir, worker_start):
...
+
+    thread_profiler = threading.Thread(
+        target=on_profile, args=())
+    thread_profiler.start()
+    jax.profiler.start_server(port)
+    y = jnp.zeros((500, 500))
+    y = jnp.dot(y, y)
+    jax.profiler.stop_server()
+    thread_profiler.join()
+    self.assertTrue(len(ret[0]) > 0)
```
skye, line 239: Bonus points: check the profile contents. You should be able to reuse the L92-102

skye (approved):

jax/collect_profile.py
skye: Hm maybe TF didn't expose it in their Python bindings. We can figure out how to add it later if it's a problem. Or maybe it's on by default already.

tests/profiler_test.py
sharadmv: I added a check for the presence of a "plugins" folder that would only be created if the profile plugin captured a trace.

sharadmv:
Thanks for the review!

tests/profiler_test.py
skye: It's also good to check that the trace actually captured something, e.g. we used to have problems where we would generate a trace but without device or Python traces. That's what the test above does. Up to you though.

sharadmv: I see, okay I used that helper function. Thanks!

PR #10968, opened 2022-06-03
Add `jax.logging` module

### LINKED ISSUE #6308: Migrate away from absl logging.
I would suggest a small change. It seems like Jax's logging facilities rely on [Abseil](https://abseil.io/docs/python/guides/logging) like tensorflow, which is an interesting google product. But, is it really necessary to use though? Python already provides a logging mechanism, would it be easy to migrate to that? It seems to be used in

1. jaxlib/tpu_client.py
2. jaxlib/xla_client.py
3. jax/experimental/host_callback.py
4. jax/interpreters/pxla.py
5. jax/interpreters/sharded_jit.py
6. jax/interpreters/xla.py
7. jax/lib/xla_bridge.py

Using abseil can confuse users who are trying to silence warnings and assume jax is using python's warning systems. For example, the famous
```
```

nicholasjng:
Fixes #6308.

This commit adds the top-level `jax.logging` module to JAX. In it, the most important routines for logging are imported, enabling their use across the codebase.

The choice which module these routines are imported from (choices are absl or Python's builtin logging) can be made through the newly added `jax_use_absl_logging` config flag.

The `jax.logging` module is intended as a drop-in replacement for `from absl import logging` imports, allowing for more flexible logging setups within JAX.

EDIT: The first approach was not dynamic enough (or, at all), but using a config value update hook that hot-reloads the logging module and reimports conditionally based on the flag value, the following works now:

```python
>>> from jax import logging
>>> from jax.config import config
>>> logging.warning("hello")
>>> WARNING:root:hello  #       <-- this is the Python builtin root logger
>>> config.update("jax_use_absl_logging", True)
>>> logging.warning("hello")
>>> WARNING:absl:hello  #       <-- this is the absl logger
```

### DIFF
--- a/jax/__init__.py
+++ b/jax/__init__.py
@@ -144,6 +144,7 @@ from jax import dtypes as dtypes
 from jax import errors as errors
 from jax import image as image
 from jax import lax as lax
+from jax import logging as logging
 from jax import nn as nn
 from jax import numpy as numpy
 from jax import ops as ops
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -18,13 +18,12 @@
 import contextlib
 import functools
 import itertools
+import logging
 import os
 import sys
 import threading
 from typing import Any, List, Callable, Hashable, NamedTuple, Iterator, Optional
 import warnings
-
-from absl import logging
 
 from jax._src import lib
 from jax._src.lib import jax_jit
@@ -522,6 +521,13 @@ def update_thread_local_jit_state(**kw):
   tls.extra_jit_context = _thread_local_state_cache.canonicalize(tmp)
 
 
+def _reload_logging_module(val):
+  # Update hook for the jax_use_absl_logging flag.
+  # Reloads JAX's logging module using importlib.
+  from importlib import reload, import_module
+  jax_logging = import_module("jax.logging")
+  jax_logging = reload(jax_logging)
+
 flags.DEFINE_integer(
     'jax_tracer_error_num_traceback_frames',
     int_env('JAX_TRACER_ERROR_NUM_TRACEBACK_FRAMES', 5),
@@ -532,6 +538,13 @@ flags.DEFINE_bool(
     'jax_pprint_use_color',
     bool_env('JAX_PPRINT_USE_COLOR', True),
     help='Enable jaxpr pretty-printing with colorful syntax highlighting.'
+)
+
+flags.DEFINE_bool(
+    'jax_use_absl_logging',
+    bool_env('JAX_USE_ABSL_LOGGING', False),
+    help='Use absl logging in JAX instead of Python builtin logging.',
+    update_hook=_reload_logging_module,
 )
 
 flags.DEFINE_bool(
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -30,12 +30,12 @@ import re
 import threading
 import warnings
 
-from absl import logging
 import numpy as np
 
 import jax
 from jax import core
 from jax import linear_util as lu
+from jax import logging
 from jax.errors import UnexpectedTracerError
 import jax.interpreters.ad as ad
 import jax.interpreters.batching as batching
--- a/jax/_src/distributed.py
+++ b/jax/_src/distributed.py
@@ -18,8 +18,9 @@ import functools
 
 from typing import Any, Optional, Union, Sequence
 
-from absl import logging
 from jax._src.clusters import ClusterEnv
+from jax import logging
+from jax._src import cloud_tpu_init
 from jax._src.config import config
 from jax._src.lib import xla_extension
 
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -26,9 +26,9 @@ import threading
 from typing import Any, Dict, List, Optional, Union
 import warnings
 
-from absl import logging
+from jax import logging
 # Disable "WARNING: Logging before flag parsing goes to stderr." message
-logging._warn_preinit_stderr = 0
+logging.set_warn_preinit_stderr(False)
 
 import jax._src.lib as lib
 from jax._src.config import flags, bool_env, int_env
@@ -127,7 +127,7 @@ def get_compile_options(
     build_options.auto_spmd_partitioning_mesh_ids = auto_spmd_partitioning_mesh_ids
   if device_assignment is not None:
     logging.vlog(
-        2,
+        logging.CPP_ERROR,
         'get_compile_options: num_replicas=%s num_partitions=%s device_assignment=%s',
         num_replicas, num_partitions, device_assignment)
     device_assignment = np.array(device_assignment)
@@ -384,7 +384,7 @@ def _init_backend(platform):
   if factory is None:
     raise RuntimeError(f"Unknown backend '{platform}'")
 
-  logging.vlog(1, "Initializing backend '%s'" % platform)
+  logging.vlog(logging.CPP_WARNING, "Initializing backend '%s'" % platform)
   backend = factory()
   # TODO(skye): consider raising more descriptive errors directly from backend
   # factories instead of returning None.
@@ -396,7 +396,7 @@ def _init_backend(platform):
                              ("process_index", backend.process_index()),
                              ("device_count", backend.device_count()),
                              ("local_devices", backend.local_devices()))
-  logging.vlog(1, "Backend '%s' initialized" % platform)
+  logging.vlog(logging.CPP_WARNING, "Backend '%s' initialized" % platform)
   return backend
 
 
--- a/jax/_src/profiler.py
+++ b/jax/_src/profiler.py
@@ -25,7 +25,7 @@ import warnings
 
 from typing import Callable, Optional
 
-from absl import logging
+from jax import logging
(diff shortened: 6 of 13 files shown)


froystig (commented):
@nicholasjng – thanks for your patience here, and for rebasing! Sorry that feedback on this PR had stalled for a while.

The change looks great along several dimensions, and the new `logging.py` is super clear to follow.

I hadn't considered using `import_module` and `reload` until reading this. Experimenting with it a bit, I wonder if aliases to attributes of `logging` that persist across the reload boundary make for unexpected behavior. For example, if I modify your example in the PR description slightly:

```python
>>> from jax.logging import warning
>>> from jax.config import config
>>> warning("hello")
>>> WARNING:root:hello  #       <-- this is the Python builtin root logger
>>> config.update("jax_use_absl_logging", True)
>>> warning("hello")
>>> WARNING:root:hello  #       <-- this is still the Python builtin root logger
``` 

In words, the reloading seems to swap out the module but not its possibly-aliased members.

It seems to me that this is behavior we'd rather avoid, although I'm not entirely certain. What do you think?

If we wanted to, could we change the implementation to make this behave analogously to your initial example (namely, respecting the dynamic change in config setting)?

nicholasjng:
Thank you for the feedback!

> It seems to me that this is behavior we'd rather avoid, although I'm not entirely certain. What do you think?

Definitely to be avoided, in my opinion. I was unaware of this, but I could see that symbols taken from an imported module are not updated upon module reload. The other option, only ever using the `jax.logging` import and never any function-level imports from it, seems like a pretty big restriction on the user to me.

> If we wanted to, could we change the implementation to make this behave analogously to your initial example (namely, respecting the dynamic change in config setting)?

Judging from browsing SO on Python hot reloading for a couple of minutes, the logging member imports in the caller module need to be reloaded as well. I do not exactly know if/how that is possible, but the machinery probably exists. My guess is it would most likely be the current solution plus a lookup of any imported `jax.logging` module members in the "right place" (where exactly is TBD), and, if found, their reimport.

For your reference: A good SO thread I found on this (because of an illustrative code example) is [this one](https://stackoverflow.com/questions/1739924/python-reload-component-y-imported-with-from-x-import-y); the example from the Python REPL there indicates the difference between whole-module imports and "dotted" (module member) imports, where the latter requires the second `from X import Y` statement for a successful reload.

So if we find a way to emulate the second import statement from the config update hook, we should be good to go.

froystig:
Also for reference, based on a quick test, [PEP 562](https://peps.python.org/pep-0562/) (`__getattr__` on modules) will not help us here either.

We should be able to handle all module-level functions, like `logging.warning`, by defining each one as a function that switches on the config setting to call the corresponding selected implementation (standard `warning` or absl `warning`). The remaining challenge is to handle the non-callable attributes. From what I can tell, all of these are logging level integers, such as `logging.DEBUG`.

Could we simply choose not to support the vlog parts of the logging interface (`vlog` and `vlog_is_on`), and instead maintain a habit of only calling `logging.info`, `logging.warning`, etc. directly?

@jakevdp @skye @hawkinsp any thoughts?

froystig:
Related discussion in #12651

nicholasjng:
Cool. I initially even planned to do it this way, but I got discouraged by the boilerplate that this added for supporting the whole logging interface.
I also opted to keep the `vlog` usage to make the change less disruptive and allow smooth adoption, but I personally think the discussed `logging.debug/info/...` APIs should be plenty - happy to change that, after a quick discussion on what the `CPP_{WARNING,ERROR}` levels should become.

As for the integers, we _could_ just use `logging`'s version of the constants, and set up a level conversion mechanism for absl APIs similar to ABSL's internal conversion of levels in `absl.log`, where we need to. But are you sure that the `if` branch there does not suffice? (I thought it might end up working like a C++ preprocessor directive)

froystig:
I discussed this a bit with @hawkinsp and @shoyer. Now that we understand the costs better, we think it's worthwhile to simplify things and just drop absl logging altogether, rather than continue trying to support both logging packages via a common interface. To do that, we'll need to migrate the jax codebase further. And while we're at it, we should also avoid using the builtin `'root'` Python logger (cf. #12526). Here's how I suspect this could go:

1. Introduce a global internal logger instance, which we can name `'jax'`.
2. Since `vlog` will no longer be available, replace current uses of `vlog` with calls to the corresponding `logging.{debug,warning,info,...}` on the `'jax'` logger.
3. Similarly, replace current uses of `vlog_is_on` with equivalent logic from Python's `logging`, e.g. using [`isEnabledFor()`](https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor) and [`getEffectiveLevel()`](https://docs.python.org/3/library/logging.html#logging.Logger.getEffectiveLevel). Note that Python logging skips string interpolation if you format with positional arguments, so some current uses of `vlog_is_on` can maybe be removed entirely, if the interpolation cost is all that they are guarding. (See [here](https://docs.python.org/3/howto/logging.html#optimization) and [here](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles).)
4. Migrate `log` calls as well, so that we only use the `'jax'` non-root logger.
5. Drop `absl` logging entirely.

We would probably no longer need a `jax.logging` module. We could consider adding an internal `jax._src.logging` for any utilities that make things easier to manage, if the need for them arises while working on the changes above.

What do you think?

Of course, let us know if this seems like more than what you'd like to take on at this point, or if you'd like to handle only a subset of such changes in this PR.

skye:
Hey sorry for not responding earlier! The key thing I like about vlog is that you can specify a vmodule to only print vlogs from a certain file. This is way better than printing all debug logging, which is often way too verbose. vlog also has the feature of introducing arbitrary debug levels, so you can control the verbosity even within a file. This is pretty useful as well, but not as critical as the per-module logging IMO.

It sounds like we should be able to do this with Python logging as well though? Maybe by specifying a different logger for each file/module? I'm in general in favor of dropping absl logging because it barely works in Python (e.g. I only know how to set a Python vmodule when running inside Google).

froystig:
I was thinking that we can look into introducing more fine-grained loggers after we migrate (e.g. one per file or module, etc.). That way we keep things simple for a transition PR and then experiment with further changes afterwards (also e.g. trying custom formatters).

We might have to write a bit of code to reinstate a few absl-like behaviors that we like, but it seems that would still be the preferred eventual situation.

hawkinsp:
@skye The idiom that a lot of folks seem to use is:

```
import logging
logger = logging(__name__)

logger.debug("hello")
```

This plays a role similar to `vlog` and `vmodule`, because in another file you can turn up the verbosity of that specific module's logs like this:
```
import logging
l = logging.getLogger("jax.core")
l.setLevel(logging.DEBUG)
```
i.e., you can enable debug logs on a per-module basis.

nicholasjng:
Sounds good! I'll work on implementing your list when I get time, I expect it could work out this week.

As for the vlog/vmodule substitution, I think I will go with @hawkinsp's suggestion, and then make a global `jax` logger available somewhere.

By the way, would you support dropping `absl-py` as a package install requirement? It can probably stay for testing, but I think there was some interest for eliminating that dependency (ping @josephrocca).

froystig:
> As for the vlog/vmodule substitution, I think I will go with @hawkinsp's suggestion, and then make a global `jax` logger available somewhere.

Sounds good. Do we need both? What would the global `'jax'` logger be used for at that point?

> By the way, would you support dropping `absl-py` as a package install requirement? It can probably stay for testing, but I think there was some interest for eliminating that dependency (ping @josephrocca).

Yes, good idea. We still support [absl flags](https://github.com/google/jax/blob/05faf0f40d376afccd9c3e957fe811b4c839aed5/jax/_src/config.py#L137) optionally, but we can make this a soft dependency: works if absl is installed, doesn't otherwise.

Our tests use it, but it doesn't need to be an install requirement for jax. We may need to add it to [build/test-requirements.txt](https://github.com/google/jax/blob/main/build/test-requirements.txt).

nicholasjng:
Sweet. Then I am good to go, except for one proposal. What do you think about adding a small `get_logger` facility in, say, the top-level `jax` namespace? Makes for easy-to-remember code (think `logger = jax.get_logger`, similar to @hawkinsp's example up top), and, maybe even nicer, would allow for some default values controllable by environment variables. 

I have a quick example [here](https://github.com/nicholasjng/pybm/blob/master/pybm/logging.py) from a personal project of mine; the `config.get_value(foo)` calls there would be `FLAGS.foo` in the JAX setting.

And last but not least, in that case, the proposed "jax" logger would be equal to the module-level logger for the `jax` top-level namespace, so the module-level solution is a perfect superset of the global one.

froystig:
> What do you think about adding a small `get_logger` facility in, say, the top-level `jax` namespace? Makes for easy-to-remember code (think `logger = jax.get_logger(...)`, similar to @hawkinsp's example up top), and, maybe even nicer, would allow for some default values controllable by environment variables.

Do you mean to add it to the public `jax` namespace? Useful utilities are a fine idea in general. If I understood correctly, `get_logger` could be internal and go under `jax._src`, either in a new module like `jax._src.logging` or in an existing one like `jax._src.util`. (?)

nicholasjng:
Closing, as #12769 got accepted. Thanks for your input!

PR #11006, opened 2022-06-07
Update testReducerNoDtype and others to not run duplicate tests.

pschuh:
ndarray.item() should return a python type (eg: float, complex, int)
and thus many PYTHON_SCALAR_SHAPE types are duplicated.

### DIFF
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -309,6 +309,12 @@ NUMPY_SCALAR_SHAPE = _NumpyScalar()
 PYTHON_SCALAR_SHAPE = _PythonScalar()
 
 
+# Some shape combinations don't make sense.
+def is_valid_shape(shape, dtype):
+  if shape == PYTHON_SCALAR_SHAPE:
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks good.

I think scalar shapes are used much more widely than just in these tests, though. What prevents the issues fixed here from arising in other tests? Could we somehow add this guardrail without having to remember to add the `is_valid_shape` clause every time we use `all_shapes` or `scalar_shapes` in a test?

jakevdp:
e.g. maybe we could change it so that our shape enumerations look like `for shape in all_shapes(dtype)`. What do you think?

pschuh:
The reason why this only affects the bfloat16 is that the test goes like so:
v = input.item()
v = np.array(v)
...
return v

There is a special case for bf16 (just these nan tests) that goes like so:
v = input.item()
v = np.array(v)
v = np.array(v, dtype=float32)
...
v = np.array(v, dtype=bfloat16)
return v

the item() change makes the type of blah.item() become float instead of bfloat16.

PR #11040, opened 2022-06-09
DOC: add FAQ section on zero gradients for rank-based operations

jakevdp:
Added this because similar questions have come up a few times in recent months (most recently in https://github.com/google/jax/discussions/11033)

Preview the new docs here: https://jax--11040.org.readthedocs.build/en/11040/faq.html#gradients-are-zero-for-rank-based-functions

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -693,6 +693,86 @@ Additional reading:
   * `How to avoid NaN gradients when using where <https://github.com/tensorflow/probability/blob/master/discussion/where-nan.pdf>`_.
 
 
+Why are gradients zero for functions based on sort order?
+---------------------------------------------------------
+
+If you define a function that processes the input using operations that depend on
+the relative ordering of inputs (e.g. ``max``, ``greater``, ``argsort``, etc.) then
+you may be surprised to find that the gradient is everywhere zero.
+Here is an example, where we define `f(x)` to be a step function that returns
+`0` when `x` is negative, and `1` when `x` is positive:
+
+  import jax
+  import numpy as np
+  import jax.numpy as jnp
+
+  def f(x):
+    return (x > 0).astype(float)
+
+  df = jax.vmap(jax.grad(f))
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -693,6 +693,85 @@ Additional reading:
   * `How to avoid NaN gradients when using where <https://github.com/tensorflow/probability/blob/master/discussion/where-nan.pdf>`_.
 
 
+Gradients are zero for rank-based functions
```
skye, line 696: I have to admit it bothers me that our FAQ doesn't contain actual questions :P But maybe that's for a different PR to resolve.

docs/faq.rst
```diff
@@ -693,6 +693,85 @@ Additional reading:
...
+-------------------------------------------
+
+If you define a function that processes the input using rank-based operators
```
skye, line 699: I don't know what "rank-based" means :(

skye (approved):
This is great!! Very good explanation, even when I don't know what a rank-based function is

docs/faq.rst
jakevdp: I tried to give examples later in the sentence – anything I could do to make that more clear?

jakevdp: To elaborate, what I had in mind is functions like `greater`, `max`, `argsort`, etc. whose output depends on comparing the order/ranking of inputs.

skye: Ooh I was thinking rank like the number of dimensions. I think "sort-order-based functions" would be clearer to me, and/or put "whose output depends on comparing the sort-order/ranking of inputs" around here.

docs/faq.rst
jakevdp: Fixed, at least for this case!

docs/faq.rst
jakevdp: Thanks - done. PTAL!

skye (approved):

PR #11043, opened 2022-06-09
Add bitwise XOR reducer to `lax.reduce`

### LINKED ISSUE #11005: Bitwise XOR missing as a `lax.reduce` operation
The `lax.reduce` function includes bitwise AND and OR as possible reduction operations, but does not include XOR. Likewise, there is no specialized XOR-reduction function in JAX.

nicholasjng:
Fixes #11005.

This commit adds handling for the `lax.bitwise_xor` operation to `lax.reduce`. It also includes a new standard reduce primitive, modeled after the existing `and`/ `or` reducer primitives.

TODO: Add testing. (The example by @jakevdp given in the issue works now, though:)

```python
>>> from jax import lax
>>> import jax.numpy as jnp
>>> lax.reduce(jnp.arange(10), 0, lax.bitwise_xor, (0,))
>>> DeviceArray(1, dtype=int32)
```

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -981,6 +981,7 @@ tf_not_yet_impl = [
     "igamma_grad_a",
     "random_gamma_grad",
     "reduce_precision",
+    "reduce_xor",
     "schur",
     "closed_call",
     "unreachable",
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -1749,6 +1749,44 @@ class LaxTest(jtu.JaxTestCase):
     self._CompileAndCheck(fun, args_maker)
 
   @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": "_op={}_shape={}_reducedims={}_initval={}_prim={}"
+       .format(op.__name__, shape, dims, init_val, prim),
+       "op": op, "init_val": init_val, "shape": shape, "dims": dims, "prim": prim}
+      for init_val, op , prim in [
+        (True, lax.bitwise_and, jax.lax.reduce_and_p),
+        (False, lax.bitwise_or, jax.lax.reduce_or_p),
+        (False, lax.bitwise_xor, jax.lax.reduce_xor_p),
+      ]
+      for shape, dims in [
+          [(3, 4, 5), (0,)], [(3, 4, 5), (1, 2)],
+          [(3, 4, 5), (0, 2)], [(3, 4, 5), (0, 1, 2)]
+      ]))
+  def testReduceBoolean(self, op, init_val, shape, dims, prim):
+    def reference_fun(operand, init_value):
+      np_op = getattr(np, op.__name__)
+      return np_op.reduce(operand, axis=dims, initial=init_val)
+
+    dtype = np.bool_
+    rng = jtu.rand_bool(self.rng())
+    init_val = np.asarray(init_val, dtype=dtype)
+    fun = lambda operand, init_val: lax.reduce(operand, init_val, op, dims)
+    args_maker = lambda: [rng(shape, dtype), init_val]
+    self._CompileAndCheck(fun, args_maker)
+    self._CheckAgainstNumpy(reference_fun, fun, args_maker)
+
+    # recheck with a static init_val
+    fun = lambda operand: lax.reduce(operand, init_val, op, dims)
+    reference_fun = partial(reference_fun, init_value=init_val)
+    args_maker = lambda: [rng(shape, dtype)]
+    self._CompileAndCheck(fun, args_maker)
+    self._CheckAgainstNumpy(reference_fun, fun, args_maker)
+
+    # check that the correct monoid reducer primitive is used inside the
+    # jaxpr. This requires the init_val (monoid identity element) to be static
+    jaxpr = jax.make_jaxpr(fun)(rng(shape, dtype))
+    self.assertEqual(jaxpr.eqns[0].primitive, prim)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_op={}.{}_arr_weak_type={}_init_weak_type={}"
        .format(op_namespace.__name__, op, arr_weak_type, init_weak_type),
        "op": op, "op_namespace": op_namespace, "arr_weak_type": arr_weak_type, "init_weak_type": init_weak_type}
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -983,6 +983,8 @@ def _get_monoid_reducer(monoid_op: Callable,
       return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_or
     elif monoid_op is bitwise_and and dtype == np.bool_:
       return np.equal(aval.val, _get_min_identity(dtype)) and _reduce_and
+    elif monoid_op is bitwise_xor and dtype == np.bool_:
+      return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_xor
     elif monoid_op is max:
       return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_max
     elif monoid_op is min:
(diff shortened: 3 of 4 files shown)


jakevdp:
Looks great so far! I'm poking around, and it doesn't look like we have any explicit test coverage of the boolean reductions for `lax.reduce`. If you're up for it, you can add those as part of this PR. If not, please let me know and I can maybe do that separately, and you can use the test as a template for the new ones here.

nicholasjng:
Yes, that stopped me a little bit. I wanted to confirm if there was any existing coverage before adding my own thing, but thank you for confirming. I can add it no problem, what do you think about adding it as a new actual test (i.e. a new function)? The reason is that the decorator over `testReduce` is already getting quite crowded, and separating the boolean ops seems like a nice semantic structure in my opinion.

Could you briefly sketch your approach on this? Otherwise I can also build off of the existing `testReduce` facility.

jakevdp:
Also, looking more closely, these special reductions are only defined for boolean types, so the example above doesn't even hit this codepath I believe. We should be able to extend these primitives to support both boolean and integer types, but that's probably a contribution for another time.

nicholasjng:
I added a separate test for the boolean reduce ops. Let me know if that works :)

tests/lax_test.py
```diff
@@ -1748,6 +1748,36 @@ def testReduce(self, op, init_val, shape, dtype, dims):
...
+    init_val = np.asarray(init_val, dtype=dtype)
+    fun = lambda operand, init_val: lax.reduce(operand, init_val, op, dims)
+    args_maker = lambda: [rng(shape, dtype), init_val]
+    self._CompileAndCheck(fun, args_maker)
+
+    # we separately test the version that uses a concrete init_val because it
+    # can hit different code paths
+    fun = lambda operand: lax.reduce(operand, init_val, op, dims)
+    args_maker = lambda: [rng(shape, dtype)]
+    self._CompileAndCheck(fun, args_maker)
```
jakevdp, line 1779: A suggestion here: you could use `jax.make_jaxpr` and assert that the operation is implemented via the expected primitive, e.g something like
```python
prim = lax.reduce_or_p
jaxpr = jax.make_jaxpr(fun)(rng(shape, dtype))
self.assertEqual(jaxpr.eqns[0].primitive, prim)
```
where you could add the expected reduction primitive to the test parameterization.

nicholasjng: Thanks for the suggestion. I did that, but the primitive is "reduce" every time. However, printing the jaxpr, it looks like the correct things are being done:

```
>>> jaxpr.eqns[0]
>>> a:bool[] = reduce[
  computation=<function bitwise_xor at 0x134349790>
  consts=()
  dimensions=(0,)
  jaxpr={ lambda ; b:bool[] c:bool[]. let d:bool[] = xor b c in (d,) }
] e f
```
The value is just kept in another entry than `primitive`. Going to look and see which one it is (debugger tells me that it might be `computation`).

nicholasjng: It is `jaxpr.params["computation"].__name__":

```
>>> jaxpr.params["computation"].__name__
>>> 'bitwise_xor'
```

jakevdp: If the bound primitive is reduce and not reduce_xor, then this test is not actually exercising the new code. We should figure out why.

nicholasjng: Are you sure? Because I am getting this also:

```python
>>> shape, dtype = (3, 4, 5), np.int32
>>> fun = lambda operand, init_val: lax.reduce(operand, init_val, lax.add, (0,))
>>> jax.make_jaxpr(fun)(rng(shape, dtype), 0)
{ lambda ; a:i32[3,4,5] b:i32[]. let
    c:i32[4,5] = reduce[
      computation=<function add at 0x127dfb940>
      consts=()
      dimensions=(0,)
      jaxpr={ lambda ; d:i32[] e:i32[]. let f:i32[] = add d e in (f,) }
    ] a b
  in (c,) }
>>> jax.make_jaxpr(fun)(a(shape, dtype), 0).eqns[0].primitive
>>> reduce
```

Maybe I am not doing something completely correctly here, though.

jakevdp: You need the init  value to be static, so it will only work for this test and not the other:
```python
import jax
import jax.numpy as jnp

def f(x):
  return jax.lax.reduce(x, 0, jax.lax.add, (0,))

jaxpr = jax.make_jaxpr(f)(jnp.arange(4))
print(jaxpr)
# { lambda ; a:i32[4]. let b:i32[] = reduce_sum[axes=(0,)] a in (b,) }

print(jaxpr.eqns[0].primitive is jax.lax.reduce_sum_p)
# True
```

jakevdp: `lax.reduce` will only produce the correct results if the `init_val` and `op` form a monoid. So we can't iterate over arbitrary init values; you have to use the correct `init_val` for the given `op`.

nicholasjng: Thanks! But that means that we could just delete these, since they violate the assumptions `lax.reduce` anyway. Is that correct?

nicholasjng:
I fixed the test failures (due to my misunderstanding of the role of `init_val`) and added `xor_reduce_p` to `jax.lax` exports.

tests/lax_test.py
jakevdp: When you say "delete these", what are you referring to?

nicholasjng: I meant "do not run tests with the wrong `init_val` for a given op" (so `False` for `reduce_and`, `True` for `reduce_{or,xor}`). The rationale is that since the monoidal assumption to `lax.reduce` was being violated by the tests inputs (specifically the wrong `init_val` for these ops), I was actually testing the wrong thing.

Do you want non-monoidal codepaths to be tested as well? Then I could add some tests for that, too (with the primitive as `reduce_p`).

jakevdp: No, i don’t think we need tests for invalid inputs

tests/lax_test.py
```diff
@@ -1748,6 +1748,37 @@ def testReduce(self, op, init_val, shape, dtype, dims):
...
+          [(3, 4, 5), (0,)], [(3, 4, 5), (1, 2)],
+          [(3, 4, 5), (0, 2)], [(3, 4, 5), (0, 1, 2)]
+      ]))
+  def testReduceBoolean(self, op, init_val, shape, dims, prim):
+    dtype = np.bool_
+    rng = jtu.rand_bool(self.rng())
+    init_val = np.asarray(init_val, dtype=dtype)
+    fun = lambda operand, init_val: lax.reduce(operand, init_val, op, dims)
+    args_maker = lambda: [rng(shape, dtype), init_val]
+    self._CompileAndCheck(fun, args_maker)
```
jakevdp, line 1770: Let's also do a `self._CheckAgainstNumpy` here and below for a suitable numpy version – we want to make sure the functions are returning the correct result!

nicholasjng: Done! I modeled it after the ReduceWindow test below.

tests/lax_test.py
```diff
@@ -1748,6 +1748,45 @@ def testReduce(self, op, init_val, shape, dtype, dims):
...
+      ]
+      for shape, dims in [
+          [(3, 4, 5), (0,)], [(3, 4, 5), (1, 2)],
+          [(3, 4, 5), (0, 2)], [(3, 4, 5), (0, 1, 2)]
+      ]))
+  def testReduceBoolean(self, op, init_val, shape, dims, prim):
+    def reference_fun(operand, init_value):
+      return lax_reference.reduce(
+        operand, init_value, computation=op, dimensions=dims,
+      )
```
jakevdp, line 1768: It would be better if the comparison were a source of truth external to the JAX library. What about something like `np.bitwise_and.reduce()`?

nicholasjng: Good call, fixed.

jakevdp (approved):
Looks great, thanks!

nicholasjng:
Thank you for the speedy review! I will try to get something done extending the binary ops to integer types as well

hawkinsp:
There's one more thing you'd need to fix to get this merged, which is add the new primitive here:
https://github.com/google/jax/blob/7098088f4eb15cf750398889e4341dbc15cda1b3/jax/experimental/jax2tf/jax2tf.py#L979

(We don't run jax2tf tests as part of our github presubmits as a measure to cutdown on presubmit times, but we do need them to pass before merging.)

nicholasjng:
Thanks, fixed.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -981,6 +981,7 @@ def _unexpected_primitive(p: core.Primitive, *args, **kwargs):
     "igamma_grad_a",
     "random_gamma_grad",
     "reduce_precision",
+    "reduce_xor_p",
```
jakevdp, line 984: The name doesn't include `_p`:
```suggestion
    "reduce_xor",
```

nicholasjng: woops. that was my bad. Fixed.

PR #11046, opened 2022-06-09
[x64] preserve weak types in promote_dtypes_inexact

jakevdp:
Why? Currently `jnp.sqrt(2.0)` returns a weakly-typed value, but `jnp.sqrt(2)` does not. This seems strange. ~I'm trying this change to see how many things it breaks.~ Edit: TGP is clean, I think this is good to go in!


jakevdp:
TGP passed

hawkinsp (approved):

PR #11052, opened 2022-06-10
[x64] make lax_scipy_test.py compatible with strict dtype promotion

jakevdp:
Part of https://github.com/google/jax/pull/10865 and https://github.com/google/jax/pull/10840.

### DIFF
--- a/jax/_src/lax/eigh.py
+++ b/jax/_src/lax/eigh.py
@@ -132,7 +132,7 @@ def _projector_subspace(P, H, n, rank, maxiter=2):
   X = _mask(X, (n, rank))
 
   H_norm = jnp_linalg.norm(H)
-  thresh = 10 * jnp.finfo(X.dtype).eps * H_norm
+  thresh = 10.0 * float(jnp.finfo(X.dtype).eps) * H_norm
 
   # First iteration skips the matmul.
   def body_f_after_matmul(X):
@@ -190,7 +190,7 @@ def split_spectrum(H, n, split_point, V0=None):
     rank: The dynamic size of the m subblock.
   """
   N, _ = H.shape
-  H_shift = H - split_point * jnp.eye(N, dtype=H.dtype)
+  H_shift = H - (split_point * jnp.eye(N, dtype=split_point.dtype)).astype(H.dtype)
   U, _, _, _ = qdwh.qdwh(H_shift, is_hermitian=True, dynamic_shape=(n, n))
   P = -0.5 * (U - _mask(jnp.eye(N, dtype=H.dtype), (n, n)))
   rank = jnp.round(jnp.trace(jnp.real(P))).astype(jnp.int32)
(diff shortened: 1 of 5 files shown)


jax/_src/lax/eigh.py
```diff
@@ -327,6 +327,7 @@ def base_case(B, offset, b, agenda, blocks, eigenvectors):
     # only do the right thing if termination_size == 1.
     H = _mask(H, (b, b), jnp.eye(B, dtype=H.dtype))
     eig_vecs, eig_vals = lax.linalg.eigh(H, sort_eigenvalues=False)
+    eig_vals = eig_vals.astype(eig_vecs.dtype)
```
hawkinsp, line 330: I guess the point of this is that we are embedding the eigenvalues (which are always real) in a complex array. One slight nit is that I might move this type cast after the masking logic.

jax/_src/lax/qdwh.py
```diff
@@ -123,7 +123,7 @@ def _qdwh(x, m, n, is_hermitian, max_iterations, eps):
   if eps is None:
     eps = float(jnp.finfo(x.dtype).eps)
   alpha = (jnp.sqrt(jnp.linalg.norm(x, ord=1)) *
-           jnp.sqrt(jnp.linalg.norm(x, ord=jnp.inf)))
+           jnp.sqrt(jnp.linalg.norm(x, ord=jnp.inf))).astype(x.dtype)
   l = eps
 
   u = x / alpha
```
hawkinsp, line 129: I'm just reflecting on this. Do we really want to force a cast when doing a `complex / real` division where the `real` type is the same on one of the complex element types?

hawkinsp (commented):
I'm not sure how I feel about requiring real -> complex casts!

jax/_src/lax/qdwh.py
jakevdp: That's a good question – I debated that in https://github.com/google/jax/pull/10824, but landed on the fact that when people say they want `strict` dtype promotion, we should probably avoid any implicit promotion rather than trying to guess which promotions are OK.

jakevdp: (I think I even posed this exact question in our x64 chat at some point)

jakevdp: I'm going to leave as-is for now; we can always make this less restrictive in the future if we wish.

jax/_src/lax/eigh.py
```diff
@@ -342,6 +343,7 @@ def recursive_case(B, offset, b, agenda, blocks, eigenvectors):
     V = _slice(eigenvectors, (0, offset), (n, b), (N, B))
 
     split_point = jnp.nanmedian(_mask(jnp.diag(jnp.real(H)), (b,), jnp.nan))  # TODO: Improve this?
+    split_point = split_point.astype(H.dtype)
```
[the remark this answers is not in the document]
jakevdp: Done, thanks!

jax/_src/lax/eigh.py
jakevdp: Done

hawkinsp (approved):

PR #11054, opened 2022-06-10
Extend Common-Gotchas docs: speed of jnp.array

### LINKED ISSUE #10662: `jnp.array` is much slower than `np.array` when called on a nested list
Consider the following example (I am using JAX version 0.3.10):
```python
>>> import jax.numpy as jnp
>>> import numpy as np
>>> x = [list(range(1000)) for i in range(1000)]    # A simple nested list
>>> %timeit np.array(x)
79.4 ms ± 6.85 ms per loop (mean ± std. dev. of 7 runs, 10 loops each
>>> %timeit jnp.array(x)
3.22 s ± 174 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

NumPy is orders of magnitude faster than JAX here. A similar problem for non-nested lists (#2919) was addressed in PR #3350.

onnoeberhard:
Closes #10662 
I extended "JAX - The Sharp Bits" a bit by a few cells to show where `jnp.array(...)` and `jnp.array(np.array(...))` differ and when it is appropriate to use the latter.

### DIFF
--- a/docs/notebooks/Common_Gotchas_in_JAX.md
+++ b/docs/notebooks/Common_Gotchas_in_JAX.md
@@ -7,7 +7,7 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.13.8
 kernelspec:
-  display_name: Python 3
+  display_name: 'Python 3.8.12 (''.venv'': venv)'
   language: python
   name: python3
 ---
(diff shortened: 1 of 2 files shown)


onnoeberhard:
Currently trying to squash these into one commit

docs/notebooks/Common_Gotchas_in_JAX.md
```diff
@@ -421,6 +421,32 @@ outputId: e31b43b3-05f7-4300-fdd2-40e3896f6f8f
 jnp.sum(jnp.array(x))
 ```
 
+Of course, writing the conversion explicitly does not make it any faster. Consider the following example of converting a nested list of integers to a JAX array:
```
jakevdp, line 424: I think this could be made much more concise, to be honest. You could say something along the lines of

> keep in mind that although `jnp.array()` accepts lists of values for convenience, passing large lists to `jnp.array()` can incur the same performance penalty that JAX prevents by disallowing list inputs to `jnp.sum` and other functions. If you are converting a large list of scalars to a JAX array outside the context of `jit` or other transforms, it can be far more performant to instead use `numpy` to convert the list to an array; e.g. `np.array(x)` rather than `jnp.array(x)`."

Less is more for these kinds of docs, I think.

jakevdp (changes requested):

docs/notebooks/Common_Gotchas_in_JAX.md
onnoeberhard: Sure, I can condense the writing. You're right that this page is already quite long and it shouldn't be filled with too much detail.
What do you think about the timeit-example? I think it's instructive to directly see just how much faster numpy is than JAX in this context.
And the last part (about the casting) isn't really a part of the performance issue, though it is related and maybe highlights a different "sharp edge". Should I cut this part?

jaymody:
Any update on getting this added to the sharp bits doc? This was a gotcha that had me confused why my code was running slow.

PR #11061, opened 2022-06-10
snapshot the threefry PRNG implementation

froystig:
Upcoming changes to the threefry RNG will affect the bits generated by a given key. We might consider this a breaking change, so snapshot the current RNG implementation and add a flag for using it.

Along the way, generalize the reference test for generated bits.


hawkinsp:
Closing as stale.

Issue #11068, opened 2022-06-11
building from source fails due to mlir BUILD referring to MlirHloModule.cc instead of MlirHloModule.cpp


randyau:
building jax from source with --enable_cuda flag, 
freshly cloned from the repo
on an WSL2 ubuntu install

gets this error in building:
ERROR: /mnt/e/linux/ubuntu/jax/jaxlib/mlir/_mlir_libs/BUILD.bazel:90:13: no such target '@org_tensorflow//tensorflow/compiler/mlir/hlo:python/MlirHloModule.cc': target 'python/MlirHloModule.cc' not declared in package 'tensorflow/compiler/mlir/hlo' (did you mean 'python/MlirHloModule.cpp'?) defined by /home/agi/.cache/bazel/_bazel_agi/e71f5b8290a4fc88939262d32d7b9edb/external/org_tensorflow/tensorflow/compiler/mlir/hlo/BUILD and referenced by '//jaxlib/mlir/_mlir_libs:_mlirHlo.so'

Issue can be fixed by going to 
/mnt/e/linux/ubuntu/jax/jaxlib/mlir/_mlir_libs/BUILD.bazel:90
and changing the file extension from .cc to .cpp

Looks like the line was recently patched from .cpp to .cc to "Fix broken builds"
not sure why there's conflicting build experiences

ludgerpaehler:
See the linked pull-request I just opened - the real change happened inside of the Tensorflow repository, where the MlirHLO compiler lies, and where cpp files were renamed to cc files. When bumping the version of the Tensorflow repo in the `WORKSPACE` file to a more current one, this issue is fixed.

ludgerpaehler:
Try again now from main - it should work now :)

hawkinsp:
This is now fixed at head due to https://github.com/google/jax/pull/11073 (Thanks for the PR @ludgerpaehler !)

PR #11073, opened 2022-06-12
Bump version of Tensorflow-Repo to fix MlirHLO build error in source builds

ludgerpaehler:
Bump version of Tensorflow repository to fix build errors in the building of the MLIR compiler where cpp files have been renamed to cc files.

Specifically, when building from source for CPU i.e.

```bash
python build/build.py --noenable_mkl_dnn --noenable_cuda
```

Bazel cannot fetch `@org_tensorflow//tensorflow/compiler/mlir/hlo:python/MlirHloModule.cc`. In order to fix this the version of the Tensorflow repository the components are pulled from needs to be bumped.

Fixes issue #11068 in the process.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -7,10 +7,10 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 #    and update the sha256 with the result.
 http_archive(
     name = "org_tensorflow",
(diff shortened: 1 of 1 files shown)


mattwalsh (approved):
LGTM

PR #11077, opened 2022-06-13
Add max step size argument to odeint

### LINKED ISSUE #10922: Add odeint max step size optional argument
It'd be nice if `odeint` had an optional argument analogous to scipy `solve_ivp`'s `max_step` argument, which controls the maximum step size. I'll copy a code example below, but I'm running into an issue with `odeint` in which it steps over a feature in the ODE `rhs` after a long-ish period for which `rhs=0.`

One minor issue is that `max_step` unfortunately has a naming collision with `odeint`'s optional arg `mxstep`, which is an integer controlling the total number of iterations `odeint` can take before exiting, so this name may not be usable without changing `mxstep` which may not be desirable.

Aside from choosing the argument name, I think implementing this is nearly trivial: just need to add the argument, and I believe change [line 199](https://github.com/google/jax/blob/bab8520d0c6c40aa3799b897a7f2629ba746c2ad/jax/experimental/ode.py#L199) to take the minimum of the max step size and the computed optimal step. The example I post below could be rolled into a unittest validating this behaviour (perhaps with comparison directly with scipy).

Thoughts? Given how simple it is I'm happy to implement this myself, but would definitely defer to the core JAX developers for what argument name to use (or for any insight into why this might be an issue with other parts of `odeint`/JAX control flow).

## Motivating example

```
import jax
jax.config.update('jax_platform_name', 'cpu')

import jax.numpy as jnp

dt = jnp.array(1.)
k = 5
def deriv(t):
    t = jnp.array(t)
    return jnp.piecewise(
        t,
        [t <= 2 * dt, (t >= k * dt) & (t <= (k + 2) * dt)],
        [lambda s: jnp.array(1.), lambda s: jnp.array(-1.), lambda s: jnp.array(0.)]
    )

def rhs(y, t):
    return deriv(t)

odeint(
    func=rhs, 
    y0=jnp.array(0.), 
    t=jnp.array([0., k*dt, 2*k*dt]),
    atol=1e-8, rtol=1e-8
)
```
Using `odeint`, the above code is essentially integrating a function which is `1.` over the interval `[0., 2*dt]`, `-1.` over the interval `[k*dt, (k+1)*dt]`, and zero elsewhere. If integrated over an interval containing both non-zero intervals, the correct result should be `0.`. However , the above code outputs:
```
DeviceArray([0.        , 1.99999565, 1.99999565], dtype=float64)
```
and changing the tolerance doesn't change this (it just yields better or worse approximations of `[0., 2., 2.]`)

The analogous code in scipy using `solve_ivp` with `method='RK45'` yields the same results, but with `solve_ivp` this can be corrected by setting `max_step=dt`. Shortening the intermediate `0.` interval by either reducing `dt` or `k` both solve the problem as well, so I'm pretty confident it's an issue of step size overshoot.

### Slightly more context

The above is a minimal example, but the similarities to the situation in which I'm using `odeint` and the above are: 
- I don't know ahead of time if a period of `0.` will be present or for how long. 
- I do however have something analogous to `dt` - basically a natural time scale over which to restrict the step so that features aren't missed, but that also isn't so short that it will totally mess with the efficiency of the solver.

DanPuzzuoli:
Closes #10922

Adds argument `hmax` to `odeint` allowing the user to control the maximum step size that the solver takes. Implemented by calling `jnp.clip(x, a_min=0., a_max=hmax)` at each point where a step size `x` is computed. Argument defaults to `jnp.inf`. This argument is set as `static` and `nondiff` argnums in appropriate places.

Also adds a test case that fails if `hmax=jnp.inf` (i.e. under existing behaviour), but passes if `hmax=1.`. The test case is explained in #10922 : it is set up so that the `rhs` is `0.` for long enough that the step size grows so large that it "steps over" a period of time in which the `rhs != 0`. Setting `hmax = 1.` (chosen based on the details of the ODE) prevents the solver step size from growing arbitrarily large, allowing the user to avoid this problematic behaviour.

The argument name choice `hmax` is taken from the corresponding argument of [odeint in scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.odeint.html). If the reviewer thinks it makes sense, we could also implement `h0` (initial step size) and `hmin` (forced minimum step size) in this PR, which would be very similar changes.

### DIFF
--- a/jax/experimental/ode.py
+++ b/jax/experimental/ode.py
@@ -142,7 +142,7 @@ def optimal_step_size(last_step, mean_error_ratio, safety=0.9, ifactor=10.0,
                       jnp.maximum(mean_error_ratio**(-1.0 / order) * safety, dfactor))
   return jnp.where(mean_error_ratio == 0, last_step * ifactor, last_step * factor)
 
-def odeint(func, y0, t, *args, rtol=1.4e-8, atol=1.4e-8, mxstep=jnp.inf, dtmax=jnp.inf):
+def odeint(func, y0, t, *args, rtol=1.4e-8, atol=1.4e-8, mxstep=jnp.inf, hmax=jnp.inf):
   """Adaptive stepsize (Dormand-Prince) Runge-Kutta odeint implementation.
 
   Args:
@@ -157,7 +157,7 @@ def odeint(func, y0, t, *args, rtol=1.4e-8, atol=1.4e-8, mxstep=jnp.inf, dtmax=j
     rtol: float, relative local error tolerance for solver (optional).
     atol: float, absolute local error tolerance for solver (optional).
     mxstep: int, maximum number of steps to take for each timepoint (optional).
-    dtmax: float, upper bound on solver step size (optional).
+    hmax: float, maximum step size allowed (optional).
 
   Returns:
     Values of the solution `y` (i.e. integrated system values) at each time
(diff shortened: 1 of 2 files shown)


DanPuzzuoli:
@mattjj sorry to bug - I'm sure you all are busy, but just wanted to check in on someone's thoughts with this PR. It's pretty minor, but whether or not `odeint` has the ability to bound step size will fully determine it's usable my applications. Thanks!

mattjj (approved):
Looks great! Thanks for the contribution. Sorry for the delay.

Issue #10982, opened 2022-06-04
Tracker: decomposing scan


sharadmv:
With jaxprs now supporting effect types, we can express side-effects like the State monad, where references can be read from and written to (i.e. mutation). We can use state to implement a simpler `scan` control flow primitive via a `for` primitive that supports reads/writes.

This issue will track the implementation progress:
- [ ] `get/swap/addupdate` primitives
   - [ ] `impl` rules
   - [ ] `abstract_eval` rules
   - [ ] `jvp` rules
   - [ ] `transpose` rules
   - [ ] `vmap` rules
- [ ] `for` primitive
  - [ ] `impl` rule
  - [ ] `abstract_eval` rule
  - [ ] MLIR lowering
  - [ ] `jvp` rule
  - [ ] `partial_eval`
    - [ ] basic implementation
    - [ ] optimizations
      - [ ] loop invariant
      - [ ] residual passthrough
      - [ ] rematerializing loop-dependent values
  - [ ] `transpose`
  - [ ] `vmap` rule
  - [ ] Miscellaneous
    - [ ] Handling closed-over refs
    - [ ] Nested for loops
  - [ ] Reimplement `scan` in terms of `for`

The "raw" version of `for` can be found [here](https://github.com/mattjj/jax/blob/decompose-scan/scan2.py). Next steps involve porting that code to JAX core and adding tests.

AriMKatz:
Hi, for a non PL/ Compiler person, can you say a bit about what this means? 

Are  there any using facing implications such as being able to express impure loops at the cost of sequential execution or being able to  write loops that can be compiled to parallel functional primitives, based on their effect type? (Kinda like what Dex does).

sharadmv:
> Hi, for a non PL/ Compiler person, can you say a bit about what this means?

We're exploring implementing a new control-flow primitive (`for`) that generalizes scan and is more flexible/expressive. We're also exploring a "state" side-effect in JAX and its ramifications. Right now, we're containing the side-effect to just in the body of this new `for` primitive. For some examples of its usage, you can see the [control flow test](https://github.com/google/jax/blob/5777c1eac26fe96e4039c343de3b2b8f04ea7339/tests/lax_control_flow_test.py#L2515).

> Are there any using facing implications such as being able to express impure loops?

Yes, potentially. The `for` implementation is still in an exploratory stage but it is a generalization of `scan` because we can express more than just `scan`-like patterns (see the `cumsum` example in the tests). 

[...]

PR #11083, opened 2022-06-13
Add stateful for loop primitives

sharadmv:
Adds a `get/swap/addupdate` primitive, along with impl, abstract_eval and jvp rules.

Tracking issue:  #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -1,0 +1,275 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Module for the `for_loop` primitive."""
+from functools import partial
+
+from typing import Any, List, Tuple
+
+from jax import core
+from jax.interpreters import ad
+from jax._src import ad_util
+from jax._src import pretty_printer as pp
+
+
+## Helpful type aliases
+Ref = Any
+Array = Any
+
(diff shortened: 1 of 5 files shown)


sharadmv:
@mattjj thoughts on how to test the get/set transpose rules?

jax/_src/lax/control_flow/for_loop.py
```diff
@@ -0,0 +1,262 @@
...
+from jax._src import ad_util
+from jax._src import pretty_printer as pp
+
+
+## Helpful type aliases
+Ref = Array = Any
+
+## State effect
+
+State = object()
```
mattjj, line 30:
```suggestion
class StateEffect: pass
State = StateEffect
```

Leads to nicer pdb printing if/when necessary!

jax/_src/lax/control_flow/for_loop.py
```diff
@@ -0,0 +1,262 @@
...
+
+## get/swap/addupdate implementations
+
+# `get` reads a value reading a value from a `Ref` type, a.k.a.:
+# a <- x[] -- a <=> get_p.bind(x, idx=())
```
mattjj, line 36: I'm not sure about these comments; the the thing to the left of `<=>` isn't jaxpr syntax, and it's not Python syntax, so what's it meant to be? And then what's `<=>` meant to mean exactly? (And is the `-- a` a typo?)

My guess is it's meant to be jaxpr syntax, but in that case I'd revise it to be more precise, like

```
# Staging out `a = get_p.bind(x, idx=())` where the aval of `x` is
# `ShapedArrayRef((3,), np.dtype('float32'))` leads to a jaxpr eqn printed like
#   b:f32[] <- x[]
```

WDYT?

mattjj: If the comment is about jaxpr pretty-printing, maybe it should be moved down to the pretty-printing section?

If it's instead about the semantics of the get operation and maybe its idx parameter, maybe it should be defined without referring to pseudo-syntax (i.e. the `a <- x[]` stuff) and instead using words?

mattjj (approved):
Woo! Nice tests.

jax/_src/lax/control_flow/for_loop.py
sharadmv: Yep, good idea.

sharadmv:
Yes looks like a typo.

PR #11090, opened 2022-06-14
Fix zoom state for next iteration in line search

david-anton:
If both variables _lo_to_j_ and _hi_to_lo_ are _True_, the state variables ending in __hi_ and __rec_ each have the same values (_a_hi=a_rec_, _phi_hi=phi_rec_) after the state is updated. Also, the information is lost at the old point _hi_ (_a_hi, phi_hi, dphi_hi)_. Since two points are identical, this causes issues with a possible cubic approximation in the next iteration (to few different conditions).

Changes have been made to correctly update the state for the next zoom iteration and to store all available information for quadratic and cubic minimization.


hawkinsp:
Closing as stale.

Issue #11055, opened 2022-06-10
Feature request: Add support for Chebyshev and other polynomials


gderossi:
I am working on a project that utilizes Chebyshev polynomials and might use other types of polynomials in the future, and I would like to leverage jit and auto differentiation through JAX. numpy supports Chebyshev and other polynomial types through the new polynomial package API, but this package has not been implemented in JAX. There was some brief discussion in #70 about whether it was even desirable to implement this interface in JAX, and I was wondering if there were any updates or a conclusion to this discussion.

I have already started to implement the Chebyshev convenience class and convenience functions (chebadd, chebmul, etc.) in a branch for internal use, but merging them into the main branch would be great for long-term support and ease of access.

Thanks,

hawkinsp:
In general, if something is in the NumPy API and it's useful, we welcome PRs! I don't know whether or not those particular APIs will be ergonomic or useful given the various constraints of JAX, but the best way to find that out would be to write a prototype, as you are doing. Let us know!

gderossi:
I think the Chebyshev code is ready for a PR, but I'm running into some issues with the tests. pytest -n auto tests/ gets to about 70-75% before one of the nodes crashes and is replaced. I can't see progress numbers after that, but the tests continue for a while longer before stalling forever or crashing. This happens both with my additions and with a clean copy of the main JAX branch. I'm working on an 8-year-old laptop, so the hardware might be the problem. Any suggestions?

hawkinsp:
I think the best thing you can do at this point is narrow it down to a single test that fails, and then share instructions to reproduce and I can take a look. Crashing is never supposed to happen, so that's a bug.

gderossi:
Running all of the tests individually one by one worked fine, for both the main branch and for my modified branch, so I think the crashes and stalling must have been hardware issues on my end. I did get one failure in pjit_test.py for both versions of the library though, which reported that "Some donated buffers were not usable" and "Donation is not implemented for cpu". This might be a bug? I would have expected this test to be skipped if it wasn't supported on CPU only, but perhaps not.

hawkinsp:
If you have a way for me to reproduce the crash, even using many tests, I'd like to try. JAX should never crash.

The buffer donation warning sounds like there's a test that should have been disabled but wasn't. Which test was it?

gderossi:
I've been unable to reproduce the crash, but the test that is failing is testLowerDonateArgnumsAvailable in pjit_test.py.

f0uriest:
[...]

PR #11093, opened 2022-06-14
Implement Chebyshev class and functions

gderossi:
Adds a Chebyshev convenience class and cheb* functions that act on arrays, as described in #11055.

### DIFF
--- a/jax/_src/numpy/polynomials/chebyshev.py
+++ b/jax/_src/numpy/polynomials/chebyshev.py
@@ -1,0 +1,680 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from functools import partial
+import operator
+import numpy as np
+
+import jax.numpy as jnp
+import jax.lax as lax
+from jax import jit
+from jax.numpy.linalg import eigvals
+from jax._src.numpy.util import promote_dtypes_inexact
+from jax._src.numpy.polynomials import polyutils as pu
+
+__all__ = [
+  'chebline', 'chebadd', 'chebsub', 'chebmulx', 'chebmul', 'chebdiv',
+  'chebpow', 'chebval', 'chebder', 'chebint', 'cheb2poly', 'poly2cheb',
+  'chebfromroots', 'chebvander', 'chebfit', 'chebroots', 'chebpts1',
+  'chebpts2', 'chebval2d', 'chebval3d', 'chebgrid2d', 'chebgrid3d',
+  'chebvander2d', 'chebvander3d', 'chebcompanion', 'chebgauss',
+  'chebweight', 'chebinterpolate'
+]
+
+@jit
+def _cseries_to_zseries(c):
+  n = c.size
+  zs = jnp.zeros(2*n-1, dtype=c.dtype)
+  zs = zs.at[n-1:].set(c/2)
+  return zs + zs[::-1]
+
+@jit
+def _zseries_to_cseries(zs):
+  n = (zs.size + 1)//2
+  c = zs[n-1:]
+  c = c.at[1:n].multiply(2)
+  return c
+
+@jit
+def _zseries_mul(z1, z2):
+  return jnp.convolve(z1, z2)
+
+@jit
+def _zseries_div(z1, z2):
+  lc1 = len(z1)
+  lc2 = len(z2)
+
+  if lc2 == 1:
+    z1 /= z2
+    return z1, z1[:1]*0
+  elif lc1 < lc2:
+    return z1[:1]*0, z1
+  else:
+    dlen = lc1 - lc2
+    scl = z2[0]
+    z2 /= scl
+    quo = jnp.zeros(dlen + 1, dtype=z1.dtype)
+
+    i = 0
+    j = dlen
+
+    def cond_fun(val):
+      return val[0] < val[1]
+
+    def body_fun(val):
+      i = val[0]
+      j = val[1]
+      quo = val[2]
+      z1 = val[3]
+      r = z1[i]
+      quo = quo.at[i].set(r)
+      quo = quo.at[j].set(r)
+      tmp = r*z2
+      z1 = _chebadd(jnp.roll(z1, -i), -tmp)
+      z1 = _chebadd(jnp.roll(z1, i-j), -tmp)
+      z1 = jnp.roll(z1, j)
+      i += 1
+      j -= 1
+      return(i, j, quo, z1)
+
+    i, j, quo, z1 = lax.while_loop(cond_fun, body_fun, (i, j, quo, z1))
+    r = z1[i]
+    quo = quo.at[i].set(r)
+    tmp = r*z2
+    z1 = _chebadd(jnp.roll(z1, -i), -tmp)
+    quo /= scl
+    rem = jnp.roll(z1, -1)[:lc2-2]
+    return quo, rem
+
+@jit
+def _poly2cheb(pol):
+  [pol] = pu._as_series([pol])
+  pol, = promote_dtypes_inexact(pol)
+  deg = len(pol) - 1
+  res = jnp.zeros(deg+1, dtype=pol.dtype)
+  pol = jnp.flip(pol)
+
+  def body_fun(res, p):
+    prd = jnp.zeros(len(res), dtype=res.dtype)
+    prd = prd.at[1].set(res[0])
+    if len(res) > 1:
+      tmp = res[1:-1]/2
+      prd = prd.at[2:].set(tmp)
+      prd = prd.at[0:-2].add(tmp)
+    prd = prd.at[0].add(p)
+    return prd, p
+
+  res, _ = lax.scan(body_fun, res, pol)
+  return res
+
+def poly2cheb(pol, trim=True):
+  if trim:
+    return _poly2cheb(pu.trimseq(pol))
+  else:
+    return _poly2cheb(pol)
+
+@jit
+def _cheb2poly(c):
+  [c] = pu._as_series([c])
+  c, = promote_dtypes_inexact(c)
+  n = len(c)
+  if n < 3:
+    return c
+  else:
+    c0 = jnp.zeros(len(c), dtype=c.dtype)
+    c0 = c0.at[0].set(c[-2])
+    c1 = jnp.zeros(len(c), dtype=c.dtype)
+    c1 = c1.at[0].set(c[-1])
+
+    def body_fun(i, val):
+      v0 = _chebsub(c[-i-2], val[1])
+      v1 = jnp.zeros(len(val[1]), dtype=val[1].dtype)
+      v1 = v1.at[1:].set(val[1][:-1])
+      v1 = _chebadd(val[0], 2*v1)
+      return(v0, v1)
+
+    c0, c1 = lax.fori_loop(-(n-1),-1, body_fun, (c0, c1))
+
+    tmp = jnp.zeros(len(c1), dtype=c1.dtype)
+    c1 = tmp.at[1:].set(c1[:-1])
+
+    return _chebadd(c0, c1)
+
+def cheb2poly(c, trim=True):
+  if trim:
+    return _cheb2poly(pu.trimseq(c))
+  else:
+    return _cheb2poly(c)
+
+@jit
+def _chebline(off, scl):
+  return jnp.array([off, scl])
+
+def chebline(off, scl, trim=True):
+  if trim:
+    return pu.trimseq(_chebline(off, scl))
+  else:
+    return _chebline(off, scl)
+
+@jit
+def chebfromroots(roots):
+  if len(roots) == 0:
+    return jnp.ones(1, dtype=roots.dtype)
+  else:
+    [roots] = pu._as_series([roots])
+    roots = roots.sort()
+    retlen = len(roots)+1
+
+    def p_scan_fun(carry, x):
+      return carry, _chebadd(jnp.zeros(retlen, dtype=x.dtype), jnp.array([-x, 1], dtype=x.dtype))
+
+    _, p = lax.scan(p_scan_fun, 0, roots)
+
+    p = jnp.asarray(p)
+    n = len(p)
+
+    def cond_fun(val):
+      return val[0] > 1
+
+    def body_fun(val):
+      m, r = divmod(val[0], 2)
+      arr = val[1]
+      tmp = jnp.array([jnp.zeros(retlen, dtype=p.dtype)]*len(p))
+
+      def inner_body_fun(i, val):
+        return val.at[i].set(_chebmul(arr[i], arr[i+m])[:retlen])
+
+      tmp = lax.fori_loop(0, m, inner_body_fun, tmp)
+      tmp = lax.cond(r, lambda x: x.at[0].set(_chebmul(x[0], arr[2*m])[:retlen]), lambda x: x, tmp)
+
+      return(m, tmp)
+
+    _, ret = lax.while_loop(cond_fun, body_fun, (n, p))
+    return ret[0]
+
+@jit
+def _chebadd(c1, c2):
+  [c1, c2] = pu._as_series([c1, c2])
+  c1, c2 = promote_dtypes_inexact(c1, c2)
+  if len(c1) > len(c2):
+    c1 = c1.at[:c2.size].add(c2)
+    ret = c1
+  else:
+    c2 = c2.at[:c1.size].add(c1)
+    ret = c2
+  return ret
+
+def chebadd(c1, c2, trim=True):
+  if trim:
+    return pu.trimseq(_chebadd(c1, c2))
+  else:
+    return _chebadd(c1, c2)
+
+@jit
+def _chebsub(c1, c2):
+  [c1, c2] = pu._as_series([c1, c2])
+  c1, c2 = promote_dtypes_inexact(c1, c2)
+  if len(c1) > len(c2):
+    c1 = c1.at[:c2.size].add(-c2)
+    ret = c1
+  else:
+    c2 = -c2
+    c2 = c2.at[:c1.size].add(c1)
+    ret = c2
+  return ret
+
+def chebsub(c1, c2, trim=True):
+  if trim:
+    return pu.trimseq(_chebsub(c1, c2))
+  else:
+    return _chebsub(c1, c2)
+
+@jit
+def _chebmulx(c):
+  [c] = pu._as_series([c])
+  c, = promote_dtypes_inexact(c)
+  prd = jnp.zeros(len(c) + 1, dtype=c.dtype)
+  prd = prd.at[1].set(c[0])
+  if len(c) > 1:
+    tmp = c[1:]/2
+    prd = prd.at[2:].set(tmp)
+    prd = prd.at[0:-2].add(tmp)
+  return prd
+
+def chebmulx(c, trim=True):
+  if trim:
+    return pu.trimseq(_chebmulx(pu.trimseq(c)))
+  else:
+    return _chebmulx(c)
+
+@jit
+def _chebmul(c1, c2):
+  [c1, c2] = pu._as_series([c1, c2])
+  c1, c2 = promote_dtypes_inexact(c1, c2)
+  z1 = _cseries_to_zseries(c1)
+  z2 = _cseries_to_zseries(c2)
+  prd = _zseries_mul(z1, z2)
+  ret = _zseries_to_cseries(prd)
+  return ret
+
+def chebmul(c1, c2, trim=True):
+  if trim:
+    return pu.trimseq(_chebmul(pu.trimseq(c1), pu.trimseq(c2)))
+  else:
+    return _chebmul(c1, c2)
+
+@jit
+def _chebdiv(c1, c2):
+  [c1, c2] = pu._as_series([c1, c2])
+  c1, c2 = promote_dtypes_inexact(c1, c2)
+  lc1 = len(c1)
+  lc2 = len(c2)
+  if lc1 < lc2:
+    return c1[:1]*0, c1
+  elif lc2 == 1:
+    return c1/c2[-1], c1[:1]*0
+  else:
+    z1 = _cseries_to_zseries(c1)
+    z2 = _cseries_to_zseries(c2)
+    quo, rem = _zseries_div(z1, z2)
+    quo = _zseries_to_cseries(quo)
+    rem = _zseries_to_cseries(rem)
+    return quo, rem
+
+def chebdiv(c1, c2, trim=True):
+  if trim:
+    quo, rem = _chebdiv(pu.trimseq(c1), pu.trimseq(c2))
+    return pu.trimseq(quo), pu.trimseq(rem)
+  else:
+    return _chebdiv(c1, c2)
+
+@partial(jit, static_argnames=('pow', 'maxpower'))
+def _chebpow(c, pow, maxpower=16):
+  [c] = pu._as_series([c])
+  c, = promote_dtypes_inexact(c)
+  power = operator.index(pow)
+  if power != pow or power < 0:
+    raise ValueError("Power must be a non-negative integer.")
+  elif maxpower is not None and power > maxpower:
+    raise ValueError("Power is too large.")
+  elif power == 0:
+    return jnp.array([1], dtype=c.dtype)
+  elif power == 1:
+    return c
+  else:
+    zs = jnp.zeros(len(c)+(len(c)-1)*(pow-1), dtype=c.dtype)
+    zs = _chebadd(zs, c)
+    zs = _cseries_to_zseries(zs)
+    prd = zs
+
+    def body_fun(i, val):
+      return jnp.convolve(val, zs, mode='same')
+
+    prd = lax.fori_loop(2, power+1, body_fun, prd)
+    return _zseries_to_cseries(prd)
+
+def chebpow(c, pow, maxpower=16, trim=True):
+  if trim:
+    return _chebpow(pu.trimseq(c), pow, maxpower=maxpower)
+  else:
+    return _chebpow(c, pow, maxpower=maxpower)
+
+@partial(jit, static_argnames=('m',))
+def chebder(c, m=1):
+  [c] = pu._as_series([c])
+  c, = promote_dtypes_inexact(c)
+  cnt = operator.index(m)
+  if cnt < 0:
+    raise ValueError("The order of derivation must be non-negative")
+  if cnt == 0:
+    return c
+
+  n = len(c)
+  if cnt >= n:
+    c = c.at[:1].multiply(0)
+    c = c[:1]
+  else:
+
+    def outer_body_fun(i, outer_val):
+      inner_start = len(outer_val)-i-1
+      tmp = jnp.zeros(len(outer_val), dtype=outer_val.dtype)
+
+      def inner_body_fun(j, inner_val):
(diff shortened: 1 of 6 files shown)


jakevdp:
Hi - thanks for the PR! In the past when we've discussed the polynomial interface, we've made the choice to avoid adding polynomial class types in favor of the more direct functional API (e.g. https://github.com/google/jax/issues/70#issuecomment-907892885). One concern I have here is that when you start to define class-based interfaces, you have to be much more careful to ensure that they're compatible with JAX's transformation model.

Is there any functionality here that cannot be expressed directly via chebyshev functions, as opposed to a full chebyshev class?

jakevdp:
I would lean toward keeping only the simplest functional APIs in JAX, partly because it makes long-term maintenance more sustainable (the class definition in this PR is a **gigantic** API surface to test and maintain). But I'm happy to let other team members weigh-in if they disagree.

gderossi:
Alright, I removed the Chebyshev class and its abstract base class to streamline the API. If it's decided later that they are worth including, I have local copies and can add them back in.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+    zs *= jnp.arange(-n, n+1)*2
+    d, r = _zseries_div(zs, ns)
+    return d
+
+@jit
+def _zseries_int(zs):
+    n = 1 + len(zs)//2
+    ns = jnp.array([-1, 0, 1], dtype=zs.dtype)
+    zs = _zseries_mul(zs, ns)
+    div = jnp.arange(-n, n+1)*2
```
jakevdp, line 93: make this `zs.dtype` to avoid unintentional dtype promotion.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+def cheb2poly(c):
+    [c] = pu.as_series([c])
+    n = len(c)
+    if n < 3:
+        return c
+    else:
+        c0 = jnp.asarray(c[-2])
+        c1 = jnp.asarray(c[-1])
+
+        for i in range(n-1, 1, -1):
```
jakevdp, line 118: Same comment here about Python control flow when the size gets large. We may want to use `lax.fori_loop` instead unless `n` is bounded by a small number.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+            prd = prd.at[0].set(0)
+            prd = prd.at[1:].set(c1)
+            c1 = chebadd(tmp, prd*2)
+        prd = jnp.empty(len(c1)+1, dtype=c1.dtype)
+        prd = prd.at[0].set(0)
+        prd = prd.at[1:].set(c1)
+        return chebadd(c0, prd)
+
+#Chebyshev default domain
+chebdomain = jnp.array([-1, 1])
```
jakevdp, line 134: We cannot create jax.numpy arrays at import time, because in some systems XLA will not yet be initialized when this is executed. Options are (1) use `np.array`, or (2) if `jnp.array` is important, define this lazily using a module `__getattr__`

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+    zs = zs.at[n+1:].divide(div[n+1:])
+    zs = zs.at[n].set(0)
+    return zs
+
+@jit
+def poly2cheb(pol):
+    [pol] = pu.as_series([pol])
+    deg = len(pol) - 1
+    res = 0
+    for i in range(deg, -1, -1):
```
jakevdp, line 104: Also here: consider `fori_loop`

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+@jit
+def chebfromroots(roots):
+    if len(roots) == 0:
+        return jnp.ones(1)
+    else:
+        [roots] = pu.as_series([roots], trim=False)
+        roots.sort()
+        p = [chebline(-r, 1) for r in roots]
+        n = len(p)
+        while n > 1:
```
jakevdp, line 161: consider `while_loop`

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+@partial(jit, static_argnames=('trim',))
+def chebmul(c1, c2, trim=False):
+    [c1, c2] = pu.as_series([c1, c2])
+    c1, c2 = _promote_dtypes_inexact(c1, c2)
+    z1 = _cseries_to_zseries(c1)
+    z2 = _cseries_to_zseries(c2)
+    prd = _zseries_mul(z1, z2)
+    ret = _zseries_to_cseries(prd)
+    if trim:
+        return pu.trimseq(ret)  #NO JIT
```
jakevdp, line 216: What does `NO JIT` here mean? You've wrapped the outer function in `JIT` so this will be executed within a jit context.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+        return c1/c2[-1], c1.at[:1].multiply(0)
+    else:
+        z1 = _cseries_to_zseries(c1)
+        z2 = _cseries_to_zseries(c2)
+        quo, rem = _zseries_div(z1, z2)
+        quo = _zseries_to_cseries(quo)
+        rem = _zseries_to_cseries(rem)
+        if trim:
+            quo = pu.trimseq(quo)  #NO JIT
+            rem = pu.trimseq(rem)  #NO JIT
```
jakevdp, line 238: Same comment here: the jit decorator above will apply to these lines.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+        return quo, rem
+
+@partial(jit, static_argnames=('pow', 'maxpower'))
+def chebpow(c, pow, maxpower=16):
+    [c] = pu.as_series([c])
+    c = _promote_dtypes_inexact(c)[0]  #Surely there is a better way to do this
```
jakevdp, line 244:
```python
c, = _promote_dtypes_inexact(c)
```

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+    c = jnp.moveaxis(c, iaxis, 0)
+    n = len(c)
+    if cnt >= n:
+        c = c.at[:1].multiply(0)
+        c = c[:1]
+    else:
+        for i in range(cnt):
+            n = n-1
+            c *= scl
+            der = jnp.empty((n,) + c.shape[1:], dtype=c.dtype)
```
jakevdp, line 282: should this be `jnp.zero` rather than `jnp.empty`? `empty` implies uninitialized values, and below it looks like you're adding numbers to them.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+    v = functools.reduce(operator.mul, vander_arrays)
+
+    return v.reshape(v.shape[:-len(deg)] + (-1,))
+
+@partial(jit, static_argnames=('deg_x', 'deg_y', 'deg_z'))
+def chebvander3d(x, y, z, deg_x, deg_y, deg_z):
+    points = (x, y, z)
+    deg = (deg_x, deg_y, deg_z)
+    # convert to the same shape and type
+    points = tuple(jnp.array(tuple(points), copy=False) + 0.0)
```
jakevdp, line 441: This looks like a roundabout way of doing `_promote_dtypes_inexact`; you should instead call that directly.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+
+    # we checked this wasn't empty already, so no `initial` needed
+    v = functools.reduce(operator.mul, vander_arrays)
+
+    return v.reshape(v.shape[:-len(deg)] + (-1,))
+
+@partial(jit, static_argnames=('deg', 'full'))
+def chebfit(x, y, deg, rcond=None, full=False, w=None):
+    x = jnp.asarray(x) + 0.0
+    y = jnp.asarray(y) + 0.0
```
jakevdp, line 459: Use `promote_dtypes_inexact` instead of adding a float.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,605 @@
...
+        van = chebvander(x, lmax)
+    else:
+        deg = jnp.sort(deg)
+        lmax = deg[-1]
+        van = chebvander(x, lmax)[:, deg]
+
+    lhs = van.T
+    rhs = y.T
+    if w is not None:
+        w = jnp.asarray(w) + 0.0
```
jakevdp, line 487: `promote_dtypes_inexact`

jax/_src/numpy/polynomials/polyutils.py
```diff
@@ -0,0 +1,81 @@
...
+#Works with JIT if static_argnames('trim',) and trim=False
+@partial(jit, static_argnames=('trim',))
+def as_series(alist, trim=False):
+    arrays = [jnp.array(a, ndmin=1, copy=False) for a in alist]
+    if min([a.size for a in arrays]) == 0:
+        raise ValueError("Coefficient array is empty")
+    if any(a.ndim != 1 for a in arrays):
+        raise ValueError("Coefficient array is not 1-d")
+    if trim:
+        arrays = [trimseq(a) for a in arrays]
```
jakevdp, line 35: This line will fail because of the `jit` decoration above and the incompatbility of `trimseq` with JIT

jax/_src/numpy/polynomials/polyutils.py
```diff
@@ -0,0 +1,81 @@
...
+    if any(a.dtype == jnp.dtype(object) for a in arrays):
+        ret = []
+        for a in arrays:
+            if a.dtype != jnp.dtype(object):
+                tmp = jnp.empty(len(a), dtype = jnp.dtype(object))
```
jakevdp, line 40: JAX does not support `dtype=object`

jax/_src/numpy/polynomials/polyutils.py
```diff
@@ -0,0 +1,81 @@
...
+                tmp = tmp.at[:].set(a[:])
+                ret.append(tmp)
+            else:
+                ret.append(a.copy())
+    else:
+        ret = [jnp.array(a, copy=True) for a in arrays]
```
jakevdp, line 46: No need to copy because JAX arrays are immutable

jakevdp (changes requested):
OK, a bunch of comments below. First-off, there is a lot in here that is incompatible or inefficent with JAX's computation model (for example, loop-based implementations, object arrays, value-dependent shapes)

Before moving forward with this, we should probably think about whether those pieces of the implementation are important; if so, I'd probably lean toward not including these APIs in JAX, because the user experience will not be good (incompatibility with JAX transforms is not great for JAX library code).

On the other hand, if the APIs can be changed in a way that they will become compatible with JAX, then we should think about making those changes before diving-in to a review at the level of granularity of my comments below.

What do you think?

gderossi:
Alright, I've cleaned up a lot of the code following your suggestions and to better match the existing polynomial functions in JAX. I've also been thinking more about JAX compatibility, and I wanted to ask a couple questions. The object arrays and value-based shapes were unnecessary, but loop implementations are necessary (or at least the natural way to implement many functions) because of the relationships between terms in a Chebyshev series. 

So, what do you consider "large" for a series or loop parameter? If a typical series is large, would replacing Python control flow  with JAX control flow primitives resolve the incompatibility/inefficiency issues? If so, I will make those changes.

jakevdp:
The issue with Python loops is that they are flattened by JAX tracing, and then passed in an unrolled state to the XLA compiler. The mechanics of the compiler are complicated, but a good rule of thumb is that compilation time will be roughly quadratic in the number of operations. So if you write a Python `for` loop over 100 elements, then compilation will be roughly 10000 times more costly than the equivalent `fori_loop` implementation. There's not really any hard line regarding what size the loops should be, but if you expect the functions to operate on larger inputs, it's something to keep in mind. Does that make sense?

gderossi:
Ok, I've removed all Python loops from chebyshev.py and replaced them with JAX control flow primitives. I've also cleaned up most of the functions to make them less convoluted where possible. Is there anything else you would like me to do?
Thank you!

jakevdp:
Thanks. Another big piece of adding this kind of functionality is testing – before we can merge this, we'll need full test coverage of the functionality. If it's helpful, you can base it off of existing tests in https://github.com/google/jax/blob/main/tests/polynomial_test.py. Let me know if you have questions about that!

Also, just FYI we have some work in progress that will add some constraints to this code (see #10840): essentially, JAX library code will no longer be able to rely on implicit dtype promotion once that is merged. You can run your tests locally with that condition by setting the flag value `JAX_NUMPY_DTYPE_PROMOTION=strict`

gderossi:
Alright, I have a couple questions. First of all, many NumPy Chebyshev functions trim trailing zeros before returning, a feature I omitted from the JAX versions for jit compatibility. When comparing against NumPy, should I just wrap JAX-based versions of these functions in a trimzeros function, or is there a better way to test that ignores length differences? 

Second, what is the desired behavior for functions under the new dtype promotion constraints? Should functions like chebadd raise an error if the input arrays have different dtypes, should they attempt explicit dtype promotion, or should they do something else?

Thank you!

jakevdp:
Excellent question - one strategy we typically use in this situation (where the numpy behavior is incompatible with JIT) is to add an additional keyword argument, set so that the default value matches the behavior of numpy, and so that it can be changed to modify the API to be compatible with JIT (you can see an example in the `strip_zeros` keyword to `jnp.roots`, which I just finished revamping in #11215)

I think you could use a very similar treatment here.

gderossi:
Yes, that all makes sense. Thank you!

jakevdp:
Regarding precision – I suspect this is due to JAX executing in float32, while numpy executes in float64.

Regarding warnings - you can filter them if they're expected. You can see examples of this elsewhere in the tests, e.g. https://github.com/google/jax/blob/4c25ef1d0039ecdd517ce2b09605322797700c2a/tests/lax_numpy_test.py#L579-L584

gderossi:
Ok, all test pass with JAX_ENABLE_x64=true and JAX_NUMPY_DTYPE_PROMOTION=strict. I also updated applicable functions with a "trim" keyword argument so that they match NumPy behavior by default. What's the next step?
Thank you!

tests/chebyshev_test.py
```diff
@@ -0,0 +1,795 @@
...
+  def testChebline(self, dtype, mul):
+    rng = jtu.rand_default(self.rng())
+
+    def args_maker():
+      off, scl = rng(2, dtype)
+      scl *= mul
+      return (off, scl)
+
+    jnp_fn = lambda off, scl: cs.chebline(off, scl)
+    np_fn = lambda off, scl: ncs.chebline(off, scl)
```
jakevdp, line 38: If there are no arguments, you don't need to use an anonymous function. `jnp_fun = cs.chebline` would suffice (or, you can skip the renaming and just use the function directly in the check below)

Same for other tests below.

tests/chebyshev_test.py
```diff
@@ -0,0 +1,795 @@
...
+    self._CheckAgainstNumpy(np_fn, jnp_fn, args_maker, check_dtypes=False,
+                            tol=1e-6)
```
jakevdp, line 40: Let's also call `self._CompileAndCheck`, here and in tests below.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,653 @@
+from functools import partial
```
jakevdp, line 1: We need a license statement at the top of each file.

Untitled.ipynb
```diff
@@ -0,0 +1,441 @@
+{
```
jakevdp, line 1: Unintentionally added?

tests/chebyshev_test.py
```diff
@@ -0,0 +1,795 @@
...
+import jax.numpy.polynomials as pu
+from jax.numpy.polynomials import chebyshev as cs
+
+from jax.config import config
+config.parse_flags_with_absl()
+
+if config.jax_enable_x64:
+  all_dtypes = ["int64", "float64", "complex128"]
+else:
+  all_dtypes = ["int32", "float32", "complex64"]
```
jakevdp, line 20: Better would be to use `jtu.dtypes`, which contains the enable_x64 logic

tests/chebyshev_test.py
```diff
@@ -0,0 +1,795 @@
...
+      leading2, trailing2),
+      "dtype": dtype, "length1": length1, "leading1": leading1, "trailing1": trailing1,
+      "length2": length2, "leading2": leading2, "trailing2": trailing2}
+      for dtype in all_dtypes
+      for length1 in [1, 3, 9, 10, 17]
+      for length2 in [1, 3, 9, 10, 17]
+      for leading1 in [0, 1, 2, 7]
+      for leading2 in [0, 1, 2, 7]
+      for trailing1 in [0, 1, 2, 7]
+      for trailing2 in [0, 1, 2, 7]))
```
jakevdp, line 57: This test generates a HUGE number of test cases (3 * 5 * 5 * 4 * 4 * 4 * 4 = 19200). The most we ever check in CI is 50 (via `jtu.cases_from_list`), but all 19200 pairs will be generated when the test file is loaded. That tends to lead to needlessly long test load times. Consider reducing the number of test cases you generate. Same comment applies below.

jakevdp (changes requested):
Thanks! A few additional comments

tests/chebyshev_test.py
gderossi: Using all_dtypes = jtu.dtypes.floating + jtu.dtypes.integer + jtu.dtypes.complex, each test runs with int32, int64, float32, float64, complex64, and complex128 dtypes when JAX_ENABLE_x64=true. Some of the tests fail when using float32 due to value mismatches. 

Removing jit decorators and replacing any compiled sections inside the function seems to fix this problem. For example, chebval fails one float32 test case and chebgrid2d (which calls chebval) fails two float32 test cases when jit decorators and JAX control flow are used. Replacing/removing these allows these tests to succeed. 

Doesn't work:
```
@partial(jit, static_argnames=('tensor',))
def chebval(x, c, tensor=True):
  [irrelevant code here]
  def body_fun(i, val):
    return (c[-i] - val[1], val[0] + val[1]*x2)

c0, c1 = lax.fori_loop(5, len(c)+1, body_fun, (c0, c1))
```

Still doesn't work:
```
@partial(jit, static_argnames=('tensor',))
def chebval(x, c, tensor=True):
  [irrelevant code here]
  for i in range(5, len(c) + 1):
    c0, c1 = (c[-i] - c1, c0 + c1 * x2)
```
Works for chebval, but chebgrid2d float32 test cases still fail:
```
def chebval(x, c, tensor=True):
  [irrelevant code here]
  def body_fun(i, val):
    return (c[-i] - val[1], val[0] + val[1]*x2)

c0, c1 = lax.fori_loop(5, len(c)+1, body_fun, (c0, c1))
```


Does work:
```
def chebval(x, c, tensor=True):
  [irrelevant code here]
  for i in range(5, len(c) + 1):
    c0, c1 = (c[-i] - c1, c0 + c1 * x2)
```

Any thoughts or suggestions for how to fix this while keeping JAX control flow and compilation? This issue only happens for a few float32 tests, and the relative difference from NumPy values is on the order of 1e-5.

gderossi:
Implemented the changes you requested, and hopefully fixed what caused the chebcompanion test to fail last time.

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,679 @@
...
+
+def cheb2poly(c, trim=True):
+  if trim:
+    return _cheb2poly(pu.trimseq(c))
+  else:
+    return _cheb2poly(c)
+
+@jit
+def _chebline(off, scl):
+  return jnp.array([off, scl])#, dtype=jnp.promote_types(off, scl))
```
jakevdp, line 162: Remove comment?

jax/_src/numpy/polynomials/chebyshev.py
```diff
@@ -0,0 +1,679 @@
...
+  else:
+    return _chebline(off, scl)
+
+@jit
+def chebfromroots(roots):
+  if len(roots) == 0:
+    return jnp.ones(1, dtype=roots.dtype)
+  else:
+    [roots] = pu._as_series([roots])
+    roots.sort()
```
jakevdp, line 176:
```suggestion
    roots = roots.sort()
```
`sort` is not an inplace operation.

tests/chebyshev_test.py
```diff
@@ -0,0 +1,812 @@
...
+      for trailing in [0, 1]
+      for pow in [0,1,2,10]))
+  def testChebpow(self, dtype, length, trailing, pow):
+    rng = jtu.rand_default(self.rng())
+
+    def args_maker():
+      c = rng((length,), dtype)
+      return (jnp.concatenate([c, jnp.zeros(trailing, c.dtype)]), pow)
+
+    jnp_fn = lambda c, _p: partial(cs.chebpow, pow=pow, trim=False)(c)
```
jakevdp, line 221: I'd use either `lambda` or `partial`, but not both together, as it is a bit confusing to parse.

Same in several places below.

jakevdp: Also, it's strange here that the tested function ignores the second input. If it needs to be static, then we shouldn't generate it as part of `args_maker`, but rather use a `partial` for both the jax and numpy version of the function. You can see that pattern elsewhere in the test suite when testing static arguments.

Same comment applies to many of the test cases below.

tests/chebyshev_test.py
```diff
@@ -0,0 +1,812 @@
...
+    rng = jtu.rand_default(self.rng())
+
+    def args_maker():
+      c = rng((length,), dtype_c)
+      if pts == 0:
+        p, = rng(1, dtype_x)
+      else:
+        p = rng((pts,), dtype_x)
+      t = tensor == 1
+      return (p, jnp.concatenate([c, jnp.zeros(trailing, c.dtype)]), t)
```
jakevdp, line 249: This is concerning - the numpy version always gets `t = 1`, while the JAX version below ignores the value of `t` and uses `tensor`, which can be zero or 1. If this test is passing, it makes me think the implementation is incorrect.

jakevdp (changes requested):
Thanks - a few comments on the test suite.

Also, I'm curious: how long does the test suite take to run? It's a lot of new tests with large parameterization spaces, and I'd be concerned this will push our CI timeout limits. You can check with a command like this (using 25 generated cases per test to match the Kokoro settings):
```
$ JAX_NUM_GENERATED_CASES=25 JAX_ENABLE_X64=1 time pytest -n auto tests/chebyshev_test.py
```

gderossi:
Made the changes you requested. Also, the Chebyshev test suite takes 75-80 seconds to run on my laptop using the command you provided.

gderossi:
After some more testing, I believe the Chebyshev tests actually contribute a much smaller percentage of total testing time: I ran the full test suite on my laptop yesterday, and it took about 2 hours and 18 minutes, or ~8300 seconds. Running the Chebyshev test file individually took 75-80 seconds, which is 0.9-1% of the total. I just ran the test suite remotely on a much more powerful computer, which completed all the tests in 1270 seconds and the Chebyshev tests alone in 11 seconds, which is ~0.866% of the total.
I'm happy to repeat these tests or try other tests you think would be helpful to verify this ratio. 1% of the testing overhead is still significant, so I've also experimented with reducing the Chebyshev tests, as I noticed that polynomial_tests.py only includes tests for jnp.roots. 
Many of the cheb* functions call other cheb* functions - for example, chebpow calls chebadd, so the chebadd tests could potentially be removed. Changes that break chebadd would (hopefully) be caught by the chebpow tests. Another category of functions, the 2d and 3d versions of chebval/chebgrid/chebvander, are simply wrappers for nested calls of the respective 1d functions. Tests for these could also be removed, as the 1d function tests should cover everything.
Removing tests for all functions that fall under the two categories described above cuts the time of the Chebyshev tests in half, and a majority of this cost is due to only a few functions: chebfit, chebdiv, and chebfromroots. Reducing the number of test cases for these particular functions or eliminating tests for them entirely could significantly speed up the Chebyshev tests.

Thank you for your time, let me know what you think!

jakevdp:
Thanks - that makes sense. Let's go ahead and merge this and we can deal with any test timeout issues separately. Thanks for all the work here!

jakevdp (approved):

jakevdp:
The test failures here look real: it looks like `chebgrid` does some implicit type promotion when it is passed integer inputs.

gderossi:
Ok, the chebroots test is now skipped on GPU and TPU. That test also no longer generates inputs with untrimmed trailing zeros, which I expect will fix the other problem.

jakevdp:
Hi - there are still a number of failures in the larger tests. I think it's easier for me to try to patch the imported PR to adjust test tolerances; I'll try to look at that today.

tests/chebyshev_test.py
```diff
@@ -0,0 +1,823 @@
...
+      off = rng(1, dtype_off)
+      scl = rng(1, dtype_scl)
+      scl  = (scl*mul).astype(dtype_scl)
+      return (off, scl)
+
+    jnp_fn = partial(cs.chebline, trim=False)
+
+    with jtu.strict_promotion_if_dtypes_match([dtype_off, dtype_scl]):
+      self._CheckAgainstNumpy(ncs.chebline, cs.chebline, args_maker, check_dtypes=False,
+                              tol=1e-6)
```
jakevdp, line 56: One reason I'm having trouble landing this, I think, is because you've hard-coded tolerances like `1E-6` throughout. When the tolerance is unspecified, it defaults to an appropriate tolerance for a given dtype. When you override this with a single number, it uses that number regardless of the dtype. I think tests may behave better if you remove these default tolerances; or if you must specify them, do something like `tol={np.float32: 1e-6}` to change the default tolerance only for a particular dtype.

gderossi: Alright, I left tolerances unspecified for all tests except for a few. For those few, I specified tolerances by dtype as done in lax-numpy-test.

gderossi:
Hi, is there anything else you'd like me to change or add before the workflow is approved?

jakevdp:
Hi - sorry I lost track of this over the summer. Let me see if I can land it now

jakevdp:
For the sake of bazel-based tests, can you add the following block for the new test file in https://github.com/google/jax/blob/main/tests/BUILD, keeping alphabetical order?
```
jax_test(
    name = "chebyshev_test",
    srcs = ["chebyshev_test.py"],
)
```

jakevdp:
We're still running into a bunch of tolerance-related failures on GPU and TPU.

f0uriest:
Any updates on this? I'm also very interested in using orthogonal polynomials in JAX and was planning on building off of this to implement some of the others from numpy

gderossi:
> Any updates on this? I'm also very interested in using orthogonal polynomials in JAX and was planning on building off of this to implement some of the others from numpy

I haven't actually had a chance to work on this since August- I started grad school and just haven't had time, but I'm hoping to investigate the tolerance issues this weekend/over the Thanksgiving holiday. Happy to chat or work with you on this if you have any suggestions or ideas.

nzolman:
Is there any plan for this to continue? I'm looking to use Chebyshev polynomials with JAX on a project.

gderossi:
Wanted to bump this for review- I've updated everything to work with the new version of JAX, and tested on CPU, GPU, and TPU. Made some small adjustments, and now all tests are passing on my end. Thank you!

jakevdp:
Hi - sorry we lost track of these. I think given the discussions around https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html, this would now be considered out-of-scope for jax itself. That said, I think there are a lot of people who would find something like this useful, and you might think about releasing this functionality as part of another package.

Sorry we weren't able to get this merged – thanks!

FelixBenning:
@gderossi did you ever get around releasing this as a package? I would be interested in using it

PR #11103, opened 2022-06-14
[x64] make api_test compatible with strict dtype promotion

jakevdp:

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -209,7 +209,7 @@ class CPPJitTest(jtu.BufferDonationTestCase):
 
     def f(x, y, z):
       side.append(None)
-      return 100 * x + 10 * y + z
+      return 100 * x + 10 * y + z.astype(y.dtype)
 
     f = self.jit(f)
     assert f(one, two, three) == 123
@@ -386,7 +386,7 @@ class CPPJitTest(jtu.BufferDonationTestCase):
   def test_jit_donate_argnums_warning_raised(self):
     x = jnp.array([1.0, 2.0], jnp.float32)
     y = jnp.array([1, 2], jnp.int32)
-    f = self.jit(lambda x, y: x.sum() + y.sum(), donate_argnums=(0, 1))
+    f = self.jit(lambda x, y: x.sum() + jnp.float32(y.sum()), donate_argnums=(0, 1))
     with warnings.catch_warnings(record=True) as w:
       warnings.simplefilter("always")
       f(x, y)
@@ -1440,9 +1440,9 @@ class APITest(jtu.JaxTestCase):
   def test_hessian_holomorphic(self):
     R = self.rng().randn
     A = R(4, 4)
-    x = R(4) * (1 + 2j)
+    x = R(4).astype('complex64') * (1 + 2j)
 
-    f = lambda x: jnp.dot(x, jnp.dot(A, x))
+    f = lambda x: jnp.dot(x, jnp.dot(A.astype(x.dtype), x))
     assert np.allclose(hessian(f, holomorphic=True)(x), A + A.T)
 
   @jtu.skip_on_devices("tpu")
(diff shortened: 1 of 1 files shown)


tests/api_test.py
```diff
@@ -222,7 +222,8 @@ def f(x, y, z):
     assert f(one, two, z=three) == 123
     assert len(side) == 2  # but should still cache
 
-    f(one, two, z=np.zeros(3))  # doesn't crash
+    with jax.numpy_dtype_promotion('standard'):
```
hawkinsp, line 225: Is this really supposed to be testing promotion?

tests/api_test.py
```diff
@@ -383,6 +384,7 @@ def f(args_list):
 
   # Jit and Donate arguments
 
+  @jax.numpy_dtype_promotion('standard')  # Test explicitly exercises implicit dtype promotion.
```
hawkinsp, line 387: Is this really supposed to be testing promotion?

In general I'd rather avoid promotion if it's not the point of the test, and it pretty clearly is not here.

tests/api_test.py
```diff
@@ -1437,6 +1439,7 @@ def test_hessian(self):
     assert np.allclose(hessian(f)(x), A + A.T)
 
   @jtu.skip_on_devices("tpu")
+  @jax.numpy_dtype_promotion('standard')  # Test explicitly exercises implicit dtype promotion.
```
hawkinsp, line 1442: Same everywhere.

tests/api_test.py
jakevdp: I don't know if it's "really" supposed to be, but it is. My aim was to modify tests as little as possible. Currently the tests run (and pass) with standard promotion, so after this change ISTM they should still run and pass with standard promotion. What do you think?

tests/api_test.py
jakevdp: In this case, since the holomorphic hessian specifically requires certain types for its inputs, we would lose test coverage if we only tested the case where all inputs are complex. It seems better to me to keep the current test coverage rather than reducing its scope.

tests/api_test.py
jakevdp: Argument donation behavior depends on the dtype of the inputs, and this test explicitly passes inputs with different types to the function. With that in mind, I think standard dtype promotion is required to test this behavior. Would you prefer I cast the inputs manually so we can test the same behavior under strict promotion?

tests/api_test.py
```diff
@@ -1828,6 +1831,7 @@ def f(z):
     expected = grad(f)(zs)
     self.assertAllClose(ans, expected)
 
+  @jax.numpy_dtype_promotion('standard')  # Test explicitly exercises implicit dtype promotion.
```
jakevdp, line 1834: Here, the test is explicitly testing heterogenous inputs (it's even in the test name!) so changing to homogeneous inputs seems like the wrong decision.

tests/api_test.py
```diff
@@ -1847,6 +1851,7 @@ def f(tup):
     jtu._check_dtypes_match(actual, desired)
     jtu.check_eq(actual, desired)
 
+  @jax.numpy_dtype_promotion('standard')  # Test explicitly exercises implicit dtype promotion.
```
jakevdp, line 1854: Same here: test explicitly is testing heterogeneous inputs, so I don't think we should change that.

tests/api_test.py
hawkinsp: I can buy a strategy of:
a) first mechanically enable dtype promotion everywhere it is needed,
b) then go and clean up its uses
if you want to do it that way.

However I don't want our tests to be littered with overrides on the promotion behavior just because we happened (casually) to use promotion in the past...

For example, in this particular case, I think we were just lazy about the type of `z`.

jakevdp: Wait, but here in the next line the comment calls out the fact that a different type is used. So either (1) we keep it as-is with the dtype promotion standard, or (2) we change the test so that it's not checking for a change in `len(side)`.

I think (2) is a bad idea because it removes test coverage. I'm going to stick with (1), as it currently is.

jakevdp:
Requested changes done; please take a look.

hawkinsp (approved):

PR #11104, opened 2022-06-14
[x64] make scipy_ndimage_test compatible with strict dtype promotion

jakevdp:

### DIFF
--- a/tests/scipy_ndimage_test.py
+++ b/tests/scipy_ndimage_test.py
@@ -96,12 +96,14 @@ class NdimageTest(jtu.JaxTestCase):
     impl_fun = (osp_ndimage.map_coordinates if impl == "original"
                 else _fixed_ref_map_coordinates)
     osp_op = lambda x, c: impl_fun(x, c, order=order, mode=mode, cval=cval)
-    if dtype in float_dtypes:
-      epsilon = max(dtypes.finfo(dtypes.canonicalize_dtype(d)).eps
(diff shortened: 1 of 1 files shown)


tests/scipy_ndimage_test.py
```diff
@@ -96,12 +103,14 @@ def args_maker():
...
                 else _fixed_ref_map_coordinates)
     osp_op = lambda x, c: impl_fun(x, c, order=order, mode=mode, cval=cval)
-    if dtype in float_dtypes:
-      epsilon = max(dtypes.finfo(dtypes.canonicalize_dtype(d)).eps
-                     for d in [dtype, coords_dtype])
-      self._CheckAgainstNumpy(osp_op, lsp_op, args_maker, tol=100*epsilon)
-    else:
-      self._CheckAgainstNumpy(osp_op, lsp_op, args_maker, tol=0)
+
+    with _strict_promotion_if_dtypes_match([dtype, int if round else coords_dtype]):
```
hawkinsp, line 107: Where exactly does the dtype promotion happen here? Why is the fix not to avoid it somewhere closer to its cause?

jakevdp: It happens in line 97 above: the test seems to be explicitly testing what happens if `coords` are converted to an int, so enabling promotion seemed appropriate here rather than changing the test so that case no longer has coverage.

hawkinsp (approved):

PR #11115, opened 2022-06-15
Enable bitwise reductions for integer dtypes in `lax.reduce`

nicholasjng:
Follow-up of #11043.

This commit extends the existing bitwise reduction primitives to work with all non-floating dtypes, including signed and unsigned integers. Tests against numpy were added, in similar style to existing test coverage, to assert the correct results for these new reductions.

Secondly, existing test coverage for `lax.reduce` was extended to check for the correct primitive being produced in the resulting jaxpr for all `lax.reduce` operations.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1016,21 +1016,29 @@ def _get_monoid_reducer(monoid_op: Callable,
   aval = core.get_aval(x)
   dtype = _dtype(x)
   if (type(aval) is ConcreteArray) and aval.shape == ():
+    # allow bitwise reductions for boolean and integer types
+    _is_intlike = dtype == np.bool_ or dtypes.issubdtype(dtype, np.integer)
     if monoid_op is add:
       return np.equal(aval.val, 0) and partial(_reduce_sum)
     elif monoid_op is mul:
       return np.equal(aval.val, 1) and _reduce_prod
-    elif monoid_op is bitwise_or and dtype == np.bool_:
-      return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_or
-    elif monoid_op is bitwise_and and dtype == np.bool_:
-      return np.equal(aval.val, _get_min_identity(dtype)) and _reduce_and
-    elif monoid_op is bitwise_xor and dtype == np.bool_:
-      return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_xor
+    elif monoid_op is bitwise_or and _is_intlike:
+      return np.equal(aval.val, _get_bitwise_or_identity(dtype)) and _reduce_or
+    elif monoid_op is bitwise_and and _is_intlike:
+      return np.equal(aval.val, _get_bitwise_and_identity(dtype)) and _reduce_and
+    elif monoid_op is bitwise_xor and _is_intlike:
+      return np.equal(aval.val, _get_bitwise_or_identity(dtype)) and _reduce_xor
     elif monoid_op is max:
       return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_max
     elif monoid_op is min:
       return np.equal(aval.val, _get_min_identity(dtype)) and _reduce_min
   return None
+
+def _get_bitwise_and_identity(dtype: DType) -> Array:
+  return np.array(-1, dtype)
+
+def _get_bitwise_or_identity(dtype: DType) -> Array:
+  return np.array(0, dtype)
 
 def _get_max_identity(dtype: DType) -> Array:
   if dtypes.issubdtype(dtype, np.inexact):
@@ -3761,25 +3769,24 @@ mlir.register_lowering(argmax_p, mlir.cache_lowering(mlir.lower_fun(
 
 
 def _reduce_logical_shape_rule(operand, *, axes):
-  if operand.dtype != np.bool_:
-    msg = "logical reduction requires operand dtype bool, got {}."
-    raise TypeError(msg.format(operand.dtype))
+  if operand.dtype != np.bool_ and not np.issubdtype(operand.dtype, np.integer):
+    raise TypeError(f"logical reduction requires operand dtype bool or int, got {operand.dtype}.")
   return tuple(np.delete(operand.shape, axes))
 
 reduce_or_p = standard_primitive(
-    _reduce_logical_shape_rule, _fixed_dtype(np.bool_), 'reduce_or',
+    _reduce_logical_shape_rule, _input_dtype, 'reduce_or',
     weak_type_rule=_strip_weak_type)
 batching.defreducer(reduce_or_p)
 
 
 reduce_and_p = standard_primitive(
-    _reduce_logical_shape_rule, _fixed_dtype(np.bool_), 'reduce_and',
+    _reduce_logical_shape_rule, _input_dtype, 'reduce_and',
     weak_type_rule=_strip_weak_type)
 batching.defreducer(reduce_and_p)
 
 
 reduce_xor_p = standard_primitive(
-    _reduce_logical_shape_rule, _fixed_dtype(np.bool_), 'reduce_xor',
+    _reduce_logical_shape_rule, _input_dtype, 'reduce_xor',
     weak_type_rule=_strip_weak_type)
 batching.defreducer(reduce_xor_p)
 
@@ -3802,11 +3809,11 @@ mlir.register_lowering(reduce_sum_p, partial(_unary_reduce_lower, mhlo.AddOp,
 mlir.register_lowering(reduce_prod_p, partial(_unary_reduce_lower, mhlo.MulOp,
                                           lambda dtype: np.array(1, dtype)))
 mlir.register_lowering(reduce_or_p, partial(_unary_reduce_lower, mhlo.OrOp,
-                                         lambda dtype: np.array(False, dtype)))
+                                          _get_bitwise_or_identity))
 mlir.register_lowering(reduce_and_p, partial(_unary_reduce_lower, mhlo.AndOp,
-                                          lambda dtype: np.array(True, dtype)))
+                                          _get_bitwise_and_identity))
 mlir.register_lowering(reduce_xor_p, partial(_unary_reduce_lower, mhlo.XorOp,
-                                          lambda dtype: np.array(False, dtype)))
+                                          _get_bitwise_or_identity))
 mlir.register_lowering(reduce_min_p, partial(_unary_reduce_lower, mlir.min_mhlo,
                                          _get_min_identity))
 mlir.register_lowering(reduce_max_p, partial(_unary_reduce_lower, mlir.max_mhlo,
--- a/tests/lax_test.py
+++ b/tests/lax_test.py
@@ -183,6 +183,27 @@ LAX_OPS = [
     op_record("lt", 2, default_dtypes, jtu.rand_small),
 ]
 
+ReducerOpRecord = collections.namedtuple(
+  "ReducerOpRecord", ["op", "reference_op", "init_val", "dtypes", "primitive"]
+)
+
+LAX_REDUCE_OPS = [
+  ReducerOpRecord(lax.add, np.add, 0, default_dtypes, lax.reduce_sum_p),
+  ReducerOpRecord(lax.mul, np.multiply, 1, default_dtypes, lax.reduce_prod_p),
+  ReducerOpRecord(lax.max, np.maximum, 0, uint_dtypes + bool_dtypes, lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, -np.inf, float_dtypes, lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, dtypes.iinfo(np.int32).min, [np.int32], lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, dtypes.iinfo(np.int64).min, [np.int64], lax.reduce_max_p),
+  ReducerOpRecord(lax.min, np.minimum, np.inf, float_dtypes, lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.int32).max, [np.int32], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.int64).max, [np.int64], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.uint32).max, [np.uint32], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.uint64).max, [np.uint64], lax.reduce_min_p),
+  ReducerOpRecord(lax.bitwise_and, np.bitwise_and, -1, int_dtypes + uint_dtypes + bool_dtypes, lax.reduce_and_p),
+  ReducerOpRecord(lax.bitwise_or, np.bitwise_or, 0, int_dtypes + uint_dtypes + bool_dtypes, lax.reduce_or_p),
+  ReducerOpRecord(lax.bitwise_xor, np.bitwise_xor, 0, int_dtypes + uint_dtypes + bool_dtypes, lax.reduce_xor_p),
+]
+
 
 class LaxTest(jtu.JaxTestCase):
   """Numerical tests for LAX operations."""
@@ -1715,28 +1736,28 @@ class LaxTest(jtu.JaxTestCase):
 
   @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_op={}_inshape={}_reducedims={}_initval={}"
-       .format(op.__name__, jtu.format_shape_dtype_string(shape, dtype), dims,
-               init_val),
-       "op": op, "init_val": init_val, "shape": shape, "dtype": dtype, "dims": dims}
-      for init_val, op, types in [
-          (0, lax.add, default_dtypes),
-          (1, lax.mul, default_dtypes),
-          (0, lax.max, all_dtypes), # non-monoidal
-          (-np.inf, lax.max, float_dtypes),
-          (dtypes.iinfo(np.int32).min, lax.max, [np.int32]),
-          (dtypes.iinfo(np.int64).min, lax.max, [np.int64]),
-          (np.inf, lax.min, float_dtypes),
-          (dtypes.iinfo(np.int32).max, lax.min, [np.int32]),
-          (dtypes.iinfo(np.int64).max, lax.min, [np.int64]),
-          (dtypes.iinfo(np.uint32).max, lax.min, [np.uint32]),
-          (dtypes.iinfo(np.uint64).max, lax.min, [np.uint64]),
-      ]
-      for dtype in types
+       .format(rec.op.__name__, jtu.format_shape_dtype_string(shape, dtype), dims,
+               rec.init_val),
+       "op": rec.op, "reference_op": rec.reference_op, "init_val": rec.init_val,
+       "shape": shape, "dtype": dtype, "dims": dims, "primitive": rec.primitive}
+      for rec in LAX_REDUCE_OPS
+      for dtype in rec.dtypes
       for shape, dims in [
           [(3, 4, 5), (0,)], [(3, 4, 5), (1, 2)],
           [(3, 4, 5), (0, 2)], [(3, 4, 5), (0, 1, 2)]
       ]))
-  def testReduce(self, op, init_val, shape, dtype, dims):
+  def testReduce(self, op, reference_op, init_val, shape, dtype, dims, primitive):
+    if not config.x64_enabled and dtype in (np.float64, np.int64, np.uint64):
+      raise SkipTest("x64 mode is disabled.")
+    def reference_fun(operand):
+      if hasattr(reference_op, "reduce"):
+        initial = np.array(init_val, dtype=dtype)
+        result = reference_op.reduce(operand, axis=dims, initial=initial)
+      else:
+        result = reference_op(operand, axis=dims)
+
+      return result.astype(dtype)
+
     rng_factory = (jtu.rand_default if dtypes.issubdtype(dtype, np.integer)
                    else jtu.rand_small)
     rng = rng_factory(self.rng())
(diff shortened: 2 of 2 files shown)


nicholasjng:
I am seeing a number of test failures locally using `JAX_NUM_GENERATED_CASES=100`, but I wanted to post this to discuss. To me, it looks like the failures can be categorized into three different types:

1) numpy vs jax dtype mismatches
2) numpy vs jax result mismatches for a given tolerance (for some uint{8,16} tests, too)
3) Wrong primitive being produced in the lowerings for some `lax.{min,max}` tests.

I would be happy about some feedback!

nicholasjng:
Would you like to review this @jakevdp? Since you handled the previous PR as well

jax/_src/lax/lax.py
```diff
@@ -3605,25 +3607,25 @@ def reducer_fn(op_val_index, acc_val_index):
 
 
 def _reduce_logical_shape_rule(operand, *, axes):
-  if operand.dtype != np.bool_:
-    msg = "logical reduction requires operand dtype bool, got {}."
+  if operand.dtype != np.bool_ and not np.issubdtype(operand.dtype, np.integer):
+    msg = "logical reduction requires operand dtype bool or int, got {}."
     raise TypeError(msg.format(operand.dtype))
```
jakevdp, line 3612: Since we're touching this code, we may as well update it to modern string formatting:
```python
  if operand.dtype != np.bool_ and not np.issubdtype(operand.dtype, np.integer):
    raise TypeError(f"logical reduction requires operand dtype bool or int, got {operand.dtype}.")
```

tests/lax_test.py
```diff
@@ -1714,26 +1714,42 @@ def testTransposeAgainstNumpy(self, shape, dtype, perm):
...
+      np_op_name = op.__name__
+      if op.__name__ == "mul":
+        np_op_name = "multiply"
+
+      np_op = getattr(np, np_op_name)
+
+      if hasattr(np_op, "reduce"):
+        return np_op.reduce(operand, axis=dims, initial=init_value)
+      else:
+        return np_op(operand, axis=dims)
```
jakevdp, line 1751: I'm a bit concerned about how this test is set up – there is some amount of "action at a distance" that makes the parameterization feel a bit brittle – for example, this reference implementation doesn't always reference the `init_val`, which means there's additional, unmentioned requirements on the input parameterization for the test to be correct. It also makes it hard to read through the test and understand what's going on.

I'd suggest using a class-record based test parameterization and adding the appropriate numpy function to those entries (I think since `init_val` is the obvious choice in each case, you don't even need to pass it to the reference function)

tests/lax_test.py
```diff
@@ -1714,26 +1714,42 @@ def testTransposeAgainstNumpy(self, shape, dtype, perm):
...
-          (np.inf, lax.min, float_dtypes),
-          (dtypes.iinfo(np.int32).max, lax.min, [np.int32]),
-          (dtypes.iinfo(np.int64).max, lax.min, [np.int64]),
-          (dtypes.iinfo(np.uint32).max, lax.min, [np.uint32]),
-          (dtypes.iinfo(np.uint64).max, lax.min, [np.uint64]),
+       "op": op, "init_val": init_val, "shape": shape, "dtype": dtype, "dims": dims, "prim": prim}
+      for init_val, op, types, prim in [
+          (0, lax.add, default_dtypes, lax.reduce_sum_p),
+          (1, lax.mul, default_dtypes, lax.reduce_prod_p),
+          (0, lax.max, float_dtypes + complex_dtypes + int_dtypes, lax.reduce_p), # non-monoidal
```
jakevdp, line 1721: Since `0` is not an identity of `lax.max`, this test case produces undefined values. I'd probably remove it.

tests/lax_test.py
```diff
@@ -1714,26 +1714,42 @@ def testTransposeAgainstNumpy(self, shape, dtype, perm):
...
+          (dtypes.iinfo(np.int32).min, lax.max, [np.int32], lax.reduce_max_p),
+          (dtypes.iinfo(np.int64).min, lax.max, [np.int64], lax.reduce_max_p),
+          (np.inf, lax.min, float_dtypes, lax.reduce_min_p),
+          (dtypes.iinfo(np.int32).max, lax.min, [np.int32], lax.reduce_min_p),
+          (dtypes.iinfo(np.int64).max, lax.min, [np.int64], lax.reduce_min_p),
+          (dtypes.iinfo(np.uint32).max, lax.min, [np.uint32], lax.reduce_min_p),
+          (dtypes.iinfo(np.uint64).max, lax.min, [np.uint64], lax.reduce_min_p),
+          (dtypes.iinfo(np.uint32).max, lax.bitwise_and, [np.uint32], lax.reduce_and_p),
+          (dtypes.iinfo(np.uint64).min, lax.bitwise_or, [np.uint64], lax.reduce_or_p),
+          (0, lax.bitwise_xor, uint_dtypes, lax.reduce_xor_p)
```
jakevdp, line 1733: This parameterization is starting to look a bit daunting. Perhaps we should define a `ReducerOpRecord` class to help organize which cases we want to test?

jakevdp (changes requested):
Thanks for looking into this! Sorry for the delay, I was out of the office for a bit.

tests/lax_test.py
nicholasjng: That test already existed in this configuration, I only amended the expected primitive. Should I still remove it?

tests/lax_test.py
nicholasjng: Fine by me, is there a pre-existing op record implementation in JAX testing that I should use?

nicholasjng:
Thanks for the initial review! I incorporated your changes in the latest commit, I hope I understood everything correctly.

tests/lax_test.py
```diff
@@ -183,6 +183,28 @@ def op_record(op, nargs, dtypes, rng_factory, tol=None):
...
+  ReducerOpRecord(lax.max, np.maximum, 0, uint_dtypes + bool_dtypes, lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, -np.inf, float_dtypes, lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, dtypes.iinfo(np.int32).min, [np.int32], lax.reduce_max_p),
+  ReducerOpRecord(lax.max, np.maximum, dtypes.iinfo(np.int64).min, [np.int64], lax.reduce_max_p),
+  ReducerOpRecord(lax.min, np.minimum, np.inf, float_dtypes, lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.int32).max, [np.int32], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.int64).max, [np.int64], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.uint32).max, [np.uint32], lax.reduce_min_p),
+  ReducerOpRecord(lax.min, np.minimum, dtypes.iinfo(np.uint64).max, [np.uint64], lax.reduce_min_p),
+  ReducerOpRecord(lax.bitwise_and, np.bitwise_and, dtypes.iinfo(np.uint32).max, [np.uint32], lax.reduce_and_p),
```
jakevdp, line 203: Could we use `-1` and `uint_dtypes` here? (since `uint(-1)` returns the maximum value)?

tests/lax_test.py
```diff
@@ -183,6 +183,28 @@ def op_record(op, nargs, dtypes, rng_factory, tol=None):
...
+  ReducerOpRecord(lax.bitwise_or, np.bitwise_or, dtypes.iinfo(np.uint64).min, [np.uint64], lax.reduce_or_p),
```
jakevdp, line 204: Could we use `0` and `uint_dtypes` here?

tests/lax_test.py
```diff
@@ -183,6 +183,28 @@ def op_record(op, nargs, dtypes, rng_factory, tol=None):
...
+  ReducerOpRecord(lax.bitwise_xor, np.bitwise_xor, 0, uint_dtypes, lax.reduce_xor_p),
+]
```
jakevdp, line 206: Can we add tests for signed integer types?

jakevdp (changes requested):

nicholasjng:
Thanks for the follow-up, I will get on it asap. What do you think about merging the bitwise test from the last PR with this one, while converting it to the `ReducerOpRecord` implementation? That way we can deduplicate some code.

jakevdp:
> What do you think about merging the bitwise test from the last PR with this one, while converting it to the ReducerOpRecord implementation? That way we can deduplicate some code.

That seems like a good plan.

nicholasjng:
I pushed your suggested changes again. It seems that now we are getting the dtype mismatch errors that I saw locally. Do you think this is related to type promotion, or something else?

tests/lax_test.py
nicholasjng: Done. The initial value is not part of the reference function signature anymore, and the parametrization has been changed to iterate over the newly created `ReducerOpRecord`s.

jakevdp:
It looks like these are dtype mismatches between the corresponding jax and numpy functions. The issue comes from this:
```python
In [1]: from jax import lax                                                                                                                                                                   

In [2]: import numpy as np                                                                                                                                                                    

In [3]: x = np.arange(10, dtype='int16')                                                                                                                                                      

In [4]: np.sum(x).dtype                                                                                                                                                                       
Out[4]: dtype('int64')

In [5]: lax.reduce([x], [0], lax.add, (0,)).dtype                                                                                                                                             
Out[5]: dtype('int16')
```
I'd handle this by ensuring that `reference_fun` returns the expected dtype.

jax/_src/lax/lax.py
```diff
@@ -1016,15 +1016,17 @@ def _get_monoid_reducer(monoid_op: Callable,
...
       return np.equal(aval.val, 1) and _reduce_prod
-    elif monoid_op is bitwise_or and dtype == np.bool_:
+    elif monoid_op is bitwise_or and _is_intlike:
       return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_or
-    elif monoid_op is bitwise_and and dtype == np.bool_:
+    elif monoid_op is bitwise_and and _is_intlike:
       return np.equal(aval.val, _get_min_identity(dtype)) and _reduce_and
-    elif monoid_op is bitwise_xor and dtype == np.bool_:
+    elif monoid_op is bitwise_xor and _is_intlike:
       return np.equal(aval.val, _get_max_identity(dtype)) and _reduce_xor
```
jakevdp, line 1030: I don't think these are the correct identities for bitwise operations in the case of signed integers. For `bitwise_or` and `bitwise_xor` the identity should be the value where all bits are zero, and for `bitwise_and` the identity should be the value where all bits are `1`. For `bool` and `uint` this corresponds to `get_min_identity` and `get_max_identity`, but for signed ints using a two's complement representation I think you need `0` and `-1`

nicholasjng: Great catch! I think the potential fix could be either of:
1. Adding more statements to that branch specializing the value compared in `np.equal` depending on whether the integer is signed or unsigned (or it is a boolean), or
2. Explicitly check equality against `dtype(0)` and `dtype(-1)`, which should evaluate to the minimum/maximum identities of uints and bools, and the bitwise identities for signed integers, in which case we would get all three cases with a single line.

What do you think? The second might be elegant, but at the same time a little less readable.

jakevdp: Yeah, seems like using `dtype(0)` for `bitwise_or`/`bitwise_xor` and `dtype(-1)` for `bitwise_and` would be the cleanest approach.

nicholasjng: I am getting errors locally of the form `numpy.dtype object is not callable`. Do you have an idea on how I could work around those? Seems like a simple "cast" of the value like I suggested does not work.

jakevdp: Use `np.array(-1, dtype=dtype)`

jakevdp (approved):
Looks great!

jakevdp:
Seems like mypy is unhappy, though that's strange because it's due to the same return pattern used elsewhere in the function. But in general, I find mypy tends to respond to butterflies flapping their wings in Brazil...

nicholasjng:
Yeah, I saw. It seems to be because unlike elsewhere, the `np.equal` aval check is called on an array as opposed to a scalar. Calling `array.item()` on each of these turns everything green again.

By the way, locally with `JAX_NUM_GENERATED_CASES=100`:
```
======================================================================
FAIL: testReduce_op=bitwise_and_inshape=int8[3,4,5]_reducedims=(0,)_initval=-1 (__main__.LaxTest)
LaxTest.testReduce_op=bitwise_and_inshape=int8[3,4,5]_reducedims=(0,)_initval=-1
testReduce_op=bitwise_and_inshape=int8[3,4,5]_reducedims=(0,)_initval=-1(op=<function bitwise_and at 0x12f3f6290>, reference_op=<ufunc 'bitwise_and'>, init_val=-1, shape=(3, 4, 5), dtype=<class 'numpy.int8'>, dims=(0,), primitive=reduce_and)
----------------------------------------------------------------------
(Traceback)

AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 4 / 20 (20%)
Max absolute difference: 6
Max relative difference: 1.
 x: array([[0, 1, 0, 0, 0],
       [0, 0, 1, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0]], dtype=int8)
 y: array([[0, 7, 0, 0, 0],
       [0, 0, 3, 0, 0],
       [2, 0, 0, 0, 0],
       [4, 0, 0, 0, 0]], dtype=int8)

======================================================================
FAIL: testReduce_op=bitwise_and_inshape=uint16[3,4,5]_reducedims=(0,)_initval=-1 (__main__.LaxTest)
LaxTest.testReduce_op=bitwise_and_inshape=uint16[3,4,5]_reducedims=(0,)_initval=-1
testReduce_op=bitwise_and_inshape=uint16[3,4,5]_reducedims=(0,)_initval=-1(op=<function bitwise_and at 0x12f3f6290>, reference_op=<ufunc 'bitwise_and'>, init_val=-1, shape=(3, 4, 5), dtype=<class 'numpy.uint16'>, dims=(0,), primitive=reduce_and)
----------------------------------------------------------------------
[...]
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 3 / 20 (15%)
Max absolute difference: 65534
Max relative difference: 32767.
 x: array([[0, 0, 1, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0]], dtype=uint16)
 y: array([[0, 2, 1, 0, 0],
       [0, 0, 0, 0, 2],
       [0, 0, 2, 0, 0],
       [0, 0, 0, 0, 0]], dtype=uint16)

======================================================================
FAIL: testReduce_op=bitwise_and_inshape=uint32[3,4,5]_reducedims=(0,)_initval=-1 (__main__.LaxTest)
LaxTest.testReduce_op=bitwise_and_inshape=uint32[3,4,5]_reducedims=(0,)_initval=-1
testReduce_op=bitwise_and_inshape=uint32[3,4,5]_reducedims=(0,)_initval=-1(op=<function bitwise_and at 0x12f3f6290>, reference_op=<ufunc 'bitwise_and'>, init_val=-1, shape=(3, 4, 5), dtype=<class 'numpy.uint32'>, dims=(0,), primitive=reduce_and)
----------------------------------------------------------------------
[...]
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2 / 20 (10%)
Max absolute difference: 4294967294
Max relative difference: 2.14748365e+09
 x: array([[0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 1, 0, 0],
       [0, 0, 0, 0, 0]], dtype=uint32)
 y: array([[0, 2, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 1, 0, 0],
       [0, 0, 0, 2, 0]], dtype=uint32)

======================================================================
FAIL: testReduce_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1 (__main__.LaxTest)
LaxTest.testReduce_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1
testReduce_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1(op=<function bitwise_and at 0x12f3f6290>, reference_op=<ufunc 'bitwise_and'>, init_val=-1, shape=(3, 4, 5), dtype=<class 'numpy.uint8'>, dims=(0,), primitive=reduce_and)
----------------------------------------------------------------------
[...]
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 20 (5%)
Max absolute difference: 254
Max relative difference: 127.
 x: array([[0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0]], dtype=uint8)
 y: array([[0, 0, 0, 0, 0],
       [0, 0, 0, 2, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0]], dtype=uint8)
```

Do you think those are flaky tests, or is there something deeper going on? (There are some other failures, but those are because of disabled x64 mode.)

jakevdp:
Thanks - regarding those failures, I don't see any obvious reason why they'd be flakes. I would suggest looking into it to try to understand where the difference is coming from – I suspect it's a real bug.

nicholasjng:
Alright. I will attempt to work more on it tomorrow. I ran all generated tests, and failures can be summarized by:

- `lax.bitwise_and` with `dims=(0,)` and `shape=[3,4,5]`, for dtypes `uint[8,16,32]` and `int[8,16,32]`. Reason is the disagreement of some values within test tolerance.
- `lax.{min,max}` for all shapes and dims, for dtypes `int64, uint64`. Reason is that the lowered primitive is `reduce_p`, probably because x64 mode was not enabled.

The first batch of errors happening, for reduction dimension (0,) only, but for all integer dtypes, is weird to me. I will keep investigating.

nicholasjng:
I opted to skip those tests for now if x64 mode is disabled, as that was done in other lax tests (and the `canonicalize_dtype` route has not worked for me so far).

For the `bitwise_and` errors, the traceback, unfortunately left out previously, points to the check against the JIT-compiled version of `lax.reduce`. So it might be possible that there is something going wrong in XLA?

```
======================================================================
FAIL: testReduceBare_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1 (__main__.LaxTest)
LaxTest.testReduceBare_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1
testReduceBare_op=bitwise_and_inshape=uint8[3,4,5]_reducedims=(0,)_initval=-1(op=<function bitwise_and at 0x13f89e290>, reference_op=<ufunc 'bitwise_and'>, init_val=-1, shape=(3, 4, 5), dtype=<class 'numpy.uint8'>, dims=(0,), primitive=reduce_and)
----------------------------------------------------------------------
  File "/Users/nicholasjunge/Workspaces/python/jax/venv/lib/python3.10/site-packages/absl/testing/parameterized.py", line 312, in bound_param_test

  File "/Users/nicholasjunge/Workspaces/python/jax/tests/lax_test.py", line 1767, in testReduceBare
    self._CompileAndCheck(fun, args_maker)   # <--- fun is lambda operand, init_val: lax.reduce(operand, init_val, op, dims)

  File "/Users/nicholasjunge/Workspaces/python/jax/jax/_src/test_util.py", line 838, in _CompileAndCheck
    self.assertAllClose(python_ans, monitored_ans, check_dtypes=check_dtypes,
  File "/Users/nicholasjunge/Workspaces/python/jax/jax/_src/test_util.py", line 787, in assertAllClose
    self.assertArraysAllClose(x, y, check_dtypes=False, atol=atol, rtol=rtol,
  File "/Users/nicholasjunge/Workspaces/python/jax/jax/_src/test_util.py", line 752, in assertArraysAllClose
  File "/Users/nicholasjunge/Workspaces/python/jax/jax/_src/public_test_util.py", line 96, in _assert_numpy_allclose
  File "/Users/nicholasjunge/Workspaces/python/jax/venv/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 1530, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/Users/nicholasjunge/Workspaces/python/jax/venv/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
  AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 20 (5%)
Max absolute difference: 254
Max relative difference: 127.
 x: array([[0, 1, 0, 0, 0],
       [0, 0, 0, 0, 1],
       [0, 0, 0, 1, 0],
       [0, 0, 0, 0, 0]], dtype=uint8)
 y: array([[0, 1, 0, 0, 0],
       [0, 0, 0, 0, 1],
       [0, 2, 0, 1, 0],
       [0, 0, 0, 0, 0]], dtype=uint8)
```

tests/lax_test.py
```diff
@@ -1715,28 +1736,28 @@ def testTransposeAgainstNumpy(self, shape, dtype, perm):
...
+      raise SkipTest("x64 mode is disabled.")
+    def reference_fun(operand):
+      if hasattr(reference_op, "reduce"):
+        initial = np.array(init_val, dtype=dtype)
+        result = reference_op.reduce(operand, axis=dims, initial=initial)
+      else:
+        result = reference_op(operand, axis=dims)
+
+      return result.astype(dtype)
+
```
jakevdp, line 1760: One issue just below: because `init_val` is passed via `args_maker`, within `CompileAndCheck` it will be traced and therefore the new reducer primitives will not come into play

nicholasjng: Thanks, that makes sense. But how is only `bitwise_and` affected?

jakevdp: Perhaps something about the monoid assumption is violated?

nicholasjng: Good idea, but I think would cause tests to fail for all shapes when they do not:

```
[ RUN      ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0, 1, 2)_initval=-1
[       OK ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0, 1, 2)_initval=-1
[ RUN      ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0, 2)_initval=-1
[       OK ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0, 2)_initval=-1
[ RUN      ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0,)_initval=-1
[  FAILED  ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(0,)_initval=-1
[ RUN      ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(1, 2)_initval=-1
[       OK ] LaxTest.testReduceBare_op=bitwise_and_inshape=int16[3,4,5]_reducedims=(1, 2)_initval=-1
```

pytest logs are similar for int8/32 as well as uint types.

jakevdp: Not necessarily, `lax.reduce` will sometimes work if the arguments do not meet the monoid condition. It's just that it's not guaranteed to return the correct output unless those conditions are met.

nicholasjng: Alright. I played with the setup a little, and the number of failing tests changes if you change the reduction dimension to `(1,)` or `(2,)`, for example. 

Also, all tests pass when the first call to `_CompileAndCheck` and the call to `_CheckAgainstNumpy` are commented out - the check of the JIT compilation for the operand-only lambda passes. So it might indeed have to do with the lambda's signature. Do you have an idea on how I could proceed?

jakevdp: It's not obvious to me where the bug is... I would probably put together a stand-alone reproduction of the error, and start using `jax.make_jaxpr` to inspect the operations that are being generated.

nicholasjng:
I updated the lowering registration. Sorry for wasting your time there, I should have caught that.

Bar two possible improvements (read polishing), I think this is ready:
- Substituting `np.array(False, dtype)` to `np.array(0, dtype)` in the `or/xor` lowerings to increase code verbosity for integer types
- Adding `_get_{or,and,xor}_identity` functions similar to the existing ones that return the identity element for these ops, again for better readability.

jax/_src/lax/lax.py
```diff
@@ -3804,7 +3805,7 @@ def _unary_reduce_lower(reducer, unit_factory, ctx, x, *, axes):
 mlir.register_lowering(reduce_or_p, partial(_unary_reduce_lower, mhlo.OrOp,
                                          lambda dtype: np.array(False, dtype)))
```
jakevdp, line 3806: This is minor, but here and below, maybe use `np.array(0, dtype)` to match the values above.

Alternatively, it may be clearer in the long run to use a helper function like `def _bitwise_or_identity(dtype): return np.array(0, dtype)` and also call this when you specify the identities above. What do you think?

nicholasjng: Fixed.

jakevdp (approved):
Looks great! Thanks for seeing this all the way through 😁

nicholasjng:
My pleasure, thank you for the help throughout!

PR #11116, opened 2022-06-15
[dynamic-shapes] initial support for dynamic shape typechecks

mattjj:
After this PR, jaxpr typechecking works with all existing dynamic shape tests. I was working on dynamic shape AD and realized I could make faster progress with some reliable jaxpr checks!

Summary of changes:
* Changed the logic in `_check_jaxpr` so as to check that type annotations are well-formed via the new `check_type` function (e.g. variable occurrences in types are in scope).
* Changed the logic in `_check_jaxpr` to handle dynamic shapes in inputs and outputs of primitive applications. For the input side, we change the signature of some primitive application typecheck rules to take input atoms rather than just types (see also last bullet in this list). That way we can check things like e.g. an `xla_call` application with a `call_jaxpr` with lambda binders like `n:i32[] x:f32[n]` against arguments like `3:i32[] y:f32[3]` (correct!) or `3:i32[] y:f32[4]` (error!) or `m:i32[] y:f32[k]` (error!). For the output side, we handle primitive application typechecks returning lists of abstract values in which the shapes may contain `InDBIdx` / `OutDBIdx` axis sizes (representing e.g. dependent tuples), in which case to check against the corresponding let-binder types (which have variables in them) we perform a substitution using `substitute_vars_in_output_ty`.
* Changed the signature of `_check_jaxpr` not to take as argument avals to which the jaxpr is to be applied; we check the called jaxpr's lambda-binder types are consistent with the arguments in the rule `_check_call` rather than in the recursive call to `_check_jaxpr`, which itself now just checks a jaxpr for internal consistency.
* Changed the type signature of custom typechecks to support checking calls and broadcasts, basically to something like `Primitive -> [Atom] -> Params -> OutType` where 
```
OutType = [AvalsWithInOutDBIdx]
AvalsWithInOutDBIdx = ShapedArray [AxisSizeOrDBIdx] DType
AxisSizeOrDBIdx = Lit Int | InDBIdx Int | OutDBIdx Int
```

Minor stuff:
* Moved `DBIdx`, `InDBIdx`, and `OutDBIdx` from partial_eval.py to core.py.
* Some typecheck rules returned `None` as the output type, in which case checking the type of the primitive application against the corresponding let-binders would just be skipped. This PR removes that, and adapts the typecheck rules which were returning `None` to return their correct output types.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import collections
 from collections import namedtuple
 from contextlib import contextmanager
+from dataclasses import dataclass
 import functools
 from functools import partial, partialmethod, total_ordering
 import gc
@@ -208,6 +209,9 @@ class JaxprEqn(NamedTuple):
 
 def new_jaxpr_eqn(invars, outvars, primitive, params, effects, source_info=None):
   source_info = source_info or source_info_util.new_source_info()
+  if config.jax_enable_checks:
+    assert all(isinstance(x, (Var, Literal)) for x in  invars)
+    assert all(isinstance(v,  Var)           for v in outvars)
   return JaxprEqn(invars, outvars, primitive, params, effects, source_info)
 
 
@@ -1032,7 +1036,6 @@ def find_top_trace(xs) -> Trace:
 
 # -------------------- abstract values --------------------
 
-
 class AbstractValue:
   __slots__: List[str] = []
 
@@ -1061,6 +1064,33 @@ class AbstractValue:
   def str_short(self, short_dtypes=False):
     return str(self)
 
+
+# For type signatures involving dynamic shapes, we use lists of abstract values
+# which may contain (reverse) de Bruijn indices in their shapes.
+class DBIdx(NamedTuple):
+  val: int
+
+@dataclass(frozen=True)
+class InDBIdx:
+  val: int
+
+@dataclass(frozen=True)
+class OutDBIdx:
+  val: int
+
+# For annotating input types of callables (i.e. linear_util.WrappedFuns), we use
+# a sequence of pairs where the first element of each pair is an AbstractValue
+# (possibly containing DBIdx instances in its shape) and the second is a boolean
+# indicating whether that argument is explicit (i.e. passed to the callable).
+InputType = Tuple[Tuple[AbstractValue, bool], ...]  # DBIdx in shapes
+
+# For annotating jaxpr output types, we use a sequence of pairs where the first
+# element of each pair is an AbstractValue (possibly containing InDBIdx and/or
+# OutDBIdx instances in its shape) and the second is a boolean indicating
+# whether that argument is explicit (i.e. returned by the callable).
+OutputType = Tuple[Tuple[AbstractValue, bool], ...]  # InDBIdx / OutDBIdx shapes
+
+
 class Bot(AbstractValue): pass
 
 bot = Bot()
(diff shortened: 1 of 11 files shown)


jax/core.py
```diff
@@ -2243,51 +2283,73 @@ def ctx_factory():
...
+      if not typecheck(x.aval, x.val):
         ctx, _ = ctx_factory()
-        raise JaxprTypeError(f"Variable '{pp_var(v, ctx)}' not defined")
-      return env[v]
+        raise JaxprTypeError(
+            f"Literal value {x.val} does not match its type annotation "
+            f"{pp_aval(x.aval, ctx)}")
+      return x
+    else:
+      assert False, "syntactically invalid jaxpr"
```
froystig, line 2309:
```suggestion
      raise JaxprTypeError(f"syntactically invalid jaxpr: attempt to read {x}")
```

jax/core.py
```diff
@@ -2316,27 +2426,41 @@ def check_eqn(prim, in_avals, params):
     out_avals = [out_avals]
   return out_avals, effects
 
-def check_call(ctx_factory, prim, in_avals, params):
+def _check_call(ctx_factory, prim, in_atoms, params):
```
froystig, line 2429: Might be a nit, but why did this pick up a leading underscore?

froystig (approved):
Nice work!

jax/core.py
mattjj: I wanted to distinguish jaxpr syntax errors from jaxpr type errors, and assume `check_jaxpr` is only given syntactically valid jaxprs (hence the assert). Same below. WDYT?

jax/core.py
mattjj: I wanted to make sure there were no existing callers so I changed the name, and also if the function is just a helper not to be used outside this module having an underscore is a nice hint. WDYT?

jax/core.py
mattjj: Put another way, we can and should have error checks for these in the `__init__` methods of the relevant classes (e.g. for this line the corresponding check would be in `JaxprEqn.__init__` asserting that each element of `invars` is an `Atom`).

froystig: I see. Yes, I agree it makes sense to distinguish.

With that in mind, should we check for syntactic validity at any point as well? If so, would we do syntax checks under `check_jaxpr`, and raise `JaxprSyntaxError` rather than `JaxprTypeError`?

This question of whether to syntax-check, and where, is broader in scope than this PR, so we can proceed with `assert` for now and return to it.

jax/core.py
froystig: No strong opinion either way – just making sure I wasn't missing some new meaning. In this case, I'd update `check_map` to `_check_map` as well to be consistent.

froystig: By the way, if our current `core` were under `_src` (maybe with select symbol exports via `jax.core`), I would slightly prefer no leading underscores.

jax/core.py
mattjj: Well, if we have the checks in the `__init__` methods, then it won't be possible to construct a syntactically invalid jaxpr (without mutation), so we may not need a syntax checker.

froystig: Sure, that'd work.

jax/core.py
mattjj: Done.

PR #11117, opened 2022-06-15
BUG: avoid warning when specifying dtype=complex in X32 mode

jakevdp:
Because of a bug in `lax.py`, currently the following produces a warning when it shouldn't:
```python
>>> import jax.numpy as jnp
>>> jnp.array(1, dtype=complex)
/github/google/jax/jax/_src/numpy/lax_numpy.py:1832: UserWarning: Explicitly requested dtype <class 'complex'> requested in array is not available, and will be truncated to dtype complex64. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  lax_internal._check_user_dtype_supported(dtype, "array")
DeviceArray(1.+0.j, dtype=complex64)
```
This PR fixes the issue and adds a test.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -4535,7 +4535,7 @@ def _abstractify(x):
 
 def _check_user_dtype_supported(dtype, fun_name=None):
   # Avoid using `dtype in [...]` because of numpy dtype equality overloading.
-  if isinstance(dtype, type) and dtype in {bool, int, float, complex}:
+  if isinstance(dtype, type) and dtype in {bool, int, float, builtins.complex}:
     return
   np_dtype = np.dtype(dtype)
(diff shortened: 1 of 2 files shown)


jax/_src/lax/lax.py
```diff
@@ -4535,7 +4535,7 @@ def _abstractify(x):
 
 def _check_user_dtype_supported(dtype, fun_name=None):
   # Avoid using `dtype in [...]` because of numpy dtype equality overloading.
-  if isinstance(dtype, type) and dtype in {bool, int, float, complex}:
+  if isinstance(dtype, type) and dtype in {bool, int, float, builtins.complex}:
```
jakevdp, line 4538: The issue is that the builtin `complex` has been clobbered by `def complex(...)` in this file.

zhangqiaorjc (approved):

PR #11123, opened 2022-06-16
[ROCm] Revert unit test enablement for ROCm (from 8d9f17df19a647577357398c7a…

rsanthanam-amd:
…57ceb6eb6b9f04).

/cc @hawkinsp


rsanthanam-amd:
@hawkinsp Please take a look at this too.  I reverted the decorator in a previous PR but that turned out to be a mistake.  Until I can figure out what is wrong, the decorator should be reinstated.

Thanks!

hawkinsp (approved):

PR #11124, opened 2022-06-16
[x64] make linalg functions & tests compatible with strict dtype promotion

jakevdp:


froystig (approved):
Nice!

PR #11127, opened 2022-06-16
Add `discharge_state` to convert stateful jaxprs into pure jaxprs

sharadmv:
Adds a method that discharges state from jaxprs.

Tracker: #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -14,13 +14,21 @@
 """Module for the `for_loop` primitive."""
 from functools import partial
 
-from typing import Any, List, Tuple
+from typing import Any, Dict, List, Sequence, Tuple
 
 from jax import core
+from jax import lax
+from jax import linear_util as lu
 from jax.interpreters import ad
+from jax.interpreters import partial_eval as pe
 from jax._src import ad_util
 from jax._src import pretty_printer as pp
+from jax._src import util
 
+## JAX utilities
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow/for_loop.py
```diff
@@ -273,3 +281,95 @@ def _swap_transpose(g, ref, x, *idx):
...
+      #    x = dus x i z
+      x, val, *idx = in_vals
+      ans = _dynamic_update_index(x, idx, val + _dynamic_index(x, idx))
+      assert isinstance(eqn.invars[0], core.Var)
+      write(eqn.invars[0], ans)
+    else:
+      # Default primitive rule, similar to `core.eval_jaxpr`. Note that here
+      # we assume any higher-order primitives inside of the jaxpr are *not*
+      # stateful.
+      # TODO(sharadmv): enable discharging state in higher order primitives
```
mattjj, line 362: I think we've set things up so that no higher-order primitives can be stateful, since the only place the state effect is allowed is in the body of for_loop and that discharges it. We should think about releasing the state effect from its current prison someday, but in the meantime I think a code TODO is too narrow than that big design question, so maybe just remove it for now.

mattjj (approved):
Nice!

jax/_src/lax/control_flow/for_loop.py
sharadmv: Done!

PR #11135, opened 2022-06-16
[x64] make host_callback_test compatible with strict dtype promotion

jakevdp:


sharadmv (approved):
Thanks

PR #11137, opened 2022-06-16
BUG: fix incorrect order of operations in line search

jakevdp:
Problem was introduced in #6785

I'm not sure the effect of this on outputs; I discovered it in the process of fixing tests for https://github.com/google/jax/pull/10840.


shoyer (approved):
good catch!

jakevdp:
I had to change the starting point for one test, because it was no longer converging.

PR #11140, opened 2022-06-17
add the 08-pjit.ipynb

maxwillzq:
the ipynb of 08-pjit is missing, while other ipynb files.


maxwillzq:
it is on main branch. will create another one for this PR

PR #11141, opened 2022-06-17
add the 08-pjit.ipynb

maxwillzq:
the ipynb of 08-pjit is missing, while other ipynb files.

### DIFF
--- a/docs/jax-101/08-pjit.ipynb
+++ b/docs/jax-101/08-pjit.ipynb
@@ -1,0 +1,809 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "800c6ef6",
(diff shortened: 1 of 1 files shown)


maxwillzq:
hi @jakevdp  Thanks a lot for your reply.
do you mean we should remove this TODO after adding back the 08-pjit.ipynb ? 
I do not understand why other files have notebook and 08-pjit does not have it.  Any special reason we break this design ?
Thanks

jakevdp:
I'm not sure why the notebook was not added originally, I suspect there was some issue with the doc build.

Regarding the TODO: the issue in the TODO explains the cause of the readthedocs failure, which comes from the fact that `08-pjit.md` is not included in that list.

PR #11146, opened 2022-06-17
Change JAX_PLATFORMS to raise an exception when platform initialization fails

sshahrokhi:
Repurposing --jax_platforms, to not only setting the priorities of which platforms to initialize, but also abort the program if the specified platforms are not initialized correctly. This will let the user make sure what platform the program is running on, without getting confused in between. If not set, still the default behavior is that we switch to CPU if other platforms are not successfully initialized. 

[TESTING] To test this flag, there are two ways to set it. 1) set it in the terminal with `export JAX_PLATFORMS=tpu,cpu` or in a written program like test.py that has this line in it: `config.update('jax_platforms', 'tpu,cpu')`. 

The test.py script could be this: 
```
import jax
from jax._src.config import config

config.update('jax_platforms', 'tpu,cpu')
print(jax.device_count())
```

If the flag is set, we expect the program to end when TPU is already in use. So I made two terminals, in one of them using `ipython` I run these two lines to get TPU running: 
```
import jax
jax.device_count()
```
and I keep ipython running. Then in the other terminal, I run the file test.py, it will raise the exception with this error: 
```
WARNING: Logging before InitGoogle() is written to STDERR
I0000 00:00:1656090003.394494 1333717 tpu_initializer_helper.cc:116] libtpu.so is already in use by process with pid 1332169. Not attempting to load libtpu.so in this process.
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 305, in backends
    backend = _init_backend(platform)
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 349, in _init_backend
    backend = factory()
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 181, in tpu_client_timer_callback
    client = xla_client.make_tpu_client()
  File "/home/shahrokhi/.local/lib/python3.8/site-packages/jaxlib/xla_client.py", line 108, in make_tpu_client
    return _xla.get_tpu_client(
jaxlib.xla_extension.XlaRuntimeError: ABORTED: libtpu.so is already in use by process with pid 1332169. Not attempting to load libtpu.so in this process.

  File "test.py", line 5, in <module>
    print(jax.device_count())
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 416, in device_count
    return int(get_backend(backend).device_count())
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 389, in get_backend
    return _get_backend_uncached(platform)
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 373, in _get_backend_uncached
    bs = backends()
  File "/home/shahrokhi/jax/jax/_src/lib/xla_bridge.py", line 330, in backends
    raise RuntimeError(f"Unable to initialize backend '{platform}': {err}")
RuntimeError: Unable to initialize backend 'tpu': ABORTED: libtpu.so is already in use by process with pid 1332169. Not attempting to load libtpu.so in this process.
```

If the flag is not set, I get this output: 
```
WARNING: Logging before InitGoogle() is written to STDERR
I0000 00:00:1656089625.706280 1313830 tpu_initializer_helper.cc:116] libtpu.so is already in use by process with pid 1298889. Not attempting to load libtpu.so in this process.
1
```

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -602,6 +602,22 @@ jax2tf_associative_scan_reductions = config.define_bool_state(
     )
 )
 
+jax_platforms = config.define_string_state(
+    name='jax_platforms',
+    default=None,
+    help=(
+        'Comma-separated list of platform names specifying which platforms jax '
+        'should initialize. If any of the platforms in this list are not successfully '
+        'initialized, an exception will be raised and the program will be aborted. '
+        'The first platform in the list will be the default platform. '
+        'For example, config.jax_platforms=cpu,tpu means that CPU and TPU backends '
+        'will be initialized, and the CPU backend will be used unless otherwise '
+        'specified. If TPU initialization fails, it will raise an exception. '
+        'By default, jax will try to initialize all available '
+        'platforms and will default to GPU or TPU if available, and fallback to CPU '
+        'otherwise.'
+        ))
+
 enable_checks = config.define_bool_state(
     name='jax_enable_checks',
     default=False,
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -35,6 +35,7 @@ from jax._src.config import flags, bool_env, int_env
 from jax._src.lib import tpu_driver_client
 from jax._src.lib import xla_client
 from jax._src import util, traceback_util
+from jax.config import config
 import numpy as np
 
 iree: Optional[Any]
@@ -64,17 +65,6 @@ flags.DEFINE_string(
     'jax_platform_name',
     os.getenv('JAX_PLATFORM_NAME', '').lower(),
     'Deprecated, please use --jax_platforms instead.')
-flags.DEFINE_string(
-    'jax_platforms',
-    os.getenv('JAX_PLATFORMS', '').lower(),
-    'Comma-separated list of platform names specifying which platforms jax '
-    'should attempt to initialize. The first platform in the list that is '
-    'successfully initialized will be used as the default platform. For '
-    'example, --jax_platforms=cpu,gpu means that CPU and GPU backends will be '
-    'initialized, and the CPU backend will be used unless otherwise specified; '
-    '--jax_platforms=cpu means that only the CPU backend will be initialized. '
-    'By default, jax will try to initialize all available platforms and will '
-    'default to GPU or TPU if available, and fallback to CPU otherwise.')
 flags.DEFINE_bool(
     'jax_disable_most_optimizations',
     bool_env('JAX_DISABLE_MOST_OPTIMIZATIONS', False),
@@ -297,9 +287,8 @@ def backends():
   with _backend_lock:
     if _backends:
       return _backends
-
-    if FLAGS.jax_platforms:
-      jax_platforms = FLAGS.jax_platforms.split(",")
+    if config.jax_platforms:
+      jax_platforms = config.jax_platforms.split(",")
       platforms = []
       # Allow platform aliases in the list of platforms.
       for platform in jax_platforms:
@@ -310,7 +299,6 @@ def backends():
       platforms_and_priorites = (
           (platform, priority) for platform, (_, priority)
           in _backend_factories.items())
-
     default_priority = -1000
     for platform, priority in platforms_and_priorites:
       try:
@@ -337,10 +325,13 @@ def backends():
         else:
           # If the backend isn't built into the binary, or if it has no devices,
           # we expect a RuntimeError.
-          logging.info("Unable to initialize backend '%s': %s", platform,
-                       err)
-          _backends_errors[platform] = str(err)
-          continue
+          err_msg = f"Unable to initialize backend '{platform}': {err}"
+          if config.jax_platforms:
+            raise RuntimeError(err_msg)
+          else:
+            _backends_errors[platform] = str(err)
+            logging.info(err_msg)
+            continue
     # We don't warn about falling back to CPU on Mac OS, because we don't
     # support anything else there at the moment and warning would be pointless.
     if (py_platform.system() != "Darwin" and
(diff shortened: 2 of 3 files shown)


sshahrokhi:
@skye could you please take a look at this PR? thanks!

jax/_src/lib/xla_bridge.py
```diff
@@ -297,21 +287,32 @@ def backends():
...
       # Allow platform aliases in the list of platforms.
       for platform in jax_platforms:
         platforms.extend(expand_platform_alias(platform))
       priorities = range(len(platforms), 0, -1)
       platforms_and_priorites = zip(platforms, priorities)
-    else:
-      platforms_and_priorites = (
-          (platform, priority) for platform, (_, priority)
-          in _backend_factories.items())
+      for platform, priority in platforms_and_priorites:
```
skye, line 299: Instead of creating new logic for when jax_platforms instead, please combine with the existing logic starting around L316. The only difference should be how to handle backend initialization errors, so you can create a new variable like `continue_on_backend_init_errors` to keep track of what behavior should be used. This logic is already a bit different from the below logic (e.g. doesn't handle the "rocm" GPU stuff).

skye (commented):
Sorry for the delay!

jax/_src/config.py
```diff
@@ -585,6 +585,22 @@ def update_thread_local_jit_state(**kw):
     )
 )
 
+jax_platforms = config.define_string_state(
+    name='jax_platforms',
+    default=None,
+    help=(
+        'Comma-separated list of platform names specifying which platforms jax '
+        'should initialize. If any of the platforms in this list is not successfully '
```
skye, line 593:
```suggestion
        'should initialize. If any of the platforms in this list aren't successfully '
```

jax/_src/config.py
```diff
@@ -585,6 +585,22 @@ def update_thread_local_jit_state(**kw):
...
+        'initialized, an exception will be raised and the program will be aborted. '
+        'The first platform in the list will be the default platform. '
+        'For example, config.jax_platforms=cpu,tpu means that CPU and TPU backends '
+        'will be initialized, and the CPU backend will be used unless otherwise '
+        'specified. If TPU initialization fails, it will abort the process returning '
+        'the error message. By default, jax will try to initialize all available '
```
skye, line 599: Say "it will raise an exception" instead of "will abort the process returning the error message"

jax/_src/lib/xla_bridge.py
```diff
@@ -297,9 +287,9 @@ def backends():
   with _backend_lock:
     if _backends:
       return _backends
-
-    if FLAGS.jax_platforms:
-      jax_platforms = FLAGS.jax_platforms.split(",")
+    default_priority = -1000
```
skye, line 290: You can move this back to where it originally was now (it's good to define variables as close to where you use them as possible)

jax/_src/lib/xla_bridge.py
```diff
@@ -340,7 +329,10 @@ def backends():
           logging.info("Unable to initialize backend '%s': %s", platform,
                        err)
           _backends_errors[platform] = str(err)
```
skye, line 331: We only need to populate _backend_errors in the else branch

jax/_src/lib/xla_bridge.py
```diff
@@ -340,7 +329,10 @@ def backends():
           logging.info("Unable to initialize backend '%s': %s", platform,
                        err)
```
skye, line 330: Let's reuse this string for both logging and raising an error (i.e. we should say "Unable to initialize backend 'foo':" at the beginning of the error message). You can just raise a RuntimeError with the new message, I don't think we need to preserve the original error type. Also you only need to log in the else case, no need to log if we're about to raise an error.

skye (commented):
This should be tested, but I'm not sure how to write a unit test. For now, can you manually test it, and put a description of what you did in the PR descriptions? A copy/paste of the exact commands + output would be ideal, or at least give an example error message.

jax/_src/lib/xla_bridge.py
sshahrokhi: I tried a bunch of things, it seems that the err will be shown anyways when we fall back to CPU and it was not writing the unable to initialize backend unless I make it a warning, then it will make a lot of warnings for TPU, GPU, etc, so I removed the logging altogether. I raise a runtime exception instead of just raising, again because the message was not propagating correctly if I didn't do that! Please let me know what you think.

jax/_src/lib/xla_bridge.py
skye: Oh sorry, I mean in the else branch below. Basically we don't want to change the existing behavior when --jax_platforms isn't set, so you still want to do the logging + populating _backends_errors as an alternative to raising a RuntimeError. You can also pull out creating the "Unable to initialize backend" string since you'll use that in both places below.

skye (changes requested):

jax/_src/lib/xla_bridge.py
sshahrokhi: But the logging does not happen! if we add or not add that line with LOG(INFO) it is not printing the message. That is why I removed it, without populating the backends the tests failed, so I brought it back :)

skye: That's because jax doesn't print INFO logs by default, but you can enable them with TF_CPP_MIN_LOG_LEVEL=0. This log message actually ends up confusing people because importing tensorflow also turns on logging, so people see the "tpu_driver" initialization failing (which is the legacy TPU backend, not the TPU VM backend) and reasonably think it will affect the TPU VM. So there is an argument for removing it now that we can get the error with JAX_PLATFORMS (being able to access the error is super important for debugging). I think we should keep it for now though, because it might still be more convenient than figuring out which platform to specify in some cases.

Please move these into the else clause below too, none of this needs to happen if we're about to raise a RuntimeError.

skye: The _backends_errors bit still needs to go in the else clause (since it's not necessary otherwise). Otherwise LGTM!

skye (approved):
Looks good! Let's get this in

PR #11149, opened 2022-06-17
Make sure libtpu is initialized before initializing TPU profiling.

skye:
The TPU tracer, which is necessary for populating TPU events in a
profile, will abort initialization if libtpu isn't initialized:
https://github.com/tensorflow/tensorflow/blob/1f4d1e5c832f8a50ea49a1659dec0db01a1fc4d8/tensorflow/core/profiler/backends/tpu/tpu_tracer.cc#L124-L127

This change makes sure all JAX backends are initialized before
creating a profiler trace, which is when the TPU tracer will be
initialized. In particular, we want to make sure that the Cloud TPU
backend is initialized if it's available (if it's not available, the
TPU tracer isn't expected to initialize).

The breakage was initially introduced in
https://github.com/tensorflow/tensorflow/pull/55429 on May 4, but was
only caught via ProfilerTest.testProgrammaticProfilingon in
profiler_test.py starting June 10. We believe something must have
changed in the test or JAX itself to cause this issue to only show up
later. I would ideally figure out how to write a more robust
regression test, but in the interest of fixing TPU profiling, this
change is only tested via the existing profiler tests.

### DIFF
--- a/jax/_src/profiler.py
+++ b/jax/_src/profiler.py
@@ -46,6 +46,15 @@ def start_server(port: int):
   global _profiler_server
   if _profiler_server is not None:
     raise ValueError("Only one profiler server can be active at a time.")
+
+  # Make sure backends are initialized before creating a profiler
+  # session. Otherwise on Cloud TPU, libtpu may not be initialized before
+  # creating the tracer, which will cause the TPU tracer initialization to
+  # fail and no TPU operations will be included in the profile.
+  # NOTE(skyewm): I'm not sure this is necessary for start_server (is definitely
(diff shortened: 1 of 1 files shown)


sshahrokhi:
Thanks for the change, LGTM, just curious when do we start the profiler server? Do we do it right before starting trace, or sometime sooner?

hawkinsp (approved):

skye:
Users start the profiler server themselves, similar to starting a trace: https://jax.readthedocs.io/en/latest/profiling.html#manual-capture-via-tensorboard

PR #11161, opened 2022-06-17
Remove spurious single quote in profiler.py

schmrlng:
Before:
```
127.0.0.1 - - [17/Jun/2022 16:08:18] "GET /perfetto_trace.json.gz' HTTP/1.1" 404 -
```
After:
```
127.0.0.1 - - [17/Jun/2022 16:35:54] "GET /perfetto_trace.json.gz HTTP/1.1" 200 -
```


jakevdp (approved):
Good catch, thanks!

PR #11173, opened 2022-06-20
Fix overflow of large prng computation

### LINKED ISSUE #11010: jax.random.uniform raises an `InconclusiveDimensionOperation` for very large shapes
```
def f():
  return jax.random.uniform(jax.random.PRNGKey(3), (308000000,128), dtype=jnp.bfloat16)

r = jax.jit(f, device=jax.devices("cpu")[0])()
```

fails with:

```
InconclusiveDimensionOperation            Traceback (most recent call last)
[google3/third_party/py/jax/_src/prng.py](http://google3/third_party/py/jax/_src/prng.py) in threefry_random_bits(key, bit_width, shape)
    543       )
    544     )
--> 545     bits = lax.reshape(bits, (np.uint32(max_count * 32 // bit_width),), (1, 0))
    546     bits = lax.convert_element_type(bits, dtype)[:size]
    547   return lax.reshape(bits, shape)

InconclusiveDimensionOperation: Cannot divide evenly the sizes of shapes (2, 19712000000) and (769294336,)
```

gnecula:
Fixes: #11010

### DIFF
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -542,7 +542,7 @@ def threefry_random_bits(key: jnp.ndarray, bit_width, shape):
         )
       )
     )
-    bits = lax.reshape(bits, (np.uint32(max_count * 32 // bit_width),), (1, 0))
+    bits = lax.reshape(bits, ((max_count * 32 // bit_width),), (1, 0))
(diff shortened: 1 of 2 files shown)


froystig (approved):
Nice. Unclear to me why the `uint32` cast was there to begin with...

PR #11179, opened 2022-06-20
fix buggy assert

mattjj:


gnecula:
Is this assert going to be fixed in a different way?

PR #11214, opened 2022-06-22
Add `for_loop` primitive and impl rule

sharadmv:
Tracker: #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -14,30 +14,38 @@
 """Module for the `for_loop` primitive."""
 from functools import partial
 
-from typing import Any, Dict, List, Sequence, Tuple
+from typing import Any, Callable, Dict, Generic, List, Sequence, Tuple, TypeVar
 
 from jax import core
 from jax import lax
 from jax import linear_util as lu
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow/for_loop.py
```diff
@@ -372,3 +386,114 @@ def write(v: core.Var, val: Any) -> None:
...
+  ```
+
+  Args:
+    nsteps: Number of iterations
+    body: A callable that takes in the iteration number as its first argument
+          and `Ref`s corresponding to `init_state` as its second argument.
+          `body` is free to read from and write to its `Ref`s. `body` should
+          not return anything.
+    init_state: A Pytree of JAX-compatible values used to initialize the `Ref`s
+                that will be passed into the for loop body.
```
mattjj, line 460:
```suggestion
    body: A callable that takes in the iteration number as its first argument
      and `Ref`s corresponding to `init_state` as its second argument.
      `body` is free to read from and write to its `Ref`s. `body` should
       not return anything.
    init_state: A Pytree of JAX-compatible values used to initialize the `Ref`s
      that will be passed into the for loop body.
```

I think something like that is how we usually indent these docstrings.

mattjj (approved):

PR #11230, opened 2022-06-23
Make JAX compatible with numpy version 1.23.0

### LINKED ISSUE #11221: autodiff fails under numpy 1.23
Something in the numpy 1.23 release broke autodiff; in particular, float0 arrays are being converted to bool:
```python
import jax
import numpy
print(numpy.__version__)
print(jax.grad(lambda x: x+0., allow_int=True)(1).dtype)
```
Newest numpy:
```
1.23.0
bool
```
previous version:
```
1.22.4
[('float0', 'V')]
```

jakevdp:
For more discussion of these changes, see #11221

Fixes #11221

### DIFF
--- a/build/test-requirements.txt
+++ b/build/test-requirements.txt
@@ -6,4 +6,3 @@ pillow>=8.3.1,<9.1.0
 pytest-benchmark
 pytest-xdist
 wheel
-numpy<1.23.0
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -957,7 +957,7 @@ def _maybe_perturbed(x: Any) -> bool:
     # happen later, but some types always have trivial tangents.
     vspace = x.aval.at_least_vspace()
     return not (vspace is core.abstract_token or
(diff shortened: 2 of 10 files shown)


jakevdp:
(related: https://github.com/numpy/numpy/issues/21838 and https://github.com/numpy/numpy/issues/21840)

build/test-requirements.txt
```diff
@@ -6,4 +6,4 @@ pillow>=8.3.1,<9.1.0
 pytest-benchmark
 pytest-xdist
 wheel
-numpy<1.23.0
+numpy
```
yashk2810, line 9: Remove this? setup.py should cover it

yashk2810 (approved):

build/test-requirements.txt
jakevdp: I don't understand - we want the tests to run on the most recent version of numpy.

yashk2810: Right. There is no upper bound on version constraint on numpy.

This numpy was not there before I added the `<1.23.0` and we still hit that error. So I think removing this should be fine.

Issue #11225, opened 2022-06-23
Reduce the size of jaxlib wheels


nicholasjng:
On my machine (Mac ARM64), installed wheels approach the 250 MB mark, which makes it the biggest wheel installed by quite a bit (the closest second is scipy with 100MB).

The wheel size is practically entirely determined by the size of three binaries, `xla_extension.so` (clocking in at 144MB), `libjaxlib_mlir_capi.dylib` (51MB) and `tpu_client_extension.so` (44MB).

Looking at potential packaging uses (think Docker or otherwise-constrained environments either in network bandwidth or storage), where every megabyte might count to get a lean solution, I wanted to ask if there are ways to reduce the size of these wheels, maybe also based on target platform. For example, I should probably not need the `tpu_client_extension` when I am on M1, right?

I know of ways to shrink wheel sizes (and thus the resulting Docker image size) for other popular scientific Python packages, e.g. by stripping debug symbols at compilation. Is there a way to do the same with jaxlib?

Possibly related to / interesting for #8943.

hawkinsp:
The easiest thing we could do is to omit the `tpu_client_extension`. *Technically* you could use it from a Mac since it talks to remote TPUs. But in practice, I very much doubt any one does, and it's mostly deprecated anyway (in favor of only supporting local TPUs, which no Mac has). This is just a build flag change in our wheel build.

It's also quite likely we can refactor things a bit to share code between the xla_extension and the jaxlib_mlir dylibs.

nicholasjng:
Sounds good. Anything I could contribute here to get it done, or is that more of an internal task?

hawkinsp:
The newly-released 0.3.14 wheels reduce the installed size of the Mac ARM wheel to 199MB by omitting the TPU extension.

I hope that helps. We welcome PRs to reduce the size further!

nicholasjng:
I wanted to check back in on this briefly, as the wheel is now back at ~220MB due to the MLIR .so file.

Would it be possible to share a quick conceptual overview of what goes into the XLA and MLIR .so files that makes them so large? That way I might get an understanding that could help with the size management :)

PR #11231, opened 2022-06-23
Add a build flag that allows disabling remote TPU builds.

hawkinsp:
Disable remote TPU by default.

The proximate reason to do this is so we don't include it in Mac builds (#11225).

For the moment, I intend to leave this enabled in Linux non-CUDA release wheels primarily for the use of colab, although this code path is deprecated in favor of the cloud TPU VM support.

Issue #11225

### DIFF
--- a/build/BUILD.bazel
+++ b/build/BUILD.bazel
@@ -14,6 +14,7 @@
 
 # JAX is Autograd and XLA
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda")
 load("@local_config_rocm//rocm:build_defs.bzl", "if_rocm")
 load("@org_tensorflow//tensorflow:tensorflow.bzl", "if_not_windows", "if_windows")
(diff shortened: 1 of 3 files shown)


build/BUILD.bazel
```diff
@@ -22,6 +23,18 @@ licenses(["notice"])  # Apache 2
 
 package(default_visibility = ["//visibility:public"])
 
+bool_flag(
+    name = "enable_remote_tpu",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "remote_tpu_enabled",
```
skye, line 32: Maybe rename `legacy_tpu_*`? Up to you

build/BUILD.bazel
```diff
@@ -22,6 +23,18 @@ licenses(["notice"])  # Apache 2
 
 package(default_visibility = ["//visibility:public"])
 
+bool_flag(
+    name = "enable_remote_tpu",
+    build_setting_default = False,
```
skye, line 28: Does this need to be enabled somewhere for the linux builds?

build/BUILD.bazel
hawkinsp: I might leave it as is... I think this is a better description of why it's different to the other "TPU".

And hopefully we can remove it sometime anyway!

PR #11234, opened 2022-06-24
jnp.ndarray: raise error for binary operations with lists and tuples

### LINKED ISSUE #2406: Behavior of `!=` operator is confusing when rhs is a list
Compare:
```python
for np in jnp, onp:
  print(np)
  print(np.array([1,2]) != [1,2])
  print(np.array([1,2]) != [3,2])
  print(np.array([1,2]) != np.array([1,2]))
  print(np.array([1,2]) != np.array([3,2]))
```
Ouptut:
```
<module 'jax.numpy' from '/usr/local/lib/python3.6/dist-packages/jax/numpy/__init__.py'>
True
True
[False False]
[ True False]
<module 'numpy' from '/usr/local/lib/python3.6/dist-packages/numpy/__init__.py'>
[False False]
[ True False]
[False False]
[ True False]
```

The strange dynamically-shaped result results in some strange downstream exceptions when dropping in jax.numpy in place of numpy, such as:
```python
print(any(onp.array([1,2]) != [1,2]))
print(any(jnp.array([1,2]) != [1,2]))
```
Output:
```
False
<ipython-input-18-e89704e72507> in <module>()
      1 print(any(onp.array([1,2]) != [1,2]))
----> 2 print(any(jnp.array([1,2]) != [1,2]))

TypeError: 'bool' object is not iterable
```
(Of course the real use-case is `some_fn_returns_array(x) != some_fn_returns_list(y)`)

jakevdp:
This fixes a potentially confusing issue when JAX arrays are compared to python sequences.

When numpy encounters a list or tuple, it implicitly casts it to an array:
```python
In [1]: import numpy as np 
   ...: print(np.array([1, 1]) == [1, 2])                                                                                                                                                                                                                                                                         
[ True False]
```
In JAX, we tend not to do this for performance reasons, and because explicit is better than implicit in this case (see e.g. the discussion in #7737). Unfortunately, this case is not explicitly handled currently in binary operators, which makes the current behavior somewhat surprising:
```python
   ...: print(jnp.array([1, 1]) == [1, 2])                                                                                                                                                                                                                                                                        
False
```
This comes from `list.__req__`, and is a surprising result.

With this PR, this kind of comparison becomes an explicit `TypeError`:
```python
In [1]: import numpy as np 
   ...: print(np.array([1, 1]) == [1, 2])                                                                                                                                                                                                                                                                         
[ True False]

   ...: print(jnp.array([1, 1]) == [1, 2])                                                                                                                                                                                                                                                                        
<ipython-input-2-5af3cb7a0607> in <module>
      1 import jax.numpy as jnp
----> 2 print(jnp.array([1, 1]) == [1, 2])

~/github/google/jax/jax/_src/numpy/lax_numpy.py in deferring_binary_op(self, other)
   4583       other = other.__jax_array__()
   4584     if isinstance(other, _rejected_binop_types):
-> 4585       raise TypeError(f"unsupported operand type(s) for {opchar}: "
   4586                       f"{type(self).__name__!r} and {type(other).__name__!r}")
   4587     if not isinstance(other, _accepted_binop_types):

TypeError: unsupported operand type(s) for ==: 'DeviceArray' and 'list'
```
~This is a draft, mainly to see if the issue catches any tests. Still todo if we choose to make this change:~

- [x] add `set` and `dict`
- [x] ensure arguments are in correct order in error message
- [x] regression tests for this behavior
- [x] CHANGELOG entry

Closes #2406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
   * `JaxTestCase` and `JaxTestLoader` have been removed from `jax.test_util`. These
     classes have been deprecated since v0.3.1 ({jax-issue}`#11248`).
   * Added {class}`jax.scipy.gaussian_kde` ({jax-issue}`#11237`).
+  * Binary operations between JAX arrays and built-in collections (`dict`, `list`, `set`, `tuple`)
+    now raise a `TypeError` in all cases. Previously some cases (particularly equality and inequality)
+    would return boolean scalars inconsistent with similar operations in NumPy ({jax-issue}`#11234`).
 
 ## jaxlib 0.3.15 (Unreleased)
 
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4575,19 +4575,21 @@ def _not_implemented(fun):
 
 _scalar_types = (int, float, complex, np.generic)
 _accepted_binop_types = (int, float, complex, np.generic, np.ndarray, ndarray)
+_rejected_binop_types = (list, tuple, set, dict)
 
-def _defer_to_unrecognized_arg(binary_op):
+def _defer_to_unrecognized_arg(opchar, binary_op, swap=False):
   # Ensure that other array types have the chance to override arithmetic.
   def deferring_binary_op(self, other):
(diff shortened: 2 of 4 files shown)


jakevdp:
We should probably add `set` and `dict` to the mix as well.

jakevdp:
Ready for review - I'll still have to land a number of downstream fixes before merging, though.

PR #11237, opened 2022-06-24
Adding an implementation of scipy.stats.gaussian_kde

### LINKED ISSUE #8954: Please consider implementing scipy.stats.gaussian_kde
Would it be possible to add [`scipy.stats.gaussian_kde`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html) to Jax?  I use this function to generate plots, but it's very slow.  If I could apply Jax's awesome jit to it, it could run much faster.

Looking at the code https://github.com/scipy/scipy/blob/master/scipy/stats/_kde.py and https://github.com/scipy/scipy/blob/master/scipy/stats/_stats.pyx#L693, it doesn't seem too crazy as most of the peripheral functions are already in Jax.

dfm:
Closes #8954

I needed this for a project that I'm working on so decided to also draft a PR. I know it's a fairly large one, so no stress if it's not a good fit, but wanted to drop it here anyways!

I wasn't aiming for performance, but the performance is pretty good on both the CPU and GPU ([source](https://gist.github.com/dfm/8d41b6e087bfb7f416e9a98b78957bf3)):

jax-kde-benchmark

Comments welcome!

### DIFF
--- a/jax/_src/scipy/stats/kde.py
+++ b/jax/_src/scipy/stats/kde.py
@@ -1,0 +1,270 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from functools import partial
+from typing import Any
+
+import numpy as np
+import scipy.stats as osp_stats
+
+import jax.numpy as jnp
+from jax import jit, lax, random, vmap
+from jax._src.numpy.lax_numpy import _check_arraylike, _promote_dtypes_inexact
+from jax._src.numpy.util import _wraps
+from jax._src.tree_util import register_pytree_node_class
+from jax.scipy import linalg, special
+
+
+@_wraps(osp_stats.gaussian_kde, update_doc=False)
+@register_pytree_node_class
+@dataclass(frozen=True, init=False)
+class gaussian_kde:
+  neff: Any
+  dataset: Any
+  weights: Any
+  covariance: Any
+  inv_cov: Any
+
+  def __init__(self, dataset, bw_method=None, weights=None):
+    _check_arraylike("gaussian_kde", dataset)
+    dataset = jnp.atleast_2d(dataset)
+    if jnp.issubdtype(lax.dtype(dataset), jnp.complexfloating):
+      raise NotImplementedError("gaussian_kde does not support complex data")
+    if not dataset.size > 1:
+      raise ValueError("`dataset` input should have multiple elements.")
+
+    d, n = dataset.shape
+    if weights is not None:
+      _check_arraylike("gaussian_kde", weights)
+      dataset, weights = _promote_dtypes_inexact(dataset, weights)
+      weights = jnp.atleast_1d(weights)
+      weights /= jnp.sum(weights)
+      if weights.ndim != 1:
+        raise ValueError("`weights` input should be one-dimensional.")
+      if len(weights) != n:
+        raise ValueError("`weights` input should be of length n")
+    else:
+      dataset, = _promote_dtypes_inexact(dataset)
+      weights = jnp.full(n, 1.0 / n, dtype=dataset.dtype)
+
+    self._setattr("dataset", dataset)
+    self._setattr("weights", weights)
+    neff = self._setattr("neff", 1 / jnp.sum(weights**2))
+
+    bw_method = "scott" if bw_method is None else bw_method
+    if bw_method == "scott":
+      factor = jnp.power(neff, -1. / (d + 4))
+    elif bw_method == "silverman":
+      factor = jnp.power(neff * (d + 2) / 4.0, -1. / (d + 4))
+    elif jnp.isscalar(bw_method) and not isinstance(bw_method, str):
+      factor = bw_method
+    elif callable(bw_method):
+      factor = bw_method(self)
+    else:
+      raise ValueError(
+          "`bw_method` should be 'scott', 'silverman', a scalar, or a callable."
+      )
+
+    data_covariance = jnp.atleast_2d(
+        jnp.cov(dataset, rowvar=1, bias=False, aweights=weights))
+    data_inv_cov = jnp.linalg.inv(data_covariance)
+    covariance = data_covariance * factor**2
+    inv_cov = data_inv_cov / factor**2
+    self._setattr("covariance", covariance)
+    self._setattr("inv_cov", inv_cov)
+
+  def _setattr(self, name, value):
+    # Frozen dataclasses don't support setting attributes so we have to
+    # overload that operation here as they do in the dataclass implementation
+    object.__setattr__(self, name, value)
+    return value
+
+  def tree_flatten(self):
+    return ((self.neff, self.dataset, self.weights, self.covariance,
+             self.inv_cov), None)
+
+  @classmethod
+  def tree_unflatten(cls, aux_data, children):
+    del aux_data
+    kde = cls.__new__(cls)
+    kde._setattr("neff", children[0])
+    kde._setattr("dataset", children[1])
+    kde._setattr("weights", children[2])
+    kde._setattr("covariance", children[3])
+    kde._setattr("inv_cov", children[4])
+    return kde
+
+  @property
+  def d(self):
+    return self.dataset.shape[0]
+
+  @property
+  def n(self):
+    return self.dataset.shape[1]
+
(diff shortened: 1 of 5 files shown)


jax/_src/scipy/stats/kde.py
```diff
@@ -0,0 +1,249 @@
...
+
+import jax.numpy as jnp
+from jax import jit, random, vmap
+from jax._src.numpy.lax_numpy import _promote_dtypes_inexact
+from jax._src.numpy.util import _wraps
+from jax.scipy import linalg, special
+
+
+@_wraps(osp_stats.gaussian_kde, update_doc=False)
+def gaussian_kde(dataset, bw_method=None, weights=None):
```
jakevdp, line 29: Let's call `_check_arraylike` before `atleast_2d` (see #7737 for context)

jax/_src/scipy/stats/kde.py
```diff
@@ -0,0 +1,249 @@
...
+  dataset = jnp.atleast_2d(dataset)
+  if not dataset.size > 1:
+    raise ValueError("`dataset` input should have multiple elements.")
+
+  d, n = dataset.shape
+  if weights is not None:
+    dataset, weights = _promote_dtypes_inexact(dataset, weights)
```
jakevdp, line 36: call `_check_arraylike` on weights

jax/_src/scipy/stats/kde.py
```diff
@@ -0,0 +1,249 @@
...
+    weights = jnp.atleast_1d(weights)
+    weights /= jnp.sum(weights)
+    if weights.ndim != 1:
+      raise ValueError("`weights` input should be one-dimensional.")
+    if len(weights) != n:
+      raise ValueError("`weights` input should be of length n")
+  else:
+    weights = jnp.full(n, 1.0 / n, dtype=dataset.dtype)
```
jakevdp, line 44: What if `dataset.dtype` is integer - should we promote to inexact first?

tests/scipy_stats_test.py
```diff
@@ -687,6 +688,174 @@ def testMultivariateNormalLogpdfBatch(self, ndim, nbatch, dtype):
...
+        lax_fun, args_maker, rtol={np.float32: 3e-07, np.float64: 4e-15})
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape], [dtype]),
+       "dtype": dtype,
+       "shape": shape}
+      for shape in [(15,), (3, 15), (1, 12)]
+      for dtype in jtu.dtypes.floating))
+  def testKdeIntegrateGaussian(self, shape, dtype):
+    @jax.numpy_dtype_promotion("standard")
```
jakevdp, line 750: This decorator shouldn't be necessary if the inputs are numpy arrays (which I believe they should be, given how this function is called)

jax/_src/scipy/stats/kde.py
```diff
@@ -0,0 +1,249 @@
...
+import jax.numpy as jnp
+from jax import jit, random, vmap
+from jax._src.numpy.lax_numpy import _promote_dtypes_inexact
+from jax._src.numpy.util import _wraps
+from jax.scipy import linalg, special
+
+
+@_wraps(osp_stats.gaussian_kde, update_doc=False)
+def gaussian_kde(dataset, bw_method=None, weights=None):
+  dataset = jnp.atleast_2d(dataset)
```
jakevdp, line 30: Should we raise an error if the dataset is complex?

tests/scipy_stats_test.py
```diff
@@ -687,6 +688,174 @@ def testMultivariateNormalLogpdfBatch(self, ndim, nbatch, dtype):
...
+    self._CompileAndCheck(
+      func, args_maker, rtol={np.float32: 3e-07, np.float64: 4e-15})
+    result = func(*args)
+    assert result.shape == (ndim,)
+
+    func = partial(resample, shape=(4,))
+    self._CompileAndCheck(
+      func, args_maker, rtol={np.float32: 3e-07, np.float64: 4e-15})
+    result = func(*args)
+    assert result.shape == (ndim, 4)
```
jakevdp, line 857: I wonder if we could use some of the strategies from `random_test.py` to confirm that the values are drawn from the expected distribution here, at least in the 1D case... for example the KS test of the CDF: https://github.com/google/jax/blob/e32373c3eaa2d8d49372cdfb8698ee2d7af162ab/tests/random_test.py#L470-L474

Though this would depend on computing an accurate cdf...

jakevdp (changes requested):
Looks really good, just a few minor comments. Thanks!

tests/scipy_stats_test.py
dfm: Good point! Turns out that we could already compute the cdf in the 1D case using `integrate_box_1d`.

tests/scipy_stats_test.py
dfm: I could have sworn that I was getting a promotion error before, but it looks like you're right.

jakevdp: If you were calling `jnp.abs` instead of `np.abs` (or something similar) on inputs, then I'd believe this error could come up

jakevdp (changes requested):
Looks great! Last couple things:

- please add an appropriate section to https://github.com/google/jax/blob/main/docs/jax.scipy.rst so this will appear in the docs
- please add a CHANGELOG entry noting this new API

Finally, a question about the design: here `gaussian_kde` is a function that returns a namedtuple with extra methods. I think this is fine in terms of typical usage, but might be a bit confusing (because, e.g. `scipy.stats.gaussian_kde.pdf` exists, while `jax.scipy.stats.gaussian_kde.pdf` does not). I wonder if `gaussian_kde` should be a class with methods and a suitable PyTree registration so instances of it can be passed to JAX functions? In some ways that's cleaner, and in some ways it's not. What do you think?

dfm:
> Finally, a question about the design: here gaussian_kde is a function that returns a namedtuple with extra methods. I think this is fine in terms of typical usage, but might be a bit confusing (because, e.g. scipy.stats.gaussian_kde.pdf exists, while jax.scipy.stats.gaussian_kde.pdf does not). I wonder if gaussian_kde should be a class with methods and a suitable PyTree registration so instances of it can be passed to JAX functions? In some ways that's cleaner, and in some ways it's not. What do you think?

Great question! I thought about this quite a bit, but couldn't decide which would be better. The main reason that I started with this was to get immutability for free, but I'm not sure how necessary that really is. But I do agree that it probably would be a lot less confusing to have a `gaussian_kde` class to match the `scipy` interface. Happy to refactor that way if you think that's ok!

jakevdp (changes requested):
I really like the dataclass approach here. A couple more comments below

jax/_src/scipy/stats/kde.py
```diff
@@ -0,0 +1,267 @@
...
+from jax._src.tree_util import register_pytree_node_class
+from jax.scipy import linalg, special
+
+
+@_wraps(osp_stats.gaussian_kde, update_doc=False)
+@register_pytree_node_class
+@dataclass(frozen=True, init=False)
+class gaussian_kde:
+  d: int
+  n: int
```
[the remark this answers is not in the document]
dfm: Good point! I'll do that.

tests/scipy_stats_test.py
```diff
@@ -684,6 +685,203 @@ def testMultivariateNormalLogpdfBatch(self, ndim, nbatch, dtype):
...
+    self._CompileAndCheck(
+        lax_fun, args_maker, rtol={np.float32: 3e-07, np.float64: 4e-15})
+
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape], [dtype]),
+       "dtype": dtype,
+       "shape": shape}
+      for shape in [(15,), (3, 15), (1, 12)]
+      for dtype in jtu.dtypes.floating))
+  @jax.numpy_dtype_promotion("standard")
```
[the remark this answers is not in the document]
dfm: When type promotion is strict, this test currently fails, but not because of this code. In fact it looks like it's actually caused by another issue in `random.choice` that I've just opened: https://github.com/google/jax/issues/11285

jakevdp (approved):
Looks great - thanks for working on this!

jakevdp:
Everything's looking good, except on TPU backends some of the `testKde` tests fail with e.g.:
```
AssertionError: 
Not equal to tolerance rtol=0.001, atol=0.001

Mismatched elements: 7 / 50 (14%)
Max absolute difference: 0.07452869
Max relative difference: 0.00547456
 x: array([ -6.917634,  -8.133598,  -7.125779,  -7.454879,  -7.296328,
        -7.365776,  -6.817926,  -9.488862,  -7.272069,  -8.460855,
       -13.688169,  -6.750305,  -6.494191,  -7.032907,  -8.428143,...
 y: array([ -6.916273,  -8.131803,  -7.12294 ,  -7.452182,  -7.291719,
        -7.362315,  -6.817333,  -9.481265,  -7.267247,  -8.457088,
       -13.613641,  -6.748472,  -6.494314,  -7.032794,  -8.427602,...
```
Can you set the rtol to 0.01 if `device_under_test` is `"tpu"`?

PR #11266, opened 2022-06-25
[ROCm] Fixes for compilation failures caused by compiler changes in R…

rsanthanam-amd:
…OCm Tensorflow fork.

/cc @hawkinsp

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -99,6 +99,7 @@ build:windows --host_linkopt=/OPT:ICF
 build:windows --experimental_strict_action_env=true
 
 build:linux --config=posix
+build:linux --copt=-Wno-unknown-warning-option
 # Workaround for gcc 10+ warnings related to upb.
 # See https://github.com/tensorflow/tensorflow/issues/39467
 build:linux --copt=-Wno-stringop-truncation
--- a/jaxlib/rocm/hipsolver_kernels.cc
+++ b/jaxlib/rocm/hipsolver_kernels.cc
@@ -126,30 +126,30 @@ static absl::Status Potrf_(hipStream_t stream, void** buffers,
       case HipsolverType::F32: {
         JAX_RETURN_IF_ERROR(JAX_AS_STATUS(hipsolverSpotrfBatched(
             handle.get(), d.uplo, d.n, static_cast<float**>(workspace), d.n,
-            static_cast<float*>(workspace + (d.batch * sizeof(float*))), d.lwork,
-            info, d.batch)));
+            reinterpret_cast<float*>(static_cast<float**>(workspace) + d.batch),
+                d.lwork, info, d.batch)));
         break;
       }
       case HipsolverType::F64: {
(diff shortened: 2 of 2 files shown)


jaxlib/rocm/hipsolver_kernels.cc
```diff
@@ -126,30 +126,30 @@ static absl::Status Potrf_(hipStream_t stream, void** buffers,
       case HipsolverType::F32: {
         JAX_RETURN_IF_ERROR(JAX_AS_STATUS(hipsolverSpotrfBatched(
             handle.get(), d.uplo, d.n, static_cast<float**>(workspace), d.n,
-            static_cast<float*>(workspace + (d.batch * sizeof(float*))), d.lwork,
-            info, d.batch)));
+            static_cast<float*>(static_cast<float*>(workspace) + (d.batch *
```
rsanthanam-amd, line 129: The following changes fix the compilation error given for arithmetic with void pointers.

.bazelrc
```diff
@@ -99,6 +99,7 @@ build:windows --host_linkopt=/OPT:ICF
 build:windows --experimental_strict_action_env=true
 
 build:linux --config=posix
+build:linux --copt=-Wno-unknown-warning-option
```
rsanthanam-amd, line 102: This is necessary since we are using gcc 9.4 and the subsequent two warning flags don't exist so the compiler gives an error.

jaxlib/rocm/hipsolver_kernels.cc
hawkinsp: I think this might be incorrect. If you've cast the base pointer to `float *`, then the pointer arithmetic shouldn't have a `sizeof(float)` scaling factor. So either cast the base pointer to `char*` or remove the `sizeof(float)` I think?

rsanthanam-amd: So I tried casting the base pointer to char* and leaving the sizeof but the compiler gave an error:

error: static_cast from 'char *' to 'float *' is not allowed

Then I tried leaving the base pointer cast to float* and removing sizeof(float*) like so:

static_cast<float*>(static_cast<float*>(workspace) + d.batch

But in that case the linalg_test failed.

The way it is written seems to make everybody happy.

hawkinsp: I think you should have cast the base type to `float**`. i.e. the structure of the workspace is:
a) `batch` float* pointers followed by
b) a workspace of floats for the use of the kernel.

So your base pointer should be `float**`, and you should step `batch` elements into it.

hawkinsp: Also the fix for the `static_cast` error is you need to use `reinterpret_cast`.

rsanthanam-amd: Fixed.  Thanks for your suggestions.

hawkinsp (approved):

PR #11268, opened 2022-06-26
[dynamic-shapes] basic linearize and grad working

mattjj:
Summary of changes:
* Added a `_broadcast_in_dim_partial_eval` rule. We can't use the standard partial eval rule for broadcasting when dynamic shapes are involved, because we need to instantiate any dynamic axis sizes in the trace (i.e. add them as inputs to the staged-out/unknown jaxpr and indicate that they should be saved as residuals) and produce an output Tracer with (lifted) Tracers in its aval's shape.
* Add a `core.get_referent` function, which is an identity function on non-Tracers and on Tracers calls a method on the Tracer instance. The purpose is to use this to decide when two values are symbolically equal even if they are not identical Python objects; such cases can arise in partial evaluation (hence why it's in an AD PR) because we might need to e.g. compare lifted JaxprTracers (representing a dynamic axis size in the jaxpr being staged out, so it's got a ConstVar recipe) to non-JaxprTracer objects (e.g. a DynamicJaxprTracer representing the same axis size but not wrapped the same way).
* Fixed some bugs around lu.WrappedFun type annotation updating. The one in partial_eval.py called `_update_annotation_known` was the only tricky one, just because extracting just the 'known part' of the type annotation requires re-indexing all the DBIdx. Relatedly, added more checks to `lu.annotate` to catch any malformed type annotations early.
* Adapt JaxprTrace, the mechanism for building jaxprs for linearize and vjp: (1) to handle Tracers in avals' shapes, e.g. in `new_arg` and `new_instantiated_const`, mainly by lifting such Tracers into the JaxprTrace (with ConstVar recipes), (2) in `JaxprTrace.process_call` to handle application of dynamic shape jaxprs, particularly using the helper `trace_to_subjaxpr_nounits_dyn` which is a bit long.
* Adapt `tracers_to_jaxpr` to handle substituting in Vars for Tracers in shapes, also improved clarity in general.
* Added some basic djax+AD tests.

Probably the most complex thing is `trace_to_subjaxpr_nounits_dyn`, but it's just bookkeeping.

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -283,8 +283,25 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
     in_type = tuple(unsafe_zip(abstract_args, itertools.repeat(True)))
     fun = lu.annotate(fun, in_type)
   else:
-    assert all(map(core.typematch, abstract_args,
-                   [a for a, k in fun.in_type if k]))
+    # Check that the provided abstract_args are consistent with in_type by first
+    # collecting values of axis size arguments, then substituting them in for
+    # DBIdx occurrences.
+    axis_sizes: Dict[core.DBIdx, int] = {}
+    abstract_args_iter = iter(abstract_args)
+    for expected_type, explicit in fun.in_type:
+      if explicit:
+        aval = next(abstract_args_iter)
+        if isinstance(expected_type, core.DShapedArray):
+          # Check the value for any DBIdx variables is consistent.
+          assert all(axis_sizes.setdefault(d1, d2) == d2
+                     for d1, d2 in zip(expected_type.shape, aval.shape)
+                     if type(d1) is core.DBIdx)
+          # Check the type matches after substitution.
+          expected_shape = [axis_sizes.get(d, d) for d in expected_type.shape]  # type: ignore
+          expected_aval = core.ShapedArray(
+              shape=tuple(expected_shape), dtype=expected_type.dtype,
+              weak_type=expected_type.weak_type)
+          assert core.typematch(expected_aval, aval)
   with log_elapsed_time(f"Finished tracing + transforming {fun.__name__} "
                         "for jit in {elapsed_time} sec"):
     jaxpr, out_type, consts = pe.trace_to_jaxpr_final2(
@@ -295,7 +312,9 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
 
   if config.jax_dynamic_shapes:
     keep_unused = True
+    has_outfeed = False
   else:
+    has_outfeed = core.jaxpr_uses_outfeed(jaxpr)
     jaxpr = apply_outfeed_rewriter(jaxpr)
 
   if not keep_unused:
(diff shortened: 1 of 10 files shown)


jax/_src/dispatch.py
```diff
@@ -283,8 +283,25 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
     in_type = tuple(unsafe_zip(abstract_args, itertools.repeat(True)))
     fun = lu.annotate(fun, in_type)
   else:
-    assert all(map(core.typematch, abstract_args,
-                   [a for a, k in fun.in_type if k]))
+    # Check that the provided abstract_args are consistent with in_type by first
```
mattjj, line 286: This check was just buggy before; we need to do some substitution to make sure the arguments to which the jaxpr is being applied (`abstract_args`) is consistent with `fun.in_type`.

jax/_src/lax/lax.py
```diff
@@ -82,6 +82,9 @@
 
 T = TypeVar("T")
 
+map, unsafe_map = safe_map, map
```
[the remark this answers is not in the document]
mattjj: I think this caught a bug in `_maybe_broadcast` also, or at least some code which was _accidentally_ working.

jax/_src/dispatch.py
mattjj: Actually, the new code here wasn't quite right either, because we don't want `abstract_args` in the cache key! This must have bitrotted while JAX+IREE was broken.

dougalm (approved):

dime10:
Hi @mattjj, while using the dynamic shapes feature I ran into a problem of undefined JAXPR variables (specifically dynamic shape variables which are constants in the current JAXPR scope). After some digging it led to me these lines of code in `DynamicJaxprTrace.process_call` from this PR:
```py
    for aval, _ in out_type:
      if type(aval) is DShapedArray:
        shape = [[*consts, *in_tracers][d.val] if type(d) is InDBIdx else
                out_tracers[d.val] if type(d) is OutDBIdx else
                d for d in aval.shape]
        aval = aval.update(shape=tuple(get_referent(d) for d in shape))
      out_tracers.append(DynamicJaxprTracer(self, aval, source_info))
    invars = map(self.getvar, in_tracers)
    constvars = map(self.getvar, map(self.instantiate_const, consts))
    outvars = map(self.makevar, out_tracers)
```

From my understanding, `get_referent`'s explicit purpose is to obtain the "original" definition of a JAXPR constant (whether that be a Python value or a tracer from another level). In this context though, where we are creating tracers, and then variables, for the JAXPR outputs, I don't see how that wouldn't always lead to an error in the variable creation when the shape contains a tracer from an outer scope. Specifically, when this function is invoked:
https://github.com/jax-ml/jax/blob/ef68063497e657d619e5a696e2a636d2efabe855/jax/_src/interpreters/partial_eval.py#L1741-L1745
`self.tracer_to_var[id(d)]` will fail because the tracer `d` is not from the current JAXPR frame.

For reference, this code is still present on main: https://github.com/jax-ml/jax/blob/03c1bf96c6ccb7049b9cd58a0ee7112f1215f408/jax/_src/interpreters/partial_eval.py#L1988

PR #11278, opened 2022-06-28
Add `jvp` rule for `for` primitive

sharadmv:
Tracker: #10982


mattjj (approved):
Nice, you made some really good improvements here.

PR #11287, opened 2022-06-28
random.choice: make compatible with strict promotion

### LINKED ISSUE #11285: Incorrect type promotion in random.choice
As discovered in #11237, `random.choice` throws a type promotion error when x64 is enabled and `p` is `float32`.

The following demonstrates the issue. Executing:

```python
import jax
import jax.numpy as jnp
with jax.numpy_dtype_promotion("strict"):
    jax.random.choice(jax.random.PRNGKey(0), 5, p=jnp.ones(5, dtype=jnp.float32))
```

results in the following type promotion error:

<details>
<summary>Traceback</summary>

```python
  File "<stdin>", line 2, in <module>
  File "/Users/dforemanmackey/src/google/jax/jax/_src/random.py", line 499, in choice
    r = p_cuml[-1] * (1 - uniform(key, shape))
  File "/Users/dforemanmackey/src/google/jax/jax/_src/numpy/lax_numpy.py", line 4586, in deferring_binary_op
    return binary_op(self, other)
  File "/Users/dforemanmackey/src/google/jax/jax/_src/numpy/ufuncs.py", line 80, in fn
    x1, x2 = _promote_args(numpy_fn.__name__, x1, x2)
  File "/Users/dforemanmackey/src/google/jax/jax/_src/numpy/util.py", line 332, in _promote_args
    return _promote_shapes(fun_name, *_promote_dtypes(*args))
  File "/Users/dforemanmackey/src/google/jax/jax/_src/numpy/util.py", line 262, in _promote_dtypes
    to_dtype, weak_type = dtypes._lattice_result_type(*args)
jax._src.dtypes.TypePromotionError: Input dtypes ('float32', 'float64') have no available implicit dtype promotion path when jax_numpy_dtype_promotion=strict. Try explicitly casting inputs to the desired output type, or set jax_numpy_dtype_promotion=standard.
```
</details>

The offending line is: https://github.com/google/jax/blob/de464fcf22c8bf7a2931182f8095bc01530df9fe/jax/_src/random.py#L499

And I think the fix is as simple as:

```diff
- r = p_cuml[-1] * (1 - uniform(key, shape))
+ r = p_cuml[-1] * (1 - uniform(key, shape, dtype=p_cuml.dtype))
```

But I haven't checked in detail!

jakevdp:
Fixes #11285

### DIFF
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -28,12 +28,12 @@ import scipy.stats
 
 import jax
 from jax import core
-from jax import dtypes
 from jax import grad
 from jax import lax
 from jax import numpy as jnp
 from jax import prng
 from jax import random
+from jax._src import dtypes
 from jax._src import test_util as jtu
 from jax import vmap
 from jax.interpreters import xla
@@ -639,6 +639,7 @@ class LaxRandomTest(jtu.JaxTestCase):
   def testChoice(self, dtype, input_range_or_shape, shape, replace, weighted, axis):
     # This is the function API that we test against (note that self.rng().choice differs)
     np_choice = np.random.default_rng(0).choice
+    p_dtype = dtypes._to_inexact_dtype(dtype)
 
     key = self.seed_prng(0)
     is_range = type(input_range_or_shape) is int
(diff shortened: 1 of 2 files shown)


tests/random_test.py
```diff
@@ -639,14 +639,19 @@ def testShuffle(self, dtype):
   def testChoice(self, dtype, input_range_or_shape, shape, replace, weighted, axis):
     # This is the function API that we test against (note that self.rng().choice differs)
     np_choice = np.random.default_rng(0).choice
+    p_dtype = dtypes._to_inexact_dtype(dtype)
```
sharadmv, line 642: Out of curiosity, why is this needed?

sharadmv (approved):
Thanks!

tests/random_test.py
jakevdp: To make the test cover the failure in the attached issue, we needed to use a non-default dtype for `p`. This seemed like a clean way to do it while maintaining coverage

PR #11288, opened 2022-06-28
Disable test for GPU/TPU warning on Mac.

hawkinsp:
We previously disabled the GPU/TPU warning on Mac so the test no longer passes. We don't show the warning because we don't support GPUs or TPUs on Mac.


jakevdp (approved):
Sounds good!

PR #11296, opened 2022-06-28
Expose ml_program dialect for MLIR builder

rsuderman:
We now have an ml_program dialect that describes global variables
including load and store operations. Expose this dialect to allow
exporting variables and constants.


hawkinsp (approved):

jax/_src/lib/mlir/dialects/__init__.py
```diff
@@ -17,6 +17,7 @@
 import jaxlib.mlir.dialects.chlo as chlo
 import jaxlib.mlir.dialects.mhlo as mhlo
 import jaxlib.mlir.dialects.func as func
+import jaxlib.mlir.dialects.ml_program as ml_program
```
hawkinsp, line 20: You'll need to make this import conditional (see below), because JAX releases its python and C++ pieces at a different cadence.

rsuderman: Done.

jax/_src/lib/mlir/dialects/__init__.py
```diff
@@ -17,9 +17,18 @@
 import jaxlib.mlir.dialects.chlo as chlo
 import jaxlib.mlir.dialects.mhlo as mhlo
 import jaxlib.mlir.dialects.func as func
+
+try:
+  import jaxlib.mlir.dialects.ml_program as ml_program
+except (ModuleNotFoundError, ImportError):
+  # TODO(ajcbik,phawkins): make this unconditional when jaxlib > 0.3.7
```
hawkinsp, line 24: Change the condition in the comment to > 0.3.14 and change the todo to `phawkins`. Aart need not do anything here.

rsuderman: Done.

PR #11298, opened 2022-06-28
Jax caches should depend on axis env.

pschuh:


mattjj (approved):
Nice!

cf. #9188. I think this PR will fix #9187.

Two requests:
1. Can we also bring over `test_caches_dont_depend_on_unnamed_axis_env` and lax_control_flow_test.py's version of `test_caches_depend_on_axis_env`? (Also I see #9188 has a change in pmap_test.py, which may be unrelated but seems worth not forgetting about, so I suggest bringing it into this PR also.)
2. Unless we apply `stash_axis_env` to more `jnp` functions (in effect promising that they don't read the axis env, see the description in #9188 for more commentary), we might in principle get a lot more cache misses. Is there anything we can do to de-risk performance impacts up front? Actually I think it's probably fine to just land this and see if anyone notices, and then make something like `stash_axis_env` a public API as needed. But I wanted to ask if you have any other thoughts on this front. WDYT?

PR #11299, opened 2022-06-28
Add JAX debugger tool

sharadmv:
Adds a JAX debugging tool (internal-only for now)

Example usage:
JAX Debugger Demo


mattjj (approved):
Beautiful code, beautiful design. This is going to be so useful...

froystig:
Yes!! Awesome work @sharadmv.

PR #11305, opened 2022-06-29
Unify jax and jaxlib versions.

hawkinsp:
Currently jax and jaxlib have separate version numbers in the JAX source
tree. It is tedious and confusing to bump both version numbers.

However, there is a simpler way to think of things: it is the *source
tree* that is versioned using a single version number, and jax/jaxlib
releases are made using that unified source version number.


skye (approved):
Thanks for doing this!

Issue #10193, opened 2022-04-08
[Better errors] `jax.vmap` with inconsistently-sized kwargs.


patrick-kidger:
```python
import jax
import jax.numpy as jnp

@jax.vmap
def f(x, y):
    return x + y

f(jnp.array([1]), y=jnp.array([1, 2]))
```

produces the rather opaque:

```
  File ".../jax/_src/api.py", line 1499, in _mapped_axis_size
    assert treedef_is_leaf(leaf)
jax._src.traceback_util.UnfilteredStackTrace: AssertionError
```

JAX version 0.3.5, jaxlib version 0.3.2

PR #11311, opened 2022-06-29
vmap: better errors for mismatched axis in keyword arguments

### LINKED ISSUE #8883: hit assertion error in api._get_axis_size
A user hit [this assertion](https://github.com/google/jax/blob/f6f6b4f04cb1cfb64d531eb41172e1783c57fe47/jax/_src/api.py#L1562). We haven't been able to make a minimal repro yet.

### LINKED ISSUE #10465: Unhelpful error messages related to pmap leading dimension requirements
`pmap` splits work based on the leading dimension of its arguments, which must be equal and cannot be larger than the number of devices. As a JAX n00b, I wasn't initially aware of this, thinking that `pmap` would internally split along the leading dimension. This led me to encounter some unhelpful error messages, one of which was triggered by using keyword arguments.  Repro details below:

```
import jax
import jax.numpy as jnp
import numpy as np
```

Here's a very simple computation that scales a (5,) array by a (1,) array's value:

```
x = np.arange(5)
w = np.array([2.])

def scale(x, w):
  return x * w

scale(x, w)
--> array([0., 2., 4., 6., 8.])
```
Note that I'm doing this on a single CPU:

```
jax.local_device_count() 
--> 1
```

Running this computation using `pmap` is inherently broken, because the first dimension of x is 5, which is greater than the number of devices.

So, we'd expect a helpful error message, but I've found that if arguments are supplied by keyword, we don't get one:

```
jax.pmap(scale)(x=x, w=w)
jax/_src/api.py in _mapped_axis_size(tree, vals, dims, name, kws)
   1499   try:
-> 1500     size, = mapped_axis_sizes
   1501     return size

ValueError: too many values to unpack (expected 1)

UnfilteredStackTrace                      Traceback (most recent call last)
6 frames
<ipython-input-7-997e0e638d4f> in <module>()
----> 1 jax.pmap(scale)(x=x, w=w)

jax/_src/traceback_util.py in reraise_with_filtered_traceback(*args, **kwargs)
    161     try:
--> 162       return fun(*args, **kwargs)
    163     except Exception as e:

jax/_src/api.py in cache_miss(*args, **kwargs)
   2011 
-> 2012     out_tree, out_flat = f_pmapped_(*args, **kwargs)
   2013     out_pytree_def = out_tree()

jax/_src/api.py in pmap_f(*args, **kwargs)
   1884   def pmap_f(*args, **kwargs):
-> 1885     p = _prepare_pmap(
   1886         fun, in_axes, out_axes, static_broadcasted_tuple, donate_tuple,

jax/_src/api.py in _prepare_pmap(fun, in_axes, out_axes, static_broadcasted_tuple, donate_tuple, global_arg_shapes, args, kwargs)
   1828       kws=True))
-> 1829   local_axis_size = _mapped_axis_size(
   1830       in_tree, args, in_axes_flat, "pmap", kws=True)

jax/_src/api.py in _mapped_axis_size(tree, vals, dims, name, kws)
   1512       tree, leaf = treedef_children(tree)
-> 1513       assert treedef_is_leaf(leaf)
   1514     # TODO(mattjj,phawkins): add a way to inspect pytree kind more directly

UnfilteredStackTrace: AssertionError

--------------------

AssertionError                            Traceback (most recent call last)
<ipython-input-7-997e0e638d4f> in <module>()
----> 1 jax.pmap(scale)(x=x, w=w)

AssertionError:
```

jakevdp:
Fixes #8883, #10465 and #10193

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1593,10 +1593,15 @@ def _mapped_axis_size(tree, vals, dims, name, *, kws=False):
     # in which case we can produce an error message based on argument indices,
     # or if it has nested containers.
     if kws:
-      # if keyword arguments are included in the tree, we make adapt the error
+      position_only_tree, leaf = treedef_children(tree)
+      if not treedef_is_leaf(leaf):
+        sizes = [x.shape[d] if d is not None else None for x, d in zip(vals, dims)]
+        sizes = tree_unflatten(tree, sizes)
+        raise ValueError(msg.format(f"the tree of axis sizes is:\n{sizes}")) from None
+      # if keyword arguments are included in the tree, we adapt the error
       # message only to be about the positional arguments
-      tree, leaf = treedef_children(tree)
-      assert treedef_is_leaf(leaf)
+      tree = position_only_tree
+
     # TODO(mattjj,phawkins): add a way to inspect pytree kind more directly
     if tree == tree_flatten((0,) * tree.num_leaves)[1]:
       lines1 = [f"arg {i} has shape {np.shape(x)} and axis {d} is to be mapped"
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -2586,6 +2586,18 @@ class APITest(jtu.JaxTestCase):
     ans = vfun(lambda x: x + 1, jnp.arange(3))
     self.assertAllClose(ans, np.arange(1, 4), check_dtypes=False)
 
+  def test_vmap_mismatched_keyword(self):
+    # https://github.com/google/jax/issues/10193
+    @jax.vmap
+    def f(x, y):
+      return x + y
+
+    with self.assertRaisesRegex(
+        ValueError, "vmap got inconsistent sizes for array axes to be mapped:\n"
+        "the tree of axis sizes is:\n"
+        r"\(\(1,\), \{'y': 2\}\)"):
+      f(jnp.array([1]), y=jnp.array([1, 2]))
+
   def test_vmap_mismatched_axis_sizes_error_message_issue_705(self):
     # https://github.com/google/jax/issues/705
     def h(a, b):


jax/_src/api.py
```diff
@@ -1595,8 +1595,16 @@ def _get_axis_size(name: str, shape: Tuple[int, ...], axis: int):
...
       # message only to be about the positional arguments
+      full_tree = tree
       tree, leaf = treedef_children(tree)
-      assert treedef_is_leaf(leaf)
+
+      # Separately ensure that keyword arguments match in size.
+      if not treedef_is_leaf(leaf):
+        sizes = [x.shape[d] if d is not None else None for x, d in zip(vals, dims)]
+        sizes = tree_unflatten(full_tree, sizes)
+        print(repr(msg.format(f"the tree of axis sizes is:\n{sizes}")))
```
sharadmv, line 1605: Remove print

jax/_src/api.py
```diff
@@ -1595,8 +1595,16 @@ def _get_axis_size(name: str, shape: Tuple[int, ...], axis: int):
     if kws:
       # if keyword arguments are included in the tree, we make adapt the error
```
sharadmv, line 1596: Is this comment relevant anymore?

jax/_src/api.py
```diff
@@ -1595,8 +1595,16 @@ def _get_axis_size(name: str, shape: Tuple[int, ...], axis: int):
...
+      full_tree = tree
       tree, leaf = treedef_children(tree)
-      assert treedef_is_leaf(leaf)
+
+      # Separately ensure that keyword arguments match in size.
+      if not treedef_is_leaf(leaf):
+        sizes = [x.shape[d] if d is not None else None for x, d in zip(vals, dims)]
+        sizes = tree_unflatten(full_tree, sizes)
+        print(repr(msg.format(f"the tree of axis sizes is:\n{sizes}")))
+        raise ValueError(msg.format(f"the tree of axis sizes is:\n{sizes}")) from None
```
sharadmv, line 1606: Should this `raise ... from e` like the error above?

sharadmv (approved):
This is a nice QOL change, thanks!

jax/_src/api.py
jakevdp: Yes, it applies to the following line where `tree` is rewritten with a copy that does not include keyword arguments.

jax/_src/api.py
jakevdp: I did this to be consistent with the existing `raise ... from None` used further below.

PR #11313, opened 2022-06-29
[dynamic shapes] revive djax+iree from bitrot

mattjj:
~#11268 is the diffbase.~

Because we hadn't released a jaxlib wheel in a while, the jaxlib and iree wheels got out of sync and so we couldn't run JAX+IREE tests in OSS for a while. (I run them manually because they're not yet covered by CI.) That led to some bit rot as we made changes to the JAX dynamic shapes stuff without updating the (untested) dispatch path.

This PR revives JAX+IREE! It also makes a batch-size-polymorphic MLP execute (i.e. compute numbers), including with autodiff, and adds basic `jnp.arange` / `lax.iota` dynamic shape support (drawn from #9426).

To run JAX+IREE tests, use `JAX_PLATFORMS=iree pytest -n auto tests/api_test.py` or similar.

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -192,7 +192,10 @@ def _device_from_arg_devices(devices: Sequence[Optional[Device]]) -> Optional[De
 def _xla_call_impl(fun: lu.WrappedFun, *args, device, backend, name,
                    donated_invars, inline, keep_unused: bool):
   del inline  # Only used at tracing time
-  arg_specs = unsafe_map(arg_spec, args)
+  if fun.in_type is None:
+    arg_specs = unsafe_map(arg_spec, args)
+  else:
+    arg_specs = [(None, getattr(x, '_device', None)) for x in args]
   compiled_fun = xla_callable(fun, device, backend, name, donated_invars,
                               keep_unused, *arg_specs)
   try:
@@ -283,25 +286,8 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
     in_type = tuple(unsafe_zip(abstract_args, itertools.repeat(True)))
     fun = lu.annotate(fun, in_type)
   else:
-    # Check that the provided abstract_args are consistent with in_type by first
-    # collecting values of axis size arguments, then substituting them in for
-    # DBIdx occurrences.
-    axis_sizes: Dict[core.DBIdx, int] = {}
-    abstract_args_iter = iter(abstract_args)
-    for expected_type, explicit in fun.in_type:
-      if explicit:
-        aval = next(abstract_args_iter)
-        if isinstance(expected_type, core.DShapedArray):
-          # Check the value for any DBIdx variables is consistent.
-          assert all(axis_sizes.setdefault(d1, d2) == d2
-                     for d1, d2 in zip(expected_type.shape, aval.shape)
-                     if type(d1) is core.DBIdx)
-          # Check the type matches after substitution.
-          expected_shape = [axis_sizes.get(d, d) for d in expected_type.shape]  # type: ignore
-          expected_aval = core.ShapedArray(
-              shape=tuple(expected_shape), dtype=expected_type.dtype,
-              weak_type=expected_type.weak_type)
-          assert core.typematch(expected_aval, aval)
+    assert abstract_args == (None,) * len(abstract_args)
+    abstract_args = [aval for aval, _ in fun.in_type]
   with log_elapsed_time(f"Finished tracing + transforming {fun.__name__} "
                         "for jit in {elapsed_time} sec"):
     jaxpr, out_type, consts = pe.trace_to_jaxpr_final2(
@@ -326,7 +312,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
                       if i in kept_var_idx]
     del kept_const_idx
   else:
-    kept_var_idx = set(range(len(abstract_args)))
+    kept_var_idx = set(range(len(fun.in_type)))
 
   nreps = jaxpr_replicas(jaxpr)
   device = _xla_callable_device(nreps, backend, device, arg_devices)
@@ -430,12 +416,10 @@ def jaxpr_has_pmap(jaxpr):
   return False
 
 def jaxpr_has_bints(jaxpr: core.Jaxpr) -> bool:
-  return (any(type(d) is core.Var for v in jaxpr.invars
-              if type(v.aval) is core.DShapedArray for d in v.aval.shape) or
-          any(type(d) is core.Var
+  return (any(type(v.aval) is core.AbstractBInt for v in jaxpr.invars) or
+          any(type(v.aval) is core.AbstractBInt
               for j in itertools.chain([jaxpr], core.subjaxprs(jaxpr))
-              for e in j.eqns for v in itertools.chain(e.invars, e.outvars)
-              if type(v.aval) is core.DShapedArray for d in v.aval.shape))
+              for e in j.eqns for v in e.outvars))
 
 def _prune_unused_inputs(
     jaxpr: core.Jaxpr) -> Tuple[core.Jaxpr, Set[int], Set[int]]:
@@ -545,7 +529,7 @@ def _input_handler(backend: Backend,
   in_avals, which_explicit = util.unzip2(in_type)
   # Check whether we actually need an input_handler.
   needs_implicit = which_explicit and not all(which_explicit)
-  needs_out_handling = any(type(d) is core.InDBIdx for a in out_type or []
+  needs_out_handling = any(type(d) is core.InDBIdx for a, _ in out_type or []
                            if type(a) is core.DShapedArray for d in a.shape)
 
   if not needs_implicit and not needs_out_handling:
@@ -565,7 +549,7 @@ def _input_handler(backend: Backend,
 
   # Precompute which input values are needed for output types.
   inputs_needed_for_out_types = out_type and [
-      d.val for aval in out_type if type(aval) is core.DShapedArray  # type: ignore
+      d.val for aval, _ in out_type if type(aval) is core.DShapedArray  # type: ignore
       for d in aval.shape if type(d) is core.InDBIdx]
 
   def elaborate(explicit_args: Sequence[Any]) -> Tuple[Tuple, Optional[Tuple]]:
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -18,8 +18,8 @@ import functools
 from functools import partial
 import itertools
 import operator
-from typing import (Any, Callable, Dict, Optional, Sequence, Tuple,
-                    List, TypeVar, Union, cast as type_cast)
+from typing import (Any, Callable, Optional, Sequence, Tuple, List, Dict,
+                    TypeVar, Union, cast as type_cast)
 import warnings
 
 import numpy as np
@@ -106,13 +106,20 @@ def _try_broadcast_shapes(
   rank, *others = {len(shape) for shape in shapes}
   if others: return None  # must have consistent rank
   if not rank: return ()  # scalar case
-  result_shape = [-1] * rank
-  for i, sizes in enumerate(zip(*shapes)):
-    non_1s = {d for d in sizes if not core.symbolic_equal_dim(d, 1)}
-    if len(non_1s) > 1:
-      return None  # must have equal sizes other than 1-sized axes
-    result_shape[i] = next(iter(non_1s), 1)
-
+  result_shape = []
+  for ds in unsafe_zip(*shapes):
+    if all(core.same_referent(d, ds[0]) for d in ds[1:]):
+      # if all axes are identical objects, the resulting size is the object
+      result_shape.append(ds[0])
+    else:
+      # if all dims are equal (or 1), the result is the non-1 size (or 1)
+      non_1s = [d for d in ds if not core.symbolic_equal_dim(d, 1)]
+      if not non_1s:
+        result_shape.append(1)
+      elif all(core.symbolic_equal_dim(non_1s[0], d) for d in non_1s[1:]):
+        result_shape.append(non_1s[0])
+      else:
+        return None
   return tuple(result_shape)
 
 def broadcast_shapes(*shapes: Tuple[Union[int, core.Tracer], ...]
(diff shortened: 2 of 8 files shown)


jax/_src/dispatch.py
```diff
@@ -283,25 +286,8 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
...
-                     for d1, d2 in zip(expected_type.shape, aval.shape)
-                     if type(d1) is core.DBIdx)
-          # Check the type matches after substitution.
-          expected_shape = [axis_sizes.get(d, d) for d in expected_type.shape]  # type: ignore
-          expected_aval = core.ShapedArray(
-              shape=tuple(expected_shape), dtype=expected_type.dtype,
-              weak_type=expected_type.weak_type)
-          assert core.typematch(expected_aval, aval)
+    assert abstract_args == (None,) * len(abstract_args)
+    abstract_args = [aval for aval, _ in fun.in_type]
```
mattjj, line 290: The previous code was mistaken because it caused cache misses. That is, the previous code correctly compared the arguments' avals (`abstract_args`, unzipped from `arg_specs`) to the type annotation on the function (`fun.in_type`), but if we pass those `avals` as part of `arg_specs` then we'll get the static shapes of arguments as part of the cache key (i.e. affecting the memoization decorator on this function). So instead when the input type is annotated, and hence we don't need to infer the type from the arguments, we don't pass in the arguments' types (see above, around line 195).

jax/_src/dispatch.py
```diff
@@ -326,7 +312,7 @@ def lower_xla_callable(fun: lu.WrappedFun, device, backend, name,
                       if i in kept_var_idx]
     del kept_const_idx
   else:
-    kept_var_idx = set(range(len(abstract_args)))
+    kept_var_idx = set(range(len(fun.in_type)))
```
mattjj, line 315: We need to keep all arguments whether implicit or explicit. Previously this was buggy in the presence of implicit arguments.

jax/_src/dispatch.py
```diff
@@ -430,12 +416,10 @@ def jaxpr_has_pmap(jaxpr):
   return False
 
 def jaxpr_has_bints(jaxpr: core.Jaxpr) -> bool:
-  return (any(type(d) is core.Var for v in jaxpr.invars
-              if type(v.aval) is core.DShapedArray for d in v.aval.shape) or
-          any(type(d) is core.Var
+  return (any(type(v.aval) is core.AbstractBInt for v in jaxpr.invars) or
```
mattjj, line 419: This check was buggy before. The logic is that we want to check all binders' annotations (i.e. the lambda binders in `jaxpr.invars` and the let binders in each `eqn.outvars`) to see if there are any `AbstractBInt`s, recursing into subjaxprs. That checks every in-scope variable.

jax/_src/dispatch.py
```diff
@@ -545,7 +529,7 @@ def _input_handler(backend: Backend,
   in_avals, which_explicit = util.unzip2(in_type)
   # Check whether we actually need an input_handler.
   needs_implicit = which_explicit and not all(which_explicit)
-  needs_out_handling = any(type(d) is core.InDBIdx for a in out_type or []
+  needs_out_handling = any(type(d) is core.InDBIdx for a, _ in out_type or []
```
mattjj, line 532: Sometimes variables named `out_type` are of type `Tuple[Tuple[AbstractValue, bool], ...]` (aka `core.OutputType`) and sometimes they are of type `Sequence[AbstractValue]` (internally in this file). Either the type of this variable changed or this was a bug before (which would've caused `needs_out_handling` to always be False). Either way, it's consistent/correct now!

jax/_src/lax/lax.py
```diff
@@ -1116,17 +1116,18 @@ def iota(dtype: DType, size: int) -> Array:
   <https://www.tensorflow.org/xla/operation_semantics#iota>`_
   operator.
   """
-  dtype = dtypes.canonicalize_dtype(dtype)
-  size, = canonicalize_shape((size,))
-  return iota_p.bind(dtype=dtype, shape=(size,), dimension=0)
+  return broadcasted_iota(dtype, (size,), 0)
```
mattjj, line 1119: This was just a trivial simplification

jax/_src/lax/lax.py
```diff
@@ -1527,9 +1528,14 @@ def broadcast_mhlo(
       assert len(aval.shape) <= len(aval_out.shape), (aval, aval_out)
       dims = mlir.dense_int_elements(
           range(len(aval_out.shape) - len(aval.shape), len(aval_out.shape)))
-      arg = mhlo.BroadcastInDimOp(
-          mlir.aval_to_ir_type(aval.update(shape=aval_out.shape)), arg,
-          dims).result
+      if any(isinstance(d, ir.Value) for d in aval_out.shape):
```
mattjj, line 1531: At the moment, applications like `add x:f32[] y:f32[n]` are legal in dynamic shape jaxprs. This `broadcast_mhlo` function was written to handle things like `add x:f32[] y:f32[3]` by calling `BroadcastInDimOp`, but to handle cases like `add x:f32[] y:f32[n]` we had to extend it to sometimes call `DynamicBroadcastInDimOp`.

jax/_src/lax/lax.py
```diff
@@ -1543,13 +1549,23 @@ def _nary_lower_mhlo(op: Callable, ctx,
       provided?
   """
   del params
-  aval_out, = ctx.avals_out
-  broadcasted_args = broadcast_mhlo(aval_out, ctx.avals_in, args)
+  avals_in, (aval_out,) = ctx.avals_in, ctx.avals_out
```
mattjj, line 1552: As per the above comment, we needed to update this function to handle lowering applications like `add x:f32[] y:f32[n]`, by making `broadcast_mhlo` call `DynamicBroadcastInDimOp`. But that means `broadcast_mhlo` needs to know the relevant axis sizes, i.e. the `ir.Value` corresponding to the `n` in the aval. Hence this code substitutes `ir.Value`s into the avals passed as arguments to `broadcast_mhlo`. (An alternative would be to make the `broadcast_mhlo` implementation generate `GetAxisSize`s to get the axis sizes it needs from the arguments themselves, but that is more complex code and seems wasteful.)

jax/_src/lax/lax.py
```diff
@@ -4325,11 +4341,49 @@ def _iota_abstract_eval(*, dtype, shape, dimension):
 iota_p.def_impl(partial(xla.apply_primitive, iota_p))
 iota_p.def_abstract_eval(_iota_abstract_eval)
 
-def _iota_lower(ctx, *, dtype, shape, dimension):
-  del dtype, shape
+def _iota_staging_rule(trace, *dyn_shape, dtype, shape, dimension):
```
mattjj, line 4344: This is essentially copied from #9426 (thanks @gnecula), though without the utility functions copied over (yet).

As with `broadcast_in_dim`, because `iota_p` can produce an output with a new dynamic axis size, we need to adapt its staging rule and typecheck rule, rather than rely on the generic implementations in terms of its abstract eval rule.

jax/_src/lax/lax.py
```diff
@@ -4325,11 +4341,49 @@ def _iota_abstract_eval(*, dtype, shape, dimension):
...
+
+def _iota_lower(ctx, *dyn_shape, dtype, shape, dimension):
+  del dtype
   aval_out, = ctx.avals_out
-  return mhlo.IotaOp(mlir.aval_to_ir_type(aval_out),
-                     mlir.i64_attr(dimension)).results
+  if dyn_shape:
+    dyn_shape_ = iter(dyn_shape)
+    shape = [next(dyn_shape_) if d is None else d for d in shape]
+    return mhlo.DynamicIotaOp(
```
mattjj, line 4379: Because MHLO has distinct `IotaOp` and `DynamicIotaOp` operations, we need to switch on which to call here in the MHLO lowering rule.

jax/interpreters/ad.py
```diff
@@ -315,34 +315,33 @@ def process_primitive(self, primitive, tracers, params):
   def process_call(self, call_primitive, f, tracers, params):
     assert call_primitive.multiple_results
     primals, tangents = unzip2((t.primal, t.tangent) for t in tracers)
-    nonzero_tangents, tangent_tree_def = tree_flatten(tangents)
-    nz_tangents = [type(t) is not Zero for t in tangents]
+    which_nz = [     type(t) is not Zero           for t in tangents]
```
mattjj, line 318: The `tangents` here contain symbolic zeros (i.e. `Zero` instances), which because they are not valid jaxtypes must be squeezed out before passed to the call primitive bind. The `traceable` wrapper applied below would ensure they were put back before the primals and tangents were put back together into `JVPTracer`s and dropped into the function being called.

However, that led to a caching issue: each `Zero` instance carries with it an aval, so in the case where we were applying a function with input type `[(i32[], False), (f32[DBIdx(0)], True)]` to the explicit primal argument `f32[N]` with some accompanying `Zero(f32[N])` for some literal value `N`, the pytree metadata would be specialized on a particular value of `N` and hence we wouldn't get cache hits for different values of `N` as we intend.

Actually, the issue was a bit worse: the `Zero`'s aval wouldn't be correct under the call, because it would need to agree with the primal aval of `f32[n]`! Something analogous happens on the output side too.

So it seems better not to propagate `Zero`s' avals through calls. Instead, we just need to propagate where the zeros are (e.g. using Nones, so we can still use the pytree machinery). We can construct the correct agrees-with-the-primal aval on the other side.

So that's what these changes do!

jax/interpreters/mlir.py
```diff
@@ -926,13 +927,19 @@ def write(v: core.Var, node: Sequence[ir.Value]):
...
-          avals_in=map(aval, eqn.invars), avals_out=map(aval, eqn.outvars),
-          tokens_in=tokens_in, tokens_out=None)
+          module_context=eqn_ctx, primitive=eqn.primitive, avals_in=avals_in,
+          avals_out=map(aval, eqn.outvars), tokens_in=tokens_in,
+          tokens_out=None)
+      if config.jax_dynamic_shapes:
+        axis_size_env = {d: read(d)[0] for a in avals_in
+                         if type(a) is core.DShapedArray for d in a.shape
+                         if type(d) is core.Var}
+        rule_ctx = rule_ctx.replace(axis_size_env=axis_size_env)
```
mattjj, line 942: As mentioned above (in the context of the changes to `mhlo_broadcast`), we extend the `rule_ctx` to include an `axis_size_env: Dict[core.Var, ir.Value]` with keys corresponding to the `Var` instances that occur in the shapes of the `avals_in`. (We might need to add those in `avals_out` also...)

jax/_src/dispatch.py
```diff
@@ -192,7 +192,10 @@ def _device_from_arg_devices(devices: Sequence[Optional[Device]]) -> Optional[De
 def _xla_call_impl(fun: lu.WrappedFun, *args, device, backend, name,
                    donated_invars, inline, keep_unused: bool):
   del inline  # Only used at tracing time
-  arg_specs = unsafe_map(arg_spec, args)
+  if fun.in_type is None:
+    arg_specs = unsafe_map(arg_spec, args)
```
sharadmv, line 196: Why unsafe_map?

sharadmv, line 196: Why unsafe_map?

jax/_src/dispatch.py
```diff
@@ -430,12 +416,10 @@ def jaxpr_has_pmap(jaxpr):
   return False
 
 def jaxpr_has_bints(jaxpr: core.Jaxpr) -> bool:
-  return (any(type(d) is core.Var for v in jaxpr.invars
-              if type(v.aval) is core.DShapedArray for d in v.aval.shape) or
-          any(type(d) is core.Var
+  return (any(type(v.aval) is core.AbstractBInt for v in jaxpr.invars) or
+          any(type(v.aval) is core.AbstractBInt
               for j in itertools.chain([jaxpr], core.subjaxprs(jaxpr))
```
sharadmv, line 421: Could be `[jaxpr, *core.subjaxprs(jaxpr)]`

jax/_src/lax/lax.py
sharadmv: Nice!

jax/_src/lax/lax.py
```diff
@@ -1116,17 +1116,18 @@ def iota(dtype: DType, size: int) -> Array:
...
-  dtype = dtypes.canonicalize_dtype(dtype)
-  size, = canonicalize_shape((size,))
-  return iota_p.bind(dtype=dtype, shape=(size,), dimension=0)
+  return broadcasted_iota(dtype, (size,), 0)
 
 def broadcasted_iota(dtype: DType, shape: Shape, dimension: int) -> Array:
   """Convenience wrapper around ``iota``."""
   dtype = dtypes.canonicalize_dtype(dtype)
   shape = canonicalize_shape(shape)
+  dynamic_shape = [d for d in shape if isinstance(d, core.Tracer)]
```
sharadmv, line 1125: This feels like it'll  be a pattern of sorts (decoupling a shape into static/dynamic portions and then stitching it back together in the impl).

It could be factored out into a pair of helper functions (one that splits up the shape and one that stitches it back together). Not sure if it's premature to do that though.

jax/_src/lax/lax.py
```diff
@@ -4325,11 +4341,49 @@ def _iota_abstract_eval(*, dtype, shape, dimension):
...
+    return trace.default_process_primitive(iota_p, (), params)
+  source_info = source_info_util.current()
+  ds = iter(dyn_shape)
+  out_shape_for_tracer = [next(ds) if d is None else d for d in shape]
+  aval = core.DShapedArray(tuple(out_shape_for_tracer), dtype, False)
+  out_tracer = pe.DynamicJaxprTracer(trace, aval, source_info)
+  invars = [trace.getvar(d) for d in dyn_shape]
+  eqn = pe.new_jaxpr_eqn(invars, [trace.makevar(out_tracer)],
+                         iota_p, params, core.no_effects, source_info)
+  trace.frame.eqns.append(eqn)
```
sharadmv, line 4356: should use `frame.add_eqn` (there are no effects here but this method is more future-proof)

sharadmv (commented):
Awesome! Thanks for the helpful comments

jax/_src/dispatch.py
mattjj: I was just keeping it how it was, but I think it's like this for speculative performance reasons. The only place `arg_specs` is consumed is just below.

jax/_src/lax/lax.py
mattjj: It's a good idea! You're totally right. Actually @gnecula did that in #9426, so we should land that. I could copy them into this PR, but I figured we might land that one imminently.

jax/_src/lax/lax.py
mattjj: Good idea, done.

jax/_src/dispatch.py
mattjj: I wrote it this way b/c in theory `itertools.chain` can be more memory efficient, but OTOH that seems like not very important here. WDYT?

jax/_src/lax/lax.py
sharadmv: Okay then could you leave a TODO?

jax/_src/dispatch.py
sharadmv: SGTM

jax/_src/dispatch.py
sharadmv: SGTM

jax/_src/lax/lax.py
mattjj: Good idea, will do.

jax/_src/lax/lax.py
gnecula: I did manage to land #9426 but note that I am still unable to run the IREE tests, presumably due to some of the same issues as here

sharadmv (approved):

PR #11343, opened 2022-07-01
Formated the file and updated to latest versions

prasadjivane:
setup updated versions

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,10 +39,10 @@ setup(
     package_data={"jax": ["py.typed"]},
     python_requires=">=3.7",
     install_requires=[
-        "absl-py>=1.1.0",
-        "numpy<1.23.0",
+        "absl-py<1.1.0",
+        "numpy>=1.19.0",
         "opt_einsum<3.3.0",
-        "scipy>=1.8.0",
+        "scipy>=1.5.0",
         "typing_extensions>=4.3.0",
         "etils[epath]",
     ],
--- a/setup.py
+++ b/setup.py
@@ -14,69 +14,71 @@
 
 from setuptools import setup, find_packages
 
-_current_jaxlib_version = '0.3.14'
+_current_jaxlib_version = "0.3.14"
 # The following should be updated with each new jaxlib release.
-_latest_jaxlib_version_on_pypi = '0.3.14'
-_available_cuda_versions = ['11']
-_default_cuda_version = '11'
-_available_cudnn_versions = ['82', '805']
-_default_cudnn_version = '82'
-_libtpu_version = '0.1.dev20220627'
+_latest_jaxlib_version_on_pypi = "0.3.14"
+_available_cuda_versions = ["11"]
+_default_cuda_version = "11"
+_available_cudnn_versions = ["82", "805"]
+_default_cudnn_version = "82"
+_libtpu_version = "0.1.dev20220627"
 
 _dct = {}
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,69 +14,71 @@
...
+    python_requires=">=3.7",
     install_requires=[
-        'absl-py',
-        'numpy>=1.19',
-        'opt_einsum',
-        'scipy>=1.5',
-        'typing_extensions',
-        'etils[epath]'
+        "absl-py>=1.1.0",
+        "numpy<1.23.0",
```
hawkinsp, line 43: This constraint doesn't look correct to me. e.g., JAX does work with NumPy 1.23. All we require is NumPy >= 1.19 at present.

Further, we have a policy about this: https://jax.readthedocs.io/en/latest/deprecation.html

setup.py
```diff
@@ -14,69 +14,71 @@
...
+    package_data={"jax": ["py.typed"]},
+    python_requires=">=3.7",
     install_requires=[
-        'absl-py',
-        'numpy>=1.19',
-        'opt_einsum',
-        'scipy>=1.5',
-        'typing_extensions',
-        'etils[epath]'
+        "absl-py>=1.1.0",
```
hawkinsp, line 42: Is this version of absl-py actually needed? (It's probably not a bad idea to set a minimum version, though, since we will most likely accidentally depend on absl-py features at some point in the latest version.)

setup.py
```diff
@@ -14,69 +14,71 @@
...
+        "numpy<1.23.0",
+        "opt_einsum<3.3.0",
```
hawkinsp, line 44: Is there a reason to add an `opt_einsum` version constraint?

setup.py
```diff
@@ -14,69 +14,71 @@
...
+        "scipy>=1.8.0",
```
hawkinsp, line 45: This constraint is too strict. The choice of 1.5 as a minimum version is intentional: while not documented as a policy, we chose it to be released around the same time as the minimum NumPy version.

hawkinsp (changes requested):
As a general rule, please do not mix reformatting changes with functional changes in the same PR. It pollutes the git history. If you want to make reformatting changes, I encourage doing them in their own PRs. (That said, I'm not sure the reformatting here buys much: it mostly looks like change for its own sake.)

The version constraints changes would not be correct as is.

hawkinsp:
Please revert the formatting changes.

setup.py
```diff
@@ -14,69 +14,71 @@
...
     install_requires=[
-        'absl-py',
-        'numpy>=1.19',
-        'opt_einsum',
-        'scipy>=1.5',
-        'typing_extensions',
-        'etils[epath]'
+        "absl-py<1.1.0",
+        "numpy>=1.19.0",
+        "opt_einsum<3.3.0",
```
hawkinsp, line 44: Revert the opt_einsum change.

setup.py
```diff
@@ -14,69 +14,71 @@
...
+        "scipy>=1.5.0",
+        "typing_extensions>=4.3.0",
```
hawkinsp, line 46: In fact, this is the only change that makes any sense to me (specify a minimum version on typing_extensions, but I wonder if it's even worth doing...)

hawkinsp:
Closing because I don't think the changes here should be merged.

PR #11367, opened 2022-07-05
Properly count sublevels when tracing xmap body

apaszke:
Otherwise it can lead to tracer leak errors. I'm not a 100% sure how
this works out, because the sublevel counting has changed since I read
it previously. This replicates the changes applied to
DynamicJaxprTrace.process_map since I last looked at it.


mattjj (approved):
This change makes sense! This is the point at which we're running the WrappedFun, which means we're "incrementing the indent" in the jaxpr.

Issue #11075, opened 2022-06-12
Generic functions with jax arrays, and incoherent `issubclass` and `isinstance` for `jnp.ndarray`


Gattocrucco:
**TLDR:** The Python type dispatch mechanism for generic functions does not work within jax tracing with `jnp.ndarray` as type. Making `issubclass(Tracer, jnp.ndarray)` come out true is still not sufficient.

I often have to work around the fact that jax arrays cannot be dealt with by numpy ufuncs because I have to write functions which accept both kind of arrays. I tried to use the Python standard [dispatch mechanism](https://docs.python.org/3.10/library/functools.html#functools.singledispatch) but failed when tracing.

```python
import functools

import jax
import jax.numpy as jnp

@functools.singledispatch
def f(_):
    print('failure')

@f.register
def _(_: jnp.ndarray):
    print('success')

f(jnp.zeros(1)) # -> success

def g(x):
    f(x)
    return x

jax.jacobian(g)(jnp.zeros(1)) # -> failure
```

This was unexpected because I used many times `isinstance(x, jnp.ndarray)` in traced code successfully. So I found out that `jax.core.Tracer` is not a subclass of `jnp.ndarray`, but the instance checking is overridden:

```python
class MyTracer(jax.core.Tracer):
    __init__ = lambda _: None
    aval = property(lambda _: jax.core.UnshapedArray(object))

print(isinstance(MyTracer(), jnp.ndarray)) # -> True
print(issubclass(jax.core.Tracer, jnp.ndarray)) # -> False
```

The code for `type(jnp.ndarray)` is:

```python
class ArrayMeta(abc.ABCMeta):
  """Metaclass for overriding ndarray isinstance checks."""

  def __instancecheck__(self, instance):
    # Allow tracer instances with avals that are instances of UnshapedArray.
    # We could instead just declare Tracer an instance of the ndarray type, but
    # there can be traced values that are not arrays. The main downside here is
    # that isinstance(x, ndarray) might return true but
    # issubclass(type(x), ndarray) might return false for an array tracer.
    try:
      return (hasattr(instance, "aval") and
              isinstance(instance.aval, core.UnshapedArray))
    except AttributeError:
      super().__instancecheck__(instance)
```

[...]

PR #11370, opened 2022-07-05
Maintain invariance to JIT and other transform when the function behavior depends on `__class__` of inputs, e.g. `functools.singledispatch`

YouJiacheng:
Solve #11075 
We override `Tracer.__class__` to return `aval.__class__` to allow tracer instances with `avals` that are instances of `UnshapedArray`
This is better than overriding ndarray isinstance checks, which can not support `functools.singledispatch` based on `x.__class__`.
Theoretically, well written code should use `x.__class__` rather than type(x) except for capturing proxy type(e.g. Tracer). Unfortunately, many users prefer `type(x)`.

### DIFF
--- a/jax/_src/numpy/ndarray.py
+++ b/jax/_src/numpy/ndarray.py
@@ -22,23 +22,7 @@ from jax.experimental import array
 import numpy as np
 
 
-class ArrayMeta(abc.ABCMeta):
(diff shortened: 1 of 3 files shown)


YouJiacheng:
@jakevdp Hi - Would you like to review and pull down this PR?

PR #11372, opened 2022-07-05
[shape_poly] Add special-case handling for operations with polynomials

gnecula:
jax2tf shape-poly works with polynomials representing the dimension sizes. We want to know how often we truly
need the power of polynomials. We know that we need to do often shape equality checking, e.g., that the reshape of
`(a, b, 4, 3)` into `(12, b, a)` is a valid one. Polynomials help here, because we can normalize easily the shape sizes as `12 * a * b`. 

I wanted to know how far we can get if we disable the use of polynomials. Can we still pass all the tests that check that `vmap(f)` is batch polymorphic? Fortunately, we have a large suite of 560 such tests for practically all the JAX primitives.

The first step was to add an internal constant `shape_poly.EXPERIMENTAL_STRICT_EQ` that makes equality be defined only for identical polynomials. With this change alone, 366 tests fail.

Then I looked at all the failures and I tried to add changes to JAX that make it unnecessary to do more powerful equality checking. I needed the following changes:

 * ensure that we create the same identical polynomial from two separate occurrences of a dimension variable (use caching)
 * add special casing for `d + 0` and `d * 1` to ensure we preserve dimension identity
 * add special casing for `dilate(d, 1) == d`
 * add special casing for `stride(d, window_size=1, window_stride=1) == d`
 
With these changes now only 21 tests fail, mostly for reshape and for `random` primitives (which use reshape internally). A typical error message is `Cannot divide evenly the sizes of shapes (b0, 3, 4, 5) and (b0, 4, 15)`. Another similar error message arises for `reshape` when one dimension is `-1`: `Cannot divide evenly the sizes of shapes (b0, 3, 4, 5) and (b0, 2, -1)`.

To handle these, I added a more elaborate implementation of `core.divide_shape_sizes` (which is used to check that
`reshape` is well-formed and to compute the missing dimension size of `reshape(-1)`) to detect the special case when all the non-constant dimensions in the divisor appear in the dividend. 

This code is somewhat ugly, because it tries to handle the cases when there are multiple occurrences of a dimension, and also to handle the case for `reshape(-1)`.

With these changes ``a single`` test fails, for `random_gamma`, which does some fancy `reshape` in its [impl rule](https://github.com/google/jax/blob/0719f986aaeda3ba216416b6b1243d9f815fefda/jax/_src/random.py#L990). The error message is `Cannot divide evenly the sizes of shapes (b0, 3, 2) and (3*b0, 2)`

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -739,7 +739,6 @@ def isrealobj(x):
   return not iscomplexobj(x)
 
 
-
 @_wraps(np.reshape, lax_description=_ARRAY_VIEW_DOC)
 def reshape(a, newshape, order="C"):
   _stackable(a) or _check_arraylike("reshape", a)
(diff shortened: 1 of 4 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -749,9 +749,20 @@ def _compute_newshape(a, newshape):
   except: iterable = False
   else: iterable = True
   newshape = core.canonicalize_shape(newshape if iterable else [newshape])
-  return tuple(- core.divide_shape_sizes(np.shape(a), newshape)
-               if core.symbolic_equal_dim(d, -1) else d
-               for d in newshape)
+  # One dimension in newshape can be -1
+  newshape_has_m1 = False
```
gnecula, line 753: This code can be simplified if we decide that we don't support `-1` dimensions in `reshape`. The user typically has a reason to believe that the reshape is correct, and hence it could write the actual missing dimension.

jax/core.py
```diff
@@ -1680,10 +1678,47 @@ def diff_shape(s1: Shape, s2: Shape) -> Shape:
...
+        prod_constant *= operator.index(d)
+      else:
+        non_constant.append(d)
+    return prod_constant, non_constant
+  s1_const, s1_non_const = split_constants(s1)
+  s2_const, s2_non_const = split_constants(s2)
+  if s1_const == s2_const == 0:
+    return 1
+  for s2_d in s2_non_const:
+    def remove_first(s1_non_const):
```
gnecula, line 1702: This code could be simplified if we decide we don't support reshape with multiple occurrences of a dimension and if we don't support `reshape` with `-1` dimensions. In that case, the only use for `divide_shape_sizes` really is to compare the shape sizes for equality.

PR #11380, opened 2022-07-06
Fix ConcretizationError in nested calls/subjaxprs.

LenaMartens:
```
@jax.jit
def f(x, y):
  if y:
    return x

@jax.jit
def g(x):
  return f(x, True)

g()
```

Before:
```
[jax/interpreters/partial_eval.py] in arg_info_pytree(fn, in_tree, has_kwargs, flat_pos)
   1840                     flat_pos: List[int]) -> str:
   1841   dummy_args = [False] * in_tree.num_leaves
-> 1842   for i in flat_pos: dummy_args[i] = True
   1843   if has_kwargs:
   1844     args, kwargs = tree_unflatten(in_tree, dummy_args)

IndexError: list assignment index out of range
```

After:
```
jax._src.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected:
Traced<ShapedArray(bool[])>with<DynamicJaxprTrace(level=0/2)> 
This concrete value was not available in Python because it depends on the value of the argument 'y'. 
```

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -1526,11 +1526,13 @@ def move_binders_to_back(closed_jaxpr: ClosedJaxpr, to_move: Sequence[bool]
   return move_binders_to_front(closed_jaxpr, map(op.not_, to_move))
 
 class DynamicJaxprTracer(core.Tracer):
-  __slots__ = ['aval']
+  __slots__ = ['aval', '_debug_info']
 
   def __init__(self, trace, aval, line_info=None):
     self._trace = trace
     self._line_info = line_info
+    # Needed for UnexpectedTracerError.
+    self._debug_info = self._trace.frame.debug_info
     self.aval = aval
 
   def full_lower(self):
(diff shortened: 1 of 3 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -1818,7 +1816,11 @@ def process_call(self, call_primitive, f, explicit_tracers, params):
     in_tracers = [*implicit_tracers, *explicit_tracers]
     # TODO(mattjj): check in_tracers are consistent with f.in_type annotation
     with core.new_sublevel():
-      jaxpr, out_type, consts = trace_to_subjaxpr_dynamic2(f, self.main)
+      # TODO(lenamartens): fill out debug_info here.
```
LenaMartens, line 1819: In a first iteration, I kept debug_info to None here, to be filled out later. This means in the nested case we don't get debug_info and leave that to a future change. However this regressed the `test_concrete_error_because_const` test cases, I think because the call to `jnp.add` then sets `debug_info` to None.

So I think we should either:
a) submit this and skip those tests, to be re-enabled later once we thread through debug_info properly
b) figure out how to populate debug_info here (maybe it's easy?). I tried, but couldn't really figure it out,

WDYT?

froystig (approved):
Ended up doing a semi-interactive review of this with @LenaMartens (resulting in the latest pushed version).

jax/interpreters/partial_eval.py
```diff
@@ -1820,9 +1820,13 @@ def process_call(self, call_primitive, f, explicit_tracers, params):
...
     # TODO(mattjj): check in_tracers are consistent with f.in_type annotation
     with core.new_sublevel():
+      # TODO(lenamartens): Make call_primitive name -> API function name mapping.
+      # (currently this will display eg. 'xla_call' instead of `jit`)
+      dbg = debug_info_final(f, call_primitive.name)
       if config.jax_check_tracer_leaks or not config.jax_experimental_subjaxpr_lowering_cache:
-        jaxpr, out_type, consts = trace_to_subjaxpr_dynamic2(f, self.main)
+        jaxpr, out_type, consts = trace_to_subjaxpr_dynamic2(f, self.main, debug_info=dbg)
       else:
+        # TODO(lenamartens): add debug_info to cached jaxpr?
```
LenaMartens, line 1829: I can figure out this TODO in this change too, up to you.

froystig (approved):

PR #11382, opened 2022-07-06
Remove deprecated function jax.tree_util.tree_multimap

jakevdp:
Finishes the deprecation in #5746


jakevdp:
Pulling in for testing - I think we'll need to update some downstream calls of the deprecated function.

skye (approved):

PR #10970, opened 2022-06-03
[ROCm] Enable XLA python callback on ROCm.

rsanthanam-amd:
/cc @hawkinsp

PR #11383, opened 2022-07-06
Enable CustomCall implementation on GPU

YouJiacheng:
https://github.com/google/jax/commit/b93706638f2a6c5f3a0270fcacf3ee80e117ad2f said HCB are waiting for the GPU implementation.
According to
https://github.com/google/jax/commit/b80d7195f695b728a91f63ab2d2b6b09692f2064 and https://github.com/google/jax/pull/10970 GPU implementation seems available after jaxlib-0.3.11

### DIFF
--- a/tests/host_callback_test.py
+++ b/tests/host_callback_test.py
@@ -2456,6 +2456,8 @@ class HostCallbackCallTest(jtu.JaxTestCase):
                                    expected_exc_txt: str):
     """Calls thunk() and checks for expected exceptions.
     """
+    if not FLAGS.jax_host_callback_outfeed:
+      raise SkipTest("TODO: implement error handling for customcall")
     if jtu.device_under_test() == "cpu":
       # On CPU the runtime crashes, and the tests are all aborted
       raise SkipTest("TODO: CPU runtime crashes on unexpected infeed")
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -537,6 +537,8 @@ def _inline_host_callback() -> bool:
 
 
 def _use_outfeed(platform: str) -> bool:
+  if jaxlib.version >= (0, 3, 15):
+    return platform == "tpu" or FLAGS.jax_host_callback_outfeed
   return (platform in ("tpu", "gpu", "cuda", "rocm") or FLAGS.jax_host_callback_outfeed)
 
 xops = xla_client._xla.ops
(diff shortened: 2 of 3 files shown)


sharadmv:
Thank you for the change! It currently looks good to me but I'll run the internal tests and see if any changes will be needed.

YouJiacheng:
It seems that there is a performance issue: https://github.com/google/jax/issues/11371#issuecomment-1176509758

sharadmv:
It looks like we're doing an unnecessary CPU copy, I'll work on a fix.

sharadmv (changes requested):
The `host_callback_custom_call_test_gpu` fails at the moment because of how it assumes errors will be thrown (see `test_call_error_callback*`).

You will need to update that for tests to pass. Alternatively I can update the tests when I pull down the PR.

YouJiacheng:
Done. All three tests using `helper_check_callback_errors` will raise error within `_outside_call_run_callback` and cannot be handled with CustomCall implementation.

sharadmv:
The `CustomCall` implementation actually *does* raise a Python error using XLA's CustomCallStatus. We should look for those errors instead.

sharadmv:
The performance issue you found has been fixed at head btw: https://github.com/tensorflow/tensorflow/commit/684199fb48754a1bc400dea6fd2bdbab859ee656. You might want to bump the jaxlib guard to 0.3.15 so we avoid using the slower implementation.

YouJiacheng:
I think error handling for CustomCall implementation should be another PR.
jaxlib version guard bumped.

YouJiacheng:
BTW, is the `mhlo.GetTupleElementOp` in `mlir.emit_python_callback` necessary, even in single result case?

sharadmv:
I suppose not but I'm unsure if there are any performance implications

tests/host_callback_test.py
```diff
@@ -2456,6 +2456,8 @@ def helper_check_callback_errors(self, thunk: Callable,
                                    expected_exc_txt: str):
     """Calls thunk() and checks for expected exceptions.
     """
+    if not FLAGS.jax_host_callback_outfeed:
+      raise SkipTest("TODO: implement error handling for customcall")
```
sharadmv, line 2460: Could you create an issue for this and link it in a comment?

sharadmv (approved):

tests/host_callback_test.py
YouJiacheng: I will create an issue. But there are 3 "replicas" of this TODO(actually I copied these 2 lines from other testcase), add a comment for each seems too verbose.

sharadmv:
FYI this was rolled back because of internal failures. I will work on fix to roll forward.

PR #11387, opened 2022-07-06
[dynamic-shapes] revive basic bounded int machinery, add tests

mattjj:
Bounded integer types are useful for lowering dynamic shape computations to XLA (though they also work with IREE). They're also necessary for the padded-and-masked raggedness representations we need when dynamic shape computations are transformed with vmap.

This PR is a first step in reviving some bounded int machinery. The next steps are to add some tests for end-to-end examples, and then to add the raggedness features.

In addition to revival, we're also making some minor changes to past BInt experiments: most importantly, we're going to maintain the invariant that _the physical representation of an array is a function only of its type_. In particular, when an array has a BInt axis size, that means the physical buffer representation is effectively padded out along that axis to the size of the BInt's bound.

There's a new `PaddedArray` type here (to be extended later to handle raggedness, and arrays with BInt element type), which is basically the runtime data representation of an array with BInts as axis sizes.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1440,6 +1440,7 @@ def unop(result_dtype, accepted_dtypes, name):
                             weak_type_rule=weak_type_rule)
   batching.defvectorized(prim)
   masking.defvectorized(prim)
+  pe.padding_rules[prim] = lambda _, __, x, **kw: [prim.bind(x, **kw)]
   return prim
 standard_unop = partial(unop, _identity)
 _attrgetter = lambda name: lambda x, **kwargs: getattr(x, name)
@@ -1515,6 +1516,7 @@ def naryop(result_dtype, accepted_dtypes, name):
                             weak_type_rule=weak_type_rule)
   batching.defbroadcasting(prim)
   masking.defnaryop(prim)
+  pe.padding_rules[prim] = lambda _, __, *xs, **kw: [prim.bind(*xs, **kw)]
   return prim
 standard_naryop = partial(naryop, _input_dtype)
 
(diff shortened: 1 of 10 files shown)


jax/core.py
```diff
@@ -1269,77 +1253,9 @@ def shape(self):
     raise TypeError(msg)
 
 
-# We have a convention of reusing AbsractValues as types, in particular reusing
```
mattjj, line 1272: This code just moved to elsewhere in the fille.

jax/_src/lax/lax.py
```diff
@@ -4425,12 +4440,46 @@ def _iota_lower(ctx, *dyn_shape, dtype, shape, dimension):
...
+  if len(eqn.params['shape']) > 1:
+    printed_params['dimension'] = eqn.params['dimension']
+  lhs = core.pp_vars(eqn.outvars, context, print_shapes=settings.print_shapes)
+  rhs = [pp.text(eqn.primitive.name),
+         core.pp_kv_pairs(sorted(printed_params.items()), context, settings),
+         pp.text(" ") + core.pp_vars(eqn.invars, context)]
+  annotation = (source_info_util.summarize(eqn.source_info)
+                if settings.source_info else None)
+  return [lhs, pp.text(" = ", annotation=annotation), *rhs]
+# core.pp_eqn_rules[iota_p] = _iota_pp_rule
```
sharadmv, line 4454: Did you want to leave this uncommented?

sharadmv (approved):

jax/_src/lax/lax.py
mattjj: I wanted to leave it commented out actually; there are a few of these rules which I use when debugging, but I don't yet want to do the work to land it (since there are probably CI tests in Google which depend on jaxpr printing).

PR #11388, opened 2022-07-06
Ensure boolean scalar weak types are treated consistently

### LINKED ISSUE #11377: `unwrap` will return different value when calling `value_and_grads`
`unwrap` will return different value when calling `value_and_grads`. 

```py
import jax
def fn(inp):
    axis = -2
    discont = 2
    period = 1
    return jax.numpy.unwrap(inp, axis=axis, discont=discont, period=period).sum()

mykey = jax.random.PRNGKey(44093535)
inp = jax.random.uniform(mykey, [2, 3, 4], jax.numpy.bfloat16, minval=-128, maxval=127)

res1 = fn(inp)
res2, _ = jax.value_and_grad(fn, (0))(inp)
print(res1)
print(res2)
# -17.0
# -20.0
```

The reason lies in `mod` with `True` as input, which will give differnt value with or without `jit`

```py
def f(dd, period):
    interval = period / 2
    ddmod = jax.numpy.mod(dd + interval, period) - interval
    return ddmod

mykey = jax.random.PRNGKey(44093535)
inp = jax.random.uniform(mykey, [2, 3, 4], jax.numpy.bfloat16, minval=-128, maxval=127)
dd = jax.numpy.diff(inp, axis=-2)
print(f(dd, True))
print(jax.jit(f)(dd, True))
```

```
[[[-0.5 -0.5 0 -0.5]
  [-0.5 -0.5 0 0]]

 [[0 -0.5 0 -0.5]
  [-0.5 0 -0.5 -0.5]]]

[[[-0.5 -0.5  0.   0. ]
  [ 0.   0.   0.   0. ]]

 [[ 0.   0.   0.  -0.5]
  [ 0.   0.  -0.5 -0.5]]]
```

It would be better to convert the `period` argument into `float` at first in `unwrap` to avoid such cases.

jakevdp:
Fixes https://github.com/google/jax/issues/11377#issuecomment-1176781291

Confirmed that the examples in that issue work as expected with the change. Also the core issue is that these outputs now match:
```python
In [1]: import jax

In [3]: jnp.array(True)
Out[3]: DeviceArray(True, dtype=bool)

In [4]: jax.jit(jnp.array)(True)
Out[4]: DeviceArray(True, dtype=bool)
```
I think the reason we went this long without noticing the issue is because it only really comes into play with functions that do type promotion between boolean scalar inputs and other weakly-typed scalar values, which I suspect is fairly rare.

### DIFF
--- a/jax/interpreters/xla.py
+++ b/jax/interpreters/xla.py
@@ -275,7 +275,10 @@ def abstractify(x) -> core.AbstractValue:
   raise TypeError(f"Argument '{x}' of type '{type(x)}' is not a valid JAX type")
 
 def _make_abstract_python_scalar(typ, val):
-  return ShapedArray((), dtypes._scalar_type_to_dtype(typ, val), weak_type=True)
+  # Note: all python scalar types are weak except bool, because bool only
+  # comes in a single width.
+  return ShapedArray((), dtypes._scalar_type_to_dtype(typ, val),
+                     weak_type=typ is not bool)
 
 pytype_aval_mappings: Dict[Any, Callable[[Any], core.AbstractValue]] = {}
 for t in device_array.device_array_types:
--- a/tests/dtypes_test.py
+++ b/tests/dtypes_test.py
@@ -329,6 +329,23 @@ class TestPromotionTables(jtu.JaxTestCase):
     else:
       self.assertIs(jax_type, np.dtype(bool))
 
+  @parameterized.named_parameters(
+      {"testcase_name": f"_{typ}", "typ": typ}
+       for typ in [bool, int, float, complex])
+  def testScalarWeakTypes(self, typ):
+    # Regression test for https://github.com/google/jax/issues/11377
+    val = typ(0)
+
+    result1 = jnp.array(val)
+    result2 = jax.jit(jnp.array)(val)
+    self.assertEqual(result1.aval, result2.aval)
+
+    with jax.numpy_dtype_promotion('standard'):
+      f = lambda x: x / 2
+      result1 = jnp.array(f(val))
+      result2 = jax.jit(f)(val)
+    self.assertEqual(result1.aval, result2.aval)
+
   def testResultTypeNone(self):
     # This matches the behavior of np.result_type(None) => np.float_
     self.assertEqual(dtypes.result_type(None), dtypes.canonicalize_dtype(dtypes.float_))


jax/_src/dtypes.py
```diff
@@ -249,6 +249,7 @@ def issubdtype(a, b):
 issubsctype = np.issubsctype
 
 # Enumeration of all valid JAX types in order.
```
skye, line 251: In what order?? (Not this change, so feel free to ignore)

jax/_src/dtypes.py
```diff
@@ -412,7 +413,7 @@ def is_weakly_typed(x):
   try:
     return x.aval.weak_type
   except AttributeError:
-    return type(x) in _weak_types
+    return type(x) in _python_scalar_types
```
skye, line 416: Do you need the `any()` clause on L290 given it checks the same thing here? Maybe that's a performance optimization?

jax/_src/dtypes.py
```diff
@@ -249,6 +249,7 @@ def issubdtype(a, b):
 issubsctype = np.issubsctype
 
 # Enumeration of all valid JAX types in order.
+_python_scalar_types = [bool, int, float, complex]
 _weak_types = [int, float, complex]
```
skye, line 253: When do you need to separate between weak and scalar types? Like what's special about bool compared to other weak types?

skye (commented):
I don't fully understand this change so I'm not approving! Feel free to answer my questions or find a different reviewer :P

jax/_src/dtypes.py
jakevdp: No, in L290 it's checking if `val is typ`, wheras here it's checking if `type(val) == typ`. You have to be careful about these things because numpy dtype objects have overloaded equality to be annoyingly polymorphic (e.g. `np.dtype('float64') == float` evaluates to `True`). Here since `type(x)` will never be a `np.dtype` object, we don't have to use object identity, and we can use the more idiomatic object equality.

jax/_src/dtypes.py
jakevdp: I think the comment was meant to reflect that some of the code below depends on the order of these enumerations.

jax/_src/dtypes.py
jakevdp: What's special about `bool` is that there is only a single `bool` width, and so there's no need to separate a strong bool from a weak bool in the context of type promotion. But this bug brought to light that we *do* need to treat boolean Python scalars as weakly-typed if we want type promotion to be invariant to JIT

jax/_src/dtypes.py
skye: :open_mouth:

jax/_src/dtypes.py
jakevdp: I expanded the comment and moved the new list above to make this more clear.

jax/_src/dtypes.py
skye: Consider adding this in a comment! Or not, in order to confuse our enemies

skye (approved):

jax/_src/dtypes.py
YouJiacheng: Should L292 be consistent with L410-411?
It might be better to add a util function like "test_membership_by_identity"?

jax/_src/dtypes.py
YouJiacheng: @jakevdp Why not make it simpler? Would adding weak bool significantly affect performance?
It seems that originally "weak type" means "bitwidth unspecified", as they named `i_, f_, c_", but currently weak type changes type promotion behavior? As a result, single bitwidth doesn't mean there is no need to separate a weak bool from strong bool?

jax/_src/dtypes.py
YouJiacheng: And what about L459?

jax/_src/dtypes.py
jakevdp: Thanks for the comment. I ended up going the other direction, and ensuring that bools are treated as strong everywhere. It's a simpler change that keeps things consistent.

jakevdp:
@skye please take another look. I ended up making things consistent in the other direction, after observing that the strong bool/weak bool promotion discrepancy has been recently fixed (see https://github.com/google/jax/issues/11377#issuecomment-1177858169). I think this ends up being a cleaner change.

jax/interpreters/xla.py
```diff
@@ -275,7 +275,9 @@ def abstractify(x) -> core.AbstractValue:
   raise TypeError(f"Argument '{x}' of type '{type(x)}' is not a valid JAX type")
 
 def _make_abstract_python_scalar(typ, val):
-  return ShapedArray((), dtypes._scalar_type_to_dtype(typ, val), weak_type=True)
+  # Note: all python scalar types are weak except bool
```
skye, line 278: Maybe add: because bool doesn't come in different sizes, so we don't need to worry about unwanted implicit type promtion [is this explanation correct??]

skye (approved):
Much cleaner!

jax/interpreters/xla.py
jakevdp: Done!

PR #11390, opened 2022-07-07
Refactor distributed GPU device initialization.

hawkinsp:
Avoid reregistering backend factories; instead simply have the usual
factory function support distributed GPU.

### DIFF
--- a/jax/_src/distributed.py
+++ b/jax/_src/distributed.py
@@ -26,6 +26,7 @@ from jax._src.lib import xla_client
 from jax._src.lib import xla_extension
 
 class State:
+  process_id: int = 0
   service: Optional[Any] = None
   client: Optional[Any] = None
   preemption_sync_manager: Optional[Any] = None
@@ -34,8 +35,8 @@ class State:
                  coordinator_address: Optional[str] = None,
                  num_processes: Optional[int] = None,
                  process_id: Optional[int] = None):
-    coordinator_address = os.environ.get('JAX_COORDINATOR_ADDRESS',
-                                         None) or coordinator_address
+    coordinator_address = (coordinator_address or
+                           os.environ.get('JAX_COORDINATOR_ADDRESS', None))
 
     if cloud_tpu_init.running_in_cloud_tpu_vm:
       worker_endpoints = cloud_tpu_init.get_metadata(
@@ -58,6 +59,8 @@ class State:
       raise ValueError('Number of processes must be defined.')
     if process_id is None:
       raise ValueError('The process id of the current process must be defined.')
+
+    self.process_id = process_id
 
     if process_id == 0:
       if self.service is not None:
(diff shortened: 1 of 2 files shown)


jax/_src/distributed.py
```diff
@@ -26,6 +26,7 @@
 from jax._src.lib import xla_extension
 
 class State:
+  process_id: int = 0
```
yashk2810, line 29: Shouldn't this be None as default to match the current behavior?

hawkinsp: It doesn't greatly matter. If you pass `None` to the underlying client init function, it defaults to 0. This is simple: always an int (after all, you have to have a node id, even if not distributed.)

jax/_src/distributed.py
```diff
@@ -34,8 +35,9 @@ def initialize(self,
                  coordinator_address: Optional[str] = None,
                  num_processes: Optional[int] = None,
                  process_id: Optional[int] = None):
-    coordinator_address = os.environ.get('JAX_COORDINATOR_ADDRESS',
-                                         None) or coordinator_address
+    self.process_id = process_id
+    coordinator_address = (coordinator_address or
```
hawkinsp, line 39: Note this change. This is intentional: if someone specifies a specific value for an arg, an environment variable should not override it.

jax/_src/lib/xla_bridge.py
```diff
@@ -210,17 +211,23 @@ def register_backend_factory(name, factory, *, priority=0):
 register_backend_factory('tpu_driver', _make_tpu_driver_client,
                          priority=100)
 
+
+def _make_gpu_client(platform_name=None):
+  return xla_client.make_gpu_client(
```
yashk2810, line 216: You should default to 0 here:

if distributed.global_state.process_id is None:
  node_id = 0
else:
  node_id = distributed.global_state.process_id

jax/_src/distributed.py
yashk2810: The current behavior is to error if process_id is None.

For TPUs we figure that thing out.

jax/_src/lib/xla_bridge.py
hawkinsp: Isn't it just fewer lines of code to default to `0` in `distributed.py`?

yashk2810: Right.. but there are 2 behaviors right now and I am confused about them.

initialize has the default None which raises an error if you don't specify process_id.

But the class variable defaults to 0. Previously, process_id was a required field for GPUs.

hawkinsp: There are two cases: single-process and distributed:
* In the single-process case, you don't call the distributed init function. The default is 0, which is the correct default.
* In the distributed case, call the init function. You must specify a `process_id` (the init function enforces this), unless you are on cloud TPU in which case it is inferred.

yashk2810 (approved):

PR #11399, opened 2022-07-07
[dynamic-shapes] add basic abstracted_axes support to jit(f, ...).lower(...)

mattjj:


jax/_src/stages.py
```diff
@@ -379,6 +380,8 @@ def runtime_executable(self) -> Optional[Any]:
     return self._executable.runtime_executable()
 
   def __call__(self, *args, **kwargs):
+    if config.jax_dynamic_shapes:
+      raise NotImplementedError
```
froystig, line 384:
```suggestion
      raise NotImplementedError(
          "invoking a pre-compiled function with dynamic shapes configured")
```

froystig (approved):

PR #11425, opened 2022-07-08
Fix definition of treedef_is_leaf.

pschuh:
Empty nodes like [] and {} have 1 node and 0 leaves. This does not make
them a leaf treedef.

Reproducer:
```
pjit.pjit(lambda x: x, None, (None, {}))((3, {'a': []}))
```

### DIFF
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -90,6 +90,9 @@ def treedef_children(treedef):
 def treedef_is_leaf(treedef):
   return treedef.num_nodes == 1
 
+def treedef_is_strict_leaf(treedef):
+  return treedef.num_nodes == 1 and treedef.num_leaves == 1
+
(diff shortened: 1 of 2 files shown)


mattjj:
This seems right to me, but @hawkinsp can you double check (b/c IIRC you wrote this code when moving pytrees out of Python)?

I think even simpler reproducers are:

```
import jax
print(jax.tree_util.treedef_is_leaf(jax.tree_util.tree_structure({})))
print(jax.tree_util.treedef_is_leaf(jax.tree_util.tree_structure([])))
```

mattjj (approved):

pschuh:
Ah, I just updated it not to modify the treedef_is_leaf because too many people rely on the broken behavior.

mattjj (approved):

PR #11429, opened 2022-07-09
Add loop invariant residual fixpoint test

sharadmv:
Tracker: #10982


mattjj (commented):
Would it be possible to write a test of just the for_loop api exercising the same thing? That is, rather than using partial_eval_jaxpr_nounits, test the code in the for_loop partial eval implementation is doing the right thing and won't lead to bad linearizations.

sharadmv:
Are you suggesting I test `_for_partial_eval` directly instead of indirectly through `pe_no_units`?

sharadmv:
PR has been updated with a test using linearize that checks that values are correct.

mattjj (approved):

PR #11441, opened 2022-07-11
Point to the right download for jaxlib

ludgerpaehler:
The download right now instructs to express the download as 

```bash
pip install "jax[cuda11_cudnn84]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```

in the code-block for jaxlib. This only downloads jax itself. Instead pointing to jaxlib as in 

```bash
pip install --upgrade "jaxlib[cuda11_cudnn84]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```

Downloads the right jaxlib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,10 +464,10 @@ version for jaxlib explicitly:
 pip install --upgrade pip
 
 # Installs the wheel compatible with Cuda >= 11.4 and cudnn >= 8.2
-pip install "jax[cuda11_cudnn82]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
+pip install "jaxlib[cuda11_cudnn82]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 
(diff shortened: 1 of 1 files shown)


YouJiacheng:
I guess
```python
pip install --upgrade "jax[cuda11_cudnn84]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
Should be the "best".
It seems that if newest `jax` was installed but `jaxlib` is stale(or without cuda), `pip install "jax[cuda11_cudnn84]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html` won't install desired `jaxlib` since `jax` requirement already satisfied.

jakevdp:
I've seen variations of this before – `pip`'s handling of optional dependences (the things in brackets) does not always interact well with `--upgrade`. To ensure you're getting the most recent versions of what you specify, I'd suggest this instead:
```
pip uninstall jax jaxlib
pip install jax[...
```

hawkinsp:
Closing; I don't think we're going to merge this.

PR #11453, opened 2022-07-11
CI: upgrade doc requirements

jakevdp:
The reasons for pinning these requirements are now stale:

- We no longer use `sphinx-rtd-theme`, so we can remove the requirement
- The `docutils` pinning was due to a poor interaction with `sphinx-rtd-theme`
- The `sphinx-autodoc-typehints` issue has been fixed in [v1.15.0](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/CHANGELOG.md#1150), although ~newer versions require setting the `set_type_checking_flag` config to avoid warnings about how we define `PyTreeDef`~ the issue fixed by #11455 causes unrelated issues for newer versions of the extension

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,16 +1,10 @@
 sphinx >=3
-sphinx_rtd_theme
+sphinx-autodoc-typehints==1.11.1  # Unpin after jaxlib 0.3.15 is released (#11455)
 sphinx-book-theme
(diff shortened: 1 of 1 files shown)


jakevdp:
The error is due to the issue fixed by #11455

yashk2810 (approved):

PR #11458, opened 2022-07-11
CI: add nightly upstream development test

jakevdp:
The idea here is to run tests once per day against the nightly numpy and scipy releases, in order to catch future errors earlier. In theory this would allow us to fix issues like https://github.com/google/jax/issues/11221 before the problematic code is released.


jakevdp:
I think this is in good shape - in order to test the scheduled run, we need to submit to the main branch.

This should be safe to submit: the worst that can happen is a failed workflow in our Actions tab, and by design this won't block any PR submissions.

Issue #11448, opened 2022-07-11
Memory leak in compiled loops closing over large constants


wiep:
I was running into an issue that I could condense to an elementary example. It seems to me as if closing over large constants in lax loops creates a memory leak while unrolling the loop avoids the issue. A disclaimer: my knowledge of how JAX is limited. Thus, the issue described may not be a memory leak but a misunderstanding on my part of how JAX or XLA buffers compiled computations.

In the MRE, each of the side-effect-free functions  `leak` and `no_leak` create in their body a large constant and later compile function which closes over the large constant. The version which includes a lax loop closing over the constant seems to leak memory. 

I'm not sure whether this issue is related to #282. However, since the function without a lax loop doesn't seem to leak memory, I think it is different. From my understanding of @jakevdp's [comment](https://github.com/google/jax/issues/7930#issuecomment-995003572) on caching of compiled functions, I'd expect that the cache is cleared and memory should be released once `g`, `jg` and `data` go out of scope or once the references are deleted manually.

However, and here I'm not sure if I use `jaxlib.xla_extension.Client` correctly, it seems that using the lax loop creates a live buffer that doesn't get dropped when both functions, the compiled function and the original function, go out of scope (see output below). I also tried to avoid the jit-compilation of the lax loop (see function `leak_no_compile`). This seems to create similar behavior of the program with the difference that, additionally to the live buffer, also a live executable can be found via the `backend.live_executables()` after the function went out of scope.

## MRE

contents of leak.py
```py
import numpy as np
import jax
import psutil


def print_mem_usage() -> None:    
    print(
        f"Resident memory: {psutil.Process().memory_info().rss // (1024 * 1024)} MB"
    )

def print_live_buffers() -> None:
    lbs = jax.lib.xla_bridge.get_backend().live_buffers()
    print(f"Number live buffers: {len(lbs)}")
    for idx, lb in enumerate(lbs):
        print(f"\tlbs[{idx}]: shape {lb.shape} at 0x{lb.unsafe_buffer_pointer():x}")


def print_num_live_executables() -> None:
    les = jax.lib.xla_bridge.get_backend().live_executables()
    print(f"Number live executables: {len(les)}")



def leak():
    # ~ 1 GB of float32s
    size = 1024 * 1024 * 1024 // 4
    data = jax.device_put(np.empty((size, ), dtype=np.float32) + 1)
    
    def g():
        return jax.lax.fori_loop(0, 1, lambda _, __: data[0], 0.0)

[...]

Issue #10023, opened 2022-03-24
lax.while_loop within jax.disable_jit() should not silently fall back to the primitive version.


YouJiacheng:
https://github.com/google/jax/blob/b43ca61a93aec9a638b5fd6b04518869d45201aa/jax/_src/lax/control_flow.py#L279-L288

It should emit a warning or raise an error.

jakevdp:
Can you say more about why you think the behavior is problematic?

YouJiacheng:
Okay. I don't really think this behavior is a bug, thus I use enhancement label.
But I think a good program should not surprise users, i.e. users should be informed if anything doesn't behave as expected, especially when `jax.disable_jit` is intended for debugging.
For example, a user may assume that within `jax.disable_jit`, any "memory leak" should have nothing to do with jit cache, and doesn't realize the "memory leak" is actually not a leak but a cache.

YouJiacheng:
Sure.
Actually it isn't a memory leak, but a cache. (the number of live buffers will stop increasing at 2048)
```python
import jax
from jax import lax
from jax import numpy as jnp

def f(max):
    def cond(val):
        return val < max

    def body(val):
        return val + 1
    return lax.while_loop(cond, body, 0)

def print_num_live_buffers():
    bs = jax.lib.xla_bridge.get_backend().live_buffers()
    print(len(bs))

for _ in range(3):
    x = jnp.full((), 2)
    f(x)
    print_num_live_buffers()
```
Out:
```
1
2
3
```
`disable_jit` works:
```python
import jax
from jax import lax
from jax import numpy as jnp

def f(max):
    def cond(val):
        return val < max

    def body(val):
        return val + 1
    return lax.while_loop(cond, body, 0)

def print_num_live_buffers():
    bs = jax.lib.xla_bridge.get_backend().live_buffers()
    print(len(bs))

for _ in range(3):
    x = jnp.full((), 2)
    with jax.disable_jit():
        f(x)
    print_num_live_buffers()
```
Out:
```
1
1
1
```
`disable_jit` silently doesn't work:
```python
import jax
from jax import lax
from jax import numpy as jnp

@jax.vmap
def f(max):
    def cond(val):
        return val < max

    def body(val):
        return val + 1
    return lax.while_loop(cond, body, 0)

def print_num_live_buffers():
    bs = jax.lib.xla_bridge.get_backend().live_buffers()
    print(len(bs))

for _ in range(3):
    x = jnp.full(1, 2)
[...]

PR #11461, opened 2022-07-11
Use weakref to avoid cache leaks in fori_loop.

pschuh:
This is in response to: https://github.com/google/jax/issues/11448.

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -19,6 +19,7 @@ import operator
 from typing import Any, Callable, List, Optional, Sequence, Tuple, TypeVar
 
 import jax
+import weakref
 from jax import core
 from jax import linear_util as lu
 from jax.config import config
@@ -51,6 +52,7 @@ from jax._src.util import (
     safe_zip,
     split_list,
     unzip2,
+    weakref_lru_cache,
     )
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


YouJiacheng:
FWIW, similar problems:
https://github.com/google/jax/issues/9714
https://github.com/google/jax/issues/10023#issuecomment-1077812932
Could this PR also close #9714 ?

jakevdp (approved):
Looks fine - is it fair to say that we should use `weakref_lru_cache` in place of `cache()` for anything that takes a function as its primary argument?

pschuh:
@jakevdp yea, I think that is reasonable. It is also good for caches that take jaxprs as the first argument.

jax/_src/lax/control_flow/loops.py
```diff
@@ -1497,18 +1499,20 @@ def _fori_cond_fun(loop_carry):
...
 
-@cache()
+@weakref_lru_cache
 def _fori_scan_body_fun(body_fun):
+  body_fun = weakref.ref(body_fun)
   def scanned_fun(loop_carry, _):
     i, x = loop_carry
-    return (i + 1, body_fun(i, x)), None
+    return (i + 1, body_fun()(i, x)), None
   return scanned_fun
```
YouJiacheng, line 1516: IIUC, these two functions are not expensive at all, why we need a cache here?

YouJiacheng:
https://github.com/google/jax/blob/86ab8a84ee9c0c1fc718f3d322f7f597bebb812b/jax/_src/lax/control_flow/loops.py#L1502-L1516
IIUC, these two functions are not expensive at all, why we need cache here?

PR #11468, opened 2022-07-12
[dynamic-shapes] Ensure that the axis_size_env is passed to sub contexts

gnecula:


gnecula:
I did not add new tests, these were discovered with the jax2tf primitives tests when run with --jax-dynamic-shapes.

mattjj (approved):

PR #11469, opened 2022-07-12
BUG: fix jvp rule for lax.rem

### LINKED ISSUE #11466: `lax.rem` has wrong gradient for `y` when `x<0`
`lax.rem` has wrong gradient when `x<0`. For example,

```py
import jax

def fn(y):
    x = jax.numpy.array(-19., jax.numpy.float64)
    return jax.lax.rem(x, y)

y = jax.numpy.array(3., jax.numpy.float64)
res = fn(y)
print(res) # -1.0

a = jax.jacfwd(fn)(y)
print(a) # 7.0
```

Actually, the gradient should be 6 instead of 7. Because `-19 = (-6)*3 + (-1)`. When the `y` is `3.1`, the results will be `-0.4`. Thus, the gradient should be 6.

By the way, another two similar APIs `numpy.remainder` and `numpy.mod` also have this issue.

```py
def fn(y):
    x = jax.numpy.array(-19., jax.numpy.float64)
    return jax.numpy.remainder(x, y)

y = jax.numpy.array(3., jax.numpy.float64)
res = fn(y)
print(res) # 2.0

a = jax.jacfwd(fn)(y)
print(a) # 8.0
```

In these two APIs, `-19 = (-7)*3 + 2`, so their gradient should be 7 instead of 8.

jakevdp:
Fixes #11466

It turns out our large test tolerances were obscuring the fact that the JVP for the second argument of `lax.rem` was incorrect when the inputs have mixed sign. This fixes the issue and makes the test more robust – I checked that the old incorrect jvp rule fails the new tighter test.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2161,7 +2161,7 @@ rem_p = standard_naryop([_int | _float, _int | _float], 'rem')
 ad.defjvp(
     rem_p,
     lambda g, x, y: _maybe_broadcast(broadcast_shapes(np.shape(x), np.shape(y)), g),
-    lambda g, x, y: mul(neg(g), floor(div(x, y))))
+    lambda g, x, y: mul(neg(g), mul(sign(div(x, y)), floor(abs(div(x, y))))))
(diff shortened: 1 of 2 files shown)


froystig (approved):
good catch!

Issue #11402, opened 2022-07-07
[jax2tf] Support shape polymorphism for strided convolutions


sdenton4:
Hello!
I'm trying to export an audio model with variable batch and/or time dimensions. This brings a couple issues.

**Case 0:** enable_xla=True, batch=None:
Seems to work! Created a saved_model from the concrete_function, and subsequently managed to load and evaluate with a different batch size. The resulting model is, of course, XLA-based, though, so can't export to TFLite.

**Case 1:** enable_xla=False, batch=None:
Fails, due to the output_shape argument in tf.nn.conv2d_transpose:  
```TypeError: Expected int32 passed to parameter 'input_sizes' of op 'Conv2DBackpropInput', got (None, 500, 1, 512) of type 'tuple' instead. Error: Expected int32, but got None of type 'NoneType'.```  

This one should be pretty easy to fix, I think: Using tf.shape(x) on a tensor creates a polymorphic shape that plays nice with conv2d_transpose. (specifically, the tf.shape(x) is a KerasTensor with a dtype of int32 and None value in some entries. Then the op knows it's eventually getting int32 values, and doesn't complain.)

**Case 2:** enable_xla=False, time=None  
The inference of padding_type from the padding values causes problems here. The call to lax.padtype_to_pads for 'SAME' padding computes the padding amount from the array shape, which causes problems when the array shape is unknown. Here's a truncated trace:
```
File "/tmp/lyra_notebook.par/google3/third_party/py/flax/linen/linear.py", line 411, in __call__  *
        y = lax.conv_general_dilated(
    File "/tmp/lyra_notebook.par/google3/third_party/py/jax/_src/lax/convolution.py", line 142, in conv_general_dilated  *
        padding = lax.padtype_to_pads(
    File "/tmp/lyra_notebook.par/google3/third_party/py/jax/_src/lax/lax.py", line 4324, in padtype_to_pads  *
        out_shape = _ceil_divide(in_shape, window_strides)
    File "/tmp/lyra_notebook.par/google3/third_party/py/jax/_src/lax/lax.py", line 4309, in _ceil_divide  *
        return -np.floor_divide(np.negative(x1), x2)
    File "/tmp/lyra_notebook.par/google3/third_party/py/jax/experimental/jax2tf/shape_poly.py", line 344, in __floordiv__  **
        return self.divmod(divisor)[0]
[...]

PR #11476, opened 2022-07-13
[jax2tf] Fixes for handling of convolutions with shape_polymorphism and enable_xla=False

gnecula:
Issue: #11402

Due to a typo we were running no tests for convolutions with shape
polymorphism and enable_xla=False.

Added a few more tests from #11402 (Thanks @sdenton4).

The main issue was that in presence of shape polymorphism we cannot
just use `x.shape` for a TF value `x` because it will contain `None`
in the place of unknown dimensions. We must use instead the JAX
abstract values.

This does not fix all issues reported in #11402, there is still the
computation of padding or padding="SAME". Commented out the
corresponding test.


jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1276,18 +1276,55 @@ def _make_harness(group_name: str, name: str,
                   jax.grad(lambda x: jnp.sum(jnp.concatenate([x, x], axis=0))),
                   [RandArg((3, 4, 5), _f32)],
                   poly_axes=[(0, 1)]),
+
+    # Issue #11402 InconclusiveDimensionOperation: Dimension polynomial '-1*t' is not a multiple of '2'
```
gnecula, line 1280: This will be fixed in an upcoming PR. This fails with enable_xla=False and True, so it is a high priority.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -97,18 +108,22 @@ def pads_to_padtype(in_shape, window_shape, window_strides, padding) -> str:
...
+    padding = no_pad + padding + no_pad
     # Add empty padding for dummy dimension, too.
-    padding = tf.concat([no_pad, padding, no_pad, no_pad], 0)
+    padding = no_pad + padding + no_pad + no_pad
   else:
-    padding = tf.concat([no_pad, padding, no_pad], 0)
-    in_shape = tf.pad(in_shape, padding)
-  return in_shape
+    padding = no_pad + padding + no_pad
+  x = tf.pad(x, padding)
```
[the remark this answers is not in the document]
sdenton4: It definitely should /not/ be indented. This is likely the source of some shape mismatches I'm seeing. Good catch!

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -217,19 +232,27 @@ def _conv_general_dilated(
...
       padding, lhs_dilation, rhs_dilation, is_conv1d)
 
-  lhs, rhs = _transpose_for_tf_conv(lhs, rhs, dimension_numbers)
-
-  in_channels = lhs.shape[-1]
-  *rhs_spatial_shapes, _, rhs_out_channel = rhs.shape
+  lhs, lhs_shape, rhs, rhs_shape = _transpose_for_tf_conv(lhs, lhs_shape,
+                                                          rhs, rhs_shape,
+                                                          dimension_numbers)
+  jax2tf._assert_matching_abstract_shape(lhs, lhs_shape)
```
sdenton4, line 252: Looks like you could move these asserts into the _transpose_for_tf_conv function to save some logic in the main branch.

jax/experimental/jax2tf/impl_no_xla.py
```diff
@@ -217,19 +232,27 @@ def _conv_general_dilated(
...
   """Implementation of lax.conv_general_dilated_p using XlaConv."""
+  # In presence of shape polymorphism, lhs.shape and rhs.shape may contain
+  # None. The actual dimension polynomial shapes are in _in_avals.
   del lhs_shape, rhs_shape, precision  # Unused arguments.
-  out_shape = jax2tf._aval_to_tf_shape(_out_aval)
-  _validate_spatial_dimensions(len(lhs.shape) - 2)
-  is_conv1d = len(lhs.shape) - 2 == 1
+  lhs_shape, rhs_shape = _in_avals[0].shape, _in_avals[1].shape
+  jax2tf._assert_matching_abstract_shape(lhs, lhs_shape)
+  jax2tf._assert_matching_abstract_shape(rhs, rhs_shape)
```
sdenton4, line 240: Consider updating _validate_spatial_dimensions to consume (lhs, lhs_shape, rhs, rhs_shape) and moving these asserts into it; they certainly with fit the function name, and it'll simplify reading the main function a little bit.

jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1117,7 +1117,7 @@ def _make_harness(group_name: str, name: str,
                   check_result=True,
                   skip_jax_run=True,
                   tol=None,
-                  enable_and_diable_xla=False,
+                  enable_and_disable_xla=False,
                   expect_error=(None, None),
                   **params) -> Union[Harness, Sequence[Harness]]:
```
sdenton4, line 1122: Can this **params be removed? 

It seems like there's an over-zealous **params somewhere which allowed the typo to pass through undetected... Might be nice to look around and ensure that the 'leaf' functions don't have **params as an argument.

sdenton4 (approved):
This looks great; massive thanks for prioritizing this work. I assume you'll want Marc to take a look as well, but this LGTM, FWIW.

gnecula:
@rxsang @maxwillzq PTAL

marcvanzee:
Really awesome change @gnecula, thanks a lot!

marcvanzee (approved):

rxsang (approved):

PR #11482, opened 2022-07-13
lax_numpy_test: make compatible with upstream numpy

### LINKED ISSUE #11480: ⚠️ Nightly upstream-dev CI failed ⚠️
[Workflow Run URL](https://github.com/google/jax/actions/runs/2670191059)

jakevdp:
This fixes the upstream-dev build breakage in #11480 (see https://github.com/google/jax/runs/7326196684?check_suite_focus=true)

~The reason for the changes to the `histogram*` APIs is that numpy has changed the order of the keyword arguments to these functions when removing the deprecated `normed` argument (see https://github.com/numpy/numpy/pull/21645).~

~Our numpy API tests check that argument order matches between JAX and numpy. While we could choose to diverge from numpy on this and keep JAX's current argument order (making an exception in the test for these APIs), in my judgment it's better to try to maintain compatibility with numpy's positional argument API as numpy evolves. What do you think?~

Opted instead to leave the APIs alone for now, and just fix the tests (with a TODO to update the APIs later)

### DIFF
--- a/.github/workflows/upstream-nightly.yaml
+++ b/.github/workflows/upstream-nightly.yaml
@@ -48,7 +48,7 @@ jobs:
             scipy
       - name: Install JAX
         run: |
-          pip install .[cpu]
+          pip install .[minimum-jaxlib]
       - name: Run tests
         if: success()
         id: status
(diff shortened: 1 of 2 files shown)


froystig (changes requested):
Tracking numpy seems worthwhile. This could break existing user code. Let's highlight that by adding a "breaking change" entry to the changelog as part of this PR?

jakevdp:
Good points... thinking more about this, it might be premature to reorder the arguments now. Let's wait until numpy 1.24 is released with the new argument order before we make changes on this.

I'll update the PR

froystig (approved):
Good call.

PR #11485, opened 2022-07-13
Populate long_description for jax & jaxlib

jakevdp:
I noticed that the PyPI pages for [jax](https://pypi.org/project/jax/) and [jaxlib](https://pypi.org/project/jaxlib/) do not contain any project descriptions. This update to the `setup.py` files should result in the README contents being propagated to those pages on the next release.

For JAX, I tested this by using the approach mentioned in the [Python docs](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/#validating-restructuredtext-markup):
```
$ python setup.py sdist
$ twine check sdist/*
```
I confirmed that this is doing the correct check by removing `long_description_content_type`, and observing that the `twine check` returns an appropriate error.

I did not do such a check for jaxlib, because I'm not set up to easily build it locally.

### DIFF
--- a/jaxlib/README.md
+++ b/jaxlib/README.md
@@ -1,0 +1,7 @@
+# jaxlib: support library for JAX
+
+jaxlib is the support library for JAX. While JAX itself is a pure Python package,
+jaxlib contains the binary (C/C++) parts of the library, including Python bindings,
+the XLA compiler, the PJRT runtime, and a handful of handwritten kernels.
+For more information, including installation and build instructions, refer to main
+JAX README: https://github.com/google/jax/.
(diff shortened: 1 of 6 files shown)


jaxlib/README.md
```diff
@@ -0,0 +1,5 @@
+# jaxlib: support library for JAX
+
+jaxlib is the support library for JAX, and packages all the compiled dependencies
```
yashk2810, line 3: I wonder if we have a better description of jaxlib somewhere

jakevdp: I didn't find anything, but this seems like it would be the canonical place for it 😁

yashk2810 (approved):

jaxlib/README.md
jakevdp: Updated based on an existing resource from @hawkinsp

PR #11486, opened 2022-07-13
Update requirements.txt

prasadjivane:
updated sphinx, sphinx-autodoc-typehints  and jupyter-sphinx .


jakevdp:
Still causing readthedocs failures; I think because sphinx v5 is incompatible with some of the sphinx extensions we use.

FWIW, when it comes to sphinx documentation builds, I usually subscribe to the "if it's not broken, don't fix it" rule.

hawkinsp:
Closing; I suspect this is not going to be merged.

PR #11487, opened 2022-07-13
upstream-dev: add failure information to CI issues

jakevdp:
The purpose of this change is to add summary information to the automated issue created when the nightly upstream-dev CI fails.

Opened from a non-fork due to github actions permissions & tested by changing `'schedule'` to `'pull_request'` in the report job, and confirming that the parsed logs end up in the relevant issue description, as seen in the latest edit to https://github.com/google/jax/issues/11480.

### DIFF
--- a/.github/workflows/parse_logs.py
+++ b/.github/workflows/parse_logs.py
@@ -1,0 +1,59 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Script used in the upstream-dev workflow in order to parse pytest logs."""
+
+
+import argparse
+import json
+
+from pytest import TestReport
+
+
+MSG_FORMAT = """\
+<details><summary>Summary of Failures</summary>
+
+```
+{summary}
+```
+
+</details>
+"""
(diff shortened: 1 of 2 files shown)


jakevdp:
Note: the failure in CI-Nightly is expected, and will be fixed by #11482

yashk2810:
What is the upstream nightly workflow? Why do we need this?

jakevdp:
This is what I added yesterday after some discussion on jax-dev – it tests JAX against development versions of numpy and scipy so we can catch failures before releases of these packages surface those failures to users. See https://github.com/google/jax/pull/11458

yashk2810:
Why not use kokoro for this?

jakevdp:
Because I already implemented it via github actions?

yashk2810:
Ok, but maybe we can switch to kokoro? Then you won't need this parse logs thing and everyone will be notified if something fails. Its a good way to track all the CI we have rather than having things in different places.

jakevdp:
We also have a fair bit of existing CI infrastructure running on github actions, and this workflow notifies everyone via the github issue tracker & contains links to raw logs there. Given that, I'm not sure what the advantage of kokoro would be.

yashk2810:
I am not sure what the advantage of github actions is. Are externally people going to look at this? (I don't think so?)

> existing CI infrastructure running on github actions
I don't think we do? But maybe I am wrong.

jakevdp:
You can see a half dozen github actions CI tests running against this PR. Since we develop open source first, I tend to think of github CI as our primary CI, and Kokoro as secondary to that, so it made sense to me to implement this via actions.

In any case, we already have a working github actions implementation of the intended CI, perhaps it's a bit late to advocate for using a different platform?

yashk2810:
> perhaps it's a bit late to advocate for using a different platform?

I didn't see the thread where this was discussed. So maybe I am not late.

> You can see a half dozen existing github actions CI tests running against this PR.

I wouldn't consider that CI. CI is continuous testing that is running post-submit. The PR stuff is presubmit.

yashk2810 (approved):

PR #11513, opened 2022-07-15
Add dlpack support to device_array and jax.numpy

### LINKED ISSUE #11496: Implement the `__dlpack__` interface
This is described at https://github.com/numpy/numpy/issues/19013, and seems to be developing as a standard

Test case would look something like this:
```python
import jax.numpy as jnp
import numpy as np

x = jnp.arange(10)
x_view = np.from_dlpack(x)
```

(related to #11494)

jakevdp:
This adds the `__dlpack__` interface to `DeviceArray`, as well as a new `jax.numpy.from_dlpack` function with the same API as [`numpy.from_dlpack`](https://numpy.org/devdocs/reference/generated/numpy.from_dlpack.html), which was added to numpy in version 1.22.0.

With this interface defined, the following is possible:
```python
In [1]: import numpy as np

In [3]: jnp.from_dlpack(np.arange(5))
Out[3]: DeviceArray([0, 1, 2, 3, 4], dtype=int64)

In [4]: np.from_dlpack(jnp.arange(5))
Out[4]: array([0, 1, 2, 3, 4], dtype=int32)
```
The cool thing is this dlpack interface is already supported by other libraries like pytorch, so this allows interop with that package as well:
```python
In [5]: import torch

In [6]: jnp.from_dlpack(torch.arange(5))
Out[6]: DeviceArray([0, 1, 2, 3, 4], dtype=int64)
```

Fixes #11496

### DIFF
--- a/docs/jax.numpy.rst
+++ b/docs/jax.numpy.rst
@@ -180,6 +180,7 @@ namespace; they are listed below.
     fromfunction
     fromiter
     fromstring
+    from_dlpack
     full
     full_like
     gcd
--- a/jax/_src/device_array.py
+++ b/jax/_src/device_array.py
@@ -266,6 +266,12 @@ for device_array in [DeviceArray]:
 
   setattr(device_array, "__array__", __array__)
 
+  def __dlpack__(self):
+    from jax.dlpack import to_dlpack  # pylint: disable=g-import-not-at-top
+    return to_dlpack(self)
+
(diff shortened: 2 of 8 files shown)


mattjj (approved):
Awesome, thanks for keeping us on top of this stuff.

(I am disappointed the PR message didn't start with "Why?" this time!)

Issue #11508, opened 2022-07-15
pretty_printer.py get_ipython() not found in latest colab version


bsense-rius:
Please:

- [ X] Check for duplicate issues.
- [X ] Provide a complete example of how to reproduce the bug, wrapped in triple backticks like this:

Running a colab notebook ([here](https://github.com/Kazuhito00/YOLOX-Colaboratory-Training-Sample/blob/main/YOLOX_Colaboratory_Training_Sample.ipynb)) based in Megvi's Yolox stopped working since 2 days ago because of an issue with the _can_use_color() from pretty_printer.py. The error is created by the exception of get_ipython().that if we print the traceback inside _can_use_color()  reports get_ipython() not found.  

Please note that for the last several weeks the _can_use_color() did not raise the exception of NameError for get_ipython, working like a charm. Might colab have changed something?

```
def _can_use_color() -> bool:
  try:
    # Check if we're in IPython or Colab
    ipython = get_ipython()  # type: ignore[name-defined]
    shell = ipython.__class__.__name__
    if shell == "ZMQInteractiveShell":
      # Jupyter Notebook
      return True
    elif "colab" in str(ipython.__class__):
      # Google Colab (external or internal)
      return True
  except NameError:
    pass
  # Otherwise check if we're in a terminal
  return sys.stdout.isatty()
```

- [ ] If applicable, include full error messages/tracebacks.

```
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/lite/python/util.py", line 51, in <module>
    from jax import xla_computation as _xla_computation

  File "/usr/local/lib/python3.7/dist-packages/jax/__init__.py", line 59, in <module>
    from .core import eval_context as ensure_compile_time_eval

  File "/usr/local/lib/python3.7/dist-packages/jax/core.py", line 47, in <module>
    import jax._src.pretty_printer as pp

  File "/usr/local/lib/python3.7/dist-packages/jax/_src/pretty_printer.py", line 61, in <module>
    CAN_USE_COLOR = _can_use_color()
                    └ <function _can_use_color at 0x7f31629d3320>

  File "/usr/local/lib/python3.7/dist-packages/jax/_src/pretty_printer.py", line 59, in _can_use_color
    return sys.stdout.isatty()
           │   └ <yolox.utils.logger.StreamToLoguru object at 0x7f353ea0e390>
           └ <module 'sys' (built-in)>

AttributeError: 'StreamToLoguru' object has no attribute 'isatty'
```

[...]

PR #11514, opened 2022-07-15
pretty_printing: handle case where stdout is patched by a logger

jakevdp:
Fixes issue introduced by https://github.com/google/jax/pull/10688 in which users who patch `sys.stdout` get an error in the recent JAX release. Example reports:

- https://github.com/google/jax/issues/11508
- https://github.com/googlecolab/colabtools/issues/2926

Technically, I think we should expect the user to provide the full interface when patching stdout, but this is a small enough change that if it saves folks some trouble, I think it's worth doing.


sharadmv (approved):
Thanks!

PR #11532, opened 2022-07-18
[CI] upstream-dev: run all tests

jakevdp:
I originally limited this to a subset of tests for initial testing; now upgrading to run all tests. The job has been running on a nightly schedule as intended (see actions dashboard: https://github.com/google/jax/actions/workflows/upstream-nightly.yaml)


jakevdp:
The failure here is strange – it's looking like the flag values are not being initialized on startup. I can reproduce locally & am taking a look.

zhangqiaorjc (approved):

jakevdp:
Failure now is expected (this is the kind of upstream error the CI was meant to catch). I'm going to merge to test again that it opens an issue when it fails tonight.

PR #11541, opened 2022-07-19
[dynamic-shapes] Add support for concatenate

gnecula:
In order to add support for concatenating dynamic shapes I add
a new parameter shape to the concatenate primitive to store the output shape.

We pass the shape as a parameter so that it can be computed once,
while lowering lax.concatenate to primitives. At that time we are
in a context where we can stage computations with shapes. This allows
us to keep the staging and typechecking rules free of shape computations.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -175,6 +175,9 @@ def _extract_tracers_dyn_shape(
   else:
     return [], list(shape)  # type: ignore
 
+def _nr_dynamic_dimensions(static_shape: Sequence[Optional[int]]) -> int:
+  return len([d for d in static_shape if d is None])
+
 def _merge_dyn_shape(
     static_shape: Sequence[Optional[int]],
     dyn_shape: Sequence[Any],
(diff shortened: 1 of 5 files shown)


jax/_src/lax/lax.py
```diff
@@ -2991,8 +2994,34 @@ def _concatenate_batch_rule(batched_args, batch_dims, *, dimension):
...
+    dyn_shape = args[- nr_dynamic_dimensions:]
+    out_shape = _merge_dyn_shape(shape, dyn_shape)
+    out_shape = [x.val if type(x) is core.Literal else x for x in out_shape]
+    out_aval = core.DShapedArray(tuple(out_shape), args[0].aval.dtype, args[0].aval.weak_type)
+    return [out_aval], core.no_effects
+core.custom_typechecks[concatenate_p] = _concatenate_typecheck_rule
+
+def _concatenate_lower(ctx, *xs, dimension, shape):
+  array_xs = xs[- _nr_dynamic_dimensions(shape):]
+  return mhlo.ConcatenateOp(array_xs, mlir.i64_attr(dimension)).results
```
gnecula, line 3023: Not yet sure this is the right lowering, I am not able to run IREE

jax/_src/lax/lax.py
```diff
@@ -2991,8 +2994,34 @@ def _concatenate_batch_rule(batched_args, batch_dims, *, dimension):
...
+  nr_dynamic_dimensions = _nr_dynamic_dimensions(shape)
+  if not nr_dynamic_dimensions:
+    out_aval, effects = concatenate_p.abstract_eval(
+        dimension=dimension, shape=shape)
+    return [out_aval], effects
+  else:
+    dyn_shape = args[- nr_dynamic_dimensions:]
+    out_shape = _merge_dyn_shape(shape, dyn_shape)
+    out_shape = [x.val if type(x) is core.Literal else x for x in out_shape]
+    out_aval = core.DShapedArray(tuple(out_shape), args[0].aval.dtype, args[0].aval.weak_type)
```
gnecula, line 3017: I do not understand the different form of DShapedArray in the staging rule and the type checking rule. I understand that in the first case we have Tracers and in the latter we don't, but isn't it confusing that DShapedArray has two forms?

mattjj: Does [this comment](https://github.com/google/jax/blob/2c72858928440be9d272fe5d04656cf1dcac20b5/jax/core.py#L1409) help explain?

The staging rule needs to put Tracers in shape tuples, while the typechecking rule needs to put Vars in shape tuples.

jax/_src/lax/lax.py
```diff
@@ -609,10 +612,31 @@ def concatenate(operands: Sequence[Array], dimension: int) -> Array:
   Returns:
     An array containing the concatenation.
   """
-  if len(operands) == 0:
-    raise ValueError("concatenate requires a non-empty sequences of arrays")
-  return concatenate_p.bind(*operands, dimension=dimension)
+  if not operands:
```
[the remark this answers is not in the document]
mattjj: Actually, we try not to do checking work in the traceable layer, as that would mean checking overheads which aren't necessary in e.g. eager mode. We should call bind as soon as possible, and leave type-checking work to type-checking rules.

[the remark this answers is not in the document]
mattjj: It seems likely that I don't understand the motivation for moving these computations to here, though!

PR #11543, opened 2022-07-19
Fix MultiProcessGpuTest test

nvcastet:
Since MultiProcessGpuTest was using 'shell=True', only the first element
of the args was executed (i.e. python). Therefore the spawn processes
never executed jax code.
Fix the test and make sure 'jax.distributed' initialize by checking
jax.device_count().

### DIFF
--- a/tests/distributed_test.py
+++ b/tests/distributed_test.py
@@ -93,16 +93,20 @@ class MultiProcessGpuTest(jtu.JaxTestCase):
       args = [
           sys.executable,
           "-c",
-          ('"import jax, os; '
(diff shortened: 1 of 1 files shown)


yashk2810:
We are hitting errors in our GPU tests right now. Once we fix that, I'll approve this so that it can be tested before we submit.

nvcastet:
@yashk2810 What is the status to merge this PR?

yashk2810 (approved):

yashk2810:
LGTM. I have started the tests. I'll pull it in after they pass.

PR #11545, opened 2022-07-19
Add flag to pass generic MLIR to downstream tools

phoenix-meadowlark:
Adds a flag and environment variable `--jax_use_generic_mlir`/`JAX_USE_GENERIC_MLIR` that configures JAX to pass generic MLIR assembly to down-steam MLIR compiler tools like IREE.

This avoids the majority of errors associated with using compiler binaries built with different versions of LLVM than the one `jaxlib` was built with, but is not guaranteed to work if new ops or attributes have been added to MHLO.

### DIFF
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -22,6 +22,7 @@ import functools
 from functools import partial
 import io
 import itertools
+import os
 import re
 import typing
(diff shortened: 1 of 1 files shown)


hawkinsp:
I think this is stale, now that `stablehlo` will take care of dialect stability.

PR #11556, opened 2022-07-20
Correct spelling on word

RuffaloLavoisier:


jax/_src/lax/lax.py
```diff
@@ -4095,7 +4095,7 @@ def _create_token_lowering(ctx, *operands):
 def after_all(*operands):
   """Merges one or more XLA token values. Experimental.
 
-  Wraps the XLA AfterAll operator."""
+  Wraps the XLA After All operator."""
```
apaszke, line 4098: AfterAll is a correct name of the XLA operator, so adding a space here is not the right thing to do

RuffaloLavoisier: Thanks ! fix it.

froystig (approved):

PR #11576, opened 2022-07-21
jnp.searchsorted: add optional method argument to control implementation

jakevdp:
This adds a keyword to access the alternative `searchsorted` implementation explored in https://github.com/google/jax/pull/9108.


shoyer (approved):
Looks great!

PR #11578, opened 2022-07-21
DOC: ensure that _wraps() generates correct links to wrapped functions

### LINKED ISSUE #11492: jax.numpy.reciprocal/remainder docstring
In https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.reciprocal.html#jax.numpy.reciprocal, the link to the numpy implementation actually self-references the jax version.

Instead, it should point to https://numpy.org/doc/stable/reference/generated/numpy.reciprocal.html

The same applies for remainder: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.remainder.html#jax.numpy.remainder

jakevdp:
The issue here is that numpy ufuncs do not have a `__module__` attribute, so we must pass the correct module explicitly.

Fixes #11492

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -302,12 +302,12 @@ def load(*args, **kwargs):
 
 ### implementations of numpy functions in terms of lax
 
-@_wraps(np.fmin)
+@_wraps(np.fmin, module='numpy')
 @jit
 def fmin(x1, x2):
   return where((x1 < x2) | isnan(x2), x1, x2)
 
-@_wraps(np.fmax)
+@_wraps(np.fmax, module='numpy')
 @jit
 def fmax(x1, x2):
   return where((x1 > x2) | isnan(x2), x1, x2)
@@ -346,7 +346,7 @@ def trapz(y, x=None, dx=1.0, axis: int = -1):
   return 0.5 * (dx * (y[..., 1:] + y[..., :-1])).sum(-1)
 
 
-@_wraps(np.trunc)
+@_wraps(np.trunc, module='numpy')
 @jit
 def trunc(x):
   _check_arraylike('trunc', x)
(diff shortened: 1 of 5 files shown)


yashk2810 (approved):

ilemhadri:
Now there is no numpy hyperlink anymore?

e.g. https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.reciprocal.html#jax.numpy.reciprocal, as of July 22nd, 10:31AM UTC.

jakevdp:
Yeah, I noticed that. I'm not sure there's anything we can do: the intersphinx reference is generated correctly, but the links don't work. Poking around the numpy documentation, I see the same behavior there for sphinx references to ufuncs (e.g. https://numpy.org/neps/nep-0013-ufunc-overrides.html#list-of-operators-and-numpy-ufuncs)

I think it's either a sphinx bug or a numpy bug of some sort, perhaps related to the fact that ufuncs don't define `__module__`. I don't have plans to look into this further, but if numpy/sphinx ever fixes the issue then the fix should propagate to the JAX documentation.

ilemhadri:
Very curious - on some links [like this one](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.reciprocal.html) there is no hyperlink at all, whereas on others [like this one](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.matrix_rank.html), there is!

jakevdp:
I believe these links are missing because there's not `.. module :: numpy` declaration at the top of https://github.com/numpy/numpy/blob/main/doc/source/reference/routines.math.rst

ilemhadri:
Created a pull request to see if it would be possible to change the numpy `routines.math.rst` accordingly: https://github.com/numpy/numpy/pull/22162

ilemhadri:
the discussion in https://github.com/numpy/numpy/pull/22162 is insightful!

rossbar (and myself) are not 100% confident that the other fix will work, and TBH I don't have an easy way to test it locally. I would welcome instructions though!

Alternatively, @jakevdp would it be fine to simply modify the role linking the numpy ufuncs in the jax docs to `:py:data:` as an easy fix and move on?

ilemhadri:
@jakevdp

PR #11612, opened 2022-07-25
Fix debugger line numbers

sharadmv:


jax/_src/debugger/core.py
```diff
@@ -63,8 +63,18 @@ def tree_unflatten(cls, info, valid_locals):
   def from_frameinfo(cls, frame_info) -> DebuggerFrame:
     try:
       _, start = inspect.getsourcelines(frame_info.frame)
-      source = inspect.getsource(frame_info.frame).split('\n')
-      offset = frame_info.lineno - start
+      source = inspect.getsource(frame_info.frame).split("\n")
+      # Line numbers begin at 1 but offsets begin at 0. `inspect.getsource` will
+      # return a partial view of the file and a `start` indicating the line
+      # number that the source code starts at. `start` is a line number meaning
```
mattjj, line 69: I don't understand the sentence "`start` is a line number meaning it begins at 1, but it returns 0, then we are at the beginning of the file.". Rephrase? Maybe "_if_ it returns zero"? But then it's still a bit confusing to say "it begins at 1" but it can return 0...

mattjj (approved):

jax/_src/debugger/core.py
```diff
@@ -63,8 +63,18 @@ def tree_unflatten(cls, info, valid_locals):
...
+      # it begins at 1, but it returns 0, then we are at the beginning of the
+      # file. Therefore, `offset` is just the `lineno - 1`. If `start` is
+      # nonzero, then we subtract it off from the `lineno` and don't need to
+      # subtract 1 since both start and lineno are 1-indexed.
+      if start:
+        offset = frame_info.lineno - start
+      else:
+        offset = frame_info.lineno - 1
```
mattjj, line 77: Is the logic here `frame_info.lineno - max(1, start)`?

jax/_src/debugger/core.py
sharadmv: I reworded to make it more clear.

jax/_src/debugger/core.py
sharadmv: Yes, thanks for the suggestion.

PR #11618, opened 2022-07-26
fix _scan_partial_eval_custom loop-hoisting bug

mattjj:
After applying the constant-hoisting optimization in the new scan partial eval rule, we need to account for the potentially new constant part of the signature of `jaxpr_known_loop` compared to `jaxpr_known`; that is, `jaxpr_known_loop` might take different constant inputs than `jaxpr_known`. Before this fix we were (re)using the stale `const_uk` and `unks_in` values to set up the `num_consts` and `linear` for the known part of the scan.

We could repro the error by making a loop where the number of constants of `jaxpr_known` is different from the number of constants for `jaxpr_known_loop`, e.g. by having two loop-invariant values computed from one constant input, as in the test added here. I verified this test fails on main.

Also passes a real-world test from @levskaya.


levskaya (approved):
thanks!

PR #11627, opened 2022-07-26
err on `None` predicate to `lax.cond`

### LINKED ISSUE #11574: jax.lax.cond evaluates None as True
When using `jax.lax.cond`, passing `None` as the first argument causes the `true_fun` argument to get called. This is contrary to the behavior one would expect for Python if statements.

For example, running the following
```python
import jax

true_fun = lambda x: True
false_fun = lambda x: False

for pred in [True, False, None]:
	print(pred, jax.lax.cond(pred, true_fun, false_fun, None))
```

gives the following output:
```
True True
False False
None True
```

I am using the newest PyPI versions of `jax` and `jaxlib` (0.3.14).

froystig:
fixes #11574

### DIFF
--- a/jax/_src/lax/control_flow/conditionals.py
+++ b/jax/_src/lax/control_flow/conditionals.py
@@ -148,7 +148,8 @@ def _cond(pred, true_fun: Callable, false_fun: Callable, *operands,
           operand=_no_operand_sentinel, linear=None):
   """Conditionally apply ``true_fun`` or ``false_fun``.
 
-  ``cond()`` has equivalent semantics to this Python implementation::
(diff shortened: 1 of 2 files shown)


jakevdp:
I think we'll need a deeper fix here... e.g. this doesn't cover the case of `jit(cond)(None, ...)`

froystig:
> this doesn't cover the case of `jit(cond)(None, ...)`

The unit test added in this PR begs to differ, no?

jakevdp:
Ah, so it does...

jakevdp (approved):

PR #11635, opened 2022-07-27
Cache subjaxprs when lowering to MLIR functions.

pschuh:
This will only actually dedup the subjaxpr when
jax_experimental_subjaxpr_lowering_cache is enabled.

### DIFF
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -398,6 +398,7 @@ class ModuleContext:
 
   # Cached primitive lowerings.
   cached_primitive_lowerings: Dict[Any, func_dialect.FuncOp]
+  cached_call_jaxpr_lowerings: Dict[Any, func_dialect.FuncOp]
 
   @property
   def axis_env(self) -> xla.AxisEnv:
@@ -416,7 +417,9 @@ class ModuleContext:
       ip: Optional[ir.InsertionPoint] = None,
       symbol_table: Optional[ir.SymbolTable] = None,
       cached_primitive_lowerings: Optional[Dict[Any,
-                                                func_dialect.FuncOp]] = None):
+                                                func_dialect.FuncOp]] = None,
+      cached_call_jaxpr_lowerings: Optional[Dict[Any,
+                                                 func_dialect.FuncOp]] = None):
     assert platform is not None
     self.context = context or make_ir_context()
     self.module = module or ir.Module.create(loc=ir.Location.unknown(self.context))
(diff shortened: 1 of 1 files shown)


jax/interpreters/mlir.py
```diff
@@ -1075,7 +1081,17 @@ def _call_lowering(fn_name, stack_name, call_jaxpr, backend, ctx, avals_in,
...
+    key = (fn_name, call_jaxpr.jaxpr, tuple(effects))
+    try:
+      func_op = ctx.cached_call_jaxpr_lowerings[key]
+    except KeyError:
+      func_op = lower_jaxpr_to_fun(ctx, fn_name, call_jaxpr, effects)
+      ctx.cached_call_jaxpr_lowerings[key] = func_op
+  else:
+    func_op = lower_jaxpr_to_fun(ctx, fn_name, call_jaxpr, effects)
+  symbol_name = func_op.name.value
   args = [*tokens_in.tokens(), *args]
```
froystig, line 1095: Maybe bundle all this maybe-cache-lookup logic up into a separate function, e.g. `lower_jaxpr_to_fun_cached`? Just to keep things tidy.

jax/interpreters/mlir.py
```diff
@@ -430,6 +433,9 @@ def __init__(
     self.channel_iterator = channel_iterator
     self.keepalives = keepalives
     self.host_callbacks = host_callbacks
+    self.cached_call_jaxpr_lowerings = ({}
+                                        if cached_call_jaxpr_lowerings is None
+                                        else cached_call_jaxpr_lowerings)
```
mattjj, line 438:
```suggestion
    self.cached_call_jaxpr_lowerings = (cached_call_jaxpr_lowerings or {})
```

does this golf work?

mattjj (approved):

jax/interpreters/mlir.py
froystig: If it does, I say we also do this to the analogous expression a few lines above. :)

pschuh: The one problem with that is that {} is a False type so it will decouple the object identity. This happens enough that it can actually thwart caching.

jax/interpreters/mlir.py
pschuh: Done.

jax/interpreters/mlir.py
mattjj: Oh man mutation is hard to reason about :P

PR #11641, opened 2022-07-28
update cond partial eval to so eqn effects match branch jaxprs'

mattjj:
Also add some new tests, including some skipped ones, for how effects should interact with jax.linearize (I think...).

### DIFF
--- a/tests/debugging_primitives_test.py
+++ b/tests/debugging_primitives_test.py
@@ -214,9 +214,40 @@ class DebugPrintTransformationTest(jtu.JaxTestCase):
 
     with capture_stdout() as output:
       x = jnp.array(1., jnp.float32)
-      jax.jvp(f, (x,), (x,))  # should print out cos(1.)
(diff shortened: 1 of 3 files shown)


tests/debugging_primitives_test.py
```diff
@@ -214,9 +214,38 @@ def f(x):
...
+    def f(x):
+      x = jnp.sin(x)
+      x = print_tangent(x)
+      return x
+
+    with capture_stdout() as output:
+      x = jnp.array(1., jnp.float32)
+      y, f_lin = jax.linearize(f, x)
+      jax.effects_barrier()
+    self.assertEqual(output(), "")
```
sharadmv, line 243: This passes right?

sharadmv (approved):
Thanks

tests/debugging_primitives_test.py
mattjj: Actually no, nothing is printed. Try it yourself!

PR #11644, opened 2022-07-28
fix lax.linalg.eigh documentation

Sohl-Dickstein:
Order of returned args was reversed relative to [NumPY documentation](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html) and relative to code behavior.

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -162,14 +162,14 @@ def eigh(x, *, lower: bool = True, symmetrize_input: bool = True,
       implementation-defined order.
 
   Returns:
-    A tuple ``(v, w)``.
-
-    ``v`` is an array with the same dtype as ``x`` such that ``v[..., :, i]`` is
-    the normalized eigenvector corresponding to eigenvalue ``w[..., i]``.
+    A tuple ``(w, v)``.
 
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks - I believe the documentation is correct as written. See the outputs here:
```python
>>> from jax import lax
>>> import numpy as np
>>> lax.linalg.eigh(np.eye(3))
(DeviceArray([[1., 0., 0.],
              [0., 1., 0.],
              [0., 0., 1.]], dtype=float32),
 DeviceArray([1., 1., 1.], dtype=float32))
```
The first returned element contains the eigenvectors `v`, while the second contains the eigenvalues `w`.

If you're looking for a function with a signature similar to `numpy.linalg.eigh`, you probably want [`jax.numpy.linalg.eigh`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.eigh.html).

Sohl-Dickstein:
Thanks for the careful consideration! And good catch. This is my bad -- I was running jax.numpy.linalg.eigh, but reading the documentation for lax.linalg.eigh. I saw the first was just a lightweight wrapper around the second, and somehow thought it also inherited its doc-string.

PR #11651, opened 2022-07-28
add cond dce rule and custom-policy partial eval rule

mattjj:
This is the `cond` analogue of the new rule(s) for `scan` landed in #10576. Luckily `cond` is a lot easier!

Context from the PR message of #9148:
> A few months ago we wrote a new version of remat to support custom rematerialization policies. The new version is also built on a simpler basic implementation. But instead of replacing the pre-existing remat implementation, we kept both around because the new remat did not yet support higher-order primitives like scan and cond. Tech debt!

This PR adds two new rules for `cond`: the DCE rule and the `_cond_partial_eval_custom` rule. The latter is to support partial evaluation parameterized by a policy, so that what gets staged out is in general a superset of what's unknown: known things can get staged out too.

The tests here are basically analogues of the scan ones.

### DIFF
--- a/jax/_src/lax/control_flow/conditionals.py
+++ b/jax/_src/lax/control_flow/conditionals.py
@@ -17,8 +17,9 @@ import functools
 from functools import partial
 import inspect
 import itertools
+import operator
 
-from typing import Callable, Sequence
+from typing import Callable, Sequence, List, Tuple
 
 from jax import core
 from jax import linear_util as lu
@@ -36,7 +37,8 @@ from jax._src import source_info_util
 from jax._src import util
 from jax._src.lax import lax
 from jax._src.traceback_util import api_boundary
-from jax._src.util import safe_map, extend_name_stack, split_list
+from jax._src.util import (safe_map, extend_name_stack, split_list,
+                           partition_list)
 from jax._src.lib.mlir import ir
 from jax._src.lib.mlir.dialects import mhlo
 import numpy as np
@@ -52,7 +54,7 @@ from jax._src.lax.control_flow.common import (
     allowed_effects,
     )
 
-_map, unsafe_map = safe_map, map
+map, unsafe_map = safe_map, map
 
 
 # For backward compatibility with a previous switch/cond calling convention,
@@ -124,7 +126,7 @@ def switch(index, branches: Sequence[Callable], *operands,
     return branches[int(index)](*operands)
 
   ops, ops_tree = tree_flatten(operands)
-  ops_avals = tuple(_map(_abstractify, ops))
+  ops_avals = tuple(map(_abstractify, ops))
 
   jaxprs, consts, out_trees = _initial_style_jaxprs_with_common_consts(
       branches, ops_tree, ops_avals, primitive_name='switch')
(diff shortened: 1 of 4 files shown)


jax/_src/lax/control_flow/conditionals.py
```diff
@@ -433,6 +435,78 @@ def _cond_partial_eval(trace, *tracers, branches, linear):
...
+  res_binders = map(newvar, all_res_avals)
+
+  # Build the known eqn.
+  ins_known, _ = partition_list(unks_in, eqn.invars)  # includes index invar
+  out_binders_known, _ = partition_list(unks_out, eqn.outvars)
+  linear_known = [l for l, uk in zip(eqn.params['linear'], ops_uk) if not uk]
+  params_known = dict(branches=branches_known, linear=tuple(linear_known))
+  eqn_known = pe.new_jaxpr_eqn(
+      ins_known, [*out_binders_known, *res_binders], cond_p, params_known,
+      core.no_effects, eqn.source_info)
```
sharadmv, line 497: Should join the branch effects here and below, right?

jax/_src/lax/control_flow/conditionals.py
```diff
@@ -509,15 +583,51 @@ def augment_jaxpr(jaxpr, res_indices):
...
+  # First, compute which inputs are used in any branch (not including `pred`).
+  used_inputs: List[bool] = [False] * (len(eqn.invars) - 1)  # -1 for pred
+  for jaxpr in branches:
+    _, used_inputs_ = pe.dce_jaxpr(jaxpr, used_outputs, instantiate=False)
+    used_inputs = map(operator.or_, used_inputs, used_inputs_)
+
+  # Next, compute DCEd branches, instantiating according to used_inputs.
+  dce_branches_ = [pe.dce_jaxpr(jaxpr, used_outputs, instantiate=used_inputs)[0]
+                   for jaxpr in branches]
+  dce_branches = [core.ClosedJaxpr(jaxpr, closed_jaxpr.consts)
```
sharadmv, line 608: Are these consts always empty? If not, can they be DCEd?

jax/_src/lax/control_flow/conditionals.py
```diff
@@ -509,15 +583,51 @@ def augment_jaxpr(jaxpr, res_indices):
...
+                  for closed_jaxpr, jaxpr in zip(closed_branches, dce_branches_)]
+
+  # Finally, update parameters and form the new eqn.
+  dce_linear = [l for l, used in zip(eqn.params['linear'], used_inputs) if used]
+  new_params = dict(eqn.params, branches=tuple(dce_branches),
+                    linear=tuple(dce_linear))
+  # TODO(mattjj,sharadmv): don't assume effects are never DCE'd?
```
sharadmv, line 615: Can you not do a join across the branch effects here as well?

sharadmv (approved):

jax/_src/lax/control_flow/conditionals.py
mattjj: Oops, thanks! Forgot to update it to reflect #11641. Hoisted by my own petard.

jax/_src/lax/control_flow/conditionals.py
mattjj: Good q. It turns out they can't be DCE'd, in the sense that `dce_jaxpr` always leaves them untouched, just as a convention of that function. So this code is safe!

jax/_src/lax/control_flow/conditionals.py
mattjj: This was stale, not updated after https://github.com/google/jax/pull/11641. Will fix...

mattjj: Ah actually, we can do a join here but `dce_jaxpr` will never eliminate effects. Really this TODO should go elsewhere, and I'll update the code here to be ready for `dce_jaxpr` potentially eliminating effects.

PR #11656, opened 2022-07-29
add while_loop custom-policy partial eval rule

mattjj:
This is the `while_loop` analogue of the new partial evaluation rule for `scan` landed in #10576, and for `cond` landed in #11651.

The custom-remat-policy partial evaluation rule for `while_loop` is actually just the same as the previous (policy-less) partial eval rule, written in a jaxpr-to-jaxpr (rather than trace-time) way. That's because we can't save residuals from `while_loop` anyway, at least not without dynamic shapes. Indeed:
* `while_loop` only supports `jax.linearize`, not `jax.vjp`, and
* with `jax.linearize` we always have to do "full rematerialization", running a full copy of the primal computation along with the linear part in the linearized body.

So the partial evaluation rule just needs to:
1. hoist out the known part of the body and cond functions, and
2. otherwise stage out a full copy of the loop.

As a result, the `_while_partial_eval_custom` function this PR adds is very close to the existing `_while_partial_eval` (except that it accepts/returns `JaxprEqn`s).

I chose to use the `partial_eval_jaxpr_custom` machinery (rather than just using the `partial_eval_jaxpr_nounits` stuff) just for consistency across the custom partial eval rules. I expect we'll do a consolidation pass across the standard- and custom-policy partial evaluation rules at some point, though for landing these I think keeping parallel implementations is good.

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -828,14 +828,14 @@ def _scan_partial_eval_custom(saveable, unks_in, inst_in, eqn):
     unks_in = const_uk   + carry_uk   + xs_uk
     jaxpr_known_, jaxpr_staged_, unks_out, inst_out, num_res = \
         pe.partial_eval_jaxpr_custom(
-            jaxpr.jaxpr, in_unknowns=unks_in, in_inst=[True] * len(unks_in),
+            jaxpr.jaxpr, in_unknowns=unks_in, in_inst=True,
             ensure_out_unknowns=carry_uk + [False] * num_ys,
             ensure_out_inst=True, saveable=saveable)
     carry_uk_out, ys_uk = split_list(unks_out, [num_carry])
     if carry_uk_out == carry_uk:
       break
     else:
-      carry_uk = _map(operator.or_, carry_uk  , carry_uk_out  )
+      carry_uk = _map(operator.or_, carry_uk, carry_uk_out)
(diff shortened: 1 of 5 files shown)


jax/_src/lax/control_flow/loops.py
```diff
@@ -1309,6 +1309,70 @@ def _while_partial_eval(trace: pe.JaxprTrace, *tracers: pe.Tracer, cond_nconsts:
...
+  params_known = dict(cond_jaxpr=cond_jaxpr_known, body_jaxpr=body_jaxpr_known,
+                      cond_nconsts=len(cond_consts_uk) - sum(cond_consts_uk),
+                      body_nconsts=len(body_consts_uk) - sum(body_consts_uk))
+  effects_known = core.join_effects(cond_jaxpr_known.effects,
+                                    body_jaxpr_known.effects)
+  eqn_known = pe.new_jaxpr_eqn(ins_known, out_binders_known, while_p,
+                               params_known, effects_known, eqn.source_info)
+
+  # Staged eqn is same as input eqn.
+  eqn_staged = eqn
```
sharadmv, line 1366: Nice so easy

sharadmv (approved):

jax/_src/lax/control_flow/loops.py
mattjj: I should've made it look harder...

PR #11672, opened 2022-07-29
Add configuration options jax_cuda_visible_devices and jax_rocm_visib…

hawkinsp:
…le_devices.

Example usage:
jax.config.update("jax_cuda_visible_devices", "1,3")

### DIFF
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -76,6 +76,14 @@ flags.DEFINE_integer(
     'Optional profile version for XLA compilation. '
     'This is meaningful only when XLA is configured to '
     'support the remote compilation profile feature.')
+flags.DEFINE_string(
+    'jax_cuda_visible_devices', 'all',
+    'Restricts the set of CUDA devices that JAX will use. Either "all", or a '
+    'comma-separate list of integer device IDs.')
+flags.DEFINE_string(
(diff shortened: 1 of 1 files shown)


jax/_src/lib/xla_bridge.py
```diff
@@ -212,20 +220,29 @@ def register_backend_factory(name, factory, *, priority=0):
                          priority=100)
 
 
-def make_gpu_client(platform_name=None):
+def make_gpu_client(*, platform_name, visible_devices_flag):
+  visible_devices = getattr(FLAGS, visible_devices_flag, "all")
+  # Pass allowed_devices unconditionally when jaxlib 0.3.15 is the minimum.
+  kwargs = {}
+  if visible_devices != "all":
+    kwargs["allowed_devices"] = {int(x) for x in visible_devices.split(",")}
```
yashk2810, line 228: Instead of just `.split()`, if someone does `"1, 2, 3"`, then you might have to strip the whitespaces.

So probably add `.strip` too?

sudhakarsingh27: probably okay since `int(x)` would ignore whitespaces?

yashk2810: Oh wow, that works!

nvcastet:
@hawkinsp @yashk2810 @skye Any update on this PR?

Issue #11501, opened 2022-07-14
Add `dtype` argument to `jax.numpy.concatenate` and friends


jakevdp:
`np.concatentate` got a `dtype` argument in numpy 1.20; other functions will get them in numpy 1.24.

- [ ] `np.concatenate`
- [ ] `np.stack`
- [ ] `np.vstack`
- [ ] `np.hstack`
- [ ] `np.row_stack`

There is also a `casting` argument that goes with this; for now we can probably ignore this because JAX's type promotion system does not have well-defined casting modes.

PR #11673, opened 2022-07-29
Add `dtype` argument to `jnp.concatenate`

jenkspt:
Getting started on #11501 by adding `dtype` argument to `jnp.concatenate`
Updates `concatenate` tests

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1650,9 +1650,9 @@ def tile(A, reps):
                         [k for pair in zip(reps, A_shape) for k in pair])
   return reshape(result, tuple(np.multiply(A_shape, reps)))
 
-def _concatenate_array(arr, axis: int):
+def _concatenate_array(arr, axis: int, dtype=None):
   # Fast path for concatenation when the input is an ndarray rather than a list.
-  arr = asarray(arr)
+  arr = asarray(arr, dtype=dtype)
   if arr.ndim == 0 or arr.shape[0] == 0:
     raise ValueError("Need at least one array to concatenate.")
   if axis is None:
@@ -1665,26 +1665,29 @@ def _concatenate_array(arr, axis: int):
   return lax.reshape(arr, shape, dimensions)
 
 @_wraps(np.concatenate)
-def concatenate(arrays, axis: int = 0):
+def concatenate(arrays, axis: int = 0, dtype=None):
   if isinstance(arrays, (np.ndarray, ndarray)):
-    return _concatenate_array(arrays, axis)
+    return _concatenate_array(arrays, axis, dtype=dtype)
   _stackable(*arrays) or _check_arraylike("concatenate", *arrays)
   if not len(arrays):
     raise ValueError("Need at least one array to concatenate.")
   if ndim(arrays[0]) == 0:
     raise ValueError("Zero-dimensional arrays cannot be concatenated.")
   if axis is None:
-    return concatenate([ravel(a) for a in arrays], axis=0)
+    return concatenate([ravel(a) for a in arrays], axis=0, dtype=dtype)
   if hasattr(arrays[0], "concatenate"):
-    return arrays[0].concatenate(arrays[1:], axis)
+    return arrays[0].concatenate(arrays[1:], axis, dtype=dtype)
   axis = _canonicalize_axis(axis, ndim(arrays[0]))
-  arrays = _promote_dtypes(*arrays)
+  if dtype is None:
+    arrays = _promote_dtypes(*arrays)
+  else:
+    arrays = [asarray(arr, dtype=dtype) for arr in arrays]
   # lax.concatenate can be slow to compile for wide concatenations, so form a
   # tree of concatenations as a workaround especially for op-by-op mode.
   # (https://github.com/google/jax/issues/653).
   k = 16
   if len(arrays) == 1:
-    return asarray(arrays[0])
+    return asarray(arrays[0], dtype=dtype)
   else:
     while len(arrays) > 1:
       arrays = [lax.concatenate(arrays[i:i+k], axis)
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -203,7 +203,9 @@ class PRNGKeyArray:
     reshaped_keys = jnp.reshape(self._keys, (*newshape, -1), order=order)
     return PRNGKeyArray(self.impl, reshaped_keys)
 
-  def concatenate(self, key_arrs, axis):
+  def concatenate(self, key_arrs, axis, dtype=None):
+    if dtype is not None:
+      raise ValueError('dtype argument not supported for concatenating PRNGKeyArray')
     axis = canonicalize_axis(axis, self.ndim)
     arrs = [self._keys, *[k._keys for k in key_arrs]]
     return PRNGKeyArray(self.impl, jnp.concatenate(arrs, axis))
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -2299,15 +2299,16 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
     self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker)
 
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_axis={}_baseshape=[{}]_dtypes=[{}]".format(
-          axis, ",".join(str(d) for d in base_shape),
+      {"testcase_name": "_axis={}_dtype={}_baseshape=[{}]_argdtypes=[{}]".format(
+          axis, dtype and np.dtype(dtype).name, ",".join(str(d) for d in base_shape),
           ",".join(np.dtype(dtype).name for dtype in arg_dtypes)),
-       "axis": axis, "base_shape": base_shape, "arg_dtypes": arg_dtypes}
+       "axis": axis, "dtype": dtype, "base_shape": base_shape, "arg_dtypes": arg_dtypes}
       for num_arrs in [3]
       for arg_dtypes in itertools.combinations_with_replacement(default_dtypes, num_arrs)
       for base_shape in [(4,), (3, 4), (2, 3, 4)]
+      for dtype in [None] + default_dtypes
       for axis in range(-len(base_shape)+1, len(base_shape))))
-  def testConcatenate(self, axis, base_shape, arg_dtypes):
+  def testConcatenate(self, axis, dtype, base_shape, arg_dtypes):
     rng = jtu.rand_default(self.rng())
     wrapped_axis = axis % len(base_shape)
     shapes = [base_shape[:wrapped_axis] + (size,) + base_shape[wrapped_axis+1:]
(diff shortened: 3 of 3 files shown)


mattjj (approved):
This is great, thank you!

jax/_src/numpy/lax_numpy.py
```diff
@@ -1665,26 +1665,29 @@ def _concatenate_array(arr, axis: int):
...
   if not len(arrays):
     raise ValueError("Need at least one array to concatenate.")
   if ndim(arrays[0]) == 0:
     raise ValueError("Zero-dimensional arrays cannot be concatenated.")
   if axis is None:
-    return concatenate([ravel(a) for a in arrays], axis=0)
+    return concatenate([ravel(a) for a in arrays], axis=0, dtype=dtype)
   if hasattr(arrays[0], "concatenate"):
-    return arrays[0].concatenate(arrays[1:], axis)
+    return arrays[0].concatenate(arrays[1:], axis, dtype=dtype)
```
mattjj, line 1679: Looks like this line causes an error on internal tests now, since `arrays[0].concatenate` may not accept a `dtype` argument with older versions of NumPy.

We could change the test to `if hasattr(arrays[0], "concatenate") and np.__version__.split('.')[:2] >= (1, 20):` or something (untested). WDYT?

jenkspt: That seems reasonable -- but It's unclear to me when `numpy.narray` or `jax.numpy.DeviceArray` will even have a "concatenate" attribute?

tests/lax_numpy_test.py
```diff
@@ -2299,25 +2299,26 @@ def testCompressMethod(self, shape, dtype, axis):
     self._CheckAgainstNumpy(np_fun, jnp_fun, args_maker)
 
   @parameterized.named_parameters(jtu.cases_from_list(
-      {"testcase_name": "_axis={}_baseshape=[{}]_dtypes=[{}]".format(
-          axis, ",".join(str(d) for d in base_shape),
+      {"testcase_name": "_axis={}_dtype={}_baseshape=[{}]_argdtypes=[{}]".format(
+          axis, np.dtype(dtype).name, ",".join(str(d) for d in base_shape),
```
jakevdp, line 2303: `np.dtype(None).name` will result in `float64`, which is misleading. Maybe use something like `dtype and np.dtype(dtype).name` ?

tests/lax_numpy_test.py
```diff
@@ -2299,25 +2299,26 @@ def testCompressMethod(self, shape, dtype, axis):
...
     wrapped_axis = axis % len(base_shape)
     shapes = [base_shape[:wrapped_axis] + (size,) + base_shape[wrapped_axis+1:]
               for size, _ in zip(itertools.cycle([3, 1, 4]), arg_dtypes)]
     def np_fun(*args):
       args = [x if x.dtype != jnp.bfloat16 else x.astype(np.float32)
               for x in args]
-      dtype = functools.reduce(jnp.promote_types, arg_dtypes)
-      return np.concatenate(args, axis=axis).astype(dtype)
-    jnp_fun = lambda *args: jnp.concatenate(args, axis=axis)
+      _dtype = functools.reduce(jnp.promote_types, arg_dtypes) if dtype is None else dtype
```
jakevdp, line 2319: shorter:
```python
_dtype = dtype or jnp.result_type(*arg_dtypes)
```

tests/lax_numpy_test.py
```diff
@@ -2299,25 +2299,26 @@ def testCompressMethod(self, shape, dtype, axis):
...
+      return np.concatenate(args, axis=axis).astype(_dtype)
```
jakevdp, line 2320: I'm a bit worried here that we're bypassing numpy's `dtype` argument here - can we find a way to use it here, to ensure that our implementation matches numpy's? e.g. pass the dtype argument when it is not `None`?

jenkspt: I can do that -- but would also require checking the numpy version as @mattjj mentioned.

jenkspt: Ok I used `packaging.version.Version` to use the dtype arg for numpy versions >= 1.20

jax/_src/numpy/lax_numpy.py
jenkspt: Ok added this check w/ `packaging.version.Version`

jax/_src/numpy/lax_numpy.py
```diff
@@ -1665,26 +1666,29 @@ def _concatenate_array(arr, axis: int):
...
     raise ValueError("Need at least one array to concatenate.")
   if ndim(arrays[0]) == 0:
     raise ValueError("Zero-dimensional arrays cannot be concatenated.")
   if axis is None:
-    return concatenate([ravel(a) for a in arrays], axis=0)
-  if hasattr(arrays[0], "concatenate"):
-    return arrays[0].concatenate(arrays[1:], axis)
+    return concatenate([ravel(a) for a in arrays], axis=0, dtype=dtype)
+  if hasattr(arrays[0], "concatenate") and (Version(np.__version__) >= Version(1.20)):
+    return arrays[0].concatenate(arrays[1:], axis, dtype=dtype)
```
jakevdp, line 1680: I don't think we should be branching based on numpy version within the implementation (the behavior of JAX should not change depending on what numpy version you happen to have installed!) it's appropriate to branch based on numpy version within the tests, however; you can see this in several existing tests: https://github.com/google/jax/blob/be8939771c9abbe384df1da5a970b306fad2e965/tests/lax_numpy_test.py#L57

jakevdp: For what it's worth, the reason for the use of `concatenate` as a method here is to support `jnp.concatenate` called on PRNGKey arrays (see https://github.com/google/jax/pull/8381)

jakevdp: We might need to modify this as part of this change: https://github.com/google/jax/blob/be8939771c9abbe384df1da5a970b306fad2e965/jax/_src/prng.py#L206-L209

jenkspt: Ok so I'm looking into `jax.random.PRNGKeyArray`. I'm seeing that the `dtype` attribute is deprecated -- but also that it is always `uint32` even when x64 is enabled (or is this implementation dependent?). 
The most recent changes just raise an error in the `hasattr(arrays[0], "concatenate")` branch if the dtype isn't `uint32` or None.

Alternatively I can update the `jax.random.PRNGKeyArray.concatenate` method to accept the dtype -- and maybe sanitize there?

jax/_src/numpy/lax_numpy.py
```diff
@@ -1665,26 +1665,32 @@ def _concatenate_array(arr, axis: int):
...
   _stackable(*arrays) or _check_arraylike("concatenate", *arrays)
   if not len(arrays):
     raise ValueError("Need at least one array to concatenate.")
   if ndim(arrays[0]) == 0:
     raise ValueError("Zero-dimensional arrays cannot be concatenated.")
   if axis is None:
-    return concatenate([ravel(a) for a in arrays], axis=0)
+    return concatenate([ravel(a) for a in arrays], axis=0, dtype=dtype)
   if hasattr(arrays[0], "concatenate"):
+    if (dtype is not None) and (dtype != np.uint32):
```
jakevdp, line 1679: I think we should forward `dtype` to the `concatenate` method here, and move this check into the PRNGKeyArray method. And we should probably error for any non-None dtype there, since `dtype` is not normally a user-specifiable parameter for PRNGKeyArray objects.

jenkspt: Should I add `PRNGKeyArray` specific concatenate tests as well?
Also need to update https://github.com/google/jax/blob/be8939771c9abbe384df1da5a970b306fad2e965/tests/lax_numpy_test.py#L2336-L2348

jax/_src/numpy/lax_numpy.py
jenkspt: > 

I don't see `numpy_version` being used anywhere in lax_numpy_test.py -- can we replace this with `Version(np.__version__)`?

jax/_src/numpy/lax_numpy.py
jakevdp: They already exist here: https://github.com/google/jax/blob/75d69725c384cd08b0ca55fa9e315fb9bb5830e4/tests/random_test.py#L1621-L1627

jax/_src/prng.py
```diff
@@ -203,10 +203,12 @@ def reshape(self, newshape, order=None):
     reshaped_keys = jnp.reshape(self._keys, (*newshape, -1), order=order)
     return PRNGKeyArray(self.impl, reshaped_keys)
 
-  def concatenate(self, key_arrs, axis):
+  def concatenate(self, key_arrs, axis, dtype=None):
+    if dtype:
```
jakevdp, line 207:
```suggestion
    if dtype is not None:
```

jax/_src/prng.py
```diff
@@ -203,10 +203,12 @@ def reshape(self, newshape, order=None):
...
+      raise ValueError('dtype argument not supported for concatenating PRNGKeyArray')
     axis = canonicalize_axis(axis, self.ndim)
     arrs = [self._keys, *[k._keys for k in key_arrs]]
-    return PRNGKeyArray(self.impl, jnp.concatenate(arrs, axis))
+    return PRNGKeyArray(self.impl, jnp.concatenate(arrs, axis, dtype=dtype))
```
jakevdp, line 211: I think passing `dtype` here is confusing given the `ValueError` above.

tests/lax_numpy_test.py
```diff
@@ -25,6 +25,7 @@
 import unittest
 from unittest import SkipTest
 import warnings
+from packaging.version import Version
```
jakevdp, line 28: No need to use `packaging.version` here: we already define `numpy_version` below.

tests/lax_numpy_test.py
```diff
@@ -2299,25 +2300,28 @@ def testCompressMethod(self, shape, dtype, axis):
...
     wrapped_axis = axis % len(base_shape)
     shapes = [base_shape[:wrapped_axis] + (size,) + base_shape[wrapped_axis+1:]
               for size, _ in zip(itertools.cycle([3, 1, 4]), arg_dtypes)]
     def np_fun(*args):
       args = [x if x.dtype != jnp.bfloat16 else x.astype(np.float32)
               for x in args]
-      dtype = functools.reduce(jnp.promote_types, arg_dtypes)
-      return np.concatenate(args, axis=axis).astype(dtype)
-    jnp_fun = lambda *args: jnp.concatenate(args, axis=axis)
+      if Version(np.__version__) < Version('1.20'):
```
jakevdp, line 2320:
```suggestion
      if numpy_version < (1, 20):
```

jakevdp (changes requested):

jakevdp (approved):
Looks great! To fix the CLA issue (and also meet the single-change commit policy) please squash your changes into a single commit; see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests

jakevdp:
Thanks for the contribution!

PR #11675, opened 2022-07-30
fix ad_checkpoint.checkpoint caching issue

mattjj:
The main action here is in ad_checkpoint.py, where I refactored the transpose rule for the new remat implementation so as to get cache hits. Without this change, the test api_test.py RematTest.test_vjp_caching was failing.

I also added a config option to switch to the new checkpoint implementation globally (default False for now), as the first step in replacing and then deleting old remat. And deleted the after_neurips flag.

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -4784,6 +4784,16 @@ class RematTest(jtu.JaxTestCase):
         f_vjp(1.)[0].block_until_ready()
     self.assertEqual(count[0], 1)  # fwd execute_trivial, backward_pass on bwd
 
+  @unittest.skipIf(not config.jax_new_checkpoint, "old remat recompiles here")
+  def test_fwd_caching(self):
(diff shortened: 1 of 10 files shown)


tests/api_test.py
```diff
@@ -5580,9 +5583,9 @@ def scanned_f(c, x):
       return out, out
 
     def f(xs):
-      return lax.scan(scanned_f, 1., xs)
+      return lax.scan(scanned_f, jnp.array(1., 'float32'), xs)
```
sharadmv, line 5586: Why is this an issue now?

sharadmv (approved):

tests/api_test.py
mattjj: Great q. It's because old remat was fast and loose with weak types, in a way that new remat noticed.

PR #11707, opened 2022-08-02
Add env variable bootstrapping to jax.distributed

nvcastet:
If you are on GPU platform, the following environment variables
  ``JAX_COORDINATOR_ADDRESS``, ``JAX_NUM_PROCESSES``, and ``JAX_PROCESS_ID``
  can be set to initialize `jax.distributed`.


hawkinsp:
Looks like this change got lost. Is this a change we still want to merge?

nvcastet:
@hawkinsp No I can close it, now that we can register cluster types to tell jax.distributed how to initialize, we probably do not need this one.
See https://github.com/google/jax/commit/412a5379c157e24fa3ffdea57df8c34e9e86afe1

PR #11717, opened 2022-08-03
Making sharding an arg to MHLO callback lowering

sharadmv:
This is just a small refactor that enables other potential callback primitives to choose different sharding strategy (say a `pure_callback` in the future).


mattjj (approved):
Could make a helper function for creating the sharding object.

Issue #8287, opened 2021-10-19
Use issue forms rather than issue templates?


jakevdp:
This is a relatively new feature, described here: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

As an example of how this may be used, take a look at scipy's new issue tabs: https://github.com/scipy/scipy/issues/new/choose

PR #11720, opened 2022-08-03
Add enhanced bug report template

jakevdp:
Addresses #8287

Unfortunately, I don't think there's any way to check the rendering on this without merging it.


froystig (approved):
We discussed/iterated on this a bit offline. Looks good!

PR #11721, opened 2022-08-03
Add multi-node multi-GPU JAX tests

### LINKED ISSUE #11648: Extend multi-process init test to multi node
A Multi-Node GPU backend init test doesn't exist and one is needed, [The multi-process test](https://github.com/nvcastet/jax/blob/7589c6d7f0e67767e6a0e2b20da75f28bfb8ff35/tests/distributed_test.py) can be extended to work on multi-node.

sudhakarsingh27:
This adds multi-node multi-GPU test for `jax.distributed.initialize`.
Presently, this is expected to run on a nightly basis. Under the hood,
SLURM is used to launch the `pytest <test_name>` commands on multiple
nodes.

Resolves: #11648

### DIFF
--- a/tests/distributed_multinode_test.py
+++ b/tests/distributed_multinode_test.py
@@ -1,0 +1,62 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import unittest
+
+import jax
+import jax._src.lib
+from jax._src import test_util as jtu
+import jax.numpy as jnp
+
+@unittest.skipIf(
+    os.environ.get("SLURM_JOB_NUM_NODES", None) != "2",
+    "Slurm environment with at least two nodes needed!")
+class MultiNodeGpuTest(jtu.JaxTestCase):
+
+  def test_gpu_multi_node_initialize_and_psum(self):
+
+    # Hookup the ENV vars expected to be set already in the SLURM environment
+    nodelist = os.environ.get("SLURM_STEP_NODELIST", None)
+    if nodelist is not None:
+      coordinator_address = nodelist.split('[')[0] + \
+                            nodelist.split('[')[1].split(',')[0]
+    num_tasks = os.environ.get("SLURM_NPROCS", None)
+    taskid = os.environ.get("SLURM_PROCID", None)
+    localid = os.environ.get("SLURM_LOCALID", None)
+
+    # fixing port since it needs to be the same for all the processes
+    port = "54321"
+
+    print(f"coord addr:port : {coordinator_address}:{port}\nTotal tasks: "
+          f"{num_tasks}\ntask id: {taskid}\nlocal id: {localid}")
+
+    self.assertEqual(
+        coordinator_address is None or num_tasks is None or taskid is None,
+        False)
+
+    jax.distributed.initialize(coordinator_address=f'{coordinator_address}:{port}',
+                               num_processes=int(num_tasks),
+                               process_id=int(taskid))
+
+    print(f"Total devices: {jax.device_count()}, Total tasks: {int(num_tasks)}, "
+          f"Devices per task: {jax.local_device_count()}")
+
+    self.assertEqual(jax.device_count(),
+                     int(num_tasks) * jax.local_device_count())
+
+    x = jnp.ones(jax.local_device_count())
+    y = jax.pmap(lambda x: jax.lax.psum(x, "i"), axis_name="i")(x)
+    self.assertEqual(y[0], jax.device_count())
+    print(y)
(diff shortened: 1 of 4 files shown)


tests/distributed_multinode_test.py
```diff
@@ -0,0 +1,63 @@
...
+    nodelist = os.environ.get("SLURM_STEP_NODELIST", None)
+    if nodelist is not None:
+      coordinator_address = nodelist.split('[')[0] + \
+                            nodelist.split('[')[1].split(',')[0]
+    num_tasks = os.environ.get("SLURM_NPROCS", None)
+    taskid = os.environ.get("SLURM_PROCID", None)
+    localid = os.environ.get("SLURM_LOCALID", None)
+
+    # fixing port since it needs to be the same for all the processes
+    port = "54321"
```
yashk2810, line 40: Use portpicker?

portpicker.unused_port()

tests/distributed_multinode_test.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2022 Google LLC
```
yashk2810, line 1: Why not just add this to distributed_test.py under a different class?

.github/workflows/nightly-ci-gpu.yml
```diff
@@ -0,0 +1,54 @@
+name: Nightly JAX CI on NVIDIA GPUs
```
yashk2810, line 1: rename this file to nightly-ci-multiprocess-gpu.yml?

tests/distributed_multinode_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    print(f"coord addr:port : {coordinator_address}:{port}\nTotal tasks: "
+          f"{num_tasks}\ntask id: {taskid}\nlocal id: {localid}")
+
+    self.assertEqual(
+        coordinator_address is None or num_tasks is None or taskid is None,
+        False)
+
+    jax.distributed.initialize(coordinator_address=f'{coordinator_address}:'
+                               f'{port}',
```
yashk2810, line 50: Keep the coordinator_addr and port on the same line. This formatting is weird. Its fine if it exceeds 80 characters.

tests/distributed_multinode_test.py
```diff
@@ -0,0 +1,63 @@
...
+                               num_processes=int(num_tasks),
+                               process_id=int(taskid))
+
+    print(f"Total devices: {jax.device_count()},Total tasks: {int(num_tasks)},"
```
yashk2810, line 54: Please add a space after commas. You can use black to format this file. I would strongly recommend that. Or use https://github.com/google/yapf with google style

tests/distributed_multinode_test.py
```diff
@@ -0,0 +1,63 @@
...
+          f"Devices per task: {jax.local_device_count()}")
+
+    self.assertEqual(jax.device_count(),
+                     int(num_tasks) * jax.local_device_count())
+
+    x = jnp.ones(jax.local_device_count())
+    y = jax.pmap(lambda x: jax.lax.psum(x, "i"), axis_name="i")(x)
+    self.assertEqual(y[0], jax.device_count())
+    print(y)
```
yashk2810, line 63: Add a new line at the end of the file.

jakevdp, line 63: Looks like flake8 is complaining about the lack of newline here.

tests/distributed_multinode_test.py
sudhakarsingh27: Multiple processes might pick different ports since the launcher is one level above python

tests/distributed_multinode_test.py
sudhakarsingh27: Sure I can do that. Just wanted to start with a separate file. Multi-node tests would have env vars like `SLURM...` and although I've added `skipIfs` to ensure that they don't interfere with other tests, I'd like to know how everyone feels here. 
@skye @hawkinsp @jakevdp

tests/distributed_multinode_test.py
sudhakarsingh27: 😄 Okay, done

tests/distributed_multinode_test.py
sudhakarsingh27: Yapf suggests the following, but it was too long and I broke it up to contain within 80 chars limit. If it's fine to cross 80 chars, I've added spaces after commas

```
print(
        f"Total devices: {jax.device_count()}, Total tasks: int(num_tasks), Devices per task: {jax.local_device_count()}"
    )
```

.github/workflows/nightly-ci-gpu.yml
sudhakarsingh27: Is it okay if we keep it generic now? Or do we need to keep it specific now and make it generic later?

yashk2810: I would prefer that (keep specific then generic). I feel that we should name the file as to what its doing. If it does more things we can always rename

tests/distributed_multinode_test.py
yashk2810: You can always split the print on 2 lines.

.github/workflows/nightly-ci-gpu.yml
sudhakarsingh27: Okay, renamed the file now.

tests/distributed_multinode_test.py
sudhakarsingh27: Sure, done that now.

tests/distributed_multinode_test.py
jakevdp: I think for now we can keep this as a separate test, since we're still iterating on its exact content & setup, but we'll revisit this in the future

yashk2810: Then let's rename this: `distributed_multiprocess_gpu_test.py`. The reason I said to merge is because distributed_test.py was created for this exact reason.

jakevdp:
Looks great! Everything is connected, and the tests are running & passing.

Once we squash the commits, I'll get this merged. Thanks for all the work on this!

sudhakarsingh27:
Squashed the commits as requested. Pls lmk if there's anything else needed from my side. Thanks for reviewing!

jakevdp (approved):
Thanks!

PR #11725, opened 2022-08-04
Throw error earlier for misformatted string in jax.debug.print

sharadmv:


jax/_src/debugging.py
```diff
@@ -156,5 +156,6 @@ def debug_print(fmt: str, *args, ordered: bool = False, **kwargs) -> None:
       w.r.t. other ordered ``debug_print`` calls.
     **kwargs: Additional keyword arguments to be formatted.
   """
+  fmt.format(*args, **kwargs)
```
mattjj, line 159: lol

mattjj (approved):

jax/_src/debugging.py
sharadmv: you gotta do what you gotta do

PR #11727, opened 2022-08-04
[call_tf] Improve error reporting for calling TF functions returning non-arrays

gnecula:
Add more checks to catch early the cases when the called TF function
returns values that are not convertible to JAX values (arrays of
numeric values). All these cases were resulting in errors even before
but sometimes these errors were deep in the stack and harder to
diagnose.


gnecula:
@rxsang PTAL

gnecula:
I forgot why I closed this PR, re-opening to look more into this.

PR #11739, opened 2022-08-04
bug-report.yaml: fix yaml syntax error

jakevdp:
Related to #11720


froystig (commented):
Markdown links aren't being formatted as we might expect?

PR #11743, opened 2022-08-04
prepare to switch to new remat

mattjj:
This commit involves a few things, which are all united in being about landing the new remat (aka new checkpoint) implementation:
  * add benchmarks for new remat eager performance, and some caching to make those benchmarks fast
  * warn when the old-remat-exclusive `concrete` feature is used, with an actionable message pointing to the new recommended approach involving static_argnums
  * add the static_argnums parameter to both new and old remat
  * update docstrings (and de-duplicate them too)
  * add new tests, especially around caching and errors/warnings

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -14,7 +14,7 @@
 
 from functools import partial
 import operator as op
-from typing import Callable, Optional, List, Tuple, Sequence, Union
+from typing import Callable, Optional, List, Tuple, Sequence, Union, Any
 import types
 
 import jax
@@ -96,8 +96,9 @@ checkpoint_policies = types.SimpleNamespace(
 
 ### Main API
 
-def checkpoint(fun: Callable, prevent_cse: bool = True,
-               policy: Optional[Callable[..., bool]] = None
+def checkpoint(fun: Callable, *, prevent_cse: bool = True,
+               policy: Optional[Callable[..., bool]] = None,
+               static_argnums: Union[int, Tuple[int, ...]] = (),
                ) -> Callable:
   """Make ``fun`` recompute internal linearization points when differentiated.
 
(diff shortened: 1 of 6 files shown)


jax/_src/ad_checkpoint.py
```diff
@@ -218,11 +253,90 @@ def fun_remat(*args, **kwargs):
...
+    msg, = e.args
+    new_msg = msg + "\n\n" + (
+        "Consider using the `static_argnums` parameter for `jax.remat` or "
+        "`jax.checkpoint`. See the `jax.checkpoint` docstring and its example "
+        "involving `static_argnums`:\n"
+        "https://jax.readthedocs.io/en/latest/_autosummary/jax.checkpoint.html"
+        "\n")
+    new_e = core.ConcretizationTypeError.__new__(core.ConcretizationTypeError)
+    new_e.args = (new_msg,)
+    raise new_e from None
```
froystig, line 339: Do we know that _all_ concretization errors that happen under here are related to the `checkpoint`'s `static_argnums`? If so, then this looks good, maybe just highlight this fact in the comment a few lines up.

jax/_src/ad_checkpoint.py
```diff
@@ -203,13 +205,46 @@ def checkpoint(fun: Callable, prevent_cse: bool = True,
   ...     f2 = recursive_checkpoint(funs[len(funs)//2:])
   ...     return lambda x: f1(jax.checkpoint(f2)(x))
   ...
+
+  If ``fun`` involves Python control flow which depends on argument values,
+  it may be necessary to use the ``static_argums`` parameter. For example,
```
froystig, line 210:
```suggestion
  If ``fun`` involves Python control flow that depends on argument values,
  it may be necessary to use the ``static_argnums`` parameter. For example,
```

jax/_src/ad_checkpoint.py
```diff
@@ -203,13 +205,46 @@ def checkpoint(fun: Callable, prevent_cse: bool = True,
...
+      if is_training:
+        ...
+      else:
+        ...
+
+  Here, the use of ``static_argnums`` is necessary because the ``if`` statement
+  depends on the value of ``is_training``. The only cost to using
+  ``static_argnums`` is more retracing overheads: in the example, ``foo`` must
+  be retraced for every new value of ``is_training``. In some cases it may
+  additionally be necessary to use ``jax.ensure_compile_time_eval``:
```
froystig, line 226:
```suggestion
  Here, the use of ``static_argnums`` allows the ``if`` statement's condition
  to depends on the value of ``is_training``. The cost to using
  ``static_argnums`` is that it introduces re-tracing overheads across calls:
  in the example, ``foo`` is re-traced every time it is called with a new value
  of ``is_training``. In some situations, ``jax.ensure_compile_time_eval``
  is useful as well:
```

froystig (approved):

jax/_src/ad_checkpoint.py
mattjj: Oh oops, great point. Actually we can check for this case! I'll add a check.

PR #11747, opened 2022-08-04
Reraise exception in test_prim for jax2tf.

shaobohou:
Without the reraise, this could potentially hide test failures


apaszke:
This fix seems to already have been applied. But thanks for sending the PR!

PR #11748, opened 2022-08-04
Annotate nn.initializers

NeilGirdhar:
This was done to expose an Initializers type annotation that can be used
in other libraries.

### DIFF
--- a/jax/_src/nn/initializers.py
+++ b/jax/_src/nn/initializers.py
@@ -18,9 +18,10 @@ used in Keras and Sonnet.
 """
 
 
-from typing import Any, Callable, Sequence, Union
+from typing import Any, Sequence, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
@jakevdp Would this fix the problems in https://github.com/google/jax/pull/9596 ?  I broadened two of the named-shape-annotated private functions.  The third seems to require `NamedShape` since it accesses `shape.total`.

jakevdp (approved):
Looks good; I think this should remedy the problems from the previous PR, but I'll run some tests to make sure. It will probably be early next week before I can get this merged.

jakevdp:
Readthedocs failure looks unrelated to this.

NeilGirdhar:
Thanks a lot!  No rush

PR #11758, opened 2022-08-05
small tweaks for bint ad

mattjj:
Also add some helper functions for padding definitions.

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -2497,6 +2497,18 @@ def _substitute_axis_sizes(env: Dict, aval: AbstractValue) -> AbstractValue:
 
 padding_rules: Dict[Primitive, Callable] = {}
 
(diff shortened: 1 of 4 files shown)


jax/interpreters/partial_eval.py
```diff
@@ -2497,6 +2497,19 @@ def _substitute_axis_sizes(env: Dict, aval: AbstractValue) -> AbstractValue:
 
 padding_rules: Dict[Primitive, Callable] = {}
 
+def def_trivial_padding(prim: Primitive) -> None:
+  if prim.multiple_results:
+    breakpoint()
```
hawkinsp, line 2502: Remove breakpoint

hawkinsp (approved):

jax/interpreters/partial_eval.py
mattjj: Oops.

PR #11759, opened 2022-08-05
JEP: Propose a new jaxlib Python package structure.

hawkinsp:


docs/jep/11759-python-package-structure.md
```diff
@@ -0,0 +1,142 @@
...
+*   Making the compiler/runtime layer of JAX pluggable makes it easier to add
+    support for new compilers (IREE) and new hardware. In particular, we can
+    support new hardware from other vendors as new plugin packages without
+    causing a further combinatorial explosion in the number of builds.
+
+## Notes
+
+*   Hardware plugins like `jaxlib-cuda` will be distributed as Python packages
+    that act as plugins for `jaxlib`. There are standard Python mechanisms for
+    [creating and discovering plugins](https://packaging.python.org/guides/creating-and-discovering-plugins/)
```
cliffwoolley, line 112: To my knowledge, the mechanisms mentioned here are for discovering plugins _that are installed_, not for discovering plugins _that exist_.  How would we notify a user who lacks such backend plugin(s) that their jaxlib will be hampered unless they further install something else (even on CPU, since apparently there would be a jaxlib-cpu optional backend as well, from what I'm understanding)?  If my understanding of all that is accurate, then doesn't this somewhat go against the opening premise that this is seeking to be
> Simpler for users: users would be able to install a complete `jax` on CPU with `pip install jax`.
?

docs/jep/11759-python-package-structure.md
```diff
@@ -0,0 +1,142 @@
...
+    between, say, CUDA or Python versions, so to make a `jaxlib` release we
+    build ~12 different copies of XLA on Linux alone. By refactoring the
+    `jaxlib` wheels, we would remove the need to build the CPU support
+    repeatedly for each CUDA version, or the need to rebuild the GPU compiler
+    for each Python version.
+*   Splitting the wheels will reduce the total size of the binary distribution.
+    The largest `jaxlib` wheels at present are the CUDA wheels (~150MB each);
+    after refactoring we would no longer need to build one CUDA wheel per Python
+    version. This would reduce the size of our GPU wheel distribution by a
+    factor of 4, and it would make it easier to ship GPU wheels on Pypi.
```
cliffwoolley, line 102: This thing of "build just one of X and not for every single python version" can also be accomplished by building one python-agnostic shared core library and then having the python bindings all in the same wheel that are minimal wrappers around that core.  See also https://github.com/tensorflow/tensorflow/pull/55941

hawkinsp: Indeed, this is the same idea just split across multiple wheels instead of having one wheel with multiple builds for different Python versions.

I think I prefer having a wheel split in here! It better describes to the Python packaging system what is happening and will degrade more gracefully when someone releases, say, a new Python version that we don't support. If we use separate packages, then `pip` will understand that it does not have a suitable wheel.

docs/jep/11759-python-package-structure.md
hawkinsp: Correct, this is for discovering the plugins you have installed.

`jaxlib-cpu` we should require as a hard dependency, at least for now.

For a CUDA build, we can do two things:
a) We can suggest the user install `jax[cuda]` that adds `jaxlib-cuda` as a dependency.
b) We could also add code to core `jax` that detects if the user has `/dev/nvidia` devices but no `jaxlib-cuda` and suggests installing `jaxlib-cuda` (and similarly for other hardware for which there exists a plugin).

docs/jep/11759-python-package-structure.md
cliffwoolley: > If we use separate packages, then pip will understand that it does not have a suitable wheel.

That's true.

Glad we're on the same basic page here, thanks.

docs/jep/11759-python-package-structure.md
cliffwoolley: > `jaxlib-cpu` we should require as a hard dependency, at least for now.

I wondered if that's how it would be done; I just didn't see that in the doc.

> `jax[cuda]`

Extras are a helpful strategy here, I agree.  (One possible gotcha is that the extra is on jax instead of on jaxlib, yet the == version match required from jaxlib to the extra and not from jax to the extra, which might get tricky.)

> We could also add code to core jax that detects if the user has /dev/nvidia devices but no jaxlib-cuda and suggests installing jaxlib-cuda (and similarly for other hardware for which there exists a plugin).

That would be helpful I think.  Please be aware that `/dev/nvidia*` device nodes, such as `/dev/nvidiactl`, are specific to Linux and to our discrete GPU drivers (Tegra drivers, for example, have `/dev/nvgpu` instead, and WSL2 has `/dev/dxg`; native Windows is another thing entirely).

ngam (commented):
Overall, I think this is a good proposal. I highlighted how we do portions of the proposed ideas in conda-forge already for reference.

Additional notes for your consideration:
- Performance: I have personally found that the TensorFlow PyPI packages often perform quite poorly compared to either optimized NGC containers form NVIDIA or recently the build from conda-forge. One of my little projects of late was to add small changes to the conda-forge TensorFlow that resulted in outperforming the highly optimized NGC containers. 
  - Relevant point: as JAX is maturing, it is important to think of the downsides of simply putting things away on PyPI --- does it make sense for your team to take a more active role in directing the conda-forge builds? (You are more than welcome!) Does it make sense to convince NVIDIA to publish an optimized container with JAX (like they do with TensorFlow and PyTorch)?

- Ease for "users": I understand the efforts to limit the bandwidth and storage and all that, but it is arguably easier for users to just download `pip install jax` and get it all at once, and not have random pieces be pulled in.
  - Relevant point: I think it is important to pause and think if the overarching "jax" package can be the one to install all the pieces together. Maybe that's what you're trying to accomplish by the first point above. For end users, and especially scientists trying to use JAX for their research, it would be easier to just have it installed without having to think about all the pieces behind the scenes. Sadly, PyPI doesn't support optimized hardware matching --- but conda-forge does to a small extent, so another case for work with conda-forge :P 
- Ease of "developers": I would say this JEP is good for developers!

docs/jep/11759-python-package-structure.md
```diff
@@ -0,0 +1,142 @@
...
+
+Status: draft
+
+## What are we proposing?
+
+This document describes a plan to change the package structure of JAX's
+open-source release, splitting the hardware-specific parts of JAX into their own
+Python packages.
+
+In open-source, JAX is currently distributed as two Python wheels:
```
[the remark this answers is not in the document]
hawkinsp: Yes. We don't use Python packages internally: we build everything, e.g., the Python interpreter, all necessary modules, etc. for each Python program, bundled as a statically-linked monolithic binary built with Bazel. It's the same code, but we bundle it very differently! This is highly unusual and unlike how most people use Python modules, and we don't expect anyone else will want to do that. The fact that we do this has little bearing on this discussion, though!

hawkinsp:
> Overall, I think this is a good proposal. I highlighted how we do portions of the proposed ideas in conda-forge already for reference.
> 
> Additional notes for your consideration:
> 
> * Performance: I have personally found that the TensorFlow PyPI packages often perform quite poorly compared to either optimized NGC containers form NVIDIA or recently the build from conda-forge. One of my little projects of late was to add small changes to the conda-forge TensorFlow that resulted in outperforming the highly optimized NGC containers.

I expect that we will see less significant differences for JAX, because it depends heavily on JIT compilation: i.e., it matters less exactly how `jaxlib` was built.

>   
>   * Relevant point: as JAX is maturing, it is important to think of the downsides of simply putting things away on PyPI --- does it make sense for your team to take a more active role in directing the conda-forge builds? (You are more than welcome!)

I think we would be willing in principle to be more involved in `conda-forge` builds, however the key thing that is stopping us being involved with `conda-forge` is the non-commercial nature of Anaconda's repositories. It might be possible to solve that.

> Does it make sense to convince NVIDIA to publish an optimized container with JAX (like they do with TensorFlow and PyTorch)?

It's possible!

> * Ease for "users": I understand the efforts to limit the bandwidth and storage and all that, but it is arguably easier for users to just download `pip install jax` and get it all at once, and not have random pieces be pulled in.

The intended end state is:
```
pip install jax   # Downloads JAX for CPU, including all necessary dependencies (i.e., `jax`, `jaxlib`, `jaxlib-cpu`)
pip install jax[cuda]  # Downloads JAX for CUDA, including all necessary dependencies (`jax`, `jaxlib`, `jaxlib-cpu`, `jaxlib-cuda`)
... etc ...
```
which I think is reasonably user friendly?

hawkinsp:
In a sense, this JEP has landed! We now ship both Google TPU and NVIDIA GPU wheels as JAX plugins (e.g., `jax-cuda12-plugin`). Closing because the exact design is slightly stale, but we've pretty much delivered on the intent, finally!

PR #11762, opened 2022-08-05
Fix jax numpy norm bug where passing `ord="inf"` always returns one

cglwn:
This fixes a silent bug where passing in `ord="inf"` always returns one. Calculating this norm requires passing `jnp.inf` instead.
```py
>>> import jax.numpy as jnp
>>> jnp.linalg.norm(jnp.array([10, 20]), ord=jnp.inf)
DeviceArray(20., dtype=float32)
>>> jnp.linalg.norm(jnp.array([10, 20]), ord="inf")
DeviceArray(1., dtype=float32)
```

Either it should calculate the infinity norm or raise an exception. With this change, we match numpy behavior and raise an exception.
```py
>>> import numpy as np
>>> np.linalg.norm(np.array([10, 20]), ord=np.inf)
20.0
>>> np.linalg.norm(np.array([10, 20]), ord="inf")
...
ValueError: Invalid norm order 'inf' for vectors
>>> jnp.linalg.norm(jnp.array([10, 20]), ord="inf") # (With this PR)
...
ValueError: Invalid order 'inf' for vector norm. Use 'jax.numpy.inf' instead.
```

This might break existing usage of `norm` that pass `"inf"`. I tried to search for instances of `jnp.linalg.norm(..., ord="inf", ...)` on GitHub and Google Search but couldn't find any matches.

### DIFF
--- a/jax/_src/numpy/linalg.py
+++ b/jax/_src/numpy/linalg.py
@@ -469,6 +469,13 @@ def norm(x, ord=None, axis : Union[None, Tuple[int, ...], int] = None,
       # code has slightly different type promotion semantics, so we need a
       # special case too.
       return jnp.sum(jnp.abs(x), axis=axis, keepdims=keepdims)
+    elif isinstance(ord, str):
+      msg = f"Invalid order '{ord}' for vector norm."
+      if ord == "inf":
+        msg += "Use 'jax.numpy.inf' instead."
+      if ord == "-inf":
+        msg += "Use '-jax.numpy.inf' instead."
+      raise ValueError(msg)
     else:
       abs_x = jnp.abs(x)
       ord = lax_internal._const(abs_x, ord)
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -539,6 +539,11 @@ class NumpyLinalgTest(jtu.JaxTestCase):
                             tol=1e-3)
     self._CompileAndCheck(jnp_fn, args_maker)
 
+  def testStringInfNorm(self):
+    err, msg = ValueError, r"Invalid order 'inf' for vector norm."
+    with self.assertRaisesRegex(err, msg):
+      jnp.linalg.norm(jnp.array([1.0, 2.0, 3.0]), ord="inf")
+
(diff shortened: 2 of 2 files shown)


jax/_src/numpy/linalg.py
```diff
@@ -469,12 +470,19 @@ def norm(x, ord=None, axis : Union[None, Tuple[int, ...], int] = None,
       # code has slightly different type promotion semantics, so we need a
       # special case too.
       return jnp.sum(jnp.abs(x), axis=axis, keepdims=keepdims)
-    else:
+    elif isinstance(ord, numbers.Number):
```
cglwn, line 473: Technically the $p$-norm is only defined if $p > 1$. To match numpy, I allow $p \leq 1$. Should we diverge from numpy and add a check that it's a valid norm? The math still works for undefined $p$ values so there aren't any numerical issues.

jax/_src/numpy/linalg.py
```diff
@@ -469,12 +470,19 @@ def norm(x, ord=None, axis : Union[None, Tuple[int, ...], int] = None,
...
       abs_x = jnp.abs(x)
       ord = lax_internal._const(abs_x, ord)
       ord_inv = lax_internal._const(abs_x, 1. / ord)
       out = jnp.sum(abs_x ** ord, axis=axis, keepdims=keepdims)
       return jnp.power(out, ord_inv)
+    else:
```
cglwn, line 479: Should we just automatically convert `"inf"` to `jnp.inf`? It might be easier to use.

jakevdp: maybe at the top we could do something like:
```python
if isinstance(ord, str):
  ord = float(ord)
```
Would that be sufficient to address the bug?

jakevdp: Oh, I see, the desire is to fail for string input. In that case I'd say detect if `ord` is a string and then raise an appropriate `ValueError`

cglwn: Sounds good, I've changed it to do that.

jakevdp (approved):
Looks great, thanks!

PR #11768, opened 2022-08-05
prototype unfettered element types in jaxpr arrays

mattjj:
Intensely co-authored with @froystig

```ocaml
{ lambda ; a:foo[3,4,5]. let
    b:f32[3,4,5] = scan[
      jaxpr={ lambda ; c:foo[4,5]. let
          d:foo[5,4] = bake c
          e:foo[4,5] = bake d
          f:f32[4,5] = take e
        in (f,) }
      length=3
    ] a
  in (b,) }
```

From where comes the set of element types in jaxprs? Historically, from NumPy and XLA element types. But why would jaxprs be constrained to those? After all, jaxprs are just symbols, my friends. Those symbols need to be grounded when we translate to another compiler's IR, or when we have input or output values with a jaxpr evaluation. So if we're lowering we need ways to map jaxpr types to lowered IR types, and also ways to map any operations allowed on these types to lowered IR operations. And we may want Python objects representing values of these types. But once we have those mappings we don't need to be limited by NumPy/XLA element types.

Within jaxprs, we also need to handle transformations with these types.

In this change we started unfettering jaxpr element types from their vestigial NumPy/XLA constraints. Concretely, that means:
  * allowing `ShapedArray` to have any object for its 'dtype' attribute
  * added `core.custom_eltype` set
  * extended existing handlers for `ShapedArray` to call the corresponding custom
    element type handlers
  * mlir lowerings of some fully-element-type-polymorphic primitives
  * tests

In this PR, we only actually use these new extension points in tests.

The applications to come that we have in mind are:
  * arrays of prngkeys (and even custom prngs, as well as reuse error checking)
  * arrays of bounded int type for dynamic shapes (and especially raggedness)
  * `float0` arrays

We do not have in mind opening these mechanisms up to users. Think of these as yet another JAX-internal extension point, like all our existing 'handler' tables.

Jargon-wise, we may want to distinguish:
  * 'eltype' meaning jaxpr element types
  * 'dtype' meaning numpy dtypes (an existing convention)
  * 'etype' meaning hlo/mhlo element types (an existing convention)

But the code doesn't model this jargon at the moment, since we left a lot of attributes and helper functions referring to 'dtype'.

We haven't yet handled all the element-type-polymorphic primitives. Here's the list we've thought of so far:
  * [x] broadcast
  * [ ] reshape
  * [x] transpose
  * [ ] pad
  * [x] slice, dynamic_slice, dynamic_update_slice
  * [ ] concatenate
  * [ ] all_to_all, gather, scatter, all_gather, collective_permute
  * [x] make empty scalar (only appears in internal-about-to-lower-jaxpr dialect)

That last one is interesting: we introduced it so that the scan lowering rule, which lowers first to a "lowered jaxpr dialect" involving only those eltypes which correspond to etypes and involving only while_loop, ds/dus, etc, can be made simpler. Otherwise we'd need scan, itself a fully-eltype-polymorphic primitive, have a more    complicated lowering rule.

We also haven't handled AD. Our main applications (at least the first two listed above) don't involve AD types, so it seemed good to skip for now.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2938,7 +2938,7 @@ def _valid_jaxtype(arg):
   try:
     xla.abstractify(arg)  # faster than core.get_aval
   except TypeError:
-    return False
+    return core.valid_jaxtype(arg)
   else:
     return True
 
--- a/jax/_src/api_util.py
+++ b/jax/_src/api_util.py
@@ -435,7 +435,11 @@ def _shaped_abstractify_slow(x):
 
   weak_type = getattr(x, 'weak_type', False)
   named_shape = getattr(x, 'named_shape', {})
-  return core.ShapedArray(np.shape(x), _dtype(x), weak_type=weak_type,
+  if hasattr(x, 'dtype'):
+    dtype = dtypes.canonicalize_dtype(x.dtype)
+  else:
+    dtype = dtypes.result_type(x)  # TODO(frostig,mattjj): why this case?
+  return core.ShapedArray(np.shape(x), dtype, weak_type=weak_type,
                           named_shape=named_shape)
 
 # TODO(mattjj,yashkatariya): replace xla.abstractify with this, same behavior
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -651,6 +651,8 @@ def array_result_handler(sticky_device: Optional[Device],
   if aval.dtype == dtypes.float0:
     return lambda _, __: np.zeros(aval.shape, dtypes.float0)
   aval = core.raise_to_shaped(aval)
+  if type(aval.dtype) in core.custom_eltypes:
+    return aval.dtype.result_handler(sticky_device, aval)
   handler = lambda _, b: _maybe_create_array_from_da(b, aval, sticky_device)
   handler.args = aval, sticky_device  # for C++ dispatch path in api.py
   return handler
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -25,6 +25,7 @@ from typing import Any, Dict, List
 
 import numpy as np
 
+import jax
 from jax._src.config import flags, config
 from jax._src.lib import xla_client
 
@@ -85,6 +86,8 @@ def _to_complex_dtype(dtype):
 @functools.lru_cache(maxsize=None)
 def _canonicalize_dtype(x64_enabled, dtype):
   """Convert from a dtype to a canonical dtype based on config.x64_enabled."""
+  if type(dtype) in jax.core.custom_eltypes:
+    return dtype
   try:
     dtype = np.dtype(dtype)
   except TypeError as e:
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -29,11 +29,13 @@ from jax.interpreters import batching
 from jax.interpreters import mlir
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
+import jax._src.pretty_printer as pp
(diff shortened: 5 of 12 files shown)


jpienaar:
My first reaction when I read description "ooh, I can have Index type" :)

froystig (approved):

shoyer:
Just my two cents, but I would consider still calling these "JAX dtypes" (at least in the public APIs) rather than inventing a new name.

froystig:
Yeah, we had decided the same! Glad you agree. Haven't followed up yet (neither in comments, nor in code).

gnecula:
While working to adapt the lowering rules to dynamic shapes I noticed that the custom lowering rules for opaque types are written to return a sequence of MHLO ops rather than just a MHLO op. I am wondering if this was intended (for some future extensibility) or it was just an accident. 

For example, I want to write a helper function for generating `mhlo.BroadcastInDimOp` or `mhlo.DynamicBroadcastInDimOp` depending on whether the output value is statically shaped or not. If this helper function also wants to handle opaque types, and call the `_rules.broadcast_in_dim` then the helper function is forced to return a sequence of `ir.Value`. This makes the helper function not usable in many other contexts where we want to broadcast during lowering, e.g., in linalg.py lowering rules. Those contexts do not know how to handle a broadcast that results in multiple ops.

As another example, currently the [broadcast_in_dim lowering rule](https://github.com/google/jax/blob/440b25bf5dbacf609401f614bd9a8e59ceb17098/jax/_src/lax/lax.py#L2896) simply returns the result of  `_rules.broadcast_in_dim_mlir`, in which case it is acceptable that the result of the latter should be a singleton list of `ir.Op`. In fact, that lowering rule would not know what to do with multiple returned ops.

I am tempted to think that the custom lowering rules should return a single op. One thing that gives me pause is that the custom `physical_avals` seem to be intended to rewrite one `aval` into multiple physical ones. I have not seen this design issue discussed anywhere, so my question is if this is really intended. If it is, not many called of these custom rules would know what to do with multiple values, e.g., from a broadcast.

gnecula:
Thinking about it some more, JAX lowering rules already support avals that are represented after lowering by multiple IR values. Do we plan to support avals that are opaque and are represented by multiple IR values? If so, then I need to rethink who to write the lowering functions that I have in mind.

froystig:
> [...] JAX lowering rules already support avals that are represented after lowering by multiple IR values. Do we plan to support avals that are opaque and are represented by multiple IR values?

Yeah, the original intent was to support lowering to multiple IR values. It isn't necessary for either of our immediate uses of opaque dtypes: RNG key types (for existing RNG implementations) and in bounded ints. But (a), as you say, it's consistent with the multi-IR-value support we have in other lowering rules. And (b), some use cases are imaginable, like if we wanted to introduce int128 or float128 dtypes in Jaxpr, or if we wanted to pursue RNG reuse checking techniques that keep an extra "used" bit for each array element.

The imagined use cases are currently more hypothetical than not, so perhaps we can revisit the multi-IR-value lowering if it's too complicated to maintain.

froystig:
@gnecula – Considering that #13562 went in as it did, are we committing to "single return value" for now?

gnecula:
At the moment we only support single value. This can be revisited.

froystig:
Sounds good. Perhaps the fact that `physical_avals` is still plural, but then asserted to be of length 1 by the slice lowerings, is a fine reminder that slice lowerings are why we decided to constrain things for now. For example:

https://github.com/google/jax/blob/aeb425874b8090d877fc50a6532d28f12be48445/jax/_src/prng.py#L408-L410

It'd good to resolve the mismatch more wholly one way or another at some point.

PR #11770, opened 2022-08-05
Error in debug print if unused arguments are provided

sharadmv:
Why?

If you do something like this:
```python
jax.debug.print(f"{x}", x=x)
```
i.e use an f-string, you'll print out a traced `x` inside of the JIT instead of the actual value. 

After this PR, we'll raise an error saying that the provided `x=x` keyword argument to `jax.debug.print` was left unused and suggest that you might be using an f-string. This can help catch this issue earlier as opposed to after JIT compilation.

Thanks to @jakevdp for the suggestion!

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -14,6 +14,7 @@
 """Module for JAX debugging primitives and related functionality."""
 import enum
 import functools
+import string
 import sys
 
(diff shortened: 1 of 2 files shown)


mattjj (approved):

sharadmv:
Thank you!

PR #11773, opened 2022-08-06
handle test failure in diffrax under new remat

mattjj:
This change is to unblock landing new remat. I don't fully understand how, but diffrax's bounded_while_loop (a tricky bit of code!) is hitting a condition I thought was impossible. So let's handle that case!

I didn't add a test for it because I don't want to encode it as supported behavior. The only way I know how to test it would be to use the partial eval APis directly (not via AD) in a way we _don't_ want to support. Luckily handling it is actually simple.

Really I just want to unblock new-remat-by-default...

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -1363,6 +1363,10 @@ def _while_partial_eval_custom(saveable, unks_in, inst_in, eqn):
   body_jaxpr_known = core.ClosedJaxpr(jaxpr_known_, body_jaxpr.consts)
   del jaxpr_known_, carry_uk_out, num_res
 
+  # Instantiate all inputs (b/c jaxpr_staged will take all inputs).
+  new_inst = [x for x, inst in zip(eqn.invars, inst_in)
+              if type(x) is core.Var and not inst]
+
   # Compute the known part of cond_fun (basically pruning inputs on known side).
   cond_unks_in = cond_consts_uk + carry_uk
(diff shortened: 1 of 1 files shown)


jax/_src/lax/control_flow/loops.py
```diff
@@ -1363,13 +1363,22 @@ def _while_partial_eval_custom(saveable, unks_in, inst_in, eqn):
...
   # Compute the known part of cond_fun (basically pruning inputs on known side).
   cond_unks_in = cond_consts_uk + carry_uk
   cond_jaxpr_known_, _, [cond_uk], _, _ = \
       pe.partial_eval_jaxpr_custom(
           cond_jaxpr.jaxpr, cond_unks_in, in_inst=True,
           ensure_out_unknowns=False, ensure_out_inst=True,
           saveable=ad_checkpoint.nothing_saveable)
+  # NOTE(mattjj): I think it should be impossible for cond_uk to be unknown, but
+  # asserting that led to one failure in all of Google's tests, in diffrax. So
+  # we handle it: if cond_uk is unknown, stage out the whole cond.
```
froystig, line 1379:
```suggestion
  # NOTE(mattjj): I think it should be impossible for the condition to be
  # unknown, but asserting that caused a test failure in diffrax. So
  # we handle it: if it is unknown, stage out the whole cond function.
```
Was confused by "cond_uk is unknown" (and tried to maintain line count).

froystig (approved):

PR #11779, opened 2022-08-06
Increase minimum NumPy version to 1.20.

hawkinsp:
Per NEP 29, support for 1.19 ended on Jun 21, 2022.


jakevdp (approved):
Looks great, thanks!

PR #11793, opened 2022-08-08
Fix sample code block formatting in the FAQ

mbrukman:
Add missing colon to format sample code as a code block rather than a quote.

-----

### Current rendering on ([live](https://jax.readthedocs.io/en/latest/faq.html#why-are-gradients-zero-for-functions-based-on-sort-order), broken):

image

-----

### Preview of the docs in GitHub with the change (fixed):

Screen Shot 2022-08-08 at 12 39 19 PM


jakevdp (approved):
Looks good, thanks for the fix!

PR #11799, opened 2022-08-08
Create JAX Enhancement Proposals (JEPs).

hawkinsp:
Migrate existing design documents to JEPs.


jakevdp:
I think there are a few remaining places in the JAX package where `design_notes` URLs are references; we should update these to the new locations.

jakevdp:
e.g. https://github.com/google/jax/blob/022fedde98ed23bf4ea636e2f8a396ede68f82c9/jax/random.py#L69

froystig (approved):
Thanks for getting this started!

hawkinsp:
> I think there are a few remaining places in the JAX package where `design_notes` URLs are references; we should update these to the new locations.

Done.

PR #11803, opened 2022-08-08
Document multiprocess GPU support.

### LINKED ISSUE #2731: Support multinode training on GPU

hawkinsp:
Fixes #2731

### DIFF
--- a/docs/multi_process.md
+++ b/docs/multi_process.md
@@ -2,71 +2,113 @@
 
 ## Introduction
 
-This guide explains how to use JAX in environments such as [Cloud
-TPU](https://cloud.google.com/tpu) pods where accelerators are spread across
-multiple CPU hosts or JAX processes. We’ll refer to these as “multi-process”
-environments.
+This guide explains how to use JAX in environments such as
+GPU clusters and [Cloud TPU](https://cloud.google.com/tpu) pods where
+accelerators are spread across multiple CPU hosts or JAX processes. We’ll refer
+to these as “multi-process” environments.
 
 This guide specifically focuses on how to use collective communication
-operations (e.g. {func}`jax.lax.psum`) in multi-process settings, although other
-communication methods may be useful too depending on your use case (e.g. RPC,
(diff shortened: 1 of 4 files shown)


docs/multi_process.md
```diff
@@ -2,72 +2,111 @@
...
 
 Key concepts:
-*   You must run at least one JAX process per host.
-*   Each process has a distinct set of _local_ devices it can address. The
-    _global_ devices are the set of all devices across all processes.
-*   Use standard JAX parallelism APIs like {func}`~jax.pmap` and
-    {func}`~jax.experimental.maps.xmap`. Each process “sees” _local_ input and
+
+  * You must run at least one JAX process per host.
+  * You must initialize the cluster with {func}`jax.distributed.initialize`.
```
jekbradbury, line 27: This is not currently the case on TPU? (As in, most people don't call `jax.distributed.initialize` on multi-process Cloud TPU, but maybe they should/could?)

hawkinsp: It's not technically mandatory, but it's strongly recommended for health-checking and for making distributed checkpointing of GDAs work.

skye: I was gonna add this to the cloud_tpu_init code, so it works automatically on Cloud TPU pods. But I guess we can keep it as-is for now and change it later when I actually do that. I'm a bit worried this will confuse existing Cloud TPU users, but they probably already read this doc.

skye (approved):

docs/multi_process.md
mjsML: @jekbradbury +1
Maybe we should try to make that "no config" experience work across the board ... I believe @nvcastet is working on making this the case for GPU clusters too. I think this is ok for now, but I guess we need to think more about this.

mjsML (approved):
LGTM, we should add more documentation links to SLURM/K8s on GPUs as they become available :)

PR #11806, opened 2022-08-09
Various debugger improvements

sharadmv:
- disable globals
- can opt out of filtering frames
- can limit number of frames

### DIFF
--- a/jax/_src/debugger/core.py
+++ b/jax/_src/debugger/core.py
@@ -80,7 +80,7 @@ class DebuggerFrame:
     return DebuggerFrame(
         filename=frame_info.filename,
         locals=frame_info.frame.f_locals,
-        globals=frame_info.frame.f_globals,
+        globals={},
         code_context=frame_info.code_context,
         source=source,
         lineno=frame_info.lineno,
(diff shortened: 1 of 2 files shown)


jax/_src/debugger/core.py
```diff
@@ -113,19 +114,45 @@ def register_debugger(name: str, debugger: Debugger, priority: int) -> None:
...
+def breakpoint(*, backend: Optional[str] = None, filter_frames: bool = True,
+               use_globals: bool = False, num_frames: Optional[int] = None,
+               ordered: bool = False, **kwargs):  # pylint: disable=redefined-builtin
+  """Enters a breakpoint at a point in a program.
+
+  Args:
+    backend: The debugger backend to use. By default, picks the highest priority
+      debugger and in the absence of other registered debuggers, falls back to
+      the CLI debugger.
+    filter_frames: Whether or not to filter out JAX internal frames.
```
mattjj, line 126:
```suggestion
    filter_frames: Whether or not to filter out JAX-internal stack frames from the traceback. Since some libraries, like Flax, also make user of JAX's stack frame filtering system, this option can also affect whether stack frames from libraries are filtered.
```

WDYT? Or some rephrasing with similar meaning.

jax/_src/debugger/core.py
```diff
@@ -113,19 +114,45 @@ def register_debugger(name: str, debugger: Debugger, priority: int) -> None:
...
+    use_globals: Whether or not to allow inspecting global variables in the
```
mattjj, line 127: I wonder if we should just delete this as an option. If you want to see a global, just assign it to a local, right? (And globals can't have JAX tracers, unless someone's doing mutation...)

jax/_src/debugger/core.py
```diff
@@ -113,19 +114,45 @@ def register_debugger(name: str, debugger: Debugger, priority: int) -> None:
...
+      debugger.
+    num_frames: The number of frames above the current to inspect in the
```
mattjj, line 129:
```suggestion
    num_frames: The number of frames above the current stack frame to inspect in the
```

mattjj: Is the 'inspect' word standard? It might be the kind of word that makes sense if you already know (or can guess) what's going on, but isn't clear enough to someone with low context.

I wonder if we can try to use language mentioned in pdb's documentation. E.g. is 'inspect' used there?

An alternative to "inspect" might be "make available for inspection in the interactive debugger" or something. Just brainstorming; up to you.

jax/_src/debugger/core.py
```diff
@@ -113,19 +114,45 @@ def register_debugger(name: str, debugger: Debugger, priority: int) -> None:
...
+      debugger.
+    ordered: A keyword only argument used to indicate whether or not the
+      staged out computation will enforce ordering of this ``debug_print``
+      w.r.t. other ordered ``debug_print`` calls.
```
mattjj, line 133:
```suggestion
      with respect to other ordered ``debug_print`` calls.
```

mattjj (approved):

jax/_src/debugger/core.py
sharadmv: Yea that's fair. I removed it for (should be easy to add back if people want it)

jax/_src/debugger/core.py
sharadmv: `inspect` does pop up in the `pdb` documentation but not frequently enough to justify using it. I removed use of "inspect" here.

sharadmv:
Thank you for the detailed review!

PR #11807, opened 2022-08-09
Add `spmd_axis_name` to vmap to allow constraining mapped PartitionSpecs.

pschuh:
This is added to allow `vmap(lambda x: with_sharding_constraint(x, None), spmd_axis_name = 'x')(x)` to lower to `with_sharding_constraint(x, P('x'))`. Previously, there was no way to specify explicitly what axis to apply to the `with_sharding_constraint`. This patch adds `spmd_axis_name` to BatchTrace so that the `with_sharding_constraint` batching rule can optionally constrain based on the provided axis.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1441,9 +1441,7 @@ def vmap(fun: F,
     axis_name: Optional, a hashable Python object used to identify the mapped
       axis so that parallel collectives can be applied.
     axis_size: Optional, an integer indicating the size of the axis to be
-      mapped. If not provided, the mapped axis size is inferred from arguments.
-    spmd_axis_name: Optional, a hashable Python object to insert into any
-      mapped PartitionSpecs.
+      mapped. If not provided, the mapped axis size is inferred from arguments..
 
   Returns:
     Batched/vectorized version of ``fun`` with arguments that correspond to
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1396,7 +1396,8 @@ def vmap(fun: F,
          in_axes: Union[int, Sequence[Any]] = 0,
          out_axes: Any = 0,
          axis_name: Optional[Hashable] = None,
-         axis_size: Optional[int] = None) -> F:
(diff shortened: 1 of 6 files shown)


mattjj (commented):
Seeing `spmd_axis_name` being plumbed through next to `axis_name` everywhere (both parts of the trace-global-data 'payload' on the MainTrace), I wonder if we could plumb this more generically. By 'generically' I mean like the `ctx` which now get plumbed into all our translation rules: it's both cleaner (fewer explicit arguments everywhere) and more future-proof (in case we need to add or remove some plumbed info) than keeping these things splatted into explicit arguments.

Maybe `axis_name`, `axis_size`, and `spmd_axis_name` could all be plumbed into the `payload`, and the payload could be plumbed through e.g. the rules/utilities for higher-order primitives parameterized by jaxprs? I'm not sure.

WDYT?

(If this is urgent to land, we could choose to land-and-iterate.)

pschuh:
How general would this be? Is it just a type for the batcher transform? In that case, i'm not so sure how useful that would be. This is slightly urgent though. People are training with patches.

mattjj:
Hm I'm not sure I understand, but the utility of what I'm proposing is avoiding having to plumb multiple explicit arguments through all these helper functions. That is, the usefulness is immediately realized in the JAX codebase, not hypothetical! I wouldn't mind landing that as a cleanup though.

jax/_src/api.py
```diff
@@ -1441,6 +1442,8 @@ def vmap(fun: F,
       axis so that parallel collectives can be applied.
     axis_size: Optional, an integer indicating the size of the axis to be
       mapped. If not provided, the mapped axis size is inferred from arguments.
+    spmd_axis_name: Optional, a hashable Python object to insert into any
+      mapped PartitionSpecs.
```
mattjj, line 1446: Can we remove the docs for now, or explicitly say this is an experimental feature and may change?

mattjj (approved):
LGTM to land and iterate.

PR #11808, opened 2022-08-09
Try flattening the locals/globals dict in the debugger and have a fallback if it fails

sharadmv:
Why? This catches issues with capturing the stack when it contains references to unflattenable dictionaries.

The debugger takes a snapshot of the stack trace at JAX trace time using `inspect.stack()`. Inside each stack frame returned by `inspect.stack()` are also snapshots of each frames local variables, which could include traced JAX values. The debugger, before this PR, would call `tree_flatten` on the locals in each stack frame and filter out non-JAX-compatible values (non-arrays and non-tracers).

The tree flattening machinery operates on dictionaries sorts the keys and treats the sorted keys as the tree's metadata. Unfortunately, for the debugger, if the locals contained "unflattenable" dictionaries (i.e. dictionaries that had keys that couldn't be compared to each other), the `tree_flatten` operation would error and throw an unhelpful error like `TypeError: '<' not supported between instances of 'function' and 'function'`.

With this change, we introduce two changes:
1. this error is caught and the unflattenable object is replaced with a sentinel indicating that it was unable to be flattened.
2. we don't tree flatten the locals dictionary directly and instead individually flatten each key-value pair. We store the flattened key-value pairs in another Pytree that allows us to convert it back into a dict easily. We do this because we want to flatten as much as possible without running into an error.

### DIFF
--- a/jax/_src/debugger/core.py
+++ b/jax/_src/debugger/core.py
@@ -14,11 +14,10 @@
 from __future__ import annotations
 
 import dataclasses
-import functools
 import inspect
 import threading
 
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, Hashable, List, Optional, Tuple
 from typing_extensions import Protocol
 
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


mattjj (commented):
Can you write more in the PR message about the motivation here? E.g.: why the alternative dict pytree registration, and how does it compare to the standard one? How did you run into this error?

jax/_src/debugger/core.py
```diff
@@ -29,35 +28,79 @@
...
+  def tree_unflatten(cls, keys, values):
+    return _DictWrapper(keys, values)
+
+
+class _CantFlatten:
+  __repr__ = lambda _: "<cant_flatten>"
+cant_flatten = _CantFlatten()
+
+def _safe_flatten_dict(dct: dict[Any, Any]) -> tuple[list[Any],
+                                                     tree_util.PyTreeDef]:
```
mattjj, line 57:
```suggestion
def _safe_flatten_dict(dct: dict[Any, Any]
                       ) -> tuple[list[Any], tree_util.PyTreeDef]:
```

jax/_src/debugger/core.py
```diff
@@ -29,35 +28,79 @@
...
+
 @tree_util.register_pytree_node_class
 @dataclasses.dataclass(frozen=True)
 class DebuggerFrame:
   """Encapsulates Python frame information."""
   filename: str
-  locals: Dict[str, Any]
-  globals: Dict[str, Any]
+  locals: Dict[Any, Any]
+  globals: Dict[Any, Any]
```
mattjj, line 77: I think these are always keyed by strings, but they might contain dicts which aren't keyed by strings, right?

mattjj: (Even for dicts which aren't keyed by strings, I think `Hashable` may be a better annotation than `Any`.)

tests/debugger_test.py
```diff
@@ -29,6 +29,7 @@
 import jax.numpy as jnp
 import numpy as np
 
+config.update("jax_traceback_filtering", "off")
```
mattjj, line 32: After this test exits, this config setting might affect subsequent tests. Maybe handle this with setUp / tearDown module-level functions or class-level methods?

jax/_src/debugger/core.py
```diff
@@ -29,35 +28,79 @@
 from jax._src import util
 import numpy as np
 
+
+@tree_util.register_pytree_node_class
+class _DictWrapper:
+  keys: list[Any]
```
mattjj, line 34:
```suggestion
  keys: list[Hashable]
```

jax/_src/debugger/core.py
sharadmv: Done.

sharadmv:
Updated the PR message.

mattjj (approved):
A++++++++++

Issue #3285, opened 2020-06-02
lax.while_loop fails with zero iterations


jakevdp:
Ran into a strange bug where a `lax.while_loop` failed unless jit was disabled; this is the distilled version:
```python
import jax.numpy as jnp
import jax

def loop_cumsum(vals):
  def body_fun(i, state):
    return state + vals[i]
  init = vals.dtype.type(0)
  return jax.lax.fori_loop(0, len(vals), body_fun, init)

with jax.disable_jit():
  print(loop_cumsum(jnp.arange(0)))
# 0

loop_cumsum(jnp.arange(0))
# IndexError: index is out of bounds for axis 0 with size 0
```
It appears that the `_initial_style_jaxpr` is indexing into the empty array.

mattjj:
Very nice short example!

I'm not sure, but upon reflection I think this might be working as intended: the loop body in some sense contains a type error, in that when `vals` has type `ShapedArray((0,), int32)` then there is no value for an integer `i` such that `vals[i]` is valid. So even though the loop body will be executed for zero iterations, the program is still invalid.

Analogously:

```c
int main(void) {
  int i = 0;
  while (0) {
    i[0] += 1;
  }
}
```

```
$ gcc foo.c
foo.c: In function ‘main’:
foo.c:4:6: error: subscripted value is neither array nor pointer nor vector
    4 |     i[0] += 1;
      |      ^
```

However, the potential inconsistency here is that in Python it's fine to have any syntactically correct loop body for a loop that executes zero iterations! So we're deviating from Python, though in a kind of reasonable way.

For this specific case, and perhaps an analogous one involving `scan`, we could detect at tracing/staging time that this `fori_loop` will execute zero iterations and thus match Python-ish behavior (essentially by skipping the loop entirely at trace time), but the issue would still be present for `while_loop` in general. Since this came up in a real use case, I think I'll try adding the one-liner (I hope) heuristic to make this work.

hawkinsp:
I hit this here too: https://github.com/google/jax/blob/dd81a8dded3191110a2db884581c1d4dacd364fb/jax/lax_linalg.py#L520

gnecula:
I think that while relaxing the constraints for a loop with no iterations makes more program executable by JAX, it also complicates the mental model by adding special cases to the general requirement that the body of while/for must be JAX traceable, and it is traced by JAX.

Issue #994, opened 2019-07-08
lax.scan fails for zero-length arrays


superbobry:
Reproducer:

```python
>>> lax.scan(lambda acc, x: (acc, None), 0.0, jnp.arange(0))
...
TypeError: slice slice_sizes must be less than or equal to operand shape, got slice_sizes (1,) for operand shape (0,).
````

I expect `lax.scan` to return `(0.0, [])` in this case.

**Comment by dynamicwebpaige** [comment]
Unable to reproduce; got the following as output:

```python
(0.0, None)
```
Feel free to reopen, if I'm mistaken. 🙂

PR #11810, opened 2022-08-09
Fix zero-length array bug in scan.

frederikwilde:
When using `scan` over an array of length zero with a function `f` that expects a specific input shape (as below), the compilation of `f` leads to an error. Here is a minimal example:
```Python
def f(carry, y):
    y1, y2 = y
    return carry + y1 * y2, None
ys = jnp.array([])
jax.lax.scan(f, 0, ys)  # raises: TypeError: iteration over a 0-d array
```
This happens because `scan` infers the abstract value of the arguments of `f` (specifically right [here](https://github.com/google/jax/blob/main/jax/_src/lax/control_flow/loops.py#L238)), which in the example above returns `ShapedArray(float32[])` which is the shape of `ys` when stripped of its first dimension, but obviously is not a valid argument shape for `f`.

The fix I implemented here seems very primitive to me, specifically because `f` won't be compiled in this case. I am not sure if this leads to any unexpected problems, but for my purposes it solves the problem.

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -220,6 +220,9 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
     else:
       length, = unique_lengths
 
+  if length == 0:
+    return init, xs
+
   if config.jax_disable_jit:
     if length == 0:
       raise ValueError("zero-length scan is not supported in disable_jit() mode because the output type is unknown.")


frederikwilde:
Just linking #994 for completeness.

jakevdp:
Thanks for looking at this – there is a similar issue in `while_loop` that we decided was working as intended: https://github.com/google/jax/issues/3285

I think the same argument could possibly be made here, that in some senses this scan body contains a type error. Maybe @mattjj can weigh in with thoughts?

jakevdp:
For example, here's a slightly modified version of your zero-length scan in which the types of the inputs match the body function:
```python
def f(carry, y): 
  y1, y2 = y 
  return carry + y1 * y2, None 
ys = jnp.empty((0, 2)) # <-- difference here
jax.lax.scan(f, 0, ys)                                                                                                                                                                                                                                                               
# (DeviceArray(0., dtype=float32), None)
```

frederikwilde:
Ah, I see. I encountered the problem in the following way (simplified for brevity):
```Python
ys = jnp.arange(10)
carry, results =  0, []
for i in range(11):
    carry, _ = scan(f, carry, ys[:i])
    carry = some_other_function(carry)
    carry, _ = scan(f, carry, ys[i:]
    results.append(carry)
```
With the solution you propose one would need to implement the edge cases `i=0` and `i=10` separately, since empty slices don't contain the full shape of the original array. I don't know whether this is reason enough to allow an -- in a strict sense -- incorrect type in this special case. Just wanted to mention how this actually appears in a practical case.

jakevdp:
Thanks - I'm not sure I understand the distinction you're making in the second example. Empty slices should maintain their shape along non-sliced axes, even for slice size zero; e.g.
```python
y = jnp.empty((10, 2))

print(y[:0].shape)
# (0, 2)

print(y[10:].shape)
# (0, 2)
```
```

frederikwilde:
Oh! Thanks. Then I did not reproduce the error correctly. I will have another look at it and write again soon.

frederikwilde:
Ok, this turned out to be a mistake. I generate several arrays `ys` over which I loop as described above, but one of them is indeed just an empty array of shape `(1, 0)`. I would say in this case I am really passing an argument of invalid type to `scan`. Thanks for the quick responses!

jakevdp:
Awesome, thanks for the update and glad you got the issue figured out!

PR #11818, opened 2022-08-09
nightly multiprocess test: create issue on failure

jakevdp:
Pushed from a branch on the main repo in order to test issue creation on forced failure (which is prevented from branches on forks)

Added a failure to the test so we can see the automatic issue creation.

### DIFF
--- a/.github/workflows/cat_slurm_logs.py
+++ b/.github/workflows/cat_slurm_logs.py
@@ -1,0 +1,45 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


jakevdp:
@sudhakarsingh27 any idea why the SLURM process is not terminating?

jakevdp:
We had a successful failure! The SLURM job failed on my `assert False` test: https://github.com/google/jax/runs/7756398222?check_suite_focus=true

And it led to the creation of this issue: https://github.com/google/jax/issues/11820

jakevdp:
It might be better if we paste the contents of the artifacts to the issue description... I might give that a try.

sudhakarsingh27:
> It might be better if we paste the contents of the artifacts to the issue description

I think it's a good idea, although there's output from multiple nodes in the artifacts. I guess uploading `output*0.txt` contents should be sufficient since if one of the processes failed, it's highly probable that the others would've failed too.

jakevdp:
I just added an attempt to put the formatted logs in the issue description... let's see if it works

jakevdp:
Reporting works! https://github.com/google/jax/issues/11822

I'm going to clean up and get ready to submit this.

sudhakarsingh27 (approved):
LGTM. Thanks for doing this!

PR #11825, opened 2022-08-10
fix type annotation on remat

mattjj:


jax/_src/api.py
```diff
@@ -3051,7 +3051,7 @@ def __init__(self, shape, dtype):
 def checkpoint(fun: Callable, *,
                concrete: bool = False,
                prevent_cse: bool = True,
-               static_argnums: Union[int, Tuple[int, ...], bool] = (),
+               static_argnums: Union[int, Tuple[int, ...]] = (),
```
mattjj, line 3054: I had temporarily entertained supporting "put a True to mean all arguments are static" but w/e.

froystig (approved):

PR #11832, opened 2022-08-10
Fix Python 3.11 compatibility problems.

hawkinsp:
Also needs https://github.com/tensorflow/tensorflow/pull/57085

In Python 3.11:
* We can't use a mutable value as a dataclass default.
* We can now set `__traceback__` to override exception tracebacks.

### DIFF
--- a/jax/_src/traceback_util.py
+++ b/jax/_src/traceback_util.py
@@ -182,10 +182,16 @@ def api_boundary(fun: C) -> C:
         e.__context__ = None
         e.__cause__ = unfiltered
 
-        # There seems to be no way to alter the currently raised exception's
-        # traceback, except via the C API. The currently raised exception
-        # is part of the interpreter's thread state: value `e` is a copy.
-        xla_extension.replace_thread_exc_traceback(filtered_tb)
+        e.__traceback__ = filtered_tb
+        # In Python < 3.11, there seems to be no way to alter the currently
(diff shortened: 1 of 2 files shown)


jax/_src/traceback_util.py
```diff
@@ -182,10 +182,14 @@ def reraise_with_filtered_traceback(*args, **kwargs):
...
-        # There seems to be no way to alter the currently raised exception's
-        # traceback, except via the C API. The currently raised exception
-        # is part of the interpreter's thread state: value `e` is a copy.
-        xla_extension.replace_thread_exc_traceback(filtered_tb)
+        e.__traceback__ = filtered_tb
+        # In Python < 3.11, there seems to be no way to alter the currently
+        # raised exception traceback, except via the C API. The interpreter
+        # keeps a separate copy of the traceback (exc_traceback) that is
+        # separate to the __traceback__ of exc_value. Python 3.11 removes
+        # exc_traceback and just setting __traceback__ is enough.
```
froystig, line 190:
```suggestion
        # exc_traceback and just setting __traceback__ is enough. Since  it
        # is no longer needed, the XLA extension no longer defines a
        # traceback-replacing method at Python 3.11 and onward.
```

froystig (approved):

jax/_src/traceback_util.py
hawkinsp: Done.

PR #11846, opened 2022-08-11
Fix bug with integer-typed Gelu

russbates:
Certain lines in gelu()  currently round down constants if called with integer types (sqrt_2_over_pi = np.sqrt(2 / np.pi).astype(x.dtype))

Cast the input array to the nearest float-like type to avoid this, as done for trigonometic functions.


hawkinsp:
Mind adding a test also? https://github.com/google/jax/blob/main/tests/nn_test.py is the right place.

russbates:
Done!

hawkinsp (approved):

PR #11854, opened 2022-08-11
Add eager pmap support

sharadmv:
JAX's `jax_debug_nan`s and `jax_disable_jit` flags are extremely useful for debugging JIT-compiled computations. Unfortunately, up to this point, they have not worked with `jax.pmap` (i.e. they do not disable the pmap compilation). This PR changes that! 

### Why is this useful/important?

After this PR (and after enabling the `jax_eager_pmap` flag), the `jax_disable_jit` flag now enables executing `jax.pmap`-ed functions **eagerly**. This makes `pmap` behave a lot more like `vmap` and therefore allows us to (at the very least) 
1. print values inside of a `pmap` with Python's `print`, 
3. use `pdb`/Python's built-in `breakpoint` while executing a `pmap`-ed function.

This is potentially very useful for debugging. When we inspect values inside of a eagerly `pmap`-ed function, we should see objects that carry around sharding information as well as their underlying arrays. Before, we'd see abstracted tracers (no values only shape/dtype information).

### How is this implemented? 

JAX works by stacking "interpreters" on top of each other. For example `jax.vmap(jax.grad(f))` works by stacking a `vmap` interpreter on top of a `grad` interpreter. There is always a default bottommost interpreter, however, called the `Eval` interpreter which just invokes the implementation rules of a primitive (i.e. `jnp.sin(x)` will evaluate sin of x).

Transformations that compile JAX functions (e.g. `jax.pmap`, `jax.jit`) swap out the `Eval` interpreter in favor of one that first builds a jaxpr (i.e. records all the primitive applications into a data structure). The jaxpr is then "lowered" into a more compiler friendly representation.

When `jax_disable_jit` is enabled and we are `pmap`-ing, this PR skips the jaxpr staging part entirely and swaps out the `Eval` interpreter in favor of a new `Map` interpreter. The `Map` interpreter keeps track of the name -> axis index mapping on each tracer and instead of calling primitive's default implementations like the `Eval` trace, it calls `jax.pmap` on them with JIT-compilation enabled.


jax/interpreters/pxla.py
```diff
@@ -918,6 +931,197 @@ def xla_pmap_impl(fun: lu.WrappedFun, *args,
...
+  return new_outvals
+
+def _map_schedule(idx: Tuple[Optional[int], ...]) -> List[Optional[int]]:
+  # In order to do a multi-map (a simultaneous map over several axes), we will
+  # nest several maps. Each time we do a map, we "remove" an input axis so we
+  # need to update the remaining map axes. For example, if we are to map over
+  # the axes 0, 3, and 4, we make three calls to pmap with in_axes as 0, 2, 2.
+  return [None if i is None else
+          i - sum(j is not None and j < i for j in idx[:l])
+          for l, i in enumerate(idx)]
```
soraros, line 987: Nit: maybe
```suggestion
  return [None if i is None else
          i - sum(j < i for j in filter(None, idx[:l]))
          for l, i in enumerate(idx)]
```

mattjj (approved):

PR #11859, opened 2022-08-11
JEP: Type Annotations

jakevdp:
This Jax Enhancement Proposal (JEP) introduces a roadmap for type annotations in JAX, addressing a number of goals, non-goals, and design decisions that need to be made.

Part of #12049.


NeilGirdhar:
I'm really happy to see this level of care about type annotations!  This is a great design document.  I love the division into three levels.

Just a few comments that are hopefully useful:

> Flexible annotation for decorators has been a [long-standing issue](https://github.com/python/mypy/issues/1927) in the mypy package, which was only recently resolved by the introduction of `ParamSpec` in [PEP 612](https://peps.python.org/pep-0612/), available starting in Python 3.10.
Because JAX follows [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), it cannot rely on Python 3.10 features until sometime after mid-2024.

Do we have to wait for Python 3.10?  `ParamSpec` is available as part of [typing-extensions](https://github.com/python/typing_extensions).  Jax already uses typing-extensions in various places.

>  When functions are decorated by jax transformations like `jit`, `vmap`, `grad`, etc. JAX will **strip all annotations**. The reason for this is that without the mechanisms of [PEP 612](https://peps.python.org/pep-0612/) there is no good way to do otherwise, and `ParamSpec` will not be available for use until Python 3.10.

This is an important point.  I have been waiting for MyPy to fix [their `ParamSpec` bugs](https://github.com/python/mypy/issues?q=is%3Aissue+is%3Aopen++label%3Atopic-paramspec+) before submitting [a fix for the annotation stripping](https://github.com/google/jax/compare/main...NeilGirdhar:jax:jit_annotation).  The MyPy team is rolling out fixes very often, so I'm not even sure, but the fix may already work.

> and functions returning a shape should always be `Tuple[int]`

Minor (common) error, this should be `Tuple[int, ...]`.

jakevdp:
> Do we have to wait for Python 3.10? ParamSpec is available as part of [typing-extensions](https://github.com/python/typing_extensions). Jax already uses typing-extensions in various places.

The sentence after the one you quoted mentions this possibility 😁

I think that given how in-flux this is (recently added, not fully supported by mypy, etc.) we can leave this kind of annotation for a later data once things have stabilized. What do you think?

jakevdp:
@NeilGirdhar – I edited the doc a bit to address some of those comments - let me know what you think

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+
+5. Conversely, outputs of functions and methods should be typed as strictly as possible: for example, for a JAX function that returns an array, the output should be annotated with `jnp.ndarray` rather than `ArrayLike`. Functions returning a dtype should always be annotated `np.dtype`, and functions returning a shape should always be `Tuple[int]` or a strictly-typed NamedShape equivalent. For this purpose, we will implement in {mod}`jax.typing` several strictly-typed analogs of the permissive types mentioned above, namely:
+
+  - `NDArray` (perhaps this could be equivalent to `jnp.ndarray`?)
+  - `DType` (perhaps this could be simply `np.dtype`?)
+  - `Shape`
+  - `NamedShape`
+  - etc.
+
+6. Aside from common typing protocols gathered in `jax.typing`, we should err on the side of simplicity, and avoid constructing overly-complex protocols for arguments passed to API functions, and instead use simple unions such as `Union[simple_type, Any]` in the case that the full type specification of the API cannot be succinctly specified. This is a comprimise that achieves the goals of Level 1 and 2 annotations, while punting on Level 3 in favor of simplicity.
```
8bitmp3, line 160:
```suggestion
6. Aside from common typing protocols gathered in `jax.typing`, we should err on the side of simplicity, and avoid constructing overly-complex protocols for arguments passed to API functions, and instead use simple unions such as `Union[simple_type, Any]` in the case that the full type specification of the API cannot be succinctly specified. This is a compromise that achieves the goals of Level 1 and 2 annotations, while punting on Level 3 in favor of simplicity.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+  One impact of this is that for the time being, when functions are decorated by jax transformations like `jit`, `vmap`, `grad`, etc. JAX will **strip all annotations**. This is because `ParamSpec` is still only partially supported; the PEP is slated for Python 3.10 (though it can be used before that via [typing-extensions](https://github.com/python/typing_extensions) and at the time of this writing mypy has a laundry-list of incompatibilities with the `ParamSpec`-based annotations (see [`ParamSpec` mypy bug tracker](https://github.com/python/mypy/issues?q=is%3Aissue+is%3Aopen++label%3Atopic-paramspec+)).
+  We will revisit this question in the future once support for such features stabilizes.
+
+3. JAX type annotations shoudl in general indicate the **intent** of APIs, rather than the implementation, so that the annotations become useful to communicate the contract of the API. This means that at times inputs that are valid at runtime may not be recognized as valid by the static type checker (a simple example is an arbitrary iterator passed in place of a shape in some function implementations).
```
8bitmp3, line 141: Smol typo:

```suggestion
3. JAX type annotations should in general indicate the **intent** of APIs, rather than the implementation, so that the annotations become useful to communicate the contract of the API. This means that at times inputs that are valid at runtime may not be recognized as valid by the static type checker (a simple example is an arbitrary iterator passed in place of a shape in some function implementations).
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+Although the two type checkers have broad overlap in their behavior, each presents its own unique corner cases, as evidenced by the numerous `#type: ignore` and `#pytype: disable` statements througout the JAX codebase.
+
+This creates friction in development: internal contributors may iterate until tests pass, only to find that on export their pytype-approved code falls afoul of mypy.
+For external contributors, it's often the opposite: a recent example is {jax-issue}`#9596` which had to be rolled-back after it failed internal Google continous checks.
+Each time we move a type annotation from Level 1 (`Any` everywhere) to Level 2 or 3 (stricter annotations), it introduces more of these potential sharp edges.
+
+### Challenge 2: Array duck-typing
+One particular challenge for annotating JAX code is its heavy use of duck-typing. An input to a function marked `Array` in general could be one of many different types: a jax `DeviceArray`, a numpy `np.ndarray`, a numpy scalar, a Python scalar, a Python sequence, an object with an `__array__` attribute, an object with a `__jax_array__` attribute, or any flavor of `jax.Tracer`.
```
8bitmp3, line 84: Nit: capitalization ("JAX", "NumPy")

```suggestion
One particular challenge for annotating JAX code is its heavy use of duck-typing. An input to a function marked `Array` in general could be one of many different types: a JAX `DeviceArray`, a NumPy `np.ndarray`, a NumPy scalar, a Python scalar, a Python sequence, an object with an `__array__` attribute, an object with a `__jax_array__` attribute, or any flavor of `jax.Tracer`.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+As we've seen above, JAX currently has type annotations that are a mish-mash of styles, and aimed at all three areas of type annotation application above.
+Partly, this comes from the fact that JAX's source code poses a number of unique challenges for Python's type annotation system. We'll outline them here.
+
+### Challenge 1: pytype, mypy and developer friction
+One challenge JAX faces is that package development must satisfy the constraints of two different static type checking systems, `pytype` (used by internal CI and internal Google projects) and `mypy` (used by external CI and external dependencies).
+Although the two type checkers have broad overlap in their behavior, each presents its own unique corner cases, as evidenced by the numerous `#type: ignore` and `#pytype: disable` statements througout the JAX codebase.
+
+This creates friction in development: internal contributors may iterate until tests pass, only to find that on export their pytype-approved code falls afoul of mypy.
+For external contributors, it's often the opposite: a recent example is {jax-issue}`#9596` which had to be rolled-back after it failed internal Google continous checks.
```
8bitmp3, line 80: Smol typo:

```suggestion
For external contributors, it's often the opposite: a recent example is {jax-issue}`#9596` which had to be rolled-back after it failed internal Google continuous checks.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+On occasion, they are due to real & subtle bugs in the behavior of pytype and/or mypy.
+In rare cases, they may be due to the fact that JAX uses Python patterns that are difficult or even impossible to express in terms of Python's static type annotation syntax.
+
+## Type Annotation Challenges for JAX
+As we've seen above, JAX currently has type annotations that are a mish-mash of styles, and aimed at all three areas of type annotation application above.
+Partly, this comes from the fact that JAX's source code poses a number of unique challenges for Python's type annotation system. We'll outline them here.
+
+### Challenge 1: pytype, mypy and developer friction
+One challenge JAX faces is that package development must satisfy the constraints of two different static type checking systems, `pytype` (used by internal CI and internal Google projects) and `mypy` (used by external CI and external dependencies).
+Although the two type checkers have broad overlap in their behavior, each presents its own unique corner cases, as evidenced by the numerous `#type: ignore` and `#pytype: disable` statements througout the JAX codebase.
```
8bitmp3, line 77: Smol typo:

```suggestion
Although the two type checkers have broad overlap in their behavior, each presents its own unique corner cases, as evidenced by the numerous `#type: ignore` and `#pytype: disable` statements throughout the JAX codebase.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+Shape = core.Shape
+
+def slice(operand: Array, start_indices: Sequence[int],
+          limit_indices: Sequence[int],
+          strides: Optional[Sequence[int]] = None) -> Array:
+  ...
+```
+For the purposes of static type checking, this use of `Array = Any` for array annotations puts no constraint on the argument values (`Any` is equivalent to no annotation at all) but it does serve as a form of useful in-code documentation for the developer.
+
+For the sake of generated documentation, the name of the alias gets lost (the [HTML docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.slice.html) for `jax.lax.slice` report operand as type Any), so the documentation benefit does not go beyond the source code.
```
8bitmp3, line 31: Nit: Use code font for `Any`

```suggestion
For the sake of generated documentation, the name of the alias gets lost (the [HTML docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.slice.html) for `jax.lax.slice` report operand as type `Any`), so the documentation benefit does not go beyond the source code.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
+# Type Annotation Roadmap for JAXS
+_jakevdp@_
+_Aug 2022_
+
+## Background
```
8bitmp3, line 5: Nit: Maybe add spacing after `#`/`##` section titles here and across the doc (ref: https://google.github.io/styleguide/docguide/style.html#add-spacing-to-headings)

For example:

```suggestion
## Background

```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+To some degree, type annotations and static type checking have become an integral part of many Python development workflows, and to this end we have added annotations in a number of places throughout the JAX API.
+The current state of type annotations in JAX is a bit patchwork, and efforts to add more have been hampered by more fundamental design questions.
+This doc attempts to summarize those issues and generate a roadmap for the goals and non-goals of type annotations in JAX.
+
+Why do we need such a roadmap? Better/more comprehensive type annotations are a frequent request from users, both internally and externally.
+In addition, we frequently receive pull requests from external users seeking to improve JAX's type annotations: it's not always clear to the JAX team member reviewing the code whether such contributions are beneficial, particularly when they introduce complex Protocols to address the challenges inherent to full-fledged annotation of JAX's use of Python.
+This document details JAX's goals and recommendations for type annotations within the package.
```
8bitmp3, line 13: General comment: Great TL;DR 👍

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Why Type Annotations?
+There are a number of reasons that a Python project might wish to annotate their code-base; I'll summarize them in this document as Level 1, Level 2, and Level 3.
```
8bitmp3, line 16: Nit: Maybe "I'll" -> "we'll", since across the doc you say "we" (e.g. "we may", "we could", "we have", etc), even though there's only one author.

```suggestion
There are a number of reasons that a Python project might wish to annotate their code-base. We'll summarize them in this document as Level 1, Level 2, and Level 3.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+
+This use of type checking requires going further than the simple aliases used above; for example, knowing that the `slice` function returns an alias of `Any` named `Array` does not add any useful information to the code completion engine. However, were we to annotate the function with a `DeviceArray` return type, the autocomplete would know how to populate the namespace of the result, and thus be able to suggest more relevant autocompletions during the course of development.
+
+JAX has begun to add this level of type annotation in a few places; one example is the `jnp.ndarray` return type within the `jax.random` package [[source](http://google3/third_party/py/jax/_src/random.py;l=359;rcl=445522278)]:
+```python
```
8bitmp3, line 41: Nit: Add new lines before and after this code block (ref: https://google.github.io/styleguide/docguide/style.html#add-spacing-to-headings)

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+On the one hand, this is nice because your static type analysis may catch faulty type annotations (for example, a case where a `DeviceArray` method is missing from the `jnp.ndarray` abstract class).
+
+On the other hand, this strictness can make the type checking process very brittle in packages that often rely on duck-typing rather than strict type-safe APIs.
+You'll currently find code comments like `#type: ignore` (for mypy) or `#pytype: disable` (for pytype) peppered throughout the JAX codebase in several hundred places.
+These typically represent cases where typing problems have arisen; they may be inaccuracies in JAX type annotations, or inaccuracies in the static type checker's ability to correctly follow the control flow in the code.
+On occasion, they are due to real & subtle bugs in the behavior of pytype and/or mypy.
+In rare cases, they may be due to the fact that JAX uses Python patterns that are difficult or even impossible to express in terms of Python's static type annotation syntax.
+
+## Type Annotation Challenges for JAX
```
8bitmp3, line 71: Nit: May be best to use sentence case. Because you use lower-case for `pytype` and `mype`, capitalizing some words may look strange. Applies here and across the entire doc. Ref: https://developers.google.com/style/capitalization?hl=en#capitalization-in-titles-and-headings

Example:

```suggestion
## Type annotation challenges for JAX
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,160 @@
...
+As we've seen above, JAX currently has type annotations that are a mish-mash of styles, and aimed at all three areas of type annotation application above.
```
8bitmp3, line 72: Nit:

- For users who aren't very proficient in English, maybe it's best to avoid using words like "mish-mash" (ref: Global audience https://developers.google.com/style/translation?hl=en#be-inclusive)
- "Seen" may not be 100% a11y-friendly. How about "demonstrated"?

Example:

```suggestion
As demonstrated above, JAX currently has type annotations that are a mixture of different styles, and aimed at all three areas of type annotation application above.
```

🤷

jakevdp:
@8bitmp3 Thanks for the comments! I addressed them in the most recent commit.

NeilGirdhar:
> The sentence after the one you quoted mentions this possibility grin

My mistake!!

> I think that given how in-flux this is (recently added, not fully supported by mypy, etc.) we can leave this kind of annotation for a later data once things have stabilized. What do you think?

Yes, of course!

docs/jep/11859-type-annotations.md
8bitmp3: @jakevdp Forgot to mention

> "... capitalize only the first word in the title, the first word in a subheading after a colon, and any proper nouns or other terms that are always capitalized a certain way."

From https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings

So:

```diff
- ## Level 1: type annotation is cool
+ ## Level 1: Type annotation is cool
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+To some degree, type annotations and static type checking have become an integral part of many Python development workflows, and to this end we have added annotations in a number of places throughout the JAX API.
+The current state of type annotations in JAX is a bit patchwork, and efforts to add more have been hampered by more fundamental design questions.
+This doc attempts to summarize those issues and generate a roadmap for the goals and non-goals of type annotations in JAX.
+
+Why do we need such a roadmap? Better/more comprehensive type annotations are a frequent request from users, both internally and externally.
+In addition, we frequently receive pull requests from external users seeking to improve JAX's type annotations: it's not always clear to the JAX team member reviewing the code whether such contributions are beneficial, particularly when they introduce complex Protocols to address the challenges inherent to full-fledged annotation of JAX's use of Python.
```
8bitmp3, line 14: Suggestion:
- Split into shorter simple sentences with `. And`, as it may be hard to comprehend (tired users, users who aren't very proficient in English).
- Add PR examples.

For example:

 ```suggestion
In addition, we frequently receive pull requests from external users (for example, [PR #10322](https://github.com/google/jax/pull/10322), [PR #9917](https://github.com/google/jax/pull/9917)) seeking to improve JAX's type annotations. And it's not always clear to the JAX team member reviewing the code whether such contributions are beneficial, particularly when they introduce complex Protocols to address the challenges inherent to full-fledged annotation of JAX's use of Python.
```

File: docs/jep/11859-type-annotations.md [hunk 14]
```diff
@@ -0,0 +1,189 @@
...
+This document details JAX's goals and recommendations for type annotations within the package.
+
+## Why type annotations?
+
+There are a number of reasons that a Python project might wish to annotate their code-base; we'll summarize them in this document as Level 1, Level 2, and Level 3.
+
+### Level 1: annotations as documentation
```
8bitmp3, line 21: Here and across the doc. Forgot to mention this:

> "... capitalize only the first word in the title, the first word in a subheading after a colon, and any proper nouns or other terms that are always capitalized a certain way."

From https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings

```suggestion
### Level 1: Annotations as documentation
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+
+When originally introduced in [PEP 3107](https://peps.python.org/pep-3107/), type annotations were motivated partly by the ability to use them as concise, inline documentation of function parameter types and return types. JAX has long utilized annotations in this manner; an example is the common pattern of creating type names aliased to `Any`. An example can be found in `lax/slicing.py` [[source](http://google3/third_party/py/jax/_src/lax/slicing.py;l=48-54;rcl=446366018)]:
```
8bitmp3, line 23: @jakevdp This g3 link won't "work" unless this is intentional 🙂

8bitmp3: This link may work for _everyone_: https://github.com/google/jax/blob/main/jax/_src/lax/slicing.py 🙂

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+Array = Any
+Shape = core.Shape
+
+def slice(operand: Array, start_indices: Sequence[int],
+          limit_indices: Sequence[int],
+          strides: Optional[Sequence[int]] = None) -> Array:
+  ...
+```
+
+For the purposes of static type checking, this use of `Array = Any` for array annotations puts no constraint on the argument values (`Any` is equivalent to no annotation at all) but it does serve as a form of useful in-code documentation for the developer.
```
8bitmp3, line 35: Suggestion:
- Split a longer sentence into two shorter ones to reduce cognitive overhead.

```suggestion
For the purposes of static type checking, this use of `Array = Any` for array annotations puts no constraint on the argument values (`Any` is equivalent to no annotation at all). But it does serve as a form of useful in-code documentation for the developer.
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+
+For the sake of generated documentation, the name of the alias gets lost (the [HTML docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.slice.html) for `jax.lax.slice` report operand as type `Any`), so the documentation benefit does not go beyond the source code.
+
+A benefit of this level of type annotation is that it is never wrong to annotate a value with `Any`, so it will provide a concrete benefit to developers and users in the form of documentation, without added complexity of satisfying the stricter needs of any particular static type checker.
+
+### Level 2: annotations for intelligent autocomplete
```
8bitmp3, line 41: Suggestion: Add a 👾 

```suggestion
### Level 2: annotations for intelligent autocomplete

```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+
+This use of type checking requires going further than the simple aliases used above; for example, knowing that the `slice` function returns an alias of `Any` named `Array` does not add any useful information to the code completion engine. However, were we to annotate the function with a `DeviceArray` return type, the autocomplete would know how to populate the namespace of the result, and thus be able to suggest more relevant autocompletions during the course of development.
+
+JAX has begun to add this level of type annotation in a few places; one example is the `jnp.ndarray` return type within the `jax.random` package [[source](http://google3/third_party/py/jax/_src/random.py;l=359;rcl=445522278)]:
```
8bitmp3, line 46: 🤷  This link may not "work". Maybe you meant this https://github.com/google/jax/blob/main/jax/_src/random.py#L359

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+
+```python
+def shuffle(key: KeyArray, x: Array, axis: int = 0) -> jnp.ndarray:
+  ...
+```
+
+In this case `jnp.ndarray` is an abstract base class that forward-declares the attributes and methods of JAX arrays ([see source](http://google3/third_party/py/jax/_src/numpy/ndarray.py;l=40;rcl=431999528)), and so PyLance in VSCode can offer the full set of autocompletions on results from this function. Here is the result in VSCode:
```
8bitmp3, line 53: As before. Maybe this will work better https://github.com/google/jax/blob/main/jax/_src/numpy/ndarray.py#L41

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,189 @@
...
+With this framing (Level 1/2/3) and JAX-specific challenges in mind, we can begin to develop our roadmap for implementing consistent type annotations across the JAX project.
+For JAX type annotation, we will be guided by the following principles:
+
+### Goal of type annotations
+
+We would like to support full, *Level 1, 2, and 3* type annotation as far as possible. In particular, this means that we should have restrictive type annotations on both inputs and outputs to public API functions.
+
+### Avoid unstable typing mechanisms
+
+In order to not add undue development friction (due to the internal/external CI differences), we would like to be conservative in the type annotation constructs we use: in particular, when it comes to recently-introduced mechanisms such as `ParamSpec` (PEP [PEP 612](https://peps.python.org/pep-0612/),), we would like to wait until support in mypy and other tools stabilizes before relying on them.
```
8bitmp3, line 157: Suggestion:
- The link here may be rendering like `(PEP PEP 612),)`.

```suggestion
In order to not add undue development friction (due to the internal/external CI differences), we would like to be conservative in the type annotation constructs we use: in particular, when it comes to recently-introduced mechanisms such as `ParamSpec` ([PEP 612](https://peps.python.org/pep-0612/)), we would like to wait until support in mypy and other tools stabilizes before relying on them.
```

docs/jep/11859-type-annotations.md
```diff
@@ -18,9 +18,9 @@ This document details JAX's goals and recommendations for type annotations withi
...
 
-### Level 1: annotations as documentation
+### Level 1: Annotations as documentation
 
-When originally introduced in [PEP 3107](https://peps.python.org/pep-3107/), type annotations were motivated partly by the ability to use them as concise, inline documentation of function parameter types and return types. JAX has long utilized annotations in this manner; an example is the common pattern of creating type names aliased to `Any`. An example can be found in `lax/slicing.py` [[source](http://google3/third_party/py/jax/_src/lax/slicing.py;l=48-54;rcl=446366018)]:
+When originally introduced in [PEP 3107](https://peps.python.org/pep-3107/), type annotations were motivated partly by the ability to use them as concise, inline documentation of function parameter types and return types. JAX has long utilized annotations in this manner; an example is the common pattern of creating type names aliased to `Any`. An example can be found in `lax/slicing.py` [[source](hhttps://github.com/google/jax/blob/2bc3e39cd9104071ee39dacac22abd51b94eb27e/jax/_src/lax/slicing.py#L47-L58)]:
```
8bitmp3, line 23: Smol typo in `https` @jakevdp

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+Broadly speaking, annotating intent better serves Level 1 type checking, while annotating implementation better serves Level 3, while Level 2 is more of a mixed bag (both intent and implementation are important when it comes to annotations in IDEs).
+
+## JAX type annotation roadmap
+
+With this framing (Level 1/2/3) and JAX-specific challenges in mind, we can begin to develop our roadmap for implementing consistent type annotations across the JAX project.
+For JAX type annotation, we will be guided by the following principles:
+
+### Purpose of type annotations
+
+We would like to support full, *Level 1, 2, and 3* type annotation as far as possible. In particular, this means that we should have restrictive type annotations on both inputs and outputs to public API functions.
```
froystig, line 153: > we should have restrictive type annotations on both inputs and outputs

What does "restrictive" mean here? Does it conflict with the section header "inputs should be permissively typed" below?

> to public API functions

I take this to mean that we're not directly concerned with annotating private stuff, for its own sake, as far as this JEP is concerned. Is that right? If so, I like that and encourage it be highlighted more prominently!

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Annotate for intent
+
+JAX type annotations should in general indicate the **intent** of APIs, rather than the implementation, so that the annotations become useful to communicate the contract of the API. This means that at times inputs that are valid at runtime may not be recognized as valid by the static type checker (a simple example is an arbitrary iterator passed in place of a shape in some function implementations).
+
+### Inputs should be permissively-typed
```
froystig, line 159:
```suggestion
### Inputs should be permissively typed
```

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+Inputs to JAX functions and methods should be typed as permissively as is reasonable: for example, while shapes are typically tuples, functions that accept a shape should accept arbitrary sequences. Similarly, functions that accept a dtype need not require an instance of class `np.dtype`, but rather any dtype-convertible object. This might include strings, built-in scalar types, or dtype-adjacent classes such as `np.float64` and `jnp.float64`. In order to make this as uniform as possible across the package, we will add a {mod}`jax.typing` module with common type specifications, starting with broad categories such as:
+
+- `ArrayLike`
+- `DtypeLike`
+- `ShapeLike`
+- etc.
+  
+Note that these will in general be simpler than the equivalent protocols used in {mod}`numpy.typing`. For example, in the case of `DtypeLike`, JAX does not support structured dtypes, so JAX can use a simpler implementation. Similarly, in `ArrayLike`, JAX generally does not support list or tuple inputs in most places, so the type definition will be simpler than the NumPy analog.
+
+### Outputs should be strictly-typed
```
froystig, line 170:
```suggestion
### Outputs should be strictly typed
```

docs/jep/11859-type-annotations.md
froystig, line 153: > to support full, *Level 1, 2, and 3* type annotation as far as possible

From reading the rest of this note, and the suggestive ordering of those numbers, it seems like you have some degree of prioritization/preference in mind (e.g. 1 typically more important than 3). Is that the case?

froystig (approved):

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+
+Conversely, outputs of functions and methods should be typed as strictly as possible: for example, for a JAX function that returns an array, the output should be annotated with `jnp.ndarray` rather than `ArrayLike`. Functions returning a dtype should always be annotated `np.dtype`, and functions returning a shape should always be `Tuple[int]` or a strictly-typed NamedShape equivalent. For this purpose, we will implement in {mod}`jax.typing` several strictly-typed analogs of the permissive types mentioned above, namely:
```
NeilGirdhar, line 172: `Tuple[int, ...]`?

froystig:
Could we write down a criterion by which we'd say that work here is complete? Something like "all public API functions have annotations of level 1 at least"? (I just made that up.)

In other words, with this written, does it make sense to open a tracking issue? If so, will we know when to close it?

If the aim is instead to set up general guidelines/recommendations going forward, maybe clarify that in the note?

froystig:
Speaking of tracking issue, I think we want one for the JEP number (see [here](https://jax.readthedocs.io/en/latest/jep/index.html#how-to-start-a-jep)).

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+
+def slice(operand: Array, start_indices: Sequence[int],
+          limit_indices: Sequence[int],
+          strides: Optional[Sequence[int]] = None) -> Array:
+  ...
+```
+
+For the purposes of static type checking, this use of `Array = Any` for array annotations puts no constraint on the argument values (`Any` is equivalent to no annotation at all), but it does serve as a form of useful in-code documentation for the developer.
+
+For the sake of generated documentation, the name of the alias gets lost (the [HTML docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.slice.html) for `jax.lax.slice` report operand as type `Any`), so the documentation benefit does not go beyond the source code.
```
superbobry, line 37: FYI the autodoc extension has [some control](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases) over this.

Another trick we [use in Haiku](https://github.com/deepmind/dm-haiku/blob/78c2984969d5b360c19a12bb7eb609712b8312f8/docs/conf.py#L36) docs is adding `from __future__ import annotations` to each .py source when building docs. Not pretty, but gets the job done :)

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+These typically represent cases where typing problems have arisen; they may be inaccuracies in JAX type annotations, or inaccuracies in the static type checker's ability to correctly follow the control flow in the code.
+On occasion, they are due to real & subtle bugs in the behavior of pytype or mypy.
+In rare cases, they may be due to the fact that JAX uses Python patterns that are difficult or even impossible to express in terms of Python's static type annotation syntax.
+
+## Type annotation challenges for JAX
+
+JAX currently has type annotations that are a mixture of different styles, and aimed at all three levels of type annotation discussed above.
+Partly, this comes from the fact that JAX's source code poses a number of unique challenges for Python's type annotation system. We'll outline them here.
+
+### Challenge 1: pytype, mypy and developer friction
```
superbobry, line 84: I think the friction is actually a result of Python not having a precise description of what the type system is and how it is supposed to work. So, mypy, pytype, pyright, pyre, IntelliJ etc all implement a slightly different type system.

@kmillikin did a talk about this on the recent Typing Summit: https://youtu.be/BNTkWQfqP_c?t=8466

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+
+One challenge JAX currently faces is that package development must satisfy the constraints of two different static type checking systems, `pytype` (used by internal CI and internal Google projects) and `mypy` (used by external CI and external dependencies).
+Although the two type checkers have broad overlap in their behavior, each presents its own unique corner cases, as evidenced by the numerous `#type: ignore` and `#pytype: disable` statements throughout the JAX codebase.
+
+This creates friction in development: internal contributors may iterate until tests pass, only to find that on export their pytype-approved code falls afoul of mypy.
+For external contributors, it's often the opposite: a recent example is {jax-issue}`#9596` which had to be rolled-back after it failed internal Google continuous checks.
+Each time we move a type annotation from Level 1 (`Any` everywhere) to Level 2 or 3 (stricter annotations), it introduces more potential for such frustrating developer experiences.
+
+### Challenge 2: array duck-typing
```
superbobry, line 93: Do you think it might be possible to reduce the degree polymorphism at least in JAX internals? Public APIs could continue to use these highly polymorphic types, but the internals would only accept `DeviceArray`s.

docs/jep/11859-type-annotations.md
superbobry: > If so, I like that and encourage it be highlighted more prominently!

Ideally, with some rationale for why type annotations are less useful for JAX internals :)

docs/jep/11859-type-annotations.md
```diff
@@ -0,0 +1,192 @@
...
+
+The advantage of #1, annotating intent, is that the annotations are more useful to the user in communicating the API contract; while for the developer the flexibility leaves room for refactoring when necessary. The down-side (particularly for gradually-typed APIs like JAX's) is that it's quite likely that user code exists which runs correctly, but would be flagged as incorrect by a type checker.
+Gradual typing of an existing duck-typed API means that the current annotation is implicitly `Any`, so changing this to a stricter type may present to users as a breaking change.
+
+Broadly speaking, annotating intent better serves Level 1 type checking, while annotating implementation better serves Level 3, while Level 2 is more of a mixed bag (both intent and implementation are important when it comes to annotations in IDEs).
+
+## JAX type annotation roadmap
+
+With this framing (Level 1/2/3) and JAX-specific challenges in mind, we can begin to develop our roadmap for implementing consistent type annotations across the JAX project.
```
superbobry, line 148: Will the roadmap/further details be published as a separate JEP?

jakevdp:
Is it required to number based on a tracking issue? e.g. [JEP 9407](https://jax.readthedocs.io/en/latest/jep/9407-type-promotion.html) is #9407, which is a PR, not an issue.

docs/jep/11859-type-annotations.md
jakevdp: This is intended to be the roadmap.

docs/jep/11859-type-annotations.md
jakevdp: That's mostly the case already.

docs/jep/11859-type-annotations.md
jakevdp: It's not an order of preference, it's an order of how complicated/specific the types must be.

jakevdp: (my own opinion is that Level 3 is usually more trouble than its worth, especially for gradual typing of established packages where APIs have developed in a free-form duck-typing world typical of historical Python practices, but our loudest users seem pretty intent on Level 3 types wherever possible)

jakevdp: "restrictive" here means a type that would disallow some inputs. `x: Any` is not restrictive. `x: int` is. When I use "permissive" below, it's referring to the same dimension: we want inputs to be more restrictive than `Any` but still relatively permissive (e.g. `x: Union[int, float, Array]` instead of `x: Any` or `x:Array`) and we want outputs to be as restrictive as possible (e.g. `x: Array` rather than `x: Union[int, float, Array]`)

jakevdp: > I take this to mean that we're not directly concerned with annotating private stuff, for its own sake, as far as this JEP is concerned. Is that right? If so, I like that and encourage it be highlighted more prominently!

I think internal annotations are useful as well, and my intent was not to recommend against them in this JEP.

docs/jep/11859-type-annotations.md
jakevdp: Thanks for the link! That's a good point – but I think JAX faces unique friction here because we use two different type-checkers in our CI, and different populations of JAX developers have access to only one or the other while writing their code.

docs/jep/11859-type-annotations.md
jakevdp: Thanks! I added this to the discussion.

docs/jep/11859-type-annotations.md
froystig: Nice – that's clarifying. I wasn't sure what the level numbers were meant to convey exactly.

> my own opinion is that comprehensive Level 3 is usually more trouble than its worth [...] but [...] users seem pretty intent on Level 3 types wherever possible

Maybe state this? We want to have concrete choices to act on (as you say, this is a "roadmap"). I'd be supportive of something like "we should at least aim for 2, though note some want 3 and that can come next" – or whatever you'd actually propose instead. We're not going to have a fully annotated codebase tomorrow.

froystig:
> @froystig Is it required to number JEPs based on a tracking issue rather than a PR? It looks like [JEP 9407](https://jax.readthedocs.io/en/latest/jep/9407-type-promotion.html) is #9407, which is a PR, not an issue.

It was @hawkinsp's idea going forward, and I think it makes sense. For retroactive JEPs we're working with whatever we already had, which in many cases means PRs. It's not super critical since the two live in the same namespace. Are you OK using this PR effectively as a tracking issue?

jakevdp:
Thanks @froystig that makes sense. I'll plan to add a tracking issue and point this there.

docs/jep/12049-type-annotations.md
```diff
@@ -197,6 +197,40 @@ We will revisit this question in the future once support for such features stabi
...
+An advantage of this approach is that it is that the API is already familiar to many; most notably `numpy.typing` provides:
+
+- `NDArray`, which is a generic array type used only for type annotations
+- `np.ndarray`, which is the array type used for instance checks, and also works for type annotations. In Python 3.9, can use constructs of the form `np.ndarray[shape, dtype]`, but this is currently poorly documented and it's unclear how well supported that is at the moment.
+
+JAX currently implements {class}`jax.numpy.ndarray` for use with runtime `isinstance` checks. It uses a metaclass override to ensure that tracers also will pass `isinstance(tracer, jnp.ndarray)`, but it does not currently have any mechanism to ensure that tracers will be valid for `jnp.ndarray` annotations. If we were to follow numpy's approach to type annotation, we could do something like the following:
+- add `jax.typing.NDArray` for use with type annotations
+- add `TYPE_CHECKING` logic to ensure that `jnp.ndarray` can be used as an annotation for tracers; and possibly add class-level `__getitem__` to match numpy's shape and dtype specification.
```
patrick-kidger, line 212: Are these two bullet points meant as mutually exclusive possibilities? (Since having both sounds redundant.)

docs/jep/12049-type-annotations.md
```diff
@@ -197,6 +197,40 @@ We will revisit this question in the future once support for such features stabi
...
+
+Python itself is slowly moving to a world of unifying instance and annotation types; for example, with class-level `getitem` support in Python 3.9, classes like `typing.Dict`, `typing.List`, `typing.Tuple`, etc. are now [deprecated](https://peps.python.org/pep-0585/#implementation) in favor of their true-typed counterparts, `dict`, `list`, and `tuple`.
+
+With this in mind, JAX could instead choose to use its (eventual) `jax.Array` type directly for both annotation and instance checks. For handling tracers, tracers could be supported using a construct like the following:
+```python
+if TYPE_CHECKING:
+  Array = Union[jax.Array, jax.Tracer]
+else:
+  Array = jax.Array
+```
```
patrick-kidger, line 228: I think this should really be the definition of `jax.Array`, i.e. the above snippet is happening in (or imported into) `jax/__init__.py`. So I don't think `jax.Array` can appear on the right hand side. Do you mean something like:

```python
# somewhere.py
Array = ...

# jax/__init__.py
from somewhere import Array

if TYPE_CHECKING:
  Array = Union[Array, Tracer]

# jax/numpy/__init__.py
from somewhere import Array as ndarray
```

docs/jep/12049-type-annotations.md
jakevdp: They are meant to outline two aspects of following numpy's precedent.

docs/jep/12049-type-annotations.md
jakevdp: Sure, that would be an option. Let me update.

docs/jep/12049-type-annotations.md
```diff
@@ -219,33 +219,41 @@ Despite the familiarity of numpy's API choices, the `ndarray` / `NDArray` / `Arr
...
 Python itself is slowly moving to a world of unifying instance and annotation types; for example, with class-level `getitem` support in Python 3.9, classes like `typing.Dict`, `typing.List`, `typing.Tuple`, etc. are now [deprecated](https://peps.python.org/pep-0585/#implementation) in favor of their true-typed counterparts, `dict`, `list`, and `tuple`.
 
-With this in mind, JAX could instead choose to use its (eventual) `jax.Array` type directly for both annotation and instance checks. For handling tracers, tracers could be supported using a construct like the following:
+With this in mind, JAX could instead choose to use its (eventual) `jax.Array` type directly for both annotation and instance checks. For handling tracers within type annotations, we could use a construct like the following:
 ```python
 if TYPE_CHECKING:
   Array = Union[jax.Array, jax.Tracer]
 else:
-  Array = jax.Array
+  Array = jax._src.array.Array
```
patrick-kidger, line 227: Nit: the `TYPE_CHECKING` branch hasn't been updated.

docs/jep/12049-type-annotations.md
```diff
@@ -219,33 +219,41 @@ Despite the familiarity of numpy's API choices, the `ndarray` / `NDArray` / `Arr
...
 ```
-For instance checks, we could use the same metaclass override we currently do in the case of `np.ndarray`. And if we would like to support more granular shape/dtype-specific annotations in the future, we could do so via Python 3.9 class-level `__getitem__`, similar to how `list[int]` and `dict[str, int]` work for Python 3.9 built-in types.
+
+For handling instance checks, we could use the same metaclass override for `jax.Array` that we currently do in the case of `jnp.ndarray`. And if we would like to support more granular shape/dtype-specific annotations in the future, we could support constructions like `jax.Array[(3, 4), int]` via Python 3.9 class-level `__getitem__`, similar to how `list[int]` and `dict[str, int]` work for Python 3.9 built-in types.
```
patrick-kidger, line 230: FYI something like `jax.Array[(3, 4), int]` will make static type checkers sad.

The best syntax I've found so far, that won't throw an error when used with static type checking, is `int32[jax.Array, "3 4"]`.

docs/jep/12049-type-annotations.md
jakevdp: Thanks!

froystig (approved):
Nice improvements since my last read!

jakevdp:
Thanks - I'm squashing the commits and going to merge.

Thanks all for the discussion & comments on this! I think we have a nice framework to use going forward.

patrick-kidger (approved):
LGTM! I like where we ended up on this.

NeilGirdhar:
With the last Jax, release it looks like we're using `jax.Array` now!  I'm just curious, is there a plan to make `Array` generic the same way that numpy arrays are generic?  That is, support something like:
```python
from jax import Array
BooleanArray = Array[jnp.bool_]
IntegralArray = Array[jnp.integer[Any]]
RealArray = Array[jnp.floating[Any]]
ComplexArray = Array[Union[jnp.floating[Any], jnp.complexfloating[Any, Any]]]
```

jakevdp:
It would be nice, but it's not currently in scope: variadic type generics are still not fully supported by mypy and other type checkers. This is briefly addressed here: https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#avoid-unstable-typing-mechanisms

NeilGirdhar:
> variadic type generics

I agree with avoiding unstable typing mechanisms, but you don't need variadic type generics to have a generic dtype.  If I understand correctly, you only need them for the shape.

patrick-kidger:
Unfortunately, using variadic generics for shape annotations doesn't really work. See my previous discussions on the topic [here](https://github.com/google/jaxtyping/blob/main/FAQ.md#does-jaxtyping-use-pep-646-variadic-generics) and [here](https://github.com/patrick-kidger/torchtyping/issues/37#issuecomment-1153294196). (Although they both say a lot of the same things.)

TL;DR: with a lot of effort it might be possible to get something very limited. But that's such an unattractive value proposition that no-one has been interested in making that happen.

If you want dtype/shape annotations then for the forseeable future I think your best bet is [jaxtyping](https://github.com/google/jaxtyping).

NeilGirdhar:
> Unfortunately, using variadic generics for shape annotations doesn't really work. 

Right, but I'm not suggesting using shape annotations.  I'm only asking dtype annotations, which numpy already exposes and both Mypy and Pyright are able to check.  Dtype annotations have already saved me many times, so it would be nice to still have them as I switch over to Jax.

PR #11866, opened 2022-08-12
simplify slicing jaxprs a little

mattjj:
**Before:**
image

**After:**
image

No more `lt`, `add`, `select`, or `convert_element_type`!

This might actually improve compilation time a bit; on CPU, **the tiny benchmark added went from ~15.5ms to 13ms on my machine.**


sharadmv (approved):
Nice side quest!

PR #11867, opened 2022-08-12
generate lax.slice instead of lax.gather for x[<int>] or x[:<int>]

mattjj:
Before, on CPU:
image
image

After, on CPU:
image
image

Following up on #11866.

I don't know if these small compile time differences matter, but hey, roofshots! And the jaxpr pretty-printing win is enough for me.


jax/_src/numpy/lax_numpy.py
```diff
@@ -3594,6 +3595,23 @@ def _rewriting_take(arr, idx, indices_are_sorted=False, unique_indices=False,
...
+    if 0 <= idx < len(arr):
+      return lax.index_in_dim(arr, idx, keepdims=False)
+  if (isinstance(idx, slice) and
+      (type(idx.start) is int or idx.start is None) and
+      (type(idx.stop)  is int or idx.stop is  None) and
+      (type(idx.step)  is int or idx.step is  None)):
+    n = len(arr)
+    start = idx.start if idx.start is not None else 0
+    stop  = idx.stop  if idx.stop  is not None else n
+    step = idx.step if idx.step is not None else 1
```
froystig, line 3610:
```suggestion
    start = idx.start if idx.start is not None else 0
    stop  = idx.stop  if idx.stop  is not None else n
    step  = idx.step  if idx.step  is not None else 1
```

froystig (approved):
Nice work!

jax/_src/numpy/lax_numpy.py
```diff
@@ -82,6 +82,7 @@
 from jax.experimental.array import Array
 
 newaxis = None
+NoneType = type(None)
```
sharadmv, line 85: What's this for?

sharadmv (approved):

jax/_src/numpy/lax_numpy.py
mattjj: Oops, stale! (I was doing `isinstance(idx.step, NoneType)` etc., but changed when mypy complained.)

PR #11871, opened 2022-08-12
Fix dictionary key sorting in `tree_util`

XuehaiPan:
Function `sorted` does not support sorting with different types.

```python
>>> sorted({1: 0, 'a': 1})
TypeError: '<' not supported between instances of 'str' and 'int'
```

This will cause errors when manipulating dictionaries that have different key types.

```python
>>> import jax.tree_util as pytree
>>> pytree.tree_flatten({1: 1, 2: 2})
([1, 2], PyTreeDef({1: *, 2: *}))
>>> pytree.tree_flatten({'a': 1, 'b': 2})
([1, 2], PyTreeDef({'a': *, 'b': *}))
>>> pytree.tree_flatten({1: 1, 'a': 2})
  File ".../venv/lib/python3.9/site-packages/jax/_src/tree_util.py", line 60, in tree_flatten
    return pytree.flatten(tree, is_leaf)
TypeError: '<' not supported between instances of 'str' and 'int'
```

Since Python 3.6, the keys are insertion ordered.

NOTE: same change should be made for the jaxlib C extension (tensorflow/tensorflow#57122).

### DIFF
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -524,6 +524,22 @@ def _register_keypaths(
     )
 
 
+def _sorted_keys(dct: dict[Any, Any]) -> Iterable[Any]:
+  try:
+    # Sort directly if possible (do not use `key` for performance reasons)
+    return sorted(dct)
+  except TypeError:  # the keys are not comparable
+    try:
+      # Add `{obj.__class__.__module__}.{obj.__class__.__qualname__}` to the key
+      # order to make it sortable between different types (e.g. `int` vs. `str`)
+      return sorted(
+        dct,
+        key=lambda o: ('{0.__module__}.{0.__qualname__}'.format(o.__class__), o),
+      )
+    except TypeError:  # cannot sort the keys (e.g. user-defined types)
+      return dct  # fallback to insertion order
+
+
 _registry_with_keypaths = {}
 
 _register_keypaths(
@@ -532,7 +548,7 @@ _register_keypaths(
 _register_keypaths(
     list, lambda xs: tuple(SequenceKey(i) for i in range(len(xs)))
 )
-_register_keypaths(dict, lambda xs: tuple(DictKey(k) for k in sorted(xs)))
+_register_keypaths(dict, lambda xs: tuple(DictKey(k) for k in _sorted_keys(xs)))
 
 _register_keypaths(
     collections.defaultdict, lambda x: tuple(DictKey(k) for k in x.keys())


apaszke:
While this might solve your use case, I'm afraid that it can lead to surprising results. The order here is used to decide the PyTree structure. If you have two dicts that are not key-sortable and have different key orders, but are otherwise equivalent, they will lead to two different and incompatible pytrees.

XuehaiPan:
> If you have two dicts that are not key-sortable and have different key orders, but are otherwise equivalent, they will lead to two different and incompatible pytrees.

Thanks for pointing out this. Do you mean there are two types that are not comparable (not support `inst1 < inst2`), but the `__eq__` / `__hash__` methods are overloaded (`inst1 == inst2`)?

apaszke:
Yes, I agree that the tree construction is deterministic. But I would claim that dictionary key order has no influence on the meaning of a Python dict (use OrderedDict for that). As you've noticed, they will always compare equal, even though they use different orders. And I think that it is important that two equal objects produce equal pytrees, which is false in your example.

XuehaiPan:
> As you've noticed, they will always compare equal, even though they use different orders. And I think that it is important that two equal objects produce equal pytrees, which is false in your example.

Despite the non-sortable case, the old implementation also does not guarantee `tree1 == tree2` -> `pytree.tree_leaves(tree1) == pytree.tree_leaves(tree2)`.

Here I add a new `__lt__` method to the example in https://github.com/google/jax/pull/11871#issuecomment-1212997687, where `KeyType2`'s order is reversed:

```python
class KeyType1(KeyTypeBase):
    def __lt__(self, other: object) -> bool:
        if isinstance(other, KeyType1):
            return self.key < other.key  # <<<
        return False

class KeyType2(KeyTypeBase):
    def __lt__(self, other: object) -> bool:
        if isinstance(other, KeyType2):
            return self.key > other.key  # >>>
        return False
```

```python
In [2]: tree1 = {KeyType1(0): 0, KeyType1(1): 1}

In [3]: tree2 = {KeyType2(1): 1, KeyType2(0): 0}

In [4]: list(tree1)
Out[4]: [KeyType1(0), KeyType1(1)]

In [5]: list(tree2)
Out[5]: [KeyType2(1), KeyType2(0)]

In [6]: tree1 == tree2
Out[6]: True

In [7]: import jax.tree_util as pytree  # jaxlib 0.3.15

In [8]: pytree.tree_flatten(tree1)
Out[8]: ([0, 1], PyTreeDef({KeyType1(0): *, KeyType1(1): *}))

In [9]: pytree.tree_flatten(tree2)
Out[9]: ([1, 0], PyTreeDef({KeyType2(1): *, KeyType2(0): *}))

In [10]: pytree.tree_leaves(tree1) == pytree.tree_leaves(tree2)
Out[10]: False
```

apaszke:
Sure, nothing is guaranteed in Python. You can always come up with contraptions like this one. Our current implementation assumes that they keys have a total order, and your example violates that.

And I think it's not only about the implementation. pytree flattening is about converting a nested data structure into a sequence, it is ill-defined to attempt to do that unless the elements of any given container can be given a total order.

XuehaiPan:
> Our current implementation assumes that they keys have a total order

What if we sort by `key = (obj.__class__.__qualname__, obj)`? It will have a total order that `key[0]` is a string and always is comparable.

XuehaiPan:
Hi @apaszke, a gentle ping for this. The XLA upstream (https://github.com/tensorflow/tensorflow/pull/57122#issuecomment-1538488523) would decide to review the corresponding change if this PR is also accepted.

Is there any progress on this? For sortable keys, this PR is fully backward compatible. While for non-sortable keys, it sorts the keys by `key=lambda o: ('{0.__module__}.{0.__qualname__}'.format(o.__class__), o)` rather than raises a `TypeError`. If the JAX community thinks we need to raise an error when the keys are not directly comparable, I can close this PR.

XuehaiPan:
Closing for now due to inactivity.

Issue #11817, opened 2022-08-09
JAX implementation of scipy.stats.multinomial


harryjulian:
Please:

- [X] Check for duplicate requests.
- [X] Describe your goal, and if possible provide a code snippet with a motivating example.

Hi all,

Whilst working on a recent project I realised that scipy.stats.multinomial doesn't have a jax.scipy implementation. 

I thought it'd be worthwhile to implement as well as a decent first contribution.

I think I'm largely done on the work [here](https://github.com/harryjulian/jax/tree/scipy.stats.multinomial) but thought to make an issue before making a PR.

Is there any reason this hasn't been implemented thus far?

jakevdp:
I don't know of any reason it hasn't been implemented - feel free to open a PR if you'd like to contribute!

PR #11879, opened 2022-08-12
JAX implementation of scipy.stats.multinomial pmf & logpmf.

harryjulian:
Implementation of scipy.stats.multinomial, as referenced by issue [#11817](https://github.com/google/jax/issues/11817).


jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,28 @@
...
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.lax_numpy import _promote_args_inexact
+from jax._src.numpy.util import _wraps
+from jax._src.scipy.special import gammaln, xlogy
+
+def _check_args(x, n, p):
+    if jnp.sum(x) != n:
+        raise ValueError("The sum of x should equal n.")
+    if jnp.sum(p) != 1:
+        raise ValueError("The values of p must sum to 1.")
```
jakevdp, line 13: These kinds of checks won't be possible for jax implementations, because the arguments may be traced. The best solution is probably to remove `_check_args` entirely.

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,28 @@
...
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+    """JAX implementation of scipy.stats.multinomial.logpmf."""
```
jakevdp, line 17: Please use two-space indentation throughout.

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,28 @@
...
+    x, n, p = _promote_args_inexact("multinomial.logpmf", x, n, p)
```
jakevdp, line 18: Since `n` is an integer, should we avoid promoting it to float?

jakevdp (changes requested):
This is a great start! The big remaining TODO is to add tests; you can see some existing examples here: https://github.com/google/jax/blob/d20dcf4b5040e05e8199ec144bae3472ebe2c24e/tests/scipy_stats_test.py#L51-L67

jax/_src/scipy/stats/multinomial.py
harryjulian: This makes sense. Have you any intuition on the best way in which I could still implement the checks, or should I just remove the entirely? Perhaps returning `jnp.inf` if the checks aren't satisfied. My initial thought would be to use `jnp.where` but the docs indicate that it can't currently be JIT compiled.

jakevdp: Returning `NaN` for invalid values is probably the best bet. The usual way to do that is using the three-argument form of `jnp.where`, which is compatible with transformations like JIT

harryjulian:
Added some changes, let me know if anything needs altering! :)

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
       lsp_stats.norm.cdf(np.full((4,), np.inf, np.float32)),
       check_dtypes=False)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape], [dtype]),
```
jakevdp, line 568: `jtu.format_test_name_suffix` is usually used when you have a name you need to format, it's strange to see it with an empty string.

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
...
+       "shape": shape}
+      for shape in [(2,), (3,), (4,), (5,), (6,), (7,), (1,2), (1, 3), (1,4), (1,5)]
+      for dtype in itertools.combinations_with_replacement(jtu.dtypes.floating, 2)))
+  def testMultinomialLogPmf(self, shape, dtype):
+    rng = jtu.rand_positive(self.rng())
+    scipy_fun = osp_stats.multinomial.logpmf
+    lax_fun = lsp_stats.multinomial.logpmf
+
+    def args_maker():
+      x = np.rint(rng(shape, dtype[0])).astype(dtype[0])
```
jakevdp, line 579: If inputs are expected to be integers, would it make more sense to generate integers directly?

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
...
+      n = np.sum(x)
+      p = rng(shape, dtype[1])
+      # Normalize the array such that it sums it's entries sum to 1 (or close enough to)
+      p = (p + np.min(p)) / np.sum(p + np.min(p))
```
jakevdp, line 583: Can you comment on why you add `np.min(p)` to the normalization? What's the problem with normalizing `p` directly?

wonhyeongseo:
Hello @harryjulian, 
may you please review the PR I've made on your branch? 
It will hopefully resolve some reviews from @jakevdp . 
I'm new to Github, so please don't hesitate to point out best practices I might be missing. 

Thanks and have a great day! 😄

tests/scipy_stats_test.py
wonhyeongseo: np.rint method does the same rounding as np.around or np.fix. We do not change to an integer because of the granularity of delta. Thank you!

harryjulian:
Hey @wonhyeongseo you need to sign the CLA :)

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,34 @@
...
+from jax import lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.lax_numpy import _promote_args_inexact
+from jax._src.numpy.util import _wraps
+from jax._src.scipy.special import gammaln, xlogy
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.logpmf."""
+  x, p = _promote_args_inexact("multinomial.logpmf", x, p)
```
jakevdp, line 26: I'm somewhat concerned with this implementation allowing floating point inputs for `x`. The function is not defined for non-integer `x` values. In numpy, this is enforced by a runtime check that effectively checks `all(x.astype(int) == x)`, but of course we can't do this in JAX because at runtime we may be on GPU or TPU where this kind of error propagation is difficult.

The path chosen here is to silently proceed with the computation even if `x` is invalid. But we have another option: if x has integer dtype, then we *know* that it's valid. So a different option would be to require an integer dtype for `x`. What do you think?

wonhyeongseo: Hello sir, I am a novice to programming and I tried to cast x to an integer in this PR for @harryjulian 's branch. May you please review it? Thank you sir!

jakevdp: If you're going to be making significant changes here and @harryjulian is no longer working on this branch, it would probably be cleaner to open a new pull request to avoid all this kind of indirection during review.

PR #11887, opened 2022-08-12
skip redundant `asarray` conversion in LAX numpy rewriting-take helper

froystig:
This isn't needed, and its removal serves to help support indexing on generally-eltyped shaped arrays (cc @mattjj).


jakevdp (approved):
Looks good! Given the two call sites, I think this change is pretty safe.

PR #11928, opened 2022-08-16
Expose pure callback and enable vectorized callbacks

sharadmv:


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 
 ## jax 0.3.17 (Unreleased)
 * [GitHub commits](https://github.com/google/jax/compare/jax-v0.3.16...main).
+* Changes
+  * Added {func}`jax.pure_callback` that enables calling back to pure Python functions from functions compiled with `jax.jit` or `jax.pmap`.
```
mattjj, line 14:
```suggestion
  * Added {func}`jax.pure_callback` that enables calling back to pure Python functions from compiled functions (e.g. functions decorated with `jax.jit` or `jax.pmap`).
```

Does it work with pjit/xmap/control flow too? I suggested a tweaked wording assuming the answer is 'yes'.

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
       return x
   return jax.tree_util.tree_map(try_to_block, x)
 
+def pure_callback(callback: Callable[..., Any], result_shape_dtypes: Any,
+                  *args: Any, **kwargs: Any):
+  """Calls a pure Python callback function from staged out JAX programs.
```
mattjj, line 3251:
```suggestion
  """Applies a functionally pure Python callable. Works under `jit`/`pmap`/etc.
```

Not sure if the suggested phrasing is ideal but I'm thinking:
1. "pure Python" isn't as clear as saying "functionally pure";
2. "staged out", while I love it, may not be as clear as saying it works with `jit`, etc", though if we had space we might want to say both;
3. I like "applies" better than "calls", esp when pure;
4. "from staged out JAX programs" could suggest it _only_ works when staged out.

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+
+  ``pure_callback`` enables calling a Python function in JIT-ed JAX functions.
+  The input ``callback`` will be passed NumPy arrays in place of JAX arrays and
+  should also return NumPy arrays. Execution takes place on the CPU host.
```
mattjj, line 3255:
```suggestion
  should also return NumPy arrays. Execution takes place on CPU, like any Python+NumPy function.
```

Also, what happens if we use jnp? Undefined behavior? Possible hangs?

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+
+  The callback is treated as "pure" meaning it can be called multiple times when
+  transformed (for example in a ``vmap`` or ``pmap``), and it can also
+  potentially be removed from JAX programs via dead-code elimination. Pure
+  callbacks can also be reordered if data-dependence allows.
```
mattjj, line 3260:
```suggestion
  The callback is treated as functionally pure, meaning it has no side-effects and its output value depends only on its argument values. As a consequence, it is safe to be called multiple times (e.g. when transformed by ``vmap`` or ``pmap``), or not to be called at all when e.g. the output of a `jit`-decorated function has no data dependence on its value. Pure callbacks may also be reordered if data-dependence allows.
```

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+
+  When ``pmap``-ed, the pure callback will be called several times (one on each axis
+  of the map). When `vmap`-ed the behavior will depend on the value of the
+  ``rank_polymorphic`` keyword argument. If the callback is indicated as rank
+  polymorphic, the callback will be called directly on batched inputs (where the
```
mattjj, line 3265: There could be more than one batch axis (with multiple applications of `vmap`), but the text here talks about _the_ batch axis. Maybe tweak to highlight that there could be arbitrarily many batch axes?

I wonder if we can make the name of the option more descriptive, like instead of `rank_polymorphic` we can call it `arbitrary_leading_batch_axes` or something. Ideas?

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+  batch axis is the leading dimension). Additionally, the callbacks should
+  return outputs that also have a leading batch axis. If not rank polymorphic,
+``callback`` will be mapped sequentially across the batched axis.
+
+  Args:
+    callback: A Python callable. The callable will be passed in NumPy arrays and
```
mattjj, line 3271:
```suggestion
    callback: A Python callable. The callable will be passed PyTrees of NumPy arrays as arguments, and
```

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+      should return a PyTree of NumPy arrays that matches
+      ``result_shape_dtypes``.
+    result_shape_dtypes: A PyTree of Python objects that have ``shape`` and
+      ``dtype`` properties that correspond to the shape and dtypes of the
+      outputs of ``callback``.
```
mattjj, line 3276:
```suggestion
    result_shape_dtypes: A PyTree with leaves that are objects with ``shape`` and
      ``dtype`` attributes which represent to the shapes and dtypes of the
      value of ``callback`` applied to ``args`` and ``kwargs``.
```

jax/_src/api.py
```diff
@@ -3245,6 +3246,50 @@ def try_to_block(x):
...
+    *args: The positional arguments to the callback. Must be PyTrees of JAX
+      types.
+    rank_polymorphic: A boolean that indicates whether or not ``callback`` is
```
mattjj, line 3279: See above suggestion about renaming `rank_polymorphic` (and avoiding that terminology).

mattjj (approved):
LGTM modulo some docs suggestions we could iterate on!

jax/_src/api.py
sharadmv: Yea I think rank-polymorphic is not the best term here. Another option is maybe "vectorized" (borrowing from `np.vectorize`)?

sharadmv (commented):
Thanks for the detailed feedback

CHANGELOG.md
sharadmv: It doesn't work with `pjit` yet.

PR #11947, opened 2022-08-16
ROCm dlpack support

rahulbatra85:
Addresses test failure related to DLPack support in array_interoperability_tests.py

### DIFF
--- a/jax/_src/dlpack.py
+++ b/jax/_src/dlpack.py
@@ -64,6 +64,14 @@ def from_dlpack(dlpack):
     gpu_backend = xla_bridge.get_backend("cuda")
   except RuntimeError:
     gpu_backend = None
+
+  # Try ROCm if CUDA backend not found
+  if gpu_backend is None: 
+    try:
+      gpu_backend = xla_bridge.get_backend("rocm")
+    except RuntimeError:
+      gpu_backend = None
+
   buf = xla_client._xla.dlpack_managed_tensor_to_buffer(
       dlpack, cpu_backend, gpu_backend)
 
--- a/tests/array_interoperability_test.py
+++ b/tests/array_interoperability_test.py
@@ -73,7 +73,6 @@ class DLPackTest(jtu.JaxTestCase):
     take_ownership=[False, True],
     gpu=[False, True],
   )
-  @jtu.skip_on_devices("rocm") # TODO(sharadmv,phawkins): see GH issue #10973
   def testJaxRoundTrip(self, shape, dtype, take_ownership, gpu):
     rng = jtu.rand_default(self.rng())
     np = rng(shape, dtype)
@@ -96,7 +95,6 @@ class DLPackTest(jtu.JaxTestCase):
     dtype=dlpack_dtypes,
   )
   @unittest.skipIf(not tf, "Test requires TensorFlow")
-  @jtu.skip_on_devices("rocm") # TODO(sharadmv,phawkins): see GH issue #10973
   def testTensorFlowToJax(self, shape, dtype):
     if not config.x64_enabled and dtype in [jnp.int64, jnp.uint64, jnp.float64]:
       raise self.skipTest("x64 types are disabled by jax_enable_x64")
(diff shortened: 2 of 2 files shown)


tests/array_interoperability_test.py
```diff
@@ -215,14 +212,13 @@ def testNumpyToJax(self, shape, dtype):
      for shape in all_shapes
      for dtype in torch_dtypes))
   @unittest.skipIf(numpy_version < (1, 22, 0), "Requires numpy 1.22 or newer")
-  @jtu.skip_on_devices("gpu")
```
hawkinsp, line 218: Revert this change.

(NumPy simply doesn't accept GPU dlpacks from any vendor... maybe let's add a comment to that effect?)

jax/_src/dlpack.py
```diff
@@ -57,6 +57,14 @@ def from_dlpack(dlpack):
     gpu_backend = xla_bridge.get_backend("cuda")
   except RuntimeError:
     gpu_backend = None
+
+  #Try ROCm if CUDA backend not found
```
hawkinsp, line 61: Nit: Space before "Try"

hawkinsp (commented):
Is this change sufficient by itself to get things working? Don't we also need a TF change?

If nothing else things like this must have the wrong vendor ID with this code:
https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/python/dlpack.cc;drc=6f62beb16126e0840dac5b405ab6284d65b716a2;l=227

jax/_src/dlpack.py
rahulbatra85: Fixed

tests/array_interoperability_test.py
rahulbatra85: Added a comment

rahulbatra85: Also regarding a corresponding TF change, I did look at the code that you pointed above, but the testcases would pass with changes just in JAX dlpack.py.

I will investigate further and see if testcases are missing some coverage here

hawkinsp: One thing to note is that the tests will pass if you have only JAX installed, because JAX will agree with itself on what the vendor ID is. Try installing TF and PyTorch with ROCM enabled and run the tests. I assume one or more of those have ROCM support and it is the interoperability between them we must test.

tests/array_interoperability_test.py
```diff
@@ -215,7 +212,7 @@ def testNumpyToJax(self, shape, dtype):
...
+  @jtu.skip_on_devices("gpu") # Numpy doesn't support DLpacks from GPU vendors
```
hawkinsp, line 215: A better way to phrase this would be:
"NumPy only accepts cpu DLPacks".

tests/array_interoperability_test.py
rahulbatra85: Made some changes on tensorflow to address the device ID for ROCm.
Here is the PR https://github.com/tensorflow/tensorflow/pull/57640

tests/array_interoperability_test.py
rahulbatra85: Done!

rahulbatra85:
FYI: Tensorflow PR that enables DLPack for ROCm was merged this morning
https://github.com/tensorflow/tensorflow/pull/57640

rahulbatra85:
@hawkinsp  Ping!

rahulbatra85:
@hawkinsp  Any feedback on this? Can we merge it?

Also, please note ROCm support for DLPack is in Tensorflow Eager and XLA, so that pre-req is met.

jax/_src/dlpack.py
```diff
@@ -64,6 +64,14 @@ def from_dlpack(dlpack):
     gpu_backend = xla_bridge.get_backend("cuda")
   except RuntimeError:
     gpu_backend = None
+
```
hawkinsp, line 67: This code will probably need to change in the future. A better design would be to supply both backends to the Python code and it can use whichever it needs. Fine for now.

hawkinsp (approved):

jax/_src/dlpack.py
rahulbatra85: @hawkinsp ok, thanks for merging!

PR #11952, opened 2022-08-16
introduce key-element-type arrays and overhaul the Python PRNG key array

froystig:
Big step for #9263.

Before this change, the Python PRNG key array was a pytree type wrapping a `uint32` array. This was a stopgap that misbehaved under `vmap`, `scan`, and even `jax.tree_map`. For a while, we thought we might rely on something like the typeclass mechanisms in development (e.g. `vmappable`) to move away from a pytree.

We're now taking a different approach: introducing key element types into our IR and other internal machinery. During staging, we map user-facing PRNG key arrays to abstract arrays such element type. Check it out:

```python
>>> import jax
>>> def f(seed):
...   key = jax.random.PRNGKey(seed)
...   key, _ = jax.random.split(key, 2)
...   return jax.random.uniform(key, (3, 7))
... 
>>> print(jax.make_jaxpr(f)(54))
{ lambda ; a:i32[]. let
    b:key<fry>[] = random_seed[impl=fry] a
    c:key<fry>[2] = random_split[count=2] b
    d:key<fry>[1] = slice[limit_indices=(1,) start_indices=(0,) strides=(1,)] c
    e:key<fry>[] = squeeze[dimensions=(0,)] d
    f:key<fry>[1] = slice[limit_indices=(2,) start_indices=(1,) strides=(1,)] c
    _:key<fry>[] = squeeze[dimensions=(0,)] f
    g:f32[1,1] = broadcast_in_dim[broadcast_dimensions=() shape=(1, 1)] 0.0
    h:f32[1,1] = broadcast_in_dim[broadcast_dimensions=() shape=(1, 1)] 1.0
    i:u32[3,7] = random_bits[bit_width=32 shape=(3, 7)] e
    j:u32[3,7] = shift_right_logical i 9
    k:u32[3,7] = or j 1065353216
    l:f32[3,7] = bitcast_convert_type[new_dtype=float32] k
    m:f32[3,7] = sub l 1.0
    n:f32[1,1] = sub h g
    o:f32[3,7] = mul m n
    p:f32[3,7] = add o g
    q:f32[3,7] = max g p
  in (q,) }
```

This leans heavily on our recently-introduced extended element type capabilities from #11768 (and #11824 and #11891).

As a consequence, `vmap`, `scan`, etc. now work. Another consequence—because the logic for this was needed either way—is that now seeding, random bit generation, and `jax.random.{split,fold_in}` are all batch-polymorphic (and `fold_in` in particular does argument broadcasting):

```python
>>> import jax.numpy as jnp
>>> def f(seed):
...   keys = jax.random.split(jax.random.PRNGKey(seed), 4)
...   msgs = jnp.arange(4, dtype=jnp.uint32)
...   a = jax.random.fold_in(keys, msgs)
...   b = jax.random.fold_in(keys[None, :], msgs[:, None])
...   c = jax.random.fold_in(keys[:, None], msgs[None, :])
...   d = jax.random.split(keys)
...   return a, b, c, d
...
>>> print(jax.make_jaxpr(f)(54))
{ lambda ; a:i32[]. let
    b:key<fry>[] = random_seed[impl=fry] a
    c:key<fry>[4] = random_split[count=4] b
    d:u32[4] = iota[dimension=0 dtype=uint32 shape=(4,)] 
    e:key<fry>[4] = random_fold_in c d
    f:key<fry>[1,4] = broadcast_in_dim[broadcast_dimensions=(1,) shape=(1, 4)] c
    g:u32[4,1] = broadcast_in_dim[broadcast_dimensions=(0,) shape=(4, 1)] d
    h:key<fry>[4,4] = random_fold_in f g
    i:key<fry>[4,1] = broadcast_in_dim[broadcast_dimensions=(0,) shape=(4, 1)] c
    j:u32[1,4] = broadcast_in_dim[broadcast_dimensions=(1,) shape=(1, 4)] d
    k:key<fry>[4,4] = random_fold_in i j
    l:key<fry>[4,2] = random_split[count=2] c
  in (e, h, k, l) }
```

We should do more argument checking as a follow-up if we want to allow this kind of usage.

A sample of changes made to introduce key-element-type arrays:

* Introduce a new element type (`prng.KeyTy`), with the requisite IR type mapping and device result handlers, as well as lowering rules for dtype-polymorphic primitive operations.
* Introduce primitives for basic RNG operations: `random_seed`, `random_bits`, `random_split`, `random_fold_in`. These primitives essentially delegate to the underlying PRNG implementation (directly so in their impl rules, and by translating their staged-out form in lowering rules).
* Also introduce `random_wrap` and `random_unwrap` for "unsafe" conversion from/to the base `uint32` array. We need this backwards compatibility, and it's useful for tests.
* Introduce some `vmap`-based helpers to adapt PRNG impls (which define basic `random_bits`, `split`, etc. on scalars) to the above batch-polymorphic primitives. Most of the primitives are vectorized, but `random_fold_in` is a broadcasting binary op.
* Update the `gamma` primitive rules to account for key-element-type abstract arrays (nice simplification here).
* Give PRNG implementation short string names ("tags") for IR pretty-printing.
* Update `lax.stop_gradient` to handle opaque dtypes (currently anything from the custom/extended set).
* Fix up loop MLIR lowering, which assumed that shaped arrays of all dtypes have the same physical shape.
* Add new tests (exercising staging, jaxprs, lowerings, ...)

A sample of changes made to rework Python-level PRNG key arrays:

* Mimic `isinstance(x, KeyArray)` checks on abstract key arrays and tracers that carry them.
* Patch (only a subset of) standard device array attributes onto PRNG key arrays.
* Implement various conversion handlers (sharding, constant-creation, `device_put`).
* Accept PRNG key arrays as input to `lax_numpy.transpose`.
* Update tests and rename some internals.

A sample of extra changes along the way:

* Disallow AD on key-typed arrays in the main API.
* Hoist `random_bits`'s named-shape-handling logic, which used to only take place in the threefry PRNG's `random_bits` implementation, up to the new `random_bits` traceable, so that we apply it consistently across PRNG implementations.

By overhauling random stuff onto the new element type extension machinery, we're also now indirectly testing the latter more heavily. Parts of this change are really about that—further general development on internal handling of element types. Most notably, I had to teach `jax2tf` about such extensions. This involved:

* Introducing a `physical_avals` view as a method on an element type-definition. This is analogous to the existing mapping to MHLO types (`aval_to_ir_types`), but where the output is an aval with a grounded element type (and hence a direct correspondence to TF and to lowerings).

* Teaching jax2tf to understand the logical vs physical aval distinction, and to trace with logical avals but work with physical ones as needed. This required fixing up various `tf_impl` rules and changing various helpers.

* Writing `tf_impl` rules for `random_{seed,split,fold_in,bits}` primitives, and teaching the jax2tf test harness to unwrap key-array-typed outputs into their physical `uint32` form when doing output comparison tests.

Finally, this change leaves some unwanted `lax` and `jax.numpy` operations superficially available on key arrays during tracing/staging (e.g. under `jit`), though not outside of it. We ultimately want to disallow these and raise useful errors, and I'm leaving that for follow-up work. For now, applying such operations under `jit` may result in downstream errors in the middle-end instead.

Everything here is still guarded by `config.jax_enable_custom_prng`, whose default setting hasn't changed (it is off).

### DIFF
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -160,7 +160,15 @@ def count_device_put():
 
   def device_put_and_count(*args, **kwargs):
     count[0] += 1
-    return device_put(*args, **kwargs)
+    # device_put handlers might call `dispatch.device_put` (e.g. on an
+    # underlying payload or several). We only want to count these
+    # recursive puts once, so we skip counting more than the outermost
+    # one in such a call stack.
+    dispatch.device_put = device_put
+    try:
+      return device_put(*args, **kwargs)
+    finally:
+      dispatch.device_put = device_put_and_count
 
   dispatch.device_put = device_put_and_count
   try:
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2963,7 +2963,7 @@ class ShapeDtypeStruct:
   __slots__ = ["shape", "dtype", "named_shape"]
   def __init__(self, shape, dtype, named_shape=None):
     self.shape = shape
-    self.dtype = np.dtype(dtype)
+    self.dtype = dtype if core.is_custom_eltype(dtype) else np.dtype(dtype)
     self.named_shape = {} if named_shape is None else dict(named_shape)
 
   size = property(lambda self: prod(self.shape))
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -1104,6 +1104,9 @@ def _check_scalar(x):
 def _check_input_dtype_revderiv(name, holomorphic, allow_int, x):
   _check_arg(x)
   aval = core.get_aval(x)
+  if core.aval_has_custom_eltype(aval):
+    raise TypeError(
+        f"{name} with input element type {core.aval_eltype(aval).name}")
   if holomorphic:
     if not dtypes.issubdtype(aval.dtype, np.complexfloating):
       raise TypeError(f"{name} with holomorphic=True requires inputs with complex dtype, "
@@ -1122,6 +1125,9 @@ _check_input_dtype_grad = partial(_check_input_dtype_revderiv, "grad")
 
 def _check_output_dtype_revderiv(name, holomorphic, x):
   aval = core.get_aval(x)
+  if core.aval_has_custom_eltype(aval):
+    raise TypeError(
+        f"{name} with output element type {core.aval_eltype(aval).name}")
   if holomorphic:
     if not dtypes.issubdtype(aval.dtype, np.complexfloating):
       raise TypeError(f"{name} with holomorphic=True requires outputs with complex dtype, "
@@ -1199,6 +1205,9 @@ def jacfwd(fun: Callable, argnums: Union[int, Sequence[int]] = 0,
 def _check_input_dtype_jacfwd(holomorphic: bool, x: Any) -> None:
   _check_arg(x)
   aval = core.get_aval(x)
+  if core.aval_has_custom_eltype(aval):
+    raise TypeError(
+        f"jacfwd with input element type {core.aval_eltype(aval).name}")
   if holomorphic:
     if not dtypes.issubdtype(aval.dtype, np.complexfloating):
       raise TypeError("jacfwd with holomorphic=True requires inputs with complex "
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -108,8 +108,13 @@ def apply_primitive(prim, *args, **params):
                                         **params)
   return compiled_fun(*args)
 
-# TODO(phawkins): update code referring to xla.apply_primitive to point here.
+# TODO(phawkins,frostig,mattjj): update code referring to
+# xla.apply_primitive to point here, or use simple_impl if that's why
+# it is using apply_primitive to begin with
 xla.apply_primitive = apply_primitive
+
+def simple_impl(prim):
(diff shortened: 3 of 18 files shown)


froystig:
@gnecula, note the jax2tf changes

jax/_src/test_util.py
```diff
@@ -808,23 +816,25 @@ def assertMultiLineStrippedEqual(self, expected, what):
                               msg=f"Found\n{what}\nExpecting\n{expected}")
 
   def _CompileAndCheck(self, fun, args_maker, *, check_dtypes=True,
-                       rtol=None, atol=None, check_cache_misses=True):
+                       rtol=None, atol=None, check_cache_misses=True,
+                       post_process=None):
```
mattjj, line 820: IIUC this new argument is only used in one (parameterized) test. It seems a little heavyweight to change this low-level method, depended on by all tests, just to make one new parameterized test work. Plus, adding behavior via kwargs is a smell, as you like to say!

How about we just make that one test in random_test.py call its own version of _CompileAndCheck, rather than editing this one?

mattjj (approved):

jax/_src/test_util.py
froystig: Thanks calling this out. You knew I'd agree! I fixed this in a way that required neither the kwarg nor a one-off `_CompileAndCheck` variant, thankfully.

PR #11961, opened 2022-08-17
Use the pjrt plugin device client.

jakeh-gc:
This pull request allows jax to make use of the proposed plugin device client interface in https://github.com/tensorflow/tensorflow/pull/57193.

This is useful for targeting experimental/out-of-tree backends.

This is disabled by default.


skye (approved):

jakeh-gc:
Now that JAX has updated the XLA version, this should build cleanly and pass the tests.

jakeh-gc:
> The failure is unrelated. I'm working on getting this PR in.

There was previously a failure because it depended on the Bazel rule I added to TF.
That's now present, so should be all clear on my end.

jakeh-gc:
Thanks 👍

PR #11963, opened 2022-08-17
Include ABI tag in jaxlib wheels.

hawkinsp:
Currently JAX wheels end up with names like:
jaxlib-0.3.15-cp39-none-manylinux2014_x86_64.whl

This PR changes the wheel names to:
jaxlib-0.3.15-cp39-cp39-manylinux2014_x86_64.whl

i.e., we include the CPython ABI tag. This simply reflects the status
quo in the wheel name, and does not change what jaxlib needs.

### DIFF
--- a/jaxlib/setup.py
+++ b/jaxlib/setup.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from setuptools import setup
+from setuptools.dist import Distribution
 import os
 
 __version__ = None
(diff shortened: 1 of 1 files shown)


yashk2810:
What is the gpu wheel name change?

hawkinsp:
> What is the gpu wheel name change?

The same. The GPU wheels differ only from the CPU wheels in their version number. This change is orthogonal.

yashk2810 (approved):

PR #11968, opened 2022-08-17
[x64] clean up internal uses of result_type()

jakevdp:
Why? For X64 deprecation (#8178) I am working on some guardrails around dtype canonicalization, so we want to avoid extra calls to `canonicalize_dtype` that will generate noise.

This change is intended to be a pure refactoring, with no user-visible impact. It relies on two facts:

- `dtypes.result_type` canonicalizes the output, so there is no need to call `canonicalize_dtype` again
- `dtypes.result_type(*args, return_weak_type_flag)` is essentially equivalent to `dtypes._lattice_result_type(*args)` followed by `dtypes.canonicalize_dtype` on the output, so we can replace this common pattern with a single `dtypes.result_type` call.

### DIFF
--- a/jax/_src/abstract_arrays.py
+++ b/jax/_src/abstract_arrays.py
@@ -33,12 +33,10 @@ raise_to_shaped = core.raise_to_shaped
 
 
 def make_shaped_array(x):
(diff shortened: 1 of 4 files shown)


jakevdp:
Closing, as this has gone stale.

Issue #6265, opened 2021-03-29
Customizable reduction in jax.lax.scatter


kho:
XLA's Scatter op allows a customizable reduction via a custom `update_computation`. Currently, different scatter reductions are implemented as separate primitives in lax (`lax.scatter`, and `lax.scatter_{add,mul,min,max}`). It would be useful for there to be a version of scatter that can be used with a custom reduction, as long as the reduction is a commutative monoid (e.g. `logsumexp`). A concrete application that I have is to implement `segment_logsumexp`.

It is not too hard to create a `scatter` jaxpr with a custom `update_jaxpr` field, e.g. the following seems to work for `logsumexp`:
```python
import jax
import jax.numpy as jnp
from jax._src.lax.lax import _abstractify, _const, _reduction_jaxpr, scatter_p


logsumexp_0 = float('-inf')


def logsumexp(a, b):
  c = jnp.maximum(a, b)
  return jnp.where(
      jnp.logical_or(a == logsumexp_0, b == logsumexp_0), c,
      c + jnp.log(jnp.exp(a - c) + jnp.exp(b - c)))

def make_scatter(reduction):

  def scatter(operand,
              scatter_indices,
              updates,
              dimension_numbers,
              *,
              indices_are_sorted=False,
              unique_indices=False):
    jaxpr, consts = _reduction_jaxpr(
        reduction,
        # This doesn't have to the monoid identity. It's simply used to trace the call to `reduction`.
        _abstractify(_const(operand, 0)))
    return scatter_p.bind(
        operand,
        scatter_indices,
        updates,
        update_jaxpr=jaxpr,
        update_consts=consts,
        dimension_numbers=dimension_numbers,
        indices_are_sorted=indices_are_sorted,
        unique_indices=unique_indices)

  return scatter

operand = jnp.ones([3, 3])
scatter_indices = jnp.array([[1, 1], [2, 2]])
updates = jnp.array([1., 2.])
dimension_numbers = jax.lax.ScatterDimensionNumbers(
    update_window_dims=(),
    inserted_window_dims=(0, 1),
    scatter_dims_to_operand_dims=(0, 1))
make_scatter(logsumexp)(operand, scatter_indices, updates, dimension_numbers)
# Output:
# DeviceArray([[1.       , 1.       , 1.       ],
#             [1.       , 1.6931472, 1.       ],
#             [1.       , 1.       , 2.3132615]], dtype=float32)
```

There are however a few remaining issues:

[...]

PR #12004, opened 2022-08-19
Add a new scatter_reduce primitive and implement segment_reduce for a custom reduction

kho:
I took a stab at implementing #6265 since I actually need `segment_logsumexp` now (although I created `segment_reduce` instead so that I can use different reductions).

Currently the jvp rule hasn't been implemented, it's fine for my use cases to leave it unimplemented for now, but I also want to run my idea about jvp by you guys:

I think an efficient general jvp implementation will require knowing the following function `f(x, z)`: let `z = computation(x, y)` for some `y`, `f(x, z)` is the derivative of `z` w.r.t. `x` computed knowing `x` and `z` but not `y`. For example `f_add(x, z) = 1`, `f_logaddexp(x, z) = exp(x-z)`, and `f_mul(x, z)` doesn't exist because we can't define it for x=0. If we have `f`, we can implement jvp for `scatter_reduce` using `scatter_add` and `mul`.

For the cases where `f` does exist, I wonder if we can allow the user to supply this function, and the jvp rule works when `f` is supplied. It'd be nice if there can be a way of automatically deriving `f` given `computation`, but I am not very optimistic about that.

### DIFF
--- a/jax/_src/lax/slicing.py
+++ b/jax/_src/lax/slicing.py
@@ -611,6 +611,66 @@ def scatter(
       indices_are_sorted=indices_are_sorted, unique_indices=unique_indices,
       mode=GatherScatterMode.from_any(mode))
 
+
+def scatter_reduce(
+    operand: Array,
+    scatter_indices: Array,
+    updates: Array,
+    computation: Callable[[Array, Array], Array],
+    dimension_numbers: ScatterDimensionNumbers,
+    *,
+    indices_are_sorted: bool = False,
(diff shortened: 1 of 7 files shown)


kho:
Also the XLA generated code performs well on GPU comparing `segment_sum` and `segment_reduce(..., computation=jnp.logaddexp)`, but not so great on TPU. Unfortunately I doubt there is not much we can do in this PR about it.

kho:
No longer relevant.

PR #12018, opened 2022-08-19
experiment: add `jax.typing` module and use it in `jax.lax`

jakevdp:
Part of #12049

Changes:

- created `jax.typing` module, following roadmap in #11859
- moved `core.Shape` definition into `jax.typing`
- moved `jax.numpy.ndarray` definition into `jax.typing`
- applied these new types throughout `jax.lax`, following the style recommended in #11859
- fixed annotations in other files that were broken by the new `jax.lax` types (what a tangled web we weave...)

Some observations:
- When implementing annotations in `jax.lax`, I used `ArrayLike` for inputs and `NDArray` for outputs. Originally I had `NDArray` aliased to the `ndarray` metaclass, but because `jax.lax` has several functions that explicitly return `Tracer` types, it seemed necessary to make `NDArray = Union[ndarray, Tracer]`.
- This change had wide-ranging ripple effects: any function annotated with `-> jnp.ndarray` and returning the result of a lax operation caused a typecheck failure (because `Tracer` is inompatible), so it meant that in many places I had to change `-> jnp.ndarray` to `-> NDarray`, which caused even more ripples.
- An alternative would be to use `-> jnp.ndarray` by convention (with the understanding that it implies compatiblity with tracers) and use `#type: ignore` statements where tracers are explicitly returned.

Happy to hear any thoughts/opinions on this.

### DIFF
--- a/jax/_src/typing/ndarray.py
+++ b/jax/_src/typing/ndarray.py
@@ -11,9 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
 import abc
-from typing import Any, Tuple, Optional, Union
+from typing import Any, Optional, Tuple, Union
+from typing_extensions import Protocol
 
 from jax import core
 from jax.interpreters import pxla
@@ -42,6 +44,7 @@ class ndarray(metaclass=ArrayMeta):
   ndim: int
   shape: Tuple[int, ...]
   size: int
+  sharding: Any
 
   def __init__(self, shape, dtype=None, buffer=None, offset=0, strides=None,
                order=None):
@@ -292,3 +295,12 @@ ndarray.register(device_array.DeviceArray)
 for t in device_array.device_array_types:
   ndarray.register(t)
 ndarray.register(pxla._SDA_BASE_CLASS)
+
+
+
+class HasArrayMethod(Protocol):
+  def __array__(self) -> np.ndarray:
+    ...
+
+NDArray = Union[ndarray, core.Tracer]
+ArrayLike = Union[bool, int, float, complex, NDArray, np.ndarray, HasArrayMethod]
--- a/jax/_src/typing/ndarray.py
+++ b/jax/_src/typing/ndarray.py
@@ -297,10 +297,12 @@ for t in device_array.device_array_types:
 ndarray.register(pxla._SDA_BASE_CLASS)
 
 
-
 class HasArrayMethod(Protocol):
   def __array__(self) -> np.ndarray:
     ...
 
-NDArray = Union[ndarray, core.Tracer]
-ArrayLike = Union[bool, int, float, complex, NDArray, np.ndarray, HasArrayMethod]
+
+ArrayLike = Union[bool, int, float, complex, ndarray, core.Tracer, np.ndarray, HasArrayMethod]
+
+# TODO(jakevdp): make this an alias of jax._src.array.Array when available (#12049)
+Array = Union[ndarray, core.Tracer]
--- a/jax/_src/typing/simple.py
+++ b/jax/_src/typing/simple.py
@@ -1,0 +1,34 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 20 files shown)


patrick-kidger:
I think folks are definitely going to get confused between `google/jax.typing` and `google/jaxtyping`. Plus I'm very wary of yet more sharding within the JAX ecosystem.

Obviously the latter has more restricted aims, but wants to stay within the static type system.

Can we see if there's a reasonable way to bring both projects together? For example switched on the `if typing.TYPE_CHECKING` flag.

jakevdp:
> Can we see if there's a reasonable way to bring both projects together? For example switched on the `typing.TYPE_CHECKING` flag.

I'm not entirely sure what you have in mind... Do you mean that `jax` should depend on `jaxtyping` for static type checks? If so I don't think this is a good idea (I discussed this briefly in the roadmap draft). In my view, `jaxtyping` is far more complicated than the level of type annotations we want in the core library, at least for now.

jakevdp:
For context, this draft PR is more about "can we use simple annotations that are slightly better than `Array = Any` without breaking mypy and pytype". My understanding of `jaxtyping` is that it's far out of the scope of that question.

jax/_src/typing/ndarray.py
```diff
@@ -292,3 +295,14 @@ def weak_type(self) -> bool: ...
...
   ndarray.register(t)
 ndarray.register(pxla._SDA_BASE_CLASS)
+
+
+class HasArrayMethod(Protocol):
+  def __array__(self) -> np.ndarray:
+    ...
+
+
+ArrayLike = Union[bool, int, float, complex, ndarray, core.Tracer, np.ndarray, HasArrayMethod]
```
YouJiacheng, line 305: `SupportsArray` might be better than `HasArrayMethod` -- it aligns with stdlib https://docs.python.org/3/library/typing.html#protocols and numpy https://github.com/numpy/numpy/blob/6094eff95c19c433648593f3cb6bca489396f9ae/numpy/_typing/_array_like.py#L36-L38.
BTW, we can add a comment about why not directly use `Union[Array, numpy.typing.ArrayLike]` -- JAX ArrayLike does NOT allow (nested) sequences, `str` or `bytes`.

jax/_src/typing/simple.py
```diff
@@ -0,0 +1,34 @@
...
+
+# This file should avoid importing from jax in order to avoid circular imports.
+from __future__ import annotations
+
+from typing import Any, Sequence, Union
+from typing_extensions import Protocol
+import numpy as np
+
+class HasDtypeAttribute(Protocol):
+  dtype: Dtype
```
YouJiacheng, line 25: Must this protocol supports writeable `dtype` variable?
`numpy.typing` counterpart has read only `dtype` variable(property) https://github.com/numpy/numpy/blob/6094eff95c19c433648593f3cb6bca489396f9ae/numpy/_typing/_dtype_like.py#L83-L87. 

[PEP 544: By default, protocol variables as defined above are considered readable and writable. To define a read-only protocol variable, one can use an (abstract) property](https://peps.python.org/pep-0544/#protocol-members)

YouJiacheng: Also, name it as `SupportsDType`?

YouJiacheng: Hmmm, I guess we can just use `numpy.typing.DTypeLike` here since we use `np.dtype(x)` for every `DTypeLike` `x`.

jakevdp: `numpy.typing.DTypeLike` is too permissive (e.g. it allows structured types) see https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#inputs-should-be-permissively-typed

jax/_src/typing/ndarray.py
jakevdp: None of this is public API yet aside from `jax.Array`, so there's no issue with adjusting names if we wish.

We don't use `numpy.typing.ArrayLike` here because it's too permissive (e.g. it allows lists).

jax/_src/typing/simple.py
YouJiacheng: I got the point! But it seems that JAX use `np.dtype(xxx)` everywhere?

jax/_src/typing/ndarray.py
YouJiacheng: Yes, JAX does not allow nested sequence. I just suggest that we can add a comment in code about this reason(and/or a link to docs).

jax/_src/typing/simple.py
jakevdp: Yes it does. Invalid inputs would generate runtime errors later in the process.

jakevdp: Thanks for the suggestions - changes in #12813

jax/_src/typing/ndarray.py
jakevdp: Thanks for the suggestions - changes in https://github.com/google/jax/pull/12813

jakevdp:
(closing in favor of #12300)

PR #12037, opened 2022-08-22
target_total_secs has type `int` but used as type `None`

luca-digrazia:
**"filename"**: "benchmarks/benchmark.py"
**"warning_type"**: "Incompatible variable type [9]",
**"warning_message"**: " target_total_secs is declared to have type `int` but is used as type `None`."
**"warning_line"**: 86
**"fix"**: int to Optional[int]

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -83,7 +83,7 @@ def benchmark(f: Callable[[], Any], iters: Optional[int] = None,
 
 
 def benchmark_suite(prepare: Callable[..., Callable], params_list: List[Dict],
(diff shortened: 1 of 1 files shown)


hawkinsp:
We'd be happy to merge this fix, but you'll have to sign the CLA before we can.

luca-digrazia:
> We'd be happy to merge this fix, but you'll have to sign the CLA before we can.

Thanks for the message. I have already signed it. However, I did today again.

hawkinsp:
Thanks, that did the trick.

hawkinsp (approved):

PR #383, opened 2019-02-15
fix nan handling in pow jvp (fixes #347)

mattjj:
See #347 for example code.

Interestingly, Autograd still has this issue:

```python
import autograd.numpy as np
from autograd import grad

def f(x):
  x = np.sqrt(np.sum(x**2, axis=1))
  return np.sum(np.where(x > 0.5, x, np.ones_like(x)))

x = np.array([[1, 2], [3, 4], [0, 0]], dtype=np.float64)
print grad(f)(x)

# [[0.4472136  0.89442719]
#  [0.6        0.8       ]
#  [       nan        nan]]
```

It comes down to the formula for the jvp of `sqrt(x)` being problematic at `x=0` when the gradient has coefficients that are also zero, since `g * x ** -0.5` is `0 * inf` there. In this case, the zero coefficients in the gradient arise from the `lax.select`/`np.where` call.

I think a zero in the gradient should win out over an inf in the Jacobian, and this PR implements that policy. I had trouble implementing it on top of the standard `lax.mul` primitive because of linearity constraints in the JVP rule, so instead I added a new `lax._safe_mul` that implements the `0*inf = 0` policy in its translation rule.

PR #12041, opened 2022-08-22
Fix JVP rule for lax.pow()

### LINKED ISSUE #12033: The second order gradient of `lax.pow` is wrong
The second order gradient of `lax.pow` is wrong. For instance, the `fn` computes `x^y`, so `grad_fn` returns `(y*x^(y-1), ln(x)* x^y)`. When `x, y = 23.0, 0.0`, for the second element `ln(x) * x^y`, the gradient w.r.t `x` should be `1/x` instead of `23.0` computed by `jacrev` or `jacfwd`. 

```python
import jax
def grad_fn(x, y):
    def fn(x, y):
        return jax.numpy.sum(jax.lax.pow(x, y))
    return jax.grad(fn, (0, 1))(x, y)
x = jax.numpy.array([23.0])
y = jax.numpy.array([0.0])

print(jax.jacrev(grad_fn, (0, 1))(x, y))
print(jax.jacfwd(grad_fn, (0, 1))(x, y))
```

```
((DeviceArray([[0.]], dtype=float32), DeviceArray([[23.]], dtype=float32)), (DeviceArray([[0.04347826]], dtype=float32), DeviceArray([[9.831324]], dtype=float32)))
```

jakevdp:
Fixes #12033

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2004,8 +2004,7 @@ mlir.register_lowering(cbrt_p, partial(_nary_lower_mhlo, mhlo.CbrtOp))
 pow_p = standard_naryop([_float | _complex, _float | _complex], 'pow')
 
 def _pow_jvp_lhs(g, ans, x, y):
-  jac = mul(y, pow(x, select(eq(y, _zeros(y)), _ones(y), sub(y, _ones(y)))))
-  return mul(g, jac)
+  return mul(g, mul(y, pow(x, sub(y, _ones(y)))))
 
 def _pow_jvp_rhs(g, ans, x, y):
   return mul(g, mul(log(_replace_zero(x)), ans))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 
 ## jax 0.3.17 (Unreleased)
 * [GitHub commits](https://github.com/google/jax/compare/jax-v0.3.16...main).
+* Bugs
+  * Fix corner case issue in gradient of `lax.pow` with an exponent of zero
+    ({jax-issue}`12041`)
 * Breaking changes
   * {func}`jax.checkpoint`, also known as {func}`jax.remat`, no longer supports
     the `concrete` option, following the previous version's deprecation; see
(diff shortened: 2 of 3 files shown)


jax/_src/lax/lax.py
```diff
@@ -2001,8 +2001,7 @@ def _abs_jvp_rule(g, ans, x):
 pow_p = standard_naryop([_float | _complex, _float | _complex], 'pow')
 
 def _pow_jvp_lhs(g, ans, x, y):
-  jac = mul(y, pow(x, select(eq(y, _zeros(y)), _ones(y), sub(y, _ones(y)))))
-  return mul(g, jac)
+  return mul(g, mul(y, pow(x, sub(y, _ones(y)))))
```
jakevdp, line 2004: It's unclear to me why this rule ever included a select statement – @mattjj, @froystig, do you have any recollection why it was originally implemented this way?

jakevdp: I found the PR that implemented the orginal logic: https://github.com/google/jax/pull/383

jakevdp: Some git blame sleuthing by @mattjj found the origin of this select statement: https://github.com/HIPS/autograd/issues/116

mattjj: Whoever wrote that line is an idiot!

jakevdp:
I double-checked against the analog of the code in https://github.com/HIPS/autograd/issues/116, which inspired the introduction of this select statement. I get the same result before and after the change in this PR:
```python
import jax
import jax.numpy as jnp

def fun(x):
    return jnp.exp(-(x*x))
d1 = jax.grad(fun)
d2 = jax.grad(d1)
d3 = jax.grad(d2)
d4 = jax.grad(d3)
print(d1(0.), d2(0.), d3(0.), d4(0.))
# -0.0 -2.0 0.0 12.0

def fun(x):
    return jnp.exp(-(x**2))
d1 = jax.grad(fun)
d2 = jax.grad(d1)
d3 = jax.grad(d2)
d4 = jax.grad(d3)
print(d1(0.), d2(0.), d3(0.), d4(0.))
# -0.0 -2.0 0.0 12.0
```

jakevdp:
So, it turns out there is one (untested) case that this PR changes:
```python
jax.grad(jax.lax.pow, 0)(0.0, 0.0)
```
On `main`, this returns `0.0`; with this PR, it returns `NaN`.

Chatting offline with @mattjj, we realized this is similar in spirit to the `sinc` zero issue in (#5054); the resolution there was to do a custom JVP rule using the maclaurin series (#5077), and I think the most rigorously-correct solution here will probably be similar.

mattjj (approved):

PR #12046, opened 2022-08-22
remove 'BUG:' title prefix from bug report template

froystig:
We already auto-label with "bug" to this end.


jakevdp (approved):
Good call!

PR #12054, opened 2022-08-23
lax.squeeze: ensure DeviceArray is returned

jakevdp:
Why? Our convention is that `lax` functions return device arrays or tracers. See for example similar logic in `convert_element_type`: https://github.com/google/jax/blob/fd799742e2b48392329d3a8187d2d0d564c431ef/jax/_src/lax/lax.py#L573-L575

Discovered in the course of experimenting with stricter type annotations in #12018


hawkinsp (approved):
Test?

jakevdp:
> Test?

Done

PR #12077, opened 2022-08-24
Add a number of missing function cross-references in the docs.

hawkinsp:


jax/_src/ops/scatter.py
```diff
@@ -211,7 +211,8 @@ def segment_sum(data: Array,
       would support all indices in ``segment_ids``, calculated as
       ``max(segment_ids) + 1``.
       Since `num_segments` determines the size of the output, a static value
-      must be provided to use ``segment_sum`` in a ``jit``-compiled function.
+      must be provided to use ``segment_sum`` in a :func:`~jax.jit`-compiled
```
jakevdp, line 214: Here and below: I have a slight-preference here for using `JIT-compiled`, since this renders as0 "[`jit()`](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html)-compiled" which looks strange

jakevdp (approved):
Looks great! I have a couple minor nit-picky comments below; feel free to skip them if you disagree 😁

PR #12079, opened 2022-08-24
Add scan to checkify docs and fix rendering of ErrorCategories in docs.

LenaMartens:
Fixes #12078


docs/debugging/checkify_guide.md
```diff
@@ -251,7 +251,42 @@ jax.grad(assert_gradient_negative)(-1.)
 # ValueError: gradient needs to be negative!
 ```
 
-<!-- TODO: scan? -->
+### `scan`/`while_loop`/`cond`
+
+Checkify works with all higher-order control flow operators, like `scan`,
```
hawkinsp, line 256: Can you make these into cross-references? I forget the markdown syntax for it; in restructured text it is :func:`~jax.lax.scan`
Similarly for while_loop and cond.

docs/debugging/checkify_guide.md
```diff
@@ -251,7 +251,42 @@ jax.grad(assert_gradient_negative)(-1.)
...
+`while_loop` and `cond`.
+
+For example, you can have a `scan` in your checked function, and the body of the
+`scan` can have a `checkify.check`:
```
hawkinsp, line 260: Make checkify.check a cross reference.

docs/debugging/checkify_guide.md
hawkinsp: I think it's
```
{func}`jax.lax.scan`
```
in markdown.

jax/_src/checkify.py
```diff
@@ -785,12 +785,19 @@ def assert_discharge_rule(error, enabled_errors, pred, code, payload, *, msgs):
 
 ErrorCategory = enum.Enum('ErrorCategory', ['NAN', 'OOB', 'DIV', 'USER_CHECK'])
 
-user_checks = frozenset({ErrorCategory.USER_CHECK})
+
+user_checks = frozenset({ErrorCategory.USER_CHECK})  #: Enables all :func:`check` assertions.
+#: Checks if all floating-point operations output non-NaN values.
```
LenaMartens, line 790: I'm actually unsure of how to add documentation to the ErrorCategory sets in a way that the autodocs generation understands. I thought I could do it with these doc-comments (`#: ...`) but they're not rendering. I'll have another look at documenting "module level variables" but maybe I'm missing something obvious?

PR #12090, opened 2022-08-24
Add callbacks to docs

sharadmv:


docs/jax.rst
```diff
@@ -99,3 +99,12 @@ Parallelization (:code:`pmap`)
     device_count
     local_device_count
     process_count
+
+Callbacks
+----------------------------
```
jakevdp, line 104: We need the underline to match the length of the text (sphinx will error otherwise)

jakevdp (changes requested):

docs/jax.rst
sharadmv: Done.

jakevdp (approved):

PR #12101, opened 2022-08-25
Add option to generate perfetto trace without generating link

sharadmv:
This is useful for generating Perfetto-compatible traces without blocking to click on a link. Users can manually upload these traces to Perfetto afterwards.

### DIFF
--- a/jax/_src/profiler.py
+++ b/jax/_src/profiler.py
@@ -72,12 +72,14 @@ class _ProfileState:
     self.profile_session = None
     self.log_dir = None
     self.create_perfetto_link = False
+    self.create_perfetto_trace = False
     self.lock = threading.Lock()
 
 _profile_state = _ProfileState()
 
 
-def start_trace(log_dir, create_perfetto_link: bool = False):
+def start_trace(log_dir, create_perfetto_link: bool = False,
+                create_perfetto_trace: bool = False):
   """Starts a profiler trace.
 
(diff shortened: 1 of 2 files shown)


jax/_src/profiler.py
```diff
@@ -96,6 +98,10 @@ def start_trace(log_dir, create_perfetto_link: bool = False):
     create_perfetto_link: A boolean which, if true, creates and prints link to
       the Perfetto trace viewer UI (https://ui.perfetto.dev). The program will
       block until the link is opened and Perfetto loads the trace.
+    create_perfetto_trace: A boolean which, if true, additionally dumps a
+      ``perfetto_trace.json.gz`` file that is compatible for upload with the
+      Perfetto trace viewer UI (https://ui.perfetto.dev). The file will also be
```
skye, line 103: Maybe add that this is useful if you wanna generate the trace without blocking execution.

jax/_src/profiler.py
```diff
@@ -209,8 +220,12 @@ def trace(log_dir, create_perfetto_link=False):
...
```
skye, line 225: Same here

skye (approved):

jax/_src/profiler.py
sharadmv: Done!

jax/_src/profiler.py
sharadmv: Done.

PR #12112, opened 2022-08-25
Fix typo in debugging docs

sharadmv:


mattjj:
Oh hey while you're at it, we should change this to say "`jit`-, `pmap`, and `jax.pjit`-decorated":
image

And here add the word "to" in "Need to potentially use many breakpoints **to** pinpoint the source of an error":
image

mattjj (approved):

sharadmv:
Done!

Issue #12119, opened 2022-08-26
`psum` fails with 1 process per GPU


sudhakarsingh27:
### Description

After assigning GPU to a process using `jax.config.update("jax_cuda_visible_devices", localid)`, `psum` fails with the following error (localid = local process id) (I added additional debug print statement in XLA code to confirm):

```
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: RET_CHECK failure (external/org_tensorflow/tensorflow/compiler/xla/service/gpu/gpu_executable_run_options.cc:75) 0 <= local_device_ordinal && local_device_ordinal < gpu_global_device_ids->size()
local_device_ordinal: 3
gpu_global_device_ids size: 1
```

Note that such a failure doesn't occur if we use `os.environ["CUDA_VISIBLE_DEVICES"]=localid`

---

Here's the repro code:

```
import jax
import os
from absl import flags, app

flags.DEFINE_string('server_addr', 'localhost:54321', help='server ip addr')
flags.DEFINE_integer('num_hosts', 1, help='num of hosts' )
flags.DEFINE_integer('host_idx', 0, help='index of current host' )
FLAGS = flags.FLAGS

def main(argv):
    print(FLAGS.server_addr, FLAGS.num_hosts, FLAGS.host_idx)

    jax.config.update("jax_cuda_visible_devices", str(FLAGS.host_idx))

    jax.distributed.initialize(FLAGS.server_addr, FLAGS.num_hosts, FLAGS.host_idx)


    print(f"Device count: {jax.device_count()}")

    print(f"Local device count: {jax.local_device_count()}")
   
    xs = jax.numpy.ones(jax.local_device_count())
    y = jax.pmap(lambda x: jax.lax.psum(x, 'i'), axis_name='i')(xs)

    print(f"output: {y}")

if __name__ == '__main__':
  app.run(main)
```

Launch 4 processes (or any number > 1) within shell as follows
```
python manual_dist.py --server_addr=127.0.0.1:54321 --num_hosts=4 --host_idx=0 &
python manual_dist.py --server_addr=127.0.0.1:54321 --num_hosts=4 --host_idx=1 &
python manual_dist.py --server_addr=127.0.0.1:54321 --num_hosts=4 --host_idx=2 &
python manual_dist.py --server_addr=127.0.0.1:54321 --num_hosts=4 --host_idx=3 &
```

### What jax/jaxlib version are you using?

jax/jaxlib 0.3.17 (built from head)

[...]

PR #12129, opened 2022-08-27
Enable one gpu per process for multinode GPU tests

sudhakarsingh27:
1. Use one process per GPU
2. Try `jax.config.update("jax_cuda_visible_devices", visible_devices)` to check multinode CI job against the fix for https://github.com/google/jax/issues/12119


yashk2810:
I'll approve on Monday. It'll also give time for nightly to be released with the fix.

yashk2810 (approved):

PR #12131, opened 2022-08-27
[ROCm] Disable CUDA specific test for ROCm.

rsanthanam-amd:
/cc @hawkinsp

### DIFF
--- a/tests/multiprocess_gpu_test.py
+++ b/tests/multiprocess_gpu_test.py
@@ -153,6 +153,13 @@ class MultiProcessGpuTest(jtu.JaxTestCase):
         'int(os.environ["NUM_TASKS"]), int(os.environ["TASK"])); '
         's = jax.pmap(lambda x: jax.lax.psum(x, "i"), axis_name="i")(jax.numpy.ones(jax.local_device_count())); '
         'print(f\'{jax.local_device_count()},{jax.device_count()},{s}\', end=""); '
+      ) if jtu.is_device_cuda() else (
+        'import jax, os; '
+        f'jax.config.update("jax_rocm_visible_devices", "{visible_devices}"); '
+        'jax.distributed.initialize('
+        'f\'localhost:{os.environ["JAX_PORT"]}\', '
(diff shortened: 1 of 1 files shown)


tests/multiprocess_gpu_test.py
```diff
@@ -111,6 +111,7 @@ def test_gpu_distributed_initialize(self):
 
   @unittest.skipIf(xla_extension_version < 91,
                    "Test requires jaxlib 0.3.17 or newer")
+  @jtu.skip_on_devices("rocm")
```
hawkinsp, line 114: FWIW, the test can probably work if you adapt it to set `jax_rocm_visible_devices` as well.

rsanthanam-amd: The recent modifications to that test did not work on ROCm.  I thought it might be cleaner to add a separate ROCm version which did things the previous way.

PTAL.

hawkinsp: I'd prefer we consolidate this with the cuda version. The only difference between these two is the name of the flag, right? Make the test parametric instead?

rsanthanam-amd: Actually, there are other differences beyond just the flag name.

Some recent changes were made to that particular subtest that did not work on ROCm.

So the ROCm specific version essentially reverts the latest changes in addition to changing the name of the flag.

rsanthanam-amd: I modified the existing test to support ROCm.

Modifications beyond just changing `jax_cuda_visible_devices` to `jax_rocm_visible_devices` were necessary because ROCm does not support the statement `s = jax.pmap(lambda x: jax.lax.psum(x, "i"), axis_name="i")(jax.numpy.ones(jax.local_device_count()));`

The reason is that a valid nccl_unique_id_callback must be supplied to `GetNcclUniqueIdCallback` from `tensorflow/compiler/xla/service/gpu/nccl_utils.cc` and this is not supported on ROCm:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/pjrt/gpu_device.cc#L386-L393

PR #12143, opened 2022-08-29
JAX implementation of scipy.stats.multinomial pmf & logpmf

### LINKED ISSUE #11817: JAX implementation of scipy.stats.multinomial

wonhyeongseo:
fix https://github.com/google/jax/issues/11817

Implementation originally made by @harryjulian.
I just made minor changes and am making a separate PR for cleaner review as @jakevdp has suggested.
I will make improvements upon review asap.

Hope you have a great week ahead :) Thanks!

### DIFF
--- a/jax/_src/scipy/stats/multinomial.py
+++ b/jax/_src/scipy/stats/multinomial.py
@@ -1,0 +1,37 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import scipy.stats as osp_stats
+from jax import lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.util import _wraps, _promote_args_inexact, _promote_args_numeric
+from jax._src.scipy.special import gammaln, xlogy
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.logpmf."""
+  p, = _promote_args_inexact("multinomial.logpmf", p)
+  x, n = _promote_args_numeric("multinomial.logpmf", x, n)
+  if not jnp.issubdtype(x.dtype, jnp.integer):
+    raise ValueError(f"x and n must be of integer type; got x.dtype={x.dtype}, n.dtype={n.dtype}")
+  x = x.astype(p.dtype)
+  n = n.astype(p.dtype)
+  logprobs = gammaln(n + 1) + jnp.sum(xlogy(x, p) - gammaln(x + 1), axis=-1)
+  return jnp.where(jnp.equal(jnp.sum(x), n), logprobs, -jnp.inf)
+
+@_wraps(osp_stats.multinomial.pmf, update_doc=False)
+def pmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.pmf."""
+  return lax.exp(logpmf(x, n, p))
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -565,6 +565,30 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
       check_dtypes=False)
 
   @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape, shape], [x_dtype, p_dtype]),
+       "x_dtype": x_dtype,
+       "p_dtype": p_dtype,
+       "shape": shape}
+      for shape in [(2), (4,), (1, 5)]
+      for (x_dtype, p_dtype) in itertools.product(jtu.dtypes.integer, jtu.dtypes.floating)))
+  def testMultinomialLogPmf(self, shape, x_dtype, p_dtype):
+    rng = jtu.rand_positive(self.rng())
+    scipy_fun = osp_stats.multinomial.logpmf
+    lax_fun = lsp_stats.multinomial.logpmf
+
+    def args_maker():
+      x = rng(shape, x_dtype)
+      n = np.sum(x, dtype=x.dtype)
+      p = rng(shape, p_dtype)
+      # Normalize the array such that it sums it's entries sum to 1 (or close enough to)
+      p = p / np.sum(p)
+      return [x, n, p]
+
+    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                            tol=5e-4)
+    self._CompileAndCheck(lax_fun, args_maker, rtol=1e-5, atol=1e-5)
+
+  @parameterized.named_parameters(jtu.cases_from_list(
       {"testcase_name": "_x={}_mean={}_cov={}".format(
           jtu.format_shape_dtype_string(x_shape, x_dtype),
           jtu.format_shape_dtype_string(mean_shape, mean_dtype)
--- a/jax/scipy/stats/__init__.py
+++ b/jax/scipy/stats/__init__.py
@@ -21,6 +21,7 @@ from jax.scipy.stats import gamma as gamma
 from jax.scipy.stats import geom as geom
 from jax.scipy.stats import laplace as laplace
 from jax.scipy.stats import logistic as logistic
+from jax.scipy.stats import multinomial as multinomial
 from jax.scipy.stats import multivariate_normal as multivariate_normal
 from jax.scipy.stats import nbinom as nbinom
 from jax.scipy.stats import norm as norm
--- a/jax/scipy/stats/multinomial.py
+++ b/jax/scipy/stats/multinomial.py
@@ -1,0 +1,19 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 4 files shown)


jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,34 @@
...
+import scipy.stats as osp_stats
+from jax import lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.util import _wraps, _promote_args_inexact
+from jax._src.scipy.special import gammaln, xlogy
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.logpmf."""
+  x = jnp.array(x.astype(int))
```
jakevdp, line 25: Rather than unconditionally converting `x` to integer type, we should error if the `x` input is not integer typed to begin with (cf. discussion at https://github.com/google/jax/pull/11879#discussion_r955320248)

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,34 @@
...
+  p = _promote_args_inexact("multinomial.logpmf", p)[0]
+  logprobs = gammaln(n + 1) + jnp.sum(xlogy(x, p) - gammaln(x + 1), axis=-1)
```
jakevdp, line 27: This is going to cause strict type promotion errors because `x` and `p` have different types. We should convert `x` to the type of `p` before attempting operations between them.

wonhyeongseo: Yes! Thank you for clarifying sir :) 
I was going back and forth with this.

jax/_src/scipy/stats/multinomial.py
wonhyeongseo: Ah, that is much better. What should the error message be like? Right now it seems a bit- unprofessional?

jax/_src/scipy/stats/multinomial.py
wonhyeongseo: @jakevdp Is promoting different from casting here? I chose this path because `_promote_args_inexact` seems to convert all of the dtypes to a subtype of `floating`. Is there a possibility for x being `float32` and p being `float64`?

jakevdp: If you run the tests locally you should see the issue: promoting here is implicit, and will cause test failures due to the `jax_numpy_dtype_promotion=strict` flag used by the test suite. Here we expect promotion because `p` is generally a float, and `x` is required to be an int.

wonhyeongseo:
@jakevdp I think I addressed the problems you have pointed out, sir. May you please review my updated code?

jakevdp:
I triggered the test runs, I believe they will fail because you're still doing implicit type promotion between ints and floats. We need explicit conversions.

jakevdp:
You should try running those tests locally, because it will make the iteration time much faster! There's some info here: https://jax.readthedocs.io/en/latest/developer.html?highlight=pytest#using-pytest

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
       lsp_stats.norm.cdf(np.full((4,), np.inf, np.float32)),
       check_dtypes=False)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape], [dtype]),
```
jakevdp, line 568: I thin this should be `[shape], dtype` because `dtype` is actually a tuple of dtypes.

This naming is relatively confusing, so perhaps below you could use `for (x_dtype, p_dtype) in itertools.product(...` and adjust the function signature accordingly.

wonhyeongseo:
Hello @jakevdp, 
image
Finally passed the type conversion test. May you please run the workflow to test the code? Thank you so much for all your guidance!

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,36 @@
...
+    raise ValueError(f"x is of dtype {x.dtype}, it should be an integer.")
+  x = jnp.asarray(x, p.dtype)
+  n = jnp.asarray(n, p.dtype)
+  logprobs = gammaln(n + 1) + jnp.sum(xlogy(x, p) - gammaln(x + 1), axis=-1)
+  return jnp.where(jnp.equal(jnp.sum(x), n), logprobs, -jnp.inf)
+
+@_wraps(osp_stats.multinomial.pmf, update_doc=False)
+def pmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.pmf."""
+  x, p = _promote_args_inexact("multinomial.pmf", x, p)
```
jakevdp, line 35: Delete this line? (all the promotion logic should be already handled in `logpmf`)

wonhyeongseo: Hello, @jakevdp! I have deleted line 35 as you have instructed. Thank you so much 😄

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,35 @@
...
+from jax._src.numpy.util import _wraps, _promote_args_inexact
+from jax._src.scipy.special import gammaln, xlogy
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.logpmf."""
+  if not jnp.issubdtype(x.dtype, jnp.integer):
+    raise ValueError(f"x is of dtype {x.dtype}, it should be an integer.")
+  x = jnp.asarray(x, p.dtype)
+  n = jnp.asarray(n, p.dtype)
```
jakevdp, line 28: Sorry to harp on this, but this still seems a bit off – for example, imagine someone puts in integer values for `p`. Then we'll convert everything to integer type, and the functions below will fail because they require floats.

There's also extra validation logic in `_promote_args_*` that is now missing from this implementation.

What if we do something like this instead:
```python
  p, = _promote_args_inexact("multinomial.logpmf", p)
  x, n = _promote_args_numeric("multinomial.logpmf", x, n)
  if not jnp.issubdtype(x.dtype, jnp.integer):
    raise ValueError(f"x and n must be of integer type; got dtype={x.dtype}")
  x = x.astype(p.dtype)
  n = n.astype(p.dtype)
```

jax/_src/scipy/stats/multinomial.py
```diff
@@ -0,0 +1,37 @@
...
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.util import _wraps, _promote_args_inexact, _promote_args_numeric
+from jax._src.scipy.special import gammaln, xlogy
+
+@_wraps(osp_stats.multinomial.logpmf, update_doc=False)
+def logpmf(x, n, p):
+  """JAX implementation of scipy.stats.multinomial.logpmf."""
+  p, = _promote_args_inexact("multinomial.logpmf", p)
+  x, n = _promote_args_numeric("multinomial.logpmf", x, n)
+  if not (jnp.issubdtype(x.dtype, jnp.integer) and jnp.issubdtype(n.dtype, jnp.integer)):
```
jakevdp, line 27: nitpick: `x` and `n` have the same dtype here (this is always the case after `_promote_args_*`, so no need to check both.

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,32 @@ def testIssue972(self):
       lsp_stats.norm.cdf(np.full((4,), np.inf, np.float32)),
       check_dtypes=False)
 
+  @parameterized.named_parameters(jtu.cases_from_list(
+      {"testcase_name": jtu.format_test_name_suffix("", [shape], [x_dtype, p_dtype]),
+       "x_dtype": x_dtype,
+       "p_dtype": p_dtype,
+       "shape": shape}
+      for shape in [(2,), (3,), (4,), (5,), (6,), (7,), (1,2), (1, 3), (1,4), (1,5)]
```
jakevdp, line 572: I think we can reduce the number of shapes here to avoid generating an excessive number of test cases. Maybe `fore shape in [(), (4,), (1, 5)]`

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
...
+
+    def args_maker():
+      x = rng(shape, x_dtype)
+      n = np.sum(x)
+      p = rng(shape, p_dtype)
+      # Normalize the array such that it sums it's entries sum to 1 (or close enough to)
+      p = p / np.sum(p)
+      return [x, n, p]
+
+    scipy_fun = osp_stats.multinomial.logpmf
```
jakevdp, line 587: Remove repeated definition

tests/scipy_stats_test.py
```diff
@@ -564,6 +564,31 @@ def testIssue972(self):
...
+    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                              tol=5e-4)
```
jakevdp, line 589: Nit: spacing is off (it should line up with the open parentheses above it

jakevdp:
Looks great! Thanks for sticking with this. The last thing we need is to squash all the commits into one, and force push to this branch (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Let me know if you have any questions

jakevdp:
I think we're ready to merge this, but we still need all the commits squashed. Let me know if you need help with that process; alternatively if I have write access to your branch I could take care of it for you.

wonhyeongseo:
Dear @jakevdp,

Hello sir! I really wanted to try rebasing this on my own, so I asked friends and watched a LOT of tutorials. I've learned to definitely keep my branches tidy from now on, but still need help with this spaghetti I've conjured up (ramen). May you please help me? https://github.com/wonhyeongseo/jax/invitations

Love and best regards,
@wonhyeongseo

jakevdp:
Hi - thanks, I'm happy to help you get this merged, but there are still a few test failures to look into. In particular, the test case is problematic because for integer inputs, `np.sum` may not return the same dtype. So I think you need to use `n = np.sum(x, dtype=x.dtype)`.

jakevdp:
That sounds good, thanks!

jakevdp:
Working on unravelling your branch's commit history... it's impressive

jakevdp (approved):
Thanks for all the hard work on this!

PR #12149, opened 2022-08-29
Add ability to close over `Ref`s in nested `for_loop`s

sharadmv:
Tracker: #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -15,7 +15,7 @@
 from functools import partial
 import operator
 
-from typing import Any, Callable, Generic, List, Optional, Sequence, Set, Tuple, TypeVar
+from typing import Any, Callable, Generic, List, Optional, Sequence, Set, Tuple, TypeVar, Union
 
 from jax import core
 from jax import lax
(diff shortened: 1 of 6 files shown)


jax/_src/lax/control_flow/for_loop.py
```diff
@@ -70,21 +70,23 @@ class Ref(Generic[T]): pass
 ### Tracing utilities
 
 def _hoist_consts_to_refs(jaxpr: core.Jaxpr) -> core.Jaxpr:
-  num_consts = len(jaxpr.constvars)
+  all_const_avals = [var.aval for var in jaxpr.constvars]
```
mattjj, line 73: Note to future selves: this function used to take a jaxpr with refs and non-ref inputs, where the non-refs were exactly closed-over consts, and produce a jaxpr with only ref inputs where the consts were read. But after this change, closed-over things could be refs (not just non-refs) introduced by an outer `for` loop, so the type signature is the same but in the implementation we don't want to e.g. wrap refs in refs.

jax/_src/lax/control_flow/for_loop.py
```diff
@@ -138,6 +141,16 @@ def for_loop(nsteps, body, init_state):
   Returns:
     A Pytree of values representing the output of the for loop.
   """
+  if isinstance(nsteps, int):
```
mattjj, line 144: Note to future selves: this is a slick way to generate nested loops at the api level (w/o making the primitive more complex), but we're not sure if keeping the loops flattened in the lowering could be advantageous, and we're also not sure if that's hard to recover from nested loops.

jax/_src/state/discharge.py
```diff
@@ -48,6 +53,81 @@ def discharge_state(jaxpr: core.Jaxpr, consts: Sequence[Any]
...
+
+def register_discharge_rule(prim: core.Primitive):
+  def register(f: DischargeRule):
+    _discharge_rules[prim] = f
+  return register
+
+def _has_refs(eqn: core.JaxprEqn):
+  return any(isinstance(v.aval, ShapedArrayRef) for v in eqn.invars)
+
+def _eval_jaxpr_discharge_state(jaxpr: core.Jaxpr, consts: Sequence[Any],
```
mattjj, line 85: Note to selves: before this PR, `_eval_jaxpr_dischage_state` had like an if/else chain for our three effectful primitives. But now that `for_loop` as a HOP can itself be effectful, it was a good time to refactor into a more general interpreter with `_discharge_rules`.

jax/_src/state/discharge.py
```diff
@@ -48,6 +53,81 @@ def discharge_state(jaxpr: core.Jaxpr, consts: Sequence[Any]
...
+                                *args: Any):
+  env = Environment({})
+
+  map(env.write, jaxpr.constvars, consts)
+  # Here some args may correspond to `Ref` avals but they'll be treated like
+  # regular values in this interpreter.
+  map(env.write, jaxpr.invars, args)
+
+  for eqn in jaxpr.eqns:
+    if _has_refs(eqn):
```
mattjj, line 95: Note to selves: Can we just check the eqn's effects tag here? Actually no! Consider a `for_loop` application which takes in a ref (eg by closing over it) but has an empty body. Because we say effects live on arrows, the empty body means there's no effect on the inner jaxpr or (given how we compute a `for_loop` application eqn's effects now) thus on the eqn. But we still want this function `_eval_jaxpr_discharge_state` to get rid of this ref!

In other words, we want this function to take a jaxpr like this:

```
{ lambda a:Ref . let
  in () } : (Ref -> ())
```

and produce a jaxpr which doesn't have any Ref-typed input binders.

So in a way we're really getting rid of Refs even more than we're discharging state.

jax/_src/lax/control_flow/for_loop.py
```diff
@@ -229,9 +242,33 @@ def _get_ref_state_effects(jaxpr: core.Jaxpr) -> List[Set[StateEffect]]:
...
+  return list(avals), nonlocal_state_effects
+
+@state.register_discharge_rule(for_p)
+def _for_discharge_rule(in_avals, *args: Any, jaxpr: core.Jaxpr,
+                        reverse: bool, which_linear: Sequence[bool],
+                        nsteps: int
+                        ) -> Tuple[Sequence[Optional[Any]], Sequence[Any]]:
+  state_avals = [state.ShapedArrayRef(v.aval.shape, v.aval.dtype) for v in
+                 jaxpr.invars[1:]]
+  jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(
```
mattjj, line 262: Maybe we don't need to make a new jaxpr! We're basically substituting ref-valued arguments with non-ref-valued arguments, but the `for_loop` itself (including its body jaxpr) can stay the same.

mattjj (approved):
Great work! Sharad and I went over this together and I wrote some notes from our conversation.

mattjj:
Another note: if we separated out `runState`, then we could have `for_loop` apply only to `Ref`s, rather than having the kind of polymorphism we have now where it can be applied to either `Ref`s or non-`Ref`s (i.e. values). But things would essentially look the same.

PR #12152, opened 2022-08-29
Add slurm multi-process auto bootstrapping interface

nvcastet:
* Generic bootstrapping interface to plug in new environments
* Slurm bootstrapping

### DIFF
--- a/docs/multi_process.md
+++ b/docs/multi_process.md
@@ -64,7 +64,9 @@ The API {func}`jax.distributed.initialize` takes several arguments, namely:
     cluster will connect.
   * `num_processes`: the number of processes in the cluster
   * `process_id`: the ID number of this process, in the range `[0 ..
-  num_processes)`.
+    num_processes)`.
+  * `local_device_ids`: Restricts the visible devices of the current process to
+    ``local_device_ids``.
 
 For example on GPU, a typical usage is:
 
@@ -76,9 +78,11 @@ jax.distributed.initialize(coordinator_address="192.168.0.1:1234",
                            process_id=0)
 ```
 
-On Cloud TPU, you can simply call {func}`jax.distributed.initialize()` with no
-arguments. Default values for the arguments will be chosen automatically using
-the TPU pod metadata:
+On Cloud TPU and Slurm environments, you can simply call {func}`jax.distributed.initialize()` with no
+arguments. Default values for the arguments will be chosen automatically.
+When running on GPUs with Slurm, it is assumed that one process is started per GPU, i.e. each process will
+be assigned only one visible local device. Otherwise it is assumed that one process is started per host,
+i.e. each process will be assigned all local devices.
 
 ```python
 import jax
--- a/jax/_src/clusters/__init__.py
+++ b/jax/_src/clusters/__init__.py
@@ -1,0 +1,24 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .cluster import ClusterEnv
+
+# Order of declaration of the cluster environments
+# will dictate the order in which they will be checked.
+# Therefore, if multiple environments are available and
+# the user did not explicitly provide the arguments
+# to :func:`jax.distributed.initialize`, the first
+# available one from the list will be picked.
+from .slurm_cluster import SlurmCluster
+from .cloud_tpu_cluster import TpuCluster
--- a/jax/_src/clusters/cluster.py
+++ b/jax/_src/clusters/cluster.py
@@ -1,0 +1,109 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List, Optional, Type, Sequence, Tuple
+from absl import logging
+from jax._src.cloud_tpu_init import running_in_cloud_tpu_vm
+
+
+class ClusterEnv:
+  """Interface for defining a cluster environment.
+
+  To enable auto bootrapping (aka :func:`jax.distributed.initialize()`),
+  cluster environments need to derive from :class:`ClusterEnv` and implement
+  :func:`is_env_present`, :func:`get_coordinator_address`,
+  :func:`get_process_count`, and :func:`get_process_id`.
+  :class:`ClusterEnv` subclasses are automatically detected when imported.
+  """
+
+  _cluster_types: List[Type['ClusterEnv']] = []
+
+  def __init_subclass__(cls, **kwargs):
+    super().__init_subclass__(**kwargs)
+    cls._cluster_types.append(cls)
+
+  @classmethod
+  # pytype: disable=bad-return-type
+  def auto_detect_unset_distributed_params(cls,
+                                           coordinator_address: Optional[str],
+                                           num_processes: Optional[int],
+                                           process_id: Optional[int],
+                                           local_device_ids: Optional[Sequence[int]]
+                                          ) -> Tuple[Optional[str], Optional[int], Optional[int],
+                                                     Optional[Sequence[int]]]:
+    if all(p is not None for p in (coordinator_address, num_processes,
+      process_id, local_device_ids)):
+      return (coordinator_address, num_processes, process_id,
+              local_device_ids)
+    env = next((env for env in cls._cluster_types if env.is_env_present()), None)
+    if env:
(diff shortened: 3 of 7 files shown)


skye (changes requested):
This is neat! Overall I like the idea, but I have some thoughts on the overall structure and some details of the Cluster API. Lemme know if you wanna discuss any of this over video or chat, feel free to email or Google chat me.

jax/_src/clusters/cluster_init.py
```diff
@@ -0,0 +1,63 @@
...
+      return coordinator_address, num_processes, process_id
+
+    @classmethod
+    @abstractmethod
+    def is_env_present(cls):
+      ...
+
+    @classmethod
+    @abstractmethod
+    def get_coordinator_address(cls):
```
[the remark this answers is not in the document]
nvcastet: Yes good point, I can add some doc here.

jax/_src/clusters/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .cluster_init import Cluster
+from .cloud_tpu_init import TpuCluster
+from .slurm_init import SlurmCluster
```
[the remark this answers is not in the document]
nvcastet: Correct, the order will dictate a priority if several environments are available. I will add a comment.

jax/_src/clusters/slurm_init.py
```diff
@@ -0,0 +1,45 @@
...
+_LOCAL_PROCESS_ID = 'SLURM_LOCALID'
+
+class SlurmCluster(Cluster):
+  @classmethod
+  def is_env_present(cls):
+    return _JOBID_PARAM in os.environ
+
+  @classmethod
+  def get_coordinator_address(cls):
+    # Pick port in ephemeral range [(65535 - 2^3 + 1), 65535]
```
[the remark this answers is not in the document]
nvcastet: Good catch, I meant to have a bigger range like the last `2^12` ports of the ephemeral range but forgot to update it. Yes you cannot use `/proc/sys/net/ipv4/ip_local_port_range` since the logic to get the port needs to lead to the exact same port on all the processes to be able to connect to the open port of the 1st process. Using `/proc/sys/net/ipv4/ip_local_port_range` may give you different answer on different nodes.

jax/_src/clusters/cluster_init.py
```diff
@@ -0,0 +1,63 @@
...
+      ...
+
+    @classmethod
+    @abstractmethod
+    def default_to_single_process_per_device(cls):
+      ...
+
+    @staticmethod
+    def is_single_process_per_device(env):
+      user_var = os.environ.get('JAX_SINGLE_PROCESS_PER_DEVICE', None)
```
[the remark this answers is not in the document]
nvcastet: @skye Correct, we already know the number of total processes of the job and also the number of processes per node.

The only case which is an issue is when the cluster tells you there are 1 process and 1 node. How do we know if the user wants to run on a single process with one device, or with all the devices, or in-between. We could make it flexible as you mentioned and have the user set `JAX_VISIBLE_DEVICES`. But what would be a good default? For GPUs, we would like to default to 1 but for other cases to all.
We don't need to worry about `CUDA_VISIBLE_DEVICES`, this is a CUDA driver setting, so the CUDA API that XLA is using will honor `CUDA_VISIBLE_DEVICES`.

nvcastet:
@skye I rebased the code and added a commit to address your comments. Let me know what you think.

jax/_src/clusters/slurm_cluster.py
```diff
@@ -0,0 +1,46 @@
...
+    return _JOBID_PARAM in os.environ
+
+  @classmethod
+  def get_coordinator_address(cls):
+    # Pick port in ephemeral range [(65535 - 2^12 + 1), 65535]
+    port = int(os.environ[_JOBID_PARAM]) % 2**12 + (65535 - 2**12 + 1)
+
+    # Parse the first hostname of the job
+    # node_list potential formats are 'node1', 'node1,host2',
+    # 'node[001-0015],host2', and 'node[001,007-015],host2'.
```
skye, line 23: I'm confused what `node` and `host` mean, can you also comment which part is the hostname we're looking for?

jax/_src/clusters/cluster.py
```diff
@@ -0,0 +1,78 @@
...
+  """
+
+  _cluster_types: List['ClusterEnv'] = []
+
+  def __init_subclass__(cls, **kwargs):
+    super().__init_subclass__(**kwargs)
+    cls._cluster_types.append(cls)
+
+  @classmethod
+  def auto_detect_unset_distributed_params(cls, coordinator_address,
```
skye, line 22: Can you put types on the params and outputs for all of these interface methods?

jax/_src/clusters/cluster.py
```diff
@@ -0,0 +1,78 @@
...
+  @classmethod
+  def get_process_count(cls):
+    raise NotImplementedError
+
+  @classmethod
+  def get_process_id(cls):
+    raise NotImplementedError
+
+  @classmethod
+  def get_local_process_id(cls):
```
skye, line 77: This can apparently return None, so I think it needs a docstring explaining what that means. And should this return None by default instead of raising NotImplemented?

jax/_src/clusters/cluster.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    raise NotImplementedError
+
+  @classmethod
+  def get_process_count(cls):
+    raise NotImplementedError
+
+  @classmethod
+  def get_process_id(cls):
+    raise NotImplementedError
```
skye, line 74: I think `get_coordinator_address`, `get_process_count`, and `get_process_id` should be abstract methods (https://docs.python.org/3/library/abc.html#abc.abstractmethod)

skye (changes requested):
It's looking good! Most of the comments are just to make the code clearer for future readers + users

jax/_src/clusters/cluster.py
nvcastet: It cannot be None. To make sure it is dealt with by the subclass, it would let the exception. I can add a docstring on what it means.

jax/_src/clusters/cluster.py
nvcastet: You may have seem that I had them as abstract methods in the first version of the PR. By trying it and reading the doc carefully, abstract method decorator in python will only help if you are instantiating subclass objects. In our case we are never creating objects, we just iterate over the subclass types.

docs/multi_process.md
```diff
@@ -64,7 +64,11 @@ The API {func}`jax.distributed.initialize` takes several arguments, namely:
     cluster will connect.
   * `num_processes`: the number of processes in the cluster
   * `process_id`: the ID number of this process, in the range `[0 ..
-  num_processes)`.
+    num_processes)`.
+  * `local_device_ids`: Restricts the visible devices of the current process to
+    ``local_device_ids``. Defaults to all local devices being visible to the
+    process except when processes are launched via Slurm on GPUs.
+    In that case, it will default to a single device per process.
```
skye, line 71: Let's actually remove the whole "Defaults to ... single deivce per process." section, since you explain that immediately below. (Sorry if I'm contradicting myself, I just looked at the rendered docs and thought it was a little awkward to read it twice in short succession.)

skye (changes requested):
Looks great! One more small round of changes and then let's get this in.

jax/_src/distributed.py
```diff
@@ -33,24 +33,18 @@ class State:
   def initialize(self,
                  coordinator_address: Optional[str] = None,
                  num_processes: Optional[int] = None,
-                 process_id: Optional[int] = None):
+                 process_id: Optional[int] = None,
+                 local_device_ids: Union[int, Sequence[int], None] = None):
```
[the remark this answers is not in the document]
nvcastet: oh, I did not know that was possible. Thanks for sharing.

jax/_src/clusters/cluster.py
```diff
@@ -0,0 +1,94 @@
...
+    cls._cluster_types.append(cls)
+
+  @classmethod
+  def auto_detect_unset_distributed_params(cls,
+                                           coordinator_address: Optional[str],
+                                           num_processes: Optional[int],
+                                           process_id: Optional[int],
+                                           local_device_ids: Optional[Union[int, Sequence[int]]]
+                                          ) -> Tuple[Optional[str], Optional[int], Optional[int],
+                                                     Optional[Union[int, Sequence[int]]]]:
```
skye, line 29:
```suggestion
                                           local_device_ids: Optional[Sequence[int]]
                                          ) -> Tuple[str, int, int, 
                                                     Optional[Sequence[int]]]:
```

Sorry I missed this earlier. These should definitely be set (except local device ids), and we normalize the local devices to a list in distributed.init right?

skye: Oh nevermind, they're None if we can't find an environment. I think the local_device_ids should always be a list and never an int though, for both the parameter and returned value.

nvcastet: correct, yes it should be none if we don't have an environment. Let me fix the `local_device_ids` type.

skye (approved):

skye:
Oh, you need to add the copyright boilerplate to the top of all new files. Copy/paste from any existing file and change the copyright year to 2022.

jax/_src/clusters/cluster.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```
yashk2810, line 1: @skye should this be NVIDIA corporation or Google LLC?

yashk2810: @hawkinsp

hawkinsp: The rules we follow here are:
https://opensource.google/documentation/reference/releasing/contributions#copyright

(I think we'd be very open to changing the copyright attribution to "The JAX authors" but we'd have to do that for the entire project.)

nvcastet: Sounds good, I just pushed a commit to follow your guideline. But yes if we don't want individual company copyright, it would make sense to switch to "The JAX authors" to match the legal reality of the files copyright.

skye (approved):

skye (approved):

skye (approved):

PR #12163, opened 2022-08-30
use dce_jaxpr in dispatch for pruning unused inputs/constants

mattjj:
TODO add trivial dispatch, or more general input forwarding, to Array path


mattjj:
We landed this in a separate PR!

PR #12172, opened 2022-08-31
Add markers to run multi node gpu tests separately

sudhakarsingh27:
Slurm launches separate processes that doesn't play well with other tests that themselves launch multiple processes (using subprocess). So, marking multi-node test to run it separately. For future multi-node tests, will be able to add this marker to run multi-node tests separately while they co-exist with other tests.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,6 @@
 [pytest]
+markers =
+    SlurmMultiNodeGpuTest: mark a test for Slurm multinode GPU nightly CI
 filterwarnings =
     error
     ignore:No GPU/TPU found, falling back to CPU.:UserWarning
--- a/tests/multiprocess_gpu_test.py
+++ b/tests/multiprocess_gpu_test.py
@@ -33,6 +33,11 @@ try:
 except ImportError:
   portpicker = None
 
+try:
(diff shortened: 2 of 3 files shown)


tests/multiprocess_gpu_test.py
```diff
@@ -148,6 +149,7 @@ def test_distributed_jax_cuda_visible_devices(self):
       self.assertEqual(proc.returncode, 0)
       self.assertEqual(out, f'{num_gpus_per_task},{num_gpus},[{num_gpus}.]')
 
+@pytest.mark.MultiNodeGpuTest
```
yashk2810, line 152: SlurmMultiNodeGpuTest?

sudhakarsingh27: Done

pytest.ini
```diff
@@ -1,4 +1,6 @@
 [pytest]
+markers =
+    SlurmMultiNodeGpuTest: mark a test for multinode GPU CI
```
yashk2810, line 3: mark a test for Slurm multinode GPU nightly CI.

yashk2810 (approved):

PR #12177, opened 2022-08-31
test the RNG key upgrade in one of our github CI runs

froystig:
I went with the "numpy-dispatch" configuration since, as @jakevdp pointed out, it seems to be our frequent choice for future-proofing.

cc #9263


jakevdp (approved):
Looks good!

PR #12178, opened 2022-08-31
[shape_poly] Add limited support for jnp.repeat with shape polymorphism

gnecula:
Supports only the case when the `repeats` parameter is a dimension polynomials
and the total_repeat_length is None.


gnecula:
@rxsang PTAL

rxsang (approved):

PR #12186, opened 2022-08-31
Throw error when 64-bit dtypes used incorrectly in pure_callback

sharadmv:


tests/python_callback_test.py
```diff
@@ -558,6 +558,20 @@ def f():
...
 
+  @jtu.skip_on_devices(*disabled_backends)
+  def test_callback_with_wrongly_specified_64_bit_dtype(self):
+    if config.jax_enable_x64:
+      raise unittest.SkipTest("Test only needed when 64-bit mode disabled.")
+
+    @jax.jit
+    def f():
+      return jax.pure_callback(lambda: np.float64(1.),
+                               core.ShapedArray((1,), np.float64))
```
jakevdp, line 569: Since the function returns a scalar:
```suggestion
                               core.ShapedArray((), np.float64)
```

jakevdp (approved):
Looks great!

PR #12205, opened 2022-09-02
[jax2tf] Refactor top-level jax2tf.convert

gnecula:
There are several goals for this refactoring:
  * improve the readability of the code: more helper functions, move big
    nested functions to top-level make make it obvious what are the
    data dependencies
  * try to be more systematic about naming: JAX entities end with _jax
    and TF entities with _tf. This is helpful because in several cases
    one function has to operate with both kinds of entities.
  * the main goal is to enable fixing the experimental_native_lowering
    for pjit. For that (future) work, we want to pass JAX callables
    to _interpret_fun_jax, rather than linear_util.WrappedFun. This will
    allow us to use fewer JAX core internals, and rely more on AOT APIs.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -202,26 +202,26 @@ def inside_call_tf():
     _thread_local_state.inside_call_tf = prev
 
 @partial(api_util.api_hook, tag="jax2tf_convert")
-def convert(fun: Callable,
+def convert(fun_jax: Callable,
             *,
             polymorphic_shapes=None,
             with_gradient=True,
             enable_xla=True,
             experimental_native_lowering="default") -> Callable:
-  """Transforms `fun` to be executed by TensorFlow.
+  """Lowers `fun_jax` into a function that uses only TensorFlow ops.
 
   See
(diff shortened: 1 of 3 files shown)


gnecula:
@rxsang PTAL

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -256,21 +256,20 @@ def convert(fun: Callable,
...
-      so the converter tries harder to use non-XLA TF ops to convert the
+      so the the lowering tries harder to use non-XLA TF ops to lower the
       function and aborts if this is not possible.
     experimental_native_lowering: DO NOT USE, for experimental purposes only.
       The value "default" defers to --jax2tf_default_experimental_native_lowering.
 
   Returns:
-    A version of `fun` that expects TfVals as arguments (or
+    A version of `fun_jax` that expects TfVals as arguments (or
     tuple/lists/dicts) thereof, and returns TfVals as outputs, and uses
```
marcvanzee, line 273:
```suggestion
    tuple/lists/dicts thereof), and returns TfVals as outputs, and uses
```

marcvanzee (approved):
Nice, very useful to explicitly name variables to either _jax or _tf.

rxsang (approved):

PR #12213, opened 2022-09-02
Add `jax.print_environment_info()` utility

jakevdp:
The idea here is that this gives a single API call that we can ask users to run when opening a bug or asking a question.

Example on Colab CPU:
```python
>>> import jax; jax.print_environment_info()
jax:    0.3.17
jaxlib: 0.3.15
numpy:  1.21.6
python: 3.7.13 (default, Apr 24 2022, 01:04:09)  [GCC 7.5.0]
jax.devices (1 total, 1 local): [CpuDevice(id=0)]
process_count: 1
```
Example on Colab GPU:
```python
>>> import jax; jax.print_environment_info()
jax:    0.3.17
jaxlib: 0.3.15
numpy:  1.21.6
python: 3.7.13 (default, Apr 24 2022, 01:04:09)  [GCC 7.5.0]
jax.devices (1 total, 1 local): [GpuDevice(id=0, process_index=0)]
process_count: 1

$ nvidia-smi
Fri Sep  2 21:26:17 2022       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 460.32.03    Driver Version: 460.32.03    CUDA Version: 11.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            Off  | 00000000:00:04.0 Off |                    0 |
| N/A   38C    P0    27W /  70W |    104MiB / 15109MiB |      4%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```
And if you have lots of hosts:
```python
$ XLA_FLAGS=--xla_force_host_platform_device_count=256 python -c "import jax; jax.print_environment_info()"
jax:    0.3.18
jaxlib: 0.3.15
numpy:  1.23.2
python: 3.8.2 (v3.8.2:7b3ab5921f, Feb 24 2020, 17:52:18)  [Clang 6.0 (clang-600.0.57)]
jax.devices (256 total, 256 local): [CpuDevice(id=0) CpuDevice(id=1) ... CpuDevice(id=254) CpuDevice(id=255)]
process_count: 1
```

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -45,7 +45,7 @@ import jax.numpy as jnp
 from jax import float0, jit, grad, device_put, jacfwd, jacrev, hessian
 from jax import core, lax
 from jax import custom_batching
-from jax._src import api, dtypes, dispatch
+from jax._src import api, dtypes, dispatch, lib
 from jax.core import Primitive
 from jax.errors import UnexpectedTracerError
 from jax.interpreters import ad
@@ -8997,6 +8997,20 @@ class CleanupTest(jtu.JaxTestCase):
       assert core.trace_state_clean()  # this is the hard one
     assert core.trace_state_clean()
 
+
+class EnvironmentInfoTest(jtu.JaxTestCase):
+  @parameterized.parameters([True, False])
+  def test_print_environment_info(self, return_string):
+    with jtu.capture_stdout() as stdout:
+      result = jax.print_environment_info(return_string=return_string)
+    if return_string:
+      self.assertEmpty(stdout())
+    else:
+      self.assertIsNone(result)
+      result = stdout()
+    assert f"jax:    {jax.__version__}" in result
+    assert f"jaxlib: {lib.version_str}" in result
+    assert f"numpy:  {np.__version__}" in result
 
 if __name__ == '__main__':
   absltest.main(testLoader=jtu.JaxTestLoader())
--- a/docs/jax.rst
+++ b/docs/jax.rst
@@ -109,3 +109,11 @@ Callbacks
 
     pure_callback
     debug.callback
+
+Miscellaneous
+-------------
+
+.. autosummary::
+  :toctree: _autosummary
+
+    print_environment_info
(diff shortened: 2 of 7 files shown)


soraros:
Maybe also platform name, `scipy`/`numpy`/... version and/or Nvidia driver version?

jakevdp:
I don't think we should mention this in the issue template until it's part of a release (or perhaps a few releases)

jakevdp:
Added numpy version and nvidia-smi when available.

soraros:
@jakevdp Ok, one very silly idea. Can we add an option like `markdown=True` to `jax.print_environment_info` so the generated information looks neat in a bug report while still being as complete as possible? Something like the following:

````
```
jax:    {...}
jaxlib: {...}
...
```

<details>

```
nvidia-smi:
  {...}
jax.config.a: {...}
jax.config.b: {...}
...
```

</details>
````

File: tests/api_test.py [hunk 1]
```diff
@@ -9013,5 +9014,19 @@ def test_call_wrapped_second_phase_cleanup(self):
     assert core.trace_state_clean()
 
 
+class EnvironmentInfoTest(jtu.JaxTestCase):
```
yashk2810, line 9017: Can you add a test where `XLA_FLAGS=--xla_force_host_platform_device_count=256` and then run this and check if the list of devices is infact truncated?

I don't see where you are truncating it.

yashk2810:
I would really like to see the process_count and local_device_count too! They are really helpful to debug multihost things instead of asking it everytime to users.

jakevdp:
Added local and process count as well, thanks!

yashk2810 (approved):

tests/api_test.py
jakevdp: I don't think you can change that flag value for a single test, can you?

jakevdp: I added an example of truncated output to the CL description.

froystig (approved):

PR #12231, opened 2022-09-06
Fix `Cmd.use_rawinput` type in debuggers

sobolevn:
Why is this an issue? In the example below:

```python
class Some:
   a = True
   b: bool = True
```

`a` and `b` have different semantics. Since `a` is initialized in a class body, tools like `mypy` assume it is a class attribute. Mypy also does inference its type to be `bool`, no annotation is needed.

But, `b` is different. It is annotated as an instance-level variable with a class-level default. (yeap, typing is a different world)

The same semantics `a = True` has can be specified by `a: ClassVar[bool] = True`, but I don't think it is worth it. Just removing type annotation is good enough.

Found in https://github.com/python/typeshed/pull/8692

### DIFF
--- a/jax/_src/debugger/cli_debugger.py
+++ b/jax/_src/debugger/cli_debugger.py
@@ -27,7 +27,7 @@ DebuggerFrame = debugger_core.DebuggerFrame
 class CliDebugger(cmd.Cmd):
   """A text-based debugger."""
   prompt = '(jdb) '
-  use_rawinput: bool = False
+  use_rawinput = False
 
   def __init__(self, frames: List[DebuggerFrame], thread_id,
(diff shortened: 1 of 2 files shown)


apaszke:
Do you have a link to the part of the Python typing docs that describes this behavior?

apaszke (commented):
The docs for `cmd` say:
> Instances of [Cmd](https://docs.python.org/3/library/cmd.html#cmd.Cmd) subclasses have some public instance variables:
> ...
> Cmd.use_rawinput
A flag, defaulting to true.

So those attributes do seem to be instance vars. Our type checkers don't seem to complain either, so I don't feel like this is a good change.

sharadmv:
I think this is now out-of-date due to https://github.com/google/jax/pull/14695.

PR #12232, opened 2022-09-06
Checkify: Remove some `err` <-not-> `pred` flipping.

LenaMartens:
Now all internal APIs deal in terms of errs (error thrown if True), and
only the check API takes a pred (error thrown if False).

This simplifies some of the internals, and will remove unnecessary ops.


mattjj (approved):
Yesssss so good.

PR #12239, opened 2022-09-06
Split State effect into Read/Write/Accum effects and tie them to Refs

sharadmv:
This change does two related, but somewhat orthogonal things.

1. We are adding in a "heap parameter" (see Haskell's ST monad and Dex's state effect) to our State effect. To do this, we are parameterizing the `StateEffect` via the `ShapedArrayRef` aval it happens to. This gives the type system more information about what effects are happening in a Jaxpr (we now have one effect per `Ref` we have, instead of one `StateEffect` for the whole jaxpr).
2. We are splitting up the `StateEffect` into a `Read/Write/AccumEffect`. The union of `Read/Write` is `State` as we commonly know it.

Why? 

1. Parameterizing the `StateEffect` with the `ShapedArrayRef` aval, i.e. a heap parameter, is needed for nesting stateful functions (like a nested for loop)_
2. Splitting up the state effect into read/write/accum this way enables us to perform optimizations more easily (e.g. hoisting loop invariant residuals) since we're exposing more information to transformations via the jaxpr type system. It also enables better lowerings (we can lower an `Accum` effect into an associative scan, for example).

Note that this also mirrors the breakdown of effect types in Dex as well.

Tracker: #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -51,7 +51,9 @@ T = TypeVar('T')
 class Ref(Generic[T]): pass
 Array = Any
 
-StateEffect = state.StateEffect
+ReadEffect = state.ReadEffect
+WriteEffect = state.WriteEffect
+AccumEffect = state.AccumEffect
 ShapedArrayRef = state.ShapedArrayRef
 ref_set = state.ref_set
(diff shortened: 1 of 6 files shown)


mattjj (approved):
Great description!

PR #12241, opened 2022-09-06
Add sound loop invariance detection for `for`

sharadmv:
Previously, we would incorrectly mark some residuals as loop invariant when they weren't. We assumed that loop-variance stems from interacting with the index argument to `body`, but it can actually stem from R/W `Ref`s as well. This PR adds this additional aspect to the fix point.

It also adds a test that fails w/o the change.

Tracker: #10982 
Stacked on: #12239


mattjj (approved):
Nice! (Plus recent discussion in GVC.)

PR #12251, opened 2022-09-07
[jax2tf] Refactor the experimental_native_lowering path in jax2tf

gnecula:
This is part of a suite of refactorings aimed towards supporting pjit by jax2tf experimental_native_lowering. The goal here is to remove many references to internal JAX core APIs, and instead use the AOT APIs: jax.jit(func_jax).lower(*args).

Only the experimental_native_lowering behavior should be affected.


gnecula:
@rxsang PTAL

marcvanzee (approved):

Issue #12049, opened 2022-08-22
Tracking Issue: JAX Type Annotations


jakevdp:
This is an issue to track progress on JAX Type Annotations.

For information, see the Type Annotation Roadmap, added in https://github.com/google/jax/pull/11859

jakevdp:
`jax.Array` is lacking a number of attributes that are defined on the array implementation. Here's a list (found by importing the contents of `basearray.pyi` and running `set(dir(jnp.arange(10)) - set(dir(basearray.Array))` and then removing properties and private attributes)
```
{
 'addressable_data',
 'block_until_ready',
 'broadcast',
 'broadcast_in_dim',
 'copy_to_host_async',
 'delete',
 'device',
 'device_buffer',
 'device_buffers',
 'devices',
 'flat',
 'global_shards',
 'is_deleted',
 'is_fully_addressable',
 'is_fully_replicated',
 'itemsize',
 'on_device_size_in_bytes',
 'split',
 'traceback',
 'unsafe_buffer_pointer'
}
```
A few of these are obvious omissions (`broadcast`, `broadcast_in_dim`, `itemsize`, `split`); I've added them in https://github.com/google/jax/pull/13967. `'flat'` is defined but raises a NotImplementedError on both arrays and tracers; the remaining are array-only attributes and methods that we should probably define on `jax.Array` and return an informative error in the case of tracers.

jakevdp:
Closing this, since the core work is complete.

PR #12256, opened 2022-09-07
jax.Array: support duck-typed isinstance checks

jakevdp:
Why? As part of the type promotion discussion in JAX (#11859 & #12049), @patrick-kidger and I have been discussing what API jax should support for both annotations and `isinstance` checks; we think the best approach would be to eventually make `jax.Array` unify all these purposes.

This is a simple enhancement to `jax.Array` that allows it to be used in `isinstance` checks in traced functions, similar to how `isinstance(obj, jnp.ndarray)` works currently.

@yashk2810, I'm curious whether you see any issues with adding a mechanism like this to the `Array` class? Do you think this idea of unifying type checking and instance checks around the eventual `jax.Array` object has merit? Do you have any hesitations here?

### DIFF
--- a/jax/experimental/array.py
+++ b/jax/experimental/array.py
@@ -105,7 +105,12 @@ def _reconstruct_array(fun, args, arr_state, aval_state):
   return jnp_value
 
 
-class Array:
+class ArrayMeta(type):
+  def __instancecheck__(cls, val):
+    return super().__instancecheck__(val) or isinstance(val, core.Tracer)
+
+
+class Array(metaclass=ArrayMeta):
   # TODO(yashkatariya): Add __slots__ here.
 
   def __init__(self, aval: core.ShapedArray, sharding: Sharding,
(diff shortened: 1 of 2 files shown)


yashk2810:
Array is in the process of being lowering to C++. How does this affect that? (given that you are adding a metaclass to Array).

Is it possible to keep the behavior of Array similar to DA in this case because DA is also in C++ and it doesn't have a metaclass on it right?

jakevdp:
`DeviceArray` does not behave this way, but we want `Array` to behave this way. If we lower `Array` to C++, we should add the same metaclass in C++.

yashk2810:
> we should add an equivalent metaclass in the C++ definition. Does that sound reasonable?

@hawkinsp or @cky9301 Is it okay if we add the metaclass to C++? I don't know if that is possible. I am fine with it existing in Python though.

jakevdp:
I've found a couple examples of overriding `__instancecheck__` within pybind11; one example is in torch: https://github.com/pytorch/pytorch/blob/31ef8ddb8c4467f5b8698ef1eb9bb8bab7056855/torch/csrc/tensor/python_tensor.cpp#L149-L177

jakevdp:
Replaced by https://github.com/google/jax/pull/12300

PR #12270, opened 2022-09-08
[jax2tf] Updates to the documentation

gnecula:
One change I made throughout is to use the verb lower instead of convert when refering to what jax2tf does. This is more accurate, because what is happening is that the lowering to HLO is replaced with a lowering to TF ops. Some users were confused by the use of the verb convert as suggesting some sort of source-to-source translation of JAX programs to TF, which sounds very complicated and risky.


jax/experimental/jax2tf/README.md
```diff
@@ -194,19 +197,20 @@ For examples of how to save a Flax model as a SavedModel see the
 
 ### Saved model and differentiation
 
-The converted code supports differentiation from TensorFlow. In order to
+The code lowered from JAX supports differentiation from TensorFlow. In order to
 ensure that the result of TensorFlow differentiation is identical to the
-one that JAX differentiation would produce, the jax2tf converter will
-annotate the converter function with a ``tf.custom_gradient`` that,
+one that JAX differentiation would produce, the jax2tf decorator will
```
marcvanzee, line 202: I am not sure if "decorator" is accurate here (and in other places), since jax2tf can be used as a decorator but also by using the function directly. Also there seems to be no decoration happening in the examples in the previous section.

marcvanzee (approved):
Looks good to me, however the function is still called `jax2tf.convert`, and it seems unlikely we will be able to change that. Also I notice you now use "lower", "stage", and "decorate" instead of just "convert". I am not entirely sure if this will lead to less confusion.

jax/experimental/jax2tf/README.md
gnecula: I have removed the references to `decorator`.

PR #12296, opened 2022-09-08
Bump the minimum jaxlib version to 0.3.15.

hawkinsp:


yashk2810:
I wanted to bump this to the jaxlib version I am going to release. Would that be okay in the future?

yashk2810 (approved):

hawkinsp:
> I wanted to bump this to the jaxlib version I am going to release. Would that be okay in the future?

The minimum must be a released version. I am comfortable bumping to 0.3.15 since it has been out for a while and seems like a good version. I would not bump to a newly released version for a few days at least.

PR #12297, opened 2022-09-08
add _device attribute to PRNGKeyArray so that computation follows key placement

mattjj:
unrelated: remove some redundant hasattr + try / except AttributeError

### DIFF
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -15,6 +15,7 @@
 
 import abc
 from functools import partial
+import operator as op
 from typing import Any, Callable, Hashable, Iterator, NamedTuple, Sequence
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


jax/_src/prng.py
```diff
@@ -169,6 +169,8 @@ def ndim(self):
   def dtype(self):
     return KeyTy(self.impl)
 
+  _device = property(op.attrgetter('_base_array._device'))
```
yashk2810, line 172: Drive-by: For jax.Array you will also need to add `_committed` attributed for it to pass on the Array path.

mattjj: Thanks!

froystig (approved):

yashk2810 (approved):

PR #12300, opened 2022-09-08
jax._src.typing: add basic types and use them in lax.py

jakevdp:
Part of #12049; this implements some of the simpler types discussed in #11859

The question of what to do with `Array` is yet to be resolved, but this PR creates the skeleton into which we can insert the eventual solution.

For now, I'm keeping these as private APIs, in `jax._src.typing`. Once we're certain we're happy with them, we can plan to export them in `jax.typing`.

### DIFF
--- a/jax/_src/typing.py
+++ b/jax/_src/typing.py
@@ -1,0 +1,63 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+`jax._src.typing`: JAX type annotations
+---------------------------------------
+
+This submodule is a work in progress; when we finalize the contents here, it will be
+exported at `jax.typing`. Until then, the contents here should be considered unstable
+and may change without notice.
+
+To see the proposal that led to the development of these tools, see
+https://github.com/google/jax/pull/11859/.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Sequence, Union
+from typing_extensions import Protocol
+import numpy as np
+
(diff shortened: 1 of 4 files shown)


jax/_src/typing.py
```diff
@@ -0,0 +1,60 @@
...
+Shape = Sequence[DimSize]
+
+# Array is a type annotation for JAX arrays and tracers.
+# For now we set it to Any; in the future this will be more restrictive
+# (see https://github.com/google/jax/pull/11859)
+# TODO(jakevdp): make this conform to the JEP 12049 plan.
+Array = Any
+
+# ArrayLike is a Union of all objects that can be implicitly converted to a JAX array.
+ArrayLike = Union[
```
froystig, line 55: Does `jax.random.KeyArray` go in here too? If so, then several of the `lax.py` annotations in this PR are unsound, e.g. you cannot `lax.sin` a key array. But if not, then several of the `lax.py` annotations are incomplete, e.g. you can `lax.expand_dims` a key array.

froystig (commented):
Does it make sense to add unit tests? For example, can/should we test that various types that we expect to check out as `HasDtypeAttribute` indeed do?

Otherwise, I think we still need to figure out how opaque-dtype arrays (like key arrays, soon `BInt` arrays too) fit in to the picture – see inline comment.

jax/_src/typing.py
jakevdp: Do we expect `KeyArray` objects to eventually be valid in all lax routines? If not, then I think we should use `Union` annotations on a case-by-case basis for the functions that support `KeyArray`.

froystig: What I wrote above is both the current and eventual intended situation: `KeyArray` is valid for some lax (and lax-numpy) operations, but not others. One cannot `sin` or `+` key arrays, but it is possible to `expand_dims`, `squeeze`, etc. A union sounds good, at least for correcting the annotations in this PR. Such a union will likely expand at some point to also account for `BInt` arrays.

> `Array` here is meant to be a lowest-common-denominator type for all JAX public APIs that accept array-like objects

Even if a union resolves this in code, it seems that this statement could be refined in light of the above (especially as it appears in documentation). When you say "[...] APIs that accept array-like objects," it seems that you do not also mean key arrays (and later `Bint` arrays), but a reader might not realize this caveat.

jakevdp: I updated the comments to add more clarity on this question - PTAL!

froystig (approved):

Issue #12016, opened 2022-08-19
Tracking bug: Integration of Array in JAX.


yashk2810:
- [x] Add tests for _committed attribute in Array.

- [x] Figure out how to resolve the local imports like (from jax.experimental import pjit) that are in dispatch.py.

- [x] Make testArrayCopy in lax_numpy_test.py work with Array. There is a TODO in that test with this bug number.

- [x] Make xmap work properly with SingleDeviceSharding Arrays or just make the behavior of xmap similar to what it was with DA and SDA.
- [x] Implement no copy indexing similar to SDA i.e. __getitem__ of SDA
- [x] Make sure that non-committed Arrays on SingleDevice (for eg. jnp.arange(10) with a mesh decorator are sharded properly in pjit and xmap.
- [x] Add a similar check to non-XLACompatibleShardings that exists in the constructor of Array (there is a TODO there).

yashk2810:
4. Make xmap work properly with SingleDeviceSharding Arrays or just make the behavior of xmap similar to what it was with DA and SDA.

yashk2810:
5. Implement no copy indexing similar to SDA.

mjsML:
I think this would be much better using the GitHub projects / boards feature :)

yashk2810:
6. Make sure that non-committed Arrays on SingleDevice (for eg. jnp.arange(10) with a mesh decorator are sharded properly in pjit and xmap.

yashk2810:
7. Make Array work properly with `view = jnp.array(buf, copy=False)` so that `buf.unsafe_buffer_pointer() == view.unsafe_buffer_pointer())`

yashk2810:
I have moved other things to this board: https://github.com/orgs/google/projects/25/views/1 but I am keeping this bug alive for the things I am tracking here.

yashk2810:
Because the boards can only be created at the project IIUC. If there is a different solution, I can use that :)

PR #12311, opened 2022-09-09
jax.Array: support fast path for lax.transpose & lax.squeeze

jakevdp:
As part of this change, I created a helper function so that the logic of type checking is in a single location. Eventually we can replace this helper function with appropriate isinstance() checks using the APIs described in #11859.

This work is part of #12016 & #12049.


yashk2810:
Wow, what a coincidence! I added support for transpose and squeeze here too: cl/473267301

yashk2810:
Let's get your PR in and I can sync my CL to that.

yashk2810 (approved):

jakevdp:
Sounds good, thanks!

PR #12314, opened 2022-09-09
Implement fixed point routine jax.scipy.optimize.fixed_point

JanLuca:
In the PR I included a general implementation of the `jax.scipy.optimize.fixed_point` function with similar API to the one of the original `scipy` package. The implementation has a custom VJP defined with the results from the implicit function theorem.

### DIFF
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -65,6 +65,7 @@ jax.scipy.optimize
 .. autosummary::
   :toctree: _autosummary
 
+   fixed_point
    minimize
    OptimizeResults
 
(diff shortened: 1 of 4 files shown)


JanLuca:
@Maintainers: I do not really see the reason why the doc generation is failing. In the log I find no error related to this PR?

froystig:
@JanLuca – what do you think of contributing this to JAXopt instead, as @mblondel suggests?

Generally speaking, when a component of `jax.scipy.optimize` involves a substantial implementation that would significantly increase the maintenance load of the jax core, we try to see if other libraries like JAXopt might take it on instead. As an existing example of this, we're planning to shed `jax.scipy.optimize.minimize` because JAXopt already covers L-BFGS.

PR #12324, opened 2022-09-11
[jax2tf] Implement jax2tf(pjit) for experimental_native_lowering

gnecula:
This implementation is for the case jax2tf.convert(pjit(f_jax)), that is, the `pjit` appears at the top-level of the function to be lowered.


gnecula:
I want to merge this PR to enable collaborators downstream to experiment with it. 

@rxsang

Issue #3009, opened 2020-05-08
Using JAX with Pandas 


harshit-2115:
Hey, 
Is there any way to use JAX to speed up and reduce memory-usage of data processing pipelines in pandas ?

jakevdp:
Hi - Jax does not have any built-in support yet for a pandas-like API, but it is an interesting idea. I'll leave this as an open issue!

dynamicwebpaige:
In the meantime, you can also [read pandas dataframes](https://www.tensorflow.org/tutorials/load_data/pandas_dataframe) directly as a `tf.data.Dataset` (which would be compatible with JAX). 

Thank you for the feature request!

99snowleopards:
@jakevdp @mattjj  @dynamicwebpaige - any updates on this enhancement? thanks,

TheAustinator:
This is a great idea!

jakevdp:
I've been playing with this idea a bit; I pushed my current work-in-progress to https://github.com/google/jax/pull/12329

vitalNohj:
Very interested! Would be a major feature to bolster JAX's already incredible advantage.

adendek:
Is there any progress on that?  Do you need open-source contributors to help?

openspaceranger:
I would like jax to replace pandas in my stack. pandas is just too slow. I know they are working to address this, but it would still be nice to have a competitor to pandas

jakevdp:
I'm not actively working on this, and I don't know of anybody who is.

AhmedThahir:
Any updates?

westurner:
- cudf.pandas calls cudf where possible and pandas where cudf has not yet implemented, and handles the dataflows between the CPU and GPU.
  - https://github.com/rapidsai/cudf/blob/branch-25.04/python/cudf/cudf/pandas/_wrappers/pandas.py#L238
  - cudf.pandas extended and re-ran the h20.ai dataframe benchmarks:
    https://docs.rapids.ai/api/cudf/stable/cudf_pandas/benchmarks/
- pandas dtype_backend="pyarrow"
  - The pyarrow backend improves the performance of pandas.
- https://github.com/jax-ml/jax/issues/19792
  - JAX and pyarrow both depend upon libgrpc
- https://github.com/NVIDIA/warp
  - warp works in JAX pipelines
- taichi support Vulkan, for CUDA-less GPU access:
[...]

PR #12329, opened 2022-09-12
pandax: experimental JAX pandas wrapper

jakevdp:
`pandax`! This is an experimental pandas-like API built on top of JAX. Very experimental work in progress. Related to #3009


jakevdp:
I'm going to close this. It's a fun experiment, but we're unlikely to add anything like this to jax core.

PR #12334, opened 2022-09-12
ensure_index: raise better error for traced inputs

### LINKED ISSUE #12331: `jnp.squeeze` will fail unexpectedly with `jit` compilation
`jnp.squeeze` will fail unexpectedly with `jit` compilation but direct invocation will succeed. For example,

```py
import jax

def fn(inp):
    axis = jax.numpy.array(1, dtype=jax.numpy.int32)
    return jax.numpy.squeeze(inp, axis)

mykey = jax.random.PRNGKey(0)
inp = jax.random.uniform(mykey, [3, 1], jax.numpy.float32, minval=0, maxval=1)

print(fn(inp))
# [0.9653214  0.31468165 0.63302994]

jax.jit(fn)(inp)
# TypeError: iteration over a 0-d array
```

I think the input is valid since the dim-`1` has shape 1.

## Version

jax 0.3.17

jakevdp:
Fixes #12331

Quick test:
```python
import jax

def f(x):
  return jax.numpy.squeeze(x, jax.numpy.array(1))

jax.jit(f)(jax.numpy.arange(5))
```
Previous error:
```
TypeError: iteration over a 0-d array
```
Error after this PR:
```
jax._src.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
expected a static index or sequence of indices.
The error occurred while tracing the function f at tmp.py:3 for jit. 
```

### DIFF
--- a/jax/_src/api_util.py
+++ b/jax/_src/api_util.py
@@ -37,6 +37,7 @@ map = safe_map
 
 def _ensure_index(x: Any) -> Union[int, Tuple[int, ...]]:
   """Ensure x is either an index or a tuple of indices."""
+  x = core.concrete_or_error(None, x, "expected a static index or sequence of indices.")
   try:
     return operator.index(x)
   except TypeError:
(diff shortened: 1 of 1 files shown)


sharadmv (approved):
Nice!

PR #12342, opened 2022-09-13
Add typing_test.py

jakevdp:
Basic tests of `jax._src.typing` objects; part of #12049

Related to #12300

### DIFF
--- a/tests/typing_test.py
+++ b/tests/typing_test.py
@@ -1,0 +1,117 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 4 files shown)


tests/typing_test.py
```diff
@@ -0,0 +1,112 @@
...
+from jax._src import typing
+from jax import lax
+import jax.numpy as jnp
+
+from absl.testing import absltest
+import numpy as np
+
+
+def dtypelike_to_dtype(x: typing.DtypeLike) -> typing.Dtype:
+  return np.dtype(x)
```
froystig, line 33: This isn't a valid conversion on opaque dtypes (as in comments on #12300, again the example is key arrays). Do we want to support those? And should we make them part of `DtypeLike`? If so an expression along the lines of:
```suggestion
  return x if jax.core.is_opaque_dtype(x) else np.dtype(x)
```
would work. If not, nevermind—it's clearly not needed in this current test file—but I'm curious whether we'll need to resolve this question more broadly, and how we would do so.

froystig (approved):

tests/typing_test.py
jakevdp: `DtypeLike` is meant to annotate anything that can be turned into a JAX-compatible dtype using `np.dtype`, so I think this is fine.

If we're introducing other things that duck-type as dtypes in the future, then we should support them via a different type annotation.

froystig: Sounds good, and works in analogy to discussion in #12300 regarding `ArrayLike`.

jakevdp: Cool, added some comments to make this more clear.

PR #12361, opened 2022-09-14
Implement unrolling for for_loop

sharadmv:
Tracker: #10982

### DIFF
--- a/jax/_src/lax/control_flow/for_loop.py
+++ b/jax/_src/lax/control_flow/for_loop.py
@@ -109,7 +109,7 @@ def val_to_ref_aval(x) -> ShapedArrayRef:
 
 def for_loop(nsteps: Union[int, Sequence[int]],
              body: Callable[[Array, Ref[S]], None], init_state: S,
-             *, reverse: bool = False) -> S:
+             *, reverse: bool = False, unroll: int = 1) -> S:
   """A for-loop combinator that allows read/write semantics in the loop body.
 
   `for_loop` is a higher-order function that enables writing loops that can be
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Nice!

mattjj (changes requested):
Actually, I'm not sure if this is computing the right thing... following up in chat.

We want to generate a `while_loop` which has trip count `nsteps // unroll`, with a body that looks like `unroll` applications of the body function. Then we want `nsteps % unroll` applications of the body function outside that `while_loop`.

I'm not sure but I think this code might be generating a `while_loop` with `nsteps - unroll` iterations?

jax/_src/lax/control_flow/for_loop.py
```diff
@@ -255,33 +263,50 @@ def _for_abstract_eval(*avals, jaxpr, **__):
...
+def _for_impl(*args, jaxpr, nsteps, reverse, which_linear, unroll):
   del which_linear
   discharged_jaxpr, consts = discharge_state(jaxpr, ())
+  def body(i, state):
+    i_ = nsteps - i - 1 if reverse else i
+    return core.eval_jaxpr(discharged_jaxpr, consts, i_, *state)
+  return _for_impl_unrolled(body, nsteps, unroll, *args)
+
+def _for_impl_unrolled(body, nsteps, unroll, *args):
+  _, remainder = divmod(nsteps, unroll)
```
mattjj, line 285: Seems like we should be using the quotient...

mattjj (approved):
Nevermind, Sharad explained it. It was right of course. The index is being incremented more than I realized! Thanks for explaining.

PR #12366, opened 2022-09-14
jax.random: warn on unsupported dtypes

### LINKED ISSUE #12364: jax.random.randint silently ignores dtype of int64 instead of generating warning
### Description

```python
import jax.numpy as jnp
import jax.random as jr
k = jr.PRNGKey(0)
a = jr.randint(k, (10,), minval=10, maxval=20, dtype=jnp.int64)  # no warning is printed!
assert a.dtype == jnp.int32
a = a.astype(jnp.int64)  # warning is printed
```

3.17

CPU

### Additional System Info

macos

jakevdp:
Fixes #12364

Explanation for the default changes: `jnp.float_` and `jnp.int_` are currently equal to `float64` and `int64`, respectively, and are cannonicalized upon use. We can't use them as default arguments here because they trigger the unsupported dtype warning when X64 is False... on the other hand, the scalar types `int` and `float` do not trigger the unsupported dtype warning, and are canonicalized to the appropriate default type, so they have the desired behavior.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -237,7 +237,7 @@ def _check_shape(name, shape: Union[Sequence[int], NamedShape], *param_shapes):
 
 def uniform(key: KeyArray,
             shape: Union[Sequence[int], NamedShape] = (),
-            dtype: DTypeLikeFloat = dtypes.float_,
+            dtype: DTypeLikeFloat = float,
             minval: RealArray = 0.,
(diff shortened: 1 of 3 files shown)


mattjj:
Is this consistent with our "remove x64 flag" plans? Or e.g. should we just switch the defaults to be 32bit?

PR #12368, opened 2022-09-14
add paper link about grad-relu-at-zero

mattjj:
Thanks @jekbradbury for pointing out this paper!

### DIFF
--- a/jax/_src/nn/functions.py
+++ b/jax/_src/nn/functions.py
@@ -45,10 +45,20 @@ def relu(x: Array) -> Array:
   .. math::
     \mathrm{relu}(x) = \max(x, 0)
 
+  except under differentiation, we take:
+
+  .. math::
+    \nabla \mathrm{relu}(0) = 0
+
+  For more information see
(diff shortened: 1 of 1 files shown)


hawkinsp:
Not rendering well in docs: https://jax--12368.org.readthedocs.build/en/12368/_autosummary/jax.nn.relu.html?highlight=relu#jax.nn.relu

mattjj:
Thanks for the catch, Peter, as always! I think it's improved now.

hawkinsp (approved):

carlosgmartin:
@mattjj Can't the implementation of [nn.relu](https://github.com/jax-ml/jax/blob/e4ca4f5a573b531c412aac87d104a52d0edf42e4/jax/_src/nn/functions.py#L52) be slightly simplified by letting
```python3
def relu(x):
  return jnp.where(x > 0, x, 0)
```

This would preserve the property that `jax.grad(relu)(0.) == 0.` without having to define a custom JVP.

PR #12370, opened 2022-09-14
[typing] overloaded type declaration for variadic lax.sort

jakevdp:
Part of #12049

Whether the return value is a tuple or an array depends on the input; leaving as a straight union causes type errors for patterns like:
```python
x = jnp.arange(10)
lax.sort(x).shape
```
because `Tuple` doesn't have a `shape` attribute.


jakevdp:
This was Approved as a subset of #12421 so I'm going to merge this first

PR #12389, opened 2022-09-16
Checkify: support batched while.

### LINKED ISSUE #11905: BUG: checkify of a vmap-ed while loop fails with an obscure error
### Description

Consider the following code:
```python
def test_checkify_with_batched_while():
    def func_to_check_with_loop(n, v):
        def while_condition(loop_state):
            counter, _ = loop_state
            return counter > 0

        def loop_body(loop_state):
            counter, value = loop_state
            checkify.check(value > 0, 'value must be positive')
            return counter - 1, value - 1

        _, result = jax.lax.while_loop(while_condition, loop_body, (n, v))
        return result

    batched_func = jax.vmap(func_to_check_with_loop)
    checked_f = checkify.checkify(batched_func)

    err, val = checked_f(jnp.asarray([1, 2, 3]), jnp.asarray([5, 2, 4]))
    err.throw()
    assert jnp.all(val == jnp.asarray([4, 0, 1]))
```

The code above fails when calling `checked_f` with the following error (truncated):
```
def _pred_bcast_select_mhlo(
        pred_aval: core.ShapedArray, pred: ir.Value, xs: Sequence[ir.Value],
        ys: Sequence[ir.Value], x_y_aval: core.AbstractValue) -> Sequence[ir.Value]:
      if x_y_aval is core.abstract_token:
        x, = xs
        y, = ys
        return [mhlo.AfterAllOp(mlir.aval_to_ir_type(x_y_aval), [x, y]).result]
      else:
        assert isinstance(x_y_aval, core.ShapedArray), x_y_aval
        x, = xs
        y, = ys
        assert x.type == y.type, (x.type, y.type)
>       assert (pred_aval.shape == x_y_aval.shape[:len(pred_aval.shape)]), (
                pred_aval.shape, x_y_aval)
E       jax._src.traceback_util.UnfilteredStackTrace: AssertionError: ((3,), ShapedArray(bool[]))
```

If one modifies the code by either removing `checkify`, or making a non-batched version by removing `vmap`, everything works as expected.

Can you please confirm that this is indeed a bug in JAX and, if so, provide a workaround?

0.3.16

CPU

### Additional System Info

Mac

LenaMartens:
Fixes #11905 

Some facts:
- A batched `while_loop` expects all its inputs and outputs to be batched values
- The `checkify` error rule adds (scalar) error values to the in- and output signatures of the `checked_cond` and `checked_body`
- The `checkify` internals expect all error values to be scalar (otherwise we get shape errors)

So to support batched `while_loops`:
- batch the error inputs to `cond` and `body` and rewrite `cond` and `body` to accept batched error values
- inside `cond` and `body`, unbatch the error value before passing it to actual `checked_cond` and `checked_body` (which only accept scalars)
- batch the error output before returning from `cond` and `body` (to adhere to the batched `while_loop` contract)
- unbatch the error output before returning from the `while_loop` error rule

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -662,18 +662,54 @@ def ignore_error_output_jaxpr(jaxpr):
   new_jaxpr = jaxpr.replace(outvars=jaxpr.outvars[3:])
   return core.ClosedJaxpr(new_jaxpr, consts)
 
+def batch_error(err, code, payload, batch_shape):
+  err = jnp.broadcast_to(err, batch_shape)
+  code = jnp.broadcast_to(code, batch_shape)
+  payload = jnp.broadcast_to(payload, batch_shape+(3,))
+  return err, code, payload
+
+def unbatch_error(err, code, payload):
+  err = err.ravel()[0]
+  code = code.ravel()[0]
+  payload = payload.reshape(-1, 3)[0]
+  return err, code, payload
+
+def trivial_batched_jaxpr(jaxpr, batch_shape, batched_err):
+  fun = core.jaxpr_as_fun(jaxpr)
+
+  def g(err, code, payload, *a):
+    err_args = unbatch_error(err, code, payload)
+    err, code, payload, *out = fun(*err_args, *a)
+    err, code, payload = batch_error(err, code, payload, batch_shape)
+    return (err, code, payload, *out)
+
+  error_avals = map(lambda x: core.raise_to_shaped(core.get_aval(x)), batched_err)
+  new_jaxpr, _, literals_out = pe.trace_to_jaxpr_dynamic(
+      lu.wrap_init(g), [*error_avals, *jaxpr.in_avals[3:]])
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Woo!

jax/_src/checkify.py
```diff
@@ -662,18 +662,49 @@ def ignore_error_output_jaxpr(jaxpr):
...
+  err = jnp.broadcast_to(err, batch_shape)
+  code = jnp.broadcast_to(code, batch_shape)
+  payload = jnp.broadcast_to(payload, batch_shape+(3,))
+  return err, code, payload
+
+def trivial_batched_jaxpr(jaxpr, batch_shape, batched_err):
+  fun = core.jaxpr_as_fun(jaxpr)
+
+  def g(err, code, payload, *a):
+    err = err.ravel()[0]
```
mattjj, line 675: Maybe abstract this into an `unbatch_error` function, which is the inverse of `batch_error` iiuc?

PR #12396, opened 2022-09-16
Jax custom call

zhangqiaorjc:


jaxlib/jax_custom_call.h
```diff
@@ -0,0 +1,23 @@
...
+extern "C" {
+#endif
+
+struct JaxCustomCallStatus;
+
+struct JaxCustomCallApi {
+  JaxCustomCallStatus* status;
+};
+
+void JaxCustomCallStatusSetSuccess(JaxCustomCallStatus* status);
```
hawkinsp, line 14: BTW, linkage may be hard here. My suggestion to Sharad was that we pass function pointers for all API entry points a custom call might need (i.e., like a c++ class implemented in C).

zhangqiaorjc: Updated to use runtime linkage instead of link-time linkage a la Numpy import_array() style.

zhangqiaorjc:
We are importing this to explore some things. @hawkinsp hold off on reviews. Thanks!

jlu-spins:
Hi! This looks really useful. Is anyone still committed to making this happen?

zhangqiaorjc:
@jlu-spins yes, we are making an improved version with better C++ typed argument support. But it creates a new dependency on some XLA component that's still WIP. We are actively working on this, and will hopefully have a version in a few weeks.

In the meanwhile, https://github.com/dfm/extending-jax is still a good guide for doing JAX custom call at head.

jlu-spins:
@zhangqiaorjc Thank you!

MoFHeka:
How create a custom class instance with several call method and its own stateful resource. Something like [Extending TorchScript with Custom C++ Classes](https://pytorch.org/tutorials/advanced/torch_script_custom_classes.html#using-the-c-class-from-python-and-torchscript)

dfm:
I'm going to close this PR since this work has been superseded by XLA's FFI API which is documented in https://github.com/google/jax/pull/22095.

@MoFHeka — Please feel free to open a new discussion topic with more details about your specific use case if it would be useful to continue the conversation.

PR #12398, opened 2022-09-16
Add the ability to pass `is_leaf` to `tree_reduce` and add a docstring.

### LINKED ISSUE #12377: Add the ability to pass an `is_leaf` argument to `tree_reduce`.
Many other functions in `jax.tree_util` allow the user to provide an `is_leaf` argument which is subsequently passed to `tree_flatten`.

It would be nice to be able to pass such a function to `tree_reduce` to give something like:
```python
def tree_reduce(function: Callable[[T, Any], T],
                tree: Any,
                initializer: Any = no_initializer,
                is_leaf: Optional[Callable[[Any], bool]] = None) -> T:
  if initializer is no_initializer:
    return functools.reduce(function, tree_leaves(tree, is_leaf))
  else:
    return functools.reduce(function, tree_leaves(tree, is_leaf), initializer)
```

My use case looks a bit like:
```python
from jax import numpy as jnp
from typing import NamedTuple
from jax.tree_utils import tree_map, tree_reduce

class Box(NamedTuple):
    a: jnp.ndarray # (N,) array.
    b: jnp.array # (N,N) array.

def sum_boxes(box1, box2):
    return tree_map(lambda x,y: x + y, box1, box2)

dim = 4
box1 = Box(a=jnp.ones(dim), b=jnp.ones((dim,dim)))
box2 = Box(a=jnp.ones(dim), b=jnp.ones((dim,dim)))
box3 = Box(a=jnp.ones(dim), b=jnp.ones((dim,dim)))

# Some PyTree containing instances of Box at various nodes.
box_tree = {"a":box1, "b":box2, "c":box3}

# I would like to be able to do this:
summed_boxes = tree_reduce(sum_boxes, box_tree, is_leaf=lambda l: isinstance(l, Box))
```
(I'm aware there are equivalent ways to achieve this but, given it exists, the `tree_reduce` syntax seems particularly clear).

In general it seems like `tree_reduce` hasn't received much love (not having a docstring), is there a particular reason for this? i.e. is there some reason why this type of operation should be avoided if possible?

gileshd:
Closes #12377.

As discussed in #12377 unlike similar functions in `jax.tree_util` the function `tree_reduce` does not provide the option to pass in an `is_leaf` callable.

This PR adds that functionality to `tree_reduce` using the approach suggested in #12377.

The PR also takes a shot at adding a docstring to `tree_reduce` taking inspiration from the `tree_map` docstring and `functools.reduce` [docs](https://docs.python.org/3/library/functools.html#functools.reduce).

### DIFF
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -262,11 +262,43 @@ def tree_reduce(function: Callable[[T, Any], T],
 
 def tree_reduce(function: Callable[[T, Any], T],
                 tree: Any,
-                initializer: Any = no_initializer) -> T:
+                initializer: Any = no_initializer,
+                is_leaf: Optional[Callable[[Any], bool]] = None) -> T:
+  """Apply function cumulatively over the leaves of a pytree, so as to reduce
+   the collection of leaves to a single value.
+
+  Args:
+    function: a function that takes two arguments, to be applied cumulatively
+      to the leaves of ``tree`` in the order returned by ``tree_leaves``.
+    tree: a pytree containing the leaves which will be passed to ``function``
+      for reduction.
+    initializer: an optional argument to be placed before the leaves
+     of ``tree`` in the calculation.
+    is_leaf: an optionally specified function that will be called at each
+      flattening step. It should return a boolean, which indicates whether
+      the flattening should traverse the current object, or if it should be
+      stopped immediately, with the whole subtree being treated as a leaf.
+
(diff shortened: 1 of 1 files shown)


gileshd:
It looks like this has now been resolved elsewhere 👍 

https://github.com/google/jax/blob/20255dce8418f820866075108af82093dab72360/jax/_src/tree_util.py#L322-L325

PR #12412, opened 2022-09-19
Ignore protobuf's DeprecationWarning during test

wookayin:
During import of tensorflow (via jax2tf), a bunch of DeprecationWarnings
from google.protobuf (<4.0, e.g. 3.19.x) would be raised:

    DeprecationWarning: Call to deprecated create function FileDescriptor().
    DeprecationWarning: Call to deprecated create function FieldDescriptor().

Pytest should not fail due to this DeprecationWarning, so we can
explicitly ignore such protobuf-related warnings.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -19,5 +19,7 @@ filterwarnings =
     # numpy uses distutils which is deprecated
     ignore:The distutils.* is deprecated.*:DeprecationWarning
     ignore:`sharded_jit` is deprecated. Please use `pjit` instead.*:DeprecationWarning
+    # protobuf
+    ignore:Call to deprecated create function .*Descriptor:DeprecationWarning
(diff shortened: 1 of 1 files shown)


wookayin:
It might be an off-topic, but should I also ignore deprecation warnings from Pillow (via keras via tensorflow)?

```
    # PIL
    ignore:is deprecated and will be removed in Pillow 10:DeprecationWarning
```

hawkinsp:
I suspect this is stale now. Sorry for not looking at it sooner!

wookayin:
Has it been fixed? If not, I can make a rebase and update the PR.

hawkinsp:
I'm not sure, but there was at least one new TF release since the PR. If you still have this problem, feel free to update and reopen it.

PR #12351, opened 2022-09-14
Jax array test

jakevdp:
Part of #12049

PR #12421, opened 2022-09-19
Add initial implementation of jax.Array base class

jakevdp:
Part of #12049

The idea here is that `jax.Array` will eventually be defined in C++ and used as a true base class of both `core.Tracer` and `jax.experimental.array.Array`, which will be renamed to `ArrayImpl` or something similar before being moved to `jax._src.array`.

~Still need to add some tests similar to the ones in https://github.com/google/jax/pull/12351 to ensure that this implementation has the desired properties.~ *Done*


jax/_src/array.py
```diff
@@ -12,12 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 import abc
```
yashk2810, line 16: Are we sure we want to rename this file right now?

I was planning to do it while moving the entire thing in experimental/array.py here and sharding.py too.

yashk2810: Currently, now there are 2 types of imports which is confusing. Can we just do the rename when the concrete class moves too?

jakevdp: Do you mean we should define `jax.Array` within `jax._src.numpy.ndarray`?

yashk2810: Yes, as a stopgap. If you want to.. maybe move the entire Array implementation here with this rename?

But I guess that will touch a lot more files :)

And then you can add a big comment, telling what to move to jax._src.array when you/I do the move.

jakevdp: I think I'll put it in `jax._src.typing` for now instead.

jax/__init__.py
```diff
@@ -35,6 +35,8 @@
 from jax import config as _config_module
 del _config_module
 
+from jax._src.basearray import Array as Array
```
yashk2810, line 38: Do we want to export this right now?

jax/_src/lax/lax.py
```diff
@@ -1330,11 +1338,11 @@ def full_like(x: Array, fill_value: ArrayLike, dtype: Optional[DTypeLike] = None
   # probably in the form of a primitive like `val = match_sharding_p.bind(x, val)`
   # (so it works in staged-out code as well as 'eager' code). Related to
   # equi-sharding.
-  if (config.jax_array and hasattr(x, 'sharding') and
-      not dispatch.is_single_device_sharding(x.sharding) and
-      not isinstance(x.sharding, sharding.PmapSharding)):
-    return array.make_array_from_callback(
-        fill_shape, x.sharding, lambda idx: val[idx])  # type: ignore[arg-type]
+  if config.jax_array and hasattr(x, 'sharding'):
+    sharding = x.sharding  # type: ignore[union-attr]
```
yashk2810, line 1342: Why this change? Do I have to use this pattern everywhere?

jax/__init__.py
jakevdp: As I understood it, the goal of this was to expose a working prototype of `jax.Array` so we could migrate users to it before landing the new experimental `Array`, rather than having to change their code twice. So yes, we'll need to export this now.

yashk2810: Ahh right! SGTM!

jax/_src/lax/lax.py
jakevdp: Because `hasattr(x, 'sharding')` is not enough for mypy to understand that `x` is now of a type that has an attribute `sharding`, so `x.sharding` results in a type check error. We could avoid this if every Array-like object declared a `sharding` attribute (e.g. `Tracer` would have to define sharding), or by using an `isinstance` check in place of the `hasattr` check (i.e. avoiding duck-typing) but I did it this way in order to not change the semantics of the function, which I admit I don't totally understand.

yashk2810: Hmm.. okay. We will eventually end up defining a sharding attributed on Tracer too I think but let's deal with that later.

yashk2810 (approved):

jakevdp:
I had been making some effort to land independent parts of this separately in order to make the change less extensive (e.g. https://github.com/google/jax/pull/12370). Do you think that's worth doing here? I worry a change like this is big enough that it might cause unexpected issues (e.g. does having `Tracer` inherit from a class without `__slots__` cause performance degredation?) So thought it would be worth landing in small bits. What do you think?

yashk2810:
> does having Tracer inherit from a class without __slots__ cause performance degredation?) 

I think the best way to know this is to write a quick benchmark and see if it regresses. The one good thing I like about this CL is that if it does break something, we can roll it back as 1 CL rather than rolling back parts of it. So I think it's fine to keep this as is.

If you are worried about perf, https://github.com/google/jax/blob/main/benchmarks/api_benchmark.py has been really handy and you can run this locally too!

jakevdp:
I'm more worried about the things I haven't thought of... and changes like #12370 are truly independent – they're correct whether or not we land the rest of this, so there wouldn't have to be rolled back together. There are a few other similar changes in here (e.g. the changes to `lax_numpy.py`)

yashk2810:
Ohh I see. Yeah splitting up SGTM.

jax/experimental/array.py
```diff
@@ -95,7 +95,7 @@ def _reconstruct_array(fun, args, arr_state, aval_state):
 
 
 @pxla.use_cpp_class(xc.Array if xc._version >= 92 else None)
-class Array:
+class Array(basearray.Array):
```
hawkinsp, line 98: Note that if the C++ class is in use (i.e., you have a new enough `jaxlib`) then subtyping won't hold for `Array` either.

hawkinsp (approved):

jax/experimental/array.py
jakevdp: Ah, right, so maybe for now we do need `basearray.Array.register(Array)`, which will take care of runtime instance checks. At static type check time, though, I don't think mypy/pytype is smart enough to figure out that the decorator has replaced the class implementation, so I believe it will still work as long as this is the method used to fall-back. What do you think?

yashk2810: Maybe just do the register below even for Python?

jax/experimental/array.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from __future__ import annotations
+from email.mime import base
```
yashk2810, line 16: What's `email.mime`?

jakevdp: Gah, I have to set up vscode to not try to automatically add imports...

I suspect I typed `base` and hit tab, so it helpfully found something named `base` to import for me

jax/experimental/array.py
jakevdp: I've done it conditionally based on the jaxlib version, let me know if that looks OK.

jakevdp: No, if we remove the inheritance then mypy/pytype will no longer recognize `Array` as a subclass of `jax.Array`

yashk2810: Ohh I see. Okay, LGTM.

mrahtz:
(Just wanted to say a huge THANK YOU for this Jake - it's something that a lot of us at DeepMind have been wanting for a long time!)

PR #12422, opened 2022-09-19
docs: avoid deprecated matplotlib axis creation

jakevdp:
Fixes readthedocs error caused by a new matplotlib deprecation.


sharadmv (approved):
Thanks!

PR #12437, opened 2022-09-21
Add multi host pjit tests

sudhakarsingh27:
Added multi host pjit tests to the existing test setup of 2 nodes and 1 process per GPU

### DIFF
--- a/tests/multiprocess_gpu_test.py
+++ b/tests/multiprocess_gpu_test.py
@@ -17,16 +17,23 @@ import subprocess
 import sys
 import threading
 import unittest
+import functools
 
 from absl.testing import absltest
 from absl.testing import parameterized
+import numpy as np
 
 import jax
+from jax import experimental
 from jax.config import config
 from jax._src import distributed
 import jax.numpy as jnp
 from jax._src.lib import xla_extension_version
 from jax._src import test_util as jtu
+from jax._src import util
+from jax.experimental import global_device_array
+from jax.experimental import maps
+from jax.experimental import pjit
 
 try:
   import portpicker
@@ -39,7 +46,6 @@ except ImportError:
   pytest = None
 
 config.parse_flags_with_absl()
-
 
 @unittest.skipIf(not portpicker, "Test requires portpicker")
 class DistributedTest(jtu.JaxTestCase):
@@ -170,6 +176,49 @@ class SlurmMultiNodeGpuTest(jtu.JaxTestCase):
   if pytest is not None:
     pytestmark = pytest.mark.SlurmMultiNodeGpuTest
 
+  def sorted_devices(self):
+    devices = sorted(jax.devices(), key=lambda d: (d.id, d.host_id))
+    if len(devices) != 16:
+      raise unittest.SkipTest(
+          "Test assumes that it runs on 16 devices (2 nodes)")
+    return devices
+
+  def create_2d_non_contiguous_mesh(self):
+    devices = self.sorted_devices()
+    device_mesh = np.array([[devices[0], devices[2]],
+                            [devices[4], devices[6]],
+                            [devices[1], devices[3]],
+                            [devices[5], devices[7]],
+                            [devices[8], devices[10]],
+                            [devices[12], devices[14]],
+                            [devices[9], devices[11]],
+                            [devices[13], devices[15]]])
+    # The mesh looks like this (the integers are process index):
+    #   0 2
+    #   4 6
+    #   1 3
+    #   5 7
+    #   8 10
+    #   12 14
+    #   9 11
+    #   13 15
+    assert [d.id for d in device_mesh.flat
+           ] == [0, 2, 4, 6, 1, 3, 5, 7, 8, 10, 12, 14, 9, 11, 13, 15]
+    return maps.Mesh(device_mesh, ("x", "y"))
+
+  def setUp(self):
+    super().setUp()
+    self.xmap_spmd_lowering_enabled = jax.config.experimental_xmap_spmd_lowering
+    jax.config.update("experimental_xmap_spmd_lowering", True)
+    self.gda_enabled = jax.config.jax_parallel_functions_output_gda
+    jax.config.update('jax_parallel_functions_output_gda', True)
+
+  def tearDown(self):
+    jax.config.update("experimental_xmap_spmd_lowering",
+                      self.xmap_spmd_lowering_enabled)
+    jax.config.update('jax_parallel_functions_output_gda', self.gda_enabled)
+    super().tearDown()
+
   def test_gpu_multi_node_initialize_and_psum(self):
 
     # Hookup the ENV vars expected to be set already in the SLURM environment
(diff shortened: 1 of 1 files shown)


yashk2810:
I don't think there's any need to repeat the tests here. Why not just subclass the original pjit tests and run them in Slurm?

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +46,80 @@
...
+  assert (jax.device_count() == int(num_tasks) * jax.local_device_count())
+
+def sorted_devices():
+  devices = sorted(jax.devices(), key=lambda d: (d.id, d.host_id))
+  if len(devices) != 16:
+    raise unittest.SkipTest("Test assumes that it runs on 16 devices (2 nodes)")
+  return devices
+
+@contextlib.contextmanager
+def use_default_mesh():
```
yashk2810, line 94: Don't use this. Use jtu.create_global_mesh. This function is an old one and there's no need to it to exist right now.

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +46,80 @@
...
+  jax_multiprocess_init()
```
yashk2810, line 95: Why do you need to init this everytime?

Can you just init it once in `def setUp(self)`?

tests/multiprocess_gpu_test.py
```diff
@@ -209,5 +289,275 @@ def test_gpu_multi_node_initialize_and_psum(self):
...
+      global_input_data = np.arange(
+          util.prod(global_input_shape)).reshape(global_input_shape)
+
+      def cb(index):
+        return global_input_data[index]
+
+      gda_obj = global_device_array.GlobalDeviceArray.from_callback(
+          global_input_shape, global_mesh, mesh_axes, cb)
+
+      jax.config.update('jax_parallel_functions_output_gda', True)
```
yashk2810, line 306: Use the context manager. This is unsafe as you never set it to False.

Or better yet, set this once in `setUp()` and set it to the original value in `tearDown()`

yashk2810:
Also add a TODO to remove the GDA tests and move them to Array.

tests/multiprocess_gpu_test.py
sudhakarsingh27: What if one of the test crashes and it messes up the multihost setup? Will the other tests continue to run? 

Right now, I'm running each test in its own subprocess using `--forked` flag in pytest which basically let's each test run unaffected if one of the other tests fails.

yashk2810: Ohh I see. ok sgtm.

tests/multiprocess_gpu_test.py
sudhakarsingh27: ah, removed it now

tests/multiprocess_gpu_test.py
sudhakarsingh27: used context manager

sudhakarsingh27:
addressed the concerns

tests/multiprocess_gpu_test.py
sudhakarsingh27: cool

sudhakarsingh27:
fixed lint and doc build issues now

tests/multiprocess_gpu_test.py
```diff
@@ -209,5 +281,251 @@ def test_gpu_multi_node_initialize_and_psum(self):
...
+    gda1 = global_device_array.GlobalDeviceArray.from_callback(
+        global_input_shape, global_mesh, mesh_axes1, cb)
+    mesh_axes2 = experimental.PartitionSpec("x")
+    gda2 = global_device_array.GlobalDeviceArray.from_callback(
+        global_input_shape, global_mesh, mesh_axes2, cb)
+    mesh_axes3 = experimental.PartitionSpec(("x", "y"))
+    gda3 = global_device_array.GlobalDeviceArray.from_callback(
+        global_input_shape, global_mesh, mesh_axes3, cb)
+
+    with parallel_functions_output_gda(True):
```
yashk2810, line 306: Please set this up in setUp once!

There's no need to use a context manager since all the tests are running with GDA

sudhakarsingh27: Since all the tests run in their own subprocess, there can't be a `setUp` for all the tests. (Do you mean doing this at the start/end of each test though?)

yashk2810: What do you mean by there can't be a setUp for all the tests?

The setUp method exists on all test classes. Can you try prototyping it please?

```
class Slurm...():
  
  def setUp(self):
    super().setUp()
    self.gda_enabled = jax.config.parallel_functions_output_gda
    jax.config.update('parallel_functions_output_gda', True)

  def tearDown(self):
    jax.config.update('parallel_functions_output_gda', self.gda_enabled)
    super().tearDown()
```

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +46,72 @@
...
 
+config.update("experimental_xmap_spmd_lowering", True)
 config.parse_flags_with_absl()
 
+X_SIZE = 4
+Y_SIZE = 2
+Z_SIZE = 2
+ALL_AXES = ("x", "y", "z")
+
+def jax_multiprocess_init():
```
yashk2810, line 57: Please don't name these functions as the generic thing. This is for slurm so name it `slurm_multiprocess_init`.

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +46,72 @@
...
+
+  print(f"coord addr:port : {coordinator_address}:{port}\nTotal tasks: "
+        f"{num_tasks}\ntask id: {taskid}\nlocal id: {localid}")
+
+  assert (coordinator_address is None or num_tasks is None or
+          taskid is None) is False
+
+  jax.config.update("jax_cuda_visible_devices", localid)
+
+  jax.distributed.initialize(
```
yashk2810, line 78: Now that Nicholas's PR is submitted (or will be soon), do you need all this setup? Can you just use `jax.distributed.initialize()`?

tests/multiprocess_gpu_test.py
sudhakarsingh27: Ah that's clearer (doing setUp/tearDown for every test and not once for all the tests). I can try prototyping this

tests/multiprocess_gpu_test.py
sudhakarsingh27: that's reasonable, will rename

tests/multiprocess_gpu_test.py
sudhakarsingh27: I think that's a good idea, will use that

sudhakarsingh27:
Done with making changes as requested

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +45,37 @@
...
+config.update("experimental_xmap_spmd_lowering", True)
 config.parse_flags_with_absl()
 
+def sorted_devices():
+  devices = sorted(jax.devices(), key=lambda d: (d.id, d.host_id))
+  if len(devices) != 16:
+    raise unittest.SkipTest("Test assumes that it runs on 16 devices (2 nodes)")
+  return devices
+
+def create_2d_non_contiguous_mesh():
```
yashk2810, line 57: Make these methods of Slurm test class. No need for this to exist outside of that class.

tests/multiprocess_gpu_test.py
```diff
@@ -38,8 +45,37 @@
 except ImportError:
   pytest = None
 
+config.update("experimental_xmap_spmd_lowering", True)
 config.parse_flags_with_absl()
 
+def sorted_devices():
```
yashk2810, line 51: Same, make this a method of Slurm test.

yashk2810:
Few minor comments, after that LGTM

sudhakarsingh27:
ping after fixing nits

yashk2810 (approved):

PR #12443, opened 2022-09-21
Fix mlir capi dll is missing chlo and stablehlo symbols

cloudhan:
In the past, only `mlir*` symbols are needed, but now, `chlo*` and `stablehlo*` are also needed now. Otherwise, there will be linker errors for targets `_chlo` and `_stablehlo` defined  [here](https://github.com/google/jax/blob/6183727acc64b0a35fc3d5842fa05cb9dee47346/jaxlib/mlir/_mlir_libs/BUILD.bazel).

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -96,7 +96,7 @@ build:windows --linkopt=/OPT:REF
 build:windows --host_linkopt=/OPT:REF
 build:windows --linkopt=/OPT:ICF
 build:windows --host_linkopt=/OPT:ICF
-build:windows --experimental_strict_action_env=true
+build:windows --incompatible_strict_action_env=true
 
 build:linux --config=posix
 build:linux --copt=-Wno-unknown-warning-option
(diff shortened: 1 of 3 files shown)


cloudhan:
@hawkinsp Could you take a look?

cloudhan:
NOTE: This only fix the branch it based.

On latest main, `external/org_tensorflow/tensorflow/core/tpu/tpu_initializer_helper.cc(18): fatal error C1083: Cannot open include file: 'dirent.h': No such file or directory`, but I have no desire to fix it. I think it should go if you update tensorflow version, maybe.

hawkinsp:
Thanks!

The `tpu_initializer_helper` should not even be linked in if TPU is disabled (e.g., on Windows). We should be able to fix that.

It looks like it is being pulled in via the new PJRT C API work:

```
$ python build/build.py --configure_only
...
TPU enabled: no
Remote TPU enabled: no
...
$ bazel cquery 'somepath(//build:build_wheel,@org_tensorflow//tensorflow/core/tpu:tpu_initializer_helper)'
....
//build:build_wheel (e92e97f)
//jaxlib:jaxlib (e92e97f)
//jaxlib:xla_extension (e92e97f)
@org_tensorflow//tensorflow/compiler/xla/python:xla_extension.so (e92e97f)
@org_tensorflow//tensorflow/compiler/xla/pjrt:pjrt_c_api_client (e92e97f)
@org_tensorflow//tensorflow/core/tpu:tpu_initializer_helper (e92e97f)
...
```

hawkinsp (approved):

PR #12444, opened 2022-09-21
Raise cond index into tracing context in case of effects

### LINKED ISSUE #12385: Problem mixing checkify.check and jax.lax.switch
### Description

I'm trying to do a switch over functions that have a checkify check inside. It goes somewhat like this:
```python
def make_branch(i):
  def branch():
    result = jnp.full((1,), i)
    checkify.check(jnp.sum(result) > 0, 'Failed!')
    return result
  return branch

index = jnp.full((), 0)
result = jax.lax.switch(index, [make_branch(i) for i in range(3)])
```
This, unfortunately, fails with `Cannot abstractly evaluate a checkify.check which was not functionalized`, presumably because `switch` compiles its branches.

I've tried transforming branches with `checkify`:
```python
def make_branch(i):
  @checkify.checkify
  def branch():
    result = jnp.full((1,), i)
    checkify.check(jnp.sum(result) > 0, 'Failed!')
    return result
  return branch

index = jnp.full((), 0)
err, result = jax.lax.switch(index, [make_branch(i) for i in range(3)])
checkify.check_error(err)
```
This approach, unfortunately, fails with another error:
```
branch 0 and 1 outputs must have same type structure, got PyTreeDef((CustomNode(Error[((4, 'Failed! (check failed at <ipython-input-14-15495f6829f6>:6 (branch))'),)], [*, *, *]), *)) and PyTreeDef((CustomNode(Error[((5, 'Failed! (check failed at <ipython-input-14-15495f6829f6>:6 (branch))'),)], [*, *, *]), *)).
```
For some reason type signatures for branches are different where error is envolved. That seems like an unexpected behavior that prohibits using checks with conditional constructs. Is there a way around it?

jax v0.3.17, jaxlib v0.3.15

GPU

### Additional System Info

Google Colab

LenaMartens:
So even if the cond is not data dependent at all, it's included in the dynamic trace.

Fixes #12385. 

But .. is it safe to do this? Not sure!

(also adds a missing checkify cond test)

### DIFF
--- a/jax/_src/lax/control_flow/conditionals.py
+++ b/jax/_src/lax/control_flow/conditionals.py
@@ -139,6 +139,10 @@ def switch(index, branches: Sequence[Callable], *operands,
   if disallowed_effects:
     raise NotImplementedError(
         f'Effects not supported in `switch`: {disallowed_effects}')
+  if joined_effects:
+    # Raise index in case of effects to allow data-dependence-based discharging
+    # of those effects (even if they don't have an explicit data dependence).
+    index = core.raise_as_much_as_possible(index)
 
   linear = (False,) * (len(consts) + len(ops))
   out = cond_p.bind(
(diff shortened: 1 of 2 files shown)


LenaMartens:
as discussed offline, only raise in case of effectful branches.

mattjj (approved):
LGTM. I think it's always safe to raise as much as possible.

Could leave a note here saying "We raise as much as possible in case of effects to allow data-dependence-based discharging of those effects in more cases."

PR #12451, opened 2022-09-21
[typing] add class-level declarations of Array members.

jakevdp:
This fixes some pytype errors associated with the changes in #12421. Part of #12049

### DIFF
--- a/jax/experimental/array.py
+++ b/jax/experimental/array.py
@@ -96,7 +96,18 @@ def _reconstruct_array(fun, args, arr_state, aval_state):
 
 @pxla.use_cpp_class(xc.Array if xc._version >= 92 else None)
 class Array:
+  """Experimental unified Array type.
+
+  This Python implementation will eventually be replaced by a C++ implementation.
+  """
   # TODO(yashkatariya): Add __slots__ here.
+
+  aval: core.ShapedArray
+  _sharding: Sharding
+  _arrays: List[DeviceArray]
+  _committed: bool
+  _skip_checks: bool
+  _npy_value: Optional[np.ndarray]
 
   @pxla.use_cpp_method
   def __init__(self, aval: core.ShapedArray, sharding: Sharding,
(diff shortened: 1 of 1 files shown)


jax/experimental/array.py
```diff
@@ -98,6 +98,13 @@ def _reconstruct_array(fun, args, arr_state, aval_state):
 class Array:
   # TODO(yashkatariya): Add __slots__ here.
 
+  aval: core.ShapedArray
```
yashk2810, line 101: Given that you are adding this, mind adding __slots__ too?

yashk2810 (approved):

jax/experimental/array.py
jakevdp: Sure, I'll give that a try here.

jax/experimental/array.py
```diff
@@ -96,7 +96,18 @@ def _reconstruct_array(fun, args, arr_state, aval_state):
 
 @pxla.use_cpp_class(xc.Array if xc._version >= 92 else None)
 class Array:
-  # TODO(yashkatariya): Add __slots__ here.
+  """Experimental unified Array type.
+
+  This Python implementation will eventually be replaced by a C++ implementation.
+  """
+  __slots__ = ['_arrays', '_committed', '_npy_value', '_sharding', 'aval']
```
yashk2810, line 103: Missing `_skip_checks`?

jax/experimental/array.py
jakevdp: Hm, turns out we can't use `__slots__` because the `pxla.maybe_cached_property` relies on the class having a `__dict__`. I suppose we could add `"__dict__"` to `__slots__`, but that would kind of defeat the purpose 😀

I'll revert the `__slots__` change for now, and leave the TODO in place.

jax/experimental/array.py
jakevdp: reverting; see https://github.com/google/jax/pull/12451#discussion_r976927974

jax/experimental/array.py
yashk2810: Sounds good! Thanks for trying!

PR #12455, opened 2022-09-22
add unused-argument-pruning and dce to pjit

mattjj:


yashk2810:
This is already handled in lower_sharding_computation. Why do we need to handle it in pjit's dispatch path?

yashk2810:
https://github.com/google/jax/blob/main/jax/interpreters/pxla.py#L2725-L2731 is the place where we DCE. What's the advantage of doing it in dispatch?

yashk2810:
AFAIK, this is how DCE happens in jit too. `jit` never tries to DCE right at the start (feel free to correct me if that's not the case).

keep_unused is always set to True for pjit though: https://github.com/google/jax/blob/main/jax/experimental/pjit.py#L964 due to some failures that we encountered in deepmind and our own pjit tests IIRC.

jax/experimental/pjit.py
```diff
@@ -339,22 +339,32 @@ def infer_params(*args, _global_avals=False, **kwargs):
...
+        in_shardings=tuple(canonicalized_in_shardings_flat),
+        out_shardings=tuple(canonicalized_out_shardings_flat),
         resource_env=resource_env,
-        donated_invars=donated_invars,
+        donated_invars=tuple(donated_invars),
         name=getattr(flat_fun, '__name__', '<unnamed function>'),
-        in_positional_semantics=in_positional_semantics,
+        in_positional_semantics=tuple(in_positional_semantics),
         out_positional_semantics=out_positional_semantics)
     return (args_flat, local_in_avals, params, in_tree, out_tree(),
```
yashk2810, line 369: local_in_avals is returned here. You need to prune that too.

jax/experimental/pjit.py
```diff
@@ -339,22 +339,32 @@ def infer_params(*args, _global_avals=False, **kwargs):
...
-        in_shardings=canonicalized_in_shardings_flat,
-        out_shardings=canonicalized_out_shardings_flat,
+        in_shardings=tuple(canonicalized_in_shardings_flat),
+        out_shardings=tuple(canonicalized_out_shardings_flat),
         resource_env=resource_env,
-        donated_invars=donated_invars,
+        donated_invars=tuple(donated_invars),
         name=getattr(flat_fun, '__name__', '<unnamed function>'),
-        in_positional_semantics=in_positional_semantics,
+        in_positional_semantics=tuple(in_positional_semantics),
```
yashk2810, line 367: Also in_tree is also returned. Is the pruning required for in_tree too?

mattjj: Yes and no... we only prune after flattening, but things like `PjitCompiled.input_shardings` seem to assume that the executable arguments are not pruned relative to the in_tree, so how does that work with the existing pruning (not the pruning in this PR)?

mattjj:
Please note this is a draft PR, not marked ready for review! But I probably shouldn't have pre-added you as a reviewer as that was confusing.

> This is already handled in lower_sharding_computation. Why do we need to handle it in pjit's dispatch path?

Actually, we were operating under some faulty hypotheses about observed behavior, thinking that the DCE wasn't happening.

An advantage of this approach over doing it in `lower_sharding_computation` is that we do it early, so any passes on the jaxpr don't need to deal with dead code.

That said, I don't think that advantage alone is worth the upside.

> AFAIK, this is how DCE happens in jit too. jit never tries to DCE right at the start (feel free to correct me if that's not the case).

Well, `pjit` forms a jaxpr up front before any transformations (it's an "initial-style higher-order primitive"), whereas `jit` only forms a jaxpr after transformations. So if we wanted to do DCE as soon as possible, i.e. as soon as a jaxpr is available, we would do it in `pjit` as in this PR, whereas with `jit` we could only do it in the impl rule.

PR #12476, opened 2022-09-22
[array] full_like: only match sharding if shape==None

jakevdp:
Also ported some of the changes required for #12421

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1311,7 +1311,7 @@ def expand_dims(array: ArrayLike, dimensions: Sequence[int]) -> Array:
 
 ### convenience wrappers around traceables
 
-def full_like(x: Array, fill_value: ArrayLike, dtype: Optional[DTypeLike] = None,
+def full_like(x: ArrayLike, fill_value: ArrayLike, dtype: Optional[DTypeLike] = None,
               shape: Optional[Shape] = None) -> Array:
   """Create a full array like np.full based on the example array `x`.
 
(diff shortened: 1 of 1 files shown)


jax/_src/lax/lax.py
```diff
@@ -1338,11 +1339,12 @@ def full_like(x: Array, fill_value: ArrayLike, dtype: Optional[DTypeLike] = None
   # probably in the form of a primitive like `val = match_sharding_p.bind(x, val)`
   # (so it works in staged-out code as well as 'eager' code). Related to
   # equi-sharding.
-  if (config.jax_array and hasattr(x, 'sharding') and
-      not dispatch.is_single_device_sharding(x.sharding) and
-      not isinstance(x.sharding, sharding.PmapSharding)):
-    return array.make_array_from_callback(
-        fill_shape, x.sharding, lambda idx: val[idx])  # type: ignore[arg-type]
+  if config.jax_array and shape is None and hasattr(x, 'sharding'):
+    sharding = x.sharding  # type: ignore[union-attr]
```
yashk2810, line 1343: I think you can call this `xs = x.sharding` and still import sharding. But I am fine either way. 

Its just if I remove PmapSharding from array, then this will break.

yashk2810 (approved):

jax/_src/lax/lax.py
jakevdp: yeah, changed it to import `PmapSharding` from `sharding`

PR #12478, opened 2022-09-23
Update debugging docs to have sharding visualization

sharadmv:


docs/jax.debug.rst
```diff
@@ -6,14 +6,28 @@ jax.debug package
...
+----------------------------
+
+Functions that enable inspecting and visualizing array shardings inside (and outside)
+staged functions.
+
+.. autosummary::
+  :toctree: _autosummary
+
+  inspect_array_sharding
+  visualize_array_sharding
```
mattjj, line 32: Why not just expose `visualize_sharding`?

sharadmv: It can't be used inside of `pjit` whereas `visualize_array_sharding` can be.

mattjj: Nevermind, I guessed wrong about what these things do.

mattjj (approved):

PR #12481, opened 2022-09-23
Allow transpose axes to be negative to match (undocumented) NumPy behavior

kho:


hawkinsp (approved):

jakevdp:
One comment here: `jax.lax` is often deliberately more strict than `jax.numpy`. I think requiring positive indices for `jax.lax.transpose` is consistent with the overall philosophy of the package. `jax.numpy.transpose` is what we expect to have a similar API to numpy.

hawkinsp (changes requested):
I think Jake is right: the change should be in the numpy layer.

kho:
Done. PTAL.

I was following `lax.squeeze` and `lax.expand_dims` which used `canonicalize_axis`. Now I see that the majority of `canonicalize_axis` calls indeed happen in `lax_numpy.py`.

jakevdp (approved):
Looks great, thanks!

PR #12497, opened 2022-09-23
fix bug in djax type signature inference logic

mattjj:


jax/interpreters/partial_eval.py
```diff
@@ -2156,24 +2159,23 @@ def _collect_implicit(
...
-
   for explicit_idx, (x, spec) in enumerate(zip(args, specs)):
     for i, name in spec.items():
       if name not in idxs and id(x.shape[i]) not in explicit_tracers:
-        idxs[name] = next(counter)
+        idxs[name] = DBIdx(next(counter))
         implicit_types.append(raise_to_shaped(get_aval(x.shape[i])))
     if isinstance(x, Tracer):
-      explicit_tracers[id(x)] = explicit_idx
+      explicit_tracers.setdefault(id(x), explicit_idx)  # use the first
```
sharadmv, line 2171: Should we add a test for this as well?

sharadmv (approved):

jax/interpreters/partial_eval.py
mattjj: That's it, I'm removing you as coauthor.

PR #12532, opened 2022-09-27
prod & sum: promote integers to default type rather than always 64

### LINKED ISSUE #10813: [x64] Reductions are too eager to promote to 64-bit
```python

In [2]: jnp.arange(4, dtype='int8').sum()
Out[2]: DeviceArray(6, dtype=int64)
```

The root cause is this line: https://github.com/google/jax/blob/d849f495193c69e8c133d0c6a424ba80d72938cb/jax/_src/numpy/reductions.py#L81

It's not an issue for floating-point types, because numpy does not promote those in most cases

jakevdp:
~This is a backward-incompatible change; previously for integer arrays x, x.sum() and x.prod() would promote to the widest available integer type, following the behavior of the NumPy equivalents. After this change, input dtypes are preserved. To recover the previous behavior, use the optional keyword promote_integers=True~

This is now a much safer change which promotes to the default dtype rather than to 64-bit dtypes. Currently the default is set by the `JAX_DEFAULT_DTYPE_BITS` flag to 64-bit (which is reduced to 32-bit unless `JAX_ENABLE_X64=true`); the net result is that this change should not be user-visible unless the user is changing the default bit config.

Followup to #12382; fixes #10813

### DIFF
--- a/jax/_src/numpy/reductions.py
+++ b/jax/_src/numpy/reductions.py
@@ -86,17 +86,21 @@ def _reduction(a, name, np_fun, op, init_val, has_identity=True,
     if not _all(core.greater_equal_dim(shape[d], 1) for d in pos_dims):
       raise ValueError(f"zero-size array to reduction operation {name} which has no identity")
 
-  result_dtype = dtypes.canonicalize_dtype(dtype or dtypes.dtype(a))
+  result_dtype = dtype or dtypes.dtype(a)
 
-  # promote_integers=True matches NumPy's behavior for sum() and prod(), which promotes
-  # all int-like inputs to the widest available dtype.
   if dtype is None and promote_integers:
+    # Note: NumPy always promotes to 64-bit; jax instead promotes to the
+    # default dtype as defined by dtypes.int_ or dtypes.uint.
     if dtypes.issubdtype(result_dtype, np.bool_):
-      result_dtype = dtypes.canonicalize_dtype(np.int64)
+      result_dtype = dtypes.int_
     elif dtypes.issubdtype(result_dtype, np.unsignedinteger):
-      result_dtype = dtypes.canonicalize_dtype(np.uint64)
+      if np.iinfo(result_dtype).bits < np.iinfo(dtypes.uint).bits:
+        result_dtype = dtypes.uint
     elif dtypes.issubdtype(result_dtype, np.integer):
-      result_dtype = dtypes.canonicalize_dtype(np.int64)
+      if np.iinfo(result_dtype).bits < np.iinfo(dtypes.int_).bits:
+        result_dtype = dtypes.int_
+
+  result_dtype = dtypes.canonicalize_dtype(result_dtype)
 
   if upcast_f16_for_computation and dtypes.issubdtype(result_dtype, np.inexact):
     computation_dtype = _upcast_f16(result_dtype)


nouiz:
Why do you want to do that change?
This will make it easy to have wrong value for int8 by default if I understand it correctly.

jakevdp:
This was discussed elsewhere, but it's part of the x64 deprecation (#8178). Silently producing 64-bit values is problematic on accelerators. Yes, it's true that `x.sum()` can overflow for `x.dtype = int8`, but that's also true of e.g. `x * 2`, in which neither numpy nor JAX will promote to a wider type, and the wraparound will happen silently. The argument is that `sum` and `prod` should similarly not promote unless it is explicitly requested.

nouiz:
If you go that route, you will have silent overflow. We have hit that in the past and so we moved to upcast by default.
Why not limit the upcast to int32 instead if the issue is only 64 bits ints?
Can you link to those dicussion?

jakevdp:
I think silent overflow is the correct behavior here, as it matches the behavior for operations like `x + y`, `2 * x`, and other operations on narrow-width integers. As in those cases, if you want to produce an output of a different type than the inputs, you should explicitly cast.

nouiz:
I think this would be a big mistake. I'll write tomorrow why in details. Can you wait to read my reasoning before deciding to merge or not?

jakevdp:
Sure, I'm waiting on your response here. I assume the issue you're concerned with is that when you sum an array of narrow integers, it's far more likely to overflow than when you do an operation like `x ** 2`, so implicity type promotion is worth doing in the case of `sum` and `prod`. I can assure you we've weighed that concern in the discussions that led to proposing this change, and still feel it's a worthwhile tradeoff to make the silent introduction of wide dtypes a rarer thing. Are there other concerns you have?

jakevdp:
Sure, I'll wait for your response here.

nouiz:
The goal to have better casting rules then NumPy given the fact that accelerators aren't as fast on int64 then on int32 is a good idea.
Having a better default is also great.
 
My objection is with the change for all reduction (prod, sum, others) not to do any upcast.
 
What is bad about this in two lines:
	• `jnp.ones(257, dtype='int8').sum() == 1` will be true.
	• `np.allclose(jnp.ones((32,16,512,512), dtype='int8').sum(axis=-1), 0) `will be true.
	 
	256 elements is way too small. It can too easily cause *silent* wrong value.
	 
In more details:
	• By default, we should get good results. Requiring small work to get even faster result is reasonable. Needing deep investigation that only ninja JAX programmer can do to get valid results isn't reasonable.
	• There is tools to help investigate performance issues, but none for those silent wrong values.
	• This goes again the philosophy of JAX: to be very user-friendly and to be ready to loose some speed for that.
	• This goes again the Zen of Python: 
		○ Explicit is better than implicit.
		○ Errors should never pass silently.
	• This goes again what user would expect: correct and fast enough by default.
	• This goes again what NumPy and *all other DL framework* to my knowledge.
		○ So porting code from any other frameworks will be way more complicated. They would just get different results and do not have any hints why. And there isn't any tool to help them investigate this.
	• Comparing sum() to add (prod to mul) is missing a very important aspect, you do one operation many times instead of just one. This is a good reason not to reuse the same principle.
	• This can break many models. Only documenting this in the release notes won't make it for a silent breaking changes. Why? Do all users know all the internal details of all the models they use? Do all users read all the release notes before installing a new version? Many users copy or clone a repo and do not know all the detail, so they will be hit by this.
		○ If we go that route, it would need deprecation with *clear warning* when people would be affected *for sometimes*.
	• There are middle grounds between this extreme decision and the upcast to int64 decision. I'm sure we can find a better one than the current extreme direction.

jakevdp:
Understood, I've weighed all of those considerations when thinking about this change.

One question: would you advocate for `np.arange(20, dtype='int8) ** 2` to also upcast? That silent truncation of values has always been present in JAX (and in numpy) and I'm not aware of any complaints about the behavior.

On the other hand, I've heard from several users surprised by the silent up-cast of `sum`.

jakevdp:
Also please realize that this change is not being proposed in a vacuum: we are revisiting this specifically in order to be able to deprecate the `jax_enable_x64` flag as painlessly as possible. That's been a HUGE issue for many, many users, so keep in mind we weigh the downsides of minor changes in corner case dtypes like this one against the eventual goal of better serving the vast majority of JAX users.

I think reasonable people can disagree with the details of this change. I stil feel strongly that this is the direction we need to go.

mattjj:
Thank you for the thoughtful input @nouiz ! Those are good points, and you have us reconsidering things.

At the risk of just repeating what's already been said (apologies!), IIUC our desiderata include:
1. behavior independent of the setting of the jax_enable_x64 config option [main motivation for making any changes at all]
2. for mitigating user surprise, consistency with (i) past jax, (ii) numpy, (iii) torch, (iv) tensorflow
3. for mitigating user surprise, don't 'silently' insert stuff
4. convenience of having a short way to express doing the right thing in the most common situations
5. allow users to express all possibilities (e.g. reducing an int8 array to an int8 result, as well as reducing an int8 array to a higher-precision result)

Some of these may be in tension with one another.

Which goals are the most important? What's the solution space for meeting the right ones?

In the OP Jake mentioned having an option like "promote_integers" by which users can make the behavior explicit. It seems like a good idea to factor things as (a) a way to be explicit and (b) a default convenience behavior when not explicit. Actually the `dtype` argument for `np.sum` may already provide such a mechanism:
```
dtype : dtype, optional
    The type of the returned array and of the accumulator in which the
    elements are summed.  The dtype of `a` is used by default unless `a`
    has an integer dtype of less precision than the default platform
    integer.  In that case, if `a` is signed then the platform integer
    is used while if `a` is unsigned then an unsigned integer of the
    same precision as the platform integer is used.
```

So perhaps another option is to just say the "default platform integer" for JAX is 32 bits wide? Then we wouldn't be changing any behavior for users when `JAX_ENABLE_X64=0` (the overwhelmingly common case AIUI), and we would be sticking close to NumPy. Indeed Jake pointed out that we have `JAX_DEFAULT_DTYPE_BITS` which may configure this default width.

FWIW on the second point, `torch` and `tensorflow` seem to differ:
```
In [2]: torch.sum(torch.tensor([1, 2, 3], dtype=torch.int8)).dtype
Out[2]: torch.int64

In [4]: tf.reduce_sum(tf.convert_to_tensor([1,2,3], dtype='int8')).dtype
Out[4]: tf.int8
```

From first principles, I can see arguments for having vs not having any automatic promotion. But at the same time the weight of numpy precedent, plus JAX's own current-behavior precedent, seem much heavier.

Anyway, I think an API where we
* keep allowing use of the `dtype` argument for an explicit result dtype, and
* when `dtype` is not specified have the default behavior on integer inputs be to return an at-least 32bit result (vs numpy's at-least 64bit)

would meet (1), 2(i), (4), and (5), and be pretty close to 2(ii) and 2(iii) modulo whether the "platform integer" is 32 vs 64 bits.

Anyway, just adding to the discussion! Maybe there are still more options to consider?

jakevdp:
A counterpoint to that: automatic promotion by `np.sum` is relatively surprising, and this really only affects people using types like `uint8`/`int8`/`uint16`/`int16` which are rarely landed upon accidentally. A user might reasonably expect that, e.g. summing the rows of an `uint8` array would return an `uint8` array. I know I've been bitten by this in the past when I do a large NUM_GENERATED_CASES test across all dtypes for some new JAX API implementation.

froystig:
@mattjj:
> when dtype is not specified have the default behavior on integer inputs be to return an at-least 32bit result (vs numpy's at-least 64bit)

The "at-least" here might reintroduce confusion, so just to be extra clear: this means "a fixed, at-least 32bit width, that does not vary with the jax_enable_x64 config option"

froystig:
> Anyway, I think an API where we
> 
> * keep allowing use of the `dtype` argument for an explicit result dtype, and
> * when `dtype` is not specified have the default behavior on integer inputs be to return an at-least 32bit result (vs numpy's at-least 64bit)

We also have `jax.lax`, where things are typically more explicit (broadly speaking, it offers less by way of automatic argument promotion and broadcasting), and where similarity to numpy or other libraries is less expected. I can imagine in `lax` an API where, if the working-and-output width is not specified, then it is not auto-widened, the way @jakevdp seems to prefer.

I think 2(ii) is important in `jax.numpy`. If nothing else, it's in the name. We do deviate from numpy on occasion, but typically only out of necessity, and minimally so.

jakevdp:
> We also have `jax.lax`, where things are typically more explicit

True, but there's no `lax.sum()`. I think the main available public API in `lax` for array summation would look something like this:
```python
lax.reduce(x, jnp.array(0, dtype=x.dtype), lax.add, (0,))
````

**Comment by jakevdp** [comment]
It sounds like most people are in favor of promotion for integer `sum` and `prod`, but only promoting to the size specified by `JAX_DEFAULT_DTYPE_BITS`. I'll update the PR to do that instead

(side note: should we remove the `promote_integers=True` option? I only added that because we had initially decided to move to non-promotion, so there's no real reason to support he keyword if we're not doing that. That being said, it has made it into a couple releases now, so it's technically a breaking change to remove it...)

**Comment by jakevdp** [comment]
In the current version, this should have no user-facing changes, unless someone is setting `JAX_DEFAULT_DTYPE_BITS=32` (which has not really been advertised yet)

**Comment by nouiz** [comment]
Sorry for the delay, I was sick.
I like the new proposal.

I think it is the background of people that will make them surprised or not of the selected behavior. C/C++ and Python(that use big int, int with infinite precision)/Numpy have different basic behavior. So if a main user of those community see a different behavior then what he is used, he will be surprised.

Should we document that in the dtype documentation of JAX to help propagate this knowledge and limit the surprise effect?

I checked the current code and it doesn't use JAX_DEFAULT_DTYPE_BITS anywhere. Is this normal?

**Comment by nouiz** [comment]
On promote_integers, removing it would be great.
As it isn't really used, why not skip the warning step of deprecation and go directly to raise a good error step before removing it?
Set promote_integers=None. If the user set it to anything, raise an error that tell how to fix it?

**Comment by jakevdp** [comment]
I'm thinking of just removing `promote_integers`. It was only added for the sake of the original plan here, and has only existed for a release or two.

Regarding the default dtype: that enters things here via `dtypes.int_` and `dtypes.float_` which are set based on the import-time value of `JAX_DEFAULT_DTYPE_BITS`.

File: jax/_src/numpy/reductions.py [hunk 1]
```diff
@@ -86,17 +86,21 @@ def _reduction(a, name, np_fun, op, init_val, has_identity=True,
...
       raise ValueError(f"zero-size array to reduction operation {name} which has no identity")
 
-  result_dtype = dtypes.canonicalize_dtype(dtype or dtypes.dtype(a))
+  result_dtype = dtype or dtypes.dtype(a)
 
-  # promote_integers=True matches NumPy's behavior for sum() and prod(), which promotes
-  # all int-like inputs to the widest available dtype.
   if dtype is None and promote_integers:
+    # Note: NumPy always promotes to 64-bit; jax instead promotes to the
+    # default dtype as defined by dtypes.int_ or dtypes.uint.
```
nouiz, line 93: Maybe comment that they are set by JAX_DEFAULT_DTYPE_BITS?

jax/_src/numpy/reductions.py
```diff
@@ -86,17 +86,21 @@ def _reduction(a, name, np_fun, op, init_val, has_identity=True,
     if not _all(core.greater_equal_dim(shape[d], 1) for d in pos_dims):
       raise ValueError(f"zero-size array to reduction operation {name} which has no identity")
 
-  result_dtype = dtypes.canonicalize_dtype(dtype or dtypes.dtype(a))
+  result_dtype = dtype or dtypes.dtype(a)
 
-  # promote_integers=True matches NumPy's behavior for sum() and prod(), which promotes
-  # all int-like inputs to the widest available dtype.
   if dtype is None and promote_integers:
```
nouiz, line 91: `promote_integers=False` by default. So by default this branch won't be taken by default.
Should some basic test be added? Just add a check to an existing test?

jakevdp: There is an existing test: https://github.com/google/jax/blob/37f9db77f72055bb3eed2b3e5e1f99a1a2ad825b/tests/lax_numpy_reducers_test.py#L301-L302

nouiz: Thanks. The PR looks good to me.
Do you want to update the documentation and the  release notes in this PR?
I'll let other people continue the review of this PR.

jakevdp: I don't think there's any need to update documentation or add a release note, because this change causes no user-visible effect. The only thing it changes is the behavior of `sum` for integers when `JAX_DEFAULT_DTYPE_BITS` is flipped from its default value, and that flag is deliberately undocumented because it's not intended to be used by users.

nouiz: So who should use that flag?
Or this PR is only to help the switch later and not now?

jakevdp: At the moment users should not be concerned about the flag. It exists for the purpose of a future migration that has not yet started, and we’ve done many PRs like this one whose only affect is to make that future migration better behaved without affecting current behavior. Does that make sense?

nouiz: Yes. Thanks.

froystig (approved):

PR #12539, opened 2022-09-27
Create jax_random_future flag & make jax.random.loggamma more robust

jakevdp:
This fixes a long-standing TODO regarding the implementation of `random.loggamma`; technically this is a breaking change as it changes the stream of random values that are produced. For this reason, I put the changes behind a `jax_random_future` flag, as suggested by @froystig to align these changes with other upcoming changes to `jax.random`.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -45,6 +45,7 @@ jobs:
             os: ubuntu-latest
             enable-x64: 0
             prng-upgrade: 0
+            random-future: 0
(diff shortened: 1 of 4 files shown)


froystig (commented):
Should we also enable `jax_random_future` in one of our github CI runs?

jakevdp:
> Should we also enable `jax_random_future` in one of our github CI runs?

Done

jakevdp:
Replaced by #16779

PR #12624, opened 2022-10-03
Fix book links

sergedurand:
Fixed two broken links for the books mentioned in documentation, in the autodiff cookbook.


jakevdp (approved):
Thanks!

PR #12628, opened 2022-10-03
Add an internal jtu.sample_product test decorator.

hawkinsp:
This decorator samples from a cartesian product of parameterized tests without materializing the full product explicitly.

Update lax_test.py to use the new decorator.

On my desktop machine, this improves the timing for `pytest --collect-only tests/lax_test.py` from 6.8s to 1.9s.

### DIFF
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -698,6 +698,41 @@ def named_cases_from_sampler(gen):
     yield case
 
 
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks great! Nice idea

mattjj (approved):
Peter, thanks so much for this.

PR #12642, opened 2022-10-03
Set JAX_PLATFORMS=tpu,cpu on TPUs

RissyRan:
* Set `JAX_PLATFORMS=tpu,cpu` automatically if libtpu is installed.
* It would make JAX raise an error if TPU cannot be initialized instead of falling back to CPU.
* Tests:
  * Run test file below on a non-TPU platform, it throws `Unable to initialize backend 'tpu': NOT_FOUND` error
  * Run test file below on a non-TPU platform with `export JAX_PLATFORMS=cpu`, and it runs with CPU (`No GPU/TPU found, falling back to CPU`).
  * Run test file on a non-TPU platform with both `export JAX_PLATFORMS=''` and `export JAX_PLATFORMS=` separately, and it executes with CPU (`No GPU/TPU found, falling back to CPU`).
  * Run test file on both non-TPU and TPU platforms with `export JAX_PLATFORMS=ccc`, and get error message below.

**test file**:
```
import jax
import os
import time

for i in range(5):
    print("checking...")
    print(os.environ.get('JAX_PLATFORMS'))
    print(jax.devices())
    time.sleep(10)
```

**error message**:
```
  File "/usr/local/google/home/ranran/jax/jax_test.py", line 8, in <module>
    print(jax.devices())
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 487, in devices
    return get_backend(backend).devices()
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 429, in get_backend
    return _get_backend_uncached(platform)
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 413, in _get_backend_uncached
    bs = backends()
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 354, in backends
    raise RuntimeError(err_msg + setup_msg)
RuntimeError: Unable to initialize backend 'ccc': Unknown backend 'ccc' (set JAX_PLATFORMS='' to automatically choose an available backend)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,12 @@ Remember to align the itemized text with the first line of an item within a list
 PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 -->
 
-## jax 0.3.22
+## jax 0.3.22 (Oct 5, 2022)
+* Changes
+  * Add `JAX_PLATFORMS=tpu,cpu` as default setting in TPU initialization, 
+  so JAX will raise an error if TPU cannot be initialized instead of falling 
+  back to CPU. Users are able to override this behavior to automatically choose 
+  an available backend by setting `JAX_PLATFORMS=''`.
 
 ## jaxlib 0.3.22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,13 @@ Remember to align the itemized text with the first line of an item within a list
 PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 -->
 
-## jax 0.3.22 (Oct 5, 2022)
+## jax 0.3.22
 * Changes
-  * Add `JAX_PLATFORMS=tpu,cpu` as default setting in TPU initialization, 
-  so JAX will raise an error if TPU cannot be initialized instead of falling 
-  back to CPU. Users are able to override this behavior to automatically choose 
-  an available backend by setting `JAX_PLATFORMS=''`.
+  * Add `JAX_PLATFORMS=tpu,cpu` as default setting in TPU initialization,
+  so JAX will raise an error if TPU cannot be initialized instead of falling
+  back to CPU. Set `JAX_PLATFORMS=''` to override this behavior and automatically
+  choose an available backend (the original default), or set `JAX_PLATFORMS=cpu`
+  to always use CPU regardless of if the TPU is available.
 
 ## jaxlib 0.3.22
 
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -350,8 +350,8 @@ def backends():
           # we expect a RuntimeError.
           err_msg = f"Unable to initialize backend '{platform}': {err}"
           if config.jax_platforms:
-            setup_msg = f" (set JAX_PLATFORMS='' to choose backend automatically)"
-            raise RuntimeError(err_msg + setup_msg)
+            err_msg += " (set JAX_PLATFORMS='' to automatically choose an available backend)"
+            raise RuntimeError(err_msg)
           else:
             _backends_errors[platform] = str(err)
             logging.info(err_msg)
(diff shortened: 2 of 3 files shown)


skye:
We should make sure it's possible for users to manually override to the default behavor (i.e. preferring TPU if available, but falling back to CPU if not). Can you test that manually setting JAX_PLATFORMS= (empty string) causes it to pickup the default behavior? I'm not totally sure if this will work, it might override the empty string (vs. just overriding if it's not set at all).

We might also wanna provide some guidance in the error message for users who hit this. I'm not totally sure how best to do this. Maybe add to err_msg [here](https://github.com/google/jax/blob/main/jax/_src/lib/xla_bridge.py#L353): `(Set JAX_PLATFORMS='' to auto fallback to CPU or JAX_PLATFORMS=cpu to always use CPU)` or something like that? You should test it out and make sure it looks reasonable.

RissyRan:
Sure! Added changes for error message.

* Ran test file on a non-TPU platform with both `export JAX_PLATFORMS=''` and `export JAX_PLATFORMS=` separately, and it executed with CPU (`No GPU/TPU found, falling back to CPU`).
* Ran test file on both non-TPU and TPU platforms with `export JAX_PLATFORMS=ccc`, and got error message below.

```
  File "/usr/local/google/home/ranran/jax/jax_test.py", line 8, in <module>
    print(jax.devices())
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 487, in devices
    return get_backend(backend).devices()
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 429, in get_backend
    return _get_backend_uncached(platform)
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 413, in _get_backend_uncached
    bs = backends()
  File "/usr/local/google/home/ranran/jax/jax/_src/lib/xla_bridge.py", line 354, in backends
    raise RuntimeError(err_msg + setup_msg)
RuntimeError: Unable to initialize backend 'ccc': Unknown backend 'ccc'(Set JAX_PLATFORMS='' to auto fallback to CPU or JAX_PLATFORMS=cpu to always use CPU)
```

jax/_src/lib/xla_bridge.py
```diff
@@ -350,7 +350,8 @@ def backends():
           # we expect a RuntimeError.
           err_msg = f"Unable to initialize backend '{platform}': {err}"
           if config.jax_platforms:
-            raise RuntimeError(err_msg)
+            setup_msg = f" (set JAX_PLATFORMS='' to choose backend automatically)"
```
skye, line 353: One more thought: how about `to automatically choose an available backend)`

Also you don't need the `f` since you're not formatting anything. Also you can use `err_msg +=` if you want

skye (changes requested):
One more tiny comment, otherwise LGTM

jax/_src/lib/xla_bridge.py
RissyRan: Sounds good.

skye (approved):

skye:
Oops two more things: update the CHANGELOG, and fix the lint error

CHANGELOG.md
```diff
@@ -8,7 +8,12 @@ Remember to align the itemized text with the first line of an item within a list
 PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 -->
 
-## jax 0.3.22
+## jax 0.3.22 (Oct 5, 2022)
```
skye, line 11: We only update the date when we actually make a release, sorry for the confusion. So please remove this date since we're not actually releasing.

CHANGELOG.md
```diff
@@ -8,7 +8,12 @@ Remember to align the itemized text with the first line of an item within a list
...
+* Changes
+  * Add `JAX_PLATFORMS=tpu,cpu` as default setting in TPU initialization, 
+  so JAX will raise an error if TPU cannot be initialized instead of falling 
+  back to CPU. Users are able to override this behavior to automatically choose 
+  an available backend by setting `JAX_PLATFORMS=''`.
```
skye, line 16: nit: I think it's a bit strange to mention "users" in a changelog since users will probably be reading it :) Also let's maybe mention JAX_PLATFORMS=cpu here. So something like: 

Set `JAX_PLATFORMS=''` to override this behavior and automatically choose an available backend (the original default), or set `JAX_PLATFORMS=cpu` to always use CPU regardless of if the TPU is available.

skye (changes requested):

skye (approved):

Issue #12618, opened 2022-10-02
Possible contribution: scipy.stats.truncnorm


adrn:
Please:

- [x] Check for duplicate requests.
- [x] Describe your goal, and if possible provide a code snippet with a motivating example.

I notice there is functionality for generating samples from a truncated normal in `jax.random.truncated_normal`, but no way of evaluating the pdf/cdf of a truncated normal (a la `scipy.stats.truncnorm`). I need this for one of my own projects, so implemented it locally, but I would be happy to contribute it here! I implemented it as a parallel implementation to `sciyp.stats.truncnorm` (and could contribute that to `jax.scipy.stats.truncnorm`), but I would also be happy to have a modified version with limits in the scaled/shifted space (`lower`, `upper` vs. `a`, `b`) like `jax.random.truncated_normal`.

jakevdp:
Hey Adrian, thanks for the request! I think we'd welcome that contribution to `jax.scipy.stats`, along the lines of the existing distributions defined there. Are you interested in contributing your implementation?

adrn:
👋 Hey! Nice to hear from you :).

Yes, happy to — I'll open a PR. thanks!

jakevdp:
Cool, feel free to request a review from me when it's ready

PR #12646, opened 2022-10-04
Implement jax.scipy.stats.truncnorm

adrn:
Following from #12618, here is an implementation of the logpdf, logcdf, and logsf (and un-log versions) for `scipy.stats.truncnorm`. Much of this is ported over from scipy itself, but I had to change some logic to use `select()` statements. I was a little confused by the test structure but think I figured out a solution to enforce that the upper limit parameter `b` is always larger than the lower limit `a`, and the evaluating values `x` are between these values.

### DIFF
--- a/jax/_src/scipy/stats/truncnorm.py
+++ b/jax/_src/scipy/stats/truncnorm.py
@@ -1,0 +1,130 @@
+# Copyright 2022 The JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import scipy.stats as osp_stats
+
+from jax import lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.util import _wraps
+from jax._src.numpy.lax_numpy import _promote_args_inexact
+from jax._src.scipy.stats import norm
+from jax._src.scipy.special import logsumexp, log_ndtr, ndtr
+
+
+def _log_diff(x, y):
+  return logsumexp(
+    jnp.array([x, y]),
+    b=jnp.array([jnp.ones_like(x), -jnp.ones_like(y)]),
+    axis=0
+  )
+
+
+def _log_gauss_mass(a, b):
+  """Log of Gaussian probability mass within an interval"""
+  a, b = jnp.array(a), jnp.array(b)
+  a, b = jnp.broadcast_arrays(a, b)
+
+  # Note: Docstring carried over from scipy
+  # Calculations in right tail are inaccurate, so we'll exploit the
+  # symmetry and work only in the left tail
+  case_left = b <= 0
+  case_right = a > 0
+  case_central = ~(case_left | case_right)
+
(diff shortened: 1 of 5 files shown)


adrn:
@jakevdp I don't have privileges to request a review from you, so pinging you instead! (and I just filled out the CLA)

tests/scipy_stats_test.py
```diff
@@ -472,6 +472,143 @@ def args_maker():
...
+    scipy_fun = osp_stats.truncnorm.logpdf
+    lax_fun = lsp_stats.truncnorm.logpdf
+
+    def args_maker():
+      x, a, b, loc, scale = map(rng, shapes, dtypes)
+
+      # hack to ensure that b > a and a < x < b
+      a = -np.abs(a).astype(a.dtype)
+      x = ((a + b) / 2).astype(x.dtype)
+      b = np.abs(b).astype(b.dtype)
```
jakevdp, line 487: What happens if these are not in order `a < x < b`? It should return `-inf`, correct? Should we just let the values be out of order so as to test that?

If we need them in order, it might be simpler to do something like this, and then cast to the desired dtype:
```python
a, x, b = rng((3, *shape), dtype).sort(0)
```

jakevdp (commented):
Looks great! Just a quick comment on tests.

tests/scipy_stats_test.py
adrn: Yea, true, I guess we can relax the need for `x` to be in bounds, but `b` must be greater than `a`. But since `a` or `b` can have different shapes, I couldn't figure out a clean way to use `np.sort()` to get this to work. My hack might be the simplest way of enforcing this, but I could think more about it!

jakevdp: It looks like when `a >= b` the scipy version returns `nan`:
```python
from scipy.stats import truncnorm
print(truncnorm.logpdf(1, a=2, b=0))
# nan
```
I believe `_CheckAgainstNumpy` will check the locations of NaN values, so it might be best in the tests to check the full range of values to ensure that we're matching that behavior.

adrn:
@jakevdp Sorry for the delay here! I removed the hack parameter value enforcement from the tests (now it tests invalid parameter values as well) and rebased because of a conflict.

jakevdp (commented):
Thanks for circling back to this - looks great! Two last things before we merge:

- please add the new functionality to the HTML docs in https://github.com/google/jax/blob/main/docs/jax.scipy.rst
- please squash all your changes into a single commit (see [single-change commits](https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests))

Thanks!

adrn:
Cool - thanks for the review! I just squashed the commits and added it to the docs.

jakevdp (approved):

adrn:
Fixed a small docs issue and pushed again, so tests might need to be approved again.

jakevdp:
The doc build is failing in a strange way, that seems to be related to our lazy loading of scipy. I'm not sure what the fix might be...

jax/_src/scipy/stats/truncnorm.py
```diff
@@ -0,0 +1,129 @@
...
+# limitations under the License.
+
+
+import scipy.stats as osp_stats
+
+from jax import lax
+from jax._src.numpy import lax_numpy as jnp
+from jax._src.numpy.util import _wraps
+from jax._src.numpy.lax_numpy import _promote_args_inexact
+from jax.scipy import special, stats
```
jakevdp, line 22: I think this is the issue: importing `stats` here leads to a circular import.

Using `from jax._src.scipy.stats import norm` should fix this.

adrn: Got it - done!

jakevdp:
Thanks for the PR!

adrn:
Thanks for the review!

PR #12654, opened 2022-10-04
[typing] add annotations to numpy.fft

jakevdp:
Part of #12049

### DIFF
--- a/jax/_src/numpy/fft.py
+++ b/jax/_src/numpy/fft.py
@@ -14,6 +14,7 @@
 
 
 import operator
+from typing import Optional, Sequence, Union
 import numpy as np
 
 from jax import dtypes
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/fft.py
```diff
@@ -268,8 +299,9 @@ def fftshift(x, axes=None):
 
 
 @_wraps(np.fft.ifftshift)
-def ifftshift(x, axes=None):
+def ifftshift(x: ArrayLike, axes: Union[None, int, Sequence[int]] = None):
```
soraros, line 302: Is the omission of return type here intentional?

jakevdp: No, not intentional, thanks for catching that!

yashk2810 (approved):

PR #12659, opened 2022-10-04
Add PYTORCH_SAME padding to JAX.

RerRayne:
Hi. We discussed adding an additional flag to padding to raise an awareness of users that there are two options. One is JAX/TF-like 'SAME' and the second one is a PyTorch-like 'SAME'.


jax/_src/lax/lax.py
```diff
@@ -4519,25 +4519,43 @@ def _dilate_shape(shape, dilation):
 def _ceil_divide(x1, x2):
   return -np.floor_divide(np.negative(x1), x2)
 
+
+jax_enum_names = [*[es.name for es in xla_client.PaddingType], 'PYTORCH_SAME']
+JaxPaddingType = enum.Enum(
```
apaszke, line 4524: I think `PaddingType` would be a better name. It clearly is defined in JAX, as opposed to `xla_client.PaddingType`

jax/_src/lax/lax.py
```diff
@@ -4519,25 +4519,43 @@ def _dilate_shape(shape, dilation):
...
   """Convert padding string to list of pairs of pad values."""
-  PaddingType = xla_client.PaddingType
-
   if isinstance(padding, str):
-    mapping = {'VALID': PaddingType.VALID, 'SAME': PaddingType.SAME}
+    mapping = {
+        'VALID': JaxPaddingType.VALID,
+        'SAME': JaxPaddingType.SAME,
+        'PYTORCH_SAME': JaxPaddingType.PYTORCH_SAME,
+    }
```
apaszke, line 4537: This is unnecessary. `mapping[some_string]` is equivalent to `JaxPaddingType[some_string]`. See [`enum` docs](https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes)

jax/_src/lax/lax.py
```diff
@@ -4519,25 +4519,43 @@ def _dilate_shape(shape, dilation):
...
+  elif padding == JaxPaddingType.VALID:
     return [(0, 0)] * len(in_shape)
+  elif padding == JaxPaddingType.PYTORCH_SAME:
+    out_shape = _ceil_divide(in_shape, window_strides)
+    pad_sizes = np.maximum(
+        0,
+        (out_shape - 1) * window_strides + window_shape - in_shape,
+        )
+    return [((pad_size + 1) // 2, pad_size - (pad_size + 1) // 2)
+            for pad_size in pad_sizes]
```
apaszke, line 4558: This seems very very similar to the `SAME` case, so perhaps consider merging them? Wouldn't this code work?
```
if padding == VALID:
  return [(0, 0)] * len(in_shape)
else:
  out_shape = ...
  pad_sizes = ...
  half_pad_sizes = (pad_size + (padding == PYTORCH_SAME)) // 2
  return [(hps, pad_size - hps) for hps in half_pad_sizes]
```

apaszke (commented):
Do you have a link to the padding calculation in PyTorch?

jax/_src/lax/lax.py
RerRayne: Renamed, thanks

jax/_src/lax/lax.py
RerRayne: Ok, no problem, I just followed the initial style. Fixed.

jax/_src/lax/lax.py
```diff
@@ -4534,23 +4534,35 @@ def _dilate_shape(shape, dilation):
...
+    pad_sizes = np.maximum(
+        0,
+        (out_shape - 1) * window_strides + window_shape - in_shape,
+    )
+    if padding == PaddingType.SAME:
+      return [(pad_size // 2, pad_size - pad_size // 2)
+              for pad_size in pad_sizes]
+    else:
+      return [((pad_size + 1) // 2, pad_size - (pad_size + 1) // 2)
+              for pad_size in pad_sizes]
```
apaszke, line 4565: Those two return statements only differ in `pad_size` vs `pad_size + 1`, so you could deduplicate them even more.

apaszke (approved):
This LGTM, but could you point me to the source of PyTorch padding size calculation that you used to derive this?

jax/_src/lax/lax.py
RerRayne: I relied on two sources.
1. [PyTorch implementation](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/conv.py#L120)
2. [haiku.pad.same](https://github.com/deepmind/dm-haiku/blob/main/haiku/_src/pad.py#L37) implementation which is equivalent to the first one.
3. [Converting tensorflow model to pytorch: issue with padding](https://discuss.pytorch.org/t/converting-tensorflow-model-to-pytorch-issue-with-padding/84224) blogpost

jax/_src/lax/lax.py
```diff
@@ -4534,23 +4534,35 @@ def _dilate_shape(shape, dilation):
...
     out_shape = _ceil_divide(in_shape, window_strides)
-    pad_sizes = np.maximum(0, (out_shape - 1) * window_strides +
-                                window_shape - in_shape)
-    return [(pad_size // 2, pad_size - pad_size // 2) for pad_size in pad_sizes]
+    pad_sizes = np.maximum(
+        0,
+        (out_shape - 1) * window_strides + window_shape - in_shape,
+    )
+    if padding == PaddingType.PYTORCH_SAME:
+      shift_pad_sizes = pad_sizes + 1
```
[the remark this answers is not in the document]
apaszke: Also, I can't prove to myself that this method is actually equivalent to what PyTorch is doing. The PyTorch code you linked does roughly this:
```
pad_size = ...
left_pad = pad_size // 2
right_pad = pad_size - left_pad
```
So, if `pad_size` is odd, the excess padding will be put on the right. But your code here favors moving it to the left, because of the `+ 1`?

apaszke (commented):
I can't verify that the code is equivalent to the implementation in PyTorch---see my comment inline. Also, please fix the lint.

jax/_src/lax/lax.py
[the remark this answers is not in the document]
RerRayne: Yeah, it doesn't because PyTorch code uses a bit different definition. However, to make sure that it works right, I added a correspondent test that contains a copy of padding from PyTorch.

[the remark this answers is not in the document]
apaszke: > Yeah, it doesn't because PyTorch code uses a bit different definitions. However, to make sure that it works right, I added a correspondent tests that contains a copy of padding from PyTorch.

Sorry, I don't follow. Can you please elaborate on why this method is equivalent?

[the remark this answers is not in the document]
RerRayne: Okay, let's consider Haiku implementation first:
```
def same(effective_kernel_size: int) -> Tuple[int, int]:
  """Pads such that the output size matches input size for stride=1."""
  return ((effective_kernel_size - 1) // 2, effective_kernel_size // 2)
```
Here ```same``` operates with ```effective_kernel_size``` which is calculated as ```effective_kernel_size=(kernel_size-1)*rate + 1``` [[Link](https://github.com/deepmind/dm-haiku/blob/main/haiku/_src/pad.py#L86)]

Now let's look at PyTorch:
```
for d, k, i in zip(dilation, kernel_size,
                                   range(len(kernel_size) - 1, -1, -1)):
  total_padding = d * (k - 1)
  left_pad = total_padding // 2
  self._reversed_padding_repeated_twice[2 * i] = left_pad
  self._reversed_padding_repeated_twice[2 * i + 1] = (total_padding - left_pad)
```

So here ```total_padding=effective_kernel_size-1```.  ```total_padding - left_pad ``` is equivalent to ```effective_kernel_size // 2``` because ```total_padding = left_p - right_p = 0.5 left_p - right_p -> right_p = 0.5 total_padding```.

I hope this helps.

PR #12673, opened 2022-10-05
[typing] add type annotations to lax.linalg functions

jakevdp:
Part of #12049

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -15,11 +15,12 @@
 import inspect
 import functools
 from functools import partial
+from typing import cast, Any, Callable, List, Optional, Tuple, TypeVar, Union, overload
 import warnings
 
 import numpy as np
+from typing_extensions import Literal
 
-import jax
 from jax._src.numpy import lax_numpy as jnp
 from jax._src.numpy.vectorize import vectorize
 from jax._src import ad_util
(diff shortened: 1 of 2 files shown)


jax/_src/lax/linalg.py
```diff
@@ -131,7 +134,8 @@ def cholesky(x, *, symmetrize_input: bool = True):
   return jnp.tril(cholesky_p.bind(x))
 
 @_warn_on_positional_kwargs
-def eig(x, *, compute_left_eigenvectors=True, compute_right_eigenvectors=True):
+def eig(x: ArrayLike, *, compute_left_eigenvectors: bool = True,
+        compute_right_eigenvectors: bool = True) -> Tuple[Array, Array]:
```
sharadmv, line 138: Does this not return a tuple of 3 arrays (also dependent on the values of the `compute_left/right` kwargs)?

jax/_src/lax/linalg.py
```diff
@@ -194,7 +198,7 @@ def lu_pivots_to_permutation(pivots, permutation_size: int):
   return permutation
 
 
-def lu(x):
+def lu(x: ArrayLike) -> Tuple[Array, Array, Array]:
```
sharadmv, line 201: Why are some functions in `lax` consume `Array` and others consuming `ArrayLike`?

jax/_src/numpy/linalg.py
```diff
@@ -531,10 +531,9 @@ def norm(x: ArrayLike, ord: Union[int, str, None] = None,
     raise ValueError(
         f"Invalid axis values ({axis}) for jnp.linalg.norm.")
 
-
 @_wraps(np.linalg.qr)
 @partial(jit, static_argnames=('mode',))
-def qr(a: ArrayLike, mode: str = "reduced") -> Tuple[Array, Array]:
+def qr(a: ArrayLike, mode: str = "reduced") -> Union[Array, Tuple[Array, Array]]:
```
sharadmv, line 536: Can we use a Union of Literals here?

sharadmv (changes requested):

jax/_src/lax/linalg.py
jakevdp: Some consume `Array` because they access array attributes (like `shape`) before passing the input to a function that converts it to an array.

jakevdp: I chose to annotate the existing implementation here rather than to change the implementation for better consistency.

jax/_src/numpy/linalg.py
jakevdp: I think we should avoid that; see response at https://github.com/google/jax/pull/12674#discussion_r989313739

jax/_src/lax/linalg.py
jakevdp: Good catch - fixed!

sharadmv (approved):

PR #12675, opened 2022-10-05
make device_put work with Sharding 2nd arg

mattjj:
@yashk2810 is co-author and reviewer:
image

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -1291,14 +1291,38 @@ def _copy_array_to_device(x: jax.Array, device: Optional[xc.Device]) -> jax.Arra
       committed=(device is not None))
 
 
-def _device_put_impl(x, device: Optional[Device] = None):
+def _device_put_impl(
+    x, device: Optional[Union[Device, jax.sharding.Sharding]] = None):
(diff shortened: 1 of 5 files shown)


yashk2810 (approved):

jax/_src/dispatch.py
```diff
@@ -1291,15 +1291,39 @@ def _copy_array_to_device(x: jax.Array, device: Optional[xc.Device]) -> jax.Arra
...
+
+  if isinstance(x, array.ArrayImpl):
+    if not x.is_fully_addressable():
+      raise ValueError(
+          "device_put's first argument must be a fully addressable array, but "
+          f"got value with devices {x.devices()}")
+    if device is None:
+      return x
+    elif is_single_device_sharding(x.sharding):
+      return _copy_array_to_device(x, device)
```
yashk2810, line 1322: I think we missed one case here:

```
if len(x.sharding.device_set) > 1:
  do what? raise an error saying you can't device_put an Array sharded over multiple devices onto a single device?

  Or bounce to host and then device_put again to the requested device?
```

mattjj: We fixed this together!

PR #12677, opened 2022-10-05
improve jit(f).lower(duck_args) and pjit(f).lower(duck_args)

mattjj:


jax/_src/stages.py
```diff
@@ -379,6 +379,20 @@ def runtime_executable(self) -> Optional[Any]:
     """
     return self._executable.runtime_executable()
 
+  @property
+  def input_shardings(self):  # PyTree[XLACompatibleSharding]
+    shardings_flat = getattr(self._executable, '_in_shardings', None)
+    if shardings_flat is None:
+      raise ValueError("no input_shardings on compiled object")
```
froystig, line 386: Under what conditions do we expect the underlying executable not to have shardings? Why err if so? Could we instead return something indicating a default/trivial sharding (assuming that's what no shardings would imply)?

Same question below for `output_shardings`.

yashk2810: There can be cases where some Executable won't be a MeshExecutable and that won't have a sharding on it. In JAX that won't/should not happen but I don't know about other cases in google.

> indicating a default/trivial sharding

I don't think that would be correct because the executable should *always* have well defined shardings which we don't infer.

froystig: Then how about a `NotImplementedError` instead that explains the above? Even an `AttributeError` seems like an improvement over `ValueError`, which suggests that the caller passed in something of invalid value. The caller did nothing wrong here – the backend is at fault.

Also, if this is something we _expect_ underlying executables to have, then we should add it to the Executable protocol.

froystig: I think we should still consider adding in/out shardings as methods to the `Exectuable` protocol to indicate that we expect it. I can push this as a change to the branch if you'd like me to.

froystig (commented):
I pushed the change I suggested. Let's add a test next?

PR #12679, opened 2022-10-06
Avoid loading scipy eagerly.

hawkinsp:
scipy accounts for around 400ms of the 900ms of JAX's import time. By loading scipy lazily, we can improve the timing of `import jax` down to about 500ms.


jakevdp (approved):
Can we add a unit test of `lazy_loader`, since it's doing some slightly nontrivial things?

hawkinsp:
> Can we add a unit test of `lazy_loader`, since it's doing some slightly nontrivial things?

Done.

jakevdp (approved):

PR #9141, opened 2022-01-08
Implemented mode function

Yash621:
this PR is respect to #8978 , @jakevdp  please review it.

I have implemented the mode function ,and have also created the _core.py file  and have placed the function there.
also i have added the license required and written unit tests as you guided under scipy_stats_test.py file.

please review my PR and let me know if any changes are required.

PR #12660, opened 2022-10-05
Migrate api_test, lax_numpy_test, and lax_vmap_test to

hawkinsp:
jtu.sample_product.

Gives a ~2x improvement in pytest --collect-only timing for lax_numpy_test.

PR #12683, opened 2022-10-06
Add JAX equivalent of scipy.stats.mode

ylamidon:
Solution for (https://github.com/google/jax/issues/8978)

The previous PR (https://github.com/google/jax/pull/9141) and (https://github.com/google/jax/pull/11475) seemed stale.

This PR is largely based on these previous contributions and will hopefully resolve the remaining issues.

### DIFF
--- a/jax/_src/scipy/stats/_core.py
+++ b/jax/_src/scipy/stats/_core.py
@@ -1,0 +1,83 @@
+# Copyright 2022 The JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+from functools import partial
+from typing import Optional, Tuple
+
+import jax.numpy as jnp
+import scipy
+from jax import jit
+from jax._src import dtypes
+from jax._src.api import vmap
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.typing import ArrayLike
+from jax._src.util import canonicalize_axis, prod
+
+ModeResult = namedtuple('ModeResult', ('mode', 'count'))
+
+@_wraps(scipy.stats.mode, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=['axis', 'nan_policy', 'keepdims'])
+def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate", keepdims: bool = False) -> ModeResult:
+  _check_arraylike("mode", a)
+  x = jnp.atleast_1d(a)
+
+  if nan_policy not in ["propagate", "omit", "raise"]:
+    raise ValueError(
+      f"Illegal nan_policy value {nan_policy!r}; expected one of "
+      "{'propoagate', 'omit', 'raise'}"
+    )
+  if nan_policy == "omit":
+    # TODO: return answer without nans included.
+    raise NotImplementedError(
+      f"Logic for `nan_policy` of {nan_policy} is not implemented"
+    )
+  if nan_policy == "raise":
+    raise NotImplementedError(
+      "In order to best JIT compile `mode`, we cannot know whether `x` contains nans. "
+      "Please check if nans exist in `x` outside of the `mode` function."
+    )
+
+  input_shape = x.shape
+  if keepdims:
+    if axis is None:
+      output_shape = tuple(1 for i in input_shape)
+    else:
+      output_shape = tuple(1 if i == axis else s for i, s in enumerate(input_shape))
+  else:
+    if axis is None:
+      output_shape = ()
+    else:
+      output_shape = tuple(s for i, s in enumerate(input_shape) if i != axis)
+
+  if axis is None:
+    axis = 0
+    x = x.ravel()
+
+  def _mode_helper(x: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
+    """Helper function to return mode and count of a given array."""
+    if x.size == 0:
+      return jnp.array(jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_)), jnp.array(jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_))
+    else:
+      vals, counts = jnp.unique(x, return_counts=True, size=x.size)
+      return vals[jnp.argmax(counts)], counts.max()
+
+  axis = canonicalize_axis(axis, x.ndim)
+  x = jnp.moveaxis(x, axis, 0)
+  x = x.reshape(x.shape[0], prod(x.shape[1:]))
+  vals, counts = vmap(_mode_helper, in_axes=1)(x)
+  return ModeResult(vals.reshape(output_shape), counts.reshape(output_shape))
--- a/jax/scipy/stats/__init__.py
+++ b/jax/scipy/stats/__init__.py
@@ -33,3 +33,4 @@ from jax.scipy.stats import chi2 as chi2
 from jax.scipy.stats import betabinom as betabinom
 from jax.scipy.stats import gennorm as gennorm
 from jax._src.scipy.stats.kde import gaussian_kde as gaussian_kde
+from jax._src.scipy.stats._core import mode as mode
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -20,14 +20,18 @@ from absl.testing import absltest
 
 import numpy as np
 import scipy.stats as osp_stats
+import scipy.version
 
 import jax
-from jax._src import test_util as jtu, tree_util
+from jax._src import dtypes, test_util as jtu, tree_util
 from jax.scipy import stats as lsp_stats
 from jax.scipy.special import expit
 
 from jax.config import config
 config.parse_flags_with_absl()
+
+scipy_version = tuple(map(int, scipy.version.version.split('.')[:3]))
+numpy_version = tuple(map(int, np.version.version.split('.')[:3]))
 
 all_shapes = [(), (4,), (3, 4), (3, 1), (1, 4), (2, 1, 4)]
 one_and_two_dim_shapes = [(4,), (3, 4), (3, 1), (1, 4)]
@@ -879,6 +883,74 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
     kde2 = tree_util.tree_unflatten(treedef, leaves)
     tree_util.tree_map(lambda a, b: self.assertAllClose(a, b), kde, kde2)
     self.assertAllClose(evaluate_kde(kde, x), kde.evaluate(x))
+
+  @jtu.sample_product(
+    [dict(shape=shape, axis=axis)
+      for shape, axis in (
+        ((0,), None),
+        ((0,), 0),
+        ((7,), None),
+        ((7,), 0),
+        ((47, 8), None),
+        ((47, 8), 0),
+        ((47, 8), 1),
+        ((0, 2, 3), None),
+        ((0, 2, 3), 0),
+        ((0, 2, 3), 1),
+        ((0, 2, 3), 2),
+        ((10, 5, 21), None),
+        ((10, 5, 21), 0),
+        ((10, 5, 21), 1),
+        ((10, 5, 21), 2),
+      )
+    ],
+    dtype=jtu.dtypes.integer + jtu.dtypes.floating,
+    contains_nans=[True, False],
+    keepdims=[True, False]
+  )
+  def testMode(self, shape, dtype, axis, contains_nans, keepdims):
+    if scipy_version < (1, 9, 0) and keepdims != True:
+      self.skipTest("scipy < 1.9.0 only support keepdims == True")
+    if numpy_version < (1, 21, 0) and contains_nans:
+      self.skipTest("numpy < 1.21.0 only support contains_nans == False")
+
+    if contains_nans:
+      rng = jtu.rand_some_nan(self.rng())
+    else:
+      rng = jtu.rand_default(self.rng())
+    args_maker = lambda: [rng(shape, dtype)]
+
+    def scipy_mode_wrapper(a, axis=0, nan_policy='propagate', keepdims=None):
+      """Wrapper to manage the shape discrepancies between scipy and jax"""
+      if scipy_version < (1, 9, 0) and a.size == 0 and keepdims == True:
+        if axis == None:
+          output_shape = tuple(1 for _ in a.shape)
+        else:
+          output_shape = tuple(1 if i == axis else s for i, s in enumerate(a.shape))
+        return (np.full(output_shape, np.nan, dtype=dtypes.canonicalize_dtype(jax.numpy.float_)),
+                np.full(output_shape, np.nan, dtype=dtypes.canonicalize_dtype(jax.numpy.float_)))
+
+      if scipy_version < (1, 9, 0):
+        result = osp_stats.mode(a, axis=axis, nan_policy=nan_policy)
+      else:
+        result = osp_stats.mode(a, axis=axis, nan_policy=nan_policy, keepdims=keepdims)
+
+      if a.size != 0 and axis == None and keepdims == True:
+        output_shape = tuple(1 for _ in a.shape)
+        return (result.mode.reshape(output_shape), result.count.reshape(output_shape))
+      return result
+
+    scipy_fun = partial(scipy_mode_wrapper, axis=axis, keepdims=keepdims)
+    scipy_fun = jtu.ignore_warning(category=RuntimeWarning,
+                                   message="Mean of empty slice.*")(scipy_fun)
+    scipy_fun = jtu.ignore_warning(category=RuntimeWarning,
+                                   message="invalid value encountered.*")(scipy_fun)
+    lax_fun = partial(lsp_stats.mode, axis=axis, keepdims=keepdims)
+    tol_spec = {np.float32: 2e-4, np.float64: 5e-6}
+    tol = jtu.tolerance(dtype, tol_spec)
+    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                            tol=tol)
+    self._CompileAndCheck(lax_fun, args_maker, rtol=tol)
 
 if __name__ == "__main__":
   absltest.main(testLoader=jtu.JaxTestLoader())


jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,70 @@
...
+from jax._src.api import vmap
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.util import canonicalize_axis
+
+ModeResult = namedtuple('ModeResult', ('mode', 'count'))
+
+@_wraps(scipy.stats.mode)
+@partial(jit, static_argnames=['axis', 'nan_policy'])
+def mode(x: jnp.ndarray, axis: int = 0, nan_policy: str = "propagate") -> ModeResult:
```
jakevdp, line 31: Let's use `x: ArrayLike` (same import as `Array` above)

Also, you check `axis is None` below, should `axis` be `Optional[int]`?

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,70 @@
...
+
+import jax.numpy as jnp
+import scipy
+from jax import jit
+from jax._src.api import vmap
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.util import canonicalize_axis
+
+ModeResult = namedtuple('ModeResult', ('mode', 'count'))
```
jakevdp, line 27: Let's use
```python
from typing import NamedTuple
from jax._src.typing import Array
class ModeResult(NamedTuple):
  mode: Array
  count: Array
```

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,70 @@
...
+      f"Logic for `nan_policy` of {nan_policy} is not implemented"
+    )
+  if nan_policy == "raise":
+    raise NotImplementedError(
+      "In order to best JIT compile `mode`, we cannot know whether `x` contains nans. "
+      "Please check if nans exist in `x` outside of the `mode` function."
+    )
+
+  if x.size == 0:
+      return ModeResult(jnp.empty((0,), x.dtype), jnp.empty((0,), dtype=jnp.int32))
```
jakevdp, line 52: 2-space indent

jax/scipy/stats/__init__.py
```diff
@@ -33,3 +33,4 @@
 from jax.scipy.stats import betabinom as betabinom
 from jax.scipy.stats import gennorm as gennorm
 from jax._src.scipy.stats.kde import gaussian_kde as gaussian_kde
+from jax._src.scipy.stats import _core as _core
```
jakevdp, line 36: `_core` is meant to be an internal-only submodule... instead we should import `mode` directly into this namespace (`jax.scipy.stats.mode` like `scipy.stats.mode`)

tests/scipy_stats_test.py
```diff
@@ -913,5 +913,41 @@ def evaluate_kde(kde, x):
...
+      for contains_nans in [True, False]
+      ))
+  def testMode(self, shape, dtype, axis, contains_nans):
+    if contains_nans:
+      rng = jtu.rand_some_nan(self.rng())
+    else:
+      rng = jtu.rand_default(self.rng())
+    args_maker = lambda: [rng(shape, dtype)]
+    scipy_fun = partial(osp_stats.mode, axis=axis)
+    lax_fun = partial(lsp_stats._core.mode, axis=axis)
```
jakevdp, line 945: This should be `lsp_stats.mode`

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,70 @@
...
+import scipy
+from jax import jit
+from jax._src.api import vmap
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.util import canonicalize_axis
+
+ModeResult = namedtuple('ModeResult', ('mode', 'count'))
+
+@_wraps(scipy.stats.mode)
```
jakevdp, line 29: Perhaps use `lax_description` to mention that custom `nan_policy` is not supported

tests/scipy_stats_test.py
```diff
@@ -913,5 +913,41 @@ def evaluate_kde(kde, x):
     tree_util.tree_map(lambda a, b: self.assertAllClose(a, b), kde, kde2)
     self.assertAllClose(evaluate_kde(kde, x), kde.evaluate(x))
 
+  @parameterized.named_parameters(jtu.cases_from_list(
```
jakevdp, line 916: We're gradually moving the tests to use `jtu.sample_product` instead of `parameteriozed.named_parameters` (see #12660). You could optionally do that here if you're feeling up for it! But if not, this looks fine.

jakevdp (commented):
Looks great! Just a few minor comments

jax/_src/scipy/stats/_core.py
ylamidon: It makes sense
I need to rename the mode input ```x``` to something like ```a``` which is of type ```ArrayLike```
So that I can use ```x = jnp.asarray(a)``` to get ```x``` of type ```Array```
And then I can use methods like ```shape``` or ```ndim``` on ```x```

jax/_src/scipy/stats/_core.py
ylamidon: ```NamedTuple``` as the ```count``` method, so I'll use ```counts``` instead. And also ```modes``` instead of ```mode```.

jakevdp: I think we should keep it as `mode` and `count` to match the numpy API. Those names don't seem to cause issues when Itry them

ylamidon: The problem occurs when using mypy

jakevdp: OK, in that case let's keep the current approach

jakevdp: Correct runtime API is more important than type annotations

tests/scipy_stats_test.py
```diff
@@ -913,5 +913,36 @@ def evaluate_kde(kde, x):
...
+        ((47, 8), None),
+        ((47, 8), 0),
+        ((47, 8), 1),
+        ((10, 5, 21), None),
+        ((10, 5, 21), 0),
+        ((10, 5, 21), 1),
+        ((10, 5, 21), 2),
+      )
+      for contains_nans in [True, False]
+    ])
```
jakevdp, line 932: To avoid generating a list of all possible combinations, use something like this:
```python
@jtu.sample_product(
    [dict(shape=shape, axis=axis)
      for shape, axis in (
        ((0,), None),
        ((7,), None),
        ((7,), 0),
        ((47, 8), None),
        ((47, 8), 0),
        ((47, 8), 1),
        ((10, 5, 21), None),
        ((10, 5, 21), 0),
        ((10, 5, 21), 1),
        ((10, 5, 21), 2),
      )
    ],
    dtype=jtu.dtypes.integer + jtu.dtypes.floating,
    contains_nans=[True, False])
```

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,75 @@
...
+from jax._src.api import vmap
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.typing import Array, ArrayLike
+from jax._src.util import canonicalize_axis
+
+
+class ModeResult(NamedTuple):
+  mode: Array
+  count: Array
```
jakevdp, line 30: If this causes mypy errors, lets switch it back to how you originally had it defined.

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,75 @@
...
+
+@_wraps(scipy.stats.mode, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=['axis', 'nan_policy'])
+def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate") -> ModeResult:
+  _check_arraylike("mode", a)
+  x = jnp.asarray(a)
```
jakevdp, line 38: Maybe make this `x = jnp.atleast_1d(a)` instead of calling it further down.

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,75 @@
...
+
+  if nan_policy not in {"propagate", "omit", "raise"}:
+    raise ValueError(
+      f"Illegal nan_policy value {nan_policy!r}; expected one of "
+      "{'propoagate', 'omit', 'raise'}"
+    )
+  if nan_policy == "omit":
+    # TODO: return answer without nans included.
```
jakevdp, line 46: Is this TODO something you'd like to do as part of this PR, or leave it for later?

jakevdp (changes requested):
Sorry for the delayed review - this is looking great! Just a couple more minor comments.

jax/_src/scipy/stats/_core.py
ylamidon: If it's not required, maybe we can leave it for later

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,72 @@
...
+ModeResult = namedtuple('ModeResult', ('mode', 'count'))
+
+@_wraps(scipy.stats.mode, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=['axis', 'nan_policy'])
+def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate") -> ModeResult:
+  _check_arraylike("mode", a)
+  x = jnp.asarray(a)
+  x = jnp.atleast_1d(x)
```
jakevdp, line 37: `atleast_1d` calls `asarray` on the input, so no need to call both here. `x = jnp.atleast_1d(a)` should be sufficient.

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,72 @@
...
+
+  if nan_policy not in {"propagate", "omit", "raise"}:
```
jakevdp, line 39: Use a list rather than a set here, because if a user passes a non-hashable value for `nan_policy` this membership check will raise an error.

jax/_src/scipy/stats/_core.py
jakevdp: Sounds good

jax/_src/scipy/stats/_core.py
jakevdp: This indentation still needs to be fixed

jax/scipy/stats/__init__.py
```diff
@@ -33,3 +33,4 @@
 from jax.scipy.stats import betabinom as betabinom
 from jax.scipy.stats import gennorm as gennorm
 from jax._src.scipy.stats.kde import gaussian_kde as gaussian_kde
+from jax._src.scipy.stats._core import mode as mode
```
jakevdp, line 36:
```suggestion
from jax.scipy.stats import mode
```
`jax.scipy.stats.mode` is the public import path for this function, so we should use it in the tests.

ylamidon: I'm not sure to understand.
In the file  ```jax/scipy/stats/__init__.py```, the current import seems ok.
In the test file (```scipy_stats_test.py```), we do:
```
from jax.scipy import stats as lsp_stats
lax_fun = partial(lsp_stats.mode, axis=axis)
```
Which seems ok.

jakevdp: Sorry, you're right. I thought this was the test file

jakevdp (commented):
Looks great, thanks for sticking with this! Last two things before we can merge:

- the comment above about indentation: all indentations should be two spaces
- once that is fixed, please squash all changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Thanks!

jakevdp (approved):
Looks great! Thanks for the contribution

jakevdp:
It looks like we're hitting a `FutureWarning` in the scipy version of the function. I'd suggest we add a `keepdims` argument that defaults to `False` (i.e. the future default behavior for the `scipy` version) and expand the test cases to cover that. What do you think?

ylamidon:
I tried to add ```keepdims``` to the mode function with:
```
  if keepdims:
    return ModeResult(vals.reshape(output_shape), counts.reshape(output_shape))
  else:
    return ModeResult(jnp.squeeze(vals.reshape(output_shape), axis=axis), jnp.squeeze(counts.reshape(output_shape), axis=axis))
```
But this produces in a ```ConcretizationTypeError```

jakevdp:
Did you somehow convert `output_shape` or one of its elements to a jax array? You should make sure it is a tuple/list of integers.

jakevdp:
Seems like it may be more straightforward to change line 67, adjusting the `output_shape` depending on the value of `keepdims`.

jakevdp:
Oh, wait, I bet the issue is probably that you passed `keepdims` as a dynamic argument. It would have to be a static argument like `axis`. But the note about `out_shape` still holds.

ylamidon:
Yes, I forgot to add ```keepdims``` to ```static_argnames```

ylamidon:
When ```shape = (0,), axis=None``` and ```keepdims``` is ```True```  or ```False```,  numpy produces a RuntimeWarning: Mean of empty slice.
So I will probably have to remove the test with ```shape = (0,), axis=None```.

jakevdp:
Don't remove the test, instead filter the expected warnings (search the codebase for `jtu.ignore_warning` to see examples of this)

ylamidon:
```jtu.ignore_warning works``` fine.
For ```x = np.zeros((0,))```
```scipy.stats.mode(x, axis=None, keepdims=True)``` returns ```ModeResult(mode=array([nan]), count=array([nan]))```.
And
```jax.scipy.stats.mode(x, axis=None, keepdims=True)``` returns ```ModeResult(mode=DeviceArray([], dtype=float32), count=DeviceArray([], dtype=int32))```

**Comment by jakevdp** [comment]
It looks like we'll have to handle zero-size arrays differently to match the semantics of `scipy.stats.mode`; for example:
```python
>>> scipy.stats.mode(np.ones((0, 2, 3)), axis=2, keepdims=True).mode.shape                                                                                                                            
(0, 2, 1)
```

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,86 @@
...
+      "Please check if nans exist in `x` outside of the `mode` function."
+    )
+
+  if x.size == 0:
+    if axis is None:
+      if keepdims:
+        output_shape = [1 for _ in range(x.ndim)]
+        return ModeResult(jnp.full(output_shape, jnp.nan), jnp.full(output_shape, jnp.nan))
+      else:
+        return ModeResult(jnp.nan, jnp.nan)
```
jakevdp, line 60: `jnp.nan` is a python scalar `NaN`... for consistency we should return a jax array in all cases. We should use `jnp.full` in both cases... maybe compute the output_shape within the `if` statement and only have a single return statement? Maybe even one `return` statement for the entire `if x.size == 0` block? The only thing that's changing is the output shape, yes?

jakevdp: Also this is a bit in the weeds, but `jnp.full(shape, jnp.nan)` returns a weakly-typed array because `jnp.nan` is a Python scalar. For consistency with the non-empty return value we should use
```python
jnp.full(shape, jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_))
```

jax/_src/scipy/stats/_core.py
```diff
@@ -0,0 +1,86 @@
...
+    """Helper function to return mode and count of a given array."""
+    vals, counts = jnp.unique(x, return_counts=True, size=x.size)
+    return vals[jnp.argmax(counts)], counts.max()
+
+  axis = canonicalize_axis(axis, x.ndim)
+  output_shape = list(x.shape)
+  if keepdims:
+    output_shape[axis] = 1
+  else:
+    del output_shape[axis]
```
jakevdp, line 82: Does this result in the correct shape for `axis=None` with `keepdims=True`?

jax/_src/scipy/stats/_core.py
jakevdp: Another thought... this may all be cleaner if you handle the empty case in `_mode_helper` instead.

jakevdp (changes requested):

jax/_src/scipy/stats/_core.py
ylamidon: ```scipy.stats.mode(np.zeros((0,)), axis=None, keepdims=False)``` and ```scipy.stats.mode(np.zeros((0,)), axis=0, keepdims=False)``` both return ```ModeResult(mode=nan, count=nan)```

jakevdp: But they don't return a Python scalar NaN, they return a numpy.float64 NaN (i.e. a numpy scalar). JAX's equivalent of a numpy scalar is a zero-dimensional array.

jakevdp: This is important because, for example, someone might write code using something like `result.mode.shape`, and if we return a Python scalar NaN in some cases, that would result in an AttributeError. For that reason we generally try to keep return types consistent for different shape inputs. Does that make sense?

jax/_src/scipy/stats/_core.py
ylamidon: Yes, I think this part of the code is fine for axis=None with keepdims=True.

jax/_src/scipy/stats/_core.py
ylamidon: Yes it makes sense.
So instead of returning ```ModeResult(jnp.nan, jnp.nan)``` in this case, I can return  
```ModeResult(jnp.full((), jnp.nan), jnp.full((), jnp.nan))``` 
?

Comment (jakevdp) [reply jax/_src/scipy/stats/_core.py:60]: Yes, basically, except we should specify the dtype: `jnp.full(shape, jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_))`

Comment (ylamidon) [reply jax/_src/scipy/stats/_core.py:60]: To make the code a bit cleaner, I can refactor it like this:
```python
  def _mode_result_nan(output_shape):
    return ModeResult(
      jnp.full(output_shape, jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_)),
      jnp.full(output_shape, jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_)))

  def _get_output_shape(shape):
    output_shape = list(shape)
    if keepdims:
      output_shape[axis] = 1
    else:
      del output_shape[axis]
    return output_shape

  if x.size == 0:
    if axis is None:
      output_shape = [1 for _ in range(x.ndim)] if keepdims else ()
      return _mode_result_nan(output_shape)
    else:
      return _mode_result_nan(_get_output_shape(x.shape))
```

jakevdp: I'd avoid that kind of extra indirection in the form of one-line helper functions. I was just saying that you might define `output_shape` within the `if`/`else` blocks and then have a single return statement at the end. But if you prefer multiple return statements, that's fine too.

jakevdp: I think the best approach would be to handle the empty case within the vmapped helper, but we've already been going a long time here and I don't want to make you refactor what you have too much!

ylamidon: So i added a helper function and pushed the changes.   
I'm not sure that's exactly what you wanted for the helper function.

jakevdp: Sorry, I think I was unclear. I don't think we need all these extra helper functions for this algorithm: it's too much indirection and makes the code much harder to follow.

When I said "handle the empty case within the vmapped helper", I meant something like this:
```python
  def _mode_helper(x: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
    """Helper function to return mode and count of a given array."""
    if x.size == 0:
      return jnp.float32(jnp.nan), jnp.float32(jnp.nan)
    else:
      vals, counts = jnp.unique(x, return_counts=True, size=x.size)
      return vals[jnp.argmax(counts)], counts.max()
```
Then I think the zero-size case would be handled naturally within the `vmap` where this function is applied.

ylamidon: Yes, but then I have error in the following cases:
```python
shape=(0,), axis=None, keepdims=False
shape=(0,), axis=None, keepdims=True
shape=(0,), axis=0, keepdims=False
shape=(0,), axis=0, keepdims=True
shape=(0, 2, 3), axis=None, keepdims=False
shape=(0, 2, 3), axis=None, keepdims=True
shape=(0, 2, 3), axis=0, keepdims=False
shape=(0, 2, 3), axis=0, keepdims=True
```

jakevdp: Without seeing the error, I suspect it's coming from `x = x.reshape(x.shape[0], -1)` which is incompatible with zero-sized arrays. you can do `x = x.reshape(x.shape[0], prod(x.shape[1:]))` instead and that will work for arrays of any size

ylamidon: Yes, thanks. But I also have errors like:
```python
ValueError: vmap was requested to map its argument along axis 1, which implies that its rank should be at least 2, but is only 1 (its shape is (0,))
```

jakevdp: Hard to guess without seeing the code. It sound like you removed the `reshape` line if your array has only one dimension when you pass it to the vmapped function

ylamidon: Sorry, my mistake. There is another error:
```
jax._src.traceback_util.UnfilteredStackTrace: UnboundLocalError: local variable 'vals' referenced before assignment
```

jakevdp: Something like this should work:
```python
def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate", keepdims: bool = False) -> ModeResult:
  _check_arraylike("mode", a)
  x = jnp.atleast_1d(a)

  if nan_policy not in ["propagate", "omit", "raise"]:
    raise ValueError(
      f"Illegal nan_policy value {nan_policy!r}; expected one of "
      "{'propoagate', 'omit', 'raise'}"
    )
  if nan_policy == "omit":
    # TODO: return answer without nans included.
    raise NotImplementedError(
      f"Logic for `nan_policy` of {nan_policy} is not implemented"
    )
  if nan_policy == "raise":
    raise NotImplementedError(
      "In order to best JIT compile `mode`, we cannot know whether `x` contains nans. "
      "Please check if nans exist in `x` outside of the `mode` function."
    )

  input_shape = x.shape
  if keepdims:
    if axis is None:
      output_shape = tuple(1 for i in input_shape)
    else:
      output_shape = tuple(1 if i == axis else s for i, s in enumerate(input_shape))
  else:
    if axis is None:
      output_shape = ()
    else:
      output_shape = tuple(s for i, s in enumerate(input_shape) if i != axis)

  if axis is None:
    axis = 0
    x = x.ravel()

  def _mode_helper(x: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
    """Helper function to return mode and count of a given array."""
    if x.size == 0:
      return jnp.array(jnp.nan, dtype=x.dtype), jnp.float32(jnp.nan)
    else:
      vals, counts = jnp.unique(x, return_counts=True, size=x.size)
      return vals[jnp.argmax(counts)], counts.max()

  axis = canonicalize_axis(axis, x.ndim)
  x = jnp.moveaxis(x, axis, 0)
  x = x.reshape(x.shape[0], prod(x.shape[1:]))
  vals, counts = vmap(_mode_helper, in_axes=1)(x)
  return ModeResult(vals.reshape(output_shape), counts.reshape(output_shape))
```
It won't match scipy for `keepdims=True` and `axis=None`, but scipy is wrong in that case.

jakevdp: I just realized we can't return `jnp.array(jnp.nan, dtype=x.dtype)` because `x` may be integer typed. I'm not sure what's best in that case.

If you want to go back to what you had yesterday (before the helper function refactor) I'd be fine with that too. Your call

jakevdp: But we should make sure we *don't* match scipy for `keepdims=True` and `axis=None`. The return shape in that case should be `(1,) * len(input_shape)`

ylamidon: Just to clarify:
```python
scipy.stats.mode(np.ones((0, 3, 2)), axis=None, keepdims=True).mode.shape
```
returns ```(1, 1, 1)```.
And
```python
scipy.stats.mode(np.ones((4, 3, 2)), axis=None, keepdims=True).mode.shape
```
returns ```(1,)```.

So, ```jax.scipy.stats.mode``` should replicate ```scipy.stats.mode``` in both cases ?

jakevdp: Sorry, I mixed that up. I think it should return `(1, 1, 1)` in both cases. From https://github.com/scipy/scipy/issues/17214#issuecomment-1276464112 it's my understanding that this is the intended API.

ylamidon: Ok. 
But then the tests where x is not empty, ```ndim>=2 , axis=None, keepdims=True``` will fail because the shape returned by ```jax.scipy.stats.mode``` is not the same as the shaped returned by ```scipy.stats.mode```.
Is there a way to deactivate the test on the shape in these cases ?

ylamidon: Another thing. Shouldn't ```_mode_helper``` return 
```python
return jnp.array(jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_)), jnp.array(jnp.nan, dtype=dtypes.canonicalize_dtype(jnp.float_))
```
when ```x.size== 0```

jakevdp: > Is there a way to deactivate the test on the shape in these cases ?

I'd probably approach this by writing a wrapper around the scipy function that detects this case and returns the expected result.

> Shouldn't _mode_helper return...

Yes, you're right

ylamidon: Ok. I made the changes and pushed the result

jakevdp (approved):
Looks great! Thanks for sticking with this & sorry for so many review comments 😀 I'm going to trigger tests and hopefully this will all pass 🤞

jakevdp:
So, one last issue: `keepdims` wasn't added to scipy until version 1.9.0, and some of our test invocations still use this older version of scipy.

What we'll want to do is compute `scipy_version = tuple(map(int, scipy.__version__.split('.')[:3]))`, and then if `scipy_version < (1, 9, 0)` avoid passing `keepdims` to `scipy.stats.mode` (and probably just test `keepdims = True` for those versions, which is the default behavior for older scipy versions)

ylamidon:
I did the dev to avoid passing ```keepdims``` to ```scipy.stats.mode``` for ```scipy_version < (1, 9, 0)```.  

There is a small problem with the 1.7.3 version of scipy that I used for testing.  
* With the scipy version 1.9.2, ```scipy.stats.mode(a, keepdims=True)``` return an array of size 1 for an input array a of size 0
* With the scipy version 1.7.3, ```scipy.stats.mode(a) ``` return an array of size 0 for an input array a of size 0  

The problem is that I cannot reshape an array of size 0 to a shape of ```(1,) * ndim```.

So maybe I can use the ```scipy_mode_wrapper``` to return a pair of ```np.full(output_shape, np.nan, dtype=dtypes.canonicalize_dtype(jax.numpy.float_)``` instead of calling ```scipy.stats.mode ``` in this case ?

jakevdp:
Sure, that sounds like a good approach.

jakevdp (approved):

jakevdp:
Looks like there's still an error on scipy 1.5:
```
=================================== FAILURES ===================================
______________________ LaxBackedScipyStatsTests.testMode4 ______________________
[gw6] linux -- Python 3.7.13 /usr/bin/python3.7
tests/scipy_stats_test.py:949: in testMode
    tol=tol)
jax/_src/test_util.py:936: in _CheckAgainstNumpy
jax/_src/test_util.py:848: in assertAllClose
    err_msg=err_msg)
jax/_src/test_util.py:856: in assertAllClose
    err_msg=err_msg)
jax/_src/test_util.py:820: in assertArraysAllClose
jax/_src/public_test_util.py:96: in _assert_numpy_allclose
E   AssertionError:
E   Not equal to tolerance rtol=0.0002, atol=0.0002
E
E   x and y nan location mismatch:
E    x: array([[[-8.991903]]], dtype=float32)
E    y: array([[[nan]]], dtype=float32)
```

ylamidon:
Yes. Actually, its fine with scipy 1.5.4 and a new version of numpy. But there is an issue with scipy 1.5.4 and numpy 1.20.3.

jakevdp:
We might be able to bump the minimum scipy version to 1.7 - give me a bit to figure out if that's possible

jakevdp:
I think it's probably fine to skip the tests for scipy versions older than 1.6. We may not bump the minimum scipy right away, but it will happen soon-ish.

ylamidon:
I noticed that the tests are ok with numpy 1.21.0 and scipy 1.5.4 but not ok with numpy 1.20.0 and scipy 1.5.4.
The problem occurs when ```contains_nans = True```.
So maybe I can skip the tests with ```contains_nans = True``` when numpy version < 1.21.0 ?

jakevdp:
Sure, that sounds like a good plan.

jakevdp (approved):

jakevdp:
Whew, nice work on this! It ended up being quite a bit more complicated than anticipated with all the numpy/scipy version effects, thanks so much for sticking with it through the whole process!

ylamidon:
Thank you. Yes, it was a bit more complicated than expected. Thanks so much for all the help and valuable feedback!

PR #12693, opened 2022-10-07
Expand support for __jax_array__ in jnp.array.

gnecula:
This relates to the long discussion in #4725 and #10065.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1862,6 +1862,8 @@ def array(object: Any, dtype: Optional[DTypeLike] = None, copy: bool = True,
   if isinstance(object, (bool, int, float, complex)):
     _ = dtypes.coerce_to_array(object, dtype)
 
+  object = tree_map(lambda leaf: leaf.__jax_array__() if hasattr(leaf, "__jax_array__") else leaf,
+                    object)
   leaves = tree_leaves(object)
   if dtype is None:
     # Use lattice_result_type rather than result_type to avoid canonicalization.
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -3574,6 +3574,13 @@ class APITest(jtu.JaxTestCase):
     for f in [jnp.isscalar, jnp.size, jnp.shape, jnp.dtype]:
       self.assertEqual(f(x), f(a))
 
+    x = AlexArray(jnp.array(1))
+    a1 = jnp.array(x)
+    self.assertAllClose(1, a1)
+
+    a2 = jnp.array(((x, x), [x, x]))
+    self.assertAllClose(np.array(((1, 1), (1, 1))), a2)
+
   def test_constant_handler_mro(self):
     # https://github.com/google/jax/issues/6129
 


gnecula:
If this is acceptable then I can use the `__jax_array__` extension mechanism to duck type shape polymorphic dimensions as JAX arrays. This would be more load put on the experimental `__jax_array__` API, from an experimental API (shape polymorphism).

jakevdp:
I'd want to get @mattjj's input on this – I'm not clear on the plans for `__jax_array__`

mattjj:
I'd really like to delete `__jax_array__`, and replace it with a more systematic mechanism, as briefly outlined in [this comment](https://github.com/google/jax/issues/10065#issuecomment-1082489298). We may need to have a replacement ready first.

What are the alternatives to this PR, for accomplishing the jax2tf shape polymorphism goals?

gnecula:
The goal for shape polymorphism is to support the following when the shape of `x` contains polynomials
```
x / x.shape[0]
jnp.array(x.shape[0])
jnp.array(x.shape)
```

With the existing `__jax_array__` support I can already achieve the first of the above by defining `__jax_array__` for dimension polynomials. This kind of makes sense because without shape polymorphism `x.shape[0]` is a constant so users know they can use it in operations that can take JAX arrays.

This PR is specifically to allow cases 2 and 3 above.

Note that when the real dynamic shapes land in JAX, the polynomials disappear and their `__jax_array__` disappears with them, because `x.shape[0]` will be a tracer that can already be used in operations with JAX arrays.

If I cannot use `__jax_array__` then my plan B is to not support case 1 above, and instead document that users have to call `jax.array(x.shape[0])` explicitly. This is a bit awkward because they need to do this only if there is shape polymorphism. And then I plan to support cases 2 and 3 above by replacing the change in this PR with:

```
object = tree_map(lambda leaf: core.dimension_as_value(leaf) if core.is_special_dim(leaf) else leaf, object)
```

mattjj:
> Note that when the real dynamic shapes land in JAX, the polynomials disappear and their __jax_array__ disappears with them, because x.shape[0] will be a tracer that can already be used in operations with JAX arrays.

Maybe we're not too far from this. In fact, a lot might already work today! Can we draw up (or can you point me to) some test cases which define success here?

gnecula:
As a start, the test cases that I'd use are:
```
JAX_DYNAMIC_SHAPES=1 JAX2TF_DEFAULT_EXPERIMENTAL_NATIVE_LOWERING=1 JAX_ENABLE_X64=0 pytest --durations=10 -r a --verbosity 1 -s jax/experimental/jax2tf/tests/shape_poly_test.py
```

There are hundreds of tests. Maybe we should move this new thread elsewhere.

mattjj:
Thanks, @gnecula !

You're right that we should stay focused in this thread. FWIW the concrete changes in this PR are super minimal and so don't count as much tech debt on their own. My only concern would be if these small changes somehow generate a behavior that we'll then get locked into supporting (and which might be hard or undesirable to support in the future). For jax2tf, it's already marked as experimental so the risk is small. This change could in theory also increase others' reliance on the `__jax_array__` mechanism, though that's a bit of an abstract concern.

Overall, I think we're on the same page and we all have the same concerns/cautions in mind, so I'd say that if this small change helps advance your work let's just do it.

mattjj (approved):

gnecula:
I agree with you that this is a small change and to the extent that users will lean on it, it will be in a way that we plan to support in the future simply because dynamic shapes will have naturally the behavior that we are forcing here.

PR #12695, opened 2022-10-07
[jax2tf] Allow the use of DimPolynomial with jnp.array and binary operations

gnecula:
Prior to this the user had to explicitly call core.dimension_as_value whenever
using a potentially polymorphic shape in the computation, e.g., x +
core.dimension_as_value(x.shape[0]). Now, some of these operations
are allowed implicitly, or the user can call `jnp.array(x.shape[0])`.


gnecula:
@rxsang PTAL

gnecula:
For a discussion about the changes in `api.py` see #12693.

rxsang (commented):
Thanks!

PR #12566, opened 2022-09-28
[dynamic-shapes] basic slicing support with dynamic shape result

mattjj:
This PR also adds `__iter__` support for dynamically-shaped arrays.

PR #12697, opened 2022-10-07
[typing] annotate lax.slicing

jakevdp:
Part of #12049

### DIFF
--- a/jax/_src/lax/slicing.py
+++ b/jax/_src/lax/slicing.py
@@ -14,7 +14,7 @@
 
 import enum
 from functools import partial
-from typing import Any, Callable, NamedTuple, Optional, Sequence, Tuple, Union
+from typing import Callable, List, NamedTuple, Optional, Sequence, Tuple, Union
 import weakref
 
 import numpy as np
@@ -40,12 +40,10 @@ from jax._src.lib.mlir import ir
 from jax._src.lib.mlir.dialects import mhlo
 from jax._src.lib import xla_bridge
 from jax._src.lib import xla_client
+from jax._src.typing import Array, ArrayLike, Shape
 
 xb = xla_bridge
 xc = xla_client
-
-Array = Any
-Shape = core.Shape
 
 map, unsafe_map = safe_map, map
 zip, unsafe_zip = safe_zip, zip
@@ -53,7 +51,7 @@ zip, unsafe_zip = safe_zip, zip
 _dtype = partial(dtypes.dtype, canonicalize=True)
 
 
-def slice(operand: Array, start_indices: Sequence[int],
+def slice(operand: ArrayLike, start_indices: Sequence[int],
           limit_indices: Sequence[int],
           strides: Optional[Sequence[int]] = None) -> Array:
   """Wraps XLA's `Slice
@@ -64,7 +62,7 @@ def slice(operand: Array, start_indices: Sequence[int],
                       limit_indices=tuple(limit_indices),
                       strides=None if strides is None else tuple(strides))
 
-def dynamic_slice(operand: Array, start_indices: Sequence[Array],
+def dynamic_slice(operand: Array, start_indices: Union[Array, Sequence[ArrayLike]],
                   slice_sizes: Shape) -> Array:
   """Wraps XLA's `DynamicSlice
   <https://www.tensorflow.org/xla/operation_semantics#dynamicslice>`_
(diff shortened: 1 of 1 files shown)


jax/_src/lax/slicing.py
```diff
@@ -64,7 +62,7 @@ def slice(operand: Array, start_indices: Sequence[int],
                       limit_indices=tuple(limit_indices),
                       strides=None if strides is None else tuple(strides))
 
-def dynamic_slice(operand: Array, start_indices: Sequence[Array],
+def dynamic_slice(operand: Array, start_indices: Sequence[ArrayLike],
```
sharadmv, line 65: Should this also be `start_indices: Union[Array, Sequence[ArrayLike]]`?

jax/_src/lax/slicing.py
```diff
@@ -686,10 +684,10 @@ def index_in_dim(operand: Array, index: int, axis: int = 0,
     return lax.squeeze(result, (axis,))
 
 
-def dynamic_slice_in_dim(operand: Array, start_index: Array,
+def dynamic_slice_in_dim(operand: Array, start_index: ArrayLike,
                          slice_size: int, axis: int = 0) -> Array:
   """Convenience wrapper around dynamic_slice applying to one dimension."""
-  start_indices = [np.zeros((), dtype=dtypes.dtype(start_index))] * operand.ndim
+  start_indices = [lax._zero(start_index)] * operand.ndim
```
sharadmv, line 690: Any reason you need to change this back? I remember we added it in https://github.com/google/jax/pull/12566 to avoid some unenecessary staging.

sharadmv (changes requested):

jax/_src/lax/slicing.py
jakevdp: `ArrayLike` includes scalars which don't have `__len__`, so it would fail to typecheck. We could use `Union[Array, Sequence[ArrayLike]]`, what do you think?

jax/_src/lax/slicing.py
jakevdp: The problem is that `np.zeros` is of type `np.ndarray`, so this was leading to a typecheck error when an `Array` was assigned to the list below. I could figure out how to annotate this differently to keep the original if you think that would be better.

jax/_src/lax/slicing.py
```diff
@@ -222,7 +220,7 @@ def from_any(s: Optional[Union[str, 'GatherScatterMode']]):
       raise ValueError(f'Unknown gather mode "{s}"')
 
 
-def gather(operand: Array, start_indices: Array,
+def gather(operand: ArrayLike, start_indices: ArrayLike,
```
soraros, line 223: Aren't `operand` and `start_indices` always of type `Array`? Same question for `scatter` etc. IMO, it would be quite surprising otherwise

jax/_src/lax/slicing.py
sharadmv: I think that would be nice but interested in hearing from @mattjj

jax/_src/lax/slicing.py
sharadmv: SGTM

jax/_src/lax/slicing.py
jakevdp: `operand` here is passed to  `_dtype` and to `Primitive.bind`, neither of which require inputs of type `Array`, but do accept the union of types that we call `ArrayLike`.

soraros: I mean the doc string does mention `operand` being an array. And I would argue that the semantic of `gather` etc. seem to imply the first two arguments to be arrays (not scalars, as `jac.ArrayLike` already ruled out types where canonical projections into `ndarray` are defined). For instance, if I can make a call `gather(m, n, ...)` where `m` and `n` are integers, I'd say it's the interface being too forgiving.

Anyhow, thanks for explaining.

jakevdp: Well, `Array` would also rule-out `np.ndarray`, a valid input to this function, while `ArrayLike` does not.

It would be impossible in Python's current implementation to make type annotations exactly match the set of valid inputs; for example `jnp.array(0)` is of type `Array` but has the wrong shape. Similarly, the scalar value `np.int32(0)` is able to be converted implicitly to `Array` (which is why it is `ArrayLike`) but has the wrong shape.

There's no perfect type annotation, but all told I think `ArrayLike` is the correct annotation here.

jakevdp: By the way, the principle I'm following here is [inputs should be permissively typed](https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#inputs-should-be-permissively-typed), but I agree there's wiggle room and it's always a judgment call.

jax/_src/lax/slicing.py
jakevdp: Done

jax/_src/lax/slicing.py
jakevdp: Switched it back to a constant; also did the same in `dynamic_update_slice`

sharadmv (approved):
Thank you!

PR #12707, opened 2022-10-08
[dynamic-shapes] implement bint arrays (opaque dtypes), add padding rules

mattjj:
This was the last bit we needed to run a batch-shape-polymorphic transformer on XLA.


sharadmv (approved):

froystig:
For reference, the "opaque dtypes" mention in the PR title here is to say that this builds on #11768 and subsequent developments.

PR #12708, opened 2022-10-08
Upgrade the state primitive traceables to handle more forms of indexing

sharadmv:
This allows us to do things like
```python
x_ref[:32, :, :48]
# AND 
x_ref[jnp.arange(32)[:, None], jnp.arange(16)[None, :]]
```

cc: @jekbradbury

### DIFF
--- a/jax/_src/state/primitives.py
+++ b/jax/_src/state/primitives.py
@@ -21,18 +21,19 @@ from jax import core
 from jax import lax
 from jax._src import ad_util
 from jax._src import pretty_printer as pp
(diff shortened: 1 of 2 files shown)


mattjj (commented):
1. How about handling slices as early as possible, namely in the `_getitem` and `_setitem` of `ShapedArrayRef`? Then the traceables can stay closer to the primitives.
2. The tests are pretty minimal, but they probably need to be thorough, as the logic is complex. Luckily we have lots of indexing tests, in lax_numpy_indexing_test.py. Maybe try to use those somehow?

PR #12717, opened 2022-10-10
Migrate more tests from jtu.cases_from_list to jtu.sample_product.

hawkinsp:


tests/errors_test.py
```diff
@@ -380,10 +380,12 @@ def test_grad_norm(self):
 
 
 class CustomErrorsTest(jtu.JaxTestCase):
-  @parameterized.named_parameters(jtu.cases_from_list(
-    {"testcase_name": f"_{errorclass}", "errorclass": errorclass}
+  @jtu.sample_product(
+    [dict(errorclass=errorclass)
      for errorclass in dir(jax.errors)
-     if errorclass.endswith('Error') and errorclass not in ['JaxIndexError', 'JAXTypeError']))
+     if errorclass.endswith('Error') and errorclass not in ['JaxIndexError', 'JAXTypeError']
```
jakevdp, line 386: Just out of curiosity, why not use `errorclass=[errorclass for errorclass in ...]` rather than the dict form?

tests/fft_test.py
```diff
@@ -288,18 +286,17 @@ def testFftErrors(self, inverse, real, hermitian):
...
+    [dict(inverse=inverse, real=real, dtype=dtype)
+     for inverse in [False, True]
+     for real in [False, True]
+     for dtype in (real_dtypes if real and not inverse else all_dtypes)
+    ],
+    shape=[(16, 8, 4, 8), (16, 8, 4, 8, 4)],
+    axes=[(-2, -1), (0, 1), (1, 3), (-1, 2)],
+    norm=FFT_NORMS,
+  )
+  def testFft2_(self, inverse, real, shape, dtype, axes, norm):
```
jakevdp, line 299: Why the trailing underscore?

tests/image_test.py
```diff
@@ -52,19 +50,15 @@
...
-       for target_shape, image_shape in itertools.combinations_with_replacement(
-        [[2, 3, 2, 4], [2, 6, 4, 4], [2, 33, 17, 4], [2, 50, 38, 4]], 2)
-       for method in ["nearest", "bilinear", "lanczos3", "lanczos5", "bicubic"]
-       for antialias in [False, True]))
+  _TF_SHAPES = [[2, 3, 2, 4], [2, 6, 4, 4], [2, 33, 17, 4], [2, 50, 38, 4]]
+
+  @jtu.sample_product(
+    dtype=float_dtypes,
+    target_shape=_TF_SHAPES,
+    image_shape=_TF_SHAPES,
```
jakevdp, line 58: Just noting: this iterates over the product of shapes rather than combinations, which I think is OK? But it's a different set of test cases than before.

I think it's fine.

tests/qdwh_test.py
```diff
@@ -129,14 +124,11 @@ def testQdwhWithUpperTriangularInputAllOnes(self, m, n, log_cond):
...
-  @parameterized.named_parameters(jtu.cases_from_list(
-      {  # pylint:disable=g-complex-comprehension
-          'testcase_name': '_m={}_by_n={}_log_cond={}_padding={}'.format(
-              m, n, log_cond, padding),
-          'm': m, 'n': n, 'log_cond': log_cond, 'padding': padding}
-      for m, n in zip([6, 8], [6, 4])
-      for padding in (None, (3, 2))
-      for log_cond in np.linspace(1, 4, 4)))
+  @jtu.sample_product(
+    [dict(m=m, n=n) for m, n in zip([6, 8], [6, 4])],
```
jakevdp, line 128: nit: `[{'m': 6, 'n': 6}, {'m': 8, 'n': 4}]` would be shorter and more direct

jakevdp (approved):
Nice! Some minor comments (take them or leave them 😁 )

tests/fft_test.py
hawkinsp: Because the test case names are formed by suffixing numbers, so there's a clash between `testFft`'s second test case and `testFft2`'s first one.

tests/image_test.py
hawkinsp: I think the intent before was to produce the full product! I think it was actually a bug to use combinations_with_replacement.

tests/qdwh_test.py
hawkinsp: Changed to avoid the `zip`, but I find it verbose to write out the names for each shape, especially if we ever add more test cases.

PR #12746, opened 2022-10-11
changelog: add missing github commit links

jakevdp:


mattjj (approved):
Thanks for catching this!

PR #12748, opened 2022-10-11
[ROCm] Fixed build instructions typo

i-chaochen:
Fixed the built jax image name in the instruction. Now it is same name as the image from the build_rocm.sh. 

@hawkinsp


mattjj (approved):
Thank you!

PR #12749, opened 2022-10-11
Port more tests from jtu.cases_from_list to jtu.sample_product.

hawkinsp:


yashk2810 (approved):

hawkinsp:
Interestingly the change in the test cases sampled revealed an interesting problem: it isn't reasonable to expect a small epsilon change to the input of QR decomposition to produce a small output. It seems that sometimes you might end up flipping some of the Q vectors around. I changed that test to compute the JVP of the QR decomposition composed with multiplying the result back together again, which should make it robust to Q direction changes.

PR #12754, opened 2022-10-11
[typing] annotate jax.scipy.signal & jax.scipy.special

jakevdp:
Part of #12049

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 from functools import partial
-from typing import Any, Optional, Tuple
+from typing import cast, overload, Any, List, Optional, Tuple, Union
 
 import numpy as np
 import scipy.special as osp_special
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/special.py
```diff
@@ -1303,14 +1320,16 @@ def _expi_pos(x):
     [_expint1, _expint2, _expint3, _expint4, _expint5, _expint6, _expint7],
   )
 
+def _expi_neg(x: Array) -> Array:
```
jakevdp, line 1323: Made this an explicit function both for ease of typing, and for better jit cacheing

mattjj (approved):

yilei:
jFYI-- https://github.com/google/jax/blob/503679d63610d985e562471d710b2b227eea5e23/jax/_src/third_party/scipy/signal_helper.py#L54-L57

If none of the `if/elif` matches, `nperseg_int` won't be defined here causing a `NameError` when this function returns.

PR #12760, opened 2022-10-11
Enable partially discharging state effects from jaxprs

sharadmv:
Previously, we might have a jaxpr like:
```python
{ lambda ; a:Ref{float32[4]} b:Ref{float32[4]}. let
    c:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
    a[:] <- c
    d:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
    b[:] <- d
  in () }
```
and when we discharge state we get:
```python
{ lambda ; a:f32[4] b:f32[4]. let
    c:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
    d:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
  in (c, d) }
```
This removes *all* state effects from the jaxpr.

However, in some applications, we might want to *partially* discharge the state, i.e. remove the state effects for a subset of the refs. This PR allows that.

Take the previous stateful jaxpr and let's see what happens when we only discharge the state effect for `b`:
```python
{ lambda ; a:Ref{float32[4]} b:f32[4]. let
    c:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
    a[:] <- c
    d:f32[4] = broadcast_in_dim[broadcast_dimensions=() shape=(4,)] 1.0
  in (d,) }
```

### DIFF
--- a/jax/_src/state/discharge.py
+++ b/jax/_src/state/discharge.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import dataclasses
 from functools import partial
 
-from typing import Any, Dict, List, Optional, Sequence, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 from typing_extensions import Protocol
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


mattjj (approved):
Nice!

PR #12766, opened 2022-10-12
Migrate remaining tests from jtu.cases_from_list to jtu.sample_product.

hawkinsp:
Delete jtu.cases_from_list.

### DIFF
--- a/tests/lax_autodiff_test.py
+++ b/tests/lax_autodiff_test.py
@@ -133,7 +133,7 @@ LAX_GRAD_OPS = [
     grad_test_spec(lax.rsqrt, nargs=1, order=2, rng_factory=jtu.rand_default,
                    dtypes=grad_complex_dtypes),
     grad_test_spec(lax.cbrt, nargs=1, order=2, rng_factory=jtu.rand_default,
-                   dtypes=grad_float_dtypes, tol={np.float64: 3e-5}),
+                   dtypes=grad_float_dtypes, tol={np.float64: 5e-3}),
     grad_test_spec(lax.logistic, nargs=1, order=2,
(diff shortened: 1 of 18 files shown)


tests/lax_autodiff_test.py
```diff
@@ -195,38 +195,43 @@ def check_grads_bilinear(f, args, order,
...
-  @parameterized.named_parameters(itertools.chain.from_iterable(
-      jtu.cases_from_list(
-        {"testcase_name": jtu.format_test_name_suffix(
-            rec.name, shapes, itertools.repeat(dtype)),
-         "op": rec.op, "rng_factory": rec.rng_factory, "shapes": shapes, "dtype": dtype,
-         "order": rec.order, "tol": rec.tol}
-        for shape_group in compatible_shapes
+  @parameterized.parameters(itertools.chain.from_iterable(
+    jtu.sample_product_testcases(
+      [dict(op=rec.op, rng_factory=rec.rng_factory, order=rec.order, tol=rec.tol)],
```
jakevdp, line 200: could we just pass `rec=[rec]` here instead and pull out its contents in the test function? I think the reason we have split this in the past is because `rec` doesn't format well in the name, but now that we're not generating any name we could simplify the parametrization. What do you think?

jakevdp (approved):
All looks reasonable. Thanks!

tests/lax_autodiff_test.py
hawkinsp: I'm going to punt this to a future PR. If it makes sense to do here, it probably makes sense to do in a number of other tests that have a similar pattern.

PR #12769, opened 2022-10-12
Migrate JAX internals to builtin Python logging

### LINKED ISSUE #6308: Migrate away from absl logging.

nicholasjng:
This commit changes the JAX codebase to use Python's builtin logging instead of ABSL logging. With the latter being used in JAX code as of now, the change to Python builtin logging is advised for the following reasons (among others):

- `absl-py` can be removed as an external dependency of JAX.
- Builtin logging brings the option of adding more log handlers, for example file handlers for log dumps or writers to different IO streams.

Logging in JAX is ported over to take place at the module level. While previously, some Python namespaces within JAX already used module-scoped logging via `absl.vlog`, the following idiom was adopted to provide the same functionality in Python builtin logging:

```
logger = get_logger(__name__)
logger.debug(...)
logger.info(...)
```

This way, all log messages emitted from JAX contain their origin information directly within the logger name. Furthermore, the builtin root logger is left untouched, which is beneficial for downstream users planning to customize the Python root logger.

The package `absl-py` was removed from JAX's install requirements, and added into its test requirements. Tests are not yet migrated in this change.

Fixes #6308.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -18,18 +18,19 @@
 import contextlib
 import functools
 import itertools
+import logging
 import os
 import sys
 import threading
 from typing import Any, List, Callable, Hashable, NamedTuple, Iterator, Optional
-import warnings
-
-from absl import logging
 
 from jax._src import lib
 from jax._src.lib import jax_jit
 from jax._src.lib import transfer_guard_lib
 from jax._src.lib import xla_client
+
+logger = logging.getLogger(__name__)
+
 
 def bool_env(varname: str, default: bool) -> bool:
   """Read an environment variable and interpret it as a boolean.
@@ -643,7 +644,7 @@ log_compiles = config.define_bool_state(
     name='jax_log_compiles',
     default=False,
     help=('Log a message each time every time `jit` or `pmap` compiles an XLA '
-          'computation. Logging is performed with `absl.logging`. When this '
+          'computation. Logging is performed with `logging`. When this '
           'option is set, the log level is WARNING; otherwise the level is '
           'DEBUG.'))
 
@@ -674,7 +675,7 @@ distributed_debug = config.define_bool_state(
     name='jax_distributed_debug',
     default=False,
     help=('Enable logging useful for debugging multi-process distributed '
-          'computations. Logging is performed with `absl.logging` at WARNING '
+          'computations. Logging is performed with `logging` at WARNING '
           'level.'))
 
 
@@ -772,7 +773,7 @@ def _validate_default_device(val):
     # TODO(skyewm): this is a workaround for non-PJRT Device types. Remove when
     # all JAX backends use a single C++ device interface.
     if 'Device' in str(type(val)):
-      logging.info(
+      logger.info(
           'Allowing non-`xla_client.Device` default device: %s, type: %s',
           repr(val), type(val))
       return
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -20,15 +20,12 @@ XLA. There are also a handful of related casting utilities.
 """
 
 from functools import partial, lru_cache
+import logging
 import os
 import platform as py_platform
 import threading
 from typing import Any, Dict, List, Optional, Union
 import warnings
-
-from absl import logging
-# Disable "WARNING: Logging before flag parsing goes to stderr." message
-logging._warn_preinit_stderr = 0
 
 import jax._src.lib as lib
 from jax._src.config import flags, bool_env, int_env
@@ -57,6 +54,8 @@ use_sharded_buffer = xla_client._version >= 90
 ShardedBuffer = Any
 
 FLAGS = flags.FLAGS
+
+logger = logging.getLogger(__name__)
 
 # TODO(phawkins): Remove jax_xla_backend.
 flags.DEFINE_string(
@@ -126,8 +125,7 @@ def get_compile_options(
     build_options.auto_spmd_partitioning_mesh_shape = auto_spmd_partitioning_mesh_shape
     build_options.auto_spmd_partitioning_mesh_ids = auto_spmd_partitioning_mesh_ids
   if device_assignment is not None:
-    logging.vlog(
-        2,
+    logger.debug(
         'get_compile_options: num_replicas=%s num_partitions=%s device_assignment=%s',
         num_replicas, num_partitions, device_assignment)
     device_assignment = np.array(device_assignment)
@@ -170,10 +168,10 @@ def get_compile_options(
 
 def _make_tpu_driver_client():
   if tpu_driver_client is None:
-    logging.info("Remote TPU is not linked into jax; skipping remote TPU.")
+    logger.info("Remote TPU is not linked into jax; skipping remote TPU.")
     return None
   if FLAGS.jax_backend_target is None:
-    logging.info("No --jax_backend_target was provided; skipping remote TPU.")
+    logger.info("No --jax_backend_target was provided; skipping remote TPU.")
     return None
   return tpu_driver_client.TpuBackend.create(worker=FLAGS.jax_backend_target)
 
@@ -354,14 +352,14 @@ def backends():
             raise RuntimeError(err_msg)
           else:
             _backends_errors[platform] = str(err)
-            logging.info(err_msg)
+            logger.info(err_msg)
             continue
     # We don't warn about falling back to CPU on Mac OS, because we don't
     # support anything else there at the moment and warning would be pointless.
     if (py_platform.system() != "Darwin" and
         _default_backend.platform == "cpu" and
         FLAGS.jax_platform_name != 'cpu'):
-      logging.warning('No GPU/TPU found, falling back to CPU. '
+      logger.warning('No GPU/TPU found, falling back to CPU. '
                       '(Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)')
     return _backends
 
@@ -371,7 +369,7 @@ def _clear_backends():
   global _backends_errors
   global _default_backend
 
-  logging.info("Clearing JAX backend caches.")
+  logger.info("Clearing JAX backend caches.")
   with _backend_lock:
     _backends = {}
     _backends_errors = {}
@@ -385,7 +383,7 @@ def _init_backend(platform):
   if factory is None:
     raise RuntimeError(f"Unknown backend '{platform}'")
 
-  logging.vlog(1, "Initializing backend '%s'" % platform)
+  logger.debug("Initializing backend '%s'", platform)
   backend = factory()
   # TODO(skye): consider raising more descriptive errors directly from backend
   # factories instead of returning None.
@@ -397,7 +395,7 @@ def _init_backend(platform):
                              ("process_index", backend.process_index()),
                              ("device_count", backend.device_count()),
                              ("local_devices", backend.local_devices()))
-  logging.vlog(1, "Backend '%s' initialized" % platform)
+  logger.debug("Backend '%s' initialized", platform)
   return backend
 
 
--- a/jax/_src/util.py
+++ b/jax/_src/util.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+from collections import namedtuple
 import functools
 from functools import partial
 import itertools as it
-from collections import namedtuple
+import logging
 import operator
 import types
 import threading
@@ -24,12 +24,13 @@ from typing import (Any, Callable, Dict, Iterable, List, Tuple, Generic,
                     TypeVar, Set, Iterator, Sequence, Optional)
 import weakref
 
-from absl import logging
 import numpy as np
 
 from jax._src.lib import xla_client as xc
 from jax._src.lib import xla_extension_version
 from jax.config import config
+
+logger = logging.getLogger(__name__)
 
 Seq = Sequence
 
@@ -510,7 +511,7 @@ def distributed_debug_log(*pairs):
       lines.append("DISTRIBUTED_DEBUG logging failed!")
       lines.append(f"{e}")
     lines.append("DISTRIBUTED_DEBUG_END")
-    logging.warning("\n".join(lines))
+    logger.warning("\n".join(lines))
 
 
 class OrderedSet(Generic[T]):
--- a/jax/experimental/host_callback.py
+++ b/jax/experimental/host_callback.py
@@ -497,13 +497,12 @@ Still to do:
 import atexit
 import functools
 import itertools
+import logging
 import threading
 import traceback
 from typing import (Any, Callable, Dict, List, Optional, Sequence,
                     Tuple, cast)
 import warnings
-
-from absl import logging
 
 from jax._src import api
 from jax import core
@@ -531,6 +530,8 @@ import numpy as np
 
 
 FLAGS = config.FLAGS
+
+logger = logging.getLogger(__name__)
 
 
 def _inline_host_callback() -> bool:
@@ -1231,9 +1232,10 @@ def _outside_call_run_callback(
   try:
     arg = api.tree_unflatten(arg_treedef, arrays)
     unpacked_transforms = _unpack_transforms(transforms)
-    if logging.vlog_is_on(2):
-      logging.vlog(2,
-                   f"Outside call invoking call_func {callback}, device={device}, transforms={unpacked_transforms}")
+    logger.debug(
+      "Outside call invoking call_func %s, device=%s, transforms=%s",
+      callback, device, unpacked_transforms
+    )
     res = callback(arg, device, unpacked_transforms)
     if identity:
       return tuple(arrays)
@@ -1250,10 +1252,9 @@ def _outside_call_run_callback(
 
       canonical_flat_results = tuple(util.safe_map(xla.canonicalize_dtype, actual_flat_results))
       actual_flat_results_aval = _values_to_avals(canonical_flat_results)
-      if logging.vlog_is_on(2):
-        logging.vlog(
-            2,
-            f"Outside call {callback} result {flat_results_aval}. Sending to infeed for device {device}."
+      logger.debug(
+        "Outside call %s result %s. Sending to infeed for device %s.",
+        callback, flat_results_aval, device,
         )
 
       if not all(ea.strip_weak_type() == ra.strip_weak_type()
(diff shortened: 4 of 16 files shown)


jax/_src/lib/xla_bridge.py
```diff
@@ -126,8 +124,7 @@ def get_compile_options(
     build_options.auto_spmd_partitioning_mesh_shape = auto_spmd_partitioning_mesh_shape
     build_options.auto_spmd_partitioning_mesh_ids = auto_spmd_partitioning_mesh_ids
   if device_assignment is not None:
-    logging.vlog(
```
hawkinsp, line 129: `logger.debug`

Anything that was a vlog should turn a debug log, as a rule.

jax/experimental/host_callback.py
```diff
@@ -1231,9 +1232,9 @@ def _unpack_transform(name, *params):
   try:
     arg = api.tree_unflatten(arg_treedef, arrays)
     unpacked_transforms = _unpack_transforms(transforms)
-    if logging.vlog_is_on(2):
-      logging.vlog(2,
-                   f"Outside call invoking call_func {callback}, device={device}, transforms={unpacked_transforms}")
+    logger.warning(
```
hawkinsp, line 1235: logger.debug

jax/experimental/host_callback.py
```diff
@@ -1250,9 +1251,7 @@ def _unpack_transform(name, *params):
 
       canonical_flat_results = tuple(util.safe_map(xla.canonicalize_dtype, actual_flat_results))
       actual_flat_results_aval = _values_to_avals(canonical_flat_results)
-      if logging.vlog_is_on(2):
-        logging.vlog(
-            2,
+      logger.warning(
```
hawkinsp, line 1254: logger.debug

jax/experimental/host_callback.py
```diff
@@ -1285,12 +1284,10 @@ def _unpack_transform(name, *params):
       # TODO: implement a proper error handling for TPU
       if device.platform != "tpu":
         canonical_flat_results = [xla.canonicalize_dtype(np.arange(12345, dtype=np.int8))]
-        if logging.vlog_is_on(2):
-          logging.vlog(2, f"Outside call consumer {callback} exception {e}. Sending to infeed the error result.")
+        logger.error(f"Outside call consumer {callback} exception {e}. Sending to infeed the error result.")
```
hawkinsp, line 1287: logger.debug

jax/experimental/host_callback.py
```diff
@@ -1285,12 +1284,10 @@ def _unpack_transform(name, *params):
...
         device.transfer_to_infeed(tuple(canonical_flat_results))
       else:
-        if logging.vlog_is_on(2):
-          logging.vlog(2, f"Outside call consumer {callback} exception {e}. On TPU we do not send infeed.")
+        logger.error(f"Outside call consumer {callback} exception {e}. On TPU we do not send infeed.")
```
hawkinsp, line 1290: logger.debug

jax/experimental/host_callback.py
```diff
@@ -1860,8 +1857,7 @@ def stop(self):
 
 # This function is called from C++; it must not allow exceptions through.
 def _callback_input_received(device, consumer_id, arrays: Tuple):
-  logging.vlog(
-      2,
+  logger.warning(
```
hawkinsp, line 1860: logger.debug

jax/experimental/host_callback.py
```diff
@@ -1920,9 +1916,8 @@ def _initialize_outfeed_receiver(
     if clients_with_outfeed:
       devices_with_outfeed = list(
         itertools.chain(*[backend.local_devices() for backend in clients_with_outfeed]))
-      if logging.vlog_is_on(2):
-        logging.vlog(
-            2,
+      if logger.isEnabledFor(logging.WARNING):
```
hawkinsp, line 1919: logger.debug and  debug priority.

Same everywhere in this file.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -636,7 +636,7 @@ def _lower_native_and_run(fun_jax: Callable,
   lowered = fun_jax_lower(*arg_specs_jax)._lowering
   mhlo_module = lowered.mhlo()
   mhlo_module_text = mlir.module_to_string(mhlo_module)
-  logging.vlog(2, f"XlaCallModule {mhlo_module_text}")
+  logger.warning(f"XlaCallModule {mhlo_module_text}")
```
hawkinsp, line 639: logger.debug

Also, string interpolation should be done via an argument, not an f-string. The point is you don't want to do the interpolation if the log isn't triggered.

jax/interpreters/pxla.py
```diff
@@ -1385,19 +1387,19 @@ def lower_parallel_callable(
   jaxpr, consts, replicas, parts, shards = stage_parallel_callable(
       pci, fun, global_arg_shapes)
 
-  if logging.vlog_is_on(2):
-    logging.vlog(2, "sharded_avals: %s", shards.sharded_avals)
-    logging.vlog(2, "global_sharded_avals: %s", shards.global_sharded_avals)
-    logging.vlog(2, "num_replicas: %d  num_local_replicas: %d",
+  if logger.isEnabledFor(logging.WARNING):
```
hawkinsp, line 1390: logger.debug

jax/_src/util.py
```diff
@@ -493,6 +493,16 @@ def maybe_named_axis(axis, if_pos, if_named):
     named = True
   return if_named(axis) if named else if_pos(pos)
 
+def get_logger(name: Optional[str] = None, level: int = logging.INFO) -> logging.Logger:
+  logger = logging.getLogger(name=name)
+  # TODO: Enable file handler attachments and level setting from config
+  stream = logging.StreamHandler()
```
hawkinsp, line 499: I think you should not add a handler. The behavior you're going to get if you don't do this is that the messages will be propagated to the root logger. That's the behavior we want! In particular, we want to interoperate with absl logging, and absl logging installs an alternative handler on the root logger. We do not want to force things to go to stderr for example if the root logger isn't configured that way.

Just remove the handler code.

hawkinsp (changes requested):

hawkinsp:
> Tests are not yet migrated in this change.

There's no point migrating the tests anyway really, since the tests have a hard absl-py dependency for other reasons anyway (e.g., they use absltest). That's fine, they are tests after all.

nicholasjng:
Thanks for the comments, I'll update the vlog situation and the commit message accordingly.

What do you think about adding optional config flags for log dumps (similar to XLA dumps via XLA_FLAGS) and setting the Python logging verbosity inside `get_logger`? I didn't include them yet, but it might be an easy way to influence JAX-wide logging in a single setting.

hawkinsp:
Well, I'm not necessarily opposed to adding a config flag, but it would have to improve over:
```
import logging
logging.getLogger("jax").setLevel(logging.DEBUG)
```

which is pretty good as is. So I'd be inclined to err on the side of not adding anything for the moment.

hawkinsp:
I think I'd be tempted to promise/document that JAX will log under the "jax" logger (and nested loggers) and let everyone use the Python APIs as they see fit.

nicholasjng:
Changes after your comments:

- Set `vlog` -> `logging.debug` everywhere, with `logging.isEnabledFor(logging.DEBUG)` priority.
- Changed logs to deferred string interpolation with `%s` style.
- Changed the `jax.util.get_logger` function to return a single logger named `"jax"`, instead of dynamically-named loggers.
- Removed formatter code, enabled logging.DEBUG priority.
- Reworked your etils PR from yesterday to use the `jax.util.get_logger` method.
- Updated the commit message to not mention test migration anymore.

jax/util.py
```diff
@@ -16,6 +16,7 @@
   HashableFunction as HashableFunction,
   as_hashable_function as as_hashable_function,
   cache as cache,
+  get_logger as get_logger,
```
hawkinsp, line 19: This should not be a public API. It serves no additional value over logging.getLogger("jax"). Just revert this change?

jax/_src/config.py
```diff
@@ -18,19 +18,21 @@
...
-
-from absl import logging
 
 from jax._src import lib
 from jax._src.lib import jax_jit
 from jax._src.lib import transfer_guard_lib
 from jax._src.lib import xla_client
 
+logger = logging.getLogger("jax")
+logger.setLevel(logging.DEBUG)
```
hawkinsp, line 33: Remove the code to change the logger level.

jax/_src/util.py
```diff
@@ -493,6 +492,18 @@ def maybe_named_axis(axis, if_pos, if_named):
...
   return if_named(axis) if named else if_pos(pos)
 
+def get_logger() -> logging.Logger:
+  """
+  Instantiate a global logger named "jax" as a direct descendant of the root logger.
+  JAX-internal code only logs messages on this logger, by default at DEBUG priority.
+  """
+
+  jax_logger = logging.getLogger("jax")
+  jax_logger.setLevel(logging.DEBUG)
```
hawkinsp, line 502: Remove the code to set the logger level.

There's no need to override it. If we do nothing, then logs get propagated to the root logger (i.e., NOTSET, see the docs). The root logger by default prints to stderr everything at priority WARNING or above, and that's exactly the behavior we want. We don't want to see debug logs by default.

jax/_src/util.py
```diff
@@ -493,6 +492,18 @@ def maybe_named_axis(axis, if_pos, if_named):
     named = True
   return if_named(axis) if named else if_pos(pos)
 
+def get_logger() -> logging.Logger:
```
hawkinsp, line 495: It's not clear this function serves any purpose over
logger = logging.getLogger(__name__)

Just inline it into its users? It would also solve the circular import problem.

hawkinsp (changes requested):
I'm not sure why you moved away from using `__name__` as the logger name. I think that's the right thing to do, e.g., see https://docs.python-guide.org/writing/logging/#logging-in-a-library

Note that loggers form a hierarchy. If you create a "jax._src.something" logger, then it propagates up the hierarchy to the "jax._src" and "jax" loggers. So if you use `__name__`, then the "jax" logger sees everything from jax, which is what we want.

We also do not want to put everything in one logger because it is very helpful to be able to selectively enable what were formerly the `vlog` messages. They are annoyingly verbose, you never want to see them all, but sometimes you really want to see all the logs from one module when debugging something.

The argument against using `__name__` is it exposes too much of the module structure of JAX. I think that's an acceptable tradeoff for the moment. All we're promising is that jax logs will be under the "jax" logger, nothing more. We could offer more structured subloggers as well in the future, but right now we only promise that top-level name.

jax/_src/util.py
hawkinsp: I think yes to the `getLogger` question.

I think no to the `NullHandler` question. I think the default behavior is the correct one. We do want these messages printed in the absence of any logging configuration (https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

jax/_src/lib/xla_bridge.py
```diff
@@ -58,6 +54,8 @@
 
 FLAGS = flags.FLAGS
 
+logger = util.get_logger()
```
hawkinsp, line 57: This one needs fixing still.

hawkinsp (changes requested):

jax/_src/lib/xla_bridge.py
nicholasjng: I see, shouldn't flake8/mypy have caught this? It was all green on my local machine, mysterious.

nicholasjng:
Thanks. I had a major misunderstanding about how `logging` works, especially about the propagation situation. I think I cleaned it up now - in particular I reverted my change to #12780, since that was the desired solution after all.

Also caught two more unconditional string interpolations along the way.

jax/experimental/host_callback.py
```diff
@@ -1285,12 +1286,12 @@ def _unpack_transform(name, *params):
       # TODO: implement a proper error handling for TPU
       if device.platform != "tpu":
         canonical_flat_results = [xla.canonicalize_dtype(np.arange(12345, dtype=np.int8))]
-        if logging.vlog_is_on(2):
-          logging.vlog(2, f"Outside call consumer {callback} exception {e}. Sending to infeed the error result.")
+        logger.debug("Outside call consumer %s exception %s. Sending to infeed the error result.",
+                     callback, e)
```
nicholasjng, line 1290: Shouldn't this be a `logger.error`, since we are logging exceptions here? Same below.

jax/_src/lib/xla_bridge.py
nicholasjng: Fixed.

jax/experimental/host_callback.py
hawkinsp: I think I'd leave it as is, because this is probably not intended to be seen by the user: note we reraise the exception below.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
       for this invocation. See `Debugging` in the module documentation.
   """
   logging_name = logging_name or ""
-  if logging.vlog_is_on(2):
-    logging.vlog(2, f"barrier_wait[{logging_name}]: start")
+  if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1961: There's no need for the `isEnabledFor` guard here because you fixed the string interpolation! So the debug call does no unnecessary work and we can just call it unconditionally.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
...
   cv = threading.Condition(lock=lock)
   devices_at_barrier = []  # Protected by lock
   def barrier_tap_received(dev_idx, _):
     device = _callback_handler_data.devices[dev_idx]
-    if logging.vlog_is_on(2):
-      logging.vlog(
-          2,
-          f"barrier_wait[{logging_name}]: at barrier_tap for device {device} "
-          f". Thread {threading.current_thread()}")
+    if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1969: Same here.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
...
+          "barrier_wait[%s]: still waiting for %s devices at barrier (%s)",
+          logging_name, len(waiting_for_devices), waiting_for_devices
+        )
       cv.notify()
 
   for d_idx, d in enumerate(_callback_handler_data.devices):
-    if logging.vlog_is_on(2):
-      logging.vlog(2,
-                   f"barrier_wait[{logging_name}]: enqueueing barrier on device {d}")
+    if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1986: Same here.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
...
+      logger.debug("barrier_wait[%s]: enqueueing barrier on device %s", logging_name, d)
     x_on_dev = api.device_put(d_idx, device=d)
     api.jit(lambda x: id_tap(barrier_tap_received, x), device=d)(x_on_dev)
-  if logging.vlog_is_on(2):
-    logging.vlog(2,
-                 f"barrier_wait[{logging_name}]: waiting for callbacks")
+  if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1990: Same here.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
...
+    logger.debug("barrier_wait[%s]: waiting for callbacks", logging_name)
   with lock:
     cv.wait_for(lambda: len(devices_at_barrier) == len(_callback_handler_data.devices))
-  if logging.vlog_is_on(2):
-    logging.vlog(2, f"barrier_wait[{logging_name}]: done")
+  if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1994: Same here.

jax/experimental/host_callback.py
```diff
@@ -1959,43 +1958,41 @@ def barrier_wait(logging_name: Optional[str] = None):
...
-          f". Thread {threading.current_thread()}")
+    if logger.isEnabledFor(logging.DEBUG):
+      logger.debug(
+        "barrier_wait[%s]: at barrier_tap for device %s. Thread %s",
+        logging_name, device, threading.current_thread()
+      )
     with lock:
       devices_at_barrier.append(device)
-      if logging.vlog_is_on(2):
+      if logger.isEnabledFor(logging.DEBUG):
```
hawkinsp, line 1976: However note this one still is useful: it guards the computation of `waiting_for_devices`.

hawkinsp (changes requested):
A couple of minor nits, but this looks pretty good to me!

jax/experimental/host_callback.py
nicholasjng: Log guard (sic) removed.

jax/experimental/host_callback.py
nicholasjng: Removed.

jax/experimental/host_callback.py
nicholasjng: Also removed.

jax/experimental/host_callback.py
nicholasjng: Removed.

jax/experimental/host_callback.py
nicholasjng: Removed.

hawkinsp (approved):

hawkinsp:
Thanks for the PR!

PR #12771, opened 2022-10-12
DOC: switch order of custom_jvp and _wraps

jakevdp:
This was causing issues with the documentation build, because the signature was not being propagated: https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.special.expn.html#jax.scipy.special.expn

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -1304,9 +1304,9 @@ def _expi_pos(x):
   )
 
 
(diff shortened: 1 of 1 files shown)


mattjj (approved):
Thanks for fixing this subtle issue!

Should custom_jvp propagate __signature__ better?

jakevdp:
> Should custom_jvp propagate signature better?

Probably? I'll look into it to try and figure out why it wasn't working. For now though, I think this is a simple enough fix.

PR #12772, opened 2022-10-12
Update version.py and CHANGELOG for jax 0.3.23 release

skye:


CHANGELOG.md
```diff
@@ -8,12 +8,14 @@ Remember to align the itemized text with the first line of an item within a list
 PLEASE REMEMBER TO CHANGE THE '..main' WITH AN ACTUAL TAG in GITHUB LINK.
 -->
 
-## jax 0.3.23
+## jax 0.3.24
```
jakevdp, line 11: Missing github commit links?

skye: We stopped doing those! Peter pointed out that with our release process, the release tarball would have the wrong URL (because we update the changelog after doing the release). So we just got rid of them. Did you find them useful? We could possibly figure out a way to have our cake and eat it too.

jakevdp: Oh, OK, I missed the memo on that 😁.

PR #12780, opened 2022-10-12
Make the etils[epath] dependency of JAX optional.

hawkinsp:
If epath is installed then we will use it for file I/O. If it is not, we will fall back to pathlib from the standard library.

### DIFF
--- a/jax/_src/path.py
+++ b/jax/_src/path.py
@@ -1,0 +1,33 @@
+# Copyright 2022 The JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import pathlib
+
+logger = logging.getLogger(__name__)
+
+try:
+  import etils.epath as epath
+except:
+  epath = None
+
+# If etils.epath (aka etils[epath] to pip) is present, we prefer it because it
+# can read and write to, e.g., GCS buckets. Otherwise we use the builtin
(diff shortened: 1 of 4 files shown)


jax/_src/path.py
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import pathlib
+
+logger = logging.getLogger(__name__)
```
yashk2810, line 18: What's this? Just use absl logging?

jax/_src/path.py
```diff
@@ -0,0 +1,33 @@
...
+
+# If etils.epath (aka etils[epath] to pip) is present, we prefer it because it
+# can read and write to, e.g., GCS buckets. Otherwise we use the builtin
+# pathlib and can only read/write to the local filesystem.
+if epath:
+  logger.debug("etils.epath found. Using etils.epath for file I/O.")
+  Path = epath.Path
+else:
+  logger.debug("etils.epath was not found. Using pathlib for file I/O.")
+  Path = pathlib.Path
```
yashk2810, line 33: If you want to use other functions from pathlib or epath, you won't be able to do that right?

jax/_src/path.py
yashk2810: Ohh wow.. I don't like the current thing. Why do you have to get the logger instead of doing just logging. We can discuss that separately though.

jax/_src/path.py
hawkinsp: Indeed. But do you?

(This appears to be all we actually use, and if we need more, we can add more later.)

yashk2810: Fair enough. But maybe its as simple as returning the module itself?

```
if epath:
  pathlib = epath
```

Then we can just do: `from jax._src.path import pathlib` and everything works as expected.

LGTM anyways.. up to you!

yashk2810 (approved):

jax/_src/path.py
hawkinsp: I'd rather keep the API minimal. But... it doesn't matter much. This is an internal API. If we need to change it later that's easy to do!

PR #12781, opened 2022-10-12
[typing] annotate jax._src.dtypes

jakevdp:
Part of #12049

Builds on #12837

### DIFF
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -21,13 +21,15 @@
 
 
 import functools
-from typing import Any, Dict, List
+from typing import overload, Any, Dict, List, Union
+from typing_extensions import Literal
 
 import numpy as np
 
 import jax
 from jax._src.config import flags, config
 from jax._src.lib import xla_client
+from jax._src.typing import DType, OpaqueDType
 
 from jax._src import traceback_util
 traceback_util.register_exclusion(__file__)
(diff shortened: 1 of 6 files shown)


jax/_src/dtypes.py
```diff
@@ -69,52 +71,55 @@
...
 
 
 @functools.lru_cache(maxsize=None)
-def _canonicalize_dtype(x64_enabled, dtype):
+def _canonicalize_dtype(x64_enabled: bool, dtype: DTypeLike) -> DType:
   """Convert from a dtype to a canonical dtype based on config.x64_enabled."""
   if jax.core.is_opaque_dtype(dtype):
-    return dtype
+    # TODO(frostig): opaque dtypes break the return type guarantee here... what to do?
+    return dtype  # type: ignore[return-value]
```
jakevdp, line 98: @froystig requesting your review in case you have thoughts about how to handle this.

froystig: Opaque dtypes are unlike dtypes—both structurally and nominally. Should we annotate the function as returning `Union[DType, jax._src.prng.KeyTy, jax.core.bint]`?

cc @mattjj

froystig: If it would be helpful to track the various opaque dtype definitions across the codebase, we could consider:
1. making them all conform to a protocol,
2. having them all derive from a common base type, or
3. manually maintaining an alias to a `Union` of them in `jax._src.typing`, e.g. named `OpaqueDType`

jakevdp: The problem with returning a union here is that it would pollute basically every dtype in JAX – we use `canonicalize_dtype` in many places where the output is required to be `DType`.

How frequently are opaque dtypes used? Could we have a special version of `canonicalize_dtype` to use in the places where opaque dtypes are expected?

froystig: Good idea to see whether we can distinguish at the call site. We'd have to investigate... The changes were introduced in #11768, which added only one explicit seemingly-relevant call to `canonicalize_dtype` (in changes to `api_util._shaped_abtractify_slow`). It's still possible that we rely on this change to `canonicalize_dtype` in other places, but this is where I'd look first to see whether modifying a single caller suffices.

jakevdp: Thanks, I'll try adding an `allow_opaque_dtypes` keyword to `canonicalize_dtype` that defaults to False. Then tests will tell us when we need to set it to True.

froystig (approved):

PR #12808, opened 2022-10-14
Add Python 3.11 as a compatible Python version.

hawkinsp:


yashk2810:
Let's wait until we actually release for it? Or is this required before we release?

hawkinsp:
This is probably good to have for the release. It's also correct: if you build the wheel yourself, it works with 3.11.

yashk2810 (approved):

PR #12837, opened 2022-10-17
add allow_opaque_dtype to dtypes.canonicalize_dtype utility

jakevdp:
Related to https://github.com/google/jax/pull/12781#discussion_r996577250

### DIFF
--- a/jax/_src/api_util.py
+++ b/jax/_src/api_util.py
@@ -438,7 +438,7 @@ def _shaped_abstractify_slow(x):
   weak_type = getattr(x, 'weak_type', False)
   named_shape = getattr(x, 'named_shape', {})
   if hasattr(x, 'dtype'):
-    dtype = dtypes.canonicalize_dtype(x.dtype)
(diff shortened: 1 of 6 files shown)


froystig (commented):
Should we make a type annotation for opaque dtypes and use it in the output annotations of `canonicalize_dtype` (in place of the `Any`s currently there)?

jakevdp:
Yeah, that's probably a good idea... do opaque dtypes have a well-defined API? Maybe I could add that as a TODO to this PR?

froystig:
Sounds good. They do have an API. We have a TODO roughly to this effect [here](https://github.com/google/jax/blob/fd2f590b3ba404f60b16fd4d58339194de1421c1/jax/core.py#L1205). The new bit to add is that we should update a few annotations accordingly. Want to add to that and hopefully @mattjj or I will take care of it as part of that TODO?

froystig (approved):

PR #12844, opened 2022-10-17
Skip unittest for stream_executor runtime 

yejingxin:
Run unit tests on CloudTPU V2-8 and found a bunch of tests fails due to call back is not supported for stream_executor runtime , disable all those tests if runtime is stream_executor .

### DIFF
--- a/tests/debugger_test.py
+++ b/tests/debugger_test.py
@@ -26,6 +26,7 @@ from jax.experimental import pjit
 from jax._src import debugger
 from jax._src import lib as jaxlib
 from jax._src import test_util as jtu
+from jax._src.lib import xla_bridge
 import jax.numpy as jnp
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


skye:
Do any of the tests pass? It might be simpler to just put the check in the test class's `setUp`?

Also can you change the commit/PR title to be a bit more descriptive? Something like: "Skip debugger unit tests if runtime is `stream_executor`"

yejingxin:
It is hard to have the logic in `setUp`, since not all the tests within a test class needs to skip. One possible way to make it more lint is add decorator like `@jtu.skip_on_devices(*disabled_backends)`, we can create one `@jtu.skip_on_running_time("stream_executor")`

skye:
Ok, this is fine then. Please do change the title though, and also squash all your commits into a single commit (otherwise we end up with merge commits and messy git history).

yejingxin:
Thanks Skye, I merged all those commits as one, and updated the commit title accordingly.

.github/workflows/ci-build.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Cancel previous
-        uses: styfle/cancel-workflow-action@0.10.1
+        uses: styfle/cancel-workflow-action@0.11.0
```
skye, line 25: I'm not sure why these changed? Revert?

yejingxin: it is automatically generate github bots. I manually removed it.

skye (approved):

Issue #12829, opened 2022-10-17
WARNING:root:Error importing mlir initializer _mlirRegisterEverything. This may happen in unclean incremental builds but is likely a real bug if encountered otherwise and the MLIR Python API may not function.


JPGoodale:
### Description

ANY piece of jax code I run returns this same error:

WARNING:root:Error importing mlir initializer _mlirRegisterEverything. This may happen in unclean incremental builds but is likely a real bug if encountered otherwise and the MLIR Python API may not function.
Traceback (most recent call last):
  File "/home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/site-packages/jaxlib/mlir/_mlir_libs/__init__.py", line 62, in process_initializer_module
    m = importlib.import_module(f".{module_name}", __name__)
  File "/home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1176, in create_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
ImportError: /home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/site-packages/jaxlib/mlir/_mlir_libs/_mlirRegisterEverything.so: undefined symbol: mlirRegisterConversionPasses
Traceback (most recent call last):
  File "/home/jpgoodale/anaconda3/envs/deeplearning/testingJAX/main.py", line 1, in <module>
    import jax
  File "/home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/site-packages/jax/__init__.py", line 35, in <module>
    from jax import config as _config_module
  File "/home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/site-packages/jax/config.py", line 17, in <module>
    from jax._src.config import config
  File "/home/jpgoodale/anaconda3/envs/deeplearning/lib/python3.10/site-packages/jax/_src/config.py", line 29, in <module>
    from jax._src import lib
[...]

PR #12845, opened 2022-10-18
bump the jaxlib version in `version.py`

froystig:
cc @yashk2810, who suggested 0.3.22.

### DIFF
--- a/jax/version.py
+++ b/jax/version.py
@@ -16,7 +16,7 @@
 # eval()-ed by setup.py, so it should not have any dependencies.
 
 __version__ = "0.3.24"
-_minimum_jaxlib_version = "0.3.15"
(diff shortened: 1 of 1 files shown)


hawkinsp:
Well, we certainly *can* bump the minimum. But I don't think it solves any particular problem, it just lets us remove any backward compatibility code.

If you want to bump the minimum, you should also grep for instances of `xla_extension_version` or `mlir_api_version` that can be removed (or even any TODOs mentioning jaxlib versions) and fix them.

froystig:
Yeah, seeing your comment on https://github.com/google/jax/issues/12829#issuecomment-1282378705, at this point this would just be an unrelated opportunity to clean up a bit. I'll try to come back to this PR and treat it that way (grepping as you suggest).

PR #12851, opened 2022-10-18
[typing] annotate util.unzip2 & util.unzip3

jakevdp:
Part of #12049


jax/interpreters/partial_eval.py
```diff
@@ -1165,7 +1165,7 @@ def ensure_instantiated(inst: bool, x: Atom) -> Atom:
         inputs = map(ensure_instantiated, inst_in, eqn.invars)
         staged_eqns.append(eqn.replace(invars=inputs))
         map(partial(write, False, True), eqn.outvars)
-  out_unknowns, out_inst = unzip2(map(read, jaxpr.outvars))
+  out_unknowns, out_inst = unsafe_map(list, unzip2(map(read, jaxpr.outvars)))
```
sharadmv, line 1168: Why doesn't `unzip2` work here? `read` could be annotated to return `Tuple[bool, bool]` right?

sharadmv (approved):

PR #12872, opened 2022-10-19
[typing] annotate jax._src.util.safe_map

jakevdp:
Part of #12049

### DIFF
--- a/jax/_src/sharding.py
+++ b/jax/_src/sharding.py
@@ -327,7 +327,7 @@ class PmapSharding(XLACompatibleSharding):
   @functools.lru_cache(maxsize=4096)
   def devices_indices_map(self, global_shape: Shape) -> Mapping[Device, Index]:
     indices = pxla.spec_to_indices(global_shape, self.sharding_spec)
-    return {d: i for d, i in safe_zip(self.devices.flat, indices)}  # type: ignore
+    return dict(safe_zip(self.devices.flat, indices))  # type: ignore[arg-type]
 
   @pxla.maybe_cached_property
   def _device_assignment(self) -> XLADeviceAssignment:
--- a/jax/experimental/sparse/transform.py
+++ b/jax/experimental/sparse/transform.py
@@ -48,7 +48,7 @@ DeviceArray([-1.2655463 , -0.52060574, -0.14522289, -0.10817424,
 
 import functools
 from typing import (
-  Any, Callable, Dict, NamedTuple, List, Optional, Sequence, Tuple, Union)
+  Any, Callable, Dict, NamedTuple, List, Optional, Sequence, Tuple)
 
 import numpy as np
 
@@ -353,11 +353,12 @@ def eval_sparse(
 ) -> Sequence[SparsifyValue]:
   env : Dict[core.Var, SparsifyValue] = {}
 
-  def read(var: core.Var) -> Union[Array, SparsifyValue]:
+  def read(var: core.Atom) -> SparsifyValue:
     # all literals are dense
     if isinstance(var, core.Literal):
       return spenv.dense(var.val)
     else:
+      assert isinstance(var, core.Var)
       return env[var]
 
   def write_buffer(var: core.Var, a: Array) -> None:
--- a/jax/interpreters/mlir.py
+++ b/jax/interpreters/mlir.py
@@ -977,13 +977,14 @@ def jaxpr_subcomp(ctx: ModuleContext, jaxpr: core.Jaxpr,
   Assumes that an MLIR context, location, and insertion point are set.
   """
   assert ctx.platform != "gpu"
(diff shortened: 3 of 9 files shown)


jax/_src/sharding.py
```diff
@@ -49,8 +49,7 @@ def device_set(self) -> Set[Device]:
     raise NotImplementedError('Subclasses should implement this method.')
 
   @abc.abstractmethod
-  def devices_indices_map(
-      self, global_shape: Shape) -> Mapping[Device, Optional[Index]]:
+  def devices_indices_map(self, global_shape: Shape) -> Mapping[Device, Index]:
```
mattjj, line 52: I think this should stay as a `Mapping[Device, Optional[Index]]`; it's part of the API. Was there an issue with that?

jax/experimental/sparse/transform.py
```diff
@@ -353,11 +353,12 @@ def eval_sparse(
 ) -> Sequence[SparsifyValue]:
   env : Dict[core.Var, SparsifyValue] = {}
 
-  def read(var: core.Var) -> Union[Array, SparsifyValue]:
+  def read(var: Union[core.Var, core.Literal]) -> SparsifyValue:
```
mattjj, line 356: I think `core.Atom` is an alias for `Union[core.Var, core.Literal]`. Could use it instead.

jax/interpreters/mlir.py
```diff
@@ -977,13 +977,13 @@ def jaxpr_subcomp(ctx: ModuleContext, jaxpr: core.Jaxpr,
   Assumes that an MLIR context, location, and insertion point are set.
   """
   assert ctx.platform != "gpu"
-  def read(v: core.Var) -> Sequence[ir.Value]:
+  def read(v: Union[core.Var, core.Literal]) -> Sequence[ir.Value]:
     if type(v) is core.Literal:
       return ir_constants(v.val, canonicalize_types=True)
     else:
-      return env[v]
+      return env[cast(core.Var, v)]
```
mattjj, line 984: Can we write an `assert isinstance(v, core.Var)` instead of `cast`? I hate `cast` :P

jax/interpreters/mlir.py
```diff
@@ -977,13 +977,13 @@ def jaxpr_subcomp(ctx: ModuleContext, jaxpr: core.Jaxpr,
   Assumes that an MLIR context, location, and insertion point are set.
   """
   assert ctx.platform != "gpu"
-  def read(v: core.Var) -> Sequence[ir.Value]:
+  def read(v: Union[core.Var, core.Literal]) -> Sequence[ir.Value]:
```
mattjj, line 980: Could use `core.Atom` here.

jax/interpreters/partial_eval.py
```diff
@@ -1166,7 +1165,9 @@ def ensure_instantiated(inst: bool, x: Atom) -> Atom:
         inputs = map(ensure_instantiated, inst_in, eqn.invars)
         staged_eqns.append(eqn.replace(invars=inputs))
         map(partial(write, False, True), eqn.outvars)
-  out_unknowns, out_inst = unsafe_map(list, unzip2(map(read, jaxpr.outvars)))
+  out_unknowns: Sequence[bool]
+  out_inst: Sequence[bool]
+  out_unknowns, out_inst = unzip2(map(read, jaxpr.outvars))
```
mattjj, line 1170: I'm not a big fan of declaring variables before they're bound. As with `cast`, it seems like just making Python more verbose.

If there's no way to annotate this on a single line, can we `# type: ignore` instead?

mattjj (approved):

jax/_src/sharding.py
jakevdp: Every concrete implementation of this method is annotated `Mapping[Device, Index]` (https://cs.github.com/google/jax?q=%22def+devices_indices_map%22), and there are several places in the code where the `None` possibility is unhandled, and these are flagged by mypy under the old implementation.

jax/interpreters/partial_eval.py
jakevdp: I changed it to not use declared variables, but I couldn't figure out how to do it in one line.

jax/interpreters/mlir.py
jakevdp: Done

jax/interpreters/mlir.py
jakevdp: Done

jax/experimental/sparse/transform.py
jakevdp: Done

jax/_src/sharding.py
jakevdp: Switched back after chatting with @yashk2810

Issue #12537, opened 2022-09-27
JEP: Unified Error Handling API


sharadmv:
Tracking issue for JAX unified error handling API

patrick-kidger:
CC'd myself as I'm very interested in this.

PR #12877, opened 2022-10-20
checkify: error types

LenaMartens:
# Checkify: upcoming changes/improvements
ref: #12537  
Authors: @LenaMartens, @sharadmv, @mattjj
### tl;dr
checkify is being updated to support more granular and rich error types. This change will (hopefully) only be an improvement to end-users and will enable future improvements to `checkify`.

### Background
Currently, `checkify` has two pieces of core functionality:

* `checkify` provides a check function that enables assertions in JIT-ted JAX code. 
checkify provides a function called checkify, which is multi-purpose. It first instruments JAX code with error checks (e.g. invalid nan checks, divide-by-zero checks, OOB indexing checks, etc.). 

* It also (simultaneously) discharges these checks into error values, returning the error value to the user. The user then can call `error.throw()` to raise a Python error corresponding to the failed check.

Internally, checkify keeps track of a bool (`err`) and an int (`code`), respectively corresponding to whether or not an error has happened yet and a unique integer pointing to the metadata about the error (a string with the source file and line number). It currently keeps track of a single `err` and `code`, corresponding to the first error which occurred (of any type).

## Proposed changes
We propose upgrading `checkify’s` internal bookkeeping, keeping track of multiple error types simultaneously. 

### New error types
`checkify` has the ErrorCategory enum, which allows users to select which checks they want enabled when calling `checkify.checkify`. All error categories (DIV, NAN, etc.) are tracked together (they all share the same err, code and payload). To enable more granular tracking, we use a new error class that tracks its own payload and static information.

We propose adding JaxError, a subclass of Python exception and Pytree that tracks both static information (such as source provenance and primitive names) and dynamic information (like indices or user-provided arrays). 

```python
class ConcreteError(Exception, metaclass=abc.ABCMeta):
  def __init__(self, source_info):
    self.source_info = source_info
    self.with_traceback(self.source_info)

  @abc.abstractmethod
  def tree_flatten(self):
    return [], self.source_info

  @abc.abstractclassmethod
  def tree_unflatten(cls, source_info, _):
    return cls(source_info)

  @abc.abstractmethod
  def get_effect_type(self) -> core.Effect:
    pass
```

For example, the DivideByZero error only encapsulates static source information, the IndexOutOfBounds error encapsulates static source information, a static primitive name (e.g. ”gather” or ”scatter”), and three dynamic ints (index, axis_size, axis) although only index needs to be dynamic. AssertionErrors track static source info, a static user format string, and a dynamic, variadic payload.

Example: OOB

```python
@register_pytree_node_class
class OOBError(ConcreteError):

  def __init__(self, source_info, primitive_name, operand_shape, payload):
    super().__init__(source_info)
    self.prim = primitive_name
    self.operand_shape = operand_shape
    self.payload = payload

  def tree_flatten(self):
    return self.payload, (self.source_info, self.prim, self.operand_shape)

  @classmethod
  def tree_unflatten(cls, metadata, payload):
    return cls(*metadata, payload)

  def __str__(self):
    return (f'Out-of-bounds indexing in {self.prim} for array of '
            f'shape {self.operand_shape}: '
            f'index {self.payload[0]} is out of bounds for axis {self.payload[1]} '
            f'with size {self.payload[2]}.')
  
  def get_effect_type(self):
    return (OOB, (jax.ShapeDtypeStruct((3,), np.int32)))
```

### Richer effect types
Currently, checks are tracked in jaxprs via effects but all checks share the same type. We are changing this so that each check will have its own effect type, created by the JaxError object. The effect types need to contain the static information about the payload (i.e. shapes and dtypes of each of the arguments). This is so that HOPs can be aware of what payloads are present in a jaxpr for transformations.

As an example, a jaxpr with a user assertion before would have an effects field that looked like `{Check}` (where Check is an enum element). This will now be changed to something like `{Error<AssertionError[]>}` where the [] contain shape and dtype information of any dynamic payloads in the assertion error (see “Assertions with more information” below).

### Full tracebacks
`checkify` currently only tracks the source file and line number in a check. We propose upgrading this to a full traceback. This just involves tracking more information and eventually raising an error that points to the traceback using the __cause__ mechanism.
One complication is BatchedError which could encompass many errors with their own trace-backs. It’s unclear which trace-back to attach to the error. Something like Python’s exception groups would make sense to use here.

Example:
The following code:
```python
import jax
import jax.numpy as jnp
from jax.experimental import checkify

@jax.jit
def f(x):
  return x / x

err, out = checkify.checkify(f, errors=checkify.div_checks)(0.)
err.throw()
```
will now throw this exception:
```python
  File "/jax/checkify_test.py", line 9, in <module>
    err, out = checkify.checkify(f, errors=checkify.div_checks)(0.)
  File "/jax/checkify_test.py", line 7, in f
    return x / x
  File "/jax/jax/_src/numpy/lax_numpy.py", line 4936, in deferring_binary_op
    return binary_op(*args)
  File "/jax/jax/_src/numpy/ufuncs.py", line 258, in true_divide
    return lax.div(x1, x2)
jax._src.checkify.DivisionByZeroError: division by zero

  File "/jax/checkify_test.py", line 10, in <module>
    err.throw()
  File "/jax/jax/_src/checkify.py", line 278, in throw
    check_error(self)
  File "/jax/jax/_src/checkify.py", line 697, in check_error
    return check_p.bind(*err_args, err_tree=tree_def)
  File "/jax/jax/core.py", line 329, in bind
    return self.bind_with_trace(find_top_trace(args), args, params)
  File "/jax/jax/core.py", line 332, in bind_with_trace
    out = trace.process_primitive(self, map(trace.full_raise, args), params)
  File "/jax/jax/core.py", line 712, in process_primitive
    return primitive.impl(*tracers, **params)
  File "/jax/jax/_src/checkify.py", line 710, in check_impl
    raise JaxRuntimeError(str(exc)) from exc
jax._src.checkify.JaxRuntimeError: division by zero

```

### Assertions with more information
With the new proposed error types, the AssertionError type can now have a variadic payload, allowing users to do things like:
`checkify.check(x != 0, "x is 0: {}", x)`

Users can now choose to embed run-time information of any shape/dtype in their error messages.

### Changes to end-users
These should be almost entirely internal changes. The `checkify` API remains the same.The only user-visible change is better tracebacks and a more flexible assertion API with variadic runtime information support.

## Future improvements
This change enables a few major improvements to `checkify`.

1. Separate instrumentation and discharging - right now, `checkify` introduces certain checks and immediately discharges them. In the future, we can explore adding in checks and represent them explicitly in jaxprs. The specific checks present in a jaxpr are immediately visible from its effect type.
2. Partial discharging - when we have many different error effects in a jaxpr, `checkify` will be able to discharge some into an error value and keep the remaining ones present. This will make `checkify` more expressive and will allow it to do things akin to try/except in Python where we only catch certain exceptions.


sharadmv (approved):
LGTM!!!!!

LenaMartens:
I ran into tracer leaks when running internal tests (under `jax_check_tracer_leaks`) because of the `traceback`s! Seems like they're holding on to values (inside of the frames?). I've reverted the `traceback` change for now so we can leave it for a future improvement (I'm sure we can filter out the values somehow?), and testing with `jax_check_tracer_leaks=True` on all CheckifyTests.

PR #12904, opened 2022-10-20
canonicalize dtypes when loading arrays via `dlpack`

### LINKED ISSUE #11895: BUG: can create 64-bit arrays using dlpack when x64 is disabled
### Description

Repro:
```python
import tensorflow as tf
import jax
import jax.dlpack

assert not jax.config.jax_enable_x64
x = jax.dlpack.from_dlpack(tf.experimental.dlpack.to_dlpack(tf.ones(5, tf.int64)))
assert x.dtype == jax.numpy.int64
```

This can cause loud and confusing errors down the line when our MHLO expects 32-bit numbers but gets 64-bit. For example, with our illegal 64-bit `x` array, try doing `x[0]`.

HEAD/0.3.15

CPU

### Additional System Info

_No response_

froystig:
... converting if needed.

Fixes #11895

### DIFF
--- a/jax/_src/dlpack.py
+++ b/jax/_src/dlpack.py
@@ -70,4 +70,5 @@ def from_dlpack(dlpack):
   xla_shape = buf.xla_shape()
   assert not xla_shape.is_tuple()
   aval = core.ShapedArray(xla_shape.dimensions(), xla_shape.numpy_dtype())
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks great!

Issue #12742, opened 2022-10-11
Add scipy.special.beta


carlosgmartin:
Add [scipy.special.beta](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.beta.html). Among other things, this is needed to compute the incomplete beta function from the _regularized_ incomplete beta function, [scipy.special.betainc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.betainc.html).

 (See #6765 for adding [scipy.special.gamma](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.gamma.html).)

PR #12918, opened 2022-10-21
Expose beta function 

tkillestein:
Adds a beta function implementation (as flagged in #12742)


jakevdp (changes requested):
Looks great! A couple additional things:

- we should add a test: in this case it should be as easy as adding a new `op_record` here: https://github.com/google/jax/blob/503679d63610d985e562471d710b2b227eea5e23/tests/lax_scipy_test.py#L113
- we need to add this to the docs by adding it to the list here: https://github.com/google/jax/blob/main/docs/jax.scipy.rst. It looks like `betaln` is missing there currently, so you could add it as well.
- and once you've done this, please be sure to squash all these changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Thanks!

jakevdp (approved):

PR #12920, opened 2022-10-21
Fix an NameError caused by #12754

yilei:


jax/_src/third_party/scipy/signal_helper.py
```diff
@@ -53,8 +53,10 @@ def _triage_segments(window: Union[ArrayLike, str, Tuple[Any, ...]], nperseg: Op
       raise ValueError('window is longer than input signal')
     if nperseg is None:
       nperseg_int = win.size
-    elif nperseg != win.size:
-      raise ValueError("value specified for nperseg is different from length of window")
+    else:
+      if nperseg != win.size:
+        raise ValueError("value specified for nperseg is different from length of window")
+      nperseg_int = win.size
```
jakevdp, line 59: I think this would be the equivalent of the original logic:
```
elif nperseg != win.size:
  raise ValueError("value specified for nperseg is different from length of window")
else:
  nperseg_int = int(nperseg)
```
i.e. the default case is that `nperseg_int` is just `nperseg` converted to an integer.

jakevdp (changes requested):
Thanks for catching this!

Could we add a regression test to cover this case?

jax/_src/third_party/scipy/signal_helper.py
yilei: Done

yilei:
Could you please give pointers how a test could be added? I don't find an existing unittest for this module, and I'm not familiar with the code at all. Thanks!

jakevdp:
This code should be covered by tests in https://github.com/google/jax/blob/main/tests/scipy_signal_test.py

Take a look - if it's not something you want to tackle, let me know and I can take care of it after merging this fix.

Thanks!

jakevdp (approved):

yilei:
This test file looks like non trivial to understand given I'm not familiar with the code. Could you take care of it? Thank you!

jakevdp:
Sounds good, thanks!

PR #12943, opened 2022-10-24
Improve Gamma sampling

adamgayoso:
This fixes:

- a todo to use lax cond in order to speed up a computation during gamma sampling

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -29,6 +29,7 @@ from jax.config import config
 from jax.core import NamedShape
 from jax._src.api import jit, vmap
 from jax._src.lax import lax as lax_internal
+from jax._src.lib import xla_bridge
 from jax._src.numpy.lax_numpy import _arraylike, _check_arraylike, _convert_and_clip_integer, _promote_dtypes_inexact
 from jax.numpy.linalg import cholesky, svd, eigh
 from jax.interpreters import ad
@@ -1005,7 +1006,10 @@ def _gamma_grad(sample, a, *, log_space):
     gamma_grad = lambda alpha, sample: lax.random_gamma_grad(alpha, sample) / sample
   else:
     gamma_grad = lax.random_gamma_grad
-  grads = vmap(gamma_grad)(alphas, samples)
+  if xla_bridge.get_backend().platform == 'cpu':
+    grads = lax.map(lambda args: gamma_grad(*args), (alphas, samples))
+  else:
+    grads = vmap(gamma_grad)(alphas, samples)
   return grads.reshape(np.shape(a))
 
 def _gamma_impl(key, a, *, log_space, use_vmap=False):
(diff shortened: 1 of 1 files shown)


jax/_src/random.py
```diff
@@ -957,7 +958,7 @@ def _cond_fn(kXVU):
     true_fun = lambda: lax.ge(lax.log(U), lax.add(lax.mul(X, one_over_two),
                                                       lax.mul(d, lax.add(lax.sub(one, V),
                                                                          lax.log(V)))))
-    false_fun = lambda: False
+    false_fun = lambda: lax.le(one, zero)
```
jakevdp, line 961: What's the reason for turning a constant into an expression? Mabye `lambda: np.bool_(False)` would be sufficient?

adamgayoso: No good reason. Fixed :)

jax/_src/random.py
```diff
@@ -927,6 +927,7 @@ def _gamma_one(key: KeyArray, alpha, log_space) -> Array:
   # https://en.wikipedia.org/wiki/Gamma_distribution#Generating_gamma-distributed_random_variables
   zero = _lax_const(alpha, 0)
   one = _lax_const(alpha, 1)
+  false = _lax_const(alpha, False)
```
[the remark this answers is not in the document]
adamgayoso: Done.

jakevdp (approved):
Looks good to me, though I'm a bit hesitant to merge this without understanding why `lax.map` was used in the first place, and confirming that that reason no longer is important. Have you been able to discern that?

adamgayoso:
cc @fehiepsi  -- looks like it was added when the grad was still implemented in this file.

I can try a quick benchmark on colab for now?

adamgayoso:
@jakevdp actually it does indeed faster to map on cpu, I had a bug in my benchmark. Also would probably good to leave these potential changes in separate PRs. Therefore I changed this PR to only include the lax cond change

jax/_src/random.py
```diff
@@ -953,13 +953,12 @@ def _gamma_one(key: KeyArray, alpha, log_space) -> Array:
...
-    # if the first condition is satisfied
-    cond = lax.bitwise_and(lax.ge(U, lax.sub(one, lax.mul(squeeze_const, lax.mul(X, X)))),
-                           lax.ge(lax.log(U), lax.add(lax.mul(X, one_over_two),
+    pred = lax.ge(U, lax.sub(one, lax.mul(squeeze_const, lax.mul(X, X))))
+    true_fun = lambda: lax.ge(lax.log(U), lax.add(lax.mul(X, one_over_two),
                                                       lax.mul(d, lax.add(lax.sub(one, V),
-                                                                         lax.log(V))))))
+                                                                         lax.log(V)))))
+    false_fun = lambda: np.bool_(False)
+    cond = lax.cond(pred, true_fun, false_fun)
```
fehiepsi, line 961: Does this change make things faster? Could you post some benchmark numbers?

jakevdp:
It seems like this has become stale. I'm going to close for now – feel free to re-open if you'd like to continue working on this!

PR #12960, opened 2022-10-24
[typing] annotate next part of lax_numpy.py

jakevdp:
Part of #12049; follows #12951

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -426,8 +426,9 @@ def histogram_bin_edges(a: ArrayLike, bins: ArrayLike = 10,
   dtype = dtypes.to_inexact_dtype(arr.dtype)
   if _ndim(bins) == 1:
     return asarray(bins, dtype=dtype)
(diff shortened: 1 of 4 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -426,7 +426,8 @@ def histogram_bin_edges(a: ArrayLike, bins: ArrayLike = 10,
   dtype = dtypes.to_inexact_dtype(arr.dtype)
   if _ndim(bins) == 1:
     return asarray(bins, dtype=dtype)
-  bins = core.concrete_or_error(operator.index, bins,
+
+  bins_int = core.concrete_or_error(operator.index, bins,
                                 "bins argument of histogram_bin_edges")
```
zhangqiaorjc, line 431: should we fix this indent?

zhangqiaorjc (approved):

jax/_src/numpy/lax_numpy.py
jakevdp: Done

PR #12962, opened 2022-10-25
Add instructions for building with ROCM.

hawkinsp:
@rsanthanam-amd

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -130,6 +130,31 @@ python .\build\build.py `
 
 To build with debug information, add the flag `--bazel_options='--copt=/Z7'`.
 
+### Additional notes for building a ROCM `jaxlib` for AMD GPUs
+
+You need several ROCM/HIP libraries installed to build for ROCM. For
+example, on a Ubuntu machine with AMD's `apt` repositories available, you need
+a number of packages installed:
+
+```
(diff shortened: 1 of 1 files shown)


rsanthanam-amd:
LGTM.  Adding @jayfurmanek for further commentary.

hawkinsp:
One question I have is: when is it necessary to use the AMD fork of TF? I'm toying with the idea of setting up a ROCM CI build (build only, we don't have the GPU hardware) and it would be very helpful if we could just use upstream TF.

rsanthanam-amd:
Currently, JAX will not build with upstream TF so the AMD fork must be used.

I need to discuss a plan with my team to enable ROCm JAX to build with upstream TF.

jayfurmanek:
> Currently, JAX will not build with upstream TF so the AMD fork must be used.

The fix is likely fairly simple. Getting it merged in TF is another story. We'd love to work mostly upstream but we have some difficulties getting getting approvals and merges unfortunately for various reason. For example the PR for ROCM5.2 support has been bouncing around since early July:
https://github.com/tensorflow/tensorflow/pull/56762]

That one will be needed if we want the CI to build on ROCM 5.2 or later.

rahulbatra85:
I created an issue to track JAX ROCm CI development

PR #12973, opened 2022-10-25
tweaks to DevicesSharding / ReshapeableDevicesSharding

mattjj:
1. rename DevicesSharding -> ReshapeableDevicesSharding
2. fix repr to print device order faithfully
3. respect shape of np.ndarray argument to __init__

### DIFF
--- a/tests/array_test.py
+++ b/tests/array_test.py
@@ -730,10 +730,25 @@ class ShardingTest(jtu.JaxTestCase):
     mesh = jtu.create_global_mesh((4, 2), ('x', 'y'))
     mps = sharding.MeshPspecSharding(mesh, pspec)
 
-    devices_sharding = sharding.DevicesSharding(jax.devices())
+    devices_sharding = sharding.ReshapeableDevicesSharding(jax.devices())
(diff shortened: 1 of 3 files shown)


tests/array_test.py
```diff
@@ -730,7 +730,7 @@ def test_devices_sharding_op_sharding_lowering(
     mesh = jtu.create_global_mesh((4, 2), ('x', 'y'))
     mps = sharding.MeshPspecSharding(mesh, pspec)
 
-    devices_sharding = sharding.DevicesSharding(jax.devices())
+    devices_sharding = sharding.ReshapeableDevicesSharding(jax.devices())
     devices_sharding = devices_sharding.reshape(shape).replicate(axes)
```
yashk2810, line 734: Maybe add a test where you don't do this reshape? To test that you are actually respecting the user shape?

In this test, you can get it by doing `mesh.devices` to get a shape that's not 1D

yashk2810 (approved):

tests/array_test.py
mattjj: Done!

PR #12977, opened 2022-10-25
Skip two unit tests about custom sharding on libtpu

yejingxin:
DETAILS:
Due to xc.register_custom_call_partitioner is not supported on libtpu, the following two tests are skipped: tests/pjit_test.py::PJitTest::test_custom_partitioner tests/debugging_primitives_test.py::InspectShardingTest::test_inspect_sharding_is_called_in_pjit

### DIFF
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -255,6 +255,9 @@ def is_device_rocm():
 def is_device_cuda():
   return xla_bridge.get_backend().platform_version.startswith('cuda')
 
+def is_cloud_tpu():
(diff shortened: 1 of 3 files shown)


skye:
Do you mind adding a `is_cloud_tpu` helper function to https://github.com/google/jax/blob/main/jax/_src/test_util.py and using that instead of copy/pasting the check? That way it's easier to reuse, update, and find tests that we need to reenable.

yejingxin:
Add a util func is_cloud_tpu to test_utils. PTAL, thanks

skye (approved):
Thanks!

PR #13008, opened 2022-10-27
Add support for using pip-installed CUDA wheels.

hawkinsp:
Change `jax[cuda]` to depend on the pip CUDA wheels. Recommend this path.
Add a `jax[cuda_using_existing_cuda_installation]` to use the system CUDA instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,18 +439,17 @@ source](https://jax.readthedocs.io/en/latest/developer.html#building-from-source
   * cuDNN 8.6 or newer. We recommend using the cuDNN 8.6 wheel if your cuDNN
     installation is new enough, since it supports additional functionality.
   * cuDNN 8.2 or newer.
-* You *must* use an NVidia driver version that is at least as new as your
+* You *must* use an NVIDIA driver version that is at least as new as your
   [CUDA toolkit's corresponding driver version](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-major-component-versions__table-cuda-toolkit-driver-versions).
-  For example, if you have CUDA 11.4 update 4 installed, you must use NVidia
+  For example, if you have CUDA 11.4 update 4 installed, you must use NVIDIA
   driver 470.82.01 or newer if on Linux. This is a strict requirement that
   exists because JAX relies on JIT-compiling code; older drivers may lead to
   failures.
   * If you need to use an newer CUDA toolkit with an older driver, for example
-    on a cluster where you cannot update the NVidia driver easily, you may be
+    on a cluster where you cannot update the NVIDIA driver easily, you may be
     able to use the
     [CUDA forward compatibility packages](https://docs.nvidia.com/deploy/cuda-compatibility/)
-    that NVidia provides for this purpose.
-
+    that NVIDIA provides for this purpose.
 
 Next, run
 
@@ -527,7 +526,7 @@ simply run
 conda install jax -c conda-forge
 ```
 
-To install on a machine with an NVidia GPU, run
+To install on a machine with an NVIDIA GPU, run
 ```bash
 conda install jaxlib=*=*cuda* jax cuda-nvcc -c conda-forge -c nvidia
 ```
--- a/jax/_src/lib/__init__.py
+++ b/jax/_src/lib/__init__.py
@@ -16,8 +16,8 @@
 # checking on import.
 
 import gc
+import pathlib
 import re
-import os
 from typing import Optional, Tuple
 
 try:
@@ -115,9 +115,22 @@ import jaxlib.gpu_rnn as gpu_rnn  # pytype: disable=import-error
 mlir_api_version = xla_client.mlir_api_version
 
 # TODO(rocm): check if we need the same for rocm.
-cuda_path: Optional[str]
-cuda_path = os.path.join(os.path.dirname(jaxlib.__file__), "cuda")
-if not os.path.isdir(cuda_path):
-  cuda_path = None
+
+def _cuda_path() -> Optional[str]:
+  _jaxlib_path = pathlib.Path(jaxlib.__file__).parent
+  # If the pip package nvidia-cuda-nvcc-cu11 is installed, it should have
+  # both of the things XLA looks for in the cuda path, namely bin/ptxas and
+  # nvvm/libdevice/libdevice.10.bc
+  path = _jaxlib_path.parent / "nvidia" / "cuda_nvcc"
+  if path.is_dir():
+    return str(path)
+  # Failing that, we use the copy of libdevice.10.bc we include with jaxlib and
+  # hope that the user has ptxas in their PATH.
+  path = _jaxlib_path / "cuda"
+  if path.is_dir():
+    return str(path)
+  return None
+
+cuda_path = _cuda_path()
 
 transfer_guard_lib = xla_client._xla.transfer_guard_lib
--- a/jaxlib/cuda/BUILD
+++ b/jaxlib/cuda/BUILD
@@ -97,6 +97,13 @@ pybind_extension(
         "-fno-strict-aliasing",
     ],
     features = ["-use_header_modules"],
+    linkopts = select({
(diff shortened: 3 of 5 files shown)


jax/_src/lib/__init__.py
```diff
@@ -131,9 +130,22 @@ def _parse_version(v: str) -> Tuple[int, ...]:
...
-cuda_path = os.path.join(os.path.dirname(jaxlib.__file__), "cuda")
-if not os.path.isdir(cuda_path):
-  cuda_path = None
+
+def _cuda_path() -> Optional[str]:
+  _jaxlib_path = pathlib.Path(jaxlib.__file__).parent
+  # If the pip package nvidia-cuda-nvcc-cu11 is installed, it should have
+  # both of the things XLA looks for in the cuda path, namely bin/ptxas and
+  # nvvm/libdevice/libdevice.10.bc
+  path = _jaxlib_path.parent / "nvidia" / "cuda_nvcc"
```
yashk2810, line 139: Don't know if its intentional but you have a double parent here. First on line 135 and then on line 139.

setup.py
```diff
@@ -91,13 +91,28 @@ def generate_proto(source):
...
+
+        # Target that does not depend on the CUDA pip wheels, for those who want
+        # to use a preinstalled CUDA.
+        'cuda_using_existing_cuda_installation': [
+          f"jaxlib=={_current_jaxlib_version}+cuda{_default_cuda_version}.cudnn{_default_cudnn_version}",
+        ],
 
         # CUDA installations require adding jax releases URL; e.g.
         # $ pip install jax[cuda11_cudnn82] -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-        # $ pip install jax[cuda11_cudnn805] -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
yashk2810, line 100: Why remove this comment if its not covered by cuda installation?

jax/_src/lib/__init__.py
hawkinsp: It is indeed intentional. The first one goes from a file name to a directory name, and the second one goes to the parent directory.

setup.py
hawkinsp: I guess it's unrelated to this change, but we can/should drop that wheel now from our build. The reason is that we can simply say "if your CUDA installation is that old, use the pip version instead".

Reverted for now, though.

README.md
```diff
@@ -419,7 +419,43 @@ On Linux, it is often necessary to first update `pip` to a version that supports
...
+
+There are two ways to install JAX with CUDA support, either by installing CUDA
+using `pip` or by using a self-installed CUDA.
+
+#### Using a `pip`-installed copy of CUDA and CuDNN (recommended)
+
+NVIDIA provides almost everything JAX needs as wheels on PyPi with the exception
+of the GPU drivers. Installing CUDA from PyPi is the easiest way to install JAX
+and we strongly recommend it. The main downside is that NVIDIA's CUDA wheels are
+quite large (approximately 2GB).
```
yashk2810, line 434: This 2GB size seems concerning to me if we are going to tell people to use this as the default path. 

Downloading such large files everywhere is not possible.

hawkinsp: Note you're going to download the same things either way. It's just whether you download it via pip or some other way (apt, for example). I want to strongly encourage users to use this path to lower our support burden.

Added some text.

yashk2810: Ohh I see. That was not clear!

LGTM

yashk2810 (approved):

README.md
```diff
@@ -419,7 +419,46 @@ On Linux, it is often necessary to first update `pip` to a version that supports
 
 ### pip installation: GPU (CUDA)
 
-If you want to install JAX with both CPU and NVidia GPU support, you must first
+JAX provides pre-built CUDA wheels for **Linux only**. Other
```
skye, line 422: nit: "Installing on other operating systems is possible, ..."

hawkinsp:
@skye @nouiz @yhtang any thoughts?

(I'm going to commit this with the README changes commented out, because they will make no sense until we make a jaxlib release.)

skye (approved):
LGTM, some minor comments on readability

README.md
hawkinsp: Done.

README.md
```diff
@@ -436,26 +475,25 @@ source](https://jax.readthedocs.io/en/latest/developer.html#building-from-source
...
+    that NVIDIA provides for this purpose.
 
 Next, run
 
 ```bash
 pip install --upgrade pip
 # Installs the wheel compatible with CUDA 11 and cuDNN 8.2 or newer.
 # Note: wheels only available on linux.
-pip install --upgrade "jax[cuda]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
+pip install --upgrade "jax[cuda_using_existing_cuda_installation]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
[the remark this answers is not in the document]
hawkinsp: Went with `existing_cuda_installation`

jaxlib/cuda/BUILD
```diff
@@ -95,6 +95,13 @@ pybind_extension(
         "-fno-strict-aliasing",
     ],
     features = ["-use_header_modules"],
+    linkopts = select({
+        "@org_tensorflow//tensorflow/compiler/xla/python:use_jax_cuda_pip_rpaths": [
+            "-Wl,-rpath,$$ORIGIN/../../nvidia/cuda_runtime/lib",
```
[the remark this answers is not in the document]
hawkinsp: Yes that's exactly it. Left alone for now.

setup.py
```diff
@@ -91,9 +91,25 @@ def generate_proto(source):
         'australis': ['protobuf>=3.13,<4'],
 
         # CUDA installations require adding jax releases URL; e.g.
```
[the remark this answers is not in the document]
hawkinsp: Reworded.

hawkinsp:
This change is blocked on a missing library path in the CuDNN `pip` packages (NV partners bug 3848864 ).

yhtang (approved):

README.md
```diff
@@ -419,6 +419,50 @@ On Linux, it is often necessary to first update `pip` to a version that supports
...
+[CUDA forward compatibility packages](https://docs.nvidia.com/deploy/cuda-compatibility/)
+that NVIDIA provides for this purpose.
+
+
+To install JAX with CUDA support, run:
+```bash
+pip install --upgrade pip
+# Installs jax, jaxlib, CUDA, and CUDNN.
+# Note: wheels are only available on linux.
+pip install --upgrade "jax[cuda]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
nouiz, line 458: I suppose this doc section is hidden for now.
This line is exactly the same as in the manual install path.
How do you plan to support both?
By creating a `jax[cuda_manual]` installation target?

hawkinsp: Yes. I didn't update the docs yet because we need to make a release, but if you look at the `setup.py` change below it adds an `existing_cuda_installation` target.

nouiz: I missed that. That looks great. Thanks.

setup.py
```diff
@@ -90,16 +89,52 @@ def generate_proto(source):
...
+          "nvidia-cusolver-cu12",
+          "nvidia-cusparse-cu12",
+        ],
+
+        # Target that does not depend on the CUDA pip wheels, for those who want
+        # to use a preinstalled CUDA.
+        'cuda11_local': [
+          f"jaxlib=={_current_jaxlib_version}+cuda11.cudnn{_default_cuda11_cudnn_version}",
+        ],
+        'cuda12_local': [
```
nouiz, line 129: I like this. Very explicit and clear.

If we want, in a follow up, we could auto detect the cuda version to install and add the target: cuda_pip.
We can detect if there is some pypi nvidia packages already installed and if so, pick that version.
If no such packages are present, we could detect the driver installed. But I'm not sure how much cross-platform that could be.

PR #13015, opened 2022-10-27
Fix result type for `InspectSharding` custom call

jurahul:
- Use the result type that will be valid for the sharding that will be attached to this custom call.


jurahul:
@sharadmv can you please review? This is in preparation of an XLA change to verify that sharding attached to any instruction is valid for the shape of that instruction.

sharadmv (approved):

PR #13022, opened 2022-10-27
leak checker: explain why a tracer is alive by printing a reference chain

mattjj:
The idea here is that if you know what data structure(s) a leaked tracer is in, it might help you understand what the side-effect was that caused the problem.

With this code:

```python
import jax
jax.config.update('jax_check_tracer_leaks', True)

class A:
  def __init__(self, dct):
    self.dct = dct

a = A({})

def sketch(x):
  a.dct['hi'] = [lambda: x]
  return x

jax.vmap(sketch)(jax.numpy.arange(3))
```

instead of just this error:

```
Traced<ShapedArray(int32[])>with<BatchTrace(level=1/0)> with
  val = DeviceArray([0, 1, 2], dtype=int32)
  batch_dim = 0
This Tracer was created on line /usr/local/google/home/mattjj/packages/jax/qiao26.py:14 (<module>)

```

we now get this error:

```
Exception: Leaked trace MainTrace(1,BatchTrace). Leaked tracer(s):

Traced<ShapedArray(int32[])>with<BatchTrace(level=1/0)> with
  val = DeviceArray([0, 1, 2], dtype=int32)
  batch_dim = 0
This BatchTracer with object id 140443431328480 was created on line:
  /usr/local/google/home/mattjj/packages/jax/qiao26.py:14 (<module>)
<BatchTracer 140443431328480> is referred to by <function 140443431341072> closed-over variable x
<function 140443431341072> is referred to by <list 140443431333184>[0]
<list 140443431333184> is referred to by <dict 140445269059648>['hi']
<dict 140445269059648> is referred to by <A 140445263828256>.dct
<A 140445263828256> is referred to by __main__.a
```

That is, when the leak checker finds (via the `gc`) a `Tracer` which has outlived its trace, we attempt to walk back up the referrer chain for that `Tracer` and describe what objects we encounter (and even at which indices/names). In the example code, that might make us notice that the global object, and our update to `dct`, is a problem

We should also add info about which transformation or primitive the `Tracer` corresponded to (i.e. point to the last line of the file here). That may be in a follow-up PR though.

### DIFF
--- a/jax/core.py
+++ b/jax/core.py
@@ -20,6 +20,7 @@ from dataclasses import dataclass
 import functools
 from functools import partial, partialmethod, total_ordering
 import gc
+import inspect
 import itertools as it
 import operator
 from operator import attrgetter
@@ -888,7 +889,9 @@ def maybe_find_leaked_tracers(x: Optional[Union[MainTrace, Sublevel]]
   """
   if not getattr(threading.current_thread(), 'pydev_do_not_trace', True):
     warnings.warn(TRACER_LEAK_DEBUGGER_WARNING)
-  # Trigger garbage collection to filter out cyclical dependency false positives
+  # Trigger garbage collection to filter out unreachable objects that are alive
+  # only due to cyclical dependencies. (We don't care about unreachable leaked
+  # tracers since they can't interact with user code and cause a problem.)
   gc.collect()
   traces = list(filter(lambda x: isinstance(x, Trace), gc.get_referrers(x)))
   tracers = list(filter(lambda x: isinstance(x, Tracer), gc.get_referrers(*traces)))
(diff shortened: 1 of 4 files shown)


jax/core.py
```diff
@@ -888,17 +889,73 @@ def maybe_find_leaked_tracers(x: Optional[Union[MainTrace, Sublevel]]
...
+    # (if this heuristic causes problems, this code block can be commented out).
+    # TODO(mattjj): after Python 3.7 is unsupported, replace with types.CellType
+    cell_type = type((lambda x: lambda: x)(10.28).__closure__[0]) # type: ignore
+    if (isinstance(parent, dict) and
+        getattr(parents(parent)[0], '__dict__', None) is parents(child)[0]):
+      parent = parents(parent)[0]
+    elif type(parent) is cell_type:
+      parent = parents(parents(parent)[0])[0]
+
+    line = f'<{type(child).__name__} {id(child)}> is referred to by '
```
LenaMartens, line 924: for my understanding: why is the object ID useful when debugging? IIUC there will be no cycles in this trace, or might there be cycles?

tests/api_test.py
```diff
@@ -3554,9 +3554,37 @@ def apply_fn(inp):
...
+        r"<function [0-9]+> is referred to by <list [0-9]+>\[0\]\n"
+        r"<list [0-9]+> is referred to by <dict [0-9]+>\['hi'\]\n"
+        r"<dict [0-9]+> is referred to by <A [0-9]+>\.dct\n"
+    )
+
+    with jax.check_tracer_leaks():
+      with self.assertRaisesRegex(Exception, msg):
+        jax.vmap(sketch)(x)
+
+    def test_default_backend(self):
```
LenaMartens, line 3585: accidental indent?

LenaMartens (approved):
nice! cool error message

tests/api_test.py
mattjj: Oops, thank you!

jax/core.py
mattjj: There shouldn't be cycles in the printed result.

I found printing the id useful because I tend to cross-reference ids against print statements I put elsewhere or against interactive pdb spelunking. Also it's a (meaningful) way to distinguish different instances of e.g. a `dict`.

jax/core.py
```diff
@@ -888,17 +889,73 @@ def maybe_find_leaked_tracers(x: Optional[Union[MainTrace, Sublevel]]
...
+
+    # special-case namespaces and closures, handled in _why_alive_container_info
+    # (if this heuristic causes problems, this code block can be commented out).
+    # TODO(mattjj): after Python 3.7 is unsupported, replace with types.CellType
+    cell_type = type((lambda x: lambda: x)(10.28).__closure__[0]) # type: ignore
+    if (isinstance(parent, dict) and
+        getattr(parents(parent)[0], '__dict__', None) is parents(child)[0]):
+      parent = parents(parent)[0]
+    elif type(parent) is cell_type:
+      parent = parents(parents(parent)[0])[0]
```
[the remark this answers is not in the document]
mattjj: Yes, that's the idea: a little shortcut. Without this we get this printed:

```
<DynamicJaxprTracer 139764213674128> is referred to by <cell 139764213646192>
<cell 139764213646192> is referred to by <tuple 139764213646096>[0]
<tuple 139764213646096> is referred to by <function 139764213640064>
<function 139764213640064> is referred to by <dict 139766096500800>['hi']
<dict 139766096500800> is referred to by <A 139766089094432>.dct
<A 139766089094432> is referred to by __main__.a
```

but with it we get:

```
<DynamicJaxprTracer 140139354034480> is referred to by <function 140139405110576> (foo) closed-over variable x
<function 140139405110576> is referred to by <dict 140141242837056>['hi']
<dict 140141242837056> is referred to by <A 140141237605664>.dct
<A 140141237605664> is referred to by __main__.a
```

I'll add a comment.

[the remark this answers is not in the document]
mattjj:
```
    # For namespaces (like modules and class instances) and closures, the
    # references may form a simple chain: e.g. instance refers to its own
    # __dict__ which refers to child, or function refers to its __closure__
    # which refers to cells which refer to child. In these cases, we can provide
    # a more intuitive description by collapsing the chain into a single
    # parent->child jump. We do that by setting `parent` here to be a
    # grandparent (or great-grandparent) of `child`, and then handling that case
    # in _why_alive_container_info.
    # To prevent this collapsing behavior, just comment out this code block.
```

Issue #12307, opened 2022-09-09
jax[cuda] installation replaces current jax version with old jax-0.2.22 version


gianlucadetommaso:
### Description

When I run
```pip install --upgrade "jax[cuda]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html```
on my laptop, the latest version of jax, currently jax-0.3.17, gets replaced by jax-0.2.22. I don't think this was happening before. Here is a representative output of the command above:

```
Looking in links: https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
Collecting jax[cuda]
  Using cached jax-0.3.17.tar.gz (1.1 MB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.16.tar.gz (1.0 MB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.15.tar.gz (1.0 MB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.14.tar.gz (990 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.13.tar.gz (951 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.12.tar.gz (947 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.11.tar.gz (947 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.10.tar.gz (939 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.9.tar.gz (937 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.8.tar.gz (935 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.7.tar.gz (944 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.6.tar.gz (936 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.5.tar.gz (946 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.4.tar.gz (924 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.3.tar.gz (924 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.2.tar.gz (926 kB)
  Preparing metadata (setup.py) ... done
  Using cached jax-0.3.1.tar.gz (912 kB)
  Preparing metadata (setup.py) ... done
[...]

Issue #7097, opened 2021-06-24
Provide AArch64 (ARM) Linux jaxlib wheels


bensworth:
Potentially related to https://github.com/google/jax/issues/6932 or https://github.com/google/jax/issues/7052.

I am using a heterogeneous cluster with various different compute options. I got Jax and Flax installed fine on the CPU system. Then I tried using a GPU node (specifically dual socket Cavium ThunderX2 99xx 32-core processors with two NVIDIA Tesla V100 GPUs), but am unable to install jaxlib:
```
$ pip install --upgrade jax jaxlib==0.1.68+cuda111 -f https://storage.googleapis.com/jax-releases/jax_releases.html
Looking in links: https://storage.googleapis.com/jax-releases/jax_releases.html
Requirement already satisfied: jax in /home/southworth/.conda/envs/jax_arch/lib/python3.7/site-packages (0.2.16)
ERROR: Could not find a version that satisfies the requirement jaxlib==0.1.68+cuda111 (from versions: none)
ERROR: No matching distribution found for jaxlib==0.1.68+cuda111
```
or even
```
pip install jaxlib
ERROR: Could not find a version that satisfies the requirement jaxlib (from versions: none)
ERROR: No matching distribution found for jaxlib
```
Below is my pip version, which is up to date. Pip also installs other libraries fine (numpy, scipy, etc.)
```
$ pip --version
pip 21.1.2 from /home/southworth/.conda/envs/jax_arch/lib/python3.7/site-packages/pip (python 3.7)
```

hawkinsp:
The issue is that we don't provide Linux AArch64 wheels for jaxlib. Cavium is an ARM machine.

I suspect you may be able to build jaxlib from source on those machines: https://jax.readthedocs.io/en/latest/developer.html#building-from-source

Try it out and see how it goes?

(I'm not sure we have the resources to make ARM builds right now, but I wouldn't rule it out for the future.)

bensworth:
Ah, that makes sense. Okay, for now I switched to a single socket Xeon E3-1585L v5 4-core processors with integrated Iris Pro P580 GPU to try and get something working. Here the wheels work fine and everything installs. I loaded a cuda/11.1 module and confirmed it is installed at `/usr/local/cuda-11.1/`, `nvcc` works as expected, and I confirmed the device via `lshw -C display`, but Jax cannot find any devices:
```
[...]

PR #13024, opened 2022-10-27
Add x86_64 dependency note to pip installation

treyra:
Currently non x86_64 linux architectures are not supported, see #7097 for request to change this. This can lead to installation confusion, as jax will install, but jaxlib will not. For example see #12307. 

If this can be more clearly phrased or explained, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,7 +413,7 @@ pip install --upgrade "jax[cpu]"
 ```
 
 On Linux, it is often necessary to first update `pip` to a version that supports
-`manylinux2014` wheels.
+`manylinux2014` wheels. Also note that for Linux, we currently release wheels for `x86_64` architectures only, other architectures require building from source. Trying to pip install with other Linux architectures may lead to `jaxlib` not being installed alongside `jax`, although `jax` may successfully install (but fail at runtime). 
 **These `pip` installations do not work with Windows, and may fail silently; see
 [above](#installation).**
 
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks! Should we add the same caveat under the GPU instructions?

mjsML:
For the current state I think this makes sense ... It would change soon-ish with Grace and Grace-Hopper

treyra:
@jakevdp good point, I added that too.

mjsML:
Forgot about jetsons too 😅

treyra:
Yeah, a Jetson Orin was what I was trying to install Jax on and had to build from source for. The good news is building from source was pretty easy! But this would have saved me some time debugging why pip appeared to work, but jaxlib didn't install.

jakevdp (approved):

README.md
```diff
@@ -413,7 +413,7 @@ pip install --upgrade "jax[cpu]"
 ```
 
 On Linux, it is often necessary to first update `pip` to a version that supports
-`manylinux2014` wheels.
+`manylinux2014` wheels. Note these existing wheels are currently for `x86_64` architectures only. Trying to pip install with other architectures may lead to `jaxlib` not being installed alongside `jax`, although `jax` may successfully install.
```
hawkinsp, line 416: The wording here is a bit misleading, because we do build and release Mac ARM wheels. So you should probably put "Linux" somewhere. Perhaps reword this change as:
"Note that on Linux, we release wheels for the `x86_64` architecture only at the moment. Other architectures require building from source."

treyra: Good point, I reworded it to emphasize that caveat applies to Linux architectures only.

hawkinsp (approved):

treyra:
Let me know if there is anything else needed from me before this is merged.

I don't seem to have permissions to re-run the Read the Docs build, but from reading the output, it seems to be just two warnings, neither related to my edits. The first was that https://docs.scipy.org/doc/scipy-1.8.1/objects.inv wasn't fetchable, which seems to have been temporary and resolved. The other was undefined label: ndimage-interpolation-modes in jax._src.scipy.ndimage.map_coordinates. If wanted I can make an empty commit to force it to re-run.

jakevdp:
Thanks @treyra - one last thing, could you please squash the changes into a single commit? (See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Thanks!

treyra:
Done! Thanks @jakevdp for walking me through how to contribute for the first time. Hopefully in the future I can make some more substantial contributions.

jakevdp:
Thanks for adding this clarification! Sorry it took so long to get this merged 😀

PR #13032, opened 2022-10-28
Declare Array.sharding & raise an error on tracers

jakevdp:
Part of #12049


yashk2810 (approved):

jakevdp:
Stalled because we have a number of `hasattr(x, 'sharding')` checks that previously returned False and now return True

PR #12798, opened 2022-10-14
Add new config `jax_persistent_cache_min_instruction_count`.

skye:
This can be used to limit the number of entries written to the persistent compilation cache.

I defaulted to setting 6 as the minimum threshold based on running the flax wmt example
(https://github.com/google/flax/tree/main/examples/wmt) and logging the instruction counts and complilation time:

name | instruction_count | compile_time_secs
---- | ----------------- | -----------------
`broadcast_in_dim` | 2 | 0.01633763313
`convert_element_type` | 2 | 0.01704716682
`reshape` | 2 | 0.01730203629
`_squareit` | 2 | 0.01730823517
`broadcast_in_dim` | 2 | 0.0182030201
`convert_element_type` | 2 | 0.01982188225
`concatenate` | 2 | 0.02102327347
`true_divide` | 2 | 0.02172231674
`broadcast_in_dim` | 2 | 0.02370619774
`broadcast_in_dim` | 2 | 0.02393102646
`broadcast_in_dim` | 2 | 0.02488565445
`broadcast_in_dim` | 2 | 0.03395628929
`broadcast_in_dim` | 2 | 0.03428125381
`broadcast_in_dim` | 2 | 0.0394551754
`shift_right_logical` | 2 | 0.06500506401
`<lambda>` | 3 | 0.01793265343
`_unstack` | 5 | 0.01975226402
`_reduce_sum` | 5 | 0.0210878849
`_reduce_sum` | 5 | 0.02416801453
`_multi_slice` | 9 | 0.09065580368
`_threefry_split` | 232 | 0.09037566185
`_threefry_split` | 232 | 0.09161829948
`<unnamed wrapped function>` | 2668 | 7.701903343
`<unnamed wrapped function>` | 3455 | 17.57672167
`<unnamed wrapped function>` | 46580 | 166.2570884
`init` | 60361 | 26.35722399
`<unnamed wrapped function>` | 78010 | 3.879326344

Also adds new int config functionality.

Fixes #12583

PR #13039, opened 2022-10-28
Add new config `jax_persistent_cache_min_compile_time_secs`.

skye:
This replaces `jax_persistent_cache_min_instruction_count` introduced in https://github.com/google/jax/pull/12798, since gating on the compile time seems strictly better than gating on the instruction count (except maybe that the instruction count is more deterministic, but I don't think that's a big deal).

I defaulted to 1 second as the minimum threshold based on the same flax wmt example
(https://github.com/google/flax/tree/main/examples/wmt) numbers from

name | instruction_count | compile_time_secs
---- | ----------------- | -----------------
`broadcast_in_dim` | 2 | 0.01633763313
`convert_element_type` | 2 | 0.01704716682
`reshape` | 2 | 0.01730203629
`_squareit` | 2 | 0.01730823517
`broadcast_in_dim` | 2 | 0.0182030201
`convert_element_type` | 2 | 0.01982188225
`concatenate` | 2 | 0.02102327347
`true_divide` | 2 | 0.02172231674
`broadcast_in_dim` | 2 | 0.02370619774
`broadcast_in_dim` | 2 | 0.02393102646
`broadcast_in_dim` | 2 | 0.02488565445
`broadcast_in_dim` | 2 | 0.03395628929
`broadcast_in_dim` | 2 | 0.03428125381
`broadcast_in_dim` | 2 | 0.0394551754
`shift_right_logical` | 2 | 0.06500506401
`<lambda>` | 3 | 0.01793265343
`_unstack` | 5 | 0.01975226402
`_reduce_sum` | 5 | 0.0210878849
`_reduce_sum` | 5 | 0.02416801453
`_multi_slice` | 9 | 0.09065580368
`_threefry_split` | 232 | 0.09037566185
`_threefry_split` | 232 | 0.09161829948
`<unnamed wrapped function>` | 2668 | 7.701903343
`<unnamed wrapped function>` | 3455 | 17.57672167
`<unnamed wrapped function>` | 46580 | 166.2570884 `init` | 60361 | 26.35722399
`<unnamed wrapped function>` | 78010 | 3.879326344

Also adds new float config functionality.

### DIFF
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -1066,9 +1066,11 @@ def compile_or_get_cached(backend, computation: ir.Module, compile_options,
       logger.info("Persistent compilation cache hit for '%s'", module_name)
       return cached_executable
     else:
+      start_time = time.monotonic()
       compiled = backend_compile(backend, serialized_computation,
                                  compile_options, host_callbacks)
-      _cache_write(computation, serialized_computation, module_name,
+      compile_time = time.monotonic() - start_time
+      _cache_write(serialized_computation, compile_time, module_name,
                    compile_options, backend, compiled)
       return compiled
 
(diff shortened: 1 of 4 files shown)


jax/_src/dispatch.py
```diff
@@ -1066,9 +1066,11 @@ def compile_or_get_cached(backend, computation: ir.Module, compile_options,
       logger.info("Persistent compilation cache hit for '%s'", module_name)
       return cached_executable
     else:
+      start_time = time.time()
```
ppwwyyxx, line 1069: nit: use `time.monotonic()`?

hawkinsp (approved):

tests/compilation_cache_test.py
```diff
@@ -335,21 +337,23 @@ def test_cache_read_warning(self):
...
+    with tempfile.TemporaryDirectory() as tmpdir, \
+         persistent_cache_min_compile_time_secs(2):
       cc.initialize_cache(tmpdir)
 
-      with persistent_cache_min_instruction_count(20):
-        # 2 instructions at time of writing
-        jit(lambda x: x * x)(2)
-        files_in_cache = len(os.listdir(tmpdir))
-        self.assertEqual(files_in_cache, 0)
+      # Takes between .02 and .08 seconds to compile as of Oct 28 2023
```
hawkinsp, line 345: I recommend against testing this way because the timing is very environment-dependent. If you have to do it, I would mock the time.

jax/_src/dispatch.py
skye: Didn't know about this, thanks! Done.

tests/compilation_cache_test.py
skye: Good idea, I mocked the time.

PR #13057, opened 2022-11-01
Allow passing empty path to compile cache

lucasb-eyer:
Quite a few codebases which use the compile cache currently follow this pattern:

```
if cache_path:
  compile_cache.initialize_cache(cache_path)
```

In these situations, it would be nicer to allow just the one-liner call without the `if`.


lucasb-eyer:
Hi, can I get a yes/no on this?

PR #13061, opened 2022-11-01
More documentation about GPU tests

nouiz:
Here is more documentation about GPU tests.

@hawkinsp

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -180,6 +180,22 @@ A number of test behaviors can be controlled using environment variables (see
 below). Environment variables may be passed to JAX tests using the
 `--test_env=FLAG=value` flag to Bazel.
 
+Some of JAX tests are for multiple accelerators (i.e. GPUs, TPUs). When JAX is already installed, you can run GPUs tests like this:
+
+```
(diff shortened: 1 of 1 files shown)


docs/developer.md
```diff
@@ -180,6 +180,22 @@ A number of test behaviors can be controlled using environment variables (see
 below). Environment variables may be passed to JAX tests using the
 `--test_env=FLAG=value` flag to Bazel.
 
+Some of JAX tests are for multiple accelerators (i.e. GPUs, TPUs). When JAX is already installed, you can run GPUs tests like this:
+
+```
+bazel test //tests:gpu_tests --jobs=4 --test_tag_filters=multiaccelerator --//jax:build_jaxlib=false --test_env=XLA_PYTHON_CLIENT_ALLOCATOR=platform
+```
+
+You can speed up single accelerator tests by running them in parallel on the same acceleator and on multiple accelarator. For GPUs, you can do it like this:
```
hawkinsp, line 189: Nit: "You can speed up single accelerator tests by running them in parallel on multiple accelarators:"

Typo and you can delete some words.

nouiz: I amended the commit. I also wanted to say that there is multiple tests on the same GPU. I said that in a new sentence to make this clearer.

hawkinsp (approved):

Issue #9960, opened 2022-03-18
Add control flow operators for iterated functions


carlosgmartin:
Add the following [control flow operators](https://jax.readthedocs.io/en/latest/jax.lax.html#control-flow-operators) for [iterated functions](https://en.wikipedia.org/wiki/Iterated_function):

```python
# iterate : ∀a. ℕ → (a → a) → (a → a)
def iterate(n, f, x):
    for _ in range(n):
        x = f(x)
    return x

# orbit : ∀a. ℕ → (a → a) → (a → [a])
def orbit(n, f, x):
    xs = [x]
    for _ in range(n):
        x = f(x)
        xs.append(x)
    return xs
```

`iterate` and [`fori_loop`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html) are mutually definable, but the former has a simpler signature and semantics that is more common in my experience, and is arguably more "basic" or "fundamental".

jakevdp:
Interesting idea - if I understand your suggestion correctly, I think the implementations would look like this:
```python
def iterate(n, f, x):
  return lax.scan(lambda x, y: (f(x), x), x, None, length=n)[0]

def orbit(n, f, x):
  return lax.scan(lambda x, y: (f(x), x), x, None, length=n + 1)[1]
```
Is that what you have in mind?

mattjj:
For `iterate` we may want `lax.scan(lambda x, _: (f(x), None), x, None, length=n)[0]`, i.e. don't have an extensive output.

carlosgmartin:
@jakevdp @mattjj Looks right:
```python
from jax.numpy import array
from jax.lax import scan

def iterate_1(n, f, x):
    for _ in range(n):
        x = f(x)
    return x

def iterate_2(n, f, x):
  return scan(lambda x, _: (f(x), x), x, None, length=n)[0]

def iterate_3(n, f, x):
    return scan(lambda x, _: (f(x), None), x, None, length=n)[0]

def orbit_1(n, f, x):
    xs = [x]
    for _ in range(n):
        x = f(x)
        xs.append(x)
    return array(xs)

def orbit_2(n, f, x):
    return scan(lambda x, _: (f(x), x), x, None, length=n + 1)[1]

n = 10
f = lambda x: 2 * x + 1
x = 3
print(iterate_1(n, f, x)) # 4095
print(iterate_2(n, f, x)) # 4095
print(iterate_3(n, f, x)) # 4095
[...]

Issue #2469, opened 2020-03-20
Implement an option for lax.while_loop to specify the maximum number of iterations, to allow reverse differentiation


gnecula:
Several forms of loops in JAX support reverse AD: `scan`, `fori_loop` with constant bounds, which is syntactic sugar for `scan`. I think it could be useful to have another syntactic sugar for bounded loops by adding a parameter `max_iterations` to `while_loop`.

tavin:
I can't say why it works, but I'm using `while_loop` with `jacrev`. Nevertheless I would like to see this enhancement, as it's a bit of a pain to take care of this in the loop state and cond/body functions.

carlosgmartin:
I'd like to see this as well. Here's a possible implementation

```python3
def while_loop(cond_fun, body_fun, init_val, max_iters):
    def f(val, _):
        val = lax.cond(cond_fun(val), lambda: body_fun(val), lambda: val)
        return val, None

    val, _ = lax.scan(f, init_val, None, length=max_iters)
    return val


def test_while_loop():
    def cond_fun(x):
        return x < 100

    def body_fun(x):
        return x * 2

    init_val = 1

    out_1 = lax.while_loop(cond_fun, body_fun, init_val)
    out_2 = while_loop(cond_fun, body_fun, init_val, 20)
    print(out_1, out_2)
```

NeilGirdhar:
JaxOpt's implementation has been ironed out: https://github.com/google/jaxopt/blob/main/jaxopt/_src/loop.py

Issue #2139, opened 2020-02-02
scan with gradient checkpointing


shoyer:
It would be great to have a version of lax.scan used a recursive gradient checkpointing (“binomial checkpointing”) that allows for differentiating through long time series with logarithmic time/space costs.

In principle this could be built on top of the experimental `remat` decorator: https://github.com/google/jax/pull/1749

shoyer:
Support for a single level of gradient checkpointing might also be useful, which requires only one extra forward pass. Apparently the optimal way to do it reduces memory usage to the square root of the number of steps, e.g., per this implementation in TensorFlow:
https://github.com/cybertronai/gradient-checkpointing

hawkinsp:
There is also a class of algorithms that optimize to a fixed memory budget: https://dl.acm.org/doi/10.1145/347837.347846

(I'm not sure they are worth it over the simpler strategies though.)

shoyer:
Yep, I realize now that that's what "binomial checkpointing" in particular means. I was originally thinking of something simpler, just using recursion.

patrick-kidger:
So [Diffrax](https://github.com/patrick-kidger/diffrax) actually implements a `bounded_while_loop` that does exactly this -- early exit by nesting scan-conds, and managing memory using recursive checkpointing. In Diffrax's case it's used to handle the stepping of a differential equation solver.

The implementation is here: https://github.com/patrick-kidger/diffrax/blob/2b4e4d863c15abc7143919bac7825090bbfe50be/diffrax/misc/bounded_while_loop.py

It's worth noting that there are a *lot* of caveats that need to be worked around in order to make something like feasible.
- If vmap is handled naively, then the `lax.cond`s used to handle whether to evaluate `body_fun` or simply perform an identity operation will get turned into `lax.select`s, and the entire point of the efficiency gains are lost. This has to be worked around with some custom vmap behaviour. (In particular an `unvmap` operation.)
- In-place updates are a nightmare. XLA:CPU fails to optimise away the copies, as per JAX issues
#8192
and
#9132.
Working around this requires changing the signature of `body_fun` with a custom way of handling in-place updates:
https://github.com/patrick-kidger/diffrax/blob/2b4e4d863c15abc7143919bac7825090bbfe50be/diffrax/misc/bounded_while_loop.py#L25
Moreover XLA clearly has some bugs because you can improve performance of nested in-place updates (in nested `bounded_while_loop`s) by adding dead code that doesn't actually evaluate to anything (!!!)
[...]

PR #13062, opened 2022-11-01
Added `scan(early_exit=...)` and `while_loop(max_steps=...)`

patrick-kidger:
TL;DR: bounded reverse-mode differentiable while loops.

Wanting to have an "early exit scan" or a "bounded while loop" is a request that has come up a lot:
https://github.com/google/jax/issues/2139
https://github.com/google/jax/issues/2469
https://github.com/google/jax/issues/5642
https://github.com/google/jax/discussions/8375
https://github.com/google/jax/issues/9960
etc.

In particular for my use-cases, this operation is the one needed inside a differential equation solver. Up until now Diffrax had had a [pretty complicated/inefficient/hacky workaround](https://github.com/google/jax/issues/2139#issuecomment-1039293633).

So anyway! I'm happy to say that this has finally become a big enough problem that I sat down to implement this properly. Building in support for this in Diffrax, I've even seen some nice compile-time improvements: the best-so-far, proportionally, is a reduction from 28 seconds to 5 seconds, just by switching out the bounded while loop implementation.

WDYT?

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -41,6 +41,7 @@ from jax._src import util
 from jax._src.lax import lax
 from jax._src.lax import slicing
 from jax._src.lax import windowed_reductions
+from jax._src.lax.control_flow import conditionals
 from jax._src.lib.mlir import ir
 from jax._src.lib.mlir.dialects import mhlo
 from jax._src.traceback_util import api_boundary
@@ -111,7 +112,9 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
          xs: X,
          length: Optional[int] = None,
          reverse: bool = False,
-         unroll: int = 1) -> Tuple[Carry, Y]:
+         unroll: int = 1,
+         early_exit: Optional[Callable[[Carry], bool]] = None,
+         ) -> Tuple[Carry, Y]:
   """Scan a function over leading array axes while carrying along state.
 
   The `Haskell-like type signature`_ in brief is
@@ -155,10 +158,6 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
   dtype (or a nested tuple/list/dict container data structure with a fixed
   structure and arrays with fixed shape and dtype at the leaves).
 
-  .. note::
-    :py:func:`scan` compiles ``f``, so while it can be combined with
-    :py:func:`jit`, it's usually unnecessary.
-
   Args:
     f: a Python function to be scanned of type ``c -> a -> (c, b)``, meaning
       that ``f`` accepts two arguments where the first is a value of the loop
@@ -181,6 +180,8 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
     unroll: optional positive int specifying, in the underlying operation of the
       scan primitive, how many scan iterations to unroll within a single
       iteration of a loop.
+    early_exit: optional callable. Called at each step; if it returns True
+      then the scan is halted early.
 
   Returns:
     A pair of type ``(c, [b])`` where the first element represents the final
@@ -191,6 +192,9 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
   """
   if not callable(f):
     raise TypeError("lax.scan: f argument should be a callable.")
+  if early_exit is None: early_exit = lambda carry: False
+  if not callable(early_exit):
+    raise TypeError("lax.scan: early_exit argument should be a callable or None.")
   xs_flat, xs_tree = tree_flatten(xs)
 
   try:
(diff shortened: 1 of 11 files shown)


jax/_src/lax/control_flow/loops.py
```diff
@@ -155,10 +158,6 @@ def scan(f, init, xs, length=None):
   dtype (or a nested tuple/list/dict container data structure with a fixed
   structure and arrays with fixed shape and dtype at the leaves).
 
-  .. note::
-    :py:func:`scan` compiles ``f``, so while it can be combined with
-    :py:func:`jit`, it's usually unnecessary.
-
```
patrick-kidger, line 161: Drive-by: I removed these whilst I was editing the docstrings, as right now JIT'd code is the only fast path, so I think this advice is contrary to good practice.

jax/_src/lax/control_flow/loops.py
```diff
@@ -219,18 +223,40 @@ def scan(f, init, xs, length=None):
...
+      pass
+
+  def aug_f(carry, x):
+    out = f(carry, x)
+    if not hasattr(out, "__len__") or len(out) != 2:
+      msg = "scan body output must be a pair, got {}."
+      raise TypeError(msg.format(out))
+    carry, y = out
+    done = early_exit(carry)
+    return done, carry, y
```
patrick-kidger, line 259: As you can see, the approach I took was to add the conditional as another output of the body, rather than as a separate jaxpr.

Upside: this really helped to minimise the amount of code changes needed elsewhere. No keeping track of separate constants for both `f` and `early_exit`, no extra fixed-point resolution needed, etc.

Downside: `early_exit` is traced/compiled twice; once at initialisation and once in the body.

Conditionals are generally small and cheap, so I felt this was a worthy trade-off here.

jax/_src/lax/control_flow/loops.py
```diff
@@ -263,106 +286,197 @@ def _create_jaxpr(init):
...
   disallowed_effects = jaxpr.effects - allowed_effects
   if disallowed_effects:
     raise NotImplementedError(
         f'Effects not supported in `scan`: {disallowed_effects}')
 
-  out = scan_p.bind(*consts, *in_flat,
+  # use default precision with no weak types
+  start = jax.numpy.array([0])[0]
+  stop = jax.numpy.array([length])[0]
+  done = jax.numpy.array(early_exit(init), dtype=np.bool_)
```
patrick-kidger, line 299: I suspect there's a better way to accomplish this but I don't know what it is.

patrick-kidger: Besides introducing the extra `done` variable, we also introduce `start` and `stop`. The iteration happens over 'xs[start:stop]`. (In particular inclusive of `start` but exclusive of `stop`.) This is needed when backpropagating through an early-exit'd scan: we would like to start from partway through the backward iterations.

jax/_src/lax/control_flow/loops.py
```diff
@@ -263,106 +286,197 @@ def _create_jaxpr(init):
...
+                    linear=(False,) * (3 + len(consts) + len(in_flat)),
                     unroll=unroll)
-  return tree_unflatten(out_tree, out)
-
-def _scan_impl_unrolled(*args, reverse, length, num_consts, num_carry, linear,
-                        f_impl, x_avals, y_avals):
-  consts, init, xs = split_list(args, [num_consts, num_carry])
+  out = out[2:]  # trim off start and stop
+  _, carry, ys = tree_unflatten(out_tree, out)  # trim off done
+  return carry, ys
```
patrick-kidger, line 307: The primitive returns `start`, `stop`, and `done` for use as residuals in the rules below. (And so they need to be trimmed off here before returning to the user.)

jax/_src/lax/control_flow/loops.py
```diff
@@ -797,58 +1111,84 @@ def _scan_dce_rule(used_outputs: List[bool], eqn: core.JaxprEqn
       used_carry_out = _map(operator.or_, used_carry_out, used_carry_in)
   else:
     assert False, "Fixpoint not reached"
-  if config.jax_enable_checks: core.check_jaxpr(jaxpr.jaxpr)
+  if config.jax_enable_checks: core.check_jaxpr(jaxpr_dce)
```
patrick-kidger, line 1114: Drive-by: I think this was just a bug in the original code.

jax/_src/lax/control_flow/loops.py
```diff
@@ -797,58 +1111,84 @@ def _scan_dce_rule(used_outputs: List[bool], eqn: core.JaxprEqn
...
+  [start_uk, stop_uk, done_uk], const_uk, carry_uk, xs_uk = split_list(unks_in, [3, num_consts, num_carry])
+  if done_uk:
+    # If conditional is unknown, just do the default processing.
+    eqn_known = None
+    eqn_staged = eqn
+    unks_out = [True] * len(eqn.outvars)
+    inst_out = [True] * len(eqn.outvars)
+    new_vars = [x for inst, x in zip(inst_in, eqn.invars)
+                if type(x) is core.Var and not inst]
+    return eqn_known, eqn_staged, unks_out, inst_out, new_vars
```
patrick-kidger, line 1146: I don't think there's a hook for this default processing, so I essentially just ended up copying what `pe.partial_eval_jaxpr_custom` does by default.

jax/core.py
```diff
@@ -2464,9 +2464,9 @@ class JaxprTypeError(TypeError): pass
 
 custom_typechecks: Dict[Primitive, Callable] = {}
 
-def _check_closed_call(*in_atoms, call_jaxpr):
+def _check_closed_call(*in_atoms, call_jaxpr, name=None):
```
patrick-kidger, line 2467: Drive-by bugfix, that meant I was otherwise getting a crash.

tests/lax_control_flow_test.py
```diff
@@ -2248,6 +2256,119 @@ def f(c, a):
         len(str(jax.xla_computation(scan)(c, xs).as_hlo_text())),
         len(str(jax.xla_computation(scan_unrolled)(c, xs).as_hlo_text())))
 
+  @parameterized.named_parameters(
+      {"testcase_name": f"_{scan_name}",
+       "scan": scan_impl}
+      for scan_impl, scan_name in SCAN_IMPLS + [(scan_reference, "reference")])
```
patrick-kidger, line 2262: The reference implementation is included to test the test.

patrick-kidger:
Hmmm, looks like there's some differences between me running pytest locally and the way it's ran here. I'll have to start digging through these test errors and figure out what's going on.

jakevdp:
It looks like an X64 test failing; you could try `JAX_ENABLE_X64=1 pytest -n auto tests`

mattjj:
I'd be happy to start taking a look, but I'm on paternity leave until end of November!

patrick-kidger:
Reporting in to say that I've now benchmarked this on our production code, and that this has improved our runtime from 10 seconds-per-iteration down to 6 seconds-per-iteration. (!)

carlosgmartin:
@patrick-kidger Could an early exit be added to [`fori_loop`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html) as well (#8375)? This would be useful for models that run for a variable number of iterations, e.g.
1. [Can You Learn an Algorithm? Generalizing from Easy to Hard Problems with Recurrent Networks](https://arxiv.org/abs/2106.04537)
2. [PonderNet: Learning to Ponder](https://arxiv.org/abs/2107.05407)
3. [End-to-end Algorithm Synthesis with Recurrent Networks: Logical Extrapolation Without Overthinking](https://arxiv.org/abs/2202.05826)

epignatelli:
I understand this is old and probably superseded. Has this been implemented?
If not, is this PR going anywhere?

patrick-kidger:
Unfortunately I believe this PR is dead. In the end the decision was that this was a change to a central part of the JAX codebase, and that was probably too risky to merge.

**Fortunately**, there is an alternative! Try `equinox.internal.while_loop(..., kind="checkpointed")`. This is a backpropagatable while loop using recursive checkpointing.

epignatelli:
That's a pity, there could be uses cases of interest. Jitting recursive sampling with a termination criterion is the first that comes to mind (LLMs?).

I'll have a look at the `equinox.internal.while_loop(..., kind="checkpointed")`, thanks @patrick-kidger !

carlosgmartin:
I'd love to see this functionality in JAX, as well.

PR #13078, opened 2022-11-02
[sparse] simplify BCOO indexing implementation

jakevdp:
This will also let us automatically inherit improvements like #12219


froystig (approved):
nice!

PR #13080, opened 2022-11-02
BUG: fix jet rule for dynamic_slice

### LINKED ISSUE #12263: jet of dynamic slice does not match jet of gather
This is blocking #12219

For example:
```python
import jax.numpy as jnp
from jax import lax
from jax.experimental.jet import jet

f_gather = lambda x: x[1:]
f_dynamic_slice = lambda x: lax.dynamic_slice(x, (1,), (len(x) - 1,))

# These two functions by design have identical behavior for 1D inputs:
x = jnp.arange(4.0)
print(f_gather(x))
# [1. 2. 3.]
print(f_dynamic_slice(x))
# [1. 2. 3.]

# However, their jet transforms do not match
h = 0.1 * x
print(jet(f_gather, (x,), ((h,),)))
# (DeviceArray([1., 2., 3.], dtype=float32), [DeviceArray([0.1, 0.2, 0.3], dtype=float32)])
print(jet(f_dynamic_slice, (x,), ((h,),)))
# (DeviceArray([1., 2., 3.], dtype=float32), [DeviceArray([0. , 0.1, 0.2], dtype=float32)])
```
It looks like the jet rule for dynamic_slice is not behaving correctly.

jakevdp:
Fixes #12263 & unblocks #12219

### DIFF
--- a/tests/jet_test.py
+++ b/tests/jet_test.py
@@ -311,7 +311,7 @@ class JetTest(jtu.JaxTestCase):
   @jtu.skip_on_devices("tpu")
   def test_cummin(self):     self.unary_check(partial(lax.cummin, axis=0))
   @jtu.skip_on_devices("tpu")
-  def test_dynamic_slice(self):     self.unary_check(partial(lax.dynamic_slice, start_indices=(0,0), slice_sizes=(1,1)))
+  def test_dynamic_slice(self): self.unary_check(partial(lax.dynamic_slice, start_indices=(1,2), slice_sizes=(1,1)))
   @jtu.skip_on_devices("tpu")
   def test_dynamic_update_slice(self): self.unary_check(partial(lax.dynamic_update_slice, start_indices=(1,2), update=jnp.arange(6.0).reshape(2, 3)))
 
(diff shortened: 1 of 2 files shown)


tests/jet_test.py
```diff
@@ -311,7 +311,7 @@ def test_cummax(self):     self.unary_check(partial(lax.cummax, axis=0))
   @jtu.skip_on_devices("tpu")
   def test_cummin(self):     self.unary_check(partial(lax.cummin, axis=0))
   @jtu.skip_on_devices("tpu")
-  def test_dynamic_slice(self):     self.unary_check(partial(lax.dynamic_slice, start_indices=(0,0), slice_sizes=(1,1)))
+  def test_dynamic_slice(self):     self.unary_check(partial(lax.dynamic_slice, start_indices=(1,2), slice_sizes=(1,1)))
```
jakevdp, line 314: Explanation here: previously the gradient terms were being computing using the gradient of `start_indices`, which is always zero, so this test only passed because all `start_indices` were zero.

mattjj (approved):

Issue #9445, opened 2022-02-04
Jax numpy array_split is slow compared to numpy


pseudo-rnd-thoughts:
I was writing Jax code, substituting all of the places where I would normally use NumPy with jax.numpy
In using jax.numpy.array_split with a large input, I found that the program was incredibly slow (for jax+numpy or jax then the program didn't finish in 5 minutes for me) 

While running there was no issue however if I ctrl+c to terminate then the follow error was thrown
`F external/org_tensorflow/tensorflow/compiler/xla/service/gpu/nvptx_compiler.cc:446] ptxas returned an error during compilation of ptx to sass: 'INTERNAL: ptxas exited with non-zero error code 2, output: '  If the error message indicates that a file could not be written, please verify that sufficient filesystem space is provided.
Aborted (core dumped)`

It seems that Jax is trying to jit/compile the array input which is slow due to the size, I just wanted to check if this was expected or not
If this is expected, could some documentation be added to note this problem on the relative functions 

Example code
```
from tqdm import tqdm
import numpy as onp
import jax.numpy as jnp

print('numpy')
a = [batch for batch in tqdm(onp.array_split(onp.arange(10_000), 1_000))]
b = [batch for batch in tqdm(onp.array_split(onp.arange(100_000), 10_000))]

print('jax + numpy')
c = [batch for batch in tqdm(jnp.array_split(onp.arange(100_000), 10_000))]
d = [batch for batch in tqdm(jnp.array_split(onp.arange(100_000), 10_000))]

print('jax')
e = [batch for batch in tqdm(jnp.array_split(jnp.arange(100_000), 10_000))]
f = [batch for batch in tqdm(jnp.array_split(jnp.arange(100_000), 10_000))]
```

Thanks for any help

jakevdp:
Thanks for the report – I think the issue is that `jnp.array_split` fails to push the numpy input to device before processing it, meaning that the splitting is effectively done on client side and then each output is pushed to device individually, incurring that overhead thousands of times instead of one time.

I'll work on a fix.

jakevdp:
Good point - my first PR gives an order of magnitude improvement by calling device_put, so I'll get that in first and then take a look at `dynamic_slice`.

jakevdp:
Re-opening to explore using `dynamic_slice` instead

jakevdp:
I tried switching to `lax.dynamic_slice`; for 1000 splits

- non-jitted execution is about 5x slower than `lax.slice`
[...]

PR #13096, opened 2022-11-03
jnp.split: use dynamic rather than static slices for speed

### LINKED ISSUE #12999: `jnp.split()` has long compile times for large numbers of splits
Reported in https://stackoverflow.com/q/74199437

```python
import jax.numpy as jnp
x = jnp.ones(5000)
%time _ = jnp.split(x, jnp.arange(0, 5000, 10))
# CPU times: user 7.29 s, sys: 101 ms, total: 7.39 s
# Wall time: 7.46 s
```
Subsequent calls are much faster, once the `XLA:Slice` operations have been cached:
```python
%time _ = jnp.split(x, jnp.arange(0, 5000, 10))
# CPU times: user 61.4 ms, sys: 0 ns, total: 61.4 ms
# Wall time: 70.3 ms
```
There are some tradeoffs here that make the solution to this non-trivial (see my answer in the above StackOverflow post), but it might be better to use `lax.dynamic_slice` in place of `lax.slice` to prevent this kind of catastrophically-bad outcome.

jakevdp:
Fixes #12999. See also #9445

This is very similar to the issue & fix in #12219

main branch:
```python
   ...: x = jnp.ones(5000) 
   ...: %time _ = jnp.split(x, jnp.arange(0, 5000, 10))                                                                   
CPU times: user 4.77 s, sys: 118 ms, total: 4.89 s
Wall time: 4.96 s
```
this branch:
```python
   ...: x = jnp.ones(5000) 
   ...: %time _ = jnp.split(x, jnp.arange(0, 5000, 10))                                                                   
CPU times: user 342 ms, sys: 8.52 ms, total: 351 ms
Wall time: 360 ms
```

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1191,9 +1191,11 @@ def _split(op: str, ary: ArrayLike, indices_or_sections: Union[int, ArrayLike],
            + ((r + 1) * (part_size + 1) - 1)])
     else:
       raise ValueError("array split does not result in an equal division")
-  starts, ends = [0] * ndim(ary), shape(ary)
+  starts, sizes = [0] * ndim(ary), shape(ary)
   _subval = lambda x, i, v: subvals(x, [(i, v)])
-  return [lax.slice(ary, _subval(starts, axis, start), _subval(ends, axis, end))
+  # Use dynamic rather than static slice to prevent slow execution of large
+  # number of splits; see https://github.com/google/jax/issues/12999
+  return [lax.dynamic_slice(ary, _subval(starts, axis, start), _subval(sizes, axis, end - start))
           for start, end in zip(split_indices[:-1], split_indices[1:])]
 
 @_wraps(np.split, lax_description=_ARRAY_VIEW_DOC)


froystig (commented):
Do we want benchmarks here, like those you added in #12219?

Does this change affect performance for a _small_ number of splits?

jakevdp:
Good question - for repeated operations after the initial run, we get something like this:

main branch:
```python
In [2]: %timeit jax.block_until_ready(jnp.split(x, jnp.arange(0, 5000, 10)))                                          
40.5 ms ± 4.22 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

this branch:
```python
In [2]: %timeit jax.block_until_ready(jnp.split(x, jnp.arange(0, 5000, 10)))                                          
116 ms ± 4.01 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
So a slowdown of ~3x on repeated calls is the tradeoff for saving ~150x on the initial call.

What do you think?

jax/_src/numpy/lax_numpy.py
```diff
@@ -1191,9 +1191,11 @@ def _split(op: str, ary: ArrayLike, indices_or_sections: Union[int, ArrayLike],
            + ((r + 1) * (part_size + 1) - 1)])
     else:
       raise ValueError("array split does not result in an equal division")
-  starts, ends = [0] * ndim(ary), shape(ary)
+  starts, sizes = [0] * ndim(ary), shape(ary)
   _subval = lambda x, i, v: subvals(x, [(i, v)])
-  return [lax.slice(ary, _subval(starts, axis, start), _subval(ends, axis, end))
+  # Use dynamic rather than static slice to prevent slow execution of large
+  # number of splits; see https://github.com/google/jax/issues/12999
+  return [lax.dynamic_slice(ary, _subval(starts, axis, start), _subval(sizes, axis, end - start))
```
jakevdp, line 1198: Another option here: we could replace this with
```python
return [ary[int(start): int(end)] for start, end in zip(split_indices[:-1], split_indices[1:])]
```
Due to #12219, this would now use dynamic rather than static slices and result in the same performance characteristics.

froystig:
From conversation with @hawkinsp: we may want to profile in order to understand why the dynamic slice approach is 3x slower, since it isn't clear that it ought to be (?)

KeAWang:
Why is `dynamic_slice` faster than static `slice` in the first place?

KeAWang:
Thanks! That explains the performance issues I've been trying to debug with slice vs dynamic_slice vs numpy slice.

jakevdp:
I can no longer reproduce these performance issues, despite this still lowering to static slice.

PR #13115, opened 2022-11-04
add `-W default` to pytest

yejingxin:
DETAILS:

without the flag '-W default', tests/compilation_cache_test.py appears to be failed due to warning messages.


jakevdp:
Hi - we typically do treat warnings as errors and use explicit ignore statements in the tests when we want the warnings ignored (e.g. https://cs.github.com/google/jax?q=jtu.ignore_warning)

How would you feel about using that approach here as well?

yejingxin:
ah, I see, also talked to @skye , it seems we are planing to move away from pytest in favor of bazel, then I will close this PR

PR #13122, opened 2022-11-04
update pytest.ini to print warning message for compilation_cache_test

yejingxin:
use default filter to avoid pytest treat the two messages in `compilation_cache_test` as errors, which causes the test fails.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -19,5 +19,8 @@ filterwarnings =
     # numpy uses distutils which is deprecated
     ignore:The distutils.* is deprecated.*:DeprecationWarning
     ignore:`sharded_jit` is deprecated. Please use `pjit` instead.*:DeprecationWarning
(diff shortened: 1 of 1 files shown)


pytest.ini
```diff
@@ -19,5 +19,8 @@ filterwarnings =
     # numpy uses distutils which is deprecated
     ignore:The distutils.* is deprecated.*:DeprecationWarning
     ignore:`sharded_jit` is deprecated. Please use `pjit` instead.*:DeprecationWarning
+    # print message for compilation_cache_test.py::CompilationCacheTest::test_cache_read/write_warning
```
skye, line 22: nit: capitalize first letter of comment (unless it's a variable name or something else that's always lowercase) (`Print`)

skye (approved):

pytest.ini
yejingxin: done

PR #13127, opened 2022-11-05
fix jax.random.logits shape argument

### LINKED ISSUE #13124: jax.random.categorical doesn't respect shape parameter with broadcasted logits
### Description

The docs for `jax.random.categorical` say that the output of `jax.random.categorical` is `shape` when `shape` is provided. However, when provided `logits` which have to be broadcasted with `shape`, it just ignores the shape parameter after checking that it's broadcast compatible. I believe this can be fixed by broadcasting logits appropriately.

```
import jax
out = jax.random.categorical(jax.random.PRNGKey(0), logits=jax.numpy.ones( (4, 1, 10)), axis=-1, shape=(4, 8))

# out.shape is (4, 1), should be (4, 8)

assert out.shape == (4, 8), f"out.shape is {out.shape}, should be (4, 8)"
```

0.3.23

CPU

### Additional system info

_No response_

### NVIDIA GPU info

_No response_

mattjj:
fixes #13124

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1327,10 +1327,12 @@ def categorical(key: KeyArray,
     shape = tuple(shape)
     _check_shape("categorical", shape, batch_shape)
 
-  sample_shape = shape[:len(shape)-len(batch_shape)]
+  shape_prefix = shape[:len(shape)-len(batch_shape)]
+  logits_shape = list(shape[len(shape) - len(batch_shape):])
+  logits_shape.insert(axis % len(logits_arr.shape), logits_arr.shape[axis])
   return jnp.argmax(
-      gumbel(key, sample_shape + logits_arr.shape, logits_arr.dtype) +
-      lax.expand_dims(logits_arr, tuple(range(len(sample_shape)))),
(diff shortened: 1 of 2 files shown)


jax/_src/random.py
```diff
@@ -1327,10 +1327,13 @@ def categorical(key: KeyArray,
     shape = tuple(shape)
     _check_shape("categorical", shape, batch_shape)
 
-  sample_shape = shape[:len(shape)-len(batch_shape)]
+  shape_prefix = shape[:len(shape)-len(batch_shape)]
+  logits_shape = list(shape[len(shape) - len(batch_shape):])
+  logits_shape.insert(axis % len(logits_arr.shape), logits_arr.shape[axis])
+  logits_arr= jnp.broadcast_to(logits_arr, logits_shape)
```
sharadmv: insert space before =

sharadmv (approved):

jax/_src/random.py
mattjj: Done! Also I deleted that line just now.

PR #13145, opened 2022-11-07
Add support for the hermitian option on jnp.linalg.pinv.

hawkinsp:
In addition:
* Improve the `pinv` implementation to avoid computing an unnecessary reduction: `svd` sorts its singular values so we don't need to use `amax()` to find the largest one.
* Avoid explicitly forming the identity matrix in the `pinv` JVP.

### DIFF
--- a/jax/_src/numpy/linalg.py
+++ b/jax/_src/numpy/linalg.py
@@ -41,6 +41,9 @@ def _H(x: ArrayLike) -> Array:
   return jnp.conjugate(jnp.swapaxes(x, -1, -2))
 
 
+def _symmetrize(x: Array) -> Array: return (x + _H(x)) / 2
+
+
 @_wraps(np.linalg.cholesky)
 @jit
 def cholesky(a: ArrayLike) -> Array:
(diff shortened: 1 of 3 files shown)


tests/linalg_test.py
```diff
@@ -804,21 +804,29 @@ def args_maker():
...
   )
   @jtu.skip_on_devices("rocm")  # will be fixed in ROCm-5.1
-  def testPinv(self, shape, dtype):
+  def testPinv(self, shape, hermitian, dtype):
     rng = jtu.rand_default(self.rng())
-    args_maker = lambda: [rng(shape, dtype)]
+    if hermitian:
+      def args_maker():
+        x = rng(shape, dtype, scale=1)
+        return [np.matmul(x, T(x.conj())) + np.eye(shape[-1], dtype=dtype)]
```
jakevdp, line 819: Since input is symmetrized, is it worth checking with non-hermitian input (e.g. to make sure the gradient is computing the right thing)?

hawkinsp: Done. (I actually had to add additional symmetrization in the JVP as well.)

hawkinsp:
Interestingly `mypy` now complains about `a.shape` because many `ArrayLike` things do not have a `shape`. However this is inside a `jit`, so mypy is wrong. I'm going to add a cast to `Array`.

jax/_src/numpy/linalg.py
```diff
@@ -406,27 +409,32 @@ def eigvalsh(a: ArrayLike, UPLO: Optional[str] = 'L') -> Array:
...
-@jit
-def pinv(a: ArrayLike, rcond: Optional[ArrayLike] = None) -> Array:
+@partial(jit, static_argnames=('hermitian',))
+def pinv(a: ArrayLike, rcond: Optional[ArrayLike] = None,
+         hermitian: bool = False) -> Array:
   # Uses same algorithm as
   # https://github.com/numpy/numpy/blob/v1.17.0/numpy/linalg/linalg.py#L1890-L1979
   _check_arraylike("jnp.linalg.pinv", a)
-  arr = jnp.conj(a)
+  arr = cast(Array, a)
```
jakevdp, line 424: Another option is `arr = asarray(a)` which avoids the casting (and will be correct even if someone removes the `jit`)

hawkinsp: Done.

jakevdp (approved):

PR #13147, opened 2022-11-07
Avoid forming identity matrix in SVD JVP.

hawkinsp:
Set the default matmul precision in the SVD JVP, and use @ to express matmuls. Also fix a flaky test failure in QR test on Mac ARM.

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -21,6 +21,7 @@ import warnings
 import numpy as np
 from typing_extensions import Literal
 
+import jax
 from jax._src.numpy import lax_numpy as jnp
 from jax._src.numpy.vectorize import vectorize
 from jax._src import ad_util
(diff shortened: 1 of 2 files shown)


jax/_src/lax/linalg.py
```diff
@@ -1580,16 +1582,16 @@ def _svd_jvp_rule(primals, tangents, *, full_matrices, compute_uv):
...
-  dV = jnp.matmul(V, F.astype(A.dtype) * (SdS + _H(SdS)))
+  dU = U @ (F.astype(A.dtype) * (dSS + _H(dSS)) + dUdV_diag)
+  dV = V @ (F.astype(A.dtype) * (SdS + _H(SdS)))
 
   m, n = A.shape[-2:]
   if m > n:
-    I = lax.expand_dims(jnp.eye(m, dtype=A.dtype), range(U.ndim - 2))
-    dU = dU + jnp.matmul(I - jnp.matmul(U, Ut), jnp.matmul(dA, V)) / s_dim.astype(A.dtype)
+    dAV = dA @ V
+    dU = dU + (dAV - U @ Ut @ dAV) / s_dim.astype(A.dtype)
```
jakevdp, line 1591: From a computational precision point of view, would there be any benefit to keeping the factored approach? e.g. what if the elements in `dAV` are large and `U @ UT` is very near to the identity: could we run into precision issues from subtracting large near-equal floats?

jakevdp (approved):

PR #13166, opened 2022-11-09
Checkify: fix nan_checks+PRNGKeys - a PRNGKey is never NaN!

LenaMartens:
Add a guard to the nan_error_rule to not call `jnp.isnan` on keys.

The primitive with a key in the output was `slice_p`, which I don't think can ever generate a NaN? I removed a few primitives from the NaN rules which I think only index or reshape, PTAL if removing those makes sense. In general, the nan_checked primitive set could use another review.

In fact, I think these kinds of "value polymorphic" primitives are the only ones we allow for keys right now (see #8381), so if we remove those we could remove the guard entirely.


jax/_src/checkify.py
```diff
@@ -796,13 +804,8 @@ def add_nan_check(prim):
 add_nan_check(lax.add_p)
 add_nan_check(lax.sub_p)
 add_nan_check(lax.convert_element_type_p)
-add_nan_check(lax.broadcast_in_dim_p)
```
sharadmv, line 799: Why do we need to remove these checks? Won't KeyArrays get caught by the `isnan` check above?

LenaMartens: agreed, I figured they don't produce nans anyway, so might be able to remove them (see PR description), but I'll leave it for a follow-up change which reviews the set (eg. are we missing any primitives).

Reverted the removals for now.

sharadmv: Oh I see. Yea these can be removed in a follow up PR.

sharadmv (approved):

PR #13176, opened 2022-11-09
Fix a test on V100.

nouiz:
Mismatched elements: 2 / 40 (5%)
Max absolute difference: 0.01554482
Max relative difference: 1.1495375e-07

Rtol very low, so it make sense to increase atol.

complex128 shouldn't have a lower tolerance then complex64. It should be the same or the opposite.

### DIFF
--- a/tests/sparse_test.py
+++ b/tests/sparse_test.py
@@ -1208,7 +1208,7 @@ class BCOOTest(jtu.JaxTestCase):
       # TODO(tianjianlu): In some cases, this fails python_should_be_executing.
       # self._CompileAndCheck(f_sparse, args_maker)
       self._CheckAgainstNumpy(f_dense, f_sparse, args_maker)
-      if dtype == np.complex128:
+      if dtype in [np.complex128, np.complex64]:
         atol = 1E-1
       else:
         atol = 1E-2


hawkinsp:
I'm concerned that the atol is massively too big in this test already. I think all of the values are smaller than 0.01, if you print them.

nouiz:
If it isn't quick to find a better solution, we must still fix the tests in some way. It isn't good that running normal jax tests fail for a long time. Here is some extra information that I found:
Here is the failing inputs:
```
[[[-3.0365543-2.4974918j   0.       +0.j          4.5742064+3.4473596j ]
  [ 0.       +0.j         -2.242873 +0.02567506j -1.7941172+1.7369398j ]]

 [[ 0.       +0.j          0.       +0.j          0.       +0.j        ]
  [-1.0199382+4.826304j    0.6967588+0.8373382j   0.       +0.j        ]]

 [[ 3.3453896+0.26961213j  2.8485265-1.9762787j  -1.8664883-0.21833457j]
  [ 4.235288 +3.5950468j  -0.8604307+2.9088175j   0.       +0.j        ]]

 [[ 0.       +0.j          0.       +0.j          0.       +0.j        ]
  [-5.130865 -2.6071038j   0.       +0.j          0.       +0.j        ]]]
[[ 1.6087320e-03+7.9499732e-05j -4.2471936e-04+1.4205376e-03j
   1.2164910e-04+7.1693110e-05j  2.0359585e-03+1.5450002e-03j
   6.9893990e-04+1.1156716e-03j]
 [-4.5874531e-04+1.4997370e-05j  6.2917796e-04-9.1430958e-04j
  -1.6131492e-04-3.5611034e-04j  1.3259083e-03+9.6132531e-04j
  -8.9901022e-04+2.2830390e-03j]
 [-6.3474494e-04+8.8330248e-04j -8.0882142e-05+1.1572879e-03j
   2.2206537e-03+1.0279124e-03j -1.3381643e-03+5.6168513e-04j
  -1.2984726e-03+3.7698406e-05j]]
```
The output diff is:
```
[[[-0.01063497-0.00240699j  0.00047791+0.00176204j
    0.0064238 +0.01183576j -0.01038104-0.01182014j
   -0.00540543-0.00943726j]
  [ 0.00063309-0.00273268j -0.00325272-0.00014996j
   -0.00539858+0.00280751j -0.00157332-0.00545413j
    0.00422188-0.00746665j]]

 [[ 0.        +0.j          0.        +0.j
    0.        +0.j          0.        +0.j
    0.        +0.j        ]
  [-0.00235669+0.00730947j -0.00521879-0.0036089j
   -0.0002843 +0.00013079j -0.00941431+0.01003039j
   -0.00863552+0.00307333j]]

 [[ 0.00546089+0.00013893j -0.0014149 -0.00135253j
   -0.00469604-0.00282638j  0.01469156+0.00507933j
    0.00642031+0.01241393j]
  [ 0.00687873+0.00477286j -0.00478751+0.00710636j
    0.00143214+0.00057815j -0.00086865+**0.01689256j**
   -0.00691809+0.00265846j]]

 [[ 0.        +0.j          0.        +0.j
    0.        +0.j          0.        +0.j
    0.        +0.j        ]
  [-0.00804692-0.00460203j  0.00588267-0.0061813j
   -0.00043725-0.000685j   -0.00641825-**0.01323514j**
   -0.00067749-0.00754657j]]]
```

The issue is only in the imaginary part. I do not know well enough complex number to say if this is normal or not. But as complex128 already had this atol, using it for complex64 shouldn't be an issue.
In fact, atol for complex64 should be bigger or the same as complex128, so the reverse of what is currently done.

hawkinsp:
I commented out this test in https://github.com/google/jax/pull/13181 as part of fixing a number of other tests so we can take our time.

My concern is that `atol=1e-1` is enough that it's larger than pretty much all of the values in the matrix. Which means the comparison is checking very little.

nouiz:
> My concern is that `atol=1e-1` is enough that it's larger than pretty much all of the values in the matrix. Which means the comparison is checking very little.

Agreed. It isn't great. Commenting the test remove all the tests for the other dtypes that were fine. This is why I preffered to not comment int. We can still work on it with a high atol instead of a commented test. But if you think a comment test is better, go with that.

the high atol doesn't botter me as rtol is low. The failing max rtol was 1.1495375e-07.

hawkinsp:
Since the rtol limit configured was 1e-6, why didn't the test pass? Maybe something is up with the comparisons for complex numbers?

nouiz:
An approximation of what is done is if any of the atol or rtol fail, the comparison fail.
Here is the exact comparison fct:
https://github.com/numpy/numpy/blob/v1.23.0/numpy/core/numeric.py#L2318

hawkinsp:
Closing this PR since the test now "passes", but I'm still quite concerned about how imprecise it is...

PR #13184, opened 2022-11-10
Conditional scan JEP

patrick-kidger:
As discussed internally: adds a JEP for bounded while loops, as implemented in #13062.

WDYT?

Tagging:
- @mattjj and @sharadmv as they're reviewers for #13062.
- @pschuh, @yashk2810, @hawkinsp, @jekbradbury as they were involved in the internal discussion, just in case they're curious.
- @shoyer, @nicholasjng, @soraros, @aespielberg, @njwfish, @shailesh1729, @dlwh, @joglekara, @gnecula, @tavin, @dcmvdbekerom, @salayatana66, as those who've engaged in discussions on this topic in the past. (And who presumably have a stated interest in this coming to pass.) I'd particularly invite you all to contribute to the discussion here, to be sure that this adequately captures your use-cases.

### DIFF
--- a/docs/jep/13062-bounded-while-loop.md
+++ b/docs/jep/13062-bounded-while-loop.md
@@ -1,0 +1,121 @@
+# Bounded while loops
+
+- *Author: patrick-kidger (GitHub) / kidger (Google)*
+
+- *Date: November 2022*
+
+## Summary
+
+We propose to introduce "bounded while loops" into JAX: these are while loops which exit unconditionally after some maximum number of iterations (are "bounded"), and are reverse-mode autodifferentiable.
+
+An implementation is available at https://github.com/google/jax/pull/13062.
+
+## Background
+
+### JAX limitations
+
+In order to reverse-mode autodifferentiate an operation, then residuals must be stored between the forward and backward pass. In the case of a while loop, which may iterate for an arbitrary number of steps, then these residuals may be of potentially unbounded size. However, it is a limitation of JAX (inherited from XLA) that the shape of all arrays be known statically in advance. For this reason, `lax.while_loop` is not reverse-mode autodifferentiable (or to be precise, not transposable).
+
+If a bound were placed on the maximum number of steps (after which the while loop exits unconditionally), then this would change: residuals may be stored in arrays of length `max_steps`, which will be partially filled according to how many steps are actually taken.
+
+### Use cases
+
+Differential equation solvers are a major use-case for this feature. A differential equation solver will typically take some variable number of steps, as determined on-the-fly by an adaptive step size controller. Differentiating through the solver is desirable: this makes it possible to fit model parameters to data via gradient updates, as is also done in deep learning. (Indeed, modern techniques frequently hybridise the two.)
+
+Other use-cases include linear and nonlinear solvers, as many of these solvers are iterative in nature, e.g. linear conjugate gradient or (quasi-)Newton methods. Sparse numerical linear algebra, such as sparse truncated SVD, is another example.
+
+(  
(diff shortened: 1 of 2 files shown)


docs/jep/13062-bounded-while-loop.md
```diff
@@ -0,0 +1,121 @@
...
+
+If a bound were placed on the maximum number of steps (after which the while loop exits unconditionally), then this would change: residuals may be stored in arrays of length `max_steps`, which will be partially filled according to how many steps are actually taken.
+
+### Use cases
+
+Differential equation solvers are a major use-case for this feature. A differential equation solver will typical take some variable number of steps, as determined on-the-fly by an adaptive step size controller. Differentiating through the solver is desirable: this makes it possible to fit model parameters to data via gradient updates, as is also done in deep learning. (Indeed, modern techniques frequency hybridise the two.)
```
dan-zheng, line 23:
```suggestion
Differential equation solvers are a major use-case for this feature. A differential equation solver will typical take some variable number of steps, as determined on-the-fly by an adaptive step size controller. Differentiating through the solver is desirable: this makes it possible to fit model parameters to data via gradient updates, as is also done in deep learning. (Indeed, modern techniques frequently hybridise the two.)
```

canyon289, line 23:
```suggestion
Differential equation solvers are a major use-case for this feature. A differential equation solver will typically take some variable number of steps, as determined on-the-fly by an adaptive step size controller. Differentiating through the solver is desirable: this makes it possible to fit model parameters to data via gradient updates, as is also done in deep learning. (Indeed, modern techniques frequency hybridise the two.)
```

docs/jep/13062-bounded-while-loop.md
```diff
@@ -0,0 +1,121 @@
...
+
+We propose to introduce "bounded while loops" into JAX: these are while loops which exit unconditionally after some maximum number of iterations (are "bounded"), and are reverse-mode autodifferentiable.
+
+An implementation is available at https://github.com/google/jax/pull/13062.
+
+## Background
+
+### JAX limitations
+
+In order to reverse-mode autodifferentiate an operation, then residuals must be stored between the forward and backward pass. In the case of a while loop, which may iterate for an arbitrary number of steps, then these residuals may be of potentially unbounded size. However, it is a limitation of JAX (inherited from XLA) that the shape of all arrays be known statically in advance. For this reason, `lax.while_loop` is not reverse-mode autodifferentiable (or to be precise, not transposable).
```
joglekara, line 17: In order to reverse-mode autodifferentiate an operation, the residuals must be stored between the forward and backward pass. In the case of a while loop, which may iterate for an arbitrary number of steps, then these residuals may be of potentially unbounded size. However, it is a limitation of JAX (inherited from XLA) that the shape of all arrays be known statically in advance. For this reason, `lax.while_loop` is not reverse-mode autodifferentiable (or to be precise, not transposable).

shoyer:
I can speak to the need for this feature. Differentiating through while loops comes up all the time in scientific computing primitives. Saving compute makes a big difference, even if we need to allocate memory for maximum static size.

In fact, I wrote a naive version of scan with early exit this just this week -- which I'm now pretty sure would not work properly.

docs/jep/13062-bounded-while-loop.md
patrick-kidger: Ty, fixed.

patrick-kidger: Ty, fixed.

nouiz:
For what it is worth, from my memory, Theano was doing that to support the gradient of scan/while loops.
So this is a tested solution that worked in another framework.

PR #13189, opened 2022-11-10
Update jax.numpy.lcm function to match with np.lcm in case of overflow

### LINKED ISSUE #12728: jax.numpy.lcm int8 overflow difference with np
### Description

The following code in jax

```python
import jax.numpy as jnp
a = jnp.array([5], dtype=jnp.int8)
b = jnp.array([26], dtype=jnp.int8)
print(jnp.lcm(a, b))
```
produces `[126]` 

while in numpy (checked in 1.23.0 and 1.21.6) (and pytorch 1.11 and 1.12 ) it gives `[-126]` keeping true to the overflow for an `int8` dtype.

Is this difference of results between **`np`** and **`jnp`** desired here?

Surely, `jax.numpy` cannot follow NumPy API in every single output but this is something achievable in this case.
The cause is perhaps that jax uses `abs` at the end of lcm operation. I suppose it should suffice to apply `abs` on `x1` and `x2` as `gcd` does.

Note: Mathematically lcm(5,26)=130, so getting a negative value (as with numpy) can at least signal a warning but getting a positive (but incorrect) value close to the correct value can easily mislead.

int8 dtype was just picked for the sake of simpler example but I am guessing this won't be restricted to int8 dtype.

If you want then I can create a PR for this and in case I m totally mistakened please do enlighten me. Thank you very much.

jax 0.3.21, jaxlib v0.3.20. Also checked with jax master repo on GitHub.

CPU

### Additional system info

Python 3.8.10, Ubuntu 20, Google Colab

### NVIDIA GPU info

_No response_

Ishticode:
Close #12728 
The details are in the issue linked.
The change produces result matching with np and passing tests locally.
CLA signed.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4367,11 +4367,12 @@ def gcd(x1: ArrayLike, x2: ArrayLike) -> Array:
 def lcm(x1: ArrayLike, x2: ArrayLike) -> Array:
   _check_arraylike("lcm", x1, x2)
   x1, x2 = _promote_dtypes(x1, x2)
+  x1, x2 = abs(x1), abs(x2)
   if not issubdtype(_dtype(x1), integer):
     raise ValueError("Arguments to jax.numpy.lcm must be integers.")
   d = gcd(x1, x2)
   return where(d == 0, _lax_const(d, 0),
-               abs(multiply(x1, floor_divide(x2, d))))
+               multiply(x1, floor_divide(x2, d)))
 
 
 @_wraps(np.extract)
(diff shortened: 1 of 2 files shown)


jakevdp (changes requested):
Thanks for the contribution! Could you add a regression test for this behavior? i.e. a test that fails with the current main branch of jax, but succeeds with your change.

The existing test for lcm is here: https://github.com/google/jax/blob/c9ebf60f4ec8531f1b87a0f5063a5d6cc10305ad/tests/lax_numpy_operators_test.py#L284; you could add a more specific test for this further down in that file. Thanks!

Ishticode:
Hi. Thanks for the review. Added test for non-small intergers with dtype int8 to detect the difference between np and jnp in overflow. Locally fails with the original and passes with the change. 🙂

jakevdp (approved):
Looks good, thanks! I'm going to run this through some further tests (off github) and if everything looks good we'll merge

jakevdp:
Could you please squash your changes into a single commit? See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests. Thanks!

PR #13191, opened 2022-11-10
[sparse] improve testing framework

jakevdp:


tests/sparse_test.py
```diff
@@ -131,7 +133,78 @@ def _is_required_cuda_version_satisfied(cuda_version):
...
+    self.assertAllClose(expected, actual, check_dtypes=check_dtypes,
+                        atol=atol or tol, rtol=rtol or tol,
+                        canonicalize_dtypes=canonicalize_dtypes)
+
+  def _CompileAndCheckSparse(self, fun, args_maker, check_dtypes=True,
+                             rtol=None, atol=None, check_cache_misses=True):
+    args = args_maker()
+
+    def wrapped_fun(*args):
+      # TODO(jakevdp): this check fails for sparse inputs. Why?
```
froystig, line 168: Any guesses? Should we try and answer this as part of this PR?

jakevdp: Ah, I figured it out: it's because `args_maker` is creating buffers whose size differs depending on the random seed, so it's triggering a cache miss the second time the args are created.

jakevdp: I'll work on a fix...

jakevdp: Done, please take a look!

froystig (approved):
Looks good! Good catch with the re-tracing.

PR #13198, opened 2022-11-11
PRNGKeyArray is now a virtual subclass of ndarray

patrick-kidger:
Right now:
```python
import jax
jax.config.update("jax_enable_custom_prng", True)
x = jax.random.PRNGKey(0)
assert isinstance(x, jax.numpy.ndarray)  # :(
```
And I'm assuming this isn't intended.

### DIFF
--- a/jax/_src/prng.py
+++ b/jax/_src/prng.py
@@ -31,6 +31,7 @@ from jax.interpreters import batching
 from jax.interpreters import mlir
 from jax.interpreters import pxla
 from jax.interpreters import xla
+from jax._src import basearray
 from jax._src.sharding import (
     MeshPspecSharding, PmapSharding, OpShardingSharding)
 
@@ -258,6 +259,7 @@ lax_numpy._set_device_array_base_attributes(PRNGKeyArray, include=[
     '__getitem__', 'ravel', 'squeeze', 'swapaxes', 'take', 'reshape',
     'transpose', 'flatten', 'T'])
 lax_numpy._register_stackable(PRNGKeyArray)
+basearray.Array.register(PRNGKeyArray)
 
 
 # TODO(frostig): remove, rerouting callers directly to random_seed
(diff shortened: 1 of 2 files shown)


yashk2810:
Its an instance of jax.Array right?

After jax.Array is enabled, your above condition should pass?

jax/_src/prng.py
```diff
@@ -258,6 +258,7 @@ def flatten(self, *_, **__)   -> 'PRNGKeyArray': assert False
     '__getitem__', 'ravel', 'squeeze', 'swapaxes', 'take', 'reshape',
     'transpose', 'flatten', 'T'])
 lax_numpy._register_stackable(PRNGKeyArray)
+jnp.ndarray.register(PRNGKeyArray)
```
yashk2810, line 261: Please use `basearray.Array.register`

yashk2810:
cc @jakevdp who change ndarray.register to `basearray.Array.register`

jax/_src/prng.py
patrick-kidger: Done.

patrick-kidger:
Nope, it doesn't work for `jax.Array`. (Before this change.)

tests/random_test.py
```diff
@@ -476,6 +476,13 @@ def test_key_array_indexing_nd(self):
     self.assertRaisesRegex(IndexError, 'Too many indices.*',
                            lambda: keys[0, 1, None, 2])
 
+  def test_isinstance(self):
+    if not config.jax_enable_custom_prng:
+      self.skipTest("test requires config.jax_enable_custom_prng")
+    key = random.PRNGKey(0)
+    self.assertIsInstance(key, jnp.ndarray)
```
yashk2810, line 483: Does this pass when jax_array is enabled? (JAX_ARRAY=1)

Maybe this change this to jax.Array?

patrick-kidger: After this change yes. Before this change no.

jakevdp (approved):
Looks good to me

jakevdp:
Fix the flake8 issue (whitespace) and let's merge this!

patrick-kidger:
Done!

PR #13215, opened 2022-11-11
Sytrd solver and SytrdDescriptor should NOT be CUDA only

rahulbatra85:
With the addition of GPU implementation of symmetric (Hermitian) tridiagonal reduction in this commit https://github.com/google/jax/commit/352b042fe9371cca84fc4e1f47b10a4c51b3ae9c, JAX build for ROCm is broken.


hawkinsp (approved):
Thanks, I did my best to make ROCM work but it's hard to get it completely correct without a CI build.

PR #13235, opened 2022-11-14
Checkify: add and remove primitives which are checked for NaN outputs.

LenaMartens:
Started from all primitives exported from `jax.lax` and removed a primitive when:
- its output is int/bool (but what if the output is complex?)
- it does not generate NaNs, ie. if the input does not contain a NaN value, the output will not contain a NaN value (eg. reshape/concatenate/..., max/..)
- it's already handled by other rules (eg. div, gather/scatter and scan/cond/while)

Compared to the previous set:

added: `{logistic, custom_linear_solve, igammac, igamma_grad_a, psum, igamma, reduce, tan, rng_uniform, lgamma, digamma, regularized_incomplete_beta, reduce_prod, reduce_window, cbrt, bessel_i0e, random_gamma_grad, bessel_i1e}`

removed: `{shift_left, concatenate, complex, shift_right_arithmetic, convert_element_type, conj, sign, round, shift_right_logical, reduce_max, bitcast_convert_type, real, max, reduce_min, rev, slice, min, imag, clamp, floor, select_n}`


LenaMartens:
I think I should add back the complex primitives but not sure, wdyt?

sharadmv (approved):

LenaMartens:
I convinced myself that complex numbers cannot be NaN so removed them, but their parts can be NaN so now I'm not sure.  I guess they fall under "does not generate NaNs" though.

PR #13251, opened 2022-11-15
Add `jax.scipy.linalg.toeplitz`.

yotarok:
A counterpart for "scipy.linalg.toeplitz" is added in "jax.scipy.linalg".

We recently have a case that requires a Jax counterpart of `tf.linalg.LinearOperatorToeplitz`, and this functionality is in fact a part of SciPy, implemented as `scipy.linalg.toeplitz`.
This also relates to the request in https://github.com/google/jax/issues/10144.

### DIFF
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -27,6 +27,7 @@ from jax import lax
 from jax._src import dtypes
 from jax._src.lax import linalg as lax_linalg
 from jax._src.lax import qdwh
+from jax._src.numpy.lax_numpy import _check_arraylike
 from jax._src.numpy.util import _wraps, _promote_dtypes_inexact, _promote_dtypes_complex
 from jax._src.numpy import lax_numpy as jnp
 from jax._src.numpy import linalg as np_linalg
@@ -1031,3 +1032,28 @@ def hessenberg(a: ArrayLike, *, calc_q: bool = False, overwrite_a: bool = False,
     return h, q
   else:
     return h
+
+@_wraps(scipy.linalg.toeplitz)
+def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
+  if r is None:
+    _check_arraylike("toeplitz", c)
+    r = jnp.conjugate(jnp.asarray(c))
+  else:
+    _check_arraylike("toeplitz", c, r)
+
+  c = jnp.asarray(c).flatten()
+  r = jnp.asarray(r).flatten()
+
+  ncols, = c.shape
+  nrows, = r.shape
+
+  if ncols == 0 or nrows == 0:
+    return jnp.empty((ncols, nrows), dtype=jnp.promote_types(c.dtype, r.dtype))
+
+  nelems = ncols + nrows - 1
+  elems = jnp.concatenate((c[::-1], r[1:]))
+  patches = lax.conv_general_dilated_patches(
+      elems.reshape((1, nelems, 1)),
+      (nrows,), (1,), 'VALID', dimension_numbers=('NTC', 'IOT', 'NTC'),
+      precision=lax.Precision.HIGHEST)[0]
+  return jnp.flip(patches, axis=0)
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -42,6 +42,7 @@ T = lambda x: np.swapaxes(x, -1, -2)
 
 float_types = jtu.dtypes.floating
 complex_types = jtu.dtypes.complex
+int_types = jtu.dtypes.all_integer
 
 
 class NumpyLinalgTest(jtu.JaxTestCase):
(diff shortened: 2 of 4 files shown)


jax/_src/scipy/linalg.py
```diff
@@ -1032,3 +1032,20 @@ def hessenberg(a: ArrayLike, *, calc_q: bool = False, overwrite_a: bool = False,
     return h, q
   else:
     return h
+
+@_wraps(scipy.linalg.toeplitz)
+def toeplitz(c, r=None):
```
jakevdp, line 1037: Add type annotations:
```python
def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
  ...
```
Also, I think the type checker will require you to use different variable names for the flattened version of these below.

tests/linalg_test.py
```diff
@@ -1679,5 +1679,42 @@ def testSchurBatching(self, shape, dtype):
       Ts, Ss = vmap(lax.linalg.schur)(args)
       self.assertAllClose(reconstruct(Ss, Ts), args, atol=1e-4)
 
+  @jtu.sample_product(
+      cshape=[(4,), (8,), (3, 7)],
+      cdtype=float_types + complex_types,
+      rshape=[(3,), (7,), (2, 1, 4)],
```
jakevdp, line 1685: You might also test shapes `()` and `(0,)` to make sure you're covering all corner cases.

jakevdp (changes requested):
Looks great! Just a few comments below and we can get this in.

Also, please add the new function to the list here so it will show up in the HTML docs: https://github.com/google/jax/blob/main/docs/jax.scipy.rst#jaxscipylinalg

tests/linalg_test.py
```diff
@@ -1679,5 +1679,42 @@ def testSchurBatching(self, shape, dtype):
...
+      rdtype=float_types + complex_types)
+  def testToeplitzConstrcution(self, rshape, rdtype, cshape, cdtype):
+      rng = jtu.rand_default(self.rng())
+      args_maker = lambda: [rng(cshape, cdtype), rng(rshape, rdtype)]
+      promotion = jax.config.jax_numpy_dtype_promotion
+      if {rdtype, cdtype} == {np.float32, np.complex64}:
+        # Test that default complex-promotion behavior is matching with osp.
+        promotion = 'standard'
+      with jax.numpy_dtype_promotion(promotion):
```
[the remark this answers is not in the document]
yotarok: done

jax/_src/scipy/linalg.py
```diff
@@ -1032,3 +1032,20 @@ def hessenberg(a: ArrayLike, *, calc_q: bool = False, overwrite_a: bool = False,
...
+  if r is None:
```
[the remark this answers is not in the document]
yotarok: done
thanks, I didn't know this is an API standard: https://github.com/google/jax/issues/7737

jax/_src/scipy/linalg.py
```diff
@@ -1032,3 +1032,20 @@ def hessenberg(a: ArrayLike, *, calc_q: bool = False, overwrite_a: bool = False,
...
+    r = jnp.conjugate(jnp.asarray(c))
+
+  c = jnp.asarray(c).flatten()
+  r = jnp.asarray(r).flatten()
+
+  ncols, = c.shape
+  nrows, = r.shape
+  nelems = ncols + nrows - 1
```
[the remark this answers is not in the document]
yotarok: done. added a short-cut for such corner cases.

jax/_src/scipy/linalg.py
yotarok: done

tests/linalg_test.py
yotarok: done

jakevdp (approved):
Looks great, thanks!

yotarok:
Looks like the numerical error is severe when TPUs/ GPUs are used because `lax.conv_general_dilated_patches` is used.
So, I switched to use `vmap` and `np.roll` that are slightly slower but should be accurate.

jakevdp:
I suspect `vmap` of `roll` is going to be much less efficient than the original approach, due to how it's implemented. Would you like to change back to the convolution-based approach along with adjusting the matmul precision?

yotarok:
The profiling results with TPU were:
- lax.conv_general_dilated_patches: ~9ms
- roll + vmap: ~15ms
- lax.conv_general_dilated_patches + Precision.HIGHEST: ~12ms
(with configuaration of `r == c`, `r.shape = (1024, 256)` and `toeplitz` is called with `vmap`).

Confirmed that with `Precision.HIGHEST`, `assertAllClose` in the test passes with the default tolerance parameters with TPU/ GPU.

yotarok:
I noticed that integers are not tested.  So, I added integer tests and some skip conditions because CuDNN doesn't support integer convolution except for int8.
PTAL again.

tests/linalg_test.py
```diff
@@ -1679,5 +1680,62 @@ def testSchurBatching(self, shape, dtype):
       Ts, Ss = vmap(lax.linalg.schur)(args)
       self.assertAllClose(reconstruct(Ss, Ts), args, atol=1e-4)
 
+  @jtu.sample_product(
```
jakevdp, line 1683: These tests should probably go under `class ScipyLinalgTestCase` rather than `class LaxLinalgTestCase`

tests/linalg_test.py
```diff
@@ -1679,5 +1680,62 @@ def testSchurBatching(self, shape, dtype):
...
+      cshape=[(), (4,), (8,), (3, 7), (0, 5, 1)],
+      cdtype=float_types + complex_types,
+      rshape=[(), (3,), (7,), (2, 1, 4), (19, 0)],
+      rdtype=float_types + complex_types + int_types)
+  def testToeplitzConstrcution(self, rshape, rdtype, cshape, cdtype):
+      if ((rdtype in [np.float64, np.complex128]
```
jakevdp, line 1689: Here and throughout, please use two-space indentations.,

jakevdp (changes requested):
Last couple minor comments - thanks!

tests/linalg_test.py
yotarok: moved.

tests/linalg_test.py
yotarok: done.
What's the style recommendation for hanging indents?
For `_src/scipy/linalg.py`, I followed Google-style, i.e. 2 spaces for normal indents, and 4 spaces for hanging indents.
However, `tests/linalg_test.py` uses 2 spaces also for hanging indents, so I followed that file-local convention only for the test.

yotarok:
Thank you for the comments, I updated the PR accordingly.

jakevdp (approved):
Thanks!

PR #13253, opened 2022-11-15
Weaken register_keypaths annotation

NeilGirdhar:
This allows definitions like:
```python
def keypaths(_: Any) -> List[AttributeKeyPathEntry]:
    return [AttributeKeyPathEntry(name) for name in dynamic_fields]

register_keypath(keypaths)
```
`List[AttributeKeyPathEntry]` is not covariant with `List[KeyPathEntry]`, but it is with `Sequence[KeyPathEntry]`.

### DIFF
--- a/jax/_src/tree_util.py
+++ b/jax/_src/tree_util.py
@@ -18,7 +18,7 @@ import functools
 from functools import partial
 import operator as op
 from typing import (Any, Callable, Dict, Hashable, Iterable, List, NamedTuple,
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
It might also be nice to expose `KeyPathEntry`.

NeilGirdhar:
Also, another option would be to use `Iterable` rather than `Sequence`, which would allow keypath-functions to be generators.

jakevdp (approved):
Seems reasonable - thanks!

PR #13258, opened 2022-11-15
Fixed `ShapeDtypeStruct((), None)` giving float64 under 32-bit mode

patrick-kidger:
As title!

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -3093,7 +3093,7 @@ class ShapeDtypeStruct:
   __slots__ = ["shape", "dtype", "named_shape", "sharding"]
   def __init__(self, shape, dtype, named_shape=None, sharding=None):
     self.shape = shape
-    self.dtype = dtype if core.is_opaque_dtype(dtype) else np.dtype(dtype)
+    self.dtype = dtypes.canonicalize_dtype(dtype, allow_opaque_dtype=True)
     if sharding is not None:
       self.sharding = sharding
     self.named_shape = {} if named_shape is None else dict(named_shape)


jakevdp:
I wonder if `dtype=None` should raise an error here instead? It seems like an unintended use to me.

patrick-kidger:
FWIW I figured this was intended. I think it's convenient to be able to specify "please use the default precision", without having to manually check `jax.config.jax_enable_x64` manually oneself.

(E.g. we use it in Diffrax [here](https://github.com/patrick-kidger/diffrax/blob/ea1bdc9a590093898ca5f75342c6f1bcbced1039/diffrax/brownian/tree.py#L78), although if you wanted we could easily change that.)

jakevdp:
Raising an error seems safer to me.

froystig:
If we were to allow it, would it always be a `float` (of the default precision), and not an integer type?

jakevdp:
Numpy chooses float:
```python
>>> np.dtype(None)
dtype('float64')
```
but I still think for a lower-level API, requiring an explicit non-None dtype is better than assuming user intent.

jakevdp:
> How would we recommend one succinctly write what @patrick-kidger is looking for, namely "give me a dtype of the default precision"?

We don't have a well-defined API for that. As currently implemented, `jnp.result_type(float)` will return the default floating-point type though (i.e. respecting the values of the JAX_DEFAULT_DTYPE_BITS and JAX_ENABLE_X64 flags)

jakevdp:
Would it make more sense for `jnp.dtype` to include canonicalization, rather than just being an alias for `np.dtype`?

froystig:
Wondering the same, but not sure. Maybe a new function, with a new and descriptive name, would be clearer and avoid surprises. I could imagine a case for having it live in the top-level `jax` module—something like `jax.canonical_dtype`—so that it can also resolve (trivially) any opaque dtypes (not in numpy).

cc @mattjj, at least for the opaque dtypes bit

patrick-kidger:
Closing due to lack of interest.

jakevdp:
> How would we recommend one succinctly write what @patrick-kidger is looking for, namely "give me a dtype of the default precision"?

I realized we do have a decent answer for this:
```python
import jax
from jax import dtypes

with jax.experimental.enable_x64():
  print(dtypes.result_type(None))
  print(dtypes.result_type(float))
  print(dtypes.result_type(int))
```
```
float64
float64
int64
```
```python
with jax.experimental.disable_x64():
  print(dtypes.result_type(None))
  print(dtypes.result_type(float))
  print(dtypes.result_type(int))
```
```
float32
float32
int32
```
I still think that for `ShapeDtypeStruct`, it should probably raise an error for `dtype = None`. I'm going to give that a try and see if it breaks anyone.

PR #13270, opened 2022-11-16
Fix less-nice type annotations for e.g. `List[jax.Array]`

patrick-kidger:
```python
from typing import List
from jax import Array
print(List[Array])
# Before: List[jax._src.basearray.Array]
# After: List[jax.Array]
```

Reassigning `__module__` is a pretty standard thing to do for this kind of thing. (See e.g. [functools.wraps](https://github.com/python/cpython/blob/f4cb8285ba7c6c71fdb459fba53e877c1678e1ab/Lib/functools.py#L32), which reassigns `__module__` along with `__name__`, `__qualname__` etc.)

### DIFF
--- a/jax/_src/basearray.py
+++ b/jax/_src/basearray.py
@@ -51,3 +51,6 @@ class Array(abc.ABC):
   @property
   def at(self):
     raise NotImplementedError("property must be defined in subclasses")
+
+
+Array.__module__ = "jax"


jakevdp (approved):
Great idea - thanks! FWIW we use this strategy elsewhere in JAX as well (see #10013)

PR #13278, opened 2022-11-16
Include which jaxlib/libtpu version failed (latest or nightly) in TPU CI chat notification

skye:
Test run: https://github.com/google/jax/actions/runs/3482341477

### DIFF
--- a/.github/workflows/cloud-tpu-ci-nightly.yml
+++ b/.github/workflows/cloud-tpu-ci-nightly.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false # don't cancel all jobs on failure
       matrix:
         python-version: ["3.10"] # TODO(jakevdp): update to 3.11 when available.
-        jaxlib-version: [latest, nightly]
+        jaxlib-version: ["latest-release", "nightly"]
     steps:
       # https://opensource.google/documentation/reference/github/services#actions
       # mandates using a specific commit for non-Google actions. We use
@@ -30,7 +30,7 @@ jobs:
       - name: Install JAX
         run: |
           pip uninstall -y jax jaxlib libtpu-nightly
-          if [ "${{ matrix.jaxlib-version }}" == "latest" ]; then
+          if [ "${{ matrix.jaxlib-version }}" == "latest-release" ]; then
             pip install .[tpu] \
               -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cloud-tpu-ci-nightly.yml
```diff
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel all jobs on failure
       matrix:
-        python-version: ["3.10"]  # TODO(jakevdp): update to 3.11 when available.
-        jaxlib-version: [latest, nightly]
+        python-version: ["3.10"] # TODO(jakevdp): update to 3.11 when available.
```
yashk2810, line 19: Update to 3.11? It is available now.

.github/workflows/cloud-tpu-ci-nightly.yml
```diff
@@ -16,8 +16,8 @@ jobs:
...
+        jaxlib-version: ["latest-release", "nightly"]
```
yashk2810, line 20: Can we add more info about what latest-release actually means? 

Maybe:

`jax (head), jaxlib/libtpu (pypi)`

.github/workflows/cloud-tpu-ci-nightly.yml
skye: I'd like to leave that for a separate change, and we should probably do it more holistically throughout our CI. We should arguably also test 3.8 on TPU, since that's the default version that comes on the VM.

.github/workflows/cloud-tpu-ci-nightly.yml
skye: It seems a little weird to put an unused variable in the matrix... can we just assume that unless otherwise noticed, all jax CI is running jax at HEAD? It's a bit odd to run anything else :)

yashk2810: Okay, but what I said is True? Latest release does mean that?

yashk2810 (approved):

.github/workflows/cloud-tpu-ci-nightly.yml
skye: Ooh I missed the part about changing it to pypi. I actually think latest-release is more accurate. Pypi is just where we host releases, including non-latest releases.

skye: But yes, jaxlib-version==latest-release does still mean we run jax at head. Just jaxlib is from a release.

PR #13288, opened 2022-11-17
Added more accurate version of the `betaln` function.

### LINKED ISSUE #13267: `betaln` is wildly inaccurate for large values
### Description

`jax.scipy.special.betaln` becomes very inaccurate for large inputs.

Screen Shot 2022-11-15 at 4 54 10 PM

 The image is comparing jax, scipy, and a jax version I made that parallels scipy's version (at the expense of duplicated work, due to the evaluation of both branches).

The more accurate jax version can be found [here](https://gist.github.com/imh/b6e3e7e2ea3224e365360cea85578240). I would be happy to open up a PR for it, if someone could answer the following:

- If this went in `jax._src.scipy.special`, where should I import `jnp.where` from to avoid circular dependencies?
- Since I just ported (two branches of) of scipy's version of `betaln`, I expect there may be be license issues including it in jax. Is that the case? Scipy's license is linked in the ported version.

0.3.23

CPU

### Additional system info

MacOS 12.6

### NVIDIA GPU info

_No response_

imh:
Resolves #13267 

I'm not sure about it style-guide-wise, and I'm just pattern-matching on how to use `_wraps`, but this version is much more accurate than the current.

I was unable to get bazel working and haven't run the tests locally, but did compare accuracy against scipy and the current implementation (see the issue) on CPU.

### DIFF
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -107,6 +107,7 @@ jax.scipy.special
   :toctree: _autosummary
 
    betainc
+   betaln
    digamma
    entr
    erf
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -29,6 +29,7 @@ from jax._src.lax.lax import _const as _lax_const
 from jax._src.numpy.lax_numpy import moveaxis, _promote_args_inexact, _promote_dtypes_inexact
 from jax._src.numpy.util import _wraps
 from jax._src.ops import special as ops_special
+from jax._src.third_party.scipy.betaln import betaln as _betaln_impl
 from jax._src.typing import Array, ArrayLike
 
 
@@ -38,10 +39,11 @@ def gammaln(x: ArrayLike) -> Array:
   return lax.lgamma(x)
 
 
-@_wraps(osp_special.betaln, module='scipy.special')
-def betaln(x: ArrayLike, y: ArrayLike) -> Array:
-  x, y = _promote_args_inexact("betaln", x, y)
-  return lax.lgamma(x) + lax.lgamma(y) - lax.lgamma(x + y)
+betaln = _wraps(
+    osp_special.betaln,
+    module='scipy.special',
+    update_doc=False
+)(_betaln_impl)
 
 
 @_wraps(osp_special.betainc, module='scipy.special')
--- a/tests/lax_scipy_test.py
+++ b/tests/lax_scipy_test.py
@@ -303,6 +303,18 @@ class LaxBackedScipyTests(jtu.JaxTestCase):
     q = np.array([1., 40., 30.], dtype=np.float32)
     self.assertAllClose(np.array([1., 0., 0.], dtype=np.float32), lsp_special.zeta(x, q))
 
+  def testIssue13267(self):
+    """Tests betaln(x, 1) across wide range of x."""
+    xs = jnp.geomspace(1, 1e30, 1000)
+    primals_out, tangents_out = jax.jvp(lsp_special.betaln, primals=[xs, 1.0], tangents=[jnp.ones_like(xs), 0.0])
+    # Check that betaln(x, 1) = -log(x).
+    # Betaln is still not perfect for small values, hence the atol (but it's close)
+    atol = jtu.if_device_under_test("tpu", 1e-3, 1e-5)
+    self.assertAllClose(primals_out, -jnp.log(xs), atol=atol)
+    # Check that d/dx betaln(x, 1) = d/dx -log(x) = -1/x.
+    self.assertAllClose(tangents_out, -1 / xs, atol=atol)
+
+
   def testXlogyShouldReturnZero(self):
     self.assertAllClose(lsp_special.xlogy(0., 0.), 0., check_dtypes=False)
 
--- a/jax/_src/third_party/scipy/LICENSE.txt
+++ b/jax/_src/third_party/scipy/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2001-2002 Enthought, Inc.  2003-2019, SciPy Developers.
+Copyright (c) 2001-2002 Enthought, Inc. 2003-2022, SciPy Developers.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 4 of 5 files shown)


jax/_src/scipy/special.py
```diff
@@ -38,10 +39,7 @@ def gammaln(x: ArrayLike) -> Array:
   return lax.lgamma(x)
 
 
-@_wraps(osp_special.betaln, module='scipy.special')
-def betaln(x: ArrayLike, y: ArrayLike) -> Array:
-  x, y = _promote_args_inexact("betaln", x, y)
-  return lax.lgamma(x) + lax.lgamma(y) - lax.lgamma(x + y)
+_wraps(betaln, module='scipy.special', update_doc=False)
```
jakevdp, line 42: this should be something like
```python
betaln = _wraps(osp_special.betaln, module='scipy.special')(betaln)
```
As currently written, the output of `_wraps` is unassigned and so it will have no effect.

jakevdp (changes requested):
Looks great, thanks! Could you add some tests that cover the problematic cases?

jax/_src/scipy/special.py
imh: Great catch, thanks

imh:
Thanks for the feedback. I added a test case testing `betaln(x, 1)` and its derivative across a wide range of x, and I think I fixed the wrapper.

A few questions:

- `betaln` doesn't currently show up in [the API docs](https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.special). I've checked that this change keeps it that way when locally rebuilding the docs. Is that the intended case, or is it intended to show up?
- I have only tested on CPU. Does the automated test suite run the tests on GPU/TPU? If not, I can spin some up and test myself.

hawkinsp:
Incidentally:

> I was unable to get bazel working and haven't run the tests locally

You don't need bazel to run the tests. You can use either `pytest` or `bazel`. See: https://jax.readthedocs.io/en/latest/developer.html#using-pytest

(We use `bazel` to build the C++ parts of JAX, but if you're changing the Python parts only you don't need to build anything and can use a prebuilt `jaxlib` and run tests with `pytest`, etc.)

> betaln doesn't currently show up in [the API docs](https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.special).

It's probably missing from the doc index. To make it show up, add it to this doc: https://github.com/google/jax/blob/main/docs/jax.scipy.rst

> I have only tested on CPU. Does the automated test suite run the tests on GPU/TPU?

Yes. A GPU test runs as a Github presubmit, and we'll also run TPU tests before merging.

docs/jax.scipy.rst
```diff
@@ -106,6 +106,7 @@ jax.scipy.special
 .. autosummary::
   :toctree: _autosummary
 
+   betaln
```
jakevdp, line 109: Move down one line for alphabetization

jakevdp (changes requested):
Thanks! A couple minor suggestions. Additionally, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

docs/jax.scipy.rst
imh: WOW. It wasn't that early in the morning that I couldn't do my ABC's ... or so I thought. 🙏

jakevdp:
Also if you rebase on the main branch the readthedocs issue should resolve itself.

imh:
Thank you again. I made the requested changes, and I think we should be ready to go now.

jakevdp (approved):
Looks great, thanks!

tests/lax_scipy_test.py
```diff
@@ -303,6 +303,17 @@ def testIssue3758(self):
     q = np.array([1., 40., 30.], dtype=np.float32)
     self.assertAllClose(np.array([1., 0., 0.], dtype=np.float32), lsp_special.zeta(x, q))
 
+  def testIssue13267(self):
+    """Tests betaln(x, 1) across wide range of x."""
+    xs = jnp.geomspace(1, 1e30, 1000)
+    primals_out, tangents_out = jax.jvp(lsp_special.betaln, primals=[xs, 1.0], tangents=[jnp.ones_like(xs), 0.0])
+    # Check that betaln(x, 1) = -log(x).
+    # Betaln is still not perfect for small values, hence the atol (but it's close)
+    assert jnp.allclose(primals_out, -jnp.log(xs), atol=1e-6)
```
jakevdp, line 312: It looks like this is failing on TPU; I'd probably just skip this test on TPU.

imh:
Out of an abundance of caution, since it's failing on TPU, I spun up a TPU and did some checks.

As expected, the version currently in JAX is inaccurate on TPU and CPU:
Old_betaln_TPU
Old_betaln_CPU

The new version in this PR isn't as accurate on TPU as it is on CPU, but it's tolerable on both:
New_betaln_TPU
New_betaln_CPU

I'll come back to this tomorrow to disable the test on TPU.

imh:
FYI, after you approved it, I made these changes:
- Added a specific error tolerance for TPU, so the tests should be working on them now
- Changed `assert jnp.allclose(...)` to `self.assertAllClose(...)`

jakevdp (approved):
Looks good, thanks!

jakevdp:
It looks like there are still some failures on CPU and GPU - are you able to see them under the checks below?

imh:
I can see the invocation log for the GPU build, but not the CPU build ("Unable to retrieve invocation log."). I can't download the `build.log` artifact for either ("Error 403, Forbidden")

- It looks like the GPU test just needs some looser tolerances, which I've just pushed (1e-6 -> 1e-5)
- If you can see the log for the CPU build, what's the failure there?

jakevdp:
We'd expect much tighter tolerance for float64 right? How about using a pattern like this to use appropriate tolerance depending on the dtype? https://github.com/google/jax/blob/7495a9e3705d67c498723240cfed3f14c5916567/tests/lax_scipy_test.py#L206

imh:
The float64 error is dominated by the function approximation error rather than floating point error, so it only cuts the absolute error in half:
New_betaln_GPU_float64 (1)

What was the CPU error? The tests pass locally, so I'm sorry to be making you hit the button again and again.

jakevdp:
Thanks for looking into that! I couldn't see the CPU error logs either. Might have been a flake.

jakevdp (approved):

PR #13300, opened 2022-11-17
[sparse] avoid re-indexing for linear unary ops

### LINKED ISSUE #13296: [sparse] Difference between BCOO matrices removes zeros
### Description

The difference operation between two `BCOO` sparse matrices implicitly removes zeros. The behaviour is different from the one observed for the addition. Consider the example below.

```python
import jax.experimental.sparse as jsp
import jax.numpy as jnp

a = jsp.BCOO((jnp.array([1.0]), jnp.array([[0, 0]])), shape=(3, 3))
b_values = jnp.array([0.0, 1.0, 0.0])
b_indices = jnp.array([[0, 0], [1, 1], [2, 2]])
b = jsp.BCOO((b_values, b_indices), shape=(3, 3))

print("Difference\n", (a - b).indices)
print("Addition\n", (a + b).indices)
```

The output is 
```
Difference
 [[0 0]
 [1 1]
 [3 3]
 [3 3]]
Addition
 [[0 0]
 [0 0]
 [1 1]
 [2 2]]
```

As you can see, in the difference operation the indices associated to null values in `a` or `b` have disappeared and they have been replaced with the out-of-bounds index `[3,3]`.

The expected behaviour can be obtained using `a + (-1)*b`.

jax v0.3.25, jaxlib v.0.3.25

CPU/GPU

### Additional system info

Python 3.10.8, ArchLinux

### NVIDIA GPU info

NVIDIA-SMI 520.56.06    Driver Version: 520.56.06    CUDA Version: 11.8

jakevdp:
Also avoid dropping explicit zeros when sparsifying unary operations. Fixes #13296

### DIFF
--- a/jax/experimental/sparse/transform.py
+++ b/jax/experimental/sparse/transform.py
@@ -70,6 +70,13 @@ from jax.experimental.sparse import BCOO
 
 sparse_rules : Dict[core.Primitive, Callable] = {}
 
+_zero_preserving_linear_unary_primitives = [
+  lax.copy_p,
+  lax.imag_p,
+  lax.neg_p,
+  lax.real_p,
+]
+
 _zero_preserving_unary_primitives = [
   lax.abs_p,
   lax.asin_p,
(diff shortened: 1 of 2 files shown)


jax/experimental/sparse/transform.py
```diff
@@ -70,26 +70,31 @@
 
 sparse_rules : Dict[core.Primitive, Callable] = {}
 
+# Zero-preserving unary primitives that commute with addition
```
froystig, line 73: Is "linear unary primitives" a correct characterization? Simpler if so.

froystig (approved):

jax/experimental/sparse/transform.py
jakevdp: Good call - Done!

PR #13317, opened 2022-11-18
Make pytest-xdist work on TPU and update Cloud TPU CI.

skye:
This change also marks multiaccelerator test files in a way pytest can understand (if pytest is installed).

By running single-device tests on a single TPU chip, running the test suite goes from 1hr 45m to 35m (both timings are running slow tests).

Sample run: https://github.com/google/jax/actions/runs/3499689262

I tried using bazel at first, which already supported parallel execution across TPU cores, but somehow it still takes 2h 20m! I'm not sure why it's so slow. It appears that bazel creates many new test processes over time, vs. pytest reuses the number of processes initially specified, and starting and stopping the TPU runtime takes a few seconds so that may be adding up. It also appears that single-process bazel is slower than single-process pytest, which I haven't looked into yet.

### DIFF
--- a/tests/xmap_test.py
+++ b/tests/xmap_test.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextlib
 import functools
 import itertools as it
 import os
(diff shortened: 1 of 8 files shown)


tests/xmap_test.py
```diff
@@ -55,6 +55,18 @@
 from jax.config import config
 config.parse_flags_with_absl()
 
+try:
```
yashk2810, line 58: Can I suggest using `contextlib.suppress(ImportError)` to achieve this rather than try;except?

(same everywhere)

skye: Nice, didn't know about that! Updated (and also removed type disable comments, pretty sure I don't need those)

yashk2810 (approved):

Issue #13291, opened 2022-11-17
Add missing distribution cdf and ppf functions to scipy.stats


carlosgmartin:
Add the following functions to [jax.scipy.stats](https://jax.readthedocs.io/en/latest/jax.scipy.html) (SciPy counterparts [here](https://docs.scipy.org/doc/scipy/reference/stats.html)):

- [ ] bernoulli.cdf
- [ ] bernoulli.ppf
- [ ] beta.cdf
- [ ] beta.ppf
- [ ] betabinom.cdf
- [ ] betabinom.ppf
- [ ] cauchy.cdf
- [ ] cauchy.ppf
- [ ] chi2.cdf
- [ ] chi2.ppf
- [ ] expon.cdf
- [ ] expon.ppf
- [ ] gamma.cdf
- [ ] gamma.ppf
- [ ] gennorm.ppf
- [ ] geom.cdf
- [ ] geom.ppf
- [ ] laplace.ppf
- [ ] pareto.cdf
- [ ] pareto.ppf
- [ ] poisson.cdf
- [ ] poisson.ppf
- [ ] t.cdf
- [ ] t.ppf
- [ ] truncnorm.ppf
- [ ] uniform.cdf
- [ ] uniform.ppf

harryjulian:
I'll submit a PR for the bernoulli functions this week, they're nearly done.

MatrixRanger98:
Please! This is useful. Is there any work around right now?

IlayMenahem:
@MatrixRanger98, for many of them there are analytical expressions

Qazalbash:
Please mark the following done as they have been implemented!

- bernoulli.cdf
- bernoulli.ppf
- beta.cdf
- beta.ppf
- cauchy.cdf
- cauchy.ppf
- chi2.cdf
- gamma.cdf
- poisson.cdf
- uniform.cdf
- uniform.ppf

carlosgmartin:
@Qazalbash I don't see `beta.ppf` [in the docs](https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.stats.beta).

Qazalbash:
Wops, mistake! I have updated the list.

crusaderky:
`stats.t.cdf(x, df)` is equivalent to `special.stdtr(df, t)`.
The latter does not exist in `jax.scipy` today, but fully supports JAX in scipy 1.15.2 through dispatch.
[...]

PR #13333, opened 2022-11-21
Implementation of scipy.stats bernoulli.cdf & bernoulli.ppf

harryjulian:
As referenced in #13291.

Initial implementations of the missing bernoulli functions w/associated tests.

Using pytest locally, I can only get most of the cdf  & _none_ of the ppf tests to work; however, locally testing in a notebook shows pretty much identical outputs (beyond types, which I'm not checking for in the tests) so I'm a little confused and wouldn't mind a little guidance here!

### DIFF
--- a/jax/_src/scipy/stats/bernoulli.py
+++ b/jax/_src/scipy/stats/bernoulli.py
@@ -36,3 +36,26 @@ def logpmf(k: ArrayLike, p: ArrayLike, loc: ArrayLike = 0) -> Array:
 @_wraps(osp_stats.bernoulli.pmf, update_doc=False)
 def pmf(k: ArrayLike, p: ArrayLike, loc: ArrayLike = 0) -> Array:
   return jnp.exp(logpmf(k, p, loc))
+
+@_wraps(osp_stats.bernoulli.cdf, update_doc=False)
+def cdf(k: ArrayLike, p: ArrayLike) -> Array:
+  k, p = jnp._promote_args_inexact('bernoulli.cdf', k, p)
+  zero, one = _lax_const(k, 0), _lax_const(k, 1)
+  conds = [
+    jnp.isnan(k) | jnp.isnan(p) | (p < zero) | (p > one),
+    lax.lt(k, zero),
+    jnp.logical_and(lax.ge(k, zero), lax.lt(k, one)),
+    lax.ge(k, one)
+    ]
+  vals = [jnp.nan, zero, one - p, one]
+  return jnp.select(conds, vals)
+
+@_wraps(osp_stats.bernoulli.ppf, update_doc=False)
+def ppf(q: ArrayLike, p: ArrayLike) -> Array:
+  q, p = jnp._promote_args_inexact('bernoulli.ppf', q, p)
+  zero, one = _lax_const(q, 0), _lax_const(q, 1)
+  return jnp.where(
+    jnp.isnan(q) | jnp.isnan(p) | (p < zero) | (p > one) | (q < zero) | (q > one),
+    jnp.nan,
+    jnp.where(lax.le(q, one - p), zero, one)
+  )
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -155,6 +155,43 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
                               tol=1e-4)
       self._CompileAndCheck(lax_fun, args_maker)
 
+  @genNamedParametersNArgs(2)
+  def testBernoulliCdf(self, shapes, dtypes):
+    rng_int = jtu.rand_int(self.rng(), -100, 100)
+    rng_uniform = jtu.rand_uniform(self.rng())
+    scipy_fun = osp_stats.bernoulli.cdf
+    lax_fun = lsp_stats.bernoulli.cdf
+
+    def args_maker():
+      x = rng_int(shapes[0], dtypes[0])
+      p = rng_uniform(shapes[1], dtypes[1])
+      return [x, p]
+
+    with jtu.strict_promotion_if_dtypes_match(dtypes):
+      self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                            tol=5e-4)
+      self._CompileAndCheck(lax_fun, args_maker)
+
+  @genNamedParametersNArgs(2)
+  def testBernoulliPpf(self, shapes, dtypes):
+    rng = jtu.rand_default(self.rng())
+    scipy_fun = osp_stats.bernoulli.ppf
+    lax_fun = lsp_stats.bernoulli.ppf
+
+    if scipy_version < (1, 9, 2):
+      self.skipTest("Scipy 1.9.2 needed for fix https://github.com/scipy/scipy/pull/17166.")
+
+    def args_maker():
+      q, p = map(rng, shapes, dtypes)
+      q = expit(q)
+      p = expit(p)
+      return [q, p]
+
+    with jtu.strict_promotion_if_dtypes_match(dtypes):
+      self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
+                             tol=5e-4)
+      self._CompileAndCheck(lax_fun, args_maker)
+
   @genNamedParametersNArgs(3)
   def testGeomLogPmf(self, shapes, dtypes):
     rng = jtu.rand_default(self.rng())
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -159,6 +159,8 @@ jax.scipy.stats.bernoulli
 
    logpmf
    pmf
+   cdf
+   ppf
 
 jax.scipy.stats.beta
 ~~~~~~~~~~~~~~~~~~~~
--- a/jax/scipy/stats/bernoulli.py
+++ b/jax/scipy/stats/bernoulli.py
@@ -18,4 +18,6 @@
 from jax._src.scipy.stats.bernoulli import (
   logpmf as logpmf,
   pmf as pmf,
+  cdf as cdf,
+  ppf as ppf
 )


jax/_src/scipy/stats/bernoulli.py
```diff
@@ -36,3 +36,29 @@ def logpmf(k: ArrayLike, p: ArrayLike, loc: ArrayLike = 0) -> Array:
...
+def cdf(k: ArrayLike, p: ArrayLike) -> Array:
+  k, p = jnp._promote_args_inexact('bernoulli.cdf', k, p)
+  zero, one = _lax_const(k, 0), _lax_const(k, 1)
+  conds = [
+    jnp.isnan(k),
+    lax.lt(k, zero),
+    jnp.logical_and(lax.ge(k, zero), lax.lt(k, one)),
+    lax.ge(k, one)
+    ]
+  vals = [jnp.nan, 0, 1-p, 1]
```
jakevdp, line 50: We should ensure the output dtype matches the input dtype. you can do this by using `lax._const` for each of the entries in this list

jax/_src/scipy/stats/bernoulli.py
```diff
@@ -36,3 +36,29 @@ def logpmf(k: ArrayLike, p: ArrayLike, loc: ArrayLike = 0) -> Array:
...
+  )
+
+@_wraps(osp_stats.bernoulli.ppf, update_doc=False)
+def ppf(q: ArrayLike, p: ArrayLike) -> Array:
+  q, p = jnp._promote_args_inexact('bernoulli.ppf', q, p)
+  one = _lax_const(q, 1)
+  return jnp.where(
+    jnp.logical_or(jnp.isnan(q), jnp.isnan(p)),
+    jnp.nan,
+    jnp.where(lax.gt(q, (one-p)), 1, 0)
```
jakevdp, line 63: Same here, the outputs should be the same types as the inputs.

tests/scipy_stats_test.py
```diff
@@ -49,6 +49,39 @@ def genNamedParametersNArgs(n):
...
+    lax_fun = lsp_stats.bernoulli.cdf
+
+    def args_maker():
+      x, logit = map(rng, shapes, dtypes)
+      x_dtype, p_dtype = x.dtype, logit.dtype
+      x = np.floor(x).astype(x_dtype)
+      p = expit(logit).astype(p_dtype)
+      return [x, p]
+
+    self._CheckAgainstNumpy(scipy_fun, lax_fun, args_maker, check_dtypes=False,
```
jakevdp, line 65: We need to use the context `with jtu.strict_promotion_if_dtypes_match(dtypes):` here or this will fail in the X64 tests (see other examples in this file). Same below.

jakevdp (changes requested):
Thanks for the PR! Just a couple requested changes; also

- please add the new functions to the list in https://github.com/google/jax/blob/main/docs/jax.scipy.rst
- please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Thanks!

harryjulian:
I've made the recommended changes and re-factored a little.

Still having a couple of problems. Namely:

1) I can't get all the scipy tests to pass. I've compared performance quickly in a [notebook](https://colab.research.google.com/drive/1jvXOAIvPkuO6xoq7o8TWNkThopY2Fec4#scrollTo=FuM23xPmoVgk) and everything looks pretty much the same (except `scipy.bernoulli.ppf` returns `-1`  when `q` and `p` both equal `0` - I'm not sure if this is a bug in _Scipy_ or not).
2) I can't get the compilation tests to work due to a `TracerArrayConversionError` - any immediate intuition as to what part of the code in either function might throw this issue?

jax/_src/scipy/stats/bernoulli.py
```diff
@@ -42,23 +42,30 @@ def cdf(k: ArrayLike, p: ArrayLike) -> Array:
...
-  vals = [jnp.nan, zero, one_minus_p, one]
+  vals = [zero, one_minus_p, one]
   return jnp.where(
-    jnp.logical_and(lax.ge(p, zero), lax.le(p, one)),
-    jnp.select(conds, vals), jnp.nan
+    jnp.logical_or(
+      jnp.logical_or(jnp.isnan(k), jnp.isnan(p)),
+      jnp.logical_or(lax.lt(p, zero), lax.gt(p, one))),
+    jnp.nan, 
+    jnp.select(conds, vals)
```
jakevdp, line 55: Is there a reason you nest `select` within `where`? Why not put all branches in the `jnp.select` call?

jax/_src/scipy/stats/bernoulli.py
```diff
@@ -42,23 +42,30 @@ def cdf(k: ArrayLike, p: ArrayLike) -> Array:
...
-  return jnp.where(
-    jnp.logical_or(jnp.isnan(q), jnp.isnan(p)),
+  zero, one, one_minus_p = _lax_const(q, 0), _lax_const(q, 1), _lax_const(q, 1-p)
+  return jnp.where(jnp.logical_or(
+      jnp.logical_or(jnp.isnan(q), jnp.isnan(p)),
+      jnp.logical_or(
+        jnp.logical_or(lax.lt(p, zero), lax.gt(p, one)),
+        jnp.logical_or(lax.lt(q, zero), lax.gt(q, one))
+        ),
+    ),
```
jakevdp, line 68: Here and above, I think it would be much more readable to use operators like `<`, `>`, and `|`. So instead of
```python
jnp.logical_or(
      jnp.logical_or(jnp.isnan(q), jnp.isnan(p)),
      jnp.logical_or(
        jnp.logical_or(lax.lt(p, zero), lax.gt(p, one)),
        jnp.logical_or(lax.lt(q, zero), lax.gt(q, one))
        ),
    )
```
you could write
```python
jnp.isnan(q) | jnp.isnan(p) | (p < 0) | (p > 1) | (q < 0) | (q > 1)
```
This approach is slightly slower at trace time, but is much easier to read and will end up lowering to the same low-level XLA operations.

tests/scipy_stats_test.py
```diff
@@ -49,6 +49,41 @@ def genNamedParametersNArgs(n):
...
+  @genNamedParametersNArgs(2)
+  def testBernoulliCdf(self, shapes, dtypes):
+    rng = jtu.rand_default(self.rng())
+    scipy_fun = osp_stats.bernoulli.cdf
+    lax_fun = lsp_stats.bernoulli.cdf
+
+    def args_maker():
+      x, logit = map(rng, shapes, dtypes)
+      x_dtype, p_dtype = x.dtype, logit.dtype
+      x = np.floor(x).astype(x_dtype)
```
jakevdp, line 61: I would use `jtu.rand_int` for `x` here rathern than `rand_default` followed by `floor`.

tests/scipy_stats_test.py
```diff
@@ -49,6 +49,41 @@ def genNamedParametersNArgs(n):
...
+      p = expit(logit).astype(p_dtype)
```
jakevdp, line 62: Is there any reason you need `p` to follow a particular distribution related to `expit`? If not perhaps `jtu.rand_uniform` would allow you to generate the appropriate range of values more directly.

jakevdp (changes requested):
Looking good! A few more comments

jakevdp:
Also, the readthedocs failure is coming from an IPython dependency issue that has already been fixed on main. Please rebase against an updated main branch to see the fix (it also looks like you have some conflicts to resolve).

harryjulian:
Updated, still having the same issues re: `TracerArrayConversionError` and `ppf` having correct looking outputs in a notebook but failing all the tests.

jakevdp:
Not sure about the `Tracer` error. Can you rebase on main and resolve conflicts so the CI tests will run and I can see the traceback?

jax/_src/scipy/stats/bernoulli.py
```diff
@@ -36,3 +36,26 @@ def logpmf(k: ArrayLike, p: ArrayLike, loc: ArrayLike = 0) -> Array:
...
+    jnp.logical_and(lax.ge(k, zero), lax.lt(k, one)),
+    lax.ge(k, one)
+    ]
+  vals = [jnp.nan, zero, one_minus_p, one]
+  return jnp.select(conds, vals)
+
+@_wraps(osp_stats.bernoulli.ppf, update_doc=False)
+def ppf(q: ArrayLike, p: ArrayLike) -> Array:
+  q, p = jnp._promote_args_inexact('bernoulli.ppf', q, p)
+  zero, one, one_minus_p = _lax_const(q, 0), _lax_const(q, 1), _lax_const(q, 1-p)
```
jakevdp, line 56: The tracer conversion error is coming from `_lax_const(q, 1-p)`. Since `p` is a traced value, a constant cannot be derived from it.

harryjulian:
Need to squash the commits together again but I believe the cdf component is now done. Still seems like the Scipy function is acting odd in the tests, nearly all `nan` out even though the same inputs produce the same outputs between scipy and this implementation in a notebook. I'll keep trying to diagnose:)

jakevdp:
When I pull-in your changes and run the tests locally they pass. I suspect you have some sort of version skew with the scipy package installed in your various environments.

Anyway - all looks good now! Once you squash the changes into a single commit, we can get this merged. Thanks for working on this!

jakevdp:
It looks like you're having trouble squashing the commits - let me know if you'd like me to take care of it for you (and if you're OK with me overwriting your branch)

harryjulian:
Yeah I seem to have made a bit of a mess. I'll have another crack at it tomorrow (and let you know if I need you to take care of it...)

jakevdp:
Just a hint: it looks like you merged against an updated main branch and then rebased against an older snapshot. You should be able to do something like this
```
$ git fetch upstream  # assuming upstream points to https://github.com/google/jax/
$ git checkout bernoulli
$ git rebase upstream/main
$ git push origin +bernoulli
```

harryjulian:
Sorted!

jakevdp (approved):

jakevdp:
Hmm, it looks like this test is failing against older versions of scipy. I think the best approach might be to add a scipy version check, and skip the test if the scipy version is too old. The test fails with scipy 1.8.1, and passes with scipy 1.9.3. It's possible that https://github.com/scipy/scipy/pull/17166 is the relevant fix, in which case the scipy version would have to be 1.9.2 or later for the test to pass

jakevdp (approved):

PR #13358, opened 2022-11-22
fix Threefry `split`/`fold_in` symmetry test under key arrays mode

froystig:


hawkinsp (approved):
This only fixes two of the three failing tests I think?

PR #13412, opened 2022-11-27
Update README.md

HilkopterBob:
Biutify


hawkinsp:
I'm not sure that actually improves things...

HilkopterBob:
It will, im sure

HilkopterBob:
IMHO it lets the explenation look way better

PR #13425, opened 2022-11-28
WIP: Add 'brute' option for searchsorted

jakevdp:
I still want to run some benchmarks before merging this...


jakevdp:
Done in #14836

PR #13427, opened 2022-11-29
[jax2tf] An alternative support for shape polymorphism for native lowering

gnecula:
[jax2tf] An alternative support for shape polymorphism for native serialization.

jax2tf already supports many cases of shape polymorphism, e.g., those where the shapes of all intermediates can be expressed as polynomials in the dimension variables appearing in input shapes. We would like to achieve the same coverage while using StableHLO as the lowering format, rather than tf.Graph.

For native serialization we will support two forms of lowering:

  * one is using the growing support in JAX for dynamic shapes, of which shape polymorphism is a special case. This implementation is enabled with the `--jax_dynamic_shapes` flag. At the moment, the JAX dynamic shapes support is still incomplete and over 500 of the 1000 jax2tf shape polymorphism tests are failing.

  * a new one (added) here in which we form a `Jaxpr` using abstract values that express dimension sizes as dimension polynomials (as for the standard jax2tf) which we then lower to StableHLO. This implementation is enabled when `--jax_dynamic_shapes` is off. With this implementation only 80 jax2tf tests fail, but they seem to fail in StableHLO shape inference.

The key contribution here is to enable lowering a Jaxpr that contains dimension polynomials in some of the shapes. Many lowering rules already have some partial support for Jaxprs where the shapes contain
`Var`s. To the extent possible, we try to write lowering rules that should cover both cases of dynamic shapes: `Var` or polynomials in shapes.

The lowering convention is that at top level we collect the sorted list of dimension variable names occurring in the input shapes, and we store it in `ModuleContext.dim_vars`. All IR functions will take `N` additional prefix arguments of int32 type containing the values of the `N` dimension variables. When generating the body of the IR function the current values of the dimension variables will be a list of `ir.Value` stored in `LoweringRuleContext.dim_var_values`.

Note that the Jaxprs are not changed to have extra `Var`s for the dimension variable values. An alternative implementation could work by transforming the Jaxpr to replace dimension polynomials into Vars.

The key code pattern used in the lowering rules is::

        if not core.is_constant_shape(shape):   # Handles both Var, and polynomials
           shape = mlir.eval_dynamic_shape(ctx, shape)
           return mhlo.DynamicXxxOp(..., shape)
        else:
           return mhlo.XxxOp(..., shape)

with `mlir.eval_dynamic_shape` handling both cases of `Var` and polynomials::

        def eval_dynamic_shape(ctx, shape):
           if config.jax_dynamic_shapes:
              # Using Var
              return ... subst using ctx.axis_size_env ...
           else:
              # Using polynomials
              return ... subst using ctx.module_context.dim_vars and ctx.dim_var_values

In order to support the above some lowering functions need to take a LoweringContext parameter, e.g., `mlir.full_like_aval`.

I expect that the changes here will improve the --jax_dynamic_shapes coverage as well.

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -411,7 +411,8 @@ def _custom_jvp_call_mlir_translation(ctx, *args, call_jaxpr, jvp_jaxpr_thunk,
   args_ = map(mlir.wrap_singleton_ir_values, args)
   consts = mlir._ir_consts(call_jaxpr.consts)
   out, tokens = mlir.jaxpr_subcomp(ctx.module_context, call_jaxpr.jaxpr,
-                                   ctx.tokens_in, consts, *args_)
+                                   ctx.tokens_in, consts, *args_,
+                                   dim_var_values=ctx.dim_var_values)
   ctx.set_tokens_out(tokens)
   return out
 mlir.register_lowering(custom_jvp_call_p, _custom_jvp_call_mlir_translation)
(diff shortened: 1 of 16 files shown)


mattjj (approved):
Great idea. Thanks for the big lift in implementing it, and unblocking lots of work while also improving things for `--jax_dynamic_shapes` too.

Issue #13429, opened 2022-11-29
 'jax.lax' has no attribute 'empty' 


joanna-yoo:
### Description

When using the flag `jax_remat_opt_barrier=False` in jax 0.3.24 / jaxlib 0.3.24, function compilation fails with the following error.

```
  File "/root/.cache/pypoetry/virtualenvs/fax-dhnsYw6V-py3.8/lib/python3.8/site-packages/jax/_src/ad_checkpoint.py", line 609, in remat_lowering
    return jax.named_call(translation_rule, name="remat")(*args, jaxpr=jaxpr)
  File "/root/.pyenv/versions/3.8.10/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/root/.cache/pypoetry/virtualenvs/fax-dhnsYw6V-py3.8/lib/python3.8/site-packages/jax/_src/ad_checkpoint.py", line 659, in _remat_translation_using_cond
    return jax.lax.cond(unif < np.float32(2), remat_comp, dummy_comp, *args)
AttributeError: module 'jax.lax' has no attribute 'empty'
```

### What jax/jaxlib version are you using?

0.3.24

### Which accelerator(s) are you using?

TPU

### Additional system info

python = 0.3.10, linux

jakevdp:
Hi, thanks for the report. Can you provide a code snippet that reproduces this error? Via a code search, the only likely candidate for the problem is this line: https://github.com/google/jax/blob/c54bc90bf48267728ffd002888cca62a66b1acd1/jax/_src/ad_checkpoint.py#L614

It looks like this function is dead code on the current main branch after merging of https://github.com/google/jax/pull/13247, so if I'm correct then I think this issue has already been fixed. Can you try running your code on the main branch to confirm? Thanks!

jakevdp:
Dead code removed in #13430

PR #13432, opened 2022-11-29
Fix issue #13429

joasyoo:
Fixes the issue `AttributeError: module 'jax.lax' has no attribute 'empty'` error when using `jax_remat_opt_barrier=False`


jakevdp:
Since this code isn't actually used anywhere, I think it's better to remove it entirely (done in https://github.com/google/jax/pull/13430)

PR #13434, opened 2022-11-29
Fix sharding for tuple created in inspect_sharding_partition

jurahul:
This ensures that the empty tuple created here has a valid sharding.


hawkinsp:
A test case would be a good idea, I suspect.

sharadmv (approved):
Yep, a test case would be nice. Thanks for the fix!

jurahul:
Note that this will be exercised by `test_inspect_sharding_is_called_in_pjit` test in debugging_primitives_test.py. When we add sharding verification in HLO instruction's set_sharding() function, this particular subtest will fail without the fix and pass with.

PR #13443, opened 2022-11-29
Drop support for Python 3.7 in jax v0.4.0

jakevdp:
In accordance with the [Python and NumPy version support policy](https://jax.readthedocs.io/en/latest/deprecation.html)

Also see #13444

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -1,7 +1,7 @@
 name: CI
 
 # We test all supported Python versions as follows:
-# - 3.7  : Documentation build
+# - 3.8  : Documentation build
 # - 3.8  : Part of Matrix
 # - 3.9  : Part of Matrix with NumPy dispatch
 # - 3.10 : Part of Matrix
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 # We test all supported Python versions as follows:
-# - 3.7  : Documentation build
+# - 3.8  : Documentation build
```
yashk2810, line 4: Double 3.8?

yashk2810 (approved):

.github/workflows/ci-build.yaml
yashk2810: Sorry, I read it wrong

jakevdp: Yeah, this list also leaves out 3.11. We should probably rework the test matrix, but I was thinking we'd do that separately from dropping 3.7 to avoid unnecessarily blocking this. What do you think?

yashk2810: Sounds good!

PR #13494, opened 2022-12-02
lax.fori_loop: allow scalar limits

jakevdp:
Previously, writing something like this would lead to a type  mismatch error:
```python 
stop = np.int16(10)
lax.fori_loop(0, stop, body, init)
```
This relaxes the type match requirement when one of the limits is expressed as a Python scalar (i.e. a weakly-typed value when traced).

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -1666,9 +1666,14 @@ def fori_loop(lower, upper, body_fun, init_val):
   lower_dtype = dtypes.canonicalize_dtype(lax.dtype(lower))
   upper_dtype = dtypes.canonicalize_dtype(lax.dtype(upper))
   if lower_dtype != upper_dtype:
-    msg = ("lower and upper arguments to fori_loop must have equal types, "
-           "got {} and {}")
-    raise TypeError(msg.format(lower_dtype.name, upper_dtype.name))
+    # Check for scalar integer inputs in case of type mismatch
+    if dtypes.is_weakly_typed(lower) and np.issubdtype(lower_dtype, np.integer):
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow/loops.py
```diff
@@ -1666,9 +1666,14 @@ def fori_loop(lower, upper, body_fun, init_val):
   lower_dtype = dtypes.canonicalize_dtype(lax.dtype(lower))
   upper_dtype = dtypes.canonicalize_dtype(lax.dtype(upper))
   if lower_dtype != upper_dtype:
-    msg = ("lower and upper arguments to fori_loop must have equal types, "
-           "got {} and {}")
-    raise TypeError(msg.format(lower_dtype.name, upper_dtype.name))
+    # Check for scalar integer inputs in case of type mismatch
+    if dtypes.is_weakly_typed(lower) and np.issubdtype(lower_dtype, np.integer):
```
hawkinsp, line 1670: Wouldn't this be problematic in the case that both are weakly typed and upper is the one with lower precision?

jakevdp: Yes, but that's part of the compromise we've made in allowing python scalar integers to implicitly promote to any strong integer type.

jakevdp: Oh, nvm, I see what you're saying...

jakevdp: Man this is tough to do cleanly in a way that preserves staticness of inputs (required for choosing `scan` vs. `while` below). i.e. we can't use `lax.convert_element_type` because of omnistaging

PR #13499, opened 2022-12-02
docs: add another example to the ConcretizationTypeError docs

jakevdp:
Added this example based on user feedback at https://stackoverflow.com/q/74640431

Preview of rendered docs here: https://jax--13499.org.readthedocs.build/en/13499/errors.html#jax-errors


skye (approved):
Nice!

PR #13539, opened 2022-12-06
Fix a false-positive typing warning on jax.default_device

wookayin:
Consider the following code where static type checkers can report an
error:

```python
CPU = jax.devices('cpu')[0]
with jax.default_device(CPU):
  ...                 # ^^^
```

```
Pyright: Argument of type "Device" cannot be assigned to parameter "new_val" of type "NoDefault"
  "Device" is incompatible with "NoDefault" (reportGeneralTypeIssues)
```

This is because `_StateContextManager.__call__` does not have a proper
type annotation on the parameter, unlike the attribute `_default_value`
which has a type annotation. Adding a `Any` to the parameter would
make the error disappear.

### DIFF
--- a/jax/_src/config.py
+++ b/jax/_src/config.py
@@ -487,7 +487,7 @@ class _StateContextManager:
     self._default_value = default_value
 
   @contextlib.contextmanager
-  def __call__(self, new_val=no_default):
+  def __call__(self, new_val: Any = no_default):
     if new_val is no_default:
       if self._default_value is not no_default:
         new_val = self._default_value  # default_value provided to constructor


jakevdp (approved):
Thanks! Looks good. We have CI to check annotations with mypy and pytype, but not pyright. Apparently pyright works differently than the other two in this case!

wookayin:
Thanks! pyright (pylance) is one of the most popular language server and static type checker many people are using in the OSS community (using VS Code or via LSP plugins), so it would be great to consider pyright as a supported type checker. It may have (and currently does) lots of many other false-positive errors though...

jakevdp:
We considered pyright in the process of [JEP 12049](https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html) but adding additional type checkers (with their own bespoke corner cases and false positives) seemed like it would over-burden development. We can revisit that again sometime, but I suspect we won't be quick to add pyright on top of the other two (which, by the way, are also open source and widely used in the OSS community)

PR #13545, opened 2022-12-07
checkify: initial style

LenaMartens:
Why?
- because initial style transforms are easier to write and maintain
- because we don't have to rely on data dependence anymore, and can delete all of the data-dependence work-arounds
- ...?

Any down-sides?
- we now stage out always, but assume `checkify` (functionalizing error values) is mainly useful in a staged-out environment


jax/_src/checkify.py
```diff
@@ -1264,9 +1062,177 @@ def checkify(fun: Callable[..., Out],
   """
   @traceback_util.api_boundary
   def checked_fun(*args, **kwargs):
-    args_flat, in_tree = tree_flatten((args, kwargs))
-    f, out_tree = flatten_fun(lu.wrap_init(fun), in_tree)
-    error, out_flat = checkify_flat(f, errors, *args_flat)
-    out = tree_unflatten(out_tree(), out_flat)
-    return error, out
+    # stage:
+    fun = lu.wrap_init(f, kwargs)
```
sharadmv, line 1066: I think we want to use `flatten_fun` here, not `flatten_fun_nokwargs` so we trace the kwargs (unless they opt out by closing over or using partial). We may want to eventually support static_argnums/names. This is consistent with other transformations.

sharadmv (approved):
This is awesome work! I'm excited we have finally addressed the data-dependence problems.

LenaMartens:
Thanks for the review! I've squashed the changes and will test internally.

PR #13556, opened 2022-12-08
Pattern-match select(inplace) into inplace(select).

### LINKED ISSUE #13522: `vmap`-`while`-`inplace` makes copies instead of updating in-place.
As title - a silent performance bug.

MWE:
```python
import jax
import jax.lax as lax
import jax.numpy as jnp
import timeit

def f(init_step, init_xs):
    def cond(carry):
        step, xs = carry
        return step < xs.size

    def body(carry):
        step, xs = carry
        xs = xs.at[step].set(1)
        return step + 1, xs

    return lax.while_loop(cond, body, (init_step, init_xs))

f = jax.jit(f)
vf = jax.jit(jax.vmap(f))

def run(size):
    args = 0, jnp.zeros(size)
    vargs = jnp.array([0]), jnp.zeros((1, size))
    time = min(timeit.repeat(lambda: f(*args), number=1, repeat=2))
    vtime = min(timeit.repeat(lambda: vf(*vargs), number=1, repeat=2))
    print(f"size={size} time={time} vtime={vtime}")

run(10000)
run(20000)
run(30000)

# size=10000 time=6.700679659843445e-05 vtime=0.12796505074948072
# size=20000 time=9.748293086886406e-05 vtime=0.5133151961490512
# size=30000 time=0.00012761494144797325 vtime=1.1480253851041198
```
(Tested on CPU.)

I think XLA is handling batch-`while` by computing the batch'd body, and then `select`ing just those batch elements that had changes. But it has a missing pattern-matching optimisation: it's not smart enough to recognise that `select(pred, xs, xs.at[step].set(new_x))` can be rewritten as `xs.at[step].set(select(pred, xs[step], new_x))` and thus that copying `xs` can be avoided.

I'd be happy to contribute this as a fix against XLA if someone can offer some guidance on where to find the XLA code / where its list of pattern-matching optimisations can be found.

Alternatively, I've been able to fix this issue in JAX, by implementing the above pattern-match optimisation at the jaxpr level, and I'd be happy to contribute my fix as a PR. My fix isn't perfect, as without being part of a fully-fledged optimising compile it still gets defeated by stuff that first requires other pattern matches (e.g. `select(pred, xs.T.T, xs.at[step].set(new_x))`), but it should at least offer a way to work around this issue.

patrick-kidger:
This fixes #13522. It adds a simple pattern-match to replace `select_n(scatter)` with `scatter(select_n)` and `select_n(dynamic_update_slice)` with `dynamic_update_slice(select_n)`.

By default this is applied to loop bodies as this seems to be the case when this issue arises in practice. (Namely in the batch rules for loops.) Perhaps we could/should apply this optimisation elsewhere as well, though.

This also tweaks the batching rule for `lax.while_loop` to explicitly use `lax.select`, so that this optimisation can be applied. In practice it looked like XLA is doing that under-the-hood with the previous implementation anyway.

The actual pattern-matcher itself is pretty simplistic, and will be defeated even by pretty straightforward cases: e.g. when one operation sits inside an `xla_call` and other one doesn't. Thus this PR is primarily to make it possible to work around the issue described in #13522 (right now it is impossible for an end-user to do so), even if it doesn't always give good performance in all cases by default.

---

This PR has one known failing test. I'm not entirely sure what's going on with this one, but I think it should be a simple fix for someone who better understands what it's testing.

### DIFF
--- a/jax/_src/lax/control_flow/loops.py
+++ b/jax/_src/lax/control_flow/loops.py
@@ -66,6 +66,7 @@ from jax._src.lax.control_flow.common import (
     _typecheck_param,
     allowed_effects,
     )
+from jax._src.lax.control_flow.pattern_match import pattern_match_inplace_select
 
 _map = safe_map
 zip = safe_zip
@@ -992,6 +993,7 @@ def _scan_pp_rule(eqn, context, settings):
 
 
 def scan_bind(*args, **params):
+  params['jaxpr'] = pattern_match_inplace_select(params['jaxpr'])
   if config.jax_enable_checks:
     avals = _map(core.get_aval, args)
     in_atoms = [core.Var(0, '', a) for a in avals]  # dummies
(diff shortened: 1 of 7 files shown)


froystig (commented):
@patrick-kidger and I discussed this offline: since this is essentially a compiler rewrite, we're thinking of having it live in XLA instead. Patrick has started work on that already.

patrick-kidger:
Closing in favour of
https://github.com/google/jax/commit/eb875cd5dd3a47cb8c99a324f37507a0ddfc622e
and
https://github.com/tensorflow/tensorflow/commit/02f2d2415252594d51f5ecf34f6fab61fceee9b5

PR #13562, opened 2022-12-08
Improves handling of opaque types for dynamic shapes

gnecula:
Improves handling of opaque types for dynamic shapes.

The immediate motivation for this is to support the lowering
to StableHLO for programs with polymorphic shapes. This requires
mixing of dynamic shapes with opaque types.

The general strategy is to push the actual selection of the MHLO ops
down into mlir module (e.g., mlir.slice_op, mlir.broadcast_in_dim)
so that we have one place where we pick whether we use the Dynamic
or static ops. These routines can also handle the opaque type.
This will result in a recursive
call to, e.g., mlir.slice_op, but the inner call will be using
the physical avals, which should not be opaque anymore.

While making this change I was confused by the fact that the
custom KeyTyRules in prng.py have lowerings that return multiple
MHLO ops. See https://github.com/google/jax/pull/11768#issuecomment-1342349102.
I changed the rules to return a single op.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -2893,14 +2893,11 @@ def _broadcast_in_dim_partial_eval(
   out_tracer.recipe = eqn
   return out_tracer
 
-def _broadcast_in_dim_lower(ctx, x, *dyn_shape, shape, broadcast_dimensions):
+def _broadcast_in_dim_lower(ctx, x, *dyn_shape, shape, broadcast_dimensions) -> Sequence[ir.Value]:
   aval_out, = ctx.avals_out
-  if core.is_opaque_dtype(aval_out.dtype):
-    return aval_out.dtype._rules.broadcast_in_dim_mlir(
-        ctx, x, *dyn_shape, shape=shape,
-        broadcast_dimensions=broadcast_dimensions)
   if dyn_shape:
     aval_out = aval_out.update(shape=_merge_dyn_shape(shape, dyn_shape))
+
 
   return [mlir.broadcast_in_dim(ctx, x, aval_out,
                                 broadcast_dimensions=broadcast_dimensions)]
@@ -3372,7 +3369,7 @@ def _transpose_batch_rule(batched_args, batch_dims, *, permutation):
 def _transpose_lower(ctx, x, *, permutation):
   aval_out, = ctx.avals_out
   if core.is_opaque_dtype(aval_out.dtype):
-    return aval_out.dtype._rules.transpose_mlir(ctx, x, permutation=permutation)
+    return [aval_out.dtype._rules.transpose_mlir(ctx, aval_out, x, permutation=permutation)]
   return mhlo.TransposeOp(x, mlir.dense_int_elements(permutation)).results
 
 transpose_p = standard_primitive(_transpose_shape_rule, _input_dtype,
@@ -4831,7 +4828,7 @@ empty_p = core.Primitive('empty')
 empty_p.def_abstract_eval(lambda *, dtype: core.ShapedArray((), dtype))
 def _empty_lower(ctx, *, dtype):
   if core.is_opaque_dtype(dtype):
-    return dtype._rules.empty_mlir(ctx)
+    return dtype._rules.empty_mlir(ctx, ctx.avals_out[0])
   return mlir.ir_constants(np.zeros((), np.dtype(dtype)))
 mlir.register_lowering(empty_p, _empty_lower)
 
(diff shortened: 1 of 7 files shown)


jax/_src/lax/lax.py
```diff
@@ -3372,7 +3369,7 @@ def _transpose_batch_rule(batched_args, batch_dims, *, permutation):
 def _transpose_lower(ctx, x, *, permutation):
   aval_out, = ctx.avals_out
   if core.is_opaque_dtype(aval_out.dtype):
-    return aval_out.dtype._rules.transpose_mlir(ctx, x, permutation=permutation)
+    return aval_out.dtype._rules.transpose_mlir(ctx, aval_out, x, permutation=permutation)
```
gnecula, line 3372: In general, we pass aval_out explicitly to the _rules methods because we should not assume that ctx.avals_out is the desired output aval for this particular op (it is the output aval for the currently running lowering rule).

jax/_src/lax/slicing.py
```diff
@@ -813,23 +813,8 @@ def _slice_batching_rule(batched_args, batch_dims, *, start_indices,
 def _slice_lower(ctx, x, *, start_indices, limit_indices, strides):
   strides = strides or [1] * len(start_indices)
   aval_out, = ctx.avals_out
-  if core.is_opaque_dtype(aval_out.dtype):
```
gnecula, line 816: Moved to mlir.slice_op all the handling of opaque types and dynamic shapes.

froystig:
I replied in https://github.com/google/jax/pull/11768#issuecomment-1343910108 about the multiple MHLO outputs when lowering.

jax/_src/lax/slicing.py
```diff
@@ -1411,10 +1381,10 @@ def _gather_lower(ctx, operand, indices, *,
                   indices_are_sorted, mode, fill_value):
   aval_out, = ctx.avals_out
   if core.is_opaque_dtype(aval_out.dtype):
-    return aval_out.dtype._rules.gather_mlir(
-        ctx, operand, indices, dimension_numbers=dimension_numbers,
+    return [aval_out.dtype._rules.gather_mlir(
+        ctx, ctx.avals_in, aval_out, operand, indices, dimension_numbers=dimension_numbers,
```
froystig, line 1385: Minor nitpick, but at least to explain our previous selves: I believe we were keeping a convention that the opaque dtype definition's lowering rule has the same signature as the primitive lowering rule (so `FooTy._rules.gather_mlir` didn't accept the in/out avals because `_gather_lower` here doesn't either). We don't have to stick to that. Any consistent convention is fine and I like this one as well, where the opaque dtype's rules consistently accept in/out avals as arguments.

froystig (approved):
In hindsight, a big part of this change makes sense as a simplification in light of the `physical_avals` method, which we only introduced later on, after having already required and implemented all the various individual lowering rules for slicing. That's nice.

(Technically, having complete control over the full lowering rules gives you more expressive power than having to commit to a uniform "physical" representation in the form of multiple avals. But I think we learned that this compromise is worthwhile for simple interaction with jax2tf and with jax.Array. This change now justifies it further.)

The change looks good overall to me. I'm approving with the assumption that we can revert back to having lowering rules that return multiple IR values, or that we are actually deciding not to do that anymore.

jax/_src/lax/slicing.py
gnecula: Understood. I think that this one is more flexible because you can then use, e.g., `broadcast_in_dim_mlir` in the lowering of some other primitive, e.g., `svd`.

gnecula:
I want to think some more about this. I think it is good to allow lowering to return multiple IR values, but we need to think what is the status of the custom lowering rules like `_rules.broadcast_in_dim_mlir`: are they meant to be used only in the lowering of broadcast_in_dim, or are they helper functions that generate code that is aware of opaque types and can be used in the lowering of other primitives? In the latter case, it is hard to deal with the situation when the helper function returns multiple IR values.

froystig:
Exercising the lowering rules of a test opaque dtype is the goal of [`CustomElementTypesTest`](https://github.com/google/jax/blob/f2c5d287a393a1df2211c74787eb6027c13400dd/tests/lax_test.py#L3011).

Do you mean to exercise specifically the lowering rules of the random key opaque dtype? I think we want similar test coverage there (in `random_test.py`). If we don't have it, we should add tests until we do. Key arrays ought to be sliceable. I can help do that. It's my mistake if current tests are insufficient.

gnecula:
I had in mind testing the actual lowering rules in KeyTyRule, but in the context of jax2tf shape polymorphism. I figured out how to do it, by setting jax_enable_custom_prng and writing a test that broadcasts/slices/... keys. See shape_poly_test:test_prng.

froystig:
Ah, I actually understand your previous question now that you've answered it! That's right, and the test in `shape_poly_test` looks good.

PR #13568, opened 2022-12-08
Make JAX tests pass under Numpy 1.24.0rc2.

hawkinsp:
* allow rc2 in numpy versions when parsed by tests.
* don't cast np.empty(), which can lead to cast errors.
* NumPy 1.24 now warns on overflowing scalar int to array casts in more places.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -1060,7 +1060,7 @@ def _get_monoid_reducer(monoid_op: Callable,
   return None
 
 def _get_bitwise_and_identity(dtype: DTypeLike) -> np.ndarray:
-  return np.array(-1, dtype)
+  return np.array(-1).astype(dtype)
 
 def _get_bitwise_or_identity(dtype: DTypeLike) -> np.ndarray:
   return np.array(0, dtype)
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -1127,3 +1127,13 @@ def strict_promotion_if_dtypes_match(dtypes):
   if all(dtype == dtypes[0] for dtype in dtypes):
     return jax.numpy_dtype_promotion('strict')
   return jax.numpy_dtype_promotion('standard')
+
+_version_regex = re.compile(r"([0-9]+(?:\.[0-9]+)*)(?:(rc|dev).*)?")
+def _parse_version(v: str) -> Tuple[int, ...]:
+  m = _version_regex.match(v)
+  if m is None:
+    raise ValueError(f"Unable to parse version '{v}'")
+  return tuple(int(x) for x in m.group(1).split('.'))
+
+def numpy_version():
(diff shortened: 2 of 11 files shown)


jax/_src/lax/lax.py
```diff
@@ -1060,7 +1060,7 @@ def _get_monoid_reducer(monoid_op: Callable,
   return None
 
 def _get_bitwise_and_identity(dtype: DTypeLike) -> np.ndarray:
-  return np.array(-1, dtype)
+  return np.array(-1).astype(dtype)
 
 def _get_bitwise_or_identity(dtype: DTypeLike) -> np.ndarray:
   return np.array(0, dtype)
```
jakevdp, line 1066: Does this need to be changed as well? Also below, the `np.array(val, dtype)` pattern is used several times.

jax/_src/numpy/lax_numpy.py
```diff
@@ -1790,6 +1792,11 @@ def concatenate(arrays: Union[np.ndarray, Array, Sequence[ArrayLike]],
   if dtype is None:
     arrays_out = _promote_dtypes(*arrays)
   else:
+    arrays_out = []
+    for arr in arrays:
+      if not can_cast(arr.dtype, dtype, casting=casting):
```
jakevdp, line 1797: I haven't put any thought into how numpy's casting rules interact with type promotion design, which differs between JAX and numpy. I'd feel more comfortable leaving this unimplemented until we take some time to think that through. For example, how do casting rules behave in the presence of weakly-typed values? If we just use `np.can_cast`, it will discard any information about weak types and just consider python scalars to be 64-bit, which may or may not be the behavior we want in JAX.

jakevdp: (if the only goal here is to make tests pass, I think you could add `'casting'` to the list of unsupported args in `lax_numpy_test.py`)

jax/_src/lax/lax.py
hawkinsp: No. It's only because it is now a DeprecationWarning to cast -1 to a uint64/uint32.

jax/_src/numpy/lax_numpy.py
hawkinsp: I added a `casting=unsafe` to the test that is failing. However, we probably need to think about this because now JAX's behavior doesn't match the default NumPy behavior: we allow more casts that NumPy would not.

jax/_src/lax/lax.py
jakevdp: I see, thanks.

jax/_src/numpy/lax_numpy.py
jakevdp: SGTM - I think we already do this in a few places

jakevdp (approved):

PR #13569, opened 2022-12-08
Small fix to scan type-check error message.

LenaMartens:


jakevdp (approved):
looks good!

jax/_src/lax/control_flow/loops.py
```diff
@@ -965,7 +965,7 @@ def _scan_typecheck(bind_time, *in_atoms, reverse, length, num_consts, num_carry
   if not all(_map(core.typecompat, x_avals_jaxpr, x_avals_mapped)):
     raise core.JaxprTypeError(
       f'scan jaxpr takes input sequence types\n{_avals_short(x_avals_jaxpr)},\n'
-      f'called with sequence of type\n{_avals_short(x_avals)}')
+      f'called with sequence of type\n{_avals_short(x_avals_mapped)}')
```
froystig, line 968:
```suggestion
      f'called with sequence whose items have type\n{_avals_short(x_avals_mapped)}')
```
Maybe something like this to resolve the ambiguity of whether it's the sequence's type or the type of its elements? Optional and up to you.

froystig (approved):

jax/_src/lax/control_flow/loops.py
LenaMartens: good suggestion, changed it to say this!

PR #13584, opened 2022-12-09
Checkify: Add documentation for adding run-time values to error message.

LenaMartens:


jax/_src/checkify.py
```diff
@@ -598,38 +598,46 @@ def checkify_fun_to_jaxpr(
...
   This is an effectful operation, and can't be staged (jitted/scanned/...).
   Before staging a function with checks, :func:`~checkify` it!
 
   Args:
     pred: if False, an error is added.
-    msg: error message if error is added.
+    msg: error message if error is added. Can be a format string.
+    *fmt_args, **fmt_kwargs: Posistional and keyword formatting arguments for
+      `msg`, eg.:
+      ``check(..., msg="check failed on values {} and {named}", x, named=y)``.
```
sharadmv, line 612: I don't think you can have positional arguments following named arguments.

sharadmv (approved):

PR #13587, opened 2022-12-09
DOC: add example of pure_callback with custom_jvp

jakevdp:
This is an example that is meant to come at the end of a more general discussion of `pure_callback`, `debug.callback`, and other callback mechanisms. This sort of question has come up frequently enough recently that I thought it would be useful to have a stub file defined into which we can fill-in details later.

Preview here: https://jax--13587.org.readthedocs.build/en/13587/notebooks/external_callbacks.html

### DIFF
--- a/docs/notebooks/external_callbacks.md
+++ b/docs/notebooks/external_callbacks.md
@@ -1,0 +1,176 @@
+---
+jupytext:
+  formats: ipynb,md:myst
+  text_representation:
+    extension: .md
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3
+  name: python3
+---
+
++++ {"id": "7XNMxdTwURqI"}
+
+# External Callbacks in JAX
+
++++ {"id": "h6lXo6bSUYGq"}
+
+This guide is a work-in-progress outlining the uses of various callback functions, which allow JAX code to execute certain commands on the host, even while running under `jit`, `vmap`, `grad`, or another transformation.
+
+This is a work-in-progress, and will be updated soon.
+
+*TODO(jakevdp, sharadmv): fill-in some simple examples of {func}`jax.pure_callback`, {func}`jax.debug.callback`, {func}`jax.debug.print`, and others.*
+
++++ {"id": "dF7hoWGQUneJ"}
+
+## Example: `pure_callback` with `custom_jvp`
+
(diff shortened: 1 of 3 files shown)


docs/notebooks/external_callbacks.md
```diff
@@ -0,0 +1,159 @@
...
+
+This is a work-in-progress, and will be updated soon.
+
+*TODO(jakevdp, sharadmv): fill-in some simple examples of {func}`jax.pure_callback`, {func}`jax.debug.callback`, {func}`jax.debug.print`, and others.*
+
++++ {"id": "dF7hoWGQUneJ"}
+
+## Example: `pure_callback` with `custom_jvp`
+
+One powerful way to take advantage of pure callbacks is to combine them with `custom_jvp` (see [Custom derivative rules](https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html) for more details on `custom_jvp`). Suppose we want to create a JAX-compatible wrapper for a scipy or numpy function that is not yet available in the `jax.scipy` or `jax.numpy` wrappers.
```
sharadmv, line 30: I think it might be worth linking the pure_callback docs here (https://jax.readthedocs.io/en/latest/_autosummary/jax.pure_callback.html)

docs/notebooks/external_callbacks.md
```diff
@@ -0,0 +1,159 @@
...
+:id: q3qh_4DrWxdQ
+:outputId: c46b0bfa-96f3-4629-b9af-a4d4f3ccb870
+:tags: [raises-exception]
+
+jax.grad(j1)(z)
+```
+
++++ {"id": "PtYeJ_xUW09v"}
+
+Let's define a custom gradient rule for this. Looking at the definition of the [Bessel Function of the First Kind](https://en.wikipedia.org/?title=Bessel_function_of_the_first_kind), we discover that there is a relatively straightforward recurrance relationship for the derivative with respect ot the argument `z`:
```
sharadmv, line 106: Nit: "ot" -> "to"

sharadmv, line 106: Nit: "recurrance" -> "recurrence"

sharadmv (approved):
Thanks Jake, this is great!

docs/notebooks/external_callbacks.md
jakevdp: Done

docs/notebooks/external_callbacks.md
jakevdp: Done

jakevdp: Done

Issue #8409, opened 2021-10-30
Jax cond evaluating too many branches.


aespielberg:
I have been playing around with converting [diffmpm from the difftaichi package](https://github.com/taichi-dev/difftaichi/blob/master/examples/diffmpm.py) into a jax version, and while the forward pass has been working wonderfully, the backward pass has been using way too much GPU memory.

Today, I was able to track down that memory usage to the grid op.  The grid op step is a  series of nested if statements.  At first, I was using jnp.where, which evaluates all branches.  That is extremely inefficient and can lead to OOM errors.  I simplified my code, and switch to jnp.cond, but my only conclusion is that cond is also evaluating both branches, otherwise I cannot see why this would run into OOM issues.

Below is a modified version of the grid op, that is composed into itself 4,000 times, like a simulation.  Even run with the XLA_PYTHON_CLIENT_PREALLOCATE=false flag, this quickly leads to the the whole GPU being used, and more if the loop length is increased.  This is _not_ true if every line from ```lin = ....``` until right before the return of grid_op is commented out.  In that case, memory usage is practically negligible.  Note that because bound = 0, literally every line written ```v_out = jax.lax.cond ... ``` evaluates to False by definition, and so most of the expressions, including the v_out_gate's and their dependencies, shouldn't even need to be evaluated in the jitted function.

Maybe I am misunderstanding cond; if so, what is the proper way to get this sparse branching behavior?  I don't want to evlauate and hang onto a bunch of expensive tensors that are never actually needed and crash my GPU with OOM, especially in an backward pass.  This is a core bottleneck to practical deployment of my code and a feature that I think should be supported.  FWIW, I am using Version: 0.1.69+cuda101

Code to reproduce is below.

```
import jax.numpy as jnp
from jax import grad, jit, vmap
from jax import random
import os
import math
import numpy as np
import matplotlib.pyplot as plt
import jax.nn as jnn
import jax.lax as jlax
import timeit
import jax

dim = 2
n_grid = 128
dt = 1e-3
gravity = 3.8


def allocate_arrays():
  global grid_m_in, grid_v_in, grid_v_out, loss, index_array
  grid_m_in = jnp.ones((n_grid, n_grid))
  grid_v_in = jnp.zeros((n_grid, n_grid, dim))
  grid_v_out = jnp.zeros((n_grid, n_grid, dim))


  index_array = np.zeros((n_grid, n_grid, dim))
  
  for i in range(n_grid):
    for j in range(n_grid):
[...]

Issue #7934, opened 2021-09-16
Efficiency of cond vs. select in lax package


ttt733:
[This comment](https://github.com/deepmind/dm-haiku/issues/216#issuecomment-920624135) on the Haiku repo introduced me to some JAX behavior that wasn't intuitive to me from the documentation - if you have "cheap" paths in a cond, it's better to use a select instead.

> AFAIK [using a select] is faster because a real cond (e.g. only execute one of the branches) on GPU requires a roundtrip of data from the GPU to the CPU in order to decide whether to take the LHS or RHS branch. While a select can be entirely computed on the GPU.

I'm not sure if this is necessarily a bug, but it was very surprising to me, and seems like something other people could easily run into without noticing. In my code, for example - which is a haiku-transformed forward function - I'm seeing a significant slowdown when using a cond. (For reference, `graph` is a pretty large array (shape=(10000, 100, 268)), in this example.)
```
# cond version: ~48 steps/sec
graph, max_index, edge_counts = jax.lax.cond(
    max_index > max_nodes,
    lambda _: (jnp.zeros(graph_shape), 0, jnp.zeros(edge_counts_shape, dtype='int32')),
    lambda _: (graph, max_index, edge_counts),
    operand=None,
)
```
```
# select version: ~40 steps/sec
graph = jax.lax.select(
    reset_graph,
    jnp.zeros(graph_shape),
    graph,
)
max_index = jax.lax.select(
    reset_graph,
    0,
    max_index,
)
edge_counts = jax.lax.select(
    reset_graph,
    jnp.zeros(edge_counts_shape, dtype='int32'),
    edge_counts,
)
```
I think that most links to the relevant parts of the documentation take people to the `cond` page - maybe it's worth leaving a note there? I'm not sure exactly how to word it since I don't understand the inner workings too well yet.

Issue #4453, opened 2020-10-05
lax.switch compilation time is super-linear in number of switch options


j-towns:
To demonstrate this, you can run the following (results below):
```python
from time import time

import matplotlib.pyplot as plt

from jax import jit, lax
import jax.numpy as np


def time_switch(n):
  # Create list of functions
  inners = []
  for i in range(n):
    inners.append(lambda x: x ** 2)
  switch = jit(lambda i, x: lax.switch(i, inners, x))
  t0 = time()
  switch(0, 3.)
  return time() - t0

def bench_switch(ns):
  return [time_switch(n) for n in ns]

ns = [256, 512, 768, 1024, 1280, 1536, 1792, 2048]
times = bench_switch(ns)
plt.plot(ns, times)
```
The bottleneck appears to be in XLA compilation; I profiled `time_switch` (with n=1500 on CPU) and almost all time is spent in [`backend_compile`](https://github.com/google/jax/blob/51538bd76d510487c137b9bffccb1e7f9ad58173/jax/interpreters/xla.py#L341-L344):
<img width="1030" alt="Screenshot 2020-10-05 at 14 09 31" src="https://user-images.githubusercontent.com/15261883/95083427-6a5aad80-0714-11eb-8f7f-7e5ebdf02097.png">.

### Benchmark results (y-axis in seconds)
CPU
![cpu](https://user-images.githubusercontent.com/15261883/95083705-c3c2dc80-0714-11eb-93a9-9008f2c5668a.png)

GPU
![gpu](https://user-images.githubusercontent.com/15261883/95083715-c9202700-0714-11eb-8dfc-43ee321fa526.png)

TPU
![tpu](https://user-images.githubusercontent.com/15261883/95084050-40ee5180-0715-11eb-942e-7bee2ac3676c.png)

j-towns:
This isn't a critical issue for me, and I think it can be worked around by using nested switches, which seem to compile faster, but I thought it was worth raising nevertheless.

hawkinsp:
To be honest, I'm not that surprised that compilation time is superlinear for this kind of program. Looking at a C++-level profile, the majority of time is spent inside LLVM. So I doubt there's an easy fix here.

j-towns:
Closing since this is apparently caused by LLVM so out of our control.

j-towns:
(and I'm not in any way blocking on it).

PR #3318, opened 2020-06-03
indexed conditionals with multiple branches

froystig:
This PR:

* Rewrites our `cond` primitive to a more general form. Instead of a boolean predicate that selects one of two branch functions, it takes an integer index into a list of branch functions.
* Introduces `lax.switch` alongside the existing `lax.cond`, as an interface to the generalized primitive.

Where the batching rule for binary `cond` previously generated a binary `select`, the batching rule for the new n-ary `cond` now generates an inline tree-of-binary-select-expressions. In a future change, I'd like to factor this implementation of "n-ary select" out of the batching rule. We can use it (or any other implementation) to generalize the `select` primitive similarly to how this PR generalizes `cond`.

Fixes #2725

PR #13589, opened 2022-12-09
DOC: clarify behavior of lax.cond & lax.select

jakevdp:
In response to questions at #13586; also #7934

Related to #8409

### DIFF
--- a/jax/_src/lax/control_flow/conditionals.py
+++ b/jax/_src/lax/control_flow/conditionals.py
@@ -155,8 +155,13 @@ def _cond(pred, true_fun: Callable, false_fun: Callable, *operands,
           operand=_no_operand_sentinel, linear=None):
   """Conditionally apply ``true_fun`` or ``false_fun``.
 
+  Wraps XLA's `Conditional
+  <https://www.tensorflow.org/xla/operation_semantics#conditional>`_
+  operator.
+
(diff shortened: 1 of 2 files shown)


skirsten:
I also just found #7934 which will probably be resolved by this.

jakevdp:
Thanks for the pointer!

jakevdp:
Updated based on offline discussion - please take a look!

bheijden:
I have a question related to

I've been noticing super linear compilation times in my application when using `jax.lax.switch` with an increasing number of branches (similar to [this](https://github.com/google/jax/issues/4453) issue). From what I read [here](https://github.com/google/jax/pull/3318), `jax.lax.cond` is implemented as a special case of `switch`.  

Could  [these lines](https://github.com/google/jax/pull/13589/files#diff-8964bb46126eb1a8e01c28fd7ca7058c589380c0df5dd4fd7f8a23f3652f9a39R172-R179) explain the increased compilation time? You state that XLA may nevertheless choose to use a `select` instead of branching if it is deemed advantageous, but could this process of inference lead to super linear compilation times? 

In my application, I chose to use a `switch` instead of a `select` because I feared long compilation times, but now it seems under the hood it is ignoring this. If this is indeed the culprit, it would be great if there was an option to force jax/xla to forgo on this optimization to lower the compilation times.

froystig (approved):

PR #13591, opened 2022-12-09
[TPU CI] Run build matrix on v3-8 as well as v4-8

skye:
We're seeing failures on v3-8 that don't appear on the current v4-8 testing. v3-8 also exposes 8 devices (vs. v4-8 exposes 4), and some tests needs 8 devices to run.

I just added a v3-8 runner VM.

Also adds a missing pip install command (I only caught this with a fresh runner since it only needs to be installed once).


skye:
Sample run: https://github.com/google/jax/actions/runs/3661198892/jobs/6189172285

I'm curious to see if this repros the random_test.py issue...

yashk2810 (approved):

PR #13618, opened 2022-12-12
Add checkify.debug_check which is a no-op outside of checkify.

LenaMartens:
`checkify.debug_check` is a check which only executes under a `checkify` transform, otherwise the check will be dropped.

```
def f(x):
  checkify.debug_check(jnp.all(x != x), "{x} cannot be {x}", x=x)
f(1.)  # no error
err, _ = checkify.checkify(f)(1.)
err.throw() #raises error
```

### Motivation

Library writers might want to add `check`s to their code without forcing their users to call `checkify`. `debug_check` allows them to encode predicates in their code which will run under "debug" mode, but will otherwise have no effect on the execution of the code. We could later expose a `debug_mode` which would run all checks, including those in library code, automatically. (eg. https://github.com/google/jax/pull/12597)

### ~Why a `pred_fun` and not a `pred`?~

Making the `pred` a callable means we can defer executing the predicate function to when we need it, and users can implement expensive validation logic which will not be run  if not necessary. Is this worth the API complexity? 
**Decision: NO, let's stick with a `pred` for now and rely on XLA to DCE.**

### ~Why a new primitive?~
Not sure, should it be a parameter on `check_p` instead? 
**Decision: YES, let's unify those primitives**

### Could this be a bad idea?
Could it be confusing to the user when these checks are silently dropped when not under `checkify`, or is the name `debug_check` clear enough?

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -243,7 +243,7 @@ class Error:
     return None
 
   def throw(self):
-    check_error(self)
+    _check_error(self)
 
   def __str__(self):
     return f'Error({self.get()})'
@@ -605,7 +605,7 @@ def check(pred: Bool, msg: str, *fmt_args, **fmt_kwargs) -> None:
   Before staging a function with checks, :func:`~checkify` it!
 
   Args:
-    pred: if False, an error is added.
+    pred: if False, a FailedCheckError error is added.
     msg: error message if error is added. Can be a format string.
     fmt_args, fmt_kwargs: Positional and keyword formatting arguments for
       `msg`, eg.:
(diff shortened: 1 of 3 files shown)


LenaMartens:
After discussion with @sharadmv: let's go for a `pred` instead of a `pred_fun`, because we can usually rely on XLA to DCE our unused validation code.
Also, let's make "debug" a parameter on `check_p` instead of making a new primitive.

jax/_src/checkify.py
```diff
@@ -725,15 +773,21 @@ class JaxRuntimeError(ValueError):
...
   error = tree_unflatten(err_tree, args)
   exc = error.get_exception()
   if exc:
     raise JaxRuntimeError(str(exc)) from exc
   return []
 
 @check_p.def_effectful_abstract_eval
-def check_abstract_eval(*args, err_tree):
+def check_abstract_eval(*args, err_tree, debug):
+  if debug:
```
sharadmv, line 788: I'm not sure if we should mark it as not effectful. Rules should still be aware of them.

I think the right approach is to have debug effects be lowerable.

sharadmv (changes requested):

sharadmv:
The main thing I'm concerned about with this implementation is the `abstract_eval` rule. I think `debug_check`s should still be "effectful" but just a lowerable effect. That would involve making a new Effect type that is lowerable and have DebugCheck emit that. Does that sound reasonable?

jax/_src/checkify.py
```diff
@@ -631,18 +631,70 @@ def check(pred: Bool, msg: str, *fmt_args, **fmt_kwargs) -> None:
...
+      If its output is False, a FailedCheck error is added.
+    msg: error message if error is added.
+    fmt_args, fmt_kwargs: Positional and keyword formatting arguments for
+      `msg`, eg.:
+      ``check(.., "check failed on values {} and {named_arg}", x, named_arg=y)``
+      Note that these arguments can be traced values allowing you to add
+      run-time values to the error message.
+      Note that tracking these run-time arrays will increase your memory usage,
+      even if no error happens.
+
```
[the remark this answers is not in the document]
LenaMartens: Not sure what you mean here, are you saying we might lose the `debug_check` under DCE? The `debug_check` will never show up at an XLA level, so I don't think it's something to worry about (maybe I'm misunderstanding your point)

jax/_src/checkify.py
sharadmv: It's the same reason we have `jax.debug.print` have an effect despite it being lowerable. Other consumers of the jaxpr might want to be aware of the effects present for their own purposes. For example, when we see "ordered effects" in a jaxpr at lowering time, we thread tokens to sequence them. The lowering needs to know about that *beforehand* so it knows to have tokens at the ready.

The analogy would be that in `jax.debug_mode` , we might run things asynchronously if they have *no* error effects. So we do want to reveal that there are `debug_check`s via the jaxpr.

jax/_src/checkify.py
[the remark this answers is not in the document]
sharadmv: Computing the predicate will show up in the XLA computation.

jax/_src/checkify.py
sharadmv: > lowerable if the experimental runtime XLA checks flag is True, and make a separate effect for debug checks, WDYT?

I don't think you need to make a separate effect, this works for now.

jax/_src/checkify.py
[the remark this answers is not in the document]
LenaMartens: yeah, so you mean a note about "XLA can DCE your predicate fun"?

jax/_src/checkify.py
LenaMartens: > The analogy would be that in jax.debug_mode , we might run things asynchronously if they have no error effects. So we do want to reveal that there are debug_checks via the jaxpr.

But either that `debug_check` will not be run (because it will never "run") in which case it should not defeat async-ness, or its effect has been discharged into an `Error` value, in which case the output error will inherently contain that effect. My prototyped debug_mode actually checked the output `Error` to decide if it would run sync or async, because you could tell from an empty Error value that no checks are active. You could imagine that check to be on the union of the Error value "effects" and jaxpr effects?

LenaMartens: from offline discussion: we can use effects as a cheap static analysis, and determine whether a (debug)-check is in the code without needing to checkify!

sharadmv (approved):

PR #13624, opened 2022-12-13
[sparse] add CheckGradsSparse utility function

jakevdp:
cc/ @tlu7


jakevdp:
The failures are real! I think there's some issue in a corner case of the transpose rule for `bcoo_todense`. I'm going to work on a fix.

jakevdp:
Fixes in #13654; I'll rebase here once that's merged.

froystig (approved):

Issue #11737, opened 2022-08-04
Add directional distributions


carlosgmartin:
Add the following [directional](https://en.wikipedia.org/wiki/Directional_statistics) distributions to [jax.random](https://jax.readthedocs.io/en/latest/jax.random.html) (for sampling) and/or [jax.scipy.stats](https://jax.readthedocs.io/en/latest/jax.scipy.html#jax-scipy-stats) (for probability density function evaluation):

- [ ] `vonmises`: [von Mises distribution](https://en.wikipedia.org/wiki/Von_Mises_distribution)
- [ ] `wrapped_normal`: [wrapped normal distribution](https://en.wikipedia.org/wiki/Wrapped_normal_distribution)
- [ ] `wrapped_cauchy`: [wrapped Cauchy distribution](https://en.wikipedia.org/wiki/Wrapped_Cauchy_distribution)
- [ ] `vonmises_fisher`: [von Mises–Fisher distribution](https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution)

huangziwei:
`jax.scipy.stats.vonmises` currently (jax + jaxlib == 0.4.2) doesn't take `loc` and `scale` as arguments. Is it intended?

stannum13:
I'd like to work on the first reviewable piece of this issue: complete von
Mises support.

My proposed PR would add a `jax.random` sampler implemented with the
Best--Fisher rejection algorithm, including broadcasting, `jit`/`vmap`, typed
keys, explicit output sharding, `kappa == 0` uniform behavior, and numerical
coverage from very small to very large concentrations. It would also make the
existing `jax.scipy.stats.vonmises.pdf/logpdf` accept `kappa == 0` and add a
circular location argument. I would leave implicit/reparameterization
gradients for a separate follow-up.

Before implementing, could a maintainer confirm three API choices?

1. Should the random function be named `von_mises` (consistent with JAX's
   underscore-separated names) or `vonmises` (matching NumPy)?
2. Is it preferable to include the small `jax.scipy.stats.vonmises` fixes in
   the sampler PR, or submit them separately?
3. For SciPy signature compatibility, should `pdf/logpdf` accept the ignored
   `scale=1` argument as SciPy does, or should JAX expose only `loc`?

The implementation will be based on Best & Fisher (1979), with NumPy and
SciPy used as behavioral references rather than requiring identical random
streams.

PR #13647, opened 2022-12-14
JAX Implementation of Power Spherical logpdf & pdf

harryjulian:
As referenced in #11737. 

Paper is [here.](https://arxiv.org/pdf/2006.04437.pdf)

Initial implementation. Tested it in notebooks re: the authors [example](https://github.com/nicola-decao/power_spherical) and the [Tensorflow Probability implementation](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/PowerSpherical) and looks good. 

This one is a little less straightforward; I've got a few of questions:

1) As these functions aren't in `scipy.stats`, should they live there? It makes the most sense to be with all other distributions imo.
2) When it comes to writing tests, as these functions aren't in `scipy.stats`, how should I go about validating their correctness without adding extra dependencies?
3) As the two arguments `mu` and `kappa` are usually of alternative shapes, what is the most appropriate way to conduct type promotion?

Edit: looks like I made a bit of a mess with this PR, I'll re-open post discussion.

### DIFF
--- a/jax/_src/scipy/stats/power_spherical.py
+++ b/jax/_src/scipy/stats/power_spherical.py
@@ -1,0 +1,34 @@
+# Copyright 2022 The JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import jax.numpy as jnp
+
+from jax import lax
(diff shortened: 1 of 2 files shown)


jakevdp:
Hi - thanks for the contribution. I think this lies in a bit of a gray area because the corresponding distribution isn't implemented in `scipy.stats`. We don't have any hard policy regarding what is in scope for `jax.scipy.stats`, but generally adding functionality beyond what is in scipy requires some discussion.

Who do you anticipate is the audience for this functionality? Can we think of a better home for it rather than including it in jax itself?

harryjulian:
Figures. Taking a quick look at the [citations](https://scholar.google.com/scholar?cites=13182117127367358760&as_sdt=2005&sciodt=0,5&hl=en) the paper has on scholar, it looks like it's mainly used in a variety of more niche ML use cases (robotics, an embedded loss function, generative modelling applications). As the work is quite new, perhaps its main use cases are yet to come?

I can't think of any other specific Jax-based libraries which would be a better fit, off the top of my head.

jakevdp:
I think I'm going to close this – it's not clear that these routines are in scope for `jax.scipy`. Thanks for the contribution, and sorry we weren't able to merge it this time!

PR #13655, opened 2022-12-14
Make input for tridiagonal test symmetric/hermitian

rahulbatra85:
Tridiagonal test input should be symmetric/hermitian and not a random matrix

### DIFF
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -1345,12 +1345,14 @@ class ScipyLinalgTest(jtu.JaxTestCase):
   def testTridiagonal(self, shape, dtype, lower):
     rng = jtu.rand_default(self.rng())
     def jax_func(a):
+      a = (a + T(a.conj())) / 2 #Make input hermitian/symmetric
       return lax.linalg.tridiagonal(a, lower=lower)
 
     real_dtype = jnp.finfo(dtype).dtype
     @partial(np.vectorize, otypes=(dtype, real_dtype, real_dtype, dtype),
             signature='(n,n)->(n,n),(n),(k),(k)')
     def sp_func(a):
+      a = (a + T(a.conj())) / 2 #Make input hermitian/symmetric
       if dtype == np.float32:
         c, d, e, tau, info = scipy.linalg.lapack.ssytrd(a, lower=lower)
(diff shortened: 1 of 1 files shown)


rahulbatra85:
@hawkinsp I may be wrong here, but I believe there is a bug in tridiagonal test. The input should be symmetric/hermitian and not a random matrix.  If you agree, then let me know if the fix makes sense.

jakevdp:
Hi - thanks for looking into this! I think the test is fine as-is, because both `lax.linalg.tridiagonal` and `scipy.linalg.lapack.ssytrd` take an argument that specifies whether the lower or upper triangle is referenced, and the rest of the matrix will be ignored. Because of this, it's actually useful if the input matrix is not hermitian/symmetric, because otherwise the test would not detect if the `lower` argument is behaving incorrectly!

With that in mind, I think we should keep the test as-is – what do you think?

rahulbatra85:
@jakevdp ok, I see your point about the lower flag, but what about the case where input matrix elements are complex numbers, but some diagonal elements are not real. Is this a valid input? For example, something like this
[[ 1+5j  2-j ]
 [ 4+j    2-j ]]

jakevdp:
It looks like for complex inputs, the function discards the imaginary part of the diagonal:
```python
import jax.numpy as jnp
from jax import lax
x = jnp.array([[1+5j, 2-1j],
               [4+1j, 5-1j]])
for output in lax.linalg.tridiagonal(x):
  print(output)
# [[ 1.       +0.j  2.       -1.j]
#  [-4.1231055+0.j  5.       +0.j]]
# [1. 5.]
# [-4.1231055]
# [1.9701426+0.24253564j]

x2 = jnp.array([[1   , 2-1j],
                [4+1j, 5   ]])
for output in lax.linalg.tridiagonal(x):
  print(output)
# [[ 1.       +0.j  2.       -1.j]
#  [-4.1231055+0.j  5.       +0.j]]
# [1. 5.]
# [-4.1231055]
# [1.9701426+0.24253564j]
```
This behavior has test coverage because the inputs are not symmetrized.

rahulbatra85:
ok, that makes sense now. Thanks!

I will close this PR

PR #13657, opened 2022-12-14
Document why 'import name as name' is used

jakevdp:
This might be excessive, but this question comes up at least a couple times a quarter, so I figure this will help clarify things for people reading the code!


sharadmv (approved):
Thanks so much!

Issue #13656, opened 2022-12-14
Bug in vjp of scatter_add with empty array


jakevdp:
Found in the process of testing #13624. Minimal reproduction:
```python
import jax
import jax.numpy as jnp

x = jnp.zeros(0)
y = jnp.ones(1)
ind = jnp.arange(1)

def f(y):
  return x.at[ind].add(y, mode='drop')

print(f(y))
# Array([], dtype=float32)

print(jax.jacfwd(f)(y))
# Array([], dtype=float32)

print(jax.jacrev(f)(y))
# Traceback (most recent call last):
#   File "tmp.py", line 16, in <module>
#     print(jax.jacrev(f, (0, 2))(x, m, y))
#   File "/Users/vanderplas/github/google/jax/jax/_src/api.py", line 1387, in jacfun
#     jac = vmap(pullback)(_std_basis(y))
# TypeError: Slice size at index 1 in gather op is out of range, must be within [0, 0 + 1), got 1.
```

jakevdp:
From @hawkinsp in chat – this might be an unfixable issue, due to poor semantics of `gather` in the presence of zero-sized arrays.  See for example this comment: https://github.com/google/jax/blob/b468e31e13f9a8695bd6fcb885b855f9dd3da7d9/jax/_src/lax/slicing.py#L1317-L1321

PR #13661, opened 2022-12-15
[sparse] fix autodiff rule for bcoo_sum_duplicates on zero-sized arrays

jakevdp:
Workaround for the issue reported in #13656


jakevdp:
Test failure here should be fixed by #13663

froystig (approved):

PR #13717, opened 2022-12-19
[sparse] improve implementation of sparse.grad & sparse.value_and_grad

jakevdp:
Rather than using the `allow_int` trick, this uses flattening of input buffers. This will pave the way for adding sparse implementations of `jacfwd` and `jacrev`.

Second commit adds wrappers for `jacfwd` and `jacrev` for computing gradients of functions that return a single array. Needs a bit more work to support multiple return values.

### DIFF
--- a/jax/experimental/sparse/ad.py
+++ b/jax/experimental/sparse/ad.py
@@ -12,90 +12,166 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import itertools
 from typing import Any, Callable, Sequence, Tuple, Union
-
-import numpy as np
 
 import jax
 from jax import core
 from jax import tree_util
 from jax._src.api_util import _ensure_index, _ensure_index_tuple
 from jax.util import safe_zip
-from jax._src.util import wraps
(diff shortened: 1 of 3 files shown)


jax/experimental/sparse/ad.py
```diff
@@ -12,90 +12,166 @@
...
+def flatten_fun_for_sparse_ad(fun, argnums: Union[int, Tuple[int]], args: Tuple[Any]):
+  argnums_tup = _ensure_index_tuple(argnums)
+  assert all(0 <= argnum < len(args) for argnum in argnums_tup)
+
+  # We do a two-step flattening to figure out how argnums maps to args_flat.
+  # First, flatten arguments to a list containing sparse and dense objects.
+  args_flat1, tree1 = tree_util.tree_flatten(args, is_leaf=is_sparse)
+  *leaf_argnums1, end = split_list(range(tree1.num_leaves),
+                                   [child.num_leaves for child in tree1.children()])
+  assert not end
```
froystig, line 40: Nice.

froystig (approved):
Nice!

PR #13728, opened 2022-12-20
[jax2tf] Fixes for handling of opaque dtypes for slice/update_slice/g…

gnecula:
…ather

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -857,8 +857,10 @@ def _jax_physical_aval(aval: core.ShapedArray) -> core.ShapedArray:
   there is only one and return it.
   """
   if core.is_opaque_dtype(aval.dtype):
-    aval, = aval.dtype._rules.physical_avals(aval)
-    return aval
+    physical_aval, = aval.dtype._rules.physical_avals(aval)
+    assert (len(physical_aval.shape) >= len(aval.shape) and
+            physical_aval.shape[:len(aval.shape)] == aval.shape), (physical_aval, aval)
+    return physical_aval
   return aval
 
 def _jax_physical_dtype(dtype):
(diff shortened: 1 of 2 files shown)


jax/experimental/jax2tf/jax2tf.py
```diff
@@ -2566,9 +2568,16 @@ def _gather(operand, start_indices, *, dimension_numbers, slice_sizes: core.Shap
         indices_are_sorted=indices_are_sorted, fill_value=fill_value,
         output_shape=_out_aval.shape, _in_avals=_in_avals, _out_aval=_out_aval)
 
+  operand_aval = _in_avals[0]
   start_indices = _maybe_cast_to_int64(start_indices)
+  if core.is_opaque_dtype(operand_aval.dtype):
```
gnecula, line 2573: WDYT about this simpler way of customizing the behavior for opaque types, instead of creating new _rules for jax2tf? Alternatively, we can create rules for adjusting the parameters of gather/slice/update_slice.

gnecula:
It was somewhat lucky that I discovered this: we had the test_prng created a few days ago, and then I discovered that many shape polymorphism tests were only doing the conversion and constructing the tf.Graph, but not running it. Once I started to run the converted function I got these failures.

I am actually concerned that there are many more places where opaque types should get involved in jax2tf, e.g., everywhere we call _eval_shape. Also in no_xla_impl.py. I wish we had a better way to deal with them, e.g., a way to rewrite a JAX eqn to eliminate opaqueness.

froystig:
To address a broader question raised here, and arguably in a recent PR or two as well, I think what we're seeing are the advantages of opaque dtypes "lowering to physical jaxpr" (i.e. without opaque dtypes), rather than lowering directly to MLIR. Basically as you say:

> I wish we had a better way to deal with them, e.g., a way to rewrite a JAX eqn to eliminate opaqueness.

I think we should try to do this. It would simplify things and still seems to capture current use cases. In a conversation we had today, @mattjj also highlighted that resolving opacity within Jaxpr is easier for supporting multiple backends. We initially added opaque dtype lowering thinking about a single backend (MLIR), but not many of them (jax2tf, etc.).

jax/experimental/jax2tf/jax2tf.py
froystig: This seems fine, and overall more consistent with thinking ahead to grounding out opaque dtypes within jaxpr (resolving jaxpr to physical-only jaxpr). If we can do that, then this sort of logic would actually just move up into the jaxpr-to-jaxpr rewrite instead. Would you agree?

I think having "rules" as we do currently would be fine if we only had one backend lowering target. With several targets, we end up with a set of rules for each, and a lot of conceptual redundancy.

froystig (approved):

jax/experimental/jax2tf/jax2tf.py
gnecula: I agree. This deserves to be resolved in a distinct pass just before lowering.

There are other things that could fall in the same category, e.g., effects.

PR #13744, opened 2022-12-21
[jax2tf] Add autograph=False everywhere we use tf.function

gnecula:


marcvanzee:
I was always wondering what the right default would be here, I suppose we turn this off since we don't want to compile control flow ops to TF ops automatically, since JAX handles them as well?

marcvanzee:
By the way, you removed me from review again but I suppose that was unintended so I approved anyway.

marcvanzee (approved):

gnecula:
We should never use autograph=True, because it will try to convert to TF the actual code in jax2tf.py (those are the conditionals it sees). We have recently discovered that this can lead to a 10x increase in TF tracing time. See b/261971172.

marcvanzee:
Interesting, thanks for the info.

PR #13639, opened 2022-12-13
Instructions to create custom operation for GPUs

seanprime7:
This tutorial explains how to define a custom operation for GPUs and use it in single and multi GPU environments.

PR #13752, opened 2022-12-21
Custom call doc

nouiz:
Follow up from https://github.com/google/jax/pull/13639/ as I can't update it directly.


nouiz:
Closing. The changes have been merged in https://github.com/google/jax/pull/13639.

PR #13761, opened 2022-12-22
Make `jax.ensure_compile_time_eval` correctly exposed as a public API

wookayin:
This function was added as a public API (#7987) but py.type static
checkers do not recognize it as a public API because of the alias name.

`jax.eval_context` exists only for backward compatibility, so the
correct import would be to import `ensure_compile_time_eval` directly
from `jax._src.core`.


mattjj:
Thank you for the expert fix!

mattjj (approved):

PR #13773, opened 2022-12-22
Added jax.random.vonmises

harryjulian:
As referenced in #11737.

This is the initial implementation, based on the reference provided in the docstring.

Half of the tests are passing at the moment, but having some dtype issues with `lax.select` and `lax.ge`. Also having some trouble ensuring the returned array is of the specified dtype.

Any guidance would be appreciated!

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1724,3 +1724,85 @@ def ball(
   g = generalized_normal(k1, p, (*shape, d), dtype)
   e = exponential(k2, shape, dtype)
   return g / (((jnp.abs(g) ** p).sum(-1) + e) ** (1 / p))[..., None]
+
+def vonmises(
+  key: KeyArray,
+  x: float,
+  kappa: float,
+  shape: Optional[Shape] = None,
+  dtype: DTypeLikeFloat = dtypes.float_
+) -> Array:
+  """Draw samples from a von Mises distribution.
+
+  Reference:
+    Best, D. J., & Fisher, N. I. (1979). Efficient Simulation of the
+    von Mises Distribution. Applied Statistics, 28(2), 152.
+    https://doi.org/10.2307/2346732
+
+  Args:
+    key: a PRNG key used as the random key.
+    x: mean of the distribution.
+    kappa: concentration of the distribution.
+    shape: shape of the output array.
+    dtype: dtype of the output array.
+
+  Returns:
+    theta: A random array of specified shape and dtype."""
+  # Run Checks
+  key, _ = _check_prng_key(key)
+  if not dtypes.issubdtype(dtype, np.floating):
+    raise ValueError(f"dtype argument to `t` must be a float"
+                     f"dtype, got {dtype}.")
+  dtype = dtypes.canonicalize_dtype(dtype)
+  if shape is not None:
+    shape = core.canonicalize_shape(shape)
+  else:
+    shape = np.shape(x)
+
+  def _true():
+    return jnp.pi * 2 * uniform(key, shape, dtype)
+
+  def _false():
+    return _vonmises(key, x, kappa, shape, dtype, 100)
+
+  # Return Distribution
+  return lax.cond(kappa <= 1e-6, _true, _false)
+
+@partial(jit, static_argnums = (3, 4, 5), inline = True)
+def _vonmises(key, x, kappa, shape, dtype, max_iters):
+
+  def _body_fn(pars):
+    iteration, key, theta_out, accepted = pars
+    key, subkey1, subkey2, subkey3 = _split(key, 4)
+
+    u1 = uniform(subkey1, shape, dtype)
+    u2 = uniform(subkey2, shape, dtype)
+    u3 = uniform(subkey3, shape, dtype)
+    z = lax.cos(jnp.pi * u1)
+    f = (1 + r * z) / (r + z)
+    c = kappa * (r - f)
+
+    accept = (lax.ge((c * (2 - c) - u2), zero)) | (lax.ge((lax.log(c / u2) + 1) - c, zero))
+    result = (jnp.sign(u3 - 0.5) * jnp.arccos(f)) + x
+    out = jnp.mod(jnp.abs(result) + jnp.pi, (2 * jnp.pi)) - jnp.pi
+    theta = jnp.sign(result) * out
+    theta_out = lax.select(accept, theta, theta_out)
+    accepted |= accept
+
+    return iteration + 1, key, theta_out, accepted
+
+  def _cond_fn(pars):
+    iteration, _, _, accepted = pars
+    return (~accepted).any() & (iteration < max_iters)
+
+  x = lax.convert_element_type(x, dtype)
+  kappa = lax.convert_element_type(kappa, dtype)
+  zero = _lax_const(x, 0)
+  a = 1 + lax.sqrt(1 + 4 * kappa * kappa)
+  b = (a - lax.sqrt(2 * a)) / (2 * kappa)
+  r = (1 + b * b) / (2 * b)
+
+  theta_init = lax.full(shape, -1, dtype)
+  accepted_init = lax.full(shape, False, jnp.bool_)
+  _, _, theta, _ = lax.while_loop(_cond_fn, _body_fn, (0, key, theta_init, accepted_init))
+  return theta.astype(dtype)
--- a/tests/random_test.py
+++ b/tests/random_test.py
@@ -1489,6 +1489,20 @@ class LaxRandomTest(jtu.JaxTestCase):
                                      axis=axis, shape=shape)
     self.assertEqual(samples.shape, shape)
 
+  @jtu.sample_product(
+    x = [-1000.0, -100.0, -10.0, -1.0, 0.0, 1.0, 10.0, 100.0, 1000.0],
+    kappa = [0.1, 0.5, 1.0, 2.0, 10.0, 100.0, 1000.0],
+    dtype=float_dtypes)
+  def testVonmises(self, x, kappa, dtype):
+    key = self.seed_prng(0)
+    rand = lambda key: random.vonmises(key, x, kappa, shape = (10000,), dtype = dtype)
+    crand = jax.jit(rand)
+
+    uncompiled_samples = rand(key)
+    compiled_samples = crand(key)
+
+    for samples in [uncompiled_samples, compiled_samples]:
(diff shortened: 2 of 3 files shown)


jakevdp:
Thanks - I'll take a closer look soon, but in the meantime it looks like there are some type mismatch errors when running in X64 mode; you can run those tests locally with
```
$ JAX_ENABLE_X64=1 pytest -n auto tests/random_test.py
```

jax/_src/random.py
```diff
@@ -1724,3 +1724,83 @@ def ball(
...
+
+  # Return Distribution
+  if kappa <= 1e-6:
+    return ((jnp.pi * 2) * uniform(key, shape, dtype)).astype(dtype)
+  else:
+    return _vonmises(key, x, kappa, shape, dtype, 100)
+
+#@partial(jit, static_argnums = (3, 4, 5), inline = True)
+def _vonmises(key, x, kappa, shape, dtype, max_iters):
+
```
soraros, line 1767: For the type issue, add 
```suggestion
  x = lax.convert_element_type(x, dtype)
  kappa = lax.convert_element_type(kappa, dtype)
```

harryjulian: Worked like a charm. Think I've found the other issues, shall update.

jax/_src/random.py
```diff
@@ -1184,7 +1184,7 @@ def _poisson_rejection(key, lam, shape, dtype, max_iters) -> Array:
   def body_fn(carry):
     i, k_out, accepted, key = carry
     key, subkey_0, subkey_1 = _split(key, 3)
-
+
```
jakevdp, line 1187: stray revert?

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+  # Run Checks
+  key, _ = _check_prng_key(key)
+  if not dtypes.issubdtype(dtype, np.floating):
+    raise ValueError(f"dtype argument to `t` must be a float"
+                     f"dtype, got {dtype}.")
+  dtype = dtypes.canonicalize_dtype(dtype)
+  shape = core.canonicalize_shape(shape)
+
+  # Return Distribution
+  if kappa <= 1e-6:
```
jakevdp, line 1760: This requires kappa to be a static parameter. If that's not your intent, you could use `lax.cond` instead.

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+    return ((jnp.pi * 2) * uniform(key, shape, dtype)).astype(dtype)
```
jakevdp, line 1761: `astype(dtype)` is redundant here and can be removed.

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+
+  def _body_fn(pars):
+    iteration, key, theta_out, accepted = pars
+    key, subkey1, subkey2, subkey3 = _split(key, 4)
+
+    u1 = uniform(subkey1, shape, dtype)
+    u2 = uniform(subkey2, shape, dtype)
+    u3 = uniform(subkey3, shape, dtype)
+    z = lax.cos(jnp.pi * u1)
+    f = (one + r * z) / (r + z)
```
jakevdp, line 1776: `1 + r * z` is fine here

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+    return (~accepted).any() & (iteration < max_iters)
+
+  x = lax.convert_element_type(x, dtype)
+  kappa = lax.convert_element_type(kappa, dtype)
+  zero = _lax_const(x, 0)
+  half = _lax_const(x, 0.5)
+  one = _lax_const(x, 1)
+  two = _lax_const(x, 2)
+  four = _lax_const(x, 4)
+  twopi = _lax_const(x, (2 * jnp.pi))
```
jakevdp, line 1800: For the most part these constants are unnecessary, and you can just use the literal values directly. The one exception is when you write `lax.ge(expression, zero)` because lax functions do not implicitly promote inputs, but you could use `expression >= 0` instead to avoid the need for strict type matching of inputs.

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+    key, subkey1, subkey2, subkey3 = _split(key, 4)
+
+    u1 = uniform(subkey1, shape, dtype)
+    u2 = uniform(subkey2, shape, dtype)
+    u3 = uniform(subkey3, shape, dtype)
+    z = lax.cos(jnp.pi * u1)
+    f = (one + r * z) / (r + z)
+    c = kappa * (r - f)
+
+    accepted = lax.ge((c * (two - c) - u2), zero) | lax.ge((lax.log(c / u2) + one) - c, zero)
```
jakevdp, line 1779: Using literal `0`, `1`, and `2` is fine here.

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
   g = generalized_normal(k1, p, (*shape, d), dtype)
   e = exponential(k2, shape, dtype)
   return g / (((jnp.abs(g) ** p).sum(-1) + e) ** (1 / p))[..., None]
+
+def vonmises(
+  key: KeyArray,
+  x: float,
+  kappa: float,
+  shape: Shape = (),
```
jakevdp, line 1732: Other distributions allow arbitrarily-sized arrays for the distribution parameters: see how `gamma` treats the `a` parameter, for example. We should follow that convention here.

jax/_src/random.py
```diff
@@ -1724,3 +1724,85 @@ def ball(
...
+  kappa = lax.convert_element_type(kappa, dtype)
+  zero = _lax_const(x, 0)
+  a = 1 + lax.sqrt(1 + 4 * kappa * kappa)
+  b = (a - lax.sqrt(2 * a)) / (2 * kappa)
+  r = (1 + b * b) / (2 * b)
+
+  theta_init = lax.full(shape, -1, dtype)
+  accepted_init = lax.full(shape, False, jnp.bool_)
+  _, _, theta, _ = lax.while_loop(_cond_fn, _body_fn, (0, key, theta_init, accepted_init))
+  return theta.astype(dtype)
```
jakevdp, line 1808: If this `astype` is necessary, I'd probably consider it a bug: the computation should be performed in the specified dtype. For example, this would be important for code running on accelerators where the user may want to ensure that computations are done in `float32`, which is more performant than `float64`.

Perhaps you can ensure the correct dtype at the start of the computation?

tests/random_test.py
```diff
@@ -1489,6 +1489,20 @@ def test_categorical_shape_argument(self, shape, logits_shape_base, axis):
...
+  def testVonmises(self, x, kappa, dtype):
+    key = self.seed_prng(0)
+    rand = lambda key: random.vonmises(key, x, kappa, shape = (10000,), dtype = dtype)
+    crand = jax.jit(rand)
+
+    uncompiled_samples = rand(key)
+    compiled_samples = crand(key)
+
+    for samples in [uncompiled_samples, compiled_samples]:
+      self._CheckKolmogorovSmirnovCDF(samples, scipy.stats.vonmises(loc = x, kappa = kappa).cdf)
```
jakevdp, line 1505: I'd add a test similar to `testGammaShape` to ensure that nontrivial combinations of `x` and `shape` produce the expected output

jakevdp (changes requested):
Thanks for making those changes! A few more suggestions below.

Also, to ensure this ends up in the docs, we should add `vonmisses` to the page here: https://github.com/google/jax/blob/main/docs/jax.random.rst

harryjulian:
Been sitting on this for a little bit now, just wondering you could offer some input!

I've done most of the re-factoring, but I'm unsure on the best way to work the conditional in with the new broadcasting. Here's the code-sample:

```
def vonmises(
  key: KeyArray,
  x: RealArray,
  kappa: RealArray,
  shape: Optional[Shape] = None,
  dtype: DTypeLikeFloat = dtypes.float_,
  max_iters: int = 100
) -> Array:
  """Draw samples from a von Mises distribution.

  Args:
    key: a PRNG key used as the random key.
    x: mean of the distribution. Must be broadcast-compatible with kappa.
    kappa: concentration of the distribution. Must be broadcast-comptible with x.
    shape: optional, shape of the output array.
    dtype: optional, dtype of the output array. (default float64 if
    max_iters: maximum number of times to run the algorithm
      that generates and accepts/rejects samples.

  Returns:
    theta: A random array of specified shape and dtype.

  Reference:
    Best, D. J., & Fisher, N. I. (1979). Efficient Simulation of the
    von Mises Distribution. Applied Statistics, 28(2), 152.
    https://doi.org/10.2307/2346732
  """
  # Check Key, Dtypes, Shape
  key, _ = _check_prng_key(key)
  if not dtypes.issubdtype(dtype, np.floating):
    raise ValueError(f"dtype argument to `t` must be a float"
                     f"dtype, got {dtype}.")
  
  if shape is None:
    shape = lax.broadcast_shapes(np.shape(x), np.shape(kappa))
  else:
    _check_shape("vonmises", shape, x, kappa)

  def _true(key: Array, x: Array, kappa: Array) -> Array:
    return jnp.pi * 2 * uniform(key, shape, dtype)

  def _false(key: Array, x: Array, kappa: Array)-> Array:
    return _vonmises(key, x, kappa, shape, dtype, 100)

  # Return Distribution
  return lax.cond(
    kappa < 1e-5, 
    lambda key, x, kappa: _true(key, x, kappa),
    lambda key, x, kappa: _false(key, x, kappa),
    key, x, kappa
  )
```

Unfortunately, `lax.select` doesn't play nicely with the` kappa < 1e-5` condition:

```
TypeError: Pred must be a scalar, got [[ True]
 [ True]
 [ True]
 [ True]
 [ True]
 [ True]
 [ True]
 [ True]
 [ True]
 [ True]] of shape (10, 1).
```
The problem is that, after broadcasting, for rows/columns where the corresponding value of kappa is below the threshold we want to apply the uniform approximation. Does lax.cond allow the activation of both conditions? Any ideas on best ways to overcome this?

jakevdp:
`lax.cond` requires a scalar condition. For a vector condition, you can use `lax.select`

jakevdp:
Sorry for losing track of this one – are you still interested in contributing this?

jakevdp:
It looks like this was addressed in #13355

isabelgaron:
[#13355](https://github.com/jax-ml/jax/pull/13355) added a log pdf function to the scipy stats package, but there is still no VonMises random sampler, or any circular samplers in jax.random

PR #13775, opened 2022-12-22
[sparse] propagate metadata through vmappable

jakevdp:
Extra medatadata had been dropped within `vmap` up to this point. This doesn't affect correctness of code, but could result in mapped operations not being lowered to cusparse on GPU.


sholtodouglas (approved):
LGTM!

PR #13785, opened 2022-12-24
Fix wrong cond-state discharge rule.

patrick-kidger:
`discharge_state` returns the original output and then the updated refs, not the other way around.


sharadmv (approved):
Thanks for the fix, would it be possible to add a test case?

sharadmv (approved):

PR #13790, opened 2022-12-25
[jax2tf] Enable Trace to handle dimension polynomials used as constants

gnecula:
This change enables the use of dimension polynomials wherever constaints are used. This would arise, e.g., when tracing `lambda x: x.shape[0]` in presence of shape polymorphism.

This won't be needed anymore once the --jax_dynamic_shapes improves its coverage to replace shape polymorphism.

The downside of this change is that it adds a code path to Trace.full_raise. An alternative would be to ask users to explicitly convert dimensions: `lambda x: core.dimension_as_value(x.shape[0])`. Both of these can be removed in the future, but the former has the advantage of being internal to JAX.

An alternative internal change in `Trace.full_raise` would be

```
if hasattr(val, "__jax_array__"): val = val.__jax_array__()
```

but I think that using `dimension_as_value` makes it clear what use case is addressed by this change.

### DIFF
--- a/jax/_src/core.py
+++ b/jax/_src/core.py
@@ -415,6 +415,8 @@ class Trace:
     self.sublevel = sublevel
 
   def full_raise(self, val) -> Tracer:
+    if hasattr(val, "dimension_as_value"):  # Used for shape_poly._DimPolynomial
+      val = val.dimension_as_value()
     if not isinstance(val, Tracer):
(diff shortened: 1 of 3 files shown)


gnecula:
@mattjj  PTAL

mattjj (approved):
Thanks!

PR #13804, opened 2022-12-27
Error on numpy masked array inputs.

### LINKED ISSUE #13795: JAX silently discards masks in numpy masked arrays
Reported in https://github.com/data-apis/array-api/issues/582
```python
>>> jnp.sin(np.ma.array([1, 2, 3], mask=[True, False, True]))
Array([0.84147096, 0.9092974 , 0.14112   ], dtype=float32)
```

jakevdp:
Fixes #13795 - there are some additional changes required for the C++ `device_put` path; I'll address those in a separate change.

### DIFF
--- a/jax/_src/abstract_arrays.py
+++ b/jax/_src/abstract_arrays.py
@@ -51,6 +51,13 @@ def canonical_concrete_aval(val, weak_type=None):
   return ConcreteArray(dtypes.canonicalize_dtype(np.result_type(val)), val,
                        weak_type=weak_type)
 
+def masked_array_error(*args, **kwargs):
+  raise ValueError("numpy masked arrays are not supported as direct inputs to JAX functions. "
+                   "Use arr.filled() to convert the value to a standard numpy array.")
+
+core.pytype_aval_mappings[np.ma.MaskedArray] = masked_array_error
+ad_util.jaxval_zeros_likers[np.ma.MaskedArray] = masked_array_error
+
 for t in array_types:
   core.pytype_aval_mappings[t] = canonical_concrete_aval
   ad_util.jaxval_zeros_likers[t] = zeros_like_array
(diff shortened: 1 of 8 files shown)


jax/_src/abstract_arrays.py
```diff
@@ -51,6 +51,13 @@ def canonical_concrete_aval(val, weak_type=None):
   return ConcreteArray(dtypes.canonicalize_dtype(np.result_type(val)), val,
                        weak_type=weak_type)
 
+def masked_array_error(*args, **kwargs):
+  raise ValueError("numpy masked arrays are not supported as direct inputs to JAX functions. "
```
froystig, line 55:
```suggestion
  raise ValueError("numpy masked arrays are not supported as direct inputs to `jax.numpy` functions. "
```
?

jax/_src/dispatch.py
```diff
@@ -1275,6 +1275,10 @@ def device_put(x, device: Optional[Device] = None) -> Tuple[Any, ...]:
 # TODO(phawkins): update users.
 xla.device_put = device_put
 
+def _device_put_masked_array(x, device: Optional[Device]):
+  raise ValueError("numpy masked arrays are not supported as direct inputs to JAX functions. "
```
froystig, line 1279:
```suggestion
  raise ValueError("numpy masked arrays are not supported as direct inputs to `jax.numpy` functions. "
```
?

froystig: etc. (several occurrences below)

froystig (approved):

jax/_src/abstract_arrays.py
jakevdp: I think the more general form is correct: it's not just `jax.numpy` functions, but *any* jax function (e.g. `jax.lax`, anything transformed with `jax.jit` or `jax.vmap`, etc.)

PR #13813, opened 2022-12-28
fix signature of shard arg handler

jakevdp:
Fix post-submit comment from #13804

### DIFF
--- a/jax/interpreters/pxla.py
+++ b/jax/interpreters/pxla.py
@@ -421,7 +421,7 @@ def _shard_token(x, devices, indices):
   return device_put(np.zeros((), dtype=np.dtype(np.bool_)), devices, replicate=True)
 shard_arg_handlers[core.Token] = _shard_token
 
-def _masked_array_error(*args, **kwargs):
+def _masked_array_error(x, devices, indices):
   raise ValueError("numpy masked arrays are not supported as direct inputs to JAX functions. "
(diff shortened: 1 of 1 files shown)


jax/interpreters/pxla.py
```diff
@@ -421,7 +421,7 @@ def _shard_token(x, devices, indices):
   return device_put(np.zeros((), dtype=np.dtype(np.bool_)), devices, replicate=True)
 shard_arg_handlers[core.Token] = _shard_token
 
-def _masked_array_error(*args, **kwargs):
+def _masked_array_error(x, devices, indices):
```
yashk2810, line 424: One question I had is which test triggers this ValueError?

You said that jax.jit(lambda x: x)(masked_array) errored on canonicalize_dtype_handler. So which test is triggering this error?

jakevdp: Nothing triggers this error currently, because canonicalize_dtype_handler catches it first. But you asked me to add this, so I did.

yashk2810 (approved):

jax/interpreters/pxla.py
yashk2810: Maybe a pjit test will trigger this, but I think we miscommunicated. I was just curious how things were working without this handler and canonicalize_dtype_handler was the answer :)

If no test can trigger this, then I don't think we need to add this! Sorry for the miscommunication :)

PR #13814, opened 2022-12-28
validate shape & dtype in ShapeDtypeStruct

jakevdp:
Followup to #13258

There was maybe some disagreement about this in #13258, but my opinion remains that passing `dtype=None` in `ShapeDtypeStruct` should be an error, rather than us trying to guess what the user might have intended. I'll be sure to run a global test on this to find out if anyone is depending on this behavior.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2988,7 +2988,9 @@ def device_get(x: Any):
 class ShapeDtypeStruct:
   __slots__ = ["shape", "dtype", "named_shape", "sharding"]
   def __init__(self, shape, dtype, named_shape=None, sharding=None):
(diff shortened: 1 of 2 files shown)


jakevdp:
Global presubmit showed several uses where shape entries are non-integer (either `DimPoly` or `None`). For simplicity, I dropped the step of mapping `operator.index` over the shape entries.

froystig:
> [...] several uses where shape entries are non-integer (either `DimPoly` or `None`)

The `DimPoly` ones should all be internal, right? Maybe we can distinguish an internal construction from a public one, and have more restrictive checks in the latter case?

froystig (approved):

PR #13816, opened 2022-12-28
cleanup: remove several unused imports across the package

jakevdp:
Technically our flake8 check should catch these, but we disable unused import checks on many files to avoid having to annotate `# skip: F401` on a line-by-line basis in top-level modules. The `setup.cfg` syntax doesn't have the flexibiilty to do this with more granularity, as far as I know.


sharadmv (approved):
Thanks!

Issue #13653, opened 2022-12-14
[sparse] bcoo_extract incorrect for zero sparse dimensions


jakevdp:
Found in the course of #13624 

Minimal repro:
```python
import jax.numpy as jnp
from jax.experimental import sparse

mat = jnp.zeros(3, dtype='float32')
indices = jnp.zeros((3, 2, 0), dtype='int32')

jax.jit(sparse.bcoo_extract)(indices, mat)  # Error
```

PR #13827, opened 2022-12-29
[sparse] fix corner case of bcoo_extract

jakevdp:
This fixes a bug that was introduced in #13653, and adds some test cases.

I discovered this via discrepancies in the JVP rule for `bcoo_fromdense`.

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -417,8 +417,9 @@ def _bcoo_extract_impl(indices, mat):
   else:
     result = mat.at[batch_ind + sparse_ind].get(mode='fill', fill_value=0)
   if n_sparse == 0 and nse != 1:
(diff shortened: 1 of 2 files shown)


froystig (approved):

jakevdp:
Digging into this further, and I think the current behavior is actually what is expected. `bcoo_extract` has the property that when duplicate indices are present, the value is extracted once for each index. This is somewhat strange in the normal course of things, as it means `bcoo_extract()` and `bcoo_todense()` are not strict inverseses of each other when indices are duplicated, but this property does ensure that `bcoo_extract()` and `bcoo_todense()` are transposes of each other!

This PR makes the two inverses in special cases, while breaking the transpose property in those cases (which is relied on in the transpose rules for both functions). We may well want to do this at some point, but it should be done more globally along with rewriting the two transpose rules.

With that in mind, I'm going to abandon this change for now.

PR #13831, opened 2022-12-31
[Typing] Add `Optional` to parameter `boundary` of `stft`.

JyChang012:
According to [doc](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.signal.stft.html#jax.scipy.signal.stft), parameter `boundary` should accept `None`, indicating no extension at both ends.
I found this issue when typing checking [one of my own project](https://github.com/JyChang012/espnet/blob/jax/asr_train/espnex/layers/stft.py) using `mypy`.


jakevdp (approved):
Looks good, thanks!

PR #13834, opened 2022-12-31
`ravel_pytree` now produces jit-compatible unravel functions

patrick-kidger:
Previously,
```python
_, unravel1 = ravel_pytree(pytree)
_, unravel2 = ravel_pytree(pytree)

def run(unravel, ...):
    ...

run(unravel1, ...)
run(unravel2, ...)
```
would unecessarily induce recompilation.

### DIFF
--- a/jax/_src/flatten_util.py
+++ b/jax/_src/flatten_util.py
@@ -12,18 +12,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
 import warnings
+from typing import Callable, Sequence
 
 import numpy as np
 
-from jax._src.tree_util import tree_flatten, tree_unflatten
-from jax._src.util import safe_zip, unzip2
+from jax._src.tree_util import tree_flatten, tree_unflatten, PyTreeDef
+from jax._src.util import safe_zip, unzip2, HashableArrayWrapper
+from jax._src.typing import DType, Shape
 
 import jax.numpy as jnp
 from jax._src import dtypes
 from jax import lax
 
 zip = safe_zip
+
+
+@dataclasses.dataclass(frozen=True)
+class UnravelPyTree:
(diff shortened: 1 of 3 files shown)


patrick-kidger:
Ping @mattjj

mattjj:
Thanks for this, Patrick. Sorry I've been so slow to respond for the last N months. (I wish I could say only "recently"!)

This is a great idea. I adapted it in #14954 to reuse the `Partial` class, and to use a tuple in place of a hashably-wraped numpy.ndarray (though we can switch to the latter if the need arises, e.g. if we want different hash functions or performance characteristics). I pasted the same test case.

Thanks for this improvement!

mattjj:
> FWIW the docs on jnp.split only mention supporting ints or arrays, not tuples. (I know that JAX departs from numpy in often disallowing arraylikes in place of arrays

Good catch!

Also I done goofed in another way: I was thinking of `HashablePartial` from shard_map.py (to be moved into util.py), not `Partial`.

patrick-kidger:
Hah, I feel like the various kinds of JAX-internal function wrappers are starting to get a bit complicated. Off the top of my head there's `Partial`, `HashablePartial`, `HashableFunction`, `_HashableCallableShim`, and callable pytrees (e.g. `checkify.Error`, or this PR).

Outside of core JAX I just use `eqx.Module` for all of these use cases. It should probably be possible to standardise on a single approach here too.

---

Side note, the issue fixed in this PR is a pretty common one in JAX -- e.g. `jax.{jit, grad, ...}` etc. all return a fresh closure each time.

mattjj:
> It should probably be possible to standardise on a single approach here too.

You might be a lumper. Internal JAX utilities are set up to make lumpers out themselves. 😛

mattjj:
We effectively merged this as #14954!

PR #13855, opened 2023-01-04
removing repeated average function implementation and docs rendering

### LINKED ISSUE #13853: The `jax.numpy.average` function is rendering repeatedly in Docs
### Description

This is being caused by repetition of the same function at the implementation side 
[reduction functions](https://github.com/google/jax/blob/a1c699f59dc6dcb594677751ada19b46386f0c26/jax/_src/numpy/reductions.py#L343)
I am going to open a pull request that can fix the mentioned issue

image

_No response_

CPU

### Additional system info

Linux

### NVIDIA GPU info

_No response_

fnhirwa:
Closes https://github.com/google/jax/issues/13853

The function `jax.numpy.average` is repeated many times on the documentation page [jax.numpy.average](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.average.html#jax-numpy-average)

This PR removes duplicated code that caused the issue
Thanks

### DIFF
--- a/jax/_src/numpy/reductions.py
+++ b/jax/_src/numpy/reductions.py
@@ -15,7 +15,7 @@
 import builtins
 from functools import partial
 import operator
-from typing import overload, Any, Callable, Literal, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Optional, Sequence, Tuple, Union
 import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks - I don't think this is the right fix. These typing overloads do serve a purpose: they ensure that when the function is used in typechecked code, the static type checker can infer the correct number of outputs. If your goal is fixing the docs, I'd suggest modifying the docs rather than the source code.

This documentation is generated via sphinx autodoc: perhaps that package has settings that control how overloaded function definitions are displayed?

hirwa-nshuti:
Yeah sure it is the issue with sphinx
you're right there might be support for this as if it is many overloads it may seem very weird

jakevdp:
#13860 improves the situation somewhat by preserving the `ArrayLike` alias rather than printing out the full union it represents.

hirwa-nshuti:
I think as these function overloading are needed we can close this PR

PR #13863, opened 2023-01-04
GPU tests: disable cusparse_lowering to avoid segfaults.

jakevdp:
We've started seeing failures in the GPU CI jobs: https://source.cloud.google.com/results/invocations/8a57fcfe-439c-4fea-9434-ae71d017625e/targets/jax%2Ftesting%2Fcpu%2Fpresubmit_github/log

I suspect this is due to increased test coverage, and similar to the other cusparse-related GPU segfaults we've been seeing.

### DIFF
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -721,7 +721,8 @@ jax_test(
 jax_test(
     name = "sparse_test",
     srcs = ["sparse_test.py"],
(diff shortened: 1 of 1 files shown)


jakevdp:
Hmm, looks like this wasn't the issue.

Now that I look more closely, it's the CPU test that's failing, not the GPU test.

I suspect it's probably a timeout rather than a cusparse issue.

PR #13865, opened 2023-01-04
[sparse] mark several slow tests

jakevdp:
Hopefully this will fix the Kokoro CPU timeout.


yashk2810 (approved):

jakevdp:
Well, this didn't fix the kokoro timeout, but I think this change is not incorrect so we'll keep it.

PR #13873, opened 2023-01-05
(add) add SmeLU activation function to nn.functions

ghost:
Hello 👋  This small PR adds Jax implementation of the Smooth ReLU activation function proposed in the paper ["Real World Large Scale Recommendation Systems Reproducibility and Smooth Activations"](https://arxiv.org/pdf/2202.06499.pdf) [arXiv 2022] by Gil I. Shamir, Dong Lin.


jax/_src/nn/functions.py
```diff
@@ -473,4 +473,24 @@ def hard_silu(x: Array) -> Array:
...
+  .. math::
+    \mathrm{smooth\_relu}(x)= \frac{(x+\beta)^2 }{4^\beta}
+  
+  Args:
+    x : input array of jnp
+    beta : beta value for Smooth ReLU
+  Return:
+    output activation array
+  """
+  assert beta >= 0., f"Beta must be equal or larger than zero, given {beta}"
```
jakevdp, line 492: We cannot do runtime value assertions of traced variables in JAX. Instead, you'll probably want to return `NaN` when `beta` is invalid, using `jnp.where` for JIT compatiblity.

jakevdp (changes requested):
Thanks for the contribution! It's looking good aside from the assertion issue mentioned below.

Also, before we merge we'll need some tests of this function. I think the best place would be in https://github.com/google/jax/blob/main/tests/nn_test.py; you can see some examples for other activation functions there. Thanks!

PR #13887, opened 2023-01-05
[sparse] generalize batch rule for bcoo_dot_general

jakevdp:
Previously, `vmap(bcoo_dot_general)` only supported `in_axes=None` or `in_axes=0` for the data and index arrays. It now supports any axis corresponding to a batch dimension of the sparse matrix.


jakevdp:
Looks great - mind if I share with the dev team for comments this morning?

froystig (approved):

PR #13917, opened 2023-01-08
[jax2tf] Fixed the conversion of a function that contains an inner pjit

gnecula:
In experimental_native_lowering when we convert a function that is not a jit or pjit, we wrap it with an implicit jit. We used to specify the backend when doing this conversion, which bypassed some logic in jit to handle the merging of jit and pjit code paths. We now drop the backend parameter to the implicit jit.

We also moved some pjit tests from jax2tf to sharding_test and dropped the old disabled test for teh GDAs, since GDAs are going away.


gnecula:
@maxwillzq @rxsang PTAL

rxsang (approved):

PR #13929, opened 2023-01-09
Add Open MPI automatic distributed initialization

nvcastet:
This PR let you initialize JAX simply with `jax.distributed.initialize()` when launching processes via `mpirun`/`mpiexec`.

### DIFF
--- a/docs/multi_process.md
+++ b/docs/multi_process.md
@@ -78,11 +78,12 @@ jax.distributed.initialize(coordinator_address="192.168.0.1:1234",
                            process_id=0)
 ```
 
-On Cloud TPU and Slurm environments, you can simply call {func}`jax.distributed.initialize()` with no
+On Cloud TPU, Slurm and Open MPI environments, you can simply call {func}`jax.distributed.initialize()` with no
 arguments. Default values for the arguments will be chosen automatically.
-When running on GPUs with Slurm, it is assumed that one process is started per GPU, i.e. each process will
+When running on GPUs with Slurm and Open MPI, it is assumed that one process is started per GPU, i.e. each process will
 be assigned only one visible local device. Otherwise it is assumed that one process is started per host,
 i.e. each process will be assigned all local devices.
+The Open MPI auto-initialization is only used when the JAX processes are launched via `mpirun`/`mpiexec`.
 
 ```python
 import jax
--- a/jax/_src/clusters/ompi_cluster.py
+++ b/jax/_src/clusters/ompi_cluster.py
@@ -1,0 +1,59 @@
+# Copyright 2023 The JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import re
+from typing import Optional
+from jax._src.clusters import ClusterEnv
+
+#  OMPI_MCA_orte_hnp_uri exists only when processes are launched via mpirun or mpiexec
+_ORTE_URI = 'OMPI_MCA_orte_hnp_uri'
+_PROCESS_COUNT = 'OMPI_COMM_WORLD_SIZE'
+_PROCESS_ID = 'OMPI_COMM_WORLD_RANK'
+_LOCAL_PROCESS_ID = 'OMPI_COMM_WORLD_LOCAL_RANK'
+
+class OmpiCluster(ClusterEnv):
+  @classmethod
+  def is_env_present(cls) -> bool:
+    return _ORTE_URI in os.environ
+
+  @classmethod
+  def get_coordinator_address(cls) -> str:
+    # Examples of orte_uri:
+    # 1531576320.0;tcp://10.96.0.1,10.148.0.1,10.108.0.1:34911
+    # 1314521088.0;tcp6://[fe80::b9b:ac5d:9cf0:b858,2620:10d:c083:150e::3000:2]:43370
+    orte_uri = os.environ[_ORTE_URI]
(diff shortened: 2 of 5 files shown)


docs/multi_process.md
```diff
@@ -78,9 +78,9 @@ jax.distributed.initialize(coordinator_address="192.168.0.1:1234",
                            process_id=0)
 ```
 
-On Cloud TPU and Slurm environments, you can simply call {func}`jax.distributed.initialize()` with no
+On Cloud TPU, Slurm and Open MPI environments, you can simply call {func}`jax.distributed.initialize()` with no
 arguments. Default values for the arguments will be chosen automatically.
-When running on GPUs with Slurm, it is assumed that one process is started per GPU, i.e. each process will
+When running on GPUs with Slurm and Open MPI, it is assumed that one process is started per GPU, i.e. each process will
```
nvlcambier, line 83: It would be worth documenting which of SLURM or OpenMPI has precedence, since both are often used in combination with each other.

nvcastet: Correct, I could clarify that, the "OpenMPI" initialization will only be used and also have higher precedence over Slurm when processes are launched via `mpirun` or `mpiexec` which will set `OMPI_MCA_orte_hnp_uri`. When launching application processes via `srun` with or without pmi/pmix feature, the Slurm initialization will be used.

jax/_src/clusters/ompi_cluster.py
```diff
@@ -0,0 +1,52 @@
...
+  def get_coordinator_address(cls) -> str:
+    # Example of orte_uri:
+    # 1531576320.0;tcp://10.96.0.1,10.148.0.1,10.108.0.1:34911
+    orte_uri = os.environ[_ORTE_URI]
+    job_id = orte_uri.split('.', maxsplit=1)[0][:5]
+    # Pick port in ephemeral range [(65535 - 2^12 + 1), 65535]
+    port = int(job_id) % 2**12 + (65535 - 2**12 + 1)
+    launcher_ip = orte_uri.split('//', maxsplit=1)[1]
+    launcher_ip = launcher_ip.split(':', maxsplit=1)[0]
+    launcher_ip = launcher_ip.split(',', maxsplit=1)[0]
```
skye, line 39: A regexp might be more readable but hard to say :)
`launcher_ip = re.search("tcp://([0-9.]+),", orte_uri).group(1)`
or
`launcher_ip = re.search("tcp://(.+?),", orte_uri).group(1)`

skye (changes requested):
LGTM overall, just a few comments + I like @nvlcambier's clarification ideas

docs/multi_process.md
nvlcambier: Ok good imho.
I didn't know that `mpirun` / `mpiexec` would set `OMPI_MCA_orte_hnp_uri`. If you have a link to OpenMPI's doc explaining that (if any), that would be worth adding somewhere.

jax/_src/clusters/ompi_cluster.py
```diff
@@ -0,0 +1,52 @@
...
+  @classmethod
+  def is_env_present(cls) -> bool:
+    return _ORTE_URI in os.environ
+
+  @classmethod
+  def get_coordinator_address(cls) -> str:
+    # Example of orte_uri:
+    # 1531576320.0;tcp://10.96.0.1,10.148.0.1,10.108.0.1:34911
+    orte_uri = os.environ[_ORTE_URI]
+    job_id = orte_uri.split('.', maxsplit=1)[0][:5]
```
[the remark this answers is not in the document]
nvlcambier: It could be useful to give `5` a name, so it's a bit more self-documenting.

docs/multi_process.md
nvcastet: I figured it out by experimenting with it.

nvlcambier: OK. Spelling it out in the code as a comment next to `_ORTE_URI = 'OMPI_MCA_orte_hnp_uri'` could be a good idea imho.

tests/multiprocess_gpu_test.py
```diff
@@ -187,6 +188,45 @@ def test_distributed_jax_visible_devices(self):
         for proc in subprocesses:
           proc.kill()
 
+  def test_gpu_ompi_distributed_initialize(self):
+    if jtu.device_under_test() != 'gpu':
+      raise unittest.SkipTest('Tests only for GPU.')
+    if shutil.which('mpirun') is None:
+      raise unittest.SkipTest('Tests only for MPI.')
```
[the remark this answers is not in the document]
nvcastet: It is a test file, so we skip Open MPI initialization test on non-MPI environments.

jax/_src/clusters/ompi_cluster.py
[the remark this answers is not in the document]
nvcastet: The jobid is always a multiple of 2^12 (which would lead to always the same port), truncate it first to 5 digits before taking the modulus

nvlcambier (approved):

jax/_src/clusters/ompi_cluster.py
```diff
@@ -0,0 +1,59 @@
...
+  @classmethod
+  def get_coordinator_address(cls) -> str:
+    # Examples of orte_uri:
+    # 1531576320.0;tcp://10.96.0.1,10.148.0.1,10.108.0.1:34911
+    # 1314521088.0;tcp6://[fe80::b9b:ac5d:9cf0:b858,2620:10d:c083:150e::3000:2]:43370
+    orte_uri = os.environ[_ORTE_URI]
+    job_id = orte_uri.split('.', maxsplit=1)[0]
+    # The jobid is always a multiple of 2^12,
+    # truncate it first to 5 digits before taking the modulus
+    job_id = job_id[-5:]
```
skye, line 40: This is the last 5 digits now, but the comment still says the first five digits, which I think makes more sense? I would also add "to ensure unique job IDs" to the comment just to really spell it out.

Now that I think about it, would it be more direct to do something like job_id = int(job_id) // 2**12, and take the modulus of that?

skye (changes requested):
Sorry one more comment about the job_id :P

jax/_src/clusters/ompi_cluster.py
nvcastet: Good idea Skye! Yes it should work.

skye (approved):
Nice!

Issue #13941, opened 2023-01-10
`jax.grad(jnp.log)` returns non-nan for out-of-domain inputs


pierreglaser:
### Description

Hi, and first of all, thanks all the amazing work on `jax` and its ecosystem!

Title says it all: MVCE below.

```python
In [1]: import jax.numpy as jnp
   ...: from jax import grad

In [2]: jnp.log(-1.)  # nans expected
Out[2]: Array(nan, dtype=float32, weak_type=True)

In [3]: grad(jnp.log)(-1.)  # nans ALSO expected!
Out[3]: Array(-1., dtype=float32, weak_type=True)
```

The above behavior contradicts the definition of a derivative of a function $f$, defined only on the domain of $f$ ( $(0, +\infty)$ in that case). I believe the [jvp rule for `jnp.log`](https://github.com/google/jax/blob/b5be92e449b2a0f61fe4ee301cc1dc1659d9d7f2/jax/_src/lax/lax.py#L1764), currently defined as 

```python
lambda g, x: div(g, x))
```
should  perform a domain-check before returning.

### What jax/jaxlib version are you using?

jax v0.4.1, jaxlib v0.4.1

### Additional system info

Python 3.9.15, CPU version, MacOS Big Sur

jakevdp:
Thanks for the report! I'll take a look

jakevdp:
Sorry this took so long to respond, but I'm coming back to this because we recently discussed the same issue in another context. The short answer is: I think this is working as expected.

We've not spelled out this contract anywhere that I'm aware of, but roughly the expectation of autodiff should be "we produce an expression which returns the correct gradient where the function is defined. Where the function is not defined, the expression has no guarantees".

A benefit of this definition is that it leaves validation up to the user: i.e. rather than `lax.log` always producing a `cond` in its gradient, it instead produces a simple expression (`reciporical(x)`) and leaves it up to the user whether to add additional checks. In practice, we think this is the best balance between strict correctness and performance.

PR #13945, opened 2023-01-10
lax.log: return NaN for out-of-bound gradient

jakevdp:
Proposed fix for #13941

This causes a number of test failures; it's not clear to me at the moment where they come from.


jakevdp:
Closing, as I don't think this is the right approach here.

PR #13961, opened 2023-01-11
[call_tf] Temporary fix for CallTfEffect being visible after lowering.

gnecula:
The CallTfEffect was added recently as an internal workaround for DCE removing arguments to call_tf. This may have been a convenient available mechanism, but there is no real "effect" here, and we should not hijack the effects mechanism for dealing with DCE.

This PR does not really solve the problem, but it removes the CallTfEffect from the externally-visible set of effects, so that users of the AOT APIs do not have to worry about the computation having real effects.

Relates to #13963


yashk2810 (approved):

sharadmv:
> there is no real "effect" here

Couldn't we have an effect? Users can write whatever side-effecting code they want in a TF function that could write to disk or something. Right now we are defensively marking it as effectful to avoid DCE.

PR #13977, opened 2023-01-12
Add `pjrt_c_api_unimplemented` pytest marker to skip unsupported tests.

skye:
Also adds `test_util.pytest_mark_if_available` helper function.

### DIFF
--- a/tests/clear_backends_test.py
+++ b/tests/clear_backends_test.py
@@ -23,6 +23,7 @@ from jax._src.lib import xla_bridge as xb
 config.parse_flags_with_absl()
 
 
+@jtu.pytest_mark_if_available('pjrt_c_api_unimplemented')  # crashes runtime
 class ClearBackendsTest(jtu.JaxTestCase):
 
   def test_clear_backends(self):
--- a/tests/debugger_test.py
+++ b/tests/debugger_test.py
@@ -55,6 +55,7 @@ def tearDownModule():
 
 foo = 2
 
+@jtu.pytest_mark_if_available('pjrt_c_api_unimplemented')  # host callback
 class CliDebuggerTest(jtu.JaxTestCase):
 
   def test_debugger_eof(self):
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -332,6 +332,17 @@ def skip_on_flag(flag_name, skip_value):
   return skip
 
 
+def pytest_mark_if_available(marker: str):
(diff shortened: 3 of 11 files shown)


tests/clear_backends_test.py
```diff
@@ -22,6 +24,10 @@
 
 config.parse_flags_with_absl()
 
+with contextlib.suppress(ImportError):
+  import pytest
+  pytestmark = pytest.mark.pjrt_c_api_unimplemented  # crashes runtime
```
jakevdp, line 29: What if you make the new `pytest_mark_if_available` function into a class with a `__call__` and `__getattr__` defined, so that here and elsewhere you could just write
```python
pytestmark = jtu.pytest_mark_if_available.pjrt_c_api_unimplemented  # crashes runtime
```

jakevdp (commented):
Looks good - I assume you're doing this so you can run things like
```
$ pytest tests/ -m "not pjrt_c_api_unimplemented"
```
If so, should we document this pattern in the developer guide?

tests/debugger_test.py
```diff
@@ -31,6 +33,11 @@
 
 config.parse_flags_with_absl()
 
+with contextlib.suppress(ImportError):
+  import pytest
+  pytestmark = pytest.mark.pjrt_c_api_unimplemented  # host callback
```
jakevdp, line 38: Other question: does just declaring this in the sope actually mark tests within the file? I've always seen `pytest.mark` used as a decorator for a test function or test class.

tests/clear_backends_test.py
jakevdp: Ah I see the global variable works: https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

This kind of magic is one of the downsides of pytest IMO... how would you feel about using it as a class decorator here instead?

skye: Switched to using class decorator, see other comment.

tests/debugger_test.py
jakevdp: Looks great!

skye:
> Looks good - I assume you're doing this so you can run things like
> $ pytest tests/ -m "not pjrt_c_api_unimplemented"
> If so, should we document this pattern in the developer guide?

Probably :) I filed https://github.com/google/jax/issues/13989 for now, since this is just for with the PJRT C API enabled, which isn't on by default yet and I don't think worth documenting until it is.

jakevdp (approved):

PR #13991, opened 2023-01-12
Migrate pytestmark usage to new `@jtu.pytest_mark_if_available` decorator.

skye:
See discussion in https://github.com/google/jax/pull/13977. Marking entire modules is magical and verbose, plus less precise than marking individual classes or tests.

I wasn't super careful on which classes to mark, and erred on the side of marking too many classes (in line with the previous behavior). It's possible some test classes don't actually benefit from multiple accelerators.

### DIFF
--- a/tests/multiprocess_gpu_test.py
+++ b/tests/multiprocess_gpu_test.py
@@ -41,11 +41,6 @@ try:
 except ImportError:
   portpicker = None
 
-try:
-  import pytest
-except ImportError:
-  pytest = None
-
 config.parse_flags_with_absl()
 
 @unittest.skipIf(not portpicker, "Test requires portpicker")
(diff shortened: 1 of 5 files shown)


tests/multiprocess_gpu_test.py
```diff
@@ -230,12 +225,9 @@ def test_gpu_ompi_distributed_initialize(self):
 @unittest.skipIf(
     os.environ.get("SLURM_JOB_NUM_NODES", None) != "2",
     "Slurm environment with at least two nodes needed!")
-@unittest.skipIf(not pytest, "Test requires pytest markers")
+@jtu.pytest_mark_if_available('SlurmMultiNodeGpuTest')
```
skye, line 228: @sudhakarsingh27 FYI

jakevdp (approved):
Looks good!

Issue #11208, opened 2022-06-22
Allow symbolic zeros in custom derivatives


patrick-kidger:
Feature request: to optionally not materialise symbolic zeros for (co)tangents to `custom_{jvp,vjp}`. Instead, pass the symbolic zero as-is and allow the user code to special-case their behaviour on which arguments have symbolic zeros.

When using `custom_{vjp,jvp}` in library code, it's common to accept a PyTree of arbitrary user-specified arguments, with respect to which derivatives are computed (my main use case is through a nonlinear solve via the implicit function theorem). Typically only some of those user-specified arguments are actually being differentiated, so the derivatives computed for the others are just a (potentially large) waste of computation.

This would also allow a solution to `jax.closure_convert` having to inefficiently hoist out all floating-point arrays: the custom derivative code will be able to determine which of these actually need a derivative and can ignore the others.

I'm imagining something like `custom_jvp(..., symbolic_zeros=True)`.

froystig:
see also #5913

patrick-kidger:
Addendum to this: the ability to return symbolic zeros from `custom_{jvp,vjp}`, for auxiliary information.

typedfemale:
I would just like to voice my support for this, since it commonly comes up when trying to write novel optimizers 😇

patrick-kidger:
Closing as this now exists! `jax.custom_jvp.defjvp(..., symbolic_zeros=True)`, and likewise for `custom_vjp`.

NeilGirdhar:
Still needs documentation, right?  I guess when `symbolic_zeros` is true, tangents and cotangents have type `custom_derivatives.SymbolicZero` when they are both static and constant?

froystig:
We had populated the `custom_jvp` and `custom_vjp` docstrings but the documentation wasn't rendering! That ought to be fixed by #16639.

Between that, #14570, and #15221, I believe that we can indeed close this.

NeilGirdhar:
@froystig Awesome, looks like a very elegant design. What is the type of the "object with two attributes instead: ``value`` and ``perturbed``", also called "two-field record objects"?

froystig:
Thanks for pointing that out. It's a `jax.custom_derivatives.CustomVJPPrimal`. #16693 should make that clearer.

PR #13993, opened 2023-01-12
Added symbolic zeros as inputs to `custom_jvp`

patrick-kidger:
Added a `jax.custom_jvp(..., symbolic_zeros=True)` flag, which causes symbolic zero tangents to be passed as a `SymbolicZero` object.

Design choices already discussed with @froystig:

1. We decided to make this a new `SymbolicZero` object since (a) `ad_util.Zero` is so far only used inside AD rules, and we didn't want to promulgate that outside of its current context, and (b) `None` doesn't carry any aval information.
2. In time we imagined that `ad_util.Zero` could potentially be removed in favour of the new `SymbolicZero`, which can be treated as a concrete array type. Special symbolic-zero-handling would move from being a thing done in AD rules to a thing done in impl/lowering rules.
3. This behaviour is gated behind a `symbolic_zeros` flag to avoid breaking everyone who depends on the current behaviour.

Design choices not discussed with @froystig:

1. The symbolic zero gets its own aval (not just `ShapedArray`), since in practice things may be wrapped in tracers (e.g. batch tracers), so the appropriate way to check for a symbolic zero in user-code is currently `isinstance(tangent.aval, ad.ZeroShapedArray)`. Hypothetically we could imagine avoiding this new aval by having every `FooTrace.process_custom_jvp_call` special-case a symbolic zero, but that strikes me as very intrusive. (E.g. `BatchTrace` would remove a dimension from `SymbolicZero` instead of wrapping it into a `BatchTracer`; `JVPTrace` would instead pass it through unchanged, etc.)

Questions for @froystig and/or @mattjj:

1. Are we happy with the above design choices?
2. Which file do we want to put `SymbolicZero` in?
3. What API do we want for downstream users to detect symbolic zeros?
    - The current `isinstance(tangent.aval, ad.ZeroShapedArray)`?
    - An `is_symbolic_zero` function?
    - We could also drop all tracer information and substitute `Traced<ZeroShapedArray>` for `SymbolicZero` immediately prior to passing to the user, to allow for a relatively-natural `isinstance(tangent, SymbolicZero)`.

[To-do for me: also add an equivalent `ShapedArrayRef -> ZeroShapedArrayRef` mapping. But I wanted to get this first version out to get some feedback first.]

---

Partially addresses #11208.

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -910,10 +910,11 @@ def custom_jvp_call_rule(in_err, enabled_errors, *in_vals, num_consts,
   @lu.wrap_init
   def jvp(*xs):
     # TODO(lenamartens, sharadmv): why not checkify here?
-    jvp_jaxpr, jvp_consts = jvp_jaxpr_thunk()
     n, ragged = divmod(len(xs), 2)
     assert not ragged
     primals, tangents = xs[num_consts:n], xs[n+num_consts:]
+    avals = tuple(map(core.get_aval, primals) + map(core.get_aval, tangents))
+    jvp_jaxpr, jvp_consts = jvp_jaxpr_thunk(*avals)
     return core.eval_jaxpr(jvp_jaxpr, jvp_consts, *primals, *tangents)
 
   jvp, jvp_out_tree = flatten_fun_output(jvp)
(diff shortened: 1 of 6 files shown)


patrick-kidger:
Closing in favour of #14570.

PR #14008, opened 2023-01-13
Don't set TPU topology env vars in cloud_tpu_init.py

skye:
This used to be necessary. However, now these are automatically set in libtpu. Beyond being redundant, the Python logic needs to be updated to avoid getting KeyErrors on new topologies and TPU versions, so better to remove it.

This also moves `get_metadata` to cloud_tpu_cluster.py since it's only used in that file now.


skye:
Just tested that jax.device_count() returns the right thing on a v2-32, so I think that means I didn't break slice initialization.

suncats (approved):

PR #14011, opened 2023-01-14
Eagerly register backend factories for PJRT C API plugins.

stellaraccident:
The prior mechanism was triggering the xla_client (in jaxlib) to load the plugins based on the environment variable, but there was nothing making them known to the JAX-side xla_bridge. Further, they were all being loaded regardless of what plugin was being requested.

This reworks the mechanism to eagerly register backend factories with the JAX-side xla_bridge so that they are availble when requested (i.e. as part of JAX_PLATFORMS env var or equivalent).

There is an opaque `TODO(b/237099479): Move to xla_bridge.py when ready.` in xla_client.py that can't be seen from the OSS side, so not sure what the intent here is. As near as I can tell, this never worked and is a temporary measure while we work out a better mechanism.

### DIFF
--- a/jax/_src/lib/xla_bridge.py
+++ b/jax/_src/lib/xla_bridge.py
@@ -257,6 +257,22 @@ if hasattr(xla_client, "make_plugin_device_client"):
 if iree is not None:
   register_backend_factory("iree", iree.iree_client_factory, priority=-100)
 
+# Register PJRT C API plugins based on environment variables.
+def _initialize_plugin_backends():
+  def make_factory(name, path):
+    def factory():
+      xla_client._xla.load_pjrt_plugin(name, path)
+      return xla_client._xla.get_c_api_client(name)
+    return factory
+
+  pjrt_plugins = xla_client._get_pjrt_plugin_names_and_library_paths()
+  for plugin_name, library_path in pjrt_plugins.items():
+    register_backend_factory(
+      plugin_name, make_factory(plugin_name, library_path))
+
+if hasattr(xla_client, "_get_pjrt_plugin_names_and_library_paths"):
+  _initialize_plugin_backends()
+
 _platform_aliases = {
   "cuda": "gpu",
   "rocm": "gpu",
@@ -328,8 +344,7 @@ def backends():
           (platform, priority) for platform, (_, priority)
           in _backend_factories.items())
     default_priority = -1000
-    if hasattr(xla_client, "maybe_load_pjrt_plugins"):
-      xla_client.maybe_load_pjrt_plugins()
+
     for platform, priority in platforms_and_priorites:
       try:
         backend = _init_backend(platform)


stellaraccident:
@skye

jax/_src/lib/xla_bridge.py
```diff
@@ -257,6 +257,22 @@ def make_gpu_client(*, platform_name, visible_devices_flag):
 if iree is not None:
   register_backend_factory("iree", iree.iree_client_factory, priority=-100)
 
+# Register PJRT C API plugins based on environment variables.
+def _initialize_plugin_backends():
+  def make_factory(name, path):
+    def factory():
+      xla_client._xla.load_pjrt_plugin(name, path)
+      return xla_client._xla.get_c_api_client(name)
```
jyingl3, line 265: @skye Do we want to expose xla_client._xla in xla_bridge?

jax/_src/lib/xla_bridge.py
```diff
@@ -257,6 +257,22 @@ def make_gpu_client(*, platform_name, visible_devices_flag):
...
+    return factory
+
+  pjrt_plugins = xla_client._get_pjrt_plugin_names_and_library_paths()
+  for plugin_name, library_path in pjrt_plugins.items():
+    register_backend_factory(
+      plugin_name, make_factory(plugin_name, library_path))
+
+if hasattr(xla_client, "_get_pjrt_plugin_names_and_library_paths"):
```
jyingl3, line 273: @skye I know this change is a short term work around. Do we want to expose _get_pjrt_plugin_names_and_library_paths (i.e. remove "_" and add to xla_client.pyi?

jax/_src/lib/xla_bridge.py
```diff
@@ -257,6 +257,22 @@ def make_gpu_client(*, platform_name, visible_devices_flag):
...
+  def make_factory(name, path):
+    def factory():
+      xla_client._xla.load_pjrt_plugin(name, path)
+      return xla_client._xla.get_c_api_client(name)
+    return factory
+
+  pjrt_plugins = xla_client._get_pjrt_plugin_names_and_library_paths()
+  for plugin_name, library_path in pjrt_plugins.items():
+    register_backend_factory(
+      plugin_name, make_factory(plugin_name, library_path))
```
jyingl3, line 271: Shall we set a high priority for plugin factories, e.g. priority=400?

jax/_src/lib/xla_bridge.py
```diff
@@ -257,6 +257,22 @@ def make_gpu_client(*, platform_name, visible_devices_flag):
 if iree is not None:
   register_backend_factory("iree", iree.iree_client_factory, priority=-100)
 
+# Register PJRT C API plugins based on environment variables.
+def _initialize_plugin_backends():
```
jyingl3, line 261: nit: maybe rename to _register_plugin_backend_factories?

jyingl3 (commented):
Thank you so much for making this change Stella!

jax/_src/lib/xla_bridge.py
skye: Ideally no, but it's not the end of the world. The right way to do this is to update [xla_client.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/python/xla_client.py) in the TF repo to expose the methods you need directly (like [this](https://github.com/tensorflow/tensorflow/blob/6d5d88041f49e866b1dd18b6f532107e4b84e650/tensorflow/compiler/xla/python/xla_client.py#L41-L42)). @stellaraccident I'd help you get a TF PR in if you wanna do that, but this is also fine for now and we can fix it later (might be easiest to do one big pass exposing everything we need after this code stops churning).

I suspect this does need an xla_client version guard though. Basically, this code has to be forward-compatible with older jaxlib versions (and xla_client is shipped with jaxlib).

Example version guard: https://github.com/google/jax/blob/e3c2602ed5174bcfd79fe3b25cf9009cdcb48bfe/jax/experimental/host_callback.py#L993-L997

Version number comes from here: https://github.com/tensorflow/tensorflow/blob/6d5d88041f49e866b1dd18b6f532107e4b84e650/tensorflow/compiler/xla/python/xla_client.py#L46

You can drop the guard once the minimum jaxlib version exports a high enough xla_client version:
https://github.com/google/jax/blob/e3c2602ed5174bcfd79fe3b25cf9009cdcb48bfe/jax/version.py#L19

So for `get_c_api_client`, blame tells me it was introduced in https://github.com/tensorflow/tensorflow/commit/77a1e8e3fc3637f7fa29d9cd8d8d33dcb707b965, and the xla_client version number was [110](https://github.com/tensorflow/tensorflow/blob/77a1e8e3fc3637f7fa29d9cd8d8d33dcb707b965/tensorflow/compiler/xla/python/xla_client.py#L46). Or you can just use the current version number of 118 if you don't wanna check. I'm gonna cut a new jaxlib soon, so this should be publicly available then.

jyingl3:
@jzhoulon FYI. After this change, there is no need to manually register a plugin backend and you can run your plugin by setting environment variables (no code changes needed).

jzhoulon:
> 
@jyingl3 Thanks for the notification. I saw the RFC, it mentioned there is a way to load the plugin through [naming convention of the Python module](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-naming-convention), in my understanding, it will find the package and may be can load the library called pjrt_xx.so in that module installed path(such as site-package/pjrt_xxx), then the loading mechanism will not require user specifiy the environment, please correct my if I was wrong:) if it is, will jax support this mechanism? I think letting user always specify the library path before running is not convenient enough . Thanks

stellaraccident:
Confirmed that the upstream changes supercede this patch.

PR #14030, opened 2023-01-17
[shape_poly] Improve error message from vmap axis size inconsistency

gnecula:
vmap tries hard to give nice error messages when the mapped axes for different arguments have different sizes, but the code to compute the error message can run into InconsistentDimensionOperation in presence of dimension polynomials. Ensure that the comparisons are done symbolically.


gnecula:
@marcvanzee This PR improves error messages for shape polymorphism on the Flax examples you are working on

marcvanzee (approved):
Thanks!

PR #14047, opened 2023-01-18
DOC: add GPU & TPU installation tabs on index page

jakevdp:
As discussed this morning.

202036b0-174f-4af4-b167-e96923e7336c

Live preview: https://jax--14047.org.readthedocs.build/en/14047/#installation

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -60,14 +60,28 @@ JAX is Autograd_ and XLA_, brought together for high-performance numerical compu
 
 Installation
 ------------
+.. tab-set::
 
-.. code-block:: bash
+    .. tab-item:: CPU
 
-   pip install "jax[cpu]"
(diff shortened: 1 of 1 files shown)


zhangqiaorjc (approved):
Super nice! Thanks Jake!

yashk2810:
Can we add a bit more customization?

CPU/GPU (2 cuda versions)/TPU/MAC?

jakevdp:
Sure, but everything is a balance between having too little information up front and having too much information up front. I tried adding extra info here and it looked busy to me. But feel free to give it a try if you disagree!

yashk2810:
Adding one more cuda version and mac tab is not too much information IMO. It would be sad if users can't find that information too :)

I think install page should be about full information. It doesn't matter if we have a couple more options IMO.

PR #14049, opened 2023-01-18
DOC: make grid placement responsive for smaller screens

jakevdp:
This change makes it so that the grid items on the main page are more responsive to page width, rather than always stacking three horizontal.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,30 +4,30 @@ JAX: High-Performance Array Computing
 JAX is Autograd_ and XLA_, brought together for high-performance numerical computing.
 
 .. grid:: 3
+   :margin: 0
+   :padding: 0
+   :gutter: 0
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -5,29 +5,26 @@ JAX is Autograd_ and XLA_, brought together for high-performance numerical compu
 
 .. grid:: 3
 
-   .. grid-item::
+   .. grid-item-card:: Familiar API
+      :columns: 12 6 6 4
```
jakevdp, line 9: The meaning of these numbers is described here: https://sphinx-design.readthedocs.io/en/latest/grids.html#introduction

Namely, on extra small screens, each card will occupy the full screen width. On medium and large screens, each card will occupy half the screen width. On extra large screens, each card will occupy 1/3 of the screen width.

PR #14066, opened 2023-01-18
Update lax_control_flow_test.py

rmlarsen:
Fix brittle scan test. Adding tan(randn) is numerically brittle because evaluating tan() near its singularities is ill-conditioned.


jakevdp (approved):
Looks reasonable, thanks!

PR #14083, opened 2023-01-19
Remove more uses of tan() in reduction tests.

rmlarsen:
This is to avoid subtly brittle tests. Tan() is an ill-conditioned function to evaluate near it's singularities.


mattjj:
Thanks, Rasmus!

mattjj (approved):

PR #14085, opened 2023-01-20
CI: adjust permissions for upstream-nightly build

jakevdp:
The report job failed with a permissions error: https://github.com/google/jax/actions/runs/3958254462/jobs/6779789502

This should fix the issue.

For what it's worth, the root cause is a failure in numpy nightly, and is tracked here: numpy/numpy/issues/23033


pschuh (approved):

jakevdp:
Successfully created issues (see links above) I need to revert some of the temporary changes.

PR #14089, opened 2023-01-20
Clarify the docstring for vjp.

hawkinsp:


mattjj (approved):
Thanks for doing this. It's a small but key improvement!

PR #14090, opened 2023-01-20
CI: update deprecated uses of set-output

jakevdp:
This deprecation has been causing warnings in our CI build; for details see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


pschuh (approved):

jakevdp:
The nightly failure is a known issue; see #14085. Looks like the fix went into numpy-dev this morning.

PR #14096, opened 2023-01-20
[sparse] refactor tests to improve runtime

jakevdp:
This effectively combines the `_CheckAgainstDense` and `_CompileAndCheckSparse` test utilities. Previously each test would do two jit and two non-jit calls of each tested function; with this change it's down to one jit and one non-jit call.

We lose some coverage in the area of recompilation checks (this is why `_CompileAndCheckSparse` would call the jit-compiled function twice) but given that test timeouts have become an issue, I think that's an OK tradeoff.


skye (approved):

jakevdp:
So, it turns out that this doesn't have an appreciable affect on the runtime of `sparse_test.py`. Perhaps the slow tests are dominated by JIT compile time, so that doing repeated operations doesn't make them appreciably slower?

Still, I like this change because it makes the test utilities cleaner, so I think we should merge it.

PR #14097, opened 2023-01-20
DOC: Adding @custom_partitioning to jax.experimental API doc

nvlcambier:
This PR adds documentation (in the form of an example directly in the docstring) for the experimental @custom_partitioning.
It also does some 100% cosmetic fixes to the existing docstring.

### DIFF
--- a/jax/experimental/custom_partitioning.py
+++ b/jax/experimental/custom_partitioning.py
@@ -186,7 +186,11 @@ class custom_partitioning:
     from jax.sharding import PartitionSpec as P
     from jax.experimental.maps import Mesh
     from jax.numpy.fft import fft
+    import regex as re
     import numpy as np
+
+    # Pattern to detect all-gather or dynamic-slice in the generated HLO
+    _PATTERN = '(dynamic-slice|all-gather)'
 
     # For an N-D input, keeps sharding along the first N-1 dimensions
     # but replicate along the last dimension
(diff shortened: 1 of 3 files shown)


jax/experimental/custom_partitioning.py
```diff
@@ -130,39 +130,190 @@ def _default_propagate_user_shardings(sharding, shape):
...
+    [    7.217285   +0.j     -3012.4937  +4287.635j   -405.83594 +3042.984j
+    ...  1422.4502  +7271.4297j  -405.84033 -3042.983j
+    -3012.4963  -4287.6343j]
+
+  In this case, the HLO of ``my_fft`` does show an all-gather and dynamic-slice, since the last dimension
+  is the dimension along which FFTs are calculated.
+
+  .. code-block::
+
+    HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(c64[16777216]{0})->c64[16777216]{0}}
```
yashk2810, line 302: Can we remove the entire HLO print here since it can get stale?

You can just do an assert in code to test for dynamic-slice, etc

nvlcambier: Can you point me to an example or the right API to use to do that?

pschuh: I think yash is suggesting just to do re.match on the output_text for dynamic-slice.

nvlcambier: Pushed a new version ; please have a look, thanks!

pschuh (approved):

jax/experimental/custom_partitioning.py
pschuh: Looks like there are still some presubmits failing.

nvlcambier: Can https://github.com/google/jax/actions/runs/3991188021/jobs/6845708206 be skipped for a particular file/section/... somehow?

Otherwise, I guess I need to put everything in a single `code-block`, which is fine, but seems a little less-nice for this specific case.

I'll fix the other two.

pschuh: I think the problem is not that they're two codeblocks its just that pytest is picking up on the

>>> code()
output

form but ignoring the code-block form.

doctests seem to carry over from one call to the next if you use the >>> form for all of them. I dunno how the docs look in that case.

nvlcambier: Fixed (WARed, really) using code-blocks. I thinks that should do it.

nvlcambier: All precommit checks are passing now

pschuh (approved):

PR #14102, opened 2023-01-20
Raise an error when attempting to mutate Jaxpr objects

jakevdp:
@sharadmv has recently run into an error that came down to a cached jaxpr being mutated in place. This is an attempt to guard against that kind of footgun by making jaxpr attributes read-only.

A more heavyweight solution might be to make `Jaxpr` and `ClosedJaxpr` either a `NamedTuple` or a frozen `dataclass`, but this is a quick fix that we can use to find any potential offending code using a global presubmit.


sharadmv (approved):
Thanks! This will avoid a lot of footguns in the future

PR #14106, opened 2023-01-21
[shape_poly] Generalize binary operations with symbolic dimensions.

gnecula:
Previously binary operations involving symbolic dimensions would work only when the other operand is convertible to a symbolic dimension, e.g., an integer. This resulted in errors when trying "x.shape[0] * 3.5" and the recourse was to ask the user to add an explicit "jnp.array(x.shape[0])".

Now we allow binary operations with any operand and the "jnp.array" is added automatically if the other operand is not an integer or a symbolic dimension. This means that instead of an error they may be an error downstream if one tries to use the result as a dimension. There is one known case where JAX works with static shapes and with the previous behavior, but will fail now. When you operate on `np.ndarray` and symbolic dimension, previously this was kept as a `np.ndarray` but not it is turned into a JAX array. The following program will now fail if `x.shape[0]` is a symbolic dimension.:

`jnp.ones(np.arange(5) * x.shape[0])`

Instead you should write

`jnp.ones([i * x.shape[0] for i in range(5)])`

We also took the opportunity to make some improvements to the shape polymorphism documentation.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1197,13 +1197,13 @@ def _split(op: str, ary: ArrayLike, indices_or_sections: Union[int, ArrayLike],
                                                  f"in jax.numpy.{op} argument 1")
     part_size, r = _divmod(size, indices_or_sections)  # type: ignore[misc]
     if r == 0:
-      split_indices = np.arange(indices_or_sections + 1,
-                                dtype=np.int64) * part_size
+      split_indices = [np.int64(i) * part_size  # type: ignore
+                       for i in range(indices_or_sections + 1)]  # type: ignore
     elif op == "array_split":
-      split_indices = np.concatenate(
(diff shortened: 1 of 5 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1197,13 +1197,13 @@ def _split(op: str, ary: ArrayLike, indices_or_sections: Union[int, ArrayLike],
                                                  f"in jax.numpy.{op} argument 1")
     part_size, r = _divmod(size, indices_or_sections)  # type: ignore[misc]
     if r == 0:
-      split_indices = np.arange(indices_or_sections + 1,
-                                dtype=np.int64) * part_size
+      split_indices = [np.int64(i) * part_size  # type: ignore
```
marcvanzee, line 1200: Did you have to change this because it uses numpy, which no longer supported for symbolic dimensions?

marcvanzee (approved):
Looks good!

PR #14137, opened 2023-01-24
Checkify: Validate format arguments to make sure they're arrays

LenaMartens:
This disallows `checkify.check("{}", 1.)`, but you probably shouldn't be doing that anyway (you can embed static info in the msg directly)


sharadmv (commented):
~What about pytrees?~ NVM

LGTM!

sharadmv (approved):

PR #14158, opened 2023-01-25
DOC: organize content in advanced guide

jakevdp:
cc @8bitmp3, @canyon289  

Preview: https://jax--14158.org.readthedocs.build/en/14158/advanced_guide.html

### DIFF
--- a/docs/advanced_guide.rst
+++ b/docs/advanced_guide.rst
@@ -2,20 +2,41 @@
 
 Advanced Tutorials
 ==================
+This section contains examples and tutorials on more advanced topics, such as Multi Core computation, Custom operations, and more in depth applications
 
 .. toctree::
+   :caption: Examples
+   :maxdepth: 1
+
+   notebooks/neural_network_with_tfds_data
+   notebooks/Neural_Network_and_Data_Loading
+   notebooks/vmapped_log_probs
+
+.. toctree::
+   :caption: Parallel Computation
+   :maxdepth: 1
+
+   multi_process
+   notebooks/Distributed_arrays_and_automatic_parallelization
+   notebooks/xmap_tutorial
+
+.. toctree::
+   :caption: Automatic Differentiation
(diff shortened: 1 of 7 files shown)


docs/advanced_guide.rst
```diff
@@ -2,20 +2,41 @@
 
 Advanced Tutorials
 ==================
+This section contains examples and tutorials on more advanced topics
```
canyon289, line 5: A short explanation of what makes these advanced would be nice, even for beginner users. That way they have a sense what advanced usage means and if they even need to be advanced

Here's a suggestion

```suggestion
This section contains examples and tutorials on more advanced topics, such as Multi Core computation, Custom operations, and more in depth applications
```

docs/advanced_guide.rst
```diff
@@ -2,20 +2,41 @@
...
 .. toctree::
+   :caption: Examples
+   :maxdepth: 1
+
+   notebooks/neural_network_with_tfds_data
+   notebooks/Neural_Network_and_Data_Loading
+   notebooks/vmapped_log_probs
+
+.. toctree::
+   :caption: Multi-core and Multi-pod Computation
```
yashk2810, line 16: Is POD an external name? i.e. something we use in cloud TPU docs?

jakevdp: I've seen it used elsewhere (e.g. https://blogs.nvidia.com/blog/2021/03/05/what-is-a-cluster-pod/). Do you think multi-cluster or something like that would be more clear?

jakevdp: Maybe "parallel and distributed computation"?

yashk2810: I think we should ask Skye or James

jakevdp: James recommended "Parallel Computation"

yashk2810 (approved):

PR #14169, opened 2023-01-26
Removed deprecated polar_unitary as per phawkins comment.

tttc3:
Noticed a comment in `jax/_src/scipy/linalg.py` to remove `polar_unitary` after 2022/08/11. 
This pull just removes that function and the left over references.


jakevdp (changes requested):
Looks good, thanks! We also should mention this removal in the CHANGELOG.

jakevdp (approved):
Looks good, thanks!

PR #14170, opened 2023-01-26
DOC: clarify purpose of the jax-array doc

jakevdp:
I noticed that the doc itself didn't mention the context of the work.

### DIFF
--- a/docs/jax_array_migration.md
+++ b/docs/jax_array_migration.md
@@ -5,6 +5,10 @@
 
 ## TL;DR
 
+JAX switched its default array implementation to the new `jax.Array` as of version 0.4.1.
+This guide explains the reasoning behind this, the impact it might have on your code,
+and how to (temporarily) switch back to the old behavior.
+
 ### What’s going on?
 
 `jax.Array` is a unified array type that subsumes `DeviceArray`, `ShardedDeviceArray`,


docs/jax_array_migration.md
```diff
@@ -5,6 +5,11 @@
 
 ## TL;DR
 
+JAX switched its default array implementation from `DeviceArray` to
```
yashk2810, line 8: Isn't this covered in What's going on below?

jakevdp: That is mentioned below, but it doesn't mention what JAX version this is relevant for, which makes it unclear to the reader if this doc is relevant to them.

jakevdp: Also, the TLDR is currently four paragraphs 😁

yashk2810: Okay, the version should be 0.4.1.

Also mention SDA and GDA (since it replaces all 3)

jakevdp: That's mentioned below, it doesn't seem necessary to mention it here?

jakevdp: What's not mentioned below is (1) the default array type is now `jax.Array` instead of `DeviceArray`, and (2) the version at which this change happened was 0.4.1. I think those are the relevant pieces to add here.

yashk2810: I think you should include all the types because that is the correct information. Jax.Array replaces all 3.

If it is mentioned below, then don't mention it here :) 

Or you can just say: JAX switched its default array implementation type to jax.Array as of version 0.4.1

This skips DA, SDA and GDA as they are mentioned below.

jakevdp: I'm mentioning something different here: that the default array type is now `jax.Array`, not `DeviceArray`. The reason I think this is important is because most users probably don't know that the default array type was `DeviceArray`, so mentioning that `jax.Array` replaces `DeviceArray`, `GlobalDeviceArray`, and `ShardedDeviceArray` does not make it clear why this is relevant to them.

Does that make sense?

yashk2810: How do you decide which was the default Array type before?

DA was the output of jit by default. But SDA was the default type from pmap. So just mentioning DA is not correct IMO (given that we had 2 default Array types).

yashk2810: And jit and pmap are both core APIs of JAX.

jakevdp: Well, `DeviceArray` was the default that 99% of users encountered. I can skip saying what the previous default was it if you think mentioning it is unclear.

yashk2810: How did you get to the 99% number? pmap is also used pretty heavily I think that those users don't see DAs.

I think that the start of the document should convey the whole picture in short rather than give partial information and then give the full picture below. Maybe we can come up with a better sentence to start the doc (basically resolve this discussion?)

JAX introduces a new jax.Array type which replaces all the other array types in JAX as of version 0.4.1. <your next sentences> (or some variant of this?)

jakevdp: I answer a lot of user questions on github and stackoverflow, and users only rarely ask about `pmap`, so I suspect among people who are finding their way to this doc page, few use pmap.

But it seems you feel very strongly about not saying `DeviceArray` here. I can remove it: the more important thing is that we have *some* framing of this document for people who land on it.

jakevdp:
Updated, PTAL

docs/jax_array_migration.md
yashk2810: Yeah, +1 for framing!

My only point is that it is not just DA that we are replacing and might give a different idea to someone who does use pmap/pjit/xmap.

yashk2810 (approved):

PR #14189, opened 2023-01-27
remove rules translating opaquely dtyped avals directly to MLIR types

froystig:
We have a rule for grounding opaquely dtyped avals into a list of basic ("physical") avals, so we can just translate as usual from there. This should help maintain consistency, and reduces code a bit.

I added some comments along the way.


mattjj (approved):
🚀

PR #14200, opened 2023-01-29
[shape_poly] Fix the hashing and equality of symbolic dimensions

gnecula:
In order to ensure soundness in presence of symbolic dimensions, JAX raises `InconclusiveDimensionOperation` for equality comparisons that cannot be resolved to the same boolean value for all possible valuations of the dimension variables, e.g., `a == 1`. Since the equality of symbolic dimensions is not total, they should not be hashable. Before this PR they were hashable and this resulted in non-deterministic failures when symbolic dimensions are used as hash keys (JAX often memoizes functions with ShapedArray and Jaxpr arguments, and those can contain symbolic dimensions).

This PR creates two types of symbolic expressions: `_DimExpr` is the main one and now it is not hashable anymore. This is the type that is safe to use in user programs. The only way a user program can get their hands on a symbolic expression is by reading the `shape` attribute of a value. Hence, we ensure that `x.shape` returns a tuple of `_DimExpr`.

The other type of symbolic expressions is `_DimExprHashable` which is a subclass of `DimExpr` defining `__hash__` and defining a total `__eq__` method (returns `False` when the symbolic dimensions are not guaranteed to be equal). We ensure that whenever we use a symbolic dimension in a `ShapedArray` or in a parameter of a `JaxprEqn`, we use `_DimExprHashable`. 

This change can break user programs that assume that `x.shape` is hashable.


gnecula:
We landed this already in #14290.

PR #14206, opened 2023-01-30
Begin deprecation of public jax.ShapedArray

jakevdp:
`jax.ShapedArray` shouldn't be a public type. This PR doesn't prevent `jax.ShapedArray` from being used, but it makes it so that some static typecheckers will flag it as a non-exported value.


yashk2810 (approved):

Balfredino123 (commented):
Encadenet,

PR #14207, opened 2023-01-30
Fix test failures under SciPy 1.10.0.

hawkinsp:
Using `jax.numpy.power` inside a callback means that the callback yields JAX arrays, which confuse SciPy.


jakevdp (approved):
Thanks!

Issue #2079, opened 2020-01-26
Implement NumPy sorting routines


shoyer:
xref https://docs.scipy.org/doc/numpy/reference/routines.sort.html

Extracted from #70:

 - [x] np.sort
 - [x] np.lexsort
 - [x] np.argsort
 - [x] np.msort
 - [x] np.sort_complex
 - [x] np.partition
 - [ ] np.argpartition

All of these should be relatively doable, though it's worth noting that I don't think XLA currently has efficient partitioning schemes.

Vaibhavs10:
@shoyer and @srvasude - Can I help any of these features into JAX?

P.S. I'm a beginner in terms of Open Source Contributions, so checking with you both before getting my hands dirty with this.

jakevdp:
Hi @Vaibhavs10 - this is a bit complicated, because sorts must be implemented in terms of [XLA operations](https://www.tensorflow.org/xla/operation_semantics#sort), which are exposed in JAX in https://github.com/google/jax/blob/master/jax/lax/lax.py.

Of the ones listed above, I think `lex_sort` and `sort_complex` would be doable using XLA's sorting of multiple inputs with a custom comparator, but unfortunately `lax.sort` doesn't yet support passing a custom comparator: that would have to be added as a precursor to this work.

jakevdp:
Actually, it looks like `np.sort_complex` would be a fairly good place to start. `lax.sort()` already handles complex input correctly as far as I can tell. What would need to happen is to define `jax.numpy.sort_complex` similarly to `sort` here: https://github.com/google/jax/blob/c1aeb8b3fe94f2a064ead43d8d9a1c35982d11d5/jax/numpy/lax_numpy.py#L3141-L3151

with modifications that make it follow the semantics of `np.sort_complex` (things like: real inputs result in complex outputs, etc.

And then add a test of `sort_complex` similar to the test of `sort` here: https://github.com/google/jax/blob/c1aeb8b3fe94f2a064ead43d8d9a1c35982d11d5/tests/lax_numpy_test.py#L2569-L2589

If you want to try this out, that would be great! If you need, feel free to open a draft pull request and tag me if you have questions.

Vaibhavs10:
Thanks @jakevdp I’ll look at this tomorrow in my day time and raise a PR hopefully by weekend.

tudorcebere:
Is this issue still open regarding partitioning schemes?

jakevdp:
[...]

PR #14209, opened 2023-01-30
Add implementation of jnp.partition

jakevdp:
This implements `jnp.partition` in terms of two calls to `lax.top_k` concatenated together. A nice feature of this approach is that XLA knows how to elide unnecessary parts of concat+slice, so the common pattern of `np.partition(x, n)[n:]` optimizes to a single top-k call under JIT:
```python
import jax
import jax.numpy as jnp

x = jnp.ones(10)
top_three = lambda x: jnp.partition(x, -3)[-3:]
print(jax.jit(top_three).lower(x).compile().as_text())
```
```
HloModule jit__lambda_, entry_computation_layout={(f32[10]{0})->f32[3]{0}}, allow_spmd_sharding_propagation_to_output=true

ENTRY %main.21 (Arg_0.1: f32[10]) -> f32[3] {
  %Arg_0.1 = f32[10]{0} parameter(0), sharding={replicated}
  %custom-call = (f32[3]{0}, s32[3]{0}) custom-call(f32[10]{0} %Arg_0.1), custom_call_target="TopK"
  ROOT %get-tuple-element = f32[3]{0} get-tuple-element((f32[3]{0}, s32[3]{0}) %custom-call), index=0
}
```

Part of #70 and #2079

### DIFF
--- a/tests/lax_numpy_test.py
+++ b/tests/lax_numpy_test.py
@@ -3604,6 +3604,30 @@ class LaxBackedNumpyTests(jtu.JaxTestCase):
       self._CompileAndCheck(jnp.msort, args_maker)
 
   @jtu.sample_product(
+    [{'shape': shape, 'axis': axis, 'kth': kth}
+     for shape in nonzerodim_shapes
+     for axis in range(-len(shape), len(shape))
+     for kth in range(-shape[axis], shape[axis])],
+    dtype=default_dtypes,
+  )
+  def testPartition(self, shape, dtype, axis, kth):
+    rng = jtu.rand_default(self.rng())
+    arg = rng(shape, dtype)
+    jnp_output = jnp.partition(arg, axis=axis, kth=kth)
+    np_output = np.partition(arg, axis=axis, kth=kth)
+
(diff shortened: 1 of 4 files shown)


tests/lax_numpy_test.py
```diff
@@ -3603,6 +3603,34 @@ def testMsort(self, dtype, shape):
...
+  )
+  def testPartition(self, shape, dtype, axis, kth):
+    print(shape, dtype, axis, kth)
+    rng = jtu.rand_default(self.rng())
+    arg = rng(shape, dtype)
+    jnp_output = jnp.partition(arg, axis=axis, kth=kth)
+    np_output = np.partition(arg, axis=axis, kth=kth)
+
+    print(np_output.shape)
+    print(jnp_output.shape)
```
sharadmv, line 3621: Delete prints

sharadmv (approved):
LGTM! Thanks

tests/lax_numpy_test.py
jakevdp: Done

PR #14211, opened 2023-01-30
Update signature of ad.defbilinear to simplify transpose rules

jakevdp:
The `lhs_shape` and `rhs_shape` parameters were left over from a very early JAX version, when the transpose rule did not have access to the input avals.

Still up for discussion: do we need `defbilinear2` or should we just modify `defbilinear` and update the handful of other places where it's called?


mattjj:
> Still up for discussion: do we need defbilinear2 or should we just modify defbilinear and update the handful of other places where it's called?

The latter option sgtm!

mattjj (approved):
nice!

PR #14215, opened 2023-01-31
Add select and scatter 2 lax

jkr26:
Adds select_and_scatter to lax.

In writing an interpreter to rehydrate HLO modules into Jaxprs, as suggested [here](https://github.com/google/jax/discussions/14147), I encountered SelectAndScatter in rehydrating the gradient of a MaxPool (I believe). Reading the documentation for SelectAndScatter, it was not obvious to me how to implement in terms of alternative lax symbols. But found this symbol, and in my experimental implementation this does indeed seem to be the thing I was looking for. So simply opening a PR adding to lax.

### DIFF
--- a/jax/_src/lax/windowed_reductions.py
+++ b/jax/_src/lax/windowed_reductions.py
@@ -195,11 +195,15 @@ def _reduce_window_logaddexp(
       window_dilation=tuple(window_dilation))
   return out
 
-def _select_and_scatter(operand: Array, select: Callable,
+def select_and_scatter(operand: Array, select: Callable,
                         window_dimensions: core.Shape,
                         window_strides: Sequence[int],
                         padding: Sequence[Tuple[int, int]], source: Array,
                         init_value: Array, scatter: Callable) -> Array:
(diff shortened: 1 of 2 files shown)


jakevdp:
Thanks - it seems reasonable to add this to the public lax API. A couple additional things we'd need:

- please add the new function to the list of operators in the docs here: http://go/jax-github/blob/main/docs/jax.lax.rst#operators
- please add a test of the new public API. I believe that the XLA implementation of this operator is somewhat incomplete, so you may find that comprehensive tests fail for some corner cases.

hawkinsp:
One thing to note that since this is a maxpool gradient, it already has something of a test here: https://github.com/google/jax/blob/b374080e855deb2af6d7c5f9decfefd547c94e95/tests/lax_autodiff_test.py#L749

If there is a public API it should be tested explicitly though.

jakevdp:
Closing for now, since we're still waiting on tests and documentation.

Feel free to re-open if you're still interested in working on this!

Issue #5687, opened 2021-02-09
map_coordinates mode='constant' not properly applied


agudallago:
jax.scipy.ndimage.map_coordinates behaves differently than scipy.ndimage.map_coordinates when using mode="constant". In this example we can see that 2.9 which should be as "off the array" as 3, produces a different output which is in fact wrong (1) but should in fact produce what the original implementation produces (0, which is cval).

```
src = jnp.array([5, 1, 6])  # Fails!
# src = jnp.array([5, 1, 5]) # Doesn't fail

# Since 2.9 is first jnp.floored, this gives "2". Which is a valid index.
coords = [[2, 3, 2.9]]


def scipy_map_coordinates():
  return ndimage.map_coordinates(src, coords, order=1, mode="constant")


def jax_map_coordinates():
  return jax.scipy.ndimage.map_coordinates(
      src, coords, order=1, mode="constant")


print(scipy_map_coordinates())
print(jax_map_coordinates())

assert jnp.array_equal(scipy_map_coordinates(), jax_map_coordinates())
```

agudallago:
5,1,5 doesn't fail cause the 5*weight gets internally rounded down to zero... matching "by luck" the original implementation output. The internal tests do a weird padding and refactor the input from [5,1,6] to [5,1,6,0] which doesn't seem to make sense with the expected behaviour.

shoyer:
I think this relates to some of the discreptancies related to boundary conditions that we noticed in the behavior of SciPy's map_coordinates: https://github.com/scipy/scipy/issues/2640

The SciPy issue is marked as closed by https://github.com/scipy/scipy/pull/12776. It sounds like the decision was to add a new mode `grid-constant` rather than to attempt to fix `constant`. I'm not sure that's the choice I would have made, but I guess it is probably the right decision for us, too.

agudallago:
Hi @shoyer, thanks for your answer!

I apologize for my description, it may have not been as clear as it sounded in my mind. 

From my experiments it seems to me that the jax version of `map_coordinates` is equivalent to `grid-constant` in SciPy 1.6. 
i.e. jax version is interpolating even outside the edges.
See https://docs.scipy.org/doc/scipy/reference/tutorial/ndimage.html#ndimage-interpolation-modes for more details of what I mean in order=1 and mode="grid_constant"

```
src = jnp.array([5,1,6])
coords = [[3]]

# This returns src[3] -> cval=0 -->>> OK!
[...]

PR #14218, opened 2023-01-31
Cubic splines

### LINKED ISSUE #3928: Implement higher order interpolation in jax's map_coordinates
Hi, I've been trying to use Jax to speed up a program that uses `scipy.ndimage.map_coordinates` extensively. Was able to use jax's `map_coordinates` to get a significant speedup, as well as using `jit` to speed up even more when looping many times. Pretty amazing.

I am aware that the current `map_coordinates` in jax only supports some "modes" and up to order "1", but our program has been using `order=3`, Is there a plan to implement such anytime soon? I am trying to implement such feature but haven't made much progress so far. 

https://github.com/scipy/scipy/blob/v1.5.2/scipy/ndimage/src/ni_splines.c

I went to the scipy.ndimage C-codes, and found the scipy.map_coordinates got the weights, but not quite entirely sure what those "weights" are, and how did they come up with such coefficients without actually solving the spline interpolation system. Would very much appreciate anybody's help!

LouisDesdoigts:
Adds cubic spline interpolation method to `scipy.ndimage.map_coordinates` with natural boundary conditions based on this paper https://doi.org/10.1007/s10614-007-9092-4.

The results given by this interpolation differ at the edges because scipy doesn't use natural boundary conditions, but does match at the inner regions. 

Similarly I have raised a `NotImplementedError` for the 'reflect' mode because it gives incorrect answers based on the coordinates output from the `_INDEX_FIXERS` with 'reflect'. I believe this could be fixed by implementing a different function to generate the output coordinates.

I also have not implemented unit tests or tests comparing to scipy (due to the boundary conditions). 

I am hoping to have some help getting this PR through soon since I need to integrate this into the software I am developing, all help is greatly appreciated!

This should also close #3928 !


jakevdp:
Hi - sorry we missed this! Can you please add some tests for the new functionality? Existing tests for this function can be found here: https://github.com/google/jax/blob/b4ec72deaefa1094891829cd0b78efb5cc19c28b/tests/scipy_ndimage_test.py#L79

LouisDesdoigts:
Hey Jake thanks for picking this up! 

So I've locally updated the tests to cover the third order interpolation however I've implemented a version with a different boundary condition, scipy uses the 'not-a-knot' condition as far as I can tell (they are not explicit about the particular algorithm they use) whereas I've implemented 'natural' boundary conditions. This gives machine precision matching results at the inner parts of the array, but differing results at the boundaries (shown below).

 
Screenshot 2023-03-08 at 2 52 24 pm


The existing tests use arrays of shapes (5,), (3, 4), (3, 4, 5) meaning that the tests comparing to scipy fail and can't just compare the central portion as the arrays are too small. I'm wondering what the way forward is here. I haven't been able to solve this problem for the 'not-a-knot' boundary conditions, which may not even be the correct one and I think will require an entirely different implementation 😬. I'm thinking as a solution the tests for third order could use larger arrays and only compare the inner regions (which isn't exactly ideal, but works) and specifying in the docs about using different boundary conditions to scipy.

Let me know your thoughts!

shoyer:
Did you see the discussion in https://github.com/google/jax/issues/5687?

The existing linear interpolation scheme is somewhat inconsistent with scipy but we would like to fix this. In particular see the proposed solution in https://github.com/google/jax/issues/5687#issuecomment-776866617

Is your PR consistent with one of SciPy (new) supported boundary modes?
https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#interpolation-boundary-handling

LouisDesdoigts:
Yes I'm aware of the current issues with the boundary mode and have been following these, my PR does not address anything to do with this. The issue you linked is to do with the boundary _mode_, how the output coordinates are mapped outside of the range of the input coordinates, which is different to the boundary _conditions_ I'm referring to in my above comments. 

The boundary conditions are a set of extra conditions at the boundary of the arrays that are needed to solve for the spline coefficients, of which there are a few although scipy doesn't provide options for this. The plot in the above comment is defined entirely within the input coordinates so there are no effects from the boundary mode.

There should be no issues with any future changes made to the boundary modes since this just changes how the output coordinates are mapped, which are then subsequently fed into the interpolation algorithm.

JamesAllingham:
I just wanted to mention that it would be great to have this merged, even if it does not match the scipy boundary conditions. Figuring out what those boundary conditions even are and then implementing them would be challenging. And having some form of cubic interpolation for `map_ccordinates` is better than having none :) 

I am happy to help out with making this happen if you need it. Let me know what needs to be done!

LouisDesdoigts:
@JamesAllingham If you need something now you can build my jax fork from source until we get this though https://github.com/LouisDesdoigts/jax. Also feel free to send me your email and I can add you to a thread I have trying to sort this out! Any help is greatly appreciated!

benjaminpope:
Just checking on this - this is a dependency for some science we are currently doing.

shoyer:
I really appreciate you putting this together, but I don't think it's a good idea for JAX's cubic spline code to diverge further from SciPy. The JAX team does not have the expertise on the numerics of splines to get this code right.

I would suggest either:

1. Work on getting these features into SciPy. Once they have been merged over there, we would be happy to consider your PR to mirror the features in JAX.
2. Release a separate library with JAX splines. We would be happy to include a link in the relevant section of the JAX docs.

jakevdp:
Thanks for this – I think given the discussion in https://github.com/google/jax/pull/18137, this functionality is probably out-of-scope for jax core. Sorry we weren't able to accept your contribution here!

Issue #14220, opened 2023-01-31
`lax.cond` excessive caching behavior, how to avoid?


thisiscam:
### Description

I noticed that when using `lax.cond`, the conditional functions are [converted to jaxprs and are cached](https://github.com/google/jax/blob/main/jax/_src/lax/control_flow/common.py#L66).  
This causes a memory-leak like behavior when the conditionals are changing across compilation runs. Here's a MWE:
```python
import jax

def f(x):
  return x**2

identity = lambda x: x

def g(x):
  return jax.lax.cond(x > 0.0, identity, jax.grad(f), operand=x)

while True:
  g1 = jax.jit(g)
  r = f1(1.)
  g1.clear_cache()
```

In this example, I expect `f1.clear_cache` to discard anything related to compiling `g`. However, I'm observing that memory keeps growing across iterations. The memory footprint only reaches a plateau after a few thousand iterations of the loop.

Here's a plot of the memory growth of the above snippet:
![mem](https://user-images.githubusercontent.com/7684342/215710748-90d11427-e377-4850-88f7-c1fc384c07d6.png)

Notice the linear memory growth from 1 ~ 80 seconds. 

My use case is that I have an `g` with varying shapes across the outer loop. 
Therefore, recompilation is required. Moreover, `clear_cache` is required, to avoid excessively caching `g1` that are used only once.

My understanding is: Each new reference produced by `jax.grad(f)` confuses `lax.cond`, causing it to cache all the jaxprs of `jax.grad(f)`.  This problem is really unfortunate for me, since my `f` is quite complicated, and a few iterations of the loop produces more than 4 GBs of cached jaxprs (that are never used again, due to the varying shapes). 

Is there a good way to prevent this behavior? Or, perhaps, this could be addressed together with #12719?

### What jax/jaxlib version are you using?

jax v0.4.2

### Which accelerator(s) are you using?

CPU

jakevdp:
I don't think this has anything to do with the jit cache; we can directly query the size of the jit cache this way, and see that it is not growing:
```python
import jax
from jax._src.api import _cpp_jit_cache

def f(x):
  return x**2

identity = lambda x: x

def g(x):
  return jax.lax.cond(x > 0.0, identity, jax.grad(f), operand=x)

for i in range(10):
  print(_cpp_jit_cache.size(), end=' ')
  g1 = jax.jit(g)
  r = g1(1.)
print()
# 0 1 2 2 2 2 2 2 2 2 
```
[...]

PR #14231, opened 2023-01-31
Use weakref_lru_cache for initial_style_jaxprs_with_common_consts

jakevdp:
Related to #14220

### DIFF
--- a/jax/_src/lax/control_flow/common.py
+++ b/jax/_src/lax/control_flow/common.py
@@ -23,7 +23,7 @@ from jax._src.core import control_flow_allowed_effects as allowed_effects
 from jax._src.lax import lax
 from jax._src import ad_util
 from jax._src import util
-from jax._src.util import cache, weakref_lru_cache, safe_map, unzip3
+from jax._src.util import weakref_lru_cache, safe_map, unzip3
 from jax.api_util import flatten_fun_nokwargs
 from jax.interpreters import partial_eval as pe
 from jax.tree_util import tree_map, tree_unflatten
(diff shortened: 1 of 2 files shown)


jax/_src/lax/control_flow/common.py
```diff
@@ -62,9 +62,9 @@ def _initial_style_jaxpr(fun: Callable, in_tree, in_avals,
   closed_jaxpr = core.ClosedJaxpr(pe.convert_constvars_jaxpr(jaxpr), ())
   return closed_jaxpr, consts, out_tree
 
-@cache()
+@weakref_lru_cache
```
mattjj, line 65: I don't think this does what we want; we want the cache keys to hold weak refs to each of the `funs`, and for the entry to disappear when one of the components of `funs` has its last strong reference dropped.

As written, we're just holding a weakref to the first of `funs` (see [this comment](https://github.com/google/jax/blob/8dc1dff61098467dbd52b72eb542f2e205543c59/jax/_src/util.py#L266)).

Or maybe I'm mistaken?

mattjj (changes requested):

jax/_src/lax/control_flow/common.py
jakevdp: Oh, I see - I thought it used weak refs to all arguments. Thanks - it looks like there's not any good way to use `weakref_lru_cache` here without digging into the C implementation.

mattjj: Sorry, I didn't articulate it very well in the other comment!

PR #14233, opened 2023-01-31
Implement jax.util.safe_zip using builtin zip with strict=True in Python

hawkinsp:
3.10+.

The builtin zip is slightly faster.


yashk2810 (approved):
Woah, nice!

hawkinsp:
Closing, the pybind11 implementation is still faster than the python-only alternative.

PR #14237, opened 2023-02-01
jax.numpy: remove private members from interface

jakevdp:
In particular, this removes `jnp._add_newdoc_ufunc`, `jnp._pyinstaller_hooks_dir`, and possibly others.

### DIFF
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -431,7 +431,7 @@ def _init():
   from jax._src import util
   # Builds a set of all unimplemented NumPy functions.
   for name, func in util.get_module_functions(np).items():
-    if name not in globals():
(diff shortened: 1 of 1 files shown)


jax/numpy/__init__.py
```diff
@@ -431,7 +431,7 @@ def _init():
   from jax._src import util
   # Builds a set of all unimplemented NumPy functions.
   for name, func in util.get_module_functions(np).items():
-    if name not in globals():
+    if name not in globals() and not name.startswith('_'):
```
yashk2810, line 434: Should this be name.startswith('_') without the `not`?

jakevdp: No, I think it's correct as written. We don't want to add `name` to globals if it starts with an underscore.

PR #14242, opened 2023-02-01
[jax2tf] Fixed constant sharing in the generated TF code

gnecula:
In order to avoid embedding large constants multiple times in the generated code, jax2tf will keep a cache of constants during the lowerig to TF. However, the separate branches of a tf.switch_case will result in separate tf.function, and they should not share constants. So, we ensure we use a fresh constant cache when lowering the different branches. Same idea for tf.while.


gnecula:
@rxsang FYI

PR #14248, opened 2023-02-01
Clean up: remove unused helper functions

jakevdp:
Used [vulture](https://github.com/jendrikseipp/vulture) to find all un-referenced function definitions whose name starts with an underscore. Also confirmed that each has no downstream callers internally.

### DIFF
--- a/jax/_src/lax/convolution.py
+++ b/jax/_src/lax/convolution.py
@@ -26,7 +26,6 @@ from jax._src import core
 from jax._src import dtypes
 from jax._src import util
 from jax._src.lax import lax
-from jax._src.lib import xla_client
 from jax._src.lib.mlir.dialects import hlo
 
 _max = builtins.max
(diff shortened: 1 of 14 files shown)


yashk2810 (approved):
Woah! Nice!

jax/_src/lax/convolution.py
```diff
@@ -747,27 +746,6 @@ def _reshape_axis_out_of(src, size1, x):
   return lax.reshape(x, shape)
 
 
-def _check_conv_shapes(name, lhs_shape, rhs_shape, window_strides):
```
soraros, line 750: This looks like a useful function to have. Shouldn't it be used in say `lax.conv`?

mattjj:
Amazing!

jax/_src/lax/convolution.py
jakevdp: Maybe it could be used? I can't find the last time this function was called in the codebase, so it would take some work to ensure the logic is correct in all corner cases.

PR #14269, opened 2023-02-02
Remove placeholder functions for unimplemented NumPy functions.

hawkinsp:
These don't seem necessary now JAX has fairly complete coverage of the NumPy API. Also removes the accidental export of _NOT_IMPLEMENTED in several modules.

### DIFF
--- a/jax/numpy/__init__.py
+++ b/jax/numpy/__init__.py
@@ -257,7 +257,6 @@ from jax._src.numpy.lax_numpy import (
     where as where,
     zeros as zeros,
     zeros_like as zeros_like,
-    _NOT_IMPLEMENTED,
 )
 
 if xla_extension_version >= 117:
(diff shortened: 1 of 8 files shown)


jax/numpy/__init__.py
```diff
@@ -432,7 +432,6 @@ def _init():
   # Builds a set of all unimplemented NumPy functions.
   for name, func in util.get_module_functions(np).items():
     if name not in globals() and not name.startswith('_'):
-      _NOT_IMPLEMENTED.append(name)
       globals()[name] = lax_numpy._not_implemented(func, module='numpy')
```
jakevdp, line 435: Perhaps we should remove this as well? I think having unimplemented functions in the namespace may have been helpful early on, but the remaining functions now are mostly ones we never plan to implement.

hawkinsp: Done.

jakevdp (approved):
Looks great!

PR #14273, opened 2023-02-02
shard_map (shmap) prototype and JEP

mattjj:
Add the `shard_map` / `shmap` JEP design doc, and prototype implementation in `jax.experimental.shard_map`.

`shmap` is a simple multi-device parallelism API which lets us write per-device code with explicit collectives, where logical shapes match per-device physical buffer shapes and collectives correspond exactly to cross-device communication. It composes well with `pjit` (which is now just `jit`), both in terms of efficiently passing a `shmap`'s outputs to a `pjit`'s inputs, and in terms of nesting a `shmap` inside a `jit`/`pjit`. In particular, `shmap` can be used inside a `jit`/`pjit` to drop temporarily into a "manual collectives" mode, letting us write explicit per-device code and collective communication (rather than relying on the compiler to automatically partition computation and insert collectives).

```python
import jax
import jax.numpy as jnp
from jax.sharding import Mesh, PartitionSpec as P
from jax.experimental import mesh_utils

devices = mesh_utils.create_device_mesh((4, 2))
mesh = Mesh(devices, axis_names=('x', 'y'))

x = jnp.arange( 8 * 16.).reshape(8, 16)
y = jnp.arange(16 * 32.).reshape(16, 32)

@partial(shmap, mesh=mesh, in_specs=(P('x', 'y'), P('y', None)),
         out_specs=P('x', None))
def matmul_basic(x_block, y_block):
  # x_block: f32[2, 8]
  # y_block: f32[8, 32]
  z_partialsum = jnp.dot(x_block, y_block)
  z_block = jax.lax.psum(z_partialsum, 'y')
  return z_block

z = matmul_basic(x, y)  # z: f32[8, 32]
```

See the JEP for more!

TODO:
* [x] split out some not-strictly-related changes into separate PRs (#14276)
* [x] add JEP doc
* [x] write pr message

### DIFF
--- a/docs/jep/14273-shard-map.md
+++ b/docs/jep/14273-shard-map.md
@@ -1,0 +1,640 @@
+# `shmap` (`shard_map`) for simple per-device code
+*sholto@, sharadmv@, jekbradbury@, zhangqiaorjc@, mattjj@*
+
+*January 2023*
+
+## Motivation
+
+JAX supports two schools of thought for multi-device programming:
+1. **Compiler, take the wheel!** Let the compiler automatically partition bulk
+   array functions over devices.
+2. **Just let me write what I mean, damnit!** Give me per-device code and
+   explicit communication collectives.
+
+We need great APIs for both, and rather than being mutually exclusive
+alternatives, they need to compose with each other.
+
+With `pjit` (now just `jit`) we have [a next-gen
+API](https://jax.readthedocs.io/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html)
+for the first school. But we haven't quite leveled-up the second school. `pmap`
+follows the second school, but over time we found it has [fatal
+flaws](#why-dont-pmap-or-xmap-already-solve-this). `xmap` solved those flaws,
+but it doesn't quite give us per-device shapes, and it includes several other
+big ideas too. Meanwhile, new demands for per-device explicit-collectives
+programming have emerged, like in [Efficiently Scaling Transformer
+Inference](https://arxiv.org/abs/2211.05102).
+
+We can level-up the second school with `shmap`. `shmap` is:
+* a simple multi-device parallelism API which lets us write per-device code with
+  explicit collectives, where logical shapes match per-device physical buffer
+  shapes and collectives correspond exactly to cross-device communication;
+* a specialization of `xmap` with scaled-back features and a few tweaks;
+* a fairly direct surfacing of the XLA SPMD Partitioner's 'manual' mode;
+* a fun-to-say Seussian name which could stand for `shard_map`,
+  `shpecialized_xmap`, `sholto_map`, or `sharad_map`.
+
+**For `pjit` users**, `shmap` is a complementary tool. It can be used inside a
+`pjit` computation to drop temporarily into a "manual collectives" mode, like an
+escape hatch from the compiler's automatic partitioning. That way, users get the
+convenience and familiar just-NumPy programming model of `pjit` for most of their
+code, along with the ability to hand-optimize collective communication with
+`shmap` wherever it's needed. It's the best of both worlds!
+
+**For `pmap` users**, `shmap` is a strict upgrade. It's more expressive,
+performant, and composable with other JAX APIs, without making basic batch data
+parallelism any harder.
+
+For more on practical use, you can jump to [When should you use `shmap` and when
+should you use `pjit`?](#when-should-you-use-shmap-and-when-should-you-use-pjit).
+If you're wondering why we need a new thing at all, or what
+the problems with `pmap` are, jump to [Why don't `pmap` or `xmap` already solve
+this?](#why-dont-pmap-or-xmap-already-solve-this).
+Or keep reading the next section to see some `shmap` examples and the API spec.
+
+
+## So, let's see `shmap`!
+
+### TL;DR example (with a more detailed explanation to follow)
+
+Sho shick:
+
+```python
+import jax
+import jax.numpy as jnp
+from jax.sharding import Mesh, PartitionSpec as P
+from jax.experimental import mesh_utils
+
+devices = mesh_utils.create_device_mesh((4, 2))
+mesh = Mesh(devices, axis_names=('x', 'y'))
+
+a = jnp.arange( 8 * 16.).reshape(8, 16)
+b = jnp.arange(16 * 32.).reshape(16, 32)
+
+@partial(shmap, mesh=mesh, in_specs=(P('x', 'y'), P('y', None)),
+         out_specs=P('x', None))
+def matmul_basic(a_block, b_block):
+  # a_block: f32[2, 8]
+  # b_block: f32[8, 32]
+  z_partialsum = jnp.dot(a_block, b_block)
+  z_block = jax.lax.psum(z_partialsum, 'y')
+  return z_block
+
+c = matmul_basic(a, b)  # c: f32[8, 32]
+```
+
+Notice:
+* no nesting needed (or `axis_index_groups`) for multiple axes of parallelism,
+  unlike `pmap`;
+* no reshapes in the caller, unlike `pmap` and hard-`xmap`, and logical shapes
+  correspond to per-device physical shapes, unlike (non-hard) `xmap`;
+* precise device placement control by using `mesh`, unlike `pmap`;
+* there's only one set of axis names for logical and physical, unlike `xmap`;
+* the result is a `jax.Array` which could be efficiently passed to a `pjit`,
+  unlike `pmap`;
+* this same code works efficiently inside a `pjit`/`jit`, unlike `pmap`;
+* this code works eagerly, so we can `pdb` in the middle and print values,
+  unlike `xmap`'s current implementation (though by design `xmap` without the
+  sequential schedule can in principle work eagerly too).
+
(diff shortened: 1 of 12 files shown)


jax/_src/dispatch.py
```diff
@@ -571,6 +572,15 @@ def jaxpr_shardings(jaxpr) -> Iterator[jax.sharding.XLACompatibleSharding]:
     elif eqn.primitive is pjit.pjit_p:
       yield from eqn.params['in_shardings']
       yield from eqn.params['out_shardings']
+    # TODO(mattjj): replace string matching with object id
```
yashk2810, line 575: This will directly affect pjit i.e. if you have a shard_map inside pjit, the NamedSharding will be taken into account for getting the device assignment. That is intended right?

jax/_src/dispatch.py
```diff
@@ -571,6 +572,15 @@ def jaxpr_shardings(jaxpr) -> Iterator[jax.sharding.XLACompatibleSharding]:
...
+    # elif eqn.primitive.name == 'device_put2':
```
yashk2810, line 576: What's device_put2? Can this be removed?

jax/_src/dispatch.py
mattjj: Yes! That's the point of the change.

I can fix the TODO now though, thanks for the reminder...

jax/_src/dispatch.py
mattjj: Yes, thanks. It's vestigial. (But `device_put2` was an implementation of unified-device-put-and-with-sharding-constraint which I had in this branch earlier...)

jax/_src/dispatch.py
mattjj: Done!

jax/_src/dispatch.py
mattjj: Done!

yashk2810: Huh! I also have an implementation sitting in a client lol.

docs/jep/14273-shard-map.md
```diff
@@ -0,0 +1,640 @@
...
+    )
+    # device 0 computes chunks 0, 1, ...
+    # device 1 computes chunks 1, 2, ...
+    update_index = (((axis_index + i) % axis_size) * chunk_size, 0)
+    accum = jax.lax.dynamic_update_slice(accum, update, update_index)
+    return accum, lhs
+
+  accum = jnp.zeros((lhs.shape[0] * axis_size, rhs.shape[1]), dtype=lhs.dtype)
+  # fori_loop cause a crash: hlo_sharding.cc:817 Check failed: !IsManual()
+  # accum, lhs = jax.lax.fori_loop(0, axis_size - 1, f, (accum, lhs))
```
sharadmv, line 435: Does this fori_loop still cause a crash?

sharadmv (approved):
Sho shick

docs/jep/14273-shard-map.md
mattjj: Oh good question... maybe @zhangqiaorjc can try it again? :D

sharadmv:
You'll need to update tests/BUILD.

jax/interpreters/mlir.py
```diff
@@ -542,6 +542,7 @@ def register_lowering(prim: core.Primitive, rule: LoweringRule,
     # this expansion.
     for p in xb.expand_platform_alias(platform):
       _platform_specific_lowerings[p][prim] = rule
+  return rule
```
yashk2810, line 545: Why do you need this return rule?

shard_map.py doesn't do anything with the returned value AIUI?

jax/experimental/shard_map.py
```diff
@@ -0,0 +1,913 @@
...
+                           memoize, partition_list, merge_lists)
+from jax.api_util import flatten_fun_nokwargs
+from jax.experimental import maps
+from jax.interpreters import batching
+from jax.interpreters import mlir
+from jax.interpreters import partial_eval as pe
+from jax.interpreters import xla
+from jax.interpreters import pxla
+from jax.interpreters import ad
+from jax.interpreters.pxla import Mesh
```
yashk2810, line 45: Use jax.sharding.Mesh

jax/interpreters/mlir.py
mattjj: That change is unrelated, and could be removed from this PR. It's just a good pattern when you might use something as a decorator, like

```python
@register_thing
def foo(...):
  ...
```

If `register_thing` doesn't return its argument then the code above binds `foo = None`.

jax/experimental/shard_map.py
mattjj: Ah yes, ty!

jax/interpreters/mlir.py
yashk2810: Ohh, cool! Thanks for explaining.

jax/experimental/shard_map.py
mattjj: Done.

jax/interpreters/mlir.py
mattjj: I don't remember what made me notice it while working on this... but maybe at one point I wanted to write

```python
@partial(mlir.register_lowering, shard_map_p)
def _shard_map_lowering(...):
  ...
```

and still have `_shard_map_lowering` bound to the actual function object afterwards.

jax/experimental/shard_map.py
```diff
@@ -0,0 +1,912 @@
...
+    bind = partial(call_primitive.bind, fun)  # TODO caching (compat w/ jaxpr())
+    fake_primitive = pxla._FakePrimitive(multiple_results=True, bind=bind)
+    _rep_rules[fake_primitive] = lambda *_, **__: set()
+    out_tracers_ = self.process_primitive(fake_primitive, tracers, params)
+    out_vals = [t.val for t in out_tracers_]
+    out_rep = _output_rep(self.mesh, jaxpr(), [t.rep for t in tracers])
+    return map(partial(ShardMapTracer, self), out_rep, out_vals)
+
+@lu.transformation_with_aux
+def _something_shady(*args):
```
yashk2810, line 581: Please please name it better?

tests/BUILD
```diff
@@ -1007,6 +1007,16 @@ jax_test(
     },
 )
 
+jax_test(
+    name = "shard_map_test",
+    srcs = ["shard_map_test.py"],
+    shard_count = {
+        "cpu": 10,
```
yashk2810, line 1014: Do you need shards? It is a good idea to add them if the tests time out without having any shards but with these tests.. I don't think it will time out.

jax/experimental/shard_map.py
mattjj: You mean with like another underscore or two on the end? 😁 

Okay, done!

jax/experimental/shard_map.py
```diff
@@ -0,0 +1,912 @@
...
+      fails, = e.args
+      msg = _rep_error(f, mesh, out_tree(), out_specs, fails)
+      raise ValueError(msg) from None
+    return tree_unflatten(out_tree(), out_flat)
+  return wrapped
+
+# Internally use AxisNames = Dict[int, Tuple[AxisName, ...]], not PartitionSpecs
+AxisName = Hashable
+AxisNames = Dict[int, Tuple[AxisName, ...]]  # TODO(mattjj): make it hashable
+def _canonicalize_spec(spec: PartitionSpec) -> AxisNames:
```
yashk2810, line 114: Isn't this pjit.array_mapping?

yashk2810: https://github.com/google/jax/blob/main/jax/_src/pjit.py#L1929 -- yes it is! Can we reuse?

jax/experimental/shard_map.py
```diff
@@ -0,0 +1,912 @@
...
+      + '\n\n'.join(msgs) + '\n\n'
+      f"Check the {prefix}_specs values passed to shard_map.")
+
+class NoFail: pass
+no_fail = NoFail()
+
+def _check_specs_vs_args(
+    f: Callable, mesh: Mesh, in_tree: PyTreeDef, in_specs: Specs,
+    in_specs_flat: List[P], xs: List) -> None:
+  in_avals = [core.raise_to_shaped(core.get_aval(x)) for x in xs]
```
yashk2810, line 142: Use api_util.shaped_abstractify? That might be faster here!

tests/BUILD
mattjj: Deleted!

jax/experimental/shard_map.py
mattjj: It looks different to me. The output of `_canonicalize_spec` is a `Dict[int, Tuple[AxisName, ...]]`, but the output of the `get_array_mapping` you linked is an ordered version of `Dict[AxisName, int]` (unioned with some other stuff). That is, I think they're perhaps flipped?

jax/experimental/shard_map.py
mattjj: Done!

jax/experimental/shard_map.py
yashk2810: Ohh yeah, I misread. Cool!

PR #14275, opened 2023-02-03
Fix jax.numpy.hsplit

xoiga123:
[numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.hsplit.html): 
> hsplit is equivalent to split with axis=1, the array is always split along the second axis except for 1-D arrays, where it is split at axis=0.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -1212,6 +1212,12 @@ def split(ary: ArrayLike, indices_or_sections: Union[int, ArrayLike], axis: int 
 def _split_on_axis(op: str, axis: int) -> Callable[[ArrayLike, Union[int, ArrayLike]], List[Array]]:
   @_wraps(getattr(np, op), update_doc=False)
   def f(ary: ArrayLike, indices_or_sections: Union[int, ArrayLike]) -> List[Array]:
+    # for 1-D array, hsplit becomes vsplit
+    nonlocal axis
+    _check_arraylike(op, ary)
+    a = asarray(ary)
+    if axis == 1 and len(a.shape) == 1:
+      axis = 0
     return _split(op, ary, indices_or_sections, axis=axis)
   return f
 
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Thanks for catching this! This looks like a good way to fix the issue.

Before we merge, can you squash your changes into a single commit? See https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests

Thanks!

jakevdp:
It looks like there are some typing and test failures - please take a look!

Alternatively, if you'd like me to dig into it instead, please let me know.

xoiga123:
Ah my bad, I'll fix that ASAP

xoiga123:
@jakevdp An alternative fix is to use `_axis = 0` and pass in `_split(..., axis=_axis)`, so tell me if you don't like using `nonlocal` in the codebase

jakevdp (approved):
Looks good, thanks!

xoiga123:
My bad again, missed the typing error 😓

xoiga123:
@jakevdp sorry for not running `pre-commit` the last few times, should be fine now

jax/_src/numpy/lax_numpy.py
```diff
@@ -1167,10 +1167,8 @@ def broadcast_to(array: ArrayLike, shape: Shape) -> Array:
   return _broadcast_to(array, shape)
 
 
-def _split(op: str, ary: ArrayLike, indices_or_sections: Union[int, ArrayLike],
+def _split(op: str, ary: Array, indices_or_sections: Union[int, ArrayLike],
            axis: int = 0) -> List[Array]:
-  _check_arraylike(op, ary)
-  ary = asarray(ary)
```
jakevdp, line 1173: Since this check must be done by every wrapper that calls this helper, wouldn't it make more sense to continue to do the check just once here?

jakevdp: If it's a type checking issue, then you can address it by changing the variable name when you cast to array. e.g.
```python
a = asarray(ary)
```
Then `a` is of type `Array` and you can do things like `a.shape` without mypy flagging it as incorrect.

xoiga123: I was kinda scared of that due to creating an extra array, but that does seem cleaner

jakevdp (approved):
Looks good, thanks!

PR #14290, opened 2023-02-04
[shape_poly] Fix the hashing and equality of symbolic dimensions

gnecula:
Make equality and disequality comparisons of symbolic expression total, to ensure proper hashability.
Previously, JAX raised `InconclusiveDimensionOperation` for equality comparisons that cannot be resolved to the same boolean value for all possible valuations of the dimension variables, e.g., `a == 1`. But symbolic dimensions need to be
hashable because JAX often memoizes functions with `ShapedArray` and `Jaxpr` arguments, and those can contain symbolic dimensions. 

However, by making equality total, we introduce unsoundness, because of the form `if x.shape[0] != 1: return 1` will now be staged to different Jaxpr when traced with certain static shapes vs. when traced with symbolic shapes.
 
We have tried in #14200 to obtain that is solution that ensures soundness and some limited hashability, but this resulted in some hashing errors in user code that were very hard to diagnose.


jax/core.py
```diff
@@ -91,13 +91,10 @@
   _check_jaxpr as _check_jaxpr,
   _check_map as _check_map,
   _compact_eqn_should_include as _compact_eqn_should_include,
-  _dim_handler_and_canonical as _dim_handler_and_canonical,
```
marcvanzee, line 94: How is this related to the rest of the PR?

marcvanzee (approved):

jax/core.py
gnecula: These are just drive-by cleanup of the jax.core namespace, for some names related to shape polymorphism

PR #14303, opened 2023-02-06
Add scipy.stats.rankdata.

### LINKED ISSUE #12621: Add scipy.stats.rankdata
Add [scipy.stats.rankdata](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rankdata.html). Examples:

```python3
>>> from scipy.stats import rankdata

>>> rankdata([0, 2, 3, 2])
array([ 1. ,  2.5,  4. ,  2.5])

>>> rankdata([0, 2, 3, 2], method='min')
array([ 1,  2,  4,  2])

>>> rankdata([0, 2, 3, 2], method='max')
array([ 1,  3,  4,  3])

>>> rankdata([0, 2, 3, 2], method='dense')
array([ 1,  2,  3,  2])

>>> rankdata([0, 2, 3, 2], method='ordinal')
array([ 1,  2,  4,  3])

>>> rankdata([[0, 2], [3, 2]]).reshape(2,2)
array([[1. , 2.5],
      [4. , 2.5]])

>>> rankdata([[0, 2, 2], [3, 2, 5]], axis=1)
array([[1. , 2.5, 2.5],
       [2. , 1. , 3. ]])
```

carlosgmartin:
Fixes https://github.com/google/jax/issues/12621. I tried to model it after https://github.com/google/jax/pull/12683.

### DIFF
--- a/jax/_src/scipy/stats/_core.py
+++ b/jax/_src/scipy/stats/_core.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The JAX Authors.
+# Copyright 2023 The JAX Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -23,7 +23,7 @@ from jax._src import dtypes
 from jax._src.api import vmap
 from jax._src.numpy.lax_numpy import _check_arraylike
 from jax._src.numpy.util import _wraps
-from jax._src.typing import ArrayLike
+from jax._src.typing import ArrayLike, Array
 from jax._src.util import canonicalize_axis, prod
 
 ModeResult = namedtuple('ModeResult', ('mode', 'count'))
@@ -39,7 +39,7 @@ def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate", k
   if nan_policy not in ["propagate", "omit", "raise"]:
     raise ValueError(
       f"Illegal nan_policy value {nan_policy!r}; expected one of "
-      "{'propoagate', 'omit', 'raise'}"
+      "{'propagate', 'omit', 'raise'}"
     )
   if nan_policy == "omit":
     # TODO: return answer without nans included.
@@ -81,3 +81,65 @@ def mode(a: ArrayLike, axis: Optional[int] = 0, nan_policy: str = "propagate", k
   x = x.reshape(x.shape[0], prod(x.shape[1:]))
   vals, counts = vmap(_mode_helper, in_axes=1)(x)
   return ModeResult(vals.reshape(output_shape), counts.reshape(output_shape))
+
+def invert_permutation(i: Array) -> Array:
+  """Helper function that inverts a permutation array."""
+  return jnp.empty_like(i).at[i].set(jnp.arange(i.size, dtype=i.dtype))
+
+@_wraps(scipy.stats.rankdata, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=["method", "axis", "nan_policy"])
+def rankdata(
+  a: ArrayLike,
+  method: str = "average",
+  *,
+  axis: Optional[int] = None,
+  nan_policy: str = "propagate",
+) -> Array:
+
+  _check_arraylike("rankdata", a)
+
+  if nan_policy not in ["propagate", "omit", "raise"]:
+    raise ValueError(
+      f"Illegal nan_policy value {nan_policy!r}; expected one of "
+      "{'propoagate', 'omit', 'raise'}"
+    )
+  if nan_policy == "omit":
+    raise NotImplementedError(
+      f"Logic for `nan_policy` of {nan_policy} is not implemented"
+    )
+  if nan_policy == "raise":
+    raise NotImplementedError(
+      "In order to best JIT compile `mode`, we cannot know whether `x` "
+      "contains nans. Please check if nans exist in `x` outside of the "
+      "`rankdata` function."
+    )
+
+  if method not in ("average", "min", "max", "dense", "ordinal"):
+    raise ValueError(f"unknown method '{method}'")
+
+  a = jnp.asarray(a)
+
+  if axis is not None:
+    return jnp.apply_along_axis(rankdata, axis, a, method)
+
+  arr = jnp.ravel(a)
+  sorter = jnp.argsort(arr)
+  inv = invert_permutation(sorter)
+
+  if method == "ordinal":
+    return inv + 1
+  arr = arr[sorter]
+  obs = jnp.insert(arr[1:] != arr[:-1], 0, True)
+  dense = obs.cumsum()[inv]
+  if method == "dense":
+    return dense
+  count = jnp.nonzero(obs, size=arr.size + 1, fill_value=len(obs))[0]
+  if method == "max":
+    return count[dense]
+  if method == "min":
+    return count[dense - 1] + 1
+  if method == "average":
+    return .5 * (count[dense] + count[dense - 1] + 1).astype(dtypes.canonicalize_dtype(jnp.float_))
+  raise ValueError(f"unknown method '{method}'")
--- a/tests/scipy_stats_test.py
+++ b/tests/scipy_stats_test.py
@@ -1129,5 +1129,26 @@ class LaxBackedScipyStatsTests(jtu.JaxTestCase):
                             tol=tol)
     self._CompileAndCheck(lax_fun, args_maker, rtol=tol)
 
+  @jtu.sample_product(
+    [dict(shape=shape, axis=axis)
+      for shape in [(0,), (7,), (47, 8), (0, 2, 3), (10, 5, 21)]
+      for axis in [None, *range(len(shape))
+    ]],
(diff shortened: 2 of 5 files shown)


tests/scipy_stats_test.py
```diff
@@ -1129,5 +1129,45 @@ def scipy_mode_wrapper(a, axis=0, nan_policy='propagate', keepdims=None):
                             tol=tol)
     self._CompileAndCheck(lax_fun, args_maker, rtol=tol)
 
+  @jtu.sample_product(
+    [dict(shape=shape, axis=axis)
+      for shape, axis in (
```
jakevdp, line 1134: Can we make this more compact?
```python
[dict(shape=shape, axis=axis)
 for shape in [(0,), (7,), (47, 8), (0, 2, 3), (10, 5, 21)]
 for axis in [None, *range(len(shape)]]
```

jax/_src/scipy/stats/_rankdata.py
```diff
@@ -0,0 +1,84 @@
...
+import scipy
+from jax import numpy as jnp, jit
+from jax._src import dtypes
+from jax._src.numpy.lax_numpy import _check_arraylike
+from jax._src.numpy.util import _wraps
+from jax._src.typing import Array
+
+def invert_permutation(i: Array) -> Array:
+    """Helper function that inverts a permutation array."""
+    return jnp.empty_like(i).at[i].set(jnp.arange(i.size, dtype=i.dtype))
```
jakevdp, line 27: Here and throughout, please use two-space indentations for consistency with the rest of the package.

jax/_src/scipy/stats/_rankdata.py
```diff
@@ -0,0 +1,84 @@
...
+    if method == "dense":
+        return dense
+    count = jnp.nonzero(obs, size=arr.size + 1, fill_value=len(obs))[0]
+    if method == "max":
+        return count[dense]
+    if method == "min":
+        return count[dense - 1] + 1
+    if method == "average":
+        return .5 * (count[dense] + count[dense - 1] + 1).astype(dtypes.canonicalize_dtype(jnp.float_))
+    assert False  # to silence mypy error "Missing return statement""
```
jakevdp, line 84: I know it's repetetive, but it would be more informative to use
```python
raise ValueError(f"unknown method '{method}'")
```

jax/_src/scipy/stats/_rankdata.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2023 The JAX Authors.
```
jakevdp, line 1: I think it would make sense to put this in `_core.py` rather than its own submodule. The reason all the distributions have their own submodules here is because the submodule structure is part of the API.

jakevdp (changes requested):
Looks great, a few changes below.

Also, please add the new function to the docs under `mode` here: https://github.com/google/jax/blob/main/docs/jax.scipy.rst#jaxscipystats

jax/_src/scipy/stats/_core.py
```diff
@@ -81,3 +81,64 @@ def _mode_helper(x: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
...
+  return jnp.empty_like(i).at[i].set(jnp.arange(i.size, dtype=i.dtype))
+
+@_wraps(scipy.stats.rankdata, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=["method", "axis", "nan_policy"])
+def rankdata(
+  a: Array,
+  method: str = "average",
+  axis: Optional[int] = None,
```
jakevdp, line 96: In scipy, it looks like the `axis` and `nan_policy` arguments are keyword only.

jax/_src/scipy/stats/_core.py
```diff
@@ -81,3 +81,64 @@ def _mode_helper(x: jnp.ndarray) -> Tuple[jnp.ndarray, jnp.ndarray]:
...
+def invert_permutation(i: Array) -> Array:
+  """Helper function that inverts a permutation array."""
+  return jnp.empty_like(i).at[i].set(jnp.arange(i.size, dtype=i.dtype))
+
+@_wraps(scipy.stats.rankdata, lax_description="""\
+Currently the only supported nan_policy is 'propagate'
+""")
+@partial(jit, static_argnames=["method", "axis", "nan_policy"])
+def rankdata(
+  a: Array,
```
jakevdp, line 94: We should generally use `ArrayLike` on function inputs (see https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#inputs-should-be-permissively-typed)

jakevdp (commented):
Looks good! A couple more minor comments

jakevdp (approved):
Looks great, thanks!

jakevdp:
It looks like there are some test failures for `min` and `max` when `contains_nans=True`. Any thoughts on what might be different between the JAX and scipy approaches?

carlosgmartin:
@jakevdp [Here](https://github.com/scipy/scipy/blob/v1.10.0/scipy/stats/_stats_py.py#L9501)'s scipy's treatment of nans. When the input contains nans and `nan_policy == 'propagate'` it returns a nan array (which is a float array, rather than an integer array). My understanding is that JAX can't have the type of the output depend on the input. Thus I simply raise a `NotImplementedError` instead. This is the same approach taken by https://github.com/google/jax/pull/12683/files for the `mode` function.

jakevdp:
Thanks - I think it must be a different issue here, because both the JAX and scipy results are float arrays, but the values do not match. e.g.
```
 (shape=(10, 5, 21), axis=2, dtype=<class 'numpy.float32'>, contains_nans=True, method='max')
```
```
AssertionError: 
Not equal to tolerance rtol=5e-06, atol=5e-06

Mismatched elements: 56 / 1050 (5.33%)
Max absolute difference: 5
Max relative difference: 0.3125
 x: array([[[10, 16, 14, ...,  7, 17,  4],
        [ 9, 21, 10, ...,  4,  6, 18],
        [ 5, 20, 10, ..., 21,  8, 15],...
 y: array([[[10, 16, 14, ...,  7, 17,  4],
        [ 9, 20, 10, ...,  4,  6, 18],
        [ 5, 17, 10, ..., 21,  8, 15],...
```

carlosgmartin:
@jakevdp Do you have access to the explicit input it's failing on? A slice may be sufficient. It looks like they differ on row 1, column 1 of this input, for example?

jakevdp:
No, but you should be able to see the failures yourself. Try this:
```
$ JAX_NUM_GENERATED_CASES=50 pytest -n auto tests/scipy_stats_test.py -k testRankData
```

carlosgmartin:
@jakevdp This might be the origin of the problem:
```python3
from jax import numpy as jnp
import numpy as np

arr = np.asarray([
       np.nan,
       0.33638692,
       -0.10554429,
       -0.9180369 ,
       0.8460967 ,
       -0.270747  ,
       -0.79438305,
       -2.001275  ,
       -0.71514213,
       np.nan,
       np.nan,
       2.1875203 ,
       -0.24735539,
       1.1461773 ,
       -1.4080595 ,
       -0.13080159,
       3.06252
])

print(jnp.argsort(arr))
print(np.argsort(arr, kind='stable'))
print(np.argsort(arr, kind='quicksort')) # version used by scipy rankdata for method != 'ordinal'
```
```
[ 7 14  3  6  8  5 12 15  2  1  4 13 11 16  0  9 10]
[ 7 14  3  6  8  5 12 15  2  1  4 13 11 16  0  9 10]
[ 7 14  3  6  8  5 12 15  2  1  4 13 11 16  9 10  0]
```

jakevdp:
I see - thanks for looking into that.

Perhaps we can fix this by wrapping the scipy function in the test so it behaves similarly to the expected behavior of jax's version. e.g. you could replace nans with `np.inf` or something like that to force scipy to produce the same results as JAX.

carlosgmartin:
@jakevdp Unfortunately, setting nan elements to inf or -inf inside a wrapper before passing it to osp_stats.rankdata still yields a different order sometimes. Should we perhaps simply use `contains_nans=[False]` for this test, instead?

jakevdp:
Sure, I think skipping NaNs is probably fine, since the function has unclear semantics for multiple nans

tests/scipy_stats_test.py
```diff
@@ -1129,5 +1129,30 @@ def scipy_mode_wrapper(a, axis=0, nan_policy='propagate', keepdims=None):
                             tol=tol)
     self._CompileAndCheck(lax_fun, args_maker, rtol=tol)
 
+  @jtu.sample_product(
+    [dict(shape=shape, axis=axis)
+      for shape in [(0,), (7,), (47, 8), (0, 2, 3), (10, 5, 21)]
+      for axis in [None, *range(len(shape))
+    ]],
+    dtype=jtu.dtypes.integer + jtu.dtypes.floating,
+    contains_nans=[False],
```
jakevdp, line 1138: If we're not going to use `contains_nans`, I'd just remove it here and below for simplicity.

jakevdp (approved):
Looks great, thanks!

PR #14355, opened 2023-02-08
tril/triu_indices: compute arrays at runtime

### LINKED ISSUE #14339: Implement `jnp.tril_indices`, `jnp.triu_indices` etc. in JAX.
Implement `jnp.tril_indices` etc. using XLA primitives instead of wrapping `numpy` implementations, so we don't materialise their output while staging.

jakevdp:
The current implementation of `jnp.tril_indices` and `jnp.triu_indices` computes the results at trace-time. This changes things to compute the values at runtime via JAX operations instead.

Fixes #14339

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -2673,9 +2673,33 @@ def _wrap_indices_function(f):
     return tuple(asarray(x) for x in f(*args, **kwargs))
   return wrapper
 
-tril_indices = _wrap_indices_function(np.tril_indices)
-triu_indices = _wrap_indices_function(np.triu_indices)
 mask_indices = _wrap_indices_function(np.mask_indices)
+
+
+def _triu_size(n, m, k):
+  if k < 0:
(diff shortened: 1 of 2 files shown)


mattjj (approved):

jax/_src/numpy/lax_numpy.py
```diff
@@ -2673,11 +2673,35 @@ def wrapper(*args, **kwargs):
...
+    return n * m - _triu_size(m, n, (1 - k))
+  elif k >= m:
+    return 0
+  else:
+    mk = _min(n, m - k)
+    return mk * (mk + 1) // 2 + mk * (m - k - mk)
+
+
+@_wraps(np.triu_indices)
+def triu_indices(n, k, m=None):
```
soraros, line 2690: `k` defaults to 0 per numpy. Also, maybe add type hint and/or jit compile?

jakevdp: Thanks - fixed in #14364

I don't think it's all that important to JIT compile here, particularly since every argument is static.

PR #14360, opened 2023-02-08
DOC: document alternatives to jnp.argpartition

jakevdp:
Part of #2079

Preview of new docs: https://jax--14360.org.readthedocs.build/en/14360/_autosummary/jax.numpy.argpartition.html


jakevdp:
Replaced by #14370

PR #14365, opened 2023-02-08
jax.numpy reductions: better validation of initial value

### LINKED ISSUE #14258: jnp.max/jnp.min allow non-scalar value for `initial`.
### Description

Hello!

I would like to report a sort of inverse feature request. Our code currently relies on the following logic working:

```python
def f(a, b):
  return jnp.max(a, axis=-1, initial=b, where=jnp.isfinite(a))

>> a = jnp.full((5,), fill_value=2)
>> b = jnp.array([1, 2, 3, 4, 5])
>> f(a, b)
Array([2, 2, 3, 4, 5], dtype=int32)
```

_Technically_ this shouldn't work since the `initial` argument of `jnp.max` is supposed to be a scalar. Numpy throws an exception `ValueError: Input object to FillWithScalar is not a scalar`, but it appears that JAX's implementation is more general and works as intended.

Can we continue to rely on this behavior?

0.4.2

GPU

### Additional system info

_No response_

### NVIDIA GPU info

_No response_

jakevdp:
Fixes #14258 via better input validation.

I didn't require `initial` to be a scalar here, because this may inadvertently break existing code. But I did require `initial` to not modify the result shape, as this seems unexpected.

### DIFF
--- a/jax/_src/numpy/reductions.py
+++ b/jax/_src/numpy/reductions.py
@@ -72,7 +72,7 @@ def _reduction(a: ArrayLike, name: str, np_fun: Any, op: ReductionOp, init_val: 
                axis: Axis = None, dtype: DTypeLike = None, out: None = None,
                keepdims: bool = False, initial: Optional[ArrayLike] = None,
                where_: Optional[ArrayLike] = None,
-               parallel_reduce: Optional[Callable[..., ArrayLike]] = None,
+               parallel_reduce: Optional[Callable[..., Array]] = None,
                promote_integers: bool = False) -> Array:
   bool_op = bool_op or op
   # Note: we must accept out=None as an argument, because numpy reductions delegate to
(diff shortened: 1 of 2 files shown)


mattjj:
Does the current code work correctly for non-scalar values of `initial`? (If we don't know for sure that it works, maybe we should consider breaking existing code, as maybe it's already broken in subtle ways! Just wanted to check.)

jakevdp:
It does work correctly, although we don't have any test coverage for it. I'd be open to requiring `initial` to be a scalar as in numpy – what do you think?

mattjj:
I like the idea of requiring `initial` to be a scalar. Not sure what that means for fixing downstream user code but it seems like the right long-term solution. Could be out of scope for this PR though!

mattjj (approved):

jakevdp:
Sounds good - I'll add that as a TODO because I suspect it would be a breaking change for some.

PR #14370, opened 2023-02-08
Implement jax.numpy.argpartition

### LINKED ISSUE #2079: Implement NumPy sorting routines

jakevdp:
Fixes https://github.com/google/jax/issues/2079

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3529,11 +3529,17 @@ def msort(a):
 
 
 @_wraps(np.partition, lax_description="""
-The jax version requires the ``kth`` argument to be a static integer rather than
-a general array. This is implemented via two calls to :func:`jax.lax.top_k`.
+The JAX version requires the ``kth`` argument to be a static integer rather than
+a general array. This is implemented via two calls to :func:`jax.lax.top_k`. If
+you're only accessing the top or bottom k values of the output, it may be more
+efficient to call :func:`jax.lax.top_k` directly.
+
+The JAX version differs from the NumPy version in the treatment of NaN entries;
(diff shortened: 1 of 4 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -3530,10 +3530,13 @@ def msort(a):
...
 @_wraps(np.partition, lax_description="""
 The jax version requires the ``kth`` argument to be a static integer rather than
-a general array. This is implemented via two calls to :func:`jax.lax.top_k`.
+a general array. This is implemented via two calls to :func:`jax.lax.top_k`. If
+you're only accessing the top or bottom k values of the output, it may be more
+efficient to call :func:`jax.lax.top_k` directly.
 """)
 @partial(jit, static_argnames=['kth', 'axis'])
 def partition(a: ArrayLike, kth: int, axis: int = -1) -> Array:
+  # TODO(jakevdp): handle NaN values like numpy.
```
mattjj, line 3539: Should we document this difference in behavior?

mattjj (approved):
Nice!

jax/_src/numpy/lax_numpy.py
jakevdp: Good idea - done!

jakevdp:
With this I think we can close Issue #70!!

PR #14373, opened 2023-02-08
[shard-map] if check_rep=False, don't call rep rules in eager

mattjj:


yashk2810 (approved):

yashk2810:
Thank you for fixing!

PR #14379, opened 2023-02-09
[shard-map] add vmap spmd_axis_name support, fix vmap rule bug

mattjj:


jax/experimental/shard_map.py
```diff
@@ -701,18 +702,25 @@ def _shard_map_batch(
   if all(bdim is batching.not_mapped for bdim in in_dims):
     return prim.bind(fun, *in_vals, mesh=mesh, in_names=in_names,
                      out_names_thunk=out_names_thunk, check_rep=check_rep)
-  if trace.spmd_axis_name is not None:
-    raise NotImplementedError  # TODO add named axis to specs
   if any(isinstance(d, batching.ConcatAxis) for d in in_dims):
     raise NotImplementedError
   fun, out_dims = batching.batch_subtrace(fun, trace.main, tuple(in_dims))
-  new_in_names = [{ax + (d is not batching.not_mapped and ax <= d): names[ax]  # type: ignore
+  new_in_names = [{ax + (d is not batching.not_mapped and d <= ax): names[ax]  # type: ignore
```
mattjj, line 708: Oops. I think `ax <= d` was wrong before.

In these lines we want to account for the fact that (for each arg and corresponding `names` dict) we're inserting a new axis at position `d` (if `d` is not `batching.not_mapped`), yet the keys in the dict refer to the old (pre-insertion) axes. So we want to bump each key `ax` up by 1 if it happens at or after the index at which we are inserting a new axis.

mattjj: Compare also to `BatchTrace.process_map`.

sharadmv (approved):

PR #14384, opened 2023-02-09
simpler pretty-print for pjit, tweak custom pp rule signature

mattjj:
Code:
```python
import jax
print(jax.make_jaxpr(jax.jit(lambda x: x[100]))(jax.numpy.arange(100)))
```

Before:
image

After:
image

(~Separately, I'm confused as to why this is a `gather` rather than something simpler like a `dynamic_slice`...~ aha, we only do a `dynamic_slice` when it's statically provable to be in-bounds!)

### DIFF
--- a/jax/_src/pjit.py
+++ b/jax/_src/pjit.py
@@ -24,7 +24,7 @@ import threading
 import warnings
 
 import jax
-from jax import core
+from jax._src import core
 from jax import stages
 from jax.errors import JAXTypeError
 from jax.experimental.global_device_array import GlobalDeviceArray as GDA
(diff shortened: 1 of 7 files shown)


yashk2810 (approved):

jax/_src/pjit.py
```diff
@@ -1841,6 +1841,29 @@ def _resource_typing_pjit(avals, params, source_info, resource_env, named_axis_r
...
+  params = dict(eqn.params)
+  del params['inline']
+  if not any(params['donated_invars']):
+    del params['donated_invars']
+  if all(p == pxla._PositionalSemantics.GLOBAL
+         for p in params['in_positional_semantics']):
+    del params['in_positional_semantics']
+  if params['out_positional_semantics'] == pxla._PositionalSemantics.GLOBAL:
+    del params['out_positional_semantics']
+  if all(s == _UNSPECIFIED for s in params['in_shardings']):
```
yashk2810, line 1854: Can you use `pxla._is_unspecified`?

mattjj: Sure!

PR #14388, opened 2023-02-09
[sparse] implement autodiff rules for bcsr primitives

jakevdp:


jax/experimental/sparse/bcsr.py
```diff
@@ -63,6 +64,25 @@ def bcsr_sum_duplicates(mat: BCSR, nse: Optional[int] = None) -> BCSR:
   return BCSR.from_bcoo(bcoo.bcoo_sum_duplicates(mat.to_bcoo(), nse=nse))
 
 
+def _bcsr_batch_dims_to_front(batched_args, batch_dims, spinfo, batch_size=None):
+  data, indices, indptr = batched_args
+  data_bdim, indices_bdim, indptr_bdim = batch_dims
+  n_batch = indices.ndim - 1 + bool(indices_bdim is None)
```
skye, line 70: Do you mean to cast to int() here?

jax/experimental/sparse/bcsr.py
```diff
@@ -231,17 +251,45 @@ def _bcsr_fromdense_abstract_eval(mat, *, nse, n_batch, n_dense, index_dtype):
...
-  if batch_dims != (0,):
-    raise NotImplementedError(f"{batch_dims=}")
-  new_n_batch = n_batch + 1
-  n_sparse = M.ndim - n_dense - new_n_batch
-  if n_sparse != 2:
-    raise ValueError("_bcsr_fromdense_batching_rule: must have 2 sparse "
-                     f"dimensions but {n_sparse} is given.")
-  return _bcsr_fromdense(M, nse=nse, n_batch=new_n_batch, n_dense=n_dense,
-                         index_dtype=index_dtype), (0, 0, 0)
+  bdim, = batch_dims
```
skye, line 254: User error can't cause this to fail right? Like you could assert it's len == 1 ?

skye (approved):

jax/experimental/sparse/bcsr.py
jakevdp: I think user input is validated at a higher level in vmap, so at this point it's safe to assume the arguments have the expected shape.

jax/experimental/sparse/bcsr.py
jakevdp: Done, thanks!

PR #14394, opened 2023-02-10
[typing] define additional methods & properties on jax.Array

jakevdp:
These are the methods that are only valid for actual materialized arrays (i.e. not Tracers) In order to simplify the experience for users, we want to maintain only a single jax.Array type, so we define all methods here and raise explicit errors on Tracer instances.

Part of #12049


jax/_src/basearray.pyi
```diff
@@ -186,3 +186,24 @@ class Array(abc.ABC):
...
+  # Methods defined on ArrayImpl, but not on Tracers
+  def addressable_data(self, index: int) -> Array: ...
+  def block_until_ready(self) -> Array: ...
+  def copy_to_host_async(self) -> None: ...
+  def delete(self) -> None: ...
+  def device(self) -> Device: ...
+  @property
+  def device_buffer(self) -> Array: ...
+  @property
+  def device_buffers(self) -> Sequence[Array]: ...
```
yashk2810, line 199: Do we want to expose these? These methods are deprecated `device_buffer` and `device_buffers`

jakevdp: Makes sense, yeah we should probably leave these out. Are there any other deprecated methods/properties in this list?

yashk2810: Rest of them are good!

PR #14358, opened 2023-02-08
added jax.scipy.special.spence function 

lucashofer:
I added jax.scipy.special.spence to JAX. The implementation below is based on the Cephes library algorithm which SciPy uses, but written in JAX code. The author of the Cephes library has already given permission with the JAX license.

PR #14408, opened 2023-02-10
added scipy special spence

lucashofer:
Hi

I had troubles with git and my previous pull request so I've fixed it all here with a single commit including all the changes mentioned in

[#14358](https://github.com/google/jax/pull/14358)

Sorry for the hassle, it's my first time contributing to an open source project.

### DIFF
--- a/jax/_src/scipy/special.py
+++ b/jax/_src/scipy/special.py
@@ -224,6 +224,9 @@ _polygamma.defjvps(None, lambda g, ans, n, x: lax.mul(g, _polygamma(n + 1, x)))
 
 # Functions "ndtr" and "ndtri" are derived from calculations made in:
 # https://root.cern.ch/doc/v608/SpecFuncCephesInv_8cxx_source.html
+# The "spence" function is also based on the Cephes library with
+# the corresponding spence.c file located in the tarball:
+# https://netlib.org/cephes/misc.tgz
 # In the following email exchange, the author gives his consent to redistribute
 # derived works under an Apache 2.0 license.
 #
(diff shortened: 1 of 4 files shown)


jakevdp (approved):
Thanks!

jax/_src/scipy/special.py
```diff
@@ -1518,3 +1521,87 @@ def exp1(x: ArrayLike, module='scipy.special') -> Array:
...
+    TypeError if elements of array `z` are not in (float32, float64).
+  Notes:
+  "There is a different convention which defines Spence's function by the
+  integral:
+  .. math::
+    \begin{equation}
+    -\int_0^z \frac{\log(1 - t)}{t}dt
+    \end{equation}
+  this is our spence(1 - z).
+  """
```
jakevdp, line 1601: It looks like this docstring is misformatted, which is causing [errors](https://readthedocs.org/projects/jax/builds/19448599/) in the documentation build:
```
/home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/14408/jax/_src/scipy/special.py:docstring of jax._src.scipy.special.spence:4: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/14408/jax/_src/scipy/special.py:docstring of jax._src.scipy.special.spence:10: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/14408/jax/_src/scipy/special.py:docstring of jax._src.scipy.special.spence:22: ERROR: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/jax/checkouts/14408/jax/_src/scipy/special.py:docstring of jax._src.scipy.special.spence:25: WARNING: Literal block ends without a blank line; unexpected unindent.
```
I think the issue is that you need a blank line before a `.. math::` block?

jax/_src/scipy/special.py
```diff
@@ -1518,3 +1521,87 @@ def exp1(x: ArrayLike, module='scipy.special') -> Array:
...
+def spence(x: Array) -> Array:
+  r"""Spence's function, also known as the dilogarithm for real values.
+  It is defined to be:
+  .. math::
+    \begin{equation}
+    \int_1^z \frac{\log(t)}{1 - t}dt
+    \end{equation}
+
+  Unlike the SciPy implementation, this is only defined for positive
+    real values of `z`. For negative values, `NaN` is returned.
```
jakevdp, line 1585: This is an unexpected indentation

jax/_src/scipy/special.py
```diff
@@ -1518,3 +1521,87 @@ def exp1(x: ArrayLike, module='scipy.special') -> Array:
...
+  Args:
```
jakevdp, line 1586: I think you also need a blank line before all of these sections (`Args:`, `Returns:`, `Raises:`)

lucashofer:
Hi Jake,

I've updated the docstring and tested it out on a private readthedocs package.

However, it also looks likes one other build failed and I can't replicate this error when I run the tests locally. I'm not too familiar with the jax.numpy type promotion, but looking at the other functions in:

[https://github.com/google/jax/blob/main/jax/_src/scipy/special.py](https://github.com/google/jax/blob/main/jax/_src/scipy/special.py)

and it looks like rather than what I'm doing with the function input:

```
def spence(x: Array) -> Array:
     x = jnp.asarray(x)
```

they use something like this:

```
def spence(x: Array) -> Array:
     x, = _promote_args_inexact("spence", x)
```

Does this look like it would solve the build/test issue?

jakevdp:
You should be able to replicate the failure locally using
```
$ JAX_ENABLE_X64=1 pytest -n auto tests/lax_scipy_test.py -k spence
```

jax/_src/scipy/special.py
```diff
@@ -1518,3 +1521,87 @@ def exp1(x: ArrayLike, module='scipy.special') -> Array:
...
+                  ])
+
+  B = jnp.array([6.90990488912553276999E-4,
+                  2.54043763932544379113E-2,
+                  2.82974860602568089943E-1,
+                  1.41172597751831069617E0,
+                  3.63800533345137075418E0,
+                  5.03278880143316990390E0,
+                  3.54771340985225096217E0,
+                  9.99999999999999998740E-1, ])
```
jakevdp, line 1544: To fix the type promotion issue, I think what you need here is to pass `dtype=w.dtype` for both the `A` and `B` arrays. Otherwise in X64 mode these will be float64, and if `w` is float32 it will cause an error in strict type promotion mode, which is what we use when running tests.

lucashofer: I've implemented this and it now passes the tests with the ``JAX_ENABLE_X64`` enabled. Thanks!

jakevdp (approved):
Nice - let's hope tests pass now!

PR #14456, opened 2023-02-13
Expose jax.typing & update docs

jakevdp:
Preview here: https://jax--14456.org.readthedocs.build/en/14456/jax.typing.html

### DIFF
--- a/jax/_src/basearray.py
+++ b/jax/_src/basearray.py
@@ -17,30 +17,20 @@
 import abc
 
 class Array(abc.ABC):
-  """Experimental Array base class for JAX
+  """Array base class for JAX
 
-  `jax.Array` is meant as the future public interface for instance checks and
-  type annotation of JAX array objects. JAX Array object types are currently in
-  flux, and this class only fully supports the new `jax.experimental.Array`, which
-  will soon replace the old-style {class}`DeviceArray`, {class}`ShardedDeviceArray`,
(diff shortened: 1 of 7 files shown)


jax/_src/basearray.py
```diff
@@ -17,30 +17,20 @@
...
+  arrays and tracers. Its main applications are in instance checks and type annotations;
+  for example::
+
+    x = jnp.arange(5)
+    isinstance(x, jax.Array)  # returns True both inside and outside traced functions.
+
+    def f(x: Array) -> Array:  # type annotations are valid for traced and non-traced types.
+      return x
+
+  Because `jax.Array` is effectively an abstact base class, the main documentation for
```
hawkinsp, line 32: sp. abstract

I'm actually a bit surprised to not find a list of methods, at least, even if they link to the numpy functions.

I went looking on the `jax.numpy` docs and didn't really find what I wanted there either.

hawkinsp (approved):
This text is now old, but we can probably clear that up in a subsequent PR.
https://jax--14456.org.readthedocs.build/en/14456/jax.numpy.html#jax-devicearray

jax/_src/basearray.py
jakevdp: Yeah, we need to figure out how to get a list of array methods in the HTML docs; there was a chat thread on that this morning. I'm going to work on that in a followup.

PR #14479, opened 2023-02-15
MeshComputation.cost_analysis() isn't implemented with PJRT C API.

skye:
This was caught via PJitTest.testLowerCostAnalysis (https://github.com/google/jax/blob/e74852f79695c9f2fcf06b4c8400b176bb899766/tests/pjit_test.py#L998). We don't need to change the test because NotImplementedError is already caught in Lowered.cost_analysis:
https://github.com/google/jax/blob/e74852f79695c9f2fcf06b4c8400b176bb899766/jax/_src/stages.py#L659-L660

### DIFF
--- a/jax/_src/interpreters/pxla.py
+++ b/jax/_src/interpreters/pxla.py
@@ -3259,8 +3259,13 @@ class MeshComputation(stages.XlaLowering):
     return self._executable
 
   def cost_analysis(self) -> Dict[str, float]:
-    return xe.hlo_module_cost_analysis(self.compile_args["backend"],
-                                       self.hlo().as_hlo_module())
+    backend = self.compile_args["backend"]
+    if xb.using_pjrt_c_api(backend):
+      raise NotImplementedError(
+          "Lowered.cost_analysis not implemented on platform "
+          f"'{backend.platform}'. Use compile().cost_analysis() for "
+          "post-compilation cost estimates.")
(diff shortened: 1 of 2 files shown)


jax/_src/interpreters/pxla.py
```diff
@@ -3259,8 +3259,12 @@ def compile(self,
     return self._executable
 
   def cost_analysis(self) -> Dict[str, float]:
-    return xe.hlo_module_cost_analysis(self.compile_args["backend"],
-                                       self.hlo().as_hlo_module())
+    backend = self.compile_args["backend"]
+    if xb.using_pjrt_c_api(backend):
+      raise NotImplementedError(
+          "Lowered.cost_analysis not implemented on platform "
+          f"'{backend.platform}'. Use compile() to access cost_analysis.")
```
froystig, line 3266: When both are implemented, we don't expect `Lowered.cost_analysis` and `Compiled.cost_analysis` to reflect quite the same costs. The former is an estimate prior to compiler optimizations, whereas the latter accounts for them. Maybe tweak to something like this?
```suggestion
          f"'{backend.platform}'. Use compile().cost_analysis() for "
          "post-compilation cost estimates.")
```

froystig (approved):

jax/_src/interpreters/pxla.py
skye: Done, thanks!

PR #14486, opened 2023-02-15
Added implementation of `null_space`

### LINKED ISSUE #14350: Implement the `null_space` function from `scipy.linalg`
Please:

I wanted to use the `null_space` function in `jax` but it's not implemented yet. The code is very straightforward because it is based on the SVD. I can try to submit a PR relatively soon.

zaccharieramzi:
This fixes #14350 .

For the unit tests I just copy-pasted and corrected the ones from `scipy` (for some reason they were not correct).

As @jakevdp was mentioning, `null_space` is not jittable, but I am not sure how to signal that to the user.

### DIFF
--- a/jax/_src/scipy/linalg.py
+++ b/jax/_src/scipy/linalg.py
@@ -14,6 +14,7 @@
 
 
 from functools import partial
+import operator
 
 import numpy as np
 import scipy.linalg
@@ -24,6 +25,7 @@ from typing import cast, overload, Any, Literal, Optional, Tuple, Union
 import jax
 from jax import jit, vmap, jvp
 from jax import lax
+from jax._src import core
 from jax._src import dtypes
 from jax._src.lax import linalg as lax_linalg
 from jax._src.lax import qdwh
@@ -1041,3 +1043,46 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
       (nrows,), (1,), 'VALID', dimension_numbers=('NTC', 'IOT', 'NTC'),
       precision=lax.Precision.HIGHEST)[0]
   return jnp.flip(patches, axis=0)
+
+
+_NULLSPACE_DOC = """\
+Because the size of the output of ``null_space`` is data-dependent, the function is not
+typically compatible with JIT. The JAX version adds the optional ``size`` argument which
+must be specified statically for ``jax.scipy.linalg.null_space`` to be used within some of JAX's
+transformations.
+"""
+_NULLSPACE_EXTRA_PARAMS = """
+size : int, optional
+    If specified, ``size`` elements of the null space are returned. If there are
+    fewer null space elements than ``size`` indicates, the return value will be padded with constant
+    0 vectors that belong by nature to the null space.
+"""
+
+@_wraps(scipy.linalg.null_space, lax_description=_NULLSPACE_DOC, extra_params=_NULLSPACE_EXTRA_PARAMS)
+def null_space(A: ArrayLike, rcond: Optional[Union[float, Array]] = None,
+               size: Optional[int] = None) -> Array:
+  u, s, vh = svd(A, full_matrices=True)
+  M, N = u.shape[0], vh.shape[1]
+  if rcond is None:
+    rcond = jnp.finfo(s.dtype).eps * max(M, N)
+  rcond = jnp.array(rcond).astype(s.dtype)
+  tol = jnp.amax(s) * rcond
+  n_null_space = vh.shape[0] - jnp.sum(s > tol, dtype=int)
+  size_int = core.concrete_or_error(
+    operator.index,
+    n_null_space if size is None else size,
+    "The size argument of null_space must be statically specified "
+    "to use null_space within jax.jit or other transformations.")
+  if size_int < 0:
+    raise ValueError("size must be positive")
+  max_size = min(size_int, vh.shape[0])
+  Q = vh[vh.shape[0] - max_size:]
+  # now we need to make sure to remove all vectors in Q that are not in the null space
+  # but were selected just because we need an array of size `size`
+  fill_value = 0
+  mask = (jnp.arange(Q.shape[0]) < max_size - n_null_space)[:, None]
+  Q = jnp.where(mask, fill_value, Q)
+  # we need to pad Q with 0s if size > max_size
+  Q = jnp.pad(Q, ((size_int - max_size, 0), (0, 0)), constant_values=fill_value)
+  Q = Q.T.conj()
+  return Q
--- a/tests/linalg_test.py
+++ b/tests/linalg_test.py
@@ -1619,6 +1619,46 @@ class ScipyLinalgTest(jtu.JaxTestCase):
       [2, 1, 4, 5],
       [3, 2, 1, 4]], dtype=np.float32))
 
+  @jtu.sample_product(
+    shape=[[2, 1], [1, 2], [2, 5], [5, 2], [4, 20], [20, 4]],
+    dtype=float_types + complex_types,
+    size=[None, 2, 9])
+  def testNullSpace(self, shape, dtype, size):
+    # test compiled version
+    rng = jtu.rand_default(self.rng())
+    X = rng(shape, dtype)
+    def scipy_fun(x):
+      result = osp.linalg.null_space(x)
+      if size is None:
+        return result
+      elif size <= result.shape[1]:
+        return result[:, -size:]
+      else:
+        all_zeros = np.zeros((result.shape[0], size), dtype=x.dtype)
+        all_zeros[:, size-result.shape[1]:] = result
+        return all_zeros
+    args_maker = lambda: [X]
+    jnp_fun = partial(jsp.linalg.null_space, size=size)
+    self._CheckAgainstNumpy(scipy_fun, jnp_fun, args_maker, check_dtypes=False)
+    if size is not None:
+      self._CompileAndCheck(jnp_fun, args_maker)
+    else:
+      err, msg = jax.errors.ConcretizationTypeError, r"Abstract tracer value encountered where concrete value is expected"
+      with self.assertRaisesRegex(err, msg):
+        self._CompileAndCheck(jnp_fun, args_maker)
+
+  @jtu.sample_product(
+    base_shape=[10, 100],
+    dtype=float_types + complex_types,
+    rcond=[1e-3, 1e-6])
+  def testNullSpaceDifficult(self, base_shape, dtype, rcond):
+    rng = jtu.rand_default(self.rng())
+    X = rng((base_shape, 5), dtype) @ rng((5, base_shape), dtype)
+    X = X + 1e-4 * rng((base_shape, 1), dtype) @ rng((1, base_shape), dtype)
+    args_maker = lambda: [X, rcond]
+    self._CheckAgainstNumpy(osp.linalg.null_space,
+                            jsp.linalg.null_space, args_maker)
+
 
 class LaxLinalgTest(jtu.JaxTestCase):
   """Tests for lax.linalg primitives."""
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -35,6 +35,7 @@ jax.scipy.linalg
    lu
    lu_factor
    lu_solve
+   null_space
    polar
    qr
    rsf2csf
--- a/jax/scipy/linalg.py
+++ b/jax/scipy/linalg.py
@@ -30,6 +30,7 @@ from jax._src.scipy.linalg import (
   lu as lu,
   lu_factor as lu_factor,
   lu_solve as lu_solve,
+  null_space as null_space,
   polar as polar,
   qr as qr,
   rsf2csf as rsf2csf,


jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1041,15 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
       (nrows,), (1,), 'VALID', dimension_numbers=('NTC', 'IOT', 'NTC'),
       precision=lax.Precision.HIGHEST)[0]
   return jnp.flip(patches, axis=0)
+
+
+@_wraps(scipy.linalg.null_space)
+def null_space(A: ArrayLike, rcond: Optional[float] = None) -> Array:
```
jakevdp, line 1047: One thing we could do here is add an optional keyword-only `size` argument (similar to what we use in [`jnp.nonzero`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.nonzero.html) and other APIs) that allows the user to specify a static size if they want the operation to be JIT-compatible. What do you think?

tests/linalg_test.py
```diff
@@ -1619,6 +1619,52 @@ def testToeplitzConstructionWithKnownCases(self):
       [2, 1, 4, 5],
       [3, 2, 1, 4]], dtype=np.float32))
 
+  @jtu.sample_product(
+    size=[1, 2, 3, 10, 100],
+    dtype=float_types + complex_types)
+  def testNullSpace(self, size, dtype):
+    rng = jtu.rand_default(self.rng())
```
jakevdp, line 1626: For tests, we generally treat the `jsp_linalg` function as the ground truth and check that JAX's output matches. That can make the tests simpler and more comprehensive. What do you think?

jax/_src/scipy/linalg.py
zaccharieramzi: Just saw what was done in `jnp.nonzero`, will do the same

tests/linalg_test.py
zaccharieramzi: Do you mean `scipy.linalg`? Otherwise I am not sure I understand.
But if so, I can definitely change the test to make sure it tests against a ground truth

jakevdp: Yes, sorry I mean treat scipy as the ground truth

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1041,43 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+"""
+_NULLSPACE_EXTRA_PARAMS = """
+size : int, optional
+    If specified, ``size`` elements of the null space are returned. If there are
+    fewer null space elements than ``size`` indicates, the return value will be padded with constant
+    ``fill_value`` vectors.
+fill_value : array_like, optional
+    When ``size`` is specified and there are fewer null space elements than the indicated number of elements,
+    the return value will be padded with constant ``fill_value`` vectors. If ``fill_value`` is not specified,
+    it defaults to zero (indeed constant 0 vectors are always in the null space).
```
jakevdp, line 1060: I wonder if we need `fill_value` here? I think a more natural fill choice is to use the already computed eigenvectors. Maybe simpler to not add `fill_value` at all?

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1041,43 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+@_wraps(scipy.linalg.null_space, lax_description=_NULLSPACE_DOC, extra_params=_NULLSPACE_EXTRA_PARAMS)
+def null_space(A: ArrayLike, rcond: Optional[float] = None,
+               size: Optional[int] = None, fill_value: Union[None, ArrayLike, Tuple[ArrayLike]] = None) -> Array:
+  u, s, vh = svd(A, full_matrices=True)
+  M, N = u.shape[0], vh.shape[1]
+  if rcond is None:
+    rcond = jnp.finfo(s.dtype).eps * max(M, N)
+  tol = jnp.amax(s) * rcond
+  num = jnp.sum(s > tol, dtype=int)
+  Q = vh[num:, :].T.conj()
```
jakevdp, line 1072: This will error under JIT. We would need to update the logic so that we don't perform this operation unless `size` is `None`

jakevdp (changes requested):

tests/linalg_test.py
zaccharieramzi: done

jax/_src/scipy/linalg.py
zaccharieramzi: yes all tests passing even with JIT, but maybe I am missing something

jax/_src/scipy/linalg.py
jakevdp: In that case, should the fill value default to zero and not be user-specifiable? In what situations would a fill value other than zero be relevant?

jax/_src/scipy/linalg.py
jakevdp: The updated version of this looks fine.

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1041,49 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+def null_space(A: ArrayLike, rcond: Optional[float] = None,
+               size: Optional[int] = None, fill_value: Union[None, ArrayLike, Tuple[ArrayLike]] = None) -> Array:
+  u, s, vh = svd(A, full_matrices=True)
+  M, N = u.shape[0], vh.shape[1]
+  if rcond is None:
+    rcond = jnp.finfo(s.dtype).eps * max(M, N)
+  tol = jnp.amax(s) * rcond
+  num = jnp.sum(s > tol, dtype=int)
+  if size is None:
+    size = int(num)
```
jakevdp, line 1073: This is the line that will error if `size=None` is used under `jit`. It may be helpful to use a more informative error message; something like this: https://github.com/google/jax/blob/fdc8864d9b0ade449a66933ee7fe0bb3dd27ede3/jax/_src/numpy/lax_numpy.py#L1371-L1373

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1041,49 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+  elif isinstance(fill_value, tuple):
+    fill_value = jnp.asarray(fill_value)
+  else:
+    fill_value = jnp.asarray(fill_value)
+  Q = lax.dynamic_slice_in_dim(vh, size, vh.shape[0] - size, axis=0)
+  # now we need to make sure to remove all vectors in Q that are not in the null space
+  # but were selected just because we need an array of size `size`
+  filler = jnp.ones_like(Q) * fill_value
+  mask = (jnp.arange(Q.shape[0]) < size-num)[:, None]
+  Q = jnp.where(mask, filler, Q)
```
jakevdp, line 1087: I think using `fill_value` directly in place of `filler` would be give the same result

tests/linalg_test.py
```diff
@@ -1619,6 +1619,39 @@ def testToeplitzConstructionWithKnownCases(self):
       [2, 1, 4, 5],
       [3, 2, 1, 4]], dtype=np.float32))
 
+  @jtu.sample_product(
+    size=[1, 2, 3, 10, 100],
```
jakevdp, line 1623: Should also test `size=None`

tests/linalg_test.py
```diff
@@ -1619,6 +1619,39 @@ def testToeplitzConstructionWithKnownCases(self):
...
+    dtype=float_types + complex_types)
+  def testNullSpace(self, size, dtype):
+    rng = jtu.rand_default(self.rng())
+    shape = (2, size)
+    X = rng(shape, dtype)
+    args_maker = lambda: [X]
+    self._CheckAgainstNumpy(jtu.promote_like_jnp(osp.linalg.null_space),
+                            jsp.linalg.null_space, args_maker)
```
jakevdp, line 1631: Add here
```python
if size is not None:
  self._CompileAndCheck(jsp.linalg.null_space, args_maker)
```
This would cover the jit-compilation checks, and we may be able to delete everything below.

tests/linalg_test.py
```diff
@@ -1619,6 +1619,39 @@ def testToeplitzConstructionWithKnownCases(self):
       [2, 1, 4, 5],
       [3, 2, 1, 4]], dtype=np.float32))
 
+  @jtu.sample_product(
+    size=[1, 2, 3, 10, 100],
+    dtype=float_types + complex_types)
+  def testNullSpace(self, size, dtype):
+    rng = jtu.rand_default(self.rng())
+    shape = (2, size)
```
jakevdp, line 1627: Can we iterate over various shapes rather than hard-coding them?

jakevdp (changes requested):

jax/_src/scipy/linalg.py
zaccharieramzi: oh right will do so

jax/_src/scipy/linalg.py
zaccharieramzi: I think you are right, will correct

tests/linalg_test.py
zaccharieramzi: ah here `size` does not correspond to the `size` argument of the `null_space` function. I was just reusing the test from scipy originally, and didnt correct this discrepancy, I will now.

tests/linalg_test.py
zaccharieramzi: I was trying to keep the structure of the original scipy test here, do you think it makes sense for the potential readers?
Otherwise I will just iterate ofc

tests/linalg_test.py
zaccharieramzi: see my comment above

jax/_src/scipy/linalg.py
zaccharieramzi: yes I think so will change to a value of 0 and clarify in the docs

tests/linalg_test.py
jakevdp: As an example, see this test: https://github.com/google/jax/blob/64145393b2c65bb011ca10d41314a8a06c025cb7/tests/linalg_test.py#L803-L823

We don't test here that the output has the properties expected of an inverse. We use scipy as the ground truth, and check against that.

jax/_src/scipy/linalg.py
zaccharieramzi: done

jax/_src/scipy/linalg.py
zaccharieramzi: done

tests/linalg_test.py
zaccharieramzi: And I did make sure to test that compilation with `size=None` returns a meaningful error

tests/linalg_test.py
zaccharieramzi: sure will update!

zaccharieramzi: done

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1042,41 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+  if rcond is None:
+    rcond = jnp.finfo(s.dtype).eps * max(M, N)
+  tol = jnp.amax(s) * rcond
+  num = jnp.sum(s > tol, dtype=int)
+  if size is None:
+    size = core.concrete_or_error(int, num,
+    "The size argument of null_space must be statically specified "
+    "to use null_space within jitted JAX transformations.")
+  else:
+    size = min(size, vh.shape[0])
```
jakevdp, line 1074: We should call `core.concrete_or_error` even if `size` is a user input, as it may not be concrete.

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1042,41 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+  Q = lax.dynamic_slice_in_dim(vh, size, vh.shape[0] - size, axis=0)
```
jakevdp, line 1075: Since `size` is never dynamic, we can use straightforward indexing here:
```
Q = vh[vh.shape[0] - size:]
```
(not `[-size:]` because `size` might be zero)

tests/linalg_test.py
```diff
@@ -1619,6 +1619,38 @@ def testToeplitzConstructionWithKnownCases(self):
...
 
+  @jtu.sample_product(
+    base_shape=[1, 2, 3, 10, 100],
+    dtype=float_types + complex_types)
+  def testNullSpace(self, base_shape, dtype):
+    for shape in [
+      (2, base_shape),
+      (base_shape, 2),
+      (1 + base_shape // 2, base_shape),
+    ]:
```
jakevdp, line 1630: Let's try to keep iterations in the test parameterization, not in the test itself. The reason is, if any of these fails it prevents subsequent iterations from running. It's better to have all test failures reported at once.

zaccharieramzi: makes total sense will correct

jax/_src/scipy/linalg.py
zaccharieramzi: right makes sense

jax/_src/scipy/linalg.py
zaccharieramzi: ah I see I hadnt realized, correcting

jax/_src/scipy/linalg.py
zaccharieramzi: but does it make sense to allow `size==0`?

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1042,48 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+  num = jnp.sum(s > tol, dtype=int)
+  n_null_space = vh.shape[0] - num
+  if size is None:
+    size = core.concrete_or_error(int, n_null_space,
+    "The size argument of null_space must be statically specified "
+    "to use null_space within jitted JAX transformations.")
+  else:
+    size = min(size, vh.shape[0])
+    size = core.concrete_or_error(
+      lambda x: min(x, vh.shape[0]),
```
jakevdp, line 1077: This is getting kind of complicated... how about something like this?
```python
n_null_space = jnp.sum(s <= tol)
if size is None:
  size = n_null_space
size = core.concrete_or_error(operator.index, size,
    "The size argument of null_space must be statically specified "
    "to use null_space within jax.jit or other transformations.")
if size < 0:
  raise ValueError("size must be positive")
size = min(size, vh.shape[0])
```

tests/linalg_test.py
```diff
@@ -1620,6 +1620,51 @@ def testToeplitzConstructionWithKnownCases(self):
       [3, 2, 1, 4]], dtype=np.float32))
 
 
+  base_shapes_small = [1, 2, 3]
+  base_shapes_big = [10, 100]
+  null_space_shapes = list(itertools.chain.from_iterable([[
+    (2, base_shape),
+    (base_shape, 2),
+    (1 + base_shape // 2, base_shape),
+  ] for base_shape in base_shapes_small + base_shapes_big]))
```
jakevdp, line 1629: We shouldn't be defining unnecessary class-level variables. You could put this logic inline, but that's probably not necessary. Couldn't we just iterate over a selection of shapes rather than constructing shapes with a particular structure? `null_space` should work for arbitrary 2D shapes, right? What if we do something like
```
@jtu.sample_product(
    shape=[[2, 1], [1, 2], [2, 5], [5, 2], [4, 20], [20, 4]],
    dtype=float_types + complex_types)
```
Do you think that would be sufficient test coverage?

tests/linalg_test.py
```diff
@@ -1620,6 +1620,51 @@ def testToeplitzConstructionWithKnownCases(self):
...
+  @jtu.sample_product(
+    shape=null_space_shapes,
+    dtype=float_types + complex_types,
+    size=[None, 2, 9])
+  def testNullSpaceCompiled(self, shape, dtype, size):
+    # test compiled version
+    rng = jtu.rand_default(self.rng())
+    X = rng(shape, dtype)
+    if size != None:
+      jax.jit(jsp.linalg.null_space, static_argnames=('size'))(X, size=size)
```
jakevdp, line 1649: We should test that the output is as expected.

One suggestion along these lines: use `CheckAgainstNumpy` but define a scipy function wrapper that returns the expected padding. For example, here's how we test `jnp.nonzero` against numpy with the `size` argument: https://github.com/google/jax/blob/7e001d842e1cbfdbf991a4bd9b236012cc40fba4/tests/lax_numpy_test.py#L308-L315

jakevdp (changes requested):

jax/_src/scipy/linalg.py
zaccharieramzi: fair this is indeed simpler

tests/linalg_test.py
zaccharieramzi: yes I think it would be, I just didn't want to make any arbitrary decisions and I thought that the less arbitrary was to just reuse the shapes scipy was testing, but I agree that going for something simpler is also a good point

tests/linalg_test.py
zaccharieramzi: ok I see will adapt my test to have something exactly like this test

zaccharieramzi: done

tests/linalg_test.py
zaccharieramzi: done

jax/_src/scipy/linalg.py
zaccharieramzi: so we can't go with `n_null_space = jnp.sum(s <= tol)` because the null singular vectors are not listed in `s`, but other than that done

tests/linalg_test.py
```diff
@@ -1620,6 +1620,55 @@ def testToeplitzConstructionWithKnownCases(self):
...
+        result = osp.linalg.null_space(x)
+        if size <= result.shape[1]:
+          return result[:, -size:]
+        else:
+          all_zeros = np.zeros((result.shape[0], size), dtype=x.dtype)
+          all_zeros[:, size-result.shape[1]:] = result
+          return all_zeros
+      args_maker = lambda: [X]
+      jnp_fun = partial(jsp.linalg.null_space, size=size)
+      self._CheckAgainstNumpy(scipy_fun, jnp_fun, args_maker, check_dtypes=False)
```
jakevdp, line 1652: You could put this `_CheckAgainstNumpy` outside the `if size is not None` block and it would basically cover everything that `testNullSpace` currently covers, right? Can we combine these two tests?

tests/linalg_test.py
```diff
@@ -1620,6 +1620,55 @@ def testToeplitzConstructionWithKnownCases(self):
...
+
+  @jtu.sample_product(
+    shape=[[2, 1], [1, 2], [2, 5], [5, 2], [4, 20], [20, 4]],
+    dtype=float_types + complex_types,
+    size=[None, 2, 9])
+  def testNullSpaceCompiled(self, shape, dtype, size):
+    # test compiled version
+    rng = jtu.rand_default(self.rng())
+    X = rng(shape, dtype)
+    if size != None:
```
jakevdp, line 1641: For comparisons with `None`, we should use object identity:
```
if size is not None
```

tests/linalg_test.py
```diff
@@ -1620,6 +1620,55 @@ def testToeplitzConstructionWithKnownCases(self):
...
+          all_zeros[:, size-result.shape[1]:] = result
+          return all_zeros
+      args_maker = lambda: [X]
+      jnp_fun = partial(jsp.linalg.null_space, size=size)
+      self._CheckAgainstNumpy(scipy_fun, jnp_fun, args_maker, check_dtypes=False)
+      self._CompileAndCheck(jnp_fun, args_maker)
+    else:
+      err, msg = jax.errors.ConcretizationTypeError, r"Abstract tracer value encountered where concrete value is expected"
+      with self.assertRaisesRegex(err, msg):
+        jax.jit(jsp.linalg.null_space, static_argnames=('size'))(X, size=size)
```
jakevdp, line 1657: Nit - this is basically `jit(jnp_fun)`, so we could simplify things by using that here.

jakevdp (changes requested):
Looking good! Just a couple more comments on streamlining the tests

tests/linalg_test.py
zaccharieramzi: ofc

tests/linalg_test.py
zaccharieramzi: agreed!! It obviously should be a single test, now that I understand the different parts it makes sense to group them into one
Fixed that

tests/linalg_test.py
zaccharieramzi: yup, corrected in the new combined test

jakevdp (changes requested):
This looks great! A couple last things:

- Please add this function to the list at https://github.com/google/jax/blob/main/docs/jax.scipy.rst#jaxscipylinalg so that it will be included in the HTML docs
- Please squash all your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests). Let me know if you need help with how to do that.

Thanks for sticking with this through all the revisions!

zaccharieramzi:
Sure no problem, thank you for your patience!
I just have issues with `mypy` that I don't see exactly how to fix in the best way possible. It seems like if I want to fix them I will make the function more difficult to read, which we tried to avoid during the whole revision process

jax/_src/scipy/linalg.py
```diff
@@ -1041,3 +1043,46 @@ def toeplitz(c: ArrayLike, r: Optional[ArrayLike] = None) -> Array:
...
+  u, s, vh = svd(A, full_matrices=True)
+  M, N = u.shape[0], vh.shape[1]
+  if rcond is None:
+    rcond = jnp.finfo(s.dtype).eps * max(M, N)
+  rcond = jnp.array(rcond).astype(s.dtype)
+  tol = jnp.amax(s) * rcond
+  n_null_space = vh.shape[0] - jnp.sum(s > tol, dtype=int)
+  if size is None:
+    size = n_null_space
+  size = core.concrete_or_error(operator.index, size,
```
jakevdp, line 1073: Maybe something like this:
```python
size_int = core.concrete_or_error(operator.index, n_null_space if size is None else size,
  ...
```
then use `size_int` below.

zaccharieramzi: seems to be working : sorry I am really not used to typing so I don't know exactly how to deal with it in a straightforward way.

zaccharieramzi:
@jakevdp For the squashing maybe I can wait until all the tests are green before doing it in order to avoid doing it multiple times

jakevdp:
Alright, looks good! Once you squash the commits, we should be able to get this merged.

zaccharieramzi:
@jakevdp all squashed

jakevdp (approved):
Thanks!

jakevdp:
It looks like there are some errors on GPU due to mismatch of the signs of eigenvectors. Seems like we may have to come up with a different testing strategy...

jakevdp:
Hi - it looks like this has become a bit out-of-date. I'm going to close for now. Feel free to re-open this if you're still interested in working on it!

PR #14531, opened 2023-02-16
[sparse] modify impl rules to always use lowering path

jakevdp:
Previously, matmuls would not use GPU lowering rules unless wrapped in JIT. This changes the setup so that GPU lowering rules are always used when available.

cc/ @tlu7

### DIFF
--- a/jax/experimental/sparse/bcoo.py
+++ b/jax/experimental/sparse/bcoo.py
@@ -41,11 +41,11 @@ import jax.numpy as jnp
 from jax.util import safe_zip, unzip2, split_list
 from jax._src import api_util
 from jax._src import core
+from jax._src import dispatch
 from jax._src.interpreters import ad
 from jax._src.interpreters import batching
 from jax._src.lax.lax import (
-  _const, ranges_like, remaining, _dot_general_batch_dim_nums,
-  DotDimensionNumbers)
+  _const, ranges_like, remaining, _dot_general_batch_dim_nums, DotDimensionNumbers)
 from jax._src.lax.slicing import GatherDimensionNumbers, GatherScatterMode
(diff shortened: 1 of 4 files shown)


jax/experimental/sparse/bcoo.py
```diff
@@ -1033,9 +1032,8 @@ def _bcoo_dot_general_batch_rule(batched_args, batch_dims, *, dimension_numbers,
 ad.defjvp(bcoo_dot_general_p, _bcoo_dot_general_jvp_lhs, None, _bcoo_dot_general_jvp_rhs)
 ad.primitive_transposes[bcoo_dot_general_p] = _bcoo_dot_general_transpose
 batching.primitive_batchers[bcoo_dot_general_p] = _bcoo_dot_general_batch_rule
-
-mlir.register_lowering(
-    bcoo_dot_general_p, _bcoo_dot_general_default_lowering)
+bcoo_dot_general_p.def_impl(partial(dispatch.apply_primitive, bcoo_dot_general_p))
```
froystig, line 1035: We have a helper function for this these days called [`dispatch.simple_impl`](https://github.com/google/jax/blob/612a94016098ae148e79f261049d1f91b24b6660/jax/_src/dispatch.py#L127), if you'd like to use it. We don't use it in too many places yet, and either way is OK.

froystig (approved):

jax/experimental/sparse/bcoo.py
jakevdp: Great idea! Done

jakevdp:
It looks like this hit some GPU test failures for `coo`/`csr` cases with padding. I'll fix those separately before landing this.

PR #14549, opened 2023-02-17
Refactor effects system to use effect types, not objects

sharadmv:
Up to this point, JAX effects were allowed/disallowed in transformations if they were members of sets that functioned as allowlists.

This meant that anytime we created a new effect object, we'd have to add it to the appropriate sets. This ended up not being a great abstraction because for things like the `State` effect, which was no longer a "singleton" effect like the debugging effects, we would be creating many effect objects, and all of them would be needed to be part of the allowlists.

This change creates the `EffectTypeSet`, which contains *types* of effects in it. It has a method that checks whether certain an effect has its type in the set.

Another part of this refactor is that it centralizes all the effect sets to a new `effects.py` file.

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 from functools import partial
-from typing import (Callable, Optional, List, Tuple, Sequence, Set, Union, Any,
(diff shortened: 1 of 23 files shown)


mattjj (approved):
Looks great!

PR #14550, opened 2023-02-17
[shard-map] add annotations and notes to shard_map_test.py

mattjj:
The goal here is to make the test file more approachable.


sharadmv (approved):
Thanks these are really helpful!

PR #14565, opened 2023-02-17
deprecation_module: Move to new internal_test_util directory.

pizzud:
Now we no longer need to mess with sys.path in deprecation_test.


pizzud:
A gentle ping? It's been nearly a week.

mattjj:
Thanks for the ping! I don't understand the motivation for the original code being deleted, but if internal tests pass with this it sounds good to me. Let's try...

mattjj:
@pizzud why move the file from tests/ to _src/?

mattjj:
Like, are you moving the file just so that we don't need to mess with `sys.path`?

If so, unless you have other information, I think it was probably intentionally not done this way to begin with: we don't want this test helper deprecation_module stuff to be included in the jax package, just in the tests.

pizzud:
There's context on the original move, including an attempt to filter _src/internal_test_util out of the distributed jax package via setup.py, in discussion on the internal cl/510260230 with @phawkins and @skye. I'm happy to restructure this if you have a different preferred structure that works internally as well, but you are correct that my goal is to clean up the code smell that is sys.path munging. I have a similar PR out for lazy_load_module that's gone unassigned so far and was planning on trying my hand at _src/test_util.py afterwards.

pizzud:
Friendly ping? It's been another week. Should I have done this internally?

mattjj (approved):

PR #14568, opened 2023-02-17
lazy_loader_module: Move to new internal_test_util directory.

pizzud:
Now we no longer need to mess with sys.path in lazy_loader_test.


hawkinsp (approved):

pizzud:
Hmm, fails internally. I'll take a look.

pizzud:
Looks like I'll have to recreate this internally in order to get the test to pass there due to the differences in environment setup.

pizzud:
Looks like I was premature in deleting this PR. The file in question could have been edited externally. I really shouldn't try to develop before I'm awake!

PR #14570, opened 2023-02-17
custom_jvp symbolic zeros support

mattjj:
follow-up on #13993

After this, we need to:
* [ ] add the custom_vjp version
* [ ] use it to adapt `closure_convert` and then fix #14561 by updating the odeint custom_vjp (and related issues)
* [ ] clarify (in code and in docs) that nondiff_argnums means non_jaxtype_argnums (or non_array_argnums / non_tracer_argnums)

### DIFF
--- a/jax/_src/ad_util.py
+++ b/jax/_src/ad_util.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from __future__ import annotations
 
+import types
 from typing import Any, Callable, Dict, TypeVar, Union, cast
 
 from jax.tree_util import register_pytree_node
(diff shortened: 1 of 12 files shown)


froystig (approved):

patrick-kidger:
Hurrah! This is great to see, thank you.

Regarding the `custom_vjp` version: I've realised that this version will actually need two things: specifying symbolic zero _tangents_, as well just symbolic zero cotangents.

The former is actually the one I find myself in more pressing need for: I have a `custom_vjp` rule accepting a callable `f` and a pytree of arrays `x`. The custom backward pass calls `vjp(f, x)`.

As `f` may not be differentiable wrt all arguments (even all floating-point arguments!), then we need to know which elements of `x` that we're actually differentiating with respect to.

With that information in hand then we could instead do `diff_x, static_x = partition(x, has_tangent); vjp(lambda _diff_x: f(combine(_diff_x, nondiff_x)), diff_x)`.

mattjj:
@patrick-kidger thanks for explaining, that makes sense. The general [Autograd](https://github.com/hips/autograd) API we had was `defvjp_argnums`, which took as one argument a tuple of integers indicating which positional arguments are being differentiated. Here's [an example](https://github.com/HIPS/autograd/blob/01eacff7a4f12e6f7aebde7c4cb4c1c2633f217d/autograd/scipy/linalg.py#L56-L66). During [the forward pass](https://github.com/HIPS/autograd/blob/01eacff7a4f12e6f7aebde7c4cb4c1c2633f217d/autograd/core.py#L36) those rules would be called like `vjpmaker(parent_argnums, value, args, kwargs)`, which would produce the pullback function to be applied to cotangents on the backward pass. I agree we need something similar here, as you say.

PR #14587, opened 2023-02-20
[jax2tf] Add support for cross-platform lowering in native serialization

gnecula:
Allow the user of native serialization to specify the platform for which the serialization to be done. This relies on newly added support for platform checking in XlaCallModule op (version 3).

The implementation here is temporary, pending having a proper cross-lowering API.

### DIFF
--- a/jax/_src/interpreters/pxla.py
+++ b/jax/_src/interpreters/pxla.py
@@ -2848,21 +2848,14 @@ def _get_and_check_device_assignment(
     arr_device_assignment = list(i._device_assignment)  # type: ignore
     if not devices:
       if first_sharding_info[0] != arr_device_assignment:
-        # Perhaps the sharding info is the same modulo using lowering_only_client
-        if not all((d1.id == d2.id and
-                    (hasattr(d1.client, "lowering_only_client") or hasattr(d2.client, "lowering_only_client")))
-                   for d1, d2 in zip(first_sharding_info[0], arr_device_assignment)):
-          raise DeviceAssignmentMismatchError([
-              DeviceAssignmentMismatch(*first_sharding_info),
-              DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
+        raise DeviceAssignmentMismatchError([
+            DeviceAssignmentMismatch(*first_sharding_info),
+            DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
     else:
       if devices != arr_device_assignment:
-        if not all((d1.id == d2.id and
-                    (hasattr(d1.client, "lowering_only_client") or hasattr(d2.client, "lowering_only_client")))
-                   for d1, d2 in zip(devices, arr_device_assignment)):
-          raise DeviceAssignmentMismatchError([
-              DeviceAssignmentMismatch(devices, MismatchType.CONTEXT_DEVICES, None),
-              DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
+        raise DeviceAssignmentMismatchError([
+            DeviceAssignmentMismatch(devices, MismatchType.CONTEXT_DEVICES, None),
+            DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
   if first_sharding_info is None and devices:
     final_device_assignment = devices
   elif first_sharding_info is None:
--- a/jax/_src/interpreters/pxla.py
+++ b/jax/_src/interpreters/pxla.py
@@ -2848,14 +2848,21 @@ def _get_and_check_device_assignment(
     arr_device_assignment = list(i._device_assignment)  # type: ignore
     if not devices:
       if first_sharding_info[0] != arr_device_assignment:
-        raise DeviceAssignmentMismatchError([
-            DeviceAssignmentMismatch(*first_sharding_info),
-            DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
+        # Perhaps the sharding info is the same modulo using lowering_only_client
+        if not all((d1.id == d2.id and
+                    (hasattr(d1.client, "lowering_only_client") or hasattr(d2.client, "lowering_only_client")))
+                   for d1, d2 in zip(first_sharding_info[0], arr_device_assignment)):
+          raise DeviceAssignmentMismatchError([
+              DeviceAssignmentMismatch(*first_sharding_info),
+              DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
     else:
       if devices != arr_device_assignment:
-        raise DeviceAssignmentMismatchError([
-            DeviceAssignmentMismatch(devices, MismatchType.CONTEXT_DEVICES, None),
-            DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
+        if not all((d1.id == d2.id and
+                    (hasattr(d1.client, "lowering_only_client") or hasattr(d2.client, "lowering_only_client")))
+                   for d1, d2 in zip(devices, arr_device_assignment)):
+          raise DeviceAssignmentMismatchError([
+              DeviceAssignmentMismatch(devices, MismatchType.CONTEXT_DEVICES, None),
+              DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
   if first_sharding_info is None and devices:
     final_device_assignment = devices
   elif first_sharding_info is None:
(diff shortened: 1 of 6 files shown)


jax/_src/pjit.py
```diff
@@ -1378,7 +1378,8 @@ def _pjit_lower(
   da = _fast_path_get_device_assignment(it.chain(in_shardings, out_shardings))
   in_shardings = SameDeviceAssignmentTuple(in_shardings, da)
   out_shardings = SameDeviceAssignmentTuple(out_shardings, da)
-  return _pjit_lower_cached(jaxpr, in_shardings, out_shardings, *args, **kwargs)
+  return _pjit_lower_cached(jaxpr, in_shardings, out_shardings, *args, **kwargs,
+                            lowering_platform_override=mlir.thread_local_state.lowering_platform_override)
```
yashk2810, line 1382: I don't think we want the compilation cache to depend on global variables like this.

In general, this is basically leaking abstraction across the pjit/jit and jax2tf boundary. pjit already has ways to tell to recompile if the platform changes via multiple routes.

Maybe we can talk about this?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -285,6 +296,26 @@ def convert(fun_jax: Callable,
   if experimental_native_lowering and not enable_xla:
     raise ValueError(
         "experimental_native_lowering is not supported with enable_xla=False")
+
+  if experimental_native_lowering and not experimental_native_lowering_platforms:
```
yashk2810, line 300: One idea: Since you already know which backend it is here since the user is telling you, why not encode that as an in_sharding (maybe a SingleDeviceSharding or any other sharding) and then pass that around?

That way this stays internal and the threading of lowering_platform_override is not needed in pjit and pxla.py?

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -213,7 +213,9 @@ def convert(fun_jax: Callable,
             polymorphic_shapes=None,
             with_gradient=True,
             enable_xla=True,
-            experimental_native_lowering="default") -> Callable:
+            experimental_native_lowering="default",
+            experimental_native_lowering_platforms=(),
+            experimental_native_lowering_strict_checks=True) -> Callable:
```
yashk2810, line 218: You mentioned this in your description: `This will be easier to do when we clean up the code deprecated by jax.Array and pjit==jit APIs.`

Why is that the case? You can just allow this new behavior only under jax.Array (via jax.config.jax_array) flag and maybe do a proper integration? You can just error out if jax.Array is not enabled for new things. 

So if this leads to a better integration, then let's do that!

jax/_src/pjit.py
gnecula: The lowering cache (not the compilation cache) must depend on the platform for which we are lowering.

Yes, we should talk.

jax/experimental/jax2tf/jax2tf.py
gnecula: I think that passing the platform in in_shardings may ultimately be the right solution. It would be quite involved, I think, because I would have to use some fake devices in the shardings (we want to be able to lower even for backends that we cannot initialize). The real reason I did not do it like that for now is that I could not follow the various places in the code where the shardings are processes. Afaik, these are in part due to supporting both jax.Array and jit==pjit and device or backend specified in the API. I would have to replicate some of these same heuristics to create shardings at the jax2tf boundary, because I have to handle the same cases that jit and pjit handle now (with device, backend, with or without shardings, ...).

jax/experimental/jax2tf/jax2tf.py
gnecula: See the comment above.

jax/experimental/jax2tf/jax2tf.py
gnecula: I should point out that this is just the first step towards multi-platform lowering. There we will lower on JAX function to a HLO module that includes the code for multiple platforms. The idea is that for the vast majority of the JAX primitives the lowering is platform-independent, but for those few cases where there is a different we will introduce a conditional in the code to check the value of a platform identifier that will be provided when the code is compiled and executed. In order to support multi-platform lowering, the code in this PR will be changed to replace `lowering_platform_override` with `lowering_platforms_override` (not the plural).

If we had a single platform to override, then we could somehow squirrel it into the `in_shardings` but I feel that this is not as clean: we would have to make fake devices when all we want is to say take the existing shardings and lowering logic and just pretend they are on a different platform. If we have multiple platforms then there is no clean way to put all the platforms into `in_shardings`.

Note that `jit` will not need to take the platform override parameters, only the lowering logic needs to.

(I should point out that I am already experimenting to verify that it is indeed possible to do multi-platform lowering in JAX without too much trouble and that the result is an HLO file that is only marginally larger than a single-platform HLO. I will be comparing with an alternative where we lower twice and we concatenate the resulting HLO.)

yashk2810: For the more general purpose thing, I would like to see a design doc before we make changes to the API.

> PR will be changed to replace lowering_platform_override with lowering_platforms_override (not the plural).

Or maybe, we can repurpose the `backend` argument that jit already takes to do multi-platform lowering? I don't think we want to introduce arguments on `lower` that only work for some primitives and not for others.

jax/_src/interpreters/pxla.py
```diff
@@ -2814,14 +2814,21 @@ def _get_and_check_device_assignment(
     arr_device_assignment = list(i._device_assignment)  # type: ignore
     if not devices:
       if first_sharding_info[0] != arr_device_assignment:
-        raise DeviceAssignmentMismatchError([
-            DeviceAssignmentMismatch(*first_sharding_info),
-            DeviceAssignmentMismatch(arr_device_assignment, s_type, source_info)])
+        # Perhaps the sharding info is the same modulo using lowering_only_client
+        if not all((d1.id == d2.id and
+                    (hasattr(d1.client, "lowering_only_client") or hasattr(d2.client, "lowering_only_client")))
```
yashk2810, line 2819: I don't really like this. Just override this check like I said in jax2tf? In the future this shouldn't be needed so I don't want to add things in the very critical piece of code that depends on something temporary.

gnecula: I did think about this. If I override this function then one would have to maintain two nearly identical copies, because we want the behavior to remain the same except for this detail. It would be hard to test that the behavior is exactly the same. I could also factor out just the check in this function into a separate function to override. But in general these kind of overrides make it very non-obvious what is going on.

I think that we will need to make changes to this function anyway to support cross-platform more cleanly. But perhaps for now this change that unblocks is acceptable?

yashk2810: > I think that we will need to make changes to this function anyway to support cross-platform more cleanly.

I don't think there will be any changes required to this function for that. But that's a long way out to think about it since the design is not even final yet.

>  If I override this function then one would have to maintain two nearly identical copies, because we want the behavior

No, my proposal was something else. I said to override this check in a very particular spot when you use LoweringOnlyDevices in jax2tf. At that point you enter into a world where this check will always fail, so it is fine to just return the backend and device_assignment you want without doing any of the other stuff. Think of it like a context manager that only affects 1 line of code. I can prototype it for you and tell you what I exactly meant. Also this function only checks the devices not the sharding info.

> It would be hard to test that the behavior is exactly the same

I don't think it would be hard if you choose to go that route. Just run a unit test that execute both the functions and add it to jax2tf and then we can never break it.

> If I override this function then one would have to maintain two nearly identical copies,

Also, if this is temporary, I would prefer maintaining 2 copies and doing the thing you want for jax2tf. We have a lot of precedent in jax for that anyways. This function hardly changes anyways. But I still think the thing that I mentioned in the chat should work out just fine.

So I don't really think adding a hack like this in this function where d1.client is lowering_only_client comes from a jax2tf internal setting is the right choice (especially since this function is executed for all ops in jax). This is basically leaking across abstraction boundary.

yashk2810: We have context managers like this: https://github.com/google/jax/blob/main/jax/_src/test_util.py#L213 that we use in tests so I am pretty confident that when you lower with fake_devices (and you only care about lowering), getting rid of all the checks and just returning a backend should work out pretty well.

Again, I can help prototype it :)

yashk2810 (approved):

PR #14594, opened 2023-02-21
[jax2tf] Enable strict platform checking for native serialized modules.

gnecula:
This takes advantage of recent changes to XlaCallModule that allow us to use a `platforms` attribute to specify what are the allowable platforms for a serialized module.

Also add a `experimental_native_lowering_strict_checks` parameter to `jax2tf.convert` to disable the check.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -35,7 +35,6 @@ from jax import tree_util
 from jax.experimental import maps
 from jax.experimental.jax2tf import shape_poly
 from jax.experimental.jax2tf import impl_no_xla
-from jax.interpreters import mlir
 from jax.interpreters import pxla
 from jax.interpreters import xla
 
@@ -55,6 +54,7 @@ from jax._src import source_info_util
 from jax._src import util
 from jax._src.global_device_array import GlobalDeviceArray
 from jax._src.interpreters import ad
+from jax._src.interpreters import mlir
 from jax._src.lax import control_flow as lax_control_flow
 from jax._src.lax import lax as lax_internal
 from jax._src.lax import linalg as lax_linalg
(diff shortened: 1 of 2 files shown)


gnecula:
@rxsang PTAL

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -729,18 +743,25 @@ def _out_type(jax_type):
...
+  call_module_attrs = dict(
       version=xla_call_module_version,
       module=mlir_serialized_module,
       Tout=out_types,
       Sout=out_shapes,
       dim_args_spec=dim_args_spec)
+  log_msg = f"version={xla_call_module_version} dim_args_spec=" + ", ".join(dim_args_spec)
+  if xla_call_module_version == 3:
+    if experimental_native_lowering_strict_checks:
+      call_module_attrs["platforms"] = (jax.default_backend().upper(),)
```
rxsang, line 755: This always set the allowed platform to current platform. Since "platforms" can be a list, should we make it configurable for users so they can add more allowed platforms? E.g. a user may want to export on CPUs and execute on TPUs. I'm wondering whether we should change the `experimental_native_lowering_strict_checks` flag to `experimental_native_lowering_allowed_platforms`.

gnecula: The cross-platform lowering is coming in a follow-up PR #14587. There is also multi-platforms support in the making PR #14595, which is why I picked a plural `platforms`. But it will take more work to land that one.

Also the generic name `strict_checks` is meant to cover a number of other checks that I will be adding soon, regarding dynamic shapes. Perhaps it may be worth adding separate parameters for enabling/disabling different checks, but I thought that this flag should be used only in an emergency and hence it is Ok to have a single flag.

rxsang (approved):

Issue #14597, opened 2023-02-21
Potential bug in Student's T random sampling


Tennessee-Wallaceh:
### Description

The current implementation of Student's-t sampling uses the same random key to generate both normal and gamma components.
See current implementation below (from jax._src.random line 1485)
```python
  df = lax.convert_element_type(df, dtype)
  key_n, key_g = _split(key)
  n = normal(key_n, shape, dtype)
  two = _lax_const(n, 2)
  half_df = lax.div(df, two)
  g = gamma(key_n, half_df, shape, dtype)
  return n * jnp.sqrt(half_df / g)
```

To me, it appears that line 1490 should use `key_g`, that is `g = gamma(key_g, half_df, shape, dtype)`.

The impact of this is not immediately clear to me, but this could be subtly  biasing any code currently sampling Student's-t.

### What jax/jaxlib version are you using?

jax v0.4.1

### Which accelerator(s) are you using?

n/a

### Additional system info

n/a

jakevdp:
That looks like a bug – are you interested in opening a PR with the fix?

Tennessee-Wallaceh:
Hi, yes I can do that now.

Tennessee-Wallaceh:
PR ready for review here - https://github.com/google/jax/pull/14598

PR #14598, opened 2023-02-21
Update student-t sampling to use correct key for gamma

Tennessee-Wallaceh:
Fix for issue #14597. 
Simple change to use correct random key in sampling procedure.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1487,7 +1487,7 @@ def _t(key, df, shape, dtype) -> Array:
   n = normal(key_n, shape, dtype)
   two = _lax_const(n, 2)
   half_df = lax.div(df, two)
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks good! Let's run the tests to see if anything needs fixing.

jakevdp:
It looks like the test of the statistical properties of the distribution is failing... that test is notoriously finnicky, because it is only probabilistically correct. You may be able to fix it by adjusting the test's random seed or number of generated elements

jakevdp (commented):
Looks good, thanks! As a last step, could you please squash your changes into a single commit? (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) thanks!

PR #14605, opened 2023-02-21
[sparse] coo - assume padded unless otherwise specified

jakevdp:
This fixes a GPU lowering bug for csr matrices: fixes half the failures from #14531


jakevdp:
Replaced by #14631

PR #14610, opened 2023-02-21
Add `JaxprInputEffect` and refactor `StateEffect`s to use it

sharadmv:
This PR adds a new type of effect to the JAX effects system: `JaxprInputEffect` (I'm open to a better name).

A `JaxprInputEffect` is an effect that's tied to the input of a jaxpr (either a constvar or invar). Specifically, it is an effect tied to the *index* of the input in jaxpr's signature. JAX's internals are therefore responsible for keeping track of how these indices are changed via transformations. 

Transformations that operate by reinterpreting a jaxpr should be fine, since the jaxpr builder correctly constructs JaxprInputEffects with the right indices. This works so long as our `abstract_eval` rules are correct.

Transformations that do jaxpr->jaxpr data structure manipulation also need to be updated to do the bookkeeping correctly because deleting an invar or constvar from a jaxpr could result in the `JaxprInputEffect` now pointing to the wrong invar.

The best example of such an effect is the state effect(s), which tie a `Read/Write/Accum` to a specific `Ref`. This PR thus makes bookkeeping the state effect easier and it no longer needs to be handled specially.

Note that the alternative to this approach would be to use heap-parameters/names. The approaches are isomorphic to each other but in some simple HOPs (like jit or closed_call), we don't need to do name/heap-parameter substitution to update the effects, so it ends up being simpler.

### DIFF
--- a/jax/interpreters/partial_eval.py
+++ b/jax/interpreters/partial_eval.py
@@ -946,10 +946,12 @@ def tracers_to_jaxpr(
       raise TypeError(r)
 
   env_vars, env_vals = unzip2(env.items())
+  invars = [*env_vars, *map(get_atom, in_tracers)]
   const_vars, const_vals = unzip2(consts.items())
-  effects = core.join_effects(*(eqn.effects for eqn in eqns))
-  jaxpr = Jaxpr(const_vars, [*env_vars, *map(get_atom, in_tracers)],  # type: ignore[list-item]
-                map(get_atom, out_tracers), eqns, effects)
+  outvars = map(get_atom, out_tracers)  # type: ignore[arg-type]
+  jaxpr_effects = make_jaxpr_effects(const_vars, invars, outvars, eqns)
+  jaxpr = Jaxpr(const_vars, invars,  # type: ignore[list-item,arg-type]
+                outvars, eqns, jaxpr_effects)
   config.jax_enable_checks and core.check_jaxpr(jaxpr)
   # del getvar  # needed to avoid cyclic-reference closure, apparently!
   return jaxpr, const_vals, env_vals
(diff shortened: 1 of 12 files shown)


mattjj (approved):
LGTM, from our discussion: we wonder if we can implement some jaxpr-signature-updating logic as a (1) translate from dbidx to heap/named form, (2) trivially update signature, (3) translate back from heap/named form to dbidx form. _If_ so that might make it easier to write such things.

jax/interpreters/partial_eval.py
```diff
@@ -1581,7 +1613,10 @@ def _const_folding_and_forwarding(jaxpr, constvals):
     # always apply invar substitutions
     eqn = eqn.replace(invars=[var_subs.get(v, v) for v in eqn.invars])
     # if any inputs are constants and we have a constant-folding rule, apply it
-    if eqn.primitive in const_fold_rules and any(v in consts for v in eqn.invars):
+    has_input_effect = any(isinstance(eff, effects.JaxprInputEffect)
+                           for eff in eqn.effects)
+    if (eqn.primitive in const_fold_rules and any(v in consts for v in eqn.invars)
+        and not has_input_effect):
```
mattjj, line 1619: This could be a little conservative, hypothetically, e.g. what if we had a constant folding rule for jit (which just constant folded inside it) and we had an eqn corresponding to an application of `f` to a constant where

```python
@jit
def f(c1):
  c2 = lax.convert_element_type(c1, jnp.dtype('float32'))
  x_ref[()] = 5.
  return c2
```

This logic would block constant folding of it.

(Sharad observed: we could rule out issues here if we implemented such transformations as make-jaxpr-of-jaxpr-interpreter.)

PR #14613, opened 2023-02-21
Add `--pre` to nightly libtpu pip install command.

skye:
This is necessary to make sure we pick up the nightly "dev" versions hosted on GCP and not the fake package at
https://pypi.org/project/libtpu-nightly/.

### DIFF
--- a/.github/workflows/cloud-tpu-ci-nightly.yml
+++ b/.github/workflows/cloud-tpu-ci-nightly.yml
@@ -41,7 +41,7 @@ jobs:
             pip install .
             pip install --pre jaxlib \
               -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-            pip install libtpu-nightly \
+            pip install --pre libtpu-nightly \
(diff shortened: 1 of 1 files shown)


.github/workflows/cloud-tpu-ci-nightly.yml
```diff
@@ -41,7 +41,7 @@ jobs:
             pip install .
             pip install --pre jaxlib \
               -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-            pip install libtpu-nightly \
+            pip install --pre libtpu-nightly \
```
yashk2810, line 44: optional: Maybe also add `-U`?

yashk2810 (approved):

.github/workflows/cloud-tpu-ci-nightly.yml
skye: Done. It's not necessary since we uninstall at the beginning, but shouldn't hurt. Also added to jaxlib for consistency.

PR #14633, opened 2023-02-23
[shard-map] add systematic vmap-of-shmap tests

mattjj:
Follow-up on #14513 (and #14550)

### DIFF
--- a/tests/shard_map_test.py
+++ b/tests/shard_map_test.py
@@ -15,6 +15,7 @@
 from functools import partial
 import itertools as it
 import math
+import operator as op
 import os
 from types import SimpleNamespace
 from typing import (Any, Sequence, Set, Iterable, Iterator, NamedTuple,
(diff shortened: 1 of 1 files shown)


yashk2810 (approved):
Should we wait for Marcello's reply on the pxla.py change?

mattjj:
> Should we wait for Marcello's reply on the pxla.py change?

No I'm already blocked on one XLA bug, don't want to block on another :P I'll leave a TODO(mattjj)

yashk2810:
> No I'm already blocked on one XLA bug, don't want to block on another

Ohh.. I was saying maybe we wait for his reply (lol) to see if this is actually correct.

We should fix this in python until HloSharding gets fixed. That's fine by me.

mattjj:
Well given the test case here I know the OpShardings in question represent replicated shardings.

mattjj:
Following up on the comment thread: it was an XLA bugs that these MANUAL shardings propagated to outputs!

PR #14636, opened 2023-02-23
[jax2tf] Preserve the the inner sharding annotations even if they are replicated

gnecula:
Previously we were dropping the replicated sharding annotations, in order to avoid having extraneous sharding annotations on inputs and output on all functions. However, we should not drop sharding annotation inside a function.


gnecula:
This seems to have been merged automatically once I removed the test. Will follow up separately with an updated test.

PR #14651, opened 2023-02-23
Move _multi_slice into pxla

jakevdp:
This avoids polluting the `jax.Array` namespace with this private method.

### DIFF
--- a/jax/_src/interpreters/pxla.py
+++ b/jax/_src/interpreters/pxla.py
@@ -49,6 +49,7 @@ import jax
 from jax.errors import JAXTypeError
 from jax.interpreters import partial_eval as pe
 from jax.tree_util import tree_flatten, tree_map
+from jax.typing import ArrayLike
 
 from jax._src import abstract_arrays
 from jax._src import api_util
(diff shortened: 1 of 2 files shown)


jax/_src/interpreters/pxla.py
```diff
@@ -436,12 +437,28 @@ def _shard_array(x, devices, indices):
 def shard_device_array(x, devices, indices):
   start_indices, limit_indices, removed_dims = unzip3(
       as_slice_indices(x, idx) for idx in indices)
-  shards = x._multi_slice(start_indices, limit_indices, removed_dims)
+  shards = jax.jit(_multi_slice, static_argnums=(1, 2, 3))(
```
yashk2810, line 440: Can we run the benchmarks and see if this does not regress it?

jax/_src/interpreters/pxla.py
```diff
@@ -436,12 +437,28 @@ def _shard_array(x, devices, indices):
...
+    x, start_indices, limit_indices, removed_dims)
   return device_put(shards, devices)
 for t in device_array.device_array_types:
   shard_arg_handlers[t] = shard_device_array
 
 
+@core.stash_axis_env()
```
yashk2810, line 447: Just jit this instead of jitting inside shard_device_array?

jakevdp: JIT is not available at the module level scope at import time, unfortunately

yashk2810: Ok, then let's run benchmarks or put it in a place where we can jit it at the top level. I am just afraid that this will cause regressions.

jakevdp: I think "put it in a place where we can jit it at the top level" was probably why this was stuck onto `jax.Array` in the first place... oh well. The other option would be to use a local import at the call-site, which is a strategy used elsewhere in this file

yashk2810: I am fine with anything you think is correct here :)

I just want to make sure we are not regressing anything because this function runs a lot of times in JAX. It is not that bad to put it on Array too but I understand that it is not ideal.

jakevdp:
Well, this is ending up being less clean than I thought. I'm going to keep things as-is.

PR #14652, opened 2023-02-24
Fix nondeterminism issue with ordered effects

sharadmv:


jakevdp (approved):
Thanks for looking into this!

Issue #14640, opened 2023-02-23
Add `random.chisquare` and `random.f` 


JiaYaobo:
For my works related to statistics part, chisquare or F distribution are often used, and there're `random.chisquare` and `random.f` in `numpy`, I'd implement these for my convenience in `JAX`, I'd like to know, would you like to add `chisquare` or `f` in future? Actually I'veimplement these for my convenience in `JAX` and pass the tests
Here's my code

```python
def chisquare(key: KeyArray,
              df: RealArray,
              shape: Shape = (),
              dtype: DTypeLikeFloat = dtypes.float_) -> Array:
  """Sample Chisquare random values with given shape and float dtype.

  Args:
    key: a PRNG key used as the random key.
    df: a float or array of floats broadcast-compatible with ``shape``
      representing the parameter of the distribution.
    shape: optional, a tuple of nonnegative integers specifying the result
      shape. Must be broadcast-compatible with ``df``. The default (None)
      produces a result shape equal to ``df.shape``.
    dtype: optional, a float dtype for the returned values (default float64 if
      jax_enable_x64 is true, otherwise float32).

  Returns:
    A random array with the specified dtype and with shape given by ``shape`` if
    ``shape`` is not None, or else by ``df.shape``.
  """
  key, _ = _check_prng_key(key)
  if not dtypes.issubdtype(dtype, np.floating):
    raise ValueError(f"dtype argument to `chisquare` must be a float "
                    f"dtype, got {dtype}")
  dtype = dtypes.canonicalize_dtype(dtype)
  shape = core.canonicalize_shape(shape)
  return _chisquare(key, df, shape, dtype)

@partial(jit, static_argnums=(2, 3))
def _chisquare(key, df, shape, dtype) -> Array:
  if shape is None:
    shape = np.shape(df)
  else:
    _check_shape("chisquare", shape, np.shape(df))
  df = lax.convert_element_type(df, dtype)
  two = _lax_const(df, 2)
  half_df = lax.div(df, two)
  log_g = loggamma(key, a=half_df, shape=shape, dtype=dtype)
  chi2 = lax.mul(jnp.exp(log_g), two)
  return chi2
  

def f(key: KeyArray,
      dfnum: RealArray,
      dfden: RealArray,
      shape: Shape = (),
      dtype: DTypeLikeFloat = dtypes.float_) -> Array:
  """Sample F-distribution random values with given shape and float dtype.

  Args:
    key: a PRNG key used as the random key.
[...]

PR #14658, opened 2023-02-24
Add `random.chisquare` and `random.f`

JiaYaobo:
As mentioned in https://github.com/google/jax/issues/14640, add `random.chisquare` and `random.f`


jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified dtype and with shape given by ``shape`` if
+    ``shape`` is not None, or else by ``df.shape``.
+  """
+  key, _ = _check_prng_key(key)
+  if not dtypes.issubdtype(dtype, np.floating):
+    raise ValueError(f"dtype argument to `chisquare` must be a float "
+                    f"dtype, got {dtype}")
```
jakevdp, line 1517: Nit: align indentation

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+  dtype = dtypes.canonicalize_dtype(dtype)
+  shape = core.canonicalize_shape(shape)
```
jakevdp, line 1519: This will raise an error if `shape = None`, which does not align with the description in the docstring

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+    dtype: optional, a float dtype for the returned values (default float64 if
+      jax_enable_x64 is true, otherwise float32).
+
+  Returns:
+    A random array with the specified dtype and with shape given by ``shape`` if
+    ``shape`` is not None, or else by ``df.shape``.
+  """
+  key, _ = _check_prng_key(key)
+  if not dtypes.issubdtype(dtype, np.floating):
+    raise ValueError(f"dtype argument to `chisquare` must be a float "
```
jakevdp, line 1516: nit: unnecessary `f`

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+
+  Returns:
+    A random array with the specified dtype and with shape given by ``shape`` if
+    ``shape`` is not None, or else by ``df.shape``.
+  """
+  key, _ = _check_prng_key(key)
+  if not dtypes.issubdtype(dtype, np.floating):
+    raise ValueError(f"dtype argument to `f` must be a float "
+                    f"dtype, got {dtype}")
+  if not np.shape(dfden) == np.shape(dfnum):
```
jakevdp, line 1564: Should we allow broadcast-compatible shapes? I believe `np.random.f` does this.

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+    raise ValueError(f"shape of `dfnum` and `dfden` must be equal "
+                    f"got {np.shape(dfnum)} != {np.shape(dfden)}")
```
jakevdp, line 1566: Nit: alignment & unused `f`

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+  dtype = dtypes.canonicalize_dtype(dtype)
+  shape = core.canonicalize_shape(shape)
```
jakevdp, line 1568: As above, this raises if `shape is None`

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
...
+  half_df = lax.div(df, two)
+  log_g = loggamma(key, a=half_df, shape=shape, dtype=dtype)
+  chi2 = lax.mul(jnp.exp(log_g), two)
+  return chi2
+
+
+def f(key: KeyArray,
+      dfnum: RealArray,
+      dfden: RealArray,
+      shape: Shape = (),
```
jakevdp, line 1539: Perhaps `shape: Optional[Shape] = None`?

jax/_src/random.py
```diff
@@ -1491,6 +1491,98 @@ def _t(key, df, shape, dtype) -> Array:
   return n * jnp.sqrt(half_df / g)
 
 
+def chisquare(key: KeyArray,
+              df: RealArray,
+              shape: Shape = (),
```
jakevdp, line 1496: Perhaps default to `shape = None`?

jakevdp (changes requested):
Looks good overall, thanks for the contribution!

A few comments below, particularly around validation of `shape` and other inputs.

Please also:
- add the new functions to the doc listing: https://github.com/google/jax/blob/main/docs/jax.random.rst#list-of-available-functions
- once everything is addressed, squash all changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests)

Thanks!

jax/random.py
```diff
@@ -187,4 +187,6 @@
   uniform as uniform,
   unsafe_rbg_key as unsafe_rbg_key,
   weibull_min as weibull_min,
+  chisquare as chisquare,
+  f as f
```
jakevdp, line 191: Also please alphabetize imports

JiaYaobo:
@jakevdp Thanks for your review, I'll meet these requests soon!

jax/_src/random.py
JiaYaobo: Hi @jakevdp, thanks for you review! I find the shape parameter for functions in `random.py` are inconsistent? For example, `multivariate_normal` requests `Optional[Shape] = None` as default, So `multivariate_normal(key, mean, cov, shape=None)` is Ok,but `cauchy`  and `double_sided_maxwell`request `shape = ()` as default, so `cauchy(..., shape=None)` or `double_sided_maxwell(..., shape=None)` raises errors, though the docstring indicates the default value,  sometimes different behavior for a sequence of these similar functions still confuse us,  shall we make a standard for this kind of parameters? For myself, I'd like to  choose `shape: Optional[Shape] = None`

jakevdp: Agreed it's inconsistent, but I lean toward defaulting to `None` because I think it makes sense that if the shape is unspecified it should respect the shape of the input parameters rather than raise an error.

jax/_src/random.py
```diff
@@ -1491,6 +1491,108 @@ def _t(key, df, shape, dtype) -> Array:
...
+    _check_shape("f", shape, params_shapes)
+  dfden = lax.convert_element_type(dfden, dtype)
+  dfnum = lax.convert_element_type(dfnum, dtype)
+  key_dfd, key_dfn = _split(key)
+  chi2_dfn = chisquare(key_dfn, dfnum, shape, dtype)
+  chi2_dfd = chisquare(key_dfd, dfden, shape, dtype)
+  # broadcast dfden and dfnum again to do div operation
+  dfden = lax.broadcast(dfden, shape[:-1])
+  dfnum = lax.broadcast(dfnum, shape[:-1])
+  num = lax.div(chi2_dfn, dfnum)
```
jakevdp, line 1590: Have you tested this with non-1D shapes? It looks like if shape is something like `(5, 6)`, this will lead to a broadcasting error, because `chi2dfn` and `dfnum` are different shapes

JiaYaobo: Hi @jakevdp I think I should misunderstand `lax.broadcast` and `jnp.broadcast_to` , thanks for your careful review, so now I would use `jnp.broadcast_arrays` and `jnp.broadcast_to` for this shape broadcasting

jax/_src/random.py
JiaYaobo: Can't agree more

jax/_src/random.py
```diff
@@ -1491,6 +1491,105 @@ def _t(key, df, shape, dtype) -> Array:
...
+  return _f(key, dfnum, dfden, shape, dtype)
+
+@partial(jit, static_argnums=(3, 4), inline=True)
+def _f(key, dfnum, dfden, shape, dtype) -> Array:
+  # broadcast dfnum and dfden to generate chi2 rvs
+  dfden, dfnum = jnp.broadcast_arrays(dfden, dfnum)
+  if shape is None:
+    shape = dfden.shape
+  else:
+    shape = dfden.shape + shape
```
jakevdp, line 1577: This seems like surprising shape semantics. Consider the analogous numpy API:
```python
import numpy as np
result = np.random.f(np.ones(5), 2, size=(5,))
print(result.shape)
# (5,)
```
Here, doing the equivalent would lead to the output having shape `(5, 5)`.

We should probably handle the `shape` parameter similarly to other functions in this file, e.g. https://github.com/google/jax/blob/f66f6ec98a2bceeb7f609cbfdbeee81dc74ad32f/jax/_src/random.py#L644-L647

jakevdp: Yeah, that API seems odd to me as well.

jakevdp: In general, it looks like the random submodule is very inconsistent in its handling of shapes and dtypes, and for the most parts the tests don't cover this. We should probably just overhaul it all in a single pass.

JiaYaobo: Ok i think i'll raise an issue and open another pr for double_side_maxwellafter modifying this pr. ：）

jax/_src/random.py
JiaYaobo: I think this time shape results are compatible with numpy

jax/_src/random.py
jakevdp: I think rather than a PR just for `double_side_maxwell`, we need a revamp of most of the `jax.random` module to make the API consistent and robust. That gets tricky because of backward compatibility and API deprecation issues; I'll plan to look into that soon.

jakevdp (approved):
Looks good, thanks!

PR #14682, opened 2023-02-25
add custom gradient for relu6 at boundary positions

MahmoudAshraf97:
Hi,
the gradient for jax.nn.relu6 at 0 and 6 is equal to 0.5, where it should be equal to 0 in order to be consistent with ReLU and ReLU6 in other frameworks such as pytorch and tensorflow


hawkinsp:
Looks like @mattjj made the same change (https://github.com/google/jax/commit/9c39b6f70c601d4ca29ccbcd28615c7179fa5016)

PR #14688, opened 2023-02-26
[typing] Use ParamSpec in JIT annotation

NeilGirdhar:
This pull request would be a huge improvement for Jax users who use type checkers like MyPy or Pyright, which now support `ParamSpec`. 

Consider:
```python
from jax import Array, jit

@jit
def f(x: Array, y: Array) -> Array:
    pass

reveal_type(f.__call__)
```
Previous to this PR, users who decorate a function with `jit` lose all annotations of the method:
```
Pyright: Type of "f.__call__" is "(*args: Unknown, **kwargs: Unknown) -> Unknown"
MyPy: Revealed type is "def (*args: Any, **kwargs: Any) -> Any"
```
After this PR, we get:
```
Pyright: Type of "f.__call__" is "(x: Array, y: Array) -> Array"
MyPy: Revealed type is "def (x: Array, y: Array) -> Array"
```
This unblocks a lot of type checking.

Part of https://github.com/google/jax/issues/12049 cc: @jakevdp

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -32,8 +32,8 @@ import dataclasses
 import enum
 from functools import partial
 import inspect
-from typing import (Any, Literal, Optional, TypeVar, overload,
-                    cast, TYPE_CHECKING)
+from typing import (Any, Literal, Optional, TypeVar, overload, cast,
+                    TYPE_CHECKING)
 import weakref
 
 import numpy as np
@@ -175,8 +175,8 @@ class NotSpecified:
     return "<not-specified>"
 
 @overload
-def jit(
-  fun: Callable, /, *,
+def jit[**Parameters, ReturnType](
+  fun: Callable[Parameters, ReturnType], /, *,
   in_shardings: Any = ...,
   out_shardings: Any = ...,
   static_argnums: int | Sequence[int] | None = ...,
@@ -188,11 +188,11 @@ def jit(
   backend: str | None = ...,
   inline: bool | Inline = ...,
   compiler_options: dict[str, Any] | None = ...,
-) -> pjit.JitWrapped:
+) -> pjit.JitWrapped[Parameters, ReturnType]:
   ...
 
 @overload
-def jit(
+def jit[**Parameters, ReturnType](
   *,
   in_shardings: Any = ...,
   out_shardings: Any = ...,
@@ -205,11 +205,12 @@ def jit(
   backend: str | None = ...,
   inline: bool | Inline = ...,
   compiler_options: dict[str, Any] | None = ...,
-) -> Callable[[Callable], pjit.JitWrapped]:
+) -> Callable[[Callable[Parameters, ReturnType]],
+              pjit.JitWrapped[Parameters, ReturnType]]:
   ...
 
-def jit(
-  fun: Callable | NotSpecified = NotSpecified(), /, *,
+def jit[**Parameters, ReturnType](
+  fun: Callable[Parameters, ReturnType] | NotSpecified = NotSpecified(), /, *,
   in_shardings: Any = sharding_impls.UNSPECIFIED,
   out_shardings: Any = sharding_impls.UNSPECIFIED,
   static_argnums: int | Sequence[int] | None = None,
@@ -221,7 +222,9 @@ def jit(
   backend: str | None = None,
   inline: bool | Inline = False,
   compiler_options: dict[str, Any] | None = None,
-) -> pjit.JitWrapped | Callable[[Callable], pjit.JitWrapped]:
+) -> (pjit.JitWrapped[Parameters, ReturnType]
+      | Callable[[Callable[Parameters, ReturnType]],
+                 pjit.JitWrapped[Parameters, ReturnType]]):
   """Sets up ``fun`` for just-in-time compilation with XLA.
 
   Args:
--- a/jax/_src/pjit.py
+++ b/jax/_src/pjit.py
@@ -20,7 +20,7 @@ from dataclasses import dataclass, replace
 from functools import partial
 import inspect
 import weakref
-from typing import NamedTuple, Any, Union
+from typing import NamedTuple, Any, Self, Union, overload
 import warnings
 
 import numpy as np
@@ -676,7 +676,8 @@ def _infer_input_type(fun: Callable, dbg_fn: Callable[[], core.DebugInfo],
   return tuple(avals)
 
 
-class JitWrapped(stages.Wrapped):
+class JitWrapped[**Parameters, ReturnType](
+    stages.Wrapped[Parameters, ReturnType]):
 
   def eval_shape(self, *args, **kwargs):
     """See ``jax.eval_shape``."""
@@ -685,12 +686,27 @@ class JitWrapped(stages.Wrapped):
   def trace(self, *args, **kwargs) -> stages.Traced:
     raise NotImplementedError
 
+  # The underlying C++ implementation is a descriptor (like a plain
+  # function), so accessing a jitted method through an instance binds it,
+  # dropping the leading `self` parameter. There's no way to express that
+  # precisely for an arbitrary ParamSpec, so the bound case falls back to
+  # `Callable[..., ReturnType]`, matching how `types.FunctionType.__get__`
+  # is typed in typeshed.
+  @overload
+  def __get__(self, instance: None, owner: type | None = None, /) -> Self: ...
+  @overload
+  def __get__(
+      self, instance: object, owner: type | None = None, /
+  ) -> Callable[..., ReturnType]: ...
+  def __get__(self, instance, owner=None):
+    raise NotImplementedError
+
 
 # in_shardings and out_shardings can't be None as the default value
 # because `None` means that the input is fully replicated.
 @partial(api_boundary, repro_api_name="pjit.pjit")
-def pjit(
-    fun: Callable,
+def pjit[**Parameters, ReturnType](
+    fun: Callable[Parameters, ReturnType],
     in_shardings: Any = UNSPECIFIED,
     out_shardings: Any = UNSPECIFIED,
     static_argnums: int | Sequence[int] | None = None,
@@ -702,7 +718,7 @@ def pjit(
     backend: str | None = None,
     inline: bool = False,
     compiler_options: dict[str, Any] | None = None,
-) -> JitWrapped:
+) -> JitWrapped[Parameters, ReturnType]:
   """`jax.experimental.pjit.pjit` has been deprecated. Please use `jax.jit`."""
   return make_jit(
       fun, in_shardings=in_shardings, out_shardings=out_shardings,
--- a/jax/_src/stages.py
+++ b/jax/_src/stages.py
@@ -889,7 +889,7 @@ def raise_lo_outs(hi_avals, lo_outs):
   return hi_outs
 
 @runtime_checkable
-class Wrapped(Protocol):
+class Wrapped[**Parameters, ReturnType](Protocol):
   """A function ready to be traced, lowered, and compiled.
 
   This protocol reflects the output of functions such as
@@ -898,7 +898,8 @@ class Wrapped(Protocol):
   to compilation, and the result compiled prior to execution.
   """
 
-  def __call__(self, *args, **kwargs):
+  def __call__(self, *args: Parameters.args, **kwargs: Parameters.kwargs
+               ) -> ReturnType:
     """Executes the wrapped function, lowering and compiling as needed."""
     raise NotImplementedError
 


NeilGirdhar:
Not sure what's going on with the type checking, but perhaps it's not running MyPy 1.0.1?

Also, I'm still working on fixing this for applying `jit` to methods, which currently works, but does not type check correctly.

setup.py
```diff
@@ -67,6 +67,7 @@ def generate_proto(source):
         'numpy>=1.20',
         'opt_einsum',
         'scipy>=1.5',
+        'typing_extensions>=4.5.0',
```
yashk2810, line 70: What's the reason for adding this as a compulsory dependency?

NeilGirdhar: I could make it a dev-dependency.  Would that be better?

NeilGirdhar: I made it a dev-dependency in a subsequence change (within this PR) so that you can assess whether you like that better.

NeilGirdhar:
After testing this a lot with my code, I think MyPy is [still not ready ](https://github.com/python/mypy/issues?q=is%3Aissue+is%3Aopen+paramspec+label%3Atopic-paramspec)to check code with complex `ParamSpec` usage (although Pyright is).

If we check something like this in, it may trip up MyPy quite a bit.  On the other hand, it has a the amazing effect of exposing plenty Jax type annotations that were previously hidden.  Plenty of functions in `jax.numpy` are stripped of perfectly good annotations by `@jit`.

Also, there are some limitations in Python typing wrt method decorators.  I've [posted about this](https://mail.python.org/archives/list/typing-sig@python.org/thread/QZX5WN4NL67L6HEPWDEJQP6L73TRYYHN/) on typing-sig.  If people are using `jit` on methods, then they will get errors, and there doesn't seem to be a universal fix.

What I will do is add typed decorators to my tjax library.  Please let me know if you'd like to get this in to Jax sooner or if we should wait on it.

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
   - id: flake8
 
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: 'v0.982'
+  rev: 'v1.0.1'
```
jakevdp, line 17: Whenever we bump the mypy version, it tends to require a ton of small fixes across the package. Because of that, it would be better to do this separately from the ParamSpec thing I think.

setup.py
jakevdp: I think it's fine to add as a dependency, but I'm also curious why it's necessary

jakevdp:
Thanks for looking at this! ParamSpec seems like it could be interesting; that said it still looks to be relatively unstable and probably the remarks [here](https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#avoid-unstable-typing-mechanisms) still apply.

JesseFarebro:
Hasn't this already been discussed here: https://github.com/google/jax/issues/10311?

NeilGirdhar:
@JesseFarebro My mistake, I didn't find that when I searched!  MyPy has progressed since then, but maybe still not enough.

.pre-commit-config.yaml
NeilGirdhar: Got it, makes sense.

setup.py
NeilGirdhar: @jakevdp We need `ParamSpec`, which was [added](https://peps.python.org/pep-0612/) in Python 3.10 (Jax's minimum version per NEP 29 is 3.8.  We could make it a dev-dependency (as I've now done) although I think it complicates the code.

NeilGirdhar:
Great news: Thanks to the recently merged https://github.com/python/mypy/pull/15837, it appears that the main issue with this pull request may have been solved: https://github.com/python/mypy/issues/12169.  Also, many of the MyPy errors that may have affected its usage may have been solved: https://github.com/python/mypy/issues/11846, https://github.com/python/mypy/issues/12986, https://github.com/python/mypy/issues/14802.

jakevdp:
We've updated mypy to 1.4.1 in the meantime, and #17147 bumps it to 1.5.0 – can you sync your PR to the current main branch?

NeilGirdhar:
@jakevdp Done.  FYI the MyPy pull I linked is merged, but it's not in any released MyPy yet.

NeilGirdhar:
@jakevdp MyPy 1.6 is out today, and it may now [support](https://mypy-lang.blogspot.com/2023/10/mypy-16-released.html) using `ParamSpec` as in this pull request.  (See the section on ParamSpec improvements.)

Is there a process or plan for upgrading Jax to use MyPy 1.6?

jakevdp:
As soon as the new mypy version is mirrored at https://github.com/pre-commit/mirrors-mypy, we can bump the version in the pre-commit configuration.

jakevdp:
It looks like the update will happen automatically in about 13 hours: https://github.com/pre-commit/mirrors-mypy/blob/08cbc46b6e135adec84911b20e98e5bc52032152/.github/workflows/main.yml#L6

jakevdp:
#18066 updates mypy to v1.6.0

NeilGirdhar:
Okay, I've rebased this to take advantage of the new MyPy version

NeilGirdhar:
Just FYI if you want to get rid of the `typing_extensions` dependency, one option would be to follow [SPEC 0](https://scientific-python.org/specs/spec-0000/), and drop Python 3.9 (since you already support Python 3.12).  Python 3.10 has `ParamSpec`.

jakevdp:
We'll drop Python 3.9 sometime after April 2024, following NEP 29.

NeilGirdhar:
> sometime after April 2024, following NEP 29 (see

Fair enough, but isn't NEP 29 now superseded by SPEC 0?

jakevdp:
> Fair enough, but isn't NEP 29 now superseded by SPEC 0?

Not in JAX... our policy still explicitly points to NEP 29. We could have a discussion about changing that, but the discussion would have to happen.

jakevdp:
I'm proposing switching to SPEC 0 in #18072

jakevdp:
Well, it looks like the result of the SPEC-0 discussion is that we're going to support old Python releases for even longer than we did previously. If the current proposal goes in, we won't drop Python 3.9 until after July 2024. But for the purpose of this work, I believe we can use `typing_extensions` until then.

jakevdp:
Hi @NeilGirdhar – what's the status of this? Do you want to keep pushing on this approach?

NeilGirdhar:
@jakevdp Yes, I'm really looking forward to something like this making it in.  I'll need to look at it tomorrow since I have plans tonight.  Were you able to update to the new MyPy?

jakevdp:
Yes, we currently run the CI on mypy 1.6.1: https://github.com/google/jax/blob/953f4670d88d2a1c168a4ad0b44ed940f6c58829/.pre-commit-config.yaml#L29-L30

NeilGirdhar:
Running MyPy 1.6.1, I'm still getting a lot of bad errors:
```

jax/experimental/sparse/linalg.py:102: error: Argument 2 to "__call__" of "Wrapped" has incompatible type "Array"; expected "P.args"  [arg-type]
jax/experimental/sparse/linalg.py:102: error: Argument 3 to "__call__" of "Wrapped" has incompatible type "int"; expected "P.args"  [arg-type]
jax/experimental/sparse/linalg.py:102: error: Argument 4 to "__call__" of "Wrapped" has incompatible type "Array | float | None"; expected "P.args"  [arg-type]
```
It seems like more MyPy bugs.   PyRight seems okay with it.  I guess we'll have to wait longer?  And someone should consider submitting this bug to MyPy?

jakevdp:
Seems like maybe a bad interaction with `functools.partial`?

jax/_src/numpy/lax_numpy.py
```diff
@@ -1958,6 +1958,14 @@ def block(arrays: ArrayLike | list[ArrayLike]) -> Array:
...
   return out
 
+@overload
+def atleast_1d(x: ArrayLike, /) -> Array:
+  ...
+
+@overload
+def atleast_1d(x: ArrayLike, y: ArrayLike, /, *arys: ArrayLike) -> list[Array]:
+  ...
+
```
jakevdp, line 1968: Split out these changes in a separate PR?

NeilGirdhar: Yes, okay, I will do that when we get close to having this checked in.   This PR will expose a lot of typing errors as code that's jitted becomes typed.

NeilGirdhar:
> Seems like maybe a bad interaction with `functools.partial`?

That makes sense.  I'll look at it as soon as I have more time.

NeilGirdhar:
Filed a MyPy bug https://github.com/python/mypy/issues/16404

NeilGirdhar:
@jakevdp I think we should wait one more release of MyPy, which should fix the above errors.  After that, although this change will not help MyPy users (the inferred type of a jit-decorated function will just be `Callable`), this change won't hurt, and it will help Pyright users.

jax/_src/numpy/lax_numpy.py
NeilGirdhar: Factored out as requested: https://github.com/google/jax/pull/18395

jakevdp:
> @jakevdp I think we should wait one more release of MyPy

Sounds good

NeilGirdhar:
@jakevdp I've rebased this now.  Is there a way to get the tests to run?  https://github.com/python/mypy/issues/1484 is still unsolved, so there may only be modest gains in MyPy.

jakevdp (approved):
Looks good! I'm going to approve and then pull it in to run internal pytype tests

jakevdp:
Pytype is failing with this message:
```
File ".../jax/_src/api.py", line 100, in <module>: argument "covariant" to TypeVar not supported yet [not-supported-yet]
```
Seems this is a known issue (https://github.com/google/pytype/issues/1471). Is there any way to do this kind of improvement without using covariant TypeVars?

NeilGirdhar:
@jakevdp 
> I'm going to approve and then pull it in to run internal pytype tests'

Thanks for taking the time!

> Is there any way to do this kind of improvement without using covariant TypeVars?

The return type does need to be covariant.  One option is to do this in two steps.  Just keep the paramspec stuff working and not type the return type.

Another option is to see if Pytype needs the covariant marker.  I have a feeling that MyPy and Pyright will fail wihtout it.  It's funny because in Python 3.12, we won't need these markers at all as the type checker can infer them, although I think we'd have to use the new PEP 695 syntax.

Should I remove the annotations on the  return type?

jakevdp:
I'm honestly not sure what the best fix is here.

jakevdp:
I think maybe the discussion in https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html#avoid-unstable-typing-mechanisms is still applicable.

NeilGirdhar:
@jakevdp No worries.  I thought we were just waiting for MyPy.  I didn't realize that we were waiting for Pytype as well!

NeilGirdhar:
@jakevdp Given [PyType's announcement](https://github.com/google/pytype/blob/main/README.md), looks like PyType will never support ParamSpec.  Should we take another look at this?

jakevdp:
Possibly, but maybe sometime later this fall. Our type checking story will likely be changing in the coming months (e.g. folks on the team are excited about `ty`) but it's too early to know exactly where we'll land.

NeilGirdhar:
Sounds great.  I'm also excited about ty!

NeilGirdhar:
> Possibly, but maybe sometime later this fall. Our type checking story will likely be changing in the coming months (e.g. folks on the team are excited about `ty`) but it's too early to know exactly where we'll land.

Any update on typing in Jax?  I'm using ty now for my own code.

jakevdp:
We've landed on pyrefly as our next typing solution. We've already replaced mypy with pyrefly for github CI checks; we still have plans to migrate away from pytype for other tests, but are not there yet.

NeilGirdhar:
Great!  Sorry to nag you about this over so many years!  pyrefly seems very capable (it even has intersections), so I guess when you're able to move away from pytype, we can hopefully revisit this :smiley:

jakevdp:
Hey @NeilGirdhar – a happy update: we now have dropped pytype and mypy, and typecheck our project via pyrefly which I believe has better support for ParamSpec and related constructs.

Would you like to jump back in on this change?

NeilGirdhar:
Sure!

NeilGirdhar:
@jakevdp Ready for your review! Here’s the commit message:

Type `jit` precisely, fix resulting errors

Give `jax.jit`'s overloads a `ParamSpec` and covariant return type instead of erasing signatures to `Any`. This makes Pyrefly precise enough to catch latent mismatches elsewhere; fixed at the source:

* `lu_solve`: narrowed to `Array` since `_lu_solve` reads `.shape`/`.ndim`.
* `_cumprod`/`_cumsum`: narrowed `axis` to `int | None`; single-axis only.
* `_chunk_iter`: builtin `divmod`, so slice size stays a static `int`.
* `_flip`: reuse the shared `Axis` alias.
* `_squeeze`/`_pad`: widened `axis`/`stat_length` to allow `None`.
* `digitize`: `TypedDict` for `**kwds` so splatting type-checks.
* `norm`/`vector_norm`: widened `ord` to include `float`.
* `ufunc.__call__`: cast to `Callable`; `jit` can't express bound-self.
* `schur`/`_sqrtm`/`sqrtm`: narrowed to `Array` since `_schur` calls `.astype`.
* `hessenberg`/`hadamard`: `jnp.concatenate` instead of nested block.
* `kde`: `rowvar=True` instead of `1`, since `rowvar` is typed `bool`.
* `eigh`: assert `subset_by_index` is not `None` before indexing.

jax/_src/lax/linalg.py
```diff
@@ -1796,8 +1796,7 @@ def _lu_tpu_lowering_rule(ctx, operand):
 register_cpu_gpu_lowering(lu_p, _lu_cpu_gpu_lowering)
 
 
-def lu_solve(lu: ArrayLike, permutation: ArrayLike, b: ArrayLike,
-             trans: int = 0) -> Array:
+def lu_solve(lu: Array, permutation: Array, b: Array, trans: int = 0) -> Array:
```
jakevdp: Public APIs should annotate their inputs as `ArrayLike`. This should be fixed by wrapping the inputs in `lax.asarray` before passing them to `_lu_solve`.

jax/_src/scipy/linalg.py
```diff
@@ -509,7 +509,7 @@ def _schur(a: Array, output: str) -> tuple[Array, Array]:
     a = a.astype(dtypes.to_complex_dtype(a.dtype))
   return lax_linalg.schur(a)
 
-def schur(a: ArrayLike, output: str = 'real') -> tuple[Array, Array]:
+def schur(a: Array, output: str = 'real') -> tuple[Array, Array]:
```
jakevdp: Same here: this should remain `ArrayLike`, but be pased through `ensure_arraylike` in order to convert to `Array` before passing to `_schur`.

jax/_src/scipy/linalg.py
```diff
@@ -2169,14 +2169,14 @@ def j_loop(j, U):
...
-def _sqrtm(A: ArrayLike) -> Array:
+def _sqrtm(A: Array) -> Array:
   T, Z = schur(A, output='complex')
   sqrt_T = _sqrtm_triu(T)
   return jnp.matmul(jnp.matmul(Z, sqrt_T, precision=lax.Precision.HIGHEST),
                     jnp.conj(Z.T), precision=lax.Precision.HIGHEST)
 
 
-def sqrtm(A: ArrayLike, blocksize: int = 1) -> Array:
+def sqrtm(A: Array, blocksize: int = 1) -> Array:
```
jakevdp: Same comment

jax/_src/scipy/linalg.py
```diff
@@ -2406,9 +2406,14 @@ def hessenberg(a: ArrayLike, *, calc_q: bool = False, overwrite_a: bool = False,
   if calc_q:
     q = lax_linalg.householder_product(a_out[..., 1:, :-1], taus)
     batch_dims = a_out.shape[:-2]
-    q = jnp.block([[jnp.ones(batch_dims + (1, 1), dtype=a_out.dtype),
-                    jnp.zeros(batch_dims + (1, n - 1), dtype=a_out.dtype)],
-                   [jnp.zeros(batch_dims + (n - 1, 1), dtype=a_out.dtype), q]])
+    top = jnp.concatenate(
```
jakevdp: `jnp.block` has the correct runtime semantics here – can we somehow fix this via typing-only changes?

jax/_src/api.py
```diff
@@ -176,7 +179,7 @@ def __repr__(self):
 
 @overload
 def jit(
```
jakevdp: Can we switch to Python 3.12-style definitions here, i.e. `def jit[Q, **V](...)`?

Same for the class definitions below.

jakevdp (changes requested):

jax/_src/scipy/linalg.py
NeilGirdhar: Yeah, maybe, but one of the reasons I went this way was to move towards the Array API's interface, which is simpler and easier to understand.  The Array API has no `block`, but it does have `concat`.  Future readers of the code who (unlike us) are less familiar with historical numpy functions like `block` might find this less confusing?

Otherwise, we have to extend `block`'s annotation to support, say, lists of lists of arrays.

Let me know what you want to do.

jax/_src/api.py
NeilGirdhar: Yes, thanks for catching that!

NeilGirdhar:
Also, because it's easy to miss: please take a look at this line: `num_chunks, tail = divmod(x.shape[0], size)` where I changed it from `ufuncs.divmod` since it seems liek the arguments are always integers, and the array return value was not accepted by downstream operations.

jax/_src/scipy/linalg.py
jakevdp: I would lean toward updating `block`'s annotation here – the current one seems incorrect! I don't know if there's any way to make it fully correct (i.e. annotate an arbitrarily nested list of `ArrayLike`) but at the very least we can go one step further and use `ArrayLike | list[ArrayLike] | list[list[ArrayLike]]`

jakevdp:
> please take a look at this line: `num_chunks, tail = divmod(x.shape[0], size)`

The change looks correct to me.

jax/_src/scipy/linalg.py
NeilGirdhar: Unforunately, `list[list[Array]]` is not assignable to a parameter of type `list[list[ArrayLike]]`.
In fact, `list[Array]` is not even assignable to a parameter of type `list[ArrayLike]`.
The problem is list invariance: the caller doesn't know that the function won't append to the list.

The options I think are to switch block's annotation to:
1. `list[Any]`
2. `list[list[Array]] | ...`.

I went with 1, but let me know if you want the narrower 2.

jax/_src/scipy/linalg.py
jakevdp: Looks like this didn't make it into the updated branch

NeilGirdhar: Sorry, fixed.

jakevdp:
One more comment on argument annotations for `schur`. Next I'd appreciate if you could move the fixes into another PR (maybe a good time to try out the new [stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) if you'd like!). Those can be merged independently to keep this change limited to the actual JIT annotation updates. Thanks!

NeilGirdhar:
Okay, I used stacked PRs as requested :smile:

jakevdp:
#39749 is merged, please rebase!

NeilGirdhar:
Done!

By the way, `trace` and `lower` don't need the same annotations, or do they?

jakevdp:
Oh, I expected that the only remaining changes here would be in `api.py`, `pjit.py`, and `stages.py`. Could we split everything else into a separate PR?

NeilGirdhar:
Okay.  See my previous question btw.

jakevdp:
> By the way, `trace` and `lower` don't need the same annotations, or do they?

That one is a bit tricky, because `trace` and `lower` can take things like `ShapeDtypeStruct` in place of arrays, so annotating them with the same signature as the wrapped function could lead to false positives. Probably better to keep them as raw `*args` and `**kwargs` for now.

NeilGirdhar:
(Rebased!)

jakevdp (approved):
Looks good!

I'll pull in the change for further testing. Since this effectively tightens annotations for any JIT-compiled function, I suspect it will require some downstream fixes before we can merge.

NeilGirdhar:
Awesome!  Thanks, for the quick and patient reviews!

jax/_src/api.py
```diff
@@ -175,8 +175,8 @@ def __repr__(self):
     return "<not-specified>"
 
 @overload
-def jit(
-  fun: Callable, /, *,
+def jit[**Q, V_co](
```
jakevdp: One potential confusion here is the naming: we use `Q, V_co` here and `P, V_co` elsewhere. We should use a common notation, but also what would you think about more descriptive names? e.g. something like:
```python
def jit[**InputType, OutputType]:
  ...
```
I think that might be more maintainable in the long run, because it would be easier to find locations where these are used.

What do you think?

NeilGirdhar: I'm happy to do whatever you like.  In the context of type variable names, the most common names are:
```python
P = ParamSpec("P")
R = TypeVar("R")
```
Unfortunately `P` was taken.  If you want more descriptive names, I propose:
```py
Parameters = ParamSpec("Parameters")  # It's not really an "input type"
ReturnType = TypeVar("ReturnType")
```

jakevdp: They're no longer explicitly defined though, so it would just be `def jit[**Paramters, ReturnType]` yes?

I think that would be a good convention to use.

NeilGirdhar: Yes, exactly!   Okay, I'll switch to that now.

jakevdp:
There seems to be an issue here: this leads to false positives when JIT wraps a method. For example:
```python
# failure.py
import jax

class Foo:
    @jax.jit(static_argnums=0)
    def bar(self, x: int) -> int:
        return x * 2
    
Foo().bar(2)
```
```
$ pyrefly check failure.py
ERROR Missing argument `x` in function `jax._src.stages.Wrapped.__call__` [missing-argument]
 --> failure.py:8:10
  |
8 | Foo().bar(2)
  |          ^^^
  |
ERROR Argument `Literal[2]` is not assignable to parameter `self` with type `Foo` in function `jax._src.stages.Wrapped.__call__` [bad-argument-type]
 --> failure.py:8:11
  |
8 | Foo().bar(2)
  |           ^
  |
 INFO 2 errors
```

jakevdp:
I think the fix is to add a descriptor protocol to `JitWrapped`; something like this:
```python
class JitWrapped[**P, R](stages.Wrapped[P, R]):

  def eval_shape(self, *args, **kwargs):
    """See ``jax.eval_shape``."""
    raise NotImplementedError

  def trace(self, *args, **kwargs) -> stages.Traced:
    raise NotImplementedError

  @overload
  def __get__(self, instance: None, owner: type) -> JitWrapped[P, R]: ...

  @overload
  def __get__(self, instance: object, owner: type | None = None) -> Callable[..., R]: ...

  def __get__(self, instance: object | None, owner: type | None = None):
    raise NotImplementedError
```
Note that `-> Callable[..., R]` rather than `-> Callable[P, R]` in the second overload is intended: this is the case where we need to tell the typechecker that the bound method does not take `self` as an argument, and `self` is part of `P`.

This does mean that parameter types for jit-wrapped methods cannot be type-checked, but I'm not sure there's any way around that with Python's current typing model.

NeilGirdhar:
Yes, you got it.  BTW, I proposed [a typing fix to this](https://github.com/python/typing/issues/1372) a few years ago, but it didn't find enough traction.  It might be time to revisit this :smile:

jakevdp:
Implementing this change and building a bunch of downstream projects, I'm finding a few common usages that are now marked as type errors with the new annotations.

## Overriding a non-jit method with a jit method (and vice versa)
This one comes up a lot: a base class method or descriptor can only be overridden by a definition of the same type. Previously the type system did not know that `jax.jit` produced a descriptor, but with this change, common usages are marked as type errors.

Simple example:
```python
# example.py
import jax

class Base:
    def method(self):
        ...

class Derived(Base):
    @jax.jit(static_argnums=0)
    def method(self):
        ...
```
```
$ pyrefly check example.py
ERROR Class member `Derived.method` overrides parent class `Base` in an inconsistent manner [bad-override]
 --> example.py:9:9
  |
9 |     def method(self):
  |         ^^^^^^
  |
  `Derived.method` and `Base.method` must both be descriptors
 INFO 1 error
```
I suspect this will be very painful for users, because JIT-compilation of methods in derived classes is very common (and the error is pretty inscrutible unless you happen to know that JIT returns a descriptor).

## Attempting to overwrite a JIT-compiled method
This one is less important I think, but there's a common pattern people use of avoiding the "static self" complications by overwriting methods with jit-compiled bound versions of themselves within the `__init__` function. If they do this *and also* jit compile the method itself, it now leads to a type check error.
```python
import jax

class Example:
    def __init__(self):
        self.method = jax.jit(self.method)

    @jax.jit(static_argnums=0)
    def method(self):
        ...
```
```
$ pyrefly check example.py
ERROR Attribute `method` of class `Example` is a read-only descriptor with no `__set__` and cannot be set [read-only]
 --> example.py:5:9
  |
5 |         self.method = jax.jit(self.method)
  |         ^^^^^^^^^^^
  |
 INFO 1 error
```
I think this may be a pyrefly error, because at runtime a descriptor with no `__set__` method can simply be overwritten with no error.

NeilGirdhar:
> ## Attempting to overwrite a JIT-compiled method
> This one is less important I think, but there's a common pattern people use of avoiding the "static self" complications by overwriting methods with jit-compiled bound versions of themselves within the `__init__` function. If they do this _and also_ jit compile the method itself, it now leads to a type check error.
> 
> ```python
> import jax
> 
> class Example:
>     def __init__(self):
>         self.method = jax.jit(self.method)
> 
>     @jax.jit(static_argnums=0)
>     def method(self):
>         ...
> ```
> 
> ```
> $ pyrefly check example.py
> ERROR Attribute `method` of class `Example` is a read-only descriptor with no `__set__` and cannot be set [read-only]
>  --> example.py:5:9
>   |
> 5 |         self.method = jax.jit(self.method)
>   |         ^^^^^^^^^^^
>   |
>  INFO 1 error
> ```
> 
> I think this may be a pyrefly error, because at runtime a descriptor with no `__set__` method can simply be overwritten with no error.

I realize that Jax has been out for a long time, and as a low level library, various people have come up with different ways of working with it.  However, this usage pattern is horrifying.

First, why would you both jit the unbound method, and also jit it after it's bound?  Yuck.

Second, I imagine the reason `self` is being treated as static is because it contains static members.  Having to carefully mark objects of type `Example` as static every time you pass them is error-prone.  It also mans you can never add a non-static field to `Example`.  I think the universally better approach is to mark the static fields as static, using [some kind of dataclass](https://github.com/NeilGirdhar/tjax/blob/2951efefec8a85a982afc0db7f1b074d1e7ddda8/tjax/_src/dataclasses/dataclass.py#L43).

Also, the pattern of `self.method = jax.jit(self.method)` is bad in general.  It means that you recompile `self.method` for every instance.  Even in the static case, it's bad because you may have two separate instances with identical data—these should not be recompiled.  And yet they will be with this construct.

> ```
> $ pyrefly check example.py
> ERROR Class member `Derived.method` overrides parent class `Base` in an inconsistent manner [bad-override]
>  --> example.py:9:9
>   |
> 9 |     def method(self):
>   |         ^^^^^^
>   |
>   `Derived.method` and `Base.method` must both be descriptors
>  INFO 1 error
> ```
>   
> 
> I suspect this will be very painful for users, because JIT-compilation of methods in derived classes is very common (and the error is pretty inscrutible unless you happen to know that JIT returns a descriptor).

Interesting.  Ty doesn't have a problem with this.  I suggest this is a bug with Pyrefly not resolving the return type of `__get__`.  Do you want to file a bug or should I?

jakevdp:
> I realize that Jax has been out for a long time, and as a low level library, various people have come up with different ways of working with it. However, this usage pattern is horrifying.

Agreed this is pretty terrible – but pretty common unfortunately. Maybe we should count this regression as a win 😁 

> Interesting. Ty doesn't have a problem with this. I suggest this is a bug with Pyrefly not resolving the return type of `__get__`. Do you want to file a bug or should I?

That would be great if you could file the bug!

NeilGirdhar:
Done: https://github.com/facebook/pyrefly/issues/4494

> Agreed this is pretty terrible – but pretty common unfortunately. Maybe we should count this regression as a win 😁

:smile: yep!

jakevdp:
> Done: [facebook/pyrefly#4494](https://github.com/facebook/pyrefly/issues/4494)

That's a solid bug report

NeilGirdhar:
FYI: Someone posted [the typing extension](https://discuss.python.org/t/proposal-callable-variant-types/108648), and it's being discussed.

NeilGirdhar:
> This one comes up a lot: a base class method or descriptor can only be overridden by a definition of the same type.

Looks like this is fixable thanks to a nice solution by @jorenham.   This should work when Pyrefly fixes [the bug I filed](https://github.com/facebook/pyrefly/issues/4592).

PR #14696, opened 2023-02-27
[shape_poly] Add support for jnp.cum{sum,prod,max,min} with shape polymorphism

gnecula:
Unfortunately, on CPU and GPU where we use associative scan, we cannot support shape polymorphism with native lowering, when the axis over which we reduce is non-constant size. Improve the error in that case.

### DIFF
--- a/jax/experimental/jax2tf/tests/shape_poly_test.py
+++ b/jax/experimental/jax2tf/tests/shape_poly_test.py
@@ -1855,6 +1855,19 @@ _POLY_SHAPE_TEST_HARNESSES = [
                 lambda x: lax_control_flow.cummax(x, axis=1, reverse=False),
                 arg_descriptors=[RandArg((3, 4, 5), _f32)],
                 poly_axes=[0]),
+    PolyHarness("jnp.cumsum", "reduce_axis=poly",
+                lambda x: jnp.cumsum(x, axis=0),
+                arg_descriptors=[RandArg((3, 5), _f32)],
+                poly_axes=[0],
+                expect_error=(
+                  (None, None) if (not config.jax2tf_default_experimental_native_lowering or
+                                   jtu.device_under_test() == "tpu") else
+                  (NotImplementedError,
+                   "associative scan over axis of non-constant size"))),
+    PolyHarness("jnp.cumsum", "reduce_axis=static",
+                lambda x: jnp.cumsum(x, axis=1),
+                arg_descriptors=[RandArg((3, 5), _f32)],
+                poly_axes=[0]),
     PolyHarness("delta", "0",
                 lambda x: lax_internal._delta(_f32, x.shape, axes=(0, 1)) + x,
                 arg_descriptors=[RandArg((3, 1), _f32)],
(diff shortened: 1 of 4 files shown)


gnecula:
@rxsang @maxwillzq PTAL

jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1855,6 +1855,19 @@ def f_jax(x):
                 lambda x: lax_control_flow.cummax(x, axis=1, reverse=False),
                 arg_descriptors=[RandArg((3, 4, 5), _f32)],
                 poly_axes=[0]),
+    PolyHarness("jnp.cumsum", "reduce_axis=poly",
+                lambda x: jnp.cumsum(x, axis=0),
+                arg_descriptors=[RandArg((3, 5), _f32)],
+                poly_axes=[0],
+                expect_error=(
+                  (None, None) if (not config.jax2tf_default_experimental_native_lowering or
```
maxwillzq, line 1863: Question:   I do not understand why need check "native_lowering" here.  Since the error message on loops.py only tell users "avoid to run it on TPU".

jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -2567,6 +2580,10 @@ def test_harness(self, harness: PolyHarness):
         raise unittest.SkipTest(
             "native lowering with shape polymorphism not implemented for JAX primitives still using HLO fallback lowering; b/261682623")
 
+      if harness.fullname == "jnp.cumsum_reduce_axis=poly" and jtu.device_under_test() == "tpu":
+        # https://github.com/openxla/stablehlo/issues/1258
+        raise unittest.SkipTest(
+            "native lowering with shape polymorphism not implemented for window_reductions on TPU")
```
maxwillzq, line 2586: same question as before.  Since shape_poly_test.py can run both non_native_lowering and native_lowering.   so Non-native-lowering with TPU will also skip this unit test. 
Do we need mention "native_lowering" in the error message?

jax/experimental/jax2tf/tests/shape_poly_test.py
gnecula: We need the check here because we want to capture the precise condition under which we get an error: native lowering and device != "TPU". Note that if I change the condition the test will fail, because an expected error is not appearing.

The error message could be more precise: you get this error if you try to use associative scan with shape polymorphism. For non-native lowering, jax2tf does not use associative scan (and jax2tf is independent of the platform). However, there is the flag `--jax2tf_associative_scan_reductions` that makes jax2tf non-native use associative scan.

With native lowering you get platform-dependent behavior, just like JAX native: will use associative scan for CPU and GPU.

jax/experimental/jax2tf/tests/shape_poly_test.py
gnecula: This whole conditional is under a bigger conditional that only applies on native lowering.

maxwillzq (approved):

rxsang (approved):

PR #14702, opened 2023-02-27
attach debug info to jaxpr, pass to mlir/mhlo

mattjj:
image

There are actually two changes entailed here:
* add argument info as a `"jax.arg_info"` attribute on the MHLO computations we generate, as shown in the screenshot above
* change some error messages about tracers to have pytree key paths, e.g. `jax.jit(lambda x: x['a'] if x['b'] > 0 else 0)` now complains about `x['b']` not just `x`

The latter is a consequence of reusing (and updating) the machinery from the latter to implement the former.

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -401,9 +401,15 @@ def saved_residuals(f, *args, **kwargs) -> List[Tuple[core.AbstractValue, str]]:
       results.append((v.aval, 'from a constant'))
 
   assert len(jaxpr.invars) == len(in_leaves)
+  dbg = pe.debug_info(f, in_tree, True, "saved_residuals")
+  arg_info = pe.arg_info_all(dbg)
   for i, v in enumerate(jaxpr.invars):
     if v in res_vars:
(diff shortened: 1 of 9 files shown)


tests/metadata_test.py
```diff
@@ -39,10 +39,10 @@ class MetadataTest(jtu.JaxTestCase):
   def test_jit_metadata(self):
     hlo = module_to_string(jax.jit(jnp.sin).lower(1.).compiler_ir())
     self.assertRegex(hlo, r'loc\("jit\(sin\)/jit\(main\)/sin"')
-    def foo(x):
-      return jnp.sin(x)
-    hlo = module_to_string(jax.jit(foo).lower(1.).compiler_ir())
-    self.assertRegex(hlo, r'loc\("jit\(foo\)/jit\(main\)/sin"')
+    # def foo(x):
```
hawkinsp: Intentional?

hawkinsp (approved):

tests/metadata_test.py
mattjj: No, accidental, fixed.

PR #14711, opened 2023-02-28
[jax2tf] Clean up the support for cross-lowering.

gnecula:
In #14587 we introduced cross-lowering support without any changes in JAX core, but at the expense of some overly complex code in jax2tf, along with overriding a JAX core function. Plus, those changes were not enough to handle some xmap and pmap cases.

Here we introduce a `_experimental_lowering_platform: Optional[str]` parameter to the `.lower()` methods and then we thread the `lowering_platform` all the way to the calls to `mlir.lower_jaxpr_to_module2`. That's it.

Note that this parameter to `.lower()` is experimental and not supposed to be used outside jax2tf. It may also gobble user kwargs.

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -789,7 +789,8 @@ def _jit_lower(fun, static_argnums, static_argnames, device, backend,
       return aval, device
 
   @api_boundary
-  def lower(*args, **kwargs) -> stages.Lowered:
+  def lower(*args, _experimental_lowering_platform: Optional[str] = None,
+            **kwargs) -> stages.Lowered:
     """Lower this function for the given arguments.
 
     A lowered function is staged out of Python and translated to a
@@ -823,13 +824,15 @@ def _jit_lower(fun, static_argnums, static_argnames, device, backend,
     if jax.config.jax_array:
       computation = dispatch.sharded_lowering(
           flat_fun, device, backend, flat_fun.__name__, donated_invars, True,
-          keep_unused, *arg_specs_and_devices)
+          keep_unused, lowering_platform=_experimental_lowering_platform,
+          *arg_specs_and_devices)
       return stages.Lowered.from_flat_info(
           computation, in_tree, in_avals, donate_argnums, out_tree())
     else:
       computation = dispatch.lower_xla_callable(
           flat_fun, device, backend, flat_fun.__name__, donated_invars, True,
-          keep_unused, *arg_specs_and_devices)
+          keep_unused, lowering_platform=_experimental_lowering_platform,
+          *arg_specs_and_devices)
       return stages.Lowered.from_flat_info(
           computation, in_tree, in_avals, donate_argnums, out_tree())
 
@@ -2471,7 +2474,8 @@ def _pmap_lower(fun, axis_name, in_axes, out_axes, static_broadcasted_tuple,
   # this might naturally be a method, with ``fun`` as a ``self`` and
   # all the other arguments stored as attributes.
   @api_boundary
-  def lower(*args, **kwargs) -> stages.Lowered:
+  def lower(*args, _experimental_lowering_platform: Optional[str] = None,
+            **kwargs) -> stages.Lowered:
     """Lower a parallel-mapped form of this function for the given arguments.
 
     A parallel-mapped and lowered function is staged out of Python and
@@ -2497,7 +2501,8 @@ def _pmap_lower(fun, axis_name, in_axes, out_axes, static_broadcasted_tuple,
         donated_invars=p.donated_invars,
         global_arg_shapes=p.global_arg_shapes_flat,
         is_explicit_global_axis_size=p.is_explicit_global_axis_size,
-        avals=abstract_args)
+        avals=abstract_args,
+        lowering_platform=_experimental_lowering_platform)
     return stages.Lowered.from_flat_info(
         computation, p.in_tree, abstract_args, donate_tuple, p.out_tree())
 
--- a/jax/_src/dispatch.py
+++ b/jax/_src/dispatch.py
@@ -326,7 +326,8 @@ def not_none_device_or_backend_on_jit(backend, device, num_ins):
 
 
 def sharded_lowering(fun, device, backend, name, donated_invars, always_lower,
-                     keep_unused, *arg_specs):
+                     keep_unused, *arg_specs,
+                     lowering_platform: Optional[str]):
   in_avals, in_shardings = util.unzip2(arg_specs)
 
   da = None
@@ -334,7 +335,7 @@ def sharded_lowering(fun, device, backend, name, donated_invars, always_lower,
     da, in_shardings = not_none_device_or_backend_on_jit(
         backend, device, len(in_shardings))
 
-  in_shardings = [pxla._UNSPECIFIED if i is None else i for i in in_shardings]
+  in_shardings = [pxla._UNSPECIFIED if i is None else i for i in in_shardings]  # type: ignore
 
   # Pass in a singleton `_UNSPECIFIED` for out_shardings because we don't know
   # the number of output avals at this stage. lower_sharding_computation will
@@ -342,19 +343,22 @@ def sharded_lowering(fun, device, backend, name, donated_invars, always_lower,
   return pxla.lower_sharding_computation(
       fun, 'jit', name, in_shardings, pxla._UNSPECIFIED, donated_invars,
       in_avals, in_is_global=(True,) * len(arg_specs), keep_unused=keep_unused,
-      always_lower=always_lower, devices_from_context=da)
+      always_lower=always_lower, devices_from_context=da,
+      lowering_platform=lowering_platform)
 
 
 def _xla_callable_uncached(fun: lu.WrappedFun, device, backend, name,
                            donated_invars, keep_unused, *arg_specs):
   if config.jax_array:
     computation = sharded_lowering(fun, device, backend, name, donated_invars,
-                                   False, keep_unused, *arg_specs)
+                                   False, keep_unused, *arg_specs,
+                                   lowering_platform=None)
     allow_prop = [True] * len(computation.compile_args['global_out_avals'])
     return computation.compile(_allow_propagation_to_outputs=allow_prop).unsafe_call
   else:
     return lower_xla_callable(fun, device, backend, name, donated_invars, False,
-                              keep_unused, *arg_specs).compile().unsafe_call
+                              keep_unused, *arg_specs,
+                              lowering_platform=None).compile().unsafe_call
 
 xla_callable = lu.cache(_xla_callable_uncached)
 
@@ -414,7 +418,8 @@ def raise_warnings_or_errors_for_jit_of_pmap(nreps, backend, name, jaxpr):
 @profiler.annotate_function
 def lower_xla_callable(
     fun: lu.WrappedFun, device, backend, name, donated_invars,
-    always_lower: bool, keep_unused: bool, *arg_specs):
+    always_lower: bool, keep_unused: bool, *arg_specs,
+    lowering_platform: Optional[str]):
   """Lower into XLA.
 
   Args:
@@ -512,7 +517,8 @@ def lower_xla_callable(
     effects.ordered_effects.filter_in(closed_jaxpr.effects))
   lowering_result = mlir.lower_jaxpr_to_module(
       module_name, closed_jaxpr, unordered_effects,
-      ordered_effects, backend, backend.platform,
+      ordered_effects, backend,
+      lowering_platform or backend.platform,
       mlir.ReplicaAxisContext(axis_env), name_stack, donated_invars)
   module, keepalive, host_callbacks = (
       lowering_result.module, lowering_result.keepalive,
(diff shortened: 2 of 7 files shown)


jax/_src/dispatch.py
```diff
@@ -326,35 +326,37 @@ def not_none_device_or_backend_on_jit(backend, device, num_ins):
 
 
 def sharded_lowering(fun, device, backend, name, donated_invars, always_lower,
-                     keep_unused, *arg_specs):
+                     keep_unused, *arg_specs,
+                     lowering_platform: Optional[str]):
```
yashk2810, line 330: Can you please move this before *arg_specs?

jax/_src/dispatch.py
```diff
@@ -326,35 +326,37 @@ def not_none_device_or_backend_on_jit(backend, device, num_ins):
...
     computation = sharded_lowering(fun, device, backend, name, donated_invars,
-                                   False, keep_unused, *arg_specs)
+                                   False, keep_unused, *arg_specs,
+                                   lowering_platform=None)
     allow_prop = [True] * len(computation.compile_args['global_out_avals'])
     return computation.compile(_allow_propagation_to_outputs=allow_prop).unsafe_call
   else:
     return lower_xla_callable(fun, device, backend, name, donated_invars, False,
-                              keep_unused, *arg_specs).compile().unsafe_call
+                              keep_unused, *arg_specs, lowering_platform=None).compile().unsafe_call
```
yashk2810, line 359: 80 lines

jax/_src/dispatch.py
```diff
@@ -326,35 +326,37 @@ def not_none_device_or_backend_on_jit(backend, device, num_ins):
...
+  in_shardings = [pxla._UNSPECIFIED if i is None else i for i in in_shardings]  # type: ignore
 
   # Pass in a singleton `_UNSPECIFIED` for out_shardings because we don't know
   # the number of output avals at this stage. lower_sharding_computation will
   # apply it to all out_avals.
   return pxla.lower_sharding_computation(
       fun, 'jit', name, in_shardings, pxla._UNSPECIFIED, donated_invars,
       in_avals, in_is_global=(True,) * len(arg_specs), keep_unused=keep_unused,
-      always_lower=always_lower, devices_from_context=da)
+      always_lower=always_lower, devices_from_context=da, lowering_platform=lowering_platform)
```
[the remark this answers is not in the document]
yashk2810: SOrry 80 character limit

jax/_src/dispatch.py
```diff
@@ -414,7 +416,8 @@ def raise_warnings_or_errors_for_jit_of_pmap(nreps, backend, name, jaxpr):
 @profiler.annotate_function
 def lower_xla_callable(
     fun: lu.WrappedFun, device, backend, name, donated_invars,
-    always_lower: bool, keep_unused: bool, *arg_specs):
+    always_lower: bool, keep_unused: bool, *arg_specs,
+    lowering_platform: Optional[str]):
```
yashk2810, line 420: experimental_lowering_platform

jax/_src/interpreters/pxla.py
```diff
@@ -1271,7 +1271,7 @@ def parallel_callable(fun: lu.WrappedFun,
   pmap_computation = lower_parallel_callable(
       fun, backend_name, axis_name, axis_size, global_axis_size, devices, name,
       in_axes, out_axes_thunk, donated_invars, global_arg_shapes,
-      is_explicit_global_axis_size, avals)
+      is_explicit_global_axis_size, avals, lowering_platform=None)
```
yashk2810, line 1274: experimental_lowering_platform (here and everywhere else)

jax/_src/interpreters/pxla.py
```diff
@@ -2874,10 +2876,12 @@ def lower_sharding_computation(
     out_shardings: Union[Sequence[Union[sharding_internal.XLACompatibleSharding, UnspecifiedValue]], UnspecifiedValue],
     donated_invars: Sequence[bool],
     global_in_avals: Sequence[core.ShapedArray],
+    *,
     in_is_global: Sequence[bool],
     keep_unused: bool,
     always_lower: bool,
-    devices_from_context: Optional[Sequence[xc.Device]] = None
+    lowering_platform: Optional[str],
```
yashk2810, line 2883: experimental_lowering_platform

Also please put this after devices_from_context

jax/_src/pjit.py
```diff
@@ -1385,7 +1386,9 @@ def _pjit_lower_cached(
     name: str,
     in_is_global: Sequence[bool],
     keep_unused: bool,
-    always_lower: bool):
+    *,
```
yashk2810, line 1389: Please remove the star

Also experimental_lowering_platform

jax/_src/pjit.py
```diff
@@ -1442,7 +1446,8 @@ def _pjit_lower_cached(
         jaxpr.in_avals, in_is_global=in_is_global, keep_unused=keep_unused,
         always_lower=always_lower,
         devices_from_context=(
-            None if mesh is None or mesh.empty else list(mesh.devices.flat)))
+            None if mesh is None or mesh.empty else list(mesh.devices.flat)),
+        lowering_platform=lowering_platform)
```
yashk2810, line 1450: I don't know how your tests passes but you read this in the wrong order in pxla.py in lower_sharding_computation

jax/_src/pjit.py
```diff
@@ -1385,7 +1386,9 @@ def _pjit_lower_cached(
     name: str,
     in_is_global: Sequence[bool],
     keep_unused: bool,
-    always_lower: bool):
+    *,
+    always_lower: bool,
+    lowering_platform: Optional[str]):
```
yashk2810, line 1391: Just default to None here so that you don't have to thread it from line 1293 too?

jax/_src/maps.py
```diff
@@ -644,7 +644,9 @@ def make_xmap_callable(fun: lu.WrappedFun,
                        in_axes, out_axes_thunk, donated_invars,
                        global_axis_sizes, axis_resources, resource_env, backend,
                        spmd_in_axes, spmd_out_axes_thunk, in_positional_semantics,
-                       out_positional_semantics, *in_avals):
+                       out_positional_semantics,
+                       lowering_platform: Optional[str],
```
yashk2810, line 648: Default to None here directly so that on lin 634 you don't have to pass None?

jax/_src/pjit.py
```diff
@@ -364,7 +364,7 @@ def post_infer_params(fun, infer_params_fn, static_argnums, static_argnames,
     wrapped = _python_pjit(fun, infer_params_fn)
 
   @api_boundary
-  def lower(*args, **kwargs):
+  def lower(*args, _experimental_lowering_platform: Optional[str] = None, **kwargs):
```
yashk2810, line 367: Please follow 80 character limit

jax/_src/dispatch.py
gnecula: Actually, I call it "experimental" only in the `lower()` API, which is public. The rest are internal APIs, and I think it mostly adds clutter.

yashk2810: I think it adds clarification for us (jax devs) as to what it is supposed to be. So I think experimental is a good choice for variables too

jax/_src/interpreters/pxla.py
```diff
@@ -2874,10 +2876,12 @@ def lower_sharding_computation(
     out_shardings: Union[Sequence[Union[sharding_internal.XLACompatibleSharding, UnspecifiedValue]], UnspecifiedValue],
     donated_invars: Sequence[bool],
     global_in_avals: Sequence[core.ShapedArray],
+    *,
```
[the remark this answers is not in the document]
gnecula: The star is simply to force this to be a kwarg, and to make it a mistake to pass is as positional. It seems that many of the arguments we pass as keyword and forcing that is safer, because you do not allow Python to mistakenly interpret the as positional. wdyt?

jax/_src/pjit.py
gnecula: This works because there is a "*" and it forces those arguments to work as kwargs and then the order does not matter.

yashk2810: Ooff.. that's confusing. Can we keep the `*` change to a separate CL and not do it here?

jax/_src/pjit.py
gnecula: I prefer to not have defaults because those can hide bugs. If there is no default value, it forces you to think what to do at each call site, and there aren't that many. I would rather not miss one. As you can see there are only a handful of places where I need to pass `None`

yashk2810: Ok, sounds good. There is one more place you are missing in pjit then in _pjit_partial_eval.

jax/_src/api.py
```diff
@@ -788,7 +788,7 @@ def arg_spec(x):
       return aval, device
 
   @api_boundary
-  def lower(*args, **kwargs) -> stages.Lowered:
+  def lower(*args, _experimental_lowering_platform: Optional[str] = None, **kwargs) -> stages.Lowered:
```
froystig, line 791: Just to note: we could make this an option to `jit` (maybe `_experimental_aot_lowering_platform`) to avoid kwarg collision here. We don't have to worry so long as the name is so long and hidden, but we should keep that in mind if this approach graduates into something more official.

jax/_src/dispatch.py
froystig: Maybe a comment can be a happy medium? For a jax dev audience, I'm less concerned at this point with labeling it experimental, and more concerned with it not being obvious what it is supposed to mean.

froystig (approved):

jax/_src/api.py
yashk2810: I don't really like adding temp arguments to jit. Since it is temporary, we can keep it on lower and then if we decide on an approach, we can pull this up to `jit`.

Adding it to `jit` has more implications (even in terms of code and threading it through) that I feel this approach currently will satisfy George's needs and keep the implementation fairly clean. WDYT?

froystig: Yep, this was a remark for our potential future selves.

jax/_src/dispatch.py
yashk2810: The problem is that there are so many codepaths currently that you would have to add a comment everywhere for it to be obvious what it means.

This code in dispatch.py is dead code (lol) and a lot of other places this CL changes. It will be deleted after March 15. That's why if the variable name also contains experimental, then it is obvious what is going on.

yashk2810: Anyways, not a big deal just wanted to see if it was reasonable. 

If you feel strongly, then lowering_platform is fine.

jax/_src/api.py
gnecula: I also do not think it should be in `jit` because that is for lowering-compilation-execution. But it can be in an intermediate place: `jit().with_lowering_platform('tpu').lower(*args)`. We could do that if this ever "graduates", but perhaps it won't ever.

jax/_src/dispatch.py
gnecula: It is after `*arg_specs` for a very precise reason: I want it to be kwarg. With that, and without a default value, you get a very clear error at a callsite where you forgot to set it that the function expects a `lowering_platform` argument. If we add it before `*arg_specs` then it will get confused with an argument and the error will be a lot more confusing. This has been very helpful when making this change and fixing all the places that need to be fixed. It is helpful also when people use these internal apis from other projects (e.g., search for `_pjit_lower(` in Google code base). Think also when we will remove this argument later on, if we make it positional and somebody forgets to remove it at the call site (e.g., these other projects) it will get gobbled as an argument, instead of getting a nice error message that the function got an unexpected `lowering_platform` argument. Do you agree?

yashk2810: Ohh.. good point! Let's keep it then! Maybe add a small comment that this is supposed to be a kwarg?

jax/_src/dispatch.py
gnecula: Adding `experimental` everywhere seems to only increase code size (`experimental_platform_lowering=experimental_platform_lowering` is the most common use). I would prefer that we do it only at the API surface.

Adding a lot of comments for these one-line uses is also not very useful, imo. The key usage is at the API surface, when I do not think we should add comments, or where it actually gets used: `lowering_platform or backend.platform`. I could add comments in those 3 places, but at that point it is very obvious what is going on. wdyt?

yashk2810: Ok, sounds good. Adding a comment in pxla.py near the call to mlir.<something> should be good enough I think.

The dispatch.py codepath is dead code and will be removed in 2 weeks anyways.

yashk2810 (approved):

jax/_src/interpreters/pxla.py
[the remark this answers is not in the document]
gnecula: Leaving in this CL per offline discussion

jax/_src/pjit.py
gnecula: I change the order at the function definition for clarity.

jax/_src/pjit.py
gnecula: Fixed the pjit_partial_eval, thanks!

jax/_src/dispatch.py
gnecula: Done

pschuh (approved):

PR #14728, opened 2023-03-01
[custom_vjp] bwd function should not be WrappedFun, may run multiple times

mattjj:
The `custom_jvp_p` primitive is kind of like the `xla_call_p` primitive underlying final-style (old) `jit`. With `xla_call_p`, we have one Python callable (wrapped in an `lu.WrappedFun`) which is called exactly once. With `custom_jvp_p` we have two functions (each wrapped in an `lu.WrappedFun`), representing either the un-differentiated primal function or its JVP rule, and exactly one is called exactly once (determined not at runtime but rather by whether an AD pass hits at tracing time).

The called-exactly-once property is useful for reasoning about values plumbed through mutable cells. Indeed one of the main reasons `lu.WrappedFun` exists is to handle populating `Store`s for such plumbing.

While for `xla_call_p` and `custom_jvp_p` we have one and two functions, respectively, wrapped as `lu.WrappedFun`s corresponding to how they're called exactly once, the `custom_vjp_p` primitive is parameterized by _three_ callables. The three callables represent the primal function, the forward pass under autodiff, and the backward pass under autodiff. And each is wrapped as an `lu.WrappedFun`.

At least, they were until this PR. But that was a bug: after differentiation, the backward pass rule need not be called exactly once. Unlike the primal and forward pass functions, that's just not a property it has. So it never should have been an `lu.WrappedFun`! Tellingly, we never _needed_ it to be a `WrappedFun` either, in that we never used any `Store`s with it.

Even though the backward-pass rule could be run more than once, for some reason we never noticed until now. Perhaps that's because we only indirectly check whether `WrappedFun`s are run more than once, by checking whether any `Store`s are populated more than once. But `Store`s only became associated with the backward pass rule when `vmap` got involved. So I guess we just hadn't hit this combination until now!

The fix is simple: just don't make the backward pass rule a `WrappedFun`. Adapt callees not to expect it.

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -560,7 +560,7 @@ class custom_vjp(Generic[ReturnValue]):
       flat_fun, out_type = _flatten_fun_nokwargs(f_, in_tree)
       flat_fwd, out_trees = _flatten_fwd(fwd, primal_name, fwd_name, in_tree,
                                          out_type)
-      flat_bwd = _flatten_bwd(bwd, in_tree, in_avals, out_trees)
+      flat_bwd = _flatten_bwd(bwd, in_tree, in_avals, out_trees).call_wrapped
       out_flat = custom_vjp_call_p.bind(flat_fun, flat_fwd, flat_bwd,
                                         *args_flat, out_trees=out_trees)
       _, (out_tree, _) = lu.merge_linear_aux(out_type, out_trees)
@@ -680,7 +680,7 @@ class CustomVJPCallPrimitive(core.CallPrimitive):
     fwd, env_trace_todo2 = process_env_traces_fwd(
       fwd, top_trace and top_trace.level, out_trees)
     tracers = map(top_trace.full_raise, args)  # type: ignore
-    bwd_ = lu.wrap_init(lambda *args: bwd.call_wrapped(*args))
+    bwd_ = lambda *args: bwd(*args)
     outs = top_trace.process_custom_vjp_call(self, fun, fwd, bwd_, tracers,
                                              out_trees=out_trees)
     fst, env_trace_todo = lu.merge_linear_aux(env_trace_todo1, env_trace_todo2)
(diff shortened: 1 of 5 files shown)


jax/_src/custom_derivatives.py
```diff
@@ -680,7 +680,7 @@ def bind(self, fun, fwd, bwd, *args, out_trees):
     fwd, env_trace_todo2 = process_env_traces_fwd(
       fwd, top_trace and top_trace.level, out_trees)
     tracers = map(top_trace.full_raise, args)  # type: ignore
-    bwd_ = lu.wrap_init(lambda *args: bwd.call_wrapped(*args))
+    bwd_ = lambda *args: bwd(*args)
```
sharadmv, line 683: don't delete this lol

jax/_src/custom_derivatives.py
```diff
@@ -560,7 +560,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> ReturnValue:  # pytype: disable
       flat_fun, out_type = _flatten_fun_nokwargs(f_, in_tree)
       flat_fwd, out_trees = _flatten_fwd(fwd, primal_name, fwd_name, in_tree,
                                          out_type)
-      flat_bwd = _flatten_bwd(bwd, in_tree, in_avals, out_trees)
+      flat_bwd = _flatten_bwd(bwd, in_tree, in_avals, out_trees).call_wrapped
```
sharadmv, line 563: Why bother wrapping it at all if you're going to unwrap it here? Do you want to avoid rewriting _flatten_bwd to not use lu?

sharadmv (approved):

froystig (approved):

jax/_src/custom_derivatives.py
mattjj: 😅

tests/api_test.py
```diff
@@ -8385,6 +8385,17 @@ def f_bwd(_, g):
 
     jax.grad(f)(A([1.]))  # doesn't crash
 
+  def test_vmap_vjp_called_twice(self):
```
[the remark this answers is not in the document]
mattjj: Good call. I usually do that. I may have triggered submit without this change going in though. I'll try to add in a follow-up.

PR #14734, opened 2023-03-01
[shape_poly, call_tf] Add support for call_tf in a shape polymorphic program

gnecula:
The use case for call_tf with shape polymorphism is when we have a JAX program that calls into TF function, and we want to serialize the JAX program with some shapes unknown. Today this use case does not work, except in the special case when the output shape of the called TF function returns statically known shapes.

The idea is that we allow the user of call_tf to specify the output shape. This can be done even in presence of shape polymorphism, by writing the output shape as an expression in terms of the input shapes. This is what other JAX primitives do, e.g., concat, so we are simply enabling call_tf to get the same behavior.

This change should be enough for old-style jax2tf, but will require more work for native serialization.

We also removed some old code that was trying to workaround some limitations in shape inference in TF. I think that those workarounds are ugly, and I am prepared to give error messages rather than keep that code. So far no tests fail.

Still to do:

 * ~~more testing~~
 * ~~check that the TF function returns values of the declared shapes~~
 * ~~documentation~~
 * ~~CHANGELOG~~

### DIFF
--- a/jax/experimental/jax2tf/README.md
+++ b/jax/experimental/jax2tf/README.md
@@ -1229,6 +1229,57 @@ DeviceArray data or for np.ndarray that are aligned on 16-byte
 boundaries) and on GPU (for DeviceArray).
 The zero-copy does not yet work on TPU.
 
+`call_tf` works even with shape polymorphism, but in that case
+the user must pass the `output_shape_dtype` parameter to `call_tf` to declare
+the expected output shapes. This allows JAX tracing to know the shape and
+dtype of the results so that it can continue tracing the rest of the program.
+When `output_shape_dtype` is not given (the default case), `call_tf` will
+form a `tf.Graph` for the called TF function and will use the inferred
+type and shape. However, in presence of dynamic shape the inferred TF
+type will contain `None` for the dynamic dimensions, which is not enough
+information for JAX shape polymorphism.
+
+For example:
+
+```python
+def fun_jax(x):
+  y_shape = (x.shape[0] * 2, y.shape[1:])
+  y = jax2tf.call_tf(
+      lambda x: tf.concat([x, x], axis=0),
+      output_shape_dype=jax.ShapeDtypeStruct(y_shape, x.dtype))(x)
+  # JAX will know the y.shape
+  return jnp.ones(y.shape, dtype=y.dtype) + y
+
+jax2tf.convert(fun_jax, polymorphic_shapes=["b, ..."])(x)
+```
+
+An even simpler example for a function that returns the same shape as the input:
+
+```python
+def fun_jax(x):
+  return jax2tf.call_tf(tf.math.sin,
+                        output_shape_dtype=x)
+  )(x)
+
+jax2tf.convert(fun_jax, polymorphic_shapes=["b, ..."])(x)
+```
+
+If all the output shapes of the TF function are static, JAX does not need the
+`output_shape_dtype` argument:
+
+```python
+def fun_tf(x):
+  return tf.math.reduce_sum(tf.math.sin(x))
+
+def fun_jax(x):
+  return jax2tf.call_tf(fun_tf)(x)
+
+# The following will not throw an error because the output shape of fun_tf is static.
+jax2tf.convert(fun_jax, polymorphic_shapes=["b, ..."])(x)
+```
+
+The shape polymorphism support for `call_tf` does not yet work for native lowering.
+
 ### Limitations of call_tf
 
 The TF function must be compileable (`tf.function(func, jit_compile=True)`)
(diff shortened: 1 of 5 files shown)


gnecula:
@rxsang @maxwillzq

jax/experimental/jax2tf/call_tf.py
```diff
@@ -113,67 +123,58 @@ def canonical_arg(v):
...
+              zip(res_tf_flat,
+                  (output_avals if output_avals is not None
+                   else (None,) * len(res_tf_flat))))]
+      return checked_res_tf_flat
 
     # Prepare a tf.function ahead of time, to cache the concrete functions. This
     # won't be used in op-by-op execution mode.
     function_flat_tf = tf.function(callable_flat_tf, autograph=False, jit_compile=True)
 
-    input_shapes_tf = [s.shape for s in args_flat_sig_tf]
```
gnecula, line 150: I think that this check does not need to be here, because it is subsumed by the `check_tf_result` that happens a bit downstream.

jax/experimental/jax2tf/call_tf.py
```diff
@@ -248,6 +249,44 @@ def fix_float0(arg_jax, ct_arg_jax):
...
+  except Exception as e:
+    msg = ("The called TF function returns a result that is not "
+           f"convertible to JAX: {r_tf}.")
+    raise ValueError(msg) from e
+
+  if r_aval is None:
+    return r_tf
+  # We convert to TF type, and canonicalize to 32-bit if necessary
+  r_aval_dtype_tf = jax2tf_internal._to_tf_dtype(r_aval.dtype)
+  # Checking shapes is trickier in presence of dynamic shapes. I wish we could
```
gnecula, line 266: @rxsang PTAL

jax/experimental/jax2tf/call_tf.py
```diff
@@ -309,39 +348,43 @@ class CallTfEffect(effects.Effect):
...
-    return (
-        tuple([
-            # We convert to JAX type, and canonicalize to 32-bit if necessary
-            core.ShapedArray(shape, jax2tf_internal._to_jax_dtype(dtype))
-            for dtype, shape in zip(concrete_function_flat_tf.output_dtypes,
-                                    concrete_function_flat_tf.output_shapes)
-        ]),
-        effects)
-
-  # There are some cases when TF shape inference is not powerful enough to
```
gnecula, line 335: I think that we do not need this workaround anymore. We should not need to go all the way to the TFXLA bridge to get the output shapes. We should be able to get them from tf.Graph, and if not, we'll ask the user to pass `output_shape_dtype`.

@rxsang PTAL

jax/experimental/jax2tf/call_tf.py
```diff
@@ -55,13 +55,15 @@
...
+TfVal = jax2tf_internal.TfVal
 
 # The platforms for which to use DLPack to avoid copying (only works on GPU
 # and CPU at the moment, and only for DeviceArray). For CPU we don't need
 # DLPack, if we are careful.
 _DLPACK_PLATFORMS = ("gpu",)
 
-def call_tf(callable_tf: Callable, has_side_effects=True) -> Callable:
+def call_tf(callable_tf: Callable, has_side_effects=True,
+            output_shape_dtype=None) -> Callable:
```
maxwillzq, line 66: can we specify the return type of output_shape_dtype here ? e.g 
output_shape_dtype: Optional[Union[jax.ShapeDtypeStruct, jax.Array]]=None.

It helps catch error if users give illegal data.

jax/experimental/jax2tf/README.md
```diff
@@ -1229,6 +1229,58 @@ DeviceArray data or for np.ndarray that are aligned on 16-byte
...
+information for JAX shape polymorphism.
+
+For example:
+
+```python
+def fun_jax(x):
+  y_shape = (x.shape[0] * 2, y.shape[1:])
+  y = jax2tf.call_tf(
+      lambda x: tf.concat([x, x], axis=0),
+      output_shape_dype=jax.ShapeDtypeStruct(y_shape, x.dtype))(x)
```
maxwillzq, line 1249: nit:  The better idea about the usage is that we write a tf_func output_infer_shape function since output_shape_and_dtype should depend on inputs.

Example
```python
def fun_jax(x):
  output_shape_dtype_infer = lambda x: (x.shape[0] * 2, x.shape[1:])
  y = jax2tf.call_tf(
      lambda x: tf.concat([x, x], axis=0),
      output_shape_dype=output_shape_dtype_infer(x))(x)
```

maxwillzq (approved):

jax/experimental/jax2tf/call_tf.py
```diff
@@ -90,6 +92,14 @@ def call_tf(callable_tf: Callable, has_side_effects=True) -> Callable:
     has_side_effects: if True then it ensures that instances of this primitive
       are not removed or replicated by JAX optimizations such as dead-code
       elimination.
+    output_shape_dtype: An optional declaration of the expected shapes and dtypes
+      from the called TensorFlow function. If given it will be used during JAX
+      tracing to form the abstract values of the results of the `call_tf`. If
+      not given then we form a `tf.Graph` for the called TensorFlow function and
+      we use the TensorFlow-inferred shapes and types. Must be a pytree matching the
+      structure of the returned pytree from the TensorFlow function, containing
```
rxsang, line 100: nit: pytree -> nested structure, since TF doesn't have the pytree concept.

rxsang (approved):

jax/experimental/jax2tf/README.md
```diff
@@ -1229,6 +1229,58 @@ DeviceArray data or for np.ndarray that are aligned on 16-byte
...
+  # JAX will know the y.shape
+  return jnp.ones(y.shape, dtype=y.dtype) + y
+  )(x)
```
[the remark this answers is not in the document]
gnecula: Good catch!

jax/experimental/jax2tf/README.md
gnecula: Actually `output_shape_dtype` cannot depend on the inputs, only on their shape. The shape of the inputs is always available to the caller and I do not see much benefit from adding that are then immediately invoked, once. If the shape computation is complex, the called can define it in a function which they call when computing the `output_shape_dtype`. This pattern has been used elsewhere in JAX, e.g., for `host_callback` and for `jax.debug` and has been sufficient for a couple of years.

jax/experimental/jax2tf/call_tf.py
gnecula: Done

jax/experimental/jax2tf/call_tf.py
gnecula: Unfortunately there is no good type annotation here to put under `Optional`, because if the function returns more than one result then it should have a matching structure. We have this problem elsewhere in JAX, and we have decided to not use typing annotations because those would cause more trouble than help in some cases. The only clearly useful part of the typing annotation is `Optional`, but I think it is not worth writing `Optional[Any]` when the default value is `None`.

PR #14754, opened 2023-03-02
DOC: clarify variable names in scan doc

jakevdp:


jax/_src/lax/control_flow/loops.py
```diff
@@ -126,8 +126,9 @@ def scan(f: Callable[[Carry, X], Tuple[Carry, Y]],
   represents the type with the same pytree structure and corresponding leaves
   each with an additional leading axis.
 
-  When ``a`` is an array type or None, and ``b`` is an array type, the semantics
-  of :func:`~scan` are given roughly by this Python implementation::
+  When ``xs`` (denoted `a` above) is an array type or None, and ``ys`` (denoted `b`
+  above) is an array type, the semantics of :func:`~scan` are given roughly by this
+  Python implementation::
```
mattjj, line 131:
```suggestion
  When the type of ``xs`` (denoted `a` above) is an array type or None, and the type of ``ys`` (denoted `b`
  above) is an array type, the semantics of :func:`~scan` are given roughly by this
  Python implementation::
```

jax/_src/lax/control_flow/loops.py
```diff
@@ -139,7 +140,7 @@ def scan(f, init, xs, length=None):
         ys.append(y)
       return carry, np.stack(ys)
 
-  Unlike that Python version, both ``a`` and ``b`` may be arbitrary pytree
+  Unlike that Python version, both ``xs`` and ``ys`` may be arbitrary pytree
   types, and so multiple arrays can be scanned over at once and produce multiple
```
mattjj, line 144: To be consistent, replace "may be arbitrary pytree types" with "may be arbitrary pytree values"? (That is, `a` and `b` stand for types while `xs` and `ys` stand for values.)

mattjj (approved):
Minor inconsistencies to clean up noted below but otherwise LGTM

jax/_src/lax/control_flow/loops.py
jakevdp: Done, thanks!

PR #14760, opened 2023-03-02
in `closure_convert`, hoist all tracers involved in staging

froystig:
with @mattjj

We need to hoist all `DynamicJaxprTracer`s in `jax.closure_convert`, not only those that are float-dtyped. Leaving any `DynamicJaxprTracer` in the closure of a `custom_{jvp,vjp}`'ed function might result in a tracer leak, as staging to jaxpr could already be complete by the time the closure is consumed.

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -21,7 +21,6 @@ from typing import (
 from jax._src import core
 from jax._src import custom_api_util
 from jax._src.custom_transpose import custom_transpose
-from jax._src import dtypes
 from jax._src import effects
 from jax._src import linear_util as lu
 from jax._src import traceback_util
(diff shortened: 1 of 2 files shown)


mattjj (approved):

froystig:
We can't quite merge this because it breaks this test:

https://github.com/google/jax/blob/15caafd93726f2f991053699c0d0ab530aa88b05/tests/api_test.py#L7354-L7362

Skimming discussion #9951 to which it links, the approach we'd intend going forward is to use `custom_jvp` (or `custom_vjp`) with the `symbolic_zeros` option. My guess is that we should try doing this ourselves in `odeint`, and then rewrite the test at a higher level, to test derivatives of `odeint` with integer arguments.

PR #14764, opened 2023-03-03
attach debug info to jaxpr, pass to mlir/mhlo

mattjj:
roll-forward of #14702 but now
1. `jax.make_jaxpr(fn)(*args, **kwargs)` no longer requires a weakref to `fn`, and
2. we only try to attach arg info if the number of parameters in the computation is equal to the number of names from the debug info (one downstream test failed this, not sure why but we can just let this extra metadata propagation fail in some cases)

### DIFF
--- a/jax/_src/ad_checkpoint.py
+++ b/jax/_src/ad_checkpoint.py
@@ -401,9 +401,15 @@ def saved_residuals(f, *args, **kwargs) -> List[Tuple[core.AbstractValue, str]]:
       results.append((v.aval, 'from a constant'))
 
   assert len(jaxpr.invars) == len(in_leaves)
+  dbg = pe.debug_info(f, in_tree, True, "saved_residuals")
+  arg_info = pe.arg_info_all(dbg)
   for i, v in enumerate(jaxpr.invars):
     if v in res_vars:
-      src = f'from {pe.arg_info_pytree(f, in_tree, True, [i])}'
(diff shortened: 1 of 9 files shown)


yashk2810:
I was going to fix this before but since it got rolled back, you should also add this info to lower_mesh_computation just like you are doing for lower_sharding_computation.

mattjj:
Can you open an issue, so someone can do it in a follow-up PR? :)

yashk2810:
Hmm.. don't worry about it. I'll do it. I don't understand the follow up part but it's fine.

mattjj:
I'm working on something else so I can't switch back to working on this right now, other than shepherding it to be merged.

Also, I think we should wait for it to stick before we add any more to it. It might get rolled back again, as there were two reasons for the previous one to roll back.

yashk2810:
Well you shouldn't have merged it then lol since the change is incomplete :). If nvidia uses xmap anywhere in their code, this won't work. But anyways, I already have a CL to add it to lower_mesh_computation so all good.

> It might get rolled back again, as there were two reasons for the previous one to roll back.
You didn't test if this actually fixed those targets?

yashk2810:
On that note, I am going to merge both those paths, so this doesn't happen again.

yashk2810:
cl/513696476 should fix the xmap case.

Issue #14751, opened 2023-03-02
Remove unnecessary broadcast from `jnp.vectorize`


Gattocrucco:
Summary: improvement to `jnp.vectorize` that reduces memory consumption and increases speed by a lot (100x in the example).

Currently `jnp.vectorize` partially broadcasts together the arguments before passing them to `jax.vmap`. As a comment says in the code:

```
# TODO(shoyer): Consider squeezing out size 1 dimensions instead, and
# doing all vectorization with vmap? This *might* be a little more
# efficient but would require more careful book-keeping.
```

The *might* is an understatement. The performance can drop a great deal with explicit broadcasting if the arrays end up being very large and fill up the memory. I modified the implementation to fix this:

```python
from jax._src.numpy.vectorize import (
  functools,
  _apply_excluded,
  map,
  jnp,
  _parse_gufunc_signature,
  _parse_input_dimensions,
  _check_output_dims,
  zip,
  api,
)

def modified_vectorize(pyfunc, *, excluded=frozenset(), signature=None):
  """Define a vectorized function with broadcasting.

  :func:`vectorize` is a convenience wrapper for defining vectorized
  functions with broadcasting, in the style of NumPy's
  `generalized universal functions <https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html>`_.
  It allows for defining functions that are automatically repeated across
  any leading dimensions, without the implementation of the function needing to
  be concerned about how to handle higher dimensional inputs.

  :func:`jax.numpy.vectorize` has the same interface as
  :class:`numpy.vectorize`, but it is syntactic sugar for an auto-batching
  transformation (:func:`vmap`) rather than a Python loop. This should be
  considerably more efficient, but the implementation must be written in terms
  of functions that act on JAX arrays.

  Args:
    pyfunc: function to vectorize.
    excluded: optional set of integers representing positional arguments for
      which the function will not be vectorized. These will be passed directly
      to ``pyfunc`` unmodified.
    signature: optional generalized universal function signature, e.g.,
      ``(m,n),(n)->(m)`` for vectorized matrix-vector multiplication. If
      provided, ``pyfunc`` will be called with (and expected to return) arrays
      with shapes given by the size of corresponding core dimensions. By
      default, pyfunc is assumed to take scalars arrays as input and output.

  Returns:
    Vectorized version of the given function.

  Here are a few examples of how one could write vectorized linear algebra
  routines using :func:`vectorize`:

  >>> from functools import partial

[...]

PR #14782, opened 2023-03-04
avoid unnecessary broadcasting in jax.numpy.vectorize

Gattocrucco:
Simple but very effective optimization of `jnp.vectorize`, which was marked `TODO`. Issue #14751.

### DIFF
--- a/jax/_src/numpy/vectorize.py
+++ b/jax/_src/numpy/vectorize.py
@@ -278,36 +278,41 @@ def vectorize(pyfunc, *, excluded=frozenset(), signature=None):
         excluded_func, dim_sizes, output_core_dims, error_context)
 
     # Rather than broadcasting all arguments to full broadcast shapes, prefer
-    # expanding dimensions using vmap when possible. By pushing broadcasting
+    # expanding dimensions using vmap. By pushing broadcasting
     # into vmap, we can make use of more efficient batching rules for
     # primitives where only some arguments are batched (e.g., for
-    # lax_linalg.triangular_solve).
+    # lax_linalg.triangular_solve), and avoid instantiating large broadcasted
+    # arrays.
 
-    vec_args = []
-    vmap_counts = []
+    squeezed_args = []
+    rev_filled_shapes = []
 
     for arg, core_dims in zip(args, input_core_dims):
-      # Explicitly broadcast the dimensions already found on each argument,
-      # because these dimensiosns might be of size 1, which vmap doesn't
-      # handle.
-      # TODO(shoyer): Consider squeezing out size 1 dimensions instead, and
-      # doing all vectorization with vmap? This *might* be a little more
-      # efficient but would require more careful book-keeping.
-      core_shape = tuple(dim_sizes[dim] for dim in core_dims)
-      full_shape = broadcast_shape + core_shape
-      vec_shape = full_shape[-arg.ndim:] if arg.ndim else ()
+      noncore_shape = arg.shape[:arg.ndim - len(core_dims)]
 
-      vec_arg = jnp.broadcast_to(arg, vec_shape)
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/vectorize.py
```diff
@@ -278,36 +278,43 @@ def wrapped(*args):
...
-      vec_args.append(vec_arg)
-
-      vmap_count = len(vec_shape) - len(core_shape)
-      vmap_counts.append(vmap_count)
+      noncore_shape = arg.shape[:arg.ndim - len(core_dims)]
+
+      pad_ndim = len(broadcast_shape) - len(noncore_shape)
+      filled_shape = pad_ndim * (1,) + noncore_shape
+      assert len(filled_shape) == len(broadcast_shape)
+      assert jnp.broadcast_shapes(broadcast_shape, filled_shape) == broadcast_shape
```
jakevdp, line 296: If these assertions are intended to catch a user error, we should instead raise an informative error in place of a raw assertion. If these checks are unrelated to user error, we could put them behind `if config.jax_enable_checks`, then the assertions would only be executed during testing, not at normal runtime.

jax/_src/numpy/vectorize.py
```diff
@@ -278,36 +278,43 @@ def wrapped(*args):
...
+      else:
+        vectorized_func = api.vmap(vectorized_func, in_axes)
+    result = vectorized_func(*squeezed_args)
+
+    if not dims_to_expand:
+      return result
+    elif output_core_dims is not None and len(output_core_dims) > 1:
+      return tuple(jnp.expand_dims(r, axis=dims_to_expand) for r in result)
+    else:
+      return jnp.expand_dims(result, axis=dims_to_expand)
```
jakevdp, line 318: Does this correctly handle the case where the function returns multiple values? e.g.
```python
@jnp.vectorize
def f(x):
  return x, x + 1
result = f(jnp.arange(4))
```

jax/_src/numpy/vectorize.py
Gattocrucco: They are internal checks, I add `if config.jax_enable_checks`.

jax/_src/numpy/vectorize.py
Gattocrucco: Yes, it works. This is tested in the unit tests. Multiple outputs are handled by the `elif output_core_dims ...` branch.

jakevdp: My reading of the code is that if `signature` is not specified, `output_core_dims` will be `None` and so that branch will be skipped. What am I missing?

tests/lax_numpy_vectorize_test.py
```diff
@@ -216,6 +217,13 @@ def test_inconsistent_output_size(self):
         ValueError, r"inconsistent size for core dimension 'n'"):
       f(jnp.zeros((2, 3)), jnp.zeros((3, 4)))
 
+  def test_expand_dims_multiple_outputs_no_signature(self):
+    f = jnp.vectorize(lambda x: (x, x))
+    x = jnp.arange(1)
+    xx = f(x)
+    self.assertAllClose(xx[0], x)
```
jakevdp, line 224: Can we also do `self.assertIsInstance(xx, tuple)`?

Gattocrucco: ok

jakevdp (changes requested):
Looks good, thanks!

The last thing we need is to squash all changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests), and then we can get this change merged. Thanks!

jakevdp (approved):

PR #14789, opened 2023-03-05
(Static type-checking fix)  Add `_ScalarMeta(dtype=...)` field.

patrick-kidger:


jax/_src/numpy/lax_numpy.py
```diff
@@ -162,6 +162,8 @@ def _dtype(x: Any) -> DType:
 # types return JAX arrays when instantiated.
 
 class _ScalarMeta(type):
+  dtype: Any
```
jakevdp, line 165: Can we make this `dtype: np.dtype`?

patrick-kidger: Ah, I missed that this was an output, not an input, to `np.dtype`. Will do.

jakevdp (commented):
Looks good! If you can squash the changes then we can get this merged.

patrick-kidger:
Done.

jakevdp (approved):

PR #14791, opened 2023-03-05
(Static type checking) Add type annotations to `curry`

patrick-kidger:


patrick-kidger:
Hmm, I'm not sure why mypy can't figure out the types for a few transformations. (For context I added this so that `jax.core.jaxpr_as_fun` was correctly typed.)

I'll just `# type: ignore` on usages of `jax.core.jaxpr_as_fun` and close this.

Issue #12141, opened 2022-08-29
jax.scipy.stats.logistic don't implement location and scale


langosco:
### Description

The docstring of `jax.scipy.stats.norm.cdf` says it takes keyword arguments `loc` and `scale`, but running
```
jax.scipy.stats.logistic.cdf(1, loc=1)
```
raises a `TypeError: cdf() got an unexpected keyword argument 'loc'`. 

### What jax/jaxlib version are you using?

jax v0.3.16, jaxlib v0.3.15

### Which accelerator(s) are you using?

CPU

### Additional System Info

Python 3.8.10, Ubuntu

jakevdp:
Thanks for the report – the reason the parameter is listed but not implemented is because the docsttring is copied from `scipy.stats` (you'll notice *"Original docstring below"* in the [doc page](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.stats.logistic.cdf.html))

We'll leave this open as a feature request - if it's something you want to contribute, feel free to open a pull request!

rajasekharporeddy:
Hi @langosco, @jakevdp 

Since the PR raised by @b0nce was merged, the issue has been resolved. 

The above mentioned code works without any error.

```
>>> jax.scipy.stats.logistic.cdf(1, loc=1)
Array(0.5, dtype=float32, weak_type=True)
```

Thank you

PR #14792, opened 2023-03-05
Fix `loc` and `scale` parameters in `scipy.logistic.*`. Add CDF and SF for several distributions.

b0nce:
Starting with [this issue](https://github.com/google/jax/issues/12141) related to missing `loc` and `scale` parameters in logistic distribution (`jax.scipy.stats.logistic`), I've discovered that a lot of rv_continuous distributions lack `.cdf` and `.sf` methods.

Note that we can't guarantee exact match of results in Cauchy for `.isf` and `.ppf`, since `.tan` functions behave different in lax and numpy near divergence points (such as pi/2).

Complete list of changes:
 - Beta distribution: `.cdf`, `.logcdf` and `.sf` methods.
 - Cauchy distribution: `.cdf`, `.logcdf`, `.sf`, `.isf` and `.ppf` methods.
 - Chi squared distribution: `.cdf`, `.logcdf` and `.sf` methods.
 - Gamma distribution: `.cdf`, `.logcdf` and `.sf` methods.
 - Logistic distribution:  `loc` and `scale` parameters have been added to `logpdf`, `pdf`, `cdf`, `logcdf`, `isf`, `sf` and `ppf` methods.
 - Normal distribution: `.sf` and `.isf` methods.
 - Tests were skipped for `.cdf` methods due to similarities with `.logcdf` implementations.

### DIFF
--- a/docs/jax.scipy.rst
+++ b/docs/jax.scipy.rst
@@ -172,6 +172,9 @@ jax.scipy.stats.beta
 
    logpdf
    pdf
+   cdf
+   logcdf
+   sf
 
 jax.scipy.stats.betabinom
 ~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -192,6 +195,11 @@ jax.scipy.stats.cauchy
 
    logpdf
    pdf
+   cdf
+   logcdf
+   sf
+   isf
+   ppf
 
 jax.scipy.stats.chi2
 ~~~~~~~~~~~~~~~~~~~~
@@ -202,7 +210,9 @@ jax.scipy.stats.chi2
 
    logpdf
    pdf
-
+   cdf
+   logcdf
+   sf
 
 
 jax.scipy.stats.dirichlet
@@ -232,6 +242,9 @@ jax.scipy.stats.gamma
 
    logpdf
    pdf
+   cdf
+   logcdf
+   sf
 
 jax.scipy.stats.gennorm
 ~~~~~~~~~~~~~~~~~~~~~~~
@@ -296,6 +309,8 @@ jax.scipy.stats.norm
    logpdf
    pdf
    ppf
+   sf
+   isf
 
 jax.scipy.stats.pareto
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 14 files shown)


docs/jax.scipy.rst
```diff
@@ -190,8 +193,13 @@ jax.scipy.stats.cauchy
 .. autosummary::
   :toctree: _autosummary
 
-   logpdf
+   logpdf as logpdf
```
jakevdp, line 196: Typo?

jax/_src/scipy/stats/logistic.py
```diff
@@ -24,29 +24,38 @@
...
   half_x = lax.div(x, two)
-  return lax.mul(lax.neg(two), jnp.logaddexp(half_x, lax.neg(half_x)))
+  return lax.sub(lax.mul(lax.neg(two), jnp.logaddexp(half_x, lax.neg(half_x))), lax.log(scale))
 
 
 @_wraps(osp_stats.logistic.pdf, update_doc=False)
-def pdf(x: ArrayLike) -> Array:
-  return lax.exp(logpdf(x))
+def pdf(x: ArrayLike, loc: ArrayLike = 0, scale: ArrayLike = 1) -> Array:
+  return lax.div(lax.exp(logpdf(x, loc, scale)), scale)
```
jakevdp, line 37: Why is this not just `lax.exp(logpdf)`?

jakevdp (changes requested):
Thanks for the contribution - looks good!

Can we add tests for non-default values of `logistic` parameters to ensure they're implemented correctly? (I have concerns about `logpdf` below)

docs/jax.scipy.rst
b0nce: Yep, fixing

jax/_src/scipy/stats/logistic.py
b0nce: Because it's also a typo :)
Fixing

jakevdp (changes requested):
Looks great, thanks!

The last thing we need is to squash all the changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) and then we can get this merged. Thanks!

b0nce:
Oh, I thought this only applies to the first group of commits, so I squashed them into one and fixes into one 😅

jakevdp (approved):

PR #14113, opened 2023-01-22
Adding info to CG and BICGSTAB

botev:

PR #14798, opened 2023-03-06
Fix `custom_linear_solve` batching rule in case of auxiliary arguments

### LINKED ISSUE #14249: Incorrect batching rule with custom_linear_solve when has_aux=True?
### Description

Since [this commit](https://github.com/google/jax/commit/73ed511d3945246467b0e3e12da326d40a3fe153), a bunch of tests that used to pass in JAXopt are now failing with `ValueError: too many values to unpack`. The issue seems to come from the fact that `custom_linear_solve` has an incorrect batching rule when `has_aux=True`. Here's a minimal repro:
```python
import numpy as np
import jax.numpy as jnp
import jax
from jax.scipy.sparse.linalg import cg

M = rng.randn(5, 5)
A = np.dot(M, M.T)
matvec = lambda x: (jnp.dot(A, x[0]), jnp.dot(A, x[1]))

def f(b):
  return cg(matvec, (b, b))[0]

b = rng.randn(5)
jax.jacrev(f)(b)
```
Error obtained:
```
  File "/Users/mblondel/Desktop/playground/cg_bug.py", line 16, in <module>
    jax.jacrev(f)(b)
  File "/Users/mblondel/Desktop/playground/cg_bug.py", line 13, in f
    return cg(matvec, (b, b))[0]
  File "/Users/mblondel/Desktop/projects/jax/jax/_src/scipy/sparse/linalg.py", line 290, in cg
    return _isolve(_cg_solve,
  File "/Users/mblondel/Desktop/projects/jax/jax/_src/scipy/sparse/linalg.py", line 229, in _isolve
    x_maybe_info = lax.custom_linear_solve(
jax._src.source_info_util.JaxStackTraceBeforeTransformation: ValueError: too many values to unpack (expected 1)

[...]

  File "/Users/mblondel/Desktop/projects/jax/jax/interpreters/batching.py", line 710, in _match_axes_jaxpr
    out_axis_dest, = out_axes_dest
jax._src.traceback_util.UnfilteredStackTrace: ValueError: too many values to unpack (expected 1)

--------------------

  File "/Users/mblondel/Desktop/playground/cg_bug.py", line 16, in <module>
    jax.jacrev(f)(b)
  File "/Users/mblondel/Desktop/projects/jax/jax/_src/api.py", line 1324, in jacfun
    jac = vmap(pullback)(_std_basis(y))
ValueError: too many values to unpack (expected 1)
```

If I come back to before the aforementioned commit, this script works, since `cg` didn't use `custom_linear_solve(..., has_aux=True)`.

Also note that I use `(b, b)` to define a pytree. When the input is just an array, there is no error.

main branch

CPU

### Additional system info

3.8, Mac

### NVIDIA GPU info

_No response_

nicholasjng:
Previously, batching in-/out axes of the wrong lengths were passed into the batched jaxpr builders for the `matvec` and `solve_t` jaxprs. This commit is a best-effort fix from debugging the axes designations in the batched jaxpr constructions of these functions.

Fixes #14249.

I also tested this change with cherry-picked #14113, and it passes. Mainly I reworked the batch axis designations by matching their input and output lengths of both the linear operators and the solves with the expected number of input/output avals of the various jaxprs.
The batching jaxpr construction was pretty tough to navigate for me, as someone who does not look at the internals too often - maybe some docstrings or small annotations would help? Happy to hear your thoughts.

cc @froystig

### DIFF
--- a/jax/_src/lax/control_flow/solves.py
+++ b/jax/_src/lax/control_flow/solves.py
@@ -387,7 +387,9 @@ def _linear_solve_batching_rule(spmd_axis_name, axis_size, axis_name, main_type,
   (matvec, vecmat, solve, solve_t) = jaxprs
   (matvec_bat, vecmat_bat, solve_bat, solve_t_bat) = params_bat
 
-  num_aux = len(solve.out_avals) - len(matvec.out_avals)
+  # number of operator out avals is assumed to be the same for matvec/vecmat
+  num_operator_out_avals = len(matvec.out_avals)
+  num_aux = len(solve.out_avals) - num_operator_out_avals
   # Fixpoint computation of which parts of x and b are batched; we need to
   # ensure this is consistent between all four jaxprs
   b_bat = orig_b_bat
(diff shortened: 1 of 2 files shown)


froystig (changes requested):
Thanks for digging into this!

Would it be possible to come up with a test that fails before this change but passes afterward?

nicholasjng:
Added a regression test in 27b2651 (it's basically just an explicit version of the breakage observed in #14249).

froystig (approved):
Thank you!

PR #14813, opened 2023-03-07
add result info to mhlo

### LINKED ISSUE #14780: Adding debug information for the return tensor of StableHLO function
PR [14702](https://github.com/google/jax/pull/14702) adds debug info to jaxpr, which will show for each stateHLO's input tensor the corresponding parameter name and pytree key (like in the example), which is quite useful. 

 
unnamed

It would be nice to include the similar info for each return tensor, i.e, the tuple index and pytree key path from the jit-function for each return tensor.  In some cases, the JAX function may return large number of tensors, knowing which tensor is correlated with return position is useful.

### LINKED ISSUE #14787: `jax.make_jaxpr` incorrectly typed
### Description

The type annotation for the return value should be a 2-tuple if `return_shape=True`. Right now it is incorrectly fixed to just a `ClosedJaxpr`.

https://github.com/google/jax/blob/2ccd785e167ab424f0c0d79de7b549a7f94aea61/jax/_src/api.py#L2912

0.4.4

mattjj:
fixes #14780, and incidentally fixes #14787

### DIFF
--- a/jax/_src/interpreters/pxla.py
+++ b/jax/_src/interpreters/pxla.py
@@ -3063,10 +3063,7 @@ def lower_sharding_computation(
   unordered_effects = list(
       effects.ordered_effects.filter_not_in(closed_jaxpr.effects))
   ordered_effects = list(effects.ordered_effects.filter_in(closed_jaxpr.effects))
-  arg_info = jaxpr.debug_info and pe.arg_info_all(jaxpr.debug_info)
-  arg_names = None if arg_info is None else [
-      f'{name}{keystr(path)}' for i, (name, path) in enumerate(arg_info)
-      if i in kept_var_idx]
+  arg_names, result_names = _debug_names(jaxpr.debug_info, kept_var_idx)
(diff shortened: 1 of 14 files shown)


jax/_src/interpreters/pxla.py
```diff
@@ -3067,6 +3067,8 @@ def lower_sharding_computation(
   arg_names = None if arg_info is None else [
       f'{name}{keystr(path)}' for i, (name, path) in enumerate(arg_info)
       if i in kept_var_idx]
+  result_info = jaxpr.debug_info and pe.result_info(jaxpr.debug_info)
```
yashk2810, line 3070: Can you add this below too to lower_mesh_computation?

mattjj: Done!

yashk2810: Thank you!

yashk2810 (approved):

PR #14836, opened 2023-03-07
Add "compare_all" method to searchsorted

shoyer:
This method can be considerably faster than "scan" and "sort" (at least on TPUs) in cases where the array being searched is very small. In particular, this allowed us to remove a bottleneck in one of our training jobs that was doing a very large number of searches where the sorted array was only of size 24.

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -4817,13 +4817,20 @@ def _searchsorted_via_sort(sorted_arr: Array, query: Array, side: str, dtype: ty
   return lax.reshape(lax.sub(index, _rank(query_flat)), np.shape(query)).astype(dtype)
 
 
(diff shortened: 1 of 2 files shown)


jakevdp (approved):
Looks great!

jakevdp:
It looks like your implementation relies on implicit dtype promotion – we should make any intended type promotion explicit so that `jax_numpy_dtype_promotion='strict'` will work correctly with this method.

shoyer:
I don't think I'm relying on implicit dtype promotion -- the arrays all get cast to a common dtype before calling my new helper function.

The test failures from internal CI all looks like some pre-existing issue to me. They are all in `tests/scipy_stats_test.py::LaxBackedScipyStatsTests`.

jakevdp:
Sorry, yeah I think you're right.

PR #14849, opened 2023-03-08
Fix implicit rank promotion.

jakeh-gc:
Fixes an implicit rank promotion in default LU decomposition implementation.

Related to #7208

### DIFF
--- a/jax/_src/lax/linalg.py
+++ b/jax/_src/lax/linalg.py
@@ -1042,7 +1042,7 @@ def _lu_unblocked(a):
     a = a.at[:, k].set(jnp.where(m_idx > k, a[:, k] / x, a[:, k]))
 
     # a[k+1:, k+1:] -= jnp.outer(a[k+1:, k], a[k, k+1:])
-    a = a - jnp.where((m_idx[:, None] > k) & (n_idx > k),
+    a = a - jnp.where((m_idx[:, None] > k) & (n_idx[None, :] > k),
                      jnp.outer(a[:, k], a[k, :]), jnp.array(0, dtype=a.dtype))
     return pivot, perm, a
 


jax/_src/lax/linalg.py
```diff
@@ -1042,7 +1042,7 @@ def body(k, state):
     a = a.at[:, k].set(jnp.where(m_idx > k, a[:, k] / x, a[:, k]))
 
     # a[k+1:, k+1:] -= jnp.outer(a[k+1:, k], a[k, k+1:])
-    a = a - jnp.where((m_idx[:, None] > k) & (n_idx > k),
+    a = a - jnp.where((m_idx[:, None] > k) & (n_idx[:, None] > k),
```
jakevdp, line 1045: I think this changes the semantics of the operation. I think it should be `n_idx[None, :]`

jakevdp (changes requested):
Thanks for looking at this!

I think the fix needs to be slightly different to not change the semantics. Perhaps we need an additional test to cover this code path?

jakevdp:
Regarding testing, it looks like the issue is that we have specialized impl rules for every platform, so this codepath never gets called in the tests: https://github.com/google/jax/blob/44082be10376805c2f80cda0dfbb917f1e916853/jax/_src/lax/linalg.py#L1202-L1217

Out of curiosity, how did you hit this error?

jax/_src/lax/linalg.py
jakeh-gc: Yes, you're correct.
That's what I get for not rerunning the tests before putting up the PR. 😅

jakeh-gc:
> Regarding testing, it looks like the issue is that we have specialized impl rules for every platform we test, so this default lowering never gets called in the tests:

Yes, this code is unreachable with what's in the repository today.

> Out of curiosity, how did you hit this error?

I'm writing a new backend, so the specialisation for cpu, gpu, and tpu isn't taken.
This happens in a few places across the codebase. I plan to put up PRs as I come across them to minimise the diff between my work and upstream.

To test this change you can comment out `jax/jax/_src/lax/linalg.py#L1206-L1208` and run the CPU tests, but I don't think that's a very robust solution for you guys.

jakevdp:
I think we'd prefer to remove this codepath, since it's essentially dead code. Would that be problematic for your project?

jakeh-gc:
> Would that be problematic for your project?

It's not fatal, as I would possibly adopt the code or something similar. The down side Is I won't be able to provide my backend as a standalone jaxlib (might not be possible anyway), it'll require python changes in the jax library.

I would like some reference implementation, but that's doesn't particularly help upstream.
I'm not sure what the expectations of the "interpreter" backend are. That will hit the default, but I don't believe it is tested.

jakevdp:
OK, thanks. Why don't we leave it in for now; I tested this fix manually by removing the CPU translation rule, and it seems correct.

Could you squash your changes into a single commit?

jakevdp (approved):

jakeh-gc:
Just like any other. Has a very low priority, so I doubt it is ever used when the CPU backend exists.
```
>>> import jax
>>> import jax.numpy as jnp
>>> jax.devices("interpreter")
[interpreter(id=0)]
>>> def foo(x):
...   return x @ x
... 
>>> f = jax.jit(foo, backend="interpreter")
>>> x = jnp.arange(16)
>>> f(x)
Array(1240, dtype=int32)
```

jakeh-gc:
We could add "interpreter" to the `ALL_BACKENDS` list in `/localdata/jakeh/jax-popit/jax/jaxlib/jax.bzl` and run as part of CI. I just wonder (and worry) how many tests will fail. 😅

jakeh-gc:
In the more narrow context of this PR, I'll write a test for this case in the morning (late in the U.K.) based on the "interpreter" backend.

jakevdp:
Well, in any case I'll merge this fix because it corrects a bug.

PR #14863, opened 2023-03-09
[shard_map] make debug_print work with shard_map (eager and jit)

mattjj:
thanks to much advice from @sharadmv

### DIFF
--- a/jax/_src/debugging.py
+++ b/jax/_src/debugging.py
@@ -121,14 +121,24 @@ ad.primitive_transposes[debug_callback_p] = debug_callback_transpose_rule
 
 def debug_callback_lowering(ctx, *args, effect, callback, **params):
 
-  if isinstance(ctx.module_context.axis_context,
-                (mlir.SPMDAxisContext, mlir.ShardingContext)):
-    # Apply maximal sharding so pjit only executes the callback on device 0.
+  axis_context = ctx.module_context.axis_context
+  if (isinstance(axis_context, mlir.SPMDAxisContext) and
+        set(axis_context.manual_axes) == set(axis_context.mesh.axis_names)):
+    # If we have fully manual sharding during lowering, that means the JAX
+    # program has per-device semantics, so we run the callback on each device.
+    sharding = xc.OpSharding()
+    sharding.type = xc.OpSharding.Type.MANUAL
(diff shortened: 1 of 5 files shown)


jax/_src/debugging.py
```diff
@@ -121,14 +121,24 @@ def debug_callback_transpose_rule(*flat_args, callback: Callable[..., Any],
...
+        set(axis_context.manual_axes) == set(axis_context.mesh.axis_names)):
+    # If we have fully manual sharding during lowering, that means the JAX
+    # program has per-device semantics, so we run the callback on each device.
+    sharding = xc.OpSharding()
+    sharding.type = xc.OpSharding.Type.MANUAL
+  elif isinstance(axis_context, (mlir.ShardingContext, mlir.SPMDAxisContext)):
+    # If we have fully automatic sharding during lowering, that means the JAX
+    # program has bulk array semantics, so we run the callback with a MAXIMAL
+    # sharding and hence execute it only once on the full logical value).
+    # If we have partially automatic sharding, we do this too... not sure why!
```
mattjj, line 135: In my first cut at this logic, it was more like: if fully manual then MANUAL, if fully auto then MAXIMAL, if not those cases but still spmd-ish ERROR, else None. But some tests started failing with that! So I went back to the old behavior: if partially manual / partially automatic we also do MAXIMAL.

jax/experimental/shard_map.py
```diff
@@ -564,11 +565,16 @@ def sublift(self, tracer):
...
-    with core.eval_context(), jax.disable_jit(False):
-      out_vals = jax.jit(f)(*in_vals)
-    rule = _rep_rules.get(prim, partial(_rep_rule, prim))
-    out_rep = rule(self.mesh, *in_rep, **params) if self.check else set()
+    eager_rule = eager_rules.get(prim)
+    if eager_rule:
+      out_vals = eager_rule(self.mesh, *in_vals, **params)
+    else:
+      f = HashablePartial(_prim_applier, prim, tuple(params.items()), self.mesh)
+      if 'debug' in str(prim): breakpoint()
```
sharadmv, line 573: Can remove this

sharadmv (approved):

jax/experimental/shard_map.py
mattjj: Oops :blob-blush:

PR #14878, opened 2023-03-09
Take seed rather than Key for some PRNG algorithms

sr5434:
For some random number generation algorithms(randint, shuffle, choice, permutation, normal, multivariate_normal, truncated_normal, bernoulli), instead of requiring a PRNGKey, the user can pass a seed into the algorithm. This means that the user doesn't have to create a random key when using these algorithms.

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -389,7 +389,7 @@ def _randint(key, shape, minval, maxval, dtype) -> Array:
 
 
 def shuffle(seed: Int, x: ArrayLike, axis: int = 0) -> Array:
-  """Shuffle the elements of an array uniformly at random along an axis.
+  """Shuffle the elements of an array uniformly at random along an axis. This is not a stateful algorithm.
 
   Args:
     seed: the seed for the random number generation.
@@ -411,7 +411,7 @@ def permutation(seed: Int,
                 x: Union[int, ArrayLike],
                 axis: int = 0,
                 independent: bool = False) -> Array:
-  """Returns a randomly permuted array or range.
+  """Returns a randomly permuted array or range. This is not a stateful algorithm.
 
   Args:
     seed: the seed for the random number generation.
@@ -473,7 +473,7 @@ def choice(seed: Int,
            replace: bool = True,
            p: Optional[RealArray] = None,
            axis: int = 0) -> Array:
-  """Generates a random sample from a given array.
+  """Generates a random sample from a given array. This is not a stateful algorithm.
 
   .. warning::
     If ``p`` has fewer non-zero elements than the requested number of samples,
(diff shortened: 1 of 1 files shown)


hawkinsp:
I suspect we are not going to accept this PR because the use of a `key` is quite intentional here, but I'll let @froystig who is the PRNG expert comment.

Note that these are not stateful PRNGs, so "seed" would be misleading at best.

sr5434:
But we are passing the seed into the Key, right? Wouldn't it be easier to use if we passed the seed for the key generator into the PRNG, and took care of generating the key for the user?

mattjj:
Thanks for the idea, but I agree with what Peter wrote. Some additional points:
1. Passing around ints means the program expresses strictly less information than if it were to pass around PRNGKeys instead. Concretely that would defeat several of our ideas for PRNGKey reuse checking (which would check reuse of PRNGKeys but not ints, so the more PRNGKey-typed values in the program the safer we can be).
2. It simultaneously makes the code less abstract and less flexible, since the PRNGKey details aren't abstracted away (and so we can't switch PRNG key generation strategies).
3. If someone wants this convenience layer today, even at the above costs, they can pretty easily just wrap all the functions. But we tend to shy away from baking tradeoff-inducing convenience wrappers into JAX itself.

sr5434:
Couldn’t we have the current functions and wrapper functions so that users can choose which function they can use?

mattjj:
Certainly we could, but we don't bake those convenience wrappers into JAX itself.

sr5434:
Oh ok.

sr5434:
> Thanks for the idea, but I agree with what Peter wrote. Some additional points:
> 
> 1. Passing around ints means the program expresses strictly less information than if it were to pass around PRNGKeys instead. Concretely that would defeat several of our ideas for PRNGKey reuse checking (which would check reuse of PRNGKeys but not ints, so the more PRNGKey-typed values in the program the safer we can be).
> 2. It simultaneously makes the code less abstract and less flexible, since the PRNGKey details aren't abstracted away (and so we can't switch PRNG key generation strategies).
> 3. If someone wants this convenience layer today, even at the above costs, they can pretty easily just wrap all the functions. But we tend to shy away from baking tradeoff-inducing convenience wrappers into JAX itself.

Can’t we trust the user to not reuse keys? Also, couldn’t we add an argument to the function call allowing them to switch key generation strategies?

froystig:
We don't do reuse checks by default, so you could say we "trust the user" all the time, but we offer a means of opting into a checker (under development). It's helpful in debugging.

I'm going to close. Discussion can always continue of course. This is also a major breaking change to the public API, and it adds complexity in making the functions in `jax.random` inconsistent with one another, so we'd need a strong case for it.

sr5434:
If we have wrappers alongside the original functions for all PRNG algorithms, would it be possible to merge?

soraros:
> - If someone wants this convenience layer today, even at the above costs, they can pretty easily just wrap all the functions. But we tend to shy away from baking tradeoff-inducing convenience wrappers into JAX itself.

@sr5434, based on @mattjj's bullet point above, I'd say highly unlikely.
Personally, I think JAX's random module is very well-designed. Maybe this [tutorial](https://jax.readthedocs.io/en/latest/jax-101/05-random-numbers.html) can convince you `KeyArray` is a better idea than NumPy-style random seed.

BTW, I suppose this discussion is better continued in [Github Discussions](https://github.com/google/jax/discussions).

PR #14887, opened 2023-03-09
Add xla garbage collection to gc.callback

### LINKED ISSUE #14882: device_put leaks references on numpy inputs
First reported in #14876; here's a minimal repro that reproduces on CPU and GPU:
```python
import sys
import jax
import numpy as np

x = np.arange(1000, dtype='float32')
print(sys.getrefcount(x))
# 2

y = jax.device_put(x)
print(sys.getrefcount(x))
# 3

del y
print(sys.getrefcount(x))
# 3
```

jakevdp:
Fixes #14882

I confirmed that the test fails without the new callback. I'm a bit worried that this could be flaky – I'm not sure what to do aside from the `time.sleep`: maybe a retry loop?

### DIFF
--- a/tests/api_test.py
+++ b/tests/api_test.py
@@ -9629,5 +9629,18 @@ class AutodidaxTest(jtu.JaxTestCase):
     autodidax_module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(autodidax_module)
 
+class GarbageCollectionTest(jtu.JaxTestCase):
+  def test_xla_gc_callback(self):
+    # https://github.com/google/jax/issues/14882
+    x_np = np.arange(10, dtype='int32')
+    x_jax = jax.device_put(x_np)
+    x_np_weakref = weakref.ref(x_np)
+
+    del x_np
+    del x_jax
+    gc.collect()
+
+    assert x_np_weakref() is None
+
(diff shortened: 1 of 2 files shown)


tests/api_test.py
```diff
@@ -9629,5 +9630,19 @@ def test_autodidax_smoketest(self):
...
+class GarbageCollectionTest(jtu.JaxTestCase):
+  def test_xla_gc_callback(self):
+    # https://github.com/google/jax/issues/14882
+    x_np = np.arange(10, dtype='int32')
+    x_jax = jax.device_put(x_np)
+    x_np_weakref = weakref.ref(x_np)
+
+    del x_np
+    del x_jax
+    gc.collect()
```
hawkinsp, line 9642: Is gc.collect actually asynchronous?

If it is, you could use a gc callback to wait for its completion.

jakevdp: OK, I'll try dropping the sleep

jakevdp: Seems like it's working – maybe I assumed incorrectly about the asynchronousity

hawkinsp (approved):
Let's try this. If the test proves flaky, we can remove the test.

PR #14891, opened 2023-03-10
Add overloads to `make_jaxpr`. Closes #14787.

### LINKED ISSUE #14787: `jax.make_jaxpr` incorrectly typed

patrick-kidger:
Pyright is very not happy with the new state of affairs.


jax/_src/api.py
```diff
@@ -2904,6 +2904,37 @@ def transposed_fun(const, out_cotangent):
...
+               abstracted_axes: Optional[Any] = None,
+               ) -> Callable[..., core.ClosedJaxpr]:
+  ...
+
+
+@overload
+def make_jaxpr(fun: Callable,
+               static_argnums: Union[int, Iterable[int]] = (),
+               axis_env: Optional[Sequence[Tuple[AxisName, int]]] = None,
+               return_shape: Literal[True] = True,
```
jakevdp, line 2921: I don't think this is valid to have different default arguments for `return_shape` here and above.

mattjj:
Thanks for providing this!

But is this really the _only_ way to make pyright happy? When I added the annotation a few days ago I considered doing this overload thing, but then decided it was so much bloat code that actually makes it _less_ human-readable on net. 

If there's really no other way to make pyright happy, I might want to say just add a `# type: ignore` or assertion in the calling code. I don't think it's our policy to add extra code to make _all_ type checkers happy.

jakevdp:
I think the default arguments need to be set up differently to be valid. You can see how we've done it elsewhere in the package: https://github.com/google/jax/blob/50408fd694a63e73af349ba22a8fb0c0f8cbc574/jax/_src/numpy/lax_numpy.py#L2305-L2324

patrick-kidger:
Okay, if the preference is to work around this in user code then I'll add some `typing.cast`s to my code to handle this.

(For the record, Jake was right -- looks like mypy and pyright have different expectations about how to handle overloads, and I was used to the other convention.)

hawkinsp:
I can't view the diff here, but I suspect another solution we could use soon (perhaps not quite yet) is to stick these kinds of detailed types in a `.pyi` file, where they will benefit type checkers but mostly not hinder humans.

mattjj:
Sorry, I meant to imply that my current opinion is weakly held and I could be talked out of it by anyone! (Though I also hate `typing.cast` for the same reasons as stated above.)

The pyi file seems like a great approach, for the reason Peter cited.

PR #14908, opened 2023-03-10
Fix jnp.sort & jnp.vdot in no-jit mode

### LINKED ISSUE #14906: `jnp.sort` and `jnp.vdot` raise error under `jax.disable_jit()`
### Description

`jnp.sort`, `jnp.vdot` and a few others in `jax.numpy` will raise under `jax.disable_jit()`. IMO, the fact that exactly *these* functions go wrong is not important, nor is the fact that `jax.disable_jit()` is involved.

Taking a look at their implementation, on can spot a pattern:

```python
@partial(jit, static_argnums=(...))
def f(a: ArrayLike):
  util._check_arraylike('f', a)
  bar = foo(a.ravel())  # << not good
  ...
```

Without jit, the scalar case will get caught by the tests, lead us to the fix naturally. These functions do array conversion implicitly on the jit boundary hides the fact that array methods might not be available.

Fixing them is easy enough, but I want to ask/discuss something more general here:
1. Maybe we should insert these explicit conversions after `_check_arraylike` everywhere, in the form of `asarray` or explicit `reshape` or `ravel`.
2. If the pattern in 1. is desirable, is it always safe to use the simpler and faster variant `util._asarray`? I think if we can, we should.
3. I see similar problems (that JAX maintainers have rather canonical solutions in JAX itself) popping up almost every time someone tries to contribute to `jax.scipy` etc. Can we have a contributor guide documenting these issues? (`array` conversion, `dtype` handling, type hinting, `jnp` vs `lax`, standard ways to deal with static arguments: `concrete_or_error`, etc.)
4. What is the best place to ask these broader questions that are also half bug reports? Should the Github discussions be preferred?

@jakevdp I bore you the most in many related issues before, so I'm gonna tag you here.

jakevdp:
Fixes #14906

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3057,7 +3057,7 @@ def vdot(a, b, *, precision=None):
   util._check_arraylike("vdot", a, b)
   if issubdtype(_dtype(a), complexfloating):
     a = ufuncs.conj(a)
-  return dot(a.ravel(), b.ravel(), precision=precision)
+  return dot(ravel(a), ravel(b), precision=precision)
 
 
 @util._wraps(np.tensordot, lax_description=_PRECISION_DOC)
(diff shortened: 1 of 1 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -3487,7 +3487,7 @@ def sort(a, axis: Optional[int] = -1, kind='quicksort', order=None):
     raise ValueError("'order' argument to sort is not supported.")
 
   if axis is None:
-    return lax.sort(a.ravel(), dimension=0)
+    return lax.sort(ravel(a), dimension=0)
   else:
     return lax.sort(a, dimension=_canonicalize_axis(axis, ndim(a)))
```
soraros, line 3492: Assuming `a: ArrayLike`, `lax.sort(a, ...)` is also a type mismatch.

hawkinsp (approved):

jax/_src/numpy/lax_numpy.py
jakevdp: Done, thanks!

PR #14912, opened 2023-03-10
jnp.argmin/max: correctly handle out argument

### LINKED ISSUE #14904: `argmax` and `argmin` doesn't raise when the argument `out` is used.
### Description

https://github.com/google/jax/blob/3fbe4e8230c60e40a1c54732d879438b9d39bd16/jax/_src/numpy/lax_numpy.py#L3401-L3410

`out` is never passed to `_argmax` where the error handling is actually implemented. I guess wrapped functions like `_argmax` shouldn't have default arguments.

Also, `keepdims` should be key word only.

jakevdp:
Fixes #14904

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -3399,14 +3399,15 @@ def argwhere(a, *, size=None, fill_value=None):
 
 
 @util._wraps(np.argmax, skip_params=['out'])
-def argmax(a, axis: Optional[int] = None, out=None, keepdims=None):
-  return _argmax(a, None if axis is None else operator.index(axis), keepdims=bool(keepdims))
-
-@partial(jit, static_argnames=('axis', 'keepdims'), inline=True)
-def _argmax(a, axis: Optional[int] = None, out=None, keepdims=False):
+def argmax(a: ArrayLike, axis: Optional[int] = None, out=None, keepdims=None) -> Array:
   util._check_arraylike("argmax", a)
   if out is not None:
     raise NotImplementedError("The 'out' argument to jnp.argmax is not supported.")
+  return _argmax(asarray(a), None if axis is None else operator.index(axis),
+                 keepdims=bool(keepdims))
+
+@partial(jit, static_argnames=('axis', 'keepdims'), inline=True)
+def _argmax(a: Array, axis: Optional[int] = None, keepdims: bool = False) -> Array:
   if axis is None:
     dims = list(range(ndim(a)))
     a = ravel(a)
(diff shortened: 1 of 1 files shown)


hawkinsp (approved):

jax/_src/numpy/lax_numpy.py
```diff
@@ -3399,14 +3399,15 @@ def argwhere(a, *, size=None, fill_value=None):
 
 
 @util._wraps(np.argmax, skip_params=['out'])
-def argmax(a, axis: Optional[int] = None, out=None, keepdims=None):
-  return _argmax(a, None if axis is None else operator.index(axis), keepdims=bool(keepdims))
-
-@partial(jit, static_argnames=('axis', 'keepdims'), inline=True)
-def _argmax(a, axis: Optional[int] = None, out=None, keepdims=False):
+def argmax(a: ArrayLike, axis: Optional[int] = None, out=None, keepdims=None) -> Array:
```
soraros, line 3402: `keepdimes` should have been keyword only.

jax/_src/numpy/lax_numpy.py
```diff
@@ -3399,14 +3399,15 @@ def argwhere(a, *, size=None, fill_value=None):
...
   util._check_arraylike("argmax", a)
   if out is not None:
     raise NotImplementedError("The 'out' argument to jnp.argmax is not supported.")
+  return _argmax(asarray(a), None if axis is None else operator.index(axis),
+                 keepdims=bool(keepdims))
+
+@partial(jit, static_argnames=('axis', 'keepdims'), inline=True)
+def _argmax(a: Array, axis: Optional[int] = None, keepdims: bool = False) -> Array:
```
soraros, line 3410: Maybe we could drop the default arguments from the inner functions, WDYT?

jakevdp: Better would be to drop the inner functions entirely... they're mostly left-over from a time when the functions returned by `jit` didn't support keyword arguments.

soraros: They do have a benefit, for `partial` doesn't work well with static type-checkers.

jakevdp: Yeah, I'm of two minds on that. It's nice to write code that's compatible with static type checkers, but if we're forced to write code in a confusing/inefficient way to appease them, I'd lean towards not jumping through those hoops.

jakevdp: We may end up just making a `pyi` file for `jax.numpy` due to issues like this.

soraros: > We may end up just making a pyi file for jax.numpy due to issues like this.

That's actually great news. And all the overloads will have a place to go.

PR #14951, opened 2023-03-13
Correct multidimensional Rosenbrock function

packquickly:
Using `np.diff` will give a function with the same minimum as the Rosenbrock function but which may be easier to optimise. In particular, the Rosenbrock function should have the term $(x_{i + 1} - x_i^2)^2$ but using `np.diff` will give $(x_{i + 1} - x_i)^2$. Both have minima at $x = \mathbf{1}$.

### DIFF
--- a/tests/scipy_optimize_test.py
+++ b/tests/scipy_optimize_test.py
@@ -28,7 +28,7 @@ config.parse_flags_with_absl()
 
 def rosenbrock(np):
   def func(x):
-    return np.sum(100. * np.diff(x) ** 2 + (1. - x[:-1]) ** 2)
+    return np.sum(100. * (x[1:] - x[:-1]**2) ** 2 + (1. - x[:-1]) ** 2)
 
   return func
 


RaderJason:
I've gone ahead and filled out the CLA as an individual.

jakevdp (approved):
Looks good, thanks for catching this!

Let's make sure the tests still pass before we merge.

jakevdp:
It looks like the optimizer chokes on this in the complex test, which is basically minimizing over the 10-dimensional coupled rosenbrock function.

Perhaps we should use the un-coupled version instead? i.e. something like this:
```python
return np.sum(100. * (x[1:2] - x[::2]**2) ** 2 + (1. - x[::2]) ** 2)
```

RaderJason:
So, in the coupled case only (L)BFGS fails the test withe the coupled optimiser. Locally I went ahead and tried the uncoupled one:
```
return np.sum(100. * (x[::2]**2 - x[1::2]) ** 2 + (x[:-1] - 1) ** 2)   
```
and (L)BFGS still fails. It seems the problem lies with BFGS here. Unfortunately, I don't have time to track down the issue what the issue is with BFGS and why it's failing this test.

jakevdp:
I don't think the expression in your last comment is correct for the uncoupled function. You should use `x[::2]` instead of `x[:-1]`

RaderJason:
Ah! Indeed, thanks for catching that. Sadly, this leads to no change in the test outcome.

PR #14954, opened 2023-03-13
use Partial to make ravel_pytree unflatteners jit-friendly

mattjj:
Based on #13834

Before this change, the `unravel` functions produced by `ravel_pytree` were not `jit`-friendly, in that they could not be returned from `jit`-decorated functions or passed in as arguments without requiring recompilation, because they were opaque Python callables. But they don't need to be opaque! We can make them transparent using `util.HashablePartial`. (This PR moves `HashablePartial` into util.py, resolving a TODO.)

This change brings `ravel_pytree`-produced callables into line with other parts of the system, e.g. the callables returned by `jax.linearize` and `jax.vjp`, which are more flexible in how they can be used with `jit`.

### DIFF
--- a/jax/_src/flatten_util.py
+++ b/jax/_src/flatten_util.py
@@ -17,7 +17,7 @@ import warnings
 import numpy as np
 
 from jax._src.tree_util import tree_flatten, tree_unflatten
-from jax._src.util import safe_zip, unzip2
+from jax._src.util import safe_zip, unzip2, HashablePartial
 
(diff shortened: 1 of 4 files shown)


sharadmv (approved):

mattjj:
Relatedly I noticed that things like `jax.vjp` are using `Partial-of-partial` where they could be using `Partial-of-HashablePartial` to get more cache hits...

PR #15069, opened 2023-03-17
explicitly convert jnp.var scalar normalizer to float (from int)

### LINKED ISSUE #15068: jnp.mean and jnp.std crash on large arrays.
### Description

I have a large array of shape float32[16,32,1280,4096].  When I call jnp.std on this array, jit crashes and fails with the following error:  "Python int 2684354560 too large to convert to int32".

Clearly the implementation of jnp.std is dividing by the number of elements in the array, which in this case does not fit into a 32-bit integer.  

Target platform: pjit on TPU. 

v0.4.2

TPU

### Additional system info

_No response_

### NVIDIA GPU info

_No response_

mattjj:
This way we don't pass a potentially-large (Python builtin) int value to an int32 JAX computation parameter and get an error.

Fixes #15068

### DIFF
--- a/jax/_src/numpy/reductions.py
+++ b/jax/_src/numpy/reductions.py
@@ -433,8 +433,9 @@ def _var(a: ArrayLike, axis: Axis = None, dtype: DTypeLike = None,
   a = lax_internal.asarray(a).astype(computation_dtype)
   a_mean = mean(a, axis, dtype=computation_dtype, keepdims=True, where=where)
   centered = lax.sub(a, a_mean)
-  if dtypes.issubdtype(centered.dtype, np.complexfloating):
+  if dtypes.issubdtype(computation_dtype, np.complexfloating):
     centered = lax.real(lax.mul(centered, lax.conj(centered)))
+    computation_dtype = centered.dtype  # avoid casting to complex below.
   else:
     centered = lax.square(centered)
 
@@ -443,13 +444,13 @@ def _var(a: ArrayLike, axis: Axis = None, dtype: DTypeLike = None,
       normalizer = core.dimension_as_value(np.size(a))
     else:
       normalizer = core.dimension_as_value(_axis_size(a, axis))
+    normalizer = lax.convert_element_type(normalizer, computation_dtype)
   else:
-    normalizer = sum(_broadcast_to(where, np.shape(a)), axis, dtype=dtype, keepdims=keepdims)
-  normalizer = normalizer - ddof
-
-  result = sum(centered, axis, keepdims=keepdims, where=where)
-  result = lax.div(result, lax.convert_element_type(normalizer, result.dtype))
-  return lax.convert_element_type(result, dtype)
+    normalizer = sum(_broadcast_to(where, np.shape(a)), axis,
+                     dtype=computation_dtype, keepdims=keepdims)
+  normalizer = lax.sub(normalizer, lax.convert_element_type(ddof, computation_dtype))
+  result = sum(centered, axis, dtype=computation_dtype, keepdims=keepdims, where=where)
+  return lax.div(result, normalizer).astype(dtype)
 
 
 def _var_promote_types(a_dtype: DTypeLike, dtype: DTypeLike) -> Tuple[DType, DType]:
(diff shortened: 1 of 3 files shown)


jax/_src/numpy/reductions.py
```diff
@@ -443,11 +443,12 @@ def _var(a: ArrayLike, axis: Axis = None, dtype: DTypeLike = None,
...
     if axis is None:
-      normalizer = core.dimension_as_value(np.size(a))
+      normalizer = float(core.dimension_as_value(np.size(a)))
     else:
-      normalizer = core.dimension_as_value(_axis_size(a, axis))
+      normalizer = float(core.dimension_as_value(_axis_size(a, axis)))
   else:
-    normalizer = sum(_broadcast_to(where, np.shape(a)), axis, dtype=dtype, keepdims=keepdims)
+    normalizer = sum(_broadcast_to(where, np.shape(a)), axis, dtype=dtype,  # type: ignore
+                     keepdims=keepdims)
```
jakevdp, line 451: If you passed e.g. `where=True` in the new test case, it would fail. Should we worry about this? Maybe cast `where` to float before summing?

mattjj: What kind of failure do you have in mind? I don't think a large Python int can be created on this line.

jakevdp: `sum` here is `jnp.sum`, so it will return an `int32` which could wrap in the situations that this PR is concerned with. e.g.:
```python
In [2]: where = True
In [3]: jnp.sum(jnp.broadcast_to(where, (16, 32, 1280, 4096)))
Out[3]: Array(-1610612736, dtype=int32)
```

mattjj: Yes I agree it could wrap (though it depends on what `dtype` might be), I just wanted to clarify because it's a different issue than the one the PR is fixing (i.e. getting an overflow error when converting a large Python int to a JAX int32, leading to an exception).

I can add a fix for this overflow issue in this PR also if you'd like. It is actually worse than the original issue this PR is fixing, in that it'd be a silently incorrect answer, rather than an exception.

mattjj: Can `dtype` (the argument to `sum`) ever be an `int32` here? (I haven't tried but looking briefly at the promotion logic in `_var_promote_types` it seems like it might not be possible.)

jakevdp: Yeah, agreed the mechanism is different but I was pattern matching on "large arrays result in incorrect behavior". May as well fix both issues here?

mattjj: Yeah you're right, so long as we're looking at this and thinking about it. I'll try to hit the issue here but I'm not convinced it's possible given `dtype`; lmk if you have a reproducer (calling `jnp.var`).

jakevdp: Yeah, it looks like this isn't an issue for `var` because `dtype` is always set to a float value. But it does come up for `mean`:
```python

In [2]: jnp.mean(jnp.full((16, 32, 1280, 4096), 0.1, 'float32'), where=True)
Out[2]: Array(-0.16677855, dtype=float32)
```

jakevdp (approved):

jax/_src/numpy/reductions.py
mattjj: SGTM!

jax/_src/numpy/reductions.py
```diff
@@ -443,11 +443,12 @@ def _var(a: ArrayLike, axis: Axis = None, dtype: DTypeLike = None,
 
   if where is None:
     if axis is None:
-      normalizer = core.dimension_as_value(np.size(a))
+      normalizer = float(core.dimension_as_value(np.size(a)))
```
jakevdp, line 446: It looks like this is causing failures in the dynamic shapes tests because `float` requires a static input. A better option may be to cast this (dynamically) to the type of the output.

jakevdp: Resolved

jakevdp:
OK @mattjj – I pushed an update to your branch. I don't think I can request you as a reviewer because you originally opened the PR, but PTAL 😀

mattjj:
Thanks @jakevdp, you're the best.

PR #15079, opened 2023-03-18
separate register_pytree_node and register_pytree_with_keys tests

mattjj:
In #14710 we added tests for `register_pytree_with_keys` etc, but because those replaced some `register_pytree_node` tests (e.g. replacing `register_pytree_node(AnObject, ...)` with `register_pytree_with_keys(AnObject, ...)`), it meant we had less test coverage of the pre-existing `register_pytree_node` path.

The aim of this PR is to restore the `register_pytree_node` tests that existed before #14710, but also to keep all the new test coverage as well. The basic approach is just to:
1. put back `register_pytree_node(AnObject, ...)` as it was before #14710;
2. introduce the `AnObject2` trivial subclass and to apply `register_pytree_with_keys(AnObject2, ...)` to it;
3. include `AnObject2` in both pathful and non-pathful pytree tests.

### DIFF
--- a/tests/tree_util_test.py
+++ b/tests/tree_util_test.py
@@ -57,10 +57,15 @@ class AnObject:
   def __repr__(self):
     return f"AnObject({self.x},{self.y},{self.z})"
 
+tree_util.register_pytree_node(AnObject, lambda o: ((o.x, o.y), o.z),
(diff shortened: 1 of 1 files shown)


IvyZX (approved):
Thank you for adding this!

PR #15085, opened 2023-03-19
make mlir arg and result names work with pmap

mattjj:
~**This PR includes the commit from #15080, i.e. that's its diffbase, so ignore this one until that one goes in!**~

This is a follow-up on #15080 to restore (and indeed fix!) how pmap builds a jaxpr with debug info (i.e. parameter names and result paths). The difference with the machinery in #15080 is just to deal with pmap being final-style (i.e.     build the jaxpr at the last second, well after pytrees have been flattened away and transformations have been applied), whereas pjit is initial style (i.e. build the jaxpr as soon as possible, so all the relevant debug information is still present). As you might imagine, plumbing for the former is a bit more long-range and subtle.

The main idea here is that we need to annotate and maintain debug info on the `lu.WrappedFun` instance, which we first form at the api.py level, then pass through transformations (which can either update or drop debug info), then finally   hand off to the impl rule to be traced to a jaxpr. It makes sense as an annotation, parallel with the `in_type` annotation on `lu.WrappedFun`s used for dynamic shapes, because the debug info has to be determined initially early on (at the api.py level) updated incrementally as transformations are applied, as they might e.g. add tangent inputs and outputs.

In more detail: with an initial-style higher-orer primitive (like pjit), a jaxpr is formed immediately. Transformations, like autodiff, are jaxpr-to-jaxpr, and so those transformations (like ad.jvp_jaxpr) need to return a new jaxpr       either with updated debug info or no debug info at all. (The initial implementation in #15080 doesn't provide updated debug info in any of those jaxpr-to-jaxpr transformation functions, so the debug info is only applied to the jaxpr and  then lowered to MLIR when the pjit as at the top level.)

For final-style, like pmap here, instead of transformations being jaxpr-to-jaxpr, they're WrappedFun-to-WrappedFun. And so, analogously, transformations, like ad.JVPTrace.process_map, would need to produce a WrappedFun with updated debug info or no debug info at all. (Also analogously to #15080, this PR only implements enough for the debug info to be preserved for top-level pmaps, and preserving debug info through transformations is left as follow-up work.)

This PR doesn't yet delete the trace-time debug info in partial_eval.py. But that'll happen next!

### DIFF
--- a/jax/_src/linear_util.py
+++ b/jax/_src/linear_util.py
@@ -64,7 +64,7 @@ data must be immutable, because it will be stored in function memoization tables
 from __future__ import annotations
 
 from functools import partial
-from typing import Any, Tuple, Callable
+from typing import Any, Tuple, Callable, Optional, NamedTuple
 import weakref
 
 from jax.tree_util import tree_map
@@ -124,14 +124,15 @@ class WrappedFun:
     params: extra parameters to pass as keyword arguments to `f`, along with the
       transformed keyword arguments.
   """
-  __slots__ = ("f", "transforms", "stores", "params", "in_type")
+  __slots__ = ("f", "transforms", "stores", "params", "in_type", "debug_info")
 
-  def __init__(self, f, transforms, stores, params, in_type):
+  def __init__(self, f, transforms, stores, params, in_type, debug_info):
     self.f = f
     self.transforms = transforms
     self.stores = stores
     self.params = params
     self.in_type = in_type
+    self.debug_info = debug_info
 
   @property
   def __name__(self):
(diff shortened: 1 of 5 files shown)


jax/_src/linear_util.py
```diff
@@ -271,6 +274,23 @@ def valid_size(d) -> bool:
   assert all(provided)
 
 
+class TracingDebugInfo(NamedTuple):
```
mattjj, line 277: This moved from api_util.py to here because we now treat it as a kind of annotation on lu.WrappedFun instances, hence we need it in this upstream file.

hawkinsp (approved):

PR #15086, opened 2023-03-20
Support extrapolation in jnp.interp

### LINKED ISSUE #14858: Support linear extrapolation in jnp.interp?
`numpy.interp` does not support extrapolation, instead leaving this functionality to the much more complex [scipy.interpolate.interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html), which uses a class based interface.

So I propose to interpret the string values `left='extrapolate'` and `right='extrapolate'` as indicating that extrapolation should be done instead of only using a constant value:

- This value would be consistent with the keyword argument used by the `fill_value` argument to scipy.interpolate.interp1d.
- There is no potential confusion about strings indicating interpolation modes versus values because `left` and `right` currently only support numeric arguments.

This would be useful and extremely easy to implement in JAX. Basically all that needs to be done to obtain linear extrapolation is to conditionally remove these calls to `where` at the end of `jnp.interp`
https://github.com/google/jax/blob/6634600c46c99e3f0861e9258aace3fdfe250baf/jax/_src/numpy/lax_numpy.py#L1042-L1043

Any thoughts? I can see a case for asking to do this in NumPy first, but I'm not sure there would be large enough benefits to the NumPy users given that they can use SciPy instead. Also it would be slightly more annoying to implement in NumPy since the implementation of `np.interp` is written in C.

shoyer:
Fixes https://github.com/google/jax/issues/14858

### DIFF
--- a/jax/_src/numpy/lax_numpy.py
+++ b/jax/_src/numpy/lax_numpy.py
@@ -940,11 +940,9 @@ def isclose(a: ArrayLike, b: ArrayLike, rtol: ArrayLike = 1e-05, atol: ArrayLike
     return lax.eq(a, b)
 
 
-@util._wraps(np.interp)
-@jit
-def interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
-           left: Optional[ArrayLike] = None,
-           right: Optional[ArrayLike] = None,
+def _interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
+           left: Union[ArrayLike, str, None] = None,
+           right: Union[ArrayLike, str, None] = None,
            period: Optional[ArrayLike] = None) -> Array:
   util.check_arraylike("interp", x, xp, fp)
   if shape(xp) != shape(fp) or ndim(xp) != 1:
@@ -952,6 +950,21 @@ def interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
   x_arr, xp_arr = util.promote_dtypes_inexact(x, xp)
   fp_arr, = util.promote_dtypes_inexact(fp)
   del x, xp, fp
+
+  if isinstance(left, str):
+    if left != 'extrapolate':
+      raise ValueError("the only valid string value of `left` is "
+                       f"'extrapolate', but got: {left!r}")
+    extrapolate_left = True
+  else:
+    extrapolate_left = False
+  if isinstance(right, str):
+    if right != 'extrapolate':
+      raise ValueError("the only valid string value of `right` is "
+                       f"'extrapolate', but got: {right!r}")
+    extrapolate_right = True
+  else:
+    extrapolate_right = False
 
   if dtypes.issubdtype(x_arr.dtype, np.complexfloating):
     raise ValueError("jnp.interp: complex x values not supported.")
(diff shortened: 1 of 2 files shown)


jax/_src/numpy/lax_numpy.py
```diff
@@ -1001,19 +1001,34 @@ def isclose(a: ArrayLike, b: ArrayLike, rtol: ArrayLike = 1e-05, atol: ArrayLike
...
+           left: Union[ArrayLike, str, None] = None,
+           right: Union[ArrayLike, str, None] = None,
            period: Optional[ArrayLike] = None) -> Array:
   util.check_arraylike("interp", x, xp, fp)
   if shape(xp) != shape(fp) or ndim(xp) != 1:
     raise ValueError("xp and fp must be one-dimensional arrays of equal size")
   x_arr, xp_arr = util.promote_dtypes_inexact(x, xp)
   fp_arr, = util.promote_dtypes_inexact(fp)
+  x_arr, xp_arr = util.promote_dtypes_inexact(x, xp)
+  fp_arr, = util.promote_dtypes_inexact(fp)
```
jakevdp, line 1014: Delete duplicate lines

jax/_src/numpy/lax_numpy.py
```diff
@@ -1036,14 +1051,43 @@ def interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
...
+
+@util._wraps(np.interp,
+  lax_description=_dedent("""
+    In addition to constant interpolation supported by NumPy, jnp.interp also
+    supports left='extrapolate' and right='extrpolate' to indicate linear
+    extrpolation instead."""))
+def interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
+           left: Union[ArrayLike, str, None] = None,
+           right: Union[ArrayLike, str, None] = None,
+           period: Optional[ArrayLike] = None) -> Array:
```
jakevdp, line 1080: What if we do this  more dynamically?
```python
static_argnames = []
if isinstance(left, str):
  static_argnames.append('left')
if  isinstance(right, str):
  static_argnames.append('right')
return jit(_interp, static_argnames=static_argnames)(x, xp, fp, left, right, period)
```

jakevdp (commented):
Looks good - out of curiosity, is this `left='extrapolate'`-style API based on anything in scipy/numpy, or something created for this JAX PR?

jax/_src/numpy/lax_numpy.py
shoyer: I am always a little nervous about call `jit` on the fly, rather than `jit` decorated functions. Would caching ensure that this is equivalent?

jax/_src/numpy/lax_numpy.py
shoyer: Done

shoyer:
> Looks good - out of curiosity, is this `left='extrapolate'`-style API based on anything in scipy/numpy, or something created for this JAX PR?

It's new for this JAX PR, but based on `fill_value='extrapolate'` from SciPy. See https://github.com/google/jax/issues/14858 for discussion.

jakevdp (commented):
Looks great! Can you squash the changes into a single commit?

jakevdp (approved):

jakevdp:
It looks like there are a few errors in the expanded test suite (older numpy versions, X64, etc.)

jax/_src/numpy/lax_numpy.py
```diff
@@ -975,15 +988,40 @@ def interp(x: ArrayLike, xp: ArrayLike, fp: ArrayLike,
...
+  static_argnames = []
+  if isinstance(left, str) or left is None:
```
shoyer, line 1015: This broke previously (only on old versions of NumPy!) because I didn't include `or left is None` for deciding to use a static argname.

jakevdp (approved):
Thanks, looks good!

PR #15088, opened 2023-03-20
`custom_vjp` symbolic zeros support

froystig:
#14570 covered custom JVPs. This one covers custom VJPs.

Symbolic zeros are meant to indicate when things that aren't involved in autodiff. In custom AD, this means indicating this to the custom rules, which can then adapt computation accordingly (for instance by saving on unnecessary computation).

For reverse-mode custom AD, there are two ways that these indications come in. As of this PR, let's say we have a python function `jax.custom_vjp(f)` for which we've set `f.defvjp(fwd, bwd, symbolic_zeros=True)`. Then:

1. The forward rule `fwd` is told which primal inputs are not involved in differentiation by a boolean indicator on every argument (more precisely, every leaf on every argument pytree). For example, wherever `f` accepts an argument of the form `(x, [y, z])` where `x`, `y`, and `z` are arrays, then `fwd` accepts `((x, x0), [(y, y0), (z, z0)])`, where `x` is the original primal array and `x0` is a boolean, indicating `True` for a "symbolic zero" (i.e. inactivity in AD).
2. The backward rule `bwd` receives `SymbolicZero` instances in place of cotangents that aren't involved in AD. This mirrors the `custom_jvp` symbolic zero interface (#14570).

Why is support for symbolic zeros simpler in `custom_jvp`? Conceptually, JVPs take primal and tangent inputs in correspondence, and each primal-tangent pair is "zero" in correspondence. Symbolic zeros indicated on the tangents (similar to item 2 above) therefore provide all indication that the rule could need.

Another way to visualize this is by thinking about rows and columns of Jacobian matrices. A primal value not involved in differentiation corresponds to zero columns in the Jacobian matrix. In JVPs, where we compute `x, u -> J(x) @ u`, zeros in the perturbation vector `u` (tangents) line up with zero columns of the Jacobian `J(x)`. In VJPs, where we compute `x, v -> transpose(J(x)) @ v`, zeros in the perturbation vector `v` (cotangents) line up with rows of the Jacobian instead, and we need to point out the zero columns separately from the zero perturbations.

Thanks to @mattjj for helping thinking through the interface here, and for how to "project" the `fwd` function into a canonical form!

### DIFF
--- a/jax/_src/custom_derivatives.py
+++ b/jax/_src/custom_derivatives.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
 from functools import update_wrapper, reduce, partial
 import inspect
 from typing import (Callable, Generic, List, Optional, Sequence, Tuple, TypeVar, Any)
@@ -31,8 +32,8 @@ from jax._src import dtypes
 from jax._src import effects
 from jax._src import linear_util as lu
 from jax._src import traceback_util
-from jax._src.ad_util import (Zero, SymbolicZero, zeros_like_aval,
-                              stop_gradient_p)
+from jax._src.ad_util import (
+    stop_gradient_p, SymbolicZero, Zero, zeros_like_aval)
 from jax._src.api_util import argnums_partial, flatten_fun_nokwargs
 from jax._src.core import raise_to_shaped
 from jax._src.interpreters import ad
@@ -163,12 +164,13 @@ class custom_jvp(Generic[ReturnValue]):
         and the second element is the tangent output. Elements of the input and
         output tuples may be arrays or any nested tuples/lists/dicts thereof.
       symbolic_zeros: boolean, indicating whether the rule should be passed
-        objects representing static symbolic zeros in its tangent tuple
-        argument; otherwise, only standard JAX types (e.g. array-likes) are
-        passed. Setting this option to True allows a JVP rule to detect whether
-        certain inputs are not involved in differentiation, but at the cost of
-        needing special handling for these objects (which e.g. can't be passed
-        into jax.numpy functions). Default False.
+        objects representing static symbolic zeros in its tangent argument in
+        correspondence with unperturbed values; otherwise, only standard JAX
+        types (e.g. array-likes) are passed. Setting this option to ``True``
+        allows a JVP rule to detect whether certain inputs are not involved in
+        differentiation, but at the cost of needing special handling for these
+        objects (which e.g. can't be passed into jax.numpy functions). Default
+        ``False``.
 
     Returns:
       None.
@@ -480,12 +482,15 @@ class custom_vjp(Generic[ReturnValue]):
     self.nondiff_argnums = nondiff_argnums
     self.fwd: Optional[Callable[..., Tuple[ReturnValue, Any]]] = None
     self.bwd: Optional[Callable[..., Tuple[Any, ...]]] = None
+    self.symbolic_zeros = False
 
   __getattr__ = custom_api_util.forward_attr
 
   def defvjp(self,
              fwd: Callable[..., Tuple[ReturnValue, Any]],
-             bwd: Callable[..., Tuple[Any, ...]]) -> None:
+             bwd: Callable[..., Tuple[Any, ...]],
+             symbolic_zeros: bool = False,
+             ) -> None:
     """Define a custom VJP rule for the function represented by this instance.
 
     Args:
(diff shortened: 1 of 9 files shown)


jax/_src/ad_util.py
```diff
@@ -95,7 +95,7 @@ def _stop_gradient_impl(x: T) -> T:
 
 
 class SymbolicZero:
-  def __init__(self, aval: core.AbstractValue) -> None:
+  def __init__(self, aval: core.AbstractValue):
```
mattjj, line 98: Quoting [PEP 484](https://peps.python.org/pep-0484/): "Note that the return type of __init__ ought to be annotated with `-> None`" 😛

mattjj: (I think it's optional these days, I just like it because `__init__` is just a regular method!)

jax/_src/interpreters/batching.py
```diff
@@ -779,8 +782,15 @@ def batch_custom_jvp_subtrace(main, in_dims, *in_vals):
 def batch_custom_vjp_bwd(bwd, axis_name, axis_size, in_dims, out_dim_dests,
                          main_type, spmd_axis_name):
   def new_bwd(*args):
+    in_dims_ = in_dims() if callable(in_dims) else in_dims
+    args = [SymbolicZero(core.mapped_aval(axis_size, dim, x.aval))
+            if type(x) is SymbolicZero else x
+            for x, dim in zip(args, in_dims_)]
+    in_dims_ = [None if type(x) is SymbolicZero else d
+               for x, d in zip(args, in_dims_)]
+    #import ipdb; ipdb.set_trace()
```
mattjj, line 791:
```suggestion
```

jax/_src/interpreters/batching.py
```diff
@@ -779,8 +782,15 @@ def batch_custom_jvp_subtrace(main, in_dims, *in_vals):
 def batch_custom_vjp_bwd(bwd, axis_name, axis_size, in_dims, out_dim_dests,
                          main_type, spmd_axis_name):
   def new_bwd(*args):
+    in_dims_ = in_dims() if callable(in_dims) else in_dims
+    args = [SymbolicZero(core.mapped_aval(axis_size, dim, x.aval))
+            if type(x) is SymbolicZero else x
+            for x, dim in zip(args, in_dims_)]
+    in_dims_ = [None if type(x) is SymbolicZero else d
+               for x, d in zip(args, in_dims_)]
```
mattjj, line 790:
```suggestion
                for x, d in zip(args, in_dims_)]
```

jax/_src/custom_derivatives.py
```diff
@@ -749,30 +791,33 @@ def _custom_vjp_call_jaxpr_abstract_eval(*_, fun_jaxpr, **__):
 def _custom_vjp_call_jaxpr_jvp(
     primals, tangents, *, fun_jaxpr: core.ClosedJaxpr,
     fwd_jaxpr_thunk: Callable[[], Tuple[core.Jaxpr, Sequence[Any]]],
```
mattjj, line 793:
```suggestion
    fwd_jaxpr_thunk: Callable[[???], Tuple[core.Jaxpr, Sequence[Any]]],
```

jax/_src/custom_derivatives.py
```diff
@@ -749,30 +791,33 @@ def _custom_vjp_call_jaxpr_abstract_eval(*_, fun_jaxpr, **__):
...
   tangents_out = map(ad.recast_to_float0, primals_out, tangents_out)
   return primals_out, tangents_out
 ad.primitive_jvps[custom_vjp_call_jaxpr_p] = _custom_vjp_call_jaxpr_jvp
 
-def _custom_vjp_call_jaxpr_vmap(spmd_axis_name,
-    axis_size, axis_name, main_type, args, in_dims, *, fun_jaxpr: core.ClosedJaxpr,
+def _custom_vjp_call_jaxpr_vmap(
+    spmd_axis_name, axis_size, axis_name, main_type, args, in_dims, *,
+    fun_jaxpr: core.ClosedJaxpr,
     fwd_jaxpr_thunk: Callable[[], Tuple[core.Jaxpr, Sequence[Any]]],
```
mattjj, line 819:
```suggestion
    fwd_jaxpr_thunk: Callable[[???], Tuple[core.Jaxpr, Sequence[Any]]],
```

jax/_src/interpreters/ad.py
```diff
@@ -387,16 +387,23 @@ def process_custom_jvp_call(self, _, __, f_jvp, tracers, *, symbolic_zeros):
...
+              for p, t in zip(primals_in, tangents_in)]
+    fwd_in = [x for pair in fwd_in for x in pair]   # flatten
+    res_and_primals_out = fwd.call_wrapped(*fwd_in)
     out_tree, res_tree = out_trees()
     res, primals_out = split_list(res_and_primals_out, [res_tree.num_leaves])
     avals_out = [raise_to_shaped(core.get_aval(x)) for x in primals_out]
+    # We don't need to handle any symbolic zeros on tangents_in or
+    # tangents_out below, because custom_lin_p is never executed and
+    # doesn't correspond to any custom user rule.
+    tangents_in = map(instantiate_zeros, tangents_in)
```
mattjj, line 403: It'd be nice to avoid instantiating zeros here... can we just prune tangents_in and bind custom_lin_p to the nonzeros?

jax/_src/custom_derivatives.py
mattjj: I think maybe `Callable[..., Tuple[core.Jaxpr, Sequence[Any]]]` is the best annotation.

jax/_src/interpreters/ad.py
mattjj: Seems like we could do that and then just adapt `_custom_lin_transpose` to handle the potentially shorter list of return values.

mattjj (approved):
It'd be nice to avoid instantiating zeros in `tangents_in` in `JVPTrace.process_custom_vjp_call`. It seems possible to sort out, but if not leave a TODO?

jax/_src/interpreters/ad.py
froystig: Actually, neither do I in hindsight! Let's take care of this in a new PR.

PR #15092, opened 2023-03-20
[jax2tf] Ensure that the gradient function is serialized natively.

gnecula:
Previously, the recursive invocation of the jax2tf.convert for the gradient function was omitting the native serialization parameters.


gnecula:
@maxwillzq @rxsang PTAL

PR #15114, opened 2023-03-21
Add `wald` distribution to  `random.py`

JiaYaobo:
Counterpart for `numpy.random.wald`

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -1880,3 +1880,59 @@ def _rayleigh(key, scale, shape, dtype) -> Array:
   sqrt_u = lax.sqrt(lax.mul(log_u, n_two))
   ray = lax.mul(scale, sqrt_u)
   return ray
+
+def wald(key: KeyArray,
+         mean: RealArray,
+         scale: RealArray,
+         shape: Optional[Shape] = None,
+         dtype: DTypeLikeFloat = dtypes.float_) -> Array:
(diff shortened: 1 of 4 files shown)


jax/_src/random.py
```diff
@@ -1880,3 +1880,58 @@ def _rayleigh(key, scale, shape, dtype) -> Array:
...
+  k1, k2 = _split(key, 2)
+  mean = mean.astype(dtype)
+  scale = scale.astype(dtype)
+  mean = jnp.broadcast_to(mean, shape)
+  scale = jnp.broadcast_to(scale, shape)
+  v = normal(k1, shape, dtype)
+  z = uniform(k2, shape, dtype)
+  two = _lax_const(mean, 2)
+  y = lax.pow(v, two)
+  mean_sq = lax.pow(mean, two)
```
jakevdp, line 1931: Consider using `lax.integer_pow` here instead (note that in contrast to `lax.pow`, the second argument to `lax.integer_pow` should be a python int)

jakevdp (commented):
Looks good, thanks!

jakevdp (approved):

jakevdp:
Thanks - looking at this more closely now in the context of some work to clarify the documentation for `jax.random` functions, I think I'd prefer to remove the `mean` and `scale` parameters from this. Other distributions in this module avoid specifying these parameters, and the output of the standard wald distribution can be trivially shifted and scaled.

PR #15121, opened 2023-03-21
hilbert transform

joglekara:
Follows SciPy's implementation except for the `ndim > 1` part.

Tested locally to reproduce `osp.signal.hilbert` for even and odd N but could use some guidance on how to implement a proper test


jakevdp:
Thanks! I think as far as testing goes, you could follow some of the FFT tests as a template: https://github.com/google/jax/blob/4bd7f408c60266641ea8db2957ea458edcdcf50d/tests/fft_test.py#L231

But the tests for functions in this file should go in https://github.com/google/jax/blob/main/tests/scipy_signal_test.py

joglekara:
I wrote some of those tests :) Looks like ya'll have improved on them substantially!

```python
E       jax._src.dtypes.TypePromotionError: Input dtypes ('float32', 'complex64') have no available implicit dtype promotion path when jax_numpy_dtype_promotion=strict. Try explicitly casting inputs to the desired output type, or set jax_numpy_dtype_promotion=standard.
```

I'm getting some of these though and I can't quite grok how to make sure both sides are the same dtype. Can I get an assist please?

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,31 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
...
+  if N % 2 == 0:
+    h = jnp.concatenate([jnp.ones(1), jnp.ones(N // 2 - 2) * 2, jnp.ones(1), jnp.zeros(N // 2)])
+    # h[0] = h[N // 2] = 1
+    # h[1:N // 2] = 2
+  else:
+    h = jnp.concatenate([jnp.ones(1), jnp.ones((N + 1) // 2 - 1) * 2, jnp.zeros(N // 2)])
+    # h[0] = 1
+    # h[1:(N + 1) // 2] = 2
+
+  if x.ndim > 1:
```
jakevdp, line 661: I would move this to the top of the function, so the error is raised before doing all the computations on `x`.

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,31 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
...
+  if jnp.iscomplexobj(x):
+    raise ValueError("x must be real.")
+  if N is None:
+    N = x.shape[axis]
+  if N <= 0:
+    raise ValueError("N must be positive.")
+
+  Xf = jnp.fft.fft(x, N, axis=axis)
+  if N % 2 == 0:
+    h = jnp.concatenate([jnp.ones(1), jnp.ones(N // 2 - 2) * 2, jnp.ones(1), jnp.zeros(N // 2)])
```
jakevdp, line 653: For type safety / efficiency, I would explicitly construct these arrays with `Xf.dtype`. And use `jnp.full(..., 2)` rather than `jnp.ones(...) * 2`

Also, is the second term correct? If `N < 4` than this is attempting to construct an array with negative length.

Does the algorithm require that `N >= 4`? If so we should check that and raise an appropriate error.

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,31 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
...
+    # h[0] = h[N // 2] = 1
+    # h[1:N // 2] = 2
+  else:
+    h = jnp.concatenate([jnp.ones(1), jnp.ones((N + 1) // 2 - 1) * 2, jnp.zeros(N // 2)])
```
jakevdp, line 657: Same broad comments here (type safety, negative array lengths. Also, I think this could probably be done without an if statement. The second array length would be `N - 1 - N // 2` in either case, yes?

tests/scipy_signal_test.py
```diff
@@ -390,5 +391,24 @@ def testIstftAgainstNumpy(self, *, shape, dtype, fs, window, nperseg,
...
+    dtype=real_dtypes,
+  )
+  def testHilbert(self, shape, N, axis, dtype):
+    rng = jtu.rand_default(self.rng())
+    args_maker = lambda: (rng(shape, dtype),)
+    name = 'hilbert'
+    jnp_op = getattr(jsp_signal, name)
+    osp_op = getattr(osp_signal, name)
+    jnp_fn = lambda a: jnp_op(a, N=N, axis=axis)
+    osp_fn = lambda a: osp_op(a, N=N, axis=axis)
```
jakevdp, line 407: Since `name` is not changing, it would be simpler to write:
```python
    jnp_fn = lambda a: jsp_signal.hilbert(a, N=N, axis=axis)
    osp_fn = lambda a: osp_signal.hilbert(a, N=N, axis=axis)
```

tests/scipy_signal_test.py
```diff
@@ -390,5 +391,24 @@ def testIstftAgainstNumpy(self, *, shape, dtype, fs, window, nperseg,
...
+    # Numpy promotes to complex128 aggressively.
+    self._CheckAgainstNumpy(osp_fn, jnp_fn, args_maker, check_dtypes=False,
+                            tol=1e-4)
+    self._CompileAndCheck(jnp_op, args_maker)
```
jakevdp, line 411: I think you meant `jnp_fn` here

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,31 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
 
   time = jnp.arange(x.shape[0], dtype=np.finfo(x.dtype).dtype) / fs
   return time, x
+
+@_wraps(osp_signal.hilbert)
+def hilbert(x, N=None, axis=-1):
+  x = jnp.asarray(x)
```
jakevdp, line 643: Please call `check_arraylike` before converting to `jnp.array`

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,31 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
 
   time = jnp.arange(x.shape[0], dtype=np.finfo(x.dtype).dtype) / fs
   return time, x
+
+@_wraps(osp_signal.hilbert)
+def hilbert(x, N=None, axis=-1):
```
jakevdp, line 642: As a public function, this should have type annotations

jakevdp (changes requested):
Looks great! A few comments below. Also please add `hilbert` here so it will appear in the HTML docs: https://github.com/google/jax/blob/main/docs/jax.scipy.rst#jaxscipysignal

jax/_src/scipy/signal.py
jakevdp: Or maybe simpler to write something like this:
```python
h = jnp.zeros(N).at[0].set(1).at[1:N // 2].set(2).at[N // 2].set(1)
```

jax/_src/scipy/signal.py
joglekara: done

jax/_src/scipy/signal.py
joglekara: hmm, I'm mostly just transcribing the `osp` implementation. It is working now, and I think it does need the `if` statement to address the last point given the suggested rewrite

tests/scipy_signal_test.py
joglekara: yup, thanks :)

tests/scipy_signal_test.py
joglekara: indeed!

jax/_src/scipy/signal.py
joglekara: done

jax/_src/scipy/signal.py
joglekara: added

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,33 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
 
   time = jnp.arange(x.shape[0], dtype=np.finfo(x.dtype).dtype) / fs
   return time, x
+
+@_wraps(osp_signal.hilbert)
+def hilbert(x: Array, N: int = None, axis: int = -1):
```
jakevdp, line 642: `N` needs to be `Optional[int]`

joglekara: I like `Optional[int]` too, lets just go with that

jakevdp: `Optional[int]` is more idiomatic

PR #15126, opened 2023-03-21
Document ShapeDtypeStruct

### LINKED ISSUE #15023: jax.ShapeDtypeStruct isn't documented
It is helpful for AOT compilation!

jakevdp:
Fixes #15023

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -2704,6 +2704,16 @@ def device_get(x: Any):
 
 
 class ShapeDtypeStruct:
+  """A container for the shape, dtype, and other static attributes of an array.
+
+  ``ShapeDtypeStruct`` is often used in conjunction with :func:`jax.eval_shape`.
+
+  Args:
+    shape: a sequence of integers representing an array shape
+    dtype: a dtype-like object
+    named_shape: (optional) a dictionary representing a named shape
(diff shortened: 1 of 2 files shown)


jax/_src/api.py
```diff
@@ -2704,6 +2704,16 @@ def device_get(x: Any):
...
 
 class ShapeDtypeStruct:
+  """A container for the shape, dtype, and other static attributes of an array.
+
+  ``ShapeDtypeStruct`` is often used in conjunction with :func:`jax.eval_shape`.
+
+  Args:
+    shape: a sequence of integers representing an array shape
+    dtype: a dtype-like object
+    naned_shape: (optional) a dictionary representing a named shape
```
sharadmv, line 2714: named_shape

sharadmv (approved):

jax/_src/api.py
mattjj: Can we not document `named_shape`? It's experimental and only partially implemented, and I'm not sure if it'll stick around forever.

mattjj: In other words, I don't think it needs to be in the public API (though I may be mistaken).

jakevdp: Done, thanks!

PR #15129, opened 2023-03-21
[jax2tf] Fix grad of pjit in native lowering.

gnecula:
Since jax2tf.convert is called recursively for the purpose of serializing the vjp function, we must ensure that if the primal function is a pjit with shardings then the vjp function must also be converted as a pjit.

Without this fix the serialization with gradients of a pjit function will fail the an error that there are shardings but not pjit at the top-level.


gnecula:
@maxwillzq @rxsang  PTAL

rxsang (approved):

PR #15143, opened 2023-03-22
WAR the dependency issue in the nightly CI container.

nouiz:
This is to help fix error like:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf 22.8.0a0+304.g6ca81bbc78.dirty requires protobuf<3.21.0a0,>=3.20.1, but you have protobuf 3.19.6 which is incompatible.
```

https://github.com/google/jax/actions/runs/4489606264


hawkinsp (approved):

jakevdp (approved):
Sounds good, let's give it a try

nouiz:
If it can be merged today, the next nightly will pick it up.

PR #15148, opened 2023-03-22
Add print statement to help debug spurious test failure

sharadmv:
Will remove once issue is fixed


skye (approved):
I don't think you need to remove it tbh

PR #15149, opened 2023-03-22
Add initial implementation of a `run_state` primitive

sharadmv:
# Background

## Side effects and types
We often say JAX is inspired by functional programming when discussing the requirement that programs in JAX be *pure* to be transformed. However, taking more inspiration from functional programming, we could lift that requirement. Specifically, JAX could allow side-effects, but only effects that JAX's type system can track and ones that have well-defined interactions with transformations.

What does it mean to track effects in the type system? Consider how the Koka programming language incorporates effects into its type system:

```koka
fun sqr    : (int) -> total int       // total: mathematical total function    
fun divide : (int,int) -> exn int     // exn: may raise an exception (partial)  
fun turing : (tape) -> div int        // div: may not terminate (diverge)  
fun print  : (string) -> console ()   // console: may write to the console  
fun rand   : () -> ndet int           // ndet: non-deterministic
```

In Koka, the type of a function not only contains its input and output types, but also a list of effect types. For example, the print function has the `console` effect and the `randr` function has the `ndet` effect. Functions can have multiple effects, which often happens when effectful functions are composed. At the Jaxpr level, JAX currently similarly tracks effect types (specifically the effects field in Jaxprs is the set of side effects associated with the Jaxpr).

We can add a primitive with the ndet effect pretty easily into JAX.

```python
random_p = core.Primitive('random')
@random_p.def_effectful_abstract_eval
def _random_abstract_eval():
  return core.ShapedArray((), jnp.float32), {"ndet"}

def f():
  return random_p.bind()
jaxpr = jax.make_jaxpr(f)()
print(jaxpr.effects)  # ⇒ Prints {"ndet"}
```

The `ndet` effect in the previous example is unhandled, meaning that if we try evaluating this jaxpr or lowering it to MHLO, we will get a type error (something along the lines of “cannot run a primitive with an `ndet` effect” or “cannot lower a jaxpr with an `ndet` effect”). In order to transform this jaxpr into something practical (something we can run, or lower to MHLO), we “discharge” the effect with a transformation.

For example, to discharge the `ndet` effect, we could convert the jaxpr into one that is deterministic but consumes a functional source of randomness, i.e. a `jax.random.PRNGKey`. We write the following “discharge” transformation:
`run_ndet :: (a -> {ndet} b) -> (PRNGKey -> a -> b)`

Pairing side-effects with discharging transformations enables their usage with other parts of JAX. For example, we could now `vmap` the function returned from `run_ndet`.

## The state effect from Dex
JAX takes inspiration from Dex and Koka for its effect types. One of the effects from Dex that is of particular interest is the *state* effect.

From the Dex prelude, we see that Dex offers a set of effectful functions (`get`, `:=` and `ask`) that each consume a special Ref type and emit a side-effect.
```dex
def Ref (r:Type) (a:Type) : Type = %Ref r a
def get  {h s} (ref:Ref h s)       : {State h} s    = %get  ref
def (:=) {h s} (ref:Ref h s) (x:s) : {State h} Unit = %put  ref x
def ask  {h r} (ref:Ref h r)       : {Read  h} r    = %ask  ref
```

For example, when we pass a `Ref h s` into `get`, we are returned a value of type `s` and emit a `State h` effect. 

Dex has already established how these effectful primitives are transposed, so we follow their lead in incorporating these stateful primitives into JAX.

# Reference types and stateful primitives
We add a new type to JAX and Jaxprs: `Ref`s.
```
data Ref = Ref Array
```

`Ref`s are basically wrapper around Array types that allow reading from and writing into the Array. Note that Arrays in JAX are immutable, so Refs offer mutation semantics.

`Ref`s can be input to Jaxprs but not outputs. An example Jaxpr is:
```
{ lambda ; a:Ref{float32[4]} b:Ref{bool[]} .
  ...
  in ...
}
```

### Why not make mutable arrays?
Why not just make Arrays mutable? Well, we can index into an Array a giving us another Array b. If we mutate b does that also mutate a? Tracking these aliased arrays can be really tricky. We can circumvent this particular aliasing problem by making a single mutable type that produces immutable types when read from.

## Effectful primitives
Refs can only be manipulated/used with 3 new JAX primitives: `get`, `swap`, and `addupdate`. All of these primitives can be provided indices that are used to manipulate subarrays in the Ref. Like in NumPy, we can index Refs using Python integers, Python slice objects, and integer Arrays.
```
type Indices = Tuple[Int | Slice | Array, ...]
```

Each of the state primitives is effectful, specifically when called, it emits an effect that is recorded into the type signature of the containing Jaxpr. Stateful primitives emit effects that are tied to their input Refs. In the notation below, we indicate this by using a “heap parameter” h. In practice, we tie the effect to the corresponding abstract reference value when determining the type of the Jaxpr. We cannot lower or evaluate Jaxprs that have state effects present.

### `get`
`get` is a JAX primitive that reads from an `Ref`. Specifically, it can do indexed reads from a `Ref` and returns an Array while emitting the Read effect.
```
get :: Ref h -> Indices -> {Read h} Array
```

Example usage:
```python
x = get(x_ref, (0, slice(0, 2)))
# Equivalent sugar:
y = x_ref[0, :2]
```

### `swap`

`swap` is a JAX primitive that first performs an indexed read from a Ref and then updates the Ref at the provided indices with a new value Array and returns the old value.
```
swap :: Ref h -> Indices -> Array -> {Write h} Array
```

Example usage:
```python
x = swap(x_ref, (0, slice(0, 2)), y)
# Analogous to: x, x_ref[0, :2] = x_ref[0, :2], y
```

We also use the shorthand set when the value returned by swap is ignored.
```python
set(x_ref, (0, slice(0, 2)), y)
# Equivalent sugar:
x_ref[0, :2] = y
```

### `addupdate`
`addupdate` is a JAX primitive that accumulates (i.e. adds) a value into a Ref at some specified indices.
```
addupdate :: Ref h -> Indices -> Array -> {Accum h} ()
```

Example usage:
```python
addupdate(x_ref, (0, slice(0, 2)), y) # Analogous to `x_ref[0, :2] += y`
```

# Creating Refs and discharging state effects

Thus far we have discussed manipulating Refs and their associated side-effects. We have not discussed how Refs are created and how the side-effects are discharged to produce functions that are compilable.

We introduce `run_state`, a higher-order function that creates Refs initialized to particular values, runs a stateful function, and returns the final values of the Refs after they have been mutated.
```
run_state :: ([Ref a] -> {Read a, Write a, Accum a | eff} ()) -> [a] -> {eff} [a]
```

`run_state` doesn’t have any state effects corresponding to the created Refs, “discharging” the state effects. Note that other effects `{eff}` will remain (e.g. error effects, or state effects corresponding to outer Refs).

An important point here is that we have not provided a way of creating Refs in “top level” user code. What we mean by this is that only run_state creates Refs, and there is no make_ref function.

## Implementation

`run_state` runs a Jaxpr interpreter (i.e. it's an initial-style primitive) that keeps track of the values associated with Refs. When the interpreter hits an effectful primitive that operates on a Ref introduced by run_state, we run a discharge rule wherein we convert the stateful semantics into value semantics. For example, the `get` primitive can be discharged into `lax.dynamic_slice` or `lax.gather` depending on the index pattern. Similarly `swap` is discharged into a `dynamic_slice/gather` and `dynamic_update_slice/scatter`. `addupdate` is discharged into a `dynamic_slice/gather`, `add`, and `dynamic_update_slice/scatter`. `run_state` is also responsible for discharging higher-order primitives that contain state effects (for example, control flow primitives or nested jits). Note that `run_state` is only responsible for discharge effects corresponding to the Refs it introduces and no others.

## Usage with control flow
State primitives should *just work* with control flow primitives. You won’t, however, be able to pass Refs into them or return Refs from them. Reading from and writing into closed-over Refs should be sufficient.

#### Implementing a simple pure function by discharging state effects

Using only one read-write Ref
```python
def f(x_ref):
  x = get(x_ref, ())
  set(x_ref, (), x * 2)

def mul2(x):
  return run_state(f)(x)
```

Using read-only and write-only Refs
```python
def f(refs):
  x_ref, y_ref = refs
  x = get(x_ref, ())
  set(y_ref, (), x * 2)

def mul2(x):
  return run_state(f)(x, 0)[1]
```

#### Array indexing
```python
def f(x_ref, i_ref, v_ref):
  i = get(i_ref, ())
  set(x_ref, (i,), get(v_ref, ()))

def set_at_index(x, i, v):
  return run_state(f)(x, i, v)[0]
set_at_index(jnp.arange(4), 2, -1)
# ⇒ Returns [0, 1, -1, 3]
```

#### Nesting state effects
Here’s an example where we nest two `run_state`s with non-overlapping state effects
```python
def square_ref(x_ref):
  x = x_ref[()]
  set(x_ref, (), jnp.square(x))

def f(x_ref, y_ref):
  x = x_ref[()]
  mul_x = run_state(square_ref)(x)
  set(y_ref, (), mul_x)
run_state(f)(2, 0)
# ⇒ Returns (2, 4)
```

Here’s an example where we mutate an outer Ref from a nested run_state
```python
def f(x_ref, y_ref):
  def set_x(z_ref):
    # Mutating a closed-over `Ref`
    set(x_ref, (), get(z_ref, ()))
  run_state(set_x)(4)
  set(y_ref, (), get(x_ref, ()))
run_state(f)(2, 0)
# ⇒ Returns (4, 4)
```

#### Using state inside of control flow
Here we write to a Ref inside of a `fori_loop` body.
```python
def f(x_ref, y_ref):
  def body(i, _):
    set(y_ref, (i,), get(x_ref, (i,)) + 1)
  lax.fori_loop(0, x_ref.shape[0], body, ())
run_state(f)(jnp.arange(4), jnp.zeros(4))[1]
# ⇒ Returns [1., 2., 3., 4.]
```

Here we conditionally write to a Ref using a `lax.cond`
```python
def f(x_ref):
  def true_fun():
    set(x_ref, (), 100)
  def false_fun():
    pass
  x = get(x_ref, ())
  lax.cond(x < 100, true_fun, false_fun)
run_state(f)(5)   # ⇒ Returns 100
run_state(f)(105) # ⇒ Returns 105
```

# Transformations of effectful functions

In general, we disallow passing `Ref`s into higher order functions. They *must* be closed over, meaning we only need to take care of some cases under transformations. However, `run_state` should always be transformable by JAX functions, since it has a pure signature (it discharges away state effects).
## Autodifferentiation

If we are differentiating a function, we can safely assume there are no Ref inputs and therefore we are not differentiating with respect to any Refs. Functions we are differentiating that close over Refs and updating them is fair game. However, we will never need to apply transformation rules to the primitives in this case since they will only ever show up as purely primal values (no associated tangents). They will similarly not show up in the backwards pass so no transpose rule needed.

In this example, we are mutating x_ref inside of a function we are taking `jax.grad` of. `x` is mutated in the forward pass and is untouched during the backward pass.
```python
def f(x_ref):
  def mul(x):
    z = x_ref[()]
    x_ref[()] += 1
    return jnp.sin(x * z)
  x = x_ref[()]
  z = jax.grad(mul)(x)
  x_ref[()] = z
run_state(f)(2)
```

Differentiating through `run_state` is something we also support, but involves dealing with transforming the state primitives. When we are differentiating w.r.t. to values that are then passed into run_state to initialize Refs, we will need to differentiate through the body of run_state, which contains state side-effects.

With these rules in place, we can mechanically differentiate run_state with little modification to JAX’s core machinery.

# Examples with transformations

Differentiating through run_state
```python
def f(x_ref):
  x_ref[()] = jnp.sin(x_ref[()])
jax.grad(lambda x: run_state(f)(x))(4.) ⇒ Returns cos(4.)
```

Digging deeper into this, the function we are differentiating is this (in jaxpr representation):
```
{ lambda ; a:f32[]. let
    b:f32[] = run_state[
      jaxpr={ lambda ; c:Ref{float32[]}. let
          d:f32[] = get c ()
          e:f32[] = sin d
          _:f32[] = swap c () e
        in () }
    ] a
  in (b,) }
```

When we take its gradient, we obtain the following function:
```
{ lambda ; a:f32[]. let
    _:f32[] b:f32[] = run_state[
      jaxpr={ lambda ; c:Ref{float32[]} d:Ref{float32[]}. let
          e:f32[] = get c ()
          f:f32[] = sin e
          g:f32[] = cos e
          _:f32[] = swap c () f
          _:f32[] = swap d () g
        in () }
    ] a
    _:f32[], h:f32[] = run_state[
      jaxpr={ lambda ; i:Ref{float32[]} j:Ref{float32[]}. let
          k:f32[] = get i ()
          l:f32[] = swap j () 0.0
          m:f32[] = mul l k
          addupdate j () m
        in () }
    ] b 1.0
  in (h,) }
```

Notice that we have two `run_state`s in our gradient function. The first corresponds to the forward pass where we compute `sin(x)` and the residual term `cos(x)`. The residual term is passed into the second `run_state` and read into the variable k. We then perform the transpose where the `get` and `swap` in the first jaxpr are turned into an `addupdate` and `swap` respectively (and run in reverse). The end result is that we write `cos(x) * 1.` into `j` and return `cos(x)` from the function.

Still TODO: vmap rule

### DIFF
--- a/jax/_src/state/discharge.py
+++ b/jax/_src/state/discharge.py
@@ -15,24 +15,34 @@
 from __future__ import annotations
 import dataclasses
 from functools import partial
+import operator
 
-from typing import Any, Dict, List, Optional, Protocol, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Protocol, Sequence, Tuple, Union
 
 import numpy as np
 
 from jax import lax
 
+from jax._src import api_util
+from jax._src import ad_util
 from jax._src import core
 from jax._src import linear_util as lu
 from jax._src.interpreters import partial_eval as pe
-from jax._src.state.types import AbstractRef
+from jax._src import source_info_util
+from jax._src import tree_util
+from jax._src.config import config
+from jax._src.interpreters import ad
+from jax._src.state.types import AbstractRef, RefEffect
 from jax._src.state.primitives import get_p, swap_p, addupdate_p
-from jax._src.util import safe_map, safe_zip, split_list
+from jax._src.state.utils import hoist_consts_to_refs
+from jax._src.util import (safe_map, safe_zip, split_list, weakref_lru_cache,
+                           partition_list, merge_lists, split_dict)
 
 ## JAX utilities
 
 map, unsafe_map = safe_map, map
 zip, unsafe_zip = safe_zip, zip
+PyTreeDef = tree_util.PyTreeDef
 
 ## Discharging state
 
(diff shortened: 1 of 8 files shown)


froystig:
This is it! We've jumped the shark!

sharadmv:
> @sharadmv Great work and thank you for the detailed write-up!

Thanks!

> BTW, I think the type of `run_state` should have been
> 
> ```haskell
> run_state :: ([Ref a] -> {Read a, Write a, Accum a | eff} ()) -> [a] -> {eff} [a]
> ```
Fixed, thanks for the catch.

krzysztofrusek:
@sharadmv Best PR doc I have ever seen !

jax/_src/state/discharge.py
```diff
@@ -241,3 +251,473 @@ def _closed_call_discharge_rule(
...
+  # "propagate" out its inner effects. Otherwise, the effects are local to this
+  # `run_state`.
+  is_ref = set(i for i, aval in enumerate(avals)
+               if isinstance(aval, AbstractRef))
+  nonlocal_effects = set()
+  for eff in jaxpr.effects:
+    if isinstance(eff, RefEffect):
+      if eff.input_index not in is_ref:
+        continue
+    nonlocal_effects.add(eff)
```
mattjj, line 291:
```suggestion
  nonlocal_effects = {e for e in jaxpr.effects if isinstance(e, RefEffect) and e.input_index in is_ref}
```

jax/_src/state/discharge.py
```diff
@@ -241,3 +251,473 @@ def _closed_call_discharge_rule(
...
+    if out_unknowns == in_unknowns:
+      break
+    in_unknowns = map(operator.or_, in_unknowns, out_unknowns)
+  else:
+    raise Exception("Invalid fixpoint")
+  del out_unknowns  # redundant since it's the same as `in_unknowns`
+  tracers = tuple(trace.instantiate_const(t) if uk else t  # type: ignore
+                  for t, uk in zip(tracers, in_unknowns))
+
+  # We use `partial_eval_jaxpr_custom` here because it won't remove effectful
```
mattjj, line 398:
```suggestion
  # We use `partial_eval_jaxpr_stateful` here because it won't remove effectful
```

jax/_src/state/discharge.py
```diff
@@ -241,3 +251,473 @@ def _closed_call_discharge_rule(
...
+    res_jaxpr_avals = [v.aval for v in res_jaxpr_.invars]
+    _, res_avals = split_list(res_jaxpr_avals, [num_known])
+    res_avals = [a.inner_aval for a in res_avals]
+    all_avals = [*res_avals, *[v.aval for v in res_jaxpr_.outvars]]
+    empty_res = map(ad.zeros_like_aval, all_avals)
+    res_jaxpr, _ = _convert_outputs_to_writes(res_jaxpr_)
+    res = run_state_p.bind(*res_args, *empty_res, jaxpr=res_jaxpr,
+                           which_linear=(False,) * (len(res_args) + len(empty_res)))
+    res = res[len(res_args):]
+    ref_res_, nonref_res_ = split_list(res, [num_res_ref])
```
mattjj, line 609:
```suggestion
    _, ref_res_, nonref_res_ = split_list(res, [len(res_args), num_res_ref])
```

jax/_src/state/discharge.py
```diff
@@ -241,3 +251,473 @@ def _closed_call_discharge_rule(
...
+
+def _run_state_abstract_eval(*avals: core.AbstractValue, jaxpr: core.Jaxpr,
+                             which_linear: tuple[bool, ...]):
+  del which_linear
+  # When we abstractly evaluate `run_state`, we want to keep track of which
+  # input avals are `Ref`s and which are not. If an aval is a `Ref`, we want to
+  # "propagate" out its inner effects. Otherwise, the effects are local to this
+  # `run_state`.
+  is_ref = set(i for i, aval in enumerate(avals)
+               if isinstance(aval, AbstractRef))
```
mattjj, line 285:
```suggestion
  is_ref = {i for i, a in enumerate(avals) if isinstance(a, AbstractRef))
```

jax/_src/state/discharge.py
```diff
@@ -241,3 +251,473 @@ def _closed_call_discharge_rule(
...
+@weakref_lru_cache
+def _initial_style_jaxpr(fun, in_tree, in_avals):
+  fun_, out_tree_thunk = api_util.flatten_fun_nokwargs(lu.wrap_init(fun), in_tree)
+  debug = pe.debug_info(fun_, in_tree, False, out_tree_thunk, 'run_state')
+  jaxpr, _, consts = pe.trace_to_jaxpr_dynamic(fun_, in_avals, debug)
+  return jaxpr, consts, out_tree_thunk()
+
+def run_state(f: Callable[..., None]):
+  def wrapped(*args):
+    singleton = len(args) == 1
```
mattjj, line 693: This feels weird to me... as we discussed!

mattjj: Two options that seem more internally coherent:

```python
# "no splatting" option, not as nice for f but nicer for caller of run_state
def f(refs):
  x_ref, = refs
  ...

x_final = run_state(f)(x_init)
assert tree_structure(x_final) == tree_structure(x_init)
```

```python
# splat but tuple output, nicer for f but not as nice for caller of run_state (must unpack output)
def f(x_ref):
  ...

x_final, = run_state(f)(x_init)
assert tree_structure(x_final) == tree_structure(x_init)
```

Or there's the current option of special-casing single input to be automatically unpacked.

jax/_src/state/primitives.py
```diff
@@ -429,6 +429,7 @@ def _get_vmap(batched_args, batched_dims, *, indexed_dims):
 batching.primitive_batchers[get_p] = _get_vmap
 
 def _swap_vmap(batched_args, batched_dims, *, indexed_dims):
+  from jax import lax
```
mattjj, line 432: Can we fix? You said in person we maybe can avoid this. Would be sick.

mattjj (approved):
Sick!

jax/_src/state/discharge.py
soraros: Anything but special-casing, please. IMO, we already have one too many special-casing (`primitive.multiple_results`).

jax/_src/state/discharge.py
sharadmv: I can't use this directly because we need to add *non*-RefEffects to `nonlocal_effects`.

jax/_src/state/discharge.py
mattjj: Someday we're going to get rid of that! 😅

sharadmv: I went w/ option 1

mjsML:
The PR description should land in a JEP :)

PR #15169, opened 2023-03-23
Correctly handle seq_lengths in lstm_ref

cgarciae:
# Changes
* Correctly returns the last valid carry according to `seq_lengths`
* Zero-out padded outputs
* Improves tests to check for NaNs in loss and grad of the `lstm` and `lstm_ref` functions.

cc @zhangqiaorjc

### DIFF
--- a/jax/experimental/rnn.py
+++ b/jax/experimental/rnn.py
@@ -289,6 +289,13 @@ def lstm_ref(x: Array, h_0: Array, c_0: Array, W_ih: Dict[int, Array],
     h = o * tanh(c)
     return (h, c), h
 
+  # here we also output the carry so that we can later slice
+  # the correct carry according to seq_lengths, while this takes more memory
+  # it is faster than using 'jnp.where' inside the scan loop
+  def scan_fn(cell, carry, x):
+    carry, y = cell(carry, x)
+    return carry, (carry, y)
+
   seq_first_y = x.transpose(1, 0, 2)
   if not bidirectional:
     final_h = []
@@ -296,8 +303,10 @@ def lstm_ref(x: Array, h_0: Array, c_0: Array, W_ih: Dict[int, Array],
     for l in range(num_layers):
       cell = partial(
           lstm_cell, W_ih=W_ih[l], W_hh=W_hh[l], b_ih=b_ih[l], b_hh=b_hh[l])
-      (h_t, c_t), seq_first_y = jax.lax.scan(cell, (h_0[l], c_0[l]),
+      cell_fn = partial(scan_fn, cell)
+      out = jax.lax.scan(cell_fn, (h_0[l], c_0[l]),
                                              seq_first_y)
+      (h_t, c_t), seq_first_y = _extract_output(seq_lengths, out)
       final_h.append(h_t)
       final_c.append(c_t)
     h_n = jnp.stack(final_h)
@@ -310,12 +319,19 @@ def lstm_ref(x: Array, h_0: Array, c_0: Array, W_ih: Dict[int, Array],
   for l in range(num_layers * 2):
     cell = partial(
         lstm_cell, W_ih=W_ih[l], W_hh=W_hh[l], b_ih=b_ih[l], b_hh=b_hh[l])
+    cell_fn = partial(scan_fn, cell)
     if l % 2 == 0:
-      (h_t, c_t), seq_first_y_fwd = jax.lax.scan(cell, (h_0[l], c_0[l]),
+      out = jax.lax.scan(cell_fn, (h_0[l], c_0[l]),
                                                  seq_first_y)
+      (h_t, c_t), seq_first_y_fwd = _extract_output(seq_lengths, out)
     else:
-      (h_t, c_t), seq_first_y_bwd = jax.lax.scan(
-          cell, (h_0[l], c_0[l]), seq_first_y, reverse=True)
+      # reverse sequence while keeping padding at the end
+      seq_first_y_reversed = _flip_sequence(seq_first_y, seq_lengths)
+      out = jax.lax.scan(
+          cell_fn, (h_0[l], c_0[l]), seq_first_y_reversed)
+      (h_t, c_t), seq_first_y_bwd = _extract_output(seq_lengths, out)
+      # align reversed sequence with original sequence
+      seq_first_y_bwd = _flip_sequence(seq_first_y_bwd, seq_lengths)
       # Inputs to next layer are concat'ed from fwd and bwd.
       seq_first_y = jnp.concatenate([seq_first_y_fwd, seq_first_y_bwd], axis=-1)  # pytype: disable=name-error
     final_h.append(h_t)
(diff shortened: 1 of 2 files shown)


hawkinsp:
(The lint failure is preexisting/unrelated)

zhangqiaorjc (approved):

zhangqiaorjc:
Thanks @cgarciae !

zhangqiaorjc:
@cgarciae needs to run linter

jax/experimental/rnn.py
```diff
@@ -322,6 +334,22 @@ def lstm_cell(carry, x, *, W_ih, W_hh, b_ih, b_hh):
...
+  # [batch, seq_len, hidden_size]
+  seq_first_y = jnp.where(
+      mask[..., None], # [seq_len, batch, 1]
+      seq_first_y,     # [seq_len, batch, hidden_size]
+      0)
+  return (h_t, c_t), seq_first_y
+
+def _select_last(seq_lengths: Array, sequence: Array):
+  last_idx = seq_lengths[None, :, None] - 1
+  return jnp.take_along_axis(sequence, last_idx, axis=0)
```
sharadmv, line 352: If I understand this correctly, seq_lengths is [batch_size] and sequence is [max_seq_len, batch_size, feature_dim].

We want an output that is [batch_size, feature_dim].

I think therefore we just need to return:
```
sequence[seq_lengths - 1, jnp.arange(sequence.shape[1])]
```

(I verified this internally)

sharadmv (changes requested):
Found a small indexing bug

jax/experimental/rnn.py
cgarciae: Thanks for catching this @sharadmv! I forgot to squeeze the first dimension when porting this code from Flax. I've confirmed your code is equivalent to:

```python
  last_idx = seq_lengths[None, :, None] - 1
  return jnp.take_along_axis(sequence, last_idx, axis=0)[0]
```
Will be keeping yours as its more readable.

tests/experimental_rnn_test.py
```diff
@@ -57,45 +59,133 @@ def test_lstm(self, batch_size: int, seq_len: int, input_size: int,
...
+    self.assertFalse(np.isnan(loss_ref))
+    self.assertFalse(np.isnan(grad_ref).any())
 
     np.testing.assert_allclose(y_ref, y, rtol=1e-05, atol=1e-5)
     np.testing.assert_allclose(h_n_ref, h_n, rtol=1e-05, atol=1e-5)
     np.testing.assert_allclose(c_n_ref, c_n, rtol=1e-05, atol=1e-5)
 
+  # @jtu.sample_product(
+  #   batch_size=[1, 4],
+  #   seq_len=[1, 4],
```
sharadmv, line 121: Could you add a TODO to enable this here and file a GH issue?

sharadmv (approved):

sharadmv:
Are you able to rebase on main?

jax/experimental/rnn.py
```diff
@@ -308,12 +317,15 @@ def lstm_cell(carry, x, *, W_ih, W_hh, b_ih, b_hh):
...
+    cell_fn = partial(scan_fn, cell)
     if l % 2 == 0:
-      (h_t, c_t), seq_first_y_fwd = jax.lax.scan(cell, (h_0[l], c_0[l]),
+      out = jax.lax.scan(cell_fn, (h_0[l], c_0[l]),
                                                  seq_first_y)
+      (h_t, c_t), seq_first_y_fwd = _extract_output(seq_lengths, out)
     else:
-      (h_t, c_t), seq_first_y_bwd = jax.lax.scan(
-          cell, (h_0[l], c_0[l]), seq_first_y, reverse=True)
+      out = jax.lax.scan(
```
sharadmv, line 326: I think there's an issue here.

The way that it's set up, the `h, c` are returned as part of the scan `ys`. However, since the `scan` has `reverse=True`, the `seq_first_y` will be in the right order but the `h_t, c_t` will be reversed.

See this example:
```python
def body(i, x):                            
  return i + 1, (i, x + 1)                 
                                           
print(jax.lax.scan(body, 0, jnp.arange(4), reverse=True))
# Prints (4, ([3, 2, 1, 0], [1, 2, 3, 4]))
```

See how the `i` in the carry was reversed but the `x + 1` is not.

Therefore when you call `_extract_output` on `h_t, c_t`, it will be selecting out the *initial* state, not the *last* state. 

There are two options for a fix here:

1) reverse `h_t, c_t` after they are returned from the scan
2) instead of using `scan(..., reverse=True)`, manually reverse `seq_first_y` and reverse `seq_first_y_bwd` at the end.

jax/experimental/rnn.py
```diff
@@ -308,12 +317,15 @@ def lstm_cell(carry, x, *, W_ih, W_hh, b_ih, b_hh):
...
+          cell_fn, (h_0[l], c_0[l]), seq_first_y, reverse=True)
+      (h_t, c_t), seq_first_y_bwd = _extract_output(seq_lengths, out)
```
sharadmv, line 328: `seq_lengths` need to be handled differently here since the sequences are reversed.

One solution would be to `jnp.roll` the sequences by `max_seq_len - seq_lengths`, then do the reversed scan, then roll them back by the same negative amount.

sharadmv (changes requested):

jax/experimental/rnn.py
cgarciae: You are totally right @sharadmv! `2` is the correct approach, however, you cannot just revert, you have to revert while keeping padding at the end. I just pushed some logic we had for this.

cgarciae: Check out the `_flip_sequence` function.

jax/experimental/rnn.py
cgarciae: This is tied to the previous comment, `_flip_sequence` should solve this.

sharadmv (approved):

jax/experimental/rnn.py
```diff
@@ -324,6 +340,38 @@ def lstm_cell(carry, x, *, W_ih, W_hh, b_ih, b_hh):
...
+  max_steps = sequence.shape[0]
+
+  # create indexes
+  idxs = jnp.arange(max_steps - 1, -1, -1)[:, None] # [seq_len, 1]
+  seq_lengths = seq_lengths[None] # [1, batch_size]
+  # calculate the reversel index order, padding will remain at the end
+  idxs = (idxs + seq_lengths) % max_steps # [seq_len, batch_size]
+  # Select the inputs in flipped order.
+  outputs = jnp.take_along_axis(sequence, idxs[:, :, None], axis=0)
+  return outputs
```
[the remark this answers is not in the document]
cgarciae: I like the idea (not sure about performance).
I'll use it and add some comments.

sharadmv (approved):
Thanks!!

cgarciae:
GPU tests will probably fail until the gradient for `rnn.lstm` if fixed.
I can rebase when that is ready

tests/experimental_rnn_test.py
```diff
@@ -37,14 +38,10 @@ class RnnTest(jtu.JaxTestCase):
...
-    batch_size = 6
-    seq_len = 7
-    input_size = 8
-    hidden_size = 12
-    num_layers = 5
     num_directions = 2 if bidirectional else 1
 
-    seq_lengths = jnp.ones((batch_size,), dtype=jnp.int32) * seq_len
+    seq_lengths = jax.random.randint(
+      jax.random.PRNGKey(0), (batch_size,), 0, seq_len)
```
sharadmv, line 44: I think this needs to have dtype=jnp.int32.

sharadmv (changes requested):

sharadmv:
LSTM should be fixed now

sharadmv (approved):

PR #15187, opened 2023-03-24
[dynamic-shapes] djax+pjit staging to jaxpr

mattjj:
This is the first in a sequence of PRs making the dynamic shape machinery work with `pjit`, aka the new "initial-style" `jit`. (Previously it only worked with the old "final-style" `jit`.)

Concretely, we want to make all the tests in `tests/dynamic_api_test.py` work again (better than ever!).

We're going to break this down into a sequence of steps:
1. [x] staging-to-jaxpr e.g. with `jax.make_jaxpr` (this PR, #15187)
3. [ ] padded-and-masked bint-based lowering and execution (e.g. with XLA, using our own padding-and-masking and/or using the HLO DynamicPadder)
4. [ ] dynamic shape compiler lowering and execution (e.g. with IREE)
5. [ ] autodiff e.g. with `jax.jvp`, `jax.linearize`, and `jax.grad`
6. [ ] batching e.g. with `jax.vmap`

Once we get these things working again, we'll be able to start landing new features!

---

This PR tackles staging to dynamic shape jaxprs. The relevant tests, moved to their own `DynamicShapeStagingTest` class and un-skipped in this PR, basically involve using `jax.make_jaxpr`, though the main work is in the underlying `partial_eval.trace_to_jaxpr_dynamic2` and the changes to `pjit`.

So most of this PR is adapting pre-existing stuff for the old `jit` to work with the new `jit`/`pjit`.

Here's an overview of the changes made in this PR and why they were needed:
* The changes in tests/dynamic_api_test.py are just code movement, plus renaming like `xla_call_p` -> `pjit` and `call_jaxpr`->`jaxpr`.
* In partial_eval.py, we needed to handle a previously unhandled case where constants closed over by a `pjit`-decorated function had `Tracer`s in their shapes (and hence those Tracers need to be lifted into the `pjit`-trace). Conceptually, because with dynamic shapes `Tracer`s can now have `Tracer`s in their shapes, whatever we did to handle closed-over `Tracer`s in the static shape world, we now also have to apply to `Tracer`s in the shapes of closed-over `Tracer`s. The reason this case wasn't handled in the previous implementation is that due to a detail of final-style tracing we would hit the `DynamicJaxprTrace.sublift` where with initial-style we now hit `DynamicJaxprTrace.new_const`; see the pre-existing comment in the body of `DynamicJaxprTrace.sublift`. The change here was just a one liner to call `DynamicJaxprTrace._lift_tracers_in_aval`.
* In pjit.py, we had to plumb through the `abstracted_axes` API option, and use it in the call to the new helper `_flat_axes_specs` which resolves the `abstracted_axes` user annotations to the internal abstracted axis size specification. We then use that specification to determine the input type on which to trace the `jit`-decorated Python callable (and the corresponding jaxpr's input type) by calling the existing utility `partial_eval.infer_lambda_input_type`.
* Also in pjit.py, we use that input type `in_type: core.InputType` in place of the tuple of avals `global_in_avals: core.AbstractValue`. The previous `global_in_avals`, representing things like `(f32[3], f32[4, 5])`, was sufficient to represent statically shaped input types on which to trace `jit`-decorated functions. But with dynamic shapes we need to generalize that to represent things more like `(n:i32[], f32[n], f32[4,5])`. That is, a tuple of avals is no longer sufficient. But it's useful to remember that the logic is staying the same: we always had to trace on an input type, it's just that dynamic shape input types are more general.
* The generalization to `core.InputType` has a corresponding generalization of the calling convention for the `pjit_p` primitive: if we apply `jit(f)(x)` for some dynamically-shaped `x` (dynamically shaped due to some outer `jit`), we need to build an input type with axis size arguments, like `(n:i32[], f32[n])`, and the convention is that the `pjit_p` primitive needs to be applied to _two_ arguments, corresponding to the pair `(x.shape[0], x)`. That is, with dynamic shapes, we need to automatically extract implicit-to-user-code axis size arguments. That's also done in pjit.py, in the function `_extract_implicit_args`, which is like an intial-style analogue of the pre-existing `partial_eval._extract_implicit_args` utility for final-style.
* The staging rule for `pjit_p` also needs to be generalized, mainly to be able to produce output `Tracer`s with `Tracer`s in their shapes, i.e. to be able to produce dynamically-shaped outputs.

All of the changes are guarded by the `jax.config.jax_dynamic_shapes` flag, so unless I've made a mistake we shouldn't be breaking anyone's code.

### DIFF
--- a/jax/_src/pjit.py
+++ b/jax/_src/pjit.py
@@ -31,9 +31,9 @@ from jax.errors import JAXTypeError
 from jax.interpreters import partial_eval as pe
 from jax.interpreters import xla
 from jax._src.interpreters.pxla import PartitionSpec
-from jax.tree_util import (
+from jax._src.tree_util import (
     tree_map, tree_flatten, tree_unflatten, treedef_is_leaf, tree_structure,
-    treedef_tuple)
+    treedef_tuple, broadcast_prefix, all_leaves)
 
 from jax._src.sharding import Sharding
 from jax._src.sharding_impls import (
@@ -65,6 +65,9 @@ from jax._src.util import (
     HashableFunction, safe_map, safe_zip, wraps,
     distributed_debug_log, split_list, tuple_insert, weakref_lru_cache,
     merge_lists)
+
+map, unsafe_map = safe_map, map
+zip, unsafe_zip = safe_zip, zip
 
 traceback_util.register_exclusion(__file__)
 
@@ -162,7 +165,7 @@ def _get_arg_names(fun, in_tree, args_flat):
 def _device_assignment_mismatch_error(fun_name, fails, args_flat, api_name,
                                       arg_names):
   arg_list = []
-  for a, n in safe_zip(args_flat, arg_names):
+  for a, n in zip(args_flat, arg_names):
     da = a.sharding._device_assignment if hasattr(a, 'sharding') else None
     arg_list.append((n, da, shaped_abstractify(a)))
 
@@ -312,9 +315,6 @@ def _resolve_axis_resources_and_shardings_arg(
 def pre_infer_params(fun, in_shardings, out_shardings,
                      donate_argnums, static_argnums, static_argnames, device,
                      backend, abstracted_axes):
-  # TODO(yashkatariya, mattjj): Remove when pjit supports dynamic shapes.
-  if jax.config.jax_dynamic_shapes:
-    raise ValueError("Dynamic shapes is not supported with pjit yet.")
   if abstracted_axes and not jax.config.jax_dynamic_shapes:
     raise ValueError("abstracted_axes must be used with --jax_dynamic_shapes")
 
@@ -414,12 +414,13 @@ class PjitInfo(NamedTuple):
   keep_unused: bool
   inline: bool
   resource_env: Any
+  abstracted_axes: Optional[Any]
 
 
 def common_infer_params(pjit_info_args, *args, **kwargs):
   (fun, user_in_shardings, user_out_shardings, static_argnums, static_argnames,
    donate_argnums, device, backend, keep_unused, inline,
-   resource_env) = pjit_info_args
+   resource_env, abstracted_axes) = pjit_info_args
 
   if kwargs and not _is_unspecified(user_in_shardings):
     raise ValueError(
@@ -435,6 +436,8 @@ def common_infer_params(pjit_info_args, *args, **kwargs):
         "Mesh context manager should not be used with jit when backend or "
         "device is also specified as an argument to jit.")
 
+  axes_specs = _flat_axes_specs(abstracted_axes, *args, **kwargs)
+
   jit_name = 'jit' if resource_env is None else 'pjit'
   dbg = debug_info(jit_name, fun, args, kwargs, static_argnums, static_argnames)
   f = lu.wrap_init(fun)
@@ -448,10 +451,10 @@ def common_infer_params(pjit_info_args, *args, **kwargs):
   # leads to wrong expansion.
   if kwargs:
     f, dyn_kwargs = argnames_partial_except(f, static_argnames, kwargs)
-    args_flat, in_tree = tree_flatten((dyn_args, dyn_kwargs))
+    explicit_args, in_tree = tree_flatten((dyn_args, dyn_kwargs))
     flat_fun, out_tree = flatten_fun(f, in_tree)
   else:
-    args_flat, in_tree = tree_flatten(dyn_args)
+    explicit_args, in_tree = tree_flatten(dyn_args)
     flat_fun, out_tree = flatten_fun_nokwargs(f, in_tree)
     dyn_kwargs = ()
   del kwargs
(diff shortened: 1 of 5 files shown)


jax/_src/pjit.py
```diff
@@ -475,21 +478,31 @@ def common_infer_params(pjit_info_args, *args, **kwargs):
...
-        canonicalized_in_shardings_flat, args_flat)
+        canonicalized_in_shardings_flat, explicit_args)
 
+  if config.jax_dynamic_shapes:
+    implicit_args = _extract_implicit_args(in_type, explicit_args)
+  else:
+    implicit_args = []
+  args_flat = [*implicit_args, *explicit_args]
   assert len(args_flat) == len(canonicalized_in_shardings_flat)
+  donated_invars = (False,) * len(implicit_args) + donated_invars
```
yashk2810, line 505: I wonder if this can be merged below on line 509?

Also if implicit_args are present, you need to handle canonicalized_in_shardings_flat too? Which can also be merged on line 507?

donated_invars = (False,) * (len(implicit_args) + len(consts)) + donated_invars

jax/_src/pjit.py
```diff
@@ -844,36 +903,36 @@ def _create_pjit_jaxpr(fun, global_in_avals, debug_info, out_paths):
...
       isinstance(orig_out_shardings, XLACompatibleSharding)):
-    out_shardings_flat = (orig_out_shardings,) * len(global_out_avals)
+    out_shardings_flat = (orig_out_shardings,) * len(out_type)
   else:
     out_shardings_flat = flatten_axis_resources(
         "pjit out_shardings", out_tree(), orig_out_shardings,
         tupled_args=False)
 
-  pjit_check_aval_sharding(out_shardings_flat, global_out_avals, "pjit outputs",
+  pjit_check_aval_sharding(out_shardings_flat, out_type, "pjit outputs",
```
yashk2810, line 920: Should this be guarded with jax.config.dynamic_shapes too as you are guarding in _process_in_axis_resources?

jax/_src/pjit.py
```diff
@@ -1341,10 +1400,43 @@ def pjit_staging_rule(trace, *args, **params):
       all(_is_unspecified(o) for o in params["out_shardings"])):
     jaxpr = params['jaxpr']
     return core.eval_jaxpr(jaxpr.jaxpr, jaxpr.consts, *args)
-  else:
+  elif not config.jax_dynamic_shapes:
     return trace.default_process_primitive(pjit_p, args, params)
+  else:
+    source_info = source_info_util.current()
```
yashk2810, line 1406: nit: add `assert config.jax_dynamic_shapes` here?

yashk2810 (approved):

jax/_src/pjit.py
mattjj: Sorry, what do you mean by "this" here?

yashk2810: Ohh sorry.. I meant `pjit_check_aval_sharding` line.

Should it be guarded with `if not config.jax_dynamic_shapes`?

jax/_src/pjit.py
mattjj: Hrm, it seems a bit redundant given the `elif` branch above is `elif not config.jax_dynamic_shapes`. That is, the structure is
```python
if ...:
  ...
elif not X:
  ...
else:
  # you want to put an `assert X` here?
  ...
```

If we didn't have the first `if` branch it'd basically be:

```python
if not X:
 ...
else:
  # put an `assert X` here?
  ...
```

Is the assertion really needed?

jax/_src/pjit.py
mattjj: (`out_type` is just the new variable name for what was previously called `global_out_avals`)

jax/_src/pjit.py
yashk2810: Not needed.. just for readability purpose. Up to you! (I don't have strong feelings)

(I missed the `elif not config.jax_dynamic_shapes`  thing on the first read)

mattjj: Maybe I'll just switch the latter two branches. I was "putting the complicated thing at the end" but maybe it's less clear, as you say.

jax/_src/pjit.py
yashk2810: Ohh I see! So out_type won't contain extra stuff like in_type does when dynamic_shapes is on?

jax/_src/pjit.py
mattjj: Done! It's more readable now, on the important axis of being able to easily evaluate "this is only hit with dynamic shapes". Good idea.

jax/_src/pjit.py
yashk2810: Ohh nvm.. I read the _pjit_jaxpr code again. 

Then why rename this? Why not just call it global_out_avals? (given that it is avals)

The out_type confused me since I associated it with in_type

mattjj: Ah yes, thank you!

jax/_src/pjit.py
mattjj: Done!

PR #15205, opened 2023-03-24
Editable jaxlib build

yhtang:
This PR adds an `--editable` option to the `build/build.py` script. When set, the build script will
1. Upon compilation completion, copy the jaxlib source folder and the compiled `.so` files from the Bazel sandbox to the output path (`dist`  by default)
2. print a message prompting users to use `pip install -e` to install jaxlib in-place at the output path.

This way, we only need to run `pip install -e` once, and subsequent Bazel builds with `--editable ` will automatically update the installed jaxlib package (e.g., in a virtual environment).

Things worth consideration:
- Would it be better to copy the compiled `.so` files directly back into the jaxlib source folder and run `pip install -e` there? This way, rebuilding is not needed if changes only apply to Python files.

This PR is adapted from the Alpa project as part of a rebasing effort, with contributions from @merrymercy, @zhisbug, @ZYHowell, @zhuohan123 among many others.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -446,6 +446,10 @@ def main():
       default=None,
       help="CPU platform to target. Default is the same as the host machine. "
            "Currently supported values are 'darwin_arm64' and 'darwin_x86_64'.")
+  parser.add_argument(
+      "--editable",
+      action="store_true",
+      help="Create an 'editable' jaxlib build instead of a wheel.")
   add_boolean_argument(
       parser,
       "configure_only",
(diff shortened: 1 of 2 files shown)


yhtang:
@mjsML @nouiz for viz.

yhtang:
Updated commit message's co-author list after communicating with the original author @merrymercy

hawkinsp (approved):

hawkinsp:
> Things worth consideration:
> 
> * Would it be better to copy the compiled `.so` files directly back into the jaxlib source folder and run `pip install -e` there? This way, rebuilding is not needed if changes only apply to Python files.

Perhaps, but I think not. In general it's nice to keep built artifacts out of the source tree.

hawkinsp:
> Things worth consideration:
> 
> * Would it be better to copy the compiled `.so` files directly back into the jaxlib source folder and run `pip install -e` there? This way, rebuilding is not needed if changes only apply to Python files.

Perhaps, but I think not. In general it's nice to keep built artifacts out of the source tree.

PR #15211, opened 2023-03-24
hilbert transform similar to that in `osp.signal`

joglekara:
hilbert transform similar to that in `osp.signal` 

Carryover + cleanup from https://github.com/google/jax/pull/15121

### DIFF
--- a/jax/_src/scipy/signal.py
+++ b/jax/_src/scipy/signal.py
@@ -637,3 +637,32 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
 
   time = jnp.arange(x.shape[0], dtype=np.finfo(x.dtype).dtype) / fs
   return time, x
+
+@_wraps(osp_signal.hilbert)
+def hilbert(x: Array, N: Optional[int] = None, axis: int = -1):
+  check_arraylike('hilbert', x)
+  x = jnp.asarray(x)
+  if x.ndim > 1:
+    raise NotImplementedError("x must be 1D.")
+    # ind = [np.newaxis] * x.ndim
+    # ind[axis] = slice(None)
+    # h = h[tuple(ind)]
+  if jnp.iscomplexobj(x):
+    raise ValueError("x must be real.")
+  if N is None:
+    N = x.shape[axis]
+  if N <= 0:
+    raise ValueError("N must be positive.")
+
+  Xf = jnp.fft.fft(x, N, axis=axis)
+  if N % 2 == 0:
+    h = jnp.zeros(N, Xf.dtype).at[0].set(1).at[1:N // 2].set(2).at[N // 2].set(1)
+    # h[0] = h[N // 2] = 1
+    # h[1:N // 2] = 2
+  else:
+    h = jnp.zeros(N, Xf.dtype).at[0].set(1).at[1:(N+1) // 2].set(2)
+    # h[0] = 1
+    # h[1:(N + 1) // 2] = 2
+
+  x = jnp.fft.ifft(Xf * h, axis=axis)
+  return x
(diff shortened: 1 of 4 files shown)


jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,32 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
...
+
+  Xf = jnp.fft.fft(x, N, axis=axis)
+  if N % 2 == 0:
+    h = jnp.zeros(N, Xf.dtype).at[0].set(1).at[1:N // 2].set(2).at[N // 2].set(1)
+    # h[0] = h[N // 2] = 1
+    # h[1:N // 2] = 2
+  else:
+    h = jnp.zeros(N, Xf.dtype).at[0].set(1).at[1:(N+1) // 2].set(2)
+    # h[0] = 1
+    # h[1:(N + 1) // 2] = 2
```
jakevdp, line 665: Here and above, can we remove these comments? It seems like they don't add much.

jax/scipy/signal.py
```diff
@@ -25,4 +25,5 @@
   istft as istft,
   stft as stft,
   welch as welch,
+  hilbert as hilbert,
```
jakevdp, line 28: alphabetize

tests/scipy_signal_test.py
```diff
@@ -56,7 +56,7 @@
     ((65, 24), 24, 7, -2, -1),
 ]
 
-
+real_dtypes = jtu.dtypes.floating + jtu.dtypes.integer + jtu.dtypes.boolean
```
jakevdp, line 59: nit: "real_dtypes" seems like a misleading name for a set that includes ints and bools.

Also, should we test complex types too?

jakevdp (changes requested):

tests/scipy_signal_test.py
joglekara: I just grabbed this from the `fft_tests`

we could test complex dtypes but the scipy (and this) routine raises an error for complex dtypes. see line 651. 

Should we try to assert the `ValueError`. If so, can you please share a code example of how we can catch that successfully?

jakevdp: Oh, i didn’t notice that complex isn’t supported

joglekara:
Let me know if I missed something!

jax/_src/scipy/signal.py
```diff
@@ -637,3 +637,25 @@ def istft(Zxx: Array, fs: ArrayLike = 1.0, window: str = 'hann',
 
   time = jnp.arange(x.shape[0], dtype=np.finfo(x.dtype).dtype) / fs
   return time, x
+
+@_wraps(osp_signal.hilbert)
+def hilbert(x: Array, N: Optional[int] = None, axis: int = -1):
+  check_arraylike('hilbert', x)
+  x = jnp.asarray(x)
+  if x.ndim > 1:
```
jakevdp, line 645: Should this be `if x.ndim != 1`? Do you intend to support scalar inputs?

jax/scipy/signal.py
```diff
@@ -20,9 +20,10 @@
   convolve2d as convolve2d,
   correlate as correlate,
   correlate2d as correlate2d,
-  detrend as detrend,
   csd as csd,
+  detrend as detrend,
   istft as istft,
+  hilbert as hilbert,
```
jakevdp, line 26: alphabetize please

jakevdp (changes requested):
Looks great! Just two very minor comments.

Also, once you've fixed these last issues, please squash your changes into a single commit (see https://jax.readthedocs.io/en/latest/contributing.html#single-change-commits-and-pull-requests) Let me know if you need assistance with that last step.

Thanks!

jakevdp (approved):
Marking "Approve" in order to trigger the extra tests.

PR #15218, opened 2023-03-25
Suppress mypy warnings about missing imports.

hawkinsp:
Many warnings are emitted if you simply run `mypy jax`; this PR suppresses them.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -25,3 +25,31 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-iree.*]
 ignore_missing_imports = True
+[mypy-rich.*]
+ignore_missing_imports = True
+[mypy-optax.*]
+ignore_missing_imports = True
+[mypy-flax.*]
+ignore_missing_imports = True
+[mypy-tensorflow.*]
+ignore_missing_imports = True
+[mypy-tensorflowjs.*]
+ignore_missing_imports = True
+[mypy-tensorflow.io.*]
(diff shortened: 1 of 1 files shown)


jakevdp:
I'm confused about why we would need suppression for packages that the JAX source never references. Perhaps it's something specific to your local environment? If you run mypy via `pre-commit run mypy --all-files` it will run mypy in an isolated environment, with the same command as our CI. Perhaps that would be a better solution?

hawkinsp:
But the JAX source *does* reference all of these packages. I don't like to use precommit because it's very slow.

jakevdp:
OK – I didn't realize jax2tf had so many extra dependencies.

Regarding slowness: mypy does a lot of cacheing, so the first time you run it (either locally or via pre-commit) it will be slow. Subsequent times that you run it in the same manner should be faster.

Do you understand why the pre-commit mypy command is not seeing these warnings? What's different about your invocation from the invocation in CI where these warnings are not seen?

hawkinsp:
Yes, the reason is that `precommit` runs with `--ignore-missing-imports` by default.

https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit

jakevdp:
Ok, thanks. That makes sense. I still think we should generally encourage running myoy via pre-submit because things tend to be very brittle across myoy versions, but I don’t see any harm in adding these warning suppressions.

jakevdp (approved):

PR #15221, opened 2023-03-26
`custom_vjp` symbolic zeros support, take two

froystig:
This change re-introduces symbolic zero support for `custom_vjp` (originally #15088, which was reverted).

This time:

1. The forward rule API is slightly different, accepting two-field records at pytree leaves rather than pairs.
2. In the default setting where `symbolic_zeros` is not set, there are no new requirements from pytree node definitions that are involved in the primal arguments. This avoids any change in behavior on the default path. In particular, custom pytree node definitions that aren't completely polymorphic in their unflattening can remain as is.
3. There is an additional test involving a custom pytree node.

As an aside, we should maybe sort out and document the polymorphism requirements of pytree node definitions at some point.

### DIFF
--- a/jax/_src/interpreters/partial_eval.py
+++ b/jax/_src/interpreters/partial_eval.py
@@ -2041,16 +2041,18 @@ def _interleave_fun(every_others, *args, **kwargs):
 
 def _memoize(fn):
   cells = {}
-  saved_state = [core.thread_local_state.trace_state.copy()]
+  saved_state = core.thread_local_state.trace_state.copy()
+  sentinel = object()
   def memoized(*args):
-    if args not in cells:
+    out = cells.get(args, sentinel)
+    if out is sentinel:
       prev_state = core.thread_local_state.trace_state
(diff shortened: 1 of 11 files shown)


mattjj (approved):
Beautiful!

jax/_src/interpreters/partial_eval.py
```diff
@@ -1971,18 +1981,23 @@ def process_custom_transpose(self, prim, call, tracers, *,
...
+
+def _memoize(fn):
+  cells = {}
   saved_state = [core.thread_local_state.trace_state.copy()]
-  def memoized():
-    if not cell:
+  def memoized(*args):
+    if args not in cells:
       prev_state = core.thread_local_state.trace_state
       core.thread_local_state.trace_state = saved_state.pop()
```
mattjj, line 1995: Just noticed: I think we can't pop here if we might reuse the same saved_state for different values of `args`.

froystig: Great catch. I was just scratching my head over this!

PR #15228, opened 2023-03-27
Fix hessian llnk

canyon289:
Current link on readme doesn't go to hessian doc any longer. Changing to direct link


jakevdp (approved):
Looks good, thanks!

PR #15229, opened 2023-03-27
Delete duplicate installation information

canyon289:
The install doc has no extra information that isn't present on the front page. Removing to reduce duplication


jakevdp:
Thanks - I think I'd prefer to keep this placeholder, mainly because it leads to there being an "installation" header in the left sidebar for people who are looking for it that way. We've also recently discussed putting more information on this page, rather than directing people to the README. What do you think?

PR #15232, opened 2023-03-27
[shape_poly] Extend the handling of jnp.arange with shape polymorphism.

gnecula:
Previously, only `arange(stop, dtype=...)` was being handled in presence of shape polymorphism. Here we extend to add support for `start` and `step` to be also present. There are still plenty of restrictions:

   * no floating point constants are allowed among start, stop and step
   * we must resolve statically if step is positive or negative
   * we must resolve statically if the distance between start and stop is negative or positive.


jakevdp:
Are the checks on lines 2223-2224 problematic here? They're asserting that `stop` and `step` are static before getting to the new code.

gnecula:
> Are the checks on lines 2223-2224 problematic here? They're asserting that `stop` and `step` are static before getting to the new code.

The `concrete_or_error` will check if the `stop` and `step` are `Tracer`. Here they are not, they are this `_DimExpr` object that duck types integers.

jakevdp (approved):

PR #15247, opened 2023-03-27
Use ml_dtypes definition for jnp.finfo

jakevdp:


.pre-commit-config.yaml
```diff
@@ -26,7 +26,7 @@ repos:
   - id: mypy
     files: (jax/|tests/typing_test\.py)
     exclude: jax/_src/basearray.py  # Use pyi instead
-    additional_dependencies: [types-requests==2.28.11, jaxlib==0.4.6, ml_dtypes==0.0.3]
+    additional_dependencies: [types-requests==2.28.11, jaxlib==0.4.6]
```
hawkinsp, line 29: Can we increment rather than removing this dependency? We actually want types.

hawkinsp (approved):

.pre-commit-config.yaml
jakevdp: I didn't understand why it was listed here, because I believe it's installed by virtue of being a `jax` dependency. Is that not the case?

jakevdp: Oh, I see my understanding was wrong. I'll add it back.

jakevdp: I suppose we should also list `numpy` and `scipy` as well – what do you think?

PR #15258, opened 2023-03-28
[shape_poly] Refactor the computation of the dimension variables in native serialization

gnecula:
Currently, JAX native serialization produces a module whose main function takes additional arguments for the values of the dimension variables. These values are then resolved in the XlaCallModule based on a dim_args_spec parameter.

We move the code that computes the dimension variables from XlaCallModule to jax_export following pretty much the same technique. This simplifies XlaCallModule and especially its API (the dim_args_spec). So far this is just a refactoring with no semantic changes, but this will allow us to improve the support for dimension variables that occur in linear polynomials, e.g., "2*b" rather than just "b".

### DIFF
--- a/jax/experimental/jax2tf/jax_export.py
+++ b/jax/experimental/jax2tf/jax_export.py
@@ -16,13 +16,13 @@
 This module is used with jax2tf, but should have no TensorFlow dependencies.
 """
 import dataclasses
+from functools import partial
 import re
 from typing import  Callable, Dict, List, Optional, Sequence, Set, Union
 
 from absl import logging
 
 import jax
-from jax import config
 from jax import sharding
 
 from jax._src import core
@@ -32,13 +32,16 @@ from jax._src.interpreters import mlir
 from jax._src.interpreters import pxla
 from jax._src.lib import xla_client
 from jax._src.lib.mlir.dialects import stablehlo
+from jax._src.lib.mlir import ir
+from jax._src.lib.mlir.dialects import hlo
+from jax._src.lib.mlir.dialects import func as func_dialect
 
 
 map = util.safe_map
 zip = util.safe_zip
 
 # These are the JAX custom call target names that are guaranteed to be stable.
-# They are tested by back_compat_test.py.
+# Their backwards compatibility is tested by back_compat_test.py.
 _CUSTOM_CALL_TARGETS_GUARANTEED_STABLE = [
     "Sharding", "SPMDFullToShardShape", "SPMDShardToFullShape",
     "ducc_fft", "cu_threefry2x32",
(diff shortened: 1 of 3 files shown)


gnecula:
@burmako @GleasonK please take a look

jax/experimental/jax2tf/jax_export.py
```diff
@@ -215,9 +169,147 @@ def serialize_native(fun_jax: Callable,
...
+  """Wraps the lowered module with a new "main" that computes the dim vars.
+
+  JAX lowering in presence of shape polymorphism produces a `module` that
+  takes one or more dimension arguments, of type i32 or i64, followed
+  by the regular array arguments. The dimension arguments correspond to the
+  dimension variables appearing in the `args_avals`, in sorted order.
+
+  Consider the lowering of a function with one array argument of type "f32[w, h]",
+  where "w" and "h" are two dimension variables. The `module` will also
+  contain two dimension arguments, corresponding to "h" and "w" respectively::
```
GleasonK, line 186: nit, extra `:` in `::`?

jax/experimental/jax2tf/jax_export.py
```diff
@@ -215,9 +169,147 @@ def serialize_native(fun_jax: Callable,
...
-  )
+      xla_call_module_version=xla_call_module_version)
+
+
+def compute_dim_vars(module: mlir.ir.Module,
+                     args_avals: Sequence[core.ShapedArray]) -> mlir.ir.Module:
+  """Wraps the lowered module with a new "main" that computes the dim vars.
+
+  JAX lowering in presence of shape polymorphism produces a `module` that
+  takes one or more dimension arguments, of type i32 or i64, followed
```
GleasonK, line 180: "that takes one or more dimension arguments, specified using 0-dimensional tensors of type i32 or i64"

https://github.com/GleasonK/stablehlo/blob/main/docs/spec.md#get_dimension_size

jax/experimental/jax2tf/jax_export.py
```diff
@@ -215,9 +169,147 @@ def serialize_native(fun_jax: Callable,
...
+
+  Returns the wrapped module.
+  """
+  dim_args_builders = get_dim_arg_builders(args_avals)
+
+  # Make a new module, do not mutate the "module" because it may be cached
+  context = mlir.make_ir_context()
+  with context, ir.Location.unknown(context):
+    # TODO: is there a better way to make a copy of the `module`?
+    new_module = ir.Module.parse(str(module))
```
GleasonK, line 215: > TODO: is there a better way to make a copy of the `module`?

I've seen the question of moving a module between Contexts's a few times lately - the consensus seems to be that re-parsing is the best option. Better to use bytecode instead of string though. I believe it would be: `ir.Module.parse(mlir.module_to_bytecode(module))`

GleasonK (commented):
LGTM.

Does this impact the Refine Shapes pass at all? I'm not too sure how the setup for that pass to be called works. @burmako

gnecula:
> LGTM.
> 
> Does this impact the Refine Shapes pass at all? I'm not too sure how the setup for that pass to be called works. @burmako

It does not, before the Refine Shapes the XlaCallModule will add a wrappper. But since dim_args_spec is empty there will be no wrapper added. Recently I have changed the XlaCallModule to run the shape refinement even if dim_args_spec is empty, expecting the change I am making now.

jax/experimental/jax2tf/jax_export.py
gnecula: Done

burmako (approved):

PR #15263, opened 2023-03-28
Add deprecation warnings for several top-level jax imports

jakevdp:


jax/__init__.py
```diff
@@ -174,6 +166,45 @@
 # Also circular dependency.
 from jax._src.array import Shard as Shard
 
+# TODO(jakevdp) update callers and remove these after June 2023
```
hawkinsp, line 169: Consider using the utility/pattern in deprecations.py for this?

Canonical example: https://github.com/google/jax/blob/main/jax/interpreters/pxla.py

One key benefit is then you can locally (e.g., in our internal copy of the JAX) accelerate the deprecations to prevent backsliding when all users we know about are gone.

jakevdp: Sounds good, thanks!

jakevdp: Done

hawkinsp (approved):

jax/__init__.py
```diff
@@ -175,6 +175,54 @@
...
+  "partial_eval": (
+    "jax.partial_eval is deprecated. Use jax.interpreters.partial_eval",
+    _deprecated_partial_eval,
+  ),
+  "pxla": (
+    "jax.pxla is deprecated. Use jax.interpreters.pxla",
+    _deprecated_pxla,
+  ),
+  "xla": (
+    "jax.ad is deprecated. Use jax.interpreters.xla",
```
yashk2810, line 197: drive-by

nit: should this be jax.xla?

jakevdp: Fixed, thanks

PR #15270, opened 2023-03-28
Turn on PJRT C API by default.

skye:
I forgot that the default setting is actually in jaxlib: https://github.com/openxla/xla/blob/fbe9a80fdb8c429e8a175962459da348cd560a50/xla/python/xla_client.py#L135

To be able to make this change as a jax-only release, I manually set the env var on Cloud TPU if it isn't already set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,25 @@ Remember to align the itemized text with the first line of an item within a list
 
 ## jax 0.4.8
 
+* Breaking changes
+  * A major component of the Cloud TPU runtime has been upgraded. This enables
+    the following new features on Cloud TPU:
+    * {func}`jax.debug.print`, {func}`jax.debug.callback`, and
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,25 @@ Remember to align the itemized text with the first line of an item within a list
 
 ## jax 0.4.8
 
+* Breaking changes
+  * A major component of the Cloud TPU runtime has been upgraded. This enables
+    the following new features on Cloud TPU:
+    * `jax.debug.print`, `jax.debug.callback`, and `jax.debug.breakpoint()` now
```
jakevdp, line 14: Optional: use e.g. ``` {func}`jax.debug.print` ``` so the generated HTML will include links to the function documentation

jakevdp (approved):
Whooo! 🎉

CHANGELOG.md
skye: Good idea, done!

PR #15274, opened 2023-03-28
Add minimal pyproject.toml specifying build system

### LINKED ISSUE #15256: pyproject.toml based wheel builds
To support modern packaging standards (aka [PEP 518](https://peps.python.org/pep-0518/)) for wheel building, specify the build requirements and backend in [pyproject.toml](https://packaging.python.org/en/latest/glossary/#term-pyproject.toml).

I've tested following the instructions in the [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use) (adding the `build-system` table), and installation from sdist works perfectly.

This will fix installation when there are no wheels provided and [setuptools](https://pypi.org/project/setuptools/) is not yet installed.

jakevdp:
Fixes #15256

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"


hawkinsp (approved):

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools"]
```
eli-schwartz, line 2: This should probably also require `wheel`. :)

PR #4879, opened 2020-11-11
TST: add explicit tests of PRNGKey seed handling.

jakevdp:
This is a set of regression tests for #4877 which highlight the current behavior. I plan to merge these prior to #4877 in order to make it absolutely clear how user-visible behavior is changed by that PR.

TODOs in this PR will be resolved by the changes in #4877.

PR #15275, opened 2023-03-28
add overflow errors for numpy.ndarray (from canonicalize_dtype)

mattjj:
Before this change, doing things like `jax.jit(lambda x: x + 0)(numpy.array([2**32]))` would silently truncate the input values. Now we raise an error.

in jax:
* add helper function dtypes.check_ndarray_for_int_overflow
* call it in mlir.py when we canonicalize constants (though deleting this line doesn't cause any tests to fail)
* call it in numpy.ndarray's shard_arg handler in pxla.py
* call it in _canonicalize_ndarray_dtype in xla.py
* call it in lax_numpy.asarray

in the tests and test utils:
* add a new test for numpy.ndarray overflows, and add more cases to Python builtin int overflow tests
* in lax_vmap_test.py, changed a test to canonicalize dtype earlier (because now the old tests raise an OverflowException)
* unrelated, change JIT_IMPLEMENTATION in test_util.py so as not to mutate jax.jit

### DIFF
--- a/jax/_src/dtypes.py
+++ b/jax/_src/dtypes.py
@@ -685,3 +685,13 @@ def check_user_dtype_supported(dtype, fun_name=None):
     fun_name = f"requested in {fun_name}" if fun_name else ""
     truncated_dtype = canonicalize_dtype(dtype).name
     warnings.warn(msg.format(dtype, fun_name , truncated_dtype), stacklevel=3)
+
+def check_ndarray_for_int_overflow(x: np.ndarray):
+  dtype = canonicalize_dtype(x.dtype)
+  if np.issubdtype(dtype, np.integer) and (
+      (x < np.iinfo(dtype).min).any() or (x > np.iinfo(dtype).max).any()):
+    raise OverflowError(
+        f"Some values in a NumPy array of dtype {x.dtype} are too large in "
+        f"magnitude to convert to dtype {dtype}. Try setting "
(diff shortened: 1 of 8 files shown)


mattjj:
It fails on tests from #4879 which apparently involved an overflow, which was silently tolerated. @jakevdp was that test behavior intentional?

jakevdp:
I think I wrote that test when we found that this kind of overflow behavior differed between jit and non-jit code. Since this PR is changing the overflow behavior, the fix would be to make sure it errors both with and without JIT.

jax/_src/dtypes.py
```diff
@@ -685,3 +685,13 @@ def check_user_dtype_supported(dtype, fun_name=None):
     fun_name = f"requested in {fun_name}" if fun_name else ""
     truncated_dtype = canonicalize_dtype(dtype).name
     warnings.warn(msg.format(dtype, fun_name , truncated_dtype), stacklevel=3)
+
+def check_ndarray_for_int_overflow(x):
```
jakevdp, line 689: I'd annotate this `x : np.ndarray` just to make clear that this is meant for numpy array inputs.

jakevdp (approved):

jax/_src/dtypes.py
mattjj: Great idea.

PR #15280, opened 2023-03-29
Use modern packaging standards for setuptools

eli-schwartz:
In setuptools 30.3.0 (December 2016), support was added for specifying setup options in setup.cfg as opposed to arguments to the setup function. Options are merged, and setup.py takes priority when disagreements occur.

setup.cfg arguments are superior for a couple reasons. They are:
- declarative -- no python code is executed to evaluate them
- easier to read
- capable of doing things that you'd need to open-code in python

Of particular note, it is possible to load files or statically parse the AST for a simple assignment, via values such as `attr:` or `file:`. This is a cfg-exclusive feature.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,35 @@
+[metadata]
+name = jax
+version = attr: jax.version.__version__
+description = Differentiate, compile, and transform Numpy code.
+author = JAX team
+author_email = jax-dev@google.com
+url = https://github.com/google/jax
+license = Apache-2.0
+classifiers =
+  Programming Language :: Python :: 3.8
+  Programming Language :: Python :: 3.9
+  Programming Language :: Python :: 3.10
+  Programming Language :: Python :: 3.11
+long_description = file: README.md
+long_description_content_type = text/markdown
+
+[options]
+install_requires =
+  ml_dtypes>=0.0.3
+  numpy>=1.21
+  opt_einsum
+  scipy>=1.7
+packages = find:
+python_requires = >= 3.8
+zip_safe = False
+
+[options.package_data]
+jax = py.typed, *.pyi, **/*.pyi
+
+[options.packages.find]
+exclude = examples, jax/src/internal_test_util
+
 [flake8]
 max-line-length = 88
 ignore =
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ import subprocess
 import os
 import sys
 
-from setuptools import setup, find_packages
+from setuptools import setup
 
 _current_jaxlib_version = '0.4.7'
 # The following should be updated with each new jaxlib release.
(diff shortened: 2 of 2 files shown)


eli-schwartz:
@jakevdp

from https://github.com/google/jax/issues/15256#issuecomment-1487730643 

> but will be somewhat painful. Our current setup process makes use of a lot of the imperative computation flexibility of `setup.py` files

Here's an example that shows you can have the best of both worlds. :)

EpicWink:
[pyproject.toml](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/) is the community standard for specifying project metadata declaritively: I would recommend using that over [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), which will [eventually be deprecated](https://github.com/pypa/setuptools/issues/3214) and removed.

In this case, you would have to add `"optional-dependencies"` to [`dynamic`](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic). All other features are supported in pyproject.toml, either in the `project` table or [`tool.setuptools`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration). `url` will become [`project.urls.homepage`](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls).

eli-schwartz:
> or [`tool.setuptools`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration).

... which is still in beta stage, unlike something that has worked robustly and stably for half a decade plus.

> over [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), which will [eventually be deprecated](https://github.com/pypa/setuptools/issues/3214) and removed.

This will almost certainly never ever ever happen, because it would cause large chunks of PyPI to become unbuildable without any hope of vanished maintainers ever fixing things. Additionally, I seem to recall the proposed "deprecation" method was actually "internally run a function that converts cfg to toml, and then run another function to load the toml", which is a bit of indirection that really boils down to "setup.cfg is still supported".

EpicWink:
`find_packages` is now unused in setup.py: I suggest removing the import

---

> > or [tool.setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration).
>
> ... which is still in beta stage

The docs aren't particularly clear, but it seems specifically automatic discovery, package data, and dynamic dependencies (from file) are in beta. I suppose its fine to wait for it to be marked as stable.

---

> > over [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), which will https://github.com/pypa/setuptools/issues/3214 and removed.
>
> This will almost certainly never ever ever happen, because it would cause large chunks of PyPI to become unbuildable without any hope of vanished maintainers ever fixing things.

Thus the deprecation. As a related example, `distutils` is in the final stages of being removed, despite it being used in ~100% of distributed projects 10 or 15 years ago.

eli-schwartz:
> `find_packages` is now unused in setup.py: I suggest removing the import

Good point. While I'm at it, distutils.spawn.find_executable() is a truly old way of performing the role of shutil.which... speaking of deprecated things.

PR #15177, opened 2023-03-23
jax.random: clean up handling of dtypes

jakevdp:
This change regularizes the handling of the `dtype` argument to `jax.random` distributions. The main changes:

1. Previously, the default behavior clobbered the dtype of the inputs. Now, if `dtype` is not specified, it will be derived from input parameters. For example, previously this was the behavior:
```python
>>> random.uniform(random.PRNGKey(0), minval=jnp.bfloat16(0), maxval=jnp.bfloat16(1))
Array(0.41845703, dtype=float32)
```
Now this is the behavior:
```python
>>> random.uniform(random.PRNGKey(0), minval=jnp.bfloat16(0), maxval=jnp.bfloat16(1))
Array(0, dtype=bfloat16)
```

2. Previously there was no warning in case of the specified dtype being truncated. For example, previously this was the behavior (with `jax_enable_x64=False`):
```python
>>> random.uniform(random.PRNGKey(0), dtype='float64')
Array(0.41845703, dtype=float32)
```
Now this is the behavior:
```python
>>> random.uniform(random.PRNGKey(0), dtype='float64')
UserWarning: Explicitly requested dtype float64 requested in uniform is not available, and will be truncated to dtype float32.
To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable.
 See https://github.com/google/jax#current-gotchas for more.

Array(0.41845703, dtype=float32)
```

PR #15282, opened 2023-03-29
add `geometric` to `random.py`

JiaYaobo:
A counterpart for  `numpy.random.geometric`

### DIFF
--- a/jax/_src/random.py
+++ b/jax/_src/random.py
@@ -2098,3 +2098,53 @@ def _wald(key, mean, shape, dtype) -> Array:
   x = mean + mean_sq * y / 2 - mean / 2 * sqrt_term
   w = lax.select(lax.le(z,  mean / (mean + x)), x, mean_sq / x)
   return w
+
+def geometric(key: KeyArray,
+              p: RealArray,
+              shape: Optional[Shape] = None,
+              dtype: DTypeLikeInt = dtypes.int_) -> Array:
+  r"""Sample Geometric random values with given shape and float dtype.
+
+  The values are returned according to the probability mass function:
+
+  .. math::
+      f(k;p) = p(1-p)^{k-1}
+
+  on the domain :math:`0 < p < 1`.
+
+  Args:
+    key: a PRNG key used as the random key.
+    p: a float or array of floats broadcast-compatible with ``shape``
+      representing the the probability of success of an individual trial.
+    shape: optional, a tuple of nonnegative integers specifying the result
+      shape. Must be broadcast-compatible with ``p``. The default
+      (None) produces a result shape equal to ``np.shape(p)``.
+    dtype: optional, a int dtype for the returned values (default int64 if
+      jax_enable_x64 is true, otherwise int32).
+
+  Returns:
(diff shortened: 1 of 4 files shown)


jax/_src/random.py
```diff
@@ -2098,3 +2098,52 @@ def _wald(key, mean, shape, dtype) -> Array:
...
+              shape: Optional[Shape] = None,
+              dtype: DTypeLikeInt = dtypes.int_) -> Array:
+  r"""Sample Geometric random values with given shape and float dtype.
+
+  The values are returned according to the probability mass function:
+
+  .. math::
+      f(k;p) = p(1-p)^{k-1}
+
+  on the domain :math:`0 < p < 1` is the probability of success of an individual trial
```
jakevdp, line 2113: Delete "is the probability of success of an individual trial" ?

jax/_src/random.py
```diff
@@ -2098,3 +2098,52 @@ def _wald(key, mean, shape, dtype) -> Array:
...
+    shape = core.canonicalize_shape(shape)
+  return _geometric(key, p, shape, dtype)
+
+@partial(jit, static_argnums=(2, 3), inline=True)
+def _geometric(key, p, shape, dtype) -> Array:
+  if shape is None:
+    shape =  np.shape(p)
+  else:
+    _check_shape("geometric", shape, np.shape(p))
+  p = lax.convert_element_type(p, np.float32)
```
jakevdp, line 2144: Instead of hard-coding float32, we should use
```python
check_arraylike("geometric", p)
p, = promote_dtypes_inexact(p)
```

jakevdp (changes requested):
Thanks!

jakevdp (approved):
Clicking "Approve" to trigger additional tests.

jax/_src/random.py
JiaYaobo: good advice! Please take a look at poisson related functions in random.py may use the hard-coding too

jakevdp (approved):

jax/_src/random.py
```diff
@@ -2098,3 +2098,53 @@ def _wald(key, mean, shape, dtype) -> Array:
...
+  return _geometric(key, p, shape, dtype)
+
+@partial(jit, static_argnums=(2, 3), inline=True)
+def _geometric(key, p, shape, dtype) -> Array:
+  if shape is None:
+    shape =  np.shape(p)
+  else:
+    _check_shape("geometric", shape, np.shape(p))
+  check_arraylike("geometric", p)
+  p = promote_dtypes_inexact(p)
```
jakevdp, line 2145: THis needs to be
```
p, = promote_dtypes_inexact(p)
```

jax/_src/random.py
jakevdp: Yeah I'm in the process of changing all of those, see https://github.com/google/jax/pull/15177. It's just taking a while to land because it breaks some corner cases.

jakevdp:
Looks like there are some precision issues in the tests

jakevdp (approved):
Thanks!

PR #15287, opened 2023-03-29
[shape_poly] Improved computation of dimension variables for native serialization

gnecula:
Previously for native serialization we could only support polymorphic_shapes where the specification was a simple dimension variable. E.g., we could not handle a specification where `polymorphic_shapes="2*b"` because there was no way to recover the value of `b` from the actual shape. (For non-native serialization we were supporting some limited equation solving.)

The above is important, e.g., for the gradient of functions like `jnp.concatenate([x, x])`, where the output shape if `2 *b`.

This is possible because in #15258 we have brought the computation of the dimension variables into jax_export.

What we do here is to even out the support for native serialization to have the same power as the non-native one. We do this by reusing the same `shape_poly.prepare_dim_var_env` that we use for non-native serialization.

After we land this, we will refactor the shape environment to be cleaner.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -399,8 +399,11 @@ def convert(fun_jax: Callable,
               native_serialization_strict_checks)
           return outs_tf, out_avals
       else:
+        def get_dimension_size(args, arg_idx: int, dim_idx: int) -> shape_poly.DimExprValue:
+          return shape_poly.dimension_size_p.bind(args[arg_idx], dimension=dim_idx)
         dim_vars, get_dim_values_jax = shape_poly.prepare_dim_var_env(
-            args_avals_flat)
+            args_avals_flat, get_dimension_size)
+
         dim_values, _ = _interpret_fun_jax(get_dim_values_jax, args_flat_tf,
                                            args_avals_flat, name_stack)
         shape_env = zip(dim_vars, dim_values)  # type: ignore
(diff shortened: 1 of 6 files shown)


gnecula:
@maxwillzq PTAL

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -399,8 +399,11 @@ def run_fun_flat_as_tf(
               native_serialization_strict_checks)
           return outs_tf, out_avals
       else:
+        def get_dimension_size(args, arg_idx: int, dim_idx: int) -> shape_poly.DimExprValue:
```
maxwillzq, line 402: So we only add this feature for non_serialization jax2tf ?

jax/experimental/jax2tf/tests/shape_poly_test.py
```diff
@@ -1007,16 +1009,25 @@ def f_jax(x):  # x: f32[w, h]
...
-  def test_non_trivial_polynomials_spec(self):
-    if config.jax_dynamic_shapes:
-      raise unittest.SkipTest("--jax_dynamic_shapes supports only trivial polynomials")
+  @parameterized.named_parameters(
+      dict(testcase_name=f"_{str(polymorphic_shapes)}",
+           polymorphic_shapes=polymorphic_shapes)
+      # The polymorphic_shapes should have three comma-separated DimExpr matching
+      # 16, 24, 32
+      for polymorphic_shapes in [
+          "b1+6,b1+14,b2",  # b1=10, b2=32
```
maxwillzq, line 1018: What happen if we use extreme value here. e.g "b1+17" so the b1=-1 ?  Does our code handle this exception correctly ?

maxwillzq (approved):

jax/experimental/jax2tf/tests/shape_poly_test.py
gnecula: Not yet, there is some code to check that all dimension variables are >= 1, but it only works in very few cases, e.g., in TF eager mode in non-native serialization. This is a known problem: 

jax/experimental/jax2tf/README.md#unchecked-assumption-that-the-dimension-variables-take-strictly-positive-values

This will be checked in native serialization, but I need a few more changes here and in XlaCallModule before we can do that.

PR #15302, opened 2023-03-29
improve pmap in_axes/out_axes pytree prefix error messages

mattjj:
Before:
image

After:
image

Additionally, this means `pmap` no longer relies on being able to unflatten user PyTree nodes with arbitrary `object()`s in them.

The key changes were:
1. use `tree_util.broadcast_prefix` in place of the older `flatten_axes`, and
2. instead of using the args pytreedef, use the actual args; instead of using the output pytreedef, use the actual output (grabbing it via a simple `lu.transformation_with_aux`).

### DIFF
--- a/jax/_src/api.py
+++ b/jax/_src/api.py
@@ -38,9 +38,10 @@ from contextlib import contextmanager, ExitStack
 import jax
 from jax._src import linear_util as lu
 from jax import stages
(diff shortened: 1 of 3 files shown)


yashk2810 (approved):
Woohoo!

mattjj:
I had an early draft of a vmap version of this in #14748, though it's bitrotted and not as nice as the approach here.

Issue #14921, opened 2023-03-10
Add streamlined `_ensure_arraylike` utility


jakevdp:
Over time we've landed on the following pattern for user-facing APIs in `jax.numpy`, `jax.scipy`, `jax.random`, etc.:
```python
import jax.numpy as jnp
from jax._src.numpy.util import _check_arraylike
from jax.typing import ArrayLike
from jax import Array

def func(x: ArrayLike, ...) -> Array:
  _check_arraylike("func", x)  # this is essentially a runtime check of isinstance(x, ArrayLike)
  x = jnp.asarray(x)  # this actually converts x to Array so that, e.g. it can be passed to lax functions
  # do some computation with x
  return out
```
There are a couple improvements we could make here:

1. We essentially never want to call `_check_arraylike` without following with `jnp.asarray` on the inputs. Whe should combine these into a single utility! This would prevent failures like the one reported in #14906.
2. `jnp.asarray` is a pretty heavy function; if we know that an input passes `_check_arraylike`, we can convert to an array much more efficiently.
3. This unified utility might also handle the `stackable` checks that are currently in numpy, via an extra keyword like `allow_stackable=True`. Also perhaps incorporate `check_no_float0s` via `allow_float0s=False`.
4. `_check_arraylike` lives in `jax._src.numpy.util`, but it is far more generally applicable. It should probably live in `jax._util`, or perhaps even be a public utility so that we can recommend this pattern for downstream APIs (e.g. here: https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices)

soraros:
@jakevdp Thanks for investigating this!

Wow, I didn't know the existence of this ([jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices](https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices)). It's such a nature place where all the array API related (`type` checking, `dtype` checking, `dtype` promotion, staticness, etc.) tools can go. Do you think we could open an umbrella issue to track the inception of such a set of tools?

jakevdp:
That doc is only a week or two old. #12049 is a relevant tracking issue

PR #15303, opened 2023-03-29
Add internal jax.lax.asarray utility

jakevdp:
`jnp.asarray` is a huge API, and we don't always need its full power when converting `ArrayLike` to `Array`.

Putting this in `lax` makes sense to avoid circular imports when breaking up the `lax_numpy.py` monolith.

### DIFF
--- a/jax/_src/lax/lax.py
+++ b/jax/_src/lax/lax.py
@@ -122,6 +122,15 @@ def _try_broadcast_shapes(
         return None
   return tuple(result_shape)
 
+def asarray(x: ArrayLike) -> Array:
+  """Lightweight conversion of ArrayLike input to Array output."""
+  if isinstance(x, Array):
+    return x
+  elif isinstance(x, np.ndarray) or np.isscalar(x):
+    return api.device_put(x)
+  else:
+    raise TypeError(f"asarray: expected ArrayLike, got {x} of type {type(x)}.")
+
 @overload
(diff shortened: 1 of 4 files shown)


soraros:
I suppose this is the canonical `asarray` to use after `check_arraylike` in the future?

jakevdp:
> I suppose this is the canonical `asarray` to use after `check_arraylike` in the future?

Yes, that's what I have in mind. We'll probably end up rolling the two together when addressing #14921

jax/_src/lax/lax.py
```diff
@@ -122,6 +122,15 @@ def _try_broadcast_shapes(
         return None
   return tuple(result_shape)
 
+def asarray(x: ArrayLike) -> Array:
+  """Lightweight conversion of ArrayLike input to Array output."""
+  if isinstance(x, Array):
+    return x
+  elif isinstance(x, np.ndarray) or np.isscalar(x):
+    return api.device_put(x)
```
pschuh, line 130: Should this work inside a trace context?

pschuh (approved):

jax/_src/lax/lax.py
jakevdp: In a traced context `isinstanc(x, Array)` will return True if `x` itself is a tracer. If `x` is a container/pytree of tracers, then it should hit the TypeError on line 131

PR #15335, opened 2023-03-31
add assertions for axis name shadowing bugs

mattjj:
Axis name shadowing in `vmap`-of-`vmap` and `vmap`-of-`pmap` is buggy, and has been for at least 2 years, and possibly for as long as we've had axis names in vmap. Damn!

The basic issue is that we look up the axis frame in the axis environment based on name, e.g. with `core.axis_frame(self.axis_name)` in methods on `BatchTrace`. But if axis names can shadow, that means computing a frame based only on `self.axis_name` could give the wrong one!

For example, if we had

```python
jax.vmap(jax.pmap(lambda x: 2 * x, axis_name='i'), axis_name='i')(jax.numpy.ones((2, 4)))
```

then in `BatchTrace.process_primitive` when we're about to call the `pjit` batching rule (since `2 * x` leads to a call to the jitted `jnp.multiply`) we get a `core.axis_frame(self.axis_name).size == 4` when we have `in_vals[0].shape[in_dims[0]] == 2`. That is, when applying `BatchTrace.process_primitive` for the (outer) `vmap`, we accidentally pick up the frame for the (inner) `pmap`. Indeed we always get the frame for the inner guy from `core.axis_frame(self.axis_name)`! The two maps have gotten confused!

An analogous thing happens with

```python
jax.vmap(jax.vmap(lambda x: 2 * x, axis_name='i'), axis_name='i')(jax.numpy.ones((2, 4)))
```

except now we're getting two `vmap`s confused with each other: again `core.axis_frame(self.axis_name)` always gives the frame for the inner `vmap`, even when we're running the outer `vmap`'s `BatchTrace.process_primitive`.

This isn't a problem for `pmap`-of-`vmap` or `pmap`-of-`pmap` because `pmap` isn't a transformation.

**Also, we only noticed this now after the jit-pjit merge because it may only be an issue for vmap-of-vmap-of-initialstyle. Actually, we noticed this as a shape error when testing vmap-of-shmap-of-initialstyle.**

This PR adds assertions as a _late_ check for whether the confusion has happened. I think a better fix would be to disallow shadowing entirely, by raising an error as soon as a shadowing axis name is attempted to be bound. Or maybe we can allow shadowing and improve our implementation. In any case, I'll send such fixes in follow-up PRs. For now I just want to rule out silently incorrect behavior.

### DIFF
--- a/jax/_src/interpreters/batching.py
+++ b/jax/_src/interpreters/batching.py
@@ -332,14 +332,19 @@ class BatchTrace(Trace):
         frame.size, frame.name, frame.main_trace.trace_type)
 
   def get_frame(self, vals, dims) -> core.AxisEnvFrame:
-    if self.axis_name is core.no_axis_name:
-      # If axis name is `no_axis_name` we can't find it via `core.axis_name` so
-      # we reconstruct it from the information we have available
+    if any(d is not not_mapped for d in dims):
       sizes = (x.shape[d] if type(d) is int else len(d.segment_lengths)
                for x, d in zip(vals, dims) if d is not not_mapped)
       axis_size, = core.dedup_referents(sizes)
+    else:
+      axis_size = None  # can't be inferred from data
+    if self.axis_name is core.no_axis_name:
+      assert axis_size is not None  # must be inferrable from data
       return core.AxisEnvFrame(self.axis_name, axis_size, self.main)
-    return core.axis_frame(self.axis_name)
+    frame = core.axis_frame(self.axis_name)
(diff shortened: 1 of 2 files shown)


sharadmv (approved):

mattjj:
I think I introduced this bug in [#4835](https://github.com/google/jax/pull/4835) by splitting out the `_main_trace_for_axis_names` function (and leaving behind a `frame = core.axis_frame(self.axis_name)`.

PR #15340, opened 2023-03-31
[shape_poly] Cleanup handling of dimension variables.

gnecula:
We unify the way we compute with dimension variables (computing their values from the shape of the actual arguments, and also using those values to evaluate shapes that contain dimension variables).

We remove DimExprValueMlir, and all computations with dimension variables and DimExpr are now done by JAX interpretation, followed by lowering to TF or StableHLO.


gnecula:
@maxwillzq PTAL

gnecula:
There should be no visible change of functionality due to this change

gnecula:
@rxsang PTAL

maxwillzq (commented):
LGTM

PR #15341, opened 2023-03-31
[jax2tf] Cleanup of handling of tf.custom_gradient

gnecula:
There are some incompatibilities between JAX and TF when it comes
to gradients for functions that take non-float arguments, or whose
arguments are unused. JAX uses float0 for those gradients, but this
is not a type that TF recognizes. Furthermore, under tf.function
context TF will pass `None` in place of cotangents for the outputs
that have non-float type.

Previously, the workarounds for these were in the JAX function
that was converted to obtain the TF gradient. Here we move
those workarounds in the TF-land of jax2tf.

This will enable us to expand jax_export with handling of gradients.
jax_export is pure JAX, and hence it is important to move the
TF workarounds outside of the converte JAX functions.

This is just a refactor.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -440,9 +440,9 @@ def convert(fun_jax: Callable,
                   _make_custom_gradient_fn_tf(
                       fun_flat_jax=fun_flat_jax,
                       args_flat_tf=args_flat_tf,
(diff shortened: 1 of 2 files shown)


gnecula:
@maxwillzq @rxsang PTAL

rxsang (approved):

maxwillzq (approved):

PR #15382, opened 2023-04-04
Add key path related guide & code to the documentation.

IvyZX:


IvyZX:
Friendly ping :)

docs/jax-101/05.1-pytrees.md
```diff
@@ -187,6 +187,22 @@ plt.scatter(xs, forward(params, xs), label='Model prediction')
 plt.legend();
 ```
 
+## Key paths
+
+In a nested pytree, each leaf has a unique "path of keys" corresponding to its place in the tree. JAX allows you to access the key paths and use them when handling a pytree. The corresponding APIs are:
```
mattjj, line 192: I think we don't guarantee the path is unique.

```suggestion
In a pytree each leaf has a _key path_. A key path for a leaf is a list of _keys_, where the length of the list is equal to the depth of the leaf in the pytree . Each _key_ is a [hashable object](https://docs.python.org/3/glossary.html#term-hashable) that represents an index into the corresponding pytree node type. The type of the key depends on the pytree node type; for example, the type of keys for `dict`s is different from the type of keys for `tuple`s.

For built-in pytree node types, the set of keys for any pytree node instance is unique. For a pytree comprising nodes with this property, the key path for each leaf is unique.

The APIs for working with key paths are:
```

docs/jax-101/05.1-pytrees.md
```diff
@@ -187,6 +187,22 @@ plt.scatter(xs, forward(params, xs), label='Model prediction')
...
+
+* `jax.tree_util.tree_flatten_with_path`
+* `jax.tree_util.tree_map_with_path`
```
mattjj, line 195: Should we add `keystr` here? We can also add an sentence that says: all we guarantee you can do with a general key path is apply `keystr` to it to get a string.

Also, can we add type annotations to these functions, if they're not too confusing? I think our doc had some.

docs/jax-101/05.1-pytrees.md
```diff
@@ -187,6 +187,22 @@ plt.scatter(xs, forward(params, xs), label='Model prediction')
...
+
+One use case is to print debugging information related to a certain leaf value:
+
+```{code-cell} ipython3
+tree = [1, {'k1': 2, 'k2': (3, 4)}, 5]
+flattened, _ = jax.tree_util.tree_flatten_with_path(tree)
+for key_path, value in flattened:
+    print(f'Value of tree{jax.tree_util.keystr(key_path)}: {value}')
+```
+
```
mattjj, line 205: Are `SequenceKey`, `DictKey`, etc in the public API? They're imported in `jax/tree_util.py`, but I forget what we decided.

If they're in the public API then we should document them. The conservative thing would be not to document them, but then should we remove them from `jax/tree_util.py`?

docs/jax-101/05.1-pytrees.md
```diff
@@ -240,9 +259,14 @@ To solve this, we need to register our container with JAX by telling it how to f
...
-def flatten_MyContainer(container) -> Tuple[Iterable[int], str]:
+def flatten_with_keys_MyContainer(container) -> Tuple[Iterable[int], str]:
   """Returns an iterable over container contents, and aux data."""
-  flat_contents = [container.a, container.b, container.c]
+  
+  # GetAttrKey is a common way to express an attribute key. Users are free
+  # to pick any other expression that fits their use cases the best.
+  flat_contents = [(jax.tree_util.GetAttrKey('a'), container.a),
+                   (jax.tree_util.GetAttrKey('b'), container.b),
+                   (jax.tree_util.GetAttrKey('c'), container.c)]
```
mattjj, line 269: WDYT about keeping the non-key API example (i.e. the existing example using `register_pytree-node`), and then appending the example which shows how to use `register_pytree_with_keys`? We don't want to un-document a lot of existing code. Moreover, this extended API is optional.

docs/jax-101/05.1-pytrees.md
```diff
@@ -255,18 +279,20 @@ def unflatten_MyContainer(
...
+jax.tree_util.register_pytree_with_keys(
+    MyContainer, flatten_with_keys_MyContainer, unflatten_MyContainer)
 
 jax.tree_util.tree_leaves([
     MyContainer('Alice', 1, 2, 3),
     MyContainer('Bob', 4, 5, 6)
 ])
 ```
 
+Note that registering both keys and values is a new feature, and you may see legacy pytrees that only has values registered (aka. registered via `register_pytree_node`). They are free to use most `tree_util` utilities except the `.*_with_path` APIs.
```
mattjj, line 291: > They are free to use most `tree_util` utilities except the `.*_with_path` APIs.

IIUC we have a decision to make here: do we make the _with_path APIs error when applied to pytrees with node types that don't opt into the extended API, or do we just handle those generically using flattened indices (i.e. the `FlattenedIndexKey` stuff)? If the latter, which I think is the current behavior of the code, we should change the second sentence not to imply that they're incompatible but just say they get a generic fallback.

mattjj, line 291: Saying "new feature" is a bit dangerous in docs because what if we leave this sentence in for a few years, and "new" no longer really means "new" :P

I think we should keep both variants documented, and tweak the language not to talk about the non-key path being "legacy" but rather the key-ful path being an "extended API".

docs/jax-101/05.1-pytrees.md
```diff
@@ -255,18 +279,20 @@ def unflatten_MyContainer(
...
+
 +++ {"id": "JgnAp7fFShEB"}
 
-Modern Python comes equipped with helpful tools to make defining containers easier. Some of these will work with JAX out-of-the-box, but others require more care. For instance:
+Modern Python comes equipped with helpful tools to make defining containers easier. Some of these will work with JAX out-of-the-box, but others require more care. For instance, `NamedTuple`:
```
mattjj, line 295:
```suggestion
Modern Python comes equipped with helpful tools to make defining containers easier. Some of these will work with JAX out-of-the-box, but others require more care. For instance, a `NamedTuple` subclass doesn't need to be registered to be considered a pytree node type:
```

mattjj (commented):
Some suggestions! TAL, WDYT?

docs/jax-101/05.1-pytrees.md
IvyZX: Done.

docs/jax-101/05.1-pytrees.md
IvyZX: Added - I didn't put type annotation in the doc to preserve the readability, but I added the docstring links (which contains type annotation and more) for each API calls.

docs/jax-101/05.1-pytrees.md
IvyZX: Added with a showcase paragraph.

docs/jax-101/05.1-pytrees.md
IvyZX: Done.

docs/jax-101/05.1-pytrees.md
IvyZX: Thanks for catching that! We do allow the fallback of _with_path API on non-keypath node types. I added a showcase paragraph.

IvyZX: Done.

docs/jax-101/05.1-pytrees.md
IvyZX: Done.

docs/jax-101/05.1-pytrees.md
mattjj: Good call!

mattjj (approved):
Looks great, thanks Ivy. Sorry for the review delay. But great work as always!

IvyZX:
> Looks great, thanks Ivy. Sorry for the review delay. But great work as always!

No worries Matt, thank you so much for reviewing!

PR #15389, opened 2023-04-04
Checkify: instrumentation context manager

LenaMartens:
Trying it out on `gather`.

notes:
- if you don't discharge an instrumented check, should you get an error? probably, but how? add an effect?
- `instrument` -> `catch`/`catch_error`

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from contextlib import contextmanager
 import dataclasses
 import functools
 import itertools as it
@@ -589,13 +590,14 @@ for _prim in nan_primitives:
 
 def gather_error_check(error, enabled_errors, operand, start_indices, *,
                        dimension_numbers, slice_sizes, unique_indices,
-                       indices_are_sorted, mode, fill_value):
+                       indices_are_sorted, mode, fill_value, check):
   out = lax.gather_p.bind(
       operand, start_indices, dimension_numbers=dimension_numbers,
       slice_sizes=slice_sizes, unique_indices=unique_indices,
-      indices_are_sorted=indices_are_sorted, mode=mode, fill_value=fill_value)
+      indices_are_sorted=indices_are_sorted, mode=mode, fill_value=fill_value,
+      check=check)
 
-  if OOBError not in enabled_errors:
+  if OOBError not in enabled_errors and not check:
     return error, out
 
   # compare to OOB masking logic in lax._gather_translation_rule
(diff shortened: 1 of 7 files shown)


jax/_src/checkify.py
```diff
@@ -589,13 +590,14 @@ def isnan(x):
...
 def gather_error_check(error, enabled_errors, operand, start_indices, *,
                        dimension_numbers, slice_sizes, unique_indices,
-                       indices_are_sorted, mode, fill_value):
+                       indices_are_sorted, mode, fill_value, check):
   out = lax.gather_p.bind(
       operand, start_indices, dimension_numbers=dimension_numbers,
       slice_sizes=slice_sizes, unique_indices=unique_indices,
-      indices_are_sorted=indices_are_sorted, mode=mode, fill_value=fill_value)
+      indices_are_sorted=indices_are_sorted, mode=mode, fill_value=fill_value,
+      check=check)
```
sharadmv, line 598: Should this be `False` since we're running the check?

jax/_src/lax/slicing.py
```diff
@@ -1244,15 +1249,17 @@ def _gather_fill(operand, indices, *, dimension_numbers, slice_sizes,
 
 def _gather_jvp_rule(g, operand, indices, *, dimension_numbers,
                      slice_sizes, unique_indices, indices_are_sorted, mode,
-                     fill_value):
+                     fill_value, check):
+  del check
   return gather(g, indices, dimension_numbers, slice_sizes,
                 unique_indices=unique_indices,
                 indices_are_sorted=indices_are_sorted, mode=mode,
                 fill_value=0)
```
sharadmv, line 1257: Why don't you add a `check=check` here?

jax/_src/checkify.py
LenaMartens: ah yeah good catch!

jax/_src/lax/slicing.py
sharadmv: I think this rule should call gather_p.join(...) and pass in check explicitly. That way we aren't relying on any global state.

If there's retracing happening then the context manager won't be active anymore so it's safer to just be explicit here.

LenaMartens: actually: wouldn't this cause issues when rules produce different primitives?

eg. `gather_transpose_rule` calls `scatter_add`. In this case `gather` and `scatter` share the same `check` set (`index_checks` checks both `gather`s and `scatter`s), but some primitives wouldn't. 

I guess the `check` arg cannot be a boolean, and needs to carry along the set of all active checks: then the primitive can be decomposed into different primitives by the transformation rules and those primitives will know if they need to be checked.

I also just realized that some (few) primitives can have multiple checks (eg. `scatter_add` has `index` and `nan` checks.

tests/checkify_test.py
```diff
@@ -1218,5 +1218,42 @@ def f(x):
...
+
+  def test_no_discharge_error(self):
+    self.skipTest("this fails")
+
+    @jax.jit
+    def f(x, i):
+      with instrument():
+        return x[i]
+
+    with self.assertRaisesRegex(ValueError, "Uncaught error!"):
```
LenaMartens, line 1245: I think we want to throw this error through effects. The instrumented checks don't call a `checkify` specific primitive (ie. no `check_p` is used), so not sure how to add that effect. Could call `check_p` there too for consistency?

LenaMartens: aha, through the primitives `abstract_eval` rule ofc!

jax/_src/lax/slicing.py
sharadmv: If a primitive in the fwd pass passes all its checks, should its primitives in the bwd pass also pass the checks?

LenaMartens: no don't think so, this is one of the examples in our documentation that only fails on the bwd (there's also no `mul` on the fwd)

```
def f(x):
 return x / (1 + jnp.sqrt(x))

grad_f = jax.grad(f)

err, _ = checkify.checkify(grad_f, errors=checkify.nan_checks)(0.)
print(err.get())
>> nan generated by primitive mul at <...>:3 (f)
```

PR #15390, opened 2023-04-04
checkify: catch OOB errors in dynamic_slice

jakevdp:
This will allow checkify tests to continue working properly for out-of-bounds indexing after #15377

### DIFF
--- a/tests/checkify_test.py
+++ b/tests/checkify_test.py
@@ -181,6 +181,23 @@ class CheckifyTransformTests(jtu.JaxTestCase):
     raises_oob(multi_idx, (5, -9), "index 5", axis0_msg)
     raises_oob(multi_idx, ((0, 9), 0), "index 9", axis0_msg)
 
+  def test_dynamic_slice_oobs(self):
+    def raises_oob(fn, x, idx, *expected_strs):
+      err, _ = checkify.checkify(jax.jit(fn), errors=checkify.index_checks)(x, idx)
+      error_txt = err.get()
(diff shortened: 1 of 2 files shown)


tests/checkify_test.py
```diff
@@ -181,6 +181,23 @@ def raises_oob(fn, idx, *expected_strs):
...
+      err, _ = checkify.checkify(jax.jit(fn), errors=checkify.index_checks)(x, idx)
+      error_txt = err.get()
+      self.assertIsNotNone(error_txt)
+      self.assertStartsWith(error_txt, "out-of-bounds indexing")
+      for s in expected_strs:
+        self.assertIn(s, error_txt)
+
+    x = jnp.ones((2, 3, 7))
+    raises_oob(partial(lax.dynamic_slice, slice_sizes=(1, 1, 1)), x, (2, 0, 0), 'index 2')
+    raises_oob(partial(lax.dynamic_slice, slice_sizes=(1, 1, 1)), x, (-3, 0, 0), 'index -1')
```
jakevdp, line 195: Negative indices are a bit weird: because `lax.dynamic_slice` normalizes negative indices before passing them to `dynamic_slice_p.bind()`, the index mentioned in the message doesn't match the index passed to `dynamic_slice`. But this is the same behavior as in `gather` (e.g. line 168 above), so I think it's working as expected.

LenaMartens (approved):
Thanks for fixing this, this LGTM!

PR #15426, opened 2023-04-06
fix custom_jvp round-tripping bug with symbolic zeros

### LINKED ISSUE #14833: custom JVP with symbolic zeros can unintentionally err under jit
From discussion in #14805.

Run this:
```python
from functools import partial
import jax

Zero = jax.custom_derivatives.SymbolicZero

@jax.custom_jvp
def f(x, y):
    return x * y

@partial(f.defjvp, symbolic_zeros=True)
def fjvp(primals, tangents):
    x, y = primals
    tx, ty = tangents
    assert type(tx) is not Zero or type(ty) is not Zero
    return f(x, y), (
        ty if type(tx) is Zero else
        tx if type(ty) is Zero else
        tx + ty)

print(jax.jacfwd(jax.jit(f))(0.1, 0.2))
```

See this:
```
TypeError: Value SymbolicZero with type <class 'jax._src.ad_util.SymbolicZero'> is not a valid JAX type
```

There is no such error if `not jax.config.jax_jit_pjit_api_merge`. It prints `1.0`.

mattjj:
fixes #14833 ~(when it's ready)~

~This is a draft I hacked on yesterday. It's not finished but I'm pushing it to share the basic idea while we iterate.~

We just weren't handling round-tripping to jaxpr and back correctly.

### DIFF
--- a/jax/_src/checkify.py
+++ b/jax/_src/checkify.py
@@ -907,17 +907,7 @@ def custom_jvp_call_rule(in_err, enabled_errors, *in_vals, num_consts,
                         call_jaxpr.consts, enabled_errors, err_tree))
   partial_checkify, f_metadata = _flatten_and_get_error_metadata_thunk(
       partial_checkify)
-
-  # Construct the default jvp function, without checkify-ing.
-  @lu.wrap_init
-  def jvp(*xs):
-    # TODO(lenamartens, sharadmv): why not checkify here?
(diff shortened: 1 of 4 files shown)


froystig (approved):

mattjj:
This was merged in 03e72e3b77cc32aaa740f340674154af4e81f693, though our bot doesn't seem to realize that.

PR #15432, opened 2023-04-06
[jax2tf] Refactor the gradient machinery for native serialization

gnecula:
In #15341 we have refactored jax2tf to separate the JAX and TF pieces
of the handling of gradients. Now we continue the refactoring and
we move the JAX-only pieces from jax2tf.py into jax_export.py. The
goal is to collect in jax_export all the pure JAX pieces needed
for serialization.

This is a pure refactoring, there should be no change in semantics.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -386,7 +386,7 @@ def convert(fun_jax: Callable,
           lowering_platform = native_serialization_platforms[0]
         else:
           lowering_platform = None
-        exported: Optional[jax_export.Exported] = jax_export.serialize_native(
+        exported: Optional[jax_export.Exported] = jax_export.export_native(
             fun_flat_jax, args_avals_flat,
             lowering_platform=lowering_platform,
             strict_checks=native_serialization_strict_checks)
@@ -394,8 +394,7 @@ def convert(fun_jax: Callable,
             args_flat_tf: Sequence[TfVal]
         ) -> Tuple[Tuple[TfVal, ...], Tuple[core.ShapedArray, ...]]:
           outs_tf, out_avals = run_exported_as_tf(
-              args_avals_flat, args_flat_tf, exported,  # type: ignore
-              native_serialization_strict_checks)
+              args_flat_tf, exported)  # type: ignore
           return outs_tf, out_avals
       else:
         dim_vars = shape_poly.all_dim_vars(args_avals_flat)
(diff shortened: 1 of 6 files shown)


jax/experimental/jax2tf/jax_export.py
```diff
@@ -211,10 +177,11 @@ def serialize_native(fun_jax: Callable,
...
-      mlir_module=mlir_module,
+      strict_checks=strict_checks,
       mlir_module_serialized=mlir_module_serialized,
       module_kept_var_idx=module_kept_var_idx,
-      xla_call_module_version=xla_call_module_version)
+      xla_call_module_version=xla_call_module_version,
+      _get_vjp=lambda exported: _export_native_vjp(wrapped_fun_jax, exported))
 
 
 def add_dim_arg_computation(module: mlir.ir.Module,
```
maxwillzq, line 187: My understanding is that this is internal API within jax_export API. can we rename "_add_dim_arg_computation"?
Same question for other functions name.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -427,6 +426,12 @@ def run_fun_flat_as_tf(
         _has_registered_tf_source_path = True
 
       if with_gradient:
+        if exported:
+          exported_vjp = exported.vjp()
+          if exported_vjp is None:
+            raise ValueError("Requested with_gradient but there is no VJP available")
```
maxwillzq, line 432: If we always request vjp() return Exported object,  in the definition, should we use "Optional["Exported"]]" ? Can we use "Exported" directly and move the exception raise there ?  When None return of `exported.vjp()` is used here?

maxwillzq (approved):

jax/experimental/jax2tf/jax_export.py
gnecula: Done. 

This is not yet as important because jax_export is private to jax2tf. But is still a good idea.

jax/experimental/jax2tf/jax_export.py
```diff
@@ -43,63 +44,71 @@
...
 
-def serialize_native(fun_jax: Callable,
-                     args_avals: Sequence[core.ShapedArray], *,
-                     lowering_platform: Optional[str],
-                     strict_checks: bool) -> Exported:
+def export_native(fun_jax: Callable,
+                  args_avals: Sequence[core.ShapedArray], *,
+                  lowering_platform: Optional[str],
+                  strict_checks: bool) -> Exported:
   arg_specs_jax = [
```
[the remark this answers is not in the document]
gnecula: Good idea, I added a whole docstring.

jax/experimental/jax2tf/jax2tf.py
gnecula: I think I understand your comment, we should move the error upstream. I have changed the handling of the "vjp()" to always return an Exported, or an exception. I think that this is in the spirit of your comment.

In truth, I do not yet understand how this will be working eventually. There is a string of several more refactorings in this area that I want to do incrementally to make it less likely that there is a regression.

PR #15457, opened 2023-04-07
[jax2tf] Refactor and simplify jax2tf.py

gnecula:
This is the next step in the refactoring and simplifying jax2tf. An additional goal is to move to jax_export.py as much pure-JAX functionalit as possible so that jax_export can start to become a standalone API for native serialization.

Previous PRs in this series:
  * #15258 and #15287: refactoring computation of dimension variables
  * #15341 and #15432: refactor handling of tf.custom_gradient (moved VJP serialization to jax_export)
  
Here we do the following:
  * we move the handling of in/out pytrees to jax_export. This allows the removal of a workaround that introduced a trampoline for the `lower` function for native lowering.
  * we move most of the code to set up the input abstract values into jax_export (`poly_spec` and `poly_specs`)
  * we separate out the implementation details for graph and native serialization into two classes `GraphSerializationImpl` and `NativeSerializationImpl`. This simplifies the core of the jax2tf.convert.

This is a pure refactoring, it should not change any existing behavior for jax2tf.

### DIFF
--- a/jax/experimental/jax2tf/jax2tf.py
+++ b/jax/experimental/jax2tf/jax2tf.py
@@ -84,6 +84,7 @@ from tensorflow.python.eager import context as tf_context  # type: ignore[import
 
 NameStack = source_info_util.NameStack
 PolyShape = shape_poly.PolyShape
+DType = Any
 
 # A temporary internal flag, to enable the wrapping of jax.jit functions
 # with tf.function(jit_compile=True). See #7389. This change has triggered a
@@ -111,7 +112,6 @@ def _sanitize_scope_name(name):
 # A value suitable in a TF tracing context: tf.Tensor, tf.Variable,
 # or Python scalar or numpy.ndarray. (A tf.EagerTensor is a tf.Tensor.)
 TfVal = Any
-DType = Any
 PrecisionType = int  # Enum xla_data.PrecisionConfig.Precision
 
 def _is_tfval(v: TfVal) -> bool:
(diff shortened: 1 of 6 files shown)


gnecula:
@maxwillzq @rxsang PTAL. This is a pretty big refactoring. I will land this when I return from vacation, after April 24th.

jax/experimental/jax2tf/jax2tf.py
```diff
@@ -334,135 +334,255 @@ def convert(fun_jax: Callable,
...
+
+  def setup_context(self):
+    self.exported = jax_export.export(
+        self.fun_jax,
+        lowering_platform=self.lowering_platform,
+        strict_checks=self.native_serialization_strict_checks
+    )(*self.args_specs, **self.kwargs_specs)
+
+  def restore_context(self):
+    pass
```
maxwillzq, line 489: Not fully understand setup_context and restore_context. My understanding is that they are reverse operations. Should we need add `del self.exported` here since we add it in setup_context ?

maxwillzq (commented):
Thanks for the effort to refactor jax2tf.
LGTM.

jax/experimental/jax2tf/jax2tf.py
gnecula: I have changed the name to `before_conversion` and `after_conversion`. The goal is to be able to setup/restore global state that may be needed. We do not need to restore changes to `self`

jax/experimental/jax2tf/jax_export.py
```diff
@@ -105,103 +116,198 @@ def vjp(self) -> "Exported":
...
+    elif "shards" in lowering.compile_args:  # for PmapComputation
+      out_avals_flat = lowering.compile_args["shards"].out_sharded_avals
+    else:
+      out_avals_flat = lowered.compile_args["out_avals"]
+
+    # Log and then check the module.
+    if logging.vlog_is_on(3):
+      mlir_module_text = mlir.module_to_string(mlir_module)
+      logmsg = f"version={xla_call_module_version} lowering_platform={lowering_platform}"
+      logging.vlog(3, "Lowered JAX module: %s\n%s", logmsg, mlir_module_text)
```
rxsang, line 285: Not related to this PR, but sometimes the MLIR module can be too large to print. Maybe consider dumping it to a file https://github.com/google/jax/blob/main/jax/_src/dispatch.py#L481-L482?

jax/experimental/jax2tf/jax_export.py
```diff
@@ -105,103 +116,198 @@ def vjp(self) -> "Exported":
...
+  """Constructs a jax.ShapeDtypeStruct with polymorphic shapes.
 
-def export_native(fun_jax: Callable,
-                  args_avals: Sequence[core.ShapedArray], *,
-                  lowering_platform: Optional[str],
-                  strict_checks: bool) -> Exported:
-  """Exports native serialization for a JAX function.
+  Args:
+    arg_shape: the shape, with possibly some unspecified dimensions.
+    arg_dtype: the jax dtype.
```
rxsang, line 131: jax->JAX

rxsang (approved):

jax/experimental/jax2tf/jax_export.py
gnecula: Good idea, I will do this in a future change

